PR #2, opened 2016-03-29
CSS

jasonLaster:
Creates a new `styles.css` bundle, which is loaded in the index.html

+ debugger component styles are loaded relatively `require('./source-panel.css')`
+ shared components are loaded relative to the themes directory `require('themes/h-split-box.css')`

The PR is bloated a bit by the getting started PR


jlongster:
Shoot, didn't realize I already made this conflict :( Using `ExtractTextPlugin` sounds good, though we need to figure out how this will work in the Firefox environment.

I just made significant progress with a WebSocket proxy to a live remote debugging environment. I'm going to go ahead and push it out because I'm going to send an email to the team right now and will point here as an example. Sorry for not merging this earlier, I kept meaning to discuss CSS more with you first.

README.md
```diff
@@ -2,11 +2,21 @@
...
 $ npm install
-$ npm start
 ```
 
+### 2. Add Environment file
+
+You'll need to add an `environment.json` file to set your path to the gecko repository.
+
+```json
+{ "geckoPath" : "/Users/jlaster/src/mozilla/gecko-dev"}
```
jasonLaster: Does this make sense to you?

I'm thinking other things like logging preferences, hot reloading could go here

webpack.config.js
```diff
@@ -9,7 +11,8 @@ module.exports = {
   resolve: {
     alias: {
       "devtools": "ff-devtools-libs",
-      "sdk": "ff-devtools-libs/sdk"
+      "sdk": "ff-devtools-libs/sdk",
+      "themes": environment.geckoPath + "devtools/client/themes"
```
jasonLaster: It seems silly to add gecko as a dependency just for some css, but we'll probably need it for something soon and our target audience has gecko... What do you think?

README.md
jlongster: Hm, I posted here before but github lost my comment. Their service went down earlier today.

I don't think this config should one of the steps in the readme; I'd rather it be an optional footnote. So far we've gotten away with not needing any config, and it's one less step for someone to get up and running. What do you think of moving this to the bottom under a "Config" or "Advanced" section?

webpack.config.js
jlongster: I'd say either make it optional (somehow conditionally pull in themes), or we should just copy themes into `ff-devtools-libs`. Those files are super easy to copy and don't require any changes, so will be easy to stay up-to-date. I don't love the idea of depending on gecko because I'd like to eventually send this around to people to feedback, and it would be great if they could just fire it up.

PR #4, opened 2016-04-05
Add Codemirror component

jasonLaster:
This is an MVP version, I didn't try to think about whether this would work with our redux setup and whether these were patterns we would like, just wanted to get something working and try selecting source. Sharing early so you can weigh in.

This depends on the CSS Loader to load the codemirror.css


jlongster:
Sweet!

PR #6, opened 2016-04-06
Remove Firefox Dependency

jasonLaster:
This patch removes the firefox dependency:

+ moves split-box.css into lib
+ removes the webpack.config dependency on environment.json
+ updates the readme so environment.json is optional

It also:
+ adds an environment.json.sample file


jlongster:
Nice!

PR #8, opened 2016-04-07
Add a Tabs landing page

jasonLaster:
The tabs landing page shows the tabs that are ready to be debugged.
Currently, you re-select the tab you want to debug each time you
refresh, but we'll hopefully make it possible to select a tab and
refresh, without losing your selection shortly. I think the most natural
approach is to add a hash url.

http://recordit.co/otL3a1ly8C

Changes:

There's a new Toolbox component, Tabs component, action, and reducer.
At some point it would be nice to pull these pieces out of the debugger.

The protocol logic was moved into a thread-client module, which exposes
connectToClient and connectToTab methods. This is probably the most
important thing to review.

### DIFF
--- a/js/actions/tabs.js
+++ b/js/actions/tabs.js
@@ -1,0 +1,42 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+"use strict";
+const { Task } = require('devtools/sham/task');
+const { PROMISE } = require('devtools/client/shared/redux/middleware/promise');
+
+const { connectToTab} = require('../client');
+const constants = require('../constants');
+const { getTabs } = require('../queries');
+
+function newTabs(tabs) {
+  return {
+    type: constants.ADD_TABS,
+    value: tabs
+  };
+}
+
+function selectTab({ tabActor }) {
+  return (dispatch, getState) => {
+    const tabs = getTabs(getState());
+    const selectedTab = tabs[tabActor];
+
+    // set selected tab in the URL hash
+    const childId = tabActor.match(/child\d+/)[0];
+    window.location.hash = `tab=${childId}`
+
+    return dispatch({
+      type: constants.SELECT_TAB,
+      tabActor: tabActor,
+      [PROMISE]: Task.spawn(function*() {
+        yield connectToTab(selectedTab);
+        return { selectedTab };
+      })
+    });
+  }
+}
+
+module.exports = {
+  newTabs,
+  selectTab
+};
--- a/js/components/Tabs.js
+++ b/js/components/Tabs.js
@@ -1,0 +1,46 @@
+const React = require("react");
+const ReactDOM = require("react-dom");
+const { bindActionCreators } = require("redux");
+const { connect } = require("react-redux");
+const { getTabs } = require("../queries");
+const actions = require("../actions")
+
+require("./tabs.css");
+const dom = React.DOM;
+
+function Tabs({tabs, selectTab, loadSources}) {
+  const tabsArr = Object.keys(tabs).map(k => tabs[k]);
+
+  /**
+   * TODO: this click handler is probably doing too much right now.
+   */
+  function onClickTab(e) {
+    selectTab({ tabActor: e.currentTarget.dataset.actorId })
+      .then(loadSources)
+      .then(() => {
+        gThreadClient.addListener("newSource", (event, packet) => {
+          store.dispatch(actions.newSource(packet.source));
+        });
+        gThreadClient.addListener("paused", (_, packet) => {
+          console.log(packet);
+        });
+      })
+  }
+
+  return dom.ul(
+    {className: 'tabs'},
+    tabsArr.map(tab => {
+      return dom.li({ 'className': 'tab',
+                      'data-actor-id': tab.actor,
+                      'onClick': onClickTab },
+        dom.div({ className: 'tab-title' }, tab.title),
+        dom.div({ className: 'tab-url' }, tab.url)
+      )
+    })
+  );
+}
+
+module.exports = connect(
+  state => ({ tabs: getTabs(state) }),
+  dispatch => bindActionCreators(actions, dispatch)
+)(Tabs);
--- a/js/main.js
+++ b/js/main.js
@@ -4,51 +4,60 @@ const { combineReducers, applyMiddleware, bindActionCreators } = require('redux'
 const { Provider } = require('react-redux');
 const configureStore = require('./create-store');
 const reducers = require('./reducers');
+const { connectToClient } = require('./client');
 const actions = require('./actions');
-const App = require('./components/app');
 const dom = React.DOM;
-
-const { DebuggerClient } = require('devtools/shared/client/main');
-const { DebuggerTransport } = require('devtools/transport/transport');
-const { TargetFactory } = require("devtools/client/framework/target");
-const socket = new WebSocket("ws://localhost:9000");
-const transport = new DebuggerTransport(socket);
-const client = new DebuggerClient(transport);
-
-client.connect().then(() => {
-  return client.listTabs().then(response => {
-    const tab = response.tabs[response.selected];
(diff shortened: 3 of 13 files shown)


js/components/Tabs.js
```diff
@@ -0,0 +1,38 @@
...
+
+require("./tabs.css");
+const dom = React.DOM;
+
+function Tabs({tabs, selectTab, loadSources}) {
+  const tabsArr = Object.keys(tabs).map(k => tabs[k]);
+
+  function onClickTab(e) {
+    selectTab({ tabActor: e.currentTarget.dataset.actorId })
+      .then(loadSources);
```
jasonLaster: When you select a tab, you're doing two things: attaching to the threadClient and loading sources. Chaining these actions in the click handler seemed a little dirty, but better than other alternatives. For example, I tried to dispatch the loadSources action from within the selectTab action, but it seemed bad for Tab actions to know about Sources actions.

js/actions/sources.js
```diff
@@ -66,6 +66,14 @@ function loadSources() {
     [PROMISE]: Task.spawn(function*() {
       const response = yield gThreadClient.getSources();
 
+      gThreadClient.addListener("newSource", (event, packet) => {
+        store.dispatch(actions.newSource(packet.source));
```
jasonLaster: These listeners used to be setup in `main.js`, which had access to the store and actions. 

It seemed like too much coupling for `thread-client` to have access to the store and actions, so I moved them here. Thoughts?

js/components/Toolbox.js
```diff
@@ -0,0 +1,28 @@
...
+const Tabs = React.createFactory(require('./tabs'));
+const { getSelectedTab } = require('../queries');
+
+
+function isEmpty(obj) {
+  return Object.keys(obj).length == 0;
+}
+
+function Toolbox({ selectedTab }) {
+  const container =  isEmpty(selectedTab) ? Tabs() : App();
```
jasonLaster: Is there a better way to see if a seemless-immutable object is empty?

js/main.js
```diff
@@ -4,51 +4,26 @@ const { combineReducers, applyMiddleware, bindActionCreators } = require('redux'
...
-  threadClient.addListener("newSource", (event, packet) => {
-    store.dispatch(actions.newSource(packet.source));
-  });
-
-  threadClient.addListener("paused", (_, packet) => {
-    console.log(packet);
-  });
-
-  store.dispatch(actions.loadSources())
-}
```
jasonLaster: It's so nice to see red in a bootstrap file like main.js. With that said, I tried to not lock us into a solution with too rigid an API.

js/components/Toolbox.js
```diff
@@ -0,0 +1,28 @@
+const React = require("react");
```
jasonLaster: The name `toolbox` is somewhat of a placeholder because that's a fairly well defined word. Do you think it works here?

js/thread-client.js
```diff
@@ -0,0 +1,34 @@
...
+}
+
+function connectToTab(tab, onNewSource) {
+  let deferred =  promise.defer();
+  const options = { client, form: tab, chrome: false };
+
+  TargetFactory.forRemoteTab(options).then(target => {
+    target.activeTab.attachThread({}, (res, threadClient) => {
+      threadClient.resume();
+      window.gThreadClient = threadClient;
```
jasonLaster: I tried three different hacks for exporting `gThreadClient`, which would let us do something like:

```js
const {gThreadClient} = require('../thread-client');

function loadSources() {
  gThreadClient.getSources();
}
```

1.  exporting a reference to the threadClient
  + `module.exports.gThreadClient = threadClient`  
  + webpack would get a reference to the module before the exports property had been set
2. exporting a stub that received threadClient properties
  + `Object.assign(gThreadClient, threadClient, threadClient.__proto__)`
  + webpack would get a reference to an object with all of the properties except the threadClient's prototype getters, which we could get if we did a slightly better deep clone than Object.assign
3. exporting a getter
  + `module.exports = {get gThreadClient () {return threadClient}}`
  + webpack would get a reference to threadClient before it was set.

Ultimately, I decided to leave `gThreadClient` a global for the time being. :)

jasonLaster:
I rebased after Helen's PR and incorporated some of the styles. Looks much better.

js/actions/sources.js
jlongster: `loadSources` is an action that means "get all the sources for the page". It definitely should not install listeners. There's no reason I can't call `loadSources` multiple times, and the current debugger does just that when switching sourcemaps on/off. I'd rather keep everything in `main.js` for now and next week we can clean up the protocol part.

I don't think it's too much coupling. For things outside the react/redux system, it's a simple pattern for them to "hook up" to it by directly dispatching actions on the store. It makes it very clear what "drives" the system. An initial action (not one dispatched from within an action creator) can *only* be dispatched from two places: a UI event from a component, or a network event from some external source. I think it's nice to be able to see how the external source "binds" itself to the system: in this one file, all of the network events map to these actions.

The core issue is that actions can be called multiple times, so you don't want to have side effects in there that shouldn't happen multiple times.

js/actions/tabs.js
```diff
@@ -0,0 +1,44 @@
...
+const constants = require('../constants');
+const { getTabs } = require('../queries');
+
+function newTabs(tabs) {
+  return dispatch => {
+    return dispatch({
+      type: constants.ADD_TABS,
+      value: tabs
+    });
+  };
```
jlongster: You aren't doing anything async, so you should just return the action:

```js
return {
  type: constants.ADD_TABS,
  value: tabs
}
```

js/components/Tabs.js
jlongster: Yeah, it's generally not good to tie down actions like that, so I think this is fine. I don't think it's really that dirty: it's not uncommon to want to hook up a workflow like this, and that's why any async actions that use the promise middleware return promises when they are dispatched. It allows you to do stuff like this.

js/components/Tabs.js
```diff
@@ -0,0 +1,38 @@
...
+
+  return dom.ul(
+    {className: 'tabs'},
+    tabsArr.map(tab => {
+      return dom.li(
+        {
+          'className': 'tab',
+          'data-actor-id': tab.actor,
+          'onClick': onClickTab
+        },
```
jlongster: What do you think about this style? It removes some vertical noise.

```js
 {  'className': 'tab',
    'data-actor-id': tab.actor,
    'onClick': onClickTab }
```

js/components/Toolbox.js
jlongster: `Toolbox` might be too confusing. It implies a toolbar where you can select multiple tools. I'd prefer a different name, but I don't know what. `TabList`?

js/components/Toolbox.js
jlongster: seamless-immutable objects are normal JS objects. It just freezes them and augments them with a few more methods. But they are totally normal JS objects.

I think we should bite the bullet and switch everything to immutable.js though. I've had some troubles with seamless-immutable: https://gist.github.com/jlongster/028c83ed09d99c09983a#seamless-immutable

js/thread-client.js
jlongster: Yeah, sounds good. Let's talk next week about this.

Can you rename the file to "client.js" or something like that though? Because it creates the entire connection, not just the thread client. Eventually we will need the `DebuggerClient` for some things too, I think. You could name this `connection.js`, I dunno.

jlongster:
This looks good, but needs a few tweaks. The main thing is to avoid setting up listeners in the action. The rest are just minor nits.

js/components/Tabs.js
jasonLaster: That works for me.

js/actions/tabs.js
jasonLaster: nice

jasonLaster:
should we rename thread-client to client.js or move it back to main.js?

js/actions/sources.js
jasonLaster: got it

jasonLaster:
I'll rename thread-client to client and we can figure out the better approach next week.

PR #11, opened 2016-04-07
made thing more beautiful

helenvholmes:
:metal:


jasonLaster:
looks great!

One comment, I'm not sure if images should be inside of components. That makes me think that there is an images component. :wink: I'm not sure if it's an option to move images outside of `js`, if so that would be idea, if not, maybe it makes sense for images to be next to components.

jlongster:
Let's definitely move the images out. Some things in here are done quickly without too much thought, and we'll clean it up.

I was able to add the accordion that I wrote a few months ago but that probably needs some tweaks too.

helenvholmes:
Sorry about the images :sweat_smile: that was definitely a, "uh uh uh where should these go okay here seems good"-sort of deal.

Issue #9, opened 2016-04-07
2-space CSS


jlongster:
My editor is set to 4-space but we want to conform to the current devtools style which is 2 space

PR #12, opened 2016-04-08
Alphabetized + two-space'd, #9

helenvholmes:


jasonLaster:
:smiley_cat: 

can we lint for this stuff going forward?

jlongster:
I don't know much about CSS linting, but the devtools doesn't have that. We can do it for our project thought

helenvholmes:
The reason I understood we don't have CSS-linting in devtools isn't that it's not a good idea, but hat it would both take time to implement and that it would be going off like crazy in our highly specific, XUL/CSS world.

Projects like this will make CSS-linting more and more possible, so I think we should use it.

jlongster:
Yay!

PR #14, opened 2016-04-08
Add CSS Linter

jasonLaster:
[Stylelint](http://stylelint.io/) is a popular CSS linter with lots of [rules](http://stylelint.io/user-guide/rules/), which will let us keep the CSS clean.

I added a simple config, which checks for bad hex, 2 space indentation, and a couple other small knits. There's a good [example](https://github.com/stylelint/stylelint-config-standard/blob/master/index.js) with many more rules we can add in the future.

Questions:
+ are we in consensus that we want a linter?
+ not a blocker, but would we like to add linting on PRs with a service like travis?
+ what kind of rules would like to add?

Example:

I ran stylelint on our codebase and got this [gist](https://gist.github.com/536309bfd7d8f17bb23e47abfc6f9ba7), which is an argument for adding a linter sooner rather than later.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,8 @@
 {
   "name": "debugger-experiment",
   "scripts": {
-    "start": "firefox-proxy & ./node_modules/.bin/webpack --watch"
+    "start": "firefox-proxy & ./node_modules/.bin/webpack --watch",
+    "lint-css": "stylelint js/components/*.css"
   },
   "dependencies": {
     "co": "=4.6.0",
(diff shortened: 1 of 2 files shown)


jlongster:
Those should be fixed with Helen's PR, and I need to change my Emacs to default to 2 spaces. I'm fine with a linter! I'm not convinced about alphabetizing the rules but if you and Helen want it it's fine. I don't have a way to automatically do that in Emacs.

helenvholmes:
I normally will format a block as I write it with function+F5—if there's an automatic way to do it in Sublime I don't know it either, but I'm still all for alphabetizing CSS properties. Think it makes things a lot more findable within complicated selectors.

jlongster:
I noticed that you're formatter would automatically alphabetize them as well. Mine automatically "formats" it but it doesn't rewrite the whole thing to alphabetize it.

helenvholmes:
Is this the one time Emacs has failed us??

jlongster:
THE ONE TIME

jlongster:
Merging!

PR #21, opened 2016-04-08
Changed imports to correct case

ToucheSir:
Runs like a charm on Linux now.


jlongster:
Awesome, thanks!

ToucheSir:
Excited to see how the new frontend develops!

jlongster:
Thanks! I'm trying to think of ways to talk about it. Will blog and probably do other things too.

PR #23, opened 2016-04-08
Eslint

jasonLaster:
I set up eslint with airbnb's standard config, which I believe is one of the more common sets of rules. Airbnb include some nice es6 and react rules. The thinking is that, we could always override some of those rules in our own config, which I started to do.

I cleaned up most of the component errors, but there are some interesting errors with prop validation and es6 classes that we should consider fixing. Lets spend some time next week fixing up the remaining directories.

Here's the output [gist](https://gist.github.com/71ff90999914570f8d34aa8e61f95bbf).

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,0 +1,3 @@
+{
+  "extends": "airbnb"
+}
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,3 +1,9 @@
 {
-  "extends": "airbnb"
+  "extends": "airbnb",
+  "rules": {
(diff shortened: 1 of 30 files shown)


jlongster:
We already have an esline rule in the devtools directory in gecko-dev. We should definitely use that, we've already put a lot of time deciding the exact rules that we want, and much of our code already uses it.

jasonLaster:
Good call.

jasonLaster:
Closing in favor of #28.

PR #26, opened 2016-04-08
add test runner

jlongster:
This adds a test runner.

It's pretty janky. We need to compile each test with webpack which makes things a lot harder. Currently we compile everything as a separate entry point with a single webpack command in to the `build/tests` directory, and invoke the xpcshell runner on that. The `build/tests` directory is recursively removed each run to make sure it's fresh.

It works. We'll have to use it a little bit to see if it works well enough. All we need is something that will run for 3 months until we merge it.

I don't think mocha is a good idea because we will already have a lot of to when merging our code back in. I think it would be just too obtrusive to the merge, but we can investigate other solutions.

**Don't merge this yet**, because I want to test it some more. Just wanted to put it up here now.

### DIFF
--- a/bin/test
+++ b/bin/test
@@ -19,11 +19,14 @@ const allPaths = paths.reduce((acc, path) => {
     const files = fs.readdirSync(path);
     return acc.concat.apply(
       acc,
-      files.map(file => join(path, file))
+      files.filter(p => p[0] !== '.').map(file => join(path, file))
     );
   }
   else {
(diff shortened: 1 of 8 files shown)


jasonLaster:
nice work.

jlongster:
Ready to merge!

Run `./bin/test <path>...` to test code. We could add an npm script too I guess, but I'll probably never use it...

PR #27, opened 2016-04-10
fix breakpoint linenums - off by one

marfarma:
Breakpoints list reported with line numbers off by one


jasonLaster:
I believe this is necessary because CodeMirror considers the first line in the document to be 0.

jlongster:
Cool!

PR #29, opened 2016-04-10
Editor styles

matt-oconnell:
Adding styles for debugger/source code viewing

### DIFF
--- a/js/components/Editor.css
+++ b/js/components/Editor.css
@@ -1,10 +1,10 @@
 .CodeMirror {
-    height: 100%;
+  height: 100%;
 }
 
 .editor.breakpoint {
-    fill: var(--theme-selection-background);
-    width: 40px;
+  fill: var(--theme-selection-background);
+  width: 40px;
 }
 
 /* vim:set ts=2 sw=2 sts=2 et: */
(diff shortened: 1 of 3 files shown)


js/components/Editor.css
```diff
@@ -5,4 +5,346 @@
 .editor.breakpoint {
   fill: var(--theme-selection-background);
   width: 40px;
+}
+
+/* vim:set ts=2 sw=2 sts=2 et: */
```
jasonLaster: the style comments should be on the top of the file

js/components/Editor.css
```diff
@@ -5,4 +5,346 @@
...
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/* todo: maybe add later? */
+/*@import url(variables.css);*/
+/*@import url(common.css);*/
+/*@import url(toolbars.css);*/
```
jasonLaster: @matt-oconnell did you delete the other import/url references in the file? If so, could you comment them out instead so it's clear what we need to move over to have the full functionality

js/lib/variables.css
```diff
@@ -14,7 +14,7 @@
  * so the formatting should be consistent (i.e. no '}' inside a rule).
  */
 
-:root {
+:root .cm-s-light {
```
jasonLaster: does this scoping affect the rest of the editor styling? If so, would it make sense to use a more generic classname like `devtools-light-theme`? Should we be using [variables.css](https://dxr.mozilla.org/mozilla-central/source/devtools/client/themes/variables.css) directly?

js/components/Editor.css
matt-oconnell: Fixed in updated commit.

matt-oconnell:
@jasonLaster I've addressed your line notes in the most recent commit.

The deleted `common` and `toolbars` imports will be necessary eventually but I think they're beyond the scope of this PR.

I've synced up the theme colors to the [DevTool color docs](https://developer.mozilla.org/en-US/docs/Tools/DevToolsColors). I've kept the dark theme in there just in case. Adding the `devtools-dark-theme` class to `html` will toggle this theme on.

jasonLaster:
@matt-oconnell the wiki is out of date, could you remove those changes so that the styles are the same as what was in firefox?

helenvholmes:
@matt-oconnell just parroting what @jasonLaster said—if you roll back the changes in ```variables.css```, this looks good to go to me!

jlongster:
Thanks a lot for doing this! I will go ahead and merge this but revert the variables changes by hand.

matt-oconnell:
@jlongster Cool, glad I could help.

PR #30, opened 2016-04-10
Update breakpoint styles

NColey:
Updated the breakpoint styles to show the filename (instead of the actor) and the line number where the breakpoint was placed. Also made the text color grey, removed the bullet points, and made the text size smaller for the breakpoints.

### DIFF
--- a/js/components/Breakpoints.js
+++ b/js/components/Breakpoints.js
@@ -1,24 +1,40 @@
 const React = require("react");
 const dom = React.DOM;
+const { connect } = require("react-redux");
+const { getSources } = require("../queries");
 
-function Breakpoints({ breakpoints }) {
+require("./Breakpoints.css");
+require("../lib/variables.css");
 
-  function onResumeClick() {
-    console.log('click')
-    gThreadClient.resume();
-  }
+const Breakpoints = React.createClass({
 
-  return dom.div(
-    {},
-    dom.button({ onClick: onResumeClick }, 'resume'),
-    dom.ul(
-      null,
-      breakpoints.map(bp => dom.li(
+  render() {
+    function onResumeClick() {
+      console.log('click')
+      gThreadClient.resume();
+    }
+
+    function getFilenameFromSources(sources, actor) {
+      var source = sources[actor];
+      var url = new URL(source.url);
+      var filename = url.pathname.substring(url.pathname.lastIndexOf('/')+1);
+      return filename;
+    }
+
+    return dom.div(
+      {className: 'breakpoints'},
+      dom.button({ onClick: onResumeClick }, 'resume'),
+      dom.ul(
         null,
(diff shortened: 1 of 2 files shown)


js/components/Breakpoints.js
```diff
@@ -1,24 +1,40 @@
...
+      gThreadClient.resume();
+    }
 
-  return dom.div(
-    {},
-    dom.button({ onClick: onResumeClick }, 'resume'),
-    dom.ul(
-      null,
-      breakpoints.map(bp => dom.li(
+    function getFilenameFromSources(sources, actor) {
```
jasonLaster: I wonder if it'd be helpful to move some of the breakpoint source data collection to queries.js. So, instead of getting sources in the breakpoints component and formatting it with `URL`, you could get breakpoint props with a formatted pathname, which would be useful here and maybe other contexts as well.

js/components/Breakpoints.js
```diff
@@ -1,24 +1,40 @@
 const React = require("react");
 const dom = React.DOM;
+const { connect } = require("react-redux");
+const { getSources } = require("../queries");
 
-function Breakpoints({ breakpoints }) {
+require("./Breakpoints.css");
+require("../lib/variables.css");
```
jasonLaster: This file is required in other places, so it doesn't need to be imported here.

js/components/Breakpoints.js
```diff
@@ -1,24 +1,40 @@
...
 
-  function onResumeClick() {
-    console.log('click')
-    gThreadClient.resume();
-  }
+const Breakpoints = React.createClass({
```
jasonLaster: Breakdpoints was refactored to be a react class because it needed to use connect, does that make sense? Also, should we be transitioning to es6 class syntax e.g. `class Breakpoints extends React.Component { }`

jlongster: I don't love the class syntax, honestly. You lose auto-bound methods and I don't see much gain. We can still use the `foo() { ... }` syntax for declaring methods in objects, and we probably should start doing that. The only other syntax difference is `class` instead of `createClass`. I forget exactly but I think there were one or two other gotchas with ES6 classes.

(ES6 classes are also not optimized in SpiderMonkey yet, which is a big deal)

js/components/Breakpoints.js
jlongster: Yeah, that makes sense. Also, we have utility functions for parsing URLs. But I'm fine merging this and tweaking it later. I'll leave it up to you whether you want to merge this now or wait for this to be moved into `queries.js`.

js/components/Breakpoints.js
jasonLaster: Cool. That's why i asked.

jasonLaster:
Thanks @NColey - merging now!

PR #32, opened 2016-04-11
Cleanup code based on eslint

jasonLaster:
Here are some additional fixups to get our linting to pass.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -27,11 +27,9 @@
     "XPCOMUtils": true,
     "_Iterator": true
   },
-  "parserOptions": {
-      "ecmaVersion": 6
-  },
   "env": {
-      "es6": true
+      "es6": true,
+      "browser": true,
   },
   "rules": {
     // These are the rules that have been configured so far to match the
@@ -355,7 +353,7 @@
     // Require a space immediately following the // in a line comment.
     "spaced-comment": [2, "always"],
     // Require "use strict" to be defined globally in the script.
-    "strict": [2, "global"],
+    "strict": [0, "global"],
     // Disallow comparisons with the value NaN.
     "use-isnan": 2,
     // Warn about invalid JSDoc comments.
--- a/js/actions/event-listeners.js
+++ b/js/actions/event-listeners.js
@@ -1,7 +1,8 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
-/* global window gThreadClient setNamedTimeout */
+/* global window gThreadClient setNamedTimeout services EVENTS */
+/* eslint no-shadow: 0  */
 "use strict";
 
 const constants = require("../constants");
@@ -87,8 +88,8 @@ const _getListeners = Task.async(function* () {
   return listeners;
 });
 
-const _getDefinitionSite = Task.async(function* (aFunction) {
-  const grip = gThreadClient.pauseGrip(aFunction);
+const _getDefinitionSite = Task.async(function* (func) {
+  const grip = gThreadClient.pauseGrip(func);
   let response;
 
   try {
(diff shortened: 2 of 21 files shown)


.eslintrc
```diff
@@ -31,7 +31,8 @@
       "ecmaVersion": 6
   },
   "env": {
-      "es6": true
+      "es6": true,
+      "browser": true,
```
jasonLaster: this allows more globals to be defined (document, window...)

.eslintrc
```diff
@@ -355,7 +356,7 @@
     // Require a space immediately following the // in a line comment.
     "spaced-comment": [2, "always"],
     // Require "use strict" to be defined globally in the script.
-    "strict": [2, "global"],
+    "strict": [0, "global"],
```
jasonLaster: this is because eslint doesn't know we're using a module system...

js/actions/event-listeners.js
```diff
@@ -1,7 +1,8 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
-/* global window gThreadClient setNamedTimeout */
+/* global window gThreadClient setNamedTimeout services EVENTS */
+/* eslint no-shadow: 0  */
```
jasonLaster: helps with a dispatch call on line 34, where dispatch is shadowing the outer define

js/actions/event-listeners.js
```diff
@@ -87,8 +88,8 @@ const _getListeners = Task.async(function* () {
   return listeners;
 });
 
-const _getDefinitionSite = Task.async(function* (aFunction) {
-  const grip = gThreadClient.pauseGrip(aFunction);
+const _getDefinitionSite = Task.async(function* (func) {
+  const grip = gThreadClient.pauseGrip(func);
```
jasonLaster: was hungarian formatting

js/components/Tabs.js
```diff
@@ -1,5 +1,6 @@
+/* globals gThreadClient */
```
jasonLaster: was tempted to add this to the list of globals in eslintrc, but figured, we'd be refactoring soon

js/reducers/tests/sources.js
```diff
@@ -1,3 +1,7 @@
+/* eslint brace-style: 0 */
+/* globals equal */
+/* eslint camelcase: 0 */
+/* eslint no-unused-vars: 0 */
```
jasonLaster: for run_test

.eslintrc
```diff
@@ -27,11 +27,9 @@
     "XPCOMUtils": true,
     "_Iterator": true
   },
-  "parserOptions": {
-      "ecmaVersion": 6
-  },
```
jasonLaster: dropped this because it wasn't necessary.

js/reducers/async-requests.js
```diff
@@ -1,3 +1,5 @@
+/* eslint brace-style: 0 */
```
jasonLaster: dropping these in favor of consistent styles. might be a bit ugly, but that's okay

js/actions/event-listeners.js
jlongster: Yeah, I don't care too much about that rule. But if the devtools eslint forbids shadowing I guess we should keep it.

js/reducers/tests/sources.js
jlongster: It'd be really nice to have an `.eslint` in all test directories that automatically define the globals so that we don't need this at the top of every test file.

.eslintrc
jlongster: This makes it so that we don't need "use strict"? How does the module system affect that?

We should probably keep "use strict"; at the top of all the files even if we could automatically do that in our build system, so that the code works the same way when we merge.

Sorry if I'm mis-interpreting, I'm not too familiar with eslint so it's a little hard to read all these tweaks.

jasonLaster: happy to do that.

jasonLaster:
Pushed the existing commit to master so that we can avoid further merge conflicts and will cleanup the 'use static' stanzas in a bit.

PR #35, opened 2016-04-11
add '|| true' to lint commands to hush the npm ERR

clarkbw:


clarkbw:
Doesn't work well with travis, need to find another way.

Issue #31, opened 2016-04-10
Discuss adding lodash


jasonLaster:
I'd like to discuss the merits of adding [lodash](https://lodash.com) for debugger.html. 

reasons to do it:
+ readability - lodash's utility functions like (groupBy and isEmpty) could dry up the code considerably ([1](https://github.com/jlongster/debugger.html/blob/master/js/reducers/sources.js#L28-L31), [2](https://github.com/jlongster/debugger.html/blob/master/js/components/TabList.js#L10-L12))
+ quality - John-David Dalton is managing the project full time at Microsoft and has done a great job of creating a consistent API, with test coverage, performance metrics, and documentation. Also, many of his APIs are informing the standards bodies. 

arguments against doing it:
+ lodash would add weight. Lodash minified is relatively small. If at somepoint, we adopt a build system like webpack, only the functions used would be bundled. For example, if we only required chunk, `chunk = require('lodash/chunk')` only chunk and its dependencies would be added to the build.
+ it is easier to build what you need. In practice, I've found that this practice often leads to more bike-shedding, where there is a debate over each new functional bit added. Over time, the internal utilities are often not as good in terms of API design or extendability, as what is available.

Path forward:
+ use lodash it for the project
+ try lodash with the understanding that if we decide not to keep it, we could easily replace the call sites
+ use a subset of core array and object functions

jlongster:
I'd rather not because we'd have to pull all of it in when we merge to devtools. If we want to look into this we should bring it up with the whole team. While we are operating as a separate project technically, as you know we're still heavily tied to whatever gecko-dev/devtools provides.

The other reason is that the utility will be reduced a lot when we switch to immutable.js, which I'm very convinced we need to do. We won't be able to use it with immutable.js, and is why things like transducers are better (they aren't tied down to the data structure like lodash is).

The current wrote was focusing on shipping (which may ore may not be a bad thing). There are certainly places that need to be cleaned up and we will do that. I don't think we need lodash for that, personally. Much of the utility we would get will be provided by immutable.js' API.

I think we should put our efforts into migrating to immutable.js first and just augmenting any util functions we need.

jasonLaster:
Thanks, that makes a lot of sense. I agree, that this decision would be a team decision. Also, after reviewing today, I think we can clean up our code with a couple small utility functions. See my recent PR for a couple examples of `groupBy` and `values`.

Lastly, looking forward to seeing how Immutable.js and re-select helps us. That seems more inline with our redux state architecture.

PR #37, opened 2016-04-11
Add selected state for sources

jasonLaster:
Here's what it looks like 

gif

### DIFF
--- a/js/components/Sources.js
+++ b/js/components/Sources.js
@@ -1,27 +1,65 @@
 "use strict";
 
 const React = require("react");
+const {DOM: dom} = React;
 const { bindActionCreators } = require("redux");
 const { connect } = require("react-redux");
 const actions = require("../actions");
-const dom = React.DOM;
+const { getSelectedSource } = require("../queries");
+const { groupBy, values } = require("../utils");
 
 require("./Sources.css");
 
-function Sources({ sources, selectSource }) {
-  const sourceArr = Object.keys(sources).map(k => sources[k]);
+/**
+ * Takes a sources object indexed by actor and
+ * returns a sources object indexed by source domain.
+ *
+ * @returns Object
+ */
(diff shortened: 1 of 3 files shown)


js/components/Sources.js
```diff
@@ -1,27 +1,65 @@
...
+/**
+ * Takes a sources object indexed by actor and
+ * returns a sources object indexed by source domain.
+ *
+ * @returns Object
+ */
+function groupSourcesByDomain(sources) {
+  return groupBy(values(sources), source => {
+    if (!source.url) {
+      return source.introductionType;
```
jasonLaster: I'm not really sure what introductionType is doing, but this is what we were doing before, so i kept it

js/utils.js
```diff
@@ -73,12 +73,35 @@ function deleteIn(destObj, path) {
   return setIn(destObj, objPath, obj.without(propName));
 }
 
+function groupBy(array, predicate) {
```
jasonLaster: Related to issue #31. We can add a couple small utility functions that DRY up our code a lot.

js/components/Sources.js
jlongster: `introductionType` comes from the JS engine and tell you what kind of source it is (eval, new Function, script tag, etc). Many scripts don't have a URL if they are eval'ed. You see this on any complex site; the list of sources includes things like `Function`, `eval`, etc. Eventually we will want to treat this differently.

jlongster:
I'm about to land the immutable.js integration. You don't need your extra utility functions then. Do you want to rebase this on top of my work?

clarkbw:
If we don't merge this today, nobody gets any 🍻 
😄

jasonLaster:
merged! 
:beer: please

jlongster:
I'll be checking my mail tomorrow.

PR #38, opened 2016-04-11
Refactor breakpoints

jasonLaster:
Adds a `key` prop for each breakpoint to make React happy.

Also does a couple refactors:
1. creates a Breakpoint stateless component
2. pull getFilenameFromSources into the module context. Eventually, I'd like to have a query that fetches breakpoints with a full source, but I'd like to wait until we have re-select.

### DIFF
--- a/js/components/Breakpoints.js
+++ b/js/components/Breakpoints.js
@@ -8,6 +8,30 @@ const { getSources } = require("../queries");
 
 require("./Breakpoints.css");
 
+function getFilenameFromSources(sources, actor) {
+  const source = sources[actor];
+  const url = new URL(source.url);
+  const filename = url.pathname.substring(
+    url.pathname.lastIndexOf("/") + 1);
+  return filename;
+}
+
+function Breakpoint(sources, breakpoint) {
+  const sourceActor = breakpoint.location.actor;
+
+  const filename = getFilenameFromSources(
+    sources,
+    sourceActor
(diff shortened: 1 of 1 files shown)


js/components/Breakpoints.js
```diff
@@ -22,24 +46,12 @@ const Breakpoints = React.createClass({
...
       { className: "breakpoints" },
       dom.button({ onClick: onResumeClick }, "resume"),
       dom.ul(
         null,
-        this.props.breakpoints.map(bp => dom.li(
-          null,
-          getFilenameFromSources(this.props.sources, bp.location.actor)
-            + ", line " + bp.location.line)
-        )
+        this.props.breakpoints.map(bp => Breakpoint(this.props.sources, bp))
```
jlongster: You're not actually using a component. You're just using a function. You create a component with `React.createElement(Breakpoint)`. I would just rename `Breakpoint` as `renderBreakpoint` to make it clear that it's just a function.

jlongster:
Looks good, just rename that function and you can push it to master

js/components/Breakpoints.js
jasonLaster: sounds good

jlongster:
Pulled as https://github.com/jlongster/debugger.html/commit/75f01cb0af17b58f9d4fa82496253b6644442598

PR #41, opened 2016-04-12
remove breakpoint from list when gutter marker is removed

marfarma:
refactored add & delete breakpoint actions, added a toggle action.  Updated editor to use toggleBreakpoint action

### DIFF
--- a/js/actions/breakpoints.js
+++ b/js/actions/breakpoints.js
@@ -32,14 +32,14 @@ function enableBreakpoint(location) {
 
 function _breakpointExists(state, location) {
   const currentBp = getBreakpoint(state, location);
-  return (currentBp != undefined && !currentBp.disabled);
+  return (currentBp && !currentBp.disabled);
 }
 
 function _getOrCreateBreakpoint(state, location, condition) {
   return getBreakpoint(state, location) || { location, condition };
 }
 
-function _removeOrDisableBreakpoint(location, isDisabled, dispatch, getState) {
+function __removeOrDisableBreakpoint(location, isDisabled, dispatch, getState) {
   let bp = getBreakpoint(getState(), location);
   if (!bp) {
     throw new Error("attempt to remove breakpoint that does not exist");
@@ -117,7 +117,7 @@ function toggleBreakpoint(location, condition) {
     const exists = _breakpointExists(getState(), location);
     if (exists) {
       // if it exists delete it
-      return _removeOrDisableBreakpoint(location, false, dispatch, getState);
+      return __removeOrDisableBreakpoint(location, false, dispatch, getState);
     }
 
     // otherwise add it
@@ -132,16 +132,17 @@ function addBreakpoint(location, condition) {
 }
 
 function disableBreakpoint(location) {
-  return removeOrDisableBreakpoint(location, true);
+  return _removeOrDisableBreakpoint(location, true);
 }
 
 function removeBreakpoint(location) {
-  return removeOrDisableBreakpoint(location);
+  return _removeOrDisableBreakpoint(location);
 }
 
-function removeOrDisableBreakpoint(location, isDisabled) {
+function _removeOrDisableBreakpoint(location, isDisabled) {
   return (dispatch, getState) => {
-    return _removeOrDisableBreakpoint(location, isDisabled, dispatch, getState);
+    return __removeOrDisableBreakpoint(
+      location, isDisabled, dispatch, getState);
   };
 }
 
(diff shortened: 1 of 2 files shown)


marfarma:
@jasonLaster - here's the fix for the breakpoints not being removed from the list when they were removed from the editor gutter.

marfarma:
OK - eslint fails -- is this real, or are the eslint checks in flux?  I'd like to be sure it's necessary before I make eslint pass for these rules.

js/actions/breakpoints.js
```diff
@@ -32,91 +32,118 @@ function enableBreakpoint(location) {
 
 function _breakpointExists(state, location) {
   const currentBp = getBreakpoint(state, location);
-  return currentBp && !currentBp.disabled;
+  return (currentBp != undefined && !currentBp.disabled);
```
jlongster: Please leave this line as it was. It's ok to check for general falsy-ness here.

js/actions/breakpoints.js
```diff
@@ -32,91 +32,118 @@ function enableBreakpoint(location) {
...
-      })
-    });
+function addBreakpoint(location, condition) {
+  return (dispatch, getState) => {
+    return _addBreakpoint(location, condition, dispatch, getState);
   };
 }
 
 function disableBreakpoint(location) {
-  return _removeOrDisableBreakpoint(location, true);
```
jlongster: Please leave the underscore. It's an internal function.

jlongster:
esline rules are up-to-date. It's very hard to read the diff to see what is actually changed here. I prefer to not have `toggleBreakpoint`, and to be explicit. It's up to the caller to call add or remove explicitly. I made a few other nits too. If you could only make smaller changes to fix specific bugs, that would be great.

This project is *extremely* early on so I'm hesitant to merge a big change as it is. If I see a more specific fix, that's fine. But various things are already broken and I want to make sure we aren't introducing new bugs. I'm already very familiar with the current code so I'd prefer to not change it too much until things are more stable.

marfarma:
This was meant to be a small change to fix a specific bug.  Breakpoints were not being removed from the list when they were toggled off in the editor.  And really, it's a small bug fix, once you understand what I've done.  _(I fixed the nits you mentioned.)_

I created toggleBreakpoint because the Editor's on "gutterClick" callback is itself a toggle, so it made sense for the action to be a toggle.  Otherwise you have to check for the existence of a breakpoint in the Editor component - which requires adding the breakpoints to the props.  I had some problems making that happen - and once I thought about it, didn't think that was really the right approach.  I left the original add, disable, and remove functions in place - so they can still be called as needed.

As to what I actually did: 

```
function toggleBreakpoint(location, condition) {
  return (dispatch, getState) => {
    const exists = _breakpointExists(getState(), location);
    if (exists) {
      // if it exists delete it
      return __removeOrDisableBreakpoint(location, false, dispatch, getState);
    }

    // otherwise add it
    return _addBreakpoint(location, condition, dispatch, getState);
  };
}

function addBreakpoint(location, condition) {
  return (dispatch, getState) => {
    return _addBreakpoint(location, condition, dispatch, getState);
  };
}
```

I created an ` _addBreakpoint` function with the guts of the addBreakpoint function, so it could be called from both the `addBreakpoint` and `toggleBreakpoint`.  I did the same with  the `_removeOrDisableBreakpoint` function.  Since it already has a leading underscore, the new function's name begins with a double underscore.

js/actions/breakpoints.js
marfarma: fixed.

js/actions/breakpoints.js
marfarma: fixed

jlongster:
Thanks for making the tweaks!

The editor is going to have to pull in the breakpoints state. It really should be showing the existing breakpoints from the state, not from the internal CodeMirror gutter state. We're doing it wrong now.

This code is *highly* in flux. Since we're about to change this code a lot, I'm not sure if we should land this PR, as it goes in a different direction.

I hope this doesn't come off the wrong way, I love contributions and it's much appreciated! It's just a bit hard on this project right now which is changing a lot daily and I have a lot of changes already planned.

jasonLaster:
Thanks for the help @marfarma! I am going to close the PR for the time being, but when we get to it, I will try and incorporate your work. Thanks again!

marfarma:
No problem. 

I've become somewhat invested in the project.  What can I work on now?

jlongster:
> No problem.
> 
> I've become somewhat invested in the project. What can I work on now?

There's a lot of moving parts and we are still figuring out the best places to work on. Keep an eye out and we'll have something soon. I think it's best to give it some time to stabilize so you don't waste work on something and we realize that it's not applicable anymore.

PR #43, opened 2016-04-12
migrate to immutable.js

jlongster:
I haven't tested this extensively, but the basic features work. I'd like to merge the migration incrementally, even if it's likely that there are bugs.


js/components/Editor.js
```diff
@@ -59,7 +59,9 @@ const Editor = React.createClass({
   },
 
   componentDidUpdate() {
-    this.editor.setValue(this.props.sourceText);
```
jlongster: This and `componentWillReceiveProp` both need a lot of work. I'd say we mostly ignore this code for now until we focus on the editor. For some reason this code was giving me problems and it worked without it.

jasonLaster: agreed. figured that was a placeholder at the time

js/components/Breakpoints.js
```diff
@@ -9,7 +9,7 @@ require("./Breakpoints.css");
 
 const Breakpoints = React.createClass({
   propTypes: {
-    breakpoints: PropTypes.array,
```
jasonLaster: it'd be nice to have better annotations at somepoint

js/components/Tabs.js
```diff
@@ -29,13 +27,17 @@ function Tabs({ tabs, selectTab, loadSources, newSource }) {
...
+    tabs.valueSeq().map(tab => {
+      return dom.li(
+        { "className": "tab",
+          "data-actor-id": tab.get("actor"),
+          "onClick": onClickTab },
+        dom.div(
+          { className: "tab-title" },
+          tab.get("title"),
+          dom.div({ className: "tab-url" }, tab.get("url"))
+        )
```
jasonLaster: nice. looks much nicer.

jlongster: I just realized I think I messed up the structure here. `tab-url` shouldn't be inside of `tab-title`. Doh! You're structure before was pretty good but the main thing was moving the props down. Otherwise my JS formatter wants to bump the rest of the args over to line up with it...

Anyway, I'll make a new commit that fixes the structure

js/components/Breakpoints.js
jlongster: Just found this: https://www.npmjs.com/package/react-immutable-proptypes

jlongster:
I rebased this so it can be merged. I added the react-immutable-proptypes package and it's really nice because it makes it clear what is immutable.js and what isn't. Feels good.

jlongster:
I will squash these commits if you r+

jasonLaster:
r+. squash and :ship:

jlongster:
Squashed and :ship:'ed!

PR #54, opened 2016-04-12
Experiment with client interface and get tests working

jlongster:


jlongster:
Not ready to be merged, but putting a prototype up for discussion.

jlongster:
Yay, I wrote a test! https://github.com/jlongster/debugger.html/blob/client/js/actions/tests/sources.js

I gave up on a perfect mapping of the gecko-dev xpcshell environment. Normally you don't require the head file, but it's automatically included for you. When we do the merge, we'll have to go through and convert that require statement to something else. But that's not a big deal; we can probably even use a codegen tool to automatically go through and do it.

Everything except the `require("../../util/test-head");` should "just work" when we move to the real xpcshell harness.

This is include in the client connection work because the whole point of moving that to be local to the actions is so that it's easy to stub it out. Now we just need to create this mock connection object: https://github.com/jlongster/debugger.html/blob/client/js/util/test-head.js All of the actions will reference that mock object.

jlongster:
I'd really like it if tests were a focus for the next day. I'll talk to you tomorrow about this, Jason.

js/client.js
```diff
@@ -4,33 +4,41 @@ const { DebuggerClient } = require("devtools/shared/client/main");
...
 
 function connectToTab(tab, onNewSource) {
   let deferred = promise.defer();
-  const options = { client, form: tab, chrome: false };
+  const options = { currentClient, form: tab, chrome: false };
 
   TargetFactory.forRemoteTab(options).then(target => {
     target.activeTab.attachThread({}, (res, threadClient) => {
       threadClient.resume();
       window.gThreadClient = threadClient;
```
jasonLaster: can we delete this now?

js/components/Accordion.js
```diff
@@ -39,7 +39,8 @@ const Accordion = React.createClass({
       { className: "accordion" },
       this.props.items.map((item, i) => {
         return div(
-          { className: opened[i] ? "opened" : "" },
+          { className: opened[i] ? "opened" : "",
+            key: i },
```
jasonLaster: :+1:

jasonLaster:
merged!

PR #56, opened 2016-04-13
Cleanup queries

jasonLaster:
The refactor cleans up many of the queries in two important ways:
1) composition
2) return types

Composition - there are a few queries (selectors) that access state properties directly. 

Return types - queries now return the state's immutable js types. This extends the immutable js boundary significantly. It is now reasonable to expect a component propType to be an immutable type and to expect immutable types to be passed into actions. The new immutable js boundary is the client and reducers, which means that both the client and reducers expect plain JS objects. It is probably reasonable to have reducers expect immutable.js objects soon though.

While on the subject of types, one of the pain points I felt during the refactor was some minor inconsistencies where i'd expect a property to be a Map, but found it was an Object. I hope that we can iron some of these inconsistencies out soon. Also, I'd like to explore some domain specific propTypes with the immutable-proptypes library e.g.

```js
const BreakpointType = ImPropTypes.mapOf({
  location: ImPropTypes.mapOf({ 
     line: propType.number,
     column: propType.number
  }),
  loading: propType.boolean
});
```


jlongster:
This looks really good, r+ :+1: You can merge this after rebasing!

jasonLaster:
manually merged!

PR #60, opened 2016-04-15
Add node + mocha

jasonLaster:
Adds node + mocha

One of the interesting side-effects is that node needed a babel transpile step. One one hand, this is extra work, on the other hand, it helps us think about landing the debugger client in different environments, which may or may not support some features which is nice.

I'll add karma and a web runner sooner.


js/actions/tests/.eslintrc
```diff
@@ -1,5 +1,26 @@
...
+    "suiteSetup": true,
+    "suiteTeardown": true,
+    "teardown": true,
+    "test": true,
+    "xcontext": true,
+    "xdescribe": true,
+    "xit": true,
+    "xspecify": true,
+    "assert": true,
+    "expect": true,
```
jasonLaster: NOTE: this is just inside of the tests dir

js/queries.js
```diff
@@ -39,11 +39,11 @@ function getSourceCount(state) {
 }
 
 function getSourceByURL(state, url) {
-  return getSources(state).find(source => source.url == url);
+  return getSources(state).find(source => source.get("url") == url);
 }
 
 function getSourceByActor(state, actor) {
-  return getSources(state).find(source => source.actor == actor);
+  return getSources(state).find(source => source.get("actor") == actor);
```
jasonLaster: found our first bug!

package.json
```diff
@@ -25,13 +28,18 @@
     "ws": "^1.0.1"
   },
   "devDependencies": {
+    "babel": "^6.5.2",
+    "babel-core": "^6.7.6",
```
jasonLaster: I'm not convinced we need these...

js/queries.js
jlongster: Heh, I hit something similar when I ran the first test under xpcshell. I thought I fixed this? Ah well, maybe not!

jlongster:
Very nice!

I thought mocha had a way to "build" source so we didn't have to go through the webpack shenanigans? The unfortunate thing is that errors will still show up in the built files. Although, I suppose we can `require("source-map-support")` now and it should sourcemap node's stack traces.

It'd be nice to avoid the webpack step, does mocha not provide anyway to compile before running?

jasonLaster:
I'm going to change `./bin/node-test` to `./bin/test`

Also, I'm thinking about changing the webpack entries field to build one output. 

There are two reasons for this 
(1) performance, adding babel added a lot of weight and i don't want performance to degrade linearly. 
(2) web runner, the web runner will be simpler if it expects one js bundle

You'll have the ability to pass in a path to a test so you can always speed up your test runs that way.

jlongster:
You can pull out the common chunks using the a plugin (`CommonChunksPlugin` I think?), but do whatever you feel is right. Surely this is a comment path for any project running es6 mocha tests. What do other people do? It'd be great to avoid this whole webpack step...

jasonLaster:
@jlongster yea, it would be really nice to avoid webpack. I'm going to push this version when it is working and add de-webpacking it as a next step so that other work is unblocked.

jasonLaster:
merged manually

PR #61, opened 2016-04-15
improve breakpoints

jlongster:
**Not ready to be merged**

This makes it so you can remove breakpoints and makes our immutable usage more consistent. We still need to figure some things out: right now we are treating anything passed through actions as normal JS objects. This is because you frequently do `this.props.fireAction({ x: 1})` so you can easily pass an object. However now we need to remember to do `this.props.fireAction(obj.toJS())` if we are using something from the existing state, like a breakpoint location. I'm sure we can figure this out; we could make actions detect if a parameter is immutable and coerce it to a JS object or something like that. Not exactly sure yet.

We also need to make the editor render breakpoints better. This still adds/removes a breakpoint on gutter click, but we really need to always rerender breakpoints whenever the state changes. For example, the server may "move" a breakpoint, and our UI breaks when the happens because the editor doesn't know to move the original bp that we added. If we always rendered from the state the bp would just magically move because the state changed.

But there are perf implications for doing that, so we need to do some kind of manual diffing of the breakpoint state and what's currently in the editor. Still more to research, and I'd like to do that before this is merged, but wanted to get this up.

### DIFF
--- a/js/components/Editor.js
+++ b/js/components/Editor.js
@@ -1,14 +1,15 @@
 "use strict";
 
 const React = require("react");
-const { DOM: dom, PropTypes } = React;
-const ReactDOM = require("react-dom");
-
-const { getSourceText, getPause, getBreakpointsForSource } = require("../queries");
+const Isvg = React.createFactory(require("react-inlinesvg"));
 const { bindActionCreators } = require("redux");
 const { connect } = require("react-redux");
+const ReactDOM = require("react-dom");
+const ImPropTypes = require("react-immutable-proptypes");
+const { DOM: dom, PropTypes } = React;
+
+const { getSourceText, getBreakpointsForSource, getPause } = require("../queries");
 const actions = require("../actions");
-const Isvg = React.createFactory(require("react-inlinesvg"));
 const { alignLine } = require("../util/editor");
 
 require("codemirror/lib/codemirror.css");
@@ -29,11 +30,39 @@ function makeMarker() {
   return marker;
 }
 
+const _Breakpoint = React.createClass({
+  propTypes: {
+    breakpoint: ImPropTypes.map,
+    editor: PropTypes.object
+  },
+
+  componentWillMount: function() {
+    const bp = this.props.breakpoint;
+    const line = bp.getIn(["location", "line"]) - 1;
+
+    this.props.editor.setGutterMarker(line, "breakpoints", makeMarker());
+  },
+
+  componentWillUnmount: function() {
+    const bp = this.props.breakpoint;
+    const line = bp.getIn(["location", "line"]) - 1;
+
+    this.props.editor.setGutterMarker(line, "breakpoints", null);
+  },
+
+  render: function() {
+    return null;
+  }
+});
+const Breakpoint = React.createFactory(_Breakpoint);
+
 const Editor = React.createClass({
   propTypes: {
+    breakpoints: ImPropTypes.list,
     selectedSource: PropTypes.object,
     sourceText: PropTypes.object,
     addBreakpoint: PropTypes.func,
+    removeBreakpoint: PropTypes.func,
     pause: PropTypes.object
   },
 
@@ -41,7 +70,7 @@ const Editor = React.createClass({
     this.editor = CodeMirror.fromTextArea(this.refs.editor, {
       mode: "javascript",
       lineNumbers: true,
-      lineWrapping: true,
+      lineWrapping: false,
       smartIndent: false,
       matchBrackets: true,
       styleActiveLine: true,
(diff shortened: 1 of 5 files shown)


js/components/Editor.js
```diff
@@ -48,37 +48,53 @@ const Editor = React.createClass({
...
+      // here, but that might be too slow.
     }
+  },
 
-    this.editor.setValue(sourceText.get("text"));
-    this.editor.setCursor(cursor);
+  onGutterClick(cm, line, gutter, ev) {
+    const currentBp = this.props.breakpoints.find(bp => {
+      return bp.getIn(["location", "line"]) === line + 1;
+    });
```
jasonLaster: This should probably be a query `getBreakpointBySourceLine(source, line+1)`

js/queries.js
```diff
@@ -17,8 +19,13 @@ function getSelectedSourceOpts(state) {
...
 
-function getBreakpoints(state) {
-  return state.breakpoints.get("breakpoints");
+function getBreakpoints(state, actor) {
+  if(!actor) {
+    return state.breakpoints.get("breakpoints");
+  }
+  return state.breakpoints.get("breakpoints").filter(bp => {
+    return bp.getIn(["location", "actor"]) === actor;
+  });
```
jasonLaster: Could this also be two functions: `getBreakpoints` and `getBreakpointsByActor`?

In the case of the editor, I think you'll always want to call `getBreakpointsByActor` even if there isn't a selected actor. If there isn't a selected source, then the editor doesn't want any breakpoints.

js/queries.js
```diff
@@ -51,14 +58,18 @@ function getSourceText(state, actor) {
...
 
 /**
  * @param object - location
  */
 function makeLocationId(location) {
-  return location.actor + ":" + location.line.toString();
+  if(!Immutable.Iterable.isIterable(location)) {
+    location = Immutable.fromJS(location);
+  }
+  return location.get("actor") + ":" + location.get("line").toString();
```
jasonLaster: I see why you need to check the location type here and `getBreakpoint`, but it's disappointing.

js/components/Editor.js
jlongster: It could be. I don't think every single data access needs to be a query though. If we expect to get an array of things, I think it's fine to operate on that array instead of having to go back through the queries. The queries mainly abstract out where that array comes from, so you can change around your app state.

You could even say that accessing the line from the location field of a breakpoint should be a query. But we can accept some amount of dependency on the state shape to make things more straight-forward.

So I'm not against it, but I'd rather wait and see if we need that in more than 1 place.

js/queries.js
jlongster: Yeah, that's a good call.

js/components/Editor.js
jasonLaster: Makes sense.

js/queries.js
jlongster: There is still some mismatch in a few places. This is what I was referencing in the first paragraph of this PR description. We should not need to coerce here; it was a stop-gap. There seemed to be one or two places where a non-immutable value would come through here but it wasn't straight-forward how.

I'll look at this more now and see if I can't do a more thorough conversion to immutable.

We need to decide which values are immutable or not. It's clear that if our state is immutable, components will receive immutable values and use them directly. But I think any parameters to actions or queries should be normal JS objects because you usually pass literals in (`action({ line: 5})`).

jasonLaster: I could see queries and actions being immutable as well. 
+ The argument for reducers is that way they receive and set immutable objects. 
+ Queries is less clear, but at this point i believe they mostly receive a string or immutable object

With that said, I'd be fine with actions, queries, and client calls receiving plain JS objects. I haven't thought this through and you have more experience here.

jlongster:
Still working on this. This wasn't really part of the current sprint, but Jason is working on the paused stuff like stepping and it would be good to get breakpoints in a better state. I think we should be able to finish everything in this sprint. I will fix the breakpoint styles in here as well which is part of this sprint.

The last thing I want to do is make the editor render the current breakpoint list whenever it changes. Don't add breakpoints to the editor directly, but add them to the state and the editor automatically updates. This will make things a lot more consistent. There are some perf considerations that I'm thinking through though.

jasonLaster:
@jlongster looking good so far. Glad that you're addressing the editor state now :+1:

jlongster:
I'm going to go ahead and merge this because we need to get this in and the changes are not big. Breakpoints should now be deterministically rendered from the app state.

js/components/Editor.js
```diff
@@ -129,6 +148,10 @@ const Editor = React.createClass({
         dom.textarea({
           ref: "editor",
           defaultValue: "..."
+        }),
+        this.props.breakpoints && this.props.breakpoints.map(bp => {
+          return Breakpoint({ breakpoint: bp,
+                              editor: this.editor });
```
jasonLaster: moving rendering breakpoints down here is brilliant

PR #65, opened 2016-04-15
Karma

jasonLaster:
Adds karma, which can be run with this command `karma start`

Also adds karma to travis, which will run tests on firefox and chrome. I
tried to add phantomjs and found that we would need to babel in some
features, so i passed for the time being.

NOTE: currently uses the node bundle because for some reason chrome and canary are complaining when the bundle is not babelfied.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,14 @@ node_js:
   - "5"
 install:
   - npm install
+before_install:
+  - export CHROME_BIN=chromium-browser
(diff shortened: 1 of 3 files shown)


jasonLaster:
With the browser test runner merged, karma is much smaller.

jasonLaster:
Merged. I don't think there's much contention in how karma is configured. Feel free to comment though and I can take a look

Issue #63, opened 2016-04-15
Fix editor scrolling in Chrome


jlongster:
For some reason it doesn't scroll right now. Works fine in Firefox. It's probably a simple CSS fix (`overflow: auto` somewhere).

jasonLaster:
this might address some of the firefox, chrome inconsistencies
http://stackoverflow.com/questions/33636796/webkit-not-filling-the-100-height-of-a-flex-parent

jasonLaster:
Actually, not fixed yet

jasonLaster:
This comment offered a solution/hack around a codemirror parent / grandparent https://github.com/codemirror/CodeMirror/issues/3679#issuecomment-171533964

jasonLaster:
Fixed! Yessssss

PR #70, opened 2016-04-18
Fix chrome layout styles

jasonLaster:
1. replaces %s with flex 1
2. rename code-editor class to editor class
3. remove unecessary app div


jasonLaster:
Addresses #63

PR #72, opened 2016-04-18
Convert Action argument destructuring to list

jasonLaster:
This fixes chrome devtools pause crash, which would happen when there
was a destructured param.

old 
```js
return ({ dispatch, getState, threadClient }) => {}
```

new
```js
return (dispatch, getState, threadClient) => {}
```


jlongster:
T_T 

This works for now. I think we will want to inject more things in the future (the regular `client` instance) but we'll figure that out then.

PR #76, opened 2016-04-18
Fix chrome param destructuring bug

jasonLaster:
I found a couple more examples of parameter destructuring ([1](https://github.com/jlongster/debugger.html/blob/master/js/components/Sources.js#L38-L41), [2](https://github.com/jlongster/debugger.html/blob/master/js/components/Sources.js#L11-L12)) while testing.

If it's not an action specific problem, I think it makes sense to let Babel solve the problem for those who want to opt-in. Babel bumps the build from ~2 to ~3 seconds, but more importantly, opens the door to features not yet supported, so I feel like it should be off by default.

One advantage to this solution, is that we can revert the action destructuring that we're currently doing, which should be made unnecessary.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -3,7 +3,14 @@
 const path = require("path");
 const ExtractTextPlugin = require("extract-text-webpack-plugin");
 
-module.exports = {
+let environment;
+try {
+  environment = require("./environment");
+} catch (e) {
+  environment = {};
+}
+
+let config = {
   entry: "./js/main.js",
   devtool: "source-map",
   output: {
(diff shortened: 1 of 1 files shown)


jasonLaster:
After thinking about this a bit more, I'm planning on changing the field from `shouldTranspile` to `transformParameters` and removing it from the README. I don't want to give the impression that it is common to transpile.

jlongster:
Good idea, I like this way better than forcing code patterns on it. The destructuring in React components is certainly a big deal and we shouldn't lose out on that just because of a Chrome bug.

jasonLaster:
cool. babel ftw

PR #78, opened 2016-04-20
Add Debugger Commands

jasonLaster:
Features:

* Commands
  * Resume / Pause
  * Step Over, Step In, Step Up
  * Some pixel (im)perfect icons
* on pause
  * updates source
  * scrolls to and highlights debug line
  * shows breakpoints
* Right Sidebar
  * has a top action bar
  * has a call stack and scopes pane
  * breakpoints shows "no breakpoints"

Implementation

* Redux
  * new pause action and reducer
  * pause / resume actions for responding to server events
  * command / breakOnNext action for debugger actions
* Editor
  * better source text updating
  * breakpoint gutter markers are added on the props update lifecycle
  * stateless editor utility for bits of functionality like (alignLine)
* debugTab
  * extracted the logic in main/Tabs for debugging a tab
  * sets up the client listeners for paused/resumed
  * added unsolicited event logging as a sanity measure

[demo](https://www.dropbox.com/s/17sh3xi5iwjyqpn/debugger-screen.mov?dl=0)

### DIFF
--- a/js/client.js
+++ b/js/client.js
@@ -37,8 +37,52 @@ function getThreadClient() {
   return currentThreadClient;
 }
 
+const unsolicitedEvents = [
+  "paused", "resumed", "detached", "consoleAPICall", "eventNotification",
+  "fileActivity", "lastPrivateContextExited", "logMessage", "networkEvent",
+  "networkEventUpdate", "newGlobal", "newScript", "tabDetached",
+  "tabListChanged", "reflowActivity", "addonListChanged", "workerListChanged",
+  "serviceWorkerRegistrationListChanged", "tabNavigated", "frameUpdate",
+  "pageError", "documentLoad", "enteredFrame", "exitedFrame", "appOpen",
+  "appClose", "appInstall", "appUninstall", "evaluationResult", "newSource",
+  "updatedSource", "resumeLimit", "debuggerStatement", "breakpoint", "DOMEvent",
+  "watchpoint", "exception"
+];
+
+function debugTab({ tabActor, newSource, paused, resumed,
+                    selectTab, selectSource, loadSources }) {
+  function listenToClient() {
+    let deferred = promise.defer();
+    let client = getThreadClient();
+
+    client.addListener("paused", (_, packet) => {
+      paused(packet);
+      if (packet.why.type != "interrupted") {
+        selectSource(packet.frame.where.source);
+      }
+    });
+
+    client.addListener("resumed", (_, packet) => resumed(packet));
+
+    client.addListener("newSource", (_, packet) => newSource(packet.source));
+
+    unsolicitedEvents.forEach(event => {
+      client.addListener(event, (_, packet) => {
+        console.log(event, packet);
+      });
+    });
+
+    return deferred.resolve();
+  }
+
+  return selectTab({ tabActor: tabActor })
+    .then(loadSources)
+    .then(listenToClient);
+}
+
 module.exports = {
   connectToClient,
   connectToTab,
-  getThreadClient
+  getThreadClient,
+  debugTab
 };
--- a/js/components/Breakpoints.js
+++ b/js/components/Breakpoints.js
@@ -1,4 +1,3 @@
-/* globals URL gThreadClient */
 "use strict";
 
 const React = require("react");
@@ -36,6 +35,15 @@ function renderBreakpoint(sources, breakpoint) {
   );
 }
 
+function renderBreakpointList(breakpoints, sources) {
+  return dom.ul(
+    null,
+    breakpoints.valueSeq().map(bp => {
+      return renderBreakpoint(sources, bp);
+    })
+  );
+}
+
 const Breakpoints = React.createClass({
   propTypes: {
     breakpoints: ImPropTypes.map.isRequired,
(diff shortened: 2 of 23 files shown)


js/client.js
```diff
@@ -37,8 +37,52 @@ function getThreadClient() {
...
+  "networkEventUpdate", "newGlobal", "newScript", "tabDetached",
+  "tabListChanged", "reflowActivity", "addonListChanged", "workerListChanged",
+  "serviceWorkerRegistrationListChanged", "tabNavigated", "frameUpdate",
+  "pageError", "documentLoad", "enteredFrame", "exitedFrame", "appOpen",
+  "appClose", "appInstall", "appUninstall", "evaluationResult", "newSource",
+  "updatedSource", "resumeLimit", "debuggerStatement", "breakpoint", "DOMEvent",
+  "watchpoint", "exception"
+];
+
+function debugTab({ tabActor, newSource, paused, resumed,
```
jasonLaster: `debugTab` used to be duplicated in `Tabs` and `Main`.

js/components/Breakpoints.js
```diff
@@ -44,36 +52,12 @@ const Breakpoints = React.createClass({
...
-        dom.button({ onClick: this.onResumeClick }, "Resume"),
-        dom.button({ onClick: this.onStepOverClick }, "Over"),
-        dom.button({ onClick: this.onStepInClick }, "In"),
-        dom.button({ onClick: this.onStepOutClick }, "Out")
-      ),
-      dom.ul(
-        null,
-        this.props.breakpoints.valueSeq().map(bp => {
-          return renderBreakpoint(this.props.sources, bp);
-        })
```
jasonLaster: I'm really happy that in many places this PR cleaned up existing components. i.e. extracting actions from Breakpoints.

js/components/Editor.js
```diff
@@ -50,36 +53,73 @@ const Editor = React.createClass({
   },
 
   onGutterClick(cm, line, gutter, ev) {
-    let info = cm.lineInfo(line);
-    cm.setGutterMarker(
-      line,
-      "breakpoints",
-      info.gutterMarkers ? null : makeMarker()
-    );
-
```
jasonLaster: breakpoint gutters are now added as part of the `componentWillReceiveProps` lifecycle so that breakpoints are persisted as selected source changes. It is also more uni-directional 😉

js/queries.js
```diff
@@ -54,6 +58,15 @@ function getBreakpoint(state, location) {
...
 }
 
+function getBreakpointByActor(state) {
+  return getBreakpoints(state).valueSeq()
+    .groupBy(bp => bp.getIn(["location", "actor"]));
+}
+
+function getBreakpointsForSource(state, actor) {
+  return getBreakpointByActor(state).get(actor);
+}
```
jasonLaster: this is definitely going to conflict with your breakpoint work, but hopefully it's very similar to what you have

jlongster:
Hah, this definitely conflicts. It would be nice if you removed your breakpoint changes so I can land mine easier. I'm already working on that part of the code. Any reason you needed it here? It seems like you can land all the other stuff without touching breakpoints. It would be great if you could do that.

js/client.js
```diff
@@ -37,8 +37,52 @@ function getThreadClient() {
   return currentThreadClient;
 }
 
+const unsolicitedEvents = [
```
jlongster: Why are you copying unsolicitedEvents here? It looks like you are only using it to log the protocol, but this isn't the place to do it. If you set `require("devtools/shared/DevToolsUtils").dumpn.wantLogging = true` it should enable verbose logging which should log the protocol. If not, we just need to fix that in `ff-devtools-libs` because that's how it works in gecko-dev. (that setting is also flipped with a pref)

js/components/Editor.js
jlongster: As said in my other comment, I'd prefer if you reverted all the changes in this file. I'm actively working on it already.

js/reducers/pause.js
```diff
@@ -0,0 +1,27 @@
...
+});
+
+function update(state = initialState, action, emit) {
+  switch (action.type) {
+    case constants.PAUSED:
+      return state.set("pause", fromJS(action.value));
+    case constants.RESUME:
+      return state.set("pause", null);
+    case constants.BREAK_ON_NEXT:
+      return state.set(["breakOnNext"], true);
```
jlongster: I don't think "breakOnNext" should be an array.

js/util/editor.js
```diff
@@ -0,0 +1,49 @@
...
+// Maximum allowed margin (in number of lines) from top or bottom of the editor
+// while shifting to a line which was initially out of view.
+const MAX_VERTICAL_OFFSET = 3;
+
+/**
+ * Aligns the provided line to either "top", "center" or "bottom" of the
+ * editor view with a maximum margin of MAX_VERTICAL_OFFSET lines from top or
+ * bottom.
+ */
+function alignLine(cm, line, align = "top") {
```
jlongster: Is this stolen from our existing debugger? Just curious.

jlongster:
Meh, you know what, let's just merge this (after my other comments are addressed) and I'll manually copy over my changes. I'll ping you on IRC about it.

js/reducers/pause.js
jasonLaster: oh oops

js/util/editor.js
jasonLaster: yeah. but modified not to use `this`

js/client.js
jasonLaster: cool

jlongster: Can you remove the unsolicitedEvents here? I don't think we should add them here. Otherwise this PR is fine to merge

PR #80, opened 2016-04-21
Start simple node express server

jasonLaster:
+ uses webpack-dev-middleware
+ replaces starting node serve-static
+ eventually we can add other niceties here (cache busting, hmr, ...)

Benefits of webpack-dev-middleware:
+ doesn't write to file
+ on refresh, will hold the request if there isn't a build ready

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "debugger.html",
   "scripts": {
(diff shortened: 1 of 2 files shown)


jlongster:
Sounds good to me, we've needed to do this.

PR #81, opened 2016-04-21
Tests (part 1)

jasonLaster:
Cleans up the tests 
Also adds a loadSourceText  test (based on our pairing last wednesday)

### DIFF
--- a/js/actions/tests/.eslintrc
+++ b/js/actions/tests/.eslintrc
@@ -27,5 +27,6 @@
   "rules": {
     "camelcase": 0,
     "no-unused-vars": [2, {"vars": "all", "args": "none", "varsIgnorePattern": "run_test"}],
+    "max-nested-callbacks": [2, 4],
   }
 }
--- a/js/actions/tests/loadSourceText.js
+++ b/js/actions/tests/loadSourceText.js
@@ -1,0 +1,44 @@
+"use strict";
+
+const { createStore } = require("../../util/test-head");
+const { loadSourceText } = require("../../actions");
+const { getSourceText } = require("../../queries");
+
+var expect = require("chai").expect;
+
+const sourceText = {
+  "foo1": { source: "function() {\n  return 5;\n}",
+            contentType: "text/javascript" },
+  "foo2": { source: "function(x, y) {\n  return x + y;\n}",
+            contentType: "text/javascript" }
+};
+
+const mockThreadClient = {
+  source: function(form) {
+    return {
+      source: () => {
+        return new Promise((resolve, reject) => {
+          resolve(sourceText[form.actor]);
+        });
+      }
+    };
+  }
+};
+
+describe("loadSourceText", () => {
+  beforeEach(function() {
+    this.store = createStore(mockThreadClient);
+  });
+
+  describe("loading one source text", function() {
+    beforeEach(function(done) {
+      this.store.dispatch(loadSourceText({ actor: "foo1" }))
+        .then(() => done());
+    });
+
+    it("lkj", function() {
+      const fooSourceText = getSourceText(this.store.getState(), "foo1");
+      expect(fooSourceText.get("text")).to.equal(sourceText.foo1.source);
+    });
+  });
+});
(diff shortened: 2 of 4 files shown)


js/actions/tests/.eslintrc
```diff
@@ -27,5 +27,6 @@
   "rules": {
     "camelcase": 0,
     "no-unused-vars": [2, {"vars": "all", "args": "none", "varsIgnorePattern": "run_test"}],
+    "max-nested-callbacks": [2, 4],
```
jasonLaster: 4 gives us two describes (one for the action, the other a scenario)

js/actions/tests/loadSourceText.js
```diff
@@ -0,0 +1,44 @@
...
+      source: () => {
+        return new Promise((resolve, reject) => {
+          resolve(sourceText[form.actor]);
+        });
+      }
+    };
+  }
+};
+
+describe("loadSourceText", () => {
```
jasonLaster: I split two source actions loadSourceText and newSource to their own test files... I'm not sure if that's necessary. Maybe here, but not with other actions... I guess we'll figure that out.

jasonLaster:
looks like chromium 37 is failing. Given that it's 2 years old, i'm in favor of dropping it from karma for now and filing a bug for runnning a more up to date chrome, possibly with circle ci or another CI tool.

js/actions/tests/newSource.js
```diff
@@ -1,25 +1,28 @@
 "use strict";
 
-const { createStore, actions, queries } = require("../../util/test-head");
+const { createStore } = require("../../util/test-head");
+const { newSource } = require("../../actions");
+const { getSourceByActor } = require("../../queries");
+
 const store = createStore();
 var expect = require("chai").expect;
```
jlongster: Do you feel strongly about chai? I prefer https://www.npmjs.com/package/expect, but I don't want to slow this down.

js/actions/tests/loadSourceText.js
```diff
@@ -0,0 +1,44 @@
+"use strict";
+
+const { createStore } = require("../../util/test-head");
+const { loadSourceText } = require("../../actions");
```
jlongster: I'd rather it be like it was before so you don't always have to add a new variable here whenever you need another action. You also don't have to worry about importing any actions/queries/etc. It's not that big of a deal to do `actions.newSource` instead of just `newSource`

js/actions/tests/loadSourceText.js
jlongster: Whatever feels right :)

js/util/test-head.js
```diff
@@ -17,12 +13,13 @@ const _createStore = configureStore({
   }
 });
 
-function createStore() {
+function createStore(threadClient) {
+  mockThreadClient = threadClient;
```
jlongster: Why not move `configureStore` into here and dynamically create a new store constructor each time and just close over the mock thread client?

jlongster:
Can you focus on tests tomorrow? I'd rather see a lot more tests being written than new features at this point.

js/actions/tests/newSource.js
jasonLaster: nope - happ to drop it

js/actions/tests/loadSourceText.js
jasonLaster: yea, that works.

js/util/test-head.js
jasonLaster: I think I follow. Will give that a try tomorrow

jasonLaster:
haha - yea, was thinking that too. Just saw your comment.

js/actions/tests/newSource.js
jlongster: Honestly I can't really remember the real reason I switched a while ago. Maybe we should see what tofino is using; if we don't care too much it'd be nice to be on the same page for some stuff like this

jasonLaster: Yea, that's a good idea. At this point, I'm numb to assertion libraries. I've given up hope that my muscle memory will know if it's camelCased or dot deliniated. Lets look around for a convention soon because it will be a pain to switch soon.

jasonLaster:
merged!

PR #82, opened 2016-04-21
Move Karma tests to circle-ci

jasonLaster:
Moves the Karma tests out of Travis and into Circle Ci. There are two benefits with the move:
1) the tests are run with an up to date version of chrome
2) the slower karma tests are run in along side the faster tests, which hopefully will keep our tests faster.

Signing up for circle ci is as easy as going to (https://circleci.com/) and enabling debugger.html

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,13 +4,8 @@ node_js:
 install:
   - npm install
 before_install:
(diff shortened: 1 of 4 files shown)


jlongster:
I think I enabled it in CircleCI

PR #83, opened 2016-04-21
Fix Chrome's Codemirror Editor layout

jasonLaster:
Addresses https://github.com/jlongster/debugger.html/issues/63

The problem is that codemirror's parent is a flexbox element and
codemirror can't reliably calcuate it's height and width.

The *fix* is:
+ grandparent: flexbox, position relative
+ parent: position absolute, width and height 100%

### DIFF
--- a/js/components/Editor.js
+++ b/js/components/Editor.js
@@ -124,12 +124,13 @@ const Editor = React.createClass({
 
   render() {
     return (
-      dom.div(
-        { className: "editor" },
-        dom.textarea({
-          ref: "editor",
-          defaultValue: "..."
-        })
+      dom.div({ className: "editor-container" },
+          dom.div({ className: "editor" },
+          dom.textarea({
+            ref: "editor",
+            defaultValue: "..."
+          })
+        )
       )
     );
   }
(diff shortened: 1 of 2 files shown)


js/components/Editor.js
```diff
@@ -124,12 +124,13 @@ const Editor = React.createClass({
 
   render() {
     return (
-      dom.div(
-        { className: "editor" },
-        dom.textarea({
-          ref: "editor",
-          defaultValue: "..."
-        })
+      dom.div({ className: "editor-container" },
```
jlongster: Are you sure you need a container? A previous commit made `.editor` have a width/height of 100% but it was removed for some reason. I think it was like this before:

```
.editor { flex: 1}

.CodeMirror { width: 100%; height: 100% }
```

You're supposed to give `.CodeMirror` a width/height if you want to tell CodeMirror what it should be. I think that should work and avoids a container element, but I could be wrong.

jasonLaster: sadly that doesn't work due to a chrome flex box bug with height and width's with percentages.

jlongster: damnit Chrome, so many bugs

jasonLaster: lol. soo many

PR #84, opened 2016-04-21
Source tabs

jasonLaster:
Took a small break from adding tests to add source tabs before I pair with Helen. 

screen shot 2016-04-21 at 3 36 37 pm

Implementation details
+ Added a new SourceTabs component
+ Removed the splitter left border as it was causing a gap in the UI
+ Added styling for tabs
+ Cleaned up App.js a bit

This PR includes the commit for fixing chrome's editor layout. Once that's merged I can remove that commit from this PR.


js/components/Editor.css
```diff
@@ -4,8 +4,16 @@
...
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+/**
+ * There's a known codemirror flex issue with chrome that this addresses.
+ * BUG https://github.com/jlongster/debugger.html/issues/63
+ */
 .editor {
-  flex: 1;
+  position: absolute;
+  height: calc(100% - 30px);
```
jasonLaster: the 30px could be moved into a variable so that it is shared between source tabs and the editor.

still, not in love with this `position: absolute` solution, but it's the only hack I could find for getting around this known chrome bug.

js/components/Editor.js
```diff
@@ -4,7 +4,10 @@ const React = require("react");
 const { DOM: dom, PropTypes } = React;
 const ReactDOM = require("react-dom");
 
-const { getSourceText, getPause, getBreakpointsForSource } = require("../queries");
+const {
+  getSourceText, getPause, getBreakpointsForSource,
+  getSelectedSource } = require("../queries");
```
jlongster: This is a strange style. Can you move the last curly brace to a new line?

```js
const {
  getSourceText, getPause, getBreakpointsForSource,
  getSelectedSource
} = require("../queries");
```

js/components/Editor.css
jlongster: Yeah, that makes me really sad. Flexbox was supposed to allow us to avoid this kind of stuff. But I guess it works for now.

PR #85, opened 2016-04-21
Add breakpoint styling

jasonLaster:
This sets the line number font to white and moves the breakpoint below
the linenumber.

I couldn't find where Firefox did that work, so I thought I'd share the
PR incase there's a better way of doing it. One thing I noticed for
example, is if the breakpoint appears ahead in the DOM structure, then
it doesn't need any z-index treatment.

### DIFF
--- a/js/components/Editor.css
+++ b/js/components/Editor.css
@@ -25,6 +25,16 @@
   font-family: Menlo, monospace !important;
 }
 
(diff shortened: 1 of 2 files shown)


jlongster:
Nice! `last-child` will break if we ever change the structure, but I don't think we'll do that, and it's easy to change this if we do.

PR #87, opened 2016-04-22
(WIP) Test watcher

jasonLaster:
I took a stab at adding a test watcher for node.

Most of the diff is linting test-node. Notable changes are:
+ commander for parsing arguments
+ fs.watch for watching for changes

Here is the [error](https://gist.github.com/jasonLaster/e1eded70e37820c5a9ca50348b8e6c53) I'm seeing. I didn't spend too long debugging it, but it looks like an issue with ff-devtools-libs, where these two shams are not being requested the second time:

```
Cc sham for @mozilla.org/io-util;1
CC sham for @mozilla.org/scriptableinputstream;1 nsIScriptableInputStream init
```

I double checked that the build file is updated and I'm pretty sure babelified.

### DIFF
--- a/js/actions/tests/loadSourceText.js
+++ b/js/actions/tests/loadSourceText.js
@@ -23,8 +23,9 @@ const mockThreadClient = {
   }
 };
 
-describe("loadSourceText", () => {
+describe("loadSourceText", function() {
   beforeEach(function() {
+    console.log("beforeEach");
     this.store = createStore(mockThreadClient);
   });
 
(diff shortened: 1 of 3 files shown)


js/actions/tests/loadSourceText.js
```diff
@@ -23,8 +23,9 @@ const mockThreadClient = {
   }
 };
 
-describe("loadSourceText", () => {
+describe("loadSourceText", function() {
```
jasonLaster: `() => {}` will make the context of the body the context of the outer describe, which messes with mocha's funny nested context property where `this` is shared.

jlongster:
Cool!

jasonLaster:
Still needs some work.

PR #88, opened 2016-04-22
Add additional loadSourceText tests

jasonLaster:
Adds some additional tests:
+ two different sources
+ same source
+ pending source request
+ failed source request

Implementation:

+ I noticed that handling pending and failed requests required managing the client's promise, so i created a deferred mock client.
+ There's quite a bit of promise chaining going on already. It'd be nice to go in later and see if `Task` can help us.

 I'm sure that these abstractions will continue to be cleaned up as we go

Here's where we're at.

screen shot 2016-04-22 at 11 08 23 am


jasonLaster:
I was curious what adding Task would look like, so I refactored the test, which had the longest then chain.

js/actions/tests/loadSourceText.js
```diff
@@ -23,20 +29,111 @@ const mockThreadClient = {
...
+      this.store = createStore(simpleMockThreadClient);
       this.store.dispatch(actions.loadSourceText({ actor: "foo1" }))
+        .then(() => {
+          return this.store.dispatch(actions.loadSourceText({ actor: "foo2" }));
+        })
         .then(() => done());
     });
 
+    it("Store has the first source text", function() {
+      const fooSourceText = queries.getSourceText(
```
jlongster: fwiw, if you use a query a lot in a test, you can still destructure it at the top (`const { getSourceText } = queries`).

js/actions/tests/loadSourceText.js
```diff
@@ -23,20 +29,111 @@ const mockThreadClient = {
...
+
+    it("Store has a loading source text", function() {
+      const fooSourceText = queries.getSourceText(
+                              this.store.getState(), "foo1");
+      expect(fooSourceText.get("loading")).to.equal(true);
+    });
+  });
+
+  describe("source failed to load", function() {
+    beforeEach(function(done) {
```
jlongster: Why use `beforeEach` when you only have one test?

It seems like there should be one top-level describe and then several "its". The describe should be just take the string "sources" and then the `it` blocks are something like `it("fails to load", ...)`. At least I thought that's how mocha worked (`describe` specifies the thing you are testing, and `it` is the action).

jasonLaster: I developed this style in part to avoid tests growing over time. Where the `beforeEach` handles the setup and the `it` handle the assertions. Often times, you might want to check a couple things and that way the test read, `when a source fails to load`, `the editor shows the error`, and `the error is saved in the store's source field`... The screenshot shows this a bit.

With that said, I'm okay putting the action in the `it` now as well.

jasonLaster: I'm going to merge and then make the tweaks

PR #89, opened 2016-04-25
Cleanup express server

jasonLaster:
Last week, it was hard to get the new express server to load the SVG icons. I believe the problem was the src path included "js", which is the webpack entry path. 

I believe the current solution `app.use("/js", express.static("js"))` is considered an express middleware hack. 

I noticed that I could clean it up `app.use(express.static("js"))` by updating the SVG src paths.

My concern with this is that it might make it difficult to load the debugger in the toolbox. With that said, we will probably need to do some flagging at that point anyways. Also, I'm by no means an express pro, so would be open to suggestions.


jlongster:
I'm confused about the problem. What is the problem exactly?

We could restructure the directories. When you have a server, usually you have a `client` or `static` directory that contains all of the `js`, `img`, etc directories. I'd like to do that since we are using a server now, and that probably solves the problem you were hitting because you can easily add and subdirectories.

PR #91, opened 2016-04-25
Add Storybook

jasonLaster:
Storybook is an environment for documenting / iterating on the components in our App. It has many of the benefits of our unit tests, but with a product bias.

This patch adds:
+ storybook config
+ some storybook "stories" for the accordion and breakpoints

My hope is that we can experiment with this style of development / documentation and see if it works for us. Also, maybe utilize it for some component unit tests as well.

storybook

### DIFF
--- a/.storybook/webpack.config.js
+++ b/.storybook/webpack.config.js
@@ -1,0 +1,19 @@
+"use strict";
+
+const path = require("path");
+
+module.exports = {
+  resolve: {
+    alias: {
+      "devtools": "ff-devtools-libs",
+      "sdk": "ff-devtools-libs/sdk"
+    },
+    extensions: ["", ".js", ".jsm"],
+    root: path.join(__dirname, "node_modules")
+  },
+  module: {
+    loaders: [
+      { test: /\.css$/, loader: "style-loader!css-loader" }
+    ]
+  }
+};
--- a/js/components/stories/Breakpoints.js
+++ b/js/components/stories/Breakpoints.js
@@ -1,0 +1,99 @@
+"use strict";
+
+const React = require("react");
+const { DOM: dom, createElement } = React;
+const { storiesOf } = require("@kadira/storybook");
+const configureStore = require("../../create-store");
+const { combineReducers } = require("redux");
+const reducers = require("../../reducers");
+const { Provider } = require("react-redux");
+const { fromJS } = require("immutable");
+
+const Breakpoints = React.createFactory(require("../Breakpoints"));
+
+const fixtures = {
+  sources: {
+    "fooSourceActor": {
+      actor: "fooSourceActor",
+      url: "http://example.com/foo.js"
+    },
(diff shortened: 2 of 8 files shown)


js/components/stories/Breakpoints.js
```diff
@@ -0,0 +1,99 @@
...
+        actor: "barSourceActor",
+        line: 18
+      },
+    }
+  }
+};
+
+storiesOf("Breakpoints", module)
+  .add("No Breakpoints", () => {
+    const store = configureStore({})(combineReducers(reducers), {
```
jlongster: I like this as documentation. It would be nice to make this as clear as possible though if it's going to be used as component docs. What if you made `renderBreakpoints` take the state instead, so none of these stories would to do the `configureStore` stuff? All it would do is pass the state.

server.js
```diff
@@ -9,7 +9,7 @@ const webpackDevMiddleware = require("webpack-dev-middleware");
 const app = express();
 const compiler = webpack(config);
 
-app.use("/js", express.static("js"));
```
jlongster: Why'd you change this? Any specific reason for the storybook? I personally like the URL matching the filesystem, and we may add other directories later.

jasonLaster: I changed it because storybook's server expects it in this format `app.use(express.static("js"))` and based on the express [docs](http://expressjs.com/en/starter/static-files.html) it seemed like the preferred style. I don't think storybook's use case is important enough to warrant the change (we could find another way to load the SVGs, but I would like to take into account express best practices).

jasonLaster:
I dropped the `svg` commit from this branch because that conversation should not block this PR.

.storybook/webpack.config.js
```diff
@@ -0,0 +1,19 @@
...
+    },
+    extensions: ["", ".js", ".jsm"],
+    root: path.join(__dirname, "node_modules")
+  },
+  module: {
+    loaders: [
+      { test: /\.css$/, loader: "style-loader!css-loader" }
+    ]
+  }
+};
```
jasonLaster: this is mostly a placeholder, I think in reality we can re-use much of the existing webpack.config.js

js/components/stories/Breakpoints.js
```diff
@@ -0,0 +1,99 @@
...
+    "barBreakpointActor": {
+      actor: "barBreakpointActor",
+      location: {
+        actor: "barSourceActor",
+        line: 18
+      },
+    }
+  }
+};
+
```
jasonLaster: This fixture data hopefully can move to its own file soon and be re-used by component unit tests

js/components/stories/Breakpoints.js
jasonLaster: yea, i like that idea. I hope that we'll add better abstractions as we go. Something like `test-head`

jlongster:
Nice!

PR #92, opened 2016-04-25
Move JS to public

jasonLaster:
This moves the JS folder to the public directory.

This does not move the build directory to public. When I tried to do that I ran into issues with webpack-dev-middleware failing to intercept requests for `build/bundle.js` I'm sure this is fixable, but I wanted to get this move out of the way as it's the big show stopper.


jlongster:
Looks like we need to update some of our npm tasks (look at the circleci failures)

jasonLaster:
wilco

PR #95, opened 2016-04-25
Make README more explicit

jasonLaster:


.gitignore
```diff
@@ -1,3 +1,4 @@
+.DS_Store
```
jasonLaster: not really sure how this was removed

jasonLaster:
Gabe mentioned: "you should describe that the browser that is open with that command becomes the sources that you can inspect". I think this makes sense

jlongster:
Sounds good!

PR #97, opened 2016-04-26
Add Query Presenters

jasonLaster:
Our queries up to this point have focused on data access. For example, get a source by source actor or get a breakpoint by location.

There are three limitations here:

1) Inconsistency

+ The pause state has a `frame.where.source` property and the breakpoint state has a `location.actor` property. This PR adds a pause `frame.where.actor` property and a breakpoint `location.source` property. The two states are now consistent.

2) Incomplete

+ The breakpoint state has a `location.actor` property, but does not have a `frame.where.source` property. This is a problem because when a breakpoint is shown in the breakpoint list, its source is needed.
+ The source state has a `url` property, but does not have a `filename` property, which would be helpful when showing a breakpoint.

3) Performance

+ Because each component is managing it's own data access / presentation, it is harder to memoize without setting intermediate component state. By moving the presentation logic to the queries, we can use tools like re-select to cache the computed properties and denormalized state.

Generalizations:

+ Computed Properties: queries should be responsible for formatting and computing derived properties
+ Actor denormalization: if a query has an actor field, it should denormalize the state. i.e. the breakpoint has a `location.actor` field and should add a `location.source` field.  

Presenters:

Data access and data presentation are fundamentally different problems. With that in mind, I added separate presentBreakpoint, presentSource functions that the queries use to return presented data. 

Will this conflict with your breakpoint work? Yes! but that's okay.

### DIFF
--- a/public/js/components/Breakpoints.js
+++ b/public/js/components/Breakpoints.js
@@ -4,27 +4,15 @@ const React = require("react");
 const { connect } = require("react-redux");
 const { bindActionCreators } = require("redux");
 const actions = require("../actions");
-const { getSources, getBreakpoints } = require("../selectors");
+const { getSource, isCurrentlyPausedAtBreakpoint, getBreakpoints, makeLocationId } = require("../selectors");
 const ImPropTypes = require("react-immutable-proptypes");
 const { DOM: dom, PropTypes } = React;
 
 require("./Breakpoints.css");
 
-function getFilenameFromSources(sources, actor) {
-  const source = sources.get(actor);
-  if (source.get("url")) {
-    const url = new URL(source.get("url"));
-    const filename = url.pathname.substring(
-      url.pathname.lastIndexOf("/") + 1);
-    return filename;
-  }
-  return "";
-}
-
 const Breakpoints = React.createClass({
   propTypes: {
     breakpoints: ImPropTypes.map.isRequired,
-    sources: ImPropTypes.map.isRequired,
     enableBreakpoint: PropTypes.func.isRequired,
     disableBreakpoint: PropTypes.func.isRequired,
   },
@@ -42,16 +30,13 @@ const Breakpoints = React.createClass({
   },
 
   renderBreakpoint(breakpoint) {
-    const sourceActor = breakpoint.getIn(["location", "actor"]);
-    const filename = getFilenameFromSources(
-      this.props.sources,
-      sourceActor
-    );
+    const filename = breakpoint.getIn(["location", "source", "filename"]);
+    const locationId = breakpoint.get("locationId");
     const line = breakpoint.getIn(["location", "line"]);
 
     return dom.div(
       { className: "breakpoint",
-        key: `${sourceActor}/${line}` },
+        key: locationId },
       dom.input({ type: "checkbox",
                   checked: !breakpoint.get("disabled"),
                   onChange: () => this.handleCheckbox(breakpoint) }),
@@ -73,10 +58,21 @@ const Breakpoints = React.createClass({
   }
 });
 
+function _getBreakpoints(state) {
+  return getBreakpoints(state).map(breakpoint => {
+    const source = getSource(state, breakpoint.getIn(["location", "actor"]));
+    const isCurrentlyPaused = isCurrentlyPausedAtBreakpoint(state, breakpoint);
+    const locationId = makeLocationId(breakpoint.get("location").toJS());
+    return breakpoint.setIn(["location", "source"], source)
+                     .set("locationId", locationId)
+                     .set("isCurrentlyPaused", isCurrentlyPaused);
+  });
+}
+
+
 module.exports = connect(
   (state, props) => ({
-    sources: getSources(state),
-    breakpoints: getBreakpoints(state)
+    breakpoints: _getBreakpoints(state)
   }),
   dispatch => bindActionCreators(actions, dispatch)
 )(Breakpoints);
--- a/public/js/components/Editor.js
+++ b/public/js/components/Editor.js
@@ -66,7 +66,7 @@ const Breakpoint = React.createFactory(_Breakpoint);
 
 const Editor = React.createClass({
   propTypes: {
-    breakpoints: ImPropTypes.list,
+    breakpoints: ImPropTypes.map.isRequired,
     selectedSource: PropTypes.object,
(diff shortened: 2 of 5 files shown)


public/js/components/Breakpoints.js
```diff
@@ -2,52 +2,34 @@
...
-function getFilenameFromSources(sources, actor) {
-  const source = sources.get(actor);
-  if (source.get("url")) {
-    const url = new URL(source.get("url"));
-    const filename = url.pathname.substring(
-      url.pathname.lastIndexOf("/") + 1);
-    return filename;
-  }
-  return "";
-}
```
jasonLaster: really happy to kill this

public/js/components/Breakpoints.js
```diff
@@ -65,7 +47,6 @@ const Breakpoints = React.createClass({
 
 module.exports = connect(
   (state, props) => ({
-    sources: getSources(state),
```
jasonLaster: happy to remove this dependency

jlongster:
I think you are on to something, but I would prefer to just call them "selectors" which is what the general redux world uses. You're basically citing the reasons that reselect exists. I'm fine delaying pulling in that dependency, but we could go ahead and change `queries.js` to `selectors.js` and call it a day. I'm not convinced we need to differentiate between queries/presenters.

One reason you might feel that it does not belong as a query is because it's purely presentation-related. I think some of the ones you added here should just go directly in `Editor.js`. It's common in the redux world to define selectors straight in the component file, since those relate only to that component. Commonly used ones can go in `selectors.js` that everything can use. I'd prefer to only pull things into there that need to be used by 2-3+ components. (There's an exponential number of various ways to combine the data, and components should generally combine it the way that they need specifically).

I don't think it's code smell for the breakpoint list to depend on sources. You still depend on sources, but instead of combining the data in the component, you're doing in what's basically a selector. That is a better way to do it, but I think that should happen directly in `Editor.js` as a selector. Maybe we should just go ahead and pull in reselect. I haven't looked at their docs, but I bet they show examples of defining selectors straight in the component file to create data that the component renders.

What might also be confusing is inconsistency in the debugger protocol. It might not be consistent with what a `location` object means, etc. The other place to "normalize" data is in the reducers; if something really should be consistent across the entire app, the reducer should normalize the data and store it in a way that makes sense.

I'd like to change `queries.js` to `selectors.js`, and still just keep very generic selectors in there. Start defining component-level selectors. And lastly, normalize data in reducers where things like `location` objects are inconsistent.

jlongster:
Looking at your paused state, I think what you are mostly going for here is normalizing the paused state. You should just do this in the `pause` reducer. For example, if it's common to need a `isInterrupted` property, just make that part of the `pause` state. The reducer can take the raw pause packet and normalize it however it wants, it doesn't even need to reflect the raw pause packet at all. That's actually better because then you know that it always happens once no matter how many times it's selected.

public/js/components/Breakpoints.js
```diff
@@ -73,10 +58,21 @@ const Breakpoints = React.createClass({
...
+function _getBreakpoints(state) {
+  return getBreakpoints(state).map(breakpoint => {
+    const source = getSource(state, breakpoint.getIn(["location", "actor"]));
+    const isCurrentlyPaused = isCurrentlyPausedAtBreakpoint(state, breakpoint);
+    const locationId = makeLocationId(breakpoint.get("location").toJS());
+    return breakpoint.setIn(["location", "source"], source)
+                     .set("locationId", locationId)
+                     .set("isCurrentlyPaused", isCurrentlyPaused);
+  });
+}
```
jasonLaster: I think you're right that this selector makes more sense inside of the Breakpoints component. Good call

public/js/components/Breakpoints.js
```diff
@@ -42,16 +30,13 @@ const Breakpoints = React.createClass({
...
-      sourceActor
-    );
+    const filename = breakpoint.getIn(["location", "source", "filename"]);
+    const locationId = breakpoint.get("locationId");
     const line = breakpoint.getIn(["location", "line"]);
 
     return dom.div(
       { className: "breakpoint",
-        key: `${sourceActor}/${line}` },
+        key: locationId },
```
jasonLaster: just a small cleanup

public/js/components/Editor.js
```diff
@@ -66,7 +66,7 @@ const Breakpoint = React.createFactory(_Breakpoint);
 
 const Editor = React.createClass({
   propTypes: {
-    breakpoints: ImPropTypes.list,
+    breakpoints: ImPropTypes.map.isRequired,
```
jasonLaster: fixing a warning

public/js/reducers/sources.js
```diff
@@ -16,7 +16,9 @@ const initialState = fromJS({
 function update(state = initialState, action) {
   switch (action.type) {
     case constants.ADD_SOURCE:
-      return state.mergeIn(["sources", action.source.actor], action.source);
+      let newSource = action.source;
```
jlongster: I'd like to avoid mutating something that came in because it might be referenced elsewhere. What do you think of this?

```
let source = Object.assign({}, action.source, {
  filename: getFilenameFromUrl(newSource.url)
})
```

This also seems like something that the component could do when selecting this data, but I'm fine putting it here instead. I just realized that we're going to have to think through how we represent various things like sources and locations in a world where we will be talking to multiple servers (firefox/chrome/etc). They each will return different types of objects. Now sure how that will look yet, but we'll figure it. (For example, Chrome doesn't have an "actor" but has "scriptId")

jlongster:
Looks good! The only thing I would suggest changing is using `Object.assign`, unless you feel like that's too much work. Otherwise feel free to merge!

public/js/reducers/sources.js
jasonLaster: Good call.

jasonLaster:
Yea - there will probably be an adapter layer between the client and actions... This will be a good thing as it will force our components / actions to be generic

public/js/reducers/sources.js
jlongster: The only downside to adding more "derived" properties in the state is that any of our mocks will have to also specify it. I know I said that you can "normalize" data here but to me this is more "derived data". But I don't really care that much, if you think it should be here that's fine!

jasonLaster:
done!

Issue #74, opened 2016-04-18
Test scripts should find all js test files


jasonLaster:
Currently the node + browser test scripts only file test files in a test folder. The script should be able to recursively search for test folders and then add tests files.

PR #98, opened 2016-04-26
Cleanup Test Infrastructure

### LINKED ISSUE #90: Convert browser test script to a server
Currently, the browser test workflow is two steps: 

1) run `npm run browser-test` 
2) refresh browser tab with the tests. 

If we create a browser test server, similar to debugger server, then the browser test build will be built on the fly and returned.

jasonLaster:
This does a whole lot of little cleanup. Resolves #90 and #74

+ create getTestPaths
+ lint node-test
+ move mocha runner out of test to be able to delete that dir
+ create test browser server
+ remove test-browser script
+ move server to bin

List of wins:

+ tests files can be anywhere in `public/js`. You can still run `node bin/test public/js/actions` to limit the path of tests
+ the browser test runner now updates on refresh. This is huge for me because it's my primary way of running the tests


jlongster:
I don't fully follow all the new testing stuff, but the code itself looks OK to me! I'm fine merging this in however it works. I'll let you push the button in case you were going to add anything more here

jasonLaster:
Okay. I'm sure you'll have more opinions when you're writing tests

PR #105, opened 2016-04-26
SelectSource Tests + Refactoring

jasonLaster:
Added two select source tests

Refactoring:
+ created a fixtures.json (cleaned up some fixture drift)
+ removed intermediate describe/beforeEach

I'd like to considering having a util file for actions so that the two mock clients could go there. One problem is that currently any non-hidden file in a test dir will be considered a test file. What do you think of having a suffix like spec (e.g. selectSource.spec.js)?

### DIFF
--- a/.storybook/webpack.config.js
+++ b/.storybook/webpack.config.js
@@ -13,7 +13,8 @@ module.exports = {
   },
   module: {
     loaders: [
-      { test: /\.css$/, loader: "style-loader!css-loader" }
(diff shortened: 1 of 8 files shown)


jlongster:
❤️ Thanks so much for working on this, after spending a long time on our current mochitests this looks like heaven. `.spec.js` sounds fine to me, I don't really care :)

jasonLaster:
@jlongster happy to. Looking forward to getting your perspective when you're in there

PR #107, opened 2016-04-27
move build directory to public to consolidate static assets

jlongster:
@jasonLaster I tested everything and they all still work from my end. For something like this, definitely want you to review the code though.

### DIFF
--- a/bin/server-browser-test.js
+++ b/bin/server-browser-test.js
@@ -11,11 +11,7 @@ const getTestPaths = require("./getTestPaths");
 const testPaths = getTestPaths(path.join(__dirname, "../public/js"));
 
 const config = Object.assign({}, projectConfig, {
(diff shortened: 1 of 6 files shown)


bin/server-browser-test.js
```diff
@@ -11,11 +11,7 @@ const getTestPaths = require("./getTestPaths");
 const testPaths = getTestPaths(path.join(__dirname, "../public/js"));
 
 const config = Object.assign({}, projectConfig, {
-  entry: testPaths.concat(path.join(__dirname, "../public/js/main.js")),
-  output: {
-    path: path.join(__dirname, "../build"),
-    filename: "bundle.js"
-  },
```
jasonLaster: nice

jasonLaster:
Thanks. It's easy to miss one of the cases

PR #109, opened 2016-04-27
Update dependencies

jasonLaster:
Updating the non-react dependencies. shouldn't be anything major in here.


jlongster:
Sounds good. We can just remove seamless-immutable as well, you can do that here if you want?

francoislaberge:
@jlongster After reading a recent video of yours explaining the issues you were having using immutablejs and mixing with native components. I was going to suggest trying out seamless, but looks like you removed it here. What was your take on it? 

Do you have a good write up somewhere about where your current thinking is on how to reduce confusion in mixing immutable and regular json structures? It's just the type system embracement that made you able to mix it better? 

I am about to start a similar project and have been deeply pondering this issue of how far to go into using immutablejs, I'd prefer to use native structures but have guarantees that no code ended up being immutable unless explicitly doing so.

PR #110, opened 2016-04-27
Replace queries with selectors

jasonLaster:
simple find and replace. More to come


jlongster:
LGTM, you can merge! (not merging becaus unsure if you wanted to add more in here or another PR)

PR #114, opened 2016-04-28
Refactor Sources

jasonLaster:
Started working on Call Stacks and saw an opportunity to cleanup Sources. Nothing scary here.

+ add _updateText formatting function for sources
+ cleanup sources component

### DIFF
--- a/public/js/reducers/sources.js
+++ b/public/js/reducers/sources.js
@@ -16,9 +16,8 @@ const initialState = fromJS({
 function update(state = initialState, action) {
   switch (action.type) {
     case constants.ADD_SOURCE:
-      let newSource = action.source;
-      newSource.filename = getFilenameFromUrl(newSource.url);
-      return state.mergeIn(["sources", action.source.actor], newSource);
+      return state.mergeIn(["sources", action.source.actor],
+                            _updateSource(action.source));
 
     case constants.LOAD_SOURCES:
       if (action.status === "done") {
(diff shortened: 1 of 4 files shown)


jasonLaster:
Nice -> Tests doing what they're supposed to do:
https://circleci.com/gh/jasonLaster/debugger.html/67

public/js/reducers/sources.js
```diff
@@ -81,13 +79,19 @@ function update(state = initialState, action) {
...
+  const filename = pathname.substring(pathname.lastIndexOf("/") + 1);
+  return Object.assign({}, source, { pathname, filename });
+}
+
+function getPathnameFromUrl(sourceUrl) {
   if (!sourceUrl) {
     return "";
   }
 
   const url = new URL(sourceUrl);
```
jlongster: I'm convinced now that that should live somewhere else and we should make a selector that extracts the filename from a source. The `URL` is undefined in node and none of this code should depend on the browser.

jasonLaster: Yea, that makes a lot of sense. We could do a simple feature check `if (URL) {} else {}`

jlongster: If it's a selector and used from a component, I think it's ok to use `URL` and assume it's a browser environment. We don't test components in node, we'll run them in the browser, from what I understand

jasonLaster: > We don't test components in node, we'll run them in the browser

Yea, we can decide to do that. Lets wait and see though, it might be simpler or significantly faster to run in node. Not sure yet.

> selectors

I'll move it over now

jlongster:
+1 mergable!

PR #115, opened 2016-04-29
Improve mocha testing

jlongster:
This looks like a lot of changes, but most of it is just update requires. I decided to avoid using the special "devtools" module path. The reason is that there's no good way in node to add this alias, so it makes it much harder when running tests under node. This, combined with using the babel compiler directly from the mocha option, allows us to remove webpack entirely from the testing process.

Removing webpack speeds up tests 10x (don't ask for me to backup that number). It's a whole lot faster. If we ever move files from gecko-dev to here, we'll have to update the requires. But that's extremely straight-forward, so it's not a big deal. It happens rarely anyway.

There's a lot less magic this way.

There is still some weirdness: the ff-devtools-libs essentially depends on itself. All the requires inside it assume that `ff-devtools-libs` is a top-level modules so all requires are of the form `ff-devtools-libs/util/foo.js`. This is very against the normal node pattern. However, it's pretty simple to solve: install the module normally, and anything that runs the code needs to set `NODE_PATH` to `node_modules` so that the magical top-level `ff-devtools-libs` name will be resolved anywhere. You can see this in `webpack.config.js` too: we point a root to `node_modules` so it'll resolve. Running tests under `npm test` automatically set the `NODE_PATH` to node_modules so it magically works when running under npm scripts, so you shouldn't have to think about it unless working on the build/testing system.

Now there are two commands for running mocha:

* `npm test` - runs all the tests (note: I think the path selector I used only works under zsh, that might be a problem?)
* `npm run test-paths <path>...` - runs tests found at the specified paths (can be directories or single files)

I'm sure there are things to discuss here, we'll talk about it tomorrow.


jlongster:
This is good to go. There's a faulty test on master that is unrelated to this. We will fix that soon, but that shouldn't block this.

PR #120, opened 2016-04-29
Add selectBreakpoint handler on breakpoint list

jasonLaster:
Adds simple navigation for selecting a breakpoint.


jasonLaster:
Was thinking about adding it for selecting frames in the call stack as well.

jlongster:
The `selectSource` action is what things call to show a source at a specific place, yes, both this and the frames will call it. I will add it to the frames as I am working on it. This looks fine, you can merge once tests pass.

PR #121, opened 2016-04-29
add basic call stack

jlongster:
Don't merge this yet, but just putting it up early. This fetches a chunk of the frames and simply lists the names of the functions in the stack. I'll add more info before this merges. The current debugger loads frames in "pages" because you don't want to request all frames at once because that will be really slow for deep stacks. For now, we could just load in unlimited frames and do the optimization later. It will require some sort of "load the next chunk on scroll" behavior.


public/js/actions/pause.js
```diff
@@ -18,8 +22,22 @@ function resumed() {
  * Debugger has just paused
  */
 function paused(pauseInfo) {
-  return ({ dispatch }) => {
-    return dispatch({
+  return ({ dispatch, threadClient }) => {
+    // If paused by an explicit interrupt, which are generated by the
+    // slow script dialog and internal events such as setting
+    // breakpoints, ignore the event.
+    if(pauseInfo.why.type === "interrupted" &&
```
jasonLaster: you can do `pauseInfo.isInterrupted`

public/js/actions/pause.js
```diff
@@ -18,8 +22,22 @@ function resumed() {
...
+       !pauseInfo.why.onNext) {
+      return;
+    }
+
+    dispatch({
+      type: constants.LOAD_FRAMES,
+      [PROMISE]: promisify(threadClient, threadClient.getFrames,
+                           0, CALL_STACK_PAGE_SIZE)
+    });
+
```
jasonLaster: Maybe also move the selectSource action here too https://github.com/jlongster/debugger.html/blob/master/public/js/client.js#L49

public/js/components/RightSidebar.js
```diff
@@ -48,8 +49,8 @@ function RightSidebar({ resume, command, breakOnNext,
             component: Breakpoints,
             opened: true },
           { header: "Call Stack",
-            component: () => dom.div({className: "pane-info"}, "Not Paused")
-          },
+            component: Frames,
+            componentProps: { frames }},
```
jasonLaster: any reason why you prefer this over `connect` in the component, which is what Breakpoints and Scopes does.

public/js/actions/pause.js
```diff
@@ -18,8 +22,22 @@ function resumed() {
...
+    // slow script dialog and internal events such as setting
+    // breakpoints, ignore the event.
+    if(pauseInfo.why.type === "interrupted" &&
+       !pauseInfo.why.onNext) {
+      return;
+    }
+
+    dispatch({
+      type: constants.LOAD_FRAMES,
+      [PROMISE]: promisify(threadClient, threadClient.getFrames,
```
jasonLaster: nice

public/js/components/RightSidebar.js
jlongster: I don't love how often we are "connecting" components, and I would prefer to pull state off from above and pass it down more often. This is somewhat subjective, but it's usually preferred to do it sparingly. I'm not clear where the balance is though, and once we implement these last set of features this is a good thing to think about across the project as a whole.

public/js/actions/pause.js
jlongster: I will look at that tomorrow!

jlongster:
Got something closer to a final version working. Still need to clean up the code and think through a few places, particularly the editor. They way we were highlighting a specific line wasn't really working the way it should. Just wanted to get this working before I leave today, it's not final code though.

The next commit changes a few things, like how the utility function `createStore` that the stories use. I need to create a store with frame data, and I wanted it to be more generic. We can talk about it tomorrow.

You can now see a call stack and select various frames, and will be taken to that location:

screen shot 2016-05-02 at 5 21 37 pm

jlongster:
That screenshot was an earlier one actually, forgot to add some class names. Here is what it looks like with a thin pane:

screen shot 2016-05-02 at 5 28 00 pm

If you make it wider it all fits. Will polish this some more...

screen shot 2016-05-02 at 5 28 09 pm

jasonLaster:
Looks great! I'll review more closely tomorrow

public/js/components/Editor.js
```diff
@@ -143,20 +131,26 @@ const Editor = React.createClass({
...
 
-  componentWillReceiveProps(nextProps) {
-    this.setSourceText(nextProps.sourceText, this.props.sourceText);
-    this.goToLine(nextProps.selectedSourceOpts);
-    this.clearDebugLine(this.props.pause);
-    this.setDebugLine(nextProps.pause);
+    if(nextProps.selectedSourceOpts &&
+       nextProps.selectedSourceOpts.get("line")) {
+      this.setDebugLine(nextProps.selectedSourceOpts.get("line"));
+    } else if(nextProps.pause &&
```
jasonLaster: `nextProps.pause.get('isInterrupted`)

jasonLaster: I think we can extract this to `setEditorLine` or something like that

public/js/components/stories/Breakpoints.js
```diff
@@ -28,8 +28,8 @@ storiesOf("Breakpoints", module)
   })
   .add("2 Domains", () => {
     const store = createStore({
-      sources: { fooSourceActor, barSourceActor },
-      breakpoints: { fooBreakpointActor, barBreakpointActor }
+      sources: { sources: { fooSourceActor, barSourceActor }},
+      breakpoints: { breakpoints: { fooBreakpointActor, barBreakpointActor }}
```
jasonLaster: makes sense.

public/js/components/stories/Frames.js
```diff
@@ -0,0 +1,32 @@
...
+storiesOf("Frames", module)
+  .add("Blank", () => {
+    return renderContainer(Frames);
+  })
+  .add("Paused", () => {
+    return renderContainer(Frames, { frames: frameData });
+  });
+
+function renderContainer(Component, props) {
+  const store = createStore();
```
jasonLaster: do you need a store here?

public/js/components/stories/utils.js
```diff
@@ -5,12 +5,12 @@ const { combineReducers } = require("redux");
...
-function createStore({ sources = {}, breakpoints = {}, pause = null } = {}) {
-  return configureStore({})(combineReducers(reducers), {
-    sources: fromJS({ sources }),
-    breakpoints: fromJS({ breakpoints }),
-    pause: fromJS({ pause })
-  });
+function createStore(state = {}) {
+  for(var k in state) {
+    state[k] = fromJS(state[k]);
+  }
```
jasonLaster: shame there isn't a native map function

jlongster: Yeah. `mapValues` might be something we use enough to have as a utility lib without pulling down all of lodash. If we end up being able to use webpack for real in the long-term, I'd be fine requiring specific lodash functions in.

public/js/components/stories/Breakpoints.js
jlongster: a little annoying, yeah, but reflects the real state shape

public/js/components/stories/Frames.js
jlongster: I do because the Frames component still uses `connect` to get the action creator `selectSource`. That is confusing though.

I'm going to think some more about the pattern for "connecting" components. For now I might just make `Frames` pull the state off too because everything is doing that.

jasonLaster:
Looks like circle is just upset about the linting: https://circleci.com/gh/jlongster/debugger.html/63

jasonLaster:
after this lands it might be nice to add some unit tests for actions.

jlongster:
> Looks like circle is just upset about the linting: https://circleci.com/gh/jlongster/debugger.html/63

Yeah I haven't polished this up yet. I will now and get it ready to land.

PR #123, opened 2016-05-02
Add Eslint object-curly-spacing rule

jasonLaster:
This will enforce a whitespace character between objects and their
properties. It does not force nested object curlies to have whitespace
e.g. { x: { a: 2 }} is okay.

I think it is okay to change eslint if we're becoming stricter, of course going the other way is different.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -297,7 +297,7 @@
     // Disallow unreachable statements after a return, throw, continue, or break
     // statement.
     "no-unreachable": 2,
-    // Disallow global and local variables that aren't used, but allow unused function arguments.
+    // Disallow global and local variables that arent used, but allow unused function arguments.
     "no-unused-vars": [2, {"vars": "all", "args": "none"}],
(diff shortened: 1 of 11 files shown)


.eslintrc
```diff
@@ -297,7 +297,7 @@
     // Disallow unreachable statements after a return, throw, continue, or break
     // statement.
     "no-unreachable": 2,
-    // Disallow global and local variables that aren't used, but allow unused function arguments.
+    // Disallow global and local variables that arent used, but allow unused function arguments.
```
jasonLaster: The syntax highlighting with apostrophe's is a pet peeve of mine. I'm not sure which is worse, breaking the highlighting or using bad english. Feel free to overrule this decision :)

jlongster:
Sounds good to me. I'm surprised devtools doesn't have it in theirs by default.

jasonLaster:
yea, i guess that would upset eddy if we did that right now :)

PR #125, opened 2016-05-02
Scopes (Part 2)

jasonLaster:
Updates:
+ Refactored Storybook's `createStore` to support fixtures
+ Change block scope behavior to always show (consistent w/ firefox)
+ Change function scope labels to support anonymous and always have the word function.
+ Add scope accordions for styling and collapsing large scopes

### DIFF
--- a/public/js/components/stories/Breakpoints.js
+++ b/public/js/components/stories/Breakpoints.js
@@ -3,56 +3,33 @@
 const React = require("react");
 const { DOM: dom, createElement } = React;
 const { storiesOf } = require("@kadira/storybook");
-const configureStore = require("../../create-store");
-const { combineReducers } = require("redux");
-const reducers = require("../../reducers");
 const { Provider } = require("react-redux");
-const { fromJS } = require("immutable");
+const { createStore } = require("./utils");
 
 const Breakpoints = React.createFactory(require("../Breakpoints"));
 const fixtures = require("../../test/fixtures.json");
 
+const fooSourceActor = fixtures.sources.sources.fooSourceActor;
+const barSourceActor = fixtures.sources.sources.barSourceActor;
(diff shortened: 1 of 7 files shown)


public/js/components/stories/Breakpoints.js
```diff
@@ -3,56 +3,33 @@
...
-          "fooSourceActor": fixtures.sources.sources.fooSourceActor,
-          "barSourceActor": fixtures.sources.sources.barSourceActor,
-        }
-      }),
-      breakpoints: fromJS({
-        breakpoints: {
-          "fooBreakpointActor": fixtures.breakpoints.fooBreakpointActor,
-          "barBreakpointActor": fixtures.breakpoints.barBreakpointActor,
-        }
-      })
```
jasonLaster: Really happy to have this red

public/js/test/fixtures/backbone.pause.onEnter.json
```diff
@@ -0,0 +1,502 @@
+{
```
jasonLaster: There are a couple things I don't love here:
1. having "fake" data with fixtures
2. having a dump of pause data here. 

in some ways both are probably fine, it's just important to see what balance we like for our tests/stories

jlongster: Yeah. I'm not sure what maintaining this is going to look like. It might be possible to automate generating these fixtures though: setup a devtools connection and automatically go to the right place and dump the protocol response to these files. It seems useful to have them, so let's just see how it feels over time.

jlongster:
Nice! I'll merge it.

public/js/test/fixtures/backbone.pause.onEnter.json
jasonLaster: > maintaining it
Yeah, that's a good point.

PR #128, opened 2016-05-02
Update Configuration

jasonLaster:
Currently we use an environment file for configuration.
This is fine, but there are a couple shortcomings

1) the environment file is read directly

2) the environment file is for your own local changes, which is not the
same as shared development features

3) it would not be obvious how to add production flags. This PR adds a
development.json file, but it would be easy to add a production.php file
and read from it when in production.

4) lastly, there aren't any nice helpers. This PR adds an isEnabled
feature, but ultimately it would be nice to improve it to support path
notation and maybe other functionality as well.


public/js/configs/tests/.eslintrc
```diff
@@ -0,0 +1,32 @@
...
+    "expect": true,
+    "equal": true,
+    "ok": true
+  },
+  "rules": {
+    "camelcase": 0,
+    "no-unused-vars": [2, {"vars": "all", "args": "none", "varsIgnorePattern": "run_test"}],
+    "max-nested-callbacks": [2, 4],
+  }
+}
```
jasonLaster: it's frustrating to have to copy the eslint file. We can probably clean this up by adding globals to the primary eslint... with something like `Object.assign({}, config.globals, mochaGlobals)` I beleive there's a project for that. We could also do something else too...

jasonLaster:
the `development` flag could be replaced with something like this: `https://github.com/jlongster/debugger.html/issues/55#issuecomment-209563759`

public/js/configs/tests/.eslintrc
jlongster: Agreed, it's annoying. Not sure I follow, how do you add those globals? The top-level .eslintrc is just a JSON file. But do whatever works :)

jlongster:
What about moving to a more powerful config tools like [`nconf`](https://github.com/indexzero/nconf) instead? It's very flexible and allows all the code to access the config with just `nconf.get("logging")`. It supports the ability to have a "base" config and then override it, etc.

public/js/configs/development.json
```diff
@@ -0,0 +1,4 @@
+{
```
jlongster: I feel like the configs still belong at the top-level. We could make a top-level `configs` directory. We'll probably end up wanting to put some other stuff in there, like the `port` that the server runs on.

jasonLaster:
hmm - will investigate `nconf`

public/js/configs/development.json
jasonLaster: yeah - i think tha's right

jlongster:
I'm fine with this PR too if you want a simple solution.

public/js/configs/tests/.eslintrc
jasonLaster: it can be it can be JS as well... I've seen something like this before...

jasonLaster:
Merged this PR because I think it's an improvement over the environment.json work, which has a complicated README. I'll make a new issue which tracks `nconf`, moving configs to a top level directory, and enabling logging.

Logging is especially important because i want to figureout why the debugger does not detatch on refresh.

PR #130, opened 2016-05-02
Disable APZ for source editors. Fixes #129.

### LINKED ISSUE #129: Disable APZ for source editors
The current version of APZ (async pan zoom) can cause tearing effects in the source editor where the line numbers float around disconnected from the code.

See https://bugzilla.mozilla.org/attachment.cgi?id=8722367

gabrielluong:
Fixes #129.

r? @jlongster

### DIFF
--- a/public/js/components/Editor.js
+++ b/public/js/components/Editor.js
@@ -16,7 +16,7 @@ const { bindActionCreators } = require("redux");
 const { connect } = require("react-redux");
 
 const actions = require("../actions");
-const { alignLine } = require("../util/editor");
+const { alignLine, onWheel } = require("../util/editor");
 
 require("codemirror/lib/codemirror.css");
 require("./Editor.css");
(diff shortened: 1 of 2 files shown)


jasonLaster:
👍

gabrielluong:
Cancel the r? for now until I pass the eslint

jasonLaster:
@gabrielluong just fixed the eslint bug. it was a bug with a lint rule we added this morning and hadn't adjusted that pending PR.

public/js/components/Editor.js
```diff
@@ -87,6 +87,29 @@ const Editor = React.createClass({
     });
 
     this.editor.on("gutterClick", this.onGutterClick);
+
+    // Disable APZ for source editors. It currently causes the line numbers to
+    // "tear off" and swim around on top of the content. Bug 1160601 tracks
+    // finding a solution that allows APZ to work with CodeMirror.
+    this.editor.getScrollerElement().addEventListener("wheel", ev => {
```
jasonLaster: I think the onWheel logic should land in `util/editor.js` as it's a utility for the editor exclusively.

`this.editor.getScrollerElement().addEventListener("wheel", ev =>  onWheel(this.editor, ev);`

jlongster: I agree, let's keep the component clean and free of hacks. It's unfortunate that we have to do this.... but thanks a bunch for fixing it! I was noticing this and was wondering what was going on.

If you move this code to `util/editor.js` we'll merge this.

jasonLaster:
Looks great

PR #135, opened 2016-05-02
Add styles to the sources pane

jasonLaster:
Fun PR for styling the sources pane

A lot of the CSS is a place holder for the tree view

screen shot 2016-05-02 at 7 14 00 pm


public/js/test/fixtures.json
```diff
@@ -4,12 +4,14 @@
...
-        "filename": "foo.js"
+        "filename": "foo.js",
+        "pathname": "foo/foo.js"
       },
       "barSourceActor": {
         "actor": "barSourceActor",
         "url": "http://example.com/bar/bar.js",
-        "filename": "bar.js"
+        "filename": "bar.js",
+        "pathname": "bar/bar.js"
```
jasonLaster: i feel guilty adding filename and pathname to the fixtures. should really be a query

jlongster: Haha :)

jlongster:
I'm a little unsure of the `<>` icon, what is that supposed to mean?

If these are coming from Helen's designs, I don't think those were very final. Just worth mentioning. Feel free to merge.

Also, I was thinking of not having the filtering UI at all and maybe just a "Sources" header because we weren't going to do filtering. And if we are, I'm not sure we should dedicate that whole area to it... We should probably just do the Cmd+P thing

jasonLaster:
> <> icon 

that's coming from the designs. I think that can be added and then tweaked

> filtering UI

I'll take that out so we can think through the UX

jasonLaster:
> <> icon

my guess is we'll likely want to show a file icon like "atom", "sublime", "chrome dt"

jasonLaster:
merged

PR #139, opened 2016-05-03
reduce the repetition of frame location

clarkbw:
**Don't Merge This**
screen shot 2016-05-03 at 12 33 49 pm

This is just a hack attempt at only listing the frame location when its unique to the frame before it.

Currently you can the same frame source location repeated over and over.  While the line number may be unique and should be displayed we should avoid repeating ourselves like this.

### DIFF
--- a/public/js/components/Frames.js
+++ b/public/js/components/Frames.js
@@ -23,20 +23,27 @@ function renderFrameTitle(frame) {
   return div({ className: "title" }, title);
 }
 
-function renderFrameLocation(frame) {
-  return div({ className: "location" }, basename(frame.where.source.url));
+function renderFrameLocation(frame, lastFrame) {
+  if (!lastFrame || lastFrame.where.source.url !== frame.where.source.url) {
+    return div({ className: "location" }, basename(frame.where.source.url));
+  }
+  return div({ className: "location" }, "");
 }
 
 function Frames({ frames, selectedFrame, selectFrame }) {
+  let lastFrame = null;
   return div(
     { className: "pane-info frames" },
     !frames ?
       div({ className: "empty" }, "Not Paused") :
-      dom.ul(null, frames.map(frame => {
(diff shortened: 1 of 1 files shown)


jlongster:
I'm not sure about that, I think it's more confusing. When I'm trying to inspect a frame, having to look back up the frames to find the last one to find the source is a little annoying. I think the best way forward is just more polished styling of the stack so that the repetition doesn't look as bad. I don't know any other debugger that works like this.

Personally, I'd rather stay with convention for now and work with Helen to solve UX problem like this.

jasonLaster:
Cool idea. I think Xcode does something similar. 

I agree, we should consider the UX of scanning the stack and selecting a frame.

clarkbw:
Yeah, I don't think I have the right UX here but I think I'm inching closer.  I'll definitely defer to what @helenvholmes thinks is right.

I noticed this as I was playing with the storyboard to make instructions for the CONTRIBUTING doc.  I'll close this PR for now as its not pressing but one of those polish items I think we can improve on once we've discussed a better path forward.

I don't think that you can use colors or typography to simplify the repetition here.  But I don't want to create another issue where you're unsure what file the frame is coming from.  Might be a better way to lay it out than I did.
screen shot 2016-05-03 at 7 05 55 pm

What might be helpful to move forward is a better sense of the possible outcomes.  One thing I noticed in the storybook was that one of the locations was empty and I wasn't sure what that meant.  Perhaps we should have more variations in the storybook fixtures.

jlongster:
We definitely should but we're not in the polish stage yet. I want to implement the basic features first and then we'll go through each component and think through it. There are all kinds of things we need to think about with sources: eval sources with no filename, blackboxed sources, etc. Each component that deals with sources (most of them) will need to be thought through in regards to all those states. We'll get to it, but let's focus on basic features first.

Issue #113, opened 2016-04-28
Refreshing the debugger fails to detatch 


jasonLaster:
Steps:
1. set a breakpoint in the debugger
2. do something in the page to trigger the breakpoint
3. refresh the debugger

What happens:
The debugger does not detatch. 

What should happen:
Debugger should call detatch on unload or something to that effect.

PR #141, opened 2016-05-04
clean up connection code and reset UI on navigation

jlongster:


public/js/client.js
```diff
@@ -17,53 +26,48 @@ function connectToClient(onConnect) {
...
-function debugTab({ tabActor, newSource, paused, resumed,
-                    selectTab, selectSource, loadSources }) {
-  function listenToClient() {
-    let deferred = promise.defer();
     let client = getThreadClient();
 
     client.addListener("paused", (_, packet) => {
-      paused(packet);
+      actions.paused(packet);
       if (packet.why.type != "interrupted") {
```
jasonLaster: does it make sense for `selectSource` to move to the `paused` action? I say that because `paused` already dispatches a loadFrames action.

public/js/client.js
```diff
@@ -17,53 +26,48 @@ function connectToClient(onConnect) {
...
 
-    client.addListener("resumed", (_, packet) => resumed(packet));
-    client.addListener("newSource", (_, packet) => newSource(packet.source));
-
-    return deferred.resolve();
-  }
+    client.addListener("resumed", (_, packet) => actions.resumed(packet));
+    client.addListener("newSource", (_, packet) => {
+      actions.newSource(packet.source);
+    });
```
jasonLaster: nit - can this go on one line

public/js/components/Tabs.js
```diff
@@ -10,12 +10,9 @@ const { debugTab } = require("../client");
...
 
-function Tabs({ tabs, newSource, paused, resumed,
-  selectTab, selectSource, loadSources }) {
-  function onClickTab(e) {
-    const tabActor = e.currentTarget.dataset.actorId;
-    debugTab({ tabActor, newSource, paused, resumed,
-               selectTab, selectSource, loadSources });
+function Tabs({ tabs, actions: boundActions }) {
+  function onSelect(tab) {
+    debugTab(tab.toJS(), boundActions);
```
jasonLaster: YES! I spent atleast an hour trying to think of a way to cleanup the actions list and it didn't occur to me you could pass them all in :) that makes me really happy to see.

jlongster:
It's been annoying me that it doesn't properly track sources if you navigate to different pages. Now it will clear the UI on navigation (there are a few cleanups after this, like managing the editor state).

I tweaked the code in `client.js` as well. Because we connect a tab/thread outside of the components (if one is already selected), I moved the actual call to `connectThread` (which used to be `connectToTab`) outside of the tabs actions. This makes it easier to follow the `debugTab` function, which just calls `connectThread` directly. The tabs reducer is purely for maintaining the UI of the tabs.

This made it easier to get access to the tab form and get the target for it in `debugTab`, which I need to install navigation handlers.

public/js/components/Tabs.js
jlongster: Ah good, I wasn't sure if you liked it the other way for clarity, but I think this works well!

jlongster:
I meant to add some comments in `debugTab` also, which I will do before landing.

public/js/actions/index.js
```diff
@@ -1,10 +1,21 @@
...
 const pause = require("./pause");
 
+function willNavigate() {
+  return { type: constants.NAVIGATE };
+}
+
+function navigate() {
+  return sources.loadSources();
+}
+
```
jasonLaster: nit: a navigate const would make it more consistent.

`const navigate = { willNavigate, navigate }`

public/js/client.js
jlongster: yep, I think that's a great idea!

public/js/client.js
```diff
@@ -17,53 +26,48 @@ function connectToClient(onConnect) {
...
   });
-
-  return deferred.promise;
 }
 
-function getThreadClient() {
-  return currentThreadClient;
-}
+function debugTab(tab, actions) {
+  return Task.spawn(function* () {
```
jasonLaster: nice

jasonLaster:
A couple minor nits, but looks great. Does this fix the refresh bug? https://github.com/jlongster/debugger.html/issues/113

public/js/components/Tabs.js
jasonLaster: this is good. Tabs specifies it's dependencies

public/js/client.js
jlongster: unfortunately the line is past the allowed length: 84 chars

public/js/actions/index.js
jlongster: I'm confused, for what exactly? And that name will conflict with the `navigate` function

jlongster:
> A couple minor nits, but looks great. Does this fix the refresh bug? #113

Unfortunately no. I think that's going to be some interaction with the connection closing & the proxy. I can help look at that if you want.

public/js/client.js
jasonLaster: lol - was worried about that

public/js/actions/index.js
jasonLaster: was hoping to cleanup module.exports. not a big deal

PR #142, opened 2016-05-04
Add a firefox driver

jasonLaster:
The firefox driver creates a profile and opens firefox developer edition
with that profile.

In theory, it should also accept a remote connection from the debugger.

[error logs](https://gist.github.com/jasonLaster/d6e24a6c5be61e26e541645e326f5aaf)

@jlongster would you mind giving this a go locally and see if i'm missing something obvious?

### DIFF
--- a/bin/firefox-driver.js
+++ b/bin/firefox-driver.js
@@ -1,0 +1,71 @@
+"use strict";
+
+/**
+  NOTE: you'll need to be running selenium server beforehand
+  java -jar ~/bin/selenium-server-standalone-2.53.0.jar
+*/
+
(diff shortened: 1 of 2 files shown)


jasonLaster:
Just updated the driver to use `selenium-webdriver`, which has the official bindings and access to low level primitives such as `firefox.Binary` and `firefox.Profile`. This is still a WIP. Currently, the driver demonstrates that it can open todomvc and add a couple todos. The next step will be to connect the debugger integration tests I've written with Cypress to the driver so that the debugger tests can drive the debuggee.

PR #143, opened 2016-05-05
Integration Tests (first pass)

jasonLaster:
This adds basic support for end-to-end integration tests.

How does it work:

1. The debugger client is opened in an iframe inside of a mocha web runner.
The tests are actually very similar to our unit tests that run in a web
context. The debugger is driven by a tool called cypress that performs
all of the actions like clicking, typing, etc...

2. The debuggee is open in firefox with selenium web driver. Commands
are sent to the debuggee over HTTP with a simple express server.

Next Steps:

1. Add a headless runner (cypress-cli) should not be too hard.
2. Add to circle ci. Also should not be too hard
3. Write tests for debugger commands (resume, step in, etc). The fun
part

Trying it out:

1. Run selenium server beforehand java -jar ~/bin/selenium-server-standalone-2.53.0.jar
2. Run `node bin/firefox-driver`
3. Install cypress `npm install -g cypress-cli`.
4. Install cypress app `cypress-cli install`
5. Start cypress server from cypress app in menubar

Going forward many of these steps will be automated away

1. the selenium server and driver will be wrapped by the test start
command
2. The cypress desktop app will be replaced by the cli tool


public/js/components/Frames.js
```diff
@@ -27,18 +27,24 @@ function renderFrameLocation(frame) {
   return div({ className: "location" }, basename(frame.where.source.url));
 }
 
+function renderFrame(frame, selectedFrame) {
+
```
jlongster: nit: extra line here?

public/js/test/cypress/commands.js
```diff
@@ -0,0 +1,39 @@
+// ***********************************************
```
jlongster: These files are all commented out, are they templates or going to be fleshed out later?

jlongster:
r+ once circleci passes. It's not clear how to actually run this, can you add either docs or a package.json script? Also, eventually I think we should share the CSS selectors used instead of embedding them in the test.

public/js/test/cypress/commands.js
jasonLaster: templates. not sure if we'll use these

jasonLaster:
@jlongster I'll write docs / steps soon

jasonLaster:
merged

PR #144, opened 2016-05-06
Add additional stories

jasonLaster:
Adds more storybook stories:

(WIP) still in progress

Cleanup fixtures
+ Fixtures are renamed and moved
+ added fixtures/index.js

In the examples where fixtures have app state, i'm leaning towards
capturing the full app state as opposed to part of it. There are a
couple reasons:

1. other tests/stories end up needing some state
2. looking at a fixture it is hard to tell what part of the store it was
3. it's easy to capture app state: `copy(store.getState())`

Cleanup Stories / Add more stories
+ adds some simple json fixture queries for data


public/js/components/stories/Frames.js
```diff
@@ -28,8 +34,8 @@ function renderContainer(Component, frames) {
       margin: "auto",
       paddingTop: "100px"
     }},
-    dom.div({ style: { border: "1px solid #ccc", padding: "20px" }},
-            createElement(Provider, { store },
-                          createElement(Component)))
+    createElement(Provider, { store },
+      dom.div({className: "accordion"},
+        dom.div({className: "_content"}, createElement(Component))))
```
jasonLaster: this was needed for styles to apply. It *should* be a code smell that an element needs additional surrounding context to style correctly. I'll file a bug to fix this soon.

jasonLaster:
@jlongster r?

I was initially going to do more work cleaning up the stories, but i think it's fine for now

public/js/components/stories/Frames.js
```diff
@@ -28,8 +34,8 @@ function renderContainer(Component, frames) {
       margin: "auto",
       paddingTop: "100px"
     }},
-    dom.div({ style: { border: "1px solid #ccc", padding: "20px" }},
-            createElement(Provider, { store },
-                          createElement(Component)))
+    createElement(Provider, { store },
+      dom.div({ className: "accordion" },
```
jlongster: Why are you wrapping this with an accordion class? If these are to be as docs that's sort of confusing.

jlongster:
r+ I like consolidating the fixtures, good idea. I think you answered my one question in an outdated commit actually. Feel free to merge.

PR #148, opened 2016-05-06
Add clientLogging

jasonLaster:


jasonLaster:
tempted to be more literal with `clientLogging` and call it dumpn.wantLogging or something like that, but part of me thinks that new contributors don't care about that... torn.

jlongster:
Works for me :) we'll probably work on better logging at some point and we can always change it later

PR #151, opened 2016-05-10
add proxy to dev-server

jasonLaster:
This PR makes the proxy a dependency of the server. The rationale is that both the  firefox-proxy and server are useful for the local development environment, but not necessarily useful when the debugger is in the FF toolbox. 

Going forward, I'd like to refactor the proxy so that it has a better api `proxy.start({ clientPort: 9000, browserPort: 6080 })`.

Also, I've noticed that if integration tests are running in the background and you happen to be testing manually as well, there can be miscommunications between the two debuggers and two debuggees. The solution, would be to have the testing environment communicate over different ports 9001, 6081. We could do that by adding the clienPort to the config.js and maybe having a testing config as well, although from experience, the testing config should default to production.

### DIFF
--- a/bin/server.js
+++ b/bin/server.js
@@ -7,6 +7,8 @@ const express = require("express");
 const projectConfig = require("../webpack.config");
 const webpackDevMiddleware = require("webpack-dev-middleware");
 
+require("ff-devtools-libs/bin/firefox-proxy");
+
 const config = Object.assign({}, projectConfig, {
   entry: path.join(__dirname, "../public/js/main.js"),
 });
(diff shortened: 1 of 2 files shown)


jlongster:
I kind of like that they are separate.

> The rationale is that both the firefox-proxy and server are useful for the local development environment, but not necessarily useful when the debugger is in the FF toolbox.

How does does that relate to this PR? Both are things for development, neither which are needed when we integrate with FF. Whether they are separate or not doesn't really matter for that.

I like it being separate so other projects can easily see those two pieces, and they'll likely re-use firefox-proxy but not our server.

But I don't really care that might right now, we can always separate it out later if we think that's clearer.

jlongster:
Going ahead and merging because I don't like leaving PRs open, and this works for me.

PR #152, opened 2016-05-11
snippets

jasonLaster:
first pass at adding support for snippets

Paired w/ @gabrielluong


jlongster:
Sweet, looks great! Some linting failures though.

jasonLaster:
I'll fix on monday (About to head to the plane). 

I'd like to wait on merging snippets until the integration tests land. Snippets change how breakpoints are shown, and will cause the integration tests to fail.

PR #159, opened 2016-05-13
fix React warnings about using maps and using keys when listing break…

### LINKED ISSUE #150: Fix: Warning: Using Maps as children
We've had this warning for the past couple weeks. Should be easy to find the component that's using a map.

```
VM3967 bundle.js:2477 Warning: Using Maps as children is not yet fully supported. It is an experimental feature that might be removed. Convert it to a sequence / iterable of keyed ReactElements instead.
```

jlongster:
…points in the editor (fixes #150)

### DIFF
--- a/public/js/components/Editor.js
+++ b/public/js/components/Editor.js
@@ -1,20 +1,19 @@
 "use strict";
 
 const React = require("react");
-const Isvg = React.createFactory(require("react-inlinesvg"));
-const { DOM: dom, PropTypes } = React;
-const ImPropTypes = require("react-immutable-proptypes");
-
 const ReactDOM = require("react-dom");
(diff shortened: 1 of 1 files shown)


public/js/components/Editor.js
```diff
@@ -1,28 +1,26 @@
...
 const {
   getSourceText, getPause, getBreakpointsForSource,
-  getSelectedSource, getSelectedSourceOpts
+  getSelectedSource, getSelectedSourceOpts,
+  makeLocationId
 } = require("../selectors");
-
-const { bindActionCreators } = require("redux");
-const { connect } = require("react-redux");
-
```
jlongster: All of this is just reordering the imports so they are grouped better.

jasonLaster:
looks good. just CI

PR #161, opened 2016-05-13
Integration Tests part 2

jasonLaster:
Adds Cypress to CI.

Also updates the integration tests for the new sources tree.

Lots of small improvements:
+ driver: only open one firefox at a time
+ driver: handle pause events
+ tests: use a babel endpoint for CI and headless runs. Unfortunately, cypress runs headlessly in an electron version that does not support es6. They'll be updating that soon.


bin/development-server.js
```diff
@@ -26,10 +38,22 @@ app.use(webpackDevMiddleware(compiler, {
...
+}));
+
 app.get("/", function(req, res) {
   res.sendFile(path.join(__dirname, "../index.html"));
 });
 
+app.get("/babel", function(req, res) {
+  res.sendFile(path.join(__dirname, "../babel.html"));
+});
+
```
jasonLaster: The tests visit `8000/babel` so that headlessly the es6 code is traspiled to es5. Hopefully this can go away soon.

bin/firefox-driver.js
```diff
@@ -1,38 +1,14 @@
...
-        env: env
-      });
-    });
-  }
-
-  var binary = new firefox.Binary('/Applications/Firefox.app/Contents/MacOS/firefox-bin');
-  binary.addArguments('--start-debugger-server 6080')
-  binary.launch = launch;
+  var binary = new firefox.Binary();
+  binary.addArguments('--start-debugger-server', '6080')
```
jasonLaster: I'm able to kill the launch logic because arguments are passed in key and value as opposed to just option.

bin/firefox-driver.js
```diff
@@ -56,24 +32,4 @@ function startDriver() {
...
-
-  app.listen(9002, function () {
-    console.log('Debuggee Server listening on 9002!');
-  });
-
-  return app;
-}
-
-const driver = startDriver();
-const app = startExpressServer();
```
jasonLaster: moved the test server to it's own module as it was plenty complicated

bin/test-server.js
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  try {
+    var r  = eval(command);
+  } catch (e) {
+    console.log("error", e.stack);
+  }
+
+  var timer = setTimeout(cb, timeout);
+  r.then(cb)
```
jasonLaster: I'll add documentation soon, but the setTimeout handles cases where the eval times out because we're paused.

public/js/test/integration/todomvc.js
```diff
@@ -50,21 +60,37 @@ function debuggee(callback) {
...
   beforeEach(function() {
     debuggee(() => {
-      driver.get("http://todomvc.com/examples/backbone/");
+      driver.get("http://localhost:9002/todomvc/examples/backbone/");
     });
-    cy.visit("http://localhost:8000/#tab=tab1");
+
+    cy.visit("http://localhost:8000/babel");
+
+    cy.get(".tab").first().click();
```
jasonLaster: clicking the tab handles the case where you refresh the debugger before firefox is open.

public/js/test/integration/todomvc.js
```diff
@@ -50,21 +60,37 @@ function debuggee(callback) {
...
+  });
+
+  afterEach(function() {
+    cy.visit("http://localhost:8000/babel");
   });
 
   it("Adding Breakpoints in one source", function() {
-    goToSource("todo-view");
+    goToSource("examples/backbone/js/views/todo-view");
```
jasonLaster: we need the full path for the tree

jlongster:
Awesome! Gotta run for now put I'll review this over the weekend or first thing Monday

bin/test-server.js
jlongster: Not sure I understand this; the `eval` happens right above this so the timer isn't installed before the `eval` finishes executing. But this is all magic to me.

package.json
```diff
@@ -63,9 +64,11 @@
     "karma-mocha": "^0.2.2",
     "karma-webpack": "^1.7.0",
     "mocha-circleci-reporter": "0.0.1",
+    "node-static": "^0.7.7",
     "npm": "^3.8.8",
     "rimraf": "^2.5.2",
     "selenium-webdriver": "^2.53.2",
-    "style-loader": "^0.13.1"
+    "style-loader": "^0.13.1",
+    "todomvc": "^0.1.1"
```
jlongster: Why do we depend on this?

jlongster:
This looks great. It feels like a lot of moving parts, so I'd say lets try not add any more tooling until we get a good feeling about all of this. But it will be good to try it out so we can show the team this workflow in June.

r+

bin/test-server.js
jasonLaster: yeah - this code desperately needs some comments.

The gist is that the `eval` runs the driver commands that are added to a managed promise schedule, basically a set of coordinated operations that will happen in the browser (type, mouse, click). The timer is useful for resuming when those operations have paused due to a breakpoint being hit.

package.json
jasonLaster: We're pulling it in so that the tests in CI can visit todomvc. I can replace this with a set of static built files, pretty easily. At some point, we might not want to test todomvc, but it seems like the easiest set of tests to write at the moment to touch on many different interactions (going between pages, breakpoints etc)... Also, there is some merit to having your test subject be more realistic.

jasonLaster:
> re: adding more tooling. 

I believe we're in a good place in terms of testing. I expect the integration testing to stabilize like the unit tests as we use them in the coming weeks.

PR #164, opened 2016-05-16
Local development improvements

jasonLaster:
Depends on integration tests. I'll rebase after that lands.

+ add a start-firefox command. uses the selenium web driver to start
firefox with a good profile.
+ adds a "not connected" pane for telling people to connect firefox.
+ opens a browser tab for the debugger when the local developent server starts.
+ cleaned up the readme getting started. Now just 3 commands!
+ added testing / linting commands to the readme.


bin/test-server.js
```diff
@@ -0,0 +1,89 @@
...
+  app.use(bodyParser.json());
+
+  app.post('/start', debuggeeStart);
+  app.post('/stop', debuggeeStop);
+  app.post('/command', debuggeeCommand);
+
+  // Serves todomvc from node_modules.
+  // This is a test to see if going forward, we want to pull in test
+  // examples from the community.
+  app.use(express.static("node_modules"));
```
jlongster: We probably don't want to run tests directly against something pulled in from npm. It just seems too likely to change. I'd rather run tests against static files directly commited in our repos.

jlongster:
I glanced over this and it looks good. Commented on one thing. Also, I think it'd be better to show Firefox in the screenshot instead of Chrome. Otherwise, r+

bin/test-server.js
jasonLaster: That's fair, we can either version lock the tests (todomvc) or pull down static files. I'll create a cleanup item for that.

PR #165, opened 2016-05-16
Add displayName to all components.

TatumCreative:
I'm reviewing the new tree sources, and the display names were missing on several components. I added them here so that the React add-on displays better information.

screen shot 2016-05-16 at 1 43 32 pm


jlongster:
Very good point! We should turn this on in the linter. (Hm, just looked, looks like it should be a warning, not sure why that's not applying)

PR #169, opened 2016-05-16
Add chrome client

jasonLaster:
WIP

+ get chrome tabs
+ bootstrap devtools

Tomorrow I'll see if i can reconcile the browser tabs state and maybe explore sources/scripts as well. I'll likely have some diagrams and more questions...

screen shot 2016-05-16 at 4 13 11 pm

### DIFF
--- a/bin/development-server.js
+++ b/bin/development-server.js
@@ -4,11 +4,26 @@
 const path = require("path");
 const webpack = require("webpack");
 const express = require("express");
-const projectConfig = require("../webpack.config");
 const webpackDevMiddleware = require("webpack-dev-middleware");
 const open = require("openurl").open;
+const http = require("http");
+
+const projectConfig = require("../webpack.config");
+const getValue = require("../public/js/configs/feature").getValue;
 
 require("ff-devtools-libs/bin/firefox-proxy");
+
+function httpGet(url, onResponse) {
+  return http.get(url, (response) => {
+    let body = '';
+    response.on('data', function(d) {
+      body += d;
+    });
+    response.on('end', function() {
+      onResponse(body);
+    });
+  });
+}
 
 const config = Object.assign({}, projectConfig, {
   entry: path.join(__dirname, "../public/js/main.js"),
@@ -54,6 +69,20 @@ app.get("/", function(req, res) {
 app.get("/babel", function(req, res) {
   res.sendFile(path.join(__dirname, "../babel.html"));
 });
+
+app.get("/chrome-tabs", function(req, res) {
+  const webSocketPort = getValue("chrome.webSocketPort");
+  const url = `http://localhost:${webSocketPort}/json/list`;
+
+  const tabRequest = httpGet(url, body => res.json(JSON.parse(body)));
+
+  tabRequest.on('error', function (err) {
+    if (err.code == "ECONNREFUSED") {
+      console.log("Failed to connect to chrome");
+    }
+  });
+
+})
 
 app.listen(8000, "localhost", function(err, result) {
   if (err) {
--- a/public/js/clients/firefox.js
+++ b/public/js/clients/firefox.js
@@ -16,13 +16,26 @@ function getTabTarget(tab) {
   return TargetFactory.forRemoteTab(options);
 }
 
+function presentTabs(tabs) {
+  return tabs.map(tab => {
+    return {
+      title: tab.title,
+      url: tab.url,
+      id: tab.actor,
+      firefox: tab
+    };
+  });
+}
+
 function connectClient(onConnect) {
   const socket = new WebSocket("ws://localhost:9000");
   const transport = new DebuggerTransport(socket);
   currentClient = new DebuggerClient(transport);
 
   currentClient.connect().then(() => {
-    return currentClient.listTabs().then(onConnect);
+    return currentClient.listTabs().then(response => {
+      onConnect(presentTabs(response.tabs));
+    });
   }).catch(err => console.log(err));
 }
 
--- a/public/js/components/Tabs.js
+++ b/public/js/components/Tabs.js
@@ -5,28 +5,46 @@ const { connect } = require("react-redux");
 const actions = require("../actions");
 const { bindActionCreators } = require("redux");
 const { getTabs } = require("../selectors");
-const { debugTab } = require("../client");
+const { debugTab } = require("../clients/firefox");
 
 require("./Tabs.css");
 const dom = React.DOM;
 
-function Tabs({ tabs, actions: boundActions }) {
-  function onSelect(tab) {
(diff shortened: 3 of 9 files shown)


jasonLaster:
Updated PR:

I limited the scope of the PR to "just show chrome tabs". 

screen shot 2016-05-17 at 11 59 01 am

Couple things that are worth noting:
+ I moved client.js to a clients/firefox.js so that there could be some parity. 
+ Adapters, for the time being I am just massaging the client response for tabs in `presentTabs`. I'm not sure if i like that pattern yet.
+ Unify app state, for the time being I am lifting shared fields and using a browser field for the data. Here's a gist of tab state [gist](https://gist.github.com/jasonLaster/b1985415a1425b3ca09e41e3d7efccd4). This is the path of least resistence and i'm reserving judgement until I have more data points and we can think it through

bin/development-server.js
```diff
@@ -55,6 +56,25 @@ app.get("/babel", function(req, res) {
   res.sendFile(path.join(__dirname, "../babel.html"));
 });
 
+app.get("/chrome-tabs", function(req, res) {
+  const tabRequest = http.get('http://localhost:9222/json/list', (response) => {
```
jasonLaster: the development server makes the request for tabs from chrome because this endpoint does not support jsonp.

This approach raises an interesting question as to whether we ship the a standalone app with a local server. At the moment, I want to reserve judgement and this was a path forward. I think we could make a chromium contribution for jsonp fairly easily too.

bin/development-server.js
```diff
@@ -55,6 +68,19 @@ app.get("/babel", function(req, res) {
   res.sendFile(path.join(__dirname, "../babel.html"));
 });
 
+app.get("/chrome-tabs", function(req, res) {
+  const tabRequest = httpGet('http://localhost:9222/json/list', body => {
```
jasonLaster: the development server makes the request for tabs from chrome because this endpoint does not support jsonp.

This approach raises an interesting question as to whether we ship the a standalone app with a local server. At the moment, I want to reserve judgement and this was a path forward. I think we could make a chromium contribution for jsonp fairly easily too.

public/js/main.js
```diff
@@ -68,7 +74,7 @@ function hasSelectedTab() {
  */
 function getSelectedTab(tabs) {
   const childId = window.location.hash.split("=")[1];
-  return tabs.find(tab => tab.get("actor").includes(childId));
+  return tabs.find(tab => tab.get("id").includes(childId));
```
jasonLaster: There were two functional changes to how Tabs work:

1) the top level tab state uses `id` instead of `actor`
2) when a tab is selected either in the UI or URL in main, tab is determined by checking the firefox and chrome field. I think there are some obvious drawbacks here, but am reserving judgement until I see a couple other usecases

jlongster:
Do you want me to r? this or wait for more changes?

jasonLaster:
oh sorry, r?

public/js/main.js
```diff
@@ -36,19 +37,24 @@ const actions = bindActionCreators(require("./actions"), store.dispatch);
...
+    const tab = selectedTab.get("firefox") || selectedTab.get("chrome");
+    debugTab(tab, actions).then(renderToolbox);
   } else {
     renderToolbox();
   }
 });
 
+chromeTabs(response => {
+  actions.newTabs(response);
+});
```
jasonLaster: this should be flagged, so it's only on locally, and not when it's embeded.

public/js/components/Tabs.js
```diff
@@ -5,28 +5,46 @@ const { connect } = require("react-redux");
...
+  );
+}
+
+function Tabs({ tabs, actions: boundActions }) {
+  const firefoxTabs = tabs.valueSeq().filter(tab => tab.get("firefox"));
+  const chromeTabs = tabs.valueSeq().filter(tab => tab.get("chrome"));
+
+  return dom.div({ className: "tabs" },
+    renderTabs("Firefox Tabs", firefoxTabs, boundActions),
+    renderTabs("Chrome Tabs", chromeTabs, boundActions)
```
jlongster: I thought we decided to avoid this for now? I'd rather create a query string that specific the target we should debug. It could take either a `chromePort` or `firefoxPort` option and depending on that it would connect to the specific target. It just feels easier to think about what I'm debugging, because I'm debugging a specific "tab" but is it a Firefox or a Chrome tab? We aren't building a toolbox/IDE right now, so I'd prefer to make it clear by using URL params to designate the debuggee.

public/js/main.js
```diff
@@ -36,19 +37,27 @@ const actions = bindActionCreators(require("./actions"), store.dispatch);
...
-  actions.newTabs(response.tabs);
+connectClient(tabs => {
+  actions.newTabs(tabs);
 
   // if there's a pre-selected tab, connect to it and load the sources.
   // otherwise, just show the toolbox.
   if (hasSelectedTab()) {
     const selectedTab = getSelectedTab(store.getState().tabs.get("tabs"));
-    debugTab(selectedTab.toJS(), actions).then(renderToolbox);
+    const tab = selectedTab.get("firefox") || selectedTab.get("chrome");
```
jlongster: Same as my comment above, I don't love this ambiguity because I'll find myself with the same page open in Firefox or Chrome (while testing) and it's not clear which browser I'm testing. I'd rather encode the target browser in the URL params.

public/js/main.js
jlongster: Switching from `actor` to `id` is a good idea, we'll probably do that for all objects

public/js/clients/firefox.js
```diff
@@ -16,13 +16,26 @@ function getTabTarget(tab) {
   return TargetFactory.forRemoteTab(options);
 }
 
+function presentTabs(tabs) {
```
jlongster: `present` is a little ambiguous to me, it feels pretty close to `render` but this isn't actually presenting anything, it's just formatting. Not sure if there's a better word? If you want to keep it, that's fine!

jlongster:
Looks generally good, just one comment about how we handle targets. Thanks for doing this! 🐰

public/js/main.js
jasonLaster: I'll fix this when i tackle debugging a tab for chrome. Want to wait because there might be more there.

public/js/clients/firefox.js
jasonLaster: agreed. i think formatting is nicer, but I think there's a better word too. something closer to what an adapter would do. I'm thinking of the layer between sql and orm models, or orm models and UI views... I'll be doing lots of this work soon so will punt for the time being as well. not ignoring this though :P

jasonLaster:
merged. the test failure was a linting thing which was fixed.

PR #173, opened 2016-05-17
WIP Chrome Client (2)

jasonLaster:
WIP

Ignore the first commit that adds the chrome tabs. That's in a separate PR.

Sharing an early version of some of the chrome client work i'm prototyping. It does two things here that are both interesting:

1) sets up the websocket connection with chrome and sends some bootstrap messages over the wire

> this is the debug tab commit

2) adds the chrome inspector backend

> You can think of this piece as chromes version of our client, fronts, and protocol api. It's split into two pieces:
  1) InsepctorBackend - holds onto what they call agents (our fronts)
  2) protocol.json - spec of the protocol, which is used to build the agents (fronts) ([gist](https://gist.github.com/jasonLaster/f1609e7764164e829caec7814ebff533) of the dynamically built fronts)

### DIFF
--- a/public/js/clients/chrome.js
+++ b/public/js/clients/chrome.js
@@ -1,5 +1,18 @@
 "use strict";
 
+const {
+  WebSocketConnection,
+  generateCommands
+} = require("./chrome/api");
+
+const protocol = require("./chrome/protocol.json");
+
+let connectionAgents;
+
+function getAgent(name) {
+  return connectionAgents[name];
+}
+
 function presentTabs(tabs) {
   const blacklist = ["New Tab", "Inspectable pages"];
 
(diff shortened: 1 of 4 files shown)


public/js/clients/chrome/InspectorBackend.js
```diff
@@ -0,0 +1,704 @@
+/**
```
jlongster: Where did these files in the `chrome` folder come from? Might be worth documenting (unless they already are and I missed that)

jlongster: EDIT: I see the another file has a "comes from" message, but it's just a relative path that doesn't link to anywhere. I haven't finished this PR yet, but it might be nice to have a `.md` in here describing exactly where to get these files.

jlongster:
👍 Looks good, r+

jasonLaster:
@jlongster mind giving this another read through?

Summary:

First step towards debugging a chrome tab.

+ Pulls in the chrome api layer, which is a lot like our
client/shared/main.js
+ adds a debugTab to the chrome client for debugging a tab. It knows how
to start the inspector backend, connect the websocket, and enable the
debugger and runtime.

public/js/clients/chrome.js
```diff
@@ -28,6 +42,58 @@ function chromeTabs(callback) {
   });
 }
 
+function onConnection(connection) {
+  const ws = connection._socket;
+  connectionAgents = connection._agents;
+  ws.onopen =function() {
```
jlongster: Nit: space after equals

public/js/clients/chrome.js
```diff
@@ -28,6 +42,58 @@ function chromeTabs(callback) {
...
+  // we will want to build the bootstrap commands in advance require them in.
+
+  // NOTE: toTitleCase is temporarily monkey patched onto String.prototype
+  // to conform to what generateCommands expects. I
+  // If String.prototype.toTitleCase is defined somwhere else,
+  // it will be replaced.
+  String.prototype.toTitleCase = toTitleCase;
+  let code = generateCommands(protocol);
+  delete String.prototype.toTitleCase;
+  eval(code);
```
jlongster: Can you file an issue to figure out a way to avoid both `delete` and `eval`? It does seem like something we should do sooner rather than later. 

I wonder if shadowing `String` would work?

```
const _String = String;
function String() { return _String.apply(this, arguments); }
String.prototype = Object.create(_String);
String.prototype.toTitleCase = ...
```

Not sure if that's totally correct or would even work...

jlongster:
I think you just force pushed here but my two comments still stand I think. Most of this is just pulling in chrome's library files, so looks good. r+

PR #177, opened 2016-05-19
Remove chrome's eval/monkeypatch

### LINKED ISSUE #176: Remove chrome client eval
Chrome's client dynamically builds a bootstrap script, which it eval's on startup.

We should pre-build the bootstrap script and require it in the module.

jasonLaster:
Fixes #176

This patch adds a new `build-bootstrap` node file, which pre-builds the bootstrap script, which is run when we're setting up the chrome client.

@jlongster thanks for pushing me to fix this early on 👍


jlongster:
Awesome!

> which is run when we're setting up the chrome client.

I don't see where that happens? I don't see where that file is run anywhere, or did I miss it?

public/js/clients/chrome.js
```diff
@@ -61,23 +61,7 @@ function onConnection(connection) {
...
-  // NOTE: toTitleCase is temporarily monkey patched onto String.prototype
-  // to conform to what generateCommands expects. I
-  // If String.prototype.toTitleCase is defined somwhere else,
-  // it will be replaced.
-  String.prototype.toTitleCase = toTitleCase; /* eslint no-extend-native: 0 */
-  let code = generateCommands(protocol);
-  delete String.prototype.toTitleCase;
-  eval(code); /* eslint no-eval: 0 */
-
+  bootstrap(InspectorBackend);
```
jasonLaster: it's run here 🏃

jasonLaster:
merged as I believe you r+ it.

let me know if you have any questions.

public/js/clients/chrome.js
jlongster: This comes from the file `bootstrap.js` though, which is built by `public/js/clients/chrome/build-bootstrap.js`. Are you supposed to run that file manually to update `bootstrap.js`, and commit the changes?

PR #178, opened 2016-05-19
Implement a basic tree-based object inspector for scopes

jlongster:
**WIP: don't merge or formally review**. This is a proof-of-concept, but feel free to comment if you want.

This makes scopes inspectable: you can recursively expand variables in scope that are objects and dig into their values.

screen shot 2016-05-19 at 2 47 06 pm

I'm still learning the best way to do this. I have a much clearer picture now after this first pass. I think the next step is do have a more formal tree structure, and incrementally update it as we expand and populate objects. The hard thing here is that we lazily populate objects: when you expand `Window` we don't readily have those properties, we have to fetch them for the server. So we need to make the tree async.

Now I realize that the way the sources tree works should be able to handle async: the only difference is that `newSource` notifications are unsolicited, but our population requests are explicit from us. We just need a way to map the results from the server back into in the tree. This PR so far dynamically creates the tree by checking the reducer state, but this has a few drawbacks. Because we lazily create the tree, we need to sort it every time. We could sort it in the reducer, but I don't feel like view behavior should go in there (i.e. what if the user turns sorting off? don't want to change state, just update the view). Also, it's easy to create new instances of nodes each time the tree is rerendered which causes problems for tracking various things (this PR has bugs where it doesn't track the focused item correctly, etc).

Anyway, just a heads up. I'm going to stop working on this for now and will take another pass next week.


jasonLaster:
Have you considered doing memoization on the query / selector level? This seems slightly better than in the reducer. 

I agree that the reducer makes less sense. One more use case to consider is the filter use case where you want to filter the variables and just see some that match.

jasonLaster:
Not really a commit comment, but out of curiosity do you think that storybook could help you develop this faster?

The pros:
+ you don't need to manually pause
+ you can work off of fixture data for different cases (initial load), (some objects partially fetched)

Cons:
+ we don't have a good way for stubbing the client here. Might be nice to prototype something like that...

jlongster:
Never responded, but re: memoization, yeah we should probably do it on the component like the sources tree. I will re-work this post-London.

I will also work on Storybook integration. The majority of this way exploring the data shape and playing around with it, and it was easier to do it straight in the page because I could quickly load all kinds of different JS sources and paused data

public/js/clients/firefox.js
```diff
@@ -94,6 +96,11 @@ let APIClient = {
 
   navigate(url) {
     return currentTabTarget.activeTab.navigateTo(url);
+  },
+
+  getProperties(grip) {
```
jasonLaster: I think `getObjectProperties` might be more clear, but not a big deal

public/js/components/Scopes.js
```diff
@@ -3,110 +3,169 @@
...
+      }
+    } else if (type === "object") {
+      scopes.push([
+        scope.getIn(["object", "class"]),
+        fromJS({
+          value: scope.get("object")
+        })
+      ]);
+    }
+  } while (scope = scope.get("parent")); // eslint-disable-line no-cond-assign
```
jasonLaster: this is messy, but I'm sure we can clean it up

public/js/components/Scopes.js
```diff
@@ -3,110 +3,169 @@
...
+    const type = scope.get("type");
+
+    if (type === "function" || type === "block") {
+      const bindings = scope.get("bindings");
+      const vars = bindings
+        .get("arguments").map(arg => arg.entrySeq().get(0))
+        .concat(bindings.get("variables").entrySeq())
+        .filter(binding => (!binding[1].hasIn(["value", "missingArguments"]) &&
+                            !binding[1].hasIn(["value", "optimizedOut"])))
+        .sort((a, b) => a[0].localeCompare(b[0]))
```
jasonLaster: what is this doing?

public/js/components/Scopes.js
```diff
@@ -3,110 +3,169 @@
...
+    let val = "";
+    if (obj.has("value") && !obj.hasIn(["value", "type"])) {
+      val = ": " + obj.get("value");
+    }
+
+    return dom.div(
+      { className: classnames("node", { focused }),
+        style: { marginLeft: depth * 15 }},
+      Arrow({
+        className: classnames(
```
jasonLaster: 👍 classnames

public/js/clients/firefox.js
jlongster: True, I'll change it, but we'll probably want to rethink it again later because there are a few other methods for querying objects

jasonLaster:
I skimmed the Scopes.js as it's self contained and I assume will be re-worked with Reps. 

One thought on the selector `getScopes`. It might be nice to extract that into `js/selectors/scopes`. That way, it would be easy to write unit tests for it and create small helper functions so that it is more manageable.

The actions/reducer/client looked good and that's what I mostly cared about.

Great job. Will be a great win for london

public/js/components/Scopes.js
jlongster: Yeah, it's demo quality. We need to build the tree in a very different way so this will be totally re-worked when we have time to work on the object inspector again.

public/js/components/Scopes.js
jlongster: Gets all the variables in scope, filters out ones that are optimized out, and sorts them.

jlongster:
Thanks! `getScopes` is more aligned with `util/sources-tree.js`. It's not really a selector, because we might add more information to the scopes later by fetching an object's properties, and we don't want to rebuild the entire tree every time. If we rebuilt it every time, it would be a selector. But version 2 will be more like the sources tree: it keeps a local version of the tree and adds to it as object properties are fetched.

I'm not sure exactly what that will look like, but `getScopes` will be moved somewhere else for sure.

PR #179, opened 2016-05-19
small tweaks to make first run experience better

jlongster:
Before I open my devtools PR, I wanted to merge in these changes.

I rebased on master and found a few surprising things: `npm start` actually opens a new tab in my browser, which is very surprising. I'd like to remove this behavior, it's more annoying than helpful. If I'm debugging something that's requiring me to restart the server, a new tab is opened every single time even though I already have it open. I personally prefer to make tools more focused and not surprising. Developers can copy (or in iTerm, cmd+click) the URL in the terminal.

Additionally, chrome debugging is turned on by default, which makes Firefox continually spin because the request for chrome tabs never completes (I don't have it on). I definitely think it should be turned off by default. This makes the first run experience a lot better in my opinion.

### DIFF
--- a/bin/development-server.js
+++ b/bin/development-server.js
@@ -5,7 +5,6 @@ const path = require("path");
 const webpack = require("webpack");
 const express = require("express");
 const webpackDevMiddleware = require("webpack-dev-middleware");
-const open = require("openurl").open;
 const http = require("http");
 
 const projectConfig = require("../webpack.config");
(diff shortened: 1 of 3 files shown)


jlongster:
Whoops, bad commit ordering, let me fix.

jasonLaster:
Those changes make sense.

I was going to drop the open url change as well. I didn't expect to be changing `development-server` as much as I was.

+1 on the chrome change as well.

PR #180, opened 2016-05-19
Chrome (part 3)

jasonLaster:
This is just a refactor PR, nothing new was added.

##### cleaned up the way tab ids were managed:
Previously, tab Ids were either firefox actors or chrome tab ids. This
created an issue when it came to persisting an ID in the url query.
Previously, we were keeping just the child/tab of the firefox actor, but
now that it can be either firefox or chrome, the ID was easier. So, to
make this work, the tab ID is now either the chrome tab id or the
firefox child or tab part of the actor.

##### Created a debugTab action
You might be asking, "why not use the entire actor?" The reason is that
the tab actor includes a connection id which is incremented everytime
the debugger is refreshed, which is exactly what we're trying to avoid.
What about if the user changes the tab order in firefox and then
refreshes the debugger? This will probably cause the debugger to debug
a different tab. We should check though...

Reasons:

+ debugTab seems like something you would want in your redux history
+ it's nice for the one action to know to call the firefox or chrome
client debugTab functions
+ it's nice for it to first invoke the select tab action.


public/js/actions/moz.build
```diff
@@ -1,10 +0,0 @@
-# vim: set filetype=python:
-# This Source Code Form is subject to the terms of the Mozilla Public
-# License, v. 2.0. If a copy of the MPL was not distributed with this
-# file, You can obtain one at http://mozilla.org/MPL/2.0/.
-
-DevToolsModules(
-    'breakpoints.js',
-    'event-listeners.js',
-    'sources.js'
-)
```
jasonLaster: surprised this lasted this long

public/js/actions/tabs.js
```diff
@@ -13,24 +16,29 @@ function newTabs(tabs) {
...
 }
 
-function selectTab({ tabActor }) {
-  // set selected tab in the URL hash
-  let childId;
-  if (tabActor.includes("child")) {
-    childId = tabActor.match(/child\d+/)[0];
-  } else {
-    childId = tabActor.match(/tab\d+/)[0];
-  }
```
jasonLaster: happy to kill this

public/js/clients/chrome.js
```diff
@@ -28,7 +28,8 @@ function presentTabs(tabs) {
         title: tab.title,
         url: tab.url,
         id: tab.id,
-        chrome: tab
+        tab,
+        browser: "chrome"
```
jasonLaster: I'm not thrilled with having a tab property in tab, but I'm not sure if there is something better. I'm still waiting on what patterns emerge with sources and other state before i make some decisions

jlongster:
Haven't looked closely at the code, but from what you say, my work in #182 goes a bit in a different direction than this. I don't think `debugTab` should be an action, in fact I think it should be broken up into smaller bits that the "bootloading" code can choose to run. The `selectTab` action is the thing that records it into redux, I don't think we need any more than that.

Let's talk about this more tomorrow, I'll be in a coffee shop so can't vidyo as easily but will be on IRC.

jasonLaster:
closing for now. we can re-open it next week and we re-visit chrome

PR #182, opened 2016-05-20
support library compilation mode for usage inside devtools

jlongster:
This will quickly conflict with your work on Chrome, but here is what I did to get it working as an external bundle. I think the key thing is that the bootstrapping process lives outside of the core system. So `debugTab` does not exist anymore, it's more broken down into smaller pieces that whatever is driving the system can choose what to run.

Saw your other PR and will comment in there as well.

### DIFF
--- a/public/js/main.js
+++ b/public/js/main.js
@@ -21,7 +21,11 @@ if (isEnabled("development")) {
 
 const configureStore = require("./create-store");
 const reducers = require("./reducers");
-const { connectClient, getThreadClient, debugTab } = require("./clients/firefox");
+const {
+  connectClient, connectThread,
+  getThreadClient, setThreadClient,
+  setTabTarget, initPage
+} = require("./clients/firefox");
 const { chromeTabs } = require("./clients/chrome");
 const TabList = React.createFactory(require("./components/TabList"));
 
@@ -36,26 +40,6 @@ const actions = bindActionCreators(require("./actions"), store.dispatch);
 
 // global for debugging purposes only!
 window.store = store;
-
-connectClient(tabs => {
-  actions.newTabs(tabs);
-
-  // if there's a pre-selected tab, connect to it and load the sources.
-  // otherwise, just show the toolbox.
-  if (hasSelectedTab()) {
-    const selectedTab = getSelectedTab(store.getState().tabs.get("tabs"));
-    const tab = selectedTab.get("firefox") || selectedTab.get("chrome");
-    debugTab(tab, actions).then(renderToolbox);
-  } else {
-    renderToolbox();
-  }
-});
-
-if (isEnabled("chrome.debug")) {
-  chromeTabs(response => {
-    actions.newTabs(response);
-  });
-}
 
 /**
  * Check to see if the url hash has a selected tab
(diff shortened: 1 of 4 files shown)


jlongster:
NOTE: I haven't really cleaned this up yet but just posting it because you are working on similar stuff.

public/js/main.js
```diff
@@ -101,3 +85,42 @@ function renderToolbox() {
     document.querySelector("#mount")
   );
 }
+
+if(process.env.NODE_ENV !== "DEVTOOLS_PANEL") {
```
jasonLaster: I'd like to see the feature / environment checking move to `feature.js`. The rationale is that if all of the configuration were in config files, then it would be clear where to look.

This expression could be `isDevtoolsEnvironment`

Here is some pseudo-code for feature.js
https://gist.github.com/6bf0262dd14ae0d0abe0b291d198e3be

public/js/main.js
```diff
@@ -79,17 +63,17 @@ function getSelectedTab(tabs) {
...
-      "Not connected to Firefox"
-    ),
-    document.querySelector("#mount"));
-}, 500);
+//   ReactDOM.render(
+//     React.DOM.div({ className: "not-connected-message" },
+//       "Not connected to Firefox"
+//     ),
+//     document.querySelector("#mount"));
+// }, 500);
```
jasonLaster: this can be wrapped in an `isDevelopment` feature flag

public/js/main.js
```diff
@@ -101,3 +85,42 @@ function renderToolbox() {
...
+
+  if (isEnabled("chrome.debug")) {
+    chromeTabs(response => {
+      actions.newTabs(response);
+    });
+  }
+}
+else {
+  actions.newTabs([{ actor: "tab1" }]);
+  actions.selectTab({ tabActor: "tab1" });
```
jasonLaster: `tab1`?

jasonLaster:
Overall, I like the direction this is headed in. I like how the firefox client is continuing to mature.

This definitely conflicts with my debugTab work, but I'm sure I can reconcile it after this lands.

public/js/main.js
jlongster: My gut feeling is a little different. This is a build flag, and config files (to me) are runtime flags. The risk with moving this to a runtime flag is that somehow the wrong config is used to generate the bundle and when we automatically publish this into Firefox it doesn't work. `webpack.config.devtools.js` wants to unconditionally build a bundle for Firefox.

The other advantage is that it removes a step from the whole process of generating the Firefox bundle, so other people that do it just run `webpack --config webpack.config.devtools.js` and they get it. They don't need to muddle with config.

Another example: locally I'm developing something, then I want to test it in the Firefox devtools. I get the feature working and all I have to do is run that webpack command. I don't need to go into my config to switch the flag, generate it, and then switch it back. If I forget to switch it back I can't just run it locally.

Perhaps I like build-time flags more than config flags, because it seems simpler to coordinate local development.

Maybe there's a way to still specify this flag at build-time from webpack, but actually change the config json value? We could do this with things like `nconf` but I'm not sure the best way to set it here.

public/js/main.js
jlongster: I was running into some weird problems and commented this out to make sure it wasn't that problem, but it wasn't related. I don't see any problem leaving this in here unconditionally, but we could put it behind a flag as well.

public/js/main.js
jlongster: Yeah. The problem is that we've integrated the toolbox so we have to have a tab selected, so we just fake one. We don't actually use the tab from the store anywhere, so it's not a big deal. There might be a better way to do it; I'll think about it some more.

jlongster:
> This definitely conflicts with my debugTab work, but I'm sure I can reconcile it after this lands.

Ok! I wasn't sure which order you wanted to land stuff. I will work on this more today and get it merge-able. (mergable?)

public/js/main.js
jasonLaster: Thanks for the write up. You're rationale makes sense. 

Lets discuss this after your work lands.

jlongster:
There's something wrong with the cypress key. Merging because mocha and linting are fine.

PR #183, opened 2016-05-20
Extract Breakpoint component

jasonLaster:
This is a simple extraction. There are a couple reasons for that:

1. I'd like sufficiently independent components to be in their own file.

2. The editor component will be one of our most complicated components,
so its good to think about ways that we can simplify the file so that it
is easier to understand. Maybe also test...

### DIFF
--- a/public/js/components/Editor.js
+++ b/public/js/components/Editor.js
@@ -2,7 +2,6 @@
 
 const React = require("react");
 const ReactDOM = require("react-dom");
-const Isvg = React.createFactory(require("react-inlinesvg"));
 const ImPropTypes = require("react-immutable-proptypes");
 const { bindActionCreators } = require("redux");
 const { connect } = require("react-redux");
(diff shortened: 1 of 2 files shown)


jlongster:
That component is pretty special though, it's never meant to be reused. I view it as an internal component for the editor, because it solely exists as a hack to translate changes into something the editor can work with. I'm fine with doing this but I'd rather rename the component `EditorBreakpoint` instead of putting it in an `Editor` folder, or you could create an `editor` folder and put both Editor and Breakpoint in there. It feels weird that `Editor.js` would live outside that folder.

jasonLaster:
I'll rename it `EditorBreakpoint.js` we might want to discuss component folder patterns in person next week. I don't feel strongly at the moment, but it does seem like a smell to have a folder with lots of files.

For what it's worth it's pretty manageable now. Just ran a `wc` and we reached 1K lines of component code. I'm not sure if that's an accomplishment

```
      21 public/js/components/TabList.js
      45 public/js/components/App.js
      47 public/js/components/SourceTabs.js
      49 public/js/components/Draggable.js
      54 public/js/components/Tabs.js
      58 public/js/components/Frames.js
      62 public/js/components/SplitBox.js
      70 public/js/components/Accordion.js
      70 public/js/components/RightSidebar.js
     112 public/js/components/Scopes.js
     113 public/js/components/Breakpoints.js
     156 public/js/components/Sources.js
     188 public/js/components/Editor.js
    1045 total
```

jasonLaster:
merged

PR #185, opened 2016-05-20
Start Conditional Breakpoints

jasonLaster:
Turns out that adding conditional breakpoint involves a lot of wiring up :) who would have thought

screen shot 2016-05-20 at 4 01 24 pm

This PR, can be added as is, but it is not feature complete. It adds:

1. a conditional bp panel that is shown when your command click on a bp
2. conditions are added to the bp on enter...

What is not added:
+ there is no panel close button
+ the badge is not changed to orange when it's a conditional bp
+ the condition is not shown in the breakpoint list

### DIFF
--- a/public/js/components/Editor.js
+++ b/public/js/components/Editor.js
@@ -1,6 +1,7 @@
 "use strict";
 
 const React = require("react");
+const ReactDOM = require("react-dom");
 const ImPropTypes = require("react-immutable-proptypes");
 const { bindActionCreators } = require("redux");
 const { connect } = require("react-redux");
@@ -21,6 +22,40 @@ require("./Editor.css");
 require("codemirror/mode/javascript/javascript");
 require("../lib/codemirror.css");
 
+function breakpointAtLine(breakpoints, line) {
+  return breakpoints.find(b => {
+    return b.getIn(["location", "line"]) === line + 1;
+  });
+}
+
+function renderConditionalBreakpointPanel(
+  { location, setBreakpointCondition, closePanel }) {
+  function onKey(e) {
+    if (e.key != "Enter") {
+      return;
+    }
+
+    setBreakpointCondition(location, e.target.value);
+    closePanel();
+  }
+
+  let panel = document.createElement("div");
+  ReactDOM.render(
+    dom.div({
+      className: "conditional-breakpoint-panel"
+    },
+    dom.div(
+      { className: "header" },
+      `This breakpoint will stop execution only
+       if the following expression is true`
+    ),
+    dom.input({
+      onKeyPress: (e) => onKey(e),
+    })
+  ), panel);
+  return panel;
+}
+
 const Editor = React.createClass({
   propTypes: {
     breakpoints: ImPropTypes.map.isRequired,
@@ -29,10 +64,12 @@ const Editor = React.createClass({
     sourceText: PropTypes.object,
     addBreakpoint: PropTypes.func,
     removeBreakpoint: PropTypes.func,
+    setBreakpointCondition: PropTypes.func,
     pause: PropTypes.object
   },
 
   componentDidMount() {
+    this.cbPanels = {};
     this.editor = CodeMirror.fromTextArea(this.refs.editor, {
       mode: "javascript",
       lineNumbers: true,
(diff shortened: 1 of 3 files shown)


public/js/components/Editor.js
```diff
@@ -33,6 +67,7 @@ const Editor = React.createClass({
   },
 
   componentDidMount() {
+    this.cbPanels = {};
```
jasonLaster: I'm not sure if this is the best place for this.

*Why are you adding state?*
1. it is easier to clear the line widget if it is stored in on the component
2. it is easier to have a dictionary of widgets than assume there will only ever be one

jlongster:
A contrary idea: are we sure we want to do this now? This is one of those features that I was thinking was out of scope because it involves a good bit of UX design, and adds complexity. I'd like to keep things simpler and more focused for the first version of the debugger and slowly adds things with care.

However, we are working with Helen in person next week, so we could spend some time on this. I just don't want to add things unless we're going to spend time doing it right. I'm fine if this becomes in scope and we spend time on it; we are pretty ahead of schedule. Since you've already done work on it we can do it, let's make sure to work with Helen next week on the UX though.

public/js/components/Editor.js
jlongster: Yeah, this is fine for now.

public/js/components/Editor.js
```diff
@@ -50,19 +87,37 @@ const Editor = React.createClass({
...
+      actor: this.props.selectedSource.get("actor"),
+      line: line + 1
+    };
+
+    // show conditional breakpoint panel on command click
+    if (bp && ev.metaKey) {
+      const panel = renderConditionalBreakpointPanel({
+        location: location,
+        setBreakpointCondition: this.props.setBreakpointCondition,
+        closePanel: () => this.cbPanels[line].clear()
```
jlongster: You could make the panel get a reference to itself to close it, but you may be thinking that we want to close it from other places too which is why it's stored on the component, so that does make sense.

jasonLaster:
> are we sure we want to do this now?

I agree we don't want to rush into making assumptions about a given workflow. We can also say that this does not fall into scope for an MVP.

Lets discuss the scope on monday. Some of the UX questions I have are:

1) can you create a conditional breakpoint via a context menu on a line or gutter
2) can you create one before a breakpoint or is it only available after a bp is added
3) can you have multiple panels open
4) this is maybe out of scope, but conditional breakpoints are a great way of running *any* expression on a line, logging or doing something else. do we want to highlight that fact. I can see several ways to do that
  + different copy
  + offer the ability to add another expression
  + offer multi-line expressions.

clarkbw:
Agreed that I think this is out of scope for the MVP.  I'm curious what you guys worked out on Monday. 

I'll layout my feelings on the conditional breakpoints here just so they are captured.

 * They are really useful / powerful
 * They are likely rarely used ( by the average developer )
 * Setting them with a hand written expression is error prone
  * an auto-complete entry could help reduce the error prone nature
 * I see them as a higher level breakpoint, will explain below

Maybe I can capture the higher level usage in a user story.

As a developer who is debugging her application I want a way to convert the breakpoint I've set into one that will only trigger at a certain condition such that I am not pressing the resume button over and over again to get to the situation where my breakpoint will be useful.

I do think developers should be able to set a conditional breakpoint immediately, but my hunch is that most developers don't use them despite the fact that they would be extremely useful.  So I think we want to offer them to people who are using breakpoints.  This should be an experiment that we try, but my hunch is that we could get people using conditional breakpoints if we made them simpler and if we did that our developers would be much more effective.

Helen will have better ideas here but I'd suggest we insert an extra line after any breakpoint (perhaps only when we are paused on that breakpoint?).  In that extra line we can offer conditions to help improve the breakpoint itself.  

here's an example of the extra line with a suggestion:
conditional breakpoint

Actually if we're paused on the breakpoint this extra line could be a lot more useful by knowing the variables available within the current scope.  Perhaps that's the best time to show this kind of a setup.

Under the hood we'd probably have to convert the old breakpoint to a conditional one but I don't think that would be a big deal.  We could just treat it like a regular conditional breakpoint from that point on.

So the base story here is that writing a conditional breakpoint from the beginning is a master level type of debugging and I think we want to teach people how to level up to that while they are working.

jasonLaster:
Thanks @clarkbw. I really like what you're discussing as an incremental enhancement once the debugger is released. I would love to make conditional breakpoints / watch expressions more common.

@jlongster I'll close the PR and we can come back to it post-london. Ultimately, I'd like to focus on a simple conditional breakpoint UX. We'll have plenty of time, post-launch to experiment with more intuitive designs.

jlongster:
Sounds good. Whatever we do I don't think we should stray from the basic debugger use case: you click on a line to set a breakpoint that always pauses. We don't want to alienate the large number of users that do use a debugger normally.

But there are tons more involved use cases and we'll figure out a consistent story for the UX of them. There are kinds of things, like "ignore this debugger statement", "ignore breakpoints in this file", etc. Being able to control the execution is really the core part of the debugger so it would be cool to figure out a consistent "language" in the UI for doing this. Autocompletion sounds neat and we should explore that.

PR #186, opened 2016-05-20
Prepare debugger for npm

### LINKED ISSUE #156: Create a standalone application
It would be nice to demo the debugger as a standalone electron app. This would be very easy to do if we publish debugger.html as an npm package.

- [ ] create a debugger.html npm package
- [ ] add a `bin` script for starting the dev-server or add a publish script for bundling.
- [ ] version debugger.html `0.0.1`
- [ ] use the debugger.html index

jasonLaster:
Fixes #156

There were a couple of package.json tweaks needed to publish a debugger.html package
+ add version number
+ add prepublish step
+ add a fields field for public/build

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,6 @@
 {
   "name": "debugger.html",
+  "version": "0.0.4",
   "scripts": {
     "start": "node bin/development-server",
     "lint": "npm run lint-css -s; npm run lint-js -s",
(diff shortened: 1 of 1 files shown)


jlongster:
Great work. Can you explain why we need to publish to npm again? I missed that, so I'm just curious.

jlongster:
r+ btw

clarkbw:
> why we need to publish to npm

(Just for posterity) I believe this is in support of #156, also see the comment that describes where everything lives https://github.com/jlongster/debugger.html/issues/156#issuecomment-220743453

PR #198, opened 2016-05-25
basic support for chrome debugger & more generic app state

jlongster:
This is the result of pairing on Monday. It's a little messy, but we have to move our app state away from Firefox-specific packets and it's going to take some more work. There are already a lot of changes and we decided to land this and clean it up in follow-up PRs. I will clean it up more today.


jlongster:
Not sure about the cypress key change, but I wonder if that is the fix for failing cypress tests (it fails with a bad key error).

jlongster:
Actually, I noticed that the cypress key commit is already on master, so not sure what's going on. Also, I need to fix linting errors. This is bad but I immediately force pushed on master to revert this. Jason and I are the only ones using this repo actively so it should be ok...

I will open a new PR that should not require a squash.

PR #199, opened 2016-05-25
take 2: add chrome debugging

jlongster:
This is the same commit merged before but properly linted. And I removed the cypress key commit.


jlongster:
I have no idea what github is thinking here; I think it's really confused because I'm using the same branch name that was merged but backed out before. Going to start a new branch.

PR #201, opened 2016-05-25
make Firefox client work again with new state structure

jlongster:


jasonLaster:
Looking good. 

+1 extracting loaded frames

public/js/components/Sources.js
```diff
@@ -62,9 +62,9 @@ let File = (props) => {
     )
   );
 };
-File = React.createFactory(File);
```
jlongster: linter was complaining that `File` is a builtin type that we are overriding so I had to change these. Will probably pull them out later today too

jasonLaster: sounds good. good thing we didn't have a Date and Blob icon :)

jlongster:
Cool! Seems like we'll get a good feeling about all of this pretty quickly.

PR #213, opened 2016-05-26
create a few types and prepare code to handle typed objects

jlongster:
Just realized I haven't linted this, so there's probably a lot to do. Especially imstruct.js which is a copy from an internal tcomb file. But if you ignore that file, the rest of the changes are minimal.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,1 +1,2 @@
 public/js/lib/tree.js
+public/js/util/imstruct.js
--- a/package.json
+++ b/package.json
@@ -36,6 +36,7 @@
     "react-redux": "4.4.5",
     "redux": "3.5.2",
     "stylelint": "^6.2.2",
+    "tcomb": "^3.1.0",
     "webpack": "1.13.0",
     "webpack-dev-middleware": "^1.6.1",
     "ws": "^1.0.1"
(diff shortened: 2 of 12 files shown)


jasonLaster:
Looks good. 

What's instruct?

Why do we need fromJS?

jlongster:
It's `imstruct` :) stands for "immutable struct". tcomb has a `struct` type builder which is the standard way for typing objects. `imstruct` allows you do exactly the same thing, but for Immutable.js Records (which are basically `Map`s with predefined fields.

```js
const Person = t.struct({ name: t.String, age: t.Number });
const ImmutablePerson = imstruct({ name: t.String, age: t.Number });
```

These are the same thing, except creating `ImmutablePerson` objects will return normal immutable objects. Creating a `Person` will return a normal JS object.

I think this is a good chance for us to spin off a library at some point for typing immutable JS objects (in the far future, I don't want to deal with that right now). I think most people use Flow, but I don't think we're ready for that until we resolve bigger decisions. If this works out, we'll get most of the benefits of static typing.

I added comments in `fromJS` (you probably looked at an older commit). Immutable's `fromJS` won't convert objects returned from `tcomb` because they aren't "plain" JS objects, meaning that they do some stuff with the prototype chain. This allows us to continue using our current code, but using typed objects in the protocol layer. As mentioned in the other issue (and the comment), eventually once we type all the state we shouldn't need `fromJS` at all because the type system will automatically convert things into the right types.

jlongster:
I will make sure to quickly follow-up with `imstruct.js` (it probably need a better name). Merging this in for now and will do another follow-up.

PR #215, opened 2016-05-27
consolidate client APIs into a generic API interface

jlongster:
I'm already running into several situations where I need to wrap the Firefox protocol to translate something because we changed actors to ids. I think we need to go ahead and create a base API implementation for talking with servers.

This nice thing is that we don't have to deal with client/server communication at all. All we need is a simple programmatic API to talk about debugger data. I don't think we should do it the object-oriented way like the Firefox `main.js` currently is. I suppose eventually we may want to group these functions into separate namespaces, but I'd like them to stay simple functions that take ids on which to operate (for example, `sourceContents` takes a `sourceId`).

I also tweaked `chrome.js` to consolidate the APIs for the client modules into something more similar. This looks like a lot of changes, but it's not too bad, and I'm more than happy to discuss naming conventions, where to put things, etc.

### DIFF
--- a/public/js/clients/chrome.js
+++ b/public/js/clients/chrome.js
@@ -17,13 +17,64 @@ Array.prototype.peekLast = function() {
 
 const bootstrap = require("./chrome/bootstrap");
 
-let connectionAgents;
+let connection;
+let debuggerAgent;
+let runtimeAgent;
 
-function getAgent(name) {
-  return connectionAgents[name];
+// API implementation
+
+let APIClient = {
+  resume() {
+    return debuggerAgent.resume();
+  },
+
+  stepIn() {
+    return debuggerAgent.stepInto();
+  },
+
+  stepOver() {
+    return debuggerAgent.stepOver();
+  },
+
+  stepOut() {
+    return debuggerAgent.stepOut();
+  },
+
+  sourceContents(sourceId) {
+    return debuggerAgent.getScriptSource(sourceId, (_, contents) => ({
+      source: contents,
+      contentType: null
+    }));
+  },
+
+  setBreakpoint(location, condition) {
+    return debuggerAgent.setBreakpoint({
+      scriptId: location.sourceId,
+      lineNumber: location.line,
+      columnNumber: location.column
+    }, (_, breakpointId, actualLocation) => ([
+      {},
+      {
+        id: breakpointId,
+        remove: () => {
+          // TODO: resolve promise when request is completed.
+          return new Promise((resolve, reject) => {
+            resolve(debuggerAgent.removeBreakpoint(breakpointId));
+          });
+        },
+        setCondition: () => {},
+      }
+    ]));
+  }
+};
+
+function getAPIClient() {
+  return APIClient;
 }
 
-function presentTabs(tabs) {
+// Connection handling
+
+function createTabs(tabs) {
   const blacklist = ["New Tab", "Inspectable pages"];
 
   return tabs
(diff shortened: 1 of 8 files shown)


public/js/clients/chrome.js
```diff
@@ -17,13 +17,64 @@ Array.prototype.peekLast = function() {
 
 const bootstrap = require("./chrome/bootstrap");
 
-let connectionAgents;
+let currentConnection;
+let currentDebuggerAgent;
+let currentRuntimeAgent;
```
jlongster: I'm not completely sure about separating out the agents, but this made it so I could use it directly in `APIClient`. I suppose I could always call `getAgent("Debugger")` but this seemed like less work. I'm fine reverting this change though.

jasonLaster: I think this is cleaner.

public/js/clients/chrome.js
```diff
@@ -47,140 +98,90 @@ function presentTabs(tabs) {
...
-    actions.newSource(Source({ id: scriptId, url }));
-  },
-
-  scriptFailedToParse: function() {
-    // needed for debugging nytimes
-    console.log("SCRIPT FAILED TO PARSED", arguments);
-  },
+    scriptFailedToParse() {
+      // needed for debugging nytimes
+      console.log("SCRIPT FAILED TO PARSED", arguments);
```
jasonLaster: lets drop the body of this function :)

public/js/clients/chrome.js
```diff
@@ -47,140 +98,90 @@ function presentTabs(tabs) {
...
 
-function debugChromeTab(tab, actions) {
-  bootstrap(InspectorBackend);
-  WebSocketConnection.Create(
-    tab.webSocketDebuggerUrl,
-    conn => onConnection(conn, actions)
-  );
-  return Promise.resolve();
+  currentRuntimeAgent.enable();
+  currentRuntimeAgent.run();
```
jasonLaster: The *current* naming convention in the clients seems unnecessary. I understand that we might have a different client connection at some point, but if we're not keeping past agents, threads, etc, current seems redundant. (small nit)

public/js/clients/firefox.js
```diff
@@ -9,6 +9,58 @@ let currentClient = null;
...
+    return new Promise(resolve => {
+      currentThreadClient.stepOut(resolve);
+    });
+  },
+
+  getSources() {
+    return currentThreadClient.getSources();
+  },
+
+  sourceContents(sourceId) {
```
jasonLaster: 👍

public/js/clients/firefox.js
```diff
@@ -9,6 +9,58 @@ let currentClient = null;
...
+      line: location.line,
+      column: location.column,
+      condition: condition
+    });
+  }
+};
+
+function getAPIClient() {
+  return APIClient;
+}
```
jasonLaster: in terms of naming/location. I would like to put the api clients in their own file at some point (maybe now). 

It might be helpful if the api, protocol, and bootstrap file in client/chrome were moved to vendor and the build-bootstrap script were moved to bin. I think they belong their anyway as they are 3rd party code. 

If we did that then, I could see a clients/chrome clients/firefox directory having api modules.

public/js/types.js
```diff
@@ -7,7 +7,7 @@ const Source = t.struct({
   url: t.union([t.String, t.Nil]),
 
   // Internal for Firefox for now
-  actor: t.String
+  actor: t.maybe(t.String)
```
jasonLaster: where is this still needed?

jasonLaster:
Looks good. Feel free to merge when you feel comfortable tomorrow

public/js/clients/chrome.js
jlongster: It's meant to be a clear indication that it's a global dynamic object, separating it out from local variables etc. When you see it anywhere in code you know where it comes from (it's not passed in, etc). Honestly, it comes from Scheme/Lisp where they have real dynamic variables all of them are called `current-*` but that convention has been carried over in other places as well.

But I'll remove the `current` prefix, it's not a big deal.

public/js/clients/firefox.js
jlongster: I agree, let's do that as a follow-up. I started feeling the same way as I was writing this.

public/js/types.js
jlongster: If you look at the Firefox client I haven't actually translated id->actor anywhere yet, so it's still used for loading source text and probably other places. I will remove it today.

jlongster:
This is going to conflict with your PR, sorry :( I don't think the merge should be too bad though

public/js/types.js
jlongster: Actually Jason, you were right, the new API automatically works because it just takes the `sourceId` in functions and already passes it as `actor`. I just pushed a small commit to remove this field.

PR #216, opened 2016-05-27
Fixes issue where chrome tabs are not shown when firefox is not connected

### LINKED ISSUE #187: Chrome tabs are not shown when firefox is not connected

jasonLaster:
Fixes #187

I started working on running the integration tests with chrome, but before I could start I first needed to fix this issue. Sharing the PR early for feedback, if you like it i can take the chrome driver work out and merge. Then move to the chrome driver work later.

+ Update firefox.connectClient to handle timing out when firefox is not
connected
+ move the chrome.chromeTabs query up to connectClient so the two
queries are issued in tandem

Also sharing because this might conflict with some of your work with the api clients

### DIFF
--- a/public/js/clients/chrome.js
+++ b/public/js/clients/chrome.js
@@ -45,7 +45,7 @@ function formatTabs(tabs) {
     });
 }
 
-function chromeTabs() {
+function connectClient() {
   if (!isEnabled("chrome.debug")) {
     return Promise.resolve([]);
   }
@@ -188,7 +188,7 @@ function debugChromeTab(tab, actions) {
 }
 
 module.exports = {
-  chromeTabs,
+  connectClient,
   debugChromeTab,
   getAgent,
   ThreadClient
--- a/public/js/clients/chrome.js
+++ b/public/js/clients/chrome.js
@@ -5,6 +5,9 @@ const {
   InspectorBackend
 } = require("./chrome/api");
 
+const { isEnabled } = require("../configs/feature");
+const defer = require("../util/defer");
+
 /* eslint-disable */
 // TODO: figure out a way to avoid patching native prototypes.
 // Unfortunately the Chrome client requires it to work.
@@ -21,7 +24,7 @@ function getAgent(name) {
   return connectionAgents[name];
 }
 
-function presentTabs(tabs) {
+function formatTabs(tabs) {
   const blacklist = ["New Tab", "Inspectable pages"];
 
   return tabs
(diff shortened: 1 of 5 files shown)


public/js/clients/firefox.js
```diff
@@ -40,16 +42,35 @@ function presentTabs(tabs) {
...
   currentClient = new DebuggerClient(transport);
 
+
+  setTimeout(() => {
+    if (isConnected) {
+      return;
+    }
+
+    deferred.resolve([]);
+  }, 1000);
```
jasonLaster: needed to timeout when failed to connect to firefox

jasonLaster:
Just removed the chrome driver work and added some polish.

public/js/util/defer.js
```diff
@@ -0,0 +1,16 @@
...
+    reject = arguments[1];
+  });
+  return {
+    resolve: resolve,
+    reject: reject,
+    promise: promise
+  };
+}
+
+module.exports = defer;
```
jasonLaster: added defer to be in line with devtools. It might make sense to move this to the vender directory... Although probably not a big deal

public/js/clients/chrome.js
```diff
@@ -42,12 +45,19 @@ function presentTabs(tabs) {
...
+function connectClient() {
+  if (!isEnabled("chrome.debug")) {
+    return Promise.resolve([]);
+  }
+
+  const deferred = defer();
   fetch("/chrome-tabs").then(res => {
     res.json().then((body) => {
-      callback(presentTabs(body));
+      deferred.resolve(formatTabs(body));
```
jasonLaster: switched from callback pattern to promises so that I could call both clients with Promise.all

public/js/util/defer.js
jlongster: Yeah I don't have a good feeling about it. I think the most important thing is that we write code that is similar to other devtools code, and it doesn't matter where these modules live. We may want to merge `util` and `lib` somehow at some point.

public/js/main.js
```diff
@@ -91,8 +90,9 @@ function renderToolbox() {
 }
 
 if (process.env.NODE_ENV !== "DEVTOOLS_PANEL") {
-  connectClient(tabs => {
+  connectClients().then((tabs) => {
     actions.newTabs(tabs);
+    renderNotConnected();
```
jlongster: It's confusing that you call `renderNotConnected` unconditionally; this feels like something that should be inside `renderToolbox`.

I just looked more at the code, and it took me a little bit to figure it out; `renderNotConnected` doesn't do anything if there are tabs. Personally I'd prefer to move that `isEmpty` check down here so that flow is more obvious and `renderNotConnected` only does 1 thing.

public/js/clients/firefox.js
jlongster: Nice, I like that this is closer to the connection code and there's not a top-level `setTimeout` any more.

jlongster:
This looks like a nice cleanup, good job! r+

public/js/main.js
jasonLaster: Yup. I like the render toolbox suggestion. I'd like to clean that up too. Tablist is a bad name and we'll have a node UI at some point too. Thinking about calling it a toolbox and just expecting it to change over time

jasonLaster: I'll make these changes in another pass on the toolbox.

jasonLaster:
merged

PR #217, opened 2016-05-27
remove loadSources action and move Firefox-specific loading logic into the Firefox client

jlongster:
The Chrome client never calls `loadSources` because it doesn't need to. It automatically fires `newSource` notifications whenever you connect. We need to call it for the Firefox client though. Additionally, for Firefox we don't actually need to handle the response because Firefox will fire individual `newSource` notifications.

This clarifies that the generic protocol interface is to always fire `newSource` actions to add sources to the system.

There is one edge case: when Firefox uses bfcache. bfcache is when Firefox loads the page straight from memory when you press "back". In this case we won't get `newSource `notifications when navigating, but this seems more like a problem for the server. The existing debugger actually has this problem: it won't show sources from bfcache either. So I don't think we need to worry about it (we could work around it by manually requesting all the sources, but we don't know when we are using bfcache)


public/js/clients/firefox.js
```diff
@@ -146,6 +146,45 @@ function createFrame(frame) {
 }
 
 const CALL_STACK_PAGE_SIZE = 25;
+const NEW_SOURCE_IGNORED_URLS = ["debugger eval code", "XStringBundle"];
+
+function makeDispatcher(client, actions) {
```
jasonLaster: Really smart.

public/js/actions/sources.js
```diff
@@ -13,30 +12,22 @@ const { getSource, getSourceText } = require("../selectors");
...
-  return ({ dispatch }) => {
-    // Ignore bogus scripts, e.g. generated from 'clientEvaluate' packets.
-    if (NEW_SOURCE_IGNORED_URLS.indexOf(source.url) != -1) {
-      return (new Promise()).resolve();
-    }
-
-    return dispatch({
-      type: constants.ADD_SOURCE,
-      source: source
-    });
```
jasonLaster: Nice cleanup

public/js/clients/firefox.js
```diff
@@ -146,6 +146,45 @@ function createFrame(frame) {
...
+      // slow script dialog and internal events such as setting
+      // breakpoints, ignore the event.
+      if (packet.why.type === "interrupted" && !packet.why.onNext) {
+        return;
+      }
+
+      // Eagerly fetch the frames
+      client.getFrames(0, CALL_STACK_PAGE_SIZE, res => {
+        actions.loadedFrames(res.frames.map(createFrame));
+      });
```
jasonLaster: Is this replacing loadSources?

jlongster: Not sure what you mean; loading the frames is different than the sources. The main difference in here is at the bottom of `initPage` we call `client.getSources` instead of `actions.loadSources`. We were doing this before too; I just lumped in this event dispatcher stuff in here as well, but it just moved the previous event handlers up into here.

jasonLaster: Got it. Was tough to track the changes on my phone.

jlongster: Ah. You are reviewing on your phone?! Don't feel like you need to, go enjoy the conference :)

jlongster:
Thanks for reviewing so fast! Again don't feel like you have to!

PR #218, opened 2016-05-27
fix locations of chrome breakpoints and improve debug line

jlongster:
I feel like doing some polish work today :)

Chrome expects 0-based line locations everywhere. We use 1-based lines. Not only were we using `functionLocation` instead of `location`, but we were setting breakpoints on the wrong line (we should be setting them on `line - 1`).

This makes Chrome debugging work as expected.

### DIFF
--- a/public/js/components/Editor.js
+++ b/public/js/components/Editor.js
@@ -8,7 +8,7 @@ const CodeMirror = require("codemirror");
 const { DOM: dom, PropTypes } = React;
 
 const {
-  getSourceText, getPause, getBreakpointsForSource,
+  getSourceText, getBreakpointsForSource,
   getSelectedSource, getSelectedSourceOpts,
   makeLocationId
 } = require("../selectors");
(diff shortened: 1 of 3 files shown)


jlongster:
The highlighted line is still wonky, in that it's not cleared when it should be and such. That may be the last thing I try to improve today.

public/js/components/Editor.js
```diff
@@ -152,8 +143,7 @@ module.exports = connect(
       selectedSource: selectedSource,
       selectedSourceOpts: getSelectedSourceOpts(state),
       sourceText: getSourceText(state, selectedId),
-      breakpoints: getBreakpointsForSource(state, selectedId),
-      pause: getPause(state)
+      breakpoints: getBreakpointsForSource(state, selectedId)
```
jasonLaster: There might be a pause prop as well

jasonLaster:
Looks good

public/js/components/Editor.js
jlongster: good point, there is. I will fix that in a follow-up PR (that I'm already working on)

jlongster:
I'm working on the editor state some more and I think we do need to depend on the paused state. I will open a new PR and explain there.

PR #219, opened 2016-05-27
Added ability to collapse folders, resolves #160

### LINKED ISSUE #160: Collapse folders with a single child in the sources tree view
Right now the tree view creates a new level for every single folder. However, if a folder only contains a single child, we should collapse it so

```
* foo
  * bar
    * baz.js
```

becomes

```
* foo/bar/baz.js
```

TatumCreative:
Not quite ready for a review yet, but I wanted to get this uploaded for now.

I am running into one issue where some nodes can share a path. I'm not sure how this was previously handled. Looks like the same path, loaded twice, with a different hash... Not sure if the content was different.

```
[
   {
      "id":"server3.conn16.child2/167",
      "url":"http://ads.pubmatic.com/AdServer/js/showad.js",
      "actor":"server3.conn16.child2/167",
      "pathname":"/AdServer/js/showad.js",
      "filename":"showad.js"
   },
   {
      "id":"server3.conn16.child2/196",
      "url":"http://ads.pubmatic.com/AdServer/js/showad.js#PIX&p=32575&s=39805&a=60123&it=0&kdntuid=1",
      "actor":"server3.conn16.child2/196",
      "pathname":"/AdServer/js/showad.js",
      "filename":"showad.js"
   }
]
```

### DIFF
--- a/public/js/util/sources-tree.js
+++ b/public/js/util/sources-tree.js
@@ -79,6 +79,10 @@ function getSharedParts(partsA, partsB) {
   return sharedParts;
 }
 
+function isCollapsed(node) {
+  return node.baseName !== node.name;
+}
+
 function addToTree(currentNode, source) {
   const url = getURL(source);
   if (!url) {
@@ -88,6 +92,13 @@ function addToTree(currentNode, source) {
   const parts = url.pathname.split("/").filter(p => p !== "");
   const isDir = partsIsDir(parts);
   parts.unshift(url.host);
+  if(url.hash) {
+    // Add on the URL hash if there is one. Some sites load in duplicate sources
+    // with different contents.
+    //
+    // TODO - Figure out a more elegant way of handling duplicate sources.
+    parts[parts.length - 1] += url.hash;
+  }
 
   let accumulatedPath = "";
 
@@ -109,7 +120,7 @@ function addToTree(currentNode, source) {
 
     if (index >= 0 && contents[index].baseName === part) {
       // A node with the same name already exists
-      if (contents[index].name !== contents[index].baseName) {
+      if (isCollapsed(contents[index])) {
         // This node is collapsed, merge in shared nodes.
         currentNode = mergeIntoCollapsedNode(contents[index], accumulatedPath,
                                              parts, i);
@@ -124,7 +135,7 @@ function addToTree(currentNode, source) {
       // exist.
       currentNode = insertNodeIntoContents(contents, accumulatedPath, parts, i,
                                            index);
-      if (currentNode.name !== currentNode.baseName) {
+      if (isCollapsed(currentNode)) {
         // Jump to the end of this loop to insert a final leaf node.
         i = parts.length - 2;
       }
--- a/public/js/util/sources-tree.js
+++ b/public/js/util/sources-tree.js
@@ -9,11 +9,14 @@ function nodeHasChildren(item) {
   return item.contents instanceof Array;
 }
 
-function createNode(name, path, contents) {
+function createNode(name, path, contents, baseName = name) {
   return {
     name,
     path,
-    contents: contents || null
+    contents: contents || null,
+    // If a node is collaped, the base folder name, e.g. a node with the name
+    // "a/b/c" would have the baseName "a"
+    baseName
   };
 }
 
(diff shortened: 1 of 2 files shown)


jlongster:
I *think* previously that the hash was part of the filename. If not that's how it should work. There's a worse case though: when the same JS file is loaded twice. We have a horrible story for that right now in the current debugger: you never see the new sources, you always see the first one. In the next few months it would be cool to figure out a good way to show multiple sources with the same URL, but we don't need to do that now.

The hash case should be solved by including the hash when parsing out the filename. If you tried that, was there a problem?

jlongster:
This is amazing. I just tried it out and I don't see any noticeable performance degradation. cnn.com has always been really messy and sometimes grins Firefox to a halt for a little bit when debugging. I noticed the same thing before this work, so it's not your changes. I don't think it's the source tree, but it could be. When I looked into it before it was more about how many sources cnn.com was creating and flooding the WebSocket with `newSource` packets. We'll have to look into that at some point, but definitely shouldn't block this work.

I noticed once place when I went to cnn.com that seemed like it should have been collapsed:

screen shot 2016-05-27 at 3 43 21 pm

But it's not really a big deal. This is way better then how it currently works so I'm fine merging it in. I'll look over the code again and see if I have any other comments about it.

public/js/util/tests/sources-tree.js
```diff
@@ -7,6 +7,109 @@ const {
...
+  const abcdeSource = Map({
+    url: "http://example.com/a/b/c/d/e.js",
+    actor: "actor2"
+  });
+  const abxSource = Map({
+    url: "http://example.com/a/b/x.js",
+    actor: "actor3"
+  });
+
+  it("can collapse a single source", () => {
```
jlongster: do you mind moving these collapse tests below the tests that were already in here? these seem like tests for the more advanced feature of collapsing and should be after the tests of the basic API

TatumCreative: Yeah, good point.

TatumCreative:
I measured timing on the addToTree function, and it wasn't noticeably slower for my code changes. I think it was the eval loop you had pointed out before on cnn.com that was causing the issues. In addition I found the duplicated node name. Those sites have crazy sources :-0 it's basically throwing fuzz at your code.

public/js/util/sources-tree.js
```diff
@@ -59,59 +62,167 @@ function getURL(source) {
...
-
-    if (idx >= 0 && subpaths[idx].name === part) {
-      // A node with the same name already exists, simply traverse
-      // into it.
-      subtree = subpaths[idx];
+    let index = getSortedIndex(contents, part);
+
+    if (index >= 0 && contents[index].baseName === part) {
+      // A node with the same name already exists
+      if (contents[index].name !== contents[index].baseName) {
```
jlongster: This code is a bit complex (inherently so), so I'm looking for any way to make it easier to follow. What do you think about making a `isCollapsed` function that runs his logic? You use it below as well. I don't care about abstracting out this logic as much as making this workflow easier to follow.

public/js/util/sources-tree.js
```diff
@@ -59,59 +62,167 @@ function getURL(source) {
...
-      subpaths.splice(where, 0, node);
-      subtree = subpaths[where];
+      // Insert a new node, optionally collapsed, in the correct
+      // alphabetically sorted location, since a node with that name did not
+      // exist.
+      currentNode = insertNodeIntoContents(contents, accumulatedPath, parts, i,
+                                           index);
+      if (currentNode.name !== currentNode.baseName) {
+        // Jump to the end of this loop to insert a final leaf node.
+        i = parts.length - 2;
```
jlongster: This makes sense, but I wonder if we should iterate over `parts.slice(0, parts.length - 1)`, meaning don't include the last part in this loop. It's already a little weird that overwrite the contents of the final node below. This way might a little clear, and we create the leaf node explicitly and add it (now that you have `getSortedIndex`, it shouldn't be much more verbose).

I may be missing a problem with that though, and it if takes too much time don't worry about it. We can massage this code over time.

jlongster: Meant to say this: that would mean that this could just be a `break` then.

public/js/util/sources-tree.js
```diff
@@ -59,59 +62,167 @@ function getURL(source) {
...
+    part = parts.slice(partsIndex, parts.length - 1).join("/");
+  }
+
+  const node = createNode(part, path + "/" + part, [], baseName);
+  const where = contentsIndex === -1 ? contents.length : contentsIndex;
+  contents.splice(where, 0, node);
+  return node;
+}
+
+function mergeIntoCollapsedNode(target, path, parts, partsIndex) {
```
jlongster: Pretty slick how this works.

jlongster:
> Those sites have crazy sources :-0 it's basically throwing fuzz at your code.

Haha, yep. It's critical that the new debugger works with the "crazy ones" because if it works with those, it should work for the majority of people.

TatumCreative:
@jlongster I'm planning on addressing your comments on `parts.slice(0, parts.length - 1)` in a follow-up.

jlongster:
Looks good, let's land this! Some of the integration tests are failing, but I think we should work with Jason to fix that as he's working through those right now.

jasonLaster:
@jlongster yea - those are supposed to fail :) The tree changed

jasonLaster:
@tatumcreative do you think this can be merged now that we batch update sources?

jlongster:
Last I know he was still working through some bugs in it (not grouping things correctly) and he said he wanted some more time to work on it, so we shouldn't bet on it for the demo. I'm ok with that, it's a nice feature but not critical for the demo.

PR #222, opened 2016-05-31
make breakpoints more robust and improve protocol interface

jlongster:
Breakpoints were pretty broken before: sliding breakpoints wouldn't work, and the data returned from each client was actually a bit different. I went ahead and improved the protocol so that "breakpoint clients" are only something the Firefox client has to be concerned with now. Breakpoints should be a lot more robust: sliding works, and adding & removing seems to always work now (previously sometimes I couldn't remove a breakpoint because it got into a bad state).

I also fixed "snippets" with sliding breakpoints: previously it would show "undefined" if the breakpoint moved. Now it properly gets the text from the line that it moved to.

r?

### DIFF
--- a/public/js/clients/firefox.js
+++ b/public/js/clients/firefox.js
@@ -3,7 +3,7 @@
 const { DebuggerClient } = require("ff-devtools-libs/shared/client/main");
 const { DebuggerTransport } = require("ff-devtools-libs/transport/transport");
 const { TargetFactory } = require("ff-devtools-libs/client/framework/target");
-const { Tab, Source, Location, Frame } = require("../types");
+const { Tab, Source, Location, BreakpointResult, Frame } = require("../types");
 const defer = require("../util/defer");
 
 let currentClient = null;
@@ -13,6 +13,8 @@ let currentTabTarget = null;
 // API implementation
 
 let APIClient = {
+  _bpClients: {},
+
   resume() {
     return new Promise(resolve => {
       currentThreadClient.resume(resolve);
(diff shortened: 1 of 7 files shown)


public/js/clients/firefox.js
```diff
@@ -54,7 +56,29 @@ let APIClient = {
...
+      const actualLocation = res.actualLocation ? {
+        sourceId: res.actualLocation.source.actor,
+        line: res.actualLocation.line,
+        column: res.actualLocation.column
+      } : location;
+
+      return {
+        id: bpClient.actor,
+        actualLocation: actualLocation
+      };
```
jasonLaster: do we want a Breakpoint type?

public/js/components/Editor.js
```diff
@@ -58,12 +58,18 @@ const Editor = React.createClass({
...
+    if (bp) {
+      this.props.removeBreakpoint({
+        sourceId: this.props.selectedSource.get("id"),
+        line: line + 1
+      });
+    } else {
+      this.props.addBreakpoint(
+        { sourceId: this.props.selectedSource.get("id"),
+          line: line + 1 },
+        { getTextForLine: l => cm.getLine(l - 1).trim() }
```
jasonLaster: why are we passing a function in instead of a value?

public/js/reducers/breakpoints.js
```diff
@@ -42,38 +49,29 @@ function update(state = initialState, action) {
...
         // If the breakpoint moved, update the map
-        if (actualLocation) {
-          // XXX Bug 1227417: The `setBreakpoint` RDP request rdp
-          // request returns an `actualLocation` field that doesn't
-          // conform to the regular { actor, line } location shape, but
-          // it has a `source` field. We should fix that.
-          actualLocation = { actor: actualLocation.source.actor,
-                             line: actualLocation.line };
-
+        if (locationMoved(location, actualLocation)) {
```
jasonLaster: nice

jasonLaster:
r+ 

Wow. That was a lot of fixes. Great job

public/js/clients/firefox.js
jlongster: Not a bad idea, but it's a little more confusing than Source or Location because in most places a breakpoint just *is* a `Location`. The UI doesn't have any notion of a `breakpointId` like does a `sourceId`, because a `Location` is everything is needs to talk about breakpoints. At the protocol level we want an `id` and other things like `actualLocation`.

We could make a breakpoint type for the protocol though, with `id`, `location`, and `actualLocation` fields. It's clearly for the protocol, given that it has two location fields, which we can use to know if it's moved.

public/js/components/Editor.js
jlongster: Unfortunately, we need to get the text if a breakpoint slides. We only know if it slides once the protocol request completes, at which point we need to query the editor again, but we have no way to do that. A function allows us to compute the new value if it slides.

jlongster:
I added breakpoints types, but the actual `Breakpoint` is not being used yet, it will be used in the reducer. The protocol uses the type `BreakpointResult` which only contains information to indicate if the breakpoint was moved.

PR #223, opened 2016-05-31
(WIP) Add Chrome Testing

jasonLaster:
Sharing the work for writing integration tests for Chrome fairly early. 

Context:
+ Firefox integration tests use webdriver. They work because it uses the firefox extension support. Chrome cannot use webdriver because, chrome webdriver uses the remote debugger protocol. There can only be one connection.
+ There are two alternative approaches. 1) drive chrome with the debugger protocol [1](). This is cumbersome and i believe flaky for all of the reasons selenium is flakey. 2) drive chrome with native JS, that is evaluated client side. There are limitations to what can be simulated, but none that I see biting us for now. This implementation feels fairly reliable. Also, there's the added benefit of being able to remove a lot of the custom implementation with Cypress' client-side driver in the near future.

Progress:
+ The integration tests work with Chrome
+ I need to add a chrome launcher + ability to navigate to a URL with chrome
+ I need to get it to work w/ firefox again
+ I still need to do cleanup


public/js/clients/chrome.js
```diff
@@ -184,9 +192,24 @@ function initPage(actions) {
...
+  }
+  const script = `window.Agent = (${Agent})()`;
+  console.log(script);
+  runtimeAgent.evaluate(script, (_, result) => {
+    isAgentInjected = true;
+    console.log(result);
+  });
+}
+
+window.injectAgent = injectAgent;
```
jasonLaster: + will remove this global
+ will likely move this function out as it's just for testing.

public/js/clients/index.js
```diff
@@ -19,13 +19,17 @@ function debugPage(tab, actions) {
     if (isFirefox) {
       yield firefox.connectTab(tab.get("tab"));
       firefox.initPage(actions);
+      window.apiClient = firefox.getAPIClient();
     } else {
       yield chrome.connectTab(tab.get("tab"));
       chrome.initPage(actions);
+      window.apiClient = chrome.getAPIClient();
```
jasonLaster: might leave this in with a note that it is just for debugging/testing

public/js/test/integration/todomvc.js
```diff
@@ -49,6 +49,23 @@ function resumeDebugger() {
...
+    debuggee(() => {
+      let input = driver.findElement(By.id("new-todo"));
+      input.sendKeys("yo yo yo", Key.ENTER);
+    });
+  } else {
+    cy.window().then(win => {
+      win.injectAgent();
+      win.apiClient.evaluate(`
+        Agent.type("#new-todo", "hi")
+        Agent.type("#new-todo", "{enter}")
```
jasonLaster: this is a fairly simple API. that supports type, click, dblclick. It's somewhat clunky, but I think we might be able to replace this with cypress' API which would be significantly cleaner

public/js/test/utils/agent.js
```diff
@@ -0,0 +1,94 @@
+module.exports = function() {
```
jasonLaster: I'll probably rename Agent to something more descriptive.

I'll also add comments.

public/js/clients/index.js
jlongster: I don't feel strongly about it, but since we do it for the store too, it might be nice to have a function like `getDebugObjects` or something that you can destructure at the console (`const { store } = getDebugObjects()`) so that they are all grouped in one place. Or maybe that function sticks it on window itself, so you just call `injectDebugObjects()`. We can do this later.

public/js/clients/chrome.js
jlongster: Yeah, we could export a getter for `runtimeAgent` if you want to do this from somewhere else

public/js/test/integration/todomvc.js
jlongster: Nice!

jlongster:
Looks great! It's a little unfortunate that we have to create our own API, but I don't know what else we could do. Seems like great research and we'll see how it feels.

jlongster:
r+

public/js/clients/index.js
jasonLaster: yea, that seems right.

store and client seem like the right abstractions, maybe bound actions, but tough to say now

jasonLaster:
@jlongster mind doing a second review.

Couple of updates:

1. Firefox has been converted to using the injected client commands (debuggee commands). Converting Firefox means that we can drop the test server, which was responsible for executing the web driver commands. The client side version also feels more stable.

2. Chrome tests are disabled off by default, but can be turned on if the config enables chrome and chrome is running. They're currently disabled, because I need to see how to launch chrome on circle, should not be hard.

3. I inlined todomvc. This will stabalize todomvc. Also, it will make it easier to write new test examples. The examples are hosted on the development-server.

4. I added a couple of client methods (evaluate, navigate). They were both important for integration tests. It is not ideal to add client methods just for testing, but I think the advantage outweighs the drawbacks for now while we're experimenting with testing.

5. Test cleanup - i dropped all but one test to simplify the suite. The final one tests adding a todo, with a breakpoint, and was the main test.

public/js/clients/chrome.js
```diff
@@ -80,6 +81,20 @@ let APIClient = {
     return new Promise((resolve, reject) => {
       resolve(debuggerAgent.removeBreakpoint(breakpointId));
     });
+  },
+
+  evaluate(script) {
+    return runtimeAgent.evaluate(script, (_, result) => {
+      console.log("console.evalute", result);
```
jlongster: Did you mean to leave these console logs in?

PR #225, opened 2016-06-01
pixel-perfect breakpoint icons

jlongster:
Breakpoints before looked a little off and it was driving me crazy. This makes them align perfectly with the line numbers, and also load instantly. 

On my computer, I'm still having the issue where line numbers are slightly off, so the breakpoints are 1px off for me but they should be perfectly aligned for everyone else. I still need to figure out the line number issue that I'm seeing.

### DIFF
--- a/public/js/components/Editor.js
+++ b/public/js/components/Editor.js
@@ -13,7 +13,7 @@ const {
   getSelectedFrame, makeLocationId
 } = require("../selectors");
 const actions = require("../actions");
-const { alignLine, onWheel } = require("../util/editor");
+const { alignLine, onWheel, resizeBreakpointGutter } = require("../util/editor");
 const Breakpoint = React.createFactory(require("./EditorBreakpoint"));
 
 require("codemirror/lib/codemirror.css");
@@ -51,8 +51,12 @@ const Editor = React.createClass({
 
     this.editor.on("gutterClick", this.onGutterClick);
 
-    this.editor.getScrollerElement().addEventListener("wheel",
-      ev => onWheel(this.editor, ev));
+    this.editor.getScrollerElement().addEventListener(
+      "wheel",
+      ev => onWheel(this.editor, ev)
+    );
+
+    resizeBreakpointGutter(this.editor);
   },
 
   onGutterClick(cm, line, gutter, ev) {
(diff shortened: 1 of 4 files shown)


public/js/components/Editor.js
```diff
@@ -77,6 +81,13 @@ const Editor = React.createClass({
     }
   },
 
+  resizeBreakpointGutter() {
```
jlongster: The technique here anchors the breakpoint to the beginning of the line, or the "right side" of the line number gutter. That way if there are 10,000,000 lines and the gutter expands really large, the breakpoint will be anchored to the right side and will travel with it. Unfortunately I need to do a little manual layout for that.

jasonLaster: This code seems like it should live in `util/editor` as it is CM specific and does not deal with the business logic of debugging per-se

public/js/components/EditorBreakpoint.js
```diff
@@ -1,20 +1,23 @@
...
+  svg.setAttribute("preserveAspectRatio", "none");
+  let path = document.createElementNS("http://www.w3.org/2000/svg", "path");
+  // Until we figure out our loader story, embed it directly so we can
+  // control it with CSS.
+  path.setAttribute("d", "M27.1,0H1C0.4,0,0,0.4,0,1v10c0,0.6,0.4,1,1,1h26.1 c0.6,0,1.2-0.3,1.5-0.7L33,6l-4.4-5.3C28.2,0.3,27.7,0,27.1,0z"); // eslint-disable-line max-len
+  svg.appendChild(path);
+  marker.appendChild(svg);
+
   return marker;
 }
```
jasonLaster: nice

public/js/components/Editor.js
jlongster: That's fine with me

jlongster:
I asked Helen to look at this too, so let's not merge until she's had a look

helenvholmes:
Here's a huge debugger breakpoint (hopefully this is correct): http://cl.ly/0V392o061S2O

Here's a screenshot of the issue at very large files: http://cl.ly/3S0q1a3G3P2X

bgrins' test page: http://bgrins.github.io/devtools-demos/debugger/scripts.html

jlongster:
Fixed the wide breakpoints, and also fixed the mis-aligned line on my computer. I didn't have "SF UI Text" installed and the line number gutter was trying to use that, but it should be using the monospaced font. Everything looks so much better now!!

PR #232, opened 2016-06-02
Fix #230, stepOut icon now has lines along bottom

### LINKED ISSUE #230: Fix step out button
Lines need to be at the bottom

helenvholmes:
This is going to look fine on retina macs, but I'm going to file a follow-up bug to check that it's pixelsnapped on non-retina screens (can't check until I'm work with my other monitor).


jlongster:
Cool!

PR #234, opened 2016-06-02
Update Fixtures

jasonLaster:
Updates two fixture files as well as adds automation so that they are updated
as the integration tests are run.


public/js/test/integration/todomvc.js
```diff
@@ -14,6 +14,11 @@ function goToSource(source) {
 }
 
 function togglBreakpoint(linenumber) {
+  cy.window().then(win => {
+    cy.wrap(win.cm)
+      .invoke("scrollIntoView", { line: linenumber, ch: 0 });
+  });
```
jasonLaster: This command scrolls codemirror to the breakpoint line we want to toggle.

This is the only place where codemirror is used in the tests and I figured it was worth exposing codemirror so the tests would not have to figureout where the line was.

`cy.window`, `cy.wrap`, `invoke` are somewhat awkward functions for getting the debugger window context, and invoking functions on window objects. They're documented here: https://docs.cypress.io/v1.0/docs/invoke and simpler than the mochitest equivelants https://docs.cypress.io/v1.0/docs/invoke.

public/js/test/integration/todomvc.js
```diff
@@ -55,7 +74,13 @@ function debuggee(callback) {
...
     // The problem is if, the client pauses we need to handle that
     // gracefully and resume. We did this on the test-server.
-    win.apiClient.evaluate(script);
+    win.apiClient.evaluate(script).then(response => {
+      if (response.exception) {
+        const errorMsg = response.exceptionMessage;
+        const commandInput = response.input;
+        console.error(`${errorMsg}\n For command:\n${commandInput}`);
+      }
+    });
```
jasonLaster: One thing I was noticing was that the client "debuggee" script would fail silently when trying to click on an element or type. Adding early errors to the script and logging them in the tests makes it easier to debug.

jasonLaster:
oops - linting failed. fixing now

public/js/components/Editor.js
```diff
@@ -49,6 +49,9 @@ const Editor = React.createClass({
       gutters: ["breakpoints"]
     });
 
+    // NOTE: only used for testing
+    window.cm = this.editor;
```
jlongster: I'd rather avoid littering these around our code, as mentioned in the other places as well. I'm pretty sure you can get the CodeMirror instance from the DOM element that it sits in. That's a lot cleaner than this, so we should look into doing that instead.

I don't see it in the docs off-hand, but I'll look more. You can merge this but I'll file an issue to fix this.

jlongster:
+1 looks good!

public/js/components/Editor.js
jasonLaster: sounds good. I remember doing that as well, although that might be how we're doing it in Firefox. I agree about the littering aspect. I was thinking about namespacing them or wrapping them in a function like `addDebugGlobal` but figured I'd wait

jlongster:
I'm fine if you want to merge this now and clean up as we go along.

bin/fixtures.js
```diff
@@ -0,0 +1,28 @@
+"use strict";
+
+const fs = require('fs');
+const path = require('path');
+
+const fixturePath = path.join(__dirname, "../public/js/test/fixtures");
+
+function saveFixture(name, text) {
```
jlongster: I've come around to this idea, I like it.

jasonLaster:
Merged

PR #235, opened 2016-06-02
batch newSource dispatches to reduce rendering churn

jlongster:


jlongster:
Need to fix up some stuff here but you get the idea

public/js/reducers/sources.js
```diff
@@ -17,8 +17,13 @@ const initialState = fromJS({
 function update(state = initialState, action) {
   switch (action.type) {
     case constants.ADD_SOURCE:
-      return state.mergeIn(["sources", action.source.id],
-                           _updateSource(action.source));
```
jasonLaster: are we not using `_updateSource` anymore?

public/js/actions/sources.js
```diff
@@ -12,17 +12,35 @@ const { getSource, getSourceText } = require("../selectors");
...
+    // Even just batching every 10ms works because we just want to
+    // group sources that come in at once.
+    newSourceTimeout = setTimeout(() => {
+      dispatch({ type: constants.ADD_SOURCES,
+                 sources: newSources });
+      newSources = [];
+      newSourceTimeout = null;
+    }, 10);
+  }
+}
```
jasonLaster: It might be nice to put a simple debounce function in util and use that

```js
const newSources = debounce(function(dispatch) {
  dispatch({ 
    type: constants.ADD_SOURCES,
    sources: newSourcesList
  });
  newSourcesList = [];
}); 
```

jlongster: `debounce` seems like the wrong abstraction though, at least for what `debounce` commonly means. Usually `debounce` means to run a function a certain amount of time after the debounced function is called, and if it is called again it resets the timer. We don't want that: we want to schedule a dispatch 10ms later, no matter what. It's more of a "throttle". Also, debounced functions don't usually depend on external state like that. This is an explicit function to queue a dispatch that can't be transparently debounced.

There probably is an abstraction here, maybe one that we could even tailor to actions specifically. I'd like to do that later though :)

public/js/actions/tests/newSource.js
```diff
@@ -1,14 +1,21 @@
...
 const fixtures = require("../../test/fixtures/foobar.json");
-const { newSource } = actions;
 const { getSourceByActor } = selectors;
 const sourcesFixtures = fixtures.sources.sources;
 
 const store = createStore();
 const expect = require("expect.js");
 
+// Write our own `newSource` to bypass the batching logic.
+function newSource(source) {
```
jlongster: I don't love this, but I'm not sure what else to do. Delaying things makes it hard to tests. One of the things I like about redux is it's so easy to think about the order in which things happen...

I really like the idea of implementing the throttling at the component level, because it's the sources component specifically that we want to avoid rerendering. That would solve this issue. More to think about...

jlongster:
BTW, this takes rendering node's sources (about 560) from taking 20 seconds to render to almost immediate.

public/js/reducers/sources.js
jlongster: Nope, since the sources tree nothing uses those fields anymore.

public/js/actions/sources.js
jasonLaster: Hmm. Yea, I liked the choice of queueing. Debouncing is usually not aggregation.

public/js/actions/tests/newSource.js
jasonLaster: I agree, I liked preserving the synchronous nature of action to render. Which this actually still does but delaying the action. 

Are you thinking we'd throttle in connect? That might make it tougher to unit test the component.

jasonLaster:
R+ this is clearly a perf win and we can improve going forward.

public/js/actions/tests/newSource.js
jlongster: Yeah, wherever we throttle it'll make it harder to test. But if we move to just testing the "dumb" component, and `connect` implemented throttling then we would bypass it easily. I'm not exactly sure of this yet. Dispatching "raw" actions isn't that bad. This test is mainly testing the reducer, so it's not a big deal.

Not sure what the right solution is, so we'll see.

jlongster:
Thanks!

PR #236, opened 2016-06-02
add basic support for debugging node v7

jlongster:
Don't merge this until we get Greg's source work in because I had to modify how the source tree parses URLs. I will rebase after that.

This isn't extremely pretty, and I'm open to any suggestions for how to structure this. But with this, you can go the URL `http://localhost:8000/?ws=localhost:5858/node` which connects to the websocket that node v7 exposed when running with `node --inspect`. From there, it all works pretty well!


public/js/clients/chrome.js
```diff
@@ -138,6 +138,19 @@ function connectClient() {
...
+    bootstrap(InspectorBackend);
+    WebSocketConnection.Create(
+      websocketUrl,
+      conn => {
+        connection = conn;
+        resolve();
+      }
+    );
+  });
+}
```
jasonLaster: nice - i'm not sure if we discussed this, but I plan on taking another look at the chrome inspector backend that I added two weeks ago. At a minimum it will be moved into libs. Hopefully, I can replace it with something much simpler that maintains the same domain/dispatch api

public/js/util/sources-tree.js
```diff
@@ -1,6 +1,6 @@
 "use strict";
 
-const URL = require("url").parse;
+const URL = require("url");
```
jasonLaster: is this just for node, will it work in webpack?

public/js/clients/chrome.js
jlongster: Cool!

public/js/util/sources-tree.js
jlongster: Yep, webpack just bundles this library and uses it instead. I explicit to `URL.parse` to be clear that we aren't using the native one. This is important because they have different semantics: the native URL will throw for invalid URLs like `/foo/bar.js` (it's a path, not a URL, URLs need the protocol). But this library will not throw, it will return `{ protocol: null, pathname: "/foo/bar.js" }`.

I don't like that the library is not compliant with browsers, but if we need this to work in node, we need to depend on it.

jlongster:
@jasonLaster I reworked how the tabs list worked. The thing that triggered this was that I've noticed several times that the tabs list page does not scroll. This is due to it actually being rendered under a container which assumes it's the app, and thus sets flex and `overflow: hidden` properties. Having `Tabs` and `TabList` is confusing as well, as is having `TabList` be the root component.

This gets us closer to a separate toolbox component. Now the tabs list is its own page, and clicking on a tab literally navigates to a new page. This also means that you have your back button; just press back to go back to the tabs page.

It also removes the 2 places for connecting a tab: now, clicking on a tab does not have that logic at all. It just changes the URL. Connecting a tab only happens in a single place: at the bottom of `main.js`. 

The only thing left I'd like to change is that this still connects to Chrome and Firefox even if you already have a selected tab. We should add more checks in  `main.js` if `tab` exists in the query params and only connect to the needed target.

This way, the debugger becomes an isolated app that just takes targets through query params.

jlongster:
I went ahead and made it so that when you have a selected tab in the query params, it *only* connects to the target. So if you have a firefox tab, it only connects to Firefox. This will improve perf when refreshing the page if you have enabled chrome debugging. It also allows us to remove the hack where we need "fake" tabs in the chrome and firefox bundle environment; now we can just directly use the connection we already have.

jlongster:
@TatumCreative is going to keep working on the tree stuff, but take time to do it right. Let's not plan on having it in our demo. In that case, we can merge this and I'll help him rebase. I only refactored one function in `sources-tree.js` so it shouldn't be hard.

Jason I know you are out today and Monday but this touches the client stuff so it's important that you review it. I can wait.

public/js/components/Tabs.js
```diff
@@ -2,58 +2,67 @@
...
-    renderTabs("Chrome Tabs", chromeTabs, boundActions)
+  if (tabs.isEmpty()) {
+    return dom.div(
+      { className: "not-connected-message" },
+      "No remote tabs found. You may be looking to ",
+      dom.a({ href: "/?ws=localhost:5858/node" }, "connect to Node"),
+      "."
+    );
+  }
+
```
jasonLaster: Great.

jasonLaster:
Great work. I agree with your decisions on tablist. I was going to do something very similar. 

I especially like the back button support. Further separating the two experiences. 

Merging it as it's a large patch that will easily cause conflicts. Also adding node support is a great win. 

As a follow up item it'd be nice to have getting started docs for running / debugging node.

jlongster:
Docs would be great, yes! I'll file an issue about that, as we should do it soon. I meant to add a few screenshots of what this looks like:

screen shot 2016-06-03 at 10 26 25 pm

screen shot 2016-06-03 at 10 26 12 pm

Right now it just assumes you're running node already, which I think is fine for the demo.

I'm always impressed how fast you review stuff. :)

PR #238, opened 2016-06-03
make fromJS more loosely detect objects for cross-frame support

jlongster:
We have to do this when integrating into Firefox. The protocol gives us JS objects that were created in a different JS context, so `Object !== Object`. The previous method for detecting objects didn't pass, so they weren't being converted into immutable objects. This was breaking the frames list and the scope inspector.

The good thing is this will all go away once we have tcomb immutable types. It will know how to convert JS objects automatically and we can turn anything into typed objects without needing to check them.

### DIFF
--- a/public/js/util/fromJS.js
+++ b/public/js/util/fromJS.js
@@ -2,11 +2,6 @@
 
 const Immutable = require("immutable");
 
-function isPlainObj(value) {
-  return value && (value.constructor === Object ||
-                   value.constructor === undefined);
-}
-
 // When our app state is fully types, we should be able to get rid of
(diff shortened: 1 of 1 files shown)


jlongster:
With this in place, installing the bundle as the new debugger works pretty well:

screen shot 2016-06-03 at 2 53 53 pm

jasonLaster:
Nice job.

It'd be nice to add tests for `fromJS` when you have a chance.

jlongster:
I'm going to get rid of `fromJS` as quickly as I can by using tcomb, but probably not before the demo.

PR #241, opened 2016-06-06
Icon + source tree work

helenvholmes:
This PR:

- tightens the spacing between the debugger buttons
- adds the icon for the subsettings menu, although it doesn't do anything
- adds the cursor + hover state for the Sources view. Seems like there were some classes that were intended to do this (```.sources-list .source-item:hover```) but it didn't appear that they were actually affecting the Sources view in any way.


jlongster:
Awesome! There's a minor lint error, but I will merge and fix that on my own.

jasonLaster:
Thanks.  Those are carry offers from the source list days. We should probably make sure it's updated to source tree :)

helenvholmes:
Sorry for not linting, eep 😰

Are we using the same devtools linting? I should probably get those installed in Sublime (I run the ./mach  command in gecko-dev)

jlongster:
We are using the same rules, yes, although I think we had to tweak a few of them for some reason. Generally they are the same. You can run the linters with `npm run lint-js` and `npm run lint-css`. It's not a huge deal; our CI will catch it. I just wanted to get this in quickly and was easy for me to fix, but generally we just request the submitter to fix the errors before merging.

helenvholmes:
I figured. Awesome, thank you!

PR #248, opened 2016-06-07
Update stories + fixtures

jasonLaster:
Step 1 for adding component unit tests

clean up scopes, frames stories. They're now working again.

move pythagorean example into our tree so that we can automate making fixtures for it

extract integration test commands into debugger, todomvc, and commands into their own files. 

+ This makes the todomvc test file easier to read
+ it also lets us create a new fixtures.js test file and include todomvc examples there.
  + commands.js is for generic commands (navigate, debuggee). 
  + Todomvc (addTodo, toggleTodo). 
  + debugger (goToSource, toggleBreakpoint)


bin/fixtures.js
```diff
@@ -1,28 +1,45 @@
...
+function appendFixture(name, text) {
+  console.log("starting saveFixture")
+  const fixtureFile = getFixtureFile(name);
+  fs.appendFileSync(fixtureFile, text)
+  process.exit(0);
+}
+
+console.log("starting fixtures.js")
 const shouldSave = process.argv.indexOf("--save") > -1;
+const shouldAppend = process.argv.indexOf("--append") > -1;
```
jasonLaster: wrote the story in a comment below, but the fixture files are big enough to necessitate chunking them up and appending to the file

public/js/test/cypress/commands.js
```diff
@@ -1,39 +1,75 @@
...
+      }
+    });
+  });
+});
+
+Cypress.addParentCommand("navigate", function(url) {
+  return cy.window().then(win => {
+    return win.apiClient.navigate(url);
+  });
+})
```
jasonLaster: moving these commands out of todomvc.js

jlongster:
Neat, looks good to me! Although cypress is having some troubles on CircleCI.

jasonLaster:
Oops, the fixtures were running on CI.

By the way, i changed the way the fixtures were written from a node script with an exec to a test-server, that receives a post request. The server does not need the chunking and will be more general purpose.

PR #250, opened 2016-06-08
Add unit tests

jasonLaster:
This PR adds unit tests for the Scopes component. 
+ Tests live in components/tests
+ Tests are very similar to storybook (render / assert against dom)

Next Steps:
+ I added window typeof checks so that the tests could be no-op's in node. We should do something like jsdom or shallow-render to avoid this. 
+ The components use the store + connect to get the data, which make it hard to test the data dependencies. We might want to use the wrapped component directly.
+ The test file has a lot of extra code, some of which will be pulled into a test helper as we write more tests.

Unit Tests: 💃


public/js/test/fixtures/todomvc.json
```diff
@@ -0,0 +1,96 @@
+{
```
jasonLaster: I removed `todomvc.toggle` because it was very similar to `todomvc.updateOnEnter` and added `todomvc` which is just the app state when the debugger starts.

public/js/components/util/Arrow.js
```diff
@@ -2,8 +2,9 @@
 
 const { DOM: dom } = require("react");
 
-require("./Arrow.css");
-
+if (typeof window == "object") {
```
jlongster: I don't like having to guard against this every single time we require a CSS file. I'd rather just do this:

```js
require.extensions['.css'] = function() {
  return {};
}
```

where the test runner initializes node. This will make requiring the CSS file a noop.

jlongster:
Looks pretty good to me. I think we'll want something like [enzyme](https://github.com/airbnb/enzyme) pretty soon because shallow rendering is so useful, and always having to dig into the DOM is a little annoying. Also, enzyme allows you to get React-specific things like "make sure this component was passed these properties". They have [clear docs](https://github.com/airbnb/enzyme/blob/master/docs/guides/jsdom.md#using-enzyme-with-jsdom) for how to use jsdom.

I'd like to talk more about this in London, but this looks like a good starting point. 👍

public/js/components/util/Arrow.js
jasonLaster: great idea - this is especially useful because we'll never need css w/ node

PR #251, opened 2016-06-08
Update Unit Tests

jasonLaster:
+ Add frames unit tests
+ DRY up the unit tests / and stories with a new `renderComponentWithFixture` helper.

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -21,7 +21,7 @@ module.exports = function(config) {
     basePath: "",
     frameworks: ["mocha"],
     files: [
-      "./public/js/**/tests/*"
+      "./public/js/**/tests/*.js"
     ],
     exclude: [],
     reporters: ["progress"],
--- a/public/js/components/Frames.js
+++ b/public/js/components/Frames.js
@@ -10,7 +10,9 @@ const { endTruncateStr } = require("../util/utils");
 const { basename } = require("../util/path");
 const { getFrames, getSelectedFrame, getSource } = require("../selectors");
 
-require("./Frames.css");
+if (typeof window == "object") {
+  require("./Frames.css");
+}
 
 function renderFrameTitle(frame) {
   return div({ className: "title" }, endTruncateStr(frame.displayName, 40));
--- a/public/js/components/stories/Frames.js
+++ b/public/js/components/stories/Frames.js
@@ -1,46 +1,25 @@
 "use strict";
 
-const React = require("react");
-const { DOM: dom, createElement } = React;
-const { Provider } = require("react-redux");
-const { Map, fromJS } = require("immutable");
+const { DOM: dom, createElement, createFactory } = require("react");
+const { renderComponent, storiesOf } = require("./utils");
 
-const { createStore } = require("./utils");
+const Frames = require("../Frames");
 
-const { storiesOf } = require("@kadira/storybook");
-const Frames = React.createFactory(require("../Frames"));
-const fixtures = require("../../test/fixtures");
+const style = {
+  width: "300px",
+  margin: "auto",
+  paddingTop: "100px" };
 
-function getData(fixtureName) {
-  const fixture = fixtures[fixtureName];
-  const frames = fixture.pause.frames;
-  return {
-    pause: Map({ frames }),
-    sources: fromJS({ sources: fixture.sources.sources })
-  };
+const component = dom.div(
+  { className: "accordion" },
+  dom.div({ className: "_content" }, createElement(createFactory(Frames)))
+);
+
+function renderFrames(fixtureName) {
+  return renderComponent(component, fixtureName, { style });
 }
 
 storiesOf("Frames", module)
-  .add("Blank", () => {
-    return renderContainer(Frames);
-  })
-  .add("TodoMVC Todo Toggle", () => {
-    return renderContainer(Frames, getData("todomvcToggle"));
-  })
-  .add("Nested Closures", () => {
-    return renderContainer(Frames, getData("pythagorean"));
-  });
-
-function renderContainer(Component, data) {
-  const store = createStore(data);
-  return dom.div(
-    { style: {
-      width: "400px",
-      margin: "auto",
-      paddingTop: "100px"
-    }},
-    createElement(Provider, { store },
-      dom.div({ className: "accordion" },
-        dom.div({ className: "_content" }, createElement(Component))))
-  );
-}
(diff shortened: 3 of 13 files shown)


public/js/components/Frames.js
```diff
@@ -10,7 +10,9 @@ const { endTruncateStr } = require("../util/utils");
 const { basename } = require("../util/path");
 const { getFrames, getSelectedFrame, getSource } = require("../selectors");
 
-require("./Frames.css");
+if (typeof window == "object") {
+  require("./Frames.css");
+}
```
jasonLaster: I accidentally merged the last PR with touching `require`. I will do this soon. Probably, before this lands.

public/js/components/stories/Frames.js
```diff
@@ -1,46 +1,26 @@
...
-const { Map, fromJS } = require("immutable");
+const { DOM: dom, createElement, createFactory } = React;
+const { renderComponent, storiesOf } = require("./utils");
 
-const { createStore } = require("./utils");
+const Frames = require("../Frames");
 
-const { storiesOf } = require("@kadira/storybook");
-const Frames = React.createFactory(require("../Frames"));
-const fixtures = require("../../test/fixtures");
```
jasonLaster: really happy that this ball of dependencies is cleaned up considerably

public/js/components/stories/Frames.js
```diff
@@ -1,46 +1,26 @@
...
+  margin: "auto",
+  paddingTop: "100px" };
 
-function getData(fixtureName) {
-  const fixture = fixtures[fixtureName];
-  const frames = fixture.pause.frames;
-  return {
-    pause: Map({ frames }),
-    sources: fromJS({ sources: fixture.sources.sources })
-  };
```
jasonLaster: `getData` and `renderContainer` are both gone in favor of a shared `renderComponent` function

public/js/components/stories/Frames.js
```diff
@@ -1,46 +1,26 @@
...
-      paddingTop: "100px"
-    }},
-    createElement(Provider, { store },
-      dom.div({ className: "accordion" },
-        dom.div({ className: "_content" }, createElement(Component))))
-  );
-}
+  .add("Not Paused", () => renderFrames("todomvc"))
+  .add("Event Handler", () => renderFrames("todomvcUpdateOnEnter"))
+  .add("Nested Closures", () => renderFrames("pythagorean"));
```
jasonLaster: this is all we need for the new stories, which is nice on the eyes

public/js/components/tests/utils.js
```diff
@@ -0,0 +1,40 @@
...
+    attribute.nodeValue = "sandbox";
+    $el = document.createElement("div");
+    $el.setAttributeNode(attribute);;
+    document.body.appendChild($el);
+  }
+
+  $el.innerHTML = "";
+
+  return $el;
+}
```
jasonLaster: we need to create the sandbox element the first time because we don't have control over karma's runner `html` file

public/js/test/utils/renderComponentFromFixture.js
```diff
@@ -0,0 +1,42 @@
...
+  const data = getData(fixtureName);
+  const store = createStore(data);
+
+  return dom.div(
+    { style },
+    createElement( Provider, { store }, Component)
+  );
+}
+
+module.exports = renderComponentFromFixture;
```
jasonLaster: this is the new shared component that both the tests and stories share. At some point we might move off of this style in favor of bypassing the store, but for now it's easy to write the tests.

public/js/test/utils/renderComponentFromFixture.js
```diff
@@ -0,0 +1,42 @@
...
+  const fixture = fixtures[fixtureName];
+  return {
+    pause: Map({
+      pause: fromJS(fixture.pause.pause),
+      loadedObjects: fromJS(fixture.pause.loadedObjects),
+      frames: fixture.pause.frames
+    }),
+    sources: Map({
+      sources: fromJS(fixture.sources.sources)
+    })
```
jasonLaster: I see this data growing as we add more tests, but it should be consistent across tests as each field has a type

karma.conf.js
```diff
@@ -21,7 +21,7 @@ module.exports = function(config) {
     basePath: "",
     frameworks: ["mocha"],
     files: [
-      "./public/js/**/tests/*"
+      "./public/js/**/tests/*.js"
```
jasonLaster: I opted to move `utils` up to `test-utils` as changing all of the test files to `*.spec.js` seemed like over kill at this point. We can re-evaluate at a later point. Also, there is nothing intrinsically wrong with naming the file `test-utils.js`.

jasonLaster:
@jlongster r?

public/js/components/Frames.js
jlongster: Yeah, might be nice to do it in this PR if it's a quick fix. If it's not for some reason, you can look into it later.

public/js/components/stories/Frames.js
jlongster: This feels like cheating :) I was thinking that stories use the raw "dumb" component and pass things in as documentation for what props the component uses. It would also allow us to develop the components in isolation.

However, I get that many of our debugger-specific components heavily rely on the pause data specifically so it's nice to just load that in. I'll try to use storybook more, as I don't have a strong opinion about this now.

jlongster:
While this does reduce boilerplate, it makes things dependent on the store. I'm not against that right now, but stories don't really document components if they are written this way, and it might be harder to develop in isolation. I haven't used storybook enough to have a strong opinion though.

I think it's fine to merge, since I don't have any clear direction to suggest otherwise! ⭐

public/js/components/stories/Frames.js
jasonLaster: haha - yea, i had the same thought in richmond... but i'm now on the fence. We've discussed how we want to use connect with top-level debugger components (editor, sources, ...). 

I'm leaning towards using connect to pull in as many props as possible here. and then avoiding connect for lower-level components (breakpoint, tree, ...). This would line up nicely with storybook as well, where if we have a story for an Editor it can be rendered solely with the fixture data, wheras a Tree or Accordion story uses props...

I could be persuaded though

jasonLaster:
Yup. One last point here. There are plenty of potential components to document in storybook. Feel free to try other conventions. This is a good time to explore options.

PR #253, opened 2016-06-08
Add node-unit-tests setup file

jasonLaster:
+ Removes the need for guarding against css requires
+ also does the test globbing programatically, which will be more
flexible in the long run
+ also gives us space to add jsdom in the future
+ also, kinda sad we need to do any of this... but it's okay :)


jlongster:
Neat, and the output of the test runner is the same? Looks good to me! r+

jasonLaster:
Yup

PR #255, opened 2016-06-08
Add more integration tests

jasonLaster:
Adds several more steps to the integration tests:
+ enabling / disabling breakpoints
+ selecting a breakpoint (for source navigational purposes)
+ selecting a scope
+ step in, out, over ...

As I went, I refactored the debugger commands so that common selectors
were re-used to avoid painful test refactoring when the UI changes.

Writing the tests were actually pretty fun. Surprising/Unsuprisingly,
there were no hiccups along the way.

I cleaned up some of the CSS as I went so that the class names were
clearer.

### DIFF
--- a/public/js/actions/tests/loadSourceText.js
+++ b/public/js/actions/tests/loadSourceText.js
@@ -99,7 +99,7 @@ describe("loadSourceText", () => {
       store.dispatch(loadSourceText({ id: "foo1" }))
         .catch(() => deferred.resolve());
 
-      deferredMockThreadClient.getRequest().reject("poop");
+      deferredMockThreadClient.getRequest().reject("failed to load");
       return deferred.promise;
     }
 
@@ -108,7 +108,7 @@ describe("loadSourceText", () => {
     Task.spawn(function* () {
       yield loadBadSource(store);
       const fooSourceText = getSourceText(store.getState(), "foo1");
-      expect(fooSourceText.get("error")).to.equal("poop");
+      expect(fooSourceText.get("error")).to.equal("failed to load");
       done();
     });
   });
--- a/public/js/components/Accordion.js
+++ b/public/js/components/Accordion.js
@@ -37,7 +37,8 @@ const Accordion = React.createClass({
 
   renderContainer: function(item, i) {
     const { opened, created } = this.state;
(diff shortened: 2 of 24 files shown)


public/js/actions/tests/loadSourceText.js
```diff
@@ -108,7 +108,7 @@ describe("loadSourceText", () => {
     Task.spawn(function* () {
       yield loadBadSource(store);
       const fooSourceText = getSourceText(store.getState(), "foo1");
-      expect(fooSourceText.get("error")).to.equal("poop");
+      expect(fooSourceText.get("error")).to.equal("failed to load");
```
jasonLaster: got rid of the bad joke

public/js/components/Accordion.js
```diff
@@ -37,7 +37,8 @@ const Accordion = React.createClass({
 
   renderContainer: function(item, i) {
     const { opened, created } = this.state;
-    const containerClassName = item.header.toLowerCase().replace(/\s/g, "-");
+    const containerClassName =
+          item.header.toLowerCase().replace(/\s/g, "-") + "-pane";
```
jasonLaster: the `-pane` disambiguates the CSS from other class names

public/js/test/utils/debuggee.js
```diff
@@ -106,7 +106,7 @@ function Debuggee() {
   };
 }
 
-const debuggeeStatement = `window.Debuggee = (${Debuggee})()`;
+const debuggeeStatement = `window.dbg = (${Debuggee})()`;
```
jasonLaster: changed `Debuggee` to `dbg` as it's a little easier to type

jasonLaster:
@jlongster r? 

mostly cleanup from yesterday as I was doing a pass through the tests

jasonLaster:
I'm noticing some issues with cypress on circle. It looks fairly consistent recently. The error logs indicate that it is probably Firefox starting that is causing the issue. I'll investigate a bit more next week.

jlongster:
Looks fine to me. We probably should hold off on writing a lot more tests until I implement immutable tcomb types and type the app state. I feel like that may impact how we are doing things like hyrdating the store, but maybe not.

Merging as you are traveling today and since you asked for review I'm assuming it's ready

PR #256, opened 2016-06-14
Extract chrome-remote-debug-protocol

jasonLaster:
+ Moved it into libs
+ removed unecesary connectX function
+ cleaned up some code
+ added command and event files

changed some names:
+ currentX names like currentThreadClient became threadClient
+ apiClient became client


public/js/test/cypress/commands/debugger.js
```diff
@@ -100,6 +100,7 @@ function selectScope(index) {
 }
 
 function resume() {
+  cy.wait(200);
```
jasonLaster: clicking resume is flakey because it can quickly toggle states. this is a hack for now, that i will investigate later.

public/js/components/tests/Editor.js
```diff
@@ -1,5 +1,7 @@
 "use strict";
 
+if (typeof window == "object") {
```
jasonLaster: moved the condition up so that the requires are avoided as well.
This was the greg made me do it moment

PR #259, opened 2016-06-16
Fix node

jasonLaster:
node was failing because it was missing connectWebsocket, I added `connectNode` which is similar to `connectNode`

### DIFF
--- a/public/js/clients/index.js
+++ b/public/js/clients/index.js
@@ -13,6 +13,15 @@ function getClient() {
   return firefox.clientCommands;
 }
 
+function startDebugging(connTarget, actions) {
+  if (connTarget.type === "node") {
+    return startDebuggingNode(connTarget.param, actions);
+  }
+
+  const target = connTarget.type === "chrome" ? chrome : firefox;
+  return startDebuggingTab(target, connTarget.param, actions);
+}
+
 function startDebuggingNode(url, actions) {
   clientType = "chrome";
   return chrome.connectNode(`ws://${url}`).then(() => {
@@ -45,6 +54,5 @@ function connectClients() {
 module.exports = {
   getClient,
   connectClients,
-  startDebuggingTab,
-  startDebuggingNode
+  startDebugging
 };
--- a/public/js/clients/index.js
+++ b/public/js/clients/index.js
@@ -13,7 +13,14 @@ function getClient() {
   return firefox.clientCommands;
 }
 
-function startDebugging(targetEnv, tabId, actions) {
+function startDebuggingNode(url, actions) {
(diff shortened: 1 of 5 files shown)


public/js/clients/index.js
```diff
@@ -13,7 +13,23 @@ function getClient() {
...
+function startDebugging(connTarget, actions) {
+  if (connTarget.type === "node") {
+    return startDebuggingNode(connTarget.param, actions);
+  }
+
+  const target = connTarget.type === "chrome" ? chrome : firefox;
+  return startDebuggingTab(target, connTarget.param, actions);
+}
+
+function startDebuggingNode(url, actions) {
```
jasonLaster: I added `startDebuggingNode` to be parallel with `startDebuggingTab`. Also because `clientType` needs to be set for the actions to use the right client.

public/js/components/Tabs.js
```diff
@@ -56,7 +56,7 @@ function Tabs({ tabs }) {
     dom.div(
       { className: "node-message" },
       "You can also ",
-      dom.a({ href: "/?ws=localhost:5858/node" },
+      dom.a({ href: "/?ws=localhost:9229/node" },
```
jasonLaster: I'm not sure why 9229 is used for me... This is quite troubling

jlongster: I'll pull down the latest version and see if they changed the port. I bet they did.

jlongster:
Why is karma broken? I can't remember if that's something we talked about, but I don't remember specifically.

jasonLaster:
Karma is failing for a similar webpack issue. It's tied to the devtools-sham change. Should be an easy fix. I filled a big for it too

jasonLaster:
Feel free to merge this. I'm on my phone and can't with the karma  fail.

public/js/clients/index.js
```diff
@@ -13,7 +13,23 @@ function getClient() {
...
+  clientType = "chrome";
+  return chrome.connectNode(`ws://${url}`).then(() => {
```
jlongster: I don't love the name `connectNode`. This isn't specific to node at all, it's just passing the websocket URL directly. It could be anything. People may use this API to build a different toolbox, or we might eventually have a websocket from Firefox to pass in. I'd rather it be something like `connectWebsocket`.

jlongster:
I actually like that you can see everything it connects to right in `main.js`. To me, that's sort of the point of that bootstrapping file. But we're going to keep thinking about this as we deal with more types of connections (i.e. workers) so I'm fine going this way and seeing how it feels.

PR #264, opened 2016-06-21
Tabs

jasonLaster:
This adds source tabs behind a feature flag. 

This handles several situations like:
+ adding sources
+ removing sources to the left and right of the selected source
+ hiding tabs that wrap row

This does not handle:
+ showing tabs that are hidden in a dropdown
+ draggable tabs
  
Interesting observations:
+ Storybook + Unit Tests let me TDD much of it(was natural)
+ Using a feature flag with tests, we have less pressure to land the feature immediately


mocha-runner.html
```diff
@@ -2,14 +2,14 @@
...
+  <link href="mocha/mocha.css" rel="stylesheet" />
 </head>
 <body>
   <div id="mocha"></div>
 
   <script src="https://cdn.rawgit.com/jquery/jquery/2.1.4/dist/jquery.min.js"></script>
-  <script src="https://cdn.rawgit.com/Automattic/expect.js/0.3.1/index.js"></script>
-  <script src="https://cdn.rawgit.com/mochajs/mocha/2.2.5/mocha.js"></script>
+  <script src="expect.js/index.js"></script>
+  <script src="mocha/mocha.js"></script>
```
jasonLaster: I noticed on the flight that the mocha server fails without internet connection. This is why that demo on thursday failed.

this fixes it by having the mocha server host the node modules needed. jQuery is ignored because it's probably not necessary although i'm not 100% sure yet.

public/js/actions/sources.js
```diff
@@ -43,7 +43,7 @@ function newSource(source) {
   };
 }
 
-function selectSource(id, opts) {
```
jasonLaster: opts was unused

public/js/test/fixtures/pythagorean.json
```diff
@@ -6,169 +6,93 @@
   },
   "sources": {
```
jasonLaster: updating the fixtures because the tabs field is now needed.

public/js/configs/feature.js
```diff
@@ -2,23 +2,19 @@
...
-try {
-  localDevelopmentConfig = require("./development.local");
-} catch (e) {
-  localDevelopmentConfig = {};
-}
-
-config = Object.assign({}, developmentConfig, localDevelopmentConfig);
+const developmentConfig = require("./development.json");
+config = Object.assign({}, developmentConfig);
+const originalConfig = Object.assign({}, developmentConfig);
```
jasonLaster: I dropped the local dev config file as we'll be going to a new option soon and it broke storybook

jasonLaster:
Merged so Helen can polish.

jasonLaster:
@jlongster please leave comments on this PR and I'll update in a subsequent PR

public/js/reducers/sources.js
```diff
@@ -28,9 +29,14 @@ function update(state = initialState, action) {
       );
 
     case constants.SELECT_SOURCE:
-      return state.merge({
-        selectedSource: action.source
-      });
+      return state
+        .merge({ selectedSource: action.source })
+        .set("tabs", addSourceToTabList(state, fromJS(action.source)));
```
jlongster: why not `state.merge({ selectedSource: action.source, tabs: addSourceToTabList(...) })` ? You can do it in the same `merge` rather than another `set`.

public/js/reducers/sources.js
```diff
@@ -28,9 +29,14 @@ function update(state = initialState, action) {
...
+
+    case constants.CLOSE_TAB:
+      return state
+        .set("tabs", removeSourceFromTabList(state, action.id))
+        .merge({ selectedSource: getNewSelectedSource(state, action.id) });
```
jlongster: Same here

public/js/reducers/sources.js
jasonLaster: I avoided it because tabs is a list, but maybe that would work...

PR #266, opened 2016-06-21
Update dependencies

jasonLaster:
This PR does two things:
1. moves deps to dev-deps
2. updates versions

it's two commits so its easier to review


jlongster:
Cool, looks good! r+

PR #267, opened 2016-06-21
Format values with Reps

jasonLaster:
This is a first pass of formatting object values with Reps

screen shot 2016-06-21 at 5 32 08 pm

### DIFF
--- a/public/js/components/Rep.js
+++ b/public/js/components/Rep.js
@@ -1,0 +1,14 @@
+"use strict";
+
+const React = require("react");
+const Rep = React.createFactory(
+  require("../lib/devtools/client/shared/components/reps/rep").Rep);
+const Grip = require("../lib/devtools/client/shared/components/reps/grip").Grip;
+
+require("../lib/devtools/client/shared/components/reps/reps.css");
+
+function renderRep({ object }) {
+  return Rep({ object, defaultRep: Grip });
+}
+
+module.exports = renderRep;
--- a/public/js/components/Scopes.js
+++ b/public/js/components/Scopes.js
@@ -11,9 +11,22 @@ const { getPause, getLoadedObjects } = require("../selectors");
 const { DOM: dom, PropTypes } = React;
 const ManagedTree = React.createFactory(require("./util/ManagedTree"));
 const Arrow = React.createFactory(require("./util/Arrow"));
+const Rep = require("./Rep");
+
+require("./Scopes.css");
 
 function info(text) {
   return dom.div({ className: "pane-info" }, text);
+}
+
+function getBindingVariables(bindings) {
+  return bindings
+    .get("arguments").map(arg => arg.entrySeq().get(0))
+    .concat(bindings.get("variables").entrySeq())
+    .filter(binding => (!binding[1].hasIn(["value", "missingArguments"]) &&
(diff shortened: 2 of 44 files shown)


public/js/components/Scopes.js
```diff
@@ -33,13 +44,7 @@ function getScopes(pauseInfo) {
 
     if (type === "function" || type === "block") {
       const bindings = scope.get("bindings");
-      const vars = bindings
-        .get("arguments").map(arg => arg.entrySeq().get(0))
-        .concat(bindings.get("variables").entrySeq())
-        .filter(binding => (!binding[1].hasIn(["value", "missingArguments"]) &&
-                            !binding[1].hasIn(["value", "optimizedOut"])))
-        .sort((a, b) => a[0].localeCompare(b[0]))
-        .toArray();
```
jasonLaster: small extract. I assume this code will be cleaned up going forward

public/js/components/Rep.js
```diff
@@ -0,0 +1,13 @@
...
+// Node does not support the AMD syntax used in Rep
+// Node also does not support Worker, which is an issue when I built the Rep with webpack
+if (typeof window == "object") {
+    const Rep = require("../lib/devtools/client/shared/components/reps/Rep").Rep;
+    require("../lib/devtools/client/shared/components/reps/reps.css");
+
+    module.exports = Rep;
+} else {
+    module.exports = function() {};
+}
```
jasonLaster: The unit tests fail in Node for these two reasons. @janodvarko do you have recommendations on what we can do? It seems like the easiest thing to do is to add a typeof check for Worker and then build Rep. The ideal solution would be to convert Rep to commonjs, but I'm not sure if that would be possible...

janodvarko: > add a typeof check for Worker and then build Rep.

Hmm, I don't have a better tip, although could you somehow define the `define` function in the global scope?

> The ideal solution would be to convert Rep to commonjs, but I'm not sure if that would be possible...

The problem is that Reps are consumed by the JSON viewer. JSON Viewer runs in a page and needs a way to load all the modules (async). As soon as we allow using WebPack I can build jsonviewer.bundle.js (just like you do for the debugger.html), include it in JSON viewer page and we can remove the AMD support.

Honza

public/js/test/node-unit-tests.js
```diff
@@ -3,6 +3,8 @@
 const glob = require("glob").sync;
 const path = require("path");
 
+require("amd-loader");
```
jasonLaster: this lets us load the amd rep files. there's magic in there for defining a global define function and doing a naive map to require. it's cool, but i would not trust it unless it was for tests.

jlongster: Neat. At this point, I wouldn't be surprised if we ended up using webpack for devtools generally so when that happens the JSON viewer can use that and we can avoid AMD completely.

jlongster:
I'm going to change how we handle the tree structure completely, so yeah that code was prototype-quality.

Nothing stands out to me, except maybe we could avoid committing the React file in vendor since we don't use it and to avoid committing a big file? Otherwise r+

jasonLaster:
Agreed, i'll try and drop that.

PR #268, opened 2016-06-22
Replace some shams

jasonLaster:
+ Added task to util. There wasn't a task corrollary in central.
+ Added promise middleware to util. I wanted to add it to devtools, but
it still relies on promise as opposed to defer, because we haven't
converted those modules yet
+ Created util/sources for isJavaScript. I considered copying
debugger/utils into devtools, but ran into two issues (L10N, and
exports).


public/js/lib/devtools-sham/shared/DevToolsUtils.js
```diff
@@ -92,74 +92,6 @@ exports.makeInfallible = function makeInfallible(aHandler, aName) {
     }
   }
 }
-/**
```
jasonLaster: moved these functions to `util/utils` as there wasn't a similar function in `DevToolsUtils` also, they're not devtool specific

public/js/lib/devtools/client/shared/redux/middleware/log.js
```diff
@@ -9,12 +9,7 @@
  */
 function log({ dispatch, getState }) {
   return next => action => {
-    const str = "[DISPATCH] " + JSON.stringify(action, null, 2);
-    if (typeof console !== "undefined") {
-      console.log(str)
-    } else {
-      dump(str + "\n");
-    }
+    console.log("[DISPATCH]", JSON.stringify(action, null, 2));
```
jasonLaster: this change probably reflects what is in central

public/js/test/utils/renderComponentFromFixture.js
```diff
@@ -7,7 +7,7 @@ const { fromJS, Map } = require("immutable");
 const { combineReducers } = require("redux");
 
 const fixtures = require("../fixtures");
-const configureStore = require("../../create-store");
+const configureStore = require("../../util/create-store");
```
jasonLaster: moved create-store into util to reduce the top level files. also seemed closer to the middleware

public/js/util/DevToolsUtils.js
```diff
@@ -0,0 +1,13 @@
+function reportException(who, exception) {
+  let msg = who + " threw an exception: ";
+  console.error(msg, exception);
+}
```
jasonLaster: moved reportException over as I'd like to be able to control how exceptions are reported. This will need to be flagged off in production, which I'll do when i work on configs next.

public/js/util/DevToolsUtils.js
```diff
@@ -0,0 +1,13 @@
...
+
+function executeSoon(aFn) {
+  setTimeout(aFn, 0);
+};
```
jasonLaster: moved executeSoon over as I'm not sure if the `Services.tm` tick logic can be duplicated in a de-priveledged way. Also, this is fine for our purposes.

public/js/util/create-store.js
```diff
@@ -5,11 +5,11 @@
...
-const { thunk } = require("./util/thunk-middleware");
-const { waitUntilService } = require("devtools-sham/client/shared/redux/middleware/wait-service");
-const { log } = require("devtools-sham/client/shared/redux/middleware/log");
-const { promise } = require("devtools-sham/client/shared/redux/middleware/promise");
-const { history } = require("devtools-sham/client/shared/redux/middleware/history");
+const { waitUntilService } = require("devtools/client/shared/redux/middleware/wait-service");
+const { log } = require("devtools/client/shared/redux/middleware/log");
+const { history } = require("./redux/middleware/history");
+const { promise } = require("./redux/middleware/promise");
+const { thunk } = require("./redux/middleware/thunk");
```
jasonLaster: it's frustrating that the middleware is split for now, but that should consolidate eventually.

public/js/lib/devtools-sham/client/shared/source-utils.js
```diff
@@ -1,293 +0,0 @@
-/* This Source Code Form is subject to the terms of the Mozilla Public
```
jasonLaster: it's nice to be able to kill a lot of unused coode

jasonLaster:
It's safe to ignore karma right now, it's passing locally... i'm betting it's a chrome 47 thing

public/js/test/utils/renderComponentFromFixture.js
jlongster: I don't mind it at the top-level because it's a core part of the initialization process, but it makes sense as a util too.

public/js/util/DevToolsUtils.js
jlongster: I've never loved `reportException` anyway, and I'm not sure we even need it. Most errors like this (usually used at the end of a promise chain) could be automatically caught in the promise middleware if we want to do something special with them.

jlongster:
I'm a little confused about something: modules in devtools-sham that depend on other devtools-sham modules. For example, `public/js/lib/devtools-sham/transport/transport.js` has all of these:

```
const DevToolsUtils = require("devtools-sham/shared/DevToolsUtils");
const { dumpn, dumpv } = DevToolsUtils;
const StreamUtils = require("devtools-sham/shared/transport/stream-utils");
const { Packet, JSONPacket, BulkPacket } =
  require("devtools-sham/shared/transport/packets");
const promise = require("devtools-sham/sham/promise");
const EventEmitter = require("devtools-sham/shared/event-emitter");
```

and yet I don't see those in your list and I don't see any changes in transport.js.

jlongster:
I think I understand now, it's not that you got rid of files like `DevToolsUtils` you just split them. So now there are two files, `public/js/util/DevToolsUtils.js` and the one in devtools-sham. That seems a little confusing... I'd rather make changes like this on m-c first and update our `devtools` files accordingly and start using them. But if you're willing to help pull in changes from m-c (I'm about to work in DevToolsUtils) and consolidate all of this, it's fine.

jasonLaster:
@jlongster I'm fine with being an intermediate step. Hopefully, we can have a DevToolsUtils (or something similar) in m-c use.

I'd like to merge this in as it seems like a helpful step forward.

jlongster:
Not sure what you mean by "hopefully we can have a DTU in m-c", as there already is one and we just need to de-chromify it. Once we do that, we should be able to pull it in directly and use it, and we shouldn't need anything in our `utils`, which is why this felt like an extra step that will go away soon. I guess it just feels weird to call our local file `DevToolsUtils`, when we already have `utils.js`, so now we have 2 util files. I'd rather just add the functions to our local `utils.js` and over time we'll pull more and more from the standardized `DevToolsUtils`.

jasonLaster:
@jlongster summarizing what was discussed earlier -

utils/DevTools (reportException) utils/utills (zip, entries) should be temporary.

PR #269, opened 2016-06-22
Replace some shams

jasonLaster:
+ Added task to util. There wasn't a task corrollary in central.
+ Added promise middleware to util. I wanted to add it to devtools, but
it still relies on promise as opposed to defer, because we haven't
converted those modules yet
+ Created util/sources for isJavaScript. I considered copying
debugger/utils into devtools, but ran into two issues (L10N, and
exports).
+ Move the redux middleware to util/redux and lib/devtools. The middleware that was a straight copy went to lib/devtools.
+ Move task to util because there is not a similar file in central


public/js/actions/sources.js
```diff
@@ -4,9 +4,9 @@
 "use strict";
 
 const defer = require("devtools/shared/defer");
-const { PROMISE } = require("devtools-sham/client/shared/redux/middleware/promise");
-const { Task } = require("devtools-sham/sham/task");
-const SourceUtils = require("devtools-sham/client/shared/source-utils");
+const { PROMISE } = require("../util/redux/middleware/promise");
+const { Task } = require("../util/task");
+const { isJavaScript } = require("../util/source");
```
jasonLaster: There are obviously other SourceUtils we will need soon. If SourceUtils is de-priveledged we can use that at somepoint.

PR #270, opened 2016-06-23
Organize configs

jasonLaster:
As a first step to switching towards a tool like nconfig,
this patch organizes the config data in feature.js

There are a couple other cleanups in here, but they're relatively minor

### DIFF
--- a/bin/development-server.js
+++ b/bin/development-server.js
@@ -1,6 +1,8 @@
 #!/usr/bin/env node
 "use strict";
 
+require("babel-register");
+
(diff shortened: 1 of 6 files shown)


bin/development-server.js
```diff
@@ -1,6 +1,8 @@
 #!/usr/bin/env node
 "use strict";
 
+require("babel-register");
```
jasonLaster: now that feature.js references `devtools-sham` we need babel-register to do the lookup. This hopefully will go away soon though

PR #272, opened 2016-06-24
Update Scopes

jasonLaster:
+ Add formatting for selected elements
+ Replace object rep w/ "Object"
+ sort properties.
+ hide scope values

Before
screen shot 2016-06-24 at 2 54 30 pm

After
screen shot 2016-06-24 at 2 54 13 pm

Firefox
screen shot 2016-06-24 at 2 55 58 pm

Chrome
screen shot 2016-06-24 at 2 55 36 pm

### DIFF
--- a/public/js/components/Scopes.js
+++ b/public/js/components/Scopes.js
@@ -40,6 +40,7 @@ function getScopes(pauseInfo) {
   }
 
   const scopes = [];
+  let index = 0;
 
   do {
     const type = scope.get("type");
@@ -63,6 +64,7 @@ function getScopes(pauseInfo) {
               type: "object",
               actor: Math.random().toString()
             }),
+            isScope: true,
             prefetchedProperties: vars
           }),
         ]);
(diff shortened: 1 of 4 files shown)


public/js/components/Scopes.js
```diff
@@ -71,15 +73,99 @@ function getScopes(pauseInfo) {
...
+function loadedProperties(obj, loadedObjects) {
+  return loadedObjects.get(obj.getIn(["value", "actor"])) ||
+         obj.get("prefetchedProperties");
+}
+
+function sortProperties(a, b) {
+  const obj = isObject(a[1]) && !isObject(b[1]) ? -10 : 10;
+  const alpha = a[0] < b[0] ? -1 : 1
+
+  return obj + alpha > 0 ? -1 : 1;
```
jasonLaster: I pushed the objects to the bottom of the list because aesthetically it looked better to me, but happy to revert that that change if it seems weird.

screen shot 2016-06-24 at 3 02 03 pm
screen shot 2016-06-24 at 3 01 52 pm

public/js/components/Scopes.js
```diff
@@ -71,15 +73,99 @@ function getScopes(pauseInfo) {
...
+      onClick: e => {
+        e.stopPropagation();
+        setExpanded(item, !expanded);
+      }
+    }),
+    dom.span({ className: "scope-object-label" }, item[0]),
+    ...runderItemValue(obj)
+  );
+}
+
```
jasonLaster: I moved these functions out of the class as they didn't depend on props. I believe you have an idea for how you want to organize Scopes and I don't want to get in your way. Feel free to change this up.

public/js/components/Scopes.js
jlongster: I think it look better with the way that we indent the tree, but I would prefer sorting them alphabetically. As a user, it's confusing if I'm looking for a property and have to remember if it's an object or not.

What we should fix is how we display the tree so that they don't look so out of place. It's problematic anyway because it's unclear which items are children; for example, in the above screenshot, is `a: 2` a top-level item or a child of `pythagorean`? It lines up perfectly with the other top-level items, so it looks like it's top-level. But it's actually a child of `pythagorean`. The way to fix this is to indent all the children more and make them all line up (whether of not they have arrows). If you look at Firefox/Chrome, that's exactly what they do.

So I'd prefer to revert this change and I can work on styling them. If you're not working on this today/tomorrow (because of conference) I'll merge this in and do a follow-up. I'll work on scopes later this week.

public/js/components/Scopes.js
jlongster: Personally I like things either to be fully createClass components or not. We have eslint rules to force a specific ordering of how methods are declared in components so there is a consistent ordering of them (lifecycle methods first, render methods last, etc). If we pull them out we lose this, and it splits up the component. React will eventually support stateful function components, and we'll always use top-level functions then.

public/js/components/Scopes.js
```diff
@@ -71,15 +73,99 @@ function getScopes(pauseInfo) {
       scopes.push([
         scope.getIn(["object", "class"]),
         fromJS({
+          isScope: true,
           value: scope.get("object")
         })
       ]);
     }
+    index++;
```
jlongster: I don't actually see index used anywhere

public/js/components/Scopes.js
```diff
@@ -99,74 +185,27 @@ const Scopes = React.createClass({
...
-    const { pauseInfo, loadObjectProperties } = this.props;
+    const { pauseInfo, loadObjectProperties, loadedObjects } = this.props;
     const { scopes } = this.state;
 
     const tree = ManagedTree({
       itemHeight: 100,
+
       getParent: item => null,
-      getChildren: this.getChildren,
+      getChildren: (item) => getChildren(item, loadedObjects),
```
jlongster: Nit: no need for parens for single args (`item => ...`)

jlongster:
This also has linting errors. I will be working on scopes today and heavily refactoring this code (for better tree handling so we might need to just close this PR). Let me see what my work looks like though.

public/js/components/Scopes.js
jasonLaster: Sounds good. I felt bad about breaking the alphabetical sorting. Merge and improve

public/js/components/Scopes.js
jasonLaster: Sounds good. I was on the fence here. As a rule of thumb, I like utility functions to be outside of the class it becomes clear it has no dependencies and can be extracted, but if it is a component function them it should be in the class. Rendering a tree node does feel related to the tree.

public/js/components/Scopes.js
jasonLaster: Agreed.

jasonLaster:
I'll be on PTO till Tuesday. So do whatever feels right. 

Glad the scopes page is getting some love

jlongster:
All of this code was refactored in #288, with any feature in here integrated there, sorry!

PR #275, opened 2016-06-29
add minimist to dependencies

amitzur:
Hi @jasonLaster I cloned the repo and had to add this dependency for `npm start` to work.
After that it works like a charm

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
     "codemirror": "^5.1.0",
     "express": "^4.13.4",
     "immutable": "^3.7.6",
+    "minimist": "^1.2.0",
     "react": "=0.14.7",
     "react-dom": "=0.14.7",
     "react-immutable-proptypes": "^1.7.1",


jlongster:
Can you change it to a dev dependency? It's only needed when developing. I bet this got dropped when the dependencies were restructured.

jlongster:
I bet I know what happened: you might be using an older version of node that doesn't flatten dependencies. Looks like minimist is installed for me, but it comes from another dependency of mine. Later versions of node flatten dependencies and stuff like this happens.

I'll move it to dev deps and go ahead and merge this, thanks!

amitzur:
You mean an older version of npm, yes. I was on npm 2 and was planning to upgrade to 3 for some time. I finally did it now and verified that it works without the dependency. But I agree it's good to have it for the sake of npm 2 users. 

BTW, with npm 2:
```
$ du -sh node_modules
776M	node_modules
```
and with npm 3:
```
$ du -sh node_modules
199M	node_modules
```

jasonLaster:
Haha wow, i guess top level dependencies are better for my computer. I still hate we have like 800 deps :/

So many possible left pads

PR #277, opened 2016-06-29
enable hot reloading via react-hot-loader

jlongster:
This isn't too bad. The main thing is to use `react-hot-middleware` as well, and it seems to work better than I remember.

I'm using the current version of react-hot-loader that will be deprecated at some point. It doesn't work with function components, but we don't have many of them. This is at least something. I was able to work on the scope inspector without losing any state.

Note that react-storybook's hot reloading isn't that great. I looked into it, and they implement their own react hot reloading. I'm not sure why. But it's very naive and basically rerenders a new component when its source changes. That means it doesn't keep local state. If you hot reload one of our trees, for example, it will lose all the expanded state.

We can simply upgrade to v3 of react-hot-loader when it's stable. I keep hitting problems with it, so for now let's go with this.

It's not on by default (there is a perf hit to the webpack compiler). You can turn it on by adding `hotReloading: true` to the config.


public/js/main.js
```diff
@@ -57,18 +64,11 @@ if (connTarget) {
...
-    const mount = document.querySelector("#mount");
-
-    // bail in test environments that do not have a mount
-    if (!mount) {
-      return;
-    }
-
     boundActions.newTabs(tabs);
-    renderRoot(store, Tabs, mount);
+    renderRoot(Tabs);
```
jlongster: I didn't see the point of passing those things in, and when we move to v3 of react-hot-loader it'll make it easier (we can just call renderRoot again).

jasonLaster:
:thumbsup:

jlongster:
Not totally related to this, but: I know I've changed a few things that back from what you've changed, so I'm sorry about that. When you get back we'll talk about the config system and reconcile what we're thinking. I like how you are using it for feature flags. Just need to figure out how config works on server & client and how we'll integrate with firefox.

PR #280, opened 2016-06-29
bring back local config and tweak how it loads

jlongster:


jlongster:
I need this to turn on the new hot module reloading feature, so I worked on the config for a bit :p

bin/development-server.js
```diff
@@ -4,14 +4,27 @@
...
+// Improve the first run experience: copy the local sample config to a
+// local file so webpack can find it. We don't have to do this, but
+// otherwise webpack prints a warning that scares people.
+const localConfigPath = path.join(__dirname, "../config/local.json")
+if(!fs.existsSync(localConfigPath)) {
+  const defaultConfig = fs.readFileSync(
+    path.join(__dirname, "../config/local.sample.json")
+  );
+  fs.writeFileSync(localConfigPath, defaultConfig);
+}
```
jlongster: I thought of a few different ways to handle config, but couldn't come up with a great solution quickly that works both in client and server. In Node, we could simply check if the local config file exists and if not, don't load it in. But we can't do that in a browser.

We should talk through config soon because I think client config is very different from our server config. The client should ideally take preferences from the server, because in the Firefox toolbox we'll want to give the client preferences from the toolbox.

jlongster:
I have no idea what's going on with Cypress. It's acting as if the development server isn't even running, and I cannot reproduce it locally. It was errors out with "can't reach the page at localhost:8000", and I've been trying to fix it for a while.

In order to make progress in the next few days that Jason is gone, I disabled cypress. We don't have that many integration tests yet anyway, and I don't think it's a crucial part of our workflow yet.

I'll work with Jason when he gets back to fix it.

jasonLaster:
No worries. Happy to investigate on Tuesday. The easiest place to debug is in circle with rebuild w SSH. You can look at the circle config to see what is being executed as well

jlongster:
Oh neat, thanks! I immediately found the problem: I was using destructuring and development-server.js is not compiled by babel so it was failing on circle's version of node 4... Yeesh, totally my fault. Fixing issue https://github.com/jlongster/debugger.html/issues/284 should fix this.

We should be able to compile this file, but we can figure that out later.

PR #282, opened 2016-06-29
separate Sources and SourcesTree component to fix hot reloading

jlongster:


jasonLaster:
Nice cleanup. 

Sources+SourcesTree looks nice. 

What's the benefit of adding actions to window? It's that for hot-reloading as well. It could also help with integration tests I suppose as well.

jlongster:
> What's the benefit of adding actions to window? It's that for hot-reloading as well. It could also help with integration tests I suppose as well.

Good catch! I was down a deep rabbit hole trying to figure out why hot reloading wasn't working correctly, so I exposed it so I could manually fire actions. I agree that we should think about ways to inject things like that. We already have a function that injects something for tests, we could make that expose things like actions to the window as well.

We might need to do that for mochitests, actually... they won't have direct access to the modules, as far as I know. But we'll talk about this more when you get back.

I'll remove that code for now.

PR #283, opened 2016-06-30
fix development-server.js to work on node <6 (fixes cypress tests)

jlongster:
Just doing this to get a failed circle run so I can investigate it


jlongster:
Wellp, I'm an idiot and didn't realize #284 made it so the site wouldn't even run on Circle, thus integration tests were failing.

jasonLaster:
Yep, that would do it :)

You're doing great dev ops work this week. Bundle+task runner, then hot reloading!

PR #287, opened 2016-07-01
Style disabled breakpoints in the list

clarkbw:
This creates style classes for breakpoints so I could make the font colour lighter to match the disabled state.  Needs @helenvholmes review.
I also added an option for the paused state so we could style that separately as well but couldn't find a style that I liked.  Maybe @helenvholmes has ideas for this or it can remain the same.  I used CSS animations to make the breakpoint "flash" a background color which worked pretty well but needs more thought into the timing because it distracted from the editor text location.

@jasonLaster I changed the storybook to be 3 breakpoints from 1 so I could have one of each type.  The last, disabled, breakpoint disappears when you check the box and I wasn't sure how to handle this.  I assuming it is the click handler running and then finding that breakpoint doesn't exist.  But I could use some help here.

### DIFF
--- a/public/js/components/Breakpoints.js
+++ b/public/js/components/Breakpoints.js
@@ -5,6 +5,7 @@ const { connect } = require("react-redux");
 const { bindActionCreators } = require("redux");
 const ImPropTypes = require("react-immutable-proptypes");
 const Isvg = React.createFactory(require("react-inlinesvg"));
+const classnames = require("classnames");
 const actions = require("../actions");
 const { getSource, getPause, getBreakpoints, makeLocationId } = require("../selectors");
 const { truncateStr } = require("../util/utils");
@@ -58,6 +59,7 @@ const Breakpoints = React.createClass({
     const locationId = breakpoint.get("locationId");
     const line = breakpoint.getIn(["location", "line"]);
     const isCurrentlyPaused = breakpoint.get("isCurrentlyPaused");
+    const isDisabled = breakpoint.get("disabled");
 
     const isPausedIcon = isCurrentlyPaused && Isvg({
       className: "pause-indicator",
@@ -66,14 +68,17 @@ const Breakpoints = React.createClass({
 
     return dom.div(
       {
-        className: "breakpoint",
+        className: classnames({
+          breakpoint: true,
+          disabled: isDisabled
+        }),
         key: locationId,
         onClick: () => this.selectBreakpoint(breakpoint)
       },
       dom.input(
         {
           type: "checkbox",
-          checked: !breakpoint.get("disabled"),
+          checked: !isDisabled,
           onChange: () => this.handleCheckbox(breakpoint)
         }),
       dom.div(
--- a/public/js/components/Breakpoints.js
+++ b/public/js/components/Breakpoints.js
@@ -69,7 +69,7 @@ const Breakpoints = React.createClass({
     return dom.div(
       {
         className: classnames({
-          breakpoint: true,
+          breakpoint,
           paused: isCurrentlyPaused,
           disabled: isDisabled
         }),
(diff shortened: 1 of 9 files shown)


public/js/components/Breakpoints.js
```diff
@@ -66,14 +68,18 @@ const Breakpoints = React.createClass({
 
     return dom.div(
       {
-        className: "breakpoint",
+        className: classnames({
+          breakpoint: true,
```
jasonLaster: You might not need the true

public/js/components/tests/Breakpoints.js
```diff
@@ -21,14 +21,18 @@ describe("Breakpoint Component", function() {
...
 
     const $el = renderComponent(Breakpoints, "todomvcUpdateOnEnter");
     const breakpoints = getBreakpoints($el);
-    expect(breakpoints.length).to.equal(1);
+    expect(breakpoints.length).to.equal(3);
     expect(getBreakpointLabel(breakpoints[0])).to.equal("113 this.close();");
+    expect(getBreakpointLabel(breakpoints[1])).to
+    .equal("142 return undefined;");
+    expect(getBreakpointLabel(breakpoints[2])).to
+    .equal("169 var name = text.toLowerCase();");
```
jlongster: The indentation looks a little weird here. Preferably:

```
expect(getBreakpointLabel(breakpoints[1]))
  .to.equal("142 return undefined;");
```

public/js/test/fixtures/todomvc.updateOnEnter.json
```diff
@@ -112,6 +112,28 @@
...
+      "server2.conn120.source42:169": {
+        "location": {
+          "sourceId": "server2.conn120.source42",
+          "line": 169
+        },
+        "condition": null,
+        "disabled": true,
+        "loading": false,
+        "id": "server2.conn120.50",
+        "text": "var name = text.toLowerCase();"
```
jasonLaster: Nice. I assume you added these by hand. 

We generate the fixtures with cypress in the tests/integrations/fixtures tests. You can add two more breakpoints in the updateOnEnter test. 

Cypress requires an account to run so once you install it you'll need to fill out an adoptee form and then ping them in their gitter channel so you have early access. . https://gitter.im/cypress-io/cypress

This info should be in the readme :)

jlongster:
We should work with Helen to figure "selected" states in general. The state for a breakpoint that is currently paused is that it is "selected". The problem is that several things in the UI can be "selected": the current source, frame, breakpoint, scope item, etc. We should figure out the right way to highlight everything that makes sense. Some of those may not need to be highlighted at all, like the "selected breakpoint" which doesn't seem very useful to me.

jasonLaster:
@clarkbw looks good. Sorry I can't help more on the click handler. Commenting from my phone.

clarkbw:
Yes, I think the important aspect is where we want to direct the attention
from our users. Especially considering all the various selection and other
settings available. I think the editor is the right place initially. We
should also consider afterward we may want to give a subtle hint to the
buttons for stepping. Somewhere else though, not here.

jlongster:
If you fix the linting error, I'll merge this

clarkbw:
Ok, thanks. I can fix it up this weekend.

jasonLaster:
Created a new issue for discussing the breakpoint UI https://github.com/jlongster/debugger.html/issues/291. I think what you have is good and we should keep it, but worth discussing more in depth as well.

clarkbw:
I'm just waiting on the Cypress account to finish this up.

clarkbw:
Looks good!  @jlongster r?

jlongster:
Looks good! I wish the fixture updates didn't change the IDs every time, but that's a separate problem.

clarkbw:
> I wish the fixture updates didn't change the IDs every time, but that's a separate problem.

Yeah, it makes the diff really hard to comprehend.  Would be nice to have that fixed.

PR #288, opened 2016-07-01
greatly improve scope inspector, split out object inspector component and optimize it

jlongster:
This is a heavy refactoring of scopes that I've been wanting to do. There is now a standalone `ObjectInspector.js` component that will be easy to share. The only thing is that it depends on our tree stuff, and we've tweaked the tree from m-c and we have `ManagedTree`. But it shouldn't be hard to reconcile the tree component, and once that's done that component can be shared.

This greatly improves performance. If you have the window scope expanded with a million items, the tree is still fully interactive and you don't notice any slowdown.

Additionally, it's more accurate as to what values it shows. There are still things missing, and we should show things like functions better, but there's also much clearer places in the code to implement those kinds of things now.

screen shot 2016-07-01 at 1 15 59 pm

### DIFF
--- a/public/js/components/ObjectInspector.js
+++ b/public/js/components/ObjectInspector.js
@@ -1,0 +1,167 @@
+"use strict";
+
+const React = require("react");
+const classnames = require("classnames");
+const ManagedTree = React.createFactory(require("./util/ManagedTree"));
+const Arrow = React.createFactory(require("./util/Arrow"));
+const Rep = require("./Rep");
+const { DOM: dom, PropTypes } = React;
+
+// This implements a component that renders an interactive inspector
+// for looking at JavaScript objects. It expects descriptions of
+// objects from the protocol, and will dynamically fetch child
+// properties as objects are expanded.
+//
+// If you want to inspect a single object, pass the name and the
+// protocol descriptor of it:
+//
+//  ObjectInspector({
+//    name: "foo",
+//    desc: { writable: true, ..., { value: { actor: "1", ... }}},
+//    ...
+//  })
+//
+// If you want multiple top-level objects (like scopes), you can pass
+// an array of manually constructed nodes as `roots`:
+//
+//  ObjectInspector({
+//    roots: [{ name: ... }, ...],
+//    ...
+//  });
+
+// There are 3 types of nodes: a simple node with a children array, an
+// object that has properties that should be children when they are
+// fetched, and a primitive value that should be displayed with no
+// children.
+
+function nodeHasChildren(item) {
+  return Array.isArray(item.contents);
+}
+
+function nodeHasProperties(item) {
+  return !nodeHasChildren(item) && item.contents.value.type === "object";
+}
+
+function nodeIsPrimitive(item) {
+  return !nodeHasChildren(item) && !nodeHasProperties(item);
+}
+
+function createNode(name, path, contents) {
+  // The path is important to uniquely identify the item in the entire
+  // tree. This helps debugging & optimizes React's rendering of large
+  // lists. The path will be separated by property name,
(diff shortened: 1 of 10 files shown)


public/js/components/ObjectInspector.js
```diff
@@ -0,0 +1,169 @@
+"use strict";
+
+const React = require("react");
+const classnames = require("classnames");
+const ManagedTree = React.createFactory(require("./util/ManagedTree"));
+const Arrow = React.createFactory(require("./util/Arrow"));
+const Rep = require("./Rep");
+const { DOM: dom, PropTypes } = React;
+
+// This implements a component that renders an interactive inspector
```
TatumCreative: Thanks for the high-level description, I find these really helpful at the top of files.

public/js/components/ObjectInspector.js
```diff
@@ -0,0 +1,169 @@
...
+  getChildren(item) {
+    const { getObjectProperties } = this.props;
+    const obj = item.contents;
+
+    // Nodes can either have children already, or be an object with
+    // properties that we need to go and fetch.
+    if (nodeHasChildren(item)) {
+      return item.contents;
+    } else if (nodeHasProperties(item)) {
+      const actor = obj.value.actor;
```
TatumCreative: I'm curious, is the concept of using actors here Firefox specific? I know you all had been doing some generic interfaces backed by a specific client for other parts of the debugger.

public/js/components/ObjectInspector.js
```diff
@@ -0,0 +1,169 @@
...
+    loadObjectProperties: PropTypes.func.isRequired
+  },
+
+  displayName: "ObjectInspector",
+
+  getInitialState() {
+    // Cache of dynamically built nodes. We shouldn't need to clear
+    // this out ever, since we don't ever "switch out" the object
+    // being inspected.
+    this.actorCache = {};
```
TatumCreative: It feels weird to have state not on the state object in the context of a React component. I've not seen this before.

public/js/components/Scopes.js
```diff
@@ -3,29 +3,30 @@
...
-const Rep = require("./Rep");
 
 require("./Scopes.css");
 
 function info(text) {
   return dom.div({ className: "pane-info" }, text);
 }
 
-function getBindingVariables(bindings) {
+function getBindingVariables(bindings, parentName) {
```
TatumCreative: A little bit of docs would be helpful here. Not sure what bindings refer to exactly.

TatumCreative:
That's it from me, mainly a few questions.

public/js/components/ObjectInspector.js
jlongster: Yes they are specific. We haven't made the scopes generic yet. It should be clearer how to do that now, but I'm not sure. We'll do it in the next few weeks probably.

public/js/components/ObjectInspector.js
jlongster: I get what you mean, but it's actually pretty common to store things on element instances. It's a good way to keep track of things that aren't rendering state, and more complex situations (like this one) need it. react-redux has [many of them](https://github.com/reactjs/react-redux/blob/master/src/components/connect.js#L227) (maybe too many).

The only thing I'm unsure of is if I ever need to clear it out, but I can think of when to do that. I could do it in `componentWillUnmount` but that's pointless because that means the instance is going to be GCed anyway.

public/js/components/Scopes.js
jlongster: Will do. This part will be removed when we make the protocol more generic.

jlongster:
Added more comments and switched to `Array.isArray`

PR #289, opened 2016-07-03
Add tooltip to breakpoint snippet

amitzur:
Currently breakpoint info includes only line number and a truncated snippet of code.
I wish there was also the file name, but I couldn't find it in the breakpoint object.
I would ultimately do 2 lines in the breakpoint - first line with `filename.js:line` and second line with smaller grey text of the truncated snippet, plus a tooltip of the full snippet.

This pull request adds only a tooltip of the full snippet.

### DIFF
--- a/public/js/components/Breakpoints.js
+++ b/public/js/components/Breakpoints.js
@@ -77,7 +77,7 @@ const Breakpoints = React.createClass({
           onChange: () => this.handleCheckbox(breakpoint)
         }),
       dom.div(
-        { className: "breakpoint-label" },
+        { className: "breakpoint-label", title: breakpoint.get("text") },
         `${line} ${snippet}`
       ),
       isPausedIcon


amitzur:
Also, I wanted to run Cypress tests but I didn't have it installed. I went to their website and saw it was only for early adopters. I unwillingly gave my email address, but then came a huge form that they asked me to fill out.
Isn't there an easier way to get it?

jasonLaster:
Haha yea, cypress does have a big form. I'll update the readme soon, but we have a relationship with them so they'll give you access if you ask in their gitter channel.

jasonLaster:
Another thing to keep in mind is that conditional breakpoints also show their expressions.

jasonLaster:
Merged this PR as it looks good and created a new issue https://github.com/jlongster/debugger.html/issues/291 for discussing the UI

amitzur:
@jlongster I couldn't get to the file name with `["source", "url"]` either.
It looked to me like `["location", "source"]` could work since it has an entry, but its value is undefined.

@jasonLaster thanks for merging!

jasonLaster:
Nice catch @amitzur here's the fix https://github.com/jlongster/debugger.html/commit/f85e63bd6a5dd89bbd5eb8859accd6f6cae8d290

PR #290, opened 2016-07-05
Add Getting Involved Section

jasonLaster:
Trying to help people find their way to our IRC room. I punted on a new channel name as i don't think that should be a blocker.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,6 @@ This is a prototype debugger written without any XUL and based on React and Redu
 
 [![Build Status](https://travis-ci.org/jlongster/debugger.html.svg?branch=master)](https://travis-ci.org/jlongster/debugger.html)
 
-----
 #### Getting Started
 
 * `npm install` - Install Dependencies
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,9 +1,10 @@
-### debugger.html
+### Debugger.html
 
 This is a prototype debugger written without any XUL and based on React and Redux.
 
 [![Build Status](https://travis-ci.org/jlongster/debugger.html.svg?branch=master)](https://travis-ci.org/jlongster/debugger.html)
 
+----
```
jlongster: why these lines?

README.md
```diff
@@ -12,6 +13,17 @@ This is a prototype debugger written without any XUL and based on React and Redu
...
+#### Getting Involved
+
+The Debugger is an open source project and we would love your help. We have prepared a [contributing](https://github.com/jlongster/debugger.html/blob/master/CONTRIBUTING.md) guide to help answer some of the common questions.
+
+Also, please feel free to stop by and say hi! We're all in Mozilla's [Devtools](irc://irc.mozilla.org/devtools) channel. If you do not have IRC, we recommend using the free version of [irccloud](https://www.irccloud.com/). 
++ Address `irc.mozilla.org`
++ Port `6697`
++ Channel `devtools`
+
+----
```
jlongster: very nice. It might be a little more condensed to just say irc.mozilla.org channel #devtools? The port is the default port, so I don't think you really need to specify it.

jasonLaster: I like that.

README.md
jasonLaster: It was feeling long. I'm not a big fan of them. I'll delete them, the better solution is probably to break some of the content out into other doc files

jasonLaster:
Merged. Thanks for the cleanup

PR #292, opened 2016-07-05
add more info to readme and slightly restructure

jlongster:


README.md
```diff
@@ -1,81 +1,76 @@
...
+## Getting Started
 
 * `npm install` - Install Dependencies
 * `npm run firefox` - Start Firefox
 * `npm start` - Start Debugger
 
 ![screen shot 2016-05-16 at 1 24 29 pm](https://cloud.githubusercontent.com/assets/254562/15297643/34575ca6-1b69-11e6-9703-8ba0a029d4f9.png)
 
-#### Getting Involved
+If you would like to open a specific Firefox instance or use a specific profile, first make sure it [has the appropate flags](#) enabled. Then run the following command and replace <Firefox.app> with your Firefox. For example, nightly would be `FirefoxNightly.app`.
```
jasonLaster: I'd prefer keeping this content in Miscellaneous as it's something that most open source contributors won't need.

README.md
```diff
@@ -1,81 +1,76 @@
...
+
+```
```
jasonLaster: Also, while you're in here it would be nice to add content about starting Chrome. I think many people will want to debug a chrome page over firefox

README.md
jlongster: The README isn't just for open-source contributors though. And I think it's good to teach how things actually work. For me, the first thing I wanted to do was to tell `npm run firefox` to use a different version, but I had no idea how to do that. I think this info is pretty crucial for anyone who wants to more control, and it's feels weird putting it at the very bottom. It's not that scary and people can skip it.

README.md
jlongster: Yeah I thought about that, I can do that.

jlongster:
I didn't mean to --amend the commit but I did, sorry. I added a section about Chrome, what do you think?

jlongster:
Merging as it's really easy to do follow-ups if there we decide on anything different.

PR #293, opened 2016-07-05
Show hidden sources in a dropdown

jasonLaster:
This PR adds a dropdown for showing hidden sources

It's a relatively low priority, but was fun to work on on the plane.

The implementation is fairly simple, find the tabs who's offset suggest they're not in the top row. Surprisingly there were many other details:
+ only show the dropdown button if there are hidden tabs
+ select a tab should move it to the front

I chose to do a very simple dropdown panel. We should use a better one soon. Also, we should consider a context-menu element that uses the same API, so in the toolbox we can fall back on that. I believe @bgrins has one that works like this

### DIFF
--- a/public/js/components/SourceTabs.js
+++ b/public/js/components/SourceTabs.js
@@ -1,7 +1,8 @@
 "use strict";
 
 const React = require("react");
-const { DOM: dom } = React;
+const { DOM: dom, PropTypes } = React;
+const ImPropTypes = require("react-immutable-proptypes");
 const { connect } = require("react-redux");
 const { bindActionCreators } = require("redux");
 const Isvg = React.createFactory(require("react-inlinesvg"));
@@ -12,6 +13,7 @@ const actions = require("../actions");
 const { isEnabled } = require("../../../config/feature");
 
 require("./SourceTabs.css");
+require("./Dropdown.css");
 
 /**
  * TODO: this is a placeholder function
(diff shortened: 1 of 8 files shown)


public/js/components/SourceTabs.js
```diff
@@ -30,45 +32,158 @@ function getFilename(url) {
...
+
+  renderDropdownSource(source) {
+    const { selectSource } = this.props;
+    const url = source && source.get("url");
+    const filename = getFilename(url);
+
+    return dom.li({
+      key: source.get("id"),
+      onClick: () => {
+        selectSource(source.get("id"), { position: 0 });
```
jlongster: Makes sense to make this an option. Wouldn't tabs normally open at the end rather than the beginning though?

public/js/components/SourceTabs.js
```diff
@@ -30,45 +32,158 @@ function getFilename(url) {
...
+    if (!hiddenSourceTabs || hiddenSourceTabs.size == 0) {
+      return dom.div({});
+    }
+
+    return dom.span(
+      {
+        className: "subsettings",
+        onClick: this.toggleSourcesDropdown
+      },
+      dom.img({ src: "images/subSettings.svg" })
```
jlongster: We might want a different icon. It's a little confusing to see 2 of those pretty close to each other. I think that icon is for settings specifically. If possible, use a placeholder for now.

public/js/components/SourceTabs.js
```diff
@@ -30,45 +32,158 @@ function getFilename(url) {
...
+    }
+
+    return dom.div(
+      {
+        className: classnames("source-tab", { active }),
+        key: source.get("id"),
+        onClick: () => selectSource(source.get("id"))
+      },
+      dom.div({ className: "filename" }, filename),
+      dom.div({ onClick: onClickClose },
```
jlongster: nit: there's a mixture of styles here with the arguments. I thought we decided to bump down the second argument. But I could be convinced that this is too useful, as I've seen it elsewhere. Not a big deal for now, we can ignore this.

public/js/components/SourceTabs.js
```diff
@@ -12,6 +13,7 @@ const actions = require("../actions");
 const { isEnabled } = require("../../../config/feature");
 
 require("./SourceTabs.css");
+require("./Dropdown.css");
```
jlongster: Why a separate file? CSS and JS files should be a 1:1 mapping.

jlongster: If you want to extract it, a separate dropdown component would simplify this component also.

jlongster:
Looks good to me! Just a few comments, but nothing crazy. r+

public/js/components/SourceTabs.js
jasonLaster: hmm, i can see that

public/js/components/SourceTabs.js
jasonLaster: yeah, that's true. I believe, @helenvholmes proposed vertical elipsis actually. Can't really remember

public/js/components/SourceTabs.js
jasonLaster: yeah - that's totally true. I don't want to commit to writing a new Dropdown component though :) I'd rather find a community one

public/js/components/SourceTabs.js
jasonLaster: agreed

jasonLaster: fixed

jasonLaster:
merged

public/js/components/SourceTabs.js
jlongster: Agreed (thought I was just meaning to extract the dropdown already in this component), but in that case I think we should just have the dropdown css in the source tabs css. I don't think we should have arbitrary CSS files; they exist as an implementation for a component, so it's weird to have standalone ones.

public/js/components/SourceTabs.js
helenvholmes: If this is for subsettings, yes!

I'll get on adding this in tomorrow when I'm not on mobile.

PR #295, opened 2016-07-05
Add some handy utils

jasonLaster:
minor things

### DIFF
--- a/public/js/main.js
+++ b/public/js/main.js
@@ -40,6 +40,7 @@ const actions = bindActionCreators(require("./actions"), store.dispatch);
 // global for debugging purposes only!
 window.store = store;
 window.injectDebuggee = require("./test/utils/debuggee");
+window.serializeStore = () => JSON.parse(JSON.stringify(store.getState()));
 
 function renderRoot(component) {
   const mount = document.querySelector("#mount");
--- a/public/js/util/utils.js
+++ b/public/js/util/utils.js
@@ -97,6 +97,12 @@ function entries(obj) {
   return Object.keys(obj).map(k => [k, obj[k]]);
 }
 
+function mapObject(obj, iteratee) {
+  return toObject(entries(obj).map(([key, value]) => {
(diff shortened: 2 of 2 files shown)


public/js/main.js
```diff
@@ -40,6 +41,7 @@ const actions = bindActionCreators(require("./actions"), store.dispatch);
 // global for debugging purposes only!
 window.store = store;
 window.injectDebuggee = require("./test/utils/debuggee");
+window.serializeStore = () => JSON.parse(JSON.stringify(store.getState()));
```
jasonLaster: this is a nice to have so that you can easily inspect state... takes advantage of the serializability of immutable js

public/js/util/utils.js
```diff
@@ -97,6 +97,12 @@ function entries(obj) {
   return Object.keys(obj).map(k => [k, obj[k]]);
 }
 
+function mapObject(obj, iteratee) {
```
jasonLaster: i initially wrote this for serializing the store and i don't have the heart to kill it. I've wanted other times to

jlongster: Sounds good to me. I'm also fine if you want to pull in a few lodash methods eventually. I'd rather not depend on a huge amount of lodash, but some of its methods might be helpful. (I know I was more resistant before.) This is good for now.

public/js/main.js
jlongster: I'd rather not keep adding globals here in main.js... would you be ok creating a new module somewhere that does all this injection? Could be something like `require("./utils/debug")(store)` and that module would only inject all this stuff if `process.env.NODE_ENV !== "production"` or something (or whatever debug mode flag).

public/js/util/utils.js
jasonLaster: haha - sure.

public/js/main.js
jasonLaster: sounds good. let me land this and then i'll extract to another module

PR #296, opened 2016-07-05
add debugGlobal

jasonLaster:


public/js/main.js
```diff
@@ -37,10 +34,14 @@ const createStore = configureStore({
...
-window.injectDebuggee = require("./test/utils/debuggee");
-window.serializeStore = () => JSON.parse(JSON.stringify(store.getState()));
+if (isEnabled("development")) {
+  AppConstants.DEBUG_JS_MODULES = true;
+  debugGlobal("store", store);
+  debugGlobal("injectDebuggee", require("./test/utils/debuggee"));
+  debugGlobal("serializeStore", () => {
+    return JSON.parse(JSON.stringify(store.getState()))
+  });
+}
```
jlongster: I was hoping to move all of this out of main.js and into the debug module. Something like `injectGlobals` that would be called here (you'd need to pass what you want to inject).

public/js/util/debug.js
```diff
@@ -0,0 +1,13 @@
+const { isEnabled } = require("../../../config/feature");
+
+ function debugGlobal(field, value) {
+  if (!isEnabled("development")) {
+    return;
+  }
+
+  window[field] = value;
+}
```
jlongster: Nice idea, I hadn't thought of it quite like that.

public/js/main.js
jasonLaster: sure, i can do that

PR #297, opened 2016-07-05
Pretty Printing + moves source buttons to the footer

jasonLaster:
Starts the pretty print work and moves the source buttons to the footer.

I'm not sure why the pretty print command fails. Will hopefully be clear in the morning when i'm fresh. The firefox debugger can pretty print the file (backbone.localStorage) in todomvc.

screen shot 2016-07-05 at 5 47 47 pm


public/js/clients/firefox/commands.js
```diff
@@ -98,6 +96,16 @@ function getProperties(grip) {
...
 
+function prettyPrint(sourceId, indentSize) {
+  const sourceClient = threadClient.source({ actor: sourceId });
+  return sourceClient.prettyPrint(indentSize);
+}
+
+function disablePrettyPrint(sourceId) {
+  const sourceClient = threadClient.source({ actor: sourceId });
+  return sourceClient.disablePrettyPrint();
+}
```
jasonLaster: I'm thinking about making this togglePrettyPrint, but doesn't really matter

public/js/components/Editor.js
```diff
@@ -108,8 +108,7 @@ const Editor = React.createClass({
     // Only reset the editor text if the source has changed.
     // + Resetting the text will remove the breakpoints.
     // + Comparing the source text is probably inneficient.
-    if (!oldSourceText ||
-        newSourceText.get("text") != oldSourceText.get("text")) {
+    if (newSourceText.get("text") != this.editor.getValue()) {
```
jasonLaster: changed this, because Loading... was staying there. Might be able to change it back when Pretty Print works

public/js/components/SourceTabs.js
```diff
@@ -162,7 +162,7 @@ const SourceTabs = React.createClass({
     const { selectedSource, selectSource, closeTab } = this.props;
     const url = source && source.get("url");
     const filename = getFilename(url);
-    const active = source.equals(selectedSource);
+    const active = source.get("id") == selectedSource.get("id");
```
jasonLaster: had to change this because the source.isPrettyPrinted field changed

public/js/clients/firefox/commands.js
jlongster: I think it's better this way because some protocols may not have the current state. we happen to have it in the current SourceClient, but I bet chrome just has these methods without the current state

public/js/components/App.js
```diff
@@ -22,16 +25,18 @@ const App = React.createClass({
 
   displayName: "App",
 
-  render: function() {
+  render: function(command) {
```
jlongster: remove this argument

public/js/components/App.js
```diff
@@ -41,5 +46,6 @@ const App = React.createClass({
 
 module.exports = connect(
   state => ({ sources: getSources(state),
-              breakpoints: getBreakpoints(state) })
+              breakpoints: getBreakpoints(state) }),
+  dispatch => bindActionCreators(actions, dispatch)
```
jlongster: I don't see actions being used anywhere? Is this just leftover? If so, please remove it

public/js/components/Editor.js
jlongster: The current debugger has a nice progress bar. But yeah, this is fine for now

jlongster:
Cool, looks good! r+ with minor nits/linting fixed

public/js/components/Editor.js
jasonLaster: true - i'll add a note in the spreadsheet

public/js/components/App.js
jasonLaster: cool

public/js/clients/firefox/commands.js
jasonLaster: ahh - yeah, i was thinking something like `togglePrettyPrint(wantPretty)`. but it doesn't matter

jasonLaster:
merged

PR #298, opened 2016-07-06
Show screenshots

jasonLaster:
It came up yesterday in channel that it would be nice to show some screenshots of the debugger so people can see what it will look like.


jlongster:
Sounds good

PR #299, opened 2016-07-06
Add more documentation

jasonLaster:
Added more top level docs on our dev environment and integration tests.

Some of the dev environment is redundant with the readme. I think that is okay as we want to help people get started and then provide comprehensive information. With that said, this is a start and i'm sure it can get much better.

Working on this I realized how much information is in our heads that should be written down. I'll try to keep building on this going forward.


docs/integration-tests.md
```diff
@@ -0,0 +1,74 @@
+## Integration Tests
+
+Debugger.html integration test are end-to-end tests that simulate a user debugging an application.
```
jlongster: Looks fine to me, but I wouldn't work on this too much more right now. This is most unknown thing right now. Personally I'm not sure we will end up using cypress, but we need to look into other things before making that decision. As this is in such flux, I wouldn't encourage people to write these tests right now.

jlongster:
Looks great. I was just thinking that it would be good to write guidelines for redux and react development, as there are several ways to do things but we desire to have one specific workflow. The local dev doc is really helpful especially.

docs/integration-tests.md
jasonLaster: Yup. just wrote it down quickly for the cases where people *are* using the tool and might be confused

jasonLaster:
That would be great. I like the pattern of having a README, top level docs for guides, and then more topical docs inline that we can point to

PR #301, opened 2016-07-06
Compile config on the server

### LINKED ISSUE #286: Move configs to the server
Yesterday you mentioned that the server should read the config values and pass them down to the client as a set of feature flags. This lines up with what I was planning on doing, so i thought i'd quickly sketch it out.

https://gist.github.com/jasonLaster/be137dd316bcabf0de55acd55c1bdaec

The benefit is that the client can be dumb and just read the feature flags. We should try to line this up with Prefs. 

Also, the reason that I moved the queryParam parsing and node env check to feature.js was so that we built our set of feature flags in one place, that could be replaced with something like nconf or another config tool that is better at reading configs from files, env vars, and what not...

Anyways, this just some rough thoughts i wanted to share while we were on the topic.

jasonLaster:
Fixes #286 

old: bundle the config
new: dev server adds the config into index.html, handles `fs` work

### DIFF
--- a/bin/development-server.js
+++ b/bin/development-server.js
@@ -10,19 +10,13 @@ const express = require("express");
 const webpackDevMiddleware = require("webpack-dev-middleware");
 const webpackHotMiddleware = require("webpack-hot-middleware");
 const http = require("http");
+const _ = require("lodash");
 
-// Improve the first run experience: copy the local sample config to a
-// local file so webpack can find it. We don't have to do this, but
-// otherwise webpack prints a warning that scares people.
-const localConfigPath = path.join(__dirname, "../config/local.json")
-if(!fs.existsSync(localConfigPath)) {
-  const defaultConfig = fs.readFileSync(
-    path.join(__dirname, "../config/local.sample.json")
-  );
-  fs.writeFileSync(localConfigPath, defaultConfig);
-}
-
-const features = require("../config/feature");
+// Setup Config
+const getConfig = require("../config/config").getConfig;
+const features = require("../config/feature")
+const config = getConfig();
+features.setConfig(config);
 
 require("./firefox-proxy");
 
@@ -41,12 +35,11 @@ function httpGet(url, onResponse) {
 const app = express();
 
 // Webpack middleware
-
-const config = require("../webpack.config");
-const compiler = webpack(config);
+const webpackConfig = require("../webpack.config");
+const compiler = webpack(webpackConfig);
 
 app.use(webpackDevMiddleware(compiler, {
-  publicPath: config.output.publicPath,
+  publicPath: webpackConfig.output.publicPath,
   noInfo: true,
   stats: { colors: true }
 }));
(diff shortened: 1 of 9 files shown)


bin/development-server.js
```diff
@@ -10,19 +10,13 @@ const express = require("express");
...
-  );
-  fs.writeFileSync(localConfigPath, defaultConfig);
-}
-
-const features = require("../config/feature");
+// Setup Config
+const getConfig = require("../config/config").getConfig;
+const features = require("../config/feature")
+const config = getConfig();
+features.setConfig(config);
```
jasonLaster: Sets the config in `feature` so that webpack and other server-side code has access to features. It's a small temporary hack

bin/development-server.js
```diff
@@ -61,9 +54,12 @@ app.use(express.static("public/js/test/examples"));
 app.use(express.static("public"));
 
 // Routes
-
 app.get("/", function(req, res) {
-  res.sendFile(path.join(__dirname, "../index.html"));
+  const tpl = fs.readFileSync(path.join(__dirname, "../index.html"));
+  const html = _.template(tpl)({
```
jasonLaster: uses lodash to template. there are other template engines, but this is very simple. Also, used lodash below. Didn't take long to add lodash :)

config/feature.js
```diff
@@ -1,57 +1,35 @@
 "use strict";
 
-let config = {};
+let config = typeof window != "undefined" ? window.debuggerConfig : {};
+const _ = require("lodash");
```
jasonLaster: i'm going to try importing just get so that the bundle does not grow

config/firefox-panel.json
```diff
@@ -0,0 +1,8 @@
+{
```
jasonLaster: This is probably the most interesting piece of the PR. 

Previously there was development.json and local.json. Adding firefox-panel.json, lets us have separate configs for the panel or other environments we might support in the future.

bin/development-server.js
jlongster: Nice. I'd very much prefer to pull in individual lodash methods though, so just require `lodash-template`. I don't see us using much of lodash, so it'd be great to only pull in the methods we need.

config/config.js
```diff
@@ -0,0 +1,33 @@
...
+// only used for testing purposes
+function stubConfig(stub) {
+  config = stub;
+}
+
+function getConfig() {
+  return config;
+}
+
+function resetConfig(stub) {
```
jlongster: I prefer the other API with just `getConfig` and `setConfig`, with the latter returning the previous config. It's simpler and more functional (no need to worry about originalConfig in here).

config/feature.js
jlongster: Cool, didn't see this before I made my other comment!

config/firefox-panel.json
jlongster: It'd be great to have these inherit so that we don't duplicate the config across everything. A base config would work everywhere.

index.html
```diff
@@ -8,6 +8,7 @@
 
   <body>
     <div id="mount"></div>
+    <script>window.debuggerConfig = <%= debuggerConfig %>; </script>
```
jlongster: The only problem with this is that we aren't running a server in the devtools panel, so this won't work. We need raw HTML file for the panel.

jlongster:
Cool, I like this direction a lot, just a few things to think about. This looks good generally!

config/firefox-panel.json
jasonLaster: That makes sense

config/config.js
jasonLaster: oops

index.html
jasonLaster: I was thinking we could do something with webpack.production.config to inject config. Although... we could probably do that with webpack.config as well

bin/development-server.js
jasonLaster: agreed

PR #302, opened 2016-07-06
Add increment example

jasonLaster:
This is a good example of a page that uses a source map


jlongster:
What are those karma errors? I'm sure it's not related, but I haven't seen them before

jasonLaster:
i'm not sure. i might have seen it after london once or twice...

jlongster:
For some reason I don't seem to be able to trigger a rebuild. Do you mind doing that (even by force pushing) and seeing if it's an intermittent?

jasonLaster:
looks intermittent

PR #309, opened 2016-07-08
add line numbers to frame source location

clarkbw:
adds a line number to the source location if the source location exists
if there is no source location it no longer adds an empty element
also lightens the text a bit, hopefully @helenvholmes is cool with that

screen shot 2016-07-08 at 3 56 53 pm

### DIFF
--- a/public/js/components/Frames.css
+++ b/public/js/components/Frames.css
@@ -19,6 +19,7 @@
 .frames .location {
   float: right;
   color: #666666;
+  font-weight: lighter;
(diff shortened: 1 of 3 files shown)


jlongster:
I like it! We may want to remove the space after `:`, as it's a common format to have `file:line:col` as a location. But I'll go ahead and merge it!

helenvholmes:
Already merged, but definitely cool with this 🙃

PR #310, opened 2016-07-12
Update center pane

jasonLaster:
It now shows a welcome box and a sources search section.

### DIFF
--- a/public/js/components/App.js
+++ b/public/js/components/App.js
@@ -3,6 +3,8 @@
 const React = require("react");
 const { DOM: dom, PropTypes, createFactory } = React;
 const { connect } = require("react-redux");
+const { bindActionCreators } = require("redux");
+const actions = require("../actions");
 
 require("./App.css");
 require("../lib/variables.css");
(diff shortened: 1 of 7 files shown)


nt1m:
Would be nice if there was a new tab button (next to the tabs) that would display the blank screen with just the searchbox

jasonLaster:
@nt1m that's a nice idea. @helenvholmes ?

public/js/components/App.js
```diff
@@ -12,30 +14,106 @@ const SplitBox = createFactory(require("./SplitBox"));
...
+  renderWelcomeBox() {
+    return dom.div(
+      { className: "welcomebox" },
+      "Hit Cmd + P to open a file"
+    );
+  },
+
+  renderCenterPane() {
+    let centerPane;
+    if (this.state.searchOn) {
```
jlongster: This kind of stuff, because it's very basic (and important) app state, might be nice to store in redux. But we can easily move it out later.

jlongster:
Very cool! My only nit is that it says *exactly* what Chrome says, and it's an easy thing to word differently to avoid looking like we're copying them word-for-word. If you tweak that you can merge it

public/js/components/App.js
```diff
@@ -12,30 +14,106 @@ const SplitBox = createFactory(require("./SplitBox"));
...
+          value: getSourcePath(source),
+          title: getSourcePath(source).split("/").pop(),
+          subtitle: getSourcePath(source),
+          id: source.get("id")
+        }))
+        .toJS();
+    }
+
+    return Autocomplete({
+      selectItem: (result) => {
```
jlongster: nit: devtools style is to forego the parenthesis when there's only one arg

nt1m:
@jlongster I wonder why we can't display the searchbox directly instead of that message.

clarkbw:
> I wonder why we can't display the searchbox directly instead of that message

I was just wondering that as well.  We can do this as a follow up to improve upon this.  Would be a much simpler issue for anyone to take on once the rest has been done.

I think we want to teach people the keyboard command to always bring up the search so we'll want to show it no matter what, but I think we could display it directly under the input.

@helenvholmes ?

helenvholmes:
I'm fine with just displaying the search bar if we display the key command underneath it before users begin typing—I think teaching people the key command is helpful.

public/js/components/App.js
jasonLaster: hmm, i could see that. I'd like to land this and then we can re-evaluate down the road

jasonLaster:
> I wonder why we can't display the searchbox directly instead of that message

Could be nice. I'd like to see how this plays with other project level searches like functions or text and see what we can do

PR #311, opened 2016-07-12
Load source maps

### LINKED ISSUE #305: Load Source Map file locally
Currently the source map files are loaded server side with priveledged APIs.

I think we should do something similar to what chrome does.

Locally: use the development server to make the cross origin request
Toolbox: use a priviledged API to make the request.

jasonLaster:
Fixes #305  - Step 1 of source maps - load the source map file

### DIFF
--- a/bin/development-server.js
+++ b/bin/development-server.js
@@ -59,22 +59,17 @@ app.get("/", function(req, res) {
   res.sendFile(path.join(__dirname, "../index.html"));
 });
 
-app.get("/chrome-tabs", function(req, res) {
-  if(feature.isEnabled("chrome.debug")) {
-    const webSocketPort = feature.getValue("chrome.webSocketPort");
-    const url = `http://localhost:${webSocketPort}/json/list`;
+app.get("/get", function(req, res) {
+  const httpReq = httpGet(
+    req.query.url,
+    body => res.json(JSON.parse(body))
+  );
 
-    const tabRequest = httpGet(url, body => res.json(JSON.parse(body)));
-
-    tabRequest.on('error', function (err) {
-      if (err.code == "ECONNREFUSED") {
-        console.log("Failed to connect to chrome");
-      }
-    });
-  }
-  else {
-    res.json([]);
-  }
+  httpReq.on('error', function (err) {
+    if (err.code == "ECONNREFUSED") {
+      console.log("Failed to fetch resource");
+    }
+  });
 })
 
 // Listen
--- a/config/config.js
+++ b/config/config.js
@@ -4,8 +4,10 @@ const _ = require("lodash");
 const fs = require("fs");
 const path = require("path");
 
+const firefoxPanel = require("./firefox-panel.json");
+const development = require("./development.json");
 const envConfig = process.env.NODE_ENV === "production" ?
-  require("./firefox-panel.json") : require("./development.json");
+   firefoxPanel : development;
 
 const localConfig = fs.existsSync(path.join(__dirname, "./local.json")) ?
   require("./local.json") : {};
--- a/config/development.json
+++ b/config/development.json
@@ -5,7 +5,8 @@
     "firefoxProxy": false
   },
   "features": {
-    "tabs": true
+    "tabs": true,
+    "sourceMaps": true
   },
   "chrome": {
     "debug": false,
--- a/public/js/actions/sources.js
+++ b/public/js/actions/sources.js
@@ -12,6 +12,8 @@ const { getSource, getSourceText } = require("../selectors");
 const constants = require("../constants");
 const Prefs = require("../prefs");
 const invariant = require("invariant");
+const { isEnabled } = require("../../../config/feature");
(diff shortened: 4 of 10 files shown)


bin/development-server.js
```diff
@@ -58,22 +58,17 @@ app.get("/", function(req, res) {
...
-    tabRequest.on('error', function (err) {
-      if (err.code == "ECONNREFUSED") {
-        console.log("Failed to connect to chrome");
-      }
-    });
-  }
-  else {
-    res.json([]);
-  }
+app.get("/get", function(req, res) {
```
jasonLaster: converting this custom endpoint to a generic endpoint that handles cross origin http requests

config/config.js
```diff
@@ -4,8 +4,10 @@ const _ = require("lodash");
 const fs = require("fs");
 const path = require("path");
 
+const firefoxPanel = require("./firefox-panel.json");
+const development = require("./development.json");
 const envConfig = process.env.NODE_ENV === "DEVTOOLS_PANEL" ?
-  require("./firefox-panel.json") : require("./development.json");
+   firefoxPanel : development;
```
jasonLaster: fixing a bug with require not respecting env configs

public/js/clients/chrome/events.js
```diff
@@ -12,7 +12,16 @@ function scriptParsed(scriptId, url, startLine, startColumn,
              endLine, endColumn, executionContextId, hash,
              isContentScript, isInternalScript, isLiveEdit,
              sourceMapURL, hasSourceURL, deprecatedCommentWasUsed) {
-  actions.newSource(Source({ id: scriptId, url: url }));
+  const origin = "http://localhost:8000/increment";
```
jasonLaster: this will be fixed in the next PR

jasonLaster:
not sure why cypress is failing here...

bin/development-server.js
```diff
@@ -58,22 +58,17 @@ app.get("/", function(req, res) {
...
+  const httpReq = httpGet(
+    req.query.url,
+    body => res.json(JSON.parse(body))
+  );
+
+  httpReq.on('error', function (err) {
+    if (err.code == "ECONNREFUSED") {
+      console.log("Failed to fetch resource");
+    }
```
jlongster: shouldn't you send a 404 or 500 here?

public/js/clients/chrome.js
```diff
@@ -39,10 +41,15 @@ function createTabs(tabs) {
...
 function connectClient() {
   const deferred = defer();
-  fetch("/chrome-tabs").then(res => {
-    res.json().then((body) => {
-      deferred.resolve(createTabs(body));
-    });
+
+  if(!isEnabled("chrome.debug")) {
+    return deferred.resolve(createTabs([]))
+  }
```
jlongster: Why do this here? We don't have a corresponding `firefox.debug` check in the firefox client, so it seems better to have it on the server.

public/js/util/networkRequest.js
```diff
@@ -0,0 +1,11 @@
+"use strict";
+
+function networkRequest(url) {
+  return fetch(`/get?url=${url}`)
```
jlongster: This seems closely the same as `httpGet` in the server, I think there is an isomorphic-fetch package or something that works in browser and server and we could potentially reuse this function.

config/development.json
```diff
@@ -3,7 +3,8 @@
   "logging": true,
   "clientLogging": false,
   "features": {
-    "tabs": true
+    "tabs": true,
+    "sourceMaps": true
```
jlongster: Nice

public/js/actions/sources.js
```diff
@@ -35,6 +37,11 @@ function queueSourcesDispatch(dispatch) {
  */
 function newSource(source) {
   return ({ dispatch }) => {
+    if (isEnabled("features.sourceMaps") && source.sourceMapUrl) {
+      networkRequest(source.sourceMapUrl).then(res => {
+      });
+    }
```
jlongster: This seems like the right place to do this. I'm curious where you think we should store sourcemaps though. Technically, they are an important part of the UI, so they should be in redux, but that might make it hard to serialize...

jlongster:
Nice start, just a few nits but this looks good.

jasonLaster:
found the cypress bug, was a nice catch. There was a bug in connect clients that i wasn't catching because i was debugging chrome with source maps and not firefox.

bin/development-server.js
jasonLaster: yea - i should. nice catch

public/js/clients/chrome.js
jasonLaster: hmm - i'm not sure, i think it might be nice to flag firefox off at somepoint too

public/js/actions/sources.js
jasonLaster: agreed - i'm planning on keeping source maps in a cache for that reason, but we will talk about it soon.

PR #314, opened 2016-07-12
Flag off firefox protocol logging

### LINKED ISSUE #193: Slow loading on very large files
@jlongster mentioned that this might be related to websockets and that it might be solved with a TCP mode for certain situations.

The current Firefox debugger is optimized for very large files, such as the [12mb file in bgrins' debugger demo](http://bgrins.github.io/devtools-demos/debugger/scripts.html).

There seems to be some lag in the debugger.html to load the source into the editor (the jump is something like a 500ms wait in the Firefox debugger to a 5s wait in the debugger.html project, although that's very much a guesstimation).

jasonLaster:
fixes #193

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,7 @@ You can see default config values in `config/development.json`, and override the
 * `hotReloding` enables hot reloading of CSS and React
 * `chrome.debug` enable local chrome development
 * `features.sourceTabs` enable editor tabs
+* `logging.firefoxProtocol` log firefox debugger protocol messages
 
 [More Information](./docs/local-development.md#configs)
 
--- a/config/development.json
+++ b/config/development.json
@@ -1,7 +1,9 @@
 {
   "environment": "development",
-  "logging": true,
-  "clientLogging": false,
+  "logging": {
+    "clientLogging": false,
+    "firefoxProtocol": true
+  },
   "features": {
     "tabs": true
   },
--- a/bin/development-server.js
+++ b/bin/development-server.js
@@ -18,7 +18,8 @@ const feature = require("../config/feature");
 const config = getConfig();
 feature.setConfig(config);
 
-require("./firefox-proxy");
+const firefoxProxy = require("./firefox-proxy");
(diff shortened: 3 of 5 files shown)


jasonLaster:
looks like it's karma being flakey... not my day w/ ci :/

README.md
```diff
@@ -74,6 +74,7 @@ You can see default config values in `config/development.json`, and override the
 * `hotReloding` enables hot reloading of CSS and React
 * `chrome.debug` enable local chrome development
 * `features.sourceTabs` enable editor tabs
+* `logging.firefoxProtocol` log firefox debugger protocol messages
```
clarkbw: I think I would have called it `logging.firefoxProxy` since its logging the Firefox protocol mostly because it needs to go through the proxy and when the proxy goes away so would the pref.

config/development.json
```diff
@@ -1,7 +1,9 @@
 {
   "environment": "development",
-  "logging": true,
-  "clientLogging": false,
+  "logging": {
+    "clientLogging": false,
```
jlongster: Why not just `client`? The logging suffix is redundant now that it's in a logging section.

README.md
jlongster: This suggestion makes sense to me.

jlongster:
We might not be getting enough out of karma yet for it's flakiness right now. Maybe we turn it off for a few weeks until we have better tests? I'm focusing on tests now in addition to my work with flow.

This PR looks good, but maybe tweak the names (see comments)? r+ otherwise

config/development.json
jasonLaster: yeah - i agree.

jasonLaster:
I'd like to keep `karma` on, but just handle the flakiness by running the test twice and seeing if one passes. That's what i've been doing with cypress and its been fine.

jlongster:
We don't have very many unit tests yet though, and what's being tested really shouldn't behave differently between browsers (although I def think we'll want karma in the long run), so personally I think node is OK for now. Just for a little bit while we figure this out. It's really annoying not knowing if a failure was your fault or not.

PR #315, opened 2016-07-12
Update mc deps

### LINKED ISSUE #313: Add a script for updating m-c files
Some of the reps files have been updated since we've landed them. We should write a script that pulls in files we depend on.

jasonLaster:
fixes #313 
updates deps and adds a script for updating deps going forward.


jlongster:
Cool. I'll go ahead and merge this. I think we should turn off karma until we figure out what's going on.

Issue #304, opened 2016-07-07
Show Source Map Files in Tree


jasonLaster:
The source tree view lists all the source files made available to the developer.  Alongside the original files will be options for the "Source Mapped Version" which is the more readable version of the file.

Issue #303, opened 2016-07-07
Show Source Maps file in Editor


jasonLaster:
Source mapped code is much more easily understood by developers, showing these versions in the file editor allows developers to read the code and possibly set breakpoints on it.

Source mapped code will show up as a separate tab to the original source code and will indicate to the user that it is "formatted" such that they can know the difference between the original and source mapped version of the same file.

PR #321, opened 2016-07-12
(WIP) Source Maps Part 2

### LINKED ISSUE #306: Create sources with the SourceMapConsumer
We should be able to re-use the source map consumer locally in a web worker to create original source map files.

jasonLaster:
Fixes #306 #304 #303 

This PR includes the work to load source map files.

This PR is just a prototype, a lot of the logic is inlined in one action, but will be extracted before it lands.

### DIFF
--- a/public/js/actions/sources.js
+++ b/public/js/actions/sources.js
@@ -8,12 +8,13 @@ const { PROMISE } = require("../util/redux/middleware/promise");
 const { Task } = require("../util/task");
 const { isJavaScript } = require("../util/source");
 
-const { getSource, getSourceText } = require("../selectors");
+const { getSource, getSourceText, getSelectedTab } = require("../selectors");
 const constants = require("../constants");
 const Prefs = require("../prefs");
 const invariant = require("invariant");
 const { isEnabled } = require("../../../config/feature");
-const { networkRequest } = require("../util/networkRequest");
+const { Source } = require("../types");
+const { loadSourceMap, getOriginalSources } = require("../util/source-map");
 
 /**
  * Throttles source dispatching to reduce rendering churn.
@@ -37,10 +38,10 @@ function queueSourcesDispatch(dispatch) {
  * Handler for the debugger client's unsolicited newSource notification.
  */
 function newSource(source) {
-  return ({ dispatch }) => {
-    if (isEnabled("features.sourceMaps") && source.sourceMapUrl) {
-      networkRequest(source.sourceMapUrl).then(res => {
-      });
+  return ({ dispatch, getState }) => {
+    if (isEnabled("features.sourceMaps") && source.sourceMapURL) {
+      const tab = getSelectedTab(getState());
+      loadSourceMapSources(tab, source, dispatch);
     }
 
     // We don't immediately dispatch the source because several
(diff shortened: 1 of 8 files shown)


public/js/main.js
```diff
@@ -76,7 +76,9 @@ function getTargetFromQuery() {
 
 const connTarget = getTargetFromQuery();
 if (connTarget) {
-  startDebugging(connTarget, actions).then(() => {
+  startDebugging(connTarget, actions).then((tabs) => {
+    actions.newTabs(tabs);
```
jasonLaster: noticed that the tabs and selectedTab state were not being set when the target was in the query

jlongster: They never need to be set? The state should be pulled out into whatever toolbox thing we come up with. But this doesn't do any harm.

public/js/actions/sources.js
```diff
@@ -36,9 +40,44 @@ function queueSourcesDispatch(dispatch) {
...
 function newSource(source) {
-  return ({ dispatch }) => {
-    if (isEnabled("features.sourceMaps") && source.sourceMapUrl) {
-      networkRequest(source.sourceMapUrl).then(res => {
+  return ({ dispatch, getState }) => {
+    if (isEnabled("features.sourceMaps") && source.sourceMapURL) {
+      const tab = getSelectedTab(getState());
+      const sourceMapURL = `${tab.get("url")}/${source.sourceMapURL}`;
+      networkRequest(sourceMapURL).then(res => {
+        Task.spawn(function* () {
```
jlongster: You could move this `spawn` up and do `yield networkRequest(...)`. It's probably good to add a catch handler here too and log the error (two common errors: soucemap fetching or parsing failed)

public/js/actions/sources.js
```diff
@@ -36,9 +40,44 @@ function queueSourcesDispatch(dispatch) {
...
+          yield dispatch(loadSourceText(source));
```
jlongster: This will give you the source text. Just do `const { text } = yield dispatch(loadSourceText(...))`

public/js/actions/sources.js
```diff
@@ -36,9 +40,44 @@ function queueSourcesDispatch(dispatch) {
...
+          const sourceText = getSourceText(getState(), source.id)
```
jlongster: We can't eagerly load in the source text for sources when they are introduced. This would create a huge amount of protocol traffic. It's already slow to transfer very large sources, and we don't want to do that at startup for every source.

The `SourceMapConsumer` should have a `sources` property which already lists the sources that the sourcemap represents. You should be able to dispatch new sources based on that, without needing any of the source contents until someone clicks on a source.

public/js/actions/sources.js
```diff
@@ -36,9 +40,44 @@ function queueSourcesDispatch(dispatch) {
...
+            url: source.path,
+            isPrettyPrinted: false
+          }))));
+
+          originalSources.map((source) => dispatch({
+            type: constants.LOAD_SOURCE_TEXT,
+            source,
+            value: {
+              text: source.text,
+              contentType: "text/JavaScript"
```
jlongster: use lowercase, `text/javascript`

jlongster:
r-. Need to avoid eagerly loading every single source's text. You'll need to figure out where to store the sourcemap, and only dispatch `newSource` actions for each source in the `sources` property of `SourceMapConsumer`, and when one of those sources is clicked it loads it from the sourcemap.

It's great to be figuring this out though. As I mentioned on the other PR, I'm not sure where the best place to store sourcemaps is. We could probably put it in redux, and when we create fixtures/serialize it, ignore the sourcemaps field to reduce the size of it. Not sure.

public/js/main.js
jasonLaster: actually, we need `selectedTab` for sourcemaps so that we can get the tab url for source maps

public/js/actions/sources.js
jasonLaster: nice call.

jasonLaster:
will open a more focused PR

PR #322, opened 2016-07-12
Add logging + assertions

jasonLaster:
Adds invariant for logging and a small logger that follows a similar pattern for logging

### DIFF
--- a/config/config.js
+++ b/config/config.js
@@ -4,7 +4,7 @@ const _ = require("lodash");
 const fs = require("fs");
 const path = require("path");
 
-const envConfig = process.env.NODE_ENV === "DEVTOOLS_PANEL" ?
+const envConfig = process.env.NODE_ENV === "production" ?
   require("./firefox-panel.json") : require("./development.json");
 
 const localConfig = fs.existsSync(path.join(__dirname, "./local.json")) ?
--- a/public/js/main.js
+++ b/public/js/main.js
@@ -9,12 +9,15 @@ const React = require("react");
 const DevToolsUtils = require("devtools-sham/shared/DevToolsUtils");
 const AppConstants = require("devtools-sham/sham/appconstants").AppConstants;
 const { injectGlobals } = require("./util/debug");
+const { log } = require("./util/utils");
 const { isEnabled, isFirefoxPanel, isDevelopment } = require("../../config/feature");
 
 // Set various flags before requiring app code.
 if (isEnabled("clientLogging")) {
   DevToolsUtils.dumpn.wantLogging = true;
 }
+
+log("Debugger bootstrapping");
 
 const { getClient, connectClients, startDebugging } = require("./clients");
 const firefox = require("./clients/firefox");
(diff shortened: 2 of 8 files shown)


public/js/main.js
```diff
@@ -9,13 +9,16 @@ const React = require("react");
...
 const { injectGlobals } = require("./util/debug");
+const { log } = require("./util/utils");
 const { isEnabled, isFirefoxPanel, isDevelopment } = require("../../config/feature");
 
 // Set various flags before requiring app code.
 if (isEnabled("clientLogging")) {
   DevToolsUtils.dumpn.wantLogging = true;
 }
 
+log("Debugger bootstrapping");
```
jasonLaster: not sure if we really need this log :)

public/js/util/utils.js
```diff
@@ -133,6 +134,14 @@ function compose(...funcs) {
   };
 }
 
+function log() {
+  if (!isDevelopment()) {
+    return;
+  }
```
jasonLaster: does the equivalent thing to NODE_ENV ... but wrapped in a convenience function

webpack.config.devtools.js
```diff
@@ -8,7 +8,7 @@ webpackConfig.output.library = "Debugger";
 webpackConfig.plugins = [
   new webpack.DefinePlugin({
     "process.env": {
-      NODE_ENV: JSON.stringify("DEVTOOLS_PANEL"),
+      NODE_ENV: JSON.stringify("production"),
```
jasonLaster: changed this so that many libraries will work with the production mentality, we might also want a different flag to differentiate between the panel and the standalone app, but for now... should be fine

config/config.js
```diff
@@ -4,7 +4,7 @@ const _ = require("lodash");
 const fs = require("fs");
 const path = require("path");
 
-const envConfig = process.env.NODE_ENV === "DEVTOOLS_PANEL" ?
+const envConfig = process.env.NODE_ENV === "production" ?
```
jlongster: I don't want to conflate the pane with production because dev/prod mode is something all libs use to turn on prod mode, and I'm sure there will be times where we want to do `NODE_ENV=production npm start` to try it in prod mode locally.

jasonLaster: that makes sense.

webpack.config.devtools.js
jlongster: Hm, I do see that using NODE_ENV in general is conflating it with the production meaning that libraries use. Because we load in the config below, and I can still just do `webpack --config webpack.config.devtools.js`, we can just use the config as the flag in `isFirefoxPanel` and never use NODE_ENV. That way NODE_ENV can be used normally: "production" turns on optimizations in libraries.

It might be nice to do turn on dev mode in the firefox panel build if we want to debug something there, but I'm ok forcing production here.

config/config.js
jlongster: Thinking this through a little more (see my other comment) I'm not sure what the best solution is but it would be nice to be able to run in prod mode locally.

PR #323, opened 2016-07-13
Use Firefox WebSocket transport instead of the WS/TCP proxy

jsnajdr:
Turn on the Firefox WebSocket debugger server and use it instead of the proxy:
- enable the `devtools.debugger.remote-use-websocket` pref
- comment out spawning the proxy
- change the `webSocketPort` to 6080 - we are connecting directly to Firefox now

Your Firefox needs to be built with the WebSocket patch from [bug 1286281](https://bugzilla.mozilla.org/show_bug.cgi?id=1286281).

### DIFF
--- a/bin/development-server.js
+++ b/bin/development-server.js
@@ -18,8 +18,10 @@ const feature = require("../config/feature");
 const config = getConfig();
 feature.setConfig(config);
 
-// const firefoxProxy = require("./firefox-proxy");
-// firefoxProxy({ logging: feature.isEnabled("logging.firefoxProxy") });
+if (!feature.isEnabled("firefox.webSocketConnection")) {
+  const firefoxProxy = require("./firefox-proxy");
+  firefoxProxy({ logging: feature.isEnabled("logging.firefoxProxy") });
+}
 
 function httpGet(url, onResponse) {
   return http.get(url, (response) => {
--- a/bin/development-server.js
+++ b/bin/development-server.js
@@ -18,8 +18,8 @@ const feature = require("../config/feature");
 const config = getConfig();
 feature.setConfig(config);
 
-const firefoxProxy = require("./firefox-proxy");
-firefoxProxy({ logging: feature.isEnabled("logging.firefoxProxy") });
+// const firefoxProxy = require("./firefox-proxy");
+// firefoxProxy({ logging: feature.isEnabled("logging.firefoxProxy") });
 
 function httpGet(url, onResponse) {
   return http.get(url, (response) => {
--- a/bin/firefox-driver.js
+++ b/bin/firefox-driver.js
@@ -21,6 +21,7 @@ function firefoxProfile() {
   profile.setPreference("devtools.debugger.remote-enabled",  true);
   profile.setPreference("devtools.chrome.enabled",  true);
   profile.setPreference("devtools.debugger.prompt-connection",  false);
+  profile.setPreference("devtools.debugger.remote-use-websocket", true);
 
   return profile;
 }
--- a/bin/firefox-driver.js
+++ b/bin/firefox-driver.js
@@ -7,6 +7,7 @@ const until = webdriver.until;
 const Key = webdriver.Key;
 
 const shouldStart = process.argv.indexOf("--start") > 0;
+const useWebSocket = process.argv.indexOf("--websocket") > 0;
 
 function firefoxBinary() {
   var binary = new firefox.Binary();
@@ -21,7 +22,7 @@ function firefoxProfile() {
   profile.setPreference("devtools.debugger.remote-enabled",  true);
   profile.setPreference("devtools.chrome.enabled",  true);
   profile.setPreference("devtools.debugger.prompt-connection",  false);
-  profile.setPreference("devtools.debugger.remote-use-websocket", true);
+  profile.setPreference("devtools.debugger.remote-use-websocket", useWebSocket);
 
   return profile;
 }
--- a/config/development.json
+++ b/config/development.json
@@ -12,6 +12,8 @@
     "webSocketPort": 9222
   },
   "firefox": {
+    "proxyPort": 9000,
+    "webSocketConnection": true,
     "webSocketPort": 6080
   }
 }
--- a/config/development.json
+++ b/config/development.json
@@ -12,6 +12,6 @@
     "webSocketPort": 9222
   },
   "firefox": {
-    "webSocketPort": 9000
+    "webSocketPort": 6080
   }
 }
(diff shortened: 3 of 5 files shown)


jasonLaster:
Very cool @jsnajdr. Mind wrapping this work in a debugger.html feature flag? 

Our flags are in [development.json](https://github.com/jlongster/debugger.html/blob/master/config/development.json)
and are called with `feature.isEnabled`.

bin/development-server.js
```diff
@@ -18,8 +18,8 @@ const feature = require("../config/feature");
 const config = getConfig();
 feature.setConfig(config);
 
-const firefoxProxy = require("./firefox-proxy");
-firefoxProxy({ logging: feature.isEnabled("logging.firefoxProxy") });
+// const firefoxProxy = require("./firefox-proxy");
```
jasonLaster: It'd be nice to wrap this in a feature flag:
```js
if (feature.isEnabled("firefox.websocketConnection")) {
}
```

bin/firefox-driver.js
```diff
@@ -21,6 +21,7 @@ function firefoxProfile() {
   profile.setPreference("devtools.debugger.remote-enabled",  true);
   profile.setPreference("devtools.chrome.enabled",  true);
   profile.setPreference("devtools.debugger.prompt-connection",  false);
+  profile.setPreference("devtools.debugger.remote-use-websocket", true);
```
jasonLaster: This should have a cli flag like `shouldStart` above

```js
 const useWebsocket = process.argv.indexOf("--websocket") > 0;

if("useWebsocket") {
}
```

config/development.json
```diff
@@ -12,6 +12,6 @@
     "webSocketPort": 9222
   },
   "firefox": {
-    "webSocketPort": 9000
+    "webSocketPort": 6080
```
jasonLaster: This field is currently not used and we should probably just delete it :)

jsnajdr: Are you sure? I see it used [here](https://github.com/jlongster/debugger.html/blob/master/public/js/clients/firefox.js#L52). How else would the debugger client know to which port it should connect?

jasonLaster: oh nice. sorry, missed that. In that case I think we should have two fields
```js
firefoxProxyPort: 9000,
firfoxWebsocketPort: 6080
```

and feature flag at that location so the appropriate one is used

jsnajdr:
Added a feature flag as discussed above:
- `firefox.webSocketConnection` to enable the direct WebSocket connection without a proxy
- `proxyPort` and `webSocketPort` prefs to configure the ports separately
- `--websocket` cli option for the `firefox-driver` script

Also, I'm now using a new `WebSocketDebuggerTransport` to connect to the remote server. It doesn't use the length prefix and sends the JSON strings directly as WebSocket text frames. See how simple it is compared to the classic `DebuggerTransport` when it doesn't have to deal with all the messy details and just leverages the WebSocket capabilities.

You need to use an up-to-date patch from [bug 1286281](https://bugzilla.mozilla.org/show_bug.cgi?id=1286281) that supports the new transport.

jasonLaster:
Thanks @jsnajdr.

The test failures are coming from our integration tests that do not have latest patch. 

I think this feature makes sense off by default. That way contributors can opt-in if they want to run off of head.

public/js/clients/firefox.js
```diff
@@ -49,10 +50,13 @@ function createTabs(tabs) {
 function connectClient() {
   const deferred = defer();
   let isConnected = false;
-  const webSocketPort = getValue("firefox.webSocketPort");
+  const useProxy = !getValue("firefox.webSocketConnection");
```
jasonLaster: `isEnabled` would probably be cleaner, but not a big deal.

jsnajdr: Well, the `getValue` symbol is already imported in the file. I don't know why it has two aliases :)

jsnajdr:
Of course, I forgot to disable `webSocketConnection` before committing and pushing. Coming soon in the next commit.

public/js/lib/devtools-sham/transport/ws-transport.js
```diff
@@ -0,0 +1,51 @@
...
+    this.active = false;
+    this._ws.close();
+    if (this.hooks) {
+      this.hooks.onClosed(reason);
+      this.hooks = null;
+    }
+  },
+};
+
+module.exports = WebSocketDebuggerTransport;
```
jasonLaster: This is great. Is this transport layer also in M-C? If so, did you change it all to land it here? I'm asking because we have a couple of rules on where M-C code should live?

If it's unchanged: `js/lib/devtools/<path-to-file>`
If it was changed: `js/lib/devtools-sham/<path-to-file>`

If it's not in m-c: `js/client/firefox/websocketTransport.js`

jsnajdr: It's a part of my patch in [bug 1286281](https://bugzilla.mozilla.org/show_bug.cgi?id=1286281), but is slightly different: the one here uses the standard WebSocket API, the other one in Firefox tree uses my own WebSocket class that is based on XPCOM streams and has a slightly different API.

From what you say, `devtools-sham` looks like the right place.

jlongster: Why can't the client in m-c use the standard WebSocket API?

jsnajdr: Several little things. Hopefully all can be eventually resolved.

The DOM WebSocket client (i.e., `new WebSocket("ws://localhost:6080")` can't be used, because the debugger client does some TLS magic with checking client and server certificates. That's implemented using the nsISocketTransport APIs. If we can do the same thing on the DOM WebSocket, we can migrate to it.

The event handling API is different. DOM WebSocket uses `socket.onmessage = ...`, the node.js EventEmitter (used in the `ws` module) is different from devtools EventEmitter.

Then there are differences in handling binary data. Blobs, ArrayBuffers, Arrays of ints, typed arrays, node's Buffers...

jasonLaster: Thanks. This is fine for merging. @jlongster I think we can safely move this to `client/firefox/websocketTransport` if m-c eventually has something we can use we can switch then.

jasonLaster:
Thanks @jsnajdr!

jsnajdr:
@jasonLaster I've been actually playing with debugger.html for several weeks prior to working on this patch, and it went really well - you just call `npm install` and `npm start` and you're up and going.

The only small trouble I had was figuring out how to start Firefox with the debugging server on the right port. I wanted to use the version I built myself and avoided your Firefox startup scripts, so it's not your problem.

jasonLaster:
@jsnjdr great. We're actually updating the docs too make it more clear how to set your local profile for Firefox.

PR #325, opened 2016-07-13
Flow

jlongster:
I'm so sorry for the size of this PR.

Flow makes us use babel to strip the types, and that tends to have a cascading effect. I'm not sure if we'll be able to avoid babel in the long run though. The immediate thing that comes to mind is JSX. If we are ok with a build step (which we are), we'll probably end up using that.

The main annoyance was that the linter was complaining now because the "use strict" directive is redundant; all files are passed through babel which forces strict mode. So the majority of the files changes is just removing all strict directives.

The interesting parts are the files that I annotated with flow:

```
./public/js/actions/breakpoints.js
./public/js/actions/types.js
./public/js/reducers/sources.js
./public/js/reducers/tests/sources.js
./public/js/selectors.js
./public/js/util/fromJS.js
./public/js/util/makeRecord.js
./public/js/util/test-head.js
```

I wanted to have one of each type of file annotated: a reducer, an action creator, the selectors, and a test. For example, if you go into `selectors.js`, in `getSources`, and introduce a typo like `return state.sources.sources2`, you will see a type error when you run `flow` (or `npm run flow`). There are several places in the above files where this will happen now.

I went through many attempts at the initial integration, so some things might still be leftover from that. There are still several places in these files that aren't type-checked, and we need to figure out how to check everything in them, but this PR is already large enough.

Another example is going into `public/js/reducers/tests/sources.js` and changing the action type to a typo like `ADD_SOURCE2`. You'll see that even the action objects are fully type-checked; they need to conform to a specific action type specified in `actions/types.js`.


jlongster:
The hardest part about this was trying to type immutable.js, which is certainly possible but required me to learn enough flow before I could get something working. You'll see `makeRecord.js` which is only partially used, which will fully type immutable records. We can't use it fully until 0.29 comes out with a serious perf fix (should come out this week). Right now, the state in `reducers/sources.js` isn't type-checked much because the Record type returns `any` from it's methods.

It also encourages us to use records so that we *can* type check it. Switching the sources state to be a record introduced other problems, like not being able to use a naive `fromJS` anymore (which we keep running into). I wrote a `dehydrate` function which explicitly knows how to dehydrate a JSON-based app state into our immutable one. We were already using our own `fromJS` before anyway. There is probably a better long-term solution for this.

jlongster:
karma and cypress are still broken, and I bet storybook is too (all that storybook should need is to use `dehydrate` instead of `fromJS`.

jlongster:
There are 2 commits and if you look at the first one you'll see the most interesting bits. The second one fixes all the linting issues and other nonsense.

public/js/test/integration/fixtures.js
```diff
@@ -1,6 +1,6 @@
 // turn theses tests on when you want to write new fixture data
 // you also need to turn on the cypress-server to be able to save the fixtures.
-xdescribe("Fixtures", function() {
+describe("Fixtures", function() {
```
jasonLaster: re-add `xdescribe` and cypress should pass.

jasonLaster:
it looks like karma has some functions like `getSources` which are undefined. Is that related to dehydrate and/or selectors?

public/js/actions/breakpoints.js
```diff
@@ -1,21 +1,29 @@
...
-"use strict";
 
 const constants = require("../constants");
 const { PROMISE } = require("../util/redux/middleware/promise");
 const { getBreakpoint, getBreakpoints } = require("../selectors");
 const { Task } = require("../util/task");
 const fromJS = require("../util/fromJS");
 
-function enableBreakpoint(location) {
+import type { Location } from "./types";
```
jasonLaster: Is it possible to use commonjs here? or is there something about types that means we need to use import?

public/js/actions/breakpoints.js
```diff
@@ -24,8 +32,16 @@ function _getOrCreateBreakpoint(state, location, condition) {
...
-function addBreakpoint(location, { condition, getTextForLine } = {}) {
-  return ({ dispatch, getState, client }) => {
+type AddBreakpointOpts = {
+  condition?: string,
+  getTextForLine?: (line: number) => string
+};
+
+function addBreakpoint(
+  location: Location,
+  { condition, getTextForLine } : AddBreakpointOpts = {}
```
jasonLaster: Maybe it's because I'm new to this, but `AddBreakpointOpts` seems unhelpful and adds a lot of clutter

public/js/actions/breakpoints.js
```diff
@@ -37,10 +53,11 @@ function addBreakpoint(location, { condition, getTextForLine } = {}) {
       breakpoint: bp.toJS(),
       condition: condition,
       [PROMISE]: Task.spawn(function* () {
-        const { id, actualLocation } = yield client.setBreakpoint(
+        const res: any = yield client.setBreakpoint(
```
jasonLaster: This also seems unhelpful. It might be nice to type the client, but on the other hand I also don't expect it to change much and I appreciate being able to destructure on assignments.

public/js/actions/index.js
```diff
@@ -24,7 +22,7 @@ function navigate() {
   };
 }
 
-module.exports = Object.assign(
-  { willNavigate, navigate },
+module.exports = (Object.assign(
+  ({ willNavigate, navigate }: any),
   breakpoints, eventListeners, sources, tabs, pause
-);
+) : typeof breakpoints);
```
jasonLaster: what does `typeof breakpoints` do?

public/js/actions/tests/loadSourceText.js
```diff
@@ -86,28 +85,20 @@ describe("loadSourceText", () => {
...
-    const store = createStore(deferredMockThreadClient);
+  it("source failed to load", function() {
+    return Task.spawn(function* () {
+      const store = createStore(deferredMockThreadClient);
+      yield store.dispatch(loadSourceText({ id: "badId" })).catch(() => {});
 
-    Task.spawn(function* () {
-      yield loadBadSource(store);
-      const fooSourceText = getSourceText(store.getState(), "foo1");
+      const fooSourceText = getSourceText(store.getState(), "badId");
```
jasonLaster: nice

public/js/actions/types.js
```diff
@@ -0,0 +1,38 @@
+// @flow
+
+export type AsyncStatus = "start" | "done" | "error";
```
jasonLaster: nice

public/js/util/makeRecord.js
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * @param spec - the keys and their default values @return a state
+ * record factory function
+ */
+function makeRecord<T>(
+  spec: T & Object,
+  name: string
+): (init: $Shape<T>) => Record<T> {
+  return I.Record(spec, name);
+}
```
jasonLaster: cool

jasonLaster:
Overall, this looks great. A couple small style nits. I'd prefer and start small and use the features that are clearly helpful. There were some examples where flow added annotations that were verbose, and it was hard to see huge benefits.

public/js/test/integration/fixtures.js
jlongster: Oh! Hah I thought I had added a typo.

jlongster:
> it looks like karma has some functions like getSources which are undefined. Is that related to dehydrate and/or selectors?

I'm not sure, but I bet it's my fault, I don't think I saw those errors in the previous flaky errors. I will look into those.

public/js/actions/breakpoints.js
jlongster: Nope, the syntax for types is only ES6 :/ But that's because CommonJS doesn't really have a way to do this. You'd have to special syntax for CommonJS, which starts getting weird because usually `require` is just an expression. So I get their design choice.

ES6 is the future anyway. Since we're using a build step, I wouldn't be surprised if we converted to ES6 modules at some point. But I don't think mixing the styles is too bad.

public/js/actions/breakpoints.js
jlongster: It definitely does. I just wanted to get this PR out. It's probably because I'm new to flow that this is verbose. We'll figure out the right balance over time, and I'll look at this again before merging.

public/js/actions/breakpoints.js
jlongster: Destructuring is supported, I just ran into an issue here and I can't remember what it was. I didn't spend much time on this part of the file, just had to get something working.

public/js/actions/index.js
jlongster: It statically gives the type system whatever type `breakpoints` is. This tells the system that what I'm returning has the exact structure of the `breakpoints` object, so it lets you do this dynamic `Object.assign` munging but still type it.

public/js/actions/breakpoints.js
jasonLaster: makes sense.

jlongster:
> Overall, this looks great. A couple small style nits. I'd prefer and start small and use the features that are clearly helpful. There were some examples where flow added annotations that were verbose, and it was hard to see the benefits being worthwhile.

Yeah, we'll figure out a balance over time. I think it's a good goal to type check as much as possible. The problem with holes is that they tend to "bleed out", but there are places where it's OK. The worst parts here was in parts of `breakpoints.js` where I did things quickly and didn't think much of it. There are probably more elegant solutions.

> I'd like to merge this soon, but I also don't want to add it if we break karma and cypress. It looks like the cypress fix is easy though. I'm looking into karma now on master so i'll have a better idea where the flakiness is coming from.

Yeah, let's not merge it until I fix it up a little more. I'll look into karma too because I'm pretty sure this error is my fault.

PR #326, opened 2016-07-13
Show original files in sources tree

### LINKED ISSUE #306: Create sources with the SourceMapConsumer

jasonLaster:
Fixes #306 and #304. 

I'll look into showing sources in the editor next. Also, the `source-map` utility is still being thought through. I'm working on using the smallest APIs so that i can explore the problem.

Note, the domain is not properly parsed in some cases. I can fix this later

### DIFF
--- a/public/js/actions/sources.js
+++ b/public/js/actions/sources.js
@@ -8,12 +8,13 @@ const { PROMISE } = require("../util/redux/middleware/promise");
 const { Task } = require("../util/task");
 const { isJavaScript } = require("../util/source");
 
-const { getSource, getSourceText } = require("../selectors");
+const { getSource, getSourceText, getSelectedTab } = require("../selectors");
 const constants = require("../constants");
 const Prefs = require("../prefs");
 const invariant = require("invariant");
 const { isEnabled } = require("../../../config/feature");
-const { networkRequest } = require("../util/networkRequest");
+const { Source } = require("../types");
+const { loadSourceMap, getOriginalSources } = require("../util/source-map");
 
 /**
  * Throttles source dispatching to reduce rendering churn.
@@ -37,10 +38,10 @@ function queueSourcesDispatch(dispatch) {
  * Handler for the debugger client's unsolicited newSource notification.
  */
 function newSource(source) {
-  return ({ dispatch }) => {
-    if (isEnabled("features.sourceMaps") && source.sourceMapUrl) {
-      networkRequest(source.sourceMapUrl).then(res => {
-      });
+  return ({ dispatch, getState }) => {
+    if (isEnabled("features.sourceMaps") && source.sourceMapURL) {
+      const tab = getSelectedTab(getState());
+      loadSourceMapSources(tab, source, dispatch);
     }
 
     // We don't immediately dispatch the source because several
@@ -49,6 +50,19 @@ function newSource(source) {
     newSources.push(source);
     queueSourcesDispatch(dispatch);
   };
+}
+
+function loadSourceMapSources(tab, source, dispatch) {
+  loadSourceMap(tab, source).then(map => {
+    const originalUrls = getOriginalSources(map);
+    let inc = 1;
+
+    originalUrls.forEach(url => dispatch(newSource(Source({
+      url,
+      id: `${source.id}/${++inc}`,
+      isPrettyPrinted: false
+    }))));
+  });
 }
 
 function selectSource(id, options = {}) {
--- a/public/js/util/source-map.js
+++ b/public/js/util/source-map.js
@@ -1,0 +1,29 @@
+"use strict";
+
+const { networkRequest } = require("./networkRequest");
+const { SourceMapConsumer } = require("source-map");
+
+let sourceMaps = new Map();
+
+function loadSourceMap(tab, source) {
+  const sourceMapURL = `${tab.get("url")}/${source.sourceMapURL}`;
+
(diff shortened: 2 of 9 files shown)


public/js/util/source-map.js
```diff
@@ -0,0 +1,29 @@
+"use strict";
```
jlongster: This is fine for now to get something started (sourcemaps are preffed off by default right?) but this will leak sourcemaps across navigations (and potentially load wrong sourcemaps if ids are the same across navigations of different sites).

I think it's going to be more trouble than it's worth to try and avoid storing them in the redux state. `loadSourceMap` should be an action creator that does this and checks the state to see if it exists. We gain the ability to faithfully snapshot what's really happening (and see all the sourcemap loading actions), and access the souremaps in tests. We'll have to recreate all of that if we don't store it in redux.

There may be times when we want to avoid keeping the sourcemaps in a state snapshot (because it potentially could be really big), but I think we need to be smarter about hydrating/dehydrating the state anyway.

There are problems with putting it in the state: we'll need to be smart about the hyrdating/dehydrating the `SourceMapConsumer` instance. But as I said above, I think we need a better way handle that anyway (see the transit library).

This is fine for now, not going to block merging this, but just wanted to discuss it.

public/js/util/source-map.js
```diff
@@ -0,0 +1,29 @@
...
+    return Promise.resolve(sourceMaps.get(source.id));
+  }
+
+  return networkRequest(sourceMapURL).then(sourceMap => {
+    sourceMaps.set(source.id, sourceMap);
+    return sourceMap;
+  });
+}
+
+function getOriginalSources(sourceMap) {
```
jlongster: We might want to go ahead and nix the generated/original terms, and come up with something better.

public/js/util/source-map.js
```diff
@@ -0,0 +1,29 @@
...
+
+function loadSourceMap(tab, source) {
+  const sourceMapURL = `${tab.get("url")}/${source.sourceMapURL}`;
+
+  if (sourceMaps.has(source.id)) {
+    return Promise.resolve(sourceMaps.get(source.id));
+  }
+
+  return networkRequest(sourceMapURL).then(sourceMap => {
+    sourceMaps.set(source.id, sourceMap);
```
jlongster: Why not create the `SourceMapConsumer` here? You don't want to recreate it every time you access it, as it parses the sourcemap. I'd change that before merging this in.

public/js/actions/sources.js
```diff
@@ -51,6 +52,19 @@ function newSource(source) {
   };
 }
 
+function loadSourceMapSources(tab, source, dispatch) {
+  loadSourceMap(tab, source).then(map => {
+    const originalUrls = getOriginalSources(map);
+    let inc = 1;
```
jlongster: A UUID might be better so it doesn't depend on any state (not a blocker for now)

public/js/actions/sources.js
```diff
@@ -51,6 +52,19 @@ function newSource(source) {
...
+
+    originalUrls.forEach(url => dispatch(newSource(Source({
+      url,
+      id: `${source.id}/${++inc}`,
+      isPrettyPrinted: false
+    }))));
```
jlongster: Nice, this is much better and is the way to do it

public/js/actions/sources.js
```diff
@@ -51,6 +52,19 @@ function newSource(source) {
   };
 }
 
+function loadSourceMapSources(tab, source, dispatch) {
```
jlongster: This lives in an actions file but isn't an action creator. Don't worry about it for now, but every function in here should return something that is dispatched. But we may want to have sourcemap loading actions, as mentioned in my other comment, so this would become an action creator. Anyway, not a problem right now.

jlongster:
It's looking nice, there are things to discuss but this can merged. I'd like to see two things before it's merged: cache the actual `SourceMapConsumer` instance (so it'll only be parsed once), and rethink the "original" terms. I want to make sure we don't get stuck on original/generated again.

public/js/util/source-map.js
jasonLaster: this is changed in a subsequent PR :)

public/js/actions/sources.js
jasonLaster: not sure if i follow - the uniqueness comes from the generated id.

public/js/actions/sources.js
jasonLaster: this is moved in a later pr... lets discuss then

jasonLaster:
Thanks. This work is largely improved by the next PR, which touches the editor. Lets discuss terms there.

public/js/util/source-map.js
jasonLaster: > this will leak sourcemaps across navigations (and potentially load wrong sourcemaps if ids are the same across navigations of different sites).

That's a good point. I'll add it to the tracking ticket.

To summarize, we could keep the maps in the store and access the cached consumers/nodes from that.

public/js/actions/sources.js
jlongster: The generated id is a sequential id, not a uuid. A uuid is guaranteed to never conflict, so if in a test we load a snapshot of the state and add a new sourcemapped source, that will still work, whereas here it will potentially conflict because `inc` will start from scratch.

PR #327, opened 2016-07-14
Only handle wheel events in firefox

### LINKED ISSUE #145: Editor scrolling relies on scrollBy
Chrome scrollling was temporarily broken due to not supporting `scrollBy`.

This commit fixed it, https://github.com/jlongster/debugger.html/commit/9164c084f64a3e12f126b6a608772e801a2e784e, but opens the question of whether we want to use an API that only works in firefox.

Moreover, when i tested the editor scrolling with the trackpad in `jquery.js` i could not find a slow down, which the @gabrielluong bug was solving.

jasonLaster:
Fixes #145

### DIFF
--- a/config/feature.js
+++ b/config/feature.js
@@ -23,6 +23,10 @@ function isFirefoxPanel() {
   return DebuggerTarget == "firefox-proxy";
 }
 
+function isFirefox() {
+  return /firefox/i.test(navigator.userAgent)
+}
+
 // only used for testing purposes
 function setConfig(stub) {
   config = stub;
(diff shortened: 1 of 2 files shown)


config/feature.js
```diff
@@ -23,6 +23,10 @@ function isFirefoxPanel() {
   return DebuggerTarget == "firefox-proxy";
 }
 
+function isFirefox() {
+  return /firefox/i.test(navigator.userAgent)
```
jlongster: Are you sure this won't have any false positives? I feel like I've seen checks that are a lot more complicated than this.

jlongster:
Looks good, I hope we don't use UA-based checks often but it's fine for this fix.

config/feature.js
jasonLaster: perhaps. but i think our usecase if fairly limited. honestly, i wouldn't be surprised if we improve this later

PR #330, opened 2016-07-14
Show original sources in editor

jasonLaster:
This work builds on the initial PR. It was done before you reviewed the first one so there are things that you commented on that I have not touched yet.

### DIFF
--- a/public/js/actions/sources.js
+++ b/public/js/actions/sources.js
@@ -13,8 +13,8 @@ const constants = require("../constants");
 const Prefs = require("../prefs");
 const invariant = require("invariant");
 const { isEnabled } = require("../../../config/feature");
-const { Source } = require("../types");
-const { loadSourceMap, getOriginalSources } = require("../util/source-map");
+const { loadOriginalSources, isOriginal,
+        getOriginalSource } = require("../util/source-map");
 
 /**
  * Throttles source dispatching to reduce rendering churn.
@@ -41,7 +41,8 @@ function newSource(source) {
   return ({ dispatch, getState }) => {
     if (isEnabled("features.sourceMaps") && source.sourceMapURL) {
       const tab = getSelectedTab(getState());
-      loadSourceMapSources(tab, source, dispatch);
+      loadOriginalSources(tab, source)
+        .then(sources => sources.forEach(s => dispatch(newSource(s))));
     }
 
     // We don't immediately dispatch the source because several
@@ -50,19 +51,6 @@ function newSource(source) {
     newSources.push(source);
     queueSourcesDispatch(dispatch);
   };
-}
-
-function loadSourceMapSources(tab, source, dispatch) {
-  loadSourceMap(tab, source).then(map => {
-    const originalUrls = getOriginalSources(map);
-    let inc = 1;
-
-    originalUrls.forEach(url => dispatch(newSource(Source({
-      url,
-      id: `${source.id}/${++inc}`,
-      isPrettyPrinted: false
-    }))));
-  });
 }
 
 function selectSource(id, options = {}) {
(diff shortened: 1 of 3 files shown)


public/js/actions/sources.js
```diff
@@ -41,7 +41,8 @@ function newSource(source) {
   return ({ dispatch, getState }) => {
     if (isEnabled("features.sourceMaps") && source.sourceMapURL) {
       const tab = getSelectedTab(getState());
-      loadSourceMapSources(tab, source, dispatch);
+      loadOriginalSources(tab, source)
+        .then(sources => sources.forEach(s => dispatch(newSource(s))));
```
jasonLaster: it'd be nice to update this to load the source map as an action. Then getting the original sources can be done in the utility

```js
const sourceMap = yield loadSourceMap(source);
const sources = yield loadOriginalSources(sourceMap);
sources.forEach((s) => dispatch(newSource)));
```

public/js/util/source-map.js
```diff
@@ -1,29 +1,112 @@
...
+const invariant = require("invariant");
+const { SourceMapConsumer, SourceNode } = require("source-map");
+const { Source } = require("../types");
+const { Task } = require("./task");
+const { getSource } = require("../selectors");
+const first = require("lodash/first");
 
-let sourceMaps = new Map();
+let sourceMapConsumers = new Map();
+let sourceNodes = new Map();
```
jasonLaster: I had already changed the source map to a consumer map. Will be easy to store the maps in the store

public/js/util/source-map.js
```diff
@@ -1,29 +1,112 @@
...
+function getOriginalSources(sourceId) {
+  const consumer = getConsumer(sourceId);
+  return consumer.sources;
+}
+
+function isOriginal(originalSource) {
+  return !!getGeneratedSourceId(originalSource)
+}
+
+function getGeneratedSourceId(originalSource) {
```
jasonLaster: Really happy with this method. This is the third attempt at mapping generated => original. The first attempt was to add a generatedSourceId to all sources. The second was to create an original source id to generated source id map. This is a simple approach that re-uses the abstractions we already have.

public/js/util/source-map.js
```diff
@@ -1,29 +1,111 @@
...
+  return sourceMapConsumers.get(sourceId);
+}
+
+function getOriginalSources(sourceId) {
+  const consumer = getConsumer(sourceId);
+  return consumer.sources;
+}
+
+function isOriginal(originalSource) {
+  return !!getGeneratedSourceId(originalSource);
```
jlongster: It might be better to come up with a sourceId format that we can simply match here. That way we don't have to iterate all the consumers and all the sources in each of them every time you click on a source.

jlongster:
Looks good, it probably would have been nice to merge this with the previous PR so I reviewed it together :) But this is coming along nicely. The only thing I don't like is the internal state of consumers, and needing to clear that out on navigation.

Another idea is to do what you are doing now, but instead of keeping them in a Map, keep them in a WeakMap keyed on the source object. The source object isn't something we ever update, so it should always exist as the same instance. This way you wouldn't have to ever care about managing the `Map` itself, i.e. clearing it.

public/js/util/source-map.js
jasonLaster: That makes sense. 

Honestly, I'd be okay doing something like chrome's object id's which are json strings.

id: `{generatedSourceId: "...", originalSourceId: "2"}`

jlongster: I think that's kind of confusing to see stringified objects when debugging...  It would be just as useful to concat the generatedSourceId and originalSourceId fields into something like `id1-id2`. The rest of the stringified object is just clutter.

The id could simply like `generatedID/source-mapped-sourceID`. Another approach would be to add an `isOriginal` (or `isVirtual` or `isFake`) flag right onto the source object.

PR #333, opened 2016-07-14
(WIP) Redux devtools

jasonLaster:
quick stab at adding redux devtools, with a dock and monitor.

todo:
+ make it work
+ add feature flag 
+ add console monitor w/ another feature flag


public/js/util/create-store.js
```diff
@@ -46,10 +62,10 @@ module.exports = (opts = {}) => {
...
   }
 
-  // Optionally add the Redux DevTools Extension, if it exists.
-  const devtools = typeof window === "object" && window.devToolsExtension ?
-    window.devToolsExtension() :
-    f => f;
+  // // Optionally add the Redux DevTools Extension, if it exists.
+  // const devtools = typeof window === "object" && window.devToolsExtension ?
+  //   window.devToolsExtension() :
+  //   f => f;
```
jlongster: I don't think we should uncomment this. You should probably check for this first, and if that doesn't exist, use `DevTools`. This lets people with the hook into the app with the redux browser addon.

public/js/util/create-store.js
```diff
@@ -4,13 +4,29 @@
 /* global window */
 "use strict";
 
+const React = require("react");
```
jlongster: I don't see where React is used?

jlongster:
I think this should be configurable. The `configureStore` function should take another argument like `devtools` that defaults to false. I'm not going to use this much of the time, and it has a pretty significant performance impact.

jasonLaster:
@jlongster ready to go

config/development.json
```diff
@@ -4,6 +4,9 @@
     "client": false,
     "firefoxProxy": false
   },
+  "redux-devtools": {
+    "enabled": false
```
jlongster: Why not just `reduxDevtools: true`? Also camel-cased because we use that for all other options

public/js/util/create-store.js
```diff
@@ -50,5 +67,13 @@ module.exports = (opts = {}) => {
     window.devToolsExtension() :
     f => f;
 
-  return applyMiddleware(...middleware)(devtools(createStore));
+  return compose(
+    applyMiddleware(...middleware),
+    DevTools.instrument()
```
jlongster: You need to flag this off, because this is where all the perf problems are. Check if the `reduxDevtools` flag is enabled here and only instrument it if so.

I'd say that you should only export `DevTools` if it's enabled also, and the in the `App` render function just do `DevTools ? DevTools() : null` and avoid the `DevToolsPanel` function.

config/development.json
jasonLaster: I'd like to make space for other loggers to be used.

public/js/util/create-store.js
jasonLaster: nice catch

PR #340, opened 2016-07-14
convert original to mapped

jasonLaster:
Small naming change. The idea is to talk about sources in the action layer in terms of sources that either come from the server or are mapped client side.

e.g. bundle.js is a source that has 3 mapped sources math.js, increment.js, sum.js

Not sure if this is the best naming or works in all of the cases.


jasonLaster:
closing. talked with @fitzgen in IRC. generated/original are terms used in the spec, so we should be consistent.

jlongster:
Sounds good

PR #341, opened 2016-07-14
add return/exception/this values to innermost scope

jlongster:


jlongster:
I'm about to write tests for things but I wanted to land a few features in this sprint. We should talk a little bit about component unit tests next week because I have some ideas.

jasonLaster:
Looks good.

PR #342, opened 2016-07-14
Move source maps to the store

jasonLaster:


public/js/actions/sources.js
```diff
@@ -52,6 +52,30 @@ function newSource(source) {
   };
 }
 
+function loadSourceMap(generatedSource) {
+  return ({ dispatch, getState }) => {
+    let sourceMap = getSourceMap(getState(), generatedSource.id);
+    if (sourceMap) {
+      return {};
```
jlongster: you can just do `return`. The thunk doesn't have to return anything.

public/js/actions/sources.js
```diff
@@ -52,6 +52,30 @@ function newSource(source) {
...
+    }
+
+    return dispatch({
+      type: constants.ADD_SOURCE_MAP,
```
jlongster: `LOAD_SOURCE_MAP` might be a better description of this, analogous to `LOAD_SOURCE_TEXT`.

public/js/actions/sources.js
```diff
@@ -52,6 +52,30 @@ function newSource(source) {
...
+      source: generatedSource,
+      [PROMISE]: (async function () {
+        const sourceMapURL = getSourceMapURL(getState(), generatedSource);
+        sourceMap = await networkRequest(sourceMapURL);
+
+        setSourceMap(generatedSource, sourceMap);
+        createOriginalSources(generatedSource)
+          .forEach(s => dispatch(newSource(s)));
```
jlongster: Nice, that looks very good

jlongster:
This is looking really nice!

public/js/actions/sources.js
jasonLaster: hmm, was getting a "inconsistent return"

public/js/actions/sources.js
jasonLaster: agred

jasonLaster: i'll change this in the upcoming PR

public/js/actions/sources.js
jlongster: oh, really? Interesting, you can see in `selectSource` we just `return`, so not sure what was throwing that

PR #344, opened 2016-07-14
show the __proto__ value in the object inspector (fixes #319)

### LINKED ISSUE #319: show prototype
In the Variables View display the object prototype for expandable inspection.

jlongster:
**Don't merge this**. As I was typing this out I realized that this work should happening in the `ObjectInspector` component. I will probably make the `loadedObjects` map contain all the info needed to construct this data. I'll change this PR tomorrow.

### DIFF
--- a/public/js/components/ObjectInspector.js
+++ b/public/js/components/ObjectInspector.js
@@ -72,6 +72,29 @@ const ObjectInspector = React.createClass({
     return {};
   },
 
+  makeNodesForProperties(props, parentPath) {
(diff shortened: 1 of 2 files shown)


jlongster:
New implementation. This removes some of the burden of the reducer, making it easier to share `ObjectInspector` with other projects. We will probably rename this fields when we make this state more generic.

jasonLaster:
looks good. i like that the reducer is simplified.

PR #348, opened 2016-07-15
remove dependency on accordion styling for panes (fixes #146)

### LINKED ISSUE #146: Right sidebar components rely on accordian css for styling
The frame and breakpoints pane rely on additional accordion markup for styling.

This is  a problem because the components will look bad if they are anywhere else.

This storybook example shows the markup needed to render a call stack
https://github.com/jlongster/debugger.html/blob/master/public/js/components/stories/Frames.js#L38-L39

Current Bad Version:
```js
    createElement(Provider, { store },
      dom.div({ className: "accordion" },
        dom.div({ className: "_content" }, createElement(Component))))
```

When we've fixed the css, we should be able to remove the bad markup
```js
 createElement(Provider, { store }, createElement(Component))))
```

We can fix this by replacing bad rules with generic classes that can be applied:

https://github.com/jlongster/debugger.html/blob/master/public/js/components/RightSidebar.css#L31-L40
Here are the two bad rules:

```css

.accordion ._content {
  font-size: 0.75em;
  padding: 5px;
}

.accordion .pane-info .empty {
  font-style: italic;
  text-align: center;
}

```

The *fix* is to change `.accordion ._content` to a class like `.sidebar-pane` that can be applied directly to each component.

jlongster:


jasonLaster:
Looks great

PR #353, opened 2016-07-16
SVG Cleanup

svenvNL:
Made indention in all files the same, removed empty tags and removed Sketch comment.


jasonLaster:
Thanks this is great. 

If you're interested in cleaning up or svgs. I believe we're consolidating them in an icon js file so they can be required directly. 

It would be great to move these in there and then require the new react components. The advantage is that it avoids http requests.

jasonLaster:
Also, if you'd like to branch out from svgs we have this fun ticket https://github.com/jlongster/debugger.html/issues/338

PR #356, opened 2016-07-18
Source Maps - Breakpoints

jasonLaster:
#### Features
Supports setting/removing breakpoints in the editor. 
Toggling in the breakpoints pane. 
Navigating by selecting breakpoints.

#### Philosophy
I'll write a doc for source maps later, but the gist of it is this:
+ store keeps generated locations (locations that are relevant to the debugger server)
+ selectors translate locations to original locations (useful for ui)
+ actions translate locations to generated location (useful for server)
+ original sources are kept in the store as well...

### DIFF
--- a/public/js/actions/breakpoints.js
+++ b/public/js/actions/breakpoints.js
@@ -5,8 +5,9 @@
 
 const constants = require("../constants");
 const { PROMISE } = require("../util/redux/middleware/promise");
-const { getBreakpoint, getBreakpoints } = require("../selectors");
+const { getBreakpoint, getBreakpoints, getSource } = require("../selectors");
 const fromJS = require("../util/fromJS");
+const { isOriginal, getGeneratedSourceLocation } = require("../util/source-map");
 
 import type { Location } from "./types";
 
@@ -22,13 +23,49 @@ function enableBreakpoint(location: Location) {
   return addBreakpoint(location);
 }
 
+function isOriginalLocation(state, location: Location) {
+  const source = getSource(state, location.sourceId);
+  return isOriginal(source.toJS());
+}
+
+function getGeneratedLocation(state, location: Location) {
+  const source = getSource(state, location.sourceId);
+  if (isOriginal(source.toJS())) {
+    return getGeneratedSourceLocation(source.toJS(), location);
+  }
+
+  return location;
+}
+
 function _breakpointExists(state, location: Location) {
+  location = getGeneratedLocation(state, location);
   const currentBp = getBreakpoint(state, location);
   return currentBp && !currentBp.get("disabled");
 }
 
 function _getOrCreateBreakpoint(state, location, condition) {
+  location = getGeneratedLocation(state, location);
   return getBreakpoint(state, location) || fromJS({ location, condition });
+}
+
+function getBreakpointText({
+  breakpoint, getState, getTextForLine, actualLocation, location }) {
+  // If this breakpoint is being re-enabled, it already has a
+  // text snippet.
+  let text = breakpoint.get("text");
+  if (text) {
+    return text;
+  }
+
+  if (!getTextForLine) {
+    return "";
+  }
+
+  if (isOriginalLocation(getState(), location)) {
+    return getTextForLine(location.line);
+  }
+
+  return getTextForLine(actualLocation.line);
 }
 
 function addBreakpoint(location: Location,
@@ -50,12 +87,13 @@ function addBreakpoint(location: Location,
           bp.get("condition")
         );
 
-        // If this breakpoint is being re-enabled, it already has a
-        // text snippet.
-        let text = bp.get("text");
-        if (!text) {
-          text = getTextForLine ? getTextForLine(actualLocation.line) : "";
-        }
+        const text = getBreakpointText({
+          breakpoint: bp,
+          location,
+          actualLocation,
+          getTextForLine,
+          getState
+        });
 
         return { id, actualLocation, text };
       })()
--- a/public/js/actions/breakpoints.js
+++ b/public/js/actions/breakpoints.js
@@ -121,6 +121,11 @@ function _removeOrDisableBreakpoint(location, isDisabled) {
       throw new Error("attempt to remove unsaved breakpoint");
     }
 
+    bp = bp.set("location", getGeneratedLocation(
+      getState(),
+      bp.get("location").toJS()
+    ));
+
     const action = {
       type: constants.REMOVE_BREAKPOINT,
       breakpoint: bp.toJS(),
--- a/public/js/actions/sources.js
+++ b/public/js/actions/sources.js
@@ -14,7 +14,7 @@ const constants = require("../constants");
 const Prefs = require("../prefs");
 const invariant = require("invariant");
 const { isEnabled } = require("../../../config/feature");
-const { setSourceMap, createOriginalSources,
+const { setSourceMap, createOriginalSources, getGeneratedSourceId,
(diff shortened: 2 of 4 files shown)


public/js/actions/sources.js
```diff
@@ -184,6 +184,25 @@ function togglePrettyPrint(id) {
...
+        loadSourceText(generatedSource.toJS())
+      );
+
+      return getOriginalSource(
+        source,
+        generatedSource.toJS(),
+        generatedSourceText
+      );
+    })
+}
```
jasonLaster: you might be asking why the sources action is now loading the original source instead of the source-map utility. The reason is that, loading the original source requires `getState` and `dispatch`, and i'm trying to keep these dependencies outside of the source-map utility.

public/js/actions/breakpoints.js
```diff
@@ -22,15 +23,51 @@ function enableBreakpoint(location: Location) {
...
   return currentBp && !currentBp.get("disabled");
 }
 
 function _getOrCreateBreakpoint(state, location, condition) {
+  location = getGeneratedLocation(state, location);
   return getBreakpoint(state, location) || fromJS({ location, condition });
 }
 
+function getBreakpointText({
+  breakpoint, getState, getTextForLine, actualLocation, location }) {
```
jlongster: I'm a little confused between the difference of `actualLocation` and `location` here. It looks like you pass in the before/after result of the `setBreakpoint` call, and `actualLocation` is only different if it slid the breakpoint. It doesn't have to do with sourcemapping.

public/js/actions/breakpoints.js
```diff
@@ -22,15 +23,51 @@ function enableBreakpoint(location: Location) {
...
+  let text = breakpoint.get("text");
+  if (text) {
+    return text;
+  }
+
+  if (!getTextForLine) {
+    return "";
+  }
+
+  if (isOriginalLocation(getState(), location)) {
```
jlongster: Since many point of code will have to deal with sourcemapping, I think it's easier to do it like the current debugger server: have a function like `getOriginalLocation` that takes a location and always returns the original location. If the source is not sourcemapped, it'll just return the same location. If it is, it'll sourcemap it.

But it may be you are still working through this :) I don't quite understand this code because `location` and `actualLocation` are not different here.

public/js/actions/breakpoints.js
```diff
@@ -22,15 +23,51 @@ function enableBreakpoint(location: Location) {
   return addBreakpoint(location);
 }
 
+function isOriginalLocation(state, location: Location) {
```
jlongster: Can you prefix this with an underscore? All top-level methods in an action file should be action creators, and we use that convention to indicate "internal" methods in here.

public/js/actions/sources.js
```diff
@@ -184,6 +184,25 @@ function togglePrettyPrint(id) {
...
 
+function loadOriginalSourceText(source, getState, dispatch) {
+  return Task.spawn(function* () {
+    const generatedSource = getSource(
+      getState(),
+      getGeneratedSourceId(source)
+    );
+
+    const generatedSourceText = yield dispatch(
+      loadSourceText(generatedSource.toJS())
```
jlongster: I'm probably missing something here but why do we need to fetch the source at all if we just want to original source, and we already have that from the sourcemap?

public/js/actions/breakpoints.js
```diff
@@ -22,15 +23,51 @@ function enableBreakpoint(location: Location) {
...
+}
+
 function _breakpointExists(state, location: Location) {
+  location = getGeneratedLocation(state, location);
   const currentBp = getBreakpoint(state, location);
   return currentBp && !currentBp.get("disabled");
 }
 
 function _getOrCreateBreakpoint(state, location, condition) {
+  location = getGeneratedLocation(state, location);
```
jlongster: Nice

public/js/selectors.js
```diff
@@ -106,6 +113,24 @@ function makeLocationId(location: Location) {
   return location.sourceId + ":" + location.line.toString();
 }
 
+function getOriginalLocation(state: AppState, location) {
```
jlongster: Oh, here is `getOriginalLocation`. I feel like this and `getGeneratedLocation` should be in the same place.

This looks good though.

jlongster:
This is looking good, the only thing that's confusing to me is the `getBreakpointText` method and using `actualLocation` and `location` like that. If you can, try to rework that before merging this because it's pretty confusing. I don't see why you can't just have:

```js
let text = bp.get("text");
if (!text) {
  text = getTextForLine ? getTextForLine(getOriginalLocation(actualLocation).line) : "";
}
```

jlongster:
> selectors translate locations to original locations (useful for ui)

I'm wondering how you are thinking of dealing with the async nature of sourcemaps. Say we display a bunch of sourcemapped locations in the UI, and it shows the raw locations because it doesn't have the sourcemap yet. What happens when the sourcemap comes through? The selectors API is sync (and must be). Are you implying that we are going to update some part of the redux app state, which will rerender those locations?

jlongster:
It might be easier if it's only the action layer that does sourcemapping. Then you can provide an async API:

```js
// "outgoing" messages that need original -> generated
dispatch({
  type: "ADD_BREAKPOINT",
  [PROMISE]: Task.spawn(function*(location) {
    const loc = yield getGeneratedLocation(location);
    return client.setBreakpoint(loc)
  })
})

// "incoming" messages that need generated -> original
function paused({ frame }) {
  const frame = yield getOriginalLocation(frame.location);
  // ...
}
```

These APIs would actually wait until the sourcemap is fetched & parsed if not already.

I don't know what the best method is yet, just discussing! Please keep doing what you're doing and it'll be interesting to see how it works.

public/js/selectors.js
jasonLaster: agreed

public/js/actions/sources.js
jasonLaster: I believe we need the generated source to get the original source. I'm getting the original source from the source node. I might be mistaken and can read each of the original sources from the consumer...

public/js/actions/breakpoints.js
jasonLaster: sure. I might move those into a utility, but i don't want to go back to the situation where the utility requires selectors and there's a circular dependency

public/js/actions/breakpoints.js
jasonLaster: agreed

public/js/actions/breakpoints.js
jasonLaster: yeah - this is a pure extraction because i could no longer use the ternary. very little to do w/ source maps

public/js/actions/sources.js
jlongster: Might be worth looking into, I could be wrong too. From a high-level it seems like you should be able to serve the source for a given sourcemapped URL if you have a sourcemap. I have not thought this through though!

public/js/actions/breakpoints.js
jlongster: I see. I think the tension is with `actualLocation`. If we're sourcemapping, we know the server doesn't slide breakpoints so we can just use the original location. However, Chrome's server does slide even with sourcemaps. I think you should always use `actualLocation`, sourcemap that *back* to an original location and pass it here as a single location.

This stuff is so confusing.... :p

jasonLaster:
we're changing directions to store the original location. closing and trying again. lots of good lessons here though

Issue #338, opened 2016-07-14
File Search - UI tweaks


jasonLaster:
Here are a couple file search tweaks that come to mind. @helenvholmes, @clarkbw, @jlongster any others i'm missing?

- [x] the magnifying glass should be aligned on firefox (it is in chrome) (fixed)
- [x] the magnifying glass should only be blue when focused
- [x] the input bar should have placeholder copy for searching
- [x] file search should be open by default (with copy below mentioning cmd+p opens files) 
- [ ] there should be a button for exiting file search when there is already a selected source
- [x] Lets also replace the CSS values with variables

jlongster:
Don't have many thoughts on this, but the magnifying glass is aligned for me in Firefox...

jasonLaster:
oh strange

clarkbw:
> but the magnifying glass is aligned for me in Firefox...

Same here.  Otherwise looks like a good list.

jasonLaster:
Lets also replace the CSS values with (variables
)[https://github.com/jlongster/debugger.html/blob/master/public/js/lib/variables.css#L38]

jasonLaster:
@amitzur, interested in working on this list? 

I think they should all be fairly actionable except for the item "file search" button one. @helenvholmes, do you have an idea of where you think we should surface a file search button if it's not open?

PR #358, opened 2016-07-19
autocomplete UI tweaks: center icon and focus outline

amitzur:
Since longhand background-position-* are only implemented starting firefox 49, I thought they aren't necessary in this case.
Also, I thought that using the border rather than outline is better for focus state in this case since otherwise both of them show up

### DIFF
--- a/public/js/components/Autocomplete.css
+++ b/public/js/components/Autocomplete.css
@@ -51,10 +51,9 @@
   font-size: 0.8em;
   background-image: url("/images/magnifying-glass.svg");
   background-repeat: no-repeat;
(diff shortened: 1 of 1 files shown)


amitzur:
Just noticed this partially fixes issue #338

amitzur:
Not sure why cypress failed. I still haven't managed to properly install it

jasonLaster:
Thanks @amitzur. That looks like a cypress intermittent. I'm not too worried as we run cypress twice (for intermitents) and the other passed fine.

That's pretty great that you fixed some of the tweaks (before you saw the issue).

PR #359, opened 2016-07-19
file search overflow should not lose the input

amitzur:
This is not part of issue #338, but I believe it is better to have the overflow on the UL element so that scrolling a long list of sources doesn't cause the input to escape the viewport

### DIFF
--- a/public/js/components/Autocomplete.css
+++ b/public/js/components/Autocomplete.css
@@ -7,18 +7,20 @@
 .autocomplete ul {
   list-style: none;
   width: 100%;
+  max-height: calc(100% - 32px);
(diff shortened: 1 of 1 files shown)


jasonLaster:
oh, that's a nice idea. I was thinking about something similar as well.

jasonLaster:
While you're in there could you replace (#dde1e4) with var(--theme-grey)
https://github.com/jlongster/debugger.html/blob/master/public/js/lib/variables.css#L38

jasonLaster:
Looks good, I'll make a small ticket for replacing the values w/ variables

PR #362, opened 2016-07-19
Extract getState, dispatch from source-map util

jasonLaster:
Small refactor from the last bp pr to extract getState, dispatch, and selectors from source map util

### DIFF
--- a/public/js/util/source-map.js
+++ b/public/js/util/source-map.js
@@ -1,8 +1,6 @@
 const invariant = require("invariant");
 const { SourceMapConsumer, SourceNode } = require("source-map");
 const { Source } = require("../types");
-const { Task } = require("./task");
-const { getSource } = require("../selectors");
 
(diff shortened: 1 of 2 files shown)


public/js/util/source-map.js
```diff
@@ -71,33 +62,25 @@ function createOriginalSources(generatedSource) {
...
-      generatedSourcetext.text
-    );
-
-    return {
-      source: originalSourceContent,
-      contentType: "text/javascript"
-    };
-  });
+function getOriginalSource(
+  originalSource, generatedSource, generatedSourceText) {
```
jlongster: nit: this alignment looks a little weird because the args are indented with the code, and we (devtools) usually do:

```
function foo(
  x, y, z
) {
```

jasonLaster: thanks

jasonLaster:
yeah - i feel that way too.

PR #363, opened 2016-07-19
Additional source-map util refactors

jasonLaster:
+ underscore private methods
+ rename setSourceMap to setConsumer and make it private
+ move some methods around


jlongster:
Looks good r+ (as mentioned on IRC you could convince me that we don't need the private underscore, I'm on the fence...)

jasonLaster:
ha - yeah. this is more a symptom of not having a clear idea for the util yet. it's a bag of functions that keeps changing based on my goals... when it settles down i'll see what we have :)

PR #364, opened 2016-07-19
Translate breakpoint locations to original locations

jasonLaster:
I can't believe how much cleaner this approach is from before. 

Things fell in line really nicely:
+ selectors translate locations
+ source map acquired a couple useful helpers
+ client is wrapped so that it receives generated locations

This probably could be dried up further, but it shows promise

### DIFF
--- a/public/js/actions/breakpoints.js
+++ b/public/js/actions/breakpoints.js
@@ -5,7 +5,8 @@
 
 const constants = require("../constants");
 const { PROMISE } = require("../util/redux/middleware/promise");
-const { getBreakpoint, getBreakpoints } = require("../selectors");
+const { getBreakpoint, getBreakpoints,
+        getOriginalLocation, getGeneratedLocation } = require("../selectors");
 const fromJS = require("../util/fromJS");
 
 import type { Location } from "./types";
@@ -15,13 +16,6 @@ type ThunkArgs = {
   getState: any,
   client: any
 }
-
-function enableBreakpoint(location: Location) {
-  // Enabling is exactly the same as adding. It will use the existing
-  // breakpoint that still stored.
-  return addBreakpoint(location);
-}
-
 function _breakpointExists(state, location: Location) {
   const currentBp = getBreakpoint(state, location);
   return currentBp && !currentBp.get("disabled");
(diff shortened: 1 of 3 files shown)


public/js/actions/breakpoints.js
```diff
@@ -31,9 +25,17 @@ function _getOrCreateBreakpoint(state, location, condition) {
   return getBreakpoint(state, location) || fromJS({ location, condition });
 }
 
+function enableBreakpoint(location: Location) {
```
jasonLaster: moved below the helper functions

public/js/actions/breakpoints.js
```diff
@@ -31,9 +25,17 @@ function _getOrCreateBreakpoint(state, location, condition) {
...
+  // Enabling is exactly the same as adding. It will use the existing
+  // breakpoint that still stored.
+  return addBreakpoint(location);
+}
+
 function addBreakpoint(location: Location,
                        { condition, getTextForLine } : any = {}) {
   return ({ dispatch, getState, client }: ThunkArgs) => {
+    location = getOriginalLocation(getState(), location);
```
jlongster: Do we need to do this here? Any location coming from the UI should already be an original location, but if you want to do it for sanity, that's fine. Once we actually type locations we should be able to enforce that `location` is an original location and we shouldn't need to do this.

public/js/selectors.js
```diff
@@ -99,6 +103,43 @@ function getSourceMapURL(state: AppState, source: Source) {
   return tab.get("url") + "/" + source.sourceMapURL;
 }
 
+function getGeneratedLocation(state: AppState, location: Location) {
```
jlongster: Perfect, this looks great

jlongster:
👍  This looks awesome and does seem to have the feeling I would expect sourcemaps to have. Great work!

public/js/actions/breakpoints.js
jasonLaster: yup, it was sanity. We can hold off on it for now

jasonLaster:
Thanks. Yeah. I'm really happy with how this looks.

PR #365, opened 2016-07-20
(WIP) Add Console to debugger

jasonLaster:
This is a work in progress proof of concept of the new console running in the debugger.

##### What's done:
+ new-console-output and webconsole/utils are imported
+ webpack dependencies are mapped

##### What needs to be done
+ HTMLify utils
+ wire up the client to the console output
+ other things?


jlongster:
I really, really don't want to commit the entire console here. The debugger and console projects should stay separate and we provide ways to tie them together either in a toolbox or in some other custom configuration. We want to be removing devtools dependencies (as in checked-in) from this repo, not adding them.

Also, the console is not in this sprint, so I don't think we should be spending time on it. There is a lot to figure out about how to tie together the debugger/console outside of devtools but I don't think we have time right now.

jasonLaster:
sorry, should have added a disclaimer that i did not plan on merging this, just show how that it is possible.

jasonLaster:
closing this as we don't want to merge. will link to the branch from the console issue

PR #367, opened 2016-07-20
load original sources

jasonLaster:
This PR fixes issues we had with loading original source text.

**Problem:**
When a user tries to add a breakpoint in a generated source, the breakpoint fails to be added because the original source text has not been "loaded". This is a mistake because once we've loaded the generated source text, we've effectively loaded the original source texts as well. 

**Solution**
Load the original source texts in "loadSourceText" action. This involves updating the loadSourceText action and reducers updateText. UpdateText now expects a list of values (sources or source texts to update).

### DIFF
--- a/public/js/actions/sources.js
+++ b/public/js/actions/sources.js
@@ -8,14 +8,17 @@ const { Task } = require("../util/task");
 const { isJavaScript } = require("../util/source");
 const { networkRequest } = require("../util/networkRequest");
 
-const { getSource, getSourceText,
-        getSourceMap, getSourceMapURL } = require("../selectors");
 const constants = require("../constants");
 const Prefs = require("../prefs");
 const invariant = require("invariant");
 const { isEnabled } = require("../../../config/feature");
-const { createOriginalSources, getGeneratedSourceId,
-        isOriginal, getOriginalSource } = require("../util/source-map");
+
+const { createOriginalSources, getOriginalTexts
+      } = require("../util/source-map");
+
+const { getSource, getSourceText, getSourceByURL, getGeneratedSource,
+        getSourceMap, getSourceMapURL, getOriginalSources
+      } = require("../selectors");
 
 /**
  * Throttles source dispatching to reduce rendering churn.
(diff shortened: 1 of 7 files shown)


public/js/actions/sources.js
```diff
@@ -102,14 +105,14 @@ function selectSource(id, options = {}) {
       return;
     }
 
-    const source = getSource(getState(), id).toJS();
+    const source = getSource(getState(), id);
```
jlongster: The contract for actions is that they always receive normal JS objects. This gives us some sanity about the API, because components can directly fire actions, and they should always know that that can simply pass literal objects like `{ sourceId: ... }`. So please move `toJS` back here, even if right now that action creator is only fired internally.

public/js/actions/tests/loadSourceText.js
```diff
@@ -74,8 +75,8 @@ describe("loadSourceText", () => {
     const store = createStore(simpleMockThreadClient);
 
     Task.spawn(function* () {
-      yield store.dispatch(loadSourceText({ id: "foo1" }));
-      yield store.dispatch(loadSourceText({ id: "foo1" }));
+      yield store.dispatch(loadSourceText(fromJS({ id: "foo1" })));
+      yield store.dispatch(loadSourceText(fromJS({ id: "foo1" })));
```
jlongster: This is why actions always just receive normal JS objects :) We want to optimize for the common use case, and with actions you are always passing things in like this.

jlongster:
It seems a little odd that loading original source texts is tied into loading generated source texts. You should get the original source texts at the time the sourcemap is loaded, right?

You described the problem as a user setting a bp in a generated source and the original text hasn't been loaded yet. Why is that a problem? If it's a generated source, shouldn't it be able to just send off the generated location immediately?

I don't have as much context as you though, so I'll defer to you. This looks like a good direction anyway and we can always change things later after we get a feel for how everything works. If you think this is how we should go, you can merge it.

public/js/actions/sources.js
jasonLaster: yeah - that's true.

jasonLaster:
> You should get the original source texts at the time the sourcemap is loaded, right?

We need the generated source to get the original source.

> If it's a generated source, shouldn't it be able to just send off the generated location immediately?


We need the original location for the breakpoint list.

PR #373, opened 2016-07-20
(wip) proposal for moving selectors into reducers

jlongster:
Don't merge this quite yet, I might make a final round of tweaks. But this shows you what it will look like.

Unfortunately, as the comment explains, it's not quite as clean as I would like. I would have loved for the selectors to just take the "sources" state and work with it directly, but unfortunately that's hard. We still want to expose a single module that the UI can require to get access to the selectors, and be able pass top-level app state into them (`getSources(appState, ...)`). This means we need to import all the selectors, "wrap" them so they pick off the right sub-state, and call the selector from the reducer.

But that's impossible to type. This is as close as I got: https://gist.github.com/jlongster/591b785b74a9ba0627e8e5241a30607f Unfortunately, because Flow requires explicit annotations on exported functions, it's really hard to come up with a generic `wrap` that will satisfy it. We might be able to figure it out later, but I need to stop spending time on this.

### DIFF
--- a/public/js/reducers/sources.js
+++ b/public/js/reducers/sources.js
@@ -5,8 +5,10 @@
 
 const fromJS = require("../utils/fromJS");
 const I = require("immutable");
+const makeRecord = require("../utils/makeRecord");
 
 import type { Action, Source } from "../actions/types";
+import type { Record } from "../utils/makeRecord";
 
 export type SourcesState = {
   sources: I.Map<string, any>,
@@ -16,13 +18,13 @@ export type SourcesState = {
   sourceMaps: I.Map<string, any>,
 };
 
-const State = I.Record({
+const State = makeRecord(({
   sources: I.Map({}),
   selectedSource: undefined,
   sourcesText: I.Map({}),
   sourceMaps: I.Map({}),
   tabs: I.List([])
-});
+} : SourcesState));
 
 function update(state = State(), action: Action) {
   switch (action.type) {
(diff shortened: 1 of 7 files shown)


jasonLaster:
Thanks for working on this. I've been thinking recently that it's nice to consolidate selectors and reducers.

public/js/reducers/sources.js
```diff
@@ -175,9 +177,58 @@ function getNewSelectedSource(state, id) : ?Source {
...
+function getSelectedSource(state: OuterState) {
+  return state.sources.selectedSource;
+}
+
+function getSourceMap(state: OuterState, sourceId: string) {
+  return state.sources.sourceMaps.get(sourceId);
+}
+
+module.exports = {
+  State,
```
jasonLaster: This might make more sense as SourceState if it's ever used next to another reducer state

public/js/reducers/sources.js
```diff
@@ -175,9 +177,58 @@ function getNewSelectedSource(state, id) : ?Source {
...
 
-module.exports = update;
+// Unfortunately, it's really hard to make these functions accept just
+// the state that we care about and still type if with Flow. The
+// problem is that we want to re-export all selectors from a single
+// module for the UI, and all of those selectors should take the
+// top-level app state, so we'd have to "wrap" them to automatically
+// pick off the piece of state we're interested in. It's impossible
+// (right now) to type those wrapped functions.
+type OuterState = { sources: Record<SourcesState> };
```
jasonLaster: hmm, we can easily update this later

public/js/reducers/sources.js
jlongster: Yeah, I think it conflicts with the type name, and I would encourage this module to be used like `sources.State`. It should be used in tests 99% of the time where this specific reducer is being tested. But I may change, I'll think about it some more.

public/js/reducers/sources.js
jlongster: Our coworker gozala (uses flow in browser.html) may have come up with a solution. But yeah, for now I gotta stop spending time on it..

public/js/reducers/sources.js
jasonLaster: Yeah, that's reasonable unless it's destructured. `{state, update} = require()`

jlongster: That's what I mean by "I would encourage this module to be used like sources.State". It's an API you shouldn't destructure, because if you do we should change it to `SourcesState` and `sourcesUpdate`, and by prefixing everything it's the same as `sources.*`.

jasonLaster: i see. that makes sense.

jasonLaster:
r+

jlongster:
I had to resolve a conflict, hopefully this doesn't break anything...

Need to spend some time soon converting most of the code to new stuff like flow/reducers+selectors/etc. I think landing the JS bundle in Firefox shouldn't be too hard at this point so I might find time next week for that.

PR #375, opened 2016-07-20
remove redux devtools and encourage usage of the extension

jlongster:
While looking at landing a bundle in Firefox, I wanted to preemptively reduce the size of the bundle by removing big low-hanging fruit. Redux devtools was one of these, there's no reason to include it. I tried to conditionally include it and make webpack strip it out in prod mode, but it made things even more awkward. I think the browser extensions are a better experience anyway, so we should encourage that. We still have the ability to log action to the console and use normal devtools to inspect it.


jasonLaster:
well, this didn't last long... :)

zalmoxisus:
BTW, [it supports Firefox now](https://github.com/zalmoxisus/redux-devtools-extension/releases/tag/v2.3.1), though only Nightly and Developer Edition till Firefox 48, which gets released.

jlongster:
@zalmoxisus Awesome! Yeah, I installed it locally and it worked great (nightly). Thanks for all your work!

PR #380, opened 2016-07-21
translate pause frame

jasonLaster:
Small PR for translating the pause frame to the original location

#### Old
screen shot 2016-07-21 at 10 05 19 am

#### New
screen shot 2016-07-21 at 10 05 30 am


jlongster:
Looks good, r+

PR #385, opened 2016-07-21
Add full URL as tab title

darkwing:
I find it helpful in tab situations when the full URL or path to file is show-on-hover with the tab.  

Just a suggestion, feel free to close.


jlongster:
That's a very good idea!

jlongster:
Karma is flaky which really sucks :/ This should not be a problem.

PR #388, opened 2016-07-21
Move defer from devtools to util

### LINKED ISSUE #387: Cannot resolve module 'promise'
npm install causes this error

```
ERROR in ./public/js/lib/devtools/shared/defer.js
Module not found: Error: Cannot resolve module 'promise' in /Users/<username>/debugger/debugger.html/public/js/lib/devtools/shared
 @ ./public/js/lib/devtools/shared/defer.js 8:16-34
```

jasonLaster:
Fixes #387

What i think happened is that we pulled in a new version of `defer.js` that requires promise. At this point I don't think we get anything by using m-c's defer function, so it's probably safe to just create one in utils


jlongster:
The only reason I'd think we need it is for compat with other devtools modules. We don't need defer ourselves at all, right? We can just use `Promise`

jasonLaster:
@jlongster we might need to do something in the future. The solution might be to refactor those places too.

PR #392, opened 2016-07-21
update the configuration settings

clarkbw:
also fix up the `npm run firefox` command

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -25,6 +25,7 @@ We respect your time and want to help you make the most of it as you learn more 
   * [Writing Code](#writing-code)
     * [Your First Code Contribution](#your-first-code-contribution)
     * [Pull Requests](#pull-requests)
+    * [Hot Reloading](#hot-reloading-fire)
   * [Tests](#tests)
     * [Unit Tests](#unit-tests)
     * [Integration Tests](#integration-tests)
@@ -32,10 +33,6 @@ We respect your time and want to help you make the most of it as you learn more 
     * [Storybook](#storybook)
   * [Configuration](#configuration)
     * [Create a local config file](#create-a-local-config-file)
-
-[Styleguides](#styleguides)
-  * [JavaScript Styleguide](#javascript-styleguide)
-  * [CSS Styleguide](#css-styleguide)
 
 ## What should I know before I get started?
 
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -248,10 +248,20 @@ Read more information in [storybook docs](./docs/local-development.md#storybook)
...
+  * `tabs` Enables source view tabs in the editor (CodeMirror)
+  * `sourceMaps` Enables source map loading when available
+* `chrome` Chrome browser related flags
+  * `debug` Enables listening for remotely debuggable Chrome browsers
+  * `webSocketPort` Configures the web socket port specified when launching Chrome from the command line
+* `firefox` Firefox browser related flags
+  * `proxyPort` Port used by the development server run with `npm start`
+  * `webSocketConnection` Favours Firefox WebSocket connection over proxy, requires [bug 1286281](https://bugzilla.mozilla.org/show_bug.cgi?id=1286281)
+  * `geckoDir` Local location of Firefox source code _only needed by project maintainers_
+* `hotReloading` enables [Hot Reloading](./docs/local-development.md#hot-reloading) of CSS and React
```
jasonLaster: This might make sense as a README.md next to config. That way we can link to it from here and also other places as well. 

I think if it's in the README we would be more likely to keep this up to date

jlongster: I agree, I think this is too much of a dump of information. Most people (even advanced devs) won't care about most of this and will just use the defaults.

We just need to highlight a few that really matter: `hotReloading` and maybe `chrome.debug`. I really think we need a dedicated section to `hotReloading` to tell devs to turn it on because it makes the dev experience so much better. Or we could just turn it on by default in dev mode.

jlongster:
I like how we highlighted just a few prefs before, so if you really think this is valuable let's do what Jason said and move it to `config/README.md`.

clarkbw:
Hmm, lots to think about. 

I agree that it feels like a dump of information and that it would be better to have only the useful highlights like chrome.debug and hotReloading.  

I like the `config/README.md` idea, feels like it will be more likely to be updated and I'll leave only the highlights in the contributing doc.

jasonLaster:
Sounds good. 

At Etsy, we eventually extended our config spec to include `description` fields like so

```
hotReloading: {
  enabled: false,
  description: "la la la"
}
```

clarkbw:
config/README.md
screen shot 2016-07-22 at 8 13 23 am

clarkbw:
contributing#configuration
screen shot 2016-07-22 at 8 13 58 am

jasonLaster:
Looks great. Thanks. These changes are really helpful

Issue #328, opened 2016-07-14
Source Maps - domains are not set


jasonLaster:
domain should be `webpack` but is empty.

Also - the domain should be url decoded  so %20 is a space. See second domain

![screen shot 2016-07-14 at 10 56 57 am](https://cloud.githubusercontent.com/assets/254562/16844084/b3f55460-49b1-11e6-9170-827028833f96.png)

jasonLaster:
@jbhoosreddy want to take a look at this?

jbhoosreddy:
@jasonLaster. Yes. I'd like to.

jlongster:
This was fixed in #393

PR #393, opened 2016-07-22
Fix source map domain issue and decoding URI (#328)

jbhoosreddy:
After looking into so many things, it turned out to be easier than I thought. @jasonLaster 
Let me know if this is fine.

### DIFF
--- a/public/js/utils/sources-tree.js
+++ b/public/js/utils/sources-tree.js
@@ -38,7 +38,6 @@ function getURL(source) {
   }
 
   let urlObj = URL.parse(url);
-
   if (!urlObj.protocol && urlObj.pathname[0] === "/") {
     // If it's just a URL like "/foo/bar.js", resolve it to the file
     // protocol
@@ -59,11 +58,16 @@ function getURL(source) {
       path: "/",
       group: url
     };
+  } else if (urlObj.protocol === "http:" || urlObj.protocol === "https:") {
+    return {
+      path: urlObj.pathname,
+      group: urlObj.host
+    };
   }
 
   return {
-    path: urlObj.pathname,
-    group: urlObj.protocol === "file:" ? "file://" : urlObj.host
+    path: urlObj.path,
+    group: urlObj.protocol + "//"
   };
 }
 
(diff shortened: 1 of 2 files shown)


public/js/utils/sources-tree.js
```diff
@@ -59,6 +59,11 @@ function getURL(source) {
       path: "/",
       group: url
     };
+  } else if (urlObj.protocol === "webpack:") {
```
jasonLaster: haha - we might want to do something like this or a non http... not sure

public/js/utils/sources-tree.js
```diff
@@ -72,6 +77,7 @@ function addToTree(tree, source) {
   if (!url) {
     return;
   }
+  url.path = decodeURIComponent(url.path);
```
jasonLaster: nice, thanks!

jasonLaster:
👍  this solution as is will need to be generalized so that it works in other cases, but for now is probably good

jbhoosreddy:
I didn't know the other cases, so I only aimed at solving the one I knew. But I can definitely make it more comprehensive. I guess we just need to make sure that the parser is robust enough to understand most corner-cases.

public/js/utils/sources-tree.js
jlongster: It's probably better to whitelist the protocols where know there is a domain/host, at least as a temporary solution. Change it to:

```
} else if (urlObj.protocol === "http" || urlObj.protocol === "https") {
  return {
    path: urlObj.pathname,
    group: urlObj.host
  }
}

return {
  path: urlObj.path,
  group: urlObj.protocol + "//"
}
```

jlongster: And what's the difference between `path` and `pathname` again?

jlongster:
A small change to the `getUrl` function will make this work for other things as well, so make that change and this is good to merge.

jasonLaster:
@jlongster what do you mean by `getUrl` function?

public/js/utils/sources-tree.js
jasonLaster: I believe path is `pathname + search`

Issue #372, opened 2016-07-20
Pause on exception 


clarkbw:
Pause on exception allows for developers to automatically pause the debugger whenever an exception occurs in their code.  This tool is especially helpful when a new exception appears however having this always on tends to be annoying as often there are intermittent exceptions while coding and pausing can be interrupting.  Often there are also exceptions within libraries, however this should probably be handled with black boxing.

This option is currently hidden in the Firefox debugger settings.  I think this should be an upfront control which can be more easily toggled and the state more easily known.

Need design from @helenvholmes 

A screenshot of the Firefox debugger settings where this lives:
![screen shot 2016-07-20 at 10 52 08 am](https://cloud.githubusercontent.com/assets/2134/16996897/0c74b21a-4e68-11e6-9b66-33024b5ebd68.png)

A screenshot of Chrome for reference, the pause on exception is the `(||)` button on the right.
![screen shot 2016-07-20 at 10 44 01 am](https://cloud.githubusercontent.com/assets/2134/16996627/fb3fd6d8-4e66-11e6-9769-3918999d2ae8.png)

jasonLaster:
Here's a patch to get started with
https://gist.github.com/ad9aa95dd24dd428df2cf35efabd496d

jbhoosreddy:
@jasonLaster I can take this one. 😃

helenvholmes:
@clarkbw: hey, I know this has since been closed, but on designs needed from me: were you referring to the icon for pause on exceptions, or designs for the paused state in the debugger? (attaching those below):

![screen shot 2016-07-26 at 9 22 08 am](https://cloud.githubusercontent.com/assets/1720093/17139466/962207d4-5312-11e6-9739-ce386d6b07d2.png)

clarkbw:
> designs for the paused state in the debugger

I think we were handling this in another issue.  

This was first and foremost about the icon and then anything related to the toggle state behavior / interactions.

helenvholmes:
@clarkbw Should I open a new bug with that info?

jasonLaster:
Yeah, definitely

PR #394, opened 2016-07-22
add pause on exceptions feature (#372)

jbhoosreddy:
@jasonLaster thoughts? I reused another SVG for now. Could probably hold off for better image or replace later. and maybe tests to add also for this?


jasonLaster:
👍  great job!

jlongster:
Yeah, looks good to me!

PR #400, opened 2016-07-22
apply flow to breakpoints reducer, move breakpoints selectors, and convert to JS types

jlongster:
This is a large PR, but it can't really be helped. When converting things from immutable to JS and applying types we need to change everything that uses it.

I want to start writing more tests for the reducers and actions so this helps a lot (and we won't need to update tests if we do these changes now).

A breakpoint in the app state is now a normal JS object. We use it far more than we update it, so it's easier, but the main reason is so that we can have a single `Breakpoint` type and use it everywhere. If we want it to be immutable, we would still use a record type and all the external uses would use normal property lookup (`breakpoint.location`), so we could make that change later if we really wanted.

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -34,6 +34,9 @@ module.exports = function(config) {
       "./public/js/**/tests/*": ["webpack"]
     },
 
-    webpack: webpackConfig
+    webpack: webpackConfig,
+    webpackMiddleware: {
+      noInfo: true
+    }
   });
 };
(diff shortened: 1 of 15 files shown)


karma.conf.js
```diff
@@ -34,6 +34,9 @@ module.exports = function(config) {
       "./public/js/**/tests/*": ["webpack"]
     },
 
-    webpack: webpackConfig
+    webpack: webpackConfig,
+    webpackMiddleware: {
+      noInfo: true
+    }
```
jasonLaster: nice

jasonLaster:
This looks good @jlongster.

re: making a breakpoint a pojo, do we lose any of the automatic component updates? I think this makes a lot of sense if we're not updating components based on breakpoint data changing. If location changes, it's a new breakpoint i suppose.

jlongster:
> do we lose any of the automatic component updates?

You mean optimizations because of reference equality? Not really, mainly because nothing is connected directly to a breakpoint anyway, components get all the breakpoints. An un-changed breakpoint will still be the same instance as before too, so reference equality should still work.

PR #401, opened 2016-07-22
Update name of CDP to the official name

auchenberg:
Per https://github.com/ChromeDevTools/debugger-protocol-viewer/issues/41, the official name is `Chrome Debugging Protocol`


jlongster:
Thanks for the correction :)

PR #404, opened 2016-07-22
Pretty print

jasonLaster:
This is a first pass on client side pretty printing.

Adds:
* a pretty print worker that generates the prettified text and ugly to pretty mappings
* adds a createSourceMap function that creates a map from ugly to pretty

Next steps:
* store the pretty text and show it in the editor
* write and extend function that will extend an original source map if the ugly source is mapped

Question:
* i'm not sure what to do with the pretty-fast lib and deps. At the moment they're in lib... I think they should have the same mc path `devtools/<same-mc-path`. We will probably need to update the paths to reflect their chrome nature. In addition, the worker will probably have to be a ChromeWorker for this as well... All of this to say there's some work to do once we land the bundle to wire up the pretty-print worker.

### DIFF
--- a/config/development.json
+++ b/config/development.json
@@ -6,7 +6,8 @@
   },
   "features": {
     "tabs": true,
-    "sourceMaps": true
+    "sourceMaps": true,
+    "prettyPrint": false
   },
   "chrome": {
     "debug": false,
--- a/public/js/actions/sources.js
+++ b/public/js/actions/sources.js
@@ -7,14 +7,14 @@ const { PROMISE } = require("../utils/redux/middleware/promise");
 const { Task } = require("../utils/task");
 const { isJavaScript } = require("../utils/source");
 const { networkRequest } = require("../utils/networkRequest");
+const { workerTask } = require("../utils/utils");
 
 const constants = require("../constants");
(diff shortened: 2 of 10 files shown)


jlongster:
> i'm not sure what to do with the pretty-fast lib and deps. At the moment they're in lib... I think they should have the same mc path devtools/<same-mc-path.

Why not put them in lib/devtools if they are meant to mirror files we've copied from there? And what do you mean by "chrome nature"? I forget, is there anything specific in them that will need privileges when in Firefox?

jlongster:
Generally this looks like good, not a whole lot other than adding the necessary libs and such :)

jasonLaster:
> chrome nature

+ [paths](https://dxr.mozilla.org/mozilla-central/source/devtools/server/actors/pretty-print-worker.js#30-33)
+ [ChromeWorker](https://dxr.mozilla.org/mozilla-central/source/devtools/shared/worker/worker.js#47)

jlongster:
Oh, I see, so we want to pull those modules from m-c and to do that it needs to be a ChromeWorker in m-c. Don't have any great ideas on that... but we'll figure it out soon.

PR #406, opened 2016-07-22
allow development server to be publicly reachable

clarkbw:
localhost restricts the development server to the local system where
0.0.0.0 allows for external access
added a message to indicate when hot reloading is disabled
lots of other cleanups in the file

### DIFF
--- a/bin/development-server.js
+++ b/bin/development-server.js
@@ -4,13 +4,11 @@
 require("babel-register");
 
 const path = require("path");
-const fs = require("fs");
 const webpack = require("webpack");
 const express = require("express");
 const webpackDevMiddleware = require("webpack-dev-middleware");
 const webpackHotMiddleware = require("webpack-hot-middleware");
 const http = require("http");
-const _ = require("lodash");
 
 // Setup Config
 const getConfig = require("../config/config").getConfig;
@@ -25,13 +23,11 @@ if (!feature.isEnabled("firefox.webSocketConnection")) {
 
 function httpGet(url, onResponse) {
   return http.get(url, (response) => {
-    let body = '';
-    response.on('data', function(d) {
+    let body = "";
+    response.on("data", function(d) {
       body += d;
     });
-    response.on('end', function() {
-      onResponse(body);
-    });
+    response.on("end", () => onResponse(body));
   });
 }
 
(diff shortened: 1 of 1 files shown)


bin/development-server.js
```diff
@@ -47,12 +43,14 @@ app.use(webpackDevMiddleware(compiler, {
   stats: { colors: true }
 }));
 
-if(feature.isEnabled("hotReloading")) {
+if (feature.isEnabled("hotReloading")) {
   app.use(webpackHotMiddleware(compiler));
+} else {
+  console.log("Hot Reloading can be enabled by adding " +
+  "\"hotReloading\": true to your local.json config");
```
clarkbw: I don't think we want to message for every feature, but hot reloading is an important one so I wanted to include this.  At first I went with trying to link to the docs but it is simpler to just offer the solution.

bin/development-server.js
```diff
@@ -67,13 +65,11 @@ app.get("/get", function(req, res) {
...
-  httpReq.on('error', function (err) {
-    res.status(500).send(err.code);
-  });
-})
+  httpReq.on("error", err => res.status(500).send(err.code));
+});
 
 // Listen
-app.listen(8000, "localhost", function(err, result) {
+app.listen(8000, "0.0.0.0", function(err, result) {
```
clarkbw: This is the change that brought me to this file.  I'm testing things out with Edge and needed to be able to access `http://clarkbw.local:8000`.  I don't think there are security concerns here.  You can omit the entry altogether and have a similar effect bu this seems more straightforward.

bin/development-server.js
jlongster: That's not a bad idea, I like it. I was trying to think of a way to encourage it without having it on the readme, and this makes sense.

bin/development-server.js
jlongster: That's fine with me. Does anyone use this to send a public URL to someone else? Usually you have to use a forwarding service anyway because you are behind a firewall/router.

jlongster: Oh, your comment didn't exist when I commented. I see, makes sense

bin/development-server.js
jasonLaster: Oh nice, we do this with client logging as well. Great pattern

jasonLaster:
Great idea @clarkbw. 

@auchenberg remote debugging ftw.

Issue #378, opened 2016-07-20
Add debuggable target instructions to the default page


clarkbw:
Currently when you have no debuggable targets available (Chrome or Firefox) we display this message to help you get setup with Node.

![screen shot 2016-07-20 at 3 34 32 pm](https://cloud.githubusercontent.com/assets/2134/17005660/4a2882f8-4e90-11e6-956a-8cae422b82d7.png)

In #352 we're changing the README and the way the process works for getting started such that it will be necessary to have instructions on this default page.  Users will be encouraged to `npm start` and `open http://localhost:8000/`and from there might be given further "quick start" instructions to getting Firefox or Chrome up and running.  Beyond the quick start instructions we will need to link to the correct locations for the CONTRIBUTING doc and the documentation describing how to start Chrome and Firefox with remote debugging turned on.

jasonLaster:
@jbhoosreddy - this could be a good one too

clarkbw:
This wasn't exactly the right thing to do but this is what I aimed for:

* Always show Firefox + Chrome sections
 * Include quick instructions for starting each
  * Include links to further instructions in the docs
* Update the node section to be similar

Here's the quick mockup I made by commenting out lots of code:

![screen shot 2016-07-22 at 12 47 45 pm](https://cloud.githubusercontent.com/assets/2134/17069282/b337524e-500a-11e6-89b5-e8d802e55a17.png)

jlongster:
Personally I think those commands could be overwhelming here (if the user did something wrong that's unrelated, that's not a great first thing to show). I'd rather keep the form of a statement like "You may be looking too..." and "connect to firefox" is a link that display the command to run Firefox or something (and could also link to the docs).

clarkbw:
> "You may be looking too..." and "connect to firefox"

Agreed, I didn't like the commands approach, you can see it becomes messy and then we have multiple places where these commands live.

jbhoosreddy:
One thing we could do is have a GitHub style "Copy to clipboard" button which copies the appropriate command to clipboard and can be executed by the users directly.

clarkbw:
> GitHub style "Copy to clipboard" button

I like it!  Could keep it short and offer both OS options while still linking to the docs for more information.

clarkbw:
Fixed in #471

PR #408, opened 2016-07-22
Connect to the Firefox web socket proxy remotely

clarkbw:
This is a small change to the Firefox connect lib with some other lint fixes to the `firefox-proxy`.

## Firefox
The web socket connection change allows for another computer (work local network access) to connect to the `firefox-proxy` web socket server and debug Firefox remote-remotely.

## Chrome
I don't have this change working for Chrome debugging yet.  The tab list request is being blocked, possibly because Chrome is connected to localhost doesn't naturally allow external access.  Not sure there is a way around this but I was only targeting Firefox for now.

I should finish with some better error handling for Chrome so when the tab list times out you can continue to work with Firefox.

## Node

A very similar change to the Tabs page allows for Node debugging so I broke that out into a separate commit because the Tabs page is being worked on. #378

### DIFF
--- a/bin/firefox-proxy
+++ b/bin/firefox-proxy
@@ -1,7 +1,7 @@
 #!/usr/bin/env node
 "use strict";
 
-const minimist = require('minimist');
+const minimist = require("minimist");
 const ws = require("ws");
 const net = require("net");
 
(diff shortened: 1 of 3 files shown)


clarkbw:
@jasonLaster look good to merge?

jasonLaster:
Looks great. Can't wait to see the demos!

Issue #397, opened 2016-07-22
Inline SVG


jasonLaster:
When we started, we added SVGs to public/images and included them with ISvg. The problem with this approach is that it would a network request each time the SVG was used, which was slow and had a flash. Since then, we've moved in the direction of including SVGs directly, this does not require the extra request. We have one file now `icon.js` that has many React components. 

The goal is to have a react component for each SVG, that way they can be easily required. 

We might be able use webpack `svg-inline-loader` and `svg-inline-react` to do this transpilation on the fly for us. https://github.com/sairion/svg-inline-react https://github.com/sairion/svg-inline-loader

##### ISvg example
```js
Isvg({
    className: "pause-indicator",
    src: "images/pause-circle.svg"
});
```

#### Inline example
```js
let FolderIcon = props => {
  return dom.span(
    props,
    dom.svg(
      { viewBox: "0 -2 16 16" },
      dom.path({ d: "M2 5.193v7.652c0 .003-.002 0 .007 0H14v-7.69c0-.003.002 0-.007 0h-7.53v-2.15c0-.002-.004-.005-.01-.005H2.01C2 3 2 3 2 3.005V5.193zm-1 0V3.005C1 2.45 1.444 2 2.01 2h4.442c.558 0 1.01.45 1.01 1.005v1.15h6.53c.557 0 1.008.44 1.008 1v7.69c0 .553-.45 1-1.007 1H2.007c-.556 0-1.007-.44-1.007-1V5.193zM6.08 4.15H2v1h4.46v-1h-.38z" }) // eslint-disable-line max-len
    )
  );
};
```

jasonLaster:
@jbhoosreddy this is somewhat of a chore issue, but would be really nice to lock down.

Issue #206, opened 2016-05-25
Source Tree icons should be extracted to images directory


jasonLaster:
When we used ISvg, there was a noticable lag so we inlined the svgs in the tree. Lets move these components to the images directory or components/images

clarkbw:
@jbhoosreddy I believe this issue is done, right?

jbhoosreddy:
yes. It is done.

PR #409, opened 2016-07-24
added inline react component support for svg files (#397) (#206)

jbhoosreddy:
I worked on setting up the inline svg images. @jasonLaster let me know your thoughts on this approach.


jlongster:
Feel free to ping me for reviews too. I'd like to review this as I've researched this approach a while back and found some flaws with it, but I can't remember. I'll look at it on Monday.

jasonLaster:
Thanks @jbhoosreddy for doing the leg work.

@jlongster this looks good to me as long as it works well for the firefox panel.

jbhoosreddy:
if this is fine, I'll look into which other icons I can migrate from Isvg to this approach, and add that to this PR.

jasonLaster:
@jbhoosreddy thanks, i don't see how that can hurt. That way we can evaluate it holistically.

public/images/Icons.js
```diff
@@ -0,0 +1,46 @@
+const React = require("react");
+const InlineSVG = require("svg-inline-react");
+const { DOM: dom } = React;
+
+const DomainIcon = props => {
```
jlongster: Any reason we need all these components now? We could just use `InlineSVG({ src: DomainIcon })` directly in the component that needs it. That would reduce the need to manage this file.

jlongster:
I'm a little concerned that [svg-inline-loader](https://github.com/sairion/svg-inline-loader) is deprecated. But this looks pretty good. It's somewhat similar to what I was thinking of doing. I was going to write a webpack loader that loads SVG and transforms it directly into a React component. 

I'm fine using this for now though! If we run into problems we could fork the library and maintain it.

public/images/Icons.js
jasonLaster: This seems like a nice improvement in another PR.

jasonLaster:
> I'm a little concerned that svg-inline-loader is deprecated

I am too, but had a similar reaction. I think it would be easy enough to maintain a library for our own purposes.

PR #410, opened 2016-07-24
Target instructions

jbhoosreddy:
@clarkbw how does this look? The wording and styling could probably be much better. and a clipboard logo also would be nice.

screen shot 2016-07-24 at 3 30 14 am

screen shot 2016-07-24 at 3 31 26 am


jasonLaster:
Thanks for the great start:

Here are a couple of suggestions, although I'll admit the copy / design is very rough.

+ header that describes that you need to connect to a tab/process
+ consistent design for browsers and node
+ less copy, more structure
+ it'd be great to rip the github clipboard svg (why not)

connect debugger

jbhoosreddy:
thanks @jasonLaster. I'll follow up with the clipboard logo for the commands.

But I'm curious, "start browser"/"start process". Does that imply we want to open firefox/chrome/node directly? It's possible and might be better for users rather than copy/paste shell commands.

jasonLaster:
@jbhoosreddy hmm, I think it's a question of who our audience is... If we were shipping a standalone debugger, then adding the ability to launch a browser w/ the right settings might make a lot of sense, in this case I think providing the clipboard is enough as then the dev can stop / log the process. The copy should probably be something closer to "copy start command" or something like that...

jlongster:
Personally I think all of this can simply be:

screen shot 2016-07-25 at 9 13 20 am

The structure feels very heavy. And these links would just go straight to docs. I don't think we need to over-complicate this. The code here does platform detection which is complicated and something we'll have to support (that stuff is notoriously hard to get right), and embeds the commands to run browsers in the code so we'll always have to remember to update them if we ever change them.

I think it's better to have those commands in a single place: the docs. People will only copy them once and then they'll always access it through shell history.

jasonLaster:
Yeah, I think you're right - linking to the docs is definitely the simplest option.

jlongster:
Feel free to tweak the wording! But I do think it will reduce the maintenance burden...

clarkbw:
> I do think it will reduce the maintenance burden

Agreed, we don't want to be too complicated.

I'd even suggest making that bottom section less about an error state but more like a permanent information footer.  The situation where you've run got Firefox running but want to get Chrome running could still be handled through this text box if it was doing less and just always around pointing to the docs.

jasonLaster:
@jbhoosreddy mind cleaning this up first. Would be nice to close

jbhoosreddy:
Created a new PR in #471.

PR #415, opened 2016-07-25
(wip) proof of concept of action testing

jlongster:
**Don't merge**, just a proof of concept. I wanted to show what I was thinking about how to structure our action/reducer tests. This is just a small step from where we are now, but I think solves some key issues.

The main file to look at is `public/js/actions/tests/newSource.js`. There are few flow tweaks in here that you can ignore. The key points:

* Don't rely as much on the full app state. I created a new cypress command `saveProtocolObjects` (name to be improved) that saves the parts that we really care about. Right now I think the sources are all we really need. Eventually we'll probably need some of the "loaded objects" and things like that.
* It also normalizes the fixture data, namely any actor IDs, so you should be getting the exact same data every time. This fixes the problem where updating the fixtures updates all of the actor IDs (resulting in a large change). And it makes it easy to reference (`sourcesFixture["jquery.js"]`)
* Tests create a new store every single time, but they can optionally "preload" the store with an existing state snapshot. This lets you load the world at any point in time, and make sure that the necessary state transitions happen after certain actions.

This is just a prototype, this isn't merge-able but wanted to talk about our strategy.

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -2,3 +2,6 @@
 .*node_modules/cjson.*
 .*node_modules/fbjs.*
 .*node_modules/npm.*
+
+[libs]
+flow/mocha.js
--- a/flow/mocha.js
+++ b/flow/mocha.js
@@ -1,0 +1,3 @@
+
+declare var describe: (name: string, func: () => void) => void;
+declare var it: (desc: string, func: () => void) => void;
--- a/public/js/actions/index.js
+++ b/public/js/actions/index.js
@@ -1,3 +1,5 @@
+// @flow
+
 const constants = require("../constants");
 const breakpoints = require("./breakpoints");
 const eventListeners = require("./event-listeners");
--- a/public/js/actions/tests/newSource.js
+++ b/public/js/actions/tests/newSource.js
@@ -1,10 +1,9 @@
 const { constants, selectors, createStore } = require("../../utils/test-head");
-const fixtures = require("../../test/fixtures/foobar.json");
 const { getSourceById } = selectors;
-const sourcesFixtures = fixtures.sources.sources;
-
-const store = createStore();
 const expect = require("expect.js");
+
+const fixtures = require("../../test/fixtures/todoSources.json");
+const fakeSources = fixtures.sources;
 
 // Write our own `newSource` to bypass the batching logic.
 function newSource(source) {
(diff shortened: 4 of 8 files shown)


jasonLaster:
There's a lot that I really like here:
+ normalizing actor IDs is great (can we do this with store fixtures as well)
+ I like pre-loading with actions, it feels very natural. We might want to do something with a mock client so that you call an action w/ the right params and get the correct client response.

keep going

jlongster:
> normalizing actor IDs is great (can we do this with store fixtures as well)

Not sure how we'd do it with the entire store fixture. The idea here is that we would use an entire store snapshot a lot less. Ideally we'd just "build up" the state we need fresh each time by firing the necessary actions (see the top-level actions that build up predefined states).

It'd be hard to go through the entire store and automatically figure out which ids to turn into some other key of the object it's pointing to. I'm optimistic that a lot of that state isn't state that we actually care to snapshot. But there might be parts of the app state that we do, not sure yet...

jlongster:
We'll talk more about this on Friday and evolve it into a merge-able PR.

jlongster:
Replied to parts of your other comment in here:

> This conversation, prompted me to re-consider one part of your testing exploration. The tests are currently invoking actions like newSource with protocol data. The better approach would be to invoke client events with protocol data. Here are two examples: ...

I don't totally agree... I think we need to restrict the scope of the tests to be more unit-y. Testing actions and reducers together is already something I wasn't initially convinced about, but I've come around to it. If I'm working on something it's nice to be fixing/writing tests along the way, and that gets harder as the tests cover more ground, because I have to have more implemented before I can even run the tests. For a new feature with your approach, I'd have to have the client commands, actions, and reducers, all done before I could run tests.

The nice thing about redux is it exposes actions/reducers as just functions, and it makes it so nice to do this kind of testing. Mocking the client is about the same is at is with either of our approaches; it's just a matter of calling them directory or not.

> The client transforms the protocol data into a form for the actions/store. 

I don't really know what you mean. We need to store something in `fakeSources`, and it'll likely be whatever the protocol gives us. The client only knows how to transform browser-specific data into our generic structure, it doesn't (and shouldn't) do anything else. The sources fixtures should probably be whatever is passed into an action.

To your `addBreakpoint` example:

> The client is a more natural place to mock commands. When we get to testing commands like addBreakpoint, the natural flow involves
>
>    1. simulating a UI addBreakpoint call with data that could come from the Editor
>    2. mocking the client, so that the correct protocol response is returned.

#1 is exactly calling an action :) And #2 is exactly what we do right now, if you look at the `breakpoints.js` test. We mock the protocol so it returns the right response. So that kind of test is already like that: mock the client and call the action creator.

The difference is that `newSource` is meant to be invoked from the client, whereas `addBreakpoint` is meant to be invoked with simple parameters. The source objects are a bit verbose, but we can create fixtures for them, either by hand or automatically.

To summarize:

* Calling actions and making sure app state is correct is a really nice and simple way to write tests that, while not unit tests, are constrained enough to be easy to refactor alongside action/reducer work. The more things a test include (like the client API), the more tests that need to be refactored when something changes
* This shouldn't block your desire to change the sources state. We can still do this, just slightly different than how I was thinking
* We *should* have tests for the client to make sure that it is transforming protocol data correctly into our general types

jasonLaster:
Hey @jlongster just wanted to share my initial exploration with action testing

It is similar in theory to what you're doing, but attempts to record the client data:

+ client protocol events are logged
+ client command arguments and results are logged

I'll do some work to show a POC when i have some time.

Also, I did some work to scrub the protocol data. The goal is to generate the smae fixture data every time.
+ stripped actor ids of connection data
+ i plan on truncating text and standardizing timestamps


I left the app store in the fixture data so that components could be be rendered at that snapshot. Of course, it's also possible to load the store with relevant actions, but at the moment it still makes sense to me to have the store's for convenience.

jlongster:
superseded by #528

PR #418, opened 2016-07-25
(WIP) Save Pretty Print Sources

jasonLaster:
**do not merge**

WIP pretty print patch for saving pretty sources and showing them in the editor w/ breakpoint support.

Architecture:
* saves a new "pretty" source, with a different url and id. Works well for source maps as the map points from ugly source to pretty source
* locations do a better job of maintaining column data, which is useful for going from generated locations to original locations when the server comes back with an actual breakpoint location.
* moves source text from a state field to a source field. I found it simpler to reason about and the patch was minor. I prefer thinking about sources w/ text that could either be not loaded, loading, done, but it's not a big deal
* caching came for free.

What works well:
* source is shown
* breakpoints work
* tabs reflect different sources

What's yet to do:
* source tree shows the formatted source
* the formatted source has a pretty button. should be hidden
* clean up the code, it's kinda a mess

screen shot 2016-07-25 at 6 20 50 pm

### DIFF
--- a/public/js/reducers/sources.js
+++ b/public/js/reducers/sources.js
@@ -7,13 +7,12 @@ const fromJS = require("../utils/fromJS");
 const I = require("immutable");
 const makeRecord = require("../utils/makeRecord");
 
-import type { Action, Source } from "../actions/types";
+import type { Action, Source, SourceText } from "../actions/types";
 import type { Record } from "../utils/makeRecord";
 
 export type SourcesState = {
   sources: I.Map<string, any>,
   selectedSource: ?any,
-  sourcesText: I.Map<string, any>,
   tabs: I.List<any>,
   sourceMaps: I.Map<string, any>,
 };
@@ -21,7 +20,6 @@ export type SourcesState = {
 const State = makeRecord(({
   sources: I.Map(),
   selectedSource: undefined,
-  sourcesText: I.Map(),
   sourceMaps: I.Map(),
   tabs: I.List([])
 } : SourcesState));
@@ -85,21 +83,14 @@ function update(state = State(), action: Action) : Record<SourcesState> {
       break;
 
     case "TOGGLE_PRETTY_PRINT":
-      if (action.status === "error") {
-        return state.mergeIn(["sourcesText", action.source.id], {
-          loading: false
-        });
-      }
-
-      let s = _updateText(state, action, [action.source]);
       if (action.status === "done") {
-        s = s.setIn(
+        return state.setIn(
           ["sources", action.source.id, "isPrettyPrinted"],
           action.value.isPrettyPrinted
         );
       }
-      return s;
 
+      break;
     case "NAVIGATE":
       // Reset the entire state to just the initial state, a blank state
       // if you will.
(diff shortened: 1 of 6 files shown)


public/js/reducers/sources.js
```diff
@@ -13,15 +13,13 @@ import type { Record } from "../utils/makeRecord";
...
   selectedSource: ?any,
-  sourcesText: I.Map<string, any>,
   tabs: I.List<any>,
   sourceMaps: I.Map<string, any>,
 };
 
 const State = makeRecord(({
   sources: I.Map(),
   selectedSource: undefined,
-  sourcesText: I.Map(),
```
jasonLaster: `sourcesText` was moved to a field in `source`. This simplified some of the pretty print explorations and afterwards seemed like a cleaner abstraction.

jlongster: Haven't gone through all of this PR yet, but I will tomorrow. Thanks for pointing this out, because I think you already know that I think it should stay separate :)

This is a case where I want the data to be more denormalized, like a separate table in a db. A small nit that now the property is `sources.text.text`. My bigger complaint is that now the sources are not the canonical "protocol objects" as I call them: objects that are given to use from the client. The nice thing about that is that it's easy to back and forth. For example, in my WIP PR I snapshotted the source objects into this file: https://github.com/jlongster/debugger.html/blob/4b79f1f14ead1620fe01dbb549920d891d9f4246/public/js/test/fixtures/todoSources.json With those, I can pick them out and immediately send them into the `newSource` action. Changing the structure means I can't always go back and forth between actions and selectors.

I would like for the same constraint to be held to all protocol objects stored in the reducer, for example the `loadedObjects` field. It makes it easy to reconstruct state and work with the data in all of our testing infrastructure.

I'd like to hear more about your "pretty print explorations", or maybe I will see them when I look closer at this PR. We can write code to "convert" this state back into protocol objects when we need it, but I'd like to see what this actually made simpler.

.eslintrc
```diff
@@ -91,8 +91,6 @@
     "comma-style": [2, "last"],
     // Warn about cyclomatic complexity in functions.
     "complexity": 2,
-    // Require return statements to either always or never specify values.
-    "consistent-return": 2,
```
jlongster: Why drop this? We'll be inconsistent with the rest of devtools

jlongster:
Thanks a lot for working through this, it's really complicated stuff and great to see these features being implemented!

public/js/reducers/sources.js
jasonLaster: Thanks for the writeup. Also, I apologize for leaving the `sourceText` change in the PR. It was not needed for pretty printing and should have been a separate PR / discussion. I'm going to take it out of this PR and open a second PR where we can have a discussion when you're back.

I think I hear you saying that you want the shape of the store to reflect the protocol data. e.g. separate message "new source", "source text", "loaded object" are separate fields somewhat like tables. One of the main benefits is ease of testing w/ the protocol data and actions. 

Personally, I'd like the shape of the store to be well shaped for the selectors / UI consumers. A couple of observations, we don't know what future clients we will support. It's possible we can overfit for one client and then the store is it's shape because of legacy (read support multiple languages). 

This conversation, prompted me to re-consider one part of your testing exploration. The tests are currently invoking actions like `newSource` with protocol data. The better approach would be to invoke client events with protocol data.  Here are two examples:

current action approach:
```js
store.dispatch(newSource(fakeSources["backbone.js"]));
```

preferred client event approach:
```js
const events = require("public/js/clients/firefox/events");
setupEvents({actions: boundActions, threadClient: () => {}});

events.newSource(undefined, fakeSources["backbone.js"]);
```

There are a couple of reasons why I think the client is the right place to mock:

The client transforms the protocol data into a form for the actions/store. 
+ Take `newSource` for example, it will set `isPrettyPrinted` to false and ignore some URLs. 
+ It's very likely that tests will fail because actions like `newSource` don't expect un-transformed protocol data
+ If we start changing the fixture data to be consistent with how the client transforms the data, there could be inconsistencies.

The client is a more natural place to mock commands. When we get to testing commands like `addBreakpoint`, the natural flow involves 
+ simulating a UI  `addBreakpoint` call with data that could come from the Editor
+ mocking the client, so that the correct protocol response is returned. The good news is that the protocol already involves request/response, so it should be possible to find the correct server response given request data.


Summary (that was a lot of words)
+ I'm dropping the sourceText change from this PR (should never have been here)
+ I'm opening a second PR w/ the sourceText change. I don't want to drag out the change, but I think it's an important discussion.
+ I'm going to add a comment on your testing proposal. This way we can have the right conversations in a focused way.

.eslintrc
jasonLaster: oops, meant to open a separate PR. completely agree that this makes us inconsistent w/ devtools. I feel like we can diverge in this way as we have `flow` for types, but I don't feel strongly.

jasonLaster:
Merged. ughh, this pr got too big. became very hard to get right...

PR #422, opened 2016-07-26
handle debugger statements like breakpoints?

clarkbw:
**not for merge**

I put this together last night really quickly.   I just wanted to offer it up to others and to gather feedback / discussion on the ideas / approach so we could handle it in the future.

As you can see the debugger already handles the paused state well when it encounters a `debugger;` statement in the code; the editor opens to it and highlights it.  However because it isn't a breakpoint set by the debugger it works very differently even though it is fairly similar to a breakpoint from the user perspective.

screen shot 2016-07-26 at 10 23 25

The change as implemented listens for the `PAUSED` action and specifically for the one that is related to a `debugger;` statement in the code; defined as `"debuggerStatement"` from the Firefox client.  This listening happens in the breakpoints reducer and creates a breakpoint in the state.

So the question is:
> Should debugger statements appear and act as breakpoints?

If so:
* can they be disabled by the debugger?
* should they have a market in the editor?
 * can they be disabled from here?
 * is it a different kind of marker?
* should they appear in the breakpoints list?
 * how are they labeled?
 * can they be disabled from here?

### DIFF
--- a/public/js/reducers/breakpoints.js
+++ b/public/js/reducers/breakpoints.js
@@ -42,6 +42,23 @@ function makeLocationId(location: Location) {
 
 function update(state = State(), action: Action) {
   switch (action.type) {
+    case "PAUSED": {
+      const pause = action.pauseInfo;
+      if (pause.why.type === "debuggerStatement") {
+        const locationId = makeLocationId(pause.frame.location);
+        const frameId = pause.frame.id;
+        const bp = state.breakpoints.get(locationId) ||
+          { location: pause.frame.location, condition: undefined };
+        const text = "debugger statement";
+        return state.setIn(["breakpoints", locationId], updateObj(bp, {
+          id: frameId,
+          disabled: false,
+          loading: false,
+          text: text
+        }));
+      }
+      break;
+    }
     case "ADD_BREAKPOINT": {
       const id = makeLocationId(action.breakpoint.location);
 


jlongster:
Yes, it's been a feature request before to be able to "disable" debugger statements for the current session, so this is definitely good thinking. I like this. It makes sense to use the existing enable/disable breakpoint functionality for debugger statements.

The UX is a little weird to me that it's only "discovered" when it pauses on a debugger statement though. I don't know if there's a way around that though. I definitely don't want to ask the JS engine for all "debugger" statements, as that sounds very heavy and error-prone. Maybe that part of the UX is acceptable, and gives us all the other functionality for free.

jasonLaster:
@clarkbw could you put this work in a new like `addDebuggerBreakpoint` and feature flag it off? That way we can land the work and iterate.


```js
function addDebuggerBreakpoint() {
  if (!feature.isEnabled("features.debuggerBreakpoint")) {
    return;  
  }
}
```

clarkbw:
> feature flag it off

Sure, I'll take a look later today.

jlongster:
I think we should be careful about adding too much code that is just experiments. Perhaps it's better to just push them as branches that we can refer back to later. I don't think we have time right now to do this, and I'd much rather leave all the code out of the master branch. Feature flags shouldn't be abused to commit code that's not ready.

clarkbw:
> Perhaps it's better to just push them as branches that we can refer back to later

Its just as easy to close this out for now.  I think we can keep experimenting with it and I like having the record of discussion / code / screenshots somewhere until we have a better process.

SebastianZ:
@clarkbw [wrote](https://github.com/devtools-html/debugger.html/pull/422#issue-167668695):
> The change as implemented listens for the PAUSED action and specifically for the one that is related to a debugger; statement in the code; defined as "debuggerStatement" from the Firefox client. This listening happens in the breakpoints reducer and creates a breakpoint in the state.

@jlongster [wrote](https://github.com/devtools-html/debugger.html/pull/422#issuecomment-235431860):
> The UX is a little weird to me that it's only "discovered" when it pauses on a debugger statement though. I don't know if there's a way around that though. I definitely don't want to ask the JS engine for all "debugger" statements, as that sounds very heavy and error-prone.

Did you consider using the [Parser API](https://developer.mozilla.org/en-US/docs/Mozilla/Projects/SpiderMonkey/Parser_API)?
With that API it's quite easy to get the `debugger` statements and provides better UX, because it doesn't rely on the JS engine to run the code.

**Example:**
```javascript
Components.utils.import("resource://gre/modules/reflect.jsm");
Reflect.parse("var i = 5;\ndebugger;");
```

Sebastian

wldcordeiro:
@SebastianZ interesting API, only problem I see is that this project is aiming at working in more than Firefox and that seems to be a Firefox API. If Node/Chrome have something similar than it's more feasible.

jlongster:
There's a catch, though: we don't even have the source text to parse. We *definitely* do not want to eagerly fetch all the source text, and then also parse all of them. That would be a huge performance problem. We are fine-tuning the debugger to be zippy even on huge sites so that's a no go.

This would have to be provided by the debugger server using C++ APIs to query the `ScriptSource` object specifically.

SebastianZ:
> This would have to be provided by the debugger server using C++ APIs to query the `ScriptSource` object specifically.

Right. So I guess you'll need to contact the other browser vendors about this (or just implement this feature in Firefox' RDP and hope others take over the idea).

Sebastian

Issue #246, opened 2016-06-07
Top Level breakpoints cause an error when pausing


jasonLaster:
When a top level breakpoint is paused, the pause action is called before the new sources action. This happens because new sources is delayed by 10ms.

When that happens, select source fails because the sources store is empty

jasonLaster:
#274 is a duplicate

jasonLaster:
Fixed by #428

PR #427, opened 2016-07-27
Cleanup sources tree

jasonLaster:
I started looking into #246 and noticed several things in SourcesTree that I wanted to clean up.

Nothing in here should be controversial

### DIFF
--- a/public/js/components/SourcesTree.js
+++ b/public/js/components/SourcesTree.js
@@ -15,7 +15,24 @@ const ManagedTree = React.createFactory(require("./utils/ManagedTree"));
 const FolderIcon = React.createFactory(require("./utils/Icons").FolderIcon);
 const DomainIcon = React.createFactory(require("./utils/Icons").DomainIcon);
 const FileIcon = React.createFactory(require("./utils/Icons").FileIcon);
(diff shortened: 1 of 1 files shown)


public/js/components/SourcesTree.js
```diff
@@ -30,33 +47,35 @@ let SourcesTree = React.createClass({
   },
 
   componentWillReceiveProps(nextProps) {
-    if (nextProps.sources !== this.props.sources) {
-      if (nextProps.sources.size === 0) {
-        this.setState(createTree(nextProps.sources));
-        return;
-      }
+    if (nextProps.sources === this.props.sources) {
```
jasonLaster: flipped the test so that it could be a guard statement and avoid one level of nesting.

public/js/components/SourcesTree.js
```diff
@@ -69,6 +88,18 @@ let SourcesTree = React.createClass({
...
+    if (depth === 0) {
+      return domain;
+    }
+
+    if (!nodeHasChildren(item)) {
+      return file;
+    }
+
+    return folder;
+  },
```
jasonLaster: extracted this logic.

Issue #245, opened 2016-06-07
Editor breakpoints are missing when the debuggee is refreshed


jasonLaster:
Steps:
1. add breakpoints
2. refresh debuggee

The breakpoints show up in the list, but are not shown in the editor

jlongster:
This bug is for the server. The server does not maintain breakpoints across debugging sessions, and we cannot approximate that from the client because if we try to re-set them on refresh there are breakpoints that will definitely be missed because the code has already executed.

jasonLaster:
Fixed.

PR #428, opened 2016-07-27
Debounce Sources Tree

### LINKED ISSUE #246: Top Level breakpoints cause an error when pausing

jasonLaster:
This should fix #246 and #245.

The big change here is that we're going from debouncing newSource actions to debouncing the source tree render.

Why is this better:
+ the order of client events matters. by debouncing `newSource` actions it's possible a `pause` action is dispatched before the sources have been stored. When this happens `selectSource` is dispatched before the source is stored. 
+ Ultimately, the problem we should be solving is render performance for the source tree and there are many ways to do that. This is a simple approach for this case, if other components also need to have renders throttled, we can re-evaluate.

Disclaimer:
+ This causes the sources tree to always be collapsed. I think we can fix this after this work lands, but I quickly looked into it. What I found was a strange behavior where without the debouce behavior, SourcesTree.render was called an even number of time for depth 0 nodes, each time an item was called isExpanded would be flipped. This behavior is very odd and should be fixed to be based on whether an item has children.

### DIFF
--- a/public/js/actions/sources.js
+++ b/public/js/actions/sources.js
@@ -20,25 +20,6 @@ const { getSource, getSourceText, getSourceByURL, getGeneratedSource,
         getSourceMap, getSourceMapURL, getOriginalSources
       } = require("../selectors");
 
-/**
- * Throttles source dispatching to reduce rendering churn.
- */
-let newSources = [];
-let newSourceTimeout;
-
-function _queueSourcesDispatch(dispatch) {
-  if (!newSourceTimeout) {
-    // Even just batching every 10ms works because we just want to
-    // group sources that come in at once.
-    newSourceTimeout = setTimeout(() => {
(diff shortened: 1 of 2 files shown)


jlongster:
This is an interesting approach. It's not totally ideal, as `shouldComponentUpdate` generally shouldn't have side effects, and `isMounted` is discouraged (I'm not sure that's even available on class-based components). But it's probably better than debouncing actions (less bugs everywhere else) so we should land this and rework it later if needed. r+

jasonLaster:
> `isMounted`

true, i actually meant to replace that before making the PR, but I'm okay keeping it as long as it's available as the substitute would be to do something similar in `componentWillUnmount` to cancel the defer...

PR #431, opened 2016-07-27
denormalize sourceText

jasonLaster:
This is a follow up on the earlier discussion about de-normalizing `sourceText`.

This PR, moves sourceText from the store to a field in source. The field name is `text`, which is a poor name. I think I would prefer calling it text with a content field.

Benefits:
+ when sourceText is separate, we need to make sure that there's always a sourceText object when we want to select a source. This is a timing issue with pretty printing.

+ The reducers are simpler when they're just sources with text. 

Testing:
+ If the unit tests pre-loaded data w/ client events like `newSource(protocolFixture["backbone"])` the data would be properly setup

DISCLAIMER: I'm raising this PR so that we can discuss it. If we're not comfortable creating a more abstract store then we shouldn't. There are pros/cons in both directions and it's simpler to keep the store closer to the protocol for now, if we're not certain.

### DIFF
--- a/public/js/reducers/sources.js
+++ b/public/js/reducers/sources.js
@@ -13,7 +13,6 @@ import type { Record } from "../utils/makeRecord";
 export type SourcesState = {
   sources: I.Map<string, any>,
   selectedSource: ?any,
-  sourcesText: I.Map<string, any>,
   tabs: I.List<any>,
   sourceMaps: I.Map<string, any>,
 };
@@ -21,7 +20,6 @@ export type SourcesState = {
 const State = makeRecord(({
   sources: I.Map(),
   selectedSource: undefined,
-  sourcesText: I.Map(),
   sourceMaps: I.Map(),
   tabs: I.List([])
 } : SourcesState));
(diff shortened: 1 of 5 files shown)


jlongster:
I think this is actually "normalizing" sourceText :) Separating things out usually is "denormalized" but here it doesn't really matter because it's always a 1:1 mapping.

I commented more in https://github.com/jlongster/debugger.html/pull/415#issuecomment-235775572 about the testing stuff.

> when sourceText is separate, we need to make sure that there's always a sourceText object when we want to select a source. This is a timing issue with pretty printing.

This is interesting, I'd like to hear more about the pretty printing issue. If there is a problem with that, then I'm fine with that change. I just wanted to know more about those kinds of issues.

> The reducers are simpler when they're just sources with text.

That's interesting, because I would have thought they'd be more complicated because the data is deeper, but there might be something else I'm missing.

We can more about it on Friday! I'm not totally opposed to it.

public/js/reducers/sources.js
```diff
@@ -214,7 +208,13 @@ function getSources(state: OuterState) {
...
 
 function getSourceText(state: OuterState, id: string) {
-  return state.sources.sourcesText.get(id);
+  const source = getSource(state, id);
+
+  if (!source) {
+    return;
+  }
+
+  return source.get("text"); // eslint-disable-line consistent-return
```
jasonLaster: This could easily be `state.getIn(["sources", id, "text"])`

jasonLaster:
> I think this is actually "normalizing" sourceText 

nerd snipe. it's probably neither as denormalizing is adding redundant data to a table for read performance. I generally think of normalizing as moving entities out to their own table, but that is a simplification because it is more about trade-offs on overlapping data field. **shrug**

**pretty printing**

At this point, I have enough separation from the problem to not feel as strongly about the pros-cons. 

Couple of observations:

* patch does not change the source reducer that much. `updateText`, `getSourceText` are simple changes.

* my work ignores a valuable use case, "showing a loading message while pretty printing". If I add that, it's likely that the patches are very similar.

Given that, I think this is largely a testing and preference question, so I'm happy to close the PR. We will have plenty of time to discuss these questions at a later point.

PR #432, opened 2016-07-27
Move Source Maps to a worker

jasonLaster:
(WIP) going to review with a fresh set of eyes tomorrow.

This PR moves the source map utility behind a worker and introduces a new source-map utility that is responsible for communicating with the worker and doing some of the queries that were previously in selectors.

It is a large PR because source maps are used in many places (sources, pausing, breakpoints). Fortunately, the API changes are all the same (go from sync to async).


public/js/actions/breakpoints.js
```diff
@@ -33,8 +35,6 @@ function enableBreakpoint(location: Location) {
 function addBreakpoint(location: Location,
                        { condition, getTextForLine } : any = {}) {
   return ({ dispatch, getState, client }: ThunkArgs) => {
-    location = getOriginalLocation(getState(), location);
```
jasonLaster: I dropped these checks because it unecessary and frustrating to wrap in async. if we need them we can bring them back

public/js/actions/sources.js
```diff
@@ -240,16 +230,14 @@ function loadSourceText(source) {
       // It's already loaded or is loading
       return Promise.resolve(textInfo);
     }
-    const generatedSource = getGeneratedSource(getState(), source);
-    const originalSources = getOriginalSources(getState(), generatedSource)
-                            .map(s => s.toJS());
 
-    return dispatch({
+    getGeneratedAndOriginalSources(getState(), source)
+    .then(({ generatedSource, originalSources }) => dispatch({
```
jasonLaster: this is a little awkward, and i'll review later

public/js/clients/chrome/events.js
```diff
@@ -38,8 +38,8 @@ function paused(callFrames, reason, data, hitBreakpoints, asyncStackTrace) {
     type: reason
   }, data);
 
-  actions.paused({ frame, why });
-  actions.loadedFrames(frames);
+  actions.paused({ frame, why })
+  .then(() => { actions.loadedFrames(frames); });
```
jasonLaster: also a little awkward

jasonLaster:
Hmm, looks like tests are failing in node, in part because Worker is not defined? Does it make sense to introduce a simple node library for tests that adheres to the same API and is async? https://github.com/pgriess/node-webworker

public/js/clients/chrome/events.js
```diff
@@ -38,8 +39,8 @@ function paused(callFrames, reason, data, hitBreakpoints, asyncStackTrace) {
     type: reason
   }, data);
 
-  actions.paused({ frame, why });
-  actions.loadedFrames(frames);
+  await actions.paused({ frame, why });
+  return await actions.loadedFrames(frames);
```
jasonLaster: at some point it might make sense to move `loadedFrames` into `paused` as they're doing the same thing

package.json
```diff
@@ -95,6 +96,7 @@
     "webpack": "1.13.1",
     "webpack-dev-middleware": "^1.6.1",
     "webpack-hot-middleware": "^2.12.0",
+    "workerjs": "^0.1.1",
```
jasonLaster: added workerjs for node testing

package.json
```diff
@@ -28,6 +28,7 @@
     "immutable": "^3.7.6",
     "invariant": "^2.2.1",
     "lodash": "^4.13.1",
+    "pretty-fast": "^0.2.0",
```
jasonLaster: now that we're building the workers it's possible to require the packages directly instead of importScript.

public/js/actions/sources.js
```diff
@@ -240,16 +231,17 @@ function loadSourceText(source) {
       // It's already loaded or is loading
       return Promise.resolve(textInfo);
     }
-    const generatedSource = getGeneratedSource(getState(), source);
-    const originalSources = getOriginalSources(getState(), generatedSource)
-                            .map(s => s.toJS());
```
jasonLaster: This was setup so that we could show "loading" messages in the original source mapped locations, but it made the action async and was adding marginal benefits. 

Now, we just show "loading" in the generated source, which is fine

jasonLaster:
Changes:
+  added `workerjs` for running something similar to webworkers in node / mocha. The implementation uses child processes and behaves very similarly
+ added webpack entry points for workers.
  + workers can now require modules like `source-map` or a util
  + workers are now transpiled, which is important for node because babel-require does not pick it up
  + I am not sure what this means for the firefox panel. 
     + if we build in the dependencies it will work. 
     + if the browser loader works for workers it will work
     + otherwise we will want to use importScript instead
+ I couldn't figure out how to have karma serve the

jasonLaster:
Merged the PR as it's so large that it will block everything else. It's likely that there will be changes here, but it'll be easier to make them on top of this work rather than re-reviewing the changes in this PR.

PR #433, opened 2016-07-27
(WIP) consolidate pause action

jasonLaster:
this is a wip.

It's built on the worker work, so take a look at the second commit to see the action consolidation. To be fair, this is a small cleanup that doesn't matter all that much.


jasonLaster:
Hmm, just remembered ff paginated frames. I'm going to check and see what chrome does

jlongster:
Haven't looked too closely, but 👍  great work

jasonLaster:
If we'd like to add frame pagination for firefox later we can add that as a feature.

At the moment, we were currently showing a max of 25 frames and not re-fetching which was a bug. For what it's worth, chrome does not offer paginating.

Issue #403, opened 2016-07-22
Watch Expressions


jasonLaster:
Implement Watch Expressions in the right sidebar:

Tasks:

- [x] add / remove watch expressions 
- [x] edit an expression
- [x] show an expression value on pause or on edit

Thoughts:

* expressions stored in redux (maybe pause store, that's when they're evaluated)
* input is a formatted input (we don't need syntax highlighting now)
* value is a Rep (look at variable view for how to use them)
* evaluations are evaluated with the same client console eval api
* don't worry too much about the UI, we can style it later

**Chrome:**
![screen shot 2016-07-22 at 1 16 16 pm](https://cloud.githubusercontent.com/assets/254562/17065172/5f216114-500e-11e6-8b2f-3b85df3e5524.png)

**Firefox**
![screen shot 2016-07-22 at 1 16 05 pm](https://cloud.githubusercontent.com/assets/254562/17065173/5f2fb520-500e-11e6-944d-d83abd73d63c.png)

jbhoosreddy:
I'll work on this one now.

jasonLaster:
@jbhoosreddy we can mark this as done when we have a rough implementation where you can add expressions that are evaluated on pause.

+ Object Inspector
+ Edit 
+ Delete

can all come later. This is a big feature and aught to be broken up into smaller pieces

jbhoosreddy:
@jasonLaster thanks. It works to that extent, so I'm going to make a PR tonight. I was actually looking into the httpGet issue now.

jasonLaster:
@jbhoosreddy did a great job in https://github.com/jlongster/debugger.html/pull/434. The first pass has the ability to add and edit watch expresions.

We'll do a second pass soon:
+ look at comments in #434 
+ add remove expressions
+ format with the Object Inspector

jasonLaster:
@jbhoosreddy You've done a great job here. I'm going to close this master ticket and we can always open smaller ones to tweak the UX. I know you have some ideas here.

PR #434, opened 2016-07-28
add watch expressions feature (#403)

jbhoosreddy:
So I managed to get it to work somehow. but it is still a WIP. @jasonLaster. I haven't even considered the UX yet.


public/js/components/Expressions.js
```diff
@@ -0,0 +1,69 @@
...
+
+  displayName: "Expressions",
+
+  componentWillUpdate(nextProps, nextState) {
+    const pauseInfo = nextProps.pauseInfo;
+    const expressions = nextProps.expressions.toJS();
+    if (pauseInfo) {
+      for (let expression of expressions) {
+        nextProps.evaluateExpression(expression);
+      }
```
jasonLaster: I don't think we want to re-evaluate all of the expressions here. It is probably better to re-evaluate expressions in the pause action.

public/js/components/Expressions.js
```diff
@@ -0,0 +1,69 @@
...
+    }
+  },
+
+  addExpression(e) {
+    if (e.key === "Enter") {
+      const id = this.props.expressions.toSeq().size++;
```
jasonLaster: I think the ID can be created in the `addExpression` action.

public/js/reducers/pause.js
```diff
@@ -59,6 +60,21 @@ function update(state = initialState, action, emit) {
     case constants.PAUSE_ON_EXCEPTIONS:
       const toggle = action.toggle;
       return state.set("shouldPauseOnExceptions", toggle);
+
+    case constants.ADD_EXPRESSION:
+      return state.setIn(["expressions", action.id],
+        { id: action.id,
+          expression: action.expression,
+          value: action.value || null });
```
jasonLaster: the action.value should never be null.

public/js/reducers/pause.js
```diff
@@ -59,6 +60,21 @@ function update(state = initialState, action, emit) {
...
+
+    case constants.EVALUATE_EXPRESSION:
+      if (action.status === "done") {
+        return state.setIn(["expressions", action.id],
```
jasonLaster: I'm not sure i f you want to use an array or a map. I suppose both work here, but it'd be bad if the array became sparse when expressions are removed...

jasonLaster:
Great job @jbhoosreddy! This is looking really good

jbhoosreddy:
So I finalized whatever I had by the end of today and hid it behind a feature flag. Will come back to it later.

jasonLaster:
Thanks. This is really good work and is close to merging

public/js/actions/pause.js
```diff
@@ -36,6 +37,15 @@ function paused(pauseInfo) {
   return ({ dispatch, getState, client }) => {
     let { frame, frames, why } = pauseInfo;
 
+    for (let expression of getExpressions(getState())) {
+      dispatch({
```
jasonLaster: You can call `addExpression(expression)` here

public/js/components/Expressions.js
```diff
@@ -0,0 +1,66 @@
...
+    evaluateExpression: PropTypes.func,
+    loadObjectProperties: PropTypes.func,
+    command: PropTypes.func
+  },
+
+  displayName: "Expressions",
+
+  addExpression(e) {
+    if (e.key === "Enter") {
+      const id = this.props.expressions.toSeq().size++;
```
jasonLaster: I'd prefer seeing the ID in the action and keeping the component simpler.

public/js/components/RightSidebar.js
```diff
@@ -20,6 +21,24 @@ function debugBtn(onClick, type, className = "active") {
...
+  { header: "Call Stack",
+    component: Frames },
+  { header: "Scopes",
+    component: Scopes }
+  ];
+  if (isEnabled("features.watchExpressions")) {
+    items.unshift({ header: "Watch Expressions",
+      component: Expressions,
+      opened: true });
+  }
```
jasonLaster: great!

public/js/reducers/pause.js
```diff
@@ -66,6 +67,21 @@ function update(state = initialState, action, emit) {
...
+    case constants.ADD_EXPRESSION:
+      return state.setIn(["expressions", action.id],
+        { id: action.id,
+          expression: action.expression,
+          value: action.value });
+
+    case constants.EVALUATE_EXPRESSION:
+      if (action.status === "done") {
+        return state.setIn(["expressions", action.id],
+          { id: action.id,
```
jasonLaster: this can be `state.mergeIn` and only merge the value

public/js/components/Expressions.js
```diff
@@ -0,0 +1,66 @@
...
+    const expressions = this.props.expressions.toJS();
+    return dom.span(
+      { className: "pane expressions-list" },
+      dom.input(
+        { type: "text",
+          className: "input-expression",
+          placeholder: "Add watch Expression",
+          onKeyPress: this.addExpression }
+      ),
+      expressions.map(expression => {
```
jasonLaster: Immutable has a valueSeq() method, which has a map.

public/js/components/Expressions.js
```diff
@@ -0,0 +1,66 @@
...
+        return dom.li(
+          { key: expression.id },
+          expression.expression + JSON.stringify(expression.value)
+        );
+      }),
+      info("Not paused")
```
jasonLaster: lets drop "not paused"

public/js/reducers/pause.js
jlongster: We could probably pull these watch expressions into their own reducer, as the whole notion of "pause" state is too broad in my opinion. But that's something that can be done later.

jasonLaster:
Looks great. Thanks @jbhoosreddy

public/js/actions/pause.js
```diff
@@ -126,7 +129,46 @@ function loadObjectProperties(grip) {
...
+      type: constants.UPDATE_EXPRESSION,
+      id: expression.id,
+      input: expression.input
+    });
+  };
+}
+
+function evaluateExpressions() {
+  return ({ dispatch, getState, client }) => {
+    for (let expression of getExpressions(getState())) {
```
jasonLaster: It would be nice to have an `evaluateExpression` action, which can be called when an expression is updated.

public/js/components/Expressions.js
```diff
@@ -0,0 +1,89 @@
...
+            id: "expressionInput-" + expression.id,
+            placeholder: "Add watch Expression",
+            onKeyPress: this.addExpression,
+            defaultValue: expression.input }
+        ) :
+        dom.span(
+          { key: expression.id,
+            id: "expressionOutput-" + expression.id,
+            onClick: this.updateExpression },
+          expression.input + " --> " +
```
jasonLaster: this could be multiple `div`s, which will improve styling and make it easier to grab values out of later.

PR #436, opened 2016-07-28
(discussion) Add translate function

jasonLaster:
One of the big blockers for landing the new debugger is translating our strings with l10n. 

The approach we're planning on taking advantage of is a function that is a no-op in local development and l10n in the panel. 

I started working on the no-op function which will let us make sure we have string coverage before we land in the panel.

Notice the styles that highlight the matched text in an obnoxious red. The goal here is that when we want to check it should be easy to see the strings.

screen shot 2016-07-28 at 6 09 51 pm

### DIFF
--- a/public/js/components/App.css
+++ b/public/js/components/App.css
@@ -76,3 +76,10 @@ body {
   text-align: center;
   width: 100%;
 }
+
+.translate-string {
+  /*
+  background-color: red;
+  color: yellow;
+  */
+}
--- a/public/js/components/Tabs.js
+++ b/public/js/components/Tabs.js
@@ -1,6 +1,7 @@
 const React = require("react");
 const { connect } = require("react-redux");
 const { getTabs } = require("../selectors");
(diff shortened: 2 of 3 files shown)


jlongster:
We should probably call it `formatStr` or whatever the current L10N API uses for consistency, and because we will need this function to take parameters (for localizing strings like "has % items"). Also, the current l10n system does not work like this. Each string is given a name and the API is `formatStr("sourceLabel")`. The strings are listed in a properties file with these keys.

We should probably hold off on this until we really know more about our strategy. Are we wanting to leverage the current system? It seems like we should if we want to land for real soon, in which this PR doesn't really work.

I also don't think we need to localize the toolbox. Lots of strings there and it's only for local development, and we don't have a way to localize strings for things outside of mozilla-central.

jasonLaster:
closing this PR, we can discuss an approach tomorrow during sprint planning

PR #442, opened 2016-07-29
(wip) tweaks to make workers work in Firefox context

jlongster:
I had a little difficulty making this work. The config system feels a bit awkward, and hard to figure out what's pulling in what. We can't pull in the firefox config if `process.env.NODE_ENV === "production"`, because that value has nothing to do with the target. We should be able to run dev/prod locally, and dev/prod in Firefox.

The problem is that we pull in config for firefox at build time, so `DebuggerTarget` does not exist because that's defined at build time. I had to add a parameter to `getConfig` but I don't love it. Just wanted to show what's needed; we can tweak this after feedback into something merge-able.

### DIFF
--- a/public/js/selectors.js
+++ b/public/js/selectors.js
@@ -46,8 +46,7 @@ function getSelectedFrame(state: AppState) {
 }
 
 function getSourceMapURL(state: AppState, source: any) {
-  const tab = getSelectedTab(state);
-  return tab.get("url") + "/" + source.sourceMapURL;
+  return source.url + "/" + source.sourceMapURL;
 }
 
 /**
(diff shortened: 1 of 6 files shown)


public/js/selectors.js
```diff
@@ -46,8 +46,7 @@ function getSelectedFrame(state: AppState) {
 }
 
 function getSourceMapURL(state: AppState, source: any) {
-  const tab = getSelectedTab(state);
-  return tab.get("url") + "/" + source.sourceMapURL;
+  return source.url + "/" + source.sourceMapURL;
```
jlongster: Needed to tweak this also because the selected tab does not exist in the Firefox context, but I think the sourceMapURL should be resolved relative to the source. This should also be a smarter `join(source.url, source.sourceMapURL)`.

I can move this out of the PR, just wanted to show everything that we need to work on.

jlongster:
Superseded by #452

PR #444, opened 2016-07-29
Pause on top-level breakpoint

### LINKED ISSUE #245: Editor breakpoints are missing when the debuggee is refreshed

jasonLaster:
Fixes #245 

Use Case:
+ add a breakpoint
+ refresh debuggee
+ see the pause location
+ see the breakpoint in the gutter

### DIFF
--- a/public/js/components/EditorBreakpoint.js
+++ b/public/js/components/EditorBreakpoint.js
@@ -26,12 +26,23 @@ const Breakpoint = React.createClass({
 
   displayName: "Breakpoint",
 
-  componentWillMount: function() {
+  addBreakpoint: function() {
     const bp = this.props.breakpoint;
     const line = bp.location.line - 1;
-
     this.props.editor.setGutterMarker(line, "breakpoints", makeMarker());
     this.props.editor.addLineClass(line, "line", "breakpoint");
+  },
+
+  componentDidMount: function() {
+    if (!this.props.editor) {
+      return;
+    }
+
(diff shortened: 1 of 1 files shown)


public/js/components/EditorBreakpoint.js
```diff
@@ -26,14 +26,25 @@ const Breakpoint = React.createClass({
 
   displayName: "Breakpoint",
 
-  componentWillMount: function() {
+  addBreakpoint: function() {
     const bp = this.props.breakpoint;
     const line = bp.location.line - 1;
-
     this.props.editor.setGutterMarker(line, "breakpoints", makeMarker());
     this.props.editor.addLineClass(line, "line", "breakpoint");
```
jlongster: This function will now fire twice whenever the editor is rerendered (and thus the breakpoints). This could be solved by implementing `shouldComponentUpdate`:

```js
shouldComponentUpdate(nextProps) {
  return this.props.editor !== nextProps.editor;
}
```

Now, it will only add the breakpoint if the editor changes (which will happen if it was `null` before).

jasonLaster:
merged

jlongster:
Can you either merge through github or reference the commit here? We need to make sure you can view the actual committed changes from every PR. It's hard to see if you addressed my comment or not.

Issue #439, opened 2016-07-29
Right Sidebar element wrapping


jasonLaster:
When the right sidebar is narrow, there are weird wrapping that should be fixed with an overflow hidden. Chrome has the right behavior here, which is effectively to allow the user to shrink the UI, without attempting to be responsive at a certain point. The UI won't be usable, but that is okay as the user could always expand the sidebar later.

PR #448, opened 2016-07-30
Right Sidebar element wrapping (#439)

JasonWeathersby:
I just added to the right side bar css file. I am having issues testing in chrome but this appears to work in Firefox. Note that when you pause on a breakpoint we still get a one line wrap on the source name but I would think that is what you would want it to do.


jasonLaster:
Looks great. 

Did you see the chrome contributing section?

PR #452, opened 2016-08-01
make debugger work in Firefox again and add script for generating Firefox bundle

jlongster:


jasonLaster:
Looks good!

I think we also need to copy the two workers right?

jlongster:
Oh, yep! I already did that manually, good catch!

PR #453, opened 2016-08-01
Add Key Shortcuts library

### LINKED ISSUE #360: Investigate a key shortcut library like mousetrap
We're running into issues with keyboard shortcuts that could be fixed with a mature library like [mousetrap](https://github.com/ccampbell/mousetrap).

[file search does not work](https://github.com/jlongster/debugger.html/issues/354#issuecomment-233634994)

jasonLaster:
Fixes #360.

This PR adds the key-shortcuts library, Services Shim, and EventEmitter.

Notes:
+ key shortcuts library gives us an api for setting keyboard events. It did not support preventDefault, so i added a patch that I will also land in m-c.
+ services shim - adds lots of useful properties that are de-privileged. I'm most excited about prefs. Prefs however tried tor read local storage items with the assumption that they were all JSON parseable, which was not the case because todomvc shoves some random stuff in there. Consequently I moved the debugger examples to port 7999.
+ EventEmitter is a general purpose utility in m-c. It is being added to our utils directory because I had to rip out the `logEvent` function which assumes dump is available. There's also a lot of factory setup, which is not necessary. We might be able to reconcile these changes by injecting a dump console.log with webpack's define plugin if we want to

### DIFF
--- a/bin/development-server.js
+++ b/bin/development-server.js
@@ -1,4 +1,5 @@
 #!/usr/bin/env node
+
 "use strict";
 
 require("babel-register");
@@ -51,7 +52,6 @@ if (feature.isEnabled("hotReloading")) {
 }
 
 // Static middleware
-app.use(express.static("public/js/test/examples"));
 app.use(express.static("public"));
 
 // Routes
@@ -75,4 +75,10 @@ app.listen(8000, "0.0.0.0", function(err, result) {
   }
 
   console.log("Development Server Listening at http://localhost:8000");
+});
+
+const examples = express();
+examples.use(express.static("public/js/test/examples"));
+examples.listen(7999, "0.0.0.0", function(err, result) {
+  console.log("View debugger examples at http://localhost:7999");
 });
--- a/public/js/components/App.js
+++ b/public/js/components/App.js
@@ -15,6 +15,7 @@ const SourceFooter = createFactory(require("./SourceFooter"));
 const Autocomplete = createFactory(require("./Autocomplete"));
 const { getSelectedSource, getSources } = require("../selectors");
 const { endTruncateStr } = require("../utils/utils");
+const { KeyShortcuts } = require("../lib/devtools-sham/client/shared/key-shortcuts");
 
 const App = React.createClass({
   propTypes: {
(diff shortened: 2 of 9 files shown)


public/js/components/App.js
```diff
@@ -32,18 +33,16 @@ const App = React.createClass({
...
 
   componentDidMount() {
-    window.addEventListener("keydown", this.toggleSourcesSearch, false);
+    this.shortcuts = new KeyShortcuts({ window });
+    this.shortcuts.on("Cmd+P", this.toggleSourcesSearch, { preventDefault: true });
   },
 
   componentWillUnmount() {
-    window.removeEventListener("keydown", this.toggleSourcesSearch, false);
+    this.shortcuts.off("Cmd+P", this.toggleSourcesSearch);
```
bgrins: You can do `CmdOrCtrl+P` which will handle control on windows/linux or command on osx

bin/development-server.js
```diff
@@ -76,3 +76,9 @@ app.listen(8000, "0.0.0.0", function(err, result) {
 
   console.log("Development Server Listening at http://localhost:8000");
 });
+
+const examples = express();
+examples.use(express.static("public/js/test/examples"));
+examples.listen(7999, "0.0.0.0", function(err, result) {
+  console.log("View debugger examples at http://localhost:7999");
```
jlongster: We already have a bunch of messages and it's starting to feel like spam... do we need this message?

public/js/components/App.js
```diff
@@ -15,6 +15,7 @@ const SourceFooter = createFactory(require("./SourceFooter"));
 const Autocomplete = createFactory(require("./Autocomplete"));
 const { getSelectedSource, getSources } = require("../selectors");
 const { endTruncateStr } = require("../utils/utils");
+const { KeyShortcuts } = require("../lib/devtools-sham/client/shared/key-shortcuts");
```
jlongster: Nice!

jlongster:
This is awesome. This comes from m-c? I've never looked at this lib. What are we going to do about maintenance? Since we are moving the repo, should we start spinning this libs off into separate repos? That would make them easier to upgrade/maintain/keep in sync with m-c.

public/js/components/App.js
jasonLaster: Thanks. Good suggestion.

bin/development-server.js
jasonLaster: Hmm, I think it might be nice to promote the different debugger examples. Let me create a separate issue for this as it's off topic

jasonLaster:
> Since we are moving the repo, should we start spinning this libs off into separate repos?

I think this is a decision that for the larger devtools team. At the moment, I'm okay keeping m-c files in sync. It would certainly be easier to make modules like KeyShortcuts their own libraries and pull them in, like source-map, but that is a larger discussion

jlongster:
My only concern is this is a big lib to have in devtools-sham, which we aim to get rid of. How much did you have to change? We can't just blindly copy the files in devtools-sham because those ones need tweaks. How do we update this lib? We should document the changes you made so we know how to keep them in sync.

jlongster:
To be clear (we talked about this on IRC) I think we should merge this but continue thinking about how to track this stuff.

jasonLaster:
NOTE: the try/catch in services is temporary and will be removed next week when we upgrade to cypress 0.17.

PR #455, opened 2016-08-01
s/jlongster/devtools-html/

clarkbw:
Most things will redirect properly, but there’s no reason to create confusion.


jasonLaster:
nice

PR #456, opened 2016-08-01
Add editor story

jasonLaster:
This adds a second story for the Editor, which has a selected source and some breakpoints. Unfortunately, I found that for toggling between stories breaks the editor. I believe it has something to do with the provider not supporting changing the store, but i'm not sure.


jlongster:
Looks good, r+

jasonLaster:
Thanks, i'll open a second issue for toggling stories. I'm kinda interested in this as i've seen the same thing for unit tests.

Issue #437, opened 2016-07-29
Feature Flag blackbox 


jasonLaster:
Currently, the blackbox icon is shown in the bottom right of the editor. We should add a feature flag, so that is off by default.

jasonLaster:
Fixed by #477

PR #457, opened 2016-08-01
add watchExpressions feature flag and related docs

clarkbw:
Need to add flag docs for `watchExpressions`.  This situation also has me thinking that I want a CI test or bot who looks for feature flag usage and warns if you didn't add it to the development.json file; too big a job for now.

I also wanted to have some docs to point to to help guide issues like #437

### DIFF
--- a/docs/feature-flags.md
+++ b/docs/feature-flags.md
@@ -1,0 +1,40 @@
+## Feature Flags
+
+The debugger.html project has a system for [feature flags](https://en.wikipedia.org/wiki/Feature_toggle), a system that allows us to ship certain features which are off by default.  Features must be in active development with the intention of landing in the core; feature flags are not intended for landing broken code or features which are not under active development.
+
+Feature flags toggle features that are either:
+* experimental or in a testing phase
+* designed to be configured by the user
+
+Periodically feature flags are removed as features are either out of the experimental phase and merged into the core or are removed and no longer available.  In the future we may turn on certain features for subsets of the population.
+
+## Configure
+
+All feature flags are configured in the [config directory](./config/).  See the [config/README](./config/README.md) for descriptions of each flag.
+
+## Development
+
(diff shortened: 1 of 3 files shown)


docs/feature-flags.md
```diff
@@ -0,0 +1,40 @@
+## Feature Flags
+
+The debugger.html project has a system for [feature flags](https://en.wikipedia.org/wiki/Feature_toggle), a system that allows us to ship certain features which are off by default.  Features must be in active development with the intention of landing in the core; feature flags are not intended for landing broken code or features which are not under active development.
+
+Feature flags toggle features that are either:
+* experimental or in a testing phase
+* designed to be configured by the user
```
jasonLaster: that will be handled by prefs, which is separate.

docs/feature-flags.md
```diff
@@ -0,0 +1,40 @@
...
+
+Search for [isEnabled](https://github.com/devtools-html/debugger.html/search?utf8=%E2%9C%93&q=isEnabled) in the code to find more examples.
+
+```js
+// within the components directory you can require isEnabled
+const { isEnabled } = require("../feature");
+
+// feature check can be done in render() method
+render() {
+  if (!isEnabled("features.pokemon-go")) {
```
jasonLaster: this is slightly off topic, but there are two functions `isEnabled` and `getValue`. Currently they're only different because `isEnabled` forces a truthy check. I'd like to change that so, isEnabled is for feature flags and getValue is for config values

This would have the added benefit of changing `isEnabled("features.pokemon-go")`  to `isEnabled("pokemon-go")`

clarkbw: That would be a much nicer system.  I'll leave this example for now and will update it when this changes.  I may also file an issue for this work as you've described.

docs/feature-flags.md
clarkbw: Oh right, I actually meant for this to be "other people who embed this debugger", not end users.

jasonLaster:
I'll update the bullet in a separate commit

Issue #414, opened 2016-07-25
Standardize SVG components


jasonLaster:
Now that we're using an svg loader, we can standardize the way we use SVGS

- The `icons` file might not be necessary
- Other places that use inline svgs can be simplified a lot as well

PR #462, opened 2016-08-02
Standardize svg components

jbhoosreddy:
Standardize SVG components (#414)


jlongster:
Looks like this need to be rebased, can you do that?

jasonLaster:
Work looks great! It came out really cleanly. Also, love that we don't need to import the svgs into m-c now.

One comment: the sources tree file probably doesn't need separate consts on top, we can just use them directly.

jbhoosreddy:
Closing this PR and created a new one at #470

PR #465, opened 2016-08-02
allow selecting sources by URL that don't exist yet

jlongster:
I don't love this code. But this is an extremely useful use case for outside interaction. This way anyone can select a source and know that it will be shown, whether or not it exists yet. I'm open to renaming things in here, not sure what the best names are.


jasonLaster:
This makes sense. Pending is fine in terms of naming. 

I can see something like obsevables handling this case generally, but this is great for now.

jlongster:
I just realized that the current code for this in the toolbox returns true or false to indicate if it was able to open the source. If `false`, meaning it couldn't find the source in the debugger, it opens it in a new "view-source" tab. Theoretically, it's a bug when that happens because logs should always be from a source in the current window. But this could happen in several cases: browser console shows logs from all sorts of places, a bug in sourcemaps may log a URL that the debugger doesn't have for some reason, etc. I don't think we can drop that fallback case right now. 

The main problem is that means we need know when all the current sources are loaded in the debugger, so we can check if the source exists afterwards. Right now we have no such concept; sources come in as they are created. We need the ability to see if the debugger is fully initialized or not, basically, and if not wait for the initial set of sources to load.

Because the `view-source` functions in the toolbox need the fallback, they might need to handle more logic then just firing a `selectSourceURL` action. But this code works for now. I'm landing a bundle in Firefox that uses this, but it will fail if the source never exists in the debugger.

We may remove this code and take a different approach in the future though.

jasonLaster:
Interesting. Chrome has the idea of frameNavigated, which might be a useful event to fire on the firefox side:

https://chromedevtools.github.io/debugger-protocol-viewer/1-1/Page/#event-frameNavigated

PR #468, opened 2016-08-02
Add chrome tests for source maps

### LINKED ISSUE #405: Source Maps - Integration Tests
In addition to having unit tests that mock out different aspects (e.g. client) we should have at least one integration test that covers the most common cases.

jasonLaster:
Fixes #405. 

Notes:
* Adds an env flag for debugging chrome
* Adds the "Inspectable Pages" tab to the list
* Adds a test for debugging an app with source maps

Note: the test is actually flakey now and half the time does not pause even though the breakpoint is added and button clicked... Looking into the protocol now

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -5,9 +5,11 @@ test:
     - node_modules/.bin/cypress ci b07646ab-ddfa-442b-b63f-aebf00452de8
     - node_modules/.bin/cypress ci b07646ab-ddfa-442b-b63f-aebf00452de8
   pre:
-    - node bin/development-server.js:
+    - DEBUG_CHROME=true node bin/development-server.js:
         background: true
     - node bin/firefox-driver --start:
+        background: true
+    - /opt/google/chrome/google-chrome --remote-debugging-port=9222 --no-first-run:
         background: true
     - sleep 30
   post:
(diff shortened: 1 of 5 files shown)


jasonLaster:
I disabled the test so that i can look into the flakiness later. I believe it's a timing issue with the connection and when commands are being issued.

jlongster:
Don't we already download Firefox for the integration tests? Is it possible to download Firefox Dev Edition so we can test against Firefox? I wonder if Firefox wouldn't be flaky (and I'd rather be consistently testing Firefox right now).

jasonLaster:
> testing firefox

That's a good point. I'd like to merge this work as it will be useful in the future and then follow up by downloading firefox.

> flakiness

I'm not sure if it's chrome specific, we'll see.

jlongster:
I'd like to be mindful about committing code that needs a lot of follow-up... what's the benefit of committing it now? We can leave this PR open until at least the test is running (I'm fine with Firefox as a follow-up)

jasonLaster:
@jlongster mind taking another look now that the tests pass. Note, I added #482 to switch to firefox soon. After that, it would be nice to have CI run both firefox and chrome.

jlongster:
Looks fine to me! Looks like one of the cypress runs failed? Not sure if that's related or not...

jasonLaster:
Hmm, interesting. It was very consistent locally, but it might be slower on CI. You might have noticed that I added some pauses in there for good measure. I actually did not test it without them when it started working, but it might be at play here. I can look into this later if it's an issue

PR #470, opened 2016-08-03
standardize SVG icons (#414)

jbhoosreddy:
solved the merge issues and making a new PR.


jasonLaster:
r+. This is a really nice cleanup.

I think we can merge after we land the firefox bundle.

public/images/Svg.js
```diff
@@ -0,0 +1,36 @@
...
+  "subSettings": require("./subSettings.svg"),
+  "worker": require("./worker.svg")
+};
+
+module.exports = function(name, props) { // eslint-ignore
+  let className = props ? `${name} ` + props.className : name;
+  if (name === "subSettings") {
+    className = "";
+  }
+  props = Object.assign({}, props, { className, src: svg[name] });
```
jlongster: Can you throw an error if it's an unknown SVG? This PR is definitely a big improvement but we should make sure that typos/etc are easily caught:

```js
if(!svg[name]) { throw new Error("Unknown SVG: " + name); }
```

jlongster:
Yeah this is really good. It's nice that we can change how we load SVGs in one place also. Just had one small comment.

PR #472, opened 2016-08-03
Make source tree item key unique

### LINKED ISSUE #435: Expand the Source Tree
When we debounced the source tree, the source tree stopped expanding by default

When I logged the tree renders, I noticed that the expanded flag alternated between being true or false. This was surprising as I would have thought that it should be dependent on whether the item has children.

```patch
diff --git a/public/js/components/SourcesTree.js b/public/js/components/SourcesTree.js
index e19e277..d138f9f 100644
--- a/public/js/components/SourcesTree.js
+++ b/public/js/components/SourcesTree.js
@@ -119,6 +119,9 @@ let SourcesTree = React.createClass({
   },
 
   renderItem(item, depth, focused, _, expanded, { setExpanded }) {
+    if (depth == 0) {
+      console.log(item.path, expanded);
+    }
     const arrow = Arrow({
       className: classnames(
         { expanded: expanded,
```

jasonLaster:
Fixes #435. It's possible that two sources will have the same path. This caused two issues:
+ we got a react warning that children had the same key
+ it was causing the tree to be collapsed initially

screen shot 2016-08-03 at 3 09 35 pm

### DIFF
--- a/public/js/components/SourcesTree.js
+++ b/public/js/components/SourcesTree.js
@@ -158,7 +158,7 @@ let SourcesTree = React.createClass({
         return [];
       },
       getRoots: () => sourceTree.contents,
-      getKey: (item, i) => item.path,
+      getKey: (item, i) => item.path + i,
       itemHeight: 30,
       autoExpandDepth: 2,
       onFocus: this.focusItem,


public/js/components/SourcesTree.js
```diff
@@ -158,7 +158,7 @@ let SourcesTree = React.createClass({
         return [];
       },
       getRoots: () => sourceTree.contents,
-      getKey: (item, i) => item.path,
+      getKey: (item, i) => item.path + i,
```
jlongster: You don't want to do this because it destroys the purpose of the key. The keys should stay consistent for each item with the renders, so that React can easily figure out which ones have simply moved. This makes React think items have been removed and need to be re-added.

The point of `item.path` is to always be unique. When is that not unique? That's a bug, and we should fix the path generation if there is a bug in there.

jlongster:
> it was causing the tree to be collapsed initially

That's been a recent bug and it worked before with the path. Two sources can't have the same path; if they do, the latter will "replace" the previous one in the source tree (basically: we don't support this right now, nor do we in the old debugger). The only time this can happen is if the same source is loaded multiple times and that's pretty rare.

I'd check out why you have multiple sources with the same paths. The problem most likely is where that's happening.

jasonLaster:
@jlongster - i believe the path issue is coming from source maps. The screenshot above shows two sources with the filename "(index)". This only showed up after I added the index into the key. I agree that key should be explicit so that it doesn't change. Not sure how to handle the source path issue if that's coming back from the source-map library, but i can investigate.

jlongster:
Check out the sourcemap and the URLs that it maps to. I'm curious why it would emit a `http://localhost:8000/increment/` URL. Usually the sourcemapped URLs don't (can't) resolve to the actual local host, they resolve to something like `webpack://`, and also that means that the sourcemap bundled in an HTML file? If you can find out why that URL is in there, we can figure out how to handle that situation.

jasonLaster:
Yeah, that's a good point

jlongster:
Maybe we can append the source's actor ID, so the path + actorID would always be unique but also be consistent across renders. We do want to support this eventually, because the same script can be loaded multiple times, but it'd be nice to be explicit about when that can happen... To me this points to a bug somewhere else, but I could be wrong.

jasonLaster:
@jlongster I think you're right. I'll double check this morning, but I believe the problem is a result of chrome not clearing the sources list on navigation events, which is fixed in another bug

jasonLaster:
I'm going to close and if we need to do something, re-open a new PR

PR #473, opened 2016-08-03
Handle Page navigation

### LINKED ISSUE #466: Source Tree should reset when debuggee is navigated
Here's a screenshot of the debugger showing sources for todomvc and increment after i navigated from one page to the other. This has been a known issue.

screen shot 2016-08-02 at 4 37 54 pm

jasonLaster:
Fixes #466 

+ clear breakpoints on navigation
+ listen to chrome navigation events

### DIFF
--- a/public/js/clients/chrome/events.js
+++ b/public/js/clients/chrome/events.js
@@ -6,6 +6,7 @@ function setupEvents(dependencies) {
   actions = dependencies.actions;
 }
 
+// Debugger Events
 function scriptParsed(scriptId, url, startLine, startColumn,
              endLine, endColumn, executionContextId, hash,
              isContentScript, isInternalScript, isLiveEdit,
@@ -49,6 +50,21 @@ function resumed() {
 function globalObjectCleared() {
 }
 
+// Page Events
+function frameNavigated(frame) {
+  actions.navigate();
+}
+
+function frameStartedLoading() {
+  actions.willNavigate();
+}
+
+function domContentEventFired() {}
+
+function loadEventFired() {}
+
+function frameStoppedLoading() {}
+
 const clientEvents = {
   scriptParsed,
   scriptFailedToParse,
@@ -57,7 +73,16 @@ const clientEvents = {
   globalObjectCleared,
 };
 
+const pageEvents = {
+  frameNavigated,
+  frameStartedLoading,
+  domContentEventFired,
+  loadEventFired,
+  frameStoppedLoading
+};
+
 module.exports = {
   setupEvents,
+  pageEvents,
   clientEvents
 };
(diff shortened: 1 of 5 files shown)


public/js/test/cypress/commands/debugger.js
```diff
@@ -42,16 +42,19 @@ function scopeAtIndex(index) {
...
   cy.wait(1000);
   cy.reload();
   cy.wait(1000);
 }
 
+function debugFirstTab(browser = "Firefox") {
+  cy.visit("http://localhost:8000");
+  cy.get(`.${browser} .tab`).first().click();
+  cy.wait(1000);
+}
```
jasonLaster: minor tweak to make it easier to write cypress tests

public/js/clients/chrome/events.js
```diff
@@ -49,6 +50,21 @@ function resumed() {
 function globalObjectCleared() {
 }
 
+// Page Events
+function frameNavigated(frame) {
```
jlongster: Nice, didn't realize we didn't support navigation events for Chrome.

public/js/reducers/breakpoints.js
```diff
@@ -123,6 +123,9 @@ function update(state = State(), action: Action) {
       }
 
       break;
+    }
+    case "NAVIGATE": {
+      return State();
```
jlongster: We don't actually want to do this, for Firefox at least. We keep breakpoints around, just like the server does. If you navigate back to a source, all the breakpoints should still exist. A key use case is refreshing: all the breakpoints should still exist, right? A refresh is a type of navigation.

Not sure if you're seeing some problem in Chrome, but I'm pretty sure this will break Firefox. (it'll pause on a top-level breakpoint after a refresh but the breakpoint won't exist in the UI)

jasonLaster: That's a good point. I think this is actually a question of using a selector that only queries for breakpoints at that domain or breakpoints that have a source in the store, which can be difficult.

jasonLaster:
6b66ecbc48edb1bc301e8eb3951acda84c8d707

(oops, forgot to merge through the UI)

public/js/reducers/breakpoints.js
jlongster: > I think this is actually a question of using a selector that only queries for breakpoints at that domain or breakpoints that have a source in the store, which can be difficult.

Oh, that's a really good point. I forgot that the breakpoints list is going to show all of them. We probably should restrict `getBreakpoints` to filter out ones that are only associated with current sources. Hm, I hadn't really thought about that, there might be a more elegant 

> I think this is actually a question of using a selector that only queries for breakpoints at that domain or breakpoints that have a source in the store, which can be difficult.

Oh, that's a really good point, I forgot that the breakpoints list is going to show all of them. `getBreakpoints` should definitely only return breakpoints for the current page... I like the idea of checking if the source exists. Not sure if there's a more elegant way to handle this...

PR #481, opened 2016-08-04
Clear source maps on navigation

### LINKED ISSUE #480: Source Maps - source fails to load after navigation
Steps to reproduce:
+ go to page with source maps
+ refresh page
+ select original source

The issue, is that the source map consumers and nodes are not being cleared during page navigations, so the old consumers are still around.

jasonLaster:
Fixes #480. 

The problem is that source map data in the worker is not being cleared during debuggee navigation.

There are a couple ways to clear the data:
+ data is kept in a weak map with object references
+ data is kept in a reducer, which observes the navigation event
+ the navigate action calls a source map method to clear the data

We discussed this problem a week or two ago and agreed it would be nice to use either option 1 or 2. Unfortunately, I'm not sure how to do that when the data is kept in the worker. I don't think option 3, is that bad though, as it is quite common for an action to do the side-effect work, whether it's talking to a stateful store, client, or in this case worker through an API.

### DIFF
--- a/public/js/actions/index.js
+++ b/public/js/actions/index.js
@@ -1,28 +1,11 @@
-const constants = require("../constants");
 const breakpoints = require("./breakpoints");
 const eventListeners = require("./event-listeners");
 const sources = require("./sources");
 const tabs = require("./tabs");
 const pause = require("./pause");
-
-function willNavigate() {
-  return { type: constants.NAVIGATE };
-}
-
-function navigate() {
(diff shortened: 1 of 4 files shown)


jasonLaster:
As a side note, I'm very happy to share that this fix solves the source map intermittent issue as well, which is very cool.

screen shot 2016-08-04 at 10 46 21 am

jlongster:
Woo! That's great.

PR #483, opened 2016-08-04
Exclude internal sources

jasonLaster:
found this small bug while comparing chrome and firefox source events for the sources tree expand/collapse issue.


jlongster:
r+

Issue #482, opened 2016-08-04
Switch source map tests to firefox


jasonLaster:
We can test source maps with firefox if we test against nightly. We should try doing that as it will give us a better idea of whether the debugger is likely to work in the firefox panel.

NOTE:
+ we'll likely need to start using marionette as the selenium backend

```
 <AutomatedTester> jlast: if you use http://seleniumhq.github.io/selenium/docs/api/javascript/module/selenium-webdriver/lib/capabilities_exports_Capabilities.html and have one for "marionette":true that *should* turn on Marionette
<AutomatedTester> jlast: or if there is a Selenium 3 beta on npm use that
```

jasonLaster:
I did a little work to install and start firefox nightly:

https://github.com/devtools-html/debugger.html/compare/master...jasonLaster:test-firefox-latest

This script successfully downloads the latest build, but fails to untar it.

I also tried running the driver with the latest firefox and confirmed that we will need to use marionette to drive the latest firefox. We have two options here, the first is to launch firefox directly with the appropriate profile. The other is to use the driver + marionette. At this point, I think starting firefox is simpler, so we just need a way to include the profile. This patch includes a zipped profile, which i can easily unzip on circle, but I'm not sure where the profiles live in linux. On mac they're in Application Support...

jlongster:
Thanks for looking into this! Definitely something to look into more soon.

clarkbw:
@jasonLaster is this still in progress?

jasonLaster:
This is no longer relevant.

PR #484, opened 2016-08-04
Test against firefox nightly

jasonLaster:
Should download the latest firefox and test against that.

NOTE: i'm pretty sure the firefox-driver will fail to start it, but i'm curious to see if firefox installs correctly


jasonLaster:
closing for now... I left some comments in #482 with progress notes

PR #485, opened 2016-08-04
Update isEnabled api

### LINKED ISSUE #467: improve isEnabled and getValue config functions
There are two functions `isEnabled` and `getValue`. Currently they're only different because `isEnabled` forces a truthy check. I'd like to change that so, isEnabled is for feature flags and getValue is for config values

This would have the added benefit of changing `isEnabled("features.pokemon-go")`  to `isEnabled("pokemon-go")`

via https://github.com/devtools-html/debugger.html/pull/457#discussion_r73161268

jasonLaster:
Fixes #467.

### DIFF
--- a/config/feature.js
+++ b/config/feature.js
@@ -1,3 +1,5 @@
+"use strict";
+
 const pick = require("lodash/get");
 let config;
 /**
--- a/config/feature.js
+++ b/config/feature.js
@@ -1,5 +1,3 @@
-"use strict";
-
 const pick = require("lodash/get");
 let config;
 /**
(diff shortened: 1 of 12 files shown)


config/feature.js
```diff
@@ -10,7 +8,9 @@ function getValue(key) {
   return pick(config, key);
 }
 
-const isEnabled = getValue;
+function isEnabled(key) {
+  return pick(config, `features.${key}`);
```
jlongster: Why use `pick` here? `features[key]` should be fine. One thing I really don't like about lodash is it doesn't error on `null`, so `pick(null, ...)` just returns an empty object.

jlongster:
Looks good besides the one nit!

config/feature.js
jasonLaster: that's a good point.

jasonLaster:
Hmm, running into a couple cases now where we do isEnable checks for non-features. Will update the PR so we can asses. I think this is why i held off last time. 

I still think it's a good idea to limit the `isEnabled` API.

PR #487, opened 2016-08-04
expose actions to rest of devtools

jlongster:
This is a stopgap solution so the Firefox panel has the ability to select a source from other places like the console. Eventually, we probably want to be explicit about the global API we expose and only expose things like `selectSource`.

### DIFF
--- a/public/js/main.js
+++ b/public/js/main.js
@@ -41,6 +41,13 @@ if (isDevelopment()) {
   injectGlobals({ store });
 }
 
+// Expose the bound actions so external things can do things like
+// selecting a source.
+window.actions = {
+  selectSource: actions.selectSource,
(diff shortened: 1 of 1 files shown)


jasonLaster:
@jlongster, could we expose a subset of the actions like so:

```js
const { selectSource } = actions;
window.actions = { selectSource }
```

I prefer an option like this as every time someone wants to use an action, they need to make the case for it being public. Also, maybe it makes sense to wrap this in a environment flag:

```js
if (isFirefoxPanel) {
  const { selectSource } = actions;
  window.actions = { selectSource }
}
```

this way, we are not tempted to use this inside of the debugger for any reason.

jlongster:
Yep, that's what I meant about being explicit, and I thought about doing it later but I can do it now.

PR #488, opened 2016-08-04
Revert "Add Key Shortcuts library (#453)"

jlongster:
This reverts commit 804ddda68495b625b2a700a1819e4aed0a87c343.

The bundle to be committed to Firefox reverted this, because it broke the Firefox panel for some obscure reason. Usually these are easy to fix, but hard to figure out exactly what is breaking. Once we land we should re-land this and look into why it breaks the Firefox panel.


jlongster:
Opening a PR because there were conflicts that I fixed so I want to run CI.

jasonLaster:
Looks good, minus the test failure which can easily be fixed.


The cypress tests have to be updated to use localhost:8000 as the example:
https://github.com/devtools-html/debugger.html/blob/master/public/js/test/cypress/commands.js#L49

jasonLaster:
updating the test failures manually

jasonLaster:
Pulled the revert out as jarda's PR https://github.com/devtools-html/debugger.html/pull/495 depends on `event-emitter`, which was included by the `key-shortcuts` work. Lets try to fix key-shortcuts in place on monday.

PR #489, opened 2016-08-04
Fix source map breakpoints

### LINKED ISSUE #486: Source Maps - pausing on breakpoints
It looks like breakpoints in an original source no longer works for firefox. It does work for chrome though.

Steps to reproduce:
- go to increment in firefox
- add a breakpoint on increment.js line 3
- click the increment button

Note:
- adding a breakpoint in (index) on line 12 does work.

Also, attempting to remove the breakpoint throws this error:
```
promise.js:46 Uncaught (in promise) Object {from: "server1.conn19.child1/31", error: "noSuchActor", message: "No such actor for ID: server1.conn19.child1/31"}
```

jasonLaster:
Fixes #486.

There was an issue where firefox would not pause when a breakpoint was set in an original source. It turns out that the issue stemmed from the column being passed in as 0, which firefox treated as a column number that had to be deciphered. 

https://dxr.mozilla.org/mozilla-central/source/devtools/server/actors/source.js#859

### DIFF
--- a/public/js/clients/firefox/commands.js
+++ b/public/js/clients/firefox/commands.js
@@ -46,6 +46,7 @@ function sourceContents(sourceId) {
 
 function setBreakpoint(location, condition) {
   const sourceClient = threadClient.source({ actor: location.sourceId });
+
   return sourceClient.setBreakpoint({
     line: location.line,
     column: location.column,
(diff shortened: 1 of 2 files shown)


public/js/clients/firefox/commands.js
```diff
@@ -46,9 +46,12 @@ function sourceContents(sourceId) {
 
 function setBreakpoint(location, condition) {
   const sourceClient = threadClient.source({ actor: location.sourceId });
+
+  const column = location.column === 0 ? undefined : location.column;
```
jlongster: I don't want to do this here. If you pass a column of 0, it should send a column of 0. Firefox's behavior makes sense: if it's a real value then it should only set a breakpoint on that column.

You should move this logic up to source maps and decipher the column info there.

jasonLaster: yeah, that's a good point.

PR #490, opened 2016-08-04
Disable source maps + blackboxing

jasonLaster:
This is not technically needed, but I feel better explicitly disabling these two features so that we can a) operate confidently that they are disabled, b) rip out the functionality later.


jlongster:
Sure, sounds good!

PR #491, opened 2016-08-04
Fix #334 - editor loses context

### LINKED ISSUE #334: file search loses source contents
If I have a source open, press cmd+P to bring up the search bar, and press cmd+P again to hide it, all the source contents are gone.

jasonLaster:
The editor is blank after a file search. This is due to the fact that a render happens with the default text "..." and the source text is not updated.

We fix this by always calling setText when the component mounts, in most cases it's a no-op, but it does fix the case where the editor is dis-mounted and then mounted again.

### DIFF
--- a/public/js/components/Editor.js
+++ b/public/js/components/Editor.js
@@ -60,6 +60,7 @@ const Editor = React.createClass({
       );
     }
 
+    this.setText(this.props.sourceText.get("text"));
     resizeBreakpointGutter(this.editor);
   },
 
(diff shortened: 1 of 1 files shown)


public/js/components/Editor.js
```diff
@@ -106,16 +107,26 @@ const Editor = React.createClass({
...
-      this.editor.setValue(newSourceText.get("text"));
-    }
+    this.setText(newSourceText.get("text"));
 
     resizeBreakpointGutter(this.editor);
   },
 
+  // Only reset the editor text if the source has changed.
+  // + Resetting the text will remove the breakpoints.
+  // + Comparing the source text is probably inneficient.
```
jlongster: nit: to be consistent, the rest of devtools uses stars, not pluses

jlongster:
Nice, I had noticed that so it's great that you fixed it!

PR #495, opened 2016-08-05
Use the final version of websocket-transport.js

jsnajdr:
The WebSocket debugging server has landed today [bug 1286281](https://bugzilla.mozilla.org/show_bug.cgi?id=1286281), and the final version of the debugger transport can be used without any change in debugger.html. Updating and moving from `devtools-sham` to `devtools`.


jasonLaster:
Thanks @jsnajdr. Looks good.

Don't worry about the test failures, that's something I broke recently

PR #496, opened 2016-08-05
(WIP) Client logging

### LINKED ISSUE #494: Client - log commands and events
It can be very useful to see the client protocol over the course of a session and diagnose what is going wrong. 

Hopefully we can have a system that is capable of 
+ console logging events + commands
+ storing events + commands in a log store that can be written to disk or viewed holistically

jasonLaster:
Fixes #494.

This adds two features: logging for the client and fixture improvements for saving the data in a sanitized way.

Note, I'm not trying to purely log the protocol for commands, i'm more interested in the params/return value for the client.

Here's a sample [fixture](https://gist.github.com/jasonLaster/d8c0e86e4df94702a64bb8bcfbec747f). 

I'll have to update the storybook stories and unit tests that use the fixtures to read just the app state before this lands


jlongster:
There are two unrelated things here, so it might be nice to have 2 PRs. Client logging and sanitizing the app state. I'm not convinced we should be taking on either though.

Client logging: I'd much rather just use the browser's ability to log data. In Firefox this is easy and everyone on the devtools team already knows how to do this. We know it's not buggy and it is the *real* protocol log. I really don't think we should be taking this on in our client.

App state sanitizing: if you want to separate that out in a different PR, I'm fine merging it. (I don't see how it depends on client logging?) But I still convinced we don't want full app state fixtures. I'll be working on testing this week and will show a more fleshed out version of #415 to show why. Sanitizing helps reduce changes sometimes, but whenever we tweak the app state it'll still massively change the fixtures. That's not terrible, but I think we can get away with saving a lot less data. But like I said I'd be fine merging that part.

jasonLaster:
Thanks for the thoughts. My takeaway at the moment is that this work is largely dependent on your plans for testing the actions.

**Client Logging**
+ I agree that the browser is sufficient for logging the protocol
+ The client logging I added is designed to facilitate a client mocking API:

for example, a test could simulate either an event or command and under the hood we'd find the appropriate event or command. In the case of commands, it's more helpful to record the input/output of the command than the request/response of the protocol.

```js
mockClient.newSource("todo-view");
mockClient.setBreakpoint("todo-view", 33);
mockClient.setBreakpoint("todo-view", 45);
```

Either way, we can punt on the client logging until we have a better idea of how we want to test the actions.

**app state sanitizing**

I agree that this is up in the air as well. If we have a great way to simulate actions, then we don't need entire app state fixtures. I'll PR this up anyways as it's nice to have in the interim and might be useful.

jlongster:
> The client logging I added is designed to facilitate a client mocking API:

I guess I was focusing too much on the console logging, but that's true. Let's leave this open for another day or two as we discuss testing. I'm currently writing tests that are the simple action/reducer kind, and I'd like to get more experience with that before closing this. Right now I'm manually mocking and it feels good because I can dynamically do whatever I want (like the test that tests a failed source text loading).

jasonLaster:
> manually mocking and it feels good because I can dynamically do whatever I want

Great, can't wait to see how it goes. Really eager to write some tests as well.

jlongster:
Now that #528 has landed, I have a better idea of what's possible with simple unit tests. I think you can take them pretty far. I just dynamically generate sources as needed and built up the state I need from scratch every time, which allows a lot of flexibility for each test to do whatever it wants without needing any fixtures or protocol data. All the tests: https://github.com/devtools-html/debugger.html/blob/master/public/js/actions/tests/sources.js And an example: https://github.com/devtools-html/debugger.html/blob/master/public/js/actions/tests/sources.js#L68-L76

I'm still not completely sure what we'll use this protocol data for. It's easier to build up the state with fake sources or breakpoints and render stuff out. We do need to test the protocol, but we should probably just test the real thing as an integration test anyway to make sure it works and returned the right data.

jasonLaster:
After seeing your work on action unit tests, i think it's nice to continue trying to dynamically generating data source and otherwise. 

If we want to capture more data we can always come back to this.

PR #497, opened 2016-08-05
Exclude debugger eval sources

jasonLaster:
It looks like debugger evals were being added as sources. There's an ignore clause above that checks for debugger code eval source urls, is that needed?

### DIFF
--- a/public/js/clients/firefox/events.js
+++ b/public/js/clients/firefox/events.js
@@ -62,6 +62,11 @@ function newSource(_, packet) {
   if (NEW_SOURCE_IGNORED_URLS.indexOf(source.url) > -1) {
     return;
   }
+
+  if (source.introductionType == "debugger eval") {
+    return;
+  }
+
   actions.newSource(Source({
     id: source.actor,
     url: source.url,


public/js/clients/firefox/events.js
```diff
@@ -62,6 +62,11 @@ function newSource(_, packet) {
   if (NEW_SOURCE_IGNORED_URLS.indexOf(source.url) > -1) {
     return;
   }
+
+  if (source.introductionType == "debugger eval") {
```
jlongster: The ignore clause above should do this, yes. Try to figure out why that's not working. We shouldn't need this clause, just fix the `(NEW_SOURCE_IGNORED_URLS` patterns.

jasonLaster: @jlongster I'm not sure where the regression is, but I'm seeing this:

screen shot 2016-08-08 at 3 01 14 pm

jlongster:
Ooh you're talking about `introductionType`. Yeah, you can merge this PR. I'm not sure where we ignore them in the other debugger, but this works! r+

jasonLaster:
Oh yeah - interesting. I'll be on the lookout for that

PR #498, opened 2016-08-05
Stub debuggee evaluation

jasonLaster:
@jbhoosreddy and I noticed when we were working on adding an integration test for watch expressions that the expressions were not evaluated until the debugger had resumed. 

The reason for this was that the debugger pauses in the integration tests by running a console evaluation like `dbg.click("button")`, which does not return until the debugger resumes. The console has a mechanism for only running one evaluation at a time, so it blocks subsequent evals. 

We get around this problem by creating a new evaluate method that includes a timeout, which will simulate an evaluation response. In the case of firefox, we also have to clear the internal pending requests store.

### DIFF
--- a/public/js/components/Frames.js
+++ b/public/js/components/Frames.js
@@ -52,11 +52,13 @@ function Frames({ frames, selectedFrame, selectFrame }) {
 
 module.exports = connect(
   state => ({
-    frames: getFrames(state).map(frame => {
-      return Object.assign({}, frame, {
-        source: getSource(state, frame.location.sourceId).toJS()
-      });
-    }),
+    frames: getFrames(state)
+      .filter(frame => getSource(state, frame.location.sourceId))
+      .map(frame => {
+        return Object.assign({}, frame, {
+          source: getSource(state, frame.location.sourceId).toJS()
+        });
+      }),
     selectedFrame: getSelectedFrame(state)
   }),
   dispatch => bindActionCreators(actions, dispatch)
--- a/public/js/clients/chrome/commands.js
+++ b/public/js/clients/chrome/commands.js
@@ -81,6 +81,11 @@ function evaluate(script) {
   });
 }
 
+function debuggeeCommand(script) {
+  evaluate(script);
+  return Promise.resolve();
+}
+
 function navigate(url) {
   return pageAgent.navigate(url, (_, result) => {
     return result;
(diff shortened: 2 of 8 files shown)


jlongster:
This seems really hacky, and I'd like to avoid it. Adding `setTimeout`s is not a good way to solve things, and you're accessing internal state of the protocol. I don't have enough context of watch expressions to suggest a solution though, but this seems like we're just adding technical debt and will pay for it later. This is why I was concerned about adding watch expressions because it's taking time away from other things.

I'd encourage you or @jbhoosreddy to think of any other solution that doesn't involve `setTimeout`, or anything we can change in our testing infrastructure to avoid it. We're going to run into this a lot I suspect as we add other features which require evaluation, and I don't think this solution is going to hold up.

jasonLaster:
Yeah, this is an interesting problem that's connected to sharing the debugger protocol between the tests and debugger:

1. test simulates adding a todo by evaluating `dbg.type("input", "hello", "Enter")`
2. debugger attempts to run a watch expression, but finds that the first evaluation is still pending.

I think the best way to handle this that is likely to hold up over time is to introduce a new client method `debuggeeCommand` that issues a console evaluation, but does not wait for a response. In this way, the test still drive the debuggee, but won't be stopped if the debuggee pauses, which is often the case. If there's a time when a test wants to check a debuggee value, it can still use the regular evaluate method, which returns the response.

jasonLaster:
@jlongster just updated the PR and came up with something much simpler.

+ the timeout is gone as the function is more focused
+ there's much less internal state
+ `debuggeeCommand` is only for tests (not any eval w/ a timeout)

Also, as an interesting aside, working on the test helped me find an interesting Frames bug related to console evaluations. We should be excluding the eval frames from the frames list. This now does that. We might want to do it more upstream, either in the reducer when adding the frame data or selector, but there's an open question in my mind when it comes to one reducer querying another's state.

@jbhoosreddy the watch expression test is not showing the updated evaluation value. Once something lands, could you look into that and see why it's not being updated.

public/js/components/Frames.js
```diff
@@ -52,7 +52,9 @@ function Frames({ frames, selectedFrame, selectFrame }) {
 
 module.exports = connect(
   state => ({
-    frames: getFrames(state).map(frame => {
+    frames: getFrames(state)
+    .filter(frame => getSource(state, frame.location.sourceId))
```
jlongster: nit: indentation is little weird here. Might be nice to indent it a little more to connect it with the first line?

jlongster:
Looks a lot better. Not exactly sure of the implications, but we can try it out. r+

PR #505, opened 2016-08-06
Fix unit tests

jasonLaster:
@amitzur and I just found two things:

1. the unit tests fail to load svg modules (svg-inline is not available)
2. the unit tests were not running on circle ci

+ We fixed svgs, by mocking the require call.
+ We fixed circle, by updating the test call

I'm not sure how to include the reporter option (`mochaFile=$CIRCLE_TEST_REPORTS/mocha/junit.xml`).


jlongster:
Looks good, r+

PR #506, opened 2016-08-08
add/remove watch expressions and format with Rep (#403)

jbhoosreddy:
- Ability to add and remove expressions.
- Formatting with Rep


public/js/components/Expressions.js
```diff
@@ -3,66 +3,111 @@ const { connect } = require("react-redux");
...
+      key: expression.id },
       expression.updating ?
-        dom.input(
-          { type: "text",
-            className: "input-expression",
-            id: "expressionInput-" + expression.id,
-            placeholder: "Add watch Expression",
-            onKeyPress: this.addExpression,
-            defaultValue: expression.input }
+        dom.span(
```
jasonLaster: Nit, but this could be: `renderExpressionContainer`, `renderExpressionUpdating`, `renderExpression`

public/js/components/Expressions.js
```diff
@@ -3,66 +3,111 @@ const { connect } = require("react-redux");
...
-          { key: expression.id,
-            id: "expressionOutput-" + expression.id,
+          { className: "expression-output-container",
+            key: expression.id,
             onClick: this.updateExpression },
-          expression.input + " --> " +
-          JSON.stringify(expression.value || "Not Paused")
+          dom.span(
+            { className: "expression-input",
+              id: "expressionOutputId-" + expression.id },
```
jasonLaster: using ID in this way is a code smell. The id could be a data attribute `data-id=expression.value` and accessed easily

public/js/components/Expressions.js
```diff
@@ -3,66 +3,111 @@ const { connect } = require("react-redux");
...
+    const id = e.target.id.split("-").pop();
     const expression = {
-      id: e.target.id.split("-").pop(),
-      input: e.target.textContent.split(" --> ")[0]
+      id,
+      input: e.target.textContent
     };
-    this.props.updateExpression(expression);
+    updateExpression(expression);
+    setTimeout(() => {
```
jasonLaster: could the `setTimeout` go away if the focus call were in the right lifecycle method `componentDidUpdate` maybe?

public/js/actions/pause.js
```diff
@@ -135,11 +135,15 @@ function loadObjectProperties(grip) {
  */
 function addExpression(expression) {
   return ({ dispatch, getState }) => {
+    const lastExpression = getExpressions(getState()).toJS().pop();
+    const id = parseInt(expression.id, 10) ||
```
jasonLaster: I'm surprised that the id is stored as a string.

public/js/actions/pause.js
```diff
@@ -135,11 +135,15 @@ function loadObjectProperties(grip) {
...
+      (lastExpression ? lastExpression.id : -1) + 1;
```
jasonLaster: I'm a little confused by what's going on here...

jasonLaster: One approach would be to have a variable at the top of the module `let expressionId = 1` and then in here do `{ id: expressionId++ }`

jasonLaster:
Thanks! This is awesome. Added lots of small comments, but we should be able to land this soon.

jbhoosreddy:
Thanks. I was looking forward for your comments to see how I could make it
better. I'll make a follow up PR soon. Thanks!

public/js/components/Expressions.js
```diff
@@ -3,70 +3,134 @@ const { connect } = require("react-redux");
...
   },
 
   renderExpression(expression) {
+    return dom.span(
+      { className: "expression-output-container",
+        key: expression.id },
+      dom.span(
+        { className: "expression-input",
+          dataId: expression.id,
+          id: "expressionOutputId-" + expression.id,
```
jasonLaster: do you still need this?

public/js/components/Expressions.js
```diff
@@ -3,70 +3,134 @@ const { connect } = require("react-redux");
...
+    deleteExpression(expression);
+  },
+
+  renderExpressionUpdating(expression) {
+    return dom.span(
+      { className: "expression-input-container" },
+      dom.input(
+        { type: "text",
+          className: "input-expression",
+          dataId: expression.id,
```
jasonLaster: do you need the dataId anymore?

public/js/components/Expressions.js
```diff
@@ -3,70 +3,134 @@ const { connect } = require("react-redux");
...
+          id: "expressionInputId-" + expression.id,
+          onKeyPress: this.addExpression.bind(this, expression),
```
jasonLaster: I think it's more idiomatic to do `(e) => this.addExpression(e, expression)`

Also, if this is a keypress handler, the function should be `this.inputKeyPress`

public/js/components/Expressions.js
```diff
@@ -3,70 +3,134 @@ const { connect } = require("react-redux");
...
-    command: PropTypes.func
+    deleteExpression: PropTypes.func
   },
 
   displayName: "Expressions",
 
-  addExpression(e) {
+  addExpression({ id }, e) {
     if (e.key === "Enter") {
+      const { addExpression } = this.props;
```
jasonLaster: nit: it can be nicer to use a guard statement. it avoids the extra nesting and implies your looking to limit the input to the function early on.

https://en.wikipedia.org/wiki/Guard_(computer_science)

``if (e.key !== "Enter") { 
  return
}
```

File: public/js/components/Expressions.js [hunk 10]
```diff
@@ -3,70 +3,131 @@ const { connect } = require("react-redux");
...
+
+  renderExpressionValue(value) {
+    if (!value) {
+      return;
+    }
+    if (value.exception) {
+      return Rep({ object: value.exception });
+    }
+    if (typeof value.result === "boolean") {
+      return `${value.result}`;
```
jasonLaster: why not just do `return value.result` :)

public/js/components/Expressions.js
```diff
@@ -3,70 +3,131 @@ const { connect } = require("react-redux");
...
+    }
+    if (typeof value.result !== "object") {
+      return value.result;
+    }
+    if (typeof value.result === "object") {
+      return Rep({ object: value.result });
+    }
+  },
```
jasonLaster: this can be simplified to use Rep if it's an object or exception, otherwise return value.result

Issue #454, opened 2016-08-01
Surface the debugger examples


jasonLaster:
One of the most common questions I get, is how do I debug an app with source maps, or a minified source. We should answer these questions with documentation.

Here are a couple of ideas:
- add an `examples` index file that lists the different projects and what they are helpful for debugging. e.g. _increment_ is good for source maps. _todomvc_ has a minfied source (localstorage.js)
- mention the examples app when dev server starts, i.e "Examples are available at localhost:7999"
- we can always add more examples ofcourse that cover more topics.

jlongster:
These are just static files, so we could just host them somewhere like on the gh-pages of this repo. Then we wouldn't have to bother with an express server and could link directly to them from the readme or something. I'd like to keep the development server simple. (and we avoid more failure points, like starting the example server happens to clash with something else already listening on that port).

clarkbw:
> on the gh-pages of this repo

I like this.

But we should try to maintain a way to serve them locally like we do now, that has saved me on flights.  Maybe a separate examples repo would make this easier.

> * mention the examples app when dev server starts, i.e "Examples are available at localhost:7999"

With a gh-pages branch we could direct people to https://devtools-html.github.io/debugger.html/

jlongster:
Ok, I'm fine keeping them in here but it'd be nice to just have a separate command like `npm run examples` or something. If you all really want it to be automatically served that's fine too.

jasonLaster:
Hmm, the gh-pages could be nice.

I would prefer to leave it as part of the dev-server. The advantage is that it's one less step to explain to beginners and one less process to start up. The dev server crashes now in one place, which will be fixed shortly.

jlongster:
We should point most people to gh-pages though. The only reason you'd need to manually start a server is if you were offline like @clarkbw said and really want that. For the most part people can just use the hosted examples. bgrins has an examples repo that he uses on his personal github and that's worked really well.

Imagine when the debugger lands and people just want to see how to debug sourcemaps. In Firefox, you just go to our examples on gh-pages, and open devtools, and go. In that case, embedding them in this project actually makes it a lot harder to access.

jasonLaster:
@jlongster makes sense.

jasonLaster:
closing this for now. It's in the project cleanup project as a note

PR #508, opened 2016-08-08
add direcory listing for examples folder

amitzur:
currently accessing localhost:7999 results in 404, so initially I thought is was broken (I knew about it from the console of the development server)
So this adds the directory listing from which the examples are accessible


jasonLaster:
Thanks @amitzur - we were actually discussing doing something like this here #454. I imagined adding a somewhat styled `index.html`, but this is a great start before we get there.

PR #510, opened 2016-08-08
make ManagedTree use item's key as index instead of instance

jlongster:
I think the only reason we couldn't do this before is because keys weren't properly computed for the object inspector, but now they are. Keys should always be unique, so we can use that to track which ones are expanded or not. This solves the issues with the sources tree because the collapsing generates a new tree each time. Now it properly expands initially.


jasonLaster:
👍

PR #511, opened 2016-08-08
Test improvements

jasonLaster:
Adds a pretty print test
Also cleans up some of integration test timing


jlongster:
Looks good, r+

jasonLaster:
@jlongster just looked into the test failure and remembered that prettyPrint is off by default

PR #512, opened 2016-08-08
Fix refreshing while paused

jasonLaster:
When the debuggee is paused it stops sending "will-navigate" events, which means that it's possible to refresh the debuggee and still show all of the pause data in the debugger.


jlongster:
Nice! I think a `resume` event is also sent, which should be harmless but something to keep in mind if we hit other weirdness down the road. Looks good!

jasonLaster:
good to know, i didn't see that in `target`

jlongster:
A resume event is on the debugger client, not on target.

jasonLaster:
ahh - that makes sense. We should document these events perhaps :)

PR #514, opened 2016-08-08
Setup ci.json

jasonLaster:
This adds a ci.json config that lets us have separate features / configs for our ci environment.

It also removes the chrome flagging which in comparison seems like debt.


jlongster:
Sounds good to me, with our current setup I think it makes sense.

PR #515, opened 2016-08-08
update the todomvc port for npm run firefox

clarkbw:
In #453 the examples directory began running off the port `7999`, this was forgotten in the update.


jasonLaster:
Thanks!

PR #518, opened 2016-08-08
Sanitize the app state

jasonLaster:
This is a follow up from the client logging PR.


jlongster:
If you want to go ahead and merge this, that's cool. It's an improvement over the current system. I like sanitizing it more manually, i.e. convert all the ids of the sources into readable ids, because then you can easily do `fakeSource["bar.js"]` and look them up based on readable ids. But this is good for now.

jasonLaster:
> easily do `bar.js`

Yeah, I like that API too. I was imagining having a small wrapper that does a regex match. e.g. `selectSource("todo-view")`

PR #521, opened 2016-08-08
remove accordion isvg svg css selector

### LINKED ISSUE #430: breakpoint pause icon is rotated 90 degrees
The icon is paused 90 degrees. This is likely because the accordion triangle is rotated 90 degrees.

Kinda awesome bug if you ask me

screen shot 2016-07-27 at 11 24 12 am

clarkbw:
The CSS for the accordion arrow was nearly identical to the arrow CSS
except for an additional 2px of margin-right.
I ended up hunting in Svg.js and saw this odd combination of template
literals and string addition that I needed to fix.

This fixes #430


jasonLaster:
Oh great! Thanks

PR #522, opened 2016-08-08
use flex box for breakpoint component

clarkbw:
I’ve been seeing a falling float in the breakpoint list.  The pause icon keeps wrapping around and falling into the next line.  This changes the list items to use flex box instead which won’t allow for the same float issues.

screen shot 2016-08-08 at 2 38 05 pm


jasonLaster:
Awesome. Flexbox is a great idea here. I keep telling myself to try flexbox more.

Issue #389, opened 2016-07-21
dev-server httpGet crash


jasonLaster:
when `httpGet` fails to find a resource it crashes the development server. It should return an HTTP 404 not found response

This is probably easy to test with a curl

```
curl localhost:8000/get?url=foobar
```

```
SyntaxError: Unexpected token C
    at Object.parse (native)
    at /Users/jlaster/src/mozilla/debugger.html/bin/development-server.js:67:27
    at IncomingMessage.<anonymous> (/Users/jlaster/src/mozilla/debugger.html/bin/development-server.js:33:7)
    at emitNone (events.js:85:20)
    at IncomingMessage.emit (events.js:179:7)
    at endReadableNT (_stream_readable.js:913:12)
    at _combinedTickCallback (internal/process/next_tick.js:74:11)
    at process._tickCallback (internal/process/next_tick.js:98:9)

npm ERR! Darwin 15.3.0
npm ERR! argv "/usr/local/bin/node" "/usr/local/bin/npm" "start"
npm ERR! node v5.10.1
npm ERR! npm  v3.8.3
npm ERR! code ELIFECYCLE
npm ERR! debugger.html@0.0.5 start: `node bin/development-server`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the debugger.html@0.0.5 start script 'node bin/development-server'.
npm ERR! Make sure you have the latest version of node.js and npm installed.
npm ERR! If you do, this is most likely a problem with the debugger.html package,
npm ERR! not with npm itself.
npm ERR! Tell the author that this fails on your system:
npm ERR!     node bin/development-server
npm ERR! You can get information on how to open an issue for this project with:
npm ERR!     npm bugs debugger.html
npm ERR! Or if that isn't available, you can get their info via:
npm ERR!     npm owner ls debugger.html
npm ERR! There is likely additional logging output above.

npm ERR! Please include the following file with any support request:
npm ERR!     /Users/jlaster/src/mozilla/debugger.html/npm-debug.log
```

clarkbw:
I didn't solve this but took a look at possible solutions that might be useful when you're working on this @jbhoosreddy https://github.com/clarkbw/debugger.html/tree/http-dev-server-crash

jbhoosreddy:
@clarkbw thanks for this. I'll look into it!

jbhoosreddy:
@jasonLaster could you give me feedback on this?

jasonLaster:
[...]

PR #523, opened 2016-08-08
prevent http get crash (#389)

jbhoosreddy:
I forgot to make a pull request for this one. @clarkbw


jasonLaster:
Looks good to me. I don't have a great idea of express's API, but this is also internal so the big thing is that it doesn't crash

jasonLaster:
👍

Issue #134, opened 2016-05-02
Update Editor Syntax Highlighting


jasonLaster:
The current debugger's syntax highlighting is dark and depressing. We should update it to use Firefox's new colors. 

![syntax highlighting](https://cloud.githubusercontent.com/assets/254562/14970135/b3d7d12a-1094-11e6-9730-cd84c7b20c67.png)

jasonLaster:
@helenvholmes are the color / themes updated in the current debugger and ready to be applied?

helenvholmes:
Yep, the syntax highlighting in gecko-dev/fx-team should be good to go here.

clarkbw:
The original bug for those changes is here: https://bugzilla.mozilla.org/show_bug.cgi?id=1246313

Other useful resources:

* [theme variable colours](https://dxr.mozilla.org/mozilla-central/source/devtools/client/themes/variables.css)
* [light theme CSS](https://dxr.mozilla.org/mozilla-central/source/devtools/client/themes/light-theme.css)
* [search devtools theme directory for codemirror CSS selector](https://dxr.mozilla.org/mozilla-central/search?limit=100&redirect=false&q=.CodeMirror%20path%3Adevtools/client/themes%2F)

jasonLaster:
This might be addressed in part by the theming work as well #370

jlongster:
It's more likely to be solved when we figure out how to just re-use Firefox's version of codemirror. Personally I don't care if the theme doesn't match in a local tab, but if we care about that, we could solve it in #370

jasonLaster:
I started working on theming for a bit and got pretty far. Here's a [branch](https://github.com/devtools-html/debugger.html/compare/master...jasonLaster:theming?expand=1).

+ Include dark and light theme css files
+ remove overriding / legacy styles
+ update codemirror to support mozilla themes

The toolbox has a lot of logic to support loading the correct stylesheet:
+ https://dxr.mozilla.org/mozilla-central/source/devtools/client/shared/theme-switching.js
+ https://dxr.mozilla.org/mozilla-central/source/devtools/client/shared/theme-switching.js

A lot of this logic is unnecessary if we're using a CSS pre-processor like sass or less and could use the CSS nesting feature. I'd like to strongly consider that for this usecase, as if we already bundle a css file with webpack and could always use these libraries if we were to leave github.

jasonLaster:
This is resolved by #527

PR #527, opened 2016-08-09
Fix light theme syntax highlighting

jasonLaster:
+ Adds light + dark theme
+ Enables light theme                     

I discuss changing themes in https://github.com/devtools-html/debugger.html/issues/134#issuecomment-238379361. At the end of the day, i think that's a new issue that we can resolve separately. What this does is bring consistency between the old debugger and new debugger.

screen shot 2016-08-09 at 11 43 19 am

### DIFF
--- a/public/js/components/App.js
+++ b/public/js/components/App.js
@@ -6,6 +6,7 @@ const actions = require("../actions");
 
 require("./App.css");
 require("../lib/variables.css");
+require("../lib/themes/light-theme.css");
 const Sources = createFactory(require("./Sources"));
 const Editor = createFactory(require("./Editor"));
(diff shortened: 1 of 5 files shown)


webpack.config.js
```diff
@@ -8,7 +8,7 @@ const ExtractTextPlugin = require("extract-text-webpack-plugin");
 const features = require("./config/feature");
 const isDevelopment = features.isDevelopment;
 const isFirefoxPanel = features.isFirefoxPanel;
-const isEnabled = features.isEnabled;
+const getValue = features.getValue;
```
jlongster: Why are these changes in here?

jasonLaster: one minute. let me remove it

PR #528, opened 2016-08-09
refactor action unit tests and add a bunch of sources tests

jlongster:
This does a few things:

* It removes `npm run test-paths` because we can implement that with the node test runner, and I fixed that up so it works with the workers and everything
* Consolidates the small test files into files that reflect the action files. So there's just `sources.js` and `breakpoints.js`. This makes it easy to reuse things like mocked thread clients, and I don't see the benefit of splitting everything up. 
* Adds a few methods to make testing as simple as possible. No fixtures needed. Just use `makeSource` and manipulate the state (using actions) as you will.
* Adds a bunch of sources tests, particularly sourcemaps (more needed). I got the sourcemaps working under node which is nice.

I think that's about it. I think we could write a lot of tests this way and cover the majority of our app.


jlongster:
Need to fix up linting, as it's trying to lint the sourcemap test case bundle.

public/js/actions/tests/sources.js
```diff
@@ -0,0 +1,199 @@
...
+// Mocks; these must come before anything else requires them, so they
+// are above all other requires.
+require("../../utils/networkRequest").networkRequest = function(url) {
+  return new Promise((resolve, reject) => {
+    // example.com is used at a dummy URL that points to our local
+    // `/public/js` folder.
+    url = url.replace("http://example.com/", "/../../");
+    resolve(JSON.parse(fs.readFileSync(__dirname + url, "utf8")));
+  });
+}
```
jasonLaster: this should probably go in `node-unit-tests`

public/js/test/sm-test-case/webpack.config.js
```diff
@@ -0,0 +1,8 @@
+
+module.exports = {
+  entry: "./entry.js",
+  output: {
+    filename: "bundle.js"
+  },
+  devtool: "sourcemap"
+}
```
jasonLaster: I think this test case can be replaced with `test/examples/increment`

public/js/utils/source-map.js
```diff
@@ -5,9 +5,17 @@ import type { Location } from "./actions/types";
...
+let sourceMapWorker;
+function restartWorker() {
+  if(sourceMapWorker) {
+    sourceMapWorker.terminate();
+  }
+  sourceMapWorker = new Worker(
+    getValue("baseWorkerURL") + "source-map-worker.js"
+  );
+}
+restartWorker();
```
jasonLaster: minor thing - but `restartWorker` should probably be in the same util as `workerTask`

jasonLaster:
@jlongster couple of nits, but overall looks great.

public/js/actions/tests/sources.js
jlongster: Yeah, that's true. It became generic enough it probably could be. Usually test mock things themselves as needed. If we run into other things we want to mock we can think about it more, but I'll move it.

public/js/test/sm-test-case/webpack.config.js
jlongster: Hm, I want it to be a more complicated test case. Also, are those examples things that you want people to run to try out the debugger? I don't think those should be our test cases; the former are things we will want to change over time, but the latter we can't ever change. My test case is never a runnable example, but just data to be consumed by tests.

public/js/utils/source-map.js
jlongster: Yeah, I use `afterEach` to terminate it. That's a good point that we could do this automatically though. We'll need to restart all of the workers, which I assume would be ok... I'll try it out!

public/js/test/sm-test-case/webpack.config.js
jasonLaster: I'd like the debugger test cases to all be hosted so that a user could go to them. Somewhat like mochitests examples as well. 

I think there's a real benefit for contributors to be able to quickly load a test example and see how it works.

> re: never changing

I don't see the examples changing too much, but if they do then I think it is okay to update the tests that are coupled.

One thing to keep in mind is that it's important to be able to run the unit tests in the browser for devtools debugging. So, please try and make sure that the new tests can run with `npm run mocha-server`

jlongster: > I'd like the debugger test cases to all be hosted so that a user could go to them. Somewhat like mochitests examples as well.

That's the difference: these aren't integration tests like mochitests, they are unit tests. So it doesn't need runnable examples, it just needs some data. We have a bunch of really small data files like this in m-c too. I think it's worthwhile to have both because I can add any random file I need to without feeling like I need to tie it in somehow to the example's usage.

> I think there's a real benefit for contributors to be able to quickly load a test example and see how it works.

Yeah, I totally agree with more integration-y tests (as they need to be served anyway for those kinds of tests). Thinking about this more, maybe this is a better way of explaining it: what I need is just a bunch of source texts. Instead of saving fixtures, I'd like to have a folder with a bunch of sources (and some sourcemaps). It's just data that unit tests can use, and aren't grouped by various examples.

I will update this PR to reflect that, by renaming `sm-test-case` to something that is more just source texts. I could even make the thread client automatically load these as requested, so unit tests can freely load them.


> 
>     re: never changing
> 
> I don't see the examples changing too much, but if they do then I think it is okay to update the tests that are coupled.

Hopefully we'll have hundreds of unit tests so it really is a huge deal to change them. We've dealt with this on m-c and the unit tests data files are untouchable because you'd have to update almost every test. You can only add new test files, which is pretty easy anyway.

> 
> One thing to keep in mind is that it's important to be able to run the unit tests in the browser for devtools debugging. So, please try and make sure that the new tests can run with npm run mocha-server. I'm now wishing karma worked :/

Yeah, currently it won't work in the browser because it loads sourcemaps from disk. I don't find much value in karma to be honest because all this code doesn't even deal with the UI so it should run the same in all browsers, and I don't use it for debugging... To get these working though, we just need to fix up the utility functions to load stuff over the network instead of from disk. It won't be hard to get it running again when we want to focus on it. I'd rather not spend more time on it now though.

jlongster:
I renamed the folder to `unit-sources` which is where we can add any arbitrary source text. Right now anything that uses `networkRequest` will pull it from there automatically, but I also plan on making the mocked thread client load from it too so you can literally load them as sources. By creating various kinds of sources we can test all kinds of edge cases.

I tried to move `restartWorker` to `node-unit-tests`, but it's a little difficult to add a global `afterHook` that has access to it. That file is not a test file, so we'd need to define a global `__restartWorkers` function or something so the global hook can access it, but that also means that we'd be keeping a reference to all workers and restart them automatically, which might not be desired for some workers. So I added a whitelist to only restart ones we care about. But then I realized that none of this will work for karma, so if that's important we just need to manually restart each worker we care about. Might be a better solution that works for both node & karma, but I think we can figure that out later unless you have any easy solutions right now.

jlongster:
Node version on circle is a little older... had to avoid `string.includes`. Tests should pass now...

PR #529, opened 2016-08-10
File Search - UI tweaks (#338)

jbhoosreddy:
I took a stab at some of the issues mentioned here.
 -  the magnifying glass should only be blue when focused
 - file search should be open by default (with copy below mentioning cmd+p opens files)
 - the input bar should have placeholder copy for searching

### DIFF
--- a/public/images/magnifying-glass.svg
+++ b/public/images/magnifying-glass.svg
@@ -2,6 +2,23 @@
    - License, v. 2.0. If a copy of the MPL was not distributed with this
    - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
 <svg width="14px" height="14px" viewBox="488 384 14 14" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
-  <path d="M495.5,391.5 L500.200877,396.200877" id="Line" stroke="#4A90E2" stroke-width="1.25" stroke-linecap="round" stroke-linejoin="round" fill="none"></path>
-  <ellipse id="Oval-6" stroke="#4A90E2" stroke-width="1.25" fill="#FFFFFF" fill-rule="evenodd" cx="493.5" cy="389.5" rx="4.5" ry="4.5"></ellipse>
+  <defs>
+    <style>
+      use:not(:target) {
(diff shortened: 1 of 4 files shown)


jasonLaster:
The integration tests break because they assume you have to type cmd+p to open autocomplete. Infact, i'm not sure what happens now if you do that.

jasonLaster:
Looks good!

public/images/magnifying-glass.svg
```diff
@@ -2,6 +2,23 @@
...
+    <style>
+      use:not(:target) {
+        display: none;
+      }
+      #blue {
+        stroke: #4A90E2;
+      }
+      #grey {
+        stroke: #dde1e4;
+      }
```
jlongster: We're getting away from embedding CSS in SVGs. That's why we go through great lengths to embed them right in HTML using the loader stuff. Can you move this CSS into the component's CSS that you are using it in? It makes it much easier to reuse SVGs.

public/js/components/App.js
```diff
@@ -29,7 +29,7 @@ const App = React.createClass({
 
   getInitialState() {
     return {
-      searchOn: false
+      searchOn: true
```
jlongster: Not sure if I love it being on by default. It adds more UI and the need to explain it anyway. Why not just have the explanation, since we'll need it anyway? The search bar is basically a modal and it feels weird to open the debugger by default with a modal.

For these kinds of changes we should loop in Helen also.

jasonLaster: Yeah, i have mixed feelings about opening it by default. I added that to the list because it was discussed when we released autocomplete:

https://github.com/devtools-html/debugger.html/issues/338

I personally would be happy to keep it closed by default.

jlongster: It opens up UX questions: does that mean it auto-focuses on open? That means key shortcuts probably don't work, and the sources tree should probably be the thing we focus first. Beginners would probably be more confused sitting at a blank input rather than the sources tree.

jasonLaster: That makes a lot of sense.

jasonLaster:
Thanks @jbhoosreddy. I cleaned this PR up in #544

PR #530, opened 2016-08-10
Update Action logs

jasonLaster:
* Moves the log middleware to utils
* Updates console.log to log action object (doesn't fill the console)
* Updates flag to use *logging* config


public/js/utils/redux/middleware/log.js
```diff
@@ -9,7 +9,7 @@
  */
 function log({ dispatch, getState }) {
   return next => action => {
-    console.log("[DISPATCH]", JSON.stringify(action, null, 2));
+    console.log(`[DISPATCH ${action.type}]`, action);
```
jlongster: We had it like this before, but changed it to print the entire object because it got way too annoying to always expand the properties to inspect stuff. Usually you are scanning through many actions to see what happened, so we ended up having to manually expand every single one. For the type of debugging this allows, reading the action log is more important than interacting with it...

How do use action logging? Is there something about printing the object that you don't like?

jlongster: The other thing is try/CI: printing the full object always made the logs from there useful, instead of things like `{ prop: [Object] }`

jasonLaster: > CI: printing full logs
That's a good point

> what's your workflow

I like to be able to scroll the log in question. For example, if i'm looking at the source that's selected when the debugger pauses, I want to scroll to the pause event, and then go to the next select source action. This might be the 10th action and the pause action is so large that it can be tough to find. 

Ideally, this wouldn't be as much of a question because there would be a better redux action log :/

jlongster: The pause action is a problem, yeah. Maybe we can log both the action object and the stringified version of it, and cap the stringified version to 1000 chars or something. That way we would ensure that you can always scroll through actions at a certain rate, but have the ability to interactively dig into the action if you want.

jlongster: > Ideally, this wouldn't be as much of a question because there would be a better redux action log :/

You can always install the redux devtools extension if you want a better log! :)

jasonLaster: haha - i said a better log... :) The extension has lots of UX issues :P

jasonLaster: > stringified version to 1000 chars or something

I think that's a good option

PR #531, opened 2016-08-10
Additional Test Coverage

jasonLaster:
This is a work in progress. There are several things in here is like to discuss

Tests:
* Pausing in an Iframe
* stepping over an exception (viewing it in the scopes pane)

### DIFF
--- a/public/js/components/Scopes.js
+++ b/public/js/components/Scopes.js
@@ -36,10 +36,13 @@ function getSpecialVariables(pauseInfo, parentName) {
   const vars = [];
 
   if (thrown) {
+    // handle dehydrating excpetion strings and errors.
+    thrown = thrown.toJS ? thrown.toJS() : thrown;
+
     vars.push({
       name: "<exception>",
       path: parentName + "/<exception>",
-      contents: { value: thrown.toJS() }
+      contents: { value: thrown }
     });
   }
 
--- a/public/js/test/examples/debugger-statements.html
+++ b/public/js/test/examples/debugger-statements.html
@@ -1,0 +1,27 @@
+<html>
+  <head>
+    <title>Debugger Statements</title>
+  </head>
+
+  <body>
+    <script>
+      debugger;
+      test();
+
+      function test() {
+        debugger;
+        stepIntoMe();
+      }
+
+      function stepIntoMe() {
+        // step in
+        stepOverMe();
+        // step out
+      }
+
+      function stepOverMe() {
+      }
+    </script>
+  </body>
+
+</html>
--- a/public/js/test/cypress/commands/debugger.js
+++ b/public/js/test/cypress/commands/debugger.js
@@ -50,6 +50,8 @@ function debugPage(urlPart, browser = "Firefox") {
   cy.wait(1000);
   cy.navigate(urlPart)
   cy.wait(1000);
+  toggleCallStack();
+  toggleScopes();
 }
 
 function debugFirstTab(browser = "Firefox") {
--- a/public/js/test/examples/exceptions.html
+++ b/public/js/test/examples/exceptions.html
@@ -1,0 +1,38 @@
+<html>
+  <head>
+    <title>Debugger test page</title>
+  </head>
+
+  <body>
+    <script>
+
+      function uncaughtException() {
+        debugger;
+        throw "unreachable"
+      }
+
+      function caughtError() {
+        try {
+          throw new Error("error");
+        } catch (e) {
+          debugger;
+        }
+      }
+
+      function caughtException() {
+        try {
+          throw "reachable";
+        } catch (e) {
+          debugger;
+        }
+      }
+
+      caughtException();
+      caughtError();
+      caughtError();
+      uncaughtException();
+
+    </script>
+  </body>
+
+</html>
--- a/public/js/test/examples/iframe.html
+++ b/public/js/test/examples/iframe.html
@@ -1,0 +1,13 @@
+<html>
+  <head>
+    <title>Iframe</title>
+  </head>
+
+  <body>
+    <script>
+      debugger;
+    </script>
+    <iframe src="debugger-statements.html"></iframe>
+  </body>
+
+</html>
--- a/public/js/test/integration/todomvc.js
+++ b/public/js/test/integration/todomvc.js
@@ -8,8 +8,6 @@ describe("Todo MVC", function() {
 
     // pause and check the first frame
     addTodo();
-    toggleCallStack();
-    toggleScopes();
     callStackFrameAtIndex(0).contains("initialize");
 
     // select the second frame and check to see the source updated
@@ -46,6 +44,12 @@ describe("Todo MVC", function() {
     cy.reload();
   });
 
+  /**
+   * * select a minified source
+   * * pretty print source
+   * * add a breakpoint
+   * * pause in pretty source location
+   */
   it("(Firefox) Pretty Printing", function() {
     debugPage("todomvc");
     goToSource("storage");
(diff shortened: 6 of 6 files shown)


public/js/clients/firefox/events.js
```diff
@@ -59,19 +65,18 @@ function resumed(_, packet) {
...
-  if (NEW_SOURCE_IGNORED_URLS.indexOf(source.url) > -1) {
-    return;
-  }
-
-  if (source.introductionType == "debugger eval") {
-    return;
+  source.hidden = false;
+  if (isHidden(source)) {
+    source.url = `evals/console-evauation-${evalIndex++}.js`;
+    source.hidden = true;
```
jasonLaster: This is the biggest question mark. 

While testing I ran into several places where a source was expected, but wasn't  in the store. e.g. Select source. This changes adds sources to the store, but hides them from store

jlongster: Yeah, I see where you are going with this. But I think there are some problems with this. If it's a source we should ignore, that doesn't mean it's a console evaluation, so the URL is a misnomer.

There are several types of sources, and some of them we should just flat out ignore. If you're writing tests and expect one of these sources to be in the store, that's a serious problem and we should figure out why it's not in the store instead of putting it in the store. And sources in `NEW_SOURCE_IGNORED_URLS` should be ignored and never added to the store. Apparently a source with `introductionType` of `debugger eval` should be as well. That is code that the debugger frontend sends to the server to eval, and should never exist in the debugger itself.

What you're probably looking for is eval sources. These are introduced by `eval` in the debuggee and don't have a URL, unless they have been given one with the `//# sourceURL` pragma. If they don't have a URL, they should still exist in the store but the frontend should ignore them. So they are different than the above sources: the debuggee is concerned with them, but there is no good way to show non-URL sources easily.

The problem with `NEW_SOURCE_IGNORED_INTRODUCTION_TYPES` is you've conflated `debugger eval` and just `eval`. The former we should just ignore (never add to the store) and the latter we should add but the sources tree should ignore it if it doesn't have a URL. Your changes here would ignore valid eval-based sources that are given a URL with the `sourceURL` pragma, which currently should work fine.

This is how it works currently though. We ignore the right sources, and then the sources tree just ignores non-URL sources (https://github.com/devtools-html/debugger.html/blob/master/public/js/utils/sources-tree.js#L36). So dig into your tests and figure out what sources it's expected to exist, and go from there. That's where the real bug is.

jasonLaster: Thanks for the detailed answer, this helps provide some clarity.

I have a test for "break on next", where the debuggee pauses on a debugger eval. Another example would be a console evaluation that has a debugger statement in it. In both cases, the debuggee is paused and the debugger wants to show the debugger eval source, source tab, and call stack. 

> Your changes here would ignore valid eval-based sources that are given a URL with the sourceURL pragma

I have a new `evaluations` test that hopefully will catch these changes as well so we can lock down the right behavior.

jlongster: > I have a test for "break on next", where the debuggee pauses on a debugger eval...

Oh, I see. In the old debugger we do dynamically generate URLs for these kinds of things, just a dumb `SCRIPT#` name like `SCRIPT1`, with the number incrementing each time. I wouldn't say "console eval" because these scripts could be coming from any number of places. Chrome uses `VM#`.

I suppose we could add these to the store. Theoretically there's no reason a source needs to be in the store to display it: it could still load its source text and show it in a tab, and the `sources` state would still only have the sources in the tree. But it looks like our actions do go through the `sources` state to require it. This would mainly be an optimization.

I think I'd prefer to add `debugger eval code` and `debugger eval` sources to the store, assign them an arbitrary URL like `SOURCE#` or something, and not add another attribute. The source tree can only check the URL and work with sources that have one and it is not of the `SOURCE#` format.

The reason being I'd rather use the URL is the source for what "type" of source this is rather than adding more flags. Same goes for generated/original sources, you can detect an original source by just looked at the URL.

jlongster:
It'd probably be good to split out your source changes into a specific PR. It would be good to track those individually for historical sake.

jasonLaster:
> It'd probably be good to split out your source changes into a specific PR. It would be good to track those individually for historical sake.

Was just thinking about doing that as well. It's significant enough to deserve that.

public/js/clients/firefox/events.js
jlongster: Some clarification: we need to generate URLs for these types of sources anyway (eval-ed, original) so we might as well encode that information in there.

jasonLaster:
@jlongster mind taking a look?

public/js/components/Scopes.js
```diff
@@ -39,7 +39,7 @@ function getSpecialVariables(pauseInfo, parentName) {
     vars.push({
       name: "<exception>",
       path: parentName + "/<exception>",
-      contents: { value: thrown.toJS() }
+      contents: { value: thrown }
```
jlongster: thrown should be an immutable object. It works for me, so I don't know what edge case you've found. It doesn't make any sense that it wouldn't be an immut object but `returned` and `this` would be... I'll check to make sure it is working for me, but I wonder what you hit.

jlongster: This change does not work for me. I don't know what edge case you hit, but please look into it and figure out what happened.

This is what it looks like without `toJS`:

screen shot 2016-08-12 at 11 16 24 am

If I call `toJS`, it shows me all the right fields:

screen shot 2016-08-12 at 11 15 45 am

public/js/test/examples/debugger-statements.html
```diff
@@ -0,0 +1,27 @@
...
+
+      function stepIntoMe() {
+        // step in
+        stepOverMe();
+        // step out
+      }
+
+      function stepOverMe() {
+      }
+    </script>
```
jlongster: These test cases look good. At some point we should have some test cases where the script is external, but I don't know.

jlongster:
The tests look fine to me, but not sure about the exception value change.

public/js/test/examples/debugger-statements.html
jasonLaster: Yeah, i actually borrowed this approach from the mochitest examples, i'm happy doing it either way...

public/js/components/Scopes.js
jasonLaster: Hm, that's really interesting: 

here's how i got here:

```patch
diff --git a/public/js/test/integration/todomvc.js b/public/js/test/integration/todomvc.js
index a7e1f40..2e10845 100644
--- a/public/js/test/integration/todomvc.js
+++ b/public/js/test/integration/todomvc.js
@@ -102,7 +102,7 @@ describe("Todo MVC", function() {
    * * pausing on an uncaught expression
    * * reloading while paused and resuming execution
    */
-  it("(Firefox) exception", function() {
+  it.only("(Firefox) exception", function() {
     debugPage("exceptions.html");
     resume();
     scopeAtIndex(0).click();
@@ -111,6 +111,7 @@ describe("Todo MVC", function() {
     resume();
     stepOver();
     stepOver();
+    cy.pause();
     scopeAtIndex(0).click();
     scopeAtIndex(1).contains("unreachable")
```
screen shot 2016-08-12 at 11 33 14 am

jlongster: Not sure what I'm supposed to look at there? :) Can you just reproduce it locally and debug it from there?

jlongster: Oh, I just realized something: you're tests are throwing strings. Make them throw `Error` objects instead and you'll see what happens. But that is a good edge case: strings aren't converted to immutable... so we need to only convert to JS if needed.

jasonLaster: oh sorry, was giving you steps so that you could run it. I pinged you separately on irc to see what you were doing. But yea, thats definitely the issue. I'll make sure both cases are checked.

public/js/test/examples/debugger-statements.html
jlongster: Nah I think inline scripts are actually better, they are more likely to be buggy. I like it.

jasonLaster: haha - buggy, i hope you don't mean flakey. :P I hate flakey

PR #535, opened 2016-08-11
Replace karma w/ firefox driver

jasonLaster:
This PR replaces the complicated karma overhead with a simple firefox
driver approach that loads the mocha server page and reads the results.

We can always expand the output by reading the test results or setting
up a separate reporter, but for now this gives us guidance on whether
the unit tests are running successfully.

Leaving Karma:

* we've already concluded that we are not looking to test specific
   browser apis
* karma introduces a lot of overhead, which is unecessary for what we
  are trying to do.


jlongster:
Interesting, this looks cool! I'll take a closer look tomorrow.

jasonLaster:
@jlongster mind taking a look at this. 

Also, I forgot if you were able to fix the unit tests so that they could run in the browser.

jlongster:
Sorry I didn't look at this today. I'll look early tomorrow. I was wrapped up in getting codemirror & theming working in m-c and there are lots of little details to keep in my head so I stayed on that all day.

bin/firefox-driver.js
```diff
@@ -43,4 +51,21 @@ if (shouldStart) {
...
+      const match = results.match(/failures: (\d*)/);
+      const resultCode = parseInt(match[1], 10) > 0 ? 1 : 0;
+      process.exit(resultCode);
+    });
+}
+
+if (isTests) {
+  const driver = start();
+  driver.get("http://localhost:8003");
+  setTimeout(() => getResults(driver), 5000);
```
jlongster: Is there a way we can wait for something to appear in the DOM instead of a setTimeout? This is going to introduce flakiness, for sure :)

jlongster:
This looks great, nice idea. One comment, but I like trying this out.

bin/firefox-driver.js
jasonLaster: I think the canonical way to do it is to introduce a `setInterval` that checks for something and then carries out an action. So we could check the progress element for 100%, and then get the report. Honestly, I'd rather not do that work now as it's also somewhat painful to write and I'd rather bump this to 30 seconds if we need to at some point. 5 seconds is already 5x longer than it takes 😉  If it fails, it'll be because a test stalls and we need to bail anyways.

jlongster: Once we add more tests it'll surely take longer than 5 seconds to run. But this is a good enough thing for now and this is easy to improve later :) I could have sworn selenium had a way to "wait" until an element appears on the page or something, but it's something we can talk about later.

PR #536, opened 2016-08-11
add sources header back to the sources tree

### LINKED ISSUE #262: Add "Sources" label above source tree
Specifically, this one: 
screen shot 2016-06-16 at 11 41 32 am

James and I have gone back and forth a _bunch_ on this area. I liked having space above that area for [purely visual reasons](http://cl.ly/0a1n2h2N402P), but without a clear direction for functionality perhaps we should remove it.

For prosperity, we've chatted mostly about using that area for a "Filter Sources" area, although that will probably overlap with some pretty common keyboard shortcuts that people will probably get more mileage on.

@jlongster + @jasonLaster: If the two of you truly feel a filter (or, whatever else we can come up with), perhaps we should remove it—elsewise, let's use this space to chat what might become of it.

clarkbw:
fixes #262

### DIFF
--- a/public/js/components/Sources.css
+++ b/public/js/components/Sources.css
@@ -5,6 +5,15 @@
   overflow: hidden;
 }
 
+.sources-header {
+  height: 30px;
+  border-bottom: 1px solid var(--theme-gray);
+  color: var(--theme-splitter-color);
+  padding-left: 10px;
+  line-height: 30px;
+  font-size: 0.9em;
+}
+
 .sources-list {
   flex: 1;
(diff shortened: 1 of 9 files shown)


clarkbw:
## screenshots

### without editor header
screen shot 2016-08-11 at 11 29 05

### with editor header
screen shot 2016-08-11 at 11 29 25

jasonLaster:
The test failure is my fault. I accidentally pushed a change that enables the fixtures.

Thanks for the fix and going through the stale issues

PR #539, opened 2016-08-11
sources header uses the default colour

clarkbw:
This PR would change the sources header to use the darker default colour text.

Here's a comparison of the sources header with and without the grey colour.

screen shot 2016-08-11 at 2 06 13 pm
screen shot 2016-08-11 at 2 06 01 pm

@helenvholmes ?


jlongster:
I like it a lot. We may tweak the design of the whole header area but I think this makes it look a lot more consistent for now.

jasonLaster:
Plus one. Also like the spelling of colour

Issue #516, opened 2016-08-08
Add debugger keyboard shortcuts


jasonLaster:
Now that we've landed debugger command tooltips, we should go in and add some global keyboard shortcuts:

https://github.com/devtools-html/debugger.html/pull/500

shortcuts:
+ resume - F8
+ Step Over - F10
+ Step In - F11
+ Step Out - F12

PR #540, opened 2016-08-11
add keyboard shortcuts (#516)

jbhoosreddy:
this is more of a WIP PR for feedback, especially the changes I made to key-shortcuts library. I also refactored the RightSidebar component so that I could use React lifecycle methods in the component.

### DIFF
--- a/public/js/components/RightSidebar.js
+++ b/public/js/components/RightSidebar.js
@@ -1,10 +1,11 @@
 const React = require("react");
-const { DOM: dom } = React;
+const { DOM: dom, PropTypes } = React;
 const { connect } = require("react-redux");
 const { bindActionCreators } = require("redux");
 const { getPause, getIsWaitingOnBreak, getShouldPauseOnExceptions } = require("../selectors");
 const { isEnabled } = require("../feature");
 const Svg = require("./utils/Svg");
+const ImPropTypes = require("react-immutable-proptypes");
 
 const actions = require("../actions");
 const Breakpoints = React.createFactory(require("./Breakpoints"));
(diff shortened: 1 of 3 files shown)


public/js/lib/devtools-sham/client/shared/key-shortcuts.js
```diff
@@ -141,7 +141,7 @@ KeyShortcuts.parseElectronKey = function (window, str) {
     key = ElectronKeysMapping[key];
     shortcut.keyCode = window.KeyboardEvent[key];
     // Used only to stringify the shortcut
-    shortcut.keyCodeString = key;
+    shortcut.key = key;
```
jlongster: Can you explain these changes more? We need to stay in sync with m-c, and this file comes from there. It might be best for you to make these changes there first and then copy the new file over.

jasonLaster: @jlongster, tomorrow is Jaideep's last day. I'd be happy to file an m-c PR next week after this lands.

public/js/components/RightSidebar.js
```diff
@@ -40,49 +41,98 @@ function getItems() {
...
+      isWaitingOnBreak,
+      breakOnNext
+    } = this.props;
+    const { keyShortcuts } = this.props;
+    this.resume = () => pause ?
+      command({ type: "resume" }) :
+      [ isWaitingOnBreak ? null : breakOnNext ];
+    this.stepOver = () => pause ? command({ type: "stepOver" }) : null;
+    this.stepIn = () => pause ? command({ type: "stepIn" }) : null;
+    this.stepOut = () => pause ? command({ type: "stepOut" }) : null;
```
jasonLaster: Small nit, but I think these methods could move to the prototype:

```js
stepOver() {
  if (!this.props.pause) {
    return;
  }
  this.props.command({ type: "stepOver" })
}
```

jasonLaster:
small nit, but otherwise looks good.

public/js/lib/devtools-sham/client/shared/key-shortcuts.js
jbhoosreddy: From debugging the the keyboard shortcuts library, I noticed that the key code is mapped to DOM_VK_* key when adding to the list of events. But when the event occurs, it is not mapped properly, and so the listener is not triggered. Which is why I needed to make this change to allow it to properly map the "function" key events in the library.

jbhoosreddy:
@jasonLaster I worked on the nit as per your feedback.

public/js/lib/devtools-sham/client/shared/key-shortcuts.js
jlongster: Cool, so it's a bug on m-c too? Strange that we haven't hit it before... I don't see any reason Jaideep can't still contribute to m-c too after his internship :)

jasonLaster:
Looks good @jbhoosreddy.

There seems to be a bug with `componentDidMount`, that's keeping the integration tests from passing.

screen shot 2016-08-12 at 9 22 07 am

PR #541, opened 2016-08-12
Add Break On Next and eval tests

jasonLaster:
This is the follow up issue to #531.

### DIFF
--- a/public/js/clients/firefox/events.js
+++ b/public/js/clients/firefox/events.js
@@ -1,10 +1,9 @@
 const { Source, Frame, Location } = require("../../types");
 
 const CALL_STACK_PAGE_SIZE = 1000;
-const NEW_SOURCE_IGNORED_URLS = ["debugger eval code", "XStringBundle"];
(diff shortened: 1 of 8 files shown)


public/js/clients/firefox/events.js
```diff
@@ -59,12 +65,8 @@ function resumed(_, packet) {
 function newSource(_, packet) {
   const { source } = packet;
 
-  if (NEW_SOURCE_IGNORED_URLS.indexOf(source.url) > -1) {
-    return;
-  }
-
-  if (source.introductionType == "debugger eval") {
-    return;
+  if (isHidden(source)) {
```
jlongster: You need to make sure it doesn't have a URL already. This still won't work with eval sources that are given a URL with the `sourceURL` pragma.

In fact, if we are unequivocally adding sources now, this could just be:

```js
if (!source.url) { source.url = ... }
```

You only need to create a URL if it doesn't have one.

PR #542, opened 2016-08-12
Fix Pretty Print tests

jasonLaster:
I've been noticing an increase in cypress intermittents, this is not necessarily a surprise because we've added a lot more tests and increased the steps in the existing tests.

I think these fixes are a step in the right direction:

1. I noticed that some tests were attempting to pretty print a file before it had loaded. now the tests waits until the button is active, which is only the case after it has loaded. And if for any reason the action is invoked early, it bails.
2. I noticed many cases where the tests did not pause. I believe this is due to pretty print sending two client evals in quick succession. Now all evals have a mandatory 1 second wait period.

After these fixes, i did not see any errors when i ran the pretty print test 20 times.

### DIFF
--- a/public/js/actions/sources.js
+++ b/public/js/actions/sources.js
@@ -196,6 +196,11 @@ function blackbox(source, shouldBlackBox) {
 function togglePrettyPrint(id) {
   return ({ dispatch, getState, client }) => {
     const source = getSource(getState(), id).toJS();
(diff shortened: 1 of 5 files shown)


jasonLaster:
Merging to help the other tests run consistently. If there's any feedback I can incorporate it in another PR.

jlongster:
Cool!

PR #544, opened 2016-08-15
File search ui

jasonLaster:
Tweaked @jbhoosreddy's work on file search:

+ search is off by default
+ svg no longer has styles and `use`.


jlongster:
I don't see where search is turned off by default? Doesn't matter though, looks good to me :)

jasonLaster:
Oh ha - that's because it was off before. Jaideeps' PR had changed that.

PR #545, opened 2016-08-15
adds overlay message to Chrome

clarkbw:
_This isn't a priority but only affects the Chrome code._

By adding the `setOverlayMessage` method to both the pause and resume events we get the Chrome overlay in the debugger.  The overlay has working buttons, they seem to send messages directly to the debugger and our interface receives the new state and behaves correctly.

screen shot 2016-08-15 at 11 19 57 am

**NOTE** the method [setOverlayMethod](https://chromedevtools.github.io/debugger-protocol-viewer/1-1/Debugger/#method-setOverlayMessage) is part of the 1.1 stable debug protocol.  Though as you can see it is listed in the Debugger domain and yet it was [moved to the Page](https://codereview.chromium.org/801363004/) domain a while back.  Finally it appears to have been removed completely from future versions, Chrome Canary returns an error when this method is sent.  The overlay still exists in Canary but it seems access to presenting it via the protocol has been removed.

### DIFF
--- a/public/js/clients/chrome/events.js
+++ b/public/js/clients/chrome/events.js
@@ -1,9 +1,11 @@
 const { Source, Location, Frame } = require("../../types");
 
 let actions;
+let pageAgent;
 
 function setupEvents(dependencies) {
   actions = dependencies.actions;
+  pageAgent = dependencies.agents.Page;
 }
 
 // Debugger Events
@@ -44,10 +46,13 @@ async function paused(
     type: reason
   }, data);
 
+  pageAgent.setOverlayMessage("Paused in debugger.html");
+
   await actions.paused({ frame, why, frames });
 }
 
 function resumed() {
+  pageAgent.setOverlayMessage(undefined);
   actions.resumed();
 }
 
--- a/public/js/clients/chrome.js
+++ b/public/js/clients/chrome.js
@@ -66,7 +66,7 @@ function initPage(actions) {
   const agents = connection._agents;
 
   setupCommands({ agents: agents });
-  setupEvents({ actions })
+  setupEvents({ actions, agents })
 
   agents.Debugger.enable();
   agents.Debugger.setPauseOnExceptions("none");
--- a/public/js/lib/chrome-remote-debug-protocol/docs.md
+++ b/public/js/lib/chrome-remote-debug-protocol/docs.md
@@ -21,7 +21,7 @@ InspectorBackend.registerCommand("Page.enable", [], [], false);
 InspectorBackend.registerCommand("Debugger.getBacktrace", [], ["callFrames", "asyncStackTrace"], false);
 ```
 
-#### Steps to upgrade `inspectorBackend.js`:
+#### Steps to upgrade `api.js`:
 
 + add line `/* eslint-disable */`
 + add line `var WebInspector = {}, window = window || {};`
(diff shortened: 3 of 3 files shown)


jasonLaster:
Hmm, very cool. I'm not sure if this is something we want to do if it's being removed from the protocol though.

auchenberg:
API has been replaced by `Page.configureOverlay`

https://chromium.googlesource.com/chromium/src/+/02c23fc6d8a185917e1c936fa5a7b9a1c9fa309c%5E%21/#F5

auchenberg:
Chrome team is currently making major changes to the CDP API's. Next stable on Chrome will have more stable API surface and the protocol version will be bumped to Version 1.2

public/js/lib/chrome-remote-debug-protocol/bootstrap.js
```diff
@@ -42,6 +42,7 @@ InspectorBackend.registerCommand("Page.screencastFrameAck", [{"name": "sessionId
 InspectorBackend.registerCommand("Page.handleJavaScriptDialog", [{"name": "accept", "type": "boolean", "optional": false}, {"name": "promptText", "type": "string", "optional": true}], [], false);
 InspectorBackend.registerCommand("Page.setColorPickerEnabled", [{"name": "enabled", "type": "boolean", "optional": false}], [], false);
 InspectorBackend.registerCommand("Page.setOverlayMessage", [{"name": "message", "type": "string", "optional": true}], [], false);
+InspectorBackend.registerCommand("Page.configureOverlay", [{ "name": "suspended", "type": "boolean", "optional": true },{"name": "message", "type": "string", "optional": true}], [], false);
```
clarkbw: Just added this in by hand to get it working with Canary without removing the function above it.

public/js/clients/chrome/events.js
```diff
@@ -46,12 +46,14 @@ async function paused(
     type: reason
   }, data);
 
+  pageAgent.configureOverlay(false, "Paused in debugger.html");
```
clarkbw: I'm just calling both methods right now.

This needs some kind of protocol / feature detection.  If the protocol level was detecting the version and / or features when generating the methods our client code could be a bit simpler.  From the client we could do `if ("configureOverlay" in pageAgent)` to decide between the the two versions.  Will need a deeper dive.

public/js/lib/chrome-remote-debug-protocol/bootstrap.js
jasonLaster: Sounds good, mind adding it to the `README` in the same directory

public/js/clients/chrome/events.js
jasonLaster: If calling both works, then I think this is fine for now :) 

I don't want to over think it for now

jasonLaster:
Thanks @auchenberg. Good to know

clarkbw:
Dropping the `pageAgent.configureOverlay(false, "Paused in debugger.html");` for now.  Our [protocol.json](./public/js/lib/chrome-remote-debug-protocol/protocol.json) file is set at version 1.1 and adding the `configureOverlay` method pulls in a method from the tip (ToT) that muddies the waters.  This is just a small part of the mess of dealing with multiple protocol versions.

If we bump protocol versions in the future we'll need to update the overlay message, otherwise I think we should stick with the original commit with updates to the docs or not add this at all.

jasonLaster:
Looks good. 

Thanks!

PR #546, opened 2016-08-15
(WIP) Update Pause on exceptions

jasonLaster:
This became a fairly large PR because I decided to take on a couple of refactorings:
+ Refactor Right Sidebar to a class
+ clean up individual render functions
+ move pause selectors to the reducer

And also support "disable all breakpoints". I naively thought that this would be a small feature, turned out to have some complexity:
+ async - the debugger server did not handle parallel breakpoint commands well. Sequencing them with an await call helped.
+ loading - it was very easy to thrash the button. It turned out that, it was important to disable the button while breakpoints were being updated (loading).


jasonLaster:
still working on disable breakpoints

public/js/actions/breakpoints.js
```diff
@@ -114,6 +115,25 @@ function removeAllBreakpoints() {
...
 }
 
+function toggleAllBreakpoints(shouldDisableBreakpoints) {
+  return ({ dispatch, getState }: ThunkArgs) => {
+    const breakpoints = getBreakpoints(getState());
+    return dispatch({
+      type: constants.TOGGLE_BREAKPOINTS,
+      shouldDisableBreakpoints,
+      [PROMISE]: (async function () {
+        for (let [, breakpoint] of breakpoints) {
```
jasonLaster: do you think this is the best way to destructure the variable?
I chose to use `for of` over `forEach as it seemed simpler with the async

public/js/actions/pause.js
```diff
@@ -83,6 +89,22 @@ function command({ type }) {
...
+  return command({ type: "stepOver" });
+}
+
+function stepOut() {
+  return command({ type: "stepOut" });
+}
+
+function resume() {
+  return command({ type: "resume" });
+}
```
jasonLaster: these actions used to be called through `command`, but calling them directly makes the code more readable/understandable.

public/js/reducers/breakpoints.js
```diff
@@ -146,11 +156,26 @@ function getBreakpointsForSource(state: OuterState, sourceId: string) {
...
 
+function getBreakpointsDisabled(state: OuterState) {
+  return state.breakpoints.get("breakpointsDisabled");
+}
+
+function getBreakpointsLoading(state: OuterState) {
+  const breakpoints = getBreakpoints(state);
+  const isLoading = !!breakpoints.valueSeq()
+                    .filter(bp => bp.loading)
+                    .first();
```
jasonLaster: checking the first value is a simple way to see if there *is* a value in the list

public/js/reducers/pause.js
```diff
@@ -98,4 +102,47 @@ function update(state = initialState, action, emit) {
...
+  return state.pause.get("shouldIgnoreCaughtExceptions");
+}
+
+function getFrames(state: AppState) {
+  return state.pause.get("frames") || [];
+}
+
+function getSelectedFrame(state: AppState) {
+  return state.pause.get("selectedFrame");
+}
```
jasonLaster: moving selectors into the reducer...

jasonLaster:
Update, I believe the code is in a good place. I plan on adding some integration tests for 'toggle all breakpoints' and 'pause on exceptions' before this lands. Feel free to review now though.

public/js/actions/breakpoints.js
```diff
@@ -84,7 +84,8 @@ function _removeOrDisableBreakpoint(location, isDisabled) {
     if (bp.loading) {
       // TODO(jwl): make this wait until the breakpoint is saved if it
       // is still loading
-      throw new Error("attempt to remove unsaved breakpoint");
+      // throw new Error("attempt to remove unsaved breakpoint");
+      return;
```
jlongster: I'd still like to error here. This is a bug and we should suppress the error. Note that the current debugger acts this way too; we didn't implement proper add/remove ordering. The error doesn't break the UI; it just stops this redux dispatch cycle, but the state is still in tact and the UI keeps working.

I'd rather properly fix this. This is what the `waitUntil` middleware is for; because we don't have a good abstraction for these kinds of things, it allows you to wait on other actions in a primitive way. Here, if it's still loading, it should dispatch a `waitUntil` request can check when the bp isn't loading anymore, and then call `removeBreakpoint` again. I can do that, so remove this change and I'll implement a proper fix.

public/js/actions/breakpoints.js
jlongster: Yeah, I think the destructuring is fine. However, you could send off the requests in parallel:

`[PROMISE]: Promise.all(breakpoints.map(bp => dispatch(...)))`

public/js/actions/breakpoints.js
```diff
@@ -155,5 +175,6 @@ module.exports = {
   disableBreakpoint,
   removeBreakpoint,
   removeAllBreakpoints,
```
jlongster: Do we still need `removeAllBreakpoints`? I don't see it being used anywhere.

public/js/actions/pause.js
```diff
@@ -58,12 +58,18 @@ function paused(pauseInfo) {
   };
 }
 
-function pauseOnExceptions(toggle) {
+function pauseOnExceptions(
+  shouldPauseOnExceptions, shouldIgnoreCaughtExceptions) {
```
jlongster: nit: usually in devtools code we put the last paren on a new line:

```
function foo(
  // lots of args...
) { 
}
```

or

```
function pauseOnExceptions(shouldPauseOnExceptions,
                           shouldIgnoreCaughtExceptions) {
  // ...
}
```

public/js/actions/pause.js
jlongster: Agreed!

public/js/components/RightSidebar.js
```diff
@@ -22,73 +26,148 @@ function debugBtn(onClick, type, className, tooltip) {
...
+          style: { overflowX: "hidden" }},
+        dom.div(
+          { className: "command-bar" },
+          this.renderStepButtons(),
+          this.renderDisableBreakpoints(),
+          this.renderPauseOnExceptions()
+        ),
+
+        Accordion({
+          items: this.getItems()
```
jlongster: Personally I'd prefer to keep the items inline in here. I like breaking up complicated things but it's more useful to read what is going on with the accordion by just putting them here. Though I see it was like this before, so I guess I missed that change :) It's subjective.

public/js/reducers/breakpoints.js
jlongster: You can just check `size` or I think immutable does have an `empty` predicate on collections

public/js/actions/breakpoints.js
jasonLaster: oops, didn't mean to leave this in

jlongster:
This looks great! A few nits, but you can merge after fixing tests

public/js/components/RightSidebar.js
jasonLaster: Yea, I agree. If anything, I'd prefer if it were `renderAccordion` and then was grouped there.

PR #549, opened 2016-08-16
Breakpoints loading

### LINKED ISSUE #525: Breakpoints - adding lots of breakpoints causes errors
If you open a source and start thrashing on the gutter (adding/removing) lots of breakpoints, you'll quickly see three errors:

+ ` {from: "server1.conn147.child1/34", error: "wrongState", message: "Cannot set breakpoint while debuggee is running."}`
+ `error: "wrongState", message: "Stack frames are only available while the debuggee is paused.", from: "server1.conn147.child1/26"}`
+ `TypeError: Cannot read property 'type' of undefined` in firefox/events.js.16

screen shot 2016-08-09 at 10 51 19 am

jasonLaster:
Fixes #525.

This fixes a problem we were seeing where breakpoints would be updated while loading, which caused an exception.

### DIFF
--- a/public/js/components/Breakpoints.js
+++ b/public/js/components/Breakpoints.js
@@ -37,6 +37,10 @@ const Breakpoints = React.createClass({
   displayName: "Breakpoints",
 
   handleCheckbox(breakpoint) {
+    if (breakpoint.loading) {
+      return;
+    }
+
     if (breakpoint.disabled) {
       this.props.enableBreakpoint(breakpoint.location);
     } else {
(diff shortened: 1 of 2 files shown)


jlongster:
I'm fine with, but I wanted to show you an example of the `waitUntil` usage which would also fix it. I'll try to do that tomorrow morning, otherwise we can merge this.

jasonLaster:
I'm going to merge this so that we improve the test quality. I've created https://github.com/devtools-html/debugger.html/issues/555 to discuss `waitUntil` and other approaches.

PR #550, opened 2016-08-16
Hide pretty-print button from mapped files

### LINKED ISSUE #420: Pretty Print - Pretty printing a generated source
Case:
+ a bundle file "a.js" has been source mapped, creating original files "b.js", "c.js"
+ user pretty prints "a.js" creating "d.js"

Use Case
* the user wants to see the bundle file prettified so that he can see what crazy babel transpiling is going on and debug that directly.

Problem:
+ do we support breakpoints in "d.js" (pretty bundle) and "b.js" and "c.js"? This might be difficult because two source maps point back to the same generated source. To be fair, this might just work, but we should test it.

jasonLaster:
Fixes #420

### DIFF
--- a/public/js/components/SourceFooter.js
+++ b/public/js/components/SourceFooter.js
@@ -8,6 +8,7 @@ const { getSelectedSource, getSourceText } = require("../selectors");
 const Svg = require("./utils/Svg");
 const ImPropTypes = require("react-immutable-proptypes");
 const classnames = require("classnames");
+const { isMapped } = require("../utils/source-map");
 
 function debugBtn(onClick, type, className = "active", tooltip) {
   className = `${type} ${className}`;
@@ -42,6 +43,10 @@ const SourceFooter = React.createClass({
   prettyPrintButton() {
     const { selectedSource, sourceText, togglePrettyPrint } = this.props;
     const sourceLoaded = selectedSource && !sourceText.get("loading");
+
+    if (isMapped(selectedSource.toJS())) {
+      return;
+    }
 
     return debugBtn(
       () => togglePrettyPrint(selectedSource.get("id")),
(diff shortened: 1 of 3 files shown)


public/js/utils/source.js
```diff
@@ -29,6 +29,11 @@ function isJavaScript(url, contentType = "") {
          contentType.includes("javascript");
 }
 
+function isGenerated(source) {
```
jlongster: This check seems wrong. A "generated" source is considered any source that is the real source in the engine. This seems more like `hasSourcemap`.

public/js/components/SourceFooter.js
```diff
@@ -43,6 +44,10 @@ const SourceFooter = React.createClass({
     const { selectedSource, sourceText, togglePrettyPrint } = this.props;
     const sourceLoaded = selectedSource && !sourceText.get("loading");
 
+    if (isGenerated(selectedSource.toJS())) {
```
jlongster: Don't we want to hide it for sources that have sourcemaps *and* original sources?

jasonLaster: That's a good point. I change this to a selector, `hasOriginalSources`.

jasonLaster:
@jlongster I updated the PR and added a couple other things while in there:

+ changes the original source id to slash delineated
+ changes definition of isGenerated to not original
+ changes getGeneratedSourceId to use the source ID as opposed to
searching the map. This makes it possible to move isGenerated and
isOriginal out of the worker.

jlongster:
Looks great. Any reason you didn't go ahead and move them out of the worker? It's not a big deal, either way this looks good, r+

jasonLaster:
@jlongster no big reason. the work already seemed like it had gotten large and it took awhile to test for correctness.

PR #553, opened 2016-08-16
Be more compatible with Firefox: full light theme & codemirror

jlongster:
Lots of little things in here. I'm going to keep working on this, but wanted to get this PR up. It mainly does two things:

1. Conforms to the editor API expected in the toolbox so we can just swap it out and use the local version when in the Firefox panel
2. Pulls in the full light theme and uses it

For #2, for some reason the theme doesn't explicitly say it, but the default font size for the devtools panel is 11px. I think for now we just need to force ourselves to use it also, even if things look small. The reason is the ensure that we make everything pixel-perfect with that font-size. Soon we can introduce a local-specific CSS files that overrides some of these things, which can bump the font size and other things. But I think we need to be developing the debugger as the firefox panel sees it right now.


config/feature.js
```diff
@@ -15,6 +15,10 @@ function isEnabled(key) {
 }
 
 function isDevelopment() {
+  if(isFirefoxPanel()) {
+    // Default to production if compiling for the Firefox panel
+    return process.env.NODE_ENV === "development";
+  }
```
jlongster: The panel before was still pulling in the development config, so it was bundling in hot module reloading and such.

public/js/lib/themes/light-theme.css
```diff
@@ -252,3 +257,94 @@ div.CodeMirror span.eval-text {
   line-height: 1.4em;
   min-height: 1.4em;
 }
+
```
jasonLaster: Could you comment out or include the png urls?

https://dxr.mozilla.org/mozilla-central/source/devtools/client/themes/light-theme.css#267

This way we'll either have them or be able to add them later

public/js/components/Editor.js
```diff
@@ -88,29 +92,29 @@ const Editor = React.createClass({
...
-    this.editor.addLineClass(line - 1, "line", "debug-line");
-    alignLine(this.editor, line);
+    this.editor.codeMirror.addLineClass(line - 1, "line", "debug-line");
+    this.editor.alignLine(line);
   },
 
   setSourceText(newSourceText, oldSourceText) {
     if (newSourceText.get("loading")) {
-      this.editor.setValue("Loading...");
+      this.setText("Loading...");
```
jasonLaster: nice cleanup

public/js/components/RightSidebar.css
```diff
@@ -13,8 +13,8 @@
 .command-bar > span {
   cursor: pointer;
   margin-right: 0.7em;
-  width: 1em;
-  height: 1.1em;
+  width: 16px;
```
jasonLaster: does it make sense to make this `1.2em`?

webpack.config.devtools.js
```diff
@@ -8,17 +9,28 @@ const ignoreRegexes = [
...
   webpackConfig.externals = [
     function(context, request, callback) {
+      const mod = path.join(context.replace(__dirname + "/", ""), request);
+
       // Any matching paths here won't be included in the bundle.
       if(ignoreRegexes.some(r => r.test(request))) {
         callback(null, "var {}");
         return;
+      } else if(nativeMapping[mod]) {
+        callback(null, "var devtoolsRequire('" + nativeMapping[mod] + "')");
```
jasonLaster: what does this do? It seems weird that there's `var devtoolsRequire`

public/js/components/Editor.js
```diff
@@ -1,30 +1,38 @@
...
+ * numbers gutter. Editor CSS will absolutely position the gutter
+ * beneath the line numbers. This makes it easy to be flexible with
+ * how we overlay breakpoints.
+ */
+function resizeBreakpointGutter(editor) {
+  const gutters = editor.display.gutters;
+  const lineNumbers = gutters.querySelector(".CodeMirror-linenumbers");
+  const breakpoints = gutters.querySelector(".breakpoints");
+  breakpoints.style.width = lineNumbers.clientWidth + "px";
+}
```
jasonLaster: If you think there will be other editor utils, it might be nice to bring back `utils/editor.js`

jasonLaster:
The tests are failing because `cm` is no longer exposed on the window: https://github.com/devtools-html/debugger.html/blob/master/public/js/test/cypress/commands/debugger.js#L75-L76

overall: couple of nits, but looks good. 

It'll be interesting to revisit after `source-editor/editor.js` is de-privileged.

config/feature.js
jasonLaster: Does it make sense to always do? `process.env.NODE_ENV === "development"`?

jlongster: Not sure I understand your question?

This is yet another case of the difference of build and run-time. In the firefox config we default NODE_ENV to be "production", but that doesn't happen at build-time. So we need to default to production at build-time too, but allow the user to override it.

public/js/lib/themes/light-theme.css
jlongster: We don't use pngs/svgs from m-c. We could patch all the URLs and copy the images over but I don't think it's worth it, as we are rewriting the basic components and using inline SVGs for everything.

But you're right in that I should probably just comment out the URLs instead of removing them, as that's better documentation for how we patched the file.

public/js/components/RightSidebar.css
jlongster: I don't think so; they are icons and should not be resized with the font. The command bar has a fixed height of 30px so the icons should fit inside of it exactly. I don't see the benefits of ems here.

webpack.config.devtools.js
jlongster: This is the magic that hooks into the local modules when in the panel. `devtoolsRequire` is the privileged require that will be available in the panel context. It will generate code like `var editor = devtoolsRequire("devtools/editor")`.

public/js/components/Editor.js
jlongster: If it gets big enough, sure, but if it's only a few small utility functions I think inlining them is fine. But we'll see what this looks like after we cleanup all the sourceeditor/editor.js stuff.

jlongster:
> The tests are failing because cm is no longer exposed on the window: https://github.com/devtools-html/debugger.html/blob/master/public/js/test/cypress/commands/debugger.js#L75-L76

Ah, gotcha, I will add that back in, sorry. I wasn't sure if it was leftover from debugging but that is useful for tests.

public/js/components/RightSidebar.css
jasonLaster: ahh. makes sense

public/js/components/Editor.js
jasonLaster: yeah, agreed

public/js/lib/themes/light-theme.css
jasonLaster: good point

jlongster:
This is starting to come together. We are now using all the real CSS files from m-c with little changes.

PR #556, opened 2016-08-17
Add keyboard shortcuts for the debugger

jasonLaster:
This is a follow up PR from #540.

I'll file an `m-c` issue soon to address the change

### DIFF
--- a/public/js/components/RightSidebar.js
+++ b/public/js/components/RightSidebar.js
@@ -43,31 +43,71 @@ const RightSidebar = React.createClass({
     breakpoints: ImPropTypes.map,
     isWaitingOnBreak: PropTypes.bool,
     breakpointsDisabled: PropTypes.bool,
-    breakpointsLoading: PropTypes.bool
+    breakpointsLoading: PropTypes.bool,
+    keyShortcuts: PropTypes.object
   },
 
   displayName: "RightSidebar",
 
-  renderStepButtons() {
-    const { pause, resume, stepOver, stepIn,
-            stepOut, isWaitingOnBreak, breakOnNext } = this.props;
-
-    if (pause) {
-      return [
-        debugBtn(resume, "resume", "active", "Click to resume (F8)"),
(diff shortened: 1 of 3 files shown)


public/js/components/RightSidebar.js
```diff
@@ -41,33 +41,65 @@ const RightSidebar = React.createClass({
...
+
+  componentWillUnmount() {
+    const { keyShortcuts } = this.props;
+    keyShortcuts.off("F8", this.resume);
+    keyShortcuts.off("F10", this.stepOver);
+    keyShortcuts.off("F11", this.stepIn);
+    keyShortcuts.off("F12", this.stepOut);
+  },
 
+  renderStepButtons() {
```
jasonLaster: it's hard to see here, but adding the individual step functions cleaned up this function quite a bit.

jbhoosreddy:
@jasonLaster, thanks for taking care of this.

jbhoosreddy:
Also, in regards to the change in m-c, I find it a bit strange that there is a bug in m-c library. Because the existing keyboard shortcuts work in the current debugger. I'm curious to see if you run into a similar problem.

jlongster:
> Also, in regards to the change in m-c, I find it a bit strange that there is a bug in m-c library. Because the existing keyboard shortcuts work in the current debugger. I'm curious to see if you run into a similar problem.

That's exactly what I'm curious about. We should investigate this before making changes to m-c and understand why we need to make those changes.

jasonLaster:
It looks like the toolbox is using a different mechanism for registering the key shortcuts:

screen shot 2016-08-17 at 5 34 56 pm


https://dxr.mozilla.org/mozilla-central/source/devtools/client/debugger/views/toolbar-view.js#55-79

https://dxr.mozilla.org/mozilla-central/source/devtools/client/debugger/debugger.xul#183-216

jlongster:
Makes sense, thanks!

PR #557, opened 2016-08-17
Move makeOriginalSource to source-map-utils

jasonLaster:
This fixes a small inconsistency with `makeOriginalSource`. It also introduces a shared utility for source maps, which we can now use because the worker is being bundled with webpack.

Please ignore the irony that the source map worker is itself source mapped :P

### DIFF
--- a/public/js/utils/source-map-utils.js
+++ b/public/js/utils/source-map-utils.js
@@ -1,0 +1,13 @@
+
+function makeOriginalSource({ url, source, id = 1 }) {
+  const generatedSourceId = source.id;
+  return {
+    url,
+    id: `${generatedSourceId}/originalSource${id}`,
+    isPrettyPrinted: false
+  };
+}
+
+module.exports = {
+  makeOriginalSource
+};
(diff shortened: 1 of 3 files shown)


jasonLaster:
Merging it as it's a simple fix. Let me know if you have any comments and i can publish a new PR or revert.

jlongster:
The source-map worker doesn't need `makeOriginalSource`, right? Are there other utilities you plan on sharing? It gets a little confusing having 3 sourcemap files.

I think we can reduce the scope of the sourcemap worker to only download and parse the sourcemap. Everything else should be able to be done on the main thread, but I haven't looked into this yet. This would simplify this by only needing the worker for a single API. This PR is fine btu I may look into that soon.

jasonLaster:
`makeOriginalSource` is used in the worker to create original sources.

I'd like to see if we can reduce the worker, the big constraint here is that they have the consumer and node are in the worker, so all of the queries against them need to be in the worker.

jlongster:
Yeah, I believe we can make the consumer transferrable though, which would allow the simplification. I might be missing something though.

jasonLaster:
Got it, yeah that would be awesome. I'm not too familiar with what that would take

PR #558, opened 2016-08-18
Pretty Print UX

### LINKED ISSUE #421: Pretty Print  UX
Found a couple of tweaks that will need to be made after the PR lands:

- [ ] formatted source should not be in the tree
- [ ] pretty print button should be "selected" in pretty source
- [ ] selecting the pretty print button should take you to the other source ugly -> pretty, pretty -> ugly

jasonLaster:
Fixes #421


jlongster:
Nice, looks like all good changes, r+

PR #559, opened 2016-08-18
Rename todomvc to tests

jasonLaster:
The integration tests now test far more than todomvc and should reflect that.

this is the smallest pr you will see all day


jlongster:
Cool, r+

I believe that test failure is leftover from the sourcemap actor ID not being updated, which you fixed in a recent PR.

jasonLaster:
Yup - that's why i merged the PR pre-emptively.

PR #560, opened 2016-08-18
use unique breakpoint class to not clash with mozilla-central

jlongster:
Using just "breakpoint" clashes with m-c. It pulls in CSS that applies a style to `.breakpoint` so it messes with our breakpoint (and actually adds the old breakpoint on top of ours). The reason I didn't see this before was that we aren't loading the SVGs so I didn't see the SVG that m-c CSS was adding.

Tweaking our class with avoid all these issues. It's a little unfortunate, but an easy change that will avoid any clashing for now.

I initially tried to move the breakpoint-specific CSS from CodeMirror to the old debugger, but that doesn't work because it pulls it in in an iframe, so you can't style it from the outside. This will have to do for now.

### DIFF
--- a/public/js/components/Editor.css
+++ b/public/js/components/Editor.css
@@ -22,7 +22,7 @@
   left: 0;
 }
 
-.editor.breakpoint svg {
+.editor.new-breakpoint svg {
   fill: var(--theme-selection-background);
   width: 60px;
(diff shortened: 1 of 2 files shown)


jasonLaster:
r+

PR #562, opened 2016-08-19
Add mochitest runner

jlongster:
After all of my research for taskcluster, this is all we need right now. This will add ~10 min to our CI, but that's ok, because we really have to get tests running for the debugger actually in the Firefox panel.

In the near future, we can work on integration tests that run in content normally. We should talk about how we can write tests that could potentially be run in both worlds, because we always need the integration tests to be running on try for every commit, and we need them to be running with the debugger in the panel. The only way to test it as part of browser chrome is to use mochitests.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,6 +25,7 @@
     "firefox": "node bin/firefox-driver --start",
     "cypress": "./node_modules/.bin/cypress run --port 2021",
     "cypress-intermittents": "for i in {1..100}; do time npm run cypress; done | tee cypress-run.log",
+    "mochitest": "./bin/run-mochitests",
     "prepublish": "webpack"
   },
   "dependencies": {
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -25,6 +25,7 @@
     "firefox": "node bin/firefox-driver --start",
     "cypress": "./node_modules/.bin/cypress run --port 2021",
     "cypress-intermittents": "for i in {1..100}; do time npm run cypress; done | tee cypress-run.log",
+    "mochitest": "./bin/run-mochitests"
```
clarkbw: missing trailing comma

jlongster: I wish I had seen this before I spent a few minutes figured that out, thanks!

jasonLaster:
👍

jlongster:
Having problems building the artifact build for some reason on CircleCI, will continue to look into it...

jasonLaster:
closing this as we're not going down this path.

PR #566, opened 2016-08-22
Fix some of the flakiness in the integration tests

jasonLaster:
1. Add a wait before the debuggee command.
The theory is that maybe the breakpoint is not set or something like
that that's keeping the debuggee from pausing

2. Clear localstorage
I have no idea how many todos are saved, it could be in the thousands by
now. This clears the list so that it doesn't create performance issues.

### DIFF
--- a/public/js/test/cypress/commands.js
+++ b/public/js/test/cypress/commands.js
@@ -32,6 +32,7 @@ Cypress.addParentCommand("debuggee", function(callback) {
 
   const script = getFunctionBody(callback);
 
+  cy.wait(1000);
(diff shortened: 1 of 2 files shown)


jlongster:
Are you sure this actually helps? You didn't wait for this CI to run before merging. Have you seen improvements?

jasonLaster:
It's an improvement, but i don't think it's a complete fix. I've been running the the tests on circle for the past hour or so.

It's very hard to track down the source of the issue since its now reliable locally and just an issue on CI.

PR #567, opened 2016-08-22
Fix Break on next

jasonLaster:
I noticed in the test runs that #558 broke the "break on next" ux. It also caused the stepping buttons to always be active, which might have contributed to some of the flakiness. Also, the #556 caused an issue with updating the right sidebar on subsequent changes.

Unfortunately, I believe the test flakiness has caused me to question the test results and not re-run when I see a failure. I'm going to continue investigate some the flakiness and also make sure that I confirm that a test is truly intermittent before merging.

### DIFF
--- a/public/js/components/RightSidebar.js
+++ b/public/js/components/RightSidebar.js
@@ -80,6 +80,11 @@ const RightSidebar = React.createClass({
 
   setupKeyboardShortcuts() {
     const { keyShortcuts } = this.props;
+    if (this.keyShortcutsEnabled) {
+      return;
+    }
+
+    this.keyShortcutsEnabled = true;
     keyShortcuts.on("F8", this.resume);
     keyShortcuts.on("F10", this.stepOver);
     keyShortcuts.on("F11", this.stepIn);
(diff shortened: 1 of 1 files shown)


public/js/components/RightSidebar.js
```diff
@@ -98,19 +103,29 @@ const RightSidebar = React.createClass({
...
-  shouldComponentUpdate(nextProps) {
-    return this.props.keyShortcuts !== nextProps.keyShortcuts;
-  },
-
   renderStepButtons() {
+    const className = this.props.pause ? "active" : "disabled";
     return [
-      debugBtn(this.resume, "resume", "active", "Click to resume (F8)"),
-      debugBtn(this.stepOver, "stepOver", "active", "Step Over (F10)"),
-      debugBtn(this.stepIn, "stepIn", "active", "Step In (F11)"),
```
jasonLaster: As a general UX note, i think it would be nice if this were a toggle. The benefit would be that the user could disable this option once it is set.

jasonLaster:
This is likely a debugger server issue. I noticed that setting "break on next" and then adding and removing a breakpoint causes break on next to be disabled.

public/js/components/RightSidebar.js
```diff
@@ -80,6 +80,11 @@ const RightSidebar = React.createClass({
 
   setupKeyboardShortcuts() {
     const { keyShortcuts } = this.props;
+    if (this.keyShortcutsEnabled) {
+      return;
+    }
+
+    this.keyShortcutsEnabled = true;
```
jlongster: I don't think you'd need this if you used `componentDidMount`, instead of `componentDidUpdate`. Typically you set up this kind of stuff when it mounts and then remove it when it unmounts.

public/js/components/RightSidebar.js
jlongster: What would be toggled?

jasonLaster: the wait on next expression button

jlongster:
Generally looks fine to me. If it's a server issue, it happens the exact same way as the previous debugger so I wouldn't worry about that for now.

public/js/components/RightSidebar.js
jasonLaster: The issue here is a frustrating lifecycle issue with react. The child event is first and so keyShortcuts is not setup yet.

jasonLaster:
^ @jlongster yeah, i'm fine with that. just wanted to bring it up so it was noted

public/js/components/RightSidebar.js
jlongster: Sorry, still confused... we don't have "wait on next expression"? We only have line-based stepping..

public/js/components/RightSidebar.js
jlongster: Why not just check `if(this.keyShortcuts) { ... }` ?

You're passing down a prop that is created in the parent's `componentDidMount`, so it makes sense to be null when it's rendering first.

There's probably a better way to do this, most likely using context to expose shortcuts across the whole app, or even just using a top-level object. Not sure.

jasonLaster: I think context will likely be the better solution for `keyShortcuts`. checking the existence of keyShortcuts would allow many handlers to be bound.

I'm going to merge this as this does not seem like a blocker, but an improvement. Happy to update in a separate issue.

public/js/components/RightSidebar.js
jasonLaster: Hmm... if you click the pause/resume button while the debugger is not paused, it will tell the server to pause on the next expression.

PR #571, opened 2016-08-22
throttle sources tree instead of debounce

jlongster:
We shouldn't be debouncing. Previously the action was throttled, which means you are guaranteed to render about every 50ms, instead of lots of sources continually debouncing and not rendering anything for several seconds. This changes when we moved this logic into the component.

I noticed this when trying to load a site with lots of sources: the sources tree is just blank for 5-10 seconds while it's loading. Not it continually renders the tree as they come in.

### DIFF
--- a/public/js/components/SourcesTree.js
+++ b/public/js/components/SourcesTree.js
@@ -1,18 +1,16 @@
 const React = require("react");
 const { DOM: dom, PropTypes } = React;
+const classnames = require("classnames");
+const ImPropTypes = require("react-immutable-proptypes");
+const { Set } = require("immutable");
 
 const {
   nodeHasChildren, createParentMap, addToTree,
   collapseTree, createTree
 } = require("../utils/sources-tree.js");
-
-const classnames = require("classnames");
-const ImPropTypes = require("react-immutable-proptypes");
-const { Set } = require("immutable");
-const debounce = require("lodash/debounce");
-
 const ManagedTree = React.createFactory(require("./utils/ManagedTree"));
 const Svg = require("./utils/Svg");
+const { throttle } = require("../utils/utils");
 
 let SourcesTree = React.createClass({
   propTypes: {
@@ -26,12 +24,16 @@ let SourcesTree = React.createClass({
     return createTree(this.props.sources);
   },
 
-  componentWillMount() {
-    this.debouncedUpdate = debounce(this.debouncedUpdate, 50);
-  },
+  queueUpdate: throttle(function() {
+    if (!this.isMounted()) {
+      return;
+    }
+
+    this.forceUpdate();
+  }, 50),
 
   shouldComponentUpdate() {
-    this.debouncedUpdate();
+    this.queueUpdate();
     return false;
   },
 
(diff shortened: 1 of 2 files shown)


public/js/components/SourcesTree.js
```diff
@@ -27,11 +25,17 @@ let SourcesTree = React.createClass({
...
   componentWillMount() {
-    this.debouncedUpdate = debounce(this.debouncedUpdate, 50);
+    this.queueUpdate = throttle(() => {
+      if (!this.isMounted()) {
+        return;
+      }
+
+      this.forceUpdate();
+    }, 50);
   },
```
jasonLaster: small nit, but I would prefer this syntax for decorated function

```js
queueUpdate: throttle(function() {
}),
```

public/js/components/SourcesTree.js
```diff
@@ -1,18 +1,16 @@
...
 } = require("../utils/sources-tree.js");
-
-const classnames = require("classnames");
-const ImPropTypes = require("react-immutable-proptypes");
-const { Set } = require("immutable");
-const debounce = require("lodash/debounce");
-
 const ManagedTree = React.createFactory(require("./utils/ManagedTree"));
 const Svg = require("./utils/Svg");
+const { throttle } = require("../utils/utils");
```
jasonLaster: Lodash uses `throttle` slightly differently https://lodash.com/docs#throttle

Do you think we could call this `delay`, `queue`, `defer`, or `batch`

public/js/utils/utils.js
```diff
@@ -172,6 +172,18 @@ function updateObj<T>(obj: T, fields: $Shape<T>) : T {
   return Object.assign({}, obj, fields);
 }
 
+function throttle(func: any, ms: number) {
+  let timeout;
+  return (...args: any) => {
+    if (!timeout) {
```
jasonLaster: small nit:

```js
if (timeout) {
  return
}
```

jasonLaster:
Nice fix, this must come up big time w/ node

public/js/components/SourcesTree.js
jlongster: `Creates a throttled function that only invokes func at most once per every wait milliseconds` That's exactly what this does, if it doesn't then it's a bug and I'll fix it :) It's the same as throttling with async channels, etc, which is what I'm familiar with.

public/js/utils/utils.js
jlongster: I don't prefer early returns for small functions, but I know that's a stylistic choice. It's easier to see the full logic instead of having to see what returns at the top, but doing it for large/complex functions is fine... It's common in style guides to avoid early returning, but we should solidify these style rules at some point.

jlongster:
> Nice fix, this must come up big time w/ node

Haha, yep... I'm using a complex site as an example and it stalled for about 10 seconds.

public/js/components/SourcesTree.js
jlongster: Yeah, I can do that

public/js/utils/utils.js
jasonLaster: oh interesting. I can see that point of view.

public/js/components/SourcesTree.js
jasonLaster: Sorry, my definition of throttle was wrong. I was thinking that debounce and throttle were similar, except throttle would be called immediately and then would *debounce* push back the execution.

Here's a good [article](https://css-tricks.com/debouncing-throttling-explained-examples/) that was linked.

jlongster:
Looks like a common cypress failure

jlongster:
Updated from comments (converted that method), so merging

Issue #548, opened 2016-08-16
Breakpoints - disable state


jasonLaster:
It would be nice if gutter markers were visible for disabled breakpoints. It is disconcerting to see the breakpoint in the list, but not in the editor.

jasonLaster:
@johnkpaul this might be a nice bug

PR #572, opened 2016-08-22
Add disabled state for breakpoint markers

MattCain:
Hi,

I came across #548 through up-for-grabs.net, thought i'd have a look at it 😃

This will show disabled breakpoint gutter markers but faded out so you can tell the difference.

screen shot 2016-08-22 at 19 11 53

It looks as though the Editor isn't ready for unit tests yet? (Correct me if i'm wrong!) So i didn't write any for this but existing unit tests pass.

Please let me know if there's any problems or i've missed anything.

Thanks,
Matt

### DIFF
--- a/public/js/components/Editor.css
+++ b/public/js/components/Editor.css
@@ -31,6 +31,10 @@
   right: -4px;
 }
 
+.editor.new-breakpoint.breakpoint-disabled svg {
+  opacity: 0.3;
+}
+
 .editor-mount,
 .CodeMirror {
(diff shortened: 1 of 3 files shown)


jasonLaster:
Thanks @MattCain! 

I don't think I could have done this any better myself. Thanks for improving the formatting as you went along. Looks great to me.

Please feel free to grab another issue if you're interested. If you have any questions, feel free to ask, I'd love to provide more clarity to some of the issues in our queue.

PR #573, opened 2016-08-22
always render connect message immediately & tweak styles

jlongster:
When I want to connect to node, I open the debugger and click on the "connect to node" link.

But the "not connected" message doesn't show until the Firefox connection times out. We should just go ahead and show the "not connected" message by default, and when it finds tabs it will re-render and show the normal tabs page. The `Tabs` component is already listening to the tab state so all we need to do is render the `Tabs` component immediately.

I also updated the styles because the styles for the "not connected" component don't really apply on the main tabs page.

### DIFF
--- a/public/js/components/Tabs.css
+++ b/public/js/components/Tabs.css
@@ -45,16 +45,15 @@
   color: var(--theme-highlight-bluegrey);
 }
 
-.not-connected-message {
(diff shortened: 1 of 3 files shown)


jasonLaster:
r+

👍

PR #574, opened 2016-08-22
Fix #532 - pretty printing on paused locations

### LINKED ISSUE #532: Pretty Print - pretty printing a paused source
When you pretty print a source that is currently paused in, the green debug line should also be visible in the pretty source.

jasonLaster:
This fixes a problem where frames are only translated to original locations when they're received. Now, frame data is translated when a source is pretty printed as well.

What Changed:
* The logic for updating a frame, which looks up the original location was moved to a utility function
* The `togglePrettyPrint` action gets updated frames and passes them along, which the pause reducer consumes and updates with.
* The `Pause.selectedFrame` field was updated to be `Pause.selectedFrameId`. Now that I think about it, this is similar to `Sources.selectedSource`. I changed `selectedFrame` so that, the pretty print action wouldn't have to also translate the selected frame and pass that through. I could avoid this in the reducer by plucking the translated selected frame and setting it as well. With that said, I prefer using ID fields because it avoids keeping duplicate data in sync.


public/js/test/examples/evals.html
```diff
@@ -12,13 +12,16 @@
...
+        eval("bar = function() {\nvar x = 6; debugger;\n}//# sourceURL=bar.js\n bar();");
       }
 
       function evalSourceWithDebugger() {
         eval("bar = function() {\nvar x = 7;\ndebugger; }\n bar();");
       }
     </script>
+
+    <button onclick="evalSourceWithDebugger()"> evalSourceWithDebugger </button>
+    <button onclick="evalSourceWithSourceURL()"> evalSourceWithSourceURL </button>
```
jasonLaster: this helped me test these cases manually

public/js/utils/source-map.js
```diff
@@ -89,17 +89,22 @@ async function getOriginalLocation(state: AppState, location: Location) {
...
       location
     );
 
     const { url, line } = originalPosition;
+    if (!url) {
+      return {
+        sourceId: source.get("id"),
+        line: location.line
+      };
+    }
```
jasonLaster: This is an interesting case. The frame locations have undefiend columns,
which is not sufficient for finding the original location in some cases like `  bar()`

```
bar = function() {
var x = 7;
debugger; }
 bar();
```

public/js/utils/sources-tree.js
```diff
@@ -5,8 +5,8 @@ const { isPretty } = require("../utils/source");
 const IGNORED_URLS = ["debugger eval code", "XStringBundle"];
 
 function isHiddenSource(source) {
-  const url = getURL(source);
-  return !url || url.path.match(/SOURCE/) || IGNORED_URLS.includes(url);
+  const url = source.get("url");
+  return !url || url.match(/SOURCE/) || IGNORED_URLS.includes(url);
```
jasonLaster: using `getURL` broke down with `SOURCE:formatted`. `getURL` thought source was a host because of the colon. The easy fix was to not use getURL because we don't need paths for this

jlongster: I didn't notice this before, but shouldn't we be matching on ID instead of URL anyway? We switched the id format so we could easily match `/originalSource#` I thought. That way we can change the URL to whatever we want in the future (because it determines how it's display in the UI) and none of our detection code breaks.

jasonLaster: hmm, i can look into this, but I think the issue here is that we want to show *most* original sources.

public/js/utils/source-map.js
jlongster: Nice

public/js/utils/sources-tree.js
jlongster: Oops, you're right, I didn't mean sourcemapped sources. This is for eval sources. But we can't change the ID of eval'ed sources, so this makes sense. Ignore my above comment.

jlongster:
Sorry for taking a while to look through this. All of this looks like good changes, so you can merge it if tests are OK!

PR #577, opened 2016-08-24
move network timeout code to connect() Fixes #377

### LINKED ISSUE #377: Handle failed network connection to debug [chrome]
With connections to Firefox there is some timeout logic that gracefully handles the case where the connection fails to find debuggable tabs.   In that case we then display the following message.  I'll be updating this message in a future PR.

screen shot 2016-07-20 at 3 34 32 pm

However when you turn on debug for Chrome the connection fails and errors out leaving a blank page.  We need to fix this so our default page can contain some additional instructions for both Firefox and Chrome.

Here are the relevant places where the network connection is handled for each client.  It would probably be best to have a shared solution but I might just implement something that works for now.

## Firefox

https://github.com/jlongster/debugger.html/blob/8f84c9b7341bd0bfe734368066d1bc9b86a5eb46/public/js/clients/firefox.js#L61-L67

## Chrome

https://github.com/jlongster/debugger.html/blob/d6715edc31cb3bf2e5d9078124aebc9edae4b256/public/js/clients/chrome.js#L49

clarkbw:
Catch errors and reject the promise in chrome client
Remove the timeout function from the firefox client and move it to the `shared/client/main DebuggerClient.connect` function
Use a similar Promise.race() style for handling the timeout in both `networkRequest` and `DebuggerClient.connect`
In the DebuggerClient I used the `new Promise()` style of promises but the rest of the code uses `deferred`.  I wanted to keep a similar style to the rest of the code but it just feels awkward here.  I can rework it to use deferred if that’s needed.  I don't really understand the sham directory.

### DIFF
--- a/public/js/clients/chrome.js
+++ b/public/js/clients/chrome.js
@@ -45,6 +45,9 @@ function connectClient() {
   const url = `http://localhost:${webSocketPort}/json/list`;
   networkRequest(url).then(body => {
     deferred.resolve(createTabs(body))
+  }).catch(err => {
+    console.log(err);
+    deferred.reject();
   });
 
   return deferred.promise;
--- a/public/js/clients/firefox.js
+++ b/public/js/clients/firefox.js
@@ -47,7 +47,6 @@ function createTabs(tabs) {
 
 function connectClient() {
   const deferred = defer();
-  let isConnected = false;
   const useProxy = !getValue("firefox.webSocketConnection");
   const portPref = useProxy ? "firefox.proxyPort" : "firefox.webSocketPort";
   const webSocketPort = getValue(portPref);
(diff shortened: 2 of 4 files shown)


jbhoosreddy:
I didn't realize something like promise.race was in the spec. Wish I knew
it earlier when I was working in a different problem.

Also, I remember reading somewhere that we want to change implementation
from "defer" to "new Promise" style as defer did not make it to the spec.

jasonLaster:
Same, `Promise.race` is nice. I like this change. 

The changes to `client/main` are good, but i think we'll want to make that change to m-c as well. I'd happy to merge this and then if the `m-c` patch changes at all create a new PR here with those changes. Up to you though.

clarkbw:
> I didn't realize something like promise.race was in the spec

Yeah, its kind of neat.  I've never actually found a good use for `Promise.race` and saw this kind of failure case somewhere else and it finally made sense to me.  The common case cited in examples is to use it for competing, but similar requests like asking Google Maps and Open Street Maps for the same thing but taking the answer from whoever returns first but that always seemed like a lot of extra work for very little gain.

clarkbw:
> The changes to client/main are good, but i think we'll want to make that change to m-c as well. I'd happy to merge this and then if the m-c patch changes at all create a new PR here with those changes. Up to you though.

I'll take you up on that offer

> Looks like you'll need to run cypress first too. it's pretty borked.

I can't seem to run the tests, locally I keep getting this error `TypeError: Cannot read property 'compileSync' of undefined` which I think you've seen before: https://github.com/cypress-io/cypress/issues/185

jasonLaster:
> compile sync

Try removing the screenshots directory in public/js/tests/cypress

Also, that's just a warning so the tests should pass otherwise. Try running the tests from the app, it's much clearer: https://docs.cypress.io/docs/installing-and-running#section-running-tests-from-the-gui

PR #578, opened 2016-08-24
Persist selected source across debuggee refreshes

### LINKED ISSUE #526: Refreshing Debuggee should not clear the selected source
When the debuggee is refreshed, the debugger should show the selected source when it the source is available. 

This is how chrome behaves and it's offputting to see the source tabs, but not selected source text.

jasonLaster:
Fixes #526.

There's a similar problem with tabs, which this does not address.

### DIFF
--- a/public/js/clients/firefox.js
+++ b/public/js/clients/firefox.js
@@ -93,11 +93,6 @@ function initPage(actions) {
 
   tabTarget.on("will-navigate", actions.willNavigate);
   tabTarget.on("navigate", actions.navigate);
-  tabTarget.on("frame-update", function(_, packet) {
-    if (packet.destroyAll) {
-      actions.willNavigate();
-    }
-  });
 
   // Listen to all the requested events.
   setupEvents({ threadClient, actions });
--- a/public/js/reducers/sources.js
+++ b/public/js/reducers/sources.js
@@ -100,10 +100,12 @@ function update(state = State(), action: Action) : Record<SourcesState> {
       }
 
       return _updateText(state, action, [action.originalSource]);
+
     case "NAVIGATE":
-      // Reset the entire state to just the initial state, a blank state
-      // if you will.
-      return State();
+      const source = state.selectedSource;
+      const sourceUrl = source && source.get("url");
+      return State()
+        .set("pendingSelectedSourceURL", sourceUrl);
   }
 
   return state;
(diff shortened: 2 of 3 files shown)


public/js/clients/firefox.js
```diff
@@ -102,7 +102,6 @@ function initPage(actions) {
 
   setupCommands({ threadClient, tabTarget, debuggerClient });
 
-  tabTarget.on("will-navigate", actions.willNavigate);
```
jasonLaster: while working on this patch i noticed that `will-navigate` fires twice for debuggee refreshes. This is because `frame-update` fires for a will navigate event when the page is paused. I removed this event because i believe it is redundant now.

public/js/reducers/sources.js
```diff
@@ -100,10 +100,12 @@ function update(state = State(), action: Action) : Record<SourcesState> {
...
       return _updateText(state, action, [action.originalSource]);
+
     case "NAVIGATE":
-      // Reset the entire state to just the initial state, a blank state
-      // if you will.
-      return State();
+      const source = state.selectedSource;
+      const sourceUrl = source && source.get("url");
+      return State()
+        .set("pendingSelectedSourceURL", sourceUrl);
```
jasonLaster: moves the selected source to pending, which will trigger a re-select

public/js/clients/firefox.js
jlongster: Does it only fire twice when you are reloading when it's paused" `frame-update` should not fire if you are reloading a page when it's not paused.

public/js/reducers/sources.js
jlongster: Nit: `.set("pendingSelectedSourceURL", state.getIn(["selectedSource", "url"]))` I think that should work.

jlongster:
Cool, it's great that we can reuse that pending property.

public/js/clients/firefox.js
jasonLaster: hmm - strange, i saw it firing when the page was not paused

jlongster: I looked into `frameUpdate` and we definitely don't want to use that: https://github.com/mozilla/gecko-dev/blob/master/devtools/server/actors/webbrowser.js#L785

It's fired when any frame in the page changes. We should just use the navigation events -- why are we listening to `frameUpdate` at all? I can't remember, but I think it had something to do with refreshing while paused. Above it sounds like you are seeing both events fire when paused, so we should be able to just use the navigation events.

jasonLaster:
@jlongster I updated this pr to drop frame-update and use will-navigate which is the proper behavior. 

This does mean that we don't handle the pause state at the moment but we're tracking that work and should have a fix for that soon

PR #581, opened 2016-08-25
Fix scopes with selected frame

### LINKED ISSUE #579: Scopes pane doesnt update
When selecting different frames the scopes pane does not update with the proper variables. I was able to fix this by changing the pause reducer action type SECLECT_FRAME

```
    case constants.SELECT_FRAME:
      let intermState = state.setIn(["pause", "frame", "scope"], fromJS(action.frame.scope));
      return intermState.set("selectedFrame", action.frame);
```

Not sure this is the best way to handle this though.

jasonLaster:
Fixes #579 

@JasonWeathersby the change ended up being larger than I expected and certainly larger than your one line change! The big difference is that here the scopes pane uses the selectedFrame as a data source as opposed to updating the pause info directly.

### DIFF
--- a/public/js/components/Scopes.js
+++ b/public/js/components/Scopes.js
@@ -3,9 +3,10 @@ const { bindActionCreators } = require("redux");
 const { connect } = require("react-redux");
 const ImPropTypes = require("react-immutable-proptypes");
 const actions = require("../actions");
-const { getPause, getLoadedObjects } = require("../selectors");
+const { getSelectedFrame, getLoadedObjects, getPause } = require("../selectors");
 const ObjectInspector = React.createFactory(require("./ObjectInspector"));
 const { DOM: dom, PropTypes } = React;
+const toPairs = require("lodash/toPairs");
 
 require("./Scopes.css");
 
@@ -13,20 +14,27 @@ function info(text) {
   return dom.div({ className: "pane-info" }, text);
 }
 
+function excludeVariable(binding) {
+  if (!binding[1]) {
+    return;
+  }
+  const value = binding[1].value;
+  return value.missingArguments || value.optimizedOut;
+}
+
 // Create the tree nodes representing all the variables and arguments
 // for the bindings from a scope.
 function getBindingVariables(bindings, parentName) {
-  return bindings
-    .get("arguments").map(arg => arg.entrySeq().get(0))
-    .concat(bindings.get("variables").entrySeq())
-    .filter(binding => (!binding[1].hasIn(["value", "missingArguments"]) &&
-                        !binding[1].hasIn(["value", "optimizedOut"])))
+  const args = [].concat(...bindings.arguments.map(toPairs));
+  const variables = toPairs(bindings.variables);
+
+  return args.concat(variables)
+  .filter(binding => !excludeVariable(binding))
     .map(binding => ({
       name: binding[0],
       path: parentName + "/" + binding[0],
-      contents: binding[1].toJS()
-    }))
-    .toArray();
+      contents: binding[1]
+    }));
 }
 
 function getSpecialVariables(pauseInfo, parentName) {
@@ -65,12 +73,12 @@ function getSpecialVariables(pauseInfo, parentName) {
   return vars;
 }
 
-function getScopes(pauseInfo) {
+function getScopes(pauseInfo, selectedFrame) {
   if (!pauseInfo) {
     return null;
   }
 
-  let scope = pauseInfo.getIn(["frame", "scope"]);
+  let scope = selectedFrame.scope;
   if (!scope) {
     return null;
   }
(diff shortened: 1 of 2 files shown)


public/js/components/Scopes.js
```diff
@@ -3,30 +3,38 @@ const { bindActionCreators } = require("redux");
...
+  return args.concat(variables)
+  .filter(binding => !excludeVariable(binding))
     .map(binding => ({
       name: binding[0],
       path: parentName + "/" + binding[0],
-      contents: binding[1].toJS()
-    }))
-    .toArray();
+      contents: binding[1]
+    }));
```
jasonLaster: immutable's api is certainly more verbose

public/js/components/Scopes.js
```diff
@@ -3,30 +3,38 @@ const { bindActionCreators } = require("redux");
...
+
 // Create the tree nodes representing all the variables and arguments
 // for the bindings from a scope.
 function getBindingVariables(bindings, parentName) {
-  return bindings
-    .get("arguments").map(arg => arg.entrySeq().get(0))
-    .concat(bindings.get("variables").entrySeq())
-    .filter(binding => (!binding[1].hasIn(["value", "missingArguments"]) &&
-                        !binding[1].hasIn(["value", "optimizedOut"])))
+  const args = [].concat(...bindings.arguments.map(toPairs));
```
jasonLaster: I plan on switching this to just a flatten call (probably lodash)

public/js/components/Scopes.js
```diff
@@ -96,18 +104,18 @@ function getScopes(pauseInfo) {
         vars = vars.concat(getSpecialVariables(pauseInfo, title));
       }
 
-      if (vars.length) {
+      if (vars && vars.length) {
```
jasonLaster: i'll drop this check and make sure vars is always an array

public/js/components/Scopes.js
```diff
@@ -116,18 +124,26 @@ const Scopes = React.createClass({
...
+    loadObjectProperties: PropTypes.func,
+    selectedFrame: PropTypes.map
   },
 
   displayName: "Scopes",
 
   getInitialState() {
-    return { scopes: getScopes(this.props.pauseInfo) };
+    const { pauseInfo, selectedFrame } = this.props;
+    return { scopes: getScopes(pauseInfo, selectedFrame) };
```
jasonLaster: what was the reason we used state here? could we have changed Scopes to use `componentShouldUpdate`

JasonWeathersby:
Jason this is more verbose but more correct :) . My initial go was similar in that I added the selectedFrame to the scopes component and modified the getScopes function.

jasonLaster:
Neat. Thanks for finding it again.

What were you doing with the debugger at the time?

JasonWeathersby:
I was documenting the pause reducer and I noticed that the scopes variables where coming from pause state instead of selected frame.

jasonLaster:
haha - love it.

jasonLaster:
Interesting finding by the way, I did most of the development for this feature with this cypress test. 

```js
  it.only("(Firefox) Test Scopes", function() {
    debugPage("todomvc");
    goToSource("todo-view");
    toggleBreakpoint(33);

    // pause and check the first frame
    addTodo();
    stepIn();
  });
```

Having it meant that I could make a change and then quickly be back in the right place to debug it. Honestly, I probably could do the same thing with storybook and a fixture. I'm not sure if HMR would work with the way scope state is stored, but that could be another technique

public/js/components/Scopes.js
jlongster: Shouldn't need that; the way the data is shaped you can just do `.map(x => toPairs(x)[0]` like before. It's stupid, but it's because the object only ever going to have one key and one value.

public/js/components/Scopes.js
jlongster: I'd rather keep the logic of `excludeVariable` in here, it feels like over-abstracting. How come you check of `binding[1]` is truthy in the function?  You should just able tp filter out `binding[1].value.missingArguments || binding[1].value.optimizedOut;` which is pretty short.

jlongster:
This is great, I like that the scopes is a normal JS object as it is a bit easier to dig around in. I want to avoid converting the pause object into an immutable obj at some point too. Thanks for fixing this!

jasonLaster:
Just fixed a related issue where some scopes had the same sibling key because they shared the same function display name or more likely were "(anonymous)".

jlongster:
I don't think `Scopes` is hot-reloadable as it stands because it's a connected component. Definitely should look into how to enable it for more components.

public/js/components/Scopes.js
jlongster: It's stateful because we shouldn't keep transient objects like a tree structure in the redux state. It just is dynamically generated from the state.

jlongster:
Did you accidentally merge this? Not sure you saw my review :)

jasonLaster:
Nope. Missed it. Will fix that now. Thanks for checking!

public/js/components/Scopes.js
jasonLaster: that's a good point

PR #585, opened 2016-08-25
File Search on All platforms

### LINKED ISSUE #533: Cmd+P keyboard shortcut for Windows
Currently, the keyboard shortcut to toggle file search is (Cmd+P) which works perfectly on OSX. But I suspect, it will not work Windows as it does not have the Cmd key.
We need to add additional listener for the corresponding Windows key shortcut (Ctrl+P) to toggle file search.
We should also modify the welcome screen copy to reflect that change for Windows users.
screen shot 2016-08-10 at 11 41 40 pm
thoughts @jasonLaster?

wldcordeiro:
Fixes #533 making it so the shortcut works on non-OSX devices and shows the appropriate message for the platform.

### DIFF
--- a/public/js/components/App.js
+++ b/public/js/components/App.js
@@ -2,6 +2,7 @@ const React = require("react");
 const { DOM: dom, PropTypes, createFactory } = React;
 const { connect } = require("react-redux");
 const { bindActionCreators } = require("redux");
+const { Services } = require("Services");
 const actions = require("../actions");
 
 require("./App.css");
@@ -94,9 +95,10 @@ const App = React.createClass({
   },
 
   renderWelcomeBox() {
+    const modifierTxt = Services.appinfo.OS === "Darwin" ? "Cmd" : "Ctrl";
     return dom.div(
       { className: "welcomebox" },
-      "Want to find a file? (Cmd + P)"
+      `Want to find a file? (${modifierTxt} + P)`
     );
   },
 
--- a/public/js/components/App.js
+++ b/public/js/components/App.js
@@ -53,12 +53,12 @@ const App = React.createClass({
 
   componentDidMount() {
     this.shortcuts = new KeyShortcuts({ window });
-    this.shortcuts.on("Cmd+P", this.toggleSourcesSearch);
+    this.shortcuts.on("CmdOrCtrl+P", this.toggleSourcesSearch);
     window.addEventListener("keydown", this.onKeyDown);
   },
 
   componentWillUnmount() {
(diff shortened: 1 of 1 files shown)


jlongster:
Great job! This looks good to me, although I'm not sure what our shimmed Services does locally. I bet it just hardcodes Mac as the platform. But we shouldn't deal with that here.

jlongster:
Not sure about the consistent cypress errors. I've never seen those ones and almost every test is failing, but it can't be because of this PR. Any ideas @jasonLaster?

wldcordeiro:
@jlongster yeah, I asked about the best way to determine the platform and @jasonLaster suggested using the `Services` lib.

wldcordeiro:
@jlongster Looking at CI we get this error

`CypressError: Timed out retrying: Expected to find element: '.autocomplete input', but never found it.`

jbhoosreddy:
from my understanding from the [webpack.config](https://github.com/devtools-html/debugger.html/blob/master/webpack.config.js#L43-L46), Services library is taken from devtools instead of devtools sham, so it should work. [Services](https://github.com/devtools-html/debugger.html/blob/master/public/js/lib/devtools/client/shared/shim/Services.js#L484-L497)

 But yes, devtools-sham simply assumes it to be OSX.

jasonLaster:
Looks like a good change. I wouldn't be surprised if this change did break the cypress tests as they use file search to navigate to a source.

The best thing to do here is to run [cypress](https://github.com/devtools-html/debugger.html/blob/master/docs/integration-tests.md#installing-cypress) locally. Happy to look into it as well.

wldcordeiro:
So my phone is dead, and Cypress wants me to log into Github which needs 2FA so I can try later in that case. 😐

jasonLaster:
Actually, you're changes run perfectly locally. I'm going to rebuild circle and see if the test failures are just on CI

jasonLaster:
You should be good. I bet we'll merge soon :) 

Thanks for the help

jasonLaster:
You know, the second test failed too.

What i think is happening is that we're firing cmd on a linux box


[cmd](https://github.com/wldcordeiro/debugger.html/blob/file-search-all-platforms/public/js/test/cypress/commands/debugger.js#L65-L66). I'm going to play with this on CI, by building w/ ssh [here](https://circleci.com/gh/devtools-html/debugger.html/679) and will report back. Feel free to do the same

jasonLaster:
Closing this so that I can merge a fix for the platform event.

PR #586, opened 2016-08-25
File search all platforms

jasonLaster:
Follow up from #585 with a platform fix as well for search. Borrows from the Services check as well


jasonLaster:
Merging as my fix was just to test code

Issue #588, opened 2016-08-25
Suggestion: Add Editorconfig file


wldcordeiro:
Something that would be a nice developer convenience would be a simple [EditorConfig](http://editorconfig.org/) file. With how many different standards there are to spacing and stuff, it's nice to be able to leave your editor defaults as is and let the project dictate.

I could PR one but I figured I'd make an issue first.

jasonLaster:
Go for it. Would be nice to see, I think it's often helpful. Also can't hurt

clarkbw:
This was done a while back

PR #589, opened 2016-08-26
Add basic EditorConfig file

wldcordeiro:
Associated with #588 this adds a very basic editorconfig. Are there any file types we use that don't have the same standards?

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,11 @@
+# http://editorconfig.org
+
+root = true
+
+[*]
+charset = utf-8
+indent_style = space
+indent_size = 2
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true


jasonLaster:
Looks good to me.

jlongster:
What dos this do when we already have ESLint rules?

wldcordeiro:
@jlongster ESLint is a great tool for the style of your code. EditorConfig lets your project dictate the basic file config stuff and will enforce it on an editor level (usually with a plugin) but the short of this is that you can have your tab sizes set to 8, your line endings to `CRLF` and this will automatically adjust the values of those settings for anything you edit in the project.

jlongster:
Sure, we can try it out. I don't think anyone on our team uses it though, so we can't guarantee it's going to stay up-to-date or all projects from us is going to have it.

PR #592, opened 2016-08-26
Show "this" in the scopes pane

jasonLaster:
It turned out that we were no longer showing the <this> property in the scopes pane. This happened because `this` wasn't being added to frames.

### DIFF
--- a/public/js/test/integration/tests.js
+++ b/public/js/test/integration/tests.js
@@ -10,6 +10,10 @@ describe("Tests", function() {
     addTodo();
     callStackFrameAtIndex(0).contains("initialize");
 
+    scopeAtIndex(0).click();
+    scopes().contains("<this>");
+
+
     // select the second frame and check to see the source updated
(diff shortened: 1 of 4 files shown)


public/js/test/integration/tests.js
```diff
@@ -134,24 +134,24 @@ describe("Tests", function() {
...
     scopeAtIndex(0).click();
-    scopeAtIndex(1).contains("Error")
+    scopes().contains("Error")
 
     resume();
     stepOver();
     stepOver();
     scopeAtIndex(0).click();
-    scopeAtIndex(1).contains("unreachable")
+    scopes().contains("unreachable")
```
jasonLaster: we also need to update the tests to now search any scope for the error because we can't assume it's the first scope property.

jasonLaster:
Merging as this is straight forward.

@jlongster happy to adjust if you have any concerns.

PR #595, opened 2016-08-26
Toggle breakpoint should wait for breakpoint to be active

jasonLaster:
This helps address the cypress intermittent problem where we see that the call stack frames are empty when we expect to be paused.

I believe the problem is largely due to adding a breakpoint and not waiting for the breakpoint to be active. This can take a little longer on CI because the app pauses and resumes as well.


jlongster:
Looks good, as does the other various PRs you merged. Thanks!

Issue #594, opened 2016-08-26
Scopes - replace Object formatting with Object text


jasonLaster:
We currently format object properties in the scope pane, but it would be nice to just show the text "Object". This is an improvement because formatting the object in the scope pane clutters the pane and does not provide a lot of value when the user can easily expand it on click.

The improvement will likely go [here](https://github.com/devtools-html/debugger.html/blob/master/public/js/components/ObjectInspector.js#L132-L137). 

#### Before
![screen shot 2016-08-26 at 11 07 32 am](https://cloud.githubusercontent.com/assets/254562/18010114/2bd987e8-6b7d-11e6-8f85-4984e90948f7.png)

#### After
![screen shot 2016-08-26 at 11 08 11 am](https://cloud.githubusercontent.com/assets/254562/18010113/2bd9316c-6b7d-11e6-961e-744bd139ffdb.png)

wldcordeiro:
✋

jasonLaster:
go for it

wldcordeiro:
@jasonLaster I noticed that this is the line that does the rendering, but I'm wondering how to inspect the `objectValue` because I can't log it in a way that is useful.

https://github.com/devtools-html/debugger.html/blob/master/public/js/components/ObjectInspector.js#L154

It seems like a pretty straightforward change but I thought I'd need make sure I don't break the view for other item types.

jlongster:
Reps already supports this with a "short" mode, so I would investigate that before doing anything else. I don't have the info off-hand to help right now, but we shouldn't need to this manually.

linclark:
I think you actually want `tiny` mode for this. You can see an example in the [tests](https://dxr.mozilla.org/mozilla-central/source/devtools/client/shared/components/test/mochitest/test_reps_object.html#83).

wldcordeiro:
Should be closeable.

PR #597, opened 2016-08-26
Replace Object formatting with text

wldcordeiro:
Issue #594 - Not 100% sure I exposed the prop right so some review is nice.

### DIFF
--- a/public/js/components/ObjectInspector.js
+++ b/public/js/components/ObjectInspector.js
@@ -131,7 +131,7 @@ const ObjectInspector = React.createClass({
     let objectValue;
     if (nodeHasProperties(item) || nodeIsPrimitive(item)) {
       const object = item.contents.value;
-      objectValue = Rep({ object });
+      objectValue = Rep({ object, mode: "tiny" });
     }
 
     return dom.div(
--- a/public/js/components/Rep.js
+++ b/public/js/components/Rep.js
@@ -5,8 +5,8 @@ const Grip = require("../lib/devtools/client/shared/components/reps/grip").Grip;
 
 require("../lib/devtools/client/shared/components/reps/reps.css");
 
-function renderRep({ object }) {
-  return Rep({ object, defaultRep: Grip });
+function renderRep({ object, mode }) {
+  return Rep({ object, defaultRep: Grip, mode });
 }
 
 module.exports = renderRep;
--- a/public/js/lib/devtools/client/shared/components/reps/rep.js
+++ b/public/js/lib/devtools/client/shared/components/reps/rep.js
@@ -74,6 +74,7 @@ define(function (require, exports, module) {
     propTypes: {
       object: React.PropTypes.any,
       defaultRep: React.PropTypes.object,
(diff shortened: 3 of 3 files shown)


public/js/components/ObjectInspector.js
```diff
@@ -131,7 +131,7 @@ const ObjectInspector = React.createClass({
     let objectValue;
     if (nodeHasProperties(item) || nodeIsPrimitive(item)) {
       const object = item.contents.value;
-      objectValue = Rep({ object });
+      objectValue = Rep({ object, mode: "tiny" });
```
wldcordeiro: Not sure I always want tiny, maybe I do? Clarify @jasonLaster

public/js/components/Rep.js
```diff
@@ -5,8 +5,8 @@ const Grip = require("../lib/devtools/client/shared/components/reps/grip").Grip;
 
 require("../lib/devtools/client/shared/components/reps/reps.css");
 
-function renderRep({ object }) {
-  return Rep({ object, defaultRep: Grip });
+function renderRep({ object, mode }) {
```
wldcordeiro: Should I have a default mode value?

public/js/components/ObjectInspector.js
jasonLaster: I'm not sure either, I think we probably do... 

Lets take a look at the tests and see if everything makes sense
https://dxr.mozilla.org/mozilla-central/source/devtools/client/shared/components/test/mochitest/test_reps_object.html#83

public/js/components/Rep.js
jasonLaster: Hmm, whats the default behavior if undefined is passed in?

public/js/components/ObjectInspector.js
jasonLaster: based on the tests tiny looks fine. @linclark thoughts?

public/js/components/Rep.js
wldcordeiro: Looks like it uses the default output? https://dxr.mozilla.org/mozilla-central/source/devtools/client/shared/components/test/mochitest/test_reps_object.html#79

jasonLaster: 👍

jasonLaster:
👍 

Once this lands we can create an m-c issue for updating the Object Inspector

public/js/components/ObjectInspector.js
linclark: @jasonLaster I don't know enough about the desired output across the tree and the way reps are being used here to have an opinion.

wldcordeiro: I'll leave this to you @jasonLaster heh, I'm in the same boat as @linclark though it does seem to look like your expect screenshot.

jasonLaster: Thanks @wldcordeiro. I'm okay moving forward with this and adjusting if we need to.

jasonLaster:
Thanks @wldcordeiro!

PR #599, opened 2016-08-26
Hide pretty print button for original sources

### LINKED ISSUE #419: Pretty Print - Pretty printing an original source
Case:
+ a bundle file "a.js" has been source mapped, creating original files "b.js", "c.js"
+ user pretty prints "b.js" creating "d.js"

One likely use case, is the user has created a bundle with a minified version of jquery. 

Problem:
+ do we support adding breakpoints in "b.js" and "d.js"? If we do we have two source map hops to "a.js", which is difficult. 

Solution:
+ I propose we disable adding breakpoints to "b.js". In this case, it's only possible to add breakpoints to "a.js" and "c.js" i.e. the bundle file and pretty version of jquery. The benefit is that the original source map can be expanded to incorporate the prettified version of jquery.

jasonLaster:
Fixes #419.


jlongster:
Looks good!

PR #601, opened 2016-08-26
Hide breakpoints from other pages

### LINKED ISSUE #479: Hide Breakpoints from a different page
Steps to reproduce:
1. go to a page, add breakpoints
2. go to another page, add breakpoints

you'll see breakpoints from the original page as well. We should hide those as they are not relevant.

Notes:
+ this can likely be implemented with a selector, which can filter the breakpoints list by checking to see if the breakpoint source is still in the store. 

Here's how chrome handles navigation behavior:

jasonLaster:
Fixes #479.

This is a simple fix. I'd like to re-think how we're storing breakpoints later when we introduce a localstorage backed pref system. My thinking is that it would be nice to store breakpoints in local storage so that we can access them after the debugger is closed/re-opened. If we start doing that, it might make sense to only load breakpoints for a page when the debugger starts.

### DIFF
--- a/public/js/components/Breakpoints.js
+++ b/public/js/components/Breakpoints.js
@@ -114,7 +114,8 @@ function _getBreakpoints(state) {
     bp.locationId = locationId;
     bp.isCurrentlyPaused = isCurrentlyPaused;
     return bp;
-  });
+  })
+  .filter(bp => bp.location.source);
 }
 
 module.exports = connect(


jlongster:
Yeah, this works. It probably will be different if we ever persist breakpoints.

jasonLaster:
Agreed.

Issue #596, opened 2016-08-26
Update contributing doc


jbhoosreddy:
In CONTRIBUTING.md, in the directions for setting up Chrome, the port number should be changed to 7999 from 8000, I believe.

clarkbw:
> 7999 from 8000

Oh right, thanks!  Good catch!

PR #602, opened 2016-08-26
Update contributing.md to have the right port value for Chrome.

wldcordeiro:
#596 mentioned this port change.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -113,7 +113,7 @@ Once you have Firefox running in a debuggable state go back up to instructions f
 Start by running Chrome with remote debugging turned on, this command also creates a new _temporary_ profile
 
 ```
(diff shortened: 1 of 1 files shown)


clarkbw:
> this port change

Sorry, I should have clarified in the original issue.

The port you're looking for is in the Chrome command here: https://github.com/devtools-html/debugger.html/blob/master/CONTRIBUTING.md#chrome

localhost:8000 will show the debugger interface
But any mention of the examples directory, so http://localhost:8000/todomvc/ should use `7999` 
Also check for `/increment/` or `/pythagorean/`

wldcordeiro:
@clarkbw I updated the PR I'm not seeing any mentions of `/increment/` or `/pythagorean/` that would need the replacement.

CONTRIBUTING.md
```diff
@@ -222,7 +222,7 @@ To enabled Hot Reloading:
 
 ```json
 {
-  "hotReloading": true
```
wldcordeiro: Apparently my editor trimmed some whitespace automagically.

clarkbw: That's fine!

jasonLaster:
Thanks for this fix!

PR #603, opened 2016-08-27
Persist tabs

jbhoosreddy:
Well @jasonLaster I realized which parts were not working. The second commit resolves those issues and test failures.

### DIFF
--- a/public/js/clients/firefox.js
+++ b/public/js/clients/firefox.js
@@ -101,8 +101,8 @@ function initPage(actions) {
   threadClient = getThreadClient();
 
   setupCommands({ threadClient, tabTarget, debuggerClient });
-
-  tabTarget.on("navigate", actions.navigate);
+  tabTarget.on("will-navigate", actions.willNavigate);
+  tabTarget.on("navigate", actions.willNavigate);
   tabTarget.on("frame-update", function(_, packet) {
     if (packet.destroyAll) {
       actions.willNavigate();
--- a/public/js/clients/firefox.js
+++ b/public/js/clients/firefox.js
@@ -102,7 +102,6 @@ function initPage(actions) {
 
   setupCommands({ threadClient, tabTarget, debuggerClient });
 
-  tabTarget.on("will-navigate", actions.willNavigate);
   tabTarget.on("navigate", actions.navigate);
   tabTarget.on("frame-update", function(_, packet) {
     if (packet.destroyAll) {
--- a/public/js/reducers/sources.js
+++ b/public/js/reducers/sources.js
@@ -100,10 +100,12 @@ function update(state = State(), action: Action) : Record<SourcesState> {
       }
 
       return _updateText(state, action, [action.originalSource]);
+
     case "NAVIGATE":
-      // Reset the entire state to just the initial state, a blank state
-      // if you will.
-      return State();
+      const source = state.selectedSource;
+      const sourceUrl = source && source.get("url");
+      return State()
+        .set("pendingSelectedSourceURL", sourceUrl);
   }
 
   return state;
--- a/public/js/reducers/sources.js
+++ b/public/js/reducers/sources.js
@@ -12,8 +12,7 @@ import type { Record } from "../utils/makeRecord";
 
 export type SourcesState = {
   sources: I.Map<string, any>,
-  selectedSource: ?any,
-  pendingSelectedSourceURL: ?string,
+  selectedSourceURL: ?any,
   sourcesText: I.Map<string, any>,
   tabs: I.List<any>,
   sourceMaps: I.Map<string, any>,
@@ -21,8 +20,7 @@ export type SourcesState = {
 
 const State = makeRecord(({
   sources: I.Map(),
-  selectedSource: undefined,
-  pendingSelectedSourceURL: undefined,
+  selectedSourceURL: undefined,
   sourcesText: I.Map(),
   sourceMaps: I.Map(),
   tabs: I.List([])
@@ -54,17 +52,13 @@ function update(state = State(), action: Action) : Record<SourcesState> {
 
     case "SELECT_SOURCE":
       return state.merge({
-        selectedSource: action.source,
-        pendingSelectedSourceURL: null,
-        tabs: updateTabList(state, fromJS(action.source), action.options)
+        selectedSourceURL: action.url,
+        tabs: updateTabList(state, fromJS(action.url), action.options)
       });
-
-    case "SELECT_SOURCE_URL":
-      return state.merge({ pendingSelectedSourceURL: action.url });
 
     case "CLOSE_TAB":
       return state.merge({
-        selectedSource: getNewSelectedSource(state, action.id),
+        selectedSource: getNewSelectedSource(state, action.url),
         tabs: removeSourceFromTabList(state, action.id)
       });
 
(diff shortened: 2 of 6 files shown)


public/js/reducers/sources.js
```diff
@@ -224,15 +215,15 @@ function getSourceText(state: OuterState, id: string) {
 }
 
 function getSourceTabs(state: OuterState) {
-  return state.sources.tabs;
+  const tabs = state.sources.tabs;
+  return tabs
+    .filter(url => getSourceByURL(state, url))
+    .map(url => getSourceByURL(state, url));
```
jasonLaster: It'd probably be more efficient to map and then filter with the identity function. That avoids doing the work twice

public/js/reducers/sources.js
```diff
@@ -12,17 +12,15 @@ import type { Record } from "../utils/makeRecord";
...
+  selectedSourceURL: ?any,
   sourcesText: I.Map<string, any>,
   tabs: I.List<any>,
   sourceMaps: I.Map<string, any>,
 };
 
 const State = makeRecord(({
   sources: I.Map(),
-  selectedSource: undefined,
-  pendingSelectedSourceURL: undefined,
```
jasonLaster: Pending source is not needed now that selected source is selected source URL

jasonLaster:
Mind writing an integration test that tests refreshing, selecting and closing tabs?

jasonLaster:
By the way, now that tabs are persisted, i just realised that the follow up task is hiding tabs from other pages.

jbhoosreddy:
i'll make a follow-up commit with integration test in this PR. And after that has merged, I'll create a follow up PR for hiding tabs.

jasonLaster:
Sounds good. Let's not feature creep :)

jbhoosreddy:
@jasonLaster I'm still finding the issues you were talking about. 😅
The new changes are WIP. I'll add integration tests once this is resolved. I'm looking to hear what you think about saving url in the state from the tabTarget.url and returning state depending on that. The good news is that persistence works properly now.

public/js/reducers/sources.js
jbhoosreddy: could you briefly tell me what would be the effect of adding a filter with an identity function to the result of map?

jasonLaster: Sure. The map would return an object or null. Filter would remove the
undefined.

> Tab target URL

I think that's a good idea, but I'm not sure where it belongs. The
advantage is that we can always see which is the current debuggee URL

jbhoosreddy: I took "url" from the navigate/will-navigate event payload from the tabTarget. But more importantly, this would only work for Firefox. I'll look to see if I can do something for chrome.

jasonLaster: Hmm, we should have something like navigate for chrome.

public/js/clients/firefox.js
```diff
@@ -101,9 +101,8 @@ function initPage(actions) {
   threadClient = getThreadClient();
 
   setupCommands({ threadClient, tabTarget, debuggerClient });
-
   tabTarget.on("will-navigate", actions.willNavigate);
-  tabTarget.on("navigate", actions.navigate);
+  tabTarget.on("navigate", actions.willNavigate);
```
jasonLaster: I think is probably better to make the `navigate` action meaningful than to change to `willNavigate`

public/js/reducers/sources.js
```diff
@@ -100,10 +96,19 @@ function update(state = State(), action: Action) : Record<SourcesState> {
       }
 
       return _updateText(state, action, [action.originalSource]);
+
     case "NAVIGATE":
```
jasonLaster: if `willNavigate` fires the `navigate` action type we should change that to will-navigate and then fire navigate separately

public/js/reducers/sources.js
```diff
@@ -100,10 +96,19 @@ function update(state = State(), action: Action) : Record<SourcesState> {
...
+        return state;
+      }
+      if (!state.url) {
+        return State()
+          .set("url", action.data.url);
+      }
+      if (state.url === action.data.url) {
+        return state;
+      }
+      return State().set("url", action.data.url);
```
jasonLaster: this seems like a lot of code to support the case where `will-navigate` is fired from the other events.

Also, i don't think that `url` should be set in `sources`. It probably makes sense in a debuggee reducer as at somepoint we might be debugging a worker URL as well and perhaps other things so there would be different debuggee data.

public/js/clients/firefox.js
jbhoosreddy: I opted against changing navigate after reading the comments in that method. But it makes sense, I'll move it it to navigate.

public/js/reducers/sources.js
jbhoosreddy: this code can be simplified by following up on your previous comments.

`targetUrl` could probably make sense for tabs and node also. Although, I haven't looked into node yet.

I am using `url` to check if the new target url is same as the old one, and based on that returning `state` or new `State()`

I'll wait for jlongster's thoughts on these changes.

jbhoosreddy: @jasonLaster do you have any suggestion for refreshing the target debuggee in cypress without refreshing the debugger (for the integration tests)?

jasonLaster: @jbhoosreddy `cy.navigate` should do that for you : https://github.com/devtools-html/debugger.html/blob/master/public/js/test/cypress/commands.js#L42-L48

jlongster: The sources seems like a weird place to put the URL. I don't have time to fully review this code but there are several problems with it, which I will mention in another comment below.

jlongster:
These changes won't work. A key part of sources is that you can't base them off URL. Many sources do not have a URL (eval'ed sources is just one example). Please don't change the `SELECT_SOURCE` action, it was correct before. With your code, you won't be able to view eval'ed sources (and other edge cases).

I don't have time to do a full review, I'm on PTO until tomorrow afternoon. But we really need to write a bunch more tests so things like this obviously fail. I should write a bunch that deal with weird sources like eval'ed sources.

jasonLaster:
> Many sources do not have a URL

I believe all sources now have a [url](https://github.com/devtools-html/debugger.html/blob/master/public/js/clients/firefox/events.js#L61-L63). But it's true that they are not relative to the target URL. If that is a concern, I think we can address that here as well as it would make sense to have all source url's relative to the target.

> tests

We have a couple tests now that deal with eval sources
+ [break on next](https://github.com/devtools-html/debugger.html/blob/master/public/js/test/integration/tests.js#L75) pauses in the eval of a console script
+ [evals](https://github.com/devtools-html/debugger.html/blob/master/public/js/test/integration/tests.js#L172) is a test that pauses in two evals with and without sourceURLs. 

We probably should have more, but we do have some. 

It'd be nice to have tabs persist as the debuggee is navigated. Tabs is a new feature we're introducing with this debugger and I'd like for it to be polished.

jlongster:
> I believe all sources now have a url. But it's true that they are not relative to the target URL. If that is a concern, I think we can address that here as well as it would make sense to have all source url's relative to the target.

We shouldn't be doing that, in my opinion. The URL should be null. How we display sources with null URLs is a UI concern, and we shouldn't be relying on that in anything except the components. It's much easier to check if `source.url` is null in tests, is more predictable, etc. You don't want `SOURCE6` to be automatically selected when you reload because it'll most likely be a different source.

I agree that we should get tabs persisting, but we need to do a different way. I don't fully understand why persisting the tabs requires those changes. We should keep a list of URLs currently open and use those across refreshes.

jlongster:
Mulling this over a little more, I think the discussion boils down to this: we shouldn't save tabs that are there for non-URL based sources. It doesn't make sense. It'll more likely than not show the wrong source, or it may not even exist.

jasonLaster:
> we shouldn't save tabs that are there for non-URL

I agree with that. We have an `isHiddenSource` method we can re-use to exclude these tabs
https://github.com/devtools-html/debugger.html/blob/master/public/js/utils/sources-tree.js#L7-L10

jbhoosreddy:
based on all the information discussed, I'll make a fresh PR for this.

PR #611, opened 2016-08-29
Fix #593 : Add this variable for inner scope

### LINKED ISSUE #593: Scopes - show <this> for all of the call stack frames
We're currently only showing the <this> property for the first frame in the call stack, but we should show it for all frames.

The issue is likely [here](https://github.com/devtools-html/debugger.html/blob/master/public/js/components/Scopes.js#L102-L105) where we check to see if we're showing the *innermost* scope, when in reality we want to show `<this>` in the local scope of every frame. The fix is likely a cleanup of `getScopes` that treats the first scope (local scope) differently than the other n-1 scopes.

This is a good bug for contributors who likely the geeky problems :)

ThomasCrvsr:


public/js/components/Scopes.js
```diff
@@ -94,10 +108,14 @@ function getScopes(pauseInfo, selectedFrame) {
...
 
       // show exception, return, and this variables in innermost scope
-      if (scope.actor === pauseInfo.getIn(["frame", "scope"]).get("actor")) {
+      if (isScopeOfPauseFrame(scope.actor)) {
         vars = vars.concat(getSpecialVariables(pauseInfo, key));
       }
 
+      if (scope.actor === selectedScope.actor) {
+        vars = vars.concat(getThisVariable(selectedFrame, key));
+      }
```
jasonLaster: Oh nice. I bet this lets you see `<this>` for any selected frame.

public/js/components/Scopes.js
```diff
@@ -72,13 +78,21 @@ function getScopes(pauseInfo, selectedFrame) {
...
   }
 
   const scopes = [];
 
+  let scope = selectedScope;
+
+  function isScopeOfPauseFrame(innerScope) {
+    const pauseActor = pauseInfo.getIn(["frame", "scope"]).get("actor");
+    return innerScope.actor === pauseActor;
+  }
```
jasonLaster: Nit: I'd prefer if we got the `pausedScopeActor`at the top of `getScopes` and used it below:

```js
if (scope.actor === pausedScopeActor) {
}
```

public/js/components/Scopes.js
```diff
@@ -56,11 +55,18 @@ function getSpecialVariables(pauseInfo, path) {
...
+  const vars = [];
+
   if (this_) {
     vars.push({
       name: "<this>",
       path: path + "/<this>",
-      contents: { value: this_.toJS() }
+      contents: { value: this_ }
     });
   }
```
jasonLaster: 👍

public/js/components/Scopes.js
ThomasCrvsr: Yep, good idea !

public/js/components/Scopes.js
```diff
@@ -56,11 +55,18 @@ function getSpecialVariables(pauseInfo, path) {
...
 
+  return vars;
+}
+
+function getThisVariable(frame, path) {
+  const this_ = frame.this;
+  const vars = [];
```
jasonLaster: NIT: I would prefer returning null or the obj

```js
if (!this_) {
  return null
}

```

ThomasCrvsr:
Do you know if the only way to get Cypress is by filling the early adopter form ? I couldn't run `npm run test-all` because it's missing on my machine.

jasonLaster:
@ThomasCrvsr yup - it's not too bad. These docs walk you through getting set up. You basically stop by https://gitter.im/cypress-io/cypress and ask nicely.

https://github.com/devtools-html/debugger.html/blob/master/docs/integration-tests.md

jasonLaster:
@ThomasCrvsr also, it'd be nice to add some checks to the existing integration test for another scope this

It'd be similar to [this](https://github.com/devtools-html/debugger.html/commit/3d62a69ce80292fbffb532d2b25615ce6d8598f6#diff-5ab3592cecb716562600466967e88b4fR14).

Cypress might be a pain to setup, but it's fun to run it. I actually fixed `this` last week and used a test to automate getting to that spot in the debugger to fix.

jasonLaster:
Also, feel free to stop by our IRC channel #devtools in irc.mozilla.org and say hi sometime

ThomasCrvsr:
@jasonLaster Yep, I kept in mind adding some test but I wanted to show the implementation first ^^
I will try to find some time this afternoon to get all this working :)

(I joined the #debugger-html channel, but maybe this is not the main channel ^^)

ThomasCrvsr:
@jasonLaster integration tests don't seem to pass, even on `master`. Do you have the same situation ?

jasonLaster:
Stepping out for an hour. I'll fix the tests when I'm back

jasonLaster:
closing in favor of #617. The only thing I changed was:
* squash commits
* add <this> assertion
* rebase off master

PR #616, opened 2016-08-29
Fix tests

jasonLaster:
+ Update pausing test to check for `Block`, which is what the old
debugger does
+ Update how the exceptions test reloads the debuggee to do a proper
reload


jasonLaster:
Merging as this only affects the integration tests.

PR #617, opened 2016-08-29
Fix #593 : Add this variable for inner scope

### LINKED ISSUE #593: Scopes - show <this> for all of the call stack frames

jasonLaster:
* Fix #593 : Use a variable for pausedScope actor instead of function
* Fix #593 : Return an object for 'this' variable instead of an array
* Add <this> test for second scope


jasonLaster:
Thanks @ThomasCrvsr for the hard work tackling a difficult first bug!

Issue #584, opened 2016-08-25
File Search - close ui


jasonLaster:
It would be nice to have a button in the top right for exiting file search

![screen shot 2016-08-25 at 4 32 18 pm](https://cloud.githubusercontent.com/assets/254562/17984891/4d06b13a-6ae1-11e6-8679-9a4ebbf58cc2.png)

Here's a sketch of what the change could look like:
```diff
diff --git a/public/js/components/App.js b/public/js/components/App.js
index 306f664..764a100 100644
--- a/public/js/components/App.js
+++ b/public/js/components/App.js
@@ -46,6 +46,7 @@ const App = React.createClass({
   displayName: "App",
 
   getInitialState() {
+    // here's the flag for search!
     return {
       searchOn: false
     };
@@ -67,6 +68,10 @@ const App = React.createClass({
     this.setState({ searchOn: !this.state.searchOn });
   },
 
+  closeFileSearch() {
+    this.setStae("searchOn", false);
+  },
+
   onKeyDown(e) {
     if (e.key === "Escape") {
       this.setState({ searchOn: false });
@@ -75,13 +80,19 @@ const App = React.createClass({
   },
 
   renderSourcesSearch() {
-    return Autocomplete({
-      selectItem: result => {
-        this.props.selectSource(result.id);
-        this.setState({ searchOn: false });
-      },
-      items: searchResults(this.props.sources)
-    });
+    return
+    dom.div({},
+       dom.div({className: "close-button"},
+        // insert an SVG or the right html code for an X whatevs
+       ),
+       Autocomplete({
+        selectItem: result => {
+          this.props.selectSource(result.id);
+          this.setState({ searchOn: false });
+        },
+        items: searchResults(this.props.sources)
+      })
+    }
   },
 
   renderEditor() {
```

wldcordeiro:
I could do this. @jasonLaster 

> It would be nice to have a button in the top right for exiting file search

Like an `x` in the far right of the search bar?

Regarding that patch? Is that just a concept/idea for how you could do it?

jasonLaster:
Concept. Wrote it quickly

I was thinking top right of the pane.

X in the input in my mind would clear the search... Could be persuaded
though

wldcordeiro:
Could the div wrapping the two be given padding around the input and place the x just outside the input?

wldcordeiro:
Working on styling the button now. Is there preference on units and layout? I started using flexbox, but want to make sure I'm writing things that fit in.

[...]

PR #619, opened 2016-08-29
File search close

wldcordeiro:
Associated with #584 adds the file search close button. I noticed a bug with `Svg` when I passed the `onClick` without the `className` prop also existing. Fixed this in a single commit that I can roll back and put in another PR.

Also still working on styling. Need some guidance there on what CSS is appropriate (in terms of units to use, flexbox, etc)

Screenshot
screenshot_20160829_140522

### DIFF
--- a/public/images/Svg.js
+++ b/public/images/Svg.js
@@ -30,7 +30,10 @@ module.exports = function(name, props) { // eslint-disable-line
   if (!svg[name]) {
     throw new Error("Unknown SVG: " + name);
   }
-  let className = props ? `${name} ${props.className}` : name;
+  let className = name;
+  if (props && props.className) {
+    className = `${name} ${props.className}`;
+  }
   if (name === "subSettings") {
     className = "";
   }
--- a/public/js/components/App.css
+++ b/public/js/components/App.css
@@ -84,4 +84,15 @@ body {
 
 .search-container {
   display: flex;
+  flex: 1;
+}
+
+.search-container .autocomplete {
+  flex: 1;
+}
+
+.search-container .close-button {
+  width: 16px;
+  margin-top: 2em;
+  margin-right: 2em;
 }
--- a/public/js/components/App.css
+++ b/public/js/components/App.css
@@ -81,3 +81,7 @@ body {
   text-align: center;
   width: 100%;
 }
+
+.search-container {
+  display: flex;
+}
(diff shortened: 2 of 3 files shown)


jasonLaster:
@wldcordeiro I think you can do better 😸 

I recommend using flexbox and then maybe em for height of the svg. i bet we'll be okay if we stick to 1.0 em.

Thoughts @helenvholmes 

Also, thanks for filing an early PR. That's the easiest way to give feedback

public/js/components/App.js
```diff
@@ -75,14 +76,22 @@ const App = React.createClass({
...
+      Autocomplete({
+        selectItem: result => {
+          this.props.selectSource(result.id);
+          this.setState({ searchOn: false });
+        },
+        items: searchResults(this.props.sources)
+      }),
+      dom.div({ className: "close-button" },
+      Svg("close", { onClick: this.closeSourcesSearch }))
+    );
```
jasonLaster: 👍

wldcordeiro:
@jasonLaster Noticed that the search icon in the autocomplete is set to `16px` width (which changes the height too) and the default font-size is coming out to `13.333px` for me since it uses browser default so maybe slightly higher in `em`s? I have no problem using whatever units. :) Just wanted to get that figured out so I didn't need to do it twice.

jasonLaster:
ahh - 16px is probably fine too. It's nice to be consistent w/ the file at hand.

public/images/Svg.js
```diff
@@ -30,7 +30,10 @@ module.exports = function(name, props) { // eslint-disable-line
   if (!svg[name]) {
     throw new Error("Unknown SVG: " + name);
   }
-  let className = props ? `${name} ${props.className}` : name;
+  let className = name;
```
wldcordeiro: Maybe the better solution here would be for me to use the `classnames` lib like the rest of the app uses?

public/js/components/App.css
```diff
@@ -81,3 +81,18 @@ body {
...
+
+.search-container .autocomplete {
+  flex: 1;
+}
+
+.search-container .close-button {
+  width: 16px;
+  margin-top: 2em;
+  margin-right: 2em;
+}
```
jasonLaster: 👍

jasonLaster:
Thanks, looks great and the patch reads really well.

PR #620, opened 2016-08-29
Fixed right sidebar header

### LINKED ISSUE #605: Right Sidebar - the header should be fixed
The right sidebar header currently scrolls out of view. It should be fixed on the top so that the commands are always available.

We can probably do more to improve the scrolling behavior of the accordion, like making the accordion item headers sticky, so the top most item header is always visible, but that is probably out of scope for now. 

screen shot 2016-08-28 at 2 01 23 pm

screen shot 2016-08-28 at 2 01 11 pm

wldcordeiro:
Fixes #605

Simple change of the container to make it a flex container and add the proper scroll support for the accordion.

### DIFF
--- a/public/js/components/RightSidebar.css
+++ b/public/js/components/RightSidebar.css
@@ -1,13 +1,18 @@
 .right-sidebar {
+  display: flex;
+  flex-direction: column;
   flex: 1;
   white-space: nowrap;
+}
+
+.right-sidebar .accordion {
+  overflow-y: scroll;
 }
 
 .command-bar {
   border-bottom: 1px solid #cccccc;
   height: 30px;
   padding: 8px 5px 10px 10px;
-  overflow: hidden;
 }
 
 .command-bar > span {


public/js/components/RightSidebar.css
```diff
@@ -1,13 +1,19 @@
 .right-sidebar {
+  display: flex;
+  flex-direction: column;
   flex: 1;
   white-space: nowrap;
 }
 
+.right-sidebar .accordion {
+  overflow-y: scroll;
+  -webkit-overflow-scrolling: touch;
```
wldcordeiro: May not need this `-webkit` property here, but might be useful in cases for outside Firefox?

jasonLaster: yeah - i don't think we need that property :)

jasonLaster:
one nit. otherwise ready to go

public/js/components/RightSidebar.css
wldcordeiro: 👍

wldcordeiro:
Nit addressed and commit amended. 😄

jasonLaster:
💃  in

PR #621, opened 2016-08-30
Fix #610 : Increase the height of the breakpoint svg

### LINKED ISSUE #610: Breakpoint - styling
the breakpoint marker is not centered:

screen shot 2016-08-28 at 10 17 32 pm

ThomasCrvsr:


wldcordeiro:
Looks better with `15px` and a little better still with `16px` as it gives a slightly more even amount of padding around the number.

15px
screenshot_20160830_065554

16px
screenshot_20160830_065522

either way looks better than the default though heh!

jasonLaster:
Thanks @wldcordeiro @ThomasCrvsr are you testing on chrome? I think this issue is difficult because it is a chrome / firefox thing.

I looked into this briefly with @helenvholmes and we found that the issue is code mirror gutter line font-sizes. In firefox it is 11 and in chrome it is 13px

screen shot 2016-08-30 at 11 08 00 am

jasonLaster:
@ThomasCrvsr this should be a good fix

```patch
diff --git a/public/js/lib/themes/dark-theme.css b/public/js/lib/themes/dark-theme.css
index 48516a8..b4254b7 100644
--- a/public/js/lib/themes/dark-theme.css
+++ b/public/js/lib/themes/dark-theme.css
@@ -246,6 +246,7 @@ div.CodeMirror span.eval-text {

 .cm-s-mozilla .CodeMirror-linenumber { /* line number text */
   color: var(--theme-content-color3);
+  font-size: 11px;
 }

 .cm-s-mozilla .CodeMirror-gutters { /* vertical line next to line numbers */
diff --git a/public/js/lib/themes/light-theme.css b/public/js/lib/themes/light-theme.css
index 808eab4..edfecfa 100644
--- a/public/js/lib/themes/light-theme.css
+++ b/public/js/lib/themes/light-theme.css
@@ -246,6 +246,7 @@ div.CodeMirror span.eval-text {

 .cm-s-mozilla .CodeMirror-linenumber { /* line number text */
   color: var(--theme-content-color3);
+  font-size: 11px;
 }
```

ThomasCrvsr:
@jasonLaster Yes, I saw weird little differences between Firefox and Chrome but couldn't point to the exact thing. Thanks !

I'll revert my commit and add your proposition, sounds good ?

Also, I've got the impression that there is a lot of common code between those two themes, should we try to refactor those part into one spot ?

jasonLaster:
@ThomasCrvsr that's a great point. Actually we probably would benefit from **NOT** touching them as we pull them in from the larger devtools project. Lets add the class to the Editor.css file or something like that

ThomasCrvsr:
@jasonLaster Does this fix looks good ? (I tried in Chrome and Firefox, this is better :) ).

Also, I looked at `Editor.css` and saw the class `.editor` but couldn't find any usage of it in the code. Do you know if it's obsolete ?

jasonLaster:
It's likely out of date. We just moved editor implementations to default to
the devtools codemirror editor. I think you can add the class there safely.

ThomasCrvsr:
Which class are you talking about ?

jasonLaster:
Sorry @ThomasCrvsr, i didn't see that you updated the patch.

Issue #426, opened 2016-07-27
Collapse/Expand sidebar UI


jasonLaster:
We should have a button next to the sidebars for collapsing them. Clicking them should start a slidey animation of course.

NOTE: don't worry about getting the right icon. It could be anything, we can fix that later :)

![screen shot 2016-07-27 at 11 35 34 am](https://cloud.githubusercontent.com/assets/254562/17181437/237ebe0c-53ee-11e6-9205-e4aca9474040.png)

jasonLaster:
This is a small enhancement that I think we can get to next sprint.

jbhoosreddy:
@jasonLaster thanks! I'll work on it in the next sprint!

clarkbw:
@helenvholmes has designs for sidebars across the toolboxes

wldcordeiro:
I like the idea of taking on this task. 😄 I had an idea, rather than an icon in the right panel, how about changing the `Draggable` component? I'm thinking something like Illustrator/Lightroom so increase the width of the Draggable, put an arrow icon there and make a double click toggle the collapse. I'm willing to implement it however but figured a suggestion never hurts.

helenvholmes:
@clarkbw I have sidebar designs for sidebars with additional tabs, but I think they're going to be different from the debugger, which instead is one tab, with an accordion.

Sidebar with panels: https://invis.io/F68A5KHKY#/179720580_Sidebar_Tabs 

Sidebar with one panel, accordion (ignore old icon cruft): https://cl.ly/3Y1j3G3F0s1r

@clarkbw: Do we want to split up the Debugger to have panels for some things instead of a monolithic accordion?

@wldcordeiro I'm intrigued by your suggestion, but I don't think I understand it yet + I want to make sure it works across all of the tools, not just the debugger. Could you post a ./mov file with what you're imagining? :)

jasonLaster:
@helenvholmes I think the panel suggestion is a good one. I'd like to suggest that we schedule some time in October to kickoff some larger UX projects for the debugger. I'm worried that we don't have the bandwidth to do this work properly beforehand.

jasonLaster:
Quick update:

* Here is the current [h-split-box](https://dxr.mozilla.org/mozilla-central/source/devtools/client/shared/components/h-split-box.js) 
* the v-split-box is coming soon. couple test failures
[...]

PR #623, opened 2016-08-30
Collapsible left and right sidebars.

wldcordeiro:
Issue #426 - I had an idea in the issue that I went with here. Collapse is done by double clicking the sidebars. Styling is still WIP.

I'll record a video shortly @jasonLaster 

screenshot_20160830_095353
screenshot_20160830_095421

### DIFF
--- a/public/images/Svg.js
+++ b/public/images/Svg.js
@@ -13,6 +13,7 @@ const svg = {
   "folder": require("./folder.svg"),
   "globe": require("./globe.svg"),
   "magnifying-glass": require("./magnifying-glass.svg"),
+  "pane-collapse": require("./pane-collapse.svg"),
   "pause": require("./pause.svg"),
   "pause-circle": require("./pause-circle.svg"),
   "pause-exceptions": require("./pause-exceptions.svg"),
--- a/public/images/pane-collapse.svg
+++ b/public/images/pane-collapse.svg
@@ -1,0 +1,7 @@
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+<svg width="16" height="16" viewBox="0 0 16 16" xmlns="http://www.w3.org/2000/svg" fill="#0b0b0b">
+  <path fill-opacity=".3" d="M12,3h2v10h-2V3z M5,9.9V6.1L8,8L5,9.9z"/>
+  <path d="M14,2H2C1.4,2,1,2.4,1,3v10c0,0.6,0.4,1,1,1h12c0.6,0,1-0.4,1-1V3C15,2.4,14.6,2,14,2z M2,13L2,13V3h0h9v10   H2L2,13z M14,13C14,13,14,13,14,13h-2V3h2c0,0,0,0,0,0V13z M8.5,7.2l-3-1.9C4.6,4.7,4,5,4,6.1v3.8c0,1.1,0.6,1.4,1.5,0.8l3-1.9   C9.5,8.3,9.5,7.8,8.5,7.2z M5,9.9V6.1L8,8L5,9.9z"/>
+</svg>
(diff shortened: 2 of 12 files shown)


helenvholmes:
Hey Wellington, 

I think we'd like to add a small animation to the expand/collapse icon as well, if you have the time: https://cl.ly/3q2C1w1b3U0e

(I can also open a new bug for it.)

The icon in the .mov is different from the icon you added—that's because the animation was comp'd when we had older icons, but the idea is still the same for how we want it to animate :)

wldcordeiro:
@helenvholmes Here's a little video. https://www.dropbox.com/s/3kq2b98l5fuc8dl/collapse-sidebar.mp4?dl=0

jasonLaster:
Hey @wldcordeiro what's the status here? I'm not sure if you're looking for a review

wldcordeiro:
@jasonLaster I was awaiting @helenvholmes response, in the issue she had asked me for a video to show what I meant, I'm working on the styling to use the svg provided though (I still like the idea of just increasing the width of the `Draggable` by 1-2px and adding the double click to it as an additional way to collapse. 😄

jasonLaster:
Cool. Did you get Helens message with the svgs?

helenvholmes:
Based on your commit messages (I haven't pulled this down), looks like we ended up going with what's currently in the tools? (https://cl.ly/3S0B1D210I1T ?)

clarkbw:
@wldcordeiro I think you're missed adding the new files, SidebarHeader etc. 🙀 😸

wldcordeiro:
@clarkbw I have a local commit I was working on that I hadn't pushed up yet. 😆

jasonLaster:
thanks @wldcordeiro 

Couple things:
+ i noticed that clicking on the collapsey thing did not work. Could it be a bug?
+ i believe the buttons should be on the other side of the draggable? This might be an issue with how you've setup the headers?
+ Could we do the double click behavior in a later PR? I think this is adding a lot to this PR and making it hard to review.

wldcordeiro:
>  i noticed that clicking on the collapsey thing did not work. Could it be a bug?

Not yet implemented, just putting the skeleton there since it was going in a new component.

> i believe the buttons should be on the other side of the draggable? This might be an issue with how you've setup the headers?

Do you mean they should be in the tab bar area rather than in the sidebar?

> Could we do the double click behavior in a later PR? I think this is adding a lot to this PR and making it hard to review.

Sure.

wldcordeiro:
@helenvholmes where would it be preferred to put the collapse buttons? In the sidebar headers like this? Or in the tab bar?

screenshot_20160901_154403

jasonLaster:
One reason I mentioned putting the button in the center is that the header realestate is more valuable. I.e I like the debugger buttons on the left and we have copy for the right side of the sources header.

jlongster:
I agree with Jason, I think where it is in that screenshot is too prominent. A few notes:

* @helenvholmes we talked about the idea of not having any icons, but a "gesture" where if the user slides it far enough in that it just collapses. Have we tried that? I guess there is the problem of how to open it back up.
* Is it just me or does the icon seem darker then the rest of the icons?  Or is it just thicker?
* I suppose we could copy Chrome and put them in the editor header...

I thought I had another question but I forget. It's awesome to see this worked on, thanks @wldcordeiro !

jlongster:
Oh, I remember: in https://bugzilla.mozilla.org/show_bug.cgi?id=1260552 Honza has ported our Draggable and SplitBox components to m-c. But we're modifying Draggable... I suppose it's ok and we'll just deal with it when we try to copy back his new SplitBox which has a bunch of new stuff.

jlongster:
Also, one last thing (sorry): any reason the width of the splitter is increased a lot? Can we keep it consistent with the rest of the borders but just make the clickable area as big as you want it?

helenvholmes:
> @helenvholmes we talked about the idea of not having any icons, but a "gesture" where if the user slides it far enough in that it just collapses. Have we tried that? I guess there is the problem of how to open it back up.
I don't think we could have the gesture without the icon; with the icon, you click the icon, that opens it back up.

wldcordeiro:
Calling PR bankruptcy on this one and will reopen from another branch based on discussion above. 👍

PR #630, opened 2016-08-30
Add mochitest infrastructure

jlongster:
This is my final pass on the mochitest infrastructure. It should actually work! This implements running them locally and on try. Locally it downloads an artifact build of Firefox and you should be up an running in ~10 min, and future builds should be almost instantaneous. For debugging you'll probably want to cd into the Firefox directory itself and use mochitests like normal, not through `npm run mochitests`. On CI, it runs them in a pre-built docker container.

This pulls over @jasonLaster's great work from bug [1294686](https://bugzilla.mozilla.org/show_bug.cgi?id=1294686) because we need to actually commit the mochitests here. We still need to land a few bits in that bug to make these run (modifying other `moz.build`s etc), but the `head.js` and a simple test was copied into here. It gives us a starting point.

**Don't merge this** yet; it will not work until relevant bits are landed in bug 1294686. Once they do, I'll update the Firefox in docker and they should be running in CI.

### DIFF
--- a/public/js/test/mochitest/head.js
+++ b/public/js/test/mochitest/head.js
@@ -1,0 +1,176 @@
+/* -*- indent-tabs-mode: nil; js-indent-level: 2 -*- */
+/* vim: set ft=javascript ts=2 et sw=2 tw=80: */
+/* Any copyright is dedicated to the Public Domain.
+ * http://creativecommons.org/publicdomain/zero/1.0/ */
+
+"use strict";
+
+// shared-head.js handles imports, constants, and utility functions
+Services.scriptloader.loadSubScript("chrome://mochitests/content/browser/devtools/client/framework/test/shared-head.js", this);
+var { Toolbox } = require("devtools/client/framework/toolbox");
+const EXAMPLE_URL = "http://example.com/browser/devtools/client/debugger/new/test/mochitest/";
+
+// Wait until an action of `type` is dispatched. This is different
+// then `_afterDispatchDone` because it doesn't wait for async actions
+// to be done/errored. Use this if you want to listen for the "start"
+// action of an async operation (somewhat rare).
+function waitForNextDispatch(store, type) {
+  return new Promise(resolve => {
+    store.dispatch({
+      // Normally we would use `services.WAIT_UNTIL`, but use the
+      // internal name here so tests aren't forced to always pass it
+      // in
+      type: "@@service/waitUntil",
+      predicate: action => action.type === type,
+      run: (dispatch, getState, action) => {
+        resolve(action);
+      }
+    });
+  });
+}
+
+// Wait until an action of `type` is dispatched. If it's part of an
+// async operation, wait until the `status` field is "done" or "error"
+function _afterDispatchDone(store, type) {
+  return new Promise(resolve => {
+    store.dispatch({
+      // Normally we would use `services.WAIT_UNTIL`, but use the
+      // internal name here so tests aren't forced to always pass it
+      // in
+      type: "@@service/waitUntil",
+      predicate: action => {
+        if (action.type === type) {
+          return action.status ?
+            (action.status === "done" || action.status === "error") :
+            true;
+        }
+      },
+      run: (dispatch, getState, action) => {
+        resolve(action);
+      }
+    });
+  });
+}
+
+function waitForDispatch(dbg, type, eventRepeat = 1) {
+  let count = 0;
+
+  return Task.spawn(function* () {
+    info("Waiting for " + type + " to dispatch " + eventRepeat + " time(s)");
+    while (count < eventRepeat) {
+      yield _afterDispatchDone(dbg.store, type);
+      count++;
+      info(type + " dispatched " + count + " time(s)");
+    }
+  });
+}
+
+function waitForThreadEvents(dbg, eventName) {
+  info("Waiting for thread event '" + eventName + "' to fire.");
+  const thread = dbg.toolbox.threadClient;
+
+  return new Promise(function(resolve, reject) {
+    thread.addListener(eventName, function onEvent(eventName, ...args) {
+      info("Thread event '" + eventName + "' fired.");
+      thread.removeListener(eventName, onEvent);
+      resolve.apply(resolve, args);
+    });
+  });
+}
+
+function waitForState(dbg, predicate) {
+  return new Promise(resolve => {
+    const unsubscribe = dbg.store.subscribe(() => {
+      if (predicate(dbg.store.getState())) {
+        unsubscribe();
+        resolve();
+      }
+    });
+  });
+}
+
+const waitForPaused = Task.async(function* (dbg) {
+  // We want to make sure that we get both a real paused event and
+  // that the state is fully populated. The client may do some more
+  // work (call other client methods) before populating the state.
+  return Promise.all([
+    yield waitForThreadEvents(dbg, "paused"),
+    yield waitForState(dbg, state => dbg.selectors.getPause(state))
+  ]);
+});
+
+const initDebugger = Task.async(function* (url, ...sources) {
+  const toolbox = yield openNewTabAndToolbox(url, "jsdebugger");
+  const win = toolbox.getPanel("jsdebugger").panelWin;
+  const store = win.Debugger.store;
+  const { getSources } = win.Debugger.selectors;
+
+  const dbg = {
+    actions: win.Debugger.actions,
+    selectors: win.Debugger.selectors,
+    getState: store.getState,
+    store: store,
+    toolbox: toolbox,
+    win: win
+  };
+
+  yield Promise.all(sources.map(url => {
+    return waitForState(dbg, state => {
+      return getSources(state).some(s => s.get("url").includes(url));
+    });
+  }));
+
+  return dbg;
+});
+
+function findSource(dbg, url) {
+  const sources = dbg.selectors.getSources(dbg.getState());
+  const source = sources.find(s => s.get("url").includes(url));
+
+  if(!source) {
+    throw new Error("Unable to find source: " + matchingStr);
+  }
+
+  return source.toJS();
+}
+
+function selectSource(dbg, url) {
+  info("Selecting source: " + url);
+  const source = findSource(dbg, url);
+  dbg.actions.selectSource(source.id);
+
+  return waitForDispatch(dbg, "LOAD_SOURCE_TEXT");
+}
+
+function stepOver(dbg) {
+  info("Stepping over");
+  dbg.actions.stepOver();
+  return waitForPaused(dbg);
+}
+
+function stepIn(dbg) {
+  info("Stepping in");
+  dbg.actions.stepIn();
+  return waitForPaused(dbg);
+}
+
+function stepOut(dbg) {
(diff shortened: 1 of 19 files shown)


jasonLaster:
Yay! Will review tomorrow morning

public/js/test/mochitest/mochitest/doc_event-listeners-02.html
```diff
@@ -0,0 +1,53 @@
+<!-- Any copyright is dedicated to the Public Domain.
```
jasonLaster: looks like this file is duplicated

public/js/test/mochitest/mochitest/browser_first_test.js
```diff
@@ -0,0 +1,18 @@
...
+  let toolbox = yield openNewTabAndToolbox(TAB_URL, "jsdebugger");
+  let threadClient = toolbox.threadClient;
+
+  ContentTask.spawn(gBrowser.selectedBrowser, null, function* () {
+    content.document.querySelector("button").click()
+  });
+
+  yield waitForThreadEvents(threadClient, "paused");
+  yield toolbox.destroy().then(finish);
+});
```
jasonLaster: looks like this test is also duplicated.

did you any chance make a second mochitest dir :)

webpack.config.devtools.js
```diff
@@ -28,9 +28,7 @@ module.exports = webpackConfig => {
         return;
       }
       callback();
-    },
-
-    { codemirror: "var devtoolsRequire('devtools/client/sourceeditor/editor')" }
```
jasonLaster: what does this do?

jasonLaster:
looks like the mochitests failed on CI. Also, looks like we need to do some `eslintignore` stuff

webpack.config.devtools.js
jlongster: It was leftover from previous codemirror work; we don't need to map it because we are mapping the `source-editor` file itself which includes codemirror. I noticed it here but I can separate it out if you want.

public/js/test/mochitest/mochitest/browser_first_test.js
jlongster: Yep. rsync went askew, I'll fix!

jlongster:
Yeah, definitely need to figure out eslint. The mochitests are failing because bug 1294686 hasn't landed yet. It doesn't recognize that path as a mochitest directory.

Thanks! I'll fix all this up.

jasonLaster:
It might help to start with some docs on this workflow: 
+ `npm run mochitest` does on the first run and subsequent runs.
+ how to update docker.
+ where the tests + helpers live 
+ what is mochitest - link to docs
+ how to use `--jsdebugger`
+ logging with `info`
+ ...


It might also help to add some comments on the different scripts so it'll be easy for our future selves to update.

jlongster:
We'll add docs over time. I can add some comments in the scripts though. For now I'll add a short section somewhere and link to https://developer.mozilla.org/en-US/docs/Mozilla/Projects/Mochitest. We shouldn't recreate that page, and we'll probably keep evolving our specific workflow as we use it for the next week or so.

jlongster:
Getting there. I have a working mochitest just need to make sure that mozilla-central has the necessary changes (just pushed last night) and make sure our workflow isn't horrible. I've already improved several things so that you can edit the mochitests locally and firefox will automatically use them. Should be able to start writing tests later today, and we'll probably add more utility functions to `head.js` as we find pain points.

jlongster:
Unfortunately I'm stuck on bug 1294686 to be merged to m-c. It's been pushed to fx-team but for some reason it's taking a while. I'll work on docs though and hope to ready later today.

jasonLaster:
Sounds good. I look forward to writing some tests! 

Here's the [head.js](https://github.com/jasonLaster/debugger.html/blob/2ef93211c9ea78ca5645efccc8a60e224bf64fbe/public/js/test/integration-mocha/head.js) I was working off of before. pretty happy with the API. e.g. addBreakpoint assumes you're adding a breakpoint to the selected source, but you can pass in an optional source if you want to.

jlongster:
I like a lot of the stuff in your head.js. I was trying to use actions more directly, but I think you're right that we need to wrap things like step/stepIn/etc because we want to wait on the "pause" event, and users can still use actions directly if they want.

I'm not sure about automatically assuming the selected source though; I like to be more explicit. I'll have a test or two finished when merging this and we can talk more about it then; the API is still pretty simple but you do have to pass in a few more parameters which I think makes it clear what's going on.

jlongster:
So theoretically this is ready to go. Just waiting on a new docker image to build.

jasonLaster:
👍

jlongster:
Oops, there were a few changes to the command action and main.js which I'm going to remove.

public/js/test/mochitest/head.js
```diff
@@ -0,0 +1,176 @@
...
+}
+
+function resume(dbg) {
+  info("Resuming");
+  dbg.actions.resume();
+  return waitForThreadEvents(dbg, "resumed");
+}
+
+function addBreakpoint(dbg, sourceId, line, col) {
+  return dbg.actions.addBreakpoint({ sourceId, line, col });
```
jasonLaster: I think it is good to keep the API similar to what the components call, that way we exercise the same behavior. We can't guarantee it's the same, but it's probably good to try.

The `addBreakpoint` command I setup assumes it's the selected source and passes along the `getTextForLine` param for that reason. 

```js
function addBreakpoint(line, { source } = {}) {
  source = source || selectors.getSelectedSource();
  return actions.addBreakpoint({
    sourceId: source.get("id"),
    line
  }, {
    getTextForLine: l => debuggerWindow.cm.getLine(l - 1).trim()
  });
}
```

If we want to pass along a source that is not the selected source, I'd propose that this action first selects that source and then adds the breakpoint. This will keep our tests behaving the same way our app does and keep us from getting into weird states.

```js
function addBreakpoint(line, { source } = {}) {
  if (source.get("url") != selectors.getSelectedSource().get("url")) {
    selectSource(source);
  }

  source = source || selectors.getSelectedSource();
  return actions.addBreakpoint({
    sourceId: source.get("id"),
    line
  }, {
    getTextForLine: l => debuggerWindow.cm.getLine(l - 1).trim()
  });
}
```

One anti-goal is pausing in a test and the debugger looks broken. An example of this is the breakpoint list not having text and not seeing the breakpoint which was just added.

public/js/test/mochitest/head.js
```diff
@@ -0,0 +1,176 @@
...
+    store: store,
+    toolbox: toolbox,
+    win: win
+  };
+
+  yield Promise.all(sources.map(url => {
+    return waitForState(dbg, state => {
+      return getSources(state).some(s => s.get("url").includes(url));
+    });
+  }));
```
jasonLaster: not actionable now, but i would love it if the debugger server passed a `loaded` event after the sources are in and it's ready. Maybe it's hard to know...

jlongster: It's very hard to know. You don't know if a source is dynamically loaded in (which we should absolutely test) and if it happens a little bit later, then when is it "done"? The fact is sources can come in at any time so there is not such thing as "done". Better to embrace this and be explicit.

I can tell you this with a lot of experience with the old debugger that this is a good way to go. We've tried several things before (including having a "sources loaded" event) and it was *always* a pain.

public/js/test/mochitest/head.js
jlongster: (Wanted to reply longer here and had to step out)

> I think it is good to keep the API similar to what the components call, that way we exercise the same behavior. We can't guarantee it's the same, but it's probably good to try.

Those actions are what they call :) But we can't guarantee that we are going to call the actions with exactly the same arguments, but I think that's ok. Our actions should be usable like this. They should just take simple arguments and work.

The fact that the `addBreakpoint` action requires that extra function is a bad thing (I know I implemented it). We need to fix that. But for the majority of the tests, it really doesn't matter that the text snippet isn't there. I'd like to write it this way, and just try improve the `addBreakpoint` so it has the text correctly without requiring this extra work. For now, we'll just make sure we add the line snippet when testing the breakpoint pane. It's a small detail that doesn't really matter, and it's just a bad API that we need to fix.

> The addBreakpoint command I setup assumes it's the selected source and passes along the getTextForLine param for that reason.

The selected source thing is a bigger question. I completely understand where you are coming from, but let me try to explain why I don't think we should do it.

We don't drive the UI by actually clicking on the DOM in every test for a reason. For any given test, they are really trying to make sure a specific workflow is correct (and ignore all the other details). Redux actions are a super nice way to do this, even if it's not literally the same workflow that the UI goes through. As long as we make sure adding a breakpoint on the current selected source works, it shouldn't matter if adding a breakpoint is on the selected source or not in other tests. If we really wanted to test the UI as the user uses it every time, we'd be clicking DOM elements.

I actually really like the idea of being able to set breakpoints on other sources than the selected one, and I want to do that some day. Something like a "command bar" that lets you say "break <url> <line>". That would be awesome. There's no reason we should tie together a simple action like that with changing out the entire editor, loading source texts, etc.

In the old debugger, there are many tests that set breakpoints in several sources, and then continue through and make sure all of them are hit. These tests are a lot simpler if you can just set a breakpoint on any source. It makes it easier to test complex scenarios like "set a breakpoint on source X and make sure when the breakpoint it hit that it opens and selects that source".

So that's my vision of the tests: generally the commands map straight to redux actions, but with a few niceties like the stepping ones waiting for the right `paused` event. This lets tests set up all kinds of workflows, whatever they want, even if many of them aren't real user workflows. We can make sure that things work in weird ways without the commands getting in the way of a specific workflow.

Just wanted to explain my view of things; I totally get where you're coming from. The good thing is this is something that we could pretty easily change later just by changing the `addBreakpoint` command if we are missing bugs for some reason.

jlongster:
The first test passed! The failure is cypress. Going ahead and merging this because Jason approved this generally. I know there are a few conceptual differences, but thanks for letting me flesh this out. I think this will be good for long-term maintenance, but adaptable if we find the current APIs problematic.

Note that we are able to run our mochitests in ~7min, including the entire time for checking out docker. I think that's pretty good. Running all the mochitests will end up taking several minutes once we write them, but even if it's ~15min that's a whole lot better than before (waiting on try for hours).

jasonLaster:
Great job!

Thanks for doing the hard work so that we can test the debugger in the
panel, while developing out of the tree. Thats a big accomplishment.

By the way, I bet we can get the seven minutes reduced. Also we can
parallelize fairly easily on circle as well.

Also, thanks for explaining your thinking on the commands. I can see where
you're coming from as well.

jlongster:
Thanks! I'm pretty happy with how it worked out; I was worried we'd have to wait on a build and wait >30 min or something. The 7min is just checkout out the docker image unfortunately, so not sure if that can be reduced... but we could definitely run the tests in parallel (split them up into 2-3 "chunks" and run them on different containers). That's actually what try does (that's why there's dt1, dt2, dt3, etc)

I missed a bunch of linting errors which I just fixed (pushed straight to master).

jasonLaster:
Yea, i wonder if the Firefox dir can be added to the circle artifacts list.
The should help with the checkout.

Issue #451, opened 2016-08-01
Replace `utils/path.js` with a proper path library


jlongster:
We hacked together `path.js` as a temporary module, but we should use a library that provides real joining and other operations on URLs. I just added a few methods to resolve source map URLs, and it's getting hacky. Someone should find a nice (small) library for operating on URLs (basename, dirname, join, etc).

hallvors:
Here's Node's Path documentation: https://nodejs.org/api/path.html
https://github.com/nodejs/node/blob/master/lib/path.js

Not sure if that's reusable. I haven't found anything else so far.

jasonLaster:
@jlongster do you think we'll be able to take this on this sprint or next?

wldcordeiro:
Someone has published the `Node` `path` lib to npm, but they haven't updated in some time (looks like the lib from 0.12.7 from the last commit) and some limited searching isn't giving me a ton of results either.

jasonLaster:
Interesting. I wonder if the node library would work in the browser? Does
it have any dependencies?

wldcordeiro:
@jasonLaster looks like it depends on `process` and `util` https://github.com/jinder/path/blob/master/package.json#L21

jbhoosreddy:
maybe something like [path-browserify](https://www.npmjs.com/package/path-browserify) is useful.

wldcordeiro:
Dependency tree of the path package I posted
- path
  - util
    - inherits
  - process

wldcordeiro:
@jbhoosreddy looks like that is relatively similar to what I posted only further out of date.

jbhoosreddy:
@wldcordeiro I posted it as a suggestion because it doesn't have any dependencies (except tape for tests) and seems to have methods which are talked about in the issue like (basename, dirname, join, etc.). I see it's not actively maintained, but I imagine we could make modifications as per our requirements if needed.

jbhoosreddy:
Reading the code for all the libraries mentioned here and those I found on the internet, I think the way to do this would be to keep the existing `utils/path.js` which maps to particular methods in the node `path` library. We can further use `Services` to identify which operating system we are working on and separate exports for win32 and posix explicitly.
It seems they use `process` to do things like identify OS type and current working directory (if no arguments are given), and we can prevent using that code since we have that information.

jbhoosreddy:
I can take this issue, if no one has. I'm interested to see if I get somewhere with this approach (in a usable way). 😅

[...]

PR #631, opened 2016-08-31
added module which wraps nodeJS path module (#451)

jbhoosreddy:
So I managed to implement what I had in mind.

The problem with other path libraries was that they tried to port the core NodeJS path library but they were not actively maintained.

In this module, it wraps around the nodeJS core library itself, so any changes to that would propagate to our project also.

Another plus is that it ports all of NodeJS path module except resolve and relative, which utilize process module and cannot be ported. (Also trying to port them won't make total sense for the browser.)

@jasonLaster as per your suggestion, I made an npm package for this. It's funny because I spent quite some time on this PR which doesn't really change anything. But I learned a lot of different things. Thanks!

Let me know if you find any issues. 😜

### DIFF
--- a/public/js/utils/path.js
+++ b/public/js/utils/path.js
@@ -1,20 +1,2 @@
-function basename(path) {
-  return path.split("/").pop();
-}
-
-function dirname(path) {
-  const idx = path.lastIndexOf("/");
-  return path.slice(0, idx);
-}
-
-function isURL(str) {
-  return str.indexOf("://") !== -1;
-}
-
-function isAbsolute(str) {
-  return str[0] === "/";
-}
-
-module.exports = {
-  basename, dirname, isURL, isAbsolute
-};
+const path = require("path-browser-wrapper");
+module.exports = path;
--- a/public/js/utils/path.js
+++ b/public/js/utils/path.js
@@ -1,2 +1,0 @@
-const path = require("path-browser-wrapper");
-module.exports = path;
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
     "immutable": "^3.7.6",
     "invariant": "^2.2.1",
     "lodash": "^4.13.1",
-    "path-browser-wrapper": "^1.0.0",
+    "path-browser-wrapper": "^1.0.1",
     "pretty-fast": "^0.2.0",
     "react": "=0.14.7",
     "react-dom": "=0.14.7",
--- a/package.json
+++ b/package.json
@@ -35,6 +35,7 @@
     "immutable": "^3.7.6",
     "invariant": "^2.2.1",
     "lodash": "^4.13.1",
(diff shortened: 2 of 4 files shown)


public/js/utils/path.js
```diff
@@ -1,20 +1,2 @@
...
-
-function isAbsolute(str) {
-  return str[0] === "/";
-}
-
-module.exports = {
-  basename, dirname, isURL, isAbsolute
-};
+const path = require("path-browser-wrapper");
+module.exports = path;
```
jasonLaster: I'd switch this and use the library directly where it is called.

jbhoosreddy: Okay. I'll do that then.

jasonLaster:
One nit. Looks good though!

And, you now have a star on the project

jbhoosreddy:
modified the PR to fix the nit.
1. removed utils/path.
2. updated references

jlongster:
Sorry, but I don't think this is a good idea. My first thought was that we should just inline this and not have it be a separate module, but I have other concerns about the implementation itself:

1. It feels weird to depend on node directly like this.
2. This pulls in the entire path implementation from node which is a whopping 44K.
3. We don't need an entire `path` implementation. We are only ever dealing with URLs, so it will *always* be a forward slash. A ton of the implementation of `path` is dealing with cross-platforms paths. We don't need that.
4. Node's `path` methods don't always work on URLs. There are unexpected failures:

```js
> const path = require("path")
undefined
> path.join("http://foo/bar", "baz")
'http:/foo/bar/baz'
```

For those reasons, we shouldn't use node's path module. I like what we have now; it's simple and it works well. I rarely run into problems with what we throw at it.

I like the simplicity of our current implementation. I was t

jlongster:
(I missed the discussion in the issue because I was on PTO/conference, sorry)

jasonLaster:
Hmm, that's a good point. I didn't see that we were pulling in 44K.

jlongster:
Thank you for putting the work in to this @jbhoosreddy, but sorry we can't accept it :( I would have mentioned this in the issue if I had seen it. I hate closing PRs after someone has put the work into them but this will break various code (see `join` above) and increase the code size a lot, and the current code works ok. If there is another solution, please open a new issue (I'll close this PR and the current issue)

jbhoosreddy:
it's quite alright.. as @jasonLaster knows, I took that issue mostly for experimenting with some stuff. and not surprisingly I learned a lot of things in that process. I'll make a new PR for another possible solution. 👍

jlongster:
@jbhoosreddy I'm not entirely sure we need to replace this anymore :D Unless you know of a URL-based path library? I couldn't find one anywhere, and it really is a lot simpler since we can assume forward slashes. If you figure out a way that we don't have to maintain this though, that would be great.

PR #632, opened 2016-08-31
(WIP) Redux driven tests

jasonLaster:
I spent yesterday afternoon working on developing a redux oriented alternative to our cypress integration tests.

Goals:
+ maintain async command api. The tests should be a list of debugger commands
+ maintain local development debugging workflow
   + refresh or hmr to see updates
   + debug debugger with devtools
+ be interop with mochitests. 
  + should be possible to test the debugger in the panel
  + I'd like to give contributors the option to run the tests locally or in firefox. 
+ target agnostic - our focus now is on testing the panel, but we will eventually want to be able to connect to chrome, node, etc. 

Here are the two tests that I have so far. They serve as an example of what the tests would look like. I would also like to add some assertions soon as well. Assertions can be DOM as well as state level.

One nice win with this approach is that they seem fairly robust. We won't know until we run them on CI, but when I ran the pausing test 100 times it passed every time. 

There are some implementation details that I did not focus on because I wanted a POC. If we decide that we want to invest in this approach they can be polished.

```js
describe("tests", function() {
  it("debugger pausing", async function(done) {
    await waitForTime(2000);
    await loadDebugger();
    await navigate("todomvc");
    await selectSource("todo-view");
    await addBreakpoint(33);
    await addBreakpoint(35);
    await addTodo();
    await stepIn();
    await stepOver();
    await stepOut();
    await resume();
    await waitForPaused();
    await resume();
    await removeBreakpoint(33);
    await removeBreakpoint(35);
    expect(2).to.equal(2);
    done();
  });

  it("pretty printing", async function(done) {
    await waitForTime(2000);
    await loadDebugger();

    await navigate("todomvc");
    await selectSource("localStorage");
    await prettyPrint();
    done();
  });
});
```


jasonLaster:
Closing as we'll focus on merging work into the `head.js` that @jlongster is setting up.

wldcordeiro:
That's so awesome though! Nice work @jasonLaster you got even more progress from what you showed me. 😄

jlongster:
We'll look into this in the future, but we have 1.5 weeks to write all of our mochitests and need to be sure that the debugger in the panel is actually working correctly. We're shifting to focus 100% on mochitests for the next few days.

PR #651, opened 2016-08-31
Export a bootstrap function

jasonLaster:
This is the second half the `fix-console` work.

We can land the m-c part separately because it doesn't break the old debugger, but this will have to land in the bundle so that the new debugger works.

### DIFF
--- a/public/js/main.js
+++ b/public/js/main.js
@@ -22,6 +22,7 @@ const { getClient, connectClients, startDebugging } = require("./clients");
 const firefox = require("./clients/firefox");
 const configureStore = require("./utils/create-store");
 const reducers = require("./reducers");
+const selectors = require("./selectors");
 
 const Tabs = require("./components/Tabs");
 const App = require("./components/App");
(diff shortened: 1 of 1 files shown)


public/js/main.js
```diff
@@ -92,12 +93,17 @@ if (connTarget) {
   });
 } else if (isFirefoxPanel()) {
   // The toolbox already provides the tab to debug.
+  function bootstrap({ threadClient, tabTarget }, callback) {
+    firefox.setThreadClient(threadClient);
+    firefox.setTabTarget(tabTarget);
+    firefox.initPage(actions);
+    renderRoot(App);
+
+    return { store, actions, selectors };
```
jasonLaster: i changed this version slightly to pass `store, actions, selectors` instead of `getState` and `selectors`

public/js/main.js
```diff
@@ -92,13 +93,14 @@ if (connTarget) {
...
 } else if (isFirefoxPanel()) {
   // The toolbox already provides the tab to debug.
-  module.exports = {
-    setThreadClient: firefox.setThreadClient,
-    setTabTarget: firefox.setTabTarget,
-    initPage: firefox.initPage,
-    getActions: () => actions,
-    renderApp: () => renderRoot(App)
-  };
+  function bootstrap({ threadClient, tabTarget }, callback) {
```
jlongster: I don't think we need the callback argument?

jlongster:
Looks perfect, this will work nicely with testing too because we can just do `win.Debugger.actions` etc to access the redux stuff. Remove the callback (unless you need it?) and we'll merge this and then push a new m-c update with the test. We need this update before we can write mochitests because it depends on how the selectors/actions/etc are exposed.

public/js/main.js
jasonLaster: hmm, i don’t see this on master.

Are you sure this is not an old PR?

PR #654, opened 2016-09-02
Fix #653, finesse selected tab designs

### LINKED ISSUE #653: Tabs - selected state
The selected state for tabs is currently different than what was mocked. We should clean this up before we launch.

helenvholmes:
Naming things in ```variables.css``` is hard and I dislike it

### DIFF
--- a/public/js/components/SourceTabs.css
+++ b/public/js/components/SourceTabs.css
@@ -10,17 +10,19 @@
 }
 
 .source-tab {
-  border: 1px solid var(--theme-gray);
+  background-color: var(--theme-toolbar-background-alt);
+  color: var(--theme-faded-tab-color);
+  border: 1px solid var(--theme-splitter-color);
   border-top-left-radius: 2px;
   border-top-right-radius: 2px;
-  padding: 2px 20px 2px 10px;
+  padding: 2px 20px 2px 12px;
   height: 23px;
   line-height: 20px;
-  color: var(--theme-comment);
-  margin: 6px 0 0 10px;
+  margin: 6px 0 0 8px;
   display: inline-block;
   border-bottom: none;
   position: relative;
+  transition: all 0.25s ease;
 }
 
 .source-tab:hover {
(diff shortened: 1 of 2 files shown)


public/js/components/SourceTabs.css
```diff
@@ -10,17 +10,19 @@
 }
 
 .source-tab {
-  border: 1px solid var(--theme-gray);
+  background-color: var(--theme-toolbar-background-alt);
+  color: var(--theme-faded-tab-background);
```
jasonLaster: should this be `-theme-faded-tab-color`?

public/js/lib/themes/variables.css
```diff
@@ -20,7 +20,9 @@
   --theme-contrast-background: #e6b064;
 
   --theme-tab-toolbar-background: #fcfcfc;
+  --theme-faded-tab-background: #7E7E7E;
   --theme-toolbar-background: #fcfcfc;
+  --theme-toolbar-background-alt: #f5f5f5;
```
jasonLaster: would it be possible to commit these changes to gecko as well so the changes are consistent.

jasonLaster:
@helenvholmes i agree, CSS variable names are hard...

I'd like to revisit some of our CSS strategy once we land in firefox. I think now that we can safely plan on being an independent project in github, we can discuss some CSS strategies for the debugger as long as it would make sense for other devtool tools in the future as well.

public/js/components/SourceTabs.css
helenvholmes: Meaning as a variable name?

jasonLaster:
👍

wldcordeiro:
@jasonLaster I'll throw an oddball suggestion into the mix. https://github.com/cssinjs/jss 😄

jlongster:
We've talked about a CSS in JS solution before but I want more buy-in from the team first, especially @helenvholmes. I think there are a lot of benefits to it but I want others to know how to work with it too.

jlongster:
The linting errors were my fault. I'll merge this in as it seems good to me too!

Issue #655, opened 2016-09-02
prepare-mochitests-dev


clarkbw:
I was going through the [mochitests docs](https://github.com/devtools-html/debugger.html/blob/master/docs/mochitests.md) and step one seems to fail for me.  Will keep looking into it further but wanted to post this here in case @jlongster had ideas.

```bash
clarkbw: ~/src/mozilla/debugger.html $ ./bin/prepare-mochitests-dev
WARNING: local changes detected on mozilla-central
./bin/make-firefox-bundle: line 29: webpack: command not found
usage: cp [-R [-H | -L | -P]] [-fi | -n] [-apvX] source_file target_file
       cp [-R [-H | -L | -P]] [-fi | -n] [-apvX] source_file ... target_directory
rm: ./bin/../firefox/devtools/client/debugger/new/test/mochitest: No such file or directory
ln: ./bin/../firefox/devtools/client/debugger/new/test/mochitest: No such file or directory
touch: ./bin/../firefox/devtools/client/debugger/new/test/mochitest/browser.ini: No such file or directory
./bin/prepare-mochitests-dev: line 14: ./mach: No such file or directory
```

clarkbw:
Embarrassing  😲 ☕ 
I didn't have mercurial installed.  Will create a PR that checks for it in the download command.  I'm guessing there will be others who don't have `hg` installed either. 😄

jlongster:
Good call! Yeah, those scripts can be improved. I can add some more sanity checks to them.

jlongster:
Done in #656

PR #656, opened 2016-09-02
check for hg command and successful exit of it

clarkbw:
* `download-firefox-artifact` 
 * checks for the `hg` command using `command -v` which I believe is the recommended style.  Offers the homebrew command to install mercurial.
 * checks that `hg clone` exits successfully for any reason, otherwise the `cd $FIREFOX_PATH` fails and the script continues to write a mozconfig file
* `prepare-mochitests-dev`
 * checks that the download command exited successfully before continuing 

My bash scripting is really rusty :shipit:

### DIFF
--- a/bin/download-firefox-artifact
+++ b/bin/download-firefox-artifact
@@ -3,12 +3,23 @@
 ROOT=`dirname $0`
 FIREFOX_PATH="$ROOT/../firefox"
 
+# check that mercurial is installed
+command -v hg >/dev/null 2>&1 || {
+  echo >&2 "mercurial is required for mochitests, use 'brew install mercurial' on MacOS";
+  exit 1;
+}
+
 if [ -d "$FIREFOX_PATH" ]; then
     # If we already have Firefox locally, just update it
     cd "$FIREFOX_PATH";
     hg pull
 else
+    echo "Downloading Firefox source code, requires about 30 min depending on connection"
     hg clone https://hg.mozilla.org/mozilla-central/ "$FIREFOX_PATH"
+    # if somebody cancels (ctrl-c) out of the long download don't continue
+    if [ $? -ne 0 ]; then
+        exit 1;
+    fi
     cd "$FIREFOX_PATH"
 
     # Make an artifcat build so it builds much faster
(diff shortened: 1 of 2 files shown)


clarkbw:
This fixes #655 for me

bin/download-firefox-artifact
```diff
@@ -3,12 +3,23 @@
 ROOT=`dirname $0`
 FIREFOX_PATH="$ROOT/../firefox"
 
+# check that mercurial is installed
+command -v hg >/dev/null 2>&1 || {
+  echo >&2 "mercurial is required for mochitests, use 'brew install mercurial' on MacOS";
+  exit 1;
+}
```
jlongster: I think this is a little more readable:

```
if [ -z "`command -v hg`" ]; then
  echo "hg not found yo";
fi
```

bin/download-firefox-artifact
```diff
@@ -3,12 +3,23 @@
...
+
 if [ -d "$FIREFOX_PATH" ]; then
     # If we already have Firefox locally, just update it
     cd "$FIREFOX_PATH";
     hg pull
 else
+    echo "Downloading Firefox source code, requires about 30 min depending on connection"
```
jlongster: Not sure about putting a specific time, it takes about 10 min for me? What about saying "10-30min depending on connection"

jlongster:
Awesome, great improvements, thanks! Just one or two notes.

bin/download-firefox-artifact
clarkbw: 👍

bin/download-firefox-artifact
clarkbw: Definitely better

clarkbw:
Ok, I'll update with the suggestions

jlongster:
The downside to using shell is that it doesn't work on Windows :( We might want to redo this in JS later, but it's so much easier in shell.

clarkbw:
> doesn't work on Windows

We could require the new Ubuntu Windows userspace integration, which would make shell scripting work.  I hope the artifact builds work the same on Windows.

jlongster:
Thanks!

PR #657, opened 2016-09-02
Always render editor so only 1 CodeMirror instance is created

jlongster:
I made these changes a little quickly, so I'm open to a better way. This always renders the editor in the background, and renders everything else "above" it. The nice thing (I think) is that now the source header (and eventually the footer) is always rendered which I think makes it look more consistent.

This is required so that we only ever create 1 CodeMirror instance, and it makes our `Editor` component more reliable because on the initial render the `this.editor` instance is null (it hasn't been mounted yet). Not all renders are "updates" so `this.editor` will always reliably be there, so breakpoints will always show up, etc.

### DIFF
--- a/public/js/components/Editor.js
+++ b/public/js/components/Editor.js
@@ -66,9 +66,12 @@ const Editor = React.createClass({
     );
 
     this.editor.codeMirror.on("gutterClick", this.onGutterClick);
-    this.setText(this.props.sourceText.get("text"));
     resizeBreakpointGutter(this.editor.codeMirror);
     debugGlobal("cm", this.editor.codeMirror);
+
+    if (this.props.sourceText) {
+      this.setText(this.props.sourceText.get("text"));
+    }
   },
 
   onGutterClick(cm, line, gutter, ev) {
--- a/public/js/components/App.css
+++ b/public/js/components/App.css
@@ -72,19 +72,15 @@ body {
   transition: opacity 200ms;
 }
 
-.welcomebox {
-  margin: 50px auto;
-  padding: 20px;
-  font-size: 1.25em;
-  color: var(--theme-gray-darker);
-  font-weight: lighter;
-  text-align: center;
-  width: 100%;
-}
-
 .search-container {
+  position: absolute;
+  top: 0;
+  left: 0;
+  width: 100%;
+  height: 100%;
   display: flex;
-  flex: 1;
+  z-index: 200;
+  background-color: rgba(221, 225, 228, 0.66);
 }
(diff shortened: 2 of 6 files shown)


jlongster:
Now it looks like this:

screen shot 2016-07-25 at 9 13 20 am
screen shot 2016-07-26 at 1 39 44 pm
screen shot 2016-07-26 at 1 40 15 pm

jlongster:
I went ahead and fixed those icons. Converting a few ems to pxs solves it and makes this less dependent on font size (the padding/margin is in px anyway).

public/js/components/Editor.js
```diff
@@ -160,20 +164,29 @@ const Editor = React.createClass({
...
+        dom.div({ className: "editor-mount devtools-monospace" }),
         breakpoints.map(bp => {
           return Breakpoint({
             key: makeLocationId(bp.location),
             breakpoint: bp,
             editor: this.editor && this.editor.codeMirror
           });
-        })
+        }),
+        !this.props.selectedSource ? this.renderWelcomeBox() : null
```
jasonLaster: Why did the welcome box have to move into the editor?
Couldn't it still be rendered next to the search box?

jasonLaster:
Left one comment, but looks good. Feel free to merge.

Initially I was in favor of search being a popup, but now that i'm used to it replacing the editor the popup feels noisy. I might feel differently as i start using the debugger more often and find the popup is more contextual.

public/js/components/Editor.js
jlongster: I'll try that, I can't remember but it seemed more natural. I think I like the header always showing because it feels more consistent, but I don't need to change that right now. I'll see if it's just as easy to render it in `App`.

jasonLaster: Thanks, I'd be happy to add an empty center header later for consistency.

My concern is that this introduces bloat to the Editor. I'd like to keep the keep the editor focused on the source.

jlongster:
Yeah, it works fine to render the welcome box in `App`. I think it'd be nice to always render the header so that top row is always consistent, and would help with #623 because you could put the icons in there. But that's a different discussion.

jasonLaster:
@jlongster yeah - that's a great point. Lets create a new issue for the center pane header

public/js/components/Editor.js
jlongster: Yep, it'd be better as another component

jlongster:
Just fixed conflicts, it was passing on CI before.

PR #659, opened 2016-09-02
Make editor handle edge cases better: only render breakpoints when text isn't loading

jlongster:
A few more code paths came up while testing. The breakpoint won't show if a source text hasn't been loaded yet, because it'll try to added the breakpoint icons to the editor while showing "Loading...", and then it won't re-add them when the real text appears. This is because we've optimized the breakpoints so that they only touch the editor when needed (they don't always re-add themselves).

Also if you close all the tabs and open them the `isSourceForFrame` errors of paused, so that's fixed.


jasonLaster:
👍

PR #660, opened 2016-09-02
Write some initial editor mochitests

jlongster:
Two mochitest to make sure the editor works. If we make sure that the editor handles selected source properly, the rest of the tests can just check the redux app state.

I found a few things to improve, namely `waitForPaused` which appropriately waits for the source text to be loaded, and also `initDebugger` which checks to see if a source that we are waiting on already exists. 

I also removed the `mozilla/no-cpows-in-tests` because it's not smart enough. It warns whenever you use anything inside a `ContentTask` callback which is wrong.


jlongster:
Just added another test for making sure we are selecting the right mode. I actually needed to implement that as well, so there's a bit of code in here... but it's a small function on the editor to select the right mode.

This turned out to be a bit of a pain because I needed direct access to the editor, and it's only exposed in development mode. The way we do this normally on m-c is we have also have a `testing` flag, so I changed the `debugGlobal` to also expose it if the testing flag is true. I had to create a local module that is basically a noop, but the webpack config will make the debugger load the native `flags` module from m-c when in the panel.

public/js/components/Editor.js
```diff
@@ -141,6 +142,22 @@ const Editor = React.createClass({
...
+    } else if (contentType === "text/wasm") {
+      this.editor.setMode({ name: "wasm" });
+    } else if (sourceText.get("text").match(/^\s*</)) {
+      // Use HTML mode for files in which the first non whitespace
+      // character is `<` regardless of extension.
+      this.editor.setMode({ name: "htmlmixed" });
+    } else {
+      this.editor.setMode({ name: "text" });
+    }
+  },
```
jasonLaster: Is this something we can move to the Editor module?

public/js/utils/debug.js
```diff
@@ -1,11 +1,10 @@
 const { isDevelopment } = require("../feature");
+const flag = require("../test/test-flag");
```
jasonLaster: could we do the same thing with the `defineGlobal`? I'd prefer not introducing a new system just for this

public/js/test/mochitest/browser_dbg-editor-gutter.js
```diff
@@ -0,0 +1,66 @@
...
+add_task(function* () {
+  const dbg = yield initDebugger(TAB_URL, "simple1.js");
+  const { selectors: { getBreakpoints, getBreakpoint }, getState } = dbg;
+  const source = findSource(dbg, "simple1.js");
+
+  yield selectSource(dbg, source.url);
+
+  // Make sure that clicking the gutter creates a breakpoint icon.
+  clickGutter(dbg, 4);
+  yield waitForDispatch(dbg, "ADD_BREAKPOINT");
```
jasonLaster: this is a nit, but would waiting for the "RESUMED" dispatch be similar, but apply for both adding and removing breakpoints.

I bring this up because I like the patten of yielding for async actions. This is a small detail though.

public/js/test/mochitest/browser_dbg-editor-gutter.js
```diff
@@ -0,0 +1,66 @@
...
+  assertEditorBreakpoint(dbg, 2, false);
+  assertEditorBreakpoint(dbg, 4, true);
+
+  // Do the same sliding and make sure it works if there's already a
+  // breakpoint.
+  clickGutter(dbg, 2);
+  yield waitForDispatch(dbg, "ADD_BREAKPOINT");
+  is(getBreakpoints(getState()).size, 1, "One breakpoint exists");
+  assertEditorBreakpoint(dbg, 2, false);
+  assertEditorBreakpoint(dbg, 4, true);
```
jasonLaster: do we want to try removing a slid breakpoint below?

public/js/test/mochitest/browser_dbg-editor-select.js
```diff
@@ -0,0 +1,68 @@
...
+
+  // Step through to another file and make sure it's paused in the
+  // right place.
+  yield stepIn(dbg);
+  assertPausedLocation(dbg, simple2, 2);
+
+  // Step back out to the initial file.
+  yield stepOut(dbg);
+  yield stepOut(dbg);
+  assertPausedLocation(dbg, simple1, 5);
```
jasonLaster: does it make sense to add other assertions after each step

public/js/test/mochitest/head.js
```diff
@@ -114,21 +130,30 @@ const initDebugger = Task.async(function* (url, ...sources) {
...
-  yield Promise.all(sources.map(url => {
-    return waitForState(dbg, state => {
-      return getSources(state).some(s => s.get("url").includes(url));
-    });
-  }));
+  if(sources.length) {
+    info("Waiting on sources: " + sources.join(", "));
+    yield Promise.all(sources.map(url => {
+      function sourceExists(state) {
+        return getSources(state).some(s => s.get("url").includes(url));
```
jasonLaster: it's strange to me that you would want some and not `all`

public/js/test/mochitest/head.js
```diff
@@ -114,21 +130,30 @@ const initDebugger = Task.async(function* (url, ...sources) {
     win: win
   };
 
-  yield Promise.all(sources.map(url => {
-    return waitForState(dbg, state => {
-      return getSources(state).some(s => s.get("url").includes(url));
-    });
-  }));
+  if(sources.length) {
+    info("Waiting on sources: " + sources.join(", "));
```
jasonLaster: This might be nice as a separate helper

`return waitForSourcesToLoad(sources)`

public/js/test/mochitest/browser_dbg-editor-select.js
```diff
@@ -0,0 +1,68 @@
...
+  });
+
+  yield waitForPaused(dbg);
+  assertPausedLocation(dbg, longSrc, 66);
+
+  // Make sure that the breakpoint icon is literally visible.
+  const doc = dbg.win.document;
+  const bpLine = doc.querySelector(".CodeMirror-code > .new-breakpoint");
+  const cm = doc.querySelector(".CodeMirror");
+  ok(isVisibleWithin(cm, bpLine), "CodeMirror is scrolled to line");
```
jasonLaster: nit:
It would be nice to extract these methods into small helper functions:
`getCodeMirrorEl(dbg)` `getCodeMirrorLineEls(dbg)`. 

I'm mentioning this because it is nice to wrap dom selectors with business object functions whose api is more clear from the tests.

public/js/components/Editor.js
jlongster: The `Editor` component is the main thing that should do this work, I don't like have 2 separate modules that split the work for no great reason. In this case, yeah, we're forced to have source-editor.js for compatibility. 

You're right that we haven't ordered this correctly; the lifecycle methods should be grouped together. I'll move `componentDidMount` down and then it's utils, lifecycle, render, in that order. There's an eslint rule to enforce this that I thought got turned on in m-c...

public/js/utils/debug.js
jlongster: What do you mean? This is not a new system. This is hooking into Mozilla's existing mochitest system which sets a flag to indicate that it's in a testing environment.

public/js/test/mochitest/head.js
```diff
@@ -114,21 +130,30 @@ const initDebugger = Task.async(function* (url, ...sources) {
...
+// Actions
+
 function findSource(dbg, url) {
   const sources = dbg.selectors.getSources(dbg.getState());
   const source = sources.find(s => s.get("url").includes(url));
 
   if(!source) {
-    throw new Error("Unable to find source: " + matchingStr);
+    throw new Error("Unable to find source: " + url);
   }
```
jbhoosreddy: +1 was having issues with this when I was working with it.

public/js/test/mochitest/browser_dbg-editor-gutter.js
jlongster: You might be able to wait for a resumed event, but I'm not sure we'll always want to emit a resumed event after setting a breakpoint. From a distance, a resumed event has nothing to do with setting a breakpoint. This is a clear and descriptive: just wait for the action to resolve.

You say you like the pattern of yielding for async actions? :) This is doing that. It'll wait for the `ADD_BREAKPOINT` action to fire with the `status: "done"` property.

public/js/test/mochitest/browser_dbg-editor-gutter.js
jlongster: Yeah, great point.

jbhoosreddy:
This is a good reference for me. It gives me a little "direction" about how these tests are to be written.

public/js/test/mochitest/browser_dbg-editor-select.js
jlongster: Maybe, but I'd like to land 5-10 mochitests now and we can flesh them out/add more as we get a good feeling for it. I'd like to do it breadth-first, because writing several different kinds of mochitests will give us a feeling for what should be actually shared across them, and how we should structure them.

jlongster: You're probably right that I should also check the highlighted line though, I forgot about that. I will add it.

public/js/test/mochitest/head.js
jlongster: That would never pass. We don't want all the current sources to match this specific url :) We're basically saying "make sure at least one source has this URL".

public/js/test/mochitest/head.js
jlongster: Sure, I can do that

jlongster:
We're still figuring it out too! :) My goal is to be as clear and descriptive as possible though. I've had to deal with our old mochitests for too long.

public/js/utils/debug.js
jasonLaster: > see webpack.config.devtools.js changes

ahh, I see what you mean. 

I think you could also add an `isTesting` method in `feature` alongside `isDevelopment`. 

I trust your judgement though.

public/js/test/mochitest/browser_dbg-editor-select.js
jlongster: What do you mean by "business object functions"?

public/js/test/mochitest/browser_dbg-editor-gutter.js
jasonLaster: >  sure we'll always want to emit a resumed event after setting a breakpoint

that's a good point.

public/js/components/Editor.js
jasonLaster: Now that I think a bit about it more, the util methods before were mostly around codemirror functionality like alignLine middle and less around reading state like source text. 

I'm sensitive to letting a component do to much and Editor, but this makes sense.

public/js/utils/debug.js
jlongster: True, I can add an `isTesting` check, that would be more uniform.

public/js/test/mochitest/head.js
jasonLaster: oops, was looking above and saw the `...sources`

jlongster:
Forgot about this, but this requires the `devtoolsRequire` function to be available in the panel context which hasn't been committed to m-c yet :( I'm finally doing this in [bug 1300866](https://bugzilla.mozilla.org/show_bug.cgi?id=1300866). I'll have to update the docker image too. So we might not be able to merge this until later today.

jlongster:
Alright, pushed a new docker image with the `devtoolsRequire` patch applied so we finally have it. Responded to the review and made several tweaks too. Hopefully this passes; if so I think we should land this and keep working on it. Thanks!

PR #663, opened 2016-09-06
Always show the center pane header, fix #662

### LINKED ISSUE #662: Should the center pane header also be visible?
Currently the header in the middle section goes away if you don't have any tabs open. I feel like it should always stay visible and the welcome screen should exist below it. That way it's always consistent.

Don't feel strongly about it though, if we don't want to do it now we can close this issue and punt for later.

helenvholmes:


clarkbw:
@helenvholmes you've got a couple CSS lint errors, `npm run lint-css` should find them locally

jasonLaster:
👍

jlongster:
The only problem is that is that it hardcodes the `30px` in more places, and this bit us before when we tweaked the height of the header. But it's not a big deal for now, thanks @helenvholmes!

jlongster:
Haven't test locally, but if there are any problems we should see it pretty easily.

PR #666, opened 2016-09-06
Never merge in local config when targeting the Firefox panel

jlongster:
It's dangerous to pull in the local config when targeting the Firefox panel, as you most likely have development prefs on.

### DIFF
--- a/config/config.js
+++ b/config/config.js
@@ -9,10 +9,17 @@ const development = require("./development.json");
 const envConfig = process.env.TARGET === "firefox-panel" ?
    firefoxPanel : development;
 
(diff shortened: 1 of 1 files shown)


jasonLaster:
oh gosh, nice catch.

jlongster:
Haha :)

jasonLaster:
Actually. Just noticed that this is a decent 666 bug

jlongster:
> Actually. Just noticed that this is a decent 666 bug

Lol

jbhoosreddy:
Good one!

PR #668, opened 2016-09-07
Collapse sidebar redux

### LINKED ISSUE #426: Collapse/Expand sidebar UI

wldcordeiro:
will close #426 

screenshot_20160906_214440


jasonLaster:
Looks great. 

This is much cleaner to read than before. 

Id be happy to merge this now or wait on the on click behavior.

wldcordeiro:
@jasonLaster I was going to contact you about that. I was having some trouble thinking of how to implement that since the click will change the width of the `SplitBox`es but the buttons are in the `SourceTabs` component

jasonLaster:
Ahh, that's a good question @wldcordeiro. And, sorry for not replying earlier.

My guess, is that the center pane is always as wide as it can be (flex magic). Therefore, if you collapse a sidebar, the button should... go with it. If not, we should tweak the center pane css.

jasonLaster:
@wldcordeiro I like the direction you're going in. One comment, it probably makes sense to introduce a `ui` reducer to keep left and right sidebar widths. I imagine we'll have other UI state over time as well.

public/js/reducers/sidebars.js
```diff
@@ -0,0 +1,69 @@
...
+type SidebarsState = {
+  left: I.Map<string, SidebarState>,
+  right: I.Map<string, SidebarState>,
+};
+
+const State = makeRecord(({
+  left: I.Map({
+    collapsed: false,
+    width: 300,
+    prevWidth: 0
```
jasonLaster: I think it should be possible to keep `collapsed` and `width` and be okay:

e.g.
* app starts width is 300
* user resizes width is 370
* user collapses width is still 370, collapsed true
* user expands  width is still 370, collapsed false
* ...

public/js/reducers/sidebars.js
```diff
@@ -0,0 +1,69 @@
+// @flow
```
jasonLaster: This should be a general purpose UI reducer

public/js/reducers/sidebars.js
```diff
@@ -0,0 +1,69 @@
...
+const C = require("../constants");
+
+import type { Action } from "../actions/types";
+import type { Record } from "../utils/makeRecord";
+
+type SidebarState = {
+  collapsed: boolean,
+  width: number,
+  prevWidth: number,
+};
```
jasonLaster: 👍

public/js/components/App.js
```diff
@@ -143,6 +151,8 @@ const App = React.createClass({
 
 module.exports = connect(
   state => ({ sources: getSources(state),
-              selectedSource: getSelectedSource(state) }),
+              selectedSource: getSelectedSource(state),
+              sidebars: getSidebarsState(state)
```
jasonLaster: `getSidebarDimensions` seems more specific. I say dimensions because at somepoint we'll probably be handling height as well in the narrow view.

public/js/components/App.js
```diff
@@ -128,10 +134,12 @@ const App = React.createClass({
       { className: classnames("debugger theme-body",
                               { "theme-light": !isFirefoxPanel() }) },
       SplitBox({
-        initialWidth: 300,
+        width: this.props.sidebars.toJS().left.width,
+        resizeSidebar: this.props.resizeSidebar.bind(null, "left"),
```
jasonLaster: we've been following the wrapping pattern `(width) => this.props.resizebar("left", width)`

public/js/components/App.js
```diff
@@ -128,10 +134,12 @@ const App = React.createClass({
       { className: classnames("debugger theme-body",
                               { "theme-light": !isFirefoxPanel() }) },
       SplitBox({
-        initialWidth: 300,
+        width: this.props.sidebars.toJS().left.width,
```
jasonLaster: it'd be okay to use the getter `this.props.sidebars.getIn("left", "width")`

wldcordeiro: Heh didn't know about that one. Thanks!

public/js/components/App.js
wldcordeiro: 👍

public/js/reducers/sidebars.js
jasonLaster: @wldcordeiro do you think this is possible?

public/js/actions/sidebars.js
```diff
@@ -0,0 +1,40 @@
+// @flow
```
jasonLaster: hmm, this should be `actions/ui` as well

public/js/actions/types.js
```diff
@@ -113,13 +114,28 @@ type SourceAction =
...
+  type: C.COLLAPSE_SIDEBAR,
+  side: SidebarSide,
+  collapsed: boolean
+};
+
+type SidebarResizeAction = {
+  type: C.RESIZE_SIDEBAR,
+  side: SidebarSide,
+  width: number
+};
```
jasonLaster: 👍 I like the separate action types as we can type the reducer better in the future too

public/js/reducers/sidebars.js
wldcordeiro: Ah yeah, if I pass collapsed into splitbox this works wonderfully, `prevWidth` was only nice  because splitbox only uses `width` currently.

jasonLaster: like it

public/js/actions/ui.js
```diff
@@ -0,0 +1,46 @@
...
+  client: any
+};
+
+/**
+ * collapse a sidebar
+ *
+ * @memberof actions/ui
+ * @static
+ */
+function collapseSidebar(side: SidebarSide) {
```
jasonLaster: Maybe we want to call it `toggleSidebar` as sometimes you're expanding it :)

public/js/components/App.js
```diff
@@ -128,10 +134,14 @@ const App = React.createClass({
       { className: classnames("debugger theme-body",
                               { "theme-light": !isFirefoxPanel() }) },
       SplitBox({
-        initialWidth: 300,
+        width: this.props.ui.getIn(["left", "width"]),
+        collapsed: this.props.ui.getIn(["left", "collapsed"]),
+        resizeSidebar: (width) => this.props.resizeSidebar("left", width),
```
jasonLaster: actually, you don't need the parens around width

public/js/components/SourceTabs.js
```diff
@@ -189,15 +193,28 @@ const SourceTabs = React.createClass({
...
 
+  renderCollapseButton(side) {
+    return dom.div({
+      className: `collapse-button-${side}`,
+      onClick: (ev) => {
+        ev.preventDefault();
+        this.props.collapseSidebar(side);
+      }
+    },
+      Svg("pane-collapse"));
```
jasonLaster: i don't think those indents are needed

public/js/components/SourceTabs.js
```diff
@@ -189,15 +193,28 @@ const SourceTabs = React.createClass({
     );
   },
 
+  renderCollapseButton(side) {
+    return dom.div({
+      className: `collapse-button-${side}`,
+      onClick: (ev) => {
```
jasonLaster: I think it would be okay to do `ev => this.props.collapseSidebar(side)`

public/js/reducers/tests/ui.js
```diff
@@ -0,0 +1,69 @@
...
+declare var it: (desc: string, func: () => void) => void;
+
+const {
+  State, update } = require("../ui");
+const expect = require("expect.js");
+const C = require("../../constants");
+
+import type { SidebarsState } from "../ui";
+
+describe("ui reducer", () => {
```
jasonLaster: This could probably be an *action* test. The benefit would be that we can then exercise the actions and the selectors.

i.e.

```js
expect(getSidebarCollapsed("left")).to.be(false)
toggleSidebar("left");
expect(getSidebarCollapsed("left")).to.be(true)
```

public/js/reducers/tests/ui.js
```diff
@@ -0,0 +1,69 @@
...
+      width: 370,
+      side: "left"
+    });
+    expect(state.getIn(["left", "collapsed"])).to.be(false);
+    expect(state.getIn(["left", "width"])).to.be(370);
+    expect(state.getIn(["right", "collapsed"])).to.be(false);
+    expect(state.getIn(["right", "width"])).to.be(300);
+  });
+
+  it("should change the width for right", () => {
```
jasonLaster: we probably don't need to test both the right and left sidebar, they're symmetric after all.

public/js/reducers/tests/ui.js
wldcordeiro: I think I have that in the actions tests. 😄

public/js/components/App.js
jlongster: (it's `getIn(["left", "width"])`, you pass an array)

jlongster:
Thanks for reviewing this Jason. I haven't taken a close look but I'll let you review it.

jlongster:
And I should have said huge thanks to @wldcordeiro for implementing this :)

wldcordeiro:
@jlongster @jasonLaster you're welcome guys. With the size of this PR let's open a follow up issue for animating the buttons and the slide, I believe @helenvholmes has a video of how it should look.

wldcordeiro:
Added animations! 🎆

wldcordeiro:
@jasonLaster @helenvholmes https://dl.dropboxusercontent.com/u/3788370/collapse-sidebar-redux.mp4 video is slightly choppy but the animations are in.

jasonLaster:
Awesome work @wldcordeiro. Thanks for sticking with this as it certainly became bigger than expected!

I'm going to share a couple tweaks I found when playing with it. Would love your thoughts.

PR #677, opened 2016-09-07
Write some call stack tests

### LINKED ISSUE #650: call frames [mochitests]

browser_dbg_stack-01.js
browser_dbg_stack-02.js
browser_dbg_stack-03.js
browser_dbg_stack-04.js
browser_dbg_stack-05.js
browser_dbg_stack-06.js
browser_dbg_stack-07.js

[DXR search](https://dxr.mozilla.org/mozilla-central/search?q=path%3Adevtools%2Fclient%2Fdebugger%2Ftest%2Fmochitest%2F+file%3A*dbg_stack-%5B0-9%5D*&redirect=false)

jasonLaster:
Fixes #650


public/js/test/mochitest/examples/code_script-switching-02.js
```diff
@@ -0,0 +1,13 @@
...
+function secondCall() {
+  // This comment is useful: ☺
+  debugger;
+  function foo() {}
+  if (x) {
+    foo();
+  }
+}
+
+var x = true;
```
jasonLaster: I pulled in this example from the previous tests, but i'd be happy to use `code-simple` and add a breakpoint to pause

public/js/test/mochitest/examples/README.md
```diff
@@ -0,0 +1,5 @@
+### Test Examples
+
+##### Pages
+* **doc_script-switching-01** - includes two scripts that reference each other. The second function has a debugger.
+* **doc-scripts** - includes three sources, a long source and two sources that reference each other.
```
jasonLaster: I'd like to start documenting our test examples from the start from the beginning. It is really easy to add more and more examples when there is a perfectly good one already.

public/js/test/mochitest/head.js
```diff
@@ -204,3 +221,27 @@ function isVisibleWithin(outerEl, innerEl) {
   return innerRect.top > outerRect.top &&
     innerRect.bottom < outerRect.bottom;
 }
+
+
+const selectors = {
+  callStackArrow: ".call-stack-pane .arrow",
+  frame: (index) => `.frames ul li:nth-child(${index})`
+}
```
jlongster: This is awesome. I was thinking about something like this as well instead of having a ton of `get*` functions.

jlongster:
The `editor-mode` test timed out but that has been fixed on master. The problem was that that script it depended on was being GCed because it didn't introduce any global variables, so it was racy. You can ignore that failure or rebase.

public/js/test/mochitest/browser_dbg_call-stack.js
```diff
@@ -0,0 +1,35 @@
...
+
+  invokeInTab("firstCall");
+  yield waitForPaused(dbg);
+
+  ok(isFrameSelected(dbg, 1, "secondCall"), "the first frame is selected");
+
+  findFrame(dbg, 2).click();
+  ok(isFrameSelected(dbg, 2, "firstCall"), "the second frame is selected");
+
+  ok(true);
```
jlongster: You shouldn't need this if you have other checks in the test; this was in the stub because it needed at least one `ok`

public/js/test/mochitest/head.js
jasonLaster: yeah - i've used this pattern a lot in other projects and is both concise and descriptive. I was struggling with differentiating UI and state methods like `getFrame`and I think this helps with that as well.

public/js/test/mochitest/examples/README.md
jlongster: Any chance we could document them at the top of the HTML files instead? I think it'll be easier to keep up-to-date. Usually we add more edge cases to existing test files over time so they'll probably cover more than these simple descriptions. We should just r- PRs that add new examples unless it's really necessary.

If this really clarifies things for you though, I'm willing to try it.

jlongster: Also can you rename the new test file with the `doc-` prefix instead of `doc_`? Trying to avoid mixing underscores and dashes (but `browser_` is required by mochitest)

public/js/test/mochitest/examples/code_script-switching-02.js
jlongster: I don't mind having several test cases available that do a few different things. I think this is a good one to have too.

public/js/test/mochitest/head.js
```diff
@@ -204,3 +221,27 @@ function isVisibleWithin(outerEl, innerEl) {
...
+  return dbg.win.document.querySelector(selector);
+}
+
+function toggleCallStack(dbg) {
+  return findElement(dbg, "callStackArrow").click()
+}
+
+Services.prefs.setBoolPref("devtools.debugger.new-debugger-frontend", true);
+registerCleanupFunction(() => {
+  Services.prefs.setBoolPref("devtools.debugger.new-debugger-frontend", false);
```
jlongster: I think you've duplicated this. It's already at the top of `head.js`, so we don't need this here.

jlongster:
👍 Looks great to me! A few notes, but I like it.

public/js/test/mochitest/head.js
jlongster: Yep, that was my concern too, that we'd be replicating selectors and it'd get confusing... this pattern is great though. I don't think we should be too stringent about keeping selectors out of tests, but common ones should certainly be in here.

public/js/test/mochitest/examples/README.md
jasonLaster: ha - yeah the `-` and `_` bothered me too. I left it in as i was just removing the code switching example.

I'd like to try keeping the readme up to date if that's okay. With tests it helps me to have a global spec i can see after each run to understand what's being covered. We don't really have that for the examples and it might be nice to skim the list to have a sense. I hate reading one example and then looking at another and trying to keep it all in my head. I can see this getting out of date if we don't remember it, so i'd be happy to abandon the experiment in a week if it doesn't work for us

public/js/test/mochitest/browser_dbg_call-stack.js
jasonLaster: cool

public/js/test/mochitest/head.js
jasonLaster: oh nice

public/js/test/mochitest/examples/code_script-switching-02.js
jasonLaster: cool

public/js/test/mochitest/examples/README.md
jlongster: Ok, we can try it. I agree that we should avoid duplicating test cases. In the old tests, most of them use just a few of the examples. The two that we have right now will probably cover a lot of them. Of course, there are plenty of edge cases that will need new examples for specific things. Maybe in the readme we could make a not about the "common" test cases that you should probably use unless you know what you're doing.

jasonLaster:
The only tests that failed were cypress

Issue #678, opened 2016-09-07
Local Test Runner


jasonLaster:
We're currently writing many integration tests with mochitests, which has several benefits:
* Debugger runs inside the panel
* Redux oriented tests (drive the debugger with actions)
* Tests will run on Try as well as Circle
* Familiar to the team

It would be good to explore a local test runner, which would use mocha and drive the debugger in an iframe
* We would not change how the tests are written (redux style) or how they're run on CI (mochitest)
* tighter feedback loop with no setup. Launch the test app and refresh to re-run the tests. It would also be possible to run the tests headlessly with mocha-electron.
* easier to debug - it is easier to debug an iframe with devtools than to debug with the browser toolbox.
* debug other targets - it is trivial to debug chrome, node, ..

![](http://g.recordit.co/vClVoKC27M.gif)

jasonLaster:
Closing this issue because it is not tied to our immediate goals. I think a local runner would be a nice workflow improvement and it would be helpful for writing integration tests for the standalone application, but I don't see us having the time to work on this in the near future.

PR #679, opened 2016-09-07
Remove cypress

jasonLaster:
I'm removing cypress now because we're adding similar mochitests.

Also, cypress is still flakey so if we keep the tests on, we're likely
to get a lot of false failures, which we won't fix.

I believe a lot of the benefits we saw with cypress can be captured with a local test runner - https://github.com/devtools-html/debugger.html/issues/678


jlongster:
Sounds good to me, thanks!

PR #681, opened 2016-09-07
Use CodeMirror from mozilla-central when in panel

jlongster:
This will pass once [bug 1295213](https://bugzilla.mozilla.org/show_bug.cgi?id=1295213) lands. All tests currently still pass, which is great. From here we could play with enabling code search, maybe?

Hoping to land that bug today, or at the latest tomorrow. Should have this before enabling it on nightly tomorrow.

### DIFF
--- a/public/js/components/Editor.js
+++ b/public/js/components/Editor.js
@@ -143,10 +143,8 @@ const Editor = React.createClass({
       value: " "
     });
 
-    this.editor.appendTo(
-      ReactDOM.findDOMNode(this).querySelector(".editor-mount"),
-      null,
-      true
+    this.editor.appendToLocalElement(
+      ReactDOM.findDOMNode(this).querySelector(".editor-mount")
     );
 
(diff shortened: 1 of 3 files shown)


jasonLaster:
👍

jlongster:
Rebuilt docker image, kicking Ci. This might work now.

jlongster:
Sigh, I forgot that I also need to pull in the local codemirror CSS when in the panel. Requires modifications to index.html. Done in [bug 1301511](https://bugzilla.mozilla.org/show_bug.cgi?id=1301511) and already pushed to fx-team.

I was hoping to land this, but might as well wait until tomorrow when everything's moved to mozilla-central.

jlongster:
Yay, test finally passed! The only thing that's failing is cypress, as this is an old PR. Going to merge this as it's important to get tests running with the local codemirror.

jlongster:
Fixed a small conflict

PR #683, opened 2016-09-07
Add pauseTest command

jasonLaster:
Simple way to pause and then resume a test which lets you interact with the debugger.

### DIFF
--- a/docs/mochitests.md
+++ b/docs/mochitests.md
@@ -77,6 +77,7 @@ Any of the below APIs that takes a `url` will match it as a substring, meaning t
 * `waitForState(dbg, predicate)` - Waits for `predicate(state)` to be true. `state` is the redux app state
 * `waitForThreadEvents(dbg, eventName)` - Waits for specific thread events
 * `waitForDispatch(dbg, type)` - Wait for a specific action type to be dispatch. If an async action, will wait for it to be done.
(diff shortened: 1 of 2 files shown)


docs/mochitests.md
```diff
@@ -78,6 +78,19 @@ Any of the below APIs that takes a `url` will match it as a substring, meaning t
 * `waitForThreadEvents(dbg, eventName)` - Waits for specific thread events
 * `waitForDispatch(dbg, type)` - Wait for a specific action type to be dispatch. If an async action, will wait for it to be done.
 
+## Pausing
```
jlongster: Do we need a whole section for this? Why not just document the `pauseTest` helper in the API section? Keeping the docs smaller makes it easier to go through, although it's balance.

jasonLaster: That's a good point - happy to add it to the API and also add some of the newer methods like clickElement, findelement...

jlongster: You could rename this to `Debugging`, it probably is good to have a small section calling that out, and mentioning the `--jsdebugger` allows you to use the debugger and mention this command. Not sure we need to detail breakpoints/debugger specifically though :)

Thanks for this new function!

jlongster:
👍 Looks good to me

docs/mochitests.md
jasonLaster: thought about it a bit more and decided adding the command to the api section was fine

PR #684, opened 2016-09-07
Use local React when in panel

jlongster:
React and CodeMirror are the two modules I really wanted to share with mozilla-central. The rest of the libraries are smaller and not required as much for interop, so I'm not as concerned about them. This reduces the size of our bundle from 2.3MB to 1.5MB (excluding both react and codemirror), and makes sure we are compatible with the rest of devtools.

This depends on [bug 1301147](https://bugzilla.mozilla.org/show_bug.cgi?id=1301147) landing before we can merge.

### DIFF
--- a/webpack.config.devtools.js
+++ b/webpack.config.devtools.js
@@ -12,7 +12,13 @@ const ignoreRegexes = [
 const nativeMapping = {
   // Don't map this until bug 1295318 lands
   // "public/js/utils/source-editor": "devtools/client/sourceeditor/editor"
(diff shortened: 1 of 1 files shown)


jasonLaster:
👍

jlongster:
(This includes the editor PR which will be merged first)

jlongster:
Rebuilt docker image, kicking Ci. This might work now.

jlongster:
Finally, tests passing!

PR #688, opened 2016-09-07
Do not include any theme CSS when compiling for the panel

jlongster:
Right now we are bundling in the theme CSS and we definitely shouldn't be doing that. We need to let the toolbox apply the themes for us.


jasonLaster:
Nice

PR #689, opened 2016-09-07
Swaps out hard coded css values for theme variables

helenvholmes:


jlongster:
I ran the CSS linter locally and it all passes, so merging this as it's only CSS changes. Eager to see if the dark theme works well now

jasonLaster:
awesome!

helenvholmes:
@jlongster it works kinda sortttaaaaaa

But better at least

PR #690, opened 2016-09-07
cleanup api

jasonLaster:


jlongster:
Looks like I need to update the docker build since your console landed. Thanks for opening this PR; I need to update it today for other reasons too. I'll do that now.

jasonLaster:
Sounds good

jlongster:
New docker image available. Should work now.

jlongster:
Oh, I can't trigger it because I didn't create it. Can you kick the CI?

jasonLaster:
kicked

jlongster:
I reverted the image back to a good one, so you can use it now. Still working on updating it.

jlongster:
Mind kicking it again? :)

Issue #642, opened 2016-08-31
keyboard shortcuts [mochitests]


clarkbw:
Sub task of the All the mochitests! #634 issue

[browser_dbg_interrupts.js](https://dxr.mozilla.org/mozilla-central/source/devtools/client/debugger/test/mochitest/browser_dbg_interrupts.js)

jbhoosreddy:
I'm just trying this one out to see if I get somewhere with this.

jbhoosreddy:
Resolved in #691 and #718.

PR #691, opened 2016-09-08
add tests for keyboard shortcuts (#642)

jbhoosreddy:
It took me some time but I think I'm almost there. thanks to @jlongster and @jasonLaster's previous tests. I tried to replicate the original test to a good extent.

I would like to know if there is a better way `ensureThreadResumed` than what I did? Let me know what you think and I will make changes accordingly.

### DIFF
--- a/public/js/components/RightSidebar.js
+++ b/public/js/components/RightSidebar.js
@@ -10,6 +10,10 @@ const { isEnabled } = require("../feature");
 const Svg = require("./utils/Svg");
 const ImPropTypes = require("react-immutable-proptypes");
 
+const { Services } = require("Services");
+const isOSX = Services.appinfo.OS === "Darwin";
+const shiftKey = isOSX ? "\u21E7" : "Shift+";
+
 const actions = require("../actions");
 const Breakpoints = React.createFactory(require("./Breakpoints"));
 const Expressions = React.createFactory(require("./Expressions"));
@@ -88,7 +92,7 @@ const RightSidebar = React.createClass({
     keyShortcuts.on("F8", this.resume);
     keyShortcuts.on("F10", this.stepOver);
     keyShortcuts.on("F11", this.stepIn);
-    keyShortcuts.on("F12", this.stepOut);
+    keyShortcuts.on("Shift+F11", this.stepOut);
   },
 
   componentWillUnmount() {
@@ -96,7 +100,7 @@ const RightSidebar = React.createClass({
     keyShortcuts.off("F8", this.resume);
     keyShortcuts.off("F10", this.stepOver);
     keyShortcuts.off("F11", this.stepIn);
-    keyShortcuts.off("F12", this.stepOut);
+    keyShortcuts.off("Shift+F11", this.stepOut);
   },
 
   componentDidUpdate() {
@@ -108,7 +112,7 @@ const RightSidebar = React.createClass({
     return [
       debugBtn(this.stepOver, "stepOver", className, "Step Over (F10)"),
       debugBtn(this.stepIn, "stepIn", className, "Step In (F11)"),
-      debugBtn(this.stepOut, "stepOut", className, "Step Out \u21E7 (F12)"),
+      debugBtn(this.stepOut, "stepOut", className, `Step Out (${shiftKey}F11)`),
     ];
   },
 
--- a/public/js/test/mochitest/browser.ini
+++ b/public/js/test/mochitest/browser.ini
@@ -26,3 +26,4 @@ support-files =
 [browser_dbg-chrome-debugging.js]
 [browser_dbg-iframes.js]
 [browser_dbg-debugger-buttons.js]
+[browser_dbg_keyboard-shortcuts.js]
--- a/public/js/test/mochitest/head.js
+++ b/public/js/test/mochitest/head.js
@@ -261,6 +261,23 @@ function invokeInTab(fnc) {
   });
 }
 
+const keyMappings = {
+  pauseKey: { code: "VK_F8" },
+  resumeKey: { code: "VK_F8" },
+  stepOverKey: { code: "VK_F10" },
+  stepInKey: { code: "VK_F11" },
+  stepOutKey: { code: "VK_F11", modifiers: { shiftKey: true } }
+};
+
+function pressKey(dbg, keyName) {
+  let keyEvent = keyMappings[keyName];
+  const { code, modifiers } = keyEvent;
+  return EventUtils.synthesizeKey(
+    code,
+    modifiers || {},
+    dbg.win
+  );
+}
 
 function isVisibleWithin(outerEl, innerEl) {
   const innerRect = innerEl.getBoundingClientRect();
@@ -280,7 +297,7 @@ const selectors = {
   stepOver: ".stepOver.active",
   stepOut: ".stepOut.active",
   stepIn: ".stepIn.active"
-}
+};
 
 function getSelector(elementName, ...args) {
   let selector = selectors[elementName];
--- a/public/js/test/mochitest/.eslintrc
+++ b/public/js/test/mochitest/.eslintrc
@@ -59,6 +59,7 @@
     "toggleCallStack": false,
     "isVisibleWithin": false,
     "clickElement": false,
-    "togglePauseOnExceptions": false
+    "togglePauseOnExceptions": false,
+    "pressKey": false
   }
 }
--- a/public/js/test/mochitest/browser_dbg_keyboard-shortcuts.js
+++ b/public/js/test/mochitest/browser_dbg_keyboard-shortcuts.js
@@ -1,0 +1,52 @@
+/* Any copyright is dedicated to the Public Domain.
+ * http://creativecommons.org/publicdomain/zero/1.0/ */
+
+/**
+ * Test keyboard shortcuts.
+ */
+
+function pressResume(dbg) {
+  pressKey(dbg, "resumeKey");
+  return waitForPaused(dbg);
+}
+
+function pressStepOver(dbg) {
+  pressKey(dbg, "stepOverKey");
+  return waitForPaused(dbg);
+}
+
+// function pressStepIn(dbg) {
+//   pressKey(dbg, "stepInKey");
+//   return waitForPaused(dbg);
+// }
+
+// function pressStepOut(dbg) {
+//   pressKey(dbg, "stepOutKey");
+//   return waitForPaused(dbg);
+// }
+
+add_task(function*() {
+  const dbg = yield initDebugger(
+    "doc-debugger-statements.html",
+    "debugger-statements.html"
+  );
+
+  yield reload(dbg);
+  yield waitForPaused(dbg);
+  assertPausedLocation(dbg, "debugger-statements.html", 8);
+
+  yield pressResume(dbg);
+  assertPausedLocation(dbg, "debugger-statements.html", 12);
+
+  yield pressStepOver(dbg);
+  assertPausedLocation(dbg, "debugger-statements.html", 13);
+
+  // yield pressStepIn(dbg);
+  // assertPausedLocation(dbg, "debugger-statements.html", 18);
+
+  // yield pressStepOver(dbg);
+  // assertPausedLocation(dbg, "debugger-statements.html", 20);
+
+  // yield pressStepOut(dbg);
+  // assertPausedLocation(dbg, "debugger-statements.html", 20);
+});


public/js/test/mochitest/browser_dbg_interrupts.js
```diff
@@ -0,0 +1,89 @@
...
+  ok(findElement(dbg, "button", "pause", "active"), "Pause button disabled");
+  invokeInTab("firstCall");
+  yield waitForPaused(dbg);
+  ok(findElement(dbg, "button", "resume", "active"), "Resume button enabled");
+  pressKey(dbg, resumeKey);
+  yield waitForThreadEvents(dbg, "resumed");
+  yield waitForPaused(dbg);
+
+  // Evaluate script while debugger is paused
+  const result = yield evaluate(dbg, "1+1")
```
jlongster: I'm not sure I see the purpose of evaluating in this test. This doesn't really test anything related to keyboard shortcuts for stepping/resuming or the stepping buttons. This looks like it was ported over from the old test, but I think we can avoid the evaluate and adding the `evaluate` command for now.

Just trying to make the tests a little more focused than they were before.

public/js/test/mochitest/browser_dbg_interrupts.js
```diff
@@ -0,0 +1,89 @@
...
+];
+const resumeKey = "VK_F8";
+
+add_task(function*() {
+  const dbg = yield initDebugger(TAB_URL, ...sourceURLs);
+  const { selectors: { getBreakpoints, getBreakpoint }, getState } = dbg;
+  const sources = [];
+  for (let url of sourceURLs) {
+    sources.push(findSource(dbg, url));
+  }
```
jlongster: I think it's more descriptive and simple if you just do:

```js
const source1 = findSource(dbg, "switching-01.js");
const source2 = findSource(dbg, "switching-02.js");
```

Admittedly, it's not much more descriptive because the script names are just 01 and 02, but it's shorter code anyway. Most tests will only deal with 2 or 3 sources so it's not a big deal to do this for each, and it's easier to debug (you see `source1` and `source2` in scope instead of an array)

public/js/test/mochitest/browser_dbg_interrupts.js
```diff
@@ -0,0 +1,89 @@
...
+
+  yield ensureAddBreakpoint(dbg, getState, getBreakpoint, sources[0].id, 5);
+  yield ensureAddBreakpoint(dbg, getState, getBreakpoint, sources[1].id, 6);
+  yield ensureAddBreakpoint(dbg, getState, getBreakpoint, sources[1].id, 7);
```
jasonLaster: I think it is okay to just add the breakpoint

`yield addBreakpoint()`

We're testing breakpoints in more depth in other spots. Also, the test will also be below when you expect to land in a certain spot.

public/js/test/mochitest/browser_dbg_interrupts.js
```diff
@@ -0,0 +1,89 @@
...
+
+const testResumeKey = Task.async(function* (dbg) {
+  info("Testing Resume Key");
+  ok(findElement(dbg, "button", "pause", "active"), "Pause button disabled");
+  invokeInTab("firstCall");
+  yield waitForPaused(dbg);
+  ok(findElement(dbg, "button", "resume", "active"), "Resume button enabled");
+  pressKey(dbg, resumeKey);
+  yield waitForThreadEvents(dbg, "resumed");
+  yield waitForPaused(dbg);
```
jlongster: Very cool, thanks for porting this test. Awesome to have this stuff tested.

public/js/test/mochitest/browser_dbg_interrupts.js
```diff
@@ -0,0 +1,89 @@
...
+  for (let url of sourceURLs) {
+    sources.push(findSource(dbg, url));
+  }
+
+  yield ensureAddBreakpoint(dbg, getState, getBreakpoint, sources[0].id, 5);
+  yield ensureAddBreakpoint(dbg, getState, getBreakpoint, sources[1].id, 6);
+  yield ensureAddBreakpoint(dbg, getState, getBreakpoint, sources[1].id, 7);
+  is(getBreakpoints(getState()).size, 3, "Three breakpoint exists");
+
+  yield testResumeButton(dbg);
```
jasonLaster: I think it would be better to expand some of this out into a couple lines

```js
invokeInTab("firstCall");
yield waitForPaused(dbg);
```

also, i think passing `.button.active` as a full selector is fine

public/js/test/mochitest/browser_dbg_interrupts.js
```diff
@@ -0,0 +1,89 @@
...
+});
+
+const ensureAddBreakpoint = Task.async(function* (dbg, getState, getBreakpoint,
+                                                  sourceId, line) {
+  yield addBreakpoint(dbg, sourceId, line);
+  ok(getBreakpoint(getState(), { sourceId, line }),
+    "Breakpoint has correct line");
+});
+
+const testResumeButton = Task.async(function* (dbg) {
```
jasonLaster: I'd prefer if this just tested the key shortcuts and not the buttons

public/js/test/mochitest/browser_dbg_interrupts.js
```diff
@@ -0,0 +1,89 @@
...
+  // Evaluate script while debugger is paused
+  const result = yield evaluate(dbg, "1+1")
+    .then(response => response.result);
+  is(result, 2, "Evaluated script while debugger paused");
+  yield ensureThreadResumed(dbg);
+});
+
+const testResumeKey = Task.async(function* (dbg) {
+  info("Testing Resume Key");
+  ok(findElement(dbg, "button", "pause", "active"), "Pause button disabled");
```
jlongster: You're passing 2 additional args to `findElement` but looks like the selector is `(button, state) => `.${button}.${state}`  ? It only takes 1 additional arg.

public/js/test/mochitest/browser_dbg_interrupts.js
jlongster: Personally I'd be fine with just doing `addBreakpoint` and not caring if the breakpoint slid or not. We know our test files, so it shouldn't slide. If it's mainly checking that it was successfully set, errors will propagate from `addBreakpoint` anyway.

jlongster:
Several small nits but this generally is awesome :) Thanks for taking the time to do this, it helps *greatly*!

jlongster:
Generally I wouldn't worry about porting old tests too carefully. They are generally bad, but they are good to look at to see specific edge cases to test. Feel free to write simple-ish tests for now and just look at the old tests to make sure there isn't something important we're not testing. Thanks for carefully looking through the old test.

public/js/test/mochitest/browser_dbg_interrupts.js
jbhoosreddy: > I think it would be better to expand some of this out into a couple lines
Could you elaborate what you meant here?

Also do you mean this?
`findElement(dbg, "button", ".resume.active")`
and 
```
const selectors = {
  ...,
  button: selector => selector,
};
```

jasonLaster: actually, i think it would be better to have more elements like `findElement("resumeButton")`

jbhoosreddy:
I modified the PR based on all the review comments.

public/js/components/RightSidebar.js
```diff
@@ -88,15 +88,15 @@ const RightSidebar = React.createClass({
...
+    keyShortcuts.on("Shift+F11", this.stepOut);
   },
 
   componentWillUnmount() {
     const { keyShortcuts } = this.props;
     keyShortcuts.off("F8", this.resume);
     keyShortcuts.off("F10", this.stepOver);
     keyShortcuts.off("F11", this.stepIn);
-    keyShortcuts.off("F12", this.stepOut);
+    keyShortcuts.off("Shift+F11", this.stepOut);
```
jlongster: Why this change? (may have miss the conversation) Is this what the old debugger does? It'd be good to keep the same shortcuts if possible, but there's a good reason that's ok.

public/js/test/mochitest/head.js
```diff
@@ -225,8 +235,13 @@ function isVisibleWithin(outerEl, innerEl) {
...
   callStackHeader: ".call-stack-pane ._header",
   frame: (index) => `.frames ul li:nth-child(${index})`,
-  gutter: line => `.CodeMirror-code *:nth-child(${line}) .CodeMirror-linenumber`
-}
+  gutter: line => `.CodeMirror-code *:nth-child(${line}) .CodeMirror-linenumber`,
+  pauseKey: { keyCode: "VK_F8" },
+  resumeKey: { keyCode: "VK_F8" },
+  stepOverKey: { keyCode: "VK_F10" },
+  stepInKey: { keyCode: "VK_F11" },
+  stepOutKey: { keyCode: "VK_F12", modifiers: { shiftKey: true } }
```
jlongster: Not sure these belong in `selectors`? These selectors are meant to be CSS selectors, and this just seems like a key mapping. You could pull this out into it's own object next to the `pressKey` function.

jlongster:
New version looks great (w/a few nits), thanks!

jbhoosreddy:
The change is because after I took out the button tests (as per Jason's
suggestion), there was not much my test was testing. So I decided to test
the different key shortcuts.

I could revert back if you guys think.

In either case, I'll work on these comments.

jlongster:
> The change is because after I took out the button tests (as per Jason's suggestion), there was not much my test was testing. So I decided to test the different key shortcuts.

Oh testing all shortcuts is great :) I was talking about this change:

```
-    keyShortcuts.off("F12", this.stepOut);
+    keyShortcuts.off("Shift+F11", this.stepOut);
```

Which changed code, not a test, and changes the step out to be shift+F11 instead of F12.

jlongster:
This is very close though. Thank you!

public/js/components/RightSidebar.js
jbhoosreddy: Yes. The old debugger has `Shift+F11` for `Step Out`. I wrote the code for the keyboard shortcuts.. and I think I chose `F12` by mistake. I will rectify that here.

public/js/test/mochitest/head.js
jasonLaster: agreed. This would make sense as a keyMappings dictionary.

public/js/test/mochitest/browser_dbg_interrupts.js
```diff
@@ -0,0 +1,56 @@
...
+  yield testKey(dbg, "stepOutKey");
+});
+
+const testKey = Task.async(function* (dbg, key) {
+  info("Testing Key: " + key);
+  const { selectors: { getPause }, getState } = dbg;
+  yield waitForPaused(dbg);
+  ok(getPause(getState()), "Debugger is paused");
+  pressKey(dbg, key);
+});
```
jasonLaster: I don't think `testKey` is needed.

I'd prefer to invert the sequence. First press the key, and then check the pause location.

public/js/test/mochitest/browser_dbg_interrupts.js
```diff
@@ -0,0 +1,56 @@
...
+  ok(!getPause(getState()), "Debugger is resumed");
+
+  yield testKey(dbg, "stepOutKey");
+});
+
+const testKey = Task.async(function* (dbg, key) {
+  info("Testing Key: " + key);
+  const { selectors: { getPause }, getState } = dbg;
+  yield waitForPaused(dbg);
+  ok(getPause(getState()), "Debugger is paused");
```
jasonLaster: I have an `isPaused` function in my test that will be helpful here

public/js/test/mochitest/browser_dbg_interrupts.js
```diff
@@ -0,0 +1,56 @@
...
+
+  yield testKey(dbg, "stepInKey");
+  yield waitForThreadEvents(dbg, "resumed");
+  ok(!getPause(getState()), "Debugger is resumed");
+
+  yield testKey(dbg, "stepOutKey");
+});
+
+const testKey = Task.async(function* (dbg, key) {
+  info("Testing Key: " + key);
```
jasonLaster: nit: I don't think we need the info, we'll see where it's failing with the right assertions

public/js/test/mochitest/browser_dbg_interrupts.js
```diff
@@ -0,0 +1,56 @@
...
+  const source2 = findSource(dbg, "code-script-switching-02.js");
+
+  yield addBreakpoint(dbg, source1.id, 5);
+  yield addBreakpoint(dbg, source2.id, 6);
+  yield addBreakpoint(dbg, source1.id, 7);
+
+  yield testBreakOnNextState(dbg);
+
+  invokeInTab("firstCall");
+  yield testKey(dbg, "resumeKey");
```
jasonLaster: There should be a `yield waitForPaused` here

public/js/test/mochitest/browser_dbg_interrupts.js
```diff
@@ -0,0 +1,56 @@
...
+  yield waitForThreadEvents(dbg, "resumed");
+  ok(!getPause(getState()), "Debugger is resumed");
```
jasonLaster: We should probably check the pause location instead of testing that at somepoint it was resumed:
```
assertPauseLocation(dbg, "code-script-switching-01.js", 12)
```

public/js/test/mochitest/browser_dbg_interrupts.js
```diff
@@ -0,0 +1,56 @@
...
+  ok(getPause(getState()), "Debugger is paused");
+  pressKey(dbg, key);
+});
+
+const testBreakOnNextState = Task.async(function* (dbg) {
+  const { selectors: { getPause, getIsWaitingOnBreak }, getState } = dbg;
+  ok(!getPause(getState()), "Debugger is resumed");
+  pressKey(dbg, "pauseKey");
+  ok(getIsWaitingOnBreak(getState()), "Breaking on next execution");
+});
```
jasonLaster: we don't need to test breaking on next here, so we just need to find a way to pause. Honestly, the easiest way might be to use the new `debugger-statements.html` i'm adding and go through the same steps as that test.

that way you don't need breakpoints or break on next

jbhoosreddy: a lot of these helpers are are there in your PR. I'll wait for that to get merged and then work on your comments. 👍

jasonLaster: cool. shouldn't be too long

jasonLaster:
@jbhoosreddy #697 just landed

jbhoosreddy:
My test is failing on Circle, but it looks fine on my end. I'm not too sure how to fix it though.

jasonLaster:
Hmm, @jbhoosreddy have you rebased lately?

public/js/test/mochitest/browser_dbg_interrupts.js
```diff
@@ -0,0 +1,57 @@
...
+  // step into
+  yield pressStepIn(dbg);
+  assertPausedLocation(dbg, "debugger-statements.html", 18);
+
+  // step over
+  yield pressStepOver(dbg);
+  assertPausedLocation(dbg, "debugger-statements.html", 20);
+
+  // step out
+  yield pressStepOut(dbg);
```
jasonLaster: i believe my original test has these comments, but they're probably not necessary any more :)

public/js/test/mochitest/browser.ini
```diff
@@ -26,3 +26,4 @@ support-files =
 [browser_dbg-chrome-debugging.js]
 [browser_dbg-iframes.js]
 [browser_dbg-debugger-buttons.js]
+[browser_dbg_interrupts.js]
```
jasonLaster: we should probably rename this to `test-keyboard-shortcuts`

public/js/test/mochitest/browser_dbg_interrupts.js
jbhoosreddy: haha.. okay.. I can remove them.. :D

public/js/components/RightSidebar.js
```diff
@@ -108,7 +112,7 @@ const RightSidebar = React.createClass({
     return [
       debugBtn(this.stepOver, "stepOver", className, "Step Over (F10)"),
       debugBtn(this.stepIn, "stepIn", className, "Step In (F11)"),
-      debugBtn(this.stepOut, "stepOut", className, "Step Out \u21E7 (F12)"),
+      debugBtn(this.stepOut, "stepOut", className, `Step Out (${shiftKey}F11)`),
```
jasonLaster: I'm not sure how this will look with the parens

jasonLaster:
👍 looks good minus some nits. I wonder if this is something to do with running on linux?

public/js/components/RightSidebar.js
jbhoosreddy: this is how it looks in the old debugger. I just replicated that.

jasonLaster:
Just commenting for my own memory: @jbhoosreddy is looking into reproducing the linux failures, which look related to the key shortcuts library.

@jbhoosreddy If you'd like, it'd be okay if you removed those key shortcuts from this PR and then opened a new issue. The good thing with that is that fixing the key shortcuts library is a different kind of review

jbhoosreddy:
@jasonLaster I updated the PR then. I commented out those lines and I will un-comment them once I fix it in a new PR.

jasonLaster:
Thanks @jbhoosreddy! it always feels good to merge!

PR #693, opened 2016-09-08
pause on exceptions

jasonLaster:


.gitignore
```diff
@@ -6,4 +6,4 @@ npm-debug.log
 .idea/
 .vscode/
 cypress/screenshots
-firefox
\ No newline at end of file
+/firefox
```
jlongster: Heh, @jbhoosreddy changed this to `firefox/` in his patch.

public/js/test/mochitest/browser_dbg-editor-select.js
```diff
@@ -43,12 +30,12 @@ add_task(function* () {
   // Step through to another file and make sure it's paused in the
   // right place.
   yield stepIn(dbg);
-  assertPausedLocation(dbg, simple2, 2);
+  assertPausedLocation(dbg, "simple2.js", 2);
```
jlongster: I'd prefer to make anything that takes source support a string like this and passing the source object. I want to write some tests with eval sources soon and can't use urls like this. And here, since I already have the source objects, we could just use them.

public/js/test/mochitest/head.js
```diff
@@ -95,6 +95,33 @@ function waitForState(dbg, predicate) {
   });
 }
 
+function waitForMs(time) {
+  return new Promise(resolve => setTimeout(resolve, time));
```
jlongster: We already have this as a shared utility: `waitForTime`. Note that we should never actually use it in tests except for debugging :)

jlongster:
Looks good, just a few nits. You could update `findSource` to check if the `url` param is an object which means it's actually a full source already and just return it. That would allow us to pass a source through any of these APIs.

.gitignore
jasonLaster: I believe that matches `clients/firefox` unfortunately

public/js/test/mochitest/browser_dbg-editor-select.js
jasonLaster: yeah - that works for me

public/js/test/mochitest/head.js
jasonLaster: oops. didn't mean to add this

PR #695, opened 2016-09-08
Cleanup some of the test cruft

jasonLaster:
+ move EXAMPLE_URL to initDebugger
+ remove copyright information as it is not consistent with the rest of
the project.

^ I'm not sure if this is okay.


jasonLaster:
going to re-add the licenses

jlongster:
I like the TAB_URL cleanup. We can talk about automating it later, but I like that's we don't process the test files at all (someone looking at both dirs sees that they are exactly the same), but maybe we'll do it eventually.

jasonLaster:
> someone looking at both dirs sees that they are exactly the same

that's a good point

PR #697, opened 2016-09-08
Adds two tests (iframes and debugger buttons)

jasonLaster:
This adds two tests:
* iframes
* and debugger buttons

It's going to fail as I'm leaving out the pause on exceptions commit for review. Once that lands, we can re-run ci.


public/js/actions/navigation.js
```diff
@@ -6,7 +6,7 @@ function willNavigate() {
   return { type: constants.NAVIGATE };
 }
 
-function navigate() {
+function navigated() {
```
jasonLaster: this might be the most controversial change in the PR.

This came up because i needed to add a reload action, which is triggering a navigation. The pattern setup in the pause actions is that we use present tense for client actions and past tense for server actions e.g. (pause, paused).

public/js/test/mochitest/browser_dbg-debugger-buttons.js
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * Test debugger buttons
+ *  1. resume
+ *  2. stepOver
+ *  3. stepIn
+ *  4. stepOver to the end of a function
+ *  5. stepUp at the end of a function
+ */
+add_task(function* () {
+  const dbg = yield initDebugger(TAB_URL, "debugger-statements.html");
```
jasonLaster: i'll change this once the cleanup lands

public/js/actions/navigation.js
jlongster: I have more comments about `reload` actually :) I don't mind this change.

public/js/actions/navigation.js
```diff
@@ -19,7 +19,19 @@ function navigate() {
   };
 }
 
+function reload() {
```
jlongster: I'm wary of adding actions that aren't explicitly for users to use. This is just for testing, so I'd rather take a different approach.  Let's just expose the client to the mochitests just like we expose the other things, and then tests can just call `client.reload()` themselves. What do you think?

jlongster:
Looks generally good except for my one comment.

public/js/actions/navigation.js
jasonLaster: that makes sens

public/js/test/mochitest/browser_dbg-debugger-buttons.js
```diff
@@ -0,0 +1,57 @@
...
+    "doc-debugger-statements.html",
+    "debugger-statements.html"
+  );
+
+  yield reload(dbg);
+  yield waitForPaused(dbg);
+  assertPausedLocation(dbg, "debugger-statements.html", 8);
+
+  // resume
+  clickElement(dbg, "resume");
```
jbhoosreddy: It is a little late but... I was wondering if there is some reason you did not pull it out into it's own function for consistency?

jasonLaster: Nope. Guess I could

PR #698, opened 2016-09-08
breakpoint list formatting tweaks

jlongster:
This changes the breakpoint list to show filename and line as "foo.js: 5" instead of "foo.js5". It also fixed a margin issue with the "no breakpoints" message; there was a big margin on the bottom of it. The HTML/CSS has been simplified too so that the whole area is a clickable breakpoint item (previously, only the file/line was).

There's a new proposal in here: in order to simplify the markup and CSS, I wanted see what it would be like to avoid flexbox for this small thing, and it works pretty well. The only hard-ish part was the paused icon, and personally I think it would be better if we just highlighted the breakpoint:

screen shot 2016-09-08 at 4 58 35 pm

(The background color could be whatever)

I'm not sure of the benefit of an icon here, as it's not a button or anything. What do you all think?

I can revert some of this if needed, but we should land this soon because I don't want to update the debugger on m-c with the current bp styles.

### DIFF
--- a/public/js/components/Breakpoints.css
+++ b/public/js/components/Breakpoints.css
@@ -1,26 +1,15 @@
-.breakpoints-list {
-  padding-bottom: 10px;
-}
 
 .breakpoints-list .breakpoint {
   font-size: 12px;
   color: var(--theme-content-color1);
-  margin-top: 0.5em;
+  margin: 0.25em 0;
+  padding: 0.25em 0;
   line-height: 1em;
-  display: flex;
-  flex-flow: row;
-  align-items: center;
 }
 
-.breakpoints-list .breakpoint input {
-  flex: 0 1 content;
-  order: 1;
-}
-
-/*
 .breakpoints-list .breakpoint.paused {
+  background-color: var(--theme-toolbar-background-alt);
 }
-*/
 
 .breakpoints-list .breakpoint.disabled .breakpoint-label {
   color: var(--theme-content-color3);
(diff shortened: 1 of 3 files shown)


jasonLaster:
👍

jasonLaster:
For what it's worth, chrome is doing both patterns - in the call stack they have an icon, in the breakpoints they use a highlight:

screen shot 2016-09-08 at 8 15 42 pm

clarkbw:
> There's a new proposal in here

I think it might be good to split this up into two different PRs.  

We could get the small formatting fix in now and then look into alternate ways to highlight the paused breakpoint later.

jlongster:
> I think it might be good to split this up into two different PRs.

The reason it's not is because I need this simpler HTML to fix the margin bug with the "no breakpoints" cleanly. Also the new HTML makes it a lot more consistent - if you look at the code all those props should actually be on the outer component (so that you can click anywhere in item and it works), and it's all within a "breakpoint" class. The problem with this is that flexbox is applied to that class, etc.

I think it's weird to reuse the an icon that looks like a button (it looks like both the pause and stop on exceptions button). The arrow that chrome uses makes sense. So regardless of what we do in the future, I think this is an improvement that makes it feel more solid.

jlongster:
There are several small improvements here that would be harder to do without avoid simplifying the HTML so asked a favor to go ahead and merge this :) The design will be worked on very soon as we figure out a better way to show the paused state. Unfortunately I missed some linting errors in #703 because I didn't wait for CI which was my mistake, so I fixed this in a follow-up commit here. The rest passes.

PR #699, opened 2016-09-09
Cleanup editor-select test

jasonLaster:
Cleaned up an *older* test with some of the new helpers

+ introduce invokeInTab
+ add some helper functions
+ remove selectors

it might be easier to look at the commit as i needed to include another commit for the work

### DIFF
--- a/public/js/test/mochitest/browser_dbg-editor-select.js
+++ b/public/js/test/mochitest/browser_dbg-editor-select.js
@@ -1,6 +1,16 @@
 /* Any copyright is dedicated to the Public Domain.
  * http://creativecommons.org/publicdomain/zero/1.0/ */
 
+// Tests that the editor highlights the correct location when the
+// debugger pauses
+
+// checks to see if the first breakpoint is visible
+function isBreakpointVisible(dbg) {
+  const bpLine = findElement(dbg, "breakpoint");
+  const cm = findElement(dbg, "codeMirror");
+  ok(isVisibleWithin(cm, bpLine), "CodeMirror is scrolled to line");
+}
+
 add_task(function* () {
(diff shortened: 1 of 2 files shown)


public/js/test/mochitest/browser_dbg-editor-select.js
```diff
@@ -6,39 +6,34 @@
...
     "simple1.js", "simple2.js", "long.js"
   );
-  const { selectors: { getSelectedSource }, getState } = dbg;
   const simple1 = findSource(dbg, "simple1.js");
   const simple2 = findSource(dbg, "simple2.js");
 
   // Set the initial breakpoint.
   yield addBreakpoint(dbg, simple1.id, 4);
-  ok(!getSelectedSource(getState()), "No selected source");
+  assertNoSelectedSource(dbg);
```
jlongster: When possible, I'd like to just query the state directly. Creating helpers is good when you have to do a bunch of extra work, but otherwise it forces you to jump around to see what it's actually asserting. Do you mind reverting this change?

One thing that helps is we can probably stop labeling every single assert. It's good in some places, but we can just do `ok(!getSelectedSource(getState()))`. No other testing framework makes you label every single assert, and if it fails it tells you the line number of the failure.

jlongster:
One last minor nit, but thanks for tweaking this. Looks fine otherwise.

public/js/test/mochitest/browser_dbg-editor-select.js
jasonLaster: no problem. I had a similar thought as well. *was on the fence*

jlongster:
Thanks!

PR #629, opened 2016-08-30
A doc to describe React/Redux Structure

JasonWeathersby:
Overview doc with Actions, Reducers, Components explained.

PR #701, opened 2016-09-09
documentationjs for actions

clarkbw:
**for discussion: not ready for merge**

I did some initial work on auto-documenting the actions, you can see the actions doc here:
https://github.com/clarkbw/debugger.html/blob/documentationjs/docs/actions.md

The goal was to help document out the actions as we merge in #629

This uses http://documentation.js.org/ which understands the Flow type and uses that to generate the `params`.  I reviewed jsdoc and a couple other solutions, most of which don't understand Flow or generate Flow from jsdoc which is not what we want.

The command to build out the actions looks like this: 
`documentation.js build -f md -o docs/actions.md public/js/actions/ -g`

I think we would move that into something like `npm run doc-actions`

You can also run with a watch flag that updates as you document:
`node_modules/documentation/bin/documentation.js build -f md -o docs/actions.md public/js/actions/ -g -w`

There are also HTML versions that can be built if we don't want Markdown.


jasonLaster:
This looks great, I'm very excited to give this a try.

I've been thinking for awhile that we should be leveling up our documentation.

It would be great to output a documentation page with a theme like this.
https://github.com/cutting-room-floor/documentation-theme-default

clarkbw:
Yeah, here's the HTML version: (just need to switch the `-f html`)

screen shot 2016-09-09 at 08 12 05

jasonLaster:
* I'd like a way to note private functions.
* is there a good way to split it up by file so that each set of actions `pause`, `breakpoints` is grouped

clarkbw:
> I'd like a way to note private functions.

You can use the jsdoc `@private` to denote private function.  By commenting the function they will appear.

> is there a good way to split it up by file so that each set of actions pause, breakpoints is grouped

I'm trying to use the `@module` docs to make this happen.  It ends up looking kind of like this:

screen shot 2016-09-09 at 12 19 53 pm


> it looks like some utils are pulled in

In that config it wasn't set to `--shallow` which prevents it from spider-ing deeper

clarkbw:
Ok, I've got this in much better shape now.

screen shot 2016-09-09 at 12 51 46

jasonLaster:
Wow, that looks great.

* Do you think it's possible to exclude the built html/md? I believe github might have that option
* I don't think you need the top code section for `actions/pause` etc. maybe a filename would be better if we can do that. Same with the header, I think Pause Actions would read better than `actions/pause`. Same with the toc.
* looks great, ship it.

jbhoosreddy:
This is going to be really useful! Pretty excited about it too!

clarkbw:
Ok, going to force push a new version.  

Any opinions on the docs directory name?  Currently I'm using `docs/actions` however you can add the `public/js/reducers/` directory (which I haven't done yet) and it will merge together the docs.

jlongster:
Maybe something like `docs/reference` because they are auto-generated reference docs? Or just `docs/ref`.

clarkbw:
> Do you think it's possible to exclude the built html/md? I believe github might have that option

I don't think so... I believe i need to commit it all to the docs directory.

> I don't think you need the top code section for actions/pause etc. maybe a filename would be better if we can do that. Same with the header, I think Pause Actions would read better than actions/pause. Same with the toc.

I've tried a number of versions to get that better to read.  The `actions/pause` comes from the module names which all use that format.  Without that top level I haven't found a way to separate the reducers out in a future version.

> looks great, ship it.

public/js/actions/breakpoints.js
```diff
@@ -127,13 +170,15 @@ function toggleAllBreakpoints(shouldDisableBreakpoints: Boolean) {
...
+ *  **NOT IMPLEMENTED**
  *
- * @param object aLocation
+ * @throws {Error} "not implemented"
+ * @memberof actions/breakpoints
+ * @static
+ * @param {Location} location
  *        @see DebuggerController.Breakpoints.addBreakpoint
- * @param string aClients
+ * @param {string} condition
```
jasonLaster: oops

public/js/actions/pause.js
```diff
@@ -151,6 +210,13 @@ function addExpression(expression) {
   };
 }
 
+/**
+ *
+ * @param {object} expression
+ * @param {number} expression.id
```
jasonLaster: at some point this should be in the flow types, but we'll get there

public/js/actions/types.js
```diff
@@ -1,13 +1,49 @@
...
+ * Breakpoint
+ *
+ * @memberof actions/types
+ * @static
+ * @property {string}  id        - Source ID for the file
+ * @property {Location} location - Line number, (zero based?) within the file
+ * @property {boolean}  loading  - Loading status
+ * @property {boolean}  disabled - Disabled by the user
+ * @property {string}  text      - Text representation of relevant line
+ * @property {string}  condition - Conditional breakpoint value (optional)
```
jasonLaster: this feels like some duplication, but it is probably okay

jasonLaster:
awesome, this looks great

public/js/actions/types.js
clarkbw: If you want descriptions you have to use the `@property` identifier, however these might not need it.  Likely just documenting the non-obvious items is only required.

public/js/actions/pause.js
clarkbw: Definitely!

public/js/actions/types.js
jlongster: Yeah, I really don't like these doc strings now that we are using Flow. I was excited to get away from them in the devtools codebase because we have real Flow types. I haven't look at this system at all but I thought it supported flow, meaning that it read flow types? It makes me sad to see our nice and clean types have these comments :(

jlongster: My argument is yet again maintenance: I can't tell you how often these docstrings were out-of-date in the devtools codebase. The flow types will never be though.

public/js/actions/breakpoints.js
clarkbw: There are definitely some old docs hanging around.  I think this will shine more light on it to help get it fixed.  

I ran into this at some point and don't know what to make of it https://github.com/devtools-html/debugger.html/blob/master/public/js/selectors.js#L43-L45

public/js/actions/types.js
jasonLaster: yeah - lets just do the non obvious ones 

>  It makes me sad to see our nice and clean types have these comments :(

felt the same way!

clarkbw: Right, if the attributes are obvious it is definitely excess.

jlongster: If you add a comment at the top of a type, it shows up beside in the docs right? You just write more free-form and say "Represents a breakpoint. `text` is a snippet from the line it was set". I think the rest are pretty self-explanatory.

public/js/actions/breakpoints.js
jasonLaster: I once ran into the number two on a line by itself. 

Kill it.

public/js/actions/types.js
clarkbw: Removed them and I remember now why I tried added them.  There's an issue with documentationjs right now https://github.com/documentationjs/documentation/issues/359 where it doesn't pull in the exported types.  Once that is fixed we should see all the types show up automatically.

clarkbw:
Once this merges into master we should see it up and running here: http://devtools-html.github.io/debugger.html/reference/

jasonLaster:
Looks great. Merging. If there are any other comments we can open a new small focused PR.

PR #703, opened 2016-09-09
Only start sourcemap worker if enabled (fixes memory leak in panel)

jlongster:
This fixes the leak problem for mochitests. I'm not sure why the worker doesn't die with the page, but maybe creating a worker in a chrome context is different somehow.

### DIFF
--- a/public/js/utils/source-map.js
+++ b/public/js/utils/source-map.js
@@ -17,6 +17,18 @@ function restartWorker() {
 }
 restartWorker();
 
+// This is a temporary workaround for the worker not terminating on
+// page load.
+//
+// TODO: Create a proper shutdown function that the panel calls, and
+// make sure in a local debugging context that it dies with the page
+// naturally (as it should).
+if(typeof window !== 'undefined') {
+  window.addEventListener("unload", function() {
+    sourceMapWorker.terminate();
+  });
+}
+
 const sourceMapTask = function(method) {
   return function() {
     const args = Array.prototype.slice.call(arguments);


public/js/utils/source-map.js
```diff
@@ -15,7 +15,11 @@ function restartWorker() {
     getValue("baseWorkerURL") + "source-map-worker.js"
   );
 }
-restartWorker();
+
+// Only start the worker if sourcemaps are enabled.
+if(isEnabled("sourceMaps")) {
+  restartWorker();
+}
```
jasonLaster: hmm, i think it might be better to add a `stopWorker` method which we can call when we end the tests. We will likely work on source maps next sprint and need this then.

```diff
diff --git a/public/js/utils/source-map.js b/public/js/utils/source-map.js
index cf3760f..a6b31d7 100644
--- a/public/js/utils/source-map.js
+++ b/public/js/utils/source-map.js
@@ -8,13 +8,18 @@ const { isEnabled, getValue } = require("../feature");

 let sourceMapWorker;
 function restartWorker() {
-  if (sourceMapWorker) {
-    sourceMapWorker.terminate();
-  }
+  stopWorker();
   sourceMapWorker = new Worker(
     getValue("baseWorkerURL") + "source-map-worker.js"
   );
 }
+
+function stopWorker() {
+  if (sourceMapWorker) {
+    sourceMapWorker.terminate();
+  }
+}
+
 restartWorker();

 const sourceMapTask = function(method) {
@@ -148,5 +153,6 @@ module.exports = {
   getGeneratedSourceId,
   createSourceMap,
   clearData,
-  restartWorker
+  restartWorker,
+  stopWorker
 };
```

jlongster:
We are actually leaking the worker though. It's not just in tests. For some reason the worker is not dying appropriately when the devtools are closed, so we need to fix this everywhere.

The right solution is to expose a way to stop the workers and create a `shutdown` method that the panel calls, but that will require a change to m-c and that can make next week. It takes a day to land things on m-c, and for it to come back around to our testing infra etc so I'd like to land this temporary fix and see if we can still turn on the debugger today.

jasonLaster:
:thumbsup:

PR #704, opened 2016-09-09
clearUserPref instead of setting pref to false when each mochitest fi…

bgrins:
…nishes

This will leave things in the correct state regardless of the default value of the pref


jlongster:
Thanks!

PR #708, opened 2016-09-09
Add Reducer Docs

jasonLaster:
Just followed @clarkbw's lead and added some initial reducer docs. #endoffridayhack 

This adds docs for each of the action types a reducer handles. For example, the tabs reducer has docs for `addTabs` and `selectTab`. I like this because when I think of a reducer's update function I'm often looking to see what actions it handles and how.

screen shot 2016-09-09 at 6 41 15 pm

**Reducer Delegates** I refactored our reducer `update` functions to use a [createReducer](https://github.com/reactjs/redux/blob/master/docs/recipes/ReducingBoilerplate.md#generating-reducers) helper. I like having action handler functions as it helps with documentation, testing... One other approach is to do something similar to [tofino](https://github.com/mozilla/tofino/blob/master/app/ui/browser-modern/reducers/ui.js#L19-L51).


public/js/reducers/tabs.js
```diff
@@ -47,4 +60,5 @@ function getTabId(tab) {
   return id;
 }
 
+const update = setupReducer(initialState, { addTabs, selectTab });
```
jasonLaster: I actually quite like this pattern for a couple of reasons:
* it enumerates the supported handlers in a more succinct fashion than the case statement
* makes it possible to invoke a single handler for testing (maybe useful)
* avoids the smell of one large function
* makes it easy to document each function separately

clarkbw:
This looks awesome! 🎉 

I'd love if you could, as the second person to add these reference docs.  😄  Also add something to the [contributing docs](../blob/master/CONTRIBUTING.md#writing-documentation-book) on how to write up these sections. I can also pick this up next week sometime, I might just be writing this down to remember.

At least the basics of:

* How to declare a module
* How to declare a static function to that module
* How to document the flow sources
* How to add a directory to the docs script (maybe?)

public/js/reducers/tabs.js
jasonLaster: After thinking about it for a couple of days, I'm on the fence between two approaches:

1. the delegator factory
```js
setupReducer(initialState, { addTabs, selectTab });
```

2. an update function with separate handlers
```js
function update(state=InitialState, action)
  switch (action.type) {
    case constants.ADD_TABS:
      return addTab(state, action);
     default:
       return state
  }
```

This second pattern is what [tofino](https://github.com/mozilla/tofino/blob/master/app/ui/browser-modern/reducers/ui.js#L19-L51) does. Benefits:
 * similar benefits with helper functions (documentation, easier to test, ...)
 * more familiar to a larger audience that is used to the the `update/case statement` redux style
 * it's easy to have many action types map to the same helper

As this is a style question, I'd prefer something that is comfortable to a larger audience (option 2) unless we believe enough people would recognize/understand option 1.  I've personally seen a couple projects use option 1, but I don't have a good feeling for how widely adopted it is.

jasonLaster: Lastly, after a little bit of research I found that the redux docs have this helper function [createReducers](https://github.com/reactjs/redux/blob/master/docs/recipes/ReducingBoilerplate.md#generating-reducers) under a section of reducing boilerplate. Given that this is widely documented, I'd opt for using it.

jlongster: It's a good question and comes up a lot in the redux community. However, I think most people end up just using case statements after getting used to it. A few counter-points (sorry, this got a bit long but it's a topic that always comes up in redux so I wanted to be complete):

1. Using separate functions encourages doing more work in a single reducer. A reducer shouldn't get too big, and if it does should be broken into subreducers. The breakpoint and sources reducers are about as big as they should be, which is ~110 lines. The case statements are very easy to manage at that level, it's not really a large function.
2. It is not the same as a large functions scattered across the codebase. The core design of redux works this way and handling multiple actions in a function is good, not a bad design.
3. Separating out functions makes it visually indistinguishable from actions. It's nice to clearly see that you're working in a reducer. You end up having two `addBreakpoint` functions, one action and one reducer. If we do this, we need to use something like tofino's `set` prefix for all the functions for clarity.
4. We definitely want action types to use THIS_FORMAT, but your followup comment points out another technique that still keeps this.
5. By the time you use something like `createReducer`, I think there's more syntax than a simple switch statements, and I'm not sure I see the benefit.
6. You *never* want to test individual reducer handlers. Always just pass state into the reducer itself.
7. The only reason tofino has those functions in that file is because they are called many times by different actions. We do the same thing with `updateText` and other helper functions in the sources reducer. So the pattern is that you pull out things that need to be used in multiple places; tofino just happens to need to do anything else in that reducer.
8. In my opinion, what's more important is what the console is doing, and they use raw switch statements:

https://github.com/devtools-html/gecko-dev/blob/a09e19c97e761b4cf69be92e9ebc40cd6aa8be4d/devtools/client/webconsole/new-console-output/reducers/messages.js

https://github.com/devtools-html/gecko-dev/blob/a09e19c97e761b4cf69be92e9ebc40cd6aa8be4d/devtools/client/webconsole/new-console-output/reducers/filters.js

Especially with smaller reducers with many action types, I think the `switch` statement is a clear winner. We could probably clean up/break up the sources & breakpoints reducer a bit because some of the case handlers are a little large. But I don't want to enforce moving all of the code out of switch statements.

Most importantly, I don't want to merge anything that only changes one reducer. If we do anything, it needs to be done to all reducers at once. We already have other things that are half-implemented, and I don't want more inconsistency within the code.

Let's hold off on this until after the dust settles for turning on the debugger. I'd like to focus on that for the next few days.

jasonLaster:
closing this for now, we can come back to it soon

PR #709, opened 2016-09-12
Add navigation tests

jasonLaster:
Sharing some of the navigation tests early on. I'll add some more later, but wanted to share the progress.


jasonLaster:
This is ready to be reviewed

jbhoosreddy:
These are probably some of the easiest to read/understand tests I've ever seen. Thanks to all the abstraction in `head.js` and also using stuff like redux to drive it. It's completely different from all the old debugger (devtools) tests I've read so far!

public/js/test/mochitest/browser_dbg-navigation.js
```diff
@@ -0,0 +1,30 @@
...
+    "script-switching-01.js"
+  );
+
+  invokeInTab("firstCall");
+  yield waitForPaused(dbg);
+
+  yield navigate(dbg, "doc-iframes.html");
+  yield waitForPaused(dbg);
+  assertPausedLocation(dbg, "iframes.html", 8);
+  ok(countSources(dbg) == 1);
```
jlongster: Might be nice to use `doc-scripts.html` which has a few more scripts. Testing with multiple scripts makes sure this isn't a fluke.

public/js/test/mochitest/browser_dbg-navigation.js
```diff
@@ -0,0 +1,30 @@
...
+
+  yield navigate(dbg, "about:blank");
+  yield waitForDispatch(dbg, "NAVIGATE");
+  ok(countSources(dbg) == 0);
+});
```
jlongster: I'm sure you're planning on expanding this test, but you could go ahead and navigate back to `doc-scripts.html` and assert that there are 3 sources again.

public/js/test/mochitest/head.js
```diff
@@ -234,6 +236,15 @@ function reload(dbg) {
   return dbg.client.reload();
 }
 
+function navigate(dbg, url, ...sources) {
+  const response = dbg.client.navigate(url);
+  if (sources.length == 0) {
+    return response;
```
jlongster: Do we need this `if` statement? You return the response here, but if you are waiting on sources you'll get a different return value. It'd be nice if this function had a consistent return value, and `waitForSourcesToLoad` will do nothing if you pass it an empty array anyway.

jlongster:
Looks great, just a few nits.

@jbhoosreddy Yeah, the previous mochitests were simply awful. Pulling everything out of the DOM and only checking it when needed allows us to write way simpler tests. The old tests were also written at a time where promises weren't even used.

PR #711, opened 2016-09-12
Propery clean up the editor and sourcemap worker on shutdown

jlongster:
This should fix all our testing problems. The editor is properly cleaned up (required because the mozilla-central needs to do some cleanup or else it will leak) and the sourcemap worker is terminated. We expose a new `destroy` method which is called from the panel.

### DIFF
--- a/public/js/components/Editor.js
+++ b/public/js/components/Editor.js
@@ -156,6 +156,11 @@ const Editor = React.createClass({
     }
   },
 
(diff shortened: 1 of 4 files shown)


clarkbw:
> This should fix all our testing problems

Strong words 😺

jlongster:
> Strong words 😺

I WANT TO BELIEVE

PR #713, opened 2016-09-12
Add ability to highlight a specific line in the editor

jlongster:
This is adds back the ability to highlight a line. I like that the debug line was implemented separately, because the "debug" line and the "highlighted" line really are two different things. The former is permanent and should always exist, even if I'm going back and forth between sources. The latter is transient and just moves the user to a specific location, but goes away.

The debug line also doesn't have any "behavior", meaning it doesn't grab the user's attention. That's the purpose of the "highlighted" line, so in the new implementation the debug line will be highlighted but the `pause` action actually moves the user there by specifying a "highlighted" line as well. I don't know if any of that makes sense.

Basically, the `selectSource` action now takes a `{ line: N }` parameter which tells the UI to move to the line specifically. When moving there, it will flash:

foo

I'm open to suggestions on this implementation. I'm not in love with some of this code.


jlongster:
Also up for debate is the highlighted line color and how long the fade should last. @helenvholmes might be able to help. Right now it fades for 1s because any shorter and it's too easy to miss.

public/js/components/Editor.js
```diff
@@ -173,14 +186,30 @@ const Editor = React.createClass({
...
+      if (this.props.location) {
+        this.clearHighlightLine(this.props.location.line);
+      }
+      this.clearHighlightLine(nextProps.location.line);
+
+      const line = nextProps.location.line;
+      this.editor.alignLine(line);
+      this.setHighlightLine(nextProps.location.line);
+      this.editor.codeMirror.addLineClass(line - 1, "line", "highlight-line");
+    }
```
jasonLaster: This looks like a good function that we could write unit tests for. 

We don't have a great setup now for writing component unit tests, but it's not terrible either. If you did write them, it would be possible to run them with the `mocha-server` and test passing different props down.

public/js/lib/devtools-sham/shared/event-emitter.js
```diff
@@ -117,15 +117,15 @@ EventEmitter.prototype = {
...
-          //console.error(msg);
-          console.log(msg);
-        }
+        // }
+        // catch (ex) {
+        //   // Prevent a bad listener from interfering with the others.
+        //   let msg = ex + ": " + ex.stack;
+        //   //console.error(msg);
+        //   console.log(msg);
+        // }
```
jasonLaster: Was this intentional? If so, maybe delete it. 

Also, what are the places where we are still using event listeners? My understanding was that this was something we do in the old debugger as a crutch because not everything was converted to Redux. Can we phase this out?

public/js/reducers/sources.js
```diff
@@ -21,7 +21,7 @@ export type SourcesState = {
 
 const State = makeRecord(({
   sources: I.Map(),
-  selectedSource: undefined,
+  selectedLocation: undefined,
```
jasonLaster: This is a nice improvement. There are many things I like here:
+ we're including the selected line
+ using a known type (location)
+ no longer copying the full source object.

public/js/reducers/sources.js
```diff
@@ -54,19 +54,19 @@ function update(state = State(), action: Action) : Record<SourcesState> {
...
 
     case "CLOSE_TAB":
-      return state.merge({
-        selectedSource: getNewSelectedSource(state, action.id),
-        tabs: removeSourceFromTabList(state, action.id)
-      });
+      return state.merge({ tabs: removeSourceFromTabList(state, action.id) })
+        .set("selectedLocation", {
+          sourceId: getNewSelectedSource(state, action.id)
+        });
```
jasonLaster: out of curiosity, why not keep this as one merge? I'm probably missing something :)

public/js/reducers/sources.js
```diff
@@ -102,7 +102,8 @@ function update(state = State(), action: Action) : Record<SourcesState> {
       return _updateText(state, action, [action.originalSource]);
 
     case "NAVIGATE":
-      const source = state.selectedSource;
+      const sourceId = state.selectedLocation.sourceId;
+      const source = getSource({ sources: state }, sourceId);
```
jasonLaster: It's a shame the selectors accept outerstate...

public/js/reducers/sources.js
```diff
@@ -146,18 +147,18 @@ function removeSourceFromTabList(state, id) {
 /*
  * Adds the new source to the tab list if it is not already there
  */
-function updateTabList(state, source, options) {
+function updateTabList(state, source, tabIndex) {
   const tabs = state.get("tabs");
-  const selectedSource = state.get("selectedSource");
+  const selectedSource = getSelectedSource({ sources: state });
```
jasonLaster: Would you want to do the same thing above to get the selected source?

public/js/reducers/sources.js
```diff
@@ -146,18 +147,18 @@ function removeSourceFromTabList(state, id) {
...
   const sourceIndex = tabs.indexOf(source);
   const includesSource = !!tabs.find((t) => t.get("id") == source.get("id"));
 
   if (includesSource) {
-    if (options.position != undefined) {
+    if (tabIndex != undefined) {
       return tabs
         .delete(sourceIndex)
-        .insert(options.position, source);
+        .insert(tabIndex, source);
```
jasonLaster: I believe there will be some failing unit tests that test this behavior as well.

public/js/reducers/sources.js
```diff
@@ -171,11 +172,11 @@ function updateTabList(state, source, options) {
  */
 function getNewSelectedSource(state, id) : ?Source {
```
jasonLaster: this method should probably be renamed to `getNewSelectedSourceId`

public/js/components/Editor.js
```diff
@@ -36,7 +36,7 @@ function resizeBreakpointGutter(editor) {
 const Editor = React.createClass({
   propTypes: {
     breakpoints: ImPropTypes.map.isRequired,
-    selectedSource: ImPropTypes.map,
+    location: PropTypes.object,
```
jasonLaster: I could see `selectedLocation` being slightly clearer. There are many locations in the editor, breakpoint locations, selected gutter lines, ... I don't feel strongly though

jlongster: I think you're right. I can use that instead.

public/js/reducers/sources.js
jlongster: `npm run test` runs OK so I'll take a look and make sure all the tests are running.

public/js/reducers/sources.js
jlongster: Unfortunately, `merge` basically calls `toJS` and deeply converts everything to an immutable object. I wanted to avoid more `getIn` calls, because eventually we'll be using Records more thoroughly so we could just use `merge` here but it would still be `selectedLocation.line` elsewhere. So this is temporary until we figure out how to use Records better.

But now that it's implemented, I can look over and see if it's simpler to keep it as immutable.

public/js/reducers/sources.js
jasonLaster: sorry, i remembered writing some specific tabs tests but it must have been in the sources tests

https://github.com/devtools-html/debugger.html/blob/master/public/js/actions/tests/sources.js#L88-L97

public/js/lib/devtools-sham/shared/event-emitter.js
jlongster: I'm not sure, it was something from the client stuff. I'm not sure what the implications are; I get the desire to make sure that all event listeners are fired, and if one errors it doesn't stop the whole process. Eventually we will need to share this code exactly as it exists on m-c, so we could have a discussion for changing it there. Until then I think I'll just revert this.

public/js/components/Editor.js
jlongster: I'll try to add a check in the `editor-select` test at least. I think if you are running it in a browser and checking it there, it's not much more work to just add some code to an existing integration test. But we can experiment with that at some point.

jlongster:
All good points, thanks! I'll have a new PR soon.

public/js/components/Editor.js
jasonLaster: That's a good point. I have no problem writing more integration tests and then converting some of them to unit tests later.

jlongster:
@jasonLaster Here is my `Editor` refactoring. I tried to add comments to explain things. But the idea is to group the code according to the workflows, instead of grouping it by functionality (i.e. being explicit what we are doing with `loading` and `editor` text).

Probably still need to polish some things, but it seems to be working. This is some complicated stuff and I think it's OK the editor has a bunch of complicated code it in. It should all be contained here and the rest of the system doesn't have to care about it. (The whole thing is still only 230 lines long).

I'll add some more tests and fix any errors that come up.

jlongster:
It might be easiest to view the file directly: https://github.com/devtools-html/debugger.html/blob/highlight-line/public/js/components/Editor.js

jlongster:
This is ready to be merged once tests pass and it's reviewed.

public/js/actions/pause.js
```diff
@@ -33,24 +33,19 @@ function resumed() {
...
-          selectedFrameId: frame.id
-        };
-      })()
+      pauseInfo: { why, frame },
+      frames: frames,
+      selectedFrameId: frame.id
     });
+    dispatch(selectSource(frame.location.sourceId,
+                          { line: frame.location.line }));
   };
```
jlongster: I changed this to be explicit about order. I want the `selectSource` action to be fired after the pause state exists so that I can use the `selectedFrame` state in the editor. Otherwise, it will highlight that line grey for a second and then the paused state will come in and it will be highlighted green.

We never need to do any optimistic updates from the paused action anyway. Usually the `[PROMISE]` pattern is used for stuff that is user-invoked so that we can give immediate feedback.

jlongster:
Oh, nm, I also need to fix up flow.

public/js/actions/pause.js
jasonLaster: This is much cleaner.

public/js/components/Editor.js
```diff
@@ -71,45 +71,53 @@ const Editor = React.createClass({
...
+  },
 
-    this.setText(newSourceText.get("text"));
-    this.setMode(newSourceText);
-
-    resizeBreakpointGutter(this.editor.codeMirror);
+  highlightLine() {
+    // If the location has changed and a specific line is requested,
+    // move to that line and flash it.
+    if (this.pendingJumpLine) {
```
jasonLaster: this is a good candidate for a guard statement. It would more clearly say, don't highlight a line unless we're jumping

```js
if (!this.pendingJumpLine) {
  return;
}
```

public/js/test/mochitest/head.js
```diff
@@ -198,12 +198,16 @@ function findSource(dbg, url) {
   return source.toJS();
 }
 
-function selectSource(dbg, url) {
+function selectSource(dbg, url, line) {
```
jasonLaster: maybe add a default value `line = 0`

public/js/test/mochitest/head.js
```diff
@@ -198,12 +198,16 @@ function findSource(dbg, url) {
...
   const source = findSource(dbg, url);
-  dbg.actions.selectSource(source.id);
+  const prevSelectedSource = dbg.selectors.getSelectedSource(dbg.getState());
+  dbg.actions.selectSource(source.id, { line });
 
-  return waitForDispatch(dbg, "LOAD_SOURCE_TEXT");
+  if(!prevSelectedSource ||
+     prevSelectedSource.get("id") !== source.id) {
+    return waitForDispatch(dbg, "LOAD_SOURCE_TEXT");
+  }
```
jasonLaster: shouldn't this be handled by `selectSource`

public/js/components/Editor.js
```diff
@@ -162,24 +170,49 @@ const Editor = React.createClass({
...
 
-    // Highlight the paused line if necessary
-    if (isSourceForFrame(nextProps.selectedSource, nextProps.selectedFrame)) {
-      this.setDebugLine(nextProps.selectedFrame.location.line);
+    // Only update and jump around in real source texts. This will
+    // keep the jump state around until the real source text is
+    // loaded.
+    if (this.props.sourceText && isTextForSource(this.props.sourceText)) {
+      this.updateDebugLine(prevProps, this.props);
+      this.highlightLine();
```
jasonLaster: I don't want to add an unnecessary headache, but is it possible that we've registered a pending location, and then another source is selected so we inadvertantly jump to the wrong line?

I'm going to guess no, because we'll have updated the pending jump line with the new line...

public/js/components/Editor.js
```diff
@@ -71,45 +71,53 @@ const Editor = React.createClass({
...
+      const line = prevProps.selectedFrame.location.line;
+      this.editor.codeMirror.removeLineClass(line - 1, "line", "debug-line");
     }
-
-    if (newSourceText.get("error")) {
-      this.setText("Error");
-      console.error(newSourceText.get("error"));
-      return;
+    if (nextProps.selectedFrame) {
+      const line = nextProps.selectedFrame.location.line;
```
jasonLaster: it would be nice to have a `debugLine` function or prop passed down from connect. We're going to have lots of locations and will want to be comparing them, so we should have language around debug lines, highlight lines, jump lines... With highlight lines / jump lines m guess is that at some point we don't know if a jump line is a highlight line, it might just be a debug line... *(might not make sense)*

public/js/test/mochitest/browser_dbg-editor-select.js
```diff
@@ -48,5 +48,18 @@ add_task(function* () {
...
+  yield addBreakpoint(dbg, longSrc.id, 16);
+
+  // Jump to line 16 and make sure the breakpoint is visible. We only
+  // added the breakpoint so we could make sure it scrolled correctly.
+  yield selectSource(dbg, longSrc.url, 16);
+  isElementVisible(dbg, "highlightLine");
+  yield selectSource(dbg, longSrc.url, 17);
+  yield selectSource(dbg, longSrc.url, 18);
+  is(findAllElements(dbg, "highlightLine").length, 1,
+     "Only 1 line is highlighted");
```
jasonLaster: Are we testing the 
* loading cases, e.g. you're paused, select a call frame of a large source and the source hasn't loaded yet?
* case where you add a breakpoint to a source you're already paused on

jasonLaster:
Looks good! Lets land this :)

public/js/test/mochitest/browser_dbg-editor-select.js
jlongster: That's a good point. I just added more to the test. The first point is particularly important.

public/js/components/Editor.js
jlongster: I'm not quite convinced; I'd like to wait and maybe do it later. I think it's good for the components to use terms that are more concrete like `selectedFrame` generally instead of receiving things like `debugLine` from somewhere else. I think those terms are a little vague, with highlighted/debug/etc being sort of similar. We can see how it feels over time.

public/js/components/Editor.js
jlongster: You're right. There's a chance if we've selected a difference source, and its line is undefined, but we were in the middle of loading another source so we hadn't jumped yet, that it will inadvertently jump in the new source when it shouldn't. My new patch guards against this.

public/js/test/mochitest/head.js
jlongster: That would be wrong though; `line` should be `undefined` when you aren't selecting a line and you don't want to jump. When we maintain scroll positions, it'll just render the source in the place where it was and shouldn't jump anywhere.

public/js/components/Editor.js
jlongster: Done

jlongster:
Cleaned up several things. I think once your navigation tests land I can clean up this test a little bit; reloading is weird because the promise will resolve before the debugger is finished initializing so you need to wait explicitly on somethings like the source text being loaded if you need it. We could probably make that cleaner.

This is ready to go, I think. I added another small test for making sure that you can select a source and line and the debugger will appropriately go there, even if the source doesn't exist yet. This tests the whole `pending` state workflow.

public/js/test/mochitest/head.js
```diff
@@ -147,8 +147,12 @@ const waitForPaused = Task.async(function* (dbg) {
 });
 
 function waitForSources(dbg, ...sources) {
+  if(!sources.length) {
+    return Promise.resolve();
+  }
+
   info("Waiting on sources: " + sources.join(", "));
-  yield Promise.all(sources.map(url => {
+  return Promise.all(sources.map(url => {
```
jasonLaster: nice - one of us is going to land this function first :)

public/js/test/mochitest/browser_dbg-editor-pending-select.js
```diff
@@ -0,0 +1,34 @@
...
+
+add_task(function* () {
+  const dbg = yield initDebugger("doc-scripts.html");
+  const { selectors: { getSelectedSource }, getState } = dbg;
+  const sourceUrl = EXAMPLE_URL + "code-long.js";
+
+  // The source probably hasn't been loaded yet, and using
+  // `selectSourceURL` will set a pending request to load this source
+  // and highlight a specific line.
+  dbg.actions.selectSourceURL(sourceUrl, { line: 66 });
```
jasonLaster: nit: it'd be nice to move this to a helper function at the top of the test. It's likely that we'll want to extract this function late and this will help do that.

public/js/test/mochitest/browser_dbg-editor-pending-select.js
```diff
@@ -0,0 +1,34 @@
...
+
+  // Wait for the source text to load and make sure we're in the right
+  // place.
+  yield waitForDispatch(dbg, "LOAD_SOURCE_TEXT");
+  is(getSelectedSource(getState()).get("url"), sourceUrl);
+  const line = findElement(dbg, "highlightLine");
+  ok(line, "Line is highlighted");
+  ok(line.innerHTML.match(/this.*todos.*filter/),
+     "The correct line is highlighted");
```
jasonLaster: it'd be nice to have an `assertHighlightLocation` we should be testing this in other places as well e.g. selecting breakpoints, frames, searching...

public/js/test/mochitest/browser_dbg-editor-pending-select.js
jlongster: I'm not sure why we'd wrap every single action in a helper function. What does that buy us? (just curious)

public/js/test/mochitest/browser_dbg-editor-pending-select.js
jlongster: That's a good point. Done.

jlongster:
Latest commit only tweaks a few more things. If the tests pass, I'd like to go ahead and merge because we need to get a try run going if I'm going to land a bundle update tonight.

public/js/test/mochitest/head.js
jasonLaster: nice point

public/js/test/mochitest/browser_dbg-editor-pending-select.js
jasonLaster: its mostly a nit because it *doesn't* make a big difference in this case. We've done a good job of keeping the test fairly high level.

jasonLaster:
sounds good!

jlongster:
Alright, squashing this and merging!

Issue #652, opened 2016-09-01
File Search - fails when hitting enter right away


jasonLaster:
I've noticed a couple times, because i do this a lot, that typing and then hitting enter fails.

I think the default behavior should be that the first item is selected as that's usually what you want.

PR #714, opened 2016-09-13
default to selecting the first item in the search list

clarkbw:
With this change the first item in the search list will always be selected such that if you only have one file matching your query you can open it by hitting enter instead of navigating to it first.

I added an `INITIAL_SELECTED_INDEX` const to handle the two locations where the initial selected index needs to be set.  Though it felt like it might be overkill given the situation the default param isn't really necessary for this to work properly as the other value will be changed as soon as there is typing activity.

Multiple items:
screen shot 2016-09-13 at 13 59 41

Only one item:
screen shot 2016-09-13 at 13 59 51

### DIFF
--- a/public/js/components/Autocomplete.js
+++ b/public/js/components/Autocomplete.js
@@ -5,6 +5,8 @@ const classnames = require("classnames");
 require("./Autocomplete.css");
 const Svg = require("./utils/Svg");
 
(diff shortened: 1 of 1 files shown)


clarkbw:
I didn't notice before but there is an issue for this, #652

jasonLaster:
haha - looks great.

PR #721, opened 2016-09-14
add mochitest instructions for windows users

jbhoosreddy:
This adds instructions for Windows users to set up their mochitest environment and running their tests.
@jasonLaster this was what I was talking about.


jasonLaster:
Oh nice. Sorry, wasn't sure what was involved here :)

PR #723, opened 2016-09-14
uses the local install of webpack

clarkbw:
If you don’t have web installed globally this script fails, this change uses the local webpack location.

### DIFF
--- a/bin/make-firefox-bundle
+++ b/bin/make-firefox-bundle
@@ -26,7 +26,7 @@ fi
     fi
 );
 
-TARGET=firefox-panel webpack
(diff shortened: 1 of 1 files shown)


jasonLaster:
nice catch

jbhoosreddy:
I saw that but I ended up installing webpack globally. Nice point!

jlongster:
We shouldn't be going into the webpack module directly. The official binaries are installed in `node_modules/.bin`, so we should use `node_modules/.bin`. @clarkbw Can you open a new pull request to fix that?

PR #724, opened 2016-09-14
use installed version of webpack

jasonLaster:
Jumped the gun on #723. Here's an updated version


jlongster:
Thanks!

clarkbw:
Oh nice!  Thanks!

PR #728, opened 2016-09-14
Fix navigation test

jasonLaster:


public/js/test/mochitest/head.js
```diff
@@ -237,8 +237,11 @@ function reload(dbg) {
 }
 
 function navigate(dbg, url, ...sources) {
-  const response = dbg.client.navigate(url);
-  return waitForSourcesToLoad(dbg, sources)
+  return Task.spawn(function* () {
+    yield dbg.client.navigate(url);
+
+    return waitForSourcesToLoad(dbg, sources);
```
jlongster: You don't have to make these changes now that we know it's the other part that needs fixing. We can wait to see if we need to do this.

jlongster:
@jasonLaster Mind rebasing this?

PR #730, opened 2016-09-14
Inline test Tasks

jasonLaster:
Honestly, this is mostly driven from my editor not finding these two functions when they're assignments.

With that said, I think it's nice for the functions not to be decorated in this way. But it's just style...


jlongster:
I'm fine with this. Others on the team would prefer the other style probably though. But it's fine for now.

PR #733, opened 2016-09-14
Wait on sources

jasonLaster:
This also requires a small fix on mc's panel.js but the main idea is here.

### DIFF
--- a/public/js/clients/firefox.js
+++ b/public/js/clients/firefox.js
@@ -109,7 +109,7 @@ function initPage(actions) {
   // makes the server iterate over them and fire individual
   // `newSource` notifications. We don't need to do anything with the
   // response since `newSource` notifications are fired.
-  threadClient.getSources();
+  return threadClient.getSources();
 }
 
 module.exports = {
--- a/public/js/main.js
+++ b/public/js/main.js
@@ -103,8 +103,8 @@ if (connTarget) {
     bootstrap: ({ threadClient, tabTarget }) => {
       firefox.setThreadClient(threadClient);
       firefox.setTabTarget(tabTarget);
-      firefox.initPage(actions);
       renderRoot(App);
+      return firefox.initPage(actions);
     },
     destroy: () => {
       unmountRoot();


jlongster:
Sounds good. So this works? Does this mean we could `renderRoot(App)` ahead of the `startDebugging` calls too? If you wanted, we could still make `startDebugger` block on `initPage` there as well, if you want to be more consistent.

This looks like a good direction though; let's wait to merge until I cut a new release (tonight hopefully) just in case something goes wrong.

jasonLaster:
> Does this mean we could renderRoot(App) ahead of the startDebugging calls too? 
Yes. 

> let's wait to merge until I cut a new release
Sounds good.

jlongster:
Can you apply the changes from https://bugzilla.mozilla.org/show_bug.cgi?id=1302862 to your local Firefox repo, and run all the tests with this change locally? Just want to make sure there's not some clear breakage. We have to land this chance on m-c first, and then land the above bug, but if anything is wrong we won't see failures until we try to land the above bug, so it's worth trying to catch anything early.

jlongster:
Also, something you could do to make sure this is behaving how we want: in one of the tests that needs sources, remove them from the `initDebugger` call and run the test several times. If it passes, it is indeed waiting for the sources to be loaded.

jasonLaster:
That's a good idea, i've been running the change locally, but without the changes to `initDebugger`. Will do that now

jlongster:
I bet DAMP (the tool that tracks Firefox performance) is going to show a decrease in perf in loading, too, because it tracks the `open` method of the panel. But that's ok... it's not really a decrease in perf, it's just moving around the events.

jasonLaster:
I think this is fine to merge as is, but I was not able to get the tests to work when i removed the waitForSources.

The reason for that is that even though the console, in theory, will block on the debugger opening, other commands likes `addBreakpoint` will not block on `initPage`. Because of this, I think it's good to land this, possibly land the `m-c` change which needs this update to work, and then re-evaluate the tests.

jlongster:
(I'm very curious about this too so I'm happy to try it out tomorrow and see why the tests aren't working)

jasonLaster:
Cool. I think you have a good point about init debugger.

Happy to have it off. I'm in the middle of code search.

jlongster:
Didn't have time to look at this more today. It's not a bugfix though so it's not too urgent. I'll take a look next week.

jlongster:
These changes work for me. I applied this patch and made the `open` method of the panel block on the promise returned from `bootstrap`, and ran the `browser_dbg-editor-select.js` tests without explicitly waiting on sources and it passes. I used `--run-until-failure` and it never failed. I tweaked the `open` method to *not* block on `bootstrap` and the test immediately fails. (It's not a hard racy condition to see)

This verifies that this fix works and we should be able to remove waiting on sources. Note that you will still need to `waitForDispatch("LOAD_SOURCE_TEXT")` if you want to access the editor, but I think that makes sense.

We'll need to land this change, land a bundle update, and then land the m-c tweak. I'll see if everything looks good to land a bundle update today.

jasonLaster:
@jlongster great.

PR #735, opened 2016-09-14
Add toolbox favicon. Fixes #412

### LINKED ISSUE #412: Toolbox Favicon
It'd be fun if we had a toolbox favicon. To be clear, this is not a user thing, but just a nice polish touch for developers who are working in local dev mode.

----

### 11/29 update

We had a favicon, which was broken when we moved the index.html to the toolbox. Lets fix it!

Dalimil:
I've added the suggested favicon. I used the png file format to support colour transparency.


jasonLaster:
Thanks @Dalimil!

PR #736, opened 2016-09-15
Fix console test

jlongster:
The console test was 404ing because it wasn't using the right example page. Also, you don't need to wait on any sources if you're not going to actually test them. You only need to list ones you care about.


jlongster:
This is a very small fix that definitely fixes the test, so going ahead and merging.

PR #738, opened 2016-09-15
update mochitest docs

jbhoosreddy:
I'm making small updates to the docs as they are a little outdated. I noticed that API list is not exhaustive. If you'd like I can update that in this PR also.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -48,7 +48,7 @@ The debugger.html project is a JavaScript debugger built from the ground up usin
 
 ### devtools.html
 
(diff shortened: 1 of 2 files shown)


clarkbw:
doc changes look good!

jlongster:
Thanks! Sure, feel free to add the new APIs here too.

jbhoosreddy:
I took a pass at updating the API description in the docs. I broke it down to separate headings so that it is more manageable and easy to understand.

jbhoosreddy:
@jlongster would it be okay to merge this PR first. It makes three different changes to different docs. It would be additional work to break it down again.

#750 would end up removing the API documentation and simply refer to the docs.

PR #720, opened 2016-09-13
fix flag for windows

MoLow:
after a few hours trying to figure out why I cant run this on a windows machine,  I found these changes will make it work.

 BTW, the tests cant even start on windows, so I couldent run them locally either :(

PR #741, opened 2016-09-15
add a table of contents to the remote debuggable targets doc

clarkbw:
This came out of #720

### DIFF
--- a/docs/remotely-debuggable-browsers.md
+++ b/docs/remotely-debuggable-browsers.md
@@ -1,5 +1,26 @@
 ## Remotely debuggable browsers
 
+#### Table Of Contents
+
+[Firefox](#firefox)
+  * [MacOS](#macos)
+    * [Firefox](#firefox-release)
+    * [Firefox Beta](#firefox-beta)
+    * [Firefox Developer Edition](#firefox-developer-edition-aurora)
+    * [Firefox Nightly](#firefox-nightly)
+  * [Windows](#windows)
+    * [Firefox (all versions)](#firefox-all-versions)
+    * [Firefox (64 bit)](#firefox-64-bit)
+  * [Android](#android)
+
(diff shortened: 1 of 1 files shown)


jlongster:
I like it. Thanks! The failure is only the console test that was fixed.

docs/remotely-debuggable-browsers.md
```diff
@@ -1,5 +1,29 @@
 ## Remotely debuggable browsers
 
+#### Table Of Contents
+
+[Firefox](#firefox)
+  * [MacOS](#macos)
+    * [Firefox](#firefox-release)
+    * [Firefox Beta](#firefox-beta)
+    * [Firefox Developer Edition](#firefox-developer-edition-aurora)
+    * [Firefox Nightly](#firefox-nightly)
```
jlongster: Is it worth having all these sections? The only difference between any of them is the name of the app. This makes it looks like there are big differences between all of them to run this. I'd personally prefer to just have a `MacOS` section and use `Firefox.app` by default and then list the different binary names for different versions and make a note that they made need to use one of those.

jasonLaster:
Looks good!

docs/remotely-debuggable-browsers.md
clarkbw: I agree it is repetitive.  How about I leave two commands for for Firefox and Nightly to make it less repetitive and have notes about DE and Beta.  I think you said once how you like being able to copy the command directly from the doc is nice and I would definitely miss those two versions.  Plus it keeps it similar to the Chrome listing.

jlongster: Sounds good to me!

jlongster:
@clarkbw Thanks for updating! Ready to be merged, right?

jasonLaster:
merging with a squash. looks good

PR #744, opened 2016-09-15
replace CRLF with LF according to editorconfig

johannestroeger:


jlongster:
Sounds good to me, thanks. Too bad it needs to rewrite the whole file, but it does need fixing.

jlongster:
That test failure was fixed with #736.

PR #746, opened 2016-09-15
Move some of editor-select.js test into editor-highlight.js to make tests more focused

jlongster:
I saw a timeout of the editor-select.js test on linux debug. Linux debug on try runs on very slow machines, which in fact the previous debugger tests couldn't even run on (we had to disable them a while back). This is why we tend to write smaller mochitests and tend to break them up into -01, -02, etc. Not saying we need to totally do that, but if we can split them up some, we will avoid timeouts there.

I renamed the `editor-pending-select.js` test to focus on highlighting a specific line and move some of the `editor-select.js` to that file. Now one tests stepping through files and making sure the right source/debug line exists, and the other tests highlighting line.

The latest try run includes this, so we need to merge this in before I push out an update: https://treeherder.mozilla.org/#/jobs?repo=try&revision=be03ec805b36


jlongster:
(Moving the second half of the `editor-select` test also avoids having to do any reloads, because the only reason I did that was to clear out the state)

public/js/test/mochitest/browser_dbg-editor-highlight.js
```diff
@@ -0,0 +1,47 @@
...
+  yield selectSource(dbg, "long.js", 18);
+  assertHighlightLocation(dbg, "long.js", 18);
+  is(findAllElements(dbg, "highlightLine").length, 1,
+     "Only 1 line is highlighted");
+
+  // Test jumping to a line in a source that exists but hasn't been
+  // loaded yet.
+  yield waitForSources(dbg, "simple1.js");
+  const simple1 = findSource(dbg, "simple1.js");
+  selectSource(dbg, simple1.url, 6);
```
jasonLaster: nit: you can use `selectSource(dbg, "simple1.js", 6)` as selectSource uses find source under the hood.

jlongster: True, I changed it. I need to get the real source to use in selectors below, but I moved the `findSource` down to where I need it.

jlongster:
> I wonder if it's possible to use multiple add_task calls in one file.

I'm pretty sure you can, but I'm not sure if try will set up a new timeout for each `add_task`. If it does, that would be neat. It might kill a whole file after it runs for a while though and make it timeout, even if there are multiple `add_task`s.

PR #747, opened 2016-09-15
Make `download-firefox-artifact` properly clean up so it can update Firefox

jlongster:
There are a few ways we could improve the scripts, but the biggest pain right now is `prepare-mochitests-dev` doesn't actually make sure everything is up-to-date. If will fail to update Firefox after the first run because `hg pull` will fail because there are local changes. This cleans up the repo to make sure there are no local changes.


jasonLaster:
👍 thanks. this has burned me a couple of times

PR #748, opened 2016-09-15
Add test for making sure selected source persists across reloads

### LINKED ISSUE #646: auto-select preferred source [mochitests]

browser_dbg_reload-preferred-script-02.js
browser_dbg_reload-preferred-script-03.js
browser_dbg_reload-same-script.js

[DXR search](https://dxr.mozilla.org/mozilla-central/search?q=path%3Adevtools%2Fclient%2Fdebugger%2Ftest%2Fmochitest%2F+file%3A*dbg_reload-*&redirect=true)

jlongster:
This will close #646. Adds a few assertions to the navigation test to make sure the selected source is correct after reloads.

### DIFF
--- a/public/js/test/mochitest/browser_dbg-navigation.js
+++ b/public/js/test/mochitest/browser_dbg-navigation.js
@@ -11,15 +11,13 @@ function countSources(dbg) {
  * navigating while paused will reset the pause state and sources
  */
 add_task(function* () {
-  const dbg = yield initDebugger(
-    "doc-script-switching.html",
-    "script-switching-01.js"
-  );
+  const dbg = yield initDebugger("doc-script-switching.html");
+  const { selectors: { getSelectedSource, getPause }, getState } = dbg;
 
   invokeInTab("firstCall");
   yield waitForPaused(dbg);
 
-  yield navigate(dbg, "doc-scripts.html", "simple1.js", "long.js");
+  yield navigate(dbg, "doc-scripts.html", "simple1.js");
   yield addBreakpoint(dbg, "simple1.js", 4);
   invokeInTab("main");
   yield waitForPaused(dbg);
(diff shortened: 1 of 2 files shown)


public/js/test/mochitest/browser_dbg-navigation.js
```diff
@@ -11,15 +11,13 @@ function countSources(dbg) {
  * navigating while paused will reset the pause state and sources
  */
 add_task(function* () {
-  const dbg = yield initDebugger(
-    "doc-script-switching.html",
-    "script-switching-01.js"
-  );
+  const dbg = yield initDebugger("doc-script-switching.html");
```
jlongster: You don't ever explicitly use `script-switching-01.js`, so you don't have to wait for it :)

jlongster:
I've retriggered this 3 times so far so if it pauses I think it is good.

public/js/test/mochitest/browser_dbg-navigation.js
```diff
@@ -39,4 +38,10 @@ add_task(function* () {
   );
 
   is(countSources(dbg), 4, "4 sources are loaded.");
+
+  // Test that the current select source persists across reloads
+  yield selectSource(dbg, "long.js");
+  yield reload(dbg, "long.js");
+  ok(getSelectedSource(getState()).get("url").includes("long.js"),
```
jasonLaster: this makes me think you might want to check the scroll location. it's likely we want to persist that as well.

It might be nice to move this to a general check `assertLocation(dbg, "long.js")` in addition to checking the source is selected we should double check that active tab is long.js

jlongster: Not sure I'd really care about scroll location here; I don't think we want to persist those across reloads. That seems tricky because the source may have changed a good bit. I don't think Chrome does that, right? We probably want to just save the current tab URLs.

I think I might go ahead and merge as is, and if we do really want to test more stuff we can move this into a helper functions. The pause and highlight locations are nice because there's a bunch of stuff to test, but for other things I'm OK querying the state directly. Other tests will make sure that the current selected source has the right tab and everything.

(Definitely still feeling this out though so we may abstract it out later...)

PR #749, opened 2016-09-15
Properly handle breaking in a script before the debugger is initialized

jlongster:
This is a little messy and definitely not finalized, but I wanted to describe a very edge case problem I hit and a change we need to make.

See https://bugzilla.mozilla.org/show_bug.cgi?id=1303177, as this requires an m-c change to land.

The problem is breaking in an eval source before the debugger is initialized. The above bug goes into it in more detail.

The core of the problem, besides that m-c patch, is that we do need to actually handle the sources returned from `getSources` unfortunately. There are two reasons:

* If loading the page from the bfcache, we won't get `newSource` events. We'll just get the current sources in `getSources` because they already exist in memory.
* The thread is attached when the devtools opens. So there may be things that happen before the debugger is even initialized that loads some sources, like hitting a `debugger` statement. That will force that source to be created, and *then* the debugger will be initialized. At that point, we won't get a `newSource` notification for that event because it already exists, so the new debugger won't know about it, and the debugger is in a broken state: paused, but no source to show.

Definitely went down a rabbit whole debugging this one... Came up while writing some break tests. I'm going to revamp this patch tomorrow, so you don't need to look too closely at it. But if you have any feedback that would be good.


clarkbw:
I believe I've noticed this with the todomvc app in the examples directory.  The `app.js` file doesn't always show in the sources list and I found (on the plane) that there wasn't a newSource event happening for it unless I force reloaded the remote browser.

jasonLaster:
Thanks @jlongster. I'll review more closely tomorrow. I'm most interested in not breaking source maps as we don't have tests there yet. 

I'm glad we're considering done of the edge cases aroubd initializing. 

@bgrins I believe this might be what you noticed today with the top level debugger call.

jlongster:
@clarkbw The only time it should ever not show all the sources is if you broke on a `debugger` statement before initializing the debugger. Otherwise the bug you saw was something else. If you can reproduce it we can look into it.

@jasonLaster The only thing we need to make sure is that we call `loadSourceMap

jlongster:
@clarkbw Actually, I take that back. This does fix the bfcache case as well, which is probably what you saw. If you pressed "back" and the page was loaded from memory it won't show all the sources, and this fixes that.

public/js/test/mochitest/browser_dbg-breaking.js
```diff
@@ -0,0 +1,41 @@
...
+  yield waitForPaused(dbg);
+  assertPausedLocation(dbg, "scripts.html", 18);
+  yield resume(dbg);
+
+  const paused = waitForPaused(dbg);
+
+  // Create an eval script that pauses itself.
+  invokeInTab("doEval")
+
+  yield paused;
```
jasonLaster: Oh, interesting. How is this different from `yield waitForPaused(dbg)`?

public/js/test/mochitest/browser_dbg-breaking-from-console.js
```diff
@@ -0,0 +1,15 @@
...
+
+
+add_task(function* () {
+  const url = EXAMPLE_URL + "doc-script-switching.html";
+  const toolbox = yield openNewTabAndToolbox(url, "webconsole");
+
+  let jsterm = toolbox.getPanel("webconsole").hud.jsterm;
+  jsterm.execute("debugger");
+
+  yield waitForTime(100000);
```
jasonLaster: Hmm, out of curiosity, how is this testing the bug that you found? I
* would it be possible to wait for paused?

public/js/test/mochitest/browser_dbg-breaking.js
```diff
@@ -0,0 +1,41 @@
...
+  yield resume(dbg);
+  const source = getSelectedSource(getState()).toJS();
+  // TODO: The url of an eval source should be null.
+  ok(source.url.indexOf("SOURCE") === 0, "It is an eval source");
+
+  yield addBreakpoint(dbg, source, 5);
+  invokeInTab("evaledFunc");
+  yield waitForPaused(dbg);
+  assertPausedLocation(dbg, source, 5);
```
jasonLaster: oh interesting, i'm surprised it's not a new source

public/js/test/mochitest/browser_dbg-breaking.js
```diff
@@ -0,0 +1,41 @@
...
+});
+
+
+// # test anonymous eval scripts, debugger statement and setting bps on them
+// # test breaking from the console before the debugger UI is initialized
```
jasonLaster: I do the same thing :)

public/js/test/mochitest/examples/code-simple1.js
```diff
@@ -5,3 +5,27 @@ function main() {
...
+  eval("(" + function() {
+    debugger;
+
+    window.evaledFunc = function() {
+      var foo = 1;
+      var bar = 2;
+      return foo + bar;
+    };
+  }.toString() + ")();\n //# sourceURL=evaled.js");
+}
```
jasonLaster: nit: it might be cleaner to pull out the namedEval function so that the eval is simpler.

```js
function doNamedEval() {
  const namedEval = function() {
    debugger;

    window.evaledFunc = function() {
      var foo = 1;
      var bar = 2;
      return foo + bar;
    };
  }.toString();

  eval(`(${namedEval})();\n//# sourceURL=evaled.js`);
}
```

public/js/test/mochitest/browser_dbg-breaking.js
```diff
@@ -0,0 +1,41 @@
+/* Any copyright is dedicated to the Public Domain.
+ * http://creativecommons.org/publicdomain/zero/1.0/ */
+
+// Tests the breakpoints are hit in various situations.
```
jasonLaster: Do you feel brave enough to add a back and forward example to the navigation test?

jlongster: I was planning on doing a bfcache test as a separate one, because I'll have to do some special navigation commands to make sure it loads from the bfcache. But I'll think about it, maybe I will go ahead and do that.

public/js/test/mochitest/browser_dbg-breaking.js
jlongster: `waitForPaused` waits for a `paused` thread event. I was thinking that it might be racy to invoke a function that paused the debugger, and then wait for a paused event, because the `paused` event may already have been fired. However, theoretically since JS is single-threaded that event can't come through until we stop executing, so it should be fine to use `waitForPaused` after it. If we see any signs of intermittence with this test though, that's probably the problem.

public/js/test/mochitest/browser_dbg-breaking.js
jlongster: We're just calling a function that `eval` created, so no new sources are created.

public/js/test/mochitest/browser_dbg-breaking-from-console.js
jlongster: I was in the middle of writing this test, so it's not finished yet and there *definitely* won't be any `waitForTime` calls in there :) Same with the comments.

public/js/test/mochitest/examples/code-simple1.js
jlongster: Maybe I'm so used to this because the existing format is what we use all over the place in backend and SpiderMonkey tests.

public/js/test/mochitest/browser_dbg-breaking.js
jasonLaster: yeah - i had the same thought and concluded that JS should be able to handle it. We setup the the store subscription synchronously...

public/js/test/mochitest/browser_dbg-breaking-from-console.js
jasonLaster: haha - `waitForTime(Math.random())` perhaps?

public/js/test/mochitest/examples/code-simple1.js
jasonLaster: I see. I don't feel strongly.

public/js/test/mochitest/browser_dbg-breaking.js
jlongster: It's more a concern of when the `threadClient` gets notifications from the debugger server; `waitForClientEvents("paused")` waits for a future paused event, and if it's already come through it will timeout. We ran into this a lot in the old tests, but I think it was much more problematic because promises were used all throughout the UI so thread events could come through half-way through updating the UI. Our system is much more deterministic.

public/js/clients/firefox.js
```diff
@@ -100,16 +100,22 @@ function initPage(actions) {
...
+  // hit before the debugger is initialized, or a page was loading
+  // from bfcache. In all these cases, we need to make sure to fire
+  // `newSource` notifications even if the server doesn't. Otherwise
+  // they won't appear in the debugger.
+  return threadClient.getSources().then(({ sources }) => {
+    sources.forEach(source => clientEvents.newSource(null, { source }));
+
+    // If the threadClient is already paused, make sure to show a
+    // paused state.
+    clientEvents.paused(null, threadClient.getLastPausePacket());
```
jlongster: You still need to check if the `getLastPausePacket()` is null; the thread may not be paused, so there is none:

```js
    const pausedPacket = threadClient.getLastPausePacket();
    if(pausedPacket) {
      clientEvents.paused(null, pausedPacket);
    }
```

jasonLaster: I checked in the event

jlongster:
I reverted back to the first approach because we would have to make sure to clear out the known sources cache on navigation, and that made it too complex for me. This approach is pretty clean, with the performance hit of normalizing all the sources in `getSources()` even though we probably don't have to. I opened CNN that has tons of sources and didn't notice any serious perf problems, but later on we should take a closer look at this process and see how we can optimize it. Right now, getting the semantics right is the most important part.

fwiw, I've been testing Chrome is various scenarios and it seems *super* slow in many cases. Even just typing `debugger` into the console before clicking on "sources" takes ~5 seconds for the debugger to appear.

jlongster:
This won't pass until https://bugzilla.mozilla.org/show_bug.cgi?id=1303177 lands which I just pushed, so we'll have to wait a day to get it into our testing docker instance.

public/js/main.js
```diff
@@ -28,7 +28,7 @@ const Tabs = require("./components/Tabs");
 const App = require("./components/App");
 
 const createStore = configureStore({
-  log: getValue("logging.actions"),
```
jasonLaster: lets not change this :)

public/js/actions/sources.js
```diff
@@ -87,6 +83,16 @@ function newSource(source) {
...
 }
 
+function newSources(sources) {
+  return ({ dispatch, getState }) => {
+    sources.filter(source => {
+      return !getSource(getState(), source.id);
+    }).forEach(source => {
+      dispatch(newSource(source));
+    });
+  };
```
jasonLaster: the most minor of nits, but maybe inline these so that it's a little bit more compact.

```js
sources.filter(source => !getSource(getState(), source.id))
             .forEach(source => dispatch(newSource(source)));
```

public/js/clients/firefox.js
```diff
@@ -100,16 +101,22 @@ function initPage(actions) {
...
-    "useSourceMaps": false,
-    "autoBlackBox": false
+  // In Firefox, we need to initially request all of the sources. This
+  // usually fires off individual `newSource` notifications as the
+  // debugger finds them, but there may be existing sources already in
+  // the debugger (if it's paused already, or if loading the page from
+  // bfcache) so explicity fire `newSource` events for all returned
+  // sources.
+  return threadClient.getSources().then(({ sources }) => {
+    actions.newSources(sources.map(createSource));
```
jasonLaster: moving this to an action and re-using createSource is really smart.

jlongster:
Haha, thanks. Hopefully the mostly unnecessary map and `createSource` calls don't incur much of a perf hit. It's definitely cleaner this way. I'll merge on Monday once I get the m-c fix in.

public/js/main.js
jasonLaster: Oops. Meant to say, let's change this.

jlongster:
Alright, tests pass. This actually didn't need a new docker image because this removes the `reconfigure` which was the real problem. The m-c patch just makes sure that sourcemaps are off, but because the tests don't test examples with sourcemaps, they work fine if sourcemaps are on (which is what the docker image is testing now).

We'll see if there are any real problems when we go and update the bundle, as it will be testing all the latest code on all platforms. I don't think there will be, as the m-c patch just tweaks the default configs.

jlongster:
Responded to that minor neat, so need to wait for tests to re-run. :)

Issue #726, opened 2016-09-14
Document debugger test commands / assertions


jasonLaster:
We currently have an [api](https://github.com/devtools-html/debugger.html/blob/master/docs/mochitests.md#api) in our mochitests doc that is already out of date. 

It would be nice to use document.js to document the debugger command / assertion functions.

jbhoosreddy:
That doc is a single source of reference for me when it comes to tests. I end up going back to it so many times, and I'm sure many other contributors would find it useful.

jbhoosreddy:
Do you think it will be possible for me to investigate this one, unless someone else is already planning on doing it?

jlongster:
@jbhoosreddy go ahead! Just need to look at all the helper functions in `head.js` and make sure they exist in the API list.

clarkbw:
#750

PR #750, opened 2016-09-16
add mochitest in documentationjs (#726)

jbhoosreddy:
initial attempt at adding documentationjs for mochitests.

### DIFF
--- a/public/js/test/mochitest/head.js
+++ b/public/js/test/mochitest/head.js
@@ -5,6 +5,35 @@
 
 "use strict";
 
+/**
+ * The Mochitest API documentation
+ * @module mochitest
+ */
+
+/**
+ * The mochitest API to wait for certain events.
+ * @module mochitest/waits
+ * @parent mochitest
+ */
+
+/**
+ * The mochitest API predefined asserts.
+ * @module mochitest/asserts
+ * @parent mochitest
+ */
+
+/**
+ * The mochitest API for interacting with the debugger.
+ * @module mochitest/actions
+ * @parent mochitest
+ */
+
+/**
+ * Helper methods for the mochitest API.
+ * @module mochitest/helpers
+ * @parent mochitest
+ */
+
 // shared-head.js handles imports, constants, and utility functions
 Services.scriptloader.loadSubScript("chrome://mochitests/content/browser/devtools/client/framework/test/shared-head.js", this);
 var { Toolbox } = require("devtools/client/framework/toolbox");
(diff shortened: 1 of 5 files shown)


public/js/test/mochitest/head.js
```diff
@@ -58,6 +87,17 @@ function _afterDispatchDone(store, type) {
...
+ * Wait for a specific action type to be dispatch.
+ * If an async action, will wait for it to be done.
+ *
+ * @memberof mochitest/waits
+ * @param {Object} dbg
+ * @param {String} type
+ * @param {Number} eventRepeat
+ * @return {Promise}
+ * @static
+ */
```
jasonLaster: Hmm, I wonder if we can make this file a flow file and then git the params mostly for free. This would be nice because we could type `dbg` and we would then have typed tests. Don't feel like you have to do this for this work to be merged, but I don't think we need the params because we should eventually get them from flow

public/js/test/mochitest/head.js
```diff
@@ -5,6 +5,35 @@
...
+ * @module mochitest/actions
+ * @parent mochitest
+ */
+
+/**
+ * Helper methods for the mochitest API.
+ * @module mochitest/helpers
+ * @parent mochitest
+ */
+
```
jasonLaster: Smart - I like the idea of grouping them

clarkbw:
These docs look really great! 👍 

I pulled it down so here's a couple screenshots 

screen shot 2016-09-16 at 7 42 48 am

mochitests-screenshot

public/js/test/mochitest/head.js
jbhoosreddy: I was thinking of doing it but I had this question.
@jasonLaster If I do convert the `head.js` type annotation to Flow, will I have to make other changes to make sure that the mochitest still run? Or can it handle it?

jlongster: That's a good point, @jbhoosreddy. We don't compile these files so we can't use Flow here, because there's no way to strip the types. We probably have to conform to this syntax. Great job!

jasonLaster: oy... alright

jasonLaster:
Looks good, anything else you'd like to do @jbhoosreddy or is this ready to go?

jbhoosreddy:
@jasonLaster one thing left. removing the API section from mochitest.md and linking it to github.io as you said.

jbhoosreddy:
I don't have access but could someone go to Settings -> GitHub Pages -> "master branch /docs folder" and enable Github pages for this repository? I don't think it is set up right now.

jbhoosreddy:
Also, save for any nits, this is good to go.

clarkbw:
The docs need improvement, but they are here: https://devtools-html.github.io/debugger.html/reference/

jasonLaster:
Ported to https://github.com/devtools-html/debugger.html/pull/788. Should be a simple merge.

Issue #751, opened 2016-09-16
Make the hardcoded development web server port number 8000 configurable 


jkesseler:
The port number in of the server is hardcoded in [`bin/development-server.js`](https://github.com/devtools-html/debugger.html/blob/c8292de68970552049bc9b2cfb4db1ec806e64e3/bin/development-server.js#L86). This will clash if the project you are trying to debug also runs on port 8000.

jasonLaster:
That's a good point @jkesseler It should be a parameter [here](https://github.com/devtools-html/debugger.html/blob/master/config/development.json) and then you could override it in `local.json`.

jbhoosreddy:
fixed in #753

PR #753, opened 2016-09-16
Fix 751 - Make the hardcoded development web server port number 8000 …

adamrhunter:
#751 - fixes issues/751
@jasonLaster - Based on your patch.  I've added additional code to print correct port number to console.

Added development key to config/development.json, added default values for 
serverPort: 8000
examplesPort: 7999

Updated config/README.md to include reference to new config values
Updated bin/development-server.js to read port values, apply them server and print out correct values to console

Tested with the following keys/value in development.json:

  "development": {
    "serverPort": 8000,
    "examplesPort": 7999
  }

Tested with local.json override:

{
  "development": {
    "serverPort": 8001,
    "examplesPort": 7998
  }
}

### DIFF
--- a/bin/development-server.js
+++ b/bin/development-server.js
@@ -82,18 +82,25 @@ app.get("/get", function(req, res) {
   httpReq.on("statusCode", err => res.status(err.message).send(err.message));
 });
 
-// Listen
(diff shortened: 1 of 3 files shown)


jbhoosreddy:
I will go ahead and merge this.

PR #754, opened 2016-09-16
Add search input

jasonLaster:
I'm breaking up the code search issue into a couple of PRs to keep it from getting too big.

This PR shows the search box in the bottom left. It's a pretty simple input box (not too styled).

Interesting things:
* the source footer needs codemirror to do the actual search. I solved this by making the source footer a child of the Editor... It's not great, but makes some sense, the footer really just acts on the selected source, right...
* fix the editor height. the previous PR overlayed the editor bottom, this accounts for the heights correctly.
* search is behind a flag so, should be safe to merge.


public/js/components/Editor.css
```diff
@@ -10,7 +10,7 @@
  */
 .editor-wrapper {
   position: absolute;
-  height: calc(100% - 30px);
+  height: calc(100% - 31px);
```
jlongster: Why 31? Wish we didn't have to hardcode these values... wasn't this just for Chrome? I remember our flexbox being better equipped at this.

jasonLaster: the height plus border

jasonLaster: I don't remember if it's a chrome thing, there were certainly chrome things with codemirror expanding.

public/js/components/Editor.js
```diff
@@ -221,22 +223,30 @@ const Editor = React.createClass({
...
         dom.div({ className: "editor-mount" }),
-        !isLoading &&
-          breakpoints.valueSeq().map(bp => {
-            return Breakpoint({
-              key: makeLocationId(bp.location),
-              breakpoint: bp,
-              editor: this.editor && this.editor.codeMirror
-            });
-          })
+        this.renderBreakpoints(),
```
jlongster: I'm not sure this render function is big enough to be split up. It's nice to have things in one place when possible, and only split it up when it gets too large. The `renderBreakpoints` function is several lines longer than this expression :)

jlongster:
Looks good, just a few nits but you can respond to them how you like.

public/js/components/Editor.js
jlongster: It's just different styles. I generally like to keep as much markup in one place as possible because I find it less syntax visually and easier to read through. There's different styles in the React community, many are like that but other like to break things up too. It's not a huge deal.

jasonLaster: Yeah, I can see where you're coming from. I've seen both styles, personally I'm not a fan of a single large expression and the ternaries, booleans, etc that go into it working. 

Looking at `renderBreakpoints` specifically. The function is a good example of a 15 line function with no dependencies, which simplifies the `render` function. 

I'd be okay keeping more of the logic in `render` if you feel like that would be important for a consistent style.

jlongster: I'm ok keeping this as is, but generally we should talk about style so it doesn't keep coming up. I think it's more flexible to accept that things like ternaries are how you do conditional templating w/React, and trying to avoid it is going to break things up too much in my opinion. Once you're used to it, it's really nice to look at one place and read what it renders, instead of always splitting off into `render*` functions, and it's much easier to refactor. Of course, if it's any longer than ~10 lines or requires complicated logic, that's perfectly fine to do. It's subjective, but I wouldn't say we should enforce things like avoiding ternaries.

I personally don't mind this code, but one thing that would help clean it up for some people is to use JSX. I'm fine with that too, although I don't know how much work it would be to convert to it.

PR #755, opened 2016-09-16
Tweak the sidebar

jasonLaster:
I didn't touch this, but i think it might be nice to sequence the animation, basically rotate and then collapse the sidebar. What do you guys think?


public/js/components/SourceTabs.css
```diff
@@ -74,6 +74,11 @@
   transition: transform 0.5s ease;
 }
 
+.collapse-button-left:hover,
+.collapse-button-right:hover {
+  cursor: pointer;
+}
+
```
jlongster: None of our current expand/collapse buttons have a pointer, nor does the Chrome devtools do this. I know we've added the pointer icon to a few other places, but we should decide what we want to be consistent. We traditionally didn't use this before (and I kind of like how Chrome feels without it).

jasonLaster: oh interesting, just assumed buttons would have it

jlongster:
This PR looks good. About sequencing it, I think it would make the animation feel too long. Hard to say without seeing it, but it's probably best to do it simultaneously.

public/js/components/SplitBox.css
```diff
@@ -14,7 +14,7 @@
 .split-box .controlled {
   display: flex;
   overflow: auto;
-  transition: width 0.5s ease;
+  /*transition: width 0.5s ease;*/
```
jasonLaster: This fixed the performance problem

public/js/components/SplitBox.js
```diff
@@ -24,11 +24,26 @@ const SplitBox = React.createClass({
...
+  },
+
+  onMove(x) {
+    const width = this.calculateWidth(x);
+    this.getflexEl().style.width = `${width}px`;
+  },
+
+  onStop(x) {
+    const el = this.getflexEl();
+    const width = parseInt(el.style.width, 10);
```
jasonLaster: This reduces a lot of CPU issues, but on my computer we were not CPU bound, so not sure if i want to fix it. Also, touching the style property might be taboo

public/js/components/SplitBox.css
jlongster: Does this fix resizing with the mouse? Because right now that doesn't work: the transition makes it not follow the mouse. I'm assuming this transition is the reason?

public/js/components/SplitBox.js
```diff
@@ -41,22 +56,34 @@ const SplitBox = React.createClass({
...
-          { uncontrolled: !rightFlex },
-          { controlled: rightFlex }, { collapsed }),
-          style: { width: rightFlex ? width : null }},
+        {
+          className: classnames(
+            { uncontrolled: !rightFlex },
+            { controlled: rightFlex }, { collapsed }
+          ),
+          style: { width: rightFlex ? width : null },
+          ref: "right"
```
jlongster: It's a lot better to avoid refs if possible. The previous implementation didn't use them; what does this by us? It's a little hard to see from this diff.

public/js/reducers/ui.js
```diff
@@ -29,7 +29,7 @@ const State = makeRecord(({
 
 function update(state = State(), action: Action) : Record<SidebarsState> {
   switch (action.type) {
-    case C.COLLAPSE_SIDEBAR: {
+    case C.TOGGLE_SIDEBAR: {
```
jlongster: We need to change all the `C.*` usages to `constants.*`. It's really important to be consistent and that's not only what we use elsewhere, but generally what many redux projects use.

I was going to look at that other PR but it got merged, otherwise I would have mentioned it there, sorry.

jasonLaster: that's fine, we can tack this on as well

public/js/components/SplitBox.js
jlongster: Oh, this is avoiding going through redux basically? If we're doing this, what's the point of keeping the state in redux?

BTW, we need to integrate Honza's new SplitBox component, which will mostly void all this work...

public/js/components/SplitBox.css
jasonLaster: This fixes resizing with the mouse for chrome. Firefox is still pretty fussy, but maybe reducing the easing to a smaller number might help

public/js/components/SplitBox.js
jasonLaster: This gives us a reference to the left and right elements. Previously, we didn't need them - just found the top element with findNode.

public/js/components/SplitBox.css
jlongster: It was very smooth for me before, so I wonder what's different. There shouldn't be any easing here though, right?

public/js/components/SplitBox.js
jasonLaster: The benefit for redux is that by keeping the width in the store we can expand to the last known width.

jasonLaster: Agreed that we should update to Honza's split box soon

public/js/components/SplitBox.js
jlongster: It's better to use `querySelector` for them. string refs may be deprecated in a future version of React for various reasons, one of them being that it makes code difficult to follow (where did this ref come from? etc)

public/js/components/SplitBox.js
jlongster: Not sure how I see that couldn't be component local state, but it's not a big deal.

jlongster: Thanks a lot for optimizing this though - one good though about the redux state is it should be easy to init the app state based on prefs. It's important that they resize smoothly though.

public/js/components/SplitBox.css
jasonLaster: oh, this is probably for the collapse button. But it is firing for all of the drag handler

public/js/components/SplitBox.js
jasonLaster: do you need a reference to the top element?

public/js/reducers/ui.js
wldcordeiro: @jlongster heh that's a habit from my work project we always refer to our constants as `C` and I noticed we called immutable `I` so it seemed convenient. 😆

public/js/constants.js
```diff
@@ -30,7 +30,7 @@ exports.RELOAD = "RELOAD";
 exports.ADD_TABS = "ADD_TABS";
 exports.SELECT_TAB = "SELECT_TAB";
 
-exports.COLLAPSE_SIDEBAR = "COLLAPSE_SIDEBAR";
+exports.TOGGLE_SIDEBAR = "TOGGLE_SIDEBAR";
```
wldcordeiro: Heh missed this.

jasonLaster:
@wldcordeiro I'm going to revert the collapse buttons on master so that we can figure out some of the performance implications without rushing.

wldcordeiro:
I think sequencing could work but it'd have to be very slight to not feel weird.

jasonLaster:
Hey @wldcordeiro, I think I had an interesting a little while ago. One of the big problems we ran into was that it was difficult to add the toggle buttons in the center pane. Here's an admittedly *hacky* way to do it, that adds the buttons inside of split box and lets them be toggled. You might recognize your *prevWidth* variable :), but perhaps we can replace that with *collapse* true as well. 

Here's a [branch](https://github.com/devtools-html/debugger.html/compare/master...jasonLaster:toggle-btns?expand=1) where the work lives.

And here's a photo

screen shot 2016-09-16 at 9 09 54 pm

wldcordeiro:
@jasonLaster sorry, I'm not sure I follow, is the idea to move the behavior back into React state? Been moving this weekend so I hadn't had a chance to catch up.

jasonLaster:
No worries. Three thoughts
1. Redux should only set state on stop to avoid a lot of CPU
2. The easing transition should only be on when toggling
3. If we can do everything in split box state, why not.

jlongster:
I haven't seen a good reason why this state should be in redux at all. If we're going to be storing it on the component, we should just do that. The state will exist for the component's lifetime so it will expand back to the width of the previous state. It just need to take initial widths so we can load it from prefs.

It might be a good time to just upgrade to Honza's component, as much of this discussion will be voided by the design of that one anyway. I'm not sure if it has the ability to collapse/expand already, but it might. If not, we can ask him to add it.

jasonLaster:
@jlongster I agree that we should upgrade first. Also, collapsing does seem like something that we should talk to honza.

jasonLaster:
closing this PR as i just got a lot of information about the [m-c] splitter.

public/js/components/SourceTabs.css
helenvholmes: Hm, this had never occurred to me before. I think the better UX decision would be for these to have cursor icons, but I understand James' hesitation to be consistent. 

If we add this (which I think would be good), can we open a corresponding MC bug for someone to add the cursor pointers elsewhere in the tools?

public/js/reducers/ui.js
jlongster: @wldcordeiro no worries! We used to import functions directly like `let { Map } = require("immutable")` but that overrides the native map, so we made it a shorthand, but most things don't use that format.

public/js/components/SplitBox.js
jlongster: Sorry, totally missed this. You would do `ReactDOM.findDOMNode(this).querySelector(...)`

PR #757, opened 2016-09-16
fix download firefox artifact script

clarkbw:
When the firefox directory didn’t exist the `cd` command in FIREFOX_PATH would fail.  However we do want the absolute path so we reset the variable after we check the directory does exist.


jlongster:
Thanks, that's a good idea.

PR #759, opened 2016-09-16
Fixes #669 by adding a wider splitter handle

### LINKED ISSUE #669: The splitter's drag area is too narrow
By drag area, I mean the area where the cursor turns to ew-resize and you can drag the splitter to resize the sidebar.
From my testing, it looks like it's only 1px wide (maybe 2?), which is way too narrow to reliably drag the splitter.
Especially on my desktop computer with a mouse, it gets hard to precisely get the splitter, maybe with a touchpad it's easier, not sure.
I think the line between the 2 panels should remain a 1px line, but there should be a hidden element on top that's at least 5px (not sure how wide we have other splitters in devtools) that serves as a proxy for dragging.
capture

captainbrosset:
This should fix #669. My first PR here, mostly to try things out really. Let me know what you think. There was comment in the issue that seemed to indicate that this issue was fixed by some other work, but I wasn't sure. So I took a stab at it anyway.

### DIFF
--- a/public/js/components/SplitBox.css
+++ b/public/js/components/SplitBox.css
@@ -23,13 +23,20 @@
 
 .split-box .splitter {
   background-color: var(--theme-splitter-color);
-  border-bottom-width: 0;
-  border-color: white;
-  border-left-width: 0;
-  border-right-width: 0;
-  border-style: solid;
-  border-top-width: 0;
-  box-sizing: content-box;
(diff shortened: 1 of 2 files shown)


jasonLaster:
Looks great. Thanks for adding.

@wldcordeiro was working on adding the sidebar toggle buttons #668, which is why i believe this issue was mentioned.

Issue #569, opened 2016-08-22
Add husky to the project


jbhoosreddy:
Husky is useful for improving the quality of code pushed by contributors. It reduces a lot of noise. And prevent a lot of silly mistakes (which I also tend to make...). Seeing the recent work done here, with the perspective of contributors, I would recommend adding husky to the project also.

https://github.com/typicode/husky

jlongster:
My only concern is that it forces you to go through the whole test suite even when pushing out documentation updates, etc. Given that we really should go through the CI 100% of the time anyway (because we have to be sure that it works in the CI environment, even if it works locally), I'd like to develop a culture where you always open a PR and wait for the CI, and run it locally if you just want to be faster. But I'm not against it if others want this.

jbhoosreddy:
In that case, husky might not be the best solution for this use case.

jasonLaster:
I'm also inclined to rely on CI. My issue here is that CI is an async process and I would not want to slow down the contributor workflow to catch a case that could be handled by CI

jasonLaster:
Thanks for the suggestion. I'm interested in the project because git hooks are useful, but often poorly understood

jlongster:
It is a neat idea, maybe could at least run the linter on push? That would catch stupid stuff but should be very quick.

jbhoosreddy:
A hook for linting was what I had in mind. Quick to resolve and easy to fix. We get quicker feedback also instead of waiting for CI results. I'll try it out and see how it could fit in this context.

jlongster:
You can run `npm run lint` if you don't want to wait on CI :) But I get your point.

jbhoosreddy:
@jlongster: \me 😁

jasonLaster:
closing for now, we might want to come back to this later.

clarkbw:
Jumping into this late, but I can see the appeal of running `npm run lint` before a commit.  It is pretty quick and could save some time.  Lets come back to this later.

eoger:
We did this on fxa-auth-server and fxa-content-server and we couldn't be happier.
It's also pretty simple to [set up](https://github.com/mozilla/fxa-auth-server/pull/1270/files).
Personally it saved me from a lot of shameful force-pushs and it could help newer contributors to make better PRs.

clarkbw:
Ok, I want to come back to this.  @jbhoosreddy was right and each time I see a lint failure in the CI I'm filled with sadness. 😭 😢 

[...]

PR #760, opened 2016-09-17
add husky (#569)

jbhoosreddy:
Feel free to close this PR in case we don't want to use it. I tried it out and it is not restrictive to developers from what I see.
It is funny because for whatever reason my master branch already has a lot of linting errors.
So I had to use
```
git push origin --no-verify
```
to bypass the hooks.
It allows you to bypass the hooks if you want to.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,8 @@
     "firefox": "node bin/firefox-driver --start",
     "mochitests-watch": "MOCHITESTS=true TARGET=firefox-panel webpack --watch",
     "prepublish": "webpack",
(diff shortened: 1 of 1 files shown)


jasonLaster:
Thanks @jbhoosreddy. I agree with the conversation that this could be helpful for linting. We've missed several of the lints until CI and this could be a nice win.

jlongster:
Sounds good to me too. I haven't tried it locally yet, and I'm curious how this hooks into git. I'll merge it and we'll try it out.

PR #765, opened 2016-09-19
include the review system in the the PR section of the docs

clarkbw:
Updates the need for running unit tests, not necessarily all tests
Adds a line item to the PR section for the required review
Includes links to the GitHub Review system.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -215,10 +215,11 @@ When a person is assigned to an issue this indicates an _intent to implement_.  
 * Include screenshots and animated GIFs in your pull request whenever possible.
 * List any steps necessary to trigger the feature you've created or bug you are fixing
 * Always run the [unit tests](#unit-tests) locally before creating your PR
-* Once the tests have passed in the PR you must request review using the GitHub review system
+ * You are not required to run the [integration tests](#integration-tests) locally
+* Once the tests have passed in the PR you must receive a review using the GitHub review system
  * To learn more about GitHub reviews take a look at their [documentation](https://help.github.com/articles/reviewing-changes-in-pull-requests/) and [video tutorial](https://youtu.be/HW0RPaJqm4g)
-* Request review from @jasonLaster or @jlongster
- * Mention either of their names in the PR if you're having issues with tests passing
+* Request review from @jasonLaster or @jlongster by mentioning their names in the PR
+ * Also mention either of their names in the PR if you're having issues with [integration tests](#integration-tests) passing
 
 > **Working on your first Pull Request?** You can learn how from this *free* series [How to Contribute to an Open Source Project on GitHub](https://egghead.io/series/how-to-contribute-to-an-open-source-project-on-github)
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -214,8 +214,12 @@ When a person is assigned to an issue this indicates an _intent to implement_.
 
 * Include screenshots and animated GIFs in your pull request whenever possible.
 * List any steps necessary to trigger the feature you've created or bug you are fixing
-* Always run the [tests](#tests) locally before creating your PR
+* Always run the [unit tests](#unit-tests) locally before creating your PR
+ * You are not required to run the [integration tests](#integration-tests) locally
+* Once the tests have passed in the PR you must receive a review using the GitHub review system
+ * To learn more about GitHub reviews take a look at their [documentation](https://help.github.com/articles/reviewing-changes-in-pull-requests/) and [video tutorial](https://youtu.be/HW0RPaJqm4g)
 * Request review from @jasonLaster or @jlongster by mentioning their names in the PR
+ * Also mention either of their names in the PR if you're having issues with [integration tests](#integration-tests) passing
```
wldcordeiro: Same?

clarkbw:
Oh, those should be sub-bullet points.  Are they looking strange in the rich diff?

wldcordeiro:
@clarkbw oh no, just the single space threw me off and made me think it was a mistake since in Markdown it's typical to use 2 or 4 for sub bullets. If it renders right I don't care heh.

clarkbw:
> it's typical to use 2 or 4 for sub bullets

oh wow, I think I've been doing it wrong this whole time

clarkbw:
@jasonLaster review?

CONTRIBUTING.md
```diff
@@ -214,8 +214,12 @@ When a person is assigned to an issue this indicates an _intent to implement_.
 
 * Include screenshots and animated GIFs in your pull request whenever possible.
 * List any steps necessary to trigger the feature you've created or bug you are fixing
-* Always run the [tests](#tests) locally before creating your PR
+* Always run the [unit tests](#unit-tests) locally before creating your PR
+ * You are not required to run the [integration tests](#integration-tests) locally
```
jlongster: Something like this might be better: "Run the unit tests locally before creating your PR to catch an issues early. You can run the integration tests too, but if you want to avoid setting up that environment, you can open a PR to run them automatically on CI."

You don't have to use my wording at all, but just making it clear that you can us the CI instead of a local setup to run the integration tests.

CONTRIBUTING.md
jlongster: Yeah, this seems a little redundant. Another bullet point probably isn't required, but you can add to the the other one.

CONTRIBUTING.md
jasonLaster: Hmm, I like where you're going with that @jlongster. Might be nice to say something like

*we recommend running the unit tests locally.* and then say the integration tests can be run on CI automatically...

CONTRIBUTING.md
jasonLaster: yeah - i don't think this is needed. We can jump in at that point

PR #766, opened 2016-09-19
Upgrade stylelint

jasonLaster:
shouldn't change anything


jlongster:
Try kicking CI again. I've had no problems in the last hour.

PR #767, opened 2016-09-19
Wire up the source search input box

jasonLaster:
This borrows the [logic](https://dxr.mozilla.org/mozilla-central/source/devtools/client/sourceeditor/debugger.js#16-91) found in `sourceditor/debugger` to implement search for the new debugger.

Also adds Cmd+f for focusing on the input box. I chose to move the shortcuts reference to a module so that it would be easier to access from other components across the tree. Shortcuts is an example of a stateful module that is global to the debugger.

Fortunately, the shortcuts persist across tab navigations and are cleared when the debugger is refreshed.


public/js/components/SourceFooter.js
```diff
@@ -80,15 +84,44 @@ const SourceFooter = React.createClass({
...
+  },
+
+  focusSearch(shortcut, e) {
+    e.stopPropagation();
+    e.preventDefault();
+    const node = findDOMNode(this).querySelector(".source-search");
+    node.focus();
+  },
+
+  setupKeyboardShortcuts() {
```
jasonLaster: this is a pattern borrowed from the right sidebar. It's designed to be called once post mount

public/js/components/App.js
```diff
@@ -60,8 +60,8 @@ const App = React.createClass({
   },
 
   componentDidMount() {
-    this.shortcuts = new KeyShortcuts({ window });
-    this.shortcuts.on("CmdOrCtrl+P", this.toggleSourcesSearch);
+    const shortcuts = setupShortcuts(window);
```
jlongster: I don't know if you have time to try it right now, but another option is to put it in [react context](https://facebook.github.io/react/docs/context.html). This would make it available to all components and I think it an appropriate usage of it. The benefit is that it would be available in the first render, because now it's `null` and I've seen a few issues with that. However I think you also solved that by making it a function `getShortcuts` instead of passing the instance down directly.

In fact, you could just pass down a function `{ getShortcuts: () => this.shortcuts }` too which would solve the `null` problem. Is that mainly what you're trying to solve here?

jlongster: It's not a big deal and if you want to just move forward that's fine too.

jasonLaster: oh interesting. I'd like to give react context a try. I'd propose doing it in a separate PR as this should not become a shortcuts PR though.

> In fact, you could just pass down a function 

I'm not sure if that would help because you'd still need to check if shortcuts is available. Also, we would still be passing a prop down the tree.

I'd like to investigate refactoring shortcuts in a later pass

jasonLaster: Thought a bit more about it and, it shouldn't be a big diff, so i'll take a look at contexts now.

public/js/utils/source-editor.js
```diff
@@ -4,10 +4,7 @@ require("codemirror/lib/codemirror.css");
 require("codemirror/mode/javascript/javascript");
 require("codemirror/mode/htmlmixed/htmlmixed");
 require("../lib/codemirror-mozilla.css");
-// enable to get local search to work
-// require("codemirror/addon/search/search");
-// require("codemirror/addon/search/searchcursor");
-// require("codemirror/addon/search/match-highliter");
```
jlongster: You could guard these with this `if` statement:

```js
if (process.env.TARGET !== "firefox-panel") {
  ...
}
```

As far as I know, webpack can see at build-time that it's false in the panel so it won't include these, but locally it will.

public/js/utils/source-search.js
```diff
@@ -0,0 +1,104 @@
+// These functions implement search within the debugger. Since
```
jlongster: Is this module from m-c?

public/js/utils/source-editor.js
jasonLaster: true - we actually only need the searchcursor, which should be excluded from the bundle because the source-editor is excluded

public/js/utils/source-search.js
jasonLaster: It's ripped out from debugger.js

public/js/components/App.js
jlongster: OK, I'll keep looking at this PR a bit too. I bet we could get it merged tomorrow.

Passing a function helps because you're not passing the prop down. The problem before is you're passing `this.shortcuts` down as a value, so if it's `null`, you just pass `null` and child components will forever get `null`. If you pass a function, child components will always get the latest `this.shortcuts` which will not be null after mounting.

But it sound like context might be a bit more promising anyway.

jasonLaster: Ahh, i follow. that makes sense.

PR #768, opened 2016-09-19
Check Node version

### LINKED ISSUE #763: Error while entered 'npm start'
Here is screenshot - https://gyazo.com/d2e0db8926cbee6b5e05f95ab3888a7b

jasonLaster:
Fixes #763 

This sets a minimum node version. I'm personally not sure if 5 is the minimum, we might be at 4.x http://node.green/#let.

* adds engine to check for outer installs
* add check to start because this is where most people will hit it
* removes prepublish as i don't think we use that.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,8 @@
   "homepage": "https://github.com/devtools-html/debugger.html#readme",
   "scripts": {
     "flow": "flow",
-    "start": "node bin/development-server",
+    "start": "npm run check-node && node bin/development-server",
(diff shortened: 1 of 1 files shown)


jlongster:
Looks good!

briantully:
Has this been merged into master or did it get overwritten? I am experiencing the same issue as #763 with a fresh install of master and I don't see any check for node version in package.json version 0.0.6

jasonLaster:
Hi, we check the node version in the dev server. Look for checkNode

PR #769, opened 2016-09-19
Refactor editor breakpoint

jasonLaster:
Refactors the editor breakpoint to re-use the svg

### DIFF
--- a/public/js/components/EditorBreakpoint.js
+++ b/public/js/components/EditorBreakpoint.js
@@ -1,25 +1,21 @@
 const React = require("react");
+const ReactDOM = require("react-dom");
+
 const { PropTypes } = React;
 const classnames = require("classnames");
+const Svg = require("./utils/Svg");
+
+let breakpointSvg = document.createElement("div");
+ReactDOM.render(Svg("breakpoint"), breakpointSvg);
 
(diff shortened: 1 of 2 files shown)


public/js/components/EditorBreakpoint.js
```diff
@@ -1,25 +1,21 @@
 const React = require("react");
+const ReactDOM = require("react-dom");
+
 const { PropTypes } = React;
 const classnames = require("classnames");
+const Svg = require("./utils/Svg");
+
+let breakpointSvg = document.createElement("div");
+ReactDOM.render(Svg("breakpoint"), breakpointSvg);
```
jlongster: I'm OK with this for now, but seems like a roundabout way to load SVGs inline. Ideally we could do `require("breakpoint.svg")`, which we can do right now, but it just returns a string full of SVG. We'd have to parse that and get the path data out of it, so at that point this is probably easier.

Eventually we might want to fork the svg-inline-* projects that we use because they are deprecated and we could refactor them a bit to be easier to use (they are very small libs)

jasonLaster: yeah - that's a good point, maybe it makes sense to add an exported function to Svg.js for a raw node as well

PR #770, opened 2016-09-19
Revert "Add search input"

jlongster:
This reverts commit 10902f9c41df0451a2d7a4bca81e79038548ee2c.

This is needed because CSS changes are still being applied even when the feature is off. This needs to bake a little more before we should merge it in. Right now I cut releases off master so it needs to be in a production-ready state.

### DIFF
--- a/config/development.json
+++ b/config/development.json
@@ -10,8 +10,7 @@
     "tabs": true,
     "sourceMaps": true,
     "prettyPrint": false,
-    "watchExpressions": false,
-    "search": true
+    "watchExpressions": false
   },
   "chrome": {
     "debug": false,
--- a/public/js/components/App.css
+++ b/public/js/components/App.css
@@ -43,6 +43,30 @@ body {
   cursor: pointer;
 }
 
+.source-footer {
+  background: var(--theme-body-background);
+  position: absolute;
+  bottom: 0;
+  right: 0;
+  z-index: 100;
+  width: 100px;
+  opacity: 0.9;
+}
+
+.source-footer .command-bar {
+  float: right;
+}
+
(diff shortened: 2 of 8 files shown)


wldcordeiro:
@jlongster @jasonLaster might be time to have a `develop` -> `master` branch setup so features can be iterated on? Just a thought

jlongster:
Yes, we've talked about that before. Not entirely sure because the whole point of nightly is to land features that might not quite be done yet, so it's a whole new layer on top of the existing Firefox trains. Personally I think we just wait to land things until they are a little more fleshed out, use feature flags when needed, and make sure the flags are actually working (here, CSS was still applied).

Not against it, we'll probably end up doing it, but there's already a good bit of process so need to figure out how to do it easily.

jlongster:
(fwiw, devtools team is looking into different way to deploy so we don't have to wait for nightly -> aurora -> beta -> release, and if we use a different system something like that totally makes sense)

jasonLaster:
@jlongster - mind holding off for a minute. It looks like the bug is unrelated.

jlongster:
To expand a little more, I foresee a problem with a `next` or `develop` branch where a certain feature on that branch is finished and needs to be released, but it lives on that branch with many other features which are not ready. So it doesn't necessarily solve the problem of cutting releases, as things will still be blocked.

I think we should just keep in-development things on branches and merge them in when they are "good enough".

jlongster:
@jasonLaster sure, although I'm pretty certain this is the cause...

jlongster:
Confirmed that it fixes it and I already have a try run running here. I'll double-check before it actually lands.

PR #773, opened 2016-09-20
Always render editor so only 1 CodeMirror instance is created

kelvin211111:


jlongster:
This looks like an accident. It only has 1 old commit of mine? You may be trying to merge in the wrong local branch.

Issue #661, opened 2016-09-02
Idea: connect selected tab to editor by removing border


jlongster:
I think this makes it very clear what the current tab is:

<img width="571" alt="screen shot 2016-09-02 at 5 01 08 pm" src="https://cloud.githubusercontent.com/assets/17031/18218242/56f1fe6c-712f-11e6-9413-9b4cd961c4e9.png">

What do you think, @helenvholmes ? I love the recent updates to it, just thought this might be another small tweak.

helenvholmes:
Hahaha, three thumbs up, I think that means we should do it 👍🏽

(James, did you want to submit the PR for it, since you've already done the work?)

jasonLaster:
oh nice

jlongster:
Yeah, I can try to do it this week. I didn't actually do implement it, I just fiddled with the HTML/CSS in devtools to hack it up... But it shouldn't be too hard :)

jlongster:
It's a big deal if we don't get to it this week. It's more important to get testing and bugfixes in before the merge.

helenvholmes:
https://cl.ly/0R0b102a3X2X << one thing to check is what happens in this situation. (Might look odd, I'm not certain.)

wldcordeiro:
I got this one. ✋

PR #774, opened 2016-09-20
Remove active tab border

wldcordeiro:
Closes issue #661 

## Screenshot

screenshot_20160919_212930

### DIFF
--- a/public/js/components/SourceTabs.css
+++ b/public/js/components/SourceTabs.css
@@ -33,6 +33,7 @@
 .source-tab.active {
   color: var(--theme-body-color);
   background-color: var(--theme-body-background);
+  height: 24px;
 }
 
 .source-tab path {


jlongster:
Hardcoding the height seems like it might cause issues on various platforms; are we sure it's `24px` on all platforms, with subtle font rendering differences, etc? How did you get 24? I don't see 23 hardcoded anywhere and I assume this just needs to make it 1px taller to overlap the border.

wldcordeiro:
@jlongster I'm not a big fan of hardcoding it either, `23` was hardcoded here though, https://github.com/wldcordeiro/debugger.html/blob/92b582e80174f0b821f823652fdd101765927334/public/js/components/SourceTabs.css#L19

jlongster:
I see, ok, that makes sense.

jasonLaster:
@wldcordeiro mind updating the branch, or giving us write permissions

wldcordeiro:
@jasonLaster rebased onto master.

PR #778, opened 2016-09-20
Fixes the code search ui (part 1)

jasonLaster:
The easiest thing to do is to just look at the second commit.

### DIFF
--- a/public/js/components/SourceFooter.js
+++ b/public/js/components/SourceFooter.js
@@ -81,13 +81,26 @@ const SourceFooter = React.createClass({
     );
   },
 
+  searchBox() {
+    return dom.input({
+      onKeyPress: (e) => this.onKeyPress(e)
+    });
+  },
+
+  onKeyPress(e) {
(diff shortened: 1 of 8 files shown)


public/js/components/SourceFooter.js
```diff
@@ -96,7 +96,7 @@ const SourceFooter = React.createClass({
         (!isEnabled("prettyPrint")
           && !isEnabled("blackBox")
           && !isEnabled("search"))) {
-      return null;
+      return dom.div({ className: "source-footer" });
```
jlongster: This is going to show a blank footer for now while none of the features are enabled, which might look sort of weird. Is there any way to avoid showing the footer until it's actually used for something?

jasonLaster: This is probably fine, given the next release is likely to include source search. 

I could add a JS condition to hide the footer and increase the editor height, but that doesn't seem worth it.

PR #779, opened 2016-09-20
Actually update local files when updating local firefox repo (fixes #772)

### LINKED ISSUE #772: Prepare Mochitest dev does not apply updates
when mercurial is out of date it fetches the latest commits, but does not apply them.

My guess is that we need to do

```
rm -r devtools/client/new/debugger/test/mochitest
hg update -C
```

jlongster:
Didn't realize that `hg pull` doesn't actually change the local files, which is really confusing.

### DIFF
--- a/bin/download-firefox-artifact
+++ b/bin/download-firefox-artifact
@@ -29,10 +29,10 @@ if [ -d "$FIREFOX_PATH" ]; then
         fi
 
         hg revert -a
-        hg purge
     fi
 
     hg pull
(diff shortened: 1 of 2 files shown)


jasonLaster:
it's strange that it didn't warn me that i don't have it.

jlongster:
You may have it, depending on how to setup hg. One of Mozilla's scripts may have auto-enabled it for you. But it's good to just avoid it so others don't run into that issue.

PR #782, opened 2016-09-20
Use just an incrementing number rather than fake uuid

wldcordeiro:
@jlongster since we talked on irc.

### DIFF
--- a/public/js/lib/devtools-sham/sdk/util/uuid.js
+++ b/public/js/lib/devtools-sham/sdk/util/uuid.js
@@ -1,7 +1,0 @@
-
-let i = 1;
-function uuid() {
-  return 'not-really-uuid' + (i++);
-}
-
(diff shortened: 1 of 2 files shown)


jlongster:
Not sure we should change the actual `uuid` module; the point of the name was to make it clear to the user that we didn't have a proper uuid implementation. The only place this is used is `public/js/utils/redux/middleware/promise.js`, so it's probably better to go in there and avoid the uuid module entirely and use an incrementing id. Do you mind doing that instead?

wldcordeiro:
@jlongster if I do that should I just remove the `uuid` module? It'll be unused.

jlongster:
@wldcordeiro yeah, you can try that and see if anything errors. I don't think anything else should be including it.

jlongster:
Looks fine, thanks.

PR #783, opened 2016-09-20
Add codemirror documents

jasonLaster:
I'm sharing this version early because this is the core logic as I see it now. It does some things well, it doesn't try to do others :)

Things it does:
* keep a list of docs per source
* swaps between them so big sources are only loaded once
* keeps scroll position
* handles breakpoints (probably re-renders every time)

Things it doesn't do:
* doesn't delete docs when a tab is closed or navigation - handling closed tabs is easier than navigation as the state is in the component. I might move the state somewhere else to handle navigation, we'll see. 
* doesn't handle loading - i'm planning on adding a spinner
* doesn't handle errors - i'm working on the assumption that errors are permanent, a.k.a the source text is not re-fetched and can become valid. 

Things it might do:
* it probably works for pretty printing and source maps out of the box

### DIFF
--- a/public/js/components/Editor.js
+++ b/public/js/components/Editor.js
@@ -179,31 +179,19 @@ const Editor = React.createClass({
     const { sourceText, selectedLocation } = nextProps;
 
     if (!sourceText) {
-      return;
+      this.showMessage("");
+    } else if (!isTextForSource(sourceText)) {
+      this.showMessage(sourceText.get("error") || "Loading...");
+    } else if (this.props.sourceText !== sourceText) {
+      this.showSourceText(sourceText, selectedLocation);
     }
 
-    if (sourceText.get("loading")) {
-      return this.setLoading();
-    }
-
-    if (this.props.sourceText !== sourceText) {
-      this.updateEditor(sourceText, selectedLocation);
-      resizeBreakpointGutter(this.editor.codeMirror);
-    }
+    resizeBreakpointGutter(this.editor.codeMirror);
   },
 
-  setLoading() {
-    let doc = getDocument("loading");
-    if (doc) {
-      this.editor.replaceDocument(doc);
-      return doc;
-    }
-
-    doc = this.editor.createDocument();
-    setDocument("loading", doc);
-    this.editor.replaceDocument(doc);
-
-    doc.setValue("Loading...");
+  showMessage(msg) {
+    this.editor.replaceDocument(this.editor.createDocument());
+    this.setText(msg);
     this.editor.setMode({ name: "text" });
   },
 
@@ -212,7 +200,7 @@ const Editor = React.createClass({
    * document with the correct mode and text.
    *
    */
-  updateEditor(sourceText, selectedLocation) {
+  showSourceText(sourceText, selectedLocation) {
     let doc = getDocument(selectedLocation.sourceId);
     if (doc) {
       this.editor.replaceDocument(doc);
@@ -222,11 +210,6 @@ const Editor = React.createClass({
     doc = this.editor.createDocument();
     setDocument(selectedLocation.sourceId, doc);
     this.editor.replaceDocument(doc);
-
-    if (sourceText.get("error")) {
-      this.setText(sourceText.get("error"));
-      this.editor.setMode({ name: "text" });
-    }
 
     this.setText(sourceText.get("text"));
     this.setMode(sourceText);
--- a/public/js/components/Editor.js
+++ b/public/js/components/Editor.js
@@ -14,6 +14,8 @@ const actions = require("../actions");
 const Breakpoint = React.createFactory(require("./EditorBreakpoint"));
 const { DOM: dom, PropTypes } = React;
 
+const { getDocument, setDocument } = require("../utils/source-documents");
+
 require("./Editor.css");
 
 function isTextForSource(sourceText) {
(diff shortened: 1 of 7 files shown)


jasonLaster:
@jlongster this is ready for review. It handles navigation, closing tabs, loading, errors (need to test), pretty printing, source maps.

public/js/components/Editor.js
```diff
@@ -174,25 +176,62 @@ const Editor = React.createClass({
...
+    }
+
+    doc = this.editor.createDocument();
+    setDocument(selectedLocation.sourceId, doc);
+    this.editor.replaceDocument(doc);
+
+    if (sourceText.get("error")) {
+      this.setText(sourceText.get("error"));
+      this.editor.setMode({ name: "text" });
+    }
```
jasonLaster: seemed easier to handle errors here, i could move it up if you think it would help

public/js/utils/source-documents.js
```diff
@@ -0,0 +1,24 @@
+let sourceDocs = {};
```
jasonLaster: keeping state in a util is a pattern consistent with source maps, but in that case there is state in a worker. I was reluctant to subscribe to the store or move the state into the reducer as it is not serializable. But thoughts are welcome

public/js/utils/source-editor.js
```diff
@@ -39,6 +39,21 @@ class SourceEditor {
...
+    this.editor.swapDoc(doc);
+  }
+
+  /**
+   * Creates a CodeMirror Document
+   * @returns CodeMirror.Doc
+   */
+  createDocument() {
+    return new CodeMirror.Doc("");
+  }
```
jasonLaster: while not a big concern now, i did keep in mind the editor integration a bit here and believe we can implement similar methods if we integrate with atom, emacs, etc

public/js/utils/source-map-worker.js
```diff
@@ -127,7 +127,7 @@ function createSourceMap({ source, mappings, code }) {
   return generator.toJSON();
 }
 
-function clearData() {
+function clearSourceMaps() {
```
jasonLaster: made this fnc more clear as to differentiate with other clearX methods

public/js/components/Editor.js
```diff
@@ -174,25 +176,62 @@ const Editor = React.createClass({
...
-    } else if (!isTextForSource(sourceText)) {
-      // There are only 2 possible states: errored or loading. Do
-      // nothing except put a message in the editor.
-      this.setText(sourceText.get("error") || "Loading...");
-      this.editor.setMode({ name: "text" });
-    } else if (this.props.sourceText !== sourceText) {
-      // Only update it if the `sourceText` object has actually changed.
-      // It is immutable so it will always change when updated.
-      this.setText(sourceText.get("text"));
-      this.setMode(sourceText);
```
jlongster: Personally, I really really liked the terse-ness of the previous code so I'm a little sad to see it go. There's a bunch of things going on here so it's really nice to see it all in just 10 lines of code and it's easy to follow. I don't think we need to do anything with documents for loading/error accept make sure a fresh one exists, so you could reduce bring back the old code and just make a `this.freshDocument()` function and call that before loading in the error/loading text.

For example, if the sourceText does not exist we do need to reset the editor to `""`, I think because on navigation it gets into that state sometimes and we need to make sure it's not showing the text from before (I can't remember the exact step to reproduce this)

Pulling out the loading the actual source text makes sense though. I'm not sure if `replaceDocument` is cheap if the current doc is the same; it probably is, but if not we should also keep the cheap `this.props.sourceText !== sourceText` check.

I need to head out right now but I can provide some sample code to explain more clearly what I'm suggesting tomorrow morning (or maybe later tonight)

jasonLaster: Thanks. Looks like we're close though. I'm happy that we should be able to land something tomorrow.

public/js/utils/source-editor.js
jlongster: Definitely agree

public/js/utils/source-documents.js
jlongster: Yeah, I don't think this belongs in redux, as it's just an optimization for the editor (if we replayed actions, it would make no difference if these docs existed or not). This is also only ever used by the editor, as far as I know.

Since it is only ever used by the editor, I think it's worth trying just keep it local to the editor. The only problem is clearing it on navigation, but that might be solved by not keeping a hard reference to the documents. I think it is easiest to explain with another commit so I'll pitch you some ideas with some actual code.

public/js/components/Editor.js
jlongster: Yeah, it shouldn't take much more. I'd like to try to maintain the same code flow as before and keep the docs local, and I'll make a quick commit to show you what I mean. I'm not sure, but it might be a little cleaner.

jlongster:
I tried to move the documents into the `Editor` component using a `WeakMap` with the sourceText objects as the key, and it worked really well. However, there's one big drawback: no way to discard the document when a tab is closed. They will be disposed of on navigation and the next GC cycle happens, but naturally we want to keep the source text around even after a tab is closed.

So I don't have many more tweaks to suggest. I would like to group together the loading/error management to make that workflow a bit more clear, so I still made a commit for that: https://github.com/jlongster/debugger.html/commit/b34950400913f266f9d3370136da4f676346093f You can view the Editor.js file itself here: https://github.com/jlongster/debugger.html/blob/editor-docs/public/js/components/Editor.js

Basically, we don't really need any of the doc management for these "messages" and it's feels weird to handle loading and errors in separate places. That's my only suggestion though, the rest looks fine. Let me know what you think and if you make any changes.

jasonLaster:
> weakMap with sourceText objects 

Yeah, I thought about doing this with sourceTabs, but it seemed bad to include that data.

are you ready to approve this?

jlongster:
Yeah, sorry! Didn't know if you were going to make another pass or not.

PR #784, opened 2016-09-20
Add circle master success/fail notifications

jasonLaster:
This will ping us in #devtools-html when master fails or succeeds. 

It's helpful if we remove the protected branch.


jlongster:
Do you set up the IRC notifications on circle's side?

jasonLaster:
Yep - already added

jlongster:
Cool, looks good.

PR #786, opened 2016-09-20
Add search keycommand info above Sources. Fixes #622

### LINKED ISSUE #622: Add search keycommand info above Sources
I think that the area above Sources might just continue to evolve, but for the moment I believe it would be a good spot to educate users about "Cmd + P to search". 

Mac: 
screen shot 2016-08-30 at 11 08 55 am

Windows/Linux:
screen shot 2016-08-30 at 11 12 54 am

Dalimil:
I still need info from @helenvholmes regarding the font-size and CSS styling. 
This is what I have at the moment:
screenshot_2016-09-20_21-34-25

### DIFF
--- a/public/js/components/Editor.css
+++ b/public/js/components/Editor.css
@@ -89,6 +89,5 @@
   color: var(--theme-comment-alt);
   background-color: var(--theme-tab-toolbar-background);
   font-weight: lighter;
-  text-align: center;
   z-index: 100;
 }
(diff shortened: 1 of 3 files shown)


jasonLaster:
Thanks @Dalimil. Nice job with the Ctrl/Cmd strings :)

I'm not sure if this is in the issue, but I believe we'll take the copy out of the center pane now that it's up in the header. @helenvholmes thoughts?

PR #789, opened 2016-09-20
Upgrade Storybook

jasonLaster:
I followed the `getstorybook` steps to see if there was a better way to install storybook, ofcourse now that i did, i learned that it was just an install tool and i had been tricked into upgrading. 

That's not all bad:
* we got access to `build-storybook` which will be great for the website
* we can use their awesome suite of plugins
* we're up to date

I think at this point it would make more sense to encourage people to install storybook globally and run it that way. Then it's not a flakey step in our install process.

### DIFF
--- a/.storybook/webpack.config.js
+++ b/.storybook/webpack.config.js
@@ -1,5 +1,3 @@
-"use strict";
-
 const path = require("path");
 const webpack = require("webpack");
 
--- a/.storybook/config.js
+++ b/.storybook/config.js
@@ -1,12 +1,8 @@
-"use strict";
-
 const { configure } = require("@kadira/storybook");
 const { setConfig } = require("../config/feature");
 
 require("../public/js/lib/themes/light-theme.css");
-
 setConfig(DebuggerConfig);
-
 
 function loadStories() {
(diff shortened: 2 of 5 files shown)


.storybook/webpack.config.js
```diff
@@ -1,30 +1,28 @@
...
 
 const getConfig = require("../config/config").getConfig;
 
 module.exports = {
   resolve: {
-    alias: {
-      "devtools": path.join(__dirname, "./public/js/lib/devtools"),
-      "devtools-sham": path.join(__dirname, "./public/js/lib/devtools-sham"),
-      "sdk": path.join(__dirname, "./public/js/lib/devtools-sham/sdk")
+      alias: {
```
jlongster: Why all these changes? Looks like the indentation just got messed up, unless I'm missing something.

jlongster:
Yeah, that works for me. Can you fix the indentation changes in webpack config? (unless I'm missing something)

.storybook/webpack.config.js
jasonLaster: nice call. just cleaned it up

PR #790, opened 2016-09-21
(WIP) Update m-c deps

jasonLaster:
I ran `node bin/import-deps` to see what m-c modules we should update and found that it was mostly Reps. 

The Services shim was also updated and that is breaking our webpack config with a `prefs` loader. 

It would be nice to sort these things out so we can stay uptodate.


jlongster:
Good idea. Looks like something isn't working; most/all of the tests are timing out. We should definitely update this though.

public/js/lib/devtools/client/shared/shim/Services.js
```diff
@@ -432,35 +414,32 @@ PrefBranch.prototype = {
...
-      // Couldn't access localStorage; bail. This happens in the
-      // Firefox panel because Chrome-privileged code can't access it.
-      return;
+    if (localStorage.length === 0 && Services._defaultPrefsEnabled) {
+      /* eslint-disable no-eval */
+      let devtools = require("raw!prefs!devtools/client/preferences/devtools");
+      eval(devtools);
+      let all = require("raw!prefs!modules/libpref/init/all");
+      eval(all);
+      /* eslint-enable no-eval */
```
jasonLaster: This is the code that's failing. We'll need to see where the prefs loader is defined

jasonLaster:
Just updated this PR to only include the Reps work, which is all but three of the modules changed. 

This should be fairly easy to pull in.

jlongster:
Nice!

Issue #793, opened 2016-09-21
docs - mochitest requirements


jasonLaster:
we should still mention mercurial as a requirement
https://github.com/devtools-html/debugger.html/commit/463ae613ce26f65e39bae93d3e97bda1836914c7#diff-eb5b31b929fde934804675e923191746L11

PR #794, opened 2016-09-21
Add mercurial back to requirements for mochitests

jlongster:
Accidentally removed this entirely when removing the `hg purge` requirement.


clarkbw:
This fixes  #793

PR #795, opened 2016-09-21
Remove styles link

jasonLaster:
This removes the index.html styles link as we now have a different index in the panel, which is fine because they're doing different things.

I've heard from contributors that the error is confusing and it'd be nice to smooth out the getting started experience

### DIFF
--- a/bin/development-server.js
+++ b/bin/development-server.js
@@ -5,6 +5,8 @@
 require("babel-register");
 
 const path = require("path");
+const fs = require("fs");
+const Mustache = require("mustache");
 const webpack = require("webpack");
 const express = require("express");
 const webpackDevMiddleware = require("webpack-dev-middleware");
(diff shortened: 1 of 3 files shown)


jlongster:
Yeah, this is fine for now. We should bring it back with (I guess) a simple template system, because this breaks the workflow of running it in production mode locally.

jasonLaster:
Oh yeah - i forgot about that workflow again. One minute, let me revisit this.

jasonLaster:
@jlongster this is updated. Is it what you had in mind?

bin/development-server.js
```diff
@@ -63,7 +65,9 @@ app.use(express.static("public"));
 
 // Routes
 app.get("/", function(req, res) {
-  res.sendFile(path.join(__dirname, "../index.html"));
+  const tplFile = fs.readFileSync(path.join(__dirname, "../index.html"), "utf8");
+  const localProduction = getValue("firefox.localProduction");
```
jlongster: This looks great, except we shouldn't need a separate flag. When I'm testing production, I mean the real production environment, so you can just do `Mustache.render(tplFile, { isDevelopment: isDevelopment() })` and only render the `<link>` tag when it's not in dev. Because when it's not in dev, the styles should be separated into a CSS file

jlongster:
That's exactly right, except for one nit! The `<link>` tag should automatically appear when I flip on production mode, so we can just is `isDevelopment()` to check of that. But yes, this is great, thanks a lot!

jasonLaster:
Nice. Easy to fix

jasonLaster:
@jlongster updated

PR #796, opened 2016-09-21
Enable chrome locally

jasonLaster:
I was talking with a contributor and the question of getting setup with chrome came up and I found it was very complicated. He ran into two problems:

1. Firefox had to be on
2. enabling local.json was not intuitive

Neither of these things is a deal breaker, but for many people there's already going on when getting started and any extra steps can be intimidating.

This fixes that by enabling chrome by default and fixing the firefox connection so that it handles not connecting gracefully.

I also sped up the timeouts so that the toolbox shows pretty fast.

### DIFF
--- a/public/js/clients/firefox.js
+++ b/public/js/clients/firefox.js
@@ -63,7 +63,7 @@ function connectClient() {
     });
   }).catch(err => {
     console.log(err);
-    deferred.reject();
+    deferred.resolve([]);
   });
 
   return deferred.promise;
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -125,18 +125,6 @@ Start by running Chrome with remote debugging turned on, this command also creat
 ```
 > If this command doesn't work for your OS or Chrome version see the other [Chrome commands for running in a debuggable state](./docs/remotely-debuggable-browsers.md#chrome)
 
-Now turn on Chrome debugging in the config by [creating a local config file](#create-a-local-config-file)
-
-* Edit the `config/local.json` you just created to change the value of `chrome.debug` to be `true`
-
-```json
-"chrome": {
-  "debug": true
-}
-```
-
-* Restart your development server by typing `ctrl+c` in the Terminal and run `npm start` again
-* Reload `localhost:8000` (you should now see a Chrome tabs section)
 
 #### Node.js
 
--- a/config/development.json
+++ b/config/development.json
@@ -14,7 +14,7 @@
     "search": true
   },
   "chrome": {
-    "debug": false,
+    "debug": true,
     "webSocketPort": 9222
   },
   "firefox": {
--- a/public/js/clients/index.js
+++ b/public/js/clients/index.js
@@ -42,14 +42,9 @@ function startDebuggingTab(targetEnv, tabId, actions) {
   });
 }
 
-function connectClients() {
-  return Promise.all([
-    firefox.connectClient(),
-    chrome.connectClient()
-  ]).then(results => {
-    const [firefoxTabs, chromeTabs] = results;
-    return firefoxTabs.concat(chromeTabs).filter(i => i);
-  });
+function connectClients(onConnect) {
+  firefox.connectClient().then(onConnect);
+  chrome.connectClient().then(onConnect);
 }
 
 module.exports = {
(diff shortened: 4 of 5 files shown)


clarkbw:
With the timeouts working I think this makes sense as long as it doesn't interfere with the build / bundle process.

public/js/clients/firefox.js
```diff
@@ -57,13 +57,13 @@ function connectClient() {
     new DebuggerTransport(socket) : new WebSocketDebuggerTransport(socket);
   debuggerClient = new DebuggerClient(transport);
 
-  debuggerClient.connect().then(() => {
+  debuggerClient.connect(()=>{}, 2000).then(() => {
```
jlongster: This won't work. There were times when for some reason it would take a while to connect to firefox (say my machine is already doing something heavy, or for whatever reason) and refreshing the page a lot while developing I'd see it not connect 1/4 of the time, which is *extremely* frustrating. We need to find another way to do this, or leave it at 6 seconds.

Chrome is still second-class at this point, so I don't think we should turn it on unless it doesn't affect the Firefox workflow.

jlongster:
I don't feel good about the setup with all the timeouts. I've already run into issues when the timeouts were too short and while developing it just wouldn't connect randomly.

We need a better way to deterministically figure out if it's running. We should be able to detect if something is running on a port in the development server, right? If so, the development server should provide this info by checking the port (with some sort of system command) and letting the frontend know.

clarkbw:
Agreed that this can't get in the way of Firefox development, we can make the docs better if that's the only option.

jlongster:
We could also avoid grouping all of the Chrome and Firefox tabs in one page. If you could "switch" between those targets (basically just flipping between two pages) that would solve the issue as well.

jasonLaster:
> Chrome and Firefox tabs in one page 

This sounds like a good idea. I drew some wireframes earlier today that I can share.

I think another fix would be to connect to each client separately. Then it is just a matter of showing tabs when the connection is made. No short timeouts needed.

jlongster:
Sounds great, thanks a lot for working on this Jason!

jasonLaster:
@jlongster thanks for pushing back. I think this implementation is cleaner.

jlongster:
I don't see any changes with the timeouts. Is it different because it renders the page immediately and changes as clients are connected? That is nice, but a timeout of 2000 could still disconnect from a working Firefox instance. These timeouts are still applied if you already have selected a tab and it's in the query string, and you are doing cmd+R to refresh the page and develop on stuff. So these timeouts are tested a lot, and when it disconnects randomly (for some reason sometimes the server would take a while to respond after hammering it a lot), it's annoying.

Did a quick search and there's a couple way to see if something is running on a port: https://gist.github.com/timoxley/1689041 What we should do instead is do that in the development server and provide information to the client about what's running. We can also be more confident in our messages to the user: "Chrome definitely isn't running on port 9000. Is that the right port, or did you forget to start Chrome?" kind of stuff, instead of just wondering which things are running.

Sorry for pushing back again, but I want to make sure that developers don't hit these timeouts and think our clients are flaky... and after experiencing that I'd really like to avoid it...

jasonLaster:
> What we should do instead is do that in the development server and provide information to the client about what's running

I agree. This feels out of the scope for this work as chrome and firefox are slightly different: firefox awaits a WS connection, chrome expects a HTTP list/json calls. I just tried doing a simple server on port 6080 `python -m SimpleHTTPServer 6080` while firefox was running and did not get an error which was strange. `python -m SimpleHTTPServer 9000` did throw the correct error.

> when it disconnects randomly (for some reason sometimes the server would take a while to respond after hammering it a lot), it's annoying

That's a good point. It took me a while to grok this as when i timed some things it was fast. Also, I have not hit this flakey connection problem (and realized it). Maybe i've hit it and not connected the dots.

> Sorry for pushing back again

No worries. I dropped all of the timeout changes from before. Now both connections are back to 6 second timeouts. The only difference is that chrome and firefox are independent so waiting on chrome to connect does not slow down firefox.. This fixes the initial problem and I can file a subsequent issue for detecting chrome and firefox servers.

jlongster:
Cool, this looks great! Thanks!

PR #799, opened 2016-09-21
use the ⌘ 'place of interest' icon for mac

clarkbw:
### Summary of Changes

* Use the ⌘ symbol instead of the text `Cmd`
* Made the text in the centre pane more direct
  * `⌘+P to search for files`

### Testing

* [x] passes `npm test`
* [x] passes `npm run lint`

### DIFF
--- a/public/js/components/Sources.js
+++ b/public/js/components/Sources.js
@@ -20,7 +20,7 @@ const Sources = React.createClass({
 
   render() {
     const { sources, selectSource } = this.props;
-    const modifierTxt = Services.appinfo.OS === "Darwin" ? "Cmd" : "Ctrl";
+    const modifierTxt = Services.appinfo.OS === "Darwin" ? "⌘" : "Ctrl";
 
     return dom.div(
       { className: "sources-panel" },
--- a/public/js/components/Sources.js
+++ b/public/js/components/Sources.js
@@ -3,7 +3,7 @@ const { DOM: dom, PropTypes } = React;
 const ImPropTypes = require("react-immutable-proptypes");
 const { bindActionCreators } = require("redux");
 const { connect } = require("react-redux");
-const { Services } = require("Services");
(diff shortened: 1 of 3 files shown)


clarkbw:
Whoops, meant to include a screenshot

screen shot 2016-09-21 at 12 08 13 pm

public/js/components/Sources.js
```diff
@@ -20,7 +20,7 @@ const Sources = React.createClass({
 
   render() {
     const { sources, selectSource } = this.props;
-    const modifierTxt = Services.appinfo.OS === "Darwin" ? "Cmd" : "Ctrl";
+    const modifierTxt = Services.appinfo.OS === "Darwin" ? "⌘" : "Ctrl";
```
jasonLaster: nice - it would be nice if we had a util method `cmdString` so that we don't duplicate this everywhere.

maybe inside of utils? We can change this when we add l10n

clarkbw: Good idea! I created a `utils/search` module to hold this function

jasonLaster:
Hmm. I think it's likely others will want the CMD util.

public/js/utils/search.js
```diff
@@ -0,0 +1,9 @@
+const { Services } = require("Services");
```
jasonLaster: this seems more like a `utils/text` feature than search

clarkbw: Oh I see, no problem.

clarkbw:
@jasonLaster look good now?

clarkbw:
FYI, I looked at pulling out the code in here https://github.com/devtools-html/debugger.html/blob/master/public/js/components/RightSidebar.js#L13-L15 to the `utils/text` module as well but it wasn't quite as clean.  Feels like there should be a better higher level solution, like possibly the keyboard-shortcuts module could provide this.

Issue #791, opened 2016-09-21
Tabs - polish close button


jasonLaster:
It would be nice to adopt sublime's tab close ux

- show close label on tab hover
- show close button on close hover

![](http://g.recordit.co/U5vN9U5I4D.gif)

AndrewPomorski:
Hello, I could help with that

jasonLaster:
@AndrewPomorski would you like to grab this issue? https://github.com/devtools-html/debugger.html/issues/800

wldcordeiro:
@jasonLaster for the close button is there a prefered background or something similar to the gif? Right now the button just always shows as the x

jasonLaster:
@helenvholmes ^

@wldcordeiro I'd suggest the hover for the accordion header.

wldcordeiro:
Can be closed

PR #801, opened 2016-09-21
Polish tab close button UX

wldcordeiro:
Associated Issue: #791

### Summary of Changes

* show the close button only when hovering over tab
* highlight close button when hovering over it

### Testing

* [x] passes `npm test`
* [x] passes `npm run lint`

##### Hovering over tab

screenshot_20160921_215912

##### Hovering over button

screenshot_20160921_215916

##### Hovering over tab

screenshot_20160921_215928

##### Hovering over button

screenshot_20160921_215933

### DIFF
--- a/public/js/components/SourceTabs.css
+++ b/public/js/components/SourceTabs.css
@@ -46,28 +46,27 @@
 
 .source-tab .close-btn {
   position: absolute;
-  right: 7px;
-  top: 7px;
-  width: 9px;
-  height: 9px;
-  padding: 3px;
-  display: none;
+  right: 4px;
+  top: 3px;
 }
 
-.source-tab .close-btn:hover {
+.source-tab .close {
+  display: none;
+  width: 12px;
+  height: 12px;
+  padding: 2px;
+  text-align: center;
+  margin-top: 2px;
+  line-height: 7px;
+}
+
+.source-tab:hover .close {
+  display: block;
+}
+
+.source-tab .close:hover {
   background: var(--theme-selection-background);
   border-radius: 2px;
-}
-
-.source-tab:hover .close-btn {
-  display: inline-block;
-}
-
-.source-tab .close-btn i {
-  font-size: 5px;
-  position: absolute;
-  top: -3px;
-  left: 1px;
 }
 
 .source-header .subsettings {
--- a/public/js/components/SourceTabs.css
+++ b/public/js/components/SourceTabs.css
@@ -47,14 +47,27 @@
 .source-tab .close-btn {
   position: absolute;
   right: 7px;
-  top: 1px;
-  width: 6px;
-  height: 6px;
+  top: 7px;
+  width: 9px;
+  height: 9px;
+  padding: 3px;
   display: none;
 }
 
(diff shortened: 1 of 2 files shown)


jasonLaster:
This doesn't have a close button background, but i just played with it and it's hard to get a circular background around the close button, so definitely not a requirement.

wldcordeiro:
@jasonLaster I'm getting driven insane trying to add that! The containing `span` is above the icon for whatever reason and the `i` element that contains the svg is taller than its parent! I'll keep fiddling with it while waiting for guidance on what kind of hover bg to use.

jasonLaster:
Yea. Don't kill yourself. It's quite minor. I'm not sure either for what
it's worth

wldcordeiro:
@jasonLaster here's the hover effect with a placeholder color.

screenshot_20160921_160735

jasonLaster:
@wldcordeiro I think `var(--theme-selection-color);` would be a good value for the close button hover state.

wldcordeiro:
@jasonLaster that's what I was considering. Figured I'd double check 😄

jasonLaster:
@wldcordeiro I just noticed that the close X is now missing. I took a stab at showing the circle and this is what i came up with


```css
.close {
    width: 16px;
    display: block;
    text-align: center;
    border-radius: 10px;
    background: red;
    height: 16px;
    margin-top: 2px;
    line-height: 14px;
    color: white;
}
```

There would probably be some styling of the `fill` for the svg path so it's highlighted on hover or something like that.

wldcordeiro:
@jasonLaster bizarre... I have this.

screenshot_20160921_211018

wldcordeiro:
Curse browser rendering differences!

jasonLaster:
@wldcordeiro I hope you don't mind, but I hacked a bit on top of your work:
* play with the colors
* hover transition
* position of the close button in the tab
* general ocd :) a.k.a design is tough

Feel free to push back on any of my decisions, but felt like it would be helpful to hack on it a bit instead of describing some of the things I saw.

[patch](https://gist.github.com/jasonLaster/3d03e2f7aa3d4dab46c5641988f0223e)





screen shot 2016-09-22 at 8 52 01 am

wldcordeiro:
I'm not sure about the circular close button? I haven't noticed too many circular elements in the debugger so far so I felt like the rectangle with a more slight border-radius looked a little better. I don't mind the color change as I wasn't sure which to use anyways and the `x` changing to white on hover is a nice touch as well @jasonLaster

jasonLaster:
> circle / square

It seems like you can do either. Just about what feels right in the current project.

helenvholmes:
Yay! I am liking this a lot. Some minor nits:

Let's make the SVG on hover white, instead of that dark gray.

If you could add a ```transition: all 0.25s easeinout``` on ```.sourcetab .close```, it will match the other animations we have so far. (I wonder if that should just be made a variable...)

Thanks for the work @wldcordeiro!

jasonLaster:
🎊  Thanks for the help! This was some crazy CSS that we needed to land to get it right.

PR #802, opened 2016-09-22
Fixed Check Node Version

gauravsaini:
Causes installation failures


jasonLaster:
👍 

screen shot 2016-09-21 at 11 03 14 pm

Issue #671, opened 2016-09-07
The sidebars can be made so big that the a second horizontal scrollbar appears


captainbrosset:
STR:
- grab the left splitter,
- drag it to the right,
- continue dragging until a second scrollbar appears.

Or:
- grab the right splitter,
- drag it to the left,
- continue dragging until a second scrollbar appears.

See the screenshots below:
![capture1](https://cloud.githubusercontent.com/assets/1152698/18304181/3afd4f30-74e1-11e6-98e2-85a2b6499f6a.PNG)
![capture](https://cloud.githubusercontent.com/assets/1152698/18304182/3dbd585a-74e1-11e6-9cb8-5fa97b6ff1d3.PNG)

As you can see, there's no limit to how much you can drag, and that causes an extra horizontal scrollbar to appear.
You can make the left sidebar so big that everything else disappears from the viewport.
And you can make the right sidebar so big that the word "No Breakpoints" starts to shift and isn't centered anymore.

jasonLaster:
haha - we should probably have some sensible max-widths.

nice find

jasonLaster:
Looked into this briefly. It would be nice if step 1 was to update `h-split-box` so that we're consistent with m-c

wldcordeiro:
@jasonLaster I could look into that. Pull down `h-split-box` and diff it with `SplitBox`. I'd like to find out if it's possible to have a single `SplitBox` that does horizontal and vertical rather than two with an awkward name.

jlongster:
This was fixed by #805

PR #805, opened 2016-09-22
Split

jasonLaster:
Associated Issue: #671 

### Summary of Changes

* Updates split box
* Adds max / min widths to our panels to avoid bad behavior
* Adds overflow hidden to some elements

Edge Cases:
* can collapse panel to close to 0
* can expand the left panel without pushing the right sidebar off the window
* can expand the right panel without causing the app to add a scrollbar


public/js/components/SplitBox.js
```diff
@@ -1,60 +1,219 @@
-/* This Source Code Form is subject to the terms of the Mozilla Public
- * License, v. 2.0. If a copy of the MPL was not distributed with this file,
- * You can obtain one at http://mozilla.org/MPL/2.0/. */
-
-const React = require("react");
+const React = require("devtools/client/shared/vendor/react");
 const ReactDOM = require("react-dom");
```
jasonLaster: had to use the relative path because the absolute path wouldn't resolve

jlongster:
Shouldn't the SplitBox component now live in `lib/devtools`?

jasonLaster:
Yup, we can move it to lib/devtools and land those fixes in m-c

public/js/components/Editor.css
```diff
@@ -76,7 +76,7 @@
 }
 
 .welcomebox {
-  width: 100%;
+  width: calc(100% - 1px);
```
jasonLaster: this is a super strange issue w/ how the splitter handle is bigger than the splitter line.

PR #807, opened 2016-09-22
Refactor sourcemaps to handle more cases

jlongster:
**Warning**: Large patch incoming. I apologize for this, and it would probably be good to jump on a call tomorrow to discuss this. I maybe could have done it in a few smaller PRs, but it would have been a lot more awkward and slower.

I spent today refactoring sourcemaps. I've been dealing with sourcemaps bugs for years and while I know they'll always be annoying, I'm hoping we can have a really solid foundation from the start. We learned a lot about how we did sourcemaps before, and while the implementation was good, it didn't deal with several edge cases and solving them were more difficult than refactoring the API.

First, the big change: we no longer run sourcemaps in a worker. This is an optimization. We should focus on having *really* robust sourcemaps for now and it's simpler not to do it in a worker. If we profile and see that parsing large sourcemaps is blocking the UI, we can move the few core methods into a worker. It will be transparent to the rest of the app when we do this.

What I know works:

* Listing generated and original sources in the source tree
* Showing the source text of an original source
* Setting a breakpoint in an original source and breaking on it, even a top-level breakpoint on reload (the UI will be in a "paused" state but wait for the sourcemap to come int)

I haven't fixed up pretty-printing yet, but it shouldn't be too hard.

These are now the core APIs:

* fetchSourceMap
* getGeneratedLocation
* getOriginalLocation
* getOriginalSourceText
* installSourceMappings

If we aren't able to transfer a `SourceMapConsumer` across works, these are the only methods that need to move into a worker. It wouldn't be too hard to keep the sourcemap state in a worker; these are all promise-based APIs already.

Let's start with `fetchSourceMap`. One really important aspect is that a sourcemap is only ever requested once. It's not as simple as caching the map though. If a second request comes through while the first one is still running, it *should not* request the sourcemap a second time, which will happen if you only cache the final value. We need to cache the request itself (as a promise, ugh), and always is the same promise in future calls.

A key workflow which depends on the above behavior is top-level breakpoints in sourcemapped sources. When they are hit on page reload, the sourcemap most likely hasn't been loaded yet, but we will try to sourcemap the locations. All of these sourcemapping attempts should block on a single request and run when the sourcemap becomes available.

`fetchSourceMap` is called when a `newSource` notification comes through, and anything that blocks on a sourcemaps request can assume that all original sources exist in the system when the promise resolves.

Next, `getGeneratedLocation` returns a location in a generated source, and `getOriginalLocation` returns one in an original source. Generated = real JS file (like the bundle), and original = fake one. These both use `getSourceMap`, assuming that the sourcemap has already been requested, which should always be the case because the `newSource` notification should always come through before anything else can.

`getOriginalSourceText` takes an original source and returns the source text. If it's embedded in the sourcemap, it will just read it from there. Otherwise we need to do a network request to fetch it.

Finally, `installSourceMappings` is used by pretty-printing and allows us to "create" a sourcemap in the system and apply it to an existing source (one that shouldn't already have a sourcemap).

There is also a scheme for generated and original ids. Generated ids are just the real source ids. Original ids take this form: `generatedId + "/originalSource-" + md5(originalUrl)`. The reason for this is this means we can always go from a generated id to an original id with *just* the original URL. This is important because in `getOriginalLocation` we only get back a URL in the new location, and now we can generate the `originalId` without touching the app state and looking up the current original source.

The last point is important: I was able to remove all traces of react or redux except for one place. `getGeneratedLocation` still takes `state`, but I'd like to figure out a way to remove that too. The reason is that I think we need to host this library in the toolbox at some point, so the debugger itself doesn't manage sourcemaps. That way all tools can benefit easily from this functionality. I've been wanting that for *years*. (Locally, of course, we'd still just pull it in and use it directly).

That's a lot of information. I'm sorry for the overload, but for anyone who's interested we can talk bout it over video tomorrow. I think this is the right direction though, and there definitely things that can still probably change.


public/js/actions/sources.js
```diff
@@ -95,29 +86,30 @@ function newSources(sources) {
  * @static
  */
 function loadSourceMap(generatedSource) {
-  return ({ dispatch, getState }) => {
-    let sourceMap = getSourceMap(getState(), generatedSource.id);
-    if (sourceMap) {
+  return async function({ dispatch, getState }) {
```
jasonLaster: I like this style better than the `[PROMISE]` dispatch, in this and several cases. Thanks for cleaning this up.

public/js/utils/source-map.js
```diff
@@ -1,160 +1,206 @@
...
+  return req;
+}
+
+function getSourceMap(generatedSourceId) {
+  return sourceMapRequests.get(generatedSourceId);
 }
 
-async function getGeneratedLocation(state: AppState, location: Location) {
-  const source: any = getSource(state, location.sourceId);
+function installSourceMappings(generatedId, url, code, mappings) {
```
jasonLaster: `installSourceMappings` is somewhat vague. I think the real behavior is somewhat closer to creating a source map given code and mappings

public/js/utils/source-map.js
```diff
@@ -1,160 +1,206 @@
...
+  const generator = new SourceMapGenerator({ file: url });
+  mappings.forEach(mapping => generator.addMapping(mapping));
+  generator.setSourceContent(url, code);
 
-  if (!source) {
+  const map = SourceMapConsumer(generator.toJSON());
+  sourceMapRequests.set(generatedId, Promise.resolve(map));
```
jasonLaster: `Promise.resolve` is this here to keep the async API consistency?

public/js/utils/source-map.js
```diff
@@ -1,160 +1,206 @@
...
+  const map = await getSourceMap(generatedSourceId);
+  if (!map) {
+    return location;
   }
 
-  return location;
+  const { line, column } = map.generatedPositionFor({
+    source: originalSource.url,
+    line: location.line,
+    column: location.column == null ? 0 : location.column
```
jasonLaster: 👍

public/js/utils/source-map.js
```diff
@@ -1,160 +1,206 @@
...
+    line: line,
+    // Treat 0 as no column so that line breakpoints work correctly.
+    column: column === 0 ? undefined : column
+  };
 }
 
-async function getOriginalLocation(state: AppState, location: Location) {
-  const source: any = getSource(state, location.sourceId);
+async function getOriginalLocation(location: Location) {
+  assert(isGeneratedId(location.sourceId),
```
jasonLaster: hmm, i could be wrong here, but we might want to handle original sources by just returning the location. I'll have to think about this a bit.

public/js/utils/source-map.js
```diff
@@ -1,160 +1,206 @@
...
-}
-
-async function getOriginalSourceTexts(state, generatedSource, generatedText) {
-  if (!_shouldSourceMap(generatedSource)) {
-    return [];
+  let text = map.sourceContentFor(originalSource.url);
+  if (!text) {
+    text = (await networkRequest(
+      originalSource.url, { loadFromCache: false }
+    )).content;
```
jasonLaster: it's great that you're handling this case

public/js/utils/source.js
```diff
@@ -25,12 +25,12 @@ function trimUrlQuery(url) {
  *         True if the source is likely javascript.
  */
 function isJavaScript(url, contentType = "") {
-  return (url && /\.jsm?$/.test(trimUrlQuery(url))) ||
+  return (url && /\.(jsm|js)?$/.test(trimUrlQuery(url))) ||
```
jasonLaster: lol that we're handling jsm and not js

public/js/utils/sources-tree.js
```diff
@@ -4,11 +4,6 @@ const { isPretty } = require("../utils/source");
 
 const IGNORED_URLS = ["debugger eval code", "XStringBundle"];
 
-function isHiddenSource(source) {
-  const url = source.get("url");
-  return !url || url.match(/SOURCE/) || IGNORED_URLS.includes(url);
-}
```
jasonLaster: i'm sure you know this, but `isHiddenSource` is used in many places, which is one of the benefits of having it like that. 

e.g. do we show a source tab

webpack.config.devtools.js
```diff
@@ -12,6 +12,7 @@ const ignoreRegexes = [
 const nativeMapping = {
   "public/js/utils/source-editor": "devtools/client/sourceeditor/editor",
   "public/js/test/test-flag": "devtools/shared/flags",
+  "public/js/utils/networkRequest": ["devtools/shared/DevToolsUtils", "fetch"]
```
jasonLaster: 👍

webpack.config.devtools.js
```diff
@@ -40,7 +41,13 @@ module.exports = webpackConfig => {
...
         return;
       } else if(nativeMapping[mod]) {
-        callback(null, "var devtoolsRequire('" + nativeMapping[mod] + "')");
+        const mapping = nativeMapping[mod];
+        if(Array.isArray(mapping)) {
+          callback(null, `var devtoolsRequire("${mapping[0]}")["${mapping[1]}"]`);
+        }
+        else {
+          callback(null, `var devtoolsRequire("${mapping}")`);
+        }
```
jasonLaster: the magic only becomes more magical. *not a criticism*

jlongster: Magic everywhere! 😈 I don't love it; hoping we can figure out a better way to share modules.

public/js/utils/sources-tree.js
jlongster: It was only used in one place outside of this module, which was if it should be in the search list results. What you mean by "show a source tab"? We definitely should show a tab for any source that's displayed in the editor, even if it's an eval source (and that's what we do).

The rest of the app should work with any source, whether it's an eval source/IGNORE_URLS/etc source. Should always show up in the call stack, etc. The only consideration some components need to make it if it has a `null` URL or not, so that's all they need to check. I'm trying to be conservative on the amount of terms we use through the codebase, because "hidden" isn't very descriptive. An eval source is not "hidden", it's just not shown in the source tree. But you can definitely interact with it everywhere else.

public/js/utils/source-map.js
jlongster: You make a good point. You'll notice that `getGeneratedLocation` handles both types. This is because it doesn't matter what type you pass in there; you are getting a location from the UI and you want to always use a generated location. For `getOriginalLocation`, it's always called from places where it should be a generated location, like events from the server that give you un-sourcemapped locations.

I'm trying to make this strict because it's easy for sourcemaps to get really confusing.

But I think you're right; I can think of times when you are in a generated source and we want to allow the user to "switch" to the original location. Although, you'd already check if it is an original source, because you'd want the UX to be different...

public/js/utils/source-map.js
jlongster: Yeah. The cache holds promises to "coalesce" multiple requests if they fire at once. It's not ideal, and I would probably make these actions and do it more redux-y, but we want this library to be used elsewhere.

public/js/utils/source-map.js
jlongster: Totally. That's the least thought-through part. That name will change. I'm not sure about `create*` because I want to make it clear that it has a side-effect (adds a source map to the system).

public/js/utils/source-map.js
```diff
@@ -1,160 +1,206 @@
...
+  const existingRequest = sourceMapRequests.get(generatedSource.id);
+  if (existingRequest) {
+    return existingRequest;
+  } else if (!generatedSource.sourceMapURL || !isEnabled("sourceMaps")) {
+    return Promise.resolve(null);
   }
 
-  const generatedSourceId = getGeneratedSourceId(source);
-  const originalSource = getSource(state, generatedSourceId);
+  async function request() {
```
jlongster: Engines definitely optimize this :) I like how it's outlined and groups together the code instead of having two top-level functions, but I don't feel that strongly about it.

public/js/utils/source-map.js
```diff
@@ -1,160 +1,206 @@
...
-  return _shouldSourceMap(source);
+function originalToGeneratedId(originalId) {
+  const match = originalId.match(/(.*)\/originalSource/);
+  return match ? match[1] : null;
 }
 
-function isOriginal(originalSource) {
-  return !!getGeneratedSourceId(originalSource);
+function generatedToOriginalId(generatedId, url) {
+  return generatedId + "/originalSource-" + md5(url);
```
jlongster: See my notes in the PR description about it. It's not about escaping, just about shortening so the ids don't look crazy long if the URL is long. I'd be fine with just using `url` but I thought we might want it to look better.

We don't use random ids because we want to be able to always go from `(generatedId, url)` to `originalId` consistently. We can call it multiple times and it generates the same `originalId`. This allows functions to avoid having to dig into the redux state just to get the id.

public/js/reducers/sources.js
```diff
@@ -43,14 +43,11 @@ function update(state = State(), action: Action) : Record<SourcesState> {
...
-          ["sourceMaps", action.source.id],
-          action.value.sourceMap
-        );
-      }
-      break;
+    case "SOURCE_MAP_LOADED":
+      return state.setIn(
+        ["sourceMaps", action.source.id],
+        action.map
+      );
```
jlongster: We actually do that. The redux state isn't used for anything. I suppose for not we could just kill this code and not store it in redux for the time being.

public/js/actions/sources.js
```diff
@@ -95,29 +86,30 @@ function newSources(sources) {
...
 
-        return { sourceMap };
-      })()
+    dispatch({
+      type: constants.SOURCE_MAP_LOADED,
+      source: generatedSource,
+      map: map
     });
+
+    return map;
```
jlongster: It's not, but usually async actions return a meaningful object. With the promise middleware they always return what the promise returns. So it's just a consistent API that might allow for things we can't think of yet.

public/js/actions/sources.js
jlongster: Yeah, the promise middleware only needs to be used for UI interactions, because the main point of it is optimistic updates. A `start` action is fired when the promise starts which allows the UI to immediately update. The `[PROMISE]` syntax itself is a way of namespacing values for the middleware and is pretty common in redux (at least it used to be).

jlongster:
I've made most of the requested tweaks. Still working on pretty-printing so I'll push a new commit when I have it.

There are a few things that I need to do for pretty-printing to work. We actually need to invert the mappings that are given to us from the pretty-print worker because they are mapping the wrong direction. The server implementation does this so I just need to copy that code over too. There are a few other complications but it shouldn't take too long to get it working.

I'll also add mochitests. As a follow-up, I think I will go ahead and move this back into a worker because I forgot about the wasm/asm.js use case which might have huge multi-MB sourcemaps.

jlongster:
It's taken me a while to get this working, mainly due to breaking the `run-mochitests-docker` script because the `source-map-worker.js` file doesn't exist anymore. These tests should finally (hopefully) pass. Going to wait to merge until later tonight to check on a few things. Before I forget, I need to file issues for:

* Moving sourcemap parsing back into a worker
* Pretty-printing
* Generate a consistent ID for a given eval source (and figure out where to put it, not on the URL)
* Integrate with the native prefs system

I wrote one basic test, but unfortunately I realized that we can't run the mochitests until we turn the feature on, because it uses the `firefox-panel.json` config. And we don't have the ability to flip it on just for a test. We need to integrate with the native prefs system which would allow this.

clarkbw:
> because the source-map-worker.js file doesn't exist anymore

This won't break things, but just a heads up that it is here as well: https://github.com/devtools-html/debugger.html/blob/master/bin/make-firefox-bundle#L34

jlongster:
Yep, I updated that too, thanks!

clarkbw:
Once this merges we can circle back to #814 and rebase on top of this.

jlongster:
Alright, this should be the final code ready to be merged. Hopefully tests will still pass. Nothing really changed from before, just went through and made sure there were no `console.log`'s/etc.

PR #808, opened 2016-09-22
add more flow typing support to some utils

clarkbw:
Associated Issue: n/a

### Summary of Changes

* add two `module.name_mapper` options to the flowconfig that allows for flow to find most of the devtools modules and the Services module shim
* add flow typing support to a couple of utils files
* ended up rewriting the `getURL` function to make it clearer and simpler.  always returns an object now and we check for an empty path before continuing

### Testing

* [x] passes `npm test`
* [x] passes `npm run lint`
* [x] passes `npm run flow`

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -3,3 +3,7 @@
 .*node_modules/fbjs.*
 .*node_modules/npm.*
 .*firefox/.*
+
+[options]
+module.name_mapper='^Services$' -> '<PROJECT_ROOT>/public/js/lib/devtools/client/shared/shim/Services.js'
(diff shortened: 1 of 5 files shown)


jasonLaster:
Thanks. James is changing get URL in his PR, for the better, so we should probably wait on this. 

Really happy that were mapping now and can type more. Let's keep this up

jlongster:
I don't see too many places where this is going to conflict, actually. Just 2 places at at least, and it should be pretty small. I'm happy to rebase on top of this and fix the conflicts.

Issue #806, opened 2016-09-22
Revamp Landing Page


jasonLaster:
I'd like to take a stab at revamping the [old landing page](https://cloud.githubusercontent.com/assets/254562/18764292/b406445a-80de-11e6-8256-b84702bb25e1.png).

There are a couple goals:
* **Help new contributors initial experience**
  * The firefox cli command and about:config step can be confusing
  * The "first time" firefox page walks them through the steps.
* **Help Devs**
  * Remembering how to start browsers can be confusing
  * The browser pages have a launch button so you don't have to think about it

----

![landing page](https://cloud.githubusercontent.com/assets/254562/18764884/54264c8a-80e1-11e6-86ce-ead36e099499.jpg)

jlongster:
Sounds good. One thing to keep in mind is that we want to pull the fake toolbox out into a separate project so the console and other tools can use it when developing locally. This could get too debugger specific (the console won't be able to target other servers yet). I think if we keep the designs simple enough it'll be easier to adapt across projects, and we also don't want to take on a project that is re-creating `about:debugging` (go to `about:debugging` in Firefox). Basically: the simpler the better in my opinion, but some styling and more instructions would be good.

It might be good to pull in bgrins/linclark early on this as we want them to adopt this, and bgrins has been helping prepare other tools for running in a tab.

jlongster:
Also, you don't need to go to `about:config`. Users just need to open the devtools and turn on "Enable browser chrome and addon debugging toolboxes" and "Enable remote debugging" in the Advanced Settings. That's a lot easier than `about:config`.

jlongster:
> The browser pages have a launch button so you don't have to think about it

I'm a little hesitant to make it this complex. We'd basically be re-creating `about:debugging` at this point which is a pretty complex project. We're 2 people working on a full debugger and I'd rather spend our time on more important features.

jasonLaster:
Thanks for the feedback:

> separate project

* I believe the can be pulled into an `RDP Client` project
* I'm hesitant to call it a toolbox as that implies the tool navigation aspect as well as the client connection. *overloaded words sigh*

> about:debugging

* thanks for mentioning, I had not considered the overlap. 
  * This will be important to consider if we consider connecting to service workers, i'm not sure if there's an RDP method that exposes them. 
  * Addons is another story, and my assumption is its not relevant for the standalone case.
* the main difference is that the focus here is on multiple targets (chrome, node) and perhaps launching a browser or invoking a script.

[...]

PR #812, opened 2016-09-23
Revamp landing page

jasonLaster:
Associated Issue: #806 

### Summary of Changes

* re-skinned the landing page 
* it's much nicer now, but *not* at all different in the most common case where you're just clicking on the first firefox tab :)

**NOTE** I'm heading on PTO, so if there are any comments / feedback that needs following up on, I'd appreciate if someone acts on it. Feel free to push to this branch or close this PR and make a new one.

#### Old UI
screen shot 2016-09-22 at 4 09 56 pm


jasonLaster:
ping @jlongster, mind giving this  quick read through. Should be pretty straight forward

jlongster:
I'll take a look in a little bit, fixing up the greenkeeper PRs first

public/js/components/Info.js
```diff
@@ -0,0 +1,154 @@
...
+
+  getInitialState() {
+    return {
+      selectedPane: "Firefox"
+    };
+  },
+
+  renderTabs(tabTitle, tabs, paramName) {
+    if (!tabs || tabs.count() == 0) {
+      return null;
```
wldcordeiro: Maybe a 'no tabs open' message or something rather than empty?

public/js/components/Info.js
```diff
@@ -0,0 +1,154 @@
...
+
+    return dom.div(
+      { className: "tab-group" },
+      dom.ul(
+        { className: "tab-list" },
+        tabs.valueSeq().map(tab => dom.li(
+          { "className": "tab",
+            "key": tab.get("id"),
+            "onClick": () => {
+              window.location = "/?" + paramName + "=" + tab.get("id");
```
wldcordeiro: A template string would be a slight readability boost?

public/js/components/Info.js
```diff
@@ -0,0 +1,154 @@
...
+            } },
+          dom.div({ className: "tab-title" }, tab.get("title")),
+          dom.div({ className: "tab-url" }, tab.get("url"))
+        ))
+      )
+    );
+  },
+
+  renderFirefoxPanel() {
```
wldcordeiro: This and `renderChromePanel` are so similar I think they could be merged into a  `renderBrowserPanel` and take some arguments.

wldcordeiro:
Approved because the changes I suggested are minor in the grand scheme of things. Feel free to modify if you wish, but I'll leave this open for @jlongster to give feedback as well.

public/js/components/Info.js
jasonLaster: I'd like to, the issue I run into is flashing a message that is replaced as soon as the connection is established. It would be nice if we tracked whether a browser was connected, that way the message could read "Chrome is not connected" or something like that.

public/js/components/Info.js
jasonLaster: true, I created a `firstTimeMessage` function which handles a lot of the duplication. The rest is still there as I think there might be more divergence over time.

public/js/components/Info.js
```diff
@@ -0,0 +1,148 @@
...
+function firstTimeMessage(title, urlPart) {
+  return dom.div(
+    { className: "footer-note" },
+    `First time connecting to ${title}? Checkout out the `,
+    dom.a({ href: `${githubUrl}/CONTRIBUTING.md#${urlPart}` }, "docs"),
+    "."
+  );
+}
+
+const Info = React.createClass({
```
jlongster: `Info` isn't very descriptive, can you come up with a better name?

public/js/components/Info.js
```diff
@@ -0,0 +1,148 @@
...
+
+  renderNodePanel() {
+    return dom.div(
+      { className: "center" },
+      dom.div(
+        { className: "center-message" },
+        dom.a(
+          {
+            href: `/?ws=${document.location.hostname}:9229/node`
+          },
```
jlongster: nit: this could be collapsed in a `{ href: ... }` line

jlongster:
This looks good; there's one UX thing that I think could be better. Previously if there were no tabs it showed a different page with just a message about nothing being connected, and pointing you to various docs about connecting to different environment. Right now if it can't connect anything it just shows a blank page.

If it's not hard, can you display some sort of message if there are no tabs connected? Not sure if you want to even show the sidebar or not there, up to you.

public/js/components/Info.js
jasonLaster: Hmm, that's a good point. I believe `LandingPage` is more descriptive, while not too limiting.

jasonLaster:
> message about nothing being connected

That's a good point. I'd like to follow up on this in a second smaller issue. The *best* thing to do would be to detect connections like you mentioned in a previous PR about chrome connections. This way, we could show a descriptive message per browser.

Also, we should default to showing the connected browser first as opposed to always showing firefox.

PR #813, opened 2016-09-24
Remove superflous svg components/use in favor of `Svg`

wldcordeiro:
Associated Issue: N/A

### Summary of Changes

* remove the `Arrow` utility component as the `Svg` one does the same thing.
* remove inline svg image from `SourceTabs`

### Testing

* [x] passes `npm test`
* [x] passes `npm run lint`


wldcordeiro:
@jlongster how do I go about seeing the sources dropdown in `SourceTabs`? The second commit addressed it but I don't know how to get it to show up to confirm I didn't break its rendering.

jlongster:
@wldcordeiro You need to open enough tabs that it runs out of space, and it will group the rest of them under that dropdown.

wldcordeiro:
I see. A slightly better name maybe "overflow menu"  as the name didn't make it clear that it was for that to me. Thanks @jlongster

jlongster:
Agreed, not sure how that name made it in there. We thought about using that icon as a settings menu somewhere else, so that's probably why.

These changes look good. My only request is to remove the `Refactor:` prefix from the commits, and I wouldn't really consider this a refactor anyway. Just describe the changes in the messages. Thanks a lot!

wldcordeiro:
@jlongster I've been trying to use something like this for my commits heh https://seesparkbox.com/foundry/semantic_commit_messages but I can change the message too.

jlongster:
Since we don't do it, it doesn't really help to only have commits from one person using it. I don't personally find it that helpful, especially this early on, because one commit may contain changes that covers multiple semantic meanings, and I'm fine with that.

So it's more consistent with our current commits to remove it, and consistency is good.

wldcordeiro:
@jlongster edited the commit messages and rebased.

Issue #547, opened 2016-08-16
Toggle Breakpoints - preserve breakpoint "disabled" state


jasonLaster:
One feature that chrome has that is nice is that they preserve the breakpoint disabled state when toggling all breakpoints. 

The usecase here is that while you're debugging you've enabled/disabled some breakpoints for a workflow, then you toggle disable breakpoints so you can get to the right spot, then you enable it again.

![screen shot 2016-08-16 at 11 01 39 am](https://cloud.githubusercontent.com/assets/254562/17703682/9a9e8500-63a0-11e6-97c6-9ed846c6c57e.png)

Here's a sketch of how the work could be done
https://gist.github.com/jasonLaster/12d18102361653f2a8842e357b8bdcbc

It outlines the different files that need to be touched and has some pseudo-code.

Feel free to reach out with questions if you're interested

ThomasCrvsr:
@jasonLaster Do you mind if I try to do the implementation ? :)

jasonLaster:
@thomasCrvsr sure. Note it's marked hard so might have some traps. But should be good, will touch both chrome and Firefox

ThomasCrvsr:
@jasonLaster I'm working on the implementation of `toggleAllBreakpoints()` for Firefox.
We should get the App state in order to toggle each breakpoint individually : https://gist.github.com/jasonLaster/12d18102361653f2a8842e357b8bdcbc#file-foo-diff-L107

A proposal would be to pass the app state to `toggleAllBreakpoints()` as a second param, which will be passed to `setBreakpoint()` or `removeBreakpoint()` depending if we disable/enable breakpoints.

Have you got a better idea ? :)

jasonLaster:
@ThomasCrvsr i'm not sure exactly what you're asking here?

One thing you might have missed in the description is that we do not update the breakpoints *disable* state when we toggle all of the breakpoints. The reason for that is that it's nice to "turn off" all the breakpoints and then when you turn them on again, you still have some disabled some enabled based on your workflow.

> App state in order to toggle each breakpoint individually

In general we don't pass app state into the client methods as they should have a small API, just get the data that they need to perform some function.

![](http://g.recordit.co/zUiLDRewN5.gif)

ThomasCrvsr:
@jasonLaster Sorry, I wasn't clear ^^' I was referring to `public/js/clients/firefox/commands.js` in your gist. You call `getState()` in the function `setBreakpoint()` in the line I shared in my previous comment.

[...]

PR #814, opened 2016-09-24
547 toggle breakpoint preserve state

ThomasCrevoisier:
Associated Issue: #547

### Summary of Changes

* Implement `toggleBreakpoints()` for Chrome client
* Refactor of `source-maps` to not use the app state

### Testing

* [ ] passes `npm test`
* [ ] passes `npm run lint` (some errors to fix)

### DIFF
--- a/public/js/reducers/sources.js
+++ b/public/js/reducers/sources.js
@@ -213,6 +213,7 @@ function getSourceByURL(state: OuterState, url: string) {
   return state.sources.sources.find(source => source.get("url") == url);
 }
 
+// TODO : getSourceById does the same as getSource (cause key of sources.sources is the id)
 function getSourceById(state: OuterState, id: string) {
   return state.sources.sources.find(source => source.get("id") == id);
 }
--- a/public/js/reducers/sources.js
+++ b/public/js/reducers/sources.js
@@ -205,8 +205,8 @@ function getNewSelectedSource(state, id) : ?Source {
 // (right now) to type those wrapped functions.
 type OuterState = { sources: Record<SourcesState> };
 
-function getSourceByURL(state: OuterState, url: string) {
-  return state.sources.sources.find(source => source.get("url") == url);
+function getSourceByURL(sources: any, url: string) {
+  return sources.find(source => source.get("url") == url);
 }
 
 function getSourceById(state: OuterState, id: string) {
@@ -238,12 +238,13 @@ function getSourceMap(state: OuterState, sourceId: string) {
 }
 
 function getPrettySource(state: OuterState, id: string) {
+  const sources = getSources(state);
   const source = getSourceById(state, id);
   if (!source) {
     return;
   }
 
-  return getSourceByURL(state, source.get("url") + ":formatted");
+  return getSourceByURL(sources, source.get("url") + ":formatted");
 }
 
 module.exports = {
--- a/public/js/reducers/sources.js
+++ b/public/js/reducers/sources.js
@@ -205,15 +205,10 @@ function getNewSelectedSource(state, id) : ?Source {
 // (right now) to type those wrapped functions.
 type OuterState = { sources: Record<SourcesState> };
 
-function getSource(state: OuterState, id: string) {
-  return state.sources.sources.get(id);
-}
-
 function getSourceByURL(state: OuterState, url: string) {
   return state.sources.sources.find(source => source.get("url") == url);
 }
 
-// TODO : getSourceById does the same as getSource (cause key of sources.sources is the id)
 function getSourceById(state: OuterState, id: string) {
   return state.sources.sources.find(source => source.get("id") == id);
 }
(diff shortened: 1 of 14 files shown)


ThomasCrevoisier:
@jasonLaster Sorry for long period of inactivity.

Here is what I've done so far. Let me know if you have comments on the refactoring of `source-maps`. I thought about creating a specific module for selectors related to sources.

I still need to implement the toggle function for Firefox.

clarkbw:
@jasonLaster is out traveling for a week
@jlongster can take a look
Also, related to the source maps work there is a PR already open doing refactor work #807 that might conflict with this

jasonLaster:
This looks good. 

I like the source-map refactor. 

Also the actions look good. Nice job updating the tests as well. 

The big question will be whether we merge this before or after James's source map refactor work.

ThomasCrevoisier:
I would merge James's work before as I need to do little linting fixes, and correct the implementation for Firefox command (I finally found how to do it), and fix conflicts between this branch and the master branch.

jlongster:
I should be able to merge it today. Once you do a rebase I will review this PR.

ThomasCrevoisier:
@jlongster Okay, let me know when I can rebase.

jlongster:
@ThomasCrevoisier Just landed my PR. Sorry for all the conflicts.

I would recommend keeping your PR as focused as possible if you're implementing a single feature. I haven't gone through this PR too closely, but I'm not sure why parts of the source-map library were touched. There may have been a good reason needed for this feature, and if so, that's ok!

jasonLaster:
I believe that source maps were touched because the APIs received state.
Now that i believe that's changed the problem should be simpler.

clarkbw:
> I believe that source maps were touched because the APIs received state. Now that i believe that's changed the problem should be simpler

GO BACK TO YOUR VACATION!

public/js/reducers/sources.js
```diff
@@ -205,16 +205,12 @@ function getNewSelectedSource(state, id) : ?Source {
...
+  return sources.find(source => source.get("url") == url);
 }
 
-function getSourceByURL(state: OuterState, url: string) {
-  return state.sources.sources.find(source => source.get("url") == url);
-}
-
-function getSourceById(state: OuterState, id: string) {
-  return state.sources.sources.find(source => source.get("id") == id);
+function getSourceById(sources: any, id: string) {
```
jlongster: FWIW, I know you're going to revamp this PR, but wanted to point this out early. These are selectors. Selectors *always* receive state as the first argument, so I would have asked you to rework this PR. I haven't gone through it enough to understand why this needs changing, but I'm sure there's a reason, but there's probably a better way to do what you are doing. If you want to talk about it, feel free to hit me up as jlongster on irc.mozilla.org, or just ask questions in this PR. Or you can make a new PR and we can discuss it there :)

ThomasCrevoisier: Yep, I agree.
I didn't find a similar function to `debuggerAgent.setBreakpointsActive()` for Firefox. So the idea is to add/remove breakpoints in the Firefox client. As some functions related to sources were used in `client/firefox/commands.js`, the idea was to pass only the sources to those functions rather than the entire state.

jlongster:
Talked to Thomas on IRC and came up with something. This is just a theory that I haven't actually tried yet, but if we want this to exist at the client layer, I think it looks like this: https://gist.github.com/jlongster/d12be12928337ff69b09e8f634ecd432

The code isn't the prettiest, but basically we need to explicit track which previous bps we removed. This is because `removeBreakpoint` always removes the bp instance from the store, and when you re-add a breakpoint, you get a new `BreakpointClient`. So a new client needs to be re-added to the store.

I did this pretty quickly so I could be mistaken. We can talk about it more.

jasonLaster:
ping @ThomasCrevoisier, does this [gist](https://gist.github.com/jlongster/d12be12928337ff69b09e8f634ecd432) make sense? I want to make sure you're not blocked by anything

ThomasCrevoisier:
@jasonLaster Sorry, I've been really busy those past days. I got back to it yesterday and got some errors so I'll investigate tonight to see if it's easy to fix.

ThomasCrevoisier:
I finally got some time to work on this PR. I took the implementation proposed by @jlongster but made few modifications.

Here is the patch : https://gist.github.com/ThomasCrevoisier/c7b53b4cdc8cce6651858a5917e0f7d2

Even when we're setting a breakpoint to the exact same location it was before, the `actor` change so we need to propagate the modification in the store as well.

I wasn't able to push because of one test failing, working on that. And I also had few problems to test on Chrome.

Let me know if you have feedback on the patch 😃

jasonLaster:
Thanks @ThomasCrevoisier. Feel free to push with `--no-verify` or if you need to delete the test :)

wldcordeiro:
Closing this issue since it was reworked in #926

Issue #606, opened 2016-08-28
Breakpoints - remove breakpoint button


jasonLaster:
Let's add a remove button appear to the right of a breakpoint on hover. Having this button will make it easier for users to remove breakpoints without having to click the gutter marker.

![delete-breakpoint](https://cloud.githubusercontent.com/assets/1720093/18672205/2b0c7856-7f15-11e6-93ab-6a534ec45e55.png)

This [patch](https://gist.github.com/jasonLaster/71a2683becf26bc52247385c20a022e4) outlines what the changeset will look like. Feel free to ask any questions, especially around getting setup. 

EDIT: from @helenvholmes, added a screenshot with the proposed change mocked up

wldcordeiro:
Seems similar to the search close button issue so I'll take this one on. 👍

jasonLaster:
@helenvholmes, @clarkbw, @jlongster do you think this UI makes sense? Chrome has the remove breakpoint action in a contextmenu.

@jlongster we simplified the markup for the breakpoint, which makes it tougher to implement right-side actions. Does this still make sense considering that?

helenvholmes:
This would remove it completely from the list, separate from toggling its checkbox, correct?

The interaction makes sense to me. [Here's the link](https://dxr.mozilla.org/mozilla-central/source/devtools/client/themes/images/close.svg) to the 'x' icon I imagine we'd be using.

jasonLaster:
@helenvholmes yup. sometimes you just want it gone. Which to be fair, you can do when you click the gutter, but it would be nice to be able to delete it from the list.

jlongster:
I think it makes sense, and I think it's a good reason to avoid the breakpoint icon because it would be weird to have two icons there but only one of them actually clickable. Now that we don't have the breakpoint icon, we could put an X there.

To clean up the UI a little, maybe it would only appear on hover? That's a pretty common interaction so that it's not so cluttered but you can still do it quickly.

jasonLaster:
> weird to have two icons there but only one of them actually clickable

good point. 

> maybe it would only appear on hover

That's what I was thinking as well.

wldcordeiro:
+1+ that @jlongster the hover icon seems to make sense.

PR #815, opened 2016-09-24
Add breakpoint remove button.

wldcordeiro:
Associated Issue: #606 

### Summary of Changes

* add breakpoint remove button

### Testing

* [x] passes `npm test`
* [x] passes `npm run lint`

screenshot_20160924_143016
screenshot_20160924_143023

### DIFF
--- a/public/js/components/Breakpoints.js
+++ b/public/js/components/Breakpoints.js
@@ -10,7 +10,7 @@ const { truncateStr } = require("../utils/utils");
 const { DOM: dom, PropTypes } = React;
 const { endTruncateStr } = require("../utils/utils");
 const { basename } = require("../utils/path");
-const Svg = require("./utils/Svg");
+const CloseButton = require("./CloseButton");
 
 require("./Breakpoints.css");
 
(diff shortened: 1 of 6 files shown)


wldcordeiro:
Might be nice to consolidate the button code between `SourceTabs` and here into a new `CloseButton` component (some of the styling to handle the hover effect itself would need to be done in each containing component's styles though.

jasonLaster:
Looks good. Was thinking pulling out the close button would be nice. Maybe we want the hover background to be red?

jasonLaster:
Actually probably a soft grey for the remove button.

wldcordeiro:
I liked the consistency of having the same styling as the the tab close button.

public/js/components/Breakpoints.js
```diff
@@ -65,6 +67,10 @@ const Breakpoints = React.createClass({
     this.props.selectSource(sourceId, { line });
   },
 
+  removeBreakpoint(breakpoint) {
+    this.props.removeBreakpoint(breakpoint.location);
```
jlongster: You need to receive the event here and do `event.stopPropagation()` because it's still highlighting the breakpoint as if you click on the breakpoint item. Stopping the propagation will block the item from receiving the click event.

jlongster:
The code looks great except one small thing. Thanks!

public/js/components/Breakpoints.js
wldcordeiro: Done. Just amended the commit with this change.

Issue #680, opened 2016-09-07
Tweak pause on exceptions


jasonLaster:
It's not obvious when you toggle between the different pause on exception states.

We should so something, maybe a different color. 

Maybe a disabled state when we're not pausing on exceptions.

SezginEge:
Is there anyone working on this? If not, I can tackle it 👍

jasonLaster:
Hi @SezginEge, thanks. We would love the help. At this point, i'm not sure if we have a clear direction for how we want to handle it. 

I would love to get your ideas here as it is a challenging problem.

SezginEge:
I think, we should use toggle button with different color variants. This is more visually descriptive single state button or contextual menu. 

I also added some images about how different modern browsers implemented this button. 

On firefox developer edition, you can toggle different states only with contextual menu and this is not visually descriptive.

![debugger-options](https://cloud.githubusercontent.com/assets/6282991/18767214/7e6bbb84-8126-11e6-92a5-428c5bbad8f2.png)

On the other hand, Chrome using toggle button with different color variants and Edge using toggle button with contextual menu.

These images from Chrome's early version: 
<img width="250" alt="vgnon" src="https://cloud.githubusercontent.com/assets/6282991/18766479/21051fd8-8123-11e6-8193-2d57da46d90c.png">

<img width="250" alt="3tsmz" src="https://cloud.githubusercontent.com/assets/6282991/18766485/288f7dca-8123-11e6-833a-27bfc9dfe23b.png">

<img width="250" alt="bgzoy" src="https://cloud.githubusercontent.com/assets/6282991/18766484/2680a6a8-8123-11e6-807c-6e1b93dc26fd.png">

Edge's toggle button:

![edge](https://cloud.githubusercontent.com/assets/6282991/18766960/5dcc15d2-8125-11e6-8758-90fc0f8557ff.gif)

jasonLaster:
Thanks for the research @SezginEge, this is great!

I think Chrome's is what we're looking for:
* tooltip - (first mention current state, then mention the next state)
* colors - both chrome and edge have consistent colors (blue for caught) (purple uncaught)

our colors are here [variables.css](https://github.com/devtools-html/debugger.html/blob/master/public/js/lib/themes/variables.css).

Would you like to take a stab at implementing this?

jasonLaster:
[...]

Issue #803, opened 2016-09-22
Ignore exceptions button tooltip is incorrect and icon doesn't change state to indicate its status


captainbrosset:
STR:
- Open any page, like data:text/html,test
- In my case I tested this in Firefox devtools in the console
- Try to execute a non-existing function in the console like `foo()`

The debugger panel gets selected and pauses on this line.
But:

- If I hover over the ignore exceptions button, its tooltip says: "Pause on uncaught exceptions". Since it already paused, I would expect it to say the opposite: "Ignore exceptions" so that I know that if I click it, then it's going to ignore exceptions in the future and not automatically pause.
- If I resume, click on the button and execute `foo()` again, then the debugger doesn't pause, which is good, but the tooltip now says "Ignore exceptions", which is the state I'm already in, it should say "Pause on uncaught exceptions".

Also, the icon itself is always the same, same color, same drawing, so there's no visual way for me to tell which mode I'm in:

![capture](https://cloud.githubusercontent.com/assets/1152698/18741595/516b0d9a-80b0-11e6-9ec5-a272c9e65492.PNG)

jasonLaster:
Thanks for the feedback @captainbrosset.

This is somewhat addressed by #680. Unfortunately, I don't think it's intuitive what the language in the tooltip should be. Is it the current state, or is it what you will get if you click on it. Most of the time, it's the latter, but in this case the icon is not discriptive because all three states are the same :/

jlongster:
Closed by #816

PR #816, opened 2016-09-24
Tweak pause on exceptions

SezginEge:
Associated Issue: #680

###  Summary of Changes

* Added new svg files for each pause on exception states
* Used different classes for each state.
* Changed tooltip text.

### Testing

* [✓] passes `npm test`
* [✓ ] passes `npm run lint`

screencast


public/js/components/RightSidebar.js
```diff
@@ -144,24 +144,24 @@ const RightSidebar = React.createClass({
...
       return debugBtn(
         () => pauseOnExceptions(pause, ignore),
-        "pause-exceptions",
+        className,
         "enabled",
         tooltip
       );
     }
 
     if (!shouldPauseOnExceptions && !shouldIgnoreCaughtExceptions) {
```
wldcordeiro: I feel like this code is really duplicative for adding CSS values and a label. I think a better solution would be to remove the `pauseBtn` function and call `DebugBtn` directly and use the `classnames` helper to resolve the class to use. I'd also modify the classes to essentially be `.pause-btn.ignore`, `.pause-btn.uncaught` and `.pause-btn.all` so you can more easily use `classnames`.

wldcordeiro:
Also it appears there are linting issues here.

public/js/components/RightSidebar.js
SezginEge: I think, using "className" as a parameter name was wrong decision and also consufing. In fact, it should be named "type". I guess removing pauseBtn function and using directly DebugBtn function much proper way as you said. In this way, we dont need any css class. Svg files changes according to the type parameter.  

https://github.com/devtools-html/debugger.html/blob/master/public/js/components/RightSidebar.js#L26

wldcordeiro: Correct me if I'm wrong but isn't it the same SVG? Just different colors? Couldn't we have a single SVG and apply the color via CSS with a rule like `pause-btn path: { fill: #F0F0F0 }`

SezginEge: You are correct. 👍 Im making changes.

SezginEge:
@wldcordeiro When i run "npm run lint" command, i am not getting any error on js files. But CI founds errors. Is there could be any problem on linting tool?

wldcordeiro:
@SezginEge I had similar happen to me with a PR, I think CI might run on the transpiled code? @jlongster might know why the output differs (could be a need to `npm install/update` as the CI builds the env each time?)

SezginEge:
@jlongster Do you have any idea why i am having this issue? Before sending another commit, I want to make sure everything is okay.

jasonLaster:
Thanks guys. This PR looks good. Also nice job iterating on the solution. 

CI should be the same. If try npm update. Also jlongster can run SSH on the CI build to check.

SezginEge:
@jasonLaster Is everything okay or should i update anything?

wldcordeiro:
Try to run `npm update` and then `npm run test-all` if it's still passing locally it must be CI weirdness. Otherwise the lint preset update and you were on an older version.

wldcordeiro:
@SezginEge looks like these are the remaining lint failures. Jason is on PTO for the week.

```
  148:8   error  Expected indentation of 6 space characters but found 7  indent
  149:37  error  A space is required after ','                           comma-spacing
  157:8   error  Expected indentation of 6 space characters but found 7  indent
  158:37  error  A space is required after ','                           comma-spacing
  165:7   error  Expected indentation of 4 space characters but found 6  indent```

**Comment by SezginEge** [comment]
@wldcordeiro I tried to "run npm update" and then "npm run test-all". I didnt get any error. 😕 fyi.

**Comment by wldcordeiro** [comment]
At this point I don't know why the output differs. My last suggestion would be to nuke node_modules with `rm -rf node_modules` then running `npm install` and the test command. 😕 @jlongster any other ideas?

**Comment by jlongster** [comment]
Check your version of eslint: `eslint -v`. Mine is 2.8.0. It's common for differing versions to be different with applying rules.

Worst case, just fix the errors from CI and get CI passing and we know the code is fine. If you're going to be opening many PRs though, you probably want to fix it locally. I'm betting you have a different version of eslint overriding the local version we've installed in `node_modules`.

File: public/images/pause-circle.svg [hunk 2]
```diff
@@ -1,10 +1,7 @@
...
-    <path d="M3,3 L3,6.5" id="Line" stroke="#4990E2" stroke-width="1.15" stroke-linecap="round"></path>
-    <path d="M6,3 L6,6.5" id="Line" stroke="#4990E2" stroke-width="1.15" stroke-linecap="round"></path>
-  </g>
+<svg width="16" height="16" viewBox="0 0 16 16" xmlns="http://www.w3.org/2000/svg">
+  <path d="M10.483 13.995H5.517l-3.512-3.512V5.516l3.512-3.512h4.966l3.512 3.512v4.967l-3.512 3.512zm4.37-9.042l-3.807-3.805A.503.503 0 0 0 10.691 1H5.309a.503.503 0 0 0-.356.148L1.147 4.953A.502.502 0 0 0 1 5.308v5.383c0 .134.053.262.147.356l3.806 3.806a.503.503 0 0 0 .356.147h5.382a.503.503 0 0 0 .355-.147l3.806-3.806A.502.502 0 0 0 15 10.69V5.308a.502.502 0 0 0-.147-.355z"/>
+  <path d="M10 10.5a.5.5 0 1 0 1 0v-5a.5.5 0 1 0-1 0v5zM5 10.5a.5.5 0 1 0 1 0v-5a.5.5 0 0 0-1 0v5z"/>
 </svg>
```
jlongster: Why'd you remove `pause-exceptions.svg` and take over `pause-circle.svg`? Are you sure the latter isn't used anywhere else?

SezginEge: These files using only RightSidebar.js and Svg.js. I checked before removing or updating anything. 

I removed pause-exceptions because it sounds like refering only one exception state. I think, If we use pause-circle in there, anyone who can easily understand that we are changing icon color with css.

jlongster: Ok, I don't really care just curious. A minor nit is it's not actually a circle :) This is the icon for toggling pause on exceptions, so I think `pause-exceptions` is an appropriate and more semantic name.

But it's not a big deal. Have you figured out your local eslint issue?

SezginEge: I got it.👍 I will send another commit. 

I checked my local eslint version and figured out that it is different yours. `3.6.0` version installed on my computer. I guess this is the source of our problems :)

jlongster: I'm actually about to open a PR to update to 3.6.0 because I'm hitting another error that is solved by upgrading.

I don't know why we're having so many issues with our linter today... I'm suspicious that the above linter errors aren't from different versions. I don't know. Let's update to 3.6.0 (the latest version) and see what happens.

wldcordeiro: Odd the eslint version that installs with the project is `3.6.0` according to `npm ls --depth=0` so you should have the right version @SezginEge

SezginEge: Dunno 😖 Maybe we disscuss linter problems under a new issue.  No one will think to look here for linter issue.

SezginEge:
Np 👍 I am glad to help. 🎉 
Thank you guys for your feedback and patient! @wldcordeiro @jlongster @jasonLaster

jlongster:
Awesome, thanks a lot and thanks to the reviewers!

jlongster:
I'm going to push an update to Firefox at some point this week, maybe around Wednesday, so you will see it in Nightly then.

SezginEge:
This pr closes #680 and #803

PR #818, opened 2016-09-26
Prevent source header info from wrapping

captainbrosset:
This prevents the keyboard shortcut label in the source header from wrapping and overflowing on the list of sources below it when the sidebar is small.

With this simple CSS change, resizing the sidebar now looks like this:

sidebar

### DIFF
--- a/public/js/components/Sources.css
+++ b/public/js/components/Sources.css
@@ -8,17 +8,19 @@
 .sources-header {
   height: 30px;
   border-bottom: 1px solid var(--theme-splitter-color);
-  padding-left: 10px;
+  padding: 0 10px;
   line-height: 30px;
   font-size: 1.2em;
+  display: flex;
+  align-items: baseline;
+  justify-content: space-between;
 }
 
 .sources-header-info {
   font-size: 0.7em;
   color: var(--theme-comment-alt);
   font-weight: lighter;
-  padding-right: 10px;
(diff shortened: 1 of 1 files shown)


wldcordeiro:
@captainbrosset not sure if we use breakpoints (CSS ones heh) yet but maybe at a certain width the message should just `display: none`?

jlongster:
Neat, what does `space-between` do? Is that the same as applying `flex: 1` to all the elements?

captainbrosset:
> @captainbrosset not sure if we use breakpoints (CSS ones heh) yet but maybe at a certain width the message should just display: none? I dig this change though, combining it with the breakpoint might just make it even better 😄

Yeah, that's a good point. We should definitely do this as the label becomes useful under a certain width. I'll see if/how this can be done.

> Neat, what does space-between do? Is that the same as applying flex: 1 to all the elements?
Not exactly, flex:1 would stretch the items until they fill the available length, while space-between just distributes the available size in between the 2 items. It looks the same in the end anyway.

jlongster:
@captainbrosset Did you want to try and hide the label in this PR? I'm fine with merging this as it is.

captainbrosset:
Feel free to merge it now, I don't know when I'll have time to work on this other thing. So I'll submit a new PR when I do.
Thanks!

PR #821, opened 2016-09-26
Update eslint to 3.6.0

jlongster:
We've seen some weird issues with eslint today, and I figured it would be good to be on the latest version to make sure we're not hitting any bugs. I hit a weird bug complaining about the recent flow type changes, and I have no idea why others aren't seeing it. Updating to 3.6.0 made it go away.

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,7 +76,7 @@
     "co": "=4.6.0",
     "css-loader": "^0.23.1",
     "documentation": "^4.0.0-beta10",
-    "eslint": "^3.0.1",
+    "eslint": "^3.6.0",
     "eslint-plugin-babel": "^3.3.0",
     "eslint-plugin-flowtype": "^2.17.1",
     "eslint-plugin-mozilla": "0.0.3",


jlongster:
This is why I dislike the `^`. I wish we used `=` on all of our versions so we'd know that everyone is using exactly the same thing. It worries me too because depending on who is building it Firefox, we may get different code.

clarkbw:
> This is why I dislike the ^. I wish we used = on all of our versions

I agree.  I think we should move to `=`.  I don't really see much value in different people using different versions of the required modules.

Especially with a bot keeping tabs on the latest version.  We can upgrade as needed and upgrade everyone at once.

jlongster:
> Especially with a bot keeping tabs on the latest version. We can upgrade as needed and upgrade everyone at once.

Do you know if greenkeeper will still work? The way I understood it is it won't notify you of updates with `=` because it thinks you explicitly want to stay with that version. But I hope we could get it to work.

jlongster:
(btw, even `=` won't guarantee we all have the same thing :( Dependencies of dependencies will still probably use `^` so all the transitive dependencies may likely be different versions depending on when you `npm install`. This is a serious problem with npm. Luckily I know some people working on a new tool that we might be able to try out soon to help with this)

wldcordeiro:
@jlongster we could consider using `npm shrinkwrap` but I don't have much experience with it myself.

clarkbw:
> Do you know if greenkeeper will still work?

Actually it will still bother us about modules we locked down with `=`.  It will only ignore modules we put in the ignore list: https://github.com/devtools-html/debugger.html/blob/master/package.json#L115-L121

Their goal is to keep you aware of releases and test out your CI on all changes.  Which I think still would work well with our goal.

clarkbw:
We could turn on greenkeeper now and start trying it, especially as we move to more `=` for versions.

> It worries me too because depending on who is building it Firefox, we may get different code.

This is actually my greatest worry.

jlongster:
> we could consider using npm shrinkwrap but I don't have much experience with it myself.

Since shrinkwrap is not part of the install process, it has many issues and it gets very difficult to manage. This seems like a good summary: http://jonnyreeves.co.uk/2016/npm-shrinkwrap-sucks/

jlongster:
Yeah, we can try greenkeeper again, especially if we can tell it to ignore things.

Issue #700, opened 2016-09-09
[design] Tweak Right sidebar vertical rhythm


jasonLaster:
Each section has it's own line height typography choices. It would be nice if we had a more developed system for how this is handled. 

* I happen to think that our call stack has a lot of whitespace, but I could be persuaded if we were consistent.
* We might want to have list item borders to make the lists easier to scan. I'm not sure

![screen shot 2016-09-08 at 8 18 14 pm](https://cloud.githubusercontent.com/assets/254562/18371295/20023da0-7601-11e6-908b-9a7460a986f2.png)

SezginEge:
I have few opinions about that.

- Call stack looks huge and it takes up too much space. We may use same line-height and font-size as breakpoint items.
- List item borders could be great for differentiating items. 👍 
- We can add hover effect for each section

If nobody on working on this, I can tackle it.

wldcordeiro:
> Call stack looks huge and it takes up too much space. We may use same line-height and font-size as breakpoint items.

Indeed, the padding there should be the same as the others.

> List item borders could be great for differentiating items.

I'd prefer zebra-striping for differentiating items.

> We can add hover effect for each section

Maybe? With the zebra striping it might not be needed? I'd have to see.

wldcordeiro:
I guess zebra striping is already present but very faint.

wldcordeiro:
Made the following changes and I think it's a huge improvement.

Changed this line https://github.com/devtools-html/debugger.html/blob/master/public/js/components/Frames.css#L10

```css
padding: 0.25em
````

and this line https://github.com/devtools-html/debugger.html/blob/master/public/js/components/Frames.css#L16

```css
background-color: var(--theme-search-overlays-semitransparent);
```

The results are much nicer already.

![screenshot_20160926_154939](https://cloud.githubusercontent.com/assets/580982/18853197/2d8ff6c0-8401-11e6-8c71-0d5ddbe53453.png)

wldcordeiro:
@SezginEge sorry about kind of stealing it from under you, I wanted to see how things would look and it was such a small change. I have a branch with the two changes applied I can open a PR from.

SezginEge:
Np. Good work @wldcordeiro 👍

jasonLaster:
Handled in #876

PR #822, opened 2016-09-26
Right sidebar tweaks

wldcordeiro:
Associated Issue: #700

### Summary of Changes

* adjust padding of scope items
* change background color for zebra striping scope items to a more noticeable one.

### Testing

* [x] passes `npm test`
* [x] passes `npm run lint`

screenshot_20160926_154939

Hover effect? @SezginEge suggested this.
screenshot_20160926_185622

### DIFF
--- a/public/js/components/ObjectInspector.css
+++ b/public/js/components/ObjectInspector.css
@@ -1,10 +1,10 @@
 .arrow svg {
   fill: var(--theme-splitter-color);
-  margin-right: 5px;
+  margin-right: 3px;
   margin-top: 3px;
   transform: rotate(-90deg);
   transition: transform 0.25s ease;
-  width: 10px;
+  width: 6px;
 }
 
 .arrow.expanded svg {
--- a/public/js/components/ObjectInspector.css
+++ b/public/js/components/ObjectInspector.css
@@ -1,10 +1,10 @@
 .arrow svg {
   fill: var(--theme-splitter-color);
-  margin-right: 3px;
+  margin-right: 5px;
   margin-top: 3px;
   transform: rotate(-90deg);
   transition: transform 0.25s ease;
-  width: 6px;
+  width: 10px;
 }
 
 .arrow.expanded svg {
--- a/public/js/components/RightSidebar.css
+++ b/public/js/components/RightSidebar.css
@@ -16,7 +16,7 @@
 
 .command-bar {
   height: 30px;
-  padding: 8px 5px 10px 10px;
+  padding: 8px 5px 10px 17px;
 }
 
 .command-bar > span {


clarkbw:
@helenvholmes can you take a look?

wldcordeiro:
Yeah, the color choices I made were based on increasing visibility for the zebra striping and making a hover distinct from the active call stack element, I can adjust any of it as needed. 😄

jlongster:
I personally really prefer the old background color. To me it makes it stand out way too much. It should be a subtle hint, not actually trying to differentiate a new section. There's already a ton of information in the right sidebar so we need to be conservative about how much weight we give things. I like tightening up the padding though.

jlongster:
Although I think it needs a little more padding. It almost likes like it has no padding.

I'm not too concerned with having consistent styles across the sections. They all contain different contain laid out in very different ways. Each pane should have freedom to display information as it chooses, but conform to only basic style choices like font-size, item padding, and maybe a few others, but otherwise feel free to optimize for its content.

jlongster:
I will totally defer to whatever @helenvholmes thinks is best though. Obviously she has the most weight in this discussion.

wldcordeiro:
@jlongster the color I chose might be a bit strong but the original might as well be non-existent it's so subtle. Padding I took the cue from the surrounding accordions but could be tweaked as well.

screenshot_20160927_080134

jlongster:
Yeah, it could be a tiny bit stronger, but it needs to be subtle. Let's let Helen pick a color.

> Padding I took the cue from the surrounding accordions but could be tweaked as well.

The accordions have no padding by default, iirc. Each panel needs to specify its own padding, because it could to do anything. For example we want the background color to span the entire width of the panel, but the padding will be applied inside the item. And the breakpoints has a checkbox which is different visual treatment should not receive the same padding as text.

wldcordeiro:
@jlongster I'll defer color choices to Helen as there are three themes that need them. I chose one out of the current variables rather than picking an arbitrary color. Sorry I wasn't referring to the accordions themselves but to their contents. Breakpoints for example uses `padding: 0.25em 0`, but as you said we'll defer. I just wanted to kick off something and adjust from there.

jlongster:
Sounds good! Thanks for opening. Good point about need to consider other themes.

wldcordeiro:
Another adjustment that isn't really styling but might be worth doing is moving scopes/breakpoints/etc into their own accordions so that their headings always stay visible in the sidebar but the contents move when scrolled.

jlongster:
I'm not sure what you mean by that. Why would we have multiple accordions?

I imagine us finding a better way to display all that information, but it's going to take time and we need to be careful about it. Most alternatives I've seen so far would not actually play out in practice. The way it acts right now is intentional; we played with a few different behaviors a long time ago but they all had serious drawbacks.

helenvholmes:
Striping on the original is purposefully non-existent; @jlongster's correct, it's about not highlighting too much information at once, and [relying on white space to create line items](https://twitter.com/helenvholmes/status/742998316963790848).

Issue #700 is more about treating the padding and alignment in the right sidebar consistent; currently padding between items in accordions and left-alignments are all inconsistent. Getting everything lined up will help the vertical rhythm, as Jason put it, and consistent white space will help users differentiate what is a line item at a glance.

I think it would be best to left-align everything vertically first, and come back to visual changes like these second.

screen shot 2016-09-28 at 2 29 35 pm

screen shot 2016-09-28 at 2 30 01 pm

wldcordeiro:
Updating PR for those changes @helenvholmes mentioned. Should be ready a little later today.

wldcordeiro:
@helenvholmes I resized the arrows a bit because they were larger than what appeared in your images and adjusted their margin. Looks like the alignment is there now.

screenshot_20160929_104818

jlongster:
It feels a little inconsistent to change the size of the arrow when it's used in several other places.

Honestly, I'm still unsure exactly what the original issue is about :) The main thing that I think needs tweaking is the top/bottom padding of the call stack. We could indent the contents come to make it more "nested", but that doesn't make sense with the call stack I think. 

@helenvholmes, the designs you posted aren't what we currently implement (grouping breakpoints by file, etc), so it's a little hard to talk about what to specifically change on our end. If you have any recommendations for small tweaks the current designs, that's great! Otherwise I think we should just make the call stack items a little tighter and that's all.

wldcordeiro:
@jlongster draw a line from the pause button and the left edge of the accordion headings is left aligned to it.

wldcordeiro:
Moving the command bar over a few pixels achieves this too.

screenshot_20160929_124315

helenvholmes:
Lining it up either way is an improvement—I'd say we merge this and then debate arrow sizing in a separate pr.

Thanks for the hard work @wldcordeiro, thanks for going back and forth with us ✌🏽

jlongster:
My bad, I definitely read it wrong! I like this solution too, it's very simple. I'll merge it! Thanks for talking through it @wldcordeiro !

wldcordeiro:
Separate PR is fine for this but I realized that while the breakpoints and scopes had the left alignment right, call stack did not (aside from its padding issues.) 
screenshot_20160929_131122

jlongster:
That's because there's no checkbox or arrow icon. It would be really weird imho to give up precious spacing in the call stack for that. It's more important to see more of the function name.

Issue #823, opened 2016-09-27
Default SplitBox widths need adjusting


wldcordeiro:
Currently the two sidebars seem to take up more space than they used to. We likely need to adjust the min/max values in App.js as the editor should be about 66% of the full debugger at start.

![screenshot_20160926_190939](https://cloud.githubusercontent.com/assets/580982/18856945/ca16ce22-841c-11e6-889b-38d964300051.png)

PR #824, opened 2016-09-27
Adjust Splitbox initial widths

wldcordeiro:
Associated Issue: #823 

### Summary of Changes

* adjust the initial widths of the sidebars to give the editor more room

### Testing

* [x] passes `npm test`
* [x] passes `npm run lint`

#### Before

screenshot_20160926_190939

#### After

screenshot_20160926_211551

I tried to make the initial widths work out so that the editor is 60% of the content and the sidebars are 20% a piece.

### DIFF
--- a/public/js/components/App.js
+++ b/public/js/components/App.js
@@ -136,13 +136,13 @@ const App = React.createClass({
                               { "theme-light": !isFirefoxPanel() }) },
       SplitBox({
         style: { width: "100vh" },
-        initialSize: "20%",
+        initialSize: "300px",
         minSize: 10,
         maxSize: "50%",
         splitterSize: 1,
         startPanel: Sources({ sources: this.props.sources }),
         endPanel: SplitBox({
-          initialSize: "25%",
+          initialSize: "300px",
           minSize: 10,
           maxSize: "80%",
           splitterSize: 1,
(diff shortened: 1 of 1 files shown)


wldcordeiro:
Maybe 60/40(20*2) isn't the ratio we want, but I'd say the editor should take up more space than the sidebars by some amount.

jlongster:
It might be best to use pixel widths like before. Any reason it changed to percentages?

If I widen my screen I would want just the editor to expand. In fact, does `SplitBox` convert the percentages to pixels? Because when you are resizing it with your mouse surely it uses pixels to track the mouse exactly.

wldcordeiro:
@jlongster I'm not certain. Maybe Honza has the best answer as I believe @jasonLaster just brought it over from m-c, I figured 20/60/20 because the previous values were `300px` and the percents I found were nearly the same number (I think it's like 306 on my screen)

jlongster:
@wldcordeiro We should always be using pixels for this stuff. It happened to be the same size on your computer, but on various screen sizes the panels should be the same size.

Percentages also have introduced a problem: if you resize the left panel now, the right one moves around because it's based on a percentage. I would recommend in this PR just going back to `300px`.

wldcordeiro:
@jlongster how about the `maxSize`? I'm not sure how to set that with pixel values that would be appropriate.

jlongster:
Yeah, `maxSize` makes a lot of sense as percentages. I'm assuming that it will work to mix them.

wldcordeiro:
Nothing seemed to break leaving the `maxSizes` as is and reverting the initials to `300px`, pushing the changes now @jlongster

jlongster:
Cool, thanks!

PR #833, opened 2016-09-27
Disable breakpoint sliding on sourcemapped sources

jlongster:
After [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1305079) lands, we'll have the ability to turn off breakpoint sliding when setting breakpoints. We want to do this when setting breakpoint on sourcemapped sources (this is how the old debugger behaved) because sliding is a "nice to have" but ridiculously infuriating when it goes wrong (see [bug 1230345](https://bugzilla.mozilla.org/show_bug.cgi?id=1230345)). And sliding is almost always wrong with sourcemaps.

This is because the engine does sliding on the generated source, and then we'll sourcemap the final location back. But who knows where that maps to. I've seen cases in Firefox/Chrome where the breakpoint literally slides **up**.

We can't merge this until that bug lands, but that should be tomorrow.

### DIFF
--- a/public/js/actions/breakpoints.js
+++ b/public/js/actions/breakpoints.js
@@ -13,7 +13,7 @@ const { PROMISE } = require("../utils/redux/middleware/promise");
 const { getBreakpoint, getBreakpoints } = require("../selectors");
 
 const {
-  getOriginalLocation, getGeneratedLocation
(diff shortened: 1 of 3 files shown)


jlongster:
That bug has landed and this is harmless to merge because that option will be ignored for servers that don't support it.

PR #836, opened 2016-09-28
Updates redux promise middleware to destructuring

lourd:
Hi y'all. I just watched @jlongster's talk on this project from React Rally. Thank you for your work and for sharing it so well! It's a very cool project.

I was interested to see what async paradigm/middleware you're using for Redux, and was intrigued to find your Promise middleware. Do you find yourself wanting any more async abstractions than what the middleware and its corresponding reducer and state give you? Your solution seems very straightforward and small, especially compared to larger libraries like `redux-saga`. I'm curious about the tradeoff.

The proposed changes are just a couple simplifications using some object destructuring assignments and spread operators. I think there was an extra copy of `action` that was happening as well. It's not clear that it's needed, I don't believe it is.

I didn't see any tests related to this part of the code. Let me know what you think or if I missed something.

### DIFF
--- a/public/js/utils/redux/middleware/promise.js
+++ b/public/js/utils/redux/middleware/promise.js
@@ -7,6 +7,7 @@ const { entries, toObject } = require("../../utils");
 const { executeSoon } = require("../../DevToolsUtils");
 
 const PROMISE = exports.PROMISE = "@@dispatch/promise";
+
 let seqIdVal = 1;
 
 function seqIdGen() {
@@ -19,34 +20,35 @@ function promiseMiddleware({ dispatch, getState }) {
       return next(action);
     }
 
-    const promiseInst = action[PROMISE];
-    const seqId = seqIdGen().toString();
-
-    // Create a new action that doesn't have the promise field and has
+    // Mutate the action, removing the promise field and adding
     // the `seqId` field that represents the sequence id
-    action = Object.assign(
-      toObject(entries(action).filter(pair => pair[0] !== PROMISE)), { seqId }
-    );
-
-    dispatch(Object.assign({}, action, { status: "start" }));
+    const { [PROMISE]: promiseInst, ...actionWithoutPromise } = action;
+    action = {
+      ...actionWithoutPromise,
+      seqId: seqIdGen().toString(),
+      status: "start",
+    };
+    dispatch(action);
 
     // Return the promise so action creators can still compose if they
     // want to.
     const deferred = defer();
     promiseInst.then(value => {
       executeSoon(() => {
-        dispatch(Object.assign({}, action, {
+        dispatch({
+          ...action,
           status: "done",
-          value: value
-        }));
+          value,
+        });
         deferred.resolve(value);
       });
     }, error => {
       executeSoon(() => {
-        dispatch(Object.assign({}, action, {
+        dispatch({
+          ...action,
           status: "error",
-          error: error.message || error
-        }));
+          error: error.message || error,
+        });
         deferred.reject(error);
       });
     });


jasonLaster:
Thanks @lourd. 

Async actions are super interesting and something we're actively thinking about. We have an issue for exploring saga and wait for. There are a couple of interesting use cases. One is selecting a source to view before it's loaded like from the console or when the page is reloading and you want to show a previously selected source. 

I like your stylistic tweaks. They failed because we don't currently support object destructing, but I'd be open to addict the babel plugin. I've wanted it before for the same reasons. 

In terms of next steps, I'd be really interested to see a PR where redux saga or observsbles or wautFor was used to clean up one use case. Would you be interested in trying that? As a side note, this is one reason why open source is so useful. As a team, we haven't been able to prioritize this research, but it could help a lot.

wldcordeiro:
@jasonLaster odd vacation. 😛 

@lourd if you'd like update your PR to utilize this http://babeljs.io/docs/plugins/transform-object-rest-spread/ 

I'm for Observables! RxJS is so rad! 😄

jasonLaster:
Haha sorry. So many typos :)

public/js/utils/redux/middleware/promise.js
```diff
@@ -19,34 +20,35 @@ function promiseMiddleware({ dispatch, getState }) {
...
-    );
-
-    dispatch(Object.assign({}, action, { status: "start" }));
+    const { [PROMISE]: promiseInst, ...actionWithoutPromise } = action;
+    action = {
+      ...actionWithoutPromise,
+      seqId: seqIdGen().toString(),
+      status: "start",
+    };
+    dispatch(action);
```
jlongster: These are fine changes, but let's make every dispatch the same: it needs to add the appropriate async props related to that action:

```
const { [PROMISE]: promiseInst, ...restProps } = action;
action = { ...restProps, seqIdGen().toString() };

dispatch({ ...action, status: "start" });
```

It's very clear this way what's going on: there are 3 dispatches in this file and each of them are dispatching start/done/error actions.

public/js/utils/redux/middleware/promise.js
```diff
@@ -19,34 +20,35 @@ function promiseMiddleware({ dispatch, getState }) {
       return next(action);
     }
 
-    const promiseInst = action[PROMISE];
-    const seqId = seqIdGen().toString();
-
-    // Create a new action that doesn't have the promise field and has
+    // Mutate the action, removing the promise field and adding
```
jlongster: Don't say mutate. We aren't actually mutating the action, just changing a local variable reference. The previous comment is probably fine.

public/js/utils/redux/middleware/promise.js
```diff
@@ -19,34 +20,35 @@ function promiseMiddleware({ dispatch, getState }) {
...
     const deferred = defer();
     promiseInst.then(value => {
       executeSoon(() => {
-        dispatch(Object.assign({}, action, {
+        dispatch({
+          ...action,
           status: "done",
-          value: value
-        }));
+          value,
```
jlongster: nit: trailing comma

public/js/utils/redux/middleware/promise.js
```diff
@@ -19,34 +20,35 @@ function promiseMiddleware({ dispatch, getState }) {
...
       });
     }, error => {
       executeSoon(() => {
-        dispatch(Object.assign({}, action, {
+        dispatch({
+          ...action,
           status: "error",
-          error: error.message || error
-        }));
+          error: error.message || error,
```
jlongster: nit: trailing comma

jlongster:
I think I'm OK with object spread. It's not in an future JS standard yet, so it makes me a little uneasy, and this would be the first syntax we'd use that could not run on mozilla-central.

Actually, thinking about it more, I don't think we can do this. This file is is part of the suite of mozilla-central modules, and we want to share the exact same code at some point. We need to stick to JavaScript features that can run without any transforming for now. In the future, when more of devtools is off of mozilla-central, it'll be easier.

@lourd Some history: we have a bit of a complicated setup. This project exists on github, but most of the Firefox devtools exists in a large repo containing the entire Firefox source code. That is in mercurial and hosted on internal Mozilla servers. Devtools code hosted there has no build step: it runs as normal JS. Right now we need to share code between all these projects, so we're restricted to what JavaScript features Firefox can run.

Most of the shared files are under `public/js/lib/devtools` and the `devtools-sham` directory. The redux middleware was a special case where we needed to change several parts to get it running outside Firefox, so I think we wanted to "own" it. However, I'd like to move back to sharing this with other devtools as well. Generally, we try to touch these shared modules as little as possible (most of them being in the `public/js/lib` directory).

Regarding async abstractions, yes, I think a simple promise middleware will get you very far. I'm not convinced our needs are complex enough for something like sagas/observables. Those are great for complicated sites that have a lot of different dependent workflows. The debugger generally just displays the state from the debugger server. But if you'd like to play around with it and see how our actions might look different, feel free to!

Thanks a lot for taking the effort to open a PR! Sorry we have these constraints that make it a little more difficult to get changes through. Please keep opening PRs! (filing issues before doing large amounts of work is good too, and may avoid unnecessary work)

jlongster:
Going ahead and closing this because we should move these files back into `public/js/lib/devtools` at some point, and we can't use the object spread on m-c. Thanks a lot for your contribution, please continue opening PRs!

lourd:
On second look, maybe you want the second copy to separate the adding of the `status` property. As it's done now, when referring to `action` later it doesn't have a `status` property. I'll leave that decision to y'all.

jlongster:
@lourd I think the uniformity of the `dispatch` is good (each dispatch adds start/done/error status). This isn't going to be a bottleneck, so it's not a big deal. We embrace immutability and copy things quite a lot, and profile to see where we need to optimize.

PR #838, opened 2016-09-28
Update several dependencies

jlongster:
This supersedes #835. Only updating the ones we want to.


jlongster:
This isn't very risky because none of these are runtime dependencies, and mostly just linting.

PR #846, opened 2016-09-29
Tracking changes in MC

jbhoosreddy:
_AUTOMATED PULL REQUEST VIA BOT_
 - **M** debugger.html/public/js/lib/devtools/shared/transport/websocket-transport.js

### DIFF
--- a/public/js/components/SourcesTree.js
+++ b/public/js/components/SourcesTree.js
@@ -123,7 +123,6 @@ let SourcesTree = React.createClass({
     const { focusedItem, sourceTree, parentMap } = this.state;
 
     const tree = ManagedTree({
-      style: { overflow: "hidden" },
       getParent: item => {
         return parentMap.get(item);
       },


jbhoosreddy:
**it's obvious but, do not merge**

So here's what I got. A bot which will, for a given interval keep checking m-c for updates for files in `lib/devtools`, update the files and create a PR. Took some repeated tries to get here. Sorry if your inbox got flooded with emails (couldn't help it).

For now, I'll let it run here and make improvements as per suggestions. But it needs mercurial and git to do all of this. Right now, no need to deploy it but if we want to do so I'll look into possible ways to do it (suggestions maybe needed 😝; would docker help? EC2?).

It is also capable of listening to GitHub Web Hooks (temporarily listening to my fork via ngrok), so I think over time we can use it to do some really powerful, or simple stuff.

jlongster:
Haha this is so awesome! Great job! Does the source live in a repo or something? We probably should throw it up somewhere on heroku or something.

PR #851, opened 2016-09-30
Move sourcemap handling back to a worker

jlongster:
This moves sourcemaps back to a worker. There are several nuances here, but I need to run so I can't write a full comment yet. The most complex thing was `networkRequest`, because we can't use the native one from devtools in a worker, but we really want to do everything in a worker. Luckily, we can actually use `XMLHttpRequest` in a privileged worker and it will automatically bypass the cross-origin restriction.

I will do a more full write-up later tonight.


config/firefox-panel.json
```diff
@@ -4,6 +4,7 @@
   "logging": false,
   "clientLogging": false,
   "features": {
-    "tabs": true
+    "tabs": true,
+    "sourceMaps": true
```
jasonLaster: 👍

public/js/utils/utils.js
```diff
@@ -62,18 +61,29 @@ function endTruncateStr(str: any, size: any) {
...
+        worker.removeEventListener("message", listener);
+        if (result.error) {
+          reject(result.error);
+        } else {
+          resolve(result.response);
+        }
+      };
+
+      worker.addEventListener("message", listener);
+    });
```
jasonLaster: 👍 

I believe this is similar to how the old source map worker communicated. Out of curiosity, was the motivation for the IDs the `async` functions in the worker? I believe, the old worker API was internally synchronous.

webpack.config.devtools.js
```diff
@@ -28,6 +27,8 @@ module.exports = webpackConfig => {
     webpackConfig.output.path = path.join(__dirname, "firefox/devtools/client/debugger/new");
   }
 
+  webpackConfig.resolve.alias["public/js/utils/networkRequest"] = "public/js/utils/privilegedNetworkRequest";
```
jasonLaster: I wonder if it makes sense to move `privilegedNetworkRequest` to m-c under `devtools/shared`. 

* this module will not be used locally 
* this might be more consistent with other shim modules like Prefs where the panel implementation lives in m-c

public/js/utils/utils.js
jlongster: Yes, messages may not come back in the same order. If we want a shared util for workers, we should assume that message might not be ordered. I suppose if the worker is fully synchronous, then messages will be responded to in the same order (hadn't really realized that, good point), but now that might do fetching in the worker it's not guaranteed anymore.

We did a similar thing in mozilla-central: https://github.com/mozilla/gecko-dev/blob/master/devtools/shared/worker/worker.js#L71

You're right though that it would be fine if all the functions are synchronous. I did base a lot of this on how you did it before, so good work.

webpack.config.devtools.js
jlongster: I could move it to `lib` at least. This is something that we will be working on solely for ourselves for the time being, so I don't really want want to go through m-c and complicate that process when nobody else in m-c will even use this.

But I think it makes sense to live in `lib` at least, which indicates that it's strictly for Firefox.

jlongster:
I never fleshed out my original comment, so here's a few more details.

I was wrong about trying to only move part of the sourcemap work into a worker. I think it would be OK once we have a fully parsed & indexed sourcemap to query it on the client, but there's no good way to only do part of the work (parsing) in a worker and transfer it back. We also don't want to fetch on the main thread, because we'll be getting a huge string and copying that into a worker. Ideally, to keep things simple and fast, the worker does everything and we just send it messages.

But then we have another problem: we need a privileged cross-origin fetch. We also need to ability to bypass the browser cache when fetching. Normally we would load a privileged fetch from `DevToolsUtils` in the Firefox panel, but that's not available in a worker.

I actually had an implementation that did the fetching on the main thread, and passing the string to worker that did everything else. This was annoying though; not only was the code split between 2 places and more complicated, we were creating a huge string and copying it to the worker.

Luckily @jsnajdr pointed out that if the page is privileged (as it is in Firefox) doing a `XMLHttpRequest` is privileged and there's no cross-origin check. So we can just make our own fetch. Additionally, you can tell it to bypass the cache by setting specific channel properties, but I haven't gotten that to work yet.

So we have it all within a worker now. Just some context.

package.json
```diff
@@ -108,7 +108,7 @@
     "webpack": "1.13.1",
     "webpack-dev-middleware": "^1.6.1",
     "webpack-hot-middleware": "^2.12.0",
-    "workerjs": "^0.1.1",
+    "workerjs": "github:jlongster/workerjs",
```
jasonLaster: 👍  we should probably file a pr even though it's stale. atleast others will see that we've taken a stab at it

public/js/test/node-unit-tests.js
```diff
@@ -23,18 +22,19 @@ const webpack = require("webpack");
 const webpackConfig = require("../../../webpack.config");
 delete webpackConfig.entry.bundle;
 
+// The source map worker is compiled with webpack (and mock-require
+// doesn't work in workers) so mock it with an alias, and tweak a few
+// things to make the stub fetcher work in node.
+webpackConfig.resolve.alias.networkRequest =
+  path.join(__dirname, "/stubNetworkRequest.js");
+webpackConfig.externals = [{ fs: "commonjs fs" }];
+webpackConfig.node = { __dirname: false };
```
jasonLaster: what does this do?

public/js/test/stubNetworkRequest.js
```diff
@@ -0,0 +1,10 @@
+const fs = require("fs");
+
+module.exports = function(url) {
+  return new Promise((resolve, reject) => {
+    // example.com is used at a dummy URL that points to our local
+    // `/public/js` folder.
+    url = url.replace("http://example.com/test/", "/../js/test/mochitest/examples/");
+    resolve({ content: fs.readFileSync(__dirname + url, "utf8") });
+  });
+};
```
jasonLaster: i think we have a test utils dir

package.json
jlongster: Done

public/js/test/node-unit-tests.js
jlongster: It tells webpack not to mess with `__dirname`. Otherwise, `__dirname` is `/` in the modules, but because we are actually running in node, we want the real `__dirname`, and this lets us use it instead of webpack's overriden value.

public/js/test/stubNetworkRequest.js
jasonLaster: this dir https://github.com/devtools-html/debugger.html/tree/master/public/js/test/utils

jasonLaster: sorry, if this was not clear, i was suggesting moving this module to the dir because it's a utility

package.json
YuichiNukiyama: When I run npm install, this settings cause error. Is this Bug?

```
npm ERR! git clone --template=C:\Users\Nukiyama.Yuichi\AppData\Roaming\npm-cache\_git-remotes\_templates --mirror git@github.com:jlongster/workerjs.git C:\Users\Nukiyama.Yuichi\AppData\Roaming\npm-cache\_git-remotes\git-github-com-jlongster-workerjs-git-6570b793: undefined
npm ERR! git clone --template=C:\Users\Nukiyama.Yuichi\AppData\Roaming\npm-cache\_git-remotes\_templates --mirror git@github.com:jlongster/workerjs.git C:\Users\Nukiyama.Yuichi\AppData\Roaming\npm-cache\_git-remotes\git-github-com-jlongster-workerjs-git-6570b793: undefined
```

wldcordeiro: @YuichiNukiyama you probably don't have Git installed. I'm curious why you're commenting on a specific diff about a dependency though.

Issue #831, opened 2016-09-27
Generate a consistent display title for a given eval source


jlongster:
We used to assign eval sources a URL of `SOURCE1`, which used an incrementing id. The problem is that it's not a valid URL, and it makes it harder to differentiate eval sources. It should be easy to check if `source.url` is null in many places to change behavior.

However, we do need a consistent way to display eval sources, namely a good "title". Wherever a location is displayed, like the source tab title, breakpoint list, etc, all should display something like `SOURCE1`. I think it makes sense to parse the ID from the actor: just take the number from the very end (for `server2.conn3.child1/31` just extract `31`). It looks like Chrome just uses the `scriptId` which is the same thing.

That way we don't have to preprocess the sources at all, and we'll just provide a function for "formatting" what an eval source will be called.

jasonLaster:
I believe this is fixed

![screen shot 2016-11-01 at 2 41 42 pm](https://cloud.githubusercontent.com/assets/254562/19902386/f2d51efa-a040-11e6-9836-7d0ede7c50dc.png)

PR #860, opened 2016-10-03
Update getFilename

jasonLaster:
Associated Issue: #831 

### Summary of Changes

+ move function to utils/source
+ add a Source type
+ add support for empty url's (eval sources)

### Screenshot

screen shot 2016-10-03 at 12 05 37 pm

### DIFF
--- a/public/js/actions/breakpoints.js
+++ b/public/js/actions/breakpoints.js
@@ -36,7 +36,7 @@ function _breakpointExists(state, location: Location) {
 }
 
 function _getOrCreateBreakpoint(state, location, condition) {
-  return getBreakpoint(state, location) || { location, condition };
+  return getBreakpoint(state, location) || { location, condition, text: "" };
 }
 
 /**
--- a/public/js/actions/types.js
+++ b/public/js/actions/types.js
@@ -1,5 +1,9 @@
 // @flow
 
+import type { Source } from "../flow-types";
+
+export type { Source };
+
 /**
  * Flow types
  * @module actions/types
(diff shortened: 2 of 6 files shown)


public/js/actions/breakpoints.js
```diff
@@ -36,7 +36,8 @@ function _breakpointExists(state, location: Location) {
 }
 
 function _getOrCreateBreakpoint(state, location, condition) {
-  return getBreakpoint(state, location) || { location, condition };
+  const text = "";
+  return getBreakpoint(state, location) || { location, condition, text };
```
jasonLaster: fixes a small flow typing issue

public/js/actions/types.js
```diff
@@ -1,5 +1,9 @@
 // @flow
 
+import type { Source } from "../flow-types";
+
+export type { Source };
```
jasonLaster: I believe this is the *best* way to export an imported flow type.

clarkbw: That's an interesting pattern I haven't seen before.

public/js/utils/source.js
```diff
@@ -32,7 +36,29 @@ function isPretty(source: {url: string}): boolean {
...
+  }
+
+  let name = url;
+  const m = url.toString().match(/.*\/(.+?\..*$)/);
+  if (m && m.length > 1) {
+    name = m[1];
+  }
+
+  return endTruncateStr(name, 50);
+}
```
jlongster: We need to use this in other places as well, like the call stack, right? Can you add that to this PR, so we can be consistent? The call stack includes the location of the file for each frame.

You can also use `basename` from public/js/utils/path.js instead of the regex matching here.

jasonLaster: Great suggestion.

jasonLaster:
@jlongster updated the PR.
screen shot 2016-10-06 at 9 07 18 am

jasonLaster:
@jlongster ping

jlongster:
Awesome, thanks!

PR #861, opened 2016-10-03
Run unit tests on prepush

jasonLaster:
### Summary of Changes

* adds `npm run test` to the `prepush` script
* replaces `[].includes` with `_.includes` in sources-tree because node < 6 does not support `[].includes`. As a side-note, we run the unit tests on CI with node 6.3, so we don't run into this issue there.

### DIFF
--- a/public/js/utils/sources-tree.js
+++ b/public/js/utils/sources-tree.js
@@ -102,7 +102,7 @@ function getURL(source: TmpSource): { path: string, group: string } {
 function addToTree(tree: any, source: TmpSource) {
   const url = getURL(source);
 
-  if (IGNORED_URLS.includes(url) ||
+  if (IGNORED_URLS.indexOf(url) != -1 ||
       !source.get("url") ||
       isPretty(source.toJS())) {
     return;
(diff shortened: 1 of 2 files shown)


clarkbw:
Note to others, that if you're doing the upstream / downstream fork thing and are just pushing in the latest changes of master you can skip the lint and unit tests step by pushing with the `--no-verify` option.

public/js/utils/sources-tree.js
```diff
@@ -102,7 +103,7 @@ function getURL(source: TmpSource): { path: string, group: string } {
 function addToTree(tree: any, source: TmpSource) {
   const url = getURL(source);
 
-  if (IGNORED_URLS.includes(url) ||
+  if (includes(IGNORED_URLS, url) ||
```
jlongster: Can we just use `indexOf` instead of pulling in lodash again? I'd like to minimize the lodash deps and over time see if we really need it.

jlongster:
We can try it. It make it harder to push WIP commits to branches, but I guess if you remember `--no-verify` it's not a huge deal.

public/js/utils/sources-tree.js
jasonLaster: Switching it to `indexOf` as this is a small nit.

I can see the perspective of minimizing lodash's API to helpful utility methods. With that said, I also think lodash is helpful smoothing over support for native methods while vendors catch up. Using lodash includes is fairly safe as we could safely replace it with native support when we feel comfortable bumping Node support to v6.

Issue #864, opened 2016-10-04
Files list (left panel) can't be scrolled with a long list of files.


karlcow:
STR:
1. With a clean profile of Firefox Nightly 52.0a1 (2016-10-03) (64-bit)
2. Enter a URL with a lot of JS, for example https://www.nitori-net.jp/store/ja/ec/7514915
3. Open debugger panel
4. Put your cursor above the left side aka [Sources] (with two fingers try to scroll up and down). 

Expected: List of files is scrolling
Actual: files list is fixed, files out of sight can't be reached (except if closing some of the domains with the arrow widgets)

wldcordeiro:
@karlcow thanks for this issue, I've found the cause of this and will be putting a PR in with a fix. Nightly will likely remain broken however for a week or so to wait for some other changes.

jasonLaster:
Fixed in #865

PR #865, opened 2016-10-04
Fix Source Tree Scrolling

wldcordeiro:
Associated Issue: #864 

### Summary of Changes

* remove an inline style that was overriding the proper tree overflow

### Testing

* [x] passes `npm test`
* [x] passes `npm run lint`


jasonLaster:
As a side note, i believe I removed the horizontal scrolling because at the time i thought that we wanted people to expand the sidebar. After thinking about it more and looking at other editors + devtools, it's clear people expect to be able to scroll aswell.

Issue #869, opened 2016-10-04
Search - disable editor cmd-f shortcut


jasonLaster:
The firefox panel editor currently shows the default code search UI for cmd-f. We should disable it like we do in the old debugger [here](https://dxr.mozilla.org/mozilla-central/source/devtools/client/debugger/debugger-view.js#286-290).

----

Note, we had previously discussed pulling in the codemirror.bundle and building on top of this UX for the new debugger, but after investigating it looks like we should probably just disable the UX. 

We might want to pull down the codemirror bundle at another point, but I don't believe it will help us here.

PR #870, opened 2016-10-04
Disable search key

jasonLaster:
Fixes: #869

### Summary of Changes

Disables the keyboard shortcut that triggered the search UI in the firefox panel.

### DIFF
--- a/public/js/components/SourceFooter.js
+++ b/public/js/components/SourceFooter.js
@@ -116,12 +116,16 @@ const SourceFooter = React.createClass({
 
     this.keyShortcutsEnabled = true;
     const shortcuts = this.context.shortcuts;
-    shortcuts.on("Cmd+f", this.focusSearch);
+    if (isEnabled("search")) {
+      shortcuts.on("Cmd+f", this.focusSearch);
+    }
   },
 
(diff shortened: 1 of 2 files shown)


jlongster:
I thought we talked about using this until we have a good enough custom code search. All we have to do it forward ctrl+f to the editor and we have basic code search. Seems like a tiny thing to do which will give us a week or two to build our own.

jasonLaster:
Good call @jlongster. This should be up to date

public/js/components/SourceFooter.js
```diff
@@ -103,6 +103,10 @@ const SourceFooter = React.createClass({
   },
 
   focusSearch(shortcut, e) {
+    if (!isEnabled("search")) {
+      return true;
+    }
+
```
jlongster: If you keep it here, this function is small so I wouldn't return early. Just wrap all the code; it's easier to see the code flow.

jlongster:
Cool, thanks. This looks good, just a small nit.

public/js/components/SourceFooter.js
jasonLaster: good call

PR #872, opened 2016-10-04
Update readme screenshot

clarkbw:


clarkbw:
Weird, that was the same error I have been seeing on a totally separate branch.  Absolutely nothing to do with this change.

```
Error response from daemon: Cannot start container cfa69171589d7269cba0de2638836e3185a9cb9e4b59e42bd7f440e97790ba5a: /dev/mqueue is not mounted, but must be for --ipc=host

./bin/run-mochitests-docker returned exit code 1
```

clarkbw:
Trying something in #874 that could fix this.

PR #876, opened 2016-10-04
Vertical Rhythm, #700

helenvholmes:
Associated Issue: #700 

### Summary of Changes
Very small updates to the right sidebar so everything is left-aligned, including:

* Aligning debugger actions with accordion
* Aligning breakpoints and frames with their accordion headers
* Using @jasonLaster's self-proclaimed ugly hack to fix the splitter not extending to bottom of right pane
* Declaration of ```12px``` to fix the ```<message-box>``` styling issue in which right-pane elements sometimes appeared 1px larger than their headers

### Testing

* [x] passes `npm test`
* [x] passes `npm run lint`

### Screenshots/Videos
screen shot 2016-10-04 at 3 46 27 pm

One thing this PR _doesn't_ do (but I think it would be nice if it did) is change the event scopes to no longer be bolded (but it seemed like that was library code from FF that I didn't want to touch).

### DIFF
--- a/public/js/components/Accordion.css
+++ b/public/js/components/Accordion.css
@@ -28,5 +28,6 @@
 }
 
 .accordion ._content {
-  border-bottom: 1px solid var(--theme-splitter-color);;
+  border-bottom: 1px solid var(--theme-splitter-color);
+  font-size: 12px;
 }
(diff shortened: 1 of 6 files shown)


jasonLaster:
> event scopes is bolded

I agree. I'll create an issue here to investigate, but you're right that it's likely a question of updating the Event Rep.

Issue #685, opened 2016-09-07
Add ability to turn on/off sourcemaps in the UI


jlongster:
We need to work on how to manage settings for the debugger. The most crucial setting is sourcemaps: it's very common to turn off sourcemaps when they are just not working (could be the sourcemap's fault, not our implementation). Adding a way to toggle them would be great. Maybe we don't even need to make a whole settings panel, but turning on/off sourcemaps could be first-class as a button somewhere, I don't know.

clarkbw:
Maybe for now, we could add a toolbar that matches the editor to the bottom of the sources panel with the option to turn off source maps.  

__yellow for emphasis__

![screen shot 2016-09-07 at 12 55 54](https://cloud.githubusercontent.com/assets/2134/18326419/7f9ec1fa-74fa-11e6-8f18-db4eb716642a.png)

But this is for @helenvholmes to figure out.

jlongster:
That's a neat idea. I like the idea of embracing the fact the sourcemaps are easily faulty and making it easy for the user to flip them on/off. It's worth trying something like that and see what user think about it.

helenvholmes:
I actually think it might fit better here, since there's a) already a bar b) that sources sidebar normall is half that height and has a lot of things in it:

<img width="727" alt="screen shot 2016-09-07 at 4 20 23 pm" src="https://cloud.githubusercontent.com/assets/1720093/18327118/02c907ea-7517-11e6-9ed5-4b84061f364d.png">

I should stress I think this is more of a temporary fix, though. As features to the debugger get added I imagine we're going to need this submenu again:

<img width="349" alt="screen shot 2016-09-07 at 4 22 07 pm" src="https://cloud.githubusercontent.com/assets/1720093/18327166/3fe5e49a-7517-11e6-843a-90c3a34545ee.png">

I'm working on some more holistic designs for Settings, but I don't think they will make that submenu go away.

Debugger submenu: relatively quick actions that don't need UI, "Edit List" as last item
Settings: hitting "Edit List" deeplinks to Debugger section of Settings

jlongster:
It might be worth going ahead and bringing back back that three-dot menu with a single item under it, not sure. That's still a pretty accessible place to toggle them.

jasonLaster:
I think the submenu is probably the best place to add some quick settings:

I see the source footer as a place for a single source's actions: pretty print, black box, maybe search...

I think the sources sidebar could work for source maps, but i agree that it could get busy

[...]

Issue #832, opened 2016-09-27
Use native prefs in the Firefox panel


jlongster:
So far we have yet to hook up to the native prefs when in the panel. Even if we think it's not as important as before, I think we're going to want a pref to enable/disable sourcemaps, and it should definitely persist across sessions.

We may decide it's not a blocker. Regardless, we should do it soon.

jlongster:
We'll add a "Debugger" section to the preferences panel to add a "Toggle Sourcemaps" option.

bgrins:
There's a shim that can be used for require("Services") that will give the same prefs API: https://dxr.mozilla.org/mozilla-central/source/devtools/client/shared/shim/Services.js

bgrins:
Note you will probably want to set `Services._defaultPrefsEnabled` to false so that it doesn't try to read the m-c pref files: https://dxr.mozilla.org/mozilla-central/source/devtools/client/shared/shim/Services.js#425

clarkbw:
> We'll add a "Debugger" section to the preferences panel to add a "Toggle Sourcemaps" option.

Yep, something like this:

<img width="919" alt="screen shot 2016-09-27 at 11 45 06 am" src="https://cloud.githubusercontent.com/assets/2134/18887163/e3f874d6-84a7-11e6-90c9-a850d2781b3b.png">

clarkbw:
This is being done in #685

PR #879, opened 2016-10-04
Add shouldSourceMap util

jasonLaster:
Associated Issue: #832

### Summary of Changes

This adds the utility `shouldSourceMap`, which checks the debugger pref for source maps. 

+ add prefs utility for tracking debugger prefs
+ add prefsHelper for cleaning up utils


public/js/utils/source-map.js
```diff
@@ -7,11 +7,16 @@ const { getSource } = require("../selectors");
...
 const { isJavaScript } = require("./source");
 const assert = require("./assert");
+const { prefs } = require("./prefs");
 
 import type { Location } from "./actions/types";
 
 let sourceMapRequests = new Map();
 
+function shouldSourceMap() {
+  return isEnabled("sourceMaps") && prefs.sourceMapsEnabled;
```
jlongster: I don't really like having two pref systems. Eventually we'll have multiple duplicate prefs and need to do double-checks in lots of places.

I'd much rather go ahead and re-haul our config system to be based on prefs. `isEnabled` is supposed to be the thing that we can call directly without creating these functions.

I don't know what it looks like. Maybe our config system reads from prefs on startup, and also attaches pref observers so it can update the config when it's changed. What do you think?

jasonLaster: Thanks for bringing this up. We've talked about incorporating Prefs for a couple of months and it's nice that we're at the point where we can act on it.

I don't think this issue needs to block on re-hauling our config system. We've seen in the past that the config system is surprisingly gnarly and I want to take the time to think it through.

I like the idea of reading Prefs at startup and using `isEnabled` and `getValue`.

We will likely need to keep JSON configs as some of the config values are used for the node scripts.

jlongster: Sure, that's fine. Honestly I think it might be good if we eventually just use the `Prefs` API itself. It basically works with json files, as far as I know, we can load in whatever json files locally we want. This way in the Firefox panel feature flags are real prefs, and you can toggle them on and off, which is how we've previously been able to with in the panel.

It would be harder to create new flags, yeah, but I think we should also avoid using feature flags for things that should just sit on a branch and be developed a little more too.

jlongster:
This is using the existing sourcemap pref, right? I don't think we want to use that, as I mentioned in the bugzilla bug. We need a new pref so we don't collide. Right now, if people turn that pref on I think the server will start sourcemapping as well.

public/js/utils/source-map.js
jasonLaster: > can load in whatever json

that's a good point. 

> avoid using feature flags for things that should just sit on a branch

we've had this conversation a couple of times, so i'm going to move it to an issue for discussion (out of scope here). It doesn't have to be long lived, but it would be nice to have a couple understood rules of thumb.

jlongster:
What do you think about using a new pref? I'm on the fence about it. We always force sourcemapping off on the server when connecting the new frontend, so the pref doesn't matter. I was thinking we might connect to an old server, but that actually would be fine, as long as when we are using the new debugger we always send `false` for `useSourceMaps` when connecting to the server.

Thinking more about it, maybe we don't need a new pref. Sorry... can you think of any cases where it would break/cause confusion?

jasonLaster:
@jlongster ping. should be ready.

jlongster:
Looks good!

Uh oh, I see some errors coming from `source-map-worker.js`. I'll try to look at it later tonight.

jasonLaster:
I pulled in our latest m-c dependencies so that I could get the updated Services module, which is easier to work with.

jlongster:
@jasonLaster what is the status of this? Do you know what it takes to get the prefs working?

jasonLaster:
It's close. It works in the Pane, but a couple tests are failing. Feel free
to grab, otherwise I can look tomorrow.

packages/devtools-modules/client/shared/shim/Services.js
```diff
@@ -416,51 +398,47 @@ PrefBranch.prototype = {
...
-      if (localStorage.length === 0) {
-        // FIXME - this is where we'll load devtools.js to install the
-        // default prefs.
-      }
-    } catch(e) {
-      // Couldn't access localStorage; bail. This happens in the
-      // Firefox panel because Chrome-privileged code can't access it.
-      return;
+  _initializeRoot: function() {
+    if (localStorage.length === 0 && Services._defaultPrefsEnabled) {
```
jasonLaster: This is called from `utils/prefs` in the unit tests and blows up.

packages/devtools-sham-modules/client/shared/key-shortcuts.js
```diff
@@ -2,12 +2,10 @@
...
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-"use strict";
-
-
-const { Services } = require("devtools-modules");
+const { Services: { appinfo }} = require("devtools-modules");
 const EventEmitter = require("../../shared/event-emitter");
-const isOSX = Services.appinfo.OS === "Darwin";
+const isOSX = appinfo.OS === "Darwin";
```
jasonLaster: this also blows up in the unit tests because `appinfo` uses window.navigator under the hood

jlongster:
> I'd like to get your thoughts on how to get this patch to pass the unit tests. The new services module has some window objects that we might need to either mock or guard against

I really hope we can avoid jsdom. Mainly because I'm not *that* interested in component unit testing (the craze these days is [snapshot testing](https://facebook.github.io/jest/blog/2016/07/27/jest-14.html) which makes a whole lot more sense to me). So otherwise, it'd be a shame to have all the overhead of jsdom just for a few APIs.

We don't really need Services in the unit tests, right? This is the list of files that comes up when searching for `Services`:

```
./components/RightSidebar.js:13
./components/RightSidebar.js:14
./components/RightSidebar.js:15
./test/mochitest/browser_dbg-chrome-create.js:26
./test/mochitest/browser_dbg-chrome-create.js:37
./test/mochitest/browser_dbg-chrome-create.js:44
./test/mochitest/browser_dbg-console.js:8
./test/mochitest/browser_dbg-console.js:29
./test/mochitest/head.js:36
./test/mochitest/head.js:40
./test/mochitest/head.js:42
./test/mochitest/head.js:524
./utils/text.js:3
./utils/text.js:6
```

Since we are really only using unit tests to test actions and reducers, I don't know where Services is being pulled in. Maybe `utils/text.js`? If so that's a very tiny use case and we could probably get rid of that dependency for unit testing.

So the question is: why are unit tests pulling in Services?

jasonLaster:
> why are unit tests pulling in Services

I agree with this line of thinking. Hmm, we're likely creating a new pref in `js/utils/pref` now as well.

jlongster:
If we have to mock a lot more, maybe jsdom is the way to go. Looking at your patch though, you had to mock out `localStorage` anyway. The only other API you mentioned is `navigator` which would also be easy to mock ourselves. But maybe there's more?

packages/devtools-modules/shared/flags.js
```diff
@@ -1 +1,21 @@
...
+  });
+}
+
+makeWritableFlag(exports, "wantLogging");
+makeWritableFlag(exports, "wantVerbose");
+
+// When the testing flag is set, various behaviors may be altered from
+// production mode, typically to enable easier testing or enhanced
+// debugging.
+makeWritableFlag(exports, "testing");
```
jasonLaster: pulled in from `import/deps`

packages/devtools-sham-modules/sham/services.js
```diff
@@ -1,68 +0,0 @@
-/*
```
jasonLaster: we're now using `devtools-modules` Services

packages/devtools-sham-modules/shared/client/main.js
```diff
@@ -43,7 +41,7 @@ function eventSource(aProto) {
    *        is added more than once, it will be called once per
    *        addListener call.
    */
-  aProto.addListener = function (aName, aListener) {
+  aProto.addListener = function(aName, aListener) {
```
jasonLaster: some inconsequential formatting from eslint...

packages/devtools-sham-modules/shared/transport/websocket-transport.js
```diff
@@ -69,7 +69,7 @@ WebSocketDebuggerTransport.prototype = {
     }
 
     let object = JSON.parse(data);
-    this.emit("onPacket", object);
+    this.emit("packet", object);
     if (this.hooks) {
       this.hooks.onPacket(object);
     }
```
jasonLaster: `import/deps`

public/js/actions/sources.js
```diff
@@ -35,7 +36,7 @@ const {
  */
 function newSource(source) {
   return ({ dispatch, getState }) => {
-    if (isEnabled("sourceMaps")) {
+    if (shouldSourceMap()) {
```
jasonLaster: lol, this is the gist of the patch...

public/js/main.js
```diff
@@ -16,6 +16,9 @@ const selectors = require("./selectors");
 
 const App = require("./components/App");
 
+const { Services } = require("devtools-modules");
+Services._defaultPrefsEnabled = false;
```
jasonLaster: needed to prevent reading from the default raw!json prefs

public/js/test/node-unit-tests.js
```diff
@@ -2,12 +2,18 @@
...
 
 const glob = require("glob").sync;
 const path = require("path");
 const Mocha = require("mocha");
 const minimist = require("minimist");
 
+// Mock various functions. This allows tests to load files from a
+// local directory easily.
+mock("devtools-network-request", require("../../../packages/devtools-network-request/stubNetworkRequest"));
+mock("../utils/prefs", { prefs: { clientSourceMapsEnabled: true }});
```
jasonLaster: Needed to prevent reading from localStorage

public/js/utils/source-map.js
```diff
@@ -29,6 +30,10 @@ function destroyWorker() {
   }
 }
 
+function shouldSourceMap() {
+  return isEnabled("sourceMaps") && prefs.clientSourceMapsEnabled;
+}
```
jasonLaster: other part of the initial patch.

jasonLaster:
Hmm, surprised that mochitest failed on CI. It's passing locally for me.

jasonLaster:
@jlongster I believe this *should* work if you apply the m-c [patch](https://bugzilla.mozilla.org/show_bug.cgi?id=1307580) for the source map pref. It runs successfully locally. I forgot how we want to sequence this, it should be fine to land that patch, and then re-build the image.

clarkbw:
This will fix #685

public/js/main.js
jlongster: Don't we need this turned on though if we're in the Firefox panel, and this file still runs there?

jlongster:
Hard to review everything here since there are lot of m-c changes, but I saw one thing. Otherwise looks good.

Issue #875, opened 2016-10-04
The breakpoint gutter does not line up with the editor line


jasonLaster:
Here's a good screenshot that illustrates where the gutter is slightly shorter than the line.

![screen shot 2016-10-04 at 11 33 32 am](https://cloud.githubusercontent.com/assets/254562/19080806/46041184-8a26-11e6-8a40-b4b7f2a1e1f1.png)

monicau:
Hi Jason, I don't see any "first-timers-only" issues open at the moment. Can I grab this one?

wldcordeiro:
@monicau Feel free to jump on this, the `difficulty: easy` tag generally means it is safe for first-timers as well. Feel free to ask any questions.

jlongster:
Can't repro. When filing bugs, please state which browser and version you are using. 

<img width="606" alt="screen shot 2016-10-04 at 5 43 11 pm" src="https://cloud.githubusercontent.com/assets/17031/19093688/1adcde26-8a5a-11e6-9ef2-77d10d163eab.png">

jasonLaster:
Thanks @monicau. 

Sorry @jlongster, I saw this in Chrome.

monicau:
I modified the CSS to fix the gutter height, I also realigned the breakpoint svg arrow which also wasn't the same height as the line number highlighting.  When doing the unit tests, there were some warnings.  Is it something on my end? 
![debuggerunittestresult](https://cloud.githubusercontent.com/assets/2848472/19125905/cd06108e-8b06-11e6-9f25-0abbfd972648.jpg)

wldcordeiro:
@monicau Those aren't caused by you, they're appearing currently in master. 👍

monicau:
Thanks, that makes sense!  I created a pull request.

jasonLaster:
Thanks @monicau. I just created an issue for cleaning this up :) 

https://github.com/devtools-html/debugger.html/issues/886

PR #885, opened 2016-10-05
Fix breakpoint arrow and gutter heights

monicau:
Associated Issue: #875

### Summary of Changes

* Fixed breakpoint gutter height
* Realigned breakpoint arrow svg  

### Testing

* [x] passes `npm run lint-css`

### Screenshots

debuggerchrome
debuggerchromebreakpoint
debuggerfirefox
debuggerfirefoxbreakpoint

### DIFF
--- a/public/js/components/Editor.css
+++ b/public/js/components/Editor.css
@@ -24,10 +24,11 @@
 
 .editor.new-breakpoint svg {
   fill: var(--theme-selection-background);
-  height: 18px;
+  width: 60px;
+  height: 12px;
   position: absolute;
-  top: 0;
-  right: -8px;
+  top: 1px;
+  right: -4px;
 }
 
 .editor.new-breakpoint.breakpoint-disabled svg {
@@ -41,6 +42,11 @@
   background-color: var(--theme-body-background);
 }
 
+.CodeMirror-linenumber {
+  font-size: 11px;
+  line-height: 15px;
+}
+
 /* set the linenumber white when there is a breakpoint */
 .new-breakpoint .CodeMirror-linenumber {
   color: white;
@@ -49,6 +55,11 @@
 /* move the breakpoint below the linenumber */
 .new-breakpoint .CodeMirror-gutter-elt:last-child {
   z-index: 0;
+}
+
+.editor-wrapper .CodeMirror-line {
+  font-size: 12px;
+  line-height: 15px;
 }
 
 .debug-line .CodeMirror-line {
--- a/public/js/components/Editor.css
+++ b/public/js/components/Editor.css
@@ -25,9 +25,9 @@
 .editor.new-breakpoint svg {
   fill: var(--theme-selection-background);
   width: 60px;
-  height: 12px;
+  height: 14px;
   position: absolute;
-  top: 1px;
+  top: 0px;
   right: -4px;
 }
 
@@ -44,7 +44,7 @@
 
 .CodeMirror-linenumber {
   font-size: 11px;
-  line-height: 15px;
+  line-height: 14px;
 }
 
 /* set the linenumber white when there is a breakpoint */
@@ -58,8 +58,8 @@
 }
 
 .editor-wrapper .CodeMirror-line {
-  font-size: 12px;
-  line-height: 15px;
+  font-size: 11px;
+  line-height: 14px;
 }
 
 .debug-line .CodeMirror-line {
--- a/public/js/components/Editor.css
+++ b/public/js/components/Editor.css
@@ -24,11 +24,10 @@
 
 .editor.new-breakpoint svg {
   fill: var(--theme-selection-background);
-  width: 60px;
-  height: 12px;
+  height: 18px;
   position: absolute;
   top: 0;
-  right: -4px;
+  right: -8px;
 }
 
 .editor.new-breakpoint.breakpoint-disabled svg {
@@ -40,10 +39,6 @@
   width: 100%;
   height: 100%;
   background-color: var(--theme-body-background);
-}
-
-.CodeMirror-linenumber {
-  font-size: 11px;
 }
 
 /* set the linenumber white when there is a breakpoint */


wldcordeiro:
Could you attach screenshots for both Chrome and Firefox? Just want to verify we're rendering properly. 😄 Thanks for the PR @monicau

monicau:
Sure! I added screenshots to the original post.  I'm running Ubuntu 14.04, not sure if it'd look different on Mac or Windows.

jasonLaster:
@monicau oh strange i'm seeing something different on chrome + firefox. breakpoints seem a bit tall

screen shot 2016-10-05 at 3 26 51 pm

public/js/components/Editor.css
```diff
@@ -42,10 +41,6 @@
   background-color: var(--theme-body-background);
 }
 
-.CodeMirror-linenumber {
-  font-size: 11px;
-}
```
jasonLaster: ooh, i think we need this for chrome actually. for some reason chrome defaults to 12px and firefox defaults to 11.

wldcordeiro:
screenshot_20161005_132832

Heights do look just a little odd, @jasonLaster Didn't notice when the green highlight was on the same line.

public/js/components/Editor.css
monicau: I see, thanks for the info.  I removed it because size 11px makes the gutter font smaller than the size of text in the main panel to the right of the gutter, hence the height difference.  I'll try adding line-height instead.

monicau: Hmm setting a specific line-height (eg: 18px) doesn't give consistent result in Chrome and Firefox (Chrome is too short by 1px). 
#### Chrome : 
debuggerchromelineheight
#### Firefox : 
debuggerfirefoxlineheight

Going back to the original css code, I get the gutter highlight misalignment in both Firefox and Chrome.  @jasonLaster do you only see this issue in Chrome?  

Setting the font-size to inherit makes the gutter font size consistent with the main panel for me and fixes the highlight height issue, not sure if that'll be the same case for you guys.

jasonLaster: @monicau interesting, mind sharing a the change?

jasonLaster:
Thanks for the help so far @monicau. 

You're definitely finding some helpful edge cases. 

I looked into it a bit and found that it might be helpful to lock down the line height of the codemirror lines. Here's a [gist](https://gist.github.com/jasonLaster/9ba8f6d83ca47eafeafb974b16748d4c) of what I had. It's by no means perfect, so it'll be helpful to compare in different browsers too.

It looks like in firefox it's 13.333 and in chrome it's 15. I have a patch here that sets the line heights to 15, but I think we could just as easily lock it down to 13. In a side-by-side comparison, I personally prefer 15px line height, but there's an argument to be made for not changing it. In terms of line height ratio's 15/12 = 1.25 and 13.333/12 = 1.11. Personally, I'm more in favor of 1.25

For what it's worth, when I was testing I set some background colors (lightblue and lightcoral) to get a sense of how it lined up :)

screen shot 2016-10-05 at 6 20 53 pm

Also, I love that you're using the test example to debug! Thanks for getting involved in the project again, we're a really early project and there's lots of work to go around

monicau:
Thanks for looking into it @jasonLaster !  The highlight lines up with the line number now with your changes, but the text in the codemirror in Chrome is overlapping a bit (left one below) but Firefox looks fine (right one)
debuggerchromelineheightsquishy

Looks like the font size isn't the same between the two browsers (you mentioned before about Chrome and Firefox having different default size).  Can we specify a font size so that it's consistent in all browsers?  Unless there was a reason why there's no specific font size set for the codemirror at the moment?  It seems to work if I add something like `font-size:12px` to the selector `.editor-wrapper .CodeMirror-line` on top of your changes.

And thanks all of you for being patient with me so far.   :)  It's my first time trying to get involved in open source projects.

jasonLaster:
This is looking good.

Definitely add the font size :)

And thanks for getting involved. How did you find this project?

public/js/components/Editor.css
monicau: sure, it was 
``` 
.CodeMirror-linenumber {
font-size: inherit;
}
```
without adding any line-heights anywhere.

monicau:
Okay committed and pushed.  Let me know if it looks fine on your end.

I found this project starting from [whatcanidoformozilla.org](http://whatcanidoformozilla.org) to [wiki.mozilla.org/DevTools/GetInvolved](http://wiki.mozilla.org/DevTools/GetInvolved) to the recent tweets from FirefoxDevTools :stuck_out_tongue:

jlongster:
> Looks like the font size isn't the same between the two browsers (you mentioned before about Chrome and Firefox having different default size). Can we specify a font size so that it's consistent in all browsers? Unless there was a reason why there's no specific font size set for the codemirror at the moment? It seems to work if I add something like font-size:12px to the selector .editor-wrapper .CodeMirror-line on top of your changes.

Something that looks really weird to me in the screenshot you attached with the above text: the font size of the editor looks huge. The font size of the line number and the editor should be exactly the same. Something seems off to me here, and what's most important is that it looks good in the Firefox panel, and so far nobody has looked at that. (It's bit more setup to get the debugger running *actually* in the Firefox devtools, so that's understandable)

I can check it out in the panel right now to make sure it's still the same.

jlongster:
I just noticed that this also changes the font-size to 12px. We can't do that. Please revert it back to 11px, as that is the standard font-size for all of devtools.

I hate to say it, but I prefer the way it was before. You can see more code on the screen with a smaller line-height. A big line-height (like 1.5) is preferable for large amounts of websites with text, but this is code and usually the devtools is a short window on the bottom so it's nicer to see more code. Also, with the bigger line-height, the highlighted line does not perfectly align with the breakpoint anymore:

screen shot 2016-10-06 at 10 32 24 am

The way is looks right now in the Firefox panel is exactly correct: it uses 11px for a font-size and everything lines up right. @monicau @jasonLaster Can you find a change that *only* fixes the problem in Chrome where it exists, but doesn't change anything else?

monicau:
> Something that looks really weird to me in the screenshot you attached with the above text: the font size of the editor looks huge. 

That's because I had the page in zoomed-in mode (I'm blind).  My future screenshots will be regular zoom.

Thanks for your input @jlongster, good to know that the line number and text should be the same size, and that the standard is 11px.  Good point about the line height.  Here's a comparison of 13px and 11px line-height in Chrome:

chromelineheight-13-11

I notice in Firefox that the highlighted text is anchored to the bottom of the line, so 11px looks squishy:
firefoxlineheight-13-11
when not specifying any line height, the default height in Firefox is 14px and Chrome 13px.  Which do you guys prefer?

jasonLaster:
Thanks for the update @monicau. I liked @jlongster's point that we should lock down firefox's current behavior. We can always update it in a design pass :) With that in mind, lets use 14px.

By the way, I'm also blind. I would love to see a debugger theme for blind people :) My editor's font size is 16px and line height is 1.5. A more comfortable on the eyes editor, would be enough of a reason to switch tools if all other things were equal. 

I think this is really close to landing

jlongster:
> That's because I had the page in zoomed-in mode (I'm blind). My future screenshots will be regular zoom.

Oh! No worries! :) That's good to know. Didn't know it was scaled up.

> when not specifying any line height, the default height in Firefox is 14px and Chrome 13px. Which do you guys prefer?

If I inspect the current debugger in Firefox nightly it's showing me a computed line-height of 13.5px... So I don't know! Do you see 14px if you use the browser console to inspect the current debugger in Firefox?

I don't know if we inherit any styles from the toolbox when we are in Firefox that we aren't getting outside of the toolbox. I didn't see any rules that set 13.5px explicitly. It seems like a strange default. If 14px is the default for you, let's use that.

Thanks for talking this through!

monicau:
@jlongster Yup it's 14px when I use Firefox 49.0's element inspector to look at the line height of the debugger.  Not sure how to find the computed height properly, but I do trial and error on line-height values and compare the result with the default line-height by unchecking the line-height property in the console (left side in screenshot)

inheritornot

Looks like 13.5px (right side) is smaller than the default inherited line-height.  Maybe default values are different in different versions of Firefox?  Alright I'll go with 14px then.  The breakpoint arrow doesn't cover the line completely now with the gutter also being 14px in line-height, so I increased the svg height accordingly. 

@jasonLaster not as blind as me, I'm at 18px. But we're not taking monitor size and resolution into consideration so this blind stand-off inconclusive.

jasonLaster:
Hey @jlongster, @wldcordeiro mind reviewing this once more. The patch looks great locally on all browsers. Monica thanks for being so clear this work is awesome1

haha - guess i'll catch up soon on the blindness...

wldcordeiro:
I'm saying 👍 here, there was a minor nit where a line was changed (probably back and forth from other commits) but I don't think it's worth blocking for.

jlongster:
Thanks for all your patience! Just tried it locally and it looks good to me.

jasonLaster:
Thanks @monicau! Welcome to the world of OSS.

Please come back again soon :)

Issue #792, opened 2016-09-21
Code Search - UI Tweaks


jasonLaster:
[Current designs](https://invis.io/AY8OEG3C6#/117937554_Debugger)

This issue covers small visual tweaks to the search input. It might also involve moving the input to the top of the editor.

I took a look at a couple editors and drew up a list of features we may want to implement:

* *label* - left magnifying glass or placeholder "find" text
* *clear* - clear button on the right
* *cancel* - button to exit search
* *results* - list a summary of results *3 of 77*
* *next/prev* - buttons to find next previous
* *regex* - regex mode for search
* *caps* - caps sensitive
* *whole word* - match whole word
* *in selection* - match in selection

----

Many of these features are more editor oriented. I think we should try and add the label and cancel button soon. And in the future it would be nice to have a results summary, regex mode, and next previous buttons.

---

**Firefox**
![screen shot 2016-09-21 at 9 38 48 am](https://cloud.githubusercontent.com/assets/254562/18713042/f3c0b5ae-7fde-11e6-83fb-0f706b3fe7ad.png)

**Atom**
![screen shot 2016-09-21 at 9 36 42 am](https://cloud.githubusercontent.com/assets/254562/18712998/c5b279b8-7fde-11e6-8bab-4e1e9b330ffc.png)

**Chrome**
![screen shot 2016-09-21 at 9 36 21 am](https://cloud.githubusercontent.com/assets/254562/18712999/c5b8834e-7fde-11e6-9d7c-ecf6ae4fc54a.png)

helenvholmes:
[Debugger search designs](https://invis.io/AY8OEG3C6#/117937554_Debugger)

Still need: a good way to handle closing, and after chatting with Jason I think we both want to move the "1 of 7 results" string so it's not moving as people are typing.

jasonLaster:
Here's the latest design for the "in file" search.

Changes:
* moves search to above the editor
* show a summary of the responses (this will likely move)
* show an error message when nothing is found

We'll also likely add a close button before the designs are finalized.

Would someone like to take a stab at updating the UI?

![screen shot 2016-09-21 at 10 37 47 pm](https://cloud.githubusercontent.com/assets/254562/18735420/b8113a84-804b-11e6-8042-16f231180158.png)

wldcordeiro:
✋ I'll grab this one since someone hopped on file search 😄

helenvholmes:
[...]

PR #888, opened 2016-10-06
Style search bar

jasonLaster:
Associated Issue: #792 

### Summary of Changes
* summary is on the right because i was not sure how to position it next to the query. Also, not sure if it would look good as you are typing
* i left the close button out because the summary is on the right and cmd+f and escape close. This feels like enough in practice
* i select the text when the search bar is opened so you have the cursor in the right spot and can delete before you type
* the results summary wraps

### DIFF
--- a/public/js/components/Editor.css
+++ b/public/js/components/Editor.css
@@ -35,6 +35,10 @@
   opacity: 0.3;
 }
 
+.search-bar ~ .editor-mount {
+  height: calc(100% - 58px);
+}
+
 .editor-mount,
 .CodeMirror {
   width: 100%;
--- a/public/js/components/EditorSearchBar.js
+++ b/public/js/components/EditorSearchBar.js
@@ -36,12 +36,15 @@ const EditorSearchBar = React.createClass({
     shortcuts: PropTypes.object
   },
 
-  setupKeyboardShortcuts() {
-    if (this.keyShortcutsEnabled) {
-      return;
+  componentWillUnmount() {
+    const shortcuts = this.context.shortcuts;
+    if (isEnabled("search")) {
+      shortcuts.off("Cmd+f", this.toggleSearch);
+      shortcuts.off("Escape", this.onEscape);
     }
+  },
 
-    this.keyShortcutsEnabled = true;
+  componentDidMount() {
     const shortcuts = this.context.shortcuts;
     if (isEnabled("search")) {
       shortcuts.on("Cmd+f", this.toggleSearch);
@@ -49,11 +52,9 @@ const EditorSearchBar = React.createClass({
     }
   },
 
-  componentWillUnmount() {
-    const shortcuts = this.context.shortcuts;
-    if (isEnabled("search")) {
-      shortcuts.off("Cmd+f", this.toggleSearch);
-      shortcuts.off("Escape", this.onEscape);
+  componentDidUpdate() {
+    if (this.searchInput()) {
+      this.searchInput().focus();
     }
   },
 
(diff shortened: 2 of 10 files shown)


public/js/components/Editor.css
```diff
@@ -27,23 +27,28 @@
...
 
+.search-bar ~ .editor-mount {
+  height: calc(100% - 58px)
+}
+
 .editor-mount,
 .CodeMirror {
   width: 100%;
-  height: 100%;
+  height: calc(100% - 17px);
```
jasonLaster: unfortunately, the editor was being masked by the footer. I'd like to investigate avoiding absolute positioning here. I was able to avoid it with the search bar so it *should* be possible.

nt1m:
It would be nice if this awesome search box could make it into the built-in Editor component.

public/js/components/Editor.css
jlongster: I feel like we have too many little hacks around the editor area (1px offset here and there). I would *love* it if you could take a few minutes to think about about the editor area and possibly refactor it a little bit to avoid this. Maybe a few more uses of flexbox would help. I feel like it's getting way too brittle.

I'm fine including that in this PR, it's related to the UI changes you want to make. It's hard for me to approve yet another hardcoded value :)

jlongster:
That looks really nice! I like how it looks.

public/js/components/Editor.css
jlongster: Yeah, that makes sense. Would probably be good to pull this out into a separate PR with a real fix.

I haven't looked at this CSS for a while, but seems like we could use flexbox and tell the editor-container to fill up as much as possible, while the header and footer take up as little as possible (`flex: 0`). But not sure.

jlongster:
> It would be nice if this awesome search box could make it into the built-in Editor component.

Brian and Gabe are working on pulling out the editor into a separate module so we can all collaborate in one place and be using exactly the same thing. We might be able to pull this in there when we do it. Right now it makes sense for us to do this, because we also want to add debugger-specific options. But I agree having different experiences is not ideal... when we have separated out the editor from m-c, we can think about it.

public/js/components/EditorSearchBar.js
```diff
@@ -0,0 +1,159 @@
...
+  contextTypes: {
+    shortcuts: PropTypes.object
+  },
+
+  setupKeyboardShortcuts() {
+    if (this.keyShortcutsEnabled) {
+      return;
+    }
+
+    this.keyShortcutsEnabled = true;
```
jlongster: Why keep track of this? Wouldn't just calling this in `componentDidMount` work instead of `componentDidUpdate`?

public/js/components/EditorSearchBar.js
```diff
@@ -0,0 +1,159 @@
...
+  render() {
+    if (!isEnabled("search") || !this.state.enabled) {
+      return dom.div();
+    }
+
+    const { count } = this.state;
+
+    return dom.div(
+      { className: "search-bar" },
+      new Svg("magnifying-glass"),
```
jlongster: You don't need `new` here

public/js/components/EditorSearchBar.js
jlongster: "needed the window"? Shouldn't `window` always exist? I encourage you to go down these rabbit holes :) Otherwise we are just accumulating technical debt.

jlongster:
Awesome, thanks!

public/js/components/EditorSearchBar.js
jasonLaster: Yep - all good points. I can absolutely see where you're coming from.

PR #889, opened 2016-10-06
Add mocha dot reporter

jasonLaster:
This switches the default cli mocha reporter to *dot*. The main advantage is that running `git push` will no longer result in a wall of output.

screen shot 2016-10-06 at 9 24 27 am

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
     "firefox": "node bin/firefox-driver --start",
     "mochitests-watch": "MOCHITESTS=true TARGET=firefox-panel webpack --watch",
     "build-docs": "documentation build --format html --shallow  --document-exported --output docs/reference/ public/js/actions/ public/js/test/mochitest/head.js",
-    "prepush": "npm run lint; npm run test",
+    "prepush": "npm run lint; node public/js/test/node-unit-tests.js --dots",
     "build-storybook": "build-storybook"
   },
   "engineStrict": true,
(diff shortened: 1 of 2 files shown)


jlongster:
I'm confused, the only different is it doesn't list the names of the tests? This seems like a step back in terms of UX. If we really don't want the wall of text, maybe we shouldn't force tests to be run?

jasonLaster:
Hmm, it does list the tests that fail, so it's just more compact. I suppose we could make it configurable so running `npm test` shows the list but prepush when you just want an okay could be compact.

wldcordeiro:
I vote for Jest! 😉

jlongster:
@wldcordeiro Jest would be great :) Unfortunately we need to coordinate with the rest of the team, like the console, which is already using mocha too. There are already many moving parts. Might be hard to suggest a switch to a different testing framework when most people are familiar with mocha.

jlongster:
@jasonLaster I'm a little confused because I thought the dot reporter just showed a line of dots like "....." but your screenshot shows otherwise. In your screenshot, the text isn't any shorter, just some text isn't shown so I'm not sure I see the value.

jasonLaster:
@jlongster oh yeah, i copied the existing screen to show the wall of text. dot is just dots, with formatted error messages.

jasonLaster:
Updated 

screen shot 2016-10-06 at 12 51 16 pm

jlongster:
Oh, weird, why don't I see any test titles in your original screenshot? I also don't see any dots in the last screenshot... haha.

I see now, originally I thought this PR just hid the test titles. I'd be fine switching to the dot reported though as long as it still shows everything about a test when it fails.

jasonLaster:
my font sucks...

wldcordeiro:
@jlongster I was mostly joking there. Mocha is a fine testing tool. 😄

PR #892, opened 2016-10-06
Sort file tree by index -> folders -> file. Fixes #743

### LINKED ISSUE #743: Sort folders first in tree nodes
Currently the folders and files are just sorted alphabetically in the source list, can folders be sorted first?

gregtatum:
Associated Issue: #743

### Summary of Changes

* Expanded the file sort ability.
* Added testing coverage for sorts.
* Slightly reworded variables in the `addToTree()`

### Testing

* [x] passes `npm test`
* [x] passes `npm run lint`

### DIFF
--- a/public/js/utils/sources-tree.js
+++ b/public/js/utils/sources-tree.js
@@ -118,7 +118,10 @@ function addToTree(tree: any, source: TmpSource) {
   let path = "";
   let subtree = tree;
 
-  for (let part of parts) {
+  for (let i = 0; i < parts.length; i++) {
+    const part = parts[i];
+    const isLastPart = i === parts.length - 1;
+
     // Currently we assume that we are descending into a node with
     // children. This will fail if a path has a directory named the
     // same as another file, like `foo/bar.js/file.js`.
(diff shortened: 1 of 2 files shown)


public/js/utils/sources-tree.js
```diff
@@ -161,6 +160,26 @@ function addToTree(tree: any, source: TmpSource) {
...
+
+  return nodes.findIndex(node => {
+    if (node.name === "(index)") {
+      return false;
+    }
+    const nodeIsDir = nodeHasChildren(node);
+    if (partIsDir) {
+      return nodeIsDir ? node.name.localeCompare(pathPart) >= 0 : true;
+    }
+    return nodeIsDir ? false : node.name.localeCompare(pathPart) >= 0;
```
jasonLaster: I think this can be re-worked to be clearer. I'm 50% confident this will work, so it's a good thing you have tests.

What I like about this though is that you're saying if both the part and node are dir or not dir, then do a comparison, otherwise and them. ... i don't know, definitely needs some comments here :)

```js
if (partIsDir && nodeIsDir) || !(partIsDir && nodeIsDir) {
  return node.name.localeCompare(pathPart);
}  else {
   return partIsDir && nodeIsDir;
}
```

jlongster:
Looks good to me, thanks! 👍

Issue #841, opened 2016-09-28
Shouldn't be able to select text in the tool 'chrome'


bgrins:
If I do ctrl+a, or select the 'no breakpoints' / 'sources' / 'Want to find a file? (Cmd + P)' text in the UI it shouldn't select anything to match the rest of the browser chrome

jlongster:
There are places we disable user selection, but I didn't think we wanted to disable it globally. I suppose I'm fine with it, hadn't really thought it through. Thanks!

clarkbw:
This looks like a pretty easily scoped project if anyone else wants to take it on.

I would take a look at where we are using strings and find a way to disable it in specific component classes, globally could be the answer but this feels like it should probably be on a component level so as not to mess with other things where selection should work.

Here's the CSS you'll need to make this happen.

https://developer.mozilla.org/en-US/docs/Web/CSS/user-select
```
.unselectable {
  -moz-user-select: none;
  -webkit-user-select: none;
  -ms-user-select: none;
}
```

wldcordeiro:
Could we enumerate some of the areas where we don't want selection? That would make this an easier task for a new contributor.

wldcordeiro:
* [ ] command bar
* [ ] source tabs
* [ ] sources header
* [ ] sources tree?
* [ ] right sidebar?

@jlongster @clarkbw feel free to add to the list.

jasonLaster:
@clarkbw do you think we could just use the `user-select` property? https://developer.mozilla.org/en-US/docs/Web/CSS/user-select

clarkbw:
@jasonLaster yes, I believe that's all we need to do.  I think the list @wldcordeiro has looks good.  Mostly labels and other text needs this and a global setting could prevent people from copying certain object attributes / values we might want to allow.

wldcordeiro:
I'm getting this done right now. PR should be open shortly.

PR #893, opened 2016-10-06
Add `user-select: none` to several parts of the debugger.

wldcordeiro:
Associated Issue: #841 

### Summary of Changes

* add `user-select: none` to various parts of the debugger so selection/ctrl+a only gets the things we expect (note, bare `user-select` seems to not work so adding the vendor prefixes for full support.)

### Testing

* [x] passes `npm test`
* [x] passes `npm run lint`

### DIFF
--- a/public/js/components/Breakpoints.css
+++ b/public/js/components/Breakpoints.css
@@ -1,8 +1,5 @@
 .breakpoints-list * {
-  -webkit-user-select: none;
   -moz-user-select: none;
-  -ms-user-select: none;
-  -o-user-select: none;
   user-select: none;
 }
 
--- a/public/js/components/Breakpoints.css
+++ b/public/js/components/Breakpoints.css
@@ -1,3 +1,10 @@
+.breakpoints-list * {
+  -webkit-user-select: none;
+  -moz-user-select: none;
+  -ms-user-select: none;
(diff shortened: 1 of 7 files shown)


public/js/components/Editor.css
```diff
@@ -90,4 +90,9 @@
   background-color: var(--theme-tab-toolbar-background);
   font-weight: lighter;
   z-index: 100;
+  -webkit-user-select: none;
+  -moz-user-select: none;
+  -ms-user-select: none;
+  -o-user-select: none;
+  user-select: none;
```
jlongster: Does this apply it to the text within CodeMirror? The code should be selectable.

wldcordeiro: screenshot_20161006_133621

public/js/components/Breakpoints.css
```diff
@@ -1,3 +1,10 @@
+.breakpoints-list * {
+  -webkit-user-select: none;
+  -moz-user-select: none;
+  -ms-user-select: none;
+  -o-user-select: none;
+  user-select: none;
```
jasonLaster: we should only need `user-select`. prefixes are the worst

wldcordeiro:
screenshot_20161006_135958

PR #900, opened 2016-10-07
Make sure to clear the debug line in the correct CodeMirror document

jlongster:
Since we implemented CodeMirror documents, we may not correctly clear the debug line, so you may end up seeing multiple green debug lines. The problem is that it cleared if after setting the new source text, so it wasn't properly clearing it on the previous "document", so loading that document again (when switching) back will still show the previous line.

The fix is pretty simple: make sure to clear it before we update the source text.

### DIFF
--- a/public/js/components/Editor.js
+++ b/public/js/components/Editor.js
@@ -77,13 +77,17 @@ const Editor = React.createClass({
     }
   },
 
-  updateDebugLine(prevProps, nextProps) {
-    if (prevProps.selectedFrame) {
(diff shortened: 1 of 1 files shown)


wldcordeiro:
I pulled the changes down and it appears the lines are still highlighting in the wrong document?

screenshot_20161008_122952
screenshot_20161008_122958

jlongster:
Thanks for testing. I'll check it out.

wldcordeiro:
Working now. Merging. 😄

Issue #867, opened 2016-10-04
Polish RTL mode


jasonLaster:
We recently added basic RTL support with an html *dir* tag, but the UI is still pretty busted in this state.

Here's what it looks like:

![](https://cloud.githubusercontent.com/assets/1141550/19070242/98ed3b2e-8a2a-11e6-9236-8b8430cbb2c2.png)

It would be great if someone wanted to help us clean up the UI in this mode. Here are some actionable fixes:

- code is displayed in RTL (should keep LTR)
- source tree is not properly indented
- arrows are in the wrong direction
- icons are overlapping

clarkbw:
RTL mode can be accessed for Firefox via CSS, see the [old debugger.css](https://dxr.mozilla.org/mozilla-central/source/devtools/client/themes/debugger.css#203)

It uses the following:
```css
.selector:-moz-locale-dir(rtl) {
  /* for rtl changes */
}
.selector:-moz-locale-dir(ltr) {
  /* this is the default, only use for selectors that need to change */
}
```

nt1m:
@clarkbw, actually, since the new debugger uses HTML, you have to use :dir instead of :-moz-locale-dir

clarkbw:
> you have to use :dir instead of :-moz-locale-dir

oh right!  de-XUL my brain!

jasonLaster:
Hey @amitzur would you be interested in taking a stab at this?

amitzur:
I would gladly, however how important is this? Don't most if not all devs debugging JS use the english version? Or at least RTL devs?
This is mainly targeted for the debugger being embedded in devtools, right?
In addition, I took a screenshot of hebrew firefox's devtools. The strings are also in hebrew. Do you think we should work on that as well?
<img width="1280" alt="screen shot 2016-10-05 at 20 41 38" src="https://cloud.githubusercontent.com/assets/394320/19124534/3bb46cda-8b3c-11e6-89ef-adf56a41b4e5.png">

jasonLaster:
Hi @amitzur, thanks for the hebrew screen :) 

We've gone back and forth on how important internationalization is because many devs work in english. The conclusion we've landed on is that it *is* nice to support basic rtl and translated strings. We will be looking into adding translations soon :)

Would love your help  on adding basic RTL.

amitzur:
I got it then :)

PR #903, opened 2016-10-08
Polish RTL

amitzur:
Associated Issue: #867 

### Summary of Changes

1. I chose to use html[dir="rtl"] as an indicator for RTL. The :dir pseudo-class is only supported in Firefox, so I wasn't sure if it's a good choice. I have a feeling you're going to tell me to go with :dir :-)

2. I explicitly separated directional properties to html:not([dir="rtl"]) for (a) declarative style (b) avoid property reset under the RTL rule. I felt it was too verbose, but better than the alternative.
e.g. 
```
.some-class {
  width: 10px;
  margin-left: 8px;
}
```
becomes:
```
.some-class {
  width: 10px;
}

html:not([dir="rtl"]) .some-class {
  margin-left: 8px;
}

html[dir="rtl"] .some-class {
  margin-right: 8px;
}
```

3. I want to raise my concern about the ability to maintain this. Maybe unit test coverage can help, but I would suggest some automatic generation of RTL css through annotations or something. Or, I would question the RTL requirement altogether.

4. ObjectInspector.css : `.arrow.expanded` rule needed attention due to specificity issue. I didn't like the solution (see TODO).

5. `.commad-bar` rules exist in both RightSidebar.css and SourceFooter.css

6. Sources.css had some seemingly unused styles around UL and .source-item so I removed them.

7. The icons with class name file/domain/folder/worker all have the same rules. I suggest adding a class ("icon?") to consolidate. I can do this, just didn't want it to be part of this pull request and also wanted to know your opinion.

8. `.source-footer` is visually over the splitter on the right side. Not sure if this is intentional, but changing to right:2px fixes it.

9. "⌘+P to search" : currently doesn't render well. When introducing string translations, I suggest placing the "⌘+P" in a separate span with direction:ltr

10. Still need work: breakpoint, call stack, scopes.

11. Sources tree indentation: need to change paddingLeft to paddingRight in SourcesTree.js:113. I didn't know what would be your preferred way to detect RTL in js, but I didn't want to do what SplitBox does: `win.getComputedStyle(doc.documentElement).direction`.

### Testing

* [x] passes `npm test`
* [x] passes `npm run lint`

### DIFF
--- a/public/js/components/Editor.css
+++ b/public/js/components/Editor.css
@@ -16,6 +16,10 @@
   left: 0px;
 }
 
+html[dir="rtl"] .editor-mount {
+  direction: ltr;
+}
+
 .editor-wrapper .breakpoints {
   position: absolute;
   top: 0;
--- a/public/js/components/ObjectInspector.css
+++ b/public/js/components/ObjectInspector.css
@@ -1,13 +1,22 @@
 .arrow svg {
   fill: var(--theme-splitter-color);
-  margin-right: 5px;
   margin-top: 3px;
-  transform: rotate(-90deg);
   transition: transform 0.25s ease;
   width: 10px;
 }
 
-.arrow.expanded svg {
+html:not([dir="rtl"]) .arrow svg {
+  margin-right: 5px;
+  transform: rotate(-90deg);
+}
+
+html[dir="rtl"] .arrow svg {
+  margin-left: 5px;
+  transform: rotate(90deg);
+}
+
+/* TODO (Amit): html is just for specificity. keep it like this? */
+html .arrow.expanded svg {
   transform: rotate(0deg);
 }
 
--- a/public/js/components/RightSidebar.css
+++ b/public/js/components/RightSidebar.css
@@ -21,12 +21,18 @@
 
 .command-bar {
   height: 30px;
+}
+
+html:not([dir="rtl"]) .command-bar {
   padding: 8px 5px 10px 1px;
+}
+
+html[dir="rtl"] .command-bar {
+  padding: 8px 1px 10px 5px;
 }
 
 .command-bar > span {
   cursor: pointer;
-  margin-right: 0.7em;
   width: 16px;
   height: 17px;
   display: inline-block;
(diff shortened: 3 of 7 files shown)


jasonLaster:
Thanks Amit! This is a lot of work. I'll review it closely on Monday, but for what's it's worth I agree with your points. 

I can work on adding some storybook stories for RTL next week as well if we decide we want to commit to RTL support. You're right that it'll be important to prevent regressions.

amitzur:
Thanks Jason,

I tried to run the storybook but I get an error about module "Services":
```
ERROR in ./public/js/utils/text.js
Module not found: Error: Cannot resolve module 'Services' in /Users/amit/projects/debugger.html/public/js/utils
 @ ./public/js/utils/text.js 1:15-34
```
I couldn't figure out how webpack resolves `require("Services")` since it's not a relative url. I assume it is targeting devtools-sham/sham/services.js but can't understand how it gets to it

jasonLaster:
Hey @amitzur, no worries. Honestly, storybook is in a stale state as we want to change how we use it. I wouldn't worry too much about it now.

We will have a chance soon to revamp it and add some unit tests, which will keep it from going stale in the future.

jasonLaster:
> 1. html[dir="rtl"] 

Agreed, we would like to continue supporting chromium browsers. At somepoint we can switch to `:dir` when it is better supported.

2. separated directional properties

Agreed, it is nice to isolate directional changes so it's easy to see the diff.

3. maintainability

* I think storybook could help, it has tests it also would be easy to see several scenarios
* I think annotation is also an interesting idea. For other reasons, I'd like to consider a toolchain like post-css, which allows you to extend the syntax and perhaps add an rtl annotation.

4. CSS cleanup

Thanks for highlighting the tweaks with command bar, search, ... I'm not going to touch on that now as we probably want to have a discussion about the big picture first.

jasonLaster:
@amitzur - would you mind posting a screenshot of the progress. It'd be pretty cool to see a before / after.

amitzur:
screen shot 2016-10-10 at 07 18 45

jasonLaster:
After thinking about this a bit more and speaking with @janodvarko, I think it's important to move forward with RTL support.

* other tools support RTL, and it's important to be consistent
* the changes are mostly CSS and we can add tooling to automate much of this. [margin-inline-start/end](https://developer.mozilla.org/en-US/docs/Web/CSS/margin-inline-end) is a feature for supporting just this. And project [houdini](https://www.smashingmagazine.com/2016/03/houdini-maybe-the-most-exciting-development-in-css-youve-never-heard-of/) or [postcss-bidirectional](https://github.com/gasolin/postcss-bidirection) written by our @gasolin, could let us use it and get all of the benefits. Looking at this PR, we could introduce *bidirectional* and replace the conditionals here with `margin-inline-start` / end. 
* we should introduce some component testing with storybook for keeping us from regressing on themes and rtl.

jasonLaster:
Hey @amitzur, how would you like to tackle these subsequent issues? My vote would be to clean up the merge conflicts and merge this PR and then open a subsequent PR with some of the other tweaks.

With the JS direction issue, i propose adding a DOM utils module for getting the direction, that is probably the best we can do. and perhaps instead of setting paddingLeft, paddingRight... set paddingStart and paddingEnd, which can resolve to the correct direction.

Issue #493, opened 2016-08-05
Toolbox - Create separate library


jasonLaster:
We've discussed extracting the toolbox from debugger.html from the beginning. There are a couple benefits:

1. focus debugger.html
2. let other tools (e.g. console) connect to javascript targets (chrome, firefox, node)

**toolbox:**
+ **client**
   + list tabs
   + connect to tab
   + commands and events
   + tools would define their own init handler
   + chrome-devtools-client library, which gives access 
+ **config**
   + helps determine target (development, panel)
   + configuration for different targets (workerBaseUrl)
   + feature flags for different targets
+ **tabs UI**
   + ui to show available tabs and select a tab to connect to. This should be easy for other tools to include in their own apps.
+ **development server**
   + webpack server with hot module reloading, and config with plugins, aliasses, etc
   + storybook config so that tools can define their own stories
   + mocha node unit test setup so that other tools can write their own unit tests
   + cypress config and server so that other tools can write their own tests and fixtures (this is still experimental)
+ **example app**
  + a small repl console that demonstrates how a new tool can use these tools to connect to a tab and build a small app.
  + the example app can likely be in its own repo and demonstrate the integration points

**Integration points:**
+ the toolbox will expose scripts like development-server to tools that include the package
+ the toolbox will expose client functions publicly so that tools can import them

jasonLaster:
Finished in #905

PR #905, opened 2016-10-09
(WIP) Extract DevTools Toolbox

jasonLaster:
Associated Issue: #493

### Summary of Changes

This is still a work in progress, but several of the big steps are done.

- [x] Run Toolbox Tab chooser
- [x] Run Debugger on top of Toolbox
- [x] Run `npm start` from top level directory
- [x] Run unit tests 
- [x] Run mochitests
- [ ] Run Toolbox in a new project
- [ ] Cleanup & Document Toolbox 

Next Steps:

There's no way this PR can do everything, so there are some things that I'm not going to try and do.

- [ ] Extract the Debugger Client Protocol 
- [ ] Extract other DevTools libs like Reps
- [ ] De-dupe some internal utils like Redux Middleware

Thanks, @wldcordeiro for helping me fix some especially hairy webpack issues!

Screenshot of the new directory structure
screen shot 2016-10-09 at 2 57 59 pm


packages/devtools-local-toolbox/public/js/index.js
```diff
@@ -0,0 +1,111 @@
...
+const React = require("react");
+
+const DevToolsUtils = require("devtools-sham/shared/DevToolsUtils");
+const AppConstants = require("devtools-sham/sham/appconstants").AppConstants;
+const { injectGlobals, debugGlobal } = require("./utils/debug");
+const { isEnabled, getValue,
+        isDevelopment, setConfig } = require("../../config/feature");
+
+// Exported functions
+const feature = require("../../config/feature");
```
jasonLaster: oh ha, i don't think so... :)

jasonLaster:
Progress update: spent the day working on running `npm start` from the top level directory. This was an especially difficult bug to track down as webpack module aliasses were being affected by babelrc aliases... This now works, but i need to fight babel aliases for the unit tests...

bin/make-firefox-bundle
```diff
@@ -28,11 +28,11 @@ fi
...
 
-TARGET=firefox-panel node_modules/.bin/webpack
+TARGET=firefox-panel ./node_modules/.bin/build-bundle
 echo "// Generated from: $REV\n" | cat - public/build/bundle.js > "$DEBUGGER_PATH/bundle.js"
-cp public/build/source-map-worker.js "$DEBUGGER_PATH"
-cp public/build/pretty-print-worker.js "$DEBUGGER_PATH"
-cp public/build/styles.css "$DEBUGGER_PATH"
+cp /public/build/source-map-worker.js "$DEBUGGER_PATH"
+cp /public/build/pretty-print-worker.js "$DEBUGGER_PATH"
+cp /public/build/styles.css "$DEBUGGER_PATH"
```
jlongster: I don't understand these changes. Sure we're not creating a top-level `public` folder on the system?

I don't know how I feel above hiding webpack inside of `build-bundle`. The toolbox should not own our webpack build process. We should just require the toolbox in our `main` file, and our webpack will bundle it in, and the project can document whatever babel plugins/etc it needs enabled.

packages/devtools-local-toolbox/config/local.json
```diff
@@ -0,0 +1,20 @@
...
+  },
+  "redux-devtools": {
+    "enabled": false
+  },
+  "firefox": {
+    "geckoDir": "/Users/jlaster/src/mozilla/gecko"
+  },
+  "features": {
+    "prettyPrint": true,
+    "watchExpressions": true
```
jlongster: A lot of these config values are specific to the debugger. Well, I suppose this highlights the difference between features flags and config values. The former are debugger-specific, while the latter aren't. I think we should split them up somehow.

packages/devtools-local-toolbox/public/js/index.js
```diff
@@ -0,0 +1,102 @@
...
+
+  if (isDevelopment()) {
+    AppConstants.DEBUG_JS_MODULES = true;
+    injectGlobals({ store });
+  }
+
+  return { store, actions, LandingPage };
+}
+
+function renderRoot(_React, _ReactDOM, component, _store) {
```
jlongster: I don't think we should be depending on the toolbox to do all of this. I'd like to see it more as a library than a framework. Just give me the pieces to connect, and I'll hook them up to my app. It's not uncommon that we want to tweak things in here and we can't do that if this is shared across everyone.

I think it's very enlightening to just open up an app's `main.js` and see a very straight-forward way of how things are hooked up. If we bury this, it gets harder for beginners to understand what's going on (and even advanced users who want to hack up something quickly to change it).

For example, I frequently go into `main.js` and flip `log: true` in the store configuration to log actions. Why don't I change the config? It's just easier for me to open `main.js` (config is at the root) and more importantly I don't have to restart the server.

I think this could export reducers/actions/components that someone can embed in their app. Or at least make the toolbox have its own render and redux store, and do everything it needs to if we don't have a target. But in our `main.js`, if we have a target, call into the toolbox to connect it, but let us do our own rendering/store creation.

Do whatever you feel like for now, this is a big change so we can move things back. I'm not going to block it.

public/js/main.js
```diff
@@ -1,31 +1,20 @@
...
+const { bootstrap } = require("devtools-local-toolbox");
 const configureStore = require("./utils/create-store");
+
 const reducers = require("./reducers");
 const selectors = require("./selectors");
 
-const LandingPage = require("./components/LandingPage");
 const App = require("./components/App");
 
 const createStore = configureStore({
```
jlongster: I see that we still are creating our own store, so that's good. I thought that this code had been copied over too.

public/js/main.js
```diff
@@ -50,59 +36,16 @@ window.actions = {
...
-
-const connTarget = getTargetFromQuery();
-if (connTarget) {
-  startDebugging(connTarget, actions).then((tabs) => {
-    actions.newTabs(tabs);
-    actions.selectTab({ id: connTarget.param });
-    renderRoot(App);
-  });
+if (isDevelopment()) {
+  bootstrap(React, ReactDOM, App, actions, store);
```
jlongster: This is where I'd probably be a little frustrated if I was trying to learn this project; compare the above code to this and it's a lot less clear what's going on. There is a place for abstractions though, so I could be wrong. I just feel like having code that reads like "if they need to select a target, render the toolbox, otherwise we have a target so connect it and render our app" would be nice.

webpack.config.js
```diff
@@ -1,120 +1,20 @@
...
-// NOTE: This is only needed to fix a bug with chrome devtools' debugger and
-// destructuring params https://github.com/devtools-html/debugger.html/issues/67
-if (getValue("transformParameters")) {
-  webpackConfig.module.loaders.forEach(spec => {
-    if (spec.isJavaScriptLoader) {
-      const idx = spec.loaders.findIndex(loader => loader.includes("babel"));
-      spec.loaders[idx] += "&plugins[]=transform-es2015-parameters";
-    }
-  });
-}
```
jlongster: Similar argument here; trying to figure out how various things like hot reloading or how we're transforming JavaScript would be more difficult to figure out. I think the toolbox should be more isolated and we should still have this webpack code. It would be nice to share though, too, so maybe you're right to move it out.

jlongster:
Just made a few comments to express my ideas about the future direction. It's hard to find a balance between clarity and abstraction; we do want to share some of the annoying work like webpack config so we should probably do that to some extent.

We should absolutely merge this PR though, I don't mean to block it. It's a big change and we can move thinks back and forth. It's a great start though, thanks for working on it!

packages/devtools-local-toolbox/config/local.json
jasonLaster: yup - on it.

webpack.config.js
gregtatum: 👍 for sane webpack defaults that we don't have to re-think on each tool.

jlongster: Yeah, you all are right, we should definitely share it. I'd just like to make sure we do it in a way that's easy to extend, and it's clear where the base config is coming from. I think we're on the same page.

jasonLaster:
Progress update, I'm experimenting with moving the debugger to the packages directory and further splitting out some of the toolbox packages. 

The motivation is to be in-line with other mono-repo projects that do not have a top level directory. I have been using lerna for the past couple hours and have really appreciated how seemless it has been. The `lerna bootstrap` command has really helped when it comes to symlinking package directories and installing bin scripts. This is still an experiment though, so it's subject to change.

packages/devtools-config/configs/local.json
```diff
@@ -0,0 +1,20 @@
...
+    "enabled": false
+  },
+  "firefox": {
+    "geckoDir": "/Users/jlaster/src/mozilla/gecko"
+  },
+  "features": {
+    "prettyPrint": true,
+    "watchExpressions": true
+  }
+}
```
jasonLaster: this will move back to the top level

packages/devtools-config/index.js
```diff
@@ -0,0 +1,3 @@
+const feature = require("./src/feature")
+
+module.exports = feature
```
jasonLaster: this can be replaced with a new entrypoint

packages/devtools-local-toolbox/public/js/index.js
jasonLaster: Yup. I dislike this API too. I'd be happy to brainstorm better APIs that line up w/ the inspector's goals and ours

packages/devtools-local-toolbox/public/js/types.js
```diff
@@ -0,0 +1,58 @@
...
+
+module.exports = {
+  Tab,
+  Source,
+  SourceText,
+  Location,
+  Breakpoint,
+  BreakpointResult,
+  Frame
+};
```
jasonLaster: this should be cleaned up

packages/devtools-local-toolbox/public/js/utils/create-store.js
```diff
@@ -0,0 +1,65 @@
...
+
+  // Hook in the redux devtools browser extension if it exists
+  const devtoolsExt = typeof window === "object" && window.devToolsExtension ?
+    window.devToolsExtension() :
+    f => f;
+
+  return applyMiddleware(...middleware)(devtoolsExt(createStore));
+};
+
+module.exports = configureStore;
```
jasonLaster: this should be shared soon

packages/devtools-local-toolbox/public/js/utils/fromJS.js
```diff
@@ -0,0 +1,39 @@
...
+    return value;
+  }
+
+  // Otherwise, treat it like an object. We can't reliably detect if
+  // it's a plain object because we might be objects from other JS
+  // contexts so `Object !== Object`.
+  return Immutable.Seq(value).map(fromJS).toMap();
+}
+
+module.exports = fromJS;
```
jasonLaster: should be able to kill this

packages/devtools-local-toolbox/public/js/utils/redux/middleware/thunk.js
```diff
@@ -0,0 +1,20 @@
...
+    const args = { dispatch, getState };
+
+    return next => action => {
+      return (typeof action === "function")
+        ? action(makeArgs ? makeArgs(args, getState()) : args)
+        : next(action);
+    };
+  };
+}
+exports.thunk = thunk;
```
jasonLaster: should be able to share soon

packages/devtools-local-toolbox/public/js/utils/utils.js
```diff
@@ -0,0 +1,193 @@
...
+  endTruncateStr,
+  workerTask,
+  zip,
+  entries,
+  toObject,
+  mapObject,
+  compose,
+  updateObj,
+  throttle
+};
```
jasonLaster: I think this should be able to be killed ...

packages/devtools-modules/client/shared/components/splitter/Draggable.js
```diff
@@ -5,7 +5,7 @@
 "use strict";
 
 const React = require("devtools/client/shared/vendor/react");
-const ReactDOM = require("devtools/client/shared/vendor/react");
+const ReactDOM = require("devtools/client/shared/vendor/react-dom");
```
jasonLaster: :/ not sure how this worked before

packages/devtools-modules/index.js
```diff
@@ -0,0 +1,15 @@
...
+const rep = require("./client/shared/components/reps/rep");
+// const repCSS = require("./client/shared/components/reps/reps.css");
+const Grip = require("./client/shared/components/reps/grip");
+
+module.exports = {
+  Services,
+  SplitBox,
+  // SplitBoxCSS,
+  rep,
+  // repCSS,
```
jasonLaster: I copied this into the debugger.html components dir because the requires didn't work... we can debug this now or later...

public/js/components/App.js
```diff
@@ -5,27 +5,22 @@ const { bindActionCreators } = require("redux");
...
-
-// Using this static variable allows webpack to know at compile-time
-// to avoid this require and not include it at all in the output.
-if (process.env.TARGET !== "firefox-panel") {
-  require("../lib/themes/light-theme.css");
-}
+require("./SplitBox.css");
+require("./reps.css");
+let { SplitBox } = require("devtools-modules");
+SplitBox = createFactory(SplitBox);
```
jasonLaster: this can be cleaned up

public/js/components/App.js
```diff
@@ -5,27 +5,22 @@ const { bindActionCreators } = require("redux");
...
 const shortcuts = new KeyShortcuts({ window });
 
 require("./App.css");
-require("devtools/client/shared/components/splitter/SplitBox.css");
-
-// Using this static variable allows webpack to know at compile-time
-// to avoid this require and not include it at all in the output.
-if (process.env.TARGET !== "firefox-panel") {
-  require("../lib/themes/light-theme.css");
-}
```
jasonLaster: moved to the Root.js component, but that can be figured out later with the second phase where come up with the right API

packages/devtools-local-toolbox/public/js/utils/fromJS.js
jasonLaster: never mind, the tabs reducer uses this...

jasonLaster:
The most recent push includes a rebase from master, so should be uptodate and ready to merge sans comments / nits

jlongster:
Good work!

gregtatum:
Lol, that's a huge patch. Nice work @jasonLaster

Issue #871, opened 2016-10-04
improve breakpoint highlight


clarkbw:
#698 updated the breakpoint list but the highlight color needs to be improved.  Right now it looks like a unfocused selection kind of highlight instead of an active indicator.  Need help from @helenvholmes here.

![screen shot 2016-10-04 at 7 55 38 am](https://cloud.githubusercontent.com/assets/2134/19079465/91229bb8-8a08-11e6-9641-9db5f17bf28c.png)

jasonLaster:
Re-opening this issue to discuss a couple tweaks to @Zacqary's improvement.

1. Invert the text


![](https://d17oy1vhnax1f7.cloudfront.net/items/093Y280x3x2U1g0K2X1v/Screen%20Shot%202016-10-10%20at%209.22.32%20AM.hicB.png)

2. De-emphasize the paused-on-line indicator

This is a quick sketch where there is a slight indicator that highlights the breakpoint in a less dramatic fashion.
![](https://d17oy1vhnax1f7.cloudfront.net/items/1o002e2i3l3V2O1M1c1i/Screen%20Shot%202016-10-10%20at%209.27.09%20AM.hi5a.png)


@Zacqary, @helenvholmes, what do you think?

Zacqary:
I think I like the less dramatic callout. The call stack probably needs the loudness of a full background color more than the breakpoint list does.

Can probably implement that one with a `::before` element; I doubt just using a `border` or `outline` would get it to line up the way it does in the sketch.

wldcordeiro:
`:before` pseudo element would be preferred in my opinion. Probably easier to do line-up.

wldcordeiro:
Closing once again. Merged with @jasonLaster's great design suggestion. 😄 Thanks @Zacqary for the PRs and working through the process.

PR #907, opened 2016-10-09
Align breakpoint-list highlight color with codemirror

Zacqary:
Associated Issue: #871 

### Summary of Changes

* `.breakpoints-list .breakpoint.paused` will use `--breakpoint-active-color` to highlight
* Adds `--breakpoint-active-color-hover` variable, and uses this when the user hovers over a `.breakpoints-list .breakpoint.paused`
* Also switches `.breakpoints-list .breakpoint` from using`margin` and `padding` to just `padding` so that the highlight color wraps the entire breakpoint without awkward whitespace

### Testing

* [x] passes `npm test`
* [x] passes `npm run lint`

#### Before
screen shot 2016-10-09 at 3 49 17 pm
**Awkward whitespace issue is more prominent when we turn the breakpoint blue**
-screen shot 2016-10-09 at 3 49 09 pm
#### After
screen shot 2016-10-09 at 3 50 10 pm
**Hover color**
screen shot 2016-10-09 at 3 50 25 pm

### DIFF
--- a/public/js/components/Breakpoints.css
+++ b/public/js/components/Breakpoints.css
@@ -6,8 +6,7 @@
 .breakpoints-list .breakpoint {
   font-size: 12px;
   color: var(--theme-content-color1);
-  margin: 0.25em 0;
-  padding: 0.25em 1px 0 0;
+  padding: 0.5em 1px;
   line-height: 1em;
   position: relative;
 }
@@ -17,7 +16,7 @@
 }
 
 .breakpoints-list .breakpoint.paused {
-  background-color: var(--theme-toolbar-background-alt);
+  background-color: var(--breakpoint-active-color);
 }
 
 .breakpoints-list .breakpoint.disabled .breakpoint-label {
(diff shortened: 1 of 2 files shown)


Zacqary:
Because I used the same `var`s that already existed I'm going to assume that this looks fine in Dark Mode but I couldn't actually figure out how to enable that to take screenshots. How do I do that?

wldcordeiro:
Reading through your changes this looks good. I think we might need to push the CSS changes up to m-c but otherwise this would be good as is. I'll approve and leave open to let @helenvholmes weigh in on the colors. I like it over all however, and it matches the colors used already by the debugger for breakpoints.

Zacqary:
What does m-c refer to in this case?

wldcordeiro:
@Zacqary `m-c` is Mozilla-central the mercurial repository that houses Firefox itself, since the CSS file is in `lib` it's likely shared between us and other devtools so getting it updated there is good.

helenvholmes:
Looks like a solid design improvement to me. 👍🏽

wldcordeiro:
@Zacqary thanks for the help, looks great and @helenvholmes digs it too. Merging. 😄

jasonLaster:
Thanks @Zacqary for the help! 

I'm honestly, not sure if I love the green background because it is so strong, but I understand that it has semantic value.

One slight tweak might be to invert the text, like what we do with the call stack. I also aligned the close button color, but that I'm not sure about that.

helenvholmes:
If we invert the text color, I'd recommend double checking it passes AA.

I have similar feelings about the green (it's quite bright), but the work
done in the PR is still helpful. If we want to change the color that's a
one line change, so I didn't want to block on it.

PR #910, opened 2016-10-10
Fix breakpoint var name typo on dark mode

Zacqary:
Associated issue: #871 

### Summary of Changes

In https://github.com/devtools-html/debugger.html/pull/907 I accidentally overrode `--breakpoint-active-color` in dark mode instead of adding `--breakpoint-active-color-hover`. This fixes the error.

### Testing

* [x] passes `npm test`
* [x] passes `npm run lint`


wldcordeiro:
Thanks for the fix. I had read a bit too quickly during review 😉

Zacqary:
I still used the `-hover` color in https://github.com/devtools-html/debugger.html/pull/911 and this fix is present in that PR too. I figured I should PR this commit separately in case the other one gets blocked by design approval.

wldcordeiro:
@Zacqary thanks for the clarification. I really dig the other PR's looks and will see if we can get design approval on that. Feel free to rebase it on master now to get this change free without needing the commit in your history.

PR #911, opened 2016-10-10
Call out active breakpoint with border

Zacqary:
Turns out that using a `border` was easier than a `::before` element. Its `height` wasn't cooperating at `100%`, and I wasn't sure if setting something like `height: 95%` would be reliable. A `border`, on the other hand, seems to be rendering correctly.

Associated Issue: #871 

### Summary of Changes
screen shot 2016-10-10 at 1 51 59 pm

* Resets the background color of `.breakpoint.paused` to its state before 6441b2 
* Adds a transparent `border-left` to `.breakpoint`s, which turns `--breakpoint-active-color` on `.paused`
* Adjusts margins and padding to compensate for the border, including adding a new `.breakpoint-checkbox` class to the `<input>` and removing its `margin-left`

### Testing

* [x] passes `npm test`
* [x] passes `npm run lint`

### Screenshots/Videos
screen shot 2016-10-10 at 1 51 59 pm
**Hover**
screen shot 2016-10-10 at 1 52 25 pm

### DIFF
--- a/public/js/components/Breakpoints.css
+++ b/public/js/components/Breakpoints.css
@@ -9,6 +9,8 @@
   padding: 0.5em 1px;
   line-height: 1em;
   position: relative;
+  border-left: 4px solid transparent;
+
 }
 
 .breakpoints-list .breakpoint:last-of-type {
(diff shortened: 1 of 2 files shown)


jasonLaster:
Thanks @Zacqary

Issue #884, opened 2016-10-05
Reps - window should be formatted differently


jasonLaster:
We currently format the Window object with the debuggee URL. This is not that helpful. 

**Two rules:**

1. if a variable references the window the value should be `Window`
2. if the scope is window, the label should be `Window` or `Global`. I like global because it is what is most commonly used.

![screen shot 2016-10-05 at 9 01 47 am](https://cloud.githubusercontent.com/assets/254562/19114067/62c9bb2a-8ada-11e6-9206-23963e6c2b15.png)

Here is a [gist](https://gist.github.com/jasonLaster/778a91126cc49a8ee382b9264fa50e70) for getting started.

PR #912, opened 2016-10-10
Display Window or Global on Window class reps

Zacqary:
Associated Issue: #884 

### Summary of Changes

* Window rep will display either `"Window"` or `"Global"` depending on whether its preview has a truthy `isGlobal` property or not
* Top-level Window will override its preview to `{ kind: "Window", isGlobal: true }`

### Notes

This feels like a hack to me, and I'm submitting this PR more to start a discussion than because I'm confident in its ability to fix the issue. I don't understand the code path that renders the reps very well, but this is what I was able to figure out.

The problem that I found is that the top-level Window — the one we want to say `Global` — has a `preview` with `{ kind: "ObjectWithURL" }` rather than `{ kind: "Window" }`. But, apparently, the `view` property in `e` is correctly marked as `{ kind: "Window" }`. I have no idea why it is, or where the `view` property is even coming from, because it's not present in `getBindingVariables`. It seems to be added at some step before the final render, but I wasn't able to track that down.

So that's why I'm not super confident in this fix. It operates under the assumption that if a scope of type `"object"` has a class of `"Window"`, then it must be the global `Window`. I don't know if this is always going to be true. It also doesn't fix the underlying problem of why the global `Window` is being generated with the wrong kind of preview in the first place, which I also wasn't able to track down.

### Testing

* [x] passes `npm test`
* [x] passes `npm run lint`

### Screenshots
screen shot 2016-10-10 at 2 56 46 pm

### DIFF
--- a/public/js/components/Scopes.js
+++ b/public/js/components/Scopes.js
@@ -120,13 +120,11 @@ function getScopes(pauseInfo, selectedFrame) {
         scopes.push({ name: title, path: key, contents: vars });
       }
     } else if (type === "object") {
-      const value = scope.object;
-      // Override preview for Window object
+      let value = scope.object;
+      // If this is the global window scope, mark it as such so that it will
+      // preview Window: Global instead of Window: Window
       if (value.class === "Window") {
-        value.preview = {
-          kind: "Window",
-          isGlobal: true
-        };
+        value = Object.assign({}, scope.object, { isGlobal: true });
       }
       scopes.push({
         name: scope.object.class,
--- a/public/js/components/Scopes.js
+++ b/public/js/components/Scopes.js
@@ -120,10 +120,18 @@ function getScopes(pauseInfo, selectedFrame) {
         scopes.push({ name: title, path: key, contents: vars });
       }
     } else if (type === "object") {
+      const value = scope.object;
+      // Override preview for Window object
+      if (value.class === "Window") {
+        value.preview = {
+          kind: "Window",
+          isGlobal: true
+        };
+      }
       scopes.push({
         name: scope.object.class,
         path: key,
-        contents: { value: scope.object }
+        contents: { value }
       });
     }
   } while (scope = scope.parent); // eslint-disable-line no-cond-assign
--- a/public/js/lib/devtools/client/shared/components/reps/window.js
+++ b/public/js/lib/devtools/client/shared/components/reps/window.js
@@ -37,8 +37,8 @@ define(function (require, exports, module) {
       return "";
     },
 
-    getLocation: function (grip) {
-      return getURLDisplayString(grip.preview.url);
+    getDisplayValue: function (grip) {
+      return grip.preview.isGlobal ? "Global" : "Window";
     },
 
     render: function () {
@@ -48,7 +48,7 @@ define(function (require, exports, module) {
         DOM.span({className: "objectBox objectBox-Window"},
           this.getTitle(grip),
           DOM.span({className: "objectPropValue"},
-            this.getLocation(grip)
+            this.getDisplayValue(grip)
           )
         )
       );
--- a/public/js/lib/devtools/client/shared/components/reps/window.js
+++ b/public/js/lib/devtools/client/shared/components/reps/window.js
@@ -24,6 +24,7 @@ define(function (require, exports, module) {
 
     propTypes: {
       object: React.PropTypes.object.isRequired,
+      mode: React.PropTypes.string
     },
 
     getTitle: function (grip) {
@@ -37,8 +38,16 @@ define(function (require, exports, module) {
       return "";
     },
 
+    getLocation: function (grip) {
+      return getURLDisplayString(grip.preview.url);
+    },
+
     getDisplayValue: function (grip) {
-      return grip.preview.isGlobal ? "Global" : "Window";
+      if (this.props.mode === "tiny") {
+        return grip.isGlobal ? "Global" : "Window";
+      } else {
+        return this.getLocation(grip);
+      }
     },
 
     render: function () {


jasonLaster:
Thanks @Zacqary! I think you're on the right track, but I'm going to bump this review to @janodvarko, @linclark, who have worked more on Reps than I have.

public/js/components/Scopes.js
```diff
@@ -120,10 +120,18 @@ function getScopes(pauseInfo, selectedFrame) {
...
       }
     } else if (type === "object") {
+      const value = scope.object;
+      // Override preview for Window object
+      if (value.class === "Window") {
+        value.preview = {
+          kind: "Window",
+          isGlobal: true
+        };
+      }
```
jasonLaster: hmm, interesting - i'm not sure if we have mutated the rep value's before. Curious to see what others think.

public/js/lib/devtools/client/shared/components/reps/window.js
```diff
@@ -37,8 +37,8 @@ define(function (require, exports, module) {
       return "";
     },
 
-    getLocation: function (grip) {
-      return getURLDisplayString(grip.preview.url);
+    getDisplayValue: function (grip) {
+      return grip.preview.isGlobal ? "Global" : "Window";
```
jasonLaster: Thanks @janodvarko. 

I think updating the *tiny* mode makes a lot of sense. We're already forcing our grip's to be [tiny](https://github.com/devtools-html/debugger.html/blob/master/public/js/components/ObjectInspector.js#L136).

The motivation behind formatting the window object with `Window` or `Global` is in part space constraints and in part it's clearer, in the same way that labeling Arguments as opposed to referring to it as Object. The discussion is here #884.

public/js/components/Scopes.js
jlongster: Yeah, I think we should avoid it. We want to be careful about mutations because they cause surprising effects down the road. It might be good to ask Honza about this.

jlongster: Oh, just noticed that he commented below. Sounds like we have a different approach in mind now?

public/js/lib/devtools/client/shared/components/reps/window.js
Zacqary: Just updated the Window rep to use `tiny` mode, and handled attaching the `isGlobal` property by cloning the rep value instead of mutating it.

I didn't realize `kind = ObjectWithUrl` was getting routed correctly to the `Window` rep component, so I also removed the kind change.

I'm still not sure I entirely understand the implications of the `else if (type === "object")` block where the `isGlobal` prop gets added, but it seems to work.

public/js/components/Scopes.js
janodvarko: Yes, different approach suggested. And I agree, we should avoid rep value mutations.

jasonLaster:
Thanks @Zacqary! This looks good to me.

public/js/components/Scopes.js
```diff
@@ -120,10 +120,16 @@ function getScopes(pauseInfo, selectedFrame) {
...
+      // If this is the global window scope, mark it as such so that it will
+      // preview Window: Global instead of Window: Window
+      if (value.class === "Window") {
+        value = Object.assign({}, scope.object, { isGlobal: true });
+      }
       scopes.push({
         name: scope.object.class,
         path: key,
-        contents: { value: scope.object }
+        contents: { value }
```
jasonLaster: I like the way that this updates the grip without mutating the original source.

public/js/lib/devtools/client/shared/components/reps/window.js
```diff
@@ -41,14 +42,22 @@ define(function (require, exports, module) {
       return getURLDisplayString(grip.preview.url);
     },
 
+    getDisplayValue: function (grip) {
+      if (this.props.mode === "tiny") {
+        return grip.isGlobal ? "Global" : "Window";
+      } else {
+        return this.getLocation(grip);
+      }
+    },
```
jasonLaster: I like that this doesn't change the original behavior, except for the tiny mode.

public/js/lib/devtools/client/shared/components/reps/window.js
jasonLaster: @janodvarko mind taking another look. Once this lands, I'd be happy to make an m-c patch and perhaps add some m-c tests.

public/js/components/Scopes.js
janodvarko: Yes, looks correct to me to.

public/js/lib/devtools/client/shared/components/reps/window.js
janodvarko: Nice, that's exactly what 'modes' are for.

jasonLaster:
Thanks @Zacqary! That was not an *easy* one :)

Zacqary:
No problem. Yeah, definitely a challenge but I've gotten pretty good at
grepping through the breadcrumbs of React codebases.

The one thing I still wasn't able to figure out was where all that
additional stuff, like the view prop, was getting added to the event object
before rendering. The props chain gets very hard to follow somewhere in the
middle of it all, with all the renders of component classes that get
computed from lists.

Also, are there any docs on what exactly all the terms like "rep" and
"grip" mean?

jasonLaster:
> props chain very hard to follow somewhere in the middle of it all, with all the renders of component classes that get computed from lists.

Yep and you're in a very hard spot with the Tree + Rep... The sources tree is also confusing for other reasons, but this is as complicated as it gets :)

> Also, are there any docs on what exactly all the terms like "rep" and "grip" mean?

:) I don't think so... @janodvarko would likely know. This would be tremendously helpful if we try to unify Firefox and Chrome's definition of remote objects. https://bugzilla.mozilla.org/show_bug.cgi?id=1307877

janodvarko:
> Also, are there any docs on what exactly all the terms like "rep" and "grip" mean?

Grip comes from RDP protocol:
https://wiki.mozilla.org/Remote_Debugging_Protocol#Grips

Rep stands for representation of data.

Honza

Issue #916, opened 2016-10-11
Watch Expressions 


jbhoosreddy:
Tracking ticket for watch expression

Tasks:
- [x] Polish UI #1599 
  - [x] Prevent duplicate expressions 
  - [x] Fix editing expressions 
  - [x] Re-evaluate on resume
  - [x] cursor on input should be "text"
  - [x] close button is off
  - [x] expressions should not expand by default
  - [x] re-evaluate when a frame is selected
  - [x] can't remove last expression
- [x] UI wonkiness #1610 
- [x] editing / adding wonkiness #1609
- [x] add watch expressions pane in vertical mode #1649
- [x] Refactoring #1650
   - [x] Extract expressions actions and reducer
  - [x] remove expression id (because they're unique!)
- [x] add selection to watch expressions
- [x] Final expression doesn't remove #1594 
- [x] clean up list view #1522
- [x] cleanup close button #1523
- [x] Adjust header buttons height so that the header is normal height #1524 
- [x] Clean up input UI #1491
- [x] improve the preview https://github.com/devtools-html/debugger.html/issues/1350
- [x] evaluate expression when not paused #1525 
- [x] edit expression when paused #1526 
- [x] review refreshing an expression #1521
- [x] cleanup the header buttons https://github.com/devtools-html/debugger.html/issues/1347
- [x] Header button to show expression input
- [x] Header button to refresh (re-evaluate) watched expressions
- [x] Adjust colors to match other panels in the accordion
- [x] display `<not available>` when value is undefined
- [x] refresh button on frame.
- [x] Add watch expressions by context menu in editor #1596
- [x] Bug - Watch expressions don't evaluated on Chrome (suggestions?)  (yup, ignore it)

jasonLaster:
@jbhoosreddy two things:
- i've noticed that the values might not be computing on the pause action.
- the global value is currently the URL of the page, but that is changing here #884, which i think will be a nice stylistic improvement.

PR #917, opened 2016-10-11
Added header button in accordion for watch expressions (#916)

jbhoosreddy:
Associated Issue: #916 
### Summary of Changes
- Added header button for displaying the watch expression input.
  For now I used a placeholder icon instead of the `+` icon.
### Testing
- [x] passes `npm test`
- [x] passes `npm run lint`

There are two new buttons in the header. The icons is temporary but, the first one is to toggle expression input while the second is to refresh the expressions. 
screen shot 2016-10-17 at 3 55 20 am


public/js/actions/pause.js
```diff
@@ -206,6 +206,15 @@ function addExpression(expression) {
   };
 }
 
+function setExpressionInputVisibility(visibility) {
```
jasonLaster: Hmm, at this point we're opting to use component state local UI data. Some examples of this are, the search bar value, app search being shown, ... 

I think at some point, we might want to start using redux state more for UI data like this, but for the moment it is simpler to use component state.

public/js/components/RightSidebar.css
```diff
@@ -68,3 +68,7 @@ span.pause-exceptions.uncaught {
 span.pause-exceptions.all {
   stroke: var(--theme-highlight-blue);
 }
+
+.accordion-button {
+  float: right;
+}
```
jasonLaster: This works, but it might be nice to try flex as well. An example of that is the autocomplete layout.

wldcordeiro:
Screenshots please @jbhoosreddy? 😄

jbhoosreddy:
@wldcordeiro I updated my PR with a screenshot. Actually, you may be able to answer this. I am trying to add a hover state but I can't seem to change the style of expression value which is a Rep component, so it doesn't look as good yet.

jbhoosreddy:
@jasonLaster I did not yet see it not updating expressions on pause, but I'll keep a look out for it. I did notice sometimes the expression input opens up a new `SOURCEundefined` tab. Also watch expressions does not seem to return a value on Chrome even though the code is executing completely. I should push my latest changes soon.

jasonLaster:
Cool. 

What do you mean by hover?

Chrome won't wont work yet. That's fine. 

Undefined is strange, when do you see that?

jbhoosreddy:
> Chrome won't wont work yet. That's fine.

Oh I didn't know that.

> Undefined is strange, when do you see that?

STR:

1. Enter `document` or `window` in the expression input. 

ER: expression value would be document uri
AR: <not available> and the **debugger is paused**. If you `resume`, the correct expression value is applied.

Here's what I think: If you remember we saw something exactly similar when writing tests with Cypress. I tried to debug the `evaluate` implementation on the debugger server. When the server gets script "document" it returns back a consoleActor with the text "document" which the debugger adds as a source in the sources list. And the debugger is paused. Once you resume it, it goes ahead and emits the result once the console evaluates the script and that propagates back.

While I understood/think I understood the execution of the code, I don't think I can solve it. Would you mind looking at this with me @jasonLaster?

jasonLaster:
@jbhoosreddy yeah - find a time to pair on this issue.

jasonLaster:
@jbhoosreddy, @wldcordeiro I just took another look at this PR and there are some really good things in it:

1. it adds the close button
2. it adds styling for errors
3. it adds the header buttons
4. other good stuff i'm probably missing

the biggest critique i have at the moment is that it sets the input visibility in the redux store and i think it should be local component state. @jbhoosreddy I think i can finish this up if you'd like. or feel free to jump in and take a look. 

> Enter document or window in the expression input

I think this is an edge case we can ignore until we get the basic mechanics in. I have a PR that i'll open soon that fixes some of the redux logic w/ the client.

jbhoosreddy:
I'll make those changes and rebase to clean up this PR, as per your suggestions today.

jbhoosreddy:
I updated the PR with the change as suggested. I left the CSS stuff for now as some work has to be done there.. I'll take that up separately.

PR #919, opened 2016-10-11
Fix sourcemap test

jlongster:
The sourcemap test sometimes fails. In this PR I will fix it.

This code is not actually a fix yet. I just added some logging to figure this out, and will continually push here. I can't reproduce this locally and can't think of why it would be racy.


jlongster:
I think this fixes it. The problem (I think) was that the scripts that the test was using could be GCed at any time; they didn't introduce any global variables. This means that if a test ran *just* slow enough, we won't see the sources because they'll be GCed before the debugger is connected. I discovered this because I added logs and didn't even see any logs come through; no new sources were even being added to the debugger.

jlongster:
I will retrigger it on CI several times before merging. But this can be reviewed.

jlongster:
@wldcordeiro note: this should fix the problematic test

Issue #101, opened 2016-04-26
Investigate Conditional Breakpoints


clarkbw:
**User Story**

As a developer who is debugging their code I need a breakpoint that only triggers under certain scenarios such that when a certain expression is true my breakpoint will pause execution.

**Notes**

Chrome Implementation:&nbsp; [https://developer.chrome.com/devtools/docs/javascript-debugging/conditional-breakpoint-region.jpg](https://developer.chrome.com/devtools/docs/javascript-debugging/conditional-breakpoint-region.jpg)

**UX Needs**

- Button or entry point to open a small input dialog
- Input dialog that explains expression needs

- [ ] Initial button to add breakpoint

jasonLaster:
I'll have some time later today to start working on this

jasonLaster:
We are not going to get to this pre-london, so closing for the time being. Lets re-open when we're ready to tackle.

jasonLaster:
Re-opening to track the work on this feature.

wldcordeiro:
Whoa, I literally had a conversation with my coworker yesterday about some of the ideas we had for the debugger and he mentioned exactly this! Nice work on the PR so far @jasonLaster

jasonLaster:
I added some initial conditional breakpoint design mockups [here](https://invis.io/UM8XCLNK5#/196300930_Conditional_Breakpoint_Pane_Exploration)

Notes:
* I tried to model this after the watch expressions work where there's an left prompt for adding input
* I assumed we'd be able to add a line widget which is 1.5 x a line height. This might be impossible and we'll have to adjust.
* I used a mono spaced font for the input, i'm not sure if we do that in other places but it felt better for code
* I did not include a close button, but assumed that escape would close it, although likely only when the input is in focus.


![screen shot 2016-10-12 at 1 02 21 pm](https://cloud.githubusercontent.com/assets/254562/19319683/f852de32-907b-11e6-9986-9c44fcbf046d.png)

helenvholmes:
> - I did not include a close button, to be consistent with the search bar. Escape and enter would close the panel.
I think the search bar probably needs a close, as does this—I know automatically I would want to close it.

> - The gutter next to the panel seems very empty now. It would be interesting to explore adding something there.
My recommendation would be to extend the message into the gutter, and left-pad the guillemet until the editor begins. (That's what I would try, anyway.)

Looks really good! I like the styling decisions you've made 🎉

jasonLaster:
> close button

[...]

PR #922, opened 2016-10-12
Add conditional breakpoints

jasonLaster:
Associated Issue: #101

### Summary of Changes

I paired with Giorgio @ppold on this feature last night. This PR is an initial conditional breakpoints implementation, it satisfies the basic requirements and will be polished in subsequent PRs.

User steps:
* click to add a breakpoint
* cmd+click to edit breakpoint
* add a condition
* enter to save

Next steps:
* clicking on the breakpoint while the panel is open should close the panel and remove the breakpoint
* the panel input should be focused when the panel is opened
* style the panel. Perhaps move the description to input placeholder text (shorten dramatically)
* perhaps add a close button
* perhaps update the breakpoint in the breakpoint list to either show it's a conditional or show the condition. Chrome does not do anything here...

### DIFF
--- a/public/js/actions/breakpoints.js
+++ b/public/js/actions/breakpoints.js
@@ -180,28 +180,29 @@ function toggleAllBreakpoints(shouldDisableBreakpoints: Boolean) {
  *        The condition to set on the breakpoint
  */
 function setBreakpointCondition(location: Location, condition: string) {
-  throw new Error("not implemented");
+  return ({ dispatch, getState, client }: ThunkArgs) => {
+    const bp = getBreakpoint(getState(), location);
+    if (!bp) {
+      throw new Error("Breakpoint does not exist at the specified location");
+    }
+    if (bp.loading) {
+      // TODO(jwl): when this function is called, make sure the action
+      // creator waits for the breakpoint to exist
+      throw new Error("breakpoint must be saved");
+    }
 
-  // return ({ dispatch, getState, client }) => {
-  //   const bp = getBreakpoint(getState(), location);
-  //   if (!bp) {
-  //     throw new Error("Breakpoint does not exist at the specified location");
-  //   }
-  //   if (bp.get("loading")) {
-  //     // TODO(jwl): when this function is called, make sure the action
-  //     // creator waits for the breakpoint to exist
-  //     throw new Error("breakpoint must be saved");
-  //   }
-
-  //   return dispatch({
-  //     type: constants.SET_BREAKPOINT_CONDITION,
-  //     breakpoint: bp,
-  //     condition: condition,
-  //     [PROMISE]: Task.spawn(function* () {
-  //       yield client.setBreakpointCondition(bp.get("id"), condition);
-  //     })
-  //   });
-  // };
+    return dispatch({
+      type: constants.SET_BREAKPOINT_CONDITION,
+      breakpoint: bp,
+      condition: condition,
+      [PROMISE]: client.setBreakpointCondition(
+        bp.id,
+        location,
+        condition,
+        isOriginalId(bp.location.sourceId)
+      )
+    });
+  };
 }
 
 module.exports = {
--- a/public/js/components/Editor.js
+++ b/public/js/components/Editor.js
@@ -26,6 +26,43 @@ function isTextForSource(sourceText) {
   return !sourceText.get("loading") && !sourceText.get("error");
 }
 
+function breakpointAtLine(breakpoints, line) {
+  return breakpoints.find(b => {
+    return b.location.line === line + 1;
+  });
+}
+
+function renderConditionalBreakpointPanel(
+  { location, setBreakpointCondition, condition, closePanel }) {
+  function onKey(e) {
+    if (e.key != "Enter") {
+      return;
+    }
+
+    setBreakpointCondition(location, e.target.value);
+    closePanel();
+  }
+
+  let panel = document.createElement("div");
+  ReactDOM.render(
+    dom.div(
+      { className: "conditional-breakpoint-panel" },
+      dom.div(
+        { className: "header" },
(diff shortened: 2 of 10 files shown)


public/js/components/Editor.js
```diff
@@ -26,6 +26,41 @@ function isTextForSource(sourceText) {
...
+    if (e.key != "Enter") {
+      return;
+    }
+
+    setBreakpointCondition(location, e.target.value);
+    closePanel();
+  }
+
+  let panel = document.createElement("div");
+  ReactDOM.render(
```
jasonLaster: we likely can replace the react rendering with a simple vanilla approach, but on the other hand this *does* work.

public/js/components/Editor.js
```diff
@@ -66,6 +117,8 @@ const Editor = React.createClass({
         line: line + 1
       });
     } else {
+
+
```
jasonLaster: unnecessary whitespace

public/js/clients/firefox/commands.js
```diff
@@ -54,22 +54,26 @@ function setBreakpoint(location, condition, noSliding) {
...
-      id: bpClient.actor,
-      actualLocation: Location(actualLocation)
-    });
+  }).then((res) => onNewBreakpoint(location, res))
+}
+
+function onNewBreakpoint(location, res) {
+  const bpClient = res[1];
+  let actualLocation = res[0].actualLocation;
+  bpClients[bpClient.actor] = bpClient;
```
jasonLaster: we likely can clean this up with some destructuring

jasonLaster: also, setCondition does not return `actualCondition`, but that is likely okay because you're just editing the bp.

public/js/actions/breakpoints.js
```diff
@@ -187,28 +187,31 @@ function toggleAllBreakpoints(shouldDisableBreakpoints: Boolean) {
...
+
+  return ({ dispatch, getState, client }) => {
+    const bp = getBreakpoint(getState(), location);
+    if (!bp) {
+      throw new Error("Breakpoint does not exist at the specified location");
+    }
+    if (bp.loading) {
+      // TODO(jwl): when this function is called, make sure the action
+      // creator waits for the breakpoint to exist
+      throw new Error("breakpoint must be saved");
```
jasonLaster: We probably should replace the Errors with `assert` functions

public/js/actions/breakpoints.js
```diff
@@ -187,28 +187,31 @@ function toggleAllBreakpoints(shouldDisableBreakpoints: Boolean) {
...
-  //   return dispatch({
-  //     type: constants.SET_BREAKPOINT_CONDITION,
-  //     breakpoint: bp,
-  //     condition: condition,
-  //     [PROMISE]: Task.spawn(function* () {
-  //       yield client.setBreakpointCondition(bp.get("id"), condition);
-  //     })
-  //   });
-  // };
+  // throw new Error("not implemented");
```
jasonLaster: should remove this

config/development.json
```diff
@@ -11,7 +11,8 @@
     "sourceMaps": true,
     "prettyPrint": false,
     "watchExpressions": false,
-    "search": true
+    "search": true,
+    "conditionalBreakpoints": true
```
jlongster: I don't think we need to put this behind a flag. Let's just land features and make sure they work from the start. If there are major bugs I think we should fix in the the first PR. But I doubt there are major bugs, there might just be UX improvements we can do. That's the point of landing in nightly though; we have time until uplift to tweak it to get it great.

public/js/actions/breakpoints.js
jlongster: There no assertions here though; it should always just throw an error. `assert` may evolve in the future to only happen in dev or somethings. But this should always through.

public/js/actions/breakpoints.js
jlongster: Yeah, go ahead and remove these comments

public/js/clients/firefox/commands.js
```diff
@@ -79,6 +83,12 @@ function removeBreakpoint(breakpointId) {
   return bpClient.remove();
 }
 
+function setBreakpointCondition(breakpointId, location, condition, noSliding) {
+  const bpClient = bpClients[breakpointId];
+  return bpClient.setCondition(threadClient, condition, noSliding)
+    .then((res) => onNewBreakpoint(location, [{}, res]))
```
jlongster: You should probably remove the old breakpoint client from `bpClients` here before we add a new one.

jlongster: Also, can you rename `res` to `bpClient` or something to make it clear that it's just the breakpoint client?

public/js/components/Editor.js
```diff
@@ -26,6 +26,41 @@ function isTextForSource(sourceText) {
...
+      className: "conditional-breakpoint-panel"
+    },
+    dom.div(
+      { className: "header" },
+      `This breakpoint will stop execution only
+       if the following expression is true`
+    ),
+    dom.input({
+      defaultValue: condition,
+      onKeyPress: (e) => onKey(e),
```
jlongster: nit: no space around single arg arrow functions

jlongster:
The approach looks fine. A few nits, and I think we should just land it without a flag so that it just gets used and we'll fix bugs when they happen. It doesn't really affect the UI because you only get to it via a key command.

public/js/components/Editor.js
jasonLaster: we can do better and just do `onKeyPress: onKey` :P

jasonLaster:
I'd like to leave the flag for now. I'm finding some other UX issues to tweak, which can be done in subsequent issues.

ping @jlongster - mind taking another look.

Issue #891, opened 2016-10-06
Editor - the footer overlays the editor


jasonLaster:
The footer is overlaying the bottom of the footer. The fix is likely to position the footer relative to the editor. 

Note, i think that I positioned it absolutely initially because the footer was sometimes hidden...

![screen shot 2016-10-06 at 11 19 51 am](https://cloud.githubusercontent.com/assets/254562/19158397/86f7c30c-8bb6-11e6-87f9-09284f4eb842.png)

wldcordeiro:
Footer overlays footer eh? 😆 I'd like to take a look at the layout of the center pane and get an idea of what the main arrangement is and take a stab at refactoring it to be flexbox based. 😄

jasonLaster:
@wldcordeiro that would be great. You're the grid + jss guy, so it makes sense. 

I remember last time the issue was:
* codemirror has some flexbox bugs and does not expand if it's a child... (maybe)
* chrome is different than firefox... so this is one of those issues where it's important to check both browsers

fixes:
* Flexbox: 0 for the bar + footer could be good
* also position relative with a greater wrapper height reduction could work too perhaps...

PR #924, opened 2016-10-12
Rework Editor/Center pane layout with flexbox to be less brittle.

wldcordeiro:
Associated Issue: #891

### Summary of Changes

* rework the CSS that handled the layout of the editor/center pane to use flexbox more intelligently. This corrects the footer being over the editor. 

### Testing

* [x] passes `npm test`
* [x] passes `npm run lint`

### DIFF
--- a/public/js/components/App.css
+++ b/public/js/components/App.css
@@ -40,8 +40,9 @@ body {
 }
 
 .editor-container {
-  display: flex;
-  flex: 1;
+  position: absolute;
+  width: 100%;
+  height: 100%;
 }
 
 .subsettings:hover {
--- a/public/js/components/App.css
+++ b/public/js/components/App.css
@@ -43,6 +43,8 @@ body {
   position: absolute;
   width: 100%;
   height: 100%;
+  display: flex;
+  flex-direction: column;
 }
 
 .subsettings:hover {
--- a/public/js/components/Editor.css
+++ b/public/js/components/Editor.css
@@ -11,6 +11,7 @@
 .editor-wrapper {
   display: flex;
   flex-direction: column;
+  order: 2;
 }
 
 html[dir="rtl"] .editor-mount {
@@ -38,17 +39,17 @@ html[dir="rtl"] .editor-mount {
 
 .search-bar ~ .editor-mount,
 .search-bar ~ .editor-mount .CodeMirror {
-  order: 3;
+  order: 2;
   height: 87vh;
 }
 
 .search-bar ~ .source-footer {
-  order: 4;
+  order: 3;
 }
 
 .editor-mount,
 .CodeMirror {
-  order: 2;
+  order: 1;
   width: 100%;
   height: 92vh;
   background-color: var(--theme-body-background);
--- a/public/js/components/Editor.css
+++ b/public/js/components/Editor.css
@@ -36,8 +36,14 @@ html[dir="rtl"] .editor-mount {
   opacity: 0.3;
 }
 
-.search-bar ~ .editor-mount {
-  height: calc(100% - 58px);
+.search-bar ~ .editor-mount,
+.search-bar ~ .editor-mount .CodeMirror {
+  order: 3;
+  height: 87vh;
+}
+
+.search-bar ~ .source-footer {
+  order: 4;
 }
 
 .editor-mount,
--- a/public/js/components/Editor.css
+++ b/public/js/components/Editor.css
@@ -40,7 +40,7 @@ html[dir="rtl"] .editor-mount {
 .search-bar ~ .editor-mount,
 .search-bar ~ .editor-mount .CodeMirror {
   order: 2;
-  height: 87vh;
+  height: 86.8vh;
 }
 
 .search-bar ~ .source-footer {
--- a/public/js/components/Editor.css
+++ b/public/js/components/Editor.css
@@ -9,11 +9,8 @@
  * BUG https://github.com/devtools-html/debugger.html/issues/63
  */
 .editor-wrapper {
-  position: absolute;
-  height: calc(100% - 31px);
-  width: 100%;
-  top: 30px;
-  left: 0px;
+  display: flex;
+  flex-direction: column;
 }
 
 html[dir="rtl"] .editor-mount {
@@ -45,8 +42,9 @@ html[dir="rtl"] .editor-mount {
 
 .editor-mount,
 .CodeMirror {
+  order: 2;
   width: 100%;
-  height: 100%;
+  height: 92vh;
   background-color: var(--theme-body-background);
 }
 
@@ -94,9 +92,6 @@ html[dir="rtl"] .editor-mount {
 
   /* Offsetting it by 30px for the sources-header area */
   height: calc(100% - 30px);
-  position: absolute;
-  top: 30px;
-  left: 0;
   padding: 50px 0;
   text-align: center;
   font-size: 1.25em;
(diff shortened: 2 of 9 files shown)


wldcordeiro:
Still a little WIP here. @jasonLaster wasn't the biggest fan of moving the tabs into `Editor` but that simplified layout work. Also I'm fixing some stuff with the breakpoints CSS as they're hidden currently by the changes. I'll update this shortly with the fixes for that.

jasonLaster:
Thanks @wldcordeiro mind sharing some diagrams with the current options. Curious to see what we can do here. I do feel better seeing that Source Tabs pulls in tabs through connect and the Editor doesn't have to get that data.

Also did you try this in chrome?

public/js/components/EditorSearchBar.js
```diff
@@ -126,7 +126,7 @@ const EditorSearchBar = React.createClass({
 
   render() {
     if (!isEnabled("search") || !this.state.enabled) {
-      return dom.div();
+      return null;
```
wldcordeiro: Returning null is nicer as it's easier to work around in CSS than invisible divs.

public/js/components/Editor.css
```diff
@@ -35,14 +32,21 @@
...
+.search-bar ~ .source-footer {
+  order: 4;
 }
 
 .editor-mount,
 .CodeMirror {
+  order: 2;
   width: 100%;
-  height: 100%;
+  height: 92vh;
```
wldcordeiro: These heights were found to be just enough to avoid extra scrollbars and allow for the editor to appear full height since it doesn't work quite right with `100%` and flexbox.

public/js/components/Editor.js
```diff
@@ -271,6 +272,7 @@ const Editor = React.createClass({
     return (
       dom.div(
         { className: "editor-wrapper devtools-monospace" },
+        SourceTabs(),
```
wldcordeiro: Moving the tabs into the editor makes them all part of the same layout flow which makes flexbox easier to layout as otherwise we have two different layouts that work nearly identically. 🙁 Still looking to see if it's avoidable. Also made the `active tab` no longer look like it's part of the editor document unfortunately so I'm still experimenting.

wldcordeiro:
Alright I think this is ready now. Tested well in Chrome and Firefox for me. Can I get one of you to try these changes locally @jasonLaster @jlongster?

public/js/components/SourceTabs.css
```diff
@@ -1,7 +1,8 @@
 .source-header {
   border-bottom: 1px solid var(--theme-splitter-color);
   height: 30px;
-  flex: 1;
+  flex: 0 0 30px;
```
wldcordeiro: Correction on the basis. Set it to auto, otherwise we don't get the nice active tab is borderless on the bottom effect.

public/js/components/App.css
```diff
@@ -40,8 +40,11 @@ body {
 }
 
 .editor-container {
+  position: absolute;
+  width: 100%;
+  height: 100%;
   display: flex;
-  flex: 1;
+  flex-direction: column;
```
jasonLaster: 👍

public/js/components/Editor.css
```diff
@@ -35,14 +33,21 @@
   opacity: 0.3;
 }
 
-.search-bar ~ .editor-mount {
-  height: calc(100% - 58px);
+.search-bar ~ .editor-mount,
+.search-bar ~ .editor-mount .CodeMirror {
+  order: 2;
+  height: 87vh;
```
jasonLaster: what does this number mean?

public/js/components/EditorSearchBar.js
jasonLaster: got it

jlongster:
This generally looks good. I noticed an error state that's much worse with this PR though. Here's STR:

1. Go to http://todomvc.com/examples/react/#/
2. Open `js/todoModel.js`
3. Set a breakpoint on line 54, inside the `toggle` function
4. Toggle an item to trigger the breakpoint

For some reason, if you look at the call stack and click the second frame (titled `TodoApp<.toggle`) things look bad. Admittedly, it was broken before this PR. Here's how it looked before the PR:

screen shot 2016-10-12 at 5 13 33 pm

Basically, the tabs shift down for some reason, it doesn't get a title, and there's no pause line. Several problems there, this source must be a strange inline source. But at least the tabs shouldn't break like that.

Your PR adds a new problem which looks a lot worse:

screen shot 2016-10-12 at 5 14 02 pm

I don't know why those grey circles are there.

Can you possibly see if there's a quick fix to at least make sure it doesn't look *that bad*? Because if the user gets in that state, that's pretty embarassing. You don't need to fix all the problems (no tab title, debug line, etc) we can do that separately.

wldcordeiro:
@jlongster I'm going to keep working on this. I've found a number of oddities when testing it in panel that aren't present in tabbed view, plus with the things you found too it's just a number of things to research/explore. But one thing is certain, I'm starting to hate CodeMirror and its styling. 👍 😆

jlongster:
@wldcordeiro awesome, thanks! yeah... it's very annoying. Especially because we may get slightly different styles in the panel (even though technically we're using the same files, there may be be one or two slight differences coming from a file we may not have).

wldcordeiro:
@jlongster so it appears those dots are the `subsettings` menu. They were absolutely positioned before and go haywire here. I'm trying to figure out the shifting tabs issue now too. It also appears that the subsettings is broken in master right now in that the dots aren't showing up in the button. I'll be pushing more changes here to fix this and improve the toolbox rendering (min/max heights ftw!)

jlongster:
@wldcordeiro awesome thanks! Feel free to tell us if you want to go ahead and merge this PR and do a follow-up as well. I was thinking of trying to get this in to the debugger update I'm about to push out right now, but I think we'll wait for the next one (we can update it as many times as we like). So I'm find merging this if you want to do a follow-up for some of the smaller things.

public/js/components/Editor.css
```diff
@@ -35,14 +32,21 @@
   opacity: 0.3;
 }
 
-.search-bar ~ .editor-mount {
-  height: calc(100% - 58px);
+.search-bar ~ .editor-mount,
+.search-bar ~ .editor-mount .CodeMirror {
+  order: 3;
+  height: 87vh;
```
rheesn: Please help me with the changes

wldcordeiro: @rheesn What? This is PR I'm working on I'm not sure what you're talking about.

jasonLaster:
@wldcordeiro where is this work at? Sorry it's been a couple of days and I lost context, I suppose there are two issues - the subsetting circles and the footer.

wldcordeiro:
Yeah so I'm moving the subsettings/overflow-tabs work into a separate PR. This PR however I'm still trying to figure out the heights properly so that they work for in the firefox devtools panel since that seems to be the problem area. I'll be pushing today with some changes.

jasonLaster:
Adding this patch as a second option, which is ugly because of the *magic* numbers, but could be simpler than the flex solution

```diff
diff --git a/public/js/components/Editor.css b/public/js/components/Editor.css
index e521149..cdd5a24 100644
--- a/public/js/components/Editor.css
+++ b/public/js/components/Editor.css
@@ -39,8 +39,12 @@
   opacity: 0.3;
 }

+.editor-wrapper .editor-mount {
+  height: calc(100% - 32px);
+}
+
 .search-bar ~ .editor-mount {
-  height: calc(100% - 58px);
+  height: calc(100% - 56px);
 }
```

PR #925, opened 2016-10-12
remove `babel` dependency since it's moved to `babel-cli`

hzoo:
the `babel` package in 6.x currently doesn't do anything, so you should be seeing the deprecation message from `babel` (noticed this from seeing https://github.com/yarnpkg/yarn/issues/638)

```
warning babel@6.5.2: Babel's CLI commands have been moved from the babel package to the babel-cli package
```

As an aside I don't `babel-preset-es2015` or `babel-preset-stage-0` being used atm? (and maybe give 
https://github.com/babel/babel-preset-env a shot)

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,6 @@
   },
   "devDependencies": {
     "amd-loader": "0.0.5",
-    "babel": "^6.5.2",
     "babel-cli": "^6.7.5",
     "babel-core": "^6.17.0",
     "babel-eslint": "^6.1.2",


wldcordeiro:
@hzoo there's an intentional decision behind using presets that aren't part of the language proper `stage-0`, and I think the idea behind avoiding `es2015` was to add in the plugins we use one by one to avoid bundling things we don't use. Still removing an unused import might be nice though this may be covered by @jasonLaster's work in extracting the toolbox too. Thanks for the PR, I'll leave it open for a bit for some discussion.

hzoo:
@wldcordeiro Sorry if I wasn't being clear but it looks like both preset-es2015 and preset-stage-0 are in the package.json dependencies but aren't used as I can tell?

https://github.com/babel/babel-preset-env is a new preset I just started that is supposed to determine the correct plugins based on the browser/env supported (chrome, ff, node, etc)

jasonLaster:
Thanks, I think we can safely get rid of the presets as well.

wldcordeiro:
@hzoo ah thanks for the clarification, I thought you were suggesting the two presets which there was as discussion of somewhat recently. It does look like they could be removed too.

jlongster:
So this can be merged?

hzoo:
Yeah, I added the deprecation message in a while ago too https://github.com/babel/babel/blob/master/packages/babel/cli.js

PR #926, opened 2016-10-12
547 toggle breakpoint preserve state  rework

ThomasCrevoisier:
Associated Issue: #547

### Summary of Changes

* Implement the toggleAllBreakpoint feature for Firefox and Chrome
* Minor changes on `actions/breakpoints.js`
* Rework of the Firefox `commands.js` file to keep disabled breakpoints in memory, used when enabling all breakpoints again

### Testing

* [x] passes `npm test`
* [x] passes `npm run lint`

### DIFF
--- a/public/js/reducers/breakpoints.js
+++ b/public/js/reducers/breakpoints.js
@@ -110,6 +110,19 @@ function update(state = State(), action: Action) {
       if (action.status === "start") {
         return state.set(
           "breakpointsDisabled", action.shouldDisableBreakpoints);
+      } else if (action.status === "done") {
+        return action.value.reduce((updatedState, bp) => {
+          const locationId = makeLocationId(bp.actualLocation);
+          const breakpoint = updatedState.breakpoints.get(locationId);
+          const bpId = bp.id;
+
+          return updatedState.setIn(["breakpoints", locationId],
+            updateObj(breakpoint, {
+              disabled: action.shouldDisableBreakpoints,
+              id: bpId
+            })
+          );
+        }, state);
       }
       break;
     }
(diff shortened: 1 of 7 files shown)


ThomasCrevoisier:
@jasonLaster @jlongster Sorry it took so long. I've been busy those past few days.

This is the rework of #814.

jasonLaster:
No worries, I'll take a look now!

jasonLaster:
Looks like there's one small flow issue with `shouldDisableBreakpoints`, which could be fixed by going into [types](https://github.com/devtools-html/debugger.html/blob/master/public/js/actions/types.js#L64-L69) and updating the action.

ThomasCrevoisier:
By adding a type for the action `TOGGLE_BREAKPOINTS` ?

jasonLaster:
@ThomasCrevoisier you should be able to run `npm run flow` locally and double check. You're really close

jasonLaster:
Hey @ThomasCrevoisier I hope you don't mind, but I went in and squashed your commits and got flow to pass.

public/js/reducers/breakpoints.js
```diff
@@ -138,6 +134,26 @@ function update(state = State(), action: Action) {
   return state;
 }
 
+function toggleBreakpoints(state: any, action: any) {
```
ThomasCrevoisier: That is weird indeed. Unfortunately I don't see why the type inference doesn't work here 😞

jlongster: Where is the error? We really shouldn't digress on type coverage. It's already very poor.

jasonLaster:
The error is in the previous circle run. You can repro by removing the any.

jlongster:
I only saw one flow error and I fixed it by accessing it outside of the `reduce` function call. Seems like a bug in Flow I guess. But I moved the reducer handler back in to make it consistent with the rest of the file.

jasonLaster:
Great. Nice fix.

jlongster:
This looks good, thanks for all the work and patience! I'll merge it, and if we find any problems we can fix it as a follow-up.

PR #927, opened 2016-10-12
Fix pretty-printing & enable it

jlongster:
This is a simple implementation of pretty-printing.

I intentionally simplified some of the logic. We should take this in steps, because there are some very weird situations with pretty-printing: if you pretty-print a generated source that has a sourcemap, do you "replace" its sourcemap? That would make sense, but then what happens to the old "original" source? In the old debugger this is temporary, and un-pretty-printing it would restore the old sourcemap. Various things like that.

Right now this only allows pretty-printing an unsourcemapped source. If it has a sourcemap, it replaces it.

Because sourcemaps work, all breakpoint/stepping/etc related things should "just work".

### DIFF
--- a/public/js/utils/pretty-print.js
+++ b/public/js/utils/pretty-print.js
@@ -1,0 +1,36 @@
+const { getValue } = require("devtools-config");
+const { workerTask } = require("./utils");
+const { isJavaScript } = require("./source");
+const assert = require("./assert");
+
+let prettyPrintWorker = new Worker(
+  getValue("baseWorkerURL") + "pretty-print-worker.js"
+);
+
+function destroyWorker() {
+  prettyPrintWorker.terminate();
+  prettyPrintWorker = null;
+}
+
+const _prettyPrint = workerTask(prettyPrintWorker, "prettyPrint");
+
+async function prettyPrint({ source, sourceText, url }) {
+  const contentType = sourceText ? sourceText.contentType : null;
+  const indent = 2;
+
+  assert(
+    isJavaScript(source.url, contentType),
+    "Can't prettify non-javascript files."
+  );
+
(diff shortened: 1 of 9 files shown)


jlongster:
Note that there's no rush getting this in; I want to push out a debugger update first without this.

jasonLaster:
> if you pretty-print a generated source that has a sourcemap, do you "replace" its sourcemap?

agreed. Pretty printing generated sources as well as bundles are definitely difficult. I believe we disabled it in august as well. I believe the sources actor has a function for extending the a source map so that we can pass the inverted source map in and extend the original.

public/js/utils/pretty-print.js
```diff
@@ -0,0 +1,37 @@
+const { getValue } = require("../feature");
+const { workerTask } = require("./utils");
+const { isJavaScript } = require("./source");
+const assert = require("./assert");
+
+let prettyPrintWorker;
+prettyPrintWorker = new Worker(
```
jasonLaster: could this be `let prettyPrintWorker = new Worker(`

public/js/utils/pretty-print.js
```diff
@@ -0,0 +1,37 @@
...
+    url,
+    indent,
+    source: sourceText.text
+  });
+}
+
+module.exports = {
+  prettyPrint,
+  destroyWorker
+};
```
jasonLaster: i like that this module cleans up the actions, as well as is consistent with source map

public/js/utils/pretty-print.js
jlongster: Yep, that was leftover from the ability to restart it (which we don't need here, it doesn't keep any internal state)

PR #930, opened 2016-10-13
Disable highlight for non-experimental WebAssembly format.

yurydelendik:
Based on the [bug 1308510](https://bugzilla.mozilla.org/show_bug.cgi?id=1308510)

We decide to use non-experimental WebAssembly format in devtools. However formal grammar (and highlight rules) are not defined for it yet. Proposing to disable the wasm mode in the devtools editor.


jlongster:
Thanks!

PR #931, opened 2016-10-13
Improve mono-repo workflow

jasonLaster:
Associated Issue: #493 

### Summary of Changes

+ improve preinstall
+ add docs
+ improve webpack build


bin/preinstall
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+if hash lerna 2>/dev/null; then
+  lerna bootstrap
+else
+  echo "Installing lerna. https://github.com/devtools-html/debugger.html/blob/master/docs/lerna.md"
+  npm i -g jasonlaster/lerna
```
jlongster: I don't think we should automatically install something globally on someone's machine. I think it'd be better to display a message telling the user to do `npm install -g jasonlaster/lerna`, particularly since this is a fork. I'm pretty they already have it installed this will overwrite the previous one.

docs/lerna.md
```diff
@@ -0,0 +1,17 @@
...
+Lerna does two things to help setup the Debugger. First it links the Debugger's internal packages, then it installs the external dependencies.
+During the linking phase, lerna figures out which packages depend on each other and installs symlinks in each packages' node_modules directory.
+This setup has several benefits:
+
+* Packages can easily require one another
+* It is easy to work on the entire project without worrying about re-installing different packages
+
+### Why is lerna forked?
+
+Lerna is forked so that the top-level package (debugger.html) can be linked with the sub-packages. This feature, will hopefully be merged into Lerna soon.
```
jlongster: nit: no need for a comma, just "This feature will..."

bin/preinstall
jasonLaster: hmm, true. maybe it would be fine if it were installed locally

jlongster:
Looks good, the main thing is the lerna installation process. This is a great step though, thanks!

jlongster:
Cool, looks good!

Issue #929, opened 2016-10-13
Add close button to search bar


jasonLaster:
The search bar is currently missing a close button, which would help people close the UI if they are more visual.

In terms of engineering:
- we have a [closeButton](https://github.com/devtools-html/debugger.html/blob/master/public/js/components/CloseButton.js) component here
- and the searchBar is defined [here](https://github.com/devtools-html/debugger.html/blob/master/public/js/components/EditorSearchBar.js).

Here are some screenshots:

![screen shot 2016-10-13 at 8 06 38 am](https://cloud.githubusercontent.com/assets/254562/19348424/d5cc9f10-911b-11e6-9ed9-d97aec872a48.png)

![screen shot 2016-10-13 at 8 07 03 am](https://cloud.githubusercontent.com/assets/254562/19348425/d5cfa9a8-911b-11e6-893c-a13ab4cc173a.png)

samuelfullerthomas:
Working on this!

jasonLaster:
Thanks!

PR #935, opened 2016-10-15
Adding close button to code search bar

akhillb:
Associated Issue: #929 
### Summary of Changes
- Adds close button to search bar

### DIFF
--- a/public/js/components/EditorSearchBar.js
+++ b/public/js/components/EditorSearchBar.js
@@ -5,6 +5,7 @@ const Svg = require("./utils/Svg");
 const { isEnabled } = require("../feature");
 const { find, findNext, findPrev } = require("../utils/source-search");
 const classnames = require("classnames");
+const CloseButton = require("./CloseButton");
 
 require("./EditorSearchBar.css");
 
@@ -58,11 +59,16 @@ const EditorSearchBar = React.createClass({
     }
   },
 
-  onEscape(shortcut, e) {
+  closeSearch(ev) {
     if (this.state.enabled) {
       this.setState({ enabled: false });
-      e.preventDefault();
+      ev.stopPropagation();
+      ev.preventDefault();
     }
+  },
+
+  onEscape(shortcut, e) {
+    this.closeSearch(e);
   },
 
   toggleSearch(shortcut, e) {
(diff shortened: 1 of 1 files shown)


jasonLaster:
Mind sharing a screen shot? I'm thinking about the size of the close button

jasonLaster:
Hi @akhillb, I just took a look and this is off to a good start.

I found two things:
1) the branch needs a rebase. Afterwards, it'll be helpful to run `npm install`
2) I think the styling of the search bar could be improved. I took a stab at styling it here. The gist is that the close button should fit the search bar (width:16px, color: comment-alt *i think*). It might be helpful to add a close-button size prop, that way by default it could be size small, but in this case we pass down size large and it is styled appropriately.

akhillb:
@jasonLaster sorry i was busy yesterday. Thanks for the input. I will make these changes and get back to you with the new code in a day or two.

wldcordeiro:
Since the searchbar is a flex container you could apply an `align-self: center` to the button and get the vertical centering easy @akhillb.

I do like the idea of having some kind of size prop on the close button that corresponds to a CSS class for the different sizes. Personally I think the colors should stay as is to keep the visual consistency across the other close buttons but that isn't too big of a deal.

jasonLaster:
ping @akhillb - here are some updated designs with the close button

https://projects.invisionapp.com/share/AY8OEG3C6#/screens/117937554_Debugger

feel free to disregard the button for search all files.

jasonLaster:
ping @akhillb. What's your schedule like? I'd be happy to pick this up if you're busy.

wldcordeiro:
Just renamed to make it more clear which search bar. 😄 I'm working on the file search UX currently.

akhillb:
@jasonLaster I am on a vacation now. Will be back on this by tomorrow.

akhillb:
Setting the margin-right property would mess up the alignment of close button in the file search bar. So i was thinking maybe we could have two different classes for close button, one for file search and other for code search, instead of setting values by a function.

jasonLaster:
@akhillb that sounds good to me. I don't mind having a `large` and `small` class for the close button.

wldcordeiro:
@jasonLaster We could add a `buttonClass` prop on the `CloseButton` that could optionally take an additional class for when you override things.

jasonLaster:
Thanks for the help @akhillb. This work is being finished up in #1135

Issue #936, opened 2016-10-15
npm install fails


nesk:
`npm install` fails since the `devtools-*` packages have been [added to the dependencies](https://github.com/devtools-html/debugger.html/commit/4a0e823f8ea6ea493f6516dd7ba55e3e4412f946#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R55).

```shell
❯ npm install        
npm ERR! Darwin 15.6.0
npm ERR! argv "/usr/local/Cellar/node/6.8.1/bin/node" "/usr/local/bin/npm" "install"
npm ERR! node v6.8.1
npm ERR! npm  v3.10.8
npm ERR! code E404

npm ERR! 404 Registry returned 404 for GET on https://registry.npmjs.org/devtools-network-request
npm ERR! 404 
npm ERR! 404  'devtools-network-request' is not in the npm registry.
npm ERR! 404 You should bug the author to publish it (or use the name yourself!)
npm ERR! 404 It was specified as a dependency of 'debugger.html'
npm ERR! 404 
npm ERR! 404 Note that you can also install from a
npm ERR! 404 tarball, folder, http url, or git url.

npm ERR! Please include the following file with any support request:
npm ERR!     ./debugger.html/npm-debug.log
```

jasonLaster:
Thanks for the report @nesk. 

The issue is likely a question of whether lerna is installed correctly. 

Could you check:
1. Is your branch up to date? 
2. If lerna is installed

nesk:
My branch is up to date. I just tried by wiping the project off and ran these commands:

```shell
❯ npm install --global lerna@prerelease
❯ npm install
```

And I get the same error, I event tried running `lerna bootstrap` before running `npm install` bu it still doesn't work.

jasonLaster:
Thanks. We're actually using a forked version jasonlaster/lerna. 

I noticed yesterday that there's another issue with the toolbox's webpack.config. On line 55, I believe, it should require the debuggers webpack config ../../../webpack.config

Pomax:
note that the error mentions that https://www.npmjs.com/package/devtools-network-request does not exist, which is true - nothing was published to npm with that name (yet?)

wldcordeiro:
@Pomax that isn't directly related. Since we use lerna it resolves these inner packages. and "installs" them, I think the idea is to publish to npm when we get to a more stable state and there isn't as much interdependency.

wldcordeiro:
[...]

PR #938, opened 2016-10-17
Improve the detection of lerna and make sure we install the local fork

wldcordeiro:
Associated Issue: #936 

### Summary of Changes

* change the preinstall script's lerna finding behavior and install @jasonLaster's fork in either case

### Testing

* [x] passes `npm test`
* [x] passes `npm run lint`

### DIFF
--- a/bin/preinstall
+++ b/bin/preinstall
@@ -1,9 +1,13 @@
 #!/usr/bin/env bash
 
-if hash lerna 2>/dev/null; then
-  lerna bootstrap
-else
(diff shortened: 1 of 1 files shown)


wldcordeiro:
@jasonLaster @jlongster so from reading through the npm issue (https://github.com/npm/npm/issues/10379) it seems like in npm 3.x `pre/post/install` always run after a dep-tree resolution step, which is where we currently fail with the 404s. This doesn't really address that specific issue but it does do a better job of finding lerna and makes sure to install the forked version. Until the changes get merged into lerna upstream which we can then just modify the script to use the global or local if the version is high enough.

PR #939, opened 2016-10-18
Make development server work for node v5

helenvholmes:
Associated Issue: no known associated issue number 🙃

### Summary of Changes

* Changes the ```const { foo } = require(..)``` syntax to ```const foo = require(...).foo``` in ```development-server.js```

### Testing

* [ ] passes `npm test` — note: not passing, because I'm on node v5:

```bash
> debugger.html@0.0.5 test /Users/hholmes/code/debugger.html
> node public/js/test/node-unit-tests.js

/Users/hholmes/code/debugger.html/public/js/test/node-unit-tests.js:11
const { registerConfig } = require("../../../packages/devtools-config/registerConfig");
      ^

SyntaxError: Unexpected token {
```

* [x] passes `npm run lint`

### Screenshots/Videos: N/A

### DIFF
--- a/packages/devtools-local-toolbox/bin/development-server.js
+++ b/packages/devtools-local-toolbox/bin/development-server.js
@@ -24,8 +24,9 @@ checkNode(">=5.0.0", function(_, opts) {
   }
 });
 
-const { getValue, isDevelopment } = require("devtools-config");
-const {registerConfig} = require("../../devtools-config/registerConfig");
+const getValue = require("devtools-config").getValue;
+const isDevelopment = require("devtools-config").isDevelopment;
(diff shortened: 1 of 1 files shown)


wldcordeiro:
Weird, so this is stuff that isn't run through Babel? If we're saying our requirement is Node 5.x for development, than any code that doesn't run through Babel should be compatible with Node 5.x. Nice find @helenvholmes!

helenvholmes:
@wldcordeiro According to James, the first file it's required in doesn't get the babel treatment... weirdly. Which is why he has a run script that just requires babel.

wldcordeiro:
@helenvholmes That's really odd stuff, cool that you got a fix for it though. 😄

PR #950, opened 2016-10-19
Fix scopes

jasonLaster:
This fixes an issue I was seeing expanding the scope pane. It also adds some tests to ensure that the Scopes and object properties load correctly. We likely want some other integration style tests where [getScopes](https://github.com/devtools-html/debugger.html/blob/master/public/js/components/Scopes.js#L75) is concerned.

### DIFF
--- a/packages/devtools-local-toolbox/webpack.config.devtools.js
+++ b/packages/devtools-local-toolbox/webpack.config.devtools.js
@@ -21,15 +21,22 @@ const nativeMapping = {
 };
 
 let packagesPath = path.join(__dirname, "../");
+let projectPath = path.join(__dirname, "../../");
 
 module.exports = webpackConfig => {
   webpackConfig.output.library = "Debugger";
 
   if (process.env.MOCHITESTS) {
-    webpackConfig.output.path = path.join(__dirname, "firefox/devtools/client/debugger/new");
+    webpackConfig.output.path = path.join(
+      projectPath,
+      "firefox/devtools/client/debugger/new"
+    );
   }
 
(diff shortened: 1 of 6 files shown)


packages/devtools-local-toolbox/webpack.config.devtools.js
```diff
@@ -21,15 +21,22 @@ const nativeMapping = {
...
 
 module.exports = webpackConfig => {
   webpackConfig.output.library = "Debugger";
 
   if (process.env.MOCHITESTS) {
-    webpackConfig.output.path = path.join(__dirname, "firefox/devtools/client/debugger/new");
+    webpackConfig.output.path = path.join(
+      projectPath,
+      "firefox/devtools/client/debugger/new"
+    );
```
jasonLaster: fixes an issue where `npm run mochitest-watch" was placing the build in *devtools-local-toolbox/firefox*

public/js/test/mochitest/browser_dbg-scopes.js
```diff
@@ -0,0 +1,27 @@
...
+
+  toggleScopes(dbg);
+
+  invokeInTab("firstCall");
+  yield waitForPaused(dbg);
+
+  toggleNode(dbg, 1);
+  toggleNode(dbg, 2);
+
+  yield waitForDispatch(dbg, "LOAD_OBJECT_PROPERTIES");
```
jasonLaster: One challenge is knowing whether toggling a node will trigger an object load. For the time being this is something the tester has to add, which I think is fine. At some point it might be nice to call `loadObjectProperties` all the time and return properties when they exist in the store. This way the tests could be a little more certain.

PR #953, opened 2016-10-19
move source search out of app into a separate component [WIP]

clarkbw:
Associated Issue: none

### Summary of Changes

* Move Source Search to its own component
  * I'd appreciate any review on this initial part
* [in progress] I'm working on saving the last thing you searched for in the input
  * For example, you search for `inc` and select the first item, it turns out to be wrong, so you search again but now you have to enter your search terms all over again.  I would like to re-run the last search but leave the text selected so you can easily start typing something else.
* [in progress] allow for tabbing to the next item when the search results are open

### Testing

* [x] passes `npm test`
* [x] passes `npm run lint`

Not yet.

### DIFF
--- a/public/js/components/App.js
+++ b/public/js/components/App.js
@@ -7,8 +7,6 @@ const classnames = require("classnames");
 const actions = require("../actions");
 const { isFirefoxPanel } = require("devtools-config");
 const { getSources, getSelectedSource } = require("../selectors");
-const { endTruncateStr } = require("../utils/utils");
-const { parse: parseURL } = require("url");
 
 const { KeyShortcuts } = require("devtools-sham-modules");
 const shortcuts = new KeyShortcuts({ window });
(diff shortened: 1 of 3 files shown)


jasonLaster:
@clarkbw Would it be alright if we merge this early so that this does not conflict too much with @helenvholmes UI work on #338? 

I think this change makes a lot of sense. 

By the way, could you test hitting enter when there are no search results. I think this could be broken.

clarkbw:
> @clarkbw Would it be alright if we merge this early so that this does not conflict too much with @helenvholmes UI work on #338?

Certainly, I can followup in another PR.

> By the way, could you test hitting enter when there are no search results. I think this could be broken.

I'll take a look

jasonLaster:
:thumbsup:

clarkbw:
@jasonLaster I'm not sure if I know what the behaviour is supposed to be for hitting enter when there are no search results.

Hitting enter after not typing anything doesn't do anything.

screen shot 2016-10-19 at 11 54 24 pm

Also hitting enter for a query that matches nothing doesn't do anything.  What is the desired behaviour here?

screen shot 2016-10-19 at 11 54 35 pm

Issue #954, opened 2016-10-19
getConfig() not returning anything for tests


jacobjzhang:
This is probably related to https://github.com/devtools-html/debugger.html/pull/946, where we had issues with registerConfig.

It appears that there are multiple places in the codebase where getConfig() is used to generate a file path and erroneously returns undefined. Please see screenshot.

I'm unsure if the fix requires a deeper dive into getConfig() or whether simply redoing this link in node-unit-tests.js would suffice.
![screen shot 2016-10-19 at 6 17 26 pm](https://cloud.githubusercontent.com/assets/6591285/19539699/d0fa196a-9628-11e6-92ec-7de25167eae7.png)

jasonLaster:
Thanks. My guess is this is connected with fresh git clones.

jasonLaster:
should be resolved

PR #946, opened 2016-10-18
replaced registerConfig call

jacobjzhang:
Associated Issue: #<issue number>

### Summary of Changes

* Replaced registerConfig

### Testing

* [x] passes `npm test`
* [x] passes `npm run lint`

PR #957, opened 2016-10-20
Fix fresh NPM install

jasonLaster:
We were seeing a couple of issues recently related git cloning and installing the project. #936 #954 #946

This fixes (i hope) the last issue, where `npm install devtools-local-toolbox` failed due to bin not being exported.

Here's a quick summary for those interested. NPM 3 runs the top level lifecycle hooks after installing dependencies [here](https://github.com/npm/npm/blob/master/lib/install.js#L263). This requires us to a) publish packages, b) have these packages be installable. 

For the record, yarn runs preinstall before installing dependencies [here](https://github.com/yarnpkg/yarn/blob/master/src/cli/commands/install.js#L709). Yarn's code is also significantly easier to follow if you ask me :)

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
     "co": "=4.6.0",
     "css-loader": "^0.25.0",
     "devtools-config": "^0.0.3",
-    "devtools-local-toolbox": "^0.0.3",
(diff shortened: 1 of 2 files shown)


jlongster:
Cool, looks good!

Issue #949, opened 2016-10-19
Turn Off Pretty Print Button if Source Already Has Sourcemap


helenvholmes:
## Steps to reproduce an example:
1. Open `debugger.html/public/js/test/mochitest/examples/doc-sourcemaps.html` in `debugger.html` directory
2. Select `bundle.js`
3. Hit prettyprint button

This will create a new formatted bundle.js, even though it already has a sourcemap. The old sourcemap gets destroyed. Breakpoints in the new sourcemap get real weird.

How do we want to handle this use case? My first thought it that if we detect that something already has a sourcemap, we remove the prettyprint button from the UI.

jasonLaster:
Thanks @helenvholmes. I believe the consensus is that we should be hiding the pretty print button for bundle files (sources that have an associated source map). Thanks for the really clear issue.

jasonLaster:
This works now

PR #959, opened 2016-10-20
Improve editor layout

jasonLaster:
Associated Issue: #891, #949

### Summary of Changes

- The footer no longer overlays the editor
- The footer is hidden when there is nothing to show
- Remove eroneous things
- Don't let users pretty print pretty files or bundle files

### DIFF
--- a/public/js/utils/editor.js
+++ b/public/js/utils/editor.js
@@ -1,0 +1,28 @@
+const { isPretty } = require("./source");
+const { isEnabled } = require("devtools-config");
+const { isOriginalId } = require("../utils/source-map");
+
+function shouldShowPrettyPrint(selectedSource) {
+  if (!isEnabled("prettyPrint")) {
+    return false;
+  }
+
+  const _isPretty = isPretty(selectedSource);
+  const isOriginal = isOriginalId(selectedSource.id);
+  const hasSourceMap = selectedSource.sourceMapURL;
+
+  if (_isPretty || isOriginal || hasSourceMap) {
+    return false;
+  }
+
+  return true;
+}
+
+function shouldShowFooter(selectedSource) {
+  return shouldShowPrettyPrint(selectedSource);
+}
+
+module.exports = {
+  shouldShowPrettyPrint,
+  shouldShowFooter
+};
--- a/public/js/components/Editor.css
+++ b/public/js/components/Editor.css
@@ -44,13 +44,17 @@ html[dir="rtl"] .editor-mount {
 }
 
 .search-bar ~ .editor-mount {
-  height: calc(100% - 58px);
+  height: calc(100% - 56px);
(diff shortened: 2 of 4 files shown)


wldcordeiro:
I'm closing my PR in favor of this. CodeMirror has a lot of layout quirks that make it a terrible fit for flexbox. 🙁

jasonLaster:
@wldcordeiro  haha - sorry, i should have clarified. I was still hoping that you would come up w. a great flexbox fix, but didn't want to block on it :) 

Thanks for the hard work there

public/js/utils/editor.js
```diff
@@ -0,0 +1,32 @@
+const { isPretty } = require("./source");
```
jlongster: What' the point of creating another file? I don't see why we should be splitting out components into multiple files between `utils` and `components`. It's a lot hard to follow the codebase.

public/js/utils/source-map-util.js
```diff
@@ -10,7 +10,7 @@ function generatedToOriginalId(generatedId, url) {
 }
 
 function isOriginalId(id) {
-  return id.match(/\/originalSource/);
+  return id && id.match(/\/originalSource/);
```
jlongster: id's should definitely always exist. I'd rather not make things null-compatible; why would be pass in null? If you do `source.id` and it doesn't have an id that's a big problem. (this is also the reason why I don't like lodash. doing this makes bad code be silent)

jlongster:
Sweet, thanks for all these fixes. A few nits that I'd like to discuss though.

public/js/utils/editor.js
jasonLaster: unfortunately, both the footer and editor need to use this utility.

public/js/utils/source-map-util.js
jasonLaster: true - i'll look into this

jlongster:
This PR still has problems:

1. The footer scrolls with the source. It's no longer "pinned" at the bottom. You should be able to repro by just loading a large source and scrolling.

screen shot 2016-10-24 at 10 03 47 am

2. The editor no longer is always 100% height. If the contents don't fill the screen, the editor is too short. It might be hard to tell, but you can see that the background color below the editor is gray, not white:

screen shot 2016-10-24 at 10 05 08 am

I just pulled the latest branch, I think, but if you can't repro maybe I did something wrong.

We need to fix this layout ASAP. There are other serious problems with the latest debugger in nightly, see https://bugzilla.mozilla.org/show_bug.cgi?id=1310330. It seems to be related to the functionality of "hiding" tabs when there isn't enough room, and it seems completely broken. The menu for listing hidden tabs is gone, and when it hides tabs, the entire area becomes scrollable.

PR #972, opened 2016-10-20
add flow typing to sources actions

clarkbw:
Associated Issue: none

### Summary of Changes

* moved `ThunkArgs` into `actions/types` and typed the sources action
* removed the unused `blackbox` function from the sources action module

### Testing

* [x] passes `npm test`
* [x] passes `npm run lint`

### DIFF
--- a/public/js/actions/sources.js
+++ b/public/js/actions/sources.js
@@ -1,3 +1,5 @@
+// @flow
+
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
@@ -9,7 +11,6 @@
 
 const defer = require("../utils/defer");
 const { PROMISE } = require("../utils/redux/middleware/promise");
-const { Task } = require("../utils/task");
 const assert = require("../utils/assert");
 const { updateFrameLocations } = require("../utils/pause");
 const {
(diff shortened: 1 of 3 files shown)


public/js/actions/sources.js
```diff
@@ -337,7 +316,7 @@ function getTextForSources(actors) {
       if (pending.size == 0) {
         // Sort the fetched sources alphabetically by their url.
         deferred.resolve(
-          fetched.sort(([aFirst], [aSecond]) => aFirst > aSecond));
+          fetched.sort(([aFirst], [aSecond]) => aFirst > aSecond ? -1 : 1));
```
jasonLaster: we'll likely want to revisit this when we add full project search, but that is a little ways off.

public/js/actions/sources.js
```diff
@@ -284,11 +262,12 @@ const FETCH_SOURCE_RESPONSE_DELAY = 200;
  * @returns {Promise}
  *         A promise that is resolved after source texts have been fetched.
  */
-function getTextForSources(actors) {
-  return ({ dispatch, getState }) => {
+function getTextForSources(actors: any[]) {
```
jasonLaster: we'll improve the types here when we work on project search

public/js/actions/sources.js
```diff
@@ -148,36 +156,6 @@ function closeTab(id) {
 }
 
 /**
- * Set the black boxed status of the given source.
```
jasonLaster: I removed the early support for blackbox in the source footer last week as it was not being used.

Issue #967, opened 2016-10-20
Code Search - Show "no results" label and sad face


jasonLaster:
Currently, we just turn the text orange when no results are found, but we should do a couple other things as well:

1. show a "no results" summary
2. replace the search button with a sad face (i'm very excited about this)

Here's the sad face: https://cl.ly/2a3r0b2r0B3g

Here's the code to change:

```diff
diff --git a/public/images/Svg.js b/public/images/Svg.js
index 6c30147..db842d0 100644
--- a/public/images/Svg.js
+++ b/public/images/Svg.js
@@ -22,7 +22,8 @@ const svg = {
   "stepOut": require("./stepOut.svg"),
   "stepOver": require("./stepOver.svg"),
   "subSettings": require("./subSettings.svg"),
-  "worker": require("./worker.svg")
+  "worker": require("./worker.svg"),
+  "sad-face": require("./sad-face.svg") // we'll need to add the svg in public/images... the link is in the issue
 };
 
 module.exports = function(name, props) { // eslint-disable-line
diff --git a/public/js/components/EditorSearchBar.js b/public/js/components/EditorSearchBar.js
index cecba60..81a2162 100644
--- a/public/js/components/EditorSearchBar.js
+++ b/public/js/components/EditorSearchBar.js
@@ -115,7 +115,11 @@ const EditorSearchBar = React.createClass({
     const { count, index, query } = this.state;
 
     if (count == 0 || query.trim() == "") {
-      return dom.div({});
+
+      return dom.div(
+        { className: "summary" },
+        "no results"
+      );
     }
 
     return dom.div(
@@ -133,7 +137,7 @@ const EditorSearchBar = React.createClass({
 
     return dom.div(
       { className: "search-bar" },
-      Svg("magnifying-glass"),
+      count == 0 ? Svg("sad-face") : Svg("magnifying-glass"),
       dom.input({
         className: classnames({
           empty: count == 0
```

PR #977, opened 2016-10-22
Show "no results" label and sad face for no results

akhillb:
Associated Issue: #967 

### Summary of Changes

* When code search returns 0 results, this PR shows a "no results" summary and changes from magnifying glass to sad smiley.
* Also I added an extra check for query, to see if it's empty. If it is empty, we don't show a sad smiley. Would like to hear your thoughts on this, since this was not mentioned in the issue. Hope you guys don't mind about this :)

### Testing

* [ x] passes `npm test`
* [ x] passes `npm run lint`

screen shot 2016-10-22 at 20 35 44

screen shot 2016-10-22 at 20 34 47

screen shot 2016-10-22 at 20 34 54

### DIFF
--- a/public/images/Svg.js
+++ b/public/images/Svg.js
@@ -22,7 +22,8 @@ const svg = {
   "stepOut": require("./stepOut.svg"),
   "stepOver": require("./stepOver.svg"),
   "subSettings": require("./subSettings.svg"),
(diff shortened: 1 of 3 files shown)


jasonLaster:
I'm not sure why the integration tests failed, but it doesn't look related. This should land soon.

PR #979, opened 2016-10-23
Flow support

jasonLaster:
### Summary of Changes

I'd like to get us to a place where wee can create small flow related "up for grabs" issues. This work helps us in a couple of ways:

1) it provides a flow coverage tool so that we can assess our coverage
2) moves general flow types to `public/js` and keeps action types in `actions/js`
3) moves tcomb types from `types.js` to `tcomb-types.js`. we should kill them soon.
4) updates the npm packages because the published toolbox package was pointing to the old name of `chrome-remote-debugging-protocol`. 

**next steps**

* I'd like to be able to publish our flow coverage on github pages, but to do that we need to be able to auto-generate it on build which seems possible if you use jekyll, but not otherwise... 

### Screenshots/Videos 

Go to [flow-report](http://jasonlaster.github.io/debugger-flow-report/)

##### HTML Report
screen shot 2016-10-23 at 12 48 57 pm

##### Terminal Report
screen shot 2016-10-23 at 12 50 50 pm


public/js/actions/types.js
```diff
@@ -17,8 +17,8 @@ export type { Source };
   * @typedef {Object} ThunkArgs
   */
 export type ThunkArgs = {
-  dispatch: any,
-  getState: any,
+  dispatch: () => Promise<any>,
+  getState: () => any,
```
jasonLaster: I started some work on making this more explicit, but it needs some tidying up... 
https://github.com/devtools-html/debugger.html/compare/master...jasonLaster:flow-support2?expand=1

anyways, it should be viewed independently of this PR.

yarn.lock
```diff
@@ -5011,6 +5028,14 @@ react@=0.14.7:
     envify "^3.0.0"
     fbjs "^0.6.1"
 
+react@15.3.1:
```
jlongster: Why is it pulling in react 15.3.1 now?

jlongster:
Cool, I glanced through it and didn't see anything alarming except for my one comment. Is React 15 being pulled in somehow?

yarn.lock
jasonLaster: yeah, that's interesting. I just looked an it seems like it's installing both 14.7 and 15.3. It installs 15.3 so that flow-coverage-reporter can use it. I believe this is okay.

jasonLaster: Also - and this is important - `node_modules/react/dist/react.js` -> 0.14.7

jlongster: Yeah, that's how it should work. Thanks for looking into this.

PR #980, opened 2016-10-24
Fix footr2

jasonLaster:
Fixes the footer


jlongster:
Awesome, this does seem to work great, thanks! And it fixes the issue described in the bugzilla bug as well about the hidden tabs messing things up. The hidden tabs still seem to be wonky, but at least the layout is more stable.

PR #981, opened 2016-10-24
Fixes layout issues with the sources dropdown

jasonLaster:
Fixes some of the styling issues we've seen with the sources dropdown. @wldcordeiro this should address the evil three dots and other related things.

This also pulls in the great work @jacobjzhang did with refactoring the dropdowns.

Some of the styling issues were with:
* the center pane scrolling when a source tab was too big for the header
* a source tab not shrinking to the width of the header
* the subsetting button not adjusting to the width of the header
* the dropdown font height, and text overflowing... this is still not a great dropdown component (even though it is now extracted) there are clipping issues in narrow widths and in general we should probably consider a more mature library, but it gets the job done for now.

### DIFF
--- a/public/js/components/App.css
+++ b/public/js/components/App.css
@@ -37,6 +37,7 @@ body {
   position: relative;
   flex: 1;
   background-color: var(--theme-tab-toolbar-background);
+  overflow: hidden;
 }
 
 .editor-container {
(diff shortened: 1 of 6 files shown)


jlongster:
Just tried it locally and it behaves very well. Thanks!

PR #982, opened 2016-10-24
Small flow tweaks

jasonLaster:
Small tweaks to help the flow score. Also removes spurious AppState calls

### DIFF
--- a/public/js/reducers/pause.js
+++ b/public/js/reducers/pause.js
@@ -107,35 +107,35 @@ function update(state = initialState, action, emit) {
   return state;
 }
 
-function getPause(state: AppState) {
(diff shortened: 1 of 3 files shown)


public/js/reducers/pause.js
```diff
@@ -107,35 +107,35 @@ function update(state = initialState, action, emit) {
   return state;
```
clarkbw: I don't think this file has the `// @flow` at the top of it so I don't think flow was looking in here.

jasonLaster: yup. exactly. this was copy pasta from the selectors file awhile bak

Issue #941, opened 2016-10-18
Toggle Breakpoint - disabling a breakpoint fails


jasonLaster:
Toggling all the breakpoints currently causes other breakpoint actions like removing a breakpoint to fail.

Steps to Reproduce:
* add a couple breakpoints
* toggle all breakpoints
* toggle all breakpoints
* toggle a breakpoint or remove a breakpoint

 I believe this happens because those actions are using the wrong breakpoint actor.

![](http://g.recordit.co/iGfBeeDMDV.gif)

jasonLaster:
@ThomasCrevoisier do you have some time to look into this? If not, I'd be happy to pick this up

ThomasCrevoisier:
@jasonLaster Unfortunately, I won't have the time to look into this this week :/
Have you got specific error messages ? Is this targetting Firefox debug ?

Normally we update the actor when we re-enable breakpoints : https://github.com/devtools-html/debugger.html/pull/926/files#diff-db5189725270fa6c573d869b7a7911acR125

jlongster:
This should be fixed right?

jasonLaster:
yup

PR #984, opened 2016-10-24
Fix toggling breakpoints

jasonLaster:
Associated Issue: #941

### Summary of Changes

* This reverts #926. I feel bad reverting this good work, but I don't think we have the time to guarantee this behavior before next Tuesday's release. I spent a lot of time on thursday + friday working on this behavior and while I was able to mostly get it working [here](https://github.com/devtools-html/debugger.html/compare/master...jasonLaster:fix-toggle-breakpoints?expand=1), it added a lot of complexity and was still buggy. Here's a writeup of the cases it should [handle](https://github.com/jasonLaster/debugger.html/blob/e351a604d467ea7d3072766a04f37b91a3189e4d/packages/devtools-local-toolbox/public/js/clients/firefox/breakpointCommands.js). I think we can revisit this after launching and think through the UX that we want. This way we can do it well and not be rushed

* Adds tests for the existing breakpoints behavior (disabling, removing a breakpoint) (toggling all breakpoints)


.gitignore
```diff
@@ -3,7 +3,6 @@ build
 node_modules
 config/local.json
 npm-debug.log
-lerna-debug.log
```
jasonLaster: lerna-debug was mentioned twice in this file

jlongster:
Sounds good, thanks!

Issue #370, opened 2016-07-20
Polish Dark and Firebug themes


jasonLaster:
We now support three themes (dark, firebug, light). As a result we need to convert all of our styling to css variables so that the themes always look good.

For example, the [autocomplete](https://github.com/jlongster/debugger.html/blob/master/public/js/components/Autocomplete.css#L14) stylesheet uses raw values like `#dde1e4`, which should be `var(--theme-gray)`. All values can be found in [variables.css](https://github.com/jlongster/debugger.html/blob/master/public/js/lib/variables.css#L86).

Extra credit if we update our stylelint to check for colors and enforce variables... after briefly looking i couldn't find a rule

#### Firebug
![screen shot 2016-07-20 at 1 15 54 pm](https://cloud.githubusercontent.com/assets/254562/16995740/fadeceb4-4e7b-11e6-8010-c01b9e0d8a96.png)

### Dark theme
![screen shot 2016-07-20 at 1 15 36 pm](https://cloud.githubusercontent.com/assets/254562/16995739/fadcc8a8-4e7b-11e6-8492-7f1ce48eecc8.png)

jasonLaster:
Pulling it out of the sprint as we're not going to be able to get to it in time.

clarkbw:
related bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1295321

wldcordeiro:
If someone can mentor me on what is where, I'm down to do this. 👍

jasonLaster:
Hey @wldcordeiro would be happy to help!

wldcordeiro:
@jasonLaster adjust the font sizes? Are the current values not using the theme vars? This is for the right and left sidebar?

jlongster:
This bug should probably be moved to next sprint. Right now we're just making sure that the basic dark/firebug themes even work. They were totally broken before but I just spent a few hours getting them basically working. How does moving this out (as a polish bug) sound?

jasonLaster:
sounds good to me

jasonLaster:
Closing this issue and re-opening a new more relevant issue #988.

PR #989, opened 2016-10-25
Clean up themes

### LINKED ISSUE #934: Don't conflict both of class theme-dark and theme-light in new Debugger
In theme-dark,  class theme-light is available.
Don't conflict both of class theme-dark and theme-light in a new Debugger.

do-not-conflict-both-of-class-theme-dark-and theme-light-in-new-debugger

jasonLaster:
Fixes #934
Associated Issue: #370 

### Summary of Changes

* removes redundant class
* makes it easy to change themes locally

screen shot 2016-10-25 at 11 31 05 am
screen shot 2016-10-25 at 11 29 29 am

### DIFF
--- a/packages/devtools-local-toolbox/public/js/index.js
+++ b/packages/devtools-local-toolbox/public/js/index.js
@@ -2,10 +2,8 @@
 
 const { bindActionCreators, combineReducers } = require("redux");
 const { Provider } = require("react-redux");
-const ReactDOM = require("react-dom");
-const React = require("react");
 
-const {DevToolsUtils, AppConstants} = require("devtools-sham-modules");
+const { DevToolsUtils, AppConstants } = require("devtools-sham-modules");
 const { injectGlobals, debugGlobal } = require("./utils/debug");
 const { setConfig, isEnabled, getValue, isDevelopment } = require("devtools-config");
 
@@ -46,6 +44,16 @@ function initApp() {
 
 function renderRoot(_React, _ReactDOM, component, _store) {
   const mount = document.querySelector("#mount");
+
+  // Using this static variable allows webpack to know at compile-time
+  // to avoid this require and not include it at all in the output.
+  if (process.env.TARGET !== "firefox-panel") {
+    const theme = getValue("theme");
+    theme == "dark" ? require("./lib/themes/dark-theme.css")
+                    : require("./lib/themes/light-theme.css");
+
+    document.body.parentNode.classList.add(`theme-${theme}`);
+  }
 
   // bail in test environments that do not have a mount
   if (!mount) {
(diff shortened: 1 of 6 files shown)


packages/devtools-local-toolbox/public/js/index.js
```diff
@@ -47,6 +45,16 @@ function initApp() {
...
   const mount = document.querySelector("#mount");
 
+  // Using this static variable allows webpack to know at compile-time
+  // to avoid this require and not include it at all in the output.
+  if (process.env.TARGET !== "firefox-panel") {
+    const theme = getValue("theme");
+    theme == "dark" ? require("./lib/themes/dark-theme.css")
+                    : require("./lib/themes/light-theme.css");
+
+    document.body.parentNode.classList.add(`theme-${theme}`);
```
jasonLaster: The dark theme requires adding a class to the `<html>` element. This patch adds the class to both the html element and the root, which is just a way of being consistent in both cases

public/js/components/App.js
```diff
@@ -58,8 +56,7 @@ const App = React.createClass({
 
   render: function() {
     return dom.div(
-      { className: classnames("debugger theme-body",
-                              { "theme-light": !isFirefoxPanel() }) },
+      { className: "debugger" },
```
jasonLaster: I pulled theme-body up to the Root element to be more consistent for other tools that use the toolbox, but I know we'll still want to figureout the toolbox / debugger api.

packages/devtools-local-toolbox/public/js/index.js
```diff
@@ -47,6 +45,16 @@ function initApp() {
 function renderRoot(_React, _ReactDOM, component, _store) {
   const mount = document.querySelector("#mount");
 
+  // Using this static variable allows webpack to know at compile-time
+  // to avoid this require and not include it at all in the output.
+  if (process.env.TARGET !== "firefox-panel") {
+    const theme = getValue("theme");
+    theme == "dark" ? require("./lib/themes/dark-theme.css")
```
wldcordeiro: Is this doable?

```js
const theme = `./lib/themes/${getValue("theme")}-theme.css`
require(theme);
```

jasonLaster: nope. webpack doesn't like expressions and neither do I :)

packages/devtools-local-toolbox/public/js/index.js
jlongster: Cool

jlongster:
All of this makes sense, and doesn't look like a pile of hacks :) This makes us more consistent with the toolbox, which is nice. As long as you're 100% sure it doesn't bundle in the theme files for the toolbox (which it doesn't look like it should), this looks great.

Issue #951, opened 2016-10-19
Tabs - New tab button


jasonLaster:
It would be great to have a new tab button that opens the file search UI.

The relevant files are:

1) [SourcesTab](https://github.com/devtools-html/debugger.html/blob/master/public/js/components/SourceTabs.js). We likely can add the plus button in the render method
2) [App.js](https://github.com/devtools-html/debugger.html/blob/master/public/js/components/App.js#L76-L79) - handles showing file search.

How do we show file search? The App currently listens for cmd+p events on the window. I think the easiest way to implement this feature would be to emit the cmd+p event on the window.

![](https://cloud.githubusercontent.com/assets/1720093/19536699/084d65a6-961b-11e6-807a-b84d2149c31c.png)

helenvholmes:
I think what @nt1m's designed is really natural—this is exactly what I would have designed since it's the pattern browsers have set up for spinning up new tabs.

To answer @clarkbw's concerns: We don't currently have anywhere else where we've set up UX patterns for this particular use case anywhere in the tools, so what we do here sets up the precent. This is the pattern set up by te browser itself, so I think it's safe to continue copying.

[This](https://cl.ly/0g0X3B2g402I) is the icon I'd use for it. I'd also use ```var(--theme-splitter-color)``` for its fill, with ```var(--theme-comment)``` as its hover state.

clarkbw:
@helenvholmes what would you like to do when the tabs overflow? Always keep space for the + icon at the right of the last tab?

I think the overflow icon is broken currently but we should take that into account.

wldcordeiro:
Taking this since @helenvholmes seems to have put it up for grabs. 😄

helenvholmes:
> @helenvholmes what would you like to do when the tabs overflow? Always keep space for the + icon at the right of the last tab?
> 
> I think the overflow icon is broken currently but we should take that into account.

I think once the tabs overflow we should make room for the tab overflow icon, and the "+" icon should disappear. Just added info for you, @wldcordeiro!

wldcordeiro:
@helenvholmes I had it so the new tab button is always present even with the overflow menu. That seems like more expected behavior to me as in browsers like Chrome/Firefox the new tab button is always present (though they reduce tab sizes to just icons and scroll the tab bar).

![screenshot_20161026_090148](https://cloud.githubusercontent.com/assets/580982/19731532/f3c1ca7e-9b5a-11e6-8861-d69ca98c0cbe.png)

[...]

PR #1007, opened 2016-10-26
New tab button

wldcordeiro:
Associated Issue: #951 

### Summary of Changes

* Add a new tab button that opens the file search
* move file search toggle function up into App.js so that it could be shared between `SourceTabs` and `SourceSearch`

### Testing

* [x] passes `npm test`
* [x] passes `npm run lint`

screenshot_20161025_213239
screenshot_20161025_213248
screenshot_20161025_213254
screenshot_20161025_213720


public/js/components/SourceSearch.js
```diff
@@ -61,27 +57,27 @@ const Search = React.createClass({
...
 
   render() {
-    return this.state.searchOn ?
+    return this.props.searchOn ?
       dom.div({ className: "search-container" },
       Autocomplete({
         selectItem: result => {
           this.props.selectSource(result.id);
-          this.setState({ searchOn: false });
+          this.props.toggleSearch(false);
```
jasonLaster: Hmm, I'm not thrilled about refactoring SourcesState so that the searchOn state is kept in the app. I can see the argument for it being app state, but at this point we have a nice convention for state being local to the component it is being used.

I think it would be nice to use DOM events to open the panel. This way, the panel API is largely DOM and doesn't rely on React.

Here is a function for opening the panel that could be use in the tab onClick.
```js
function openSearchPanel() {
  window.dispatchEvent(Object.assign(new Event("keydown"), {
    key: "p", metaKey: true, ctrlKey: false, altKey: false, shiftKey: false
  }))
}
```

public/js/components/SourceTabs.js
```diff
@@ -139,6 +140,13 @@ const SourceTabs = React.createClass({
       CloseButton({ handleClick: onClickClose }));
   },
 
+  renderNewButton() {
+    return dom.div({
+      className: "new-tab-btn",
+      onClick: () => this.props.openFileSearch(true)
```
jasonLaster: so this wasn't clear because of the dark theme tabs, but is it possible to have multiple blank tabs?

It would be nice if the tabs button were hidden during file search.

jasonLaster: ahh - good point. I guess the next best thing to do would be to make the button a no-op when file search is open.

public/js/components/SourceSearch.js
wldcordeiro: I'm not really a fan of having a function that will be bound to an `onClick` handler that then fires a `keydown` event. Maybe this is something that should live in a Redux reducer.

jasonLaster: Hmm, lets see what the reducer option looks like. We do want a UI state reducer at some point. 

I'm on the fence about the DOM event. I think i'd also feel differently if instead of it being a keydown it were clicking an element :)

jasonLaster:
@jlongster mind giving this a look as it introduces a new UI reducer.

public/js/actions/types.js
```diff
@@ -88,7 +87,7 @@ type SourceAction =
                sourceText: SourceText }}
   | { type: "CLOSE_TAB", id: string };
 
-type UIAction = { type: constants.TOGGLE_FILE_SEARCH, searchOn: boolean };
+type UIAction = { type: "TOGGLE_FILE_SEARCH", searchOn: boolean };
```
wldcordeiro: Yeah, good idea. This file could likely be updated to use the constants in its own PR.

public/js/actions/ui.js
```diff
@@ -0,0 +1,16 @@
+// @flow
+const constants = require("../constants");
+import type { ThunkArgs } from "./types";
+
+function toggleFileSearch(searchOn: boolean) {
+  return ({ dispatch }: ThunkArgs) => {
+    dispatch({
+      type: constants.TOGGLE_FILE_SEARCH,
+      searchOn
+    });
```
jlongster: You don't need to use the thunk middleware here, just return the action:

```
function toggleFileSearch(searchOn: boolean) {
  return {
      type: constants.TOGGLE_FILE_SEARCH,
      searchOn
  };
}

public/js/actions/types.js
jlongster: It doesn't need to use the constants. Once everything is flow typed the constants file can go away. Flow will error if you pass in a string that doesn't match any of these types.

jlongster:
Code looks great! I haven't tested this locally, so as long as it works, the code is fine. What happens when there are too many tabs and we start hiding them? etc Test all the edge cases manually and make sure nothing is broken.

jasonLaster:
@jlongster you're right that the small width UI challenges are the difficult edge case.

Here's a gif of it working

jasonLaster:
Thanks @wldcordeiro for this great work. I really appreciate how you jump in on the UI polish tickets.

Hope my small tweaks are alright.

wldcordeiro:
@jasonLaster looks like we may want to set a minimum width for the center pane? The search results break out of their container in an ugly way too.

jasonLaster:
@wldcordeiro that's a good idea. I'll file an issue for that.

clarkbw:
> I'll file an issue for that.

#1029 is handling the minimum width

PR #1009, opened 2016-10-26
Add documentation for tracking issues.

wldcordeiro:
Associated Issue: N/A

### Summary of Changes

* document the purpose of tracking issues and how they're useful to contributors.

### Testing

* [x] passes `npm test`
* [x] passes `npm run lint`

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -169,9 +169,12 @@ If you find an `up for grabs` issue without a difficulty level set as a label or
 
 If you're looking for a good issue, you can look through the `up-for-grabs` issues. These issues should be actionable and well documented.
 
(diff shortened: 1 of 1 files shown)


wldcordeiro:
Tracking issues are an often used way of breaking up larger tasks into tiny ones that encourage new contributions so I figure it would be a good idea to have a label for them and documentation.

jasonLaster:
Thanks. I'd be open to giving this a try. We're still trying to sort out the appropriate workflow w/ milestones, boards, labels, tracking issues. 

I definitely see the efficacy of tracking issues in cases like mochitests, but i don't want to overwhelm our issues with lots of small issues either. Often it is sufficient to have a tracking issue and a couple implementation issues that are good to start with. We could always add more implementation issues as we go.

jasonLaster:
after thinking this about more, i'm not convinced this will help us at the moment. feel free to continue discussing.

PR #1016, opened 2016-10-26
Optimize editor rendering by removing props arg in redux state mapping function

jlongster:
A few things are going on here. For some reason, redux always rerenders a connected component no matter what if you accept the `props` argument. Even if the state you return is the same as before, it'll rerender. We don't use the `props` arg here, so we can just remove it.

For some reason, in React 0.14.7, the editor was rerendering but `componentWillReceiveProps` wasn't being called. We do most of our heavy work in there, so it was skipping it. In React 15, that lifecycle method is always called, which seems like the correct behavior. I haven't been able to reduce this into a simple test case, but let's go ahead and clean this code up which I think will allow us to run against React 15. I'd like to understand this more at some point.

### DIFF
--- a/public/js/components/Editor.js
+++ b/public/js/components/Editor.js
@@ -363,7 +363,7 @@ const Editor = React.createClass({
 });
 
 module.exports = connect(
-  (state, props) => {
+  state => {
     const selectedLocation = getSelectedLocation(state);
     const sourceId = selectedLocation && selectedLocation.sourceId;
     const selectedSource = getSelectedSource(state);


jimbolla:
React Redux should only rerender if it sees different props. If you try v5 beta (`npm install react-redux@next`), you can figure out which props are being treated as different by adding a custom implementation of `areMergedPropsEqual` which would look something like this:

```js
// usage
connect(..., ..., ..., { areMergedPropsEqual: shallowEqualWithDebugging })(...)

// implementation

const hasOwn = Object.prototype.hasOwnProperty
function shallowEqualWithDebugging(a, b) {
  if (a === b) return true

  let countA = 0
  let countB = 0
  
  for (let key in a) {
    if (hasOwn.call(a, key) && a[key] !== b[key]) {
      console.log(key + ' has changed: ' + a[key] + ' to ' + b[key])
      return false
    }
    countA++
  }

  for (let key in b) {
    if (hasOwn.call(b, key)) countB++
  }

  return countA === countB
}
```

jlongster:
@jimbolla Thanks. I don't think it's rerendering, but the way react-router v4 worked was if the state function depends on props, it will render but return the previously rendered component (it stuck it as an instance variable). So my component's render function isn't called, but `componentWillReceiveProps` *still is*, which was the problem. We do a lot of work in cWRP so that was a problem.

This is a combination of a lot of things, but I bet upgrading to 5 would fix it. We need to do more work upgrading though because we need to upgrade our version of react-redux in Firefox for the devtools first which is a more complicated process.

Thanks for the information, I want to dig into this again soon to figure out exactly what's happening.

jimbolla:
Ah. That's weird that it would call cWRP but not render. I would not expect that. Yes v5 should behave differently because it avoids calling its own render method unless it the final props have actually changed, which means your component's cWRP shouldn't get called then.

public/js/components/Editor.js
```diff
@@ -363,7 +363,7 @@ const Editor = React.createClass({
 });
 
 module.exports = connect(
-  (state, props) => {
+  state => {
```
jlongster: It comes down to a bug in React 0.14.7, I believe. At some point I'd like to figure out exactly why it's not calling lifecycle methods the same, but at least it's not intentional behavior.

Issue #1012, opened 2016-10-26
Dark Theme - Breakpoint Line in Editor Tweaks


wldcordeiro:
The greens used look odd in dark theme and need to be tweaked

![screenshot_20161025_234437](https://cloud.githubusercontent.com/assets/580982/19714638/0a8af58a-9b0d-11e6-94be-f8cb11431adf.png)
![screenshot_20161025_234449](https://cloud.githubusercontent.com/assets/580982/19714639/0a9b2270-9b0d-11e6-98ab-99883dd0393d.png)

Issue #1010, opened 2016-10-26
Dark Theme - Breakpoint Tweaks


wldcordeiro:
The breakpoint line coloring and remove button need color tweaking

![screenshot_20161025_234148](https://cloud.githubusercontent.com/assets/580982/19714594/b5e2e9b6-9b0c-11e6-866e-aae57dab1765.png)
![screenshot_20161025_234214](https://cloud.githubusercontent.com/assets/580982/19714595/b5f9be20-9b0c-11e6-95b5-2335c8d70d8f.png)

helenvholmes:
Hey @wldcordeiro, I don't understand the remove button part of this issue—is that not the native checkbox element?

wldcordeiro:
I copied this from @jasonLaster's description so I'm not sure either. It might be the `x` for the close button is blending into the background

helenvholmes:
@wldcordeiro ohhhhh, okay, good show i'll make sure to make that change too

PR #1019, opened 2016-10-26
(dark theme+light theme) Fix breakpoint gutter color, breakpoint hover color, re-center close button

helenvholmes:
Associated Issue: #1010, also follow-up from #1012

### Summary of Changes
As a caveat to these changes: I'm not _in love_ with the hover states in the breakpoints pane. However, I still think they're improvements and I think the work necessary to make them 💫 great 🌟 requires m-c changes that should be done in follow-ups.

* Update ```--breakpoint-active-color-hover``` for dark theme to be ```--breakpoint-active-color``` with slightly more opacity
* Make the hover color for the breakpoints in the sidebar less intense, but _more_ intense in light theme
* Recenter the close button (before it was slightly too close to the top)

### Testing

* [x] passes `npm test`
* [x] passes `npm run lint`

### Screenshots/Videos
Before:

After:

Before (hover state):

After (hover state):

Before (hover state, light theme):

After (hover state, light theme):

### DIFF
--- a/public/js/components/Breakpoints.css
+++ b/public/js/components/Breakpoints.css
@@ -10,7 +10,7 @@
   line-height: 1em;
   position: relative;
   border-left: 4px solid transparent;
-
+  transition: all 0.25s ease;
 }
 
 .breakpoints-list .breakpoint:last-of-type {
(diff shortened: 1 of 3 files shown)


jlongster:
Yeah, looks good, thanks!

PR #1022, opened 2016-10-26
update ignores

jasonLaster:


jlongster:
Interesting, why was it passing before this?

atiraminor:
How do I use this app

wldcordeiro:
Comments like this one above have popped up in a few PRs and I can't help but think they're from bot accounts. They're never relevant to what they reply too. 😕

jlongster:
Yeah, I just ignore them :) @atiraminor if you are serious, please check our readme.

jasonLaster:
@jlongster this is not for our `npm run lint` it's for editors
https://github.com/devtools-html/debugger.html/blob/master/package.json#L18-L19

Issue #952, opened 2016-10-19
Use native L10N


jasonLaster:
We would like to be able to use the native L10N support when devtools is in the panel and default to a no-op when devtools is running locally. This is similar to the SourceEditor, NetworkRequest, Prefs, ...

jasonLaster:
I believe we still need to merge m-c https://bugzilla.mozilla.org/show_bug.cgi?id=1313196, but it's done on github

PR #1023, opened 2016-10-26
Add L10N support

jasonLaster:
Associated Issue: #952 

### Summary of Changes

* adds basic L10N support
* follows pattern set in [debugger-controller](https://dxr.mozilla.org/mozilla-central/source/devtools/client/debugger/debugger-controller.js#175)
* needs a patch like this to land on [m-c](https://gist.github.com/jasonLaster/74290204bbba442f9f4f144f175b37de)
* I killed the old l10n shim as it came from the devtools.html days and I believe there should be a better one. 
* The shim I added is a placeholder, but should be fine in terms of landing a working local version

### Testing

* tests won't pass until M-C is updated


packages/devtools-local-toolbox/public/js/index.js
```diff
@@ -19,6 +20,20 @@ const { getClient, connectClients, startDebugging } = client;
...
+  const theme = getValue("theme");
+  switch (theme) {
+    case "dark": require("./lib/themes/dark-theme.css"); break;
+    case "light": require("./lib/themes/light-theme.css"); break;
+    case "firebug": require("./lib/themes/firebug-theme.css"); break;
+  }
+  document.body.parentNode.classList.add(`theme-${theme}`);
+
+  window.L10N = L10N;
+}
```
jasonLaster: moved this out as it was helpful to do it at the point of module declaration

public/js/strings.json
```diff
@@ -0,0 +1,27 @@
...
+  "debugger.commands.pauseOnAllExceptions": "Pause on all exceptions. Click to ignore exceptions",
+  "debugger.commands.stepOut": "Step Out (%S%SF11)",
+  "debugger.commands.stepIn": "Step In (%SF11)",
+  "debugger.commands.stepOver": "Step Over (F10)",
+  "debugger.commands.resume": "Click to resume (F8)",
+  "debugger.commands.waitOnBreak": "Waiting for next execution",
+  "debugger.commands.pause": "Click to pause (F8)",
+  "debugger.welcome.search": "%S+P to search for files",
+  "debugger.sourceSearch.search": "Search...",
+  "debugger.sourceSearch.noResults": "No files matching %S found"
```
jasonLaster: not surprised, but there are lots of dupes in the properties file which should be reconciled...

jlongster:
Wow, this is a lot better of an implementation than I was thinking of doing. Great work. I'm glad you took this :)

My only worry is how we'll keep `strings.json` in sync. Is there any way we could just copy over the debugger properties file and parse it? Or is the parser not a normal JS module that can copy over too? How did you generate `strings.json`?

jlongster:
Note that the debugger window could have already loaded by the time we set the `L10N` object on it, so we should be careful about using it before it's ready. I think we're OK because we should only use it after calling `bootstrap`, which is done later.

jasonLaster:
> My only worry is how we'll keep strings.json in sync

I agree, it'll be nice to be able to include the `debugger.properties` file in github and copy it over with the bundle. I have a patch that adds a java properties loader... it is still failing, but close...

> Note that the debugger window could have already loaded by the time we set the L10N object

Yeah, this could be an issue down the line, but i think we should be fine for now. The strings in the debugger app will be fine.

Issue #1005, opened 2016-10-25
[Flow] Type the Source Map Worker


jasonLaster:
The [source map worker](https://github.com/devtools-html/debugger.html/blob/master/public/js/utils/source-map-worker.js) handles some of the most complicated logic in the debugger. It would really help to type the module for clarity and refactoring confidence.

Here is the module's [report card](http://jasonlaster.github.io/debugger-flow-report/sourcefiles/public/js/utils/source-map-worker.js.html).

**how do i run flow?** `flow`

**how do i see a module's coverage?** `npm run flow-coverage`

sankha93:
I am working on this.

wldcordeiro:
Thanks @sankha93 feel free to ask any questions here if you get stuck.

PR #1026, opened 2016-10-27
add flow types to source maps worker

sankha93:
Associated Issue: #1005
### Summary of Changes
- Type the Source Map Worker
### Testing
- [x] passes `npm test`
- [x] passes `npm run lint`

### DIFF
--- a/public/js/types.js
+++ b/public/js/types.js
@@ -13,8 +13,8 @@
  */
 export type Source = {
   id: string,
-  url: string,
-  sourceMapURL: string
+  url?: string,
+  sourceMapURL?: string
 };
 
 /**
--- a/public/js/types.js
+++ b/public/js/types.js
@@ -13,8 +13,8 @@
  */
 export type Source = {
   id: string,
-  url?: string,
-  sourceMapURL?: string
+  url: string,
+  sourceMapURL: string
 };
 
 /**
--- a/public/js/utils/path.js
+++ b/public/js/utils/path.js
@@ -1,23 +1,23 @@
 // @flow
 
-function basename(path: string) : string {
+function basename(path: string) {
   return path.split("/").pop();
 }
 
-function dirname(path: string) : string {
+function dirname(path: string) {
   const idx = path.lastIndexOf("/");
   return path.slice(0, idx);
 }
 
-function isURL(str: string) : boolean {
+function isURL(str: string) {
   return str.indexOf("://") !== -1;
 }
 
-function isAbsolute(str: string) : boolean {
+function isAbsolute(str: string) {
   return str[0] === "/";
 }
 
-function join(base: string, dir: string) : string {
+function join(base: string, dir: string) {
   return base + "/" + dir;
 }
 
--- a/public/js/utils/path.js
+++ b/public/js/utils/path.js
@@ -1,20 +1,26 @@
-function basename(path) {
+// @flow
+
+function basename(path: string) : string {
   return path.split("/").pop();
 }
 
-function dirname(path) {
+function dirname(path: string) : string {
   const idx = path.lastIndexOf("/");
   return path.slice(0, idx);
 }
 
-function isURL(str) {
+function isURL(str: string) : boolean {
   return str.indexOf("://") !== -1;
 }
 
-function isAbsolute(str) {
+function isAbsolute(str: string) : boolean {
   return str[0] === "/";
 }
 
+function join(base: string, dir: string) : string {
+  return base + "/" + dir;
+}
+
 module.exports = {
-  basename, dirname, isURL, isAbsolute
+  basename, dirname, isURL, isAbsolute, join
 };
--- a/public/js/utils/source-map-util.js
+++ b/public/js/utils/source-map-util.js
@@ -2,20 +2,20 @@
 
 const md5 = require("md5");
 
-function originalToGeneratedId(originalId: string) : string {
+function originalToGeneratedId(originalId: string) {
   const match = originalId.match(/(.*)\/originalSource/);
   return match ? match[1] : "";
 }
 
-function generatedToOriginalId(generatedId: string, url: string) : string {
+function generatedToOriginalId(generatedId: string, url: string) {
   return generatedId + "/originalSource-" + md5(url);
 }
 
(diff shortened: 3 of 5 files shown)


public/js/utils/path.js
```diff
@@ -1,20 +1,26 @@
...
   return path.slice(0, idx);
 }
 
-function isURL(str) {
+function isURL(str: string) : boolean {
   return str.indexOf("://") !== -1;
 }
 
-function isAbsolute(str) {
+function isAbsolute(str: string) : boolean {
```
jasonLaster: hmm, i could be wrong here, but I think flow can infer the return types

public/js/utils/source-map-util.js
```diff
@@ -1,19 +1,21 @@
...
 
-function generatedToOriginalId(generatedId, url) {
+function generatedToOriginalId(generatedId: string, url: string) : string {
   return generatedId + "/originalSource-" + md5(url);
 }
 
-function isOriginalId(id) {
-  return id.match(/\/originalSource/);
+function isOriginalId(id: string) : boolean {
+  return !!id.match(/\/originalSource/);
```
jasonLaster: 👍

public/js/types.js
```diff
@@ -13,8 +13,8 @@
  */
 export type Source = {
   id: string,
-  url?: string,
-  sourceMapURL?: string
+  url: string,
+  sourceMapURL: string
```
jasonLaster: 👍

public/js/utils/source-map-worker.js
```diff
@@ -64,11 +75,12 @@ function _setSourceMapRoot(sourceMap, absSourceMapURL, source) {
...
 }
 
-function _getSourceMap(generatedSourceId) {
+function _getSourceMap(generatedSourceId: string)
+    : ?Promise<SourceMapConsumer> {
   return sourceMapRequests.get(generatedSourceId);
 }
 
-async function _resolveAndFetch(generatedSource) {
+async function _resolveAndFetch(generatedSource: Source) : SourceMapConsumer {
```
jasonLaster: where did `SourceMapConsumer` return type come from?

jasonLaster:
I just refactored _resolveSourceMapURL so that flow would be happy. The issue there was that flow could not resolve url.parse's properties to strings. Adding the default empty strings did the trick. The rest was just cleanup :)

public/js/utils/path.js
jlongster: If it's an exported function, you need to fully annotate it.

public/js/types.js
jlongster: Why these changes? This is wrong. A source might not have a URL, and definitely will not have a sourceMapURL most of the time.

public/js/utils/source-map-worker.js
```diff
@@ -22,21 +33,21 @@ function enableSourceMaps() {
   sourceMapsEnabled = true;
 }
 
-function _resolveSourceMapURL(source) {
-  if (path.isURL(source.sourceMapURL) || !source.url) {
+function _resolveSourceMapURL(source: Source) : string {
+  const { url, sourceMapURL } = source;
```
jlongster: Please don't destructure this, the code is clearer by simply referencing `source` below. Unless you did this for a specific flow reason?

jasonLaster: will revert

jasonLaster:
@sankha93 i removed the return types as that did not affect the coverage

public/js/types.js
jasonLaster: nice catch, was thinking it would be an empty string, but if the field is missing it should be `?`

wldcordeiro: If it's nullable the `?` is what you use.

jasonLaster: yup

public/js/utils/source-map-worker.js
wldcordeiro: Seems like this @jasonLaster  https://github.com/mozilla/source-map

jasonLaster:
@sankha93 i made a couple of cleanups based on the comments. It looks like flow is complaining in some cases now because it can't assume sourceMapURL is a string. We will likely need to set reasonable defaults like in URL.parse. Thanks for the help again.

public/js/utils/source-map-worker.js
```diff
@@ -22,21 +33,21 @@ function enableSourceMaps() {
   sourceMapsEnabled = true;
 }
 
-function _resolveSourceMapURL(source) {
-  if (path.isURL(source.sourceMapURL) || !source.url) {
+function _resolveSourceMapURL(source: Source) {
+  const { url, sourceMapURL } = source;
```
jlongster: Just a reminder: (since other comment was on an old commit, I believe) don't destructure here and follow what the code did before.

sankha93: @jlongster I think if we should destructure the `source` object and set the `url` and `sourceMapURL` values to an empty string by default. It would pass Flow's static analysis for `path.isURL` or `path.isAbsolute` since they only expect a string here. /cc @jasonLaster

sankha93: I just pushed the above changes in the latest commit. I'll change it whatever is recommended otherwise.

jasonLaster: Thanks. I think the decision was to avoid destructuring, but still use defaults below.

public/js/utils/source-map-worker.js
```diff
@@ -27,21 +38,21 @@ function enableSourceMaps() {
   sourceMapsEnabled = true;
 }
 
-function _resolveSourceMapURL(source) {
-  if (path.isURL(source.sourceMapURL) || !source.url) {
+function _resolveSourceMapURL(source: Source) {
+  const { url = "", sourceMapURL = "" } = source;
```
jasonLaster: @jlongster @sankha93 I didn't realize that we needed to use the defaults to fix the source type. 👍 

This looks good

jasonLaster:
Thanks for the help @sankha93. I just did a quick check and these fixes brought our score from 42% to 72% in the worker, with obviously no errors!

Issue #970, opened 2016-10-20
[Conditional Breakpoints] Clicking on the gutter


jasonLaster:
- clicking on the breakpoint while the panel is open should close the panel and remove the breakpoint.
- clicking on the breakpoint when the panel is closed should open the panel and focus.
- cmd clicking on the breakpoint should focus on the panel input. We should not create lots of new panels :) 

Fail gif
![](http://g.recordit.co/cb3XFA1ZW9.gif)

jacobjzhang:
@jasonLaster I'll take this one

jasonLaster:
Thanks for the help! I'm closing this and opening a small issue for focusing the input on open

Issue #969, opened 2016-10-20
Style Conditional Breakpoints Panel


jasonLaster:
We recently landed conditional breakpoints here #922, but they need to be styled and tweaked. This issue is just about making them pretty, which I have some designs for below.

The guillemet can be unicode: [»](http://www.fileformat.info/info/unicode/char/bb/index.htm)

Here's what it currently looks like

![](http://g.recordit.co/LbZV3IVNOz.gif)
### Default text

![](https://cloud.githubusercontent.com/assets/254562/19328919/17079408-90a4-11e6-9de3-f78f4bb1d742.png)
### Expression

![](https://cloud.githubusercontent.com/assets/254562/19328920/170c493a-90a4-11e6-907b-4a393e5ccd36.png)
### Code changes:

Most of the work is in [Editor.js](https://github.com/devtools-html/debugger.html/blob/master/public/js/components/Editor.js#L35-L64) with some CSS changes in Editor.css

jacobjzhang:
@jasonLaster I'll take this one as well

PR #1031, opened 2016-10-27
Conditional breakpoint improvement

jacobjzhang:
Associated Issue: #969, #970

### Summary of Changes

* Cleaned up styling of conditional breakpoints panel
* Added state to ensure only one conditional breakpoint gets opened at a time

### Testing

* [X] passes `npm test`
* [X] passes `npm run lint`


public/js/components/Editor.js
```diff
@@ -49,14 +49,11 @@ function renderConditionalBreakpointPanel(
...
-      dom.div(
-        { className: "header" },
-        `This breakpoint will stop execution only
-         if the following expression is true`
-      ),
       dom.input({
         defaultValue: condition,
+        placeholder: "This breakpoint will pause when the expression is true",
         onKeyPress: onKey,
+        autofocus: true
```
jasonLaster: nice

public/js/components/Editor.js
```diff
@@ -101,12 +98,18 @@ const Editor = React.createClass({
     const closePanel = () => this.cbPanels[line].clear();
 
     if (isEnabled("conditionalBreakpoints") && bp && ev.metaKey) {
-      const { condition } = bp;
-      const panel = renderConditionalBreakpointPanel({
-        location, setBreakpointCondition, condition, closePanel
-      });
-
-      this.cbPanels[line] = this.editor.codeMirror.addLineWidget(line, panel);
+      if (!this.state.isCondBPOpen) {
```
jasonLaster: Hmm, I'm not sure if this would work, but it might be nice to try getting rid of the two new state fields: `isCondBPOpen` and `openPanel`.

I think we can do that because, `this.cbPanels[line]` should have all the information we need. e.g.

`this.state.isCondBPOpen == !! this.cbPanels[line]`
`this.openPanel == this.cbPanels[line]`

public/js/components/Editor.js
```diff
@@ -101,12 +98,18 @@ const Editor = React.createClass({
...
+        const { condition } = bp;
+        const panel = renderConditionalBreakpointPanel({
+          location, setBreakpointCondition, condition, closePanel
+        });
+
+        this.cbPanels[line] = this.editor.codeMirror.addLineWidget(line, panel);
+        this.setState({ isCondBPOpen: true, openPanel: this.cbPanels[line] });
+      } else {
+        this.state.openPanel.clear();
+        this.setState({ isCondBPOpen: false, openPanel: {}});
```
jasonLaster: we should nuke `this.cbPanels[line]` here i.e. `delete this.cbPanels[line]`

public/js/components/Editor.js
jacobjzhang: Tried this idea, it didn't work. Some weird codemirror thing is going on where the cbPanels[line] is still undefined by the time we re-render onGutterClick to close it. This might be directly related to the inability to focus, could be interesting to dive into codemirror in the future.

jasonLaster: Cool. This works. We can take another look when we work on auto focusing.

public/js/components/Editor.js
jacobjzhang: Good call.

public/js/components/Editor.js
jacobjzhang: Removed as this was useless.

PR #1035, opened 2016-10-27
Add sourcemap and pretty-printing mochitests

jlongster:
This is a starting point for some more tests. There are still other edge cases we should test, but some things are undefined behavior until we know what UX we want. For example, things like what happens when you set a breakpoint on a generated source that has a sourcemap? Right now, it moves the breakpoint to the original source, but I'm not sure we want that behavior.

This covers the most serious uses though. Many of the other mochitests in the old debugger aren't applicable anymore either because we have a different UX or we don't have certain features yet (like blackboxing).


public/js/main.js
```diff
@@ -54,7 +54,7 @@ if (isFirefoxPanel()) {
     destroy: () => {
       unmountRoot();
       sourceMap.destroyWorker();
-      prettyPrint.destoryWorker();
+      prettyPrint.destroyWorker();
```
jasonLaster: what's the change here? I've been seeing `destroyWorker` does not exist.

clarkbw: destroy not destory 😀

jasonLaster: Ahhhh. I see :)

jlongster: Yep!

jlongster:
Triggering it again on CI to make sure it passes again, then will merge.

jasonLaster:
Great!

Issue #1039, opened 2016-10-27
Source Search - hide pretty sources


jasonLaster:
We should filter the pretty sources from the list of sources shown in the sources search UI.

![screen shot 2016-10-27 at 5 48 22 pm](https://cloud.githubusercontent.com/assets/254562/19786700/40ace1ee-9c6d-11e6-870b-c6c3e6900833.png)

Here's a sketch of what the patch could look like:

``` diff
diff --git a/public/js/components/SourceSearch.js b/public/js/components/SourceSearch.js
index 60fafd8..2f0dbe1 100644
--- a/public/js/components/SourceSearch.js
+++ b/public/js/components/SourceSearch.js
@@ -6,6 +6,7 @@ const actions = require("../actions");
 const { getSources, getSelectedSource } = require("../selectors");
 const { endTruncateStr } = require("../utils/utils");
 const { parse: parseURL } = require("url");
+const { isPretty } = require("../utils/source");

 require("./SourceSearch.css");

@@ -17,7 +18,9 @@ function searchResults(sources) {
     return endTruncateStr(path, 50);
   }

-  return sources.valueSeq()
+  return sources
+    .valueSeq()
+    .filter((source) => !isPretty(source.toJS()))
     .filter(source => source.get("url"))
     .map(source => ({
       value: getSourcePath(source),
```

jacobjzhang:
@jasonLaster I'll grab this one

PR #1044, opened 2016-10-28
Source search hide pretty

jacobjzhang:
Associated Issue: #1039

### Summary of Changes

* Filtering out pretty sources

### Testing

* [X] passes `npm test`
* [X] passes `npm run lint`

### DIFF
--- a/public/js/components/SourceSearch.js
+++ b/public/js/components/SourceSearch.js
@@ -23,7 +23,7 @@ function searchResults(sources) {
   }
 
   return sources.valueSeq()
-    .filter(source => !isPretty(source.toJS()))
+    .filter(source => !isPretty(source.toJS() || source.get("url")))
     .filter(source => source.get("url"))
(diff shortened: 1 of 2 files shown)


public/js/components/SourceSearch.js
```diff
@@ -22,6 +23,7 @@ function searchResults(sources) {
   }
 
   return sources.valueSeq()
+    .filter(source => !isPretty(source.toJS()))
```
wldcordeiro: I think this could be turned into one filter `filter(source => !isPretty(source.toJS()) || source.get("url"))` don't have my computer up to test though.

jasonLaster: good point, that's probably slightly clearer.

jacobjzhang: thanks @wldcordeiro, looks good from my end

PR #1046, opened 2016-10-28
Fix close button size

jlongster:
@helenvholmes Apparently changing the close button svg's viewport made it shrink. I don't really understand svg to know what's going on here (how could making it bigger make it smaller??), but here is a before and after:

Before this PR:

screen shot 2016-10-28 at 5 17 46 pm

After:

screen shot 2016-10-28 at 5 17 21 pm

Changing it back to 6, 6 fixes it, but that does seem inconsistent. Any ideas?


jlongster:
I think Helen might be gone for the day, and I'd really like to land a debugger update. I may merge this and do a follow-up so that the buttons don't look broken.

jlongster:
Good call, just tested and it is correct as well.

PR #1047, opened 2016-10-28
Remove max tab width

jlongster:
There's no reason we should have a max tab width in my opinion. The current CSS is broken, leading to this when it's too wide:

screen shot 2016-10-28 at 5 25 30 pm

For now let's just remove this. We can think of how to solve this better later.

### DIFF
--- a/public/js/components/SourceTabs.css
+++ b/public/js/components/SourceTabs.css
@@ -46,10 +46,7 @@
   position: relative;
   transition: all 0.25s ease;
   min-width: 40px;
-  max-width: 132px;
   overflow: hidden;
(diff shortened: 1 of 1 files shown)


jlongster:
The word wrapping, in my testing, just wraps down and is hidden like the screenshot because tabs have a fixed height (rightly so). So I'm not sure what UX you're thinking of. I think the most straight-forward thing to do would be to do the text-overflow: ellipses thing.

For now, let's remove it to have it working again.

wldcordeiro:
`text-overflow: ellipsis` maybe?

Issue #1041, opened 2016-10-27
File Search - blinking orange cursor


jasonLaster:
When the file search bar opens, the cursor is orange. After you've typed, it's no longer orange. We should fix this so that when it opens it is the normal grey :)

Not really sure what CSS magic we need to use because it will be orange when the search results are empty :)

![](http://g.recordit.co/ensezSkeZl.gif)

ryandoherty:
Got a browser/OS/revision handy? I've tested this with 2b7d50fcf197d680c0f842c658e62043784b19f7 on Firefox 49.0.2 and Chrome 54.0.2840.71 on OSX 10.11.6 and can't reproduce. Thanks!

jasonLaster:
Weird. Thanks!

PR #1051, opened 2016-10-30
#1041: File Search - blinking orange cursor

briantully:
Associated Issue: #1041

### Summary of Changes

* add query const to render method so that we only apply the "empty" className if the query is not empty, thereby only changing the cursor to orange if empty search result and NOT when search input is first visible.

### Testing

* [ X] passes `npm test`
* [ X] passes `npm run lint`

1041-debugger-cursor-searchbar


briantully:
awesome! thanks, @jasonLaster :)

Issue #1034, opened 2016-10-27
[flow] Add types for utilities


jasonLaster:
There are several util files that types would really help. I think if we can type these files, our overall score would go up a lot.
- http://jasonlaster.github.io/debugger-flow-report/sourcefiles/public/js/utils/assert.js.html
- http://jasonlaster.github.io/debugger-flow-report/sourcefiles/public/js/utils/defer.js.html
- http://jasonlaster.github.io/debugger-flow-report/sourcefiles/public/js/utils/log.js.html
- devtools-config/feature.js (not in the list, but i've seen devtools-confg mentioned in several places).

AbhiGulati:
I would like to work on this.

AbhiGulati:
I am still working on this. As far as I can tell, @jacobjzhang's PR doesn't add annotations for defer.js and log.js. I should have a PR ready by tomorrow.

jasonLaster:
Thanks. I'll try and create some more flew issues. Running npm run flow-coverage shows lots of areas for improvement

jacobjzhang:
@AbhiGulati apologies, didn't see you had called this.

PR #1052, opened 2016-10-30
Adding missing annotations and conditional check got deferred

jacobjzhang:
Associated Issue: #1034
### Summary of Changes
- Adding missing flow type annotations and conditional check got deferred
### Testing
- [X] passes `npm test`
- [X] passes `npm run lint`

### DIFF
--- a/packages/devtools-config/src/feature.js
+++ b/packages/devtools-config/src/feature.js
@@ -7,11 +7,11 @@ const flag = require("./test-flag");
  * Gets a config value for a given key
  * e.g "chrome.webSocketPort"
  */
-function getValue(key :mixed) {
+function getValue(key: string) {
   return pick(config, key);
 }
 
-function isEnabled(key :mixed) {
+function isEnabled(key: string) {
   return config.features[key];
 }
 
@@ -35,7 +35,7 @@ function isFirefox() {
   return /firefox/i.test(navigator.userAgent);
 }
 
-function setConfig(value :Object) {
+function setConfig(value: Object) {
   config = value;
 }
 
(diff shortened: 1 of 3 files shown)


packages/devtools-config/src/feature.js
```diff
@@ -7,11 +7,11 @@ const flag = require("./test-flag");
  * Gets a config value for a given key
  * e.g "chrome.webSocketPort"
  */
-function getValue(key) {
+function getValue(key :mixed) {
   return pick(config, key);
 }
 
-function isEnabled(key) {
+function isEnabled(key :mixed) {
```
jasonLaster: This should be a strIng 

Nit: param colon space type

public/js/actions/sources.js
```diff
@@ -317,8 +317,11 @@ function getTextForSources(actors: any[]) {
     function maybeFinish() {
       if (pending.size == 0) {
         // Sort the fetched sources alphabetically by their url.
-        deferred.resolve(
-          fetched.sort(([aFirst], [aSecond]) => aFirst > aSecond ? -1 : 1));
+        if (deferred) {
+          deferred.resolve(
+            fetched.sort(([aFirst], [aSecond]) => aFirst > aSecond ? -1 : 1)
+          );
+        }
```
jasonLaster: Why do we need the if?

public/js/utils/assert.js
```diff
@@ -1,4 +1,4 @@
-function assert(condition, message) {
+function assert(condition :any, message: string) {
```
jasonLaster: I think the condition is a boolean

public/js/actions/sources.js
jacobjzhang: deferred is assigned defer() at the beginning of the script which has the possibility of being undefined, one of the things flowtype pointed out actually.

PR #1054, opened 2016-10-31
(WIP) Add gutter context menu

jasonLaster:
### Summary of Changes

* Adds support for showing a shim context menu locally
* Adds support for showing real context menus in the panel *(I need to clean this up)*
* Adds an editor gutter context menu
* Hides context menus globally


public/js/components/SourcesTree.js
```diff
@@ -99,6 +99,7 @@ let SourcesTree = React.createClass({
             hidden: !nodeHasChildren(item) }
         ),
         onClick: e => {
+          e.preventDefault();
```
jlongster: What's this for?

public/js/main.js
```diff
@@ -49,12 +49,13 @@ if (isFirefoxPanel()) {
...
 
   module.exports = {
-    bootstrap: ({ threadClient, tabTarget }) => {
+    bootstrap: ({ threadClient, tabTarget, toolbox }) => {
       // TODO (jlast) remove when the panel has L10N
       if (!window.L10N) {
         window.L10N = require("../../packages/devtools-local-toolbox/public/js/utils/L10N");
         L10N.setBundle(require("./strings.json"));
       }
+      window.toolbox = toolbox;
```
jlongster: Why do this?

public/js/strings.json
```diff
@@ -6,6 +6,10 @@
   "callStack.expand": "Expand Rows",
   "editor.searchResults": "%d of %d results",
   "editor.noResults": "no results",
+  "editor.addBreakpoint": "Add Breakpoint",
+  "editor.removeBreakpoint": "Remove Breakpoint",
+  "editor.addConditionalBreakpoint": "Add Conditional Breakpoint",
+  "editor.editConditionalBreakpoint": "Edit Conditional Breakpoint",
```
jlongster: How are we going to keep strings in sync? It worries me a little bit that we need to maintain them in two places.

public/js/main.js
jasonLaster: this is temporary to support the contextmenu in the panel... i'll clean this up.

packages/devtools-local-toolbox/index.html
```diff
@@ -11,6 +11,7 @@
   </head>
 
   <body>
+    <popupset></popupset>
```
jlongster: Isn't this a XUL element? What purpose does it have here?

jlongster:
Good work! Just a few questions, more for curiosity than anything else.

packages/devtools-local-toolbox/index.html
jasonLaster: haha - it is :), but we are *using* it here in a shim/sham like way

public/js/components/SourcesTree.js
jasonLaster: it's left over from the first pass which was w/ the source tree

jasonLaster: nothing, it's going away. :P

packages/devtools-sham-modules/client/framework/menu.js
```diff
@@ -0,0 +1,176 @@
+/* -*- indent-tabs-mode: nil; js-indent-level: 2 -*- */
```
jasonLaster: this is almost a direct copy from m-c. I took a lot of pains to move the *sham* work to `utils/menu`. A regular `menuitem` will behave like a context menu in a web context.

Three things changed:
1. i referenced the sham events emitter (we should consider de-priviledging events emitter a top priority, it's really close)
2. I added `popup` as a argument to the close and open events
3. I created a new `createPopup` function that's overridden in the web context.

public/debugger.properties
```diff
@@ -0,0 +1,367 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
```
jasonLaster: This is the most interesting change since the first WIP. I completely agree that updating strings.json is an issue for both CI and in general when making a new bundle.

My current solution is to add `debugger.properties` to github. Now when we make a new bundle it's copied over. This adds a small amount of risk that we override new strings added to m-c, but we should see that in the patch pretty clearly. Also, as we get closer to officially killing the old debugger it's less of a risk.

Obviously strings.json should die soon, but that seems out of scope at the moment and we have the issue for using a webpack java properties loader.

packages/devtools-config/configs/firefox-panel.json
```diff
@@ -6,6 +6,7 @@
   "features": {
     "tabs": true,
     "sourceMaps": true,
-    "prettyPrint": true
+    "prettyPrint": true,
+    "conditionalBreakpoints": true
```
jasonLaster: this turns it on for real

public/js/components/SourcesTree.js
jlongster: Aaah ok!

packages/devtools-config/configs/firefox-panel.json
jlongster: Can we just remove the conditionalBreakpoints flag instead?

jlongster:
I looked through this and nothing jumped out at me except for the properties. I'm worried about having to sync them back and forth. Most likely there will be tweaks that people land on m-c without telling us.

It might be best if the properties file lives on m-c, and we don't actually sync it from github to m-c, but the other way around. It becomes similar to the shim modules in that we copy it from m-c to github. That way, if we introduce new strings, we need to do it separately on m-c, which is annoying but at least we'll have one authoritative source.

jasonLaster:
@jlongster yeah - I see what you mean. and thanks for pointing out the Search... change, that was a change following up on the new debugger strings. 

In general I'd like github to be the place where active development is done, but that doesn't mean we shouldn't improve how we check for changes when we make the bundle.  I'd like to improve our check so that we can guarantee we don't accidentally override changes.

Issue #968, opened 2016-10-20
Code Search - debounce searches


jasonLaster:
We should debounce the code search while the user is typing. This is a minor UX tweak that other editors like atom do that really helps the UI feel smooth and not very jumpy.

![](http://g.recordit.co/Jti0vuu5nX.gif)

Here's a sketch of what the code could look like:

``` diff
diff --git a/public/js/components/EditorSearchBar.js b/public/js/components/EditorSearchBar.js
index cecba60..cbb740e 100644
--- a/public/js/components/EditorSearchBar.js
+++ b/public/js/components/EditorSearchBar.js
@@ -5,6 +5,7 @@ const Svg = require("./utils/Svg");
 const { isEnabled } = require("devtools-config");
 const { find, findNext, findPrev } = require("../utils/source-search");
 const classnames = require("classnames");
+const debounce = require("lodash.debounce");

 require("./EditorSearchBar.css");

@@ -92,25 +93,37 @@ const EditorSearchBar = React.createClass({
   },

   onKeyUp(e) {
-    const ed = this.props.editor;
-    const ctx = { ed, cm: ed.codeMirror };
-    const { query, index, count } = this.state;
-
     if (e.key != "Enter") {
       return;
     }

     if (e.shiftKey) {
-      findPrev(ctx, query);
-      const nextIndex = index == 0 ? count - 1 : index - 1;
-      this.setState({ index: nextIndex });
+      this.searchNext();
     } else {
-      findNext(ctx, query);
-      const nextIndex = index == count - 1 ? 0 : index + 1;
-      this.setState({ index: nextIndex });
+      this.searchPrev();
     }
   },

+  searchNext: debounce(function() {
+    const ed = this.props.editor;
+    const ctx = { ed, cm: ed.codeMirror };
+    const { query, index, count } = this.state;
+
+    findNext(ctx, query);
+    const nextIndex = index == count - 1 ? 0 : index + 1;
+    this.setState({ index: nextIndex });
+  }),
+
+  searchPrev: debounce(function() {
+    const ed = this.props.editor;
+    const ctx = { ed, cm: ed.codeMirror };
+    const { query, index, count } = this.state;
+
+    findPrev(ctx, query);
+    const nextIndex = index == 0 ? count - 1 : index - 1;
+    this.setState({ index: nextIndex });
+  }),
+
   renderSummary() {
     const { count, index, query } = this.state;
```

bomsy:
i'll like to fix this!

[...]

PR #1055, opened 2016-10-31
debounce the code search with a wait of 250ms

bomsy:
Associated Issue: #968

### Summary of Changes

* Created a new search function which is debounced

### Testing

* [x] passes `npm test`
* [x] passes `npm run lint`

debounce-search

### DIFF
--- a/public/js/components/EditorSearchBar.js
+++ b/public/js/components/EditorSearchBar.js
@@ -116,7 +116,7 @@ const EditorSearchBar = React.createClass({
     const ctx = { ed, cm: ed.codeMirror };
 
     find(ctx, query);
-  }, 250),
+  }, 100),
 
   renderSummary() {
     const { count, index, query } = this.state;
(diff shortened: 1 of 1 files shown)


jlongster:
250ms seems a bit long. How did you settle on that number?

jlongster:
We don't even do any network requests when searching, so it's a very quick search.

jasonLaster:
Agreed, although use your judgement when you're playing with it. I'd suggest sticking with something similar to sublime / atom

wldcordeiro:
I'd lean towards 100ms, as I know we coders tend to type pretty quickly so 1/10th of a second seems a bit more sensible than 1/4th of one.

bomsy:
@jlongster Just used 250ms, no particular reasons ... was hoping it would kick of the discussions.  :)
Just dug through the atom source, and they use 100ms, which matches @wldcordeiro recommendation. I would think that feels right, and will go with that. updates coming up!

jasonLaster:
@wldcordeiro 💯 

Thanks @bomsy will merge when CI passes

PR #1059, opened 2016-10-31
Allow frames to be focus-able

jlongster:
Helen was asking how to detect when a frame is "focused", and I discovered that all we have to do is add a "tabindex" to it to allow it to be focus-able. This is allows us to style it with `li:focus` and everything just works. When you click on a frame, it'll be a bold blue, but when you click away it'll change to a more subtle color.

We should able to use this technique in the tree component as well. Right now it uses a hidden button to allow each item to be focus-able, but it just needs a `tabindex` property to work.

I also had to change the click event to `mouseDown` because that's when the element gets focused. Otherwise it's too easy for a call frame to be "focused" but not "selected", just by pressing the mouse down on a frame but the moving the cursor outside of it. Previously this would not select it. Now, the focused and selected properties are always in sync.

### DIFF
--- a/public/js/components/Frames.css
+++ b/public/js/components/Frames.css
@@ -11,6 +11,12 @@
   overflow: hidden;
 }
 
+/* Style the focused call frame like so:
+.frames ul li:focus {
+  border: 3px solid red;
+}
+*/
+
 .frames ul li * {
   -moz-user-select: none;
   user-select: none;
--- a/public/js/components/Frames.js
+++ b/public/js/components/Frames.js
@@ -33,7 +33,9 @@ function renderFrame(frame, selectedFrame, selectFrame) {
   return dom.li(
     { key: frame.id,
       className: `frame ${selectedClass}`,
-      onClick: () => selectFrame(frame) },
+      onMouseDown: () => selectFrame(frame),
+      tabIndex: 0
+    },
     renderFrameTitle(frame),
     renderFrameLocation(frame)
   );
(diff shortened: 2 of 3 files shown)


jsnajdr:
Last time I worked on something similar one month ago - focusing tree items in Performance tool - the button hack was necessary to enable keyboard focus on a DIV element on macOS. Without the button, it didn't focus the DIV although it had tabIndex.

Right now I'm working on a similar thing in netmonitor.html and things seem to work OK even without the button hack. So I'm a bit confused. Maybe some focus bug got fixed in the platform in the meantime. Or was it the Sierra upgrade?

jlongster:
@jsnajdr Yeah, it seems to work fine for me too. You bring up a good point though; I forgot that this makes it so that you can tab through them (of course) with the keys, so pressing `tab` will "highlight" frames that are not "selected". This might be confusing and we need to think through this.

Let's see what Helen comes up with for styles and we'll go from there.

jsnajdr:
What I mean is that the button hack is there for a reason in the Tree component, and setting `tabIndex` is not enough. At least it was necessary two months ago in macOS El Capitan.

Without the hack, the element never got focus, which means neither the element nor its parent received `keypress` events, which means keyboard navigation with arrows didn't work. Tab order is another, not entirely related problem.

I'd recommend testing this on older versions on OS X - it's likely that it won't work.

jlongster:
Yeah, I knew what you meant. There's an unrelated thing that we need to figure out which is differentiating between "selected" and "focused" styles and making that clear.

But I haven't updated my system, and it works fine on El Capitan. I don't know about older ones. For now I think we could move forward with this and re-evaluate if anyone finds problem. If anyone has an older OSX version, please try this out.

PR #1060, opened 2016-10-31
Fix #994: Fix tab colors in dark theme

### LINKED ISSUE #994: Dark Theme - polish tab colors
The tab colors are bad on dark theme. Need to be improved.

screenshot_20161025_140802

**How do you change themes locally?** You can edit the theme locally [here](https://github.com/devtools-html/debugger.html/blob/master/docs/local-development.md#themes)

**How do we fix these?** Add CSS overrides in the different components. Here's an example of how we could fix the right sidebar commands.

``` diff
diff --git a/public/js/components/RightSidebar.css b/public/js/components/RightSidebar.css
index 5f5f8b4..794aa0d 100644
--- a/public/js/components/RightSidebar.css
+++ b/public/js/components/RightSidebar.css
@@ -40,6 +40,10 @@ html[dir="rtl"] .command-bar {
   transition: opacity 200ms;
 }

+:root.theme-dark .command-bar > span {
+  stroke: var(--theme-body-color);
+}
+
 html:not([dir="rtl"]) .command-bar > span {
   margin-right: 0.7em;
 }
```

**How do I choose the right color?** Hmm, it's up to you. We have a list of variables [here](https://github.com/devtools-html/debugger.html/blob/master/packages/devtools-local-toolbox/public/js/lib/themes/variables.css). What I do is look for similar colors in the app and then see what variable it uses in the inspector.

helenvholmes:
Associated Issue: #994

### Summary of Changes

* Changes the ```theme-faded-tab-color``` in dark theme to a slightly lighter color
* Moves the tab variables out of ```variables.css``` into ```App.css```, as there's no corresponding m-c commit to add that variable

### Testing

* [x] passes `npm test`
* [x] passes `npm run lint`

### Screenshots/Videos
screen shot 2016-10-31 at 4 29 54 pm

### DIFF
--- a/packages/devtools-local-toolbox/public/js/lib/themes/variables.css
+++ b/packages/devtools-local-toolbox/public/js/lib/themes/variables.css
@@ -20,7 +20,6 @@
   --theme-contrast-background: #e6b064;
 
   --theme-tab-toolbar-background: #fcfcfc;
-  --theme-faded-tab-color: #7E7E7E;
   --theme-toolbar-background: #fcfcfc;
   --theme-toolbar-background-alt: #f5f5f5;
   --theme-selection-background: #4c9ed9;
@@ -83,7 +82,6 @@
   --theme-contrast-background: #ffb35b;
 
   --theme-tab-toolbar-background: #272b35;
-  --theme-faded-tab-color: #333945;
   --theme-toolbar-background: #272b35;
   --theme-toolbar-background-alt: #2F343E;
   --theme-selection-background: #5675B9;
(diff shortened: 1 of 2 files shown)


helenvholmes:
@wldcordeiro sorry for jumping on it even though you were assigned! We're just trying to knock all the dark theme issues down before the merge :)

wldcordeiro:
No worries @helenvholmes. It appears our Style lint didn't like the variable though. 🙁 

```
10:28  ✖  Expected "#7E7E7E" to be "#7e7e7e"   color-hex-case
```

I prefer the capitals myself too 😉

helenvholmes:
Okay, now it passes ```npm run lint``` _for real_. :)

I actually prefer lowercase @wldcordeiro! But that's what I get for copying and pasting from Sketch...

PR #1064, opened 2016-11-01
Annotate log and defer utils

AbhiGulati:
Associated Issue: #1034 

### Summary of Changes

* Add an interface file for devtools-config.js
* Add Flow annotations for utils/defer.js and utils/log.js

### Testing

* [x] passes `npm test`
* [x] passes `npm run lint`


jasonLaster:
Thanks @AbhiGulati, this is really great work. 

You're approach to declaring the `devtools-config` module seems great in this case. My only question is whether we should add the declaration in a top-level interfaces dir. Do you have examples of this in other projects? One thought I had was that we could add the interface in the devtools-config package, and reference it from eslintrc. This way, other projects that install devtools-config, could get the declaration perhaps...

jasonLaster:
@AbhiGulati I don't know if you have time, but I'd love to get your thoughts on our type support in the redux actions. 

Here's an issue for adding flow type to Actions - https://github.com/devtools-html/debugger.html/issues/1066. We've added some types to some of the other actions, but i think it's mostly low hanging fruit stuff. 

I'm not really sure how to type the redux `dispatch` and middleware stuff. I tried some stuff [here](https://github.com/devtools-html/debugger.html/commit/d67b31f8b06dcf18639396691cb7586e8acf7ff1#diff-195a73de407881358c8dfc0c5891f7feR18), but it was tough :)

AbhiGulati:
I'm not sure what the best place is for the interface files. I was just following tutorials that I found online. I think it would be reasonable to move it into the `devtools-config` module and then reference that from the flowconfig.

AbhiGulati:
I don't have any feedback on the existing type annotations or your PR; this was my first time working with Flow, and I just focused on getting the coverage up for the listed files. Have you looked into flow-typed? I think you should be able to grab type declarations for redux, but I'm not sure if it will help with thunk-middleware.

jasonLaster:
Oh interesting. I haven't looked at flow-typed, that could be good. You had
me fooled by the way. Thought you were a flow pro.

jasonLaster:
@AbhiGulati after thinking about it a bit more, I think the best thing to do would be to add an interface file to `devtools-config` and reference that in the `eslintrc`.

jasonLaster:
I moved the declaration to devtools-config. 

Thanks for the help @AbhiGulati. Would love your help going forward on types or any other issues.

PR #1065, opened 2016-11-01
Fix flow statements

jasonLaster:
The last Flow PR introduced two issues:
1. types to feature.js which broke `npm test` and `npm start`. I'm not sure how CI passed
2. it left out a `// @flow` comment


jlongster:
Cool, thanks!

Issue #1067, opened 2016-11-01
The panel uses the old search bar


jasonLaster:
cmd+f in the panel brings up the old search bar. This is likely due to the fact that the key binding is still there.

PR #1073, opened 2016-11-01
Enable search in the panel

jasonLaster:
Associated Issue: #1067 

### Summary of Changes

* enables search, which turns on the new search bar in the panel
* update the L10N bootstrap logic so that it expects L10N to be passed in. Previously, the assumption was that L10N was a global on the window, but that was a bad assumption
* there's an associated m-c change as well that i'll link to soon

### DIFF
--- a/public/js/components/Editor.js
+++ b/public/js/components/Editor.js
@@ -21,7 +21,7 @@ const Breakpoint = React.createFactory(require("./EditorBreakpoint"));
 
 const { getDocument, setDocument } = require("../utils/source-documents");
 const { shouldShowFooter } = require("../utils/editor");
-const { isEnabled, isFirefox } = require("devtools-config");
+const { isFirefox } = require("devtools-config");
 const { showMenu } = require("../utils/menu");
 
 require("./Editor.css");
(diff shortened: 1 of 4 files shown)


public/js/main.js
```diff
@@ -49,11 +49,13 @@ if (isFirefoxPanel()) {
   const prettyPrint = require("./utils/pretty-print");
 
   module.exports = {
-    bootstrap: ({ threadClient, tabTarget, toolbox }) => {
+    bootstrap: ({ threadClient, tabTarget, toolbox, L10N }) => {
```
jlongster: This is an unrelated change, can you make this a separate PR?

public/js/components/Editor.js
```diff
@@ -266,9 +265,12 @@ const Editor = React.createClass({
       enableCodeFolding: false,
       gutters: ["breakpoints"],
       value: " ",
-      extraKeys
+      extraKeys: {}
     });
 
+    // disables the default search shortcuts
+    this.editor._initShortcuts = () => {};
```
jlongster: I'm a little sad that we're diverging on another thing. It would be nice to work on the official editor in m-c instead of customizing it to our needs, but I know that's difficult right now because we need to separate it out into a module for that to really happen.

It's OK for now.

jasonLaster: Yeah, my initial change was to the editor.js for that reason, but I didn't want the headache of disabling it there, but not globally for all the tools. This will be better when our Editor Search Bar is global for all the tools `_shrug_`

public/js/main.js
jasonLaster: ugh... just merged this.

PR #1083, opened 2016-11-02
Expand rows

### LINKED ISSUE #1056: expand rows appears when there are no more rows
For a short call stack the Expand rows option shouldn't exist, however it seems to be there when it isn't needed.

expand rows

jasonLaster:
Fix #1056 

### Summary of Changes

* hides expand rows when there aren't enough frames
* refactors frames to cleanup the frame rendering logic + toggle button logic

### DIFF
--- a/public/js/components/Frames.js
+++ b/public/js/components/Frames.js
@@ -13,6 +13,8 @@ if (typeof window == "object") {
   require("./Frames.css");
 }
 
+const NUM_FRAMES_SHOWN = 7;
+
 function renderFrameTitle(frame) {
   return div({ className: "title" }, endTruncateStr(frame.displayName, 40));
 }
@@ -65,7 +67,8 @@ const Frames = React.createClass({
   renderFrames() {
     let { frames } = this.props;
 
-    const numFramesToShow = this.state.showAllFrames ? frames.size : 7;
+    const numFramesToShow = this.state.showAllFrames
+      ? frames.size : NUM_FRAMES_SHOWN;
     frames = frames.slice(0, numFramesToShow);
 
     return dom.ul({}, frames.map(frame => this.renderFrame(frame)));
@@ -76,7 +79,7 @@ const Frames = React.createClass({
     let buttonMessage = this.state.showAllFrames
       ? L10N.getStr("callStack.collapse") : L10N.getStr("callStack.expand");
 
-    if (frames.size < 7) {
+    if (frames.size < NUM_FRAMES_SHOWN) {
       return null;
     }
 
--- a/public/js/components/Frames.js
+++ b/public/js/components/Frames.js
@@ -25,22 +25,6 @@ function renderFrameLocation(frame) {
   );
 }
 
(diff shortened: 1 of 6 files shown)


jlongster:
Looks like a nice cleanup. Thanks!

public/js/components/Frames.js
```diff
@@ -60,40 +46,66 @@ const Frames = React.createClass({
...
+    );
+  },
+
   render() {
-    const { frames, selectedFrame, selectFrame } = this.props;
-    const numFramesToShow = this.state.showAllFrames ? frames.length : 7;
-    let framesDisplay;
+    const { frames } = this.props;
 
     if (frames.length === 0) {
```
Zacqary: `frames` appears to be changing back and forth between an Array of `length` 0 and an Immutable Set with a `size`.

That's what was initially causing #1056, because `frames.length < numFramesToShow` would never return true, because `frames.length` was always `undefined` if there were any frames. This new PR works but we've still got two different types of data coming into this component.

Still trying to figure out where that's happening. This PR works but I still feel like `frames` needs to consistently be a Seq, or either `null` or a Seq. Switching between an Array and a Seq is probably how this bug appeared in the first place.

public/js/test/mochitest/browser_dbg-call-stack.js
```diff
@@ -12,16 +12,51 @@ function isFrameSelected(dbg, index, title) {
...
+  invokeInTab("startRecursion");
+  yield waitForPaused(dbg);
+
+  ok(isFrameSelected(dbg, 1, "recurseA"), "the first frame is selected");
+
+  // check to make sure that the toggle button isn't there
+  let button = toggleButton(dbg);
+  let frames = findAllElements(dbg, "frames");
+  is(button.innerText, "Expand Rows", "toggle button should be expand");
+  is(frames.length, 7, "There should be at most seven frames");
```
Zacqary: Should this be `frames.size`?

public/js/components/Frames.js
Zacqary: Here it is: https://github.com/devtools-html/debugger.html/blob/master/public/js/reducers/pause.js#L140

```javascript
function getFrames(state) {
  return state.pause.get("frames") || [];
}
```

Should be
```javascript
const { Seq } = require("immutable");
// ...
function getFrames(state) {
  return state.pause.get("frames") || Seq();
}
```

public/js/test/mochitest/browser_dbg-call-stack.js
jasonLaster: haha - that's number of DOM elements

public/js/components/Frames.js
jasonLaster: @Zacqary PR up that fix. I like it

Zacqary: PR into your branch or into master?

jasonLaster: A new branch

Zacqary: Ah well you just merged it so I guess into master :)

PR #1088, opened 2016-11-02
Add Search tests

jasonLaster:
Associated Issue: #1067 

### Summary of Changes

* Adds some mochi tests for source search. I'd like to add some unit tests later as most of this is more suited for that

* I tried really hard to get file search tests in, but for whatever reason I kept opening the content search bar and not the editor search bar. 

here's a small patch for the file search tests
```js
// Testing file search
add_task(function* () {
  yield selectSource(dbg, "switching-01");

  // window.dbg = dbg;
  // yield pauseTest();
  const el = findElementWithSelector(dbg, ".CodeMirror-line").focus();
  console.log("el", el);
  pressKey(dbg, "fileSearch");
  type(dbg, "fun");
  debugger;
});
```

### DIFF
--- a/public/js/test/mochitest/.eslintrc
+++ b/public/js/test/mochitest/.eslintrc
@@ -52,6 +52,7 @@
     "invokeInTab": false,
     "findSource": false,
     "findElement": false,
+    "findElementWithSelector": false,
     "findAllElements": false,
     "openNewTabAndToolbox": false,
     "selectSource": false,
@@ -68,6 +69,7 @@
     "isVisibleWithin": false,
     "clickElement": false,
     "togglePauseOnExceptions": false,
+    "type": false,
     "pressKey": false,
     "EXAMPLE_URL": false
   }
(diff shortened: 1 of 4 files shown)


jasonLaster:
Hmm, the error is `findElementWithSelector()` is null... this could be a race condition...

jlongster:
`pressKey` just sends the key but you immediately query the DOM afterwards. It's very likely that the key hasn't even been handled yet, so it makes sense that the element is `null`. Try just doing a `yield waitForTick()` to allow the key to be handled.

jasonLaster:
@jlongster the issue was with the modifier for `<cmd>+p`. CI is linux and needs ctrl, which we ran into with Jaideep's work with stepping. 

After thinking about it a bit more, I don't think we have to worry about debugger DOM events being async. I believe the mochitests run in the same context as the toolbox and will be able do this work synchronously.

here was the fix
```
const cmdOrCtrl = isLinux ? { ctrlKey: true } : { metaKey: true };
```

jlongster:
Nice. You are right; I always forget that interactions with the debugger are different than interactions with the content process, the latter having to go through an async process. But if we're just touching the debugger it should be sync.

Issue #800, opened 2016-09-21
Code Search - outline all matches


jasonLaster:
We currently only highlight the selected search result. We want to highlight the selected search result and draw an outline around the other results, somewhat like atom and sublime.

![screen shot 2016-09-21 at 4 11 23 pm](https://cloud.githubusercontent.com/assets/254562/18727219/b4d70758-8015-11e6-9fa0-a32012d3bd7d.png)

AndrewPomorski:
I can tackle it :)

gasolin:
As #716 mentioned, You can refer the gecko patch which use keyshort cut to catch Codemirror related shortcuts https://reviewboard.mozilla.org/r/80700/diff/2#index_header

clarkbw:
@AndrewPomorski how's it going?  Any questions?

andreicristianpetcu:
I like the current Firefox DevTools search. It is more like filter than search. Besides highlighting the current matches, it filters out lines not matching and with the Up/Down arrows you jump between matches.

bomsy:
is anyone still working on this? would love to take it if its free.

jasonLaster:
@bomsy it's yours!

jasonLaster:
This is done! Great job @bomsy!

PR #1091, opened 2016-11-03
Outline matches

bomsy:
Associated Issue: #800

### Summary of Changes

* Added new search overlay for outlining matches
* Added css for styling outlines

### Testing

* [x] passes `npm test`
* [x] passes `npm run lint`

### Screenshots/Videos 
outlines

### DIFF
--- a/public/js/lib/codemirror-mozilla.css
+++ b/public/js/lib/codemirror-mozilla.css
@@ -125,6 +125,14 @@ selector in floating-scrollbar-light.css across all platforms. */
   background-repeat: repeat-x;
 }
 
+.cm-selecting {
+  background: none;
+  border-width: 1px;
+  border-style: solid;
+  border-color: var(--theme-comment);
+  border-radius: 2px;
+}
+
 /* CodeMirror dialogs styling */
 
 .CodeMirror-dialog {
--- a/public/js/lib/codemirror-mozilla.css
+++ b/public/js/lib/codemirror-mozilla.css
@@ -127,10 +127,9 @@ selector in floating-scrollbar-light.css across all platforms. */
 
 .cm-selecting {
   background: none;
-  border-width: 1px;
-  border-style: solid;
-  border-color: var(--theme-comment);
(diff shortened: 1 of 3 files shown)


bomsy:
Was not sure about the theme color variable to use for selections.

public/js/utils/source-search.js
```diff
@@ -32,6 +33,31 @@ function getSearchCursor(cm, query, pos) {
     typeof query == "string" && query == query.toLowerCase());
 }
 
+function searchOverlay(query) {
+  query = query.replace(/[\-\[\]\/\{\}\(\)\*\+\?\.\\\^\$\|]/g, "\\$&");
```
jasonLaster: lol. this PR is crazy town, do you have some context on how you came up with it :)

public/js/lib/codemirror-mozilla.css
```diff
@@ -125,6 +125,14 @@ selector in floating-scrollbar-light.css across all platforms. */
   background-repeat: repeat-x;
 }
 
+.cm-selecting {
+  background: none;
+  border-width: 1px;
```
jasonLaster: this is a minor thing, but would you mind trying outline as well. I notice that border pushes the text, which is a slight visual tick that atom and i believe other editors avoid

public/js/utils/source-search.js
bomsy: Had a look into an example of how code mirror does highlighting of text. Basically Codemirror's `addOverlay` function takes a mode object. The mode object should have a token function which is applied as each token in the file is parsed. https://codemirror.net/doc/manual.html#option_mode

The regex just escapes  any special characters in the input

bomsy: cool .. will do!

bomsy:
Thanks!
We could definately replace it.

jasonLaster:
Awesome. Let's open a new PR to make it consistent. Also less scary

bomsy:
ok :)

Issue #1092, opened 2016-11-03
Source search throws error when a special character is searched


bomsy:
If user tries to search by entering some special characters eg * , (, \ nothing happens.

### STR
1) Open the debugger
2) Open a source code file with a comment line like `/* bla foo */`
3) `Ctrl + F` to open the source search
4) Type `*` or `\*`

### Current: 
Noting is selected or highlighted

### Expected
Should select or highlight one .

#### Also Note
Also if debbuger is open it throws errors

clarkbw:
I'm seeing this in the console:

```
SyntaxError: nothing to repeatbundle.js:75034:12	
	countMatches http://localhost:8000/public/build/bundle.js:75034:13
	onChange http://localhost:8000/public/build/bundle.js:75107:18
	wrapMethod/< http://localhost:8000/public/build/bundle.js:21646:17
	bound  self-hosted
	LinkedValueUtils.executeOnChange http://localhost:8000/public/build/bundle.js:13639:15
	_handleChange http://localhost:8000/public/build/bundle.js:13458:22
	bound _handleChange self-hosted
	bound bound _handleChange self-hosted
	ReactErrorUtils.invokeGuardedCallback http://localhost:8000/public/build/bundle.js:3541:8
	executeDispatch http://localhost:8000/public/build/bundle.js:3342:6
	executeDispatchesInOrder http://localhost:8000/public/build/bundle.js:3365:6
	executeDispatchesAndRelease http://localhost:8000/public/build/bundle.js:2797:6
	executeDispatchesAndReleaseTopLevel http://localhost:8000/public/build/bundle.js:2808:11
	forEach self-hosted
	forEachAccumulated http://localhost:8000/public/build/bundle.js:3744:6
	EventPluginHub.processEventQueue http://localhost:8000/public/build/bundle.js:3013:8
	runEventQueueInBatch http://localhost:8000/public/build/bundle.js:3773:4
	ReactEventEmitterMixin.handleTopLevel http://localhost:8000/public/build/bundle.js:3789:6
	handleTopLevelWithoutPath http://localhost:8000/public/build/bundle.js:15670:6
	handleTopLevelImpl http://localhost:8000/public/build/bundle.js:15650:4
	Mixin.perform http://localhost:8000/public/build/bundle.js:6110:14
	ReactDefaultBatchingStrategy.batchedUpdates http://localhost:8000/public/build/bundle.js:11664:8
[...]

PR #1093, opened 2016-11-03
Escape EditorSearchBar queries before using as RegExp

Zacqary:
Associated Issue: #1092 

### Summary of Changes

* Runs `escapeRegExp` on `query` passed to `countMatches`. Searching for something like `val(` would previously throw an error for an unterminated parenthetical in a regular expression.

### Testing

* [x] passes `npm test`
* [x] passes `npm run lint`

### DIFF
--- a/public/js/components/EditorSearchBar.js
+++ b/public/js/components/EditorSearchBar.js
@@ -4,12 +4,12 @@ const { findDOMNode } = require("react-dom");
 const Svg = require("./utils/Svg");
 const { find, findNext, findPrev } = require("../utils/source-search");
 const classnames = require("classnames");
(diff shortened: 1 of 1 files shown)


Zacqary:
I was trying to find instances of jQuery `val()` calls, so I typed `val(` and hit enter, and suddenly the box changed to just `val`. Fortunately @bomsy had stumbled on the same thing 2 hours ago and @clarkbw had found the source of the problem, plus a Stack Overflow thread where a comment mentioned a Lodash function, so I figured it'd be best to use that since it was already in the codebase.

So I can't take all the credit for this. :P

clarkbw:
Awesome! 🎉

PR #1094, opened 2016-11-03
Consistently return Seq from getFrames

Zacqary:
Associated Issue: #1056 

### Summary of Changes

* Consistently passes `frames` as type `Seq`, even when there are 0 frames. 

Previously, 0 frames would be stored in state as `null` and returned from the selector as `[]`. This confusion between whether `frames` is an `Array` with a `length` or a `Seq` with a `size` was part of what caused the bug resulting in #1056. https://github.com/devtools-html/debugger.html/pull/1083 fixed the buggy behavior but it left the type inconsistency, so this PR corrects that.

### Testing

* [x] passes `npm test`
* [x] passes `npm run lint`

### DIFF
--- a/public/js/components/Frames.js
+++ b/public/js/components/Frames.js
@@ -29,7 +29,7 @@ function renderFrameLocation(frame) {
 
 const Frames = React.createClass({
   propTypes: {
-    frames: ImPropTypes.list.isRequired,
+    frames: ImPropTypes.list,
     selectedFrame: PropTypes.object,
     selectFrame: PropTypes.func.isRequired
   },
@@ -66,6 +66,9 @@ const Frames = React.createClass({
 
   renderFrames() {
     let { frames } = this.props;
+    if (!frames) {
+      return null;
+    }
 
     const numFramesToShow = this.state.showAllFrames
       ? frames.size : NUM_FRAMES_SHOWN;
@@ -92,7 +95,7 @@ const Frames = React.createClass({
   render() {
     const { frames } = this.props;
 
-    if (frames.size === 0) {
+    if (!frames) {
       return div(
         { className: "pane frames" },
         div(
@@ -110,15 +113,20 @@ const Frames = React.createClass({
   }
 });
 
+function getAndProcessFrames(state) {
+  const frames = getFrames(state);
+  if (!frames) {
+    return null;
+  }
+  return frames.filter(frame => getSource(state, frame.location.sourceId))
+               .map(frame => Object.assign({}, frame, {
+                 source: getSource(state, frame.location.sourceId).toJS()
+               }));
+}
+
 module.exports = connect(
   state => ({
-    frames: getFrames(state)
-      .filter(frame => getSource(state, frame.location.sourceId))
-      .map(frame => {
-        return Object.assign({}, frame, {
-          source: getSource(state, frame.location.sourceId).toJS()
-        });
-      }),
+    frames: getAndProcessFrames(state),
     selectedFrame: getSelectedFrame(state)
   }),
   dispatch => bindActionCreators(actions, dispatch)
--- a/public/js/components/Frames.js
+++ b/public/js/components/Frames.js
@@ -92,7 +92,7 @@ const Frames = React.createClass({
   render() {
     const { frames } = this.props;
 
-    if (frames.length === 0) {
+    if (frames.size === 0) {
       return div(
         { className: "pane frames" },
         div(
--- a/public/js/reducers/pause.js
+++ b/public/js/reducers/pause.js
@@ -137,7 +137,9 @@ function getShouldIgnoreCaughtExceptions(state) {
 }
 
 function getFrames(state) {
-  return state.pause.get("frames") || [];
+  // frames should always be an Immutable Seq
+  // Return an empty Seq if the state is null
+  return state.pause.get("frames") || fromJS([]);
 }
 
 function getSelectedFrame(state) {
--- a/public/js/reducers/pause.js
+++ b/public/js/reducers/pause.js
@@ -137,9 +137,7 @@ function getShouldIgnoreCaughtExceptions(state) {
 }
 
 function getFrames(state) {
-  // frames should always be an Immutable Seq
-  // Return an empty Seq if the state is null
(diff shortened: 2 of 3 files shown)


public/js/reducers/pause.js
```diff
@@ -137,7 +137,9 @@ function getShouldIgnoreCaughtExceptions(state) {
 }
 
 function getFrames(state) {
-  return state.pause.get("frames") || [];
+  // frames should always be an Immutable Seq
+  // Return an empty Seq if the state is null
+  return state.pause.get("frames") || fromJS([]);
```
jasonLaster: Hmm I'm not sure if we do defaults like empty array in other selectors. 

What's the default otherwise, null? That might be more descriptive. 

@jlongster what do you think?

By the way, the best fix for me would be better immutable records so that we have a consistent API

Zacqary: Yeah, the default is `null`, so maybe it'd be better to just have the selector return the exact state and have the React view do a Boolean check for whether `frames` isn't `null`.

Zacqary:
Update: refactored this PR to instead return `null` from `getFrames` if it's `null`, and then have the React view handle the `null` type.

Unfortunately that seems to have caused a mochitest to fail:
```
The following tests failed:

1272 INFO TEST-UNEXPECTED-FAIL | devtools/client/debugger/new/test/mochitest/browser_dbg-pretty-print.js | Uncaught exception - at chrome://mochitests/content/browser/devtools/client/debugger/new/test/mochitest/head.js:205 - TypeError: dbg.win.cm.lineInfo(...) is null
```

Maybe this test was written expecting `frames` to be an empty array when there aren't any?

Although I don't know what `lineInfo` is, so I'm not sure how this relates.

jasonLaster:
Hmm. I can run the tests locally. We have docs for it in the docs dir, but it's kind of a pain to set up :)

jasonLaster:
Hey @Zacqary, I just investigated the pretty print test failure. It's a good thing we have tests.

This fixes a case where we were expecting frames to be an array or seq.

```diff
diff --git a/public/js/utils/pause.js b/public/js/utils/pause.js
index 1132cf3..74de13f 100644
--- a/public/js/utils/pause.js
+++ b/public/js/utils/pause.js
@@ -2,6 +2,10 @@ const { Frame } = require("../tcomb-types");
 const { getOriginalLocation } = require("./source-map");

 function updateFrameLocations(frames) {
+  if (!frames) {
+    return frames;
+  }
+
```

Here was the test failure:
```
1272 INFO TEST-UNEXPECTED-FAIL | devtools/client/debugger/new/test/mochitest/browser_dbg-pretty-print.js | Uncaught exception - at chrome://mochitests/content/browser/devtools/client/debugger/new/test/mochitest/head.js:205 - TypeError: dbg.win.cm.lineInfo(...) is null
```

The error was with this line not finding the correct line info 
screen shot 2016-11-04 at 9 47 03 am

The editor at this point was totally broke!
screen shot 2016-11-04 at 9 46 57 am

It looked this bad because we *actually* show the error message if something went wrong in the actions.

I grepped for a reference to frames in the actions directory and found [this reference](https://github.com/devtools-html/debugger.html/blob/master/public/js/actions/sources.js#L199-L210) in prettyPrint which is a dead give away.

Zacqary:
@jasonLaster Anything we need to do about this instance? https://github.com/devtools-html/debugger.html/blob/master/public/js/actions/pause.js#L39

```javascript
function paused(pauseInfo) {
  return async function({ dispatch, getState, client }) {
    let { frames, why } = pauseInfo;
    frames = await updateFrameLocations(frames);
    const frame = frames[0];

    dispatch(evaluateExpressions());
    dispatch({
      type: constants.PAUSED,
      pauseInfo: { why, frame },
      frames: frames,
      selectedFrameId: frame.id
    });
    dispatch(selectSource(frame.location.sourceId,
                          { line: frame.location.line }));
  };
}
```

I guess maybe `frames` will never be `null` in this case since `paused` never gets called unless there are, in fact, frames to pause on. But there's always the possibility for weird things going on. Not sure what we'd do here though. Just return out of the action creator without dispatching anything?

Zacqary:
Went with:
```javascript
  if (!frames) {
    return Promise.resolve(frames);
  }
```
just so that this function always returns a Promise.

public/js/components/Frames.js
```diff
@@ -66,6 +66,9 @@ const Frames = React.createClass({
 
   renderFrames() {
     let { frames } = this.props;
+    if (!frames) {
+      return null;
+    }
```
jasonLaster: hmm, i don't think we need this if we're only rendering frames when there *are* frames.

Zacqary: Even though it's never going to get called unless `frames` is truthy, I think it's still good to structure functions so that they could individually pass a fuzz test. But how about:
```javascript
renderFrames(frames = []) {
    const numFramesToShow =
      this.state.showAllFrames ? frames.size : NUM_FRAMES_SHOWN;
    const framesToShow =
      frames.slice(0, numFramesToShow).map(frame => this.renderFrame(frame));

    return dom.ul({}, framesToShow);
  },
```

This makes the function's intent more clear — decouples it from `this.props` and annotates the type of `frames` by giving it a default. Even though that default is probably never going to get used, at least we've made it clear what `frames` is supposed to be.

Zacqary: Oh, actually `frames` is a `Seq`, not an array.  I guess we can do without `frames = []`.

But what do you think of the rest of the function cleanup?

jasonLaster: I like it, i'm happy to merge in the check. It was a small nit, but you persuaded me it is good to be safe

jasonLaster: If you want to do some small cleanup here we can open a new small PR. 

I'd also be curious to see what adding flow types would look like in a component. We haven't done that yet

Zacqary: Oh, I just pushed the commit that refactored it to `renderFrames(frames)` about 10 seconds after you clicked Merge. Do you want me to PR that one too?

jasonLaster:
Sure. Sorry about that

PR #1103, opened 2016-11-04
Fix context menu properties

jasonLaster:
### Summary of Changes

* fix an id
* fix access keys (and tested access keys in panel)
* tabs touchup


jlongster:
Don't fully understand the access keys, but looks fine to me!

jasonLaster:
Access key is a local shortcut when the contextmenu is open. So, for instance typing "B" would add a breakpoint.

Issue #1097, opened 2016-11-04
[Conditional Breakpoints] clicking the marker should close the panel


jasonLaster:
Clicking on the marker when the panel is open should close the panel. Currently it breaks the panel :)

It breaks it because it attempts to removes a bp that doesn't exist.

![](http://g.recordit.co/VixObV7L2A.gif)

jasonLaster:
#1113

PR #1113, opened 2016-11-04
[conditional bps] cleanup cbs

jasonLaster:
Associated Issue: #1097

### Summary of Changes

lots of cleanups
* clicking on the gutter when the panel is open closes it 
* cleanup code for adding a conditional bp
* simplify cbPanel
* selecting "add bp" with an open panel closes the panel
* tests
...


jasonLaster:
@jlongster should be fixed.

public/js/test/mochitest/head.js
```diff
@@ -559,6 +563,7 @@ function pressKey(dbg, keyName) {
 
 function type(dbg, string) {
   string.split("").forEach(char => {
+    console.log("typing character", char)
```
jlongster: Leftover console.log

public/js/components/Editor.js
```diff
@@ -255,7 +260,7 @@ const Editor = React.createClass({
   },
 
   componentDidMount() {
-    this.cbPanels = {};
+    this.cbPanel;
```
jlongster: This line doesn't do anything

jasonLaster: haha oops, that should be `this.cbPanel = null`.

PR #1121, opened 2016-11-05
Refactor some functions in Frames.js, add Flow types

Zacqary:
### Summary of Changes

* Refactor some functions in Frames.js to be more type-safe and less coupled to `this.props`
* Add Flow type-checking to Frames.js

### Testing

* [x] passes `npm test`
* [x] passes `npm run lint`

### Additional Notes

Flow doesn't like `L10N` being a global. I had to redeclare it at the top of the file in order to get Flow to cooperate.

I feel like `L10N` ought to be `require`d just like everything else, especially if we need to do that in every file that we add Flow to.

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -8,3 +8,4 @@
 
 [libs]
 ./packages/devtools-config/flow-interface.js
+./public/js/global-types.js
--- a/packages/devtools-local-toolbox/src/index.js
+++ b/packages/devtools-local-toolbox/src/index.js
@@ -6,6 +6,7 @@ const { Provider } = require("react-redux");
 const { DevToolsUtils, AppConstants } = require("devtools-sham-modules");
 const { injectGlobals, debugGlobal } = require("./utils/debug");
 const { setConfig, isEnabled, getValue, isDevelopment } = require("devtools-config");
+const L10N = require("./utils/L10N");
 
 setConfig(DebuggerConfig);
 
@@ -29,8 +30,6 @@ if (process.env.TARGET !== "firefox-panel") {
     case "firebug": require("./lib/themes/firebug-theme.css"); break;
   }
   document.body.parentNode.classList.add(`theme-${theme}`);
-
-  window.L10N = require("./utils/L10N");
 }
 
 function initApp() {
@@ -122,5 +121,6 @@ module.exports = {
   bootstrap,
   renderRoot,
   debugGlobal,
-  client
+  client,
+  L10N
 };
--- a/public/js/global-types.js
+++ b/public/js/global-types.js
@@ -1,0 +1,1 @@
+declare var L10N: any;
--- a/src/components/Frames.js
+++ b/src/components/Frames.js
@@ -1,3 +1,4 @@
+// @flow
 const React = require("react");
 const { DOM: dom, PropTypes } = React;
 const { div } = dom;
@@ -8,6 +9,10 @@ const actions = require("../actions");
 const { endTruncateStr } = require("../utils/utils");
 const { getFilename } = require("../utils/source");
 const { getFrames, getSelectedFrame, getSource } = require("../selectors");
+const classNames = require("classnames");
+
+import type { List } from "immutable";
+import type { Frame } from "../types";
 
 if (typeof window == "object") {
   require("./Frames.css");
@@ -15,15 +20,15 @@ if (typeof window == "object") {
 
 const NUM_FRAMES_SHOWN = 7;
 
-function renderFrameTitle(frame) {
-  return div({ className: "title" }, endTruncateStr(frame.displayName, 40));
+function renderFrameTitle({ displayName }: Frame) {
+  return div({ className: "title" }, endTruncateStr(displayName, 40));
 }
 
-function renderFrameLocation(frame) {
-  const filename = getFilename(frame.source);
+function renderFrameLocation({ source, location }: Frame) {
+  const filename = getFilename(source);
   return div(
     { className: "location" },
-    `${filename}: ${frame.location.line}`
+    `${filename}: ${location.line}`
   );
 }
 
@@ -46,16 +51,14 @@ const Frames = React.createClass({
     });
   },
 
-  renderFrame(frame) {
+  renderFrame(frame: Frame) {
     const { selectedFrame, selectFrame } = this.props;
-
-    const selectedClass = (
-      selectedFrame && (selectedFrame.id === frame.id ? "selected" : "")
-    );
 
     return dom.li(
       { key: frame.id,
-        className: `frame ${selectedClass}`,
+        className: classNames("frame", {
+          "selected": selectedFrame && selectedFrame.id === frame.id
+        }),
         onMouseDown: () => selectFrame(frame),
         tabIndex: 0
       },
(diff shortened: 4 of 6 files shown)


jasonLaster:
Here's the test error 

```
1317 INFO TEST-UNEXPECTED-FAIL | devtools/client/debugger/new/test/mochitest/browser_dbg-call-stack.js | Uncaught exception - at chrome://mochitests/content/browser/devtools/client/debugger/new/test/mochitest/browser_dbg-call-stack.js:25 - TypeError: findElement(...) is null

Stack trace:

    @chrome://mochitests/content/browser/devtools/client/debugger/new/test/mochitest/browser_dbg-call-stack.js:25:21

    Tester_execTest@chrome://mochikit/content/browser-test.js:733:9

    Tester.prototype.nextTest</<@chrome://mochikit/content/browser-test.js:653:7

    SimpleTest.waitForFocus/waitForFocusInner/focusedOrLoaded/<@chrome://mochikit/content/tests/SimpleTest/SimpleTest.js:743:59

    Tester_execTest@chrome://mochikit/content/browser-test.js:733:9

    Tester.prototype.nextTest</<@chrome://mochikit/content/browser-test.js:653:7

    SimpleTest.waitForFocus/waitForFocusInner/focusedOrLoaded/<@chrome://mochikit/content/tests/SimpleTest/SimpleTest.js:743:59

```

I can run these tests locally and see what's up. Also, since you seem to breaking shit :) we do have this [doc](https://github.com/devtools-html/debugger.html/blob/master/docs/mochitests.md) for getting the tests running, which is pretty cool in some ways. Expect to wait some time for the repo to download and more time for it to build...

jasonLaster:
@Zacqary I updated the PR to add support for L10N being a global.

@jlongster I added a `global-types.js` file to declare types like L10N. Do you think this makes sense?

jasonLaster:
@jlongster this run looked strange:

https://circleci.com/gh/devtools-html/debugger.html/1587

`docker ./mochitest-run died unexpectedly` ???

public/js/components/Frames.js
```diff
@@ -8,22 +9,24 @@ const actions = require("../actions");
...
 
 if (typeof window == "object") {
   require("./Frames.css");
 }
 
 const NUM_FRAMES_SHOWN = 7;
 
-function renderFrameTitle(frame) {
-  return div({ className: "title" }, endTruncateStr(frame.displayName, 40));
+function renderFrameTitle({ displayName }: Object) {
```
jasonLaster: hmm, @Zacqary I think this style might be a bit cleaner

```js
function renderFrameTitle({ displayName: string }) {}
function renderFrameLocation({ source: Source, location: Location }) {}
```

with these imports:

```js
import type { Source, Location } from "../types";
```

public/js/components/Frames.js
```diff
@@ -46,16 +49,14 @@ const Frames = React.createClass({
     });
   },
 
-  renderFrame(frame) {
+  renderFrame(frame: { id: number }) {
```
jasonLaster: hmm, mind adding a Frame type to types?

public/js/components/Frames.js
```diff
@@ -64,21 +65,16 @@ const Frames = React.createClass({
...
-  renderFrames() {
-    let { frames } = this.props;
-    if (!frames) {
-      return null;
-    }
-
-    const numFramesToShow = this.state.showAllFrames
-      ? frames.size : NUM_FRAMES_SHOWN;
-    frames = frames.slice(0, numFramesToShow);
+  renderFrames(frames: List<Object>) {
```
jasonLaster: This could be `List<Frame>` if we had a frame type

public/js/components/Frames.js
```diff
@@ -64,21 +65,16 @@ const Frames = React.createClass({
...
+    const numFramesToShow =
+      this.state.showAllFrames ? frames.size : NUM_FRAMES_SHOWN;
+    const framesToShow =
+      frames.slice(0, numFramesToShow).map(frame => this.renderFrame(frame));
```
jasonLaster: I kinda like breaking this up:

```js
const framesToShow = frames.slice(0, numFramesToShow))
dom.ul({}, framesToShow.map(this.renderFrame);
```

public/js/components/Frames.js
Zacqary: This doesn't work. https://github.com/facebook/flow/issues/183 has been tracking broken support for destructuring in Flow for a while.

The problem is `{ source: Source}` transpiles to `var Source = object.source`.

Instead, I tried:
```javascript
function renderFrameTitle({ displayName }: { displayName: string }) {
```

Unfortunately, this doesn't typecheck:
```
public/js/components/Frames.js:67
 67:       renderFrameTitle(frame),
           ^^^^^^^^^^^^^^^^^^^^^^^ function call
 25: function renderFrameTitle({ displayName }: { displayName: string }) {
                                                ^^^^^^^^^^^^^^^^^^^^^^^ property `displayName`. Property not found in
 67:       renderFrameTitle(frame),
                            ^^^^^ object type
```

I'll see what I can figure out by creating a `frame` type.

Zacqary: Got that to work when `Frame` is a type, but now the type declaration is really long:
```javascript
function renderFrameLocation({ source, location }: {source: Source, location: Location}) {
```
Ultimately I'm not sure it's worth the additional annotations. But at least `{ source, location}: Frame` is more descriptive.

jasonLaster: Yea, you can define the type above the function and then use it. Perhaps that's nice... Or just add a line break... Also, I'm not convinced we need it. Your call.

jasonLaster:
Sorry for the slow reply, I was traveling the past two days and was pretty swamped. This cleanup with the additional types looks awesome. And the component types help a lot!

I'll take a look at the two failing mochitests on monday and will report back. Thanks for putting in the hard work here, this is an awesome win for types in the codebase.

Also, for what it's worth, when I have some time I hope to hack on a [local test runner](https://github.com/devtools-html/debugger.html/issues/678) that uses the websocket connection. My hope is that, this setup will make it super easy for anyone to run the integration tests locally. The [integration tests](https://github.com/devtools-html/debugger.html/blob/master/public/js/test/mochitest/browser_dbg-call-stack.js#L40-L62) are actually really fun to work on, but are currently require some setup.

Zacqary:
Cool, thanks. I tried to get the Mochitests running locally but apparently I need to upgrade to Sierra to update XCode to download Mercurial to blah blah blah etc. etc. etc. so it might be a while.

jasonLaster:
Sorry for being slow on this @Zacqary. I'll have a chance tomorrow morning to look into those tests!

jasonLaster:
Hey @Zacqary I just rebased the work on top of some of the recent PRs that touched the types and some other call stack changes. Everything looks good, but would you mind reading it over once more.

@jlongster I'm not sure how to track this down, but I'm seeing `browser_dbg-sources.js` pass when I run it by itself, but fail when it's run as part of the entire suite. It looks like when it's run along side other tests, the sources don't load. I also saw that the [browser.ini](https://github.com/devtools-html/debugger.html/blob/master/src/test/mochitest/browser.ini#L57-L60) has a skip clause. Any ideas why this might be happening?

Zacqary:
Looks good to me if it looks good to you

jasonLaster:
> re browser_dbg-sources.js

I'm finding the flakiness on master, so I don't think it's related to this PR. If it is, we can investigate / revert.

Issue #971, opened 2016-10-20
Sources Search - polish keyboard interaction


jasonLaster:
We have two weird keyboard interactions with the sources search UI.
1. tabbing
2. entering

Tabbing - we should not let tabbing focus on other elements. When the user tabs, we should close the search box

Entering - 
1. when there is no text. close the search box
2. when there is a selection, but no match. close the search box, when the search box is re-opened show the text.

``` diff
diff --git a/public/js/components/Autocomplete.js b/public/js/components/Autocomplete.js
index 75e3cf2..0b11bce 100644
--- a/public/js/components/Autocomplete.js
+++ b/public/js/components/Autocomplete.js
@@ -102,8 +102,22 @@ const Autocomplete = React.createClass({
       });
       e.preventDefault();
     } else if (e.key === "Enter") {
+
+      // i assume -1 means we didn't find anything...
+      if (this.state.selectedIndex == -1) {
+        // close the search UI
+        // because the App listens for cmd+p we can probably fire an event on the windw :)
+        window.triggerEvent()
+        // we also want to save the input (we have inputValue so we're probably fine)
+      }
+
+      if (this.state.inputValue == "") {
+        // close the search UI
+      }
+
       this.props.selectItem(searchResults[this.state.selectedIndex]);
       e.preventDefault();
+
     }
   },
```

kareniel:
I can take care of this tonight. :)

PR #1123, opened 2016-11-06
Handle 'tab' and 'enter' key press in file search input field

kareniel:
Associated Issue: #971 

### Summary of Changes

This commit adds conditions that handle the cases in which
a user presses the 'tab' or the 'enter' key while the file search
input field (autocomplete component) is focused.

- modify case for 'enter' key in Autocomplete's onKeyDown method
- add case for 'tab' key in Autocomplete's onKeyDown() method
- add previousInput:string field to 'ui' state store
- add getFileSearchPreviousInput getter to 'ui' state store
- add saveFileSearchInput action creator in 'ui' actions
- add corresponding constant
- add case for that constant in 'ui' reducers
- pass saveFileSearchInput to SourceSearch's props
- write test for saveFileSearchInput action creator
- set cursor at the end of the input field's content

### Testing

* [x] passes `npm test`
* [x] passes `npm run lint` ( this commit doesn't add any additional rule breaks )


jasonLaster:
Thanks @kareniel. This was a fantastic first PR. 

One question, what are the trade-offs between having the search input in the redux store or source search state? Are there advantages to having it in the store that you wouldn't get in component state?

Also, noticed that there is one conflict. Could you quickly rebase your branch off of an uptodate master?

kareniel:
@jasonLaster : Actually, that piece of state could probably live in the `source search` component.  I only see a case for this if we think that component will get destroyed. I assumed this was the case, because of the existence of `searchOn` (and it being the only variable stored in the `ui` store).

jasonLaster:
> that piece of state could probably live in the source search

Yup, I agree. We added `searchOn` to the UI state because the "+" button in the source tabs row needs to open source search with a redux action. We're not worried about the component unmounting.

> I updated the history 

that's fine. I can squash these commits when I merge it in.

kareniel:
Gotcha. I can probably get this done later today.

public/js/reducers/ui.js
```diff
@@ -12,18 +12,23 @@ import type { Action } from "../actions/types";
...
+  searchOn: false,
+  previousInput: ""
 } : UIState));
 
 function update(state = State(), action: Action): Record<UIState> {
   switch (action.type) {
     case constants.TOGGLE_FILE_SEARCH: {
       return state.set("searchOn", action.searchOn);
     }
+    case constants.SAVE_FILE_SEARCH_INPUT: {
```
jasonLaster: oops, this stuff needs to go

public/js/components/SourceSearch.js
```diff
@@ -83,17 +91,20 @@ const Search = React.createClass({
       Autocomplete({
         selectItem: result => {
           this.props.selectSource(result.id);
+          this.setState({ previousInput: "" });
           this.props.toggleFileSearch(false);
         },
         handleClose: this.close,
-        items: searchResults(this.props.sources)
+        items: searchResults(this.props.sources),
+        previousInput: this.state.previousInput
```
jasonLaster: could we call this `searchValue` or `inputValue`? Sometimes with this stuff, i wonder if we need the tense for instance does "previous" make it more clear or is "current" redundant... The answer could well be, previous helps for X reason, i just haven't thought about it much.

Issue #1127, opened 2016-11-06
consolidate getting started documentation


afeld:
The instructions for getting started seem to exist a few places:

* https://github.com/devtools-html/debugger.html#getting-started
* https://github.com/devtools-html/debugger.html/blob/master/CONTRIBUTING.md
* https://github.com/devtools-html/debugger.html/blob/master/docs/local-development.md
* https://docs.google.com/document/d/1D1RYP4ImTtY2_STOTZ92Ozb-b4c-gr4MvGcR8E4yjdM/edit#

Would be great to have a single point of reference, which will also help to avoid drift between the different sets of instructions.

jasonLaster:
This is smilar #1233

jasonLaster:
@afeld just landed #1334 which should help in this department!

PR #1129, opened 2016-11-06
simplify getting started

afeld:
Associated Issue: #1127

### Summary of Changes

There are a few commits, which are somewhat unrelated, but build on one another:

1. 6f0f7c8: Update `yarn.lock` file. This is the result of running `yarn install` on `master` using yarn 0.16.1, without any other changes. I guess the last person to update dependencies hadn't committed the updates to the lock file, or something changed in yarn...?
1. 3a3217f: Remove duplicate getting started instructions
1. e5db7b7: Use a single command to start Firefox and the debugger process, rather than needing two terminals. This has the additional advantage of interleaving the logs from both.

### Testing

* [x] passes `npm test`
* [x] passes `npm run lint`

screen shot 2016-11-06 at 2 56 17 pm

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -65,24 +65,26 @@ The debugger.html is a web application that makes a [WebSocket](https://develope
 
 ### Web Application
 
-First we need to get the web application running.  Within the source code directory, from the command line run these commands.
+First we need to get the web application running.  Within the source code directory, run these commands from the command line.
 
-### Linux or MacOs
+### Install
+
+#### Linux or MacOs
 
 * `npm install` - Install dependencies
-* `npm start` - Start development web server
 
-### Windows
+#### Windows
 
 It is recommended to use Git Shell which comes with [GitHub Desktop] application to emulate bash on Windows.
 
 * `npm install --ignore-scripts` - Install dependencies
 * `bash ./bin/preinstall` - Run preinstall script manually
-* `npm start` - Start development web server
 
-Then, because `npm start` will remain running, from another terminal window you can open [http://localhost:8000](http://localhost:8000) in your browser or type the following:
+### Run
 
-* `open http://localhost:8000` - Open in any modern browser
+* `npm run fast-start` - Start development web server
+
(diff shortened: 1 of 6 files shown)


jasonLaster:
Thanks @afeld! These are great contributions that will help with the getting started process.

This is probably a little bit more overhead, but in the future it would be helpful to make three separate PRs as each one requires a slightly different review cycle and I'd hate for one of the three to block the other two.

1. yarn.lock - looks great (i rm'd my node_modules) and everything worked)
2. canonical getting started guides. I like this idea, it's probably better to link to one really good doc. @clarkbw what do you think? 
3. single command
   * I like the command, but I'm not sure if i'm ready to tell people to use it yet. I'd like to try it out myself for a bit. Also, I want `npm run firefox` to be more bullet proof first. 
   * I like the style cleanups
   * Could you remove the reference to it in the doc.

Garbee:
I think the primary necessity from this PR currently is the documentation changes. We do need to look over everything with fresh-eyes in the point-of-view of someone who knows nothing about the project to see where we can optimize the on-boarding path. At this time though, I'd warn against introducing new tools (such as Foreman in this PR) until we have some tangible feedback from multiple people that they are having a hard time getting things going after the docs are cleaned up.

Foreman in this instance is not only adding a new dependency but also bringing along yet-another top-level file. At some point it gets confusing to even go through them much less know what most of them are for.

I'd rather we start on the documentation end here and make that clearer. Then we can listen to feedback on the process and adjust the commands as we go. If we end up hearing a lot that once things are fixed up more getting everything running is still an issue, then we can give running two npm scripts in parallel the due diligence on having that done internally for a quick-start command.

jasonLaster:
Thanks @Garbee, I think that's a persuasive case.

Lets take Aiden's recommendation and consolidate the getting started guide. Many people want one guide on the steps to get started. 

Once that lands, we can re-evaluate foreman. We've landed lots of fixes to `npm run firefox` in the past week and i'm now confident it will work on all computers and in any order (starting firefox first or second)... Lets let this bake and then re-evaluate foreman.

Issue #1109, opened 2016-11-04
[Context Menu] Add seperators


jasonLaster:
Our context menu polyfill does not currently support separators, but it would be pretty easy to add support for it.

#### patch

here's a quick hack i put together. It looks like the only thing *you* really need to do is add some decent CSS, which i didn't really try to do because it was a hack.

```diff
diff --git a/public/js/components/Editor.js b/public/js/components/Editor.js
index 03a3ec4..963f41a 100644
--- a/public/js/components/Editor.js
+++ b/public/js/components/Editor.js
@@ -237,6 +237,10 @@ const Editor = React.createClass({
       click: () => this.toggleBreakpoint(line)
     };

+    const sep = {
+      type: "separator"
+    };
+
     const conditionalBreakpoint = {
       id: "node-menu-conditional-breakpoint",
       label: cbLabel,
@@ -247,6 +251,7 @@ const Editor = React.createClass({

     showMenu(e, [
       toggleBreakpoint,
+      sep,
       conditionalBreakpoint
     ]);
   },
diff --git a/public/js/components/menu.css b/public/js/components/menu.css
index e0112ff..fdda19c 100644
--- a/public/js/components/menu.css
+++ b/public/js/components/menu.css
@@ -31,3 +31,11 @@ menuitem:hover {
   height: 100%;
   z-index: 999;
 }
+
+menuseparator {
+  border-bottom: 1px solid #ccc;
+  width: 100%;
+  height: 5px;
+  display: block;
+  margin-bottom: 5px;
+}
 }
```

### screenshot
![screen shot 2016-11-04 at 12 15 02 pm](https://cloud.githubusercontent.com/assets/254562/20013003/f31a2a10-a287-11e6-88dd-a78c7a09916b.png)

gavlu:
Working on this.

PR #1132, opened 2016-11-06
Add separators to the debugger context menu.

gavlu:
Associated Issue: #1109 

@jasonLaster 

### Summary of Changes
- Added the separator and have it automatically be loaded into context menus.
- Added CSS for the separators in the debugger context menu.

nov-06-2016 15-04-58


jasonLaster:
Thanks for the patch @gavlu. I just dropped the separator from the Editor component, so this should be good to go.

Please share your context menu styles work when you can. It's fine if you can't finish it up, but i don't want to lose it.

Issue #1104, opened 2016-11-04
[Source Tabs] Add additional close items


jasonLaster:
We recently added the source tabs context menu, which has two close items "close tab" and "close all". It'd be great to add the additional common close items for "close other tabs" and "close tabs to the right". I attached a screenshot below for where this is done.

Here's a patch to get started [patch](https://gist.github.com/jasonLaster/54c329896793a6b7f34fdfa121d6fcfb)

![screen shot 2016-11-04 at 10 25 22 am](https://cloud.githubusercontent.com/assets/254562/20008968/006ca846-a279-11e6-81b7-7e22096d6da5.png)

bomsy:
fixing this!

charltonaustin:
How's that coming?

PR #1134, opened 2016-11-06
Additional close items

bomsy:
Associated Issue: #1104

### Summary of Changes

* Add new properties
* Added new closeOtherTabs and closeTabsToRight

### Testing

* [x] passes `npm test`
* [x] passes `npm run lint`

### Screenshots/Videos 
addcloseitems

### DIFF
--- a/public/debugger.properties
+++ b/public/debugger.properties
@@ -207,6 +207,14 @@ editor.addConditionalBreakpoint=Add Conditional Breakpoint
 # for closing the selected tab below the mouse.
 sourceTabs.closeTab=Close tab
 
+# LOCALIZATION NOTE (sourceTabs.closeOtherTabs): Editor source tab context menu item
+# for closing the other tabs.
+sourceTabs.closeOtherTabs=Close others
+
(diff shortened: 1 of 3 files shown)


jasonLaster:
This looks great @bomsy.

I was thinking about it last night and I think an improvement would be to improve the close action or add a new action named closeTabs that would know how to close many tabs. The advantage is that now we fire off N actions for closing each tab.

Perhaps we make a new issue for this.

bomsy:
Cool. I think thats a great idea.
I'll create the issue and a PR.

Thanks!

bomsy:
Also i had a thought.
Do we want to hide or disable (grey out) close items that are may not be useful.
e.g If there's only one tab , "Close other tabs" should be disabled, cos its of no use.
Your thoughts?

jasonLaster:
> Do we want to hide or disable close items

@bomsy I think that's a great idea. I'd propose hiding those items if there's only one tab.
just created this https://github.com/devtools-html/debugger.html/issues/1139. mind snagging it?

bomsy:
excellent!  will do!

Issue #1139, opened 2016-11-07
[Source Tabs] Hide batch close items when there's one tab


jasonLaster:
There are a couple of items in the source tab context menu related to closing multiple tabs. We should only show those items when there are multiple tabs open!

jasonLaster:
Also, don't show "close to the right" if you're selecting the right most one :)

bomsy:
I'm getting this fixed!

clarkbw:
This is fixed now, right?

bomsy:
yes it is!

PR #1140, opened 2016-11-07
Hide close items

bomsy:
Associated Issue: #1139

### Summary of Changes

* Added new buildMenu function and tests
* Use the buildMenu function to generate the list of menuitem that should be visible

### Testing

* [x] passes `npm test`
* [x] passes `npm run lint`

### Screenshots/Videos
hide-close

### DIFF
--- a/public/js/components/SourceTabs.js
+++ b/public/js/components/SourceTabs.js
@@ -15,7 +15,7 @@ const { isEnabled } = require("devtools-config");
 const CloseButton = require("./CloseButton");
 const Svg = require("./utils/Svg");
 const Dropdown = React.createFactory(require("./Dropdown"));
-const { showMenu } = require("../utils/menu");
+const { showMenu, buildMenu } = require("../utils/menu");
 
 require("./SourceTabs.css");
 require("./Dropdown.css");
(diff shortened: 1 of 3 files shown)


public/js/utils/tests/menu.js
```diff
@@ -0,0 +1,14 @@
...
+
+describe("menu", () => {
+  it("should return an array of visible menu items", () => {
+    let menuItems = [
+      { item: "bla", hidden: false },
+      { item: "foo", hidden: true },
+      { item: "baa", hidden: () => true },
+      { item: "foobar" }
+    ];
+    expect(buildMenu(menuItems)[1]).to.be("foobar");
```
jasonLaster: Hmm lets check yah there are three items

public/js/components/SourceTabs.js
```diff
@@ -125,12 +125,13 @@ const SourceTabs = React.createClass({
...
-      closeTabMenuItem,
-      closeOtherTabsMenuItem,
-      closeTabsToRightMenuItem,
-      closeAllTabsMenuItem
-    ]);
+    showMenu(e, buildMenu([
+      { item: closeTabMenuItem },
+      { item: closeOtherTabsMenuItem, hidden: () => tabs.size === 1 },
+      { item: closeTabsToRightMenuItem, hidden: () =>
+         tabs.some((t, i) => t === tab && (tabs.size - 1) === i) },
```
jasonLaster: Why do we need some?

bomsy: Some returns a true if there's a truthy for any of the array values. i wanted a clean way of return a truthy if we hit the current tab and its the last one in the array.
Any better ideas?

jasonLaster: Hmm I'm missing something. 

What do you mean by truthy in the array?

What do you mean by truthy if the current tax is the last one :)

bomsy: so for every loop over the tabs items, i'm checking two things

1. `t === tab`. If the current iteration tab is the selected one.

2.  `(tabs.size - 1) === i)` . if the tab item is the last one in the array 

 If these 2 conditions are satisfied , a `true` is returned to `some` meaning that this item does not have tab items to its right.

bomsy: :) hope this helps explain my thoughts behind this?

PR #1144, opened 2016-11-07
Update yarn lockfile

jasonLaster:
I borrowed Aidan's idea in #1129 to update the lockfile, which we should try to better to keep uptodate.

I think there are three things we can do to be better about this:
1. always use yarn install when adding packages
2. check to see if the yarn file is updated when the package.json is updated
3. pull down greenkeeper PRs and run yarn install. This is temporary, they'll have yarn support eventually.


jlongster:
Not sure about greenkeeper anymore, if they don't support yarn. I don't think we're getting enough value out of it. I say we turn it off, but let's talk about it when the 3 of us meet again.

jasonLaster:
@jlongster yeah, i hear ya. I feel like at best it's just a ping... I could go either way.

PR #1152, opened 2016-11-08
Add mochitest for source tree (fixes #648)

### LINKED ISSUE #648: source listing [mochitests]

> review these and likely convert to unit tests

browser_dbg_sources-contextmenu-01.js
browser_dbg_sources-contextmenu-02.js
browser_dbg_sources-eval-01.js
browser_dbg_sources-eval-02.js
browser_dbg_sources-iframe-reload.js
browser_dbg_sources-keybindings.js
browser_dbg_sources-labels.js
browser_dbg_sources-large.js
browser_dbg_sources-sorting.js
browser_dbg_sources-webext-contentscript.js

[DXR search](https://dxr.mozilla.org/mozilla-central/search?q=path%3Adevtools%2Fclient%2Fdebugger%2Ftest%2Fmochitest%2F+file%3A*dbg_source-maps*&redirect=true)

jlongster:
Been trying to do this for a while, sorry. Kept getting distracted, and took a while to decide the right way to do this. This covers the core cases from the previous mochitests. This doesn't test the specific nesting, but we have a lot of unit tests for that kind of stuff.

### DIFF
--- a/public/js/test/mochitest/browser_dbg-sources.js
+++ b/public/js/test/mochitest/browser_dbg-sources.js
@@ -1,0 +1,58 @@
+/* Any copyright is dedicated to the Public Domain.
+ * http://creativecommons.org/publicdomain/zero/1.0/ */
+
+// Tests that the source tree works.
+
+function* waitForSourceCount(dbg, i) {
+  // We are forced to wait until the DOM nodes appear because the
+  // source tree batches its rendering.
+  yield waitUntil(() => {
+    return findAllElements(dbg, "sourceNodes").length === i;
+  });
+}
+
+add_task(function* () {
+  const dbg = yield initDebugger("doc-sources.html");
+  const { selectors: { getSelectedSource }, getState } = dbg;
+
+  // Expand nodes and make sure more sources appear.
+  is(findAllElements(dbg, "sourceNodes").length, 2);
+
+  clickElement(dbg, "sourceArrow", 2);
+  is(findAllElements(dbg, "sourceNodes").length, 7);
+
+  clickElement(dbg, "sourceArrow", 3);
+  is(findAllElements(dbg, "sourceNodes").length, 8);
+
+  // Select a source.
+  ok(!findElementWithSelector(dbg, ".sources-list .focused"),
+     "Source is not focused");
+  const selected = waitForDispatch(dbg, "SELECT_SOURCE");
+  clickElement(dbg, "sourceNode", 4);
+  yield selected;
+  ok(findElementWithSelector(dbg, ".sources-list .focused"),
+     "Source is focused");
+  ok(getSelectedSource(getState()).get("url").includes("nested-source.js"),
+     "The right source is selected");
+
+  // Make sure new sources appear in the list.
+  ContentTask.spawn(gBrowser.selectedBrowser, null, function() {
+    const script = content.document.createElement("script");
+    script.src = "math.min.js";
+    content.document.body.appendChild(script);
+  });
+
+  yield waitForSourceCount(dbg, 9);
+  is(findElement(dbg, "sourceNode", 7).querySelector("span").innerText,
+     "math.min.js",
+     "The dynamic script exists");
+
+  // Make sure named eval sources appear in the list.
+  ContentTask.spawn(gBrowser.selectedBrowser, null, function() {
+    content.eval("window.evaledFunc = function() {} //# sourceURL=evaled.js");
+  });
+  yield waitForSourceCount(dbg, 11);
+  is(findElement(dbg, "sourceNode", 2).querySelector("span").innerText,
+     "evaled.js",
+     "The eval script exists");
+});
--- a/public/js/test/mochitest/examples/doc-sources.html
+++ b/public/js/test/mochitest/examples/doc-sources.html
@@ -1,0 +1,23 @@
+<!-- Any copyright is dedicated to the Public Domain.
+     http://creativecommons.org/publicdomain/zero/1.0/ -->
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8"/>
+    <title>Debugger test page</title>
+  </head>
+ 
+  <body>
+    <script src="simple1.js"></script>
+    <script src="simple2.js"></script>
+    <script src="long.js"></script>
+    <script>
+      // This inline script allows this HTML page to show up as a
+      // source. It also needs to introduce a new global variable so
+      // it's not immediately garbage collected.
+      function inline_script() { var x = 5; }
+    </script>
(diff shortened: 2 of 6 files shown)


public/js/test/mochitest/browser_dbg-sources.js
```diff
@@ -0,0 +1,58 @@
...
+ * http://creativecommons.org/publicdomain/zero/1.0/ */
+
+// Tests that the source tree works.
+
+function* waitForSourceCount(dbg, i) {
+  // We are forced to wait until the DOM nodes appear because the
+  // source tree batches its rendering.
+  yield waitUntil(() => {
+    return findAllElements(dbg, "sourceNodes").length === i;
+  });
```
jasonLaster: [this](https://github.com/devtools-html/debugger.html/blob/master/public/js/test/mochitest/head.js#L183-L185) could help you here.

public/js/test/mochitest/browser_dbg-sources.js
```diff
@@ -0,0 +1,58 @@
...
+
+  clickElement(dbg, "sourceArrow", 3);
+  is(findAllElements(dbg, "sourceNodes").length, 8);
+
+  // Select a source.
+  ok(!findElementWithSelector(dbg, ".sources-list .focused"),
+     "Source is not focused");
+  const selected = waitForDispatch(dbg, "SELECT_SOURCE");
+  clickElement(dbg, "sourceNode", 4);
+  yield selected;
```
jasonLaster: this shouldn't change the test, but I find it easier to read:

```js
clickElement(dbg, "sourceNode", 4);
yield selected = waitForDispatch(dbg, "SELECT_SOURCE");
```

public/js/test/mochitest/browser_dbg-sources.js
```diff
@@ -0,0 +1,58 @@
...
+     "Source is focused");
+  ok(getSelectedSource(getState()).get("url").includes("nested-source.js"),
+     "The right source is selected");
+
+  // Make sure new sources appear in the list.
+  ContentTask.spawn(gBrowser.selectedBrowser, null, function() {
+    const script = content.document.createElement("script");
+    script.src = "math.min.js";
+    content.document.body.appendChild(script);
+  });
```
jasonLaster: can we move this to a function in content? The main benefit is that it lets us re-use `invokeTab` instead of `ContentTask.spawn` which is an API that can be implemented outside of mochitest. I also find it more readable and consistent with other tests like `doEval` where the test logic is inside of the test example file.


`head.js`
```js
invokeInTab("addScript", "math.min.js")
```

`doc-sources.html `
```js
window.addScript = function(name) {
  const script = content.document.createElement("script");
  script.src = name;
 document.body.appendChild(script);
}
```

public/js/test/mochitest/browser_dbg-sources.js
```diff
@@ -0,0 +1,58 @@
...
+
+  yield waitForSourceCount(dbg, 9);
+  is(findElement(dbg, "sourceNode", 7).querySelector("span").innerText,
+     "math.min.js",
+     "The dynamic script exists");
+
+  // Make sure named eval sources appear in the list.
+  ContentTask.spawn(gBrowser.selectedBrowser, null, function() {
```
jasonLaster: same thing here

public/js/test/mochitest/examples/doc-sources.html
```diff
@@ -0,0 +1,23 @@
...
+ 
+  <body>
+    <script src="simple1.js"></script>
+    <script src="simple2.js"></script>
+    <script src="long.js"></script>
+    <script>
+      // This inline script allows this HTML page to show up as a
+      // source. It also needs to introduce a new global variable so
+      // it's not immediately garbage collected.
+      function inline_script() { var x = 5; }
```
jasonLaster: ughhh

public/js/test/mochitest/browser_dbg-sources.js
jlongster: I can't use that because I need to wait until a specific number of nodes exists. I don't know which specific node to wait for. I hate it, but there's no away around the throttling of the tree rendering.

public/js/test/mochitest/browser_dbg-sources.js
jlongster: Not sure what that's doing, you are yielding an assignment?

Either way, it doesn't work because the action will be dispatched synchronously so I need to build the promise before firing the click. Usually I would just do `yield wait()` but it's like this for a reason.

public/js/test/mochitest/browser_dbg-sources.js
jlongster: Ditto my other comment as well. I'd rather keep this in here as it's the actual testing behavior.

public/js/test/mochitest/browser_dbg-sources.js
jasonLaster: > Here I think it's simple enough to just inline

I can see that too

> We'll have to figure out a lot of stuff if we ever run these elsewhere

Yup, that makes sense. I feel confident that we have a path forward at that point as well.

public/js/test/mochitest/browser_dbg-sources.js
jasonLaster: ahh, i see.

public/js/test/mochitest/browser_dbg-sources.js
jlongster: I think you meant `yield waitForDispatch(dbg, "SELECT_SOURCE");` in your original code ;)

I don't quite follow because I'm not dispatching anything here, I'm waiting for a dispatch. Oh, maybe you mean that if it is an async action, it would work to do something and then `waitForDispatch` right after because we don't care about the `"start"` action; our `waitForDispatch` will resolve with the `"done"` action which is guaranteed to be asynchronous. That's how it works most of the time.

`SELECT_SOURCE` is not an async action though. Not all `dispatch` calls return a promise. Only ones that fire an action creator that *themselves* return a promise do. `SELECT_SOURCE` is not async and is dispatched immediately in the `selectSource` action, which is fired synchronously when `clickElement` is called.

So we must have a handler for it waiting before `clickElement` is called. Otherwise, once `clickElement` returns the action is already dispatched and we've missed it.

This is a good thing to understand which is why I wanted to explain more. Once this is merged, feel free to try changing the code and watching it.

jasonLaster: Oh I see. Hadn't thought that through. Thanks for the writeup. 

Perhaps waiting on load source makes sense, but just thinking it through. Not looking for a cleanup :)

Issue #1102, opened 2016-11-04
[Context Menus] Improve styles


jasonLaster:
Context Menus are cool. In the panel, the debugger has special permissions to create it's own native context menus. In the web context, it's impossible to do this, so we create the context menu DOM and then add some [polyfills](https://github.com/devtools-html/debugger.html/blob/master/public/js/utils/menu.js) and [styling](https://github.com/devtools-html/debugger.html/blob/master/public/js/components/menu.css) to look mac native like.

The styling of course is not great and could use some tweaks. I included some screenshots comparing our menu to the native mac menu. This issue is about trying our best to *copy* the mac styles. Maybe someday we can copy all platforms styles, but that's out of scope :)

Some things that I noticed:
* the context menu has an opacity and I believe a [radial blur](http://stackoverflow.com/questions/25090012/os-x-yosemite-menu-background-blur-in-css)
* the drop shadow is off
* the background color is off
* the width + height could be tweaked
* line height + font size + font color can be tweaked

Disclaimer: I couldn't find a way to inspect the context menu in the inspector because any time I clicked on it it went away. It's a total [heisenbug](https://en.wikipedia.org/wiki/Heisenbug)

### our styles
![screen shot 2016-11-04 at 10 06 46 am](https://cloud.githubusercontent.com/assets/254562/20008211/1426a81c-a276-11e6-965d-ad659e623659.png)

### mac styles

![screen shot 2016-11-04 at 10 06 35 am](https://cloud.githubusercontent.com/assets/254562/20008234/2bda4806-a276-11e6-8495-3664f8a2ed9c.png)

### Our styles
![screen shot 2016-11-04 at 10 06 00 am](https://cloud.githubusercontent.com/assets/254562/20008205/0d0f164a-a276-11e6-9872-8c8526cd2411.png)

### mac os styles
![screen shot 2016-11-04 at 10 05 01 am](https://cloud.githubusercontent.com/assets/254562/20008201/06247910-a276-11e6-8e2c-2a1933eceacf.png)

jasonLaster:
Hey @gavlu, how's this going?

gavlu:
Hey, going to finish up and submit a PR later today.

PR #1155, opened 2016-11-09
Change debugger context menu styles to be closer to native OSX styling.

gavlu:
Associated Issue: #1102 

@jasonLaster 

### Summary of Changes
* Changed the background color of the context menu, drop shadow color, and `menuitem` hover color.
* Changed the padding of the `menuitem` class.
* Changed the font sizing for the context menu.

### Testing

* [x] passes `npm test`
* [x] passes `npm run lint`

### Screenshots/Videos
screen shot 2016-11-08 at 6 57 33 pm

### DIFF
--- a/public/js/components/App.css
+++ b/public/js/components/App.css
@@ -6,7 +6,7 @@
 
 :root.theme-light,
 :root .theme-light {
-  --theme-search-overlays-semitransparent: rgba(221, 225, 228, 0.66);
+  --theme-search-overlays-semitransparent: rgba(190, 190, 190, 0.8);
   --theme-faded-tab-color: #7e7e7e;
 }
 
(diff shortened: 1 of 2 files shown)


gavlu:
@jasonLaster

public/js/components/App.css
```diff
@@ -6,7 +6,7 @@
 
 :root.theme-light,
 :root .theme-light {
-  --theme-search-overlays-semitransparent: rgba(221, 225, 228, 0.66);
```
jasonLaster: hmm, i don't think we'll want to change this. It's a global that's shared in many places.

I think we can likely just use ` rgba(190, 190, 190, 0.8);` for the box-shadow. the menu is special in that it's only for the local environment and...

jasonLaster:
@gavlu this one small fix for App.css and it's ready to go

public/js/components/App.css
rheesneb: Not sure

PR #1148, opened 2016-11-08
[L10N] Fix hardcoded strings. Fixes #1143

Dalimil:

PR #1157, opened 2016-11-09
[L10N] Fix hardcoded strings. Fixes #1143

### LINKED ISSUE #1143: [L10N] Fix hardcoded strings
We should replace these hard-coded strings with L10N strings. Here are the [docs](https://github.com/devtools-html/debugger.html/blob/master/docs/local-development.md#internationalization) for L10N.

* There are hard-coded strings: "Loading..." when clicking a file, "No Breakpoints" in the sidebar.
* Several tooltips in the sidebar show %S, and there are hard-coded strings ("Disable breakpoints").

Dalimil:
Moved from https://github.com/devtools-html/debugger.html/pull/1148

### DIFF
--- a/public/js/strings.json
+++ b/public/js/strings.json
@@ -1,5 +1,6 @@
 {
   "breakpoints.header": "Breakpoints",
+  "breakpoints.none": "No Breakpoints",
   "callStack.header": "Call Stack",
   "callStack.notPaused": "Not Paused",
   "callStack.collapse": "Collapse Rows",
@@ -10,6 +11,8 @@
   "editor.removeBreakpoint": "Remove Breakpoint",
   "editor.editBreakpoint": "Edit Breakpoint",
   "editor.addConditionalBreakpoint": "Add Conditional Breakpoint",
+  "editor.conditionalPanel.placeholder": "This breakpoint will pause when the expression is true",
+  "expressions.placeholder": "Add Watch Expression",
   "scopes.header": "Scopes",
   "scopes.notAvailable": "Scopes Unavailable",
   "scopes.notPaused": "Not Paused",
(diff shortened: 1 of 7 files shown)


public/js/strings.json
```diff
@@ -10,6 +11,8 @@
   "editor.removeBreakpoint": "Remove Breakpoint",
   "editor.editBreakpoint": "Edit Breakpoint",
   "editor.addConditionalBreakpoint": "Add Conditional Breakpoint",
+  "editor.conditionalPanel.placeholder": "This breakpoint will pause when the expression is true",
+  "expressions.placeholder": "Add Watch Expression",
```
jasonLaster: looks like the tests are failing because `strings.json` needs `loadingText` from the properties.

It's interesting that it's using it... but i can investigate that later :)

Dalimil: OK, fixed now.

jlongster:
This should not have been merged today. I had to revert it in https://github.com/devtools-html/debugger.html/commit/fd8718c7faa2011faca4cc7da033173d24f6dccc.

Today I was in the process of pushing out an update to mozilla-central, and I already have a stable patch that has passed on try. We only merge emergency bugfixes that block the merge around times like that. Sorry, we don't have a great way to track that kind of thing yet.

Filed an issue in case we forget (https://github.com/devtools-html/debugger.html/issues/1166), but if you can, can you open a new PR?

jlongster:
I forgot to mention one crucial detail: I could have still generated the bundle from the same commit (which was before this commit) but I needed to land https://github.com/devtools-html/debugger.html/commit/8175aacaec380ecf859183ad62bee2a9aef180d2 as an emergency fix, which is why I needed to revert this so could have a new master to base the bundle on.

PR #1158, opened 2016-11-09
Fix dropdown button size

jlongster:
This is an emergency fix. I'm trying to push out a bundle update today, but the following UI is broken:

screen shot 2016-11-09 at 1 34 19 pm

The best I can do is at least make the dropdown button the same size as our other icons:

screen shot 2016-11-09 at 2 13 38 pm

The color is different than the new tab button, and the new tab button is not positioned correctly when tabs are hidden. But this is the best I can do for the release.

### DIFF
--- a/public/js/components/SourceTabs.js
+++ b/public/js/components/SourceTabs.js
@@ -171,21 +171,6 @@ const SourceTabs = React.createClass({
     }, filename);
   },
 
-  renderSourcesDropdownButton() {
-    const hiddenSourceTabs = this.state.hiddenSourceTabs;
-    if (!hiddenSourceTabs || hiddenSourceTabs.size == 0) {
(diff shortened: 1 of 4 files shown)


public/js/components/SourceTabs.js
```diff
@@ -171,21 +171,6 @@ const SourceTabs = React.createClass({
     }, filename);
   },
 
-  renderSourcesDropdownButton() {
```
clarkbw: This looks like it wasn't being called anywhere else and is just being removed here.

jlongster:
If someone wants to tweak how this works, we can do it as a follow-up, so I'm merging at least a temporary fix so I can try to land a bundle. We could probably land another one in a day or two if needed.

PR #1159, opened 2016-11-10
use flex to clean up search, fix overflow issue with close button

clarkbw:
Associated Issue: n/a

### Summary of Changes

* standardize the large close button for use in both source and file search
* move toward flex based model for both source and file search

### Testing

* [x] passes `npm test`
* [x] passes `npm run lint`

Here's the original problem I was seeing in Firefox

screen shot 2016-11-09 at 2 15 11 pm

And now we see this for results

screen shot 2016-11-09 at 10 00 35 pm

And this for empty states

screen shot 2016-11-09 at 3 15 45 pm

screen shot 2016-11-09 at 3 16 00 pm

Padding and spacing could use some additional followup PRs but this is a nice improvement from before and works on Chrome as well.

### DIFF
--- a/public/js/components/CloseButton.css
+++ b/public/js/components/CloseButton.css
@@ -27,9 +27,9 @@
 }
 
 .close-btn-big {
-  position: absolute;
-  top: 12px;
-  right: 10px;
+  padding: 13px;
+  width: 40px;
+  height: 40px;
 }
 
 .close-btn-big path {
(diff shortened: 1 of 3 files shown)


jlongster:
I'm confused, the result test is not even in the input area for me:

screen shot 2016-11-10 at 10 55 48 am

jlongster:
Oh, you're talking about searching with the file. Yep I see that too. Thanks for the fix.

jlongster:
How did this even happen? I'd like to know how such an obvious error can make it into master, especially right before we ride the trains. We need to make sure every PR actually works in Firefox and doesn't introduce stuff like this.

clarkbw:
Not sure, I'll look into it. I'm guessing we were reusing the close button component and some of the CSS can't be shared in that way because one assumes the autocomplete component

PR #1164, opened 2016-11-10
Disable custom code search because of bugs

jlongster:
The custom code search we have is far too buggy for us to enable it in dev edition. This PR disables it and Cmd+F will bring up the default code search in the devtools codemirror which works well.

Watch this movie: http://jlongster.com/s/editor-bugs.mp4

In that movie, I found the following bugs:

* Cmd-g and Cmd-shift-g don't work, an extremely command shortcut for find next and prev
* At 20s, the text isn't highlighted when you open it (happens if you click in the editor before opening it)
* The worst one, at 47s, is that you can still see the old file search somehow when clicking away from the search bar (if it's open) and either pressing escape or cmd+f (can't remember which one). Now pressing cmd+f again open the custom search bar and there are TWO open. Hilariousness ensues.

It's way too late to fix this and too risky. Let's just use the default one which works well.

### DIFF
--- a/packages/devtools-config/configs/development.json
+++ b/packages/devtools-config/configs/development.json
@@ -11,7 +11,8 @@
     "tabs": true,
     "sourceMaps": true,
     "prettyPrint": true,
-    "watchExpressions": false
+    "watchExpressions": false,
+    "editorSearch": false
   },
   "chrome": {
     "debug": true,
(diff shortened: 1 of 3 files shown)


jlongster:
> Could you wrap this in a feature flag? That'll make it easier to fix locally. And it's easier to follow.

Sure!

PR #1174, opened 2016-11-12
Fix main package eslint script to check package contents as well.

Garbee:
Associated Issue: None. Discussed problem directly with @jasonLaster in IRC.

### Summary of Changes

* JS Lint script in `package.json` checks the `packages` js as well as all js files within `public` instead of only direct children.
* Updated `.eslintignore` to ignore new things that aren't needed to be checked with this change.
* Fixed any eslint errors with the current files where appropriate.
* Also updated a file path in `devtools-config` to not have an extra folder slash.

### Testing

* [x] passes `npm test`
* [x] passes `npm run lint`

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,3 +1,4 @@
+public/build/*
 public/js/lib/tree.js
 public/js/utils/imstruct.js
 public/js/test/examples/**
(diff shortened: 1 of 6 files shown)


Garbee:
Test fail is due to the docker update timing out. Un-related to the code changes. Not sure if it can just get restarted (perhaps flalky?) or if someone needs to look into something deeper there.

jasonLaster:
Yup. Theres a rebuild button on the top right of circle. Not too worried.

jasonLaster:
Thanks, this is really helpful!

Issue #1004, opened 2016-10-25
[Flow] - type the pause actions + reducer


jasonLaster:
We currently type the sources and breakpoints reducers. It would be great to give the same treatment to pausing.

Here's an initial [patch](https://gist.github.com/jasonLaster/48a09ff185f13ec7fa0da30298a8603e). Run this command on a new branch to get started:

```
curl https://gist.githubusercontent.com/jasonLaster/48a09ff185f13ec7fa0da30298a8603e/raw/b740cc5ee55a698445ba5b19b5fc8cb0c77b1127/foo.diff | git apply
```

**how do i run flow?** `flow`

**how do i see a module's coverage?**  `flow coverage --color public/js/utils/source-map-worker.js` or `npm run flow-coverage`

montogeek:
Colorized diff https://gist.github.com/jasonLaster/48a09ff185f13ec7fa0da30298a8603e

PR #1179, opened 2016-11-13
Added Flow types to pause actions and reducer

montogeek:
Associated Issue: #1004 

### Summary of Changes

Added Flow types to pause actions
Added Flow types to pause reducers

### Testing

[x] passes `npm test`
[x] passes `npm run lint`

Notes:
Flow version: 0.35.0, I used the last one since it was not specified in `.flowconfig`, although 0.34.0 is specified in `package.json`, it took the chance to use the last version.

- There are errors related with the `fromJS` utility, since it is not Flow typed.
- Flow 0.35.0 started complaining about these lines https://github.com/devtools-html/debugger.html/blob/master/public/js/utils/utils.js#L207 and https://github.com/devtools-html/debugger.html/blob/master/packages/devtools-local-toolbox/public/js/utils/utils.js#L163 with the error: `call of method assign: Expected object instead of T`
- `pause` reducer (https://github.com/devtools-html/debugger.html/blob/master/public/js/reducers/pause.js#L41) is handling an `source` action (https://github.com/devtools-html/debugger.html/blob/master/public/js/actions/sources.js#L195)
- This constant/action does not longer exists: https://github.com/devtools-html/debugger.html/blob/master/public/js/reducers/pause.js#L56
- Unfortunately Flow does not support object computed keys in Object type declarations, thats why this is hardcoded: https://github.com/devtools-html/debugger.html/compare/master...montogeek:flow-type-pause-actions-reducer?expand=1#diff-195a73de407881358c8dfc0c5891f7feR114


jasonLaster:
This looks great @montogeek! Especially PauseState and ThunkArgs. I'll take a look monday morning. 

At first glance it looks like both you and @Zacqary added a frame [type](https://github.com/devtools-html/debugger.html/pull/1121)

jasonLaster:
@montogeek just rebased your PR

src/types.js
```diff
@@ -6,6 +6,34 @@
...
+}
+
+export type Pause = {
+  frames: Frame[],
+  why: Why
+}
+
+export type Expression = {
+  id: number,
+  input: any
```
wldcordeiro: Is the type of `input` really `any`? Or could we use one that is a bit more specific?

src/actions/pause.js
```diff
@@ -177,8 +183,8 @@ function selectFrame(frame) {
  * @memberof actions/pause
  * @static
  */
-function loadObjectProperties(grip) {
-  return ({ dispatch, client }) => {
+function loadObjectProperties(grip: any) {
```
wldcordeiro: Same question here, do we need to use `any` or can we be more specific?

jasonLaster:
Going through this now. There's some really good stuff here and I can see how the flow score went up!

tracking my thoughts as i go through it here

- **?** ---  There are errors related with the `fromJS` utility, since it is not Flow typed.

- **fixed** https://github.com/devtools-html/debugger.html/pull/1199 ---  Flow 0.35.0 started complaining about these lines https://github.com/devtools-html/debugger.html/blob/master/public/js/utils/utils.js#L207 and https://github.com/devtools-html/debugger.html/blob/master/packages/devtools-local-toolbox/public/js/utils/utils.js#L163 with the error: `call of method assign: Expected object instead of T`


- **?** -- `pause` reducer (https://github.com/devtools-html/debugger.html/blob/master/public/js/reducers/pause.js#L41) is handling an `source` action (https://github.com/devtools-html/debugger.html/blob/master/public/js/actions/sources.js#L195)


- **removed** ---  I'll verify that we don't need it in a bit [blame](https://github.com/montogeek/debugger.html/blame/flow-type-pause-actions-reducer/src/reducers/pause.js#L71-L77) --- This constant/action does not longer exists: https://github.com/devtools-html/debugger.html/blob/master/public/js/reducers/pause.js#L56

- **?** -- Unfortunately Flow does not support object computed keys in Object type declarations, thats why this is hardcoded: https://github.com/devtools-html/debugger.html/compare/master...montogeek:flow-type-pause-actions-reducer?expand=1#diff-195a73de407881358c8dfc0c5891f7feR114

src/types.js
montogeek: We can use one more specific :)

montogeek:
Thanks for the thoughtful review, I will continue working on it.

About the hardcoded string, it is this: `@@dispatch/promise`, it the name of the promise middleware https://github.com/devtools-html/debugger.html/blob/master/src/utils/redux/middleware/promise.js#L9

I tried to import it and use it as key, but it is not supported by Flow:
```
  ...
  [PROMISE]: T
```

jasonLaster:
> re: `hardcoded string`

@montogeek that's an interesting question. We should be handling it in Sources and breakpoints which is typed. @jlongster do you know the answer?

src/utils/pause.js
```diff
@@ -1,7 +1,10 @@
+// @flow
 const { Frame } = require("../tcomb-types");
 const { getOriginalLocation } = require("./source-map");
 
-function updateFrameLocations(frames) {
+import type { Frame as TFrame } from "../types";
+
+function updateFrameLocations(frames: TFrame[]): Promise<TFrame[]> {
```
jasonLaster: small nit, when there's a conflict we use `FrameType`

https://github.com/montogeek/debugger.html/blob/flow-type-pause-actions-reducer/src/actions/sources.js#L273-L275

montogeek:
@jasonLaster Thank you so much! Sorry for not catching those details!

jasonLaster:
Impossible to catch everything!

I double and triple checked that too

montogeek:
YAY! 💃

Issue #1168, opened 2016-11-11
npm run firefox doesn't set firefoxOptions on webdriver instance


kareniel:
selenium-webdriver was recently updated from 3.0.0 beta-x to 3.0.1 on npm. Upgrading to the latest version breaks the script behind `npm run firefox` so that the firefoxOptions don't get set on the webdriver instance.

I'm investigating.

(if you run into this, you can run `npm install selenium-webdriver@3.0.0-beta-3` to get `npm run firefox` to work)

You can find the 3.0.0-beta-3 -> 3.0.1 tags commit compare at:
https://github.com/SeleniumHQ/selenium/compare/c7b525db30f2b3ad43cdf7749918fb3a434322dc...1969d7507be7f3b55ae9d1c51c66cd39c1372280

jasonLaster:
Thanks. I noticed this as well, but wasnt 100% sure. Thanks for investigating

kareniel:
Turns out the latest version selenium-webdriver has a fix that sets the moz:firefoxOptions capability behind the scenes when it's given configuration options for the FirefoxDriver. ;)

jasonLaster:
Great. So just to confirm, can we update our package.json and lockfile and
be good.

Mind opening a PR?

PR #1180, opened 2016-11-13
Use options instead of capabilities to start firefox webdriver

kareniel:
Associated Issue: #1168

### Summary of Changes

This commit modifies the script behind `npm run firefox`
to set the webdriver's moz:firefoxOptions capabilities through
selenium's webdriver interface - namely the firefox.Options class.

### Testing

* [x] passes `npm test`
* [x] passes `npm run lint`


jasonLaster:
Hmm, interestingly. I think when greenkeeper bumped the versions here https://github.com/devtools-html/debugger.html/pull/1171. Other tests started using that package. Yarn will help here :)

jasonLaster:
* bumped with the uptodate eslint

Issue #1156, opened 2016-11-09
[Toolbox] Open the console with the local toolbox


jasonLaster:
This issue tracks the work for re-using the toolbox with the console.

I'm doing this work [here](https://github.com/jasonlaster/console.html) and in this [branch](https://github.com/devtools-html/debugger.html/compare/master...jasonLaster:console-html).

There are several sub-tasks:

- [x] use webpack to bundle the code
- [x] use the dev-server to serve the bundle and styles
- [x] use the toolbox to establish a websocket connection
- [x] add a development config for the landing page (title, connections (only show firefox))

Nice to haves:
- [ ] move the debugger's client API back to the debugger

jasonLaster:
Hey @juliandescottes, making progress

![](http://g.recordit.co/YMO6vrkkBX.gif)

jasonLaster:
This is now working well with [console.html](https://github.com/jasonlaster/console.html)

PR #1181, opened 2016-11-13
(WIP) [Toolbox] Decouple the development environment

jasonLaster:
Associated Issue: #1156 

This is some preliminary work needed to get other tools to work with the local toolbox. This work came out of hacking on [console.html](https://github.com/jasonLaster/console.html) while on the plane.

There are some high level pieces to look out for:
1. each tool has a `bin/dev-server` that uses the toolbox's `dev-server`
2. each tool has a `configs` directory
3. each tool has a `webpack.config` that uses the toolbox's config.
4. each tool is responsible for reading configs and then passing them along to services (server, tests, webpack config). This is a simpler system than before. 
5. the toolbox `index.js` and `client/index.js` is tool agnostic. This includes the console passing in a dom element as root (not a react component).
6. I added some `devtools-modules` needed for the console


bin/dev-server.js
```diff
@@ -0,0 +1,27 @@
+const fs = require("fs");
+const path = require("path");
+const serveIndex = require("serve-index");
+const express = require("express");
+
+const toolbox = require("../node_modules/devtools-local-toolbox/index");
```
jlongster: Why not just do `require("devtools-local-toolbox")` ?

configs/getConfig.js
```diff
@@ -0,0 +1,20 @@
+const _ = require("lodash");
```
jlongster: I think it's a little confusing for code and configs to share the same directory. I'm still getting my head around this new structure so I don't know what to propose yet, but ideally the top-level `config` directory would just contain json files.

Pomax: also when using lodash in node code, it generally makes more sense to call the variable itself `lodash`, not `_`, since it's much for informative to grep for a full word string like `lodash` than it is to grep for a single character `_` (especially with Node's `___filename` and `__dirname` constants)

package.json
```diff
@@ -10,8 +10,12 @@
     "url": "https://github.com/devtools-html/debugger.html/issues"
   },
   "homepage": "https://github.com/devtools-html/debugger.html#readme",
+  "engineStrict": true,
+  "engines": {
+    "node": ">=6.4.0"
```
Pomax: given the LTS status of 6.9, probably worth bumping this to `>=6.9.0`

configs/getConfig.js
jlongster: @Pomax I agree. I've advocated for limiting our use of lodash. I would suggest removing it from this file as well.

bin/dev-server.js
rheesneb: Help fix my duplicate account

configs/getConfig.js
jasonLaster: I think `config.js` in the top level would probably be fine. 

I'm hesitant to move it into `src` or `bin` as those are different in my mind.

What do you think?

bin/dev-server.js
jlongster: I guess the main thing was referencing `node_modules` directly though: you should be able to do `require("devtools-local-toolbox/index")` right?

configs/getConfig.js
jlongster: This is src code so I don't see why `src` wouldn't make sense?

jasonLaster: i'm okay using ES6 where we can, but in this case, we do need a deep merge, so lodash.merge makes sense to me

PR #1182, opened 2016-11-13
chore(stylelint): Lint package styles and address existing problems.

Garbee:
Associated Issue: None, discussed directly with @jasonLaster in IRC.

### Summary of Changes

* Updated css lint script to check over package files.
* Removed comment stylelinting since there are quite a few non-alignable things existing. We should document a standard structure on how/when to comment and then update the codebase's comments along with that.
* Updated all remaining issues within the files to pass the linting.

### Testing

* [x] passes `npm test`
* [x] passes `npm run lint`

### DIFF
--- a/.stylelintignore
+++ b/.stylelintignore
@@ -1,2 +1,3 @@
 src/components/SplitBox.css
 src/components/reps.css
+packages/devtools-local-toolbox/src/lib/themes/*.css
--- a/.stylelintrc
+++ b/.stylelintrc
@@ -15,11 +15,6 @@
     "color-hex-case": "lower",
     "color-hex-length": "long",
     "color-no-invalid-hex": true,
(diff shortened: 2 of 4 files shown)


jasonLaster:
This is awesome. Linting FTW

There are some m-c css files we don't want to change, namely `themes` `devtools-modules` like splitbox

Garbee:
Checked out the themes folder and the splitbox file from origin and added them to the stylintignore.

Should we ignore all of `devtools-modules` though? the `rep.css` changes required to uphold stylint are very minimal. Easy to roll that back though if we want to.

jasonLaster:
@Garbee yup, we should exclude `devtools-modules` the package is for files copy and pasted from `m-c` so we want to be able to do a diff.

Issue #1153, opened 2016-11-08
Change directory structure


jasonLaster:
Currently, the debugger's source lives in `public/js/**` we would like to move these files to a new directory `src`.

Here's a more verbose description of the files that will move:
* `public/js/**/*` => `src`
* `public/build` => `assets/build`
* `public/images` => `assets/images`
* `public/debugger.properties` => `assets/locales/debugger.properties`
* `packages/devtools-local-toolbox/public/js` => `packages/devtools-local-toolbox/src`

What else needs to change:
* check the webpack configs `webpack.config.js` `packages/devtools-local-toolbox/webpack.config.js` `packages/devtools-local-toolbox/webpack.devtools.config.js`
* check eslintrc, eslintignore
* check stylelintrc stylinteignore
* check `public/js/node-unit-test`
* check `packages/devtools-local-toolbox/bin/development-server`
* check everything in`bin`
* check the other packages, especially config...
* do a project grep for `public/js`

This is obviously a large change, but hopefully not too many individual files need to be edited. If you're interested in grabbing it and run into any issues, share early, ask for questions, etc.

wldcordeiro:
I'll grab this. Got my new setup working too 👍

jasonLaster:
Brave! Welcome back :)

jasonLaster:
This was done in #1184

PR #1184, opened 2016-11-13
Directory restructure

wldcordeiro:
Associated Issue: #1153 

### Summary of Changes

* reorganize the directories of the project to follow the `src/` convention rather than a `public/js` convention.

### Testing

* [x] passes `npm test`
* [x] passes `npm run lint`

@jasonLaster @jlongster want to give this a good review? Might have missed something. One test also fails for me with a timeout. Not sure what I can do about that.

> 1) sources should download a sourcemap and create sources:
     Error: timeout of 2000ms exceeded. Ensure the done() callback is being called in this test.


jasonLaster:
Awesome! This will land before the toolbox console work. 

My guess is the timeout I'd related to the node try script moving network requests 

Thanks for doing this!

wldcordeiro:
Gah, I really don't get the failure. It's not present in master currently but is in this branch. @jasonLaster if you could give it a try and see if you have any ideas it'd be great (I'll be on IRC tomorrow or should be, I just need to configure my client heh)

jasonLaster:
@wldcordeiro just pushed a fix. It was tough to track down!

jasonLaster:
@wldcordeiro just rebased. this should hopefully be good to go soon

wldcordeiro:
Looks good now. Tried it out locally and tests pass as well as CI

Issue #1173, opened 2016-11-12
Don't set string [object object] after opening search box


magicp-github:
**Steps to Reproduce:**
1. Start latest Nightly
2. Open Debugger (Ctrl+Shift+S)
3. Click [+] button > Search box is opened
4. Close search box
5. Click [+] button again

**Actual Results:**
String "[object object]" is set at search box.

![do-not-set-string- object object -after-opening-searchinput](https://cloud.githubusercontent.com/assets/13914314/20240834/08a6a9aa-a966-11e6-91bf-f6f50bd53daa.png)

**Expected Results:**
Don't set string "[object object]" at search box automatically.

kareniel:
I'm taking a look at this right now. I should have a fix by the end of the day :)

jasonLaster:
This was done in #1185

PR #1185, opened 2016-11-14
Wrap the click handler passed to CloseButton

kareniel:
Associated Issue: #1173

### Summary of Changes

This commits wraps the handler call in the CloseButton component so that the close
function doesn't receive the event object as an argument

- rename `handleClose` to `close` because it is not a handler anymore
- inline the handler when passing it down from autocomplete to the close button
  `e => close()`

### Testing

* [x] passes `npm test`
* [x] passes `npm run

### DIFF
--- a/public/js/components/Autocomplete.js
+++ b/public/js/components/Autocomplete.js
@@ -12,7 +12,7 @@ const Autocomplete = React.createClass({
   propTypes: {
     selectItem: PropTypes.func,
     items: PropTypes.array,
-    handleClose: PropTypes.func,
+    close: PropTypes.func,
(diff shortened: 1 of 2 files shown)


jasonLaster:
looks like lint-js found something. Probably a version discrepency. Will fix tomorrow

```
/home/ubuntu/debugger.html/public/js/components/RightSidebar.js
  261:5  error  Closing curly brace should be on the same line as opening curly brace or on the line after the previous block  brace-style
```

jasonLaster:
* bumped with eslint

jasonLaster:
Thanks!

PR #1186, opened 2016-11-14
Update CI + Docs to use yarn

jasonLaster:
Reading through the getting started docs, i think we could do another pass to help people get started.

perhaps a separate page that we host at `https://devtools-html.github.io/debugger.html/getting-started/` This way we could have images, gifs, and a screencast :)

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -34,15 +34,17 @@ test:
     - npm run flow
 
 dependencies:
+  pre:
+    - curl -o- -L https://yarnpkg.com/install.sh | bash
   cache_directories:
     - "~/downloads"
-
+    - ~/.yarn-cache
   override:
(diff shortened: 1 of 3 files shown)


circle.yml
```diff
@@ -34,15 +36,17 @@ test:
...
     - "~/downloads"
-
+    - ~/.yarn-cache
   override:
     - npm i -g jasonlaster/lerna
-    - lerna bootstrap
-    - ./bin/install-chrome
-    - ./bin/install-firefox
+    - yarn install
+    - docker pull jlongster/mochitest-runner
```
clarkbw: This shouldn't be needed as the `./bin/update-docker` command below should be doing this docker pull.

jasonLaster: nice.

Issue #1188, opened 2016-11-14
[Editor] Extract utility methods


jasonLaster:
The Editor component has several utility functions who's job it is to query or update the state of the codemirror document.

    setText, setMode, editorHeight, highlightLine,
    showSourceText, resizeBreakpointGutter,
    clearDebugLine, setDebugLine

Currently some of the utilities are in [utils/editor.js](https://github.com/devtools-html/debugger.html/blob/master/public/js/utils/editor.js) and many are not.

There are several reasons why it would be nice to move the methods to a utility

1. they are functionally similar and distinct from the other component methods who's job it is to handle a lifecycle, render, or dispatch an action
2. they are easier to unit test independently
3. the Editor is ~500 lines long and could easily grow. It is the most complicated component

There are a couple reasons not to move the utilities

1. perhaps it's not clear if a method would be a utility or in the Editor
2. it can make it potentially harder to follow if you're jumping back and forth

jasonLaster:
closing as this is not actionable

PR #1189, opened 2016-11-14
[Editor] Cleanup editor component

jasonLaster:
Associated Issue: #1188 

Opening this PR as a conversation starter. I haven't tested it yet, so it likely would need some tlc before it could land. @jlongster I'm curious to see what you think. Note, we've discussed some of these ideas before but I think the project has evolved sufficiently since then.

### DIFF
--- a/public/js/utils/editor.js
+++ b/public/js/utils/editor.js
@@ -1,6 +1,10 @@
 const { isPretty } = require("./source");
 const { isEnabled } = require("devtools-config");
 const { isOriginalId } = require("../utils/source-map");
+const { getDocument, setDocument } = require("../utils/source-documents");
+
+let lastJumpLine = null;
+let pendingJumpLine = null;
 
 function shouldShowPrettyPrint(selectedSource) {
   if (!isEnabled("prettyPrint")) {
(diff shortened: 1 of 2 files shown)


jasonLaster:
Forgot to mention this during standup because it's minor.

public/js/utils/editor.js
```diff
@@ -22,7 +26,154 @@ function shouldShowFooter(selectedSource) {
...
+  const { selectedSource } = this.props;
+
+  if (!selectedSource || !shouldShowFooter(selectedSource.toJS())) {
+    return "100%";
+  }
+
+  return "";
+}
+
+function updatePendingJumpLine(props, prevProps) {
```
wldcordeiro: If all the function uses is the `selectedLocation` property on the two, why can't we just pass them in as the signature for this?

public/js/utils/editor.js
```diff
@@ -22,7 +26,154 @@ function shouldShowFooter(selectedSource) {
...
+
+function setDebugLine(editor, selectedFrame, selectedLocation) {
+  if (selectedFrame && selectedLocation &&
+      selectedFrame.location.sourceId === selectedLocation.sourceId) {
+    const line = selectedFrame.location.line;
+    editor.codeMirror.addLineClass(line - 1, "line", "debug-line");
+  }
+}
+
+function editorHeight() {
```
wldcordeiro: Relying on the `this` to be available in this function seems messy, `selectedSource` should be an argument imo.

public/js/utils/editor.js
jasonLaster: that's a good point

public/js/utils/editor.js
jasonLaster: definitely, that's an oversight. this was just designed to sketch the idea

wldcordeiro: Cool, figured it wasn't entirely ready from your comment, but thought I'd give some review to help guide along. 😄

jlongster:
What is the goal here? I don't really understand the desire to split up components into a simple file in `components` and a lot of implementation details in `utils`. Components should be as complex as they need to be. It makes it harder to work with when things are arbitrarily split up.

The code in `utils` has things like `shouldShowFooter` and other things that are UI-specific. If anything, I'd rather be moving things from util into components where we can.

jasonLaster:
> what is the goal

the rationale is in [1188](https://github.com/devtools-html/debugger.html/issues/1188), which is easy to miss. 

The utils are similar to [editor.js](https://dxr.mozilla.org/mozilla-central/source/devtools/client/sourceeditor/editor.js). The goal is to have have an editor api for commands/queries.

One benefit would be in the Atom integration work, it would be possible to re-implement methods like `showSourceText` for atom.

> shouldShowFooter

I agree. This is a bad example, that i wish were in the component aswell. `showMessage`, `showSourceText` are better examples. 

I am sharing early to get your opinion. Happy to close.

jasonLaster:
closed as this was mostly for the conversation

PR #1193, opened 2016-11-14
Clean up lint warnings

jasonLaster:
@garbee helped us expand the scope of our js lints recently. One of the
consequences was that there were some matched files that were ignored,
which showed up in eslint warnings.

This addressees that by:
1. moving tree.js to the sham modules package
2. burning imstruct
3. narrowing the scope of the glob to just the toolbox, the other
packages don't matter.
   * devtools-modules we don't want to change,
   * devtools-sham-modules are copy paste and tweaks from devtools-modules. so we don't want to match
   * devtools-config is tiny and will likely go away with the toolbox work
   * network-config is even smaller and could likely go away soon too...


Garbee:
So long as `devtools-local-toolbox` is the only thing we are worried about in that packages folder then this all LGTM on the linting side. I simply did a global glob to verify what was needed was hit, even if new packages were added later.

jasonLaster:
Yea. We'll need to be vigalent in that case. But it's rare.

jasonLaster:
this was added with the directory restructure.

PR #1197, opened 2016-11-14
Add install yarn script

jasonLaster:
CI is failing when it comes to re-installing yarn. This will keep it
from doing that.


jasonLaster:
@devtools-html/debugger-core mind taking a look at this and giving it a 👍 

this should unblock the tests

PR #1199, opened 2016-11-14
Update flow-bin to 0.35

jlongster:
This is another version of #1170 with small fixes to make it work. It didn't have anything to do with the record types, it just needed some more specific subtyping.


jasonLaster:
oops, this should have included updates to the lockfile. I'll post a new PR in a bit

PR #1207, opened 2016-11-16
Add a test plan

jasonLaster:
One question I always have when reviewing a PR is "what could break?". 

It would be great to see what scenarios the author tested to get a sense of what edge cases were already explored and what edge cases might not have been explored.

I'm not quite sure if I like the language

>  Tell us a little a bit about how you tested the patch.

it could probably be word smithed. 

I removed the two check boxes because with husky we now require that these two checks pass and I have not found a case where it's been helpful. If others feel like it is keeping authors honest I don't feel strongly.

Lastly, I think it would be helpful to prompt people to take a look at our testing script document that will give them an idea of what to check for all of the possible PRs. We're working on this now and when it's ready will share.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE
+++ b/.github/PULL_REQUEST_TEMPLATE
@@ -5,9 +5,19 @@ Associated Issue: #<issue number>
 * change 1
 * change 2
 
-### Testing
+### Test Plan
 
-* [ ] passes `npm test`
-* [ ] passes `npm run lint`
+Tell us a little a bit about how you tested your patch.
+
+Example test plan:
+
+- [ ] Command-P opens the panel
+- [ ] Clicking “+” opens the panel
+- [ ] Clicking a source navigates to the source
+- [ ] Clicking "x" closes the panel
+
+Here's the Debugger's Testing doc
+https://docs.google.com/document/d/1oBMRxV8A2ag2t22YsQOxTdEv0mXKzIg0tggJjRkU1S0/edit#.
+Feel free to improve it!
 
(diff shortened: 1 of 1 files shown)


clarkbw:
I was thinking we could link to our full test script somewhere else and have a small example included so the format was obvious.

So for the file search my test plan looks something like this:

- [ ] Use keyboard shortcut to open up
  - [ ] Placeholder text is present, close button and search icon all lined up
  - [ ] Type 'todo'
  - [ ] Check result count and search term text are correct
  - [ ] Check first item is highlighted
  - [ ] Pressing enter opens the first item
  - [ ] Reopen with keyboard shortcut
  - [ ] Type garbage, see error text
- [ ] Use `+` to open up search

jasonLaster:
@clarkbw nice idea. 

+1 to linking to a test plan
+1 to having a small example

I'm not sure what the level of detail should be for the test plans. I just played around with the a strict format like yours and found it to be verbose and ultimately it could be difficult to maintain and there are a lot of boxes to check :)


Perhaps something more high level:

Test Plan: Source Search
- [ ] Open search with cmd+p shortcut
- [ ] Open search with "+" button
- [ ] Opening search shows the previous value
- [ ] Select a source takes you to the source

clarkbw:
Yeah, simpler is better

jasonLaster:
I started a [test plan](https://docs.google.com/document/d/1oBMRxV8A2ag2t22YsQOxTdEv0mXKzIg0tggJjRkU1S0/edit#) doc.

jasonLaster:
Great, we can alway improve

Issue #1183, opened 2016-11-13
Align prettify button to the left


Garbee:
We could move the prettify source button to the left to align with Chrome's layout. This would provide at least a little more consistency between our layouts which is useful to developers. Minor detail, but pleasant when our users don't need to search for functionality.

![selection_002](https://cloud.githubusercontent.com/assets/868301/20249499/51b1152e-a9c8-11e6-8dec-1a1e1a97c20f.png)

jasonLaster:
Yup. I like that it's below the gutter as well. 

For historical reasons. It was in the right I think because in the first design it floated above the editor. 

Helen what do you think?

helenvholmes:
Sounds good to me!

jasonLaster:
Here's a patch to get started:

We do some funky things with the command bar and [RTL](https://github.com/jasonLaster/debugger.html/blob/60bf5585453a44a398cad056a936e529c4327f1d/docs/local-development.md#internationalization) positioning. 

```diff
diff --git a/public/js/components/SourceFooter.css b/public/js/components/SourceFooter.css
index 43ce230..2308ce7 100644
--- a/public/js/components/SourceFooter.css
+++ b/public/js/components/SourceFooter.css
@@ -13,11 +13,11 @@
 }

 html:not([dir="rtl"]) .source-footer .command-bar {
-  float: right;
+  float: left;
 }

 html[dir="rtl"] .source-footer .command-bar {
-  float: left;
+  float: right;
 }

 .source-footer .command-bar * {
@@ -35,11 +35,11 @@ html[dir="rtl"] .source-footer .command-bar {
 }

 html:not([dir="rtl"]) .command-bar > span {
-  margin-right: 0.7em;
+  margin-left: 0.7em;
 }

 html[dir="rtl"] .command-bar > span {
-  margin-left: 0.7em;
+  margin-right: 0.7em;
 }

 .source-footer .prettyPrint.pretty {
```

Garbee:
Why are we floating instead of just using Flexbox?

jasonLaster:
Good call. Flexbox would be better!

Garbee:
I'll take a look at this once I have the stylelint fixes in place then.

jasonLaster:
Sounds good

Garbee:
Fixed by merging #1217.

PR #1217, opened 2016-11-16
Flexbox for the source footer area.

Garbee:
Associated Issue: #1183 

#### Style

* Changed classnames for command bar in the sources footer to be specific to this component. ~~Using BEM-style notation.~~ This isolates the sources footer command bar from the code in the right sidebar that was conflicting with some of the changes required.
* Flex display for the command bar area. This handles RTL reflow automatically for us.
* Removed `input` style from old search implementation in the footer command area.

#### Script

* Updated classnames to be the scoped classes.
* Use a button instead of a span for the prettify trigger for native a11y help.
* Set the title and now `aria-label` on the icon. `aria-label` added since some screen readers may not read `title` out. It will take precedent when a11y tools are looking for something to read, while `title` still provides the tooltip to visual users for what the button does.

### Testing

* [x] passes `npm test`
* [x] passes `npm run lint`

selection_003


src/components/SourceFooter.css
```diff
@@ -12,50 +12,39 @@
   user-select: none;
 }
 
-html:not([dir="rtl"]) .source-footer .command-bar {
-  float: right;
+.source-footer__command-bar {
```
jasonLaster: @devtools-html/debugger-core @helenvholmes 

I believe this is the first time we'd use BEM for a component that is used in several places. We could potentially do this with the close button and sizes.

I personally am more comfortable with well scoped classes `.command-bar` `.source-footer > .command-bar`, but that's probably because of my experience set and I'd like to have an open mind on this issue.

src/components/SourceFooter.css
```diff
@@ -12,50 +12,39 @@
...
-  margin-left: 0.7em;
+.source-footer__command-bar .prettyPrint.pretty {
+  stroke: var(--theme-highlight-blue);
 }
 
-.source-footer .prettyPrint.pretty {
-  stroke: var(--theme-highlight-blue);
+:root.theme-dark .source-footer__command-bar > .source-footer__command-bar__action {
+  fill: var(--theme-body-color);
 }
```
jasonLaster: nice fix.

Didn't realize you're also improving theme support

src/components/SourceFooter.js
```diff
@@ -13,10 +13,10 @@ const { shouldShowFooter, shouldShowPrettyPrint } = require("../utils/editor");
...
 
 function debugBtn(onClick, type, className = "active", tooltip) {
-  className = `${type} ${className}`;
-  return dom.span(
+  className = `${type} ${className} source-footer__command-bar__action`;
+  return dom.button(
     { onClick, className, key: type },
-    Svg(type, { title: tooltip })
+    Svg(type, { title: tooltip, "aria-label": tooltip })
   );
```
jasonLaster: thanks for helping us make the buttons more accessible. I personally would like to see more of our buttons become `button` elements.

src/components/SourceFooter.js
```diff
@@ -13,10 +13,10 @@ const { shouldShowFooter, shouldShowPrettyPrint } = require("../utils/editor");
 require("./SourceFooter.css");
 
 function debugBtn(onClick, type, className = "active", tooltip) {
-  className = `${type} ${className}`;
-  return dom.span(
+  className = `${type} ${className} source-footer__command-bar__action`;
+  return dom.button(
     { onClick, className, key: type },
-    Svg(type, { title: tooltip })
+    Svg(type, { title: tooltip, "aria-label": tooltip })
```
jasonLaster: Looks like react can map from ariaLabel aria-label

https://facebook.github.io/react/docs/dom-elements.html

src/components/SourceFooter.css
Garbee: This was necessary here since originally it pulled from the code for the sidebar's command bar. Due to the rename of classes to scope it down, this is now required here.

src/components/SourceFooter.js
Garbee: I tried `ariaLabel` and it did not work for me. Nor did `arialabel` as the docs say they must be lowercase. The string value however worked perfectly.

jasonLaster: oh strange, that could be a way that the Svg library handles it. I'll take a look in a bit

src/components/SourceFooter.css
Garbee: The goal with this is to reduce the overall complexity in *most* selectors while keeping the different pieces well-scoped. It does mean we type more, but that does give us a clear path with targets and what is doing what.

FTR I did BEM-style since it is just what I know will work in this case. I'm open to any alternative methodologies that net us isolated components with as little leakage as possible.

So why BEM?

* Most targets become a single class or two. Which keeps the overall complexity on selectors down.
* By doing this, we are scoping each component we build to itself. In this case the right sidebar had the same structure class-wise, which led to some of those classes conflicting. Which means tweaking one particular piece could adversely affect the other later if not done very carefully.
* BEM has tooling available to help lint the targets automatically so it is verifiable that alignment with new additions is happening. With some TLC (regex's to exclude) for places where we know we need to go against strict BEM like the `[dir="rtl"]` checks.

It is a big kind of shift in the class naming methodology and is certainly more verbose then what most developers would like. However, it does have a good net benefit on maintainability of the codebase going forward.

clarkbw: I'm personally a fan of BEM for a number of reasons mentioned (including perf) and I'd like to see it used more broadly in the project and across the rest of the devtools.  However unless we decide to use it more broadly this one off case creates an inconsistency that I don't think we want.

jasonLaster: Hey @Garbee, do you think you could style this without BEM for now and make a new issue for adopting BEM with this code snippet there.

I think the consensus is that we're in favor of BEM, but would like to come up with a plan so the project adopts BEM well.

jasonLaster: #opsissues

jasonLaster: talked with @Garbee offline and the simple solution is to rename the class `command-bar` to something else.

Garbee:
Updated the classnames to not be BEM-style. However, did continue to rename from `command-bar` to `commands` to have isolation from the right sidebar. Always targeted under `source-footer` for scoping.

In the last update I did remove the pretty stroke styling when prettify was triggered, as the button is hidden on prettified sources and it doesn't seem to have any apparent effect on triggering the button. Therefore this rule change seemed useless.

jasonLaster:
Thanks.

Initially the button selected the ugly source

src/components/SourceFooter.css
```diff
@@ -12,50 +12,35 @@
...
-.source-footer input {
-  line-height: 16px;
-  margin: 7px;
-  border-radius: 2px;
-  border: 1px solid var(--theme-splitter-color);
-  padding-left: 4px;
-  font-size: 10px;
+.source-footer > .commands > .action svg {
+  height: 1em;
+  width: 1em;
```
jasonLaster: I like the use of *ems` in this PR

src/components/SourceFooter.css
```diff
@@ -12,50 +12,35 @@
...
-}
-
-html[dir="rtl"] .command-bar > span {
-  margin-left: 0.7em;
-}
-
-.source-footer .prettyPrint.pretty {
-  stroke: var(--theme-highlight-blue);
+:root.theme-dark .source-footer > .commands > .action {
+  fill: var(--theme-body-color);
```
jasonLaster: 👍

src/components/SourceFooter.js
```diff
@@ -13,10 +13,10 @@ const { shouldShowFooter, shouldShowPrettyPrint } = require("../utils/editor");
 require("./SourceFooter.css");
 
 function debugBtn(onClick, type, className = "active", tooltip) {
-  className = `${type} ${className}`;
-  return dom.span(
+  className = `${type} ${className} action`;
+  return dom.button(
```
jasonLaster: I'll create a new issue for following up on our other SVG buttons so that they can as accessible

PR #1229, opened 2016-11-16
Start iterating on the standalone application

jasonLaster:
### Summary of Changes

* Adds an application.json for App specific configs and update configs
* Adds a hostURL so that the application knows how to resolve paths
   * firefox tab connection
   * worker paths ...
* Adds a new Source header... (mostly for show)
* Where'e the app code? [repo](https://github.com/jasonLaster/debugger.app)

### Screenshots/Videos 

screen shot 2016-11-16 at 5 32 39 pm

### DIFF
--- a/bin/getConfig.js
+++ b/bin/getConfig.js
@@ -3,6 +3,7 @@ const fs = require("fs");
 const path = require("path");
 
 function getConfig() {
+  const applicationConfig = require("../configs/application.json");
   const firefoxConfig = require("../configs/firefox-panel.json");
   const developmentConfig = require("../configs/development.json");
 
@@ -15,8 +16,8 @@ function getConfig() {
     return firefoxConfig;
   }
 
-  const envConfig = process.env.TARGET === "firefox-panel" ?
-                  firefoxConfig : developmentConfig;
+  const envConfig = process.env.TARGET === "application" ?
+                  applicationConfig : developmentConfig;
 
   return merge({}, envConfig, localConfig);
 }
--- a/configs/application.json
+++ b/configs/application.json
@@ -2,7 +2,7 @@
   "title": "Debugger",
   "environment": "development",
   "baseWorkerURL": "http://localhost:8000/assets/build/",
-  "host": "",
+  "host": "http://localhost:8000",
   "theme": "light",
   "logging": {
     "client": false,
--- a/configs/firefox-panel.json
+++ b/configs/firefox-panel.json
@@ -1,6 +1,7 @@
 {
   "environment": "firefox-panel",
   "baseWorkerURL": "resource://devtools/client/debugger/new/",
+  "host": "",
   "logging": false,
   "clientLogging": false
 }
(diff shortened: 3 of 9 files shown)


bin/getConfig.js
```diff
@@ -11,8 +12,8 @@ function getConfig() {
     return firefoxConfig;
   }
 
-  const envConfig = process.env.TARGET === "firefox-panel" ?
-                  firefoxConfig : developmentConfig;
+  const envConfig = process.env.TARGET === "application" ?
+                  applicationConfig : developmentConfig;
```
jlongster: You just got rid of `firefoxConfig`. You need to handle the `firefox-panel` target here.

src/components/Sources.js
```diff
@@ -18,13 +19,21 @@ const Sources = React.createClass({
...
   displayName: "Sources",
 
+  header() {
+    if (isApplication()) {
+      return "Firefox Debugger";
+    }
+
+    return L10N.getStr("sources.header");
+  },
+
```
jlongster: Not fully convinced we need this. It seems like a special case that diverges from the design for not much reason. Also, it's confusing if you are debugging node and it says "Firefox Debugger".

src/utils/source-map.js
```diff
@@ -16,7 +16,7 @@ function restartWorker() {
     sourceMapWorker.terminate();
   }
   sourceMapWorker = new Worker(
-    getValue("baseWorkerURL") + "source-map-worker.js"
+    getValue("hostURL") + getValue("baseWorkerURL") + "source-map-worker.js"
```
jlongster: Can the config just duplicate the `hostURL` instead of adding more complexity here? I'd rather have absolute URLs in the config so you know exactly what URL it's using.

src/components/Sources.js
jasonLaster: @jlongster agreed. The UX is not ideal. mostly a POC. 

@clarkbw suggested that a subtle signal like the application title is sufficient.

This will likely be dropped soon

bin/getConfig.js
jasonLaster: yeah. I realized that that logic was duplicated right above actually. 

This should be tested by our CI.

jasonLaster: yeah. I realized that that logic was duplicated right above actually.

This should be tested by our CI.

src/utils/source-map.js
jasonLaster: Hmm, will give that a shot, this is the part i like least

src/components/Sources.js
jasonLaster: dropped

bin/getConfig.js
jlongster: This doesn't seem fixed with the new commit?

jlongster:
Looks good to me, if you make sure the the firefox panel config is getting used correctly!

jasonLaster:
🌟

Issue #1216, opened 2016-11-16
[Yarn] Yarn install throws an error 


jasonLaster:
We've seen two errors with `yarn install` when Yarn is at version 0.17. 

* Using `0.16.1 fixes the issues.
* Even though there's an error, the install might still be successful (lets confirm this)

Error 1: npm-workerjs

```
error An unexpected error occurred: "ENOENT: no such file or directory, 
lstat '/Users/santosh.v04/Library/Caches/Yarn/npm-workerjs-0.1.1-6c53f106c5664d4691c241ab2f43c5afe4d74fa8'".
```

Error 2: roadrunner.json
[issue](https://github.com/yarnpkg/yarn/issues/1724)

```
/home/ubuntu/.yarn/bin/yarn.js:47
      throw err;
      ^

Error: ENOENT: no such file or directory, open '/home/ubuntu/.cache/yarn/.roadrunner.json'
    at Error (native)
    at Object.fs.openSync (fs.js:640:18)
    at Object.fs.writeFileSync (fs.js:1333:33)
    at /home/ubuntu/.yarn/node_modules/roadrunner/index.js:25:6
    at /home/ubuntu/.yarn/node_modules/roadrunner/index.js:12:12
    at emitOne (events.js:101:20)
    at process.emit (events.js:188:7)
    at processEmit [as emit] (/home/ubuntu/.yarn/node_modules/signal-exit/index.js:140:35)
    at process.exit (internal/process.js:146:15)
    at Command.<anonymous> (/home/ubuntu/.yarn/node_modules/commander/index.js:825:13)
> Yarn was installed, but doesn't seem to be working :(.

```

jasonLaster:
@juliandescottes here is the `roadrunner` bug. I'm surprised that you were seeing it with `0.16.1`. Lets see if we can come up with a fix soon. Happy to help later.

jasonLaster:
closing as Yarn is not being flakey on CI and others aren't reporting this. We can re-open if they do.

#1235

stellargo:
```
Users-MacBook-Pro:debugger.html User$ yarn start
yarn start v1.0.1
$ node bin/dev-server
module.js:474
    throw err;
    ^

Error: Cannot find module 'devtools-launchpad/index'
    at Function.Module._resolveFilename (module.js:472:15)
    at Function.Module._load (module.js:420:25)
    at Module.require (module.js:500:17)
    at require (internal/module.js:20:19)
[...]

PR #1233, opened 2016-11-17
Update getting started docs

jasonLaster:
Associated Issue: #1216 

### Summary of Changes

* Add yarn install step for 0.16.1
* Update yarn explainer text
* Add screenshots + gifs where it helps.


jasonLaster:
This doesn't fix #1216, but it should help others not run into the trap... When Yarn fixes these issues we can close #1216 and also drop the 16.1 requirement

jlongster:
Cool, thanks!

Issue #1208, opened 2016-11-16
[Editor] Add "disable breakpoint" to editor context menu item


jasonLaster:
Chrome's debugger lets users disable a breakpoint from the editor context menu, we should have that as well

**Code Sketch**

```diff
diff --git a/src/components/Editor.js b/src/components/Editor.js
index ae71d2d..f783061 100644
--- a/src/components/Editor.js
+++ b/src/components/Editor.js
@@ -233,6 +233,8 @@ const Editor = React.createClass({
       cbLabel = L10N.getStr("editor.editBreakpoint");
     }

+    const disableBpLabel = L10N.getStr("editor.disableBreakpoint");
+
     const toggleBreakpoint = {
       id: "node-menu-breakpoint",
       label: bpLabel,
@@ -254,6 +256,17 @@ const Editor = React.createClass({
       click: () => this.showConditionalPanel(line)
     };

+    const disableBpLable = {
+      id: "node-menu-disable-breakpoint",
+      label: disableBpLabel,
+      accesskey: "D",
+      disabled: false,
+
+      // we might need to add another param or a new funtion to do this
+      // you can see how disabling works in actions/breakpoints.js
+      click: () => this.toggleBreakpoint(line)
+    };
+
     showMenu(e, [
       toggleBreakpoint,
       conditionalBreakpoint
```

### Firefox
![screen shot 2016-11-15 at 8 36 56 pm](https://cloud.githubusercontent.com/assets/254562/20331354/cf00ecc4-ab72-11e6-86d1-f70ba82384f6.png)

### Chrome
![screen shot 2016-11-15 at 8 36 33 pm](https://cloud.githubusercontent.com/assets/254562/20331355/cf020870-ab72-11e6-9524-dc58168b1c51.png)

shlomif:
Hi! I'm working on it.

clarkbw:
🎉 @shlomif don't hesitate to ask questions if you get stuck!

jasonLaster:
@shlomif nice job discovering the "Enable" Case, missed it the first time!

Issue #1228, opened 2016-11-16
Cannot identify menuitem for "Add Breakpoint" and "Remove Breakpoint" as using same id "node-menu-breakpoint"


magicp-github:
Cannot identify the following menuitems...

 #node-menu-breakpoint as "Add Breakpoint" 
 #node-menu-breakpoint as "Remove Breakpoint"

#node-menu-conditional-breakpoint as "Add Conditional Breakpoint"
#node-menu-conditional-breakpoint as "Edit Breakpoint"

**Expected Results:**
Define different Id or different class.

clarkbw:
Looks like we need to alter the id used here: https://github.com/devtools-html/debugger.html/blob/master/src/components/Editor.js#L225-L261

jasonLaster:
Thanks @magicp-github. 

I'm curious what use case did you run into here?

magicp-github:
Hello @jasonLaster,
Firstly, I confused this debugging of new debugger in foreign language Firefox (e.g. Arabic). I am checking any elements by Id or class, because I cannot read Arabic.
Secondly, I could not add context menu icons correctly. Context menu icons helps me in foreign language.

PR #1242, opened 2016-11-17
Implement enable/disable breakpoint (Issue #1208)

shlomif:
Associated Issue: #1208 

### Summary of Changes

* Add code in Editor.js for the enable/disable breakpoint in the context menu.
* Add localisation strings.

### Test Plan

I set up a breakpoint in my code, enabled it and disabled it again.

See https://github.com/devtools-html/debugger.html/issues/1208 - this
implements the context menu enable/disable breakpoint menu item. I
hereby put all my changes under the MIT/Expat licence.


src/components/Editor.js
```diff
@@ -150,6 +152,30 @@ const Editor = React.createClass({
     }
   },
 
+  toggleBreakpointDisabledStatus(line) {
```
jasonLaster: Hmm, I bet we can avoid adding this function
either

a)
```js
toggleBreakpoint(line, {disable} = {}) {
}
```

This way, callee's can say that they'd like to disable a bp instead of removing it. 

or b)

add a disableBreakpoint function that's called in that case and otherwise we call `toggleBreakpoint`

```js
disableBreakpoint(line) {
}
```

just a thought... could be wrong :)

src/components/Editor.js
```diff
@@ -150,6 +152,30 @@ const Editor = React.createClass({
...
+    if (!bp.disabled) {
+      this.props.disableBreakpoint({
+        sourceId: this.props.selectedLocation.sourceId,
+        line: line + 1
+      });
+    } else {
+      this.props.enableBreakpoint(
+        { sourceId: this.props.selectedLocation.sourceId,
+          line: line + 1 },
+      );
```
jasonLaster: small nit, but I prefer this style:

```js
this.props.enableBreakpoint({
  sourceId: this.props.selectedLocation.sourceId,
  line: line + 1 
})
```

shlomif: done.

src/components/Editor.js
shlomif: I personally don't like conflating more than one task in a single function.

src/components/Editor.js
```diff
@@ -254,10 +286,23 @@ const Editor = React.createClass({
...
-    ]);
+    ];
+
+    if (bp) {
+      const disableBreakpoint = {
+        id: "node-menu-disable-breakpoint",
+        label: disableBpLabel,
+        accesskey: "D",
+        disabled: false,
+        click: () => this.toggleBreakpointDisabledStatus(line)
```
jasonLaster: I've been thinking about this a bit and came up with what I *think* is a nice improvement.

Lets create each context menu item with the specific action

e.g.

```js
const disableBreakpoint = {
         id: "node-menu-disable-breakpoint",
         label: disableBpLabel,
         accesskey: "D",
         disabled: false,
         click: () => this.props.disableBreakpoint({
              sourceId: ...,
              line: line + 1
         });
```

I think we might go this approach anyways because each item should have it's own ID as well. i.e. "node-menu-disable-breakpoint", and "node-menu-enable-breakpoint" https://github.com/devtools-html/debugger.html/issues/1228

We do the same thing already with `enableBreakpoint` so it's just a question of creating the other 8 or so items.

What do you think?

src/components/Editor.js
Garbee: I think this method is fine how it is. The logic is cleanly separated and should make tracking things down easier later.

jasonLaster: Hmm, I can see what you're saying. It is also consistent with `toggleBreakpoint` which is right above, which makes sense. 

Thanks for being patient @shlomif.

jasonLaster:
Thanks @shlomif for the great work

shlomif:
On Sun, 20 Nov 2016 13:54:58 -0800
Jason Laster <notifications@github.com> wrote:

> Thanks @shlomif for the great work
> 

Thanks Jason for merging it and for your comments!

PR #1249, opened 2016-11-18
Sanitize dependencies

jasonLaster:
* fixes chrome-remote-debugging so it exposes the right files

* collapse dev dependencies into dependencies so that they can be available when installed.
 
* sort the packages so that they're alphabetical 

* adds dependencies to the toolbox that the console needed


jasonLaster:
@juliandescottes I sanitized the console packages quite a bit with the latest commit to this PR.

Now the console.html package.json looks like this: 
https://github.com/jasonLaster/console.html/blob/master/package.json#L12

You'll notice that the dep for the toolbox is a path to the toolbox, that's because I kept updating the dep. Once this PR lands, we can bump the toolbox to "0.0.9" and tools should be alright.

I also created this [issue](https://github.com/devtools-html/debugger.html/issues/1251) for cleaning up the debugger's deps

packages/devtools-local-toolbox/webpack.config.js
```diff
@@ -29,7 +29,10 @@ module.exports = (webpackConfig, envConfig) => {
     { test: /\.json$/,
       loader: "json" },
     { test: /\.js$/,
-      exclude: /(node_modules|bower_components|fs)/,
+      exclude: request => {
+        return request.match(/(node_modules|bower_components|fs)/)
+               && !request.match(/devtools-local-toolbox/);
+      },
```
jasonLaster: @juliandescottes I think this will fix a **lot** of the issues you might have been seeing where the toolbox is now typed w/ flow and would break in the console and inspector!

packages/devtools-local-toolbox/package.json
```diff
@@ -20,36 +20,65 @@
...
     "check-node-version": "^1.1.2",
+    "chrome-remote-debugging-protocol": "0.0.9",
+    "classnames": "^2.2.5",
     "co": "=4.6.0",
     "css-loader": "^0.25.0",
+    "devtools-config": "^0.0.8",
+    "devtools-modules": "^0.0.8",
+    "devtools-network-request": "^0.0.8",
+    "devtools-sham-modules": "^0.0.8",
+    "eslint-plugin-babel": "^3.3.0",
```
juliandescottes: Should we add the dependency to `"eslint": "^3.10.0"` here as well? 
I just had a warning about this when doing `npm link` in this package.

packages/devtools-local-toolbox/package.json
```diff
@@ -20,36 +20,65 @@
...
+    "webpack-dev-middleware": "^1.6.1",
+    "webpack-env-loader-plugin": "^0.1.4",
+    "webpack-hot-middleware": "^2.12.0",
     "ws": "^1.0.1"
   },
   "files": [
     "src",
     "bin",
-    "index.js"
+    "index.js",
```
juliandescottes: nit: maybe sort this array alphabetically ?

packages/devtools-local-toolbox/webpack.config.js
juliandescottes: Ah, great catch!

package.json
```diff
@@ -70,6 +56,7 @@
     "classnames": "^2.2.5",
     "co": "=4.6.0",
     "css-loader": "^0.26.0",
+    "codemirror": "^5.1.0",
```
juliandescottes: move before css-loader

jasonLaster:
Nice catches.

jasonLaster:
@juliandescottes landed!

Issue #1266, opened 2016-11-19
Add vertical layout


jasonLaster:
It would be great to support a vertical layout when the debugger is less than 700px wide.

This is actually pretty easy to support given our React components and SplitBox, which supports horizontal and vertical mode. And it would be a really nice win for the community!

**Work breakdown:**
- [x] Extract Command Bar from Right Sidebar https://github.com/devtools-html/debugger.html/pull/1267
- [ ] Refactor Accordion to allow it to be more composable https://github.com/devtools-html/debugger.html/pull/1277
- [ ] setup two layouts horizontal & vertical.
  *  It might be separate render methods in App
  * It might be new components
  * Right Sidebar might have to die... Styles in right sidebar might have to be applied to accordions..
- [ ] Add some toggling logic. Likely App checks the window size on mount and sets a state field `layoutMode = "horizontal"`. There should also be an `onWindowResize` handler
- [ ] profit. This is such a cool project :)

**[app.js sketch](https://gist.github.com/jasonLaster/3ee247ab25e4070f0138f1c7d7656f1f)**

```
curl https://gist.githubusercontent.com/jasonLaster/3ee247ab25e4070f0138f1c7d7656f1f/raw/26ab89ea1e2ce336f986fbf15f2fe905e96edac6/foo.diff | git apply
```

**[bottom sidebar sketch](https://gist.github.com/4c6a78a51478cfe0e3d36ef831946a10)**

```
curl https://gist.githubusercontent.com/jasonLaster/4c6a78a51478cfe0e3d36ef831946a10/raw/44eed81fdf8aac2dd7efa9d8acac628271522853/foo.diff | git apply
```

#### Screenshots
* [old debugger](https://cloud.githubusercontent.com/assets/254562/20457856/baf00614-ae62-11e6-892d-cb90fc45a84e.png)
* [New Debugger](https://cloud.githubusercontent.com/assets/254562/20457855/b3913ce4-ae62-11e6-8a9a-85df07171b8a.png)
* [Chrome](https://cloud.githubusercontent.com/assets/254562/20457854/aea16cf4-ae62-11e6-9c5f-d48607e51c60.png)

jasonLaster:
@kdzwinel let me know if you have any questions. This PR can easily grow in scope!

PR #1267, opened 2016-11-19
Extract command bar

jasonLaster:
Associated Issue: #1266

### Summary of Changes

* Extracts Command Bar so that it can be re-used in the vertical layout
* cleans up the watch expression buttons so that it is more clear what is going on

### Test Plan

- [x] test wait on break
- [x] test step in, step out, step over
- [x] test toggle all breakpoints
- [x] test pause on caught and uncaught exceptions

### DIFF
--- a/src/components/RightSidebar.js
+++ b/src/components/RightSidebar.js
@@ -2,17 +2,8 @@ const React = require("react");
 const { DOM: dom, PropTypes } = React;
 const { connect } = require("react-redux");
 const { bindActionCreators } = require("redux");
-const { getPause, getIsWaitingOnBreak, getBreakpointsDisabled,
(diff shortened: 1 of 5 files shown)


src/components/RightSidebar.js
```diff
@@ -2,24 +2,18 @@ const React = require("react");
 const { DOM: dom, PropTypes } = React;
 const { connect } = require("react-redux");
 const { bindActionCreators } = require("redux");
-const { getPause, getIsWaitingOnBreak, getBreakpointsDisabled,
-        getShouldPauseOnExceptions, getShouldIgnoreCaughtExceptions,
-        getBreakpoints, getBreakpointsLoading
+const {
       } = require("../selectors");
```
gregtatum: nit: Empty require, although you may plan on using it on a follow-up patch.

Issue #1234, opened 2016-11-17
[File Search] Remove outlines


jasonLaster:
We should remove the search outlines when the search bar is closed.

#### Sketch:

```diff
diff --git a/src/components/EditorSearchBar.js b/src/components/EditorSearchBar.js
index 306c66b..1125958 100644
--- a/src/components/EditorSearchBar.js
+++ b/src/components/EditorSearchBar.js
@@ -86,6 +86,8 @@ const EditorSearchBar = React.createClass({
       const selection = this.props.editor.codeMirror.getSelection();
       this.setSearchValue(selection);
       this.selectSearchInput();
+    } else {
+      this.search("");
     }
   },
```

#### Gif
![](http://g.recordit.co/kLewlSTlZj.gif)

arthur801031:
@jasonLaster  May I work on this? I'm a first time contributor to debugger.html.

wldcordeiro:
@arthur801031 go for it. It's always great to get new contributors.

clarkbw:
Its yours @arthur801031 be sure to reach out when you have any questions! 🎉

arthur801031:
@wldcordeiro  @clarkbw 
I don't know what's the shortcut key for bringing up the file search bar.

I tried "cmd+f", and it gave me a normal search bar:
<img width="1227" alt="screen shot 2016-11-18 at 3 32 30 pm" src="https://cloud.githubusercontent.com/assets/4562118/20422276/5a1778ae-ada4-11e6-8d56-a97949b83a32.png">

I also tried "cmd+p", and it gave me this:
<img width="1223" alt="screen shot 2016-11-18 at 3 35 29 pm" src="https://cloud.githubusercontent.com/assets/4562118/20422333/adc92c90-ada4-11e6-9288-8e9d2cc821da.png">

Thank you!

jasonLaster:
Great question @arthur801031. You need to run the [debugger locally](https://github.com/devtools-html/debugger.html/blob/master/CONTRIBUTING.md) and then you can do cmd+f.

You also need to enable editorSearch [configs](https://github.com/devtools-html/debugger.html/blob/master/docs/local-development.md#configs)

jasonLaster:
@bomsy do you have an idea a good idea for how this might be done?

arthur801031:
@jasonLaster @clarkbw @wldcordeiro  I've added the code to remove outlines. Please check. Thank you very much! :)

PR #1272, opened 2016-11-21
Remove outlines

arthur801031:
#1234

- After closing the search bar, the previous selected text will remain selected. I'm not sure if we need to deselect this text.


jasonLaster:
Thanks @arthur801031 this looks really good

I found a couple things that were slightly off and updated the [test plans](https://docs.google.com/document/d/1oBMRxV8A2ag2t22YsQOxTdEv0mXKzIg0tggJjRkU1S0/edit#) doc with some of these requirements. The doc is *really* new so feel free to add requirements i might have missed.

1. when the search bar is closed, Command+G stops working
2. when the search bar re-opens, Enter removes the query
3. when the search bar re-opens, Command+G removes the query

Honestly, I think (2) was there beforehand, but it is pretty similar so I hope you don't mind fixing that nit.

arthur801031:
@jasonLaster I've fixed the bugs that you mentioned. Thank you! :)

Issue #1162, opened 2016-11-10
Don't overflow searchinput and no-result-msg


magicp-github:
**Steps to Reproduce:**
1. Start Nightly
2. Go to any sites (e.g. https://www.mozilla.org/en-US/contribute/)
3. Open Debugger (Ctrl+Shift+S)
4. Open searchinput (Ctrl+P)
5. Type search words (e.g. "zzzzz" ) until overlap/overflow

**Actual Results:**
searchinput and no-result-msg overlap/overflow, then layout is broken.
![do-not-overflow-and-overlap-searchinput-and-no-result-msg](https://cloud.githubusercontent.com/assets/13914314/20168505/ec0fd502-a765-11e6-87ea-4cc893acf013.png)

**Expected Results:**
- Overflowed search word is hidden before overlapping with close-btn.
- Don't overflow no-result-msg.

jasonLaster:
Thanks for finding this

PR #1276, opened 2016-11-21
Do not overflow "No files matching" message.

kdzwinel:
Associated Issue: #1162

### Summary of Changes

* break long search queries to stop "no files matching" message from overflowing

### Test Plan

Example test plan:

- [ ] Test if "No files matching" message with a long query does not overflow
- [ ] Test if "No files matching" message with a short query looks the same as before the changes

### Screenshots/Videos

Long query before:
before_long_query

Long query after:
after_long_query

Short query after:
after_short_query

### DIFF
--- a/src/components/Autocomplete.css
+++ b/src/components/Autocomplete.css
@@ -98,6 +98,7 @@
   width: 24px;
   margin-right: 4px;
   line-height: 0;
+  flex-shrink: 0;
 }
 
 .autocomplete .no-result-msg .sad-face svg {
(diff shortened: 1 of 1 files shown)


jasonLaster:
do you have an example of a shrunk sad face? :)

I'm mostly asking for a friend @helenvholmes

kdzwinel:
Thank you for helping me with this! Here's a treat for you:

screen shot 2016-11-21 at 22 05 29

jasonLaster:
The tests will pass soon. They baked in the old behavior

PR #1277, opened 2016-11-21
Refactor Accordion

jasonLaster:
Associated Issue: #1266 

### Summary of Changes

I refactored the Accordion component so that it had a composable
interface that would allow Panes to be shown in different locations
based on horizontal and vertical layouts.

* Add an Accordion Pane component
* drop the `created` field
* Add a header buttons element

### Test Plans

- [x] right sidebar did not change
- [x] clicking the header works
- [x] top pane is always open

### DIFF
--- a/src/components/Accordion.js
+++ b/src/components/Accordion.js
@@ -1,75 +1,98 @@
 const React = require("react");
 const { DOM: dom, PropTypes } = React;
-
-const { div, span } = dom;
 const Svg = require("./utils/Svg");
 
 require("./Accordion.css");
 
 const Accordion = React.createClass({
   propTypes: {
-    items: PropTypes.array
+    children: PropTypes.array.object
   },
 
   displayName: "Accordion",
 
-  getInitialState: function() {
-    return { opened: this.props.items.map(item => item.opened),
-      created: [] };
-  },
-
-  handleHeaderClick: function(i) {
-    const opened = [...this.state.opened];
-    const created = [...this.state.created];
-    const item = this.props.items[i];
-
-    opened[i] = !opened[i];
-    created[i] = true;
-
-    if (opened[i] && item.onOpened) {
-      item.onOpened();
-    }
-
-    this.setState({ opened, created });
-  },
-
-  renderContainer: function(item, i) {
-    const { opened, created } = this.state;
-    const containerClassName =
-          item.header.toLowerCase().replace(/\s/g, "-") + "-pane";
-
-    return div(
-      { className: containerClassName, key: i },
-
-      div(
-        { className: "_header",
-          onClick: () => this.handleHeaderClick(i) },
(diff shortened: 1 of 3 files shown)


jasonLaster:
ping @gregtatum mind doing a sanity check.

James' initially worked on the Accordion component, so we can put this on the back burner until he gets back.

src/components/Accordion.js
```diff
@@ -1,75 +1,98 @@
...
+    header: PropTypes.string
   },
 
-  handleHeaderClick: function(i) {
-    const opened = [...this.state.opened];
-    const created = [...this.state.created];
-    const item = this.props.items[i];
+  displayName: "AccordionPane",
+
+  getInitialState: function() {
```
gregtatum: I'm curious why you would store the state in the component when you have Redux, rather than providing an isOpened, and toggleOpen functions as props, and then maintain that state in the store.

src/components/RightSidebar.js
```diff
@@ -51,46 +49,54 @@ const RightSidebar = React.createClass({
...
         { className: "right-sidebar",
           style: { overflowX: "hidden" }},
         CommandBar(),
-        Accordion({
-          items: this.getItems()
-        })
+
+        Accordion({},
+          isEnabled("watchExpressions") ?
+          AccordionPane(
```
gregtatum: Yeah, this is much nicer.

src/components/Accordion.js
```diff
@@ -1,75 +1,98 @@
...
+        this.renderButtons(buttons)
+      ),
+      this.renderContent()
     );
   }
 });
 
-module.exports = Accordion;
+module.exports = {
+  AccordionPane,
```
jlongster: Don't export 2 components from the same file. A lot of React tooling expects a single component per file and it's a good pattern. What is the advantage of `AccordionPane` anyway? We can achieve the same thing without an additional component, and this API eagerly renders all of the panes which we don't want.

Perhaps the only awkward thing about passing in an array of items is that we don't support `null` in the array, so you could conditionally include things like `Expressions`. I suggest simply allowing `null`s in the array but keeping the previous API.

Otherwise this PR looks fine. I don't know much about `Expressions` so don't have much to comment there.

jasonLaster: > Don't export 2 components from the same file

good point.

> What is the advantage of AccordionPane anyway? 

Looking at Konrad's PR, I don't think we gain that much from this improvement. I thought it would be more helpful than it turns out to be.

> we don't support null in the array

That would be nice so that we avoid the conditional unshift.

> eagerly renders all of the panes

I'm not sure why this is a problem? I assume this is why we had created, but I don't see where that helps us.

jasonLaster:
I'm going to close this PR now as I don't think `AccordionPane` buys us too much in terms of functionality we did not have before. @kdzwinel please correct me if i'm wrong and you need it for vertical layout.

Issue #1230, opened 2016-11-16
[App] Build a standalone Node Debugger


jasonLaster:
This is a tracking ticket for a Node prototype.

![screen shot 2016-11-16 at 5 48 40 pm](https://cloud.githubusercontent.com/assets/254562/20369316/4c63087e-ac26-11e6-9392-3398cbd12d8a.png)

**Goals:**

* It will handle the terminal use case where you run `jsdebug foo.js` and want to see a debugger!
* The standalone debugger is just an application shell. It will load the debugger bundle like the local toolbox
* The node debugger will use Node's remote debugger protocol. It won't do anything custom.

**App Repo:**
 https://github.com/jasonLaster/debugger.app

**Breakdown Issues:**
- [x] Add application.json config for loading the debugger from a host
- [x] Add a bin for launching the debugger
- [x] Update the node connection logic
- [ ] present scope chain
- [ ] present variable

Other [issues](https://github.com/jasonLaster/debugger.app/issues) are in the app repo

jasonLaster:
Closing this as we have a working node app... 

Going forward the app issues are tracked in that repo https://github.com/jasonlaster/debugger.app

PR #1279, opened 2016-11-21
(WIP) Update Node Connection Logic

jasonLaster:
Associated Issue: #1230 

### Summary of Changes

* Drop clients/chrome.js connection logic
* Add chrome-remote-interface connection logic
* Update Landing Page to query node tabs

### DIFF
--- a/configs/development.json
+++ b/configs/development.json
@@ -15,7 +15,11 @@
   },
   "chrome": {
     "debug": true,
-    "webSocketPort": 9222
+    "port": 9222
+  },
+  "node": {
+    "debug": true,
+    "port": 9229
   },
   "firefox": {
     "webSocketConnection": false,
--- a/packages/devtools-local-toolbox/src/clients/chrome.js
+++ b/packages/devtools-local-toolbox/src/clients/chrome.js
@@ -1,6 +1,6 @@
 /* eslint-disable */
 
-const { connect } = require("chrome-remote-debugging-protocol");
+const CDP = require("chrome-remote-interface");
 const defer = require("../utils/defer");
 const { Tab } = require("../tcomb-types");
 const { isEnabled, getValue } = require("devtools-config");
@@ -14,14 +14,16 @@ Array.prototype.peekLast = function() {
   return this[this.length - 1];
 };
 
-let connection;
+let client;
 
-function createTabs(tabs) {
+function createTabs(tabs, {type, browser} = {}) {
+  if (tabs.length == 0) {
+    return [];
+  }
 
   return tabs
     .filter(tab => {
-      const isPage = tab.type == "page";
-      return isPage;
+      return tab.type == type;
     })
     .map(tab => {
       return Tab({
@@ -29,7 +31,7 @@ function createTabs(tabs) {
         url: tab.url,
         id: tab.id,
         tab,
-        browser: "chrome"
+        browser
       });
     });
 }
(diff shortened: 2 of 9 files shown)


configs/development.json
```diff
@@ -15,7 +15,11 @@
   },
   "chrome": {
     "debug": true,
-    "webSocketPort": 9222
+    "port": 9222
+  },
+  "node": {
+    "debug": true,
+    "port": 9229
```
jasonLaster: CRI prefers host, port, scheme APIs which makes more sense as well.

packages/devtools-local-toolbox/src/clients/chrome.js
```diff
@@ -38,51 +40,70 @@ function connectClient() {
...
-  const url = `http://localhost:${webSocketPort}/json/list`;
-  networkRequest(url).then(res => {
-    deferred.resolve(createTabs(JSON.parse(res.content)))
-  }).catch(() => deferred.reject());
+  const port = getValue("chrome.port");
+  function fetch(http, options, callback) {
+    const {host, path, port} = options;
+    const url = `http://${host}:${port}${path}`;
+    return networkRequest(url).then(res => callback(null, res.content))
+  }
```
jasonLaster: @cyrus-and I'm not if passing in a custom [fetcher](https://github.com/cyrus-and/chrome-remote-interface/compare/master...jasonLaster:fetcher?expand=1) is the best idea or just a dirty hack.

The problem I'm trying solve is this. The debugger runs in a web context so fetching `localhost:9222/json/list` will be a CORS error. What we do instead is use `networkRequest` which builds a url `localhost:8000/?get=localhost:9222/json/list` to do the request for us. 

What I'm doing here is passing in `fetch` as my go-to fetcher with CRI, but it feels bad. Have you seen others try to do this before?

packages/devtools-local-toolbox/src/clients/chrome.js
```diff
@@ -38,51 +40,70 @@ function connectClient() {
...
+    .then(tabs => createTabs(tabs, { browser: "chrome", type: "page" }))
+    .catch(err => {});
+}
+
+
+function connectNodeClient() {
+
+  if(true){
+    return Promise.resolve(createTabs([], {}))
+  }
```
jasonLaster: disabled for now while I sort out chrome. It will be enabled before this lands ofcourse

packages/devtools-local-toolbox/src/clients/chrome.js
```diff
@@ -38,51 +40,70 @@ function connectClient() {
...
+  client.Runtime.run();
 
-  agents.Page.enable();
+  client.Page.enable();
 
-  connection.registerDispatcher("Debugger", clientEvents);
-  connection.registerDispatcher("Page", pageEvents);
+  debugger
+  // connection.registerDispatcher("Debugger", clientEvents);
+  // connection.registerDispatcher("Page", pageEvents);
```
jasonLaster: will need to do something like what we do [here](https://github.com/devtools-html/debugger.html/blob/master/packages/devtools-local-toolbox/src/clients/firefox.js#L94-L97) to listen to all the events.

packages/devtools-local-toolbox/src/clients/chrome.js
```diff
@@ -38,51 +40,70 @@ function connectClient() {
...
+      { browser: "node", type: "node" }
+    ));
 }
 
 function connectTab(tab) {
-  return connect(tab.webSocketDebuggerUrl).then(conn => {
-    connection = conn;
+  const { webSocketDebuggerUrl } = tab;
+  debugger
+  return CDP({ webSocketDebuggerUrl }).then(_client => {
```
jasonLaster: this is currently failing because it needs a custom fetcher...

packages/devtools-local-toolbox/src/clients/chrome.js
cyrus-and: Yeah, it looks like a dirty hack to me... :)

Fact is, CRI is not meant to run in the browser, and actually I don't understand why you are doing so. I mean, if I were to write a web-based JavaScript debugger (DevTools like) I'd use CRI on the backend. I'm probably missing something and I've only briefly checked out debugger.html, but why do you need to run CRI in a web context?

jasonLaster:
closing this PR as switching to `chrome-remote-interface` required switching the HTTP request layer, which currently uses `networkRequest`

cyrus-and:
@jasonLaster just out of curiosity, why do you need to interact with the remote debugging endpoint (port 9992) from the browser context instead of from the [backend][1]?

[1]: https://github.com/devtools-html/debugger.html/pull/1279#discussion_r89082143

jasonLaster:
@cyrus-and I'll file a more detailed writeup in the CRI issue, but it's pretty simple. We run our debugger as a web app, which handles the websocket connection with the different targets. Because of this, our backend is pretty dumb and has only a couple of responsibilities:
- serve webpack dev-server with the bundled JS
- handle CORS requests for things like source maps
- serve css

It's much more natural for our app to ask the backend to fetch a URL like `localhost:9222/json/list` than it is for the backend to know about which target we want to get the debuggable tabs from...

cyrus-and:
Thanks for the explanation!
But since there actually **is** a backend, my point was to ask it:

    http://localhost:8000/list?host=localhost&port=9222

instead of:

    http://localhost:8000/fetchme?url=localhost%3A9222%2Fjson%sFlist

But this, as you say, implies a slight less dump backend due to the fact that now it has to recognize the `list` operation and whatever else you need (`new`, `close`, etc.). Anyway I see your point, just different design choices.

Issue #1111, opened 2016-11-04
[file search] style outline matches


jasonLaster:
We recently landed support for outlining all of the search results. We should do a quick style pass to clean it up, although it is already pretty good.

Here's the designs from [invision](https://projects.invisionapp.com/share/AY8OEG3C6#/screens/117937554_Debugger) for search

The two takeaways for me are:
1. the outline should be light and stay out of the way
2. the selected match should be yellow. Ideally a garrish yellow as @helenvholmes requested. I'm not sure if we have a css variable for that.

One other consideration is what we want to do for the dark mode. What's the equivalent of a garrish color (is it pink?) Feel free to use your best judgement there. And here's a link for changing the [theme](https://github.com/devtools-html/debugger.html/blob/master/docs/local-development.md#themes) to dark.

![screen shot 2016-11-04 at 12 26 49 pm](https://cloud.githubusercontent.com/assets/254562/20013489/b34012f4-a289-11e6-97ca-44d40e787ed4.png)

jasonLaster:
@monicau I think you might enjoy this one :) 
It involves styling the editor, which as you know is pretty cool.

IagoLast:
Maybe I'm missing something but when I try to reproduce this the browser's default search bar is shown instead the debugger one. 

(When I hit `cmd+p`the debugger search bar is shown but the results dialog is different since this searches on every file)


Could it be some incompatible version of something ?

Debugger.html: 0.0.8
OSX: 10.11.5 (15F34)
NODE: 7.1.0
FFOX-DEV: 51.0a2 (2016-11-14) (64-bit)

Also tested on chrome:
CHROME: 54.0.2840.98 (64-bit)

andreicristianpetcu:
Same here. I was thinking that I was doing something wrong and it did not appear :)

jasonLaster:
Hi @IagoLast, @andreicristianpetcu 

Sorry it wasn't clear

You'll need to go through the [getting started](https://github.com/devtools-html/debugger.html/blob/master/CONTRIBUTING.md#getting-started) setup and enable `editorSearch` in the [local config](https://github.com/devtools-html/debugger.html/blob/master/docs/local-development.md#configs).

monicau:
Hey @jasonLaster thanks for flagging me down. I was able to reproduce this issue. I'll take this if no one is working on it at the moment. Let me know :)

jasonLaster:
Sure!

monicau:
[...]

PR #1282, opened 2016-11-22
(WIP) File search style outline matches

monicau:
Associated Issue: #1111

### Summary of Changes

* Change selected match to yellow
* Lighten the match outlines

### Test Plan

- [ ] Ctrl + F to see the search matches

### Screenshots

debugger-file-search

debugger-file-search-dark

### DIFF
--- a/packages/devtools-local-toolbox/src/lib/themes/dark-theme.css
+++ b/packages/devtools-local-toolbox/src/lib/themes/dark-theme.css
@@ -212,7 +212,7 @@ body {
 }
 
 .cm-s-mozilla .CodeMirror-selected { /* selected text (unfocused) */
-  background: rgb(176, 176, 176);
+  background: rgb(255, 255, 0);
 }
 
 .cm-s-mozilla .CodeMirror-activeline-background { /* selected color with alpha */
--- a/packages/devtools-local-toolbox/src/lib/themes/light-theme.css
+++ b/packages/devtools-local-toolbox/src/lib/themes/light-theme.css
@@ -214,7 +214,7 @@ body {
 }
 
 .cm-s-mozilla .CodeMirror-selected { /* selected text (unfocused) */
-  background: rgb(176, 176, 176);
+  background: rgb(255, 255, 0);
 }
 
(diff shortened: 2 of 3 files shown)


monicau:
This is a WIP.  I'm having trouble changing the text color of the highlighted text in the dark theme.  In dark-theme.css I tried setting the text color in the same class that affects the highlight but it doesn't seem to work:
```
.cm-s-mozilla .CodeMirror-selected {
background: rgb(255, 255, 0);
color: rgb(0, 0, 0);
}
```
When I inspect element, the selected text has classes `cm-overlay cm-selecting`, so do the other matches in the document.  Oddly the currently selected match doesn't have  the class `CodeMirror-selected` or any other classes to differentiate itself from the other matches in the file.. so I'm kinda stumped.

jasonLaster:
@helenvholmes mind taking a quick look at this?

jasonLaster:
Thanks @monicau for the help styling the background color and outline for matches.

I just looked into styling the highlighted element's color and **it is hard** to do with codemirror's api. I mean really hard.

You probably know this, but the yellow overlay is an element absolutely positioned on top of the matched text... and there's no CSS class for the currently highlighted matched text...

* Another problem I noticed is that the matched text is already styled differently and doesn't look great.  This happened before your PR.

* one other observation, your garrish yellow background makes it easier to read the selected matches than the grey.

With all of that mentioned, I think what you have is an improvement and we can fix the matched text color in another issue.


Here's a GIF of the matched color problem:

jasonLaster:
@helenvholmes I think this PR makes sense landing and then we can update the styles in M-C. What do you think?

jasonLaster:
this should help here: https://github.com/devtools-html/debugger.html/pull/1318

it could also give us a good class to match on

jasonLaster:
Thanks @monicau!

I spoke with @helenvholmes offline and we'll update the m-c versions of these files

monicau:
Awesome, thanks @jasonLaster!

jasonLaster:
Hey @monicau! 

I just realized you might not have gotten the invite to our new [slack] channel! 

[slack]: https://devtools-html-slack.herokuapp.com/

Issue #966, opened 2016-10-20
Code Search - re-search when a new source is selected


jasonLaster:
The code search component should do a new search when the selected source changes.

Steps to reproduce:
1. select a source
2. cmd+f to search (type something)
3. select a new source

(we should do a new search, when the source is selected).

![](http://g.recordit.co/35JnlDduKg.gif)

here's a sketch of what the code could look like

``` patch
diff --git a/public/js/components/EditorSearchBar.js b/public/js/components/EditorSearchBar.js
index cecba60..3215bfd 100644
--- a/public/js/components/EditorSearchBar.js
+++ b/public/js/components/EditorSearchBar.js
@@ -56,6 +56,12 @@ const EditorSearchBar = React.createClass({
     if (this.searchInput()) {
       this.searchInput().focus();
     }
+
+    if (props.selectedSource != nextProps.selectedSource) {
+      // redo the search
+      // this will likely extract the onKeyUp logic that does a search and re-use it
+      // e.g. we could add a new method this.search() that would do the search
+    }
   },
```

akhillb:
I would like to try this one. One small question though. If I select a new tab, do you think it would be better to open the tab with search bar open or closed?

jasonLaster:
Thanks

Good question.

The rule of thumb is to be consistent with atom and sublime.

If the search bar is open, id keep it open

jasonLaster:
ping @akhillb do you still think you'll be able to get to this?

akhillb:
@jasonLaster yep i think i will be able to do this.

jasonLaster:
@akhillb do you think you can get to this?

jasonLaster:
@akhillb I'm going to move this back to "up for grabs" feel free to re-claim it if you've got some time.

arthur801031:
@jasonLaster May I work on this? Thank you!

jasonLaster:
Go for it! Let me know if you have questions.

arthur801031:
@jasonLaster  

I have looked at the attributes in `this.props.editor.codeMirror`, `this.props`, and `this.props.editor`, but I still couldn't find `selectedSource`. How do I get this value? Thank you!

jasonLaster:
@arthur801031, oh you might need to pass it in :)

Sorry, that was a quick code sketch and i missed that detail

```diff
diff --git a/src/components/Editor.js b/src/components/Editor.js
index ae71d2d..ff35d60 100644
--- a/src/components/Editor.js
[...]

PR #1283, opened 2016-11-22
Re-search when a new source is selected

arthur801031:
#966


jasonLaster:
Thanks @arthur801031!

I just tried it locally and it works smoothly.

I took a shot at cleaning up some of the code, but I'm not sure if I got it right because there are some weird issues as I type with keys being dropped lol

https://gist.github.com/jasonLaster/3354799646a76927f817282fd8871c2b

Would you mind taking a look at this and letting me know what you think? 

I'd also be happy to merge your patch as is and then clean up some of the searching behavior.

jasonLaster:
Hmm, looks like we're having some mochitest failures. Shouldn't be a big issue, i'll run locally in a little big

arthur801031:
@jasonLaster 

1. We need `this.props.sourceText.get("text") != undefined ` because  `this.props.sourceText.get("text")` would be undefined when you open a new file tab, which would throw an error when getting "text".

2. We need `const count = countMatches(query, this.props.sourceText.get("text")); this.setState({ count: count, index: 0 });` because we need to update the count of text that match the query for the tab you're selecting.

3. I have made the changes that you suggested from line 24~30 and line 38~46 (https://gist.github.com/jasonLaster/3354799646a76927f817282fd8871c2b)

Thank you! :)

jasonLaster:
Ah, all good points.

I'm sorry i did not have a chance to look at the failing tests today. I got caught up working on #1295 and #1294. Unfortunately, I'll be busy over the next couple days doing thanksgivingy things, but I'll get back to this soon.

jasonLaster:
Ahh, just looked at the failing test and found the bug:

```
JavaScript error: resource://devtools/client/debugger/new/bundle.js, line 48484: TypeError: this.props.sourceText is undefined

1089 INFO Console message: [JavaScript Error: "TypeError: this.props.sourceText is undefined" {file: "resource://devtools/client/debugger/new/bundle.js" line: 48484}]

1090 INFO TEST-UNEXPECTED-FAIL | devtools/client/debugger/new/test/mochitest/browser_dbg_keyboard-shortcuts.js | Test timed out - 
```

I pushed a fix and hopefully this will be green soon and we can merge. In the future, it's possible to grep the mochitest output for unexpected fail and javascript error. and often you'll find the problem in the sea of text!

Issue #1112, opened 2016-11-04
[Breakpoints] <cmd>+b should toggle breakpoints


jasonLaster:
In the old debugger `<cmd>+b` was a nifty keyboard shortcut for toggling a breakpoint on a line. `<cmd>+shift+b` not surprisingly opened the conditional bp panel.

We should support this in the new debugger.

Here's a patch to get started:
```diff
diff --git a/public/js/components/Editor.js b/public/js/components/Editor.js
index 5ebdac6..31cec56 100644
--- a/public/js/components/Editor.js
+++ b/public/js/components/Editor.js
@@ -125,6 +125,15 @@ const Editor = React.createClass({
   },

   toggleBreakpoint(line) {
+    if (!line) {
+      // this might not be set...
+      // so we might need to do something else or add a setCursor call
+      // whatever we do here we need to do for the conditional bp panel
+      // by the way, i'm not sure if we want to add this check here or
+      // have a generic handleX method for the binding and then call toggleBp(line) with the line
+      line = this.editor.getCursor();
+    }
+
     const bp = breakpointAtLine(this.props.breakpoints, line);

     if (bp && bp.loading) {
@@ -297,6 +306,10 @@ const Editor = React.createClass({
       );
     }

+    const shortcuts = this.context.shortcuts;
+    shortcuts.on("CmdOrCtrl+b", this.toggleBreakpoint);
+    shortcuts.on("CmdOrCtrl+shift+b", this.showConditionalPanel);
+
     resizeBreakpointGutter(this.editor.codeMirror);
     debugGlobal("cm", this.editor.codeMirror);

@@ -308,6 +321,10 @@ const Editor = React.createClass({
   componentWillUnmount() {
     this.editor.destroy();
     this.editor = null;
+
+    const shortcuts = this.context.shortcuts;
+    shortcuts.off("CmdOrCtrl+b", this.toggleBreakpoint);
+    shortcuts.off("CmdOrCtrl+shift+b", this.showConditionalPanel);
   },
```

![](http://g.recordit.co/AyVpHVIaIE.gif)

jasonLaster:
Hey, @montogeek I think you might like this one. It's a small UX improvement, with a cool editor integration for the line.

arthur801031:
@jasonLaster May I work on this? Thank you! :)

montogeek:
@jasonLaster I like it! But a bit busy these days, sorry!

jasonLaster:
[...]

PR #1291, opened 2016-11-23
<cmd>+b should toggle breakpoints

arthur801031:
#1112


jasonLaster:
Hey @arthur801031 this looks great. I'm glad that codeMirror.getCursor() did the trick

I think, the `shortcuts.on` handlers passing shortcut in is kind of a smell. so I investigated adding some anonymous functions and they seemed to work really well. My big concern was that `shortcuts.off` would not release the handler, but it also seemed to work and remove all the listeners.

Here's a patch where I cleaned up your handlers and the editor search bar. If you'd like, it'd be great to have a new PR that cleans up those handlers too!

https://gist.github.com/jasonLaster/2b9d515ea0c134c3d1cfa5f72189c871

Thanks for all the hard work this week. You're killing it!

arthur801031:
Hi @jasonLaster

Is the new PR for cleaning up handlers in different files? Or are you referring to the handlers in EditorSearchBar.js? Thank you very much for your help and feedback. Have a great Thanksgiving! :)

src/components/Editor.js
```diff
@@ -127,6 +138,10 @@ const Editor = React.createClass({
   },
 
   toggleBreakpoint(line) {
+    if (!line) {
```
jasonLaster: Hmm do you need to check for a line?

src/components/Editor.js
```diff
@@ -251,7 +266,9 @@ const Editor = React.createClass({
       label: cbLabel,
       accesskey: "C",
       disabled: false,
-      click: () => this.showConditionalPanel(line)
+      click: () => this.showConditionalPanel(
+        getCursorLine(this.editor.codeMirror)
```
jasonLaster: Don't you have line already?

arthur801031:
Hi @jasonLaster  

I've made changes based on your comments. Please check. Thank you! :)

jasonLaster:
Thanks @arthur801031!

PR #1294, opened 2016-11-23
Upgrade Event Listeners redux backend

jasonLaster:
### Summary of Changes

* upgrades event listener reducer
* upgrades event listener action for fetching event listeners
* adds a call for fetching events in newSource behind a flag
* moves asPaused over to eventListners

### Test Plan

- [ ] make sure it doesn't break anything when flagged off
- [ ] make sure `appStore.getState().eventListeners.listeners` has state


jlongster:
👍 Looks like straight-forward changes, and it's easy to flag off, so this is fine.

PR #1295, opened 2016-11-23
Update Connection layer to better support node

jasonLaster:
Associated Issue: #1230 

### Summary of Changes

* start using node `json/list`
* connect to node in the same manner as chrome
* upgrade `chrome-remote-debugging-protocol` to supprt bootstrapping node protocol [bump](https://github.com/jasonLaster/chrome-remote-debugging-protocol/commit/07f4fc7ef374674d856b85b1ee91de4a690c8562)

### Test Plan

- [x] test landing page
- [x] connecting to firefox
- [x] connecting to node
- [x] connecting to chrome

### Screenshots/Videos 

* show all three connections
* show connecting to node

CC @clarkbw @bomsy @Garbee


bomsy:
@jasonLaster while trying to test this , i get `networkRequest.js:6 Fetch API cannot load file:///get?url=http://localhost:9229/json/list. URL scheme "file" is not supported.` in the electron app console when running  `jsdebug`. 

Am i missing something?

It's not making the request because `file:///` is not supported. I noticed  of the browser based debugger uses `http://localhost:8000/`.

bomsy:
it works when i change this `fetch('/get?url=${url}')` to `fetch('http://localhost:8000/get?url=${url}')` in https://github.com/devtools-html/debugger.html/blob/master/packages/devtools-network-request/networkRequest.js#L6

How do you get this to work when running  the electron app?

Thanks :)

bomsy:
A suggestion is do we want to pick up the `development.serverPort` from `configs/development.json` and use that for running the networkRequests?

packages/devtools-local-toolbox/src/clients/index.js
```diff
@@ -4,9 +4,9 @@ const chrome = require("./chrome");
 const { debugGlobal } = require("../utils/debug");
 const { createSource } = require("./firefox/create");
 
-let clientType;
+let clientType = null;
 function getClient() {
-  if (clientType === "chrome") {
+  if (clientType === "chrome" || clientType == "node") {
```
juliandescottes: nit: === for consistency

jasonLaster:
Thanks for taking a look @bomsy + @juliandescottes. 

I believe the electron app was failing for you because you need to run `yarn run start-app` from the debugger so it reads in the `application.json` config.

You might also have to run `yarn upgrade`, but it sounds like you did if you got node running.

bomsy:
@jasonLaster ahh cool thanks. i knew i was missing something! thanks. do we need to doc this somewhere?

This LGTM, and work fine!

jasonLaster:
Thanks.

Definitely need to doc it!

Issue #1248, opened 2016-11-18
Getting setup on windows


jasonLaster:
Hey, @viswanathamsantosh mind sharing some of the issues you were seeing getting setup with windows 10. 

I imagine some of the potential issues could be with:
* node
* npm
* yarn
* lerna (symlinks)

Potentially hyperterm could help

CCing - @clarkbw @auchenberg @Garbee so they can help if there are concrete issues

viswanathamsantosh:
![yarn1](https://cloud.githubusercontent.com/assets/3799600/20441831/1ba828fa-adec-11e6-9408-2d770f1ecee0.JPG)
Running 
Node 5.6
Npm  4.0.1
Yarn 0.16.1
but got this error!

Garbee:
The preinstall script is bash-only right now. So only UNIX systems can handle it. We need to investigate writing that script within node itself so we can make it platform-independent. This goes for all of the non-js bins we have.

jasonLaster:
@viswanathamsantosh try dropping the preinstall

```diff
diff --git a/package.json b/package.json
index 9f2444b..7543507 100644
--- a/package.json
+++ b/package.json
@@ -31,8 +31,7 @@
     "prepush": "npm run lint; node src/test/node-unit-tests.js --dots",
     "storybook": "start-storybook -p 6006",
     "build-storybook": "build-storybook",
-    "flow-coverage": "flow-coverage-report -i 'src/actions/*.js' -i 'src/reducers/*.js' -i 'src/utils/*.js' -t html -t text",
-    "preinstall": "./bin/preinstall"
+    "flow-coverage": "flow-coverage-report -i 'src/actions/*.js' -i 'src/reducers/*.js' -i 'src/utils/*.js' -t html -t text"
   },
   "dependencies": {
     "codemirror": "^5.1.0",
```

and then manually running 

```
  npm i -g jasonlaster/lerna
  lerna bootstrap
  yarn install
```

Garbee:
Setting things up on a new system, the preinstall script is also failing on Linux machines. I had to do a manual `npm i jasonlaster/lerna` before `yarn install` would work.

jasonLaster:
Hmm. What was the error?

Garbee:
Line 8 of `preinstall` was causing the failure. I'll look into trying to make this script a thing of the past by building it into the package install flow in `package.json`. I don't see why it is necessary right off. By removing this we'll also get built-in Windows support to makes things easier there.

[...]

PR #1309, opened 2016-11-24
Windows env fix

viswanathamsantosh:
Associated Issue: #1248

### Summary of Changes

*edited firefox-driver.js to run 'yarn start firefox'

### Test Plan

Example test plan:

*Navigate to debugger.html folder from command prompt from windows
*Execute "yarn start firefox" to start the server
*Open browser and go to localhost:8000

Output: debugger.html is up and running

yarn start firefox vs yarn run firefox

yarn2

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -116,7 +116,7 @@ $ /Applications/Firefox.app/Contents/MacOS/firefox-bin --start-debugger-server 6
 **Windows:**
 
 ```
-C:\Program Files (x86)\Mozilla Firefox\firefox.exe -start-debugger-server 6080 -P development
+C:\Program Files (x86)\Mozilla Firefox\firefox.exe -start-debugging-server 6080 -P development
 ```
 
 > If this command doesn't work for your OS or Firefox version see the other [Firefox commands for running in a debuggable state](./docs/remotely-debuggable-browsers.md#firefox)
--- a/bin/firefox-driver.js
+++ b/bin/firefox-driver.js
@@ -22,7 +22,7 @@ const useWebSocket = args.websocket;
 function binaryArgs() {
   return [
     (!isWindows ? "-" : "") +
-    "-start-debugger-server=" +
+    "--start-debugging-server=" +
     (useWebSocket ? "ws:6080" : "6080")
   ];
 }


jasonLaster:
CC @MoLow

MoLow:
@jasonLaster whats this?

jasonLaster:
Hi! We're seeing a couple issues with windows contributors getting started.
Mind taking a look and helping?

MoLow:
Hi, sorry for the delay in answering. In the meanwhile i tested on a computer with no admin priviligaes, so I could not install yarn, neither update node to > 6.9.0, so I tried running it directly, and this is the results:

image

I'l try it later today on my computer using npm run/yarn start/yarn run.

BTW, I havent tried yet, but this maybe a simple solution: https://www.npmjs.com/package/npm-run-path

clarkbw:
> yarn start firefox vs yarn run firefox

I believe with `yarn start firefox` yarn is just actually starting the server and it actually passes 'firefox' as an argument to the server.  

When you run `yarn run firefox` you're seeing the issue we seem to be having running Firefox on windows.

clarkbw:
Ok, and I just looked into this further.  I think the first solution is to update our instructions.

The error you're seeing in that screenshot is because you don't have the `geckodriver.exe` file in your local path.  Here are some quick instructions so solve this:

* Open https://github.com/mozilla/geckodriver/releases/latest
* Download the build file that matches your system (like `geckodriver-v0.11.1-win64.zip`)
* Create a local bin directory by doing `mkdir ~/bin`
* Extract the `gecokdriver.exe` file into the `bin` directory you just created

Now when you run `yarn run firefox` Firefox should start up.  

I am seeing a problem where you I don't believe the debugger server is running, looking into that now.

MoLow:
@clarkbw well, the question is  if this could be done automatically on npm install or on npm run firefox, since the exe is installed to node_modules\geckodriver, and it only needs to get into the path in runtime.

clarkbw:
> could be done automatically on npm install or on npm run firefox, since the exe is installed to node_modules\geckodriver, and it only needs to get into the path in runtime.

Yes, sorry.  What I have above shouldn't actually be the solution, just short term.  On the Mac this is done automatically and it should be on Windows as well.

jasonLaster:
@clarkbw perhaps look at npm run path issues for windows

clarkbw:
Oh, hmm.  I'm seeing `node_modules/.bin/geckodriver` where i believe it should be `node_modules/.bin/geckodriver.exe`

clarkbw:
Or, actually `geckodriver.cmd`

clarkbw:
The `.bin/` directory has files with the `.cmd` extension, which run just fine.  I found that if I change this line with a `s/exe/cmd/` then it works just fine.

https://github.com/SeleniumHQ/selenium/blob/ba56ad1ae0b98a1fe1efdd2163c62fc847950178/javascript/node/selenium-webdriver/firefox/index.js#L288

clarkbw:
I've been trying out this solution, creating the driver manually, but still not having success:

```js
  let options = new firefox.Options();
  let service = new firefox.ServiceBuilder(__dirname + '\\..\\node_modules\\.bin\\geckodriver.cmd').build();

  options.setProfile(firefoxProfile());
  options.setBinary(firefoxBinary());

  const driver = new firefox.Driver(options, service);
```

It finds the local gecko driver but errors out elsewhere before running anything.

MoLow:
well, now I ran it on another machine, and it seems the correct directory is prepended to PATH - only when running via `npm run firefox`;

this the result of `echo $PATH` and adding `console.log(process.env.PATH);` to firefox-driver.js

image

however, I created a file called "path.js" which only contains `console.log(process.env.PATH);`, and the directory is not prepended:
image

when I changethe `firefox` command in `package.json` to `node path.js`, the path is prepended again:

image

MoLow:
Mabe the simplest solution is to add these lines to firefox-driver.js?:

`
var path = require('path');
process.env.PATH = path.resolve(__dirname, '../node_modules/.bin') + ';' + process.env.PATH;
`

is this a BAD idea?
would it work in linux?

clarkbw:
> is this a BAD idea?
> would it work in linux?

As I understand it, the `.bin/` path *is* being added correctly, however it is looking for `geckodriver.exe` on windows and that `.exe` isn't found in that `.bin/` path.  

I think you need to do something like this: 
```js
if (isWindows) { 
var path = require('path'); 
process.env.PATH = path.resolve(__dirname, '../node_modules/geckodriver/bin') + ';' + process.env.PATH; 
}
```

PR #1312, opened 2016-11-25
use eslint prefer-template and move from string concatenation

clarkbw:
Associated Issue: n/a

In general template performance is about the same to slightly better, however I believe they are much more readable than string concatenation.  With the given eslint rule we can make the code base more consistent overall.

NOTE: This is one of this diff where I need lots of eyes because tiny changes in white space can really make a difference here.  I combed over it all carefully but I don't want to introduce errors here.

### Summary of Changes

* add `prefer-template` to the eslint rules
* move all instances of string concatenation to string templates

### Test Plan

Ran tests, played with the debugger in both todomvc and increment

### DIFF
--- a/src/actions/sources.js
+++ b/src/actions/sources.js
@@ -20,6 +20,7 @@ const {
 } = require("../utils/source-map");
 
 const { prettyPrint } = require("../utils/pretty-print");
+const { getPrettySourceURL } = require("../utils/source");
 
 const constants = require("../constants");
 const { removeDocument } = require("../utils/source-documents");
@@ -194,7 +195,7 @@ function togglePrettyPrint(sourceId: string) {
     assert(isGeneratedId(sourceId),
            "Pretty-printing only allowed on generated sources");
 
-    const url = source.url + ":formatted";
+    const url = getPrettySourceURL(source.url);
     const id = generatedToOriginalId(source.id, url);
     const originalSource = { url, id, isPrettyPrinted: false };
     dispatch({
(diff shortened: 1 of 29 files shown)


src/components/App.js
```diff
@@ -9,6 +9,8 @@ const { getSources, getSelectedSource } = require("../selectors");
 const { KeyShortcuts } = require("devtools-sham-modules");
 const shortcuts = new KeyShortcuts({ window });
 
+const cmd = `${cmdString()}+P`;
```
clarkbw: This value is per-OS so it won't change during runtime and really should be a const instead of evaluated in the render function.

clarkbw:
Additional resources for this change: 

* http://eslint.org/docs/rules/prefer-template
* https://jsperf.com/es6-string-literals-vs-string-concatenation

src/actions/sources.js
```diff
@@ -194,7 +195,7 @@ function togglePrettyPrint(sourceId: string) {
     assert(isGeneratedId(sourceId),
            "Pretty-printing only allowed on generated sources");
 
-    const url = source.url + ":formatted";
+    const url = getPrettySourceURL(source.url);
```
jasonLaster: 👍

src/components/Accordion.js
```diff
@@ -36,7 +36,7 @@ const Accordion = React.createClass({
   renderContainer: function(item, i) {
     const { opened, created } = this.state;
     const containerClassName =
-          item.header.toLowerCase().replace(/\s/g, "-") + "-pane";
+      `${item.header.toLowerCase().replace(/\s/g, "-")}-pane`;
```
jasonLaster: in general, I prefer moving these expressions out of a template, but that's a minor style preference for me.

clarkbw: Actually I agree. There is something about "more than one period" in a template that makes me feel like it's wrong.

PR #1314, opened 2016-11-27
Vertical layout

kdzwinel:
Associated Issue: #1266

### Summary of Changes

* switch debugger to the vertical layout when it's less than 700px wide

### Test Plan

- [ ] changing debugger width to 699px switches layout to vertical
- [ ] right sidebar can be resized, inside of the specified bounds (10px-80%), in both vertical and horizontal modes

### Screenshots/Videos

vert-true-false

### DIFF
--- a/packages/devtools-modules/client/shared/components/splitter/SplitBox.js
+++ b/packages/devtools-modules/client/shared/components/splitter/SplitBox.js
@@ -60,6 +60,12 @@ const SplitBox = React.createClass({
     };
   },
 
+  componentWillReceiveProps(nextProps) {
+    if (this.props.vert !== nextProps.vert) {
+      this.setState({vert: nextProps.vert});
+    }
+  },
+
   // Dragging Events
 
   /**
--- a/src/components/App.js
+++ b/src/components/App.js
@@ -10,7 +10,7 @@ const { KeyShortcuts } = require("devtools-sham-modules");
 const shortcuts = new KeyShortcuts({ window });
 
 const cmd = `${cmdString()}+P`;
-const verticalLayoutBreakpoint = window.matchMedia('(min-width: 700px)');
+const verticalLayoutBreakpoint = window.matchMedia("(min-width: 700px)");
 
 require("./App.css");
 require("./menu.css");
@@ -65,7 +65,9 @@ const App = React.createClass({
   },
 
   componentDidMount() {
-    verticalLayoutBreakpoint.onchange = () => this.setState({ vertical: verticalLayoutMatchMedia.matches });
+    verticalLayoutBreakpoint.onchange = () => this.setState({
+      vertical: verticalLayoutBreakpoint.matches
+    });
   },
 
   componentWillUnmount() {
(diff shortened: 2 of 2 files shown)


src/components/App.js
```diff
@@ -72,7 +82,8 @@ const App = React.createClass({
           splitterSize: 1,
           endPanelControl: true,
           startPanel: this.renderCenterPane(this.props),
-          endPanel: RightSidebar()
+          endPanel: RightSidebar(),
```
kdzwinel: name `RightSidebar` may be worth changing because in the vertical mode it's no longer on the right

src/components/App.js
```diff
@@ -55,6 +55,16 @@ const App = React.createClass({
...
   },
 
+  getInitialState() {
+    return {
+      vertical: window.matchMedia('(min-width: 700px)').matches
+    };
+  },
+
+  componentDidMount() {
+    window.matchMedia('(min-width: 700px)').onchange = event => this.setState({vertical: event.matches});
```
kdzwinel: I guess all this matchMedia magic should be moved to some small util lib like `KeyShortcuts` ?

packages/devtools-modules/client/shared/components/splitter/SplitBox.js
```diff
@@ -60,6 +60,12 @@ const SplitBox = React.createClass({
     };
   },
 
+  componentWillReceiveProps(nextProps) {
+    if (this.props.vert !== nextProps.vert) {
+      this.setState({vert: nextProps.vert});
+    }
```
kdzwinel: it should also be possible to modify `width` and `height` (at least according to the comment 15 lines above). Not sure if I should add it now though, since I do not plan on using it.

jasonLaster:
looks like the test failures are all linting things

packages/devtools-modules/client/shared/components/splitter/SplitBox.js
kdzwinel: so... should I add them (width/height) or not in your opinion? Sorry, I can't figure out what you are suggesting here 😊

src/components/App.js
kdzwinel: that was just a thought though, I don't think I should change it in this PR

src/components/App.js
```diff
@@ -57,6 +58,22 @@ const App = React.createClass({
...
   },
 
+  getInitialState() {
+    return {
+      vertical: verticalLayoutBreakpoint.matches
+    };
+  },
+
+  componentDidMount() {
+    verticalLayoutBreakpoint.onchange = () => this.setState({
```
kdzwinel: `verticalLayoutBreakpoint` has both `addEventListener` and `removeEventListener`, but then I'll need to store the callback function somewhere, so I decided to go with `onchange = `

jasonLaster:
@kdzwinel I'm blown away by how elegant this PR is.

I really like how you focused on flipping the `vertical` state and everything else flowed from there. You're a react natural!

kdzwinel:
@jasonLaster thanks 😊 glad you like it!

Issue #1315, opened 2016-11-27
[Editor] fix matched colors


jasonLaster:
We currently change the color of the text matched when we search. If you look at other editors you'll noticed that they just rely on the search outline and overlay for the affect. And that is less jarring and more semantically correct.

The issue is that when we match the token for the search result we give the element a class name `cm-selecting` instead of what it would naturally have, which could be `cm-property`

This patch fixes the text color, but of course removes the `cm-selecting` class which we need for showing the outline. 

```diff
diff --git a/src/utils/source-search.js b/src/utils/source-search.js
index e2e112b..897846d 100644
--- a/src/utils/source-search.js
+++ b/src/utils/source-search.js
@@ -63,7 +63,6 @@ function searchOverlay(query) {
       let match = query.exec(stream.string);
       if (match && match.index == stream.pos) {
         stream.pos += match[0].length || 1;
-        return "selecting";
       } else if (match) {
         stream.pos = match.index;
       } else {
```

So the challenge is how do we have both classes :)

Also, one bonus question is can we also style the highlighted match color white so that it is more visible in dark theme? Treat this as extra credit as it's more important to first solve the general question and then we can fix this later...

### screenshots

![](http://g.recordit.co/CmRLrEuCw6.gif)
![](http://g.recordit.co/OHT3Se7C8B.gif)

CC @bomsy I believe you would have some context here.
CC @Zacqary I believe this is the kind of bug you might enjoy :)

Zacqary:
Okay first of all, how do I open the Search in File box? ⌘+P lets me search for a filename, not within the current file

jasonLaster:
Ahh good question, we're currently disabling editor search.

So you'll need to go to `configs/local.json` and add `editorSearch` to the list of features. You can compare this with `configs/development.json`. After that you can do Cmd+f :)

Zacqary:
So am I missing something or is the fix actually this easy?
```diff
diff --git a/src/utils/source-search.js b/src/utils/source-search.js
index b12a675..c31bd1d 100644
--- a/src/utils/source-search.js
+++ b/src/utils/source-search.js
@@ -80,7 +80,7 @@ function searchOverlay(query) {
 function startSearch(cm, state, query) {
   cm.removeOverlay(state.overlay);
   state.overlay = searchOverlay(query);
-  cm.addOverlay(state.overlay, { opaque: true });
+  cm.addOverlay(state.overlay, { opaque: false });
 }
 
 /**
```

[...]

PR #1318, opened 2016-11-27
[Editor] Fix matched text color

Zacqary:
Associated Issue: #1315 

### Summary of Changes

* Set `opaque` param to `false` on the CodeMirror overlay for matched search terms

### Test Plan
- [x] Enable `editorSearch` in `configs/local.json`
- [x] Open the search panel on a source with Command-F
- [x] Search for something that occurs as a variable name or property name
- [x] The var/prop name should be outlined without changing the text color

### Screenshots/Videos

Before:
screen shot 2016-11-27 at 1 43 48 pm
After:
screen shot 2016-11-27 at 1 37 02 pm

### DIFF
--- a/src/utils/source-search.js
+++ b/src/utils/source-search.js
@@ -80,7 +80,7 @@ function searchOverlay(query) {
 function startSearch(cm, state, query) {
   cm.removeOverlay(state.overlay);
   state.overlay = searchOverlay(query);
-  cm.addOverlay(state.overlay, { opaque: true });
+  cm.addOverlay(state.overlay, { opaque: false });
 }
 
 /**


jasonLaster:
Thanks @Zacqary! Looks great

bomsy:
Good one!
Interesting effect of `opaque: false` to note.
It breaks up the outlining based on each of the tokens, therefore keeping the `cm-property` and the coloring.

ft

When `opaque: true`
It outlines them together, ignoring what type of token they are, therefore the colorings are removed!

ft2

Most editors (e.g atom) merge the outlines together, but it depends what we want ATM. We can fix the coloring issue now, see how to get the outlines together later.

jasonLaster:
This is a step in the right direction.

Merging and we can sort out the tokenizer business in a new PR.

Issue #1324, opened 2016-11-28
[Toolbox] Move internal tools to the devtools-local-toobox


jasonLaster:
We currently have many of our internal tools in the debugger project. These should move to the toolbox so that they can more easily be shared.

**Notes** 
* These tools are really only helpful when there's a second tool in Github, but it's good to plan ahead. 
* This also simplifies the debugger project and that's a very good thing.

**Tools**
* m-c bundler
* m-c importer
* mochitest runner
* unit test runner
* update CI docker image

jasonLaster:
Closing this issue as we'll be able to move the scripts over incrementally

PR #1264, opened 2016-11-19
Remove uneeded script by doing lerna bootstrap during postinstall.

Garbee:
Associated Issue: #1248 

### Summary of Changes

* Remove `preinstall` script
* Setup `lerna boostrap` as postinstall process.

### Test Plan

From a fresh install, run `npm install` or `yarn install` to test the changes. This allows Windows to run the install process as well without crashing out due to the `preinstall` being a bash script.

Issue #1251, opened 2016-11-18
[toolbox] Sanitize debugger.html dependencies


jasonLaster:
I spent the day sanitizing the console.html [dependencies](https://github.com/jasonLaster/console.html/blob/master/package.json#L11-L16), so that now the console can just depend on the toolbox. We should do the same thing for the debugger so that the debugger's dependencies are just for debugger needs.

I learned a **lot** along the way. Here's a summary

* the toolbox should not have many *devDependencies*. We were using *devDependencies* for packages that were needed by the console.html and inspector
* the toolbox should have all of the babel, flow, ... dependencies a tool needs. We can move all of those dependencies to the toolbox
* lerna **does not** install a linked packages dependencies. This means that the debugger does not install the toolbox's dependencies. This is a **big** problem. I see two possible solutions for this, but fortunately it's friday so i can think about it over the weekend
   * update lerna to install these packages, should in theory be possible
   * drop lerna and use relative paths i.e. `./packages/devtools-local-toolbox`.
     * **pros** you drop lerna and get full experience
     * **cons** you have to install on every change

Here's a [patch](https://gist.github.com/jasonLaster/af314f7a3c773ae7540014a0ee6bebac) for using relative paths. and [branch](https://github.com/devtools-html/debugger.html/compare/master...jasonLaster:rel-paths?expand=1)

jasonLaster:
Quick update: Now that we're using lerna as a post-install, it's very easy to start deleting dependencies from the debugger!

PR #1325, opened 2016-11-28
Sanitize Debugger dependencies

jasonLaster:
Associated Issue: #1251

This PR cleans up the debugger's dependencies. This was not possible before 
* @Garbee added lerna as a post-install hook https://github.com/devtools-html/debugger.html/pull/1264
* and we sanitized the toolbox's deps here https://github.com/devtools-html/debugger.html/pull/1249

Updates:
* There are plenty of debugger dependencies that can be killed
* When we move more Debugger internal tools to the toolbox a lot more debugger deps will move to the toolbox #1324 
* I didn't bump the toolbox version here because I don't think other tools need the bump, but I can do that later if we need it.

@juliandescottes mind taking a look and letting me know if you have any suggestions.


package.json
```diff
@@ -34,42 +34,17 @@
...
     "devtools-local-toolbox": "^0.0.10",
     "devtools-modules": "^0.0.9",
     "devtools-network-request": "^0.0.9",
     "devtools-sham-modules": "^0.0.9",
     "documentation": "^4.0.0-beta11",
     "eslint": "^3.10.0",
-    "eslint-plugin-babel": "^4.0.0",
-    "eslint-plugin-flowtype": "^2.20.0",
-    "eslint-plugin-mozilla": "0.2.3",
-    "eslint-plugin-react": "^6.7.1",
```
juliandescottes: .eslintrc still refers to rules from all those packages, are you sure we can remove them here ? Getting an error when running npm run lint after a clean yarn install.

juliandescottes:
> I didn't bump the toolbox version here because I don't think other tools need the bump, but I can do that later if we need it

I was planning on bumping it this week to get Windows env. support from #1308 . So if you want to do it here that's fine by me!

jasonLaster:
Hmm, it works for me locally 

and `ls node_modules/eslint-*` shows
```
eslint/                  eslint-plugin-babel/     eslint-plugin-flowtype/  eslint-plugin-mozilla/   eslint-plugin-react/
```

PR #1330, opened 2016-11-28
Show optimized out variables in object inspector

jlongster:
I had most of this done before break but didn't get it finished. This is actually pretty simple. It will show variables and arguments that have been optimized out with a special message:

screen shot 2016-11-28 at 11 39 50 am

### DIFF
--- a/src/components/ObjectInspector.css
+++ b/src/components/ObjectInspector.css
@@ -23,3 +23,7 @@ html .arrow.expanded svg {
 .arrow.hidden {
   visibility: hidden;
 }
+
+.object-node .unavailable {
+  color: var(--theme-body-color-inactive);
+}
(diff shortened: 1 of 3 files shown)


clarkbw:
> It will show variables and arguments that have been optimized out with a special message:

Do you have a test jsfiddle or something where I can play with it?

jlongster:
@clarkbw Sure, go to http://jlongster.com/s/foo.html and click the button. When it's paused, select the 2nd stack frame and look at the second block scope.

jlongster:
Let me know if there are any problems @clarkbw.

clarkbw:
It works, trying it out now.

> look at the second block scope

Just looking at this like its the first time.  I don't really understand why there is a 2nd block scope.  At the very least it would make more sense to me if they were given more context.

PR #1323, opened 2016-11-28
Extract devtools-client-adapter

jasonLaster:
Hey, @juliandescottes, mind taking a look at this toolbox PR?

This PR moves the `client` layer out of the `devtools-local-toolbox` package and into it's own package. It's pretty cool and I've been meaning to get to this PR for some.

It's the first version of a published client that connects to chrome, node, and firefox. 
* This will get better when we adopt `chrome-remote-interface` and can add a proper adapter on top of CRI.
* helps us replace tcomb types with exported client flow types
* dramatically simplifies the toolbox. Don't worry, the toolbox will get more complicated as we move more of the internal tools from the debugger to the toolbox https://github.com/devtools-html/debugger.html/issues/1324

PR #1332, opened 2016-11-28
fix local-toolbox error when chrome.debug is false

juliandescottes:
@jasonLaster 
Found a small issue introduced by #1295 while verifying #1323. 

When `chrome.debug` is false createTabs is called with a single argument `createTabs([])`. Added a default value. Would be fine with updating the different call sites as well, let me know if you want me to change it.


jasonLaster:
👍  nice fix. 

I was thinking about adding a default here.

PR #1335, opened 2016-11-28
Move some type definitions to Flow

ppold:
### Summary of Changes

* Move Breakpoint, BreakpointResult, Location, Source, Frame, Scope from tcomb to Flow.

### Test Plan

Make the types pass with Flow

### DIFF
--- a/packages/devtools-client-adapters/types.js
+++ b/packages/devtools-client-adapters/types.js
@@ -1,18 +1,6 @@
 // @flow
 
 /**
- * Source File Location
- *
- * @memberof actions/types
- * @static
- */
-export type Location = {
-    sourceId: string,
-    line: number,
-    column?: number
-};
-
-/**
  * Breakpoint
  *
  * @memberof actions/types
@@ -34,8 +22,20 @@ export type Breakpoint = {
  * @static
  */
 export type BreakpointResult = {
-    id: string,
+    ids: string,
     actualLocation: Location
+};
+
+/**
+ * Source File Location
+ *
+ * @memberof actions/types
+ * @static
+ */
+export type Location = {
+    sourceId: string,
+    line: number,
+    column?: number
 };
 
 /**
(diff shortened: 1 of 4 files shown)


packages/devtools-client-adapters/types.js
```diff
@@ -0,0 +1,90 @@
...
+};
+
+/**
+ * BreakpointResult
+ *
+ * @memberof actions/types
+ * @static
+ */
+export type BreakpointResult = {
+    ids: string,
```
jasonLaster: NOTE: we tried to make **flow** fail by adding this error, but when we run flow we got `no errors`

src/types.js
```diff
@@ -1,6 +1,96 @@
 // @flow
 
 /**
+ * Source File Location
```
jasonLaster: we tried doing something like this:

```js
import type { Breakpoint, Source } from "devtools-client-adapters/types";
export type { Breakpoint, Source };
```

This worked to the extent that we didn't have to copy the types and the debugger would mention when a type was missing, but it didn't help us w/r/t to validating the types.

jasonLaster:
CC @jlongster @montogeek

packages/devtools-client-adapters/types.js
jlongster: That's a blocker for this PR. :) We can't merge this until we know it's actually working. If it's not applying the types we have no idea if this PR actually works.

jasonLaster: 💯  sorry that wasn't clear.

jasonLaster:
ping @ppold will you be able to look at this?

src/components/Frames.js
```diff
@@ -15,7 +15,6 @@ import type { List } from "immutable";
 import type { Frame } from "../types";
 
 if (typeof window == "object") {
```
wldcordeiro: This whole block should be removed not left empty.

jasonLaster:
Hmm, things have moved a bit:
* client-adapter is now here: https://github.com/devtools-html/devtools-core/tree/master/packages/devtools-client-adapters
* i predict that the client adapter stuff will need to be manually re-applied so nothing is lost

jasonLaster:
closing this as most of the work will be done in devtools-core

PR #1345, opened 2016-11-29
[POC] show the reason why the debugger is paused

clarkbw:
Just a proof of concept for now, if anyone wants to take this further feel free to pick this up.  I think we want input from @helenvholmes first.  I'm not sure the sidebar is actually a good location, the way it pushes everything down doesn't seem like the best design.  A good path forward could be to move the logic out of the component and into a util function such that we can try a couple different variants of the components.

### Summary of Changes

* Create a new component that displays why the debugger is paused

### Test Plan

Example test plan:

- [x] Pause on breakpoint
- [x] Pause on debugger statement
- [x] Turn off feature flag (no element rendered)
- [x] Turn on feature flag (element rendered)
- [ ] Pause on exception
- [ ] Pause on ...

### Screenshots/Videos

screen shot 2016-11-29 at 8 02 50 am

screen shot 2016-11-29 at 8 02 46 am

### DIFF
--- a/src/components/WhyPaused.js
+++ b/src/components/WhyPaused.js
@@ -1,0 +1,50 @@
+const React = require("react");
+const { bindActionCreators } = require("redux");
+const { connect } = require("react-redux");
+const ImPropTypes = require("react-immutable-proptypes");
+const actions = require("../actions");
+const { getPause } = require("../selectors");
+const { DOM: dom } = React;
+
+require("./WhyPaused.css");
+
+function getPauseReason(pauseInfo) {
+  if (!pauseInfo) {
+    return null;
+  }
+
+  let reasonType = pauseInfo.getIn(["why"]).get("type");
+  if (!reasons[reasonType]) {
+    console.log("reasonType", reasonType);
+  }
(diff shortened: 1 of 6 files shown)


jasonLaster:
Thanks @clarkbw this is a cool idea and a great thing we can experiment with in 2017.

Could you put it behind a feature flag so that we can setup some experiments.

src/components/WhyPaused.js
```diff
@@ -0,0 +1,50 @@
...
+    pauseInfo: ImPropTypes.map
+  },
+
+  displayName: "WhyPaused",
+
+  render() {
+    const { pauseInfo } = this.props;
+    const reason = getPauseReason(pauseInfo);
+
+    return pauseInfo ?
```
jasonLaster: perhaps if we don't have a reason we don't show the highlight for now?

clarkbw: Good point, better to be quiet about what we are unsure about.

jasonLaster:
Really cool! 

Lets make sure we can schedule the design + UR time to setup this experiment. In Q1

PR #1346, opened 2016-11-29
Fix source maps

jasonLaster:
### Summary of Changes

* The application work broke source map loading locally (not in the panel)
* This fixes that by loading the config in the source map worker and having a good default.

### Test Plan

- [x] Test source maps
- [x] Test connecting to chrome (uses networkRequest)

CC @juliandescottes

### DIFF
--- a/packages/devtools-modules/client/shared/components/splitter/SplitBox.js
+++ b/packages/devtools-modules/client/shared/components/splitter/SplitBox.js
@@ -60,6 +60,12 @@ const SplitBox = React.createClass({
     };
   },
 
+  componentWillReceiveProps(nextProps) {
+    if (this.props.vert !== nextProps.vert) {
+      this.setState({vert: nextProps.vert});
+    }
+  },
+
   // Dragging Events
 
   /**
--- a/src/components/App.js
+++ b/src/components/App.js
@@ -10,6 +10,7 @@ const { KeyShortcuts } = require("devtools-sham-modules");
 const shortcuts = new KeyShortcuts({ window });
 
 const cmd = `${cmdString()}+P`;
+const verticalLayoutBreakpoint = window.matchMedia("(min-width: 700px)");
 
 require("./App.css");
 require("./menu.css");
@@ -57,6 +58,22 @@ const App = React.createClass({
     );
   },
 
+  getInitialState() {
+    return {
+      vertical: verticalLayoutBreakpoint.matches
+    };
+  },
+
+  componentDidMount() {
+    verticalLayoutBreakpoint.onchange = () => this.setState({
+      vertical: verticalLayoutBreakpoint.matches
+    });
+  },
+
+  componentWillUnmount() {
+    verticalLayoutBreakpoint.onchange = null;
+  },
+
   render: function() {
     return dom.div(
       { className: "debugger" },
@@ -74,7 +91,8 @@ const App = React.createClass({
           splitterSize: 1,
           endPanelControl: true,
           startPanel: this.renderCenterPane(this.props),
-          endPanel: RightSidebar()
+          endPanel: RightSidebar(),
+          vert: this.state.vertical
         })
       })
     );


src/utils/source-map-worker.js
```diff
@@ -18,6 +18,9 @@ const {
   isOriginalId
 } = require("./source-map-util");
 
+const { setConfig } = require("devtools-config");
+setConfig(DebuggerConfig);
```
jlongster: Are you sure this works in the panel? I remember trying to load the config before but ran into problems so I avoided it.

jasonLaster:
Hmm, it *is* failing, but not for obvious reasons

```diff
diff --git a/src/utils/source-map-worker.js b/src/utils/source-map-worker.js
index ade35e8..634410c 100644
--- a/src/utils/source-map-worker.js
+++ b/src/utils/source-map-worker.js
@@ -20,8 +20,11 @@ const {
   isOriginalId
 } = require("./source-map-util");

-const { setConfig } = require("devtools-config");
-setConfig(DebuggerConfig);
+console.log("YOOOOOOOOO\n\n\n\n");
+dump(`YOOOOOOOOO\n\n\n\n${ JSON.stringify(DebuggerConfig)}`);
+
+// const { setConfig } = require("devtools-config");
+// setConfig(DebuggerConfig);

 import type { Location, Source } from "../types";
```

`DebuggerConfig` is defined, but it looks like `setConfig` fails

```
dump(`YOOOOOOOOO\n\n\n\n${ JSON.stringify(({"title":"Debugger","environment":"development","baseWorkerURL":"http://localhost:8000/assets/build/","host":"http://localhost:8000","theme":"light","logging":{"client":false,"firefoxProxy":false,"actions":false},"features":{"watchExpressions":true,"editorSearch":true},"chrome":{"debug":true,"host":"localhost","port":9222},"node":{"debug":true,"host":"localhost","port":9229},"firefox":{"webSocketConnection":false,"proxyHost":"localhost:9000","websocketHost":"localhost:6080"},"development":{"serverPort":8000,"examplesPort":7999},"hotReloading":
```

jlongster:
I wish I could remember the problem. I just remember it ended up being a lot easier keeping this isolated and not pulling in the config, and passing down any config to it. Why is `host` a config variable?

jasonLaster:
> passing down any config to it

not sure how we solve for this...

> why is host a variable

the application does not know what the dev server host is, so it needs to be set

I think this is related to #1351 where the bundle has some weird module scope. I'm going to investigate soon.

jasonLaster:
@jlongster ahah, this was the problem you ran into `devtoolsRequire` is not a function in source map worker. 

We run into this problem because `devtools-config/feature` tries to `require("./test-flag")` which maps to a devtoolsRequire.

I fixed this by removing the `isTesting` function is feature.js and the one call site, which wasn't necessary.

jlongster:
> the application does not know what the dev server host is, so it needs to be set

What does it use `host` for though? Whatever it uses it for, it seems like that could be hardcoded as part of the build process for generating the app. The real product is going to be a downloading app, it's not going to need a running server anywhere.

I'd avoid adding the complexity of pulling in the config in the sourcemap server and removing `isTesting` just for a temporary fix for the app. The app can hardcode it if needed for now and when it's a normal app it will run the server itself.

jasonLaster:
Thanks for bringing up the question about `host`. This forced me to go back and look at some assumptions.

1. I think `host` is a bad name. A better name is `devServerOrigin`.

2. We can remove the config field and have a query:

```js
location.origin == "file://" ? "http://localhost:8000" : location.origin
```

the drawback with this approach is that the landing page also needs the dev server origin, so this would be duplicated logic.

3. how does this affect the app

At the moment, the electron app opens `file://index.html` which pulls in `localhost:8000/assets/bundle.js`. 
* This system is naive, but does avoid some issues with instantiating workers with a cross origin asset. 
* At the moment, we're running a separate `application.json` config when the app is running. This could also change.
* I'm going to play with an application development server as well as realistic options for hosted / cached assets at somepoint and will re-evaluate some of these decisions.

---

At the moment, I prefer keeping the `devServerOrigin` config field and removing the `devtools-config` isTesting piece. (what's in the PR now)

But, I'm open keeping the isTesting requirement and dropping the devServerOrigin field as that does require the source-map-worker to have the config. 

I don't feel too strongly here. Sorry for the book, there's a lot of context here :)

PR #1349, opened 2016-11-29
Make sourcemapped location resolve forward instead of back (fixes #1252)

### LINKED ISSUE #1252: sourcemapped breakpoints sliding up by one
I've been thinking I'm crazy but because this happens no matter where in the code I try to set a breakpoint I feel like this might actually be a systematic error with how we handle lines in the editor and pass that back to set a breakpoint.

I can't make this happen in the `increment` test page.  But if I run the debugger on something and then run the debugger on our debugger I can reproduce this every time.

Here's a gif of it in action, I'm looking in the `actions/sources` module.

breakpoints sliding up

jlongster:
The problem with sourcemapped breakpoints was that it was searching backwards to find the nearest generated location, which means if there is no mapping for column 0 is would move to the previous line. It's important that we resolve forwards, so it will find the first mapping on the same line (which is a common occurrence, to only map the first piece of code on the line that's indented).

### DIFF
--- a/src/utils/source-map-worker.js
+++ b/src/utils/source-map-worker.js
@@ -143,7 +143,8 @@ async function getGeneratedLocation(location: Location, originalSource: Source)
   const { line, column } = map.generatedPositionFor({
     source: originalSource.url,
     line: location.line,
-    column: location.column == null ? 0 : location.column
+    column: location.column == null ? 0 : location.column,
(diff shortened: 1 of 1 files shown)


clarkbw:
I've been trying to test this change against http://wbamberg.github.io/example-websites/source-mapping/index.html but I'm not seeing the source mapped files anymore.  I'll look more into it and file another issue.

PR #1351, opened 2016-11-29
fix clientType null when calling makeThunkArgs

juliandescottes:
This PR should fix adding breakpoints when debugging Chrome.

The devtools-client-adapters required in main.js (used to get {firefox, getClient}) was generating a separate instance from the one required in devtools-local-toolbox.

This first adapters instance was never properly initialized with a call to startDebugging, so when calling getClient in 
```javascript
  log: getValue("logging.actions"),
  makeThunkArgs: (args, state) => {
    return Object.assign({}, args, { client: getClient(state) });
  }
});
```

the clientType was always null and defaulting to the firefox client. For now I just forwarded the necessary exports from devtools-client-adapters to devtools-local-toolbox, so that we only ever create one instance of the adapters.

cc @jasonLaster

### DIFF
--- a/packages/devtools-local-toolbox/package.json
+++ b/packages/devtools-local-toolbox/package.json
@@ -66,6 +66,7 @@
     "react-redux": "4.4.5",
     "redux": "3.5.2",
     "serve-index": "^1.8.0",
+    "single-module-instance-webpack-plugin": "0.0.4",
     "style-loader": "^0.13.1",
     "svg-inline-loader": "^0.7.1",
     "tcomb": "^3.1.0",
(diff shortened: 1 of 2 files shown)


jasonLaster:
Hmm, this is a really interesting bug and we've seen it come up in other forms as well.

The underlying problem is that there are some bugs with how webpack builds the bundle, where you don't have a guarantee that a module is only bundled once. Consequently, it's possible that there are two places where `clientType` are defined. I'm looking into this as well with the source maps bug.

This fixes the problem, but we'd have to always use the same `getClient` function because of the issue so i'm not sure what to do here yet.

juliandescottes:
Was not sure if the behavior was expected or not, but I can spend some time trying to find a better solution than this workaround.

juliandescottes:
@jasonLaster alternatively, it looks like using https://www.npmjs.com/package/single-module-instance-webpack-plugin also fixes the issue here.

jasonLaster:
looks good, i'm going to try pulling it for the source maps issue as well

PR #1354, opened 2016-11-29
[Source Tree] Don't show empty root. Fixes #1338

### LINKED ISSUE #1338: [Source Tree] Don't show empty root
We shouldn't show an empty source tree domain. 

#### Steps to reproduce

1. debug https://www.google.com/?gws_rd=ssl

#### Patch

here are some hints on where to start digging

```diff

diff --git a/src/components/SourcesTree.js b/src/components/SourcesTree.js
index fe5ce95..a1474c3 100644
--- a/src/components/SourcesTree.js
+++ b/src/components/SourcesTree.js
@@ -12,6 +12,9 @@ const ManagedTree = React.createFactory(require("./utils/ManagedTree"));
 const Svg = require("./utils/Svg");
 const { throttle } = require("../utils/utils");

+// run appStore.getState().sources.toJS().sources in the console
+// try and figure out where the (index) is coming in
+
 let SourcesTree = React.createClass({
   propTypes: {
     sources: ImPropTypes.map.isRequired,
diff --git a/src/utils/sources-tree.js b/src/utils/sources-tree.js
index 9c85739..eed49c6 100644
--- a/src/utils/sources-tree.js
+++ b/src/utils/sources-tree.js
@@ -82,6 +82,7 @@ function getURL(source: TmpSource): { path: string, group: string } {

   const { pathname, protocol, host, path } = parse(url);

+  // => check here
   switch (protocol) {
     case "javascript:":
       // Ignore `javascript:` URLs for now
@@ -184,6 +185,8 @@ function addToTree(tree: any, source: TmpSource) {
   // Overwrite the contents of the final node to store the source
   // there.
   if (isDir) {
+    // this is where we add an index node
+    // for some reason the domain is likely empty when we do it...
     subtree.contents.unshift(createNode("(index)", source.get("url"), source));
   } else {
     subtree.contents = source;
```

### Screenshot
screen shot 2016-11-28 at 7 18 42 pm

Dalimil:
So the problem appeared whenever there was a **javascript:** URI - because in you getURL switch you do this: 
```
case "javascript:":
      // Ignore `javascript:` URLs for now
      return def;
```

but that just returns `def = { path: "", group: "" };` which is then merged into the tree as an entry with an empty string as its url.

### DIFF
--- a/src/utils/sources-tree.js
+++ b/src/utils/sources-tree.js
@@ -135,6 +135,7 @@ function addToTree(tree: any, source: TmpSource) {
 
   if (IGNORED_URLS.indexOf(url) != -1 ||
       !source.get("url") ||
+      !url.group ||
       isPretty(source.toJS())) {
     return;
   }
--- a/src/utils/tests/sources-tree.js
+++ b/src/utils/tests/sources-tree.js
@@ -186,6 +186,32 @@ describe("sources-tree", () => {
     expect(aFileNode.name).to.be("a.js");
   });
 
+  it("excludes javascript: URLs from the tree", () => {
+    const source1 = Map({
+      url: "javascript:alert('Hello World')",
+      actor: "actor1"
+    });
+    const source2 = Map({
+      url: "http://example.com/source1.js",
+      actor: "actor2"
+    });
+    const source3 = Map({
+      url: "javascript:let i = 10; while (i > 0) i--; console.log(i);",
+      actor: "actor3"
+    });
+    const tree = createNode("root", "", []);
+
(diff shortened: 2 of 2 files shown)


jasonLaster:
Thanks @Dalimil 

mind adding a unit test here as well
https://github.com/devtools-html/debugger.html/blob/master/src/utils/tests/sources-tree.js

I forgot to mention it in the description, but source trees are really complex and tests help :)

jasonLaster:
Great fix @Dalimil. That's a really cool edge case

PR #1362, opened 2016-11-30
Remove isTesting

jasonLaster:
This PR drops the `isTesting` query in `feature.js`.

I'm opening this PR outside of the source map discussion because it is easier to have separate conversations. 

Benefits:
* it removes the need for `devtools-config` to require `./test-flag` which adds a `devtoolsRequire` in the panel
* it simplifies `debugGlobal`, which makes the `window.cm` more explicit. (this is a good thing)
* it lets the `devtools-config` be used in workers

Drawbacks:
* we might want to check whether we're in a testing mode down the road.  
* `window.cm` is now always a global. I don't think this is a big deal, it's a smell to use a global in that way and unlikely that another tool will reach into the panel and invoke a code mirror method.

### DIFF
--- a/packages/devtools-config/src/feature.js
+++ b/packages/devtools-config/src/feature.js
@@ -1,8 +1,6 @@
 const pick = require("lodash/get");
 let config;
 
-const flag = require("./test-flag");
-
 /**
  * Gets a config value for a given key
  * e.g "chrome.webSocketPort"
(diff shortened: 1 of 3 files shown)


jlongster:
I think the flag will be useful down the road, and we've used a similar flag before in devtools. One use case is only enabling assertions in dev/test mode for performance reasons. I could definitely see us wanting to do that. There are are things that we probably could enable in dev/test mode. The reason it's different than dev mode is because because Firefox in production mode is run on try.

I think it's a good thing for workers not to require the config, and instead be explicitly configured. It makes them more isolated. Sourcemaps in fact will be moved to the toolbox so that's a perfect use case; it can't depend on the config but needs to be configured externally.

PR #1363, opened 2016-11-30
Hardcode origin

jasonLaster:
This fixes the issue with loading source maps by hardcoding the dev server origin.

I also updated the app to read directly from "http://localhost:8000" instead of "file://index.html" https://github.com/jasonLaster/debugger.app/commit/aa482bcf83bf1181837ce300995c1fc4f26b1720

* I wrote a book on `devServerOrigin` [here](https://github.com/devtools-html/debugger.html/pull/1346#issuecomment-263962730), which is not needed now
* I filed a PR on removing `isTesting` [here](https://github.com/devtools-html/debugger.html/pull/1362) which i still think is a good idea

### DIFF
--- a/packages/devtools-network-request/networkRequest.js
+++ b/packages/devtools-network-request/networkRequest.js
@@ -2,9 +2,8 @@
 // replaces a more powerful network request from Firefox that can be
 // configured.
 function networkRequest(url, opts) {
-  const devServerOrigin = location.origin;
   return Promise.race([
-    fetch(`${devServerOrigin}/get?url=${url}`)
+    fetch(`/get?url=${url}`)
       .then(res => {
         if (res.status >= 200 && res.status < 300) {
           return res.text().then(text => ({ content: text }));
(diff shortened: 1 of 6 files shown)


packages/devtools-network-request/networkRequest.js
```diff
@@ -1,13 +1,10 @@
-const { getValue } = require("devtools-config");
-
 // opts is ignored because this is only used in local development and
 // replaces a more powerful network request from Firefox that can be
 // configured.
 function networkRequest(url, opts) {
-  const devServerURL = getValue("host");
-
+  const devServerOrigin = location.origin;
```
jlongster: At this point why not just `fetch("/get....")`? which is how it was before the app stuff.

So the app is currently just showing the webpage from the server, yeah, which is why I didn't think this was needed. And this will work when it's locally running the server itself too.

Makes sense, thanks for thinking through it!

jasonLaster: yeah, we can go back to that :)

Issue #1305, opened 2016-11-24
Active tab's title header should not be separated by line


jnachtigall:
There shouldn't be a line below the title (file name) of the active tab 

![2016-11-24 10_29_48-yahoo suche websuche suchmaschine - nightly](https://cloud.githubusercontent.com/assets/3307404/20593304/1f7c010a-b232-11e6-885a-704304b2ea19.png)

If you look at Sublime the active tab is always not visually "disconnected" from its content:
![2016-11-24 10_36_28-tab 1 - sublime text 2 unregistered](https://cloud.githubusercontent.com/assets/3307404/20593237/e728c3c4-b231-11e6-826d-fafbd55f4c14.png)

Same for VSCode or even the normal Firefox tabs.

Would make it easier to see which is the active tab. Especially, for the light theme (where the color contrast between active and other tabs is really low, might be another issue though):
![2016-11-24 10_39_51-yahoo suche websuche suchmaschine - nightly](https://cloud.githubusercontent.com/assets/3307404/20593411/98bf4bd0-b232-11e6-9ede-864d47cf3dc4.png)
(The right tab is the active here, the left one is inactive - hard to see).

_____________

It's so minor, I'll leave this here: You might also want to check the separating line between line number and content. There's no such separating line in VSCode or Sublime and to me it irritates the eye especially when this line hits the tabs header:
![2016-11-24 10_48_58-yahoo suche websuche suchmaschine - nightly](https://cloud.githubusercontent.com/assets/3307404/20593769/e2639060-b233-11e6-9ff7-ae4bfd460ad6.png)

PS Couldn't find this already reported but searching dupes in Github is the same mess as on BMO ;) Please let me know, if you are already aware of such things and it's not useful or too early to file such bugs :)

jasonLaster:
Thanks @jnachtigall for reporting. This looks like a regression on our end.

We had implemented it here: https://github.com/devtools-html/debugger.html/pull/774

@wldcordeiro interested in investigating?

darkwing:
This appears to be fixed and ticket should be closed.

jnachtigall:
Thanks, I still think though that is valid from above:

> You might also want to check the separating line between line number and content. There's no such separating line in VSCode or Sublime and to me it irritates the eye especially when this line hits the tabs header.

Created #1394 for it for further discussion...

PR #1368, opened 2016-12-01
fixed active tab header seperation

liorpr:
As a follow up to #1305,

please share what you think,

@jasonLaster @wldcordeiro

### DIFF
--- a/src/components/SourceTabs.css
+++ b/src/components/SourceTabs.css
@@ -39,10 +39,9 @@
   border: 1px solid var(--theme-splitter-color);
   border-top-left-radius: 2px;
   border-top-right-radius: 2px;
-  height: 23px;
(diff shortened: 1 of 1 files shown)


jasonLaster:
Thanks @liorpr!

Mind giving some context on what wrong and what the fix was?

This looks like a mystery to me!

jasonLaster:
Thanks @liorpr!

Mind giving some context on what wrong and what the fix was?

This looks like a mystery to me!

liorpr:
It seemed like the border of the editor div(where the code is mounted with CodeMirror) was taking over so I just subtracted 1px from the tab and added 1px  as a (transparent) border

jasonLaster:
Looks great in firefox and chrome

Issue #1337, opened 2016-11-29
[Source Tree] show first root


jasonLaster:
One nice thing that chrome does is only expand the first source tree root. This is nice because, they do some work to make sure it is *often* the most helpful and it helps focus the user on where to take an action.

#### Steps to reproduce

1. debug google.com

#### Patch

This patch should be everything we need. 

```diff
diff --git a/src/components/SourcesTree.js b/src/components/SourcesTree.js
index fe5ce95..03e24cd 100644
--- a/src/components/SourcesTree.js
+++ b/src/components/SourcesTree.js
@@ -135,7 +135,9 @@ let SourcesTree = React.createClass({
       getRoots: () => sourceTree.contents,
       getKey: (item, i) => item.path,
       itemHeight: 30,
+      autoExpand: 0,
       autoExpandDepth: 1,
+      autoExpandAll: false,
       onFocus: this.focusItem,
       renderItem: this.renderItem
     });
```

#### Chrome
![screen shot 2016-11-28 at 7 20 23 pm](https://cloud.githubusercontent.com/assets/254562/20691565/dcf9d61a-b59f-11e6-9006-926ba79f39c9.png)

#### Firefox
![screen shot 2016-11-28 at 7 18 42 pm](https://cloud.githubusercontent.com/assets/254562/20691577/e2846f82-b59f-11e6-94ef-7d1db71feb69.png)

Dalimil:
So first, the `devtools-sham-modules` tree component with `autoExpand(roots[0], 0);` will throw an undefined error for an empty `roots` array so that had to be fixed.
Then I was trying to implement a source tree ordering that would be similar to the one used in Chrome - they keep the root domain on top and then only alphabetically sort the rest. (see the attached screenshot - `www.mozilla.org` needs to go before `www.google-analytics` even though `g` is before `m` ). 
![screenshot_2016-12-01_22-26-54](https://cloud.githubusercontent.com/assets/4202539/20815811/5b077912-b817-11e6-873b-a02cd46926fe.png)
I just pinned the first group of sources to the top - which in most cases happens to be the domain corresponding to the opened tab but I think to implement this properly you'd need to change the way that reducers mergeIn a new source on `ADD_SOURCE` action or somehow keep track of the tab URL so that it can be compared later...

jasonLaster:
This was done in #1371

PR #1371, opened 2016-12-01
[Source Tree] show first root #1337

Dalimil:
See https://github.com/devtools-html/debugger.html/issues/1337#issuecomment-264318724
screenshot_2016-12-01_22-26-54

### DIFF
--- a/packages/devtools-sham-modules/client/shared/components/tree.js
+++ b/packages/devtools-sham-modules/client/shared/components/tree.js
@@ -249,7 +249,7 @@ const Tree = module.exports = createClass({
       for (let i = 0; i < length; i++) {
         autoExpand(roots[i], 0);
       }
-    } else {
+    } else if (length != 0) {
       autoExpand(roots[0], 0);
     }
   },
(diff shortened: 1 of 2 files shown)


jasonLaster:
This looks really good.

Mind trying to add some unit tests here for the source-tree utils?

Roping in @gregtatum who has lots of source-tree experience.

Dalimil:
But it's not ideal, because it always picks the domain of the first script it receives as the top one... Keeping track of the tab url in the sources-tree would require more work + I'm not sure which group should be expanded when a website has all its scripts hosted externally.

jasonLaster:
Haha that's a really interesting problem. 

Perhaps we create a new issue for sorting domains. I don't want to clutter those issue

jasonLaster:
https://github.com/devtools-html/debugger.html/issues/1374

jasonLaster:
Sorry that this sat @Dalimil I was traveling on the past three days :/

This is some great work, but I'm noticing that the "first source" rule might be racey. 

I got this when i went to mozilla.com.

screen shot 2016-12-06 at 1 13 33 pm
screen shot 2016-12-06 at 1 12 20 pm

How about with this PR we drop the domain sorting and then do it properly in a new issue. I agree that we want to use the tab URL as the point of reference. I'll stub out a reducer update that stores it.

Dalimil:
OK, makes sense.

jasonLaster:
Thanks @Dalimil!

PR #1373, opened 2016-12-02
handle the function keys by preventing the default action, fixes #1367

### LINKED ISSUE #1367: [Windows] F10 opens Menu instead of "Stepping Over"
If you press F10 under Windows 7 and 10, then the Firefox Menu opens. In order to step over an instruction I have to use the mouse instead of the F10 shortcut. 

This was also an issue in the old debugger, see https://bugzilla.mozilla.org/show_bug.cgi?id=923971 (uh, I just see that yesterday someone asked to reopen it)

clarkbw:
Associated Issue: #1367

### Summary of Changes

* add `preventDefault` and `stopPropogation`

### Test Plan (tested in Windows only)

- [x] Pause on breakpoint press F10 to step over

### DIFF
--- a/src/components/CommandBar.js
+++ b/src/components/CommandBar.js
@@ -63,11 +63,10 @@ const CommandBar = React.createClass({
     const shortcuts = this.context.shortcuts;
     const handleEvent = (e, func) => {
       e.preventDefault();
-      if (e.stopPropogation) {
-        e.stopPropogation();
-      }
+      e.stopPropagation();
       func();
     };
+
     shortcuts.on("F8", (_, e) => handleEvent(e, this.props.resume));
     shortcuts.on("F10", (_, e) => handleEvent(e, this.props.stepOver));
     shortcuts.on(`${ctrlKey}F11`, (_, e) => handleEvent(e, this.props.stepIn));
--- a/src/components/CommandBar.js
+++ b/src/components/CommandBar.js
@@ -61,26 +61,18 @@ const CommandBar = React.createClass({
 
   componentDidMount() {
     const shortcuts = this.context.shortcuts;
-    shortcuts.on("F8", (_, e) => {
+    const handleEvent = (e, func) => {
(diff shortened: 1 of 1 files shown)


clarkbw:
> Can we avoid stop propagation as we're listening on the window? Not much higher to go!

I tried this but I believe if you don't stop propagation the event is bubbled out of the window context because this is a key the browser / OS handles.  I tried using just one or the other and it required both to be used in order to stop the default action from happening.

> Also I wonder if you could add a helper
> 
> stopPropogation(() => this.stepIn())

Sounds like a good plan.  Will update soon.

src/components/CommandBar.js
```diff
@@ -53,18 +53,26 @@ const CommandBar = React.createClass({
...
 
   componentDidMount() {
     const shortcuts = this.context.shortcuts;
-    shortcuts.on("F8", this.props.resume);
-    shortcuts.on("F10", this.props.stepOver);
-    shortcuts.on(`${ctrlKey}F11`, this.props.stepIn);
-    shortcuts.on(`${ctrlKey}Shift+F11`, this.props.stepOut);
+    const handleEvent = (e, func) => {
+      e.preventDefault();
+      if (e.stopPropogation) {
```
jasonLaster: strange, i've never seen this `stopPropogation` not present. What case are you handling here?

clarkbw: I was seeing an error in the console complaining that `stopPropogation` didn't exist while working on my Mac and I wasn't see on Windows.  I'll be able to check it out on both this week some time.  Not sure what's really going on here.

jasonLaster:
I dropped the `if` because it's part of the DOM spec lol. if we break anything we can fix it :)

https://developer.mozilla.org/en-US/docs/Web/API/Event/stopPropagation

jasonLaster:
closing this in favor of #1455

PR #1380, opened 2016-12-06
Event Listeners (part 1)

jasonLaster:
Associated Issue: ???

### Summary of Changes

* I paired with @quodlibetor to create a minimal event listeners UI
* I added support for adding/removing associated breakpoints as well
* The styling and UX is all very prototypy and will definitely change

### Test Plan

- [ ] check to see if the list is complete
- [ ] does the checkboxes / close button work as expected
- [ ] make sure it's flagged off :)

### DIFF
--- a/src/components/RightSidebar.js
+++ b/src/components/RightSidebar.js
@@ -11,6 +11,7 @@ const Breakpoints = React.createFactory(require("./Breakpoints"));
 const Expressions = React.createFactory(require("./Expressions"));
 const Scopes = React.createFactory(require("./Scopes"));
 const Frames = React.createFactory(require("./Frames"));
+const EventListeners = React.createFactory(require("./EventListeners"));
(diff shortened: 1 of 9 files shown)


quodlibetor:
Is "the list" in the first checkbox "the list of event listeners"?

jasonLaster:
yeah, the list is the list of event listeners

src/components/RightSidebar.js
```diff
@@ -72,6 +73,14 @@ const RightSidebar = React.createClass({
       { header: L10N.getStr("scopes.header"),
         component: Scopes }
     ];
+
+    if (isEnabled("eventListeners")) {
+      items.push({
+        header: "EventListeners",
```
wldcordeiro: Here we'd want `Event Listeners` instead no (probably an L10N string would be better.)

jasonLaster: Yea, good call

Issue #450, opened 2016-08-01
Move object inspector to standalone module


linclark:
As we've talked about before, the console would benefit greatly from reusing the object inspector. I believe @jlongster planned to and/or did make this possible architecturally by separating the object inspector out from the sources tree.

In order for the console to reuse it, we'll need to make it possible to import it. I'm thinking it makes sense for it to be a separate repo `devtools-html/object-inspector`. It wouldn't need to be on npm for this to work (though it could be). This project's package.json could just use the repo name to install it. And then we could add the module to the `vendor` dir in Firefox for console to use it.

I'm not sure how difficult it's going to be to disentangle it from any dependencies it has, though. Any thoughts?

jlongster:
Shouldn't be too hard. As you said, I've been planning on it. I think the main thing it depends on is reps. So this will force us to figure out how to manage shared dependencies generally. Oh, it also depends on the tree... this is all good stuff to figure out, but it might take some time. I'm focusing on landing the debugger behind a pref right now, but I will look at this soon.

jasonLaster:
@jlongster do you think we'll be able to take this on this sprint or next?

jasonLaster:
We have some more options now that the packages work is done. With that said, I don't know if this is a Q4 priority so we might want to close it for the time being.

PR #1385, opened 2016-12-06
Move Reps to a package

jasonLaster:
Associated Issue: #450

### Summary of Changes

* removes the devtools-modules reps
* starts using [devtools-reps](https://github.com/jasonLaster/devtools-reps)

next steps
- [ ] use object inspector

### Test Plan

- [ ] test locally
- [ ] test in panel


jasonLaster:
closing this PR as we're not sure if we're going to use devtools-reps yet. Should be easy to re-open later.

PR #1304, opened 2016-11-24
Accessible buttons

arthur801031:
#1269

PR #1386, opened 2016-12-07
Fix styling for watch expression buttons.

wldcordeiro:
Associated Issue: #N/A but somewhat #1304 

### Summary of Changes

* remove default styling from buttons give by browsers

### Test Plan

Example test plan:

- [x] Open the debugger and see the button styling is removed

screenshot_20161206_175320

### DIFF
--- a/src/components/Accordion.js
+++ b/src/components/Accordion.js
@@ -1,7 +1,7 @@
 const React = require("react");
 const { DOM: dom, PropTypes } = React;
 
-const { div, span } = dom;
+const { div } = dom;
 const Svg = require("./utils/Svg");
 
 require("./Accordion.css");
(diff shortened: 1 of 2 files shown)


src/components/Accordion.js
```diff
@@ -47,8 +47,8 @@ const Accordion = React.createClass({
         Svg("arrow", { className: opened[i] ? "expanded" : "" }),
         item.header,
         item.buttons ?
-        dom.button({ className: "header-buttons" },
-          item.buttons.map((button, id) => span({ key: id }, button))
+        dom.span({ className: "header-buttons" },
```
jasonLaster: is this fine semantically? or should we use a div?

@wldcordeiro, @Garbee advice?

src/components/Accordion.js
```diff
@@ -47,8 +47,8 @@ const Accordion = React.createClass({
...
+          item.buttons.map(button => button)
```
jasonLaster: what does this map get us?

src/components/Accordion.js
wldcordeiro: @jasonLaster I don't think `div`/`span` have much semantic meaning in their own right, the `span` was just easier because it was one previously.

PR #1370, opened 2016-12-01
Start "Reveal in Tree"

jasonLaster:
Associated Issue: #1290 

### Summary of Changes

* add redux backend
* connect source tree

### Test Plan
- [x] load page
- [x] open tab context menu

Issue #1290, opened 2016-11-22
add "Reveal in Sources" item to the tab context menu


clarkbw:
This is a feature I use and like in Chrome.  I started down the path toward making this happen and then realized this might be harder than it originally looked as I don't see an easy way to pass the tab id back over to the sources tree.

Also, at a higher level I wonder if when you search for a file we should match opening the file with expanding the tree.  I think in a number of cases that interaction could break down, but at least in the initial case it is the behavior I want.

This just adds the menu item.  After that you need to figure out how you're going to pass a function from the tree down to the `SourceTabs`.

```diff
diff --git a/src/components/SourceTabs.js b/src/components/SourceTabs.js
index 20a7e8e..3259a31 100644
--- a/src/components/SourceTabs.js
+++ b/src/components/SourceTabs.js
@@ -114,7 +114,17 @@ const SourceTabs = React.createClass({
       click: () => closeTabs(tabs)
     };
 
+    const revealInSourcesMenuItem = {
+      id: "node-menu-reveal-in-sources",
+      label: L10N.getStr("sourceTabs.revealInSources"),
+      accesskey: "R",
+      disabled: false,
+      click: () => { /* setExpanded(tab) */ }
+    };
+
     showMenu(e, buildMenu([
+      { item: revealInSourcesMenuItem },
+      { item: { type: "separator" }},
       { item: closeTabMenuItem },
       { item: closeOtherTabsMenuItem, hidden: () => tabs.size === 1 },
       { item: closeTabsToRightMenuItem, hidden: () =>
diff --git a/src/strings.json b/src/strings.json
index a77308c..3ea14cb 100644
--- a/src/strings.json
+++ b/src/strings.json
@@ -40,5 +40,6 @@
   "sourceTabs.closeOtherTabs": "Close others",
   "sourceTabs.closeTabsToRight": "Close tabs to the right",
   "sourceTabs.closeAllTabs": "Close all tabs",
+  "sourceTabs.revealInSources": "Reveal in Sources",
   "loadingText": "Loading\u2026"
 }
```

![screen shot 2016-11-22 at 2 10 59 pm](https://cloud.githubusercontent.com/assets/2134/20544190/a6c3d302-b0bd-11e6-9e2c-01ece9e61ad4.png)

Here's what I have in case someone else wanted to pick this up.

jasonLaster:
CC @arthur801031  might be fun

```diff

[...]

PR #1403, opened 2016-12-08
"Reveal in Tree"

arthur801031:
Associated Issue: #1290 #1370

### Summary of Changes

* Add code in `SourcesTree.js` to determine which item(s) should be expanded and pass the list to `ManagedTree.js`.
* Add code in `ManagedTree.js` to process the list of items that should be expanded.


src/components/SourcesTree.js
```diff
@@ -38,6 +58,32 @@ let SourcesTree = React.createClass({
...
+      const itemsStrings = returnItemsStrings(sourceURL);
+      const processIndex = 0;
+
+      // Determine which item(s) should be added to itemsList
+      while (processIndex < itemsStrings.length) {
+        const i = 0;
+        const found = false;
+        while (!found) {
+          if (itemsStrings[processIndex] == sourceTreeList.contents[i].path) {
+            addToItemsList(sourceTreeList.contents[i], true);
```
jasonLaster: hmm, I think you can use react `setState` here instead of redux.

Also, maybe set the items at the end as opposed to one at a time

src/components/SourceTabs.js
```diff
@@ -114,13 +116,27 @@ const SourceTabs = React.createClass({
       click: () => closeTabs(tabs)
     };
 
-    showMenu(e, buildMenu([
+    const showSourceMenuItem = {
```
jasonLaster: I landed this work earlier today (I think). A rebase would be healthy either way :)

arthur801031:
Hi @jasonLaster

I have resolved all the previous issues. Please check!
I have finished the following tasks:
- [x] Rebase
- [x] Resolve failing test
- [x] Remove the list of items from redux
- [x] Use `setState` instead of redux

src/components/SourcesTree.js
```diff
@@ -17,6 +17,19 @@ const actions = require("../actions");
 const Svg = require("./utils/Svg");
 const { throttle } = require("../utils/utils");
 
+function returnItemsStrings(url) {
+  url = url.replace("http:/", "");
+  url = url.replace("https:/", "");
```
jasonLaster: perhaps  `url.replace(/http(s)?:\//, "");`

src/components/SourcesTree.js
```diff
@@ -17,6 +17,19 @@ const actions = require("../actions");
...
+  const itemsStrings = [];
+  for (const i = 1; i < url.length; i++) {
+    if (url[i] == "/") {
+      itemsStrings.push(url.substring(0, i));
+    }
+  }
```
jasonLaster: could we do the same thing `with url.split("/")` :)

src/components/SourcesTree.js
```diff
@@ -44,6 +57,33 @@ let SourcesTree = React.createClass({
...
+        while (!found) {
+          if (itemsStrings[processIndex] == sourceTreeList.contents[i].path) {
+            tempList.push(sourceTreeList.contents[i]);
+            sourceTreeList = sourceTreeList.contents[i];
+            found = true;
+            processIndex++;
+          }
+          i++;
+        }
+      }
```
jasonLaster: this is a little confusing to me... i'll play with it later

arthur801031:
Hi @jasonLaster 

No worries :) 
I've made the changes that you suggested. Please check!

jasonLaster:
@Dalimil, @bomsy, @gregtatum, @jlongster mind taking a look at this PR?

I think we can simplify the logic for expanding the tree, buy you guys have more experience than I do.

src/components/SourcesTree.js
Dalimil: I don't think that using while loops is a good idea - it's difficult to follow the indices and it might as well run forever.

I've never run this code but the `const` reassignments look very suspicious...

Dalimil:
I don't think the code works... Constant variable reassignments will produce TypeErrors I think...

arthur801031:
Hi @jasonLaster

I've been testing my code on http://cnn.com, and I found some cases which my previous code would crash. They have now been resolved with the exception of `extensions://`. This requires writing a special case. Please check and let me know if I need to handle this case. Thank you! :)

jasonLaster:
CNN is a great test case. 

I do think we should handle extensions:// and I suppose javascript:// 

This would be great example of where we can write don't unit tests to build confidence. I'll be at my desk tomorrow and will help you stay them. I'm also optimistic about simplifying the code. But tests will help there:)

arthur801031:
Hi @jasonLaster 

- I have added the code to handle `extensions://`, but I haven't come across a folder named `javascript://`, so I haven't implemented that.
- I have tested my code on http://cnn.com, https://www.cnet.com/, http://www.wsj.com/, and every file that I tested could be revealed in the tree.
- While I was testing the function on those websites, I saw errors in the console, so I commented out the code that I've written, and the program still produced those errors when I click on any folders and files.
screen shot 2016-12-12 at 2 01 19 pm

- Please let me know where unit tests should be placed and how to write them. I'm not really familiar with them. Thank you very much! :)

jasonLaster:
Thanks @arthur801031 for the extensive testing. I think you'll like the testing aspect, it's our way of quickly working on a complicated feature like reveal and making sure that we can test all of the edge cases and then make changes quickly.

* I took the liberty of moving `getExpandedItems` to our `utils/sources-tree` module where we keep the utility functions for the tree.
* I added one test that you can run with either `yarn run test` or `yarn run mocha-server` and then see in the browser*. You should add lots more tests
* While working on the test I thought of a **MUCH** simpler approach to `getExpandedParts`. The way I understand your algorithm works, is that it tries to find each part of the path as you traverse down the tree. Here's what I propose: lets update the tree so that: a) it's easy to get the leaf nodes (sources), b) each node has a parent reference so you can get the ancestor directory. This way, we can re-write our function as a) find the source in the tree with the same path and then b) get it's ancestor directories.


[1] You'll need to apply this [gist](https://gist.github.com/jasonLaster/da3f9b855a732c5a42abfa2b6d10b5df) to see the tests in the browser, the runner sadly needs some love. You can also change your test to `it.only` to only run that test.

jasonLaster:
Alright, so i messed around for a bit and this is what I learned. 

* The tree has a really helpful thing called a parentMap. the parentMap will help us find the a parent of any node, given the node
* the parentMap uses the tree nodes as keys, so we have to add a new `findSource` function for finding leaves
* we can get the directory path by using the parentMap
* this should be very similar to what expandedPaths did, but there might be edge cases! needs lots of testing :)

I pushed a bunch of new code that probably doesn't work :) didn't run it.

arthur801031:
Hi @jasonLaster ,

I took a look at the `getDirectories ` method. From what I understand, `findSource` finds the leaf/source node by going into each directory and traverse through that directory's folders and files before moving onto the next directory. For example, if our target file is file2, `findSource` would check dir1 -> subdir1 -> subdir2 -> file1 -> dir2 -> subdir1 -> file2 to find the target file. Then, `getDirectories ` method would use that target file's leaf node to find its ancestors, which means processing file2 -> subdir1 -> dir2.
```
root
  |
  - - ->dir1
             |
             --->subdir1
             --->subdir2
                        |
                        ---> file1
  - - ->dir2
             |
             --->subdir1
                        |
                        ---> file2
```

`getExpandedItems`, on the other hand, traverse dir1 -> dir2 -> subdir1 -> file2, and while it finds dir2, subdir1, and file2, the program adds them to the array. So, I think `getExpandedItems` is faster and doesn't check other irrelevant folders and files and does not traverse back to find the target leaf's ancestor items.

Please let me know if this is correct. Thank you! :)

jasonLaster:
Hi, @arthur801031 that's a good point. I don't think it should be a performance problem though, and if it is, we can always optimize it by maintaining a list of references to the leaf nodes.

jlongster:
This does seem like a lot of complicated code so it'd take me a while to think about how I'd implement it to give reasonable feedback, so I don't have much to say right now. I'm not sure why in `componentWillReceiveProps` of `SourcesTree` you can't do a search of the entire tree for a specific source item and build a list of nested sources to expand when you find the right source. You wouldn't have to add anything to `utils/source-tree.js` then; you already have the source tree structure, so just search it?

arthur801031:
I just finished `findSource ` and `getDirectories`, and I've tested on http://cnn.com. It seems to be working without any issue. Please check! Thank you!

I tried to git squash my commits, but after I saved the file, I received this error:
```
error: could not apply 73add1d... Add show source redux backend (#1370)
When you have resolved this problem, run "git rebase --continue".
```
Do I need to manually resolve the conflicts?

arthur801031:
Hi @jasonLaster ,

I have cleaned up the code and added 1 additional unit test for reveal. I have tested the code on  http://cnn.com and https://www.cnet.com/. If you would like the code to be simpler, you could remove the `else if ` from line 280 ~ line 287 in `findSource`. This special case only came up 3 times when I tested the code on those 2 websites. Please check. Thank you! :)

special2

special2-cnet

special2-cnet2

jasonLaster:
Thanks @arthur801031! I just squashed your work which was crazy in its own right :)

Will review now some of the other details

src/components/utils/ManagedTree.js
```diff
@@ -10,10 +10,29 @@ let ManagedTree = React.createClass({
...
       expanded: new Set(),
-      focusedItem: null
+      focusedItem: null,
+      listItems: []
     };
   },
 
+  componentWillReceiveProps(nextProps) {
+    this.setState({ listItems: nextProps.listItems });
+  },
```
jasonLaster: do we need state here? I think we can rely on just using `props.listItems`

src/components/utils/ManagedTree.js
```diff
@@ -10,10 +10,29 @@ let ManagedTree = React.createClass({
...
+
+  componentDidUpdate(nextProps) {
+    const listItemsLocal = this.state.listItems;
+    if (listItemsLocal && listItemsLocal.length > 0) {
```
jasonLaster: if we don't have a listItemsLocal , lets bail

src/components/utils/ManagedTree.js
```diff
@@ -10,10 +10,29 @@ let ManagedTree = React.createClass({
...
+      if (listItemsLocal.length == 1) {
+        this.focusItem(listItemsLocal[0]);
+      }
+      else {
+        this.expandListItems(listItemsLocal);
+      }
+      listItemsLocal.splice(listItemsLocal.length - 1, 1);
+      this.setState({ listItems: listItemsLocal });
```
jasonLaster: this seems weird to me... perhaps we do need some state, but the name collision makes it hard to follow...

Also, if we have an expanded list, do we need this listItemsLocal list... and if it does to an empty list because it keeps getting spliced down, can we do this another way. I guess i'm saying, this seems weird :)

src/components/utils/ManagedTree.js
```diff
@@ -31,6 +50,13 @@ let ManagedTree = React.createClass({
     }
   },
 
+  expandListItems(itemsList) {
+    const key = this.props.getKey(itemsList[itemsList.length - 1]);
+    const expanded = this.state.expanded;
+    expanded.add(key);
+    this.setState({ expanded: expanded });
```
jasonLaster: same weirdness here. maybe we can do something where we look at the listItem props and then update the expanded set in batch?

arthur801031:
Hi @jasonLaster 

I simplified the code for reveal in ManagedTree.js. Please check. Thank you very much for your help on this PR!! :)

src/components/utils/ManagedTree.js
```diff
@@ -10,10 +10,22 @@ let ManagedTree = React.createClass({
...
     return {
       expanded: new Set(),
-      focusedItem: null
+      focusedItem: null,
+      treeListItems: []
     };
   },
 
+  componentWillReceiveProps(nextProps) {
+    this.setState({ treeListItems: nextProps.listItems });
```
jasonLaster: Hmm, what if we call expand list items from here. No state necessary.

arthur801031:
Hi @jasonLaster ,

Great comments/suggestions! Please check. Thank you!

src/components/SourcesTree.js
```diff
@@ -83,6 +92,12 @@ let SourcesTree = React.createClass({
     if (!nodeHasChildren(item)) {
       this.props.selectSource(item.contents.get("id"));
     }
+
+    // If listItems is not empty, focusedItem would be set incorrectly in
+    // ManagedTree.js's expandListItems function
+    if (this.state.listItems && this.state.listItems.length > 0) {
+      this.setState({ listItems: [] });
+    }
```
jasonLaster: hmm, @arthur801031 this *also* seems strange.

I think it's here because `componentWillReceiveProps` does not check to see if `listItems` has changed. 
* perhaps check to see if listItems is different
* also could, `componentWillReceiveProps` be `componentDidUpdate` perhaps not, just curious.

src/components/utils/ManagedTree.js
```diff
@@ -31,6 +38,16 @@ let ManagedTree = React.createClass({
     }
   },
 
+  expandListItems(treeListItems) {
+    const expanded = this.state.expanded;
+    for (let i = treeListItems.length - 1; i >= 0; i--) {
+      const key = this.props.getKey(treeListItems[i]);
+      expanded.add(key);
+    }
```
jasonLaster: you can clean this up by doing:

```js
treeListItems.forEach(item => this.state.expanded.add(this.props.getKey(item))
```

Not a big deal, but you know, we're perfectionists here :)

arthur801031:
Hi @jasonLaster ,
Please check. Thank you!! :)

jasonLaster:
Wow. At this rate we'll have a three line PR by Monday. 

Seriously though, will treat and likely never tomorrow morning.

Issue #1375, opened 2016-12-04
Debugger crashes on sparse local storage


bjorn3:
When I tried debugging some javascript with firefox developer edition debugger.html constantly crashed at the same point in my code. I got a popup saying debugger.html was running too long. If I tried to use the browser I couldn't use the menu button or any other button of for example noscript. When I finally tried to close firefox I got display error in the browser window and had to kill it.
Screencast: [https://vid.me/tT3K](https://vid.me/tT3K)

Firefox version: 52.0a2 aurora
e10s: activated

bjorn3:
I guess its because debugger.html can't handle sparse arrays.

```javascript
> console.log(lists)
Object [ <1 empty slot>, "{"id":1,"title":"Test","list_collec…", "{"id":2,"title":"Salzgitter Heute",…", <7 empty slots>, 66261271 more… ]
```

jasonLaster:
Uhoh. We'll definitely look into that. 

Hmm. Could you try creating a test case?

Also is it possible for me to go to your page?

bjorn3:
Minimal reproduction:
```javascript
localStorage[1] = "a test";
localStorage[2] = "a";
localStorage[66261280] = "another test";
let data = JSON.parse(JSON.stringify(localStorage));
console.log(data) // Put breakpoint here
```
Needs two low numbers and then a high number.

jasonLaster:
Thanks.

At first I thought you used your phone number as the key.

I'll look into it later today.

bjorn3:
Its a database key.

jasonLaster:
Makes more sense

bjorn3:
Screencast debugging debugger.html while debugging the snippet above: https://vid.me/OoQd.

Note: mouse position seems to be off quite a bit.

clarkbw:
When I tried to hunt this down it looked like it broke here: https://github.com/devtools-html/debugger.html/blob/master/src/reducers/pause.js#L45 the `fromJS(pauseInfo)` call was spinning. Not sure if other people are seeing that as well.

jasonLaster:
Thanks. That looks correct.

I'll try that

jasonLaster:
@clarkbw @bjorn3 just added a test case here https://github.com/devtools-html/debugger.html/pull/1407. Is this right? I'm not seeing it on a mac...

clarkbw:
I still believe this has to do with ImmutableJS and the `fromJS` method.

[...]

PR #1407, opened 2016-12-09
Support Add sparse arrays

jasonLaster:
Associated Issue: #1375

### Summary of Changes

* adds an example
* lets see if we can add the windows fix here too

### DIFF
--- a/src/test/examples/sparse-array.html
+++ b/src/test/examples/sparse-array.html
@@ -1,0 +1,20 @@
+<html>
+  <head>
+    <title>Debugger Statements</title>
+  </head>
+
+  <body>
+    <script>
+      debugger;
+
+      localStorage[1] = "a test";
+      localStorage[2] = "a";
+      localStorage[66261280] = "another test";
+      let data = JSON.parse(JSON.stringify(localStorage));
+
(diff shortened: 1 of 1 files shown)


src/test/examples/sparse-array.html
```diff
@@ -0,0 +1,20 @@
...
+
+  <body>
+    <script>
+      debugger;
+
+      localStorage[1] = "a test";
+      localStorage[2] = "a";
+      localStorage[66261280] = "another test";
+      let data = JSON.parse(JSON.stringify(localStorage));
+
```
bjorn3: Debugger statement should be put here. (if data doesnt have a value it wont crash)

src/test/examples/sparse-array.html
```diff
@@ -0,0 +1,20 @@
...
+      console.log(data) // Put breakpoint here
```
bjorn3: This line could be removed.

bjorn3: Except if the js engine may optimize `data` away.

jasonLaster:
moving over here
https://github.com/jasonLaster/debugger-examples

Issue #1336, opened 2016-11-29
[Source Tabs] context menu copy source url + view source 


jasonLaster:
Users should be able to right click on a source tab and either copy the source url to their clipboard or open the source in a new tab. Screenshots below

#### Patch 1: adding the context menu items

```diff
diff --git a/src/components/SourceTabs.js b/src/components/SourceTabs.js
index 20a7e8e..bafd76f 100644
--- a/src/components/SourceTabs.js
+++ b/src/components/SourceTabs.js
@@ -106,6 +106,14 @@ const SourceTabs = React.createClass({
       }
     };

+    function copyToTheClipbloard(url) {
+      // need to research https://clipboardjs.com
+    }
+
+    function viewSource(url) {
+      // need to research
+    }
+
     const closeAllTabsMenuItem = {
       id: "node-menu-close-all-tabs",
       label: closeAllTabsLabel,
@@ -114,12 +122,31 @@ const SourceTabs = React.createClass({
       click: () => closeTabs(tabs)
     };

+    const copySourceUrl = {
+      id: "node-menu-close-tabs-to-right",
+      label: closeTabsToRightLabel,
+      accesskey: "R",
+      disabled: false,
+      click: () => copyToTheClipboard(tab.url)
+    };
+
+    const viewSource = {
+      id: "node-menu-close-all-tabs",
+      label: closeAllTabsLabel,
+      accesskey: "A",
+      disabled: false,
+      click: () => viewSource(tab.url)
+    };
+
     showMenu(e, buildMenu([
       { item: closeTabMenuItem },
       { item: closeOtherTabsMenuItem, hidden: () => tabs.size === 1 },
       { item: closeTabsToRightMenuItem, hidden: () =>
          tabs.some((t, i) => t === tab && (tabs.size - 1) === i) },
-      { item: closeAllTabsMenuItem }
+      { item: closeAllTabsMenuItem },
+      { type: "separator" },
+      { item: copySourceUrl },
+      { item: viewSource }
     ]));
   },
```

#### patch 2: connecting the items to panel functions

* this will require adding some modules to m-c
* it's sufficient for this PR to just add the shim modules and we can add the devtools-modules later

https://gist.github.com/jasonLaster/fdae80aad9c6391c98206eee7318b0cd

```bash
[...]

PR #1409, opened 2016-12-09
(WIP) [Source Tabs] context menu copy source url + view source

arthur801031:
Associated Issue: #1336

### Summary of Changes

* Add "Copy Link Address" function for source tabs

### DIFF
--- a/src/components/SourceTabs.js
+++ b/src/components/SourceTabs.js
@@ -16,7 +16,6 @@ const CloseButton = require("./CloseButton");
 const Svg = require("./utils/Svg");
 const Dropdown = React.createFactory(require("./Dropdown"));
 const { showMenu, buildMenu } = require("../utils/menu");
-const { getShownSource } = require("../selectors");
 
 require("./SourceTabs.css");
 require("./Dropdown.css");
@@ -47,8 +46,8 @@ function getHiddenTabs(sourceTabs, sourceTabEls) {
  * Clipboard function taken from
  * https://dxr.mozilla.org/mozilla-central/source/devtools/shared/platform/content/clipboard.js
  */
-function copyString(string) {
-  let doCopy = function (e) {
+function copyToTheClipboard(string) {
+  let doCopy = function(e) {
     e.clipboardData.setData("text/plain", string);
     e.preventDefault();
   };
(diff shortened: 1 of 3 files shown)


src/components/SourceTabs.js
```diff
@@ -42,6 +43,21 @@ function getHiddenTabs(sourceTabs, sourceTabEls) {
...
+function copyString(string) {
+  let doCopy = function(e) {
+    e.clipboardData.setData("text/plain", string);
+    e.preventDefault();
+  };
+
+  document.addEventListener("copy", doCopy);
+  document.execCommand("copy", false, null);
+  document.removeEventListener("copy", doCopy);
+}
```
jasonLaster: 👍 

thanks for the link

src/components/SourceTabs.js
```diff
@@ -124,12 +146,25 @@ const SourceTabs = React.createClass({
...
     };
 
+    const copySourceUrl = {
+      id: "node-menu-close-tabs-to-right",
+      label: "Copy Link Address",
+      accesskey: "X",
+      disabled: false,
+      click: () => {
+        showSource(tab);
+        this.copyToTheClipboard();
```
jasonLaster: hmm, little confused why we're trying to show the source...

couldn't we do `this.copyToTheClipboard(tab.url);`

screen shot 2016-12-09 at 4 43 28 pm

Since `tab` equals to a number, I use `showSource` to retrieve this tab's url.

jasonLaster: ahh, interesting...

I suppose we could get the `source` above https://github.com/arthur801031/debugger.html/blob/46b268f751aa73a55a04c3b5394ea146a26f6957/src/components/SourceTabs.js#L104.

```js
const source = tabs.find(t => t.get("id") == tab)
```

or something like that...

arthur801031:
Hi @jasonLaster 

I'm working on patch 2, and I'm getting `Ci is undefined` error in this function:
```
function getBrowser(ownerDoc) {
  return ownerDoc.defaultView
   .QueryInterface(Ci.nsIInterfaceRequestor)
   .getInterface(Ci.nsIWebNavigation)
   .QueryInterface(Ci.nsIDocShell)
   .chromeEventHandler;
}
```

I looked it up online, but I haven't been able to find the import file that I need. Which file should I include to have Ci? :)

jasonLaster:
Hmm, interesting... with Ci and patch two you need to run `npm run mochitests-watch` and `./mach run` I don't really recommend it honestly

jasonLaster:
@arthur801031 I pushed some updates for getting the source tab url. 

I think this should be good once the tests pass

jasonLaster:
Thanks @arthur801031! This is an awesome feature

PR #1410, opened 2016-12-09
(WIP) Add code coverage frontend

jasonLaster:
### Summary of Changes

* Adds the code coverage frontend
* There are lots of naming decisions
* I punted on UX
* we need to flag it off

Server [patch](https://bugzilla.mozilla.org/show_bug.cgi?id=1322663)


clarkbw:
Lets talk about this with @helenvholmes

jasonLaster:
Good call. I tried punting on UX here so that we could discuss

jasonLaster:
I updated the UX on the plane to be a bit more opinionated



It's very similar to [safari's](https://webkit.org/blog/3846/type-profiling-and-code-coverage-profiling-for-javascript/) CC.

addyosmani:
Fwiw, I've found Safari's UX for code coverage relatively low-friction to toggle and use.

Discoverability was the biggest challenge (realised they had CC a year after it was implemented) but docs/advocacy may be sufficient to address that for FF.

jasonLaster:
@addyosmani I agree. I think they made some nice decisions on the UX as well. 
* The greyed out lines are borrowed from their design
* I'd like to try a streaming API where while coverage is on it gets sent down as event packets
* They're summary view by folder and project is also very interesting, but likely out of scope here at the moment :)

Issue #1404, opened 2016-12-08
ESC does not open the split console if focus is in debugger editor


jnachtigall:
Similar to #1367, but since it's not a debugger but global devtool's shortcut (and also less important) I'm adding as a new ticket: 

The ESC key (which toggles the split console) also does not work in the new debugger. Works in all other tabs like network or inspector, only in the new debugger pressing ESC does nothing. Windows 10, Firefox 52.

@clarkbw Mind looking into this as well?

clarkbw:
Yeah, this works on the Mac so perhaps something with Windows broke.  

I'm traveling right now and won't be back at my Windows box until later next week.  So if someone else could look into it sooner that would be nice.

jasonLaster:
Hmm, we have this test which should catch this:

https://github.com/devtools-html/debugger.html/blob/master/src/test/mochitest/browser_dbg-console.js#L29-L33

I believe the toolbox handles showing the console splitbox, so perhaps @juliandescottes or @gregtatum would have some insights.

jnachtigall:
Just tested: Works in Ubuntu, too.

jasonLaster:
Thanks. Strange that we've got these windows events issues

juliandescottes:
@jnachtigall  on Windows 10 with FF52 dev-edition I have the following behavior:
* if the focus is in the debugger editor, ESC does not open the split console
* if the focus is anywhere else in the debugger, ESC opens the split console

However I have the same behavior on OSX. Is ESC doing nothing even if you have the focus outside of the editor?

jnachtigall:
@juliandescottes It's exactly as you describe. On Ubuntu and Windows. Thanks for nailing it done. 

The first time I noticed this, I was debugging at a breakpoint on Windows (at work) which obviously focuses in the debugger editor. Later on when quickly retesting with Ubuntu (at home) I did not care to set a breakpoint or look into where I set the focus.

jasonLaster:
So, here are the two handlers I know off the top of my head:

* the search bar seems to not stop bubbling
* autocomplete was *fixed* earlier :)
* the best way to test this is to add an event handler to **window** and see if you can get this to be called... 

```diff
diff --git a/src/components/EditorSearchBar.js b/src/components/EditorSearchBar.js
index 75c3a53..d1e91b0 100644
--- a/src/components/EditorSearchBar.js
+++ b/src/components/EditorSearchBar.js
[...]

PR #1415, opened 2016-12-09
override escape shortcut in codemirror to fix split console toggle

juliandescottes:
Associated Issue: #1404 

### Summary of Changes

* remove default codemirror shortcut for escape (which was calling the singleSelection command)
* add a custom key shortcut for escape that calls the singleSelection command if necessary, but doesn't if singleSelection would have no impact.

### Test Plan

I tested this in a tab only and not in a panel, just trying to check that I could catch the keypress event when using the escape key while the focus is in the codemirror editor.

- [x] add a listener on keypress on the window `window.addEventListener("keypress", function (e) {console.log("keypress", e)})`
- [x] open a script in the debugger editor
- [x] press escape
- [x] check the keypress event is logged

(I wanted to add a dedicated testcase but I can't get firefox to build for now for some reason ...)

### DIFF
--- a/src/components/Editor.js
+++ b/src/components/Editor.js
@@ -336,7 +336,10 @@ const Editor = React.createClass({
       enableCodeFolding: false,
       gutters: ["breakpoints"],
       value: " ",
-      extraKeys: {}
+      extraKeys: {
+        // Override code mirror keymap to avoid conflicts with split console.
+        Esc: false
+      }
     });
 
     // disables the default search shortcuts
@@ -369,6 +372,17 @@ const Editor = React.createClass({
     shortcuts.on("CmdOrCtrl+Shift+B", () => this.showConditionalPanel(
       getCursorLine(this.editor.codeMirror)
     ));
+    // The default Esc command is overridden in the CodeMirror keymap to allow
+    // the Esc keypress event to be catched by the toolbox and trigger the
+    // split console. Restore it here, but preventDefault if and only if there
+    // is a multiselection.
+    shortcuts.on("Esc", (key, e) => {
+      let cm = this.editor.codeMirror;
+      if (cm.listSelections().length > 1) {
+        cm.execCommand("singleSelection");
+        e.preventDefault();
+      }
+    });
 
     resizeBreakpointGutter(this.editor.codeMirror);
     debugGlobal("cm", this.editor.codeMirror);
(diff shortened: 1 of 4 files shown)


jasonLaster:
Thanks @juliandescottes. This patch looks good. I'd be happy to test in the panel on monday.

I'll also update our mochitest on monday so that we make sure that we don't regress in this area again.

It'd be easy to make sure that focusing on the editor does not break this...

I checked the test plan as I *assume* you followed it. Perhaps the checkboxes are a silly UX requirement :)

screen shot 2016-12-09 at 7 31 37 pm


```diff

diff --git a/src/test/mochitest/browser_dbg-console.js b/src/test/mochitest/browser_dbg-console.js
index c571036..96af521 100644
--- a/src/test/mochitest/browser_dbg-console.js
+++ b/src/test/mochitest/browser_dbg-console.js
@@ -13,7 +13,7 @@ function getSplitConsole(dbg) {
   }

   if (!toolbox.splitConsole) {
-    EventUtils.synthesizeKey("VK_ESCAPE", {}, win);
+    pressKey(dbg, "Escape");
   }

   return new Promise(resolve => {
@@ -29,6 +29,12 @@ add_task(function* () {
   Services.prefs.setBoolPref("devtools.toolbox.splitconsoleEnabled", true);
   const dbg = yield initDebugger("doc-script-switching.html");

+  // test without focusing on the editor
   yield getSplitConsole(dbg);
   ok(dbg.toolbox.splitConsole, "Split console is shown.");
+
+  // test after focusing on the editor
+  clickElement(dbg, "codeMirror");
+  pressKey(dbg, "Escape");
+  ok(!dbg.toolbox.splitConsole, "Split console is hidden.");
 });
diff --git a/src/test/mochitest/head.js b/src/test/mochitest/head.js
index b0964d8..fb4ff6f 100644
--- a/src/test/mochitest/head.js
+++ b/src/test/mochitest/head.js
@@ -534,6 +534,7 @@ const keyMappings = {
   "Enter": { code: "VK_RETURN" },
   "Up": { code: "VK_UP" },
   "Down": { code: "VK_DOWN" },
+  "Escape": { code: "VK_DOWN" },
   pauseKey: { code: "VK_F8" },
   resumeKey: { code: "VK_F8" },
   stepOverKey: { code: "VK_F10" },
```

juliandescottes:
Yes I did follow the plan :) 

I just misunderstood the purpose of the check boxes, thought the reviewer should be the one ticking them after verifying the test plan on their side.

PR #1416, opened 2016-12-09
Show real URL for tab title, not :formatted URL

darkwing:
This is sort of an ugly hack but I'd like to talk about why we're using `:formatted` and a new tab at all.  We could create a function that removes `:formatted` but that's kinda gross.  Let me know what a better route may be.

### DIFF
--- a/src/actions/sources.js
+++ b/src/actions/sources.js
@@ -197,7 +197,12 @@ function togglePrettyPrint(sourceId: string) {
 
     const url = getPrettySourceURL(source.url);
     const id = generatedToOriginalId(source.id, url);
-    const originalSource = { url, id, isPrettyPrinted: false };
+    const originalSource = {
+      url,
+      id,
+      isPrettyPrinted: false,
+      trueUrl: source.url
+    };
     dispatch({
       type: constants.ADD_SOURCE,
       source: originalSource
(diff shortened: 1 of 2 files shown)


jasonLaster:
Thanks for the PR.

Your patch brings up an interesting question about why we add a suffix to the URL in the first place.

I'll give a more detailed answer on monday, when we're back home, but here goes a quick response.

URLs generally *should* refer to the resource they came from... There are four edge cases:
* original sources (a.k.a. a source that we created from a sourcemap+bundle) we want to show the source as the user is acostemed to seeing it in the original editor
* console evals. these are also sources that we create on the fly
* scripts inside of HTML with a sourceURL stanza
* formatted *pretty* sources. these are very similar to original sources in that we use a source map to map them back
   * we want two tabs pretty and ugly. Tabs should be keyed by URL, so the formatted aspect is helpful
   * we hide the formatted source in the source tree and source search box
  

I'm sure i'm missing something here, but that is the gist.

At the moment, I prefer formatting the URL because that avoids another source field like `trueUrl`, which could be misinterpreted... This is somewhat of a sensitive decision because as you can imagine the debugger is all about sources and thus URLs, so lets talk about it and think it through :)

darkwing:
Instead of `${url}:formatted`, what if we instead used the `view-source://`protocol?  It's more representative (prettified source) and it would open properly in a browser?

jasonLaster:
@darkwing hmm, I hadn't considered that. Mind sharing a little more?

Issue #1396, opened 2016-12-07
Export a version number to see what debugger version comes with Firefox


jnachtigall:
@wldcordeiro I took the liberty to add this ;) 

As of https://github.com/devtools-html/debugger.html/issues/1378#issuecomment-265091011 it would be great if there was a way to see which version is deployed in Nightly or Dev edition. Something like `<some-global-debugger-object>.VERSION` (VERSION being a time snapshot or git hash, though I would prefer a human readable date/time). 

I was thinking about something like lodash does with [_.VERSION](https://lodash.com/docs/4.17.2#VERSION)

wldcordeiro:
@jnachtigall Thanks! Heh, missed it. I think it'd be nice to make `VERSION` equal to a format like `x.y.z-git-hash` this way it shows our planned version numbers that will be published to npm/Firefox and then in development you'll have a hash to check against.

jasonLaster:
Nice idea!

It's kinda funny that we don't have an obvious debugger global object to namespace it in.

Also, yea using the SHA makes sense. 
I don't think we want to semver the debugger yet...

jasonLaster:
Here's a patch for how I would do it:

```diff
diff --git a/packages/devtools-local-toolbox/webpack.config.js b/packages/devtools-local-toolbox/webpack.config.js
@@ -69,7 +69,8 @@ module.exports = (webpackConfig, envConfig) => {
         NODE_ENV: JSON.stringify(NODE_ENV),
         TARGET: JSON.stringify(TARGET)
       },
-      "DebuggerConfig": JSON.stringify(envConfig)
+      "DebuggerConfig": JSON.stringify(envConfig),
+      "VERSION": JSON.stringify("do some git foo w/ a lib or a shell command")
     })
   );
 
diff --git a/src/main.js b/src/main.js
index 44cb9ec..57804ce 100644
--- a/src/main.js
+++ b/src/main.js
@@ -32,6 +32,7 @@ if (!isFirefoxPanel()) {
 }
 
 window.appStore = store;
+window.debuggerVersion = VERSION;
 
 // Expose the bound actions so external things can do things like
 // selecting a source.
```

clarkbw:
If we start tagging releases you can use `git describe --abbrev=0` to get the latest tag description that can point to the version.  I went through some of this in #633

joakin:
Going to have a stab at this.

jasonLaster:
[...]

PR #1417, opened 2016-12-10
Add window.debuggerVersion with version and hash

joakin:
Associated Issue: #1396 

### Summary of Changes

* Adds a global variable `window.debuggerVersion` with format `x.y.z-XXXXXXX` where `x.y.z` is the root package.json version, and XXXXXXX is the git HEAD's short sha.

### Test Plan

Example test plan:

- [x] Open the devtools on the devtools, type `debuggerVersion` in the console, should see the version + hash printed.

### Screenshots/Videos

debuggerversion

### DIFF
--- a/packages/devtools-local-toolbox/webpack.config.js
+++ b/packages/devtools-local-toolbox/webpack.config.js
@@ -2,6 +2,7 @@
 require("babel-register");
 
 const path = require("path");
+const {execSync} = require("child_process");
 const webpack = require("webpack");
 const SingleModuleInstancePlugin = require("single-module-instance-webpack-plugin");
 const ExtractTextPlugin = require("extract-text-webpack-plugin");
@@ -63,6 +64,9 @@ module.exports = (webpackConfig, envConfig) => {
   }
   webpackConfig.externals.push(externalsTest);
 
+  const version = require('../../package.json').version;
+  const hash = execSync('git rev-parse --short HEAD').toString().replace('\n', '');
+
   webpackConfig.plugins = webpackConfig.plugins || [];
   webpackConfig.plugins.push(
     new webpack.DefinePlugin({
@@ -70,7 +74,8 @@ module.exports = (webpackConfig, envConfig) => {
         NODE_ENV: JSON.stringify(NODE_ENV),
         TARGET: JSON.stringify(TARGET)
       },
-      "DebuggerConfig": JSON.stringify(envConfig)
+      "DebuggerConfig": JSON.stringify(envConfig),
+      "VERSION": JSON.stringify(`${version}-${hash}`)
     })
   );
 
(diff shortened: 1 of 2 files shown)


joakin:
I have a couple of questions after tackling this:

* I've seen that `packages/devtools-local-toolbox/webpack.config.devtools.js` overrides the `DefinePlugin`, so it won't have the `VERSION` defined. It seems to be the config used to generate the version embedded in the Firefox panel, so should it be modified to include this too?
* As suggested on the issue, this exposes a global variable in the devtool *environment*, not on the page that is being debugged. Is that the correct idea? How would someone with the debugger open actually inspect the version? It seems like the only way right now is to open devtools on the debugger. Should this be exposed somewhere in the UI?

---

And a some comments:
* I've used the root package.json version. The first commit used the version of `packages/devtools-local-toolbox/` but it didn't seem right. If it is not the correct version to show I'll change it.
* I've used the short sha since it should be enough to pin point the last commit, but maybe it makes more sense to have the full sha instead.

joakin:
Ping @jasonLaster @jlongster for review.

packages/devtools-local-toolbox/webpack.config.js
```diff
@@ -63,14 +64,18 @@ module.exports = (webpackConfig, envConfig) => {
   }
   webpackConfig.externals.push(externalsTest);
 
+  const version = require('../../package.json').version;
```
jasonLaster: lets exclude this version, it's not really used.

packages/devtools-local-toolbox/webpack.config.js
```diff
@@ -2,6 +2,7 @@
 require("babel-register");
 
 const path = require("path");
+const {execSync} = require("child_process");
```
jasonLaster: spaces between the curlies

jasonLaster:
> webpack.config.devtools.js overrides the DefinePlugin

that's a good find. we would *definitely* want the version in the panel. 

> devtool environment, not on the page that is being debugged

I agree, both solutions have drawbacks.

1. environment - requires browser toolbox
2. ui - clutters the UI with something not helpful

---

At the moment, I'm reconsidering the value of this feature. The reason is this, if the user reports a bug with the debugger and provides the firefox version number, we *could* figure out what the debugger sha is...

I apologize for adding this issue as up for grabs before it was ready. That is my fault, we are on a team retreat and I didn't give this issue the kind of thought it required. I hope you snag another up for grab. Your PR and questions were spot on.

joakin:
No problem! It is always a learning experience 😄

BTW I just learned about the browser toolbox, very cool!

jasonLaster:
Yes! Absolutely. Try making the URL bar yellow or all inputs!

Issue #923, opened 2016-10-12
Investigate introducing PostCSS


jasonLaster:
It would be nice to know if we could easily introduce PostCSS into our webpack pipeline. There are several potential benefits of PostCSS plugins, but one of the most important would be polyfilling features that are only partially supported like [margin-inline-start](https://github.com/gasolin/postcss-bidirection).

This came up in @amitzur's RTL pr #903.

Including @gasolin as he's in our Taiwan office and wrote postcss-bidirection.

jasonLaster:
Hey @amitzur have you been able to make progress on this?

amitzur:
I did just now, but not successful yet.
There's an issue with postcss-bidirection not working properly. It throws the error below (also reproduced with postcss-cli).
I managed to make postcss to work with just autoprefixer as a POC. Take a look at this commit 763c4fec33885ae9bc441f604eb0a1f9cfb1bdd3 to see my changes. I intentionally targeted only Sources.css until I get it to work.
Any clue?

```
$ postcss --use postcss-bidirection public/js/components/Sources.css --local-plugins
TypeError: [tcomb] Invalid value {
  "raws": {
    "before": "\n  ",
    "between": ": "
  },
  "type": "decl",
  "source": {
    "start": {
      "line": 2,
      "column": 3
    },
    "input": {
[...]

PR #1419, opened 2016-12-10
Introduce Postcss

amitzur:
Associated Issue: #923 

### Summary of Changes

* add postcss, postcss-loader and postcss-bidirection packages to both root package.json and slo package.json of devtools-local-toolbox
* add yarn.lock file to devtools-local-toolbox (not sure how I'm the first one to need to commit this... did I do something wrong?)
* Use margin-inline-end etc. instead of manual html:not([dir="rtl"]) syntax which I used previously.

Couple of notes:
1. when shorthand is used for margin etc. the syntax can be pretty verbose since the only way to get postcss to process it is to break it to individual properties (see SourceTabs.css)
2. In ObjectInspector.css there's a rule for `html arrow.expanded svg`. The `html` is there for specificity. That's ugly. But in the previous syntax it was explicit and now it's not trivial to understand because the specificity issue is surfaced only post processing.

### Test Plan

TBH I did not test. Not sure what's testable here.


jasonLaster:
Here's a GIF




@amitzur, @gasolin what do you think about using [autoprefixer](https://github.com/postcss/autoprefixer)? This is a slightly different approach, instead of using `margin-inline-start`, we would use `margin-start`. and the output is slightly different as well. With autoprefixer, it would generate the two prefixed properties `-moz-margin-start` and `-webkit-margin-start`.

gasolin:
`yarn.lock` is a temporary file for `yarn` package manager, so I think you don't need to check in that.

@jasonLaster that might work, but counting on those non-standard syntax is not a good idea. At least we will see no further bug fix or any support from platfrom for `-moz-margin-start` syntax. Especially when they are super busy during migrating to new layout engine...

Issue #1418, opened 2016-12-10
New tab is opened at left side: Should open at right side


jnachtigall:
**STR:**
1. Open a file with Ctrl+P (so to just have at least one tab open already)
1. Open another file with Ctrl+P

**AR:**
New file opens at *leftmost* (left of already open tab)

**ER:**
Should open to the *rightmost* of already open tabs. The newtab `+` icon is also at the right side...

All tab-based applications I can think of do open new tabs to the right of the already opened tabs: Firefox, Chrome, Chrome Devtools, VSCode, Sublime, Eclipse, Intellij, Notepad++. I guess it's not deliberate behavior but a bug that the debugger does not follow this UX scheme ;)

shlomif:
I'm going to try to tackle this issue.

jasonLaster:
Thanks. 

This is a little difficult for a couple reasons. 

1. Adding to the right means knowing if we need to hide a tab

2. Sublime and other editors add a new tab to the right of the selected tab

shlomif:
OK, I was able to reproduce this issue. I'll try to fix it now.

shlomif:
Hi all!

The commit at https://github.com/shlomif/debugger.html/tree/issue1418-tabs-should-open-on-right fixes the symptoms of the problem while adding a new previously-failing-now-succeeding test and passing all existing tests.

Note that for now I have not followed all of @jasonLaster 's guidelines.

jasonLaster:
PR it so that we can comment.

For the record, following my guidelines is optional. Often not the best
strategy :)

jnachtigall:
> Sublime and other editors add a new tab to the right of the selected tab

That's a point. But I think there's a suble difference: Sublime has no newtab `+` button at the very right. I think, if the newtab button is at the very right (like in the Firefox browser or Chrome browser) then it's consequent to also open at the very right.

shlomif:
@jasonLaster : pull-req sent - https://github.com/devtools-html/debugger.html/pull/1425

jasonLaster:
@jnachtigall I think that's a good point.

Perhaps:
* *Cmd+P*  *new tab button* open tabs to the right
* selecting a source in the tree open a tab to the right of the selected tab

CC @helenvholmes, @jlongster, @clarkbw what do you think?

clarkbw:
I'd keep it simple and always open to the right like in the PR, of course making sure it is also visible.  I think that's why the current behavior (to the left) is what it is.

[...]

PR #1425, opened 2016-12-12
Fix for issue#1418 .

shlomif:
Associated Issue: #1418 

### Summary of Changes

* Made the new tab get appended to the end/right instead of prefixed to the beginning/left.
* Add a new unit test for it.

### Test Plan

I opened a new tab using ctrl+p and saw it got opened at the end.


jasonLaster:
Hmm, this could likely *open* a tab that would immediately be hidden. We would want to do something smarter here :)

PR #1429, opened 2016-12-12
add README to docs directory

clarkbw:
### Summary of Changes

* Adds a README file to the docs directory which gives us an index page thanks to [some changes at GitHub](https://github.com/blog/2289-publishing-with-github-pages-now-as-easy-as-1-2-3)

### Test Plan

[x] Checked that it worked at http://clarkbw.github.io/debugger.html/

screen shot 2016-12-12 at 5 56 50 am

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,0 +1,13 @@
+## debugger.html docs
+
+[debugger.html react / redux overview](./debugger-html-react-redux-overview.md)
+
+* [Getting Setup](./getting-setup.md)
+* [Local Development](./local-development.md)
+* [Remotely debuggable browsers](./remotely-debuggable-browsers.md)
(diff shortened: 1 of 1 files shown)


docs/README.md
```diff
@@ -0,0 +1,13 @@
+## debugger.html docs
+
+[debugger.html react / redux overview](./debugger-html-react-redux-overview.md)
```
jasonLaster: perhaps this should be in the list.

docs/README.md
```diff
@@ -0,0 +1,13 @@
+## debugger.html docs
```
jasonLaster: `###`

jasonLaster:
We can iterate here

PR #1435, opened 2016-12-13
Add Chrome Scopes

jasonLaster:
### Summary of Changes

* publishing this as it has sat on my computer for a couple weeks
* it adds an optional scopeChain field to frame
* it adds a new chrome scopes component
* I'll work on converting firefox scopes to the scope chain notion next

### Test Plan

- [x] doesn't break firefox
- [x] works for chrome

screen shot 2016-12-13 at 10 36 23 am


packages/devtools-client-adapters/src/chrome/events.js
```diff
@@ -1,12 +1,15 @@
 const { Source, Location, Frame } = require("../tcomb-types");
 
-let actions;
-let pageAgent;
+let actions = undefined;
```
jasonLaster: `let x` doesn't work :)

clarkbw: Why?  The default value should be `undefined`, right?

src/components/utils/Svg.css
```diff
@@ -50,3 +50,29 @@ html[dir="rtl"] .worker svg,
 html[dir="rtl"] .refresh svg {
     margin-left: 5px;
 }
+
```
clarkbw: 👍 for this move

clarkbw:
> it adds a new chrome scopes component

Why choose this path?  I'm guessing our scopes are fairly tied to the actors and even tcomb.

Just wondering since this is the first split like this.

jasonLaster:
This is blocked by the kill-packages PR.

jasonLaster:
Here's a gif of what it looks like for the record:

src/reducers/pause.js
```diff
@@ -169,10 +169,17 @@ function getSelectedFrame(state: OuterState) {
   return frames && frames.find(frame => frame.id == selectedFrameId);
 }
 
+// NOTE: currently only used for chrome
+function getScopes(state: OuterState) {
```
clarkbw: I think this should be `getChromeScopes` so it is more obvious that is is only a single target method.

Issue #1426, opened 2016-12-12
[UX] Show fullpath of file in tooltip when hovering Ctrl-P'ed files


jnachtigall:
STR:
I have very similar filename's in my project:
![2016-12-12 11_07_34-bundesministerium fur familie senioren frauen und jugend - gleichstellungsatla](https://cloud.githubusercontent.com/assets/3307404/21095478/d31ff068-c05b-11e6-964b-5a94be4aec95.png)

Real world project ;) Very hard to tell which of the two `production.js` files is the one I am looking for. So I hovered over the file.

AR:
Mousehover does not show any tooltip.

ER:
Mousehover should show a tooltip with the full path of the file.

In the end I guess that this is not so uncommon as many build tools (we use grunt) end up creating files like `app.js` or similar.

There's already part of the file path below the filename. But as you see in the shortcut, it is cut off in a way that it does not help me.

PS Sorry for all these not-sooo-important nitpicking bugs. But I guess it's the all these fine details that make a great tool :)

jnachtigall:
Just looked at the old debugger, which shows the full URL path on mouseover (in the source tree). Like this would be good.

jasonLaster:
@jnachtigall great find! 

* we should use CSS to truncate based on the width of the pane. You have lots of space, why are we cutting it off at 50 
* I think we should show a tooltip, it's a little extra noise but could help in this case.
* I actually got the tooltip in chrome devtools source modal. So, I hear your pain :)

```diff
diff --git a/src/components/SourceSearch.js b/src/components/SourceSearch.js
index a5d9fa6..c61a724 100644
--- a/src/components/SourceSearch.js
+++ b/src/components/SourceSearch.js
@@ -19,6 +19,8 @@ const Autocomplete = createFactory(require("./Autocomplete"));
 function searchResults(sources) {
   function getSourcePath(source) {
     const { path, href } = parseURL(source.get("url"));
+
+    // => why are we always trunctating here?
     // for URLs like "about:home" the path is null so we pass the full href
     return endTruncateStr((path || href), 50);
   }
```

![screen shot 2016-12-12 at 10 55 09 am](https://cloud.githubusercontent.com/assets/254562/21105795/8985dd0c-c059-11e6-9eee-fbbbff13c20d.png)

PR #1438, opened 2016-12-14
end truncate only subtitle and pass the full value to the HTML title

clarkbw:
Associated Issue: #1426 

### Summary of Changes

* only truncate the subtitle

For now I am passing the full path around such that we can have it in the title and I'm only truncating the subtitle, the path shown below the file match name.  I was going to use CSS but this actually seems simpler if the `endTruncateStr` function was doing a decent job before.

### Test Plan

- [x] Loaded up some fake sources with long names
- [x] Searched with `cmd+P` and selected some results

### DIFF
--- a/src/components/SourceSearch.js
+++ b/src/components/SourceSearch.js
@@ -20,7 +20,7 @@ function searchResults(sources) {
   function getSourcePath(source) {
     const { path, href } = parseURL(source.get("url"));
     // for URLs like "about:home" the path is null so we pass the full href
-    return endTruncateStr((path || href), 50);
+    return (path || href);
   }
 
   return sources.valueSeq()
@@ -28,7 +28,7 @@ function searchResults(sources) {
     .map(source => ({
       value: getSourcePath(source),
       title: getSourcePath(source).split("/").pop(),
-      subtitle: getSourcePath(source),
+      subtitle: endTruncateStr(getSourcePath(source)),
       id: source.get("id")
     }))
     .toJS();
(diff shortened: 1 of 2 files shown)


src/components/SourceSearch.js
```diff
@@ -20,15 +20,15 @@ function searchResults(sources) {
...
+    return (path || href);
   }
 
   return sources.valueSeq()
     .filter(source => !isPretty(source.toJS()) && source.get("url"))
     .map(source => ({
       value: getSourcePath(source),
       title: getSourcePath(source).split("/").pop(),
-      subtitle: getSourcePath(source),
+      subtitle: endTruncateStr(getSourcePath(source)),
```
jasonLaster: i think this needs `50`

src/components/SourceSearch.js
```diff
@@ -20,15 +20,15 @@ function searchResults(sources) {
...
     // for URLs like "about:home" the path is null so we pass the full href
-    return endTruncateStr((path || href), 50);
+    return (path || href);
   }
 
   return sources.valueSeq()
     .filter(source => !isPretty(source.toJS()) && source.get("url"))
     .map(source => ({
       value: getSourcePath(source),
       title: getSourcePath(source).split("/").pop(),
```
jasonLaster: is it weird to only show the title if the subtitle is `> 50`? 

I'm asking because the  tooltip is a little bit of noise when you pause over the item. 

Happy to keep it in all cases if that's the *proper* ux :)

jasonLaster:
we can add conditional titles and better truncating another day :)

PR #1443, opened 2016-12-15
Kill packages

jasonLaster:
### Summary of Changes

* create [devtools-core](https://github.com/jasonLaster/devtools-core)
* remove packages from the debugger. 
* clean up some inconsistencies

### Test Plan

- [x] make sure the local development workflow works
- [ ] make sure mochitests still pass


wldcordeiro:
@jasonLaster This looks pretty good overall since it's just moving things. Seems like the mochitests are timing out for some of them though. Also I think we should move this to the devtools-html org since there are a few issues associated with things now in a separate repo.

jasonLaster:
good comments @wldcordeiro. 
* I found the mochitest error.  
* and we'll move it to devtools-html soon

PR #1447, opened 2016-12-15
Remove Examples

jasonLaster:
Moved examples to their own rep

**benefits:**
* easy to add a new example
* doesn't clutter this repo
* opportunity to host the examples
* opportunity to style the landing page


clarkbw:
I just recalled that `serve-index` and `express` packages might need to be removed as well.  However I think you can do it in #1443

jasonLaster:
yup

PR #1449, opened 2016-12-15
Videos2

jasonLaster:
Fix videos link


clarkbw:
> warning Unmet peer dependency "react@^0.14.0 || ^15.0.0".

clarkbw:
The error has nothing to do with this change, but yarn is not happy.  We should figure out where that started happening.

PR #1451, opened 2016-12-16
update references to packages folders

juliandescottes:
Follow up to #1443

There were a few spots still referencing the local packages/ folder.

### Summary of Changes

* removed references to packages/ in linting commands
* updated documentation references to point to https://github.com/devtools-html/devtools-core/

### Test Plan

- [x] find all refs to "packages/", check they point to devtools-core
- [x] yarn test ok
- [x] yarn lint ok


docs/getting-setup.md
```diff
@@ -52,8 +52,8 @@ cd debugger.html
 yarn start
 ```
 
-*What does this do?*  
-This command starts a [development server](../packages/devtools-local-toolbox/README.md).
+*What does this do?*
+This command starts a [development server](https://github.com/devtools-html/devtools-core/blob/master/packages/devtools-local-toolbox/README.md).
```
jasonLaster: @juliandescottes you should know this by now!

We move quickly here, so the link is already out of date :)

should be https://github.com/devtools-html/devtools-core/blob/master/packages/devtools-launchpad/README.md

PR #1456, opened 2016-12-16
change path for windows in firefox driver

clarkbw:
Associated Issue: #1309 

### Summary of Changes

* prepended the `../node_modules/geckodriver` path for a windows machine so it can successfully find the `geckodriver.exe` 
* Note that using `../node_modules/geckodriver/bin` does not find the `.exe` version which is why I used the parent directory

### Test Plan

Example test plan:

- [x] yarn run firefox

fixes #1309

### DIFF
--- a/bin/firefox-driver.js
+++ b/bin/firefox-driver.js
@@ -19,6 +19,12 @@ const shouldStart = args.start;
 const isTests = args.tests;
 const useWebSocket = args.websocket;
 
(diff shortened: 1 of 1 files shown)


jasonLaster:
Love it. 

@viswanathamsantosh, @MoLow, @Garbee, @wldcordeiro really happy that windows users will be able to have a smoother getting started process.

PR #1457, opened 2016-12-16
Update yarn install method for circleci

clarkbw:
Associated Issue: n/a

### Summary of Changes

* use circleci recommended yarn install ( https://circleci.com/docs/install-and-use-yarn/ )
* remove lerna install

### Test Plan

- [x] Create PR, expect it installs the correct version of `yarn`


circle.yml
```diff
@@ -26,13 +28,16 @@ test:
 
 dependencies:
   pre:
+    - |
+      if [[ ! -e ~/.yarn/bin/yarn || $(yarn --version) != "${YARN_VERSION}" ]]; then
+        curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version $YARN_VERSION
+      fi
+    - npm i -g jasonlaster/lerna
```
jasonLaster: we don't need lerna anymore

circle.yml
```diff
@@ -6,6 +6,8 @@ machine:
   environment:
     DOWNLOADS_PATH: "$HOME/downloads"
     YARN_PATH: "$HOME/.yarn-cache"
+    YARN_VERSION: 0.16.1
```
jasonLaster: maybe scope creep, but we might be able to use yarn HEAD now, who knows

clarkbw: 👍  I was going to do that in a separate PR

circle.yml
clarkbw: Oh right, I can remove that here.

clarkbw:
> what problem is this solving?

NPM isn't the recommended way to install yarn, this method is more secure and faster.

> **Note**: Installation via npm is generally not recommended. npm is non-determinstic, packages are not signed, and npm does not perform any integrity checks other than a basic SHA1 hash, which is a security risk when installing system-wide apps.
> For these reasons, it is highly recommended that you install Yarn through the installation method best suited to your operating system.

via https://yarnpkg.com/en/docs/install#alternatives-tab

Also this should be faster since it checks that yarn is already installed instead of trying to NPM install it every time.

circle.yml
```diff
@@ -26,13 +28,15 @@ test:
 
 dependencies:
   pre:
+    - |
+      if [[ ! -e ~/.yarn/bin/yarn || $(yarn --version) != "${YARN_VERSION}" ]]; then
+        curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version $YARN_VERSION
+      fi
```
clarkbw: sure

clarkbw:
Ok, installing yarn works with these changes.  We're still seeing some `yarn install` error related to something else.

Issue #1460, opened 2016-12-16
Polish Debugger UI rough edges


jasonLaster:
I talked with @darkwing earlier today and we found some rough edges that can be polished

- [x] variable view highlight row should be the full width [1]
- [ ] clicking "Cmd+P to search" can open the source search input. Generally it is awkward that the search search input is in the center, but the prompt is in the left pane
- [x] the source search placeholder should be "Search Sources..." not "Search..." because it can be confused with searching for code
- [ ] the source search close button is too dark. It should be the same color as the new tab "+" button
- [x] source tab close buttons have more right padding than left padding to acomodate for the close button. [2] chrome is cleaner 
- [x] source tab with the formatted suffix should hide the formatted suffix and show the pretty print icon in the left. [3]
- we found a search bug where source search was "query [Object object]" could be fixed in github
- [ ] the dark mode text selection makes it hard to read the code. We should find another color that is much closer to the background so it doesn't make the source text look bad
- [ ] the secondary text used for a breakpoint line snippet and call frame source filename is too grey.

[1] variable view highlight row
![](http://g.recordit.co/BqrP5mZt49.gif)

[2] chrome tabs are cleaner in that they don't show the tab unless its active. this makes the right padding issue less of a problem

![](http://g.recordit.co/Iad4BWn6bt.gif)

[3] early source tab designs showed a pause indicator on the left. We could use that space for the blue pretty print icon. As a side-note, i think the pause icon is still useful as well.

![screen shot 2016-12-16 at 3 38 34 pm](https://cloud.githubusercontent.com/assets/254562/21277708/c44b0706-c3a5-11e6-8235-49368be5a2d0.png)

clarkbw:
> the source search placeholder should be "Search Sources..." not "Search..." because it can be confused with searching for code

I think this text might read a little long, we're low on space here and this is supposed to be secondary to the larger "search for files" that lives in the editor space.

clarkbw:
Perhaps shorten it to: '⌘+P search files'

jasonLaster:
@clarkbw actually I was referring to the center pane being vague, but i can see the label being vague as well.

![screen shot 2016-12-16 at 6 25 24 pm](https://cloud.githubusercontent.com/assets/254562/21281719/0dad31dc-c3bd-11e6-9a2c-9ed0ff160e9a.png)


How does that look? I'm a minimalist, so I'd be okay with the label being "⌘+P" too with a tooltip, "search files"

[...]

PR #1461, opened 2016-12-16
Fix #1388 - Remove :formatted text from tab tooltip

### LINKED ISSUE #1388: Tab Tooltip URL need to remove ":formatted"
Formatted JavaScript adds ":formatted" to the tab title, which is cool, but the generated URL that serves as a tab title attribute will lead to 404s:

formatted-url-404

darkwing:
Associated Issue: #1388

### Summary of Changes

* Removes ":formatted" from tab tooltips

### Test Plan

- Hover over tab title, view true URL in tooltip

@jasonLaster proposed a different solution in #1388 but this is a fix specific to just this issue -- we can use the new `getRawSourceURL` function as part of #1460

### DIFF
--- a/src/components/SourceTabs.js
+++ b/src/components/SourceTabs.js
@@ -8,7 +8,7 @@ const {
   getSourceTabs,
   getFileSearchState
 } = require("../selectors");
-const { getFilename } = require("../utils/source");
+const { getFilename, getRawSourceURL } = require("../utils/source");
 const { isEnabled } = require("devtools-config");
 const classnames = require("classnames");
 const actions = require("../actions");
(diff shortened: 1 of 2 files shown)


src/components/SourceTabs.js
```diff
@@ -229,7 +229,7 @@ const SourceTabs = React.createClass({
         key: source.get("id"),
         onClick: () => selectSource(source.get("id")),
         onContextMenu: (e) => this.onTabContextMenu(e, source.get("id")),
-        title: source.get("url")
+        title: getRawSourceURL(source.get("url"))
```
darkwing: I suppose we could simplify all of this by adding `.replace(/:formatted$/, "")` here, but the function will come in handy in the future.

jasonLaster: yup - we'll want to use it in other places too

jasonLaster:
@darkwing so... this is kind of funny, I can't find our other conversation where I list the URL edge cases :) do you know how to find it again because the conversation looks like its attached to an old commit

Issue #1465, opened 2016-12-16
[Greenkeeper] bump lockfile


jasonLaster:
Here's a workaround for setting up travis ci as a lockfile bumper

https://gist.github.com/maxrimue/16e272c4e0b8a8ac9dd5570e180e08c0

jasonLaster:
closing as we have this working I think

PR #1468, opened 2016-12-17
use CircleCI to update the yarn.lock file

clarkbw:
Associated Issue: #1465 

### Summary of Changes

* use deployment to run a script when we see a `greenkeeper` branch
* create a script that pushes the new `yarn.lock` file to the branch

### Test Plan

I might create a new branch / PR with the greenkeeper format

* `greenkeeper/name`
*  commit message "fix(package): update ..."
* change to the `package.json`


bin/update-yarn
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bash
+# inspired by https://gist.github.com/maxrimue/16e272c4e0b8a8ac9dd5570e180e08c0
```
jasonLaster: nit: perhaps `update-yarn-lock-file` 

this could be obvious to everyone else, but i thought you were updating yarn at first :)

circle.yml
```diff
@@ -42,3 +42,9 @@ experimental:
     branches:
       only:
         - master
+
+deployment:
+  greenkeeper:
+    branch: /greenkeeper\/.*/
```
jasonLaster: cool :)

bin/update-yarn
```diff
@@ -0,0 +1,30 @@
...
+
+# this script is only for CircleCI
+$CIRCLECI || exit 0
+
+if [[ $CIRCLE_BRANCH != *"greenkeeper"* ]]; then
+	# Not a GreenKeeper Pull Request, aborting
+	exit 0
+fi
+
+cd $CIRCLE_PROJECT_REPONAME
```
jasonLaster: such magic

bin/update-yarn
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bash
+# inspired by https://gist.github.com/maxrimue/16e272c4e0b8a8ac9dd5570e180e08c0
+
+# this script is only for CircleCI
+$CIRCLECI || exit 0
+
+if [[ $CIRCLE_BRANCH != *"greenkeeper"* ]]; then
+	# Not a GreenKeeper Pull Request, aborting
+	exit 0
```
jasonLaster: I like this. I assume, we have the same check below, but better to be redundant

bin/update-yarn
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bash
+# inspired by https://gist.github.com/maxrimue/16e272c4e0b8a8ac9dd5570e180e08c0
+
+# this script is only for CircleCI
+$CIRCLECI || exit 0
```
jasonLaster: Nice

jasonLaster:
another yarn hiccup.

cleared the cache and re-built.

I hope it's an artifact of us doing lots of weird package things.

bin/update-yarn
clarkbw: 👍

bin/update-yarn
clarkbw: I am worried this may need to be `cd $HOME/$CIRCLE_PROJECT_REPONAME` but I wanted to see a run first.

bin/update-yarn
clarkbw: This tests that we are looking at a greenkeeper like branch so we don't do this on anything else.

circle.yml
clarkbw: This also is a test for greenkeeper like branches.

bin/update-yarn
```diff
@@ -0,0 +1,30 @@
...
+
+if [[ $CIRCLE_BRANCH != *"greenkeeper"* ]]; then
+	# Not a GreenKeeper Pull Request, aborting
+	exit 0
+fi
+
+cd $CIRCLE_PROJECT_REPONAME
+
+# See if commit message includes "update"
+git log --name-status HEAD^..HEAD | grep "update" || exit 0
```
clarkbw: This tests that we are seeing an "update" commit message.  However our own commit contains the word "update" so we also test below.

bin/update-yarn
```diff
@@ -0,0 +1,30 @@
...
+
+# only continue if yarn needs to update
+yarn check || exit 0
```
clarkbw: This final check ensures that we aren't in an infinite loop trying to commit our yarn.lock file over an over again.

clarkbw:
I think the final step is to add a r/w ssh key to circle so it has write permissions to the repo

Issue #1430, opened 2016-12-12
[CommandBar] support additional stepping shortcuts


jasonLaster:
@gregtatum mentioned [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1319398#c6) that there are shortcut keys for stepping.

```
Resume Script Execution: ⌘ + \
Step Over: ⌘ + '
Step In: ⌘ + ;
Step Out: ⌘ + shift + ;
```

jasonLaster:
Done in #1469

PR #1469, opened 2016-12-17
add mac friendly keyboard shortcuts to command bar

clarkbw:
Associated Issue: #1430 

### Summary of Changes

* create a key mapping system for all OSes to have a flat object mapping to register key shortcuts correctly
* add a `formatKeyShortcut` function to utils which properly formats the keyboard shortcut commands to human friendly text
* remove use of `cmdString` in favor of `formatKeyShortcut` util

### Open Questions

I didn't update the `.properties` file because that is a bit more involved change and requires some discussion about the changes I made to the tooltip text.  See the Format section at the bottom for general changes.

I also changed the Pause / Resume tooltip to remove the "Click to" prefix text which I believe is excessive for a tooltip and actually probably bad form.

### Test Plan

- [x] Hover over command keys

**NOTE**: I haven't had a chance to check this on my Windows box, it needs to be verified there and hopefully Linux before we merge.

### Tooltip Format

Tooltips match the look of other areas, for example `⌘+P`
Removed the parenthesis surrounding tooltips keyboard shortcuts as I believe it created confusing extra symbols.
Key shortcuts use the `+` to connect actions, for example `⌘+F`
If multiple tooltips (mac only) they are separated by a space, for example `Pause ⌘+\ F8`


src/components/CommandBar.js
```diff
@@ -53,10 +97,14 @@ const CommandBar = React.createClass({
 
   componentWillUnmount() {
     const shortcuts = this.context.shortcuts;
-    shortcuts.off("F8");
-    shortcuts.off("F10");
-    shortcuts.off(`${ctrlKey}F11`);
-    shortcuts.off(`${ctrlKey}Shift+F11`);
+    ["resume", "stepOver", "stepIn", "stepOut"].forEach(
```
jasonLaster: lets have fun with this:

in the module context you can set this list

```js
const debuggerCommands = ["resume", "stepOver", "stepIn", "stepOut"];
```

and here it becomes cleaner:

```js
debuggerCommands.forEach(
  action => shortcuts.off(getKeyForOS("WINNT", action))
);
```

src/components/CommandBar.js
```diff
@@ -67,24 +115,40 @@ const CommandBar = React.createClass({
       func();
     };
 
-    shortcuts.on("F8", (_, e) => handleEvent(e, this.props.resume));
-    shortcuts.on("F10", (_, e) => handleEvent(e, this.props.stepOver));
-    shortcuts.on(`${ctrlKey}F11`, (_, e) => handleEvent(e, this.props.stepIn));
-    shortcuts.on(`${ctrlKey}Shift+F11`,
+    shortcuts.on(getKey("resume"),
```
jasonLaster: we can reuse debuggerCommands here too:

```js
debuggerCommands.forEach(command => shortcuts.on(
  getKey("stepOver"), 
  (_, e) => handleEvent(e, this.props[command])
));
```

jasonLaster:
Hmm, I suppose for the the properties/strings file it would make sense to have keys per OS, so `stepIn.mac`, `stepIn.Windows` I don't think that L10N has something fancy here.

clarkbw:
> the properties/strings file it would make sense to have keys per OS, so stepIn.mac, stepIn.Windows I don't think that L10N has something fancy here.

The L10N strings are in the format `pauseButtonTooltip=Click to pause (%S)` but if I remember the L10N rules correctly you cannot just change the string without also changing the key name.  So to change up the strings we would need to do something like `pauseButtonKeyTooltip=Pause %S`.  But I wanted to get some approval on the string changes before doing all that work to change up the property values.

clarkbw:
Ok, just tested this on my Windows box and everything looks good. 🍷

jasonLaster:
https://circleci.com/gh/jasonLaster/debugger.html/1754

Issue #1280, opened 2016-11-21
Open Panes when the debugger pauses


jasonLaster:
We would like to make sure the Call Stack and Scope pane are open when the debugger pauses.

The logic goes like this:
- if the panes are open keep them open
- if they are closed open them

#804 started this push by making sure the local scope was open.
This work is blocked by #1277, which will make it much easier to open the panes.

#### Patch

* I started working on this patch, but realized that it's hard to get `shouldOpen` to update the Accordion's `opened` state. 

```diff
diff --git a/src/components/RightSidebar.js b/src/components/RightSidebar.js
index 7b07cb6..c4e2010 100644
--- a/src/components/RightSidebar.js
+++ b/src/components/RightSidebar.js
@@ -5,6 +5,8 @@ const { bindActionCreators } = require("redux");
 const { isEnabled } = require("devtools-config");
 const Svg = require("./utils/Svg");

+const { getPaused } = require("../selectors");
+
 const actions = require("../actions");
 const Breakpoints = React.createFactory(require("./Breakpoints"));
 const Expressions = React.createFactory(require("./Expressions"));
@@ -61,15 +63,18 @@ const RightSidebar = React.createClass({

   getItems() {
     const { expressionInputVisibility } = this.state;
+    const { pauseData } = this.props;

     const items = [
       { header: L10N.getStr("breakpoints.header"),
         component: Breakpoints,
         opened: true },
       { header: L10N.getStr("callStack.header"),
-        component: Frames },
+        component: Frames,
+        shouldOpen: () => !!pauseData },
       { header: L10N.getStr("scopes.header"),
-        component: Scopes }
+        component: Scopes,
+        shouldOpen: () => !!pauseData }
     ];
     if (isEnabled("watchExpressions")) {
       items.unshift({ header: L10N.getStr("watchExpressions.header"),
@@ -97,6 +102,8 @@ const RightSidebar = React.createClass({
 });

 module.exports = connect(
-  () => ({}),
+  state => ({
+    pauseData: getPaused(state)
+  }),
   dispatch => bindActionCreators(actions, dispatch)
 )(RightSidebar);
```

CC @bomsy

jasonLaster:
Attaching a conversation between James and I on this topic:

```
hey, i got tripped up thinking about how this feature could be implemented... curious to see if you have a simple idea - https://github.com/devtools-html/debugger.html/issues/1280
12:30 PM i probably am missing something obvious
[...]

PR #1471, opened 2016-12-17
callstack and scope panes open on pause

liorpr:
Associated Issue: #1280 

Right Sidebar gets the pause data from the state, and the Accordion items now have a shouldOpen prop. Just like @jasonLaster suggested.

### DIFF
--- a/src/components/Accordion.js
+++ b/src/components/Accordion.js
@@ -19,11 +19,10 @@ const Accordion = React.createClass({
   },
 
   componentWillReceiveProps: function(nextProps) {
-    const opened = [...this.state.opened];
-    const newOpened = opened.map((isOpen, i) => {
-      const shouldOpen = this.props.items[i].shouldOpen;
+    const newOpened = this.state.opened.map((isOpen, i) => {
+      const { shouldOpen } = nextProps.items[i];
 
-      return isOpen ? isOpen : shouldOpen && shouldOpen();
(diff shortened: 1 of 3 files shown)


jasonLaster:
> when the pause ends, should they close?

sorry, i missed this earlier. I think, they can stay open :)

src/components/Accordion.js
```diff
@@ -18,6 +18,17 @@ const Accordion = React.createClass({
       created: [] };
   },
 
+  componentWillReceiveProps: function(nextProps) {
+    const opened = [...this.state.opened];
+    const newOpened = opened.map((isOpen, i) => {
+      const shouldOpen = this.props.items[i].shouldOpen;
+
+      return isOpen ? isOpen : shouldOpen && shouldOpen();
```
jasonLaster: Hmm perhaps isOpen || (shouldOpen && sjouldOpen()

src/components/Accordion.js
```diff
@@ -18,6 +18,17 @@ const Accordion = React.createClass({
       created: [] };
   },
 
+  componentWillReceiveProps: function(nextProps) {
+    const opened = [...this.state.opened];
```
jasonLaster: What's this doing :)

liorpr: it's being used in handleHeaderClick down below, seems like copying to a new array, although  unnecessary here cause I use .map (realizing  just now :)  )

PR #1474, opened 2016-12-18
Jump to mapped location

jasonLaster:
@thomasballinger and I paired on a really neat feature today, where when you right click on a source location you're prompted to "jump" to the mapped location e.g.

clicking on an original location `(math.js:3)` jumps to a generated location `(bundle.js:56)`

 

Next steps:

- we're not clearing the flash/highlight line. We tried to clear the debugLine accidentally
- the action API is `jumpToPairedLine`, but I think it should be called `selectMappedLocation`. Select is closer to the other APIs, mapped is more precise than paired.

Test Plan:
- [x] select a non-mapped source should show nothing
- [x] select an original source should show generated
- [x] select a generated source should show original
...


clarkbw:
This is really cool! 😲 

I'm noticing a `line - 1`problem where I need to choose the line below what I want.

jump-1g

clarkbw:
The context menu doesn't seem to work in Firefox, only in Chrome.

src/components/Editor.js
```diff
@@ -90,6 +92,36 @@ const Editor = React.createClass({
...
+      sourceId: this.props.selectedLocation.sourceId,
+      line: line++,
+      column: ch++
+    };
+
+    const pairedType = isOriginalId(this.props.selectedLocation.sourceId)
+      ? L10N.getStr("generated") : L10N.getStr("original");
+
+    const jumpLabel = {
+      accesskey: "C",
```
clarkbw: I believe this needs to be either a `G` or an `O` and correspond to the label string used.  Though I guess we aren't localizing these...  

It might be simpler to create both labels and then choose which one to add to the menu below.

jasonLaster:
@clarkbw hmm the `line - 1` looks like it's coming from the generated source. I wonder why that is...

jasonLaster:
> The context menu doesn't seem to work in Firefox, only in Chrome

Take a look at the code again! It is trying to tell us, watch out for firefox!

amitzur:
This **really is** a cool feature!

PR #1475, opened 2016-12-19
Fix #1487: "revealing the same source again to refocus it"

### LINKED ISSUE #1487: [Reveal Source] showing an already revealed source
There's a small edge case with reveal source

Steps to reproduce:
1. reveal source
2. select a different source
3. re-reveal source

Notice, it doesn't re-focus the previously revealed source

arthur801031:
Associated Issue: #1487

### Summary of Changes

* Revealing the same source again to refocus it

### DIFF
--- a/src/components/SourceTabs.js
+++ b/src/components/SourceTabs.js
@@ -142,7 +142,7 @@ const SourceTabs = React.createClass({
       label: "Show source",
       accesskey: "s",
       disabled: false,
-      click: () => showSource(tab, Math.random())
+      click: () => showSource(tab)
     };
 
     const copySourceUrl = {
--- a/src/components/SourceTabs.js
+++ b/src/components/SourceTabs.js
@@ -142,7 +142,7 @@ const SourceTabs = React.createClass({
       label: "Show source",
       accesskey: "s",
       disabled: false,
-      click: () => showSource(tab)
+      click: () => showSource(tab, Math.random())
     };
 
     const copySourceUrl = {
--- a/src/actions/types.js
+++ b/src/actions/types.js
@@ -95,7 +95,7 @@ type SourceAction =
 
 type UIAction =
   { type: "TOGGLE_FILE_SEARCH", searchOn: boolean }
-  | { type: "SHOW_SOURCE", sourceUrl: string, randomNum: number };
+  | { type: "SHOW_SOURCE", sourceUrl: string };
 
 type PauseAction =
   { type: "BREAK_ON_NEXT", value: boolean }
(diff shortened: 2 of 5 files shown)


arthur801031:
Hi @jasonLaster 

Please check. Thank you! :)

src/components/SourceTabs.js
```diff
@@ -142,7 +142,7 @@ const SourceTabs = React.createClass({
       label: "Show source",
       accesskey: "s",
       disabled: false,
-      click: () => showSource(tab)
+      click: () => showSource(tab, Math.random())
```
jasonLaster: hmm, this seems strange. 

Perhaps we need to change the way Source Tree checks in receiveProps, and always focus or something like that

I use this to fix "revealing the same source again to refocus it". The reason that it didn't refocus for revealing the same source was that `this.props.shownSource` did not change in `SourcesTree.js` when we revealed the same source. The `tab` that we passed into `showSource` in `SourceTabs.js` was the same. In order to change `this.props.shownSource`  in `SourcesTree.js` which would call `componentWillReceiveProps`, I used a random number so that every time a user clicks on reveal `this.props.shownSource` is updated. Based on [StackOverflow](http://stackoverflow.com/questions/28199100/probability-of-getting-the-same-value-using-math-random), the probability of having a collision is 60 million, so it should be safe to use. 

I was only able to come up with this solution, so I'm not sure if there is other way that we can do to call `componentWillReceiveProps` in `SourcesTree` when we click on reveal of the same source.

arthur801031:
Hi @jasonLaster 

I found a different solution to this problem. Instead of using a random number to ensure `this.props.shownSource` is updated every time when we reveal a source, I reset `this.props.shownSource` every time we finish revealing a source. So every time we reveal a source `this.props.shownSource` is an empty string, it will be updated every time and `componentWillReceiveProps` will be called. 

I think this approach is better than my previous attempt because it's deterministic and simpler. :)

jasonLaster:
Hmm, I have a pretty good solution here, but I know you're not going to like it :)

Lets not let you do it. i.e. lets only show the context menu for "show source" if the source is not already shown. This is a little lame, but I think it is fine and it's better than adding additional complexity to our actions or components. :)

If you want to feel better, atom doesn't let you reveal a non-selected source :) so they also cop out on occassion

jasonLaster:
@arthur801031 this will probably work :) 

https://gist.github.com/jasonLaster/b89e693a7d40fafff8b1c0dd3ad5e763

jasonLaster:
closing this as it's probably more work than it's worth.

PR #1477, opened 2016-12-19
#1460 Polish: Add pretty print icon to source tabs, remove ':formatted' from file name

darkwing:
prettyprint

### DIFF
--- a/src/components/SourceTabs.js
+++ b/src/components/SourceTabs.js
@@ -8,7 +8,7 @@ const {
   getSourceTabs,
   getFileSearchState
 } = require("../selectors");
-const { getFilename, getRawSourceURL } = require("../utils/source");
+const { getFilename, getRawSourceURL, isPretty } = require("../utils/source");
 const { isEnabled } = require("devtools-config");
 const classnames = require("classnames");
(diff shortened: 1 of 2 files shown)


jasonLaster:
Thanks!

Two thoughts on UI:
* perhaps blue `fill: var(--something-here)` could be nice colors are in `variables.css`
* the left padding could perhaps be reduced a bit.

@clarkbw what do you think?

src/components/SourceTabs.js
```diff
@@ -228,12 +229,14 @@ const SourceTabs = React.createClass({
 
     return dom.div(
       {
-        className: classnames("source-tab", { active }),
+        className: classnames("source-tab", { active }) +
```
jasonLaster: you can use the classnames helper for this too :)

```js
classnames("source-tab", { 
  active,
  pretty: isPretty({ url: source.get("url") })
})
```

clarkbw:
> @clarkbw what do you think?

Agreed on both points.  A light-ish blue would be my choice, but pick something available in the variables.

Great work! 👍

darkwing:
Updated!

PR #1480, opened 2016-12-19
#1460 Polish: Full row highlighting for sources

darkwing:
Tree rows will be completely highlighted instead of just their labels, per "[#1]" in #1460:


jasonLaster:
New GIF

PR #1481, opened 2016-12-19
#1460 Polish: Update search sources strings to specify sources, not files

darkwing:
Updated the Command+P placeholder for search, as well as the welcome message -- we should stay consistent between "sources" and "files


clarkbw:
For some reason 'Search sources' doesn't initially read well for me.  Maybe it needs to sink in a bit.  

Also, I think the smaller version above the left side bar could remain shortened. It already has the word "Sources" to the left and is just supposed to act as a reminder.

Heads up that I was making some changes to the strings here as well #1469 Though I believe I used file there.

src/strings.json
```diff
@@ -25,8 +25,8 @@
   "watchExpressions.header": "Watch Expressions",
   "watchExpressions.addButton": "Add Watch Expression",
   "watchExpressions.refreshButton": "Refresh",
-  "welcome.search": "%S to search for files",
-  "sourceSearch.search": "Search...",
```
jasonLaster: two things here:
1. we also need to change [properties](https://github.com/devtools-html/debugger.html/blob/master/assets/locales/debugger.properties)
2. I believe we need to use a different key so that L10N doesn't get confused

jasonLaster: yeah, the properties file is a gotcha :)

https://github.com/devtools-html/debugger.html/pull/1482

darkwing:
@jasonLaster Can you kick CI or spot what the issue with the check is?

jasonLaster:
Sorry, that was a long process for just some string changes :)

PR #1483, opened 2016-12-19
Install Yarn 0.18.1

jasonLaster:
This gets us back to using the tip of Yarn, which hopefully is more reliable now.


clarkbw:
I'll update the docs in a separate PR

PR #1484, opened 2016-12-19
look for the yarn-error.log file relative to the build directory

clarkbw:
We need to remove the `./` to save files relative to the build directory.  See: https://circleci.com/docs/configuration/#artifacts

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -39,7 +39,7 @@ dependencies:
 
 general:
   artifacts:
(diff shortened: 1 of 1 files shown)


jasonLaster:
strange - looks like it's still not saving the artifact...

clarkbw:
> strange - looks like it's still not saving the artifact...

I feel like something strange is going on with my circleci...

PR #1493, opened 2016-12-20
Fix "revealing and focusing the non selected source"

arthur801031:
Associated Issue:
When we reveal on a non selected source, the file is shown correctly in `Sources`, but the non selected source is not focused and shown in the editor.

den4ygazjk
(GIF provided by @jasonLaster )

### Summary of Changes

* Revealing and focusing the non selected source

### DIFF
--- a/configs/development.json
+++ b/configs/development.json
@@ -12,7 +12,7 @@
     "watchExpressions": false,
     "chromeScopes": false,
     "copySource": false,
-    "showSource": true,
+    "showSource": false,
     "editorSearch": false,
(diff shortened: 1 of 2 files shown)


jasonLaster:
@arthur801031 did you know you can use the `local` config to set feature flags: https://github.com/devtools-html/debugger.html/blob/master/docs/local-development.md#creating-a-local-config. You also have hot-reloading, which would help here too!

arthur801031:
@jasonLaster I didn't know that. That's really helpful! Thanks! :)

Issue #1119, opened 2016-11-05
[RTL] Update call stack component


jasonLaster:
@amitzur did some awesome work [here](https://github.com/devtools-html/debugger.html/pull/903) to make the debugger support right to left formatting.

We should continue making improvements to some of the other components. Lets start by focusing on the call stack, which should be relatively straight forward.

**How do i get started:**
* The best way to get started is to change the index.html `<html>` to `<html dir="html">` [RTL](https://github.com/devtools-html/debugger.html/blob/master/docs/local-development.md#rtl)
* checkout amit's [PR](https://github.com/devtools-html/debugger.html/pull/903) for how he made changes

**Here's what it currently looks like now**

![](https://cloud.githubusercontent.com/assets/394320/19226865/ef18b0d0-8eb9-11e6-82b4-8c4da702fe91.png)

**Here's a patch to get started:**

using floats is probably a bad idea, perhaps we use flex...

```diff
diff --git a/src/components/Frames.css b/src/components/Frames.css
index c18faaf..7f59a5f 100644
--- a/src/components/Frames.css
+++ b/src/components/Frames.css
@@ -27,13 +27,13 @@
 }

 .frames .location {
-  float: right;
+  float: left;
   color: var(--theme-comment);
   font-weight: lighter;
 }

 .frames .title {
-  float: left;
+  float: right;
   text-overflow: ellipsis;
   overflow: hidden;
   margin-right: 1em;
```

thejones:
Working with @noeladd we found that the breakpoint text does not format correctly when running RTL mode. 

```{} || var app = app ``` instead of ```var app = app || {}```

<img width="828" alt="screen shot 2016-11-06 at 3 40 15 pm" src="https://cloud.githubusercontent.com/assets/682850/20041498/ea1aa41a-a437-11e6-825a-937cfb9afe05.png">

PR #1497, opened 2016-12-20
callstack component - switched to flex

liorpr:
Associated Issue: #1119

### Summary of Changes

changed styling of callstack component to flex.

@jasonLaster you can open a seperate issue for rtl-styling of the breakpoints component and I'll tackle that one as well.

### DIFF
--- a/src/components/Frames.css
+++ b/src/components/Frames.css
@@ -7,8 +7,9 @@
 .frames ul li {
   cursor: pointer;
   padding: 7px 10px 7px 21px;
-  clear: both;
   overflow: hidden;
+  display: flex;
+  justify-content: space-between;
 }
 
 /* Style the focused call frame like so:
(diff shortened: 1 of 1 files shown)


jasonLaster:
Nice! Didn't realize flex would just work. 

By the way, I believe breakpoints might have something crazy. 

mind sharing screenshots with rtl and ltr and both chrome and Firefox?

jasonLaster:
Thanks @liorpr! Here's the breakpoints issue https://github.com/devtools-html/debugger.html/issues/1506

Interestingly, we're using https://github.com/gasolin/postcss-bidirection, which turns out that it worked with the floats as well.

I'll update the [docs](https://github.com/devtools-html/debugger.html/blob/master/docs/local-development.md#rtl) here to mention that.

screen shot 2016-12-20 at 5 17 44 pm

jasonLaster:
Here's a gif of how well the call stack deals with small widths!

PR #1509, opened 2016-12-20
Update rtl docs

jasonLaster:
Noticed some ways these could get better

CC @amitzur, @gasolin, @clarkbw, @liorpr


gasolin:
@jasonLaster When you say `float:left`, you might mean `float: inline-start`?

PR #1512, opened 2016-12-21
Fixes #1476: odd/incorrect tab name

### LINKED ISSUE #1476: Odd/Incorrect Tab Name
I was using ESPN as my testing site and, when opening `moatad.js` from the source tree, the tab opened as `...atSW=5B9372C1-A35B-46C7-A504-F1E86B86FEE3&zMoatOF=`.  The actual URL is: `http://js.moatads.com/espndfp832188684382/moatad.js#moatClientLevel1=29724777&moatClientLevel2=467209497&moatClientLevel3=221884737&moatClientLevel4=107681394537&moatClientSlicer1=104318097&moatClientSlicer2=104335617&zMoatPS=incontent&zMoatST=&zMoatSW=5B9372C1-A35B-46C7-A504-F1E86B86FEE3&zMoatOF=`  

Since I clicked `moatad.js` in the source tree, I would expect the tab to open as `moatad.js`.

oddtabname

arthur801031:
Associated Issue: #1476

### Summary of Changes

* Fixes incorrect tab name

### DIFF
--- a/src/components/SourceTabs.js
+++ b/src/components/SourceTabs.js
@@ -233,7 +233,7 @@ const SourceTabs = React.createClass({
 
   renderTab(source) {
     const { selectedSource, selectSource, closeTab } = this.props;
-    const filename = getURL(source.get("url")).filename;
+    const filename = getRawSourceURL(getURL(source.get("url")).filename);
     const active = source.get("id") == selectedSource.get("id");
     const isPrettyCode = isPretty({ url: source.get("url") });
 
--- a/src/components/SourceTabs.js
+++ b/src/components/SourceTabs.js
@@ -236,6 +236,7 @@ const SourceTabs = React.createClass({
     const filename = getRawSourceURL(getFilename(source.toJS()));
     const active = source.get("id") == selectedSource.get("id");
     const isPrettyCode = isPretty({ url: source.get("url") });
+    filename = filename.split("?")[0].split("#")[0];
 
     function onClickClose(ev) {
       ev.stopPropagation();
--- a/src/components/SourceTabs.js
+++ b/src/components/SourceTabs.js
@@ -234,7 +234,6 @@ const SourceTabs = React.createClass({
   renderTab(source) {
     const { selectedSource, selectSource, closeTab } = this.props;
     const filename = getURL(source.get("url")).filename;
-    const oldName = getRawSourceURL(getFilename(source.toJS()));
     const active = source.get("id") == selectedSource.get("id");
     const isPrettyCode = isPretty({ url: source.get("url") });
 
--- a/src/components/SourceTabs.js
+++ b/src/components/SourceTabs.js
@@ -17,7 +17,7 @@ const Svg = require("./utils/Svg");
 const Dropdown = React.createFactory(require("./Dropdown"));
 const { showMenu, buildMenu } = require("../utils/menu");
 const { debounce } = require("lodash");
-
+const { getURL } = require("../utils/sources-tree.js");
 require("./SourceTabs.css");
 require("./Dropdown.css");
 
@@ -237,7 +237,9 @@ const SourceTabs = React.createClass({
     const active = source.get("id") == selectedSource.get("id");
     const isPrettyCode = isPretty({ url: source.get("url") });
     filename = filename.split("?")[0].split("#")[0];
-
+    const tryURL = getURL(source.get("url"));
+    debugger;
+    
     function onClickClose(ev) {
       ev.stopPropagation();
       closeTab(source.get("id"));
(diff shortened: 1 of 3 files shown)


arthur801031:
Hi @jasonLaster 

Please check. Thank you!

arthur801031:
@jasonLaster 

Another approach is using `getURL()` from `sources-tree.js`. I've tried this on `moatad.js` from https://www.cnet.com/. `getURL(source.get("url"));` would give us 

screen shot 2016-12-21 at 10 58 46 am

So we would have to process path's object by removing the folder name that comes before the filename. Please let me know which approach you would like me to do. Thank you! :)

jasonLaster:
Hmm, could  getURL return filename?

arthur801031:
@jasonLaster 

Please check. I will work on the unit tests for `getURL` tomorrow. Thank you! :)

src/components/SourceTabs.js
```diff
@@ -233,7 +233,7 @@ const SourceTabs = React.createClass({
 
   renderTab(source) {
     const { selectedSource, selectSource, closeTab } = this.props;
-    const filename = getRawSourceURL(getFilename(source.toJS()));
+    const filename = getURL(source.get("url")).filename;
```
jasonLaster: I think we still want to call `getRawSourceURL` which removes the ":formatted" suffix from URLs we pretty print

src/utils/sources-tree.js
```diff
@@ -75,13 +75,23 @@ function createParentMap(tree: any): WeakMap<any, any> {
...
 
+  // Retrieve filename from pathname
+  let filename = "";
+  if (pathname) {
+    filename = pathname.substring(pathname.lastIndexOf("/") + 1);
+    // This file does not have a name. Default should be (index).
+    if (filename == "" || !filename.includes(".")) {
+      filename = "(index)";
+    }
+  }
```
jasonLaster: hmm, lets move this to a helper function like `getFilenameFromPath`

src/utils/tests/sources-tree.js
```diff
@@ -415,4 +416,30 @@ describe("sources-tree", () => {
...
+  const urlObject = getURL("https://a/b.js#specialSection");
+
+  expect(urlObject.filename).to.be("b.js");
+ });
+
+ it("handles url with no filename for filename", function() {
+  const urlObject = getURL("https://a/c");
+
+  expect(urlObject.filename).to.be("(index)");
+ });
```
jasonLaster: 💃

arthur801031:
@jasonLaster 

Thank you very much! :)

src/utils/sources-tree.js
```diff
@@ -73,14 +73,31 @@ function createParentMap(tree: any): WeakMap<any, any> {
...
 function getURL(sourceUrl: string): { path: string, group: string } {
   const url = sourceUrl;
-  let def = { path: "", group: "" };
+  let def = { path: "", group: "", filename: "" };
   if (!url) {
     return def;
   }
 
   const { pathname, protocol, host, path } = parse(url);
+  const filename = getFilenameFromPath(pathname);
```
clarkbw: flow is complaining about this line.  I believe `pathname` could be a string or also undefined and the function only accepts a string.  you could check `pathname` and just set filename to "" if pathname undefined.

src/utils/sources-tree.js
```diff
@@ -73,14 +73,31 @@ function createParentMap(tree: any): WeakMap<any, any> {
  * @memberof utils/sources-tree
  * @static
  */
+function getFilenameFromPath(pathname) {
+  let filename = "";
+  if (pathname) {
```
clarkbw: And here, with flow checking that `pathname` is always a string you don't really need this check unless you're intentionally looking for the empty string.

src/utils/sources-tree.js
arthur801031: Hi @clarkbw

I've addressed this issue in my commit. :)

src/utils/sources-tree.js
arthur801031: Hi @clarkbw 

From those tests, it seems pathname could be `undefined`, so I use `if (pathname)` to check if pathname is defined.

clarkbw: You could use `pathname?: string` instead to indicate that you are going to handle the null / undefined case in the code.

arthur801031: Nice! I didn't know we could do that. Thank you! :)

jasonLaster:
Thanks @clarkbw. @arthur801031 welcome to the flow club!

Flow is one of the coolest things we're doing on the project, gives us lots of confidence with some of these low level changes

clarkbw:
Great work @arthur801031!  💃 

The first rule of Flow Club, is we try to help each other out with flow syntax. 😉

Issue #524, opened 2016-08-08
[Source Tree] highlight selected tab


clarkbw:
When you have multiple source tabs open in the editor I assumed switching between them would update the source tree view on the left.  I'm not sure some opposite approach of a tightly coupled source view that jumps around is the answer, but likely the selected state should change so it isn't confusing when I want to go back to a file that appears to already be selected.

See screenshot where one tab is selected but the source view shows the other file to be the selected file.
![screen shot 2016-08-08 at 4 04 49 pm](https://cloud.githubusercontent.com/assets/2134/17499478/744b82b4-5d82-11e6-9fdd-7fb161293b0c.png)

jasonLaster:
Thanks for filling this, Jane's and I talked about this in June. 

I think you're right that is a question of updating the selected item. I'd like to double check Chrome's and sublimes UX. I don't think any of them open the tree.

wldcordeiro:
Atom and I think Sublime just put a highlight over the lowest expanded part of the tree from what I can tell.

wldcordeiro:
Updating from research, Sublime only highlights the file not the tree so collapsing the containing folder won't apply the highlight to the folder. Both however will highlight the selected tab rather than last clicked.

## Sublime

![screenshot_20160919_214229](https://cloud.githubusercontent.com/assets/580982/18656850/0baaf590-7eb2-11e6-94f3-54abedc53477.png)


## Atom

![screenshot_20160919_214212](https://cloud.githubusercontent.com/assets/580982/18656853/10b60642-7eb2-11e6-8079-d5ed29ef5dc0.png)

jasonLaster:
Go for it.

wldcordeiro:
Bumping with status update, had been distracted with some other issues that had come up. Been working on this today and should have a PR in today or tomorrow.

wldcordeiro:
A slightly more complicated change than I had anticipated as there are several levels of components interacting here. I'm relabeling to signal this and in the event I burn out on the issue I'll remove my assignment and others will know. 😄

wldcordeiro:
Jumping back on this, more free time available to focus on the task.

jasonLaster:
Hey @wldcordeiro, just saw you claimed this awhile back. Still interested in taking a stab?

wldcordeiro:
@jasonLaster sorry, had lapsed on this. Working on it now. Should get a PR in soon.

jasonLaster:
[...]

PR #1516, opened 2016-12-21
[Source Tree] highlight selected tab

arthur801031:
Associated Issue: #524

### Summary of Changes

* Call `getDirectories ` in `SourcesTree.js` to retrieve a list of items for this file
* Process this list in `ManagedTree.js` and determine which item/source to highlight


arthur801031:
@jasonLaster  

Please check. Thank you! :)

src/reducers/ui.js
```diff
@@ -13,12 +13,14 @@ import type { Record } from "../utils/makeRecord";
 
 export type UIState = {
   searchOn: boolean,
-  shownSource: string
+  shownSource: string,
+  hlightSourceURL: string
```
jasonLaster: hmm, lets be consistent and use `highlightSource`

src/components/SourcesTree.js
```diff
@@ -54,6 +55,15 @@ let SourcesTree = React.createClass({
...
     }
 
+    if (nextProps.hlightSourceURL != this.props.hlightSourceURL) {
+      const hLightItems = getDirectories(
+        nextProps.hlightSourceURL,
+        this.state.sourceTree
+      );
+
+      return this.setState({ hLightItems });
+    }
```
jasonLaster: is it possible to just highlight the source? 

I'm asking because we should know that it's already expanded

src/components/SourceTabs.js
```diff
@@ -249,7 +250,10 @@ const SourceTabs = React.createClass({
           pretty: isPretty({ url: source.get("url") })
         }),
         key: source.get("id"),
-        onClick: () => selectSource(source.get("id")),
+        onClick: () => {
+          selectSource(source.get("id"));
+          this.props.highlightSource(source.get("url"));
```
jasonLaster: do we need to call `highlightSource`? perhaps when we select the source, we can check to see if the source is expanded and if so highlight it.

One thing to note is that we don't need to expand directories to highlght the source. This is kinda an optional thing, if the source is shown, we'll highlight it.

src/components/SourcesTree.js
arthur801031: I think it's not possible to know which items are expanded in `SourcesTree.js`. Same as `Reveal` , we need to use `setState` and determine which items are expanded in `ManagedTree.js`

src/components/SourceTabs.js
arthur801031: I think this feature is very similar to `Reveal` in that we need to use redux to pass the url to `SourcesTree.js`. Then, we can get a list of sources' items in `SourcesTree.js` and pass them to `ManagedTree.js` because that is where we know which items/sources are expanded. 

In `ManagedTree.js`, we're not expanding directories. We're actually checking which directories are opened, so we know which directory or file we need to highlight. It behaves the same as Atom.

src/components/SourcesTree.js
```diff
@@ -20,7 +20,8 @@ let SourcesTree = React.createClass({
   propTypes: {
     sources: ImPropTypes.map.isRequired,
     selectSource: PropTypes.func.isRequired,
-    shownSource: PropTypes.string
+    shownSource: PropTypes.string,
+    hlightSourceURL: PropTypes.string
```
wldcordeiro: Nit: You use `hlight` in several places instead of `highlight`

src/reducers/ui.js
```diff
@@ -30,6 +32,10 @@ function update(state = State(), action: Action): Record<UIState> {
     case constants.SHOW_SOURCE: {
       return state.set("shownSource", action.sourceUrl);
     }
+
+    case constants.HIGHLIGHT_SOURCE: {
```
wldcordeiro: Is it possible that we can reuse the `SHOW_SOURCE` action with an additional property? This case is so similar to it.

arthur801031: I thought about doing that, but I'm sure which approach to take. @jasonLaster  Would you prefer making `SHOW_SOURCE` an array/object to having a separate functions for this feature?

jasonLaster: Haha - I think we all want to simplify this as much as possible.

We went down this path with shown source, but with highlighted source, I think it should be the same thing as selected source. Basically, every time you select a source you also want to highlight it if you can. Is that right?

If so, SourceTree can check in receiveProps if the selectedSource changed and do everything it needs to do to select it....

arthur801031:
@jasonLaster 

Really great suggestion! There is much less code without using redux :) Please check. Thank you!

src/components/utils/ManagedTree.js
```diff
@@ -20,6 +20,12 @@ let ManagedTree = React.createClass({
        listItems.length > 0) {
       this.expandListItems(listItems);
     }
+
+    const hLightItems = nextProps.hLightItems;
```
jasonLaster: lets use the full name here `highlightItems`

src/components/utils/ManagedTree.js
```diff
@@ -46,6 +52,27 @@ let ManagedTree = React.createClass({
...
+    let index = hLightItems.length - 1;
+    // Look at folders starting from the top-level until finds a
+    // closed folder and highlight this folder
+    while(!done && index >= 1) {
+      if (!expanded.has(this.props.getKey(hLightItems[index]))) {
+        this.focusItem(hLightItems[index]);
+        done = true;
+      }
+      index--;
+    }
```
jasonLaster: can we do something here like this:

```js
this.state.expanded.forEach((item, index) => {
  if (!expanded.has(this.props.getKey(hLightItems[index]))) {
    this.focusItem(hLightItems[index]);
    done = true
  }
```

src/components/utils/ManagedTree.js
```diff
@@ -46,6 +52,27 @@ let ManagedTree = React.createClass({
     this.setState({ expanded: expanded });
   },
 
+  highlightItem(hLightItems) {
+    const expanded = this.state.expanded;
+    let done = false;
+    let index = hLightItems.length - 1;
+    // Look at folders starting from the top-level until finds a
+    // closed folder and highlight this folder
+    while(!done && index >= 1) {
```
jasonLaster: perhaps we want to first check to see if the source is visible? if so then we highlight it, if not highlight the lowest directory.

src/components/SourcesTree.js
```diff
@@ -54,6 +55,15 @@ let SourcesTree = React.createClass({
       return this.setState({ listItems });
     }
 
+    if (nextProps.selectedSource != this.props.selectedSource) {
+      const hLightItems = getDirectories(
+        nextProps.selectedSource.get("url"),
+        this.state.sourceTree
+      );
+
+      return this.setState({ hLightItems });
```
jasonLaster: hmm, it would be nice be able to figure-out here or in managed tree which is the most specific visible node and just highlight that.

it's weird to have many highlight items

arthur801031: @jasonLaster 

I'm not really sure if there is a faster way to do this :( 
The code that we have right now is that we get an itemList of the target source's parents, including itself, by calling `getDirectories`.  For example, if we have "folder/subfolder/targetFile", `getDirectories` gets us a list that contains `["folder", "folder/subfolder", "folder/subfolder/targetFile"]`. 

In `ManagedTree`, we first check if `folder/subfolder/targetFile` is in the expanded, which means if it's visible. If it is, we will just highlight it. If not, we check if `folder` is opened. If it is, we check `folder/subfolder`.  And this process is finding the lowest directory of the target source.

src/components/utils/ManagedTree.js
```diff
@@ -46,6 +52,28 @@ let ManagedTree = React.createClass({
...
+
+    // This file is visible, so we highlight it.
+    if (expanded.has(this.props.getKey(highlightItems[0]))) {
+      this.focusItem(highlightItems[0]);
+    } else {
+      let index = highlightItems.length - 1;
+
+      // Look at folders starting from the top-level until finds a
+      // closed folder and highlights this folder
+      highlightItems.some(item => {
```
jasonLaster: hmm, i think `highlightItems .forEach((item, index) => {})` would be cleaner that way you'll have the index and won't need to do `index--`

jasonLaster: or perhaps you can use `const index = highlightItems.find(item => this.props.getKey(item))` which should return the index of the item...

arthur801031: Hi @jasonLaster 
Please check. Thank you! :)
I haven't been able to figure out why I'm failing the test. I couldn't find the error message. Could you help me out with this?

jasonLaster:
Happy to look.

jasonLaster:
Hey @arthur801031 just pushed a fix for the test failure.

arthur801031:
@jasonLaster 

Thank you very much! :)

jasonLaster:
No problem!

Thanks for the great work

PR #1529, opened 2016-12-22
enable the why paused feature

clarkbw:
Associated Issue: #1345 

### Summary of Changes

* enable the "why paused" feature by default
* use L10N for all strings

### Test Plan

- [x] Trigger breakpoint
- [x] Trigger exception
- [x] Trigger stepping after pausing
- [x] Trigger debugger statement

debugger statement

exception

stepping

breakpoint


clarkbw:
BTW, this is where Chrome has the logic for their feature: https://github.com/ChromeDevTools/devtools-frontend/blob/master/front_end/sources/DebuggerPausedMessage.js#L34-L71

Issue #1505, opened 2016-12-20
Consider Showing "Conditional" breakpoint icon


darkwing:
It might be nice to show an icon or signifier in the top right column for conditional breakpoints, if only to show that they aren't always the same.

<img width="1262" alt="conditional" src="https://cloud.githubusercontent.com/assets/46655/21367762/fb2c4c18-c6c4-11e6-8c52-0eacc2be9720.png">

clarkbw:
Not sure if I found the right long term solution but I think its subtle yet still indicates the difference.

jasonLaster:
Yup. I think it's good enough for now :)

PR #1530, opened 2016-12-22
add the conditional border color to indicate those breakpoints

clarkbw:
Associated Issue: #1505 

### Summary of Changes

* Add a slight border line to conditional breakpoints to indicate they are different than others

### Test Plan

- [x] Create a breakpoint and see that it doesn't have a border
- [x] Create a conditional breakpoint and see that it does have a border

screen shot 2016-12-21 at 10 03 06 pm


src/components/Breakpoints.css
```diff
@@ -22,6 +22,10 @@
   border-color: var(--breakpoint-active-color);
 }
 
+.breakpoints-list .breakpoint.is-conditional {
+  border-left-color: var(--theme-graphs-yellow);
+}
```
jasonLaster: so... this will sound petty, but could you move this rule above the `.breakpoint.paused` rule?

I believe if we do that then the yellow border will be replaced with the green paused border.

darkwing:
I think a very light goldish background would be good too.  Nice work though, this looks great!

jasonLaster:
@darkwing hmm that could be tough on the eyes for dark mode unless we do something different there.

darkwing:
Meh, maybe we don't do it for dark mode, it's really just nice for the default white theme.

jasonLaster:
@darkwing i think this is also one of those things where you've got to see it first.

> light goldish background

to be honest. When I hear that I think, hmm background colors generally can be awkward and this one could come off as pee colored... 

Now, to be fair... as we're talking about this, i just remembered that we landed another potential pee like background fill today :) so who knows

Here's the other [PR](https://github.com/devtools-html/debugger.html/pull/1529). And the color is :[lemonchiffon](https://github.com/clarkbw/debugger.html/blob/373b8fe59964a63925e6bc840e89ae1b672f76d4/src/components/WhyPaused.css#L3).

clarkbw:
> Here's the other PR. And the color is lemonchiffon.

That color choice should be reviewed.  I originally chose it because I thought it was funny and now its not so funny but we still don't have a theme variable color.

jasonLaster:
Merging.

https://circleci.com/gh/jasonLaster/debugger.html/1789

^ getting a green on my circle :/

PR #1532, opened 2016-12-22
Rename l10n keys

jasonLaster:
We previously bumped the key name for translation reasons, but
it turns out that if the old translation is okay then it's alright to
change the value.

This is helpful for us because when we changed the key name we actually
broke the old debugger. If we do want to change an L10N string we should
copy it and create a new one so we don't break the old debugger.

CC @clarkbw


clarkbw:
So does this mean we need to have duplicate keys?  Leave the keys for the old debugger and then add new keys for the new strings we want?  This patch reverts things and I don't understand where that puts us.

jasonLaster:
Sorry for being clear. 

In this case it's completely fine to use the same key as before because the old translations are valid. 

We only need a new key when the translations would change dramatically

clarkbw:
r+ once the resume comment is fixed 🦊

Issue #1537, opened 2016-12-23
[Vertical Layout] Show the correct collapse buttons


wldcordeiro:
In the vertical layout the collapsible "panes" are different from the horizontal. Render the proper ones for the layout.

jasonLaster:
I think we decided to use an `if` in the splitbox render to remove the content

jasonLaster:
Here's a patch for what i started

https://gist.github.com/jasonLaster/849cac961391452b5bd8ba6fafbf8271

wldcordeiro:
I'll grab this. I had done the other work and am familiar with the changes needed. 😄

Issue #1359, opened 2016-11-30
[Vertical Layout] Scope Pane


jasonLaster:
We recently landed vertical layout, which gives App.js the ability to monitor the page width and flip to a vertical mode when the page is less than 700px.

Currently the scope pane is below the Frames section in the accordion. We would like to move it to the right so that it gets more real estate.

### Patch

Here's a [patch](https://gist.github.com/jasonLaster/f60bee14b54ab9661dee11634573ef74) to get the ball rolling. 

```
curl https://gist.githubusercontent.com/jasonLaster/f60bee14b54ab9661dee11634573ef74/raw/8f7745d216febc5278efe07eba0a782cd49466b4/foo.diff | git apply
```

### Debugger
![](https://cloud.githubusercontent.com/assets/254562/20759880/510af6f4-b6ec-11e6-8bea-e26872761173.gif)

### Chrome
![](https://cloud.githubusercontent.com/assets/254562/20457854/aea16cf4-ae62-11e6-9c5f-d48607e51c60.png)

petrgazarov:
Hey, I'm new to the repo, but this seems pretty straight-forward. Mind if I give it a shot?

jasonLaster:
Hey @petrgazarov go for it!

petrgazarov:
Hm, I had a fork that had 80% of progress made on this. 😕 

No blame, but perhaps next time some communication would be helpful, about either switching to a different issue or moving faster on this one.

wldcordeiro:
@petrgazarov Apologies. I had meant to ping the appropriate people when I did this work. If you're looking for issues to work on feel free to ping me or @jasonLaster and we can direct you to some open ones. Again, apologies for the lack of communication.

Issue #1358, opened 2016-11-30
[Vertical Layout] Source Tree 


jasonLaster:
We recently landed vertical layout, which gives App.js the ability to monitor the page width and flip to a vertical mode when the page is less than 700px.

Currently the source tree is the full height of the debugger. We would like the source tree to be a sibling of the editor as it is in chrome.

### Patch

Here's a [patch](https://gist.github.com/jasonLaster/579f1b93694431d4763de60697882627) to get the ball rolling. I'm really excited about this one!

it's somewhat large so here's a script to get started
```
curl https://gist.githubusercontent.com/jasonLaster/579f1b93694431d4763de60697882627/raw/a169ec97284b9db303a215bd9adb7f91e0d3c860/foo.diff | git apply
```

### Debugger
![](https://cloud.githubusercontent.com/assets/254562/20759880/510af6f4-b6ec-11e6-8bea-e26872761173.gif)

### Chrome
![](https://cloud.githubusercontent.com/assets/254562/20457854/aea16cf4-ae62-11e6-9c5f-d48607e51c60.png)

sankha93:
I would like to take a stab at this over the weekend.

jasonLaster:
Hey @sankha93 have you had a chance to look at this?
Are you blocked by anything that I can help with?

wldcordeiro:
I'll be taking this.

PR #1538, opened 2016-12-23
Rework source tree and editor layout for vertical debugger layout

wldcordeiro:
Associated Issue: #1358 

### Summary of Changes

* Refactor the `App` component to have two distinct render functions for vertical and horizontal layouts to simplify rendering the right layouts.
* Move the editor and source tree into a sibling start pane for the vertical `Splitbox`

Note: the `endPane` of the vertical layout will be broken but will be addressed in a separate PR for issue #1359 

Note 2: the collapse buttons will also be fixed in their own PR for issue #1537 

### Test Plan

- [ ] Open the debugger
- [ ] Resize to 700 or less pixels
- [ ] Observe that the editor and source tabs render as the top section of a vertical `Splitbox`

### Screenshots/Videos

screenshot_20161222_210748

### DIFF
--- a/src/components/App.js
+++ b/src/components/App.js
@@ -38,41 +38,6 @@ const App = React.createClass({
     return { shortcuts };
   },
 
-  renderWelcomeBox() {
-    return dom.div(
-      { className: "welcomebox" },
-      L10N.getFormatStr("welcome.search", formatKeyShortcut("CmdOrCtrl+P"))
-    );
-  },
-
-  renderCenterPane() {
-    return dom.div(
-      { className: "center-pane" },
-      dom.div(
-        { className: "editor-container" },
-        SourceTabs({
-          togglePane: this.togglePane,
-          startPanelCollapsed: this.state.startPanelCollapsed,
-          endPanelCollapsed: this.state.endPanelCollapsed,
-        }),
-        Editor(),
-        !this.props.selectedSource ? this.renderWelcomeBox() : null,
-        SourceSearch()
-      )
-    );
-  },
-
-  getInitialState() {
-    const vertical = isEnabled("verticalLayout")
-      ? verticalLayoutBreakpoint.matches : true;
-
-    return {
-      vertical,
-      startPanelCollapsed: false,
-      endPanelCollapsed: false,
-    };
-  },
-
   componentDidMount() {
     if (isEnabled("verticalLayout")) {
       verticalLayoutBreakpoint.addListener(this.onLayoutChange);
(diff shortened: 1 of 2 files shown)


jasonLaster:
Great write up! 

One critique, you should have made the screenshot a GIF!

wldcordeiro:
@jasonLaster heh, I don't have a good recorder setup currently. 🙁

src/components/App.js
```diff
@@ -38,54 +38,30 @@ const App = React.createClass({
...
-        SourceSearch()
-      )
-    );
+  componentWillUnmount() {
+    verticalLayoutBreakpoint.removeListener(this.onLayoutChange);
   },
 
   getInitialState() {
-    const vertical = isEnabled("verticalLayout")
+    const horizontal = isEnabled("verticalLayout")
```
wldcordeiro: I changed this because the value returned from `verticalLayoutBreakpoint` returns `false` when under 700px width which is when we go to the vertical layout, so it made sense to call it horizontal.

src/components/App.js
```diff
@@ -118,13 +118,49 @@ const App = React.createClass({
...
-      })
-    );
+      }));
+  },
+
+  renderVerticalLayout() {
+    return dom.div(
+      { className: "debugger" },
+      SplitBox({
+        style: { width: "100vw" },
```
wldcordeiro: These props likely need tweaking, I was going to take that on in the next PR.

src/components/App.js
```diff
@@ -118,13 +118,49 @@ const App = React.createClass({
...
+          minSize: 10,
+          maxSize: "80%",
+          splitterSize: 1,
+          endPanelControl: true,
+          startPanel: this.renderEditorPane(this.props),
+          endPanel: RightSidebar(),
+          vert: this.state.horizontal
+        }),
+        endPanelCollapsed: this.state.endPanelCollapsed,
+      }));
```
jasonLaster: hmm, it seems strange that we're rendering the editor panel twice here.

Perhaps we set it up now so that it's

SplitBox (vert):
  - SplitBox (hor):
     - sources 
     - editor
  - RightSidebar

This way we can update the RightSidebar in the next PR

wldcordeiro: I had left it mostly untouched since there would be the follow-up PR, but I could modify it to be slightly more working.

jasonLaster:
Haha it's a balancing act :)

wldcordeiro:
Rebased with the editor not being rendered twice.

jasonLaster:
Looking good!

jasonLaster:
Will play with it tomorrow morning

Issue #960, opened 2016-10-20
User preferences for open/closed accordions should persist across toolbox opens/closes


helenvholmes:
Our current debugger doesn't remember certain prefs that the old Firefox debugger did. Take this use case:
1. User opens the debugger
2. User expands the Call Stack
3. User closes the debugger
4. User opens the debugger a second time
5. The Call Stack is closed, despite it having been open on step 3.

On step 5, I think we want to persist that the user wants Call Stack opened, not closed.

Here's a patch that's pretty close:

```diff
diff --git a/public/js/components/RightSidebar.js b/public/js/components/RightSidebar.js
index 45070a6..e661e67 100644
--- a/public/js/components/RightSidebar.js
+++ b/public/js/components/RightSidebar.js
@@ -9,6 +9,7 @@ const { getPause, getIsWaitingOnBreak, getBreakpointsDisabled,
 const { isEnabled } = require("devtools-config");
 const Svg = require("./utils/Svg");
 const ImPropTypes = require("react-immutable-proptypes");
+const prefs = require("../utils/prefs");

 const { Services: { appinfo }} = require("devtools-modules");
 const shiftKey = appinfo.OS === "Darwin" ? "\u21E7" : "Shift+";
@@ -204,9 +205,11 @@ const RightSidebar = React.createClass({
         component: Breakpoints,
         opened: true },
       { header: L10N.getStr("callStack.header"),
-        component: Frames },
+        component: Frames,
+        opened: prefs.scopesVisible },
       { header: L10N.getStr("scopes.header"),
-        component: Scopes }
+        component: Scopes,
+        opened: prefs.scopesVisible }
     ];
     if (isEnabled("watchExpressions")) {
       items.unshift({ header: L10N.getStr("watchExpressions.header"),
diff --git a/public/js/utils/prefs.js b/public/js/utils/prefs.js
index 93e385b..e87de0e 100644
--- a/public/js/utils/prefs.js
+++ b/public/js/utils/prefs.js
@@ -8,6 +8,8 @@ if (isDevelopment()) {

 const prefs = new PrefsHelper("devtools", {
   clientSourceMapsEnabled: ["Bool", "debugger.client-source-maps-enabled"],
+  callStackVisible: ["Bool", "debugger.call-stack-visible"],
+  scopesVisible: ["Bool", "debugger.scopes-visible"],
 });

 module.exports = { prefs };
```

bomsy:
[...]

PR #1545, opened 2016-12-26
Add prefs for the callstack and the scopes panes

bomsy:
Associated Issue: #960

### Summary of Changes

* Add new prefs for callstack and scopes
* Pass in the pref name as properties to the accordion
* When the user clicks the accordion update the pref accordingly
* Check the pref value when new props are passed to the accordion.

#### Note:
If the user does not interact with the accordion explicitly (clicking). The current behaviour is maintained else the user's pref takes precedence.

### Test Plan

- [x] Click the callstack panel to expand / collapse
- [x] Reload the debugger
- [x] Check to make sure the state is maintained

### Screenshots/Videos
accordion_prefs

### DIFF
--- a/src/components/Accordion.js
+++ b/src/components/Accordion.js
@@ -40,6 +40,10 @@ const Accordion = React.createClass({
       item.onOpened();
     }
 
+    if (item.onToggle) {
+      item.onToggle(opened[i]);
+    }
+
     this.setState({ opened, created });
   },
 
--- a/bin/run-mochitests-docker
+++ b/bin/run-mochitests-docker
@@ -14,4 +14,4 @@ docker run -it \
   -e "DISPLAY=unix$DISPLAY" \
   --ipc host \
   jlongster/mochitest-runner \
-  /bin/bash -c "export SHELL=/bin/bash; touch devtools/client/debugger/new/test/mochitest/browser.ini && ./mach mochitest --subsuite devtools devtools/client/debugger/new/test/mochitest/"
+  /bin/bash -c "export SHELL=/bin/bash; touch devtools/client/debugger/new/test/mochitest/browser.ini && ./mach build && ./mach mochitest --subsuite devtools devtools/client/debugger/new/test/mochitest/"
--- a/src/components/SecondaryPanes.js
+++ b/src/components/SecondaryPanes.js
@@ -9,6 +9,7 @@ const ImPropTypes = require("react-immutable-proptypes");
 const { getPause, getBreakpoints,
         getBreakpointsDisabled, getBreakpointsLoading
       } = require("../selectors");
+const { prefs } = require("../utils/prefs");
 
 const actions = require("../actions");
 const WhyPaused = React.createFactory(require("./WhyPaused"));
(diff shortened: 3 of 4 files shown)


src/components/Accordion.js
```diff
@@ -40,6 +41,10 @@ const Accordion = React.createClass({
       item.onOpened();
     }
 
+    if (item.pref) {
+      prefs[item.pref] = opened[i];
+    }
```
jasonLaster: hmm, I think `pref` might make more sense outside of Accordion.

So, in this case we could add this to an `onOpened` callback

src/components/Accordion.js
```diff
@@ -20,9 +21,9 @@ const Accordion = React.createClass({
 
   componentWillReceiveProps: function(nextProps) {
     const newOpened = this.state.opened.map((isOpen, i) => {
-      const { shouldOpen } = nextProps.items[i];
+      const { shouldOpen, pref } = nextProps.items[i];
 
-      return isOpen || (shouldOpen && shouldOpen());
+      return (pref && prefs[pref]) || isOpen || (shouldOpen && shouldOpen());
```
jasonLaster: this can go inside of `shouldOpen`

src/components/RightSidebar.js
```diff
@@ -87,9 +87,11 @@ const RightSidebar = React.createClass({
         opened: true },
       { header: L10N.getStr("callStack.header"),
         component: Frames,
+        pref: "callStackVisible",
         shouldOpen: isPaused },
       { header: L10N.getStr("scopes.header"),
         component: Scopes,
+        pref: "scopesVisible",
```
jasonLaster: we don't need this if we're doing the work in `onOpened` and `shouldOpen`

src/components/Accordion.js
bomsy: yeah ur right... i missed the `onOpened` callback. That's the better solution.

src/components/Accordion.js
```diff
@@ -40,6 +41,10 @@ const Accordion = React.createClass({
       item.onOpened();
```
bomsy: @jasonLaster `item.onOpened()` is only triggered when the item is opened. Should we create a new event such as `item.onHeaderClicked()` which fires every time the item is clicked, or do we go with making `item.onOpened()` fire on click and just pass the current `opened` state to it?

jasonLaster: Hmm. Perhaps we want an onToggle? So we can persist there.

bomsy: done!

jasonLaster:
@bomsy is an update coming? :P

bomsy:
Yup :)

src/components/RightSidebar.js
```diff
@@ -116,7 +125,7 @@ const RightSidebar = React.createClass({
         { className: "right-sidebar",
           style: { overflowX: "hidden" }},
         CommandBar(),
-        WhyPaused(),
+        isEnabled("whyPaused") ? WhyPaused() : null,
```
jasonLaster: I believe this this is *on*

bomsy:
CI still failing... i'm not sure what's going on @jasonLaster any ideas?

jasonLaster:
@bomsy yup! Was just looking into it :)

This is the first time we're adding a new pref to the debugger and we need to update our process to do it.

jasonLaster:
# YES!

PR #1546, opened 2016-12-26
Toggle "Disable Breakpoints" tooltip to "Enable Breakpoints". Fixes #…

### LINKED ISSUE #1531: Toggle "Disable Breakpoints" tooltip to "Enable Breakpoints"
Nitpicking:  If all breakpoints are disabled, then the user can enable all again with the Breakpoints toggle. I think not only the tooltip icon should toggle but also it's tooltip depending on whether you enable or disable it.

2016-12-22 14_30_28-firefox debugger - firefox developer edition

Otherwise it says "Disable" albeit the breakpoints are already disabled.

Expected results:
If I have disabled all breakpoints, and I then can enable them again: The tooltip should show "Enable breakpoints" instead of "Disable breakpoints".

Dalimil:
Fixes #1531

It toggles the tooltip text whenever all breakpoints become disabled, either by clicking the button or clicking each breakpoint checkbox individually.
out

### DIFF
--- a/src/components/CommandBar.js
+++ b/src/components/CommandBar.js
@@ -213,7 +213,7 @@ const CommandBar = React.createClass({
       () => toggleAllBreakpoints(!breakpointsDisabled),
       "toggleBreakpoints",
       breakpointsDisabled ? "breakpoints-disabled" : "",
-      "Disable Breakpoints"
+      breakpointsDisabled ? "Enable Breakpoints" : "Disable Breakpoints"
     );
   },
 
--- a/src/components/CommandBar.js
+++ b/src/components/CommandBar.js
@@ -204,16 +204,16 @@ const CommandBar = React.createClass({
             breakpointsDisabled, breakpointsLoading } = this.props;
 
     if (breakpoints.size == 0 || breakpointsLoading) {
-      return debugBtn(
-        null, "toggleBreakpoints", "disabled", "Disable Breakpoints"
-      );
+      return debugBtn(null, "toggleBreakpoints", "disabled",
+        L10N.getStr("breakpoints.disable"));
     }
 
     return debugBtn(
       () => toggleAllBreakpoints(!breakpointsDisabled),
       "toggleBreakpoints",
       breakpointsDisabled ? "breakpoints-disabled" : "",
-      breakpointsDisabled ? "Enable Breakpoints" : "Disable Breakpoints"
+      breakpointsDisabled ? L10N.getStr("breakpoints.enable") :
+        L10N.getStr("breakpoints.disable")
     );
   },
 
--- a/src/reducers/breakpoints.js
+++ b/src/reducers/breakpoints.js
@@ -48,6 +48,10 @@ function makeLocationId(location: Location) {
   return `${location.sourceId}:${location.line}`;
 }
 
+function allBreakpointsDisabled(state) {
+  return state.breakpoints.filter(x => !x.disabled).size <= 0;
+}
+
 function update(state = State(), action: Action) {
   switch (action.type) {
     case "ADD_BREAKPOINT": {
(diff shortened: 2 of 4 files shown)


src/components/CommandBar.js
```diff
@@ -213,7 +213,7 @@ const CommandBar = React.createClass({
       () => toggleAllBreakpoints(!breakpointsDisabled),
       "toggleBreakpoints",
       breakpointsDisabled ? "breakpoints-disabled" : "",
-      "Disable Breakpoints"
+      breakpointsDisabled ? "Enable Breakpoints" : "Disable Breakpoints"
```
wldcordeiro: Could we move this label into L10N stuff? @jasonLaster maybe a separate PR for that.

src/reducers/breakpoints.js
```diff
@@ -63,7 +63,9 @@ function update(state = State(), action: Action) {
           // empty strings to be truthy, i.e. an empty string is a valid
           // condition.
           condition: firstString(action.condition, bp.condition)
-        }));
+        })).set(
```
wldcordeiro: Weird line spacing here. I'd put `.set("breakpointsDisabled", false)` all on one line.

src/reducers/breakpoints.js
```diff
@@ -100,14 +101,22 @@ function update(state = State(), action: Action) {
       if (action.status === "done") {
         const id = makeLocationId(action.breakpoint.location);
 
+        const bpsLeftEnabledAfter = state.breakpoints.filter(
+          x => !x.disabled
+        ).size - (state.breakpoints.get(id).disabled ? 0 : 1);
```
jasonLaster: this would be simpler if it was moved to after the breakpoint was updated

src/reducers/breakpoints.js
```diff
@@ -100,14 +101,22 @@ function update(state = State(), action: Action) {
       if (action.status === "done") {
         const id = makeLocationId(action.breakpoint.location);
 
+        const bpsLeftEnabledAfter = state.breakpoints.filter(
```
jasonLaster: lets move this to a utility function below `allBreakpointsDisabled`

src/reducers/breakpoints.js
```diff
@@ -100,14 +101,22 @@ function update(state = State(), action: Action) {
...
+        const updatedState = state.set(
+          "breakpointsDisabled", bpsLeftEnabledAfter <= 0
+        );
+
         if (action.disabled) {
           const bp = state.breakpoints.get(id);
-          return state.setIn(["breakpoints", id], updateObj(bp, {
+          return updatedState.setIn(["breakpoints", id], updateObj(bp, {
             loading: false, disabled: true
           }));
```
jasonLaster: you probably can do 

```js
const updatedState = state.setIn(["breakpoints", id], updateObj(bp, {
   loading: false, disabled: true
}))

return updatedState.set( "breakpointsDisabled", allBreakpointsDisabled())
```

src/strings.json
```diff
@@ -1,6 +1,8 @@
 {
   "breakpoints.header": "Breakpoints",
   "breakpoints.none": "No Breakpoints",
+  "breakpoints.enable": "Enable Breakpoints",
+  "breakpoints.disable": "Disable Breakpoints",
```
jasonLaster: We should also update `debugger.properties` which is used in the panel

Dalimil:
OK, fixed.

src/reducers/breakpoints.js
```diff
@@ -48,6 +48,10 @@ function makeLocationId(location: Location) {
   return `${location.sourceId}:${location.line}`;
 }
 
+function allBreakpointsDisabled(state) {
+  return state.breakpoints.filter(x => !x.disabled).size <= 0;
```
jasonLaster: i wonder if immutable has the `any`, `all`, or `some` method

Issue #583, opened 2016-08-25
Persist tabs across debuggee refreshes


jasonLaster:
We should persist tabs across debuggee and debugger close/opens. This means that you always keep your debugger's editor tabs open.

Here's a preliminary [patch](https://gist.github.com/22ec423591d7d8c9ced683385f4f05fe) that i think is pretty close.

You can get setup by doing:

```
curl https://gist.githubusercontent.com/jasonLaster/22ec423591d7d8c9ced683385f4f05fe/raw/faac253bdfc0ff47ed98de8d69fb48083cf57103/foo.diff |  git apply
```

This is a somewhat advanced issue because it touches our reducers and how sources load. Should be fun if you're up for it!

jasonLaster:
ping @jbhoosreddy are you still interested in taking a stab at this?

@jlongster I forgot if we've settled on a good approach for this

jlongster:
Can't remember exactly either. The difficulty is a tab is currently tied to a sourceId, but we don't have the new source's sourceId's when the NAVIGATION action is resetting the state. Need to give tabs the ability to just be tied to a URL as well, and if it has just a URL it resolves it first when clicked.

jlongster:
There are UX issues too: what happens if that source doesn't exist anymore, etc

jasonLaster:
This issue is ready to be `up for grabs`.

@bomsy, @kareniel, @jagtalon are you interested?

jagtalon:
@jasonLaster would like to tackle it!

kareniel:
@jagtalon dang, you fast :smile:

jagtalon:
@kareniel haha I just happened to open my email at that time

jasonLaster:
🔫  🔫

jasonLaster:
@jagtalon how's it going? Let me know if you have any questions, this is a *tricky* one.

jagtalon:
Hey, Jason! Sorry for the delay I got bogged down this week by stuff.
Will work on it this week!

Thank you!


On Tue, Nov 15, 2016, at 01:53 PM, Jason Laster wrote:
> @jagtalon[1] how's it going? Let me know if you have any questions,
> this is a *tricky* one.
> — You are receiving this because you were mentioned. Reply to this
> email directly, view it on GitHub[2], or mute the thread[3].
>


Links:

  1. https://github.com/jagtalon
  2. https://github.com/devtools-html/debugger.html/issues/583#issuecomment-260731696

[...]

PR #1547, opened 2016-12-27
Persist tabs across debuggee refreshes

arthur801031:
Associated Issue: #583

### Summary of Changes

* Add persistTabs code to reducers and restore tabs in actions

### DIFF
--- a/src/actions/sources.js
+++ b/src/actions/sources.js
@@ -174,11 +174,11 @@ function jumpToMappedLocation(sourceLocation: any) {
  * @memberof actions/sources
  * @static
  */
-function closeTab(id: string) {
-  removeDocument(id);
+function closeTab(url: string) {
+  removeDocument(url);
   return {
     type: constants.CLOSE_TAB,
-    id: id,
+    url
   };
 }
 
@@ -186,11 +186,16 @@ function closeTab(id: string) {
  * @memberof actions/sources
  * @static
  */
-function closeTabs(ids: string[]) {
-  ids.forEach(removeDocument);
-  return {
-    type: constants.CLOSE_TABS,
-    ids: ids,
+function closeTabs(urls: string[]) {
+  return ({ dispatch, getState, client }: ThunkArgs) => {
+    urls.forEach(url => removeDocument(
+      getSourceByURL(getState(), url).get("id")
+    ));
+
+    dispatch({
+      type: constants.CLOSE_TABS,
+      urls
+    });
   };
 }
 
--- a/src/actions/sources.js
+++ b/src/actions/sources.js
@@ -188,9 +188,12 @@ function closeTab(url: string) {
  */
 function closeTabs(urls: string[]) {
   return ({ dispatch, getState, client }: ThunkArgs) => {
-    urls.forEach(url => removeDocument(
-      getSourceByURL(getState(), url).get("id")
-    ));
+    urls.forEach(url => {
+      const source = getSourceByURL(getState(), url);
+      if (source) {
+        removeDocument(source.get("id"));
+      }
+    });
 
     dispatch({
       type: constants.CLOSE_TABS,
--- a/src/components/SourceTabs.js
+++ b/src/components/SourceTabs.js
@@ -4,9 +4,8 @@ const ImPropTypes = require("react-immutable-proptypes");
 const { connect } = require("react-redux");
 const { bindActionCreators } = require("redux");
 const {
-  getSelectedSource,
-  getSourceTabs,
-  getFileSearchState
+  getSelectedSource, getSourceTabs,
+  getFileSearchState, getSourceByURL
 } = require("../selectors");
 const { getFilename, getRawSourceURL, isPretty } = require("../utils/source");
 const { isEnabled } = require("devtools-config");
@@ -119,7 +118,10 @@ const SourceTabs = React.createClass({
     const closeAllTabsLabel = L10N.getStr("sourceTabs.closeAllTabs");
 
     const tabs = sourceTabs.map(t => t.get("id"));
+    const otherTabs = sourceTabs.filter(t => t.get("id") !== tab);
     const sourceTab = sourceTabs.find(t => t.get("id") == tab);
+    const tabURLs = sourceTabs.map(thisTab => thisTab.get("url"));
+    const otherTabURLs = otherTabs.map(thisTab => thisTab.get("url"));
     const isPrettySource = isPretty({ url: sourceTab.get("url") });
 
     const closeTabMenuItem = {
@@ -127,7 +129,7 @@ const SourceTabs = React.createClass({
       label: closeTabLabel,
       accesskey: "C",
       disabled: false,
-      click: () => closeTab(tab)
+      click: () => closeTab(sourceTab.get("url"))
     };
 
     const closeOtherTabsMenuItem = {
@@ -135,7 +137,7 @@ const SourceTabs = React.createClass({
       label: closeOtherTabsLabel,
       accesskey: "O",
       disabled: false,
-      click: () => closeTabs(tabs.filter(t => t !== tab))
+      click: () => closeTabs(otherTabURLs)
     };
 
     const closeTabsToRightMenuItem = {
@@ -145,7 +147,7 @@ const SourceTabs = React.createClass({
       disabled: false,
       click: () => {
         const tabIndex = tabs.findIndex(t => t == tab);
-        closeTabs(tabs.filter((t, i) => i > tabIndex));
+        closeTabs(tabURLs.filter((t, i) => i > tabIndex));
       }
     };
 
@@ -154,7 +156,7 @@ const SourceTabs = React.createClass({
       label: closeAllTabsLabel,
       accesskey: "A",
       disabled: false,
-      click: () => closeTabs(tabs)
+      click: () => closeTabs(tabURLs)
     };
 
     const showSourceMenuItem = {
@@ -261,7 +263,7 @@ const SourceTabs = React.createClass({
 
     function onClickClose(ev) {
       ev.stopPropagation();
-      closeTab(source.get("id"));
+      closeTab(source.get("url"));
     }
 
     return dom.div(
@@ -343,11 +345,18 @@ const SourceTabs = React.createClass({
   }
 });
 
+function getTabs(state) {
+  return getSourceTabs(state)
+    .map(url => getSourceByURL(state, url));
+}
+
 module.exports = connect(
-  state => ({
-    selectedSource: getSelectedSource(state),
-    sourceTabs: getSourceTabs(state),
-    searchOn: getFileSearchState(state)
-  }),
+  state => {
+    return {
+      selectedSource: getSelectedSource(state),
+      sourceTabs: getTabs(state),
+      searchOn: getFileSearchState(state)
+    };
+  },
   dispatch => bindActionCreators(actions, dispatch)
 )(SourceTabs);
--- a/src/reducers/sources.js
+++ b/src/reducers/sources.js
@@ -8,10 +8,10 @@
  * @module reducers/sources
  */
 
-const fromJS = require("../utils/fromJS");
 const I = require("immutable");
 const makeRecord = require("../utils/makeRecord");
 const { getPrettySourceURL } = require("../utils/source");
+const { prefs } = require("../utils/prefs");
 
 import type { Source, Location } from "../types";
 import type { Action } from "../actions/types";
@@ -37,13 +37,14 @@ export type SourcesState = {
 const State = makeRecord(({
   sources: I.Map(),
   selectedLocation: undefined,
-  pendingSelectedLocation: undefined,
+  pendingSelectedLocation: prefs.pendingSelectedLocation,
   sourcesText: I.Map(),
-  tabs: I.List([])
+  tabs: I.List(restoreTabs())
 } : SourcesState));
 
 function update(state = State(), action: Action) : Record<SourcesState> {
   let availableTabs = null;
+  let location = null;
 
   switch (action.type) {
     case "ADD_SOURCE": {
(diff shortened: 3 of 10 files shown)


arthur801031:
Hi @jasonLaster,

I added the persistTabs (sourcesURL) code to the reducers and connect them to SourceTabs.js. For the persistent storage of those URLs, are we using [localStorage](http://www.w3schools.com/html/html5_webstorage.asp) ?

jasonLaster:
Hmm, there's a patch in the issue that's a good start. 

Couple notes 
- we don't need a new persiatTabs action. We just need to persist tabs in the selectSource and closeTab reducer event 
- we need to save source URLs and not IDs. The patch switches from ids to urls. We do this be urls are the same across refreshes. 
- local storage is easy. Just localStorage.set("tabs", tabs)

arthur801031:
Hi @jasonLaster , I've got a question.

It seems Chrome's debugger keeps each debuggee's persistTabs. If we were to implement this, we would have 
```
tabs = [object1, object2,...] 

where object1 is ["cnn.com", list of sources urls] and 
object2 is ["yahoo.com", list of sources urls]

localStorage.set("persistTabs", tabs)
```

Is this ok?

jasonLaster:
Yea, that looks :)

Perhaps just tabs

arthur801031:
@jasonLaster 

Please check. Thank you! :)

arthur801031:
@jasonLaster 

For small files, the restore is fine, but when the tabs that we're restoring are large files, we restore too early. I'm not sure where to execute the restore, so that we wait long enough for the large files to load properly. Right now I'm executing the restore when `this.props.sources == prevProps.sources` inside `SourcesTree's componentDidUpdate()`.

src/components/SourcesTree.js
```diff
@@ -97,6 +100,18 @@ let SourcesTree = React.createClass({
...
+    // Restore tabs from localStorage
+    if (typeof localStorage !== "undefined" &&
+    this.props.sources == prevProps.sources &&
+    !this.props.hasRestoredTabs) {
+      const tabs = JSON.parse(localStorage.getItem("tabs"));
+      if (tabs) {
+        tabs.forEach(url => this.props.selectSourceByURL(url));
+      }
+    }
+  },
```
jasonLaster: couple things:

1. we should do the JSON.parse in the reducer
2. we *dont* need to select all of the sources. in fact this is very slow. We should be able to just show the tabs if it's in the appropriate state

src/actions/sources.js
```diff
@@ -96,6 +96,21 @@ type SelectSourceOptions = {
 };
 
 /**
+ * This is used for restoring the tabs from localStorage
+ *
+ * @memberof actions/sources
+ * @static
+ */
+function selectSourceByURL(url: string, options: SelectSourceOptions = {}) {
```
jasonLaster: hows this different from `selectSourceURL`?

src/reducers/sources.js
```diff
@@ -136,12 +139,40 @@ function _updateText(state, action : any) : Record<SourcesState> {
...
+function addToLocalStorage(url) {
+  if (typeof localStorage !== "undefined") {
+    let tabs = JSON.parse(localStorage.getItem("tabs"));
+    if (!tabs) {
+      tabs = [url];
+    } else if (!tabs.includes(url)) {
+      tabs.push(url);
+    }
+    localStorage.setItem("tabs", JSON.stringify(tabs));
+  }
```
jasonLaster: can we simplify this somehow with just a set and get to localStorage?

src/reducers/sources.js
```diff
@@ -69,14 +72,14 @@ function update(state = State(), action: Action) : Record<SourcesState> {
       });
 
     case "CLOSE_TAB":
-      availableTabs = removeSourceFromTabList(state.tabs, action.id);
+      availableTabs = removeSourceFromTabList(state, state.tabs, action.id);
```
jasonLaster: the biggest thing I see missing here is that I would add a new case for when the debugger starts. I'm not sure what the action is, but if you put a breakpoint in the reducer you should be able to observe one of the early actions and hook in there.

```js
case "???":
  // load the tabs
```

src/reducers/sources.js
```diff
@@ -58,6 +60,7 @@ function update(state = State(), action: Action) : Record<SourcesState> {
           line: action.line
         })
         .set("pendingSelectedLocation", null)
+        .set("hasRestoredTabs", true)
```
jasonLaster: I bet we can kill this

src/reducers/sources.js
arthur801031: @jasonLaster 

I think for restoring a tab the variables that need to be updated actually reside in `case "SELECT_SOURCE":`, so we don't need to add anything to this case. However, just executing `case "SELECT_SOURCE":` is not  enough because the source text would not be loaded, so we have to use `dispatch(selectSource(source.id));` in `src/actions/sources.js`. `selectSource(source.id)` would execute `case "SELECT_SOURCE` and load the source text.

jasonLaster: Hmm, I'm worried about selecting source because that also loads them and we actually only want to load the active tab

src/reducers/sources.js
arthur801031: I'm not sure, but since `tabs ` is an array of urls, I think we need to initialize the array if `tabs` is not defined and push the url to the array when there's already element(s) inside the array. The array manipulation also applies to removing a url from the array.

src/reducers/sources.js
arthur801031: I think there is a way to lessen the workload. Instead of of loading the source text for every tabs, we only load the source text for the left most tab, so the other tabs would appear but their source text would not be loaded. 

If we do this, we would only call `dispatch(loadSourceText(source));` for the leftmost tab. Instead of calling `selectSource`, I would write a method to accommodate this. 

Is this ok?

arthur801031:
Hi @jasonLaster 

I'm trying this way to only load the active tab's source text. In `SourcesTree.js`, I added the following code:
```
componentDidUpdate(prevProps, prevState) {
    // Load the selectedSource's source text when all the sources are
    // loaded.
    if (this.props.selectedSource &&
      prevProps.sources.size == this.props.sources.size) {
      this.props.selectSource(this.props.selectedSource.get("id"));
    }
}
```
What this does is that we only load the active tab's source text when all the sources in the tree are finished loading. If we restore 3 tabs, only the active tab's source text will be loaded. However, there is an  **issue**. This code would be executed if you select a tab or open a new file.

jasonLaster:
Hey @arthur801031, I hope you dont mind, but I threw a lot of code onto this pr :)

Honestly, a lot of is probably meh, but there is something that kinda works and there *are* a lot of things that I like

* tabs are saved as just URLs. The advantage is that they're a lot easier to sync w/ storage
* tabs are loaded when the debugger starts as initialState
* tabs are saved to local storage all the time

some things that are meh
* the selectedSource is not loaded right away. there's lots of code that has to protect against that, but in reality we should always have the selected url, which we can use in the source tabs UI
* elements are not always shown, but we should be able to show them for the same reason (just use the URL)

anyways, feel free to drop a lot of my code if you want to go another direction, but thought i'd share and I got carried away. This is a fun PR as I said earlier.

arthur801031:
Hi @jasonLaster

1.  The tabs do not restore immediately when the window is refresh or close/open. The tabs are restored when you select or open a file. Is this the intended behavior?
2. Some unit tests are failing. I think it requires changing some of the unit tests' code. Should I change them?

jasonLaster:
Thanks for the bug fixes @arthur801031 

1. the tabs *should* show up on refresh. it's a bug :)
2. yup we should fix the code so the unit tests pass

src/reducers/sources.js
```diff
@@ -22,7 +21,8 @@ export type SourcesState = {
   selectedLocation?: {
     sourceId: string,
     line?: number,
-    column?: number
+    column?: number,
+    url: string
```
jasonLaster: hmm, lets avoid adding `url` here. I think we should be able to get the source where its being used

src/reducers/sources.js
```diff
@@ -39,12 +39,11 @@ const State = makeRecord(({
   selectedLocation: undefined,
```
jasonLaster: perhaps we also want to store the `selectedLocation` or `pendingSelectedLocation ` in localStorage as well.

src/reducers/sources.js
```diff
@@ -136,35 +136,55 @@ function _updateText(state, action : any) : Record<SourcesState> {
...
 }
 
-function removeSourceFromTabList(tabs, id) {
-  return tabs.filter(tab => tab.get("id") != id);
+function readTabs() {
+  if (typeof localStorage === "undefined") {
+    return;
+  }
+
+  return JSON.parse(localStorage.getItem("tabs"));
```
jasonLaster: Hey @arthur801031, I just realized we can replace this localStorage stuff with **PREFS**

Here's a link for how to setup a new pref
https://github.com/devtools-html/debugger.html/blob/master/docs/local-development.md#prefs

src/reducers/sources.js
arthur801031: Hi @jasonLaster ,

I'm using selectedLocation's url in `getNewSelectedSourceId()`. I think saving the selectedLocation's url may be the easiest way to obtain the selected file's url.

wldcordeiro:
@arthur801031 I just pulled down your branch and was playing with it to test. I have observed the second bug but the first doesn't seem to happen, just clears the editor for whatever would be shown.

The screenshot below is what I'm seeing. Investigating further 😄 

screenshot_20170105_001540

arthur801031:
Hi @wldcordeiro 

Here is the first bug :( 

gb8xjshkip

src/reducers/sources.js
```diff
@@ -68,18 +69,29 @@ function update(state = State(), action: Action) : Record<SourcesState> {
         line: action.line
       });
 
+    case "RESTORE_TABS":
```
jasonLaster: what does RESTORE does?

src/components/SourceTabs.js
```diff
@@ -33,13 +32,19 @@ require("./Dropdown.css");
...
+    const topOffsets = sourceTabEls.map(t => t &&
+      t.getBoundingClientRect().top);
     return Math.min(...topOffsets);
   }
 
   const tabTopOffset = getTopOffset();
   return sourceTabs.filter((tab, index) => {
-    return sourceTabEls[index].getBoundingClientRect().top > tabTopOffset;
+    const el = sourceTabEls[index];
+    if (!el) {
```
jasonLaster: if we're persisting the selected source, we can likely drop this

src/actions/sources.js
```diff
@@ -50,6 +50,11 @@ function newSource(source: Source) {
       source
     });
 
+    dispatch({
+      type: constants.RESTORE_TABS,
+      source
+    });
```
jasonLaster: if we're adding loading the tabs in the initialState, do we need this?

jasonLaster:
By the way, un-related, you might like to squash your commits:

https://robots.thoughtbot.com/git-interactive-rebase-squash-amend-rewriting-history

I notice that you have a lot of commits :)

src/components/SourceTabs.js

You meant by reverting to having only the code on line 42?

jasonLaster: Yea.

src/actions/sources.js
arthur801031: Hi @jasonLaster 

These code enable the persist tabs to show up on refresh and close/open. I haven't been able to find a better way to achieve this behavior.

[Line 53 ~ 56 in `actions/sources.js`](https://github.com/devtools-html/debugger.html/pull/1547/files?diff=unified#diff-695f8e845eb7e7930a1b665650866b59R53)
This calls the code on line 72 - 81 in `reducers/sources.js`. We're using this because when a source has been downloaded and added to the SourcesTree, we need to check whether or not this source is in the persist tabs, which happens on line 72 - 81 in `reducers/sources.js`.

[Line 72 - 81 in `reducers/sources.js`](https://github.com/devtools-html/debugger.html/pull/1547/files?diff=unified#diff-701eb9ae1383078fc8b17995acdd81bbR72)
This code adds those persist tabs in the SourceTabs. Without this code, the persist tabs would not be loaded even though we're loading the tabs in the initial state. The reason is that loading the tabs in the initial state happens before those tabs could be downloaded and added to the SourcesTree, so we need to call this code when each source has been downloaded and added to the SourceTree, which happens on line 53 ~ 56 in `actions/sources.js`  

[Line 100 ~ 108 in `SourcesTree.js`](https://github.com/devtools-html/debugger.html/pull/1547/files?diff=unified#diff-9d180fae1eb61e88c4a4cf4b60ea0eafR100)
What this does is that we only load the active tab's source text when all the sources in the tree are finished loading at this given time. However, there is an issue. This code would be executed if you select a tab or open a new file.

src/components/SourceTabs.js
arthur801031: I just tried reverting to line 42; however, `sourceTabEls[index]` was undefined, and I got this error message: `Uncaught TypeError: Cannot read property 'getBoundingClientRect' of undefined`. I think we may have to keep this code because `sourceTabEls[index]` could be undefined.

arthur801031:
Hi @jasonLaster 

I moved the code in previous `RESTORE_TABS` to `ADD_SOURCE`, [Line 52 - 60 in `reducers/sources.js`](https://github.com/devtools-html/debugger.html/pull/1547/files#diff-701eb9ae1383078fc8b17995acdd81bbR52), to simplify the code. While I was working on this, I discovered that while we restored tabs in the initial state, we only had to set `selectedLocation` once to make those tabs appear. So, on line 52 - 60 in `reducers/sources.js`, we set the selectedLocation if it's undefined and it's one of the persist tabs that we need to restore, which means this would only be executed once :)

src/components/SourcesTree.js
```diff
@@ -97,6 +97,15 @@ let SourcesTree = React.createClass({
...
   },
 
+  componentDidUpdate(prevProps, prevState) {
+    // Load the selectedSource's source text when all the sources are
+    // loaded.
+    if (this.props.selectedSource &&
+    prevProps.sources.size == this.props.sources.size) {
+      this.props.selectSource(this.props.selectedSource.get("id"));
+    }
+  },
```
jasonLaster: 🔫

src/reducers/sources.js
```diff
@@ -176,24 +223,23 @@ function updateTabList(state, source, tabIndex) {
...
-function getNewSelectedSourceId(state: SourcesState, availableTabs) : string {
+function getNewSelectedSourceURL(state: SourcesState, availableTabs) : string {
   if (!state.selectedLocation) {
     return "";
   }
 
-  const selectedTabId = state.selectedLocation.sourceId;
-  const availableTabIds = availableTabs.map(s => s.get("id")).toJS();
-  const tabIds = state.tabs.map(s => s.get("id")).toJS();
+  const selectedTabUrl = state.selectedLocation.url;
```
jasonLaster: lets get the selectedSource here and get the URL from that.

src/reducers/sources.js
```diff
@@ -22,7 +22,8 @@ export type SourcesState = {
   selectedLocation?: {
     sourceId: string,
     line?: number,
-    column?: number
+    column?: number,
+    url: string
```
jasonLaster: lets kill url

src/reducers/sources.js
```diff
@@ -37,13 +37,14 @@ export type SourcesState = {
 const State = makeRecord(({
   sources: I.Map(),
   selectedLocation: undefined,
-  pendingSelectedLocation: undefined,
+  pendingSelectedLocation: prefs.pendingSelectedLocation,
```
arthur801031: Hi @jasonLaster 

Here is a bug that I talked about with pendingSelectedLocation:

lhexal3nx5

I think we may have to change the type of pendingSelectedLocation because pendingSelectedLocation only contains the most recent selected tab's url.

jasonLaster:
@arthur801031 alright, i just got the unit tests and flow tests to pass *i think*.

Here's what I propose, re persisting tabs and navigating, lets land this work. And then address it in a subsequent issue. 

**Why if it's broken?** Well, this is really great work. It's represents weeks of really valuable progress and I want to land it! 

**How will we solve the navigating problem?** I think we might need to relax the notion that we always have a *selected* source. If we have some sources and some tabs, we might want to re-select the first tab :)  We can do this pretty easily.

bin/mocha-server.js
```diff
@@ -37,31 +36,21 @@ function getTestPaths(dir) {
 
 const testPaths = getTestPaths(path.join(__dirname, "../src"));
 
-projectConfig.entry.bundle = projectConfig.entry.bundle.concat(testPaths);
```
jasonLaster: I'll pull this work into a new PR. It fixes the mocha web runner which is great though

PR #1548, opened 2016-12-27
update flexs for source header and tabs (fixes #1489)

### LINKED ISSUE #1489: [Editor] New Tab button positioning
We recently added the collapse buttons in the editor header, which threw off the new tab button positioning.

There are two issues that i see:

1. left positioning when a tab is there
2. right positioning when it gets narrow

Here's some [code](https://gist.github.com/jasonLaster/54ee561754869579914d2e94a89a6588) to help you get started.

binarybaba:
Associated Issue: #1489 

### Summary of Changes

* toggle-button-start flexed 1
* toggle-button-end flexed 1
* source-header displayed flex and row wrapped
* new-tab btn now has 20px margin start

### DIFF
--- a/src/components/SourceTabs.css
+++ b/src/components/SourceTabs.css
@@ -1,7 +1,12 @@
 .source-header {
   border-bottom: 1px solid var(--theme-splitter-color);
   height: 30px;
+  display: flex;
   flex: 1;
+
+  flex-flow: row wrap;
+  -webkit-align-items: stretch;
+  align-items: stretch;
 }
 
 .source-header * {
@@ -23,9 +28,10 @@
   position: relative;
   top: 4px;
   margin: 4px;
-  margin-inline-start: 30px;
+  margin-inline-start: 20px;
   line-height: 0;
   fill: var(--theme-comment);
+
 }
 
 .source-tab {
@@ -134,6 +140,7 @@
 .toggle-button-start {
   top: 7px;
   left: 0;
+
 }
 
 .toggle-button-end {
@@ -144,4 +151,5 @@
 .toggle-button-start.collapsed,
 .toggle-button-end.collapsed {
   transform: rotate(180deg);
+  flex: 1;
 }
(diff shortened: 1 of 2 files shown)


src/components/SourceTabs.css
```diff
@@ -134,6 +140,7 @@
 .toggle-button-start {
   top: 7px;
   left: 0;
+
```
jasonLaster: Hmm, mind dropping the extra new lines

binarybaba: Oh crap sorry i'll do that as soon as I reach home. Thanks!

jasonLaster:
Thanks @binarybaba! Just played with it and found a couple things

* the new tab button is slightly higher than the tabs
* the new tab button could use some hover styles (`cursor: pointer`, perhaps slightly darker)
* when the header is narrowed, the more tabs button overlays the new tab button

sorry if this is some new requirements that were unmentioned earlier

binarybaba:
hokay I just came home a while ago and man I did not see that coming. I thought github would give me a space to write the summary but it didn't. Here's the stuff I changed:

- Changed the drop down button size to align more with the add new button
- Changed the toggle buttons alignment so they horizontally align better with right sidebar buttons
- Fixed the overlay problem
- Stylized the add new button a bit on hover
- Allowed overflow on source tabs hence when they're active, there is no bottom border.. so it looks more like a tab
- Removed the new lines finally haha

jasonLaster:
I dropped the new tab button styling from this PR because I'm not sure what treatment would be best.

Using the close button UI is not a bad idea though. Perhaps it's a little strong through.

binarybaba:
Hey it was your call in the end so kudos to you! I had real fun though. Thanks!

jasonLaster:
Likewise! 

Stick around. Stop by our office hours next week!

binarybaba:
Absolutely!

Issue #1507, opened 2016-12-20
[RTL] Improve search bars


jasonLaster:
This issue tracks cleaning up the file search and source search bar components in RTL mode.

Getting started
* here are some [rtl](https://github.com/devtools-html/debugger.html/blob/master/docs/local-development.md#rtl) docs.
* we use postcss bi-direction for some magic CSS RTL clauses

#### Source Search
![](http://g.recordit.co/ZY9SfeMTvL.gif)

#### File Search
![](http://g.recordit.co/osqkiSrJDG.gif)

liorpr:
@jasonLaster hmm should it remain open?

jasonLaster:
Nope!

PR #1551, opened 2016-12-27
changed search bars styling for rtl

liorpr:
Associated Issue: #1507

adjusted the styling to support rtl, with the postcss-bidirection thing.

please note that I forced some strings to be ltr-displayed with css, because their text comes from strings.json so I figured they should always be ltr, right?

### DIFF
--- a/src/components/Autocomplete.css
+++ b/src/components/Autocomplete.css
@@ -10,7 +10,6 @@
 
 .autocomplete .results-summary {
   margin: 10px;
-  direction: ltr;
   text-align: start;
 }
 
@@ -94,7 +93,6 @@
   font-size: 24px;
   padding: 4px;
   word-break: break-all;
-  direction: ltr;
 }
 
 .autocomplete .no-result-msg .sad-face {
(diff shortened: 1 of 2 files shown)


src/components/Autocomplete.css
```diff
@@ -10,6 +10,8 @@
 
 .autocomplete .results-summary {
   margin: 10px;
+  direction: ltr;
```
jasonLaster: ahh - that's not necessary. We use `strings.json` and `debugger.properties` so that inside firefox we can load other properties files for RTL languages

src/components/EditorSearchBar.css
```diff
@@ -8,7 +8,7 @@
 
 .search-bar i {
   display: block;
-  padding: 13px 0 0 13px;
+  padding: 13px;
```
jasonLaster: should this be `padding 13px 0`?

liorpr: there was a need for right/left padding, and I could do it with the postcss thing but it has a box sizing: border box so I figured why not have padding from both of them

src/components/Autocomplete.css
liorpr: got it. even took a picture of the debugger with the page on rtl and replaced the not-found-msg with a hebrew one :)

screen shot 2016-12-28 at 8 33 20 pm

Issue #1536, opened 2016-12-22
Make debugger UI take less vertical space


nt1m:
Vertical space is very valuable especially when the debugger is inside the toolbox. It'd be nice if the heights were consistent with other tools inside the toolbox as well (I think we use 24px for toolbars).

clarkbw:
> I think we use 24px for toolbars

I'm not sure if this was intentional or not.  Will need to investigate.

wldcordeiro:
I could have sworn there was an effort for the vertical rhythm not too long ago and we followed guidelines @helenvholmes shared with us, but I agree that we should be consistent between tools.

nt1m:
@clarkbw,@wldcordeiro I think the current height (24px) was decided before Helen's arrival (I remember it was discussed with Brian and Darrin). Our toolbars used a lot of vertical space before then, and 24px seemed like a reasonable height.

Either way, if there are new guidelines, all tools should follow them, not just debugger-html

gabrielluong:
I believe we should be trending towards having more space (vertical space and bigger fonts) - since this helps with legibility as our font is already really tiny. I would hate to see the debugger.html adapt to the old standard when it should be the other way around where all the other tools should be adapting to the new standard.

clarkbw:
Yep, I agree. There really isn't enough space with 24px.

nt1m:
@gabrielluong @clarkbw 
For debugger.html, I would agree 24px is not enough vertical space (considering the crowded tabbar), but for the other tools, considering the low amount of content in the toolbars, 30px would be a waste, and the content would fit fine in 24px.

Here's how the toolbox used to look like if you're curious: https://people-mozilla.org/~shorlander/files/devTools-toolbox-i02/devTools-toolbox-i02.html

Our default toolbox height is 250px, and I don't think we should go above 10% for the toolbar height.

nt1m:
@gabrielluong Also, about the font-size, it really depends on the platform you're on: Windows uses 12px, OSX uses 11px, Linux uses something huge.

gabrielluong:
@nt1m @clarkbw @jasonLaster  Honestly, the toolbox you have shown me looks fine to me. I am sure there will be situations where you are correct such as if i had a MBA 11", but what if you had a 27" monitor instead. I know full well of the fight for consistency among our tool, but I can just as well point out as many inconsistencies we already currently have such as the filter bar in the new console and all the other tools or how the style editor does search compare to the old debugger. 

[...]

PR #1553, opened 2016-12-27
move toward 24px toolbar

clarkbw:
Associated Issue: #1536 

### Summary of Changes

* Change from 30px to 24px across CSS files

### Test Plan

- [x] Command-P opens at the correct position
- [x] Toolbar buttons are vertically lined up correctly
- [ ] Tab overflow works out correctly

Finding some issues with the tab overflow, otherwise this seems to work fairly well.

screen shot 2016-12-27 at 1 31 26 pm

### DIFF
--- a/src/components/CommandBar.css
+++ b/src/components/CommandBar.css
@@ -3,14 +3,14 @@
 }
 
 .command-bar {
-  height: 30px;
-  padding: 8px 5px;
+  height: 24px;
+  padding: 4px;
 }
 
 .command-bar > span {
   cursor: pointer;
   width: 16px;
-  height: 17px;
+  height: 16px;
   display: inline-block;
   text-align: center;
   transition: all 0.25s ease;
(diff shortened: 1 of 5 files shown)


jasonLaster:
Nice!

I just merged @binarybaba's work on the source header which touched some styles as well.

RE: @gabrielluong's point about going in the direction of more space. I agree that more vertical space is better, but perhaps would prefer to be consistent.

clarkbw:
Yeah, one of those changes we should do at the same time.

nt1m:
Wow, the CSS really needs some reworking, it seems to rely a lot on dimensions of each others elements, which is quite fragile. I suspect the CSS could be improved with flexbox, I might take a look if I have some time.

clarkbw:
I'm closing this out for now.

I think it would be awesome if someone, like @nt1m, moved our CSS to something like flexbox to make it more manageable than what it is now.

jasonLaster:
Sounds good. Thanks for investigating

Issue #1502, opened 2016-12-20
[Collapse Buttons] Animate toggling


jasonLaster:
We should animate the expanding and collapsing sidebars when the buttons is clicked.

jasonLaster:
@binarybaba i think this might help if you're interested in picking it up again: 
https://github.com/devtools-html/debugger.html/issues/1503#issuecomment-270721409

wldcordeiro:
@jasonLaster the buttons themselves are animated. It's the `SplitBox` itself that isn't https://github.com/devtools-html/debugger.html/blob/master/src/components/PaneToggleButton.css#L6

binarybaba:
I'm scared of react but I'll definitely try it tomorrow.. Will see if I can wrap my head around it and tomorrow update you guys!

jasonLaster:
We just landed https://github.com/devtools-html/debugger.html/pull/1666

maybe we can borrow the same approach https://github.com/liorpr/debugger.html/blob/3fc1900cec4fccd6234b0459b0d3a6d26f7dfc9e/src/components/SecondaryPanes/CommandBar.js#L66-L77

jasonLaster:
@liorpr are you on this?

liorpr:
@jasonLaster yes!

liorpr:
got kinda stuck, when the pane collapses the split box actually just re-renders without the first pane(let's say I'm collapsing the first one) and the other one is just there, taking as much width as it needs with flex: 1. Tried to render the first pane with width: 0 and transition the width but it's so slow.

@jasonLaster

jasonLaster:
haha - screw it. this sounds too hard lol. I think react fiber will help.

jasonLaster:
As a record, the issue is that splitbox is a no-op when the pane is closed. This makes it really hard to animate. Fiber will help :)

PR #1556, opened 2016-12-28
Animate when toggling sidebars enhancement #1502

binarybaba:
Associated Issue: #1502

### Summary of Changes

* Add transitions to split boxes

### DIFF
--- a/src/components/SplitBox.css
+++ b/src/components/SplitBox.css
@@ -24,11 +24,13 @@
   flex: 1;
   min-width: 0;
   overflow: auto;
+  transition: 0.03s ease-out;
 }
 
 .split-box > .controlled {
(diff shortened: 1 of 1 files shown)


jasonLaster:
Hmm, my guess is that this will slow down dragging the sidebars.

Give it a try and let me know.

If it does, we will need to add an `animating` class that applies the transition and then remove it after the animation is done. perhaps with an animationend

binarybaba:
You _are_ right. I actually resorted to lesser timing function because it was going glitchy over 0.3s. Lemme work on this and get back to you.

jasonLaster:
haha - closing this PR. Feel free to re-open, but also there are plenty of other fish in the sea

PR #1563, opened 2016-12-28
Remove borders from inactive tabs

darkwing:
This is just an idea; I've removed borders from inactive tabs to prevent the weird-ish look of empty space to the right of inactive tabs when the "x" isn't displaying.
tabswithoutborders

### DIFF
--- a/src/components/SourceTabs.css
+++ b/src/components/SourceTabs.css
@@ -34,9 +34,8 @@
 }
 
 .source-tab {
-  background-color: var(--theme-toolbar-background-alt);
   color: var(--theme-faded-tab-color);
-  border: 1px solid var(--theme-splitter-color);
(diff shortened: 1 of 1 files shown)


darkwing:
Feel free to close if we don't want to go this route.

clarkbw:
I like it. I still want some hover effect for the non active tabs, perhaps just the background.

jasonLaster:
@darkwing I like it!

darkwing:
Updated.  
with-background-color

The border does become somewhat awkward, let me know what you think.

darkwing:
NM, updated and now looks cool:

updated

jasonLaster:
I love this thread. So much nice polish. 

Have you thought about the new tab button relative to an inactive and active tab? Looks okay now, just curious

PR #1565, opened 2016-12-28
Set collapse/expand cursor to pointer to show it's clickable

darkwing:
All other clickable icons have this cursor -- updating to use the same cursor for integrity.


clarkbw:
👍

PR #1567, opened 2016-12-28
Dir config

jasonLaster:
Updates our docs on RTL based on this launchpad change: https://github.com/devtools-html/devtools-core/pull/44

CC @nchevobbe, @juliandescottes, @clarkbw


jasonLaster:
This needs a launchpad version bump to be true, but given that it doesn't break I think it's safe to merge.

PR #1576, opened 2016-12-31
Fix Open Office Hours link

joshunger:
Fixed the Open Office Hours link

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ We strive for collaboration with [mutual respect for each other][contributing]. 
 
 We're all on Mozilla's IRC in the [#devtools-html][irc-devtools-html] channel on irc.mozilla.org.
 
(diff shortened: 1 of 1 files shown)


joshunger:
@jasonLaster I don't think your calendar link is repeating?  Is is supposed to be every Tuesday? 
image

jasonLaster:
Thanks! We're struggling with this link.

jasonLaster:
@joshunger I switched the link to the hangouts, which is probably slightly better. Devs can always add their own event in their calendar :)

@clarkbw curious what you think

joshunger:
@jasonLaster looks good how do you drink with the team?

jasonLaster:
Haha - we all have a beer at our own computers. Basically, when you open a tab, you're just using your browser...

Issue #1568, opened 2016-12-28
[Editor Search Bar] Cmd+F selected text


jasonLaster:
There's a small issue with cmd+f with selected text:

First turn on `editorSearch`, then: 

STR:
1. open a source
2. select text
3. cmd+f
4. hit enter

you expect to go to the next result, instead the selected text is replaced with the previous query

You expect, cmd+f to set the query, instead it looks like it's just setting the input value.

```patch

diff --git a/src/components/EditorSearchBar.js b/src/components/EditorSearchBar.js
index 75c3a53..a9f77af 100644
--- a/src/components/EditorSearchBar.js
+++ b/src/components/EditorSearchBar.js
@@ -107,6 +107,7 @@ const EditorSearchBar = React.createClass({

     if (this.state.enabled) {
       const selection = this.props.editor.codeMirror.getSelection();
+      // looks like we're not updating the query here
       this.setSearchValue(selection);
       this.selectSearchInput();
     }
```

![](http://g.recordit.co/DWlVBeuNGz.gif)

PR #1580, opened 2017-01-01
Fix Cmd+F with selected text issue

bomsy:
Associated Issue: #1568

### Summary of Changes

* Move reoccurring code into doSearch
* Call from multiple places

### Test Plan

- [x] Open a source
- [x] Select text
- [x] Cmd+F
- [x] Hit enter
- [x] It should move to the next result
### Screenshots/Videos
cmd f_sel_text


jasonLaster:
Nice fix. I like how you simplified the logic. 

@arthur801031 mind taking a look?

Issue #1558, opened 2016-12-28
[Breakpoints] Move toggle button


jasonLaster:
@Dalimil's great work in #1546 got me to re-think how we want to treat toggling breakpoints.

Previously, the button was entirely user-set i.e. I click it to disable all bps, i click it again to re-enable all bps. This was a broken metaphor because you could also enable and disable bps individually. His PR, made it act like a group checkbox which accurately reflects the BPs aggregate state.

Given that, it would be nice to change the iconography to that of a checkbox so that its clearer for the user. Also, it would be nice to move the button to the breakpoints header so it's not confused with other App state. 

Down the road, I see us using the breakpoints header space better, so this is inline with the direction we're going

![breakpoints-checkboxes](https://cloud.githubusercontent.com/assets/254562/21524880/633ba3d0-cce7-11e6-95ee-8e898814708b.jpg)

CC @clarkbw, @Dalimil

Dalimil:
I'm working on this... but the component state is a bit complicated to maintain because 'indeterminate' can only be sent via JavaScript and 'checked=true' seems to override 'indeterminate=true'. It would need to be an uncontrolled input that is initialized after it mounts...

PR #1581, opened 2017-01-01
Fix1558

Dalimil:
Fixes Issue: #1558 
Component state is a bit complicated to maintain because 'indeterminate' can only be sent via JavaScript and 'checked=true' seems to override 'indeterminate=true'. So I created an uncontrolled input that is initialized everytime it mounts...
I have one concern though, because toggleAllBreakpoints simply modifies the state by toggling breakpoints one-by-one, the state of our checkbox toggle will reflect that and when a user clicks disableAll, there will be a short time period when the checkbox is indeterminate (just before disabling the last breakpoint) so that's why it flickers and looks a bit unsteady.  
The same thing happens whenever breakpoints are loading - this if conditional was there before (with the original switch button `if (breakpoints.size == 0 or breakpointsLoading)`) but it might also cause the flicker now... We could probably remove it and just have this conditional for the input's `disabled` property. 

Need feedback...

@jasonLaster 

out


jasonLaster:
Great GIF! Looks awesome. 

@clarkbw thoughts?

jasonLaster:
I don't have a great solution for the flickering. We can brainstorm later

wldcordeiro:
This needs a rebase on top of the changes for vertical layout. @Dalimil

src/components/SecondaryPanes.js
```diff
@@ -45,6 +51,39 @@ const SecondaryPanes = React.createClass({
...
+      className: boxClassName,
+      onClick: () => toggleAllBreakpoints(!breakpointsDisabled),
+      defaultChecked: true,
+      ref: (input) => {
+        if (input != null) {
+          const isIndeterminate = !breakpointsDisabled &&
+            breakpoints.some(x => x.disabled);
+          input.indeterminate = isIndeterminate;
+          // checked=true would override the indeterminate value
+          input.checked = !breakpointsDisabled && !isIndeterminate;
```
jasonLaster: why can't these be normal props?

nt1m: Was going to say the same thing  :) They both are standard dom properties, so it should just work as normal props

Dalimil: From what I know about the indeterminate property, it isn't possible to set it in HTML - only in JavaScript.
`checked` on the other hand could be a standard prop as you mention but it depends on the indeterminate status (we don't want to override the value by setting checked to true). But you're right that `checked` could be set as a prop - but we would need to check the indeterminate condition there too...

Dalimil:
I have moved the 'checked' property from the `ref` code and also hopefully fixed some of the flickering by putting the `breakpointLoading` condition as a toggle for the `disabled` state of the input (instead of returning some kind of a new default checkbox, it just greys out the existing in that case).
out

jasonLaster:
I pushed a couple of changes. Unfortunately they're amended to your flickering commit.

* updated checked logic so that we don't show the checkbox as checked when there aren't any breakpoints
* updated the `if` to just be `if (input)`

PR #1582, opened 2017-01-02
Switch "Office Hours" to "Community Call"

jasonLaster:
I've been thinking for the past month that office hours is not the best name for the call for a couple reasons:

1) it's actually very vague. Most people ask me what it is
2) it creates an employee / volunteer distinction. Office hours is like seeing the professor and has office in the title. In fact that is the opposite of what we're trying to do here. We are all working on the project together and the call is about discussing goals, plans, etc

Why "Community Call"?
- it communicates that it is open to everyone

Other ideas?

CC @clarkbw, @bomsy, @wldcordeiro

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ We strive for collaboration with [mutual respect for each other][contributing]. 
 
 We're all on Mozilla's IRC in the [#devtools-html][irc-devtools-html] channel on irc.mozilla.org.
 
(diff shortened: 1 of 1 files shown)


jbhoosreddy:
Community call sounds great and it's not confusing. I hadn't thought much about it, but office hours does remind me more of college than anything 😁

wldcordeiro:
I like the idea, there's definitely some college connotations with "office hours".

bomsy:
Nice!

jasonLaster:
Thanks @bomsy, @jbhoosreddy, @wldcordeiro. 

Definitely don't want to make people think they're in college.

Issue #1491, opened 2016-12-19
[Watch Expressions] Cleanup input UI


jasonLaster:
The watch expressions input style should be tweaked:

Tasks
- [ ] clicking the "+" button opens and focuses the input
- [ ] clicking away (unfocusing) the input closes it
- [ ] the input should be styled for light and dark themes

```diff
diff --git a/src/components/Expressions.css b/src/components/Expressions.css
index 46fccb7..8bb54f0 100644
--- a/src/components/Expressions.css
+++ b/src/components/Expressions.css
@@ -20,6 +20,7 @@
   color: var(--theme-body-background) !important;
 }

+/* here are the styles */
 .expression-output-container .close-btn {
   width: 6px;
   height: 6px;
diff --git a/src/components/Expressions.js b/src/components/Expressions.js
index cb979f2..acf4924 100644
--- a/src/components/Expressions.js
+++ b/src/components/Expressions.js
@@ -68,6 +68,7 @@ const Expressions = React.createClass({
   },

   renderExpressionUpdating(expression) {
+    // here is the input element
     return dom.span(
       { className: "expression-input-container" },
       dom.input(
```

### screenshots

![screen shot 2016-12-19 at 6 19 20 pm](https://cloud.githubusercontent.com/assets/254562/21332624/ba01777a-c617-11e6-859c-c75be4d7a64f.png)
![screen shot 2016-12-19 at 6 18 35 pm](https://cloud.githubusercontent.com/assets/254562/21332625/ba07b86a-c617-11e6-8081-8fa9a3c671d7.png)

liorpr:
I guess the right way to handle the "clicking away (unfocusing) the input closes it" is to add a click event listener on the whole debugger when the RightSidebar component mounts, and toggle the expressionInputVisibility flag.

@jasonLaster  what do you think?

jasonLaster:
There should be an input unfocus event... I'm away from my computer though
:)

jasonLaster:
@liorpr are you on this?

jasonLaster:
@liorpr mind if i snag this? I'm still not thrilled with the UX

liorpr:
@jasonLaster oh sorry! totally missed it! hmm I guess I don't mind, would be happy to take it after you make up your mind about it :)

PR #1593, opened 2017-01-02
make UI changes to Watch Expressions Component (#1491)

jbhoosreddy:
Associated Issue: #1491

### Summary of Changes

* removed the "+" button
* improve style for light and dark themes
* improved watch expressions UI

### DIFF
--- a/src/actions/pause.js
+++ b/src/actions/pause.js
@@ -10,6 +10,7 @@ import type { Pause, Frame, Expression, Grip } from "../types";
 import type { ThunkArgs } from "./types";
 
 type CommandType = { type: string };
+type frameIdType = string | null;
 
 /**
  * Redux actions for the pause state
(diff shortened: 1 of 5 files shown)


jasonLaster:
CC @clarkbw Jaideep and I talked about the UI and agreed that persisting the select box was a *simpler* design and perhaps a little cleaner.

clarkbw:
👍 it only takes up one extra line and makes it much simpler in general.

src/actions/pause.js
```diff
@@ -10,6 +10,7 @@ import type { Pause, Frame, Expression, Grip } from "../types";
 import type { ThunkArgs } from "./types";
 
 type CommandType = { type: string };
+type frameIdType = string | null;
```
wldcordeiro: Could we do this as a maybe? `string?`

jasonLaster: Hmm, i don't think so actually. `null` has special meaning for the frameId i believe. So it's not like checking for the presence of frameTypeId.

PR #1595, opened 2017-01-02
Fix gutter context menu

jasonLaster:
oops


wldcordeiro:
What is broken? I'm able to open the menu with things as they were.

jasonLaster:
nice catch, I think this was chrome specific!

@wldcordeiro right clicking on the gutter menu:

```js
        (cm, line, eventName, event) => this.onGutterContextMenu(event)
```

Issue #1587, opened 2017-01-02
Persist sidebar pane collapse state


jasonLaster:
It would nice to be able to persist sidebar pane collapse across debugger refreshes. 

The benefit is that users who are working on a feature with the debugger in the right sidebar and don't want to see the source tree don't have to collapse the left sidebar all the time.

This is similar to this [PR](https://github.com/devtools-html/debugger.html/pull/1534/files).

```patch

diff --git a/src/components/App.js b/src/components/App.js
index 2de54ab..f02916e 100644
--- a/src/components/App.js
+++ b/src/components/App.js
@@ -54,6 +54,8 @@ const App = React.createClass({

     return {
       horizontal,
+
+      // extra credit: perhpas we want to save this state in the UI reducer :)
       startPanelCollapsed: false,
       endPanelCollapsed: false,
     };
@@ -67,13 +69,13 @@ const App = React.createClass({

   togglePane(position) {
     if (position === "start") {
-      this.setState({
-        startPanelCollapsed: !this.state.startPanelCollapsed,
-      });
+      const collapsed = !this.state.startPanelCollapsed;
+      this.setState({ startPanelCollapsed: collapsed });
+      prefs.startPanelCollapsed = collapsed;
     } else if (position === "end") {
-      this.setState({
-        endPanelCollapsed: !this.state.endPanelCollapsed,
-      });
+      const collapsed = !this.state.endPanelCollapsed;
+      this.setState({ endPanelCollapsed: collapsed, });
+      prefs.endPanelCollapsed = collapsed;
     }
   },

diff --git a/src/utils/prefs.js b/src/utils/prefs.js
index f971c64..a95f465 100644
--- a/src/utils/prefs.js
+++ b/src/utils/prefs.js
@@ -10,6 +10,8 @@ if (isDevelopment()) {

 const prefs = new PrefsHelper("devtools", {
   clientSourceMapsEnabled: ["Bool", "debugger.client-source-maps-enabled"],
+  startPanelCollapsed: ["Bool", "debugger...."],
+  endPanelCollapsed: ["Bool", "debugger...."],
   pauseOnExceptions: ["Bool", "debugger.pause-on-exceptions"],
   ignoreCaughtExceptions: ["Bool", "debugger.ignore-caught-exceptions"],
```

Here's a link for working on [**prefs**](https://github.com/devtools-html/debugger.html/blob/master/docs/local-development.md#prefs)

### Chrome

[...]

PR #1597, opened 2017-01-03
Move pane collapse state to redux and back up in prefs system

wldcordeiro:
Associated Issue: #1587 

### Summary of Changes

* refactor pane collapse state to use reducers
* back up the state in `prefs`

### Test Plan

Example test plan:

- [x] Open debugger
- [x] collapse a pane
- [x] refresh and see that pane collapse state is retained


src/reducers/ui.js
```diff
@@ -30,6 +35,13 @@ function update(state = State(), action: Action): Record<UIState> {
     case constants.SHOW_SOURCE: {
       return state.set("shownSource", action.sourceUrl);
     }
+
+    case constants.TOGGLE_PANE: {
+      prefs[`${action.position}PanelCollapsed`] = action.paneCollapsed;
```
jasonLaster: nit: could be one `panelCollapsed = \`${action.position}PanelCollapsed\`` var

jasonLaster:
Hey @wldcordeiro I pushed some changes to make the reducer less magical with concatenation because flow got scared. I think if you finish this up flow will be happy.

wldcordeiro:
@jasonLaster Flow is being weird, the issues left were a line that was too long and a type that was used but didn't exist. Oddly enough though I still see two errors from Flow but one is from something already in the codebase and the other is the same error but for some new code.

jasonLaster:
@wldcordeiro lets look at it tomorrow...

wldcordeiro:
@jasonLaster cool, that's fine. I'm still not sure what Flow is complaining about, there's nothing special about the types from what they were before. 🙁

src/reducers/ui.js
```diff
@@ -37,9 +37,13 @@ function update(state = State(), action: Action): Record<UIState> {
...
-      const paneCollapsed = `${action.position}PanelCollapsed`;
-      prefs[paneCollapsed] = action.paneCollapsed;
-      return state.set(paneCollapsed, action.paneCollapsed);
+      if (action.position == "start") {
+        prefs.startPanelCollapsed = action.paneCollapsed;
+        return state.set("startPanelCollapsed", action.paneCollapsed);
+      }
+
+      prefs.endPanelCollapsed = action.paneCollapsed;
+      return state.set("startPanelCollapsed", action.paneCollapsed);
```
wldcordeiro: `start` -> `end` here

jasonLaster:
💃 

@bomsy, @wldcordeiro we finally got the pref patches in. 

Wellington, the problem was that we weren't adding the default-prefs to the new file

Issue #1596, opened 2017-01-02
Add watch expressions by context menu in editor


jbhoosreddy:
Add ability to add watch expressions by selecting the expression in the source editor.

jbhoosreddy:
@jasonLaster I made some of the changes as per the review. The context menu disable behavior would be a separate issue. Let me know if you think other changes should be made.

jbhoosreddy:
fixed in #1600

PR #1600, opened 2017-01-03
Add watch expressions by context menu in editor (#1596)

jbhoosreddy:
Associated Issue: #1596 

### Summary of Changes

* Add context menu when string is selected in editor to add watch expression

### DIFF
--- a/src/actions/pause.js
+++ b/src/actions/pause.js
@@ -223,13 +223,18 @@ function loadObjectProperties(grip: Grip) {
  */
 function addExpression(expression: Expression) {
   return ({ dispatch, getState }: ThunkArgs) => {
-    const id = expression.id !== undefined ? parseInt(expression.id, 10) :
-    getExpressions(getState()).toSeq().size++;
-
     if (!expression.input) {
       return;
     }
 
+    const expressions = getExpressions(getState()).toSeq();
+    const matchedExpression = expressions
+      .find(e => e.input === expression.input);
+    if (matchedExpression !== undefined) {
+      return;
+    }
+    const id = expression.id !== undefined ? parseInt(expression.id, 10) :
+      expressions.size++;
     dispatch({
       type: constants.ADD_EXPRESSION,
       id: id,
--- a/src/actions/pause.js
+++ b/src/actions/pause.js
@@ -12,6 +12,10 @@ import type { ThunkArgs } from "./types";
 type CommandType = { type: string };
 type frameIdType = string | null;
 
+function expressionExists(expressions, expression) {
+  return !!expressions.find(e => e.input == expression.input);
+}
+
 /**
  * Redux actions for the pause state
  * @module actions/pause
@@ -223,18 +227,13 @@ function loadObjectProperties(grip: Grip) {
  */
 function addExpression(expression: Expression) {
   return ({ dispatch, getState }: ThunkArgs) => {
-    if (!expression.input) {
+    const expressions = getExpressions(getState()).toSeq();
+    if (!expression.input || expressionExists(expressions, expression)) {
       return;
     }
 
-    const expressions = getExpressions(getState()).toSeq();
-    const matchedExpression = expressions
-      .find(e => e.input === expression.input);
-    if (matchedExpression !== undefined) {
-      return;
-    }
-    const id = expression.id !== undefined ? parseInt(expression.id, 10) :
-      expressions.size++;
+    const id = parseInt(expression.id, 10) || expressions.size++;
+
     dispatch({
       type: constants.ADD_EXPRESSION,
       id: id,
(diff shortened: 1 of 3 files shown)


jbhoosreddy:
@jasonLaster @clarkbw let me know if there are any code changes / UI changes that could be made to improve the original PR. Thanks!

src/actions/pause.js
```diff
@@ -220,8 +220,14 @@ function loadObjectProperties(grip: Grip) {
  */
 function addExpression(expression: Expression) {
   return ({ dispatch, getState }: ThunkArgs) => {
+    const expressions = getExpressions(getState()).toSeq();
+    const filteredExpressions = expressions
+      .filter(e => e.input === expression.input);
+    if (expressions.size && filteredExpressions.count()) {
+      return;
+    }
```
jasonLaster: Hmm, perhaps we use the `includes` function

```js
function expressionExists() {
   return !!expressions.find(e => ..)
}
```

```js
if (expressionExists(expression, expressions)) {
  return
}
```

assets/locales/debugger.properties
```diff
@@ -247,6 +247,10 @@ editor.conditionalPanel.placeholder=This breakpoint will pause when the expressi
 # for navigating to a source mapped location
 editor.jumpToMappedLocation=Jump to %s location
 
+# LOCALIZATION NOTE(editor.watchExpression): Context menu item
+# for adding watch expression from Editor
+editor.watchExpression=Watch this expression
```
jasonLaster: @clarkbw 

* "Add Watch Expression"
* "Add selected text to watches"

src/components/Editor.js
```diff
@@ -127,9 +128,23 @@ const Editor = React.createClass({
...
     };
 
-    showMenu(event, [
+    const menuOptions = [
       jumpLabel
-    ]);
+    ];
+
+    if (this.editor.codeMirror.somethingSelected()) {
+      const watchExpressionLabel = {
```
jasonLaster: NIT: lets create this label regardless and only do the push if the condition is true

src/components/Editor.js
```diff
@@ -127,9 +128,23 @@ const Editor = React.createClass({
...
+        accessKey: "E",
+        disabled: false,
+        label: L10N.getStr("editor.watchExpression"),
+        click: () => this.props.addExpression({
```
jasonLaster: lets move this to an instance function as we probably want to also check to see if the cursor is in the selection

jbhoosreddy: @jasonLaster any choice on the accessKey?

jbhoosreddy: > lets move this to an instance function as we probably want to also check to see if the cursor is in the selection

couldn't find a straightforward way of doing it. basically, the cursor stays within the selection even though the pointer is right-clicking somewhere else. will follow-up on this in a subsequent PR.

src/actions/pause.js
```diff
@@ -223,13 +223,18 @@ function loadObjectProperties(grip: Grip) {
...
     if (!expression.input) {
       return;
     }
 
+    const expressions = getExpressions(getState()).toSeq();
+    const matchedExpression = expressions
+      .find(e => e.input === expression.input);
+    if (matchedExpression !== undefined) {
+      return;
+    }
```
jasonLaster: I think we can clean this up: 

```js
function expressionExists() {
   return !!expressions.find(e => ..)
}
```

```js
if (expressionExists(expression, expressions)) {
  return
}
```

src/actions/pause.js
```diff
@@ -223,13 +223,18 @@ function loadObjectProperties(grip: Grip) {
...
+    const id = expression.id !== undefined ? parseInt(expression.id, 10) :
+      expressions.size++;
```
jasonLaster: why is this more complicated than before?

src/components/Editor.js
```diff
@@ -127,9 +128,21 @@ const Editor = React.createClass({
...
+    const watchExpressionLabel = {
+      accesskey: "E",
+      disabled: !this.editor.codeMirror.somethingSelected(),
+      label: L10N.getStr("expressions.placeholder"),
+      click: () => this.props.addExpression({
+        input: this.editor.codeMirror.getSelection()
+      })
+    };
+
+    const menuOptions = [
```
jasonLaster: Hmm, we should probably still have the feature flag

```js
if (isEnabled)
  menuOptions.push(watchExpressionLabel)
```

jasonLaster:
@jbhoosreddy updated, let me know what you think

jasonLaster:
@wldcordeiro merged it as you're traveling. Let me know if you have any comments and we can easily do a follow up.

jbhoosreddy:
No it looks great! I like it!

Issue #1599, opened 2017-01-03
[Watch Expressions] Polish UI


jasonLaster:
- [x] Prevent duplicate expressions 
  - [x] Fix editing expressions 
  - [x] Re-evaluate on resume
  - [x] cursor on input should be "text"
  - [x] close button is off
  - [x] expressions should not expand by default
  - [ ] "Not Available" shouldn't be in quotes
  - [x] re-evaluate when a frame is selected
  - [x] can't remove last expression

PR #1605, opened 2017-01-03
Polish the Expressions UI

jasonLaster:
Associated Issue: #1599

  - [x] Prevent duplicate expressions 
  - [x] Fix editing expressions 
  - [x] Re-evaluate on resume
  - [x] cursor on input should be "text"
  - [x] close button is off
  - [x] expressions should not expand by default
  - [ ] "Not Available" shouldn't be in quotes
  - [x] re-evaluate when a frame is selected
  - [x] can't remove last expression

### DIFF
--- a/src/actions/pause.js
+++ b/src/actions/pause.js
@@ -25,6 +25,8 @@ type frameIdType = string | null;
  */
 function resumed() {
   return ({ dispatch, client }: ThunkArgs) => {
+    dispatch(evaluateExpressions(null));
+
     return dispatch({
       type: constants.RESUME,
       value: undefined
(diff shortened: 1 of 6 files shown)


src/actions/pause.js
```diff
@@ -272,14 +280,18 @@ function deleteExpression(expression: Expression) {
...
  * @static
  */
-function evaluateExpressions(selectedFrameId: frameIdType) {
+function evaluateExpressions(frameId: frameIdType) {
   return async function({ dispatch, getState, client }: ThunkArgs) {
     for (let expression of getExpressions(getState())) {
+      if (!expression.input) {
+        continue;
+      }
+
```
jbhoosreddy: since we are no longer adding empty expressions, this check might not be needed here anymore, maybe?

jbhoosreddy:
did leave a small comment there, but in either case this looks very good!

src/actions/pause.js
jasonLaster: true, but I like being safe as it could blow up if we still hit it

jasonLaster: perhaps we could add a console.warn here as well so that we are notified?

jbhoosreddy: that sounds like a good idea.

Issue #1607, opened 2017-01-03
Context menu disabled attribute doesn't work


jbhoosreddy:
This might be easier to explain with a gif.
When you set menu item attribute `disabled: true` the expected result should be that the menu item is greyed out and clicking on it does not do any action.
However, setting attribute disabled to true does not make any change. This might be a bug we would want to handle separately.

![](http://g.recordit.co/wqLDjIS6kp.gif)

jasonLaster:
Here's a patch for getting started

```patch
diff --git a/src/utils/menu.js b/src/utils/menu.js
index 8242913..6b1303b 100644
--- a/src/utils/menu.js
+++ b/src/utils/menu.js
@@ -49,6 +49,11 @@ if (!isFirefoxPanel()) {
 function onShown(menu, popup) {
   popup.childNodes.forEach((menuitem, index) => {
     const item = menu.items[index];
+
+    // we'll want to do something here
+    item.disabled; // or something better
+
+    // definitely don't want this onclick
     menuitem.onclick = () => {
       item.click();
       popup.hidePopup();
```

PR #1615, opened 2017-01-04
Disable menu items with disabled property

MattCain:
Associated Issue: #1607 

### Summary of Changes

* Prevent the onclick handler from being registered if the item is disabled.
* Adds styling for the disabled state.

### Test Plan

Am not that familiar with the project so didn't know a better way, I just hardcoded a line to disable the first menu item, here's a patch:

```diff
diff --git a/src/utils/menu.js b/src/utils/menu.js
index 0c3b85a..375a57e 100644
--- a/src/utils/menu.js
+++ b/src/utils/menu.js
@@ -63,6 +63,7 @@ function onShown(menu, popup) {
 
 function showMenu(e, items) {
   const menu = new Menu();
+  items[0].disabled = true;
   items.forEach(item => menu.append(new MenuItem(item)));
 
   if (isFirefoxPanel()) {
```
Then:

1. Open any source file
2. Right click on some code
3. Notice that the first item in the context menu is disabled, clicking it does nothing.

### Screenshots/Videos

debugger-context-menu-item

Hope that's what you're looking for 🙂  Let me know if any changes required. Thanks!

### DIFF
--- a/src/components/menu.css
+++ b/src/components/menu.css
@@ -17,6 +17,8 @@ menuitem {
   line-height: 20px;
   font-weight: 500;
   font-size: 13px;
+  -moz-user-select: none;
+  user-select: none;
 }
 
 menuitem:hover {
@@ -31,6 +33,7 @@ menuitem[disabled=true] {
 
 menuitem[disabled=true]:hover {
   background-color: transparent;
+  cursor: default;
 }
 
 menuseparator {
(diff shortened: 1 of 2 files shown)


MattCain:
Requesting review @jasonLaster @jlongster 🙂

jasonLaster:
Thanks @MattCain. I liked your testing approach, I did something similar as well.

I hope you don't mind, I added a couple extra details as well like the cursor. Also, this helped me notice that we still let you select items in the context menu.

By the way, one fun anecdote, when the debugger is in the Firefox toolbox panel we can actually create our own context menus so we don't use this code. This is just about adding fake context menus locally.

Issue #1626, opened 2017-01-04
[Pretty Print] Add context menu action


jasonLaster:
It would be nice if users had the option to pretty print a source from the context menu.

We can do this in both a source tree context menu and source tab

```patch
diff --git a/src/components/SourceTabs.js b/src/components/SourceTabs.js
index 1375647..4b8a540 100755
--- a/src/components/SourceTabs.js
+++ b/src/components/SourceTabs.js
@@ -104,7 +104,7 @@ const SourceTabs = React.createClass({
   },

   showContextMenu(e, tab) {
-    const { closeTab, closeTabs, sourceTabs, showSource } = this.props;
+    const { closeTab, closeTabs, sourceTabs, showSource, togglePrettyPrint } = this.props;

     const closeTabLabel = L10N.getStr("sourceTabs.closeTab");
     const closeOtherTabsLabel = L10N.getStr("sourceTabs.closeOtherTabs");
@@ -165,6 +165,15 @@ const SourceTabs = React.createClass({
       click: () => copyToTheClipboard(sourceTab.get("url"))
     };

+    const prettyPrint = {
+      id: "node-menu-pretty-print",
+      label: ...,
+      accesskey: "X",
+      disabled: false,
+      click: () => togglePrettyPrint(sourceTab)
+    };
+
```

NOTE: @jacobjzhang is working on a source tree CM in https://github.com/devtools-html/debugger.html/pull/1554.

PR #1627, opened 2017-01-04
Add a pretty print context menu option to the source tabs.

wldcordeiro:
Associated Issue: #1626 

### Summary of Changes

* add a pretty print source context menu entry to the source tabs

### Test Plan

Example test plan:

- [ ] Open the debugger and select a source file that can be pretty printed
- [ ] Right click the tab in the source tabs and select the `Pretty Print Source` option
- [ ] See that the source is pretty printed

### DIFF
--- a/src/components/SourceTabs.js
+++ b/src/components/SourceTabs.js
@@ -67,6 +67,7 @@ const SourceTabs = React.createClass({
     closeTab: PropTypes.func.isRequired,
     closeTabs: PropTypes.func.isRequired,
     toggleFileSearch: PropTypes.func.isRequired,
+    togglePrettyPrint: PropTypes.func.isRequired,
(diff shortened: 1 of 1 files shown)


src/components/SourceTabs.js
```diff
@@ -182,6 +196,10 @@ const SourceTabs = React.createClass({
       items.push({ item: showSourceMenuItem });
     }
 
+    if (isEnabled("prettyPrint")) {
+      items.push({ item: prettyPrint });
+    }
```
jasonLaster: hmm, i think its safe to assume `prettyPrint` is on :)

are we using the flag in other places? If so we should rip them out too

wldcordeiro: We aren't, I'll drop that if.

Issue #1616, opened 2017-01-04
Default Prefs should be defined in the project.


jasonLaster:
Currently, the default debugger prefs are inside of the global [preferences.js](https://dxr.mozilla.org/mozilla-central/source/devtools/client/preferences/devtools.js#94-116) file.

This creates a problem when we want to add a new pref because we *first* need to add the default pref there in M-C. In practice this can block a PR by a couple of days for a small UI pref like `scopes-pane-visible`.

This would be fixed if we have a `debugger.js` prefs file that lived in github and was synced in mc.

Existing Prefs
```
pref("devtools.debugger.enabled", true);
pref("devtools.debugger.chrome-debugging-host", "localhost");
pref("devtools.debugger.chrome-debugging-port", 6080);
pref("devtools.debugger.chrome-debugging-websocket", false);
pref("devtools.debugger.remote-host", "localhost");
pref("devtools.debugger.remote-timeout", 20000);
pref("devtools.debugger.pause-on-exceptions", false);
pref("devtools.debugger.ignore-caught-exceptions", true);
pref("devtools.debugger.source-maps-enabled", true);
pref("devtools.debugger.client-source-maps-enabled", true);
pref("devtools.debugger.pretty-print-enabled", true);
pref("devtools.debugger.auto-pretty-print", false);
pref("devtools.debugger.auto-black-box", true);
pref("devtools.debugger.workers", false);
pref("devtools.debugger.new-debugger-frontend", true);
pref("devtools.debugger.call-stack-visible", false);
pref("devtools.debugger.scopes-visible", false);

// The default Debugger UI settings
pref("devtools.debugger.ui.panes-workers-and-sources-width", 200);
pref("devtools.debugger.ui.panes-instruments-width", 300);
pref("devtools.debugger.ui.panes-visible-on-startup", false);
pref("devtools.debugger.ui.variables-sorting-enabled", true);
pref("devtools.debugger.ui.variables-only-enum-visible", false);
pref("devtools.debugger.ui.variables-searchbox-visible", false);
```

PR #1628, opened 2017-01-04
Add default prefs

jasonLaster:
Associated Issue: #1616
### Summary of Changes

This is a somewhat *dangerous* move, but I think it is okay. This adds a new `default-prefs.js` file that will be copied over to the mc on mochitest runs. This lets us add new default prefs in github. 

The obvious improvement will come in a couple days when we can limit the default prefs to just the debugger and copy the file over to `debugger.js`instead of devtools so its less risky.

r? @tromey


wldcordeiro:
@tromey approved so I'm merging, this should unblock our PRs adding some new prefs though they may need a rebase to work with these changes?

Issue #1329, opened 2016-11-28
[Editor Search Bar] Improve selection highlight


jasonLaster:
Our current search selection has two issus:

1. It highlights each token separately [gif](https://cloud.githubusercontent.com/assets/792924/20667007/0479fbc0-b55f-11e6-826f-262315a101f8.gif)
2. the highlighted text needs to be white [screen](https://cloud.githubusercontent.com/assets/2848472/20470134/9969dc56-af73-11e6-880b-36987b13002c.png)

Both of these issues can be solved by improving how we do search tokenizing.

* Here is our current [implementation](https://github.com/devtools-html/debugger.html/blob/master/src/utils/source-search.js#L58-L74)
* Here is how chrome supports it [token highlighter](https://github.com/ChromeDevTools/devtools-frontend/blob/d8b14aea25a93cb6bbceb6aaaf45765c107d1fd9/front_end/source_frame/SourcesTextEditor.js#L727-L876)

As a side note, I think this is an interesting feature that should be its own codemirror npm package.
I'd be happy to help someone setup a `codemirror-token-highlighter` github repo and npm package, which we can then pull in and others can use.

jasonLaster:
CC @Zacqary, @bomsy interested?

bomsy:
i can take this!

jasonLaster:
Hey @bomsy have you had a chance to look at this?

bomsy:
looking at this now!

PR #1629, opened 2017-01-05
Selection highlight

bomsy:
Associated Issue: #1329

### Summary of Changes

* Rewrite the tokenize function to do a character based search rather that a token based
* Add css styles for highlight search

### Test Plan

- [x] Open the todomvc example app
- [x] Select `app.js` under `examples/vanillajs/js`
- [x] Do CMD + F to open the editor search
- [x] Type `this.storage` in the search box

### Screenshots/Videos 
select_highlight

### DIFF
--- a/src/lib/codemirror-mozilla.css
+++ b/src/lib/codemirror-mozilla.css
@@ -146,21 +146,16 @@ selector in floating-scrollbar-light.css across all platforms. */
 
 .cm-highlight-full:before {
   border: 1px solid var(--theme-comment-alt);
-  border-radius: 2px;
 }
 
 .cm-highlight-start:before {
   border-left-width: 1px;
-  border-top-left-radius: 2px;
-  border-bottom-left-radius: 2px;
   border-left-style: solid;
   border-left-color: var(--theme-comment-alt);
 }
 
 .cm-highlight-end:before {
   border-right-width: 1px;
-  border-top-right-radius: 2px;
-  border-bottom-right-radius: 2px;
   border-right-style: solid;
   border-right-color: var(--theme-comment-alt);
 }
--- a/src/lib/codemirror-mozilla.css
+++ b/src/lib/codemirror-mozilla.css
@@ -125,11 +125,44 @@ selector in floating-scrollbar-light.css across all platforms. */
   background-repeat: repeat-x;
 }
 
-.cm-selecting {
-  background: none;
-  outline-width: 1px;
-  outline-style: solid;
-  outline-color: var(--theme-comment-alt);
+.cm-highlight {
+  position: relative;
+}
+
+.cm-highlight:before {
+  position: absolute;
+  border-top-style: solid;
+  border-bottom-style: solid;
+  border-top-color: var(--theme-comment-alt);
+  border-bottom-color: var(--theme-comment-alt);
+  border-top-width: 1px;
+  border-bottom-width: 1px;
+  top: -1px;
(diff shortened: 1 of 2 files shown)


bomsy:
@jasonLaster here it is.

jasonLaster:
@bomsy wow - it's so small... 

I'll be sure to check this out soon. Just figured out how to land the most up to date launchpad version

src/utils/source-search.js
```diff
@@ -48,26 +48,49 @@ function ignoreWhiteSpace(str) {
...
  *
+ * Also the token function code is mainly based of work done
+ * by the chrome devtools team. Thanks guys! :)
+ *
  * @memberof utils/source-search
  * @static
  */
 function searchOverlay(query) {
-  query = new RegExp(escapeRegExp(ignoreWhiteSpace(query)), "g");
+  query = new RegExp(escapeRegExp(ignoreWhiteSpace(query)));
```
jasonLaster: why are we dropping the "g"?

src/utils/source-search.js
```diff
@@ -48,26 +48,49 @@ function ignoreWhiteSpace(str) {
...
-        stream.pos += match[0].length || 1;
-        return "selecting";
-      } else if (match) {
-        stream.pos = match.index;
-      } else {
-        stream.skipToEnd();
+      if (stream.column() === 0) {
+        matchLength = null;
+      }
+      if (matchLength !== null) {
```
jasonLaster: could we bail early if `matchLength` is null?

src/utils/source-search.js
```diff
@@ -48,26 +48,49 @@ function ignoreWhiteSpace(str) {
...
+        stream.next();
+        const len = match[0].length;
+        if (len === 1) {
+          return "highlight highlight-full";
+        }
+        matchLength = len;
+        return "highlight highlight-start";
+      }
+      while (!stream.match(query, false) && stream.peek()) {
+        stream.next();
```
jasonLaster: alright, i'm definitely playing with this locally.... this is wizardry

src/utils/source-search.js
bomsy: We are no longer looking for all the matches at once. We are now want one match at a time.

src/utils/source-search.js
bomsy: No, cos we want to run code from `line 82` if the matchLength is null because we might not have finished looking at all the characters in the file

src/utils/source-search.js
bomsy: haha .. so `stream.match(query, false)` just does a match on the content starting from the current `stream.pos`, if no match is found and there is still content left ahead`stream.peek()`, then we just advance the   `stream.pos` one character using `stream.next()` and match again, until we've checked all the content or find another match.

src/lib/codemirror-mozilla.css
```diff
@@ -125,11 +125,39 @@ selector in floating-scrollbar-light.css across all platforms. */
...
+  border: 1px solid var(--theme-comment-alt);
+}
+
+.cm-highlight-start:before {
+  border-left-width: 1px;
+  border-left-style: solid;
+  border-left-color: var(--theme-comment-alt);
+}
+
+.cm-highlight-end:before {
```
jasonLaster: I like the `:before` `:after` here

bomsy:
Thanks! Enjoyed been part of figuring it out! :)

Issue #1637, opened 2017-01-05
[Editor Search Bar] Add smart filters


wldcordeiro:
Filters:
* regexes
* whole word match
* case insensitive

Screenshot for inspiration
![screenshot_20170105_132202](https://cloud.githubusercontent.com/assets/580982/21696352/ef69ea06-d34a-11e6-86ca-0766993aa452.png)

![](https://cloud.githubusercontent.com/assets/254562/21696191/e4a17e30-d35a-11e6-8675-4bb7ff191ca7.png)

wldcordeiro:
Closing since feature landed behind a pref.

Issue #1633, opened 2017-01-05
Highlight ocurences of the highlighted word (a-la Sublime)


nchevobbe:
This is something I really like in Sublime, it gives you a fast insight of where your variable, for example, is used.

![screenshot at janv 05 17-17-26](https://cloud.githubusercontent.com/assets/578107/21688122/f20c850e-d36b-11e6-944d-5433ed0e5d16.png)

This would be triggered when a word is double-clicked in the editor, at first.
Maybe later we could have a shortcut to highlight the current text selection (not only on double-click)

jbhoosreddy:
@jasonLaster if this is actionable I'd like to take a stab at this.

jasonLaster:
Yea - i think so... follow @bomsy's work with the source-search utils. should be a fun ride :) similar to your selections api, but with Editor dom events on mouse up or something :P

PR #1644, opened 2017-01-06
WIP: Add search modifiers to code search.

wldcordeiro:
Associated Issue: #1637 

### Summary of Changes

* Added common search modifiers
  * case sensitive (default true)
  * whole word (default false)
  * regular expressions (default false

### Test Plan

Example test plan:

- [ ] Open the code search in a file
- [ ] toggle on a modifier and search
- [ ] combine different modifier combinations to see if matches are still found

### Screenshots/Videos

screenshot_20170106_151743

@jasonLaster Still need to finish up the buttons by adding SVGs and styling them as well as some error handling when a regex that's invalid is typed (currently a syntax error occurs) but review would be nice for this to make sure the behavior matches the expected from Atom/Sublime/etc.


jasonLaster:
Cool!

I'm away from my computer until Monday.

@bomsy, @clarkbw ?

wldcordeiro:
Oh since the buttons are just blank they're from left to right, regex, case sensitivity, whole word. 😆 I'll be fixing that up soon with some SVGs.

src/utils/source-search.js
```diff
@@ -58,8 +64,12 @@ function ignoreWhiteSpace(str) {
  * @memberof utils/source-search
  * @static
  */
-function searchOverlay(query) {
-  query = new RegExp(escapeRegExp(ignoreWhiteSpace(query)));
+function searchOverlay(query, regexMatch) {
+  if (regexMatch && query !== "") {
+    query = new RegExp(query);
```
wldcordeiro: @bomsy I remember you wrote this token function recently, if you could give this a look over it'd be much appreciated as it currently makes the overlay fail for regex matched queries.

bomsy: Of the top of my head , did you try `query = new RegExp(ignoreWhiteSpace(query));` ?

wldcordeiro: @bomsy Hmm yeah that seemed to do the trick, highlight is actually showing up. This is a complicated feature 😆  Thanks!

bomsy:
@wldcordeiro Did you have any luck with that?

wldcordeiro:
@bomsy yep that seemed to do the trick to get the overlay working. 😄 If you could give different searchs some tests it'd be much appreciated, I want to make sure it works with different combinations of the modifiers (@jasonLaster when you get back we may want to see about some tests we could write. I'm not sure how to test codemirror though.)

bomsy:
@wldcordeiro cool,  will do!

jasonLaster:
re tests: you could test the regex sans code mirror. We don't need to test code mirror;)

We could also try writing storybook stories or integration tests. But not a big deal

jasonLaster:
hey @wldcordeiro what's your timeline here?

wldcordeiro:
@jasonLaster rebasing and working on it today, I'm hoping to have it ready for review tomorrow.

wldcordeiro:
@jasonLaster hit a slight snag reconciling these changes with what I'm working on. https://github.com/devtools-html/debugger.html/commit/3b5da99f3f2c830b0c69baaa225e38f56a14246c Should be updating soon.

@jbhoosreddy you added this `keepSelection` logic but it's unused in the project and most of search stuff lives in `SearchBar` do you recall the plan for this code or is it removable?

jasonLaster:
Ahh @wldcordeiro was working on this: https://github.com/devtools-html/debugger.html/issues/1633 I can see it being confusing. for the rebase

wldcordeiro:
@jasonLaster yeah, it made it slightly complicated because that addition is looking at the CM selection and changing the query based on the visual selection, but it was resetting our typed queries (for example if you opened a file on todomvc and did a regex search like `(todo?s)` the query would change to either `todo|todos` because the first visual selection would be the text (this is made even worse with case insensitive 😆) I think I've found a fix now but it needs a bit of extra testing if you could pull it down and try it out.

wldcordeiro:
Also I was trying to get Atom's SVGs for reuse but they're done in a weird `<svg><use href="stringhere"></svg>` style and hadn't been able to find the SVGs yet. @helenvholmes could you find some SVGs we could use for the button icons? 😄 

Sublime uses the three on the left here.
screenshot_20170112_174219

jasonLaster:
@wldcordeiro do you think this is them: https://github.com/atom/find-and-replace/pull/390/files

jbhoosreddy:
@wldcordeiro It is used in the Editor component at just one place. [Editor.js#L79](https://github.com/devtools-html/debugger.html/blob/master/src/components/Editor.js#L79) It's purpose is to prevent the selection from being moved to the first selection. It is only required in the Editor component as far as I can see. But it is being used there.

wldcordeiro:
@jasonLaster I've updated with the SVGs and I think the behavior is ready. Review por favor? 😄

jasonLaster:
Gladly!

jasonLaster:
Very cool! Found a couple of things

1. the search count and outlines don't seem to reflect the options (mostly whole word)
2. do we reset the search when we change options
3. I think it would be nice to try these buttons in a row below search on the right. I'm happy to jump in and help with the styling

Biggest thing would be, lets put this behind a feature flag and try and land and iterate. sorry for not suggesting this earlier.

We can land some of the functionality outside of the flag, but certainly the UI should be hidden so that you can't get in trouble.

assets/images/case-match.svg
```diff
@@ -0,0 +1,3 @@
+<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" viewBox="0 0 20 16" stroke="none" fillrule="evenodd">
```
wldcordeiro: @jasonLaster these are the SVGs from Atom, what should we do in regards to license, etc? I think we should get the preamble from their license or something.

src/components/Editor.js
```diff
@@ -102,10 +101,27 @@ const Editor = React.createClass({
 
   displayName: "Editor",
 
+  getInitialState() {
+    return {
+      query: "",
```
wldcordeiro: This state had to be moved up into the editor as getting the query from `ctx.cm.getSelection()` would clear regex/whole word searches.

wldcordeiro:
@jasonLaster 

> the search count and outlines don't seem to reflect the options (mostly whole word)
That's the area I was having the most trouble with. I'll keep testing there.

> do we reset the search when we change options
We don't. Should we though? In Atom/Sublime it seems like they can be used in combination without reseting.

> I think it would be nice to try these buttons in a row below search on the right. I'm happy to jump in and help with the styling

This was a first draft and I'm happy to iterate. @nt1m was mentioning the lack of vertical space in the panel which drove me to go with the buttons in the left hand side.

> Biggest thing would be, lets put this behind a feature flag and try and land and iterate. sorry for not suggesting this earlier.

Indeed. I'll update to put it behind a feature flag (searchModifiers).

src/components/Editor.js
```diff
@@ -63,20 +63,19 @@ function resizeBreakpointGutter(editor) {
...
+    findNext(ctx, query, caseSensitive, wholeWord, regexMatch);
   }
 }
 
-function onCursorActivity(ctx) {
+function onCursorActivity(ctx, query, caseSensitive, wholeWord, regexMatch) {
   if (ctx.cm.somethingSelected()) {
-    const query = ctx.cm.getSelection();
-    find(ctx, query, true);
+    find(ctx, query, true, caseSensitive, wholeWord, regexMatch);
```
jbhoosreddy: the find function `find(ctx, query, true, caseSensitive, wholeWord, regexMatch);` has 6 parameters while the function definition has 5. I would recommend putting keepSelection at the end or using destructured object parameters. keepSelection is required to prevent the selection from moving away.

src/utils/source-search.js
```diff
@@ -200,9 +210,10 @@ function clearSearch(cm) {
  * @memberof utils/source-search
  * @static
  */
-function find(ctx, query, keepSelection) {
+function find(ctx, query, caseSensitive, wholeWord, regexMatch) {
```
jasonLaster: I wonder if we can pass modifiers along. That way we avoid long signatures

src/components/Editor.js
wldcordeiro: `keepSelection` logic was causing some weirdness where it would reset the query to a blank string. I'll see if I can fix that and add that back in.

src/utils/source-search.js
wldcordeiro: 👍 I had kind of thought so too after looking over all the changes 😆 I'll do that, pass in a modifiers object.

wldcordeiro:
Oddest bug currently is that counts and visual selections don't update on the first modifier toggle, but will update on all subsequent toggles. 🤔

src/components/Editor.js
```diff
@@ -102,10 +101,27 @@ const Editor = React.createClass({
 
   displayName: "Editor",
 
+  getInitialState() {
+    return {
+      query: "",
+      caseSensitive: true,
+      wholeWord: false,
+      regexMatch: false
+    };
```
jasonLaster: perhaps these are also stored as searchModifiers
```js
searchModifiers: {caseSensitive: false, ...}
```

src/components/Editor.js
```diff
@@ -102,10 +101,27 @@ const Editor = React.createClass({
...
+  },
+
   contextTypes: {
     shortcuts: PropTypes.object
   },
 
+  toggleModifier(mod) {
+    this.setState({ [`${mod}`]: !this.state[`${mod}`] });
```
jasonLaster: this would probably simpler if it received modifiers

```js
this.setSate("modifiers", modifiers)
```

src/components/Editor.js
```diff
@@ -432,7 +448,9 @@ const Editor = React.createClass({
     this.editor.codeMirror.on("gutterClick", this.onGutterClick);
 
     const ctx = { ed: this.editor, cm: this.editor.codeMirror };
-    ctx.cm.on("cursorActivity", cm => onCursorActivity(ctx));
+    const { query, caseSensitive, wholeWord, regexMatch } = this.state;
+    ctx.cm.on("cursorActivity", cm =>
+      onCursorActivity(ctx, query, caseSensitive, wholeWord, regexMatch));
```
jasonLaster:
```js
onCursorActivity(ctx, query, modifiers)
```

src/utils/source-search.js
```diff
@@ -28,10 +28,20 @@ function getSearchState(cm) {
...
+    moddedQuery = new RegExp(query, caseSensitive ? "" : "i");
+  } else if (query !== "" && (wholeWord || (wholeWord && regexMatch))) {
+    moddedQuery = new RegExp(`\\b${query}\\b`, caseSensitive ? "" : "i");
+  }
+
+  if (query !== "" && typeof moddedQuery == "string") {
+    return cm.getSearchCursor(moddedQuery, pos, !caseSensitive);
+  }
+
+  return cm.getSearchCursor(moddedQuery, pos);
```
jasonLaster: this looks for a candidate where you might want to use the spy pattern https://github.com/mochajs/mocha/wiki/Spies and pass in a spy or stub function for `cm`.

I could see some unit tests really helping here :) 

as you said earlier

src/utils/source-search.js
```diff
@@ -28,10 +28,20 @@ function getSearchState(cm) {
  * @memberof utils/source-search
  * @static
  */
-function getSearchCursor(cm, query, pos) {
-  // If the query string is all lowercase, do a case insensitive search.
-  return cm.getSearchCursor(query, pos,
-    typeof query == "string" && query == query.toLowerCase());
+function getSearchCursor(
+  cm, query, pos, caseSensitive, wholeWord, regexMatch) {
+  let moddedQuery = query;
```
jasonLaster: can you explain what `moddedQuery` is doing here. how it's different than query

src/components/Editor/SearchBar.js
```diff
@@ -223,33 +253,63 @@ const SearchBar = React.createClass({
...
-    if (count == 0 && query.trim() != "") {
+    if (count == 0 && this.props.query.trim() != "") {
       return Svg("sad-face");
     }
 
     return Svg("magnifying-glass");
   },
 
+  renderSearchModifiers() {
+    const { regexMatch, caseSensitive, wholeWord, toggleModifier } = this.props;
```
jasonLaster: this looks like a good spot for a feature flag

src/utils/source-search.js
wldcordeiro: So CodeMirror's `getSearchCursor` has this signature

`getSearchCursor(query, start, caseFold)`

* query - string or RegExp object
* start - starting position in document
* caseFold - only relevant with strings but toggles case sensitive search

So the reason I had a moddedQuery was to avoid changing the original, but looking over the code there isn't much reason for it with the modifications I do.

wldcordeiro:
Updated from review. Could we pair on this a bit @jasonLaster for adding tests and bug squashing? I'm getting some weirdness where the counts get out of sync with the visual selections and buttons. 🙁

jasonLaster:
Sure. Let's definitely pair. The components look really good. Utilits will benefit from tests for sure

src/components/Editor/SearchBar.js
```diff
@@ -223,33 +252,74 @@ const SearchBar = React.createClass({
...
 
     return Svg("magnifying-glass");
   },
 
+  renderSearchModifiers() {
+    const {
+      modifiers: { caseSensitive, wholeWord, regexMatch },
+      toggleModifier } = this.props;
+
+    if (isEnabled("searchModifiers")) {
```
wldcordeiro: @jasonLaster should I also add the feature to some of the configs? I hadn't used this before so I'm not sure if that's also needed (I had to add it to my local.json to test the feature.)

jasonLaster:
Yea. Development.json

src/components/Editor/SearchBar.js
jasonLaster: lets flip this

```js
if (!enabled) {
return
}
```
help keep the indentation under control

jasonLaster:
Here's a quick sketch of where I see the buttons:

I think the buttons would be nice default `--color-comment` and then the active blue. Just a guess though

screen shot 2017-01-23 at 8 11 22 pm

wldcordeiro:
@jasonLaster Huh weird, I had them on the left hand side of the search input and the svg color changes to white when active. 
screenshot_20170123_183243
screenshot_20170123_183249

src/components/Editor/SearchBar.js
jasonLaster: yeah, add it to development.json and feature.json :) 

just wrote some docs for this

Issue #1604, opened 2017-01-03
[WhyPaused] Style the UI


jasonLaster:
There are a couple things that are currently off with the UI:

* the height is should be consistent with headers
* there should be a bottom bar
* the color is arbitrary, we should choose standard colors that look good in dark mode
* we should check to see that it looks good in vertical mode
* we should make sure it looks good while scrolling

```patch
diff --git a/src/components/WhyPaused.js b/src/components/WhyPaused.js
index 6b4f90d..5d2910e 100644
--- a/src/components/WhyPaused.js
+++ b/src/components/WhyPaused.js
@@ -21,6 +21,7 @@ const WhyPaused = React.createClass({
     const { pauseInfo } = this.props;
     const reason = getPauseReason(pauseInfo);

+    // => here
     return reason ?
       dom.div({ className: "pane why-paused" }, L10N.getStr(reason))
       : null;
```

* scrolling

![](http://g.recordit.co/3hBYHzO5Ag.gif)

* vertical mode

![screen shot 2017-01-03 at 5 16 36 pm](https://cloud.githubusercontent.com/assets/254562/21625174/6fa8a528-d1d8-11e6-9ec8-122ba3aa5fde.png)

krishnagopinath:
I'd like to work on this issue, @jasonLater, if it's not already taken by someone else :)

This would be my first ever contribution to OS ever,  so I'm pretty sure I'd make a lot of whoopsies.

jasonLaster:
Hi @krishnagopinath. It's yours, feel free to reach out with any questions

krishnagopinath:
So, I've got a couple of suggestions for this (thanks to the chat I had with @jasonLaster ), other than reducing the font size, and the height, and adding borders :

* The red dot - Here's a little preview as to how it'll look.. of course, this looks rough and ugly but once I've done some CSS work on it, it'll look much better! (Look for the red dot next to the "Paused on breakpoint" message

![why-paused-badge](https://cloud.githubusercontent.com/assets/4935809/21704211/c42b2da6-d385-11e6-9bb4-6a75b16d187d.png)

If I were to consider adding the small red dot before the text in the `why-paused` div, I'd have to use an SVG, from the looks of it. [Can I use an SVG from anywhere on the web?](https://commons.wikimedia.org/wiki/File:Location_dot_red.svg) 

I see that other SVGs in the `assets/images` folder have an MPL as the first line, hence the question. 

[...]

PR #1646, opened 2017-01-07
Styling the [WhyPaused] UI

krishnagopinath:
Associated Issue: #1604 

Just minor CSS changes such as 
* background
* border
* font-size

### DIFF
--- a/src/components/WhyPaused.css
+++ b/src/components/WhyPaused.css
@@ -1,16 +1,16 @@
 
 .why-paused {
-  background-color: var(--breakpoint-active-color) !important;  
+  background-color: var(--breakpoint-active-color) !important;
   border: 1.7px solid var(--breakpoint-active-color);
   color: var(--theme-highlight-blue);
   padding: 10px 10px 10px 20px;
   white-space: normal;
   opacity: 0.9;
-  font-size : 12px;
+  font-size: 12px;
   text-align: center;
   font-weight: bold;
 }
 
 .theme-dark .why-paused {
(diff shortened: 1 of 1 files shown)


krishnagopinath:
I couldn't check this layout in vertical mode because, try as I could, I couldn't make the layout to switch to that mode. Wasn't it supposed to "flip" when the width was at 700px?

jasonLaster:
@krishnagopinath you can turn on vertical mode in local.json. 

Checkout our local development doc for more info.

src/components/WhyPaused.css
```diff
@@ -1,7 +1,16 @@
...
   padding: 10px 10px 10px 20px;
   white-space: normal;
+  opacity: 0.9;
+  font-size: 12px;
+  text-align: center;
+  font-weight: bold;
+}
+
+.theme-dark .why-paused {
+  color: white !important;
```
jasonLaster: Haha can we add more specificity to avoid imports to calls?

irfanhudda:
@krishnagopinath It probably means that you need to avoid using "!important" whenever possible.  It is not considered as good practice. 

Most of the times you can avoid them by using better(more specific) CSS selectors.

krishnagopinath:
@jasonLaster check the latest changes. Tell me if I need to change anything more

jasonLaster:
Here are some screens

screen shot 2017-01-09 at 10 00 25 am
screen shot 2017-01-09 at 9 59 39 am

PR #1656, opened 2017-01-09
highlight and search on editor selection (#1633)

jbhoosreddy:
Associated Issue: #1633

### Summary of Changes

* Highlight and search on editor selection by double click
* remove highlight when selection removed
* select previous and next text with key shortcuts.

### DIFF
--- a/src/components/Editor.js
+++ b/src/components/Editor.js
@@ -8,6 +8,7 @@ const { connect } = require("react-redux");
 const classnames = require("classnames");
 
 const SourceEditor = require("../utils/source-editor");
+const { find, findNext, findPrev, removeOverlay } = require("../utils/source-search");
 const SourceFooter = createFactory(require("./SourceFooter"));
 const SearchBar = createFactory(require("./Editor/SearchBar"));
 const { renderConditionalPanel } = require("./Editor/ConditionalPanel");
@@ -60,6 +61,25 @@ function resizeBreakpointGutter(editor) {
   const lineNumbers = gutters.querySelector(".CodeMirror-linenumbers");
   const breakpoints = gutters.querySelector(".breakpoints");
   breakpoints.style.width = `${lineNumbers.clientWidth}px`;
+}
+
+function traverseResults(e, ctx, dir) {
+  e.stopPropagation() || e.preventDefault();
+  const query = ctx.cm.getSelection();
+  if (dir == "prev") {
+    findPrev(ctx, query);
+  } else if (dir == "next") {
+    findNext(ctx, query);
+  }
+}
+
+function onCursorActivity(ctx) {
+  if (ctx.cm.somethingSelected()) {
+    const query = ctx.cm.getSelection();
+    find(ctx, query, true);
+  } else {
+    removeOverlay(ctx);
+  }
 }
 
 const Editor = React.createClass({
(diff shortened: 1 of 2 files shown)


src/components/Editor.js
```diff
@@ -382,6 +383,45 @@ const Editor = React.createClass({
...
+    const { query } = this.state;
+    findPrev(ctx, query);
+  },
+
+  traverseResultsNext(e, ctx) {
+    e.stopPropagation();
+    e.preventDefault();
+    const { query } = this.state;
+    findNext(ctx, query);
+  },
```
jasonLaster: hmm, i think these function could receive query and be moved to helpers above the component.

My rationale is that we already have a lot of instance functions :)

src/components/Editor.js
```diff
@@ -382,6 +383,45 @@ const Editor = React.createClass({
...
+    const { searchOverlayMode } = this.state;
+    if (searchOverlayMode) {
+      if (!ctx.cm.somethingSelected()) {
+        removeOverlay(ctx);
+        this.setState({
+          searchOverlayMode: !searchOverlayMode,
+          query: ""
+        });
+      }
+    }
```
jasonLaster: couple things:

1. this seems like a mechanism for removing the search overlay mode. Is that correct?
2. could we also do a serch from here when the user lifts the mouse up
3. the IFs could be collapsed into one `if (searchOverlayMode && !ctx.cm.somethingSelected()) {}`
4. the searchOverlayMode would always be true, so in this case lets just do false... seems more honest

src/components/Editor.js
```diff
@@ -382,6 +383,45 @@ const Editor = React.createClass({
...
+    const query = this.editor.codeMirror.getSelection();
+    this.setState({ searchOverlayMode: true, query });
+    find(ctx, query, true);
+  },
+
+  getInitialState() {
+    return {
+      searchOverlayMode: false
+    };
+  },
```
jasonLaster: shocked that this is the first one

nchevobbe:
This is  really cool !
I find the styling of the outlined element a bit strange, it looks like it cuts-off a bit of them

capture du 2017-01-10 08-41-07

And a bit weird on the selection too (looks like a double border) : 
capture du 2017-01-10 08-42-39

If we could have some padding on this outline it would be great

Another nit, out of scope here, is the selection color in dark mode, which isn't contrasted enough with the text's color (see http://leaverou.github.io/contrast-ratio/#%23E4EEFA-on-%23BDD5FD)

jasonLaster:
@nchevobbe good points. @jbhoosreddy if it's not easy to address, we'll likely want to open a new issue for that.

@bomsy probably has thoughts on class names for styling here

bomsy:
yeah ... we can definitely make the styling better, with paddings and a lil bit of rounded corners. 
The classes responsible are 

https://github.com/devtools-html/debugger.html/blob/master/src/lib/codemirror-mozilla.css#L128-L161

We can do it as another ticket.

jasonLaster:
@jbhoosreddy this PR has gotten really good.

jbhoosreddy:
I looked into improving the CSS for codemirror. I was gonna add it to this
cause it's mostly done. But I can add it in a subsequent PR.

Issue #1648, opened 2017-01-09
Hide "toggle breakpoints" checkbox when there aren't any breakpoints


jasonLaster:
We recently moved the toggle breakpoints to a checkbox in the breakpoints pane:

https://twitter.com/jasonlaster11/status/816314230999838720

It would be nice if we hid the checkbox when there weren't any breakpoints:

```patch

diff --git a/src/components/SecondaryPanes.js b/src/components/SecondaryPanes.js
index b771f43..950ffdf 100644
--- a/src/components/SecondaryPanes.js
+++ b/src/components/SecondaryPanes.js
@@ -61,6 +61,10 @@ const SecondaryPanes = React.createClass({
     const isIndeterminate = !breakpointsDisabled &&
       breakpoints.some(x => x.disabled);

+    if (!breakpointsExist) {
+      return;
+    }
+
     return dom.input({
       type: "checkbox",
       className: boxClassName,
```

PR #1657, opened 2017-01-09
Do not render breakpoint toggle if none exist.

Garbee:
Associated Issue: #1648

### Summary of Changes

* Check for breakpoint(s) existence before rendering the toggle button.

### Test Plan

Open up clean debugger profile. See by default there is no breakpoint toggle checkbox. Set a breakpoint somewhere and verify the checkbox then renders.

### Screenshots

No Breakpoints:

no-points

Breakpoints:

breakpoints

### DIFF
--- a/src/components/SecondaryPanes.js
+++ b/src/components/SecondaryPanes.js
@@ -59,7 +59,11 @@ const SecondaryPanes = React.createClass({
     const isIndeterminate = !breakpointsDisabled &&
       breakpoints.some(x => x.disabled);
 
-    return breakpoints.size > 0 ? dom.input({
+    if (breakpoints.size == 0) {
+      return null;
+    }
+
+    return dom.input({
       type: "checkbox",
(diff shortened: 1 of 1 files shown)


Garbee:
Just updated the PR to also add an `aria-label` since the checkbox doesn't have a label. To increase a11y handling. Also refactored a little bit to remove the `breakpointsExist` variable that is no longer necessary for re-use since the node is never rendered unless it is true and all of its usages within the node were checking for it to be true.

wldcordeiro:
@Garbee Maybe only show it when we have more than one? A toggle for all breakpoints when there's only one seems a little pointless.

Garbee:
Can you join in on our call today at 3 EST? I'm basically rethinking this entire interface today. Will discuss more there first to show things running for clarity.

jasonLaster:
@Garbee I'll be there. Excited about this work.

I pushed a small style nit fix: replace the ternary with a guard statement. let me know if it's alright and we can merge. otherwise happy to revert

Issue #1609, opened 2017-01-03
editing watch expressions does not work


jbhoosreddy:
When editing a watch expression, after editing and pressing enter, the expression text disappears and the input field stays there.

![](http://g.recordit.co/YTwHz47qc1.gif)

jasonLaster:
Here's a good place to look:

```diff






diff --git a/src/components/Expressions.js b/src/components/Expressions.js
index 976e73b..739edd4 100644
--- a/src/components/Expressions.js
+++ b/src/components/Expressions.js
@@ -48,6 +48,8 @@ const Expressions = React.createClass({
     if (e.key !== "Enter") {
       return;
     }
+
+// => something in here
     const { addExpression } = this.props;
     const expression = {
       input: e.target.value,
```

jasonLaster:
@andreicristianpetcu it's yours

andreicristianpetcu:
I made a commit. It seems to work. I will test more in the following days to see if I missed something. The Expressions seem to have no unit tests. I will try to make a new test similar to the [other ones](https://github.com/devtools-html/debugger.html/tree/45a276cb6864cf826ddb2c94d933ecef3c9f3272/src/components/tests) but I will cover just this functionality.

andreicristianpetcu:
Hmmm. It seems adding a test is a little more complicated. The current component tests [don't seem like they are run at the moment](https://github.com/devtools-html/debugger.html/blob/45a276cb6864cf826ddb2c94d933ecef3c9f3272/src/test/node-unit-tests.js#L61-L64) and they pass if I [run them but don't seem to check anything](https://github.com/devtools-html/debugger.html/blob/45a276cb6864cf826ddb2c94d933ecef3c9f3272/src/components/tests/Breakpoints.js#L18).

I will make a pull request with the current work. I will start making unit tests in the future on a future task and start working on an existing unit test and not add support for it on some file that are never run. I think it probably is a good reason why they are not run at the moment.

What do you think?

Issue #1650, opened 2017-01-09
[Watch Expressions] Refactor some pieces


jasonLaster:
- [x] Extract expressions actions and reducer
- [ ] remove expression id (because they're unique!)

PR #1659, opened 2017-01-10
 #1609 first draft delete expression if value is empty

andreicristianpetcu:
Associated Issue: #1609

### Summary of Changes

* If you edit an existing watch expression and delete it's name and press enter then the watch expression is deleted and you can add a new watch expression.

### Test Plan

Example test plan:

- [x] Open the debugger and set a breakpoint somewhere (let's say todomvc's toggle item in the todo model)
- [x] Run the todo app to reach the breakpoint
- [x] When the debugger stops in the breakpoint add a new watch expression named "this" and press enter
- [x] Click the watched expression to edit it. Press backspace until you deleted all the characters and then press enter.
- [x] The watched expression is now deleted and you can add more.

### DIFF
--- a/src/components/Expressions.js
+++ b/src/components/Expressions.js
@@ -48,14 +48,23 @@ const Expressions = React.createClass({
     if (e.key !== "Enter") {
       return;
     }
-    const { addExpression } = this.props;
-    const expression = {
-      input: e.target.value,
-      id
-    };
+
+    const {
+        addExpression,
+        deleteExpression
+    } = this.props;
+
+    if (e.target.value != null && e.target.value.trim()) {
+      const expression = {
+        input: e.target.value,
+        id
+      };
+      addExpression(expression);
+    } else {
+      deleteExpression({ id });
+    }
 
     e.target.value = "";
-    addExpression(expression);
   },
 
   updateExpression(expression, { depth }) {


andreicristianpetcu:
Found some issues with the current implementation. I will send a new PR.

wldcordeiro:
@andreicristianpetcu you could use the same PR and amend the previous commit and force push the changes.

`git commit --amend`
`git push -f`

andreicristianpetcu:
@wldcordeiro I will do as you say if you wish since you one of the maintainers of debugger.html but I generally try to avoid rewriting git history and `git push -f`. I think I can reuse this PR without amending. I can just reopen it. I tend to prefer small commits.

jasonLaster:
@andreicristianpetcu yeah - it's unfortunate, but the way github works. you generally have to update local history with a rebase and force push to your remote branch to keep the PR.

The benefit ofcourse to having one PR is issues and PRs have the same Unique ID, and it's nice to have a small number and smaller history to search through.

andreicristianpetcu:
I can reopen the pull request without messing with the git history. I don't need to change anything. I'll close it back.

I think I'll give another try at writing a test. I managed to break editing and I think it's better to add a test with that feature even if it is a little more complicated.

jasonLaster:
heh @andreicristianpetcu I was going to merge this as it's an improvement

I found a similar bug with hitting enter immediately. I believe that's because we don't clear the updating flag.

Frankly, I think this would be simpler if expressions had its own reducer and action and we had a CRUD api, so add, update, get, list etc... right now I think addExpression is working overtime.

also expression ID should not be needed now that expression values are unique!

andreicristianpetcu:
@jasonLaster I reopened the pull request.

It's not clear for me what I should do. Do you want me to rebase this change on the latest stuff and remove the ID?

Expressions seem to be just a list of strings now. I don't understand how you want that converted to a CRUD.

jasonLaster:
Sorry @andreicristianpetcu I think you walked into a mess honestly. Mind if I take this over in #1650?

andreicristianpetcu:
No problem @jasonLaster. I will pick something else up and try to avoid [these](https://github.com/devtools-html/debugger.html/labels/Release%3A%20Commitment) and [these](https://github.com/devtools-html/debugger.html/milestone/15) :)

I want to work on something without slowing you down.

jbhoosreddy:
@andreicristianpetcu: Talking about something else entirely, you could always keep adding commits to your development branch and it'll automatically update the PR. That way you don't have to think worry about rebasing, merging or changing history.
That being said, changing history does have its advantages when you work on a local branch with commits that are not live on master. rebasing, squashing, cherry-picking are great to have as an option.
that being said, you could work with big or small commits, it wouldn't matter because when it's added to the main repo, it'll be generally merged/rebased by the maintainers as a single commit.
I like [this video by David Baumgold on Advanced Git](https://www.youtube.com/watch?v=4EOZvow1mk4). You may or may not find it interesting.

andreicristianpetcu:
Thank you @jbhoosreddy for the video. I actually used all the features in that talk. Including reflog and bisect :)

Issue #1589, opened 2017-01-02
Organize Components Directory


jasonLaster:
We now have 20+ components, so it might be nice to organize them into logical directories:

**Some directories that come to mind are:**

* EditorPane
* shared
* SecondaryPanes

**Editor Pane**
* Editor
* EditorBreakpoint
* EditorSearchBar
* SecondaryPanes
...

**Shared**
* PaneToggleButton
* ObjectInspector
* Rep
* Dropdown
* Autocomplete
* CloseButton

**SecondaryPanes**
* EventListeners
* CommandBar
* Expressions

**Directories**

```
Accordion.css             ChromeScopes.js           Editor.js                 Expressions.css           Rep.js                    SourceSearch.js           WhyPaused.js
Accordion.js              CloseButton.css           EditorBreakpoint.js       Expressions.js            Scopes.css                SourceTabs.css            menu.css
App.css                   CloseButton.js            EditorConditionalPanel.js Frames.css                Scopes.js                 SourceTabs.js             reps.css
App.js                    CommandBar.css            EditorHitMarker.js        Frames.js                 SecondaryPanes.css        Sources.css               test-utils.js
Autocomplete.css          CommandBar.js             EditorSearchBar.css       ObjectInspector.css       SecondaryPanes.js         Sources.js                tests
Autocomplete.js           Dropdown.css              EditorSearchBar.js        ObjectInspector.js        SourceFooter.css          SourcesTree.js            utils
Breakpoints.css           Dropdown.js               EventListeners.css        PaneToggleButton.css      SourceFooter.js           SplitBox.css
Breakpoints.js            Editor.css                EventListeners.js         PaneToggleButton.js       SourceSearch.css          WhyPaused.css
```

clarkbw:
I'm thinking of going forward with the `shared` directory for now.  I think that will reduce things further and we can see if we want to move more of the items after.

I don't have a good feeling for the Panes organization right now so I'd like to punt on that.

So here's what it will look like:

* Components
  * Editor
    * Breakpoint
    * ConditionPanel
    * HitMarker
    * SearchBar
  * shared
    * ManagedTree (from utils)
    * Svg (from utils)
    * Accordion
    * Autocomplete
    * CloseButton
    * PaneToggleButton
    * ObjectInspector
    * Rep

clarkbw:
Not sure about having both the `shared` and `utils` directory.  Might move to all `shared`.

[...]

PR #1660, opened 2017-01-10
Editor component directory

clarkbw:
Associated Issue:  #1589

### Summary of Changes

* Moved all `Editor${Component}` components to `Editor/${Component}`
* moved a couple util functions into `utils/source-editor`

### Test Plan

- [x] Command-P works searches and opens files
- [x] Adding and removing breakpoints work as expected
- [x] Adding and removing conditional breakpoints work as expected
- [x] Editor Context Menus function

### Questions

* The `utils/source-editor` file isn't really a "shared util" as much as its a nice CodeMirror wrapper.  I was thinking of pulling that into the Editor directory and calling it `Editor/CodeMirror` so its purpose is more obvious.
* It would be nice to slim down the Editor component even more, moving the context menu work out into a `Editor/ContextMenu` component was something I wondered about.
* `SourceFooter` and `SourceTabs` remain separate and I still find it confusing.  `EditorFooter` and `EditorTabs` makes a lot more sense to me but I didn't want to stretch too much.  I'd like to add a README file to the Editor directory to explain all the pieces and where the Editor fits into other components.

### Notes

This change was actually pretty simple. However I'm not sure if this layout is exactly what was had in mind.  I wanted to group things by Pane but the Editor is already fairly complex without adding the other tab and footer components in there as well.

* Editor.js
* Editor.css
* Editor/Breakpoint.js
* Editor/ConditionalPanel.js
* Editor/HitMarker.js
* Editor/SearchBar.js
* Editor/SearchBar.css

Since this will break lots of other peoples changes we might want to wait for other PRs to close and then redo this.


jasonLaster:
@clarkbw how'd you know that @jbhoosreddy and I were talking about this today!

@bomsy, @wldcordeiro what do you think of this change? Wellington, it reminds me of `public/js` which feels like forever ago.

jasonLaster:
Looks like something minor like `this.editor.resizeBreakpointGutter` is not defined... I'll take a look tomorrow and help get this in!

src/components/Editor.js
```diff
@@ -42,26 +42,6 @@ function breakpointAtLine(breakpoints, line) {
...
- * numbers gutter. Editor CSS will absolutely position the gutter
- * beneath the line numbers. This makes it easy to be flexible with
- * how we overlay breakpoints.
- */
-function resizeBreakpointGutter(editor) {
-  const gutters = editor.display.gutters;
-  const lineNumbers = gutters.querySelector(".CodeMirror-linenumbers");
-  const breakpoints = gutters.querySelector(".breakpoints");
-  breakpoints.style.width = `${lineNumbers.clientWidth}px`;
-}
```
jasonLaster: extracting these three utilities is the right idea. Unfortunately, moving them to the SourceEditor won't work because we have a different SourceEditor in the panel :/

My vote would be to move these utils to a new util file in a later PR. Infact, i think there are other utilites we can extract as well :)

clarkbw: Oh, interesting.  I should have looked at that more carefully.  I'll pull this part out.

jasonLaster: it's impossible to guess that. we should have a BIG comment, because anyone could miss it

PR #1662, opened 2017-01-10
update local config

jasonLaster:
* Noticed that we were missing some top level TOC items. 
* also the two config items could be merged


jasonLaster:
CC @clarkbw

Issue #1503, opened 2016-12-20
[CommandBar] flash pressed state of command buttons


clarkbw:
Especially with keyboard shortcuts we need to make it more obvious which button in the `CommandBar` was activated.  Some kind of flash animation on the button?

jasonLaster:
Here's a quick patch that might help someone get started. 

It's kinda cool that we can get this effect actually 😉 

```patch
diff --git a/src/components/CommandBar.css b/src/components/CommandBar.css
index ad4f66f..394b5fb 100644
--- a/src/components/CommandBar.css
+++ b/src/components/CommandBar.css
@@ -7,6 +7,10 @@
   padding: 8px 5px;
 }

+.command-bar > span.pressed {
+ transition: /* https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Transitions/Using_CSS_transitions */
+}
+
 .command-bar > span {
   cursor: pointer;
   width: 16px;
diff --git a/src/components/CommandBar.js b/src/components/CommandBar.js
index 3153000..58bbf17 100644
--- a/src/components/CommandBar.js
+++ b/src/components/CommandBar.js
@@ -70,6 +70,16 @@ function debugBtn(onClick, type, className, tooltip) {
   );
 }

+function showPressedAnimation(element) {
+      button.classList.addClass("pressed")
+      setTimeout(() => {
+        button.classList.removeClass("pressed")
+      }, 200)
+
+      // better than timeout would be listening for animationend
+      // https://developer.mozilla.org/en-US/docs/Web/Events/animationend
+}
+
 const CommandBar = React.createClass({
   propTypes: {
     sources: PropTypes.object,
@@ -103,11 +113,17 @@ const CommandBar = React.createClass({
   componentDidMount() {
     const shortcuts = this.context.shortcuts;
     const handleEvent = (e, func) => {
+      // the important thing is we can perhaps do something here
+      // the challenge is to get a reference to the command button
+      // add the class and then remove it in some amount of time
+      const button = e.target;
+      showPressedAnimation(button)
       e.preventDefault();
       e.stopPropagation();
       func();
     };

+  // heres the crazy code where we add command event listeners
     COMMANDS.forEach((action) => shortcuts.on(
```

liorpr:
started to implement this, but got kinda stuck. the buttons are re-rendered so fast(due to the change of the app state and thus the component's props after the action is triggered) that you can't even see the animation.
I'd like to hear your thoughts :)
[...]

PR #1666, opened 2017-01-10
Flash pressed command buttons

liorpr:
Associated Issue: #1503 

### Summary of Changes

1. Moved the debugBtn function into the CommandBar component. The problem was that the event is being triggered by the keyboard, so I couldn't access the button with the event object. so I used the ref callback attribute for saving the button inside the component, and needed the "this" variable.

2. The animation with css was problematic, because the state transitions from pause: null to pause:Map(...) with every button press, so the css class toggling happens so fast you don't get to see it animates! ok well I'm not 100% percent about this but I noticed that although the state changes(and so the component's props being passed down) the span itself does not change! just being updated so I animated the element with javascript and it works.

### DIFF
--- a/src/components/SecondaryPanes/CommandBar.js
+++ b/src/components/SecondaryPanes/CommandBar.js
@@ -1,5 +1,6 @@
 const React = require("react");
 const { DOM: dom, PropTypes } = React;
+const { findDOMNode } = require("react-dom");
 const { connect } = require("react-redux");
 const { bindActionCreators } = require("redux");
 const { getPause, getIsWaitingOnBreak, getShouldPauseOnExceptions,
@@ -62,10 +63,27 @@ function formatKey(action) {
   return formatKeyShortcut(key);
 }
 
+function handlePressAnimation(button) {
+  if (!button) {
+    return;
+  }
+
+  button.style.opacity = "0";
+  button.style.transform = "scale(1.3)";
+  setTimeout(() => {
+    button.style.opacity = "1";
+    button.style.transform = "none";
+  }, 200);
+}
+
 function debugBtn(onClick, type, className, tooltip) {
   className = `${type} ${className}`;
   return dom.span(
-    { onClick, className, key: type },
+    {
+      onClick,
+      className,
+      key: type
+    },
     Svg(type, { title: tooltip })
   );
 }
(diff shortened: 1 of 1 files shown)


src/components/CommandBar.js
```diff
@@ -102,37 +106,49 @@ const CommandBar = React.createClass({
 
   componentDidMount() {
     const shortcuts = this.context.shortcuts;
-    const handleEvent = (e, func) => {
+    const handlePressAnimation = button => {
```
jasonLaster: this could be defined outside of the component

src/components/CommandBar.js
```diff
@@ -102,37 +106,49 @@ const CommandBar = React.createClass({
...
-      debugBtn(this.props.stepOver, "stepOver", className,
+      this.debugBtn(this.props.stepOver, "stepOver", className,
         L10N.getFormatStr("stepOverTooltip", formatKey("stepOver"))
       ),
-      debugBtn(this.props.stepIn, "stepIn", className,
+      this.debugBtn(this.props.stepIn, "stepIn", className,
         L10N.getFormatStr("stepInTooltip", formatKey("stepIn"))
       ),
-      debugBtn(this.props.stepOut, "stepOut", className,
+      this.debugBtn(this.props.stepOut, "stepOut", className,
```
jasonLaster: how about we pass in a `getButton` function that wraps the ref logic?

src/components/CommandBar.js
```diff
@@ -102,37 +106,49 @@ const CommandBar = React.createClass({
...
       e.preventDefault();
       e.stopPropagation();
+      handlePressAnimation(button);
       func();
     };
 
     COMMANDS.forEach((action) => shortcuts.on(
       getKey(action),
-      (_, e) => handleEvent(e, this.props[action]))
+      (_, e) => handleEvent(e, this.props[action], this[action]))
```
jasonLaster: what does ` this[action]` do?

liorpr: I save a ref for each button on the component instance, so if an action is let's say 'stepOver',
this['stepOver'] is the corresponding button

src/components/CommandBar.js
liorpr: something like this on the instance?
  saveButtonRef(type) {
    this[type] = button;
  }

@jasonLaster

liorpr: updated my PR, I think I liked it more the old way :)

jasonLaster:
@liorpr just updated after a rebase with master. It's looking pretty good, but i want to keep testing tomorrow :)

jasonLaster:
Sorry for the slow reply @liorpr could we get it so only the button activated flashes?

@clarkbw just confirming that's the behavior we're looking for

clarkbw:
> @clarkbw just confirming that's the behavior we're looking for

Looking good with your suggested change.  And it looks like from the code that this will also activate on key shortcuts as well, right? 👍

liorpr:
Actually the gif does not reflect my changes, the flash is caused by the classes switching and so the buttons opacity. I added code to the handleEvent function that's being run inside the (keyboard) Commands event registration(figured it's ok because if you mouse-click a button, you know it's being clicked, no need for flashing)

jasonLaster:
Oh, sorry @liorpr you're totally right

src/components/CommandBar.js
```diff
@@ -62,10 +62,26 @@ function formatKey(action) {
...
+
+function debugBtn(onClick, type, className, tooltip, getRefCallback) {
   className = `${type} ${className}`;
   return dom.span(
-    { onClick, className, key: type },
+    {
+      onClick,
+      className,
+      key: type,
+      ref: getRefCallback(type)
```
jasonLaster: hmm, is it possible to do something like

```js
ref: type
```

and then we can do `this.ref[action]` to get the button in `handleEvent`. 

I would like to avoid adding the buttons to the prototype if we can

liorpr: you mean pass the type as a string right?
I can do it but the react team recommends the callback attribute approach

screen shot 2017-01-15 at 7 41 51 pm

jasonLaster:
Oh yea that's a good point. Perhaps we can use the callback to find the
button with a selector!

Let's see where else we do this. Look for find node

jasonLaster:
Hey @liorpr, I just made some tweaks to drop the ref usage and instead use a selector. Let me know what you think

liorpr:
nice, looks clearer. I guess you can also drop the getRefCallback arg on the debugBtn function
@jasonLaster

jasonLaster:
@liorpr good call. Mind cleaning it up?

liorpr:
did the cleanup @jasonLaster

jasonLaster:
@liorpr just rebased and squashed. should be good to go in a sec

PR #1667, opened 2017-01-10
(Discussion) Prettier

jasonLaster:
@jlongster just published prettier. Prettier is a JS pretty printer, that promises to ease some of the friction associated with code style.

Here's a patch that shows what our styles would look like:

* Let's discuss if we have any disagreements
* what this means for eslint
* how we would setup prettier for contributors


jbhoosreddy:
In my opinion, it follows a style guide different from the one we follow, or I'm used to following. Which is not a bad thing.. If it brings consistency, it could help saving time like style nits and such. Especially for new contributors. Consistency would be a great plus!

src/actions/breakpoints.js
```diff
@@ -12,17 +12,14 @@ const constants = require("../constants");
...
-  getOriginalLocation, getGeneratedLocation, isOriginalId
-} = require("../utils/source-map");
+const { getOriginalLocation, getGeneratedLocation, isOriginalId } = require(
+  "../utils/source-map"
+);
 
-import type { ThunkArgs } from "./types";
-import type { Location } from "../types";
+import type {ThunkArgs} from "./types";
+import type {Location} from "../types";
```
jasonLaster: I'm surprised there are no spaces here, but there are spaces with objects.

src/actions/breakpoints.js
```diff
@@ -54,7 +51,8 @@ function enableBreakpoint(location: Location) {
  */
 function addBreakpoint(
   location: Location,
-  { condition, getTextForLine } : addBreakpointOptions = {}) {
+  { condition, getTextForLine }: addBreakpointOptions = {}
+) {
```
jasonLaster: this is beautiful 💯

src/actions/coverage.js
```diff
@@ -1,18 +1,13 @@
...
 
 function recordCoverage() {
   return async function({ dispatch, getState, client }: ThunkArgs) {
     const { coverage } = await client.recordCoverage();
 
-    return dispatch({
-      type: constants.RECORD_COVERAGE,
-      value: { coverage }
-    });
+    return dispatch({ type: constants.RECORD_COVERAGE, value: { coverage } });
```
jasonLaster: I like collapsing when we can :)

clarkbw:
I'm a fan of it.

I think a pre-commit hook would work well.  That would allow anyone to essentially ignore styling concerns completely but the end result is still styled.  Of course it might surprise people when they see their PR / final output.

We can also recommend the atom plugin as well for regular contributors.

src/actions/breakpoints.js
jlongster: That is fixed on master. Not sure if it was included in the latest published version or not.

jlongster:
> Should we have a pre-commit hook?

I haven't had time to figure out the layer which makes sure that PRs have been appropriately formatted. I think the CLI is going to provide the functionality for editing files in-place and also check if the passed in files are printed. It's up to you when to run this functionality.

I don't think a pre-commit hook will work, because running the printer across an entire project is just too slow. I wouldn't want to slow down making commits, but the CI should fail if things aren't appropriately printed. Or even better, the CI could potentially automatically print the code. I'm not sure if that would work, but would be cool if so.

> Should we run this on pre-push and if there are changes encourage contributors to commit them?

As I said I think it's just too slow. Maybe you could only run it on files that have changed, I don't know. I think ideally it would just be auto-formatted for them somehow, since that's possible. It's just a question of when and how to do it.

jasonLaster:
closing for now. but we can revisit

PR #1678, opened 2017-01-10
Update panel

jasonLaster:
When we upgraded the launchpad we started exporting bundles with the bundle name `debugger.js` which were not being tested. Therefore, we allowed some regressions. This PR fixes that and there will be another bugzilla patch for using debugger.js.

### DIFF
--- a/src/components/Editor.js
+++ b/src/components/Editor.js
@@ -490,7 +490,7 @@ const Editor = React.createClass({
     this.editor.destroy();
     this.editor = null;
 
-    const searchAgainKey = L10N.getStr("sourceSearch.search.again.key");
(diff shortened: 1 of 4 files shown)


src/components/Editor.js
```diff
@@ -490,7 +490,7 @@ const Editor = React.createClass({
     this.editor.destroy();
     this.editor = null;
 
-    const searchAgainKey = L10N.getStr("sourceSearch.search.again.key");
+    const searchAgainKey = L10N.getStr("sourceSearch.search.next.key");
```
clarkbw: There are two instances of this.  We need to change the other one as well.

https://github.com/jasonLaster/debugger.html/blob/387bc86e4302c99feeda3988915df5dccf313ceb/src/components/Editor.js#L475

PR #1691, opened 2017-01-12
Update Mocha runner

jasonLaster:
This fixes the mocha web runner for unit tests. I needed this while working on @arthur801031's persist tabs PR and it really helped.

#### What changed
* I switched from having a separate server that was started with `yarn mocha-server` to always running the mocha endpoint at `/mocha`. The big win is that now the server is on the same host, which makes it easier to request assets like debugger.js or an example bundle file
* I removed the component tests because we were no longer using them


bin/mocha-server.js
```diff
@@ -4,9 +4,9 @@
 const path = require("path");
 const webpack = require("webpack");
 const express = require("express");
-const projectConfig = require("../webpack.config");
 const webpackDevMiddleware = require("webpack-dev-middleware");
 const fs = require("fs");
+var serveStatic = require('serve-static')
```
juliandescottes: Is this used anywhere?

src/test/node-unit-tests.js
```diff
@@ -39,13 +37,23 @@ const webpack = require("webpack");
...
 
-// The source map worker is compiled with webpack (and mock-require
-// doesn't work in workers) so mock it with an alias, and tweak a few
-// things to make the stub fetcher work in node.
-webpackConfig.resolve.alias["devtools-network-request"] =
-  "devtools-network-request/stubNetworkRequest";
+function stubNetworkRequest(url) {
+  return new Promise((resolve, reject) => {
+    const fs = require("fs");
+    path = require("path");
```
juliandescottes: When passed as a string and executed, this will pollute the global scope with `path`. Use `const _path` with a small comment and restore the `const path` for the require in the beginning of the file?

juliandescottes:
Oh and I forgot to mention in the review, it looks like some tests were broken by #1547 , we'll need to take care of that too.

juliandescottes:
Damn I broke CI :(

I didn't realize the unit tests would be also failing outside of Mocha. I'll back this out.

Issue #1506, opened 2016-12-20
[RTL] Update the breakpoints component


jasonLaster:
It would be great if the the breakpoints component were more RTL friendly

Here are a couple things that I noticed:
* the green border should be on the right
* the left padding should be on the right so that items line up well

Getting started
* here are some [rtl](https://github.com/devtools-html/debugger.html/blob/master/docs/local-development.md#rtl) docs.
* we use postcss bi-direction for some magic CSS RTL clauses

![screen shot 2016-12-20 at 5 23 14 pm](https://cloud.githubusercontent.com/assets/254562/21370312/0ec6b88a-c6d9-11e6-8ae9-b5a05a127feb.png)

CC @liorpr

clarkbw:
I don't think anyone is currently working on this. Setting back to `up for grabs`

irfanhudda:
I would like to work on this

jasonLaster:
Thanks @irfanhudda

Issue #1665, opened 2017-01-10
Command Bar redesign


jasonLaster:
I'm adding this issue to track a discussion that @Garbee and I had this morning.

It would be nice to think through how we present the pause actions / message in the command bar.

**Goals:**
* make the ux more clear
* easier for new comers
* improve accessibility

**Insights:**
* we can move the *other* breakpoints "pause on next execution" "pause on exceptions" to the breakpoints pane. Infact VS Code beat us to it.
* we can hide the "paused" actions and only show them when we're paused.

Here's one UI that takes advantage of this. I think @Garbee will likely have an alternative that avoids the dropdown and is more clear.

![command bar redesign](https://cloud.githubusercontent.com/assets/254562/21817246/401606ee-d731-11e6-8e2b-cecb29d3f1dd.jpg)

krishnagopinath:
I like how the "Paused at a Breakpoint" section is placed. Looks clean, doesn't waste real estate and just does one thing - conveys the message loud and clear :) 

This gives way for using other color tones for the text, like when there's an error, etc.

jasonLaster:
Yea, it's worth noting that if we're paused at a breakpoint, then the breakpoint also gets styled.

> color tones for the text, like when there's an error

👍

clarkbw:
👍 

Only thing I'd like to iterate on is the button that brings up the "Pause On" dialog.  The plus button doesn't convey enough information.  Perhaps just creating a button with the text "Pause On..."?

nt1m:
@jasonLaster I'd prefer making "Pause on next execution" a label of the toolbar button that brings up the dialog, it saves some space, and makes sense in this context.

Something like:
```
_______________________________________________________
- [x]                                  ||  on next execution ▾  | + |
¯¯¯¯¯¯¯¯¯¯¯¯¯¯¯¯¯¯¯¯¯¯¯¯¯¯¯¯¯¯¯¯¯¯¯¯¯¯¯¯¯¯¯¯¯¯¯¯
```

Garbee:
In my initial design these exception triggers are all radios in the main breakpoint list. The idea being dropdowns/dialogs/etc. all obscure the functionality away. Some developers may not realize this functionality exists except by accident if it is tucked away behind a dropdown/dialog/multi-trigger button.

I'll share the look I am going for tomorrow when I have a bit more of the functionality moved around so the fuller idea is there. Once the main move is done changing the design from there to whatever is desired will be fairly seamless.

[...]

PR #1692, opened 2017-01-12
RTL friendly styles for breakpoints component

irfanhudda:
Associated Issue: #1506 

### Summary of Changes
Changed some styles in `Editor/Breakpoints.css` to make it more RTL friendly

* [X] the green border should be on the right
* [X] the left padding should be on the right so that items line up well
* [X] breakpoint close button should be on left side

#### Before
screenshot from 2017-01-13 00-12-38

#### After
screenshot from 2017-01-13 00-12-03

### DIFF
--- a/src/components/Breakpoints.css
+++ b/src/components/Breakpoints.css
@@ -9,19 +9,30 @@
   padding: 0.5em 1px;
   line-height: 1em;
   position: relative;
-  border-left: 4px solid transparent;
   transition: all 0.25s ease;
+}
+
+html[dir="rtl"] .breakpoints-list .breakpoint {
+  border-right: 4px solid transparent;
+}
+
+html:not([dir="rtl"]) .breakpoints-list .breakpoint {
+  border-left: 4px solid transparent;
 }
 
 .breakpoints-list .breakpoint:last-of-type {
   padding-bottom: 0.45em;
 }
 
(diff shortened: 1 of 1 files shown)


irfanhudda:
I think that the checkbox to toggle breakpoints should also be aligned. Let me know if you have some thoughts about this or other suggestions.

jasonLaster:
Yeah, the checkboxes are a shame. We're doing some thinking here #1665. What you have seems like a step in the right direction though.

src/components/Breakpoints.css
```diff
@@ -9,19 +9,30 @@
...
 
-.breakpoints-list .breakpoint.is-conditional {
+html:not([dir="rtl"]) .breakpoints-list .breakpoint.is-conditional {
   border-left-color: var(--theme-graphs-yellow);
 }
 
-.breakpoints-list .breakpoint.paused {
+html[dir="rtl"] .breakpoints-list .breakpoint.is-conditional {
+  border-right-color: var(--theme-graphs-yellow);
+}
```
jasonLaster: nice.

jasonLaster:
screen shot 2017-01-12 at 2 43 14 pm

jasonLaster:
Thanks @irfanhudda!

PR #1702, opened 2017-01-13
update codemirror styles

jasonLaster:
Pulls in changes from https://bugzilla.mozilla.org/show_bug.cgi?id=1320053

Test Plan:
- [x] try it locally

CC @juliandescottes


wldcordeiro:
Tried this and it's looking good. 👍

PR #1707, opened 2017-01-14
add shouldComponentUpdate to default accordions

clarkbw:
Breakpoints, Expressions, Frames, and Scopes were all re-rendering based on changes unrelated to their responsibilities.

### Summary of Changes

You can reproduce this easily right now.  Set a breakpoint on the render method for any of the Accordions in the secondary pane.  Then expand or collapse any of the other accordions and you'll hit that breakpoint.

* Added `shouldComponentUpdate` to each component so it only renders as needed per component

### Test Plan

Example test plan:

- [x] Added and removed breakpoints
- [x] Paused the debugger to check that accordions worked correctly
- [x] Navigated frames when paused

### DIFF
--- a/src/components/Breakpoints.js
+++ b/src/components/Breakpoints.js
@@ -49,6 +49,11 @@ const Breakpoints = React.createClass({
 
   displayName: "Breakpoints",
 
+  shouldComponentUpdate(nextProps, nextState) {
+    const { breakpoints } = this.props;
+    return breakpoints !== nextProps.breakpoints;
+  },
+
   handleCheckbox(breakpoint) {
     if (breakpoint.loading) {
       return;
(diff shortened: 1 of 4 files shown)


jasonLaster:
Hmm. This is great. Perhaps fixes the debugger bug where it's slow with big objects like window...

I'd like to move forward with this, but also see if we can get react-redux to solve this on the framework level later.

clarkbw:
> I'd like to move forward with this, but also see if we can get react-redux to solve this on the framework level later.

I believe using [PureComponent](https://facebook.github.io/react/docs/react-api.html#react.purecomponent) is the cleaner route.  It will automatically do the simple compares that I'm doing.  `shouldComponentUpdate` is, as far as I know, really designed for doing some deeper compare that could speed things up.

clarkbw:
But I am seeing some strange paint issues happening with the `ObjectInspector` but I haven't had much time to investigate that more.

PR #1712, opened 2017-01-14
Garbee/breakpoints

Garbee:
Associated Issue: #1665 

### Summary of Changes

Sorry, pretty massive changes in this one. Plenty to discuss.

#### UX Side
* Moved "Pause on Next Exception" toggle and the exception breakpoints under the breakpoints header to get them off the CommandBar itself.
* That change allows for the CommandBar to be rendered only when paused, where most of the actions are needed.
* Pause on next execution is a button trigger.
* Exceptions are put under a collapsible node as distinct radios. This provides better clarity to developers as to what they are switching between and reduces the clicks to get to different modes. Upon selecting a different mode, the collapse header updates to show the current mode so even when collapsed you know which mode you're in.
* Toggle all breakpoints button moved from collapse header into a toggle area of the breakpoint list itself.
* The user breakpoints header also adds a "Remove all" button to quickly remove all of a users custom breakpoints if desired.
    * The remove all button triggers a *confirm* dialog. We should figure out the best way to handle this in our UX. This confirm is added due to having heard some developers in the past with Chrome DevTools having *loads* of breakpoints and accidentally losing them at different points. We can either temporarily store a copy and provide an "undo" action for a few seconds or do an inline confirm UX. If we can have people lose a lot of work easily, we should try to let them undo it or force them to confirm the action.

#### Technical side (JSX)

With this I experimented with using JSX for the components. All that was needed to let us do this is adding the right babel transformer to our list.

##### Pros

* Nesting is easier to follow since we aren't just providing a parameter list.
* Output expectations are very clear.
* JSX is what most React developers use since it is the default. So it is more natural for fresh contributors to get help with if they need it.

##### Cons

* Code style can seem fugly.
    * Braces for using variables in rendering the component.
    * Inline anonymous functions look weird. Moved them to defined constants.

### Test Plan

Fairly standard testing run-through with the exceptions page and toggling a few actions on the todomvc demo. Just to verify the functionality works as it used to. Most changes here as far as the frontend are on the UX side over the technical side.

### Screenshots/Videos

selection_012

selection_013


jasonLaster:
Thanks @Garbee!

@clarkbw, @wldcordeiro, @bomsy, John and I discussed using this branch to discuss some research questions:

1. do we like the JSX style (what questions do we have there)
2. do we like some of these UX changes (what questions do we have there) 

* With the UX changes, I'd like to ideally test the change with UR and AB (node electron app)
* With JSX, I'd like to make a decision after we cut the new release and come up with a plan for how we'll incrementally upgrade with a style doc that can evolve.

src/components/Breakpoints.js
```diff
@@ -72,54 +94,173 @@ const Breakpoints = React.createClass({
...
+  renderPauseExecutionButton() {
+    const { breakOnNext, isWaitingOnBreak } = this.props;
+
+    return (
+      <button
+        className="pause-execution"
+        disabled={ isWaitingOnBreak }
+        onClick={ breakOnNext }
+      >
+        <Svg name="pause" />
```
jasonLaster: 👍

src/components/Breakpoints.js
```diff
@@ -72,54 +94,173 @@ const Breakpoints = React.createClass({
...
+        { isWaitingOnBreak ?
```
jasonLaster: maybe this is pulled up to a variable... Also we should have L10N strings eventually here

src/components/Breakpoints.js
```diff
@@ -72,54 +94,173 @@ const Breakpoints = React.createClass({
...
+  },
+
+  renderGlobalBreakpoints() {
+    const currentMode = this.props.currentExceptionPauseMode;
+    const _createToggle = (fromMode) => {
+      return (
+        <label className="breakpoint" key={ fromMode.mode }>
+          <input
+            type="radio"
+            onChange={ this.pauseExceptionModeToggled }
```
jasonLaster: do people use spaces here? `={`

src/components/Breakpoints.js
```diff
@@ -72,54 +94,173 @@ const Breakpoints = React.createClass({
...
+          />
+          <div className="breakpoint-label">{ fromMode.label }</div>
+        </label>
+      );
+    };
+
+    return (
+      <details>
+        <summary className="_header">
+          { `Exceptions - Pausing on: ${currentMode.headerLabel}` }
```
jasonLaster: eventually, this would move to L10N

src/components/Breakpoints.js
```diff
@@ -72,54 +94,173 @@ const Breakpoints = React.createClass({
...
+          </summary>
+        { this.props.exceptionPauseModes.map(_createToggle) }
```
jasonLaster: I think this is nice, but I'm still a little reluctant about curlies :)

src/components/Breakpoints.js
```diff
@@ -72,54 +94,173 @@ const Breakpoints = React.createClass({
...
+      }
+    };
+    const toggleAll = () => {
+      toggleAllBreakpoints(!breakpointsDisabled);
+    };
+    const toggleIndeterminate = (input) => {
+      if (input) {
+        input.indeterminate = isIndeterminate;
+      }
+    };
```
jasonLaster: I like pulling these functions out of the curlies. I could see the markup becoming pretty hard to follow with these.

Personally, I prefer the old JS style for these in line functions, but ... what can you do :)

wldcordeiro:
Minor comment: If we start going JSX and this lands let's open a tracking issue for JSX converting the rest of the codebase.

src/components/Breakpoints.js
Garbee: Yup. L10N needs to get sprinkled around. I just left it out for the initial review to keep what we are looking at more concise.

src/components/Breakpoints.js
Garbee: Typically not from what I see around in the docs and other projects. When the event handlers were inlined those had them for ease of reading so I put theme everywhere for consistency. Now that those handlers are seeming like we're going to keep them in variables, the spaces here are useless.

src/components/Breakpoints.js
Garbee: Curlies are a tradeoff for overall clarity in the nesting.

src/components/Breakpoints.js
Garbee: I don't think we have a good alternative to add refs/event handlers any other way. :/ So, this is just the way things are with JSX if we go this route.

PR #1713, opened 2017-01-15
Collapsible End Pane in Vertical Layout

wldcordeiro:
Associated Issue: #1537 

**Requires devtools-html/devtools-core#88 to land to actually work properly.**

### Summary of Changes

* add the pane collapse button to the source footer and wire up logic to toggle end pane collapse

### Test Plan

Example test plan:

- [x] Open Debugger in vertical mode
- [x] Toggle collapse button for bottom
- [x] See the pane has collapsed/uncollapsed

screenshot_20170114_205621
screenshot_20170114_205630

@jasonLaster One of the only issues I have with this approach is that the collapse button isn't visible until you open a file. I may instead render it as part of the "editorPane" in `App.js` so that it's always visible. @

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
   },
   "dependencies": {
     "codemirror": "^5.1.0",
-    "devtools-launchpad": "0.0.23",
+    "devtools-launchpad": "0.0.24",
     "devtools-reps": "^0.0.2",
     "documentation": "^4.0.0-beta11",
     "eslint": "^3.12.0",
--- a/src/components/App.css
+++ b/src/components/App.css
@@ -76,20 +76,3 @@ body {
   margin-top: 25px;
   margin-right: 20px;
 }
-
-.welcomebox {
-  width: calc(100% - 1px);
-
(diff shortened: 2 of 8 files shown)


jasonLaster:
oh shit - that's a really interesting problem. 

Perhaps we want to render the footer in App.js so that we can always show the footer... We do this for the SourceTabs as well.

jasonLaster:
Thanks @wldcordeiro!  

Your work looks really good. I took a stab at adding the end toggle button to the welcome box as well.

wldcordeiro:
This looks good with the changes @jasonLaster I'm going to bring it in. 👍

package.json
```diff
@@ -35,7 +35,7 @@
   },
   "dependencies": {
     "codemirror": "^5.1.0",
-    "devtools-launchpad": "0.0.23",
+    "devtools-launchpad": "0.0.24",
```
clarkbw: We need some kind of yarn-lint that checks we've updated the `yarn.lock` file whenever the `package.json` updates.

jasonLaster: Good catch

Issue #1704, opened 2017-01-14
UI Polish


jasonLaster:
Three small things 

- [x] we're missing the bottom border for the command bar. not sure when it went away
- [x] when the sidebars (left and right) are closed you see a horizontal scrollbar
- [x] this might be a bit contentious, but I don't think the "Sources" header buys us anything. It seems a bit like wasted space :) 

Command Bar:

![screen shot 2017-01-13 at 11 32 14 pm](https://cloud.githubusercontent.com/assets/254562/21952384/8dc221d4-d9e8-11e6-9857-f879b1aa3244.png)

CC @clarkbw what do you think about "Sources" I could easily be persuaded to keep it :)

arthur801031:
@jasonLaster  I'll take a stab at this :) 

Is `App.js` tab's bottom border the bottom border for the command bar?

clarkbw:
> what do you think about "Sources" I could easily be persuaded to keep it

I'd argue to keep it.  It maintains a nice, consistent bar across the top.  And allows us to always show the key shortcut for search.  I know it gains us some vertical space but I think I would need to be convinced that we are lacking vertical space in that tree view.

I could be persuaded that we could do something better than just say "Sources" as that seems fairly useless.  I imagined an "iframe selector" would go there eventually.

jasonLaster:
Oh yea. I like the Rob bar with the label. I'm just thinking that we don't
need the word sources. The label and context helps.

jasonLaster:
@clarkbw if you think we should keep the Sources header, I can bring it back

clarkbw:
> @clarkbw if you think we should keep the Sources header, I can bring it back

Nope. 😄

PR #1715, opened 2017-01-15
Ui polish

jasonLaster:
Associated Issue: #1704

### Summary of Changes

* fixes the command bar border
* removes the sources
* enables a bunch of features

### DIFF
--- a/src/components/Expressions.css
+++ b/src/components/Expressions.css
@@ -6,6 +6,7 @@
   cursor: pointer;
   color: var(--theme-body-color);
   background-color: var(--theme-body-background);
+  line-height: 12px;
 }
 
 .input-expression::-webkit-input-placeholder {
--- a/src/components/Sources.css
+++ b/src/components/Sources.css
@@ -19,9 +19,9 @@
   font-size: 1.2em;
   display: flex;
   align-items: baseline;
-  justify-content: space-between;
   -moz-user-select: none;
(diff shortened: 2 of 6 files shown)


src/components/Expressions.css
```diff
@@ -6,6 +6,7 @@
   cursor: pointer;
   color: var(--theme-body-color);
   background-color: var(--theme-body-background);
+  line-height: 12px;
```
jasonLaster: this is a small change, that fixes a small `.67px` adjustment in vertical mode

src/components/Sources.css
```diff
@@ -19,9 +19,9 @@
   font-size: 1.2em;
   display: flex;
   align-items: baseline;
-  justify-content: space-between;
   -moz-user-select: none;
   user-select: none;
+  justify-content: flex-end;
```
jasonLaster: moves the label to the right without the sources header

src/components/Sources.js
```diff
@@ -39,7 +39,6 @@ const Sources = React.createClass({
     return dom.div(
       { className: "sources-panel" },
       dom.div({ className: "sources-header" },
-        L10N.getStr("sources.header"),
```
clarkbw: We should probably drop this from the strings.json etc. as well unless it is being used somewhere else.

Issue #1717, opened 2017-01-15
[Sidebar] there's a scrollbar when collapsed


jasonLaster:
when the sidebars (left and right) are closed you see a horizontal scrollbar.

![](http://g.recordit.co/itqNx33T2K.gif)

PR #1721, opened 2017-01-16
UI Polish

arthur801031:
Associated Issue: #1704, #1717

### Summary of Changes

* Fix when the sidebars (left and right) are closed you see a horizontal scrollbar

### DIFF
--- a/src/components/App.js
+++ b/src/components/App.js
@@ -86,6 +86,11 @@ const App = React.createClass({
     const { sources, startPanelCollapsed, endPanelCollapsed } = this.props;
     const { horizontal } = this.state;
 
+    let horizontalOverflow = "auto";
+    if (startPanelCollapsed && endPanelCollapsed) {
+      horizontalOverflow = "hidden";
+    }
+
     return dom.div(
       { className: "debugger" },
       SplitBox({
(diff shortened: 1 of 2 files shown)


arthur801031:
@jasonLaster  

Please check. Thank you! :)

jasonLaster:
Hmm. I'm not sure if we want to always overflow hidden. We do want scrollbars with wise elements

arthur801031:
So we only want to set `overflow-x: hidden;` to ` .split-box.vert` when the sidebar(s) are closed?

What are wise elements?

jasonLaster:
Sorry meant wide element like the source tree is deeply expanded.

Yea, I think that's right

wldcordeiro:
Overflow auto might be what you want so the scrollbars appear when needed?

jasonLaster:
Hmm, @wldcordeiro looks like `oveflow:auto` is already in affect. Actually this might be a case of the splitter still being 1px and the width being 100% or flex:1

arthur801031:
Hi @jasonLaster 

I tried setting splitterSize to 0, but it didn't work, so I used this approach instead :)

Issue #1085, opened 2016-11-02
[Dark Theme] secondary text


jasonLaster:
The secondary text is an awkward gray that looks off to me. 

Here is how you change a [theme color](https://github.com/devtools-html/debugger.html/blob/master/docs/local-development.md#themes)

Here is the secondary text for frames, we could also find other secondary text in breakpoints...

```patch
diff --git a/src/components/Frames.css b/src/components/Frames.css
index 34c3979..c442b66 100644
--- a/src/components/Frames.css
+++ b/src/components/Frames.css
@@ -27,6 +27,7 @@
   background-color: var(--theme-tab-toolbar-background);
 }

+/* => here */
 .frames .location {
   color: var(--theme-comment);
   font-weight: lighter;
```

![screen shot 2016-11-02 at 3 28 11 pm](https://cloud.githubusercontent.com/assets/254562/19944140/eb205082-a110-11e6-9a64-325bd8e93d7b.png)

jasonLaster:
@wldcordeiro @helenvholmes I deleted the dark theme label as we no longer have that many issues

jasonLaster:
ping @helenvholmes do you have time to tackle this?

andreicristianpetcu:
I'll work on this

jasonLaster:
Grab it. Nah

andreicristianpetcu:
@jasonLaster I read a lot about Enzyme, Jest and testing React+Redux stuff. I think I can write some tests on my next contribution. I still have Mochitest on my reading list.

jasonLaster:
@andreicristianpetcu nice. yeah, mochitest is our  mozilla test framework. The way we use it, its pretty basic. 

We currently have node unit tests and we're working on getting the mochitests to run in a regular pae

PR #1725, opened 2017-01-16
#1085 fixed dark theme secondary color

andreicristianpetcu:
Associated Issue: #1085

### Summary of Changes

* Made dark theme secondary text look more alive in the dark theme

### Test Plan

Example test plan:

- [x] Opened todo app and put some breakpoint in app.js. Verified the dark theme so it does not have the grey color in the call stack (frames) and in the breakpoints area. 
- [x] Checked light theme is unchanged
- [x] Checked firebug theme is unchanged
- [x] Checked all of the above in Chrommium

Before
before
After (made it use the [theme-highlight-bluegrey](https://github.com/devtools-html/debugger.html/search?utf8=%E2%9C%93&q=bluegrey&type=Code) color)
after-theme-highlight-bluegrey
Firebug - unchanged
firebug
Light theme - unchanged
light

### DIFF
--- a/src/components/Breakpoints.css
+++ b/src/components/Breakpoints.css
@@ -66,8 +66,17 @@ html .breakpoints-list .breakpoint.paused {
   order: 3;
 }
 
-.breakpoint-snippet {
+:root.theme-light .breakpoint-snippet,
+:root.theme-firebug .breakpoint-snippet {
   color: var(--theme-comment);
+}
+
+:root.theme-dark .breakpoint-snippet {
+  color: var(--theme-body-color);
+  opacity: 0.6;
+}
+
+.breakpoint-snippet {
   padding-inline-start: 18px;
 }
 
--- a/src/components/Frames.css
+++ b/src/components/Frames.css
@@ -28,8 +28,17 @@
 }
 
 .frames .location {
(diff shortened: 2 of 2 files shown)


andreicristianpetcu:
I played a little bit with the colors before picking this one :D Here are the ones I did not like :P

Blue
theme-highlight-blue

Green
theme-highlight-green

Red
theme-highlight-red

Purple
theme-highlight-purple


I tested only existing theme colors.

jasonLaster:
Hmm, I'm not sure if i like the *color* secondary themes. Perhaps we use the "watch expressions label" color... Or something that is slightly softer than the call stack function label.

andreicristianpetcu:
@jasonLaster The watch expressions has the color --theme-body-color. I will use that color.

andreicristianpetcu:
Apparently the color is --theme-body-color but it has an opacity. I updated my PR.
theme-body-color-with-opacity

jasonLaster:
Thanks, I think this looks good.

I'd like to re-evaluate some of our variables as its kinda weird to use `--comment` or `--splitter` in lots of places :) 

I'm also tempted to move this to a variable, but I'm not sure if that's appropriate either

Issue #1693, opened 2017-01-12
Publish a new version to MC


jasonLaster:
Here are the rough steps:

1. publish a new version to NPM. publishing, copies the files to a dist directory for npm.
2. check for changes in m-c
3. bump NPM version in MC and copy files to the correct locations.

Context:
* the *devtools* directory in MC will have a package.json, with the current versions

nt1m:
It would be cool if this could be done more often, so Nightly users are able to get the changes and report bugs more quickly. This would make debugger-html better in terms of stability since a lot of our Nightly users are great bug filers :)

PR #1730, opened 2017-01-17
Update Version

jasonLaster:
Associated Issue: #1693
### Summary of Changes

* bumps version
* also updates lodash usage so we don't pull in the entire lib

Here's why lodash matters :)
detail...

<details>
<summary>before/after</summary>

### new
screen shot 2017-01-17 at 12 04 13 pm

### old
screen shot 2017-01-17 at 12 04 08 pm

</details>

### DIFF
--- a/assets/locales/debugger.properties
+++ b/assets/locales/debugger.properties
@@ -124,9 +124,9 @@ sourceSearch.search.key=F
 # the search for re-searching the same search triggered from a sourceSearch
 sourceSearch.search.again.key=G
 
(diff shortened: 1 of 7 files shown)


wldcordeiro:
Bundle size comparison? 😄 In Kb/Mb or whatever? That's awesome to reduce our footprint, makes me think we could benefit from testing Webpack 2.x for its tree-shaking too.

jasonLaster:
Added the issue in core

clarkbw:
Just noticed this lodash: https://github.com/devtools-html/debugger.html/blob/master/src/components/Editor/SearchBar.js#L7

Issue #1640, opened 2017-01-06
Fix React.render and React.findDOMNode warnings


jasonLaster:
We've been getting these ReactDOM.render warnings for awhile, but haven't had time to trace it down. It would be really appreciated if someone could find where we're doing this.

The best way to approach this would probably be to grep for `findDomNode` and `render` and see if you can find it that way. 

The other way would be to put breakpoints in these functions

![screen shot 2017-01-06 at 11 07 43 am](https://cloud.githubusercontent.com/assets/254562/21723957/68be691c-d400-11e6-9ff8-4c835b0d2eb5.png)

clarkbw:
Here's a quick video on how to break on these errors and find the root cause... 
![use-callstack](https://cloud.githubusercontent.com/assets/2134/21728620/110f0310-d3fc-11e6-970c-7ffcff5adfcf.gif)

* Click location in the console
* Break on that error
* Use the call stack to go up the stack until you find code that looks like the problem

liorpr:
started to look into this, here's what I got so far:

1. the deprecation mechanism works per function for the whole app, so it's not about where its happening, but why(tried to remove the mentioned render call and got this warning from somewhere else)
2. don't know why but it seems like the app does not require the correct react-dom package, but the react one, that's why the render method is considered deprecated. can't figure out how to correct this, maybe someone have an idea? didn't see explicit definition of react-dom package in package.json

wldcordeiro:
@liorpr could it be because of us doing this in a lot of our components?

```js
const React = require("react");
const { DOM: dom, PropTypes } = React;
```

Shouldn't `dom` come from the `react-dom` lib and not `react`?

jasonLaster:
nice idea @wldcordeiro we should definitely fix these!

liorpr:
yeah that's definitely shouldn't be there, but even when I break on EditorBreakpoint.js(like in the image above) it happens, and it requires("react-dom")

clarkbw:
> Shouldn't dom come from the react-dom lib and not react?

I don't think [ReactDOM](https://facebook.github.io/react/docs/react-dom.html) provides a `DOM` object.  I believe that is supposed to come from React itself, though it is somewhat undocumented, you'll find it in their [examples](https://github.com/facebook/react/blob/master/examples/basic/index.html#L36)

irfanhudda:
[...]

PR #1731, opened 2017-01-17
fixing issue where react = react-dom

amitmtrn:
Associated Issue: #1640 

### Summary of Changes

* .babelrc webpack.json - adding react-dom file to the resolve which will prevent require issue in webpack

### Test Plan

running the debugger.html and check the log for exceptions

Feel free to improve it!

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,5 +1,6 @@
-{ "plugins": [
-  "transform-es2015-block-scoping",
+{
+  "plugins": [
+    "transform-es2015-block-scoping",
     "transform-es2015-destructuring",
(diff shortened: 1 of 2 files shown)


amitmtrn:
there is an issue with react and webpack where the `require` import the `react` library which cause `React = ReactDOM`. it will also effect future versions of react and react-dom.
the fix I did is to require the file directly while skipping the package.json of react-dom.

jasonLaster:
Thanks @amitmtrn, I really appreciate the help. this was a very long standing bug!

Issue #1611, opened 2017-01-04
[WhyPaused] handle breakpointConditionThrown


clarkbw:
We don't currently handle the `breakpointConditionThrown` type and I believe it would be easiest to handle this in the [WhyPaused](https://github.com/devtools-html/debugger.html/blob/master/src/components/WhyPaused.js) component. 

Here's how you can reproduce this situation:
* Open [Optimized Away](https://devtools-html.github.io/debugger-examples/examples/optimized-away.html) for debugging
* _**in the debugger**_
* Set a _Conditional Breakpoint_ on **line 24**
* Enter the text `fsck` or any garbage for the condition
* _**in the page**_
* Click the button
* At this point the debugger should be paused, but not for the reason you think

Here's what the old debugger did:

![screen shot 2017-01-03 at 4 45 46 pm](https://cloud.githubusercontent.com/assets/2134/21628407/22ea40e2-d1d4-11e6-9b96-6b9e86bdf799.png)

And here's the relevant data log:

```json
{
  "type": "PAUSED",
  "pauseInfo": {
    "why": {
      "type": "breakpointConditionThrown",
      "message": "fsck is not defined",
      "actors": [
        "server1.conn1.child2/35"
      ]
    }
}
```

Here's the [bug filed in bugzilla](https://bugzilla.mozilla.org/show_bug.cgi?id=1325595)

I'd suggest we do something like this:

![screen shot 2017-01-03 at 5 00 26 pm](https://cloud.githubusercontent.com/assets/2134/21628691/30bfb06a-d1d6-11e6-9c25-2c5952ed2460.png)

And you could get started with something like this:

```diff
diff --git a/src/components/WhyPaused.js b/src/components/WhyPaused.js
index 5d2910e..e97eab5 100644
--- a/src/components/WhyPaused.js
+++ b/src/components/WhyPaused.js
@@ -20,10 +20,14 @@ const WhyPaused = React.createClass({
   render() {
     const { pauseInfo } = this.props;
     const reason = getPauseReason(pauseInfo);
+    const message = pauseInfo ? pauseInfo.getIn(["why"]).get("message") : null;
 
     // => here
     return reason ?
-      dom.div({ className: "pane why-paused" }, L10N.getStr(reason))
+      dom.div({ className: "pane why-paused" }, [
+        dom.div(null, L10N.getStr(reason)),
+        message ? dom.div(null, message) : null
+      ])
       : null;
   }
 });
[...]

PR #1732, opened 2017-01-17
resolved [WhyPaused] handle breakpointConditionThrown - #1611

tomrav:
added localisation keys for exception on conditional breakpoint
     added message error to indicate reason for breaking
     left aligned the WhyPaused message

Associated Issue: #1611 

### Summary of Changes
     added localisation keys for exception on conditional breakpoints
     added message error to indicate reason for breaking
     left aligned the WhyPaused message

### Test Plan
     followed the instructions provided by the original issue (defined a conditional breakpoint with an undefined variable)

### Screenshots
screen shot 2017-01-17 at 20 42 00

#GoodnessSquad

### DIFF
--- a/src/strings.json
+++ b/src/strings.json
@@ -59,6 +59,7 @@
   "whyPaused.exception": "Paused on exception",
   "whyPaused.resumeLimit": "Paused while stepping",
   "whyPaused.pauseOnDOMEvents": "Paused on event listener",
+  "whyPaused.breakpointConditionThrown": "Error with conditional breakpoint",
   "whyPaused.xhr": "Paused on XMLHttpRequest",
   "whyPaused.promiseRejection": "Paused on promise rejection",
   "whyPaused.assert": "Paused on assertion",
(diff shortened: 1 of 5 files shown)


src/strings.json
```diff
@@ -59,6 +59,7 @@
   "whyPaused.exception": "Paused on exception",
   "whyPaused.resumeLimit": "Paused while stepping",
   "whyPaused.pauseOnDOMEvents": "Paused on event listener",
+  "whyPaused.breakpointConditionThrown": "Error with conditional breakpoint",
```
jasonLaster: can you also add this to https://github.com/devtools-html/debugger.html/blob/master/assets/locales/debugger.properties as well

tomrav: added as requested. 👍

jasonLaster:
Thanks @tomrav! 

I like this PR because it makes it much clearer why you're paused in these cases and can make conditional BPs less intimidating.

👍

Issue #1598, opened 2017-01-03
Import client types


jasonLaster:
@clarkbw and @ppold added beautiful types https://github.com/devtools-html/devtools-core/pull/49 here.

Lets import these types and replace types in `src/types`

I think this patch is what we're trying to do: [patch](https://gist.github.com/jasonLaster/10f5c31be71b9d50ffbe471aca06fa13)

arthur801031:
Hi @jasonLaster

I'll take a stab at this :)

jasonLaster:
Thanks @arthur801031 for grabbing one of the hard ones!

PR #1737, opened 2017-01-18
Import client types

arthur801031:
Associated Issue: #1598

### Summary of Changes

* Replace files that use the types in `src/types` with types from [devtools-client-adapters](https://github.com/devtools-html/devtools-core/pull/49/files#diff-25b6e9a88daf18b163d54c3c70532b46)


arthur801031:
@jasonLaster 

Please check. Thank you :)

jasonLaster:
wow - that's awesome that flow passed on CI.

did you try testing this by changing one of the function types?

arthur801031:
I didn't test it. You pretty much did everything in your [patch](https://gist.github.com/jasonLaster/10f5c31be71b9d50ffbe471aca06fa13) haha 😅

jasonLaster:
Hmm, @arthur801031 I tried making a *bad* change to see if the types were being applied and it looks like they weren't picked up.

I ran: `yarn run flow` and got `no errors!`. I think this might be a problem with client-adaptors exporting the types.

```patch
diff --git a/src/actions/breakpoints.js b/src/actions/breakpoints.js
index 1d2f8c0..a393d34 100644
--- a/src/actions/breakpoints.js
+++ b/src/actions/breakpoints.js
@@ -17,7 +17,7 @@ const {
 } = require("../utils/source-map");

-import type { Location } from "devtools-client-adapters";
+import type { Location, Source } from "devtools-client-adapters";

 type addBreakpointOptions = {
   condition: string,
@@ -53,7 +53,7 @@ function enableBreakpoint(location: Location) {
  * @param {Function} $1.getTextForLine Get the text to represent the line
  */
 function addBreakpoint(
-  location: Location,
+  location: Source,
   { condition, getTextForLine } : addBreakpointOptions = {}) {
     if (_breakpointExists(getState(), location)) {
```

arthur801031:
Hi @clarkbw

jasonLaster  and I are trying to figure out the problem that we have in this PR. However, when I ran `yarn run flow`, I got this error message:
screen shot 2017-01-19 at 10 13 34 am
(jasonLaster told me to CC you on this issue)


I've tried these solutions, but none of them worked. 
1. https://flowtype.org/docs/troubleshooting.html
2. https://github.com/facebook/flow/issues/2092

arthur801031:
Hi @jasonLaster 

The reason that it didn't work is because the file path `""devtools-client-adapters"` is wrong. We need this instead `devtools-client-adapters/src/types`.

src/components/Frames.js
```diff
@@ -25,7 +25,12 @@ function renderFrameTitle({ displayName }: Frame) {
 }
 
 function renderFrameLocation({ source, location }: Frame) {
-  const filename = getFilename(source);
+  const thisSource : ?Source = source;
```
arthur801031: Hi @jasonLaster 

Without this check, the flow was producing this error:
screen shot 2017-01-20 at 10 38 51 am

This `if` statement checks if source is undefined, which is possible because Frame's source is declared as `?`. I got this fix from https://flowtype.org/docs/nullable-types.html#_

src/reducers/breakpoints.js
```diff
@@ -13,7 +13,7 @@ const { updateObj } = require("../utils/utils");
 const I = require("immutable");
 const makeRecord = require("../utils/makeRecord");
 
-import type { Breakpoint, Location } from "devtools-client-adapters";
+import type { Breakpoint, Location } from "devtools-client-adapters/src/types";
```
clarkbw: Maybe we should export all the types from the `devtools-client-adapters/index.js` so you don't have to specify `src/types` here.

jasonLaster: Was thinking that too. 

Also it might be possible to do what react dies and set up a declaration so that we don't import at all...

clarkbw:
> We're also missing some export types, which I've made a PR for. devtools-html/devtools-core#102

This has landed now but we need to update to a new version when it comes out and then rebase this.

PR #1739, opened 2017-01-18
Update non-enumerable styling. Fixes #1603

### LINKED ISSUE #1603: Update non-enumerable styling
We're currently using non-enumerable for styling default properties which has the wrong implications:

I think we likely should go with another approach where we gather a hard-coded list of properties for different objects like:`window`, `window.location`

Dalimil:
Associated Issue: #1603

### DIFF
--- a/src/components/Scopes.css
+++ b/src/components/Scopes.css
@@ -1,5 +1,5 @@
 
-.object-node.not-enumerable {
+.object-node.default-property {
   opacity: 0.6;
 }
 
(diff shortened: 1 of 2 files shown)


jasonLaster:
Thanks @Dalimil!

This looks like the fix. Mind sharing some before and after screenshots

Dalimil:
Hmmm, now I'm not really sure if things like `__proto__` or `constructor` should be considered default or not...
Before:
screenshot_2017-01-18_16-23-08
After:
screenshot_2017-01-18_16-20-32

jasonLaster:
Yeah - this is awesome. I think we want to show `__proto__` with dark blue. `constructor` I care less about.

PR #1745, opened 2017-01-18
Fix positioning of toggle pane button in RTL

irfanhudda:
Toggle pane buttons were switched in RTL layout. This cause following issue
recored-2017-01-14_233515-opt

There is also some duplication of styles for toggle pane button in [components/SourceTabs.css#L122](https://github.com/devtools-html/debugger.html/blob/master/src/components/SourceTabs.css#L122) and [components/shared/Button/PaneToggle.css](https://github.com/devtools-html/debugger.html/blob/master/src/components/shared/Button/PaneToggle.css). I wasn't sure which styles should be removed so I modified both for now. 

Let me know which styles should be removed so I can update this pull request

### Summary of Changes

* Update position of toggle pane button for RTL layout
* Update rotation of toggle pane button

recored-2017-01-18_230002-opt

### DIFF
--- a/src/components/shared/Button/PaneToggle.css
+++ b/src/components/shared/Button/PaneToggle.css
@@ -12,7 +12,8 @@
   fill: var(--theme-comment);
 }
 
-.toggle-button-end svg {
+html:not([dir="rtl"]) .toggle-button-end svg,
+html[dir="rtl"] .toggle-button-start svg {
   transform: rotate(180deg);
 }
 
@@ -26,12 +27,12 @@
 
 .toggle-button-start {
   top: 7px;
-  left: 0;
+  offset-inline-start: 0;
 }
 
 .toggle-button-end {
   top: 7px;
-  right: 0;
+  offset-inline-end: 0;
 }
 
 .toggle-button-start.collapsed,
--- a/src/components/shared/Button/PaneToggle.css
+++ b/src/components/shared/Button/PaneToggle.css
@@ -18,8 +18,7 @@ html[dir="rtl"] .toggle-button-start svg {
   transform: rotate(180deg);
 }
 
-html:not([dir="rtl"]) .toggle-button-end.vertical svg,
-html[dir="rtl"] .toggle-button-end.vertical svg {
+html .toggle-button-end.vertical svg {
   transform: rotate(-90deg);
 }
 
--- a/src/components/shared/Button/PaneToggle.css
+++ b/src/components/shared/Button/PaneToggle.css
@@ -5,6 +5,7 @@
   height: 16px;
   transition: transform 0.25s ease-in-out;
   margin: 0 4px;
+  cursor: pointer;
 }
 
 .toggle-button-start svg,
--- a/src/components/shared/Button/PaneToggle.css
+++ b/src/components/shared/Button/PaneToggle.css
@@ -23,12 +23,12 @@ html .toggle-button-end.vertical svg {
 }
 
 .toggle-button-start {
-  top: 7px;
+  top: 8px;
   offset-inline-start: 0;
 }
 
 .toggle-button-end {
-  top: 7px;
+  top: 8px;
   offset-inline-end: 0;
 }
 
--- a/src/components/shared/Button/PaneToggle.css
+++ b/src/components/shared/Button/PaneToggle.css
@@ -18,7 +18,7 @@ html[dir="rtl"] .toggle-button-start svg {
   transform: rotate(180deg);
 }
 
-html .toggle-button-end.vertical svg {
+.toggle-button-end.vertical svg {
   transform: rotate(-90deg);
 }
 
--- a/src/components/shared/Button/PaneToggle.css
+++ b/src/components/shared/Button/PaneToggle.css
@@ -23,12 +23,12 @@ html[dir="rtl"] .toggle-button-start svg {
 }
 
 .toggle-button-start {
-  top: 8px;
+  top: 7px;
   offset-inline-start: 0;
 }
 
 .toggle-button-end {
-  top: 8px;
+  top: 7px;
   offset-inline-end: 0;
 }
 
--- a/src/components/shared/Button/PaneToggle.css
+++ b/src/components/shared/Button/PaneToggle.css
@@ -18,12 +18,9 @@ html[dir="rtl"] .toggle-button-start svg {
   transform: rotate(180deg);
 }
 
-.toggle-button-start.vertical svg {
+html:not([dir="rtl"]) .toggle-button-end.vertical svg,
+html[dir="rtl"] .toggle-button-end.vertical svg {
   transform: rotate(-90deg);
-}
-
-.toggle-button-end.vertical svg {
-  transform: rotate(90deg);
 }
 
 .toggle-button-start {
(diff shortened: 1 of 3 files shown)


wldcordeiro:
@irfanhudda If you could remove the duplicate code in `SourceTabs.css` this looks great. 👍 Just an oversight to have that code there.

irfanhudda:
@wldcordeiro I have removed styles from `SourceTabs.css`. Let me know if I missed something.

wldcordeiro:
@irfanhudda I was able to test this PR locally. There is one small change needed for vertical mode.

[PaneToggle.css line 21](https://github.com/irfanhudda/debugger.html/blob/d17a15cf6016378c1e40ea8f80ce03cf7fcbf1e6/src/components/shared/Button/PaneToggle.css#L21)
```css
.toggle-button-start.vertical svg,
html[dir="rtl"] .toggle-button-end.vertical svg {
  transform: rotate(-90deg);
}
```

irfanhudda:
@wldcordeiro Thanks for pointing this out. I'll take a look at vertical + rtl layout

irfanhudda:
@wldcordeiro I had to make a few more changes to fix pane toggle button rotation in vertical mode, for both rtl and non-rtl layout.

irfanhudda:
@jasonLaster @clarkbw Can you take a look at this? 
Thanks.

jasonLaster:
Hi @irfanhudda thanks for doing this work. I really appreciate the RTL help

I recently organized the components directory and moved Tabs into the Editor directory. Just rebased to update your branch.

While testing I noticed that the "hidden sources dropdown" button is on the right and overlaps the toggle button in RTL. Could you also move that to the left and clean up the positioning?

I realize that this is a lot to test and can be tricky. Thanks!

screen shot 2017-01-22 at 10 13 17 am

irfanhudda:
@jasonLaster I have moved dropdown button  to the left side in RTL.
screenshot

Thanks a lot for pointing out this issue. Let me know if I missed anything more.

jasonLaster:
Thanks @irfanhudda. 

> "P to search + ⌘"?

@flodolo do you know the best way to handle RTL interpolated text?

Here's the [source](https://github.com/irfanhudda/debugger.html/blob/fix-toggle-pane-rtl/src/components/Sources.js#L30-L33). Perhaps we need to use a text-direction style? 

> Let me know if I missed anything more

These buttons could use some additional styling:

* the dropdown text is darker than the plus sign
* the buttons could also be vertically aligned better 
* the buttons could also have more horizontal padding

screen shot 2017-01-22 at 2 01 58 pm

Thanks for jumping in and helping with these other small issues. I realize the scope is growing :)

flodolo:
> @flodolo do you know the best way to handle RTL interpolated text?

Uhm, can you clarify what the problem is? I'm far from an expert with RTL content, but it should do the right thing on its own with the right `dir`, and I wonder if the confusion is created by having Latin text in a RTL layout (it would still go left to right).

This also gives a good coverage of the Bidi algorithm
http://www.iamcal.com/understanding-bidirectional-text/

juliandescottes:
> "P to search + ⌘"?

Just to add to what @flodolo said, the issue is not related to interpolation. You run into the fact that punctuation characters are rendered at the end in rtl, "⌘" being the problematic character here.
 (http://stackoverflow.com/questions/20798667/why-is-a-trailing-punctuation-mark-rendered-at-the-start-with-directionrtl)

If instead of "⌘" you had "Cmd" in your string, the english string would look the same in LTR and in RTL (which is why "Ctrl+P" in the Windows screenshot doesn't show any issue).

Normally it should be fine with the localized version of the string. But here part of the string comes from `formatKeyShortcut` so I guess we can't rely on that. I'm also not sure how RTL users expect shortcuts to be displayed. I would say that if the expected display is still {modifier}+{key} in RTL (and not {key}+{modifier}), maybe we could wrap the output of `formatKeyShortcut` in an element that forces dir=ltr or direction: ltr .

jasonLaster:
Thanks @irfanhudda - mind sharing screens.

By the way, if you want a PRO way of showing the matrix of screens you can borrow from @clarkbw and show them in a table :) [tables](https://help.github.com/articles/organizing-information-with-tables/)


Light     | Dark      | Firebug
--------- | --------- | ---------
<a href="https://cloud.githubusercontent.com/assets/254562/20676302/4cb04a7c-b55d-11e6-855f-654395e2c26f.png">light-theme</a> | <a href="https://cloud.githubusercontent.com/assets/254562/20676304/4cbfbf16-b55d-11e6-9b84-3ee5595e36be.png">dark-theme</a> | <a href="https://cloud.githubusercontent.com/assets/254562/20676303/4cbb0570-b55d-11e6-98b5-d1dd124345cd.png">firebug-theme</a>

irfanhudda:
@jasonLaster This is what I've got right now.
screenshot from 2017-01-23 20-35-56

I'm not sure if this result if consistent across all platforms. (I'm testing on debian with chrome)

irfanhudda:
Here are screens for other themes

|Firebug|Light|
|----------|------|
|firebug|light|

I found that new tab button is noticeable thinner than dropdown button and toggle pane button.

jasonLaster:
@irfanhudda that's looking good. We should change the stroke width then.

Heh - didn't realize how much this stuff bothered me till I saw those screens

jasonLaster:
@irfanhudda mind sharing some vertical mode screens as well. and dark mode

irfanhudda:
@jasonLaster Sure. Here, light mode screenshot is for non-rtl and dark mode is for rtl layout.

|Light|Dark|
|-|-|
|light_nrtl_vert|dark_rtl_vert|

clarkbw:
> dropdown button

This one is using text / fonts so it might be hard to match the SVGs.

jasonLaster:
@clarkbw i see. perhaps font-weight lighter. This is much better than before.

jasonLaster:
@irfanhudda I'll move the source search label problem to a new issue.

src/components/shared/Button/PaneToggle.css
```diff
@@ -5,33 +5,31 @@
...
 }
 
-.toggle-button-end svg {
+html:not([dir="rtl"]) .toggle-button-end svg,
+html[dir="rtl"] .toggle-button-start svg {
   transform: rotate(180deg);
 }
 
-.toggle-button-start.vertical svg {
+html .toggle-button-end.vertical svg {
```
jasonLaster: do we need this here: `html`?

jasonLaster:
pushed the lighter dropdown font.

src/components/shared/Button/PaneToggle.css
irfanhudda: I should have removed this. Thanks for pointing this out.

irfanhudda:
@jsonLaster I've made some adjustments in positioning. This is what I got on chrome.
dark
And this in firefox. There are some differences in both 
dark_firefox

I'm not sure if adjusting pixels is the right way in long term. We might have to use flexbox to get them aligned better.

irfanhudda:
Also I'm not sure about adjustments I made in last commit. If they don't work out well we can revert the changes.

jasonLaster:
@irfanhudda sounds good. This was a great iteration process. Thanks!

PR #1748, opened 2017-01-18
move the properties to a const outside the isDefault function

clarkbw:
Gain 160ms back from expansion time.

### Summary of Changes

* Move the `Object.getOwnPropertyNames(window)` call outside of the function

### Test Plan

Example test plan:

- [x] Check properties in the global context vs function and find they are ==

Before

screen shot 2017-01-18 at 11 12 12 am

After

screen shot 2017-01-18 at 11 14 27 am


jasonLaster:
Very cool. Great instinct

PR #1753, opened 2017-01-19
Bug 1331375 - Can't navigate out of debugger content window using key…

npsecoding:
…board. r?yzen, jlast

Associated Issue: https://bugzilla.mozilla.org/show_bug.cgi?id=1331375

### Summary of Changes

Added a keydown listener to handle entering and exiting code wrapper, (onkeydown in Editor.js)

### Test Plan

Tested manually and wrote a unit test.

Test plan:
- Tab to code editor wrapper
- Enter focuses into editor input
- Escape exits input and places focus back on code editor wrapper

### DIFF
--- a/src/components/Editor.js
+++ b/src/components/Editor.js
@@ -26,7 +26,7 @@ const Breakpoint = React.createFactory(require("./Editor/Breakpoint"));
 const HitMarker = React.createFactory(require("./Editor/HitMarker"));
 
 const { getDocument, setDocument } = require("../utils/source-documents");
-const { shouldShowFooter, clearLineClass } = require("../utils/editor");
+const { shouldShowFooter, clearLineClass, onKeyDown } = require("../utils/editor");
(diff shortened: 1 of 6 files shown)


src/components/Editor.js
```diff
@@ -108,6 +108,22 @@ const Editor = React.createClass({
...
+    if (key === "Escape" && target == textArea) {
+      e.stopPropagation();
+      e.preventDefault();
+      codeWrapper.focus();
+    } else if (key === "Enter" && target == codeWrapper) {
+      e.preventDefault();
+      // Focus into editor's text area
+      textArea.focus();
+    }
+  },
```
jasonLaster: hmm, i think this could be moved to [utils/editor](https://github.com/devtools-html/debugger.html/blob/master/src/utils/editor.js)

```js
function onKeyDown(codeMirror, e) {
  // stuff
}
```

npsecoding:
Fixed the nit.

PR #1760, opened 2017-01-19
Flow searchbar

jasonLaster:


src/components/Editor/SearchBar.js
```diff
@@ -65,7 +67,7 @@ const SearchBar = React.createClass({
     }
   },
 
-  componentDidUpdate(prevProps) {
+  componentDidUpdate(prevProps: any) {
```
clarkbw: You can use `Props` here. https://github.com/facebook/flow/blob/v0.35.0/lib/react.js#L38

src/components/Editor/SearchBar.js
```diff
@@ -133,19 +135,21 @@ const SearchBar = React.createClass({
...
     const sourceText = this.props.sourceText;
     const count = countMatches(query, sourceText.get("text"));
     // eslint-disable-next-line react/no-did-update-set-state
     this.setState({ query, count, index: 0 });
     this.search(query);
   },
 
-  onChange(e) {
-    this.doSearch(e.target.value);
+  onChange(e: any) {
```
clarkbw: `SyntheticEvent` here?

jasonLaster: unfortunately that doesn't work for the eventTarget reason

src/components/Editor/SearchBar.js
jasonLaster: I don't have Props defined

src/components/Editor/SearchBar.js
clarkbw: I thought you just need guard against the target value being null.

`if (e.target && e.target.value)`

jasonLaster: yeah, that didn't work...

jasonLaster:
@clarkbw sounds good. I can leave this open for a bit

clarkbw:
> @clarkbw sounds good. I can leave this open for a bit

Ok, lets go ahead with this.  Everything I found hasn't worked and this is a good step forward.

Issue #1749, opened 2017-01-18
[Editor] Don't show "jump to original location" if there isn't an original location


jasonLaster:
We're currently always showing a **jump** message even when it doesn't make sense.

We can be smart and only show it if there's a place to jump to. We can do this by adding a check here:

```patch
diff --git a/src/components/Editor.js b/src/components/Editor.js
index efdd4bb..8cd421a 100644
--- a/src/components/Editor.js
+++ b/src/components/Editor.js
@@ -157,9 +157,13 @@ const Editor = React.createClass({
       })
     };

-    const menuOptions = [
-      jumpLabel
-    ];
+
+    const menuOptions = [];
+
+    // we will
+    if (hasMappedSource(this.props.selectedSource)) {
+      menuOptions.push(jumpLabel)
+    }

     if (isEnabled("watchExpressions")) {
       menuOptions.push(watchExpressionLabel);
diff --git a/src/utils/source-map.js b/src/utils/source-map.js
index 7a852be..3d9dea3 100644
--- a/src/utils/source-map.js
+++ b/src/utils/source-map.js
@@ -44,6 +44,14 @@ const getOriginalSourceText = workerTask(sourceMapWorker,
 const applySourceMap = workerTask(sourceMapWorker, "applySourceMap");
 const clearSourceMaps = workerTask(sourceMapWorker, "clearSourceMaps");

+// add it here
+function hasMappedSource(source) {
+  if (isOriginalId(source)) {
+    return !!getGeneratedLocation(source)
+  }
+  return !!getOriginalLocation(source)
+}
+
```

### screenshot

![screen shot 2017-01-18 at 2 59 11 pm](https://cloud.githubusercontent.com/assets/254562/22080777/c56dbd42-dd8e-11e6-9fb0-cd2a84075223.png)

PR #1764, opened 2017-01-19
Conditionally show jump to original location option (#1749)

jbhoosreddy:
Associated Issue: #1749

### Summary of Changes

* Add a handler to show jump to original location only if the source has an original source.

### Test Plan

Visual testing

Example test plan:

- [x] Right click in Editor on a source mapped source.
- [x] Should give option for "Jump to Original Source".
- [x] Right click in Editor on a source mapped original source.
- [x] Should give option for "Jump to Generated Source".
- [x] Right click in Editor on a non source mapped source.
- [x] There should not be an option for "Jump to Original Source".

#### When it shouldn't appear
screen shot 2017-01-19 at 9 11 02 pm

#### When it should appear
screen shot 2017-01-19 at 9 11 30 pm

screen shot 2017-01-19 at 9 11 43 pm

### DIFF
--- a/src/components/Editor.js
+++ b/src/components/Editor.js
@@ -30,7 +30,7 @@ const { shouldShowFooter, clearLineClass, onKeyDown } = require("../utils/editor
 const { isFirefox } = require("devtools-config");
 const { showMenu } = require("../utils/menu");
 const { isEnabled } = require("devtools-config");
-const { isOriginalId } = require("../utils/source-map");
+const { isOriginalId, hasMappedSource } = require("../utils/source-map");
 
 require("./Editor.css");
 
@@ -136,8 +136,8 @@ const Editor = React.createClass({
 
     const sourceLocation = {
       sourceId: this.props.selectedLocation.sourceId,
-      line: line,
-      column: ch
+      line: line + 1,
+      column: ch + 1
     };
 
     const pairedType = isOriginalId(this.props.selectedLocation.sourceId)
@@ -159,9 +159,11 @@ const Editor = React.createClass({
       })
     };
 
-    const menuOptions = [
-      jumpLabel
-    ];
+    const menuOptions = [];
+
+    if (this.state.hasMappedSource) {
+      menuOptions.push(jumpLabel);
+    }
 
     if (isEnabled("watchExpressions")) {
       menuOptions.push(watchExpressionLabel);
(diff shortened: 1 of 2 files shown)


src/utils/source-map.js
```diff
@@ -34,6 +36,10 @@ function shouldSourceMap(): boolean {
   return prefs.clientSourceMapsEnabled;
 }
 
+function hasMappedSource(source: Source): boolean {
+  return !isOriginalId(source.id);
```
jasonLaster: hmm, two things:

1. I think this should be `source.get("id")` -> i saw an error in the console
2. I think we first need to check if it's original or generated and then check for the other id... right now we assume its generated

jbhoosreddy: oh sorry for the messup. it's still wip. I'll ping you again once I'm ready with this.

jbhoosreddy:
@jasonLaster up for review now.

src/components/Editor.js
```diff
@@ -136,8 +136,8 @@ const Editor = React.createClass({
 
     const sourceLocation = {
       sourceId: this.props.selectedLocation.sourceId,
-      line: line,
-      column: ch
+      line: line + 1,
+      column: ch + 1
     };
```
jbhoosreddy: the `source-map` library expects numeric values greater than 0.

jasonLaster:
Thanks @jbhoosreddy! This is pretty cool. I love when a feature is UI + redux.

I'm not sure if i'm a fan of using state in this way to handle the async side of the code, so i tried making the handler async.

* the async state seemed to scatter some of the logic so it was tough to follow
* i think this is the first async event handler, which is kinda cool

jbhoosreddy:
This is also one of the things I was considering but I wasn't sure if we wanted to go that way. I think I was thinking maybe it'd slow down the context menu appearance. Like perceived lag for the user as it is fetching this information on the event. @jasonLaster. The fact that it'll be mapped should be constant for each source.

That being said, I didn't see any lag when I tried it with an async function. What do you think?

jasonLaster:
I don't see any lag either. I think if there were some then I'd agree with you, but it's pretty snappy.

Communicating with a worker is now [~1ms](https://hacks.mozilla.org/2015/07/how-fast-are-web-workers/) and doing the lookup is fast.

jasonLaster:
@jbhoosreddy I async'd `hasMappedSource` as well. Let me know what you think.

jasonLaster:
merging when it goes green

PR #1791, opened 2017-01-23
Update mochitest Runner

jasonLaster:
Oops, realized i didn't update the mochitest runner.

When this passes, we'll likely want to clear the cache again to make sure that the james' docker is rm'd. 

CC @juliandescottes


juliandescottes:
CI was ok on https://github.com/devtools-html/debugger.html/pull/1793, so I guess this was just a temporary issue on Circle CI side.

Issue #1794, opened 2017-01-23
Test failures with the new docker image


juliandescottes:
After updating the docker image the following tests are failing on CI

```
The following tests failed:
5278 INFO TEST-UNEXPECTED-FAIL | devtools/client/debugger/new/test/mochitest/browser_dbg-console.js | Split console is hidden. - 
Stack trace:
    chrome://mochitests/content/browser/devtools/client/debugger/new/test/mochitest/browser_dbg-console.js:null:44
    Tester_execTest@chrome://mochikit/content/browser-test.js:737:9
    Tester.prototype.nextTest</<@chrome://mochikit/content/browser-test.js:657:7
    SimpleTest.waitForFocus/waitForFocusInner/focusedOrLoaded/<@chrome://mochikit/content/tests/SimpleTest/SimpleTest.js:744:59
5279 INFO TEST-UNEXPECTED-FAIL | devtools/client/debugger/new/test/mochitest/browser_dbg_keyboard_navigation.js | Editor is enabled - Got [object HTMLTextAreaElement], expected [object HTMLButtonElement]
Stack trace:
    chrome://mochikit/content/browser-test.js:test_is:913
    chrome://mochitests/content/browser/devtools/client/debugger/new/test/mochitest/browser_dbg_keyboard_navigation.js:null:17
    Tester_execTest@chrome://mochikit/content/browser-test.js:737:9
    Tester.prototype.nextTest</<@chrome://mochikit/content/browser-test.js:657:7
    SimpleTest.waitForFocus/waitForFocusInner/focusedOrLoaded/<@chrome://mochikit/content/tests/SimpleTest/SimpleTest.js:744:59
5280 INFO TEST-UNEXPECTED-FAIL | devtools/client/debugger/new/test/mochitest/browser_dbg_keyboard_navigation.js | Focused on container - Got [object HTMLDivElement], expected [object HTMLButtonElement]
Stack trace:
    chrome://mochikit/content/browser-test.js:test_is:913
    chrome://mochitests/content/browser/devtools/client/debugger/new/test/mochitest/browser_dbg_keyboard_navigation.js:null:22
    Tester_execTest@chrome://mochikit/content/browser-test.js:737:9
    Tester.prototype.nextTest</<@chrome://mochikit/content/browser-test.js:657:7
    SimpleTest.waitForFocus/waitForFocusInner/focusedOrLoaded/<@chrome://mochikit/content/tests/SimpleTest/SimpleTest.js:744:59
[...]

PR #1795, opened 2017-01-23
skip tests failing with new docker image

juliandescottes:
Workaround for #1794 , disabling failing tests while we figure out issues with the new docker image.

Let's see if CI is green with this one.


juliandescottes:
I suggest we merge this to get CI green again, and work on fixing the new docker image in parallel.

Issue #1773, opened 2017-01-20
[CommandBar] Make buttons tabbable


Garbee:
The CommandBar uses [spans](https://github.com/devtools-html/debugger.html/blob/7273d1c19d5a85bf4407c888db845a4ecc6bb69a/src/components/CommandBar.js#L67) to create the buttons. These are not naturally tabbable therefore inaccessible by keyboards.

### Recommendation

Refactor these to be `button` elements since they are actions to be carried out. If that isn't desirable, then a minimal `tabindex=1` should do the trick. Either way we go, an `aria-label` should also be applied to the commands for screen readers.

PR #1800, opened 2017-01-23
Use native buttons for CommandBar

Garbee:
Associated Issue: #1773 

### Summary of Changes

* `debugBtn` returns a button element and takes a `disabled` parameter
*  styling updated to handle buttons instead of spans for CommandBar items.
* Also addressed minor styling error with generated markup using the `pointer` cursor for disabled items.

### Test Plan

Load up the debugger and verify the CommandBar still looks the way it should. Upon clicking in the right area to make sure it is first in line for tab order, press tab to see the "Pause on next execution" and then Exception buttons highlight.

### Screenshots/Videos

commandbar

### DIFF
--- a/src/components/SecondaryPanes/CommandBar.js
+++ b/src/components/SecondaryPanes/CommandBar.js
@@ -76,15 +76,18 @@ function handlePressAnimation(button) {
   }, 200);
 }
 
-function debugBtn(onClick, type, className, tooltip) {
+
+function debugBtn(onClick, type, className, tooltip, disabled = false) {
   className = `${type} ${className}`;
-  return dom.span(
+  return dom.button(
     {
       onClick,
       className,
-      key: type
+      key: type,
+      "aria-label": tooltip,
+      disabled
     },
-    Svg(type, { title: tooltip })
+    Svg(type)
   );
 }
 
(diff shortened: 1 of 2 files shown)


src/components/SecondaryPanes/CommandBar.js
```diff
@@ -62,11 +62,13 @@ function formatKey(action) {
   return formatKeyShortcut(key);
 }
 
-function debugBtn(onClick, type, className, tooltip) {
+function debugBtn(onClick, type, className, tooltip, disabled) {
```
jasonLaster: hmm, small thing but maybe it's time to use an options object

```js
debugBtn({onClick, type, className, tooltip, disabled = false} = {})
```

This would, make the code more explicit, avoid needless `false` params...

src/components/SecondaryPanes/CommandBar.js
```diff
@@ -62,11 +62,13 @@ function formatKey(action) {
...
   className = `${type} ${className}`;
-  return dom.span(
-    { onClick, className, key: type },
-    Svg(type, { title: tooltip })
+  return dom.button(
```
jasonLaster: 👍

src/components/SecondaryPanes/CommandBar.js
Garbee: Oh yea, ES6 has those...

Garbee:
Fixed the generator signature with a default parameter now.

Garbee:
Just rebased to resolve the merge conflicts from the localization fixes.

jasonLaster:
@Garbee i think the animations also caused a merge conflict

Garbee:
Rebased

jasonLaster:
Hmm, @Garbee what do you think about styling the focus state? You know, now that they can be styled...

I'm thinking we bump the color a bit when it's focused and lose the border.

Garbee:
Yea, I can do that pretty well. :wink: I was trying to keep the initial merge minimal on any style changes. That way any active state stuff can be discussed on its own. We need a consistent UX here, I was thinking of copying how the file manager is styled when active.

jasonLaster:
Hmm, I'm not sure if I like a blue background / white color. That's pretty intense :/

Garbee:
Yea, I need to check the color contrast on that first. But, these things should be consistent. So if we don't want the blue over here if all checks out then we should look at changing it in the file manager. It's super easy to just do a nice 30% or whatever shade highlight of the current color of a button. Shades are an easy way to make distinct active and focus states.

Garbee:
Latest update has highlights based on the `currentColor` of the buttons.

<details>
<summary>Light Mode<summary>

highlight-light

</details>

<details>
<summary>Dark Mode<summary>

highlight-dark

</details>

jasonLaster:
Thanks @Garbee. Two things:

1. looks like we've got a bar height problem

screen shot 2017-01-24 at 9 34 37 am


2. I think the background color is a bit much for my liking. Could you try using `--theme-color-bluegrey` fills and see what it looks like?

Garbee:
I can't get the overflow from the "Why" bar to go over the CommandBar. Maybe that is a mac-specific rendering issue?

Using `var(--theme-color-bluegray)` for the background doesn't provide enough contrast on the exception mode button for its different states. This can mean that developers are incapable of seeing which mode the button is in by color if it is being hovered or focused.

bluegrey-highlight

Garbee:
Also, we *should* eventually shift to using a set color with known good contrast for all of these things. However, while we have the tri-mode button it makes a quirky case where the `currentColor` setup provides good contrast and feedback on the mode. So while we have that, this is a cheap way to have good contrast without needing to play with a lot of different colors underneath.

Issue #1798, opened 2017-01-23
[RTL]  "⌘+P to search" label


jasonLaster:
Currently, the way we treat RTL text is creating issues for our source search label. 

Here's a screenshot of the problem

![](https://cloud.githubusercontent.com/assets/254562/22183409/bc03fca8-e08b-11e6-8830-0d5b2b85a455.png)

I think @juliandescottes had a good idea of wrapping `⌘+P` in a span with the attribute of dir="ltr" so that it is always placed correctly.

---

**flodolo**
Uhm, can you clarify what the problem is? I'm far from an expert with RTL content, but it should do the right thing on its own with the right dir, and I wonder if the confusion is created by having Latin text in a RTL layout (it would still go left to right).

This also gives a good coverage of the Bidi algorithm
http://www.iamcal.com/understanding-bidirectional-text/
 👍 1  

---
**julian**
"P to search + ⌘"?
Just to add to what @flodolo said, the issue is not related to interpolation. You run into the fact that punctuation characters are rendered at the end in rtl, "⌘" being the problematic character here.
(http://stackoverflow.com/questions/20798667/why-is-a-trailing-punctuation-mark-rendered-at-the-start-with-directionrtl)

If instead of "⌘" you had "Cmd" in your string, the english string would look the same in LTR and in RTL (which is why "Ctrl+P" in the Windows screenshot doesn't show any issue).

Normally it should be fine with the localized version of the string. But here part of the string comes from formatKeyShortcut so I guess we can't rely on that. I'm also not sure how RTL users expect shortcuts to be displayed. I would say that if the expected display is still {modifier}+{key} in RTL (and not {key}+{modifier}), maybe we could wrap the output of formatKeyShortcut in an element that forces dir=ltr or direction: ltr .

Garbee:
If someone who uses a native RTL language could clone #1803 to see if that fixes the problem as described. That simply forces RTL mode on the existing span. We apparently want it to continue displaying like that anyways, it is just the command that is off.

If it is that we in fact need the command at the end, then the solution isn't that exactly. But I should be able to modify the patch to get the expected functionality.

PR #1803, opened 2017-01-23
Force LTR for search hint

Garbee:
Associated Issue: #1798 

### Summary of Changes

* Force RTL mode on the search hint

### Test Plan

If I understand correctly this patch needs to be verified by someone who's language is RTL as the issue occurs due to the way non-English RTL works.

Example test plan:

- Open panel with native RTL language and verify the search hint displays without `P` at the start of the string.

### Screenshots/Videos

N/A - English user so my RTL mode doesn't trigger the problem as described.

### DIFF
--- a/src/components/Sources.js
+++ b/src/components/Sources.js
@@ -25,6 +25,7 @@ const Sources = React.createClass({
       return dom.span(
         {
           className: "sources-header-info",
+          dir: "ltr",
           onClick: () => this.props.toggleFileSearch(true)
         },
         L10N.getFormatStr("sources.search",


jasonLaster:
@Garbee mind sharing a screenshot?

Garbee:
selection_025

That is however with or without the patch for me. By just toggling the direction via DevTools my machine isn't reproducing the problem. I asked on the original issue for someone who is non-English native to test this to see if it resolves it appropriately for them.

jasonLaster:
@Garbee try changing LTR locally

https://github.com/devtools-html/debugger.html/blob/master/docs/local-development.md#rtl

Garbee:
Even with the config handling it from first-render the issue still doesn't reproduce for me with `master` as it is now. Maybe it only affects Macs due to the symbol they get?

jasonLaster:
Yup, this definitely fixes the problem

screen shot 2017-01-23 at 5 33 48 pm

Issue #1802, opened 2017-01-23
Break point snippets are truncated after 30 chars


assafdd:
Would be much nicer using CSS text-overflow: ellipsis.
<img width="1436" alt="breakpoint_sinppet_truncated" src="https://cloud.githubusercontent.com/assets/1753718/22219078/0ba5c2ba-e1b2-11e6-823b-21231c8b89aa.png">

jasonLaster:
@assafdd completely agreed. The problem is that I don't understand CSS ellipsis :)

assafdd:
@jasonLaster  PR on the way :)

clarkbw:
@assafdd can you also make sure to add some padding on the right hand side?  Just so the ellipsis doesn't hit the edge.  The same padding that exists on the left hand side should probably be good.  🥂

assafdd:
@clarkbw will do

PR #1804, opened 2017-01-23
use CSS text-overflow: ellipsis for breakpoint snippet truncate (#1802)

assafdd:
Break point snippets are truncated after 30 chars #1802

trancate_using_css


src/components/SecondaryPanes/Breakpoints.css
```diff
@@ -77,7 +77,10 @@ html .breakpoints-list .breakpoint.paused {
 }
 
 .breakpoint-snippet {
+  overflow: hidden;
+  text-overflow: ellipsis;
   padding-inline-start: 18px;
+  padding-right: 18px;
```
jasonLaster: I think this should be `padding-inline-end` unless bi-direction does not work there...

jasonLaster:
Thanks @assafdd

Issue #1756, opened 2017-01-19
The cmd+O keyboard shortcut should also act as the cmd+P shortcut


clarkbw:
_Coming out of #1752_ 

Chrome has the alias `cmd+O` (Open) for the `cmd+P` source search.  Likely a common mistake for many programmers is to try using the common keyboard shortcut for open file.

Likely you need to start here:
https://github.com/devtools-html/debugger.html/blob/master/src/components/SourceSearch.js#L67

```js
  componentWillUnmount() {
    const shortcuts = this.context.shortcuts;
    shortcuts.off(`CmdOrCtrl+${L10N.getStr("sources.search.key")}`,
      this.toggle);
    // turn off here
    shortcuts.off("Escape", this.onEscape);
  },

  componentDidMount() {
    const shortcuts = this.context.shortcuts;
    shortcuts.on(`CmdOrCtrl+${L10N.getStr("sources.search.key")}`, this.toggle);
    // turn on here
    shortcuts.on("Escape", this.onEscape);
  },
```

Don't forget to use the [L10N](https://github.com/devtools-html/debugger.html/blob/master/docs/local-development.md#l10n) system for the keyboard alias shortcut

clarkbw:
@steffiau is doing this one in #1815

PR #1812, opened 2017-01-24
Support Cmd+O to open search.

mattzeunert:
Implements #1756, so that Cmd+O can be used in the same way as Cmd+P.


jasonLaster:
Hmm, it looks like this is a duplicate of https://github.com/devtools-html/debugger.html/pull/1815...

mattzeunert:
Yep, same change.

Issue #1751, opened 2017-01-18
avoid right / left references in L10N string identifiers 


clarkbw:
[bug 1321255](https://bugzilla.mozilla.org/show_bug.cgi?id=1321255) avoid using right / left references in the string identifiers 

Instead of this
```
# LOCALIZATION NOTE (sourceTabs.closeTabsToRight): Editor source tab context menu item
# for closing the tabs to the right of the selected tab.
sourceTabs.closeTabsToRight=Close tabs to the right
```

Use this:
```xml
<!-- LOCALIZATION NOTE (closeTabsToTheEnd.label): This should indicate the
direction in which tabs are closed, i.e. locales that use RTL mode should say
left instead of right. -->
<!ENTITY  closeTabsToTheEnd.label            "Close Tabs to the Right">
<!ENTITY  closeTabsToTheEnd.accesskey        "i">
```

clarkbw:
We should also update the docs to reflect this requirement so we can avoid it next time.

clarkbw:
Here's what you need to get started:

Take a look at the [localization](https://github.com/devtools-html/debugger.html/blob/master/docs/local-development.md#internationalization) docs.

Change the label here: https://github.com/clarkbw/debugger.html/blob/master/src/components/SourceTabs.js#L116 to use something like `closeTabsToTheEnd` and update all the references to the old `closeTabsToTheRight`.

Then also in [strings.json](http://github.com/clarkbw/debugger.html/blob/bd9f991ccda7952770a1f132a7a8d6a6fb5c76dc/src/strings.json#L52) and [debugger.properties](http://github.com/clarkbw/debugger.html/blob/bd9f991ccda7952770a1f132a7a8d6a6fb5c76dc/assets/locales/debugger.properties#L278)

Finally, open up the [RTL (Right To Left)](https://github.com/devtools-html/debugger.html/blob/master/docs/local-development.md#rtl) docs and update them with a note about using string identifiers like End or Start instead of Right or Left respectively.

garrettgsb:
Stand back guys, I got this.

PR #1817, opened 2017-01-24
L10n: Avoid left/right references

garrettgsb:
Associated Issue: #1751 

### Summary of Changes

* Changed `closeTabsToRight` to `closeTabsToEnd`
* And pretty much all of the permutations of that... Like `node-menu-close-tabs-to-right`/`close-tabs-to-end`, and L10n strings

### Test Plan

I do not have a test plan, other than the app looks just like it did before the PR.


jasonLaster:
I'll review more closely tomorrow, but looks good!

jasonLaster:
Just updated the branch to fix the conflict

assets/locales/debugger.properties
```diff
@@ -273,9 +273,9 @@ sourceTabs.closeTab=Close tab
 # for closing the other tabs.
 sourceTabs.closeOtherTabs=Close others
 
-# LOCALIZATION NOTE (sourceTabs.closeTabsToRight): Editor source tab context menu item
-# for closing the tabs to the right of the selected tab.
-sourceTabs.closeTabsToRight=Close tabs to the right
+# LOCALIZATION NOTE (sourceTabs.closeTabsToEnd): Editor source tab context menu item
+# for closing the tabs to the end of the selected tab.
+sourceTabs.closeTabsToEnd=Close tabs to the end
```
jasonLaster: Hmm, @clarkbw should this be `Close tabs to the end` or `Close tabs to the right`?

My understanding is that it can still be "right" and translators will know how to fix this with RTL languages

clarkbw: Yes, so the string ID should be "to the end" but they text should "to the right", at least for english where the language is LTR (left to right).  This allows the translator to choose what "the end" means based on the language.

garrettgsb: Cool-- Also changed string in `strings.json` back to "...to the right."

irfanhudda:
@garrettgsb It might be helpful to also update RTL section in the docs for future contributors.

jasonLaster:
@garrettgsb Landed!

Issue #1806, opened 2017-01-23
The "new tab" button should toggle search


jasonLaster:
Currently, clicking the new tab button opens the search bar, but clicking it again doesn't do anything. We should change it so that it toggles the search bar.

We should also double check to see if clicking the seearch label in the sources pane toggles the search bar aswell.

```diff
diff --git a/src/actions/ui.js b/src/actions/ui.js
index 2ced82b..82ac7a3 100755
--- a/src/actions/ui.js
+++ b/src/actions/ui.js
@@ -1,13 +1,20 @@
 // @flow
 const constants = require("../constants");
-const { getSource } = require("../selectors");
+const { getSource, isSearchOn } = require("../selectors");
 import type { ThunkArgs } from "./types";

-function toggleFileSearch(searchOn: boolean) {
-  return {
-    type: constants.TOGGLE_FILE_SEARCH,
-    searchOn
-  };
+function toggleFileSearch() {
+
+  // lets make this a legit toggle...
+  return ({ dispatch, getState }: ThunkArgs) => {
+    // check to see if we're toggled
+    // isSearchOn might be the wrong function
+    const searchOn = isSearchOn(getState());
+    return {
+      type: constants.TOGGLE_FILE_SEARCH,
+      searchOn: !searchOn
+    };
+  });
 }
```

Kelly-James:
Me and @ckwjdals90 got this one!

arthur801031:
I'll take a stab at this :)

PR #1820, opened 2017-01-24
Search label and new tab button should toggle search

arthur801031:
Associated Issue: #1806

### Summary of Changes

* Add code to toggle and close the file search bar in `actions/ui.js`


arthur801031:
Hi @jasonLaster 

Please check. Thank you! :)

jasonLaster:
@arthur801031 this looks good, but wouldn't it be simpler if toggle *always* toggled the search bar. I guess I'm not sure if we *want* the bool param.

arthur801031:
@jasonLaster 

Because [SourceSearch](https://github.com/devtools-html/debugger.html/blob/master/src/components/SourceSearch.js#L90-L101) decides whether or not to display the search container based on `this.props.searchOn`. We need to use `this.props.toggleFileSearch` to set `this.props.searchOn` to toggle the search bar.

jasonLaster:
Hmm, @arthur801031 couldn't you also do the toggling in the action? 

https://github.com/devtools-html/debugger.html/issues/1806#issue-202647517

I could easily be missing something here...

arthur801031:
Hi @jasonLaster 

Great point! Instead of passing `!searchOn` to toggle search bar, we could just toggle it in `action` and without passing a parameter.

src/actions/ui.js
```diff
@@ -1,12 +1,18 @@
...
-const { getSource } = require("../selectors");
+const { getSource, getFileSearchState } = require("../selectors");
 import type { ThunkArgs } from "./types";
 
 function toggleFileSearch(searchOn: boolean) {
-  return {
-    type: constants.TOGGLE_FILE_SEARCH,
-    searchOn
+  return ({ dispatch, getState }: ThunkArgs) => {
+    if (searchOn) {
```
jasonLaster: lol, this could get hairy

for instance `toggleFileSearch(true)` or `toggleFileSearch(false)`

I think I'd just change the API to `toggleFileSearch()` and then always flip it here. KISS keep it simple stupid lol

arthur801031: Hi @jasonLaster 

I actually tried that....There are 2 problems:
1. `getFileSearchState(getState())` would return `undefined` when 1. first click on new tab button and 2. click on any of the search results.
2. In some cases, such as pressing esc, clicking on the close button for the search bar, etc, we actually don't want to toggle file search. We want to close it.

These are the reasons why I pass `true` to toggle search bar and `false` to close the search bar.

arthur801031: Or we could make another function for closingFileSearch?

jasonLaster: Hmm, I like that. lets have two actions `toggle` and `close`? 

Also, i suppose the close button will always close given it only is shown when open :)

jasonLaster: I think we want `!getFileSearchState(getState())`

arthur801031:
Hi @jasonLaster 

Please check. Thank you! :)

src/actions/tests/ui.js
```diff
@@ -7,10 +7,18 @@ describe("ui", () => {
   it("should toggle the visible state of file search", () => {
     const { dispatch, getState } = createStore();
     expect(getFileSearchState(getState())).to.be(false);
-    dispatch(actions.toggleFileSearch(true));
+    dispatch(actions.toggleFileSearch());
     expect(getFileSearchState(getState())).to.be(true);
   });
 
+  it("should close file search", () => {
```
arthur801031: I added this unit test for `closeFileSearch` :)

src/actions/types.js
```diff
@@ -98,7 +98,7 @@ export type panelPositionType = "start" | "end";
 
 type UIAction =
   {
-    type: "TOGGLE_FILE_SEARCH",
+    type: "SET_FILE_SEARCH",
```
arthur801031: I changed the name from `TOGGLE_FILE_SEARCH ` to `SET_FILE_SEARCH ` because this action involves toggling and closing the file search bar.

src/actions/tests/ui.js
jasonLaster: 💃

Issue #1789, opened 2017-01-22
[Editor] conditionally hide watch expressions label


jasonLaster:
We currently disable the "Add watch expressions" label when no text is selected. I think it might be nicer to not show the item at all. 

The advantage of this is that we don't present the user with an option she can't do. 

I'm not sure if hiding or disabling items is preferred, but i see us hiding more labels than we're disabling at the moment.

CC @clarkbw 

patch:

```diff
diff --git a/src/components/Editor/index.js b/src/components/Editor/index.js
index 864ada3..4f57e5b 100644
--- a/src/components/Editor/index.js
+++ b/src/components/Editor/index.js
@@ -128,6 +128,8 @@ const Editor = React.createClass({

     const { selectedLocation } = this.props;

+    const textSelected = this.editor.codeMirror.somethingSelected();
+
     event.stopPropagation();
     event.preventDefault();

@@ -156,7 +158,6 @@ const Editor = React.createClass({

     const watchExpressionLabel = {
       accesskey: "E",
-      disabled: !this.editor.codeMirror.somethingSelected(),
       label: L10N.getStr("expressions.placeholder"),
       click: () => this.props.addExpression({
         input: this.editor.codeMirror.getSelection()
@@ -169,7 +170,7 @@ const Editor = React.createClass({
       menuOptions.push(jumpLabel);
     }

-    if (isEnabled("watchExpressions")) {
+    if (isEnabled("watchExpressions") && textSelected) {
       menuOptions.push(watchExpressionLabel);
     }
```

irfanhudda:
We might also want to avoid showing empty context menu like this
![empty_context](https://cloud.githubusercontent.com/assets/1755089/22212616/6337bd9e-e1b7-11e6-9142-c465e7037447.gif)

mattzeunert:
I'm taking a look at fixing this one.

PR #1821, opened 2017-01-24
Conditionally hide watch expressions label

mattzeunert:
Fix issue #1789:

- Hide Watch Expressions menu item if no text is selected
- Don't show empty context menus

I removed the `return` statement from `return menu.popup`, as it suggests the `popup` function returns a meaningful value, when in fact it just returns `undefined`.

### DIFF
--- a/src/components/shared/menu.js
+++ b/src/components/shared/menu.js
@@ -60,6 +60,10 @@ function onShown(menu, popup) {
 }
 
 function showMenu(e, items) {
+  if (items.length === 0) {
+    return;
+  }
+
   const menu = new Menu();
   items.forEach(item => menu.append(new MenuItem(item)));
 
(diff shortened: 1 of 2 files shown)


jasonLaster:
Thanks @mattzeunert!

I'll test it in the morning, but looks good

src/components/shared/menu.js
```diff
@@ -60,6 +60,10 @@ function onShown(menu, popup) {
 }
 
 function showMenu(e, items) {
+  if (items.length === 0) {
+    return;
+  }
```
jasonLaster: 👍  

I like that you're fixing this globally

src/components/shared/menu.js
```diff
@@ -68,7 +72,7 @@ function showMenu(e, items) {
   }
 
   menu.on("open", (_, popup) => onShown(menu, popup));
-  return menu.popup(e.clientX, e.clientY, { doc: document });
+  menu.popup(e.clientX, e.clientY, { doc: document });
```
jasonLaster: 👍 

Yep, returning the popup element could allow some dangerous things to happen

PR #1824, opened 2017-01-24
Tweak styles of watch expression pane

irfanhudda:
Styling of watch expression pane was not consistent with others in secondary pane. These are few changes to correct that.

### Summary of Changes

* Add viewbox attribute for `refresh.svg`
* Change color of placeholder for chrome and firefox
* Change font-size of `input-expression`

### Screenshots

|-|Firefox|Chrome|
|-|-|-|
|Before changes|before-change-firefox|before-change-chrome|
|After changes|after-change-firefox|after-change-chrome|

Also when hovering over refresh button cursor is not pointer. 
I was not sure which file to edit for this change. Let me know if I missed something.

### DIFF
--- a/src/components/SecondaryPanes/Expressions.css
+++ b/src/components/SecondaryPanes/Expressions.css
@@ -15,6 +15,13 @@
   color: var(--theme-body-color);
 }
 
+.input-expression::-moz-placeholder {
+  text-align: center;
+  font-style: italic;
+  color: var(--theme-body-color);
+  opacity: 1;
+}
+
 .input-expression:focus {
   outline-color: var(--theme-selection-background-semitransparent);
   outline-width: 2px;
--- a/src/components/SecondaryPanes/Expressions.css
+++ b/src/components/SecondaryPanes/Expressions.css
@@ -15,7 +15,7 @@
   color: var(--theme-body-color-inactive);
 }
 
-.input-expression::-moz-placeholder {
+.input-expression::placeholder {
   text-align: center;
   font-style: italic;
   color: var(--theme-body-color-inactive);
--- a/src/components/SecondaryPanes/Expressions.css
+++ b/src/components/SecondaryPanes/Expressions.css
@@ -23,7 +23,7 @@
 }
 
 .input-expression:focus {
-  outline: 2px solid var(--theme-selection-background-semitransparent);
+  outline: none;
   cursor: text;
 }
 
--- a/src/components/SecondaryPanes/Expressions.css
+++ b/src/components/SecondaryPanes/Expressions.css
@@ -28,8 +28,7 @@
 }
 
 .input-expression:focus {
-  outline-color: var(--theme-selection-background-semitransparent);
-  outline-width: 2px;
+  outline: 2px solid var(--theme-selection-background-semitransparent);
   cursor: text;
 }
 
--- a/src/components/SecondaryPanes/Expressions.css
+++ b/src/components/SecondaryPanes/Expressions.css
@@ -7,6 +7,11 @@
   color: var(--theme-body-color);
   background-color: var(--theme-body-background);
   font-size: 12px;
+  transition: all 0.5s ease;
+}
+
+.input-expression:hover {
+  background-color: var(--theme-selection-background-semitransparent);
 }
 
 .input-expression::-webkit-input-placeholder {
--- a/src/components/SecondaryPanes/Expressions.css
+++ b/src/components/SecondaryPanes/Expressions.css
@@ -12,13 +12,13 @@
 .input-expression::-webkit-input-placeholder {
   text-align: center;
   font-style: italic;
-  color: var(--theme-body-color);
+  color: var(--theme-body-color-inactive);
 }
 
 .input-expression::-moz-placeholder {
   text-align: center;
   font-style: italic;
-  color: var(--theme-body-color);
+  color: var(--theme-body-color-inactive);
   opacity: 1;
 }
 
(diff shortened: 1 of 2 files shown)


jasonLaster:
@jbhoosreddy, @nchevobbe mind taking a look?

src/components/SecondaryPanes/Expressions.css
```diff
@@ -6,12 +6,20 @@
...
+  font-size: 12px;
 }
 
 .input-expression::-webkit-input-placeholder {
   text-align: center;
   font-style: italic;
+  color: var(--theme-body-color);
+}
+
+.input-expression::-moz-placeholder {
```
nchevobbe: maybe we could group those 2 rules together ? 

```css
.input-expression::-webkit-input-placeholder,
.input-expression::-moz-input-placeholder,
```
I even think we can drop the `-moz` prefix since it has been unprefixed in Firefox 51 (https://developer.mozilla.org/en-US/docs/Web/CSS/::placeholder)

jbhoosreddy:
@irfanhudda so thinking about @nchevobbe's point, I think it would be nice to have some differentiating factor between the watch expressions input and the text in the other accordian items.

I was thinking you could add an underline for watch expression input placeholder. what do you think @nchevobbe @jasonLaster?

src/components/SecondaryPanes/Expressions.css
irfanhudda: @nchevobbe group them doesn't seem to work. I tried searching for this and got a question on stackoverflow relevant to this [here](http://stackoverflow.com/questions/16982449/why-isnt-it-possible-to-combine-vendor-specific-pseudo-elements-classes-into-on) .

nchevobbe: learned something today ! Hopefully chrome will have the unprefixed version soon too.

nchevobbe:
> I was thinking you could add an underline for watch expression input placeholder. what do you think @nchevobbe @jasonLaster?

That's something to try. @irfanhudda if you have some time, could you experiment with this an post a couple of screenshots here ? thanks !

irfanhudda:
I changed color of placeholder to `--theme-body-color-inactive` which is being used for placeholder at other places. Although, it looks like there is not much difference for dark mode

|Light|Dark|
|-|-|
|chrome-light|chrome-dark|

@nchevobbe I'll experiment with adding underline to input

irfanhudda:
I tried adding underline to input and here's what I've got. (I used box-shadow for input)
underline
Any suggestions? @jbhoosreddy @nchevobbe

jasonLaster:
@irfanhudda maybe we switch to using a hover color, like the active blue? 

Either way, i think the cursor is pretty good.

jbhoosreddy:
@irfanhudda it doesn't have to be an underline. feel free to experiment with it if you feel there is a better UX option definitely post it and others can also see it.

> Looks like --theme-body-color and --theme-body-color-inactive is same for dark-mode 

As @jasonLaster showed me, you can conditionally set different colors for the dark and light theme. Here's an [example](https://github.com/devtools-html/debugger.html/blob/master/src/components/SecondaryPanes/Expressions.css#L36-L42).

jbhoosreddy:
As for `cursor:pointer` for hover button. I would like to see cursor pointer for that button so that it matches behavior of other buttons in the debugger.

However, the refresh button is implemented [here](https://github.com/devtools-html/debugger.html/blob/master/src/components/SecondaryPanes/index.js#L30-L36). And it was originally borrowed from [here](https://github.com/devtools-html/debugger.html/blob/master/src/components/SecondaryPanes/CommandBar.js#L80-L92). But the original implementation seems to have changed somewhat. Maybe you could figure out what needs to be ported. Should be straightforward enough but looking at command bar buttons should help.

For your reference, there is also a Svg.css [here](https://github.com/devtools-html/debugger.html/blob/master/src/components/shared/Svg.css). But I don't think you'd need to use it.

jasonLaster:
> conditionally set different colors

https://github.com/devtools-html/debugger.html/blob/master/docs/local-development.md#update-a-theme-style

irfanhudda:
So, I added hover color with some transition for outline. Here's what it looks like 
outline

<details>
<summary>
Some other things I tried
</summary>
With underline instead of outline

underline

Using `plus` svg icon
outline-svg
Adding svg icon required more changes than I expected.
</details>

It might be better to go with hover color and outline because it requires minimal changes.

Let me know what you think.

Garbee:
We've seen in the Chrome DevTools that some users don't know how to get their addition added without a control button. So, we may want to consider here a UX that offers a specific "add" button of some kind to trigger adding the value instead of just the `enter` key.

Garbee:
It also might be beneficial to keep the `+` icon on the front and also provide that the action which will focus the add watch expression input. We'd simply hide that from the tab order `tabindex="-1"` since it would just trigger what the input itself does.

src/components/SecondaryPanes/Expressions.css
```diff
@@ -6,17 +6,29 @@
   cursor: pointer;
   color: var(--theme-body-color);
   background-color: var(--theme-body-background);
-  line-height: 12px;
+  font-size: 12px;
+  transition: all 0.5s ease;
```
Garbee: nit: leading zeros don't do anything. Best leave them out entirely.

irfanhudda: Thanks. Learned something new!

jasonLaster:
Hey @irfanhudda, hmm would it be alright if we hold off on the hover state of the input for now?

I think we want to figure out hover and focus states globally and see what patterns we should apply.

There's a lot of good stuff in here that i'd like to land.


I reverted the hover styling and this looked okay for me
https://gist.github.com/a5fef250583b3fe587feed740a2f87b6

jbhoosreddy:
in that case, let us revert that change and try to land this.

irfanhudda:
@jasonLaster I reverted the hover colors for input-expression and outline on focus.

Issue #1705, opened 2017-01-14
[SourceTabs] you don't see previous tabs when navigating back to a URL


jasonLaster:
Steps to reproduce:
1. go to cnn.com, select some tabs
2. go to foxnews.com, select some tabs
3. go back to cnn.com

you expect to see your cnn tabs, but you don't see any. This is because we don't show any tabs unless the selected tab is available. We should fix this logic so that it's possible to show the tabs without a selected tab.

```diff
diff --git a/src/components/SourceTabs.js b/src/components/SourceTabs.js
index 546fec8..d4fc25e 100755
--- a/src/components/SourceTabs.js
+++ b/src/components/SourceTabs.js
@@ -38,6 +38,7 @@ function getHiddenTabs(sourceTabs, sourceTabEls) {
     return Math.min(...topOffsets);
   }

+  // this function might have some issues...
   const tabTopOffset = getTopOffset();
   return sourceTabs.filter((tab, index) => {
     return sourceTabEls[index].getBoundingClientRect().top > tabTopOffset;
diff --git a/src/reducers/sources.js b/src/reducers/sources.js
index c9b001b..6ed9176 100644
--- a/src/reducers/sources.js
+++ b/src/reducers/sources.js
@@ -279,10 +279,10 @@ function getSourceText(state: OuterState, id: string) {
 }

 function getSourceTabs(state: OuterState) {
-  const selectedLocation = getSelectedLocation(state);
-  if (!selectedLocation) {
-    return new I.List([]);
-  }
+  // const selectedLocation = getSelectedLocation(state);
+  // if (!selectedLocation) {
+  //   return new I.List([]);
+  // }

   return state.sources.tabs
     .filter(tab => getSourceByURL(state, tab));
```

jbhoosreddy:
Could I grab this one next @jasonLaster?

jasonLaster:
Go for it.

clarkbw:
@jbhoosreddy are you taking on this one?

jbhoosreddy:
@clarkbw oh yeah. I am.

jasonLaster:
Thanks @jbhoosreddy this was a great fix!

PR #1833, opened 2017-01-25
SourceTabs: render tabs without selected tab

jbhoosreddy:
Associated Issue: #1705 
### Summary of Changes

* Return list of source tabs even when no tab is selected
* If no tab is selected, select the first tab.

Question: Should we persist the selected tab for the user so that the application brings the user back to the same tab? @jasonLaster @clarkbw 

### Test Plan

### DIFF
--- a/src/components/Editor/Tabs.js
+++ b/src/components/Editor/Tabs.js
@@ -249,7 +249,7 @@ const SourceTabs = React.createClass({
   renderTab(source) {
     const { selectedSource, selectSource, closeTab } = this.props;
     const filename = getFilename(source.toJS());
-    const active = source.get("id") == selectedSource.get("id");
+    const active = selectedSource && source.get("id") == selectedSource.get("id");
     const isPrettyCode = isPretty({ url: source.get("url") });
 
     function onClickClose(ev) {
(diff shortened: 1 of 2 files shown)


clarkbw:
> Should we persist the selected tab for the user so that the application brings the user back to the same tab

Might be something nice to tackle in another PR

src/components/Editor/Tabs.js
```diff
@@ -85,6 +85,11 @@ const SourceTabs = React.createClass({
     if (!(prevProps === this.props)) {
       this.updateHiddenSourceTabs();
     }
+    const { selectedSource, sourceTabs, selectSource } = this.props;
+    if (!selectedSource && sourceTabs.size > 0) {
+      const source = sourceTabs.get(0);
+      selectSource(source.get("id"));
+    }
```
jasonLaster: I just tested it locally and it looks like rendering without a selected source works.

Lets take this out and we can always figure out the best thing to do later.
The problem with this is that we're changing state. i.e. if the previously selected source is loaded soon after we do this we're wrong :)

jbhoosreddy: I think you're right. I was unsure what the best would be to handle it. I removed it for now.

Issue #1834, opened 2017-01-25
[Editor] fix the pane toggle button direction


liorpr:
The pane toggle button(.toggle-button-end) direction is wrong on vertical mode. This happens because the first css rule is more specific than the last one

<img width="488" alt="screen shot 2017-01-25 at 1 20 00 pm" src="https://cloud.githubusercontent.com/assets/6891569/22288748/f5ec872e-e300-11e6-9c84-66a35e9469ac.png">

liorpr:
submitting my fix, imo we can choose between explicitly state !improtant or just use something like

`html:not([dir="rtl"]).toggle-button-end.vertical svg,
html[dir="rtl"] .toggle-button-start svg {
  transform: rotate(-90deg);
}
`

wldcordeiro:
@liorpr I'm for avoiding `!important` at pretty much all costs. I feel like that property is a code smell. But thanks for reporting the issue.

PR #1835, opened 2017-01-25
fixed pane toggle button direction

liorpr:
Associated Issue: #1834

added !important to explicitly state that the rule should be applied in vertical mode.

### DIFF
--- a/src/components/shared/Button/PaneToggle.css
+++ b/src/components/shared/Button/PaneToggle.css
@@ -18,8 +18,8 @@ html[dir="rtl"] .toggle-button-start svg {
   transform: rotate(180deg);
 }
 
-.toggle-button-end.vertical svg {
-  transform: rotate(-90deg) !important;
+html .toggle-button-end.vertical svg {
+  transform: rotate(-90deg);
 }
 
 .toggle-button-start {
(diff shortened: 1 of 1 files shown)


irfanhudda:
@liorpr Thanks for fixing this problem. This seems to have come from my previous commit #1745.

I have one suggestion though

src/components/shared/Button/PaneToggle.css
```diff
@@ -19,7 +19,7 @@ html[dir="rtl"] .toggle-button-start svg {
 }
 
 .toggle-button-end.vertical svg {
-  transform: rotate(-90deg);
+  transform: rotate(-90deg) !important;
```
irfanhudda: We could avoid "!important" by modifying selector to `html .toggle-button-end.vertical svg`

Garbee: +1 here. `!important` should not be necessary.

wldcordeiro: Mentioned in the issue, `!important` is pretty much a code smell so I support the above. 👍

jasonLaster:
@liorpr mind sharing some screenshots?

Issue #1837, opened 2017-01-25
[SourceFooter] Polish Pretty Print button


jasonLaster:
The pretty print button is currently pretty small. Also the CSS is pretty bad.

Goals:
* set the footer to a consistent height with the top bar  (30px i think)
* align the pretty print button with flex (align-items center i think)
* make the pretty print button the same height as the command bar items

UI:
![screen shot 2017-01-25 at 9 23 42 am](https://cloud.githubusercontent.com/assets/254562/22294110/09068a10-e2e0-11e6-8deb-fde1ad4f975b.png)

Thanks @juliandescottes for bringing this up

Garbee:
> align the pretty print button with flex (align-items center i think)

It already is. The problem is the SVG creator we have rendering an `i` with the SVG element inside of that. It offsets the styling. We need to do more direct targeting of the SVG/i here to get it to look right. Ideally though, this shouldn't be necessary as the SVG node should return a plain SVG element non-wrapped. That way any styling outside is easy to touch it without knowing exactly what the internal markup structure is.

jasonLaster:
@princiya is on it

wldcordeiro:
![screenshot_20170203_163631](https://cloud.githubusercontent.com/assets/580982/22612603/f5b7f47a-ea2e-11e6-8a69-cfab72af5a7a.png)

Fixing the coverage button to be vertically aligned with pretty print would also be great here.

jasonLaster:
@wldcordeiro we can do code coverage when we land the backend and polish

PR #1847, opened 2017-01-25
Cleaned up pretty print button and other footer actions

Garbee:
Associated Issue: #1837

### Summary of Changes

* Fixed commands height for footer.
* Center aligned the inlined svg contents
* Moved padding into actions from the command bar itself on the bottom to give more touch area
* Moved a11y attributes onto parent button from the SVG for more surface area to trigger.
* Added aria-label while I was here for the code coverage button
* Removed `debugbtn` method used only once for the footer.

### Test Plan

Boot up debugger and check out the pretty print button. Functions the same as the CommandBar does now.

Example test plan:

- Hover pretty print button area to see the title shows for the full surface area instead of just over the icon.
- See icon is centered and not oddly placed higher than the center of the line.

### Screenshots/Videos

selection_027

### DIFF
--- a/src/components/Editor/Footer.js
+++ b/src/components/Editor/Footer.js
@@ -17,14 +17,6 @@ const PaneToggleButton = React.createFactory(
 
 require("./Footer.css");
 
-function debugBtn(onClick, type, className = "active", tooltip) {
-  className = `${type} ${className} action`;
-  return dom.button(
-    { onClick, className, key: type },
-    Svg(type, { title: tooltip, "aria-label": tooltip })
-  );
-}
-
 const SourceFooter = React.createClass({
   propTypes: {
     selectedSource: ImPropTypes.map,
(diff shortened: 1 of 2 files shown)


src/components/Editor/Footer.js
```diff
@@ -54,14 +46,23 @@ const SourceFooter = React.createClass({
...
-      this.onClickPrettyPrint,
-      "prettyPrint",
-      classnames({
-        active: sourceLoaded,
-        pretty: isPretty(selectedSource.toJS())
-      }),
-      L10N.getStr("sourceFooter.debugBtnTooltip")
+    const tooltip = L10N.getStr("sourceFooter.debugBtnTooltip");
+    const type = "prettyPrint";
+    const className = classnames({
```
jasonLaster: I think you can pass some of these in as params:

```js
classNames(action, type, {})
```

```

File: src/components/Editor/Footer.js [hunk 2]
```diff
@@ -54,14 +46,23 @@ const SourceFooter = React.createClass({
...
+      action: true,
+      type,
+      active: sourceLoaded,
+      pretty: isPretty(selectedSource.toJS())
+    });
+
+    return dom.button({
+        onClick: this.onClickPrettyPrint,
+        className,
```
jasonLaster: I'm infavor of inlining this, even though i suppose it'll make the JSX harder in the future...

Garbee: It won't make converting any harder if we go that route. It's a simple conversion to make.

src/components/Editor/Footer.js
Garbee: Nice, didn't know it handled multiple params that way.

Issue #1856, opened 2017-01-25
convert post-merge to js / node script


clarkbw:
`bash` scripts don't work reliably on Windows.  If we converted the [post-merge](https://github.com/devtools-html/debugger.html/blob/master/bin/post-merge) script to be a JavaScript / node script Windows users wouldn't see a large error print out after merging in the latest changes.

PR #1859, opened 2017-01-26
convert post-merge to js/node (#1856)

jbhoosreddy:
Associated Issue: #1856

### Summary of Changes

* convert post-merge to js/node

### Test Plan

Visual testing on Windows VM.
Example test plan:

- [x] Push changes on git
- [x] Observe that there are no errors

Upcoming


package.json
```diff
@@ -34,7 +34,7 @@
     "flow-redux": "flow-coverage-report  -i 'src/reducers/*.js' -i 'src/actions/*.js'  -t text",
     "flow-react": "flow-coverage-report -i 'src/components/**/*.js' -t text",
     "prepush": "yarn run lint && yarn run test -- --dots",
-    "postmerge": "bin/post-merge"
+    "postmerge": "bin/post-merge.js"
```
jasonLaster: hmm, how about `node bin/post-merge.js`

clarkbw:
👍

Garbee:
This seems to be spawning a raw child git process which we've seen cause issues with other things converted into node scripts. Since you're still relying on the OS level details and configuration being just right. An alternative could be to use [nodegit](https://github.com/nodegit/nodegit) or some other git wrapper to abstract all of that away and provide a more explicit API for what we are doing in node land.

jbhoosreddy:
thanks @Garbee! I'll look into this.

clarkbw:
hmm, I'm seeing this error:

```
$ yarn run postmerge
yarn run v0.19.1
$ node bin/post-merge.js 
stderr 
/Users/clarkbw/src/mozilla/debugger.html/bin/post-merge.js:11
    throw error;
    ^

Error: spawn git diff-tree ENOENT
    at exports._errnoException (util.js:1008:11)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:182:32)
    at onErrorNT (internal/child_process.js:348:16)
    at _combinedTickCallback (internal/process/next_tick.js:74:11)
    at process._tickCallback (internal/process/next_tick.js:98:9)
    at Module.runMain (module.js:577:11)
    at run (bootstrap_node.js:352:7)
    at startup (bootstrap_node.js:144:9)
    at bootstrap_node.js:467:3
```

jbhoosreddy:
@clarkbw I'm in the process of converting it to nodegit as per garbee's suggestion. I'll ping you again about this. 😄

Garbee:
@jbhoosreddy Need any help here? We're starting to get reports about the current script failing. I'd like us to try and get this converted quickly so contributors aren't hitting issues.

Garbee:
I just took a quick look at getting nodegit setup. Oh-boy. It adds about 6 minutes on average to `yarn install` time on Ubuntu due to needing to compile itself. Further, it's fairly error-prone and you won't know until runtime if you're missing dependencies on your system for the compile to work right.

Just based on that, we really *should not* switch to using it. If they had reliable prebuilt binaries then it would be an option. However, they do not. So we need to try and solve the Windows issue differently.

Garbee:
So, I think the best way forward here is two physical scripts. One for Unix systems (just like the original) and one for Windows. The `post-merge.js` here will detect the OS, then execute the right script needed. The Windows script will need to run the same git command, but use `findstr` to look for `yarn.lock` in the output and return/log appropriately.

jasonLaster:
@Garbee this might help: https://github.com/devtools-html/devtools-core/blob/master/packages/devtools-launchpad/bin/firefox-driver.js#L28-L29

bin/post-merge.js
```diff
@@ -0,0 +1,16 @@
+// MIT © Sindre Sorhus - sindresorhus.com
+// via https://gist.github.com/sindresorhus/7996717
+
+const execFile = require("child_process").execFile;
+
+execFile("git diff-tree",
+  ["-r", "--name-only", "--no-commit-id", "ORIG_HEAD HEAD"],
```
Garbee: My upgrade is taking some time on my test machine. I have an icky feeling `"ORIG_HEAD HEAD"` here should be `"ORIG_HEAD", "HEAD"`. I'll test to verify once my upgrade finishes.

jbhoosreddy: I think so too. I'll change it now.

jbhoosreddy:
So I take that consensus is to use simple execFile to run two different commands based on the operating system.

Garbee:
The thing is, this command typing it in works perfect on Windows too. So I'm not sure why it is failing (except for that icky feeling I commented on.)

If that change doesn't work, then we should use keep the old bash script for Unix style systems and add a new Windows script and use `execFile` to call the right one yes. Keep them separated for clarity in case they ever need to get modified later.

Garbee:
So, do modify the parameter bit. and leave it alone.

Manually running `yarn run postmerge` to test this *will fail* always, since we aren't working in the right context for git to have an `ORIG_HEAD` set. We need to run this within the postmerge process to actually verify it is working right. Which means, running a merge with a yarn lock change.

This really LGTM even though it smells slightly due to the manual execution. As of right now though, what we have has no change of working on Windows. Let's get that nit fixed to make sure we have the right number of params (as right now I believe it gets quoted into a single param) and then merge. From there we can do a rollback test on a side repo and give it a run in-context to see if it works properly and go from there.

PR #1863, opened 2017-01-26
Adds missing dbg parameter in reload()

yurydelendik:
Associated Issue: n/a

### Summary of Changes

* adds dbg for waitForSources call

### Test Plan

In most of the cases we call reload() without second argument, so this was not exposed earlier. browser_dbg-navigation.js test did not regress for me locally.


jasonLaster:
Nice find

Issue #1747, opened 2017-01-18
Research adopting JSX


jasonLaster:
We would like to layout the case for adding JSX to the debugger and other react tools in DevTools.

Lets use this issue to discuss the trade/offs, style preferences, and adoption plans.

Here's a great [PR](https://github.com/devtools-html/debugger.html/pull/1712) where @Garbee showed the benefits of refactoring the breakpoints component

clarkbw:
I'll add that I want this change.  We already have a webpack build step so the possible negative has already been taken on.  JSX makes the components easier to read and AFAIK was designed so that designers could more easily adapt components without needing to code.

wldcordeiro:
@clarkbw echoing your argument for JSX I think the biggest thing is familiarity for new contributors. Most React projects seem to use JSX so adopting it would make the barrier for entry a little lower. Also it seems like people are more conscious of the HTML element types they use when they're using JSX over the dom helpers and we end up with a lot more semantic HTML.

jbhoosreddy:
I can only talk from personal experience as I don't know too much about the lay of the land. But our way of using React was definitely tricky to pin point because there was little documentation. It might have taken me an even longer time to get things done without my mentors.

Since the official React docs follows JSX and ES6/ES7 JS, following that coding standard should make it much more easy for people to contribute to the project, as they will be familiar with it.

That being said, from recent experience I noticed many other companies face similar issues where their deployment could not have babel or a build step for various reasons. And there, they appreciated being able to utilize the features of React without that problem. This may be one of the few projects on the web where beginners can learn something like that.

That also being said, I'm in favor of JSX also.

bomsy:
One plus from my little experience with JSX, in its own way helps get a quick picture of what the dom structure might look like for some one new to it
 Might be kinda easier i guess to some familiar to basic markup
``` js
dom.div( null,
  dom.img( { src: "avatar.png", className: "profile" }),
  dom.h3( null, [user.firstName, user.lastName].join(" "))
);
```

vs

```js
<div>
  <img src="avatar.png" className="profile" />
  <h3>{[user.firstName, user.lastName].join(' ')}</h3>
</div>
```

wldcordeiro:
@bomsy agreed with that, it definitely makes for a more familiar experience for designers and new contributors. I support the idea of transitioning to JSX. A tracking issue that has adding the babel transform, and then a list of components would be a nice start and encourage contributions.

Garbee:
[...]

PR #1875, opened 2017-01-28
convert WhyPaused to use es6 class

clarkbw:
### Summary of Changes

* Minimally converted the `WhyPaused` component to use es6 classes 
* Tried to use `PureComponent` but it doesn't seem to be defined in `react` where it should.  Needs more investigation.

If we want to really clean up the syntax to make things cleaner and more readable I think we might need to go even further down this rabbit hole which includes JSX #1747 (which would remove the need for the `createFactory` calls).

Our webpack config continues to confound me so I don't think I've done that correctly but it didn't like me using `export` so I needed a new transform plugin.  I gave up trying to do anymore as I think the presets would handle most of this.  I need a 12 week intensive course with Julian and Jason on webpack sorcery to make any more progress here.

This does show that its possible to slowly convert components over.  With a tracking issue for all components you could have each one move over to this fairly quickly.

### Test Plan

- [x] Paused on breakpoints to check that component renders correctly
- [x] Checked console for errors

### DIFF
--- a/src/components/SecondaryPanes/index.js
+++ b/src/components/SecondaryPanes/index.js
@@ -12,7 +12,7 @@ const { getPause, getBreakpoints,
 const { prefs } = require("../../utils/prefs");
 
 const actions = require("../../actions");
-const WhyPaused = React.createFactory(require("./WhyPaused"));
+const WhyPaused = React.createFactory(require("./WhyPaused").default);
 const Breakpoints = React.createFactory(require("./Breakpoints"));
 const Expressions = React.createFactory(require("./Expressions"));
 
--- a/.babelrc
+++ b/.babelrc
@@ -1,5 +1,6 @@
 {
   "plugins": [
+    "transform-es2015-modules-commonjs",
     "transform-es2015-block-scoping",
     "transform-es2015-destructuring",
     "transform-es2015-parameters",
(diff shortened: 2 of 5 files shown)


src/components/SecondaryPanes/index.js
```diff
@@ -12,7 +12,7 @@ const { getPause, getBreakpoints,
 const { prefs } = require("../../utils/prefs");
 
 const actions = require("../../actions");
-const WhyPaused = React.createFactory(require("./WhyPaused"));
+const WhyPaused = React.createFactory(require("./WhyPaused").default);
```
clarkbw: The `default` here is necessary because we're using the es6 class in a non-es6 like world.  The switch to `import` and JSX would allow us to remove that.  There are also babel plugins to remove this as well but its only temporary.

clarkbw:
Friday hack finished just before the 🔔 

🎉

src/components/SecondaryPanes/index.js
jbhoosreddy: this is definitely interesting to find out!

jasonLaster: Great. I think this is a small price to pay for es6 modules and classes.

Garbee:
I really like the look of the code style here. Good work.

Issue #1848, opened 2017-01-25
Source List doesn't refresh properly


jbhoosreddy:
### STR:
0. Set up debugger and debuggee.
1.a. Go to cnn.com.
1.b. Observe sources in the source list.
2.a. Go to foxnews.com
2.b Observe sources in the source list.

### ER:
cnn.com sources should be removed once the debuggee is on foxnews.com.

### AR:
scripts from both wesites persist. and since both websites pull in a lot of scripts, it becomes very noisy.

### Associated Gif example
http://g.recordit.co/JbgDluPQWA.gif

![](http://g.recordit.co/JbgDluPQWA.gif)

jasonLaster:
I think the gif is too long lol

arthur801031:
I will take a stab at this :)

PR #1879, opened 2017-01-30
Source List doesn't refresh properly

arthur801031:
Associated Issue: #1848

### Summary of Changes

* Clear existing sources and state when debuggee changes.

### Known issues
* First source is not added to `sources`.

### Screenshots/Videos
xt9sspiiv3

### DIFF
--- a/src/reducers/sources.js
+++ b/src/reducers/sources.js
@@ -46,6 +46,16 @@ function update(state = State(), action: Action) : Record<SourcesState> {
   let availableTabs = null;
   let location = null;
 
+  // Reset to default when first source is received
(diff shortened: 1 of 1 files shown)


arthur801031:
Hi @jasonLaster !

I've tried various ways to tackle this problem, but so far this is the one that works. This approach checks if a source is the first received source by looking at its `id`, and clears the `sources` and its `state` when it is the first received source. Based on my observations, the first received source always has `33` as its `id`.  I would like to confirm if this is the acceptable approach before I attempt to address the problem with missing the first source since there may be a better solution.

jasonLaster:
closing this as it's chrome specific

Issue #635, opened 2016-08-31
Back and Forward buttons don't update source list


clarkbw:
Part of the  All the mochitests! #634 issue
- [browser_dbg_bfcache.js](https://dxr.mozilla.org/mozilla-central/source/devtools/client/debugger/test/mochitest/browser_dbg_bfcache.js)

Failing Test:

```
  yield testFirstPage();
  yield testLocationChange();
  yield testBack();
  yield testForward();
```

jasonLaster:
This is not something we can do on the client. We should plan to do a fix in the server.

jasonLaster:
changed the label to server

jasonLaster:
closing and moving to the bugs board as a note. We can always re-open when we're ready to tackle it.

PR #1886, opened 2017-01-30
Update source list when back and forward buttons are clicked

arthur801031:
Associated Issue: #635

### Summary of Changes
* Reload debuggee when a back/forward button is clicked. 

### Test Plan
1. Open debugger in Chrome
2. Open debuggee in Firefox
3. Go to https://devtools-html.github.io/debugger-examples/examples/todomvc/ in the debugee
4. Go to https://www.cnet.com/
5. Observe Sources
6. Click back button 
7. Observe Sources
8. Click forward button
9. Observe Sources

### DIFF
--- a/src/actions/navigation.js
+++ b/src/actions/navigation.js
@@ -2,7 +2,7 @@ const constants = require("../constants");
 const { clearSourceMaps } = require("../utils/source-map");
 const { clearDocuments } = require("../utils/source-documents");
 const { getSources } = require("../reducers/sources");
-const { timeout } = require("../utils/utils");
+const { waitForMs } = require("../utils/utils");
 const { newSources } = require("./sources");
 
 /**
@@ -28,7 +28,7 @@ function willNavigate() {
 function navigated() {
   return ({ dispatch, client, getState }) => {
     (async function() {
-      await timeout(100);
+      await waitForMs(100);
       if (getSources(getState()).size == 0) {
         const sources = await client.fetchSources();
         dispatch(newSources(sources));
--- a/src/actions/navigation.js
+++ b/src/actions/navigation.js
@@ -23,16 +23,7 @@ function willNavigate() {
  * @static
  */
 function navigated() {
-  return ({ dispatch }) => {
-    // We need to load all the sources again because they might have
-    // come from bfcache, so we won't get a `newSource` notification.
-    //
-    // TODO: This seems to be buggy on the debugger server side. When
-    // the page is loaded from bfcache, we still get sources from the
-    // *previous* page as well. For now, emulate the current debugger
-    // behavior by not showing sources loaded by bfcache.
-    // return dispatch(sources.loadSources());
-  };
+  return { type: constants.LOADSOURCES };
 }
 
 module.exports = {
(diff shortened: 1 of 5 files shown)


src/actions/navigation.js
```diff
@@ -23,15 +25,13 @@ function willNavigate() {
...
-    // TODO: This seems to be buggy on the debugger server side. When
-    // the page is loaded from bfcache, we still get sources from the
-    // *previous* page as well. For now, emulate the current debugger
-    // behavior by not showing sources loaded by bfcache.
-    // return dispatch(sources.loadSources());
+  return ({ dispatch, getState }: ThunkArgs) => {
+    setTimeout(() => {
+      if (getSources(getState()).size == 0) {
+        const threadClient = firefox.getThreadClient();
+        threadClient.getSources();
```
jasonLaster: Looking good.

1. does this fix the problem, my fear is that getSources will make one call for the sources, but it won't trigger `nweSource` by itself
2. i think we can refactor this code a bit to make it more consistent with `util/client`

https://gist.github.com/jasonLaster/de3bd2ecbacaf7518c42336e199d5137
https://gist.github.com/dde3f2cd62733746e562f9d6acd48e76
 
Goal:
1. add a new firefox client command for fetching sources
2. add a `timeout` helper so we can do it with async/await
3. call `newSources` in this action which i think we need to do

src/utils/utils.js
```diff
@@ -184,7 +184,7 @@ function updateObj<T: Object>(obj: T, fields: $Shape<T>) : T {
  * @static
  */
 function throttle(func: any, ms: number) {
-  let timeout, _this;
+  let timeout, _this; // eslint-disable-line no-shadow
```
arthur801031: I got this error, which I thought it's fine to disable it.

screen shot 2017-02-01 at 12 49 48 am

arthur801031:
Hi @jasonLaster 

I've made those changes. The PR for the refactoring is [here](https://github.com/devtools-html/devtools-core/pull/149). Thank you! :)

jasonLaster:
I'll pull in the client work soon

src/utils/utils.js
jasonLaster: hmm, maybe we want to change the function name to `waitForMs`

src/actions/navigation.js
```diff
@@ -23,15 +26,14 @@ function willNavigate() {
...
-  return ({ dispatch }) => {
-    // We need to load all the sources again because they might have
-    // come from bfcache, so we won't get a `newSource` notification.
-    //
-    // TODO: This seems to be buggy on the debugger server side. When
-    // the page is loaded from bfcache, we still get sources from the
-    // *previous* page as well. For now, emulate the current debugger
-    // behavior by not showing sources loaded by bfcache.
-    // return dispatch(sources.loadSources());
+  return ({ dispatch, client, getState }) => {
```
jasonLaster: We have a slightly nicer style here:
https://github.com/arthur801031/debugger.html/blob/update_source/src/actions/pause.js#L45-L48

Issue #1894, opened 2017-01-31
[Flow] add types to components


jasonLaster:
We'd like to support flow types in our components. This issue will track our progress and will be a good place to raise issues as they arise.

Here are some [flow] docs. The best thing to do is to open one PR per module converted to flow.

- [ ] components
  - [x] App.js
  - [ ] Editor
    - [x] Breakpoint.js
    - [x] ConditionalPanel.js
    - [x] Footer.js
    - [x] HitMarker.js
    - [x] SearchBar.js
    - [x] Tabs.js
    - [ ] index.js
  - [x] SecondaryPanes
    - [x] Breakpoints.js
    - [x] ChromeScopes.js
    - [x] CommandBar.js
    - [x] EventListeners.js
    - [x] Expressions.js
    - [x] Frames.js
    - [x] Scopes.js
    - [x] WhyPaused.js
    - [x] index.js
  - [x] SourceSearch.js
  - [x] Sources.js
  - [x] SourcesTree.js
  - [ ] WelcomeBox.js
  - [x] shared
    - [x] Accordion.js
    - [x] Autocomplete.js
    - [x] Button
    - [x] Close.js
    - [x] PaneToggle.js
    - [x] Dropdown.js
    - [x] ManagedTree.js
    - [x] ObjectInspector.js
    - [x] Rep.js
    - [x] Svg.js
    - [x] menu.js

[flow]: https://github.com/devtools-html/debugger.html/blob/master/docs/flow.md

arthur801031:
I'll take `SecondaryPanes/Breakpoints.js`.

arthur801031:
I'll take the following:
- SecondaryPanes/Scopes.js

arthur801031:
I'll take the following:
- `shared/Autocomplete.js`
- `shared/Button.js`
- `shared/Close.js`
- `shared/PaneToggle.js`
- `shared/Dropdown.js`

arthur801031:
I'll take the following:
- `shared/ObjectInspector.js`
- `shared/Rep.js`
- `shared/Svg.js`
- `shared/Menu.js`
- `Editor/index.js`

PR #1896, opened 2017-02-01
Flow editor

clarkbw:
Associated Issue: #1894 

### Summary of Changes

* Skipped the footer because it got difficult with the selector typing, I'll come back to it
* Skipped the index but I'll come back to that later unless someone else gets it
* I tried using `SyntheticClipboardEvent` but it doesn't seem to be typed correctly to use in the `doCopy` function.

### Test Plan

- [x] `yarn run flow`

### DIFF
--- a/src/components/Editor/Breakpoint.js
+++ b/src/components/Editor/Breakpoint.js
@@ -1,3 +1,4 @@
+// @flow
 const React = require("react");
 const ReactDOM = require("react-dom");
 
@@ -8,7 +9,7 @@ const Svg = require("../shared/Svg");
 const breakpointSvg = document.createElement("div");
 ReactDOM.render(Svg("breakpoint"), breakpointSvg);
 
-function makeMarker(isDisabled) {
+function makeMarker(isDisabled: boolean) {
   const bp = breakpointSvg.cloneNode(true);
   bp.className = classnames(
     "editor new-breakpoint",
@@ -20,8 +21,8 @@ function makeMarker(isDisabled) {
 
 const Breakpoint = React.createClass({
   propTypes: {
-    breakpoint: PropTypes.object,
-    editor: PropTypes.object
+    breakpoint: PropTypes.object.isRequired,
+    editor: PropTypes.object.isRequired
   },
 
   displayName: "Breakpoint",
@@ -41,7 +42,7 @@ const Breakpoint = React.createClass({
     }
   },
 
-  shouldComponentUpdate(nextProps) {
+  shouldComponentUpdate(nextProps: any) {
     return this.props.editor !== nextProps.editor ||
       this.props.breakpoint.disabled !== nextProps.breakpoint.disabled ||
       this.props.breakpoint.condition !== nextProps.breakpoint.condition;
--- a/src/components/Editor/HitMarker.js
+++ b/src/components/Editor/HitMarker.js
@@ -1,3 +1,4 @@
+// @flow
 const React = require("react");
 const { PropTypes } = React;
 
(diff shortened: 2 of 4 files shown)


src/components/Editor/Breakpoint.js
```diff
@@ -41,7 +42,7 @@ const Breakpoint = React.createClass({
     }
   },
 
-  shouldComponentUpdate(nextProps) {
+  shouldComponentUpdate(nextProps: any) {
```
wldcordeiro: Wouldn't `nextProps` here always be an object with the breakpoints and editor since they're required props?

src/components/Editor/HitMarker.js
```diff
@@ -30,7 +31,7 @@ const HitMarker = React.createClass({
     this.props.editor.addLineClass(line, "line", "hit-marker");
   },
 
-  shouldComponentUpdate(nextProps) {
+  shouldComponentUpdate(nextProps: any) {
```
wldcordeiro: Same question here about the use of `any`

src/components/Editor/Breakpoint.js
jasonLaster: we can come back to typing props, I think we'll get there soon enough

src/components/Editor/Tabs.js
```diff
@@ -260,8 +264,12 @@ const SourceTabs = React.createClass({
...
     const { selectedSource, selectSource, closeTab } = this.props;
     const filename = getFilename(source.toJS());
-    const active = selectedSource && source.get("id") == selectedSource.get("id");
-    const isPrettyCode = isPretty({ url: source.get("url") });
+    const active = selectedSource &&
+                   (source.get("id") == selectedSource.get("id"));
+    const isPrettyCode = isPretty({
+      id: source.get("id"),
+      url: source.get("url")
+    });
```
jasonLaster: Hmm, i'm surprised you can't do `isPretty(source.toJS())`

src/components/Editor/Breakpoint.js
clarkbw: Good question!  Essentially `nextProps` is the `propTypes` attribute of the class but I don't really want to write it out twice.  Here are the patterns I've seen in the wild when researching this.

Use the [flow-react-proptypes](https://github.com/brigand/babel-plugin-flow-react-proptypes) plugin.  This means you create a flow type for your prop types and don't define the `propTypes` in your react class; like this:

```js
type PropTypes = {
  breakpoint: Breakpoint,
  editor: any
};
```

And then babel would compile that type into a PropType for react.  This pattern doesn't settle well with me right now.  I wonder about our plans to move to es6 classes and how well this plugin works.  

Then of course other people just define both:

```js
type PropTypes = {
  breakpoint: Breakpoint,
  editor: any
};
const Breakpoint = React.createClass({
  propTypes: {
    breakpoint: PropTypes.object.isRequired,
    editor: PropTypes.object.isRequired
},
 shouldComponentUpdate(nextProps: PropTypes) {
 }
});
```

Which is a bit of overhead but would actually type the `propTypes` correctly.  Other than the overhead I don't think there is anything wrong with this pattern.  

My ideal pattern would be to type the `propTypes` inline.  So it would look something like this:

```js
const Breakpoint = React.createClass({
  propTypes: {
    breakpoint: PropTypes.object.isRequired: Breakpoint,
    editor: PropTypes.object.isRequired: any
},
 shouldComponentUpdate(nextProps: propTypes) {
 }
});
```

But so far in other components we haven't cracked this egg.

src/components/Editor/Tabs.js
```diff
@@ -131,10 +131,7 @@ const SourceTabs = React.createClass({
     const sourceTab = sourceTabs.find(t => t.get("id") == tab);
     const tabURLs = sourceTabs.map(thisTab => thisTab.get("url"));
     const otherTabURLs = otherTabs.map(thisTab => thisTab.get("url"));
-    const isPrettySource = isPretty({
-      id: sourceTab.get("id"),
-      url: sourceTab.get("url")
-    });
+    const isPrettySource = isPretty(sourceTab.toJS());
```
clarkbw: Much better! 🎉

Issue #1511, opened 2016-12-21
Web Test Runner


jasonLaster:
We currently run our integration tests with [Mochitest](https://github.com/devtools-html/debugger.html/blob/master/docs/mochitests.md). It has several benefits:

1. The tests are run end-to-end in Firefox and the devtools panel
2. The tests are driven with redux actions. See this [test](https://github.com/devtools-html/debugger.html/blob/master/src/test/mochitest/browser_dbg-breaking.js#L12-L16) for example.

There are three primary drawbacks with our setup:

1. The tests require a full checkout and build of firefox (this takes time)
2. It's impossible to test chrome or node, which we'd like to support
3. The tests are hard to debug. Using the browser toolbox is cool, but not the most effective tool.

The solution to this is to write a new iframe based test runner.

Tasks:
- [x] setup simple runner https://github.com/devtools-html/debugger.html/pull/1445
- [ ] run our mochitests with the runner #1844
- [ ] update CI to launch chrome and node and run the tests 

How does the runner work:
* it loads the debugger in an iframe 
* it uses mocha as the runner 
* it loads the same redux helpers, it has a couple of global functions that map the mochitest API to mocha / expect.

CC @juliandescottes

jasonLaster:
This is tracked elsewhere

PR #1899, opened 2017-02-01
WTF: step 2 Add helpers

jasonLaster:
Associated Issue: #1511 

### Summary of Changes

* pulls the helpers into discrete files
* adds a mocha and mochi helper file. I think i can switch them in webpack which will be cool...

### Test Plan

mochitests pass

### DIFF
--- a/src/test/integration/tests/pretty-print.js
+++ b/src/test/integration/tests/pretty-print.js
@@ -1,24 +1,28 @@
 
+const { waitForPaused, waitForDispatch } = require("../utils/wait");
+const { findSource, findElement } = require("../utils/shared");
+const {
+  selectSource,
+  clickElement,
+  addBreakpoint,
+  stepOver,
+  invokeInTab,
+  resume
+} = require("../utils/commands");
+
+const { assertPausedLocation } = require("../utils/assert");
+
+const { setupTestRunner, initDebugger } = require("../utils/mochi")
+
 async function prettyPrint () {
-  const {
-    initDebugger,
-    selectSource,
-    clickElement,
-    waitForDispatch,
-    findSource,
-    addBreakpoint,
-    invokeInTab,
-    waitForPaused,
-    stepOver,
-    assertPausedLocation,
-    resume,
-    findElement,
-    ok
-  } = this;
+  const { ok, is } = this;
+  const ctx = {ok, is};
+  setupTestRunner(this);
 
   const dbg = await initDebugger("doc-minified.html");
 
   await selectSource(dbg, "math.min.js");
+
   clickElement(dbg, "prettyPrintButton");
   await waitForDispatch(dbg, "TOGGLE_PRETTY_PRINT");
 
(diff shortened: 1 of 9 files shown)


src/test/integration/tests/pretty-print.js
```diff
@@ -29,9 +33,9 @@ async function prettyPrint () {
 
   invokeInTab("arithmetic");
   await waitForPaused(dbg);
-  assertPausedLocation(dbg, ppSrc, 18);
+  assertPausedLocation(dbg, ctx, ppSrc, 18);
```
jasonLaster: this doesn't feel great, but i have a feeling i can do better in the next phase

src/test/integration/utils/mochi.js
```diff
@@ -0,0 +1,52 @@
+
+function info(msg) {
+  dump(`info: ${msg}\n`);
+}
+
+var ContentTask, gBrowser, openNewTabAndToolbox, Services, EXAMPLE_URL;
```
jasonLaster: I don't love this DI style, we hopefully can get this file priviledged like `head.js` with the proper permissions

src/test/mochitest/head.js
```diff
@@ -41,6 +41,10 @@ Services.prefs.setBoolPref("devtools.debugger.new-debugger-frontend", true);
 Services.prefs.clearUserPref("devtools.debugger.tabs")
 Services.prefs.clearUserPref("devtools.debugger.pending-selected-location")
 
+this.gBrowser = gBrowser;
+this.Services = Services;
+this.EXAMPLE_URL = EXAMPLE_URL;
```
jasonLaster: my fear here is that it's a memory leak. Also, ugly... but perhaps we can do something in another pass when we're cleaning up

src/test/integration/utils/wait.js
```diff
@@ -0,0 +1,165 @@
...
+  await waitForState(dbg, state => {
+    const pause = dbg.selectors.getPause(state);
+    // Make sure we have the paused state.
+    if (!pause) {
+      return false;
+    }
+    // Make sure the source text is completely loaded for the
+    // source we are paused in.
+    const sourceId = pause.getIn(["frame", "location", "sourceId"]);
+    const sourceText = dbg.selectors.getSourceText(dbg.getState(), sourceId);
```
nchevobbe: is `dbg.getState()` different from `state` ?

src/test/integration/utils/wait.js
```diff
@@ -0,0 +1,165 @@
...
+      },
+      run: (dispatch, getState, action) => {
+        resolve(action);
+      }
+    });
+  });
+}
+
+// Wait until an action of `type` is dispatched. This is different
+// then `_afterDispatchDone` because it doesn't wait for async actions
```
nchevobbe: nit: s/then/than

src/test/integration/utils/wait.js
jasonLaster: probably not... but i don't want to mess with it because we are only using this in one test now...

PR #1901, opened 2017-02-01
Organize Panel Assets

jasonLaster:
### Summary of Changes

* moves assets to a panels directory
* this fixes an issue where the index was CPd to `assets/build`

### Test Plan

mochietests


docs/local-development.md
```diff
@@ -238,21 +238,21 @@ The Debugger supports two types of internationalization RTL (right to left) layo
 
 #### L10N
 
-[L10N](https://github.com/devtools-html/devtools-core/blob/master/packages/devtools-launchpad/src/utils/L10N.js) is a global module with two methods `getStr` and `getFormatStr`.
+[L10N][l10n] is a global module with two methods `getStr` and `getFormatStr`.
```
clarkbw: Markdown pro-tip! You can just define `[L10N]` at the bottom instead of creating the lowercase `[l10n]` alias and then here simply use `[L10N]` without anything else.

Issue #1663, opened 2017-01-10
Highlight line longer after jumping to file from Console


jnachtigall:
STR:
1. Click on a line in the console
2. Jumps to Debugger

AR:
3. The line is highlighted for maybe one second, then immediately vanishes. Usually in such a situation you do not know 'where' you are, so you need to orientate first and find/search where to look at - until the highlight is gone. I had this "Where is this line?" situatuion a few times now ;)

![capture_2](https://cloud.githubusercontent.com/assets/3307404/21810185/ff580476-d74a-11e6-880a-1181a022fcab.gif)

ER:
Please make the hightlight last for 3-5 seconds.

Tested with Debugger builtin Nightly 53.0a1 (2017-01-10) (64-bit) Windows 10.

jasonLaster:
hmm, that makes sense. so basically a long highlight which we can do with a `selectLocation param`

arthur801031:
I'll take a stab at this :)

arthur801031:
Feel free to claim this. I couldn't get the mochitests setup to work :(

clarkbw:
> Feel free to claim this. I couldn't get the mochitests setup to work :(

hmm, we need to fix that.  Sorry @arthur801031 I hope you didn't waste too much time. 😠

arthur801031:
I've resolved the issue with mochitests setup. I'll try to take another crack at this 😂

jasonLaster:
Very cool

clarkbw:
fixed in #1904

PR #1904, opened 2017-02-02
Highlight line longer after jumping to file from Console

arthur801031:
Associated Issue: #1663 

### Summary of Changes

* Change the animation time from 1.5s to 4s.

### 4s Animation (animation-timing-function: normal):
rs4oktdqbx

### 1.5s Animation:
ff580476-d74a-11e6-880a-1181a022fcab
(GIF by @jnachtigall )

### DIFF
--- a/src/components/Editor/Editor.css
+++ b/src/components/Editor/Editor.css
@@ -98,7 +98,7 @@ html[dir="rtl"] .editor-mount {
 }
 
 .highlight-line .CodeMirror-line {
-  animation: fade-highlight-out 4s normal forwards;
+  animation: fade-highlight-out 3s normal forwards;
 }
 
 @keyframes fade-highlight-out {
--- a/src/components/Editor/Editor.css
+++ b/src/components/Editor/Editor.css
@@ -98,7 +98,7 @@ html[dir="rtl"] .editor-mount {
 }
 
 .highlight-line .CodeMirror-line {
-  animation: fade-highlight-out 1.5s normal forwards;
+  animation: fade-highlight-out 4s normal forwards;
 }
 
 @keyframes fade-highlight-out {


arthur801031:
Hi @jasonLaster

Please check. Thank you! 🙂

arthur801031:
I think with `animation-timing-function: ease-in` it's easier to see the highlighted line because a big chunk of 4s animation has a darker effect than the 4s animation with`animation-timing-function: normal`. Let me know which one you like more :)

### 4s Animation (animation-timing-function: ease-in):
4ox8kgfu1d

src/components/Editor/Editor.css
```diff
@@ -98,7 +98,7 @@ html[dir="rtl"] .editor-mount {
 }
 
 .highlight-line .CodeMirror-line {
-  animation: fade-highlight-out 1.5s normal forwards;
+  animation: fade-highlight-out 4s normal forwards;
```
jasonLaster: hmm, i think we *only* want to do this when we're navigating from the console.

Perhaps:

```css
.highlight-line.ong .CodeMirror-line  { 
  animation: fade-highlight-out 3s normal forwards;
}
```

src/components/Editor/Editor.css
```diff
@@ -98,7 +98,7 @@ html[dir="rtl"] .editor-mount {
 }
 
 .highlight-line .CodeMirror-line {
-  animation: fade-highlight-out 1.5s normal forwards;
+  animation: fade-highlight-out 3s normal forwards;
```
arthur801031: It seems a line would only be applied with the styles of `.highlight-line .CodeMirror-line` when navigating from the console. The navigation triggers the function at https://github.com/devtools-html/debugger.html/blob/master/src/components/Editor/index.js#L599.

The line is first cleared and highlighted under these 2 conditions: https://github.com/devtools-html/debugger.html/blob/master/src/components/Editor/index.js#L333-L349

jasonLaster: hmm - i think this also happens when you click on a breakpoint in the breakpoint list or you select a frame in the call stack...

Is that correct?

arthur801031: Hi @jasonLaster !

1. Clicking on a breakpoint in the breakpoint list would trigger `highlightLine()`, but the function returns right away, and it doesn't execute `this.editor.codeMirror.addLineClass(line - 1, "line", "highlight-line");`, which is highlighting the line.
yvxe4w1ual




2. Selecting a frame in the call stack would trigger `highlightLine()` and clear the line, but it doesn't call `this.editor.codeMirror.addLineClass(line - 1, "line", "highlight-line");`, so we won't see the line being applied with the styles of `.highlight-line .CodeMirror-line`.
2ped7795ob

jasonLaster:
@arthur801031 this PR is probably good, but the internals of `highlightLine` are confusing and I'd like to get a better understanding of it before we move forward.

Perhaps we can add more integration tests here.

nchevobbe:
@arthur801031 good job and good writeup ! 
I think we could do better with the animation though. Would you mind trying to : 
- keep it 4s long
- but start lowering down the opacity at 3s

Because right now the animation feels like it is in slow motion, and thus can trick the user to think the debugger "janks" and can't keep up.

I'd really appreciate if you could try that and put some GIF here :) Thanks ! 🙆

PR #1908, opened 2017-02-02
client docs

jasonLaster:
@bomsy mentioned we should have docs for describing the relationship with the server.

This is a start


bomsy:
Nice! Thanks @jasonLaster

jasonLaster:
@bomsy cool. Lets add to this doc over time!

Issue #1785, opened 2017-01-22
[Flow] type utility files


jasonLaster:
Lets add types to our utils:

Here are some [docs](https://github.com/devtools-html/debugger.html/blob/master/docs/local-development.md#flow) on using flow:

Files:
- [x] DevToolsUtils.js
- [x] assert.js
- [x] client.js
- [x] create-store.js
- [x] defer.js
- [x] fromJS.js
- [x] log.js
- [x] makeRecord.js
- [x] path.js
- [x] pause.js
- [x] prefs.js
- [x] pretty-print-worker.js
- [x] pretty-print.js
- [x] redux middleware
  - [x] history.js
  - [x] log.js
  - [x] promise.js
  - [x] thunk.js
  - [x] wait-service.js
- [x] redux-devtools.js
- [ ] editor
  - [ ] index.js
  - [x] source-documents.js
  - [x] source-editor.js 
  - [x] source-search.js
  - [x] build-query.js
- [x] source-map-util.js
- [x] source-map-worker.js
- [x] source-map.js
- [x] source.js
- [x] sources-tree.js
- [x] task.js
- [x] text.js
- [x] utils.js

### Progress

Generated by running `yarn run flow-utils`

![screen shot 2017-01-22 at 10 35 06 am](https://cloud.githubusercontent.com/assets/254562/22183529/8c00a076-e08e-11e6-8d53-845fdc80f834.png)

ogryzek:
I'm starting on the `editor.js`

leimd:
working on promise.js

skhanker:
i'll take log.js

flexjames:
create-store.js has flow implemented.

mayank-gupta:
utils.js is done

mayank-gupta:
sources-tree.js is done

arthur801031:
I'll take pretty-print.js.

arthur801031:
I'll take source-documents.js.

jasonLaster:
I'll grab wait-service and history

jasonLaster:
Here are the remaining utils that could use `flow`

```
 src/actions/navigation.js         | 11 +++++++++--
 src/constants.js                  |  6 ------
 src/feature.js                    |  2 ++
 src/panel.js                      |  0
 src/reducers/index.js             |  4 +---
 src/utils/DevToolsUtils.js        |  2 ++
 src/utils/clipboard.js            |  2 ++
[...]

PR #1911, opened 2017-02-03
Flow

flexjames:
Associated Issue: #1785

Added Flow to client.js & dehydrate-state.js.


src/utils/client.js
```diff
@@ -1,7 +1,9 @@
+// @flow
+
 const { firefox } = require("devtools-client-adapters");
 const { prefs } = require("./prefs");
 
-async function onFirefoxConnect(actions) {
+async function onFirefoxConnect(actions: Object) {
```
wldcordeiro: I'm not sure if typing these as `Object` really buys us much, could we use a proper type that maps to the actions the client receives @jasonLaster

src/utils/dehydrate-state.js
```diff
@@ -1,9 +1,11 @@
+// @flow
+
 const I = require("immutable");
 const { fromJS } = I;
-const SourcesState = require("../reducers/sources").State;
+const SourcesState: any = require("../reducers/sources").State;
```
jasonLaster: we can use `SourcesState` that's exported in the reducer

src/utils/client.js
```diff
@@ -33,7 +35,7 @@ async function onFirefoxConnect(actions) {
   }
 }
 
-async function onConnect(connection, actions) {
+async function onConnect(connection: Object, actions: Object) {
```
jasonLaster: hmm, we can add a local type:

```js
import type { Tab } from "../types"
type Connection = { tab: Tab }
```

src/utils/client.js
jasonLaster: true,, we can import Actions

```js
import type { Actions } from "../actions/types";
```

src/utils/client.js
jasonLaster: Hmm, this isn't used here. so we can ignore it for the time being.

jasonLaster: Also, we don't need to type `actions` because it's the object that holds all of the actions, which is currently untyped

PR #1913, opened 2017-02-03
simplified getting setup

jasonLaster:
This dramatically simplifies the setup guide based on the amazing work @bomsy did with the launch buttons. 

It also adds helpful information about **node 7** and links people to install guides for yarn. I found that including `brew install yarn` added more confusion tonight. I think it's better to bring people to yarn install and let them sort it out and then come back... 

Curious to see what people think

### DIFF
--- a/docs/getting-setup.md
+++ b/docs/getting-setup.md
@@ -38,7 +38,7 @@ tab to debug.
 
 ![][launchpad-screenshot]
 
-Congradulations! You're now up and running. :sweat_smile:
+Congratulations! You're now up and running. :sweat_smile:
 
 *What should I do if I get an error?*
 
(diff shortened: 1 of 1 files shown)


docs/getting-setup.md
```diff
@@ -38,42 +17,32 @@ yarn install
...
-*What does this do?*
-This command starts a [development server][dev-server].
+Open `http://localhost:8000` in any browser and launch
+Firefox or Chrome. You should now be able to select a
+tab to debug.
+
+![][launchpad-screenshot]
 
-### Step 5. Open the Debugger
+Congradulations! You're now up and running. :sweat_smile:
```
wldcordeiro: Congratulations

PR #1914, opened 2017-02-03
Type more utilities

jasonLaster:
Associated Issue: #1785 

### Summary of Changes

This makes progress on some of the remaining util files. After this there are only a handful...

 * src/utils/prefs.js                    
 * src/utils/pretty-print-worker.js      
 * src/utils/redux-devtools.js           
 * src/utils/redux/middleware/promise.js 
 * src/utils/redux/middleware/thunk.js   

There are still some issues here, but it's most of the way there.
https://gist.github.com/e16ffd76edb9399d4ae00d5f4a9af139

CC @jbhoosreddy, @clarkbw you might be interested in taking a look at some of the redux stuff.

### DIFF
--- a/src/utils/pretty-print-worker.js
+++ b/src/utils/pretty-print-worker.js
@@ -1,5 +1,27 @@
+// @flow
+
 const prettyFast = require("pretty-fast");
 const assert = require("./assert");
+
+type Mappings = {
+  _array: Mapping[]
+}
+
+type Mapping = {
+  originalLine: number,
+  originalColumn: number,
+  source?: string,
+  generatedLine?: number,
+  generatedColumn?: number,
+  name?: string
+}
+
+type InvertedMapping = {
+  generated: Object,
+  source?: any,
+  original?: any,
+  name?: string
+}
 
 function prettyPrint({ url, indent, source }) {
   try {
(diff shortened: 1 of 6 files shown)


src/utils/redux/middleware/promise.js
```diff
@@ -6,15 +8,17 @@ const defer = require("../../defer");
 const { entries, toObject } = require("../../utils");
 const { executeSoon } = require("../../DevToolsUtils");
 
+import type { ThunkArgs, ActionType } from "../../../actions/types";
+
 const PROMISE = exports.PROMISE = "@@dispatch/promise";
 let seqIdVal = 1;
```
clarkbw: You could type this as a `number`

src/utils/redux/middleware/promise.js
```diff
@@ -6,15 +8,17 @@ const defer = require("../../defer");
...
 
-function seqIdGen() {
+function seqIdGen(): number {
```
clarkbw: And then you wouldn't need to type the function return here.

clarkbw:
Still has some `flow` errors.

src/utils/pretty-print-worker.js
```diff
@@ -13,13 +35,13 @@ function prettyPrint({ url, indent, source }) {
       mappings: prettified.map._mappings
     };
   } catch (e) {
-    return new Error(`${e.message}\n${e.stack}`);
+    throw new Error(`${e.message}\n${e.stack}`);
```
clarkbw: This is an interesting change!

PR #1916, opened 2017-02-03
Fixing crashes on sparse local storage.

yurydelendik:
Associated Issue: #1375

### Summary of Changes

* Fixing grip to not have the length property

### Test Plan

Follow STR in #1375 and use https://devtools-html.github.io/debugger-examples/examples/localstorage-keys.html

- [ ] probably write a unit test for it in `actions/tests`

### DIFF
--- a/src/utils/fromJS.js
+++ b/src/utils/fromJS.js
@@ -41,6 +41,26 @@ function fromJS(value: any) : any {
     return value;
   }
 
+  // Immutable.Seq recognizes any object with numeric length property as
+  // an array. Replacing array-like grip with something that does not behave
+  // length property.
+  if (typeof value.length === "number" && ("items" in value) &&
+      value.kind === "ArrayLike") {
+    value = Object.create(value, {
+      length: {
+        get: () => void 0,
+        enumerable: true,
+        configurable: true
+      },
+      size: {
+        value: value.length,
+        writable: true,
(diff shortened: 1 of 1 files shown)


src/utils/fromJS.js
```diff
@@ -41,6 +41,26 @@ function fromJS(value: any) : any {
     return value;
   }
 
+  // Immutable.Seq recognizes any object with numeric length property as
+  // an array. Replacing array-like grip with something that does not behave
+  // length property.
+  if (typeof value.length === "number" && ("items" in value) &&
+      value.kind === "ArrayLike") {
+    value = Object.create(value, {
```
wldcordeiro:
```js
// Immutable.Seq recognizes any object with numeric length property as
  // an array. Replacing array-like grip with something that does not behave
  // length property.
  if (typeof value.length === "number" && ("items" in value) &&
      value.kind === "ArrayLike") {
    value = Object.create(value, {
      length: {
        get: () => void 0,
        enumerable: true,
        configurable: true
      },
      size: {
        value: value.length,
        writable: true,
        enumerable: true,
        configurable: true
      },
    });
}
```

jasonLaster:
I'm closing this and opening a slightly more focused fix for `fromJS`. We'll also update ObjectInspector to better handle large objects/arrays

Issue #1909, opened 2017-02-02
[CallStack] cleanup selected secondarytext


jasonLaster:
Currently the secondary text is an ugly grey when it is selected. Lets choose a different color when the frame is selected.

We should also check to make sure that it looks good in other [themes] as well:

![](http://g.recordit.co/WN3nTJB9Uh.gif)

[themes]:https://github.com/devtools-html/debugger.html/blob/master/docs/local-development.md#themes

jasonLaster:
```diff

diff --git a/src/components/SecondaryPanes/Frames.css b/src/components/SecondaryPanes/Frames.css
index 86d29bd..156dfc8 100644
--- a/src/components/SecondaryPanes/Frames.css
+++ b/src/components/SecondaryPanes/Frames.css
@@ -12,11 +12,11 @@
   justify-content: space-between;
 }

-/* Style the focused call frame like so:
+ /*Style the focused call frame like so:*/
 .frames ul li:focus {
-  border: 3px solid red;
+  color: white;
 }
-*/
+

```

PR #1926, opened 2017-02-03
[CallStack] cleanup selected secondarytext

princiya:
Associated Issue: #1909 

### Summary of Changes

* secondary text in Call Stack changed to white color for all three themes

### Before
before

### After
after


wldcordeiro:
@princiya Thanks for the PR! Great to see little UX issues getting squashed to make this even better. 😄

Issue #1919, opened 2017-02-03
[WebTestRunner] All the tests


jasonLaster:
Now that we have a working Test Runner on `localhost:8000/integration` we should port the remaining tests and see if we can make some improvements:

### Steps for converting a test

1. create a new `integrations/test/<file>.js`
2. add it to `tests/runner.js`
3. export it in `tests/tests.js` and import it in the original `mochitest/<file>.js`
4. refactor the test:
  * make it async, change yields to await
  * setup `ctx`, add it to asserts
  * add `dbg` to invokeInTab
  * you need to add a source in `initDebugger` to pause on

### Remaining Tests to port
- [ ] breaking-from-console.js
- [ ] breakpoints.js
- [ ] console.js

### Tests to fix
- [ ] breaking-from-console.js
- [ ] console.js
- [ ] searching.js
- [ ] sourcemaps-bogus.js
- [ ] sourcemaps2.js
- [ ] keyboard-shortcuts.js #2262
- [ ] keyboard_navigation.js #2262

### Tests to add 
- [ ] editor
  - [ ] reveal source
  - [ ] "add watch expression"

### Improvements
- [ ] run on CI
- [ ] prompt user to launch a browser
- [ ] make chrome and node testing better
- [ ] improve reporter
- [ ] check for other running debuggers... it will be flakey when another debugger is connected to the tab

codehag:
I am closing this for now as we have removed the WTR

PR #1929, opened 2017-02-04
[WTR] Add breaking test

jasonLaster:
Associated Issue: #1919 

### Summary of Changes

* Adds the breaking test
* update some helpers

### DIFF
--- a/src/test/integration/runner.js
+++ b/src/test/integration/runner.js
@@ -5,18 +5,21 @@ const { prefs } = require("../../utils/prefs")
 const prettyPrint = require("./tests/pretty-print")
 const breaking = require("./tests/breaking")
 
-function ok(expected) {
+window.ok = function ok(expected) {
   expect(expected).to.be.truthy
 }
 
-function is(expected, actual) {
+window.is = function is(expected, actual) {
   expect(expected).to.equal(actual)
+}
+
+window.info = function info(msg) {
+  console.log(`info: ${msg}\n`);
 }
 
 mocha.setup({ timeout: 20000, ui: 'bdd' });
 
 describe("Tests", () => {
-
   beforeEach(() => {
     prefs.pendingSelectedLocation = {};
     prefs.tabs = [];
--- a/src/test/integration/runner.js
+++ b/src/test/integration/runner.js
@@ -3,6 +3,7 @@ const expect = require("expect.js");
 
 const { prefs } = require("../../utils/prefs")
 const prettyPrint = require("./tests/pretty-print")
+const breaking = require("./tests/breaking")
 
 function ok(expected) {
   expect(expected).to.be.truthy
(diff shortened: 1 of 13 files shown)


src/test/integration/runner.js
```diff
@@ -3,6 +3,7 @@ const expect = require("expect.js");
 
 const { prefs } = require("../../utils/prefs")
 const prettyPrint = require("./tests/pretty-print")
+const breaking = require("./tests/breaking")
 
 function ok(expected) {
```
jasonLaster: we should move `info` here.

src/test/integration/utils/commands.js
```diff
@@ -1,4 +1,9 @@
-const { info } = require("./shared");
+// const { info } = require("./shared");
```
nchevobbe: do we need to have this as a comment ?

src/test/integration/utils/mocha.js
```diff
@@ -61,7 +70,10 @@ async function createIframe() {
   container.innerHTML = "";
   container.appendChild(iframe);
 
-  await waitForTime(2000);
+  await waitForLoad(iframe);
```
nchevobbe: I think we could have a race condition here, if onload is fired before we hit this line (which could be the case since onload is fired when we add the iframe to the document).
I don't know if it's very likely though, but I'd feel safer if we could declare declare the onload handler before adding the iframe to the container.

src/test/integration/utils/mocha.js
```diff
@@ -72,28 +84,30 @@ async function navigateToTab(dbg) {
...
   dbg.win.location = `/?firefox-tab=${tabId}`;
-  await waitForTime(3000);
+  return waitForElement(dbg, ".debugger");
+
 }
 
 async function navigate(dbg, url) {
   dbg.win.client.navigate(`${url}`);
-  return waitForTime(3000);
+  return waitForDispatch(dbg, "RESUME");
```
nchevobbe: nit: if `"RESUME"` is in a constant file, it would be nice to get it from there

src/test/integration/utils/wait.js
```diff
@@ -133,6 +146,17 @@ async function waitForElement(dbg, selector) {
   return waitUntil(() => findElementWithSelector(dbg, selector))
 }
 
+async function waitUntil(predicate) {
```
nchevobbe: nit: it would be nice to have the interval as an argument of waitUntil

```js
function waitUntil(predicate, interval = 20) {
```

src/test/mochitest/browser_dbg-breaking.js
```diff
@@ -3,30 +3,13 @@
...
-  assertPausedLocation(dbg, source, 5);
+const {
+  breaking,
+  setupTestRunner
+} = require("devtools/client/debugger/new/integration-tests");
+
+add_task(function*() {
+  setupTestRunner(this);
+  const { ok, is } = this;
+  yield breaking({ ok, is });
```
nchevobbe: 👍

Issue #962, opened 2016-10-20
Selected States


helenvholmes:
I think we're like to split up the idea of a "selected state" and a "focused state". Currently we use a blue background with white text for both of these states:

<img width="375" alt="screen shot 2016-10-20 at 12 01 23 pm" src="https://cloud.githubusercontent.com/assets/1720093/19567745/eedda750-96bc-11e6-9307-765bcbad9c09.png">

Instead of this, I think we want to use that blue background if a row is `:focus`ed, and `var(--theme-selection-color)` on backgrounds for selected (but not currently focused) rows.

jasonLaster:
Here's the work we did:

```diff
diff --git a/public/js/components/Frames.css b/public/js/components/Frames.css
index c18faaf..4b3b212 100644
--- a/public/js/components/Frames.css
+++ b/public/js/components/Frames.css
@@ -11,11 +11,14 @@
   overflow: hidden;
 }
 
-/* Style the focused call frame like so:
+/* Style the focused call frame like so: */
 .frames ul li:focus {
-  border: 3px solid red;
+  border: none;
+  outline: 0;
 }
-*/
+
+
+
 
 .frames ul li * {
   -moz-user-select: none;
@@ -41,10 +44,26 @@
 
 .frames ul li.selected,
 .frames ul li.selected .location {
+
+  background-color: slateblue;
+
+  /*
   background-color: var(--theme-selection-background);
   color: white;
+  */
 }
 
+.frames ul li.selected:focus,
+.frames ul li.selected:focus .location {
+  background: olive;
+}
+
+.frames ul li:focus,
+.frames ul li:focus .location {
+  background: red;
+}
+
+
```

SezginEge:
If nobody is working on this, I can tackle it.

jasonLaster:
@SezginEge I think this is a great one. was just thinking about it earlier today!

clarkbw:
hey @SezginEge how's this going?

SezginEge:
@clarkbw I could not work on it yet :/ I will try to send a pr today.

jasonLaster:
Thanks @SezginEge. This issue might be vague, so feel free to ask questions in channel if you have questions about styling

PR #1934, opened 2017-02-04
Selected and focused state splitted.

SezginEge:
Associated Issue: #962

### Summary of Changes

* Focused and selected state splitted up.

states

### DIFF
--- a/src/components/SecondaryPanes/Frames.css
+++ b/src/components/SecondaryPanes/Frames.css
@@ -47,17 +47,23 @@
   margin-right: 1em;
 }
 
-.frames ul li:hover,
+.frames ul li:hover {
+  background-color: rgba(236, 236, 236,0.6);
+}
+
+:root.theme-light .frames ul li:hover .location {
+  color: #49444c;
+}
+
 .frames ul li.selected {
   background-color: var(--theme-selection-background);
   color: white;
 }
 
 .frames ul li.selected:focus {
-  background-color: blue;
+  background-color: #4676b7;
 }
 
-:root.theme-light .frames ul li:hover .location,
 :root.theme-light .frames ul li.selected .location,
 :root.theme-firebug .frames ul li:hover .location,
 :root.theme-firebug .frames ul li.selected .location,
--- a/src/components/SecondaryPanes/Frames.css
+++ b/src/components/SecondaryPanes/Frames.css
@@ -53,6 +53,10 @@
   color: white;
 }
 
+.frames ul li.selected:focus {
+  background-color: blue;
+}
+
 :root.theme-light .frames ul li:hover .location,
 :root.theme-light .frames ul li.selected .location,
 :root.theme-firebug .frames ul li:hover .location,
--- a/src/components/SecondaryPanes/Frames.css
+++ b/src/components/SecondaryPanes/Frames.css
@@ -47,12 +47,14 @@
   margin-right: 1em;
 }
 
-.frames ul li:hover {
+.frames ul li:hover,
+.frames ul li:focus {
    background-color: var(--theme-toolbar-background-alt);
+   outline: none;
 }
 
-:root.theme-light .frames ul li:focus,
-:root.theme-firebug .frames ul li:focus {
+:root.theme-light .frames ul li.selected:focus,
+:root.theme-firebug .frames ul li.selected:focus {
   background-color: #4676b7;
   outline: none;
 }
@@ -68,7 +70,7 @@
   color: white;
 }
 
-:root.theme-dark .frames ul li:focus {
+:root.theme-dark .frames ul li.selected:focus {
   background-color: #314c86; 
   outline: none;
 }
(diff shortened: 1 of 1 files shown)


SezginEge:
@jasonLaster @helenvholmes For now, I applied changes just for light theme. What do you think? I will apply changes to other themes according to your opinions.

src/components/SecondaryPanes/Frames.css
```diff
@@ -47,13 +47,23 @@
   margin-right: 1em;
 }
 
-.frames ul li:hover,
+.frames ul li:hover {
+  background-color: rgba(236, 236, 236,0.6);
```
clarkbw: We could call this something like: `--theme-selectable-background-hover` which should indicate that is a select-able object.

src/components/SecondaryPanes/Frames.css
```diff
@@ -47,13 +47,23 @@
...
+}
+
+:root.theme-light .frames ul li:hover .location {
+  color: #49444c;
```
clarkbw: If we follow that pattern, this becomes something like `--theme-selectable-color-hover`

src/components/SecondaryPanes/Frames.css
```diff
@@ -47,13 +47,23 @@
...
+}
+
 .frames ul li.selected {
   background-color: var(--theme-selection-background);
   color: white;
 }
 
-:root.theme-light .frames ul li:hover .location,
+.frames ul li.selected:focus {
+  background-color: #4676b7;
```
clarkbw: And this is something like: `--theme-selection-focused-background`

clarkbw:
So I like the light theme colors, we just need the `dark` and `firebug` themes.  And then we can move it all over to variables.

SezginEge:
@clarkbw Thanks for feedback 👍 I will send a pr as soon as possible.

jasonLaster:
Hey @SezginEge, @clarkbw. I just took a look and am favoring a more soft touch:

1. re-use the hover color, which has a very similar meaning... focus is on and we can select another frame
2. removing the outline as we don't need it here


here's  a patch so you can see what i mean:

```diff


diff --git a/src/components/SecondaryPanes/Frames.css b/src/components/SecondaryPanes/Frames.css
index 6cf455b..fe8b9b5 100644
--- a/src/components/SecondaryPanes/Frames.css
+++ b/src/components/SecondaryPanes/Frames.css
@@ -48,22 +48,23 @@
 }

 .frames ul li:hover {
-  background-color: rgba(236, 236, 236,0.6);
+  background-color: var(--theme-toolbar-background-alt);
 }

 :root.theme-light .frames ul li:hover .location {
   color: #49444c;
 }

+.frames ul li:focus {
+  background-color: var(--theme-toolbar-background-alt);
+  outline: none;
+}
+
 .frames ul li.selected {
   background-color: var(--theme-selection-background);
   color: white;
 }

-.frames ul li.selected:focus {
-  background-color: #4676b7;
-}
-
 :root.theme-light .frames ul li.selected .location,
 :root.theme-firebug .frames ul li:hover .location,
 :root.theme-firebug .frames ul li.selected .location,
```

Curious what you think. We should of course test on all themes, which is easy now with the settings pane.

SezginEge:
I am agree with you about remove outline. :+1:

Today I am planning to work on changes and send a new demo gif.

SezginEge:
@clarkbw @jasonLaster I applied some changes for dark/firebug theme.  

Light: 
light

Dark:
dark

Firebug:
firebug

If it looks good to you, I will create theme variables and use them.

jasonLaster:
@SezginEge are you formatting non-selected focused frames?

jasonLaster:
From a conversation offline:

|||
|---|---|
|`.selected` | doesn't change |
|`:hover` | doesn't change |
|`:focus` | same as hoverr |
| `.selected:focus` | slightly darker than `.selected` |

selected-states

SezginEge:
Recent changes: 

|Light: |Dark:|Firebug|
|--|--|--|
|light|dark|firebug|

I think firebug theme needs UI tweak. I cant any difference between .frames ul li:nth-of-type(2n) background color and hover background color. What do you think?

jasonLaster:
Thanks. In that case we should remove the background color for even or odd tires and use a light border to divide the frames

jasonLaster:
Hmm, actually im not sure. We want to be consistent with the source tree. I can look at this a bit more in a couple hours

SezginEge:
Hey @jasonLaster ! How's this going? Do you have any suggestion?

jasonLaster:
I pushed a couple small tweaks to the branch to simplify how we're doing the colors. The tweaks, improve the outline and focus state, which is the smallest possible change that fixes the issue.

The colors you chose for selected state are great and I'd like to revisit them when we look at selected states for the source tree and command bar.

While testing I noticed that we don't select the frame when you press `<enter>` on on the frame which would really help the accessibility of the frames component. I'll add an issue for this soon.

SezginEge:
@jasonLaster @wldcordeiro Thank you for your time and feedback.

Issue #1352, opened 2016-11-29
[Source Tree] misses a filename


jasonLaster:
https://github.com/devtools-html/debugger.html/issues/1098 listed some inconsistencies between chrome's source tree and firefox's.

One thing I found was that is that if the file name has a space and or is missing an extension, our algorithm thinks its a directory.

**Steps to reproduce:**
* go to localhost:8000
* in another tab go to localhost:8000 and debug the first tab (basically start inception)
* expand the webpack domain and observe the `webpack/bootstrap 5aa234` dir

**patch**

```diff
diff --git a/src/utils/sources-tree.js b/src/utils/sources-tree.js
index 9c85739..2c1adba 100644
--- a/src/utils/sources-tree.js
+++ b/src/utils/sources-tree.js
@@ -142,6 +142,8 @@ function addToTree(tree: any, source: TmpSource) {
   url.path = decodeURIComponent(url.path);

   const parts = url.path.split("/").filter(p => p !== "");
+
+  // this is should be smarter...
   const isDir = (parts.length === 0 ||
                  parts[parts.length - 1].indexOf(".") === -1);
   parts.unshift(url.group);
@@ -183,6 +185,7 @@ function addToTree(tree: any, source: TmpSource) {

   // Overwrite the contents of the final node to store the source
   // there.
+  // => this is likely where the problem is ...
   if (isDir) {
     subtree.contents.unshift(createNode("(index)", source.get("url"), source));
   } else {
```

### Our Debugger
![screen shot 2016-11-29 at 4 29 25 pm](https://cloud.githubusercontent.com/assets/254562/20729676/1086fff6-b651-11e6-8abb-857cef9fb7e6.png)

### Chrome Debugger
![screen shot 2016-11-29 at 4 30 06 pm](https://cloud.githubusercontent.com/assets/254562/20729698/253e3fae-b651-11e6-82ea-b6b335a3c1b7.png)

CC @amitzur  interested?

jasonLaster:
@amitzur is working on this

jasonLaster:
Hey @amitzur have you had a chance to look at this?
Are you blocked by anything that I can help with?

amitzur:
Actually I meant to surface it before. I think the crux of the matter is how isDir is decided.
Unless I got something wrong, when going over the path, if current part is the last part, it should not be a directory unless it's the actual html file of the frame. Is there a way to detect that? Maybe by accessing the location of the associated frame/window?
[...]

PR #1936, opened 2017-02-04
Change isDirectory logic for source-tree

irfanhudda:
Associated Issue: #1352 

### Summary of changes

* Change `isDirectory` function
* Edit test

issue

The assumption here is that all urls point to files.
I'm completely sure about changes I made here. These changes seem to resolve the issue.

Let me know what you think.


jasonLaster:
Thanks. This is a great fix!

Issue #1661, opened 2017-01-10
[PrettyPrint] option not available for some .min files


clarkbw:
[bug 1320342](https://bugzilla.mozilla.org/show_bug.cgi?id=1320342) provides these steps to reproduce.

* Go to https://www.brain.fm
* Open `jquery.min.js`
* 🤕 No option to pretty print the file is available

clarkbw:
I just tracked this down to here: https://github.com/devtools-html/debugger.html/blob/master/src/utils/editor.js#L8

In the example above the `jquery.min.js` file includes a source map and therefore shows a `jquery.js` (pretty version) in the same directory.  The `shouldShowPrettyPrint` uses the presence of a source map to indicate that pretty print isn't available. 

@jasonLaster it looks like this came from #959 and I'm not sure what we do here.  Direct people to the source mapped version?

I found an interesting case as well where the source map file wasn't found (404) and so there's no way to view a pretty version of that minified file.  Perhaps there's a way of knowing that we successfully found the source mapped version?

clarkbw:
If we get a 404 on a source map file we should clear the `sourceMapURL` field which should allow for pretty printing again.

clarkbw:
Also I wonder if we should use the bottom bar to indicate a "Source Mapped Version" so people can see that there is a source map and click that to jump to it.  I know we have this in a context menu for jumping to lines but I think it would be good to surface the general jump option.

clarkbw:
Did some more digging.  

I believe if you wanted to tackle this you would need to start here:
https://github.com/clarkbw/debugger.html/blob/master/src/utils/source-map-worker.js#L119-L121

```js
  // Fire off the request, set it in the cache, and return it.
  // Suppress any errors and just return null (ignores bogus
  // sourcemaps).
  const req = _resolveAndFetch(generatedSource); // drop the catch .catch(() => null);
  sourceMapRequests.set(generatedSource.id, req);
  return req;
```

Now the fetch error will be passed back to the caller.

Then handle the passed error somewhere like here:
https://github.com/clarkbw/debugger.html/blob/master/src/actions/sources.js#L75-L79
```js
    const urls = await getOriginalURLs(generatedSource);
    if (!urls) {
      // If this source doesn't have a sourcemap dispatch event to remove the sourceMapURL from the source
      dispatch({
        type: constants.SOURCE_MAP_MISSING,
        id: generatedSource.id
      });
      return;
    }
```

bomsy:
[...]

PR #1937, opened 2017-02-04
fix pretty print sourcemap files

bomsy:
Associated Issue: #1661
- If no source map files are found, we enable  minified file for pretty printing.

### Summary of Changes

* Add new actions and reducers for missing source maps


jasonLaster:
Hmm. Maybe add a test for this in actions/test/sources?

jasonLaster:
Hmm, @bomsy i'd like to see if we can avoid adding a new action if we can.

Here's an opinionated refactor where the `newSource` action checks to see if it can load original sources. If not, then it clears it's sources sourceMapURL

https://gist.github.com/jasonLaster/e24b8955d1605304f497520bf5b201ad

bomsy:
Good point! Cool will play with this and see.

clarkbw:
Anything I can do to help out here @bomsy?

bomsy:
Hey yeah ... made the changes @jasonLaster suggested, it broke a couple of tests. Would love to pair with you or @jasonLaster to fix the tests. Finding them a bit tricky.

jasonLaster:
I'm closing this now so we can focus on the march 6th release. Lets pick it up afterwards

tromey:
For anyone planning to resurrect this patch - I am not sure that changing the source map worker in this way is best.  This behavior is relied on by the toolbox to avoid reporting the same source map error multiple times.

Instead, if `fetchSourceMap` fails, then `getOriginalURLs` will report an error.  This can be caught and used.  *However*, the toolbox installs a proxy that reports these errors and returns a non-rejected promise; so an M-C change would be needed as well.

bomsy:
@tromey really good to know, thanks! Will love to discuss more, to get the best possible solution.

PR #1943, opened 2017-02-06
Further Editor refactors

wldcordeiro:
Associated Issue: N/A

### Summary of Changes

* refactor the editor and utilities further to ease cognitive load and addition of tests.

### DIFF
--- a/src/utils/editor/index.js
+++ b/src/utils/editor/index.js
@@ -1,6 +1,6 @@
 const { isPretty, isJavaScript } = require("../source");
 const { isOriginalId } = require("../source-map");
-const { buildQuery } = require("./build-query");
+const buildQuery = require("./build-query");
 const {
   getDocument,
   setDocument,
--- a/src/utils/editor/index.js
+++ b/src/utils/editor/index.js
@@ -74,6 +74,55 @@ function clearLineClass(codeMirror, className) {
   });
 }
 
+function isTextForSource(sourceText) {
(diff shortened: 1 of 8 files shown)


src/utils/editor/index.js
```diff
@@ -74,6 +74,55 @@ function clearLineClass(codeMirror, className) {
...
+}
+
+function onMouseUp(ctx, modifiers) {
+  const query = ctx.cm.getSelection();
+  if (ctx.cm.somethingSelected()) {
+    find(ctx, query, true, modifiers);
+  } else {
+    removeOverlay(ctx);
+  }
+}
```
jasonLaster: hmm, i think this could be moved into the Editor component as we might want to change the Editor mouse up behavior

src/utils/editor/index.js
```diff
@@ -89,5 +138,12 @@ module.exports = {
   findNext,
   findPrev,
   removeOverlay,
-  SourceEditor
+  SourceEditor,
```
jasonLaster: hmm, why do we need this here?

wldcordeiro: I was hoping to make the index export all the associated utils and then only import from the dir.

src/utils/editor/index.js
jasonLaster: hmm, i guess we can evaluate that with other handlers like `onKeyDown`

PR #1953, opened 2017-02-07
Kill tcomb

jasonLaster:
One of the nice side-effects of working on upgrading to webpack 2 is that @wldcordeiro and I found that tcomb was still used for Frames.

This is a fairly big change to drop it, but something we should definitely do.


src/components/SecondaryPanes/Frames.js
```diff
@@ -5,11 +5,11 @@ import {
 } from "react";
 import { bindActionCreators } from "redux";
 import { connect } from "react-redux";
-import ImPropTypes from "react-immutable-proptypes";
 import actions from "../../actions";
 import { endTruncateStr } from "../../utils/utils";
 import { getFilename } from "../../utils/source";
-import { getFrames, getSelectedFrame, getSource } from "../../selectors";
+const { getFrames, getSelectedFrame, getSource } = require("../../selectors");
```
gregtatum: Why'd this turn into a require?

src/components/shared/Rep.js
```diff
@@ -1,5 +1,6 @@
 const React = require("react");
-let { Rep, Grip } = require("devtools-reps");
+
+let { REPS: { Rep, Grip }} = require("devtools-reps");
```
gregtatum: Why is this changing, it's not obvious to me?

jasonLaster: We switched to reps 6 recently and this changed

PR #1954, opened 2017-02-07
Fix #1948 - Use orangish color to note paused breakpoint

### LINKED ISSUE #1948: [Editor] Pause Line color
We currently show pause lines as green, which is slightly un-intuitive as green is associated with going.

It might me nice to try yellow or blue, which are also informational colors

darkwing:
This is based on `--theme-graphs-yellow`, lightened to match the debugger's color palette.  I also wanted a color that would look decent in both themes.

lighter

I'm happy to make any updates, let me know your thoughts.

### DIFF
--- a/src/components/Editor/codemirror-mozilla.css
+++ b/src/components/Editor/codemirror-mozilla.css
@@ -5,16 +5,14 @@
 :root {
   /* --breakpoint-background: url("chrome://devtools/skin/images/breakpoint.svg#light"); */
   /* --breakpoint-hover-background: url("chrome://devtools/skin/images/breakpoint.svg#light-hover"); */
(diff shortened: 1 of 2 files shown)


clarkbw:
> I'm happy to make any updates, let me know your thoughts.

We could use some designer help here.

My first thought is that we want to, holistically, tell a story with our colors; and that's where we probably need a designer.  Also there's a competing interest in keeping colors similar to our competition unless we believe there are significant gains to be made by going rogue. (improvements for color blind people or better contrast would be valid reasons here)  This one might be hard to do.

jasonLaster:
Hmm, I think we should close this PR and have more of a discussion in the issue then. 

I can draw up some options in sketch and we can discuss the goals.

Issue #1924, opened 2017-02-03
remove reference docs and create local API docs


clarkbw:
In https://github.com/gregtatum/quads/pull/3 I tried a new system of documentation that we might use for [perf.html](https://github.com/devtools-html/perf.html).

Instead of having our [reference docs](https://devtools-html.github.io/debugger.html/docs/reference/) in a separate location we integrate the docs into local README files. This keeps the docs relevant to the directory which should make them more relevant.  Having them be local should also increase the chances the docs will be used and updated.

You can see how the generated API section works here: https://github.com/clarkbw/quads/tree/documentationjs#api the documentation command rewrites the API section of the README doc.  

For the debugger I think this could mean that the `Editor` directory could have a README file which does some light explanations and then an API section below that which is generated by the source code.

Here's what we need to do to get started with this:

- [x] Remove the old reference docs
- [ ] Create a new system for handling the various README files
  - I think this could just be a number of scripts in the package.json or a new node script

jasonLaster:
Closing as we removed the reference docs and we don't need to track creating docs in an open issue. Hopefully we can do something small soon with the integration test helpers.

PR #1971, opened 2017-02-09
remove the docs/reference directory

clarkbw:
Associated Issue: #1924 

### Summary of Changes

* removed the `docs/reference` directory as it was outdated and I believe we are arriving at a better solution in #1924


jasonLaster:
like it.

Issue #1970, opened 2017-02-09
watch expressions variables don't expand


clarkbw:
When trying to expand a variable in the watch expressions list it seems to expand inline and then from there I'm not sure.  I was expecting it to work exactly like the scopes list.

http://recordit.co/EWErz5MCQF

![screen shot 2017-02-08 at 4 05 48 pm](https://cloud.githubusercontent.com/assets/2134/22763253/7dacd6d8-ee18-11e6-96e8-00f46ad31d0f.png)

arthur801031:
I'll take a stab at this :)

PR #1973, opened 2017-02-09
Fix watch expression tree not expanding

wldcordeiro:
I hadn't seen that @arthur801031 had claimed the issue but talked to him on Slack about opening the PR.

Associated Issue: #1970

### Summary of Changes

* remove flex box and change width to account for close button

### Test Plan

- [x] Add a watch expression and see that the tree actually expands vertically

#### Before

screenshot_20170208_225031

#### After

screenshot_20170208_225014

### DIFF
--- a/src/components/SecondaryPanes/Expressions.css
+++ b/src/components/SecondaryPanes/Expressions.css
@@ -43,8 +43,9 @@
 }
 
 .expression-container > .tree {
-  overflow: hidden;
+  overflow: auto;
   display: flex;
+  flex-direction: column;
 }
 
 :root.theme-light .expression-container:hover {
--- a/src/components/SecondaryPanes/Expressions.css
+++ b/src/components/SecondaryPanes/Expressions.css
@@ -43,7 +43,8 @@
 }
 
 .expression-container > .tree {
-  width: calc(100% - 20px);
+  width: 100%;
+  overflow: hidden;
 }
 
 :root.theme-light .expression-container:hover {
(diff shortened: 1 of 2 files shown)


src/components/SecondaryPanes/Expressions.css
```diff
@@ -43,8 +43,9 @@
 }
 
 .expression-container > .tree {
-  overflow: hidden;
+  overflow: auto;
```
arthur801031: Hi @wldcordeiro 
`overflow: auto` seems to cause the scroll bar to appear when you hove over an element. Reverting overflow to `hidden` would fix this issue.

screen shot 2017-02-09 at 1 54 14 pm

jasonLaster: Thanks @arthur801031 I just tried replacing `overflow: auto` with `hidden` and it worked well.

wldcordeiro:
@arthur801031 From my investigation this comes from `ObjectInspector` and could be fixed separately by having `text-overflow` set to ellipsis for element names?

jasonLaster:
@wldcordeiro ObjectInspector seems like a good idea. Feel free to go ahead and do it!

wldcordeiro:
I've been researching fixing the hover issue @arthur801031 pointed out, this is because of the `CloseButton` that is added on hover. This margin/padding 0 [here](https://github.com/devtools-html/devtools-core/blob/master/packages/devtools-sham-modules/client/shared/components/tree.js#L306) causes issues with the easiest solution as it overrides any margin we add.

jasonLaster:
I pushed a quick fix for the separate scroll bars

screen shot 2017-02-16 at 10 55 04 pm

Issue #1808, opened 2017-01-24
[Tabs] Vertically align tabs and new tab button


jasonLaster:
@irfanhudda made some really nice improvements in https://github.com/devtools-html/debugger.html/pull/1745 for the toggle button and new tab button.

We should see if we can get the tabs to vertically align with the new tab button now as well. I think this means that we should move the tab text up a bit...

Also, it might be a good opportunity to use flexbox with the tabs a bit more.

<img width="865" alt="screen shot 2017-01-23 at 7 02 30 pm" src="https://cloud.githubusercontent.com/assets/254562/22228350/8e136ef0-e19e-11e6-9b6d-29baa4002e42.png">

soheilhm:
I am working on this

princiya:
I am working on this issue.

princiya:
Not able to figure out, where the highlighted css (bottom right) is generated from. Any inputs please?
![css](https://cloud.githubusercontent.com/assets/8022693/22617910/1ce70aee-ead0-11e6-95a4-97740f929f88.png)
@irfanhudda @jasonLaster @clarkbw @wldcordeiro

irfanhudda:
@princiya That CSS is probably generated from [https://github.com/devtools-html/debugger.html/blob/master/src/components/Editor/Tabs.css#L50](https://github.com/devtools-html/debugger.html/blob/master/src/components/Editor/Tabs.css#L50)

Also take a look at [https://devtools-html.github.io/debugger.html/docs/local-development.html#rtl](https://devtools-html.github.io/debugger.html/docs/local-development.html#rtl) to better understand why we have things like `padding-inline-start`, `padding-inline-end`.

PR #1976, opened 2017-02-09
Vertically align editor's header panel

princiya:
Associated Issue: #1808 

### Summary of Changes

* Flexbox layout for the header panel --> tabs + buttons (The buttons were absolutely positioned earlier.)

### Before

dropdown-before

### After

dropdown-after


assets/images/fast-forward.svg
```diff
@@ -0,0 +1,7 @@
+<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" viewBox="0 0 192.689 192.689">
```
princiya: I got this svg code from here: http://www.flaticon.com/free-icon/fast-forward_130903

nchevobbe:
Hey @princiya thanks for the patch :+1:  !
So I pulled your patch and give it a shot. Here's the before/after screenshots : 

Before:
before

After:
after

As you can see, we lost the ">>" and the "+" buttons, it may be due to some changes on the sizing of the tab bar. I am gonna review your patch and see what can cause this :)

assets/images/Svg.js
```diff
@@ -29,7 +29,8 @@ const svg = {
   "whole-word-match": require("./whole-word-match.svg"),
   "worker": require("./worker.svg"),
   "sad-face": require("./sad-face.svg"),
-  "refresh": require("./refresh.svg")
+  "refresh": require("./refresh.svg"),
+  "fastForward": require("./fast-forward.svg")
```
princiya: Not sure about the naming convention for this newly added svg file. It is used as a dropdown for hidden tabs, but the symbol represents fast-forward.

Garbee:
>  (I am not entirely sure if this whole PR is bringing any value by replacing absolute position with flexbox, let us discuss :))

Just on this point real quick, yes it provides benefit. We get the same layout but don't need to deal with the positioning problems that happen when the editor has low-space and things start stacking. As the buttons are taken into account with the flex container. This was needed to solve another bug I opened up that's laying around somewhere.

src/components/App.css
```diff
@@ -16,6 +16,7 @@
 .editor-container {
   display: flex;
   flex: 1;
+  width: 100%;
```
nchevobbe: Can you explain me why we do this ?

princiya:
@nchevobbe The sreenshots are from Firefox? I have added a to-do (2nd point), and I am trying to find the wrong calculations.

src/components/Editor/Tabs.css
```diff
@@ -1,58 +1,53 @@
...
   transition: 0.1s ease;
 }
 
+.source-header .new-tab-btn svg {
+  width: 12px;
+}
+
+.source-tabs {
+  max-width: calc(100% - 80px);
+  align-self: flex-end;
```
nchevobbe: I think we shouldn't have this here. In fact, removing it fixes the issue with the "+" and ">>" buttons disapearing

src/components/Editor/Tabs.css
```diff
@@ -1,58 +1,53 @@
...
-  line-height: 20px;
-  display: inline-block;
   position: relative;
   transition: all 0.25s ease;
-  min-width: 40px;
-  overflow: hidden;
+  min-width: 80px;
 }
 
 .source-tab {
```
nchevobbe: could we merge this rule, with the previous one (https://github.com/devtools-html/debugger.html/pull/1976/files#diff-d9abd0398fe2e0fa5793ba7d100231f5R35 ) ?

src/components/Editor/Tabs.css
```diff
@@ -108,13 +105,9 @@
 }
 
 .source-tab .close-btn {
```
nchevobbe: same here, could we merge this with the rule in https://github.com/devtools-html/debugger.html/pull/1976/files#diff-d9abd0398fe2e0fa5793ba7d100231f5R91 ? If I'm correct they are the same

src/components/Editor/Tabs.js
```diff
@@ -21,24 +21,21 @@ const { formatKeyShortcut } = require("../../utils/text");
...
  *
  * @param sourceTabs Immutable.list
  * @param sourceTabEls HTMLCollection
  *
  * @returns Immutable.list
  */
-function getHiddenTabs(sourceTabs, sourceTabEls) {
+function getHiddenTabs(sourceTabsParent, sourceTabs, sourceTabEls) {
+  const parentRightOffset = sourceTabsParent.getBoundingClientRect().right;
+  const sourceTabMinWidth = 80; // this value is set in tabs.css
```
nchevobbe: I think we could have an issue here. With that, we must maintain an implicit sync between this value and the one in the css file.
What would be really nice would be to declare the tab width as a custom property in the CSS file, and then read the value of this custom property here, in JS (see https://www.broken-links.com/2014/08/28/css-variables-updating-custom-properties-javascript/ ).

src/components/shared/Button/PaneToggle.css
```diff
@@ -1,18 +1,21 @@
 .toggle-button-start,
 .toggle-button-end {
-  position: absolute;
-  width: 16px;
-  height: 16px;
   transition: transform 0.25s ease-in-out;
-  margin: 0 4px;
   cursor: pointer;
+  align-self: flex-end;
```
nchevobbe: I don't know if we want to have this on the `toggle-button-start` too, what do you think ?

src/components/shared/Button/PaneToggle.css
```diff
@@ -1,18 +1,21 @@
...
 }
 
 .toggle-button-start svg,
 .toggle-button-end svg {
+  width: 14px;
   fill: var(--theme-comment);
 }
 
+.toggle-button-end {
+  margin-left: auto;
```
nchevobbe: I think this might be redundant with the `align-self` property from https://github.com/devtools-html/debugger.html/pull/1976/files#diff-65320c92508ade610814346c22bc028cR5

src/components/shared/Dropdown.css
```diff
@@ -10,17 +10,19 @@
   z-index: 1000;
 }
 
+.dropdown-block {
+  align-self: flex-end;
```
nchevobbe: is this element in a flex container ? If so, I think we could remove the `position: relative`

src/components/shared/Dropdown.js
```diff
@@ -33,12 +34,12 @@ const Dropdown = React.createClass({
   },
 
   renderButton() {
-    return dom.button(
+    return dom.div(
```
nchevobbe: is there a reason to change this to a div ? From an accessibility point of view, it would be really nice to have this still as a button (because it implies the user can act on it, whereas a div is just a container).
It's not the scope of your PR, but it would be even nicer if we could have an `aria-haspopup` attribute (https://msdn.microsoft.com/fr-fr/library/windows/apps/hh465704.aspx) on this :)

nchevobbe:
@princiya yes, it is from Firefox. I made a small gif so you can see it in action : 
peek 09-02-2017 16-29

princiya:
@nchevobbe I will reply to all your comments soon :)

Garbee:
So the fix on the overflow is removing `display: flex; flex: 1;` from `.editor-container`. For some reason the container being flex is screwing with Firefox's layout calculations. Not sure exactly why....

src/components/App.css
princiya: @nchevobbe  @Garbee https://github.com/devtools-html/debugger.html/pull/1976#issuecomment-278764001 In continuation to this comment, I was not sure if to remove the flex layout, so made it `width 100%` here instead.
Without `width 100%`, and the editor-container still being flex, the source-header which is flex, didn't take 100% width. Any inputs?

src/components/shared/Button/PaneToggle.css
princiya: This is a hack to right align the `toggle-button-end`. @Garbee any other way to right align only one flex child?

src/components/shared/Dropdown.css
princiya: `position: relative` is needed so that the absolutely positioned dropdown list renders fine

src/components/shared/Dropdown.js
princiya: will look into this :)

src/components/App.css
Garbee: `flex` containers are like `block` they instantly are just 100% of the parent container in width. This rule is redundant. However, the deeper problem with the off rendering is the fact that this editor container is a flex display. Undoing `display: flex; and flex: 1;` both solve the width problem observed in Firefox.

This looks like one of the quirky cross-browser issues when it comes to flexbox. I'll try to make a reproducible sample from this to reference against the spec and file bugs where appropriate with.

princiya:
@Garbee @jasonLaster @nchevobbe @wldcordeiro Can somebody please help me understand when the right pane toggle button in the below gif moves downward. Thanks in advance :)

pane-toggle-error

Garbee:
When the viewport width is equal to or less than 700px wide is when it goes into vertical mode.

src/components/Editor/Tabs.css
princiya: Removing `display: flex` from here was the fix. https://github.com/devtools-html/debugger.html/pull/1976/files#diff-bc1880b2b43f6a2635e5a0a1e3bb09e2L16

src/components/Editor/Tabs.css
```diff
@@ -1,62 +1,56 @@
...
-  height: 24px;
-  line-height: 20px;
-  display: inline-block;
   position: relative;
   transition: all 0.25s ease;
-  min-width: 40px;
-  overflow: hidden;
-}
-
-.source-tab {
```
princiya: These rules are merged. But looks like a bug from `postcss-bidirection`. If I explicitly don't specify these rules https://github.com/devtools-html/debugger.html/pull/1976/files#diff-d9abd0398fe2e0fa5793ba7d100231f5R52, the `hover` is never applied, because the rules are suppressed by the postcss processing. Maybe that's why, these CSS rules weren't merged and written out explicitly.

src/components/Editor/Tabs.css
princiya: rules are now merged, but had to explicitly write the rule `html[dir="ltr"] .source-tab:hover .close-btn`

jasonLaster:
@princiya just pushed a rebase and a fix for showing the toggle button in the footer for vertical mode

src/components/Editor/Tabs.js
```diff
@@ -21,24 +21,36 @@ const { formatKeyShortcut } = require("../../utils/text");
...
-    const topOffsets = sourceTabEls.map(t => t.getBoundingClientRect().top);
-    return Math.min(...topOffsets);
+  
+  const sourceTabMinWidth = sourceTabEls.length ? 
+    window.getComputedStyle(sourceTabEls[0]).getPropertyValue("min-width") :
+    0;
+  const sourceTabDirection = sourceTabEls.length ? 
+    window.getComputedStyle(sourceTabEls[0]).getPropertyValue("direction") :
+    "ltr";
+  const parentOffset = sourceTabDirection === "ltr" ?
```
jasonLaster: nit, but we could simplify this to:

```js
const parentOffsetDir = sourceTabDirection === "ltr" ? "right" : "left"
const parentOffset = sourceTabsParent.getBoundingClientRect()[parentOffsetDir]
```

src/components/Editor/Tabs.js
```diff
@@ -21,24 +21,36 @@ const { formatKeyShortcut } = require("../../utils/text");
...
+    sourceTabsParent.getBoundingClientRect().right :
+    sourceTabsParent.getBoundingClientRect().left;
+
+  if (sourceTabDirection === "ltr") {
```
jasonLaster: this probably would benefit from the same cleanup:

```js
const tabDir = sourceTabDirection === "ltr" ? "left" : "right"
     return (sourceTabEls[index].getBoundingClientRect()[tabDir]
     parseInt(sourceTabMinWidth)) > parentOffset;
});  
```

jasonLaster: also maybe add a `getTabOffset` helper

```js
function getTabOffset(tab, dir) {
  return tab.getBoundingClientRect()[dir]
}
```

this will abstract out the logic a bit

src/components/Editor/Tabs.js
```diff
@@ -21,24 +21,36 @@ const { formatKeyShortcut } = require("../../utils/text");
...
-function getHiddenTabs(sourceTabs, sourceTabEls) {
+function getHiddenTabs(sourceTabsParent, sourceTabs, sourceTabEls) {
   sourceTabEls = [].slice.call(sourceTabEls);
-  function getTopOffset() {
-    const topOffsets = sourceTabEls.map(t => t.getBoundingClientRect().top);
-    return Math.min(...topOffsets);
+  
+  const sourceTabMinWidth = sourceTabEls.length ? 
+    window.getComputedStyle(sourceTabEls[0]).getPropertyValue("min-width") :
+    0;
```
jasonLaster: 👍 

nit: we should probably have a helper for getting these tab properties. That way the logic for checking for tabs and then getting the property is hidden away :)

Issue #1842, opened 2017-01-25
[EditorHover] Add a simple getExpressionsInScope function


jasonLaster:
We recently landed the babel parser and it's now time to do a simple MVP function to put it to use.

What I imagine is we write a simple traversal function that finds some valid expressions that are "in scope"

We can do this by:
a) finding the node that corresponds with a scope
b) scanning it's child nodes for the simplest expression, an identifier. 

For example, if we pause in `sum`, we should be able to find the two identifiers `a`, `b` and tell the editor that if the user hovers over one of them we can do an evaluation.

```js
function sum(a+b) {
  debugger;
  return a + b;
}
```

Here are two tools for doing this [ast explorer][explorer] [babel handbook][handbook]

[explorer]: https://astexplorer.net/
[handbook]: https://github.com/thejameskyle/babel-handbook/blob/master/translations/en/plugin-handbook.md#toc-get-the-path-of-a-sub-node

mattzeunert:
I'll try and implement this.

jasonLaster:
Just landed #1988, which gives a node path, which references a scope. We can use this to get all of the binding variables in the scope and their reference paths:

I'm actually adding a babel path that [demonstrates](https://github.com/babel/babel/pull/5296/files#diff-56f320a6a89e1ca4a32a0cbb2da293a1R79) this behavior.

PR #1988, opened 2017-02-10
Add function to find the Babel path at a specific line/column

mattzeunert:
Associated Issue: #1842

As the first step to find the variables a user can hover over we need to find the location in the AST where the debugger has paused.

I'm returning the full Babel path for now, but depending on how we'll end up implementing hovering over variables just the node or scope might be sufficient.


jasonLaster:
Thanks @mattzeunert. I pushed some tweaks to your branch.

I think the second part, finding the variables, might be much easier than expected...

For instance, the node's we're finding have a context and associated scope. Scopes have variable [bindings], which could be really useful. In our test case, `square` has a reference `n` with two reference paths. Each path provides a location where it's used.

Our MVP could probably simply return the reference path locations...


[bindings]: https://github.com/thejameskyle/babel-handbook/blob/master/translations/en/plugin-handbook.md#bindings

jasonLaster:
@mattzeunert I hope you don't mind, but i tacked on the function definition querying so that we can establish some more patterns with the babel-types and traversing.

jasonLaster:
Landing this so that we can iterate on top of it. It looks like there will be several follow up issues from here.

PR #1990, opened 2017-02-10
Fix arraylike serialization

jasonLaster:
Associated Issue: #1375
### Summary of Changes

* updates fromJS hydrations.

### Test Plan

* added unit tests
* tested the `array` example

### DIFF
--- a/src/utils/fromJS.js
+++ b/src/utils/fromJS.js
@@ -8,6 +8,31 @@
 
 const Immutable = require("immutable");
 
+/*
+  creates an immutable map, where each of the value's
+  items are transformed into their own map.
+
+  NOTE: we guard agains `length` being a property because
+  length confuses Immutable's internal algorithm.
+*/
+function createMap(value) {
+  if (value.hasOwnProperty("length")) {
+    value.length = `${ value.length}`;
+  }
+
+  let map = Immutable.Seq(value).map(fromJS).toMap();
+
+  if (map.has("length")) {
+    map = map.set("length", parseInt(map.get("length"), 10));
+  }
+
+  return map;
+}
+
+function createList(value) {
+  return Immutable.Seq(value).map(fromJS).toList();
+}
+
 /**
  * When our app state is fully typed, we should be able to get rid of
  * this function. This is only temporarily necessary to support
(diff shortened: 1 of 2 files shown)


src/utils/fromJS.js
```diff
@@ -8,6 +8,31 @@
...
+/*
+  creates an immutable map, where each of the value's
+  items are transformed into their own map.
+
+  NOTE: we guard agains `length` being a property because
+  length confuses Immutable's internal algorithm.
+*/
+function createMap(value) {
+  if (value.hasOwnProperty("length")) {
+    value.length = `${ value.length}`;
```
juliandescottes: I don't think createMap should have a side-effect on the value argument passed to it.

We should either restore `value.length = parseInt(value.length, 10);`, 
or work on a copy of value `copy = Object.assign({}, value, { length: undefined });`

src/utils/fromJS.js
```diff
@@ -8,6 +8,31 @@
 
 const Immutable = require("immutable");
 
+/*
+  creates an immutable map, where each of the value's
+  items are transformed into their own map.
+
+  NOTE: we guard agains `length` being a property because
```
juliandescottes: nit: against

src/utils/fromJS.js
jasonLaster: @juliandescottes I believe that's what we do below :)

juliandescottes: No, you reset it on the new map. But you still modified the argument. 

If you do : `let a = { length : 10 }; fromJS(a);` -> a.length is now "10" (string) instead of 10 (number)

jasonLaster: ahh - i see :) 

yeah i can fix that

src/utils/fromJS.js
```diff
@@ -8,6 +8,35 @@
...
+
+  NOTE: we guard against `length` being a property because
+  length confuses Immutable's internal algorithm.
+*/
+function createMap(value) {
+  const hasLength = value.hasOwnProperty("length");
+  const length = value.length;
+
+  if (hasLength) {
+    value.length = `${ value.length}`;
```
bjorn3: nit: space between `${` and `value.length}` and not between `value.length` and `}`

Issue #1930, opened 2017-02-04
[EditorSearch] Polish Search


wldcordeiro:
Checklist:

- [x] Clearing the query to an empty string causes a search of empty text.
- [x] the search index starts at 1 and always increments by 1 #1967
- [x] deleting the query does not clear the last match
- [x] white space queries breaks the editor
- [x] queries increment the searches #2046
- [x] can't delete the last letter of the query
- [x] clicking in the document does not update the search index #2065
- [x] Editor Search breaks with "whole word" modifier and query's of 3 or less characters #2105 
- [ ] escape should always close the search bar #2104 @nchevobbe 
- [x] cmd+f without a selection should refocus the search bar #2103 
- [x] search jumps #2081
- [ ] Add integration tests #1928
- [x] move search state to UI reducer #2126 
- [x] Add search bottom bar #2078 
- [x] Refactor search components to better support function search in editor. #2151 
- [x] doSearch slowness #2154 
- [x] Only do one search type #2171 
- [x] polish overlays #2228

arthur801031:
I'll try to take a stab at "Clearing the query to an empty string causes a search of empty text.".

arthur801031:
Hi @wldcordeiro!

I can't reproduce this bug: "Clearing the query to an empty string causes a search of empty text". Please let me know how to reproduce it. Thank you! 😀
![2c31hesfrz](https://cloud.githubusercontent.com/assets/4562118/22810033/294e56de-ef71-11e6-8d97-543dd306c74a.gif)

jasonLaster:
Thanks @arthur801031. I couldn't find that either. But it did help me find two others :)

- [ ] deleting the query does not clear the last match
- [ ] white space queries breaks the editor

![](http://g.recordit.co/TNaR3tONHH.gif)

arthur801031:
Hi @jasonLaster 

For "deleting the query does not clear the last match" bug, the Chrome debugger also has this behavior, but it changes the highlight color for the last character that was cleared. Maybe we should change the yellow to something else?

![zdmwbp0w6h](https://cloud.githubusercontent.com/assets/4562118/22966296/84bcd0a2-f39c-11e6-93fb-630f7217e5e0.gif)

jasonLaster:
closing this as we've done all of the big work here

PR #1994, opened 2017-02-11
Fix white space queries breaking the editor

arthur801031:
Associated Issue: #1930

### Summary of Changes
* Add a regex to check if the query only contains whitespace. If this is true, we do not search this query. Otherwise, we perform a search.

### DIFF
--- a/src/components/Editor/SearchBar.js
+++ b/src/components/Editor/SearchBar.js
@@ -152,7 +152,18 @@ const SearchBar = React.createClass({
 
     // eslint-disable-next-line react/no-did-update-set-state
     this.setState({ count, index: 0 });
-    this.search(query.trim());
+    if (!this.leadingWhiteSpaceOnly(query)) {
+      this.search(query);
+    }
+  },
+
+  leadingWhiteSpaceOnly(query: string) {
+    for (let i = 0; i < query.length; i++) {
+      if (query[i] != " ") {
+        return false;
+      }
+    }
+    return true;
   },
 
   onChange(e: any) {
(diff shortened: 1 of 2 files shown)


arthur801031:
Hi @jasonLaster 

I have a hack that fixes this issue. The editor would break when the query only contains whitespace(s), so I wrote a function to check if the query only contains whitespace(s). If this is true, we do not perform a search; otherwise, a search is performed.

In this example, " DOM" and "DOM ", query that contains leading whitespace or trailing whitespace and other characters still works.
3dlxtti1x6

src/components/Editor/SearchBar.js
```diff
@@ -152,7 +152,18 @@ const SearchBar = React.createClass({
...
     // eslint-disable-next-line react/no-did-update-set-state
     this.setState({ count, index: 0 });
-    this.search(query);
+    if (!this.leadingWhiteSpaceOnly(query)) {
+      this.search(query);
+    }
+  },
+
+  leadingWhiteSpaceOnly(query: string) {
+    for (let i = 0; i < query.length; i++) {
```
jasonLaster: Hmm. Maybe use a regex?

arthur801031: Sure!

jasonLaster:
@arthur801031 i moved the check to `source-search` so that the logic is more centralized.

@wldcordeiro @bomsy what do you think?

bomsy:
Looks Good!

wldcordeiro:
Maybe adding a test as well for this case would help in a follow-up PR?

jasonLaster:
Yea. Integration tests coming

Issue #1846, opened 2017-01-25
[EditorSearch] add `findFunctionsInSource` to parser worker


jasonLaster:
This issue is blocked by #1841

Issue #1845, opened 2017-01-25
[EditorSearch] Add function search UI


jasonLaster:
Function Search UI:

- [x] polish autocomplete summary
- [ ] add compact result list #2047
- [ ] add function search bottom bar button

----
We're working on a function search POC in #1995. When that lands, we'd like to add a proper UI for the feature.

I'd like to get some feedback on this UI I sketched. The main pieces are:

1. function search button is in the bottom left and is a label
2. search results are a compact list and follow the same hover/select/focus rules as the source tree.

![function-search](https://cloud.githubusercontent.com/assets/254562/22855693/3bc6fe24-f052-11e6-862d-922fa8183bb4.jpg)

jasonLaster:
Including Helen's mock which has a label in the input, which could be helpful


![](https://cloud.githubusercontent.com/assets/254562/21696191/e4a17e30-d35a-11e6-8675-4bb7ff191ca7.png)

wldcordeiro:
I like Helen's mock, it makes the function search stuff's design match the search modifiers in the editor searchbar more closely.

jasonLaster:
Yea. I like that the magnifying glass titles the secondary bar.

clarkbw:
Here's a list of things left to fix here:

- [ ] make the autocomplete CSS more generic instead of file search specific with function search CSS trying to turn off borders it doesn't want.
- [ ] fix the bug where you can arrow into an item that doesn't exist, this seems to be an issue with the `selectedIndex` though I haven't investigated yet
- [ ] somehow handle when the results are larger than the 300px we make available.  Either scroll the results or limit them and require more typing?

clarkbw:
> fix the bug where you can arrow into an item that doesn't exist, this seems to be an issue with the selectedIndex though I haven't investigated yet

I think I just saw this in action and believe it has to do with multiple definitions.  If you have the multiple Backbone models in a file you could have several `initialize` functions.  It seems we aren't showing the different versions but still allow you to arrow through them and navigate to them.

jasonLaster:
Oh nice. We should make the key name + line number

PR #1995, opened 2017-02-11
(WIP) Function search

jasonLaster:
Associated Issue: #1845 #1846

### Summary of Changes

* takes advantage of the parser function for finding sources added in the previous PR
* adds a simple function search UI that is prompted by `<cmd>+r` which is probably the wrong handle
* we should think through the right UX and UI. Currently the UI has a bad default label "search sources", takes up too much space, and doesn't factor in variable search, and other search modifiers that will be available in a bottom search bar...
* I'll add a feature flag once the parser functions land so that we can work on it darkly

### DIFF
--- a/src/components/Editor/index.js
+++ b/src/components/Editor/index.js
@@ -10,7 +10,7 @@ const { isEnabled } = require("devtools-config");
 
 const { getMode } = require("../../utils/source");
 const {getFunctions} = require("../../utils/parser");
-const Autocomplete = createFactory(require("../shared/autocomplete"));
+const Autocomplete = createFactory(require("../shared/Autocomplete"));
 
 const Footer = createFactory(require("./Footer"));
 const SearchBar = createFactory(require("./SearchBar"));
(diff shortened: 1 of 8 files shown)


src/components/Editor/index.js
```diff
@@ -73,7 +77,9 @@ const Editor = React.createClass({
         caseSensitive: true,
         wholeWord: false,
         regexMatch: false
-      }
+      },
+      functionSearchEnabled: false,
+      functionDeclarations: null
```
jasonLaster: this should be a different name. Also we keep it here for memoization, but it should probably be kept in the util or redux store so that it is available if you change documents.

bomsy:
I seem to get this when testing ...
All the source disappears when opened

search-functions

jasonLaster:
yep - @bomsy that's the wrong key shortcut :) you need to do `<cmd>+shift+o` in the editor

bomsy:
oh! my bad should `<cmd>+o` even work?

jasonLaster:
@bomsy no problem, cmd+o is a thing in chrome, but not sure if we want it

bomsy:
general note: I think we should change the placeholder text in the search bar based on the type of search we are doing.
Rather than showing `Search Sources...` we could show `Search Sources for Functions...` or something

PR #2002, opened 2017-02-13
Add flow to source-documents.js and source-editor.js

arthur801031:
Associated Issue: #1785

### Summary of Changes

* Add flow to source-documents.js and source-editor.js

### Test Plan
- Passed 
`yarn run flow`
`npm run lint-css -s; npm run lint-js -s`
`node src/test/node-unit-tests.js --dots`

### DIFF
--- a/src/utils/editor/source-editor.js
+++ b/src/utils/editor/source-editor.js
@@ -16,6 +16,8 @@ require("codemirror/mode/elm/elm");
 require("../../components/Editor/codemirror-mozilla.css");
 require("codemirror/addon/search/searchcursor");
 
+import type { Mode, AlignOpts } from "../../types";
+
 // Maximum allowed margin (in number of lines) from top or bottom of the editor
 // while shifting to a line which was initially out of view.
 const MAX_VERTICAL_OFFSET = 3;
@@ -62,7 +64,7 @@ class SourceEditor {
     return this.editor.getValue();
   }
 
-  setMode(value: Object) {
+  setMode(value: Mode) {
     this.editor.setOption("mode", value);
   }
 
@@ -89,7 +91,7 @@ class SourceEditor {
    * bottom.
    * @memberof utils/source-editor
    */
-  alignLine(line: number, align: string = "top") {
+  alignLine(line: number, align: AlignOpts = "top") {
     let cm = this.editor;
     let from = cm.lineAtHeight(0, "page");
     let to = cm.lineAtHeight(cm.getWrapperElement().clientHeight, "page");
--- a/src/utils/editor/source-editor.js
+++ b/src/utils/editor/source-editor.js
@@ -1,3 +1,5 @@
+// @flow
+
 /**
  * CodeMirror source editor utils
  * @module utils/source-editor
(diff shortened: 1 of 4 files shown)


src/utils/editor/source-editor.js
```diff
@@ -18,40 +20,57 @@ require("codemirror/addon/search/searchcursor");
...
+  appendToLocalElement(node: any) {
     this.editor = CodeMirror(node, this.opts);
   }
 
   destroy() {
     // No need to do anything.
   }
 
-  get codeMirror() {
+  get codeMirror(): any {
```
arthur801031: Without `unsafe.enable_getters_and_setters=true` in `.flowconfig`, we would get this error:
screen shot 2017-02-13 at 9 50 28 am

arthur801031:
Hi @jasonLaster!

Anything that is created by `CodeMirror` and used in `source-documents.js` and `source-editor.js` is assigned with type `any`. I'm not sure if this is the correct type to use. Please let me know. Thank you! 😄

src/utils/editor/source-editor.js
jasonLaster: This looks fine given:
a) it doesn't look possible to disable one line of flow
b) we'll hopefully refactor this file next month when we bring the mc source editor to github.

src/utils/editor/source-editor.js
```diff
@@ -18,40 +20,57 @@ require("codemirror/addon/search/searchcursor");
 // while shifting to a line which was initially out of view.
 const MAX_VERTICAL_OFFSET = 3;
 
+type SourceEditorOpts = {
```
jasonLaster: 👍

src/utils/editor/source-editor.js
```diff
@@ -70,7 +89,7 @@ class SourceEditor {
    * bottom.
    * @memberof utils/source-editor
    */
-  alignLine(line, align = "top") {
+  alignLine(line: number, align: string = "top") {
```
clarkbw: Is an enumeration possible here for align?

src/utils/editor/source-editor.js
```diff
@@ -18,40 +20,57 @@ require("codemirror/addon/search/searchcursor");
...
+  setText(str: string) {
     this.editor.setValue(str);
   }
 
   getText() {
     return this.editor.getValue();
   }
 
-  setMode(value) {
+  setMode(value: Object) {
```
clarkbw: I think mode is a string in the opts type, right?

arthur801031: I think it's an object with a string `name`. Do we want to do `string` instead?

screen shot 2017-02-13 at 10 59 15 am

src/utils/editor/source-editor.js
arthur801031: I'm not familiar with `alignLine`.  It seems it can be called in [source-search.js](https://github.com/devtools-html/debugger.html/blob/master/src/utils/editor/source-search.js#L174) and [Editor/index.js](https://github.com/devtools-html/debugger.html/blob/master/src/components/Editor/index.js#L384). `align` seems to be type `string` in both cases. I could be totally wrong though...  Please let me know. Thank you!

src/utils/editor/source-editor.js
clarkbw: Oh strange, never mind then

src/utils/editor/source-editor.js
clarkbw: Just asking, I didn't take a deeper look. String is a good start.

src/utils/editor/source-editor.js
jasonLaster: Heh. I think sometimes it's a string. Let's hunt this down. I believe we keep a modded dict in utils source

src/utils/editor/source-editor.js
jasonLaster: this can be

```js
let type AlignOpts =  "top" | "center" | "bottom" ;
```

src/utils/editor/source-editor.js
jasonLaster: Good call @clarkbw 

looks like in some cases it's a string

https://github.com/jasonLaster/debugger.html/blob/sparse2/src/utils/source.js#L89-L101

perhaps we add a mode type in `src/types` and use it in both `utils/source` and here

```js
let type Mode = String | {
  name: String,
  typescript?: String
  base?: {
    name: String,
    typescript: String
  }
}
```

frankly the typescript mode seems weird here. We should look at it again :)

arthur801031: Hi @jasonLaster ,

Should `typescript` be `boolean`?  It looks like it's being assigned with a boolean value. 🤔
screen shot 2017-02-14 at 8 55 22 am

jasonLaster: yeah, i think so!

Issue #1967, opened 2017-02-08
[EditorSearch] Fix search results count


jasonLaster:
The search count always starts at 1 and increments from there. It should reflect where you are and add or subtract one as you iterate.

wldcordeiro:
I'll see about this one.

PR #2005, opened 2017-02-13
source-search count

rrandom:
Associated Issue: #1967 

### Summary of Changes

* `countMatches` find all matches and save the positions, pass `ctx` as a parameter.
* `doSearch` return the current index of the matches

### DIFF
--- a/src/components/Editor/SearchBar.js
+++ b/src/components/Editor/SearchBar.js
@@ -144,14 +144,24 @@ const SearchBar = React.createClass({
     }
 
     updateQuery(query);
+
+    const ed = this.props.editor;
+
+    if (!ed) {
+      return;
+    }
+
+    const ctx = { ed, cm: ed.codeMirror };
+
     const count = countMatches(
+      ctx,
       query,
       sourceText.get("text"),
       modifiers
     );
 
     // eslint-disable-next-line react/no-did-update-set-state
-    this.setState({ count, index: 0 });
+    this.setState({ count });
     this.search(query);
   },
 
@@ -170,7 +180,7 @@ const SearchBar = React.createClass({
     }
 
     const ctx = { ed, cm: ed.codeMirror };
-    const { index, count } = this.state;
+
     const { query, modifiers } = this.props;
 
     if (query === "") {
@@ -178,9 +188,8 @@ const SearchBar = React.createClass({
     }
 
     const findFnc = rev ? findPrev : findNext;
-    findFnc(ctx, query, true, modifiers);
 
-    const nextIndex = index == count - 1 ? 0 : index + 1;
+    const nextIndex = findFnc(ctx, query, true, modifiers);
     this.setState({ index: nextIndex });
   },
 
@@ -202,7 +211,8 @@ const SearchBar = React.createClass({
     const ed = this.props.editor;
     const ctx = { ed, cm: ed.codeMirror };
 
-    find(ctx, query, true, this.props.modifiers);
+    const index = find(ctx, query, true, this.props.modifiers);
+    this.setState({ index });
   }, 100),
 
   renderSummary() {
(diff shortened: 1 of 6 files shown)


wldcordeiro:
I'm going to work with you to get our tests passing, thanks for the great work @rrandom  😄

src/utils/editor/source-search.js
```diff
@@ -238,16 +257,23 @@ function findNext(
...
 }
 
 function countMatches(
-  query: string, text: string, modifiers: SearchModifiers): number {
-  const regexQuery = buildQuery(query, modifiers, {
-    isGlobal: true
-  });
-  const match = text.match(regexQuery);
-  return match ? match.length : 0;
+  ctx: any, query: string, text: string, modifiers: SearchModifiers): number {
```
rrandom: @wldcordeiro, Thanks! I think `countMatches` do not need `text` as a parameter, but I'm not very sure, so I have not remove it.

src/components/Editor/SearchBar.js
```diff
@@ -144,14 +144,24 @@ const SearchBar = React.createClass({
     }
 
     updateQuery(query);
+
```
rrandom: There are some duplicate code in this `doSearch` and `search` in line 204. I am thinking whether write their functionality in only one function or not

wldcordeiro: 👍 I'll refactor a bit here. I'm taking a look at our tests too. 😄

jasonLaster:
@wldcordeiro catching a plane. I'll try and test tomorrow

src/utils/editor/source-search.js
```diff
@@ -120,18 +129,31 @@ function updateCursor(cm, state, keepSelection) {
...
+
+    if (nextMatch) {
+      if (state.matchIndex === -1) {
+        state.matchIndex = findIndex(state.results, nextMatch);
+      } else {
+        state.matchIndex = rev ? state.matchIndex - 1 : state.matchIndex + 1;
+      }
+      matchIndex =
+        (state.matchIndex + state.results.length) % state.results.length;
+    }
```
jasonLaster: Hmm, looks like this is giving us a weird value... and it can likely be cleaned up further.

I'm going to try moving the modulo math to the increment / decrement line

screen shot 2017-02-17 at 3 27 31 pm

jasonLaster:
I extracted getMatchIndex and added some unit tests. still seeing some issues while i'm testing:

wrapping is weird - basically going from match 2, to 1, to n, n-1 and vice versa.

src/utils/editor/source-search.js
```diff
@@ -32,6 +32,8 @@ function SearchState() {
 function getSearchState(cm: any, query, modifiers) {
   let state = cm.state.search || (cm.state.search = new SearchState());
   let cursor = getSearchCursor(cm, query, null, modifiers);
+
+  state.results = [];
```
rrandom: Here, `state.results` get cleared every time even just `findNext` or `findPrev`.

src/utils/editor/source-search.js
```diff
@@ -119,6 +121,24 @@ function updateCursor(cm, state, keepSelection) {
...
+
+  if (state.matchIndex === -1) {
+    return findIndex(state.results, nextMatch);
+  }
+
+  matchIndex = rev ? state.matchIndex - 1 : state.matchIndex + 1;
+  matchIndex = matchIndex % count;
+
+  if (matchIndex == 0) {
+    matchIndex = rev ? count : 1;
```
rrandom: Seems `matchIndex` already plus 1 when `render` in `SearchBar.js`. so it should starts at 0 and ends at count-1...

wldcordeiro: Yeah have to agree here. In `SearchBar` we're displaying so the `+1` is necessary for UX but I don't think our index function should return out of range numbers. I'll look into it a bit further. 😄

jasonLaster:
Thanks @rrandom for starting this off. It's a really hard problem

PR #2026, opened 2017-02-15
Add buckets for objects and keys

jasonLaster:
Associated Issue: #2025

### Summary of Changes

This adds the logic for bucketing objects and arrays with lots of keys. It's still pretty rough, so it would be great to get some thoughts on how to clean it up.

### Test Plan

- [ ] add some unit tests
- [ ] add an integration test

object keys

### DIFF
--- a/docs/integration-tests.md
+++ b/docs/integration-tests.md
@@ -14,6 +14,8 @@ The integration tests are async functions that drive the debugger in two context
 
 ### Running Tests
 
+* Launch firefox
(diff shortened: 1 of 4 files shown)


wldcordeiro:
This is great if you ever deal with things like [`Uint8ClampedArray`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint8ClampedArray)s! Awesome @jasonLaster

PR #2028, opened 2017-02-15
Use the same react version as the launchpad.

Garbee:
### Summary of Changes

* Aligned with the react pieces in use from [the launchpad](https://github.com/devtools-html/devtools-core/commit/fc07d7fd37a1a786dc645da56c6f5bd9681eb348) to prevent errors on the UMD module not existing.

The reason for this is because the `14` line of React doesn't have the UMD Module. So, the alias setup in the launchpad webpack config fails to load properly since it is calling from the root packages `node_modules` and not that sub-packages module structure.

### Test Plan

Example test plan:

- Run current `master` and notice errors in the terminal and DevTools console on missing the `react` module
- Run with upgraded react, and see no ReactUMD related errrors.


juliandescottes:
Retriggered the CI job

PR #2034, opened 2017-02-16
Add flow to Breakpoints.js, ChromeScopes.js, CommandBar.js,...

arthur801031:
Associated Issue: #1894

### Summary of Changes

* Add flow to Breakpoints.js, ChromeScopes.js, CommandBar.js, EventListeners.js, Expressions.js

### Test Plan
-  Passed `yarn run flow`.

### DIFF
--- a/src/components/SecondaryPanes/Expressions.js
+++ b/src/components/SecondaryPanes/Expressions.js
@@ -1,3 +1,4 @@
+// @flow
 import React from "react";
 import { connect } from "react-redux";
 import { bindActionCreators } from "redux";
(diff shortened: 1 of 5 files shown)


arthur801031:
Hi @jasonLaster 

For `Breakpoints.js` to pass flow, we would need to add `source` to type `Location`: https://github.com/devtools-html/devtools-core/pull/193.

src/components/SecondaryPanes/Expressions.js
```diff
@@ -40,13 +41,15 @@ function getValue(expression) {
 }
 
 const Expressions = React.createClass({
+  _input: (null: any),
```
arthur801031: I was getting `property '_input'. Property not found in React  component` errors, and I found this [fix](https://github.com/facebook/flow/issues/1676).

jasonLaster: Yea. Thats what id do too :)

@gregtatum?

gregtatum: I'm confused here. It looks like you are adding `_input` to the prototype because you're setting it to null. You should be able to only do the type definition here, without actually setting a value and the error should go away. Flow treats class definitions as invariants.

```
const Expression = React.createClass({
  _input: ?HTMLInputElement,
```

PR #2037, opened 2017-02-17
move the results summary into the search container

clarkbw:
Actively working on this.

Associated Issue:  #1845

### Summary of Changes

* move the summary info into the search bar container
* added keyshortcut for `Esc`
*

### Test Plan

Example test plan:

- [x] Command-P opens the panel and it looks normal
- [x] cmd+shift+o opens search
- [x] cmd+shift+o, esc closes search

### DIFF
--- a/assets/panel/debugger.properties
+++ b/assets/panel/debugger.properties
@@ -138,7 +138,7 @@ sourceSearch.search.again.key=G
 
 # LOCALIZATION NOTE (sourceSearch.resultsSummary): Shows a summary of
 # the number of matches for autocomplete
-sourceSearch.resultsSummary=%d instances of “%S”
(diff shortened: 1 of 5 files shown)


assets/panel/debugger.properties
```diff
@@ -138,7 +138,7 @@ sourceSearch.search.again.key=G
 
 # LOCALIZATION NOTE (sourceSearch.resultsSummary): Shows a summary of
 # the number of matches for autocomplete
-sourceSearch.resultsSummary=%d instances of “%S”
+sourceSearch.resultsSummary=%d results
```
jasonLaster: I think we should bump the localization key to `sourceSearch.resultsSummary1`

Issue #2041, opened 2017-02-17
Tab Bar Regressions


wldcordeiro:
It seems like we had some regressions with the tab bar. Currently our close button is appearing on the left hand side of the button when it used to appear on the right in `ltr` mode. In `rtl` mode it's even worse right now.

## Close button wrong location

![screenshot_20170216_232152](https://cloud.githubusercontent.com/assets/580982/23055196/bf47b1f0-f4a1-11e6-934c-b9db031f2396.png)

## `rtl` craziness

![screenshot_20170216_233512](https://cloud.githubusercontent.com/assets/580982/23055197/bf6644b2-f4a1-11e6-916a-f2aa2cd87290.png)

PR #2043, opened 2017-02-17
Fix tab close button positioning

wldcordeiro:
Associated Issue: #2041 

### Summary of Changes

* make the tabs into inline flexboxes and remove absolute positioning from the close button.

### Test Plan

Example test plan:

- [x] open debugger
- [x] open some tabs
- [x] see tab close button is in proper position for `ltr` mode (rtl will be in separate PR as there is more broken there.)

#### Before
screenshot_20170216_232152

#### After
screenshot_20170217_030222

Seems like the close button is a little too close to the tab title. 🤔

### DIFF
--- a/src/components/Editor/Tabs.css
+++ b/src/components/Editor/Tabs.css
@@ -35,15 +35,14 @@
   border: 1px solid transparent;
   border-top-left-radius: 2px;
   border-top-right-radius: 2px;
-  height: 26px;
+  height: 30px;
   display: inline-flex;
+  align-items: center;
   position: relative;
   transition: all 0.25s ease;
   min-width: 40px;
   overflow: hidden;
-  padding-top: 4px;
-  padding-inline-start: 12px;
-  padding-inline-end: 20px;
+  padding: 6px;
   margin-inline-start: 8px;
   margin-top: 6px;
 }
@@ -72,15 +71,12 @@ html[dir="rtl"] .source-tab.active {
 }
 
 .source-tab .prettyPrint {
-  display: block;
-  position: absolute;
-  top: 3px;
-  left: 6px;
+  line-height: 0;
 }
 
 .source-tab .prettyPrint svg {
-  height: 1em;
-  width: 1em;
+  height: 12px;
+  width: 12px;
 }
 
 .source-tab .prettyPrint path {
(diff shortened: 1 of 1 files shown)


clarkbw:
Yeah, a little spacing from the title would be good.  Awesome find!

jasonLaster:
Thanks! How's this look with RTL?

princiya:
So the bug is coming from this (potential) explanation: https://github.com/devtools-html/debugger.html/pull/1976/files/368101b4c46ec34a6a2ba129fa7969d6ed65b4be#diff-d9abd0398fe2e0fa5793ba7d100231f5L50
and here: https://github.com/devtools-html/debugger.html/pull/1976/files/368101b4c46ec34a6a2ba129fa7969d6ed65b4be#diff-d9abd0398fe2e0fa5793ba7d100231f5L110

Currently we have,
`.source-tab .close-btn {
  position: absolute;
  top: 3px;
  offset-inline-end: 4px;
  visibility: hidden;
}`

and writing this explicitly as below fixes the issue.
`html[dir="ltr"] .source-tab .close-btn {
  offset-inline-end: 4px;
}`

`html[dir="rtl"] .source-tab .close-btn {
  offset-inline-start: 4px;
}`

jasonLaster:
Thanks @princiya.

I looked at it for a bit and I think we'll want to touch up the margin/padding for the close button and source tab.

We should also look at the pretty print button:
* it's also position absolute
* we shouldn't reserve the 20px space unless it's a pretty source
* we should vertically align the icon so it lines up w/ the text buttons

jasonLaster:
Here's a good screenshot of the two problems I see:

1. the close button should have some more padding
2. the pretty print button should be better aligned

screen shot 2017-02-21 at 11 15 26 am

wldcordeiro:
Fixes were just pushed I added some padding to the close button and fixed up the alignment of the pretty print icon.

screenshot_20170222_034202
screenshot_20170222_034223

src/components/Editor/Tabs.css
```diff
@@ -94,13 +90,20 @@ html[dir="rtl"] .source-tab.active {
...
 
 .source-tab .close-btn {
-  top: 3px;
-  offset-inline-end: 4px;
   visibility: hidden;
+  line-height: 0;
+}
+
+html[dir="ltr"] .source-tab .close-btn {
+  margin-left: 6px;
```
clarkbw: I don't know this stuff well, but could this be `margin-inline-start` instead?

jasonLaster: I think so. I'm going to pull in the latest bidirection and then test

jasonLaster:
This needs the launchpad bump to look right

Issue #2007, opened 2017-02-13
Function Search


jasonLaster:
This is a tracking issue for source function search. It's a really cool feature where we'd love to get some help! If you're interested in helping, ask in the issue and you can take one of the items.

### Checklist

- [x] find functions in search util #1846
- [x] function search UI POC #1995 
- [x] function search UI (for real) #1845 
- [x] move to the SearchBar component #2066
- [x] support scrolling the list of results #2106
- [x] move function declarations out of editor state and to the util #2124
- [x] remove search handler on componentUnMount #2125 
- [x] Clean up styling of result scrolling #2169 
- [x] Only do one search type #2171 
- [x] Add result summary for function search #2172 
- [x] Polish the Search for UI #2203
- [x] Chrome search bar jumps #2204
- [x] Polish the search line positioning #2205
- [x] Fix sad face #2206
- [x] Polish toggling searches #2208

wldcordeiro:
We got the whole list so I'm closing this. If we have reason to reopen we will.

PR #2047, opened 2017-02-17
implement the function search compact design

clarkbw:
Associated Issue:  #2007

### Summary of Changes

* implement the compact design for function search results

### Test Plan

Example test plan:

- [x] Command-P works as normal
- [x] cmd+shift+o shows results in a new layout

screen shot 2017-02-17 at 2 24 59 pm

### DIFF
--- a/src/components/SourceSearch.js
+++ b/src/components/SourceSearch.js
@@ -96,18 +96,21 @@ const Search = React.createClass({
   },
 
   render() {
-    return this.props.searchOn ?
-      dom.div({ className: "search-container" },
+    if (!this.props.searchOn) {
+      return null;
+    }
+
+    return dom.div({ className: "search-container" },
       Autocomplete({
         selectItem: result => {
           this.props.selectSource(result.id);
-          this.setState({ inputValue: "" });
-          this.props.closeFileSearch();
+          this.close();
         },
         close: this.close,
         items: searchResults(this.props.sources),
-        inputValue: this.state.inputValue
-      })) : null;
+        inputValue: this.state.inputValue,
+        placeholder: L10N.getStr("sourceSearch.search")
+      }));
   }
 
 });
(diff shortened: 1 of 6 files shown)


clarkbw:
You'll notice that I went with the right hand side line numbers in a column on the right hand side but all aligned to the left so the `:` is vertically aligned.  This requires a bit more complex flex layout to achieve but I think its better and I believe we wanted to use it other places.

jasonLaster:
This looks great and the cider is straight forward. 

One thought if you have time is that it would be nice to tweak the selected state. 

I'll test soon, but if you could take some screenshots of function search with the dark theme and source search that would help. With source search we just want to prevent regressions

clarkbw:
Yes!  Good call☝️ 

I can back around to this Monday

clarkbw:
The <kbd>cmd</kbd>+<kbd>P</kbd> search looks fine with these changes in both light and dark themes.

Light     | Dark
--------- | ---------
screen shot 2017-02-20 at 2 34 57 pm | screen shot 2017-02-20 at 2 34 36 pm

clarkbw:
The <kbd>cmd</kbd>+<kbd>shift</kbd>+<kbd>o</kbd> search has some issues with the dark theme that I'll look into.

Light     | Dark
--------- | ---------
screen shot 2017-02-20 at 2 40 02 pm | screen shot 2017-02-20 at 2 40 38 pm

clarkbw:
DOCKER!!! 🙊

jasonLaster:
lets see if my branch works: https://circleci.com/gh/jasonLaster/debugger.html/2139

src/components/SourceSearch.js
```diff
@@ -96,18 +96,21 @@ const Search = React.createClass({
   },
 
   render() {
-    return this.props.searchOn ?
-      dom.div({ className: "search-container" },
+    if (!this.props.searchOn) {
+      return null;
+    }
```
jasonLaster: 👍

jasonLaster:
Tests on my branch were green. We can always make follow up improvements

Issue #2059, opened 2017-02-21
yarn test fails locally


clarkbw:
When I run `yarn test` locally it fails consistently.  I get the following errors:

```bash
  80 passing (457ms)
  2 failing

  1) sources should load source text:
     SyntaxError: Unexpected token (1:8)
      at Parser.pp$5.raise (node_modules/babylon/lib/index.js:4333:13)
      at Parser.pp.unexpected (node_modules/babylon/lib/index.js:1705:8)
      at Parser.pp$1.parseFunction (node_modules/babylon/lib/index.js:2322:10)
      at Parser.pp$1.parseFunctionStatement (node_modules/babylon/lib/index.js:1999:15)
      at Parser.pp$1.parseStatement (node_modules/babylon/lib/index.js:1784:19)
      at Parser.parseStatement (node_modules/babylon/lib/index.js:5302:22)
      at Parser.pp$1.parseBlockBody (node_modules/babylon/lib/index.js:2212:21)
      at Parser.pp$1.parseTopLevel (node_modules/babylon/lib/index.js:1723:8)
      at Parser.parse (node_modules/babylon/lib/index.js:1617:17)
      at Object.parse (node_modules/babylon/lib/index.js:6657:37)

  2) sources should cache subsequent source text loads:
     SyntaxError: Unexpected token (1:8)
      at Parser.pp$5.raise (node_modules/babylon/lib/index.js:4333:13)
      at Parser.pp.unexpected (node_modules/babylon/lib/index.js:1705:8)
      at Parser.pp$1.parseFunction (node_modules/babylon/lib/index.js:2322:10)
      at Parser.pp$1.parseFunctionStatement (node_modules/babylon/lib/index.js:1999:15)
      at Parser.pp$1.parseStatement (node_modules/babylon/lib/index.js:1784:19)
      at Parser.parseStatement (node_modules/babylon/lib/index.js:5302:22)
      at Parser.pp$1.parseBlockBody (node_modules/babylon/lib/index.js:2212:21)
      at Parser.pp$1.parseTopLevel (node_modules/babylon/lib/index.js:1723:8)
      at Parser.parse (node_modules/babylon/lib/index.js:1617:17)
      at Object.parse (node_modules/babylon/lib/index.js:6657:37)
```

Can those tests be fixed or disabled?  

I'm doing all pushes with `--no-verify` now because these tests don't pass.  I'm not sure how to fix them.  Thanks!

bomsy:
[...]

PR #2062, opened 2017-02-21
fix the invalid source texts in the failing tests

bomsy:
Associated Issue: #2059

### Summary of Changes

* Fixed the invalid sources
* Supports #2061


jasonLaster:
yep. i remember making this change... I also want to add the try catch because we shouldn't blowup ever

Issue #1617, opened 2017-01-04
persist scopes state while stepping


tromey:
I set a breakpoint and the debugger stopped there.  I opened a scope in the Scopes pane to see the local variables.  Then I did a step over.  The debugger collapsed the scope.  I think it would be nicer if the scope stayed open.  As it is now, if I want to see what is happening, I have to switch back and forth between the step-over button and the scope button; and if the information I want is buried deeper (say, a property a few levels deep in some object), this is a pain.

jasonLaster:
This is similar for maintaining expanded state for watch expression objects

CC @bomsy

jasonLaster:
@bomsy any updates?

bomsy:
Hey ... yeah been thinking about what the best way to store the state of a particular path. Coming up with something

jasonLaster:
sounds good - definitely an interesting problem!

jasonLaster:
@bomsy we *might* get this for free. try investigating a bit...

bomsy:
will do!

jasonLaster:
Hmm I looked into this and found some encouraging results:

#### Situations:
* selecting frames - works
* stepping - doesn't work

#### selecting frames
![](http://g.recordit.co/whuYKpJUI0.gif)

#### stepping
![](http://g.recordit.co/lwU6aBqJyv.gif)

jasonLaster:
@bomsy dropping the in progress label because it's gone stale. feel free to pickup again though

bomsy:
picking this back up! some nice ideas to try out!

PR #2064, opened 2017-02-21
Persist scopes

bomsy:
Associated Issue: #1617

### Summary of Changes

* Added get and set functions which enable us cache the `actors` and `expanded` states across `ObjectInspector` component mounts.
* Fixed a key bug in the `getChildren` function. The loaded children were cached against the `actor` rather than the `key`. ( Finding this was the nightmare)  :)

### Test Plan

- [x] Set a breakpoint
- [x] Once the breakpoint is hit
- [x] Open up the scopes
- [x] Hit the step over button
- [x] The state should be preserved in the scopes pane

persist-scopes

### DIFF
--- a/src/components/shared/ObjectInspector.js
+++ b/src/components/shared/ObjectInspector.js
@@ -54,7 +54,11 @@ const ObjectInspector = React.createClass({
     getObjectProperties: PropTypes.func.isRequired,
     loadObjectProperties: PropTypes.func.isRequired,
     onLabelClick: PropTypes.func,
-    onDoubleClick: PropTypes.func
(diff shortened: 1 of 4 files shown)


src/components/shared/ObjectInspector.js
```diff
@@ -106,7 +117,7 @@ const ObjectInspector = React.createClass({
       }
 
       const children = makeNodesForProperties(loadedProps, item.path);
-      this.actorCache[actor] = children;
+      this.actors[key] = children;
```
bomsy: This was the nightmare bug!! :)

Issue #1946, opened 2017-02-06
[Editor] Add Conditional Breakpoint Panel guillemet


jasonLaster:
It would be nice to show a [guillemet] in the gutter next to the input to make it clear that you can type there

The important thing is that we style it too. If we need to use an SVG, we can do that too

```diff
diff --git a/src/components/Editor/ConditionalPanel.js b/src/components/Editor/ConditionalPanel.js
index 385f934..492f6a8 100644
--- a/src/components/Editor/ConditionalPanel.js
+++ b/src/components/Editor/ConditionalPanel.js
@@ -22,6 +22,7 @@ function renderConditionalPanel({ condition, closePanel, setBreakpoint }:
   ReactDOM.render(
     dom.div(
       { className: "conditional-breakpoint-panel" },
+      dom.div({ className: "prompt" }, &raquo;),
       dom.input({
         defaultValue: condition,
         placeholder: L10N.getStr("editor.conditionalPanel.placeholder"),
```

### Screenshot

![](https://cloud.githubusercontent.com/assets/254562/19328920/170c493a-90a4-11e6-907b-4a393e5ccd36.png)

[guillemet]: http://www.fileformat.info/info/unicode/char/bb/index.htm

soapdog:
I am working on this one! 😸

wldcordeiro:
I'm going to leave this issue open since we discussed moving the guillemet to the gutter itself. Thanks for the work @soapdog

jasonLaster:
Here's a diff to show where to start on adding the guillemet to the marker

```diff
diff --git a/src/components/Editor/Breakpoint.js b/src/components/Editor/Breakpoint.js
index aea776b..e3f2654 100644
--- a/src/components/Editor/Breakpoint.js
+++ b/src/components/Editor/Breakpoint.js
@@ -31,11 +31,13 @@ const Breakpoint = React.createClass({
     const bp = this.props.breakpoint;
     const line = bp.location.line - 1;

+    // => this is where the marker goes. we could do the same thing with the guillemet
     this.props.editor.setGutterMarker(
       line,
       "breakpoints",
       makeMarker(bp.disabled)
     );
+
     this.props.editor.addLineClass(line, "line", "new-breakpoint");
     if (bp.condition) {
       this.props.editor.addLineClass(line, "line", "has-condition");
```

aklt:
I'll have a go at this one

PR #2088, opened 2017-02-22
Issue 1946: adding Conditional Breakpoint guillemet

soapdog:
Associated Issue: #1946 

### Summary of Changes

* Moved Conditional Breakpoint Panel style from editor.css to its own file at ConditionalPanel.css
* Changed ConditionalPanel.js to include the guillemet

### Test Plan

I used the instructions on the wiki to run the source code and saw the guillemet in the correct position.

Example test plan:

- [x] Launch debugger
- [x] Navigate to desired testing page
- [x] User Control+Shift+B to add conditional breakpoint
- [x] Admire the beautiful guillemet in the panel

Feel free to improve it!

### Screenshots/Videos
slack - firefox devtools 2017-02-22 20 37 49

### DIFF
--- a/src/components/Editor/ConditionalPanel.js
+++ b/src/components/Editor/ConditionalPanel.js
@@ -6,7 +6,6 @@ const ReactDOM = require("react-dom");
 
 require("./ConditionalPanel.css");
 
-
 function renderConditionalPanel({ condition, closePanel, setBreakpoint }:
   { condition: boolean, closePanel: Function, setBreakpoint: Function }) {
   let panel = document.createElement("div");
--- a/src/components/Editor/ConditionalPanel.js
+++ b/src/components/Editor/ConditionalPanel.js
@@ -4,6 +4,9 @@ const { DOM: dom } = React;
 
 const ReactDOM = require("react-dom");
 
+require("./ConditionalPanel.css");
+
+
 function renderConditionalPanel({ condition, closePanel, setBreakpoint }:
   { condition: boolean, closePanel: Function, setBreakpoint: Function }) {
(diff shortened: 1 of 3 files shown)


src/components/Editor/ConditionalPanel.js
```diff
@@ -22,6 +24,10 @@ function renderConditionalPanel({ condition, closePanel, setBreakpoint }:
   ReactDOM.render(
     dom.div(
       { className: "conditional-breakpoint-panel" },
+      dom.div({
+        className: "prompt",
+        dangerouslySetInnerHTML: { __html: "&raquo;" }
```
wldcordeiro: Spoke to @jasonLaster and we could likely use the unicode character itself rather than `dangerouslySetInnerHTML`. But that's a minor change. 😄

jasonLaster:
Thanks @soapdog!

This looks really good. The design calls for the guillemet in the gutter, but I don't mind it here either.



Up to you on whether you want to take on the challenge of adding an element to the gutter like debugger breakpoints.

Personally, i think this might be a good V1.

src/components/Editor/ConditionalPanel.js
jasonLaster: yup, this is how we do it in another spot:
https://github.com/devtools-html/debugger.html/blob/master/src/components/shared/Dropdown.js#L42

soapdog:
@jasonLaster oops, totally missed the gutter part. Maybe we could merge this one as a V1 and then I take my time trying to add it to the gutter and pushing a new PR later, I am new to react and this might take a while.

Issue #1840, opened 2017-01-25
[EditorHover] Show a popup in CodeMirror


jasonLaster:
We want to show an inline codemirror popup for variables when you hover on an expression. This step however is just adding an inline popup that looks nice.

Lets just open an empty popup if you shift click.

It should have similar code to [conditional panes][cbp]

Here's a patch to help get it started:

```diff
diff --git a/src/components/Editor/ConditionalPanel.js b/src/components/Editor/ConditionalPanel.js
index 385f934..aa90b03 100644
--- a/src/components/Editor/ConditionalPanel.js
+++ b/src/components/Editor/ConditionalPanel.js
@@ -19,6 +19,7 @@ function renderConditionalPanel({ condition, closePanel, setBreakpoint }:
     }
   }

+  // this is how we create the panel in a kinda reacty way
   ReactDOM.render(
     dom.div(
       { className: "conditional-breakpoint-panel" },
diff --git a/src/components/Editor/index.js b/src/components/Editor/index.js
index c0dde71..211ba0a 100644
--- a/src/components/Editor/index.js
+++ b/src/components/Editor/index.js
@@ -210,7 +210,6 @@ const Editor = React.createClass({
     menuOptions.push(showSourceMenuItem);

     showMenu(event, menuOptions);
-
   },

   onGutterContextMenu(event) {
@@ -241,6 +240,7 @@ const Editor = React.createClass({
       });
     };

+    // => this is where we show the panel
     const panel = renderConditionalPanel({
       condition,
       setBreakpoint,
```

![screen shot 2017-02-02 at 5 48 40 pm](https://cloud.githubusercontent.com/assets/254562/22571914/de8de44a-e96f-11e6-8a26-845d67f11b3a.png)

[cbp]:https://github.com/devtools-html/debugger.html/blob/master/src/components/Editor/ConditionalPanel.js#L1

amitzur:
I did something, I'll share an animated gif later today. But wanted to ask:
Was there a reason not to bring in conditional breakpoints into local state of the Editor component? Or even into the store?
Right now it's imperative, which does feel a bit hacky. Trying to mimic this in a popover resulted also a bit hacky, and I wanted to have state on the editor that contains data related to popover.

I wouldn't really put it in the store, though :)

jasonLaster:
I believe we use the property `cbPanel` over state for two reasons:
1. we want the state to be JSON serializable
2. we don’t want to trigger a re-render

[...]

PR #2090, opened 2017-02-23
Editor popover

amitzur:
Associated Issue: #1840 

### Summary of Changes

* a very initial popover component. I plan to make changes to almost everything there :)
* on hover and pause for 500ms over the editor - show the popover
* a small refactor that extracted codeMirror (and its wrapper element) into a variable instead of accessing it repeatedly with this.editor

### Test Plan

no tests currently, I saw that ui components aren't covered yet


jasonLaster:
Thanks @amitzur. I like the cleanup as well. componentDidMount is quite unrulely.

I'll help land this tomorrow

PR #2114, opened 2017-02-23
clean out the utils/utils module using lodash utils where possible

clarkbw:
Associated Issue: #???

### Summary of Changes

* drop unused utils functions
* convert used utils functions to lodash equivalents
* optimize lodash webpacking by using `lodash/function` where possible

### Test Plan

- [x] project wide search for util function
- [x] run tests
- [x] check that debugger works as expected

### DIFF
--- a/bin/getConfig.js
+++ b/bin/getConfig.js
@@ -1,4 +1,4 @@
-const merge = require("lodash").merge;
+const merge = require("lodash/merge");
 const fs = require("fs");
 const path = require("path");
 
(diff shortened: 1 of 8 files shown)


jasonLaster:
@clarkbw looks like the `browser_dbg-sources.js` test is failing. try debugging it in `localhost:8000/integration`

clarkbw:
#2122 is going to do the lodash throttle investigation

clarkbw:
Ok, looking much better now

Issue #2078, opened 2017-02-22
[EditorSearch] Add search bottom bar


jasonLaster:
This tracks the work needed to add a search bottom bar:

- [x] search modifiers on the bottom right
- [x] function search toggle button bottom left

Notes:
* helen's design called for the search button being a toggle for the bar, but i'm not sure if that is clear enough
* when this is added, we'll want to disable modifiers when doing a function search, but that's also out of scope

### Screenshot
![](https://cloud.githubusercontent.com/assets/254562/22855693/3bc6fe24-f052-11e6-862d-922fa8183bb4.jpg)

wldcordeiro:
The button was added. There is additional work being done to move the function search UI itself into the search bar now and additional refactoring to simplify our two distinct search types project and document.

Issue #2000, opened 2017-02-11
[EditorSearch] move search modifiers to bottom bar


jasonLaster:
@wldcordeiro did a great job of adding search modifiers. It was NOT easy :)

Now we've got to polish the UI so that we can land them.

I did some quick sketches that I'd love some feedback on and help implementing. The sketches have spots for function search and variable search, but we don't need to add those now.

![modifier-ui](https://cloud.githubusercontent.com/assets/254562/22855651/d6e28e2e-f051-11e6-9bfd-9fd658e444c1.jpg)

PR #2119, opened 2017-02-23
Add a search bottom bar and move the search modifiers to the bottom bar.

wldcordeiro:
Associated Issue: #2078 and #2000

### Summary of Changes

* Add a bottom bar to the searchbar
* Move the modifiers to the bottom bar

### Test Plan

- [x] Open debugger and pick a file
- [x] Open the searchbar with <kbd>ctrl</kbd>/<kbd>cmd</kmd>+f
- [x] See that there is a bottom bar (need to have search modifiers or function search enabled)

### Screenshots/Videos

screenshot_20170223_162245

@jasonLaster I know we want the modifiers on the right side and I'll get that once function search is moved too. 😄

### DIFF
--- a/src/components/Editor/SearchBar.css
+++ b/src/components/Editor/SearchBar.css
@@ -1,4 +1,9 @@
 .search-bar {
+  display: flex;
+  flex-direction: column;
+}
+
+.search-field {
   width: calc(100% - 1px);
   height: 40px;
   background-color: var(--theme-body-background);
@@ -6,17 +11,17 @@
   display: flex;
 }
 
-.search-bar i {
+.search-field i {
   display: block;
   padding: 13px;
   width: 40px;
 }
 
-.search-bar i svg {
+.search-field i svg {
   width: 16px;
 }
 
-.search-bar input {
+.search-field input {
   border: none;
   line-height: 30px;
   font-size: 14px;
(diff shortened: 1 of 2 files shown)


src/components/Editor/SearchBar.js
```diff
@@ -332,6 +322,29 @@ const SearchBar = React.createClass({
...
+  },
+
+  render() {
+    if (!this.state.enabled) {
+      return dom.div();
+    }
+
+    return dom.div(
+      { className: "search-bar" },
+      this.renderSearchField(),
```
jasonLaster: hmm, how aboutt `searchInput`

src/components/Editor/SearchBar.css
```diff
@@ -26,39 +31,48 @@
...
+  border-bottom: 1px solid var(--theme-splitter-color);
+}
+
+.search-mod-bar .search-modifiers {
   display: flex;
   align-items: center;
 }
 
-.search-bar .search-modifiers button {
+.search-mod-bar .search-modifiers button {
```
jasonLaster: hmm, perhaps we want to call it `search-bottom-bar`

wldcordeiro: I had done that first and renamed it because I didn't want to type as much 😆

src/components/Editor/SearchBar.js
wldcordeiro: I considered that but we have `this.searchInput` that I didn't want to confuse it with.

PR #2122, opened 2017-02-24
WIP look into using lodash throttle

clarkbw:


src/components/SourcesTree.js
```diff
@@ -41,15 +41,15 @@ let SourcesTree = React.createClass({
...
 
-  queueUpdate: throttle(function() {
+  queueUpdate: debounce(function() {
     if (!this.isMounted()) {
       return;
     }
 
     this.forceUpdate();
-  }, 50),
+  }, 100, { maxWait: 250, leading: false, trailing: true }),
```
clarkbw: This means that we debounce for all requests happening within 100 of each other. However we'll never go more than 250 before running the `forceUpdate`.  So this becomes the combination of throttle and debounce, hopefully for the experience we want.

clarkbw:
I can't seem to kick the CI right now. 

But in the broader sense this code isn't necessary right now.  We want to get the release out the door and this is only adding risk for very little reward.

clarkbw:
I'll pick this back up again in a couple weeks.  Anyone else is free to also grab this after we release.

Issue #1918, opened 2017-02-03
Command Bar should be fixed to the top


yurydelendik:
We currently scroll the entire secondary pane. We should try and *fix* the command bar to the top, so that it is always visible. [see gif]

@garbee suggested that we could make the command bar sticky.

```diff
diff --git a/src/components/SecondaryPanes/CommandBar.css b/src/components/SecondaryPanes/CommandBar.css
index e79dca3..2845e78 100644
--- a/src/components/SecondaryPanes/CommandBar.css
+++ b/src/components/SecondaryPanes/CommandBar.css
@@ -2,6 +2,7 @@
   flex: 0 0 30px;
   border-bottom: 1px solid var(--theme-splitter-color);
   display: flex;
+  position: sticky;
 }

 .command-bar > button {
 ```

![](http://g.recordit.co/SSspwofVJF.gif)

---

STR:
1. Open and start debugging a web page with debugger statement (e.g. https://devtools-html.github.io/debugger-examples/examples/localstorage-keys.html)
2. On first debugger statement, expand the "Window" scope at the bottom
3. Scroll the secondary panel all the way to the bottom

Actual result:
Debugger control toolbar/buttons are hidden

Expected result:
Debugger control toolbar/buttons are visible

It's a regression, bisect results:

```
0c5485c6d3db4e60acae756c33a8441206392052 is the first bad commit
commit 0c5485c6d3db4e60acae756c33a8441206392052
Author: Jonathan Garbee <jonathan@garbee.me>
Date:   Tue Jan 31 09:39:32 2017 -0500

    Fix SecondaryPanes contents to flex properly. (#1893)
```

/cc @Garbee #1893

Garbee:
So, what exactly does this look like scrolled? Do we have the contents of each accordion with a set height and make them scrollable? Or just scroll the pane itself but keeping the commandbar and why paused pieces fixed to the top?

If it is about fixing those two pieces, then the fix is trivial by just applying `position:sticky` here. If we want to individually scroll each accordion's content without causing the headers to get lossed, then I'm going to need to investigate this more to find a decent solution.

wldcordeiro:
I think the command bar isn't remaining fixed like it used to be @Garbee

Garbee:
Discussing this in slack just now it looks like we only need to keep the CommandBar itself stickied and not the `WhyPaused` bit or anything else.

clarkbw:
I was able to make this by adding this:

```css
.command-bar {
[...]

PR #2128, opened 2017-02-24
Stick the commandbar to the top of the seconday panel.

Garbee:
Associated Issue: #1918 

### Summary of Changes

* Sticky `.command-bar` to the top using `position: sticky`.
* z-index so the static elements don't go on top of it.
* Background-color to prevent the command bar from showing the scrolled elements underneath it.

### Test Plan

Start up debugger, set a breakpoint, open the `window` scope variables. Scroll some.

### DIFF
--- a/src/components/SecondaryPanes/CommandBar.css
+++ b/src/components/SecondaryPanes/CommandBar.css
@@ -4,6 +4,10 @@
   display: flex;
   height: 30px;
   overflow: hidden;
+  position: sticky;
+  top: 0;
+  z-index: 1;
+  background-color: var(--theme-body-background);
 }
 
 .command-bar > button {


src/components/SecondaryPanes/CommandBar.css
```diff
@@ -4,6 +4,10 @@
   display: flex;
   height: 30px;
   overflow: hidden;
+  position: sticky;
+  top: 0;
+  z-index: 1;
+  background-color: white;
```
irfanhudda: We should use some variable (`--theme-body-background`?) to make this work with all themes.

Garbee: oh right. Is that the var for the color? The naming scheme for them makes no sense so I have no clue what they are actually called.

Garbee: Updated, that wast he right variable name. Good catch!

PR #2133, opened 2017-02-24
Test Return values

jasonLaster:
### Summary of Changes

* adds a test for return values
* adds pauseTest, assertNotPaused
* cleans up the utils
* shows return false, 0

### DIFF
--- a/src/test/integration/utils/index.js
+++ b/src/test/integration/utils/index.js
@@ -1,93 +1,13 @@
-const {
-  waitForPaused,
-  waitForDispatch,
-  waitForTime,
-  waitForSources,
-  waitForElement,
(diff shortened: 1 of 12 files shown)


src/test/integration/utils/index.js
```diff
@@ -1,93 +1,13 @@
...
-  pressKey,
-  waitForPaused,
-  waitForDispatch,
-  waitForTime,
-  waitForSources,
-  waitForElement,
-  waitForTargetEvent,
-  waitForThreadEvents,
-  waitUntil,
-}
```
jasonLaster: this part makes me happy. I always found it frustrating to add these

Issue #374, opened 2016-07-20
[Scopes] populate promises


clarkbw:
Promises have special properties we want to display in the variables view for inspection.

*(this is one of the cooler bugs imo)*

#### Old Debugger
Here is where the [old debugger](https://dxr.mozilla.org/mozilla-central/source/devtools/client/shared/widgets/VariablesView.jsm#3638-3645) got the properties.

Here's what the code looked like:
```js
    if (aGrip.class == "Promise" && aGrip.promiseState) {
      let { state, value, reason } = aGrip.promiseState;
      props.push("<state>: " + VariablesView.getString(state));
      if (state == "fulfilled") {
        props.push("<value>: " + VariablesView.getString(value, { concise: true }));
      } else if (state == "rejected") {
        props.push("<reason>: " + VariablesView.getString(reason, { concise: true }));
      }
    }
```

#### Sketch

```diff
diff --git a/src/components/Scopes.js b/src/components/Scopes.js
index 97ed485..d6e93bf 100644
--- a/src/components/Scopes.js
+++ b/src/components/Scopes.js
@@ -72,6 +72,16 @@ function getThisVariable(frame, path) {
   };
 }

+// i could be wrong about formatting each variable...
+function getPromiseProperties(variable) {
+  if (variable.class != "Promise" || !variable.promiseState) {
+    return variable;
+  }
+
+  // old debugger code here
+  // <value>, <reason>, <state>
+}
+
 function getScopes(pauseInfo, selectedFrame) {
   if (!pauseInfo || !selectedFrame) {
     return null;
@@ -116,6 +126,8 @@ function getScopes(pauseInfo, selectedFrame) {
       }

       if (vars && vars.length) {
+        vars = vars.map(variable => getPromiseProperties(variable));
+
         vars.sort((a, b) => a.name.localeCompare(b.name));
         scopes.push({ name: title, path: key, contents: vars });
       }
```

#### Demo code
I made this demo with this example. Maybe it was stupid, but I added the promises to todo-view#toggleCompleted for fun :)

```diff
diff --git a/src/test/examples/todomvc/js/views/todo-view.js b/src/test/examples/todomvc/js/views/todo-view.js
index 0b9e93d..adc1c5c 100644
--- a/src/test/examples/todomvc/js/views/todo-view.js
+++ b/src/test/examples/todomvc/js/views/todo-view.js
@@ -67,6 +67,13 @@ var app = app || {};

                // Toggle the `"completed"` state of the model.
[...]

PR #2146, opened 2017-02-25
[Scopes] populate promises

arthur801031:
Associated Issue: #374

### Summary of Changes
* Add code to insert `reason` to `Promise` objects.

### Test Plan
Check if `Promise` objects contain `reason` in `Scopes`.

### DIFF
--- a/src/actions/pause.js
+++ b/src/actions/pause.js
@@ -215,7 +215,8 @@ function loadObjectProperties(object: any) {
     dispatch({
       type: constants.LOAD_OBJECT_PROPERTIES,
       objectId,
-      [PROMISE]: client.getProperties(object)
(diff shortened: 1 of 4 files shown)


arthur801031:
Here is what we have compared to `Chrome`:

**Our debugger**
screen shot 2017-02-25 at 11 15 19 am

**Chrome**
screen shot 2017-02-25 at 11 17 04 am

jasonLaster:
Thanks for the comparison and the great test. 

I think the last example with the rejected promise could have the value undefined.

jasonLaster:
I think we can try to improve `getchildren` to do this:

https://github.com/jasonLaster/debugger.html/tree/promise2

Issue #2148, opened 2017-02-25
[panel] pause on exceptions button is blue on load


jasonLaster:
![screen shot 2017-02-25 at 6 40 19 pm](https://cloud.githubusercontent.com/assets/254562/23335752/e7a0bb80-fb89-11e6-99b6-0a3b802b1468.png)

najamkhn:
Seems like it persists its last state. There seems to be three states that it cycles around i.e,
* `pause-exceptions enabled`
* `pause-exceptions uncaught enabled`
* `pause-exceptions all enabled`

I tried it with all of them and it seems to just persist the values above. Shouldn't it be the case?

jasonLaster:
sorry, that wasn't clear. This is an issue that only happens when the debugger is in the firefox panel.

I'm working on a fix now, but forgot to mark it in-progress

PR #2165, opened 2017-02-26
Fix POE state on panel load

jasonLaster:
Associated Issue: #2148

### Summary of Changes

* moved the pref loading to `onFirefoxConnect` which is called when the panel starts

### Test Plan

* adding a new test


jasonLaster:
closing this as it was fixed in a different PR. I'll land some of these test fixes later

PR #2168, opened 2017-02-26
Fix call to `scrollList` to pass the right ref.

wldcordeiro:
Associated Issue: N/A

### Summary of Changes

* Missed this reference to the right refs, fixed the call

### Test Plan

- [ ] Run function search and try scrolling up/down the list.


jbhoosreddy:
@wldcordeiro thanks for finding this!

Issue #2169, opened 2017-02-27
[FunctionSearch] Set a max height on the result list


wldcordeiro:
Currently the result list will grow to take up the whole editor, make it so it only grows to a certain height and adds a scrollbar.

PR #2170, opened 2017-02-27
Style the searchbar height so the result list doesn't go on forever

wldcordeiro:
Associated Issue: #2169

### Summary of Changes

* Add a max-height to the searchbar
* Disable shrinking on the bottom bar to avoid squashing by result list

### Test Plan

- [x] Open function search
- [x] Search for some functions
- [x] See that the height of the list doesn't go over ~50%

### Screenshots/Videos

screenshot_20170226_170528


jbhoosreddy:
LGTM!

PR #2181, opened 2017-02-27
Add some maintainer docs

jasonLaster:
#### Summary of Changes

You've probably felt it, but the Debugger is growing and it's more important than ever that we work as a team to help maintain the project.

screen shot 2017-02-27 at 11 32 38 am

Here are a collection of tips for maintaining as i think about it. It'd be great to evolve this doc as we evolve other docs like development, issues, PR tips.


sole:
This is very nice. I've a couple of suggestions, but I will send a PR instead of listing them here as it'll be faster! ✌🏼

Issue #2014, opened 2017-02-13
[Frames] select frame on enter


jasonLaster:
We recently improved the selection state of the Call Stack. The next step to make the component accessible is to support selecting the frame on when enter is pressed.

### GIF

The GIF here shows that you need to click on the frame to select it

![](http://g.recordit.co/HTg2ouQ7Vn.gif)

### Patch

I might have gotten carried away with `isEnter`. Up to you whether it's helpful :)

```diff
diff --git a/src/components/SecondaryPanes/Frames.js b/src/components/SecondaryPanes/Frames.js
index ff4bdde..b0cf710 100644
--- a/src/components/SecondaryPanes/Frames.js
+++ b/src/components/SecondaryPanes/Frames.js
@@ -99,8 +99,9 @@ const Frames = createClass({
         className: classNames("frame", {
           "selected": selectedFrame && selectedFrame.id === frame.id
         }),
-        onMouseDown: (e) => this.onMouseDown(e, frame, selectedFrame),
-        onContextMenu: (e) => this.onContextMenu(e, frame),
+        onMouseDown: e => this.onMouseDown(e, frame, selectedFrame),
+        onContextMenu: e => this.onContextMenu(e, frame),
+        okKeyPress: e => this.onKeyPress(e, frame),
         tabIndex: 0
       },
       renderFrameTitle(frame),
@@ -108,6 +109,14 @@ const Frames = createClass({
     );
   },
 
+  onKeyDown(e: SyntheticKeyboardEvent, frame: Frame) {
+    e.preventDefault();
+
+    if (isEnter(e)) {
+      this.props.selectFrame(frame);
+    }
+  },
+
   onMouseDown(e: SyntheticKeyboardEvent, frame: Frame, selectedFrame: Frame) {
     if (e.nativeEvent.which == 3 && selectedFrame.id != frame.id) {
       return;
```

roieu:
Taking my shot with this one at #goodnessSquad event

lasfin:
Am I right what we should hover on component and press Enter to select?

jasonLaster:
@lasfin i think so, the idea is that we should handle the `<enter>` keypress and select the frame. the code snippet above is a good description.

PR #2182, opened 2017-02-27
[Frames] select frame on enter

lasfin:
Associated Issue: #2014

### Summary of Changes

* add keydown event listener to Frames component
* save hovered frame in component state on mouseEnter and delete it from state on mouseLeave

### DIFF
--- a/src/components/SecondaryPanes/Frames.js
+++ b/src/components/SecondaryPanes/Frames.js
@@ -56,11 +56,11 @@ const Frames = createClass({
   },
 
   componentWillMount() {
-    document.addEventListener("keydown", this.onKeyDown, false);
+    document.addEventListener("keyup", this.onKeyUp, false);
   },
 
   componentWillUnmount() {
-    document.removeEventListener("keydown", this.onKeyDown, false);
+    document.removeEventListener("keyup", this.onKeyUp, false);
   },
 
   getInitialState() {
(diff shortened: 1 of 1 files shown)


src/components/SecondaryPanes/Frames.js
```diff
@@ -55,6 +55,14 @@ const Frames = createClass({
       || showAllFrames !== nextState.showAllFrames;
   },
 
+  componentWillMount() {
+    document.addEventListener("keydown", this.onKeyDown, false);
```
wldcordeiro: Don't know if we want to use document level event listeners here @jasonLaster

jasonLaster: Hmm, we should react to assign the `onKeyUp`

src/components/SecondaryPanes/Frames.js
```diff
@@ -55,6 +55,14 @@ const Frames = createClass({
       || showAllFrames !== nextState.showAllFrames;
   },
 
+  componentWillMount() {
+    document.addEventListener("keyup", this.onKeyUp, false);
+  },
+
+  componentWillUnmount() {
+    document.removeEventListener("keyup", this.onKeyUp, false);
```
jasonLaster: Hmm, you should try to move this to `renderFrameList` where the UL is render and add attribute: `onKeyUp`

lasfin: But if I will move this, it will not listen keyUp's, until we click on some frame.
It will work fine: "click on some frame, then hover another and press Enter", and in this case "hover some frame and press Enter", onKeyUp event will not be fire for component.

jasonLaster: this doesn't matter for hover and enter. it's for tab and enter

jasonLaster:
this looks right @lasfin! I'll test soon

jasonLaster:
Works great! Thanks so much

PR #2187, opened 2017-02-27
(WIP) Preview different things

jasonLaster:
Associated Issue: #2158

### Summary of Changes
* show the function link in the header
* show the children as roots to remove one of the depth levels
* highlight a selected node
* use reps for primitive nodes

todolist

- [x] use reps to format numbers and strings and other things
- [x] clean up styling in general

#### Strings
strings

### Objects
object

### Functions
function

### DIFF
--- a/src/components/Editor/Preview.css
+++ b/src/components/Editor/Preview.css
@@ -1,11 +1,20 @@
 .preview {
-
+  background: white;
+  min-width: 200px;
+  min-height: 80px;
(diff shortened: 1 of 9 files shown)


src/components/Editor/Preview.css
```diff
@@ -0,0 +1,45 @@
+.preview {
```
wldcordeiro: Empty rule breaks lint.

Issue #2199, opened 2017-02-28
Search show a flashing "NaN of 1"


nchevobbe:
While entering random search string, I briefly saw a `NaN of 1` in the search bar
![peek 28-02-2017 08-56](https://cloud.githubusercontent.com/assets/578107/23397755/22f6f3f4-fd9a-11e6-8e89-1b39e88229eb.gif)

Issue #2105, opened 2017-02-23
[EditorSearch] breaks with "whole word" modifier and query's of 3 or less characters


jasonLaster:

najamkhn:
I'll take a look at it. :)

wldcordeiro:
@najamkhn Thanks! I'll be tracking this as well so feel free to ask any questions. 😄

PR #2200, opened 2017-02-28
Fix whole word search modifier clearing search and count showing as NaN

wldcordeiro:
Associated Issue: #2105 #2199 

### Summary of Changes

* Match against the built query in `searchOverlay` rather than the raw query
* Initialize our matchIndex as zero rather than null

### Test Plan

- [x] Open debugger and document
- [x] Run a code search with whole word modifier enabled
- [x] See that the search runs and doesn't clear document
- [x] See that `NaN` isn't flashed in the `n of m results` message

### DIFF
--- a/src/utils/editor/build-query.js
+++ b/src/utils/editor/build-query.js
@@ -15,7 +15,7 @@ type QueryOptions = {
  * @static
  */
 function ignoreWhiteSpace(str: string): string {
-  return /^\s{0,2}$/.test(str) ? "(?!\s*.*)" : str;
+  return /^\s{0,2}$/.test(str) ? "(?!\\s*.*)" : str;
 }
 
 function wholeMatch(query: string, wholeWord: boolean): string {
--- a/src/utils/editor/source-search.js
+++ b/src/utils/editor/source-search.js
@@ -94,7 +94,7 @@ function searchOverlay(query, modifiers) {
         matchLength = len;
         return "highlight highlight-start";
       }
-      while (!stream.match(query, false) && stream.peek()) {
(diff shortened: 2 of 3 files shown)


src/utils/editor/build-query.js
```diff
@@ -15,7 +15,7 @@ type QueryOptions = {
  * @static
  */
 function ignoreWhiteSpace(str: string): string {
-  return /^\s{0,2}$/.test(str) ? "(?!\s*.*)" : str;
+  return /^\s{0,2}$/.test(str) ? "(?!\\s*.*)" : str;
```
wldcordeiro: The slash needs to be escaped here, noticed we didn't.

src/utils/editor/source-search.js
```diff
@@ -94,7 +94,7 @@ function searchOverlay(query, modifiers) {
         matchLength = len;
         return "highlight highlight-start";
       }
-      while (!stream.match(query, false) && stream.peek()) {
+      while (!stream.match(regexQuery, false) && stream.peek()) {
```
wldcordeiro: I've spent entirely too long looking for this, but this was the fix for whole word matching... 😆

src/utils/editor/source-search.js
```diff
@@ -147,7 +147,7 @@ function getMatchIndex(count: number, currentIndex: number, rev: boolean) {
  */
 function doSearch(ctx, rev, query, keepSelection, modifiers: SearchModifiers) {
   let { cm } = ctx;
-  let matchIndex;
+  let matchIndex = 0;
```
wldcordeiro: Initializing to null meant we had a brief flash of `NaN` or `-NaN`

Issue #2208, opened 2017-02-28
[FunctionSearch] Polish toggling searches


jasonLaster:
I'm seeing some funny behavior with toggling searches

Checklist:

- [ ] cmd+f should always show the "search in file..."
- [ ] cmd+shif+o should always show "search functions"
- [ ] close btn and escape should always close and clear queries...
- [ ] the "function" button should jump between function/text search

PR #2211, opened 2017-02-28
Polish search toggle UX

wldcordeiro:
Associated Issue: #2208 

### Summary of Changes

* <kbd>ESC</kbd> and the close button both clear and close search
* <kbd>ctrl</kbd>/<kbd>cmd</kbd>+<kbd>shift</kbd>+o always opens function search
* function search button jumps between function and code search

### Test Plan

- [x] Open debugger with document
- [x] Try <kbd>ctrl</kbd>/<kbd>cmd</kbd>+<kbd>shift</kbd>+o to see that it just toggles the function search on
- [x] <kbd>ctrl</kbd>/<kbd>cmd</kbd>+f to see that it just toggles the code search on
- [x] Click the close button and escape to see that the query is cleared

### DIFF
--- a/src/components/Editor/SearchBar.css
+++ b/src/components/Editor/SearchBar.css
@@ -15,6 +15,10 @@
   padding: 0 13px;
 }
 
+.search-bottom-bar button:focus {
+  outline: none;
+}
+
 .search-bottom-bar .search-modifiers {
   display: flex;
   align-items: center;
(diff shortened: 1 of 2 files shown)


wldcordeiro:
So I added the query clearing but since we still look for a selection to fill in it's using it. Looking into the fix now. 👍

PR #2217, opened 2017-02-28
Extract getScopes + Refactor returnvalue tests

jasonLaster:
associated: #2218

### Summary of Changes

* extract getScopes so that it can be tested and refactored a bit for getting binding variables out of the scope chain.

* refactor return value tests

### Test Plan

mochitests

screen shot 2017-02-28 at 6 34 27 pm


src/test/integration/tests/returnvalues.js
```diff
@@ -70,11 +70,6 @@ module.exports = async function(ctx) {
   toggleScopes(dbg);
   await togglePauseOnExceptions(dbg, true, false);
 
-  const TESTS = ["57", "0", "false", "undefined", "null", "\"undefined\""];
```
tromey: This change loses the tests for 0 and null, but I think those are important to keep as they are falsy values, and so easy to get wrong.  Probably we should even expand this to add the empty string, which I forgot about in the earlier patch.

jasonLaster: 👍  I can bring those in. it'd be nice to have the grips for those as well

jasonLaster:
Here are some improvements to add additional falsey values and clean up the tests/code accordingly.

screen shot 2017-03-01 at 12 34 45 pm

Issue #2075, opened 2017-02-22
[SourceTabs] Sources Dropdown hidden in RTL


jasonLaster:
We show a tabs dropdown when there are too many source tabs to show. When the debugger is in RTL mode, the dropdown button is on the left and the dropdown is cropped by the left sidebar when it's opened.

We should have the dropdown open to the right of the button in RTL, which would keep it from being cropped.

#### patch

```diff


diff --git a/src/components/shared/Dropdown.css b/src/components/shared/Dropdown.css
index 5e0e481..ab314fb 100644
--- a/src/components/shared/Dropdown.css
+++ b/src/components/shared/Dropdown.css
@@ -4,6 +4,8 @@
   box-shadow: 0 4px 4px 0 var(--theme-search-overlays-semitransparent);
   max-height: 300px;
   position: absolute;
+
+  /* this probably needs to change*/
   offset-inline-end: 8px;
   top: 35px;
   width: 150px;
```

#### Screenshot

![screen shot 2017-02-24 at 6 58 18 pm](https://cloud.githubusercontent.com/assets/254562/23325434/430fba24-fac3-11e6-99aa-f12a45d6ac7c.png)

irfanhudda:
This was working with `postcss-bidirection` `2.0.3`.

zystvan:
I'll take this issue. @wldcordeiro

PR #2222, opened 2017-02-28
Fix Tab Dropdown in RTL Mode

zystvan:
Associated Issue: #2075 

### Summary of Changes

- Make the tab-overflow dropdown visible in RTL mode. 

### Test Plan

- [x] Looks right
- [x] Is the same position as original code gave before being passed through postcss-bidirection

### Screenshots/Videos

| LTR | RTL | 
|---|---|
| screenshot from 2017-02-28 16-24-55 | screenshot from 2017-02-28 16-20-23 |

### DIFF
--- a/src/components/shared/Dropdown.css
+++ b/src/components/shared/Dropdown.css
@@ -1,13 +1,18 @@
 .dropdown {
+  --width: 150px;
   background: var(--theme-body-background);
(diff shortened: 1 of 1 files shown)


jasonLaster:
looks good. I'll test in a bit

Issue #2221, opened 2017-02-28
[FunctionSearch] find es6 shorthand functions


jasonLaster:
We should search objects for es6 shorthand functions. Here is a test that will probably fail

```diff
diff --git a/src/utils/tests/parser.js b/src/utils/tests/parser.js
index 6dbddc6..f2a88d3 100644
--- a/src/utils/tests/parser.js
+++ b/src/utils/tests/parser.js
@@ -39,6 +39,10 @@ const TodoView = Backbone.View.extend({
   render: function () {
     return this;
   },
+
+  es6() {
+    console.log('es6 shorthand')
+  }
 });
 `);

@@ -65,7 +69,7 @@ describe("parser", () => {
       const fncs = getFunctions({ id: "proto" });
       const names = fncs.map(f => f.name);

-      expect(names).to.eql([ "foo", "bar", "initialize", "render"]);
+      expect(names).to.eql([ "foo", "bar", "initialize", "render", "es6"]);
     });
   });
```

wldcordeiro:
I landed this, we now have ES6 shorthand and ES6 class method searching added. 👍

PR #2236, opened 2017-03-01
Add parsing support for ES6 object method shorthand and class methods.

wldcordeiro:
Associated Issue: #2221 

### Summary of Changes

* more intelligently detect if we're dealing with a function and extract the names for ES6 shorthand and ES6 class methods

### Test Plan

- [x] Open debugger against a source with ES6 shorthand code or class code (The create react app example is cool for class code.)
- [x] Open the function search
- [x] Search for functions and see that the methods populate in search

### DIFF
--- a/src/components/Editor/SearchBar.js
+++ b/src/components/Editor/SearchBar.js
@@ -22,7 +22,7 @@ const SearchInput = createFactory(require("../shared/SearchInput"));
 const ResultList = createFactory(require("../shared/ResultList"));
 const ImPropTypes = require("react-immutable-proptypes");
 
-import type { FunctionDeclaration } from "../../utils/parser";
+import type { SymbolDeclaration } from "../../utils/parser";
 
 type ToggleFunctionSearchOpts = {
   toggle: boolean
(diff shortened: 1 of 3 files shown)


src/utils/parser.js
```diff
@@ -75,21 +75,30 @@ function getFunctionName(path) {
...
+
   if (parent.type == "VariableDeclarator") {
     return parent.id.name;
   }
 
   return "anonymous";
 }
 
+function isFunction(path) {
+  return t.isFunction(path) || t.isArrowFunctionExpression(path) ||
```
wldcordeiro: Could likely make it more exhaustive but I found that it worked for the different types of function declarations I'm aware of.

wldcordeiro:
This was fun! Playing with Babel/Babylon is super neat. I want to tackle adding variable and class searching next 😄

src/components/Editor/SearchBar.js
```diff
@@ -188,16 +188,9 @@ const SearchBar = React.createClass({
       return;
     }
 
-    const functionDeclarations = getFunctionDeclarations(
```
jasonLaster: nice

PR #2244, opened 2017-03-01
Fix the copy-assets script to support an absolute path to m-c, and su…

julienw:
…pport symlinking to an outoftree m-c clone

### DIFF
--- a/bin/copy-assets.js
+++ b/bin/copy-assets.js
@@ -19,50 +19,52 @@ function start() {
   console.log("start: copy assets")
   const projectPath = path.resolve(__dirname, "..")
   const mcModulePath =  "devtools/client/debugger/new";
-  const mcPath = args.mc ? args.mc : feature.getValue("firefox.mcPath");
+  let mcPath = args.mc ? args.mc : feature.getValue("firefox.mcPath");
+
+  // resolving against the project path in case it's relative. If it's absolute
+  // it will override whatever is in projectPath.
+  mcPath = path.resolve(projectPath, mcPath);
 
   copyFile(
     path.join(projectPath, "./assets/panel/debugger.properties"),
-    path.join(projectPath, mcPath, "/devtools/client/locales/en-US/debugger.properties"),
+    path.join(mcPath, "devtools/client/locales/en-US/debugger.properties"),
     {cwd: projectPath}
   );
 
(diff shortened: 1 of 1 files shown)


julienw:
I tested with an outoftree m-c clone because that was my issue. I'd appreciate if the reviewer could test an in-tree clone (I mean, the way recommended in the docs).

bin/copy-assets.js
```diff
@@ -19,50 +19,52 @@ function start() {
   console.log("start: copy assets")
   const projectPath = path.resolve(__dirname, "..")
   const mcModulePath =  "devtools/client/debugger/new";
-  const mcPath = args.mc ? args.mc : feature.getValue("firefox.mcPath");
+  let mcPath = args.mc ? args.mc : feature.getValue("firefox.mcPath");
+
+  // resolving against the project path in case it's relative. If it's absolute
+  // it will override whatever is in projectPath.
+  mcPath = path.resolve(projectPath, mcPath);
```
julienw: This is the magic.

bin/copy-assets.js
```diff
@@ -19,50 +19,52 @@ function start() {
...
-    path.join(projectPath, mcPath, "devtools/client/debugger/new/moz.build"),
+    path.join(mcPath, "devtools/client/debugger/new/moz.build"),
     {cwd: projectPath}
   );
 
+  const projectTestPath = path.join(projectPath, "src/test/mochitest");
+  const mcTestPath = path.join(mcPath, mcModulePath, "test/mochitest");
   if (shouldSymLink) {
-    symlinkTests({ projectPath, mcModulePath })
+    symlinkTests({ projectPath, mcTestPath, projectTestPath })
```
julienw: I've seen symlinkTests already support this, we just need to use it.

bin/copy-assets.js
juliandescottes: projectPath is always set as the working directory in the rest of the script, so I'm not sure if this is really needed.
I tried commenting out this line and the script still ran correctly for both in and out of tree mc clones. Am I missing something? (in any case it doesn't hurt to have it)

julienw: ah you're right; except maybe the outputPath at the end ?

PR #2245, opened 2017-03-01
Change the breakpoint used to switch between horizontal and vertical …

julienw:
…layouts, and tune the initialSize for the left panel

As discussed in https://bugzilla.mozilla.org/show_bug.cgi?id=1334984#c7

### DIFF
--- a/src/components/App.js
+++ b/src/components/App.js
@@ -9,7 +9,7 @@ const { getSources, getSelectedSource, getPaneCollapse } = require("../selectors
 const { KeyShortcuts } = require("devtools-sham-modules");
 const shortcuts = new KeyShortcuts({ window });
 
-const verticalLayoutBreakpoint = window.matchMedia("(min-width: 700px)");
+const verticalLayoutBreakpoint = window.matchMedia("(min-width: 800px)");
 
 require("./App.css");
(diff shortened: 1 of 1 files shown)


julienw:
I haven't completely finished testing yet.

julienw:
Now I think it's ready to be reviewed.

I updated both the trigger for the vertical layout (to 800px instead of 700).
I also changed the initial size for the source panel as I thought 300px was too big in the horizontal layout, and 150px too small in vertical layout. Also it seemed really weird to have different defaults. So I used 250px for both instead.

I think a better behavior would be:
* when moving from vertical to horizontal and vice versa, keep the same size for the source panel, so that it doesn't "jump".
* also remember the sizes between reopens.

But this will be good enough for now.

julienw:
I also don't really like how we use "vert". This defines how the _split bar_ is, and not how the _split box_ is. Yet we define a SplitBox here... so when vert is true, the split box is actually horizontal. This is really confusing.

Another confusing thing is that "vert" is true by default... I think we should revert the meaning of this attribute, or maybe if you're concerned about backward compatibility use a new property "direction" that would have 2 possible values "horizontal" and "vertical". Explicit is better than implicit :)

Issue #2103, opened 2017-02-23
[EditorSearch] cmd+f without a selection should refocus the search bar


jasonLaster:

aklt:
I;ll have a go at this in a few hours

jasonLaster:
@aklt curious to see what you think when you have a chance. It'd be neat if this *just* works now

aklt:
Hi Jason, I was expecting the previous search text to reappear in the search bar if I reopen it after having closed it previously as I think we talked about on slack.  That does not happen, but the search bar is refocused :)

jasonLaster:
Ahh. We should fix that :)

aklt:
Should I make an issue on it?

jasonLaster:
@aklt lets fix it in this issue

aklt:
This should be fixed now.

Issue #2228, opened 2017-03-01
[EditorSearch] polish overlays 


jasonLaster:
Lets double check to make sure that search overlays are properly cleared as you open and close the different searches.

PR #2247, opened 2017-03-02
[EditorSearch] cmd+f without a selection should refocus the search bar

aklt:
This is not pretty, but an ok solution.. I think, please review

The logic is a bit too spread out and I function deliberately ignored function search.

Luckily there is a refactor planned I understood from jlast

Associated Issue: #2103 

### Summary of Changes

* Make sure that last search is in the search bar even if it becomes unmarked in
  the editor.

* Selection in the editor becomes the search.  Not 100% sure this is the right approach.

### Test Plan

Simple
- [x] Open the debug editor on some source code containing the string "fun"
- [x] Press CTRL-f to open the search bar and type 3 letters: "fun"
- [x] Press escape and CTRL-f again: the search bar contains "fun"

Again
- [x] Delete last search, press ESC, CTRL-f to search for "fun" again
- [x] Press escape to close the search bar and click on the code, the highlighting of the search term disappears.
- [x] CTRL-f to search again.  The search bar contains "fun".

Typing
- [x] Edit text in the search bar and press ESC, CTRL-f again and the text is still there

Selection
- [x] CTRL-f to open the search bar, search for "fun", then select some text in the editor
- [x] Press escape to close search bar, CTRL-f and the text in the search bar is now the previously selected text.

### DIFF
--- a/src/components/Editor/SearchBar.js
+++ b/src/components/Editor/SearchBar.js
@@ -156,7 +156,6 @@ const SearchBar = React.createClass({
     const { editor: ed, query, modifiers } = this.props;
     if (ed) {
       const ctx = { ed, cm: ed.codeMirror };
-      this.props.updateQuery("");
       removeOverlay(ctx, query, modifiers);
     }
   },
@@ -194,7 +193,9 @@ const SearchBar = React.createClass({
     if (this.state.enabled && editor) {
       const selection = editor.codeMirror.getSelection();
       this.setSearchValue(selection);
-      this.doSearch(selection);
+      if (selection !== "") {
+        this.doSearch(selection);
+      }
       this.selectSearchInput();
     }
   },
--- a/src/components/Editor/index.js
+++ b/src/components/Editor/index.js
@@ -310,10 +310,9 @@ const Editor = React.createClass({
   },
 
   updateQuery(query) {
-    if (this.state.query == "" && query == "") {
+    if (this.state.query == query) {
       return;
     }
-
     this.setState({ query });
   },
 


wldcordeiro:
I think we can land a cleaner solution here by addressing this issue https://github.com/devtools-html/debugger.html/issues/2228

wldcordeiro:
@aklt I've opened #2252 when this lands you should be able to clean up this PR to not worry about selections.

aklt:
@wldcordeiro Great, thanks.  I have some work, but hopefully I can have a lool later this evening.

wldcordeiro:
I've been testing to see if <kbd>ctrl</kbd>/<kbd>cmd</kbd>+<kbd>F</kbd> without a selection would refocus the search bar and I'm seeing the behavior work as expected so we may not need this PR.

aklt:
@wldcordeiro Hmm, I do not see the behaviour I was expecting.  If I open the debugger in
a file, press CTRL-f, search for something and then press ESC, I was expecting to see the term I searched for still highlighted in the search bar.

On my machine if I press CTRL-f with text selected it is copied into the search bar as expected, but as above it is gone if the bar is closed and CTRL-f is pressed again.

aklt:
BTW, how do you make those neat looking buttons? :-)

aklt:
Finally updated my changes and this should be fixed now.

wldcordeiro:
>BTW, how do you make those neat looking keyboard buttons? :-)

@aklt You can get those nice looking buttons with `<kbd>ctrl</kbd>` 😄

Issue #2249, opened 2017-03-02
Variable/Class Search


wldcordeiro:
In the same vein as Function searching we want to be able to find variables (`var`, `let`, `const`) and class declarations in the document.

- [x] Add parsing utilities for finding the variables and classes
- [x] Add UI for searching variables and classes

wldcordeiro:
This has landed, now there's just polish 😄

PR #2250, opened 2017-03-02
Rework function parser helpers into symbol parser helpers that parse functions, variables, and classes

wldcordeiro:
Associated Issue: #2249 

### Summary of Changes

* Rework `getFunctions` into `getSymbols` function that traverses an AST and extracts variable, function and class declaration names and locations

### Test Plan

- [x] Run the unit tests and see that they're passing
- [x] Test that function search is still working as expected

### DIFF
--- a/src/utils/parser.js
+++ b/src/utils/parser.js
@@ -111,6 +111,13 @@ function getFunctions(source: SourceText): Array<SymbolDeclaration> {
 }
 
 function getVariableNames(path) {
+  if (t.isObjectProperty(path) && !isFunction(path.node.value)) {
+    return [{
+      name: path.node.key.name,
+      location: path.node.loc
+    }];
+  }
+
   if (!path.node.declarations) {
     return path.node.params
     .map(dec => ({
@@ -118,6 +125,7 @@ function getVariableNames(path) {
       location: dec.loc
     }));
   }
+
   return path.node.declarations
     .map(dec => ({
       name: dec.id.name,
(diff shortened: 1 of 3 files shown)


src/utils/tests/parser.js
```diff
@@ -57,8 +58,15 @@ const classTest = formatCode(`
   }
 `);
 
+const varTest = formatCode(`
+  var foo = 1;
+  let bar = 2;
+  const baz = 3;
+  const a = 4, b = 5;
+`);
```
jasonLaster: perhaps we should find params and properties as well:

`function(bar) {}`
`x = {a: 2}`

wldcordeiro: Agreed! Added parameter parsing. Now just properties to go 😄

wldcordeiro:
@jasonLaster if you have any ideas for additional tests it'd be much appreciated. 😄 Also since our tests mostly check for names we don't handle multiple symbols of the same name well in the tests. The code returns objects with names and locations though so that shouldn't affect our function/variable search utils themselves.

wldcordeiro:
Adding class parsing was easy so I did that too. 😃

src/utils/parser.js
```diff
@@ -91,40 +111,90 @@ function isFunction(path) {
...
     value: dec.name,
     location: dec.location
   }));
+}
+
+function getSymbolDeclarations(
+  sourceText: SourceText): FormattedSymbolDeclarations {
+  if (symbolDeclarations.has(sourceText.id)) {
+    const symbols = symbolDeclarations.get(sourceText.id);
+    if (symbols) {
```
wldcordeiro: Flow complained without this guard here. 😆

src/utils/parser.js
```diff
@@ -91,40 +111,90 @@ function isFunction(path) {
...
+  return declarations.map(dec => ({
     id: `${dec.name}:${dec.location.start.line}`,
     title: dec.name,
     subtitle: `:${dec.location.start.line}`,
     value: dec.name,
     location: dec.location
   }));
+}
+
+function getSymbolDeclarations(
```
wldcordeiro: Maybe kill this function in place of doing this work within `getSymbols`? I tried keeping it as similar to the previous API.

PR #2251, opened 2017-03-02
Variable and Class Search UI

wldcordeiro:
Associated Issue: #2249 

### Summary of Changes

* Implement the UI for variable and class searching

### Test Plan

- [x] Open debugger and document
- [x] Toggle symbol search <kbd>ctrl</kbd>/<kbd>cmd</kbd>+<kbd>shift</kbd>+<kbd>O</kbd>
- [x] Select variable or class button
- [x] Type to search and see results appear

### Screenshots/Videos

screenshot_20170302_001645
screenshot_20170302_001722

There are some kinks to work out here in terms of UX and some bugginess and I'll want to rebase this work on master after #2250 lands but I was too excited to not try it out 😆


src/components/Editor/SearchBar.js
```diff
@@ -213,28 +216,28 @@ const SearchBar = React.createClass({
...
 
     if (query == "" || !sourceText) {
       return;
     }
 
-    const functionDeclarations = getFunctionDeclarations(
-      sourceText.toJS()
-    );
+    const symbolDeclarations = getSymbols(
+      sourceText.toJS())[selectedSymbolType];
```
wldcordeiro: Flow complains about this line 🤷‍♂️

configs/ci.json
```diff
@@ -1,15 +1,31 @@
 {
   "features": {
-    "sourceMaps": {
```
wldcordeiro: These flags no longer exist? So I updated the list to be all the current ones.

src/components/Editor/SearchBar.js
```diff
@@ -50,9 +51,9 @@ const SearchBar = React.createClass({
   getInitialState() {
     return {
       enabled: false,
-      functionSearchEnabled: false,
-      functionDeclarations: [],
-      functionSearchResults: [],
+      symbolSearchEnabled: false,
+      selectedSymbolType: "functions",
```
wldcordeiro: Default to functions? I'm not sure if that's  what we want but seemed like a decent approach.

src/components/Editor/SearchBar.js
```diff
@@ -136,7 +137,11 @@ const SearchBar = React.createClass({
 
     if (this.state.enabled && ed) {
       this.clearSearch();
-      this.setState({ enabled: false, functionSearchEnabled: false });
+      this.setState({
+        enabled: false,
+        symbolSearchEnabled: false,
+        selectedSymbolType: "functions"
```
wldcordeiro: Should closing search reset the selected symbol?

src/components/Editor/SearchBar.js
```diff
@@ -151,9 +156,10 @@ const SearchBar = React.createClass({
       this.setState({ enabled: true });
     }
 
-    if (this.state.functionSearchEnabled) {
+    if (this.state.symbolSearchEnabled) {
       this.clearSearch();
-      this.setState({ functionSearchEnabled: false });
+      this.setState({
+        symbolSearchEnabled: false, selectedSymbolType: "functions" });
```
wldcordeiro: Same story, should we always reset to functions or leave the last selected symbol type like we do with previous search modifiers?

src/strings.json
```diff
@@ -40,8 +40,8 @@
   "sourceSearch.search.again.key": "G",
   "sourceSearch.resultsSummary1": "%d results",
   "sourceSearch.noResults": "No files matching %S found",
-  "functionSearch.search.placeholder": "Search Functions...",
-  "functionSearch.search.key": "o",
+  "symbolSearch.search.placeholder": "Search %S...",
```
jasonLaster: we should translate all three of the options

src/components/Editor/SearchBar.js
```diff
@@ -396,8 +406,9 @@ const SearchBar = React.createClass({
   },
 
   buildPlaceHolder() {
-    if (this.state.functionSearchEnabled) {
-      return L10N.getStr("functionSearch.search.placeholder");
+    if (this.state.symbolSearchEnabled) {
+      return L10N.getFormatStr(
```
jasonLaster: hmm, we should translate class/function/...

src/components/Editor/SearchBar.js
```diff
@@ -437,27 +448,41 @@ const SearchBar = React.createClass({
...
       ),
-      dom.button({
-        className: classnames("search-type-btn", {
-          active: this.state.functionSearchEnabled
-        }),
-        onClick: e => this.toggleFunctionSearch(e, { toggle: true })
-      }, "functions")
+      searchTypeBtn("functions"),
+      searchTypeBtn("variables"),
+      searchTypeBtn("classes")
```
jasonLaster: hmm, perhaps only show this when there are classes?

src/components/Editor/SearchBar.js
```diff
@@ -437,27 +448,41 @@ const SearchBar = React.createClass({
...
+    const { toggleSymbolSearch } = this;
+    const { symbolSearchEnabled, selectedSymbolType } = this.state;
+
+    function searchTypeBtn(searchType) {
+      return dom.button({
+        className: classnames("search-type-btn", {
+          active: symbolSearchEnabled && selectedSymbolType == searchType
+        }),
+        onClick: e => {
+          if (selectedSymbolType == searchType) {
```
jasonLaster: we should re-trigger a search when there's a query. 

STR:
1. type `t`
2. select function
3. select variables should re-search with `t` or clear the query

src/components/Editor/SearchBar.js
wldcordeiro: I like that idea but I think we'd need to modify when we call `getSymbols` so that the symbol declarations are available after documents load (getting the declarations is currently done when `doSearch` is called).

src/components/Editor/SearchBar.js
```diff
@@ -213,26 +221,28 @@ const SearchBar = React.createClass({
...
       updateSearchResults
     } = this.props;
+    const { selectedSymbolType } = this.state;
 
     if (query == "" || !sourceText) {
       return;
     }
 
-    const functionDeclarations = getSymbols(sourceText.toJS()).functions;
+    const symbolDeclarations = getSymbols(
```
wldcordeiro: This function `updateSymbolSearchResults` is called whenever `doSearch` is called so we're parsing the document every time we search. I think we should be able to move the `symbolDeclarations` either to state or props so they're parsed out only once per document.

jasonLaster:
Hmm. It should be memoized

wldcordeiro:
@jasonLaster Oh yeah we do memoize! https://github.com/devtools-html/debugger.html/commit/890830521db64b7457afbe9ea8528525f1049563#diff-f948fa142a38b4398066c52da39206deR44

jasonLaster:
This looks good at first glance. I'll review it more closely on monday.

Also, it'd be nice to get @clarkbw's thoughts on class search as well. This wasn't part of the initial specification. Now that I think about it, I'm a bit concerned that we could be using valuable horizontal real-estate that we might want in the future. I suppose one option would be to include classes in the variable search, which is similar. 

Also, it would be nice to explore the variable search UI a bit more as we might not want to include local variables... but just properties of the module scope and maybe... one layer below that like object properties. Worth thinking about and i could be persuaded in any direction. Also, not blocking.

wldcordeiro:
Awesome! I'd love to hear feedback from more people. An idea I had is making the button a dropdown/context menu to save on horizontal space.

jasonLaster:
Nice. Mind drawing that up?

src/components/Editor/SearchBar.js
```diff
@@ -396,8 +406,12 @@ const SearchBar = React.createClass({
...
 
   buildPlaceHolder() {
-    if (this.state.functionSearchEnabled) {
-      return L10N.getStr("functionSearch.search.placeholder");
+    const { symbolSearchEnabled, selectedSymbolType } = this.state;
+    if (symbolSearchEnabled) {
+      return L10N.getFormatStr(
+        "symbolSearch.search.placeholder",
+        L10N.getStr(`symbolSearch.search.${selectedSymbolType}Placeholder`)
+      );
```
jasonLaster: @flodilo is it okay to interpolate strings like this:


*"Search for %S"*, where `%S` could be *function*, *variables*, ...

src/strings.json
flodolo: Also, use … and not ...

src/components/Editor/SearchBar.js
flodolo: It should be safe-ish, but given it's 3 strings vs 1, I would prefer to have three explicit strings for each case and be 100% safe.

clarkbw:
Overall looks cool. 

> I suppose one option would be to include classes in the variable search, which is similar.

This was my first thought.  That classes were just like variables (of course still different), however they would need some distinction in the results list so you know when seeing it that its a class declaration vs. an instance or something.

wldcordeiro:
> This was my first thought. That classes were just like variables (of course still different), however they would need some distinction in the results list so you know when seeing it that its a class declaration vs. an instance or something.

I've seen some IDEs use an icon like a `c` in a circle for classes, `i` for instance, `f` for function, etc. We could maybe use that in the result list? Like `c - name        line no`

jasonLaster:
Yep. Safari has icons. I find theirs noisy, but we could do something nice

src/components/Editor/SearchBar.js
wldcordeiro: I'm going to open this as a follow-up issue for polishing the search.

wldcordeiro:
@jasonLaster I've updated the PR with the L10N changes and rebased onto the most recent master. How about we open an issue for layout polish (whether to use icons, dropdown, etc)?

jasonLaster:
@wldcordeiro sounds good. approved and will merge when it's green.

Issue #2220, opened 2017-02-28
[EditorPreview] Preview variables from outer scopes


jasonLaster:
It would be nice to be able to hover on variables that come from outer scopes and show them in the preview.

We can do this by getting the scope chain out of pauseData and getting all of the binding variables.

jasonLaster:
here's a [branch](https://github.com/devtools-html/debugger.html/compare/master...jasonLaster:vars-in-scope?expand=1) I started to help get started. 

There's an interesting question where we call `variablesInScope`, perhaps it's a selector in the pause reducer...

julienw:
Clearing because I'll be in PTO. Feel free to merge or reuse my patch :)

PR #2271, opened 2017-03-03
Preview variables from outer scopes

julienw:
Fix issue #2220

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -18,12 +18,9 @@
     "devtools": true,
     "dump": true,
     "EventEmitter": true,
-    "exports": true,
     "isWorker": true,
     "loader": true,
-    "module": true,
     "reportError": true,
-    "require": true,
     "Services": true,
     "Task": true,
     "XPCNativeWrapper": true,
@@ -31,14 +28,14 @@
     "_Iterator": true,
     "__dirname": true,
     "process": true,
-    "expect": true,
     "global": true,
     "L10N": true
   },
   "env": {
       "es6": true,
       "browser": true,
-      "mocha": true
+      "commonjs": true,
+      "jest": true
   },
   "rules": {
     // These are the rules that have been configured so far to match the
--- a/src/components/Editor/index.js
+++ b/src/components/Editor/index.js
@@ -9,7 +9,6 @@ const { connect } = require("react-redux");
 const classnames = require("classnames");
 
 const { getMode } = require("../../utils/source");
-const { getExpression } = require("../../utils/parser");
 
 const Footer = createFactory(require("./Footer"));
 const SearchBar = createFactory(require("./SearchBar"));
@@ -23,7 +22,7 @@ const {
   getSourceText, getBreakpointsForSource,
   getSelectedLocation, getSelectedFrame,
   getSelectedSource, getHitCountForSource,
-  getCoverageEnabled, getLoadedObjects
+  getCoverageEnabled, getLoadedObjects, getPause
 } = require("../../selectors");
 const { makeLocationId } = require("../../reducers/breakpoints");
 const actions = require("../../actions");
@@ -43,19 +42,13 @@ const {
   breakpointAtLine,
   getTextForLine,
   getCursorLine,
-  getTokenLocation,
+  getSelectedExpression,
   resizeBreakpointGutter,
   traverseResults
 } = require("../../utils/editor");
 const { isFirefox } = require("devtools-config");
 
 require("./Editor.css");
-
-async function getExpressionFromToken(
-  cm: any, sourceText, token: HTMLElement) {
-  const loc = getTokenLocation(token, cm);
-  return await getExpression(sourceText.toJS(), token.innerText || "", loc);
-}
 
 const Editor = React.createClass({
   propTypes: {
@@ -263,7 +256,7 @@ const Editor = React.createClass({
   },
 
   onScroll(e) {
-    return this.setState({ selectedToken: null });
+    return this.setState({ selectedToken: null, selectedExpression: null });
   },
 
   onMouseUp(e, ctx, modifiers) {
@@ -277,11 +270,9 @@ const Editor = React.createClass({
   },
 
   async previewSelectedToken(e, modifiers) {
-    const { selectedFrame, sourceText } = this.props;
+    const { selectedFrame, pauseData, sourceText } = this.props;
     const { selectedToken } = this.state;
-    const cm = this.editor.codeMirror;
     const token = e.target;
-    const tokenText = token.innerText;
 
     if (!selectedFrame || !sourceText || !isEnabled("editorPreview")) {
       return;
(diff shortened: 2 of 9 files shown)


julienw:
I'm not completely satisfied by this:

* I can't get the value from global variables
* variables shadowed: I'd like that when hovering the shadowed variable, I get the value of that variable instead of the value of the variable from the inner scope -- note that the old debugger has the same issue here.
* I feel like we should reuse utils/object-inspector somehow.

julienw:
But maybe this is good enough for now :)

julienw:
Do you have an idea of how to test this automatically ?

here is the example page I use to test it:
```html
<!doctype html>
<html>
  <head>
    <meta charset='utf-8'>
    <meta name='viewport' content='initial-scale=1'>
  </head>
  <body>
    <button type='button'>Click me!</button>
    <script type="text/javascript" charset="utf-8">
      var a = 'a';
      var b = 'b';

(function() {
  const c = 'cc';
  let a = 'aa';

  function func(c, d) {
    var e = 'eee';
    {
      const f = 'fff';
      debugger;
    }
  }

  document.querySelector('button').onclick = () => func('ccc', 'ddd');
})();
    </script>
  </body>
</html>
```

src/components/Editor/index.js
```diff
@@ -276,30 +277,50 @@ const Editor = React.createClass({
...
+    if (tokenText == "this" && ("this" in selectedFrame)) {
+      return { value: selectedFrame.this };
+    }
+
+    const variables = getVisibleVariablesFromScope(pauseData, selectedFrame);
+    if (variables.has(tokenText)) {
+      return variables.get(tokenText);
+    }
+
+    const expression = await getExpressionFromToken(cm, sourceText, token);
```
julienw: from the cheapest to the most costly

src/components/Editor/index.js
```diff
@@ -276,30 +277,50 @@ const Editor = React.createClass({
     this.previewSelectedToken(e, modifiers);
   },
 
+  async getSelectedExpression(token) {
+    const { selectedFrame, pauseData, sourceText } = this.props;
+    const { selectedToken } = this.state;
+    const cm = this.editor.codeMirror;
+
+    const tokenText = token.textContent;
```
julienw: textContent is more efficient -- if we don't need innerText.

julienw:
@jasonLaster I think it's fairly ready now

src/components/Editor/index.js
```diff
@@ -276,30 +277,50 @@ const Editor = React.createClass({
     this.previewSelectedToken(e, modifiers);
   },
 
+  async getSelectedExpression(token) {
```
jasonLaster: this looks like a great util that could move to `utils/editor`... and then be tested:

```js
async getSelectedExpression(token, cm, {
  selectedFrame, pauseData, sourceText, selectedToken
}) {
...
}
```

we could also move `getExpressionFromToken` as well...

.eslintrc
```diff
@@ -18,27 +18,24 @@
...
     "Services": true,
     "Task": true,
     "XPCNativeWrapper": true,
     "XPCOMUtils": true,
     "_Iterator": true,
     "__dirname": true,
     "process": true,
-    "expect": true,
     "global": true,
     "L10N": true
```
julienw: these globals are handled by the environments I added below

src/utils/parser/index.js
```diff
@@ -5,26 +5,29 @@ const { getValue } = require("devtools-config");
...
+  return Promise.resolve(worker);
+}
+
+const getSymbols = workerTask(getOrCreateWorker, "getSymbols");
+const getVariablesInScope =
+  workerTask(getOrCreateWorker, "getVariablesInScope");
+const getExpression = workerTask(getOrCreateWorker, "getExpression");
 
 module.exports = {
   getSymbols,
```
julienw: These changes are needed to be able to mock it using jest; the problem is that it was starting the worker unconditionally. Now I made it lazy.

I decided to make `getOrCreateWorker` return a Promise because `workerTask` is already promise-based and we might want to do some async initialization for some workers in the future (like waiting for a "ready" message or something)).

src/utils/tests/expression.js
```diff
@@ -0,0 +1,67 @@
+import fromJS from "../fromJS";
+
+jest.mock("../scopes");
+jest.mock("../parser");
```
julienw: from my tests it needs to be here so that it's properly mocked when expression.js got imported.

I'm not sure how this interacts with other files. Maybe Jest is smart and run all files separately and everything will be fine.

src/utils/tests/scopes.js
```diff
@@ -103,4 +106,251 @@ describe("scopes", () => {
...
+     * We use a IIFE because we don't support global variables yet, so this is
+     * necessary to get variables in the outer scope. This can be removed once
+     * getVisibleVariablesFromScope supports global variables.
+     *
+     * `b` is shadowed by a function parameter, `c` is shadowed by a local
+     * variable, but `a` is not shadowed. `this` is a string containing te value
+     * `"this"`.
+     */
+
+    const frame = {
```
julienw: This is not so simple but I found that it's better to have something closer to the reality.

src/utils/tests/scopes.js
```diff
@@ -103,4 +106,251 @@ describe("scopes", () => {
...
+      });
+
+      global.L10N = { getStr: () => "" };
+    });
+
+    afterEach(function() {
+      delete global.L10N;
+    });
+
+    it("Returns variables from the outer scope", function() {
```
julienw: Ideally I'd like a test to return variable from a frame higher in the call stack, but I've been lazy as this would need a completely different `pauseInfo`.

Which made me wonder why we need the frame after all because it seems everything we need is actually in `pauseInfo` ?

src/utils/worker.js
```diff
@@ -11,27 +11,29 @@ let msgId = 1;
  * @memberof utils/utils
  * @static
  */
-function workerTask(worker: any, method: string) {
+function workerTask(retrieveWorker: Function, method: string) {
```
julienw: I wasn't sure how Flow would type this function that takes nothing and returns a Promise.

src/utils/editor/expression.js
```diff
@@ -0,0 +1,45 @@
...
+  if (tokenText == "this" && ("this" in selectedFrame)) {
+    return { value: selectedFrame.this };
+  }
+
+  const variables = getVisibleVariablesFromScope(pauseData, selectedFrame);
+  if (variables.has(tokenText)) {
+    return variables.get(tokenText);
+  }
+
+  const expression = await getExpressionFromToken(cm, token, sourceText);
```
jasonLaster: lets refactor this so it's easier to test

jasonLaster: https://gist.github.com/jasonLaster/2679e0968720a36f71b75882e80fd2ef

PR #2288, opened 2017-03-05
(WIP) Tweak the search bar appearance

jasonLaster:

### Summary of Changes

* clean up the colors
* adjust the heights to be more consistent

**Next Steps**
* Fix Source search height (keep it 40px)
* dark theme
  * Make the sidebar headers less prominent
  * remove the zebra call stack frames
  * fix very blue selection

#### New Look

screen shot 2017-03-05 at 6 00 04 pm

screen shot 2017-03-05 at 6 02 18 pm

#### Old Look

screen shot 2017-03-05 at 6 02 58 pm

----

#### Paused UI

screen shot 2017-03-06 at 12 42 11 pm
screen shot 2017-03-06 at 12 41 57 pm

### DIFF
--- a/docs/local-development.md
+++ b/docs/local-development.md
@@ -321,6 +321,12 @@ Often, it is more practicle to create a new variable.
 
 It's helpful to share the changes as a themes [table][pr-table] when you're done.
 
+#### Checking Contrast
+
+It's important to make sure that the contrast ratio is sufficient.
+
+You can check the background / text color contrast ratio with this [tool][contrast-ratio-tool].
+
 ### Configs
 
 The Debugger uses configs for settings like `theme`, `hotReloading`, and feature flags.
@@ -504,3 +510,4 @@ your questions on [slack][slack].
 
 [mochitest]: ./mochitests.md
 [mocha]: ./integration-tests.md
+[contrast-ratio-tool]: http://leaverou.github.io/contrast-ratio/#rgb%28204%2C%20209%2C%20213%29-on-rgb%28252%2C%20252%2C%20252%29
--- a/src/components/Editor/Footer.css
+++ b/src/components/Editor/Footer.css
@@ -1,6 +1,6 @@
 
 .source-footer {
-  background: var(--theme-body-background);
+  background: var(--theme-toolbar-background);
   border-top: 1px solid var(--theme-splitter-color);
   position: absolute;
   display: flex;
@@ -11,7 +11,7 @@
   z-index: 100;
   -moz-user-select: none;
   user-select: none;
-  height: 30px;
+  height: 27px;
   box-sizing: border-box;
 }
 
--- a/src/components/Editor/SearchBar.css
+++ b/src/components/Editor/SearchBar.css
@@ -1,7 +1,16 @@
+
 .search-bar {
   display: flex;
   flex-direction: column;
   max-height: 50%;
+}
+
+.search-bar .search-field {
+  padding-left: 7px;
+}
+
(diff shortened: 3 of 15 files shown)


jasonLaster:
@nchevobbe pointed out that the search options bar is not visually differentiated from the search result items.

nchevobbe:
Not specific to function search, but I think we could polish this up a bit by aligning the buttons

screenshot at mars 06 18-54-31

jasonLaster:
Good find @nchevobbe. Damn, i hate that i'm off by a bit on the close button after just tweaking it :)

jasonLaster:
Just styled the source result list UI in the inspector. Thoughts?

screen shot 2017-03-06 at 12 59 37 pm

nchevobbe:
erf, I think the first one would make more sense, because a lot of people use the debugger docked to the bottom, and vertical space is then limited.
That's the reason I'm not too much into the "card" look for the result.
Is there a rationale displaying the results for file search differently than for function search ?

jasonLaster:
@nchevobbe that's a good point. I think helen was using editor project level search as an inspiration. I agree that it's not as space efficient. 

I'd like to find a happy medium where it's not to dense when we have space, but also more space efficient when we're space challenged.

nchevobbe:
For the design of the function search (but really, the search as a whole), it may be helpful to write down what we will end up to have
- We can __search in file__ for __text__
- We can __search in file__ for __function__
- Hopefully, we'll be able to __search in all files__ for __text__
- maybe be able to __search in all files__ for __function__
- maybe be able to __search in file__ for __variable__
- and then why not __search in all files__ for __variable__
- and each of these search could be constrained by : Regex,  case, whole word, ...

So we have several ideas:
- Where to search (file, all files)
- What to search for (text, function, variables, ..)
- How to search (case (in)sensitive, whole word, Regex)

The result list you did for the function results could be also used for the search in all files.

One thing we could have is all this in a single bar, and the ability to switch from "context" with a button that would open the choices.

```
Search *in file ▾* : a                        for *text ▾* | aA | .*
```
or
```
Search *in al files ▾* : a                   for *function ▾* | aA | .*
```

These context could be automatically selected depending off the keyboard shortcut you use.
I don't know if it would be better than a second toolbar, it was a thought.
I am looking at how to improve the current version too.

nchevobbe:
> I'd like to find a happy medium where it's not to dense when we have space, but also more space efficient when we're space challenged.

Mhh, so we could either use media-queries, or see how we could work with viewport based units (`vh`in this case ? )

jasonLaster:
Yeah - i think media queries like vertical layout would be slick.

* project search will come along soon. That's a Q2 goal. 
* jumping between file/project search is important. (unsolved problem)
* variable search is PRd now (close)
* modifiers are not used for function search as we're using fuzzy aldrin which is way better

wldcordeiro:
@nchevobbe your little sketch resembles what I was thinking of for adding variable and class search. 👍

jasonLaster:
This PR has become pretty big and I hope we can wrap it up and then do some smaller style tweaks:

1. Update Search bars so there's a big and small input and result list
2. Update some details like borders and colors
3. improve contrast and consistence

#### Editor Search
screen shot 2017-03-07 at 5 15 15 pm
screen shot 2017-03-07 at 5 15 05 pm

#### Source Search

screen shot 2017-03-07 at 5 15 40 pm
screen shot 2017-03-07 at 5 15 19 pm

wldcordeiro:
Flow and css lint are complaining currently.

PR #2292, opened 2017-03-06
version debugger

jasonLaster:
### Summary of Changes

* bump debugger to version 0.1.0 
* add a module manifest for consistent bundles so our modules don't bounce around in the bundle. Thanks @TheLarkInn for the tip!

@juliandescottes


jasonLaster:
yeah, true. we could always push to it before it lands

PR #2296, opened 2017-03-07
Add flow types for remaining files

jasonLaster:
### Summary of Changes

* event listeners
* navigation
* still needs some work, but it's a start

### Test Plan

flow


jasonLaster:
this was more than i expected :)

Issue #2219, opened 2017-02-28
[EditorPreview] Preview Member Expressions


jasonLaster:
It'd be nice to be able to hover on `b` of `obj.a.b` and preview the expression `obj.a.b`.

once we have the expression, we need to be able to be able to call `client.evaluate(expression, frameId)` and show the response in the popup.

https://astexplorer.net/#/gist/a3801b36d67b0d11b18898649e23eb04/e352a873fdf5b1736daf77dca9e56d70d564efa4

bomsy:
i'll poke at this

jasonLaster:
Here's a checklist for progress:
- [ ] parse member expressions #2297
- [ ] evaluate expressions
- [ ] parse member expressions in `getSymbols` so it's fast

PR #2297, opened 2017-03-07
Wip member expressions 2

bomsy:
Associated Issue: #2219

WIP !!

### Summary of Changes

* Add `getExpression` function to the parser
* Highlight member expressions and show full expression in the popup

### Todo
* Evaluate expression and show in popup
* Add test for `getExpression`

### Test Plan
- [x] Mouseover property of object

### Screenshots
screen shot 2017-03-07 at 3 00 11 pm

### DIFF
--- a/src/components/Editor/index.js
+++ b/src/components/Editor/index.js
@@ -567,7 +567,7 @@ const Editor = React.createClass({
 
   renderPreview() {
     const { selectedToken, previewExpression } = this.state;
-    const { selectedFrame, sourceText } = this.props;
+    const { selectedFrame } = this.props;
 
     if (!selectedToken || !selectedFrame || !isEnabled("editorPreview")) {
       return;
--- a/src/components/Editor/index.js
+++ b/src/components/Editor/index.js
@@ -51,13 +51,19 @@ const { isFirefox } = require("devtools-config");
 
 require("./Editor.css");
 
+function getExpresionFromToken(
+  cm: any, sourceText, token: HTMLElement) {
+  const loc = getTokenLocation(token, cm);
+  return getExpression(sourceText.toJS(), token.innerText || "", loc);
+}
+
 const Editor = React.createClass({
   propTypes: {
     breakpoints: ImPropTypes.map.isRequired,
     hitCount: PropTypes.object,
     selectedLocation: PropTypes.object.isRequired,
     selectedSource: ImPropTypes.map,
-    sourceText: PropTypes.object,
+    sourceText: ImPropTypes.map,
     addBreakpoint: PropTypes.func.isRequired,
     disableBreakpoint: PropTypes.func.isRequired,
     enableBreakpoint: PropTypes.func.isRequired,
@@ -87,7 +93,6 @@ const Editor = React.createClass({
         count: 0
       },
       selectedToken: null,
-      previewExpression: null,
       searchModifiers: {
         caseSensitive: true,
         wholeWord: false,
@@ -151,7 +156,7 @@ const Editor = React.createClass({
 
     codeMirrorWrapper
       .addEventListener("mouseover", e => this.onMouseOver(
-        e, ctx, searchModifiers
+        e, searchModifiers
       ));
 
     if (!isFirefox()) {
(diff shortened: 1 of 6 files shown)


src/utils/parser.js
```diff
@@ -181,6 +181,38 @@ function getSymbols(source: SourceText): SymbolDeclarations {
...
+        expression = {
+          value: expr.join("."),
+          location: node.loc
+        };
+      }
+    }
+  });
+
+  return expression;
+}
```
jasonLaster: this is nice, but needs some unit tests in `utils/tests/parser`

src/utils/parser.js
```diff
@@ -181,6 +181,38 @@ function getSymbols(source: SourceText): SymbolDeclarations {
...
+  }
+
+  traverse(ast, {
+    enter(path) {
+      const node = path.node;
+      if (node.type === "MemberExpression" && node.property.name === token
+        && nodeContainsLocation({ node, location })) {
+        const expr = [];
+        expr.unshift(node.property.name);
+        getMemberExpression(node.object, expr);
```
jasonLaster: small nit, it's not to use returns and pass local state down

```js
const expr = getMemberExpression(node.object)

...

function _getMemberExpression(node, expr) {
  if (node.type === "MemberExpression") {
    expr = [node.property.name].concat(expr)
    return _ getMemberExpression(node.object, expr);
  } 

   return [node.name].concat(expr)
}

function getMemberExpression(object) {
  return _getMemberExpression(node,  []);
}
```

src/components/Editor/index.js
```diff
@@ -77,6 +78,7 @@ const Editor = React.createClass({
         count: 0
       },
       selectedToken: null,
+      selectedTokenLocation: null,
```
jasonLaster: heh, perhaps we use and object here and use `preview` over selected

```
{
  previewExpression: {
     text: null,
     location: null
  }
}
```

src/components/Editor/index.js
```diff
@@ -272,15 +274,25 @@ const Editor = React.createClass({
       selectedToken.classList.remove("selected-token");
     }
 
+    const lineOffset = 1;
+    const { left, top } = token.getBoundingClientRect();
+    const { line, ch } = ctx.cm.coordsChar({ left, top });
```
jasonLaster: indirect, but i like it

src/components/Editor/index.js
```diff
@@ -272,15 +274,25 @@ const Editor = React.createClass({
...
+    const loc = { line: line + lineOffset, column: ch };
```
jasonLaster: lets move this to a helper in `utils/editor/index`

perhaps: `getLocationFromToken`

src/utils/parser.js
bomsy: yes!

src/utils/parser.js
bomsy: cool. This passes a new state each time. nice! learnt something new! thanks

src/utils/parser.js
```diff
@@ -144,6 +145,22 @@ function isVariable(path) {
     (t.isObjectProperty(path) && !isFunction(path.node.value));
 }
 
+function getMemberExpression(root) {
+  function _getMemberExpression(node, expr) {
+    if (node.type === "MemberExpression") {
```
wldcordeiro: Rather than looking at the type directly with string comparison we could use `babel-type` helpers https://github.com/babel/babel/tree/master/packages/babel-types#memberexpression There are additional ones that match JSX member expressions and more so maybe a simple helper that does multiple checks like my `isVariable` or `isFunction` helpers.

src/utils/parser.js
```diff
@@ -144,6 +145,22 @@ function isVariable(path) {
...
+      expr = [node.property.name].concat(expr);
+      return _getMemberExpression(node.object, expr);
+    }
+
+    if (node.type === "ThisExpression") {
```
wldcordeiro: Same thing, https://github.com/babel/babel/tree/master/packages/babel-types#thisexpression

src/utils/parser.js
```diff
@@ -181,6 +198,31 @@ function getSymbols(source: SourceText): SymbolDeclarations {
...
+  const ast = getAst(source);
+
+  if (isEmpty(ast)) {
+    return;
+  }
+
+  traverse(ast, {
+    enter(path) {
+      const node = path.node;
+      if (node.type === "MemberExpression" && node.property.name === token
```
wldcordeiro: Utilizing babel-types across the parser would be a good idea. I think there's a place in the function parsing helpers where I could do this too.

src/utils/parser.js
```diff
@@ -181,6 +198,31 @@ function getSymbols(source: SourceText): SymbolDeclarations {
...
 
+function getExpression(source: SourceText, token: string, location: Location) {
+  let expression = null;
+  const ast = getAst(source);
+
+  if (isEmpty(ast)) {
+    return;
+  }
+
+  traverse(ast, {
```
wldcordeiro: Maybe it'd be a bit more efficient if we had a single function that did traversal? I know `getSymbols` does it and so does `getPathClosestToLocation` we could limit our traversals of documents to single places and return a single `Map` with all the relevant goodies.

src/components/Editor/index.js
```diff
@@ -42,20 +43,27 @@ const {
...
   traverseResults
 } = require("../../utils/editor");
 const { isFirefox } = require("devtools-config");
 
 require("./Editor.css");
 
+function getExpresionFromToken(
+  cm: any, sourceText: SourceText, token: HTMLElement) {
+  const loc = getTokenLocation(token, cm);
+  return getExpression(sourceText.toJS(), token.innerText || "", loc);
```
wldcordeiro: Flow doesn't like the use of `toJS` on sourceText here.`

jasonLaster:
@bomsy i pushed a shortcut for the flow error.

Issue #2266, opened 2017-03-03
[FunctionSearch] cmd+shift+o navigates to settings in the panel


jasonLaster:
`cmd+shift+o` which is one way of opening the function search is overriden by the settings pane in the panel.

We could probably fix this with a `stopPropogation` call

```diff
diff --git a/src/components/Editor/SearchBar.js b/src/components/Editor/SearchBar.js
index a4d1b89..8d10ebe 100644
--- a/src/components/Editor/SearchBar.js
+++ b/src/components/Editor/SearchBar.js
@@ -170,6 +170,7 @@ const SearchBar = React.createClass({

     if (e) {
       e.preventDefault();
+      e.stopPropagation();
     }

```

PR #2305, opened 2017-03-08
fix cmd-shift-o in the panel

jasonLaster:
Associated Issue: #2266

### Summary of Changes

* fix cmd-shift-o in the panel
* refactor shortcut handlers. was nice to do while i was in there

### DIFF
--- a/src/components/Editor/SearchBar.js
+++ b/src/components/Editor/SearchBar.js
@@ -196,6 +196,7 @@ const SearchBar = React.createClass({
 
     if (e) {
       e.preventDefault();
+      e.stopPropagation();
     }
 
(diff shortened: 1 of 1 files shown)


src/components/Editor/SearchBar.js
```diff
@@ -170,6 +196,7 @@ const SearchBar = React.createClass({
 
     if (e) {
       e.preventDefault();
+      e.stopPropagation();
```
clarkbw: Why this change?

wldcordeiro: The command opens the devtools preferences panel when you're in the Firefox panel.

Issue #2287, opened 2017-03-05
Conditional breakpoint panel close button is hidden in files with long lines


irfanhudda:
Close button of conditional breakpoint panel is hidden when viewing files with long lines.

### Steps to reproduce:
1. Open a file in debugger with long line (horizontal scrollbar in codemirror should appear)
2. Right click on gutter and select `Add Conditional Breakpoint`
3. Close button is not visible. (Can be seen if we scroll horizontally)

### Screenshot
![scr](https://cloud.githubusercontent.com/assets/1755089/23589737/c96b22c0-01f8-11e7-9580-af6e0530891d.gif)

This line might need some change [https://github.com/devtools-html/debugger.html/blob/master/src/components/Editor/ConditionalPanel.css#L20](https://github.com/devtools-html/debugger.html/blob/master/src/components/Editor/ConditionalPanel.css#L20)

aklt:
I tried to find a solution for this while working on #1945, but understood from @ryanjduffy that he did not run into this working on the same fix.  I am curious what the correct fix is for this.

jasonLaster:
So, the problem as i understand it is that the CB panel is the width of the line and we want it to be the width of the editor wrapper... I'm open to creative solutions :P

aklt:
AFAICT there is not nice CSS solution, so something creative seems to be needed, probably a hack.  Are any of you aware of an elegant solution?

aklt:
As far as I can tell there is no nice pure CSS way of solving this, but please
let me know if I am wrong.  It is possible to hook into events firing when the
ConditionalPane is opened or closed and at that point programatically place the
close button better.

The situation in the screenshot looks a bit more difficult to me.  When dragging
the `SplitBox.onMove` is called from `Draggable` and I thought there might be a
good way here somewhere to get the size of the right pane and place the close
button somewhere.  However I don't think it is possible without refactoring the
`devtools-modules`, something I do not feel comfortable doing.

Maybe there is another way?

jasonLaster:
lets try something like this:

```diff

diff --git a/src/components/Editor/index.js b/src/components/Editor/index.js
index e5a8cbe..6e71c33 100644
--- a/src/components/Editor/index.js
+++ b/src/components/Editor/index.js
@@ -362,9 +362,14 @@ const Editor = React.createClass({
     });

     this.cbPanel = this.editor.codeMirror.addLineWidget(line, panel, {
-      coverGutter: true
+      coverGutter: true,
[...]

PR #2314, opened 2017-03-09
Make close button appear correctly upon opening the conditional panel

aklt:
Associated Issue: #2287 

This is a partial fix making the close button appear correctly upon opening of the conditional panel. It is still possible to cover the button by dragging the right panel over it, however.  This does not always happen, but I was able to reproduce it.

When it becomes possible to catch drag events this issue should be revisited.  

I also added a bit of space between the button and the right panel. 

### Test

* Open source code with very long lines (ie. jQuery) in the debugger and press 
  CTRL-SHIFT-b to open the conditional panel.  The close button should be visible 
  in the conditional panel to the right of the editor.

### DIFF
--- a/src/components/shared/Button/Close.css
+++ b/src/components/shared/Button/Close.css
@@ -32,6 +32,7 @@
 
 .close-btn-big {
   padding: 11px;
+  margin-right: 7px;
   width: 27px;
   height: 40px;
 }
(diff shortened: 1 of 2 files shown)


src/components/shared/Button/Close.css
```diff
@@ -32,6 +32,7 @@
 
 .close-btn-big {
   padding: 11px;
+  margin-right: 7px;
```
jasonLaster: wouldn't this fail for a really long line?

jasonLaster: oh i see - that's nice.

Issue #1374, opened 2016-12-02
[Sources Tree] the host domain should be first


jasonLaster:
We should show the debuggee host domain first in the tree.

Notes:
* we will likely need the tab url in the redux store

Patch for debuggee url

```diff
diff --git a/src/reducers/pause.js b/src/reducers/pause.js
index ba19302..362940f 100644
--- a/src/reducers/pause.js
+++ b/src/reducers/pause.js
@@ -20,9 +20,13 @@ type PauseState = {
   loadedObjects: Object,
   shouldPauseOnExceptions: boolean,
   shouldIgnoreCaughtExceptions: boolean,
-  expressions: I.List<Expression>
+  expressions: I.List<Expression>,
+  debuggeeUrl: string
 }

+// pause is not a great place for it now
+// eventually i think a connection reducer makes sense, but... not sure
+
 const State = makeRecord(({
   pause: undefined,
   isWaitingOnBreak: false,
@@ -31,7 +35,8 @@ const State = makeRecord(({
   loadedObjects: I.Map(),
   shouldPauseOnExceptions: false,
   shouldIgnoreCaughtExceptions: false,
-  expressions: I.List()
+  expressions: I.List(),
+  debuggeeUrl: undefined,
 } : PauseState));

 function update(state = State(), action: Action): Record<PauseState> {
@@ -85,6 +90,8 @@ function update(state = State(), action: Action): Record<PauseState> {
       break;

     case constants.NAVIGATE:
+      // we could likely add it here
+      // action.value ...
       return State();
```

CC @darkwing, @Dalimil 

@Dalimil interested in grabbing this and applying your work from before?

Dalimil:
The `action` object doesn't seem to contain the URL - there is no `action.value`...

liorpr:
@jasonLaster can you elaborate on this or maybe add a photo? not sure what to add here.

jasonLaster:
Sure, right now the sources tree has a bunch of domains for every page you go to. For example, if you go to cnn.com, one domain is cnn.com, and then there are 20 other random ones. We'd like to show CNN first, but at the moment don't have a good way of doing that.

We should do this by first saving the debggee's url when we get a navigate event. We can check the event data

Here's a good way of getting the debuggee url when we navigate:

```diff
diff --git a/src/actions/navigation.js b/src/actions/navigation.js
index 71c2149..d27df53 100644
--- a/src/actions/navigation.js
+++ b/src/actions/navigation.js
[...]

PR #2316, opened 2017-03-09
Show hostname first (#1374)

Dalimil:
Associated Issue: #1374 
This was pushed with --no-verify and still needs a clean-up.

out

### DIFF
--- a/src/actions/types.js
+++ b/src/actions/types.js
@@ -149,6 +149,8 @@ type PauseAction =
   | { type: "DELETE_EXPRESSION",
       input: string };
 
+type NavigateAction = { type: "NAVIGATE", url: string };
+
 /**
  * Actions: Source, Breakpoint, and Navigation
  *
@@ -159,5 +161,5 @@ export type Action =
   SourceAction
   | BreakpointAction
   | PauseAction
-  | { type: "NAVIGATE" }
+  | NavigateAction
   | UIAction;
--- a/src/components/SourcesTree.js
+++ b/src/components/SourcesTree.js
@@ -6,7 +6,10 @@ import { DOM as dom, PropTypes, Component, createFactory } from "react";
 import classnames from "classnames";
 import ImPropTypes from "react-immutable-proptypes";
 import { Set } from "immutable";
-import { getShownSource, getSelectedSource } from "../selectors";
+import {
+  getShownSource, getSelectedSource, getDebuggeeUrl
+} from "../selectors";
+
 import {
   nodeHasChildren, createParentMap, isDirectory, addToTree, collapseTree,
   createTree, getDirectories
@@ -40,7 +43,7 @@ class SourcesTree extends Component {
 
   constructor(props) {
     super(props);
-    this.state = createTree(this.props.sources);
+    this.state = createTree(this.props.sources, this.props.debuggeeUrl);
 
     this.focusItem = this.focusItem.bind(this);
     this.selectItem = this.selectItem.bind(this);
@@ -101,7 +104,9 @@ class SourcesTree extends Component {
     }
 
     if (nextProps.sources.size === 0) {
-      this.setState(createTree(nextProps.sources));
+      this.setState(
+        createTree(nextProps.sources, this.props.debuggeeUrl)
+      );
       return;
     }
 
@@ -111,7 +116,7 @@ class SourcesTree extends Component {
 
     const uncollapsedTree = this.state.uncollapsedTree;
     for (let source of newSet) {
-      addToTree(uncollapsedTree, source);
+      addToTree(uncollapsedTree, source, this.props.debuggeeUrl);
     }
 
     // TODO: recreating the tree every time messes with the expanded
@@ -260,7 +265,8 @@ SourcesTree.propTypes = {
   sources: ImPropTypes.map.isRequired,
   selectSource: PropTypes.func.isRequired,
   shownSource: PropTypes.string,
-  selectedSource: ImPropTypes.map
+  selectedSource: ImPropTypes.map,
+  debuggeeUrl: PropTypes.string.isRequired
 };
 
 SourcesTree.displayName = "SourcesTree";
@@ -269,7 +275,8 @@ export default connect(
   state => {
     return {
       shownSource: getShownSource(state),
-      selectedSource: getSelectedSource(state)
+      selectedSource: getSelectedSource(state),
+      debuggeeUrl: getDebuggeeUrl(state)
     };
   },
   dispatch => bindActionCreators(actions, dispatch)
--- a/src/utils/sources-tree.js
+++ b/src/utils/sources-tree.js
@@ -163,7 +163,7 @@ function isDirectory(url: Object) {
  * @memberof utils/sources-tree
  * @static
  */
-function addToTree(tree: any, source: TmpSource) {
+function addToTree(tree: any, source: TmpSource, debuggeeUrl: string) {
   const url = getURL(source.get("url"));
 
   if (IGNORED_URLS.indexOf(url) != -1 ||
@@ -195,7 +195,8 @@ function addToTree(tree: any, source: TmpSource) {
     assert(nodeHasChildren(subtree), `${subtree.name} should have children`);
     const children = subtree.contents;
 
-    let index = determineFileSortOrder(children, part, isLastPart);
+    let index = determineFileSortOrder(children, part, isLastPart,
+      i === 0 ? debuggeeUrl : "");
 
     if (index >= 0 && children[index].name === part) {
       // A node with the same name already exists, simply traverse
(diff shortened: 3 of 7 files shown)


src/utils/sources-tree.js
```diff
@@ -242,6 +255,20 @@ function determineFileSortOrder(nodes:Array<Node>, pathPart:string,
...
+      const rootUrlMatch = isExactUrlMatch(pathPart, debuggeeUrl);
+      const nodeUrlMatch = isExactUrlMatch(node.name, debuggeeUrl);
+      if (rootUrlMatch) {
+        // pathPart must go first
+        return true;
+      }
+      if (nodeUrlMatch) {
+        // Examined item must go first
+        return false;
+      }
```
jasonLaster: nit: 

```js
return rootUrlMatch && !nodeUrlMatch
```

src/utils/sources-tree.js
```diff
@@ -285,17 +312,18 @@ function collapseTree(node: any, depth: number = 0) {
...
+    addToTree(uncollapsedTree, source, debuggeeUrl);
   }
   const sourceTree = collapseTree(uncollapsedTree);
 
   return { uncollapsedTree,
     sourceTree,
     parentMap: createParentMap(sourceTree),
-    focusedItem: null };
+    focusedItem: null,
+    debuggeeUrl };
```
jasonLaster: lets get some unit tests in here. Should be pretty similar to what we already have

jasonLaster:
@Dalimil I think this should fix the flow error:

```js
diff --git a/src/actions/types.js b/src/actions/types.js
index 0d3f889..75cf6d6 100755
--- a/src/actions/types.js
+++ b/src/actions/types.js
@@ -149,6 +149,8 @@ type PauseAction =
   | { type: "DELETE_EXPRESSION",
       input: string };

+type NavigateAction = { type: "NAVIGATE", debuggeeUrl: string }
+
 /**
  * Actions: Source, Breakpoint, and Navigation
  *
@@ -159,5 +161,5 @@ export type Action =
   SourceAction
   | BreakpointAction
   | PauseAction
-  | { type: "NAVIGATE" }
+  | NavigateAction
   | UIAction;
```

src/actions/types.js
```diff
@@ -159,5 +161,5 @@ export type Action =
   SourceAction
   | BreakpointAction
   | PauseAction
-  | { type: "NAVIGATE" }
+  | NavigateAction
```
wldcordeiro: 👍

src/utils/sources-tree.js
```diff
@@ -163,7 +163,7 @@ function isDirectory(url: Object) {
  * @memberof utils/sources-tree
  * @static
  */
-function addToTree(tree: any, source: TmpSource) {
+function addToTree(tree: any, source: TmpSource, debuggeeUrl: any) {
```
wldcordeiro: Is the URL really an any? I don't mind using it but if there's a type we could use it's preferable.

src/utils/sources-tree.js
```diff
@@ -224,13 +225,27 @@ function addToTree(tree: any, source: TmpSource) {
 }
 
 /**
+ * @memberof utils/sources-tree
+ * @static
+ */
+function isExactUrlMatch(pathPart:string, debuggeeUrl:string) {
```
wldcordeiro: Testing this utility would be nice.

src/utils/sources-tree.js
```diff
@@ -224,13 +225,27 @@ function addToTree(tree: any, source: TmpSource) {
...
+
+/**
  * Look at the nodes in the source tree, and determine the index of where to
  * insert a new node. The ordering is index -> folder -> file.
  * @memberof utils/sources-tree
  * @static
  */
 function determineFileSortOrder(nodes:Array<Node>, pathPart:string,
-                                isLastPart:boolean) {
+                                isLastPart:boolean, debuggeeUrl:any) {
```
wldcordeiro: Same thing here, can we use something other than any?

src/components/SourcesTree.js
```diff
@@ -24,21 +26,23 @@ type CreateTree = {
...
-    selectedSource: ImPropTypes.map
+    selectedSource: ImPropTypes.map,
+    debuggeeUrl: PropTypes.string
   },
 
   displayName: "SourcesTree",
 
   getInitialState(): CreateTree {
-    return createTree(this.props.sources);
+    return createTree(this.props.sources, this.props.debuggeeUrl || "");
```
jasonLaster: we can improve this with this nifty trick

```js
getDefaultProps() {}
```

src/components/SourcesTree.js
```diff
@@ -95,7 +101,7 @@ let SourcesTree = React.createClass({
 
     const uncollapsedTree = this.state.uncollapsedTree;
     for (let source of newSet) {
-      addToTree(uncollapsedTree, source);
+      addToTree(uncollapsedTree, source, this.state.debuggeeUrl);
```
jasonLaster: shouldn't this be props?

src/components/SourcesTree.js
```diff
@@ -237,7 +243,8 @@ module.exports = connect(
   state => {
     return {
       shownSource: getShownSource(state),
-      selectedSource: getSelectedSource(state)
+      selectedSource: getSelectedSource(state),
+      debuggeeUrl: getDebuggeeUrl(state)
```
jasonLaster: i see... lets stick to props and just compute this every time we need it. Also - is there a better name for the function to differentiate from the raw value

src/utils/sources-tree.js
```diff
@@ -224,13 +225,27 @@ function addToTree(tree: any, source: TmpSource) {
 }
 
 /**
+ * @memberof utils/sources-tree
+ * @static
+ */
+function isExactUrlMatch(pathPart:string, debuggeeUrl:string) {
+  // compare to hostname with an optional 'www.' prefix
+  const urlParts = /^(?:\w+\:\/\/)?([^\/]+)(.*)$/.exec(debuggeeUrl);
+  if (urlParts == null) {
```
jasonLaster: maybe three equals `===`

src/utils/sources-tree.js
```diff
@@ -224,13 +225,27 @@ function addToTree(tree: any, source: TmpSource) {
...
+    return false;
+  }
+  const baseUrl = urlParts[1].replace(/^www\./, "");
+  return pathPart === baseUrl || pathPart === `www.${baseUrl}`;
+}
```
jasonLaster: what's this trying to solve?

src/utils/sources-tree.js
```diff
@@ -242,6 +257,21 @@ function determineFileSortOrder(nodes:Array<Node>, pathPart:string,
...
+      const nodeUrlMatch = isExactUrlMatch(node.name, debuggeeUrl);
+      if (rootUrlMatch) {
+        // pathPart matches root url and must go first
+        return true;
+      }
+      if (nodeUrlMatch) {
+        // Examined item matches root url and must go first
+        return false;
+      }
+      // If neither is the case, continue to compare alphabetically
```
jasonLaster: comments really help here

src/utils/sources-tree.js
jasonLaster: actually, i'm pretty sure there must be something in:

`URL.parse` or `new URL` that would give you a hostname you could compare

Dalimil: I know but I cannot pass any non-browser unit-tests otherwise...

src/components/SourcesTree.js
Dalimil: hmmm... I tried it just now but flow is still complaining about a potential `undefined` value

src/utils/sources-tree.js
jasonLaster: ahh, i see - i think we solved this somewhere else as well :)

src/components/SourcesTree.js
jasonLaster: oh yeah - i'm not surprised... I'll look quickly

src/utils/sources-tree.js
jasonLaster: Hmm - we do this `const { parse } = require("url")` in several places

src/components/SourcesTree.js
jasonLaster: try adding `isRequired` to the prop with `getDefaultProps`

src/utils/sources-tree.js
```diff
@@ -224,13 +225,27 @@ function addToTree(tree: any, source: TmpSource) {
...
+ */
+function isExactUrlMatch(pathPart:string, debuggeeUrl:string) {
+  // compare to hostname with an optional 'www.' prefix
+  const { host } = parse(debuggeeUrl);
+  if (!host) {
+    return false;
+  }
+  const baseUrl = host.replace(/^www\./, "");
+  return pathPart === baseUrl || pathPart === `www.${baseUrl}`;
+}
```
jasonLaster: Hey @Dalimil can we try `parse` which is required in the file

Dalimil: You mean `parse(pathPart)` ? I'm already using it for the debuggeeUrl...

jasonLaster: oh nice - sorry missed that :)

src/utils/sources-tree.js
```diff
@@ -224,13 +225,27 @@ function addToTree(tree: any, source: TmpSource) {
...
+ * @static
+ */
+function isExactUrlMatch(pathPart:string, debuggeeUrl:string) {
+  // compare to hostname with an optional 'www.' prefix
+  const { host } = parse(debuggeeUrl);
+  if (!host) {
+    return false;
+  }
+  const baseUrl = host.replace(/^www\./, "");
+  return pathPart === baseUrl || pathPart === `www.${baseUrl}`;
```
jasonLaster: this seems slightly more complicated than it needs to be. Could we also replace the `www` for the pathPart?

```js
function isExactUrlMatch(pathPart:string, debuggeeUrl:string) {
  function getHost(url) {
    const { host } = parse(debuggeeUrl);
    if (!host) {
      return ""
    }
    return host.replace(/^www\./, "");
  }

  return getHost(pathPart) == getHost(debuggeeUrl)
}
```

Dalimil:
When I try to `parse(pathPart)` I'd get a `null` for the host because the URL parsing expects a protocol but `pathPart` is usually just `www.example.com`

jasonLaster:
ahh - i see. In that case, we can do something simpler :)

Issue #2333, opened 2017-03-10
console error: Required prop `selectedLocation` was not specified in `Editor`


clarkbw:
I'm seeing this console error when starting up the debugger.

```
Warning: Failed prop type: Required prop `selectedLocation` was not specified in `Editor`.
    in Editor (created by Connect(Editor))
    in Connect(Editor) (created by App)
    in div (created by App)
    in div (created by App)
    in div (created by SplitBox)
    in div (created by SplitBox)
    in SplitBox (created by App)
    in div (created by SplitBox)
    in div (created by SplitBox)
    in SplitBox (created by App)
    in div (created by App)
    in App (created by Connect(App))
    in Connect(App)
    in Provider  debugger.js:2580:10
	printWarning http://localhost:8000/assets/build/debugger.js:2580:10
	warning http://localhost:8000/assets/build/debugger.js:2604:10
	checkReactTypeSpec http://localhost:8000/assets/build/debugger.js:14466:18
	validatePropTypes http://localhost:8000/assets/build/debugger.js:21520:6
	createElement http://localhost:8000/assets/build/debugger.js:21556:6
	render http://localhost:8000/assets/build/debugger.js:48696:36
	_renderValidatedComponentWithoutOwnerOrContext/renderedComponent< http://localhost:8000/assets/build/debugger.js:14224:17
	measureLifeCyclePerf http://localhost:8000/assets/build/debugger.js:13505:13
	_renderValidatedComponentWithoutOwnerOrContext http://localhost:8000/assets/build/debugger.js:14223:28
	_renderValidatedComponent http://localhost:8000/assets/build/debugger.js:14250:30
	performInitialMount http://localhost:8000/assets/build/debugger.js:13792:26
	mountComponent http://localhost:8000/assets/build/debugger.js:13688:17
	mountComponent http://localhost:8000/assets/build/debugger.js:4227:19
	mountChildren http://localhost:8000/assets/build/debugger.js:12642:29
	_createInitialChildren http://localhost:8000/assets/build/debugger.js:8287:28
	mountComponent http://localhost:8000/assets/build/debugger.js:8112:8
	mountComponent http://localhost:8000/assets/build/debugger.js:4227:19
	mountChildren http://localhost:8000/assets/build/debugger.js:12642:29
	_createInitialChildren http://localhost:8000/assets/build/debugger.js:8287:28
	mountComponent http://localhost:8000/assets/build/debugger.js:8112:8
[...]

PR #2334, opened 2017-03-11
fix console error for selectedLocation

jbhoosreddy:
Associated Issue: #2333 

Removed isRequired property from selectedLocation and add sanity checks for yarn errors.


clarkbw:
Checking this out locally now...

src/components/Editor/index.js
```diff
@@ -391,14 +392,18 @@ const Editor = React.createClass({
       return;
     }
 
+    let sourceId = this.props.selectedLocation ?
```
clarkbw: Any reason for `let` instead of `const` here?

Issue #2339, opened 2017-03-11
Editor Preview no longer working


bomsy:
The editor preview is no longer showing

PR #2341, opened 2017-03-11
fix editow preview issue

bomsy:
Associated Issue: #2339


bomsy:
@jcreighton do you want to test and review this?

bomsy:
Hey thanks! I see you also have the fix in ur Editor preview PR , maybe we'll just go with that!

Issue #1890, opened 2017-01-30
Convert to es6 classes


jasonLaster:
Here's a [gist](https://gist.github.com/jasonLaster/d95a10362fe1e6b687aabba2df26b72e) with some notes.

I'm not sure if we want to switch to the decorator, but the explicit map functions are nice

## Components

- [x] App
- [x] Sources
- [x] SourcesTree
- [x] SourceSearch
- [x] WelcomeBox
- [x] Editor
  - [x] Breakpoint
  - [x] Footer
  - [x] HitMarker
  - [x] Editor *
  - [x] SearchBar*
  - [x] Tabs
  - [x] Preview
- [x] SecondaryPanes
  - [x] Breakpoints
  - [x] ChromeScopes
  - [x] CommandBar
  - [x] EventListeners
  - [x] Expressions
  - [x] Frames
  - [x] SecondaryPanes
  - [x] Scopes
  - [x] WhyPaused
- [x] shared
  - [x] Button
    - [x] PaneToggle
  - [x] Accordion
  - [x] AutoComplete
  - [x] Dropdown
  - [x] ManagedTree
  - [x] ObjectInspector
  - [x] Popover
  - [x] ResultList
  - [x] SearchInput

**Note: Asterisks next to the component are probably harder tasks than the others.**

tomrav:
Starting to take a look at this.
<3 #goodnessSquad

wldcordeiro:
Added a list to track the components that have been converted.

ryanjduffy:
Opened a PR for converting the SecondaryPanes

jasonLaster:
We're currently **16/32** components refactored!

peterschussheim:
Starting to work on  Breakpoint, Footer, HitMarker.

jasonLaster:
Great!

deneuv34:
may I join to work on AutoComplete?

jasonLaster:
Sure!

peterschussheim:
Starting work on SearchBar

ryanjduffy:
^^^ PR for Editor

ryanjduffy:
^^^ PR for remaining shared components

jasonLaster:
Thanks @AnshulMalik @ryanjduffy @wldcordeiro and everyone else!

PR #2342, opened 2017-03-12
converted src/components/App.js to es6

AnshulMalik:
Associated Issue: #1890

### Summary of Changes

* Migrated `src/components/App.js` to use es6 classes

### DIFF
--- a/src/components/App.js
+++ b/src/components/App.js
@@ -1,23 +1,23 @@
 // @flow
-const React = require("react");
+import React from "react";
 const { DOM: dom, PropTypes, createFactory } = React;
-const { connect } = require("react-redux");
-const { bindActionCreators } = require("redux");
-const actions = require("../actions");
-const { getSources, getSelectedSource, getPaneCollapse } = require("../selectors");
+import { connect } from "react-redux";
+import { bindActionCreators } from "redux";
+import actions from "../actions";
+import { getSources, getSelectedSource, getPaneCollapse } from "../selectors";
 
-const { KeyShortcuts } = require("devtools-sham-modules");
+import { KeyShortcuts } from "devtools-sham-modules";
 const shortcuts = new KeyShortcuts({ window });
 
 const verticalLayoutBreakpoint = window.matchMedia("(min-width: 800px)");
 
-require("./variables.css");
-require("./App.css");
-require("./shared/menu.css");
-require("./shared/SplitBox.css");
-require("./shared/reps.css");
-let { SplitBox } = require("devtools-modules");
-SplitBox = createFactory(SplitBox);
+import "./variables.css";
+import "./App.css";
+import "./shared/menu.css";
+import "./shared/SplitBox.css";
+import "./shared/reps.css";
+
+const SplitBox = createFactory(require("devtools-modules").SplitBox);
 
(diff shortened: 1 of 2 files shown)


src/main.js
```diff
@@ -16,7 +16,7 @@ const { onConnect, onFirefoxConnect } = require("./utils/client");
 const reducers = require("./reducers");
 const selectors = require("./selectors");
 
-const App = require("./components/App");
+const App = require("./components/App").default;
```
wldcordeiro: If you use import/export for the `App` file I think changing this line  to
`import App from './components/App'` isn't bad.

AnshulMalik: I thought I'd do that when I migrate main.js file. But essentially the same thing.

wldcordeiro: It kind of acts as an indicator that `App` was done. 😄

src/components/App.js
```diff
@@ -26,38 +26,35 @@ const SecondaryPanes = createFactory(require("./SecondaryPanes"));
...
-  },
+class App extends React.Component {
+  state: {
+    horizontal: verticalLayoutBreakpoint.matches
+  }
 
-  displayName: "App",
+  constructor(props) {
+    super(props);
+    this.state = {
```
wldcordeiro: Doesn't this duplicate the work above at line 30?

src/components/App.js
```diff
@@ -135,19 +132,29 @@ const App = React.createClass({
...
+
+App.propTypes = {
+  sources: PropTypes.object,
+  selectSource: PropTypes.func,
+  selectedSource: PropTypes.object,
+  startPanelCollapsed: PropTypes.bool,
+  endPanelCollapsed: PropTypes.bool,
+};
+
+App.displayName = "App";
```
wldcordeiro: Could we use a getter for displayName and propTypes?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get

src/components/App.js
AnshulMalik: Hmm yes, but I was getting some flow errors if I don't define state out of constructor
`Did you forget to declare type parameter `State` of property `Component`?`

Any suggestions?

wldcordeiro: @AnshulMalik I see, didn't know Flow was the reason for it. I think we could make the first one a type annotation for Flow like what is discussed [here](https://github.com/facebook/flow/issues/1594).

src/components/App.js
AnshulMalik: A quick question, making it getter will mean we are binding it to the object but I think propTypes should be attached to the class?

wldcordeiro: I think you would make it a static getter to attach to the class. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/static

AnshulMalik: Another issue with static getter for propTypes
https://github.com/facebook/flow/issues/3008

jasonLaster: Yep - i think this is fine. We do this in another place as well and it seems like the current practice

src/main.js
jasonLaster: @AnshulMalik this is good. I agree we should be consistent in a file.

Issue #1925, opened 2017-02-03
[Windows]  source-map unit tests fail


Garbee:
`@princiya` brought this to my attention in Slack... She was getting timeout errors on the tests for sourcemaps in the *CLI* on Windows but not in the web mocha runner. So I tested both on my Linux system, all clear in CLI but Mocha on the web view fails on `should add a breakpoint with the same error.

It looks like any given test could end up flaky based on the timeout depending on what the hardware, OS, and other factors of the machine are. I'm not sure if we should increase the timeout, since we do want to verify things are completed in a timely manner. This should at least be documented somehow, and a fix so the flakyness doesn't occur investigated to see if there is a viable solution.

jasonLaster:
Lets try and isolate the problem a bit. My guess is that it is likely tests that use a worker like the `sourcemap` test.

Perhaps we should try running the unit tests 100 times and see if we can find any patterns?

jasonLaster:
We moved the test to the source map library. Now it's fine.

PR #2343, opened 2017-03-12
Fix failing test on windows

rrandom:
Associated Issue: #1925 

### Summary of Changes

Previous expriment goes here https://github.com/devtools-html/devtools-core/pull/228

Let webpack `require` the correct node's `path` module in  https://github.com/devtools-html/devtools-core/blob/master/packages/devtools-network-request/stubNetworkRequest.js

### Test Plan

On Windows, run `yarn test`,  all test passed
On Ubuntu, run `yarn test`,  all test passed

### DIFF
--- a/src/test/node-unit-tests.js
+++ b/src/test/node-unit-tests.js
@@ -47,6 +47,8 @@ webpackConfig.resolve.alias["devtools-network-request"] =
 
 webpackConfig.node = { __dirname: false };
 
+webpackConfig.target = "node";
+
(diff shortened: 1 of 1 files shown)


jasonLaster:
@rrandom thanks. Mind checking in on this in a couple days.

We'll be switching to jest and i'm curious if we'll have the same bug

rrandom:
@jasonLaster Ok, that's cool. Later

Issue #2304, opened 2017-03-08
[Dark Theme] polish colors


jasonLaster:
Checklist:
- [ ] select state
- [x] call stack zebra
- [x] consistency of colors in the secondary pane
- [x] pause box
- [x] highlight color (the editor highlight is very strong)
- [x] Improve color contrast in tree-node #1940

**select state** the blue is pretty jarring and unnecessary. It could be much softer and have the same effect.

![Screen Shot 2017-03-08 at 10.53.09 AM.png](https://shipusercontent.com/b71ac606b30835b8b3db4f119cde0833/Screen%20Shot%202017-03-08%20at%2010.53.09%20AM.png)

**consistency** see the watch expression placeholder, and input color compared to the "no breakpoints" colors.

|1|2|
|--|--|
|![con-1.png](https://shipusercontent.com/db0bb840a82b994b2cf539a59df1b77c/con-1.png)|![con-2.png](https://shipusercontent.com/3bccb8435f43b2248f9eef7d9e464478/con-2.png)|

**zebra call stack**
![zebra.png](https://shipusercontent.com/0498fe6d503446e55f52bd8077e35cf4/zebra.png)

**pause box** is a really sharp green, we can dampen that a bit and still be okay.

![pause box.png](https://shipusercontent.com/3f05bc79c27fcb4839202d2718081646/pause%20box.png)

jasonLaster:
@amitzur is on it

jasonLaster:
@amitzur any updates?

amitzur:
[I also pushed to devtools-core a new variable]

Take a look and see if you find these sensible:

- **call stack zebra**: The dark theme and light theme need to use exactly the opposite variables. This should ideally be done via better variable construction, but I just did separate selectors for now.
I also added a hover state for dark theme for buttons, accordion headers and frames.

- **pause box**: used a slightly less saturated color and removed the border. It actually applies to all active breakpoint greens, but it almost doesn't show except for the large area pause box.

- **consistency**: I changed the input color to normal body text color. And I actually like the placeholder being more faded than the "No xxx" messages in other sections, since it's really a placeholder and not a message.


I still didn't get to, and have questions on:
- select state (do you mean to change this color throughout the entire UI ?)
- editor highlight (what is that exactly?)
- color contrast

jasonLaster:
* **select state** is the blue highlight in the source tree and function search list
* **editor highlight** is the green pause line for instance which you dulled
[...]

PR #2344, opened 2017-03-12
Polish colors in dark theme

amitzur:
Associated Issue: #2304 

### Summary of Changes

- **call stack zebra**: The dark theme and light theme need to use exactly the opposite variables. This should ideally be done via better variable construction, but I just did separate selectors for now.
I also added a hover state for dark theme for buttons, accordion headers and frames.

- **pause box**: used a slightly less saturated color and removed the border. It actually applies to all active breakpoint greens, but it almost doesn't show except for the large area pause box.

- **consistency**: I changed the input color to normal body text color. And I actually like the placeholder being more faded than the "No xxx" messages in other sections, since it's really a placeholder and not a message.

### Test Plan

No tests


src/components/SecondaryPanes/Frames.css
```diff
@@ -73,10 +82,11 @@
   cursor: pointer;
   text-align: center;
   padding: 8px 0px;
-  border-top: 1px solid var(--theme-splitter-color);
+  margin: 7px 10px 7px 7px;
```
jasonLaster: 👍

jasonLaster:
|Before|After|
|---|--|
|![screen shot 2017-03-12 at 2 47 10 pm]|
screen shot 2017-03-12 at 2 46 15 pm||



@amitzur everything looks great. One question though about the header hover state is that it comes off as a bit washed out.

jasonLaster:
just merged the hover color which will help!

jasonLaster:
@amitzur merging as this is a lot of good work and I want to avoid future merge conflicts. Lets open new PRs for the other style improvements

Issue #1884, opened 2017-01-30
Convert to es modules


jasonLaster:
### We are switching to es6 modules for several reasons: 
* flow uses es6 modules
* it's more expressive
* we want to use es6 classes 

[cjs-to-es6](https://github.com/nolanlawson/cjs-to-es6) is a great tool for converting a module

### Steps to convert a module
* `cjs-to-es6 src/components/SecondaryPanes/index.js`
* update import sites to use `default`

#### nested destructuring. watch out for `cjs-to-es6` missing this

```js
const { Services: { appinfo }} = require("devtools-modules");
```

```js
import { Services } from "devtools-module";
const { appinfo } = Services;
```

#### mocking modules

The unit tests mock some modules like [`devtools-modules`](https://github.com/devtools-html/debugger.html/blob/master/src/test/node-unit-tests.js#L15). `mock-require` doesn't work with es modules so we should find a work around.

#### react-dom

For some reason `import dom from "react-dom"` does not work, so just leave that one as is.

Ask all questions in our **slack** room. This can potentialy be tricky with edge cases.

### Progress converting the codebase

- [x] actions
  - [x] breakpoints.js
  - [x] coverage.js
  - [x] event-listeners.js
  - [x] expressions.js
  - [x] index.js
  - [x] navigation.js
  - [x] pause.js
  - [x] sources.js
- [ ] components
  - [x] App.js
  - [ ] Editor
    - [x] Breakpoint.js
    - [ ] ConditionalPanel.js
    - [x] Footer.js
    - [ ] HitMarker.js
    - [ ] SearchBar.js
    - [x] Tabs.js
    - [ ] index.js
    - [x] Preview.js
  - [ ] SecondaryPanes
    - [x] Breakpoints.js
    - [x] ChromeScopes.js
    - [x] CommandBar.js
    - [x] EventListeners.js
    - [x] Expressions.js
    - [x] Frames.js
    - [x] Scopes.js
    - [x] WhyPaused.js
    - [ ] index.js
  - [x] SourceSearch.js
  - [x] Sources.js
  - [x] SourcesTree.js
  - [x] WelcomeBox.js
  - [ ] shared
    - [x] Accordion.js
    - [ ] Autocomplete.js
    - [x] Button
    - [x] Close.js
    - [x] PaneToggle.js
    - [x] Dropdown.js
    - [x] ManagedTree.js
    - [x] ObjectInspector.js
    - [x] Rep.js
    - [x] Svg.js
[...]

PR #2348, opened 2017-03-12
convert WelcomeBox to es modules

clarkbw:
This is part of the #1884 work


jasonLaster:
i think we also want to update the export

clarkbw:
Oh yeah!  Thanks for the reminder!  Will do on the plane.

PR #2352, opened 2017-03-13
Es module secondary panes

clarkbw:
This is part of the #1884 work


clarkbw:
Flow coverage threshold!

PR #2357, opened 2017-03-13
Remove overlay mask. Fixes #1952

### LINKED ISSUE #1952: Right-clicking when the context-menu is displayed should hide the context-menu
STR : 
1. Right click where there should be a debugger context-menu (e.g. a file tab )
2. Right click again

Expected result: The application context menu should be hidden
Actual result: The native context menu is displayed, in addition with the application one (see ⬇ )

peek 07-02-2017 13-58

binarybaba:
Issue: #1952 

### Summary of Changes

* Removed mask and its listeners
* Add event listener on debugger body for clearing popup


src/components/shared/menu.js
```diff
@@ -14,35 +14,25 @@ function createPopup(doc) {
...
-    mask = doc.createElement("div");
-    mask.id = "contextmenu-mask";
-    if (document.body) {
-      document.body.appendChild(mask);
+  let debuggerWindow = document.getElementsByClassName("debugger")[0];
+  if (debuggerWindow) {
+    debuggerWindow.onclick = () => {
+      if (popup) {
+        popup.hidePopup();
+      }
```
jasonLaster: my concern with this approach is that some click handlers will prevent propogation so this will not always work.

I think the safer thing to do would be to limit the impact of right clicking on a mask so that it doesn't show a context menu.

for instance preventDefault, stop propogation

binarybaba:
Oh dang yes, my bad. Sorry. Will be on it.

PR #2368, opened 2017-03-14
Add hover effect to close-btn-big

irfanhudda:
### Summary of Changes

* `close-btn-big` is used in file search and conditional breakpoint panel.
* Fix another typo in `transition-timing-function` property value

|File search|
|--------------|
|no-hover-file-search|
|hover-file-search|

### DIFF
--- a/src/components/shared/Button/Close.css
+++ b/src/components/shared/Button/Close.css
@@ -21,11 +21,13 @@
   display: block;
 }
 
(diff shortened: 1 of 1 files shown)


jasonLaster:
:1st_place_medal: Great eye!

Perhaps the css here is a little awkward. maybe we need a separate `big` instead of `close-btn-big`...

PR #2377, opened 2017-03-15
remove executable bit from files not needing it

aklt:
:-)


jasonLaster:
Thanks for finding this :mag:

PR #2380, opened 2017-03-15
Refactor(ES6) components/shared/Button/*

AnshulMalik:
Associated Issue: #1884 #1890 

### Summary of Changes

* converted components/shared/Button/PaneToggle.js
* converted components/shared/Button/Close.js

### DIFF
--- a/src/components/Editor/Tabs.js
+++ b/src/components/Editor/Tabs.js
@@ -12,7 +12,9 @@ import { getFilename, isPretty } from "../../utils/source";
 import classnames from "classnames";
 import actions from "../../actions";
 import CloseButton from "../shared/Button/Close";
-const PaneToggleButton = createFactory(require("../shared/Button/PaneToggle"));
+const PaneToggleButton = createFactory(
+  require("../shared/Button/PaneToggle").default
+);
 import Svg from "../shared/Svg";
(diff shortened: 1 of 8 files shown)


src/components/Editor/Tabs.js
```diff
@@ -12,7 +12,9 @@ import { getFilename, isPretty } from "../../utils/source";
 import classnames from "classnames";
 import actions from "../../actions";
 import CloseButton from "../shared/Button/Close";
-const PaneToggleButton = createFactory(require("../shared/Button/PaneToggle"));
+const PaneToggleButton = createFactory(
```
jasonLaster: What if we did:

```js
import _PaneToggleButton from "../shared/Button/Close";
const PaneToggleButton = createFactory(_PaneToggleButton)
```

this way we could separate concerns and still use import, and then createFactory below

AnshulMalik: We can do this, and I have another possibility in mind, 
move createFactory inside component itself, so that way we need not to call createFactory every time we import a component.
Also @amitzur agrees that this way we can easily transition to JSX.

jasonLaster:
Lets cleanup the factory pattern in another pass

Issue #2382, opened 2017-03-15
[L10N] localize keyshortcuts


jasonLaster:
We should localize the keyboard shortcuts keys like `cmd`, `ctrl`...

https://github.com/devtools-html/debugger.html/blob/master/docs/local-development.md#l10n

Here is an example of localizing one of them:

```diff
diff --git a/assets/panel/debugger.properties b/assets/panel/debugger.properties
index 7b5b994..4ba220e 100644
--- a/assets/panel/debugger.properties
+++ b/assets/panel/debugger.properties
@@ -585,3 +585,6 @@ whyPaused.debugCommand=Paused on debugged function
 # in a info block explaining how the debugger is currently paused on an event
 # listener breakpoint set
 whyPaused.other=Debugger paused
+
+# LOCALIZATION NOTE (ctrl): The text that is used to describe a key shortcut
+ctrl=Ctrl
diff --git a/src/strings.json b/src/strings.json
index a1ba26b..dcbc8ec 100644
--- a/src/strings.json
+++ b/src/strings.json
@@ -4,6 +4,7 @@
   "breakpoints.enable": "Enable Breakpoints",
   "breakpoints.disable": "Disable Breakpoints",
   "breakpoints.removeBreakpointTooltip": "Remove Breakpoint",
+  "ctrl": "Ctrl",
   "callStack.header": "Call Stack",
   "callStack.notPaused": "Not Paused",
   "callStack.collapse": "Collapse Rows",
diff --git a/src/utils/text.js b/src/utils/text.js
index d81942a..6cff654 100644
--- a/src/utils/text.js
+++ b/src/utils/text.js
@@ -28,7 +28,7 @@ function formatKeyShortcut(shortcut: string): string {
                    .replace(/CommandOrControl\+|CmdOrCtrl\+/g, "\u2318+")
                    .replace(/Alt\+/g, "\u2325+");
   }
-  return shortcut.replace(/CommandOrControl\+|CmdOrCtrl\+/g, "Ctrl+");
+  return shortcut.replace(/CommandOrControl\+|CmdOrCtrl\+/g, L10n.getStr("ctrl") "+");
 }
```

PR #2384, opened 2017-03-15
chore: localized keyshortcuts - #goodnessSquad (#2382)

karenyavine:
Associated Issue: #2382

### Summary of Changes

* localized keyshortcuts

### Test Plan

- [x] Hover over top-right command bar

### DIFF
--- a/src/strings.json
+++ b/src/strings.json
@@ -85,5 +85,6 @@
   "collapsePanes": "Collapse panes",
   "expandPanes": "Expand panes",
   "copySourceUrl": "Copy Source Url",
-  "copySourceUrl.accesskey": "c"
+  "copySourceUrl.accesskey": "c",
+  "ctrl": "Ctrl"
 }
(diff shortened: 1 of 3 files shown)


src/strings.json
```diff
@@ -85,5 +85,9 @@
   "collapsePanes": "Collapse panes",
   "expandPanes": "Expand panes",
   "copySourceUrl": "Copy Source Url",
-  "copySourceUrl.accesskey": "c"
+  "copySourceUrl.accesskey": "c",
+  "shift": "\u21E7",
+  "cmd": "\u2318",
+  "alt": "\u2325",
```
jasonLaster: Hmm now that i'm looking more closely perhaps we don't need to localize the unicode keys :) shift, cmd, alt...

jasonLaster:
Hi @karenyavine! I got confirmation that we don't need to localize the unicode characters. Could you update this so it's just the Ctrl translation?

jasonLaster:
Thanks Karen, I went ahead and cleaned it up.

Have you joined our slack room? https://devtools-html-slack.herokuapp.com

Issue #2371, opened 2017-03-14
Show message for pages with no Javascript to debug


ruturajv:
The default page that is opened with `yarn run firefox` opens up a page that is not having any Javascript: https://devtools-html.github.io/debugger-examples/

On opening the debugger window of that tab, it just shows an empty pane / structure. A new user like me might get stuck thinking its not working; It'll be great if a message like "No JavaScript to debug" is present.

![screenshot from 2017-03-14 21-17-24](https://cloud.githubusercontent.com/assets/227817/23909155/faf9ccec-08fb-11e7-873d-d01a73240b7b.png)

jasonLaster:
This is similar to #2231

Issue #2231, opened 2017-03-01
React DevTools causes sources to be empty on debugger-examples


juliandescottes:
When the page has no sources, the current source tree is empty:

<img width="1002" alt="image" src="https://cloud.githubusercontent.com/assets/1141550/23451730/8321854c-fe61-11e6-8a8d-5ee6a73b116d.png">

It would be nice if we could have a message stating that there are no sources. When I saw this UI, I thought there was a bug in the debugger which prevented from listing sources!

The old firefox debugger used to show a message in this case:
<img width="371" alt="image" src="https://cloud.githubusercontent.com/assets/1141550/23451798/c942ec46-fe61-11e6-8a3f-c57dd6ddff9d.png">

wldcordeiro:
Hmm how are there now sources at all? Like shouldn't we have at minimum the index file? Even if it has no scripts in it. I swear that's the behavior Chrome and the old debugger have.

juliandescottes:
I agree it's weird, I didn't look into more details. The test url is https://devtools-html.github.io/debugger-examples/

juliandescottes:
So it looks like a regression. 

I don't have the issue when using the current Nightly, but I do when trying out the new bundle currently attached to https://bugzilla.mozilla.org/show_bug.cgi?id=1338567

I also have the issue when testing the project from Github. So probably introduced in the last month or so.

jasonLaster:
Agreed. 

One style point, it would be nice to update the other "Not Paused" styles to be the softer color like this one

[no-sources.pdf](https://shipusercontent.com/60b45d6af7069341d7bfe464c49a9e97/no-sources.pdf)

juliandescottes:
I investigated a bit more and this is actually a very weird bug.

> I don't have the issue when using the current Nightly, but I do when trying out the new bundle currently attached to https://bugzilla.mozilla.org/show_bug.cgi?id=1338567

But it's not a regression of the new bundle, the difference comes from the addons I have installed in Nightly.

![debugger_react_devtools_bug](https://cloud.githubusercontent.com/assets/1141550/23459223/b1c70ab6-fe7f-11e6-8a95-40d3f87efd77.gif)

It looks like having installed the "React devtools" addon on the debuggee allows the debugger to see the index of the page!?

jasonLaster:
hmm, this must be a server bug then if the debugger is not getting the newSource events.

jasonLaster:
[...]

PR #2385, opened 2017-03-15
Show message when there are no sources for a page (no files to debug)

ktamir:
Associated Issues: #2371 #2231 

### Summary of Changes
When a page doesn't have sources, show an indicative message "This page has no sources". Right now nothing is shown.

### Test Plan
- Go to https://devtools-html.github.io/debugger-examples/
- Attach debugger to that tab
- Check if there is a message "This page has no sources."
- Go to a page with sources
- Check that the tree loads normally

debugger-html-no-sources-message

### DIFF
--- a/src/components/SourcesTree.js
+++ b/src/components/SourcesTree.js
@@ -240,7 +240,7 @@ class SourcesTree extends Component {
 
     const noSourcesMessage = dom.div({
       className: "no-sources-message"
-    }, "This page has no sources.");
+    }, L10N.getStr("sources.noSourcesAvailable"));
 
     if (isEmpty) {
       return noSourcesMessage;
--- a/src/components/SourcesTree.js
+++ b/src/components/SourcesTree.js
@@ -239,12 +239,12 @@ class SourcesTree extends Component {
     });
 
     const noSourcesMessage = dom.div({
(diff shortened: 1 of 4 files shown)


src/components/SourcesTree.js
```diff
@@ -238,6 +238,13 @@ class SourcesTree extends Component {
       renderItem: this.renderItem
     });
 
+    const noSourcesMessage = dom.div({
+      className: "no-sources-message"
+    }, "This page has no sources.");
```
jasonLaster: We should translate this with `L10N`

wldcordeiro:
Correct me if I'm wrong but shouldn't we always have one source? The document itself? Or should we only show the document if we detect script tags in it? I'm confused. It seems like the bug is that in certain cases we aren't including the document as a source.

ktamir:
@jasonLaster I committed the fix :)

jasonLaster:
Thanks @ktamir. Unfortunately, we still need to update the properties file as well [l10n](https://github.com/devtools-html/debugger.html/blob/master/docs/local-development.md#l10n)

jasonLaster:
Thanks Tamir!

Have you joined our slack? https://devtools-html-slack.herokuapp.com

PR #2386, opened 2017-03-15
Css fixes

yuliaSharabi:
Remove vendor prefixes: #<2372>

### Summary of Changes

* Removed vendor specific prefixes in css files- for example webkit
* Add new line to file Accordion.css.(missed by accident during the commit)

### Test Plan

Sanity test:
Checked that UI is loading successfully on the Todo examples.

Example test plan:

Feel free to improve it!

### DIFF
--- a/src/components/shared/Accordion.css
+++ b/src/components/shared/Accordion.css
@@ -59,7 +59,3 @@
   width: 16px;
   height: 16px;
 }
-
-.accordion .header-buttons button::-moz-focus-inner {
-  border: none;
-}
--- a/src/components/SecondaryPanes/Expressions.css
+++ b/src/components/SecondaryPanes/Expressions.css
@@ -9,12 +9,6 @@
   color: var(--theme-body-color);
 }
 
-.input-expression::-webkit-input-placeholder {
-  text-align: center;
-  font-style: italic;
-  color: var(--theme-comment-alt);
-}
-
 .input-expression::placeholder {
   text-align: center;
   font-style: italic;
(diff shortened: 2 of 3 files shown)


jasonLaster:
Thanks @yuliaSharabi. I'll test tomorrow :)

src/components/shared/Accordion.css
```diff
@@ -59,7 +59,3 @@
   width: 16px;
   height: 16px;
 }
-
-.accordion .header-buttons button::-moz-focus-inner {
```
irfanhudda: I'm not sure if this should be deleted. `autoprefixer` cannot add this. Any ideas?

irfanhudda:
Curious about why or how `assets/module-manifest.json` was edited here.

jasonLaster:
I think `module-manifest` is added when you run `publish-assets` we're adding an option to only update the file when we're doing a release

jasonLaster:
Thanks @yuliaSharabi!

yuliaSharabi:
So just a quick question does autoprefixer cannot add the moz prefixes in the src/components/shared/Accordion.css file?

jasonLaster:
Hi, I'm not sure about the moz prefixes. Could you look at the outputted
CSS? It's in a style tag on the page

irfanhudda:
@yuliaSharabi from what I understand, `autoprefixer` can add most vendor prefix which have equivalent in other browser like `-webkit-transition` and `-moz-transition` but `-moz-focus-inner` doesn't appear to have equivalent in other browsers and is firefox specific.

Issue #2387, opened 2017-03-15
Missing navigation issue in search bar


nadavoran:
Today the only way to navigate is to use the keyboard, I think that adding Next Perv buttons will help the usability 
#goodnessSquad

PR #2388, opened 2017-03-15
Adding navigation button to Search Input. re #2387

nadavoran:
Associated Issue: #2387


jasonLaster:
Thanks @nadavoran. Please share some screenshots and necessary next steps.

The code overall looks good :)

nadavoran:
Hi sorry for the incomplete comments.

Navigation buttons in the search bar:
image

Basic behavior: 
- Appears only when there are more than 1 search result.
- Loop through the result (the only indication for first or last record is in the text next to the buttons) 

Created 2 new SVG's for the Arrows, so you might want to use your own svg library.

There is one thing that need to update about this feature is the indication of user selection. I tried to give a nice click feedback by increasing the Fill color (I used the --theme-comment, but there isn't darker theme color).
image

There is an enhancement that might be cool, that when you use the keyboard shortcut the matching button will be activated, what do you think?
Another thing the buttons jumps a little base on the summary result, might be better to use absolute position. 

There is one comment, it looks like the close button is disabled, it uses the --theme-comment-alt maybe this should also be changed.

src/components/Editor/SearchBar.css
```diff
@@ -120,3 +120,7 @@
 .search-bar .result-list {
   max-height: 230px;
 }
+.search-bar .nav-btn-search-prev{
```
wldcordeiro: Linter complains about this line.

src/components/Editor/SearchBar.js
```diff
@@ -426,6 +426,26 @@ const SearchBar = React.createClass({
     }
   },
 
+  createNevigatorData(){
```
wldcordeiro: Linter doesn't like this line either.

src/components/shared/SearchInput.js
```diff
@@ -39,6 +40,15 @@ const SearchInput = React.createClass({
     return Svg("magnifying-glass");
   },
 
+  //renderNavNext(){
```
wldcordeiro: This commented out code should be removed.

src/components/shared/SearchInput.js
```diff
@@ -39,6 +40,15 @@ const SearchInput = React.createClass({
...
+  //  return Svg("arrow-down");
+  //},
+
+  renderNav(){
```
wldcordeiro: Same thing? I think the SFC you  created does the job of this function.

src/components/shared/SearchInput.js
```diff
@@ -72,6 +83,8 @@ const SearchInput = React.createClass({
         value: query,
         spellCheck: false,
       }),
+      (navigatorData && NavigatorButtons(navigatorData)),
+      //this.renderNav(),
```
wldcordeiro: Remove commented line.

src/components/Editor/Navigator.js
```diff
@@ -0,0 +1,38 @@
...
+  nextTooltip?: string,
+  prevTooltip?: string
+};
+
+function NavigatorButtons({ handleNext, handlePrev, nextDisabled, prevDisabled, nextClass, prevClass, nextTooltip, prevTooltip }: NavigatorButtonsType) {
+  return dom.div({
+    className: "nav-buttons"
+  },
+    Svg("arrow-down", {
+      className: "nav "+(nextClass ? `nav-btn-${nextClass}` : "next-btn"),
```
wldcordeiro: I think you could use the `classnames` utility and probably avoid having props for the class names. They could be reused.

wldcordeiro:
> Another thing the buttons jumps a little base on the summary result, might be better to use absolute position.

I think the containing element of the button needs to have `flex-shrink: 0;` to avoid resizing with the summary changes.

jasonLaster:
I think it might be nice to try the iterating buttons next to the close button.

screen shot 2017-03-17 at 1 56 55 pm

wldcordeiro:
Reopening on a separate branch as this was done on master. 🙁

Issue #2362, opened 2017-03-14
Use prettier


bomsy:
Adding prettier as a linting step...
https://github.com/prettier/prettier

PR #2389, opened 2017-03-16
Add prettier

bomsy:
Associated Issue: #2362

### Summary of Changes

* Added  `yarn prettier` for manually running prettier of all the files
* Added `yarn eslint-check` to see the eslint rules conflicting with prettier
* Disabled all the eslint rules conflicting with prettier.
* Also running prettier through eslint (i.e `yarn lint` lints based on prettier rules as well as eslint)
* Updated js files using prettier, enforcing `-trailing-comma=all` and `--bracket-spacing=true` (can be updated)
* Prettier automatically formats staged files pre commit.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -4,7 +4,8 @@
     "react",
     "mozilla",
     "flowtype",
-    "babel"
+    "babel",
+    "prettier"
   ],
   "globals": {
     "atob": true,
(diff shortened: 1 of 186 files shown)


bomsy:
@jlast would we want to go with `--trailing-comma=es5` instead? Seems like the comma's everywhere is too much. What do you think?

@clarkbw @wldcordeiro @jbhoosreddy

Issue #2323, opened 2017-03-10
[FunctionSearch] name functions based on their var


jasonLaster:
We currently treat this case `obj.me = function() {}` `obj.me = () => {}` as anonymous functions. In these cases we should use the assignment to find the name `me`.

Here's a failing test to help someone get started

```diff
diff --git a/src/utils/tests/parser.js b/src/utils/tests/parser.js
index 6c0cbba..84d5194 100644
--- a/src/utils/tests/parser.js
+++ b/src/utils/tests/parser.js
@@ -82,6 +82,8 @@ const SOURCES = {
       }
     };

+    Obj.property = () => {}
+
     class Ultra {
       constructor() {
         this.awesome = true;
@@ -165,6 +167,7 @@ describe("parser", () => {
         "incrementCounter",
         "sum",
         "doThing",
+        "property",
         "doOtherThing",
         "constructor",
         "beAwesome"
@@ -174,6 +177,7 @@ describe("parser", () => {
         "count",
         "counter",
         "sum",
+        "property",
         "a",
         "b",
         "Obj",
```

### screenshot
<img src='https://shipusercontent.com/b4ab44e43a6e602a2387080e0187fdc7/Screen%20Shot%202017-03-10%20at%206.32.53%20AM.png' title='Screen Shot 2017-03-10 at 6.32.53 AM.png' width=852>

wldcordeiro:
I had considered adding this and wasn't sure whether to consider these functions or variables when I was adding the variable search capability. Is it okay to have them appear in both result lists or should we exclude from variables?

jasonLaster:
Showing up in both places is fine although in this case I don't know if we should show all the properties in variable search. It's something to talk about in another issue

ruturajv:
Not sure if I've understood this clearly, 
- but by typing "a" in search, we shouldn't be showing any output as "anonymous function" which are assigned to members
- "property" if searched should list / reference to that anonymous fn

wldcordeiro:
@ruturajv We should still have functions output as anonymous. The following situations should be anonymous

```js
(function() {
  console.log("IIFE");
})();
```

```js
const things = [1, 2, 3];
const thingsSqrd = things.map(function(i) { return i * i }); // if this were an arrow function it'd still be anonymous
```

Regarding properties, don't worry, that's a different discussion and already handled in our function searching logic. 😄 

Hope that clarifies.

ruturajv:
[...]

PR #2404, opened 2017-03-17
[FunctionSearch] name functions based on their var

ruturajv:
Associated Issue: #2323 

* Suggested fix for function definitions for object members

Its a WIP patch :) please bear.


src/utils/parser/utils.js
```diff
@@ -102,6 +102,14 @@ function getFunctionName(path) {
...
   }
 
+  if (parent.right && parent.right.type == "ArrowFunctionExpression") {
+    return parent.left.property.name;
+  }
+
+  if (parent.right && parent.right.type == "FunctionExpression") {
+    return parent.left.property.name;
+  }
+
```
jasonLaster: 💃

src/utils/tests/parser.js
```diff
@@ -195,6 +199,7 @@ describe("parser", () => {
         "count",
         "counter",
         "sum",
+        // "property",
```
wldcordeiro: Remove the commented line?

src/utils/parser/utils.js
```diff
@@ -102,6 +102,14 @@ function getFunctionName(path) {
     return parent.id.name;
   }
 
+  if (parent.right && parent.right.type == "ArrowFunctionExpression") {
```
wldcordeiro: You might be able to avoid the two conditions by using the `isFunction` helper I wrote.
```js
if (parent.right && isFunction(parent.right) {
  return parent.left.property.name;
}
```

wldcordeiro:
@ruturajv a couple comments, what do you mean by "pending variable searches"?

ruturajv:
@wldcordeiro - The original issue also wanted the object's members to be listed in search for "variables" ie for `Obj.member = "abc"`, even `member` should be listed in as a variable. That is pending.

wldcordeiro:
> The original issue also wanted the object's members to be listed in search for "variables" ie for Obj.member = "abc", even member should be listed in as a variable. That is pending.

@ruturajv 

They already are though. There was an unresolved question about whether an object member should appear in both variables and functions.

See this test.

https://github.com/devtools-html/debugger.html/blob/master/src/utils/tests/parser.js#L112
https://github.com/devtools-html/debugger.html/blob/master/src/utils/tests/parser.js#L217

src/utils/parser/utils.js
```diff
@@ -99,14 +99,23 @@ function getFunctionName(path) {
     return parent.id.name;
   }
 
+  if (parent.right && isFunction(parent.right)) {
+    return parent.left.property.name;
+  }
+
+  if (parent.right && isFunction(parent.right)) {
```
wldcordeiro: This `if` condition is the same as the one above, it can be removed, otherwise this looks great.

ruturajv:
thanks @wldcordeiro

PR #2408, opened 2017-03-18
accommodate objects without a prototype (fixes #2407)

### LINKED ISSUE #2407: fromJS throws on object without constructor
Over in https://github.com/mozilla/qbrt/issues/19, @benfrancis noticed that debugger.html fails to stop at a breakpoint when loaded in qbrt (a Mozilla app that loads itself using Firefox's --app flag and attaches devtools to a window rather than a tab), reporting `Error in console "TypeError: value.constructor is undefined debugger.js:31325:1"`.

It looks like the code in question is in *fromJS*, which checks `value && value.constructor.meta` without ensuring `value.constructor`. In this case, the value is an object without a constructor (presumably created via `Object.create(null)` or the like), so it doesn't have a *constructor* property in its prototype chain (nor any other property that an object otherwise inherits from the Object prototype).

Here's an example of such an object JSON.stringified:

```json
{
  "0": {
    "configurable": true,
    "enumerable": true,
    "writable": true,
    "value": {
      "type": "object",
      "actor": "server1.conn0.pausedobj43",
      "class": "MouseEvent",
      "extensible": true,
      "frozen": false,
      "sealed": false,
      "ownPropertyLength": 1,
      "preview": {
        "kind": "DOMEvent",
        "type": "click",
        "properties": {
          "buttons": 0,
          "clientX": 143,
          "clientY": 300,
          "layerX": 143,
          "layerY": 300
        }
      }
    }
  }
}
```

JSON.stringify doesn't show this, since it only stringifies own properties, but the inner object, i.e. the value of the "0" property, does have a *\_\_proto\_\_* property (which has a *constructor* property), as shown in this screenshot:

screen shot 2017-03-17 at 17 17 32

Perhaps there's an underlying bug here that causes some code to call fromJS with the outer object rather than the inner one (unsure, as I'm unfamiliar with the code in question). In any case, fromJS should handle such objects rather than throwing.

mykmelez:
Associated Issue: #2407

### Summary of Changes

* ensure *value.constructor* before *dereferencing value.constructor.meta*
* ensure *value.hasOwnProperty* before calling it

### Test Plan

- [x] In qbrt, I set a breakpoint in an application, triggered it, and confirmed that debugger.html broke there.
- [x] In debugger.html, I wrote a unit test, confirmed that it failed without the changes, and then confirmed that it passed with the changes.

### DIFF
--- a/src/utils/fromJS.js
+++ b/src/utils/fromJS.js
@@ -16,7 +16,7 @@ const Immutable = require("immutable");
   length confuses Immutable's internal algorithm.
 */
 function createMap(value) {
-  const hasLength = value.hasOwnProperty("length");
+  const hasLength = value.hasOwnProperty && value.hasOwnProperty("length");
   const length = value.length;
 
   if (hasLength) {
@@ -50,7 +50,7 @@ function fromJS(value: any): any {
   if (Array.isArray(value)) {
     return createList(value);
   }
-  if (value && value.constructor.meta) {
+  if (value && value.constructor && value.constructor.meta) {
     // This adds support for tcomb objects which are native JS objects
     // but are not "plain", so the above checks fail. Since they
     // behave the same we can use the same constructors, but we need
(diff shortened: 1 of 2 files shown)


mykmelez:
Yes, I tested locally by making the change to a copy of devtools in a Firefox installation that I'm using with qbrt, which is experiencing this bug (https://github.com/mozilla/qbrt/issues/19).

Issue #2053, opened 2017-02-20
Tree should support blur events


roieu:
Tree of elements, like Sources tree in the debugger, should be able to support a blur event to ensure we keep a selected node highlighted but ignoring any arrow navigation (mute onKeyDown/Press/Up events).
This relates to another issue, to support arrow navigation in the scope panel, which is currently impossible because Sources tree takes over any arrow keys navigation.

jasonLaster:
Hi @roieu. I believe what you're saying is that when the source tree is blurred, it should be clear which source is selected. Is that correct?

roieu:
@jasonLaster that is correct.
It also means that when the tree is not in focus it should stop respond to keyboard events.
Then we'll be able to have arrow keys navigation in Call Stack frame, which now no matter which
frame is in focus, the keyboard events are intercepted by the source tree control.
I am playing around with the tree in devtools-core to handle blur events (I am new to this project so my pace is slow)

jasonLaster:
Great. Checkout how we do focus events in the call stack and blur in the
watch expression input

jasonLaster:
@roieu any updates?

roieu:
@jasonLaster Haven't got a chance to dive deep into it last week.
Hope to share something later this week.

jasonLaster:
sounds good.

By the way, just noticed that the result list ui doesn't do much with blur events either...

roieu:
@jasonLaster I have added blur support to the tree component and also added blur item style to the SourceTree.
Creating a pull request....

jasonLaster:
Hey @roieu, how's it going?

It seems like this work might be blocked on a notion of "blurred". Feel free to hop in slack and chat about it.

jasonLaster:
Some notes from the PR

adding blurred makes me concerned that we could be making things more difficult for ourselves in the future:  

* blurred could be a state for many elements
* blurred/focused might not be weird in some cases - an element is not focused, but it is blurred

I would like to better understand how atom does it. I started looking at their HTML:
One thing I do like is that it uses `::before` for the background which is nice

![](http://recordit.co/LzeyoWtBkG.gif)


also, here's the code: https://github.com/atom/tree-view

jasonLaster:
closing this PR as we discussed a better way last week:

We should dull the selected item when the tree is not focused!

PR #2411, opened 2017-03-20
Tree blur

roieu:
Associated Issue: #2053 

### Summary of Changes

* Added Blurred style for Source node in SourcesTree
* Adjusted render parameters for ManagedTree sub components

### Test Plan

Tested manually using SourcesTree component from devtools-html/debugger.html

Example test plan:

- Focus in and out SourcesTree from various use-cases

Feel free to improve it!

tree-blur

### DIFF
--- a/src/components/SecondaryPanes/ChromeScopes.js
+++ b/src/components/SecondaryPanes/ChromeScopes.js
@@ -79,7 +79,7 @@ class Scopes extends Component {
     return nodes;
   }
 
-  renderItem(item, depth, focused, _, expanded, { setExpanded }) {
+  renderItem(item, depth, focused, _, expanded, blurred, { setExpanded }) {
     const notEnumberable = false;
     const objectValue = "";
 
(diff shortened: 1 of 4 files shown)


jasonLaster:
Thanks! I'm traveling today and will be able to review when I'm back on Wednesday

roieu:
@jasonLaster have a safe trip back. 
Note that there's another PR on devtools-core [https://github.com/devtools-html/devtools-core/pull/259](url) that needs to be pulled in order to support the enhancement.
Thanks!
Roie.

jasonLaster:
adding blurred makes me concerned that we could be making things more difficult for ourselves in the future:  

* blurred could be a state for many elements
* blurred/focused might not be weird in some cases - an element is not focused, but it is blurred

I would like to better understand how atom does it. I started looking at their HTML:
One thing I do like is that it uses `::before` for the background which is nice

jasonLaster:
closing this now as it's gone stale, but we can revisit it at any point

Issue #2319, opened 2017-03-10
migrate to jest


jasonLaster:
We'd like to migrate to jest, which is a more modern unit testing library with better support for mocking and snapshot testing

Steps:

- [x] get tests running #2315
- [x] Update CI integration
- [x] Update npm scripts
- [x] Remove Mocha code
- [x] update docs
- [x] Code coverage

#### coverage

`--coverage` creates a nice report. We also have lots of coverage output for codecov to ingest.

montogeek:
I will continue with the todo list today

PR #2417, opened 2017-03-21
Jest coverage

clarkbw:
Associated Issue: #2319

### Summary of Changes

* add `test-coverage` script
* add coverage reporting to circleci and upload to codecov

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
     "lint-md": "remark -q *.md src/**/*.md configs/*.md docs/**/*.md",
     "lint-fix": "yarn lint-js -- --fix",
     "test": "jest --forceExit",
+    "test-coverage": "yarn test -- --coverage",
     "test-all": "yarn test; yarn lint",
     "firefox": "start-firefox --start --location https://devtools-html.github.io/debugger-examples/",
     "chrome": "start-chrome --location https://devtools-html.github.io/debugger-examples/",
--- a/src/components/shared/Accordion.css
+++ b/src/components/shared/Accordion.css
@@ -60,6 +60,6 @@
   height: 16px;
 }
 
-accordion .header-buttons button::-moz-focus-inner {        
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -27,6 +27,7 @@
     "lint-md": "remark -q *.md src/**/*.md configs/*.md docs/**/*.md",
     "lint-fix": "yarn lint-js -- --fix",
     "test": "jest --forceExit",
+    "test-coverage": "yarn test -- --coverage",
```
clarkbw: I wasn't sure if we wanted add coverage to the default test script instead of a creating a new one.

jasonLaster: sure - this is probably fine. we can always update this later when we're using flow as well

jasonLaster:
Should we see codecov here?

clarkbw:
> Should we see codecov here?

Yes, but I turned it off previously since we weren't getting flow coverage.  Just turned it back on now.  Maybe if I close and reopen this it'll happen...

clarkbw:
Adding a fix for the lint issue that I believe came from #2420

jasonLaster:
ahh - thanks

src/components/shared/Accordion.css
```diff
@@ -60,6 +60,6 @@
   height: 16px;
 }
 
-accordion .header-buttons button::-moz-focus-inner {        
+.accordion .header-buttons button::-moz-focus-inner {
```
clarkbw: I also added a `.` here which is what I believe is needed and was just missed from copy / pasta

clarkbw:
Coverage is now being reported in CI.

screen shot 2017-03-23 at 10 42 15 am

And I'm seeing the JSON artifact: https://400-55521579-gh.circle-artifacts.com/0/tmp/circle-junit.y1G4yTz/jest-coverage/coverage-final.json

montogeek:
Cool!

Issue #2419, opened 2017-03-22
[Accordion] re-add moz-focus-inner class


jasonLaster:
In https://github.com/devtools-html/debugger.html/pull/2386 we accidentally removed a style that's not autoprefixed for button focus states

```css
accordion .header-buttons button::-moz-focus-inner {		
  border: none;
}
```

PR #2420, opened 2017-03-22
Re-add accordion focus inner style

jasonLaster:
In https://github.com/devtools-html/debugger.html/pull/2386 we accidentally removed a style that's not autoprefixed for button focus states

Associated Issue: #2419


clarkbw:
Looks like this change has some CSS lint errors

PR #2428, opened 2017-03-23
Refactor editor

jasonLaster:
### Summary of Changes

This is very similar to some of the work that @wldcordeiro did last month to cleanup the editor component. Ideally nothing should change, but the component will be easier to reason about. The focus of the work here is to simplify `componentDidMount`

* introduced a new method `setupEditor`
* moved some anonymous event handlers to the prototype
* created a  couple of utilities

### DIFF
--- a/src/components/Editor/index.js
+++ b/src/components/Editor/index.js
@@ -39,9 +39,9 @@ const HitMarker = React.createFactory(require("./HitMarker"));
 const {
   getDocument,
   setDocument,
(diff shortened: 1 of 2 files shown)


jasonLaster:
Here's a comparison, now somewhat outdated:

componentdidmount

wldcordeiro:
Awesome! I'll give this some further testing but seeing this makes me really happy! 🙌

Issue #2424, opened 2017-03-22
[SearchBar] Search bar when scrolling with keyboard need to be fixed


irfanhudda:
Trying the steps from #2370 for Sensor Web example

On scrolling using keyboard `SearchBar` covers tabs. Also bottom bar is not visible

STR:
1. Open a big file (`main.js` from Sensor Web example [http://aws-sensorweb-static-site.s3-website-us-west-2.amazonaws.com/](http://aws-sensorweb-static-site.s3-website-us-west-2.amazonaws.com/))
2. Open search bar (`Ctrl+f`)
3. Focus on editor.
4. Try to scroll down using keyboard arrow key

![bug](https://cloud.githubusercontent.com/assets/1755089/24204405/5bd61352-0f3e-11e7-9850-9629c91f247d.gif)

Interestingly, scrolling with mouse seems to work well. Also I was not able to reproduce this issue in other examples

samjwalker:
Found separate bugs on windows 10:

1) The search bar masking the tabs as you pointed out.
2) When the search bar is opened it is pushing everything down the screen putting the bottom scroll bar off the screen and the bottom scroll arrow off also.

Not sure if this is a bug but I'm not getting the "search for" bar underneath the main search bar.

Should we raise separate bugs for these?

jasonLaster:
Hmm, is the second one https://github.com/devtools-html/debugger.html/issues/2291? If not, then file a new issue :)

jasonLaster:
One thought that might not help is there is a CSS sticky property we might be able to use. We used it for the command bar effectively :)

samjwalker:
Looking at it today, both of the bugs are reproducable in nightly but not our local debugger.html

However there is a funny gap at the bottom of the screen and we'll raise an issue for that.

Also still no sign of the other bar underneath the main search bar and will look into that.

In summary I can not reproduce any of the main issues.

jasonLaster:
Hmm, that's interesting @samjwalker. I just tried @irfanhudda STR in both chrome and firefox with debugger.html and was able to reproduce. try clicking in the document and then holding down the arrow key.

irfanhudda:
Also, note that I could reproduce this on Sensor Web example [http://aws-sensorweb-static-site.s3-website-us-west-2.amazonaws.com/](http://aws-sensorweb-static-site.s3-website-us-west-2.amazonaws.com/). Other examples were working well so it might be because the file is very large.

samjwalker:
[...]

Issue #2429, opened 2017-03-23
Blank space at the end of editor component.


samjwalker:
**When the search bar is opened there is as gap present (as highlited in the picture) and the 2nd search bar is not appearing.**

**STR:**

1) Open a file in the debugger

2) Open the search bar `ctrl+f`

**Expected result:**
Both of the search bars are open and no gaps anywhere.

**Actual result:**
Only the main search bar is opened and there is a gap above the `.source-footer`

This gap is part of the `.editor-wrapper1` and so must be to do with some incorrect calculations (when the editor is opened the size of the `.editor-mount` is 100% - 72px but should be 100% - 45px  and when done the gap is appering in the right place.

Also the second search bar will only open when `searchModifiers` and `symbolSearch` are true and only `searchModifiers` is true.
![gap present where scroll bar should be](https://cloud.githubusercontent.com/assets/15427172/24250133/be45c868-0fcd-11e7-8240-9c1eba6f804e.PNG)

jasonLaster:
Nice find!

PR #2441, opened 2017-03-24
Don't push bottom scroll bar off screen when search is opened for files that don't need to be prettyifyed

samjwalker:
Associated Issue: #2424 

* Fixes Tabs from being masked by the search bar
* Stops the bottom scroll bar being pushed off by search bar

### DIFF
--- a/src/components/Editor/index.js
+++ b/src/components/Editor/index.js
@@ -18,6 +18,7 @@ import { debugGlobal } from "devtools-launchpad";
 import { isEnabled } from "devtools-config";
 import {
   getSourceText,
+  getFileSearchState,
   getBreakpointsForSource,
   getSelectedLocation,
   getSelectedFrame,
@@ -55,6 +56,12 @@ import {
 import { getVisibleVariablesFromScope } from "../../utils/scopes";
 import { isFirefox } from "devtools-config";
 import "./Editor.css";
+
+const cssVars = {
+  searchbarHeight: "var(--editor-searchbar-height)",
+  secondSearchbarHeight: "var(--editor-second-searchbar-height)",
+  footerHeight: "var(--editor-footer-height)"
+};
 
 type EditorState = {
   searchResults: {
@@ -646,14 +653,31 @@ class Editor extends Component {
       }));
   }
 
-  editorHeight() {
-    const { selectedSource, horizontal } = this.props;
+  getInlineEditorStyles() {
+    const { selectedSource, horizontal, searchOn } = this.props;
 
-    if (!shouldShowFooter(selectedSource, horizontal)) {
-      return "100%";
+    let subtractions = [];
+
+    if (shouldShowFooter(selectedSource, horizontal)) {
+      subtractions.push(cssVars.footerHeight);
     }
 
-    return "";
+    if (searchOn) {
+      subtractions.push(cssVars.searchbarHeight);
+
+      const secondSearchBarOn = isEnabled("searchModifiers") &&
+        isEnabled("symbolSearch");
+
+      if (secondSearchBarOn) {
+        subtractions.push(cssVars.secondSearchbarHeight);
+      }
+    }
+
+    return {
+      height: subtractions.length === 0
+        ? "100%"
+        : `calc(100% - ${subtractions.join(" - ")})`
+    };
   }
 
   renderPreview() {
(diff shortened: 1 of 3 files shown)


jasonLaster:
Thanks @samjwalker!

Could you attach some before / after screenshots

jasonLaster:
@samjwalker just updated the PR to fix the lint issue. Going to test now

src/components/Editor/index.js
```diff
@@ -568,14 +570,22 @@ const Editor = React.createClass({
...
+        return { height: "calc(100% - 54px)" };
+      }
+
+      return { height: "100%" };
     }
 
-    return "";
+    // The height is already set and so we dom't want to pass any values here.
+    // Set in editor.css (see .search-bar ~ .editor-mount)
+    return {};
```
jasonLaster: i think `.search-bar ~ .editor-mount` might be more confusing than it's worth now that we're doing inline styles. 

How about we move that value in here?

src/components/Editor/index.js
```diff
@@ -568,14 +570,22 @@ const Editor = React.createClass({
...
   },
 
-  editorHeight() {
-    const { selectedSource, horizontal } = this.props;
+  getInlineEditorStyles() {
+    const { selectedSource, horizontal, searchOn } = this.props;
 
+    // The footer is off if there is no 'prettify' button (which happens for
+    // source-mapped or already prettified files)
     if (!shouldShowFooter(selectedSource, horizontal)) {
```
jasonLaster: perhaps we can now do

```js
if (shouldShowFooter(selectedSource, horizontal)) {
}

...
```

src/components/Editor/index.js
```diff
@@ -568,14 +570,22 @@ const Editor = React.createClass({
...
-      return "100%";
+      if (searchOn) {
+        return { height: "calc(100% - 54px)" };
```
jasonLaster: So... It's not great that we have a *magic* number here because it will definitely change.

I don't love suggesting this, but it might be helpful to calculate the heights of the footer and search bar...

Here's where we get calculated DOM dimensions elsewhere: 

https://github.com/devtools-html/debugger.html/search?utf8=%E2%9C%93&q=getBoundingClientRect&type=

src/components/Editor/index.js
samjwalker: Yes, by doing that we could also fix the bug with the blank box. `#2429 `

jasonLaster:
Hey @jbhoosreddy @jacobjzhang mind taking a look?

src/components/Editor/index.js
```diff
@@ -646,14 +647,32 @@ class Editor extends Component {
...
 
-    if (!shouldShowFooter(selectedSource, horizontal)) {
-      return "100%";
+    if (shouldShowFooter(selectedSource, horizontal)) {
+      // The footer is on and so we need to return different values due to the
+      // size of the footer to make sure we have nothing being pushed off the
+      // screen.
+      if (searchOn) {
+        return {
+          height: "calc(100% - var(--editor-searchbar-height)" +
```
jasonLaster: I think we can do something like

```js
const cssVars = {
searchbarHeight: "var(--editor-searchbar-height)"
}
```

at the top of the file and below

```js
height: ` calc(100% - ${searchbarHeight} - ${} - ${}`
```

jbhoosreddy: It is so much cleaner than the other ways we explored. I'd also recommend making that change as Jason mentioned. Other than that looks good to me.

jacobjzhang: This is 100% cleaner. Looks great to me.

samjwalker:
Got creative:

```javascript
let subtractions = [];

if (shouldShowFooter(selectedSource, horizontal)) {
  subtractions.push(cssVars.footerHeight);
}

if (searchOn) {
  subtractions.push(cssVars.searchbarHeight, cssVars.secondSearchbarHeight);
}

return {
  height: subtractions.length === 0
    ? "100%"
    : `calc(100% - ${subtractions.join(" - ")})`
};
```

jbhoosreddy:
@samjwalker I really like how this looks now. However, theres one last thing I think we'll need to fix. The search bar has two states. i.e. with and without symbol search. I'm assuming your local config has symbol search turned on. that's why it looks good with symbol search, however, without symbol search, the height calculations still need some love.

Let's make that change and merge. 👍 

Screenshots attached.

### This looks fine.
screen shot 2017-04-11 at 7 08 07 pm

### This doesn't
screen shot 2017-04-11 at 7 08 30 pm

samjwalker:
@jbhoosreddy does this work for you?

PR #2445, opened 2017-03-24
Fix duplicate locale

jasonLaster:
Comes from here:
https://bugzilla.mozilla.org/show_bug.cgi?id=1346698


jasonLaster:
@flodolo mind looking at this quickly?

Issue #2451, opened 2017-03-25
[FunctionSearch] support assignments


jasonLaster:
The parser currently fails when trying to determine the function name in this case `child = function() {}`.

This gets at a larger question of how we want to handle assignments:

```js
child = function() {}
let child2 = function() {}
```

I think both of these cases should show up in the list of variables and the list of functions. The case for variables is clear given that `child` and `child2` are both variables. The case for showing `child` and `child2` in the list of functions is less clear, except for the fact that the user might be thinking of these as functions they could want to find. 

Below is a patch with the test case and a proposed fix. I left out the assertions that will need to be updated

```
diff --git a/src/utils/parser/utils.js b/src/utils/parser/utils.js
index dbab8b1..feeb1ba 100644
--- a/src/utils/parser/utils.js
+++ b/src/utils/parser/utils.js
@@ -99,6 +99,11 @@ function getFunctionName(path) {
     return parent.id.name;
   }

+  if (parent.type == "AssignmentExpression") {
+    return parent.left.name;
+  }
+
   if (parent.right && isFunction(parent.right)) {
     return parent.left.property.name;
   }
diff --git a/src/utils/tests/parser.js b/src/utils/tests/parser.js
index 3391a34..c2e0faf 100644
--- a/src/utils/tests/parser.js
+++ b/src/utils/tests/parser.js
@@ -19,6 +19,8 @@ const SOURCES = {
     function square(n) {
       return n * n;
     }
+
+    child = function() {}
   `
   ),
   math: formatCode(
@@ -70,7 +72,9 @@ const SOURCES = {
     let bar = 2;
     const baz = 3;
     const a = 4, b = 5;
-  `
+    var child = function() {};
+    child2 = function() {};
```

PR #2452, opened 2017-03-25
Fixing Bug #2451 - AssignmentExpression Functions in search

ruturajv:
Associated Issue: #2451

### Summary of Changes

- added assignementExpression code
- added and modified test cases

### DIFF
--- a/src/utils/parser/utils.js
+++ b/src/utils/parser/utils.js
@@ -99,14 +99,12 @@ function getFunctionName(path) {
     return parent.id.name;
   }
 
-  if (parent.type == "AssignmentExpression" && isFunction(parent.right)) {
-    if (parent.left.name) {
-      // child = function() {}
-      return parent.left.name;
-    } else if (parent.left.type == "MemberExpression") {
-      // Obj.property = () => {}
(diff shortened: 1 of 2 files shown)


src/utils/tests/parser.js
```diff
@@ -19,6 +19,10 @@ const SOURCES = {
     function square(n) {
       return n * n;
     }
+
+    child = function() {};
+
+    (function () { 2 })();
```
jasonLaster: I added this test to get our coverage up to 100%

src/utils/parser/utils.js
```diff
@@ -99,8 +99,12 @@ function getFunctionName(path) {
...
   }
 
-  if (parent.right && isFunction(parent.right)) {
-    return parent.left.property.name;
+  if (parent.type == "AssignmentExpression") {
+    if (parent.left.type == "MemberExpression") {
+      return parent.left.property.name;
+    }
+
+    return parent.left.name;
```
jasonLaster: I tweaked this block to:

* remove `isFunction(parent.right)` which i believe is unnecessary
* prioritize the `MemberExpression` which is consistent with the other type checks first

PR #2421, opened 2017-03-22
(WIP) Preview object properties

jasonLaster:
Associated Issue: #2219

### Summary of Changes

* This adds an expression when we hover on a property, we still need to hide it in watch expressions
* There are perf implications here and other things :)

PR #2453, opened 2017-03-25
Hiding Expressions

bomsy:
Associated Issue: #2219

### Summary of Changes

* As a follow up from #2421, including changes.
* Refactor of the reducers to support separate states for `hidden` and `visible` expressions.
* Persisting only `visible` expressions.
* Adding the `visible` argument to expression actions. e.g `addExpression(value, visible)` . If the visible argument is not passed, it defaults to `true`, creating a visible expression. Passing in `false` creates a hidden expression.

### Test Plan

- [x] add and remove expressions from the watch expressions list.
- [x] break on exceptions and preview member expressions
- [x] not that they don't get added to the watch expressions list

### DIFF
--- a/src/actions/expressions.js
+++ b/src/actions/expressions.js
@@ -21,8 +21,7 @@ function expressionExists(expressions, input) {
  * @memberof actions/pause
  * @static
  */
-
-export function addExpression(input: string, { visible = true } = {}) {
+export function addExpression(input: string, { visible = true }: Object = {}) {
   return ({ dispatch, getState }: ThunkArgs) => {
     const expressions = getExpressions(getState());
     if (!input || expressionExists(expressions, input)) {
--- a/src/actions/expressions.js
+++ b/src/actions/expressions.js
@@ -22,12 +22,12 @@ function expressionExists(expressions, input) {
  * @static
  */
 export function addExpression(input: string, { visible = true }: Object = {}) {
-  return ({ dispatch, getState }: ThunkArgs) => {
+  return async ({ dispatch, getState }: ThunkArgs) => {
     const expressions = getExpressions(getState());
     if (!input || expressionExists(expressions, input)) {
       const expression = getExpression(getState(), input);
       if (!expression.visible && visible) {
-        dispatch(deleteExpression(expression));
+        await dispatch(deleteExpression(expression));
       } else {
         return;
       }
--- a/src/actions/expressions.js
+++ b/src/actions/expressions.js
@@ -2,8 +2,7 @@
 
 import constants from "../constants";
 import { PROMISE } from "../utils/redux/middleware/promise";
-
-import { getExpressions, getSelectedFrame } from "../selectors";
+import { getExpression, getExpressions, getSelectedFrame } from "../selectors";
 
 import type { Expression } from "../types";
 import type { ThunkArgs } from "./types";
@@ -22,11 +21,17 @@ function expressionExists(expressions, input) {
  * @memberof actions/pause
  * @static
  */
-export function addExpression(input: string, visible: boolean = true) {
+
+export function addExpression(input: string, { visible = true } = {}) {
   return ({ dispatch, getState }: ThunkArgs) => {
-    const expressions = getExpressions(getState(), visible);
+    const expressions = getExpressions(getState());
     if (!input || expressionExists(expressions, input)) {
-      return;
+      const expression = getExpression(getState(), input);
+      if (!expression.visible && visible) {
+        dispatch(deleteExpression(expression));
+      } else {
+        return;
+      }
     }
 
     dispatch({
@@ -51,6 +56,7 @@ export function updateExpression(input: string, expression: Expression) {
       type: constants.UPDATE_EXPRESSION,
       expression,
       input: input,
+      visible: expression.visible,
     });
 
     const selectedFrame = getSelectedFrame(getState());
@@ -81,9 +87,9 @@ export function deleteExpression(expression: Expression) {
  * @param {number} selectedFrameId
  * @static
  */
-export function evaluateExpressions(frameId: frameIdType, visible: boolean = true) {
+export function evaluateExpressions(frameId: frameIdType) {
   return async function({ dispatch, getState, client }: ThunkArgs) {
-    const expressions = getExpressions(getState(), visible).toJS();
+    const expressions = getExpressions(getState()).toJS();
     for (let expression of expressions) {
       await dispatch(evaluateExpression(expression, frameId));
     }
@@ -100,7 +106,7 @@ function evaluateExpression(expression, frameId: frameIdType) {
     return dispatch({
       type: constants.EVALUATE_EXPRESSION,
       input: expression.input,
-      visible: expression.visible == undefined ? true : expression.visible,
+      visible: expression.visible,
       [PROMISE]: client.evaluate(expression.input, { frameId }),
     });
   };
--- a/src/actions/expressions.js
+++ b/src/actions/expressions.js
@@ -22,9 +22,9 @@ function expressionExists(expressions, input) {
  * @memberof actions/pause
  * @static
  */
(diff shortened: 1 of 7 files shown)


src/actions/expressions.js
```diff
@@ -22,21 +22,22 @@ function expressionExists(expressions, input) {
  * @memberof actions/pause
  * @static
  */
-function addExpression(input: string) {
+function addExpression(input: string, visible: boolean = true) {
```
jasonLaster: I've seen secondary options in a trailing options param in some other actions like `addBreakpoint` which might be nice here.

```js
addExpression(input: string, { visible: boolean = true }) {
  // stuff
}
```

src/actions/expressions.js
```diff
@@ -80,9 +81,10 @@ function deleteExpression(expression: Expression) {
  * @param {number} selectedFrameId
  * @static
  */
-function evaluateExpressions(frameId: frameIdType) {
+function evaluateExpressions(frameId: frameIdType, visible: boolean = true) {
```
jasonLaster: Why would we want to only evaluate visible / hidden expressions?

src/actions/expressions.js
```diff
@@ -80,9 +81,10 @@ function deleteExpression(expression: Expression) {
...
   return async function({ dispatch, getState, client }: ThunkArgs) {
-    for (let expression of getExpressions(getState())) {
+    const expressions = getExpressions(getState(), visible).toJS();
+    for (let expression of expressions) {
```
jasonLaster: 👍

src/actions/expressions.js
```diff
@@ -98,6 +100,7 @@ function evaluateExpression(expression, frameId: frameIdType) {
     return dispatch({
       type: constants.EVALUATE_EXPRESSION,
       input: expression.input,
+      visible: expression.visible == undefined ? true : expression.visible,
```
jasonLaster: hmm, i wonder if we should force `expression.visible` to be a boolean and then not worry about these other states... Flow to the rescue?

src/components/Editor/index.js
```diff
@@ -204,6 +206,8 @@ const Editor = React.createClass({
     } else if (this.props.sourceText) {
       this.setText(this.props.sourceText.get("text"));
     }
+
+    this.previewSelectedToken = debounce(this.previewSelectedToken, 50);
```
jasonLaster: I think 50 should bumped up to 100.

src/components/Editor/index.js
```diff
@@ -278,10 +276,15 @@ const Editor = React.createClass({
       sourceText
     );
 
+    if (expressionFromToken) {
+      addExpression(expressionFromToken.value, false);
```
jasonLaster: `addExpression(expressionFromToken.value, { visible: false });` is clearer

src/components/Editor/index.js
```diff
@@ -587,18 +590,26 @@ const Editor = React.createClass({
...
 
     const token = selectedToken.textContent;
     selectedToken.classList.add("selected-token");
 
-    const value = selectedExpression.value || selectedExpression.contents.value;
+    let value = get(selectedExpression, "contents.value");
+    if (!value) {
+      const exp = this.props.getExpression(selectedExpression.value, false);
+      value = get(exp, "value.result");
+    }
```
jasonLaster: hmm - lets move this to a selector: `getExpressionResult` seems simpler :)

src/reducers/expressions.js
```diff
@@ -9,38 +9,59 @@ import type { Expression } from "../types";
...
+  expressions: I.Map<string, Expression>,
 };
 
 const State = makeRecord(
   ({
-    expressions: I.List(restoreExpressions()),
+    expressions: I.Map({
+      visible: I.List(restoreExpressions()),
+      hidden: I.List(),
+    }),
```
jasonLaster: nit: it might be simpler to keep one list and then filter the list when persisting and selecting values. It's *slightly* less efficient but will be easier when we want all expressions.

src/reducers/expressions.js
```diff
@@ -81,27 +104,44 @@ function updateItemInList(
...
+  storeExpressions(newState, path);
   return newState;
 }
 
-function deleteExpression(state: State, input: string) {
-  const index = getExpressions({ expressions: state }).findKey(
+function deleteExpression(
+  state: State,
+  input: string,
+  visible: boolean = true
```
jasonLaster: `visible` is less important if we keep expressions in one less

src/reducers/expressions.js
```diff
@@ -81,27 +104,44 @@ function updateItemInList(
...
-  return state.expressions.get("expressions");
+function getExpressions(state: OuterState, visible: boolean = true) {
+  return state.expressions.getIn(["expressions", expressionType(visible)]);
+}
+
+function getExpression(
+  state: OuterState,
+  input: string,
+  visible: boolean = true
+) {
```
jasonLaster: haha - thank you  prettier

jasonLaster:
This is such a nice feature!

src/components/Editor/index.js
```diff
@@ -587,18 +590,26 @@ const Editor = React.createClass({
...
-    const value = selectedExpression.value || selectedExpression.contents.value;
+    let value = get(selectedExpression, "contents.value");
+    if (!value) {
+      const exp = this.props.getExpression(selectedExpression.value);
+      value = get(exp, "value.result");
+    }
+
+    if (!value) {
+      return;
+    }
```
jasonLaster: This could be a util: `getExpressionResult`

```js
let value = get(selectedExpression, "contents.value");
if (!value) {
   const exp = this.props.getExpression(selectedExpression.value);
   value = get(exp, "value.result");
 }
 if (!value) {
   return;
 }
```

src/components/SecondaryPanes/Expressions.js
```diff
@@ -223,7 +223,7 @@ Expressions.displayName = "Expressions";
 export default connect(
   state => ({
     pauseInfo: getPause(state),
-    expressions: getExpressions(state),
+    expressions: getExpressions(state, true),
```
jasonLaster: nit: ` true` is vague `getExpressions(state, true),`

src/reducers/expressions.js
```diff
@@ -96,12 +102,18 @@ function deleteExpression(state: State, input: string) {
 
 type OuterState = { expressions: Record<ExpressionState> };
 
-function getExpressions(state: OuterState) {
-  return state.expressions.get("expressions");
+function getExpressions(state: OuterState, visible: boolean = false) {
+  const expressions = state.expressions.get("expressions");
+  return visible ? expressions.filter(e => e.visible) : expressions;
+}
```
jasonLaster: This is a minor preference, but I think we could likely split this into two selectors: `getExpressions` and `getVisibleExpressions`

src/actions/expressions.js
```diff
@@ -22,21 +26,27 @@ function expressionExists(expressions, input) {
...
  */
-function addExpression(input: string) {
+function addExpression(input: string, { visible = true }: Object = {}) {
   return ({ dispatch, getState }: ThunkArgs) => {
     const expressions = getExpressions(getState());
     if (!input || expressionExists(expressions, input)) {
-      return;
+      const expression = getExpression(getState(), input);
+      if (!expression.visible && visible) {
+        dispatch(deleteExpression(expression));
```
jasonLaster: hmm, maybe we want to use `await` here which will mean the outer function `return ({ dispatch, getState }: ThunkArgs) => {` should be async. There are examples of this elsewhere

src/components/Editor/index.js
bomsy: hmm, it looks `getExpression` is tied to the state.

jasonLaster:
functions are assumed to be read-only
https://github.com/facebook/flow/issues/1397#issuecomment-183428756

Issue #2455, opened 2017-03-25
 broken formatting in the CONTRIBUTING.md


abhishekg785:
Broken table in the 'About Us'  section of the CONTRIBUTING.md.

PR #2456, opened 2017-03-25
broken table fixed

abhishekg785:
Fix for the issue #2455


jasonLaster:
I'm now scared that we'll keep breaking tables :)

PR #2461, opened 2017-03-26
[SearchBar] Fix cursor for disabled search modifier

irfanhudda:
Cursor for disabled search modifier was not correct
cursor

### Summary of changes

* Change cursor of disabled search modifier to `default`.
* Add hover background for search modifier.

Hover style
hover2

### DIFF
--- a/src/components/Editor/SearchBar.css
+++ b/src/components/Editor/SearchBar.css
@@ -66,6 +66,10 @@
   cursor: pointer;
 }
 
+.search-bottom-bar .search-modifiers button:hover {
+  background: var(--theme-toolbar-background-hover);
+}
+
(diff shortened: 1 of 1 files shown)


jasonLaster:
I love this polish!

Is this consistent with what we do with disabled stepping butons?

irfanhudda:
@jasonLaster Background color on hover is same as the one we use for stepping buttons

https://github.com/devtools-html/debugger.html/blob/master/src/components/SecondaryPanes/CommandBar.css#L32

Although the shape of border is not exactly same

| | |
|----|----|
|hover2|hover3|

Hover background is on `Step over` button in second screenshot

PR #2469, opened 2017-03-28
[WIP] converting the shared/menu component to es modules

clarkbw:
Associated Issue: #1884

### Summary of Changes

* converted first require to import

### DIFF
--- a/src/components/shared/menu.js
+++ b/src/components/shared/menu.js
@@ -92,9 +92,7 @@ function buildMenu(items: Array<any>) {
     .filter(itm => itm !== null);
 }
 
-// export default Dropdown;
-
-module.exports = {
+export default {
   showMenu,
   buildMenu,
 };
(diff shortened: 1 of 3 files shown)


gregtatum:
Hey Bryan, thanks for the PR. The first thing that caught my eye was that you can export individual functions instead of using one big export default. See https://github.com/devtools-html/debugger.html/blob/3e86312a46a0e37450785b27a90224be647a58da/src/utils/editor/expression.js#L7

src/components/shared/menu.js
```diff
@@ -92,7 +92,9 @@ function buildMenu(items: Array<any>) {
     .filter(itm => itm !== null);
 }
 
-export default {
+// export default Dropdown;
```
clarkbw: I was trying this but then got console errors.

gregtatum:
https://github.com/devtools-html/debugger.html/blob/3e86312a46a0e37450785b27a90224be647a58da/src/components/App.js#L22-L26

Also, default exports from ES modules require this `.default` at the end of the requires. Check out the example above.

jasonLaster:
Thanks @gregtatum

PR #2478, opened 2017-03-28
(WIP) Add client to debugger

jasonLaster:
Associated Issue: https://github.com/devtools-html/devtools-core/issues/287

### Summary of Changes

This moves the part of client adapters responsible for debugger commands and events back to debugger. The rationale for this move is:

* other tools like the inspector, console, don't need the debugger client.
* patches to the debugger occasionally touch the client
* versioning and packages is overhead that slows us down
* moving the debugger logic out frees the client adapter to fold into the launchpad and make the launchpad a devtool tab launcher! (find tabs and connect to them)

Yes, this patch scares the hell out of me :)


src/client/chrome/commands.js
```diff
@@ -0,0 +1,130 @@
...
+}
+
+function pauseOnExceptions(
+  shouldPauseOnExceptions: boolean,
+  shouldIgnoreCaughtExceptions: boolean
+) {
+  if (!shouldPauseOnExceptions) {
+    return debuggerAgent.setPauseOnExceptions({ state: "none" });
+  }
+  const state = shouldIgnoreCaughtExceptions ? "uncaught" : "all";
```
nchevobbe: Maybe not in the scope of this review, but it would be nice if the `state` possible values (`none`, `uncaught`, `all`, ..) were in constants somewhere

src/client/chrome/commands.js
```diff
@@ -0,0 +1,130 @@
...
+  const { result } = await runtimeAgent.getProperties({
+    objectId: object.objectId,
+  });
+
+  const loadedObjects = result.map(createLoadedObject);
+
+  return { loadedObjects };
+}
+
+function evaluate(script: string) {
```
nchevobbe: could we rename `script` to `expression` so it matches what we passe in `evaluate` ?

src/client/chrome/commands.js
```diff
@@ -0,0 +1,130 @@
...
+  return runtimeAgent.evaluate({ expression: script });
+}
+
+function debuggeeCommand(script: string) {
+  evaluate(script);
+  return Promise.resolve();
```
nchevobbe: out of curiosity, why are we returning a self-resolving promise here ?

src/client/chrome/create.js
```diff
@@ -0,0 +1,48 @@
...
+      line: serverLocation.lineNumber + 1,
+      column: serverLocation.columnNumber,
+    };
+  }
+}
+
+function toServerLocation(location: Location): ServerLocation {
+  return {
+    scriptId: location.sourceId,
+    lineNumber: location.line - 1,
```
nchevobbe: shouldn't we have `columnNumber: location.column` too ?

src/client/chrome/events.js
```diff
@@ -0,0 +1,132 @@
...
+
+  actions.newSource({
+    id: scriptId,
+    url,
+    sourceMapURL,
+    isPrettyPrinted: false,
+  });
+}
+
+function scriptFailedToParse() {}
```
nchevobbe: could we add a comment explaining why we must have this ?

src/client/chrome/events.js
```diff
@@ -0,0 +1,132 @@
...
+async function paused(
+  {
+    callFrames,
+    reason,
+    data,
+    hitBreakpoints,
+    asyncStackTrace,
+  }: any
+) {
+  const frames = callFrames.map(createFrame);
```
nchevobbe: could we check that callFrames is a non-empty array

src/client/chrome/events.js
```diff
@@ -0,0 +1,132 @@
...
+
+  const objectId = frame.scopeChain[0].object.objectId;
+  const { result } = await runtimeAgent.getProperties({
+    objectId,
+  });
+
+  const loadedObjects = result.map(createLoadedObject);
+
+  if (clientType == "chrome") {
+    pageAgent.configureOverlay({ message: "Paused in debugger.html" });
```
nchevobbe: I guess we should localize the message

src/client/chrome/events.js
```diff
@@ -0,0 +1,132 @@
...
+
+function resumed() {
+  if (clientType == "chrome") {
+    pageAgent.configureOverlay({ suspended: false });
+  }
+
+  actions.resumed();
+}
+
+function globalObjectCleared() {}
```
nchevobbe: add a comment explaining why we need this

src/client/chrome/events.js
```diff
@@ -0,0 +1,132 @@
...
+// Page Events
+function frameNavigated(frame: any) {
+  actions.navigated();
+}
+
+function frameStartedLoading() {
+  actions.willNavigate();
+}
+
+function domContentEventFired() {}
```
nchevobbe: add a comment explaining why we need those

src/client/firefox/events.js
```diff
@@ -0,0 +1,72 @@
...
+
+let threadClient: ThreadClient;
+let actions: Actions;
+
+function setupEvents(dependencies: Dependencies) {
+  threadClient = dependencies.threadClient;
+  actions = dependencies.actions;
+
+  Object.keys(clientEvents).forEach(eventName => {
+    if (threadClient) {
```
nchevobbe: could we move the forEach call inside the if ? 

```js
if (threadClient) {
  Object.keys(clientEvents).forEach(eventName => {
    threadClient.addListener(eventName, clientEvents[eventName]);
  });
}
```

src/client/firefox/types.js
```diff
@@ -0,0 +1,364 @@
...
+ * BreakpointClient
+ * @memberof firefox
+ * @static
+ */
+export type BreakpointClient = {
+  actor: ActorId,
+  remove: () => void,
+  location: Location,
+  setCondition: (ThreadClient, boolean, boolean) => Promise<BreakpointClient>,
+  // getCondition: () => any,
```
nchevobbe: why are those commented ?

src/components/Editor/index.js
```diff
@@ -254,11 +254,11 @@ const Editor = React.createClass({
   },
 
   /*
-   * The default Esc command is overridden in the CodeMirror keymap to allow
-   * the Esc keypress event to be catched by the toolbox and trigger the
-   * split console. Restore it here, but preventDefault if and only if there
-   * is a multiselection.
-  */
+     * The default Esc command is overridden in the CodeMirror keymap to allow
```
nchevobbe: is this change needed ? I fell like the previous version was correct

src/components/Editor/index.js
```diff
@@ -549,10 +549,10 @@ const Editor = React.createClass({
   },
 
   /**
-   * Handle getting the source document or creating a new
-   * document with the correct mode and text.
-   *
-   */
+     * Handle getting the source document or creating a new
```
nchevobbe: is this change needed ? I fell like the previous version was correct

assets/panel/panel.js
```diff
@@ -16,49 +15,45 @@ function DebuggerPanel(iframeWindow, toolbox) {
 }
 
 DebuggerPanel.prototype = {
-  open: Task.async(function* () {
+  open: Task.async(function*() {
```
jryans: I believe it should be safe to do so here... Firefox supports async / await natively now.

assets/panel/panel.js
```diff
@@ -68,10 +63,10 @@ DebuggerPanel.prototype = {
     return { frames, selected };
   },
 
-  destroy: function () {
+  destroy: function() {
```
jryans: DevTools ESLint style in m-c requires `function (` with the space. Should we update the debugger `.eslintrc` so that they match on this point?

jasonLaster: oops, this change was accidental. 

We could update the eslint to be consistent, but we started using prettier on the debugger for the styling and I like the defaults that it has. (even though it's not consistent with devtools in this case)

jasonLaster:
@nchevobbe many of your comments were on chrome/firefox files that were copied over. I'll reply, but we shouldn't change these things now

src/client/firefox/types.js
jasonLaster: hmm, i don't remember.

src/client/chrome/events.js
jasonLaster: it's a chrome message, but yea

src/client/chrome/create.js
jasonLaster: yep. good point. this will break w/ source maps

src/client/chrome/commands.js
jasonLaster: yep :) 

It's a weird function that's only for testing.

src/client/chrome/commands.js
jasonLaster: yeah - that would be clearer

nchevobbe:
@jasonLaster sure

PR #2481, opened 2017-03-28
Convert actions to use ES6 import/exports

wldcordeiro:
Associated Issue: #1884

### Summary of Changes

* modify the actions to use ES6 imports/exports

### Test Plan

- [x] Run the test suites
- [x] Test out basic features of debugger
  - add/remove/enable/disable breakpoints
  - toggle UI states
  - add watch expressions
  - etc

### DIFF
--- a/src/actions/index.js
+++ b/src/actions/index.js
@@ -1,15 +1,15 @@
 // @flow
 
-const breakpoints = require("./breakpoints");
-const expressions = require("./expressions");
-const eventListeners = require("./event-listeners");
-const sources = require("./sources");
-const pause = require("./pause");
-const navigation = require("./navigation");
-const ui = require("./ui");
-const coverage = require("./coverage");
+import * as breakpoints from "./breakpoints";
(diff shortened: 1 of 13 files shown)


src/actions/index.js
```diff
@@ -1,15 +1,15 @@
...
-const ui = require("./ui");
-const coverage = require("./coverage");
+import * as breakpoints from "./breakpoints";
+import * as expressions from "./expressions";
+import * as eventListeners from "./event-listeners";
+import * as sources from "./sources";
+import * as pause from "./pause";
+import * as navigation from "./navigation";
+import * as ui from "./ui";
+import * as coverage from "./coverage";
```
jasonLaster: wow what does that do?

wldcordeiro: When you have no default export this is roughly the same as

```js
import { foo, bar, baz} from "./mod";
const mod = { foo, bar, baz };
```

PR #1885, opened 2017-01-30
Es6 modules

jasonLaster:
Associated Issue: #1884

### Summary of Changes

* worked on Secondary Panes

### Test Plan

trusting integration tests and an initial page refresh

PR #2483, opened 2017-03-28
Fix ES6 actions imports

wldcordeiro:
Associated Issue: #1885 

### Summary of Changes

* Fix some additional require calls

I had missed these require calls that needed to import the default.


wldcordeiro:
@jasonLaster don't know what codecov is failing on but this should fix the things missed in the last PR.

Issue #2467, opened 2017-03-27
[FunctionSearch] NaN results 


jasonLaster:
@jacobjzhang has a `NaN` in his recent GIF. I'm not quite sure how he got it.

STR:
1. go to `todo-view.js` 
2. select function search
3. search for `close`

?

![](http://g.recordit.co/KEEn6tvCCX.gif)

jacobjzhang:
Looks like after toggling function search off, there is no count or index passed as props.
<img width="1393" alt="screen shot 2017-03-27 at 5 57 36 pm copy" src="https://cloud.githubusercontent.com/assets/6591285/24379915/300d4ba0-1317-11e7-8f37-23627ef9d20a.png">

bomsy:
looks like this is fixed!

PR #2484, opened 2017-03-29
Fix NaN bug when toggling function search

jacobjzhang:
Associated Issue: #2467

### Summary of Changes

* Passed default necessary index value in symbolic search as prop.

### Test Plan

1) Command-F in Editor opens Search.
2) Choose function search and look for a function.
3) Click the function and watch it highlight.
4) Escape out of the search box and Command-F again.
5) Check upper right hand corner.

Before: http://g.recordit.co/VNsJHS8jhC.gif

### DIFF
--- a/src/components/Editor/SearchBar.js
+++ b/src/components/Editor/SearchBar.js
@@ -307,7 +307,8 @@ const SearchBar = React.createClass({
       { key: "value" }
     );
 
-    updateSearchResults({ count: symbolSearchResults.length });
+    // symbol search doesn't use index so we pass the default value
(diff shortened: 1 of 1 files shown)


src/components/Editor/SearchBar.js
```diff
@@ -307,7 +307,8 @@ const SearchBar = React.createClass({
       { key: "value" }
     );
 
-    updateSearchResults({ count: symbolSearchResults.length });
+    // symbol search doesn't use index so we pass the default value
+    updateSearchResults({ count: symbolSearchResults.length, index: -1 });
```
jasonLaster: that's awesome!

PR #2488, opened 2017-03-29
Polish updateSearchResults

jasonLaster:
Small tweaks


jacobjzhang:
Looks good, contributors in the future can tell what's required, and provides more stability.

Issue #2004, opened 2017-02-13
TabBar drop down button (>>) is not displayed when resizing the sidebar


nchevobbe:
STR: 
1. Open 2 tabs 
2. Resize the right sidebar until one of the tab is hidden

Expected result:
The tab is hidden and the drop down button (>>) is displayed

Actual result:
The tab is hidden, but the drop down button isn't displayed

GIF : 
![peek 13-02-2017 12-11](https://cloud.githubusercontent.com/assets/578107/22881187/be65409a-f1e5-11e6-836c-d262a94edc83.gif)

I guess we don't check if we should show the button when resizing the sidebars.

_Bug found by @princiya , thanks !_

jasonLaster:
Hmm. Good catch. We do have an on resize check so it should work. I think @princiya is looking at this with her flex changes. It's most likely how we calculate hidden tabs

jasonLaster:
i believe we fixed this with @princiya's work.

princiya:
This isn't fixed yet. It works with window resize event, not CSS div resize. From the gif, it's a CSS div resize happening there. I was looking at this last week, but couldn't finish it up ;) Let me know the priority for this and I will get back to this soon :)

jasonLaster:
Ahh. Thanks for clarifying. 

No pressure.

jasonLaster:
@AnshulMalik is working on this

jasonLaster:
one option


```diff
diff --git a/src/components/App.js b/src/components/App.js
index 3d4e53d..6c2d1db 100644
--- a/src/components/App.js
+++ b/src/components/App.js
@@ -79,6 +79,12 @@ class App extends Component {
     const { horizontal } = this.state;

     const overflowX = endPanelCollapsed ? "hidden" : "auto";
+    const onResize = (panel, size) => {
+      if (panel == "start") {
+        this.setState({startPanelSize: size})
+      }
+      this.setState({endPanelSize: size})
+    }

     return dom.div(
       { className: "debugger" },
@@ -88,6 +94,7 @@ class App extends Component {
         minSize: 10,
         maxSize: "50%",
         splitterSize: 1,
+        onResize,
         startPanel: Sources({ sources, horizontal }),
         startPanelCollapsed,
         endPanel: SplitBox({
@@ -96,6 +103,7 @@ class App extends Component {
           minSize: 10,
           maxSize: "80%",
           splitterSize: 1,
+          onResize,
           endPanelControl: true,
[...]

PR #2492, opened 2017-03-29
Fixes tabs dropdown button now shown onResize

AnshulMalik:
Associated Issue: #2004 

### Summary of Changes

* Added onResize with splitter

### Test Plan

- [x] After resize, dropdown button appears

### DIFF
--- a/src/components/App.js
+++ b/src/components/App.js
@@ -28,7 +28,7 @@ class App extends Component {
   state: {
     horizontal: boolean,
     startPanelSize: number,
-    endPanelSize: number,
+    endPanelSize: number
   };
   onLayoutChange: Function;
   getChildContext: Function;
@@ -40,7 +40,7 @@ class App extends Component {
     this.state = {
       horizontal: verticalLayoutBreakpoint.matches,
       startPanelSize: 0,
-      endPanelSize: 0,
+      endPanelSize: 0
     };
 
     this.getChildContext = this.getChildContext.bind(this);
@@ -76,7 +76,7 @@ class App extends Component {
           startPanelCollapsed,
           endPanelCollapsed,
           horizontal,
-          endPanelSize,
+          endPanelSize
         }),
         Editor({ horizontal }),
         !this.props.selectedSource ? WelcomeBox({ horizontal }) : null,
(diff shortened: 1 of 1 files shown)


src/components/App.js
```diff
@@ -88,6 +108,7 @@ class App extends Component {
         minSize: 10,
         maxSize: "50%",
         splitterSize: 1,
+        onResize: onResize.bind(this, "start"),
```
jasonLaster: hmm, maybe we should change this to be: 

```js
onResize: size => this.setState({startPanelSize: size})
```

it's a little clenaer

src/components/App.js
```diff
@@ -27,15 +27,23 @@ const WelcomeBox = createFactory(require("./WelcomeBox").default);
...
   renderEditorPane: Function;
   renderVerticalLayout: Function;
 
   constructor(props) {
     super(props);
-    this.state = { horizontal: verticalLayoutBreakpoint.matches };
+    this.state = {
+      horizontal: verticalLayoutBreakpoint.matches,
+      startPanelSize: 0,
+      endPanelSize: 0,
```
jasonLaster: I think @wldcordeiro might recommend that we put this state in the UI state. 

What do you think?

src/components/App.js
```diff
@@ -61,12 +69,17 @@ class App extends Component {
...
     return dom.div(
       { className: "editor-pane" },
       dom.div(
         { className: "editor-container" },
-        EditorTabs({ startPanelCollapsed, endPanelCollapsed, horizontal }),
+        EditorTabs({
+          startPanelCollapsed,
+          endPanelCollapsed,
+          horizontal,
+          endPanelSize,
```
jasonLaster: is `endPanelSize` used?

AnshulMalik:
I'll create a PR in devtools core, then it'll work fine.

src/components/App.js
wldcordeiro: Yeah. 😛  https://github.com/devtools-html/debugger.html/pull/2492/files#diff-57aa11691094c3295e1f9dcbe8432ba2R113

src/components/App.js
wldcordeiro: I wouldn't be against that, we already store that collapse state there.

src/components/App.js
AnshulMalik: It's working with updated devtools-core splitter package :)

AnshulMalik:
Thanks for the tweaks :)

PR #2494, opened 2017-03-30
improve readibility of readme

NikoLewis:
Associated Issue: #None

### Summary of Changes

* Improved setup instruction.
* Clarified how to verify it's working.

### Test Plan

n/a


jasonLaster:
Thanks @NikoLewis. I've updated the README, which had gotten out of date. 

I think the screenshot is a good idea, but that image is in the longer setup guide

PR #2495, opened 2017-03-30
Update Gecko in CI Docker image to include devtools-source-map v0.3.0

jryans:
* Gecko in CI Docker image includes devtools-source-map v0.3.0, which will be needed to use source maps from the toolbox (upcoming work)
* Changed to referencing the Docker image repo by hash, which should avoid the need to create a new repo name for each update


jryans:
Looks like CircleCI failed because CircleCI 1.0's Docker is too old and doesn't support pulling an image by hash like I tried. So for now, I'll go back to just the Docker repo name, but let's try to finish upgrading to CircleCI 2.0 soon.

Issue #2475, opened 2017-03-28
remove yarn version from setup docs


clarkbw:
Right now we recommend people install yarn version `0.19.1` which isn't the latest version of yarn and so after they run `yarn install` yarn complains that it is outdated. 

I don't believe we need to stick to a specific version of yarn like we used to so I think we can start removing this requirement.

PR #2497, opened 2017-03-30
Remove yarn version from docs and circle config

clarkbw:
Associated Issue: #2475

### Summary of Changes

I don't believe we need to ensure a version of `yarn` anymore.  Now that the package has stabilized we can likely trust the latest stable version instead of a custom `0.19.1` version.

* removed yarn version requirements from circle ci config
* removed yarn version suggestions in docs

### Test Plan

- run this on circle
- try to pass local tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ debugger.html is a hackable debugger for modern times, built from the ground up 
 > Or take a look at our detailed [getting started][getting-started] instructions.
 
 ```bash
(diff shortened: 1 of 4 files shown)


clarkbw:
Strange, we don't need to do that for `devtools-core` https://github.com/devtools-html/devtools-core/blob/master/circle.yml

clarkbw:
Oh right, I remember now.  We are on a linux build that works for our Docker system but doesn't have yarn installed by default.  Will put those pieces back.

PR #2504, opened 2017-03-30
Fix worker test failures

jasonLaster:
### Summary of Changes

We were seeing `DebuggerConfig` errors in our unit tests because the new source map worker was using DebuggerConfig. The new version of the debugger bypasses that with a new explicit start/stop call to the workers.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -20,6 +20,7 @@ test:
     - mkdir -p $CIRCLE_TEST_REPORTS/mocha
     - mkdir -p $CIRCLE_TEST_REPORTS/flow-coverage
     - mkdir -p $CIRCLE_TEST_REPORTS/jest-coverage
+    - node ./node_modules/.bin/jest --testResultsProcessor jest-junit-reporter --coverage --coverageDirectory=$CIRCLE_TEST_REPORTS/jest-coverage
     - yarn lint-css
     - yarn lint-js
     - yarn lint-md
--- a/package.json
+++ b/package.json
@@ -51,7 +51,7 @@
     "codemirror": "^5.1.0",
     "devtools-launchpad": "^0.0.61",
     "devtools-reps": "^0.5.0",
-    "devtools-source-map": "0.3.1",
+    "devtools-source-map": "0.3.3",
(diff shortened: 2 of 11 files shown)


circle.yml
```diff
@@ -20,6 +20,7 @@ test:
     - mkdir -p $CIRCLE_TEST_REPORTS/mocha
     - mkdir -p $CIRCLE_TEST_REPORTS/flow-coverage
     - mkdir -p $CIRCLE_TEST_REPORTS/jest-coverage
+    - node ./node_modules/.bin/jest --testResultsProcessor jest-junit-reporter --coverage --coverageDirectory=$CIRCLE_TEST_REPORTS/jest-coverage
```
jasonLaster: this was removed during the client refactor, which forced us to upgrade the source map worker with the bad config.

package.json
```diff
@@ -123,8 +123,9 @@
       "node_modules/(?!devtools-)"
     ],
     "setupFiles": [
-      "<rootDir>/test/tests-boot.js"
+      "<rootDir>/test/tests-setup.js"
     ],
+    "setupTestFrameworkScriptFile": "<rootDir>/test/tests-framework.js",
```
jasonLaster: the new script runs post setup.

src/test/tests-framework.js
```diff
@@ -0,0 +1,28 @@
...
+  stopParserWorker,
+} = require("../utils/parser");
+
+const { getValue } = require("devtools-config");
+
+beforeAll(() => {
+  startSourceMapWorker(getValue("workers.sourceMapURL"));
+  startPrettyPrintWorker(getValue("workers.prettyPrintURL"));
+  startParserWorker(getValue("workers.parserURL"));
+});
```
jasonLaster: I would re-use the bootstrapWorker function, but unfortunately, `bootstrap.js` pulls in `devtools-launchpad` which has `DebuggerConfig`. At some point it would be nice to use babel to replace `DebuggerConfig` with the appropriate config, but that is difficult given how jest uses babel.

PR #2506, opened 2017-03-30
Use source map service from toolbox in Firefox

jryans:
### Summary of Changes

* When used as a Firefox panel, the debugger gets a `sourceMapService` object from the toolbox, which manages the worker itself, etc. Those details are hidden from the debugger.
* When used in a tab, the debugger starts the source map worker like it did before.

### Test Plan

- [x] Verified source maps work when running in a tab
- [x] Verified source maps work when running as a panel

### DIFF
--- a/src/actions/navigation.js
+++ b/src/actions/navigation.js
@@ -1,5 +1,4 @@
 import constants from "../constants";
-import { clearSourceMaps } from "devtools-source-map";
 import { clearDocuments } from "../utils/editor";
 import { getSources } from "../reducers/sources";
 import { waitForMs } from "../utils/utils";
@@ -15,12 +14,14 @@ import { newSources } from "./sources";
  * @static
  */
 export function willNavigate(_, event) {
-  clearSourceMaps();
-  clearDocuments();
+  return async function({ dispatch, getState, client, sourceMaps }: ThunkArgs) {
+    await sourceMaps.clearSourceMaps();
+    clearDocuments();
 
-  return {
-    type: constants.NAVIGATE,
-    url: event.url,
+    dispatch({
+      type: constants.NAVIGATE,
+      url: event.url,
+    });
   };
 }
 
--- a/src/actions/pause.js
+++ b/src/actions/pause.js
@@ -42,9 +42,9 @@ export function resumed() {
  * @static
  */
 export function paused(pauseInfo: Pause) {
-  return async function({ dispatch, getState, client }: ThunkArgs) {
+  return async function({ dispatch, getState, client, sourceMaps }: ThunkArgs) {
     let { frames, why, loadedObjects } = pauseInfo;
-    frames = await updateFrameLocations(frames);
+    frames = await updateFrameLocations(frames, sourceMaps);
     const frame = frames[0];
 
     dispatch({
--- a/src/client/index.js
+++ b/src/client/index.js
@@ -6,7 +6,7 @@ const { prefs } = require("../utils/prefs");
 const {
   bootstrapApp,
   bootstrapStore,
-  bootstrapWorker,
+  bootstrapWorkers,
 } = require("../utils/bootstrap");
 
 function loadFromPrefs(actions: Object) {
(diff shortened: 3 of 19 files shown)


src/utils/teardown.js
```diff
@@ -1,9 +0,0 @@
-const { stopSourceMapWorker } = require("devtools-source-map");
-const { stopPrettyPrintWorker } = require("../utils/pretty-print");
-const { stopParserWorker } = require("../utils/parser");
-
-export function teardownWorkers() {
-  stopSourceMapWorker();
-  stopPrettyPrintWorker();
-  stopParserWorker();
-}
```
jryans: I thought it was helpful to see both bootstrap and teardown in the same file. For example with this change, we need to use the same Firefox conditional for the source map worker. With it all in one file, it's easy to see they work the same way.

src/actions/navigation.js
```diff
@@ -15,12 +14,14 @@ import { newSources } from "./sources";
  * @static
  */
 export function willNavigate(_, event) {
-  clearSourceMaps();
-  clearDocuments();
+  return async function({ dispatch, getState, client, sourceMaps }: ThunkArgs) {
+    sourceMaps.clearSourceMaps();
```
jasonLaster: should this use `await`

src/actions/pause.js
```diff
@@ -42,9 +42,9 @@ export function resumed() {
  * @static
  */
 export function paused(pauseInfo: Pause) {
-  return async function({ dispatch, getState, client }: ThunkArgs) {
+  return async function({ dispatch, getState, client, sourceMaps }: ThunkArgs) {
     let { frames, why, loadedObjects } = pauseInfo;
-    frames = await updateFrameLocations(frames);
+    frames = await updateFrameLocations(frames, sourceMaps);
```
jasonLaster: 👍

src/client/index.js
```diff
@@ -21,16 +21,16 @@ function getClient(connection: any) {
...
   if (!connection) {
     return;
   }
 
   const client = getClient(connection);
-  const { store, actions, selectors } = bootstrapStore(client);
+  const { store, actions, selectors } = bootstrapStore(client, services);
 
-  bootstrapWorker();
+  bootstrapWorkers();
```
jasonLaster: 👍

src/components/Editor/EditorMenu.js
```diff
@@ -67,9 +65,8 @@ async function EditorMenu(
 
   const menuOptions = [];
 
-  if (isMapped) {
-    menuOptions.push(jumpLabel);
-  }
+  // TODO: Find a new way to only add this for mapped sources?
+  menuOptions.push(jumpLabel);
```
jasonLaster: yeah - this will be tough, the way we're doing it now. But that is okay. This will just be a no-op in the context menu.

src/main.js
```diff
@@ -11,27 +11,40 @@ const {
...
     destroy: () => {
       unmountRoot(ReactDOM);
       teardownWorkers();
     },
   };
 } else {
-  bootstrap(React, ReactDOM).then(onConnect);
+  window.L10N = L10N;
+  // $FlowIgnore:
+  window.L10N.setBundle(require("../assets/panel/debugger.properties"));
```
jasonLaster: 👍

src/utils/bootstrap.js
```diff
@@ -47,8 +56,20 @@ export function bootstrapApp(connection, { store, actions }) {
...
+    // When used in Firefox, the toolbox manages the source map worker.
+    startSourceMapWorker(getValue("workers.sourceMapURL"));
+  }
   startPrettyPrintWorker(getValue("workers.prettyPrintURL"));
   startParserWorker(getValue("workers.parserURL"));
 }
+
+export function teardownWorkers() {
+  if (!isFirefoxPanel()) {
+    // When used in Firefox, the toolbox manages the source map worker.
```
jasonLaster: 👍

src/actions/navigation.js
jryans: Hmm, probably so since it goes through the worker messaging. I'll update it.

PR #2507, opened 2017-03-31
Fix up Docker mapping to use latest files again

jryans:
When I made the latest Docker image, I edited the name of the `WORKDIR` without realizing it needed to match the paths over here, so the mochitests weren't being run against the latest changes.

To make it easier to track these things, I created a tiny repo with the Docker stuff: https://github.com/devtools-html/debugger-docker. We might trash it soon if the approach changes, but seems useful to share for now.

The test run here should be broken for now (that's expected). Tomorrow we can look into adding fixes to make the tests pass again.

### DIFF
--- a/bin/run-mochitests-docker
+++ b/bin/run-mochitests-docker
@@ -3,18 +3,18 @@
 node ./bin/publish-assets
 
 docker run -it \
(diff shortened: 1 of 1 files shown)


jryans:
The tests are failing now, as expected. I'll land this so they report correctly. @jasonLaster is working on fixing them up in #2508.

PR #2508, opened 2017-03-31
Fix mochitests - add testing vars

jasonLaster:
I started working on getting the tests passing. I'm not sure how far i'll get, but I thought I'd share early on.

### DIFF
--- a/assets/panel/panel.js
+++ b/assets/panel/panel.js
@@ -16,16 +16,17 @@ function DebuggerPanel(iframeWindow, toolbox) {
 }
 
 DebuggerPanel.prototype = {
-  open: Task.async(function* () {
+  open: async function () {
     if (!this.toolbox.target.isRemote) {
-      yield this.toolbox.target.makeRemote();
+      await this.toolbox.target.makeRemote();
     }
 
     const {
       actions,
       store,
       selectors,
-    } = yield this.panelWin.Debugger.bootstrap({
+      client,
+    } = await this.panelWin.Debugger.bootstrap({
       threadClient: this.toolbox.threadClient,
       tabTarget: this.toolbox.target,
       debuggerClient: this.toolbox.target._client,
@@ -34,9 +35,19 @@ DebuggerPanel.prototype = {
     this._actions = actions;
     this._store = store;
     this._selectors = selectors;
+    this._client = client;
     this.isReady = true;
     return this;
-  }),
+  },
+
+  getVarsForTests () {
+    return {
+      store: this._store,
+      selectors: this._selectors,
+      actions: this._actions,
+      client: this._client,
+    };
+  },
 
   _getState: function () {
     return this._store.getState();
--- a/src/test/integration/tests/editor-highlight.js
+++ b/src/test/integration/tests/editor-highlight.js
@@ -29,7 +29,6 @@ module.exports = async function(ctx) {
   // Wait for the source text to load and make sure we're in the right
   // place.
   await waitForDispatch(dbg, "LOAD_SOURCE_TEXT");
-  assertHighlightLocation(dbg, ctx, "long.js", 66);
 
   // Jump to line 16 and make sure the editor scrolled.
   await selectSource(dbg, "long.js", 16);
--- a/src/test/mochitest/browser_dbg-editor-highlight.js
+++ b/src/test/mochitest/browser_dbg-editor-highlight.js
@@ -5,7 +5,7 @@
 // matter if the source text doesn't exist yet or even if the source
 // doesn't exist.
 
-add_task(function*() {
+add_task(function* () {
   const dbg = yield initDebugger("doc-scripts.html");
   const { selectors: { getSourceText }, getState } = dbg;
   const sourceUrl = EXAMPLE_URL + "long.js";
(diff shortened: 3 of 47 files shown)


jasonLaster:
I updated this so atleast most of the tests are passing. I can finish this up tomorrow.

assets/panel/panel.js
```diff
@@ -16,7 +15,7 @@ function DebuggerPanel(iframeWindow, toolbox) {
 }
 
 DebuggerPanel.prototype = {
-  open: Task.async(function* () {
+  open: Task.async(function*() {
```
bomsy: Do we want move to async await here?

jasonLaster:
This has two big fixes:

1. fixed the panel and vars
2. reverted the prettier work on the mochitests dir

src/test/mochitest/examples/long.js
```diff
@@ -5,70 +5,66 @@ var app = {};
...
-  this.todos = this.todos.concat([
-    {
-      id: Utils.uuid(),
-      title: title,
-      completed: false,
-    },
-  ]);
+app.TodoModel.prototype.addTodo = function (title) {
+  this.todos = this.todos.concat([{
+	id: Utils.uuid(),
```
jryans: Something's wrong with the indent here and in other parts of this file. It seems like... 5 spaces?

src/test/mochitest/examples/math.min.js
```diff
@@ -1,27 +1,3 @@
...
-    sub(a, 2, function(b) {
-      mul(b, 3, function(c) {
-        div(c, 2, function(d) {
-          console.log(d);
-        });
-      });
-    });
-  });
-}
+function add(a,b,k){var result=a+b;return k(result)}function sub(a,b,k){var result=a-b;return k(result)}function mul(a,b,k){var result=a*b;return k(result)}function div(a,b,k){var result=a/b;return k(result)}function arithmetic(){
```
jryans: What happened here? The previous one is much more readable...

src/test/mochitest/examples/simple1.js
```diff
@@ -7,33 +7,25 @@ function main() {
 }
 
 function doEval() {
```
jryans: Some of these files seem to use source maps, are we sure it's correct to rewrite them?

src/test/mochitest/examples/sourcemaps2/main.js
```diff
@@ -1,23 +1,15 @@
-var message = document.getElementById("message");
+var message = document.getElementById('message');
```
jryans: Can we use `"` to match DevTools style on m-c?

src/test/mochitest/examples/math.min.js
jryans: Oh, maybe you are reverting it to how it's supposed to be for mapping? If so, sounds good.

jasonLaster: yup - that's how it was...

src/test/mochitest/examples/simple1.js
jasonLaster: this is going back to how they were...

src/test/mochitest/examples/sourcemaps2/main.js
jasonLaster: this a test file, so i'd prefer not change it until we know the tests are passing

src/test/mochitest/head.js
```diff
@@ -305,8 +305,12 @@ function createDebuggerContext(toolbox) {
  */
 function initDebugger(url, ...sources) {
   return Task.spawn(function* () {
+    Services.prefs.clearUserPref("devtools.debugger.pause-on-exceptions", false);
+    Services.prefs.clearUserPref("devtools.debugger.ignore-caught-exceptions", false);
```
jryans: This is `clear` so there is no second arg.

src/test/integration/tests/editor-highlight.js
```diff
@@ -29,7 +29,6 @@ module.exports = async function(ctx) {
   // Wait for the source text to load and make sure we're in the right
   // place.
   await waitForDispatch(dbg, "LOAD_SOURCE_TEXT");
-  assertHighlightLocation(dbg, ctx, "long.js", 66);
```
jryans: Why is this being removed? I assume it used to work...

src/test/mochitest/browser_dbg-editor-highlight.js
```diff
@@ -18,7 +18,6 @@ add_task(function* () {
   // Wait for the source text to load and make sure we're in the right
   // place.
   yield waitForDispatch(dbg, "LOAD_SOURCE_TEXT");
-  assertHighlightLocation(dbg, "long.js", 66);
```
jryans: This one too, not sure I follow?

jasonLaster: We stopped highlighting lines on debugger open last week.

I think we *do* want to highlight lines when we jump from the console, but it's not a trivial fix so i'm taking the test out now and will revisit later.

src/test/integration/tests/editor-highlight.js
jasonLaster: same as below. the test is temporarily duplicated because the web version was more intermittent in debug builds...

src/test/mochitest/head.js
jasonLaster: good call

Issue #2394, opened 2017-03-16
[Scopes] expanded properties are not updated while stepping


jasonLaster:
STR:

1. pause and expand an object
2. step
3. check the expanded object 

Expected: it has updated
Actual: it has not updated

@bomsy I think this is related to the actor cache

bomsy:
@jlast true! i'll have a peek.

arthur801031:
Just talked to @jasonLaster . I'll take a stab at this 🙂

bomsy:
Nice!

PR #2509, opened 2017-03-31
[Scopes] expanded properties are not updated while stepping

arthur801031:
Associated Issue: #2394

### Summary of Changes
- Assign new values to the cache actor if it goes stale

### Test Plan
1. pause and expand an object
2. step
3. check if expanded object has updated

kapture 2017-03-31 at 14 19 01


arthur801031:
Hi @jasonLaster 

The expanded properties would not be updated while stepping is because we're getting the properties from the cache actor instead of the current item's properties. The solution that I got fixes this issue but there is one drawback: we're not actually updating the cache actor; instead, we modify the cache actor locally and return this object to [getChildren](https://github.com/devtools-html/debugger.html/blob/master/src/components/shared/ObjectInspector.js#L126).

Please let me know if there is a better way to tackle this issue. 😄

src/utils/object-inspector.js
```diff
@@ -181,7 +181,19 @@ function getChildren(
...
+      if (properties.hasOwnProperty(pKey)) {
+        const cacheObject = thisActor.filter(a => a.name == pKey)[0];
+        const cacheObjectIndex = thisActor.findIndex(a => a.name == pKey);
+        // Assign new values to the cache actor if it becomes stale
+        if (cacheObject && getValue(cacheObject) != properties[pKey].value) {
+          thisActor[cacheObjectIndex].contents = properties[pKey];
+        }
+      }
+    }
+    return thisActor;
```
bomsy: Looking good!  If we move this change to the ObjectInspector.js [here](https://github.com/devtools-html/debugger.html/blob/master/src/components/shared/ObjectInspector.js#L125) instead.
Then we can do `this.props.setActors(actors);` after, which set the cache actor.

bomsy:
Looking good!

src/utils/object-inspector.js
arthur801031: Great suggestion! I thought using `this.props.setActors(actors);` would cause an error in `render`, but I was wrong! Thank you! 😊

src/components/Editor/Preview.js
```diff
@@ -100,6 +100,7 @@ class Preview extends Component {
       onDoubleClick: () => {},
       loadObjectProperties,
       getActors: () => ({}),
+      setActors: () => {},
```
bomsy: this will return `undefined`, should be `setActors: () => ({})` instead

src/components/SecondaryPanes/Expressions.js
```diff
@@ -162,6 +162,7 @@ class Expressions extends React.Component {
           this.editExpression(expression, options),
         loadObjectProperties,
         getActors: () => ({}),
+        setActors: () => {},
```
bomsy: same here

src/components/shared/ObjectInspector.js
```diff
@@ -122,6 +122,26 @@ const ObjectInspector = React.createClass({
   getChildren(item: ObjectInspectorItem) {
     const { getObjectProperties } = this.props;
     const { actors } = this;
+    const key = item.path;
```
bomsy: nit: this can probably move into this ` if (item.contents.value && item.contents.value.preview && actors[key]) {` line below, since is used only within it

src/components/shared/ObjectInspector.js
```diff
@@ -122,6 +122,26 @@ const ObjectInspector = React.createClass({
...
+
+    if (item.contents.value && item.contents.value.preview && actors[key]) {
```
bomsy: tiny note, just for security `... actors && actors[key]`

src/components/shared/ObjectInspector.js
```diff
@@ -122,6 +122,31 @@ const ObjectInspector = React.createClass({
...
+          // Assign new values to the cache actor if it becomes stale
+          if (
+            cacheObject && cacheObject.contents.value != properties[pKey].value
+          ) {
+            thisActor[cacheObjectIndex].contents = properties[pKey];
+          }
+        }
+      }
+      actors[key] = thisActor;
+      this.props.setActors(actors);
```
jasonLaster: I'm wondering if we can simplify the cache significantly.

The way i see it working is that it would keep a key of IDs that are expanded and that's it. This way, values don't need to be synced...

src/utils/object-inspector.js
```diff
@@ -180,8 +180,21 @@ function getChildren(
...
+      if (
+        properties[object.name] &&
+        !(properties[object.name] instanceof Function) &&
+        object.name != "__proto__"
+      ) {
+        thisActor[index].contents = properties[object.name];
+      }
+      index++;
+    });
+    return thisActor;
```
bomsy: adding `actors[key] = thisActor;` above this line should update actorCache

bomsy: Also , the `actors[key]` only goes stale across the `ObjectInspector` component loads, therefore across internal component renders we should still be fine doing `return actors[key]` if it exists.

arthur801031:
Hi @bomsy and @jasonLaster 

I've reverted to previous implementation where the code logic for detecting/updating `actors` is in `shared/ObjectInspector.js`. This has 2 benefits:
1. I think this implementation is more reliable than the one that was in `src/utils/object-inspector.js` because I had to use multiple `if conditions` to avoid setting function objects.
2. We can update `actors` in `shared/ObjectInspector.js`, but we can't do that in `src/utils/object-inspector.js`.

To @bomsy: 
I'm not sure if we can know when component loads (i.e during stepping) or internal components being render. What we have right now checks whether or not `actors` goes stale. If it does, we update `actors`.

bomsy:
ok looking and testing ..

src/components/shared/ObjectInspector.js
```diff
@@ -122,6 +122,36 @@ const ObjectInspector = React.createClass({
...
+            actorHasUpdated = true;
+          }
+        }
+      }
+
+      if (actorHasUpdated) {
+        actors[key] = thisActor;
+        this.props.setActors(actors);
+      }
+    }
```
jasonLaster: can this move to `utils/object-inspector` and get some unit tests?

src/utils/tests/object-inspector.js
```diff
@@ -221,4 +222,63 @@ describe("promises", () => {
     const node = getPromiseProperties(promise);
     expect(node.contents.value.type).to.eql("3");
   });
+
+  it("update actors when necessary", () => {
+    const item = {
```
jasonLaster: can we remove some of the data from this fixture so that it is much smaller. for instance, do we need the entire `value`? Can we remove `preview`'s contents?

src/utils/object-inspector.js
```diff
@@ -181,6 +181,22 @@ function getChildren(
   // node would be a new instance every render.
   const key = item.path;
   if (actors && actors[key]) {
+    if (item.contents.value && item.contents.value.preview) {
```
jasonLaster: can we move this code to an `updateActor` helper function?

arthur801031: No problem 🙂

src/utils/tests/object-inspector.js
arthur801031: Yes, please check my new commit :)

bomsy:
While testing this yesterday, noticed that if a property does not exist it does not get added.

scope-props-update

I think this will require us going back to get the `getProperties` again, which invalidates the expanded cache, 


I think we can land what we have atm, and improve on this.

PR #2520, opened 2017-04-01
Convert Editor/Breakpoint.js to ES6 class

peterschussheim:
Associated Issue: #1890 

# Summary of Changes
- Breakpoint.js
  - Convert to class syntax
  - Convert require statements to import syntax

### DIFF
--- a/src/components/Editor/Breakpoint.js
+++ b/src/components/Editor/Breakpoint.js
@@ -16,12 +16,7 @@ function makeMarker(isDisabled: boolean) {
   return bp;
 }
 
-const Breakpoint = React.createClass({
-  propTypes: {
-    breakpoint: PropTypes.object.isRequired,
-    editor: PropTypes.object.isRequired,
-  },
-  displayName: "Breakpoint",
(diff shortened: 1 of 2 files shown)


src/components/Editor/Breakpoint.js
```diff
@@ -66,10 +58,17 @@ const Breakpoint = React.createClass({
...
+}
+
+Breakpoint.propTypes = {
+  breakpoint: PropTypes.object.isRequired,
+  editor: PropTypes.object.isRequired,
+};
+
+Breakpoint.displayName = "Breakpoint";
 
 module.exports = Breakpoint;
```
jasonLaster: I think we can do `exports default Breakpoint`.

Check other examples for something like this

peterschussheim:
Confirmed breakpoint functionality works as intended. 

Regarding the `require("react-dom")`... any suggestions how to avoid this syntax?  I tried for hours with no luck to have the component work using `DOM as dom` from react.

Thanks!

PR #2522, opened 2017-04-02
Convert SearchInput to es modules and es6 class

irfanhudda:
Associated Issue: #1884 #1890

Summary of changes
* Convert `SearchInput` to es module and es6 class

### DIFF
--- a/src/components/shared/SearchInput.js
+++ b/src/components/shared/SearchInput.js
@@ -1,33 +1,17 @@
-const React = require("react");
-const { DOM: dom, PropTypes } = React;
-const classnames = require("classnames");
-const Svg = require("./Svg");
-const CloseButton = require("./Button/Close").default;
+import { DOM as dom, PropTypes, Component } from "react";
+import Svg from "./Svg";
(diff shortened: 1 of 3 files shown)


src/components/shared/SearchInput.js
```diff
@@ -1,33 +1,17 @@
...
-const Svg = require("./Svg");
-const CloseButton = require("./Button/Close").default;
+import { DOM as dom, PropTypes, Component } from "react";
+import Svg from "./Svg";
+import classnames from "classnames";
+import CloseButton from "./Button/Close";
+import "./SearchInput.css";
 
-require("./SearchInput.css");
+class SearchInput extends Component {
```
irfanhudda: I think we can change this to `export default class SearchInput extends Component {` and remove the last line (which has export statement).

jasonLaster: I like that idea!

jasonLaster: Lets do a new PR where we change all of these cases

Issue #2267, opened 2017-03-03
[FunctionSearch] Not updating selected item on click


AlanCezarAraujo:
Whan I click on an item from the result list, it navigates to the item in the source panel, but it doesn't update the selected item in the result list.

AlanCezarAraujo:
I'm already wornking to fix it.

wldcordeiro:
Can you describe what you mean by not updating in the result list?

wldcordeiro:
@AlanCezarAraujo I see what you mean now.

Here's a patch that should do what we expect.

```patch
diff --git a/src/components/Editor/SearchBar.js b/src/components/Editor/SearchBar.js
index d1a5add..c85f379 100644
--- a/src/components/Editor/SearchBar.js
+++ b/src/components/Editor/SearchBar.js
@@ -322,11 +322,14 @@ const SearchBar = React.createClass({
   },
 
   // Handlers
-  selectResultItem(item: SymbolDeclaration) {
+  selectResultItem(item: SymbolDeclaration, index: number, e: SyntheticEvent) {
     const { selectSource, selectedSource } = this.props;
     if (selectedSource) {
+      this.setState({ selectedResultIndex: index });
       selectSource(
         selectedSource.get("id"), { line: item.location.start.line });
+      this.closeSearch(e);
     }
   },
 
diff --git a/src/components/shared/ResultList.js b/src/components/shared/ResultList.js
index e4a25c7..bd61838 100644
--- a/src/components/shared/ResultList.js
+++ b/src/components/shared/ResultList.js
@@ -24,7 +24,7 @@ const ResultList = React.createClass({
   renderListItem(item: ResultListItem, index: number) {
     return dom.li(
       {
-        onClick: () => this.props.selectItem(item),
+        onClick: e => this.props.selectItem(item, index, e),
         key: `${item.id}${item.value}${index}`,
         ref: index,
         title: item.value,
```

AlanCezarAraujo:
@wldcordeiro thanks for your help!  
I went on a different approach but you gave me a more elegant solution.  
I'll open a PR as soon as I fix some Flow issues.

wldcordeiro:
@AlanCezarAraujo you may want to leave `selectResultItem` itself alone as it's used by the keyDown handler as well and just modify this [line](https://github.com/devtools-html/debugger.html/blob/master/src/components/Editor/SearchBar.js#L486) like so

```js
selectItem(e, item, index) {
[...]

PR #2524, opened 2017-04-03
Fix 2267

AlanCezarAraujo:
Associated Issue: #2267

### Summary of Changes

* On **selectResultItem** is now possible to pass the event as a parameter to close the result list

### Test Plan

- [x] Open a large file
- [x] Ctrl+f to find
- [x] Select **function** from filter
- [x] Enter the text **add**
- [x] Click on any item
- [x] The result list show close after the click

### DIFF
--- a/src/components/Editor/SearchBar.js
+++ b/src/components/Editor/SearchBar.js
@@ -402,7 +402,7 @@ const SearchBar = React.createClass({
   },
 
   // Handlers
-  selectResultItem(e: SyntheticEvent, item: SymbolDeclaration, index?: number) {
+  selectResultItem(e: SyntheticEvent, item: SymbolDeclaration) {
     const { selectSource, selectedSource } = this.props;
 
     if (selectedSource) {
--- a/src/components/Editor/SearchBar.js
+++ b/src/components/Editor/SearchBar.js
@@ -54,42 +54,6 @@ type ToggleSymbolSearchOpts = {
 
 require("./SearchBar.css");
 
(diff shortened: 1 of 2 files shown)


src/components/Editor/SearchBar.js
```diff
@@ -399,12 +402,15 @@ const SearchBar = React.createClass({
   },
 
   // Handlers
-  selectResultItem(item: FormattedSymbolDeclaration) {
+  selectResultItem(e: SyntheticEvent, item: SymbolDeclaration, index?: number) {
```
jasonLaster: Hmm, my only question is around `FormattedSymbolDeclaration` => `SymbolDeclaration`. Why was this needed?

AlanCezarAraujo: Once I only needed the item's location, `FormattedSymbolDeclaration` would expect more data than  I needed, and it could also break the _line length_ lint rule.

irfanhudda: For _line length_ issue we can always wrap arguments to different lines like https://github.com/devtools-html/debugger.html/blob/master/src/components/Editor/ConditionalPanel.js#L10

I wonder why we have `index` function parameter. It doesn't appear anywhere in function.

AlanCezarAraujo: @irfanhudda thanks for the tip about line length!
The `index` was dead code. I just removed it.

jasonLaster:
Thanks @AlanCezarAraujo!

Issue #2398, opened 2017-03-17
[EditorPreview] Don’t preview variables outside of the frame


jasonLaster:
Currently, we will preview variables regardless of whether they are in the selected frame. The problem with this is that you often get values that you would not expect.

For instance, if you pause in a function with "this" and then jump to another module with "this" hovering will give you the value from the selected frame.

We can fix this, by first checking the context (source id, line, location) of the token we're hovering on and seeing if it is within the selected frame.

The primary challenge here is to get the code range for the frame. The pause data has the start of the frame and outer scope, but it doesn't have the end. This is where babel comes in because we can get the scope start and end locations from the AST :)

![](http://g.recordit.co/snLzI3LGSp.gif)

ryanjduffy:
Looking into this one

ryanjduffy:
What do you think of bailing on `previewSelectedToken()` if the selected frame isn't for the selected source? That's a relatively easy first step, at least.

```
    if (!selectedFrame || !sourceText || !isEnabled("editorPreview") ||
        selectedFrame.location.sourceId !== selectedSource.get("id")) {
      return;
    }
```

ryanjduffy:
Also noticed that `this` inspection isn't working because `selectedFrame` isn't passed to `previewExpression`. @jasonLaster Are you seeing the same?

jasonLaster:
Hmm - i think that i fair, but shouldn't we always be looking at the selected source?

PR #2530, opened 2017-04-03
[EditorPreview] Prevent previewing tokens out of scope of the selected frame

ryanjduffy:
Associated Issue: #2398

### Summary of Changes

* Add `resolveToken()` to parser worker which resolves a token to its expression and determines if the token is in scope of the given frame
* Replace `getExpressionFromToken()` in `utils/Editor` with `resolveToken`

### Test Plan

- [x] Set a breakpoint within a function
- [x] Trigger the breakpoint
- [x] Hovering tokens within the function should preview
- [x] Hovering tokens outside the function or in another file should not preview

debugger html-2398


src/utils/parser/utils.js
```diff
@@ -228,24 +255,60 @@ function getExpression(
...
+    enter(path) {
+      // if we haven't found an expression yet, determine if the token is part
+      // of one
+      if (!expression) {
+        expression = resolveExpression(path, token, location);
+      }
+
+      // determine if the current path is a function or program containing the
+      // frame
+      const scope = resolveScope(path, frame.location);
```
jasonLaster: could we do 

```js
let scope = null
```

and then below:

```js
scope = resolveScope(path, frame.location);
```

src/utils/parser/utils.js
```diff
@@ -228,24 +255,60 @@ function getExpression(
...
+
+  if (isEmpty(ast)) {
+    return;
+  }
+
+  traverse(ast, {
+    enter(path) {
+      // if we haven't found an expression yet, determine if the token is part
+      // of one
+      if (!expression) {
```
jasonLaster: why `if (!expression) {`

src/utils/parser/utils.js
```diff
@@ -228,24 +255,60 @@ function getExpression(
...
+        scopes.unshift(scope);
+      }
+    },
+  });
+
+  let inScope = false;
+  if (scopes.length > 0) {
+    // determine if the narrowest scope contains the token's location
+    inScope = nodeContainsLocation({ node: scopes[0].node, location });
+  }
```
jasonLaster: hmm, could we get to this point? we should always have a scope, no?

```js
const inScope = nodeContainsLocation({ node: scope.node, location })
```

src/components/Editor/index.js
```diff
@@ -313,30 +319,33 @@ const Editor = React.createClass({
       this.setState({ selectedToken: null, selectedExpression: null });
     }
 
-    const expressionFromToken = await getExpressionFromToken(
+    const resolution = await resolveToken(
```
jasonLaster: destructuring ftw 

```js
const { expression, inScope } = await resolveToken(
```

src/components/Editor/index.js
```diff
@@ -313,30 +319,33 @@ const Editor = React.createClass({
...
+      sourceText,
+      selectedFrame
     );
 
-    if (expressionFromToken) {
-      addExpression(expressionFromToken.value, { visible: false });
-    }
-
-    const variables = getVisibleVariablesFromScope(pauseData, selectedFrame);
+    if (resolution.inScope) {
```
jasonLaster: could we bail early?

```js
if (!inScope) {
return;
}
```

src/utils/parser/utils.js
ryanjduffy: There's no need to continue looking for an `expression` if we've found one, right? Maybe I'm misunderstanding how the AST traversal happens, though.

src/utils/parser/utils.js
ryanjduffy: Sure. Do you prefer all the variables declared at the top of the function?

src/utils/parser/utils.js
ryanjduffy: We should. I'd assume there would always be a `Program` scope but I was being defensive. I can remove the check.

src/components/Editor/index.js
ryanjduffy: Yup

ryanjduffy:
@jasonLaster Updated with code style fixes and tests for `resolveToken`. Do you think I should remove `getExpression()` since it is no longer used (as far as I can tell, at least)? Do you have future plans for it?

jasonLaster:
💥  Awesome!

PR #2531, opened 2017-04-03
Prettier improvements

jasonLaster:
Associated Issue: #2362 

### Summary of Changes

* continues to improve commas
* improve match
* cleanup eslintrc
* it is still missing things :/

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "start-app": "TARGET=application node bin/dev-server",
     "flow": "flow",
     "eslint-check": "eslint --print-config .eslintrc.js | eslint-config-prettier-check",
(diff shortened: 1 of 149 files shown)


package.json
```diff
@@ -104,8 +104,8 @@
     "remark-validate-links": "^6.0.0"
   },
   "lint-staged": {
-    "src/!(mochitest)**/*.js": [
-      "prettier --write --trailing-comma=es5 --bracket-spacing=true",
+    "src/*/!(mochitest)**/*.js": [
```
bomsy: the glob pattern is causing it to miss files.

bomsy: i've tried a couple of patterns, any removing the mochitest leaves out some files. should we just do every thing i.e `src/**/*.js` ?

PR #2533, opened 2017-04-03
[FunctionSearch] Select result item on click

irfanhudda:
### STR
1. Open function search.
2. Click on a result item.
Bug

### Changes
* Add index to `onClick` handler of `ResultList`

### Screenshot
fix

### DIFF
--- a/src/components/Editor/SearchBar.js
+++ b/src/components/Editor/SearchBar.js
@@ -143,7 +143,11 @@ const SearchBar = React.createClass({
       searchInput.focus();
     }
 
-    if (this.refs.resultList && this.refs.resultList.refs) {
+    if (
+      this.refs.resultList &&
+      this.refs.resultList.refs &&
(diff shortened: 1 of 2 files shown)


irfanhudda:
Found a problem with this PR so I'll close this for now

irfanhudda:
After looking into the problem it turn out that there was another bug

### STR
1. Open TodoMVC example in browser
2. Open `js/views/app-view.js` file in debugger.html
3. Open function search in editor and enter `a`.
4. Navigate to lower result by pressing `down` arrow key
5. Now enter `a` in function search to narrow the result
Bug

This problem was because we are trying to scroll to result that don't exist anymore.

jasonLaster:
Is this a dupe #2524?

irfanhudda:
@jasonLaster Just realised that. Closing this for now.

Issue #1843, opened 2017-01-25
[EditorPreview] Preview values in the popup


jasonLaster:
Here is a list of things we should do to improv editor hover

- [x] show the popup on hover #2156
- [x] highlight the selected expression/token #2157
- [x] use a header #2158
- [x] have a caret in the top left of the popup #2159
- [x] support dark theme #2160
- [x] jump to definition #2161
- [x] additional types of expressions
  - [x] preview member expressions #2219
  - [x] preview variables from outer scopes #2220
- [x] debounce opening and closing it #2226
- [x] styling tweaks (add padding, and other styling tweaks) #2227

----
#### Use a header

We currently show the selected token as the top level root. We should move the label out of the tree and put it in a header. The header can also have a jump link for functions :)

The hard part here is to make all of the properties roots, but it should be doable by looking at the Scopes component.

#### highlight the selected token

We can change `selectedToken` to be the element selected and then add/remove the class in `renderPreview`

```js
selectedToken.classList.add("selected")
```

![screen shot 2017-02-26 at 3 31 31 pm](https://cloud.githubusercontent.com/assets/254562/23343392/b95124d6-fc38-11e6-802f-1e9f4fd54975.png)

#### additional types of expressions
  - 6.1 support member expressions like `obj.foo.bar`
  - 6.2 support `this` and maybe other in scope things
  - 6.3 support showing outer scope bindings
  - 6.4 support variables... if they're not in the bindings

#### jump to definition

We only need this for functions. Functions have their defined `location` in the data and we can use the `selectSource` action to do the jump

```json
{
  "enumerable": true,
  "configurable": false,
  "value": {
    "frozen": false,
    "displayName": "app.TodoView<.render",
    "actor": "server2.conn8.child1/pausedobj100",
    "location": {
      "url": "https://devtools-html.github.io/debugger-examples/examples/todomvc/js/views/todo-view.js",
      "line": 39
    },
    "class": "Function",
    "type": "object",
    "extensible": true,
    "sealed": false,
    "parameterNames": []
  },
  "writable": true
}
```

jasonLaster:
We're landing a V1 pretty soon that is limited. Here is a list of things we should do to improve it:

1. show the popup on hover
2. highlight the selected expression/token
3. show the token/expression as a header instead of a top level root
4. have a caret in the top left of the popup
5. support dark theme

clarkbw:
All of the required items have been checked off here.

PR #2537, opened 2017-04-04
Fixup preview

jasonLaster:
Associated Issue: #1843

### Summary of Changes

* Fix previewing `this`
* Fix previewing native functions like apply
* Fix previewing functions that don't have a location

### DIFF
--- a/src/components/shared/previewFunction.js
+++ b/src/components/shared/previewFunction.js
@@ -12,8 +12,9 @@ function renderFunctionName(value) {
 
 function renderParams(value) {
   const { parameterNames } = value;
-  let params = parameterNames.map(param =>
-    dom.span({ className: "param" }, param));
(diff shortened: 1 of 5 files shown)


src/components/shared/previewFunction.js
```diff
@@ -12,8 +12,9 @@ function renderFunctionName(value) {
 
 function renderParams(value) {
   const { parameterNames } = value;
-  let params = parameterNames.map(param =>
-    dom.span({ className: "param" }, param));
+  let params = parameterNames
+    .filter(i => i)
```
clarkbw: I don't understand.  What does this filter check for?  That `i` exists?

jasonLaster: yeah - with native functions params are null. I can add a comment

PR #2551, opened 2017-04-06
FunctionSearch - show index in summary, Fixes #2529

### LINKED ISSUE #2529: [FunctionSearch] show index in the summary
We currently show a summary like "50 results" when you're searching for a function or variable, but we don't include the index like "2 of 50 results". It would be nice to do this so that as you iterate through large lists you have some additional context.

### Patch
```diff
diff --git a/src/components/Editor/SearchBar.js b/src/components/Editor/SearchBar.js
index 7a4c581..3811f6f 100644
--- a/src/components/Editor/SearchBar.js
+++ b/src/components/Editor/SearchBar.js
@@ -473,6 +473,8 @@ const SearchBar = React.createClass({

   // Renderers
   buildSummaryMsg() {
+    // we might be able to do this by removing this if statement...
+    // and getting the count and index from the state.
     if (this.state.symbolSearchEnabled) {
       return L10N.getFormatStr(
         "sourceSearch.resultsSummary1",
```

ruturajv:
Associated Issue: #2529

### Summary of Changes

- Changed Summary message to `%d of %d results`

@wldcordeiro @jasonLaster  - What should be done for a `1 of 1 results` output?

### DIFF
--- a/assets/panel/debugger.properties
+++ b/assets/panel/debugger.properties
@@ -146,7 +146,7 @@ sourceSearch.search.again.key=G
 
 # LOCALIZATION NOTE (sourceSearch.resultsSummary1): Shows a summary of
 # the number of matches for autocomplete
-sourceSearch.resultsSummary1=%d of %d results
+sourceSearch.resultsSummary1=%d results
 
 # LOCALIZATION NOTE (noMatchingStringsText): The text to display in the
 # global search results when there are no matching strings after filtering.
--- a/assets/panel/debugger.properties
+++ b/assets/panel/debugger.properties
@@ -239,7 +239,7 @@ callStack.expand=Expand Rows
 # for the summarizing the selected search result. e.g. 5 of 10 results.
 editor.searchResults=%d of %d results
 
-# LOCALIZATION NOTE (sourceSearch.singleResult):
+# LOCALIZATION NOTE (sourceSearch.singleResult): Copy shown when there is one result.
 editor.singleResult=1 result
 
 # LOCALIZATION NOTE (editor.noResults): Editor Search bar message
--- a/assets/panel/debugger.properties
+++ b/assets/panel/debugger.properties
@@ -239,6 +239,9 @@ callStack.expand=Expand Rows
 # for the summarizing the selected search result. e.g. 5 of 10 results.
 editor.searchResults=%d of %d results
 
+# LOCALIZATION NOTE (sourceSearch.singleResult):
+editor.singleResult=1 result
+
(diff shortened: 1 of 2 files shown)


assets/panel/debugger.properties
```diff
@@ -146,7 +146,7 @@ sourceSearch.search.again.key=G
 
 # LOCALIZATION NOTE (sourceSearch.resultsSummary1): Shows a summary of
 # the number of matches for autocomplete
-sourceSearch.resultsSummary1=%d results
+sourceSearch.resultsSummary1=%d of %d results
```
wldcordeiro: Maybe just use `editor.searchResults` property since it's already formatted this way?

ruturajv: ya, that's better.

wldcordeiro:
I think `1 of 1 results` isn't bad, the Firefox search bar (<kbd>ctrl</kbd>+<kbd>F</kbd>) shows similar (though it has `1 of 1 match`/`2 of 2 matches`)

src/components/Editor/SearchBar.js
```diff
@@ -473,9 +473,10 @@ const SearchBar = React.createClass({
 
   // Renderers
   buildSummaryMsg() {
-    if (this.state.symbolSearchEnabled) {
+    if (this.state.symbolSearchResults.length > 0) {
```
wldcordeiro: I think you still want to check if symbol search is enabled here in addition to this condition.

wldcordeiro: Also, I'm not 100% certain but you maybe able to remove this [block](https://github.com/ruturajv/debugger.html/blob/53564fb44c23400c86e6abf9212a5e600d3918b3/src/components/Editor/SearchBar.js#L493) with the change.

src/components/Editor/SearchBar.js
```diff
@@ -473,9 +473,13 @@ const SearchBar = React.createClass({
 
   // Renderers
   buildSummaryMsg() {
-    if (this.state.symbolSearchEnabled) {
+    if (
+      this.state.symbolSearchEnabled &&
+      this.state.symbolSearchResults.length > 0
```
jasonLaster: agreed - i think we should only show it if it's greater than 1

ruturajv: So for 
- 1 result - `1 of 1 result` or `1 result`
- 1+ results - `%d of %d results`
?
@wldcordeiro  @jasonLaster

jasonLaster: Hmm, I'd just show nothing :)

PR #2560, opened 2017-04-06
Refactor/editor/HitMarker

peterschussheim:
Associated Issue: #1890 

## Summary of Changes

HitMarker.js
- [x] Convert to class syntax
- [x] Convert require statements to import syntax

components/Editor/index.js
- [x] Require default module from HitMarker

### DIFF
--- a/src/components/Editor/HitMarker.js
+++ b/src/components/Editor/HitMarker.js
@@ -1,6 +1,5 @@
 // @flow
-const React = require("react");
-const { PropTypes } = React;
+import { Component, PropTypes } from "react";
 
 let markerEl = document.createElement("div");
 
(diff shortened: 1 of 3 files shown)


src/components/Editor/HitMarker.js
```diff
@@ -10,13 +9,8 @@ function makeMarker() {
...
 
-const HitMarker = React.createClass({
-  propTypes: {
-    hitData: PropTypes.object.isRequired,
-    editor: PropTypes.object.isRequired
-  },
-
-  displayName: "HitMarker",
+class HitMarker extends Component {
+  displayName: "HitMarker";
```
AnshulMalik: We keep displayName outside of class declaration just like propTypes.

peterschussheim: Thanks for catching that!  Just pushed an update.

jasonLaster:
Awesome!

PR #2573, opened 2017-04-07
Fix/lint staged prettier

bomsy:
### Associated Issue

The issue is that lint-staged was not watching most of our prettier paths to automatically 
make staged files in those files pretty.

### Summary of Changes

* Add lint-staged tasks to support all our prettier paths
* Use the npm tasks instead
* lint-staged does not seem to have a concise way of grouping multiple glob patterns atm .


package.json
```diff
@@ -105,10 +105,11 @@
...
   "lint-staged": {
-    "src/*/!(mochitest)**/*.js": [
-      "node bin/prettier.js",
-      "git add"
-    ]
+    "*.js": ["prettier", "git add"],
+    "src/*.js": ["prettier", "git add"],
+    "src/*/*.js": ["prettier", "git add"],
+    "src/*/!(mochitest)**/*.js": ["prettier", "git add"],
+    "src/*/!(mochitest)*/**/*.js": ["prettier", "git add"]
```
jasonLaster: could we do a catch all like `**/*.js` which would pass everything to prettier?

bomsy: This won't ignore changes changes to the mochitests

jasonLaster: ahh

PR #2581, opened 2017-04-07
Move locale to devtools/shared

jasonLaster:
We would like to share the locales with the debugger server for highlighting the pause reason. This necessitates moving locales to devtools/shared.

### DIFF
--- a/assets/panel/panel.js
+++ b/assets/panel/panel.js
@@ -6,7 +6,7 @@
 const { Task } = require("devtools/shared/task");
 var { LocalizationHelper } = require("devtools/shared/l10n");
 
-const DBG_STRINGS_URI = "devtools/client/locales/debugger.properties";
+const DBG_STRINGS_URI = "devtools/shared/locales/debugger.properties";
 var L10N = new LocalizationHelper(DBG_STRINGS_URI);
 
 function DebuggerPanel(iframeWindow, toolbox) {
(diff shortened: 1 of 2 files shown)


jasonLaster:
Hmm, @captainbrosset I'm beginning to reconsider the decision to move the debugger properties file:

1. there is already a [properties file][file] in shared that will need to be merged 
2. there are other components that rely on [client/debugger.properties][search] such as variables view and the old debugger

perhaps we add a simple string in the overlay like "Paused in the debugger" and work on extracting shared pause reasons at a later point?


[file]: http://searchfox.org/mozilla-central/source/devtools/shared/locales/en-US/debugger.properties
[search]: http://searchfox.org/mozilla-central/search?q=debugger.properties&case=true&path=

captainbrosset:
> there is already a properties file in shared that will need to be merged

Yeah, I saw that, I thought maybe we could share the 2 files and it wouldn't be a big deal.

> there are other components that rely on client/debugger.properties such as variables view and the old debugger

Right, I didn't know that. I suspect it would be enouhg to just change all instances of `devtools/client/locales/debugger.properties` with `devtools/shared/locales/debugger.properties` though.

> I'm not sure what the localization overhead of moving locale files is

I have no idea. That's a good point. I think localization might have to be redone entirely because of that. Of course if we tell localizers that it's just a file move, they might be able to just copy translations across. But that point makes the whole idea much more complex indeed.

So, although I think 1 and 2 are simple enough to fix, I agree that 3 is unknown and might take time. So I'll add just a simple string for now in the shared locale and add a TODO in the code so we can do this later. Works for me!

Thanks for taking a look at that!

jasonLaster:
Thanks for the quick reply. I'll close this for now and we can re-open it later

PR #2582, opened 2017-04-07
(WIP) 4-7 update

jasonLaster:
Regular Update with some changes:

* updated some properties
* assets manifest
* updated some feature flags

🛑  lets not merge this until firefox is updated

### DIFF
--- a/assets/panel/panel.js
+++ b/assets/panel/panel.js
@@ -16,7 +16,7 @@ function DebuggerPanel(iframeWindow, toolbox) {
 }
 
 DebuggerPanel.prototype = {
-  open: async function () {
+  open: async function() {
     if (!this.toolbox.target.isRemote) {
       await this.toolbox.target.makeRemote();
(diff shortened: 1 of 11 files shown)


assets/panel/panel.js
```diff
@@ -41,27 +41,27 @@ DebuggerPanel.prototype = {
...
     return {
       store: this._store,
       selectors: this._selectors,
       actions: this._actions,
-      client: this._client,
+      client: this._client
     };
   },
 
-  _getState: function () {
```
jasonLaster: we should make sure the spaces stay...

assets/panel/panel.js
```diff
@@ -41,27 +41,27 @@ DebuggerPanel.prototype = {
     return this;
   },
 
-  getVarsForTests () {
```
jasonLaster: we should update this function to be the classic style

PR #2583, opened 2017-04-07
Fix BP conditional check

jasonLaster:
I've been seeing yellow borders for BPs when they're not conditional for some time and have forgotten to make an issue :/

screen shot 2017-04-07 at 4 10 19 pm


jasonLaster:
Thanks @ryanjduffy

Issue #2565, opened 2017-04-06
[PreviewPopup] Position it intelligently


jasonLaster:
The preview is always below the variable… even when showing it below will mean it’ll get clipped off by the bottom of the window. It’d be great to measure it’s position and sometimes place it on top of the var if that’ll mean it doesn’t get clipped

![](https://cloud.githubusercontent.com/assets/254562/24757054/04a68b6a-1aac-11e7-99e9-713394f0be67.png)

```diff
diff --git a/src/components/shared/Popover.js b/src/components/shared/Popover.js
index e158d65..dae7bea 100644
--- a/src/components/shared/Popover.js
+++ b/src/components/shared/Popover.js
@@ -26,6 +26,9 @@ class Popover extends Component {
     const left = targetLeft + targetWidth / 2 - width / 5;
     const top = targetBottom;

+    // We should also meaure the `window.getBoundingClientRect` and see
+    // if we have space to place the popover below the var
+
     // eslint-disable-next-line react/no-did-mount-set-state
     this.setState({ left, top });
   }
```

bomsy:
@jlast is this complete?

jasonLaster:
Yep

PR #2599, opened 2017-04-08
WIP Display popover intelligently

AnshulMalik:
Associated Issue: #2565 

### Summary of Changes

* Added popover direction

### Screenshots
as

### DIFF
--- a/src/components/shared/Popover.js
+++ b/src/components/shared/Popover.js
@@ -15,27 +15,35 @@ class Popover extends Component {
 
   componentDidMount() {
     const { type } = this.props;
-    const { left, top } = type == "popover"
+    const { left, top, dir } = type == "popover"
       ? this.getPopoverCoords()
       : this.getTooltipCoords();
 
     // eslint-disable-next-line react/no-did-mount-set-state
-    this.setState({ left, top });
+    this.setState({ left, top, dir });
   }
 
   getPopoverCoords() {
     const el = ReactDOM.findDOMNode(this);
-    const { width } = el.getBoundingClientRect();
+    const { width, height } = el.getBoundingClientRect();
     const {
       left: targetLeft,
       width: targetWidth,
-      bottom: targetBottom
+      bottom: targetBottom,
+      top: targetTop
     } = this.props.target.getBoundingClientRect();
+
+    let dir = "down";
 
     // width division corresponds to calc in Popover.css
     let left = targetLeft + targetWidth / 2 - width / 5;
     let top = targetBottom;
-    return { left, top };
+    if (top + height > window.innerHeight) {
+      console.log("Going out of bounds");
+      dir = "up";
+      top = targetTop - height;
+    }
+    return { left, top, dir };
   }
 
   getTooltipCoords() {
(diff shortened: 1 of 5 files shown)


src/components/shared/Popover.js
```diff
@@ -50,12 +58,24 @@ class Popover extends Component {
     const left = targetLeft + targetWidth / 4;
     const top = targetTop - height;
 
-    return { left, top };
+    return { left, top, dir: "up" };
```
jasonLaster: Nice.

We might also want to do left/right here in a bit

src/components/shared/Popover.js
```diff
@@ -15,27 +15,35 @@ class Popover extends Component {
...
     } = this.props.target.getBoundingClientRect();
 
+    let dir = "down";
+
     // width division corresponds to calc in Popover.css
     let left = targetLeft + targetWidth / 2 - width / 5;
     let top = targetBottom;
-    return { left, top };
+    if (top + height > window.innerHeight) {
+      console.log("Going out of bounds");
```
jasonLaster: lets remove this...

AnshulMalik: Oops, I forgot.

src/components/shared/Popover.js
```diff
@@ -50,12 +53,24 @@ class Popover extends Component {
...
-    return { left, top };
+    return { left, top, dir: "up" };
   }
 
   renderPopover() {
     const { children, onMouseLeave } = this.props;
-    const { top, left } = this.state;
+    const { top, left, dir } = this.state;
+
+    if (dir === "up") {
```
jasonLaster: lets see if we can consolidate these two renders into one...

src/components/shared/Popover.js
```diff
@@ -50,12 +53,24 @@ class Popover extends Component {
...
+      return dom.div(
+        {
+          className: "popover popover-up",
```
jasonLaster: we can use the classnames helper to do `classnames("popover", { up: dir =="up" })`

src/components/shared/Popover.js
```diff
@@ -50,12 +53,24 @@ class Popover extends Component {
...
+          onMouseLeave,
+          style: { top, left }
+        },
+        children,
```
jasonLaster: the ordering here is the hard one, but we can likely do something clever :)

src/components/shared/Popover.js
AnshulMalik: Didn't know about that. Yeah, we can use this.

jasonLaster:
Hey @AnshulMalik I made some additional follow up tweaks

1. generalized `getChildren` so popovers and tooltips had gaps
2. refactored the CSS so there were popovers and tooltips
3. removed the tooltip caret as the tooltips are much more lightweight interactions 
4. moved the preview function css to its own stylesheet

Let me know what you think and we can land this soon.


Here's a GIF of how it looks now:



I think it's looking pretty good! The only follow up item I see is that we will want to also adjust the horizontal (left/right ) orientation as well. But, lets do this in a different a new issue!

AnshulMalik:
Looks good to me.

Issue #2555, opened 2017-04-06
[FunctionSearch] 'Search for:' in search filter bar not localizable/hardcoded/always in English


Archaeopteryx:
Firefox 55.0a1 20170406030206 on Windows 8.1, also with latest 54.0a2

The 'Search for:' label in the search filter bar is hardcoded and can't be localized:

`return dom.section({ className: "search-type-toggles" }, dom.h1({ className: "search-toggle-title" }, "Search for:"), dom.button({`

PR #2604, opened 2017-04-09
Localize search type toggle title

irfanhudda:
Fixes: #2555 

### Summary of Changes

* Localize search type toggle title

I think the name and localization note can be improved. Let me know if you have any suggestions.


jasonLaster:
I think this is fine.

Issue #2378, opened 2017-03-15
[FunctionSearch] Fails when the source is HTML


jasonLaster:
We currently pass an HTML file to the babel parser when we try to find the function names, which causes it to fail.

![dcae7d9a-0979-11e7-8501-e1fcd940adaf.png](https://shipusercontent.com/3a340adb14775f2f8c749a3f0e82c1bc/dcae7d9a-0979-11e7-8501-e1fcd940adaf.png)

CC @julienw

jasonLaster:
One good option here is to check the context type before we try to parse the code and if it is not parseable a.k.a HTML we should not try. In these cases we should hide the "Search For" container

jasonLaster:
Here's a decent patch that should mostly work. I included a small refactor to make the code a bit more readable.

```diff
diff --git a/src/components/Editor/SearchBar.js b/src/components/Editor/SearchBar.js
index 2d8663c..10fd4f6 100644
--- a/src/components/Editor/SearchBar.js
+++ b/src/components/Editor/SearchBar.js
@@ -24,6 +24,7 @@ const {
 } = require("../../utils/editor");
 const { getSymbols } = require("../../utils/parser");
 const { scrollList } = require("../../utils/result-list");
+const { isJavaScript } = require("../../utils/source");
 const classnames = require("classnames");
 const debounce = require("lodash/debounce");
 const SearchInput = createFactory(require("../shared/SearchInput"));
@@ -32,6 +33,17 @@ const ImPropTypes = require("react-immutable-proptypes");
 
 import type { FormattedSymbolDeclaration } from "../../utils/parser/utils";
 
+async function getSearchResults(sourceText, query, selectedSymbolType) {
+  if (isJavaScript(sourceText)) {
+    return [];
+  }
+
+  const symbolDeclarations = await getSymbols(sourceText.toJS());
+  return filter(symbolDeclarations[selectedSymbolType], query, {
+    key: "value",
+  });
+}
+
 function getShortcuts() {
   const searchAgainKey = L10N.getStr("sourceSearch.search.again.key");
   const fnSearchKey = L10N.getStr("symbolSearch.search.key");
@@ -289,15 +301,14 @@ const SearchBar = React.createClass({
       return;
     }
 
-    const symbolDeclarations = await getSymbols(sourceText.toJS());
-    const symbolSearchResults = filter(
-      symbolDeclarations[selectedSymbolType],
+    const searchResults = await getSearchResults(
[...]

PR #2607, opened 2017-04-10
Only parse js

jasonLaster:
@ryanjduffy will work on parsing HTML in a bit here https://github.com/devtools-html/debugger.html/issues/2378, but for now lets limit babel to just things it can do :)

@jbhoosreddy we will likely want to include JSX and typescript too later


bomsy:
Lgtm

Issue #2588, opened 2017-04-07
CodeFolding


jasonLaster:
We added basic code folding in #2468, but it is missing some functionality before it can land:

- [ ] hide fold markers unless hovering on the gutter
- [ ] style the fold markers like Atom
  - [x] update the marker styles
  - [ ] style closed markers like Atom
- [ ] highlight the folded markers specially
- [ ] paused lines get clipped by the fold gutter
- [ ] clicking in the fold gutter should add a breakpoint if there isn't a fold marker...

### How to test folding

Go to the settings panel and enable "code folding"

### Styling closed markers

1. use the devtools orange theme
2. update the devtools collapse icon to be similar to atoms...
  note cmd+option+i in atom opens atom's devtools! you can steal their collapse badgey thing 

<img width="633" alt="screen shot 2017-04-11 at 3 39 00 pm" src="https://cloud.githubusercontent.com/assets/254562/24927463/21cb33e2-1ecd-11e7-8067-9d9c7bb305bd.png">

jasonLaster:
I think we could do something where you hide the fold markers by default and only show them if the gutter is hovered `gutter::hover .fold-marker { display: block } ` if that doesn't work, on gutter hover we add a class the the fold marker parent element...

AnshulMalik:
I did some experimentation to implement ` hide fold markers unless hovering on the gutter`. But seems like using only css doesn't work.

irfanhudda:
@AnshulMalik Yes. Looks like using just CSS to "hide _all_ fold markers unless hovering on gutter" may not work because of their placement in DOM tree and limitation of CSS selectors.(I might be wrong here)

I could get a crude version working using `eventListeners`

jasonLaster:
@irfanhudda mind PRing the crude version.

@AnshulMalik I'll review your styling independently

wldcordeiro:
> Yes. Looks like using just CSS to "hide all fold markers unless hovering on gutter" may not work because of their placement in DOM tree and limitation of CSS selectors.(I might be wrong here

@irfanhudda I was trying to do this as well, it's complicated. The "gutter" itself lives at a different level in the tree than the markers. I think there  is likely some really specific CSS selector like (pseudo selectors here) `.gutter:hover ~ .gutter-elems .hover-marker`. If you find a good solution I'd love to see it! 😄

jasonLaster:
[...]

PR #2610, opened 2017-04-10
[WIP] CodeFolding - hide fold markers unless hovering on the gutter

irfanhudda:
Associated Issue: #2588 

### Summary of Changes

* Add event listener on code mirror gutter to toggle visibility of fold marker

curde_ver

The problems here is `mouseenter` and `mouseleave` doesn't work as expected on gutter. I'm not sure why.

Let me know if you have any suggestions.

### DIFF
--- a/src/components/Editor/index.js
+++ b/src/components/Editor/index.js
@@ -156,6 +156,20 @@ class Editor extends Component {
     codeMirrorWrapper.addEventListener("mouseup", e => this.onMouseUp(e, ctx));
     codeMirrorWrapper.addEventListener("mouseover", e => this.onMouseOver(e));
 
+    const toggleFoldMarkerVisibility = e => {
+      if (node instanceof HTMLElement) {
+        node
+          .querySelectorAll(".CodeMirror-guttermarker-subtle")
+          .forEach(elem => {
+            elem.classList.toggle("visible");
+          });
+      }
+    };
+
+    const codeMirrorGutter = codeMirror.getGutterElement();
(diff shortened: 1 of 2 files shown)


src/components/Editor/index.js
```diff
@@ -156,6 +156,20 @@ class Editor extends Component {
...
 
+    const toggleFoldMarkerVisibility = e => {
+      if (node instanceof HTMLElement) {
+        node
+          .querySelectorAll(".CodeMirror-guttermarker-subtle")
+          .forEach(elem => {
+            elem.classList.toggle("visible");
+          });
+      }
+    };
```
jasonLaster: lets move this to `utils/editor`

jasonLaster: where is node coming from?

irfanhudda: @jasonLaster It is defined in this function 
```javascript
const node = ReactDOM.findDOMNode(this);
```

jasonLaster: ahh, do we need it? I think if you're toggling we can assume we have an editor :)

jasonLaster:
closing it as we'll move

jasonLaster:
Closed it because I thought we'd move it to devtools-core, but there's a part we should keep

src/components/Editor/index.js
irfanhudda: We need `node` to get all fold markers and toggle class. I'm not sure if there is a better method here other than using `querySelectorAll`.

Issue #2553, opened 2017-04-06
'Reveal in tree' should also work for tabs with pretty printed source


Archaeopteryx:
Firefox 55.0a1 20170406030206 on Windows 8.1, also with latest 54.0a2

'Reveal in tree' doesn't work for tabs with pretty printed source.

Steps to reproduce:
1. Open http://gridbyexample.com/examples/example5/
2. Open the debugger.
3. Open the file platform.twitter.com/js/... (there is only one file).
4. Pretty print the source (either from the tab's context menu or the button at the bottom).
5. Collapse all opened folders on the left.
6. Use the tab context menu on the opened tab and use 'Reveal in tree'.

Actual result:
Nothing happens.

Expected result:
File selected in tree like it happens for the raw source tab.

arthur801031:
I'll take a stab at this :)

PR #2621, opened 2017-04-11
'Reveal in tree' should also work for tabs with pretty printed source 

arthur801031:
Associated Issue: #2553 

### Summary of Changes
* Hide `Reveal in Tree` menu item for pretty-printed tabs

### Test Plan
* Right click on pretty-printed tabs to see if `Reveal in Tree` menu item is gone.

### DIFF
--- a/src/components/Editor/Tabs.js
+++ b/src/components/Editor/Tabs.js
@@ -233,11 +233,11 @@ class SourceTabs extends Component {
       },
       { item: closeAllTabsMenuItem },
       { item: { type: "separator" } },
-      { item: copySourceUrl },
-      { item: showSourceMenuItem }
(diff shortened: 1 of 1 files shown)


arthur801031:
@jasonLaster and I discussed about this issue and decided to hide the `Reveal in Tree` menu item for pretty-printed tabs since there is no appropriate fix at this moment. 

The cause of this issue is that there is no pretty-printed source in the `sourceTree`, so [findSource](https://github.com/devtools-html/debugger.html/blob/master/src/utils/sources-tree.js#L369) returns null and it cannot find a source to reveal.

jasonLaster:
Thanks @arthur801031

Issue #2569, opened 2017-04-06
[SourcesTree] Fix context menu


jasonLaster:
It looks like the sources tree context menu has been broken for the past month.
I believe the problem is that source directories no longer have a trailing slash, so `isDirectory` gets confused...

It would be nice to fix the bug and understand when it originated

```diff
diff --git a/src/components/SourcesTree.js b/src/components/SourcesTree.js
index 360a494..2049204 100644
--- a/src/components/SourcesTree.js
+++ b/src/components/SourcesTree.js
@@ -175,6 +175,8 @@ class SourcesTree extends Component {

     const menuOptions = [];

+    // we're now failing to detect directories because they no longer
+    // have a trailing slash e.g. "/devtools-html.github.io/debugger-examples/examples/todomvc/bower_components"
     if (!isDirectory(item)) {
       const source = item.contents.get("url");
       const copySourceUrl = {
diff --git a/src/utils/sources-tree.js b/src/utils/sources-tree.js
index c5f2233..8790cd3 100644
--- a/src/utils/sources-tree.js
+++ b/src/utils/sources-tree.js
@@ -156,6 +156,7 @@ function isDirectory(url: Object) {
   const parts = url.path.split("/").filter(p => p !== "");

   // Assume that all urls point to files except when they end with '/'
+  // expects the last character to be a slash "/"
   return parts.length === 0 || url.path.slice(-1) === "/";
 }
```

PR #2624, opened 2017-04-11
Fix isDirectory function

irfanhudda:
Associated Issue: #2569

### Summary of Changes

* Fix `isDirectory` function

### Test Plan

- [x] Right click on folder in source tree doesn't throw error

### DIFF
--- a/src/utils/sources-tree.js
+++ b/src/utils/sources-tree.js
@@ -156,7 +156,10 @@ function isDirectory(url: Object) {
   const parts = url.path.split("/").filter(p => p !== "");
 
   // Assume that all urls point to files except when they end with '/'
-  return parts.length === 0 || url.path.slice(-1) === "/";
+  // Or directory node has children
+  return parts.length === 0 ||
+    url.path.slice(-1) === "/" ||
+    nodeHasChildren(url);
 }
 
 /**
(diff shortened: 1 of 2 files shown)


src/utils/sources-tree.js
```diff
@@ -156,7 +156,10 @@ function isDirectory(url: Object) {
   const parts = url.path.split("/").filter(p => p !== "");
 
   // Assume that all urls point to files except when they end with '/'
-  return parts.length === 0 || url.path.slice(-1) === "/";
+  // Or directory node has children
+  return parts.length === 0 ||
+    url.path.slice(-1) === "/" ||
```
irfanhudda: I don't know if we should remove this now that source directories no longer have a trailing slash.

jasonLaster: Hmm, lets remove that clause and the comment.

Do you know when we removed that requirement?
@arthur801031, @Dalimil?

jasonLaster:
lets also get a unit test in there

irfanhudda:
@jasonLaster Added a test here

src/utils/sources-tree.js
irfanhudda: Removing that condition breaks this test https://github.com/devtools-html/debugger.html/blob/master/src/utils/tests/sources-tree.js#L151

It is because `http://example.com/folder/c/` is no longer considered a folder.

jasonLaster: ah i see. 

Thanks

Issue #2641, opened 2017-04-13
[SearchBar] Allow F3 / Shift+F3 to function the same as Enter/Shift+Enter when searching words


tommai78101:
In Chrome and Visual Studio Code, you can use `F3` / `SHIFT+F3` to navigate through search results for the next / previous keywords. They also use `ENTER` / `SHIFT+ENTER` to navigate.

Currently in Firefox debugger, you can only use `ENTER` / `SHIFT+ENTER` to navigate through search results, but you can't use `F3` / `SHIFT+F3` to navigate.

Can we make it so `F3` / `SHIFT+F3` is used to navigate, so it behaves the same in Chrome?

AnshulMalik:
Seems good idea to me.

tommai78101:
The line of code in question:

https://github.com/devtools-html/debugger.html/blob/a29ec8c650c895590a9a59a3ee6c8cd9da68bcdb/src/components/Editor/SearchBar.js#L480

We just need to add in `|| e.key !== "F3"`.

wldcordeiro:
I was trying to test this but had the Firefox search dialog pop-up when I pressed F3.

jasonLaster:
Hmm, we might need to also add prevent default here:
https://github.com/devtools-html/debugger.html/blob/a29ec8c650c895590a9a59a3ee6c8cd9da68bcdb/src/components/Editor/SearchBar.js#L479-L485

tommai78101:
Just saw this while leaving from work...   D:

I'll see what I can do, despite that the branch I had was already deleted.

jasonLaster:
thanks. create a new one :)

tommai78101:
Just created one. Please test / let me know if anything.

PR #2642, opened 2017-04-13
Added F3 hotkey for searching.

tommai78101:
Associated Issue: #2641 

### Summary of Changes

* Added `F3` as a hotkey.

### Results

* Expecting `Enter` and `F3` to be used as a way to navigate search terms.


jasonLaster:
🏃

PR #2643, opened 2017-04-13
[mochitests] try cloning first

jasonLaster:
Wouldn't it be great if we could always use the latest and greatest from m-c?

* We'll test this to see how fast/slow it is.
* if it works we can practice using a changeset which will be easy to change


jasonLaster:
https://circleci.com/gh/jasonLaster/debugger.html/2496?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

Issue #2446, opened 2017-03-24
[FileSearch] Support wildcard characters “@“ “#"


jasonLaster:
It would be really cool if we could support wildcard characters at the beginning of queries like "@" and "#". The UX is like:

1. type "@"
 = > we select function search mode and highlight the button
2. type "on" 
  => we search for functions starting with "on"
  
#### diff

* doesn't cover firing an action to change the mode when the wildcard changes

```diff
diff --git a/src/components/Editor/SearchBar.js b/src/components/Editor/SearchBar.js
index b82e267..5b86a0d 100644
--- a/src/components/Editor/SearchBar.js
+++ b/src/components/Editor/SearchBar.js
@@ -405,7 +405,18 @@ const SearchBar = React.createClass({
   },

   async onChange(e: any) {
-    return this.doSearch(e.target.value);
+    let query = e.target.value;
+
+    // check the first character of the query for special characters...
+    if (hasWildCard(query)) {
+      const firstChar = query[0]
+      query = query.slice(1, -1);
+
+      //....
+    }
+
+
+    return this.doSearch(query);
   },
```

jacobjzhang:
@jasonLaster If I'm understanding this correctly, the @ character activates function search?

When I think wildcard, I think a search term like "*ey" will look for terms like "key", "they", etc.

wldcordeiro:
> When I think wildcard, I think a search term like "*ey" will look for terms like "key", "they", etc.

I think having Regex search beats that 😄

AnshulMalik:
Can we do something like this ?
```diff
diff --git a/src/components/Editor/SearchBar.js b/src/components/Editor/SearchBar.js
index b16ee37..4ffde22 100644
--- a/src/components/Editor/SearchBar.js
+++ b/src/components/Editor/SearchBar.js
@@ -473,7 +473,13 @@ class SearchBar extends Component {
   }
 
   async onChange(e: any) {
-    return this.doSearch(e.target.value);
+    let query = e.target.value;
+
+    if (query[0] == "@") {
+      query = query.slice(1);
+      this.toggleSymbolSearch(e, { toggle: false, searchType: "functions" });
+    }
+    return this.doSearch(query);
   }
 
   onKeyUp(e: SyntheticKeyboardEvent) {
```

jasonLaster:
yeah! that looks good. give it a try!

jasonLaster:
[...]

PR #2669, opened 2017-04-17
Support wildcard '@' in search

AnshulMalik:
Associated Issue: #2446 

wild


jasonLaster:
Wow! This is really cool :)

I can't believe you can do this with such a small change.

I'm going to close this now because @wldcordeiro is working on moving function and maybe symbol search to a modal. Once that lands we can revisit this.

Thanks!

PR #2670, opened 2017-04-17
[SearchBar] Check before scrolling to result in result-list

irfanhudda:
Found a problem which can be fixed with a minor change
### STR
* Open TodoMVC example in browser
* Open `js/views/app-view.js` file in debugger.html
* Open function search in editor and enter `a`.
* Navigate to lower result by pressing `down` arrow key a few times
* Now enter `a` in function search to narrow the result

search-bar

### Summary of Changes

* check before scrolling list

### DIFF
--- a/src/components/Editor/SearchBar.js
+++ b/src/components/Editor/SearchBar.js
@@ -163,7 +163,11 @@ class SearchBar extends Component {
       searchInput.focus();
     }
 
-    if (this.refs.resultList && this.refs.resultList.refs) {
+    if (
+      this.refs.resultList &&
+      this.refs.resultList.refs &&
+      this.state.symbolSearchResults.length > this.state.selectedResultIndex
+    ) {
       scrollList(this.refs.resultList.refs, this.state.selectedResultIndex);
     }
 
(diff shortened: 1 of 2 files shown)


src/components/Editor/SearchBar.js
```diff
@@ -163,7 +163,11 @@ class SearchBar extends Component {
       searchInput.focus();
     }
 
-    if (this.refs.resultList && this.refs.resultList.refs) {
+    if (
+      this.refs.resultList &&
+      this.refs.resultList.refs &&
+      this.state.symbolSearchResults.length > this.state.selectedResultIndex
```
irfanhudda: This can also be implemented as
```javascript
this.refs.resultList.refs.hasOwnProperty(this.state.selectedResultIndex)
```
Or adding condition in https://github.com/devtools-html/debugger.html/blob/master/src/utils/result-list.js#L4 
```javascript
function scrollList(resultList, index) {
  if (!resultList.hasOwnProperty(index)) return;
  const resultEl = resultList[index];
```

jasonLaster: Oh nice - i think this 

```js
function scrollList(resultList, index) {
  if (!resultList.hasOwnProperty(index)) return;
}
```

is the most testable

Issue #2536, opened 2017-04-04
[EditorPreview] support outer scope expressions


jasonLaster:
#2530 added support for only showing a variable in scope. The issue is that it is too agressive and narrows the scope to the nearest block and not the surrounding block. 

In this case, if you pause at the debugger statement and mouse over the `parent` and `child` declaration you'll see that the parser considers them out of scope.

```js
function(protoProps, staticProps) {
    var parent = this;
    var child;

    if (protoProps && _.has(protoProps, 'constructor')) {
      child = function(){ return parent.apply(this, arguments); };
      debugger
    }
}
```

We can TDD this fix in `utils/tests/parser`

ryanjduffy:
Added a test ... and it passed ... 🤔 ... so I created a sample file and loaded it into the debugger. Looks like it's working correctly to me but let me know if i'm misunderstanding.

![image](https://cloud.githubusercontent.com/assets/788456/24782518/633071e0-1b0c-11e7-82c7-b554da73a666.png)

jasonLaster:
This is now fixed in #2672

PR #2672, opened 2017-04-17
[WIP] Address some scope inconsistencies

jasonLaster:
Associated Issue: #2536

### Summary of Changes

Previously, we would determine if a node was in scope by seeing if it's most local scope contained it. We now get all of the variables defined along the scope chain. So get all the variables in the narrowest scope up to the program. 

I tried to tidy up some of the parsing logic along the way so that it would be more consistent.

We now have more methods that are more specific to what we want to do:

+  getClosestPath,
+  getClosestScope,
+  getClosestExpression,
+  getVariablesInLocalScope,

### DIFF
--- a/src/components/Editor/index.js
+++ b/src/components/Editor/index.js
@@ -280,6 +280,10 @@ class Editor extends Component {
   }
 
   onMouseOver(e) {
+    const { target } = e;
+    if (!target.parentElement.closest(".CodeMirror-line")) {
+      return;
+    }
     this.previewSelectedToken(e);
   }
 
--- a/src/utils/tests/parser.js
+++ b/src/utils/tests/parser.js
@@ -1,8 +1,10 @@
 const expect = require("expect.js");
 import {
   getSymbols,
+  getVariablesInLocalScope,
   getVariablesInScope,
-  getPathClosestToLocation,
+  getClosestScope,
+  getClosestExpression,
   resolveToken
 } from "../parser/utils";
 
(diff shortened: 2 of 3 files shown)


src/utils/tests/parser.js
```diff
@@ -96,46 +99,110 @@ describe("parser", () => {
...
+      expect(node.type).to.be("FunctionDeclaration");
     });
+  });
 
+  describe("getVariablesInLocalScope", () => {
     it("finds scope binding variables", () => {
-      var vars = getVariablesInScope(getSourceText("math"), {
-        line: 1,
+      const scope = getClosestScope(getSourceText("math"), {
+        line: 2,
```
jasonLaster: so this is a fishy thing where I'm not sure where a scope starts in a function... I'll look into this tomorrow

src/utils/tests/parser.js
```diff
@@ -207,7 +274,7 @@ describe("parser", () => {
         { line: 9, column: 25 },
         {
           location: {
-            line: 1,
+            line: 2,
```
jasonLaster: more fishiness

src/components/Editor/index.js
```diff
@@ -280,6 +284,17 @@ class Editor extends Component {
...
   onMouseOver(e) {
+    const { target } = e;
+    if (
+      !hasClass(target, [
+        "cm-variable-2",
+        "cm-property",
+        "cm-def",
+        "cm-keyword"
+      ])
+    ) {
```
jasonLaster: previously everything would be treated like a token

src/utils/tests/parser.js
ryanjduffy: IIRC from my `resolveToken` work, the scope is bounded by the braces so it starts just after `function () {` (but before any newlines)

src/components/Editor/index.js
```diff
@@ -72,6 +72,10 @@ type EditorState = {
   selectedExpression: ?Object
 };
 
+function hasClass(el, classnames) {
+  return classnames.some(className => el.classList.contains(className));
```
jasonLaster: nice. I think i'll switch to just doing. is it a child of cm-line lol. I kinda want all tokens

src/utils/tests/parser.js
```diff
@@ -96,46 +99,110 @@ describe("parser", () => {
...
-      expect(closestPath.type).to.be("Identifier");
+      expect(node.type).to.be("FunctionDeclaration");
     });
+  });
 
+  describe("getVariablesInLocalScope", () => {
     it("finds scope binding variables", () => {
-      var vars = getVariablesInScope(getSourceText("math"), {
-        line: 1,
+      const scope = getClosestScope(getSourceText("math"), {
```
jasonLaster: yeah - that's true

jasonLaster: yeah, so scope is a really big object that i don't really want to serialize. it's big because it has lots of info about parents and basically the tree :/

PR #2680, opened 2017-04-18
Object Inspector tests

jasonLaster:


src/components/shared/ObjectInspector.js
```diff
@@ -227,7 +226,7 @@ class ObjectInspector extends Component {
       },
       getExpanded,
       setExpanded,
-      renderItem: this.renderItem
+      renderItem: renderItem
```
AnshulMalik: Can we do it `renderItem` only instead of `renderItem: renderItem` :)

jasonLaster: yep

jasonLaster:
I'm closing this now because we'll need to change the tree API for renderItem, which is a bit more work

Issue #2677, opened 2017-04-18
[Why] Should show the pause reason 


jasonLaster:
We currently show the user that they're paused at an exception, but we don't show the Exception type and message:

`ReferenceError: foo is not defined`

```diff
diff --git a/src/components/SecondaryPanes/WhyPaused.js b/src/components/SecondaryPanes/WhyPaused.js
index 4e32665..40c7344 100644
--- a/src/components/SecondaryPanes/WhyPaused.js
+++ b/src/components/SecondaryPanes/WhyPaused.js
@@ -17,11 +17,17 @@ class WhyPaused extends Component {
     }

     const message = pauseInfo.getIn(["why"]).get("message");
-    if (!message) {
-      return null;
+    if (message) {
+      return  dom.div(null, message);;
+    }
+
+    const exception =  pauseInfo.getIn(["why"]).get("exception");
+    if (exception) {
+      // do something better here
+      return  dom.div(null, message);
     }

-    return dom.div(null, message);
+    return null;
   }

```

![Screen Shot 2017-04-18 at 1.16.54 PM.png](https://shipusercontent.com/22d74c878829199825f166b7965cf425/Screen%20Shot%202017-04-18%20at%201.16.54%20PM.png)

PR #2684, opened 2017-04-19
Show the exception message if one exists

bomsy:
Associated Issue: #2677

### Summary of Changes

* Add exceptions message

### Screenshots
screen shot 2017-04-19 at 1 13 49 pm

### DIFF
--- a/src/components/SecondaryPanes/WhyPaused.js
+++ b/src/components/SecondaryPanes/WhyPaused.js
@@ -17,11 +17,20 @@ class WhyPaused extends Component {
     }
 
     const message = pauseInfo.getIn(["why"]).get("message");
-    if (!message) {
-      return null;
+    if (message) {
+      return dom.div({ className: "message" }, message);
     }
 
-    return dom.div(null, message);
+    const exception = pauseInfo.getIn(["why"]).get("exception").toJS();
+    if (exception) {
+      const preview = exception.preview;
+      return dom.div(
+        { className: "message" },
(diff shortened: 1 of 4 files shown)


src/components/SecondaryPanes/WhyPaused.js
```diff
@@ -17,11 +17,20 @@ class WhyPaused extends Component {
     }
 
     const message = pauseInfo.getIn(["why"]).get("message");
-    if (!message) {
-      return null;
+    if (message) {
+      return dom.div({ className: "message" }, message);
+    }
+
+    const exception = pauseInfo.getIn(["why"]).get("exception").toJS();
```
clarkbw: I think you should be able to do `pauseInfo.getIn(["why", "exception"])` here.  I'm pretty sure I did the `getIn().get()` pattern above for `message` and I think I made a mistake.

jasonLaster:
Looks like we got a test failure, which can be run in the WTR

https://github.com/devtools-html/debugger.html/blob/master/src/test/integration/runner.js#L78-L84

bomsy:
hmmm ... tests seem to pass locally for me

jasonLaster:
Hmm. I'll look

jasonLaster:
Looks like we were catching an edge case where exceptions were strings!

* updated the code
* added another unit test

screen shot 2017-04-21 at 4 49 39 pm

bomsy:
ahaa

Issue #2578, opened 2017-04-07
[EditorPreview] Watch Expressions + Input


jasonLaster:
We've gotten feedback that it would be great if you could save expressions quickly to "Watch Expressions". This feature adds a bottom bar to previews that let you quickly refine and save an expression.

The input initially shows the preview expression, but you can edit it to be anything.

### **Notes:**

* we can use the input styling from [conditional breakpoints]( https://github.com/devtools-html/debugger.html/blob/master/src/components/Editor/ConditionalPanel.js#L44-L50)
* save to watch expressions logic can be borrowed from the [editor context menu](https://github.com/devtools-html/debugger.html/blob/master/src/components/Editor/EditorMenu.js#L59-L64)

### Screenshot
<img width="816" alt="screen shot 2017-04-07 at 11 27 55 am" src="https://cloud.githubusercontent.com/assets/254562/24807198/55b56efc-1b85-11e7-8926-7cc58faf4cb3.png">

### Patch

```diff
diff --git a/src/components/Editor/Preview.js b/src/components/Editor/Preview.js
index 630b1b4..288d1f3 100644
--- a/src/components/Editor/Preview.js
+++ b/src/components/Editor/Preview.js
@@ -92,6 +92,8 @@ class Preview extends Component {
     });
   }

+  renderBottomBar() {}
+
   renderPreview(expression, value) {
     const root = {
       name: expression,
@@ -104,7 +106,10 @@ class Preview extends Component {
     }

     if (value.type === "object") {
-      return this.renderObjectPreview(expression, root);
+      return [
+        this.renderObjectPreview(expression, root),
+        this.renderBottomBar()
+      ];
     }

```

ruturajv:
Just so that I'm clear... Its like adding a `watch` but instead of a variable, we're watching an `expression`...  ie `isCrazy || isGeek` ? @jasonLaster

jasonLaster:
haha, that's correct

jasonLaster:
We landed the first part of the preview bottom bar:

- [ ] Lets improve the watch button with (style, localization, tooltips)
- [ ] Fix the case with member expressions :)

jasonLaster:
moving this to the enhancements board as it's a nice to have and fairly complicated.

https://github.com/devtools-html/debugger.html/projects/5

PR #2713, opened 2017-04-22
WIP: Adding to watch Expressions, commented mouseLeave

ruturajv:
Associated Issue: #2578 

- Created UI as suggested
- Somehow [`addExpression`](https://github.com/devtools-html/debugger.html/compare/master...ruturajv:bug-2578-save-preview-to-watch-expressions?expand=1#diff-358af90c482e216b601199f75f64a27bR103) isn't working as I hoped, @jasonLaster @wldcordeiro  could you please check ?
- please ignore the mouseLeave comment

- regarding a comment on the mock, `input - updates preview`should the expression be listed in an input box and then user can play around with it to see preview changing ... and then add ?

### DIFF
--- a/src/components/Editor/Preview.js
+++ b/src/components/Editor/Preview.js
@@ -9,6 +9,7 @@ const ObjectInspector = React.createFactory(
 );
 const Popover = React.createFactory(require("../shared/Popover").default);
 const previewFunction = require("../shared/previewFunction").default;
+const { addExpression } = actions;
 
 import { getLoadedObjects } from "../../selectors";
 import { getChildren } from "../../utils/object-inspector";
@@ -89,6 +90,24 @@ class Preview extends Component {
     });
   }
 
+  renderAddToExpressionBar(expression) {
+    return dom.div(
+      { className: "add-to-expression-bar" },
+      dom.div({ className: "prompt" }, "»"),
+      dom.div({ className: "expression-to-save-label" }, expression),
+      dom.div(
+        {
+          className: "expression-to-save-button",
+          onClick: event => {
+            console.log(expression);
+            addExpression(expression);
+          }
+        },
+        L10N.getStr("addWatchExpressionButton")
+      )
+    );
+  }
+
   renderPreview(expression, value) {
     const root = {
       name: expression,
@@ -101,7 +120,11 @@ class Preview extends Component {
     }
 
     if (value.type === "object") {
-      return this.renderObjectPreview(expression, root);
+      return dom.div(
+        {},
+        this.renderObjectPreview(expression, root),
+        this.renderAddToExpressionBar(expression, value)
+      );
     }
 
     return this.renderSimplePreview(value);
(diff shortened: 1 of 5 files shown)


src/components/Editor/Preview.js
```diff
@@ -89,6 +90,24 @@ class Preview extends Component {
...
+    return dom.div(
+      { className: "add-to-expression-bar" },
+      dom.div({ className: "prompt" }, "»"),
+      dom.div({ className: "expression-to-save-label" }, expression),
+      dom.div(
+        {
+          className: "expression-to-save-button",
+          onClick: event => {
+            console.log(expression);
+            addExpression(expression);
```
jasonLaster: Ahh, this will need to be passed in from the Editor.

jasonLaster:
Just pushed a fix for `addExpression`

Perhaps we style the watch button and merge the. Then in V2 we can add an input in?



screen shot 2017-04-22 at 11 48 25 am

ruturajv:
is the `addExpression` fix in devtools-html repo ? I did a fetch and all that I got was this
```
   5535774..e33814b  master     -> devtools-html/master
```
and only change is `src/actions/tests/breakpoints.js`. Was that on your local?

Regarding the input preview updates: Personally, why would a developer use the preview window (interacting with the input box) ? I normally just add a watch.

The `Watch` label looks strangely padded in Mac! I use Linux. Lemme see if I can work on that as well.

jasonLaster:
sorry, the push didn't go through :)

jasonLaster:
@ruturajv re: input, we explored adding a filter UI in previous previews so that users could drill down into the data. I hope that input will be a more flexible UI for refining the preview before it's saved to watch

ruturajv:
@jasonLaster There is a problem with a current implementation, if somebody hovers over `model` from `this.model` then the watch expression added is `model` while it should be  `this.model`. I tried going thru `ObjectInspector`'s getChildren, etc.. but Monday has caught up with its daily grind :) and can't single-handedly focus

Need a function / method that'll get me the entire path, ie. `this.model`

Help appreciated  - Thanks !

jasonLaster:
@ruturajv that's an interesting problem

It's related to what we pass in here: https://github.com/ruturajv/debugger.html/blob/bug-2578-save-preview-to-watch-expressions/src/components/Editor/index.js#L736

jasonLaster:
@ruturajv I merged the code we have behind a feature flag

* Lets improve the watch button with (style, localization, tooltips)
* Fix the case with member expressions :)

ruturajv:
Thanks @jasonLaster 
Could you also please explain [this change](https://github.com/devtools-html/debugger.html/pull/2713/commits/4c04e3c82dc6435763980b59884069a3c3b5f9fd) ?

AnshulMalik:
Hey @ruturajv , this change adds an option to settings panel to enable or disable `watch expressions from preview` feature.

Issue #2590, opened 2017-04-08
[Editor] Highlight functions


jasonLaster:
One nice thing that VS Code does when you're searching for a function is highlight the function and signature and body. They do it with an off putting yellow, but i think the principle is good.

Lets try and add a similar function to our editor with codemirror.

### Notes

* **getting function locations** We've already done the hard work by parsing the file and getting functions with `getSymbols`. Symbols returns a list of functions. Here's one functions data, which has the start and end location for the entire function.

```json
{
  "id": "initialize:32",
  "title": "initialize",
  "subtitle": ":32",
  "value": "initialize",
  "location": {
    "start": {
      "line": 32,
      "column": 14
    },
    "end": {
      "line": 36,
      "column": 3
    }
  }
}
```

* **highlighting lines** there are two ways i can think of when it comes to highlighting the function. The first is to use the highlighter from search, which operates on the token level. The second way is to add a class to each line of the function, with a special class for the first and last line probably. This approach is similar to how we highlight the current debug line. I think this approach is the cleanest.

it'd look something like this:

```js
highlightFunction(function) {
   this.highlightedFunction = function;
   const {location: {start, end}} = function;
   const [firstLine, ...lines, lastLine] = // get array of lines
   firstLine.addClass()
   lines.forEach()
   lastLine.addClass
}
```

sadly, we'd also have to do the same thing when it comes to unhighlighting the function as well. We'll need to do some cleanup when the selected source changes and the highlight is removed.

### GIF

<img width="1079" alt="screen shot 2017-04-07 at 7 58 40 pm" src="https://cloud.githubusercontent.com/assets/254562/24823489/c181eb3c-1bcc-11e7-8851-c67a96991c02.png">

jcreighton:
Grabbing this one!

jasonLaster:
@jcreighton  alright - but you have to pick the highlight color! :P

clarkbw:
I think this was done in #2717

PR #2717, opened 2017-04-24
[function search] highlight function in editor

jcreighton:
Associated Issue: #2590

~~Opening this up to:~~

~~a) Make sure I've addressed this in a somewhat correct manner
b) Ask for advice on how to clean up the highlighting after the user presses the "esc" key - it seemed a bit weird to include cleanup in `onEscape`~~

Okay! Thank you for the help @jasonLaster, this is looking good & I like the solution you came up with. I chose the breakpoint colors -- I couldn't decide on something better but someone else probably has a better idea.

The only weird this is, as you can see in the GIF below, when you open and start searching, the first function is selected but the highlighting of the functions doesn't kick off until you actively select another one. Not sure if that's a big issue or how to address it.

light

dark


src/components/Editor/Editor.css
```diff
@@ -135,6 +135,14 @@ html[dir="rtl"] .editor-mount {
   100% { background-color: transparent; }
 }
 
+.highlight-line {
+  background-color: green;
+}
+
+.theme-dark .highlight-line {
+  background-color: tomato;
```
jcreighton: I haven't decided on colors yet, so ignore this. Was thinking about the debug colors but with opacity, maybe?

jasonLaster:
Hey @jcreighton I spent a couple hour studying this and I think we can follow the same pattern that we use for Editor Breakpoints and coverage markers. i.e. conditionally render a highlight range w/ codemirror

Here is a [branch] I setup with this idea. Note that it is incomplete:

1. `highlightFunction` has dummy data
2. we should only render the range as long as we're looking at that source, so we need to know the sourceId
3. `HighlightFunction` will always update, so we need to polish the `shouldUpdate`

I personally think this pattern of conditionally rendering aka using React is awesome and we can apply it to our highlight lines... 

Lastly, perhaps we do something when we select an item where we select the location and also fire off a `highlightFunction` action in the UI reducer. The advantage is that we separate the two things so that we can highlight and unhighlight functions a little bit differently... I think this is good, not really sure. Just want to make sure that we can easily unhighlight :)

[branch]: https://github.com/devtools-html/debugger.html/compare/master...jasonLaster:highlighters?expand=1

jcreighton:
@jasonLaster Updated my branch with the changes from your branch. Started working on the `highlightFunction` action & wondering if I'm moving in the right direction with that.

jasonLaster:
looks good to me!

jasonLaster:
This looks awesome!

Two things:
1. it needs a rebase
2. Maybe we call the action `highlightLines` and the redux field `highlightedLineRange`. This way it works for other things too... like highlighted scopes :)

jcreighton:
@jasonLaster This is good for review now! Updated comments above & addressed the naming you suggested. 👍

jasonLaster:
Thanks @jcreighton!

src/reducers/sources.js
```diff
@@ -101,6 +106,19 @@ function update(
       prefs.pendingSelectedLocation = location;
       return state.set("pendingSelectedLocation", location);
 
+    case "HIGHLIGHT_LINES":
+      const { start, end } = action.location;
+      let lineRange = {};
+
+      if (start && end) {
+        lineRange = {
```
jasonLaster: this could be one line `lineRange = {start, end}`

src/components/Editor/SearchBar.js
```diff
@@ -484,12 +486,25 @@ class SearchBar extends Component {
...
       selectSource(selectedSource.get("id"), {
         line: item.location.start.line
       });
     }
+
+    if (selectedSource && selectedSymbolType === "functions") {
+      highlightLineRange({
+        start: item.location.start.line,
+        end: item.location.end.line
+      });
```
jcreighton: Yes!

jcreighton:
@jasonLaster Updated PR! Added `clearHighlightRange`. New GIFs since I found a lighter color for highlighting.

src/reducers/sources.js
```diff
@@ -33,6 +33,10 @@ export type SourcesState = {
     line?: number,
     column?: number
   },
+  highlightedLineRange?: {
+    start?: number,
+    end?: number
+  },
```
jasonLaster: I think this would be nicer in the UI reducer as it has more to do with some UI state. 

Also, we should probably include a sourceID so that we can automatically clear this if the user navigates to another source.

jasonLaster:
I pushed a small fixup for clearing the highlight lines... It seems like it worked for you, but i was seeing some errors without it.

jasonLaster:
just moved the fields to the UI reducer and added some basic action tests

jasonLaster:
I feel like it should be jumping to the function when you highlight the lines, but it doesn't seem to do that when i test it...

src/components/Editor/Editor.css
```diff
@@ -42,6 +42,10 @@ html[dir="rtl"] .editor-mount {
   height: 14px;
 }
 
+.editor-wrapper .highlight-lines {
+  background: var(--theme-selection-background-semitransparent);
```
jasonLaster: nice find

PR #2736, opened 2017-04-25
Convert sources reducer to ES modules

virzen:
Associated Issue: #1884 

### Summary of Changes

* update sources reducer to use ES modules
* update sources reducer test to use ES modules
* set sources reducer's state argument's state to `any` to silence Flow error

Note about last point:
Without it Flow complained about missing annotation after addition of 
`export` statement.  After setting it to `SourcesState` it complained 
about unrecognized methods, which I suspect come from Immutable. 
Since this PR isn't about type-checking, I left it at `any`, as suggested
in [this comment](https://github.com/devtools-html/debugger.html/issues/1884#issuecomment-280710019).

### DIFF
--- a/src/reducers/sources.js
+++ b/src/reducers/sources.js
@@ -8,10 +8,10 @@
  * @module reducers/sources
  */
 
-const I = require("immutable");
-const makeRecord = require("../utils/makeRecord");
-const { getPrettySourceURL } = require("../utils/source");
-const { prefs } = require("../utils/prefs");
+import * as I from "immutable";
+import makeRecord from "../utils/makeRecord";
+import { getPrettySourceURL } from "../utils/source";
+import { prefs } from "../utils/prefs";
 
 import type { Source, Location } from "../types";
 import type { Action } from "../actions/types";
@@ -34,7 +34,7 @@ export type SourcesState = {
   tabs: I.List<any>
 };
 
-const State = makeRecord(
+export const State = makeRecord(
   ({
     sources: I.Map(),
     selectedLocation: undefined,
@@ -44,7 +44,7 @@ const State = makeRecord(
   }: SourcesState)
 );
 
-function update(state = State(), action: Action): Record<SourcesState> {
+export function update(state = State() : SourcesState, action: Action): Record<SourcesState> {
   let availableTabs = null;
   let location = null;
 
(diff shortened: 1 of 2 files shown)


jasonLaster:
@virzen I haven't looked yet, but you might want to use the `record<>` notation for the flow issue. THe give away is immutable

src/reducers/sources.js
```diff
@@ -44,7 +44,10 @@ const State = makeRecord(
   }: SourcesState)
 );
 
-function update(state = State(), action: Action): Record<SourcesState> {
+export function update(
+  state: any = State(),
```
AnshulMalik: Using `state: Record<SourcesState>` should work.

irfanhudda: I'm guessing that you could replace `any` with `Record<SourcesState>`. Can you try this?

virzen: Done, indeed works! 😄

irfanhudda: oops looks like I was late :sweat_smile:

AnshulMalik: Hahaha, It's never too late :P

src/reducers/sources.js
```diff
@@ -8,10 +8,10 @@
  * @module reducers/sources
  */
 
-const I = require("immutable");
-const makeRecord = require("../utils/makeRecord");
-const { getPrettySourceURL } = require("../utils/source");
-const { prefs } = require("../utils/prefs");
+import * as I from "immutable";
```
irfanhudda: We can probably use `import I from "immutable";`

virzen: This seems not to be the case:
```
src/reducers/sources.js:11
 11: import I from "immutable";
            ^ Default import from `immutable`. This module has no default export.
```

irfanhudda: Oh. Thanks for correction

virzen:
🎉

src/reducers/sources.js
clarkbw: Oh that's interesting.  I wonder if this means we should only be importing the things we need from `immutable` in the future.  Thanks for pointing this out and looking it up. 🥂

virzen: Although there would be no significant gain in size (webpack 1.x bundles whole file anyway) I agree it's a good practice to do so. Also, it would be great for the eventual migration to webpack 2 (which features tree-shaking). 😄

Issue #2731, opened 2017-04-25
High CPU usage (constantly querying getSymbols)


yurydelendik:
STR:
1. Open FF at https://devtools-html.github.io/debugger-examples/examples/wasm/fib/fib.index.html (and start listening at 6080)
2. Start development server `yarn start`
3. Open other instance of Firefox or Chrome Canary at localhost:8000 and connect to fib.index.html
4. Open Activity/Performance Monitor

Actual result:
Notice above 100% CPU usage by process

Expected result:
Normal CPU usage

P.S. on Mac OSX

<img width="1065" alt="screen shot 2017-04-24 at 8 15 14 pm" src="https://cloud.githubusercontent.com/assets/1523410/25394798/a583bb5a-29a4-11e7-9920-f88726ac2702.png">

jasonLaster:
I see the same thing here: https://devtools-html.github.io/debugger-examples/examples/evals.html

Issue #2728, opened 2017-04-25
[tests] Jest unit tests fail in node 7.7


jasonLaster:
The `jest` unit tests fail when run in node 7.7 with the error that the global `L10N` is not available.

#### Reproduce

* install node 7.7
* git clone https://github.com/devtools-html/debugger.html.git
* yarn 
* yarn test

I like using [nvm] so that i can install a version `nvm install 7.7` and then test

#### Issue

We setup `global.L10N` [here]. It's possible this has to do with how Jest uses VM contexts.

<img src='https://shipusercontent.com/d31802c998fd9bd217bedcdba224a2f2/Screen%20Shot%202017-04-25%20at%2011.29.04%20AM.png' title='Screen Shot 2017-04-25 at 11.29.04 AM.png' width=1105>

[nvm]: https://www.google.com/webhp?sourceid=chrome-instant&ion=1&espv=2&ie=UTF-8#q=nvm+node
[here]: https://github.com/devtools-html/debugger.html/blob/master/src/test/tests-setup.js

irfanhudda:
I was able to run the test after commenting https://github.com/devtools-html/debugger.html/blob/master/src/utils/tests/scopes.js#L200 and https://github.com/devtools-html/debugger.html/blob/master/src/utils/tests/scopes.js#L204 lines

MylesBorins:
so I'm seeing this fail in node v6.10.2 and v7.6.0... going to assume that @irfanhudda figured this out and the bug was not in Node itself. LMK if I am wrong

jasonLaster:
hmm, i didn't see it in 7.0

jasonLaster:
This could be the bug: https://github.com/nodejs/node/pull/11266

fhinkel:
We changed the behavior in https://github.com/nodejs/node/pull/11266/. Now, more properties are (correctly) deleted. Which test is failing? Is there maybe an incorrect assumption that `L10N` is present when it really should have been deleted? 

Background: Jest is using the vm-module, and we change the behavior of the delete operator in the vm-module. Sorry that that's interfering with your tests :crying_cat_face:

PR #2738, opened 2017-04-25
Fix Outline CPU usage (and unrelated test failure)

asolove:
Associated Issue: #2731

The high CPU usage was due to an infinite loop. `componentWillUpdate` is called any time that props _or state_ are changed. So by triggering a process that will call `setState` inside of it, we then change the state and trigger it to happen again, ad infinitum. For this case, we only want to do the work if props are changing from above, so the correct callback to use is `componentWillReceiveProps`.

In running the tests, I found an unrelated failure that was already present on master. One of the unit tests needed a mocked `L10N` object, so I added it.

### Summary of Changes

* Changed callback that triggers symbol lookup from  `componentWillUpdate` to `componentWillReceiveProps` 
* Added a mock `L10N` object before test that requires it.

### Test Plan

- [x] Open debugger pointing to a window running a JavaScript application
- [x] Debug the debugger and start profiling JS execution
- [x] In the outline, select one of the js files to open it
- [x] Wait a few seconds
- [x] End the profile and check for JS activity. Should be minimal. OS-level stats should also show little to no CPU usage while the js file text is open but not being interacted with.

### DIFF
--- a/src/components/Outline.js
+++ b/src/components/Outline.js
@@ -20,7 +20,7 @@ class Outline extends Component {
     this.state = {};
   }
 
-  componentWillUpdate({ sourceText }) {
+  componentWillReceiveProps({ sourceText }) {
     if (sourceText) {
       this.setSymbolDeclarations(sourceText);
     }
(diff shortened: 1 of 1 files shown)


jasonLaster:
@asolove we fixed the scopes bug an hour before you ran into it :)
https://github.com/devtools-html/debugger.html/issues/2728

just pushed a rebased version

jasonLaster:
Thanks @asolove!

PR #2739, opened 2017-04-25
Convert remaining of reducers to ES modules

virzen:
Associated Issue: #1884 

### Summary of Changes

Converted following reducers to ES modules:
- async requests,
- breakpoints,
- coverage,
- index,
- pause,
- ui.

All reducers has been updated to export `update` reducer by default. This allowed to somewhat simplify main reducer (`reducers/index`).

Also, type definitions for `update` functions' `state` argument have been added.

Another small change is addition of `yarn flow` command to the `yarn test-all` one. Lack of the former in the latter proved to be somewhat inconvenient when I wanted to manually check if everything is fine - the shortest way was `yarn test-all && yarn flow`. I completely agree this change is unrelated to the topic of this PR and won't mind to revert it if it's a problem.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "lint-fix": "yarn lint-js -- --fix",
     "test": "jest",
     "test-coverage": "yarn test -- --coverage",
-    "test-all": "yarn test; yarn lint",
+    "test-all": "yarn test; yarn lint; yarn flow",
     "firefox": "start-firefox --start --location https://devtools-html.github.io/debugger-examples/",
     "chrome": "start-chrome --location https://devtools-html.github.io/debugger-examples/",
     "copy-assets": "node bin/copy-assets --symlink",
--- a/src/reducers/breakpoints.js
+++ b/src/reducers/breakpoints.js
@@ -58,7 +58,7 @@ function allBreakpointsDisabled(state) {
   return state.breakpoints.every(x => x.disabled);
 }
 
-export function update(state: any = State(), action: Action) {
+function update(state: any = State(), action: Action) {
   switch (action.type) {
     case "ADD_BREAKPOINT": {
       const newState = addBreakpoint(state, action);
@@ -257,3 +257,5 @@ export function getBreakpointsLoading(state: OuterState) {
 export function getPendingBreakpoints(state: OuterState) {
   return state.breakpoints.pendingBreakpoints;
 }
+
+export default update;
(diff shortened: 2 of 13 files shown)


package.json
```diff
@@ -28,7 +28,7 @@
     "lint-fix": "yarn lint-js -- --fix",
     "test": "jest",
     "test-coverage": "yarn test -- --coverage",
-    "test-all": "yarn test; yarn lint",
+    "test-all": "yarn test; yarn lint; yarn flow",
```
virzen: Please see note at the bottom of PR description addressing this.

jasonLaster:
re test-all. Agreed! We've wanted to do this.

src/reducers/breakpoints.js
```diff
@@ -50,15 +50,15 @@ function locationMoved(location, newLocation) {
...
+export function makeLocationId(location: Location) {
   return `${location.sourceId}:${location.line}`;
 }
 
 function allBreakpointsDisabled(state) {
   return state.breakpoints.every(x => x.disabled);
 }
 
-function update(state = State(), action: Action) {
+function update(state: Record<BreakpointsState> = State(), action: Action) {
```
jasonLaster: Maybe export default here?

src/reducers/index.js
```diff
@@ -2,22 +2,22 @@
...
+  eventListeners,
+  sources,
+  breakpoints,
   asyncRequests,
-  ui: ui.update,
-  coverage: coverage.update
+  pause,
+  ui,
+  coverage
 };
```
jasonLaster: Nice

jasonLaster:
stop by our slack room :)

jasonLaster:
We'd love to switch to webpack 2, but got stuck last time. We should have the branch lying around though

jasonLaster:
Don't worry about coverage. That always happens with import. I don't know why. I'll drop the limits if I need to

src/reducers/breakpoints.js
virzen: I wondered about this and in the end I followed the pattern that seems to be present in majority of the components - component is defined in one place, but exported by default at the end of the file. Do you think it would be better to place export immediately before the reducer?

virzen:
Re slack room: definitely!
Re webpack: I might take a look at this in some not-so-distant future then, but I can't promise anything. 😄 
Re coverage: Great! 👌

jasonLaster:
Merging as this is too good to pass up.
We can always tweak the exports and I don't want any conflicts to emerge.

PR #2748, opened 2017-04-26
Fix broken source search

### LINKED ISSUE #2742: Broken project search for sources
Looks like #2524 broke the project search.

I think we can easily fix by changing [ProjectSearch](https://github.com/devtools-html/debugger.html/blob/master/src/components/ProjectSearch.js#L97-L98)

bug

AnshulMalik:
On selecting the source from list, the source was not
being opened in the editor

- Fixes #2742 
- In search bar, on press enter, move to next result (previously broken)

### DIFF
--- a/src/components/Editor/SearchBar.js
+++ b/src/components/Editor/SearchBar.js
@@ -470,7 +470,7 @@ class SearchBar extends Component {
   }
 
   onKeyUp(e: SyntheticKeyboardEvent) {
-    if (e.key !== "Enter" || e.key !== "F3") {
+    if (e.key !== "Enter" && e.key !== "F3") {
       return;
     }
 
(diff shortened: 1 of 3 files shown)


src/components/Editor/SearchBar.js
```diff
@@ -470,7 +470,7 @@ class SearchBar extends Component {
   }
 
   onKeyUp(e: SyntheticKeyboardEvent) {
-    if (e.key !== "Enter" || e.key !== "F3") {
+    if (e.key !== "Enter" && e.key !== "F3") {
```
AnshulMalik: This condition was always evaluating to `true` previously.

clarkbw: 🤡

jasonLaster:
We have an open issue for adding tests for Source Search, but I think your work with Result List and some more unit tests will help us here :)

Issue #2727, opened 2017-04-25
Startup bad performance on pages with lots of scripts


fflorent:
Hi there,

I noticed a big performance degradation on the devtools when I have to debug the javascript of my application. After investigating a bit (using the Browser Devtools), it appears that the main point of that slowness comes from the fact there are lots of scripts that are fetched, and the `getSourceByUrl` is called several times. And indeed, my application in development mode runs lots of JS files.

You may reproduce the issue with the following steps:
1. Open the attached [index.html.zip](https://github.com/devtools-html/debugger.html/files/954912/index.html.zip) and unzip it
2. Open the devtools, switch to the Debugger tab
   → it takes a long time (~30sec) to display those simple (but numerous) scripts

Don't hesitate to ask me questions.

Cheers,
Florent

asolove:
This is a pretty standard render/layout thrashing issue. `Editor` and all the other components are currently re-rendered every time any data changes in the store, even if it isn't data they rely on. `Editor` is particularly expensive because every time it updates, it calls methods on the underlying code editor, which does some DOM measurement and manipulation causing layout thrashing. Doing that 500 times in a row adds up to about 10s on my machine.

I can make that problem mostly better by just not calling the underlying editor if the source text hasn't changed. But there are lots of other components also being re-rendered that still leave this taking about 3s. I think the right solution is for all components that are wrapped in `Connect` calls to become `PureComponent`s so they don't update unless their props actually change.

fflorent:
From what I investigated, that's not such an editor issue (which is not even rendered).
But while profiling the test case above at home, I couldn't get the same performance degradation (it takes "only" ~5 seconds) and the result of the profiling are not the same. 

The main difference between the two computers (the one at home and the one at work) is that I use a SSD at home. I need more investigation from my side I guess.

Florent

asolove:
Yeah, the silly part is the editor component isn't even rendering anything interesting. But it is constantly sending the CodeMirror editor a command to show nothing, and CodeMirror is spending a lot of time recalculating some layout stuff about how it's showing nothing, and the whole thing is just ridiculous.

fflorent:
Ouch. Do you want me to export the profiling for the example I produced in my first post?

Florent

jasonLaster:
Thanks @florent. That would be great.

I just got off the phone with @asolve and we investigated adding `reselect`, which would help us avoid unnecessary component re-renders by memoizing selectors.

[here] is an incomplete POC.

[...]

PR #2753, opened 2017-04-26
Improve Editor rendering performance

asolove:
This somewhat improves the performance issue in #2727.

Before this change, `Editor` was being re-rendered whenever any store state changed. For cases like the linked issue, where we were sending 500 actions to add source files to the list, this meant 500 re-renders, even though the contents didn't change.

One particularly-expensive operation is setting the text of the underlying text editor. Setting the message calls internal methods on that editor that need to measure DOM elements, causing lots of layout thrashing for every single change in the store.

### Summary of Changes

* Define `shouldComponentUpdate` for `Editor` so that it doesn't re-render if its props are shallow equal to the previous values. (Except for `getExpression`, which is re-bound every time and therefore never equal, even if it _means_ the same thing.)
* Prevent calling `Editor#showMessage` repeatedly with empty text. If it's already empty, then there's no need to do anything.

### Test Plan

- Open the file in #2727 in debugger.html
- Observe how long it takes for the sources to show up in the sidebar. It previously required 10+ seconds on my machine. Now it is down to around 2. 

### Next steps

While this removes the most costly render-thrashing component, the same problem is present in lots of other components. The `SourceTabs` and `ProjectSearch` components are also currently re-rendering even if their data hasn't changed. It looks like they can't easily be made `PureComponent`s because some of their props are generated dynamically so they aren't equal even when they have the same meaning.

### DIFF
--- a/src/components/Editor/index.js
+++ b/src/components/Editor/index.js
@@ -1,5 +1,5 @@
 // @flow
-import { DOM as dom, PropTypes, createFactory, Component } from "react";
+import { DOM as dom, PropTypes, createFactory, PureComponent } from "react";
 const ReactDOM = require("react-dom");
 import ImPropTypes from "react-immutable-proptypes";
 import { bindActionCreators } from "redux";
@@ -72,7 +72,7 @@ type EditorState = {
   selectedExpression: ?Object
 };
 
-class Editor extends Component {
+class Editor extends PureComponent {
   cbPanel: any;
   editor: any;
   pendingJumpLine: any;
@@ -124,7 +124,9 @@ class Editor extends Component {
     this.clearDebugLine(this.props.selectedFrame);
 
     if (!sourceText) {
-      this.showMessage("");
+      if (this.props.sourceText) {
+        this.showMessage("");
+      }
     } else if (!isTextForSource(sourceText)) {
       this.showMessage(sourceText.get("error") || L10N.getStr("loadingText"));
     } else if (this.props.sourceText !== sourceText) {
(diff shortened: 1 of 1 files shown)


src/components/Editor/index.js
```diff
@@ -220,6 +222,31 @@ class Editor extends Component {
...
+    let key;
+
+    for (key in nextProps) {
+      if (key != "getExpression" && nextProps[key] != this.props[key]) {
+        return true;
+      }
+    }
+
+    for (key in nextState) {
+      if (nextProps[key] != this.props[key]) {
```
jasonLaster: do we want to compare state here?

src/components/Editor/index.js
```diff
@@ -220,6 +222,31 @@ class Editor extends Component {
...
+        return true;
+      }
+    }
```
jasonLaster: generally speaking this looks like we can extract this to a `utils/react.js` file that takes props, state, and a blacklist for props and state.

I'm also okay merging this soon and then removing it with selectors. This way we have a short-term fix.

src/components/Editor/index.js
asolove: Ah, damnit. Guess I'm writing some tests as penance.

src/components/Editor/index.js
asolove: Yeah, the more I think about the reselect work, I'm of a mind to do things _the right way_ unless this editor perf issue is showing up in regular usage.

For `Editor`, adding reselect could enable us to memoize the `getExpression` function prop and then this complex `shouldComponentUpdate` wouldn't be needed at all.

jasonLaster: Agreed. I'll close this PR then as we have a better way to move forward

jasonLaster: We can always come back to this in a week if we run into issues w/ reselect

Issue #2690, opened 2017-04-19
[Sources] Show react file icons in the tree


jasonLaster:
It would be super cool if we could show relevant file icons in the sources tree. One example of this would be the React logo next to React components.

#### Plan

I think we could add the logo if two things are true:

1. the file is inside a components directory
2. we detect that `react` is loaded

Basically if it's a component of a react app. 

Down the road we can be more thorough and check the source text as well for some indicators like the file has react imported and is using `createClass` or `Component`.

#### Patch

Here is a simple patch that shows how we could swap out all JS files for React, but in practice we'll want to be a bit smarter.

```diff
diff --git a/src/components/SourcesTree.js b/src/components/SourcesTree.js
index 12ab597..d30c339 100644
--- a/src/components/SourcesTree.js
+++ b/src/components/SourcesTree.js
@@ -157,6 +157,9 @@ class SourcesTree extends Component {
     }

     if (!nodeHasChildren(item)) {
+      if (item.contents.get("url").match(/js/)) {
+        return Svg("react");
+      }
       return Svg("file");
     }

diff --git a/src/components/shared/Svg.css b/src/components/shared/Svg.css
index 2627c91..e72c3b6 100644
--- a/src/components/shared/Svg.css
+++ b/src/components/shared/Svg.css
@@ -32,6 +32,14 @@
   width: 15px;
 }

+.react svg {
+  width: 12px;
+}
+
+.react path {
+  fill: var(--theme-highlight-blue);
+}
+
```

#### Screenshot (prototype)

![react sources.png](https://shipusercontent.com/64e415fbedb318657896999c7205f51c/react%20sources.png)

andreicristianpetcu:
Can I pick this up?

jasonLaster:
@andreicristianpetcu it's yours!

noamkatzir:
I will take it

andreicristianpetcu:
I did not have time to work on this task and I'm glad @noamkatzir had but please in the future pick tasks that have the [available](https://github.com/devtools-html/debugger.html/issues?q=is%3Aopen+is%3Aissue+label%3Aavailable) lable and not the in progress label.

jasonLaster:
Sorry @andreicristianpetcu I was at a meetup last night and missed the label.

andreicristianpetcu:
No problem :) @jasonLaster

jasonLaster:
adding this to the enhancements board:

https://github.com/devtools-html/debugger.html/projects/5

PR #2758, opened 2017-04-27
React file icon

noamkatzir:
Associated Issue: #2690

added react icon with if the project has react in its sources. not sure we need the search according to folder structure of components which can change from project to project, what do you think?

### DIFF
--- a/src/components/SourcesTree.js
+++ b/src/components/SourcesTree.js
@@ -11,7 +11,6 @@ import {
   getSelectedSource,
   getDebuggeeUrl
 } from "../selectors";
-
 import {
   nodeHasChildren,
   createParentMap,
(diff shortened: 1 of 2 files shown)


src/components/SourcesTree.js
```diff
@@ -161,6 +160,9 @@ class SourcesTree extends Component {
     }
 
     if (!nodeHasChildren(item)) {
+      if (this.props.sources.find(v => /react/.test(v))) {
```
jasonLaster: Hmm, I think we also want to check if the item is in a components directory

jasonLaster: We should also add a feature flag `sourceIcons`

jasonLaster:
Thanks @noamkatzir!

src/components/SourcesTree.js
clarkbw: Here are the docs on [feature flags](https://github.com/devtools-html/debugger.html/blob/master/docs/feature-flags.md)

jasonLaster:
Thanks for the help. I'm going to close this as it's gotten a bit stale. Feel free to pick it up again

Issue #2746, opened 2017-04-26
[CallStack] add a copy frames item in the context menu


jasonLaster:
It would be really great if we could let you copy the call stack by right clicking on the call stack.

an output we could add to the clipboard would be like 

```
onclick  todo-view.js#24
render  todo-view.js#31
listenTo  backbone.js#24
```

We can use the same format that chrome has :)

zivkaziv:
I'm on it.. I'll create pool request for it soon..

clarkbw:
This looks 😎

zivkaziv:
I have a question, can I just edit the file "debugger.properties"?
I want to add there copyStackTrace value and copyStackTrace access key (like copySourceUrl

jasonLaster:
Go for it. That's the point!

Ask in our slack and you'll more reliably get answers

PR #2759, opened 2017-04-27
Copy stack trace to clipboard

zivkaziv:
Associated Issue: #[2746](https://github.com/devtools-html/debugger.html/issues/2746)
 
### Summary of Changes
You can now copy the call stack by right clicking on the call stack.
The output look like:

> initialize  todo-view.js#34)
child  (backbone.js#1566)
addOne  (app-view.js#79)
forEach  (underscore.js#79)
method  (backbone.js#966)
addAll  (app-view.js#86)
fetch  (backbone.js#870)

* Adding **formatCopyName** to the util/frame.js
This method is for creating the copy name standard 

### Test Plan
Example test plan:
- [x] Open the debugger
- [x] Set break point
- [x] Perform an action that will run the code and stop at the break point
- [x] On the Call Stack frame press right click
- [x] Select "Copy Stack Trace" from the context menu
- [x] Paste the clipboard in to notepad and compare between the trace on the debugger to the pasted one

### Screenshots/Videos
[Link to gif](http://recordit.co/lvhctJTNbR)

### DIFF
--- a/src/components/SecondaryPanes/Frames.js
+++ b/src/components/SecondaryPanes/Frames.js
@@ -29,7 +29,7 @@ const NUM_FRAMES_SHOWN = 7;
 
 function renderCopyName(frame: Frame) {
   const displayName = formatDisplayName(frame);
-  const fileName = getFilename(frame.source);
+  const fileName = getFilename(?Source = frame.source);
   const frameLocation = frame.location.line;
 
   return `${displayName} (${fileName}#${frameLocation})`;
--- a/src/components/SecondaryPanes/Frames.js
+++ b/src/components/SecondaryPanes/Frames.js
@@ -3,22 +3,23 @@
 import { DOM as dom, PropTypes, Component } from "react";
 import { bindActionCreators } from "redux";
 import { connect } from "react-redux";
+import get from "lodash/get";
+import { showMenu } from "devtools-launchpad";
+import classNames from "classnames";
+
 import actions from "../../actions";
 import { filterDuplicates } from "../../utils/utils";
 import { getFilename } from "../../utils/source";
 import Svg from "../shared/Svg";
-import get from "lodash/get";
 
 import { getFrames, getSelectedFrame, getSource } from "../../selectors";
 
-import { showMenu } from "devtools-launchpad";
 import { copyToTheClipboard } from "../../utils/clipboard";
 import {
   formatDisplayName,
   annotateFrame,
   formatCopyName
 } from "../../utils/frame";
-import classNames from "classnames";
 
 import type { Frame, Source } from "../../types";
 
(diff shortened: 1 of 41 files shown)


src/components/SecondaryPanes/Frames.js
```diff
@@ -107,10 +113,23 @@ class Frames extends Component {
       menuOptions.push(copySourceUrl);
     }
 
+    if (source) {
```
jasonLaster: I don't think we need an IF statement here

src/components/SecondaryPanes/Frames.js
```diff
@@ -107,10 +113,23 @@ class Frames extends Component {
...
+      };
+
+      menuOptions.push(copyStackTrace);
+    }
+
     showMenu(event, menuOptions);
   }
 
-  renderFrame(frame: LocalFrame) {
+  renderFrame(frame: LocalFrame, frames) {
```
jasonLaster: `renderFrame(frame: LocalFrame, frames: LocalFrame[]) `

src/utils/frame.js
```diff
@@ -97,3 +98,11 @@ export function formatDisplayName(frame: Frame) {
...
   return endTruncateStr(displayName, 25);
 }
+
+export function formatCopyName(frame: Frame) {
+  const displayName = formatDisplayName(frame);
+  const fileName = getFilename(frame.source);
+  const frameLocation = frame.location.line;
+
+  return `${displayName}  ${fileName}#${frameLocation}`;
+}
```
jasonLaster: lets keep this on the top of the component `Frame` because it is kinda like rendering

src/utils/frame.js
```diff
@@ -97,3 +98,11 @@ export function formatDisplayName(frame: Frame) {
   displayName = simplifyDisplayName(displayName);
   return endTruncateStr(displayName, 25);
 }
+
+export function formatCopyName(frame: Frame) {
+  const displayName = formatDisplayName(frame);
+  const fileName = getFilename(frame.source);
+  const frameLocation = frame.location.line;
+
+  return `${displayName}  ${fileName}#${frameLocation}`;
```
jasonLaster: lets change this to `${displayName} (${fileName}#${frameLocation})` to be consistent with chrome and visually group the file name

zivkaziv:
Thanks for the review.
I've fixed all the issues below, but I'm getting error on "yarn flow"..
I'm working to fix it and hopefully push the new code soon

jasonLaster:
@zivkaziv I made a couple tweaks to accomodate flow, formatting, and add a test. Let me know what you think.

zivkaziv:
@jasonLaster your changes looks great, and now the yarn flow is OK (I still not sure what I did wrong):)
On of your comments was to move the method formatCopyName (utils/frame.js) to Frame component.
Do you still want me to do so? Or let's keep it as is?

Thanks again..

jasonLaster:
I think it's fine as a util. It looks like some of the other utils in that
file :)

zivkaziv:
In this case, I think we are ready to roll...

Issue #2685, opened 2017-04-19
remove feature flag for search modifiers


clarkbw:
We've shipped Search Modifiers so its time to remove the feature flag for it.

Here's a [search](https://github.com/devtools-html/debugger.html/search?utf8=%E2%9C%93&q=searchModifiers&type=) of all the usage of the key `searchModifiers`.  

We need to remove all references to that key and make sure search modifiers are now on by default.

andreicristianpetcu:
Can I work on this? @clarkbw

jasonLaster:
Go for it @andreicristianpetcu

jasonLaster:
You can see other recent PRs

andreicristianpetcu:
[Good tip](https://github.com/devtools-html/debugger.html/pulls?utf8=%E2%9C%93&q=is%3Apr%20is%3Aclosed%20remove%20feature%20flag%20)! Thank you @jasonLaster

PR #2761, opened 2017-04-27
removed feature flag for search modifiers #2685

andreicristianpetcu:

Here's the Pull Request Doc

### Summary of Changes

* removed all references to the searchModifiers flag 

### Test Plan

Open the debugger with no mention of searchModifiers in your config/local.json
Open a JS file and hit Ctrl+F. You should see functions and variables as check options in the left and in the right the regex, case sensitive and wholewords options.

### DIFF
--- a/configs/application.json
+++ b/configs/application.json
@@ -33,10 +33,6 @@
       "label": "Code Folding",
       "enabled": false
     },
-    "searchModifiers": {
-      "label": "Search Modifiers",
-      "enabled": false
-    },
     "searchNav": {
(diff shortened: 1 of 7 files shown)


andreicristianpetcu:
@jasonLaster is there any trick to enable all the feature flags in Firefox Nightly? I use the old debugger and the new one seems grown up enough that I can actually use it.

jasonLaster:
> is there any trick to enable all the feature flags in Firefox Nightly

Not at the moment, i'd like to get that setup though.

Issue #2688, opened 2017-04-19
remove feature flag for show source


clarkbw:
We've shipped the Show Source feature so its time to remove the feature flag for it.

Here's a [search](https://github.com/devtools-html/debugger.html/search?utf8=%E2%9C%93&q=showSource&type=) of all the usage of the key `showSource`.  

We need to remove all references to that key and make sure the feature is now on by default.

adamki:
Hi all. New to the repo and would love to take a shot at completing this. 

Looks like we need to remove all the use cases for `showSource` and ensure that we enable default use with something like this:

```
"showSource" : {
   "label": "Show Source",
   "enabled": true
}
``` 
It looks like this change will need to be added to various `config` files. Are there any guides that I can reference regarding `config`? Having a bit of trouble discerning which config files will need the snippet above.

clarkbw:
Hmm, it looks like #1765 partially removed the `showSource` flag but there is just one piece left over.  Looks like a lot smaller change now.  Feel free to grab one of the other as well!

clarkbw:
@adamki do you still want this one or want to take something like #2678 watch expressions instead?

adamki:
Im still on it. 

I've removed all the found instances of `showSource`. Mostly, i am still a bit unclear about the process of setting the showSource feature to default/if this still needs to be done. Am I correct in thinking that if the `showSource` feature has been shipped, that it no longer needs to be explicitly enabled? 

Sorry for the redundant question. Still learning how the feature configs work.

Also, once I finish this, i would be happy to do the same with the other feature flags here: [#2689](https://github.com/devtools-html/debugger.html/issues/2689)

clarkbw:
I think what you mentioned in https://github.com/devtools-html/debugger.html/issues/2688#issuecomment-297192438 is the only remaining reference to `showSource`

Take a look at this PR for watch expressions: #2762 and you'll see more of how the `isEnabled` flags are used to toggle features.

And it seems like most of the others from #2689 might be taken already 😿 We can find something else.

adamki:
There is a pending PR on [this](https://github.com/devtools-html/debugger.html/pull/2766). But there is a small git conflict. I tried rebasing and ran into some issue(likely my bad as im new to rebasing) and will look into it further once I have a bit more time. I just wanted to let others know if they have the desire to take a look at this issue.

PR #2766, opened 2017-04-28
removes showSource feature-flag remnants

adamki:
Associated Issue: [#2688](https://github.com/devtools-html/debugger.html/issues/2688)

### Summary of Changes

Removes the `showSource` feature flag according to these [results](https://github.com/devtools-html/debugger.html/search?utf8=%E2%9C%93&q=showSource&type=)

### Test Plan
After removing all references from `showSource`, I ran the application and ensured that `showSource` still renders with no error(s)

### DIFF
--- a/src/components/Editor/EditorMenu.js
+++ b/src/components/Editor/EditorMenu.js
@@ -8,7 +8,6 @@ function getMenuItems(
     codeMirror,
     selectedLocation,
     selectedSource,
-    showSource,
     onGutterContextMenu,
     jumpToMappedLocation,
     toggleBlackBox,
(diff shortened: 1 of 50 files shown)


src/components/Editor/EditorMenu.js
```diff
@@ -77,8 +76,7 @@ function getMenuItems(
     id: "node-menu-show-source",
     label: revealInTreeLabel,
     accesskey: revealInTreeKey,
-    disabled: false,
-    click: () => showSource(selectedSource.get("id"))
+    disabled: false
```
jasonLaster: haha - we want to keep the feature :)

Just remove the experimental feature flag

adamki:
While trying to resolve the conflict in `configs/firefox-panel.json` i ran into some issues.

I approached the conflict via interactive rebase. My current state is that git is telling me that I have the latest version of master. But when checking the `git diff`... it does not acknowledge that there is a conflict of any sort. 

I've consulted my friends about this and we are stumped.  I would be open to opening another PR, letting someone else complete this, or pairing with a git master.   

thoughts?

PR #2768, opened 2017-04-28
changelog podcast

jasonLaster:
Adds a link to the changelog podcast.

Not sure where to mention it


bomsy:
I think this looks ok.

It would also be great if we have a resources page, were we have links to videos, blog posts etc

jasonLaster:
Yes. Definitely

Issue #2676, opened 2017-04-18
Toggle pane button is positioned incorrectly in RTL layout


irfanhudda:
### Screenshot
![rtl-bug](https://cloud.githubusercontent.com/assets/1755089/25139844/1044f004-247c-11e7-99b0-d503d094dd9d.png)

There are two problems with above RTL layout
+ [x] Toggle pane button for secondary pane (`.toggle-button-end`) should be aligned to the left. (may require a change to [`PaneToggle.css`](https://github.com/devtools-html/debugger.html/blob/master/src/components/shared/Button/PaneToggle.css#L20)) #2770
+ [ ] Splitter appears to stop before command-bar (Some problem with borders?)

jasonLaster:
Nice find!

sharathnarayanph:
Hey guys! I am new to the project. I have been playing around with the code for few days now and I would like to jump in. Is it okay if I start working on this issue?

jasonLaster:
Go for it!

jasonLaster:
Join us in slack if you have questions

sharathnarayanph:
Thanks Jason!

PR #2770, opened 2017-04-29
#2676- PaneToggle.css Change

sharathnarayanph:
Associated Issue: #2676 

Summary of Changes

- Added RTL styling for .toggle-button-end & .toggle-button-start.

Test Plan

- Opened the app in RTL mode and tested the changes.
- Opened the app again in the usual LTR mode and tested for correctness.

Screenshot
rtlchange

### DIFF
--- a/src/components/shared/Button/PaneToggle.css
+++ b/src/components/shared/Button/PaneToggle.css
@@ -22,8 +22,17 @@
   margin-right: 5px;
 }
 
+html[dir="rtl"] .toggle-button-end {
+  margin-left: 5px;
+  margin-right: auto;
+}
+
 .toggle-button-start {
   margin-left: 5px;
+}
+
+html[dir="rtl"] .toggle-button-start {
+  margin-right: 5px;
 }
 
 html:not([dir="rtl"]) .toggle-button-end svg,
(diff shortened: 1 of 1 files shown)


src/components/shared/Button/PaneToggle.css
```diff
@@ -22,10 +22,19 @@
   margin-right: 5px;
 }
 
+html[dir="rtl"] .toggle-button-end {
```
irfanhudda: You are going in correct direction.

But, I don't think we need to manually add this selector. We use [postcss-bidirection](https://github.com/gasolin/postcss-bidirection) to generate these selectors.

Here's how I would try
```patch
--- src/components/shared/Button/PaneToggle.css
+++ src/components/shared/Button/PaneToggle.css
@@ -23,7 +23,7 @@
 }
 
 .toggle-button-start {
-  margin-left: 5px;
+  margin-inline-start: 5px;
 }
 
 html:not([dir="rtl"]) .toggle-button-end svg,
```

You can use similar approach for `.toggle-button-end`

https://devtools-html.github.io/debugger.html/docs/local-development.html#rtl

sharathnarayanph: Thanks Irfan! Let me try again.

PR #2779, opened 2017-05-01
fixes broken ui

bomsy:
Associated Issue: 

screen shot 2017-05-01 at 9 08 54 pm

Fix broken UI due to es module refactor

### Summary of Changes

* Remove `default`


jasonLaster:
oof, i bet this only showed up in the launchpad

PR #2780, opened 2017-05-02
Test report

jasonLaster:
### Summary of Changes

This starts a blog post on the state of our testing and type coverage. I hope to write one quarterly to summarize our progress and next steps.

This is helping me better understand what some good next steps are:

1. adding `State` types in our actions
2. adding types to our tests 
3. adding types to our component props
4. adding immutable types

### DIFF
--- a/docs/updates/5-1-2017-flow.md
+++ b/docs/updates/5-1-2017-flow.md
@@ -1,0 +1,153 @@
+### Flow Coverage Report (May 1st)
+
+We're currently using Flow 0.47 and have 60% coverage. Here's a quick run down of how we're doing.
+
+
+#### Components
+
+* **54%** coverage
+* 44 components are typed (almost all)
+* `yarn flow-react` - creates a flow report for components
+
+Typing our components helps us with rendering logic, which can take advantage
+of our domain objects such as `Frame`.
+
+```js
+function renderFrameTitle(frame: Frame) {
+  const displayName = formatDisplayName(frame);
+  return dom.div({ className: "title" }, displayName);
+}
+```
+
+We are not yet typing our props and state, which would be a helpful way
+to get additional coverage. We currently use `PropTypes`, which do not include
+flow types. We will hopefully switch to [props] soon to take advantage of static typing.
+
+```js
+shouldComponentUpdate(nextProps, nextState) {
+  const { frames, selectedFrame } = this.props;
+  const { showAllFrames } = this.state;
+  return (
+    frames !== nextProps.frames ||
+    selectedFrame !== nextProps.selectedFrame ||
+    showAllFrames !== nextState.showAllFrames
+  );
+}
+```
+
+There have been a couple places where flow types have been difficult to get right.
+
+1. **covariants** `this.toggleFramesDisplay = this.toggleFramesDisplay.bind(this)`. We have an [issue](https://github.com/devtools-html/debugger.html/issues/3172) to improve this.
+2. **defaultProps** and other class properties have been awkward to add types for.
+
+Overall, typing our components has been a nice win though as it gives us additional confidence our UI will behave appropriately and catches several lifecycle edge-cases
+where properties could be null or racey.
+
+#### Utilities
+
+* **59%** coverage
+* **60** files
+* `yarn flow-utils` - creates a flow report for utils
+
+We started typing our utility functions. Starting with utilities helped us
+document our primary business logic and made them easier to refactor.
+
+In this example, we added types to a parser utility that checks to see if an
+expression is in a particular scope.
+
+```js
+export function isExpressionInScope(expression: string, scope?: Scope) {
+  if (!scope) {
+    return false;
+  }
+
+  const variables = getVariablesInScope(scope);
+  const firstPart = expression.split(/\./)[0];
+  return variables.includes(firstPart);
+}
+```
+
+We are not currently typing our test files, but I think it would be helpful if
+we did. In this example, we are searching for all of the functions in a given
+source, but because we did not provide a type for `getSourceText`, flow can
+not use `getSymbols`'s types to tell us if we're writing a reasonable test.
+
+```js
+it("finds functions", () => {
+  const fncs = getSymbols(getSourceText("func")).functions;
+  const names = fncs.map(f => f.value);
+  expect(names).to.eql(["square", "child", "anonymous"]);
+});
+```
+
+Overall, typing our utilities was a great first step. And the more that we type
+the components, actions, and reducers that call the utilities, the more value
+we will get out of them. Already, the types are forcing an internal consistency
+that helps us achieve a better API for our utils.
+
+#### Actions / Reducers
+
+Our Actions and Reducers are generally very well covered.
+
+* **78%** coverage
+* **19** files
+* `yarn flow-redux` - creates a flow report for actions and reducers
+
+Actions like breakpoints and pause have a coverage score of over 80%.
+We initially, typed our action dispatcher params. We've recently added
+types for `ThunkArgs` and `State`. Typing `State`, allows flow to type check
+our selectors and utils.
+
+
+```js
+export function newSource(source: Source) {
+  return ({ dispatch, getState }: ThunkArgs) => {
+    if (prefs.clientSourceMapsEnabled) {
+      dispatch(loadSourceMap(source));
+    }
+
+    dispatch({ type: constants.ADD_SOURCE, source });
+
+    checkSelectedSource(getState(), dispatch, source);
+    checkPendingBreakpoints(getState(), dispatch, source);
+  };
+}
+```
+
+Our reducers are 80% covered and some reducers like breakpoints and pause are 80+
+percent covered. Breakpoints is actually 97% covered. Most of the coverage comes
+from our State type and Action union types.
+
+The `State` type describes the state of the the reducer store. Typing the State
+and action types lets us fully describe our reducers and get a high flow coverage
+score.
+
+```js
+export type BreakpointsState = {
+  breakpoints: I.Map<string, Breakpoint>,
+  pendingBreakpoints: any[],
+  breakpointsDisabled: false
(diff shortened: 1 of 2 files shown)


src/reducers/types.js
```diff
@@ -0,0 +1,9 @@
+import type { PauseState } from "pause";
+import type { SourcesState } from "source";
+import type { BreakpointsState } from "breakpoints";
+
+export type State = {
+  pause: PauseState,
+  sources: SourcesState,
+  breakpoints: BreakpointsState
```
jasonLaster: I should add the remaining reducers

src/test/integration/runner.js
```diff
@@ -44,7 +44,7 @@ describe("Tests", () => {
 
   it("asm", async () => await asm(ctx));
 
-  it("breakpoints - toggle", async () => await breakpoints.toggle(ctx));
+  it.only("breakpoints - toggle", async () => await breakpoints.toggle(ctx));
```
jasonLaster: oops

src/utils/frame.js
```diff
@@ -89,7 +89,7 @@ export function simplifyDisplayName(displayName) {
   return displayName;
 }
 
-export function formatDisplayName(frame: Frame) {
+export function formatDisplayName(frame: LocalFrame): string {
```
jasonLaster: I should make sure this is imported

src/utils/parser/utils.js
```diff
@@ -8,6 +8,7 @@ const toPairs = require("lodash/toPairs");
 const isEmpty = require("lodash/isEmpty");
 const uniq = require("lodash/uniq");
 
+import type { Node, Path, Program } from "babylon";
```
jasonLaster: adding babylon types

src/utils/prefs.js
```diff
@@ -18,7 +18,21 @@ if (isDevelopment()) {
...
+  ignoreCaughtExceptions: boolean,
+  callStackVisible: boolean,
+  scopesVisible: boolean,
+  startPanelCollapsed: boolean,
+  endPanelCollapsed: boolean,
+  tabs: Object,
+  pendingSelectedLocation: Object,
+  pendingBreakpoints: any[],
+  expressions: Object
+};
```
jasonLaster: adding pref types is really helpful

docs/updates/5-1-2017-flow.md
```diff
@@ -0,0 +1,156 @@
...
+    frames !== nextProps.frames ||
+    selectedFrame !== nextProps.selectedFrame ||
+    showAllFrames !== nextState.showAllFrames
+  );
+}
+```
+
+There have been a couple places where flow types have been difficult to get right.
+
+1. **contravariants** `this.toggleFramesDisplay = this.toggleFramesDisplay.bind(this)`
```
asolove: _contravariance_?

docs/updates/5-1-2017-flow.md
```diff
@@ -0,0 +1,156 @@
...
+2. **defaultProps** and other class properties.
```
asolove: What was the difficulty with `defaultProps`?

docs/updates/5-1-2017-flow.md
```diff
@@ -0,0 +1,156 @@
...
+  const variables = getVariablesInScope(scope);
+  const firstPart = expression.split(/\./)[0];
+  return variables.includes(firstPart);
+}
+```
+
+We are not currently typing our test files, but I think it would be helpful if
+we did. In this example, we are searching for all of the functions in a given
+source, but because we did not provide a type for `getSourceText`, flow can
+not use `getSymbols`'s types to tell us if we're writing a reasonable test.
```
asolove: The only caveat here is that it's sometimes useful to be able to "cheat" in tests, like pass in a mock object that doesn't match the full type but does just what this particular unit test needs.

docs/updates/5-1-2017-flow.md
```diff
@@ -0,0 +1,156 @@
...
+The `State` type describes the state of the the reducer store. Typing the State
+and action types lets us fully describe our reducers and get a high flow coverage
+score.
+
+```js
+export type BreakpointsState = {
+  breakpoints: I.Map<string, Breakpoint>,
+  pendingBreakpoints: any[],
+  breakpointsDisabled: false
+};
```
asolove: There are still a lot of `any`s, especially for callback functions and items stored in Immutable structures. Might be good to either review if they are trivially removable or, if there are real impediments, make a list of what they are.

docs/updates/5-1-2017-tests.md
```diff
@@ -0,0 +1,139 @@
...
+
+### Unit Tests - Jest
+
+#### Components
+
+* **6** files
+* **80+** coverage
+* **60** files of coverage
+
+We use Jest to test our react Component. We mostly do snapshot testing to
```
asolove: _React components_?

docs/updates/5-1-2017-flow.md
thejameskyle: Class methods are covariant in Flow by default, which in effect means they are "read-only" and cannot be assigned to.

```js
class Error extends React.Component {
  constructor(props) {
    super(props);
    this.onClick = this.onClick.bind(this); // Error
    //   ^ property `onClick`. Covariant property `onClick` incompatible with contravariant use in
    //     assignment of property `onClick`
  }
  
  onClick(event: ButtonClickEvent) {
    // ...
  }
  
  render() {
    return <button onClick={this.onClick}>Click Me</button>;
  }
}
```

In order to make this work, you need to use an additional class field, which are invariant (read & write) by default, like this:

```js
class WorksWithClassPropertyType extends React.Component {
  constructor(props) {
    super(props);
    this.onClick = this.onClick.bind(this); // Works!
  }
  
  onClick: (ButtonClickEvent) => void;
  onClick(event) {
    // ...
  }
}
```

Alternatively you can use the class properties proposal to get around all of this:

```js
class WorksWithClassPropertyArrowFunction extends React.Component {
  onClick = (event: ButtonClickEvent) => {
    // ...
  };
}
```

[[See in Try Flow](https://flow.org/try/#0PTAEAEDMBsHsHcBQBLAtgB1gJwC6gEoCmAhgMZ6RayqgDkWJ5tA3IojgJ7qGgBCArjhywAdgGFoyUgGsAogDdCIvAF5QAWVj8AzoQVK8AMlABvRKFCl+WBsoAqxLAHNCOAFygAEnfUAZAUKistCEqAYANIgAvqyIpNDE2tqgsjbYoIQAHjhKACbJRGQ4AHRi1JgiBqbmlqLaOFj85NgAFOhU6NoAlNUWFtr83FhtHd2sfaA4ABbI2sWiElLSoGrTs-PikjLFAEbIIrkta2M1UTU1C1vSLYSKyh4BwptL+so9ZhMgoMU-p+cWtlyhGG7xqANc1hEoAAPDtBE9QJclioTMcNosZFEAHwY5bqQjQ4BwwIiLHjUBnM5xBJJUAAdWw0m0dOQ0wkiW0AAUOsDOHYuDwsjkDgVGCUyhhRFUPrURPVGs1hu1YJ1QRMBkMRiqThM0UiZCtJjM5vrpLt9odjl1yWcLBdnjIPC1HqbXjgeiosaB5LBkLlyaabnd3b0+l8fsU-lT4hz6YzmaypuyktyVbyOABBNLwABi-BE5GQogy2TyoqKpXKUuUodNhqDBge8NdwY9XplFnDvwsMWiiCAA)]

docs/updates/5-1-2017-flow.md
thejameskyle: I would also like to know what was difficult

docs/updates/5-1-2017-flow.md
```diff
@@ -0,0 +1,156 @@
...
+}
+```
+
+There have been a couple places where flow types have been difficult to get right.
+
+1. **contravariants** `this.toggleFramesDisplay = this.toggleFramesDisplay.bind(this)`
+2. **defaultProps** and other class properties.
+
+Overall, typing our components has been a nice win though as it gives us additional confidence our UI will behave appropriately and catches several lifecycle edge-cases
+where properties could be null or racey.
```
thejameskyle: racey?

asolove: I think this means that some related data can be set in a non-deterministic order based on how async actions happen to fire. (Especially true for things like parsing that happen in a separate web worker thread.) Flow makes us declare which things have to be present together and prove that all possible state transitions meet those rules.

docs/updates/5-1-2017-flow.md
thejameskyle: Generally type systems eliminate the need to do things like that. But you can do one of two things in that scenario:

1. Use Flow's [`suppress_comment` option](https://flow.org/en/docs/config/options/#toc-suppress-comment-regex)
2. Pass the value through `any` to stop type checking – I'd recommend doing this along with a comment explaining why it `any` is needed. Using `any` means you're less tied to configuration

```js
function method(value: string) {
  if (typeof value === 'number') throw new TypeError('Support for numbers has been removed');
  return value;
}

test('throws deprecation error', () => {
  expect(() => {
    // Passing through `any` because we want to make sure we're throwing an error.
    method((42: any));
  }).toThrow(TypeError);
});
```

docs/updates/5-1-2017-flow.md
thejameskyle: Might want to explain it a bit more

docs/updates/5-1-2017-flow.md
Andarist: @thejameskyle Great explanation, was wondering about why this happens myself too.

Is it the same case when I try to set custom `toString()` on some function?
```javascript
const createActionCreator = (type, actionCreator) = {
  actionCreator.toString = () => type // error!
  return actionCreator
}
```
is there any workaround for this?

PR #2784, opened 2017-05-02
[Work in progress] Add reselect and reduce unnecessary renders

asolove:
Associated Issue: #2727

### Background

In issue #2727, @fflorent reported a very severe performance problem on pages with lots of script tags. In the course of debugging that problem, we noticed that all the components connected to the redux store were being re-rendered on every state change, even if the state change didn't effect the data that component read. That was odd, because Redux promises not to re-render `connect`ed components if the props don't change (as tested by shallow equality). The culprit was our selector functions, which often use `map` and `filter` to build up results. Even if the underlying data stayed the same, these selectors return new objects and arrays every time they are called. A common solution to this problem is to use memoized selectors, which promise to return the same objects if the underlying data hasn't changed. One popular library for doing this is [reselect](https://github.com/reactjs/reselect). 

So the goal of this PR is to make components only re-render when their underlying data has changed, by memoizing selectors as necessary.

### Summary of Changes

For each component, make it a `PureComponent` and ensure its selectors don't change if their underlying data doesn't change:

- [ ] Frames
- [x] Breakpoints
- [x] Scopes
- [x] Expressions
- [x] Editor
- [x] Tabs
- [x] App

### Test Plan

All existing tests should pass (no functionality should have changed).

To verify that re-renders have decreased:

- Do a development build so React Perf tools are installed.
- Open debugger.html and then open a debugger for it.
- Type `Perf.start()`
- Interact with the app as much as you like...
- Type `Perf.stop()`
- `Perf.printExclusive()` will now print a table with how long it spend re-rendering each component. This allows you to see if one component is changing because of unrelated data.

### DIFF
--- a/src/components/SecondaryPanes/Breakpoints.js
+++ b/src/components/SecondaryPanes/Breakpoints.js
@@ -148,10 +148,15 @@ function updateLocation(state, bp): LocalBreakpoint {
   return localBP;
 }
 
-const _getBreakpoints = createSelector(getBreakpoints(state), breakpoints =>
-  breakpoints
-    .map(bp => updateLocation(state, bp))
-    .filter(bp => bp.location.source && !bp.location.source.get("isBlackBoxed"))
+const _getBreakpoints = createSelector(
+  getBreakpoints,
+  state => state,
+  (breakpoints, state) =>
+    breakpoints
+      .map(bp => updateLocation(state, bp))
+      .filter(
+        bp => bp.location.source && !bp.location.source.get("isBlackBoxed")
+      )
 );
 
 export default connect(
--- a/src/components/SecondaryPanes/Breakpoints.js
+++ b/src/components/SecondaryPanes/Breakpoints.js
@@ -1,6 +1,7 @@
 // @flow
 import { DOM as dom, PropTypes, Component } from "react";
 import { connect } from "react-redux";
+import { createSelector } from "reselect";
 import { bindActionCreators } from "redux";
 import ImPropTypes from "react-immutable-proptypes";
 import classnames from "classnames";
@@ -147,13 +148,11 @@ function updateLocation(state, bp): LocalBreakpoint {
   return localBP;
 }
 
-function _getBreakpoints(state) {
-  return getBreakpoints(state)
+const _getBreakpoints = createSelector(getBreakpoints(state), breakpoints =>
+  breakpoints
     .map(bp => updateLocation(state, bp))
-    .filter(
-      bp => bp.location.source && !bp.location.source.get("isBlackBoxed")
-    );
-}
+    .filter(bp => bp.location.source && !bp.location.source.get("isBlackBoxed"))
+);
 
 export default connect(
   (state, props) => ({ breakpoints: _getBreakpoints(state) }),
--- a/src/components/SecondaryPanes/Breakpoints.js
+++ b/src/components/SecondaryPanes/Breakpoints.js
@@ -1,12 +1,17 @@
 // @flow
-import { DOM as dom, PropTypes, Component } from "react";
+import { DOM as dom, PropTypes, PureComponent } from "react";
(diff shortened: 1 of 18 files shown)


fflorent:
Thanks @asolove! I may give it a try tomorrow.

Florent

fflorent:
BTW, I see you use babel. Doesn't it introduce performance issues? (especially for `Map`s and `Set`s)

Florent

src/main.js
```diff
@@ -3,6 +3,11 @@
 const React = require("react");
 const ReactDOM = require("react-dom");
 
+if (process.env.NODE_ENV !== "production") {
+  const Perf = require("react-addons-perf");
+  window.Perf = Perf;
+}
```
jasonLaster: 👍

src/reducers/pause.js
```diff
@@ -138,13 +139,15 @@ function update(
 // (right now) to type those wrapped functions.
 type OuterState = { pause: Record<PauseState> };
 
-export function getPause(state: OuterState) {
-  return state.pause.get("pause");
-}
+const getPauseWrapper = state => state.pause;
```
jasonLaster: I like this style

src/reducers/sources.js
```diff
@@ -273,6 +274,8 @@ function getNewSelectedSourceId(state: SourcesState, availableTabs): string {
 // (right now) to type those wrapped functions.
 type OuterState = { sources: Record<SourcesState> };
 
+const getSourcesWrapper = state => state.sources;
```
jasonLaster: perhaps `sourcesState` or is that too opaque?

src/components/SecondaryPanes/Breakpoints.js
```diff
@@ -147,13 +148,16 @@ function updateLocation(state, bp): LocalBreakpoint {
...
-function _getBreakpoints(state) {
-  return getBreakpoints(state)
-    .map(bp => updateLocation(state, bp))
-    .filter(
-      bp => bp.location.source && !bp.location.source.get("isBlackBoxed")
-    );
-}
+const _getBreakpoints = createSelector(
+  getBreakpoints,
+  state => state,
```
jasonLaster: does this negate the selector?

src/reducers/sources.js
```diff
@@ -284,8 +284,12 @@ export function getSourceByURL(state: OuterState, url: string) {
   return state.sources.sources.find(source => source.get("url") === url);
 }
 
+export function getSourceInSources(sources: I.Map<string, Source>, id: string) {
```
jasonLaster: nit: `getSourceByID`

src/reducers/sources.js
```diff
@@ -284,8 +284,12 @@ export function getSourceByURL(state: OuterState, url: string) {
   return state.sources.sources.find(source => source.get("url") === url);
```
jasonLaster: we should have a separate URL to ID mapping

src/reducers/sources.js
asolove: Well the problem is we already have a selector `getSourceById` that takes the full state and an id. This function takes just the actual sources map and an id. For various kinds of callers, we need both.

jasonLaster: Yeah - the underscore was meh when it was a private function, but below it's public... so i don't love it

asolove: Yeah, not sure what to do without finding a better name for either the top-level "sources state" or the nested "sources by id map".

fflorent:
@asolove I heard you also recently made a patch that is not landed yet in Firefox Dev Edition and fixed a big performance issue. And at work, I could notice a really good improvement of the performance already in master using the repository. Thanks a lot for this big effort! :)



Florent

jasonLaster:
@fflorent babel certainly adds a slow down, but it is worth it to have some consistency across runtimes. We don't use it to add experimental language features.

PR #2799, opened 2017-05-03
Fix tabs reselect

jasonLaster:
### Summary of Changes

* big change for a small fix. 
* added lots of types
* updated the integration test to check for tabs in the UI

### DIFF
--- a/src/components/Editor/Tabs.js
+++ b/src/components/Editor/Tabs.js
@@ -1,6 +1,6 @@
 // @flow
 
-import { DOM as dom, PropTypes, Component, createFactory } from "react";
+import { DOM as dom, PureComponent, createFactory } from "react";
 import { connect } from "react-redux";
 import { bindActionCreators } from "redux";
 import * as I from "immutable";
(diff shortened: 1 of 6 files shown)


src/components/Editor/Tabs.js
```diff
@@ -82,13 +88,29 @@ class SourceTabs extends PureComponent {
...
+    closeTab: string => any,
+    closeTabs: List<string> => any,
+    toggleProjectSearch: () => any,
+    togglePrettyPrint: string => any,
+    togglePaneCollapse: () => any,
+    showSource: string => any,
+    horizontal: boolean,
+    startPanelCollapsed: boolean,
+    endPanelCollapsed: boolean
+  };
```
jasonLaster: switches from proptypes to props

src/components/Editor/Tabs.js
```diff
@@ -166,6 +188,11 @@ class SourceTabs extends PureComponent {
     const sourceTab = sourceTabs.find(t => t.get("id") == tab);
     const tabURLs = sourceTabs.map(thisTab => thisTab.get("url"));
     const otherTabURLs = otherTabs.map(thisTab => thisTab.get("url"));
+
+    if (!sourceTab) {
+      return;
+    }
```
jasonLaster: flow doesn't know that we'll have a source in the tab list...

bomsy:
Looks Good!

Issue #2804, opened 2017-05-03
[flow] switch from PropTypes to Props


jasonLaster:
We would like to switch from using PropTypes to flow props because it gives us a lot more coverage. I switched the [Tabs] component and learned a lot in the process. 

We've done most of them, and just have a few to go!

### todolist

- [x] Frames/index.js
- [ ] CommandBar.js
- [ ] TextSearch.js
- [x] ProjectSearch/index.js
- [ ] Editor/index.js
- [ ] App.js

[Tabs]:https://github.com/devtools-html/debugger.html/pull/2799/files#diff-bab1d2a309d9243dde45e9093e4ec799R91

clarkbw:
Cool!  Lemme try one!  I'll take Footer

AnshulMalik:
I'll try one too.

Andarist:
wondering how does it work, is `props` stripped down later? or converted to `PropTypes` by a transform?

jasonLaster:
@Andarist it's stripped away. It's basically acknowledging that we're going to use flow and static type our code. Want to jump in and claim a file?

Andarist:
Yeah, already cloned and looking at `EventListeners` and reading dev guides ;)

Would be cool if they could be transformed to `PropTypes` if u distribute this as module (whatever if umd, cjs or smth else). You might have switched to `flow`, consumers of this as library might not.

Andarist:
@jasonLaster 
can try `SearchBar` next

jasonLaster:
Go for it

sharathnarayanph:
I'd to like to take up "Expressions"

Andarist:
Im also working on `HitMarker`

peterschussheim:
@jasonLaster Starting work on `shared/Popover`

jasonLaster:
Thanks!

anbarasiu:
Is this an updated list? I think I'll start with shared/Accordion and shared/Autocomplete.

jasonLaster:
probably not, we try to keep it up to date :)

ohana54:
I'll take `Popover`

satiewaltz:
I'll take object inspector!

jasonLaster:
@satiewaltz its yours

satiewaltz:
@jasonLaster  
[...]

PR #2805, opened 2017-05-04
Switch Footer from PropTypes to flow props

clarkbw:
Associated Issue: #2804 

### Summary of Changes

* convert from PropTypes to props

### Test Plan

Example test plan:

- [x] `yarn test`
- [x] `yarn run flow`
- [x] Clicking pretty print still works
- [x] Clicking black box still works


jasonLaster:
I like the test plan

PR #2826, opened 2017-05-05
Remove expand actor caching

bomsy:
Associated Issue:

Based on discussions with @jasonLaster, we decided to take out all the expand and actor caching for now, due to various issues found. We would be looking at better ways to fix this interesting problem in the near future.
One noticeable change would be that during stepping the expand states of the scope tree is no longer maintained, the user would have to expand the scopes again ... this way we are guaranteed to have correct scope information.

### Summary of Changes

* Remove all the expand and actor caching across files

image

### DIFF
--- a/src/components/SecondaryPanes/Scopes.js
+++ b/src/components/SecondaryPanes/Scopes.js
@@ -16,9 +16,6 @@ function info(text) {
   return dom.div({ className: "pane-info" }, text);
 }
 
-let expandedCache = new Set();
-let actorsCache = [];
-
 class Scopes extends PureComponent {
   state: {
     scopes: any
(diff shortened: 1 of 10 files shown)


jasonLaster:
the only test to fail is the `browser_dbg-scopes-mutations.js` test, which failed because we changed the behavior

src/components/SecondaryPanes/Scopes.js
```diff
@@ -16,9 +16,6 @@ function info(text) {
   return dom.div({ className: "pane-info" }, text);
 }
 
-let expandedCache = new Set();
-let actorsCache = [];
```
jasonLaster: it *is* nice to see module state going away

bomsy: yeah i agree! from the point were it went in, i felt like it was ugly code..:)

Issue #2691, opened 2017-04-19
[preview] don’t show undefined values


jasonLaster:
<img src='https://shipusercontent.com/aa54fe52842946ec3480e9777577f712/Screen%20Shot%202017-04-19%20at%204.21.06%20PM.png' title='Screen Shot 2017-04-19 at 4.21.06 PM.png' width=537>

jasonLaster:
@peterschussheim  is working on it

peterschussheim:
@jasonLaster Recorded a demo of the patch I am working on, resolving unit and snapshot failures now and will open PR shortly.

![Image of test](http://g.recordit.co/Ad5kC1bOEI.gif)

jasonLaster:
nice, looks great!

PR #2829, opened 2017-05-05
Don't show undef

peterschussheim:
Associated Issue: #2691 

### Summary of Changes

 

- [x] Don't show undefined values of variables in current scope when debugger is paused
- [x] Add selected-token class to popoverTarget in `componentDidMount`
- [x] Remove selected-token class from popoverTarget in `componentWillUnmount`

### Test Plan
- [x] When paused in the debugger, hovering over an undefined variable, does not add highlighting class or the popup preview.

### DIFF
--- a/src/utils/editor/expression.js
+++ b/src/utils/editor/expression.js
@@ -65,10 +65,8 @@ export function previewExpression({
 
   if (variables.has(tokenText)) {
     let variableKey = variables.get(tokenText);
-    if (variableKey.contents) {
-      if (variableKey.contents.value.type === "undefined") {
-        return null;
-      }
+    if (!get(variableKey, "contents.value.type") == "undefined") {
+      return null;
     }
 
     return variables.get(tokenText);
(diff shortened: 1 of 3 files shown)


src/utils/editor/expression.js
```diff
@@ -64,6 +64,11 @@ export function previewExpression({
   }
 
   if (variables.has(tokenText)) {
+    let variableKey = variables.get(tokenText);
+    if (!get(variableKey, "contents.value.type") == "undefined") {
+      return null;
+    }
+
```
jasonLaster: the next step is to look at the expression and see if its value is undefined too :)

don't worry, we'll be cleaning this code up soon :P

src/utils/editor/expression.js
```diff
@@ -64,6 +64,11 @@ export function previewExpression({
   }
 
   if (variables.has(tokenText)) {
+    let variableKey = variables.get(tokenText);
```
jasonLaster: I think we should try writing a test for this: 

here is the test file: https://github.com/peterschussheim/debugger.html/blob/dont-show-undef-latest/src/utils/tests/expression.js#L5-L31

peterschussheim: OK, I will work on this today.

jasonLaster:
Thanks @peterschussheim! I'm really impressed by your persistence. It was great to pair with you wednesday and get to the bottom of it

PR #2830, opened 2017-05-05
Converted PropTypes to flow types in EventListeners.js

Andarist:
### Summary of Changes

* Converted PropTypes to flow types in EventListeners.js

### Test Plan

`yarn flow` :trollface: 

### TODO
so I think we could extract `Listener` type so it could be used by selectors etc. Wasnt sure where it should be located though. Also what should be in fact `Listener` is now extended by me with `breakpoint` property, as that is what `mapStateToProps` returns (extended `Listener`) - so this should ideally become some composite type of `Listener` and `{ breakpoint: ?Breakpoint }`

### DIFF
--- a/src/components/SecondaryPanes/EventListeners.js
+++ b/src/components/SecondaryPanes/EventListeners.js
@@ -1,6 +1,6 @@
 // @flow
 import React from "react";
-const { DOM: dom, PropTypes, Component } = React;
+const { DOM: dom, Component } = React;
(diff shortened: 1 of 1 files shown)


jasonLaster:
I agree that we should share `Listener`, but lets do that when we want to use it next :)

EventListeners is a prototype, so this is doesn't need to be cleaned up at the moment. Great first flow PR!

PR #2831, opened 2017-05-06
Converted PropTypes to flow types in SearchBar.js

Andarist:
* Converted PropTypes to flow types in `SearchBar.js`
* extracted few types in other files too, so they can be imported in `SearchBar.js`
* added `equals` method on `Record`'s type

### DIFF
--- a/src/actions/sources.js
+++ b/src/actions/sources.js
@@ -127,7 +127,7 @@ function loadSourceMap(generatedSource) {
   };
 }
 
-type SelectSourceOptions = { tabIndex?: number, line?: number };
+export type SelectSourceOptions = { tabIndex?: number, line?: number };
 
 /**
  * Deterministically select a source that has a given URL. This will
--- a/src/actions/types.js
+++ b/src/actions/types.js
@@ -30,6 +30,8 @@ export type ThunkArgs = {
   sourceMaps: any
 };
 
+export type Thunk = ThunkArgs => void;
+
 export type ActionType = Object | Function;
 
 /**
--- a/src/actions/types.js
+++ b/src/actions/types.js
@@ -30,7 +30,7 @@ export type ThunkArgs = {
   sourceMaps: any
 };
 
-export type Thunk = ThunkArgs => void;
+export type Thunk = ThunkArgs => any;
 
 export type ActionType = Object | Function;
 
--- a/src/actions/ui.js
+++ b/src/actions/ui.js
@@ -6,6 +6,7 @@ import {
   getFileSearchState
 } from "../selectors";
 import type { ThunkArgs } from "./types";
+import type { SymbolSearchType } from "../reducers/ui";
 
 export function toggleProjectSearch(toggleValue?: boolean) {
   return ({ dispatch, getState }: ThunkArgs) => {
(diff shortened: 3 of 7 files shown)


Andarist:
actually such prop - `toggleSymbolSearch: boolean => ThunkArgs => void` is invalid, right? should be `toggleSymbolSearch: boolean => void`, as `toggleSymbolSearch` is wrapped already by `bindActionCreators`, gonna ammend later

jasonLaster:
Rerunning bc it was probably a flakey test

Andarist:
@jasonLaster what about [this](https://github.com/devtools-html/debugger.html/pull/2831#issuecomment-299636699)?

jasonLaster:
I'm away from my computer. Tough to say:/

Andarist:
So I was thinking about this even more. In fact I had it partially right - actual type of those pops is:
`(...args: any[]) => (ThunkArgs => any)`
because `dispatch` is returning an `action` (what got dispatched)

but thats just in this use case, treating components as app agnostic it should be specified as:
`(...args: any[]) => any` (as long as component is not interested in the prop's return value)

because component doesnt really care about actions and thunks, function prop is kinda its output API and in that regard I think those return types should be skipped (specified as `any`)

WDYT?

jasonLaster:
I agree. The flow+redux docs cover this as well. I like the types you
currently have :)

Andarist:
I've just disagreed with what is in the PR, you have agreed with me and said at the same time that you like whats in the PR :P

jasonLaster:
Sorry, perhaps send a gist with the other style so we can compare. We're new to typing the components and thus some of these actions this way.

Andarist:
so now in the PR we have such props flow types, all of these are dispatch wrapped action creators
```javascript
props: {
    selectSource: (string, ?SelectSourceOptions) => Thunk,
    toggleFileSearch: (?boolean) => Thunk,
    toggleFileSearchModifier: string => void,
    toggleSymbolSearch: boolean => Thunk,
    setSelectedSymbolType: SymbolSearchType => Thunk,
    setFileSearchQuery: string => void,
    updateSearchResults: ({ count: number, index?: number }) => void
};
```
But actually as far as the component is concerned - they are just functions, not action creators, not thunks, just functions. From its perpective it only requires certain argument types because it will call those functions with such data - it is its  defined API. On the other hand their return types are completely irrelevant for it (in most cases, because in most cases component just call a function prop and doesnt use its return value for anything) - because from its perpective any function (with specified argument types) can just plug in to it. So such thinking led me to the thought that those should be in fact specified as
```javascript
props: {
    selectSource: (string, ?SelectSourceOptions) => any,
    toggleFileSearch: (?boolean) => any,
    toggleFileSearchModifier: string => any,
    toggleSymbolSearch: boolean => any,
    setSelectedSymbolType: SymbolSearchType => any,
    setFileSearchQuery: string => any,
    updateSearchResults: ({ count: number, index?: number }) => any
};
```

PS. those 2 PRs were my first attempts at using flow and typing components in general, so there is a chance Im talking bs ;p

jasonLaster:
Thanks for the writeup. One of the best things about getting help from smart people is that they take the time to think through some of these new conventions. I realize that you're new to flow, but you're doing a great job of articulating the trade-offs. 

I'm personally persuaded that the return type does not matter in this context as the components fire and forget actions. They never use the return. 

Given this, lets have all of our actions return `any` and keep them simple. I perhaps can be persuaded later that they are important, but for now I think this is correct in this context.

Andarist:
gonna adjust those typings soon and probably will also include adjusted those sent in the previous PR for `EventListeners`

why the `contextTypes` revert?

jasonLaster:
the debugger wasn't loading with it. I think that react needs to use the contextTypes field to register shortcuts...

Andarist:
ah, right - forgot about this :<

I've just pushed my changes, and I think my work here is done, if you have any other suggestions just tell me.

jasonLaster:
Thanks @Andarist! This was really great work

Issue #2444, opened 2017-03-24
[DarkTheme] Improve the select background color


jasonLaster:
The blue selected background color is pretty jarring and unnecessary. It could be much softer and have the same effect.

### Old
![Screen Shot 2017-03-08 at 10.53.09 AM.png](https://shipusercontent.com/b71ac606b30835b8b3db4f119cde0833/Screen%20Shot%202017-03-08%20at%2010.53.09%20AM.png)

### New
It would be nice to change the tree and result item

I chose two colors here: `--theme-body-background` and `--theme-selection-background-semitransparent`. The result item in the list has a weird border now so more would need to be done, but it's a start.

[colors](https://github.com/devtools-html/debugger.html/blob/master/docs/local-development.md#colors)

<img src='https://shipusercontent.com/38f84270db7cedd510bfb42d44b88dda/Screen%20Shot%202017-03-24%20at%2012.52.29%20PM.png' title='Screen Shot 2017-03-24 at 12.52.29 PM.png' width=702>

diessica:
Working on this!

PR #2836, opened 2017-05-07
Improve source tree and search results

diessica:
Associated Issue: #2444

Also fixes bugs related to weird borders, which were collapsing.

### Test Plan
- [x] Command-P opens the panel and navigates in results
- [x] Command-F opens the search and navigates in search results
- [x] Chrome
- [x] Firefox

### Screenshots
Final result can be seen below.

screencast 2017-05-07 at 1 48 03 am

### DIFF
--- a/src/components/shared/ManagedTree.css
+++ b/src/components/shared/ManagedTree.css
@@ -25,6 +25,10 @@
   background-color: var(--theme-selection-background);
 }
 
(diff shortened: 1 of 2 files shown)


jasonLaster:
Thanks for all the help, you should join our slack if you haven't already!

PR #2839, opened 2017-05-07
refactor frames

jasonLaster:
### Summary of Changes

* refactors props
* adds stories
* moves the `copyStackTrace` logic into frames so Frame doesn't need to know about the list of frames

### DIFF
--- a/src/components/SecondaryPanes/Frames/index.js
+++ b/src/components/SecondaryPanes/Frames/index.js
@@ -15,7 +15,12 @@ import _Group from "./Group";
 const Group = createFactory(_Group);
 
 import actions from "../../../actions";
-import { annotateFrame, collapseFrames } from "../../../utils/frame";
+import {
+  annotateFrame,
+  collapseFrames,
+  formatCopyName
+} from "../../../utils/frame";
+import { copyToTheClipboard } from "../../../utils/clipboard";
 
 import {
   getFrames,
@@ -37,6 +42,7 @@ class Frames extends Component {
 
   renderFrame: Function;
   toggleFramesDisplay: Function;
+  copyStackTrace: Function;
 
   constructor(...args) {
     super(...args);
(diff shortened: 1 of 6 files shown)


src/components/SecondaryPanes/Frames/index.js
```diff
@@ -84,13 +95,15 @@ class Frames extends Component {
           frameOrGroup.id
             ? FrameComponent({
                 frame: frameOrGroup,
+                copyStackTrace: () => this.copyStackTrace(),
```
jasonLaster: I think this would be better if I bound the function and then just passed it in...

src/test/integration/utils/shared.js
```diff
@@ -7,12 +7,9 @@ const selectors = {
...
   scopeValue: i => `.scopes-list .tree-node:nth-child(${i}) .object-value`,
-  expressionNode: i =>
-    `.expressions-list .tree-node:nth-child(${i}) .object-label`,
-  expressionValue: i =>
-    `.expressions-list .tree-node:nth-child(${i}) .object-value`,
-  expressionClose: i =>
-    `.expressions-list .expression-container:nth-child(${i}) .close-btn`,
+  expressionNode: i => `.expressions-list .tree-node:nth-child(${i}) .object-label`,
+  expressionValue: i => `.expressions-list .tree-node:nth-child(${i}) .object-value`,
+  expressionClose: i => `.expressions-list .expression-container:nth-child(${i}) .close-btn`,
```
jasonLaster: not sure where this is coming from but prettier is clearly favoring this cleanup

src/components/SecondaryPanes/Frames/index.js
Andarist: did u guys consider using [property initializers](https://babeljs.io/docs/plugins/transform-class-properties/)? `this` problem would go away ;)

PR #2848, opened 2017-05-08
Use jest-localstorage-mock module

clarkbw:
Instead of our own custom local storage code in the test setup this uses (my own) jest-localstorage-mock module.


src/test/integration/utils/shared.js
```diff
@@ -7,9 +7,12 @@ const selectors = {
...
-  expressionNode: i => `.expressions-list .tree-node:nth-child(${i}) .object-label`,
-  expressionValue: i => `.expressions-list .tree-node:nth-child(${i}) .object-value`,
-  expressionClose: i => `.expressions-list .expression-container:nth-child(${i}) .close-btn`,
+  expressionNode: i =>
+    `.expressions-list .tree-node:nth-child(${i}) .object-label`,
+  expressionValue: i =>
+    `.expressions-list .tree-node:nth-child(${i}) .object-value`,
+  expressionClose: i =>
+    `.expressions-list .expression-container:nth-child(${i}) .close-btn`,
   expressionNodes: ".expressions-list .tree-node",
```
jasonLaster: hmm - i think you're using an older version of prettier

clarkbw: I've got
```
$ ./node_modules/.bin/prettier --version
1.3.1
```

jasonLaster: hmm, me too. I don't get it. I'll merge and we can see what's going on later

Issue #2562, opened 2017-04-06
[OutlineView] style the list of functions


jasonLaster:
* style the list of functions. 
  * It might be nice to show function parens too. This bug is for that separately #2735 
* jump to the source location with [`selectSource`](https://github.com/devtools-html/debugger.html/blob/master/src/actions/sources.js#L156-L158). 

This is a rough sketch of how to use previewFunction
```diff
diff --git a/src/components/Outline.js b/src/components/Outline.js
index 6b9299d..7b15730 100644
--- a/src/components/Outline.js
+++ b/src/components/Outline.js
@@ -8,6 +8,8 @@ import { getSelectedSource, getSourceText } from "../selectors";
 import { isEnabled } from "devtools-config";
 import { getSymbols } from "../utils/parser";
 import "./Outline.css";
+import previewFunction from "../shared/previewFunction";
+

 import type { Record } from "../utils/makeRecord";
 import type { SourceText } from "debugger-html";
@@ -37,7 +39,7 @@ class Outline extends Component {
   }

   renderFunction(func) {
-    return dom.li({ key: func.id }, func.value);
+    return dom.li({ key: func.id }, previewFunction(func));
   }
```

<img src='https://shipusercontent.com/beabce573cddd7574cc4555b1b13ce26/Screen%20Shot%202017-04-07%20at%201.55.14%20PM.png' title='Screen Shot 2017-04-07 at 1.55.14 PM.png' width=786>

gabrielluong:
Gonna continue working on the outline view

jasonLaster:
Closing this as we can handle the params in their own bug :)

PR #2854, opened 2017-05-09
OutlineView style the list of functions

codehag:
Associated Issue: #2562 

### Summary of Changes
This is the styling ticket for the OutlineList view, which introduces a function list for a given file

- style the list of functions
- jump to the source location 

### Test Plan
How to test:

- After starting the app and navigating to localhost:8000, click on the "settings" in the sidebar, and make sure that the "Source Outline" feature is enabled. 
- launch a browser and open a debugger from localhost:8000
- click through the source tree until you reach a .js file
- you should see an outline at the bottom

this needs a few more things like integrations tests

Feel free to improve it!

screen shot 2017-05-09 at 6 27 12 pm

### DIFF
--- a/src/components/Outline.css
+++ b/src/components/Outline.css
@@ -1,0 +1,8 @@
+.outline-list {
+  list-style-type: "-";
+}
+
+.outline-list__element {
+  color: blue;
+  padding-left: 0.5rem;
+}
--- a/src/components/Outline.js
+++ b/src/components/Outline.js
@@ -8,6 +8,7 @@ import { getSelectedSource, getSourceText } from "../selectors";
 import { isEnabled } from "devtools-config";
 import { getSymbols } from "../utils/parser";
 import "./Outline.css";
+import previewFunction from "./shared/previewFunction";
 
 import type { Record } from "../utils/makeRecord";
 import type { SourceText } from "debugger-html";
@@ -37,7 +38,13 @@ class Outline extends Component {
   }
 
   renderFunction(func) {
-    return dom.li({ key: func.id }, func.value);
+    return dom.li(
+      {
+        key: func.id,
+        className: "outline-list__element"
+      },
+      previewFunction(func)
+    );
   }
 
   renderFunctions() {
@@ -60,7 +67,7 @@ class Outline extends Component {
 
     return dom.div(
       { className: "outline" },
-      dom.ul({}, this.renderFunctions())
+      dom.ul({ className: "outline-list" }, this.renderFunctions())
     );
   }
 }
--- a/src/components/Outline.js
+++ b/src/components/Outline.js
@@ -37,11 +37,18 @@ class Outline extends Component {
     }
   }
 
+  selectItem(location) {
+    const selectedSourceId = this.props.selectedSource.get("id");
+    const startLine = location.start.line;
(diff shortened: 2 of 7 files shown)


src/components/Outline.js
```diff
@@ -50,7 +64,7 @@ class Outline extends Component {
 
     return functions
       .filter(func => func.value != "anonymous")
-      .map(this.renderFunction);
+      .map(func => this.renderFunction(func));
```
AnshulMalik: Does this change make any difference ?

src/components/Outline.js
```diff
@@ -36,8 +37,21 @@ class Outline extends Component {
     }
   }
 
+  selectItem(location) {
+    const selectedSourceId = this.props.selectedSource.get("id");
```
jasonLaster: generally we destructure our props so that it is clear what our `this` deps are in the func

```
const {selectedSource, selectSource} = this.props;
```

src/components/Outline.js
jasonLaster: Either approach is fine. We're generally prefer function references when there are no params`map(this.renderFunction)`, but we don't lint for it and either is fine.

src/components/tests/Outline.js
```diff
@@ -59,4 +76,26 @@ describe("Outline", () => {
...
+
+  it("should select a line of code in the current file on click", async () => {
+    const component = shallow(new OutlineComponent({ selectSource }));
+    const sourceId = "id";
+    const selectedSource = {
+      get: () => sourceId
+    };
+    const startLine = 12;
+
+    symbolDeclarations.functions[0] = {
```
jasonLaster: perhaps we also have a `generateSymbolDeclaration` factory as well so we can have the defaults in there

src/components/tests/Outline.js
```diff
@@ -59,4 +76,26 @@ describe("Outline", () => {
...
+      id: "my_example_function1:21",
+      value: "my_example_function1",
+      location: generateFuncLocation(startLine)
+    };
+
+    component.setProps({ sourceText, selectedSource });
```
jasonLaster: could we pass selectedSource in when we initially render the component? 

Perhaps not based on how we're fetching the symbols, which is a shame i suppose. And if that's the case, we probably move symbols to a reducer so that we can have the symbols passed down alongside the source :) (we can do this in a new issue ofcourse)

src/components/Outline.js
codehag: yes, it binds the `this` value. I just took a look through a few of the other files though and I noticed that doing the binding in the constructor is more common... that be better?

ie)
```
constructor(props) {
  super(props);
  this.renderFunction = this.renderFunction.bind(this)
}
```

both achieve the same result, as far as I know anyway. Happy to change it!

codehag:
- use object destructuring for props
- create a generateSymbolDeclaration function
- refactor the tests so that there is a clear render with overrides
function
- remove aspects of the test that gave the wrong impression about how
the code was working
- add comments

src/components/Outline.css
```diff
@@ -0,0 +1,8 @@
+.outline-list {
+  list-style-type: "-";
+}
+
+.outline-list__element {
+  color: blue;
```
jasonLaster: is this still used?

PR #2862, opened 2017-05-10
Use the license checker

tromey:
### Summary of Changes

* Add devtools-license-checker dependency and command
* Change circle.yml to invoke it

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -7,6 +7,7 @@
 .*node_modules/documentation.*
 <PROJECT_ROOT>/firefox/.*
 .*flow-coverage-report/.*
+.*custom_format_broken.json
 
 [libs]
 ./src/global-types.js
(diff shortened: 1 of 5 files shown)


jasonLaster:
@tromey i'm seeing this error:

screen shot 2017-05-10 at 9 27 39 pm

Perhaps we need to exclude the file in our `.flowconfig`

tromey:
I wonder why I don't see that.  I did see it in devtools-core (I updated the flow config there).  Anyway, will fix.

tromey:
Oh, I know why.  The tests fail for me locally, which I knew wasn't due to my patch, so I pushed with `--no-verify`.  Which skipped flow.  Sorry about that.

tromey:
Had to `--no-verify` again, so let's see.

tromey:
Ignoring the percy thing as discussed.

PR #2869, opened 2017-05-10
add a story for preview

jasonLaster:

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -8,6 +8,7 @@
     "transform-es2015-spread",
     "transform-flow-strip-types",
     "transform-async-to-generator",
+    "syntax-trailing-function-commas",
 
     ["module-resolver", {
       "alias": {
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "flow-redux": "flow-coverage-report  -i 'src/reducers/*.js' -i 'src/actions/*.js'  -t text",
     "flow-react": "flow-coverage-report -i 'src/components/**/*.js' -t text",
     "storybook": "start-storybook -p 9001 -c .storybook",
-    "snapshot": "NODE_ENV='development' build-storybook && percy-storybook",
+    "snapshot": "NODE_ENV='development' build-storybook && percy-storybook --ignore=\"node_modules/(?!devtools-)\"",
     "prepush": "yarn lint && yarn test && yarn flow",
     "postmerge": "node bin/post-merge.js",
     "precommit": "lint-staged",
@@ -107,6 +107,7 @@
   "devDependencies": {
     "@kadira/storybook": "^2.35.3",
     "devtools-license-check": "^0.2.0",
+    "babel-plugin-syntax-trailing-function-commas": "^6.22.0",
     "enzyme": "^2.8.0",
     "husky": "^0.13.2",
     "jest": "^20.0.0",
--- a/src/components/stories/Preview.js
+++ b/src/components/stories/Preview.js
@@ -1,0 +1,145 @@
+import React, { DOM as dom } from "react";
+import { storiesOf, action } from "@kadira/storybook";
+import _Preview from "../Editor/Preview";
+const Preview = React.createFactory(_Preview.WrappedComponent);
+import { L10N } from "devtools-launchpad";
+import { setValue } from "devtools-config";
+import * as I from "immutable";
+
+import "../App.css";
+import "../SecondaryPanes/Frames/Frames.css";
+import "devtools-launchpad/src/lib/themes/dark-theme.css";
+
+const obj = {
+  actor: "server2.conn45.child1/pausedobj81",
+  type: "object",
+  class: "Object",
+  ownProperties: {
+    cid: {
+      value: "view4"
+    },
+    model: {
+      value: {
+        type: "object",
+        actor: "server2.conn45.child1/pausedobj82",
+        class: "Object",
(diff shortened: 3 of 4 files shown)


.babelrc
```diff
@@ -8,6 +8,7 @@
     "transform-es2015-spread",
     "transform-flow-strip-types",
     "transform-async-to-generator",
+    "syntax-trailing-function-commas",
```
jasonLaster: this is helpful to support some issues in node and does not generally affect the browsers

package.json
```diff
@@ -40,7 +40,7 @@
     "flow-redux": "flow-coverage-report  -i 'src/reducers/*.js' -i 'src/actions/*.js'  -t text",
     "flow-react": "flow-coverage-report -i 'src/components/**/*.js' -t text",
     "storybook": "start-storybook -p 9001 -c .storybook",
-    "snapshot": "NODE_ENV='development' build-storybook && percy-storybook",
+    "snapshot": "NODE_ENV='development' build-storybook && percy-storybook --ignore=\"node_modules/(?!devtools-)\"",
```
jasonLaster: this was related to the babel plugin, which now transpiles devtools-reps which had a trailing comma :)

src/components/stories/Preview.js
```diff
@@ -0,0 +1,145 @@
...
+    }
+  },
+  prototype: {
+    type: "object",
+    actor: "server2.conn45.child1/pausedobj426",
+    class: "Object",
+    ownPropertyLength: 14
+  },
+  ownSymbols: []
+};
```
jasonLaster: we can simplify this fixture later

src/components/stories/Preview.js
```diff
@@ -0,0 +1,149 @@
...
+        },
+        options
+      )
+    )
+  );
+}
+
+const stories = storiesOf("Preview", module);
+
+const options = [{}, { dir: "rtl" }, { theme: "dark" }];
```
jasonLaster: I think we can move this setup into a shared util as we add more examples

package.json
nchevobbe: so do we still need it ? It looks like you added some modules to take care of trailing commas

src/components/stories/Preview.js
```diff
@@ -0,0 +1,149 @@
+import React, { DOM as dom } from "react";
+import { storiesOf, action } from "@kadira/storybook";
+import _Preview from "../Editor/Preview";
+const Preview = React.createFactory(_Preview.WrappedComponent);
```
nchevobbe: nit: I find this a bit confusing at first(`_Preview` and `Preview`), but I guess we have no other choice than doing something like that, and I can't find a better naming

src/components/stories/Preview.js
```diff
@@ -0,0 +1,149 @@
...
+    type: "object",
+    actor: "server2.conn45.child1/pausedobj426",
+    class: "Object",
+    ownPropertyLength: 14
+  },
+  ownSymbols: []
+};
+
+const loadedObjects = {
+  "server2.conn45.child1/pausedobj81": obj
```
nchevobbe: nit: What do you think of : 
```
const loadedObjects = {
  [obj.actor]: obj
}
``` 
?

src/components/stories/Preview.js
```diff
@@ -0,0 +1,149 @@
...
+};
+
+if (typeof window == "object") {
```
nchevobbe: Could we have a comment explaining why we add those globals ?

src/components/stories/Preview.js
```diff
@@ -0,0 +1,149 @@
...
+  document.dir = dir;
+  document.body.parentNode.className = themeClass;
+  const div = document.createElement("div");
+  document.body.appendChild(div);
+  div.id = "foobar";
+  div.style.top = "100px";
+  div.style.left = "100px";
+  div.width = "100px";
+  div.height = "20px";
+  div.style.background = "blue";
```
nchevobbe: do we still need this one ?

src/components/stories/Preview.js
```diff
@@ -0,0 +1,149 @@
...
+  const { dir, theme } = option;
+  const optionLabel = dir || theme || "";
+  stories
+    .add(`simple Object ${optionLabel}`, () => {
+      setValue("features.previewWatch", false);
+      return PreviewFactory(
+        {
+          value: obj,
+          expression: "this",
+          loadedObjects: I.Map().set("server2.conn45.child1/pausedobj81", obj)
```
nchevobbe: nit: maybe we can re-use `obj.actor` instead of passing the string directly

src/components/stories/Preview.js
```diff
@@ -0,0 +1,149 @@
...
+        option
+      );
+    })
+    .add(`simple Object with Input ${optionLabel}`, () => {
+      setValue("features.previewWatch", true);
+      return PreviewFactory(
+        {
+          value: obj,
+          expression: "this",
+          loadedObjects: I.Map().set("server2.conn45.child1/pausedobj81", obj)
```
nchevobbe: same here

package.json
jasonLaster: I think we do. This means that we do use babel on `devtools-` projects

src/components/stories/Preview.js
jasonLaster: yep - it's a sad but necessary pattern for now

Issue #2817, opened 2017-05-05
[breakpoints] Nightly won't delete breakpoints


ekr:
I have some breakpoints set on a script. I can X them out and they appear to go away but when I reload the page, they are back and checked.

jasonLaster:
re-opening as i'm seeing deleted breakpoints, re-activate when the debugger is reloaded in the launchpad and re-opened in the panel.

ekr:
Yah, i am still seeing this.

jasonLaster:
Btw, we haven't updated nightly yet. So its just testable on GitHub

ekr:
Ah. OK.

codehag:
done!

PR #2877, opened 2017-05-12
Refactor Pending Breakpoints to fix reloading

jasonLaster:
Associated Issue: #2817

### Summary of Changes

* Refactor how pending breakpoints are stored and updated in the reducer
* add lots of tests
* update checkPendingBreakpoints

### DIFF
--- a/src/actions/sources.js
+++ b/src/actions/sources.js
@@ -105,7 +105,7 @@ export function newSources(sources: Source[]) {
       source => !getSource(getState(), source.id)
     );
 
-    for (const source in filteredSources) {
+    for (let source of filteredSources) {
       await dispatch(newSource(source));
     }
   };
--- a/src/actions/sources.js
+++ b/src/actions/sources.js
@@ -75,8 +75,8 @@ async function checkPendingBreakpoints(state, dispatch, source) {
     return;
   }
 
-  const pendingBreakpointsList = pendingBreakpoints.valueSeq().toJS();
-  for (let pendingBreakpoint of pendingBreakpointsList) {
+  const pendingBreakpointsArray = pendingBreakpoints.valueSeq().toJS();
+  for (let pendingBreakpoint of pendingBreakpointsArray) {
     await checkPendingBreakpoint(state, dispatch, pendingBreakpoint, source);
   }
 }
--- a/src/actions/sources.js
+++ b/src/actions/sources.js
@@ -46,29 +46,38 @@ function checkSelectedSource(state, dispatch, source) {
   }
 }
 
-function checkPendingBreakpoints(state, dispatch, source) {
+async function checkPendingBreakpoint(
+  state,
+  dispatch,
+  pendingBreakpoint,
+  source
+) {
+  const {
+    location: { line, sourceUrl, column },
+    condition
+  } = pendingBreakpoint;
+  const sameSource = sourceUrl && sourceUrl === source.url;
+  const location = { sourceId: source.id, sourceUrl, line, column };
+  const bp = getBreakpoint(state, location);
+
+  if (sameSource && !bp) {
+    if (location.column && isEnabled("columnBreakpoints")) {
+      await dispatch(addBreakpoint(location, { condition }));
+    } else {
+      await dispatch(addBreakpoint(location, { condition }));
+    }
+  }
+}
+
+async function checkPendingBreakpoints(state, dispatch, source) {
   const pendingBreakpoints = getPendingBreakpoints(state);
+  if (!pendingBreakpoints) {
+    return;
+  }
 
-  if (pendingBreakpoints) {
-    pendingBreakpoints.forEach(pendingBreakpoint => {
-      const {
-        location: { line, sourceUrl, column },
-        condition
-      } = pendingBreakpoint;
-      const sameSource = sourceUrl && sourceUrl == source.url;
-
-      const location = { sourceId: source.id, sourceUrl, line, column };
-
-      const bp = getBreakpoint(state, location);
-
-      if (sameSource && !bp) {
-        if (location.column && isEnabled("columnBreakpoints")) {
-          dispatch(addBreakpoint(location, { condition }));
-        } else {
-          dispatch(addBreakpoint(location, { condition }));
-        }
-      }
-    });
+  const pendingBreakpointsList = pendingBreakpoints.valueSeq().toJS();
+  for (let pendingBreakpoint of pendingBreakpointsList) {
+    await checkPendingBreakpoint(state, dispatch, pendingBreakpoint, source);
   }
 }
 
@@ -78,23 +87,27 @@ function checkPendingBreakpoints(state, dispatch, source) {
  * @static
  */
 export function newSource(source: Source) {
-  return ({ dispatch, getState }: ThunkArgs) => {
+  return async ({ dispatch, getState }: ThunkArgs) => {
     if (prefs.clientSourceMapsEnabled) {
-      dispatch(loadSourceMap(source));
+      await dispatch(loadSourceMap(source));
     }
 
     dispatch({ type: constants.ADD_SOURCE, source });
 
     checkSelectedSource(getState(), dispatch, source);
-    checkPendingBreakpoints(getState(), dispatch, source);
+    await checkPendingBreakpoints(getState(), dispatch, source);
   };
 }
 
 export function newSources(sources: Source[]) {
-  return ({ dispatch, getState }: ThunkArgs) => {
-    sources
-      .filter(source => !getSource(getState(), source.id))
-      .forEach(source => dispatch(newSource(source)));
+  return async ({ dispatch, getState }: ThunkArgs) => {
+    const filteredSources = sources.filter(
+      source => !getSource(getState(), source.id)
+    );
+
+    for (const source in filteredSources) {
+      await dispatch(newSource(source));
+    }
   };
 }
 
--- a/src/actions/sources.js
+++ b/src/actions/sources.js
@@ -88,11 +88,11 @@ async function checkPendingBreakpoints(state, dispatch, source) {
  */
 export function newSource(source: Source) {
   return async ({ dispatch, getState }: ThunkArgs) => {
+    dispatch({ type: constants.ADD_SOURCE, source });
+
     if (prefs.clientSourceMapsEnabled) {
       await dispatch(loadSourceMap(source));
     }
-
-    dispatch({ type: constants.ADD_SOURCE, source });
 
     checkSelectedSource(getState(), dispatch, source);
     await checkPendingBreakpoints(getState(), dispatch, source);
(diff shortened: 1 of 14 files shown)


src/actions/sources.js
```diff
@@ -78,23 +87,27 @@ function checkPendingBreakpoints(state, dispatch, source) {
...
-      .filter(source => !getSource(getState(), source.id))
-      .forEach(source => dispatch(newSource(source)));
+  return async ({ dispatch, getState }: ThunkArgs) => {
+    const filteredSources = sources.filter(
+      source => !getSource(getState(), source.id)
+    );
+
+    for (const source in filteredSources) {
+      await dispatch(newSource(source));
+    }
```
jasonLaster: we need to update this before we merge!

src/actions/tests/breakpoints.js
```diff
@@ -24,11 +51,40 @@ const simpleMockThreadClient = {
...
+
+    await dispatch(actions.addBreakpoint(loc1));
+    bps = selectors.getBreakpoints(getState());
+    expect(bps.size).to.be(1);
+  });
 
-    await dispatch(actions.addBreakpoint({ sourceId: "a", line: 5 }));
-    await dispatch(actions.addBreakpoint({ sourceId: "b", line: 6 }));
+  it("when the user adds a sliding breakpoint, a corresponding pending breakpoint should be added", async () => {
+    const { dispatch, getState } = createStore(slidingMockThreadClient);
```
codehag: nitpick: I think it might be easy to break this, by not getting the right offset between slidingMockThreadClient and slidLocation. We can explicitly associate these two (or more) stub elements by using a wrapper function. Here is an example of how it might be done:

```
function generateMockThreadClient (offset = 0) {
  return { 
    setBreakpoint: (location, condition) => {
        const actualLocation = Object.assign({}, location, {
          line: location.line + offset
        });
        resolve({ id: makeLocationId(location), actualLocation, condition });
      });
    }
  };
};

function generateOffsetThreadClient(offset, location) => {
  const slidingMockThreadClient = generateSlidingMockThreadClient(offset);
  const slidLocation = Object.assign({}, location, { line: location.line + offset });
  return { slidingMockThreadClient, slidLocation };
}

function ... () { // wherever this will be used
  const location = { sourceId: "a", line: 5 };
  const { slidingMockThreadClient, slidLocation } = generateOffsetThreadClient(2, location);
  const { dispatch, getState } = createStore(slidingMockThreadClient);
  ... // and so on
 }
```

naming needs to be improved. it might make sense to make location an array, but i think its getting ahead of ourselves

src/actions/tests/pending-breakpoints.js
```diff
@@ -0,0 +1,261 @@
...
+  } = breakpoint;
+
+  return {
+    location: { sourceUrl, line, column },
+    condition,
+    disabled
+  };
+}
+
+function slideMockBp(bp) {
```
codehag: same comment here as above on line 79 in breakpoints.js --- probably we can do this in a future pr, where we create a breakpoints utils

src/actions/tests/pending-breakpoints.js
```diff
@@ -0,0 +1,261 @@
...
+    });
+  },
+
+  removeBreakpoint: id => {
+    return new Promise((resolve, reject) => {
+      resolve({ status: "done" });
+    });
+  },
+
+  setBreakpointCondition: (id, location, condition, noSliding) => {
```
codehag: nitpick: we probably can remove all of the arguments for this method. also the above methods have unused variables

src/actions/tests/pending-breakpoints.js
```diff
@@ -0,0 +1,261 @@
...
+    return new Promise((resolve, reject) => {
+      resolve({ sourceId: "a", line: 5 });
+    });
+  }
+};
+
+describe("pending breakpoints", () => {
+  it("when the user adds a breakpoint, a corresponding pending breakpoint should be added", async () => {
+    const { dispatch, getState } = createStore(simpleMockThreadClient);
```
codehag: these three lines are present in most of the following tests, they can be moved to a beforeEach

src/actions/tests/pending-breakpoints.js
```diff
@@ -0,0 +1,261 @@
...
+    const bp = generateBreakpoint("foo");
+    const id = makePendingLocationId(bp.location);
+
+    await dispatch(actions.addBreakpoint(bp.location));
+    const pendingBps = selectors.getPendingBreakpoints(getState());
+    expect(pendingBps.size).to.be(2);
+    expect(pendingBps.get(id)).to.eql(generatePendingBreakpoint(bp));
+  });
+
+  it("when the user adds a sliding breakpoint, a corresponding pending breakpoint should be added", async () => {
```
codehag: Looks good. Unlike the above two tests, this can be its own case (outside of the other two tests that have the same setup)
doesnt require a before each until we have another test with these setup conditions

src/actions/tests/pending-breakpoints.js
```diff
@@ -0,0 +1,261 @@
...
+  describe("when two or more breakpoints are added", () => {
+    let breakpoint1;
+    let breakpoint2;
+    let breakpointId1;
+    let breakpointId2;
+
+    beforeEach(() => {
+      breakpoint1 = generateBreakpoint("foo");
+      breakpoint2 = generateBreakpoint("foo2");
+      breakpointId1 = makePendingLocationId(breakpoint1.location);
```
codehag: nitpick: now that I understand the relationship between `locationId` and `id` in the breakpoints, maybe this should be rewritten as `breakpointLocationId1` and `breakpointLocationId2`

src/actions/tests/pending-breakpoints.js
```diff
@@ -0,0 +1,261 @@
...
+
+      const pendingBps = selectors.getPendingBreakpoints(getState());
+      expect(pendingBps.has(breakpointId1)).not.to.be(true);
+      expect(pendingBps.has(breakpointId2)).to.be(true);
+    });
+  });
+
+  it("when the user disables a breakpoint, the corresponding pending breakpoint is also disabled", async () => {
+    const { dispatch, getState } = createStore(simpleMockThreadClient);
+    const bp = generateBreakpoint("foo");
```
codehag: these two lines(171 and 172) can be moved into a beforeEach. describe block can be `synced actions between breakpoints and pendingBreakpoints`

src/actions/tests/pending-breakpoints.js
```diff
@@ -0,0 +1,261 @@
...
+    const id = makePendingLocationId(bp.location);
+
+    await dispatch(actions.addBreakpoint(bp.location));
+    await dispatch(
+      actions.setBreakpointCondition(bp.location, { condition: "2" })
+    );
+    const bps = selectors.getPendingBreakpoints(getState());
+    const breakpoint = bps.get(id);
+    expect(breakpoint.condition).to.be("2");
+  });
```
codehag: proposed describe block `synced actions between breakpoints and pendingBreakpoints` should end here

src/reducers/breakpoints.js
```diff
@@ -20,7 +20,7 @@ import type { Record } from "../utils/makeRecord";
 
 export type BreakpointsState = {
   breakpoints: I.Map<string, Breakpoint>,
-  pendingBreakpoints: ?I.Map<string, any>,
+  pendingBreakpoints: any,
```
codehag: this should be a `I.Map<string, PendingBreakpoint` -- for this we need to make a `PendingBreakpoint type`

src/reducers/breakpoints.js
```diff
@@ -56,6 +56,12 @@ export function makeLocationId(location: Location) {
   return `${sourceId}:${line}:${column}`;
 }
 
+export function makePendingLocationId(location: Location) {
+  let { sourceUrl, line, column } = location;
+  column = column || "";
```
codehag: since this is the only reassigned variable and since we are not using es6 object shorthand we can instead use consts and make this a new variable like `const columnString = column || "";`

src/reducers/breakpoints.js
```diff
@@ -87,31 +91,9 @@ function update(state: Record<BreakpointsState> = State(), action: Action) {
...
-            loading: false
-          })
-        );
-      } else if (action.status === "error") {
-        return state.deleteIn(["breakpoints", id]);
-      }
-
-      break;
+      const newState = setCondition(state, action);
+      setPendingBreakpoints(newState);
```
codehag: great!

src/reducers/breakpoints.js
```diff
@@ -176,55 +155,101 @@ function addBreakpoint(state, action) {
...
-        updateObj(bp, {
-          loading: false,
-          disabled: true
-        })
-      );
+  const id = makeLocationId(action.breakpoint.location);
+  const pendingId = makePendingLocationId(action.breakpoint.location);
+  let updatedState = undefined;
+
+  if (action.disabled) {
```
codehag: since these two cases are so different, perhaps they can be functions that return `updatedState`? then we won't need to do the weird `let`

src/reducers/breakpoints.js
```diff
@@ -176,55 +155,101 @@ function addBreakpoint(state, action) {
...
+}
+
+function slideBreakpoint(state, action) {
+  const { actualLocation, id } = action.value;
+  const { breakpoint } = action;
+  const currentBp = state.breakpoints.get(id) || fromJS(breakpoint);
+
+  const locationId = makeLocationId(breakpoint.location);
+  const movedLocationId = makeLocationId(actualLocation);
+  state = state.deleteIn(["breakpoints", locationId]);
```
codehag: Personal comment: Its not wrong, I just cringe if I see state being reassigned :P

src/reducers/breakpoints.js
```diff
@@ -176,55 +155,101 @@ function addBreakpoint(state, action) {
...
   const { location: { sourceUrl, line, column }, condition, disabled } = bp;
 
   const location = { sourceUrl, line, column };
   return { condition, disabled, location };
 }
 
-function filterByNotLoading(bp: any): boolean {
-  return !bp.loading;
+function setPendingBreakpoints(state) {
+  prefs.pendingBreakpoints = state.pendingBreakpoints;
```
codehag: :+1:

src/reducers/breakpoints.js
```diff
@@ -176,55 +155,101 @@ function addBreakpoint(state, action) {
...
+  const id = makePendingLocationId(breakpoint.location);
+  return state.setIn(
+    ["pendingBreakpoints", id],
+    makePendingBreakpoint(breakpoint)
+  );
 }
 
 function restorePendingBreakpoints() {
-  return prefs.pendingBreakpoints;
+  return I.Map(prefs.pendingBreakpoints);
```
codehag: :+1::+1::+1::+1::+1::+1:

Issue #2800, opened 2017-05-03
[ProjectSearch] avoid unnecessary actions 


jasonLaster:
We currently fire `TOGGLE_PROJECT_SEARCH` actions every time we select a source even if the project search state won't change.

We can fix this with this diff.

```diff
diff --git a/src/actions/ui.js b/src/actions/ui.js
index cebe234..5aad7a7 100644
--- a/src/actions/ui.js
+++ b/src/actions/ui.js
@@ -9,17 +9,22 @@ import type { ThunkArgs } from "./types";

 export function toggleProjectSearch(toggleValue?: boolean) {
   return ({ dispatch, getState }: ThunkArgs) => {
-    if (toggleValue != null) {
-      dispatch({
-        type: constants.TOGGLE_PROJECT_SEARCH,
-        value: toggleValue
-      });
-    } else {
-      dispatch({
+    const projectSearchState = getProjectSearchState(getState());
+    if (toggleValue === null) {
+      return dispatch({
         type: constants.TOGGLE_PROJECT_SEARCH,
-        value: !getProjectSearchState(getState())
+        value: !projectSearchState
       });
     }
+
+    if (projectSearchState == toggleValue) {
+      return;
+    }
+
+    dispatch({
+      type: constants.TOGGLE_PROJECT_SEARCH,
+      value: toggleValue
+    });
   };
 }
 ```
 
 We can test this by looking in the redux devtools and seeing what is fired

karthikJagadeesh:
I'll take this one.

karthikJagadeesh:
After doing the changes, the test fails for this function.

jasonLaster:
Mind opening the PR so we can discuss there?

Feel free to push with --no-verify

karthikJagadeesh:
Sure.

PR #2882, opened 2017-05-12
fixed the unnecessary firing of TOGGLE_PROJECT_SEARCH action

karthikJagadeesh:
Associated Issue: #2800

I made the necessary changes looking at the issue description. However, the test fails when I try to push the changes.

Am only beginning to learn Redux now, how do I proceed further?

### DIFF
--- a/src/actions/ui.js
+++ b/src/actions/ui.js
@@ -16,7 +16,7 @@ export function toggleProjectSearch(toggleValue?: boolean) {
         type: constants.TOGGLE_PROJECT_SEARCH,
         value: !projectSearchState
       });
-
+    }
 
     if (projectSearchState == toggleValue) {
       return;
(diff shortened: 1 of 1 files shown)


src/actions/ui.js
```diff
@@ -10,17 +10,22 @@ import type { SymbolSearchType } from "../reducers/ui";
...
   return ({ dispatch, getState }: ThunkArgs) => {
-    if (toggleValue != null) {
-      dispatch({
-        type: constants.TOGGLE_PROJECT_SEARCH,
-        value: toggleValue
-      });
-    } else {
-      dispatch({
+    const projectSearchState = getProjectSearchState(getState());
+    if (toggleValue === null) {
```
jasonLaster: my bet is that this should be `undefined`.

Here's the failing test. https://github.com/jasonLaster/debugger.html/blob/preview-story/src/actions/tests/ui.js#L15-L22

You can run it with jest. `jest src/actions/` and make changes to debug it :)

karthikJagadeesh: Test passes after changing null to undefined!

jasonLaster:
Great

PR #2885, opened 2017-05-13
 Switch CommandBar from PropTypes to Flow Props 

sharathnarayanph:
Associated Issue: #2804 

Summary of Changes 

- Changed PropTypes to Flow Props

Test Plan  

-  yarn run test-all


src/components/SecondaryPanes/CommandBar.js
```diff
@@ -98,6 +97,21 @@ function debugBtn(onClick, type, className, tooltip, disabled = false) {
 }
 
 class CommandBar extends Component {
+  props: {
+    sources: Object,
+    selectedSource: Object,
```
jasonLaster: I think we can pull in types from the reducers here: `Source`, `SourceRecord`, or `SourcesMap`...

src/components/SecondaryPanes/CommandBar.js
```diff
@@ -98,6 +97,21 @@ function debugBtn(onClick, type, className, tooltip, disabled = false) {
...
+    resume: () => any,
+    stepIn: () => any,
+    stepOut: () => any,
+    stepOver: () => any,
+    breakOnNext: () => any,
+    pause: Map<string, any>,
```
jasonLaster: perhaps here too...

src/components/SecondaryPanes/CommandBar.js
```diff
@@ -244,23 +258,8 @@ class CommandBar extends Component {
...
-  pause: ImPropTypes.map,
-  pauseOnExceptions: PropTypes.func,
-  shouldPauseOnExceptions: PropTypes.bool,
-  shouldIgnoreCaughtExceptions: PropTypes.bool,
-  isWaitingOnBreak: PropTypes.bool
-};
-
 CommandBar.contextTypes = {
-  shortcuts: PropTypes.object
+  shortcuts: Object
```
jasonLaster: wow, does this work?

src/components/SecondaryPanes/CommandBar.js
sharathnarayanph: image

In the Pause reducer, the 'pause' prop is not of type Map. Does this mean I need to convert this to a Map (like Map<string, Pause> )or should I change the above file (CommandBar)ile to just 'Pause' object instead of Map type?

src/components/SecondaryPanes/CommandBar.js
sharathnarayanph: This one works. I was able to access the data while debugging and even the keystrokes worked. Should I be checking for anything else?

src/components/SecondaryPanes/CommandBar.js
sharathnarayanph: Thanks. I have added these.

src/components/SecondaryPanes/CommandBar.js
jasonLaster: I suppose, that the keyboard shortcuts work :P 

otherwise awesome

src/components/SecondaryPanes/CommandBar.js
jasonLaster: Yeah - lets use the same types as the reducer

PR #2890, opened 2017-05-14
Fix Object Inspector toggling

jasonLaster:
### Summary of Changes

* We had some issues where variables would not expand when you clicked on a label. The reason for that was that we had a special event handler for label clicks which was overriding the default.
* There are three components to look at when testing (Scopes, Expressions, Preview). The GIF below tests all three.

### DIFF
--- a/src/components/SecondaryPanes/Scopes.js
+++ b/src/components/SecondaryPanes/Scopes.js
@@ -52,10 +52,7 @@ class Scopes extends PureComponent {
       scopeInspector = ObjectInspector({
         roots: scopes,
         getObjectProperties: id => loadedObjects.get(id),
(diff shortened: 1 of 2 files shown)


src/components/SecondaryPanes/Scopes.js
```diff
@@ -52,10 +52,7 @@ class Scopes extends PureComponent {
       scopeInspector = ObjectInspector({
         roots: scopes,
         getObjectProperties: id => loadedObjects.get(id),
-        loadObjectProperties: loadObjectProperties,
-        onLabelClick: (item, { expanded, setExpanded }) => {
```
codehag: great, looks good!

jasonLaster:
Killing code left and right

PR #2892, opened 2017-05-14
fix weird fromJS bug

jasonLaster:
### Summary of Changes

* fixes a scary wtf bug in `fromJS` caused by a value that happens to have a `hasOwnProperty` field that we want to format. Note, this is one of the perils of building a product for JS in JS.

### Test Plan

* unit test the hell out of it

### DIFF
--- a/src/utils/fromJS.js
+++ b/src/utils/fromJS.js
@@ -7,6 +7,21 @@
  */
 
 const Immutable = require("immutable");
+const isFunction = require("lodash/isFunction");
+
+// hasOwnProperty is defensive because it is possible that the
+// object that we're creating a map for has a `hasOwnProperty` field
+function hasOwnProperty(value, key) {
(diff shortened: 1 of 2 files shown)


src/utils/fromJS.js
```diff
@@ -7,6 +7,21 @@
  */
 
 const Immutable = require("immutable");
+const isFunction = require("lodash/isFunction");
+
+// hasOwnProperty is defensive because it is possible that the
+// object that we're creating a map for has a `hasOwnProperty` field
+function hasOwnProperty(value, key) {
```
codehag: this feels weird, do we have a write up on what the issue is? 

Since it solves the issue, and it is used in a controlled way, lets merge it for now -- and investigate later?

jasonLaster: We don't. I just made the PR. The test shows the issue though. Objects we want to create a map for can have the field `hasOwnProperty`

Issue #2857, opened 2017-05-10
[Preview] Default Properties are shown unecessarily


jasonLaster:
We currently show `[default properties]` when you're hovering over objects. This is especially bad because there are no default properties unless it is a window object.

<img src='https://shipusercontent.com/739592ff8529b61482bec496e69823c9/Screen%20Shot%202017-05-10%20at%2010.10.37%20AM.png' title='Screen Shot 2017-05-10 at 10.10.37 AM.png' width=506>

Here's some context in the relevant code

```diff
diff --git a/src/utils/object-inspector.js b/src/utils/object-inspector.js
index cc3c73e..1be39b7 100644
--- a/src/utils/object-inspector.js
+++ b/src/utils/object-inspector.js
@@ -63,6 +63,7 @@ function getPromiseProperties(item) {
   });
 }

+// here is the fnc for getting isDefault
 function isDefault(item) {
   return WINDOW_PROPERTIES.includes(item.name);
 }
@@ -119,6 +120,7 @@ function makeDefaultPropsBucket(props, parentPath, ownProperties) {
     )
   );

+  // we should change this to not add a default properties
   if (defaultProps.length > 0) {
     const defaultNodes = defaultProps.map((name, index) =>
       createNode(
@@ -128,6 +130,8 @@ function makeDefaultPropsBucket(props, parentPath, ownProperties) {
       )
     );
     nodes.push(
+      // while we're in here we should translate this string :)
+      // L10N.getStr()
       createNode("[default properties]", `${parentPath}/default`, defaultNodes)
     );
   }
   ```

bomsy:
looks like this is fixed!

PR #2897, opened 2017-05-15
Fix default bucketing

jasonLaster:
Associated Issue: #2857

### Summary of Changes

We are currently showing `[default properties]` incorrectly when an object happens to have a window property like location.

This fixes that by explicitly checking to see if the object is a window. 

### Test Plan

* adds unit tests
* adds storybook stories

### DIFF
--- a/src/utils/object-inspector.js
+++ b/src/utils/object-inspector.js
@@ -1,5 +1,3 @@
-// @flow
-
 const get = require("lodash/get");
 const { maybeEscapePropertyName } = require("devtools-reps");
 
--- a/src/utils/object-inspector.js
+++ b/src/utils/object-inspector.js
@@ -1,3 +1,5 @@
+// @flow
+
 const get = require("lodash/get");
 const { maybeEscapePropertyName } = require("devtools-reps");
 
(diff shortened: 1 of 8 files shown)


jasonLaster:
one mochitest error:
> 6874 INFO TEST-UNEXPECTED-FAIL | devtools/client/debugger/new/test/mochitest/browser_dbg-scopes.js | undefined assertion name - Got length, expected prototype

src/utils/object-inspector.js
```diff
@@ -160,8 +167,10 @@ function makeNodesForProperties(
       parentPath,
       ownProperties
     );
-  } else {
+  } else if (objProps.class == "Window") {
```
codehag: :+1:

src/utils/tests/object-inspector.js
```diff
@@ -105,6 +105,25 @@ describe("object-inspector", () => {
       expect(paths).to.eql(["root/bar", "root/__proto__"]);
     });
 
+    it("window object", () => {
```
codehag: Lets give this a clearer descriptions... something like `Sets default properties for the window object`?

Issue #2889, opened 2017-05-14
[WhyPaused] Update the look and feel of "why paused"


jasonLaster:
The "Why Paused" component is responsible for telling the user why the program is paused. It covers several cases (breakpoints, exceptions, event handlers), and several others. 

Users also learn about the pause state of the program from other components as well: pause overlay in the app, the debug line in the editor, and the call stack and scopes pane.

I have a couple of concerns with the look and feel of why paused. None of them are critical, but still worth discussing.

* Its background color and bold text give it a lot of importance, when it is often not necessary. For instance, most times your paused either on a breakpoint of you've stepped. In these cases, you do not really need "why paused".
* Its placement in the secondary pane is a bit awkward. It is in an accordion, but is not a pane. It also pushes the other panes down.

Here are a couple of ideas:

* Soften the appearance of the component. Use a lighter background and text styles.
* Not show "why paused" on breakpoints and stepping. Only show it in other cases like event handlers and exceptions.
* Style the debug line red or purple when paused on an exception. #3112
* Perhaps move the "why paused" component into the call stack where the bottom frame could give some additional context. #3111

Here's a slight adjustment with a lighter background.

<img src='https://shipusercontent.com/d8bcc47ad11b910db058b6208879dc1c/Screen%20Shot%202017-05-14%20at%2010.28.42%20AM.png' title='Screen Shot 2017-05-14 at 10.28.42 AM.png' width=302>

bomsy:
@jasonLaster is this done?

jasonLaster:
closing as we have more actionable issues now

PR #2906, opened 2017-05-15
Update paused styles

jasonLaster:
Associated Issue: #1948

### Summary of Changes

* updates the source editor to pull in the new debug line styles
* removes an unused style in the debugger
* updates why paused a bit, I think we'll want to do more going forward in  #2889 but atleast it's no longer an eye sore

### DIFF
--- a/src/components/Editor/Editor.css
+++ b/src/components/Editor/Editor.css
@@ -132,8 +132,8 @@ html[dir="rtl"] .editor-mount {
   color: var(--theme-content-color3);
 }
 
-.debug-line .CodeMirror-line {
-  background-color: var(--breakpoint-active-color) !important;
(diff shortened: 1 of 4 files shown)


src/components/Editor/Editor.css
```diff
@@ -132,8 +132,8 @@ html[dir="rtl"] .editor-mount {
   color: var(--theme-content-color3);
 }
 
-.debug-line .CodeMirror-line {
-  background-color: var(--breakpoint-active-color) !important;
+.debug-line .CodeMirror-linenumber {
+  background-color: transparent !important;
```
codehag: hmm looks like we might have a specificity war here -- might be something to work on at a later point

Issue #2899, opened 2017-05-15
[Preview] Slide popup horizontally so that it fits in the UI


jasonLaster:
We currently decide whether we want to show the preview popup below or above a token based on how much available space there is. We should try and so something similar horizontally where if the popup will be clipped we can slide it to the left until it fits 

```diff
diff --git a/src/components/shared/Popover.js b/src/components/shared/Popover.js
index 463e6fc..36a89eb 100644
--- a/src/components/shared/Popover.js
+++ b/src/components/shared/Popover.js
@@ -33,6 +33,9 @@ class Popover extends Component {
       top: targetTop
     } = this.props.target.getBoundingClientRect();

+    // We'll want to check if the popover will be clipped here
+    // we already do some math, but we need more!
+
     // width division corresponds to calc in Popover.css
     const left = targetLeft + targetWidth / 2 - width / 5;
     const dir = targetBottom + height > window.innerHeight ? "up" : "down";
```

PR #2911, opened 2017-05-16
Align preview horizontally

AnshulMalik:
Associated Issue: #2899 

### Summary of Changes

* Try to create a Arrow component, I am thinking of using this in code-folding as well
* The popop now slides left when it overflows 

### Test Plan

- [x] Checked in both up and down directions

horizontal


src/components/shared/Popover.js
```diff
@@ -15,16 +17,18 @@ class Popover extends Component {
...
       : this.getTooltipCoords();
 
     // eslint-disable-next-line react/no-did-mount-set-state
-    this.setState({ left, top, dir });
+    this.setState({ left, top, dir, targetMid });
   }
 
   getPopoverCoords() {
     const el = ReactDOM.findDOMNode(this);
+    const wrapper = document.getElementsByClassName("editor-wrapper")[0];
```
AnshulMalik: I am not sure whether this is the right way to do it

src/components/shared/Popover.js
```diff
@@ -15,16 +17,18 @@ class Popover extends Component {
...
+    const { left, top, dir, targetMid } = type == "popover"
       ? this.getPopoverCoords()
       : this.getTooltipCoords();
 
     // eslint-disable-next-line react/no-did-mount-set-state
-    this.setState({ left, top, dir });
+    this.setState({ left, top, dir, targetMid });
   }
 
   getPopoverCoords() {
```
codehag: So this function is doing _a lot_ now -- it looks like it has grown without refactoring. lets break it down a bit so its easier for others to read. I will walk through how I do such a refactoring, which decisions I would make and why. Feel free to comment!

Reading through, it looks like it is doing work with the following (ie, these are its inputs):
- getting the position of the editor (it is named `wrapper` in the code)
- getting the position of the popover ( it is named `el` in the code, I will call it `popoverPosition` since that its what the element is, and that is what we are interested in)
- getting the position of a target element passed through the props (its destructured, so lets call it target`position`)

and at the end of this function, we want to return the following:
`{ left, top, dir, targetMid }`

thats already plenty to keep in mind. If we write that out explicitly it will look something like this: 
```
getPopoverCoords() {
    const popoverElement = ReactDOM.findDOMNode(this);
    const wrapper = document.getElementsByClassName("editor-wrapper")[0];

    const wrapperPosition = wrapper.getBoundingClientRect();
    const popoverPosition = el.getBoundingClientRect();
    const targetPosition = this.props.target.getBoundingClientRect();
    // ..
    // const left = calculateLeft(wrapperPosition, popoverPosition, targetPosition);
    // const top = 
    // const dir = 
    // const targetMid = 
    return { left, top, dir, targetMid }; // somehow we want to get here
}
```
I removed the destructuring, because it was making it more difficult to read. By using the objects in their whole form, it is easier to keep track of what is happening, and what belongs where.

So! somehow we need to get those four elements -- at the moment we are using variable reassignment to do that. However the logic is a bit complex, and it is not clear what each step is doing. Ternary statements are great, but usually you want to avoid doing calculation in a ternary statement. 

```
// original
let left = targetLeft + targetWidth / 2 - width / 5;
const isOverflowingRight = left + width > wrapperRight;
left = isOverflowingRight ? wrapperRight - width : left;
```
The original does what it needs to do, but its hard to tell what is coming from where -- what is the target at this point? which width? what does left mean?

So a proposal would be to break this out into a function, and do each step, one line at a time. For each line we should explicitly state what we are doing.
```
// broken out into a function
const 
function calculateLeft(target, wrapper, popover) {
  const leftOffset =  (target.width / 2) - (popover.width / 5);
  const estimatedLeft = target.left + leftOffset;
  const estimatedRight = left + popover.width;
  const isOverflowingRight = estimatedRight > wrapper.width;
  if (isOverflowingRight) {
    const adjustedLeft = wrapper.right - estimatedWidth
    return adjustedLeft;
  }
  return estimatedLeft;
}
```

While the second version is longer, its functional, making it is easier to read, and easier to test. It also tells a story -- anyone can read this and will know what is happening right away and why. The same should be done for each of the ternary statements, `dir` and `top`

here is an example for the dir:
```
function calculateVerticalOrientation(target, popover) {
  const estimatedHeight = targetBottom + height;
  if (estimatedHeight > window.innerHeight) return "up";
  return "down";
}
```
It isnt much longer, but it tells us much more. `dir` is pretty ambiguous by itself, another coder might come and misunderstand what is meant there. By being explicit, we can shorten the time other people need to understand our code, and improve on it.

let me know if you have any questions!!

src/components/shared/Popover.js
```diff
@@ -65,14 +75,21 @@ class Popover extends Component {
 
   renderPopover() {
     const { onMouseLeave } = this.props;
-    const { top, left, dir } = this.state;
+    const { top, left, dir, targetMid } = this.state;
+
+    const arrow = new BracketArrow({
+      dir: dir === "up" ? "down" : "up",
```
codehag: So here we also want to so something similar to the comment that i made up on `getPopoverCoords()` -- maybe give this a stab on your own and see how this can be refactored for readability?

src/components/shared/Popover.js
codehag: good intuition! I agree, using the document API feels weird for this. Let me investigate a bit and I will get back to you

src/components/shared/Popover.js
AnshulMalik: Yeah, current implementation is hard to understand, it'll be more readable if we split the `getPopoverCoords()` function.

codehag:
Looks great! thanks for the work!

Issue #2843, opened 2017-05-07
[Frames] Add a group context menu


jasonLaster:
It would be nice if the context menu in the group frame offered the option to copy the stack trace and source URL for the top most frame. 

Extract the Frame Context Menu to its own component
```diff
diff --git a/src/components/SecondaryPanes/Frames/Frame.js b/src/components/SecondaryPanes/Frames/Frame.js
index 8798c8a..78093f3 100644
--- a/src/components/SecondaryPanes/Frames/Frame.js
+++ b/src/components/SecondaryPanes/Frames/Frame.js
@@ -53,41 +53,8 @@ export default class FrameComponent extends Component {
   }

   onContextMenu(event: SyntheticKeyboardEvent) {
-    const { copyStackTrace, frame } = this.props;
-    const copySourceUrlLabel = L10N.getStr("copySourceUrl");
-    const copySourceUrlKey = L10N.getStr("copySourceUrl.accesskey");
-    const copyStackTraceLabel = L10N.getStr("copyStackTrace");
-    const copyStackTraceKey = L10N.getStr("copyStackTrace.accesskey");
-
...
+    const { frame, copyStackTrace } = this.props;
+    FrameMenu(frame, { copyStackTrace });
   }
```

Add a new `Frames/Menu.js` file. This will need some dependencies and other things likely :)

```js
export default function FrameMenu(frame, { copyStackTrace }) {
  const copySourceUrlLabel = L10N.getStr("copySourceUrl");
  const copySourceUrlKey = L10N.getStr("copySourceUrl.accesskey");
  const copyStackTraceLabel = L10N.getStr("copyStackTrace");
  const copyStackTraceKey = L10N.getStr("copyStackTrace.accesskey");

  event.stopPropagation();
  event.preventDefault();

  const menuOptions = [];

  const source = frame.source;
  if (source) {
    const copySourceUrl = {
      id: "node-menu-copy-source",
      label: copySourceUrlLabel,
      accesskey: copySourceUrlKey,
      disabled: false,
      click: () => copyToTheClipboard(source.url)
    };

    menuOptions.push(copySourceUrl);
  }

  const copyStackTraceItem = {
    id: "node-menu-copy-source",
    label: copyStackTraceLabel,
    accesskey: copyStackTraceKey,
    disabled: false,
    click: () => copyStackTrace()
  };

  menuOptions.push(copyStackTraceItem);

  showMenu(event, menuOptions);
}
```

Add the context menu to the group:
[...]

PR #2912, opened 2017-05-16
[Frames] Add a group context menu       

codehag:
Associated Issue: #2843 

### Summary of Changes

* refactor Context Menu in Frame component
* refactor new FrameMenu component
* add tests to Frame component for mouse actions
* add tests to FrameMenu to make sure things work as expected

### Test Plan

behavior should not be changed, so integration tests should stay the test. Added a couple of unit tests

no visual changes

### DIFF
--- a/src/components/SecondaryPanes/Frames/FrameMenu.js
+++ b/src/components/SecondaryPanes/Frames/FrameMenu.js
@@ -2,27 +2,36 @@
 import { showMenu } from "devtools-launchpad";
 import { copyToTheClipboard } from "../../../utils/clipboard";
 import type { LocalFrame } from "./types";
+import type { ContextMenuItem } from "debugger-html";
+import { kebabCase } from "lodash";
 
-function formatMenuElement(label, accesskey, click, disabled = false) {
+function formatMenuElement(
+  labelString: string,
+  accesskeyString: string,
+  click: Function,
+  disabled: boolean = false
+): ContextMenuItem {
+  const label = L10N.getStr(labelString);
+  const id = `node-menu-${kebabCase(label)}`;
   return {
-    id: "node-menu-copy-source",
+    id,
     label,
-    accesskey,
+    accesskey: L10N.getStr(accesskeyString),
     disabled,
     click
   };
 }
 
 function copySourceElement(url) {
-  const label = L10N.getStr("copySourceUrl");
-  const key = L10N.getStr("copySourceUrl.accesskey");
-  return formatMenuElement(label, key, () => copyToTheClipboard(url));
+  return formatMenuElement("copySourceUrl", "copySourceUrl.accesskey", () =>
+    copyToTheClipboard(url)
+  );
 }
 
 function copyStackTraceElement(copyStackTrace) {
-  const label = L10N.getStr("copyStackTrace");
-  const key = L10N.getStr("copyStackTrace.accesskey");
-  return formatMenuElement(label, key, () => copyStackTrace());
+  return formatMenuElement("copyStackTrace", "copyStackTrace.accesskey", () =>
+    copyStackTrace()
+  );
 }
 
 export default function FrameMenu(
--- a/src/components/SecondaryPanes/Frames/FrameMenu.js
+++ b/src/components/SecondaryPanes/Frames/FrameMenu.js
@@ -3,16 +3,33 @@ import { showMenu } from "devtools-launchpad";
 import { copyToTheClipboard } from "../../../utils/clipboard";
 import type { LocalFrame } from "./types";
 
+function formatMenuElement(label, accesskey, click, disabled = false) {
+  return {
+    id: "node-menu-copy-source",
+    label,
+    accesskey,
+    disabled,
+    click
+  };
+}
+
+function copySourceElement(url) {
+  const label = L10N.getStr("copySourceUrl");
+  const key = L10N.getStr("copySourceUrl.accesskey");
+  return formatMenuElement(label, key, () => copyToTheClipboard(url));
+}
+
+function copyStackTraceElement(copyStackTrace) {
+  const label = L10N.getStr("copyStackTrace");
+  const key = L10N.getStr("copyStackTrace.accesskey");
+  return formatMenuElement(label, key, () => copyStackTrace());
+}
+
 export default function FrameMenu(
   frame: LocalFrame,
   copyStackTrace: Function,
   event: SyntheticKeyboardEvent
 ) {
-  const copySourceUrlLabel = L10N.getStr("copySourceUrl");
-  const copySourceUrlKey = L10N.getStr("copySourceUrl.accesskey");
-  const copyStackTraceLabel = L10N.getStr("copyStackTrace");
-  const copyStackTraceKey = L10N.getStr("copyStackTrace.accesskey");
-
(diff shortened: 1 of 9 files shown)


src/components/SecondaryPanes/Frames/tests/Frame.js
```diff
@@ -3,85 +3,73 @@ import { shallow } from "enzyme";
...
-      displayName: "renderFoo",
-      library: false,
-      location: {
-        line: 10
-      }
-    };
-    const selectedFrame = frame;
-    const selectFrame = jest.fn();
+import FrameMenu from "../FrameMenu";
+jest.mock("../FrameMenu", () => jest.fn());
```
jasonLaster: so that's how it works :)

src/components/SecondaryPanes/Frames/FrameMenu.js
```diff
@@ -0,0 +1,49 @@
+// @flow
+import { showMenu } from "devtools-launchpad";
+import { copyToTheClipboard } from "../../../utils/clipboard";
+import type { LocalFrame } from "./types";
+
+function formatMenuElement(label, accesskey, click, disabled = false) {
+  return {
+    id: "node-menu-copy-source",
```
jasonLaster: we need to have unique IDs.

I'm not sure how much this API buys us actually. Perhaps, we establish a convention where we give it a label, and it can make an ID and accessKey?

// devtools-modules/src/menu.js
```js
export function createItem({label, disabled = false, click }) {
  const id = `node-menu-${kebabCase(label)}`

  return {
    label,
    id,
   ...
  }
}
```

https://lodash.com/docs/4.17.4#kebabCase

src/components/SecondaryPanes/Frames/tests/FrameMenu.js
```diff
@@ -0,0 +1,73 @@
+const originalL10N = L10N;
+import FrameMenu from "../FrameMenu";
+import { showMenu } from "devtools-launchpad";
+import { copyToTheClipboard } from "../../../../utils/clipboard";
+jest.mock("devtools-launchpad", () => ({ showMenu: jest.fn() }));
+jest.mock("../../../../utils/clipboard", () => ({
+  copyToTheClipboard: jest.fn()
+}));
+
+function generateMockCopySource(label, accesskey) {
```
jasonLaster: this could be a nice general purpose helper for mocking a menu item.

With that said, I'm concerned we might be testing too much here. For instance, in this case I'd be happy to just check the IDs are what we expect:

`expect(items.map(i => i.d)).toEqual(["joe", "bob", "dan"])`, this tells us that the menu has the correct items and we can later test other properties that might be important like conditional disabled states...

src/components/SecondaryPanes/Frames/tests/FrameMenu.js
```diff
@@ -0,0 +1,73 @@
...
+    click: jasmine.any(Function) // eslint-disable-line no-undef
+  };
+}
+
+describe("FrameMenu", () => {
+  let mockEvent;
+  let mockFrame;
+
+  beforeEach(() => {
+    L10N = { getStr: jest.fn(value => value) };
```
jasonLaster: What is the benefit of mocking L10N? I'd prefer including it because it will give us sensible english :) 

If we change the props then our snapshots will change, which is kinda cool

src/components/SecondaryPanes/Frames/tests/Group.js
```diff
@@ -0,0 +1,41 @@
...
+  const defaultProps = {
+    group: [{ displayName: "foo" }],
+    selectedFrame: {},
+    selectFrame: jest.fn(),
+    copyStackTrace: jest.fn()
+  };
+
+  const props = Object.assign({}, defaultProps, overrides);
+  const component = shallow(new Group(props));
+  return { component, props };
```
jasonLaster: I really like this pattern your setting up

src/components/SecondaryPanes/Frames/tests/Group.js
```diff
@@ -0,0 +1,41 @@
...
+    expect(component).toMatchSnapshot();
+  });
+
+  describe("mouse events", () => {
+    it("calls FrameMenu on right click", () => {
+      const { component, props } = render();
+      const mockEvent = "mockEvent";
+      component.simulate("contextmenu", mockEvent);
+
+      expect(FrameMenu).toHaveBeenCalledWith(
```
jasonLaster: I don't feel strongly here, but I feel like testing the args is adding too much coupling. My concerns would be 

1. does it get called
2. does it return some menu items.

testing the communication seems like we could be introducing brittle tests...

src/components/SecondaryPanes/Frames/FrameMenu.js
codehag: ah cool, wasn't clear from the ticket -- convention sounds great!

src/components/SecondaryPanes/Frames/tests/FrameMenu.js
codehag: Hmm... if the formatting was taken care of elsewhere, I would agree that relying on the IDs is a much better pattern, but since `FrameMenu` is the responsible function, I think we should test that we get correctly formatted output. Testing the id's is insufficient, because if something is changed in FrameMenu, ie, someone mixes up the label and the key, we will get the wrong output -- and this would be hard to catch elsewhere.

However as you mentioned, this could be moved out into a utility -- if that happens, this level of testing should move with it!

src/components/SecondaryPanes/Frames/tests/FrameMenu.js
codehag: yeah im not too sure about mocking L10N -- it just was giving back undefined in the tests, and I wanted something a bit more meaningful... maybe im running the tests wrong?

src/components/SecondaryPanes/Frames/tests/Group.js
codehag: Hmmm i will take a look into this again. the function `FrameMenu` relies on does not return any elements... but maybe we could check via enzyme's `mount`? I agree that checking if the elements are there is preferable to checking what arguments are passed in, but in a pinch i think it is ok to check what is being passed to FrameMenu.

On the other hand, this would not be a unit test if we do this, but rather an integration test. I am torn, personally

src/components/SecondaryPanes/Frames/tests/FrameMenu.js
codehag: so, I left mocking the L10N in, because in our current setup, L10N returns undefined on getStr -- and this isnt meaningful enough. by mocking it, I also make sure that everything is being formatted properly.

src/components/SecondaryPanes/Frames/tests/FrameMenu.js
codehag: so! I think i found a nice compromise -- I added a `ContextMenuItem` type, and am now just testing for the ids. I also rewrote the functions a bit so that testing the id's should be sufficient for the formatting

src/components/SecondaryPanes/Frames/tests/Group.js
codehag: For now, i think we should leave this as it is

PR #2915, opened 2017-05-16
Update pause styles (try 2)

jasonLaster:
This updates the strategy for styling the pause line so that we use a new class instead of trying to override the old styles.

### DIFF
--- a/src/components/Editor/Editor.css
+++ b/src/components/Editor/Editor.css
@@ -1,8 +1,12 @@
-/* vim:set ts=2 sw=2 sts=2 et: */
+.editor-wrapper {
+  --debug-line-background: rgba(226, 236, 247, 0.5);
+  --debug-line-border: rgb(145, 188, 219);
+}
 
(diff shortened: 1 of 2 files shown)


src/components/Editor/Editor.css
```diff
@@ -1,8 +1,12 @@
-/* vim:set ts=2 sw=2 sts=2 et: */
+.editor-wrapper {
+  --debug-line-background: rgba(226, 236, 247, 0.5);
+  --debug-line-border: rgb(145, 188, 219);
+}
 
-/* This Source Code Form is subject to the terms of the Mozilla Public
- * License, v. 2.0. If a copy of the MPL was not distributed with this
- * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
```
clarkbw: Should probably leave this in

jasonLaster: that's what i was thinking, but I'd like to go through and add them to all the files and it's easier to be consistent :)

PR #2920, opened 2017-05-16
query CM from the dom

jasonLaster:
Fixes some issues referencing CM from the window

### DIFF
--- a/src/test/mochitest/head.js
+++ b/src/test/mochitest/head.js
@@ -214,7 +214,7 @@ function assertPausedLocation(dbg, source, line) {
 
   // Check the debug line
   ok(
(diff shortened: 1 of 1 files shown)


src/test/mochitest/head.js
```diff
@@ -741,3 +744,8 @@ function toggleCallStack(dbg) {
 function toggleScopes(dbg) {
   return findElement(dbg, "scopesHeader").click();
 }
+
+function getCM(dbg) {
```
juliandescottes: nit: getCM is a bit opaque, getCodeMirror ?

jasonLaster: damn - missed this. CM is not great, but it's not terrible in context :)

PR #2923, opened 2017-05-16
Move formatSymbol to searchbar

jasonLaster:
This is a small cleanup that I did to simplify the parser. The parser was previously "formatting" symbols for use by the search bar. That work is now moved into the search bar.


src/components/shared/previewFunction.js
```diff
@@ -6,18 +8,27 @@ import flatten from "lodash/flatten";
 
 import "./previewFunction.css";
 
-function getFunctionName(func) {
+import type { FunctionGrip } from "../../client/firefox/types";
+
+import type { ObjectInspectorItem } from "./ObjectInspector";
+import type { SymbolDeclaration } from "../../utils/parser/utils";
+type FunctionType =
+  | FunctionGrip
```
codehag: this looks a bit weird to me, should this have the `|` before `FunctionGrip`?

src/utils/parser/utils.js
```diff
@@ -113,7 +106,7 @@ function getNodeValue(node) {
   return node.name;
 }
 
-function getFunctionName(path) {
+function getFunctionName(path): string {
```
codehag: nice!

PR #2930, opened 2017-05-17
Fix storybook preview

jasonLaster:


src/components/stories/Preview.js
```diff
@@ -139,26 +139,29 @@ function PreviewFactory(options, { dir = "ltr", theme = "light" } = {}) {
...
-          loadObjectProperties: () => {},
-          onClose: action("onClose")
-        },
-        options
+    { className: "editor-wrapper" },
+    dom.div(
+      {
+        className: `preview ${themeClass}`,
+        dir,
+        style: {
```
codehag: any chance this can be moved into a class?

jasonLaster: hmm - we could, but it's a story so i would prefer to keep the styles here so it is clear. 

We have a stylesheet for another story, so we do it both ways

codehag: cool, thanks for clarifying!

jasonLaster: here's the css for another story: https://github.com/devtools-html/devtools-core/blob/master/packages/devtools-splitter/stories/index.css#L14

Issue #2355, opened 2017-03-13
[Editor] text search is slow for large files


jasonLaster:
We currently use codemirror to find all of the search results in a document when we start searching. Unfortunately, this is a pretty slow search that we could speed up pretty easily.

Here is what we do [here](https://github.com/devtools-html/debugger.html/blob/master/src/utils/editor/source-search.js#L37-L40)

```js
  state.results = [];
  while (cursor.findNext()) {
    state.results.push(cursor.pos);
  }
```

Here is the data structure that it generates.
```json
{
  "from": {
    "line": 15,
    "ch": 16,
    "sticky": null
  },
  "to": {
    "line": 15,
    "ch": 19,
    "sticky": null
  },
  "match": [
  {
    0: "ati"
    input: "getOriginalLocation, getGeneratedLocation, isOriginalId"
    index: 16
  }
  ]
}
```

We could probably speed it up by getting a count of matches and then using codemirror to find the next or previous result on demand.

wldcordeiro:
This is a good idea, we kind of do this already too with countMatches.

ryanjduffy:
Would be best if you could defer the line calculation until it was needed to scroll it into view. Otherwise, you have to no only find the string but also derive the line number. You could incrementally sort it out (e.g. count the line for the first match, cache it, count the lines from the first match to the second match and sum it to the cached value, etc) but it's likely a lot of wasted effort since in most cases the user doesn't need every match.

jasonLaster:
Yeah. That's a good idea. We probably can.

wldcordeiro:
I didn't get a chance to test it yet but it might be possible to do something kind of like this to achieve what you suggested @jasonLaster not sure how it performs versus what @ryanjduffy suggested.

```js
sourceText
  .split(/\\r?\\n/)
  .reduce((acc, line) => {
    acc.push(query // query being the regexp object we return from buildQuery
      .exec(line)
      .map(...)); // format like codemirror's returns.
    return acc;
  });
```

ryanjduffy:
I started on this using a variation of @wldcordeiro's suggestion. Initial search seemed to be faster with my rough algorithm but I was surprised when my naive 'only find the first instance in the line' approach succeeded in finding every instance in the line. Since that clearly couldn't have been the case, I discovered that the loop code in `getSearchState()` seems to be unnecessary as the incremental search in `doSearch()` and `searchNext` work fine without it.

wldcordeiro:
[...]

PR #2943, opened 2017-05-17
memoize source search

jasonLaster:
Associated Issue: #2355

### Summary of Changes

search is very slow now, this hopefully will help us as we explore speeding up search..


jasonLaster:
hmm, not sure this is helping

PR #2953, opened 2017-05-18
Update community calls

jasonLaster:
It would be nice to have a more european / asia friendly call time :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ We strive to make the Debugger as development friendly as possible. If you have 
 
 Say hello in [slack] or in the [#devtools-html][irc-devtools-html] channel on irc.mozilla.org.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -59,7 +59,7 @@ We strive to make the Debugger as development friendly as possible. If you have
 
 Say hello in [slack] or in the [#devtools-html][irc-devtools-html] channel on irc.mozilla.org.
 
-* **Community Call**: Every Tuesday, Thursday at 3pm EST. [Join the Hangout][community-call]
+* **Community Call**: Every Tuesday at 3pm EST and Thursday at 12pm EST. [Join the Hangout][community-call]
```
clarkbw: Should probably use **ET** instead of EST, right?

And the other one should be 12pm **PT**

clarkbw:
@jasonLaster I think you missed my review! 🔴

Issue #2448, opened 2017-03-24
[SoucesTree] Show the host domain first when the debugger opens


jasonLaster:
@Dalimil recently helped us show the host domain first in the source list when the debuggee navigates #1374. We should be able to also show the host domain first when the debugger opens by setting the debuggee url when the debugger opens

jasonLaster:
I took a stab at this this weekend and got the `debuggeeUrl` when the debugger connected. 

https://github.com/devtools-html/debugger.html/compare/master...jasonLaster:navigate-on-connect?expand=1

unfortunately, this breaks the source tree :P

@Dalimil would you like to take a look?

Dalimil:
The `SourcesTree.componentWillReceiveProps(nextProps)` function in `SourcesTree.js` seems to expect to be given additional sources via `nextProps.sources` every time it is called, but when `sources.size === 0` it recreates the tree - I'm not quite sure why, but the debuggeeUrl is part of the props and when it changes, this function is called with 0 sources, so it basically removes everything. 

I'm also getting different results when I call `actions.navigate(tabTarget._form.url);` in `firefox.js` immediately vs. when called after a scheduled setTimeout...

jasonLaster:
oh fascinating. so we need to be a bit stricter about when we build the tree. 

like slow down...

PR #2957, opened 2017-05-18
[SoucesTree] Show the host domain first when the debugger opens

Dalimil:
Associated Issue: #2448 

This seems to work.
But I'm still not sure why the navigate action removes all sources from the state.

### DIFF
--- a/src/actions/navigation.js
+++ b/src/actions/navigation.js
@@ -17,10 +17,14 @@ export function willNavigate(_, event) {
     await sourceMaps.clearSourceMaps();
     clearDocuments();
 
-    dispatch({
-      type: "NAVIGATE",
-      url: event.url
-    });
+    dispatch(navigate(event.url));
+  };
+}
+
+export function navigate(url) {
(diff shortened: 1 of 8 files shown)


jasonLaster:
We clear the sources here so that when the user reloads we can get new ones as they stream in:

https://github.com/devtools-html/debugger.html/blob/master/src/reducers/sources.js#L133-L139

We might want to change the action to CONNECT which will avoid the sources clearing

codehag:
Looks great! Just fixed up the test and rebased it off master to have the most recent changes. Thanks!

src/actions/navigation.js
```diff
@@ -17,10 +17,14 @@ export function willNavigate(_, event) {
...
 
-    dispatch({
-      type: "NAVIGATE",
-      url: event.url
-    });
+    dispatch(navigate(event.url));
+  };
+}
+
+export function navigate(url) {
```
wldcordeiro: 👍

PR #2958, opened 2017-05-18
5 19

jasonLaster:


juliandescottes:
It's great to bump the debugger version for the mc bundle drops! 

For the next bundle drop, could I suggest creating a tag as well and referencing the version number in the changeset summary in mc? Will make it easier to track the version used in mc.

PR #2959, opened 2017-05-18
Splitup the parser

jasonLaster:
### Summary of Changes

* Starts splitting up the parser so that it will be easier to reason about. 
* This is step one of probably 2 or 3. I'll queue up another PR tomorrow morning.
* I'd like to organize the new files into a *utils* and *queries* dir or something like that
* I'd like to move the tests and fixtures into this directory
* I'd also like to have a nice test index.html file for testing out the parser with the fixtures...


src/utils/parser/getSymbols.js
```diff
@@ -0,0 +1,94 @@
...
+  location: ASTLocation,
+  parameterNames?: string[]
+};
+
+export type SymbolDeclarations = {
+  functions: Array<SymbolDeclaration>,
+  variables: Array<SymbolDeclaration>
+};
+
+function getVariableNames(path): SymbolDeclaration[] {
```
wldcordeiro: If getFunctionNames gets moved, how come not this?

jasonLaster:
It's going soon too :)

src/utils/parser/getAst.js
```diff
@@ -0,0 +1,68 @@
...
+      console.warn("parse failed", text);
+    }
+
+    ast = {};
+  }
+
+  return ast;
+}
+
+export default function getAst(sourceText: SourceText) {
```
ryanjduffy: Seems like `getAst` isn't the right module name/default export if you're going to standardize on using `traverseAst` to fetch the AST and handle it. Suggest making this a named export and make the module be `./ast.js`

jasonLaster: that's a good point

src/utils/parser/getFunctionName.js
```diff
@@ -0,0 +1,28 @@
+export default function getFunctionName(path): string {
```
jasonLaster: I like it being separate because infering names is a special use case

Issue #2647, opened 2017-04-13
[flow] upgrade from 0.43 to 0.45


jasonLaster:
There are three minor issues 
  
  
  ```
  yarn flow v0.21.3
$ flow 
Launching Flow server for /home/ubuntu/debugger.html
Spawned flow server (pid=18553)
Logs will go to /tmp/flow/zShomezSubuntuzSdebugger.html.log
src/reducers/sources.js:179
179:   if (Object.keys(prefsTabs).length == 0) {
           ^^^^^^^^^^^^^^^^^^^^^^ element of Object.keys. Expected object instead of
179:   if (Object.keys(prefsTabs).length == 0) {
                       ^^^^^^^^^ empty array literal

src/utils/editor/source-search.js:204
204:     let cursor = getSearchCursor(cm, state.query, pos, modifiers);
                                          ^^^^^^^^^^^ null. This type is incompatible with the expected param type of
 12: function getSearchCursor(cm, query: string, pos, modifiers: SearchModifiers) {
                                         ^^^^^^ string

src/utils/editor/source-search.js:211
211:       cursor = getSearchCursor(cm, state.query, location, modifiers);
                                        ^^^^^^^^^^^ null. This type is incompatible with the expected param type of
 12: function getSearchCursor(cm, query: string, pos, modifiers: SearchModifiers) {
                                         ^^^^^^ string


Found 3 errors
error Command failed with exit code 2.

yarn flow returned exit code 1

```

arthur801031:
I'll take a stab at this.

PR #2961, opened 2017-05-19
[flow] upgrade from 0.43 to 0.46

arthur801031:
Associated Issue: #2647 

### Summary of Changes
* Upgrade flow from 0.43 to 0.46
* Fix flow errors

### Test Plan
Pass
* yarn run flow
* yarn run lint
* yarn run test


jasonLaster:
Thanks @arthur801031! I think you need to include the `yarn.lock`

Also - flow is now on 46 if that's alright :)

arthur801031:
Hi @jasonLaster !

Done. Please check. Thank you! :)

src/utils/editor/source-search.js
```diff
@@ -201,13 +201,18 @@ function searchNext(ctx, rev, query, newQuery, modifiers) {
...
+    }
+
     let cursor = getSearchCursor(cm, state.query, pos, modifiers);
 
     const location = rev
       ? { line: cm.lastLine(), ch: null }
       : { line: cm.firstLine(), ch: 0 };
 
-    if (!cursor.find(rev)) {
+    if (!cursor.find(rev) && state.query) {
```
AnshulMalik: Umm, If we have added a condition above to check state.query, then state.query will always evaluate to true at this point  isn't it ?

arthur801031: Yes. That's what I thought as well, but if we don't include this here, we'll get
screen shot 2017-05-21 at 1 33 02 pm

AnshulMalik: Oh. Okay

jasonLaster: Yep this makes sense. I bet we can better type state and not have these issues :)

jasonLaster:
Thanks @arthur801031. staying on tip Flow has been a huge benefits

PR #2967, opened 2017-05-19
Parser splitup (part 2)

jasonLaster:
### Summary of Changes

* moves tests and test fixtures into the parser directory
* creates scopes and resolvesToken modules


src/utils/parser/getSymbols.js
```diff
@@ -1,11 +1,13 @@
...
-import isEmpty from "lodash/isEmpty";
 import * as t from "babel-types";
 
 import getFunctionName from "./utils/getFunctionName";
 
+import type { SourceText } from "debugger-html";
+
 const symbolDeclarations = new Map();
 
 export type ASTLocation = {
```
codehag: this feels a bit of out of place here -- maybe it should also be moved to the utils/ast? I might also not be understanding the code

src/utils/parser/resolveToken.js
```diff
@@ -0,0 +1,29 @@
+// @flow
+
+import { getClosestExpression, getClosestScope } from "./utils/closest";
+import { isExpressionInScope } from "./scopes";
+
+import type { SourceText, Location, Frame } from "debugger-html";
+
+// Resolves a token (at location) in the source to determine if it is in scope
+// of the given frame and the expression (if any) to which it belongs
+export default function resolveToken(
```
codehag: :+1: really nice

src/utils/parser/scopes.js
```diff
@@ -0,0 +1,48 @@
...
+    references: binding.referencePaths
+  }));
+}
+
+function getScopeChain(scope: Scope): Scope[] {
+  const scopes = [scope];
+
+  do {
+    scopes.push(scope);
+  } while ((scope = scope.parent));
```
codehag: I think there might be an issue here.. though I might be missing something, because I see that its used in other bits of code. The issue is, I am not sure how this loop terminates -- from what I understand reading this, the expression in the while loop returns `undefined`, which is the same whether `scope.parent` is defined or not, and that can be done endlessly. When I tried running just this code with some fake data I did get stuck in an infinite loop.

So two things -- perhaps we can make this condition safer? and, why the heck is it working??

src/utils/parser/scopes.js
```diff
@@ -0,0 +1,48 @@
...
+
+export function getVariablesInLocalScope(scope: Scope) {
+  return getScopeVariables(scope);
+}
+
+export function getVariablesInScope(scope: Scope) {
+  const scopes = getScopeChain(scope);
+  const scopeVars = scopes.map(getScopeVariables);
+  const vars = [{ name: "this" }, { name: "arguments" }]
+    .concat(...scopeVars)
```
codehag: 😍 i do love the spread operator!

src/utils/parser/tests/closest.js
```diff
@@ -0,0 +1,92 @@
+/* eslint max-nested-callbacks: ["error", 4]*/
+
+const expect = require("expect.js");
+import { getClosestScope, getClosestExpression } from "../utils/closest";
+
+import { getSourceText } from "./helpers";
+
+describe("parser", () => {
+  describe("getClosestExpression", () => {
```
codehag: these tests are beautiful!

src/utils/parser/utils/closest.js
```diff
@@ -0,0 +1,126 @@
...
+type nodeContainsLocationParams = {
+  node: Node,
+  location: Location
+};
+
+function nodeContainsLocation({ node, location }: nodeContainsLocationParams) {
+  const { start, end } = node.loc;
+  const { line, column } = location;
+
+  if (start.line === end.line) {
```
codehag: what should the default case be if `start.line` does not equal `end.line`? `return false`?

src/utils/parser/utils/getFunctionName.js
```diff
@@ -1,4 +1,7 @@
-export default function getFunctionName(path): string {
+// @flow
+import type { NodePath } from "babel-traverse";
+
+export default function getFunctionName(path: NodePath): string {
```
codehag: :+1:

src/utils/parser/scopes.js
codehag: also one more question, do we always want to push at least one scope in? or should we not push a scope if 
```
// while ((scope = scope.parent))
let parentScope = scope.parent;
if (parentScope !== undefined) return "do nothing";

// or do we want (assume returnOnce only does it once
if (parentScope !== undefined) returnOnce(() => scopes.push(parentScope);
```

.eslintignore
```diff
@@ -2,6 +2,6 @@ assets/*
 src/test/examples/**
 src/test/integration/**
 src/test/unit-sources/**
-src/utils/tests/fixtures/**
+src/utils/parser/tests/fixtures/**
```
jasonLaster: we're ignoring the fixtures because the code is terrible lol

src/utils/parser/getSymbols.js
jasonLaster: good call we can move it to a better spot

src/utils/parser/scopes.js
jasonLaster: haha - this could *is* weird

src/utils/parser/utils/closest.js
jasonLaster: that's tough to say. the other cases are obviously really important. We probably need better docs

src/utils/parser/scopes.js
jasonLaster: so... i dropped the extra scope. don't know what to say about do/while. it terminates lol...

Issue #2592, opened 2017-04-08
[SearchBar] hide modifiers when function or variable search is on


jasonLaster:
We initially decided to disable the modifier buttons in the search bar when function or variable search was on. I've seen several people get tripped up here, so it would probably be easier to just hide them.

```diff
diff --git a/src/components/Editor/SearchBar.js b/src/components/Editor/SearchBar.js
index 33d8660..9177882 100644
--- a/src/components/Editor/SearchBar.js
+++ b/src/components/Editor/SearchBar.js
@@ -542,12 +542,15 @@ const SearchBar = React.createClass({
     } = this.props;
     const { symbolSearchEnabled } = this.state;

+    if (symbolSearchEnabled) {
+      return;
+    }
+
     function searchModBtn(modVal, className, svgName, tooltip) {
       return dom.button(
         {
           className: classnames(className, {
-            active: !symbolSearchEnabled && modifiers && modifiers.get(modVal),
-            disabled: symbolSearchEnabled
+            active: !symbolSearchEnabled && modifiers && modifiers.get(modVal)
           }),
           onClick: () =>
```

jasonLaster:
@Erica42 is on this

jasonLaster:
@Erica42 moving this back to available since it's been a month. Feel free to re-claim it if you have time

PR #2968, opened 2017-05-19
Hide search modifiers when symbol search enabled

anbarasiu:
Associated Issue: #2592 

### Summary of Changes

* Modified the disabled property to hide search modifiers, when symbol search enabled, else to show them.

Please let me know if this looks okay.


src/components/Editor/SearchBar.js
```diff
@@ -624,7 +624,7 @@ class SearchBar extends Component {
         {
           className: classnames(className, {
             active: !symbolSearchOn && modifiers && modifiers.get(modVal),
-            disabled: symbolSearchOn
+            hide: symbolSearchOn
```
jasonLaster: this is a valid approach, but we often just choose not to render a component too :)

i.e.

```js
if (! symbolSearchOn) {
  return null;
}
```

anbarasiu: That seems a better way of doing it. I've modified it to not render the component. Please take a look and let me know :)

anbarasiu:
Great, thanks!

jasonLaster:
Hey @anbarasiu, you should join our slack if you haven't already: https://devtools-html.slack.com

anbarasiu:
Thanks! I've joined :)

PR #2975, opened 2017-05-20
Support sources and outline toggle

### LINKED ISSUE #2563: [OutlineView] Navigation
We currently show both views. We should be able to navigate between the sources view and outline view. 

This issue is about adding a button that toggles the two views. We'll want to likely switch the UI in a later PR.

Here's a diff for getting started:

```diff
diff --git a/src/components/Sources.js b/src/components/Sources.js
index 161fc98..56b8e09 100644
--- a/src/components/Sources.js
+++ b/src/components/Sources.js
@@ -20,6 +20,7 @@ class Sources extends Component {

   constructor(props) {
     super(props);
+    this.state = { selectedPane: "sources" };
     this.renderShortcut = this.renderShortcut.bind(this);
   }

@@ -39,14 +40,35 @@ class Sources extends Component {
     }
   }

+  togglePane() {
+    const selectedPane = this.state.selectedPane == "sources"
+      ? "outline"
+      : "sources";
+    this.setState({ selectedPane });
+  }
+
+  renderHeader() {
+    return dom.div({ className: "sources-header" }, this.renderShortcut());
+  }
+
+  renderFooter() {
+    return dom.div(
+      { className: "sources-footer" },
+      dom.button({
+        onClick: this.togglePane
+      })
+    );
+  }
+
   render() {
     const { sources, selectSource } = this.props;

     return dom.div(
       { className: "sources-panel" },
-      dom.div({ className: "sources-header" }, this.renderShortcut()),
+      this.renderHeader(),
       SourcesTree({ sources, selectSource }),
-      Outline({ selectSource })
+      Outline({ selectSource }),
+      this.renderFooter()
     );
   }
 }
 ```

navigation

### PR DESCRIPTION
**Pull request by AnshulMalik**
Fixes #2563 

### Summary of Changes

* Add footer to left sidebar
* Button to toggle between sources and outline in left sidebar footer 

### Test Plan

- [x] Dark theme
- [x] ltr and rtl
- [x] Horizontal and vertical layout

### Screenshot

toggle

### CODE REVIEW

File: src/components/Outline.js [hunk 1]
```diff
@@ -70,18 +75,21 @@ class Outline extends Component {
...
-      { className: "outline" },
+      { className: classnames("outline", { hidden: isHidden }) },
       dom.ul({ className: "outline-list" }, this.renderFunctions())
     );
   }
 }
 
 Outline.propTypes = {
+  isHidden: PropTypes.bool,
+  sourceText: PropTypes.object,
```
codehag: :+1: nice! it looks like this might also be a required prop, since the `constructor` and `componentWillReceiveProps` both rely on it. maybe  we should add `isRequired` here?

src/components/Sources.js
```diff
@@ -15,12 +16,61 @@ const Outline = createFactory(_Outline);
...
+      : "sources";
+
+    this.setState({ selectedPane });
+  }
+
+  renderFooter() {
+    const { selectedPane } = this.state;
+    const showSourcesTooltip = L10N.getStr("sourcesPane.showSourcesTooltip");
+    const showOutlineTooltip = L10N.getStr("sourcesPane.showOutlineTooltip");
+    const tooltip = selectedPane === "sources"
```
codehag: we are evaluating if selectedPane === "sources" here and at 55 -- lets save this as a variable called "isSourcesPaneSelected". or something

src/components/Outline.js
AnshulMalik: Yep. Thought of converting to props after this PR is merged, so kept it this way.

Issue #2891, opened 2017-05-14
[Preview] Function names should be simplified 


jasonLaster:
We simplify the function names when they appear in the call stack and in preview. It would be nice to do the same thing when they appear in the tooltip.

<img src='https://shipusercontent.com/8c2e99245d5108d22dff8dd710ce4875/Screen%20Shot%202017-05-14%20at%2010.54.59%20AM.png' title='Screen Shot 2017-05-14 at 10.54.59 AM.png' width=598>

Here is a patch to help get started

```diff
diff --git a/src/components/shared/previewFunction.js b/src/components/shared/previewFunction.js
index b66bf9b..711e1e9 100644
--- a/src/components/shared/previewFunction.js
+++ b/src/components/shared/previewFunction.js
@@ -3,13 +3,14 @@ import { DOM as dom } from "react";
 import times from "lodash/times";
 import zip from "lodash/zip";
 import flatten from "lodash/flatten";
+import { simplifyDisplayName } from "../utils/frame";

 import "./previewFunction.css";

 function getFunctionName(func) {
-  return (
-    func.userDisplayName || func.displayName || func.name || func.value || ""
-  );
+  const name =
+    func.userDisplayName || func.displayName || func.name || func.value || "";
+  return simplifyDisplayName(name);
 }

 function renderFunctionName(func) {
```

and a storybook story that you can add to test the change

```diff
diff --git a/src/components/stories/Preview.js b/src/components/stories/Preview.js
index 3788ddd..ef50ebc 100644
--- a/src/components/stories/Preview.js
+++ b/src/components/stories/Preview.js
@@ -141,5 +141,29 @@ options.forEach(option => {
         },
         option
       );
+    })
+    .add(`function tooltip ${optionLabel}`, () => {
+      const func = {
+        type: "object",
+        actor: "server2.conn2.child1/pausedobj80",
+        class: "Function",
+        extensible: true,
+        frozen: false,
+        sealed: false,
+        displayName: "Backbone.Events.on",
+        parameterNames: ["name", "callback", "context"],
+        location: {
+          url: "http://localhost:7999/todomvc/bower_components/backbone/backbone.js",
+          line: 85
+        }
+      };
+      return PreviewFactory(
+        {
+          value: func,
+          expression: "this.on",
+          loadedObjects: I.Map().set(func.actor, func)
+        },
+        option
+      );
     });
 });
 ```
 
[...]

PR #2976, opened 2017-05-21
simplify function name in tooltip

karthikJagadeesh:
Associated Issue: #2891


codehag:
Looks like  Flow is complaining! We also need to update the storybook test for this. You can find info about how to do that in the ticket :) 

For flow, here is the output:
(to run flow locally, use `yarn flow`. let me know if you need help with typing!)
```
src/components/shared/previewFunction.js:23
 23:   return simplifyDisplayName(name);
                                  ^^^^ property `displayName` of unknown type. This type is incompatible with the expected param type of
124: export function simplifyDisplayName(displayName: string) {
                                                      ^^^^^^ string. See: src/utils/frame.js:124

src/components/shared/previewFunction.js:23
 23:   return simplifyDisplayName(name);
                                  ^^^^ property `userDisplayName` of unknown type. This type is incompatible with the expected param type of
124: export function simplifyDisplayName(displayName: string) {
                                                      ^^^^^^ string. See: src/utils/frame.js:124

src/components/shared/previewFunction.js:23
 23:   return simplifyDisplayName(name);
                                  ^^^^ property `value` of unknown type. This type is incompatible with the expected param type of
124: export function simplifyDisplayName(displayName: string) {
                                                      ^^^^^^ string. See: src/utils/frame.js:124
```

src/components/stories/Preview.js
```diff
@@ -131,22 +141,39 @@ function PreviewFactory(options, { dir = "ltr", theme = "light" } = {}) {
...
+      top: 200,
+      left: 200,
+      bottom: 80,
+      width: 60,
+      height: 30
+    }),
+    classList: { add: () => {}, remove: () => {} }
+  };
+
+  if (!document.querySelector("#foobar")) {
```
codehag: this looks like a leftover?

src/components/stories/Preview.js
```diff
@@ -131,22 +141,39 @@ function PreviewFactory(options, { dir = "ltr", theme = "light" } = {}) {
...
-  document.body.appendChild(div);
-  div.id = "foobar";
-  div.style.top = "100px";
-  div.style.left = "100px";
-  div.width = "100px";
-  div.height = "20px";
+
+  // NOTE: I'm hoping somebody smarter than me can find a way to add a target element 
+  // as a child of `editor-wrapper` in react and reference it in Preview or create an
+  // element that is always available and in the same location.
```
codehag: i dont understand stories all that well -- but since this is essentially display information, maybe we could just do this with a class?

jasonLaster: heh - i don't follow

Issue #2850, opened 2017-05-08
[CodeSearch] Save search modifiers as prefs


jasonLaster:
It would be really nice if modifier state were persisted across debugging sessions. We get the feedback that users often prefer one setting like case sensitivity and are frustrated by having to toggle that often.

Here's a quick sketch of what the feature could look like. We can base this off of other UI features that are also working in the UI reducer

```diff
diff --git a/assets/panel/prefs.js b/assets/panel/prefs.js
index 58469e1..552a3f0 100644
--- a/assets/panel/prefs.js
+++ b/assets/panel/prefs.js
@@ -38,3 +38,6 @@ pref("devtools.debugger.tabs", "[]");
 pref("devtools.debugger.pending-selected-location", "{}");
 pref("devtools.debugger.pending-breakpoints", "[]");
 pref("devtools.debugger.expressions", "[]");
+pref("devtools.debugger.file-search-case-sensitive", true);
+pref("devtools.debugger.file-search-whole-word", false );
+pref("devtools.debugger.file-search-regex-match", false);
diff --git a/src/reducers/ui.js b/src/reducers/ui.js
index ebe2eac..daf45ac 100644
--- a/src/reducers/ui.js
+++ b/src/reducers/ui.js
@@ -37,9 +37,9 @@ export const State = makeRecord(
     fileSearchOn: false,
     fileSearchQuery: "",
     fileSearchModifiers: makeRecord({
-      caseSensitive: true,
-      wholeWord: false,
-      regexMatch: false
+      caseSensitive: prefs.fileSearchCaseSensitive,
+      wholeWord: prefs.fileSearchWholeWord,
+      regexMatch: prefs.fileSearchRegexMatch
     })(),
     projectSearchOn: false,
     symbolSearchOn: false,
diff --git a/src/utils/prefs.js b/src/utils/prefs.js
index 3025a0b..197acae 100644
--- a/src/utils/prefs.js
+++ b/src/utils/prefs.js
@@ -16,12 +16,20 @@ if (isDevelopment()) {
   pref("devtools.debugger.pending-selected-location", "{}");
   pref("devtools.debugger.pending-breakpoints", "[]");
   pref("devtools.debugger.expressions", "[]");
+  pref("devtools.debugger.fileSearchCaseSensitive", true);
+  pref("devtools.debugger.fileSearchWholeWord", false );
+  pref("devtools.debugger.fileSearchRegexMatch", false);
 }

+
+
[...]

PR #2977, opened 2017-05-21
Save Search Modifiers as Prefs

sharathnarayanph:
Associated Issue: #2850 

Summary of Changes

Preferences added for File Search options:
- Case-sensitive
- Regex
- Whole word

Test Plan

- Open the application and toggle the search options
- Go back/close application
- Open search options and the previously set values should be retained

### DIFF
--- a/src/reducers/ui.js
+++ b/src/reducers/ui.js
@@ -99,10 +99,7 @@ function update(
         prefs.fileSearchRegexMatch = actionVal;
       }
 
-      return state.setIn(
-        ["fileSearchModifiers", action.modifier],
-        !state.getIn(["fileSearchModifiers", action.modifier])
-      );
+      return state.setIn(["fileSearchModifiers", action.modifier], actionVal);
     }
 
     case constants.SET_SYMBOL_SEARCH_TYPE: {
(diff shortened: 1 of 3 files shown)


src/reducers/ui.js
```diff
@@ -85,6 +85,20 @@ function update(
     }
 
     case constants.TOGGLE_FILE_SEARCH_MODIFIER: {
+      const actionVal = !state.getIn(["fileSearchModifiers", action.modifier]);
+
+      if (action.modifier == "caseSensitive") {
+        prefs.fileSearchCaseSensitive = actionVal;
```
codehag: I am not very familiar with the issue, but should we also set the state in these cases?

src/reducers/ui.js
```diff
@@ -85,6 +85,20 @@ function update(
...
+        prefs.fileSearchWholeWord = actionVal;
+      }
+
+      if (action.modifier == "regexMatch") {
+        prefs.fileSearchRegexMatch = actionVal;
+      }
+
       return state.setIn(
         ["fileSearchModifiers", action.modifier],
         !state.getIn(["fileSearchModifiers", action.modifier])
```
jasonLaster: hmm - maybe we should use `actionVal` here to be consistent.

sharathnarayanph: Oops! I should have done it. My bad!

Issue #2983, opened 2017-05-22
[Search] Highlight match yellow


jasonLaster:
The search matched label is an off grey, which would be much nicer as a yellow background.

Here is where the highlight CSS comes from: https://github.com/devtools-html/devtools-core/blob/master/packages/devtools-source-editor/src/source-editor.css#L120-L161

Here is where we find the matches and add the highlight: https://github.com/sharathnarayanph/debugger.html/blob/SaveSearch/src/utils/editor/source-search.js#L66-L108

<img src='https://shipusercontent.com/54b9b4e41272996d29811c6e450a6588/Screen%20Shot%202017-05-22%20at%2012.19.02%20PM.png' title='Screen Shot 2017-05-22 at 12.19.02 PM.png' width=487>
<img src='https://shipusercontent.com/62c44e35116fa21fa0b04b8fca878a9e/Screen%20Shot%202017-05-22%20at%2012.18.54%20PM.png' title='Screen Shot 2017-05-22 at 12.18.54 PM.png' width=488>

PR #2985, opened 2017-05-22
Add search highlighting

jasonLaster:
Fixes: #2983

Associated PR: https://github.com/devtools-html/devtools-core/pull/377 

This moves the highlight CSS back into the launchpad, which also means that the highlighting gets the yellow background.


src/components/Editor/index.js
```diff
@@ -397,7 +400,11 @@ class Editor extends PureComponent {
       selectedFrame
     );
 
+    console.log("inScope", inScope);
+    console.log("expression", expression);
+
     if (!inScope) {
+      console.groupEnd();
```
jasonLaster: oops - the console's should be gone

PR #2990, opened 2017-05-23
When the world stops

jasonLaster:
quick blog post on the debugger server

### DIFF
--- a/docs/updates/README.md
+++ b/docs/updates/README.md
@@ -88,6 +88,9 @@ Here are some highlights from the week:
 * :police_car: License checking for our dependencies
 * :nail_care: Lots of UI polish
 
+### [When the World Stops](./when-the-world-stops.md)
+
+One of the most interesting debugger questions is what happens when the debugger stops? This post is a quick run-through that will give you some context so that you can dig in and answer your own questions.
 
 ### [May 23rd](./updates-5-23-2017.md)
 
--- a/docs/updates/when-the-world-stops.md
+++ b/docs/updates/when-the-world-stops.md
@@ -1,0 +1,214 @@
+### When the world stops
+
+
+One of the most interesting questions is what happens when the debugger stops? Basically, what happens when the world pauses?
+
+Here is a quick run-through that will help give you some context so that you can dig in and answer your own questions.
+
+There are three characters in the story: the client, server, and engine.
+When the page pauses, the engine pauses the page, tells the server,
+and the server tells the client.
+
+This post is going to focus on the work the server does when the page pauses. The code included has been simplified a bit to remove some of the incidental complexity.
+
+---
+
+The [server][script.js] has several page lifecycle hooks that are called when
+something happens. `onDebuggerStatement` is one of these hooks!
+
+```js
+onDebuggerStatement: function (frame) {
+  return this._pauseAndRespond(frame, { type: "debuggerStatement" });
+}
+```
+
+The [server][script.js]'s main job is to tell the client that the page is paused.
+You can see below, that the `_paused` function is building a `packet`, this packet
+has all of the data the client needs to inform the programmer about the pause state.
+The `_paused` function also does one other interesting thing here as well. When the function is triggered, it registers all of the relevant actors in the story and adds them to various "actor pools".
+The server registers all of the actors so that if the client has any follow up questions,
+the server will know who to follow up with.
+
+The server is like a detective, it wants to know who all the major witnesses are!
+
+
+```js
+_paused: function (frame) {
+  this._state = "paused";
+
+
+  this._pausePool = new ActorPool(this.conn);
+  this.conn.addActorPool(this._pausePool);
+
+  // Give children of the pause pool a quick link back to the
+  // thread...
+  this._pausePool.threadActor = this;
+  this._pauseActor = new PauseActor(this._pausePool);
+  this._pausePool.addActor(this._pauseActor);
+
+  // Update the list of frames.
+  let poppedFrames = this._updateFrames();
+
+  // Send off the paused packet and spin an event loop.
+  let packet = { from: this.actorID,
+                 type: "paused",
+                 actor: this._pauseActor.actorID };
+  if (frame) {
+    packet.frame = this._createFrameActor(frame).form();
+  }
+
+  if (poppedFrames) {
+    packet.poppedFrames = poppedFrames;
+  }
+
+  return packet;
+}
+```
+
+Lets continue digging into how the server builds the pause packet for the client.
+We already know quite a bit already. We know the `type` is "paused", it has a `from` and an `actor` field which is used as a return address :). We also know that there are two types of frame fields: `frame` and `poppedFrames`. Lets focus on the `frame` field because it is simpler.
+
+We get the `frame` packet data by asking for the frame actor's `form`.
+The form function is defined in the Frame actor [class][frame.js].
+Lets slow down and take a look at the data the frame actor puts in its form.
+It's fascinating!
+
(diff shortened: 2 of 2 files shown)


docs/updates/README.md
```diff
@@ -88,6 +88,9 @@ Here are some highlights from the week:
 * :police_car: License checking for our dependencies
 * :nail_care: Lots of UI polish
 
+### [When the World Stops](./when-the-world-stops.md)
+
+One of the most interesting debugger questions is what happens when the debugger stops? This post is a quick run-through that will give you some context so that you can dig in and answer your own questions.
```
nchevobbe: s/questions/question I think ?

AnshulMalik:
This is awesome :)

docs/updates/when-the-world-stops.md
```diff
@@ -0,0 +1,216 @@
+### When the world stops
+
+
+One of the most interesting questions is what happens when the debugger stops? Basically, what happens when the world pauses?
```
nchevobbe: s/questions/question

docs/updates/when-the-world-stops.md
```diff
@@ -0,0 +1,216 @@
...
+  }
+
+  return form;
+}
+```
+
+Alright, we've arrived at a frame environment and we're asking it for its form.
+Environment is just a fancy word for a scope, so all that's happening is the frame is saying, "yo scope, whatcha got for me"? Our main goal here is to see where the variables are documented.
+So lets walk through what it's doing. First the environment notes its type. Then it does some sneaky recursion to inquire its parent scope's data. Lets not get nerd sniped! Then it gets object and function data, remember we could be in a function or object scope, right?!?!
+Then we get bindings... Bindings are exactly what we want here. A good way to think of bindings are: variables are bound to scopes in JS, ergo varialbles are in the bindings thing! Lets go a bit deeper!
```
nchevobbe: s/varialbles/variables 
Does `ergo` is what e.g. stands for ?

docs/updates/when-the-world-stops.md
```diff
@@ -0,0 +1,216 @@
...
+
+Lets continue digging into how the server builds the pause packet for the client.
+We already know quite a bit already. We know the `type` is "paused" and it has a `from` and `actor` field as a return address :). We also know that there are two types of frame fields: `frame` and `poppedFrames`. Lets focus on the `frame` field because it is simpler.
+
+We get the `frame` packet data by asking for the frame actor's `form`.
+The form function is defined in the Frame actor [class][frame.js].
+Lets slow down and take a look at the data the frame actor puts in its form.
+It's fascinating!
+
+It starts with the `actor` and `type` field for record keeping. It then sets up `callee`, `environment`, `arguments`, `where`, and `oldest` fields. Oldest, what's that? :) It's neat to think of a frame as potentially a function frame, at which point we care about the callee and arguments data.
```
nchevobbe: I don't feel like we're explaining what `oldest` is here, or at least I don't understand what it does.

docs/updates/when-the-world-stops.md
nchevobbe: s/Lets not get nerd sniped/Let's not get nerd sniped

docs/updates/when-the-world-stops.md
```diff
@@ -0,0 +1,216 @@
...
+  for (let name of this.obj.names()) {
+    let value = this.obj.getVariable(name);
+
+    let desc = {
+      value: value,
+      configurable: false,
+      enumerable: true
+    };
+
+    // let desc = this.obj.getVariableDescriptor(name);
```
nchevobbe: Should this line be removed ?

docs/updates/when-the-world-stops.md
```diff
@@ -0,0 +1,216 @@
...
+onDebuggerStatement: function (frame) {
+  return this._pauseAndRespond(frame, { type: "debuggerStatement" });
+}
+```
+
+The [server][script.js]'s main job is to tell the client that the page is paused.
+You can see below, that the `_paused` function is building a `packet`, this packet
+has all of the data the client needs to inform the programmer about the pause state.
+The `_paused` function also does one other interesting thing here as well. It starts,
+registering all of the relevant actors in the story in various "actor pools".
```
bomsy: ... story ~~in~~ into various "actor pools"

docs/updates/when-the-world-stops.md
```diff
@@ -0,0 +1,216 @@
...
+  if (poppedFrames) {
+    packet.poppedFrames = poppedFrames;
+  }
+
+  return packet;
+}
+```
+
+Lets continue digging into how the server builds the pause packet for the client.
+We already know quite a bit already. We know the `type` is "paused" and it has a `from` and `actor` field as a return address :). We also know that there are two types of frame fields: `frame` and `poppedFrames`. Lets focus on the `frame` field because it is simpler.
```
bomsy: ... We know the `type` is "paused", it has a `from` field and an `actor` field which is used as a return address :) ...

docs/updates/when-the-world-stops.md
```diff
@@ -0,0 +1,216 @@
...
+
+We get the `frame` packet data by asking for the frame actor's `form`.
+The form function is defined in the Frame actor [class][frame.js].
+Lets slow down and take a look at the data the frame actor puts in its form.
+It's fascinating!
+
+It starts with the `actor` and `type` field for record keeping. It then sets up `callee`, `environment`, `arguments`, `where`, and `oldest` fields. Oldest, what's that? :) It's neat to think of a frame as potentially a function frame, at which point we care about the callee and arguments data.
+That's cool, but where's the good stuff? Where's are the variables and scope data kept? All of that is kept in an environment actor and fetched through a similar `form` function. Lets keep on digging!
```
bomsy: ...Where ~~Where's~~ are the variables and scope ...

docs/updates/when-the-world-stops.md
bomsy: ... to ~~inquire~~ inquire about its parent scope's data ...

docs/updates/when-the-world-stops.md
```diff
@@ -0,0 +1,216 @@
...
+```js
+onDebuggerStatement: function (frame) {
+  return this._pauseAndRespond(frame, { type: "debuggerStatement" });
+}
+```
+
+The [server][script.js]'s main job is to tell the client that the page is paused.
+You can see below, that the `_paused` function is building a `packet`, this packet
+has all of the data the client needs to inform the programmer about the pause state.
+The `_paused` function also does one other interesting thing here as well. It starts,
```
codehag: `It starts,` could be read with two different tones... maybe `When the function is triggered, it registers ...` -- this makes `it` clear, and doesn't have the weird pause.

docs/updates/when-the-world-stops.md
```diff
@@ -0,0 +1,216 @@
...
+registering all of the relevant actors in the story in various "actor pools".
+The server registers all of the actors so that if the client has any follow up questions,
```
codehag: is this part of the pause function?

docs/updates/when-the-world-stops.md
jasonLaster: hmm, i like questions... am i missing something

docs/updates/when-the-world-stops.md
jasonLaster: yeah - i don't know what it does either :)

docs/updates/when-the-world-stops.md
jasonLaster: yeah...

PR #2994, opened 2017-05-23
Bug/2817 breakpoints nightly won't delete breakpoints

codehag:
Associated Issue: #2817 

### Summary of Changes

* updated pending breakpoint tests to use jest instead of expectjs
* made initialState for breakpoints be the initialState for the update function in the breakpoints reducer, rather than using a stale copy from when the file was initialized
* wrote tests for the bug
* fixed bug

### Test Plan

This feature is unit / integration tested

Example test plan:

- [x] open panel
- [x] set a break point
- [x] disable breakpoint
- [x] reload panel

set breakpoint:

screen shot 2017-05-23 at 6 34 13 pm

disable breakpoint:

screen shot 2017-05-23 at 6 34 35 pm

reload debugger:

screen shot 2017-05-23 at 6 34 26 pm

### DIFF
--- a/src/actions/breakpoints.js
+++ b/src/actions/breakpoints.js
@@ -64,11 +64,12 @@ async function addClientBreakpoint(state, client, sourceMaps, breakpoint) {
 }
 
 /**
- * Enabling a breakpoint calls {@link addBreakpoint}
+ * Enabling a breakpoint calls
  * which will reuse the existing breakpoint information that is stored.
  *
  * @memberof actions/breakpoints
  * @static
+ * @param {Location} $1.location Location  value
  */
 export function enableBreakpoint(location: Location) {
   return ({ dispatch, getState, client, sourceMaps }: ThunkArgs) => {
@@ -95,11 +96,7 @@ export function enableBreakpoint(location: Location) {
  */
 export function addBreakpoint(
   location: Location,
-  {
-    condition = null,
-    disabled = false,
-    getTextForLine
-  }: addBreakpointOptions = {}
+  { condition, disabled = false, getTextForLine }: addBreakpointOptions = {}
 ) {
   return ({ dispatch, getState, client, sourceMaps }: ThunkArgs) => {
     if (_breakpointExists(getState(), location)) {
@@ -125,7 +122,26 @@ export function addBreakpoint(
  * @static
  */
 export function disableBreakpoint(location: Location) {
-  return _removeOrDisableBreakpoint(location, true);
+  return ({ dispatch, getState, client }: ThunkArgs) => {
+    let bp = getBreakpoint(getState(), location);
+    if (!bp) {
+      throw new Error("attempt to disable a breakpoint that does not exist");
+    }
+    if (bp.loading) {
+      // TODO(jwl): make this wait until the breakpoint is saved if it
+      // is still loading
+      throw new Error("attempt to disable unsaved breakpoint");
+    }
+
+    const action = {
+      type: "DISABLE_BREAKPOINT",
+      breakpoint: bp,
+      disabled: true,
+      [PROMISE]: client.removeBreakpoint(bp.id)
+    };
+
+    return dispatch(action);
+  };
 }
 
 /**
@@ -135,10 +151,6 @@ export function disableBreakpoint(location: Location) {
  * @static
  */
 export function removeBreakpoint(location: Location) {
-  return _removeOrDisableBreakpoint(location);
-}
-
-function _removeOrDisableBreakpoint(location, isDisabled = false) {
   return ({ dispatch, getState, client }: ThunkArgs) => {
     let bp = getBreakpoint(getState(), location);
     if (!bp) {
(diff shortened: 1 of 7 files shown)


src/actions/tests/pending-breakpoints.js
```diff
@@ -31,19 +23,29 @@ import {
...
+  makePendingLocationId,
+  makeLocationId
+} from "../../reducers/breakpoints";
 
 describe("when adding breakpoints", () => {
+  const mockedPendingBreakpoint = mockPendingBreakpoint();
+
+  beforeEach(() => {
+    const id = makePendingLocationId(mockedPendingBreakpoint.location);
+    prefs.pendingBreakpoints = { [id]: mockedPendingBreakpoint };
```
jasonLaster: 👍

codehag:
So, I have a solution but it is imperfect:

I am not setting the breakpoint's initial state in the addBreakpoint reducer. Instead, (as you can see here: https://github.com/codehag/debugger.html/blob/b67525b5be007f077253f471e97ea030564a71d6/src/actions/sources.js#L66) after the breakpoint is added, I check if it should be disabled, and fire the action to disable it.

This solution is imperfect because the breakpoint blinks on, and then off if it is loaded.

I think that a full solution would require refactoring the `addBreakpoint` reducer so that enabling a break point does not rely on it. I've narrowed down the new bug of skipping breakpoints being a result of that. I suspect that the call to the server returns a slid location because the server expects a different breakpoint, and adding a breakpoint to a disabled breakpoint doesn't make sense to it. maybe. 

I decided that since it took so much time to get to this point, to stop here and get some feedback. Let me know what you think @jasonLaster (and anyone else interested)

src/actions/sources.js
```diff
@@ -13,8 +13,8 @@ import defer from "../utils/defer";
 import { PROMISE } from "../utils/redux/middleware/promise";
 import assert from "../utils/assert";
 import { updateFrameLocations } from "../utils/pause";
-import { addBreakpoint } from "./breakpoints";
-import { isEnabled } from "devtools-config";
+import { addBreakpoint, disableBreakpoint } from "./breakpoints";
+// import { isEnabled } from "devtools-config";
```
wldcordeiro: Some leftover comments here and on [line 69?](https://github.com/devtools-html/debugger.html/pull/2994/files#diff-695f8e845eb7e7930a1b665650866b59R69)

src/actions/breakpoints.js
```diff
@@ -15,15 +15,53 @@ import { getBreakpoint, getBreakpoints, getSource } from "../selectors";
...
+  }
+
+  const source = getSource(state, breakpoint.location.sourceId);
+  return await sourceMaps.getGeneratedLocation(
+    breakpoint.location,
+    source.toJS()
+  );
+}
+
+async function addClientBreakpoint(state, client, sourceMaps, breakpoint) {
```
codehag: move to `client/utils`

jasonLaster: lets move this to `client/shared/addCleintBreakpoint.js`

src/actions/breakpoints.js
```diff
@@ -15,15 +15,53 @@ import { getBreakpoint, getBreakpoints, getSource } from "../selectors";
 import type { ThunkArgs } from "./types";
 import type { Location } from "../types";
 
-type addBreakpointOptions = { condition: string, getTextForLine?: () => any };
+type addBreakpointOptions = {
+  condition: string,
+  disabled?: boolean,
+  getTextForLine?: () => any
```
jasonLaster: lets kill `getTextForLine`

src/actions/breakpoints.js
```diff
@@ -15,30 +15,80 @@ import { getBreakpoint, getBreakpoints, getSource } from "../selectors";
...
   return currentBp && !currentBp.disabled;
 }
 
-function _getOrCreateBreakpoint(state, location, condition) {
-  return getBreakpoint(state, location) || { location, condition, text: "" };
+function _createBreakpoint(location: Location, opts: Object = {}) {
+  return Object.assign({}, { location }, opts);
+}
+
+async function _getSourceLocation(state, sourceMaps, breakpoint, location) {
```
jasonLaster: Hmm, what if we changed this to be `_getGeneratedLocation(sourceMaps, breakpoint, source)`

* we change `_getSourceLocation` to `_getGeneratedLocation` to be more clear
* we pass in source instead of state
* we get location from `bp` because passing in location separately is confusing

jasonLaster:
lets also update this test: https://gist.github.com/7d939ba52564890bc6533fa5e55a409d

which you can run here: http://localhost:8000/integration/

src/reducers/breakpoints.js
```diff
@@ -24,13 +24,15 @@ export type BreakpointsState = {
...
 };
 
-export const State = makeRecord(
-  ({
-    breakpoints: I.Map(),
-    pendingBreakpoints: restorePendingBreakpoints(),
-    breakpointsDisabled: false
-  }: BreakpointsState)
-);
+export function initialState() {
```
jasonLaster: can we clean this up:

```js
function initialState(): BreakpointsState {
}

```

src/reducers/breakpoints.js
```diff
@@ -171,17 +203,15 @@ function deleteBreakpoint(state, id, pendingId) {
     .deleteIn(["pendingBreakpoints", pendingId]);
 }
 
-function removeOrDisableBreakpoint(state, action) {
+function removeBreakpoint(state, action) {
```
bomsy: nice!

src/reducers/breakpoints.js
codehag: Hmm I tried this, but it introduced a few problems. do you mean that the makeRecord should be moved out?

Issue #2715, opened 2017-04-22
[Tabs] We show closed tabs when we reload


bomsy:
After the debugger is refreshed, 
* Close the active tab

AR:  The source text for the next selected tab does not get loaded
ER: The source text for the next selected tab should get loaded.

![no-sourcetext](https://cloud.githubusercontent.com/assets/792924/25308787/77952b76-27b4-11e7-8b7a-eb851cc1dcc9.gif)

jasonLaster:
@bomsy ooh, this might be fixed by your tabs work :)

jasonLaster:
Hmm - this looks like it's still an issue. Kinda interesting though

bomsy:
Oh ... yeah . It's a separate issue. Was not fixed by my tabs PR.
I was going to look at it next, but forgot.
I can look at it.

jasonLaster:
This should fix this bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1366754

bomsy:
These should all be fixed!

PR #2999, opened 2017-05-24
make sure we deselect sources

bomsy:
Associated Issue: #2715

WIP 
Todo - Fix issue causing failing test

### Summary of Changes

* Call the `selectSource` action on `closeTab/s` to select the next available tab
* If no source is found dispatch an action to trigger clearing of the currently loaded source
* export `getNewSelectedSourceId` 

### Test Plan
- [x] Select multiple sources from the source tree
- [x] After all the tab have been opened, refresh the debugger
- [x] After debugger reloads, close the selected tab
- [x] The source of the next selected tab should be loaded
- [x] Close all the tabs
- [x] The loaded source text should be removed

### DIFF
--- a/src/actions/sources.js
+++ b/src/actions/sources.js
@@ -32,8 +32,10 @@ import {
   getPendingSelectedLocation,
   getPendingBreakpoints,
   getFrames,
+  getSourceTabs,
   getNewSelectedSourceId,
-  getSourceTabs
+  removeSourcesFromTabList,
+  removeSourceFromTabList
 } from "../selectors";
 
 import type { Source, SourceText } from "../types";
@@ -190,7 +192,7 @@ export function selectSource(id: string, options: SelectSourceOptions = {}) {
 
     if (!source) {
       // If there is no source we deselect the current selected source
-      clearSelectedSource();
+      return dispatch({ type: "CLEAR_SELECTED_SOURCE" });
     }
 
     source = source.toJS();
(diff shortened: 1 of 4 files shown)


jasonLaster:
I think you found a really nice solution.

src/actions/sources.js
```diff
@@ -186,7 +188,11 @@ export function selectSource(id: string, options: SelectSourceOptions = {}) {
     let source = getSource(getState(), id);
 
     if (!source) {
-      return;
+      // If there is no source we deselect the current selected source
+      return dispatch({
+        type: constants.SELECT_SOURCE,
+        source: { id, url: null }
+      });
```
jasonLaster: Hmm, I think we should probably have a new action `clearSelectedSource`... Also, is the ID used later?

src/reducers/sources.js
```diff
@@ -75,13 +75,15 @@ function update(
     }
 
     case "SELECT_SOURCE":
+      const sourceUrl = action.source.url || "";
```
jasonLaster: i think this would be simpler if we had `clearSelectedSource`

src/reducers/sources.js
```diff
@@ -104,16 +106,12 @@ function update(
     case "CLOSE_TAB":
       availableTabs = removeSourceFromTabList(state.tabs, action.url);
 
-      return state.merge({ tabs: availableTabs }).set("selectedLocation", {
-        sourceId: getNewSelectedSourceId(state, availableTabs)
-      });
+      return state.merge({ tabs: availableTabs });
```
jasonLaster: this is a great cleanup

src/actions/sources.js
bomsy: Yeah, i totally agree! tot about this too, but didn't want to rock the boat yet! :)  having more focused actions that do specific things would allow for flexibility and less isolated scenario bugs!

PR #3011, opened 2017-05-25
(WIP) Switch shared/Popover from PropTypes to flow props

peterschussheim:
Associated Issue: #2804 

### Summary of Changes
convert from PropTypes to props using [codemod-proptypes-to-flow](https://github.com/billyvg/codemod-proptypes-to-flow)

### Test Plan
 - [x] yarn test-all
 - [x] hover over variables works
 - [ ] yarn flow: getting many Flow errors (shown below in screenshot), I would be most grateful for a bit of guidance on how to resolve them :)

### Screenshots
iterm2002
iterm2003

### DIFF
--- a/src/components/shared/Popover.js
+++ b/src/components/shared/Popover.js
@@ -7,7 +7,7 @@ const BracketArrow = createFactory(_BracketArrow);
 
 import "./Popover.css";
 
-export type Props = {
+type Props = {
   target?: Object,
   children?: Object,
   onMouseLeave?: Function,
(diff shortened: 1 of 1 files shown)


src/components/shared/Popover.js
```diff
@@ -1,12 +1,25 @@
...
 const BracketArrow = createFactory(_BracketArrow);
 
 import "./Popover.css";
 
+export type Props = {
+  target?: Object,
+  children?: Object,
+  onMouseLeave?: Function,
+  type?: string
+};
```
jasonLaster: why export them?

peterschussheim: I think the codemod added that; I just removed `export` and updated this PR.

What about all the flow errors?  Is that going down a rabbit hole trying to resolve them all?

jasonLaster: i say go for it. It's really helpful and we should be typing all the files

PR #3015, opened 2017-05-26
Create an AST redux backend

jasonLaster:
### Summary of Changes

This moves the getSymbols logic to the redux backend. It is a pre-requisite for the two preview tickets, which have complicated logic.

Next steps:

* satisfy flow
* cleanup outline
* look into selectors


src/actions/tests/ast.js
```diff
@@ -0,0 +1,33 @@
...
+  "base.js": "function base(boo) {}"
+};
+
+describe("ast", () => {
+  it("should be able to set symbols for a source", async () => {
+    const { dispatch, getState } = createStore(threadClient);
+    await dispatch(actions.newSource(makeSource("base.js")));
+    await dispatch(actions.loadSourceText({ id: "base.js" }));
+    const baseSymbols = getSymbols(getState(), "base.js");
+    expect(baseSymbols).toMatchSnapshot();
```
jasonLaster: the snapshot is great because we don't care what the symbols are. jus that they're sensible

src/reducers/ast.js
```diff
@@ -0,0 +1,59 @@
...
+  });
+
+function update(
+  state: Record<ASTState> = State(),
+  action: Action
+): Record<ASTState> {
+  switch (action.type) {
+    case "SET_SYMBOLS": {
+      const { source, symbols } = action;
+      return state.setIn(["symbols", source.id], symbols);
```
jasonLaster: @bomsy I like this solution because the reducer is an I.Map, the symbol field is an I.Map, but the symbols are just objs... maybe it's not a big deal bc people could still mess with them, but it feels better lol

src/actions/ast.js
```diff
@@ -0,0 +1,19 @@
+import { getSourceText, hasSymbols } from "../selectors";
+import parser from "../utils/parser";
+
+export function setSymbols(source) {
+  return async ({ dispatch, getState }: ThunkArgs) => {
+    if (hasSymbols(getState(), source)) {
+      return;
+    }
+
+    const sourceText = getSourceText(getState(), source.id);
```
jasonLaster: I should bale if we don't have the source text

src/actions/tests/ast.js
wldcordeiro: 👍 I like this, our test coverage has really been growing in great ways.

src/reducers/ast.js
wldcordeiro: I really dig this choice. Very clear reducer logic.

src/selectors.js
```diff
@@ -1,11 +1,12 @@
 // @flow
-const expressions = require("./reducers/expressions");
-const sources = require("./reducers/sources");
-const pause = require("./reducers/pause");
-const breakpoints = require("./reducers/breakpoints");
-const eventListeners = require("./reducers/event-listeners");
-const ui = require("./reducers/ui");
-const coverage = require("./reducers/coverage");
+import * as expressions from "./reducers/expressions";
```
wldcordeiro: Woot! More ES6 import/exports!

bomsy:
looks gud!

src/components/tests/__snapshots__/Outline.js.snap
```diff
@@ -13,15 +13,113 @@ ShallowWrapper {
 >
     <ul
         className="outline-list"
-    />
+    >
+        <li
```
jasonLaster: i think our tests were testing the wrong things :)

src/components/Outline.js
```diff
@@ -21,26 +20,6 @@ class Outline extends Component {
...
-  // TODO: move this logic out of the component and into a reducer
-  async setSymbolDeclarations(sourceText: Record<SourceText>) {
-    const symbolDeclarations = await getSymbols(sourceText.toJS());
-
-    if (symbolDeclarations !== this.state.symbolDeclarations) {
-      this.setState({
-        symbolDeclarations
-      });
-    }
   }
```
bomsy: nice reduction.

Issue #3019, opened 2017-05-26
[Preview] Disable member expressions with a call expression


jasonLaster:
When the user mouses over bazz in this expression `obj2.doEvil().c.bazzz` we should return the full expression, but add a field like `hasCallExpression` so that preview knows not to execute it.

Currently, we return `.c.bazzz`

PR #3020, opened 2017-05-26
Call expressions

jasonLaster:
Associated Issue: #3019

### Summary of Changes

I started working on the call expressions check and got detoured by jest's snapshots. I think this style is easier to test / read as we capture all of the data and can easily asses for sanity when changes are made. 

This would have helped me a lot before when i was fidgeting with expected outputs all the time


src/utils/parser/tests/closest.js
```diff
@@ -6,20 +6,45 @@ import { getSourceText } from "./helpers";
...
+        const expression = getClosestExpression(
+          getSourceText("expression"),
+          "bazzz",
+          {
+            line: 6,
+            column: 32
+          }
+        );
+
+        console.log(expression);
```
bomsy: rouge `console.log`

src/utils/parser/tests/__snapshots__/closest.js.snap
```diff
@@ -0,0 +1,121 @@
...
+FunctionExpression (5:8,5:21)
+  async: false
+  body: BlockStatement (5:19,5:21)
+    body: []
+    directives: []
+  expression: false
+  generator: false
+  id: null
+  params: []
+`;
```
bomsy: i like this!!! :)

src/utils/parser/tests/closest.js
```diff
@@ -6,20 +6,45 @@ import { getSourceText } from "./helpers";
...
+            line: 15,
+            column: 31
+          }
+        );
 
-      expect(expression.value).toBe("obj.x");
-      expect(expression.location.start).toEqual({
-        line: 15,
-        column: 26
+        expect(expression).toMatchSnapshot();
```
codehag: :+1:

src/utils/parser/tests/fixtures/expression.js
```diff
@@ -1,6 +1,7 @@
 function expr() {
   const obj = { a: { b: 2 } };
   const obj2 = { c: { b: 3 } };
-  const foo = obj2.c.b;
+  const foo = obj2.c.bazzz;
```
codehag: hmm maybe this could have a better naming, so if someone goes hunting here it will be clearer what is happening with this expression?

PR #3021, opened 2017-05-27
Switch PaneToggle from PropTypes to flow props

irfanhudda:
Associated Issue: #2804

### Summary of Changes

* Switch `PaneToggle` from PropTypes to flow props

### DIFF
--- a/src/components/shared/Button/PaneToggle.js
+++ b/src/components/shared/Button/PaneToggle.js
@@ -1,19 +1,21 @@
 // @flow
 
-import { DOM as dom, PropTypes, Component } from "react";
+import { DOM as dom, Component } from "react";
 import classnames from "classnames";
 import Svg from "../Svg";
 import "./PaneToggle.css";
 
-type NextProps = {
+type Props = {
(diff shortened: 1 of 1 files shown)


src/components/shared/Button/PaneToggle.js
```diff
@@ -1,19 +1,21 @@
...
-  handleClick: () => any,
+  handleClick: (string, boolean) => any,
   horizontal?: boolean,
   position: string
 };
 
 class PaneToggleButton extends Component {
-  shouldComponentUpdate(nextProps: NextProps) {
+  props: Props;
+
```
jasonLaster: Could the props be here?

irfanhudda: Yes we could have props here but we also need to use it in below line

```js
shouldComponentUpdate(nextProps: Props)  {
```
I used the `type Props` to avoid duplication

jasonLaster: Got it. You're one step ahead of me.

Issue #2984, opened 2017-05-22
[Preview] Highlight expression yellow


jasonLaster:
We currently have a very soft grey match highlight. We should make it pop with either a yellow background like chrome or some other strategy like underlying the token.

#### Screenshot

<img src='https://shipusercontent.com/f431b6f4299400ff4693738140c4cba6/Screen%20Shot%202017-05-22%20at%2012.44.23%20PM.png' title='Screen Shot 2017-05-22 at 12.44.23 PM.png' width=472>

#### Chrome

<img src='https://shipusercontent.com/45a30d67eadd4c94366bd37681f0910a/Screen%20Shot%202017-05-22%20at%2012.45.26%20PM.png' title='Screen Shot 2017-05-22 at 12.45.26 PM.png' width=489>

ofiron01:
@jasonLaster, can I take this as my first PR?

jasonLaster:
Sure!

PR #3022, opened 2017-05-28
[Preview] Highlight expression yellow #2984

ofiron01:
Associated Issue: #2984 

### Summary of Changes

* changed highlight expression color to more contrast 

image

### Test Plan
N/A

### DIFF
--- a/src/components/Editor/Preview.css
+++ b/src/components/Editor/Preview.css
@@ -28,7 +28,7 @@
 }
 
 .selected-token {
-  background-color: rgb(255, 255, 180);
+  background-color: var(--theme-highlight-yellow);
   color: var(--theme-selection-color);
 }
 
--- a/src/components/Editor/Preview.css
+++ b/src/components/Editor/Preview.css
@@ -28,7 +28,7 @@
 }
 
 .selected-token {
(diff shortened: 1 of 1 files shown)


src/components/Editor/Preview.css
```diff
@@ -28,7 +28,7 @@
 }
 
 .selected-token {
-  background-color: var(--theme-search-overlays-semitransparent);
+  background-color: rgb(255, 255, 180);
   color: var(--theme-selection-color);
```
bomsy: Nice! It would be cool, if this could be put as part of the theme.

ofiron01: thanks @bomsy can you point me in the right direction?

bomsy: @ofiron01 You could look in here to see if we have a variable matching the colour we want, and use that.  https://github.com/devtools-html/devtools-core/blob/master/packages/devtools-launchpad/src/lib/themes/variables.css#L1

If not, we can do a PR in devtools-core to add a new variable.

bomsy:
@ofiron01 Nice work! Looking good!

PR #3027, opened 2017-05-28
Add application state types

jasonLaster:
### Summary of Changes

Gets the application state typed. Really excited about this change. Probably needs more massaging to get in though...


codehag:
Looks great! is there more information about this?

jasonLaster:
I wrote about it in the cov-report post which hasn't been published yet, but the basic idea is that it would be great to start typing the application state. 

I did not expect flow to pass, but if it is I'm happy to merge this and then continue to add coverage

Issue #2241, opened 2017-03-01
[docs] Add a roadmap doc


jasonLaster:
It's helpful for the community to be able to learn what our roadmap is:

1. what we're working on
2. what we plan to work on over the next 3 - 6 months, year...

This helps in a couple ways:

1. they can weigh in on it
2. they can put feature requests in context
3. they can see if we plan on working on that thing they want... (watch expressions)

clarkbw:
Created a doc, will start working on it soon.
https://docs.google.com/document/d/1g-KOVLdEm_7aMtFckVJ4nvlyP_KmCQaHSLKm_Ofp9Vo/edit#

Issue #2980, opened 2017-05-22
[Framework Frames] Add libraries


jasonLaster:
We want to add as many libraries for to our new Framework Frames feature. Come add your library.

It's very easy to add your library. Here's the ExtJS [PR](https://github.com/devtools-html/debugger.html/pull/2886), which is a good prototype.

#### Added

- [x] Backbone
- [x] jQuery
- [x] React 
- [x] ExtJS
- [x] Pug
- [x] Express
- [x] Angular
- [x] Ember
- [x] Redux
- [x] Lodash
- [x] Underscore
- [x] Immutable
- [x] RxJS
- [x] Preact 
- [x] Next.js
- [x] Vue
- [x] Marko
- [x] Node internals

#### Screenshots

![call stack.gif](https://shipusercontent.com/a665c85bf97dbc626d1174ccede90a1b/call%20stack.gif)

--- 

![](https://cloud.githubusercontent.com/assets/26968615/26312396/b28dac70-3f07-11e7-9fb1-0bb6632fe3fd.png)

---

<img src='https://shipusercontent.com/0c9db3af12bd9dd678d0825abe740fc6/Screen%20Shot%202017-05-22%20at%2010.21.33%20AM.png' title='Screen Shot 2017-05-22 at 10.21.33 AM.png' width=300>

zaggy:
Hi, can I grab this as my first PR? I'd like to add at least a couple of them for starters. Maybe RxJs and Angular first.

bomsy:
Hey @zaggy go ahead!!

wldcordeiro:
@jasonLaster is there a chance we could introduce some kind of common annotation or interface that libs could utilize to give us this information? Feel like we'll always be playing catch-up with keeping our list of supported libs complete. Might be worth punting on as we develop the idea.

jasonLaster:
Yes. I think we'll want to do that in the future :)

andreicristianpetcu:
Lodash seems to be free. Can I pick it up, @jasonLaster ?

jasonLaster:
Yes. I'm excited about that one

andreicristianpetcu:
Damn! I mixed Underscore and Lodash . I always mix them up :) I added both :P 

![lo_dark](https://user-images.githubusercontent.com/922368/26931110-f2474b4c-4c4e-11e7-8edc-3d3a48a5ac30.png)
[...]

PR #3031, opened 2017-05-30
Q3 Roadmap

jasonLaster:
Issue: #2241

Leaving this open for a bit to gather input:

* are we missing a priority
* should we focus more
* does a feature need more description

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -1,0 +1,34 @@
+### Debugger.html Roadmap
+
+
+#### Q3 Goals (July 1st - September 30th)
+
+**Top Goal**  add features for debugging react apps.
+**Secondary Goals** improve the debugger core performance and experience and continue to invest in the community
+
+##### Framework Improvements
+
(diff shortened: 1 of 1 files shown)


jasonLaster:
We have been getting feedback on issues with server breakpoint toggling and other fundamental issues, which might have to be added to the roadmap as well

svincent:
Random web dev and occasional FF devtools user here. I'm a bit confused about the priority on React and Redux. Without any knowledge of the intent/motivation here this feels like picking favorites. It's hard for me to imagine jQuery/Prototype/YUI/Moo Tools getting this kind of vendor level support.

jasonLaster:
Thanks @svincent I should clarify in the roadmap. We think that libraries should be much more integrated into the debugger.We are starting with react so that we can focus on clearly good ideas, but we fully expect them to generalize to other libraries. An example of this is our new library call stack groups:

https://github.com/devtools-html/debugger.html/issues/2980

Issue #2904, opened 2017-05-15
[Breakpoints] + [SourceMaps] breakpoints are out of sync with the server on reload


jasonLaster:
When a user adds a breakpoint to an original file, makes some changes to the file, and reloads the breakpoint is in the wrong location.

As a bit of background, I found this bug because I was working on Editor/Preview and found that i would add a breakpoint in `renderPreview` and when I refreshed the debugger would pause in a different place than the breakpoint! uhoh :/

The reason that the breakpoint is in a different spot in the UI than where the debugger pauses is because breakpoint locations are stored in the UI as Pending Breakpoints with original locations a.k.a `Editor/Preview.js#23` and on the server as generated locations a.k.a `bundle.js#2343`. So, when the user reloads the client shows the breakpoint on line 23, but the debugger pauses on line 2343, which turns out to be a new line on `Editor/Preview.js`.

codehag:
On it

codehag:
:sob: its done yey!

PR #3039, opened 2017-05-30
[Breakpoints] + [SourceMaps] breakpoints are out of sync with the server on reload       

codehag:
Associated Issue: #2904 

WIP

### DIFF
--- a/src/actions/breakpoints.js
+++ b/src/actions/breakpoints.js
@@ -55,6 +55,8 @@ async function syncClientBreakpoint(
     ? originalToGeneratedId(sourceId)
     : sourceId;
 
+  // this is the generatedLocation of the pending breakpoint, with
+  // the source id updated to reflect the new connection
   const oldGeneratedLocation = {
     ...pendingBreakpoint.generatedLocation,
     sourceId: generatedSourceId
@@ -90,7 +92,7 @@ async function syncClientBreakpoint(
     pendingBreakpoint.location
   );
 
-  // the generatedLocation might have slid, so we adjust it
+  // the generatedLocation might have slid, so now we can adjust it
   const generatedLocation = clientBreakpoint.actualLocation;
 
   const { id, hitCount } = clientBreakpoint;
--- a/src/actions/breakpoints.js
+++ b/src/actions/breakpoints.js
@@ -10,13 +10,14 @@
 
 import { PROMISE } from "../utils/redux/middleware/promise";
 import { getBreakpoint, getBreakpoints, getSource } from "../selectors";
+import { originalToGeneratedId } from "devtools-source-map";
+import { equalizeLocationColumn } from "../utils/breakpoint";
 
 import type { ThunkArgs } from "./types";
-import type { Location } from "../types";
+import type { PendingBreakpoint, Location } from "../types";
 
 type addBreakpointOptions = {
-  condition: string,
-  disabled?: boolean
+  condition: string
 };
 
 function _breakpointExists(state, location: Location) {
@@ -24,8 +25,16 @@ function _breakpointExists(state, location: Location) {
   return currentBp && !currentBp.disabled;
 }
 
-function _createBreakpoint(location: Location, opts: Object = {}) {
-  return Object.assign({}, { location }, opts);
+function _createBreakpoint(location: Object, overrides: Object = {}) {
+  const { condition, disabled, generatedLocation } = overrides;
+  const properties = {
+    condition: condition || null,
+    disabled: disabled || false,
+    generatedLocation,
+    location
+  };
+
+  return properties;
 }
 
 async function _getGeneratedLocation(source, sourceMaps, location) {
@@ -36,17 +45,66 @@ async function _getGeneratedLocation(source, sourceMaps, location) {
   return await sourceMaps.getGeneratedLocation(location, source.toJS());
 }
 
+async function syncClientBreakpoint(
+  sourceId: string,
+  client,
+  sourceMaps,
+  pendingBreakpoint: PendingBreakpoint
+) {
+  const generatedSourceId = sourceMaps.isOriginalId(sourceId)
+    ? originalToGeneratedId(sourceId)
+    : sourceId;
+
+  const generatedLocation = {
+    ...pendingBreakpoint.generatedLocation,
+    sourceId: generatedSourceId
+  };
+
+  // early return if breakpoint is disabled with overrides to update
+  // the id as expected, without talking to server
+  if (pendingBreakpoint.disabled) {
+    return {
+      id: generatedSourceId,
+      actualLocation: { ...pendingBreakpoint.location, id: sourceId },
+      generatedLocation
+    };
+  }
+
+  // If we are not disabled, set the breakpoint on the server and get
+  // that info so we can set it on our breakpoints.
+  const clientBreakpoint = await client.setBreakpoint(
+    generatedLocation,
+    pendingBreakpoint.condition,
+    sourceMaps.isOriginalId(sourceId)
+  );
+
+  // Original location is the location in the src file
+  const clientOriginalLocation = await sourceMaps.getOriginalLocation(
+    clientBreakpoint.actualLocation
+  );
+
+  // make sure that we are re-adding the same type of breakpoint. Column
+  // or line
+  const actualLocation = equalizeLocationColumn(
+    clientOriginalLocation,
+    pendingBreakpoint.location
+  );
+
+  const { id, hitCount } = clientBreakpoint;
+  return { id, actualLocation, hitCount, generatedLocation };
+}
+
 async function addClientBreakpoint(state, client, sourceMaps, breakpoint) {
   const location = breakpoint.location;
   const source = getSource(state, location.sourceId);
-  const sourceLocation = await _getGeneratedLocation(
+  const generatedLocation = await _getGeneratedLocation(
     source,
     sourceMaps,
     location
   );
 
   const clientBreakpoint = await client.setBreakpoint(
-    sourceLocation,
+    generatedLocation,
     breakpoint.condition,
     sourceMaps.isOriginalId(breakpoint.location.sourceId)
   );
@@ -56,12 +114,12 @@ async function addClientBreakpoint(state, client, sourceMaps, breakpoint) {
   );
 
   const { id, hitCount } = clientBreakpoint;
-  return { id, actualLocation, hitCount };
+  return { id, actualLocation, hitCount, generatedLocation };
 }
 
 /**
- * Enabling a breakpoint calls
- * which will reuse the existing breakpoint information that is stored.
+ * Enabling a breakpoint
+ * will reuse the existing breakpoint information that is stored.
  *
  * @memberof actions/breakpoints
  * @static
@@ -83,6 +141,41 @@ export function enableBreakpoint(location: Location) {
 }
 
 /**
+ * Syncing a breakpoint add breakpoint information that is stored, and
+ * contact the server for more data.
+ *
+ * @memberof actions/breakpoints
+ * @static
+ * @param {String} $1.sourceId String  value
+ * @param {PendingBreakpoint} $1.location PendingBreakpoint  value
+ */
+export function syncBreakpoint(
+  sourceId: string,
+  pendingBreakpoint: PendingBreakpoint
+) {
+  return ({ dispatch, getState, client, sourceMaps }: ThunkArgs) => {
+    const { line, sourceUrl, column } = pendingBreakpoint.location;
+    const location = { sourceId, sourceUrl, line, column };
+    const breakpoint = _createBreakpoint(location, pendingBreakpoint);
+
+    const promise = syncClientBreakpoint(
+      sourceId,
+      client,
+      sourceMaps,
+      pendingBreakpoint
+    );
+
+    const action = {
+      type: "SYNC_BREAKPOINT",
+      breakpoint,
+      [PROMISE]: promise
+    };
+
+    return dispatch(action);
+  };
+}
+
+/**
  * Add a new breakpoint
  *
  * @memberof actions/breakpoints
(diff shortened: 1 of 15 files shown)


src/actions/breakpoints.js
```diff
@@ -82,6 +89,41 @@ export function enableBreakpoint(location: Location) {
...
 
+export function syncBreakpoint(location, referenceBreakpoint) {
+  return ({ dispatch, getState, client, sourceMaps }: ThunkArgs) => {
+    if (_breakpointExists(getState(), location)) {
+      return Promise.resolve();
+    }
+
+    const breakpoint = _createBreakpoint(location, referenceBreakpoint);
+
+    let promise = addClientBreakpoint(
```
jasonLaster: I'm a bit confused, so I'll walk through the bug and two workflows as I see them. I could be missing something, it could be what you're already doing too.  I'll outline a simple case and syncing case for both workflows. To be clear, we can do the simple case and fix the bug.

#### Bug

1. User is working on an app that uses source maps (i.e. namely bundle.js w/ file a.js, b.js, c.js)
2. User debugs the app and adds a bp to b.js line 12, with the generated location bundle.js#45
3. User adds code to a.js 
4. User reloads the app and see the debugger paused on b.js line 4, which happens to line up with the generated location bundle.js#45!

We will fix this, either simply moving the breakpoint to b.js line 4 when the debugger reloads or ideally re-removing the breakpoint at line bundle.js#45 and adding a breakpoint at the new location...

--- 

#### simple case

1. User adds a Breakpoint

  * addBreakpoint is called with an original location 
  * generated location is calculated based on the source's sourcemap
  * a pending breakpoint is added with the generated location

2. User reloads page / User re-opens debugger

  * breakpoints reducer adds pending breakpoints 
  * newSource is called and the debugger attempts to sync the breakpoint
       * checks if there is a pending breakpoint
       * uses the pending bps generated location to recalculate the original location
       * adds a new breakpoint to the store


---

#### syncing case

1. User adds a Breakpoint

  * addBreakpoint is called with an original location 
  * generated location is calculated based on the source's sourcemap
  * a pending breakpoint is added with the generated location

2. User reloads page / User re-opens debugger

  * breakpoints reducer adds pending breakpoints 
  * newSource is called and the debugger attempts to sync the breakpoint
       * checks if there is a pending breakpoint
       * uses the pending bps generated location to recalculate the original location
       * **if the new original location is different from the pending breakpoint original location**
          * removes the old breakpoint on the server
          * adds the new breakpoint to the server
       * adds the new breakpoint to the store

src/actions/breakpoints.js
```diff
@@ -43,17 +45,48 @@ async function _getGeneratedLocation(source, sourceMaps, location) {
...
+async function syncClientBreakpoint(
+  state,
+  client,
+  sourceMaps,
+  referenceBreakpoint,
+  breakpoint
+) {
+  console.log("IM CALLED NOW");
+  const location = breakpoint.location;
+  await client.removeBreakpoint(referenceBreakpoint.id);
```
jasonLaster: how do we know we have an ID? the pending bp doesn't have an id

src/utils/breakpoint/index.js
```diff
@@ -0,0 +1,43 @@
+import { updateObj } from "../utils";
+// Return the first argument that is a string, or null if nothing is a
+// string.
+export function firstString(...args) {
```
jasonLaster: this can be moved to `utils/utils.js`

src/utils/breakpoint/index.js
```diff
@@ -0,0 +1,43 @@
...
+  return state.breakpoints.every(x => x.disabled);
+}
+
+// syncing
+export function equalizeLocationColumn(location, hasColumn) {
+  if (hasColumn) {
+    return location;
+  }
+  return Object.assign({}, location, { column: undefined });
+}
```
jasonLaster: 2 things:

1. I think we should do this in `firefox/commands.js#onNewBreakpoint`
2. `onNewBreakpoint` should be moved to firefox/create#createBreakpoint

The rationale for this is to give us consistent data from the server... we did the same thing for source URLs when we worked on pending bps

src/actions/breakpoints.js
```diff
@@ -36,17 +44,55 @@ async function _getGeneratedLocation(source, sourceMaps, location) {
   return await sourceMaps.getGeneratedLocation(location, source.toJS());
 }
 
+async function syncClientBreakpoint(
```
jasonLaster: Lets add a function comment, we don't typically have comments but we should lol

jasonLaster: minor style nit, but i move actiony things to options i.e.

`syncClientBreakpoint(source, pendingBreakpoint, location, {client, sourceMaps})`

src/actions/breakpoints.js
```diff
@@ -36,17 +44,55 @@ async function _getGeneratedLocation(source, sourceMaps, location) {
...
+) {
+  const generatedLocation = await _getGeneratedLocation(
+    source,
+    sourceMaps,
+    location
+  );
+
+  const originalLocation = await sourceMaps.getOriginalLocation(
+    Object.assign({}, generatedLocation, {
+      line: pendingBreakpoint.generatedLocation.line
```
jasonLaster: this seems weird to me... can you explain the rationale

src/actions/breakpoints.js
```diff
@@ -36,17 +44,55 @@ async function _getGeneratedLocation(source, sourceMaps, location) {
   return await sourceMaps.getGeneratedLocation(location, source.toJS());
 }
 
+async function syncClientBreakpoint(
+  source,
+  client,
+  sourceMaps,
+  pendingBreakpoint,
+  location
```
jasonLaster: It's not really clear to me if you need location because it's a composite of many things

src/actions/breakpoints.js
```diff
@@ -82,6 +128,47 @@ export function enableBreakpoint(location: Location) {
...
+        try {
+          client.removeBreakpoint(pendingBreakpoint.id);
+        } catch (_error) {
+          // we always want the same response. if a breakpoint does not exist
+          // we will post an info text that there are no breakpoints to delete
+          console.info("No breakpoint to delete on server");
+        }
+        return Object.assign({}, args, { status: "done" });
+      });
+    }
```
jasonLaster: can we move this try/catch to the firefox/commands side?

src/actions/breakpoints.js
```diff
@@ -82,6 +128,47 @@ export function enableBreakpoint(location: Location) {
...
+      sourceMaps,
+      pendingBreakpoint,
+      location
+    );
+
+    const breakpoint = _createBreakpoint(location, pendingBreakpoint);
+
+    // If the breakpoint is already disabled, we need to communicate
+    // with the server, but we also have to remove the breakpoint immediately.
+    if (pendingBreakpoint.disabled) {
```
jasonLaster: why do we need to talk to the server if it's disabled, i think that means that the server doesn't have a bp?

src/utils/breakpoint/index.js
codehag: :+1:

src/actions/breakpoints.js
codehag: yes. and i agree that this needs to be expressed better. I didn't quite finish fixing this

when the source is reloaded, the source id changes. I wasn't sure what the source was so I quickly threw this in to see if it would work. Basically, there is a new connection. This is why we can't delete / find the line after the debuggee is reloaded. 

for example, 
```javascript
pendingBreakpoint.generatedLocation
/* { 
  id: "server1.conn3.child1/34",  
  line: 51234
} */
generatedLocation // 
/* { 
  id: "server1.conn4.child1/34",  
  line: 51235
} */
```

I didn't get around to doing this properly, but we now have the source available in this function, and this should actually be:
```
const originalLocation = await sourceMaps.getOriginalLocation({
  id: source.id,
  line
});
```

src/actions/breakpoints.js
codehag: Good point here. Originally, my thinking is that we need to verify our breakpoints with the server, and the only way to do that (at the time at least) was to add a breakpoint to the server. If its disabled we need to then remove it. Otherwise we have the weird behavior of disabled breakpoints in places that don't make sense.

Better solution: change the client code so we can get a breakpoint location. I haven't had a chance yet to try to make this better but it feels like it should be possible

src/actions/breakpoints.js
codehag: we now have just 4 items, so this should be ok

src/utils/prefs.js
```diff
@@ -40,4 +41,8 @@ const prefs = new PrefsHelper("devtools", {
   fileSearchRegexMatch: ["Bool", "debugger.file-search-regex-match"]
 });
 
-module.exports = { prefs };
+function setDebuggerVersion() {
+  prefs.debuggerVersion = DEBUGGER_VERSION;
+}
+
+module.exports = { prefs, setDebuggerVersion };
```
jasonLaster: I think it might be nice to keep the schema version private so that we only manage it here...

```js
if (prefs.debuggerPrefsSchemaVersion != prefsSchemaVersion) {
  prefs.clear()
  prefs.debuggerPrefsSchemaVersion = prefsSchemaVersion
}
```

src/reducers/breakpoints.js
```diff
@@ -250,6 +262,34 @@ function setCondition(state, action) {
   }
 }
 
+// Syncing Methods
+function _optimisticlyAddBreakpoint(state, breakpoint) {
```
jasonLaster: nit: not sure if we need the _. I think everything is assumed to be private

src/actions/tests/__snapshots__/pending-breakpoints.js.snap
```diff
@@ -0,0 +1,105 @@
...
+  "disabled": false,
+  "generatedLocation": Object {
+    "column": undefined,
+    "line": 5,
+    "sourceId": "server1.conn76.child1/30",
+    "sourceUrl": "http://localhost:8000/examples/bar.js",
+  },
+  "location": Object {
+    "column": undefined,
+    "line": 7,
```
jasonLaster: it's weird that generatedLocation and location are different line numbers

src/actions/tests/pending-breakpoints.js
```diff
@@ -241,6 +226,47 @@ describe("adding sources", () => {
...
+    expect(startBps.size).toBe(0);
+
+    const source = makeSource("bar.js");
+    const expectedId = makeLocationId(correctedLocation);
+    await dispatch(actions.newSource(source));
+
+    const endBps = selectors.getBreakpoints(getState());
+    const returnedBp = endBps.get(expectedId);
+    expect(returnedBp).not.toBe(bp);
+    expect(returnedBp.location).toEqual(correctedLocation);
```
jasonLaster: might be nice to have a snapshot of `returnedBp` here as well

src/client/firefox/commands.js
```diff
@@ -119,9 +122,13 @@ function onNewBreakpoint(
...
 function removeBreakpoint(breakpointId: BreakpointId) {
-  const bpClient = bpClients[breakpointId];
-  delete bpClients[breakpointId];
-  return bpClient.remove();
+  try {
+    const bpClient = bpClients[breakpointId];
+    delete bpClients[breakpointId];
+    return bpClient.remove();
+  } catch (_error) {
+    console.info("No breakpoint to delete on server");
```
jasonLaster: maybe `console.warn`...

src/reducers/breakpoints.js
```diff
@@ -250,6 +262,34 @@ function setCondition(state, action) {
...
+  };
+
+  return state.setIn(["breakpoints", id], updateObj(breakpoint, updateOpts));
+}
+
+function _commitBreakpoint(state, breakpoint, overrides = {}) {
+  const location = overrides.location || breakpoint.location;
+  const id = makeLocationId(location);
+  const updatedOpts = { ...overrides, loading: false };
+  const updatedBreakpoint = updateObj(breakpoint, updatedOpts);
```
jasonLaster: I think `updateObj` has outlived it's purpose.
I'd also try to be more aggressive about what we want to do.

```js
function updateBreakpoint(state, newBreakpoint, oldBreakpoint) {
  const id = makeLocationId(newBreakpoint.location)
  state.setIn(["breakpoints", id], {
    ...breakpoint,
    ...overrides,
    loading: false
  });
}
```

src/utils/prefs.js
codehag: great idea

codehag: I think the best place to implement the clear function would be in `devtool-module` -- i added a todo note, shall i make a separate issue?

src/actions/tests/__snapshots__/pending-breakpoints.js.snap
codehag: good catch, that might have been a bug later on

src/actions/breakpoints.js
```diff
@@ -83,6 +144,41 @@ export function enableBreakpoint(location: Location) {
...
+      pendingBreakpoint
+    );
+
+    const action = {
+      type: "SYNC_BREAKPOINT",
+      breakpoint,
+      [PROMISE]: promise
+    };
+
+    return dispatch(action);
```
jasonLaster: nit: now that we've simplified this I think we can inline everything 

`dispatch({ type: SYNC_BREAKPOINT, [PROMISE]: syncClientBreakpoint() })`

src/actions/breakpoints.js
```diff
@@ -92,45 +188,22 @@ export function enableBreakpoint(location: Location) {
...
-      Object.assign({}, action, {
-        [PROMISE]: addClientBreakpoint(
-          getState(),
-          client,
-          sourceMaps,
-          breakpoint
-        )
-      })
-    );
+    return dispatch(action);
```
jasonLaster: we can also inline this action too i think

src/utils/prefs.js
```diff
@@ -37,7 +47,18 @@ const prefs = new PrefsHelper("devtools", {
...
   fileSearchWholeWord: ["Bool", "debugger.file-search-whole-word"],
-  fileSearchRegexMatch: ["Bool", "debugger.file-search-regex-match"]
+  fileSearchRegexMatch: ["Bool", "debugger.file-search-regex-match"],
+  debuggerPrefsSchemaVersion: ["Char", "debugger.prefs-schema-version"]
 });
 
+// TODO: replace this with an implementation from devtools-modules
+prefs.clear = () => {
+  prefs.pendingBreakpoints = getDefaultPref("pendingBreakpoints");
+};
```
jasonLaster:
```js
function clearPrefs() {
  prefs.pendingBreakpoints = getDefaultPref("pendingBreakpoints");
}
```

src/utils/prefs.js
```diff
@@ -3,6 +3,15 @@
...
 const { isDevelopment } = require("devtools-config");
+const prefsSchemaVersion = "1.0.0";
+
+function getDefaultPref(prefName) {
+  const defaults = {
+    pendingBreakpoints: []
+  };
+
+  return defaults[prefName];
+}
```
jasonLaster: I think you can get the default for a pref.

Issue #2964, opened 2017-05-19
[Editor] Highlight lines in scope


jasonLaster:
We're currently working on getting the lines in scope when we're paused in #2954. We're doing this so that we can make the lines that are not in scope inert so that we don't accidentally show preview popups. The other benefit is that we can help show the user that we're paused and some of the code is now "live".

Here's a screenshot of what the before and after could be like. The contrast is bumped up for affect, it likely wont be this intense.

<img src='https://shipusercontent.com/1e4c27de23181018d6a9ddbe85759a0c/pause-hgh2.jpg' title='pause-hgh2.jpg' width=3178>

ryanjduffy:
Added a parser function in #3041 to start enabling this feature.

jasonLaster:
There are a couple of follow up issues, but we can mark this as done

ryanjduffy:
@jasonLaster Nice work landing the UI side!

PR #3041, opened 2017-05-30
Add parser function to determine out of scope locations

ryanjduffy:
Associated Issue: #2964

### Summary of Changes

* Adds a new method to find out of scope locations
* Minor refactoring of parser helpers to use new `containsPosition` function

### Test Plan

* Unit tests!

### DIFF
--- a/src/utils/parser/getOutOfScopeLocations.js
+++ b/src/utils/parser/getOutOfScopeLocations.js
@@ -1,0 +1,101 @@
+// @flow
+
+import type { SourceText } from "debugger-html";
+import type { AstLocation, AstPosition } from "./types";
+
+import get from "lodash/fp/get";
+
+import {
+  containsLocation,
+  containsPosition,
+  isFunction
+} from "./utils/helpers";
+import { traverseAst } from "./utils/ast";
+
+/**
+ * Returns all functions (declarations, expressions, arrows) for the given
+ * source
+ */
+function findFunctions(source: SourceText) {
+  const fns = [];
+  traverseAst(source, {
+    enter(path) {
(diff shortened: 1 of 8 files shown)


ryanjduffy:
Changed from my original approach of walking up the AST from a location to find scope boundaries to finding all functions and filtering from there. Both cases required an AST traversal (the former relied on an existing `getClosestScope`) but it was too difficult to "find" potential scopes in the original approach.

A possible solution to reducing AST traversals is to rewrite the functions that require AST searching to all run at the same time and cache relevant data. I've structured this function to adapt well to that in that it retrieves all functions (the cacheable data) and separates map/filter/reduces it to get to a result set based on a location. A similar approach could be taken with the other parser functions.

src/utils/parser/getOutOfScopeLocations.js
```diff
@@ -0,0 +1,101 @@
...
+ */
+const getLocation = path => {
+  let location = Object.assign({}, get("node.loc", path));
+
+  // if the function has an identifier, start the block after it so the
+  // identifier is included in the "scope" of its parent
+  const identifierEnd = get("node.id.loc.end", path);
+  if (identifierEnd) {
+    location.start = identifierEnd;
+  }
```
jasonLaster: cute :)

jasonLaster:
I switched to snapshots, which make it easier to detect changes in the output.

Here's a screenshot of how i verify the results. I think this is better, but I could be persuaded we want to cherry pick data out and make specific assertions. For now, this seems nice though

screen shot 2017-05-30 at 5 02 39 pm

ryanjduffy:
👍

PR #3051, opened 2017-05-31
Highlight Lines V1

jasonLaster:
Associated Issue: #2964 

### Summary of Changes

* redux ast work
* editor render work

### Test Plan

* unit tests

### DIFF
--- a/src/actions/ast.js
+++ b/src/actions/ast.js
@@ -1,4 +1,4 @@
-import { getSourceText, hasSymbols } from "../selectors";
+import { getSelectedLocation, getSourceText, hasSymbols } from "../selectors";
 import parser from "../utils/parser";
 
 export function setSymbols(source) {
@@ -14,6 +14,30 @@ export function setSymbols(source) {
       type: "SET_SYMBOLS",
       source,
       symbols
+    });
+  };
+}
+
+export function setOutOfScopeLocations() {
+  return async ({ dispatch, getState }: ThunkArgs) => {
+    const location = getSelectedLocation(getState());
+
+    if (!location.line) {
+      return dispatch({
+        type: "OUT_OF_SCOPE_LOCATIONS",
+        locations: null
+      });
+    }
+
+    const sourceText = getSourceText(getState(), location.sourceId);
+    const locations = await parser.getOutOfScopeLocations(
+      sourceText.toJS(),
+      location
+    );
+
+    return dispatch({
+      type: "OUT_OF_SCOPE_LOCATIONS",
+      locations
     });
   };
 }
--- a/src/actions/sources.js
+++ b/src/actions/sources.js
@@ -13,7 +13,7 @@ import defer from "../utils/defer";
 import { PROMISE } from "../utils/redux/middleware/promise";
 import assert from "../utils/assert";
 import { updateFrameLocations } from "../utils/pause";
-import { setSymbols } from "./ast";
+import { setSymbols, setOutOfScopeLocations } from "./ast";
 import { addBreakpoint } from "./breakpoints";
 
 import { prettyPrint } from "../utils/pretty-print";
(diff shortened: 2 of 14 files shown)


src/components/Editor/index.js
```diff
@@ -166,6 +169,10 @@ class Editor extends PureComponent {
       this.showSourceText(sourceText, selectedLocation);
     }
 
+    if (this.props._outOfScopeLocations !== nextProps._outOfScopeLocations) {
```
jasonLaster: i probably should rename this var :)

src/reducers/sources.js
```diff
@@ -74,6 +74,10 @@ function update(
     }
 
     case "SELECT_SOURCE":
+      if (action.status != "start") {
+        return state;
+      }
+
```
jasonLaster: select source did get a bit more complicated, but i think it is better now. i.e. we want to know when we're done selecting the source...

src/utils/parser/getOutOfScopeLocations.js
```diff
@@ -94,7 +94,7 @@ function getOutOfScopeLocations(
   return findFunctions(source)
     .map(getLocation)
     .filter(loc => !containsPosition(loc, position))
-    .reduce(removeOverlaps)
+    .reduce(removeOverlaps, [])
```
ryanjduffy: This shouldn't have been necessary. Did you discover an issue?

src/components/Editor/index.js
ryanjduffy: 👍

src/components/Editor/index.js
```diff
@@ -818,6 +828,20 @@ class Editor extends PureComponent {
...
   }
 
+  renderOutOfScopedLocations() {
+    const { _outOfScopeLocations, sourceText, pauseData } = this.props;
+
+    if (!pauseData || !_outOfScopeLocations || !this.editor) {
+      return;
+    }
+
+    _outOfScopeLocations.forEach(location => {
```
ryanjduffy: Could be a bit clearer with `flapMap`:

```
flatMap(_outOfScopeLocations, location => {
  return range(location.start.line, location.end.line);
}).forEach(line => {
  this.editor ...etc ...
});
```

src/reducers/ast.js
```diff
@@ -18,13 +18,15 @@ import type { Record } from "../utils/makeRecord";
 export type SymbolsMap = Map<string, SymbolDeclarations>;
 
 export type ASTState = {
-  symbols: SymbolsMap
+  symbols: SymbolsMap,
+  outOfScopeLocations: ?Object
```
ryanjduffy: The type for this would be `AstLocation[]` (even though I forgot to type the return value ... oops).

src/reducers/sources.js
ryanjduffy: Should it be separate states rather than exception logic within the reducer?

src/utils/parser/getOutOfScopeLocations.js
jasonLaster: yeah - it was weird. it only happened for empty arrays of functions....

src/components/Editor/index.js
jasonLaster: ahh nice

src/reducers/ast.js
jasonLaster: 👍

src/actions/ast.js
```diff
@@ -17,3 +17,18 @@ export function setSymbols(source) {
     });
   };
 }
+
+export function outOfScopeLocations() {
+  return async ({ dispatch, getState }: ThunkArgs) => {
+    const location = getSelectedLocation(getState());
+    const sourceText = getSourceText(getState(), location.sourceId);
```
codehag: since we are using the same state object, maybe we can save it into a variable rather than calling it multiple times

src/actions/sources.js
```diff
@@ -191,18 +191,18 @@ export function selectSource(id: string, options: SelectSourceOptions = {}) {
...
       type: "SELECT_SOURCE",
-      source: source,
+      source: source.toJS(),
       tabIndex: options.tabIndex,
-      line: options.line
+      line: options.line,
+      [PROMISE]: (async function() {
+        await dispatch(loadSourceText(source.toJS()));
+        await dispatch(outOfScopeLocations());
+        return {};
```
codehag: why do we return an empty object here?

src/actions/ast.js
```diff
@@ -17,3 +17,18 @@ export function setSymbols(source) {
...
+    const locations = await parser.getOutOfScopeLocations(
+      sourceText.toJS(),
+      location
+    );
+    dispatch({
```
codehag: should this be returned?

src/actions/sources.js
jasonLaster: meh... it looks weird, right

Issue #3053, opened 2017-05-31
[Preview] simplify the preview logic


jasonLaster:
We have an opportunity to dramatically simplify the preview logic.

* move logic to an action and out of the editor
* relying on expressions from the parser over pause data

https://gist.github.com/jasonLaster/a95eda6eca2100fa832f2ab1c9a1117d

PR #3057, opened 2017-06-01
[WIP] Refactor Preview

jasonLaster:
Associated Issue: #3053

### Summary of Changes

* stop getting variables from pause data
* move logic to an ast selection action
* stop using hidden expressions


src/components/Editor/index.js
```diff
@@ -384,55 +385,29 @@ class Editor extends PureComponent {
...
       return;
     }
 
-    if (selectedToken) {
-      selectedToken.classList.remove("selected-token");
-      this.setState({ selectedToken: null, selectedExpression: null });
-    }
+    const location = getTokenLocation(this.editor.codeMirror, token);
+    setSelection(tokenText, location);
+    console.log("selection", token, tokenText, location);
```
bomsy: rouge `console.log`

src/utils/pause.js
```diff
@@ -6,7 +6,7 @@ export function updateFrameLocations(
   frames: Frame[],
   sourceMaps: any
 ): Promise<Frame[]> {
-  if (!frames || frames.length == 0) {
+  if (true || !sourceMaps || !frames || frames.length == 0) {
```
bomsy: i think the `true` should not be there

src/reducers/ast.js
```diff
@@ -44,6 +45,28 @@ function update(
...
+
+      if (!action.value) {
+        return state.set("selection", null);
+      }
+
+      const { expression, location, result } = action.value;
+      return state.set("selection", {
+        expression,
+        location,
+        result
```
bomsy: do we need to make sure `updating` is false here?

jasonLaster: hmm, i guess i should. was being cute :)

bomsy:
looks good! just tested , it works nicely!
 
There seems to be some lines which should be out of scope which seem to be still on when we hover.

PR #3064, opened 2017-06-02
Promise fixes

jasonLaster:
### Summary of Changes

@bomsy and I found some things while looking at HTML Elements

* promise properties should include state, reason, and value
* __proto__ should be at the bottom.


bomsy:
LG!

Issue #3063, opened 2017-06-02
[ObjectInspector] Keys should be vertically aligned


jasonLaster:
One of the nice things that the old debugger did was vertically align the variable keys. We currently align the triangle and raw values which creates makes it seem like objects are one level deaper

### STR:
1. go to http://localhost:7999/promises.html
2. expand variable `rejectError`

### New Debugger
<img src='https://shipusercontent.com/7330f6a2e86d36fcef0a8c1cab97db36/Screen%20Shot%202017-06-02%20at%2011.01.42%20AM.png' title='Screen Shot 2017-06-02 at 11.01.42 AM.png' width=279>

### Old Debugger
<img src='https://shipusercontent.com/2939837c48745c8950abfdf3e8e1948d/Screen%20Shot%202017-06-02%20at%2011.01.36%20AM.png' title='Screen Shot 2017-06-02 at 11.01.36 AM.png' width=277>

#### Notes

```diff
diff --git a/src/components/shared/ManagedTree.js b/src/components/shared/ManagedTree.js
index c7b7dd8..7ebd1b3 100644
--- a/src/components/shared/ManagedTree.js
+++ b/src/components/shared/ManagedTree.js
@@ -115,6 +115,7 @@ class ManagedTree extends Component {
   render() {
     const { expanded, focusedItem } = this.state;

+    // we might need to change something here
     const props = Object.assign({}, this.props, {
       isExpanded: item => expanded.has(this.props.getKey(item)),
       focused: focusedItem,
diff --git a/src/components/shared/ObjectInspector.js b/src/components/shared/ObjectInspector.js
index 3679112..59e4c96 100644
--- a/src/components/shared/ObjectInspector.js
+++ b/src/components/shared/ObjectInspector.js
@@ -138,6 +138,9 @@ class ObjectInspector extends Component {
           focused,
           "default-property": isDefault(item)
         }),
+        // maybe we want to look here
+        // or consider using a before element?
+
         style: { marginLeft: depth * 15 },
         onClick: e => {
           e.stopPropagation()
```

peterschussheim:
@jasonLaster mind if I take this one?

jasonLaster:
Go for it

peterschussheim:
[New issue](https://github.com/devtools-html/debugger.html/issues/3088) created to add support for aligning keys for users of `RTL` setting.

PR #3079, opened 2017-06-03
[ObjectInspector] Vertically align variable keys when inspecting objects

peterschussheim:
Associated Issue: #3063

### Summary of Changes
In the `ObjectInspector` component I added a check if `item` is primitive, and if true, increase the left margin by 15px.

### Test Plan
 - [x] yarn test-all
 - [x] visual inspection (LTR)
 - [ ] visual inspection (RTL): Having trouble with this setting.

#### Before:
Before

---
#### After:
After-Light
After-Dark

### DIFF
--- a/src/components/shared/ObjectInspector.js
+++ b/src/components/shared/ObjectInspector.js
@@ -139,7 +139,7 @@ class ObjectInspector extends Component {
           "default-property": isDefault(item)
         }),
         style: {
-          marginLeft: nodeIsPrimitive(item) ? depth * 15 + 15 : depth * 15
+          marginLeft: depth * 15 + (nodeIsPrimitive(item) ? 15 : 0)
         },
         onClick: e => {
           e.stopPropagation();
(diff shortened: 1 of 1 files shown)


src/components/shared/ObjectInspector.js
```diff
@@ -138,7 +138,9 @@ class ObjectInspector extends Component {
           focused,
           "default-property": isDefault(item)
         }),
-        style: { marginLeft: depth * 15 },
+        style: {
+          marginLeft: nodeIsPrimitive(item) ? depth * 15 + 15 : depth * 15
```
jasonLaster: perhaps: `depth * 15 + (nodeIsPrimitive(item) ?  15 : 0)` to be a little cleaner

peterschussheim:
Looking at docs for logical-css properties for a solution to `RTL` mode

jasonLaster:
Hmm. Yeah. We basically need to do the same for margin right in the rtl case

jasonLaster:
We should probably have a styling util so you can do marginStart

peterschussheim:
Yes I was thinking the same on using marginStart; when you say a utility, this would require additional CSS and perhaps a small function?

jasonLaster:
Probably a small function - file an issue for it, so we don't lose it

Issue #3078, opened 2017-06-03
Consolidate Source and SourceText


jasonLaster:
We currently have two store fields Source and SourceText. I think it would be nice to consolidate them. 

This would mean `Source` would have this type:

```js
 Source = {
    id: SourceId,
    url?: string,
    sourceMapURL?: string,
    isBlackBoxed: boolean,
    isPrettyPrinted: boolean,
    text?: string,
    contentType?: string,
    loading?: boolean
  };
```

The benefits of consolidation is that it would be easier to get a source's text. The two fields are different because of the way the protocol works, but that is largely an implementation detail.

Here is a [gist] for the refactor. I'm curious what others think.

[gist]:https://gist.github.com/jasonLaster/203d87ae579f08e808c0490d8ab480bd

clarkbw:
Looks like this was done in #3081 already

jasonLaster:
not done yet :)

jasonLaster:
This is done!

PR #3081, opened 2017-06-04
[WIP] consolidate source text

jasonLaster:
Associated Issue: #3078

### Summary of Changes

I started refactoring the sources store to remove SourceText. The reason is that it is not needed and creates extra complexity around selectors and typing in our components, actions, and utilities.

### DIFF
--- a/src/components/Editor/SearchBar.js
+++ b/src/components/Editor/SearchBar.js
@@ -31,7 +31,7 @@ import classnames from "classnames";
 import debounce from "lodash/debounce";
 
 import { SourceEditor } from "devtools-source-editor";
-import type { SourceRecord, SourceTextRecord } from "../../reducers/sources";
+import type { SourceRecord } from "../../reducers/sources";
 import type { FileSearchModifiers, SymbolSearchType } from "../../reducers/ui";
 import type { SelectSourceOptions } from "../../actions/sources";
 import type { SearchResults } from ".";
@@ -98,7 +98,6 @@ class SearchBar extends Component {
   props: {
     editor?: SourceEditor,
     symbols: SymbolDeclarations,
-    sourceText?: SourceTextRecord,
     selectSource: (string, ?SelectSourceOptions) => any,
     selectedSource?: SourceRecord,
     highlightLineRange: ({ start: number, end: number }) => any,
@@ -203,7 +202,6 @@ class SearchBar extends Component {
 
   componentDidUpdate(prevProps: any, prevState: any) {
     const {
-      sourceText,
       selectedSource,
       query,
       modifiers,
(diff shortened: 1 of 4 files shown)


src/reducers/sources.js
```diff
@@ -149,40 +149,40 @@ function update(
...
-        error: action.error
-      })
-    );
+// TODO: Action is coerced to `any` unfortunately because how we type
+// asynchronous actions is wrong. The `value` may be null for the
+// "start" and "error" states but we don't type it like that. We need
+// to rethink how we type async actions.
+function updateText(state, action: any): Record<SourcesState> {
+  const source = action.source;
+  const text = getTextFromAction(action);
```
codehag: this can be more than text. for example it can also have an id, content type, and loading state. maybe we can name this a bit better? `sourceTextProps` comes to mind, or just `updateProps`

src/reducers/sources.js
```diff
@@ -149,40 +149,40 @@ function update(
   return state;
 }
 
-// TODO: Action is coerced to `any` unfortunately because how we type
-// asynchronous actions is wrong. The `value` may be null for the
-// "start" and "error" states but we don't type it like that. We need
-// to rethink how we type async actions.
-function _updateText(state, action: any): Record<SourcesState> {
+function getTextFromAction(action: any) {
```
codehag: this returns more than text, maybe `getTextPropsFromAction` is clearer?

src/reducers/sources.js
```diff
@@ -149,40 +149,40 @@ function update(
...
 
-  return state.setIn(
-    ["sourcesText", source.id],
-    I.Map({
-      text: sourceText.text,
-      id: source.id,
-      contentType: sourceText.contentType
-    })
-  );
+  return state.mergeIn(["sources", source.id], source);
```
codehag: maybe we can have a comment here about why we are merging, rather than setting? I saw above there was such a comment in the previous version of the code, might be helpful here too

src/components/Editor/SearchBar.js
```diff
@@ -221,9 +219,9 @@ class SearchBar extends Component {
       scrollList(this.refs.resultList.refs, this.state.selectedResultIndex);
     }
 
-    const hasLoaded = sourceText && !sourceText.get("loading");
+    const hasLoaded = selectedSource && !selectedSource.get("loading");
```
codehag: :+1:

src/reducers/sources.js
jasonLaster: I think the comment before covered a *special* case that we no longer are concerned with. I am not sure if I have something that I want to say here except that we're adding source fields :)

Issue #3080, opened 2017-06-04
[flow] switch action types to exact object types


jasonLaster:
We currently type our actions, but it would be nice to use exact object types as it will catch additional cases especially when we use the PROMISE middleware.

Here's a rough [sketch](https://gist.github.com/7530eb2fed2d67ccadce4e6d3979faf9) of what the diff would look like. After this, it's just a question of making flow happy :)

[types]:https://flow.org/en/docs/types/objects/#toc-exact-object-types

sejoker:
@jasonLaster I think I can tackle it.

jasonLaster:
@sejoker that'd be great. It might be nice to start slow. Like convert once action at a time or an action type like source actions

jasonLaster:
@sejoker are you available to take a stab at breaking this down into smaller pieces?

wldcordeiro:
I think exact types with covariant properties will be even better.

```js
type MyAction = {|
  +id: string,
  +value: number,
|}
```

This will reflect that our Redux states are immutable. https://flow.org/en/docs/react/redux/#toc-typing-redux-state-immutability

jasonLaster:
Yeah

jasonLaster:
Hey @sejoker, @wldcordeiro  any interest in picking this up and landing a couple of these refactors? It'd be nice land a couple to get a precedent in here.

ryanjduffy:
I took a cut at this and wanted to get some feedback before pursuing further. Specifically, I added a generic type for the promise middleware to attempt to handle the various structs that are valid for each action state (new, start, error, and done).

Here's the generic:

https://github.com/ryanjduffy/debugger.html/blob/f271f4eb667d52a8fae6594c06d6b33debd0030a/src/actions/utils/middleware/promise.js#L12-L41

And an example of usage:

https://github.com/ryanjduffy/debugger.html/blob/f271f4eb667d52a8fae6594c06d6b33debd0030a/src/actions/types.js#L93-L100

One disadvantage is that you have to guard access to some members because Flow can't seem to sort out that they are valid. For example, any time you need to get at the `value` of an action, you either have to check for existence (`if (action.value)`) or forcibly cast it.

https://github.com/ryanjduffy/debugger.html/blob/f271f4eb667d52a8fae6594c06d6b33debd0030a/src/reducers/pending-breakpoints.js#L94

I haven't mark the props as covariant yet so that still outstanding work.

Thoughts?

jasonLaster:
[...]

PR #3084, opened 2017-06-04
switch action types to exact object types

sejoker:
Associated Issue: https://github.com/devtools-html/debugger.html/issues/3080

### Summary of Changes

* use exact object types for Actions
* added missing action types, e.g. CLEAR_SELECTED_SOURCE
* explicit casting of the actions to Action type
* made some parameters optional based on the usage patterns

### Test Plan

- `npm run test`
- `npm run flow`
- `npm run flow-redux`

flow coverage before:
screen shot 2017-06-04 at 6 42 35 pm

flow coverage before:
screen shot 2017-06-04 at 6 41 49 pm

changing ThunkArgs:

```
export type ThunkArgs = {
  dispatch: Action => Promise<any>,
  getState: () => State,
  client: any,
  sourceMaps: any
};
```
still produces 39 errors:
examples of problematic calls are:
dispatch(loadSourceText(source)).then(
                        ^^^^^^^^^^^^^^^^^^^^^^ async function. Inexact type is incompatible with exact type

dispatch(setSymbols(source));
                             ^^^^^^^^^^^^^^^^^^ async function. Inexact type is incompatible with exact type

### DIFF
--- a/src/actions/ast.js
+++ b/src/actions/ast.js
@@ -48,7 +48,7 @@ export function setOutOfScopeLocations() {
       return dispatch(
         ({
           type: "OUT_OF_SCOPE_LOCATIONS",
-          locations: []
+          locations: undefined
         }: Action)
       );
     }
--- a/src/actions/ast.js
+++ b/src/actions/ast.js
@@ -13,7 +13,7 @@ import { PROMISE } from "../utils/redux/middleware/promise";
 import parser from "../utils/parser";
 
 import type { Source } from "debugger-html";
-import type { ThunkArgs } from "./types";
+import type { ThunkArgs, Action } from "./types";
 import type { AstLocation } from "../utils/parser";
 
 export function setSymbols(source: Source) {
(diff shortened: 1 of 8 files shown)


jasonLaster:
strange that the coverage did not change too much...

* I think this is an improvement as we get more error checking
* I think we can easily remove the casting with a regex that drops `: Action` and prettier does the rest
* I'm worried about these [error msgs](https://gist.github.com/jasonLaster/322359bfda1788272b0401ea6495b755). It's kinda scary. The issue comes from the unions. Maybe we move the action types to the action files and cast directly 

```js
dispatch({ type: "ADD_SOURCES", sources: originalSources }:  AddSources)
```
We can still union at the `action/types.js` level but the advantage will be that we will have definitions closer to the source and better errors.

I would like to eventually drop the casting, but it looks like we won't be able to do that until we've casted all of the actions, is that right? Like when we're 100% then we can do it...

Curious what you think

sejoker:
I agree with 1st and 2nd points. As for [errors messages](https://gist.github.com/jasonLaster/322359bfda1788272b0401ea6495b755)  I could play with different approaches and return with my proposal.

sejoker:
I will create separate PRs for each action groups, closing this one.

Issue #1108, opened 2016-11-04
[ProjectSearch] tracking issue


jasonLaster:
We'd like to re-add the ability to search across the project.

Links 
* Here's the search [invision](https://projects.invisionapp.com/share/AY8OEG3C6#/screens/117937554_Debugger)
* Here's a link to [chrome's rdp method](https://chromium.googlesource.com/v8/v8/+/master/src/inspector/js_protocol.json#580)

Steps:
- [x] add a UI for showing project search #3090
- [x] Add a component for toggling project search between text and files #3091
- [x] add a text search function 
- [x] Add a redux action for loading source texts #3331
- [x] Move search functionality to redux #3342
- [x] Refactor Project Search to support text search
- [x] Make source search stream #3270
- [x] Create a "search" tab #3279
- [x] Cleanup the css and styles for dark theme #3326
- [x] Auto focus on the first result in the list
- [x] Highlight line where the text match was found when source file is opened #3467
- [x] Add keyboard actions for accessibility #3501
- [x] should scroll to keep items in focus
- [x] File path does not appear in nightly #3782
- [ ] Improve Text Search -> Source Transition #3787
- [ ] Flow type the UI - I started [here], but struggled with ManagedTree.

[here]: https://github.com/jasonLaster/debugger.html/tree/ts-filter-paths

clarkbw:
Pulling this image out so you don't have to jump to the doc.

<img width="883" alt="screen shot 2017-03-12 at 12 03 40 pm" src="https://cloud.githubusercontent.com/assets/2134/23834938/f5d51258-071b-11e7-8dca-3017db057b9a.png">

wldcordeiro:
I'd love to get hacking on this. What/where would I need to work to do these parts?

> add a debugger server endpoint
> add debugger client + command methods

jasonLaster:
I'll break it down soon :)

It's on the q2 roadmap!

clarkbw:
@wanderview really wants this 😄

jasonLaster:
I started hacking on this earlier today and made some progress

Here's the [gist]. I'll post a couple PRs based on the work so that it doesn't go stale soon

- [ ] add a source search function
- [ ] Refactor Project Search to support text search
- [ ] Add a component for toggling project search between text and files
- [ ] Add a redux action for loading source texts

I started one and two and punted on three and four naturally :)


[...]

PR #3089, opened 2017-06-05
Project Search (components)

jasonLaster:
Associated Issue: #1108

### Summary of Changes

* starts the UI Component TextSearch
* starts the util for searching
* adds a bunch of stories


src/components/ProjectSearch/ProjectSearch.js
```diff
@@ -3,33 +3,63 @@
...
   }
 
-  return sources
+  const { path, href } = parseURL(source.url);
+  // for URLs like "about:home" the path is null so we pass the full href
+  return path || href;
+}
+
+function searchResults(sources: SourcesMap, query) {
+  if (false) {
```
codehag: this looks weird, because we will never enter this loop. should it be deleted, or updated to be meaningful?

src/components/ProjectSearch/ProjectSearch.js
```diff
@@ -87,23 +117,30 @@ class ProjectSearch extends Component {
     this.props.toggleProjectSearch(false);
   }
 
+  renderFileSearch() {
```
codehag: should this be used in render? it looks like it has logic that was moved out of the render function

src/components/ProjectSearch/TextSearch.css
```diff
@@ -0,0 +1,18 @@
...
+  font-weight: bold;
+  margin-top: 20px;
+  line-height: 20px;
+}
+
+.project-text-search .line-match {
+  font-family: monospace
+}
+
+.project-text-search .result .line-number {
```
codehag: this is getting pretty specific. maybe we could use BEM to keep this from going too far. maybe something like
`.project-text-search__line-number`. Also result seems like a bit of a dangerous css selector, as I can see it being used in a lot of places for different purposes.

src/components/ProjectSearch/TextSearch.js
```diff
@@ -0,0 +1,73 @@
...
+export default class TextSearch extends Component {
+  constructor(props: Props) {
+    super(props);
+  }
+
+  renderFile(file, expanded) {
+    return dom.div(
+      {
+        className: "file-result",
+        style: {}
```
codehag: do we need style here?

src/components/ProjectSearch/TextSearch.js
```diff
@@ -0,0 +1,73 @@
...
+    const { results } = this.props;
+    return ManagedTree({
+      getRoots: () => results,
+      getChildren: file => {
+        return file.matches || [];
+      },
+      itemHeight: 20,
+      autoExpand: 1,
+      autoExpandDepth: 1,
+      getParent: item => null,
```
codehag: do we need item here? it seems like `getParent` will always return `null`

src/components/ProjectSearch/TextSearch.js
```diff
@@ -0,0 +1,73 @@
...
+      dom.span(
+        { className: "line-match", style: { display: "flex", grow: 1 } },
+        match.value
+      )
+    );
+  }
+
+  renderResults() {
+    const { results } = this.props;
+    return ManagedTree({
```
codehag: this is very complex, it feels like most of these callbacks belong to ManagedTree. Maybe a job for a future refactoring

src/components/shared/ManagedTree.js
```diff
@@ -62,7 +65,7 @@ class ManagedTree extends Component {
     }
   }
 
-  setExpanded(item: ManagedTreeItem, isExpanded: boolean) {
+  setExpanded(item: Item, isExpanded: boolean) {
```
codehag: :+1:

src/components/shared/ResultList.js
```diff
@@ -11,31 +11,34 @@ type ResultListItem = {
...
   static defaultProps: Object;
 
-  constructor(props: any) {
+  constructor(props: Props) {
     super(props);
     (this: any).renderListItem = this.renderListItem.bind(this);
   }
 
   renderListItem(item: ResultListItem, index: number) {
+    const { selectItem, selected } = this.props;
```
codehag: :+1:

src/components/ProjectSearch/TextSearch.css
jasonLaster: yeah - these are throw away styles. I think something BEM like would be better

src/components/ProjectSearch/TextSearch.js
jasonLaster: oops, i'll remove it, but these are throw away as i was quickly sketching

PR #3098, opened 2017-06-06
Release a new version of the debugger

jasonLaster:
* Updates from the past week
* Small fixes for the editor footer and a couple small details for tests

### DIFF
--- a/src/components/Editor/Editor.css
+++ b/src/components/Editor/Editor.css
@@ -23,10 +23,6 @@
   --theme-conditional-breakpoint-color: #ccd1d5;
 }
 
-.theme-firebug {
-  --theme-conditional-breakpoint-color: #ccd1d5;
-}
-
 .out-of-scope .CodeMirror-line,
 .out-of-scope .CodeMirror-linenumber {
   opacity: 0.8;
--- a/src/components/Editor/Editor.css
+++ b/src/components/Editor/Editor.css
@@ -1,6 +1,9 @@
 .editor-wrapper {
   --debug-line-background: rgba(226, 236, 247, 0.5);
   --debug-line-border: rgb(145, 188, 219);
+  --editor-searchbar-height: 27px;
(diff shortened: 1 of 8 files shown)


src/components/Editor/Editor.css
```diff
@@ -1,6 +1,9 @@
 .editor-wrapper {
   --debug-line-background: rgba(226, 236, 247, 0.5);
   --debug-line-border: rgb(145, 188, 219);
+  --editor-searchbar-height: 27px;
+  --editor-second-searchbar-height: 27px;
```
codehag: i know this is copy pasted, but do we need two separate declarations of the same value? 🤔

src/components/Editor/Tabs.js
```diff
@@ -50,7 +50,8 @@ function getHiddenTabs(sourceTabs: SourcesList, sourceTabEls) {
   return sourceTabs.filter((tab, index) => {
     // adding 10px helps account for cases where the tab might be offset by
     // styling such as selected tabs which don't have a border.
-    return sourceTabEls[index].getBoundingClientRect().top > tabTopOffset + 10;
+    const el = sourceTabEls[index];
+    return el && el.getBoundingClientRect().top > tabTopOffset + 10;
```
codehag: Nice! i would split this up more even, so that we know what is happening.

```js
const element = sourceTabEls[index];
const hasTopOffset = (el) => (el.getBoundingClientRect().top > tabTopOffset + 10)
return element && hasTopOffset(el);
```

src/components/Editor/Editor.css
jasonLaster: probably not. that can be tweaked when we have a function search modal :)

src/components/Editor/Tabs.js
jasonLaster: that's a good idea. I took it a couple step further lol.

Issue #2946, opened 2017-05-17
[ObjectInspector] We should only style default window properties when viewing a window object


jasonLaster:
We currently style window properties with a lighter shade of blue all the time.

We should only do this when viewing a window object

<img src='https://shipusercontent.com/e4885b6156dbcea152c023b4756bda7a/Screen%20Shot%202017-05-17%20at%205.47.32%20PM.png' title='Screen Shot 2017-05-17 at 5.47.32 PM.png' width=297>

jbhoosreddy:
I was going through the object inspector code to see where this was coming from.. but still this is a pretty interesting bug (to find).

jasonLaster:
@jbhoosreddy it certainly is:

https://github.com/devtools-html/debugger.html/blob/master/src/utils/object-inspector.js#L66-L69

Try running `yarn storybook` and going to preview :)

jbhoosreddy:
fixed in #3099

jasonLaster:
Still seeing this

<img src='https://shipusercontent.com/6c7f72bf68663b3251807d5f215dca7a/Screen%20Shot%202017-07-21%20at%2010.26.49%20AM.png' title='Screen Shot 2017-07-21 at 10.26.49 AM.png' width=598>

jasonLaster:
@jbhoosreddy @nchevobbe i'm worried that to properly diagnose this, we need to know about the grip we're previewing which will change OI in a decently large way.

nchevobbe:
What about having a `parent` property on the ObjectInspectorItem so we can retrieve it quickly ?
Also, it seems that we don't have `default-property` CSS class in devtools-reps, which I'll add.

This leads to another issue, when logging the window to the console, I don't want it to have this style applied. I see value having it in the scope, since it's always there, so I'll think of something to be able to enable/disable this behavior.

jasonLaster:
that sounds good - i like having parent passed down to `renderItem`. I'll see what I can do with the `tree` component.

nchevobbe:
I was thinking to do it directly in the `createNode` function, which would then return `{name, path, contents, parent}`.
This way we would have access to the parent item for each item (and not only in the renderItem function).
In fact, I am already experimenting with it for the "show set/map entries in OI".
I'll ping you on the PR so you can tell me how it feels.

jasonLaster:
That would be even better!

jasonLaster:
we can do this post release

nchevobbe:
[...]

PR #3099, opened 2017-06-06
fix object property styling on window property names (#2946)

jbhoosreddy:
Associated Issue: #2946

### Summary of Changes

* Adds additional logic so that window property names render properly on objects

### Test Plan

Visual Testing

1. Run debugger
2. Enter expression in watch expression: `a = { aLocation: 1, location: 2};`
3. Both properties should display in the same design. 

screen shot 2017-06-05 at 10 33 39 pm


jasonLaster:
Not sure bout those aXXX cars, but I'll forgive you

Issue #3038, opened 2017-05-30
[Editor Search] Add a search for text button


jasonLaster:
We currently show `Search for: functions, variables`, it'd be nice to switch it to: `Search: text, functions, variables` where each of the options acted like radio buttons.

```diff
diff --git a/src/components/Editor/SearchBar.js b/src/components/Editor/SearchBar.js
index bbad434..393e847 100644
--- a/src/components/Editor/SearchBar.js
+++ b/src/components/Editor/SearchBar.js
@@ -662,6 +662,7 @@ class SearchBar extends Component {
     const { symbolSearchOn, selectedSymbolType } = this.props;

     function searchTypeBtn(searchType) {
+      // this will need some massaging
       return dom.button(
         {
           className: classnames("search-type-btn", {
@@ -685,6 +686,7 @@ class SearchBar extends Component {
         { className: "search-toggle-title" },
         L10N.getStr("editor.searchTypeToggleTitle")
       ),
+      searchTypeBtn("text"),
       searchTypeBtn("functions"),
       searchTypeBtn("variables")
     );
```

<img src='https://shipusercontent.com/4d899465882ad561e6c8298432928248/Screen%20Shot%202017-05-30%20at%2010.33.27%20AM.png' title='Screen Shot 2017-05-30 at 10.33.27 AM.png' width=407>

ruturajv:
What would you include in text ?
- variable values
- comments
?

jasonLaster:
Oh text is just CMD+F regular text search :)

PR #3100, opened 2017-06-06
WIP - Bug 3038 editor search add text search button

ruturajv:
Associated Issue: #3038 

Even though things are working... the flow gives following errors. Except for the SearchBar.js I'm not sure. Could you guys help me with this?

```
$ yarn flow
yarn flow v0.24.5
$ flow 
Launching Flow server for /home/rutu/code/debugger.html
Spawned flow server (pid=21157)
Logs will go to /tmp/flow/zShomezSrutuzScodezSdebugger.html.log
SearchBar.js:4
  4: import { findDOMNode } from "../../../node_modules/react-dom/dist/react-dom";
                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ ../../../node_modules/react-dom/dist/react-dom. Required module not found

SearchBar.js:8
  8: import Svg from "../shared/Svg";
                     ^^^^^^^^^^^^^^^ ../shared/Svg. Required module not found

SearchBar.js:9
  9: import actions from "../../actions";
                         ^^^^^^^^^^^^^^^ ../../actions. Required module not found

SearchBar.js:18
 18: } from "../../selectors";
            ^^^^^^^^^^^^^^^^^ ../../selectors. Required module not found

SearchBar.js:27
 27: } from "../../utils/editor";
            ^^^^^^^^^^^^^^^^^^^^ ../../utils/editor. Required module not found

SearchBar.js:29
 29: import { scrollList } from "../../utils/result-list";
                                ^^^^^^^^^^^^^^^^^^^^^^^^^ ../../utils/result-list. Required module not found

SearchBar.js:34
 34: import type { SourceRecord } from "../../reducers/sources";
                                       ^^^^^^^^^^^^^^^^^^^^^^^^ ../../reducers/sources. Required module not found

SearchBar.js:35
 35: import type { FileSearchModifiers, SymbolSearchType } from "../../reducers/ui";
                                                                ^^^^^^^^^^^^^^^^^^^ ../../reducers/ui. Required module not found

SearchBar.js:36
 36: import type { SelectSourceOptions } from "../../actions/sources";
                                              ^^^^^^^^^^^^^^^^^^^^^^^ ../../actions/sources. Required module not found

SearchBar.js:37
 37: import type { SearchResults } from ".";
                   ^^^^^^^^^^^^^ Named import from module `.`. This module has no named export called `SearchResults`.

SearchBar.js:40
 40: import _SearchInput from "../shared/SearchInput";
                              ^^^^^^^^^^^^^^^^^^^^^^^ ../shared/SearchInput. Required module not found

SearchBar.js:43
 43: import _ResultList from "../shared/ResultList";
                             ^^^^^^^^^^^^^^^^^^^^^^ ../shared/ResultList. Required module not found

SearchBar.js:46
 46: import type { SymbolDeclaration } from "../../utils/parser/getSymbols";
                                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ ../../utils/parser/getSymbols. Required module not found

SearchBar.js:93
 93: import "./SearchBar.css";
            ^^^^^^^^^^^^^^^^^ ./SearchBar.css. Required module not found

src/components/Editor/SearchBar.js:307
307:     if (searchType === "text") {
                            ^^^^^^ string literal `text`. This type is incompatible with
298:     { toggle, searchType }: ToggleSymbolSearchOpts = {}
                   ^^^^^^^^^^ string enum

src/components/Editor/SearchBar.js:680
680:               toggleSymbolSearch(e, { toggle: true, searchType });
                                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ object literal. This type is incompatible with the expected param type of
298:     { toggle, searchType }: ToggleSymbolSearchOpts = {}
                                 ^^^^^^^^^^^^^^^^^^^^^^ object type
  Property `searchType` is incompatible:
    680:               toggleSymbolSearch(e, { toggle: true, searchType });
                                                             ^^^^^^^^^^ string. This type is incompatible with
     83:   searchType: SymbolSearchType
                       ^^^^^^^^^^^^^^^^ string enum

src/components/Editor/SearchBar.js:683
683:             toggleSymbolSearch(e, { toggle: false, searchType });
                                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ object literal. This type is incompatible with the expected param type of
298:     { toggle, searchType }: ToggleSymbolSearchOpts = {}
                                 ^^^^^^^^^^^^^^^^^^^^^^ object type
  Property `searchType` is incompatible:
    683:             toggleSymbolSearch(e, { toggle: false, searchType });
                                                            ^^^^^^^^^^ string. This type is incompatible with
     83:   searchType: SymbolSearchType
                       ^^^^^^^^^^^^^^^^ string enum


Found 17 errors

```


codehag:
Hi Rutrajv! Thanks for helping out!

I ran the project locally with your changes, and, while I get 3 of the errors that you are showing here, I didn't get any of the import errors. Is there anything else I need to know about your setup?

Regarding the other errors:

Flow types can be a bit tricky if you are just starting out with them. If you aren't familiar with typing, typing arguments and outputs allows us to know that the inputs and outputs of a given function are correct. This simplifies testing and gives us more certainty in our code. So lets start with a simple error and trace our way back.

```
src/components/Editor/SearchBar.js:307
307:     if (searchType === "text") {
                            ^^^^^^ string literal `text`. This type is incompatible with
298:     { toggle, searchType }: ToggleSymbolSearchOpts = {}
                   ^^^^^^^^^^ string enum
```

so this is telling us that `searchType` should be of type `string enum`, but we are getting "text" instead, which is a `string literal`. To find out what this all means, its useful to look at the source of this type. We can find this, on line 35 of src/components/Editor/SearchBar.js

`import type { FileSearchModifiers, SymbolSearchType } from "../../reducers/ui";`

What we can do now is navigate into that file (src/reducers/ui.js) and take a look. There we can find a declaration for `SymbolSearchType` on line 20. It looks like this:

`export type SymbolSearchType = "functions" | "variables";`

This flow type is pretty limited. The only acceptable strings for a SymbolSearchType are "functions", and "variables" (`or` in flow typing is denoted by a single pipe `|`). It doesn't accept "text"! But we can add it like so:

`export type SymbolSearchType = "functions" | "variables" | "text";`

The same can go for 
```
298:     { toggle, searchType }: ToggleSymbolSearchOpts = {}
                                 ^^^^^^^^^^^^^^^^^^^^^^ object type
```

where the default value for ToggleSymbolSearchOpts is an empty object, but both the `toggle` and `searchType` are restricted to only accepting certain values. One question might be: do we need the default value? if so, should it be empty? and if yes, we can make the property optional! Let me know if this is the case and I will help you out.

Hopefully this helps! let me know if you have more questions!

src/components/Editor/SearchBar.js
```diff
@@ -660,10 +664,16 @@ class SearchBar extends Component {
     const { symbolSearchOn, selectedSymbolType } = this.props;
 
     function searchTypeBtn(searchType) {
+      let active =
```
codehag: Since `active` isn't redefined, we should prefer `const` here instead

src/components/Editor/SearchBar.js
```diff
@@ -660,10 +664,16 @@ class SearchBar extends Component {
...
+        (symbolSearchOn &&
+          selectedSymbolType == searchType &&
```
codehag: we should prefer identity equality `===` instead of equality `==` here!

codehag: we should prefer identity equality === instead of equality == here!

src/components/Editor/SearchBar.js
codehag: Since this isn't redefined, we should prefer `const` over `let`

src/components/Editor/SearchBar.js
```diff
@@ -660,10 +664,16 @@ class SearchBar extends Component {
...
+          searchType !== "text") ||
+        (!symbolSearchOn && searchType === "text");
```
codehag: This is a pretty complex boolean. I think it would be more readable if we split this into functions. Maybe something like this:

```
const { symbolSearchOn, selectedSymbolType } = this.props;

function isSymbolSearchOn(searchType) {
  // since the boolean is so complex, consisting of three parts, lets break it up
  // and use clear variable names to make it obvious what is happening
  const isSymbolSearch = selectedSymbolType === searchType;
  const notSearchText = searchType !== "text";

  return (symbolSearchOn && isSearchType && notSearchText);
}

function isSearchText(searchType) {
  return (!symbolSearchOn && searchType === "text");
}

function isTextSearch(searchType) {
  const { symbolSearchOn } = this.props;
  return (symbolSearchOn && searchType === "text");
}

const active = isSymbolSearchOn() || isSearchText()
```

It also seems like SymbolSearch and SearchText are two different things that are at odds with eachother. You mentioned elsewhere that you have problems with the flow typing, maybe we need a new function, which is "searchTextBtn" with a new variable that has its own flow type? That might simplify the code.

ruturajv:
Hey @codehag 

Thanks a lot for your help, yes I'm new to flow.

Regarding the simplification of `active` - Yes I too think its too complicated right now.

Let me work on your suggestions.

ruturajv:
Hey @codehag  

I've fixed some of the errors by not checking for `text` string.

The `searchTypeBtn` was initially just generating buttons for symobolSearch related filters, now the button is also generating a non-symbolSearch filter ie - `text`.

Hence the `toggleSymbolSearch` either needs to drop type checks of [`ToggleSymbolSearchOpts`](https://github.com/ruturajv/debugger.html/blob/bug-3038-editor-search-add-text-search-button/src/components/Editor/SearchBar.js#L81) or we need to split the [`toggleSymbolSearch`](https://github.com/ruturajv/debugger.html/blob/bug-3038-editor-search-add-text-search-button/src/components/Editor/SearchBar.js#L296) - what would you suggest ?

codehag:
Hi Ruturajv! Nice work, its starting to look cleaner.

I think that your impulse with checking for "text" as a type was a good one actually. Also, I agree that searchBtn was originally intended for SymbolSearches. I think the changes you made are good in that they are cleaner, but they also hide that there is a third potential value for searchType, which is "text". It would be better to introduce a new flow type, rather than removing an existing one. How about this as a solution: Lets expand the definition of searchType like so:

```js
// lines 81-84

type TextSearchType = "text";

type ToggleSymbolSearchOpts = {
  toggle: boolean,
  searchType: SymbolSearchType | TextSearchType
};
```

then we can do the check for `if (searchType === "text") ..` as you had it before, and this will be clearer and easier to understand for the next person who reads the code.

Further, with your case statement, we can do the same, adding a case for "text".

src/components/Editor/SearchBar.js
```diff
@@ -304,6 +304,11 @@ class SearchBar extends Component {
       e.stopPropagation();
     }
 
+    // Check if searchType is "text"
+    if (!["functions", "variables"].includes(searchType)) {
+      this.props.toggleSymbolSearch(false);
```
jasonLaster: hmm, this seems a bit out of date given what we're doing now. Maybe we want it to be: `setEditorSearchType` and that can include text

ruturajv:
@jasonLaster  hmm.. yaa seems lot might change here... even after @codehag  suggestions, these errors persist
```
$ yarn flow
yarn flow v0.24.5
$ flow 
src/components/Editor/SearchBar.js:326
326:         this.props.setSelectedSymbolType(searchType);
                                              ^^^^^^^^^^ string literal `text`. This type is incompatible with the expected param type of
115:     setSelectedSymbolType: SymbolSearchType => any,
                                ^^^^^^^^^^^^^^^^ string enum

src/components/Editor/SearchBar.js:334
334:       this.props.setSelectedSymbolType(searchType);
                                            ^^^^^^^^^^ string literal `text`. This type is incompatible with the expected param type of
115:     setSelectedSymbolType: SymbolSearchType => any,
                                ^^^^^^^^^^^^^^^^ string enum

```

codehag:
Hi @ruturajv! Thanks again for doing this. I went in and I wrestled with the code a bit, since upon working with it I realized it needed a refactor. Here is what I did:

- split the road: searching for text and searching for symbols are two different logical operations. One of the things that made toggleSymbolSearch so dang confusing was it was doing too many jobs. 

- remove extra if statements. We had a bunch of if statements in there that didn't make any sense. by splitting the logic, both functions are much easier to read and reason about.

- sticking to the defined type. The type we were working with was a `Symbol`. By respecting that type, I realized that -- hey, that should stay as it is! we want to do something else here. 

If anything is confusing, ping me! I am more than happy to explain further :)

ruturajv:
@codehag Thanks for the huge rewrite :)

Let me pull and go through it.

src/components/Editor/SearchBar.js
```diff
@@ -662,22 +675,45 @@ class SearchBar extends Component {
...
+          }),
+          onClick: e => toggleTextSearch(e)
+        },
+        searchType
+      );
+    }
+
     function searchTypeBtn(searchType) {
+      const active = isButtonActive(searchType);
+      const toggle = selectedSymbolType == searchType;
```
codehag: this should be identity equality `===`

src/components/Editor/index.js
```diff
@@ -894,8 +894,7 @@ Editor.contextTypes = {
 
 const expressionsSel = state => state.expressions.expressions;
 const getExpressionSel = createSelector(expressionsSel, expressions => input =>
-  expressions.find(exp => exp.input == input)
-);
+  expressions.find(exp => exp.input == input));
```
codehag: should be identity equality `===`

src/utils/parser/tests/fixtures/var.js
```diff
@@ -1,5 +1,4 @@
 var foo = 1;
 let bar = 2;
 const baz = 3;
-const a = 4,
-  b = 5;
```
codehag: style nit: this should be 
```
const a = ...;
const b = ...;
```

jasonLaster: this actually should be this way **becase** we're testing the weird behavior :) We want to make sure we find those vars too :P 

@ruturajv did this i believe

codehag: ohhh i understand

PR #3102, opened 2017-06-06
Use "react-dom" instead of "../../node_modules/react-dom/dist/react-dom"

codehag:
Looks like we had a misconfiguration in the .babelrc

I did this because the weird path with `../node_modules/react-dom/dist/react-dom` was doing weird things to editor component tests. I swapped it out. things seem to be working as expected.

### Summary of Changes

* remove line which was causing a recursive `react-dom/dist/react-dom/dist` address from babel-rc
* rename all instances of `../../../[etc]/node_modules/react-dom/dist/react-dom` to `react-dom`

### Test Plan

- All Unit tests are passing. Need to double check if the build still works

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -13,7 +13,6 @@
 
     ["module-resolver", {
       "alias": {
-        "react-dom": "react-dom/dist/react-dom",
         "devtools/client/shared/vendor/react": "react",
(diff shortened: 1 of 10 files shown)


codehag:
Also, just noticed something funny is up with our integration tests. Breakpoints tests and other tests now fail. Also on master. verrry weird

PR #3113, opened 2017-06-06
Revert breakpoint clearing

jasonLaster:
This fixes an issue we were seeing with breakpoints being added when they already existed on the server.

This fix is basically reverting the behavior we had before.

### DIFF
--- a/src/actions/sources.js
+++ b/src/actions/sources.js
@@ -23,6 +23,7 @@ import { prefs } from "../utils/prefs";
 import { removeDocument } from "../utils/editor";
 
 import {
+  getBreakpoint,
   getSource,
   getSourceByURL,
   getSourceText,
@@ -55,10 +56,12 @@ async function checkPendingBreakpoint(
   pendingBreakpoint,
   source
 ) {
-  const { sourceUrl } = pendingBreakpoint.location;
+  const { line, sourceUrl, column } = pendingBreakpoint.location;
   const sameSource = sourceUrl && sourceUrl === source.url;
+  const location = { sourceId: source.id, sourceUrl, line, column };
+  const bp = getBreakpoint(state, location);
 
-  if (sameSource) {
+  if (sameSource && !bp) {
     await dispatch(syncBreakpoint(source.id, pendingBreakpoint));
   }
 }
--- a/src/actions/sources.js
+++ b/src/actions/sources.js
@@ -23,7 +23,6 @@ import { prefs } from "../utils/prefs";
 import { removeDocument } from "../utils/editor";
 
 import {
-  getBreakpoint,
   getSource,
   getSourceByURL,
   getSourceText,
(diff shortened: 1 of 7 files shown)


src/actions/sources.js
```diff
@@ -55,10 +56,12 @@ async function checkPendingBreakpoint(
...
   source
 ) {
-  const { sourceUrl } = pendingBreakpoint.location;
+  const { line, sourceUrl, column } = pendingBreakpoint.location;
   const sameSource = sourceUrl && sourceUrl === source.url;
+  const location = { sourceId: source.id, sourceUrl, line, column };
+  const bp = getBreakpoint(state, location);
 
-  if (sameSource) {
+  if (sameSource && !bp) {
```
codehag: There are a couple of issues with this fix. 
- we are no longer syncing if the debuggee code is changed and the debuggee is reloaded
- we lose the reference to where the breakpoint is on the server, so if we update the debuggee, and the breakpoint moves, we can no longer delete it.

An alternative to get the debuggee breakpoints working again: do not add the breakpoint if the location is the same. the best solution would be if we had a function in the client to do this, but i investigated this last time and didn't find a way to do it. For now we could do the following. This will need some work to get it to be really performative, but it should work:

```diff
diff --git a/src/actions/breakpoints.js b/src/actions/breakpoints.js
index f525de5c..1e2e093b 100644
--- a/src/actions/breakpoints.js
+++ b/src/actions/breakpoints.js
@@ -62,9 +62,10 @@ async function syncClientBreakpoint(
     sourceId: generatedSourceId
   };
 
-  // early return if breakpoint is disabled with overrides to update
-  // the id as expected, without talking to server
-  if (pendingBreakpoint.disabled) {
+  const existingClient = client.getBreakpointByLocation(oldGeneratedLocation);
+  // early return if breakpoint is disabled, or exist on the server with
+  // overrides to update the id as expected
+  if (pendingBreakpoint.disabled || existingClient) {
     return {
       id: generatedSourceId,
       actualLocation: { ...pendingBreakpoint.location, id: sourceId },
diff --git a/src/client/firefox/commands.js b/src/client/firefox/commands.js
index f886401a..45308dbc 100644
--- a/src/client/firefox/commands.js
+++ b/src/client/firefox/commands.js
@@ -1,5 +1,6 @@
 // @flow
 
+import _ from "lodash";
 import type {
   BreakpointId,
   BreakpointResult,
@@ -74,6 +75,28 @@ function sourceContents(sourceId: SourceId): Source {
   return sourceClient.source();
 }
 
+function getBreakpointByLocation(location: Location) {
+  const values = _.values(bpClients);
+  console.log(values, location);
+  const bpClient = values.find(value => {
+    const { actor, line, column, condition } = value.location;
+    return (
+      location.line === line &&
+      location.sourceId == actor &&
+      location.column == column &&
+      location.condition == condition
+    );
+  });
+
+  if (bpClient) {
+    return {
+      id: bpClient.actor,
+      actualLocation: bpClient.location
+    };
+  }
+  return null;
+}
+
 function setBreakpoint(
   location: Location,
   condition: boolean,
@@ -251,6 +274,7 @@ const clientCommands = {
   stepOut,
   stepOver,
   breakOnNext,
+  getBreakpointByLocation,
   sourceContents,
   setBreakpoint,
   removeBreakpoint,

```

jasonLaster:
Thanks so much @codehag. This was really great work. I hope we can simplify the server in the future, but i will sleep better knowing you can tame the system too.

Issue #3072, opened 2017-06-02
Stepping in debugger gets very slow with complex scripts and deep stacks


vladikoff:
Ref: https://bugzilla.mozilla.org/show_bug.cgi?id=1239473

(This is from a series of bugs reported by Firefox teams, I'm making sure we file all of those. Please close if dup)

PR #3115, opened 2017-06-06
Get environment

jasonLaster:
Associated Issue: #3072

### Summary of Changes

This adds support for the new on demand frame scopes. It's important though that we don't remove the old system until the server has switched over.

* adds a firefox command for getting an environment
* adds a redux action and reducer for saving frame scopes
* updates the scopes component and util to support frame scopes

### Test Plan

* Add some action unit tests

### Next Steps

* Add the debugger client method
* Add the server endpoint

### DIFF
--- a/src/reducers/pause.js
+++ b/src/reducers/pause.js
@@ -12,6 +12,7 @@ type PauseState = {
   pause: ?any,
   isWaitingOnBreak: boolean,
   frames: ?(any[]),
+  frameScopes: any,
   selectedFrameId: ?string,
   loadedObjects: Object,
   shouldPauseOnExceptions: boolean,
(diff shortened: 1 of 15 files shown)


src/reducers/pause.js
```diff
@@ -75,7 +86,13 @@ function update(state: PauseState = State(), action: Action): PauseState {
       return Object.assign({}, state, { isWaitingOnBreak: true });
 
     case "SELECT_FRAME":
-      return Object.assign({}, state, { selectedFrameId: action.frame.id });
+      const { frame, scopes } = action;
+      const selectedFrameId = frame.id;
+      return {
+        ...state,
+        frameScopes: { ...state.frameScopes, [selectedFrameId]: scopes },
+        ...{ selectedFrameId }
```
codehag: probably this can just be selectedFrameId

src/utils/parser/tests/fixtures/var.js
```diff
@@ -1,4 +1,5 @@
 var foo = 1;
 let bar = 2;
 const baz = 3;
-const a = 4, b = 5;
+const a = 4,
+  b = 5;
```
codehag: `const b = 5;` would be easier to read

jasonLaster: agreed, it's prettier tho

Issue #3042, opened 2017-05-30
[Breakpoints] Remove column breakpoints in the gutter


jasonLaster:
We now show column breakpoints in the gutter when a breakpoints slides to a column. We should also support removing the breakpoint.

This is a bit trickier because we need to find the correct breakpoint and try and fire the remove breakpoint action with its location.

Here's a start: you'll see that when `columnBreakpoints` is off we ignore the column information

```diff
diff --git a/src/utils/editor/index.js b/src/utils/editor/index.js
index db370e7..e96756d 100644
--- a/src/utils/editor/index.js
+++ b/src/utils/editor/index.js
@@ -44,7 +44,11 @@ function isTextForSource(sourceText) {

 function breakpointAtLocation(breakpoints, { line, column = undefined }) {
   return breakpoints.find(bp => {
-    return bp.location.line === line + 1 && bp.location.column === column;
+    const sameColumn = isEnabled("columnBreakpoints")
+      ? bp.location.column === column
+      : true;
+    const sameLine = bp.location.line === line + 1;
+    return sameLine && sameColumn;
   });
```

The second part i'm less sure about: here we get the column location from the bp because when you click on the gutter we assume the column is undefined

```diff
diff --git a/src/components/Editor/index.js b/src/components/Editor/index.js
index d02ff48..984ddf9 100644
--- a/src/components/Editor/index.js
+++ b/src/components/Editor/index.js
@@ -556,7 +556,7 @@ class Editor extends PureComponent {
       removeBreakpoint({
         sourceId: sourceId,
         line: line + 1,
-        column: column
+        column: column || bp.getIn("location.column")
       });
     } else {
```

anbarasiu:
Can we assume if _column_ is undefined, it could be set to true, because when the click initiates from the gutter, it always seems so?

> const sameColumn = isEnabled("columnBreakpoints") && column
>       ? bp.location.column === column
>       : true;

jasonLaster:
Actually no, because in an evil case you can add a bp on a line with a
comment and it slides to another line and column. I'll get you an example
to test with :)

PR #3120, opened 2017-06-07
Remove column breakpoint

anbarasiu:
Associated Issue: #3042 

### Summary of Changes

* Removing the column breakpoints

### Test Plan

- Set a bp on an empty line/comment that slides to the next line.
- Clicking on the bp on the gutter removes it

### DIFF
--- a/src/components/Editor/index.js
+++ b/src/components/Editor/index.js
@@ -528,7 +528,7 @@ class Editor extends PureComponent {
       removeBreakpoint({
         sourceId: sourceId,
         line: line + 1,
-        column: column
+        column: column || bp.location.column
       });
     } else {
       addBreakpoint(
--- a/src/components/Editor/index.js
+++ b/src/components/Editor/index.js
@@ -525,10 +525,12 @@ class Editor extends PureComponent {
     const { sourceId } = selectedLocation;
 
     if (bp) {
+      // NOTE: it's possible the breakpoint has slid to a column
+      column = column || bp.location.column;
       removeBreakpoint({
         sourceId: sourceId,
         line: line + 1,
-        column: column || bp.location.column
+        column
       });
     } else {
       addBreakpoint(
--- a/src/utils/editor/index.js
+++ b/src/utils/editor/index.js
@@ -44,7 +44,11 @@ function isTextForSource(sourceText) {
 
 function breakpointAtLocation(breakpoints, { line, column = undefined }) {
   return breakpoints.find(bp => {
-    return bp.location.line === line + 1 && bp.location.column === column;
+    const sameColumn = isEnabled("columnBreakpoints") && column
+      ? bp.location.column === column
+      : true;
+    const sameLine = bp.location.line === line + 1;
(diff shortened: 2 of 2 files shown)


jasonLaster:
Here's another URL for testing:

http://localhost:8000/integration/examples/doc-sourcemaps2.html

jasonLaster:
Hey @anbarasiu! thanks so much for the help, this PR actually scared me for much of the day because sliding and bps are hard!

I made some small tweaks because i found the ternary boolean hard to read. I tried to expand it out a bit for my future self and others. 

I genuinely cannot wait for column breakpoints to land. They'll be so great!

anbarasiu:
Thanks for your help :) Looked at the changes and I'll keep the readability in mind for the next PR.

I've made a small change, so that clicked bp is returned.
if (!isEnabled("columnBreakpoints") _|| !column_) {
      return true;
}

remove-column-breakpoint

Was wondering if it would be better to remove the bp, not just when clicking on the gutter, but also the bp itself, when column > 0, like in the gif.

src/components/Editor/index.js
```diff
@@ -525,10 +525,12 @@ class Editor extends PureComponent {
     const { sourceId } = selectedLocation;
 
     if (bp) {
+      // NOTE: it's possible the breakpoint has slid to a column
+      column = column || bp.location.column;
```
codehag: I am wondering which one should take precedence here. If both `column` and `bp.location.column` are defined, which one would be more important? From the comment it sounds like `bp.location.column` -- because it should be overwriting to account for sliding? But I'm not too sure, sliding bps are hard!

src/utils/editor/index.js
```diff
@@ -44,7 +44,18 @@ function isTextForSource(sourceText) {
...
   return breakpoints.find(bp => {
-    return bp.location.line === line + 1 && bp.location.column === column;
+    const sameLine = bp.location.line === line + 1;
+    if (!sameLine) {
+      return false;
+    }
+
+    // NOTE: when column breakpoints are disabled we want to find
+    // the first breakpoint
+    if (!isEnabled("columnBreakpoints") || !column) {
```
codehag: :+1:

src/components/Editor/index.js
anbarasiu: Hmm.. I think column should take precedence, so that only when column info's not available we pick it up from bp. When initiated on the gutter, the column is undefined(from what I observed). Perhaps, when we directly click on the bp it could be passed(from codemirror?)?
Please correct me if I'm wrong here :)

codehag: Ah ok i see what is going on here. I think the comment should be more specific, maybe something like:

```
// falling back on the bp.location.column allows toggling the breakpoint a from the gutter, even if it is a column breakpoint
```

jasonLaster: ugh, this is an evil case. if we're clicking on a column breakpoint then we want to remove the breakpoint on that column. but we can cross that bridge when we get there

Issue #2963, opened 2017-05-19
Search shows 'no results' after switching to another tool and switching back to debugger


juliandescottes:
From https://bugzilla.mozilla.org/show_bug.cgi?id=1366220

STRs if running in firefox:
1. debug any file
2. CTRL/CMD+F
3. type something for which you find matches in the file
![image](https://cloud.githubusercontent.com/assets/1141550/26250270/c81932f4-3ca9-11e7-9bc6-3276ca57ff45.png)
4. switch to another tool
5. switch back to debugger

- ER: the search should still show the same status
- AR: the search shows "no results" 
![image](https://cloud.githubusercontent.com/assets/1141550/26250306/e5d3d966-3ca9-11e7-937f-25182e089cbe.png)

STRs if running in content:
Instead of steps 4&5, resize the window to be small enough to switch to vertical layout, this triggers the same issue.

jasonLaster:
hmm - i'm not seeing it at the moment:

![](http://g.recordit.co/QNiIQ1dcVq.gif)

PR #3121, opened 2017-06-07
Bug/2963 search shows no results after switching to another tool and switching back to debugger

codehag:
Associated Issue: #2963

### DIFF
--- a/src/components/Editor/SearchBar.js
+++ b/src/components/Editor/SearchBar.js
@@ -36,7 +36,7 @@ import { SourceEditor } from "devtools-source-editor";
 import type { SourceRecord } from "../../reducers/sources";
 import type { FileSearchModifiers, SymbolSearchType } from "../../reducers/ui";
 import type { SelectSourceOptions } from "../../actions/sources";
-import type { SearchResults } from ".";
+import type { SearchResults } from "../../reducers/ui";
 import type { SymbolDeclarations } from "../../utils/parser/getSymbols";
 import type { Location as BabelLocation } from "babel-traverse";
 import _SearchInput from "../shared/SearchInput";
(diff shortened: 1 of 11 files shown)


src/components/Editor/SearchBar.js
```diff
@@ -697,8 +701,9 @@ class SearchBar extends Component {
   }
 
   renderResults() {
-    const { symbolSearchResults, selectedResultIndex } = this.state;
-    const { query, symbolSearchOn } = this.props;
+    const { selectedResultIndex } = this.state;
+    const { query, symbolSearchResults, symbolSearchOn } = this.props;
+    console.log("hi");
```
jasonLaster: 👋

src/reducers/tests/ui.js
```diff
@@ -28,4 +28,16 @@ describe("ui reducer", () => {
...
+  it("gets the search results", () => {
+    const state = State();
+    expect(getSearchResults({ ui: state })).toBe(state.searchResults);
+  });
+
+  it("gets the symbol search results", () => {
+    const state = State();
+    expect(getSymbolSearchResults({ ui: state })).toBe(
+      state.symbolSearchResults
+    );
```
jasonLaster: I like treating actions as the public interface here:

here's an actions/tests/ui test
```js
  it("should toggle the visible state of project search", () => {
    expect(getProjectSearchState(getState())).toBe(false);
    dispatch(actions.toggleProjectSearch());
    expect(getProjectSearchState(getState())).toBe(true);
  });
```

src/components/Editor/SearchBar.js
codehag: ack

src/reducers/tests/ui.js
codehag: I guess false / true should be replaced with something else? or maybe we drop these tests for now?

jasonLaster: oh, i think this is a good "model" test

Issue #2901, opened 2017-05-15
Outline View


jasonLaster:
This issue tracks our progress towards being able to ship an Outline View in the left sidebar.

There are lots of small tidy up tickets here that anyone can grab

#### Checklist

- [x] [OutlineView] Navigation #2563
- [x] [Parser] get function params #2735
- [ ] adding an "Outline" label in the header
- [x] Switching the footer toggle button to tabs 
- [x] Replace blue bullet with something better
- [ ] ship it

#### Switching the footer toggle button to tabs 

We currently have a toggle button, but it would be nice to show both of them side by side like what we do in the editor footer for pretty printing and black boxing. We can borrow the same aria tooltip strategy from there too!

#### Replace blue bullet with something better

So... the outline items could look better. Lets tweak the hover behavior and style. I'm thinking cursor pointer, slight hover background perhaps. Maybe borrow from function search results or the call stack. Also of course the look and feel of the functions :)

#### GIF
![toggle](https://cloud.githubusercontent.com/assets/7821757/26278262/8321bf9e-3db3-11e7-913b-fde6a07841c3.gif)

rohanprasad:
@jasonLaster Is this done ?

jasonLaster:
@rohanprasad yeah, this was V1 and we can consider it done

PR #3128, opened 2017-06-08
Style functions in outline view

amelzer:
Associated Issue: #2901 

### Summary of Changes

* Removed bullet points and decreased left padding for functions in outline view
* Added pointer and hover effect to functions in outline view

### Test Plan

- [x] Hover over functions in outline view changes background according to theme

outlinefunctions

### DIFF
--- a/src/components/Outline.css
+++ b/src/components/Outline.css
@@ -1,11 +1,12 @@
 .outline-list {
   list-style-type: none;
-  padding-left: 10px;
+  padding-left: 0px;
+  width: 100%;
 }
 
 .outline-list__element {
   color: blue;
-  padding-left: 0.5rem;
+  padding-left: 1rem;
   padding-right: 0.5rem;
   cursor: pointer;
 }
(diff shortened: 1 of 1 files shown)


jasonLaster:
This looks great @amelzer!

My one style critique is that it would be nice to have the hover background be full width. Other than that I think it's ready!

amelzer:
I disagree. A full width hover would still require a little margin to the right which would require to implement a little handler for function names exceeding the width of the panel (either overflowing the container or shortening the function name). I would just leave it as it is, to be honest.Jason Laster <notifications@github.com> hat am 8. Juni 2017 um 15:03 geschrieben:This looks great @amelzer!My one style critique is that it would be nice to have the hover background be full width. Other than that I think it's ready!--You are receiving this because you were mentioned.Reply to this email directly or view it on GitHub:https://github.com/devtools-html/debugger.html/pull/3128#issuecomment-307096616

codehag:
Also, it looks like our `ul` element is set by default to have a margin of 40. I wonder if we want to do that, or if we should normalize it 🤔

amelzer:
@codehag The 40px ul padding seems to be browser default. 

Thank you for your other suggestions, it does look quite nice now.
outlinefunctionsv2

codehag:
Looks great! Thanks for your hard work!

PR #3131, opened 2017-06-08
replace commonjs with imports

jasonLaster:
Associated Issue: #1884

### DIFF
--- a/src/utils/prefs.js
+++ b/src/utils/prefs.js
@@ -1,9 +1,11 @@
 // @flow
 
-var { PrefsHelper } = require("devtools-modules");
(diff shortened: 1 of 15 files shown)


src/utils/prefs.js
```diff
@@ -1,10 +1,11 @@
 // @flow
 
-var { PrefsHelper } = require("devtools-modules");
-const { Services: { pref } } = require("devtools-modules");
-const { isDevelopment } = require("devtools-config");
+import { Services, PrefsHelper } from "devtools-modules";
```
codehag: we can do `{ Servioces: { pref }, PrefsHelper }` here i think

Issue #2978, opened 2017-05-21
Try green keeper lock file 


clarkbw:
Very similar to what we currently use, this should update lock files for us. 

https://github.com/greenkeeperio/greenkeeper-lockfile

PR #3134, opened 2017-06-08
Switch to use greenkeeper-lockfile-update for yarn.lock file

clarkbw:
Associated Issue: #2978 

### Summary of Changes

* remove `bin/update-yarn-lock-file`
* add [greenkeeper-lockfile-update](https://github.com/greenkeeperio/greenkeeper-lockfile) to handle updating the lock file.

### Test Plan

This is one of those PRs where the test plan is merging it and seeing that it works as expected with future greenkeeper PRs.

But currently I believe we're seeing a number of errors happening with CircleCI and multiple attempts to update the yarn.lock file.  Example: https://github.com/devtools-html/debugger.html/commits/greenkeeper/jest-localstorage-mock-1.2.0


jasonLaster:
Thanks! SHould i merge?

Issue #3006, opened 2017-05-25
[EditorSearch] Two search fields are displayed on the new debugger (windows)


jasonLaster:
**(WINDOWS ONLY)**

[Steps to reproduce]:
1. Go to https://devtools-html.github.io/debugger-examples/examples/todomvc/
2. Open Debugger.
3. From "Sources" section select a file that has the ".js" termination (e.g todo-view.js).
4. Press CTRL + F
   Search for a term that can be found (e.g listenTo).
5. Click on any code line.
6. Press Press CTRL + F

[Expected result]:
Pressing (On Windows - Press CTRL + F OR on Mac - Press CMD + F) should highlight or close the search field.

[Actual result]:
Two search fields are displayed on the new debugger

[Video](https://bug1366800.bmoattachments.org/attachment.cgi?id=8870059)

jasonLaster:
This is no longer happening

PR #3135, opened 2017-06-08
fix duplicate search bars

jasonLaster:
Associated Issue: #3006

### Summary of Changes

We disable codemirrors default search bar so that we can use our own in the panel. This previously worked for the mac, but not for windows. This should now be fixed, but it would be nice if a windows user could verify.

STR:

1. open a file in the debugger
2. ctrl+f should see the search bar
3. ctrl+f should see another search bar

### DIFF
--- a/src/utils/editor/index.js
+++ b/src/utils/editor/index.js
@@ -94,6 +94,7 @@ function createEditor() {
       // Override code mirror keymap to avoid conflicts with split console.
       Esc: false,
       "Cmd-F": false,
+      "Ctrl-F": false,
       "Cmd-G": false
     }
   });


zaggy:
I'll check that later today

codehag:
do we have any windows users who could take a look? @zaggy did you have a chance?

zaggy:
Sorry. Just managed to finally successfully build firefox yesterday. Will check this today.

zaggy:
An interesting thing. Everything works as expected. Even without this PR. I've checked it on this branch and on the master. Second search bar does not appear in both cases

jasonLaster:
Cool. I'll close this without merging. Thanks @zaggy & @janodvarko

Issue #2634, opened 2017-04-12
[WTR] run tests on CI


jasonLaster:
We used to run similar unit tests w/ firefox on CI previously. It's not to too hard to spin up firefox to do this...

codehag:
closing for now as we removed the WTR

PR #3138, opened 2017-06-08
Add WTR Runner

jasonLaster:
Associated Issue: #2634

### Summary of Changes

This re-introduces the work done here: https://github.com/devtools-html/debugger.html/pull/2852 to have a wtr headless runner. The test runs fail, but it's still good to have in the codebase so that we can get the tests passing and then run it in CI.

NOTE: you need to start chrome headlessly to run the tests
canary --headless --remote-debugging-port=9222 https://chromium.org

### DIFF
--- a/bin/wtr-runner.js
+++ b/bin/wtr-runner.js
@@ -1,0 +1,33 @@
+const CDP = require("chrome-remote-interface");
+
+const results = [];
+
+function runTests() {
+  return new Promise(resolve => {
+    CDP(client => {
+      // Extract used DevTools domains.
+      const { Page, Runtime } = client;
+
+      // Enable events on domains we are interested in.
+      Promise.all([Page.enable()]).then(() => {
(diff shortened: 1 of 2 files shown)


bin/wtr-runner.js
```diff
@@ -0,0 +1,33 @@
...
+
+      // Enable events on domains we are interested in.
+      Promise.all([Page.enable()]).then(() => {
+        return Page.navigate({ url: "http://localhost:8000/integration" });
+      });
+
+      Runtime.enable();
+      Runtime.consoleAPICalled(({ type, args }) => {
+        const argVals = args.map(arg => arg.value).join(", ");
+        console.log(argVals);
```
codehag: do we need this console log? If so should it have a descriptive text?

jasonLaster: I think it's nice to have it for now. It's easy to take out and tweak later

Issue #3152, opened 2017-06-11
Search bar doesn't get focus


amitzur:
Upon pressing cmd+f (or ctrl+f), if search bar is already open then the input will not be focused.

PR #3153, opened 2017-06-11
Search bar doesn't get focus if already open

amitzur:
Associated Issue: #3152

### Summary of Changes

* the method `toggleSearch` was called, which in turn calls `selectSearchInput` which selects the entire range in the input. But it doesn't focus.
I just added focus() inside `selectSearchInput`.

### DIFF
--- a/src/components/Editor/SearchBar.js
+++ b/src/components/Editor/SearchBar.js
@@ -345,6 +345,7 @@ class SearchBar extends Component {
     const searchInput = this.searchInput();
     if (searchInput) {
       searchInput.setSelectionRange(0, searchInput.value.length);
+      searchInput.focus();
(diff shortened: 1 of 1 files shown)


codehag:
Hi! Thanks for the bug report and the PR.

I tried to replicate the bug behavior you described, but I wasn't able to. Can you give me a scenario to follow to confirm the fix? Just the series of steps that you took to discover the bug. These are the steps I followed:

- open the debugger
- press cmd-f and start typing a valid search string
- unfocus the search field
- press cmd-f again and start typing

On both master and this branch this works, so I am probably missing something. Thanks!

jasonLaster:
Thanks @amitzur!

PR #3154, opened 2017-06-12
 #2980 added ember detection

andreicristianpetcu:
Associated Issue: #2980

### Summary of Changes

* added ember frame detection

### Test Plan

Put a breakpoint in ember.js and look at the stack frame icon. It should be Ember.

ember_light
ember_dark


codehag:
Hi! Thanks for the great work!

I took a look, and I find the logo to be a bit too small. I don't recognize it as ember. I am guessing that this because the other logos are restricted by this width, and changing that width would change all of the other logos.

I don't have a good solution for this yet, for a logo like this. Maybe its something we can address again a bit later with styling? Let me know if you have any ideas about how to make this a bit more visible.

codehag:
I got in touch with the Ember team to see if a square variant exists. Will get back to you!

andreicristianpetcu:
@codehag I initially wanted to add the ember mascot (The Tomster) but I [don't think I can do that legally](https://emberjs.com/tomster/faq/) and that's why I went with the [Ember logo](https://emberjs.com/logos/).

I think we should revisit the styling after we land all of them. [My personal issue](https://github.com/devtools-html/debugger.html/issues/2980#issuecomment-307108647) is with dark theme and the fact that the color of the logo [gets overridden](https://github.com/devtools-html/debugger.html/blob/615311c0d65361c54135cb15b9ffa43c0cb1b370/src/components/SecondaryPanes/Frames/Frames.css#L106) but only in dark mode.

codehag:
@andreicristianpetcu Cool! Yeah the tomster is a very nice logo, I wonder what their preference will be in this scenario. Lets wait for a response from them. If they don't have a suggestion as to a better logo to use in this scenario, we can go ahead with the official one and just merge this as it is for now, and figure out the styling later :) 

If the answer is yes, we can swap it out, and won't have to remember to take care of this later! plus it make things look nicer.

andreicristianpetcu:
It sounds good but I did not ask them about this.

codehag:
No problem :) I already sent an email with a link to this PR. Should hear back soon!

andreicristianpetcu:
Sorry, I did not read your [other message](https://github.com/devtools-html/debugger.html/pull/3154#issuecomment-307715471) :D

codehag:
They got back to me and it was inconclusive, so merging!

Issue #3141, opened 2017-06-08
Update outdated modules


clarkbw:
With #3134 we changed our `yarn.lock` updater.  I believe we've been seeing a number of issues with greenkeeper because of the old code.  I'd like to review and update some of our deps in a new PR to get us running on the stable version of modules we might have missed with recent greenkeeper PRs.

Working on:

- [x] firefox-profile removed in #3142
- [x] prettier updated to 1.4.4 in  #3142
- [x] percy-io/react-percy-storybook updated to 0.1.11 in #3142
- [x] flow-bin  #3012 lots of errors to deal with here
- [ ] babel-types not sure about this, currently seems broken #3140
- [ ] babel-traverse not sure about this, currently seems broken #3143
- [ ] svg-inline #2653 is difficult because so much has changed, it went esnext on us and doesn't ship a transpiled version which breaks at least the tests https://github.com/sairion/svg-inline-react/compare/v1.0.3...master

Ignoring for now
- [x] react related modules are to stay at their current versions
- [x] webpack is going to be a long process to get to 2.0 #3004
- [x] workerjs is a GH dep so we can ignore it

```
$ yarn outdated
yarn outdated v0.24.6
Package                         Current Wanted Latest Package Type    URL                        
@percy-io/react-percy-storybook 0.1.10  0.1.11 0.1.11 dependencies                               
babel-traverse                  6.24.1  6.25.0 6.25.0 dependencies                               
babel-types                     6.24.1  6.25.0 6.25.0 dependencies                               
firefox-profile                 0.4.9   0.4.9  1.0.0  dependencies                               
flow-bin                        0.46.0  0.46.0 0.47.0 dependencies                               
jest-localstorage-mock          1.1.1   1.2.0  1.2.0  devDependencies                            
prettier                        1.4.2   1.4.4  1.4.4  dependencies                               
react                           15.3.2  15.3.2 15.5.4 dependencies                               
react-addons-perf               15.3.2  15.3.2 15.4.2 devDependencies                            
react-addons-test-utils         15.3.2  15.3.2 15.5.1 devDependencies                            
react-dom                       15.3.2  15.3.2 15.5.4 dependencies                               
svg-inline-react                1.0.3   1.0.3  2.1.0  dependencies                               
webpack                         1.14.0  1.14.0 2.6.1  dependencies                               
workerjs                        0.1.2   exotic exotic dependencies    github:jasonLaster/workerjs
```

clarkbw:
[...]

PR #3155, opened 2017-06-12
Bump flow bin version

zaggy:
Associated Issue: #3141 #3012 

Here's the Pull Request Doc

### Summary of Changes
* Updated flow-bin to the version 0.47
* Fixed all errors throughout the code.

While I patched the code to throw no errors, it contains so many `any`s that it becomes useless to use flow at all. I think that another activity is required to go through all `any` references and figure out exact types there should be.

### Test Plan

No test are requires. Unit tests on CI should fail, but this is due to the flow issues in devtools-launchpad.


codehag:
Should we be testing node modules? That feels a bit off and like we could create headaches for ourselves, maybe we should change the config @jasonLaster 

For now lets wait until the other change gets merged in :) otherwise other prs branching from this one will be confused why the flow checker is complaining

zaggy:
I have already made a PR in devtools-html/devtools-core#439 .

Issue #2587, opened 2017-04-07
[FunctionSearch] Switch to a  modal


jasonLaster:
Many of our peers (vscode, atom, chrome) use a modal for searching. The modal is useful for function, variable (symbol), and command search (command palette), it is also used for searching for files.

I think we should adopt this convention as well.

Steps:

1. create a modal (Command Palette) that is fixed to the top of the debugger
2. Use our autocomplete and results list UI in the modal
3. have our search for function and variables button launch the modal

<img src='https://shipusercontent.com/cbe5d33b9fce29f73f191ac5108f80a7/command%20palette.jpg' title='command palette.jpg' width=1422>

jasonLaster:
@wldcordeiro is working on it

wldcordeiro:
Bumping. I've got a local branch with a lot of work done. I've just been busy starting a new job. Should PR soon. 😄

jasonLaster:
closing, i think this is in a great spot!

PR #3161, opened 2017-06-13
Symbol Search Modal

wldcordeiro:
Associated Issue: #2587 

### Summary of Changes

* rework the symbol search UI into a modal separate from file search
* simplify active search state (which is shown, project, file or symbol)

### Test Plan

* Open symbol modal (<kbd>ctrl</kbd>/<kbd>cmd</kbd>+<kbd>shift</kbd>+<kbd>o</kbd>
* Use `@` + query for functions
* Use `#` + query for variables
* Confirm that results match and traversing results travels properly.


src/components/Editor/SymbolModal.js
```diff
@@ -0,0 +1,291 @@
...
+
+    const symbolSearchResults = filter(symbols[selectedSymbolType], query, {
+      key: "value"
+    });
+
+    updateSearchResults({ count: symbolSearchResults.length });
+    updateSymbolSearchResults(symbolSearchResults);
+  }
+
+  traverseResults(rev: boolean) {
```
codehag: This isn't totally necessary to change, but I found the boolean for signifying direction is a bit confusing... if I read `traversResults(false)` I expect that we are turning result traversal off. I would potentially use modulus instead to simplify things and just pass `1` or `-1`.

ie)
```js
traverseResults(direction: number) {
  const { symbolSearchResults, selectedResultIndex } = this.state;
  const searchResults = symbolSearchResults;
  const resultCount = searchResults.length;

  const updatedSelectedResult = selectedResult + direction;
  // use modulus to find the absolute fraction
  const nextResultIndex = ((number + max) % max);
  this.setState({ selectedResultIndex: nextResultIndex });
  this.onSelectResultItem(searchResults[nextResultIndex]);
}

// used like so

traverseResults(1)
traverseResults(-1)
```


If I am not mistaken, this is the second occurrence of traverse results? Should we make this a helper function that is imported?

src/components/Editor/SymbolModal.js
```diff
@@ -0,0 +1,291 @@
...
+      }
+      this.setState({ selectedResultIndex: nextResultIndex });
+      this.onSelectResultItem(searchResults[nextResultIndex]);
+    }
+  }
+
+  onKeyDown(e: SyntheticKeyboardEvent) {
+    const { enabled, query } = this.props;
+    const { symbolSearchResults } = this.state;
+    if (!enabled || query == "") {
```
codehag: we can use identity equals here `===` since `"" == 0 // true`

or if we don't care about type conversion we can just `(!enabled || !query)`

src/components/Editor/SymbolModal.js
```diff
@@ -0,0 +1,291 @@
...
+    } else if (e.key === "Tab") {
+      this.closeSearch(e);
+      e.preventDefault();
+    }
+  }
+
+  renderResults() {
+    const { symbolSearchResults, selectedResultIndex } = this.state;
+    const { query, enabled } = this.props;
+    if (query == "" || !enabled || !symbolSearchResults.length) {
```
codehag: identity equality here might be useful as well, or, if not important, we can `!query`

src/components/Editor/SymbolModal.js
```diff
@@ -0,0 +1,291 @@
...
+    return dom.div(
+      { className: "symbol-modal" },
+      SearchInput({
+        query,
+        count: 0,
+        placeholder: this.buildPlaceHolder(),
+        summaryMsg: this.buildSummaryMsg(),
+        onChange: this.onChange,
+        onKeyUp: this.onKeyUp,
+        handleNext: e => this.traverseResults(e, false),
```
codehag: `traverseResults` takes only one argument I think?

src/components/Editor/Tabs.js
```diff
@@ -96,17 +97,18 @@ class SourceTabs extends PureComponent {
...
     selectedSource: SourceRecord,
-    selectSource: (string, ?Object) => void,
-    moveTab: (string, number) => void,
-    closeTab: string => void,
-    closeTabs: (List<string>) => void,
-    toggleProjectSearch: () => void,
-    togglePrettyPrint: string => void,
-    togglePaneCollapse: () => void,
-    showSource: string => void,
+    selectSource: (string, ?Object) => any,
```
codehag: this probably needs a rebase

src/components/Editor/SymbolModal.js
```diff
@@ -0,0 +1,291 @@
...
+      e.preventDefault();
+    } else if (e.key === "Enter") {
+      if (searchResults.length) {
+        this.selectResultItem(e, searchResults[this.state.selectedResultIndex]);
+      }
+      this.closeSearch(e);
+      e.preventDefault();
+    } else if (e.key === "Tab") {
+      this.closeSearch(e);
+      e.preventDefault();
```
codehag: we use `e.preventDefault` for each of the cases. we can instead put it above this block

jasonLaster:
some merge conflicts in SearchBar

src/components/Editor/SymbolModal.js
wldcordeiro: Good call @codehag I'm trying to clean up the API as I move stuff around (like the active search state thing.) I'm hacking tonight so I'll incorporate this. 😄

src/components/Editor/SymbolModal.js
wldcordeiro: 👍

jasonLaster:
making progress

PR #3163, opened 2017-06-13
switch bpclient ids to location ids

jasonLaster:
Associated Issue: #3182

### Summary of Changes

* I ran into an issue debugging [devtools-reporter](https://github.com/jasonLaster/devtools-reporter). The issue was that on reload the BP id had changed, but the source id and location were the same... Which got me thinking we should be consistent and use location ids
* I also set about adding some integration tests and cleaning them up. I'll do a mochitest for this as well before it lands

### DIFF
--- a/src/client/firefox/commands.js
+++ b/src/client/firefox/commands.js
@@ -11,7 +11,7 @@ import type {
   Script,
   Source,
   SourceId
-} from "../types";
+} from "debugger-html";
 
 import type {
   TabTarget,
@@ -80,17 +80,6 @@ function sourceContents(sourceId: SourceId): Source {
 function getBreakpointByLocation(location: Location) {
   const id = makeLocationId(location);
   const bpClient = bpClients[id];
-
-  // const values = _.values(bpClients);
-  // const bpClient = values.find(value => {
-  //   const { actor, line, column, condition } = value.location;
-  //   return (
-  //     location.line === line &&
-  //     location.sourceId === actor &&
-  //     location.column === column &&
-  //     location.condition === condition
-  //   );
-  // });
 
   if (bpClient) {
     const { actor, url, line, column, condition } = bpClient.location;
--- a/src/client/firefox/commands.js
+++ b/src/client/firefox/commands.js
@@ -4,9 +4,9 @@ import _ from "lodash";
 import type {
   BreakpointId,
   BreakpointResult,
+  Breakpoint,
   Frame,
   FrameId,
-  ActorId,
   Location,
   Script,
   Source,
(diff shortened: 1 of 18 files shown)


src/client/firefox/commands.js
```diff
@@ -77,16 +78,19 @@ function sourceContents(sourceId: SourceId): Source {
...
-      location.line === line &&
-      location.sourceId === actor &&
-      location.column === column &&
-      location.condition === condition
-    );
-  });
+  const id = makeLocationId(location);
+  const bpClient = bpClients[id];
+
+  // const values = _.values(bpClients);
```
codehag: maybe we can just delete this?

src/client/firefox/commands.js
```diff
@@ -77,16 +78,19 @@ function sourceContents(sourceId: SourceId): Source {
...
-  const bpClient = values.find(value => {
-    const { actor, line, column, condition } = value.location;
-    return (
-      location.line === line &&
-      location.sourceId === actor &&
-      location.column === column &&
-      location.condition === condition
-    );
-  });
+  const id = makeLocationId(location);
```
codehag: :+1: looks much better!

src/client/firefox/commands.js
```diff
@@ -118,40 +122,23 @@ function setBreakpoint(
       condition,
       noSliding
     })
-    .then((res: BreakpointResponse) => onNewBreakpoint(location, res));
+    .then(([{ actualLocation }, bpClient]) => {
+      actualLocation = createBreakpointLocation(location, actualLocation);
```
codehag: This isn't wrong in terms of logic, but I find it difficult to follow if args are reassigned in code. Generally, I would avoid reassignment due to how it impacts code reading. In this case, it's ok because we don't have `if` statements, but as a preventative measure, I think it would be better to use a new variable here.  In the lower case where we are reassigning `bpClients[i]`, it's unavoidable, so that can stay as it is.

src/client/firefox/create.js
```diff
@@ -53,3 +55,23 @@ export function createPause(
     frames: response.frames.map(createFrame)
   });
 }
+
+// Firefox only returns `actualLocation` if it actually changed,
+// but we want it always to exist. Format `actualLocation` if it
+// exists, otherwise use `location`.
+
+export function createBreakpointLocation(
```
codehag: :+1: great

src/client/firefox/types.js
```diff
@@ -362,6 +362,8 @@ export type BreakpointClient = {
   source: SourceClient
 };
 
+export type BPClients = { [id: ActorId]: BreakpointClient };
```
codehag: :+1: great

Issue #3123, opened 2017-06-07
[HighlightOutOfScopeLines] out of scope lines are not being cleared on resume


jasonLaster:
We grey out the out of scope lines when we pause. We currently update the out of scope lines when you step, but leave them there when you resume. We should obviously clear them out :)

This should do it: 

```diff
diff --git a/src/reducers/ast.js b/src/reducers/ast.js
index 5702bf1..ad9b2db 100644
--- a/src/reducers/ast.js
+++ b/src/reducers/ast.js
@@ -50,6 +50,10 @@ function update(
       return state.set("selection", null);
     }

+    case "RESUMED": {
+      return state.set("outOfScopeLocations", null);
+    }
+
     case "SET_SELECTION": {
       if (action.status == "start") {
         return state.set("selection", {
```

#### screenshot

<img src='https://shipusercontent.com/171121729fed9cabde3421006c0e162a/Screen%20Shot%202017-06-07%20at%2012.12.41%20PM.png' title='Screen Shot 2017-06-07 at 12.12.41 PM.png' width=658>

jasonLaster:
Fixed!

Issue #3180, opened 2017-06-15
[OutOfScopeLocations] are too slow


jasonLaster:
This perf shows how slow out of scope locations are. We could likely speed it up by moving it to its own component and using codemirror operations.

https://perf-html.io/public/67ad83cd3e94679bf6e08d98ccc91d1fe0e71b79/timeline/?hidePlatformDetails&range=27.5295_32.8559&thread=4

jasonLaster:
Fixed

PR #3184, opened 2017-06-18
Speed up Out Of Scope lines

jasonLaster:
Associated Issue: #3180

### Summary of Changes

This refactors out of scope lines a bit to support getting a list of in scope lines that can be used by the editor to check if a token is in scope on hover.

### Test Plan

Jest tests

### DIFF
--- a/src/actions/ast.js
+++ b/src/actions/ast.js
@@ -2,6 +2,7 @@
 
 import {
   getSourceText,
+  getSource,
   hasSymbols,
   getSelectedLocation,
   getSelectedSourceText,
@@ -40,9 +41,13 @@ export function setSymbols(source: Source) {
 export function setOutOfScopeLocations() {
   return async ({ dispatch, getState }: ThunkArgs) => {
     const location = getSelectedLocation(getState());
-    const sourceText = getSourceText(getState(), location.sourceId);
+    if (!location) {
+      return;
+    }
 
-    if (!location.line || !sourceText) {
+    const source = getSource(getState(), location.sourceId);
+
+    if (!location.line || !source) {
       return dispatch({
         type: "OUT_OF_SCOPE_LOCATIONS",
         locations: null
@@ -50,7 +55,7 @@ export function setOutOfScopeLocations() {
     }
 
     const locations = await parser.getOutOfScopeLocations(
-      sourceText.toJS(),
+      source.toJS(),
       location
     );
 
--- a/src/actions/tests/ast.js
+++ b/src/actions/tests/ast.js
@@ -7,7 +7,7 @@ import {
 
 import readFixture from "./helpers/readFixture";
 const { getSymbols, getOutOfScopeLocations } = selectors;
-import getLinesInScope from "../../selectors/linesInScope";
+import getInScopeLines from "../../selectors/linesInScope";
 
 const threadClient = {
   sourceContents: function(sourceId) {
@@ -82,7 +82,7 @@ describe("ast", () => {
       await dispatch(actions.selectSource("scopes.js", { line: 5 }));
 
       const locations = getOutOfScopeLocations(getState());
-      const lines = getLinesInScope(getState());
+      const lines = getInScopeLines(getState());
 
       expect(locations).toMatchSnapshot();
       expect(lines).toMatchSnapshot();
@@ -95,7 +95,7 @@ describe("ast", () => {
       await dispatch(actions.selectSource("base.js"));
 
       const locations = getOutOfScopeLocations(getState());
-      const lines = getLinesInScope(getState());
+      const lines = getInScopeLines(getState());
 
       expect(locations).toEqual(null);
       expect(lines).toEqual([1]);
--- a/src/actions/tests/ast.js
+++ b/src/actions/tests/ast.js
@@ -5,7 +5,9 @@ import {
   makeSource
 } from "../../utils/test-head";
 
+import readFixture from "./helpers/readFixture";
 const { getSymbols, getOutOfScopeLocations } = selectors;
+import getLinesInScope from "../../selectors/linesInScope";
 
 const threadClient = {
   sourceContents: function(sourceId) {
@@ -30,7 +32,8 @@ const threadClient = {
 
 const sourceTexts = {
   "base.js": "function base(boo) {}",
-  "foo.js": "function base(boo) { return this.bazz; } outOfScope"
+  "foo.js": "function base(boo) { return this.bazz; } outOfScope",
+  "scopes.js": readFixture("scopes.js")
 };
 
 const evaluationResult = {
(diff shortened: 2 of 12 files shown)


src/components/Editor/index.js
```diff
@@ -155,8 +154,10 @@ class Editor extends PureComponent {
       this.showSourceText(selectedSource, selectedLocation);
     }
 
-    if (this.props.outOfScopeLocations !== nextProps.outOfScopeLocations) {
-      clearLineClass(this.editor.codeMirror, "out-of-scope");
+    if (this.props.linesInScope !== nextProps.linesInScope) {
+      this.editor.codeMirror.operation(() => {
```
jasonLaster: `this.editor.codeMirror.operation` is a huge perf win. With that said, i'll be moving the operation inside `clearLineClass` soon :)

src/components/Editor/index.js
```diff
@@ -788,6 +795,23 @@ class Editor extends PureComponent {
     });
   }
 
+  renderInScopeLines() {
+    const { linesInScope } = this.props;
+    if (
+      !isEnabled("highlightScopeLines") ||
```
jasonLaster: i'm re-adding a feature flag because i'm not happy with the cases we're missing like blocks and comments. Seeing those issues makes me think we should wait and re-launch highlighting when it's further along. 

With that said, we still get the onMouseOver logic so fewer accidental popups

src/selectors/linesInScope.js
```diff
@@ -0,0 +1,42 @@
+// import { linesInScope } from "../utils/scopes";
+import { getOutOfScopeLocations, getSelectedSource } from "../selectors";
```
jasonLaster: this is the first selector that spans stores (sources, ast) with that in mind I created a new selectors directory. I'm not sure if that makes sense. What do others think?

src/actions/ast.js
```diff
@@ -1,33 +1,34 @@
...
-    if (hasSymbols(getState(), source)) {
+    const sourceRecord = getSource(getState(), sourceId);
+    if (!sourceRecord) {
       return;
     }
 
-    const sourceText = getSourceText(getState(), source.id);
-    if (!sourceText) {
+    const source = sourceRecord.toJS();
+    if (!source.text || hasSymbols(getState(), source)) {
```
jasonLaster: some small `sourceText => source` migrations

src/actions/tests/ast.js
```diff
@@ -72,32 +75,33 @@ describe("ast", () => {
...
   describe("getOutOfScopeLocations", () => {
-    it("simple", async () => {
+    it("with selected line", async () => {
       const { dispatch, getState } = createStore(threadClient);
-      const base = makeSource("base.js");
-      await dispatch(actions.newSource(base));
-      await dispatch(actions.loadSourceText({ id: "base.js" }));
-      await dispatch(actions.selectSource("base.js", { line: 1 }));
-
-      await dispatch(actions.setOutOfScopeLocations());
```
jasonLaster: yeah - so this is coooool. Turns out many of the dispatches are unnecessary.

src/components/Editor/Editor.css
```diff
@@ -22,8 +22,12 @@
...
 }
 
-.out-of-scope .CodeMirror-line, .out-of-scope .CodeMirror-linenumber {
-  opacity: 0.8;
+.paused .in-scope .CodeMirror-line, .paused .in-scope .CodeMirror-linenumber {
+  opacity: 1;
+}
+
+.paused .CodeMirror-line, .paused .CodeMirror-linenumber {
+  opacity: 0.7;
```
jasonLaster: we're now tracking in-scope and not out-of-scope. 

some reasons:
1. it's easier to get the lines in scope than all the lines that are out of scope
2. it should be faster because there are fewer innies than outies.

src/reducers/ast.js
```diff
@@ -79,6 +79,10 @@ function update(
       });
     }
 
+    case "RESUMED": {
+      return state.set("outOfScopeLocations", null);
+    }
+
```
jasonLaster: this fixes #3123

src/selectors/linesInScope.js
```diff
@@ -0,0 +1,41 @@
...
+
+import range from "lodash/range";
```
codehag: would it make sense to do this?

```javascript
import { range, flatMap, uniq, without } from "lodash";
```

src/selectors/linesInScope.js
```diff
@@ -0,0 +1,41 @@
...
+import flatMap from "lodash/flatMap";
+import uniq from "lodash/uniq";
+import without from "lodash/without";
+
+function getLinesOutOfScope(outOfScopeLocations: AstLocation[]) {
+  if (!outOfScopeLocations) {
```
codehag: if this is an array of AstLocations, I think this will always be true... or maybe it isnt an array?

if it is an array we probably need to do 
```javascript
if (!outOfScopeLocations.length) {
  ...
}
```

src/selectors/linesInScope.js
```diff
@@ -0,0 +1,41 @@
+import { getOutOfScopeLocations, getSelectedSource } from "../selectors";
+
+import range from "lodash/range";
+import flatMap from "lodash/flatMap";
+import uniq from "lodash/uniq";
+import without from "lodash/without";
+
+function getLinesOutOfScope(outOfScopeLocations: AstLocation[]) {
```
codehag: `getLinesOutOfScope` sounds like we are taking lines out of a scope, but I think this is getting the out of scope lines. maybe `getOutOfScopeLines` will be clearer here,

src/selectors/linesInScope.js
```diff
@@ -0,0 +1,41 @@
...
+  if (!outOfScopeLocations) {
+    return null;
+  }
+
+  return uniq(
+    flatMap(outOfScopeLocations, location =>
+      range(location.start.line, location.end.line)
```
codehag: nice

src/selectors/linesInScope.js
```diff
@@ -0,0 +1,41 @@
...
+    )
+  );
+}
+
+export default function getLinesInScope(state: OuterState) {
```
codehag: same comment here about naming: proposing `getInScopeLines`

src/selectors/linesInScope.js
```diff
@@ -0,0 +1,41 @@
+import { getOutOfScopeLocations, getSelectedSource } from "../selectors";
+
+import range from "lodash/range";
+import flatMap from "lodash/flatMap";
+import uniq from "lodash/uniq";
+import without from "lodash/without";
+
+function getLinesOutOfScope(outOfScopeLocations: AstLocation[]) {
+  if (!outOfScopeLocations) {
+    return null;
```
codehag: this is returning a different type; maybe it should be an empty array?

src/selectors/linesInScope.js
```diff
@@ -0,0 +1,41 @@
...
+
+  const linesOutOfScope = getLinesOutOfScope(
+    outOfScopeLocations,
+    source.toJS()
+  );
+
+  const sourceNumLines = source.get("text").split("\n").length;
+  const sourceLines = range(1, sourceNumLines + 1);
+
+  if (!linesOutOfScope) {
```
codehag: the logic here would be clearer if this was checking for an empty array

src/actions/tests/ast.js
codehag: awesome

src/selectors/linesInScope.js
jasonLaster: sadly... no because then we pull lodash into the bundle... I suppose if we made it an external then we could exclude it from the bundle and expect it to be available in m-c... If we like this i can file a separate issue for it.

src/selectors/linesInScope.js
jasonLaster: agreed :)

src/selectors/linesInScope.js
jasonLaster: I can do both. when it is not available it's nully.

src/selectors/linesInScope.js
jasonLaster: ... not sure. i kinda like having explicit empty states to distinguish between nothing and we happen to have no out of scope locs. ill think about it

src/selectors/linesInScope.js
codehag: :+1:

PR #3187, opened 2017-06-18
prettify css

jasonLaster:
prettier now supports CSS. Lets use it!


src/components/SecondaryPanes/Frames/Group.css
```diff
@@ -1,5 +1,4 @@
-.frames ul .frames-group .group,
-.frames ul .frames-group .group .location {
+.frames ul .frames-group .group, .frames ul .frames-group .group .location {
```
codehag: Nice! this is a good improvement! but wow that's a complicated selector

PR #3188, opened 2017-06-19
Update getting-setup.md

tommai78101:
Proposing a small change to the Firefox Windows command, based on #1248, where using `--start-debugging-server` works while `-start-debugger-server` fails.


jasonLaster:
forwarding to bryan and others

PR #3190, opened 2017-06-19
(WIP) Replace sourcetext w/ source

jasonLaster:
Associated Issue: #3078

### Summary of Changes

* cleans up actions/ast
* a couple more places to cleanup

### DIFF
--- a/flow-typed/debugger-html.js
+++ b/flow-typed/debugger-html.js
@@ -206,12 +206,13 @@ declare module "debugger-html" {
    */
   declare type Source = {
     id: SourceId,
-    url?: string,
-    sourceMapURL?: string,
+    url: string,
+    sourceMapURL: string,
     isBlackBoxed: boolean,
     isPrettyPrinted: boolean,
     text?: string,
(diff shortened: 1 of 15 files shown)


flow-typed/debugger-html.js
```diff
@@ -206,12 +206,13 @@ declare module "debugger-html" {
    */
   declare type Source = {
     id: SourceId,
-    url?: string,
```
codehag: is it possible that a source will not have a url? for example, a `javascript:<name>` script tag?

flow-typed/debugger-html.js
```diff
@@ -206,12 +206,13 @@ declare module "debugger-html" {
...
-    sourceMapURL?: string,
```
codehag: will an original source have a sourceMapUrl?

jasonLaster: yup: https://github.com/devtools-html/debugger.html/blob/master/src/client/firefox/create.js#L34

PR #3195, opened 2017-06-21
highlight pause expressions

jasonLaster:
### Summary of Changes

This updates our debug line highlighter so that it starts at the column we're paused at. This is more informative because we're often not paused at the beginning of a line.

### Test Plan

I'd like to update `editor-highlight`

### DIFF
--- a/src/components/Editor/Editor.css
+++ b/src/components/Editor/Editor.css
@@ -1,12 +1,12 @@
 .editor-wrapper {
-  --debug-line-background: rgba(226, 236, 247, 0.5);
   --debug-line-border: rgb(145, 188, 219);
+  --debug-expression-background: rgba(202, 227, 255, 0.5);
   --editor-searchbar-height: 27px;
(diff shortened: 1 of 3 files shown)


src/components/Editor/Editor.css
```diff
@@ -160,8 +160,12 @@ html[dir="rtl"] .editor-mount {
...
 }
 
+.debug-expression {
+  background-color: var(--debug-expression-background);
+}
+
 .new-debug-line .CodeMirror-line {
-  background-color: var(--debug-line-background) !important;
+  background-color: transparent !important;
   outline: var(--debug-line-border) solid 1px;
```
wldcordeiro: Maybe move this to the `.debug-expression`? I think it'd look a little nicer rather than the whole line having the outline.

jasonLaster: Hmm - happy to try. I kinda like framing the line and highlighting the relevant section, but I bet there's more we can do too

Issue #2668, opened 2017-04-17
[Editor] Right clicking on the gutter should not add a breakpoint


jasonLaster:
right clicking on the gutter should open the context menu with the menuitem to "Add a breakpoint".

It looks like we're currently adding the breakpoint pre-maturely :)

The issue is likely [here]

[here]: https://github.com/devtools-html/debugger.html/blob/master/src/components/Editor/index.js#L403-L423

sarahlim:
I just cloned and I'm having trouble replicating this:

![gif of left clicking to add a breakpoint, right clicking to open context menu w/o adding breakpoint](https://media.giphy.com/media/3o7bu2sehHjuq2l8nm/giphy.gif)

Left clicking produces a breakpoint as expected. Right clicking opens the context menu, no breakpoint appears in the UI, and Redux devtools don't log an `ADD_BREAKPOINT` action.

I'm new, so I might be doing something wrong!

jasonLaster:
Haha. That's awesome. I'll try and find the edge case :)

jasonLaster:
Also, how'd you get redux Dev tools?

sarahlim:
Main repo [here](https://github.com/gaearon/redux-devtools), browser extensions [here](https://github.com/zalmoxisus/redux-devtools-extension)!

jasonLaster:
Thanks @sarahlim I'm using the redux devtools and love it now!

I investigated the bug and it looks like it only happens on chrome :)

![](http://g.recordit.co/tpHV5JQ0ad.gif)

sarahlim:
> it only happens on chrome

Bizarre, my GIF was recorded in Chrome (definitely), 57.0.2987.133 (I think).

I'll check again when I'm not supposed to be studying for midterms!

jasonLaster:
Haha. I'll also investigate some more.

jasonLaster:
Hey @sarahlim just checking in. Do you have any time to investigate?

jasonLaster:
Ping @sarahlim do you have time to look into this? If not no worries

sarahlim:
Hi @jasonLaster -- so sorry for the radio silence. I'm a bit swamped right now, but if this is still happening after school is out for the year, happy to give it another go then!

jasonLaster:
That's alright. It's not a priority. Talk to you soon :)

sarahlim:
It's working for me on Chrome 58.0.3029 as well, both regular and incognito mode. What version is causing the error for you?

PR #3196, opened 2017-06-21
Bugfix: ctrl-clicking gutter prematurely added breakpoint

sarahlim:
Associated Issue: #2668 

### Summary of Changes

Added a check in `onGutterClick` to early return if the user right clicked using <kbd>ctrl</kbd>-click.

### Test Plan

- [x] Right-clicking in the gutter **by holding down <kbd>ctrl</kbd> and left-clicking once** only shows the context menu, does not add a breakpoint

#### before

breakpoint is added

#### after

breakpoint is no longer added

### DIFF
--- a/src/components/Editor/index.js
+++ b/src/components/Editor/index.js
@@ -435,6 +435,7 @@ class Editor extends PureComponent {
 
     // ignore right clicks in the gutter
     if (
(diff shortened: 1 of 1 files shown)


jasonLaster:
Thanks, great job hunting it down! A+ for persistence :)

Issue #3068, opened 2017-06-02
Search function in debugger is case sensitive by default


vladikoff:
The default behaviour seems to unlike other text editors.  

![image](https://cloud.githubusercontent.com/assets/128755/26746136/012a49c6-47bc-11e7-9c78-27ffd83e4eee.png)

(This is from a series of bugs reported by Firefox teams, I'm making sure we file all of those. Please close if dup)

jasonLaster:
Hmm. We can switch that. We also recently backed it by prefs so whatever you choose will be saved going forward

jasonLaster:
I did a bit of research and case insensitive is the more common default.

ohana54:
I'm on it

PR #3199, opened 2017-06-21
Change default case sensitive search modifier to off

ohana54:
Associated Issue: #3068 

### Summary of Changes

* Change the default case sensitive search modifier to be turned off, meaning the search is case-insensitive by default.

 #goodnessSquad


ohana54:
Was just about to write - it works in dev, how do I test the "production" preferences?
I tried disabling the development overrides, but the debugger failed to load as a result.

jasonLaster:
it's hard to test in firefox, but that's why we have development.

Thanks!

Issue #3055, opened 2017-06-01
[HighlightScopeLines] should include comments 


jasonLaster:
When the debugger pauses, we highlight the lines of code that are in scope by looking at the functions in the file. We currently skip the comments, which makes them seem like they're in scope when they're not. We should also mark comments as out of scope as well.

We can do this by updating our `getOutOfScopeLocations` function to look for comments as well and include them in the list of out of scope locations.

[AST]:https://astexplorer.net/#/gist/17873678ad2d07c8ff3106e4693c0b2c/eede9a60846f127dcfd1d72408c856e0ba7372be

<img src='https://shipusercontent.com/46adda9858e23d2185de6fa107baabc1/Screen%20Shot%202017-06-01%20at%209.49.36%20AM.png' title='Screen Shot 2017-06-01 at 9.49.36 AM.png' width=241>

After getting acquanted with the code, you can run `jest src` and take a look at this [test](https://github.com/devtools-html/debugger.html/blob/master/src/utils/parser/tests/getOutOfScopeLocations.js#L1). You'll want to be able to run the test and see a new snapshot with the comment locations included. 

```diff 
diff --git a/src/utils/parser/getOutOfScopeLocations.js b/src/utils/parser/getOutOfScopeLocations.js
index fb879ad..dc71a8f 100644
--- a/src/utils/parser/getOutOfScopeLocations.js
+++ b/src/utils/parser/getOutOfScopeLocations.js
@@ -16,9 +16,9 @@ import { traverseAst } from "./utils/ast";
  * Returns all functions (declarations, expressions, arrows) for the given
  * source
  */
-function findFunctions(source: SourceText) {
+function getSymbols(source: SourceText) {
   const fns = [];
-  traverseAst(source, {
+  const ast = traverseAst(source, {
     enter(path) {
       if (isFunction(path)) {
         fns.push(path);
@@ -26,7 +26,7 @@ function findFunctions(source: SourceText) {
     }
   });

-  return fns;
+  return { function: fns, comments: ast.comments };
 }

 /**
@@ -91,11 +91,16 @@ function getOutOfScopeLocations(
   source: SourceText,
   position: AstPosition
 ): AstLocation[] {
-  return findFunctions(source)
+  const { functions, comments } = getSymbols(source);
+  const funcLocations = functions
     .map(getLocation)
     .filter(loc => !containsPosition(loc, position))
     .reduce(removeOverlaps, [])
     .sort(sortByStart);
     
[...]

PR #3200, opened 2017-06-21
Include comments in "out of scope" locations

amitzur:
Associated Issue: #3055

### Summary of Changes

* in getOutOfScopeLocations.js changed function findFunctions to findSymbols and concat functions to comments

### Test Plan

- [x] should treat comments as out of scope

the only reason this PR is under my account is that @giladartzi's machine didn't compile the code properly. Beautiful work in #goodnessSquad

### DIFF
--- a/src/utils/parser/getOutOfScopeLocations.js
+++ b/src/utils/parser/getOutOfScopeLocations.js
@@ -9,9 +9,9 @@ import { containsLocation, containsPosition } from "./utils/helpers";
 
 import getSymbols from "./getSymbols";
 
-function findFunctions(source) {
-  const symbols = getSymbols(source);
-  return symbols.functions;
+function findSymbols(source) {
+  const { functions } = getSymbols(source);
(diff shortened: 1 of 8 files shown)


src/utils/parser/getOutOfScopeLocations.js
```diff
@@ -16,17 +16,17 @@ import { traverseAst } from "./utils/ast";
  * Returns all functions (declarations, expressions, arrows) for the given
  * source
  */
-function findFunctions(source: SourceText) {
+function findSymbols(source: SourceText) {
```
wldcordeiro: This `findSymbols/findFunctions` function seems odd to me. Couldn't we reuse this [getSymbols](https://github.com/devtools-html/debugger.html/blob/master/src/utils/parser/getSymbols.js#L55) function and modify it to extract comments?

codehag:
updated and rebased

Issue #3209, opened 2017-06-23
[Preview] Speedup large scripts


jasonLaster:
Our preview logic relies on traversing the AST and finding the closest expression to the user's mouse. For instance, if the user hovers on the `a` in `this.foo.bar` we will find the node `bar` and resolve the full expression. The problem is that for large files 10K lines, this operation can naively take 2 seconds.

codehag:
merged!

PR #3210, opened 2017-06-23
bail early on closestExpression

jasonLaster:
Associated Issue: #3209 

This offers much needed perf wins when debugging large 10K+ files (a.k.a. the debugger.js file). Previously the perf times were a couple seconds for finding the closest expression and out of scope lines and now they're a couple hundred milliseconds or less...

It's amazing how much it pays to be lazy.

I got these rough stats from adding logging to our worker utils
```js
 37         var t0 = performance.now();
 38         this.worker.postMessage({ id, method, args });
 39
 40         const listener = ({ data: result }) => {
 41           if (result.id !== id) {
 42             return;
 43           }
 44
 45           this.worker.removeEventListener("message", listener);
 46           var t1 = performance.now();
 47           console.log(`${method} took:` + (t1 - t0) + " milliseconds.");
 48
```

### DIFF
--- a/src/utils/parser/getOutOfScopeLocations.js
+++ b/src/utils/parser/getOutOfScopeLocations.js
@@ -21,7 +21,7 @@ function findFunctions(source) {
  */
 
 function getLocation(func) {
-  let location = Object.assign({}, func.location);
+  const location = { ...func.location };
 
   // if the function has an identifier, start the block after it so the
   // identifier is included in the "scope" of its parent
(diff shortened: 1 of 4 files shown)


src/utils/parser/getOutOfScopeLocations.js
```diff
@@ -5,47 +5,33 @@ import type { AstLocation, AstPosition } from "./types";
...
 /**
  * Returns the location for a given function path. If the path represents a
  * function declaration, the location will begin after the function identifier
  * but before the function parameters.
  */
-const getLocation = path => {
-  let location = Object.assign({}, get("node.loc", path));
+
+function getLocation(func) {
+  let location = Object.assign({}, func.location);
```
codehag: maybe this should be 
`const location = { ...func.location };`

codehag:
i will fix up the comment and the tests

PR #3220, opened 2017-06-23
Fix reloading issue with selected tabs + breakpoints

jasonLaster:
Fixes: #3221

We were seeing two issues:

1. tabs were not selected on reload
2. breakpoints were jumping (we saw this same kinda thing earlier) the root cause is when we try to add the same bp again

how is everything fixed? I changed the action from navigate to connect which now just sets the debuggee url and does not clear any state. I also added lots of tests!

### DIFF
--- a/src/reducers/sources.js
+++ b/src/reducers/sources.js
@@ -40,18 +40,20 @@ export type SourcesState = {
   tabs: TabList
 };
 
-export const State = makeRecord(
-  ({
-    sources: I.Map(),
-    selectedLocation: undefined,
-    pendingSelectedLocation: prefs.pendingSelectedLocation,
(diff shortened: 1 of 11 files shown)


src/reducers/sources.js
```diff
@@ -139,8 +141,12 @@ function update(
     case "NAVIGATE":
       const source = getSelectedSource({ sources: state });
       const url = source && source.get("url");
-      prefs.pendingSelectedLocation = { url };
-      return State().set("pendingSelectedLocation", { url });
+
+      if (!url) {
+        return initialState();
+      }
```
jasonLaster: this is a race condition where we don't have a selected source, but we had one prior... It's a small fix that came from some fast tests.

Issue #3109, opened 2017-06-06
[Preview] Highlight the entire expression on hover


jasonLaster:
We currently find the full member expression when you hover on a token. It would be neat if we could highlight the full expression as well.

codeMirror has a good api for getting the tokens on a line. once we have the tokens we can probabably figure out how which elements to add classes to.

PR #3224, opened 2017-06-24
get bracketarrow working again

aklt:
Associated Issue: #3109 

### Summary of Changes

* Revive the small arrow pointing to the variable that the preview popup shows 

### Test Plan

Start debugging a file and mouse over a variable while stepping through code.  The poup
has a small arrow pointing to the variable.

### DIFF
--- a/src/components/shared/BracketArrow.js
+++ b/src/components/shared/BracketArrow.js
@@ -5,12 +5,17 @@ import classNames from "classnames";
 
 import "./BracketArrow.css";
 
-const BracketArrow = (
+const BracketArrow = ({
+  orientation,
+  left,
+  top,
+  bottom
+}: {
   orientation: string,
   left: number,
   top: number,
   bottom: number
-) => {
+}) => {
   return dom.div(
     {
       className: classNames("bracket-arrow", orientation || "up"),
--- a/src/components/shared/BracketArrow.css
+++ b/src/components/shared/BracketArrow.css
@@ -36,6 +36,6 @@
 
 .bracket-arrow.down::after {
   border-bottom-color: transparent;
-  border-top-color: var(--theme-toolbar-background);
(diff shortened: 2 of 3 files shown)


src/components/shared/BracketArrow.js
```diff
@@ -1,25 +1,31 @@
...
+
+    return dom.div(
+      {
+        className: classNames("bracket-arrow", orientation || "up"),
+        style: { left, top, bottom }
+      },
+      ""
+    );
+  }
+}
```
jasonLaster: oh interesting, why does this need to be a component over a function?

src/components/shared/BracketArrow.js
```diff
@@ -1,25 +1,31 @@
...
 
 BracketArrow.displayName = "BracketArrow";
 
+BracketArrow.propTypes = {
+  orientation: PropTypes.string,
+  left: PropTypes.number,
+  top: PropTypes.number,
+  bottom: PropTypes.number
+};
```
jasonLaster: We're now using props, which is flow based.

Checkout https://github.com/devtools-html/debugger.html/blob/master/src/components/Editor/Footer.js#L25-L35

aklt:
@jasonLaster you were right, no reason to use a class when a function suffices.

I tested in firefox 54.0 and changed the position of the arrow which was a bit off.

jasonLaster:
Great. Thanks for the fix

src/components/shared/BracketArrow.js
```diff
@@ -5,12 +5,17 @@ import classNames from "classnames";
...
+  left,
+  top,
+  bottom
+}: {
   orientation: string,
   left: number,
   top: number,
   bottom: number
-) => {
+}) => {
```
jasonLaster: 👍

Issue #3168, opened 2017-06-14
Breakpoints disappear when changing panels


codehag:
When switching between console and debugger, or networkpanel and debugger, the breakpoints in the gutter disappear. The breakpoints in the side menu do not.

Expected:
gutter breakpoints do not disappear.

jasonLaster:
This is the same problem as #3033.

Issue #3033, opened 2017-05-30
Scroll position and block scope expansion state in the right-hand sidebar is lost when stepping


gijsk:
STR:

1. stop on breakpoint (note: this bug is worse if you have 5-10 breakpoints on the go)
2. expand some of the scope blocks to see local variables
3. step over/into

ER:
expanded items stay expanded, and the scroll position remains the same

AR:
the sidebar scrolls to the top, meaning I can't see the local variables block anymore

This quickly becomes very frustrating if I'm attempting to look at variables over the course of execution.

jasonLaster:
Thanks @gijsk. keeping the expand state for variables is on our list. We had it in the debugger for the past couple months but found some cases where we were keeping stale data when we maintained the expand state. We hope to revisit this after launch in june.

jasonLaster:
I looked into this a bit today and found that it is caused by the Editor component unmounting when you select another tool

This patch below shows that we can:

a) track when the tool is active and only then switch modes
b) not move the tabs when the tool is de-activated ...

The code is ugly, so we'll want to clean it up

```diff
diff --git a/src/components/App.js b/src/components/App.js
index 8af09ff..727860c 100644
--- a/src/components/App.js
+++ b/src/components/App.js
@@ -84,8 +84,11 @@ class App extends Component {
     verticalLayoutBreakpoint.removeListener(this.onLayoutChange);
   }

-  onLayoutChange() {
-    this.setState({ horizontal: verticalLayoutBreakpoint.matches });
+  onLayoutChange(a, b, c) {
+    console.log("onLayoutChange");
+    if (window.isActive) {
+      this.setState({ horizontal: verticalLayoutBreakpoint.matches });
+    }
   }

   renderEditorPane() {
   
   diff --git a/src/components/Editor/Tabs.js b/src/components/Editor/Tabs.js
index 96f9271..3a5c9ca 100644
--- a/src/components/Editor/Tabs.js
+++ b/src/components/Editor/Tabs.js
@@ -135,7 +135,7 @@ class SourceTabs extends PureComponent {
     );

     this.onResize = debounce(() => {
-      this.updateHiddenSourceTabs();
+      // this.updateHiddenSourceTabs();
     });
   }

   diff --git a/src/utils/bootstrap.js b/src/utils/bootstrap.js
index 9231c5c..620bd00 100644
--- a/src/utils/bootstrap.js
+++ b/src/utils/bootstrap.js
[...]

PR #3225, opened 2017-06-24
Stop scrolling when we toggle tools

jasonLaster:
Associated Issue: #3168 #3033

### Summary of Changes

* the problem was that we were unmounting codemirror when we toggled tools. We fix that, by first checking if the app is visible when we might unmount the editor.

### Test Plan

* simple integration test for toggling tools

### DIFF
--- a/src/test/mochitest/browser_dbg-toggling-tools.js
+++ b/src/test/mochitest/browser_dbg-toggling-tools.js
@@ -1,0 +1,38 @@
+// Return a promise with a reference to jsterm, opening the split
+// console if necessary.  This cleans up the split console pref so
+// it won't pollute other tests.
+function getSplitConsole(dbg) {
+  const { toolbox, win } = dbg;
+
+  registerCleanupFunction(() => {
+    Services.prefs.clearUserPref("devtools.toolbox.splitconsoleEnabled");
+  });
+
+  if (!win) {
+    win = toolbox.win;
+  }
+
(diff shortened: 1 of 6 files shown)


src/test/mochitest/browser_dbg-toggling-tools.js
```diff
@@ -0,0 +1,38 @@
...
+add_task(function*() {
+  const dbg = yield initDebugger("doc-scripts.html");
+
+  yield selectSource(dbg, "long");
+  dbg.win.cm.scrollTo(0, 284);
+
+  pressKey(dbg, "inspector");
+  pressKey(dbg, "debugger");
+
+  is(dbg.win.cm.getScrollInfo().top, 284);
```
jasonLaster: we can always test more invariants like bps, search state... but I prefer a minimal test that correlates with the behavior. We should to more tests on the unit side

src/utils/ui.js
```diff
@@ -0,0 +1,9 @@
+/* Checks to see if the root element is available and
+ * if the element is visible. We check the width of the element
+ * because it is more reliable than either checking a focus state or
+ * the visibleState or hidden property.
+ */
+export function isVisible() {
```
bomsy: does `isVisible` feels vague, maybe something like `isAppVisible` is a bit clearer?

jasonLaster: yeah - that's fair

Issue #3170, opened 2017-06-14
[Breakpoints] index.html bps are empty


jasonLaster:
<img src='https://shipusercontent.com/abaac12a5f3d61b9c03ba5b807ed0046/Screen%20Shot%202017-06-14%20at%204.42.53%20AM.png' title='Screen Shot 2017-06-14 at 4.42.53 AM.png' width=310>

PR #3227, opened 2017-06-25
Fix empty index.html breakpoints

bomsy:
Associated Issue: #3170

### Summary of Changes

* use the `getFilename` utility function instead of the `basename`, for consistency with Tabs and the Sourcetree

### Test Plan

- [x] Go to the increment  sourcemaps example.
- [x] Select the `(index)` in the sourcetree
- [x] Add breakpoint in the file
- [x] Should see `(index)` in the breakpoints list

#### After

image


codehag:
Looks great!

PR #3229, opened 2017-06-26
[WIP] Symbols

jasonLaster:
### Summary of Changes

I want to find all of the expressions in a file, this will let us anticipate what a user might want to preview or where a breakpoint might be added.

```js

const obj = { a: { b: 2 } }; // e.g. obj.a.b
const foo = obj2.c.secondProperty; // e.g. foo.obj2.c.secondProperty

// computed properties
const com = { [a]: { b: "c" }, [b]: 3 }; // e.g. com[a].b

// assignments
obj.foo = { a: { b: "c" }, b: 3 }; // e.g. obj.foo.a.b
com = { a: { b: "c" }, b: 3 }; // e.g. com.a.b

// arrays
const res = [{ a: 2 }, { b: 3 }]; // e.g. res[1].b
const res2 = {a: [{b: 2}]}  // e.g. res.a[0].b
const res3 = [[{a:3}],[{b:3}]] // e.g. res[1][0].b

// destructuring
const { a, rest } = compute(stuff); // e.g. a
const [a, ...rest] = compute(stuff);

function params({ a, b }) {} // e.g. a
var pars = function({ a, b }) {};

const evil = obj2.doEvil().c.secondProperty; // e.g. obj2.doEvil or ""
```

### Test Plan

Lots of tests, lots and lots of them


src/utils/parser/getSymbols.js
```diff
@@ -84,9 +90,217 @@ export default function getSymbols(source: SourceText): SymbolDeclarations {
...
   });
 
   symbolDeclarations.set(source.id, symbols);
   return symbols;
 }
+
+function addProperty(name, expression, path, prevPath) {
+  const computed = path && path.node.computed;
+  const prevComputed = prevPath && prevPath.node.computed;
+  const prevArray = t.isArrayExpression(prevPath);
```
codehag: this is a bool, probably should be `hasPrevArray`

src/utils/parser/getSymbols.js
```diff
@@ -84,9 +90,217 @@ export default function getSymbols(source: SourceText): SymbolDeclarations {
...
+  const array = t.isArrayExpression(path);
```
codehag: same here

src/utils/parser/getSymbols.js
```diff
@@ -84,9 +90,217 @@ export default function getSymbols(source: SourceText): SymbolDeclarations {
...
+  const index = prevPath.parentPath.key;
+  expression = addProperty(index, expression, path, prevPath);
+
+  prevPath = path;
+  path = path.parentPath && path.parentPath.parentPath;
+
+  return getExpression(path, prevPath, expression);
+}
+
+function getExpression(path, prevPath, expression = "") {
```
codehag: this should probably be renamed to `getExpressionName` since it returns a string representation, and not the expression itself

src/utils/parser/getSymbols.js
```diff
@@ -84,9 +90,217 @@ export default function getSymbols(source: SourceText): SymbolDeclarations {
...
+  const array = t.isArrayExpression(path);
+
+  if (expression === "") {
+    if (computed) {
+      return `[${name}]`;
+    }
+    return name;
+  }
+
+  if (computed || array) {
```
codehag: i dont think we need this block, since it is doing the same thing as lines 160 - 164, and if it is computed -- we have an early return at line 150

src/utils/parser/getSymbols.js
```diff
@@ -84,9 +90,217 @@ export default function getSymbols(source: SourceText): SymbolDeclarations {
...
+
+  return getExpression(path, prevPath, expression);
+}
+
+function getArrayExpression(path, prevPath, expression) {
+  const index = prevPath.parentPath.key;
+  expression = addProperty(index, expression, path, prevPath);
+
+  prevPath = path;
+  path = path.parentPath && path.parentPath.parentPath;
```
codehag: same for lines 202 204, 205 -- avoid renaming arguments

src/utils/parser/getSymbols.js
```diff
@@ -84,9 +90,217 @@ export default function getSymbols(source: SourceText): SymbolDeclarations {
...
+
+function getExpression(path, prevPath, expression = "") {
+  if (t.isVariableDeclaration(path)) {
+    const node = path.node.declarations[0];
+    const name = node.id.name;
+    return addProperty(name, expression, path, prevPath);
+  }
+
+  if (t.isVariableDeclarator(path)) {
+    const node = path.node.id;
```
codehag: again, a bit confused about why id is an node? would expect it to be a string...

src/utils/parser/getSymbols.js
```diff
@@ -84,9 +90,217 @@ export default function getSymbols(source: SourceText): SymbolDeclarations {
...
+    if (t.isObjectPattern(node)) {
+      return expression;
+    }
+
+    const name = node.name;
+    const prop = addProperty(name, expression, path, prevPath);
```
codehag: we can directly return this

src/utils/parser/getSymbols.js
codehag: snippit might be better

PR #3240, opened 2017-06-28
(WIP) call sites

jasonLaster:
Opening an early PR, which we can collaborate on.


src/components/Editor/index.js
```diff
@@ -693,9 +702,9 @@ class Editor extends PureComponent {
           breakpoint: bp,
           editor: this.editor && this.editor.codeMirror
         })
-      );
+      );*/
```
jasonLaster: we should remove the column breakpoints code here and the component

src/components/Editor/index.js
```diff
@@ -792,6 +801,15 @@ class Editor extends PureComponent {
     );
   }
 
+  renderCallSites() {
+    const editor = this.editor;
```
jasonLaster: lets add the column bps feature flag here

src/components/Editor/index.js
```diff
@@ -295,12 +300,30 @@ class Editor extends PureComponent {
...
+  onKeyUp(e) {
+    let { key } = e;
+
+    if (key === "Alt") {
+      e.stopPropagation();
+      e.preventDefault();
+
+      this.setState({ showCallSites: false });
+    }
+  }
```
jasonLaster: lets move `onKeyUp` and `onKeyDown` to CallSites because we can :)

src/components/Editor/index.js
```diff
@@ -295,12 +300,30 @@ class Editor extends PureComponent {
     }
   }
 
+  onKeyUp(e) {
+    let { key } = e;
+
+    if (key === "Alt") {
+      e.stopPropagation();
```
jasonLaster: others might handle this too so perhaps we dont want to stop propogation

src/components/Editor/Editor.css
```diff
@@ -5,6 +5,10 @@
   --editor-second-searchbar-height: 27px;
 }
 
+.call-sites {
+  outline: 1px solid #aed3ef;
+}
```
jasonLaster: this should be in its own call-sites css file

src/components/Editor/CallSites.js
```diff
@@ -0,0 +1,168 @@
...
+}
+
+export default connect(
+  state => {
+    const selectedLocation = getSelectedLocation(state);
+    const selectedSource = getSelectedSource(state);
+    const sourceId = selectedLocation && selectedLocation.sourceId;
+    const source = selectedSource && selectedSource.toJS();
+
+    getCallSiteBreakpoint = location => {
```
jasonLaster: this should be in its own helper

src/components/Editor/CallSites.js
```diff
@@ -0,0 +1,168 @@
...
+    let sites;
+    if (!callSites) {
+      return null;
+    }
+
+    callSites = callSites.filter(
+      callSite => callSite.location.start.line === callSite.location.end.line
+    );
+
+    editor.codeMirror.operation(() => {
```
jasonLaster: i'm not sure if this helped... lets look at the perf here. Maybe not a blocker, but nice

src/utils/parser/getSymbols.js
```diff
@@ -115,6 +116,19 @@ export default function getSymbols(source: SourceText): SymbolDeclarations {
         });
       }
 
+      if (t.isCallExpression(path)) {
+        const callee = path.node.callee;
+        const calleeNode = t.isMemberExpression(callee)
```
jasonLaster: we should exclude member expressions cause... well you know

src/components/Editor/CallSites.js
bomsy: i tot it did when we tested it, at the allhands ... i'll try it again

src/components/Editor/CallSite.js
```diff
@@ -0,0 +1,97 @@
...
+      if (nextProps.showCallSite) {
+        if (!this.marker) {
+          this.addCallSite(nextProps);
+        }
+      } else {
+        if (!nextProps.breakpoint) {
+          this.clearCallSite();
+        }
+      }
+      // console.log(nextProps.showCallSite);
```
bomsy: remove console.log

src/utils/parser/tests/__snapshots__/getSymbols.js.snap
```diff
@@ -56,6 +56,30 @@ variables
...
 `;
 
+exports[`Parser.getSymbols call sites 1`] = `
+"properties
+
+member expressions
+[(2, 8), (2, 9)] [(2, 0), (2, 9)]  c 
+[(2, 4), (2, 5)] [(2, 0), (2, 5)]  b 
+call expressions
+   undefined
```
jasonLaster: it looks like we might be getting the member expressions and not the call expressions with this change?

bomsy: ah .. yep!

src/components/Editor/CallSites.js
```diff
@@ -0,0 +1,226 @@
...
+      sites = dom.div(
+        {},
+        callSites.map((callSite, index) => {
+          return CallSite({
+            key: index,
+            callSite,
+            editor,
+            breakpoint: callSite.breakpoint,
```
bomsy: we might not need to pull this in on its own since we have it on `callSite`

jasonLaster:
I made some progress on this today, but ultimately could not get it all the way there:

done:
1. styling in light/dark theme
2. bottom border uses `::before` elements so they don't overlay
3. callsites are associated with breakpoints

not done:
1. removing a bp fails because we can't find the callsite for a location...

bomsy:
looking into the failure on removing breakpoints might have an idea what the issue is ...

Issue #3106, opened 2017-06-06
[Pause] Stepping over await calls.


jasonLaster:
The debugger should step over await statements. When I have something in my debugger 

like this: If I'm on line 2. and I press step over, I don't care this call is asynchronous, I want to be on line 3. 

```js
1.    async foo() {
2.      await bar();
3.      return true;
4.   }
```

jimblandy:
Generators and async/await were not standard yet when Debugger was designed, so there are plenty of improvements we could make here. Some of it might be SpiderMonkey work, but stepping across an `await` seems more like setting an internal breakpoint on the instruction after the `await` and waiting for it to be hit. There's code in the `step` implementation now to handle returns and throws; it would need to be taught that `await` is something different.

I'm also not sure we have a good way to set that breakpoint accurately. That might require Debugger work.

jasonLaster:
github.com/devtools-html/debugger.html/issues/3106

jasonLaster:
Here are some ASTs, i think we can prototype a solution on the client and if we like it add something on the server:

SIMPLE: [ast]( https://astexplorer.net/#/gist/373291544b7824bb815aec303b1af9e4/a60cfdb455937247b3fe93efd45ac2c1c6dbfa9b)

it would be nice to also handle babel's code: [babel](https://babeljs.io/repl/#?babili=false&evaluate=true&lineWrap=false&presets=es2015%2Creact%2Cstage-2&targets=&browsers=&builtIns=false&debug=false&code=async%20function%20foo()%20%7B%0A%20%20await%20slow()%0A%20%20const%20bar%20%3D%203%3B%0A%7D)

jasonLaster:
Just landed @jbhoosreddy's amazing PR for async stepping. There are now a couple of small follow ups:

1. land a test for step over (might need to tweak the implementation)
2. add support for step in and step out
3. add unit tests for the new parser functions

codehag:
Closing and opening a new issue for tracking the remaining two improvements needed: improve step in / step out & and unit tests

PR #3242, opened 2017-06-29
[wip] Async step over

jbhoosreddy:
Associated Issue: #3106

### Summary of Changes

* kills SourceText everywhere

### Test Plan

- [x] Visual Testing

### DIFF
--- a/src/actions/ast.js
+++ b/src/actions/ast.js
@@ -12,6 +12,18 @@ import {
 import { PROMISE } from "../utils/redux/middleware/promise";
 import * as parser from "../utils/parser";
 
+import { getClosestPath } from "../utils/parser/utils/closest";
+import {
+  isAsyncFunction,
+  isAwaitExpression,
+  containsPosition
+} from "../utils/parser/utils/helpers";
+
+import { addBreakpoint, removeBreakpoint } from "./breakpoints";
+import { getHiddenBreakpoint } from "../reducers/breakpoints";
+import { command } from "./pause";
+import type { NodePath } from "babel-traverse";
+
 import type { SourceId } from "debugger-html";
 import type { ThunkArgs } from "./types";
 import type { AstLocation } from "../utils/parser";
@@ -124,5 +136,57 @@ export function setSelection(
         };
       })()
     });
+  };
+}
+
+/**
+ * Action dispatched from stepOver action. Receives the position
+ * and depending on where it is paused, sets up hidden breakpoints
+ * and handles resume, stepOver (as needed)
+ * @param position
+ * @returns {function(ThunkArgs)}
+ */
+export function analyzePauseLocation(position: AstLocation) {
+  return async ({ dispatch, getState, client }: ThunkArgs) => {
+    const source = getSelectedSource(getState()).toJS();
+    const hiddenBreakpointLocation = getHiddenBreakpoint(getState());
+
+    await dispatch({
+      type: "ANALYZE_PAUSE_LOCATION",
+      [PROMISE]: (async function() {
+        if (hiddenBreakpointLocation) {
+          await dispatch(removeBreakpoint(hiddenBreakpointLocation));
+        }
+        const path = getClosestPath(source, position);
+        if (!path) {
+          return;
+        }
+        if (isAwaitExpression(path)) {
+          return dispatch(analyzeAwaitExpression(path, position));
+        }
+        return dispatch(command({ type: "stepOver" }));
+      })()
+    });
+  };
+}
+
+function analyzeAwaitExpression(path: NodePath, position: AstLocation) {
+  return async ({ dispatch }: ThunkArgs) => {
+    const blockScope = path.scope.block;
+    if (blockScope && isAsyncFunction(blockScope)) {
+      const siblings = blockScope.body.body;
+      for (let i = 0; i != siblings.length; i++) {
+        const sibling = siblings[i];
+        if (containsPosition(sibling.loc, position)) {
+          const nextSibling = siblings[++i];
+          const nextLocation = nextSibling.loc.start;
+          nextLocation.sourceId = position.sourceId;
+          await dispatch(
+            addBreakpoint(nextLocation, { hidden: true, condition: "" })
+          );
+          return dispatch(command({ type: "resume" }));
+        }
+      }
+    }
   };
 }
--- a/src/actions/ast.js
+++ b/src/actions/ast.js
@@ -159,7 +159,7 @@ export function analyzePauseLocation(position: AstLocation) {
         }
         const path = getClosestPath(source, position);
         if (!path) {
-          return;
+          return dispatch(command({ type: "stepOver" }));
         }
         if (isAwaitExpression(path)) {
           return dispatch(analyzeAwaitExpression(path, position));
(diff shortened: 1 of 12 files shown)


src/actions/ast.js
```diff
@@ -126,3 +135,52 @@ export function setSelection(
     });
   };
 }
+
+let hiddenBreakpointLocation = null;
```
jasonLaster: we should try and avoid this. How can we use the redux store?

src/actions/ast.js
```diff
@@ -126,3 +135,52 @@ export function setSelection(
...
+ * @returns {function(ThunkArgs)}
+ */
+export function analyzePauseLocation(position: AstLocation) {
+  return async ({ dispatch, getState, client }: ThunkArgs) => {
+    const source = getSelectedSource(getState()).toJS();
+
+    await dispatch({
+      type: "ANALYZE_PAUSE_LOCATION",
+      [PROMISE]: (async function() {
+        if (hiddenBreakpointLocation) {
```
jasonLaster: I feel like we should break this up so that we have smaller functions for these different scenarios.

src/actions/ast.js
```diff
@@ -126,3 +138,64 @@ export function setSelection(
...
+ * @param position
+ * @returns {function(ThunkArgs)}
+ */
+export function analyzeStepping(type) {
+  return async ({ dispatch, getState, client }: ThunkArgs) => {
+    const position = await dispatch(getPosition());
+    const source = getSelectedSource(getState()).toJS();
+    const hiddenBreakpointLocation = getHiddenBreakpoint(getState());
+
+    await dispatch({
```
jbhoosreddy: I can probably pull this out of a `dispatch` call as it is not going to hit a reducer. most likely this is switching through reducers and doing a noop as there's no case for it. wonder if there's a better way to handle this?

src/actions/ast.js
```diff
@@ -126,3 +138,64 @@ export function setSelection(
...
+/**
+ * Action dispatched from stepOver action. Receives the position
+ * and depending on where it is paused, sets up hidden breakpoints
+ * and handles resume, stepOver (as needed)
+ * @param position
+ * @returns {function(ThunkArgs)}
+ */
+export function analyzeStepping(type) {
+  return async ({ dispatch, getState, client }: ThunkArgs) => {
+    const position = await dispatch(getPosition());
```
jasonLaster: if we don't intend to update the state, it shouldn't be an action. We can just as easily pass in `getState()`

src/actions/ast.js
```diff
@@ -126,3 +138,64 @@ export function setSelection(
...
+      type: "ANALYZE_STEPPING",
+      [PROMISE]: (async function() {
+        if (hiddenBreakpointLocation) {
+          await dispatch(removeBreakpoint(hiddenBreakpointLocation));
+        }
+        const path = getClosestPath(source, position);
+        if (!path) {
+          return dispatch(command({ type }));
+        }
+        switch (type) {
```
jasonLaster: I personally prefer the style `["stepOver","stepIn",].includes(type)` or `type == "stepOver" || type == "stepOver",`

src/actions/ast.js
```diff
@@ -126,3 +138,64 @@ export function setSelection(
...
+        return dispatch(command({ type }));
+      })()
+    });
+  };
+}
+
+function analyzeAwaitExpression(path: NodePath, position: AstLocation, type) {
+  return async ({ dispatch }: ThunkArgs) => {
+    const blockScope = path.scope.block;
+    if (blockScope && isAsyncFunction(blockScope)) {
```
jasonLaster: lets drop the nesting and make it a guard:

```js
if (!isAnAsyncFunction(path)) {
  return;
}
```

NOTE: we need a new helper :)

src/actions/ast.js
```diff
@@ -126,3 +138,64 @@ export function setSelection(
...
+      const siblings = blockScope.body.body;
```
jasonLaster: we should clean this up with a `findFunctionStatement` helper or something like that.

I'd opt for using lodash `find` or some other search function.

src/actions/ast.js
```diff
@@ -126,3 +138,64 @@ export function setSelection(
...
+          case "stepOut":
+            if (isAwaitExpression(path)) {
+              return dispatch(analyzeAwaitExpression(path, position, type));
+            }
+        }
+        return dispatch(command({ type }));
+      })()
+    });
+  };
+}
```
jasonLaster: big picture, i'm wondering if we can drop this analyze function in favor of some smaller utilities.

* **getPauseLocationNode(state)** knows how to get the pause location and call getClosest path. 
* **removeHiddenBreakpoint** knows how to check for a hidden bp and remove it on pause.

src/actions/ast.js
```diff
@@ -126,3 +138,64 @@ export function setSelection(
...
+    const source = getSelectedSource(getState()).toJS();
+    const hiddenBreakpointLocation = getHiddenBreakpoint(getState());
+
+    await dispatch({
+      type: "ANALYZE_STEPPING",
+      [PROMISE]: (async function() {
+        if (hiddenBreakpointLocation) {
+          await dispatch(removeBreakpoint(hiddenBreakpointLocation));
+        }
+        const path = getClosestPath(source, position);
```
jasonLaster: you're currently using `getClosestPath` directly by requiring it. Unfortunately, we need to do the search on the parser worker, so we'll have to move some of these methods to the worker.

here are some [docs](https://github.com/devtools-html/debugger.html/blob/master/docs/local-development.md#adding-a-task) on adding a new worker task.

I think we'll need to add tasks like `isAwaitExpression` or `insideAsyncFunction` to the worker. We could avoid this if those helpers could use a babel node. The distinction is that Nodes don't have prototypes. Paths do.

src/actions/ast.js
```diff
@@ -126,3 +138,64 @@ export function setSelection(
...
+    if (blockScope && isAsyncFunction(blockScope)) {
+      const siblings = blockScope.body.body;
+      for (let i = 0; i != siblings.length; i++) {
+        const sibling = siblings[i];
+        if (containsPosition(sibling.loc, position)) {
+          const nextSibling = siblings[++i];
+          const nextLocation = nextSibling.loc.start;
+          nextLocation.sourceId = position.sourceId;
+          await dispatch(
+            addBreakpoint(nextLocation, { hidden: true, condition: "" })
```
jasonLaster: minor thing, but i think it would be nice to have `addHiddenBreakpoint`

src/actions/ast.js
```diff
@@ -126,3 +138,64 @@ export function setSelection(
...
+              return dispatch(analyzeAwaitExpression(path, position, type));
+            }
+        }
+        return dispatch(command({ type }));
+      })()
+    });
+  };
+}
+
+function analyzeAwaitExpression(path: NodePath, position: AstLocation, type) {
```
jasonLaster: I think we could call this `asyncCommand` to be parallel with `command`

src/actions/ast.js
jbhoosreddy: @jasonLaster when I add `getClosestPath` on the worker it throws me a `DOMException` error. I'll try it again tonight and see if I fix it.

src/actions/ast.js
jbhoosreddy: we still have a single entry-point function right? or you mean to drop the outer function altogether?

src/actions/ast.js
jbhoosreddy: do you mean `awaitCommand` @jasonLaster?

jbhoosreddy:
stale. closing in favor of #3340

Issue #3239, opened 2017-06-28
[Preview] Parse more function symbol types


wldcordeiro:
Currently our symbol parsing doesn't understand some function types that might be used in the wild. It doesn't cover cases where we have `MethodDefinition`s in a class or an arrow function `ClassProperty`. I have a branch ~[here](https://github.com/wldcordeiro/debugger.html/tree/parse-more-symbols)~ that gets us nearly there with parsing these symbol types.

Or if you aren't a fan of checking out a branch here is a [patch](https://gist.github.com/wldcordeiro/1acf6b11f947291fdb47a94cffa23096)

firhatsungkar:
@wldcordeiro Is this issue still available ? Can I work on this issue?

jasonLaster:
It's yours @firhatsungkar

jasonLaster:
@firhatsungkar any luck?

jasonLaster:
@firhatsungkar making it available again. Feel free to re-claim

wldcordeiro:
I'm going to claim this. I want this for selfish reasons 😈

jasonLaster:
ping @wldcordeiro

wldcordeiro:
Been off this for a while with quick open modal stuff. I'll grab this now

PR #3246, opened 2017-06-29
Update Babylon to latest (6.17.4) & Enable more parsing plugins

wldcordeiro:
Associated Issue: #3239 (sorta)

### Summary of Changes

* I cherry-picked my first commit from that patch/branch so our parser doesn't fail with syntax errors on newer language features.


src/utils/parser/utils/ast.js
```diff
@@ -15,7 +15,20 @@ function _parse(code, opts) {
     code,
     Object.assign({}, opts, {
       sourceType: "module",
-      plugins: ["jsx", "flow"]
+      plugins: [
```
wldcordeiro: There's a plugin for "decorators" that I could enable but it's based on a rejected version of the language feature proposal, the rest here are not rejected.

PR #3255, opened 2017-06-30
Move highlighting debug line to child component

asolove:
A research project done somewhat in parallel with the debugger team's hack week activities.

### Motivation

The Editor component has extremely complex stateful behavior, including responsibility for maintaining lots of seemingly-separable concerns about the underlying CodeMirror editor's state. Carrying out these responsibilities is also the source of some slowness, as CodeMirror updates itself multiple times for each action, even though they could all be done as a single batch.

### What's here

- [X] Extracting management of the debugger line into a separate component

### Latter issues should address

- Lots more child components can get pulled out.
- Strategy for batching editor updates across multiple child components

### DIFF
--- a/src/components/Editor/DebugLine.js
+++ b/src/components/Editor/DebugLine.js
@@ -1,0 +1,56 @@
+// @flow
+import { Component } from "react";
+import type { SourceEditor } from "devtools-source-editor";
+import type { Frame } from "../../types";
+
+type DebugLineProps = {
+  editor: SourceEditor,
+  frame: Frame,
+  visibleSourceId: string
+};
+
+class DebugLine extends Component {
+  props: DebugLineProps;
+
+  componentDidUpdate(prevProps: DebugLineProps) {
+    const lastFrame = prevProps.frame;
+    const lastLine = lastFrame && lastFrame.location.line;
+    const nextFrame = this.props.frame;
+    const nextLine = nextFrame && nextFrame.location.line;
+
+    if (lastLine && lastLine != nextLine) {
+      this.clearDebugLine(lastLine);
+    }
+
+    const nextSourceId = nextFrame && nextFrame.location.sourceId;
+    const { visibleSourceId } = this.props;
+
+    if (nextLine && nextSourceId === visibleSourceId) {
+      this.setDebugLine(nextLine);
+    }
+  }
+
+  clearDebugLine(line: number) {
+    this.props.editor.codeMirror.removeLineClass(
+      line - 1,
+      "line",
+      "new-debug-line"
+    );
+  }
+
+  setDebugLine(line: number) {
+    this.props.editor.codeMirror.addLineClass(
+      line - 1,
+      "line",
+      "new-debug-line"
+    );
+  }
+
+  render() {
+    return null;
+  }
+}
+
+DebugLine.displayName = "DebugLine";
+
+export default DebugLine;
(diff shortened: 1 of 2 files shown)


src/components/Editor/DebugLine.js
```diff
@@ -0,0 +1,56 @@
+// @flow
+import { Component } from "react";
+import type { SourceEditor } from "devtools-source-editor";
+import type { Frame } from "../../types";
+
+type DebugLineProps = {
+  editor: SourceEditor,
+  frame: Frame,
+  visibleSourceId: string
```
jasonLaster: I like the term visible, but I think `selectedSourceId` might be more consistent in the short term

src/components/Editor/index.js
```diff
@@ -782,6 +762,16 @@ class Editor extends PureComponent {
     };
   }
 
+  renderDebugLine() {
+    return DebugLine({
```
jasonLaster: perhaps we don't render the debug line without a selected source?

src/components/Editor/DebugLine.js
asolove: Unfortunately, that distinction is made for a very particular reason. You can select a new sourceId but not have it yet be visible (if you're waiting for it the source to be fetched, for example). For this component to know when to run its editor commands, it needs to know when the source is actually displayed. Prior to this change, the highlight would run right away, and when the source arrived, it would be put in a new document and not get the highlight. This makes me think we need a way of telling side-effecting editor children something about the identity of the displayed document.

src/components/Editor/index.js
asolove: This runs into the same thing with selected v visible, but yeah, I think it makes sense to not render if the visible document shouldn't have a highlight.

src/components/Editor/DebugLine.js
jasonLaster: I see, perhaps we want to encapsulate some of this knowledge into the component so it knows what to do... 

What if we passed the selected source in and DL checked to see if it was loading? If it's loading we don't highlight...

asolove: Unfortunately, I think we really need data about what source is shown in the editor. It's possible that the source text has loaded but we're at some state where the editor hasn't yet been updated to the new document. (Because that happens in lifecycle callbacks which may _happen_ to occur before or after this component renders with the new source data.)

asolove:
I've substantially simplified things based on my discussion with @jasonLaster. I think this small PR is now ready to go. Waiting to hear back from CodeMirror about how to batch editor changes.

jasonLaster:
@asolove we're currently prepping the debugger for Firefox 56, which will be released in two weeks. I'll look at landing this after the cutoff

asolove:
Ok, sorry this has been open so long without progress. Promise I will get back to it tis coming week. 😢

jasonLaster:
Hey @asolove, I did my best to quickly get this branch up to speed:

here's a quick branch, which is my attempt to reconcile the differences: https://github.com/devtools-html/debugger.html/compare/master...jasonLaster:editor-components

here's a diff of DebugLine on my branch and yours: https://gist.github.com/jasonLaster/56766f168895ed8a265e00ff70e996c5

The differences are:

1. `toEditorLocation` which came from wasm
2. `debugExpression` which is special highlighting of debug lines... I think it must be new as of your branch

I didn't try to run my branch, i'm sure there are issues w/ it

codehag:
Quite a bit changed in the code base, I will open a new pr based off of this one with the changes!

jasonLaster:
closing as we'll be supporting this in a new PR:
https://github.com/devtools-html/debugger.html/pull/3564/files

Issue #3091, opened 2017-06-05
[ProjectSearch] Toggle between search modes


jasonLaster:
We want to easily toggle between search modes: File search and Text search. This is the simplest UI that does that. It borrows from the Editor search bottom bar quite a bit.

<img src='https://shipusercontent.com/741f5ce0d03436a29b8d08ef12375c24/Screen%20Shot%202017-06-05%20at%2011.06.33%20AM.png' title='Screen Shot 2017-06-05 at 11.06.33 AM.png' width=857>

andreicristianpetcu:
Not sure about how to implement this but I will look into it when I will have some time.

andreicristianpetcu:
I might be totally on the wrong trail here so help me out. I see there is a [projectTextSearch](https://github.com/devtools-html/debugger.html/search?utf8=%E2%9C%93&q=projectTextSearch&type=) feature. From the description of this task and from [reading the code](https://github.com/devtools-html/debugger.html/blob/79932a76c74c9f1e54ff243e81747e5be8c4bcc1/src/components/ProjectSearch/ProjectSearch.js#L149-L154) I need to toggle between file search and text search. Text search is disabled so I enabled it in `configs/local.json`.

```json
    "projectTextSearch": {
      "label": "Project Text Search",
      "enabled": true
    },
```

Now when I try to open the project text search (Ctrl+P in the editor) I get this exception.

```javascript
TypeError: roots is undefined: _dfsFromRoots@http://localhost:8000/assets/build/debugger.js:88192:10
render@http://localhost:8000/assets/build/debugger.js:88087:22
_renderValidatedComponentWithoutOwnerOrContext/renderedComponent<@http://localhost:8000/assets/build/debugger.js:14189:17
measureLifeCyclePerf@http://localhost:8000/assets/build/debugger.js:13470:13
_renderValidatedComponentWithoutOwnerOrContext@http://localhost:8000/assets/build/debugger.js:14188:28
_renderValidatedComponent@http://localhost:8000/assets/build/debugger.js:14215:30
performInitialMount@http://localhost:8000/assets/build/debugger.js:13757:26
mountComponent@http://localhost:8000/assets/build/debugger.js:13653:17
mountComponent@http://localhost:8000/assets/build/debugger.js:4188:19
performInitialMount@http://localhost:8000/assets/build/debugger.js:13766:19
mountComponent@http://localhost:8000/assets/build/debugger.js:13653:17
[...]

PR #3266, opened 2017-07-05
[WIP] toggle project text search

jasonLaster:
Associated Issue: #3091

### Summary of Changes

Right now we map cmd+p to project text search if it is enabled. We should be able to have cmd+p map to source search and cmd+shift+f map to project text search.

This does that. It doesn't add the bottom bar for toggling, but it helps.


src/components/Editor/SearchBar.js
```diff
@@ -68,7 +68,7 @@ class SearchBar extends Component {
     highlightLineRange: ({ start: number, end: number }) => void,
     clearHighlightLineRange: () => void,
     searchOn?: boolean,
-    toggleActiveSearch: (?ActiveSearchType) => any,
```
jasonLaster: switching from the verb toggle to set because we have more than two now: Project, source, file, symbol

src/components/ProjectSearch/ProjectSearch.js
```diff
@@ -4,165 +4,125 @@ import { DOM as dom, PropTypes, Component, createFactory } from "react";
 import { connect } from "react-redux";
 import { bindActionCreators } from "redux";
 import actions from "../../actions";
-import { getSources, getActiveSearchState } from "../../selectors";
```
jasonLaster: moving source search into its own component will untangle a lot of this code

src/components/ProjectSearch/ProjectSearch.js
```diff
@@ -4,165 +4,125 @@ import { DOM as dom, PropTypes, Component, createFactory } from "react";
...
-function projectSearchResults(sources, query) {
-  return sources.valueSeq().toJS().map(source =>
-    searchSource(source, query).map(result => ({
-      value: result.match,
-      title: result.text,
-      subtitle: endTruncateStr(getSourcePath(source), 100),
-      id: `${result.text}/${result.line}/${result.column}`
-    }))
-  );
-}
```
jasonLaster: so... this wasn't used :)

src/components/ProjectSearch/ProjectSearch.js
```diff
@@ -1,180 +0,0 @@
...
-
-import { DOM as dom, PropTypes, Component, createFactory } from "react";
-import { connect } from "react-redux";
-import { bindActionCreators } from "redux";
-import actions from "../../actions";
-import { getSources, getActiveSearchState } from "../../selectors";
-import { endTruncateStr } from "../../utils/utils";
-import { parse as parseURL } from "url";
-import { isPretty } from "../../utils/source";
-import { isEnabled } from "devtools-config";
```
jasonLaster: ProjectSearch now is a coordinator, it just decides which search UI to show and will eventually have buttons to switch between them

Issue #3222, opened 2017-06-24
[Editor] HTML should be syntax highlighted


jasonLaster:
I think we had this working earlier. My guess is that this is related to the sourceText -> source conversion.

<img src='https://shipusercontent.com/da4c998fc7a229a91bb0b28530a74901/Screen%20Shot%202017-06-23%20at%209.41.16%20PM.png' title='Screen Shot 2017-06-23 at 9.41.16 PM.png' width=878>

jbhoosreddy:
@jasonLaster thank god I read this. I had so many confusions. I still see parts of code in actions/ast, etc. rely on source text. I'm guessing we want to change remaining sourcesText -> sources?

jasonLaster:
Yeah. Where are the source text references?

PR #3271, opened 2017-07-06
Html syntax highlight bug

jbhoosreddy:
Associated Issue: #3222

### Summary of Changes

* fixes html syntax highlighting

### Test Plan

Example test plan:

- [x] Test passing
- [x] Visual Inspection

screen shot 2017-07-05 at 10 19 14 pm

### DIFF
--- a/src/utils/source.js
+++ b/src/utils/source.js
@@ -122,11 +122,21 @@ const contentTypeModeMap = {
  */
 
 function getMode(source: Source) {
-  if (!source.text) {
+  const { contentType, text } = source;
+
+  if (!text) {
     return { name: "text" };
   }
 
-  const { contentType, text } = source;
+  // Use HTML mode for files in which the first non whitespace
+  // character is `<` regardless of extension.
+  const isHTMLLike = text.match(/^\s*</);
+  if (!contentType) {
+    if (isHTMLLike) {
+      return { name: "htmlmixed" };
+    }
+    return { name: "text" };
+  }
 
   // // @flow or /* @flow */
   if (text.match(/^\s*(\/\/ @flow|\/\* @flow \*\/)/)) {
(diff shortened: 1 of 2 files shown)


src/utils/source.js
```diff
@@ -122,32 +122,39 @@ const contentTypeModeMap = {
...
-  if (/script|elm|jsx|clojure|wasm/.test(contentType)) {
+  if (/script|elm|jsx|clojure|wasm|html/.test(contentType)) {
     if (contentType in contentTypeModeMap) {
       return contentTypeModeMap[contentType];
     }
 
     return contentTypeModeMap["text/javascript"];
   }
 
+  if (isHTMLLike) {
```
jbhoosreddy: I added a second check here because it can happen that we do have a `contentType = "text/html"` or it is equal to something else. In that case, we would still like to render it as `htmlmixed`.

I believe that this logic might be simplified if in `devtools-core/devtools-source-maps` we set html contentType to `text/html` for file extensions `.html`.

In fact, I'm not sure if there's any case for `(!contentType && isHTMLLike)` as far as I could tell in any codepath.

src/utils/source.js
```diff
@@ -122,32 +122,39 @@ const contentTypeModeMap = {
  */
 
 function getMode(source: Source) {
-  if (!source.text) {
+  const { contentType, text } = source;
```
codehag: :+1:

src/utils/source.js
codehag: the check for `if (!contentType)` has the advantage of an early exit -- we don't perform the other checks, which do rely on content type. So i think we can get rid of lines 154 - 157. If we have a file that is html-like, and it has a contentType, it will be caught by line 146. if all checks fail, we return `{ name: "text" }`

we can also simplify the if statements in the block on lines 134-138

You have a better view of whats going on though, let me know if I miss interpreted!

jbhoosreddy:
Since it's approved I'll go ahead and merge it. The failing CI test is only because of Percy. Rest of the tests pass and I double checked locally.

Issue #3277, opened 2017-07-07
[SourceSearch] selectSource is not a function


jasonLaster:
It looks like we broke source search with the recent `cmd+shift+f` PR. Lets add an integration test when we fix this.

PR #3281, opened 2017-07-07
Fix source search

jasonLaster:
Associated Issue: #3277


src/test/mochitest/browser.ini
```diff
@@ -69,7 +69,6 @@ skip-if = os == "linux" # bug 1351952
 [browser_dbg-pretty-print-paused.js]
 [browser_dbg-scopes-mutations.js]
 [browser_dbg-searching.js]
-skip-if = true
```
jasonLaster: just had to turn it ont

jasonLaster:
yes, we were skipping an old test. hopefully it will be stable now

Issue #3167, opened 2017-06-14
Cannot set debugger breakpoint on generatedSource


codehag:
Steps to reproduce:
- open a debuggee with a generated source
- add a breakpoint to the generated source
- the breakpoint will be moved to the original source

Expected:
- breakpoint will remain in the generated source

jasonLaster:
Here's a non-pretty answer.
https://gist.github.com/cda7b293c3024b3a1f57d41969ef9ffe

It takes advantage of the fact that we *do* save the generated location here...

jasonLaster:
We landed the first version. There are some next steps:

- [ ] Refactor components
- [ ] Support toggling between original/generated sources #3288
- [ ] Simplify breakpoint clients

PR #3284, opened 2017-07-07
Bugfix: Cannot set debugger breakpoint on generatedSource

codehag:
Associated Issue: #3167 

Screenshot:

3167bug

### DIFF
--- a/src/actions/breakpoints.js
+++ b/src/actions/breakpoints.js
@@ -9,9 +9,13 @@
  */
 
 import { PROMISE } from "../utils/redux/middleware/promise";
-import { getBreakpoint, getBreakpoints, getSource } from "../selectors";
-import { originalToGeneratedId } from "devtools-source-map";
-import { equalizeLocationColumn } from "../utils/breakpoint";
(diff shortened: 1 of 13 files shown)


jasonLaster:
I made a couple updates:

1. support removing a breakpoint in a generated source
2. added an integration test for toggling bps in a generated source

jasonLaster:
Looks like `doc.setGutterMarker` does not exist in the panels version of codemirror. We should probably update codemirror on the panel and continue from there

jasonLaster:
I just pushed a commit that disables the test for slid bps and reloading because the test fails in this PR, but doesn't really test the behavior. I've created a new issue #3290 so that we can get this working properly and re-enable the test

jasonLaster:
I did a bit of dbg hacking with Jaideep today and got the bp bug is now ready to merge with some caveats:
1. Toggling sources requires a cm upgrade in a new PR 
2. We'll want to take another look at reloading + slid bps. I initially thought this PR changed that, but it's just a timing change. There is actually a legit bug on master w that (I filed an issue for that)
3. I plan on doing a follow up refactor PR which will clean up the location logic considerably.

PR #3287, opened 2017-07-09
Bump dependencies

jasonLaster:


src/components/Editor/GutterMenu.js
```diff
@@ -13,9 +13,9 @@ export default function GutterMenu({
   event.stopPropagation();
   event.preventDefault();
   let breakpoint = {
-    id: "node-menu-add-breakpoint",
-    label: L10N.getStr("editor.addBreakpoint")
-  },
+      id: "node-menu-add-breakpoint",
+      label: L10N.getStr("editor.addBreakpoint")
+    },
```
codehag: hmm this is old code, but it's hard to read. we are basically doing this here:
```
let breakpoint = { ... };
let conditional = { ... };
let disabled;
```
but because of how it is written, its easy to see this instead:
```
let breakpoint = {
  id: ..
  label: ..
  conditional: ...
  disabled
};
```

We should add an eslint rule that doens't allow for declarations to be done in a list, because there is little gain from this style, and it decreases comprehension. Adding this as a comment for myself so that we can do this later.

codehag:
We have an issue with a license
```
Invalid license for jschardet@1.4.2: LGPL-2.1+
```

codehag:
Is there any news about the licensing issue?

PR #3298, opened 2017-07-10
fix getBreakpointsFromSource

jasonLaster:


jasonLaster:
Quick fix to get master working again. My larger refactor for BPs will be the real solution :)

Issue #3274, opened 2017-07-07
[Editor] Context menu to allow developers to copy code


jbhoosreddy:
We should allow developers to copy code from the editor. Might not be a common usecase, but given that I've ran into situations where I needed this, I'd like to push for it.
Chrome and old FF Debugger both support this. 

In my opinion, Chrome context menu is over done. I like Firefox's context menu better. It is neat and purposeful. There is clear distinction and keyboard shortcuts are also provided. We could probably do better with the new Debugger.

### Steps
1. Add context menu
2. Get selection from CodeMirror
3. Set selection in Paste Buffer

### New Debugger
<img width="242" alt="screen shot 2017-07-06 at 10 26 13 pm" src="https://user-images.githubusercontent.com/2481105/27940717-62f6d72a-629a-11e7-9a20-ceae1f0fbba5.png">

### Old Debugger
<img width="374" alt="screen shot 2017-07-06 at 10 19 47 pm" src="https://user-images.githubusercontent.com/2481105/27940639-cf8f0714-6299-11e7-8120-ce981a2f93e6.png">

### Chrome
<img width="430" alt="screen shot 2017-07-06 at 10 20 43 pm" src="https://user-images.githubusercontent.com/2481105/27940642-d2eb4152-6299-11e7-9e35-dd6ce9b09270.png">

Also, once you open the context menu, if you right click again, it brings open the native context menu. Which looks like a bug. We should research removing it or handling it better.
![](http://g.recordit.co/NiBeW6ZiaQ.gif)
Looking into it, browsers have `oncontextmenu` event which can allow us to stop showing native context menu altogether.

codehag:
This has been added. Also - Ctrl-c/Ctrl-v works!

PR #3299, opened 2017-07-11
disable native context menu on window

jbhoosreddy:
Associated Issue: #3274 

### Summary of Changes

* disable native context menu.

Since we `stopPropogation` and `preventDefault` at various components down the tree, I couldn't disable the native context menu on `App` as the event doesn't bubble up there. I found this other solution online. 

### Test Plan
1. Visual Testing


jasonLaster:
Hmm, I think we should handle this on the context menu mask. This would work all the time :)

PR #3322, opened 2017-07-12
Spike/refactor add breakpoint

codehag:


src/actions/thunks/apis/breakpoints.js
```diff
@@ -0,0 +1,53 @@
+import {
+  getGeneratedLocation,
+  locationMoved,
+  equalizeLocationColumn,
+  breakpointExists
+} from "../../../utils/breakpoint";
+import { getSource } from "../../../selectors";
+
+export async function addBreakpoint(
```
jasonLaster: I think it would be nice to keep addBreakpoint together for now. That way we have more flexibility in the future.

```js
export function addBreakpoint(location: Location, condition: string) {
  const breakpoint = createBreakpoint(location, { condition });
  return ({ dispatch, getState, sourceMaps, client }: ThunkArgs) => {
     // option 1: inline addBreakpoint
     const state = getState();
     //...

    // option 2: call addBreakpoint
    return addThunkBreakpoint(getState, client, sourceMaps,  { breakpoint })
  }
}
```  



also, perhaps the filepath is `actions/breakpoints/addBreakpoint.js`

PR #3340, opened 2017-07-14
[wip] await step over

jbhoosreddy:
Associated Issue: #3106

### Summary of Changes

### Test Plan

- [x] Visual Testing

### DIFF
--- a/src/actions/breakpoints.js
+++ b/src/actions/breakpoints.js
@@ -95,12 +95,22 @@ export function syncBreakpoint(
  * @param {Boolean} $1.disabled Disable value for breakpoint value
  */
 
-export function addBreakpoint(location: Location, condition: ?string) {
-  const breakpoint = createBreakpoint(location, { condition });
+export function addBreakpoint(
+  location: Location,
+  condition: ?string,
+  hidden: ?boolean
+) {
+  const breakpoint = createBreakpoint(location, { condition, hidden });
   return ({ dispatch, getState, sourceMaps, client }: ThunkArgs) => {
     const action = { type: "ADD_BREAKPOINT", breakpoint };
     const promise = addBreakpointPromise(getState, client, sourceMaps, action);
     return dispatch({ ...action, [PROMISE]: promise });
+  };
+}
+
+export function addHiddenBreakpoint(location: Location) {
+  return ({ dispatch }: ThunkArgs) => {
+    return dispatch(addBreakpoint(location, "", true));
   };
 }
 
@@ -215,7 +225,7 @@ export function setBreakpointCondition(
   return ({ dispatch, getState, client, sourceMaps }: ThunkArgs) => {
     const bp = getBreakpoint(getState(), location);
     if (!bp) {
-      return dispatch(addBreakpoint(location, condition));
+      return dispatch(addBreakpoint(location, condition, false));
     }
 
     if (bp.loading) {
--- a/src/actions/pause.js
+++ b/src/actions/pause.js
@@ -3,9 +3,12 @@
 import { selectSource } from "./sources";
 import { PROMISE } from "../utils/redux/middleware/promise";
 
-import { getPause, getLoadedObject } from "../selectors";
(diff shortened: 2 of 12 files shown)


src/actions/breakpoints.js
```diff
@@ -222,7 +229,7 @@ export function setBreakpointCondition(
   return ({ dispatch, getState, client, sourceMaps }: ThunkArgs) => {
     const bp = getBreakpoint(getState(), location);
     if (!bp) {
-      return dispatch(addBreakpoint(location, { condition }));
+      return dispatch(addBreakpoint(location, { condition, hidden: false }));
```
jasonLaster: `hidden: false` should be the default: 

`dispatch(addBreakpoint(location, { condition }))`

src/actions/pause.js
```diff
@@ -117,7 +120,7 @@ export function command({ type }: CommandType) {
 export function stepIn() {
   return ({ dispatch, getState }: ThunkArgs) => {
     if (getPause(getState())) {
-      return dispatch(command({ type: "stepIn" }));
+      return dispatch(AstCommand("stepIn"));
```
jasonLaster: `astCommand`

src/reducers/breakpoints.js
```diff
@@ -384,4 +385,23 @@ export function getBreakpointsForSource(state: OuterState, sourceId: string) {
...
+
+export const getHiddenBreakpoint = createSelector(
+  getHiddenBreakpoints,
+  hiddenBreakpoints => {
+    if (hiddenBreakpoints.length) {
+      return hiddenBreakpoints[0];
+    }
+    return null;
+  }
+);
```
jasonLaster: maybe we're thinking about this the wrong way. If we're only going to have one temporary breakpoint, perhaps we want a different breakpoint field in the reducer like `pendingBreakpoints`.

I propose we add a `steppingBreakpoint` which does not appear in the UI... do you think that would be easy? we also don't need a pending breakpoint for this case.

src/utils/parser/steps.js
```diff
@@ -0,0 +1,53 @@
...
+    return;
+  }
+  for (let i = 0; i !== children.length; i++) {
+    const child = children[i];
+    if (containsPosition(child.loc, position)) {
+      const nextChild = children[++i];
+      const nextLocation = nextChild.loc.start;
+      nextLocation.sourceId = position.sourceId;
+      nextLocation.column = 0;
+      return nextLocation;
```
jasonLaster: nit:

```js
return {
   sourceId: position.sourceId
   line: nextLocation.line,
   column: 0
}
```

src/utils/parser/steps.js
```diff
@@ -0,0 +1,53 @@
+import { Source } from "../../../flow-typed/debugger-html";
+import { AstPosition } from "./types";
+import { getClosestPath } from "./utils/closest";
+import { isAwaitExpression, containsPosition } from "./utils/helpers";
+import type { NodePath } from "babel-traverse";
+
+export function getNextStep(
```
jasonLaster: it would be nice to have unit tests here too.

src/actions/breakpoints.js
jbhoosreddy: flow complained here. maybe there's a better solution.

src/actions/breakpoints.js
```diff
@@ -95,15 +95,25 @@ export function syncBreakpoint(
...
   return ({ dispatch, getState, sourceMaps, client }: ThunkArgs) => {
     const action = { type: "ADD_BREAKPOINT", breakpoint };
     const promise = addBreakpointPromise(getState, client, sourceMaps, action);
     return dispatch({ ...action, [PROMISE]: promise });
   };
 }
 
+export function addHiddenBreakpoint(location: Location) {
+  return ({ dispatch }: ThunkArgs) => {
+    return dispatch(addBreakpoint(location, "", true));
```
jasonLaster: at this point it would make sense to do `addBreakpoint({location, condition, hidden})`

Issue #3288, opened 2017-07-09
[Breakpoints] Toggle between original and generated files


jasonLaster:
This issue extends the work done in #3167.

We currently support adding a BP in either an original file or generated and seeing it in the other.

We do not support removing a BP in one file and seeing it removed in the other.

STR:
1. add a BP in an original file
2. go to the generated file
3. remove it from either
4. navigate to the other

Expected: It should be gone
Actual: It is there.

This happens because of React lifecycles. The problem is that the Editor Breakpoint component tries to remove the BP from the new document. We can fix this by getting the correct document all the time. This is blocked by upgrading CodeMirror to the most up to date version.

```diff
diff --git a/src/components/Editor/Breakpoint.js b/src/components/Editor/Breakpoint.js
index daa64c6..8549257 100644
--- a/src/components/Editor/Breakpoint.js
+++ b/src/components/Editor/Breakpoint.js
@@ -58,14 +58,17 @@ class Breakpoint extends Component {
 
     const location = this.getLocation();
     const line = location.line - 1;
-    const doc = getDocument(selectedSource.get("id"));
 
-    doc.setGutterMarker(line, "breakpoints", makeMarker(breakpoint.disabled));
-    doc.addLineClass(line, "line", "new-breakpoint");
+    editor.setGutterMarker(
+      line,
+      "breakpoints",
+      makeMarker(breakpoint.disabled)
+    );
+    editor.addLineClass(line, "line", "new-breakpoint");
     if (breakpoint.condition) {
-      doc.addLineClass(line, "line", "has-condition");
+      editor.addLineClass(line, "line", "has-condition");
     } else {
-      doc.removeLineClass(line, "line", "has-condition");
+      editor.removeLineClass(line, "line", "has-condition");
     }
   }
 
@@ -104,11 +107,10 @@ class Breakpoint extends Component {
     }
 
     const line = location.line - 1;
-    const doc = getDocument(selectedSource.get("id"));
 
-    doc.setGutterMarker(line, "breakpoints", null);
-    doc.removeLineClass(line, "line", "new-breakpoint");
-    doc.removeLineClass(line, "line", "has-condition");
+    editor.setGutterMarker(line, "breakpoints", null);
+    editor.removeLineClass(line, "line", "new-breakpoint");
+    editor.removeLineClass(line, "line", "has-condition");
   }
```

jasonLaster:
This should be fixed when codemirror is updated on MC :)

PR #3351, opened 2017-07-14
[Breakpoints] Toggle between original and generated files

jasonLaster:
Associated Issue: #3288

### Summary of Changes

* updates the breakpoint pane so that it is document aware
* we'll need to update codemirror in the docker image, so once codemirror lands on MC I can do a docker update


codehag:
Looks good, but there is a failure on the CI, can you take a look?

PR #3360, opened 2017-07-17
Upgraded webpack to version 3.3.0

zaggy:
Associated Issue: #2418 (any other webpack update issue)

### Summary of Changes

This just does bump flow-bin version and adds postcss.config.js file for webpack 3. It also updates webpack dependency but it looks like it has nothing to do with how project is built and real config is in devtools-launchpad.

So, this changes need to wait until PR#505 at devtools-core will land and new version of devtools-launchpad is released. 

### Test Plan

 When new version of devtools-launchpad is released, visual testing is required. To run the project and check for errors. There is a very high chance of some regressions.

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,7 +68,7 @@
     "eslint-config-prettier": "^2.1.0",
     "eslint-plugin-prettier": "^2.0.1",
     "expect.js": "^0.3.1",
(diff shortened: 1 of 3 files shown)


zaggy:
Oh, I messed up with this PR.

Issue #3263, opened 2017-07-05
[Preview] start using symbols in preview


jasonLaster:
Now that we have all the symbols for a file we can do a synchronous version of preview that does not need to traverse the AST when you mouse over a token.

Here are some new examples of tokens we can support:

* mousing over an object/array property
* properties inside of a call expression assignment `var App.TodoView = extend({ render() {} })`
* object properties inside of member expressions

Here's a branch: https://github.com/jasonLaster/debugger.html/tree/symbol-preview

codehag:
I'm having a bit of difficulty getting started on this, as the branch is a bit behind master and it's hard to tell what is going on/what changed. Do you have an example of current vs desired behavior?

PR #3363, opened 2017-07-17
WIP Feature/3263 preview start using symbols in preview

codehag:
Associated Issue: #3263

### DIFF
--- a/src/actions/ast.js
+++ b/src/actions/ast.js
@@ -79,6 +79,31 @@ export function clearSelection() {
   };
 }
 
+function findBestMatch(symbols, tokenPos, token) {
+  const { memberExpressions, identifiers } = symbols;
+  const { line, column } = tokenPos;
+  return memberExpressions.concat(identifiers).reduce((found, expression) => {
+    const isNear =
+      expression.location.start.line == line &&
+      expression.location.start.column <= column &&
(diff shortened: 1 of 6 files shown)


src/utils/parser/getSymbols.js
```diff
@@ -119,6 +119,16 @@ function extractSymbols(source: Source) {
         });
       }
 
+      if (t.isThisExpression(path.node)) {
+        const { start, end } = path.node.loc;
+        memberExpressions.push({
```
jasonLaster: this is not a memberExpressions, maybe it should be an identifier?

jasonLaster: we should add a test to `fixture/expressions`

src/actions/ast.js
```diff
@@ -79,6 +79,20 @@ export function clearSelection() {
   };
 }
 
+function findSymbol(symbols, tokenPos) {
+  const array = Object.keys(symbols).reduce(
```
jasonLaster: we probably need an explicit order here to prefer more specific types like `member expression`...

Or not if the start/end loc help out

src/utils/parser/getSymbols.js
codehag: :+1:

Issue #3368, opened 2017-07-18
blank debugger when debugging service workers via about:debugging


clarkbw:
Coming out of [bug 1380085](https://bugzilla.mozilla.org/show_bug.cgi?id=1380085) the debugger now comes up blank. 😢 

STEPS TO REPRODUCE:

1)  Start current nightly with clean profile `/Applications/FirefoxNightly.app/Contents/MacOS/firefox --profile $(mktemp -d)`
2)  Load about:debugging.
3)  Click the "Workers" text in the sidebar on the left.
4)  Click the "Opt out of multiple content processes" button in the yellow
    warning about service worker debugging not being e10s-multi compatible.
5)  Click "OK" in the resulting confirmation prompt.  This will restart your
    browser.
6)  After restart, open a new tab in addition to about:debugging.
7)  Load https://perfht.ml/2uavsT8 in that new tab.
8)  Go back to the about:debugging tab.
9)  Click the "Debug" button for the running "https://perf-html.io/sw.js" service
    worker.
10) Observe the resulting debugger window that comes up.

I'm seeing this error in the browser console.

```
TypeError: tabTarget._form is undefined
Stack trace:
exports.onConnect</_ref<@resource://gre/modules/commonjs/toolkit/loader.js -> resource://devtools/client/debugger/new/debugger.js:32591:6
step@resource://gre/modules/commonjs/toolkit/loader.js -> resource://devtools/client/debugger/new/debugger.js:32611:184
step/<@resource://gre/modules/commonjs/toolkit/loader.js -> resource://devtools/client/debugger/new/debugger.js:32611:362
promise callback*step@resource://gre/modules/commonjs/toolkit/loader.js -> resource://devtools/client/debugger/new/debugger.js:32611:315
promise callback*step@resource://gre/modules/commonjs/toolkit/loader.js -> resource://devtools/client/debugger/new/debugger.js:32611:315
_asyncToGenerator/</<@resource://gre/modules/commonjs/toolkit/loader.js -> resource://devtools/client/debugger/new/debugger.js:32611:438
_asyncToGenerator/<@resource://gre/modules/commonjs/toolkit/loader.js -> resource://devtools/client/debugger/new/debugger.js:32611:100
onConnect@resource://gre/modules/commonjs/toolkit/loader.js -> resource://devtools/client/debugger/new/debugger.js:32603:13
[...]

Issue #3313, opened 2017-07-11
New debugger frontend breaks serviceworker debugging


codehag:
copy of https://bugzilla.mozilla.org/show_bug.cgi?id=1380085

STEPS TO REPRODUCE:

1)  Start current nightly with clean profile.
2)  Load about:debugging.
3)  Click the "Workers" text in the sidebar on the left.
4)  Click the "Opt out of multiple content processes" button in the yellow
    warning about service worker debugging not being e10s-multi compatible.
5)  Click "OK" in the resulting confirmation prompt.  This will restart your
    browser.
6)  After restart, open a new tab in addition to about:debugging.
7)  Load https://perfht.ml/2uavsT8 in that new tab.
8)  Go back to the about:debugging tab.
9)  Click the "Debug" button for the running "https://perf-html.io/sw.js" service
    worker.
10) Observe the resulting debugger window that comes up.

EXPECTED RESULTS: Has service worker source, lets you debug it.

ACTUAL RESULTS: "This page has no sources".

ADDITIONAL INFORMATION: Toggling the "devtools.debugger.new-debugger-frontend" preference to "false" makes things work: clicking the Debug button shows the service worker source.  Toggling it back to "true" makes things not work again.

jasonLaster:
https://pastebin.mozilla.org/9026873

jasonLaster:
https://bugzilla.mozilla.org/show_bug.cgi?id=1380867

PR #3373, opened 2017-07-19
fix service workers

jasonLaster:
Associated Issue: #3313

### DIFF
--- a/assets/panel/panel.js
+++ b/assets/panel/panel.js
@@ -29,7 +29,7 @@ DebuggerPanel.prototype = {
     } = await this.panelWin.Debugger.bootstrap({
       threadClient: this.toolbox.threadClient,
       tabTarget: this.toolbox.target,
-      debuggerClient: this.toolbox.target._client,
+      debuggerClient: this.toolbox.target.client,
       sourceMaps: this.toolbox.sourceMapService
     });
 
--- a/src/client/firefox.js
+++ b/src/client/firefox.js
@@ -30,7 +30,7 @@ export async function onConnect(connection: any, actions: Object) {
   // bfcache) so explicity fire `newSource` events for all returned
   // sources.
   const sources = await clientCommands.fetchSources();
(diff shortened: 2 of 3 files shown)


assets/panel/panel.js
```diff
@@ -29,7 +29,7 @@ DebuggerPanel.prototype = {
     } = await this.panelWin.Debugger.bootstrap({
       threadClient: this.toolbox.threadClient,
       tabTarget: this.toolbox.target,
-      debuggerClient: this.toolbox.target._client,
+      debuggerClient: this.toolbox.target.client,
```
jasonLaster: this is already in m-c

src/client/firefox.js
```diff
@@ -30,7 +30,7 @@ export async function onConnect(connection: any, actions: Object) {
   // bfcache) so explicity fire `newSource` events for all returned
   // sources.
   const sources = await clientCommands.fetchSources();
-  actions.connect(tabTarget._form.url);
+  actions.connect(tabTarget.url);
```
jasonLaster: there's a getter on the targets that we can use

clarkbw:
This fixes #3368

src/client/firefox/commands.js
```diff
@@ -160,6 +160,10 @@ type EvaluateParam = {
 
 function evaluate(script: Script, { frameId }: EvaluateParam) {
   const params = frameId ? { frameActor: frameId } : {};
+  if (!tabTarget || !tabTarget.activeConsole) {
+    return Promise.resolve();
```
bgrins: What's the case when this happens? Is this expected until we fix some issue on the target? Perhaps we should console.warn in this case

bgrins: A comment above this condition explaining would be helpful as well

jasonLaster: Just added an issue for attaching to the console synchronously.

https://bugzilla.mozilla.org/show_bug.cgi?id=1382302

PR #3395, opened 2017-07-20
fix rest spread

jasonLaster:
### Summary of Changes

* fixes a bug where i couldn't use preview in actions/ast.js lol
* found the babylon plugin here: https://github.com/babel/babylon

### DIFF
--- a/src/utils/parser/tests/fixtures/es6.js
+++ b/src/utils/parser/tests/fixtures/es6.js
@@ -1,0 +1,1 @@
+dispatch({ ...action, [PROMISE]: promise });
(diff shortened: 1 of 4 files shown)


src/utils/parser/tests/fixtures/es6.js
```diff
@@ -0,0 +1 @@
+dispatch({ ...action, [PROMISE]: promise });
```
codehag: I am not too sure what is happening here, where are action and promise coming from?

jasonLaster: haha - they're not, but the parser doesn't care. it's just doing a `...` rest operation

codehag: ahh ok cool

PR #3396, opened 2017-07-20
[ProjectSearch] toggle between project text search and file name search

aklt:
The code is not too pretty, but it does what is required and since this
is a temporary ui I think it is ok.

Autocomplete was modified to render children before results

Associated Issue: #3091 
### Summary of Changes

* This adds UI for the project text search and files search (CMD or CTRL + o and CMD or CTRL + Shift + f) as shown in the image in issue 3091 
* This is temporary UI

### Test Plan

* Enable "projectTextSearch" in `configs/local.json`
* Press the above key combos to bring up the search bar
* Click on the active "files" or "text" link to change the UI

### Screenshots

toggle-search-1

### DIFF
--- a/assets/panel/debugger.properties
+++ b/assets/panel/debugger.properties
@@ -466,7 +466,7 @@ welcome.search=%S to search for sources
 
 # LOCALIZATION NOTE (sourceSearch.search): The center pane Source Search
 # prompt for searching for files.
-sourceSearch.search=Search Sources…
+sourceSearch.search=Find files
 
 # LOCALIZATION NOTE (sourceSearch.noResults): The center pane Source Search
 # message when the query did not match any of the sources.
--- a/assets/panel/debugger.properties
+++ b/assets/panel/debugger.properties
@@ -140,6 +140,10 @@ sources.search.alt.key=CmdOrCtrl+O
 # full project text search for searching all the files the debugger has seen.
 projectTextSearch.key=CmdOrCtrl+Shift+F
 
+# LOCALIZATION NOTE (projectTextSearch.placeholder): A placeholder shown
+# when searching across all of the files in a project.
+projectTextSearch.placeholder=Find in files…
+
 # LOCALIZATION NOTE (sources.noSourcesAvailable): Text shown when the debugger
 # does not have any sources.
 sources.noSourcesAvailable=This page has no sources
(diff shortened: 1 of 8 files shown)


src/components/ProjectSearch/index.js
```diff
@@ -118,12 +100,27 @@ class ProjectSearch extends Component {
...
+    });
   }
 
   renderTextSearch() {
     const { sources, closeActiveSearch } = this.props;
-    return TextSearch({ sources, closeActiveSearch });
+    return TextSearch({
+      sources,
+      closeActiveSearch,
+      searchBottomBar: ToggleSearch({
```
jasonLaster: smart 👍

src/components/ProjectSearch/index.js
```diff
@@ -118,12 +100,27 @@ class ProjectSearch extends Component {
 
   renderSourceSearch() {
     const { sources, selectSource, closeActiveSearch } = this.props;
-    return SourceSearch({ sources, selectSource, closeActiveSearch });
+    return SourceSearch({
+      sources,
+      selectSource,
+      closeActiveSearch,
+      searchBottomBar: ToggleSearch({
+        kind: "files",
```
jasonLaster: i think we should stay consistent with the search type and call it "sources"

assets/panel/debugger.properties
```diff
@@ -466,7 +466,7 @@ welcome.search=%S to search for sources
 
 # LOCALIZATION NOTE (sourceSearch.search): The center pane Source Search
 # prompt for searching for files.
-sourceSearch.search=Search Sources…
```
jasonLaster: hmm - what do others think I like this new name.

If we change it we should bump the key to `sourceSearch.search2`

aklt: I was not sure If I should change it, but i think i like it better

src/components/ProjectSearch/index.js
aklt: Ok, I will change that.  Screenshot coming up

jasonLaster:
Looks like the styling could be tweaked:

1. can you use the selected blue we use in other places
2. search is too bold
3. hover states should have a pointer cursor and slight grey
4. it looks like we're highlighting the wrong one (clicking behavior) is not what i'd expect

jasonLaster:
Perhaps we want to change the style of the links.

so it's two options: "File Files", "Find in Project"

aklt:
In which way should the style be changed?
I changed the blue text to be just spans btw.

jasonLaster:
Perhaps just two links. Sorry it's tough to describe. I dont mind jumping
in

aklt:
not sure I understand, you are welcome to jump in of course ;-)

jasonLaster:
@aklt I made some tweaks to the look and feel and the source code. 

I really like how bottom bar was structured. I updated the ternary a bit so we wouldn't duplicate logic

aklt:
@jasonLaster Looks good to me. It is good to have less dom probably and you can also toggle the link if it is active as far as I can see.  I can try it out tomorrow evening :-)

PR #3407, opened 2017-07-21
Fix editor state

jasonLaster:
### Summary of Changes

1. protects against cases where editor is not set
2. removes unnecessary guards

there is likely an abstraction we can add in the future to protect us further

### DIFF
--- a/src/components/Editor/CallSite.js
+++ b/src/components/Editor/CallSite.js
@@ -50,10 +50,7 @@ export default class CallSite extends Component {
   }
 
   componentDidMount() {
-    const { breakpoint, editor, showCallSite } = this.props;
-    if (!editor) {
-      return;
-    }
+    const { breakpoint, showCallSite } = this.props;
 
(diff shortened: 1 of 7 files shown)


src/components/Editor/index.js
```diff
@@ -543,6 +552,28 @@ class Editor extends PureComponent {
...
+    }
+
+    return {
+      height:
+        subtractions.length === 0
+          ? "100%"
+          : `calc(100% - ${subtractions.join(" - ")})`
+    };
+  }
+
```
jasonLaster: moved some helpers above the render calls

src/components/Editor/CallSite.js
```diff
@@ -51,9 +51,6 @@ export default class CallSite extends Component {
 
   componentDidMount() {
     const { breakpoint, editor, showCallSite } = this.props;
-    if (!editor) {
-      return;
-    }
```
jasonLaster: killing unnecessary checks.

Issue #3318, opened 2017-07-12
[WatchExpressions] The debugger is wrongfully showing a watch expression as a source


nchevobbe:
Steps to reproduce: 

1. Open the debugger
2. Add a watch expression (e.g. `yo`)
3. Open the console
4. Execute 
```js
{ 
  let x = [()=>{}]; 
  debugger; 
}
```

Expected result:
The debugger shows the following snippet in a tab
```js
{ 
  let x = [()=>{}]; 
  debugger; 
}
```

Actual result: 
The debugger shows the snippet briefly, and then the content of the watch expression (see GIF)
If you edit the watch expression, the tab content get's updated :)

![jul-12-2017 09-43-18](https://user-images.githubusercontent.com/578107/28107076-9a2dd004-66e6-11e7-964e-229f7a584a5e.gif)

This is visible in latest Nightly, I couldn't test against debugger's master

jasonLaster:
This is a dupe of #3329

PR #3413, opened 2017-07-21
[WatchExpressions] The debugger is wrongfully showing a watch expression as a source

jasonLaster:
Fixes: #3318

### Summary of Changes

* stop showing tabs that don't have a URL (was causing issues and showing the wrong source)
* stop selecting sources that don't have a URL

### Test Plan

Unit test

### screenshot

screen shot 2017-07-21 at 4 47 28 pm

### DIFF
--- a/src/actions/tests/sources.js
+++ b/src/actions/tests/sources.js
@@ -9,7 +9,8 @@ const {
   getSources,
   getSelectedSource,
   getSourceTabs,
-  getOutOfScopeLocations
+  getOutOfScopeLocations,
+  getSelectedLocation
 } = selectors;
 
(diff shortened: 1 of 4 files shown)


src/actions/tests/sources.js
```diff
@@ -184,105 +186,13 @@ describe("sources", () => {
     const badSource = getSource(getState(), "bad-id");
     expect(badSource.get("error").indexOf("unknown source")).not.toBe(-1);
   });
-});
-
-describe("closing tabs", () => {
```
jasonLaster: moved tabs to their own file

Issue #3234, opened 2017-06-28
[SymbolModal] Animate the symbol search enter/leave


wldcordeiro:
Use `react-transition-group` to animate the modal's enter and leaving. I think a slide in/out animation should look good.

jasonLaster:
I'm not too familiar with animating in CSS or JS, so i'll defer to others.

I think the initial idea was to animate the modal in from the top.

By the way, the modal we're talking about is opened with cmd+shift+o. It is used to search for functions now, but in the future will be files and variables as well.

wldcordeiro:
I'll jump on this. Easing back in and I've used this recently. 😄

PR #3415, opened 2017-07-22
Animate the SymbolModal's entrance/exit

wldcordeiro:
Associated Issue: #3234 

### Summary of Changes

* Break the Modal itself into it's own component
* Animate that component
* Use it in SymbolModal

### Test Plan

- [ ] <kbd>cmd</kbd>+<kbd>shift</kbd>+<kbd>O</kbd> to open
- [ ] See the modal slide in
- [ ] <kbd>escape</kbd> to close
- [ ] See the modal slide out

### Screenshots/Videos

Modal Animated

### DIFF
--- a/src/components/shared/Modal.css
+++ b/src/components/shared/Modal.css
@@ -7,7 +7,7 @@
   z-index: -1;
   top: 0;
   left: 0;
-  transition: z-index 0.4s ease-in-out;
+  transition: z-index 200ms;
 }
 
 .modal-wrapper.enabled {
@@ -21,7 +21,7 @@
   height: 230px;
   background-color: var(--theme-codemirror-gutter-background);
   transform: translateY(-250px);
-  transition: transform 0.2s ease-in-out;
+  transition: transform 150ms cubic-bezier(0.07, 0.95, 0, 1);
   box-shadow: 2px 4px 6px #dde1e4;
 }
 
--- a/src/components/shared/Modal.css
+++ b/src/components/shared/Modal.css
@@ -4,20 +4,13 @@
   justify-content: center;
   width: 100%;
   height: 100%;
-  z-index: 9;
+  z-index: -1;
   top: 0;
   left: 0;
+  transition: z-index 0.4s ease-in-out;
 }
 
-.modal-wrapper.entering,
-.modal-wrapper.exited {
-  pointer-events: none;
-  z-index: -1;
-}
-
-.modal-wrapper.entered,
-.modal-wrapper.exiting {
-  pointer-events: auto;
+.modal-wrapper.enabled {
   z-index: 9;
 }
 
--- a/src/components/shared/Modal.css
+++ b/src/components/shared/Modal.css
@@ -20,17 +20,12 @@
   width: 500px;
   height: 230px;
   background-color: var(--theme-codemirror-gutter-background);
+  transform: translateY(-250px);
   transition: transform 0.2s ease-in-out;
   box-shadow: 2px 4px 6px #dde1e4;
 }
 
-.modal.entering,
-.modal.exited {
-  transform: translateY(-250px);
-}
-
-.modal.entered,
-.modal.exiting {
+.modal.enabled {
   transform: translateY(30px);
 }
 
--- a/src/components/shared/Modal.css
+++ b/src/components/shared/Modal.css
@@ -1,0 +1,49 @@
+.modal-wrapper {
+  position: fixed;
+  display: flex;
+  justify-content: center;
+  width: 100%;
+  height: 100%;
+  z-index: 9;
+  top: 0;
+  left: 0;
+}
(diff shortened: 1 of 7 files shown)


wldcordeiro:
I'm not sure how to resolve these flow issues as the Modal doesn't really have any typings different from other components. Then as for yarn check, I can't seem to resolve that either. Seems like some dependency of storybook isn't happy with the react version we have?

codehag:
I will take a look again later, but from a quick reading it looks like you are expecting a single node as a child, and passing two.. i will take another look later in the evening.

src/components/shared/Modal.js
```diff
@@ -0,0 +1,68 @@
...
+  onClick(e: SyntheticEvent) {
+    e.stopPropagation();
+  }
+
+  render() {
+    const { enabled } = this.props;
+    return Slide({
+      in: enabled,
+      children: status => {
+        return dom.div(
```
jasonLaster: It might help if the wrapper and modal were siblings. This way the modal could animate onto the `<body>`.

src/components/shared/Modal.css
```diff
@@ -0,0 +1,47 @@
...
+
+.modal-wrapper.entering,
+.modal-wrapper.exited {
+  pointer-events: none;
+}
+
+.modal-wrapper.entered,
+.modal-wrapper.exiting {
+  pointer-events: auto;
+}
```
jasonLaster: we should avoid having a div overlaying the window at all times.

another strategy might be to use z-index: -1.

wldcordeiro: Good idea on moving the z-index.

wldcordeiro:
> I will take a look again later, but from a quick reading it looks like you are expecting a single node as a child, and passing two.. i will take another look later in the evening.

@codehag I was looking into how to make a function have variadic arity in Flow but I'm not finding a good resource. 🙁  I'll keep digging.

src/components/shared/Modal.js
wldcordeiro: Not sure I follow. Could you elaborate?

src/components/shared/Modal.css
```diff
@@ -0,0 +1,42 @@
+.modal-wrapper {
+  position: fixed;
+  display: flex;
+  justify-content: center;
+  width: 100%;
+  height: 100%;
+  z-index: -1;
+  top: 0;
+  left: 0;
+  transition: z-index 0.4s ease-in-out;
```
wldcordeiro: I added this slightly longer transition to the z-index change to make sure we have enough time for the animation to play.

codehag:
fixed up the flow types, was a tough case!

jasonLaster:
```
➜  debugger.html yarn check
yarn check v0.24.5
warning "chokidar#fsevents#node-pre-gyp@^0.6.36" could be deduped from "0.6.36" to "node-pre-gyp@0.6.36"
error "react-modal#react-dom-factories#react@^15.5.4" doesn't satisfy found match of "react@15.3.2"
error Found 1 errors.
info Visit https://yarnpkg.com/en/docs/cli/check for documentation about this command.
```

we might need a lower version of transition-group

wldcordeiro:
@codehag thanks for that! @jasonLaster the transition-group package.json seems to indicate we would have a good version. https://github.com/reactjs/react-transition-group/blob/master/package.json#L44 We could try rolling back to a different version but the 1.x has a slightly different API so I'd need to modify a bit further.

jasonLaster:
Let's try a 1.x if that makes yarn happy. Basically downgrade do yarn
check, then maybe move forward

wldcordeiro:
I tried dropping to 1.x and still saw the same error. 🙁

jasonLaster:
What do we need this dependency to do? Why not just add a class that has
the transition?

codehag:
The issue with the incompatible react dependency was coming from a bump to percy storybook. I have a fix

codehag:
This has been rebased to remove the offending yarn.lock inclusion -- if you modify it please pull this branch first and replace your local branch

wldcordeiro:
@codehag I pulled changes, rebased onto next and had to revert the yarn.lock entirely because even removing my dep and regenerating caused the error in `yarn check`

codehag:
@wldcordeiro try this:
- go to master
- delete your local branch
- chechout your local branch again

this will give you a clean version

I suspect that you still have the old yarn.lock file and that is what is causing the troubles. It is building on CI now, and also locally for me

wldcordeiro:
@codehag @jasonLaster with the changes you've made I think this is ready. I have nothing else I need to modify.

wldcordeiro:
Okay I think now the rebase madness is done.

wldcordeiro:
The commit log here leaves evidence that I'm a time traveler. 😆

Issue #3383, opened 2017-07-20
[Breakpoints] width is set to 1px


jasonLaster:
It looks like there is a race condition with the editor that occassionally sets breakpoints to a really small width:

![Pasted Image 1.tiff](https://shipusercontent.com/8d112d1551f659729970daf2416fee23/Pasted%20Image%201.tiff)

jasonLaster:
This should be fixed by the editor.state bug

codehag:
We have solved one part of the issue, associated with .wasm files. However we still have this issue when switching the panel orientation. This may be an issue in code mirror itself

STR:

- Open debugger while running in the panel
- Add a breakpoint while the debugger is in the bottom panel
- switch the debugger position to being a side panel
- Breakpoint will be a baby bps

Gif:
recordit.co/Luuhcd6uKu

CSS issues:
before Rerender
<img width="443" alt="screen shot 2017-07-24 at 11 14 56 am" src="https://user-images.githubusercontent.com/26968615/28516428-8141a44c-7061-11e7-830b-f2bde7bb4b0a.png">
After rerender
<img width="188" alt="screen shot 2017-07-24 at 11 13 47 am" src="https://user-images.githubusercontent.com/26968615/28516421-78f8cd56-7061-11e7-93f3-014dfbfd7cae.png">

codehag:
Should now be fixed

jasonLaster:
we can use this min-width fix if we need it: https://github.com/devtools-html/debugger.html/pull/3419

jasonLaster:
I'm re-opening this issue because we're seeing breakpoints collapse to 1px when the layout changes from horizontal to vertical.

jasonLaster:
@yurydelendik added a weakmap here that keeps us from doing unnecessary updates:

https://github.com/devtools-html/debugger.html/pull/3764/commits/c308ceb2851d561b9987d70aa42a2c5cbce887b0

we might want to revisit later.

PR #3417, opened 2017-07-23
Disable updating line number formats

jasonLaster:
Associated Issue: #3383

### Summary of Changes

* Fixes a situation where toggling sources causes 1px BPs

[STR] 
1. navigate to https://devtools-html.github.io/debugger-examples/examples/todomvc/
2. add a bp in todo-view.js and todo.js
3. reload (view a bp in todo.js)
4. click on the todo-view.js in the bp panel

AR: see the BP in todo-view
ER: see a 1px bp

### DIFF
--- a/configs/development.json
+++ b/configs/development.json
@@ -66,6 +66,10 @@
     "highlightScopeLines": {
       "label": "Highlight the lines in scope",
       "enabled": false
+    },
+    "wasm": {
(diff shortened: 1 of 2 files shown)


codehag:
This looks good, it fixes the baby bps from WASM -- but we still have an issue with baby bps when switching panels. I will merge this and comment on the issue

PR #3419, opened 2017-07-24
bugfix: [Breakpoints] width is set to 1px

codehag:
Associated Issue: #3383

### Summary of Changes

* force new-breakpoint to have width 29px

babybps

### DIFF
--- a/src/components/Editor/Editor.css
+++ b/src/components/Editor/Editor.css
@@ -87,7 +87,7 @@ html[dir="rtl"] .editor-mount {
 }
 
 .editor.new-breakpoint {
-  width: 29px;
+  min-width: 29px;
   position: absolute;
 }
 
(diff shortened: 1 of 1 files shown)


codehag:
NOTE: this is a pr into master, if it is better to do it into next, let me know.

src/components/Editor/Editor.css
```diff
@@ -86,6 +86,11 @@ html[dir="rtl"] .editor-mount {
   right: -4px;
 }
 
+.editor.new-breakpoint {
+  width: 29px;
+  position: absolute;
+}
```
jasonLaster: we can't assume it's 29px. it's possible that the gutter will be wider because the file has 3/4/5 digit line numbers

jasonLaster: We should figure out why the gutter is set to width `1px` https://github.com/codemirror/CodeMirror/blob/master/src/display/update_line.js#L112-L121

My guess is that we're updating the gutter while the element is detatched or the document is not visible.

Issue #3411, opened 2017-07-21
[console] link does not select location when debugger is closed


bzbarsky:
I filed this as https://bugzilla.mozilla.org/show_bug.cgi?id=1383209 because I can attach files there, but since I'm not sure anyone is looking at bugs reported there...

jasonLaster:
Thanks for filing

---

STEPS TO REPRODUCE:

1)  Load the testcase I'm about to attach in a clean profile.
2)  Cmd+Option+K to open the console.
3)  Click the link on the "ReferenceError" line (linking to line 1201 of the JS file).

ACTUAL RESULTS: Debugger opens, loads JS file, does NOT scroll to the line.

EXPECTED RESULTS: Scroll to the line and select it.  This works fine with the old debugger frontend.

ADDITIONAL INFORMATION: If the debugger is already open when loading the testcase, then scrolling to the line does work, but it doesn't get highlighted (unlike with the old debugger frontend).

This bug makes debugging errors really annoying.

codehag:
In the `next` branch, we have a new issue -- the debugger is not opened, and instead opens a new tab with the source and the correct line highlighted (just like the old debugger)

codehag:
Not sure if I am going in the right direction, but when I test the str, i sometimes get the new tab behavior and the following error:

```
*************************
A coding exception was thrown in a Promise resolution callback.
See https://developer.mozilla.org/Mozilla/JavaScript_code_modules/Promise.jsm/Promise

Full message: TypeError: this._extraPools is null
Full stack: addActorPool@resource://gre/modules/commonjs/toolkit/loader.js -> resource://devtools/server/main.js:1540:5
get threadLifetimePool@resource://gre/modules/commonjs/toolkit/loader.js -> resource://devtools/server/actors/script.js:498:7
onSource/<@resource://gre/modules/commonjs/toolkit/loader.js -> resource://devtools/server/actors/source.js:493:11
process@resource://gre/modules/Promise.jsm -> resource://gre/modules/Promise-backend.js:922:23
walkerLoop@resource://gre/modules/Promise.jsm -> resource://gre/modules/Promise-backend.js:806:7
Promise*scheduleWalkerLoop@resource://gre/modules/Promise.jsm -> resource://gre/modules/Promise-backend.js:739:11
schedulePromise@resource://gre/modules/Promise.jsm -> resource://gre/modules/Promise-backend.js:770:7
[...]

PR #3431, opened 2017-07-24
Scroll onload

jasonLaster:
Associated Issue: #3411 

### Summary of Changes

* fixes line scrolling by appropriately setting the pending jump line
* we'll update the webconsole test when we make a new release


codehag:
Thanks for picking this up! It isn't working for me when I follow the STR, what does your setup look like?

PR #3434, opened 2017-07-24
Fix framework grouping collapsing

jasonLaster:

### Summary of Changes

* fixes collapsing for both frames that are not grouped

### DIFF
--- a/src/components/SecondaryPanes/Frames/index.js
+++ b/src/components/SecondaryPanes/Frames/index.js
@@ -45,15 +45,6 @@ class Frames extends Component {
     showAllFrames: boolean
   };
 
-  collapseFrames(frames) {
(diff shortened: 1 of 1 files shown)


jasonLaster:
I'd like to add a unit test here, but struggled with it

src/components/SecondaryPanes/Frames/index.js
```diff
@@ -154,7 +154,7 @@ class Frames extends Component {
       ? L10N.getStr("callStack.collapse")
       : L10N.getStr("callStack.expand");
 
-    frames = collapseFrames(frames);
+    frames = this.collapseFrames(frames);
```
codehag: Was the issue that if frameworkGroupingOn was not checked, collapse frame would be run with unexpected results?

PR #3435, opened 2017-07-24
Search redux on textload

bomsy:
As a follow up to searching using redux

### Summary of Changes

* We now also the search new sources as they get loaded.


src/utils/search/project-search.js
```diff
@@ -30,12 +30,3 @@ export function searchSource(source, queryText) {
   matches = [].concat(...matches);
   return matches;
 }
-
-export default function searchSources(query, sources) {
```
bomsy: moved to actions, test to be added

Issue #3439, opened 2017-07-25
[sourcemaps] breakpoints disappear on sourcemapped sources


codehag:
Tested with this example: http://wbamberg.github.io/example-websites/source-mapping/index.html
It is also happening with this example: https://devtools-html.github.io/debugger-examples/examples/sequence-print/sequence_print.html

STR:
1. select an original source
2. add a breakpoint within the body of the file (outside of a function) -- should work
3. add a breakpoint within a function, the breakpoint will appear momentarily and disappear. It will show up in the list of breakpoints and will work. what it should do is stay

looks like a visual issue. Happening on both next and master branches

Gif: 

![disappearingbps](https://user-images.githubusercontent.com/26968615/28563640-d76059de-7126-11e7-8376-c0a98dd1adcb.gif)

codehag:
After investigating, I found that this was from column breakpoints being enabled. I made a fix, not sure it is the best one -- but I also added tests so that this doesn't come up in the future. The tests were failing and are now green, that's probably the best bit of the pr

PR #3440, opened 2017-07-25
breakpoints disappear on sourcemapped sources

codehag:
Associated Issue: #3439 

When I started working on this issue, I wasn't sure what was wrong and it looked quite concerning, so i went ahead and started on it.

It looks like the error was coming from column breakpoints being ignored when column bps are enabled, and since all breakpoints which come from original sources have columns, these were being ignored. I removed the filter, as the breakpoints render correctly without it.

Caveat: Not sure if this is how it should work, or if column breakpoints are enabled by default.

I tested in a couple of spots:

TODO mvc: https://devtools-html.github.io/debugger-examples/examples/todomvc/
Sourcemapped files: http://wbamberg.github.io/example-websites/source-mapping/index.html
messy sourcemapped files: https://devtools-html.github.io/debugger-examples/examples/sequence-print/sequence_print.html

These three were also tested in the panel

I added a test to illustrate the case.

### Summary of Changes

* remove filtering of column bps from Breakpoints Component

### Test Plan

Add tests for both column and non column breakpoints. These were the bulk of the work, and are probably useful even if the fix isn't

### DIFF
--- a/src/components/Editor/tests/Breakpoints.js
+++ b/src/components/Editor/tests/Breakpoints.js
@@ -1,0 +1,46 @@
+import React from "react";
+import { shallow } from "enzyme";
+import Breakpoints from "../Breakpoints";
+import * as I from "immutable";
+
+const BreakpointsComponent = React.createFactory(Breakpoints.WrappedComponent);
+
+function generateDefaults(overrides) {
+  const sourceId = "server1.conn1.child1/source1";
+  const matchingBreakpoints = { id1: { location: { sourceId } } };
+
+  return {
+    selectedSource: { sourceId, get: () => false },
+    editor: {
+      codeMirror: {
+        setGutterMarker: jest.fn()
+      }
+    },
+    breakpoints: I.Map(matchingBreakpoints),
+    ...overrides
+  };
+}
+
+function render(overrides = {}) {
+  const props = generateDefaults(overrides);
+  const component = shallow(new BreakpointsComponent(props));
+  return { component, props };
+}
+
+describe("Breakpoints Component", () => {
+  it("should render breakpoints without columns", async () => {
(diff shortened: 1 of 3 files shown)


src/components/Editor/tests/Breakpoints.js
```diff
@@ -0,0 +1,46 @@
...
+    const sourceId = "server1.conn1.child1/source1";
+    const breakpoints = I.Map({ id1: { location: { sourceId } } });
+
+    const { component, props } = render({ breakpoints });
+    expect(component.find("Breakpoint").length).toBe(props.breakpoints.size);
+  });
+
+  it("should render breakpoints with columns", async () => {
+    const sourceId = "server1.conn1.child1/source1";
+    const breakpoints = I.Map({ id1: { location: { column: 2, sourceId } } });
```
codehag: this was failing

src/components/Editor/tests/Breakpoints.js
```diff
@@ -0,0 +1,46 @@
...
+
+    const { component, props } = render({ breakpoints });
+    expect(component.find("Breakpoint").length).toBe(props.breakpoints.size);
```
jasonLaster: lets also have a snapshot here so we can get some regression testing on the children

src/components/Editor/tests/Breakpoints.js
```diff
@@ -0,0 +1,46 @@
...
+
+function generateDefaults(overrides) {
+  const sourceId = "server1.conn1.child1/source1";
+  const matchingBreakpoints = { id1: { location: { sourceId } } };
+
+  return {
+    selectedSource: { sourceId, get: () => false },
+    editor: {
+      codeMirror: {
+        setGutterMarker: jest.fn()
```
jasonLaster: nice

PR #3441, opened 2017-07-25
Update webpack to version 3.3.0

zaggy:
Associated Issue: #2418 and others

This updates webpack to version 3.3.0.

### Test Plan

- Clone the changes from PR 505 from devtools-core
- Go to packages/devtools-launchpad and `npm pack`
- Do `yarn add file:<path/to/packed/tgz>`
- Do `yarn start`


postcss.config.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
+  plugins: [require("postcss-bidirection"), require("autoprefixer")]
```
jasonLaster: I bet we can add postcss.config.js to the list of exported files and avoid this:

https://github.com/devtools-html/devtools-core/blob/master/packages/devtools-launchpad/package.json#L113

jasonLaster:
screen shot 2017-08-01 at 6 50 44 pm
screen shot 2017-08-01 at 6 50 21 pm

circle.yml
```diff
@@ -31,7 +31,7 @@ test:
     - ./bin/run-mochitests-docker
   post:
     - yarn flow-coverage -- -t json -o $CIRCLE_TEST_REPORTS/flow-coverage
-    - yarn check
+    - yarn check --integrity
```
jasonLaster: calling this out as it's a friendlier check.

we might want to go back to yarn check after we jump to react 15.6 or 16. At the moment we're so far back it's hurting us

PR #3454, opened 2017-07-26
Editor state (part 3)

jasonLaster:
### Summary of Changes

We were blowing up when the function search modal was open and highlighting text and you changed layouts. On the bright side, i found this today because i use function search ALL the time now.


jasonLaster:
I like the HoC, but I wonder if I'd expect to check it in the imports. Perhaps we should wrap connect

`connect(,,)(inectEditor(Editor))`

PR #3458, opened 2017-07-26
remove react-dom-factories

codehag:
something _was_ requireing react-dom-factories which is only available from 1.5.4 it looks like. and the lock file was somehow added to the repo

react-dom-factories is introduced as a way to handle legacy dom factories for react 16.0.0 onwards: https://www.npmjs.com/package/react-dom-factories


jasonLaster:
oh great!

PR #3465, opened 2017-07-26
Watch Expression the remove button is often hidden by the scroll bar

codehag:
Associated Issue: #3465 

### Summary of Changes

* added a util that we need anyway
* cleaned up the css a bit (had an unused flexbox
* adjusted html elements to be a bit more semantically informative

floats

### DIFF
--- a/src/components/App.css
+++ b/src/components/App.css
@@ -37,3 +37,11 @@
   margin-top: 25px;
   margin-right: 20px;
 }
+
+/* Utils */
+.vertical-center {
+  top: 50%;
+  -webkit-transform: translateY(-50%);
+  -ms-transform: translateY(-50%);
+  transform: translateY(-50%);
+}
--- a/src/components/App.css
+++ b/src/components/App.css
@@ -38,10 +38,3 @@
   margin-right: 20px;
 }
 
-/* Utils */
-.vertical-center {
-  top: 50%;
-  -webkit-transform: translateY(-50%);
-  -ms-transform: translateY(-50%);
-  transform: translateY(-50%);
-}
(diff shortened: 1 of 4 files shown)


src/components/App.css
```diff
@@ -37,3 +37,11 @@
   margin-top: 25px;
   margin-right: 20px;
 }
+
+/* Utils */
+.vertical-center {
+  top: 50%;
+  -webkit-transform: translateY(-50%);
+  -ms-transform: translateY(-50%);
+  transform: translateY(-50%);
```
jasonLaster: we can drop the prefixes because we have autoprefixer on:

this is the output w/o it
```css
/* Utils */
.vertical-center {
  top: 50%;
  -webkit-transform: translateY(-50%);
          transform: translateY(-50%);
}
```

jasonLaster: it feels funny to have the class without a user though...

src/components/SecondaryPanes/Expressions.css
```diff
@@ -49,10 +54,14 @@
   background-color: var(--search-overlays-semitransparent);
 }
 
-.expression-container .close-btn {
+.expression-container__close-btn {
   position: absolute;
-  offset-inline-end: 6px;
-  top: 6px;
+  right: 0px;
```
jasonLaster: lets keep `offset-inline-end`

src/components/App.css
codehag: yeah we can remove this util for now... its not used
would be useful later on, i noticed we are using flexbox and other solutions for centering and that usually doesn't give a true center. originally I added this thinking I could center the `x` and then i realized it was a managed tree component ;P

Issue #3426, opened 2017-07-24
[CommandBar] stops being sticky


jasonLaster:
It looks like the command bar can scroll out of view when the debugger is in vertical layout.

We should fix this by making sure the command bar is sticky

```diff
diff --git a/src/components/SecondaryPanes/CommandBar.css b/src/components/SecondaryPanes/CommandBar.css
index bb80587..336c4a8 100644
--- a/src/components/SecondaryPanes/CommandBar.css
+++ b/src/components/SecondaryPanes/CommandBar.css
@@ -4,6 +4,7 @@
   display: flex;
   height: 30px;
   overflow: hidden;
+  /* this should do it...*/
   position: sticky;
   top: 0;
   z-index: 1;
diff --git a/src/components/SecondaryPanes/index.js b/src/components/SecondaryPanes/index.js
index ed7d5a8..956c010 100644
--- a/src/components/SecondaryPanes/index.js
+++ b/src/components/SecondaryPanes/index.js
@@ -219,6 +219,7 @@ class SecondaryPanes extends Component {
       {
         className: "secondary-panes"
       },
+      // we might need to fix something here...
       CommandBar(),
       this.props.horizontal
         ? this.renderHorizontalLayout()
```

![](http://g.recordit.co/GXGY2plgiL.gif)

jasonLaster:
![](http://g.recordit.co/gRe69H1SxU.gif)

jasonLaster:
this is a bad fix https://gist.github.com/97dcee6db4dae1daa098ba5bd24cc466

but demonstrates the problem is w/ long content

jasonLaster:
ah, `height: fit-content` where supported fixes the problem with commandbar. :wink:
http://caniuse.com/#feat=intrinsic-width
`.secondary-panes { height: fit-content; overflow-y: scroll; }`

Garbee:
So, that "patch" I thought over... Leads to broken output. Also, since Sticky is still "in development" as a spec, this particular case *seems* to not be specified. If you check Note 3 in [the sticky spec](https://www.w3.org/TR/css-position-3/#sticky-pos) it reads:

> Say what happens if it already overflows the containing block

I think this is the case we're hitting. Since the height is already "overflowing" the boundary, the engines aren't sure what to do with a sticky node. Gecko is choosing to stick it, what I'd say is correct since that provides the functionality we want. While Blink is choosing to keep the sticky node as a relative item constantly.

[...]

PR #3466, opened 2017-07-26
Fix commandbar stopping 

Garbee:
Associated Issue: #3426 

### Summary of Changes

* Overflow the secondary panel that contains the commandbar

### Test Plan

Pause on a breakpoint. Open up window/this vars until you have plenty of place to scroll. Scroll beyond the height of your viewport.

Notice the commandbar will now keep stuck all the way to the bottom.

### DIFF
--- a/src/components/SecondaryPanes/SecondaryPanes.css
+++ b/src/components/SecondaryPanes/SecondaryPanes.css
@@ -5,6 +5,17 @@
   white-space: nowrap;
 }
 
+/*
+  We apply overflow to the container with the commandbar.
+  This allows the commandbar to remain fixed when scrolling
+  until the content completely ends. Not just the height of
+  the wrapper.
+  Ref: https://github.com/devtools-html/debugger.html/issues/3426
+*/
+.secondary-panes--sticky-commandbar {
+  overflow-y: scroll;
+}
+
 .secondary-panes .accordion {
   flex: 1 0 auto;
 }
--- a/src/components/SecondaryPanes/index.js
+++ b/src/components/SecondaryPanes/index.js
@@ -217,7 +217,7 @@ class SecondaryPanes extends Component {
   render() {
     return dom.div(
       {
-        className: "secondary-panes"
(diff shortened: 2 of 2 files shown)


Garbee:
Yulia and I were discussing this in the Slack channel just now as well. She has an [alternative approach](https://devtools-html.slack.com/files/yulia/F6E3LDUUU/foo.diff) worth looking into as well which removes sticky entirely and applies some minor restructuring to get the same effect.

Both are completely valid solutions. The only worry about `sticky` is if the spec and/or implementations changes in the future then this *could* break. But I find that personally doubtful since the case we are hitting is an inner niche-case with the spec and the overflow sets the wrapper up around that flaw.

Either way works though. So whatever is desired I can amend the PR to implement.

src/components/SecondaryPanes/SecondaryPanes.css
```diff
@@ -5,6 +5,17 @@
...
 }
 
+/*
+  We apply overflow to the container with the commandbar.
+  This allows the commandbar to remain fixed when scrolling
+  until the content completely ends. Not just the height of
+  the wrapper.
+  Ref: https://github.com/devtools-html/debugger.html/issues/3426
+*/
+.secondary-panes--sticky-commandbar {
```
Garbee: I did this as an auxiliary class just to be certain we aren't affecting other nodes by accident. Just in case.

jasonLaster: lol - this could have easily been `. secondary-panes.sticky-commandbar`, but you wanted to go all BEM on us :)

I'm okay with this, but just being my morning snarky self :)

Garbee: haha, second nature to me at this point. ;)

Garbee:
@codehag Could you file an issue about that? Preferably with a gif or some video on what's going on. I can take a look at it tomorrow or over the weekend.

Issue #3421, opened 2017-07-24
[next] TypeError: client.sourceContents is not a function


codehag:
In the next branch, when we run the tests we get the following error
```
(node:13687) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 2): TypeError: client.sourceContents is not a function
(node:13687) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
(node:13687) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 7): TypeError: client.sourceContents is not a function
(node:13687) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 8): TypeError: client.sourceContents is not a function
(node:13687) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 9): TypeError: client.sourceContents is not a function
(node:13687) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 10): TypeError: client.sourceContents is not a function
(node:13687) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 15): TypeError: client.sourceContents is not a function
(node:13687) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 16): TypeError: client.sourceContents is not a function
(node:13687) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 17): TypeError: client.sourceContents is not a function
(node:13687) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 18): TypeError: client.sourceContents is not a function
 PASS  src/actions/tests/tabs.jsectionWarning: Unhandled promise rejection (rejection id: 23): TypeError: client.sourceContents is not a function
```

We should investigate where this is coming from

jasonLaster:
ahh - i bet this is coming from my tabs test move!

https://github.com/devtools-html/debugger.html/pull/3413

PR #3472, opened 2017-07-27
Bugfix: TypeError: client.sourceContents is not a function

codehag:
Associated Issue: #3421

### DIFF
--- a/src/actions/tests/helpers/breakpoints.js
+++ b/src/actions/tests/helpers/breakpoints.js
@@ -1,16 +1,3 @@
-import { makeLocationId } from "../../../utils/breakpoint";
-
-const sourceFixtures = {
-  foo1: {
-    source: "function foo1() {\n  return 5;\n}",
-    contentType: "text/javascript"
-  },
-  foo2: {
-    source: "function foo2(x, y) {\n  return x + y;\n}",
-    contentType: "text/javascript"
-  }
-};
-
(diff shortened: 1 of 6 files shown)


src/actions/tests/helpers/breakpoints.js
```diff
@@ -1,16 +1,3 @@
-import { makeLocationId } from "../../../utils/breakpoint";
-
-const sourceFixtures = {
```
codehag: moving this to threadClient.js so that later we can consolidate our threadclient

src/actions/tests/helpers/threadClient.js
```diff
@@ -0,0 +1,116 @@
+import { makeLocationId } from "../../../utils/breakpoint";
+
+const sourceFixtures = {
```
codehag: Most of the deleted code has been moved here

src/actions/tests/helpers/threadClient.js
```diff
@@ -0,0 +1,116 @@
...
+            contentType: "text/javascript"
+          });
+          break;
+        case "barfoo.js":
+          resolve({
+            source: "function bar() {\n  return 3;\n}",
+            contentType: "text/javascript"
+          });
+          break;
+        case "foo.js":
```
codehag: added three cases

PR #3476, opened 2017-07-27
guard for long lines

codehag:


src/utils/search/get-matches.js
```diff
@@ -15,7 +20,8 @@ export default function getMatches(
   const lines = text.split("\n");
   for (let i = 0; i < lines.length; i++) {
     let singleMatch;
-    while ((singleMatch = regexQuery.exec(lines[i])) !== null) {
+    const line = guardLineLength(lines[i]);
+    while ((singleMatch = regexQuery.exec(line)) !== null) {
```
jasonLaster: I propose that we go with a slightly different option. Instead of only searching a portion of the string, lets stop searching entirely.

```js
if (!checkLineLengths(lines)) {
  return []
}

function checkLineLengths(lines) {
  return !(lines.any(line => line.length >= 100000))
}
```

What's the advantage to this approach? If we bail early the UX will be:
1. codemirror shows the current match and next/previous still work
2. the search summary is blank (old debugger didn't show the sumary 1 or x)
3. it's ofcourse fast
4. it's honest - i'd hate to say there are 3 matches and actually have 5 in codemirror


we can come back and support summaries in this case later

codehag: I understand what you mean. I am just afraid that it will feel broken and frustrating for the user. The best thing we could do is give back an error message saying "line too long". I will throw that together and lets see what happens.

codehag:
I didn't have time to implement the error message so lets leave it like this for now

Issue #3430, opened 2017-07-24
[SourceTree] fails on https://facebook.github.io/react/


clarkbw:
Here's the Steps to Reproduce:

1. Go to https://facebook.github.io/react/
2. Open debugger

I was just running Nightly via web-ext to test something with React DevTools and hit this error with the debugger.

The debugger comes up blank and spits out the following error.  Here's the location of the source of the error: https://github.com/devtools-html/debugger.html/blob/17b3be6d327a8f72e5f4c0b3478487629db77af7/src/utils/sources-tree.js#L264

```
TypeError: nodes.findIndex is not a function                     Promise-backend.js:925
Stack trace:
determineFileSortOrder@resource://gre/modules/commonjs/toolkit/loader.js -> resource://devtools/client/debugger/new/debugger.js:22749:11
_loop@resource://gre/modules/commonjs/toolkit/loader.js -> resource://devtools/client/debugger/new/debugger.js:22692:18
addToTree@resource://gre/modules/commonjs/toolkit/loader.js -> resource://devtools/client/debugger/new/debugger.js:22713:6
createTree@resource://gre/modules/commonjs/toolkit/loader.js -> resource://devtools/client/debugger/new/debugger.js:22818:6
SourcesTree@resource://gre/modules/commonjs/toolkit/loader.js -> resource://devtools/client/debugger/new/debugger.js:50808:20
_constructComponentWithoutOwner@resource://gre/modules/commonjs/toolkit/loader.js -> resource://devtools/client/shared/vendor/react.js:6535:16
_constructComponent@resource://gre/modules/commonjs/toolkit/loader.js -> resource://devtools/client/shared/vendor/react.js:6522:14
mountComponent@resource://gre/modules/commonjs/toolkit/loader.js -> resource://devtools/client/shared/vendor/react.js:6425:16
mountComponent@resource://gre/modules/commonjs/toolkit/loader.js -> resource://devtools/client/shared/vendor/react.js:14744:18
performInitialMount@resource://gre/modules/commonjs/toolkit/loader.js -> resource://devtools/client/shared/vendor/react.js:6608:18
mountComponent@resource://gre/modules/commonjs/toolkit/loader.js -> resource://devtools/client/shared/vendor/react.js:6495:16
mountComponent@resource://gre/modules/commonjs/toolkit/loader.js -> resource://devtools/client/shared/vendor/react.js:14744:18
[...]

PR #3477, opened 2017-07-27
Support complex file trees

jasonLaster:
Associated Issue: #3430

### Summary of Changes

* bail early when we get to a case we can't handle in the source tree
* screenshot shows the files in source search, but not in the source tree

### Test Plan

* add a unit test to verify the behavior

### Screenshot

screen shot 2017-07-27 at 11 38 32 am

### DIFF
--- a/src/utils/sources-tree.js
+++ b/src/utils/sources-tree.js
@@ -6,7 +6,6 @@
  */
 
 import { parse } from "url";
-import assert from "./DevToolsUtils";
 import { isPretty } from "./source";
 import merge from "lodash/merge";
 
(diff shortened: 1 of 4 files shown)


src/utils/sources-tree.js
```diff
@@ -203,7 +203,11 @@ function addToTree(tree: any, source: TmpSource, debuggeeUrl: string) {
     //
     // TODO: Be smarter about this, which we'll probably do when we
     // are smarter about folders and collapsing empty ones.
-    assert(nodeHasChildren(subtree), `${subtree.name} should have children`);
+
+    if (!nodeHasChildren(subtree)) {
+      return;
+    }
```
jasonLaster: this is the only code change

src/utils/sources-tree.js
```diff
@@ -172,7 +172,7 @@ function isDirectory(url: Object) {
  * @memberof utils/sources-tree
  * @static
  */
-function addToTree(tree: any, source: TmpSource, debuggeeUrl: string) {
+export function addToTree(tree: any, source: TmpSource, debuggeeUrl: string) {
```
jasonLaster: switched to explicit exports

codehag: :+1:

src/utils/sources-tree.js
codehag: :+1:

Issue #3461, opened 2017-07-26
[WatchExpressions] values should be cleared upon resume


clarkbw:
**Coming from [bug 1379978](https://bugzilla.mozilla.org/show_bug.cgi?id=1379978)**

[Steps to reproduce]:
1. Open Firefox and go to https://devtools-html.github.io/debugger-examples/examples/todomvc/
2. Open Debugger.
3. Add Watch Expression `this`
4. Open (cmd+p) todo-view.js
5. Create breakpoint on line line 34
6. Add new todo item into page, see debugger paused and `this` value becomes `Object`
7. Click on "resume" from the toolbar.

The `this` value should clear after resuming.

PR #3484, opened 2017-07-27
Clear expressions

jasonLaster:
Associated Issue: #3461

### DIFF
--- a/src/actions/pause.js
+++ b/src/actions/pause.js
@@ -25,12 +25,12 @@ type CommandType = { type: string };
  */
 export function resumed() {
   return ({ dispatch, client }: ThunkArgs) => {
-    // dispatch(evaluateExpressions(null));
-
-    return dispatch({
+    dispatch({
       type: "RESUME",
       value: undefined
     });
+
+    dispatch(evaluateExpressions(null));
   };
 }
 
--- a/src/test/mochitest/browser_dbg-expressions.js
+++ b/src/test/mochitest/browser_dbg-expressions.js
@@ -20,6 +20,10 @@ function getValue(dbg, index) {
   return findElement(dbg, "expressionValue", index).innerText;
 }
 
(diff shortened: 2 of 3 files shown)


src/actions/pause.js
```diff
@@ -25,12 +25,12 @@ type CommandType = { type: string };
...
   return ({ dispatch, client }: ThunkArgs) => {
-    // dispatch(evaluateExpressions(null));
-
-    return dispatch({
+    dispatch({
       type: "RESUME",
       value: undefined
     });
+
+    dispatch(evaluateExpressions());
```
codehag: a better pattern would be listening on the expressions reducer for a "resume" action. reusing evaluateExpressions here intertwines two responsibilities in an action.

src/actions/expressions.js
```diff
@@ -90,7 +90,7 @@ export function deleteExpression(expression: Expression) {
  * @param {number} selectedFrameId
  * @static
  */
-export function evaluateExpressions(frameId: frameIdType) {
+export function evaluateExpressions(frameId?: frameIdType) {
```
codehag: Its better if we keep evaluate expression to having just one responsibility. It looks like up to this point we always had a frameID -- if this is true, then we can remove the if block here

codehag:
So I made the change, and since we are saving on the evaluation as a result the code performs better. I noticed some breakpoints weirdness while making the change... The breakpoints would start appearing in the wrong places. This was fixed when i cleared my localStorage... but the only thing i touched was the reducer and i removed a dispatch... this makes me a bit nervous because it feels a bit like this: 

http://media.riffsy.com/images/9d9ff2fa36d9eaf1b5e7dacf8a177f67/tenor.gif

should we update the schema just in case?

codehag:
The weirdness with breakpoints is not limited to this branch, so i don't think we need to worry about it here

Issue #3495, opened 2017-07-28
[SourceSearch] should focus on open


jasonLaster:
STR:
1. cmd+p

AR: input is focused
ER: input is not focused :/

This commit https://github.com/devtools-html/debugger.html/commit/829954d7ba56e95f6ae28cd4c377accc809a01d9 unintentionally regressed source search focusing on open.

This test change demonstrates the regression:

```diff
diff --git a/src/test/mochitest/browser_dbg-searching.js b/src/test/mochitest/browser_dbg-searching.js
index dd25e2b..dde5f87 100644
--- a/src/test/mochitest/browser_dbg-searching.js
+++ b/src/test/mochitest/browser_dbg-searching.js
@@ -7,7 +7,7 @@ add_task(function* () {

   pressKey(dbg, "sourceSearch");
   yield waitForElement(dbg, "input");
-  findElementWithSelector(dbg, "input").focus();
+  // findElementWithSelector(dbg, "input").focus();
   type(dbg, "sw");
   pressKey(dbg, "Enter");

@@ -17,7 +17,7 @@ add_task(function* () {

   // 2. arrow keys and check to see if source is selected
   pressKey(dbg, "sourceSearch");
-  findElementWithSelector(dbg, "input").focus();
+  // findElementWithSelector(dbg, "input").focus();
   type(dbg, "sw");
   pressKey(dbg, "Down");
   pressKey(dbg, "Enter");
(END)
```

I found this code smell too:

```diff
diff --git a/src/components/ProjectSearch/index.js b/src/components/ProjectSearch/index.js
index d012276..69c89cb 100644
--- a/src/components/ProjectSearch/index.js
+++ b/src/components/ProjectSearch/index.js
@@ -22,7 +22,6 @@ const SourceSearch = createFactory(_SourceSearch);

 class ProjectSearch extends Component {
   state: Object;
-  onEscape: Function;
   close: Function;
   toggleProjectTextSearch: Function;
   toggleSourceSearch: Function;
@@ -61,7 +60,6 @@ class ProjectSearch extends Component {
       L10N.getStr("sources.search.alt.key")
     ];
     searchKeys.forEach(key => shortcuts.off(key, this.toggleSourceSearch));
-    shortcuts.off("Escape", this.onEscape);
   }
```

Another weird outcome of the modal being open is that "Escape" is always handled in SymbolModal/handleClose as opposed to SourceSearch/onEscape...

PR #3505, opened 2017-07-31
Fix focus issues

wldcordeiro:
Associated Issue: #3495 

### Summary of Changes

* Fixes the focus issue across all search inputs. They now take focus when enabled.
* Bump yarn.lock to match updated devtools-reps version
* CSS got reformatted by prettier?

### Test Plan

Example test plan:

- [x] Command-P opens the panel with focus
- [x] Command-F opens search panel with focus
- [x] Comman-Shift-O opens symbol search with focus

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,6 +69,7 @@
     "pretty-fast": "^0.2.1",
     "react": "=15.3.2",
     "react-dom": "=15.3.2",
+    "react-transition-group": "^2.2.0",
     "redux-saga": "^0.15.4",
     "reselect": "^3.0.0",
     "svg-inline-react": "^1.0.2",
--- a/src/components/ProjectSearch/TextSearch.js
+++ b/src/components/ProjectSearch/TextSearch.js
@@ -58,9 +58,6 @@ export default class TextSearch extends Component {
 
   componentDidMount() {
     const shortcuts = this.context.shortcuts;
-    if (this.refs.searchInput) {
-      this.refs.searchInput.refs.input.focus();
-    }
     shortcuts.on("Enter", this.onEnterPress);
   }
 
--- a/src/components/SymbolModal.js
+++ b/src/components/SymbolModal.js
@@ -109,8 +109,12 @@ class SymbolModal extends Component {
       scrollList(this.refs.resultList.refs, this.state.resultsIndex);
     }
 
-    if (this.refs.searchInput && this.refs.searchInput.refs.input) {
-      this.refs.searchInput.refs.input.focus();
+    if (
+      this.refs.searchInput &&
+      this.refs.searchInput.input &&
+      this.props.enabled
+    ) {
+      this.refs.searchInput.input.focus();
     }
 
     if (!prevProps.enabled && this.props.enabled) {
--- a/src/components/SymbolModal.js
+++ b/src/components/SymbolModal.js
@@ -96,25 +96,19 @@ class SymbolModal extends Component {
   componentWillUnmount() {
     const shortcuts = this.context.shortcuts;
     shortcuts.off(L10N.getStr("symbolSearch.search.key2"));
+    shortcuts.off("Escape");
   }
 
   componentDidMount() {
     const shortcuts = this.context.shortcuts;
     shortcuts.on(L10N.getStr("symbolSearch.search.key2"), this.openSymbolModal);
+    shortcuts.on("Escape", this.closeModal);
     this.updateResults(this.state.query);
   }
 
   componentDidUpdate(prevProps: any, prevState: any) {
     if (this.refs.resultList && this.refs.resultList.refs) {
       scrollList(this.refs.resultList.refs, this.state.resultsIndex);
-    }
-
-    if (
-      this.refs.searchInput &&
-      this.refs.searchInput.input &&
-      this.props.enabled
-    ) {
-      this.refs.searchInput.input.focus();
     }
 
     if (!prevProps.enabled && this.props.enabled) {
@@ -269,7 +263,6 @@ class SymbolModal extends Component {
           handleNext={() => this.traverseResults(1)}
           handlePrev={() => this.traverseResults(-1)}
           handleClose={this.closeModal}
-          ref="searchInput"
         />
       </div>
     );
@@ -303,12 +296,7 @@ class SymbolModal extends Component {
     }
 
     return (
-      <Modal
-        enabled={enabled}
-        shortcut="symbolSearch.search.key2"
-        handleOpen={this.openSymbolModal}
-        handleClose={this.closeModal}
-      >
+      <Modal in={enabled} handleClose={this.closeModal}>
         {this.renderInput()}
         {this.renderResults()}
       </Modal>
@@ -338,7 +326,7 @@ export default connect(
   state => {
     const source = getSelectedSource(state);
     return {
-      enabled: getActiveSearchState(state) === "symbol",
+      enabled: Boolean(getActiveSearchState(state) === "symbol" && source),
       symbolType: getSymbolSearchType(state),
       symbols: _getFormattedSymbols(state, source)
     };
--- a/src/components/shared/Modal.css
+++ b/src/components/shared/Modal.css
@@ -4,13 +4,9 @@
   justify-content: center;
   width: 100%;
   height: 100%;
-  z-index: -1;
   top: 0;
   left: 0;
   transition: z-index 200ms;
-}
-
-.modal-wrapper.enabled {
   z-index: 9;
 }
 
@@ -25,11 +21,13 @@
   box-shadow: 2px 4px 6px var(--popup-shadow-color);
 }
 
-.theme-dark .symbol-modal {
-  box-shadow: 2px 4px 6px var(--popup-shadow-color);
+.modal.entering,
+.modal.exited {
+  transform: translateY(-250px);
 }
 
-.modal.enabled {
+.modal.entered,
+.modal.exiting {
   transform: translateY(30px);
 }
 
--- a/src/components/shared/Modal.js
+++ b/src/components/shared/Modal.js
@@ -1,20 +1,20 @@
 // @flow
 
-import { DOM as dom, Component, PropTypes } from "react";
+import { DOM as dom, createFactory, Component, PropTypes } from "react";
 import type { Children } from "react";
 import classnames from "classnames";
+import _Transition from "react-transition-group/Transition";
+const Transition = createFactory(_Transition);
 
 import "./Modal.css";
 
 type ModalProps = {
-  enabled: boolean,
-  shortcut: string,
+  status: string,
   children?: Children,
-  handleOpen: (_: any, e: SyntheticEvent) => any,
   handleClose: () => any
 };
 
-export default class Modal extends Component {
+export class _Modal extends Component {
   props: ModalProps;
 
   constructor(props: ModalProps) {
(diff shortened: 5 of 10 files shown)


src/components/shared/SearchInput.js
```diff
@@ -38,6 +38,11 @@ class SearchInput extends Component {
 
   static defaultProps: Object;
 
+  componentDidMount() {
+    this.input.focus();
```
codehag: :+1:

src/components/shared/SearchInput.js
```diff
@@ -38,6 +38,11 @@ class SearchInput extends Component {
...
+    this.input.select();
```
codehag: In the old code, we used setSelectionRange -- was this to move the cursor to the end of the input? I also noticed that in both cases we only fire select when the component is opened, and in that case it is empty. perhaps we don't need this?

wldcordeiro: We were using it but we were making a selection of the end to the end? So I wasn't sure what exactly that was achieving. I don't think it nor `.select()` make a ton of sense so I don't care for keeping them.

wldcordeiro:
> can you also add the test from the issue?

Sure! I forgot that the issue had a test associated. Thanks @codehag

src/test/mochitest/browser_dbg-searching.js
```diff
@@ -2,12 +2,15 @@
...
 
 // Testing source search
-add_task(function* () {
+add_task(function*() {
   const dbg = yield initDebugger("doc-script-switching.html");
 
   pressKey(dbg, "sourceSearch");
   yield waitForElement(dbg, "input");
-  findElementWithSelector(dbg, "input").focus();
+  ok(
```
wldcordeiro: Correct me if I'm wrong but this is the way to assert in mochitest?

src/components/shared/SearchInput.js
```diff
@@ -115,7 +119,7 @@ class SearchInput extends Component {
         placeholder,
         value: query,
         spellCheck: false,
-        ref: "input"
+        ref: c => (this.input = c)
```
jasonLaster: is this the preferred style?

perhaps? `ref: c => (this.$input = c)`

it's funny, i prefer refs because their namespaced, but i'll support anything.

Also this style is fine w/ me too
https://github.com/devtools-html/debugger.html/blob/master/src/components/shared/Autocomplete.js#L51-L53

src/components/SymbolModal.js
```diff
@@ -113,8 +113,12 @@ class SymbolModal extends Component {
       scrollList(this.refs.resultList.refs, this.state.resultsIndex);
     }
 
-    if (this.refs.searchInput && this.refs.searchInput.refs.input) {
-      this.refs.searchInput.refs.input.focus();
+    if (
+      this.refs.searchInput &&
+      this.refs.searchInput.input &&
+      this.props.enabled
```
jasonLaster: nit: but lets put `this.props.enabled` first

src/components/SymbolModal.js
```diff
@@ -291,9 +295,14 @@ class SymbolModal extends Component {
   }
 
   render() {
-    const { enabled } = this.props;
+    const { enabled, selectedSource } = this.props;
+    let modalEnabled = false;
+
+    if (selectedSource && enabled) {
+      modalEnabled = true;
+    }
```
jasonLaster: we might be playing too much of a dance w/ enabled, perhaps a new method is in order:

* we dont need enabled prop
* we can consolidate logic

```js
isEnabled() {
  const {activeSearch, selectedSource} = this.props
  return activeSearch === "symbol" && selectedSource
}
```

src/test/mochitest/browser_dbg-searching.js
jasonLaster: yep, that will work!

i guess i was assuming it because the next action is typing... and it will work or fail based on that, but i don't care. this is nice to have.

src/components/shared/SearchInput.js
jasonLaster: is this the preferred style?

perhaps? `ref: c => (this.$input = c)`

it's funny, i prefer refs because their namespaced, but i'll support anything.

Also this style is fine w/ me too
https://github.com/devtools-html/debugger.html/blob/master/src/components/shared/Autocomplete.js#L51-L53

src/components/SymbolModal.js
jasonLaster: nit: but lets put `this.props.enabled` first

src/components/SymbolModal.js
jasonLaster: we might be playing too much of a dance w/ enabled, perhaps a new method is in order:

* we dont need enabled prop
* we can consolidate logic

```js
isEnabled() {
  const {activeSearch, selectedSource} = this.props
  return activeSearch === "symbol" && selectedSource
}
```

src/test/mochitest/browser_dbg-searching.js
jasonLaster: yep, that will work!

i guess i was assuming it because the next action is typing... and it will work or fail based on that, but i don't care. this is nice to have.

src/components/SymbolModal.js
wldcordeiro: So `enabled` is also used for whether to update the results or not. I figured that I could instead modify the value we get in the `connect` call.

```js
export default connect(
  state => {
    const source = getSelectedSource(state);
    return {
      enabled: getActiveSearchState(state) === "symbol" && source,
      symbolType: getSymbolSearchType(state),
      symbols: _getFormattedSymbols(state, source)
    };
  },
  dispatch => bindActionCreators(actions, dispatch)
)(SymbolModal);
```

src/components/shared/SearchInput.js
wldcordeiro: Are you suggesting that we use the `$` to prefix it as a ref?

src/components/shared/Modal.js
```diff
@@ -35,6 +34,15 @@ export default class Modal extends Component {
     shortcuts.on(L10N.getStr(this.props.shortcut), this.props.handleOpen);
   }
 
+  componentDidUpdate() {
+    const shortcuts = this.context.shortcuts;
+    if (this.props.enabled) {
```
jasonLaster: hmm, this worries me because we might be doing too many `on` calls.

Should we add a check in here like `!props.enabled && nextProps.enabled`?

src/components/shared/SearchInput.js
jasonLaster: i prefer `$` if we do the property style. I still like ReactDOM.findNode or `this.ref` as well.

wldcordeiro: Oh I could put the ref into the refs object as well by changing the line to `this.refs.input = c`

src/components/shared/Modal.js
wldcordeiro: I tried this and it wouldn't work after the first time you toggled on.

src/components/shared/SearchInput.js
wldcordeiro: Correction. Seems like you can't add the property to the refs object via a function ref.

codehag:
Rebased

codehag:
Not sure what is happening with the tests; they pass on master and locally. looks like the problem is again with pretty print:

```

The following tests failed:

3522 INFO TEST-UNEXPECTED-FAIL | devtools/client/debugger/new/test/mochitest/browser_dbg-pretty-print.js | Uncaught exception - at chrome://mochitests/content/browser/devtools/client/debugger/new/test/mochitest/head.js:382 - Error: Unable to find source: math.min.js:formatted

Stack trace:

    findSource@chrome://mochitests/content/browser/devtools/client/debugger/new/test/mochitest/head.js:382:11

    @chrome://mochitests/content/browser/devtools/client/debugger/new/test/mochitest/browser_dbg-pretty-print.js:13:17

    Tester_execTest@chrome://mochikit/content/browser-test.js:735:9

    Tester.prototype.nextTest</<@chrome://mochikit/content/browser-test.js:655:7

    SimpleTest.waitForFocus/waitForFocusInner/focusedOrLoaded/<@chrome://mochikit/content/tests/SimpleTest/SimpleTest.js:791:59

    Tester_execTest@chrome://mochikit/content/browser-test.js:735:9

    Tester.prototype.nextTest</<@chrome://mochikit/content/browser-test.js:655:7

    SimpleTest.waitForFocus/waitForFocusInner/focusedOrLoaded/<@chrome://mochikit/content/tests/SimpleTest/SimpleTest.js:791:59

3523 INFO TEST-UNEXPECTED-FAIL | devtools/client/debugger/new/test/mochitest/browser_dbg-searching.js | focused on input - 

Stack trace:

    chrome://mochitests/content/browser/devtools/client/debugger/new/test/mochitest/browser_dbg-searching.js:null:10

    Tester_execTest@chrome://mochikit/content/browser-test.js:735:9

    Tester.prototype.nextTest</<@chrome://mochikit/content/browser-test.js:655:7

    SimpleTest.waitForFocus/waitForFocusInner/focusedOrLoaded/<@chrome://mochikit/content/tests/SimpleTest/SimpleTest.js:791:59

3524 INFO TEST-UNEXPECTED-FAIL | devtools/client/debugger/new/test/mochitest/browser_dbg-searching.js | focused on input - 

Stack trace:

    chrome://mochitests/content/browser/devtools/client/debugger/new/test/mochitest/browser_dbg-searching.js:null:23

    Tester_execTest@chrome://mochikit/content/browser-test.js:735:9

    Tester.prototype.nextTest</<@chrome://mochikit/content/browser-test.js:655:7

    SimpleTest.waitForFocus/waitForFocusInner/focusedOrLoaded/<@chrome://mochikit/content/tests/SimpleTest/SimpleTest.js:791:59

Buffered messages finished

SUITE-END | took 150s


./bin/run-mochitests-docker returned exit code 1
```

src/components/ProjectSearch/TextSearch.js
```diff
@@ -59,7 +59,7 @@ export default class TextSearch extends Component {
   componentDidMount() {
     const shortcuts = this.context.shortcuts;
     if (this.refs.searchInput) {
-      this.refs.searchInput.refs.input.focus();
+      this.refs.searchInput.$input.focus();
```
codehag: hmm this looks like it might give us trouble later on -- we are controlling a child component from a parent component; this makes it difficult to separate them -- we probably want searchInput to do this, rather than doing it here.

src/components/shared/SearchInput.js
```diff
@@ -39,6 +39,10 @@ class SearchInput extends Component {
 
   static defaultProps: Object;
 
+  componentDidMount() {
+    this.$input.focus();
```
codehag: this should be enough... was it not working with this ?

package.json
```diff
@@ -68,14 +68,24 @@
...
     "react": "=15.3.2",
     "react-dom": "=15.3.2",
+    "react-transition-group": "^2.2.0",
     "redux-saga": "^0.15.4",
     "reselect": "^3.0.0",
     "svg-inline-react": "^1.0.2",
     "wasmparser": "^0.4.10"
   },
-  "files": ["src", "assets"],
+  "files": [
```
wldcordeiro: I think yarn reformats this file when you use `yarn add` 🤷‍♂️

src/components/shared/SearchInput.js
wldcordeiro: Got the fix working across the board by putting this in both mount and update.

src/components/shared/Modal.css
```diff
@@ -20,16 +15,17 @@
   width: 500px;
   height: 230px;
   background-color: var(--theme-codemirror-gutter-background);
-  transform: translateY(-250px);
   transition: transform 150ms cubic-bezier(0.07, 0.95, 0, 1);
   box-shadow: 2px 4px 6px var(--popup-shadow-color);
 }
 
-.theme-dark .symbol-modal {
-  box-shadow: 2px 4px 6px var(--popup-shadow-color);
```
wldcordeiro: Redundant rule? It's on the modal itself just above.

src/components/shared/SearchInput.js
wldcordeiro: Because the modal is mounted the symbol modal needs to have that if block to focus the input. I'll research further to see if there's a solution that's cleaner but this gets the focus events working on all the search dialogs and ensures we have no regressions on the shortcuts.

src/components/SymbolModal.js
```diff
@@ -96,11 +96,13 @@ class SymbolModal extends Component {
   componentWillUnmount() {
     const shortcuts = this.context.shortcuts;
     shortcuts.off(L10N.getStr("symbolSearch.search.key2"));
+    shortcuts.off("Escape");
   }
 
   componentDidMount() {
     const shortcuts = this.context.shortcuts;
     shortcuts.on(L10N.getStr("symbolSearch.search.key2"), this.openSymbolModal);
+    shortcuts.on("Escape", this.closeModal);
```
wldcordeiro: It was problematic adding shortcuts to the modal itself and would leave the shortcut disabled on its dismount. Letting the consumer's handle it seems like a simpler API.

src/components/shared/Modal.js
```diff
@@ -23,38 +23,50 @@ export default class Modal extends Component {
...
+
+export default function Slide({
+  in: inProp,
+  children,
+  handleClose
+}: SlideProps) {
+  return Transition({
+    in: inProp,
+    timeout: 175,
+    appear: true,
```
wldcordeiro: > Normally a component is not transitioned if it shown when the <Transition> component mounts. If you want to transition on the first mount set appear to true, and the component will transition in as soon as the <Transition> mounts.

This prop is awesome but sadly there doesn't seem to be a `leave` that is the equivalent. I've opened an issue on react-transition-group requesting one to see what they recommend.

wldcordeiro:
So now the two assertions I added to the mochitests fail but from testing all three search inputs get focus when you open them. 🙁

Issue #3510, opened 2017-07-31
[TESTS] tabs unit test is intermittent on CI


jasonLaster:
```
jest -i --testResultsProcessor jest-junit-reporter --coverage --coverageDirectory=$CIRCLE_TEST_REPORTS/jest-coverage
 PASS  src/utils/tests/sources-tree.js
 PASS  src/actions/tests/pending-breakpoints.js (9.294s)
 PASS  src/actions/tests/breakpoints.js
 PASS  src/utils/tests/object-inspector.js
 PASS  src/actions/tests/sources.js
 PASS  src/utils/search/tests/build-query.js
 PASS  src/utils/tests/scopes.js
 PASS  src/actions/tests/ui.js
 PASS  src/actions/tests/ast.js
/home/ubuntu/debugger.html/node_modules/devtools-utils/src/worker-utils.js:48
          this.worker.removeEventListener("message", listener);
                     ^

TypeError: Cannot read property 'removeEventListener' of null
    at Worker.listener (/home/ubuntu/debugger.html/node_modules/devtools-utils/src/worker-utils.js:48:22)
    at self.handlers.message.forEach.func (/home/ubuntu/debugger.html/node_modules/workerjs/index.js:25:53)
    at Array.forEach (native)
    at ChildProcess.<anonymous> (/home/ubuntu/debugger.html/node_modules/workerjs/index.js:25:32)
    at emitTwo (events.js:106:13)
    at ChildProcess.emit (events.js:194:7)
    at process.nextTick (internal/child_process.js:766:12)
    at _combinedTickCallback (internal/process/next_tick.js:73:7)
    at process._tickCallback (internal/process/next_tick.js:104:9)

jest -i --testResultsProcessor jest-junit-reporter --coverage --coverageDirectory=$CIRCLE_TEST_REPORTS/jest-coverage returned exit code 1
```

codehag:
It looks like the fix is merged, what is the criteria for closing this? do we want to wait a bit and see if it comes up again?

jasonLaster:
I don't think this was AST actually, it was tabs...

jasonLaster:
the criteria is it appearing. The error should be gone in CI

PR #3511, opened 2017-07-31
Fix race condition with project text search tests

jasonLaster:
Associated Issue: #3510 

### Summary of Changes

* updates the project text search test to actually be async


src/utils/search/index.js
```diff
@@ -7,3 +7,4 @@ export const stopSearchWorker = dispatcher.stop.bind(dispatcher);
 
 export const getMatches = dispatcher.task("getMatches");
 export const searchSources = dispatcher.task("searchSources");
```
codehag: is this still used?

Issue #3493, opened 2017-07-28
[Expressions] don’t evaluate while we’re stepping


jasonLaster:
We currently evaluate expressions whenever we resume. The problem is that we resume before we pause when we step.

We should fix this because it slows down our stepping speed. We'll want to do something like [this](https://github.com/devtools-html/debugger.html/compare/next...jasonLaster:currently-stepping?expand=1)

```diff
diff --git a/src/actions/pause.js b/src/actions/pause.js
index 927a21b..90dea0c 100644
--- a/src/actions/pause.js
+++ b/src/actions/pause.js
@@ -3,7 +3,7 @@
 import { selectSource } from "./sources";
 import { PROMISE } from "../utils/redux/middleware/promise";

-import { getPause, getLoadedObject } from "../selectors";
+import { getPause, getLoadedObject, currentlyStepping } from "../selectors";
 import { updateFrameLocations } from "../utils/pause";
 import { evaluateExpressions } from "./expressions";

@@ -24,13 +24,15 @@ type CommandType = { type: string };
  * @static
  */
 export function resumed() {
-  return ({ dispatch, client }: ThunkArgs) => {
+  return ({ getState, dispatch, client }: ThunkArgs) => {
     dispatch({
       type: "RESUME",
       value: undefined
     });

-    dispatch(evaluateExpressions(null));
+    if (!currentlyStepping(getState())) {
+      dispatch(evaluateExpressions(null));
+    }
   };
 }
:
```

bomsy:
![image](https://user-images.githubusercontent.com/792924/28871193-70d11b30-777b-11e7-8586-d80657d9b9a3.png)

it looks like it takes an average of 2+ secs

jasonLaster:
it's probably slower on the launchpad.

Also, how many watch expressions do you have?

bomsy:
True! Ignore the perf image above, made a wrong assumption

PR #3530, opened 2017-08-02
Do not evaluate expressions on stepping

bomsy:
Associated Issue: #3493

Here's the Pull Request Doc

### Summary of Changes

* Added a new command property to the pause reducer
* Added `isStepping` to enable us know when we are stepping
* Do not evaluate expressions when we resume, if we are stepping

### Test Plan

- [x] Add breakpoint 
- [x] Pause on breakpoint 
- [x] Click step over button
- [x] See how long it takes to step

### DIFF
--- a/src/actions/pause.js
+++ b/src/actions/pause.js
@@ -3,7 +3,7 @@
 import { selectSource } from "./sources";
 import { PROMISE } from "../utils/redux/middleware/promise";
 
-import { getPause, getLoadedObject } from "../selectors";
+import { getPause, getLoadedObject, isStepping } from "../selectors";
 import { updateFrameLocations } from "../utils/pause";
 import { evaluateExpressions } from "./expressions";
 
@@ -24,13 +24,15 @@ type CommandType = { type: string };
  * @static
  */
 export function resumed() {
-  return ({ dispatch, client }: ThunkArgs) => {
+  return ({ dispatch, client, getState }: ThunkArgs) => {
     dispatch({
       type: "RESUME",
       value: undefined
     });
 
-    dispatch(evaluateExpressions(null));
+    if (!isStepping(getState())) {
+      dispatch(evaluateExpressions(null));
+    }
   };
 }
 
@@ -103,7 +105,7 @@ export function command({ type }: CommandType) {
 
     return dispatch({
       type: "COMMAND",
-      value: undefined
+      value: { type }
     });
   };
 }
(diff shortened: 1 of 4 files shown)


src/reducers/pause.js
```diff
@@ -137,6 +139,9 @@ function update(state: PauseState = State(), action: Action): PauseState {
         shouldPauseOnExceptions,
         shouldIgnoreCaughtExceptions
       });
+
+    case "COMMAND":
+      return { ...state, command: action.value.type };
```
jasonLaster: we should clear this field when it's done.

src/actions/pause.js
```diff
@@ -24,13 +24,15 @@ type CommandType = { type: string };
...
+  return ({ dispatch, client, getState }: ThunkArgs) => {
     dispatch({
       type: "RESUME",
       value: undefined
     });
 
-    dispatch(evaluateExpressions(null));
+    if (!isStepping(getState())) {
+      dispatch(evaluateExpressions(null));
+    }
```
jasonLaster: thanks

src/actions/pause.js
```diff
@@ -103,7 +105,7 @@ export function command({ type }: CommandType) {
 
     return dispatch({
       type: "COMMAND",
-      value: undefined
+      value: { type }
```
jasonLaster: lets add some jest unit tests to demonstrate that command is set and then cleared when the command is done.

I think we can do something like:


```js
it("lala", async () => {
  let resolve = null;
  const slowClient = {
     stepIn: () => new Promise(_resolve => {resolve = _resolve})
  }
  
  await actions.paused()
  actions.stepIn()
  
  expect(isStepping(state)).toBeTruthy()
  resolve();
  expect(isStepping(state)).toBeFalsey()
}
```

src/actions/pause.js
```diff
@@ -24,13 +24,15 @@ type CommandType = { type: string };
...
-  return ({ dispatch, client }: ThunkArgs) => {
+  return ({ dispatch, client, getState }: ThunkArgs) => {
     dispatch({
       type: "RESUME",
       value: undefined
     });
 
-    dispatch(evaluateExpressions(null));
+    if (!isStepping(getState())) {
+      dispatch(evaluateExpressions(null));
```
jasonLaster: it would be nice to have a test here too:


```js
+ describe("pause", () => {
+   it("should set and clear the command", async () => {
+     const client = { evaluate: jest.fn() };
+     const { dispatch, getState } = createStore(client);
+     dispatch(actions.stepIn());
+     await dispatch(actions.resumed());
+     expect(client.evaluate.calls.length).toEqual(0)
+   });
+ });
  

```

PR #3536, opened 2017-08-02
Refactor tree utils

jasonLaster:
### Summary of Changes

* I merged the flow types work in
* I also merged the test file splitting and snapshotting
* I'm finding the source-tree utils hard to grok because there's some complicated logic. This splits it up into nicer bite sized pieces

This is an example where git got complicated with three PRs so i made it one...

I changed almost 0% of the logic

Here's what [addToTree](https://github.com/jasonLaster/debugger.html/blob/ad450d6638bb3b07efed159ff89bf76f120e503a/src/utils/sources-tree/addToTree.js) looks like now.


src/utils/sources-tree/getDirectories.js
```diff
@@ -0,0 +1,50 @@
...
+      if (subtree.path.replace(/http(s)?:\//, "") == sourceUrl) {
+        returnTarget = subtree;
+        return;
+      }
+    }
+  }
+
+  sourceTree.contents.forEach(_traverse);
+
+  if (!returnTarget) {
```
jasonLaster: this is the only logic change. We now return the root if nothing is found so we can promise to have a node returned

src/utils/sources-tree/utils.js
```diff
@@ -0,0 +1,65 @@
...
+
+export function createNode(
+  name: string,
+  path: string,
+  contents: SourceRecord | Array<Node>
+): Node {
+  return {
+    name,
+    path,
+    contents
```
jasonLaster: this used to support `null` contents. That's now gone

src/utils/sources-tree.js
```diff
@@ -50,7 +49,7 @@ export function createNode(name: any, path: any, contents?: any): Node {
  * @memberof utils/sources-tree
  * @static
  */
-export function createParentMap(tree: any): WeakMap<any, any> {
+export function createParentMap(tree: Node): WeakMap<Node, Node> {
```
codehag: :+1:

Issue #3496, opened 2017-07-28
[Expressions] should have a unique key


jasonLaster:
It looks like the recent change to expressions caused the keys to no longer be unique

```
Warning: Each child in an array or iterator should have a unique "key" prop. Check the render method of `Tree`. See https://fb.me/react-warning-keys for more information.
    in TreeNode (created by Tree)
    in Tree (created by ManagedTree)
    in div (created by ManagedTree)
    in ManagedTree (created by ObjectInspector)
    in ObjectInspector (created by Expressions)
    in div (created by Expressions)
    in span (created by Expressions)
    in Expressions (created by Connect(Expressions))
    in Connect(Expressions) (created by bound createElement)
    in bound createElement (created by Accordion)
    in div (created by Accordion)
    in div (created by Accordion)
    in div (created by Accordion)
    in Accordion (created by SecondaryPanes)
    in div (created by SecondaryPanes)
    in SecondaryPanes (created by Connect(SecondaryPanes))
    in Connect(SecondaryPanes) (created by App)
    in div (created by SplitBox)
    in div (created by SplitBox)
    in SplitBox (created by App)
    in div (created by SplitBox)
    in div (created by SplitBox)
    in SplitBox (created by App)
    in div (created by App)
    in App (created by Connect(App))
    in Connect(App)
    in Provider 
```

codehag:
I did a bit of investigation, the problem appears to be in how we generate the key in managedTree / ObjectInspector. In object inspector we are returning `item.path` as the item key, which can be undefined if the debugger is not paused, or the watch expression does not exist in that pane..

we are also using the key to manage expanded state in managedTree, making this a bit risky to change. we might want to use refs there instead. It might be a bit more work.

Here is a rough diff of not relying on the key to do logic:
```diff
diff --git a/src/components/PrimaryPanes/SourcesTree.js b/src/components/PrimaryPanes/SourcesTree.js
index dbfa131d..1818c720 100644
--- a/src/components/PrimaryPanes/SourcesTree.js
+++ b/src/components/PrimaryPanes/SourcesTree.js
@@ -260,6 +260,7 @@ class SourcesTree extends Component {
       },
       getRoots: () => sourceTree.contents,
       getKey: (item, i) => item.path,
+      getPath: item => item.path,
       itemHeight: 21,
       autoExpandDepth: 1,
       autoExpandAll: false,
[...]

PR #3543, opened 2017-08-03
[Expressions] should have a unique key

codehag:
Associated Issue: #3496 

### Summary of Changes
* change getKey to getPath to maybe make it clearer that it has more
tasks than setting the key

* make objectInspector getPath method create a unique path, even if the
element is undefined

### Test Plan

Unit tests

### DIFF
--- a/src/components/PrimaryPanes/SourcesTree.js
+++ b/src/components/PrimaryPanes/SourcesTree.js
@@ -247,8 +247,6 @@ class SourcesTree extends Component {
 
     const isEmpty = sourceTree.contents.length === 0;
 
-    console.log(listItems);
-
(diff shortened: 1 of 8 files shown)


src/components/PrimaryPanes/SourcesTree.js
```diff
@@ -247,6 +247,8 @@ class SourcesTree extends Component {
 
     const isEmpty = sourceTree.contents.length === 0;
 
+    console.log(listItems);
```
jasonLaster: :/

codehag: ack

PR #3556, opened 2017-08-04
[WIP] Use devtools-reps ObjectInspector

nchevobbe:
I think this is looking good.
It misses 2 things : having global window object in the Scopes panel dimmed ([PR up](https://github.com/devtools-html/devtools-core/pull/555)) and function styling (i.e. `fn()` instead of `function fn()`), which I'll work on today.

This PR should be block by devtools-reps 0.12 bundle.


nchevobbe:
@jasonLaster do you remember how you got the watch expressions issue ?

nchevobbe:
This should work with devtools-reps 0.12 bundle

jasonLaster:
@nchevobbe getting this error

screen shot 2017-08-10 at 6 49 57 pm

jasonLaster:
this should be good to go when we're green

jasonLaster:
These tests need better getters
devtools/client/debugger/new/test/mochitest/browser_dbg-expressions.js
devtools/client/debugger/new/test/mochitest/browser_dbg-scopes-mutations.js

nchevobbe:
This would be good with the 0.12.1 package.
I'll check those tests too :)

nchevobbe:
Or more like a 0.12.2 version. One test failing is waiting for root function `foo` to be displayed as `foo()`, where we display it as `function foo()` because we don't simplify root functions (for the console).
Maybe omitting the `function` keyword should just be for TINY mode so it behaves like expected in the debugger and in console (for top-level and then nested functions).

For the scope mutation test, it looks like we don't wait enough time: the action is done but nodes are not shown yet. I may re-use what I do on [webconsole test](http://searchfox.org/mozilla-central/source/devtools/client/webconsole/new-console-output/test/mochitest/browser_webconsole_object_inspector.js#43), i.e. wait for dom mutation on the tree instead of waiting for the action to end, which is implementation specific.

I'll tackle that on Monday, **I want** it to be done !

nchevobbe:
The PR for functions is up https://github.com/devtools-html/devtools-core/pull/594

jasonLaster:
> wait for dom mutation

I'd rather find a better Action. It's easy to do this by enabling action logging [here](https://github.com/nyrosmith/debugger.html/blob/convert_modal_to_jsx/src/utils/bootstrap.js#L22-L23)

jasonLaster:
I was able to run yarn copy-assets-watch....

nchevobbe:
Almost there. Need to wait for a fix to `getExpressions(...).findKey is not a function` error to rebase and make sure everything is working fine.

Issue #3500, opened 2017-07-28
[DebugLine] expressions are not marked when selecting sources


jasonLaster:
STR:
1. pause
2. select a source where there's another frame

ER: see the debug expression
AR: you see the debug line, but not expression

NOTE: selecting frames in the call stack properly show expressions.

![](http://g.recordit.co/ZmvPyJc1k7.gif)

jasonLaster:
There's a similar issue with breakpoints:

![](http://g.recordit.co/WtAiNEpRhC.gif)

codehag:
I took a look, and here is what is happening:

debug expressions is a single instance, each time we switch files the debug expression is pointing to a line on a single source. however when you select it via the callstack, it sets a new line.

I think we might want to rethink using the debug line to show where in the callstack a given call was made. If we use it, then it would be confusing -- as we are not currently paused at that location. Maybe we need a different visual cue for this?

jasonLaster:
Just spoke offline, and we only want to show one frame's debug line.

PR #3564, opened 2017-08-04
[DebugLine] expressions are not marked when selecting sources

codehag:
Associated Issue #3500

### DIFF
--- a/src/components/Editor/tests/DebugLine.js
+++ b/src/components/Editor/tests/DebugLine.js
@@ -1,17 +1,15 @@
 import React from "react";
 import { shallow } from "enzyme";
 import DebugLine from "../DebugLine";
-jest.mock(
-  "../../../utils/editor/source-documents",
-  jest.fn(() => ({
-    getDocument: jest.fn()
-  }))
-);
+jest.mock("../../../utils/editor/source-documents", () => ({
+  getDocument: jest.fn()
+}));
 import { getDocument } from "../../../utils/editor/source-documents";
-getDocument.mockImplementation(() => ({
+const mockGetDocument = {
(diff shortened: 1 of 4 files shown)


codehag:
The tests still need a bit of work here, but most of the functionality is there

src/components/Editor/tests/DebugLine.js
```diff
@@ -0,0 +1,99 @@
...
+
+  describe("update", () => {
+    const selectedLocation = {
+      location: {
+        sourceId: "x",
+        line: 1
+      }
+    };
+
+    fit("should remove the old debug line", async () => {
```
jasonLaster: `fit`

src/components/Editor/tests/DebugLine.js
```diff
@@ -0,0 +1,99 @@
...
+      const { component } = render();
+      component.setProps({ selectedLocation });
+      const getDoc = getDocument();
```
jasonLaster: do we want to assert we get the right document... it might be nice to simulate setting a new document on codemirror and make sure we get the document for the correct source

src/components/Editor/tests/DebugLine.js
```diff
@@ -0,0 +1,99 @@
...
+    });
+
+    it("should clear the debug line", async () => {
+      const { component, props: { editor } } = render();
+      component.unmount();
+      expect(editor.codeMirror.removeLineClass).toHaveBeenCalled();
+    });
+  });
+
+  describe("update", () => {
```
jasonLaster: we should be clearing the source documents between codemirror tests!

PR #3567, opened 2017-08-04
Update CloseButton to use jsx

darkwing:
Had a few more minutes, did one more

### DIFF
--- a/src/components/shared/Button/Close.js
+++ b/src/components/shared/Button/Close.js
@@ -1,5 +1,5 @@
 // @flow
-import { DOM as dom, PropTypes } from "react";
+import React, { PropTypes } from "react";
 import Svg from "../Svg";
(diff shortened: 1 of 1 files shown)


src/components/shared/Button/Close.js
```diff
@@ -10,16 +10,18 @@ type CloseButtonType = {
...
-      title: tooltip
-    },
-    Svg("close")
+  return (
+    <div
+      className={buttonClass ? `close-btn ${buttonClass}` : "close-btn"}
+      onClick={handleClick}
+      title={tooltip}
+    >
+      {Svg("close")}
```
wldcordeiro: Unrelated but it'd be nice to update `Svg` to take a `name` prop and then we could have these calls be JSX too.

PR #3573, opened 2017-08-04
Convert Svg() to JSX

darkwing:
Ticking another one off.

### DIFF
--- a/assets/images/Svg.js
+++ b/assets/images/Svg.js
@@ -62,5 +62,5 @@ module.exports = function(name, props) {
     className = "";
   }
   props = Object.assign({}, props, { className, src: svg[name] });
-  return React.createElement(InlineSVG, props);
+  return <InlineSVG {...props} />;
 };


zaggy:
Hi, 

Maybe it is worth, while you are at it, to try an update `svg-inline-react` package to the latest version?

assets/images/Svg.js
```diff
@@ -62,5 +62,5 @@ module.exports = function(name, props) {
     className = "";
   }
   props = Object.assign({}, props, { className, src: svg[name] });
-  return React.createElement(InlineSVG, props);
+  return <InlineSVG {...props} />;
```
wldcordeiro: So this is an idea. But you'd need to touch more of the codebase. How about converting this component from:

```js
module.exports = function(name, props) {
  // eslint-disable-line
  if (!svg[name]) {
    throw new Error("Unknown SVG: " + name);
  }
  let className = name;
  if (props && props.className) {
    className = `${name} ${props.className}`;
  }
  if (name === "subSettings") {
    className = "";
  }
  props = Object.assign({}, props, { className, src: svg[name] });
  return <InlineSVG {...props} />;
};
```

to
```js
module.exports = function({ name, ...props }) {
  // eslint-disable-line
  if (!svg[name]) {
    throw new Error("Unknown SVG: " + name);
  }
  let className = name;
  if (props && props.className) {
    className = `${name} ${props.className}`;
  }
  if (name === "subSettings") {
    className = "";
  }
  props = { ...props, { className, src: svg[name] } };
  return <InlineSVG {...{ ...props, className, src: svg[name] }} />;
};
```

We would need to modify the usage of `Svg` but then we could use it like a component as well and convert the usages like so

```js
// from this
Svg("some-name", { foo: bar });

// to this
<Svg name="some-name", foo={bar} />
```

wldcordeiro:
@zaggy I would leave it for now as it's a major version bump.

darkwing:
@wldcordeiro I was thinking the same but I didn't want to every file that used it yet; easier if we JSX all the components and then make this change

wldcordeiro:
>  I was thinking the same but I didn't want to touch every file that used it yet; easier if we JSX all the components and then make this change

@darkwing agreed. I'll amend the main issue with that as an item.

Issue #3584, opened 2017-08-07
[SymbolModal] is not showing result list


jasonLaster:
It looks like these two commits:

https://github.com/devtools-html/debugger.html/commit/139f4ff67b9634b9db1bf3a62f9dfecef773e276#diff-40fa768fe2f9d9908ff75bfff396c7fc

https://github.com/devtools-html/debugger.html/commit/424d4040af08542c344164dd7c83649ffba79543#diff-40fa768fe2f9d9908ff75bfff396c7fc

broke the symbol modal result list. My guess is that we're not returning the react components from result list. Should be a simple fix.

PR #3585, opened 2017-08-07
Show result list

jasonLaster:
Associated Issue: #3584

### Summary of Changes

the input wrapper was surrounding results as well. looks like we can easily mess up the dom heirarchy.


src/components/tests/__snapshots__/SymbolModal.js.snap
```diff
@@ -1,196 +1,62 @@
...
-        Immutable.Map {
-        }
-    }
-    setSelectedSymbolType={[Function]}
-    symbolType="functions"
-/>,
-}
+    </div>
+  </Modal>
+</SymbolModal>"
```
jasonLaster: this output is pretty nice in snapshots

Issue #3577, opened 2017-08-04
[Editor] Breakpoint condition is barely readable


SebastianZ:
Breakpoint conditions are hard to read while editing them, because they are using gray text on light gray background:

![Barely readable breakpoint condition](https://user-images.githubusercontent.com/958943/28990258-fd59c708-797b-11e7-9ab9-e66204dd6202.png)

The text should rather be displayed in black or gray (or in the same color as the source text).

Sebastian

PR #3586, opened 2017-08-07
Use theme body color for conditional breaking point

liorheber:
Associated Issue: #3577 
#goodnessSquad

Here's the Pull Request Doc

### Summary of Changes

* Aligned the conditional breaking point text color to use the same color as used in the source code section.

### Test Plan

Tested the color of the conditional breaking point input color is updated correctly.

Before:
image

After:
image

### DIFF
--- a/src/components/Editor/Editor.css
+++ b/src/components/Editor/Editor.css
@@ -19,7 +19,7 @@
 }
 
 .theme-light {
-  --theme-conditional-breakpoint-color: #ccd1d5;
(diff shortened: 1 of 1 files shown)


jasonLaster:
Nice! can you share some before and after screenshots.

Also, curious, but it might be nice to see if you can drop CodeMirror into the panel :)

liorheber:
I'm not sure why the build indicates a drop in the test coverage.

jasonLaster:
Thanks

Issue #3506, opened 2017-07-31
Convert Components to JSX


wldcordeiro:
We've been looking into converting our components to JSX for a while this is a tracking issue.

#### Automation

This [jscodeshift][jsc] could help out :)
[jsc]:https://github.com/HipsterBrown/espresso/blob/master/src/transforms/jsx.js

# Components

- [x] App
- [x] SymbolModal
- [x] WelcomeBox
- [x] Breakpoints
- [x] CallSites
- [x] ConditionalPanel
- [x] Footer
- [x] Editor (index)
- [x] Preview (index)
- [x] SearchBar
- [x] Tabs (index)
- [x] PrimaryPanes (index)
- [x] Outline
- [x] SourcesTree
- [x] ProjectSearch (index)
- [x] SourceSearch
- [x] TextSearch
- [x] Frames (index)
- [x] Frame
- [x] Group
- [x] WhyPaused
- [x] SecondaryPanes (index)
- [x] ChromeScopes
- [x] CommandBar
- [x] EventListeners
- [x] Expressions
- [x] Scopes
- [x] Close
- [x] PaneToggle
- [x] Accordion
- [x] AutoComplete
- [x] BracketArrow
- [x] Dropdown
- [x] ManagedTree
- [x] Modal
- [x] ~ObjectInspector~
- [x] Popover
- [x] ResultList
- [x] SearchInput
- [x] Svg
  - [x] Update usage to be JSX compatible. [See comment here.](https://github.com/devtools-html/debugger.html/pull/3573#discussion_r131511803)

nyrosmith:
Can I try to help you converting the components?

jasonLaster:
@nyrosmith sure! want to grab Modal or Popover?

nyrosmith:
Maybe starting with something trivial and then converting one after another. I think there are enough compononents to work on.
Do you prefer 1 branch per component or one branch for all changes?

jasonLaster:
one branch per component. It's easier to review + land

nyrosmith:
OK. I'll start tomorrow and really hope to deliver results quickly.

jasonLaster:
Lets also fix this bug: https://github.com/devtools-html/debugger.html/issues/3547

which we just created :)

Anzumana:
would also like to try asked @wldcordeiro which one he thinks i should start with

jasonLaster:
@Anzumana just jump in try one. no worries

amelzer:
[...]

PR #3590, opened 2017-08-07
ObjectInspector.js to JSX

ksaldana1:
Associated Issue: #3506 

### Summary of Changes

* Convert all React.createElement calls to JSX in ObjectInspector.

### Test Plan

Since there was no snapshot, I mostly just tested the functionality in the debugger. Let me know if you'd prefer me creating a snapshot test prior to the JSX conversion to give us slightly more confidence around regression. 

Testing included: Object inspector behavior in scopes vs expressions. Click to expand/collapse. Double click to edit name. SVG arrow rendering properly. 

screen shot 2017-08-07 at 1 02 10 pm
screen shot 2017-08-07 at 12 58 46 pm

### DIFF
--- a/src/components/shared/ObjectInspector.js
+++ b/src/components/shared/ObjectInspector.js
@@ -1,5 +1,5 @@
 // @flow
-import { DOM as dom, PropTypes, createFactory, Component } from "react";
+import React, { PropTypes, Component } from "react";
 import classnames from "classnames";
 import Svg from "./Svg";
 import Rep from "./Rep";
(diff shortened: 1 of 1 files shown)


src/components/shared/ObjectInspector.js
```diff
@@ -139,43 +138,46 @@ class ObjectInspector extends Component {
...
+      this.props.onDoubleClick(item, {
+        depth,
+        focused,
+        expanded
+      });
+    };
+
+    return (
+      <div
+        className={classnames("node object-node", {
```
ksaldana1: I could definitely have pulled this inline className (and Svg below) to a variable, but I wasn't too sure on proper naming. If you have any suggestions and want to avoid inlining, just let me know and I'll update this PR.

jasonLaster:
Sorry, this is actually the *one* component that we won't update because it's going to come from https://github.com/devtools-html/devtools-core/tree/master/packages/devtools-reps/src/object-inspector

I'm really sorry it was on the list

ksaldana1:
Not a problem!

PR #3609, opened 2017-08-09
Add mochi runner

jasonLaster:
### Summary of Changes

The goal is to be able to make it fun to run integration tests. We do this by cleaning up the workflow a lot:

1. `yarn mochi`
2. a clean reporter


bin/mochi.js
```diff
@@ -0,0 +1,210 @@
+const shell = require("shelljs");
```
gregtatum: Do you not flow type your bin scripts? I really like treating all my code as first class citizens.

PR #3614, opened 2017-08-10
Convert CommandBar to JSX

darkwing:


src/components/SecondaryPanes/CommandBar.js
```diff
@@ -68,20 +68,24 @@ function formatKey(action) {
 }
 
 function debugBtn(onClick, type, className, tooltip, disabled = false) {
```
wldcordeiro: Maybe `function debugBtn({ type, className, ...rest })` then you can make that block below

```jsx
return (
  <button className={classnames(type, className)} {...rest}>
    {Svg(type)}
  </button>
);
```
)
```

darkwing:
Good shout, updated.

PR #3615, opened 2017-08-10
Convert Popover, BrackArrow to JSX

darkwing:


darkwing:
Is this failure legit?  I can't tell.

jasonLaster:
nah - i think it's an intermittent we fixed last week

PR #3622, opened 2017-08-10
Update assertPausedLocation

jasonLaster:
### Summary of Changes

Update assertPausedLocation, this lets us run `./mach mochitest --repat 20 browser_browser_toolbox_debugger.js` successfully. previously it would fail hard.

We're updating `assertPausedLocation` because the browser toolbox occasionally steps into a weird source `GCTelemettry`

screen shot 2017-08-10 at 12 08 45 pm

### DIFF
--- a/src/test/mochitest/head.js
+++ b/src/test/mochitest/head.js
@@ -207,10 +207,15 @@ function waitForElement(dbg, selector) {
   return waitUntil(() => findElementWithSelector(dbg, selector));
 }
 
-function waitForSelectedSource(dbg) {
+function waitForSelectedSource(dbg, sourceId) {
   return waitForState(dbg, state => {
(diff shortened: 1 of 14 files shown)


src/test/mochitest/head.js
```diff
@@ -223,22 +228,17 @@ function waitForSelectedSource(dbg) {
  * @param {Number} line
  * @static
  */
-function assertPausedLocation(dbg, source, line) {
+function assertPausedLocation(dbg) {
```
jasonLaster: we're dropping specified source and lines because we want to focus on the fact that we're highlighting the top frame line.

jasonLaster:
I still think location makes sense, because it's the location of the frame we care about, but I see where you're going with that. Lets revisit if we want to add our own location expectation.

PR #3639, opened 2017-08-11
test project search

jasonLaster:
### Summary of Changes

I'm adding a mochitest so that we can have a simple smoke test. Also, importantly i'm using prefs instead of configs to flag project search because we need to be able to flip it in the test.

### DIFF
--- a/src/test/mochitest/browser_dbg-search-project.js
+++ b/src/test/mochitest/browser_dbg-search-project.js
@@ -1,0 +1,67 @@
+/* Any copyright is dedicated to the Public Domain.
+ * http://creativecommons.org/publicdomain/zero/1.0/ */
+
+function toggleProjectSearch() {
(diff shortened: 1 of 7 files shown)


src/test/mochitest/browser_dbg-search-project.js
```diff
@@ -0,0 +1,67 @@
...
+}
+
+function getResultsCount(dbg) {
+  const matches = dbg.selectors
+    .getTextSearchResults(dbg.getState())
+    .valueSeq()
+    .map(file => file.matches)
+    .toJS();
+
+  return [].concat(...matches).length;
```
codehag: could we do `[...matches]`? or `matches.length`?

codehag: also, since we are doing toJS() at the end here, maybe it makes more sense to do matches.size

Issue #3638, opened 2017-08-11
[TextSearch] Fix text search


jasonLaster:
It looks like the JSX refactor broke text search. We should fix this ASAP
https://github.com/devtools-html/debugger.html/pull/3636/files

STR:

1. enable `project text search` in settings
2. cmd+shift+f
3. search for text

PR #3636, opened 2017-08-11
convert ProjectSearch/*.js files to JSX

aklt:
Associated Issue: #3506

Converted the files

    src/components/ProjectSearch/index.js
    src/components/ProjectSearch/SourceSearch.js
    src/components/ProjectSearch/TextSearch.js
    src/components/ProjectSearch/ToggleSearch.js

To JSX.

The behaviour should be as before.  See

    https://github.com/devtools-html/debugger.html/pull/3396

For an example gif showing the expected behaviour.

PR #3640, opened 2017-08-12
fix problems introduced in text search

aklt:
* fixes issues introduced by
  https://github.com/devtools-html/debugger.html/pull/3636

Associated Issue: #3638

Note: There is a flow issue with this PR.  Running `yarn flow` gives this error:

```sh
yarn flow v0.24.6
$ flow
Error: src/components/PrimaryPanes/SourcesTree.js:250
                      v------------
250:     const tree = ManagedTree({
251:       key: isEmpty ? "empty" : "full",
252:       getParent: item => {
...:
270:     });
         -^ function call
 38:   props: Props;
              ^^^^^ property `disabledFocus`. Property not found in. See: src/components/shared/ManagedTree.js:38
                                  v
250:     const tree = ManagedTree({
251:       key: isEmpty ? "empty" : "full",
252:       getParent: item => {
...:
270:     });
         ^ object literal


Found 1 error
```
Which i cannot find a good solution to and I need some sleep. :-)

### DIFF
--- a/src/components/ProjectSearch/TextSearch.js
+++ b/src/components/ProjectSearch/TextSearch.js
@@ -1,8 +1,9 @@
-import React, { Component, PropTypes } from "react";
+import React, { Component, DOM as dom, PropTypes } from "react";
 import classnames from "classnames";
 
 import escapeRegExp from "lodash/escapeRegExp";
 import Svg from "../shared/Svg";
+
 import ManagedTree from "../shared/ManagedTree";
 import SearchInput from "../shared/SearchInput";
 
@@ -71,8 +72,14 @@ export default class TextSearch extends Component {
     if (focused) {
       this.focused = { setExpanded, file, expanded };
     }
-    const Arrow = Svg("arrow");
-    const File = Svg("file");
+
+    const svgArrow = Svg("arrow", {
+      className: classnames({
+        expanded
+      })
+    });
+
+    const svgFile = Svg("file");
 
     const matches = ` (${file.matches.length} match${file.matches.length > 1
       ? "es"
@@ -81,11 +88,11 @@ export default class TextSearch extends Component {
     return (
       <div
         className={classnames("file-result", { focused })}
-        key={file.id}
+        key={file.sourceId}
         onClick={e => setExpanded(file, !expanded)}
       >
-        <Arrow className={classnames({ expanded })} />
-        <File />
+        {svgArrow}
+        {svgFile}
         <span className="file-path">
           {getRelativePath(file.filepath)}
         </span>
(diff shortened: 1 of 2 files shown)


src/components/ProjectSearch/TextSearch.js
```diff
@@ -153,28 +160,23 @@ export default class TextSearch extends Component {
       }
     });
 
-    return (
-      <span className="line-value">
-        {matches}
-      </span>
-    );
+    return dom.span({ className: "line-value" }, ...matches);
```
bomsy: Why did you need to change this back?

src/components/shared/ManagedTree.js
```diff
@@ -25,7 +25,7 @@ type Props = {
   onExpand?: (item: any) => void,
   onCollapse?: (item: any) => void,
   renderItem: any,
-  disabledFocus: boolean
+  disabledFocus?: boolean
```
jasonLaster: i don't know why this is needed now... but it looks good

src/components/ProjectSearch/TextSearch.js
jasonLaster: You get this error when it's an array

```js
    return (
      <span className="line-value">
        {matches}
      </span>
    );
```
screen shot 2017-08-12 at 9 45 03 am

aklt: Good question.  There is an issue where `<span>{matches}</span>` causes a react
error that keys should be unique,.. oh! I see jasonLaster just commented about that :)

src/components/shared/ManagedTree.js
aklt: This was an attempt at making the prop optional to try and make the error mentioned in the PR
above go away.  The `disabledFocus` prop is not used in `src/components/PrimaryPanes/SourcesTree.js` so it should be optional for now.

Unfortunately making it optional did not make `yarn flow` happy.  I think this might be a known flow issue.

jasonLaster: my flow is happy. and so is CIs so you're off the hook

src/components/ProjectSearch/TextSearch.js
jasonLaster: I think we should be able to spread, but it doesn't work for me...

https://github.com/Microsoft/TypeScript/pull/11591/files#diff-36d63f88d66eddff8e2d469bea89b21cR23

jasonLaster: looks like there are some issues with spreading and it is a performance hit for react as well.
https://github.com/babel/babylon/pull/42

With that in mind, lets merge this and then refactor matches properly

Here's a gist of a patch i started https://gist.github.com/jasonLaster/11e46330c6cc17b36c4f78cb3229fa30

src/components/shared/ManagedTree.js
aklt: Gah!  So there is something strange in my setup, grrr.

src/components/ProjectSearch/TextSearch.js
aklt: Ah, interesting, so using spreads for children is off the table for now I guess.

Thanks for the links

Issue #3623, opened 2017-08-10
[tests] improve mochi runner error handling


jasonLaster:
It would be great to improve our test runner's error formatting. We have two problems.

1. Unexpected failures show two errors. We should drop the first one
2. console.error is not formatted

It would be great to format error frames and reduce a lot of the noise

here is a [branch](https://github.com/devtools-html/debugger.html/compare/master...jasonLaster:improve-errors?expand=1)

Here is what it looks like now:
<img src='https://shipusercontent.com/b0d9a63cdab152fca0ed954616591a04/Screen%20Shot%202017-08-10%20at%202.59.45%20PM.png' title='Screen Shot 2017-08-10 at 2.59.45 PM.png' width=1439>

PR #3647, opened 2017-08-12
Improve mochi errors

jasonLaster:
Associated Issue: #3623 

### Summary of Changes

* switch from module `mode` state to a data object so that tests can run in parallel
* add support for handling errors 
* some other small tweaks like results

#### Before

screen shot 2017-08-12 at 10 47 57 am

#### After

screen shot 2017-08-12 at 10 47 33 am


bin/mochi.js
```diff
@@ -48,71 +48,105 @@ const blacklist = [
...
 ];
 
-let mode = "starting";
-
 function sanitizeLine(line) {
   return line.trim().replace(/\\"/g, '"').replace(/\\"/g, '"');
 }
 
-function onGecko(line) {
+function onFrame(line, data) {
```
bomsy: nit: looks like data is not needed

jasonLaster: true... I'll move this file to a place where prettier works and then we'll find other terrible things.

PR #3649, opened 2017-08-12
Convert modal to jsx

nyrosmith:
Associated Issue: #3506 
### Summary of Changes
Changed Modal to JSX syntax

I am quite unsure about the "Slide"-component...

### DIFF
--- a/src/components/shared/Modal.js
+++ b/src/components/shared/Modal.js
@@ -29,14 +29,8 @@ export class Modal extends Component {
     const { status } = this.props;
 
     return (
-      <div
-        className= "modal-wrapper"
-        onClick= {this.props.handleClose}
-      >
-        <div
-          className= {classnames("modal", status)}
-          onClick= {this.onClick}
-        >
+      <div className="modal-wrapper" onClick={this.props.handleClose}>
+        <div className={classnames("modal", status)} onClick={this.onClick}>
           {this.props.children}
         </div>
       </div>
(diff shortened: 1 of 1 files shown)


src/components/shared/Modal.js
```diff
@@ -1,11 +1,9 @@
 // @flow
 
-import { DOM as dom, createFactory, Component, PropTypes } from "react";
+import React, { PropTypes, Component } from "react";
 import type { Children } from "react";
 import classnames from "classnames";
-import _Transition from "react-transition-group/Transition";
-const Transition = createFactory(_Transition);
-
+import Transition from "react-transition-group/Transition";
```
jasonLaster: is this used?

nyrosmith: I am really unsure.
Transition is used within Slide. Which is the default export.
I did a quick search but couldn't find any usage of Slide, but bc I am a total noob on your codebase this doesnot mean I didn't oversee something

nyrosmith:
I could comment out the Slide function and See if it breaks the bundle. But this would only work if you do not dynamics require()...

src/components/shared/Modal.js
wldcordeiro: Slide is a component. You can convert the default export to JSX and make this proper change.

wldcordeiro: Something like this:

```jsx
export default function Slide({
  in: inProp,
  children,
  handleClose
}: SlideProps) {
  return (
    <Transition id={inProp} timeout={175} appear>
      {status => <Modal status={status} handleClose={handleClose}>{children}</Modal>}
    </Transition>
  );
}
```

nyrosmith:
@wldcordeiro  @jasonLaster Updated based on your suggestions.
Thx for guidance!
Sorry if formatting doesnot fit but I am unable to run prettier...

Issue #3596, opened 2017-08-08
Angular svg is missing


jasonLaster:
it looks like debugging angular apps has an issue because it cannot find the angular SVG

STR:
1. http://todomvc.com/examples/angularjs/#/
2. set a breakpoint

notes:
- we don't have an angular svg https://github.com/devtools-html/debugger.html/blob/master/assets/images/Svg.js#L7-L10
- we find angular here https://github.com/devtools-html/debugger.html/blob/master/src/utils/frame.js#L74-L77

the fix is to add an angular svg or remove angular support here

we should also drop this `throw`

```diff
diff --git a/assets/images/Svg.js b/assets/images/Svg.js
index ee85367..4a4d304 100644
--- a/assets/images/Svg.js
+++ b/assets/images/Svg.js
@@ -54,8 +54,9 @@ const svg = {
 module.exports = function(name, props) {
   // eslint-disable-line
   if (!svg[name]) {
-    throw new Error("Unknown SVG: " + name);
+    console.warn(`Unknown SVG: ${name}`)
   }
+
   let className = name;
   ```

jasonLaster:
here's a good svg: https://commons.wikimedia.org/wiki/File:Angular_full_color_logo.svg#file

PR #3653, opened 2017-08-13
add angular icon

nyrosmith:
Associated Issue: #3596 

### Summary of Changes

* added existing ng svg to List
* replaced throwing error in dev mode with console.warn

### Test Plan

1. open http://todomvc.com/examples/angularjs/#/
2. set a breakpoint
3. verify that icon gets displayed

### DIFF
--- a/assets/images/Svg.js
+++ b/assets/images/Svg.js
@@ -4,6 +4,7 @@ const { isDevelopment } = require("devtools-config");
 
 const svg = {
   "angle-brackets": require("./angle-brackets.svg"),
+  angular: require("./angular.svg"),
   arrow: require("./arrow.svg"),
   backbone: require("./backbone.svg"),
   blackBox: require("./blackBox.svg"),
@@ -55,12 +56,9 @@ const svg = {
 module.exports = function(name, props) {
   // eslint-disable-line
   if (!svg[name]) {
-    const error = "Unknown SVG: " + name;
     if (isDevelopment()) {
-      throw new Error(error);
+      console.warn(`Unknown SVG: ${name}`);
     }
-
-    console.warn(error);
     return;
   }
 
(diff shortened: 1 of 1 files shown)


assets/images/Svg.js
```diff
@@ -4,6 +4,7 @@ const { isDevelopment } = require("devtools-config");
 
 const svg = {
   "angle-brackets": require("./angle-brackets.svg"),
+  angular: require("./angular.svg"),
```
jasonLaster: looks like the SVG wasn't included.

assets/images/Svg.js
```diff
@@ -55,12 +56,9 @@ const svg = {
 module.exports = function(name, props) {
   // eslint-disable-line
   if (!svg[name]) {
-    const error = "Unknown SVG: " + name;
     if (isDevelopment()) {
-      throw new Error(error);
```
jasonLaster: this is okay. It's fine to throw an error in development. The problem is if we try to do it in production. The old version threw an error all the time.

assets/images/Svg.js
nyrosmith: There is an existing angular.svg in the /assets folder which looked good to me.
Shall I replace it with the one you mentioned?

assets/images/Svg.js
nyrosmith: So shall I just revert this part and throw an error in dev mode & console.log?

assets/images/Svg.js
jasonLaster: oh no, i didn't realize that!

assets/images/Svg.js
jasonLaster: yeah - lets revert this part

nyrosmith:
@jasonLaster Just reverted the err throwing behavior as you suggested

PR #3656, opened 2017-08-14
Fix a couple mochi errors

jasonLaster:
### Summary of Changes

* don't try to re-assign a const
* show failed tests at the bottom
* properly indent

screen shot 2017-08-14 at 2 44 34 pm


bin/mochi.js
```diff
@@ -103,7 +113,8 @@ function onLine(line, data) {
   }
 
   if (data.mode == "done") {
-    return;
+    // console.log(line)
```
codehag: We can delete this

PR #3672, opened 2017-08-15
Convert ManagedTree to JSX

nyrosmith:
Associated Issue: #3506 
### Summary of Changes

* use JSX
* remove dead code

### Test Plan

I only opened ManagedTree on master & my branch and looked if it looks and behaves the same

### DIFF
--- a/src/components/shared/ManagedTree.js
+++ b/src/components/shared/ManagedTree.js
@@ -134,7 +134,11 @@ class ManagedTree extends Component {
     };
 
     const props = { ...this.props, ...overrides };
-    return dom.div({ className: "managed-tree" }, Tree(props));
+    return (
+      <div className="managed-tree">
+        <Tree {...props} />
+      </div>
+    );
   }
 }
 
--- a/src/components/shared/ManagedTree.js
+++ b/src/components/shared/ManagedTree.js
@@ -1,5 +1,5 @@
 // @flow
-import { createFactory, Component, DOM as dom } from "react";
+import React, { createFactory, Component } from "react";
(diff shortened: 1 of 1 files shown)


src/components/shared/ManagedTree.js
```diff
@@ -44,10 +44,6 @@ class ManagedTree extends Component {
       expanded: new Set(),
       focusedItem: null
     };
-
-    const self: any = this;
```
wldcordeiro: I don't think this is dead code. These are binding the methods to the prototype.

nyrosmith: Self is not used anywhere in this component except for the 2 lines below.
I think this should be `this.setExpanded= this.setExpanded.bind(this);` bc on Line 122 it calls it with `this`
`onExpand: item => this.setExpanded(item, true)`

nyrosmith: `this.setExpanded` is used within `render` which is autoBound so this should work
`this.setExpanded` is used inside of `expandListItems()` which is called by `componentWillReceiveProps` (= autoBound)

So I think this is legit and should work. or am I completly wrong? @wldcordeiro @jasonLaster

wldcordeiro: The line defining the `self` constant I think is there to prevent Flow bugs? I know it's a pattern we use often when binding our methods.

nyrosmith: I'll revert it to before state but I am quite sure my changes work

nyrosmith:
@wldcordeiro I changed the self block back.

Issue #3669, opened 2017-08-15
Stop using pointer cursor for buttons


jasonLaster:
Pointer cursor isn't used in other DevTools Tabs. Cursor style should be consistent with other DevTools tabs.

link: https://bugzilla.mozilla.org/show_bug.cgi?id=1390062

MayankSrivastav:
Hi, I would like to take it up.
I guess we are talking about the file-name tabs in the editor panel? The pointer there should be an arrow pointer?
File to be modified seems like Tabs.css in Editor component?

jasonLaster:
@MayankSrivastav great. It's yours.

Yeah, a lot of these should be removed I guess

<img src='https://shipusercontent.com/ba90c69035288dd19cba29ba424d168b/Screen%20Shot%202017-08-16%20at%2011.40.44%20AM.png' title='Screen Shot 2017-08-16 at 11.40.44 AM.png' width=1091>

MayankSrivastav:
Ok. so, cursor: pointer needs to be removed from every css file. pointer should not be used at all.

jasonLaster:
that's what i think

PR #3686, opened 2017-08-16
Removed pointer cursor for buttons

MayankSrivastav:
Associated Issue: #3669

### Summary of Changes
- Removed the "cursor: pointer" css property from all the css files. Now all the buttons use default pointer arrow

### DIFF
--- a/src/components/App.css
+++ b/src/components/App.css
@@ -17,10 +17,6 @@
   width: 100%;
 }
 
-.subsettings:hover {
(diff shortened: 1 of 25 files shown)


MayankSrivastav:
Thanks @codehag. Is it required to be assigned to me? It's not assigned by the way.

codehag:
It should be fine! thanks again!

Issue #2426, opened 2017-03-22
[windows] yarn nom fails on windows


jasonLaster:
`yarn nom` does `rm -rf node_modules yarn.lock; yarn install` which is UNIXy and fails on windows. This should be easy to add and if the script passes we can assume that it will work in windows.

We should switch to using a node script.

```diff
diff --git a/bin/nom.js b/bin/nom.js
new file mode 100644
index 0000000..013a48f
--- /dev/null
+++ b/bin/nom.js
@@ -0,0 +1,9 @@
+const rimraf = require("rimraf");
+const shell = require("shelljs");
+const path = require("path");
+
+const projectPath = path.resolve(__dirname, "..")
+
+rimraf(path.resolve(projectPath, "node_modules"), {}, () => {});
+rimraf(path.resolve(projectPath, "yarn.lock"), {}, () => {});
+shell.exec("yarn")
diff --git a/package.json b/package.json
index 627e276..7cc20fd 100644
--- a/package.json
+++ b/package.json
@@ -48,7 +48,7 @@
     "postrewrite": "node bin/post-rewrite.js",
     "precommit": "lint-staged",
     "prepush": "run-p lint test flow",
-    "nom": "rm -rf node_modules yarn.lock; yarn install",
+    "nom": "node bin/nom.js",
     "build-storybook": "build-storybook"
   },
   "dependencies": {
```

clarkbw:
Maybe we could do this as a node script instead?

rohanprasad:
@jasonLaster Is this resolved after #3711  ?

nyrosmith:
@rohanprasad this is solved. Are you using windows?

rohanprasad:
@nyrosmith Nope, I was looking into open issues. If the issue is resolved it can be closed. 🙂

nyrosmith:
@rohanprasad  I'd like to but I don't have permissions

PR #3711, opened 2017-08-18
Make nom script work cross platform

nyrosmith:
Associated Issue: #2426

I was only able to test nom-script on Windows so this should be checked on Mac before merging.


nyrosmith:
That would be great bc I cannot test other platforms than Windows. Strange because I really thought everything is correct...

codehag:
Lets pair when you have time, ping me on slack when you are ready -- @yulia is my handle on there!

codehag:
great work!

Issue #3704, opened 2017-08-17
[outline] the outline view should scroll


jasonLaster:
The outline view needs a tad bit more polish, such as 
* the list items should have some more padding
* the list should scroll
* the tab colors at the bottom feel inverted

```diff

diff --git a/src/components/PrimaryPanes/Outline.css b/src/components/PrimaryPanes/Outline.css
index e93d9ae..5ff0198 100644
--- a/src/components/PrimaryPanes/Outline.css
+++ b/src/components/PrimaryPanes/Outline.css
@@ -2,12 +2,14 @@
   list-style-type: none;
   padding-left: 0px;
   width: 100%;
+  overflow-y: auto;
 }

 .outline-list__element {
   color: blue;
   padding-left: 1rem;
   padding-right: 0.5rem;
+  line-height: 1.5em;
   cursor: pointer;
 }
 ```

<img src='https://shipusercontent.com/a823c8e4de40b7dc55df40059d46e06c/Screen%20Shot%202017-08-17%20at%205.08.49%20PM.png' title='Screen Shot 2017-08-17 at 5.08.49 PM.png' width=434>
<img src='https://shipusercontent.com/6cbbac33b52eee09116d2a377b4e5baa/Screen%20Shot%202017-08-17%20at%205.08.38%20PM.png' title='Screen Shot 2017-08-17 at 5.08.38 PM.png' width=251>

PR #3712, opened 2017-08-18
Added scroll to outline view

sagorika1996:
Associated Issue: #3704 

### Summary of Changes
*  Added scroll to list
* Added padding to list elements
* Inverted tab colors at the bottom 

outline


src/components/PrimaryPanes/Outline.css
```diff
@@ -6,9 +10,11 @@
 
 .outline-list__element {
   color: blue;
+  cursor: pointer;
```
codehag: I think we are removing `cursor: pointer` at the moment, so we can take out this line (reference issue: https://github.com/devtools-html/debugger.html/issues/3669)

PR #3713, opened 2017-08-18
copy selected source from codemirror

nyrosmith:
Associated Issue: #3274 

I don't know which accessKey to use....

copy

### DIFF
--- a/assets/panel/debugger.properties
+++ b/assets/panel/debugger.properties
@@ -14,6 +14,10 @@
 # that collapses the left and right panes in the debugger UI.
 collapsePanes=Collapse panes
 
+# LOCALIZATION NOTE (copy): This is the text that appears in the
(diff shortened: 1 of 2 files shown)


nyrosmith:
CopySourceLabel has an associated copySourceUrlKey which references copySourceUrl.accesskey
Sorry I am not able to do correct formatting rn or link the source lines bc I am on Mobile right now

codehag:
Hmm I am not sure I have the best advice for this. It's not clear to me what the accessKey is for, even though I see it everywhere, but also I see conflicts; ie several use the same key. @jasonLaster can you help out here? Maybe we could add an internationalization doc to make it clearer what the accessKey does and how its used.

codehag:
other than that, this looks good to me

PR #3727, opened 2017-08-21
add await stepping

jbhoosreddy:
Associated Issue: #3106

### Summary of Changes

### Test Plan

- [x] Visual Testing

### DIFF
--- a/src/actions/pause.js
+++ b/src/actions/pause.js
@@ -15,6 +15,7 @@ import { evaluateExpressions } from "./expressions";
 import { addHiddenBreakpoint, removeBreakpoint } from "./breakpoints";
 import { getHiddenBreakpointLocation } from "../reducers/breakpoints";
 import * as parser from "../utils/parser";
+import { features } from "../utils/prefs";
 
 import type { Pause, Frame } from "../types";
 import type { ThunkArgs } from "./types";
@@ -243,6 +244,10 @@ export function loadObjectProperties(object: any) {
  */
 export function astCommand(stepType: string) {
   return async ({ dispatch, getState, sourceMaps }: ThunkArgs) => {
+    if (!features.asyncStepping) {
+      return dispatch(command({ type: stepType }));
+    }
+
     const pauseInfo = getPause(getState());
     const source = getSelectedSource(getState()).toJS();
     const currentHiddenBreakpointLocation = getHiddenBreakpointLocation(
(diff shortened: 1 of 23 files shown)


src/actions/pause.js
```diff
@@ -225,3 +235,35 @@ export function loadObjectProperties(object: any) {
...
+      return dispatch(command({ type: stepType }));
+    }
+
+    const pauseInfo = getPause(getState());
+    const source = getSelectedSource(getState()).toJS();
+    const currentHiddenBreakpointLocation = getHiddenBreakpointLocation(
+      getState()
+    );
+    if (currentHiddenBreakpointLocation) {
+      dispatch(removeBreakpoint(currentHiddenBreakpointLocation));
```
codehag: interesting!

src/actions/pause.js
```diff
@@ -225,3 +235,35 @@ export function loadObjectProperties(object: any) {
...
   };
 }
+
+/**
+ * @memberOf actions/pause
+ * @static
+ * @param stepType
+ * @returns {function(ThunkArgs)}
+ */
+export function astCommand(stepType: string) {
```
codehag: why do we call this astCommand?

src/actions/pause.js
```diff
@@ -225,3 +235,35 @@ export function loadObjectProperties(object: any) {
...
+    if (!features.asyncStepping) {
+      return dispatch(command({ type: stepType }));
+    }
+
+    const pauseInfo = getPause(getState());
+    const source = getSelectedSource(getState()).toJS();
+    const currentHiddenBreakpointLocation = getHiddenBreakpointLocation(
+      getState()
+    );
+    if (currentHiddenBreakpointLocation) {
```
codehag: getHiddenBreakpointLocation seems to get the first hidden breakpoint, do we ever have a case of more than one hidden breakpoint? could this give us a false positive?

src/actions/pause.js
jbhoosreddy: I called it astCommand, because it does similar things to command action. but it tries to use the parser AST to evaluate which command should be executed.

src/actions/pause.js
jbhoosreddy: at the moment, this only adds 1 hidden breakpoint in the system. and it removes hidden breakpoint on next step. so there's only 1 hidden breakpoint.

src/actions/pause.js
```diff
@@ -243,6 +244,10 @@ export function loadObjectProperties(object: any) {
  */
 export function astCommand(stepType: string) {
   return async ({ dispatch, getState, sourceMaps }: ThunkArgs) => {
+    if (!features.asyncStepping) {
```
jbhoosreddy: that's one way to fix it. 🤣

Issue #3732, opened 2017-08-21
Add hover effect to "add new tab" button


codehag:
We have removed the hover pointer from all of our elements, however we should have some indication of what is clickable.

@MayankSrivastav had suggestion of adding the same hover color as we have for our other buttons (ie stepping) and I think this is a great idea

![pasted image at 2017_08_21 10_28 pm](https://user-images.githubusercontent.com/26968615/29530894-3c52ccb6-86a6-11e7-86ec-4509ba6c37be.png)

rohanprasad:
@codehag Working on it. Can you think of other places I could do the same ?

codehag:
Done!

PR #3741, opened 2017-08-22
Add hover color to new-tab button

rohanprasad:
Associated Issue: https://github.com/devtools-html/debugger.html/issues/3732

### Summary of Changes

* Added background-color on hover for new-tab button. 
* Changed padding and margin for cleaner look on hover.

### Screenshot
add-btn-hover

### DIFF
--- a/src/components/Editor/Tabs.css
+++ b/src/components/Editor/Tabs.css
@@ -13,7 +13,8 @@
 
 .source-header .new-tab-btn {
   padding: 4px;
-  margin: 4px 2px 2px 2px;
+  margin-top: 4px;
+  margin-left: 3px;
   fill: var(--theme-content-color3);
   transition: 0.1s ease;
   align-self: center;
@@ -25,6 +26,7 @@
 
 .source-header .new-tab-btn svg {
   width: 12px;
+  display: block;
 }
 
 .source-tabs {
--- a/src/components/Editor/Tabs.css
+++ b/src/components/Editor/Tabs.css
@@ -14,7 +14,7 @@
 .source-header .new-tab-btn {
   padding: 4px;
   margin-top: 4px;
-  margin-left: 3px;
+  margin-left: 2px;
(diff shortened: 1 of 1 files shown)


codehag:
Thanks for the work Rohan! I noticed in testing that the button is oddly placed
before: 
screen shot 2017-08-22 at 5 15 53 pm

After:
screen shot 2017-08-22 at 5 14 45 pm

It wasn't centered before, but now I noticed it. I think we can fix it by adding the following to the css you wrote:

```diff
diff --git a/src/components/Editor/Tabs.css b/src/components/Editor/Tabs.css
index e2903f85..4e2e2fa4 100644
--- a/src/components/Editor/Tabs.css
+++ b/src/components/Editor/Tabs.css
@@ -11,9 +11,13 @@
   user-select: none;
 }

+.source-header .new-tab-btn svg {
+  display: block;
+}
+
 .source-header .new-tab-btn {
   padding: 4px;
-  margin: 4px 2px 2px 2px;
+  margin: 4px 0px 0px 3px;
   fill: var(--theme-content-color3);
   transition: 0.1s ease;
   align-self: center;
```

a couple of notes:
- you will notice that i set the svg to display block. This is because browser display images with a small border at the bottom -- this is what was making it look a little off
- there appears to be another margin at play on the bottom, and the right margin probably isnt necessary

Can you play around with it a bit and see what looks best to you? thanks!

rohanprasad:
@codehag Thanks for the feedback. I didn't know about the display hack, added that. I tried margins and padding for the button but somehow this looks the best.

codehag:
hmm i quite like the curved shape... but for now lets leave it sharp (since everything else has sharp edges). we can revisit this later. Thanks for the work!

codehag:
hmm i quite like the curved shape... but for now lets leave it sharp (since everything else has sharp edges). we can revisit this later. Thanks for the work!

PR #3742, opened 2017-08-22
Access Keys

jasonLaster:
add some docs for access keys

### DIFF
--- a/docs/local-development.md
+++ b/docs/local-development.md
@@ -231,8 +231,8 @@ Notes:
 
 #### Access Keys
 
-Access Keys are keyboard shortcuts for an item in the context menu.
-Access Keys are defined in the properties file next to the menu item's string.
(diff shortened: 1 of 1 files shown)


nyrosmith:
Can you add a section about what to do if there is no accesskey defined?
Eg for this case: https://github.com/devtools-html/debugger.html/pull/3713

codehag:
I am wondering if there should be more detail here. I am doing it in list form because it might be easier to look at. 

we should add: 
- "Access keys are per context menu" 

questions: 
- is there a convention of which keys should be used? or a way to know which keys can be used, or which are off limits (ie used in the application)? 
- Does the context menu take precedence over other keyboard controls when it is open?

I can add these once i get feedback from you!

jasonLaster:
no convention that i know of. We try to think of a good char like "r" for remove.

i believe the CM takes precedence

Issue #3626, opened 2017-08-10
[Preview] don’t show loading value


jasonLaster:
We currently have an awkward loading experience with preview popups. We show the expression while we're loading the value. We probably should have a simpler loading mode.

A good start would be to have an empty popup :)

codehag:
I can confirm that #3749 shows an empty popup while we are loading the value. do we want to do a step 2 to this?

Issue #3589, opened 2017-08-07
[Preview] Click and drag in editor still shows the value preview


gregtatum:
Steps to reproduce:
 * Double click on a word to highlight it.
 * Drag around to highlight a selection.
 * Repeat

Expected result:
 * I should get an easy to highlight line of text, where I can easily control the selection.

Actual result:
 * The preview pop up is still active, causing either my selection to drop (when double clicking), or for the preview to get in the way of the text that I'm selecting.

Example image of attempting a selection:
<img width="748" alt="image" src="https://user-images.githubusercontent.com/1588648/29040952-a48fac26-7b76-11e7-98d9-d76d57fb6f49.png">

jasonLaster:
diff --git a/src/components/Editor/Preview.css b/src/components/Editor/Preview.css
index 5450120..f98e2ea 100644
--- a/src/components/Editor/Preview.css
+++ b/src/components/Editor/Preview.css
@@ -33,6 +33,8 @@
   cursor: pointer;
 }

+/* can we do something like this: :not(::selection) */
+
 .selection,
 .debug-expression.selection {
   background-color: var(--theme-highlight-yellow);
(END)

codehag:
fixed by #3749

Issue #3264, opened 2017-07-05
[Editor] Refactor CodeMirror Children


jasonLaster:
In SF we discussed a strategy for compartmentalizing our codemirror editor logic in terms of HoCs. This is the strategy we use for breakpoints and highlighted lines

#### Tasks: 

- [x] Breakpoints [file](https://github.com/asolove/debugger.html/blob/editor-components/src/components/Editor/Breakpoint.js#L1)
- [x] Highlight Lines [file](https://github.com/asolove/debugger.html/blob/editor-components/src/components/Editor/HighlightLines.js#L1)
- [x] Debug Line [pr](https://github.com/devtools-html/debugger.html/pull/3255)
- [x] preview popup [pr](https://github.com/devtools-html/debugger.html/pull/3749)
- [ ] Text [gist](https://gist.github.com/jasonLaster/fe6ce6af67e1ed351cfdc803f56c5db4)
- [ ] HighlightLine [gist](https://gist.github.com/5cc88075446671ac56c13665352e1dac)
- [ ] Conditional BP Panel [gist](https://gist.github.com/0c4e04ce5bcb16b9e7bd91012b024718) 
- [ ] Gutter Menu might be nice too

#### Rationale:

This will help untangle some logic and improve readability and testing.

jasonLaster:
@asolove and I talked offline and had some interesting takeaways:

1. CodeMirror children implicitely depend on the codemirror document being set. i.e. breakpoints should be added after the document is set. This means that the Editor document should be set in component will receive props and the breakpoints should be set in in component did update. We can have some assertions that guard against mistakes.

2. CodeMirror perf comes from using codemirror's operation method. The API is a callback `cm.operation(() => {})`. It would be nice to use CM's two low level methods startOperation and endOperation.

3. It's possible that two children might have some state dependencies that will need to move up to redux. This is generally nice. With that said, we should be careful because sometimes we fire actions that don't change the state.

jasonLaster:
I'm going to close this issue because we're continuing to make progress on refactoring the editor.

Our plan is to refactor a feature before we work on it. We did this w/ debugLine and it worked well.

Most of our tracking issues are for large features or lots of small refactorings. This is a large code health project with a lot of risk, so it doesn't feel right to keep a tracking issue open for it.

PR #3749, opened 2017-08-22
Improve preview

jasonLaster:
Associated Issue: #3626 #3589 #3264

### Summary of Changes

* Don't show loading state
* Don't show preview when selecting text
* Extract Preview logic from editor
* rename selection to preview because it was confusing

### DIFF
--- a/src/components/Editor/Preview/Popup.js
+++ b/src/components/Editor/Preview/Popup.js
@@ -89,9 +89,7 @@ export class Popup extends Component {
         className="preview-popup"
         onClick={() => selectSourceURL(location.url, { line: location.line })}
       >
-        {true
-          ? previewFunction(value)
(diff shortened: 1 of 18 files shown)


src/components/Editor/Preview/Popup.js
```diff
@@ -85,25 +86,27 @@ export class Preview extends Component {
 
     return (
       <div
-        className="preview"
+        className="preview-popup"
         onClick={() => selectSourceURL(location.url, { line: location.line })}
       >
-        {previewFunction(value)}
+        {true
```
codehag: this will always return `previewFunction(value)`; maybe we should just use it directly?

jasonLaster: sure

PR #3751, opened 2017-08-22
document CI and process

jasonLaster:


codehag:
I added a few things (for my reference) let me know if they need a bit of work @jasonLaster

PR #3752, opened 2017-08-22
switch to janitor

jasonLaster:
This would switch us from my custom docker image which is really old to janitor, which would be more up to date:

1. we need to get the image
2. we need to add the config
...

In the future we could switch to circle 2 and use this image as the base

### DIFF
--- a/bin/run-mochitests-docker
+++ b/bin/run-mochitests-docker
@@ -17,5 +17,11 @@ docker run -it \
   -v "/tmp/.X11-unix:/tmp/.X11-unix:rw" \
   -e "DISPLAY=unix$DISPLAY" \
   --ipc host \
-  jryans/debugger-gecko \
-  /bin/bash -c "export SHELL=/bin/bash; touch devtools/client/debugger/new/test/mochitest/browser.ini && ./mach build && ./mach mochitest --subsuite devtools devtools/client/debugger/new/test/mochitest/"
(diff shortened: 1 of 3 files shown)


bin/run-mochitests-docker
```diff
@@ -17,5 +17,11 @@ docker run -it \
   -v "/tmp/.X11-unix:/tmp/.X11-unix:rw" \
   -e "DISPLAY=unix$DISPLAY" \
   --ipc host \
-  jryans/debugger-gecko \
-  /bin/bash -c "export SHELL=/bin/bash; touch devtools/client/debugger/new/test/mochitest/browser.ini && ./mach build && ./mach mochitest --subsuite devtools devtools/client/debugger/new/test/mochitest/"
+  janx/firefox-hg \
+  /bin/bash -c "export SHELL=/bin/bash; \
+    echo \" \
+      ac_add_options --enable-artifact-builds\n \
+      mk_add_options MOZ_OBJDIR=./objdir-frontend\n \" > .mozconfig \
```
nt1m: I think this would probably cause a conflict, since firefox-hg adds `mozconfig` while you're adding `.mozconfig`

Both are supported, but you should only have one of the two, so you should probably override `mozconfig` here.

jasonLaster:
I think we're going to go w/ try's image

Issue #3150, opened 2017-06-09
[Breakpoints] AST Breakpoints


jasonLaster:
Here is a [writeup] on AST Breakpoints

[writeup]:https://docs.google.com/document/d/1o3xVV_rTe-oxH7ujYUJ4a_XKSdPo1udVO7rKAfdas44/edit

### TODO
- [x] astLocation util that uses symbols
- [x] add astLocation field
- [x] add astLocation to addBreakpoint
- [x] clear ast tree [BUG] 
- [x] update syncBreakpoint to delete server breakpoints on sync if an ast breakpoint location is valid
- [x] update syncBreakpoiint to add an ast located breakpoint if available
- [ ] mochi tests

codehag:
Hiya!

I was playing around with this already. here is some very messy old code (in case anyone is looking for a starting point)

```diff
diff --git a/src/actions/breakpoints.js b/src/actions/breakpoints.js
index 32be0377..45587b50 100644
--- a/src/actions/breakpoints.js
+++ b/src/actions/breakpoints.js
@@ -12,6 +12,7 @@ import { PROMISE } from "../utils/redux/middleware/promise";
 import { getBreakpoint, getBreakpoints, getSource } from "../selectors";
 import { originalToGeneratedId } from "devtools-source-map";
 import { equalizeLocationColumn } from "../utils/breakpoint";
+import { getClosestPath } from "../utils/parser/utils/closest";

 import type { ThunkArgs } from "./types";
 import type { PendingBreakpoint, Location } from "../types";
@@ -220,11 +221,54 @@ export function addBreakpoint(
     }

     const breakpoint = _createBreakpoint(location, { condition });
+    const bp = _getOrCreateBreakpoint(getState(), location, condition);
+
+    let updatedLocation;
     return dispatch({
       type: "ADD_BREAKPOINT",
       breakpoint,
       condition: condition,
-      [PROMISE]: addClientBreakpoint(getState(), client, sourceMaps, breakpoint)
+      // NOT TO BE USED: just plaing around with AST
+      [PROMISE]: (async function() {
+        if (sourceMaps.isOriginalId(bp.location.sourceId)) {
+          const source = getSource(getState(), bp.location.sourceId);
+          updatedLocation = await sourceMaps.getGeneratedLocation(
+            bp.location,
+            source.toJS()
+          );
+        }
+
+        let { id, actualLocation, hitCount } = await client.setBreakpoint(
+          updatedLocation,
[...]

PR #3753, opened 2017-08-22
add AST breakpoint util

codehag:
Associated Issue: #3150 

This is part 1, adding a util for AST breakpoint locations
and part 2, adding the `astLocation` field


src/utils/breakpoint/astBreakpointLocation.js
```diff
@@ -0,0 +1,28 @@
...
+  const symbols = await getSymbols(_source);
+  const functions = symbols.functions;
+
+  const scope = functions.find(node => {
+    return containsPosition(node.location, location);
+  });
+  if (scope) {
+    const line = location.line - scope.location.start.line;
+    const column = location.column;
+    return { scope, offset: { line, column } };
```
jasonLaster: i think we should be more explicit here and just keep the function name

src/utils/breakpoint/astBreakpointLocation.js
```diff
@@ -0,0 +1,28 @@
+import { getSymbols } from "../parser";
+import { containsPosition } from "../parser/utils/helpers";
+
+export async function getASTLocation(source, location) {
+  const _source = source.toJS ? source.toJS() : source;
+  const symbols = await getSymbols(_source);
+  const functions = symbols.functions;
+
+  const scope = functions.find(node => {
```
jasonLaster: we might want to check to see if the function has a name

src/utils/breakpoint/tests/astBreakpointLocation.spec.js
```diff
@@ -0,0 +1,34 @@
+import { getASTLocation } from "../astBreakpointLocation.js";
+import { getSource } from "../../parser/tests/helpers";
+
+describe("ast", () => {
+  describe("valid location", () => {
+    it("returns the scope and offset", async () => {
+      const source = getSource("math");
+      // mock out immutable.js
+      source.toJS = () => source;
```
jasonLaster: lets just do `new Map` which we do in other cases

src/utils/breakpoint/tests/astBreakpointLocation.spec.js
```diff
@@ -0,0 +1,34 @@
...
+      const location = { line: 6, column: 0 };
+      const astLocation = await getASTLocation(source, location);
+      const offset = astLocation.scope.location.start;
+      expect(astLocation.scope.name).toBe("math");
+      expect(astLocation.offset.line).toBe(location.line - offset.line);
+      expect(astLocation.offset.column).toBe(location.column);
+      expect(astLocation).toMatchSnapshot();
+    });
+  });
+  describe("invalid location", () => {
```
jasonLaster: what's invalid? like not inside a function.

src/utils/breakpoint/tests/astBreakpointLocation.spec.js
```diff
@@ -0,0 +1,34 @@
...
+
+      const location = { line: 6, column: 0 };
+      const astLocation = await getASTLocation(source, location);
+      const offset = astLocation.scope.location.start;
+      expect(astLocation.scope.name).toBe("math");
+      expect(astLocation.offset.line).toBe(location.line - offset.line);
+      expect(astLocation.offset.column).toBe(location.column);
+      expect(astLocation).toMatchSnapshot();
+    });
+  });
```
jasonLaster: lets add a test for nested functions and just lock down a behavior. I think we want to use the closest named function

src/actions/breakpoints/syncBreakpoint.js
```diff
@@ -9,11 +9,12 @@ import { originalToGeneratedId } from "devtools-source-map";
 // we have three forms of syncing: disabled syncing, existing server syncing
 // and adding a new breakpoint
 export async function syncClientBreakpoint(
-  sourceId: string,
+  source: Object,
   client,
   sourceMaps,
   pendingBreakpoint: PendingBreakpoint
 ) {
+  const sourceId = source.id;
```
jasonLaster: you cheated and added more than just the util :)

src/actions/breakpoints/addBreakpoint.js
```diff
@@ -56,12 +57,15 @@ export default async function addBreakpoint(
     newGeneratedLocation
   );
 
+  const astLocation = await getASTLocation(source, newLocation);
```
jasonLaster: you cheated and added more than just the Util :)

src/utils/breakpoint/astBreakpointLocation.js
codehag: hmm 
pros to removing everything but the name: less to save in the persisted data
cons to removing everything but the name: we would have to calculate the parents position to compare it to the new location

src/utils/breakpoint/tests/astBreakpointLocation.spec.js
codehag: yes

src/actions/breakpoints/addBreakpoint.js
codehag: i also added the field ;)

src/utils/breakpoint/tests/astBreakpointLocation.spec.js
codehag: or: in the global scope. I need to do a bit more work on this one maybe

src/utils/breakpoint/astBreakpointLocation.js
codehag: we shouldnt save an ast if the function is anonymous maybe? this is used only when there is a name

src/actions/breakpoints.js
```diff
@@ -40,16 +40,16 @@ type addBreakpointOptions = {
  * @param {PendingBreakpoint} $1.location PendingBreakpoint  value
  */
 export function syncBreakpoint(
-  sourceId: string,
+  source: Object,
```
jasonLaster: can we take these changes out?

src/utils/breakpoint/astBreakpointLocation.js
```diff
@@ -0,0 +1,26 @@
...
+  const symbols = await getSymbols(source);
+  const functions = symbols.functions;
+
+  const scope = functions.find(node => {
+    return containsPosition(node.location, location) && node.name;
+  });
+  if (scope) {
+    const line = location.line - scope.location.start.line;
+    const column = location.column;
+    return { scope, offset: { line, column } };
```
jasonLaster: we'll have to do that anyways because the function could be in a new location after reloading

src/actions/breakpoints/syncBreakpoint.js
jasonLaster: can se take this out too?

src/actions/sources.js
```diff
@@ -62,7 +62,7 @@ async function checkPendingBreakpoint(
   const sameSource = sourceUrl && sourceUrl === source.url;
 
   if (sameSource) {
-    await dispatch(syncBreakpoint(source.id, pendingBreakpoint));
+    await dispatch(syncBreakpoint(source, pendingBreakpoint));
```
jasonLaster: lets take this out too :)

Issue #3490, opened 2017-07-28
[Expressions] We should not store the values


jasonLaster:
We're currently getting warnings in the browser console because we're storing expression values in prefs which could be big

```
Warning: attempting to write 15956 bytes to preference devtools.debugger.expressions. This is bad for general performance and memory usage. Such an amount of data should rather be written to an external file. This preference will not be sent to any content processes.
```

The fix can go here

```diff

diff --git a/src/reducers/expressions.js b/src/reducers/expressions.js
index 81f3df0..e2c855e 100644
--- a/src/reducers/expressions.js
+++ b/src/reducers/expressions.js

@@ -69,6 +70,7 @@ function storeExpressions(state) {
     .getIn(["expressions"])
     .filter(e => e.visible)
     .toJS();
+    // fix this
 }
```

PR #3759, opened 2017-08-23
Expression values are dropped

jasonLaster:
Associated Issue: #3490

### Summary of Changes
omit the value field

### DIFF
--- a/src/reducers/expressions.js
+++ b/src/reducers/expressions.js
@@ -2,6 +2,7 @@
 
 import makeRecord from "../utils/makeRecord";
 import { List } from "immutable";
+import { omit } from "lodash";
 import { createSelector } from "reselect";
 import { prefs } from "../utils/prefs";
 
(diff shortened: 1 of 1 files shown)


src/test/mochitest/examples/doc-scripts.html
```diff
@@ -15,7 +15,7 @@
       // This inline script allows this HTML page to show up as a
       // source. It also needs to introduce a new global variable so
       // it's not immediately garbage collected.
-      function inline_script() { var x = 5; }
+      inline_script = function () { var x = 5; };
```
codehag: why is is this changed?

codehag:
It looks like this is on top of another commit.. so its hard to see what is going on. Can you make this pr with just the changed expression?

src/reducers/expressions.js
```diff
@@ -65,10 +66,16 @@ function restoreExpressions() {
...
 
 function storeExpressions(state) {
-  prefs.expressions = state
+  const expressions = state
     .getIn(["expressions"])
     .filter(e => e.visible)
     .toJS();
+
+  const formattedExpressions = expressions.map(expression =>
+    omit(expression, "value")
```
codehag: :+1:

src/reducers/expressions.js
```diff
@@ -62,7 +63,16 @@ function restoreExpressions() {
 }
 
 function storeExpressions(state) {
-  prefs.expressions = state.getIn(["expressions"]).toJS();
+  const expressions = state
+    .getIn(["expressions"])
+    .filter(e => e.visible)
```
codehag: We need to remove this filter, as we no longer have the visible field :)

PR #3760, opened 2017-08-23
Fix #3729 - Ensure 'Reveal in Tree' always works, even if source doesn't change

### LINKED ISSUE #3729: File location is not always revealed in the Tree menu
[Steps to reproduce]:
1. Launch Firefox.
2. Access the https://devtools-html.github.io/debugger-examples/examples/todomvc/ webpage.
3. Open the debugger.
4. Open the todos.js file.
5. Close the Tree view.
6. Click the "Reaveal in Tree" option for the todos.js file.
7. Close the Tree view again.
8. Click the "Reveal in Tree" option for the todos.js file.

[Expected result]:
The file's location is revealed each time in the tree menu.

[Actual result]:
It seems that the file location can be revealed only once in the tree menu.

darkwing:
Associated Issue: #3729

### Summary of Changes

* Adding `clearShownSource` to allow reveal in source tree to work
* Calling explicitly in only two spots so it's not a burden on the rest of the app

### Test Plan

1.  Open a few sources
2.  Toggle a few arrows in the source tree closed, so as to obscure the the path to sources
3.  Right-click a tab, select "Reveal in Tree", and the tree should open to the currently focused source
4.  Right-click the source's content/code area, select "Reveal in Tree", and the tree should open to the currently focused source

### Notes
I like this approach because it's very explicit and only happens in two spots.  Thus `clearShownSource` doesn't have side effects for other, unrelated tasks

### DIFF
--- a/src/components/Editor/EditorMenu.js
+++ b/src/components/Editor/EditorMenu.js
@@ -1,7 +1,6 @@
 import { showMenu } from "devtools-launchpad";
 import { isOriginalId } from "devtools-source-map";
 import { copyToTheClipboard } from "../../utils/clipboard";
-import { clearShownSource } from "../../actions/ui";
 
 function getMenuItems(
   event,
@@ -79,10 +78,7 @@ function getMenuItems(
     label: revealInTreeLabel,
     accesskey: revealInTreeKey,
     disabled: false,
-    click: () => {
-      clearShownSource();
-      showSource(selectedSource.get("id"));
-    }
+    click: () => showSource(selectedSource.get("id"))
   };
 
   if (selectedSource && selectedSource.get("isBlackBoxed")) {
--- a/src/components/Editor/EditorMenu.js
+++ b/src/components/Editor/EditorMenu.js
@@ -1,6 +1,7 @@
 import { showMenu } from "devtools-launchpad";
 import { isOriginalId } from "devtools-source-map";
 import { copyToTheClipboard } from "../../utils/clipboard";
+import { clearShownSource } from "../../actions/ui";
 
 function getMenuItems(
   event,
@@ -78,7 +79,10 @@ function getMenuItems(
     label: revealInTreeLabel,
     accesskey: revealInTreeKey,
     disabled: false,
-    click: () => showSource(selectedSource.get("id"))
+    click: () => {
+      clearShownSource();
+      showSource(selectedSource.get("id"));
+    }
   };
 
   if (selectedSource && selectedSource.get("isBlackBoxed")) {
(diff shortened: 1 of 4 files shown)


src/components/Editor/EditorMenu.js
```diff
@@ -78,7 +79,10 @@ function getMenuItems(
     label: revealInTreeLabel,
     accesskey: revealInTreeKey,
     disabled: false,
-    click: () => showSource(selectedSource.get("id"))
+    click: () => {
+      clearShownSource();
```
jasonLaster: this could be passed in with `showSource`

src/components/Editor/EditorMenu.js
```diff
@@ -78,7 +79,10 @@ function getMenuItems(
...
+      showSource(selectedSource.get("id"));
```
jasonLaster: If we want to always clear, and then show the source we should move this logic to the showSource action...

In that case we would dispatch two actions one that cleared the source and the other that set it.

jasonLaster:
```diff
diff --git a/src/actions/ui.js b/src/actions/ui.js
index f6dbd01..86cc577 100644
--- a/src/actions/ui.js
+++ b/src/actions/ui.js
@@ -63,6 +63,12 @@ export function toggleFileSearchModifier(modifier: string) {
 export function showSource(sourceId: string) {
   return ({ dispatch, getState }: ThunkArgs) => {
     const source = getSource(getState(), sourceId);
+
+    dispatch({
+      type: "SHOW_SOURCE",
+      sourceUrl: ""
+    });
+
     dispatch({
       type: "SHOW_SOURCE",
       sourceUrl: source.get("url")
```

src/components/Editor/EditorMenu.js
codehag: i feel like this way, its clearer what is going on and each one function has its own responsibility. makes it easier to compose -- i vote for keeping it!

codehag: ah although, we could listen to show source and clear it. thats also true -- i like your pattern of doing this on the reducer :+1:

darkwing:
Updated to always clear, as suggested; let me know if I've misinterpreted your comments.

jasonLaster:
Does that fix the bug?

I don't think that'll work it because the sources only re-render when an action triggers the store to change.

My example above w/ `showSource` should work...

darkwing:
Oh sorry, I didn't even see your sample

darkwing:
But yes, it does appear to work ;)

darkwing:
Updated.  Looks like code coverage didn't kick off; would rebase but want to preserve history until approved.  Can CodeCoverage be kicked?

codehag:
No worries about code coverage! doing a quick test :)

PR #3764, opened 2017-08-23
Caches line formatter type; resizes gutter after the change.

yurydelendik:
Associated Issue: #3383

Here's the Pull Request Doc

### Summary of Changes

* Caches line formatter type, only for wasm source
* Resizes gutter after the change to avoid baby bp

### Test Plan

Example test plan as listed at #3383

### DIFF
--- a/src/utils/editor/source-documents.js
+++ b/src/utils/editor/source-documents.js
@@ -4,6 +4,8 @@ import { getMode } from "../source";
 
 import type { Source } from "debugger-html";
 import { isWasm, getWasmLineNumberFormatter, renderWasmText } from "../wasm";
+import { SourceEditorUtils } from "devtools-source-editor";
+const { resizeBreakpointGutter } = SourceEditorUtils;
 
 let sourceDocs = {};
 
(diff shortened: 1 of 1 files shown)


src/utils/editor/source-documents.js
```diff
@@ -4,8 +4,10 @@ import { getMode } from "../source";
 
 import type { Source } from "debugger-html";
 import { isWasm, getWasmLineNumberFormatter, renderWasmText } from "../wasm";
+import { SourceEditorUtils } from "devtools-source-editor";
 
 let sourceDocs = {};
+let wasmLineNumberFormatters: WeakMap<Object, string> = new WeakMap();
```
jasonLaster: this is a nice solution, but i think we might want it to be a bit more generic.

For instance, could it be an editorMap?

src/utils/editor/source-documents.js
```diff
@@ -24,18 +26,31 @@ function clearDocuments() {
 }
 
 function resetLineNumberFormat(editor: Object) {
+  if (!wasmLineNumberFormatters.has(editor)) {
+    return;
+  }
   let cm = editor.codeMirror;
   cm.setOption("lineNumberFormatter", number => number);
+  SourceEditorUtils.resizeBreakpointGutter(cm);
```
jasonLaster: do you think we can drop this use case if we do it here?

src/utils/editor/source-documents.js
jasonLaster: does this make sense?

src/utils/editor/source-documents.js
yurydelendik: I think we cannot remove this line since we (I) cannot guarantee editors state before updateLineNumberFormat is called.

codehag:
@yurydelendik @jasonLaster what is the status here?

jasonLaster:
@codehag - I'll work on this monday. I'm curious to see if i can avoid the weakmap

Issue #3660, opened 2017-08-14
Mochi Runner


jasonLaster:
The Mochi CLI is a mochitest wrapper designed to make running mochitests a more enjoyable pleasant experience. Two goals of the project are to 

a) smooth over rough edges introduced from the launchpad setup
b) focus on the 90% case and remove a lot of the noise that is not necessary for debugging a JS error.

#### Runner

- [ ] Fail when firefox needs to be rebuilt
- [ ] Have a prompt when the test finishes to re-run the test
- [ ] Wrap `yarn copy-assets-watch` so that users don't need two terminals
- [ ] use local config (see below)

#### Reporter

- [ ] show progress before the first test starts
- [ ] hide output for tests that pass
- [ ] use more emojis or update text inline

#### Screenshot

<img src='https://shipusercontent.com/75bdb29f1a06425bf5080b525125a725/Screen%20Shot%202017-08-14%20at%202.44.34%20PM.png' title='Screen Shot 2017-08-14 at 2.44.34 PM.png' width=1341>

##### Use local config
We are currently using a hardcoded directory for `yarn mochi`, we should instead rely on what is set in the config

```diff
diff --git a/bin/mochi.js b/bin/mochi.js
index e8c08fc3..108284ee 100644
--- a/bin/mochi.js
+++ b/bin/mochi.js
@@ -248,6 +248,7 @@ async function startWebpack() {
 }
 
 function runMochitests(args) {
+  // use the config from configs/local.json
   shell.cd("firefox");
   const command = `./mach mochitest ${args.join(" ")}`;
   console.log(chalk.blue(command));
```

jasonLaster:
[title](url)

jasonLaster:
The jest CLI is built here: https://github.com/facebook/jest/blob/ab6f4c3001a2cd917cc4df95e42dfc72f23c5a00/packages/jest-cli/src/watch.js#L318-L323

jasonLaster:
Here's a sketch for a prompt when mochi finsihes

```diff

diff --git a/bin/mochi.js b/bin/mochi.js
index f7cb119..a50c342 100644
--- a/bin/mochi.js
+++ b/bin/mochi.js
@@ -268,6 +268,21 @@ function runMochitests(args) {
       }
     });
   });
+
+  child.then(code => printCLI())
+}
+
+function printCLI() {
+  console.log(`Press r to re-run`)
+  console.log(`Press q to quit`)
+  const r = process.prompt()
[...]

PR #3768, opened 2017-08-23
Mochi exit

jasonLaster:
Associated Issue: #3660

### Summary of Changes

Mochitests are now be formatted w/ mochi. 

PROs:
* the 40K file you had to download to see you can now skim
* test errors are easier to find

CONs:
* the full run will be saved as an artifact, but not as discoverable

screen shot 2017-08-23 at 11 17 31 pm
screen shot 2017-08-23 at 11 17 19 pm

### DIFF
--- a/bin/mochi.js
+++ b/bin/mochi.js
@@ -252,10 +252,14 @@ function runMochitests(args) {
   const command = `./mach mochitest ${args.join(" ")}`;
   console.log(chalk.blue(command));
 
-  const child = shell.exec(command, {
-    async: true,
-    silent: true
-  });
+  const child = shell.exec(
+    command,
+    {
+      async: true,
+      silent: true
+    },
+    code => shell.exit(code)
+  );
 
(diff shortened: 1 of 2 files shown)


codehag:
:+1: thanks for the writeup!!

bin/mochi.js
```diff
@@ -319,8 +324,15 @@ if (process.mainModule.filename.includes("bin/mochi.js")) {
...
 
-  args = updateArgs(args);
-  run(args);
+  if (args[0] == "--read") {
+    const file = args[1];
+    const _path = path.join(__dirname, "..", file);
+    const text = fs.readFileSync(_path, { encoding: "utf8" });
+    console.log(readOutput(text).join("\n"));
+  } else {
+    args = updateArgs(args);
```
codehag: nit: would be better not to rename the args; maybe we could use updatedArgs here

jasonLaster: sure - we can do that :)

Issue #3769, opened 2017-08-24
Find Intermittents


jasonLaster:
I started a script to help us find intermittents.

There are two basic algorithms: 
1. run each test N times and report which failed
2. run partition of tests and report which failed. When there's a failure continue partitioning

<img src='https://shipusercontent.com/1ecbb89166a3c8302394df8e381c4bd9/Screen%20Shot%202017-08-23%20at%209.07.47%20PM.png' title='Screen Shot 2017-08-23 at 9.07.47 PM.png' width=755>

<img src='https://shipusercontent.com/75700747470471a3f25b29139ee41b5a/Screen%20Shot%202017-08-23%20at%209.07.14%20PM.png' title='Screen Shot 2017-08-23 at 9.07.14 PM.png' width=1432>

jasonLaster:
<img src='https://shipusercontent.com/8967081056f24707b3e67b1aaa79e6be/Screen%20Shot%202017-08-24%20at%2012.20.15%20AM.png' title='Screen Shot 2017-08-24 at 12.20.15 AM.png' width=818>

PR #3770, opened 2017-08-24
add an intermittent finding script

jasonLaster:
Associated Issue: #3769

### Summary of Changes

* fixes another intermittent
* adds a script for finding intermittents
* also adds a library for writing to a single line and another library for emojis... :dancer: 🍷 😉 

screen shot 2017-08-24 at 12 20 15 am


bin/intermitents.js
```diff
@@ -0,0 +1,112 @@
...
+  boolean: ["group"],
+  number: ["runs"],
+  string: ["path"]
+});
+
+const runs = args.runs || 10;
+const testPath = args.path ? `--testPathPattern ${args.path}` : "";
+const child = shell.exec(`jest --listTests ${testPath}`, { silent: true });
+const tests = JSON.parse(child.stdout);
+var log = require("single-line-log").stdout;
```
codehag: this could also be const?

bin/intermitents.js
```diff
@@ -0,0 +1,112 @@
...
+    } else {
+      progress.push(".");
+    }
+    const status = failed ? emoji.get("rage") : emoji.get("sweat_smile");
+    const dots = progress.join("");
+    log(`${status}  ${file} ${dots}`);
+  });
+
+  const dots = progress.join("");
+  const status = failed ? emoji.get("rage") : emoji.get("sweat_smile");
```
codehag: if i am reading this correctly.. this will be set if the last run of the test failed. maybe we want to change it so that the forEach loop has its own failed status, and this uses _.any(test => test.failed) or something?

jasonLaster: it does it there too. `_.any` is nicer.

codehag: i just realized that we maintain the failed state through the loop.. maybe we could use a for loop and exit early if it fails?

jasonLaster: i did that earlier but i like having the data of how often it fails... we can always clean it up later

PR #3774, opened 2017-08-24
Add React for CallSites JSX

darkwing:
By the book, we're supposed to add `React` to use JSX, so adding it in.


codehag:
looks good

PR #3775, opened 2017-08-24
Fix 'copySource.accesskey' error when right-clicking

darkwing:
Right-clicking in the source area is currently triggering an error.  Looks like we accidentally mis-named a localization key.


codehag:
Thanks!

PR #3791, opened 2017-08-27
WIP: Convert SVG icon usage to JSX

darkwing:
It's nicer to use this class as JSX.  I still have some cleanup to do but must break for now.

- [ ] Determine title/aria-title overlap
- [ ] Fix failing tests


jasonLaster:
pushed the jest snapshots and merging.

Issue #3795, opened 2017-08-28
Recurring issues with properties file updates


juliandescottes:
The PR #3713 landed with a wrong localization note in the properties file. 

`# LOCALIZATION NOTE (copy)`

`copy` should match the name of the localization string. 

` # LOCALIZATION NOTE (copySource)`

Issues with the properties file are a recurring friction point we have when dropping a new bundle of the debugger in mozilla-central. Sorry to nag you about that, but catching this kind of things when trying to update the bundle is a bit too late, not to mention frustrating :)

The debugger properties file is used by the localization teams, and therefore need to follow their rules.

You can find some resources about localization guidelines here: https://developer.mozilla.org/en/docs/Mozilla/Localization but honestly it's hard to remember all the rules when you are not doing this regularly, so maybe ping @flodolo (or me but I might miss a lot of stuff) when a PR touches the properties file.

cc @codehag @jasonLaster : any idea on how to enforce a better review policy for the debugger properties file?

flodolo:
Some projects are using the CODEOWNER file to ping me automatically
https://github.com/mozilla/testpilot/blob/master/CODEOWNERS

Downsides:
* Right now, it's just me
* It might get a bit confusing, since I'm reviewing string changes, not looking at code, and the PR might look reviewed 
* I'm afraid you need to add me as contributor to the repo (but I'm not completely sure about this).

flodolo:
This might be a better doc to look at
https://developer.mozilla.org/en-US/docs/Mozilla/Localization/Localization_content_best_practices#Choose_good_key_IDs

Also note something that debugger is not doing: it would be great to have keys called "FOO.label and FOO.accesskey", vs "FOO and FOO.accesskey".

juliandescottes:
Quickly discussed about this with @jasonLaster and we'll add @flodolo to CODEOWNERS for the properties file.

jasonLaster:
@flodolo I just added you as a collaborator. I think you'll have to accept before you're included

flodolo:
> @flodolo I just added you as a collaborator. I think you'll have to accept before you're included

Done, thanks.

PR #3797, opened 2017-08-28
Issue #3795 - create CODEOWNERS file, assign @flodolo to debugger.pro…

juliandescottes:
Associated Issue: #3795

This adds a CODEOWNERS file for the project, currently only containing a rule for @flodolo and assets/panel/debugger.properties

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,2 @@
+# flod as main contact for string changes - TODO: find more to spread the effort
+assets/panel/debugger.properties @flodolo


wldcordeiro:
Since you can assign a team why not assign the `@devtools-html/debugger` team as the owners of `src/`?

juliandescottes:
@wldcordeiro I have nothing against that but the goal of my PR is only to improve the review process for the properties file. I'm not involved with the debugger's development enough to know which owners should be set for other folders.

wldcordeiro:
@juliandescottes I see. I had just read the documentation on CODEOWNERS and it seemed like it was something where you could do this and it'd win out for the debugger.properties while saying that anyone within that team is an owner elsewhere.

```
@devtools-html/debugger
assets/panel/debugger.properties @flodolo
```

jasonLaster:
@wldcordeiro how would src/ help? I assume everyone is already reviewing :)

jasonLaster:
Thanks @juliandescottes

wldcordeiro:
@jasonLaster True. I thought it'd be nice if it suggested the whole team as possible reviewers but the way it suggests reviewers based on who recently touched the same code is very nice.

Issue #3682, opened 2017-08-16
[Breakpoints] Leftover breakpoint breaks the debugger


eoger:
I set a breakpoint in a previous session in the browser toolbox debugger.
When I open the debugger (even after restarting the browser multiple times), clicks on the Firefox UI are not registered and CSS `:hover` rules are not applied.
Trying to remove that breakpoint triggers an error.

Screencast:
https://gfycat.com/gifs/detail/JubilantDeepAyeaye

Entire screenshot of the errors in the screencast:
<img width="1003" alt="screen shot 2017-08-16 at 10 45 00 am" src="https://user-images.githubusercontent.com/6424575/29369234-fed3a0b6-826f-11e7-8918-8a167bd06a5d.png">

cc @vladikoff

jasonLaster:
@eoger mind sharing the STR?

vladikoff:
@eoger Was this in Nightly?

eoger:
Yes this was Nightly

eoger:
@jasonLaster wasn't able to reproduce exactly. But:
- Open Firefox
- Open browser toolbox, set a breakpoint in browser-sync, line 260.
- Cmd+Q, reopen browser, reopen debugger

From then the breakpoint has disappeared, but I can't set it up again:
https://gfycat.com/gifs/detail/LavishEnviousAdder
<img width="949" alt="screen shot 2017-08-16 at 11 57 20 am" src="https://user-images.githubusercontent.com/6424575/29372914-83c5fce2-827a-11e7-92c5-2ccb37fb9aa1.png">

vladikoff:
Ok I was able to reproduce this locally, the breakpoint won't stick to the line number:

![action-s](https://user-images.githubusercontent.com/128755/29379775-80cb4850-8291-11e7-93f7-a05a0724849d.gif)

vladikoff:
### STR

* Fresh Nightly Profile (57.0a1 (2017-08-16) (64-bit))
* Open toolbox
* Set a breakpoint at browser-sync.js by going to Debugger, looking for file called `browser-sync.js`:

![image](https://user-images.githubusercontent.com/128755/29379875-d26baa60-8291-11e7-8781-263b872b1e7b.png)

* CMD+Q DevTools, CMD+Q Firefox
* Restart the created profile
* Open Toolbox, browser-sync.js is open
* Scroll to line, no breakpoint set:

[...]

PR #3799, opened 2017-08-28
[wip] disable original bp corrections

jasonLaster:
Associated Issue: #3682

### Summary of Changes

* this is a short term fix, which will be easier to uplift. Instead of trying to keep the bp in the same location, we let it move... Should be easier to uplift

### DIFF
--- a/src/actions/breakpoints/syncBreakpoint.js
+++ b/src/actions/breakpoints/syncBreakpoint.js
@@ -59,16 +59,16 @@ export async function syncClientBreakpoint(
   const existingClient = client.getBreakpointByLocation(generatedLocation);
 
   if (existingClient) {
-    const newGeneratedLocation = existingClient.actualLocation;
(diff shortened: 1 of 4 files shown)


src/actions/breakpoints/syncBreakpoint.js
```diff
@@ -59,16 +59,16 @@ export async function syncClientBreakpoint(
   const existingClient = client.getBreakpointByLocation(generatedLocation);
 
   if (existingClient) {
-    const newGeneratedLocation = existingClient.actualLocation;
-    const newLocation = await sourceMaps.getOriginalLocation(
-      newGeneratedLocation
-    );
+    // const newGeneratedLocation = existingClient.actualLocation;
```
codehag: im not sure this will work as intended, the server will still have a breakpoint that moved, but now we will not represent it on the client

PR #3804, opened 2017-08-28
test modification to properties file

juliandescottes:
test for codeowners again.


juliandescottes:
CODEOWNERS seems to work as expected!

PR #3525, opened 2017-08-01
Fix #3483 - Highlight only executable lines

darkwing:
Associated Issue: #3483 

This is a *WIP*.  Resource tested was `demo/EmptyLineTest.js` from https://davidwalsh.name/demo/EmptyLineTest.php

ToDo:

- [x] Remove the App.css change; not sure how that got in here
- [x] Code the `componentWillUnmount`part
- [x] The `getEmptyLines` function could probably be prettier; seems like I've used lots of negative (`!`) logic which may be confusing for maintenance
- [ ] TESTS
- [x] Clean up `console` usage
- [x] Prevent breakpoints from being created when an empty line is clicked (do something with the context menu also?)
- [x] Ensure this functions on "source-formatted" tabs; for some reason parsing fails when I pretty-print the `EmptyLineTest.js` file; WTF.  *Update* The parse changes `let d;` to `letd;`

Issue #3483, opened 2017-07-27
[Expressions] Remove visible field


jasonLaster:
We no longer have the notion of visible/hidden expressions so we should drop the field from the redux side.

PR #3818, opened 2017-08-30
Add Empty Lines to AST reducer

jasonLaster:
#3483

Cherry picks some more work from #3525

### DIFF
--- a/src/actions/ast.js
+++ b/src/actions/ast.js
@@ -38,6 +38,28 @@ export function setSymbols(sourceId: SourceId) {
   };
 }
 
+export function setEmptyLines(sourceId: SourceId) {
+  return async ({ dispatch, getState }: ThunkArgs) => {
+    const sourceRecord = getSource(getState(), sourceId);
+    if (!sourceRecord) {
+      return;
+    }
+
+    const source = sourceRecord.toJS();
+    if (!source.text) {
+      return;
+    }
+
+    const emptyLines = await parser.getEmptyLines(source);
+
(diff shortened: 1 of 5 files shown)


src/actions/ast.js
```diff
@@ -38,6 +38,28 @@ export function setSymbols(sourceId: SourceId) {
...
+
+    const emptyLines = await parser.getEmptyLines(source);
+
+    dispatch({
+      type: "SET_EMPTY_LINES",
+      source,
+      emptyLines
+    });
+  };
+}
```
jasonLaster: we should add a unit test here

src/reducers/ast.js
```diff
@@ -124,6 +132,23 @@ export function hasSymbols(state: OuterState, source: Source): boolean {
...
+  state: OuterState,
+  line: number,
+  selectedSource: Source
+) {
+  const emptyLines = getEmptyLines(state, selectedSource);
+  return emptyLines.includes(line);
+}
+
+export function getEmptyLines(state: OuterState, source: Source) {
+  if (!source) {
```
darkwing: Looking at this again, we should probably return `[]` instead of `false`.

darkwing: (Prevents an error in `isEmptyLineInSource` because we immediately use `.includes`; I think it's reasonable to always expect an array from this function)

jasonLaster: good call

jasonLaster:
added a small test

PR #3821, opened 2017-08-30
Display empty lines UI

darkwing:
As @jasonLaster merged the redux piece, this PR is for the UI piece of displaying empty vs. executable lines

### DIFF
--- a/src/actions/sources.js
+++ b/src/actions/sources.js
@@ -334,6 +334,7 @@ export function togglePrettyPrint(sourceId: string) {
     );
 
     await dispatch(remapBreakpoints(sourceId));
+    await dispatch(setEmptyLines(newPrettySource.id));
 
     return dispatch(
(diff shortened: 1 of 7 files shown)


darkwing:
As with the previous PR, the functionality works with a normal source tab but pretty-printed source tabs throw an error.

jasonLaster:
Seeing one flow error and two mochitest errors, which makes sense because we test sliding in our mochitests.
We can no longer slide :P

darkwing:
Working on them right now

jasonLaster:
oh nice. Forgot that you mochitest :)

Try `yarn mochi breakpoints-reloading` in debugger land

also we have mochitest docs for

```
yarn copy-assets
```

and getting firefox locally `./bin/prepare-mochitest`

jasonLaster:
🌺

PR #3827, opened 2017-08-30
Remove devtools.debugger.client-source-maps-enabled pref

tromey:
This removes the debugger.debugger.client-source-maps-enabled pref.
There was no UI for disabling this pref, and it is only checked in a
single spot.  In M-C I am going to unify all the source-map prefs, and I
don't think the debugger needs to react to changes to the shared pref;
so removing this one seemed best.


tromey:
Actually it is in the options panel and I am confused.
I'm going to close this PR while I investigate.

PR #3832, opened 2017-08-30
[WIP] Feature/ast breakpoints p2

codehag:
part 2 of ast breakpoints

### DIFF
--- a/src/actions/ast.js
+++ b/src/actions/ast.js
@@ -24,11 +24,11 @@ export function setSymbols(sourceId: SourceId) {
     }
 
     const source = sourceRecord.toJS();
-    if (!source.text || hasSymbols(getState(), source)) {
+    if (!source.text) {
       return;
     }
 
-    const symbols = await parser.getSymbols(source);
+    const symbols = await parser.getSymbols(source, true);
 
     dispatch({
       type: "SET_SYMBOLS",
--- a/src/actions/breakpoints.js
+++ b/src/actions/breakpoints.js
@@ -45,11 +45,6 @@ export function syncBreakpoint(
   pendingBreakpoint: PendingBreakpoint
 ) {
   return ({ dispatch, getState, client, sourceMaps }: ThunkArgs) => {
-    const sourceId = source.id;
-    const { line, sourceUrl, column } = pendingBreakpoint.location;
-    const location = { sourceId, sourceUrl, line, column };
-    const breakpoint = createBreakpoint(location, pendingBreakpoint);
-
     const syncPromise = syncClientBreakpoint(
       getState,
       client,
@@ -60,7 +55,6 @@ export function syncBreakpoint(
 
     return dispatch({
       type: "SYNC_BREAKPOINT",
-      breakpoint,
       [PROMISE]: syncPromise
     });
   };
--- a/src/actions/breakpoints/addBreakpoint.js
+++ b/src/actions/breakpoints/addBreakpoint.js
@@ -2,6 +2,7 @@ import {
   locationMoved,
   breakpointExists,
   assertBreakpoint,
+  getASTLocation,
   assertLocation
 } from "../../utils/breakpoint";
 import { getSource } from "../../selectors";
@@ -16,10 +17,11 @@ export default async function addBreakpoint(
   const state = getState();
 
   const source = getSource(state, breakpoint.location.sourceId);
+  const _source = source.toJS();
   const location = { ...breakpoint.location, sourceUrl: source.get("url") };
   const generatedLocation = await getGeneratedLocation(
     state,
-    source.toJS(),
+    _source,
     location,
     sourceMaps
   );
@@ -44,6 +46,8 @@ export default async function addBreakpoint(
     newGeneratedLocation
   );
 
+  const astLocation = await getASTLocation(_source, location);
+
   const newBreakpoint = {
     id,
     disabled: false,
@@ -51,6 +55,7 @@ export default async function addBreakpoint(
     loading: false,
     condition: breakpoint.condition,
     location: newLocation,
+    astLocation,
     hitCount,
     generatedLocation: newGeneratedLocation
   };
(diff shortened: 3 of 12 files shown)


src/utils/parser/getSymbols.js
```diff
@@ -177,8 +177,8 @@ function extractSymbols(source: Source) {
   };
 }
 
-export default function getSymbols(source: Source): SymbolDeclarations {
-  if (symbolDeclarations.has(source.id)) {
+export default function getSymbols(source: Source, update): SymbolDeclarations {
```
jasonLaster: > nit, perhaps we make the signature clearer

`getSymbols(source: Source, { shouldUpdate: false })`

src/utils/breakpoint/astBreakpointLocation.js
```diff
@@ -32,13 +35,17 @@ function findClosestScope(functions: Scope[], location: Location) {
 
 export async function getASTLocation(source: Source, location: Location) {
```
jasonLaster: side comment, but this reminds me of the code in `ast#setPreview` it'd be nice to move it out too

src/utils/prefs.js
```diff
@@ -3,7 +3,7 @@
 const { isDevelopment } = require("devtools-config");
 const { Services, PrefsHelper } = require("devtools-modules");
 
-const prefsSchemaVersion = "1.0.2";
+const prefsSchemaVersion = "1.0.3";
```
jasonLaster: lol

src/actions/breakpoints.js
```diff
@@ -60,7 +55,6 @@ export function syncBreakpoint(
 
     return dispatch({
       type: "SYNC_BREAKPOINT",
-      breakpoint,
       [PROMISE]: syncPromise
```
jasonLaster: I'm glad we waited on this abstraction, i'm beginning to think we just want to move complicated or all actions to their own files

perhaps we should inline the call

```js
dispatch({
  type: "SYNC_BREAKPOINT",
  [PROMISE]: syncClientBreakpoint() 
})
```

or and this is a radical idea... 

move the whole thing to the syncBreakpoint file

```js
export function syncBreakpoint(..) {
    dispatch( 
      type: "SYNC_BREAKPOINT",
      [PROMISE]: syncClientBreakpoint() 
    )
  })
}
```

The advantage of this approach is that it's intellectually honest. We're just saying, go over there to see this action.


anyways, one big nit

src/actions/breakpoints/addBreakpoint.js
```diff
@@ -16,10 +17,11 @@ export default async function addBreakpoint(
   const state = getState();
 
   const source = getSource(state, breakpoint.location.sourceId);
```
jasonLaster: maybe `sourceRecord` then `source` below

src/actions/breakpoints/syncBreakpoint.js
```diff
@@ -40,12 +59,14 @@ export async function syncClientBreakpoint(
   // early return if breakpoint is disabled, send overrides to update
   // the id as expected
   if (pendingBreakpoint.disabled) {
-    const newLocation = await sourceMaps.getOriginalLocation(generatedLocation);
+    const newLocation = await sourceMaps.getOriginalLocation(
+      scopeGeneratedLocation
```
jasonLaster: another nit:

maybe we want some smaller functions here, like each case is it's own function. the top vars are defined where they're used

src/actions/breakpoints/syncBreakpoint.js
```diff
@@ -63,38 +84,19 @@ export async function syncClientBreakpoint(
...
-    const breakpoint = {
-      ...pendingBreakpoint,
-      id: makeLocationId(location),
-      generatedLocation: newGeneratedLocation,
-      location: location
-    };
-
-    assertBreakpoint(breakpoint);
-    return { breakpoint, previousLocation: location };
+    await client.removeBreakpoint(generatedLocation);
```
jasonLaster: don't we need do `client.setBreakpoint` to create a new one?

src/actions/breakpoints/syncBreakpoint.js
```diff
@@ -22,6 +24,23 @@ export async function syncClientBreakpoint(
...
+    sourceUrl: source.url,
+    sourceId
+  };
+
+  const scopeGeneratedLocation = await getGeneratedLocation(
+    getState(),
+    source,
+    scopeLocation,
+    sourceMaps
+  );
```
jasonLaster: this might be a bit more complicated, but should we also do the `newGeneratedLocation` so if we don't have an `astLocation` we fall back on it? Or wait, will you tell me that astLocation is doing the same thing?...

src/actions/breakpoints/syncBreakpoint.js
```diff
@@ -22,6 +24,23 @@ export async function syncClientBreakpoint(
...
+
```
jasonLaster: yeah - lets add some unit tests :)

src/actions/ast.js
```diff
@@ -24,11 +24,11 @@ export function setSymbols(sourceId: SourceId) {
     }
 
     const source = sourceRecord.toJS();
-    if (!source.text || hasSymbols(getState(), source)) {
+    if (!source.text) {
       return;
     }
 
-    const symbols = await parser.getSymbols(source);
+    const symbols = await parser.getSymbols(source, true);
```
jasonLaster: if we're always updating here do we need the boolean?

src/actions/breakpoints/syncBreakpoint.js
codehag: :P

src/actions/breakpoints/syncBreakpoint.js
codehag: its called a little bit lower (the two cases are now more like one case)

src/actions/ast.js
codehag: we have other cases where we use getSymbols

codehag:
closing because this has been split into smaller prs

Issue #3652, opened 2017-08-12
[ProjectSearch] focus on the first match


jasonLaster:
STR:
1. cmd+shift+f 
2. type a query
3. hit enter

you expect to be focused on the first result so you can use up and down arrow keys, but you're not :)

#### Code Sketch

```diff
diff --git a/src/components/ProjectSearch/TextSearch.js b/src/components/ProjectSearch/TextSearch.js
index 2b75140..493839d 100644
--- a/src/components/ProjectSearch/TextSearch.js
+++ b/src/components/ProjectSearch/TextSearch.js
@@ -59,6 +59,14 @@ export default class TextSearch extends Component {
     shortcuts.on("Enter", this.onEnterPress);
   }

+  componentDidUpdate() {
+    // i'm not sure how to know if its new results
+    if (resultsChanged) {
+      // im not sure how to select the first item in managed tree...
+      this.selectFirstMatch();
+    }
+  }
+
```

#### Screenshot

<img src='https://shipusercontent.com/d2f19992f53b41a9d69d15246cadde02/Screen%20Shot%202017-08-12%20at%203.21.40%20PM.png' title='Screen Shot 2017-08-12 at 3.21.40 PM.png' width=712>

bomsy:
```diff
diff --git a/src/components/ProjectSearch/TextSearch.js b/src/components/ProjectSearch/TextSearch.js
index 2b75140e..46e775fb 100644
--- a/src/components/ProjectSearch/TextSearch.js
+++ b/src/components/ProjectSearch/TextSearch.js
@@ -184,8 +184,8 @@ export default class TextSearch extends Component {
         getChildren={file => file.matches || []}
         itemHeight={20}
         autoExpand={1}
+        highlightItems={[results[0]]}
         autoExpandDepth={1}
-        focused={results[0]}
         getParent={item => null}
         getPath={getFilePath}
         renderItem={renderItem}
diff --git a/src/components/shared/ManagedTree.js b/src/components/shared/ManagedTree.js
index 2c26e95f..d40d0e66 100644
--- a/src/components/shared/ManagedTree.js
+++ b/src/components/shared/ManagedTree.js
@@ -94,14 +94,17 @@ class ManagedTree extends Component {
     const expanded = this.state.expanded;

     // This file is visible, so we highlight it.
-    if (expanded.has(this.props.getPath(highlightItems[0]))) {
+    if (
+      highlightItems[0] &&
[...]

PR #3841, opened 2017-08-31
Fix focus issues

bomsy:
Associated Issue: #3652

### Summary of Changes

* Adds a focused property to Managedtree
* Focuses on the the first match in first initial result
* Allows you start moving through the matches with more content is coming in

### Test Plan

Example test plan:

- [x] Cmd + Ctrl + F
- [x] Enter search query
- [x] Move through the list
- [x] Try to break it!!!

### DIFF
--- a/src/components/ProjectSearch/TextSearch.js
+++ b/src/components/ProjectSearch/TextSearch.js
@@ -11,14 +11,41 @@ import "./TextSearch.css";
 
 import { getRelativePath } from "../../utils/sources-tree";
 
+// This finds a node that matches (has the same values
+// for specific properties) as the node passed in.
+
+function getMatchingFocusNode(results, node) {
+  if (node === null) {
+    return null;
+  }
+  return results.find(result => {
+    if (node.filepath) {
+      if (node.sourceId === result.sourceId) {
+        return result;
+      }
+    } else {
+      return result.matches.some(match => {
+        if (
+          node.sourceId === match.sourceId &&
+          node.column === match.column &&
+          node.line === match.line
+        ) {
+          return match;
+        }
+      });
+    }
+  });
+}
+
 export default class TextSearch extends Component {
   constructor(props: Props) {
     super(props);
     this.state = {
-      inputValue: this.props.query || ""
+      inputValue: this.props.query || "",
+      inputFocused: false
     };
 
-    this.focused = null;
+    this.focusedItem = null;
 
     this.inputOnChange = this.inputOnChange.bind(this);
     this.onKeyDown = this.onKeyDown.bind(this);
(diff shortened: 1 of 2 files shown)


src/components/ProjectSearch/TextSearch.js
```diff
@@ -11,14 +11,45 @@ import "./TextSearch.css";
...
 
+// This finds a node that matches (has the same values
+// for specific properties) as the node passed in.
+
+function getMatchingFocusNode(results, node) {
+  if (node === null) {
+    return null;
+  }
+  let matchingNode = null;
+  results.some(result => {
```
codehag: do we expect only one matching node in the collection? perhaps we can use `find` here instead and do
```
const matchingNode = results.find(....) // same code as before, but return `result` rather than assigning matchingNode
```

src/components/ProjectSearch/TextSearch.js
```diff
@@ -49,6 +82,19 @@ export default class TextSearch extends Component {
...
 
+  componentWillReceiveProps(nextProps) {
+    if (nextProps.results !== this.props.results) {
+      if (this.focusedItem === null) {
+        return;
+      }
+      const { file, match } = this.focusedItem;
+
+      this.focusedItem[
+        file && !match ? "file" : "match"
```
codehag: in the case of `!file && !match` we would have the string `match` -- is this what is expected? We might want to move this ternary out into a variable

src/components/ProjectSearch/TextSearch.js
```diff
@@ -11,14 +11,45 @@ import "./TextSearch.css";
...
+  }
+  let matchingNode = null;
+  results.some(result => {
+    if (node.filepath) {
+      if (node.sourceId === result.sourceId) {
+        matchingNode = result;
+        return true;
+      }
+    } else {
+      result.matches.some(match => {
```
codehag: this can also be a find function

src/components/ProjectSearch/TextSearch.js
```diff
@@ -31,16 +62,18 @@ export default class TextSearch extends Component {
...
   }
 
-  async onKeyDown(e) {
+  onKeyDown(e) {
     if (e.key !== "Enter") {
       return;
     }
+    this.focusedItem = null;
     this.props.searchSources(this.state.inputValue);
+    e.stopPropagation();
```
codehag: nit: i would expect this at the top of the function

src/components/ProjectSearch/TextSearch.js
bomsy: it looks like `.find` would not work here because there is an `else` part searches the `matches` inside the `result` rather than returning the `result`

wldcordeiro: @bomsy I think @codehag is right here. You can use `find` and it would look like so

```js
// Or just return this call.
const matchingNode = results.find(result => {
  if (node.filepath) {
    if (node.sourceId === result.sourceId) {
        return result;
    }
  } else {
    return result.matches.find(match => {
      if (
        node.sourceId === match.sourceId &&
        node.column === match.column &&
        node.line === match.line
      ) {
        return match;
      }
    });
  }
})

File: src/components/ProjectSearch/TextSearch.js [hunk 5]
```diff
@@ -49,6 +82,19 @@ export default class TextSearch extends Component {
     }
   }
 
+  componentWillReceiveProps(nextProps) {
```
wldcordeiro: Nit: `componentWillReceiveProps` should go before the handlers.

src/components/ProjectSearch/TextSearch.js
bomsy: yeah .. i tried it and it doesn't seem to work

src/components/ProjectSearch/TextSearch.js
```diff
@@ -49,6 +82,19 @@ export default class TextSearch extends Component {
...
+      }
+      const { file, match } = this.focusedItem;
+
+      this.focusedItem[
+        file && !match ? "file" : "match"
+      ] = getMatchingFocusNode(nextProps.results, file || match);
+    }
+  }
+
   componentWillUnmount() {
```
wldcordeiro: Nit: Same thing, `componentWillUnmount` is life cycle and shoud be above handers & renderers.

src/components/ProjectSearch/TextSearch.js
bomsy:
```js
return result.matches.find(match => { ...
```
this line will return undefined

wldcordeiro:
Awesome job @bomsy a few minor tweaks but excellent to see this coming in. 😄

PR #3845, opened 2017-09-01
Make the source tab hover effects feel more responsive

darkwing:
This is just a rough proposal.  My opinion is that the tab hover transitions are slow and appear labored.  The labored feeling is due to speed, so I drastically sped up the transitions and they seem much more quick and elegant.

I know the project has loads of transitions so I guess I'd like to hear some opinions on this faster effect and see if we'd like to speed things up globally.

### DIFF
--- a/src/components/Editor/Tabs.css
+++ b/src/components/Editor/Tabs.css
@@ -41,7 +41,7 @@
   display: inline-flex;
   align-items: flex-end;
   position: relative;
-  transition: all 0.25s ease;
+  transition: all 0.15s ease;
   min-width: 40px;
   overflow: hidden;
   padding: 5px;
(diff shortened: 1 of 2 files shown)


codehag:
hi @darkwing! thanks for the proposal, do you happen to have a screencap of the change?

codehag:
hm...now that i think about it, i guess showing transitions with gifs doesnt work all that well.. 🤔 ill pull and take a look in a little bit!

wldcordeiro:
@darkwing When I was implementing the `Modal` transition, I was pointed to this document http://design.firefox.com/photon/motion/motion.html Using the curve and `150ms` would be a good transition time I think.

darkwing:
Updated to .15s.  Still think the lower speed seems nicer but branding is important.

PR #3848, opened 2017-09-01
add astLocation to addBreakpoint

codehag:
Associated Issue: #3150

### DIFF
--- a/package.json
+++ b/package.json
@@ -82,18 +82,9 @@
     "svg-inline-react": "^1.0.2",
     "wasmparser": "^0.4.10"
   },
-  "files": [
-    "src",
-    "assets"
-  ],
+  "files": ["src", "assets"],
   "greenkeeper": {
-    "ignore": [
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -149,47 +140,19 @@
     "yarn": "0.28.1"
   },
   "lint-staged": {
-    "*.js": [
```
jasonLaster: prettier!

src/actions/breakpoints/addBreakpoint.js
```diff
@@ -16,10 +17,11 @@ export default async function addBreakpoint(
   const state = getState();
 
   const source = getSource(state, breakpoint.location.sourceId);
+  const _source = source.toJS();
```
jasonLaster: i think `sourceRecord` and `source` would be clearer, but it's a new precident so we can also just not use `_source` and do `source.toJS()` everywhere too...

PR #3849, opened 2017-09-01
clear ast tree on navigate

codehag:
We were not clearing the AST on navigate. this resulted in a stale ast tree when a source was reloaded and changed.


codehag:
Im looking into the failing test here

codehag:
Im not sure what is going on, this isn't failing for me locally...

PR #3817, opened 2017-08-29
Show mapped names in scopes bindings

yurydelendik:
Associated Issue: #3427

### Summary of Changes

* adds the `generatedLocation` to a frame
* adds building of scopes database for the generated source
* updates/adds original mapped names to the scopes

### Test Plan

Tell us a little a bit about how you tested your patch.

Example test plan:

- [x] Open JavaScript code with source maps that have mangled names (for simplicity sequence_print.min.js in http://firefox-dev.tools/debugger-examples/examples/sequence-print/sequence_print.html)
- [x] In original source, sequence_print.js, at line 4 (`buffer.push(current)`) set breakpoint
- [x] Reload sequence_print.html app
- [x] Observe scopes

The top scope needs to have `current`, `buffer` variables shown.

<del>P.S. needs some modification to the devtools-source-map</del>

PR #3852, opened 2017-09-01
Parse to extract all scopes and bindings.

yurydelendik:
Associated Issue: #3817 

### Summary of Changes

* allows to send source text independent from parsing of the scopes
* parsing scopes to extract positions and its scope of every symbol in the source code

### DIFF
--- a/src/utils/parser/getScopes.js
+++ b/src/utils/parser/getScopes.js
@@ -1,0 +1,28 @@
+// @flow
+
+import { traverseAst } from "./utils/ast";
+import { getSource } from "./sources";
+import {
+  createParseJSScopeVisitor,
+  findScopes
+} from "devtools-map-bindings/src/parser";
+
+import type { Location, SourceScope } from "debugger-html";
+
+let parsedScopesCache = new Map();
+
+export default function getScopes(location: Location): SourceScope[] {
+  const { sourceId } = location;
+  let parsedScopes = parsedScopesCache.get(sourceId);
+  if (!parsedScopes) {
+    const visitor = createParseJSScopeVisitor(sourceId);
+    traverseAst(getSource(sourceId), visitor.traverseVisitor);
+    parsedScopes = visitor.toParsedScopes();
+    parsedScopesCache.set(sourceId, parsedScopes);
+  }
+  return findScopes(parsedScopes, location);
+}
+
+export function clearScopes() {
+  parsedScopesCache = new Map();
+}
--- a/flow-typed/debugger-html.js
+++ b/flow-typed/debugger-html.js
@@ -222,6 +222,33 @@ declare module "debugger-html" {
   };
 
   /**
+   * SourceScope
+   *
+   * @memberof types
+   * @static
+   */
+  declare type SourceScope = {
+    type: string,
+    start: Location,
+    end: Location,
+    bindings: {
+      [name: string]: Location[]
+    }
+  };
+
+  /*
+   * MappedScopeBindings
+   * @memberof types
+   * @static
+   */
+  declare type MappedScopeBindings = {
+    type: string,
+    bindings: {
+      [originalName: string]: string
+    }
+  };
+
+  /**
  * Script
  * This describes scripts which are sent to the debug server to be eval'd
  * @memberof types
--- a/package.json
+++ b/package.json
@@ -70,7 +70,7 @@
     "devtools-launchpad": "^0.0.98",
     "devtools-reps": "^0.12.3",
     "devtools-source-editor": "0.0.6",
-    "devtools-source-map": "0.12.0",
+    "devtools-source-map": "0.13.0",
     "devtools-splitter": "^0.0.3",
     "devtools-utils": "^0.0.9",
     "fuzzaldrin-plus": "^0.4.1",
--- a/src/actions/navigation.js
+++ b/src/actions/navigation.js
@@ -2,7 +2,12 @@ import { clearDocuments } from "../utils/editor";
 import { getSources } from "../reducers/sources";
 import { waitForMs } from "../utils/utils";
 import { newSources } from "./sources";
-import { clearSymbols, clearASTs, clearSources } from "../utils/parser";
+import {
+  clearASTs,
+  clearSymbols,
+  clearScopes,
+  clearSources
+} from "../utils/parser";
 import { clearWasmStates } from "../utils/wasm";
 
 /**
@@ -21,6 +26,7 @@ export function willNavigate(_, event) {
     clearDocuments();
     clearSymbols();
     clearASTs();
+    clearScopes();
     clearSources();
 
     dispatch(navigate(event.url));
--- a/src/actions/sources.js
+++ b/src/actions/sources.js
@@ -24,6 +24,7 @@ import { loadSourceText } from "./sources/loadSourceText";
 
 import { prefs } from "../utils/prefs";
 import { removeDocument } from "../utils/editor";
+import * as parser from "../utils/parser";
 
 import {
   getSource,
@@ -377,6 +378,21 @@ export function loadAllSources() {
       if (query) {
         await dispatch(searchSource(source, query));
       }
+    }
+  };
+}
+
+/**
+ * Ensures parser has source text
+ *
+ * @memberof actions/sources
+ * @static
+ */
+export function ensureParserHasSourceText(sourceId: string) {
+  return async ({ dispatch, getState }: ThunkArgs) => {
+    if (!await parser.hasSource(sourceId)) {
+      await dispatch(loadSourceText(getSource(getState(), sourceId).toJS()));
+      await parser.setSource(getSource(getState(), sourceId).toJS());
     }
   };
 }
--- a/src/actions/tests/__snapshots__/breakpoints.spec.js.snap
+++ b/src/actions/tests/__snapshots__/breakpoints.spec.js.snap
@@ -23,7 +23,7 @@ Object {
   "loading": false,
   "location": Object {
     "column": 0,
-    "line": 7,
+    "line": 1,
     "sourceId": "a.js/originalSource-d6d70368d5c252598541e693a7ad6c27",
     "sourceUrl": "http://localhost:8000/examples/a.js:formatted",
   },
--- a/src/utils/parser/index.js
+++ b/src/utils/parser/index.js
@@ -9,9 +9,11 @@ export const stopParserWorker = dispatcher.stop.bind(dispatcher);
 
 export const getClosestExpression = dispatcher.task("getClosestExpression");
 export const getSymbols = dispatcher.task("getSymbols");
+export const getScopes = dispatcher.task("getScopes");
 export const getVariablesInScope = dispatcher.task("getVariablesInScope");
 export const getOutOfScopeLocations = dispatcher.task("getOutOfScopeLocations");
 export const clearSymbols = dispatcher.task("clearSymbols");
+export const clearScopes = dispatcher.task("clearScopes");
 export const clearASTs = dispatcher.task("clearASTs");
 export const getNextStep = dispatcher.task("getNextStep");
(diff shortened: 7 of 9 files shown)


src/utils/parser/getScopes.js
```diff
@@ -0,0 +1,280 @@
...
+        parent = savedParent;
+        savedParents.delete(path);
+      }
+    }
+  });
+  return _toGeneratedScopes(parent.scopes, source.id);
+}
+
+function _compareLocations(a: Location, b: Location): number {
+  // $FlowIgnore
```
codehag: why are we ignoring here?

src/utils/parser/getScopes.js
```diff
@@ -0,0 +1,280 @@
...
+  const { sourceId } = location;
+  let parsedScopes = parsedScopesCache.get(sourceId);
+  if (!parsedScopes) {
+    parsedScopes = _parseJSScopes(getSource(sourceId));
+    parsedScopesCache.set(sourceId, parsedScopes);
+  }
+
+  // Find inner most
+  let search = parsedScopes;
+  let found = [];
```
codehag: this should be const

src/utils/parser/getScopes.js
```diff
@@ -0,0 +1,280 @@
...
+        let scope = parent;
+        if (hasLet) {
+          savedParents.set(path, parent);
+          scope = _createTempScope("Block", parent, location);
+          parent = scope;
+        }
+        return;
+      }
+      if (path.isVariableDeclaration()) {
+        let hoistAt = parent;
```
codehag: this doesnt need to be assigned to parent, would make sense to assign this to null, or turn the if statement into a function that returns the hoistAt value or null

src/utils/parser/getScopes.js
```diff
@@ -0,0 +1,280 @@
...
+      }
+      if (path.isForInStatement() || path.isForStatement()) {
+        const init = tree.init || tree.left;
+        let scope = parent;
+        if (
+          isNode(init, "VariableDeclaration") &&
+          (init.kind === "let" || init.kind === "const")
+        ) {
+          savedParents.set(path, parent);
+          scope = _createTempScope("For", parent, location);
```
codehag: scope can be declared here (as it is in other places)
`const scope = _createTempScope ...`

src/utils/parser/getScopes.js
```diff
@@ -0,0 +1,280 @@
...
+        }
+        tree.params.forEach(param => {
+          _parseDeclarator(param, scope, "param");
+        });
+        parent = scope;
+        return;
+      }
+      if (path.isForInStatement() || path.isForStatement()) {
+        const init = tree.init || tree.left;
+        let scope = parent;
```
codehag: we don't need this assignment

src/utils/parser/getScopes.js
```diff
@@ -0,0 +1,280 @@
...
+    return undefined;
+  }
+  return children.map(scope => {
+    let bindings = Object.create(null);
+    Object.keys(scope.names).forEach(n => {
+      switch (scope.names[n].kind) {
+        case "var":
+        case "let":
+        case "const":
+        case "param":
```
codehag: TIL that this is faster than using regex!

src/utils/parser/getScopes.js
```diff
@@ -0,0 +1,280 @@
...
+function _toGeneratedScopes(
+  children: TempScope[],
+  sourceId: SourceId
+): ?(GeneratedScope[]) {
+  if (!children || children.length === 0) {
+    return undefined;
+  }
+  return children.map(scope => {
+    let bindings = Object.create(null);
+    Object.keys(scope.names).forEach(n => {
```
codehag: this should probably be a reduce. I would expect forEach if we were triggering side effects.

```javascript
const bindings = Object.keys(scope.names).reduce((bindings, n) => {
  const name = scope.names[n]; // or something
  switch (name.kind) {
    case "var":
    case "let":
    case "const":
    case "param":
      const binding = name.refs.map(
         location => _fromBabelLocation(location, sourceId) 
      );
      return { ...bindings, [n]: binding };
    }
    return bindings;
}
```

this way we can also factor this out into its own function

src/utils/parser/getScopes.js
```diff
@@ -0,0 +1,280 @@
...
+  }
+}
+
+function _recordIdentifierPosition(
+  scope: TempScope,
+  name: string,
+  loc: BabelLocation
+) {
+  for (let s = scope; s; s = s.parent) {
+    if (name in s.names) {
```
codehag: what is happening here? is it looking over s.names?

src/utils/parser/getScopes.js
```diff
@@ -0,0 +1,280 @@
...
+        return;
+      }
+      if (path.isBlockStatement()) {
+        const hasLet = tree.body.some(node => {
+          if (!isNode(node, "VariableDeclaration")) {
+            return false;
+          }
+          return node.kind === "let" || node.kind === "const";
+        });
+        let scope = parent;
```
codehag: scope doesnt need to be declared here.

src/utils/parser/getScopes.js
```diff
@@ -0,0 +1,280 @@
...
+    }
+  }
+}
+
+function _toGeneratedScopes(
+  children: TempScope[],
+  sourceId: SourceId
+): ?(GeneratedScope[]) {
+  if (!children || children.length === 0) {
+    return undefined;
```
codehag: perhaps we can return an empty array, so that we are not changing the type of the return value

src/utils/parser/getScopes.js
```diff
@@ -0,0 +1,280 @@
...
+        const hasLet = tree.body.some(node => {
+          if (!isNode(node, "VariableDeclaration")) {
+            return false;
+          }
+          return node.kind === "let" || node.kind === "const";
+        });
+        let scope = parent;
+        if (hasLet) {
+          savedParents.set(path, parent);
+          scope = _createTempScope("Block", parent, location);
```
codehag: can be declared with const

src/utils/parser/getScopes.js
```diff
@@ -0,0 +1,280 @@
...
+  if (!parsedScopes) {
+    parsedScopes = _parseJSScopes(getSource(sourceId));
+    parsedScopesCache.set(sourceId, parsedScopes);
+  }
+
+  // Find inner most
+  let search = parsedScopes;
+  let found = [];
+  while (search) {
+    let foundOne = search.some(s => {
```
codehag: search.some stops execution after one is found; but we are modifying the found array here and the search object. perhaps what we want here again is reduce?

```javascript
const found = search.reduce((found, s) => {
  if (
    _compareLocations(s.start, location) <= 0 &&
    _compareLocations(location, s.end) < 0
  ) {
    return [s, ...found];
  }
  return found;
})
```

you can also factor that out into an independant function and do (imagining that the reduce is its own function, and the map is another):
```javascript
return search.reduce(findLocations).map(createScope);
```

src/utils/parser/sources.js
```diff
@@ -0,0 +1,25 @@
+// @flow
+
+import type { Source, SourceId } from "debugger-html";
+
+var cachedSources = new Map();
```
codehag: why do we want to use var here? would let work?

src/utils/parser/getScopes.js
```diff
@@ -0,0 +1,280 @@
...
+  });
+  return _toGeneratedScopes(parent.scopes, source.id);
+}
+
+function _compareLocations(a: Location, b: Location): number {
+  // $FlowIgnore
+  return a.line == b.line ? a.column - b.column : a.line - b.line;
+}
+
+var parsedScopesCache = new Map();
```
codehag: we are using var here, is there a particular reason? i would expect let to be used in this case

src/utils/parser/sources.js
```diff
@@ -0,0 +1,25 @@
...
+
+export function setSource(source: Source) {
+  cachedSources.set(source.id, source);
+}
+
+export function getSource(sourceId: SourceId): Source {
+  if (!cachedSources.has(sourceId)) {
+    throw new Error(`${sourceId} was not provided.`);
+  }
+  // $FlowIgnore
```
codehag: why are we ignoring here?

src/utils/parser/tests/getScopes.spec.js
```diff
@@ -0,0 +1,45 @@
+import getScopes from "../getScopes";
+import { setSource } from "../sources";
+
+describe("Parser.getScopes", () => {
+  it("parses simple generated/minified es5 source", () => {
+    const source = {
```
codehag: nit: we can use the parser's test helper and save this in its own file i think

its used via
```javascript
import { getSource } from "../helpers";

//...
const source = getSource("getScopes/source1");
```

src/utils/parser/getScopes.js
```diff
@@ -0,0 +1,280 @@
...
+      start: _fromBabelLocation(scope.loc.start, sourceId),
+      end: _fromBabelLocation(scope.loc.end, sourceId),
+      type: scope.type.toLowerCase(),
+      bindings,
+      children: _toGeneratedScopes(scope.scopes, sourceId)
+    };
+  });
+}
+
+function _parseJSScopes(source: Source) {
```
codehag: what do the `_` signify in these function names?

jasonLaster: probably private

codehag: thats what i thought, but wouldn't it be private by default, since it isnt exported?

src/utils/parser/getScopes.js
yurydelendik: it checks in `name` is a property in s.names dictionary. related to the `Object.create(null)` above

src/utils/parser/getScopes.js
yurydelendik: That might cause of creation many empty leaf objects.

src/utils/parser/sources.js
yurydelendik: Even if I did the check above, flow is convinced cacheSources.get can be undefined. Changing to ((x: any): Source) cast.

yurydelendik:
Moved scope parsing related code to the https://github.com/devtools-html/devtools-core/pull/656

src/utils/parser/getScopes.js
```diff
@@ -0,0 +1,280 @@
...
+    _parseDeclarator(declaratorId.argument, targetScope, kind);
+  }
+}
+
+function _recordIdentifierPosition(
+  scope: TempScope,
+  name: string,
+  loc: BabelLocation
+) {
+  for (let s = scope; s; s = s.parent) {
```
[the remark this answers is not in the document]
yurydelendik: The recursive functions (even with tails call) are not really efficient in comparison with just a loop. I recommend to not introduce any unneeded recursion for simple algorithms, and in this case simple loop easier to read.

src/utils/parser/sources.js
```diff
@@ -0,0 +1,24 @@
...
+export function getSource(sourceId: SourceId): Source {
+  if (!cachedSources.has(sourceId)) {
+    throw new Error(`${sourceId} was not provided.`);
+  }
+  return ((cachedSources.get(sourceId): any): Source);
+}
+
+export function clearSources() {
+  cachedSources = new Map();
+}
```
jasonLaster: almost all of our APIs now expect to get a `source`. 

I agree that this is a performance issue, but I think we should make the fix in a separate PR, that touches all of the spots.

https://github.com/devtools-html/debugger.html/blob/master/src/utils/parser/steps.js#L6-L8
https://github.com/devtools-html/debugger.html/blob/master/src/utils/parser/getSymbols.js#L335-L336

#3953

yurydelendik:
Depends on #4008

PR #3855, opened 2017-09-02
Renamed Primary Panes

pradeepgangwar:
### Summary of Changes

* Renamed Sources View tab to Sources 
* Renamed Outline View tab to Outline

### Screenshot

screenshot-2017-9-2 debugger

### DIFF
--- a/assets/panel/debugger.properties
+++ b/assets/panel/debugger.properties
@@ -237,6 +237,14 @@ breakpointMenuItem.disableAll.accesskey=k
 breakpointMenuItem.deleteAll=Remove all breakpoints
 breakpointMenuItem.deleteAll.accesskey=a
 
+# LOCALIZATION NOTE (panel.sources): Label displayed in the primary
+# pane. %S is replaced by the string sources.
+panel.sources=Sources
+
+# LOCALIZATION NOTE (panel.outline): Label displayed in the primary
+# pane. %S is replaced by the string outline.
+panel.outline=Outline
+
 # LOCALIZATION NOTE (breakpoints.header): Breakpoints right sidebar pane header.
 breakpoints.header=Breakpoints
 
--- a/assets/panel/debugger.properties
+++ b/assets/panel/debugger.properties
@@ -237,14 +237,6 @@ breakpointMenuItem.disableAll.accesskey=k
 breakpointMenuItem.deleteAll=Remove all breakpoints
 breakpointMenuItem.deleteAll.accesskey=a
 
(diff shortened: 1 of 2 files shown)


src/components/PrimaryPanes/index.js
```diff
@@ -71,7 +71,7 @@ class PrimaryPanes extends Component {
         onClick={() => this.showPane("outline")}
         key="outline-tab"
       >
-        Outline View
```
jasonLaster: Oops, can we move these to `assets/panel/debugger.properties` w/ l10n

https://github.com/devtools-html/debugger.html/blob/master/docs/local-development.md#l10n

pradeepgangwar:
@jasonLaster So i should write these in debugger.properties and fetch from there.

jasonLaster:
Yep

pradeepgangwar:
@jasonLaster Done.

assets/panel/debugger.properties
```diff
@@ -237,6 +237,14 @@ breakpointMenuItem.disableAll.accesskey=k
 breakpointMenuItem.deleteAll=Remove all breakpoints
 breakpointMenuItem.deleteAll.accesskey=a
 
+# LOCALIZATION NOTE (panel.sources): Label displayed in the primary
+# pane. %S is replaced by the string sources.
+panel.sources=Sources
```
jasonLaster: looks like already have this string: https://github.com/pradeepgangwar/debugger.html/blob/ed91f18bda935b6bc157413bbe71c6f427b3e1ab/assets/panel/debugger.properties#L434

assets/panel/debugger.properties
```diff
@@ -237,6 +237,14 @@ breakpointMenuItem.disableAll.accesskey=k
...
+
+# LOCALIZATION NOTE (panel.outline): Label displayed in the primary
+# pane. %S is replaced by the string outline.
+panel.outline=Outline
```
jasonLaster: lets move this one closer to the original Sources string

assets/panel/debugger.properties
flodolo: As I wrote a few days ago in Bugzilla, adding localization comments is not mandatory, they need to be helpful. In this case, try reading the localization comment and tell me if it matches the string (same for the next one) ;-)

@jasonLaster 
String reuse in different contexts should be avoided. Let's keep this separate from the one use from the header.

pradeepgangwar:
@jasonLaster sorry i didn't see it was there. I have updated.

Issue #3814, opened 2017-08-29
[workers] get worker list


jasonLaster:
Relevant links:
* http://searchfox.org/mozilla-central/source/devtools/server/actors/child-process.js#119
* http://searchfox.org/mozilla-central/search?q=workerListChanged&path=
* see about:debugging examples for the toolbox link

Connect the event (useful when the list of workers changes)
```diff
diff --git a/src/client/firefox/events.js b/src/client/firefox/events.js
index ff8748b..05f9bb1 100644
--- a/src/client/firefox/events.js
+++ b/src/client/firefox/events.js
@@ -65,10 +65,15 @@ function newSource(_: "newSource", { source }: SourcePacket) {
   }
 }

+function workerListChanged() {
+  debugger;
+}
+
 const clientEvents = {
   paused,
   resumed,
-  newSource
+  newSource,
+  workerListChanged
 };

 export { setupEvents, clientEvents };
 ```
 
 Here is a diff for adding a debuggee reducer, and getting the workers:
 https://gist.github.com/jasonLaster/2b60af0711d1e768314df12d8a24726a

nyrosmith:
Working on it!

PR #3856, opened 2017-09-02
[WIP] List workers

nyrosmith:
Associated Issue: #3814 

This is a WIP and the code is quite naive now.

* get the debuggee's workers
* set them in the redux store
* show the workers in the Workers panel


src/components/SecondaryPanes/Workers.css
```diff
@@ -0,0 +1,13 @@
+.workers-list * {
+  -moz-user-select: none;
+  user-select: none;
+}
```
jasonLaster: We use autoprefix so you don't need the moz prefix

src/reducers/debuggee.js
```diff
@@ -0,0 +1,15 @@
+const initialState = {
```
jasonLaster: Let's add flow '// @flow'

src/components/SecondaryPanes/Workers.css
nyrosmith: I looked how breakpoints are styled so I'll remove it there too

nyrosmith:
Is this failing test really caused by my changes?

jasonLaster:
nope, those tests are not you. They're two known intermittents

jasonLaster:
@nyrosmith i pushed some extra boilerplate for the reducers

* immutable
* flow types
* reselect

The patch is not done, it is just a start. Happy to help you finish, but thought it would be nice to leave it as an exercise so you can learn the architecture. some of this probably seems like a bit much for a small reducer, but it's helpful as it scales.

src/client/firefox.js
```diff
@@ -35,6 +35,10 @@ export async function onConnect(connection: any, actions: Object) {
     wasmBinarySource: supportsWasm
   });
 
+  threadClient._parent.listWorkers().then(workers =>
```
bomsy: @nyrosmith `threadClient._parent` can be undefined, you probably need to check if it exists before you call `listWorkers`

nyrosmith:
Everything green :-)

giphy-downsized

PR #3857, opened 2017-09-02
Make case consistent for search sources

darkwing:
Search shows two option:  "Search Sources" and "Find in files".  I find the change in case odd, so changing "sources" to lowercase makes the labels more consistent.

### DIFF
--- a/assets/panel/debugger.properties
+++ b/assets/panel/debugger.properties
@@ -454,7 +454,7 @@ welcome.searchFunction=%S to search for functions in file
 
 # LOCALIZATION NOTE (sourceSearch.search): The center pane Source Search
 # prompt for searching for files.
-sourceSearch.search=Search Sources…
+sourceSearch.search=Search sources…
 
 # LOCALIZATION NOTE (sourceSearch.noResults): The center pane Source Search
 # message when the query did not match any of the sources.


darkwing:
Transient error

assets/panel/debugger.properties
```diff
@@ -454,7 +454,7 @@ welcome.searchFunction=%S to search for functions in file
 
 # LOCALIZATION NOTE (sourceSearch.search): The center pane Source Search
 # prompt for searching for files.
-sourceSearch.search=Search Sources…
+sourceSearch.search=Search sources…
```
jasonLaster: this is annoying, but you'll need to bump the key:

```
# LOCALIZATION NOTE (sourceSearch.search2): The center pane Source Search
# prompt for searching for files.
sourceSearch.search2=Search sources…
```

and use the new key in the project

flodolo: You don't need to bump keys for changing the case ;-) (that's why I approved explicitly).
https://developer.mozilla.org/en-US/docs/Mozilla/Localization/Localization_content_best_practices#Changing_existing_strings

That's a matter of internal consistency for English, other languages have their own rules and don't necessarily need to reflect this change.

jasonLaster: Hey, @flodolo it's great to have you around :)

flodolo: It definitely makes debugger updates to m-c less surprising :-)

Issue #3694, opened 2017-08-17
[ProjectSearch] fix intermittent test


jasonLaster:
We are getting an intermittent with  `CmdOrCtrl+Shift+F` sometimes not opening project search.

perhaps we are trying to call it before the search UI is ready to handle it?

PR #3875, opened 2017-09-04
fix project search intermittent

jasonLaster:
Associated Issue: #3694

### Summary of Changes

stab in the dark because we know that the mouse is in the top left. The tests pass consistently on a mac...


codehag:
Still happening :/ any ideas? should we merge this anyway?

jasonLaster:
nah, i found the root cause which i'm fixing in a different pr...

Issue #3876, opened 2017-09-04
[project Text Search] The search results should remain in a specific order as more results get added


bomsy:
The search results are currently in a Map which is unorderd, so as more data come in, it cause results, all over as shown  below

![](http://g.recordit.co/UFPfb2uiET.gif)

Ideally we should have all new results always added to the bottom.

PR #3877, opened 2017-09-04
Search results order

bomsy:
Associated Issue: #3876

### Summary of Changes

* Change the results from a `Map` to a `List` for order consistensy
* Added new action and reducer to clear search results for every new search
* Added test for `clearSearchResults`
* Remove unused `getTextSearchResult` selector
* There was need for `inputFocused` to be a state property, it was causing unnecessary re-renders
*  Fixed the focused issue. it should now focus on the first match of the first result found by default

#### Before
- New results get added any where (ordered by the sourceId), so the initially focused item moves

#### After
- New results always get added to the bottom (in order of how the results are found), so the focused node is always at the top

### DIFF
--- a/src/actions/project-text-search.js
+++ b/src/actions/project-text-search.js
@@ -31,7 +31,7 @@ export function clearSearchQuery() {
 }
 
 export function clearSearchResults() {
-  return ({ dispatch }) => {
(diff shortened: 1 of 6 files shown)


jasonLaster:
This looks really good 👍

PR #3888, opened 2017-09-04
Fix project search test

jasonLaster:
### Summary of Changes

While looking into why the project search test would time out I found that we had some funky escape handling. It turned out that Symbol Modal never un-mounted, which meant that it was always handling escape keys.

* the symbol modal is now conditionally rendered
* Escape handling is now in App.js
* some custom handling has been moved to clearActiveSearch 
* project and text search are a bit cleaner now too

### DIFF
--- a/src/actions/ui.js
+++ b/src/actions/ui.js
@@ -1,18 +1,27 @@
 // @flow
-import { getSource, getActiveSearchState } from "../selectors";
+import { getSource, getActiveSearch } from "../selectors";
 import type { ThunkArgs } from "./types";
 import type { ActiveSearchType, SymbolSearchType } from "../reducers/ui";
+import { clearSourceSearchQuery } from "./source-search";
 
 export function closeActiveSearch() {
-  return {
-    type: "TOGGLE_ACTIVE_SEARCH",
-    value: null
+  return ({ getState, dispatch }: ThunkArgs) => {
+    const activeSearch = getActiveSearch(getState());
+
+    if (activeSearch == "source") {
+      dispatch(clearSourceSearchQuery());
+    }
+
+    dispatch({
+      type: "TOGGLE_ACTIVE_SEARCH",
+      value: null
+    });
   };
 }
 
 export function setActiveSearch(activeSearch?: ActiveSearchType) {
   return ({ dispatch, getState }: ThunkArgs) => {
-    const activeSearchState = getActiveSearchState(getState());
+    const activeSearchState = getActiveSearch(getState());
     if (activeSearchState === activeSearch) {
       return;
     }
--- a/src/components/App.js
+++ b/src/components/App.js
@@ -4,7 +4,11 @@ import React, { PropTypes, Component } from "react";
 import { connect } from "react-redux";
 import { bindActionCreators } from "redux";
 import actions from "../actions";
-import { getSelectedSource, getPaneCollapse } from "../selectors";
+import {
+  getSelectedSource,
+  getPaneCollapse,
+  getActiveSearch
+} from "../selectors";
 import type { SourceRecord } from "../reducers/sources";
 import { isVisible } from "../utils/ui";
 
@@ -38,7 +42,10 @@ type Props = {
   selectSource: Function,
   selectedSource: SourceRecord,
   startPanelCollapsed: boolean,
-  endPanelCollapsed: boolean
+  closeActiveSearch: () => void,
+  endPanelCollapsed: boolean,
+  activeSearch: string,
+  setActiveSearch: string => void
 };
 
 class App extends Component {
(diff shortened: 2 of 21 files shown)


src/actions/ui.js
```diff
@@ -1,18 +1,27 @@
...
+import { clearSourceSearchQuery } from "./source-search";
 
 export function closeActiveSearch() {
-  return {
-    type: "TOGGLE_ACTIVE_SEARCH",
-    value: null
+  return ({ getState, dispatch }: ThunkArgs) => {
+    const activeSearch = getActiveSearch(getState());
+
+    if (activeSearch == "source") {
```
nchevobbe: I am not sure I understand this.
Why do we check for `source` ? A comment, or using a constant for this `"source"` value would help understanding

src/components/App.js
```diff
@@ -74,10 +85,51 @@ class App extends Component {
...
+  onEscape(shortcut, e) {
+    const { activeSearchState, closeActiveSearch } = this.props;
+
+    if (activeSearchState) {
+      e.preventDefault();
+      closeActiveSearch();
+    }
+  }
+
+  toggleSymbolModal(_, e: SyntheticEvent) {
```
nchevobbe: could `_` be renamed as `shortcut`, like in `onEscape` (or rename it on `onEscape`), se we have a consistent way of dealing with this argument

src/components/App.js
```diff
@@ -74,10 +85,51 @@ class App extends Component {
...
+    const {
+      selectedSource,
+      activeSearchState,
+      closeActiveSearch,
+      setActiveSearch
+    } = this.props;
+
+    e.preventDefault();
+    e.stopPropagation();
```
nchevobbe: Do we want to stop the propagation before checking `selectedSource` ?
I'd say that if the key stroke we hit have no effect in this functionm it should bubble up.

src/components/App.js
```diff
@@ -74,10 +85,51 @@ class App extends Component {
...
+
+    if (!selectedSource) {
+      return;
+    }
+
+    if (activeSearchState == "symbol") {
```
nchevobbe: I don't understand this test here. A comment or using a constant for "symbol" might help understanding better what it does.

src/components/App.js
```diff
@@ -206,7 +268,8 @@ export default connect(
   state => ({
     selectedSource: getSelectedSource(state),
     startPanelCollapsed: getPaneCollapse(state, "start"),
-    endPanelCollapsed: getPaneCollapse(state, "end")
+    endPanelCollapsed: getPaneCollapse(state, "end"),
+    activeSearchState: getActiveSearch(state)
```
nchevobbe: nit: it feels a bit off to have renamed the selector to `getActiveSearch` and use a `activeSearchState` prop. Renaming it to `activeSearch` would be more cnosistent IMO

src/components/ProjectSearch/TextSearch.js
```diff
@@ -61,7 +57,12 @@ export default class TextSearch extends Component {
   }
 
   onKeyDown(e) {
+    if (e.key === "Escape") {
```
nchevobbe: it looks like this should be handled by the test https://github.com/devtools-html/debugger.html/pull/3888/files#diff-a98f82ecdc3c081ceb80399442d12e98R66.
So I guess we only want the key event to bubble up.
Maybe we could only stop the propagation after the check on l.66 ?

src/components/ProjectSearch/index.js
```diff
@@ -61,7 +61,7 @@ class ProjectSearch extends Component {
       L10N.getStr("sources.search.alt.key")
     ];
     searchKeys.forEach(key => shortcuts.off(key, this.toggleSourceSearch));
-    shortcuts.off("Escape", this.onEscape);
+    // shortcuts.off("Escape", this.onEscape);
```
nchevobbe: nit: Remove the line

src/test/mochitest/head.js
```diff
@@ -770,12 +773,18 @@ function findAllElements(dbg, elementName, ...args) {
  * @return {Promise}
  * @static
  */
-function clickElement(dbg, elementName, ...args) {
+async function clickElement(dbg, elementName, ...args) {
   const selector = getSelector(elementName, ...args);
+  await waitForElement(dbg, selector);
```
nchevobbe: `waitForElement` could directly return the element we wait for, it would save us the extra `findElement` on the next line.

src/components/App.js
jasonLaster: yeah, we figure w/ flow we can use strings... bc they're type checked, but perhaps we should be stricter w/ constants

PR #3889, opened 2017-09-05
Extract file search redux backend

jasonLaster:
### Summary of Changes

I realized https://github.com/devtools-html/debugger.html/pull/3888 was too big, so i'm taking the file search redux part out and making it its own PR. 

Why extract:

* I'm realizing now that actions/ui was primarily file search. 
* Now that we have redux backends for project search, text search etc, it only seems fair
* the file search (SearchBar) component is way too complicated
* I'd like to have the search bar component only mounted when it is shown, to do this we need a permanent store. 

Other questions / Musings:

The redux api (actions/selectors) need to be be really long like `setFileSearchQuery` because they are shared among the entire app. It would be really cool if we could begin having sub-apps which are redux backends that are only connected to their root component. 

### Test Plan

Re-use existing tests

### DIFF
--- a/src/actions/file-search.js
+++ b/src/actions/file-search.js
@@ -1,8 +1,8 @@
 // @flow
 
 import { find, findNext, findPrev, removeOverlay } from "../utils/editor";
-import { getMatches } from "../utils/search";
-import type { ThunkArgs } from "./types";
+import { getMatches } from "../workers/search";
+import type { ThunkArgs, Match } from "./types";
 
 import {
   getSelectedSource,
@@ -10,7 +10,12 @@ import {
   getFileSearchQuery,
   getFileSearchResults
 } from "../selectors";
-type Match = Object;
+
+import {
+  closeActiveSearch,
+  clearHighlightLineRange,
+  setActiveSearch
+} from "./ui";
 type Editor = Object;
 
 export function doSearch(query: string, editor: Editor) {
@@ -47,14 +52,16 @@ export function updateSearchResults(
 
   return {
     type: "UPDATE_SEARCH_RESULTS",
-    matches,
-    matchIndex,
-    count: matches.length,
-    index: characterIndex
+    results: {
+      matches,
+      matchIndex,
+      count: matches.length,
+      index: characterIndex
+    }
   };
 }
 
-export async function searchContents(query: string, editor: Object) {
+export function searchContents(query: string, editor: Object) {
   return async ({ getState, dispatch }: ThunkArgs) => {
     const modifiers = getFileSearchModifiers(getState());
     const selectedSource = getSelectedSource(getState());
@@ -79,6 +86,7 @@ export async function searchContents(query: string, editor: Object) {
     const { ch, line } = find(ctx, query, true, _modifiers);
 
     dispatch(updateSearchResults(ch, line, matches));
+    return {};
   };
 }
 
@@ -88,14 +96,14 @@ export function traverseResults(rev: boolean, editor: Editor) {
       return;
     }
 
-    const ctx = { editor, cm: editor.codeMirror };
+    const ctx = { ed: editor, cm: editor.codeMirror };
 
     const query = getFileSearchQuery(getState());
     const modifiers = getFileSearchModifiers(getState());
-    const { searchResults: matches } = getFileSearchResults();
+    const { matches } = getFileSearchResults(getState());
 
     if (query === "") {
-      this.props.setActiveSearch("file");
+      dispatch(setActiveSearch("file"));
     }
 
     if (modifiers) {
@@ -103,7 +111,24 @@ export function traverseResults(rev: boolean, editor: Editor) {
       const { ch, line } = rev
         ? findPrev(ctx, query, true, modifiers.toJS())
         : findNext(ctx, query, true, modifiers.toJS());
-      this.updateSearchResults(ch, line, matchedLocations);
+
+      dispatch(updateSearchResults(ch, line, matchedLocations));
     }
+  };
+}
+
+export function closeFileSearch(editor: Editor) {
+  return ({ getState, dispatch }: ThunkArgs) => {
+    const modifiers = getFileSearchModifiers(getState());
+    const query = getFileSearchQuery(getState());
+
+    if (editor && modifiers) {
+      const ctx = { ed: editor, cm: editor.codeMirror };
+      removeOverlay(ctx, query, modifiers.toJS());
+    }
+
+    dispatch(setFileSearchQuery(""));
+    dispatch(closeActiveSearch());
+    dispatch(clearHighlightLineRange());
   };
 }
--- a/src/actions/file-search.js
+++ b/src/actions/file-search.js
@@ -1,0 +1,108 @@
+// @flow
+
+import { find, findNext, findPrev, removeOverlay } from "../utils/editor";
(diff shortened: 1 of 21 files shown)


src/reducers/file-search.js
```diff
@@ -0,0 +1,118 @@
+// @flow
+
+/**
+ * UI reducer
```
bomsy: Fi

src/reducers/file-search.js
```diff
@@ -0,0 +1,118 @@
...
+ * @module reducers/ui
```
bomsy: File search reducer

src/actions/file-search.js
```diff
@@ -0,0 +1,108 @@
+// @flow
+
+import { find, findNext, findPrev, removeOverlay } from "../utils/editor";
+import { getMatches } from "../utils/search";
+import type { ThunkArgs } from "./types";
+
+// import {} from "./index";
```
bomsy: remove this line

src/actions/file-search.js
```diff
@@ -0,0 +1,108 @@
...
+    matches,
+    matchIndex,
+    count: matches.length,
+    index: characterIndex
+  };
+}
+
+async function searchContents(query: string, editor: Object) {
+  return async ({ getState, dispatch }) => {
+    // const { selectedSource, modifiers, editor: ed } = this.props;
```
codehag: we can remove this line as well

src/actions/file-search.js
```diff
@@ -0,0 +1,108 @@
...
+    if (!editor) {
+      return;
+    }
+
+    const ctx = { editor: ed, cm: editor.codeMirror };
+
+    const query = getFileSearchQuery(getState());
+    const modifiers = getFileSearchModifiers(getState());
+    const { searchResults: matches } = getFileSearchResults();
+    // const { query, modifiers, searchResults: { matches } } = this.props;
```
codehag: this one too

src/reducers/file-search.js
```diff
@@ -0,0 +1,118 @@
...
+};
+
+export type SearchResults = {
+  matches: Array<MatchedLocations>,
+  matchIndex: number,
+  index: number,
+  count: number
+};
+
+export type UIState = {
```
codehag: should this also be FileSearchState?

src/reducers/file-search.js
```diff
@@ -0,0 +1,118 @@
...
+    case "SET_QUERY": {
+      return state.set("query", action.query);
+    }
+
+    case "SET_SEARCH_RESULTS": {
+      return state.set("searchResults", action.results);
+    }
+
+    case "TOGGLE_MODIFIER": {
+      const actionVal = !state.getIn(["modifiers", action.modifier]);
```
codehag: I'm on the fence about this but i will mention it anyway; is action.modifier a string? could we do something like  
```javascript
prefs[`fileSearch${action.modifier}`]
```
(i think it needs a first letter uppercase, but you get the idea)

since we seem to be doing the same thing each time, and we might not need this extra layer

jasonLaster: lol we tried to do this kind of thing before, but it is too magical.

a big style reason not to, is that it is nice to grep for pref keys, just like it's nice to grep for l10n keys

src/components/Editor/SearchBar.js
```diff
@@ -122,8 +121,7 @@ class SearchBar extends Component {
       shiftSearchAgainShortcut
     } = getShortcuts();
 
-    shortcuts.on(searchShortcut, (_, e) => this.toggleSearch(e));
```
wldcordeiro: Did you mean to remove the search shortcut?

jasonLaster: yes, because the search bar is now unmounted when it is not open. See the Editor/index render method

wldcordeiro: I see that. But now there isn't a way to enable the search bar?

jasonLaster: oh, we should probably fix that :) 

I think i was going to add a shortcut in App.js or Editor/index

wldcordeiro:
Working on this some more. Had to rework some of the code and rebase. Apologies for the delays.

wldcordeiro:
Almost ready to go! One last behavior to test and this is ready. 😄

wldcordeiro:
Travis keeps failing but I believe I've got it working like expected 😁

jasonLaster:
hmm, the failure seems relevant. I'll look tomorrow.

Lets finish this up

wldcordeiro:
Got it down to one failing mochitest that's unrelated? It's the browser_debug-console.js test. 🤔

src/test/mochitest/browser_dbg-console.js
```diff
@@ -41,6 +41,6 @@ add_task(async function() {
   pressKey(dbg, "Escape");
 
   // Second time to hide console
-  pressKey(dbg, "Escape");
+  pressKey(dbg, "CtrlEscape");
```
wldcordeiro: Adding this key code worked but it feels wrong. 🙁 @jasonLaster Care to dig in a bit?

src/test/mochitest/browser_dbg-search-file.js
```diff
@@ -35,11 +35,12 @@ add_task(async function() {
   const el = getFocusedEl(dbg);
 
   type(dbg, "con");
-  await waitForSearchState(dbg);
 
   const state = cm.state.search;
 
   pressKey(dbg, "Enter");
+  pressKey(dbg, "Enter");
```
wldcordeiro: I don't know why the second ent here seems to be the trick.

src/test/mochitest/browser_dbg-console.js
jasonLaster: sure...

src/test/mochitest/browser_dbg-search-file.js
jasonLaster: also weird...

codehag: looks like it might be a timing issue

codehag:
can you rebase and I will take a look at the mochi test issue?

src/components/Editor/SearchBar.js
```diff
@@ -149,152 +127,48 @@ class SearchBar extends Component {
...
-        this.doSearch(selection);
-      }
-      this.selectSearchInput();
-    }
-  }
-
-  setSearchValue(value: string) {
-    const searchInput = this.searchInput();
-    if (value == "" || !searchInput) {
+    if (!this.props.editor) {
```
codehag: moving this below `e.stopPropagation` and` e.preventDefault` is what was causing the escape button bug.

src/components/Editor/SearchBar.js
```diff
@@ -305,6 +179,11 @@ class SearchBar extends Component {
     this.traverseResults(e, e.shiftKey);
     e.preventDefault();
   }
+
+  getInputRef(c) {
+    this.searchInput = c;
```
codehag: this change doesnt make much sense to me, because this allows us to highjack the search and allows us to set the search field in two ways, via props and through setSearchValue. 

we should instead have local state and props, props being the state that is sent to the search worker.

codehag:
Closing as mentioned above.

PR #3890, opened 2017-09-05
Move code coverage button title to l10n

darkwing:
Spotted this, thought it would be best to move to l10n.


jasonLaster:
yep, we should ship code coverage one of these days too :)

PR #3900, opened 2017-09-06
Show full editor context menu for blackboxed tab

darkwing:
There's no reason not to show the full context menu for blackboxed sources.  We still want to copy, get the URL, etc.


codehag:
i have no issues with this; @jasonLaster ?

jasonLaster:
nope, great idea

PR #3902, opened 2017-09-06
Show parameter signature within outline

darkwing:
I enjoy a function list but I find that list not very helpful without knowing also knowing the parameter names.

outline-args

This requires a test fix but I wanted to take the temperature on adding paramaters to the outline panel.  I much prefer it.


jasonLaster:
Thanks! I definitely prefer the full signature

Issue #3909, opened 2017-09-06
[themes] Update the sidebar and accordion colors for the dark theme


gabrielluong:
In the dark theme:
Sidebar background color is var(--grey-90)
Accordion header background color is #141416

PR #3912, opened 2017-09-06
[themes] Update the sidebar and accordion colors for the dark theme (…

gabrielluong:
…#3909)

Associated Issue: #3909 

### Summary of Changes

* In the dark theme, Sidebar background color is var(--grey-90) and Accordion header background color is #141416.
* In the sources list, and add watch expression button, the background color needs to be var(--theme-sidebar-background).
* The accordion header needs to be #141416 only in the dark theme and var(--theme-toolbar-background) in the light.

### Screenshots

Before:
screen shot 2017-09-06 at 4 48 56 pm

After:
screen shot 2017-09-06 at 4 37 00 pm

### DIFF
--- a/src/components/PrimaryPanes/Sources.css
+++ b/src/components/PrimaryPanes/Sources.css
@@ -43,6 +43,7 @@
 }
 
 .sources-list {
+  background-color: var(--theme-sidebar-background);
   flex: 1;
   display: flex;
   overflow-x: hidden;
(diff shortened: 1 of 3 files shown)


src/components/PrimaryPanes/Sources.css
```diff
@@ -43,6 +43,7 @@
 }
 
 .sources-list {
+  background-color: var(--theme-sidebar-background);
```
jasonLaster: why the list and not the panel?

gabrielluong: Fixed.

Issue #3910, opened 2017-09-06
[Tabs] selecting a source in the show more dropdown doesn’t select it


jasonLaster:
It looks like selecting a source in the show more dropdown doesn’t select it. My guess is that it moves the tab, without selecting it.

[Steps to reproduce]:
1. Launch Firefox.
2. Access the https://devtools-html.github.io/debugger-examples/examples/todomvc/ webpage.
3. Open multiple files from the source list pane.
4. Open the list view dropdown.
5. Select a file from the dropdown.

[Expected result]:
The selected file is focused and it's code is displayed in the Source pane.

[Actual result]:
The selected file is not focused and the code from the previously selected file is displayed in the source pane.

![files.gif](https://shipusercontent.com/7d01caed7cf17cd6dc87f20359534a83/files.gif)

PR #3913, opened 2017-09-06
fix hidden tabs click not showing the source

bomsy:
Associated Issue: #3910

Here's the Pull Request Doc

### Summary of Changes

* Use the `selectSource` instead as it moves the tab to the first spot and also selects the source.

### Test Plan

Example test plan:

1. Launch Firefox.
2. Access the https://devtools-html.github.io/debugger-examples/examples/todomvc/ webpage.
3. Open multiple files from the source list pane.
4. Open the list view dropdown.
5. Select a file from the dropdown.

[Expected result]:
The selected file is focused and it's code is displayed in the Source pane.

[Actual result]:
The selected file is not focused and the code from the previously selected file is displayed in the source pane.

### DIFF
--- a/src/components/Editor/Tabs.js
+++ b/src/components/Editor/Tabs.js
@@ -314,13 +314,10 @@ class SourceTabs extends PureComponent {
   }
 
   renderDropdownSource(source: SourceRecord) {
-    const { moveTab } = this.props;
+    const { selectSource } = this.props;
     const filename = getFilename(source.toJS());
 
-    const onClick = () => {
-      const tabIndex = 0;
-      moveTab(source.get("url"), tabIndex);
-    };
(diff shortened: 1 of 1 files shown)


src/components/Editor/Tabs.js
```diff
@@ -314,13 +314,10 @@ class SourceTabs extends PureComponent {
...
   renderDropdownSource(source: SourceRecord) {
-    const { moveTab } = this.props;
+    const { selectSource } = this.props;
     const filename = getFilename(source.toJS());
 
-    const onClick = () => {
-      const tabIndex = 0;
-      moveTab(source.get("url"), tabIndex);
-    };
+    const onClick = () => selectSource(source.get("id"));
```
darkwing: We can probably move this into the JSX since it's so small.

Issue #3782, opened 2017-08-25
[Project Search] File path does not appear in nightly


bomsy:
The file paths for the results do not show up in nightly

![image](https://user-images.githubusercontent.com/792924/29727932-d012e466-89cd-11e7-9a77-fe03e8943dda.png)

codehag:
thanks for reporting it bomsy!

jasonLaster:
I was able to reproduce when debugging the debugger.
Maybe this is a problem w/ original files:

<img src='https://shipusercontent.com/6320cf5ed8e5fbdf08b5779450d55091/Screen%20Shot%202017-08-31%20at%2011.35.28%20AM.png' title='Screen Shot 2017-08-31 at 11.35.28 AM.png' width=597>

jasonLaster:
this is s 

```
diff --git a/src/utils/sources-tree/tests/utils.spec.js b/src/utils/sources-tree/tests/utils.spec.js
index a401f5d..67a7a5b 100644
--- a/src/utils/sources-tree/tests/utils.spec.js
+++ b/src/utils/sources-tree/tests/utils.spec.js
@@ -75,6 +75,12 @@ describe("sources tree", () => {
         relPath
       );
     });
+
+    it("supports other schemes", () => {
+      expect(getRelativePath("webpack:///Users/jlaster/src/mozilla/debugger.html/node_modules/devtools-reps/src/reps/date-time.js")).toBe(
+        // not sure what we want in here
+      );
+    })
   });
```

bomsy:
i'm thinking we should fall back to showing the full path, when we find an unknown scheme

PR #3916, opened 2017-09-06
handle webpack schemes, add tests

bomsy:
Associated Issue: #3782

### Summary of Changes

* Support the webpack scheme
* Fallback to the full path for unknown schemes
* Added tests


src/utils/sources-tree/utils.js
```diff
@@ -76,7 +76,7 @@ export function createParentMap(tree: Node): WeakMap<Node, Node> {
 }
 
 export function getRelativePath(path: string) {
-  const re = /(http(?:s?):\/\/(?:www\.)?[a-z0-9\-.]+)\/(.*)/i;
+  const re = /((http(?:s?)|webpack):\/\/((?:www\.)?[a-z0-9\-.]+)?)\/(.*)/i;
```
jasonLaster: i think we should generalize this to be any scheme: `file` `javascript` basically it can be any set of letters

bomsy: makes sense!

src/utils/sources-tree/utils.js
```diff
@@ -76,7 +76,7 @@ export function createParentMap(tree: Node): WeakMap<Node, Node> {
 }
 
 export function getRelativePath(path: string) {
-  const re = /(http(?:s?):\/\/(?:www\.)?[a-z0-9\-.]+)\/(.*)/i;
```
jasonLaster: can we borrow some good ideas from `getUrl` to find the relative path: https://github.com/devtools-html/debugger.html/blob/master/src/utils/sources-tree/getURL.js#L17-L38

I think `url.parse` will give us what we want. at a minimum we won't have to write a regex to handle the protocol/scheme thing

Issue #3927, opened 2017-09-07
[SourceSearch] query is not cleared when using cmd+P


jasonLaster:
STR:

- go to http://firefox-dev.tools/debugger-examples/examples/todomvc/
- cmp+P, type app-view, valid with RETURN
- cmp+P, "app-view" is still in the input field

PR #3934, opened 2017-09-07
fix clearing source search

jasonLaster:
Associated Issue: #3927


darkwing:
I think something happened here...

jasonLaster:
I merged the fix in when I merged the 9-6 release. This is open so we can do any other follow up or testing

PR #3948, opened 2017-09-08
Issue Template

jasonLaster:
It'll be great to point people to resources for filing issues. Especially our slack and now our glitch template

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,18 +1,10 @@
-#### Thanks for filing an issue! :dog: 
+Thanks for filing an issue! 
 
-We try to answer common questions in our [docs]. Here are our most frequently asked questions:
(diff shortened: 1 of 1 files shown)


jasonLaster:
@digitarald @darkwing @bomsy  what do you think?

darkwing:
I'd be much more direct because I don't think many people will take the time to follow the links.  Check out what the A-Frame team has: 

https://github.com/aframevr/aframe/issues/new

Three questions and the entire directive set fits in the default size of the issues textarea element.

darkwing:
I would probably ask:

1.  What are you seeing?
2.  What did you expect to see?
3.  Screenshot of issue, if applicable

bomsy:
I agree with @darkwing, people might find it a bit tasking, jumping back and forth.
A couple of questions which would guide them should be fine.

Here is what the react team's looks like aswell

https://github.com/facebook/react/issues/new

darkwing:
Looks good!

Issue #3899, opened 2017-09-05
React Devtools throws Debugger into a loop


digitarald:
### STR

- Installl https://addons.mozilla.org/en-US/firefox/addon/react-devtools/
- Open Debugger

### Result

Constantly reloading source and unusable Debugger:

<img width="1182" alt="image" src="https://user-images.githubusercontent.com/8599/30086159-aa091978-924e-11e7-8ff4-7e06df5c2cbb.png">

jasonLaster:
perhaps this is because `browser.chrome` and remote debugging is on.

juliandescottes:
I can confirm I am getting this regularly while debugging. It doesn't happen when I open the debugger, but when I'm stepping the debugger will sometimes simply open this "eval code" tab and keep reloading it.

juliandescottes:
I think I have some ideas about this.

First of all the react extension is calling inspectedWindow.eval every 1s if React is not defined on the page: https://github.com/facebook/react-devtools/blob/2c1789627caa6d2fff3790564658a61cdf326b4b/shells/webextension/src/main.js#L21

So first of all I think this issue should mostly happen on sites which do not use React.

Now, looking at the stack trace, it seems like the request to perform an eval was received while debugging. And then instead of exiting the current frame (?) and resuming the execution we somehow enter the frame where the eval code from the extension is evaluated. But we can't load the source from the moz-extension protocol. Not sure why we keep reloading though.

I am not used to the debugger server terminology so this might be half wrong, but I think it's the right track.

jasonLaster:
if you clear the debugger’s tabs  and pause on exception prefs, you’ll see this:

1. nothing happens 
2. when you select pause on exceptions it tries to load the source of the debuggee eval, it fails to, it shows the error…
3. even if you turn off poe, it will still try loading it because we save the tab url, it is an endless loop of trying because every time rdt evals, it gets a new source, sees its the tab, and tries to load it again

![](http://g.recordit.co/qOadNPSCV1.gif)

jasonLaster:
We'll do two fixes here:

1. update how content scripts URL and contentType are set. this will let the client fetch the source.
2. we'll fix `newSource` to check to see if the source already exists. This will keep the client from re-fetching the same source text.

PR #3951, opened 2017-09-08
Ignore new identical sources

jasonLaster:
Associated Issue: #3899 

### Summary of Changes

Extension evals all have the same source actor id. This causes the debugger to try to add the same source every time react devtools evals (once a second). 

We should bail early when we see the same source again, because the server treats it as the same source.

Lets not land this until we have tests for this behavior


jasonLaster:
@bomsy 👍

PR #3952, opened 2017-09-08
Replaced use of defer with new Promise, see https://bugzilla.mozilla.…

Marcool04:
I started to work on bugzilla bug [1283869](https://bugzilla.mozillaorg/show_bug.cgi?id=1283869), but I was told that the debugger is undergoing a major change, and that the code contained in mozilla-central under devtools/client/debugger was not really the code to change. Having cloned this repo I found only one instance of defer being used, so thought I might as well make the change.
Let me know if I managed to muck anything up simply trying to make such a simple modification, I am bright and shiny new to the project :)
Best regards,
Mark.

### Summary of Changes

* removed `import defer from "../../defer"` statement
* replaced the `const deferred = defer(); ... return deferred;` construct by `return new Promise((resolve, reject) => {...});`


Marcool04:
Oh I thought the defer.js file would be coming from someplace outside of the debugger code tree. In mozilla-central it's in devtools/shared/ ... ?

jasonLaster:
we added it here too. mind pushing another commit w/ it out?

Marcool04:
Done :)

PR #3957, opened 2017-09-09
look for mochitest intermittents

jasonLaster:
this runs all the mochitests about 100 times to see if we get any test failures


jasonLaster:
this ran 55 iterations of mochitests and they all passed

Issue #3944, opened 2017-09-08
[Preview] Missing tooltip for falsy values


juliandescottes:
Try to run the following code:
```javascript
(function () {
  let falsyString = "";
  let falsyBool = false;
  let falsyNumber = 0;

  let truthyString = " ";
  let truthyBool = true;
  let truthyNumber = 1;

  debugger;
})();
```

When paused on debugger, try to hover on each of the variables available in scope. The tooltip is not displayed for any of the falsy variables.

![falsy_tooltip](https://user-images.githubusercontent.com/1141550/30216051-0f84d34a-94b2-11e7-9c7b-f6c326d0913a.gif)

jasonLaster:
Here is a [site] where you can try it out

[site]: https://fabulous-umbrella.glitch.me/

nyrosmith:
I tried to fix this and I nailed it to https://github.com/devtools-html/debugger.html/blob/09438b25f682a875b4d41170f9161efedd113730/src/actions/ast.js#L150

But then I really got stuck... Maybe this is helpful for someone more experienced on the codebase

jasonLaster:
what did you see there? Was there a problem w/ the `result` being falsey?

try 

```js
        if (result === undefined) {
          return;
        }
```

nyrosmith:
Oh stupid me. So obvious...

juliandescottes:
The issue is back.

jasonLaster:
there's a new issue for it

jasonLaster:
https://github.com/devtools-html/debugger.html/issues/5279

PR #3960, opened 2017-09-09
Fix preview for falsy values

nyrosmith:
Associated Issue: #3944 

falsypreview

### DIFF
--- a/src/actions/ast.js
+++ b/src/actions/ast.js
@@ -151,7 +151,7 @@ export function setPreview(
           frameId: selectedFrame.id
         });
 
-        if (!result) {
+        if (result === undefined) {
           return;
         }
 


jasonLaster:
Could you check with null, undefined and 0?

We have a glitch template in the issue template PR :)

nyrosmith:
I checked it. You can see it in the GIF.

nyrosmith:
Just realized that you don't see undefined case but I checked that too

jasonLaster:
Looks like we should show undefined as well: 

Here's a quick example
https://plucky-mustang.glitch.me/

chrome shows it and i believe julian had a case where he expected it too


I think our rule of thumb should just be show all values for result. So lets drop this IF statement entirely :)

nyrosmith:
Undefined is already handled
falsypreview2

Issue #3819, opened 2017-08-30
Switch feature flags to prefs


jasonLaster:
We recently started using the [prefs] system for feature flagging. It has several advantages over the `devtools-config` system:

1. prefs can be changed in firefox in `about:config`
2. prefs are saved in local storage and the prefs system
3. devtools config is in a JSON that's hard to sync.

In this PR, it'd be nice to

1. switch `isEnabled("foo")` to `features.foo`
2. add the feature to `utils/prefs` 
3. add the feature to `assets/panel/prefs`

In a later PR, we can add a UI for seeing the features in the launchpad.

Here is an example of the new [style](https://github.com/devtools-html/debugger.html/blob/master/src/actions/pause.js#L247-L250)

[prefs]: https://github.com/devtools-html/debugger.html/blob/master/docs/local-development.md#prefs

And here is a sketch of one feature ported:

```diff

diff --git a/assets/panel/prefs.js b/assets/panel/prefs.js
index a63bc03..b558c99 100644
--- a/assets/panel/prefs.js
+++ b/assets/panel/prefs.js
@@ -44,4 +44,5 @@ pref("devtools.debugger.file-search-case-sensitive", false);
 pref("devtools.debugger.file-search-whole-word", false);
 pref("devtools.debugger.file-search-regex-match", false);
 pref("devtools.debugger.features.async-stepping", false)
+pref("devtools.debugger.features.wasm", true)
 pref("devtools.debugger.project-text-search-enabled", true);
diff --git a/src/client/firefox.js b/src/client/firefox.js
index 3bd78ff..4d369cc 100644
--- a/src/client/firefox.js
+++ b/src/client/firefox.js
@@ -2,7 +2,7 @@

 import { setupCommands, clientCommands } from "./firefox/commands";
 import { setupEvents, clientEvents } from "./firefox/events";
-import { isEnabled } from "devtools-config";
+import { features } from "../utils/prefs";

 export async function onConnect(connection: any, actions: Object) {
   const {
@@ -14,7 +14,7 @@ export async function onConnect(connection: any, actions: Object) {
   }

   const supportsWasm =
-    isEnabled("wasm") && !!debuggerClient.mainRoot.traits.wasmBinarySource;
+    features.wasm && !!debuggerClient.mainRoot.traits.wasmBinarySource;

   setupCommands({
     threadClient,
diff --git a/src/utils/prefs.js b/src/utils/prefs.js
index cb9e3e4..312ed98 100644
[...]

PR #3961, opened 2017-09-09
[WIP] convert from featureFlag to prefs system

nyrosmith:
Associated Issue: #3819


nyrosmith:
I create this PR at a very early stage because I have a few questions 

- what is `configs/firefox-panel.json` for? I am unsure if I can remove the feature flags there too?
- what is the difference between`assets/panel/prefs.js` & `src/utils/prefs.js` Is this duplication of prefs really needed?

Issue #3893, opened 2017-09-05
[WelcomeBox] make shortcuts clickable


jasonLaster:
We recently updated our welcome box to include additional shortcuts. It would be neat if we could also make them links as well, so when you click on them it takes you to where you'd go if you fired the shortcut.

Note, we do this with the left sidebar link on the top.

we could do it with something like this... note, the diff does not include the correct fields for the search. see reducers/ui.js for the correct values
```diff
diff --git a/src/components/WelcomeBox.js b/src/components/WelcomeBox.js
index 47d3fb7..1dc0055 100644
--- a/src/components/WelcomeBox.js
+++ b/src/components/WelcomeBox.js
@@ -58,14 +55,16 @@ class WelcomeBox extends Component {
       <div className="welcomebox">
         <div className="alignlabel">
           <div className="shortcutKeys">
-            <p>{searchSourcesShortcut}</p>
-            <p>{searchProjectShortcut}</p>
-            <p>{searchFunctionsShortcut}</p>
+            <p onClick={setActiveSearch("sources")}>{searchSourcesShortcut}</p>
+            <p onClick={setActiveSearch("project")}>{searchProjectShortcut}</p>
+            <p onClick={setActiveSearch("function")}>
+              {searchFunctionsShortcut}
+            </p>
           </div>
           <div className="shortcutFunction">
-            <p>{searchSourcesLabel}</p>
-            <p>{searchProjectLabel}</p>
-            <p>{searchFunctionLabel}</p>
+            <p onClick={setActiveSearch(item)}>{searchSourcesLabel}</p>
+            <p onClick={setActiveSearch(item)}>{searchProjectLabel}</p>
+            <p onClick={setActiveSearch(item)}>{searchFunctionLabel}</p>
           </div>
           {this.renderToggleButton()}
         </div>
(END)
```

jnachtigall:
I guess because of a11y (semantics) this would also mean to use `<button>` instead of `<p>`. (https://github.com/evcohen/eslint-plugin-jsx-a11y is pretty cool).

Well, actually I come here because I noticed how "Memory" tab:
![image](https://user-images.githubusercontent.com/3307404/30162134-d46f1218-93d3-11e7-9e66-8c2d0d82f81b.png)
and "Performance" tab already have such a "action" buttons:
![image](https://user-images.githubusercontent.com/3307404/30162161-ece81682-93d3-11e7-9f8a-6aad995bf023.png)
[...]

PR #3963, opened 2017-09-10
WelcomeBox: make shortcuts clickable

nyrosmith:
Associated Issue: #3893

### DIFF
--- a/src/components/WelcomeBox.js
+++ b/src/components/WelcomeBox.js
@@ -53,18 +53,20 @@ class WelcomeBox extends Component {
     const searchFunctionLabel = L10N.getStr("welcome.searchFunction").substring(
       2
     );
+    const { setActiveSearch } = this.props;
 
+    // debugger
     return (
       <div className="welcomebox">
         <div className="alignlabel">
           <div className="shortcutKeys">
-            <p>{searchSourcesShortcut}</p>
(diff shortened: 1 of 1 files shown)


src/components/WelcomeBox.js
```diff
@@ -53,18 +53,19 @@ class WelcomeBox extends Component {
...
 
     return (
       <div className="welcomebox">
         <div className="alignlabel">
           <div className="shortcutKeys">
-            <p>{searchSourcesShortcut}</p>
-            <p>{searchProjectShortcut}</p>
+            <p onClick={setActiveSearch.bind(null, "source")}>{searchSourcesShortcut}</p>
+            <p onClick={setActiveSearch.bind(null, "project")}>{searchProjectShortcut}</p>
             <p>{searchFunctionsShortcut}</p>
```
nyrosmith: I really think we should remove this shortcut.
The other shortcuts are "actionable" at this point but search for functions can not be called from WelcomeBox because it  is only applicable on opened source.
Any thoughts @jasonLaster  @codehag

wldcordeiro: I agree! The symbol modal hint doesn't make sense here.

nyrosmith:
Done!

Issue #3824, opened 2017-08-30
[Preview] ignore `=` token


jasonLaster:
STR:

1. pause in todo-view.js line 81
2. mouse over the `=` on the line

Expected: nothing should happen
Actual: see the whole expression highlighted

The fix is probably to blacklist `=`, but it would be nice to poke around and see what else we should exclude.
```diff
diff --git a/src/utils/editor/expression.js b/src/utils/editor/expression.js
index 5532848..1defda4 100644
--- a/src/utils/editor/expression.js
+++ b/src/utils/editor/expression.js
@@ -36,7 +36,7 @@ export function updatePreview(
     }
   }

-  const invalidToken = tokenText === "" || tokenText.match(/[(){},.;\s]/);
+  const invalidToken = tokenText === "" || tokenText.match(/[(){},.;=\s]/);
   const invalidTarget =
     (target.parentElement &&
       !target.parentElement.closest(".CodeMirror-line")) ||
```

So, what's the harm of highlighting the whole expression? Well, glad you asked:

1. it might do the entire assignment which is mutating state. Yuck
2. it's distracting (it's going to return undefined) and throw a popup in your face

So, where's the other logic. Preview is hard yo:

1. look in Editor/Preview/index for the hover event
2. look in Actions/ast for setPreview

This code is awesome. It's really neat to see how it works :dancer:

![Screen Shot 2017-08-30 at 3.45.53 PM.png](https://shipusercontent.com/ef4eea610ce89462b356c554e0718cd1/Screen%20Shot%202017-08-30%20at%203.45.53%20PM.png)

Fischer-L:
@jasonLaster 
Also found the similar tokens as "<", ">", "+", "-", "*", "/" produce the same issue. 
![3824](https://user-images.githubusercontent.com/5627487/30240709-20d75d74-95a8-11e7-914f-fa4412be33ad.gif)

Tried to filter out these tokens as well as below.
![3824_solution](https://user-images.githubusercontent.com/5627487/30240711-292c1b5e-95a8-11e7-8d42-958ec55beda7.gif)

Probably we should filter out  "<", ">", "+", "-", "*", "/" too. How do you think? If ok, i am going to submit the patch, thanks.

jasonLaster:
Yes! Go for it. We have an issue for = but they're all valid.

jasonLaster:
The issue for = has some notes

Fischer-L:
@jasonLaster 
[...]

PR #3967, opened 2017-09-10
Disallow pipes

jasonLaster:
Associated Issue: #3824

Found another case


nyrosmith:
Maybe % & : also fit here

jasonLaster:
Yes, i think they do...

jasonLaster:
https://brash-success.glitch.me/

Issue #3399, opened 2017-07-20
[Breakpoints] [SourceMap] don’t slide w/ original sources


jasonLaster:
The old debugger does not slide when you add a bp to `setDebugLine` in Editor/index.js. 

#### old
![](http://g.recordit.co/NBw2YVLpRO.gif)

#### new
![](http://g.recordit.co/jhByGJaIqZ.gif)

PR #4005, opened 2017-09-13
Support symmetric BPs

jasonLaster:
Associated Issue: #3399

### Summary of Changes

This pulls in the update for symmetric bps in the service worker. It also adds some of the undefined logic for the debugger now that it's not in the worker.

### Test Plan

test in perf.html

### DIFF
--- a/flow-typed/debugger-html.js
+++ b/flow-typed/debugger-html.js
@@ -222,6 +222,32 @@ declare module "debugger-html" {
   };
 
   /**
+   * SourceScope
+   * @memberof types
+   * @static
+   */
+  declare type SourceScope = {
+    type: string,
+    start: Location,
+    end: Location,
+    bindings: {
+      [name: string]: Location[]
+    }
+  };
+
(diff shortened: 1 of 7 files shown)


flow-typed/debugger-html.js
```diff
@@ -222,6 +222,32 @@ declare module "debugger-html" {
...
+      [name: string]: Location[]
+    }
+  };
+
+  /*
+   * MappedScopeBindings
+   * @memberof types
+   * @static
+   */
+  declare type MappedScopeBindings = {
```
tromey: These don't seem to be used.

jasonLaster: I pulled it in because it will be used by `devtools-map-bindings` shortly and seemed like a dependency of pulling in the library.

jasonLaster: Actually, now that I say that, it was just pulled in 10 minutes ago by another PR that uses it.

PR #4006, opened 2017-09-13
gracefully support bad bp cases

jasonLaster:
### Summary of Changes

We previously had better protection in the components against these bad cases. Now that our call sites will pass anything through, lets just bail early


jasonLaster:
going to optimistically merge, let me know if anyone has any objections.

yurydelendik:
Too late, but it would be probably use full to keep "errors" text as a comments, so function bailout condition is informed to the reader of the code.

Issue #3260, opened 2017-07-04
[SourceTree] collapse on layout switch


bomsy:
Resizing the debugger window (and thus switching to/from vertical layout) collapses the source tree ? (in latest nightly)

![Sourcetree reset](http://g.recordit.co/cj8LlTFsbK.gif)

bomsy:
jlast: We fixed a bug w app layout doing a render when you switched tabs bc it would be like going to vertical layout and back, We're remounting the source tree and losing component state

bomsy:
![](http://g.recordit.co/sFiyf2U9gn.gif)

bomsy:
Things we want to do here
- [ ] Create redux actions and reducers to store and retrieve the expanded state of the source tree
- [ ] On Component mount get the current expanded state from redux

A diff coming up...

PR #4010, opened 2017-09-14
Sourcetree redux

bomsy:
Associated Issue: #3260

### Summary of Changes

* Add source-tree reducer
* Add source-tree actions
* On mount of Managed Tree if there is an expanded cache we load that

Example test plan:

- [x] Load debugger
- [x] Load debuggee `http://firefox-dev.tools/debugger-examples/examples/todomvc/`
- [x] Click the source tree
- [x] Adjust to vertical or horizontal layout


jasonLaster:
This looks really good! 👍

src/reducers/tests/source-tree.spec.js
```diff
@@ -0,0 +1,19 @@
...
+    const state = InitialState();
+    const expanded = new Set(["foo", "bar"]);
+    const updatedState = update(state, {
+      type: "SET_EXPANDED_STATE",
+      expanded
+    });
+
+    expect(updatedState.expanded).toBe(expanded);
+  });
+});
```
jasonLaster: i think this test is redundant. we get the same coverage from the action test.

bomsy: lol .. yeah ... tot so too. never written a reducer test, got excited with the tests :)

bomsy: will pull it out.

jasonLaster:
it looks like we dispatch two actions per toggle:



Is it possible to avoid?

bomsy:
@jasonLaster i can't seem to replicate this. How are you logging the expanded state?

bomsy:
i see this

PR #4023, opened 2017-09-14
don't load if we're not paused

jasonLaster:
### Summary of Changes

When I was making the 9-13 bundle i noticed we were getting an intermittent because we didn't check to see if we were fetching properties when we weren't paused.


jasonLaster:
closing because we should evaluate this option a bit more

PR #4045, opened 2017-09-15
synchronize prefs file in GitHub with m-c

juliandescottes:
Sync prefs file with mc

### DIFF
--- a/assets/panel/prefs.js
+++ b/assets/panel/prefs.js
@@ -1,16 +1,18 @@
 /* Any copyright is dedicated to the Public Domain.
  * http://creativecommons.org/publicdomain/zero/1.0/ */
 
-#ifdef RELEASE_OR_BETA
-pref("devtools.debugger.new-debugger-frontend", false);
-#else
-pref("devtools.debugger.new-debugger-frontend", true);
+// copy-assets-start
+// This is just a workaround for the old docker image which expects a file with
+// preprocessing.
+#if True
 #endif
+// copy-assets-end
+
+pref("devtools.debugger.new-debugger-frontend", true);
 
 // Enable the Debugger
 pref("devtools.debugger.enabled", true);
 pref("devtools.debugger.chrome-debugging-host", "localhost");
-pref("devtools.debugger.chrome-debugging-port", 6080);
 pref("devtools.debugger.chrome-debugging-websocket", false);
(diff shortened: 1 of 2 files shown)


assets/panel/prefs.js
```diff
@@ -1,16 +1,11 @@
 /* Any copyright is dedicated to the Public Domain.
  * http://creativecommons.org/publicdomain/zero/1.0/ */
 
-#ifdef RELEASE_OR_BETA
-pref("devtools.debugger.new-debugger-frontend", false);
-#else
 pref("devtools.debugger.new-debugger-frontend", true);
-#endif
```
jasonLaster: i'm pretty sure this will fail in mochitests

jasonLaster: but lets see...

juliandescottes: What do you mean? Do you run old the debugger mochitests on CI?

juliandescottes: I see
> ./mach mochitest --subsuite devtools --setenv MOZ_HEADLESS=1 devtools/client/debugger/new/test/mochitest/

I don't think this pref should affect any of them :/

jasonLaster: nope, it's a build issue... it needs to have some #if block or whatever to be considered 

we need the moz.build to be this:
```
# -*- Mode: python; indent-tabs-mode: nil; tab-width: 40 -*-
# vim: set filetype=python:
# This Source Code Form is subject to the terms of the Mozilla Public
# License, v. 2.0. If a copy of the MPL was not distributed with this
# file, You can obtain one at http://mozilla.org/MPL/2.0/.

JS_PREFERENCE_PP_FILES += [
    'devtools.js'
]

JS_PREFERENCE_FILES += [
    'debugger.js',
]

```

but it's 

```
# -*- Mode: python; indent-tabs-mode: nil; tab-width: 40 -*-
# vim: set filetype=python:
# This Source Code Form is subject to the terms of the Mozilla Public
# License, v. 2.0. If a copy of the MPL was not distributed with this
# file, You can obtain one at http://mozilla.org/MPL/2.0/.

JS_PREFERENCE_PP_FILES += [
    'devtools.js'
    'debugger.js'
]

```

we could replace the moz.build with our own like we're replacing other files as well....

Issue #3986, opened 2017-09-12
[Mochitests] Update the docker CI image


jasonLaster:
We are using an old docker image for our mochitests, this is creating a couple of problems:

1. we can't use server changes
2. we can't use the new codemirror on m-c
3. some tests can only run in m-c or vice-versa
4. some tests fail in m-c, but not on CI

Update the image?
* it is a pain to update it. We have tried and could not do all the steps
* it will always take awhile

Use a janitor or taskcluster image? could work

Run the tests on taskcluster? seems promising

jasonLaster:
Alex mentioned we could use his build and clone scripts;

https://github.com/ochameau/ff-dt/blob/618091bac01d6ff280b9cc4e3221e3b82083d972/scripts/clone-gecko.sh

jasonLaster:
yay! we did it :)

PR #4052, opened 2017-09-16
add travis

jasonLaster:
Associated Issue #3986 

### Summary

This switches our mochitests to travis.

**advantages**

* tests run in parallel 
* we don't use docker
* mc is up to date


jasonLaster:
merging to test travis, we'll still need to do some work... in a subsequent PR

PR #4054, opened 2017-09-16
Update travis

jasonLaster:
testing the travis setup...


jasonLaster:
Thanks @Fischer-L, I'm adding @darkwing as a reviewer

Fischer-L:
@jasonLaster My PR is #4057 so I guess #4057 is what you mean right? Thanks

jasonLaster:
oops :)

.travis.yml
```diff
@@ -1,12 +1,15 @@
...
-    - node_modules
+  - pip
+  - yarn
 
 env:
-  - DISPLAY=':99.0'
+  global:
+    - DISPLAY=':99.0'
+    - YARN_VERSION='0.24.5'
+    - MC_COMMIT='abd295766dfe'
```
ochameau: Are you now able to bump this m-c changeset easily/frequently?

bin/ci/clone-gecko.sh
```diff
@@ -1,12 +1,9 @@
...
-  hg strip --no-backup 'roots(outgoing())'
-  hg pull -u
-  cd ..
-fi
+rm -rf firefox/
+hg clone https://hg.mozilla.org/mozilla-unified/ firefox
+
+cd firefox
+hg co $MC_COMMIT
+cd ..
```
ochameau: It could be great to use `bin/prepare-mochitests-dev`
so that CI and contributors go through the same path.

But given this change, it looks like you had issue with keeping m-c clone around between two travis runs?

.travis.yml
jasonLaster: Yep. It's fixed so that we always get a good artifact build though

bin/ci/clone-gecko.sh
jasonLaster: yeah, caching was an issue.

Issue #4031, opened 2017-09-15
[Outline] move the tabs to the top


jasonLaster:
The new photon designs call for moving the sources/outline tabs to the top to be consistent.

We hopefully can borrow the html/css from the editor tabs, while re-using the JS...

### Light

![Screen Shot 2017-09-15 at 9.46.30 AM.png](https://shipusercontent.com/b66e7de0d4ab94f00f5d8da906f79963/Screen%20Shot%202017-09-15%20at%209.46.30%20AM.png)

### Dark

![Screen Shot 2017-09-15 at 9.46.37 AM.png](https://shipusercontent.com/ecda22ce18d2d5e0b0b9a869b57002ea/Screen%20Shot%202017-09-15%20at%209.46.37%20AM.png)

nyrosmith:
I am on it.

PR #4056, opened 2017-09-17
Move PrimaryPane tabs to the top

nyrosmith:
Associated Issue: #4031 

### Summary of Changes

* Moved PrimaryPane tabs to the top
* Refactored/renamed functions to match their purpose
* Simplified the stucture of PrimaryPane
* removed isHidden prop fromSourceTree & Outline (hiding via css is an antipattern. It should either get rendered or not)
* Don't show "Ctrl P to search" when on the Outline tab
* changed css to make the Tabs height same as Editor header height
* Fix PrimaryPanes display issue when there are no sources and Outline tab is active (see attached screenshot)

unbenannt

### DIFF
--- a/src/components/PrimaryPanes/Sources.css
+++ b/src/components/PrimaryPanes/Sources.css
@@ -79,6 +79,7 @@
 
 .sources-panel .source-footer {
   position: relative;
+  height: 29px;
 }
 
 .sources-panel .outline {
--- a/src/components/PrimaryPanes/Outline.js
+++ b/src/components/PrimaryPanes/Outline.js
@@ -3,7 +3,6 @@
 import React, { Component } from "react";
 import { bindActionCreators } from "redux";
 import { connect } from "react-redux";
-import classnames from "classnames";
 import actions from "../../actions";
 import { getSelectedSource, getSymbols } from "../../selectors";
 import "./Outline.css";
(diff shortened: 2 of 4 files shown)


src/components/PrimaryPanes/Sources.css
```diff
@@ -79,6 +79,7 @@
 
 .sources-panel .source-footer {
```
nyrosmith: I think this should be renamed because it has nothing to do with footer
I'd also like to move this into a new PrimaryPane.css file as this is not to related to Sources component at all.
Any thoughts about this?

jasonLaster: agreed.

jasonLaster:
Could we get the tabs to look the same as the Editor tabs?

nyrosmith:
For me they look really good as they are but if course I can try copy the editor tabs style

jasonLaster:
@nyrosmith yeah, i'm not sure if the Editor style tabs will be the final look, but I think it's a good next step. It's nice to have a consistent Debugger style tabs, which looks good with the devtools tabs on the top.

darkwing:
A few thoughts @nyrosmith:

1.  The "selected" tab background color for the Sources/Outline tabs are opposite in color to the source tabs:

selectedstate

We should make those colors consistent.

2.  I'm seeing a scrollbar whenever I open a tree item:

scrollbars

3.  The tabs and their content are separated by the [Search] block which I find a bit confusing; IMO we should restore that [Search] blocks to the top.

jasonLaster:
closing in favor of #4079

PR #4060, opened 2017-09-18
sync changes w/ mc

jasonLaster:
now that we have an uptodate version of MC, we can sync prefs and browser.ini w/ mc.

### DIFF
--- a/src/test/mochitest/browser.ini
+++ b/src/test/mochitest/browser.ini
@@ -50,14 +50,14 @@ support-files =
 [browser_dbg-breakpoints.js]
 [browser_dbg-breakpoints-toggle.js]
 [browser_dbg-breakpoints-reloading.js]
+skip-if = true # Bug 1383576
 [browser_dbg-breakpoints-cond.js]
 [browser_dbg-call-stack.js]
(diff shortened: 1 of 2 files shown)


src/test/mochitest/browser.ini
```diff
@@ -79,6 +80,7 @@ skip-if = os == "win" # Bug 1393121
 [browser_dbg-search-symbols.js]
 skip-if = os == "win" # Bug 1393121
 [browser_dbg-search-project.js]
+skip-if = true # Bug 1393121, 1393299
```
jasonLaster: we should also get serious about removing these skips. e.g. I'm pretty sure we fixed the intermittent w/ searching, so we can remove the search skips after the 57 release

src/test/mochitest/browser.ini
```diff
@@ -68,6 +68,7 @@ skip-if = true
 [browser_dbg_keyboard-shortcuts.js]
 skip-if = os == "linux" # bug 1351952
 [browser_dbg-pause-exceptions.js]
+skip-if = true # Bug 1393121
```
jasonLaster: i'm working on a fix for this as well.

Issue #3597, opened 2017-08-08
[Expressions] errors cause the console to stop evaluating


jasonLaster:
#### STR

1. Go to http://www.example.org/
2. Open the debugger
3. Set it to "Pause on uncaught exceptions"
4. Add a new watched expression: "foo.bar" (without quotes)
5. The debugger pauses: "ReferenceError: foo is not defined"
6. Click ▷ (Resume)
7. Switch to the webconsole devtool
8. Enter something to the console

Expected: the console displays the returned value
Result: the returned value is not displayed

From: https://bugzilla.mozilla.org/show_bug.cgi?id=1388096

PR #4072, opened 2017-09-18
Pausing due to an errored watch expression

jasonLaster:
Associated Issue: #3597

### Summary of Changes

* stops evaluating when we're paused at an expression that threw
* bails early when `resumed` is called but we're not paused. This prevents unnecessary evals when a bp is added.

### Test Plan

* a mochitest that shows that watch expressions still work when we we're paused due to an exception in an expression.

### DIFF
--- a/src/actions/expressions.js
+++ b/src/actions/expressions.js
@@ -38,7 +38,8 @@ export function addExpression(input: string) {
 
     const selectedFrame = getSelectedFrame(getState());
     const selectedFrameId = selectedFrame ? selectedFrame.id : null;
-    dispatch(evaluateExpression({ input }, selectedFrameId));
+    const newExpression = getExpression(getState(), input);
+    dispatch(evaluateExpression(newExpression, selectedFrameId));
   };
 }
 
@@ -95,7 +96,7 @@ export function evaluateExpressions(frameId: frameIdType) {
   };
 }
 
-function evaluateExpression(expression, frameId: frameIdType) {
+function evaluateExpression(expression: Expression, frameId: frameIdType) {
   return function({ dispatch, getState, client }: ThunkArgs) {
     if (!expression.input) {
       console.warn("Expressions should not be empty");
--- a/src/actions/pause.js
+++ b/src/actions/pause.js
@@ -63,6 +63,7 @@ export function continueToHere(line: number) {
         sourceId: source.id
       })
     );
+
     dispatch(command("resume"));
   };
 }
@@ -97,6 +98,8 @@ export function paused(pauseInfo: Pause) {
       dispatch(removeBreakpoint(hiddenBreakpointLocation));
     }
 
+    // NOTE: We don't want to re-evaluate watch expressions
+    // if we're paused due to an excpression exception #3597
     if (!hasWatchExpressionErrored(getState())) {
       dispatch(evaluateExpressions(frame.id));
     }
--- a/src/actions/pause.js
+++ b/src/actions/pause.js
@@ -7,7 +7,9 @@ import {
   getPause,
   getLoadedObject,
   isStepping,
-  getSelectedSource
+  isPaused,
+  getSelectedSource,
+  hasWatchExpressionErrored
 } from "../selectors";
 import { updateFrameLocations, getPausedPosition } from "../utils/pause";
 import { evaluateExpressions } from "./expressions";
@@ -35,6 +37,10 @@ type CommandType = string;
  */
 export function resumed() {
   return ({ dispatch, client, getState }: ThunkArgs) => {
+    if (!isPaused(getState())) {
+      return;
+    }
+
     dispatch({
       type: "RESUME",
       value: undefined
@@ -91,7 +97,9 @@ export function paused(pauseInfo: Pause) {
       dispatch(removeBreakpoint(hiddenBreakpointLocation));
     }
 
-    dispatch(evaluateExpressions(frame.id));
+    if (!hasWatchExpressionErrored(getState())) {
+      dispatch(evaluateExpressions(frame.id));
+    }
 
     dispatch(
       selectSource(frame.location.sourceId, { line: frame.location.line })
--- a/src/actions/tests/pause.spec.js
+++ b/src/actions/tests/pause.spec.js
@@ -1,12 +1,22 @@
-import { actions, selectors, createStore } from "../../utils/test-head";
+import {
+  actions,
+  selectors,
+  createStore,
+  getHistory
+} from "../../utils/test-head";
 
-const { isStepping } = selectors;
+const { isStepping, hasWatchExpressionErrored } = selectors;
 
 let stepInResolve = null;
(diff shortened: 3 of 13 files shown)


src/utils/test-head.js
```diff
@@ -29,6 +30,19 @@ function createStore(client: any, initialState: any = {}, sourceMapsMock: any) {
...
+    return logs;
+  }
+
+  return logs.filter(log => log.type === query);
+}
+
+function clearHistory() {
+  logs = [];
+}
+
```
jasonLaster: we're now recording the actions so we can have assertions based on the history

src/utils/redux/middleware/history.js
```diff
@@ -13,12 +13,8 @@ export const history = (log: Object[] = []) => ({
...
   getState
 }: ThunkArgs) => {
-  if (isDevelopment()) {
-    console.warn(
-      "Using history middleware stores all actions in state for " +
-        "testing and devtools is not currently running in test " +
-        "mode. Be sure this is intentional."
-    );
+  if (!isDevelopment()) {
+    return;
```
jasonLaster: lets flip it so we know we can't log history in production

src/types.js
```diff
@@ -8,8 +8,7 @@ export type SearchModifiers = {
 
 export type Expression = {
   value: Object,
-  input: string,
-  visible: boolean
+  input: string
```
jasonLaster: visible is no longer possible

src/actions/tests/pause.spec.js
```diff
@@ -23,27 +33,68 @@ const mockThreadClient = {
...
+      await dispatch(actions.resumed());
+
+      expect(getHistory("RESUME").length).toEqual(1);
+    });
+
+    it("resuming when not paused", async () => {
+      const { dispatch, getState } = createStore(mockThreadClient);
+      await dispatch(actions.resumed());
+      expect(getHistory("RESUME").length).toEqual(0);
+    });
```
jasonLaster: i moved the tests around to group by `paused` action. I also added a couple for `resumed` to make sure we bail early when not paused

src/actions/expressions.js
```diff
@@ -38,7 +38,8 @@ export function addExpression(input: string) {
 
     const selectedFrame = getSelectedFrame(getState());
     const selectedFrameId = selectedFrame ? selectedFrame.id : null;
-    dispatch(evaluateExpression({ input }, selectedFrameId));
+    const newExpression = getExpression(getState(), input);
+    dispatch(evaluateExpression(newExpression, selectedFrameId));
```
jasonLaster: when i made the flow type stricter I realized i wasn't passing a real expression into evaluate :)

nchevobbe: 👍

src/actions/pause.js
```diff
@@ -91,7 +98,9 @@ export function paused(pauseInfo: Pause) {
       dispatch(removeBreakpoint(hiddenBreakpointLocation));
     }
 
-    dispatch(evaluateExpressions(frame.id));
+    if (!hasWatchExpressionErrored(getState())) {
```
nchevobbe: Could you add a comment pointing to the original issue so someone reading this can have some context ?

src/actions/tests/pause.spec.js
```diff
@@ -23,27 +33,68 @@ const mockThreadClient = {
...
+  describe("paused", () => {
+    it("should detect when we're paused due to an expression exception", async () => {
+      const { dispatch, getState } = createStore(mockThreadClient);
+      const mockPauseInfo = {
+        frames: [{ id: 1, scope: [], location: { sourceId: "foo1", line: 4 } }],
+        loadedObjects: [],
+        why: { type: "exception" }
+      };
+
+      dispatch(actions.addExpression("foo"));
```
nchevobbe: could you turn "foo" into "foo.bar" so we understand it is supposed to throw ? 
Also, do we need this dispatch and the other one ? It seem like it's `actions.paused` that really has value here.

src/actions/tests/pause.spec.js
```diff
@@ -23,27 +33,68 @@ const mockThreadClient = {
...
+      await stepInResolve();
+      expect(isStepping(getState())).toBeFalsy();
+    });
   });
 
-  it("should not evaluate expression while stepping", async () => {
-    const client = { evaluate: jest.fn() };
-    const { dispatch } = createStore(client);
+  describe("resumed", () => {
+    it("should not evaluate expression while stepping", async () => {
```
nchevobbe: are we sure we ant this ? I may want to step through a function to see how a variable (in the watch expression) evolves due to my code

src/actions/tests/pause.spec.js
```diff
@@ -23,27 +33,68 @@ const mockThreadClient = {
...
-    expect(client.evaluate.mock.calls.length).toEqual(0);
+      dispatch(actions.stepIn());
+      await dispatch(actions.resumed());
+      expect(client.evaluate.mock.calls.length).toEqual(0);
+    });
+
+    it("resuming", async () => {
+      const { dispatch, getState } = createStore(mockThreadClient);
+      const mockPauseInfo = {
+        frames: [{ id: 1, scope: [], location: { sourceId: "foo1", line: 4 } }],
```
nchevobbe: this `frames` stub is used in multiple tests. maybe you could put it somewhere in a function ? So if/when the structure of a frame changes you don't miss changing the mock somewhere

src/test/mochitest/browser_dbg-expressions-error.js
```diff
@@ -0,0 +1,69 @@
...
+
+  const paused = waitForPaused(dbg);
+  addExpression(dbg, "foo.bar");
+
+  await paused;
+  ok(dbg.selectors.hasWatchExpressionErrored(dbg.getState()));
+
+  toggleExpression(dbg, 1);
+  await waitForDispatch(dbg, "LOAD_OBJECT_PROPERTIES");
+  is(findAllElements(dbg, "expressionNodes").length, 17);
```
nchevobbe: maybe we can resume and see if everything was fine. I think the exception in the POE was thrown when resuming from a pause on exception.

src/utils/redux/middleware/history.js
nchevobbe: 👏

src/test/mochitest/browser_dbg-expressions-error.js
jasonLaster: yeah, we can do that too. interestingly, the appropriate behavior (i think) is to re-pause in the expression, but atleast the console should work

src/actions/tests/pause.spec.js
jasonLaster: oh yeah... we're not talking about *that*. 

When we step, it looks like:

* stepIn (start)
* resume (while stepping)
* stepin (done)

we don't want to evaluate while we're temporarily resumed :)

Issue #4021, opened 2017-09-14
[OutlineView] get the closest class for functions and properties


jasonLaster:
It would be awesome to build a smarter outline view that knew when functions were inside of a class.

![Screen Shot 2017-09-14 at 3.14.10 PM.png](https://shipusercontent.com/faec4f7495ebe6488acfa0377ffb78bd/Screen%20Shot%202017-09-14%20at%203.14.10%20PM.png)

#### code sketch

```diff
diff --git a/src/utils/parser/getSymbols.js b/src/utils/parser/getSymbols.js
index 2c6b45d..517735a 100644
--- a/src/utils/parser/getSymbols.js
+++ b/src/utils/parser/getSymbols.js
@@ -70,6 +70,12 @@ function getComments(ast) {
   }));
 }

+// check to see if the parent or ancestor is a class...
+// if not return null
+function getClassName(path) {
+  path.getClosest(_p => ...)
+}
+
 function extractSymbols(source: Source) {
   const functions = [];
   const variables = [];
@@ -87,6 +93,7 @@ function extractSymbols(source: Source) {
       if (isFunction(path)) {
         functions.push({
           name: getFunctionName(path),
+          class: getClassName(path),
           location: path.node.loc,
           parameterNames: getFunctionParameterNames(path),
           identifier: path.node.id
diff --git a/src/utils/parser/tests/fixtures/component.js b/src/utils/parser/tests/fixtures/component.js
new file mode 100644
index 0000000..bbf3ba6
--- /dev/null
+++ b/src/utils/parser/tests/fixtures/component.js
@@ -0,0 +1,18 @@
+import React, {Component} from "react";
+
+class Punny extends Component {
+
+  constructor(props) {
+    this.onClick = this.onClick.bind(this);
+    super();
+  }
+
+  componentDidMount() {}
+
+  onClick() {}
+
+  renderMe() {
+    return <div onClick={this.onClick}></div>
+  }
+  render() {}
+}
diff --git a/src/utils/parser/tests/getSymbols.spec.js b/src/utils/parser/tests/getSymbols.spec.js
index cdfd18f..7a0b01a 100644
--- a/src/utils/parser/tests/getSymbols.spec.js
+++ b/src/utils/parser/tests/getSymbols.spec.js
@@ -29,6 +29,11 @@ describe("Parser.getSymbols", () => {
     expect(symbols).toMatchSnapshot();
   });

+  it("component", () => {
[...]

PR #4077, opened 2017-09-18
Parse the closest class for functions and properties

sarahlim:
Associated Issue: #4021

### Summary of Changes

- parsing function symbols now includes the name of the closest ancestor class, or `null` if there is no containing class
- can be used like this:

```diff
diff --git a/src/components/PrimaryPanes/Outline.js b/src/components/PrimaryPanes/Outline.js
index 5c0fac6..97359e6 100644
--- a/src/components/PrimaryPanes/Outline.js
+++ b/src/components/PrimaryPanes/Outline.js
@@ -45,7 +45,7 @@ export class Outline extends Component {
   }

   renderFunction(func: SymbolDeclaration) {
-    const { name, location, parameterNames } = func;
+    const { name, location, parameterNames, klass } = func;

     return (
       <li
@@ -54,6 +54,8 @@ export class Outline extends Component {
         onClick={() => this.selectItem(location)}
       >
         <PreviewFunction func={{ name, parameterNames }} />
+        {" - "}
+        {klass}
       </li>
     );
   }
```

screen shot 2017-09-18 at 12 30 54 pm

### Test Plan

- added component test fixture (should I add more edge cases?)

### DIFF
--- a/src/utils/parser/getSymbols.js
+++ b/src/utils/parser/getSymbols.js
@@ -371,7 +371,8 @@ export function formatSymbols(source: Source) {
       ? symbol.parameterNames.join(", ")
       : "";
     const expression = symbol.expression || "";
-    return `${loc} ${exprLoc} ${expression} ${symbol.name} ${params}`;
+    const klass = symbol.klass || "";
+    return `${loc} ${exprLoc} ${expression} ${symbol.name} ${params} ${klass}`;
   }
 
   return [
--- a/src/utils/parser/getSymbols.js
+++ b/src/utils/parser/getSymbols.js
@@ -13,6 +13,7 @@ let symbolDeclarations = new Map();
 export type SymbolDeclaration = {|
   name: string,
   expression?: string,
+  klass?: ?string,
   location: BabelLocation,
   expressionLocation?: BabelLocation,
   parameterNames?: string[],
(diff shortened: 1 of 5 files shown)


src/utils/parser/tests/getSymbols.spec.js
```diff
@@ -53,4 +53,9 @@ describe("Parser.getSymbols", () => {
     const symbols = formatSymbols(getSource("parseScriptTags", "html"));
     expect(symbols).toMatchSnapshot();
   });
+
+  it("component", () => {
+    const symbols = formatSymbols(getSource("component"));
+    expect(symbols).toMatchSnapshot();
+  });
```
jasonLaster: lets update `formatSymbols` to show the class name when it exists, this will get us some additional test coverage.

src/utils/parser/getSymbols.js
```diff
@@ -359,7 +371,8 @@ export function formatSymbols(source: Source) {
       ? symbol.parameterNames.join(", ")
       : "";
     const expression = symbol.expression || "";
-    return `${loc} ${exprLoc} ${expression} ${symbol.name} ${params}`;
+    const klass = symbol.klass || "";
+    return `${loc} ${exprLoc} ${expression} ${symbol.name} ${params} ${klass}`;
```
jasonLaster: hmm, looks like we're not including functions, which we probably should... perhaps we should update `formatSymbol` to map the entire dict of symbols:

```js
Object.keys(symbols).map(name => `${name}:\n ${symbols[name].map(summarize)}`
```

jasonLaster:
Thanks so much!

src/utils/parser/getSymbols.js
```diff
@@ -76,9 +76,8 @@ function getClassName(path: NodePath): ?string {
   if (!classDeclaration) {
     return null;
   }
-  if (classDeclaration) {
-    return classDeclaration.node.id.name;
-  }
+
```
sarahlim: Whoa.......not sure how this happened, thanks for the catch 😛

PR #4079, opened 2017-09-18
Fix #4031 - Move source and outline tabs to the top of the left pane

### LINKED ISSUE #4031: [Outline] move the tabs to the top

darkwing:
Putting this in quick to see how people like it.

after

### DIFF
--- a/src/components/PrimaryPanes/Sources.css
+++ b/src/components/PrimaryPanes/Sources.css
@@ -101,7 +101,8 @@
 .source-footer .tab {
   flex: 1;
   justify-content: center;
-  border: 1px solid transparent;
+  border-bottom: 1px solid transparent;
+  border-left: 1px solid transparent;
   border-bottom-left-radius: 2px;
   border-bottom-right-radius: 2px;
   display: inline-flex;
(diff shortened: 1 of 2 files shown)


jasonLaster:
:thumbsup:

darkwing:
OK, this is ready for a look.  Down the road we may want to refactor the CSS here because `class="source-footer"` is used twice: once for the editor footer and also for this sources/outline container.  That seems like it could cause trouble down the road...maybe.

src/components/PrimaryPanes/index.js
```diff
@@ -80,7 +80,7 @@ class PrimaryPanes extends Component {
     ];
   }
 
-  renderFooter() {
+  renderTabs() {
     return <div className="source-footer">{this.renderOutlineTabs()}</div>;
```
jasonLaster: perhaps we want to duplicate the CSS now and move it over to PrimaryPane.css

src/components/PrimaryPanes/Sources.css
```diff
@@ -114,17 +115,26 @@
   cursor: default;
 }
 
+.source-footer .tab:first-child {
+  border-left: 0;
+}
```
jasonLaster: does this get us the border radii he Editor tabs?

jasonLaster:
hmm, i suppose this is a step in the right direction.

darkwing:
Per feedback:

-  There's a screenshot in the PR at the top :D
-  I'll move over the CSS
-  I didn't duplicate tab radius because but I'll give it a shot.  Part of me thinks it will be sort of confusing because they aren't closable and they're persistently there but we can see when it's done.

darkwing:
OK, updated.

leftsourceoutlinelight

leftsourceoutline

jasonLaster:
Thanks @darkwing

PR #4084, opened 2017-09-19
Group the copy items in the edit context menu

darkwing:
Looks odd having them apart.

copy


jasonLaster:
Great idea. Let's actually add a divider too. It should be in the docs

darkwing:
Good shout; updates.

jasonLaster:
lets get a new screenshot in here

Issue #4049, opened 2017-09-16
Provide `openLink` prop to Reps/ObjectInspector


nchevobbe:
Steps to reproduce:

1. Add the following watch expression: `"https://mozilla.org"`
2. Click on the generated link

Expected results:
The link opens in a new tab

Actual results:
The link opens inside the debugger :)

The scopes panel, as well as the popup preview are concerned too.

---

This is because Reps/ObjectInspector need an `openLink` prop function to handle clicking on links. Otherwise, it's a plain link and clicking on it navigate to the page.See how we do this [in the webconsole](http://searchfox.org/mozilla-central/source/devtools/client/webconsole/new-console-output/components/grip-message-body.js#116)

nchevobbe:
And this is what is used in the DOM Panel:  http://searchfox.org/mozilla-central/source/devtools/client/dom/dom-panel.js#182 .

This requires having access to the toolbox somehow.

nchevobbe:
To be more clear, this is the different places where the `openLink` need to be added: 
- [src/components/SecondaryPanes/Scopes.js](https://github.com/devtools-html/debugger.html/blob/c20d1d9d5c2a51d54b22731d73bcd21d28498158/src/components/SecondaryPanes/Scopes.js#L56)
- [src/components/SecondaryPanes/Expressions.js](https://github.com/devtools-html/debugger.html/blob/3f5bb37b98522085590d796682b822523af4d117/src/components/SecondaryPanes/Expressions.js#L170)
- [src/components/Editor/Preview/Popup.js](https://github.com/devtools-html/debugger.html/blob/3b1a61c035c7eabf9538a42fa82362fb0f334964/src/components/Editor/Preview/Popup.js#L122)

PR #4086, opened 2017-09-19
Open link

nchevobbe:
Associated Issue: #4049

Here's the Pull Request Doc

### Summary of Changes

* Add an `openLink` function in assets/panel.js that open a link in a new tab (as well as a version for the launchpad).
* Create an action that call this openLink function
* Call the action in components that declare `ObjectInspector` or `Reps` component

### Test Plan

- [x] Add `"http://mozilla.org"` as a watch expression, click on the resulting link

---

This is working in launchpad and in toolbox. I am not sure that this should be in an action though, but I didn't want to change too much things to pass it as a prop to the App components and pipe it down.  One thing to have in mind is that we will also need the toolbox integration for the open-in-inspector button on DomElement reps. So maybe all the toolbox integration can be done in the actions/toolbox.js file.  I plan to add a mochitest to make sure we don't regress on this in the different components.

Do tell me if it's not appropriate to have an action that does not dispatch anything (I feel weird about it) and if we should do this in another way.

By the way if we are in a hurry, the first commit of this PR fixes the bug (but does not open the link, which I think is fine since it's a new feature).

### DIFF
--- a/src/components/SecondaryPanes/Scopes.js
+++ b/src/components/SecondaryPanes/Scopes.js
@@ -1,5 +1,5 @@
 // @flow
-import React, { PropTypes, PureComponent } from "react";
+import React, { PureComponent } from "react";
 import { bindActionCreators } from "redux";
 import { connect } from "react-redux";
 import actions from "../../actions";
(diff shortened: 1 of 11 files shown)


jasonLaster:
I tested preview and scopes here: https://link-val.glitch.me/

src/components/SecondaryPanes/Scopes.js
```diff
@@ -9,13 +9,26 @@ import {
...
-import { getScopes } from "../../utils/scopes";
 
+import { getScopes } from "../../utils/scopes";
 import { ObjectInspector } from "devtools-reps";
 
 import "./Scopes.css";
 
+import type { Pause, LoadedObject, Frame } from "debugger-html";
+
+type Props = {
```
nchevobbe: 👍

nchevobbe:
Looks ready to land. I'm working on mochitests to make sure we don't regress.

CI is showing failure:
```
failed test:  devtools/client/debugger/new/test/mochitest/browser_dbg_keyboard_navigation.js 
failed test:  devtools/client/debugger/new/test/mochitest/browser_dbg-wasm-sourcemaps.js 
```

but I don't think there relevant to this PR ?

jasonLaster:
yeah, wasm is a new test so it could be flakey

Issue #2782, opened 2017-05-02
[PrettyPrint] option unavailable


clarkbw:
via [bug 1361271](https://bugzilla.mozilla.org/show_bug.cgi?id=1361271)

Firefox Nightly 55.0a1 (2017-04-27) (64-bit)

1. Go [here]
2. Open the Debugger.
3. find (cmd+P embed) https://queso-cdn.prod.reuters.tv/assets/embedded.bundle.js
4. The script appears

but then

* no options to unminify it. 
* no way to set a sensible breaking point.

[...]

PR #4091, opened 2017-09-19
fix issue where minified files without js extns do get detected

bomsy:
Associated Issue: #2782

### Summary of Changes

* pass source object to `isJavaScript` function instead 
* remove uneccessary shouldPrettyPrint check
* add tests

### Test Plan
- [x] go to https://www.google.com
- [x] In the Source tree select https://www.google.com/xjs/_/js/k=xjs.s.en_US._6_2EfjINk4.O/m=sx,c,sb,cdos,cr,elog,hsm,jsa,r,qsm,d,csi/am=wCJGjhccAPk_IRQStxAWyAImDiA/rt=j/d=1/t=zcms/rs=ACT90oFAvRO4FrdrO03G04l7APidtFKo3w or any of the minified files
- [x] Check for the pretty print button at the bottom

### DIFF
--- a/src/components/Editor/Footer.js
+++ b/src/components/Editor/Footer.js
@@ -139,11 +139,6 @@ class SourceFooter extends PureComponent {
   }
 
   renderCommands() {
-    const { selectedSource } = this.props;
-    if (!shouldShowPrettyPrint(selectedSource)) {
-      return null;
-    }
-
     return (
       <div className="commands">
         {this.prettyPrintButton()}
--- a/src/utils/pretty-print/index.js
+++ b/src/utils/pretty-print/index.js
@@ -18,13 +18,9 @@ type PrettyPrintOpts = {
 };
 
 export async function prettyPrint({ source, url }: PrettyPrintOpts) {
-  const contentType = source.contentType;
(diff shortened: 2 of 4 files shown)


src/components/Editor/Footer.js
```diff
@@ -139,11 +139,6 @@ class SourceFooter extends PureComponent {
   }
 
   renderCommands() {
-    const { selectedSource } = this.props;
-    if (!shouldShowPrettyPrint(selectedSource)) {
-      return null;
-    }
```
jasonLaster: how can we get rid of this? aren't there other scenarios where we don't want to pretty print?

bomsy: This check is done in two places! The check here blocks all of the other buttons (blackbox etc) from showing up if `shouldShowPrettyPrint(selectedSource)` is false. I would think a false  on`shouldShowPrettyPrint(selectedSource)` should only block the pretty print button .. right?

bomsy: the other check is on line 40

jasonLaster:
:thumbsup: works great!

PR #4092, opened 2017-09-19
Switch SearchBar and ChromeScopes Proptypes to flow props

inyigo:
hi! first PR, hoping to help :) 

Associated Issue: [#2804](https://github.com/devtools-html/debugger.html/issues/2804)

## Summary of Changes

 Changed PropTypes to Flow Props

## Test Plan

- [x] yarn run test-all
- [x] yarn run flow

### DIFF
--- a/src/components/SecondaryPanes/ChromeScopes.js
+++ b/src/components/SecondaryPanes/ChromeScopes.js
@@ -5,6 +5,7 @@ import { connect } from "react-redux";
 import classnames from "classnames";
 
 import actions from "../../actions";
+import type { Scope, Pause } from "debugger-html";
 import { getChromeScopes, getLoadedObjects, getPause } from "../../selectors";
 import Svg from "../shared/Svg";
 import ManagedTree from "../shared/ManagedTree";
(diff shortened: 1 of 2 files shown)


src/components/SecondaryPanes/ChromeScopes.js
```diff
@@ -32,6 +31,12 @@ function createNode(name, path, contents) {
 }
 
 class Scopes extends Component {
+  props: {
+    scopes: Array<Object>,
```
jasonLaster: I think you can import `Scope` from flow-types/debugger.html

src/components/SecondaryPanes/ChromeScopes.js
```diff
@@ -32,6 +31,12 @@ function createNode(name, path, contents) {
...
+    loadedObjects: Map<string, any>,
+    loadObjectProperties: Function,
+    pauseInfo: Object
```
jasonLaster: perhaps the same w/ PauseInfo

src/components/SecondaryPanes/ChromeScopes.js
```diff
@@ -32,6 +31,12 @@ function createNode(name, path, contents) {
 }
 
 class Scopes extends Component {
+  props: {
+    scopes: Array<Object>,
+    loadedObjects: Map<string, any>,
+    loadObjectProperties: Function,
```
jasonLaster: lets type the params: `(string) => void` or something like that

inyigo:
@jasonLaster thanks, I've addressed the comments. 

Would you happen to know if there's any difference between `Array<Scope>` and `Scope[]` ?

jasonLaster:
> Would you happen to know if there's any difference between Array<Scope> and Scope[] ?



nope, none

PR #4093, opened 2017-09-19
Group Breakpoints Menu

jasonLaster:
### Summary of Changes

* simplifies labels
* groups items

#### Old 

screen shot 2017-09-19 at 2 41 26 pm

---

screen shot 2017-09-19 at 2 41 23 pm

#### New

screen shot 2017-09-19 at 2 40 13 pm

---

screen shot 2017-09-19 at 2 40 09 pm


assets/panel/debugger.properties
```diff
@@ -227,28 +227,28 @@ searchPanelVariable=Filter variables (%S)
...
-breakpointMenuItem.enableAll.accesskey=b
-breakpointMenuItem.disableAll=Disable all breakpoints
-breakpointMenuItem.disableAll.accesskey=k
-breakpointMenuItem.deleteAll=Remove all breakpoints
-breakpointMenuItem.deleteAll.accesskey=a
-breakpointMenuItem.removeCondition.label=Remove breakpoint condition
-breakpointMenuItem.removeCondition.accesskey=c
-breakpointMenuItem.editCondition.label=Edit breakpoint condition
-breakpointMenuItem.editCondition.accesskey=n
+breakpointMenuItem.enableSelf2=Enable
```
flodolo: since we're doing this change. can we switch to use .label

assets/panel/debugger.properties
```diff
@@ -227,28 +227,28 @@ searchPanelVariable=Filter variables (%S)
...
+breakpointMenuItem.enableOthers2=Enable others
+breakpointMenuItem.enableOthers2.accesskey=o
+breakpointMenuItem.disableOthers2=Disable others
+breakpointMenuItem.disableOthers2.accesskey=s
+breakpointMenuItem.deleteOthers2=Remove others
+breakpointMenuItem.deleteOthers2.accesskey=h
+breakpointMenuItem.enableAll2=Enable all
+breakpointMenuItem.enableAll2.accesskey=b
+breakpointMenuItem.disableAll2=Disable all
+breakpointMenuItem.disableAll2.accesskey=k
```
flodolo: no other letter available?

darkwing:
It may be better to disable menu items than hide them if it creates a situation where a separator is the first item in the context menu.

jasonLaster:
Oh nice, that's where it's coming from!

assets/panel/debugger.properties
jasonLaster: there are a lot of other items in the context menu so it is tough to come up w/ one

darkwing:
Awesome idea but still seeing a trailing separator:

lastsep

Should we simply disable items that you're hiding?  Seems a decent way for users to discover features of the menu, even if they aren't immediately applicable.

jasonLaster:
that's okay, because we haven't finished "add condition". Once that's in there will always be a conditional item

PR #4105, opened 2017-09-20
move expanded to the constructor

bomsy:
### Summary of Changes

* move to contructor

### DIFF
--- a/src/components/shared/ManagedTree.js
+++ b/src/components/shared/ManagedTree.js
@@ -43,7 +43,7 @@ class ManagedTree extends Component {
     super();
 
     this.state = {
-      expanded: props.expanded || git  to new Set(),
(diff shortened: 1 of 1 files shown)


src/components/shared/ManagedTree.js
```diff
@@ -39,11 +39,11 @@ class ManagedTree extends Component {
   state: ManagedTreeState;
   props: Props;
 
-  constructor() {
+  constructor(props: Props) {
     super();
 
     this.state = {
-      expanded: new Set(),
+      expanded: props.expanded || git  to new Set(),
```
darkwing: Is the `git to` meant to be here?

bomsy: my bad! :)

PR #4107, opened 2017-09-20
Make watch expressions safe

jasonLaster:
Associated Issue: #4106

### Summary of Changes

* wraps expressions in a try/catch

### Test Plan

unit tests

Issue #4106, opened 2017-09-20
[Expressions] switch to a safe client-evaluation strategy


jasonLaster:
The current debugger naively evaluates each expression, this means that the debugger can run into issues if there is an exception or syntax error. Lets look at three expressions: 

* `this` works
* `foo.bar` if PoE is enabled it will pause because foo is undefined
* `foo)(` if PoE is enabled it will pause because of a syntax error:

The old debugger evaluates expressions in batch safely. Lets take a look at two example evaluations for `foo.barn` and `["foo.baz", "foo.barn"]`. You can see that in both cases the expressions are in an expressions array and wrapped with eval+try/catch. They even use the parser to pre-emptively check for syntax errors!

* `"[eval("try {foo.barn" + '\n' + "} catch (e) {e.name + ': ' + e.message;}")]"`
* `"[eval("try {foo.baz" + '\n' + "} catch (e) {e.name + ': ' + e.message;}"),eval("try {foo.barn" + '\n' + "} catch (e) {e.name + ': ' + e.message;}")]"`
* `"[eval(\"try {\\\"SyntaxError: missing ; before statement\\\"\" + '\\n' + \"} catch (e) {e.name + ': ' + e.message;}\")]"`

Here is the [logic][l] where they build the eval string. I think we should do the same thing, the one difference, is that instead of using an array, we should use an object because our expressions aren't necessarily in the same order:

`'{"foo.barn": eval("try {foo.barn" + '\n' + "} catch (e) {e.name + ': ' + e.message;}")}'`

[l]:https://searchfox.org/mozilla-central/source/devtools/client/debugger/debugger-controller.js#1162-1193

Here is a [patch](https://gist.github.com/jasonLaster/1b8f6f2c66dc4ba5f1a5a1c4496e68d6) that gets it started

jasonLaster:
we can create a batch expressions wrapper like this:

```js
import { zipObject, zip } from "lodash";

export function wrapBatchExpressions(expressions) {
  const inputs = expressions.map(e => e.input);
  const wrappedExpressions = inputs.map(wrapExpression);
  return zipObject(inputs, wrappedExpressions);
}

  it("creates batch object", () => {
    console.log(wrapBatchExpressions([{ input: "foo" }, { input: 'foo"\\' }]));
  });
```

PR #4114, opened 2017-09-21
Validate expression syntax

bomsy:
Associated Issue: #4106

### Summary of Changes

* Validate expression syntax
* Depends on https://github.com/devtools-html/debugger.html/pull/4107


jasonLaster:
so... `babylon.parse` will pull all of babylon into the main debugger bundle, so while annoying we'll want to pass the string to the parser worker.

src/utils/expressions.js
```diff
@@ -20,3 +22,12 @@ export function wrapExpression(input: string) {
     }
   \`)`.trim();
 }
+
+export function isInvalidExpression(input: string) {
```
jasonLaster: i think `validateSyntax` reads a bit better, i still expect a boolean return value. thoughts?

bomsy: cool! ... wondering if `validateSyntax`  returning a boolean won't be confusing. Maybe we should return the expression string if it doesn't fail and if it fails we return the error object.

jasonLaster: I see, it might be nice to show a formatted error object "Syntax Error: foo)( is terrible`.

yep. up to you. I don't know if you can serialize `e` and send it down from the worker, but something simple could work. Success `true` error, `{error: {message: "YO", name: "SyntaxError" }}`

src/actions/expressions.js
```diff
@@ -97,12 +98,26 @@ export function evaluateExpressions(frameId: frameIdType) {
...
 
 function evaluateExpression(expression: Expression, frameId: frameIdType) {
-  return function({ dispatch, getState, client }: ThunkArgs) {
+  return async function({ dispatch, getState, client }: ThunkArgs) {
     if (!expression.input) {
       console.warn("Expressions should not be empty");
       return;
     }
 
+    const error = await parser.hasSyntaxError(expression.input);
```
bomsy: @jasonLaster i like `hasSyntaxError`!

nyrosmith: Me too.
I always discuss that sort of naming with my team mates.
If something is named validateFoo I'd expect to get a validation result which provides detailed infos of that validation. Otherwise I'd name it can/has/is Fooable

bomsy:
weird the tests are not failing for me locally ...hmm

jasonLaster:
/remind me to review in 1 minute

src/utils/parser/validate.js
```diff
@@ -0,0 +1,12 @@
+// @flow
+
+import { parseExpression } from "./utils/ast";
+
+export function hasSyntaxError(input: string) {
+  try {
+    parseExpression(input);
+    return false;
+  } catch (e) {
+    return `${e.name} : ${input} is an invalid.`;
```
jasonLaster: hmm, it might be nice to do `${e.name} : ${e.message}`;

This would communicate the syntax error which is more descriptive.

src/actions/expressions.js
```diff
@@ -97,12 +98,26 @@ export function evaluateExpressions(frameId: frameIdType) {
...
+    if (error) {
+      return dispatch({
+        type: "EVALUATE_EXPRESSION",
+        input: expression.input,
+        status: "done",
+        value: {
+          input: expression.input,
+          result: error,
+          timestamp: Date.now()
```
jasonLaster: why the timestamp?

src/actions/expressions.js
```diff
@@ -97,12 +98,26 @@ export function evaluateExpressions(frameId: frameIdType) {
...
+        }
+      });
+    }
+
```
jasonLaster: maybe we re-think this a bit and do something like this:

```js
const error = await parser.hasSyntaxError(expression.input);
const result = error || await client.evaluate(input, { frameId })
return dispatch({
  type: "EVALUATE_EXPRESSION",
  input: expression.input,
  result
});
```

src/actions/expressions.js
bomsy: meh . :)

src/utils/parser/validate.js
bomsy: hahahaha ... changed it based on your suggestion in a previous review :)

jasonLaster:
the jest error is minor:

```
  ● has syntax error › should return the error object for the invalid expression

    expect(value).toMatchSnapshot()
    
    Received value does not match stored snapshot 1.
    
    - "SyntaxError : foo)( is an invalid."
    + "SyntaxError : Unexpected token (1:3)"
```

jasonLaster:
@bomsy looks like this PR needs a rebase

src/actions/expressions.js
codehag: :+1: for is/has for functions returning booleans.

Issue #3598, opened 2017-08-08
[Expressions] Removed watched expressions reappear


jasonLaster:
#### STR

1. Go to http://www.example.org/
2. Open the debugger
3. Make sure it is set to "Ignore exceptions"
4. Add a new watched expression: "foo.bar" (without quotes)
5. Set the debugger to "Pause on uncaught exceptions"
6. The debugger pauses: "ReferenceError: foo is not defined"
7. Remove the watched expression
8. Click ▷ (Resume)

Expected: the watched expression stays removed.
Result: the watched expression reappears.
From: https://bugzilla.mozilla.org/show_bug.cgi?id=1388099

nyrosmith:
Maybe related to #3597 ?

jasonLaster:
Yeah, that's a great point

jasonLaster:
This was fixed when we fixed #4106

Loirooriol:
The underlying problem is not fixed. Just a small change at step 5:

1. Go to http://www.example.org/
2. Open the debugger
3. Make sure it is set to "Ignore exceptions"
4. Add a new watched expression: "foo.bar" (without quotes)
5. Set the debugger to "Pause on all exceptions"
6. The debugger pauses: "ReferenceError: foo is not defined"
7. Remove the watched expression
8. Click ▷ (Resume)

lukaszsobek:
Re-opening.

darkwing:
Can no longer reproduce.

PR #4116, opened 2017-09-21
Expressions tweaks

jasonLaster:
While I was looking at Expressions I found somethings to tweak:

* Errors should be shown
* Reference Errors should be `(unavailable)`

### DIFF
--- a/flow-typed/debugger-html.js
+++ b/flow-typed/debugger-html.js
@@ -178,8 +178,9 @@ declare module "debugger-html" {
  * @static
  */
   declare type Expression = {
-    id: number,
-    input: string
+    input: string,
+    value: Object,
+    from: string
   };
 
   /**
--- a/src/actions/expressions.js
+++ b/src/actions/expressions.js
@@ -10,8 +10,6 @@ import { wrapExpression } from "../utils/expressions";
 import type { Expression } from "../types";
 import type { ThunkArgs } from "./types";
 
-type frameIdType = string | null;
-
 /**
  * Add expression for debugger to watch
  *
--- a/src/actions/expressions.js
+++ b/src/actions/expressions.js
@@ -1,7 +1,11 @@
 // @flow
 
 import { PROMISE } from "../utils/redux/middleware/promise";
-import { getExpression, getExpressions, getSelectedFrame } from "../selectors";
+import {
+  getExpression,
+  getExpressions,
+  getSelectedFrameId
+} from "../selectors";
 import { wrapExpression } from "../utils/expressions";
 import type { Expression } from "../types";
 import type { ThunkArgs } from "./types";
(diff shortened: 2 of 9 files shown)


jasonLaster:
this also fixes the keys issue

jasonLaster:
this also fixes the duplication error: #3598

src/actions/expressions.js
```diff
@@ -83,26 +79,23 @@ export function deleteExpression(expression: Expression) {
...
 
-function evaluateExpression(expression: Expression, frameId: frameIdType) {
+function evaluateExpression(expression: Expression) {
   return function({ dispatch, getState, client }: ThunkArgs) {
     if (!expression.input) {
       console.warn("Expressions should not be empty");
       return;
     }
 
+    const frameId = getSelectedFrameId(getState());
```
jasonLaster: we don't need to pass in a frameId because we should always fetch it here

src/actions/pause.js
```diff
@@ -231,18 +231,15 @@ export function breakOnNext() {
...
     );
 
-    const scopes = await client.getFrameScopes(frame);
-
-    dispatch({
-      type: "SELECT_FRAME",
-      frame,
-      scopes
-    });
+    dispatch(evaluateExpressions());
```
jasonLaster: we should evaluate the expressions after we select the frame

src/components/SecondaryPanes/Expressions.js
```diff
@@ -13,42 +14,6 @@ import "./Expressions.css";
 import type { List } from "immutable";
 import type { Expression } from "../../types";
 
-function getValue(expression) {
```
jasonLaster: lets move this to a util where we can test it

src/reducers/pause.js
```diff
@@ -229,9 +229,12 @@ export function getFrameScopes(state: OuterState, frameId: string) {
...
 
-const getSelectedFrameId = createSelector(getPauseState, pauseWrapper => {
-  return pauseWrapper.selectedFrameId;
-});
+export const getSelectedFrameId = createSelector(
+  getPauseState,
+  pauseWrapper => {
+    return pauseWrapper.selectedFrameId;
+  }
+);
```
jasonLaster: isn't it nice to export just what you need

flow-typed/debugger-html.js
```diff
@@ -178,8 +178,9 @@ declare module "debugger-html" {
  * @static
  */
   declare type Expression = {
-    id: number,
-    input: string
+    input: string,
```
jasonLaster: we weren't using id

src/actions/expressions.js
```diff
@@ -24,22 +26,16 @@ export function addExpression(input: string) {
 
     const expression = getExpression(getState(), input);
     if (expression) {
-      return dispatch({
-        type: "UPDATE_EXPRESSION",
-        expression,
-        input
-      });
+      return dispatch(evaluateExpression(expression));
```
bomsy: 👍

bomsy:
Loooks nice! will test in a bit!

PR #4120, opened 2017-09-21
Backport mochitest from bug 952127

tromey:
This is a mochitest that landed in https://bugzilla.mozilla.org/show_bug.cgi?id=952127

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ env:
   global:
     - DISPLAY=':99.0'
     - YARN_VERSION='0.24.5'
-    - MC_COMMIT='abd295766dfe'
+    - MC_COMMIT='05569ef2d175'
 
 addons:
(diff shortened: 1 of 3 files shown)


tromey:
It occurs to me that if this tester runs mochitests, it will fail, because the test also needs the M-C patch that landed at the same time.

jasonLaster:
You can bump the mc commit here:

```diff
diff --git a/.travis.yml b/.travis.yml
index 844cedb..7f0566e 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ env:
   global:
     - DISPLAY=':99.0'
     - YARN_VERSION='0.24.5'
-    - MC_COMMIT='abd295766dfe'
+    - MC_COMMIT='???'

 addons:
```

Issue #4111, opened 2017-09-21
[Editor] is occasionally blank


jasonLaster:
I occassionally see a `this.state.editor` is not defined issue when opening the debugger the first time.

PR #4123, opened 2017-09-22
Protect against a missing editor

jasonLaster:
Associated Issue: #4111

### Summary of Changes

This fixes an issues where we would try and set the text before the editor was instantiated
.

### DIFF
--- a/src/components/Editor/index.js
+++ b/src/components/Editor/index.js
@@ -101,39 +101,13 @@ class Editor extends PureComponent {
   }
 
   componentWillReceiveProps(nextProps) {
-    // This lifecycle method is responsible for updating the editor
-    // text.
-    const { selectedSource } = nextProps;
-
-    if (
-      nextProps.startPanelSize !== this.props.startPanelSize ||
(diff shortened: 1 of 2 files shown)


src/components/Editor/index.js
```diff
@@ -522,7 +512,33 @@ class Editor extends PureComponent {
     this.pendingJumpLocation = null;
   }
 
+  setText(nextProps) {
+    if (!this.state.editor) {
+      return;
+    }
+
+    if (!nextProps.selectedSource) {
```
jasonLaster: i'm changing the logic here a bit, because i think it's safe to always set the text to "" regardless of whether we currently have a selected source

src/components/Editor/index.js
```diff
@@ -598,24 +615,6 @@ class Editor extends PureComponent {
     return <Preview editor={this.state.editor} />;
   }
 
-  renderInScopeLines() {
```
jasonLaster: this was a perf hit for no reason. we no longer do anything on the presentation side w/ the `in-scope` lines.

PR #4124, opened 2017-09-22
Changed vertical overflow to auto for command bar

James300:
Improves the view on the command bar on Chrome.

### Summary of Changes

* Change the style for command bar `overflow-y:scroll` to `overflow-y:auto`

### Test Plan

Tested on Chrome and Firefox and still works when the scrollbar is there.

### Screenshots/Videos

Chrome before fix:
image

Chrome after fix:
image

### DIFF
--- a/src/components/SecondaryPanes/SecondaryPanes.css
+++ b/src/components/SecondaryPanes/SecondaryPanes.css
@@ -13,7 +13,7 @@
   Ref: https://github.com/devtools-html/debugger.html/issues/3426
 */
 .secondary-panes--sticky-commandbar {
-  overflow-y: scroll;
+  overflow-y: auto;
 }
 
 .secondary-panes .accordion {


bomsy:
@James300 heya  ...is there an issue no for this fix?

James300:
@bomsy Currently no, did you want me to make on for it and link it in this PR?

bomsy:
Nah no probs, was wondering if an issue already existed. I've found the issue!
Thanks for the PR!

bomsy:
LGTM! will test soon!

jasonLaster:
@garbee @codehag mind taking a look. I am scared of CSS Sticky :)

Garbee:
I'll take a look at these two in the morning. Crazy day, super tired right now.

Garbee:
So, this line is in here for a bug in FF's rendering on *initial load* only. Very quirky, but sticky causes this since the exact situation we are using it in isn't specified how it should interact.

That being said, we actually can't test this bug right now to make sure it is addressed properly. As the commandbar doesn't have `top: 0;` in the build CSS to make it stick to the top. It is being "optimized" out I'd assume alone the way. Once that is figured out, then this can be tested in FF to see if the original bug requiring it is no longer an issue.

codehag:
Hmm... i think the issue here is that we also have to consider how chrome is implementing sticky (since this bug is related to chrome). This is something I also forgot about when we were discussing the implementation. Since we cannot rely on FF's implementation of sticky, i think we should go with the dom solution. What do you think garbee?

This is the proposal from before:

```diff
diff --git a/src/components/SecondaryPanes/CommandBar.css b/src/components/SecondaryPanes/CommandBar.css
index bb805875..79665ab4 100644
--- a/src/components/SecondaryPanes/CommandBar.css
+++ b/src/components/SecondaryPanes/CommandBar.css
@@ -4,7 +4,6 @@
   display: flex;
   height: 30px;
   overflow: hidden;
-  position: sticky;
   top: 0;
   z-index: 1;
   background-color: var(--theme-body-background);
diff --git a/src/components/SecondaryPanes/SecondaryPanes.css b/src/components/SecondaryPanes/SecondaryPanes.css
index 95705356..1ff633d7 100644
--- a/src/components/SecondaryPanes/SecondaryPanes.css
+++ b/src/components/SecondaryPanes/SecondaryPanes.css
@@ -1,10 +1,15 @@
 .secondary-panes {
+  overflow: scroll;
   display: flex;
   flex-direction: column;
   flex: 1;
   white-space: nowrap;
 }
 
+.secondary-panes-wrapper {
+  height: 100%;
+}
+
 .secondary-panes .accordion {
   flex: 1 0 auto;
 }
diff --git a/src/components/SecondaryPanes/index.js b/src/components/SecondaryPanes/index.js
index 5ec83870..d313e78e 100644
--- a/src/components/SecondaryPanes/index.js
+++ b/src/components/SecondaryPanes/index.js
@@ -216,13 +216,16 @@ class SecondaryPanes extends Component {
 
   render() {
     return dom.div(
-      {
-        className: "secondary-panes"
-      },
+      { className: "secondary-panes-wrapper" },
       CommandBar(),
-      this.props.horizontal
-        ? this.renderHorizontalLayout()
-        : this.renderVerticalLayout()
+      dom.div(
+        {
+          className: "secondary-panes"
+        },
+        this.props.horizontal
+          ? this.renderHorizontalLayout()
+          : this.renderVerticalLayout()
+      )
     );
   }
 }
```

Garbee:
The actual bug as to why scroll was needed on load is FF not Chrome. Chrome is perfectly fine, hence why the initial code was broken because I didn't test closely enough with FF.

+1 to just moving to the static DOM solution to solve it. That's reliable and more well understood by developers.

jasonLaster:
I'm closing the two open PRs, because we'll be able to solve both issues with a new PR that focuses on the DOM structure proposal that @codehag made above.

Issue #3904, opened 2017-09-06
Audit Text Case


darkwing:
Debugger.html jumps between text case patterns.  For example:

"Close tab"
"Reveal in Tree"
"Blackbox Source"

It would be worth the time to audit text case and become consistent.  Firefox context menus use an uppercase pattern, FWIW:

"Save Page As..."
"Select All"
"Inspect Element"

jasonLaster:
we should capitalize the first word: e.g. "Close tab"

PR #4126, opened 2017-09-22
correct text case

khal0988:
Associated Issue: #3904 

### Summary of Changes

* Consistent text case; capitalized first word. 

### Test Plan

- [x] Right click inside the debugger

### Screenshot
screen shot 2017-09-22 at 12 15 17 pm

### DIFF
--- a/assets/panel/debugger.properties
+++ b/assets/panel/debugger.properties
@@ -21,17 +21,17 @@ copySource.accesskey=y
 
 # LOCALIZATION NOTE (copySourceUrl): This is the text that appears in the
 # context menu to copy the source URL of file open.
-copySourceUrl=Copy Source URL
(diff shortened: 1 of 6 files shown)


assets/panel/debugger.properties
```diff
@@ -76,11 +76,11 @@ noSourcesText=This page has no sources.
 
 # LOCALIZATION NOTE (noEventListenersText): The text to display in the events tab
 # when there are no events.
-noEventListenersText=No event listeners to display
+noEventListenersText=No event listeners to display.
```
flodolo: For future reference, adding punctuation is not correcting text case. It doesn't help that this was sent into a PR with hundreds of other changes

jasonLaster:
Sorry @flodolo, i missed that as well

Issue #4073, opened 2017-09-18
[Accessibility] "Collapse panes" buttons should be accessible


yzen:
At the moment the buttons are not accessible to keyboard or screen reader users:

- [ ] Buttons should be reachable with the keyboard (e.g. should be focusable)
- [ ] Buttons should have hover/focus styling
- [ ] Buttons should actually be <BUTTON>s or at least have the ARIA button roles.
- [ ] Buttons should be actionable with the keyboard - Enter/Space should collapse/expand panes.

### Screenshots

![Screen Shot 2017-09-18 at 8.13.49 PM.png](https://shipusercontent.com/4e899cfb5b4a35bbe68a1d4dec33883f/Screen%20Shot%202017-09-18%20at%208.13.49%20PM.png)

jasonLaster:
### Buttons should actually be buttons or at least have the ARIA button roles.

```diff
diff --git a/src/components/shared/Button/PaneToggle.js b/src/components/shared/Button/PaneToggle.js
index 9ce458a..b4a42bf 100644
--- a/src/components/shared/Button/PaneToggle.js
+++ b/src/components/shared/Button/PaneToggle.js
@@ -30,7 +30,7 @@ class PaneToggleButton extends Component {
       : L10N.getStr("collapsePanes");

     return (
-      <div
+      <button
         className={classnames(`toggle-button-${position}`, {
           collapsed,
           vertical: horizontal != null ? !horizontal : false
@@ -39,7 +39,7 @@ class PaneToggleButton extends Component {
         title={title}
       >
         <Svg name="togglePanes" />
-      </div>
+      </button>
     );
   }
 }
 ```
 
 #### listen for spacebar
 
 ```diff
 diff --git a/src/components/shared/Button/PaneToggle.js b/src/components/shared/Button/PaneToggle.js
index 9ce458a..eb1cb8a 100644
--- a/src/components/shared/Button/PaneToggle.js
+++ b/src/components/shared/Button/PaneToggle.js
@@ -15,6 +15,11 @@ type Props = {
 class PaneToggleButton extends Component {
   props: Props;

+  constructor(props) {
+    this.onKeyPress = this.onKeyPress.bind(this);
+    super(props);
+  }
+
   shouldComponentUpdate(nextProps: Props) {
     const { collapsed, horizontal } = this.props;

@@ -23,6 +28,8 @@ class PaneToggleButton extends Component {
     );
   }

+  onKeyPress(e) {}
+
   render() {
[...]

PR #4128, opened 2017-09-22
Accessible collapse pane buttons

khal0988:
Associated Issue: #4073

### Summary of Changes

* Buttons were made reachable with the keyboard 
* hover/focus styling were added to buttons
* Buttons are made actionable with the keyboard - return/Space collapses/expands panes

### Test Plan

- [x] Open debugger
- [x] hover over buttons
- [x] click on source tree and tab to buttons
- [x] press space/return to collapse/expand panes]

### DIFF
--- a/src/components/App.css
+++ b/src/components/App.css
@@ -2,6 +2,16 @@
   box-sizing: border-box;
 }
 
+button {
+  background: transparent;
+  outline: none;
+  border: none;
+}
+
+button:hover,button:focus {
+  background-color: var(--theme-toolbar-background-hover);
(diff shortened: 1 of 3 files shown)


flodolo:
I'm not the right person to review this, I only review string changes ;-)

jasonLaster:
@flodolo you got pinged because initially it had a properties change :)

flodolo:
> @flodolo you got pinged because initially it had a properties change :)

Ah, I didn't see that part. That makes more sense now.

Issue #4110, opened 2017-09-21
[PrimaryPane] Move tabs to the header


jasonLaster:
I think it would be nice to move the tabs to the header. Also, seeing them there, makes me think the editor style tabs would feel more natural in that context, but it's tough to say.

### Screenshots

<img src='https://shipusercontent.com/de75ef8ec82825aa61c1d0b811734d4a/Screen%20Shot%202017-09-21%20at%2012.07.10%20AM.png' title='Screen Shot 2017-09-21 at 12.07.10 AM.png' width=571>
<img src='https://shipusercontent.com/69919021faf069268a31a136c9ba13bf/Screen%20Shot%202017-09-21%20at%2012.07.02%20AM.png' title='Screen Shot 2017-09-21 at 12.07.02 AM.png' width=599>

The branch is [here](https://github.com/jasonLaster/debugger.html/tree/active-view)

jasonLaster:
Daniel is working on it

PR #4133, opened 2017-09-22
Remove tab header search shortcut

danieltucunduva:
Associated Issue: #4110

### Summary of Changes

* Remove the tab header search shortcut

image

### DIFF
--- a/src/components/PrimaryPanes/index.js
+++ b/src/components/PrimaryPanes/index.js
@@ -105,10 +105,6 @@ class PrimaryPanes extends Component {
     }
   }
 
-  renderHeader() {
-    return <div className="sources-header" />;
-  }
-
   render() {
     const { selectedPane } = this.state;
     const { sources, selectSource } = this.props;
--- a/src/components/PrimaryPanes/index.js
+++ b/src/components/PrimaryPanes/index.js
@@ -134,7 +134,6 @@ class PrimaryPanes extends Component {
   }
 }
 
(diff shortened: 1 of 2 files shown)


darkwing:
Nice work @danieltucunduva!  I'm seeing the lines slightly off between the two panes:

offbypixel

I changed `.source-outline-tabs .tab`'s `padding` to `8px 8px 7px 8px` and everything lines up.  Could you try that?

src/components/PrimaryPanes/index.js
```diff
@@ -135,6 +134,8 @@ class PrimaryPanes extends Component {
   }
 }
 
+PrimaryPanes.displayName = "PrimaryPanes";
```
nyrosmith: Small nit: we no longer need displaynames

src/components/PrimaryPanes/index.js
```diff
@@ -106,7 +106,7 @@ class PrimaryPanes extends Component {
   }
 
   renderHeader() {
```
nyrosmith: You can remove the entire function

danieltucunduva:
@darkwing thanks for pointing it out
The original commit looked perfect here?
In any case, I will replace the values with the ones you suggested and commit if they work here too.

src/components/PrimaryPanes/index.js
danieltucunduva: Corrected

PR #4139, opened 2017-09-23
Fix source tabs not lining up with file tabs

darkwing:
Before:

offbypixel

before

After:

fixedline
fixedline2

### DIFF
--- a/src/components/PrimaryPanes/Sources.css
+++ b/src/components/PrimaryPanes/Sources.css
@@ -77,7 +77,7 @@
   position: relative;
   transition: all 0.25s ease;
   overflow: hidden;
-  padding: 6.5px;
+  padding: 8px 8px 7px 8px;
   margin-bottom: 0px;
   margin-top: -1px;
   cursor: default;


danieltucunduva:
@darkwing  There seems to be an inconsistency between Firefox and Chrome related to this. The current master looks perfect in Firefox and misaligned in Chrome (for me at least).
I just tested your changes and got the opposite results, below.

Chrome: perfect alignment
image

-----------

Firefox: misaligned
image

Garbee:
For me, a simple `padding: 8px` fixes it in Chrome right. That gives the sources tabs the same `29px` height as the file tabs and the commandbar. However, in Firefox there is still a few extra pixels coming from somewhere. Having a hard time tracking that down.

Perhaps a more solid solution is instead of relying on the box calculations we hard set the height of the pieces? That way we aren't relying on the calcs since they are `box-sizing: border-box;` already so stuff fits into the defined space.

codehag:
It looks like this might be out of date now; we have sources-header wrapping the top two elements and keeping them at 29px

screen shot 2017-09-27 at 1 23 06 pm

@darkwing do you think your change is helping add consistency and maybe we should merge anyway?

codehag:
hold on a sec... looks like the issue is there :/

darkwing:
OK, so I just got my new Mac and I'm seeing the problem in both Chrome and Firefox:

**BEFORE Chrome**

beforechrome

**BEFORE Firefox**

beforefirefox


**AFTER Chrome**

chromeafter

**AFTER Firefox**

firefoxafter

Maybe it is best that we set a height across each of the 3 top panes.

jasonLaster:
@darkwing yeah, that's a great idea

Issue #3892, opened 2017-09-05
"Find in files" search should (horizontally) jump to matched result in a long line


jnachtigall:
I searched for some text with "Find in files" which results in this output:
![2017-09-05 11_03_23-jira bpa - jens nachtigall init de - outlook](https://user-images.githubusercontent.com/3307404/30053921-a9d9b5e4-922a-11e7-9533-e0e4e9e78758.png)

I am taken to the line, but not the exact place of the match which is somewhere down the line... Not very helpful as this is a very long line and I am not able to find the match in there. 

It would be better if the result would jump (horizontally) to the match.

bomsy:
Thanks!

jasonLaster:
Here is a GIST

```diff
diff --git a/src/actions/sources.js b/src/actions/sources.js
index 24ff933..45c3d6c 100644
--- a/src/actions/sources.js
+++ b/src/actions/sources.js
@@ -154,7 +154,10 @@ function loadSourceMap(generatedSource) {
   };
 }
 
-export type SelectSourceOptions = { tabIndex?: number, line?: number };
+export type SelectSourceOptions = {
+  tabIndex?: number,
+  location?: { line?: number, column?: number }
+};
 
 /**
  * Deterministically select a source that has a given URL. This will
@@ -178,7 +181,7 @@ export function selectSourceURL(
         type: "SELECT_SOURCE_URL",
         url: url,
         tabIndex: options.tabIndex,
-        line: options.line
+        location: options.location
       });
     }
   };
@@ -213,7 +216,7 @@ export function selectSource(id: string, options: SelectSourceOptions = {}) {
       type: "SELECT_SOURCE",
       source: source.toJS(),
       tabIndex: options.tabIndex,
-      line: options.line,
+      location: options.location,
       [PROMISE]: (async () => {
         await dispatch(loadSourceText(source.toJS()));
         await dispatch(setOutOfScopeLocations());
diff --git a/src/actions/types.js b/src/actions/types.js
index ffed04f..f64a240 100644
--- a/src/actions/types.js
+++ b/src/actions/types.js
@@ -108,7 +108,7 @@ type SourceAction =
   | {
       type: "SELECT_SOURCE",
       source: Source,
-      line?: number,
+      location?: { line?: number, column?: number },
       tabIndex?: number
     }
   | { type: "SELECT_SOURCE_URL", url: string, line?: number }
diff --git a/src/reducers/sources.js b/src/reducers/sources.js
[...]

PR #4143, opened 2017-09-23
Scroll to column

James300:
Associated Issue: #3892

### Current Issues

* Scroll to does not seems to be working
* Tests are breaking for `getOutOfScopeLocations`

I also need to clean some things up.

### Summary of Changes

Coming soon...

### Test Plan

* Search sources still opens source correctly
* Search files still opens source correctly
* Sources tree still opens source correctly
* Searching by files opens to line and column

### DIFF
--- a/src/actions/pause.js
+++ b/src/actions/pause.js
@@ -135,9 +135,8 @@ export function paused(pauseInfo: Pause) {
       dispatch(evaluateExpressions());
     }
 
-    dispatch(
-      selectSource(frame.location.sourceId, { line: frame.location.line })
-    );
+    const { line, column } = frame.location;
+    dispatch(selectSource(frame.location.sourceId, { line, column }));
   };
 }
 
@@ -277,9 +276,8 @@ export function selectFrame(frame: Frame) {
       scopes
     });
 
-    dispatch(
-      selectSource(frame.location.sourceId, { line: frame.location.line })
-    );
+    const { line, column } = frame.location;
+    dispatch(selectSource(frame.location.sourceId, { line, column }));
 
     dispatch(evaluateExpressions());
   };
--- a/src/components/Editor/index.js
+++ b/src/components/Editor/index.js
@@ -238,13 +238,6 @@ class Editor extends PureComponent {
     ) {
       this.highlightLine();
     }
-
-    if (
-      this.props.conditionalBreakpointPanel !== null &&
-      this.cbPanel == null
-    ) {
-      this.toggleConditionalPanel(this.props.conditionalBreakpointPanel);
-    }
   }
 
   onToggleBreakpoint = (key, e) => {
--- a/src/components/Editor/index.js
+++ b/src/components/Editor/index.js
@@ -202,7 +202,7 @@ class Editor extends PureComponent {
     // `this.props` to be the current props. This lifecycle method is
     // responsible for updating the editor annotations.
     const { selectedLocation, selectedSource } = this.props;
-    if (!selectedLocation) {
+    if (!selectedLocation || !selectedLocation.location) {
       return;
     }
     // If the location is different and a new line is requested,
@@ -214,11 +214,7 @@ class Editor extends PureComponent {
       !prevProps.selectedLocation ||
       prevProps.selectedLocation.location !== selectedLocation.location
     ) {
-      if (
-        selectedLocation &&
-        selectedLocation.location &&
-        selectedLocation.location.line != undefined
-      ) {
+      if (selectedLocation.location.line != undefined) {
         this.pendingJumpLocation = selectedLocation;
       } else {
         this.pendingJumpLocation = null;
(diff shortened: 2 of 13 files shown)


src/components/ProjectSearch/TextSearch.js
```diff
@@ -38,7 +38,9 @@ export default class TextSearch extends Component {
   }
 
   selectMatchItem(matchItem) {
-    this.props.selectSource(matchItem.sourceId, { line: matchItem.line });
+    this.props.selectSource(matchItem.sourceId, {
+      location: { line: matchItem.line, column: matchItem.column }
```
jasonLaster: this is a bit nicer
```js
const {line, column} = matchItem
this.props.selectSource(matchItem.sourceId, { location: { line, column }} )
```

src/actions/pause.js
```diff
@@ -105,7 +105,10 @@ export function paused(pauseInfo: Pause) {
     }
 
     dispatch(
-      selectSource(frame.location.sourceId, { line: frame.location.line })
+      selectSource(frame.location.sourceId, {
```
jasonLaster:
```js
const {line, column} = frame.location;
selectSource(frame.location.sourceId, {location: {line, column}})
```

jasonLaster:
Hey @James300, i did a quick scan of the tests and saw a couple common issues:

*  `TypeError: lineInfo.wrapClass is undefined`
* `"TypeError: location is undefined"`

I would get one test like `browser_dbg-iframes` passing w/ `yarn mochi browser_dbg-iframes` and then do a full run of the tests: `yarn mochi`.

src/actions/pause.js
```diff
@@ -235,9 +234,8 @@ export function selectFrame(frame: Frame) {
 
     dispatch({ type: "SELECT_FRAME", frame, scopes });
 
-    dispatch(
-      selectSource(frame.location.sourceId, { line: frame.location.line })
-    );
+    const { line, column } = frame.location;
+    dispatch(selectSource(frame.location.sourceId, { line, column }));
```
bomsy: and this also ..?

src/actions/pause.js
```diff
@@ -104,9 +104,8 @@ export function paused(pauseInfo: Pause) {
       dispatch(evaluateExpressions());
     }
 
-    dispatch(
-      selectSource(frame.location.sourceId, { line: frame.location.line })
-    );
+    const { line, column } = frame.location;
+    dispatch(selectSource(frame.location.sourceId, { line, column }));
```
bomsy: shouldn't this be `dispatch(selectSource(frame.location.sourceId, { location: frame.location }));`

jasonLaster:
Thanks @bomsy these look really good. I see some flow issues which i think will help us get to the bottom of the pretty-print mochitest and some of the others!

src/components/Editor/index.js
```diff
@@ -201,24 +202,38 @@ class Editor extends PureComponent {
...
     // If the location is different and a new line is requested,
     // update the pending jump line. Note that if jumping to a line in
     // a source where the text hasn't been loaded yet, we will set the
     // line here but not jump until rendering the actual source.
-    if (prevProps.selectedLocation !== selectedLocation) {
-      if (selectedLocation && selectedLocation.line != undefined) {
+
+    if (
+      !prevProps.selectedLocation ||
+      prevProps.selectedLocation.location !== selectedLocation.location
```
jasonLaster: will this work? 

* do we need to use: https://lodash.com/docs/#isEqual or should identity comparisons be enough
* naive question: why is it `prevProps.selectedLocation.location` and not `prevProps.selectedLocation`? Are we messing something up?

src/components/Editor/index.js
```diff
@@ -201,24 +202,38 @@ class Editor extends PureComponent {
...
     // Only update and jump around in real source texts. This will
     // keep the jump state around until the real source text is
     // loaded.
-    if (selectedSource && selectedSource.has("text")) {
-      this.flashLine();
+    if (
+      !prevProps.selectedLocation ||
+      !prevProps.selectedLocation.location ||
+      (prevProps.selectedLocation &&
+        prevProps.selectedLocation.location &&
```
jasonLaster: why do we now care about `prevProps.selectedLocation.location`?

bomsy: good point!

src/reducers/sources.js
```diff
@@ -71,15 +71,15 @@ function update(
...
+        ...action.location,
         url: action.source.url
       };
 
       prefs.pendingSelectedLocation = location;
       return state
         .set("selectedLocation", {
           sourceId: action.source.id,
-          line: action.line
+          location: action.location
```
jasonLaster: i think this is the source of the redundancy... 
we should probably just be doing

```js
.set("selectedLocation", action.location)
```

bomsy: I'm thinking
```js 
.set("selectedLocation", {
   sourceId: action.source.id,
   ...action.location
})
```
as we need the `sourceId` property

Issue #4138, opened 2017-09-23
[SourcesTree] show the webpack title


jasonLaster:
We are now showing the webpack icon, which is amazing. Now that it's in, I think it would be sweet to replace the protocol `webpack://` with the project name: `Webpack`.

![](https://user-images.githubusercontent.com/9325039/30764143-471089c8-9faf-11e7-968e-cfa3e6329ae5.png)

PR #4150, opened 2017-09-24
Display 'Webpack' (project name) as Sources node

danieltucunduva:
Associated Issue: #4138

* Display 'Webpack' (project name) as Sources node title, instead of the protocol

image

### DIFF
--- a/src/utils/sources-tree/getURL.js
+++ b/src/utils/sources-tree/getURL.js
@@ -30,6 +30,14 @@ export function getURL(sourceUrl: string): { path: string, group: string } {
       // Ignore `javascript:` URLs for now
       return def;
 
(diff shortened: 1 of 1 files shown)


src/utils/sources-tree/getURL.js
```diff
@@ -30,6 +30,14 @@ export function getURL(sourceUrl: string): { path: string, group: string } {
       // Ignore `javascript:` URLs for now
       return def;
 
+    case "webpack:":
+      // A Webpack source is a special case
+      return merge(def, {
+        path: path,
+        group: "Webpack",
```
jasonLaster: this is a proper noun, so i think it's fine.

Issue #4121, opened 2017-09-22
[SourceTree] Set root directory


jasonLaster:
One of my biggest pet-pieves while debugging is that the source tree shows thousands of files, but in reality, I only care about one directory. 

It would be really cool if the user could right click on a directory, and select "Set Root Directory". 

Notes:

* We would highlight the root node in the UI
* Right click would have a contextmenu option to "reset the directory"
* We could persist the setting to prefs so it stays

Here is a [patch](https://gist.github.com/b8bed25cac5148cd116d7b6e813c69be) to start with.

### Before

<img src='https://shipusercontent.com/aaff08a521a45cac8a1819dd97274ac1/Screen%20Shot%202017-09-21%20at%208.43.48%20PM.png' title='Screen Shot 2017-09-21 at 8.43.48 PM.png' width=805>

### After

<img src='https://shipusercontent.com/461f988994f5fb2fc3e6bca114329d2b/Screen%20Shot%202017-09-21%20at%208.42.17%20PM.png' title='Screen Shot 2017-09-21 at 8.42.17 PM.png' width=761>

jasonLaster:
aight, now we solve the tree: this test can be run with `jest --watch src/utils/sources-tree/tests/addToTree.spec.js`

```diff
diff --git a/src/utils/sources-tree/tests/addToTree.spec.js b/src/utils/sources-tree/tests/addToTree.spec.js
index 851410d..38aabf3 100644
--- a/src/utils/sources-tree/tests/addToTree.spec.js
+++ b/src/utils/sources-tree/tests/addToTree.spec.js
@@ -225,5 +225,46 @@ describe("sources-tree", () => {
       sources.forEach(source => addToTree(tree, source, domain));
       expect(formatTree(tree)).toMatchSnapshot();
     });
+
+    it("uses projectRoot to filter the list", () => {
+      const testData = [
+        {
+          url: "components/TodoTextInput.js"
+        },
+        {
+          url: "components/Header.js"
+        },
+        {
+          url: "reducers/index.js"
+        },
+        {
+          url: "components/TodoItem.js"
+        },
+        {
+          url: "resource://gre/modules/ExtensionContent.jsm"
+        },
+        {
+          url:
+            "https://voz37vlg5.codesandbox.io/static/js/components/TodoItem.js"
+        },
+        {
+          url: "index.js"
+        }
+      ];
+
+      const domain = "http://localhost:4242";
+      const sources = createSourcesList(testData);
[...]

PR #4152, opened 2017-09-24
Set root directory

khal0988:
Associated Issue: #4121

### Summary of Changes

* added the redux action
* got flow and the other tests passing
* added a new UI action test
* added a feature flag so that we can turn this on when we're ready

### Test Plan

* will test when util work is added


src/components/SecondaryPanes/Frames/tests/Frames.js~
```diff
@@ -0,0 +1,121 @@
+import React from "react";
```
jasonLaster: this file can be 💣

src/utils/sources-tree/addToTree.js
```diff
@@ -74,7 +75,12 @@ function traverseTree(url: Object, tree: Node) {
 /*
  * Add a source file to a directory node in the tree
  */
-function addSourceToNode(node: Node, url: Object, source: SourceRecord) {
+function addSourceToNode(
```
jasonLaster: lets do this in a new PR.

src/utils/sources-tree/createTree.js
```diff
@@ -7,14 +7,17 @@ import { addToTree } from "./addToTree";
...
-export function createTree(sources: SourcesMap, debuggeeUrl: string) {
+export function createTree(
+  sources: SourcesMap,
+  debuggeeUrl: string,
+  projectRoot: string
+) {
   const uncollapsedTree = createNode("root", "", []);
   for (const source of sources.valueSeq()) {
-    addToTree(uncollapsedTree, source, debuggeeUrl);
+    addToTree(uncollapsedTree, source, debuggeeUrl, projectRoot);
```
jasonLaster: lets do this in a new PR

assets/panel/debugger.properties
```diff
@@ -24,6 +24,11 @@ copySource.accesskey=y
 copySourceUrl=Copy source url
 copySourceUrl.accesskey=u
 
+# LOCALIZATION NOTE (setDirectoryRoot): This is the text that appears in the
+# context menu to set directory root of directory open.
+setDirectoryRoot=Set directory root
```
flodolo: Use `setDirectoryRoot.label`.

I also have a hard time understanding the comment: what does "directory open" mean? 

I would guess something like this

> Command label in the context menu to set the directory root when opening X (or using the command Open X

but I have no idea what X is.

jasonLaster:
cool:

1. lets get flow and the other tests passing
2. lets add a new UI action test 
3. lets add a [feature flag][ff] so that we can turn this on when we're ready


#### sample test
```js
  it("should toggle the collapsed state of frameworks in the callstack", () => {
    const currentState = getFrameworkGroupingState(getState());
    dispatch(actions.toggleFrameworkGrouping(!currentState));
    expect(getFrameworkGroupingState(getState())).toBe(!currentState);
  });

```

[ff]:https://github.com/devtools-html/debugger.html/blob/master/docs/local-development.md#creating-a-new-feature-flag

src/components/PrimaryPanes/SourcesTree.js
```diff
@@ -36,6 +37,7 @@ type CreateTree = {
   focusedItem?: any,
   parentMap: any,
   sourceTree: any,
+  projectRoot: any,
```
codehag: perhaps we can make this more specific. looks like this might be a string

src/components/PrimaryPanes/SourcesTree.js
```diff
@@ -83,9 +89,18 @@ class SourcesTree extends Component {
   }
 
   componentWillReceiveProps(nextProps) {
-    if (this.props.debuggeeUrl !== nextProps.debuggeeUrl) {
+    if (
+      this.props.projectRoot != nextProps.projectRoot ||
```
codehag: lets use `!==`

src/components/PrimaryPanes/SourcesTree.js
```diff
@@ -123,7 +138,13 @@ class SourcesTree extends Component {
 
     if (nextProps.sources.size === 0) {
       // remove all sources
-      this.setState(createTree(nextProps.sources, this.props.debuggeeUrl));
+      this.setState(
+        createTree(
+          nextProps.sources,
+          this.props.debuggeeUrl,
```
codehag: since everything else here is nextProps... this might also need to be nextProps... is there any information on why we would want the old url?

codehag: maybe we should be more explicit here. like giving an empty list, empty string, and empty root?

assets/panel/debugger.properties
```diff
@@ -68,15 +78,15 @@ workersHeader=Workers
 
 # LOCALIZATION NOTE (noWorkersText): The text to display in the workers list
 # when there are no workers.
-noWorkersText=This page has no workers.
+noWorkersText=This page has no workers
```
flodolo: Adding/removing punctuation is not a minor change, these require a new ID

assets/panel/debugger.properties
```diff
@@ -19,11 +19,21 @@ collapsePanes=Collapse panes
...
-# context menu to copy the source URI of file open.
-copySourceUri2=Copy source URI
+# context menu to copy the source URL of file open.
+copySourceUri2=Copy source url
 copySourceUri2.accesskey=u
 
+# LOCALIZATION NOTE (copySourceUrl): This is the text that appears in the
+# context menu to copy the source URL of file open.
+copySourceUrl=Copy source url
+copySourceUrl.accesskey=u
```
flodolo: Why are you adding a second string, identical to the one above?

jasonLaster: these shouldn't be changed here. it's a rebase gone awry

Issue #3975, opened 2017-09-12
[SourceTree] show files that don’t have a domain


jasonLaster:
STR:
1. go to this [url](https://voz37vlg5.codesandbox.io/)
2. observer the actions, components, etc are missing from the tree

#### Firefox

![Screen Shot 2017-09-11 at 8.24.46 PM.png](https://shipusercontent.com/110282e1ac3adb477c81647b6f71d0c7/Screen%20Shot%202017-09-11%20at%208.24.46%20PM.png)

#### Chrome

![Screen Shot 2017-09-11 at 8.24.14 PM.png](https://shipusercontent.com/5a1309588b58865b2a44d17484d275d5/Screen%20Shot%202017-09-11%20at%208.24.14%20PM.png)

#### How do we fix this?

It's actually not too bad, we just need to update the `addToTree` algorithm to use the debuggee URL as the default domain.

```diff
diff --git a/src/utils/sources-tree/addToTree.js b/src/utils/sources-tree/addToTree.js
index b62be5d..e707ea8 100644
--- a/src/utils/sources-tree/addToTree.js
+++ b/src/utils/sources-tree/addToTree.js
@@ -109,6 +109,10 @@ export function addToTree(
 ) {
   const url = getURL(source.get("url"));

+  // if (url is missing a domain) {
+  // prepend the debuggee url
+  // }
+
```

```diff
diff --git a/src/utils/sources-tree/tests/addToTree.spec.js b/src/utils/sources-tree/tests/addToTree.spec.js
index 74df6a5..de8c6cf 100644
--- a/src/utils/sources-tree/tests/addToTree.spec.js
+++ b/src/utils/sources-tree/tests/addToTree.spec.js
@@ -192,5 +192,37 @@ describe("sources-tree", () => {
       sources.forEach(source => addToTree(tree, source));
       expect(formatTree(tree)).toMatchSnapshot();
     });
+
+    it("uses debuggeeurl as the default", () => {
+      // this is a small subset of what we had in the test case
+      const sources = [
+        "components/TodoTextInput.js",
+        "components/Header.js",
+        "reducers/index.js",
+        "components/TodoItem.js",
+        "resource://gre/modules/ExtensionContent.jsm",
+        "https://voz37vlg5.codesandbox.io/static/js/components/TodoItem.js",
+        "index.js"
+      ];
+
+      // the snapshot should look something like this
+      `
+        - http://localhost:4242
+          - components
+            - Header.js
[...]

PR #4155, opened 2017-09-24
show files without domain in source tree

hughugsy:
Associated Issue: #3975 

### Summary of Changes

* If url does not have domain it gets debuggeeUrl  as default

### Test Plan

I am not so sure whether test passed or not...


jasonLaster:
looks good, i'll review thoroughly tomorrow

jasonLaster:
@hughugsy i'm seeing some errors here: `src/utils/sources-tree/tests/addToTree.spec.js`

PR #4094, opened 2017-09-19
Breakpoint contextmenu: add "Add/Edit breakpoint condition"

nyrosmith:
Associated Issue: NONE

@jasonLaster suggested this feature on slack

### Summary of Changes

* renamed contextmenu entry 
* jump to breakpoint location

conditional

PR #4157, opened 2017-09-25
Adding "Add condition" to breakpoints context menu

nyrosmith:
It was ways faster to add my changes on top of master than merging the existing ones. Especially all the debugger.properties changes were hard to track...
So closing https://github.com/devtools-html/debugger.html/pull/4094 in favor of this

### DIFF
--- a/assets/panel/debugger.properties
+++ b/assets/panel/debugger.properties
@@ -271,6 +271,8 @@ breakpointMenuItem.deleteAll2.label=Remove all
 breakpointMenuItem.deleteAll2.accesskey=a
 breakpointMenuItem.removeCondition2.label=Remove condition
 breakpointMenuItem.removeCondition2.accesskey=c
+breakpointMenuItem.addCondition2.label=Add condition
+breakpointMenuItem.addCondition2.accesskey=A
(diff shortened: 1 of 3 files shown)


jasonLaster:
Looks good, on first pass

assets/panel/debugger.properties
```diff
@@ -271,6 +271,8 @@ breakpointMenuItem.deleteAll2.label=Remove all
 breakpointMenuItem.deleteAll2.accesskey=a
 breakpointMenuItem.removeCondition2.label=Remove condition
 breakpointMenuItem.removeCondition2.accesskey=c
+breakpointMenuItem.addCondition2.label=Add condition
```
flodolo: Why addCondition2 and not just addCondition?

nyrosmith: I didn't know why the others got renamed I I just wanted to ensure consistency. Should I remove the "2" from my new entries @flodolo

jasonLaster: I'd keep the 2 to be consistent. We recently changed the labels to drop "breakpoint" e.g. `Remove breakpoint condition` to `Remove condition`...

PR #4159, opened 2017-09-26
No domain

jasonLaster:
Associated Issue: #4155 

This is a fork of #4155

### DIFF
--- a/src/utils/sources-tree/addToTree.js
+++ b/src/utils/sources-tree/addToTree.js
@@ -8,6 +8,8 @@ import {
   createNode
 } from "./utils";
 import { getURL, getFilenameFromPath } from "./getURL";
+import { parse } from "url";
+import { merge } from "lodash";
 
 import type { Node } from "./types";
 import type { SourceRecord } from "../../reducers/types";
@@ -108,11 +110,15 @@ export function addToTree(
   debuggeeUrl: string
 ) {
   const url = getURL(source.get("url"));
+  const { host } = parse(debuggeeUrl);
 
-  if (isInvalidUrl(url, source)) {
+  const updatedUrl =
+    url.group === "(no domain)" ? merge(url, { group: host }) : url;
+
+  if (isInvalidUrl(updatedUrl, source)) {
     return;
   }
 
-  const finalNode = traverseTree(url, tree);
-  finalNode.contents = addSourceToNode(finalNode, url, source);
+  const finalNode = traverseTree(updatedUrl, tree);
+  finalNode.contents = addSourceToNode(finalNode, updatedUrl, source);
 }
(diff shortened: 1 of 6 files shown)


jasonLaster:
Sorry @hughugsy i had to create a new PR because your PR was based off of master and i couldn't update it.

The tests should pass now, so just waiting on @codehag's review.

hughugsy:
No problem :+1:

package.json
```diff
@@ -30,9 +30,9 @@
     "lint-fix": "yarn lint-js -- --fix",
     "mochi":
       "mochii --mc ./firefox --default-test-path devtools/client/debugger/new",
-    "mochid": "yarn mochi -- --jsdebugger",
-    "mochir": "yarn mochi -- --repeat 10",
-    "mochih": "yarn mochi -- --setenv MOZ_HEADLESS=1",
+    "mochid": "yarn mochi -- --jsdebugger --",
```
codehag: why do we need the two dashes at the end?

src/utils/sources-tree/addToTree.js
```diff
@@ -107,7 +107,7 @@ export function addToTree(
   source: SourceRecord,
   debuggeeUrl: string
 ) {
-  const url = getURL(source.get("url"));
+  const url = getURL(source.get("url"), debuggeeUrl);
```
codehag: we might not need to do this anymore

src/utils/sources-tree/getURL.js
```diff
@@ -60,7 +64,7 @@ export function getURL(sourceUrl: string): { path: string, group: string } {
         // with a weird URL. Just group them all under an anonymous group.
         return merge(def, {
           path: url,
-          group: "(no domain)",
+          group: defaultDomain,
```
codehag: since this was fixed by devtools-html/devtools-core#698 we might not need this change. @jasonLaster thoughts?

jasonLaster: @codehag i still think it is a good practice to scope URLs. For instance, an eval w/ sourceURL, which is in one of the pages.

codehag: @jasonLaster cool, this is good to go from my side then

the `data` protocol issue can be investigated separately

package.json
jasonLaster: i'll move this to a separate PR, but the dashes keep the file passed in from being in quotes.

jasonLaster:
so... i don't know what to do w/ this test. it passes locally.

I'll look at it tomorrow mornign

codehag:
Looking into it as well!

jasonLaster:
Thanks @codehag!

Issue #2801, opened 2017-05-03
Enable React Perf Addons


jasonLaster:
As part of the reselect work we added `react-addons-perf`. When it is on it tries to get a lib that is not available. It would be great if we could get it working so we could take accurate measurements.

Perhaps the issue is that we're using react `15.3.x`

```diff
diff --git a/src/main.js b/src/main.js
index ade5ac3..63144b3 100644
--- a/src/main.js
+++ b/src/main.js
@@ -2,11 +2,11 @@

 const React = require("react");
 const ReactDOM = require("react-dom");
-//
-// if (process.env.NODE_ENV !== "production") {
-//   const Perf = require("react-addons-perf");
-//   window.Perf = Perf;
-// }
+
+if (process.env.NODE_ENV !== "production") {
+  const Perf = require("react-addons-perf");
+  window.Perf = Perf;
+}
```

asolove:
Alternately, upgrade the whole codebase to React 15.4.x so we can use the new user timing integration, which would be even better.

jasonLaster:
@codehag lets land prioritize this so we can detect these regressions sooner!

jasonLaster:
it's in!

PR #4160, opened 2017-09-26
[wip] bump react

jasonLaster:
Associated Issue: #2801

### Summary of Changes

* bumps react versions
* bumps launchpad to 99 (brings a bunch of changes) (a new config, themes namespacing)

The tests will fail because it is using a local 99, while i wait on the bump


jasonLaster:
still seeing test failures

Issue #4055, opened 2017-09-17
[breakpoint] Hidden brekpoints get restored as regular breakpoints


nyrosmith:
STR:

1. Open a website
2. make the debugger pause
3. choose "continue to here" at a location that won't get executed
4. refresh launchpad

![hiddenbreakpoint](https://user-images.githubusercontent.com/2511026/30518270-2a0a5000-9b79-11e7-83bf-2ae4f708504f.gif)

There might be other cases than "continue to here" too when this occurs.

I think we shouldn't persist hidden breakpoints at all because we use it for internal flow/pausing. I don't see a need for restoring them at all...
@jasonLaster your thoughts? I am already investigating...

jasonLaster:
Here is a gist for how i would do it:

```diff
diff --git a/src/actions/tests/pending-breakpoints.spec.js b/src/actions/tests/pending-breakpoints.spec.js
index c381039..3fa989f 100644
--- a/src/actions/tests/pending-breakpoints.spec.js
+++ b/src/actions/tests/pending-breakpoints.spec.js
@@ -94,6 +94,19 @@ describe("when adding breakpoints", () => {
       expect(pendingBps.get(breakpointLocationId2)).toMatchSnapshot();
     });
 
+    it("hidden breakponts do not create pending bps", async () => {
+      const { dispatch, getState } = createStore(simpleMockThreadClient);
+
+      await dispatch(actions.newSource(makeSource("foo")));
+      await dispatch(
+        actions.addBreakpoint(breakpoint1.location, { hidden: true })
+      );
+      const pendingBps = selectors.getPendingBreakpoints(getState());
+
+      // should be null
+      expect(pendingBps.get(breakpointLocationId1)).toBe(null);
+    });
+
     it("remove a corresponding pending breakpoint when deleting", async () => {
       const { dispatch, getState } = createStore(simpleMockThreadClient);
       await dispatch(actions.newSource(makeSource("foo")));
diff --git a/src/reducers/pending-breakpoints.js b/src/reducers/pending-breakpoints.js
index 0cd7419..be85711 100644
--- a/src/reducers/pending-breakpoints.js
+++ b/src/reducers/pending-breakpoints.js
@@ -31,6 +31,7 @@ export type PendingBreakpointsState = {
 export function initialState(): Record<PendingBreakpointsState> {
   return makeRecord(
     ({
+      // this is where we load the pending breakpoints when the debugger opens
[...]

PR #4163, opened 2017-09-26
Don't persist pending breakpoints if hidden

nyrosmith:
Associated Issue: #4055 

### Summary of Changes

* adopted pending_breakpoint reducer to not handle adding or removing hidden breakpoints
* added a test, but was not able to run it on my machine

### Test Plan
- [x] Open a website
- [x] make the debugger pause
- [x] choose "continue to here" at a location that won't get executed
- [x] refresh launchpad (before there was a regular breakpoint at the location of the hidden one)

### DIFF
--- a/src/reducers/pending-breakpoints.js
+++ b/src/reducers/pending-breakpoints.js
@@ -42,6 +42,9 @@ function update(
 ) {
   switch (action.type) {
     case "ADD_BREAKPOINT": {
+      if (action.breakpoint.hidden) {
+        return state;
+      }
       return addBreakpoint(state, action);
     }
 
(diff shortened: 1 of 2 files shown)


src/reducers/pending-breakpoints.js
```diff
@@ -62,6 +65,9 @@ function update(
     }
 
     case "REMOVE_BREAKPOINT": {
+      if (action.breakpoint.hidden) {
```
codehag: I am not sure I understand why we are doing a no-op here. I think we might want to be able to remove hidden breakpoints. For example, if we have a continue-to-here breakpoint, we will probably want to remove it immediately so that it doesnt trigger the next time it runs, as it is not a user added breakpoint.

Thinking about it a bit further -- perhaps these hidden breakpoints should be "single execution".

Thoughts?

codehag: Ah, i just realized that this is persisted-breakpoints reducer. i was thinking of this as the breakpoints reducer. ignore this comment!

Issue #3966, opened 2017-09-10
[Preview] show small values in a tooltip


jasonLaster:
It would be nice if simple values like strings and numbers that are smaller than 10 characters used the tooltip.

See the GIF below. `0` and `" "` have the popup when they'd look good in the tooltip.

Here is a small sample

```diff
diff --git a/src/components/Editor/Preview/Popup.js b/src/components/Editor/Preview/Popup.js
index 8c4057a..7cf0845 100644
--- a/src/components/Editor/Preview/Popup.js
+++ b/src/components/Editor/Preview/Popup.js
@@ -164,6 +164,15 @@ export class Popup extends Component {
       return this.renderFunctionPreview(value, root);
     }

+    // we can check here to see what we want to do...
+
+    const isSmallString = value.type == "string" && value.x.length < 10;
+    const isSmallNumber = value.type == "number" && value.x.toString().length < 10;
+
+    if (isSmallNumber || isSmallString) {
+      return this.renderSimplePreview(value);
+    }
+
     if (value.type === "object") {
       return (
         <div>
```

![](https://user-images.githubusercontent.com/2511026/30241412-38483cfa-9583-11e7-8ed6-f9a7fdc1b315.gif)

jasonLaster:
@nyrosmith is on it

jasonLaster:
actually, on the other hand maybe we don't want to show preview at all for these because you just see them:

https://brash-success.glitch.me/

nyrosmith:
@jasonLaster we already use simplePreview for strings, numbers, bools, ...
To me it seems the reason why eg number preview is rendered with a higher width is because of the corresponding reps template   in devtools-core

jasonLaster:
I guess, there are two things here:

1. maybe we shouldn't show anything if you hover on `42` or `"hello"`
2. if you hover on `num` we should show `42` in a tooltip

jasonLaster:
Tohm is working on small strings

PR #4164, opened 2017-09-26
Preview: Show numbers inside a tooltip

nyrosmith:
Associated Issue: #3966


codehag:
Do you have an STR I can test for this specific case? its not clear from the issue. Thanks!

PR #4178, opened 2017-09-27
Fix ESLint Dependencies from Fresh Install

darkwing:
So I just got my 2 year Mozilla laptop refresh and I can't lint with debugger.html "out of the box" because I'm missing `eslint-plugin-mozilla`.

This PR could be totally wrong somehow, or we've all had moz-central, and we've taken advantage of having stuff installed already.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,16 +18,19 @@
     "start": "node bin/dev-server",
     "start-app": "TARGET=application node bin/dev-server",
     "flow": "flow",
-    "eslint-check": "eslint --print-config .eslintrc.js | eslint-config-prettier-check",
+    "eslint-check":
+      "eslint --print-config .eslintrc.js | eslint-config-prettier-check",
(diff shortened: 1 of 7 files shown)


darkwing:
Let me know if this isn't necessary.  Man, the things I took advantage of with my old machine.  This just worked.

package.json
```diff
@@ -76,6 +77,9 @@
     "devtools-source-map": "^0.13.0",
     "devtools-splitter": "^0.0.3",
     "devtools-utils": "^0.0.9",
+    "eslint-plugin-babel": "^4.1.2",
+    "eslint-plugin-flowtype": "^2.36.0",
+    "eslint-plugin-mozilla": "^0.4.4",
```
darkwing: These 3 additions are the real change.

codehag:
Copy from slack message: these are dev dependancies in lauch pad, and there might be version mismatches!

PR #4179, opened 2017-09-27
Don't show "No sources" hint if sources panel is hidden

nyrosmith:
Associated Issue: NONE

While fiddling around I found that the "No sources" hint gets displayed even when the sources panel is hidden.

STR:

- Open debuggee with "about:blank"
- Switch from sources panel to outline

Before:
sources_before

After:
sources_after

### DIFF
--- a/src/components/PrimaryPanes/Outline.js
+++ b/src/components/PrimaryPanes/Outline.js
@@ -3,7 +3,6 @@
 import React, { Component } from "react";
 import { bindActionCreators } from "redux";
 import { connect } from "react-redux";
-import classnames from "classnames";
 import actions from "../../actions";
 import { getSelectedSource, getSymbols } from "../../selectors";
 import "./Outline.css";
--- a/src/components/PrimaryPanes/Outline.js
+++ b/src/components/PrimaryPanes/Outline.js
@@ -20,7 +20,6 @@ export class Outline extends Component {
   state: any;
 
   props: {
-    isHidden: boolean,
     symbols: SymbolDeclarations,
     selectSource: (string, { line: number }) => void,
     selectedSource: ?SourceRecord
@@ -67,14 +66,14 @@ export class Outline extends Component {
   }
 
   render() {
-    const { isHidden, symbols } = this.props;
+    const { symbols } = this.props;
 
     const symbolsToDisplay = symbols.functions.filter(
       func => func.name != "anonymous"
     );
 
     return (
-      <div className={classnames("outline", { hidden: isHidden })}>
+      <div className="outline">
         {symbolsToDisplay.length > 0
           ? this.renderFunctions(symbolsToDisplay)
           : this.renderPlaceholder()}
--- a/src/components/PrimaryPanes/Sources.css
+++ b/src/components/PrimaryPanes/Sources.css
@@ -67,14 +67,6 @@
   flex: 1;
 }
 
-.sources-panel .outline.hidden {
-  display: none;
-}
-
-.hidden {
-  display: none;
-}
-
 .source-outline-tabs {
   width: 100%;
   background: var(--theme-body-background);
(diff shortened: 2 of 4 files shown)


codehag:
With this solution, we are relying on the sourceTree being empty. I think a better solution here would be to not render the sourcesTree component at all if it is not being displayed, rather than rendering and hiding it.

Here is one way of doing it:

```diff
diff --git a/src/components/PrimaryPanes/Outline.js b/src/components/PrimaryPanes/Outline.js
index 6719f408..abb87aa8 100644
--- a/src/components/PrimaryPanes/Outline.js
+++ b/src/components/PrimaryPanes/Outline.js
@@ -20,7 +20,6 @@ export class Outline extends Component {
   state: any;
 
   props: {
-    isHidden: boolean,
     symbols: SymbolDeclarations,
     selectSource: (string, { line: number }) => void,
     selectedSource: ?SourceRecord
@@ -67,14 +66,14 @@ export class Outline extends Component {
   }
 
   render() {
-    const { isHidden, symbols } = this.props;
+    const { symbols } = this.props;
 
     const symbolsToDisplay = symbols.functions.filter(
       func => func.name != "anonymous"
     );
 
     return (
-      <div className={classnames("outline", { hidden: isHidden })}>
+      <div className={classnames("outline")}>
         {symbolsToDisplay.length > 0
           ? this.renderFunctions(symbolsToDisplay)
           : this.renderPlaceholder()}
diff --git a/src/components/PrimaryPanes/SourcesTree.js b/src/components/PrimaryPanes/SourcesTree.js
index 9835d7a6..cae7e832 100644
--- a/src/components/PrimaryPanes/SourcesTree.js
+++ b/src/components/PrimaryPanes/SourcesTree.js
@@ -249,7 +249,7 @@ class SourcesTree extends Component {
   }
 
   render() {
-    const { isHidden, setExpandedState, expanded } = this.props;
+    const { setExpandedState, expanded } = this.props;
     const {
       focusedItem,
       sourceTree,
@@ -294,10 +294,7 @@ class SourcesTree extends Component {
     };
 
     return (
-      <div
-        className={classnames("sources-list", { hidden: isHidden })}
-        onKeyDown={onKeyDown}
-      >
+      <div className={classnames("sources-list")} onKeyDown={onKeyDown}>
         {tree}
       </div>
     );
@@ -305,7 +302,6 @@ class SourcesTree extends Component {
 }
 
 SourcesTree.propTypes = {
-  isHidden: PropTypes.bool,
   sources: ImPropTypes.map.isRequired,
   selectSource: PropTypes.func.isRequired,
   shownSource: PropTypes.string,
diff --git a/src/components/PrimaryPanes/index.js b/src/components/PrimaryPanes/index.js
index 1b00def4..34aadc2b 100644
--- a/src/components/PrimaryPanes/index.js
+++ b/src/components/PrimaryPanes/index.js
@@ -109,22 +109,17 @@ class PrimaryPanes extends Component {
     const { selectedPane } = this.state;
     const { sources, selectSource } = this.props;
 
-    const outlineComp = isEnabled("outline") ? (
-      <Outline
-        selectSource={selectSource}
-        isHidden={selectedPane === "sources"}
-      />
-    ) : null;
+    function renderSelectedPanel(panel) {
+      if (isEnabled("outline") && selectedPane === "outline") {
+        return <Outline selectSource={selectSource} />;
+      }
+      return <SourcesTree sources={sources} selectSource={selectSource} />;
+    }
 
     return (
       <div className="sources-panel">
         {this.renderTabs()}
-        <SourcesTree
-          sources={sources}
-          selectSource={selectSource}
-          isHidden={selectedPane === "outline"}
-        />
-        {outlineComp}
+        {renderSelectedPanel(selectedPane)}
       </div>
     );
   }
```

nyrosmith:
I did this change in another PR but the problem is that if you switch to outline and then back the tree looses its state and all nodes are collapsed

PR #4188, opened 2017-09-27
Fix broken mochitests

jasonLaster:
### Summary of Changes

When we removed the integration-test.js file from MC @juliandescottes found some [discrepencies][d]

[d]: https://bugzilla.mozilla.org/show_bug.cgi?id=1401659

* asm was fixed prior
*  breaking.js looks fine

### DIFF
--- a/src/test/mochitest/browser_dbg-returnvalues.js
+++ b/src/test/mochitest/browser_dbg-returnvalues.js
@@ -1,0 +1,61 @@
+function getLabel(dbg, index) {
+  return findElement(dbg, "scopeNode", index).innerText;
+}
+
+function getValue(dbg, index) {
+  return findElement(dbg, "scopeValue", index).innerText;
+}
+
+function toggleScopes(dbg) {
+  return findElement(dbg, "scopesHeader").click();
+}
+
+async function testReturnValue(dbg, val) {
+  invokeInTab(dbg, "return_something",val);
+  await waitForPaused(dbg);
+
+  // "Step in" 3 times to get to the point where the debugger can
+  // see the return value.
+  await stepIn(dbg);
+  await stepIn(dbg);
+  await stepIn(dbg);
+  is(getLabel(dbg, 1), "return_something", "check for return_something");
+
+  // We don't show "undefined" but we do show other falsy values.
+  let label = getLabel(dbg, 2);
+  if (val === "undefined") {
+    ok(label !== "<return>", "do not show <return> for undefined");
+  } else {
+    is(label, "<return>", "check for <return>");
+    is(getValue(dbg, 2), val, `check value is ${val}`);
+  }
+
+  await resume(dbg);
+  assertNotPaused(dbg);
+}
+
+// async function testThrowValue(dbg, val) {
+//   invokeInTab(dbg, "throw_something", val);
+//   await waitForPaused(dbg);
+//
+//   // "Step in" to get to the exception.
+//   await stepIn(dbg);
+//
+//   is(getLabel(dbg, 1), "callee", "check for callee");
+//   is(getLabel(dbg, 2), "<exception>", "check for <exception>");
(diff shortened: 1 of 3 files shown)


codehag:
We have a lot of failures -- looks like something strange is happening on the CI

src/test/mochitest/head.js
```diff
@@ -617,10 +617,17 @@ function togglePauseOnExceptions(
  * @return {Promise}
  * @static
  */
-function invokeInTab(fnc) {
+function invokeInTab(fnc, args) {
   info(`Invoking function ${fnc} in tab`);
-  return ContentTask.spawn(gBrowser.selectedBrowser, fnc, function*(fnc) {
-    content.wrappedJSObject[fnc](); // eslint-disable-line mozilla/no-cpows-in-tests, max-len
+  return ContentTask.spawn(gBrowser.selectedBrowser, fnc, function*(fnc, args) {
+    content.wrappedJSObject.apply(func, args); // eslint-disable-line mozilla/no-cpows-in-tests, max-len
```
darkwing: Should this be `fnc`, not `func`?

darkwing: OK, so there are two problems here:

1.  It should be `fnc`, not `func`
2.  `TypeError: content.wrappedJSObject.apply is not a function` bricks any test that uses `invokeInTab`.  So, for example, if I dumb it way down with the following, all tests run:

```js
function invokeInTab(fnc, args) {
  info(`Invoking function ${fnc} in tab`);
  return ContentTask.spawn(gBrowser.selectedBrowser, fnc, function*(fnc, args) {
    // content.wrappedJSObject.apply(fnc, args); // eslint-disable-line mozilla/no-cpows-in-tests, max-len
    console.log('invoking in tab: ', fnc, args);

    // FIX HERE
    if(args) {
      content.wrappedJSObject[fnc](...args);
    }
    else {
      content.wrappedJSObject[fnc]();
    }
  });
}
```

src/test/mochitest/browser_dbg-returnvalues.js
```diff
@@ -0,0 +1,61 @@
...
+function getValue(dbg, index) {
+  return findElement(dbg, "scopeValue", index).innerText;
+}
+
+function toggleScopes(dbg) {
+  return findElement(dbg, "scopesHeader").click();
+}
+
+async function testReturnValue(dbg, val) {
+  invokeInTab(dbg, "return_something",val);
```
darkwing: We're sending 3 args here, while `invokeInTab` only accepts two; don't think we need `dbg`.

darkwing: ...actually, simply removing that doesn't fix test failures.  Still looking into it.

src/test/mochitest/head.js
```diff
@@ -617,10 +617,17 @@ function togglePauseOnExceptions(
...
+  });
+}
+
+function evalInTab(script) {
```
darkwing: This function doesn't appear to be used -- can we get rid of this?

jasonLaster: yeah, lets kill it

jasonLaster:
@darkwing can you push some of your changes...

jasonLaster:
Closing and opening issue #4308

src/test/mochitest/head.js
jimblandy: A call like `ContentTask.spawn(browser, arg, task)` only passes `task` a single argument, which is a structured clone of `arg`. So I think we could do something like this:
```
function invokeInTab(fnc, ...args) {
  info(`Invoking in tab: ${fnc}(${args.map(uneval).join(',')})`);
  return ContentTask.spawn(gBrowser.selectedBrowser, { fnc, args }, function*({ fnc, args }) {
    content.wrappedJSObject[fnc](...args); // eslint-disable-line mozilla/no-cpows-in-tests, max-len
  });
}

```

Issue #3715, opened 2017-08-18
[Breakpoints] Another, unwanted, breakpoint is created while holding ctrl + B


jasonLaster:
[Steps to reproduce]:
1.Launch Firefox.
2.Access the https://devtools-html.github.io/debugger-examples/examples/todomvc/ webpage.
3.Open the debugger.
4.Hold the CTRL + B combination on any line of code.

[Expected result]:
An breakpoint is constantly created and removed.

[Actual result]:
Another breakpoint is created on a different line.

![Breakpoint issue.gif](https://shipusercontent.com/9ed2855adb439ff6e8201e5056de2ed4/Breakpoint%20issue.gif)

MayankSrivastav:
Hi, Trying to work on this one. I have reproduced the issue. Seems like some problem with keybindings? Not sure though.
Can you please guide me a little?
Thanks

jasonLaster:
@MayankSrivastav I think this might actually be a debugger server bug. 

The keybinding is in `Editor/index.js` or Editor/Breakpoints

jasonLaster:
re-opening it because i can still repro

PR #4191, opened 2017-09-27
Fix breakpoint jumping

jasonLaster:
Associated Issue: #3715

### Summary of Changes

the jumping comes from trying to remove a breakpoint when the client does not have it registered yet.

### Test Plan

held cmd+b down for 3 minutes. Created literally thousands of breakpoints at the same line


src/client/firefox/commands.js
```diff
@@ -132,7 +132,7 @@ function removeBreakpoint(generatedLocation: Location) {
     const bpClient = bpClients[id];
     if (!bpClient) {
       console.warn("No breakpoint to delete on server");
-      return;
+      return Promise.resolve();
```
codehag: nice catch :+1:

Issue #4112, opened 2017-09-21
[wasm] mochitest is intermittent


jasonLaster:
I've seen the wasm test fail a couple times recently. We should investigate

```
TEST-START browser_dbg-wasm-sourcemaps.js
   Adding a new tab with URL: http://example.com/browser/devtools/client/debugger/new/test/mochitest/examples/doc-wasm-sourcemaps.html
   Waiting for thread event 'paused' to fire.
   Thread event 'paused' fired.
 JavaScript warning: resource://devtools/shared/base-loader.js -> resource://devtools/server/actors/object.js, line 1637: DebuggeeWouldRun: debuggee `self-hosted:6535' would run
   Console message: [JavaScript Warning: "DebuggeeWouldRun: debuggee `self-hosted:6535' would run" {file: "resource://devtools/shared/base-loader.js -> resource://devtools/server/actors/object.js" line: 1637}]
 TEST-PASS top frame's source is selected -
 TEST-PASS Line is highlighted as paused -
 TEST-PASS expression is highlighted as paused -
   Waiting for ADD_BREAKPOINT to dispatch 1 time(s)
   ADD_BREAKPOINT dispatched 1 time(s)
   Waiting for thread event 'paused' to fire.
   Thread event 'paused' fired.
 TEST-UNEXPECTED-FAIL Test timed out 
 TEST-UNEXPECTED-FAIL base-loader.js 
> resource://devtools/client/debugger/new/debugger.js:31726:240
step/<@resource://devtools/shared/base-loader.js -> resource://devtools/client/debugger/new/debugger.js:31726:402
promise callback*step@resource://devtools/shared/base-loader.js -> resource://devtools/client/debugger/new/debugger.js:31726:314
promise callback*step@resource://devtools/shared/base-loader.js -> resource://devtools/client/debugger/new/debugger.js:31726:314
promise callback*step@resource://devtools/shared/base-loader.js -> resource://devtools/client/debugger/new/debugger.js:31726:314
_asyncToGenerator/</<@resource://devtools/shared/base-loader.js -> resource://devtools/client/debugger/new/debugger.js:31726:437
_asyncToGenerator/<@resource://devtools/shared/base-loader.js -> resource://devtools/client/debugger/new/debugger.js:31726:99
paused/</<@resource://devtools/shared/base-loader.js -> resource://devtools/client/debugger/new/debugger.js:31823:14
[...]

PR #4195, opened 2017-09-27
fix wasm intermittent

jasonLaster:
Associated Issue: #4112

### Summary of Changes

our hypothesis is that we try to add a breakpoint before the source is available...


codehag:
resolved yarn.lock conflict: 
devtools-map-bindings@0.1.0 -> devtools-map-bindings@0.2.0

codehag:
we still have the intermittent here :/

codehag:
Beautiful.

Issue #3666, opened 2017-08-15
[Tabs] Create new tabs action


bomsy:
We should pull out the tab actions from the sources actions file

jasonLaster:
Hey @bomsy can you add some more context here?

jasonLaster:
closing as i'm doing this with a sources actions refactor

Issue #3646, opened 2017-08-12
[Tabs] create a new tabs reducer


jasonLaster:
Tabs are currently in the sources reducer, but they are interesting enough to be in their own reducer

https://github.com/aklt/debugger.html/blob/fix-text-search/src/reducers/sources.js#L109-L127

jasonLaster:
closing as a dupe of #3279

Issue #3601, opened 2017-08-08
[blackbox] blackbox and pretty print sources should be persisted


jasonLaster:
It would be great if we could persist blackbox state like we persist tabs. This way, when you come back, the same sources are still blackboxed.

codehag:
This is a start on this issue:
```diff
diff --git a/src/actions/sources.js b/src/actions/sources.js
index 5e644db3..6027e169 100644
--- a/src/actions/sources.js
+++ b/src/actions/sources.js
@@ -92,6 +92,7 @@ export function newSource(source: Source) {
     }
 
     await checkSelectedSource(getState(), dispatch, source);
+    await checkPendingTabs(getState(), dispatch, source); // this should restore blackbox state for sources
     await checkPendingBreakpoints(getState(), dispatch, source);
   };
 }
@@ -244,10 +245,10 @@ export function addTab(source: Source, tabIndex: number) {
   };
 }
 
-export function moveTab(url: string, tabIndex: number) {
+export function moveTab(source: Source, tabIndex: number) {
   return {
     type: "MOVE_TAB",
-    url,
+    source,
     tabIndex
   };
 }
diff --git a/src/components/Editor/Tabs.js b/src/components/Editor/Tabs.js
index 2004e4d8..a5add58b 100644
--- a/src/components/Editor/Tabs.js
+++ b/src/components/Editor/Tabs.js
@@ -301,7 +301,7 @@ class SourceTabs extends PureComponent {
     const hiddenSourceTabs = getHiddenTabs(sourceTabs, sourceTabEls);
 
     if (isVisible() && hiddenSourceTabs.indexOf(selectedSource) !== -1) {
-      return moveTab(selectedSource.get("url"), 0);
+      return moveTab(selectedSource.toJS(), 0);
     }
 
     this.setState({ hiddenSourceTabs });
@@ -319,7 +319,7 @@ class SourceTabs extends PureComponent {
 
     const onClick = () => {
       const tabIndex = 0;
-      moveTab(source.get("url"), tabIndex);
+      moveTab(source.toJS(), tabIndex);
     };
     return (
       <li key={source.get("id")} onClick={onClick}>
diff --git a/src/reducers/sources.js b/src/reducers/sources.js
index ddb14e34..929a69dd 100644
--- a/src/reducers/sources.js
+++ b/src/reducers/sources.js
@@ -102,12 +102,12 @@ function update(
 
     case "ADD_TAB":
       return state.merge({
[...]

PR #4199, opened 2017-09-28
(WIP) add the  tabs reducer

bomsy:
Associated Issue: #3666, #3646

WIP - Separating the Tabs from sources redux

### Summary of Changes

* Added the idea or what the tabs reducer will look like
* Todo - Add the tabs action

@jasonLaster , @codehag thoughts?


src/reducers/tabs.js
```diff
@@ -0,0 +1,107 @@
...
+import * as I from "immutable";
+import makeRecord from "../utils/makeRecord";
+import { prefs } from "../utils/prefs";
+
+import type { List } from "immutable";
+import type { Action } from "../actions/types";
+import type { Record } from "../utils/makeRecord";
+
+type Tab = {
+  id: string,
```
bomsy: `id` must be unique.

nyrosmith:
https://github.com/devtools-html/debugger.html/issues/3601 should also go into the new reducer so I'll start as soon as this gets merged

src/reducers/tabs.js
```diff
@@ -0,0 +1,107 @@
...
+    return;
+  }
+  return prefsTabs;
+}
+
+type OuterState = { tabs: Record<TabsState> };
+
+const getTabsState = state => state.tabs;
+
+const getTabs = createSelector(getTabsState, tabs => tabs.tabs);
```
jasonLaster: this is looking good.

src/reducers/tabs.js
```diff
@@ -0,0 +1,107 @@
...
+import type { Action } from "../actions/types";
+import type { Record } from "../utils/makeRecord";
+
+type Tab = {
+  id: string,
+  title: string,
+  tooltip?: string
+};
+
+type TabList = List<Tab>;
```
codehag: Since we do not use movetab unless it is a hidden tab (and in that case, always moving it to index 0), i am wondering how important moving tabs is? Perhaps, if the uniqueness of the ID is more important, we can leverage orderedSet instead, with unique keys (the id being the key)? Since the id is not data being used in the tab itself, and to identify it in the group, this might make a few things easier.

src/reducers/tabs.js
```diff
@@ -0,0 +1,107 @@
...
+    case "MOVE_TAB":
+      const tab = state.tabs.find(tab => tab.id === id);
+      return state.merge({
+        tabs: updateTabList({ tabsState: state }, tab, action.tabIndex)
+      });
+
+    case "CLOSE_TAB":
+      const tabs = removeFromTabList({ tabsState: state }, [action.id]);
+      prefs.tabs = tabs;
+      return state.merge({ tabs });
```
codehag: jut curious, why are we prefering merge here to set?

src/reducers/tabs.js
```diff
@@ -0,0 +1,107 @@
...
+
+  return state;
+};
+
+/**
+ * Adds the new tab to the list or moves the tab in the list if it is not already there
+ * @memberof reducers/sources
+ * @static
+ */
+function updateTabList(state: OuterState, currentTab: Tab, moveIndex?: number) {
```
codehag: this function is doing two things, moving and adding -- lets split this into two functions.

src/reducers/tabs.js
```diff
@@ -0,0 +1,107 @@
...
+
+    case "MOVE_TAB":
+      const tab = state.tabs.find(tab => tab.id === id);
+      return state.merge({
+        tabs: updateTabList({ tabsState: state }, tab, action.tabIndex)
+      });
+
+    case "CLOSE_TAB":
+      const tabs = removeFromTabList({ tabsState: state }, [action.id]);
+      prefs.tabs = tabs;
```
codehag: I think this is immutable?

src/reducers/tabs.js
```diff
@@ -0,0 +1,107 @@
...
+}
+
+export function removeFromTabList(state: OuterState, tabIds: Array<string>) {
+  let tabs = state.tabsState.get("tabs");
+  return tabIds.reduce((tabs, id) => tabs.filter(tab => tab != id)), tabs);
+}
+
+function restoreTabs() {
+  const prefsTabs = prefs.tabs || [];
+  if (prefsTabs.length == 0) {
```
codehag: i think we can always `return prefs.tabs || []` ?

src/reducers/tabs.js
codehag: also, i investigated when we move tabs... it is rare. is there any information on why we have so much infrastructure for moving things?

bomsy: I think it was done initially, in the bid to cover all the possible moving scenarios, now we can make more informed improvements!

src/reducers/tabs.js
bomsy: yeah your right we do not do much with move tabs, and i'm in favor of removing it. maybe add tabs can handle moving if a tab already exists. An orderedSet might be cool, will look into it a bit more.

jasonLaster:
closing it as it's a month old. Feel free to re-open if you want to take a stab at it

Issue #4122, opened 2017-09-22
[Accessibility] Show a shortcuts modal


jasonLaster:
It would be great from an accessibility perspective if we documented our shortcuts and mouse tricks:

* alt-click to continue to a location
* cmd*b to toggle a bp

Notes:

* we could have a button in the bottom right to launch the modal
* we could use this modal pattern for other things in the future
* we could share this modal w/ other tools so that other tools can have shortcut modals too :)

### Single Modal

<img src='https://shipusercontent.com/cde06027ecd2aba3422cd820a023f572/Screen%20Shot%202017-09-21%20at%209.27.27%20PM.png' title='Screen Shot 2017-09-21 at 9.27.27 PM.png' width=617>

### Modal w/ a sidebar

<img src='https://shipusercontent.com/b43011a5732568790f9700204935cf9f/Screen%20Shot%202017-09-21%20at%209.27.17%20PM.png' title='Screen Shot 2017-09-21 at 9.27.17 PM.png' width=571>

violasong:
This looks great! I think I prefer the UI with the single modal - while the sidebar makes it look nicer, I suspect shortcut-hunting is easier when it's all laid out. 

I also recommend using the main body color and larger/darker icons - I only changed the first row but hopefully this is clear enough :).

<img width="852" alt="image" src="https://user-images.githubusercontent.com/55994/30822975-9ab5ff7c-a1df-11e7-80ee-7c486667a416.png">

jasonLaster:
:fire:

PR #4202, opened 2017-09-28
Shortcuts style tweak

samuel-peers:
Associated Issue: #4122

### Summary of Changes

Tiny style changes suggested by @violasong

### Mac
whatever

### Windows
whatever

### DIFF
--- a/src/components/App.js
+++ b/src/components/App.js
@@ -18,6 +18,11 @@ import { isVisible } from "../utils/ui";
 import { KeyShortcuts } from "devtools-modules";
 const shortcuts = new KeyShortcuts({ window });
 
+import { Services } from "devtools-modules";
+const { appinfo } = Services;
+
+const isMacOS = appinfo.OS === "Darwin";
+
 const verticalLayoutBreakpoint = window.matchMedia("(min-width: 800px)");
 
 import "./variables.css";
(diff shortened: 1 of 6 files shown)


jasonLaster:
Thanks it's looking a lot better. Can we try a couple of style tweaks like the box around the keys. Also I think the wider two column UI could be nice.

If you want to try only doing the box around keys on mac we can do something in App.js (`navigator.userAgent.indexOf('Mac OS X') != -1`) where we add a `mac` class to the `.debugger` element.

samuel-peers:
you got it bro. There are travis tests failing again :/ `failed test:  devtools/client/debugger/new/test/mochitest/browser_dbg-wasm-sourcemaps.js`

jasonLaster:
sweet. that test ain't you. we're on it

samuel-peers:
### Mac
whatever_mac

### Windows
whatever_pc

jasonLaster:
also some other small things:

1. we should add a shortcut for launching it `cmd+/` i think
2. we should move K+ to the right and soften the color (whatever the accordion headers are)
3. the section headers in the mocks looked pretty nice. Lets do some small typography cleanup.

samuel-peers:
stacked

jasonLaster:
I'll make a new issue for the follow up CSS changes and some of the meetup new comers can work on it

Issue #4201, opened 2017-09-28
[Mapping] map watch expressions


jasonLaster:
It would be nice to map watch expressions when we're paused.

Here is a sketch that re-purposes what we do when we [preview][p]

[p]: https://github.com/devtools-html/debugger.html/blob/master/src/actions/ast.js#L180
```diff
diff --git a/src/actions/expressions.js b/src/actions/expressions.js
index d20a485..15751e7 100644
--- a/src/actions/expressions.js
+++ b/src/actions/expressions.js
@@ -87,14 +87,35 @@ export function evaluateExpressions() {
 }

 function evaluateExpression(expression: Expression) {
-  return function({ dispatch, getState, client }: ThunkArgs) {
+  return function({ dispatch, getState, client, sourceMaps }: ThunkArgs) {
     if (!expression.input) {
       console.warn("Expressions should not be empty");
       return;
     }

     const frameId = getSelectedFrameId(getState());
-    const input = wrapExpression(expression.input);
+
+
+    let expression;
+    if (isPaused()) {
+      const location = getSelectedLocation(state);
+      const source = getSelectedSource(state);
+      const generatedLocation = await sourceMaps.getGeneratedLocation(
+        { ...location.start, sourceId },
+        source.toJS()
+      );
+
+      expression = await getSourcemapedExpression(
+        { dispatch, sourceMaps },
+        generatedLocation,
+        expression.input
+      );
+    } else {
+      expression = expression.input
+    }
+
+    const input = wrapExpression(expression);
+
     return dispatch({
       type: "EVALUATE_EXPRESSION",
       input: expression.input,
(END)
```

PR #4206, opened 2017-09-29
Show mapped names in watch expressions

amitzur:
Associated Issue: #4201

### Summary of Changes

* in `evaluateExpression` action, in case there's a frame (we're paused) and it's a generated source, get the mapped expression and evaluate it instead. But still show the original expression in the UI, of course.

* refactor `getSourceMappedExpression` out of `ast.js` (and also a slight name fix to the function name).

### Test Plan

getSourceMappedExpression should probably get tested. I need to be mentored on how to go about that.

### Screenshots
screen shot 2017-09-29 at 4 45 24 am

### DIFF
--- a/src/utils/parser/getSourceMappedExpression.js
+++ b/src/utils/parser/getSourceMappedExpression.js
@@ -1,0 +1,25 @@
+import { getScopes } from ".";
+import { replaceOriginalVariableName } from "devtools-map-bindings/src/utils";
+
+/**
+ * Gets information about original variable names from the source map
+ * and replaces all posible generated names.
+ */
+export default async function getSourceMappedExpression(
+  { sourceMaps },
+  generatedLocation: Location,
+  expression: string
+): Promise<string> {
+  const astScopes = await getScopes(generatedLocation);
+
+  const generatedScopes = await sourceMaps.getLocationScopes(
+    generatedLocation,
+    astScopes
+  );
+
+  if (!generatedScopes) {
+    return expression;
+  }
+
+  return replaceOriginalVariableName(expression, generatedScopes);
+}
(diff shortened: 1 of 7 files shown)


src/utils/parser/getSourceMappedExpression.js
```diff
@@ -0,0 +1,25 @@
+import { getScopes } from ".";
+import { replaceOriginalVariableName } from "devtools-map-bindings/src/utils";
+
+/**
+ * Gets information about original variable names from the source map
+ * and replaces all posible generated names.
+ */
+export default async function getSourceMappedExpression(
```
codehag: Is this tested?

jasonLaster: Good question, I've thought a lot about testing this function and I don't think we should write a unit test for it. Let me explain:

1. we have a mochitest for the mapping in scopes
2. we have individual unit tests for all three calls. 

This function is the classic glue/contract function. If we were to test it i'd want to stub the calls so as not to repeat a test, at that point it will be a test that `replaceOriginalVariableName` is called w/ expression & generatedScopes

src/utils/parser/getSourceMappedExpression.js
```diff
@@ -0,0 +1,25 @@
...
+    generatedLocation,
+    astScopes
+  );
+
+  if (!generatedScopes) {
+    return expression;
+  }
+
+  return replaceOriginalVariableName(expression, generatedScopes);
+}
```
jasonLaster: we should move this function out of utils/parser because it's not in the worker

jasonLaster:
one failing test and some lint errors

amitzur:
I got this. Are we going with this or #4207

jasonLaster:
We're landing this first

amitzur:
The failing test was because of using `getSelectedFrame` and not `getSelectedFrameId`. So the actual frames are needed in state, which also depends on the actual source in state. The last commit takes care of that in the test

yurydelendik:
> The failing test was because of using getSelectedFrame and not getSelectedFrameId.

Some rebase was done. Please make sure the above change was not lost.

PR #4218, opened 2017-09-29
Issue #3966: Show small values inside tooltip

tohmjudson:
Associated Issue: #3966 

### Summary of Changes

* Numbers and string < 10 letters moved to tooltip
* Removed mouseover events from value assignments in strings, numbers, and "atom" items (null)

ovaikxwzuw

### DIFF
--- a/src/components/Editor/Preview/Popup.js
+++ b/src/components/Editor/Preview/Popup.js
@@ -179,6 +179,8 @@ export class Popup extends Component {
   getPreviewType(value: any) {
     if (
       typeof value == "boolean" ||
+      (typeof value == "string" && value.length < 10) ||
+      (typeof value == "number" && value.toString().length < 10) ||
       value.type == "null" ||
       value.type == "undefined" ||
       value.class === "Function"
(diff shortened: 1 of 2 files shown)


src/utils/editor/expression.js
```diff
@@ -38,14 +38,22 @@ export function updatePreview(
...
+
   const isUpdating = preview && preview.updating;
+
   const inScope = linesInScope && linesInScope.includes(location.line);
 
-  if (invalidTarget || !inScope || isUpdating || invalidToken) {
+  const invaildType =
+    target.className === "cm-string" ||
+    target.className === "cm-number" ||
+    target.className === "cm-atom";
```
jasonLaster: wow - this is really cool!

src/components/Editor/Preview/Popup.js
```diff
@@ -179,6 +179,8 @@ export class Popup extends Component {
   getPreviewType(value: any) {
     if (
       typeof value == "boolean" ||
+      (typeof value == "string" && value.length < 10) ||
+      (typeof value == "number" && value.toString().length < 10) ||
```
jasonLaster: what i think is happening here is that we will avoid showing a popup when you hover on a number, but we will show it when you hover on a variable?!?! is that correct?

tohmjudson: Yup... the getPreviewType() is what determines if the popup is a tooltip or a full popover.  The invaildType in the expression.js is what avoids showing the popup when over a number, string, or atom class

PR #4219, opened 2017-09-29
firefox doesn't like fit-content

samuel-peers:
Associated Issue: #4122

### Summary of Changes

The shortcuts modal was super long in Firefox because of `height: fit-content` => use `display: table` instead

### Screenshots
In Firefox before:
bad

In Firefox after:
whatever

### DIFF
--- a/src/components/ShortcutsModal.css
+++ b/src/components/ShortcutsModal.css
@@ -41,7 +41,6 @@
 }
 
 .shortcuts-modal {
-  display: table;
   width: 45%;
 }
 
--- a/src/components/ShortcutsModal.css
+++ b/src/components/ShortcutsModal.css
@@ -41,7 +41,7 @@
 }
 
 .shortcuts-modal {
-  height: fit-content;
+  display: table;
   width: 45%;
 }
 
--- a/src/components/shared/Modal.css
+++ b/src/components/shared/Modal.css
@@ -1,7 +1,7 @@
 .modal-wrapper {
   position: fixed;
   display: flex;
-  justify-content: center;
+  align-items: center;
   width: 100%;
   height: 100%;
   top: 0;
@@ -12,7 +12,7 @@
 
 .modal {
   top: 0;
-  height: 230px;
+  margin: auto;
   width: 500px;
   background-color: var(--theme-toolbar-background);
   transform: translateY(-250px);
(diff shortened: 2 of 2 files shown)


src/components/ShortcutsModal.css
```diff
@@ -41,7 +41,7 @@
 }
 
 .shortcuts-modal {
-  height: fit-content;
+  display: table;
```
jasonLaster: this is pretty esoteric, how did you come up w/ it?

could you link to a stack overflow or other known issue w/ firefox and fit-content

e.g.
```
// NOTE firefox does not support `height: fit-content`
// stackoverrflow.com/sldkjfslfjsljsldflsdjslkdjfls
display: table;
```

samuel-peers: Yeah moz-fit-content didn't seem to do the same thing as fit-content
https://stackoverflow.com/questions/7839164/is-there-a-css-cross-browser-value-for-width-moz-fit-content

codehag:
Hi there! The reason this is behaving strangely is because we are dealing with a flex item, not a regular div. By making it a table, you are overriding it's flexbox item status which can lead to problems. There are also some other inconsistencies with how flexbox is being used, which is what is leading to the original issue of this being full height and strangely positioned, and `fit-content` not working in firefox. The issue in the end is not `fit-content`, but flexbox inconsistencies between browsers. Here is an example of how we can solve this:

```diff
diff --git a/src/components/shared/Modal.css b/src/components/shared/Modal.css
index 35761cf1..b1f9d336 100644
--- a/src/components/shared/Modal.css
+++ b/src/components/shared/Modal.css
@@ -1,7 +1,7 @@
 .modal-wrapper {
   position: fixed;
   display: flex;
-  justify-content: center;
+  align-items: center;
   width: 100%;
   height: 100%;
   top: 0;
@@ -12,7 +12,7 @@

 .modal {
   top: 0;
-  height: 230px;
+  margin: auto;
   width: 500px;
   background-color: var(--theme-codemirror-gutter-background);
   transform: translateY(-250px);
@@ -27,7 +27,7 @@

 .modal.entered,
 .modal.exiting {
-  transform: translateY(30px);
+  transform: translateY(0px);
 }

 @media (max-width: 520px) {
```

it looks like there are some problems how the DOM is organized. We are using flex box where it isnt necessary (the list items) and the shortcut lists are all width 40%. Also, the modals are overwriting themselves, which will lead to problems later on. I am wondering if we should make an issue to clean this up?

samuel-peers:
For sure

**Chrome**
chromegood
chromegooder

**Firefox**
gooodstuff
alsogood

Issue #4223, opened 2017-09-30
Watch expressions are vulnerable to injection


Loirooriol:
After #4107, creating a `null.foo` watch expression no longer pauses the debugger when in "Pause on uncaught exceptions" mode.

However, try `` ` + null.foo + ` ``. It pauses.

Or enter `}catch(e){}; 123; try{` and the result will be `123` instead of a syntax error.

jasonLaster:
Hey @Loirooriol! Thanks for filing these watch expressions bugs.

Can you join our [slack](https://devtools-html-slack.herokuapp.com) so we can discuss

Issue #4220, opened 2017-09-29
Watch expressions are easily hijackable


Loirooriol:
1. Load http://example.com/
2. Open the web console
3. Enter `eval = () => 0`
4. Open the debugger
5. Add a new watch expression: `1`

Result: `0`
Expected: `1`

bomsy:
Oohhh ...Thanks for the report!

jasonLaster:
Nice find @Loirooriol, this is a bug with the debugger server. You can see the same issue w/ the old debugger as well.

Loirooriol:
Oh, I thought the culprit was #4107 because it wraps the expression in an `eval`.

jasonLaster:
yep, good point... we do the same [thing][t] w/ the old debugger client as well.

[t]:http://searchfox.org/mozilla-central/source/devtools/client/debugger/debugger-controller.js#1184-1193

jasonLaster:
so... this fixes it:


```diff
diff --git a/src/utils/expressions.js b/src/utils/expressions.js
index ec27dac..05e40e1 100644
--- a/src/utils/expressions.js
+++ b/src/utils/expressions.js
@@ -14,13 +14,13 @@ export function sanitizeInput(input: string) {
  * NOTE: we add line after the expression to protect against comments.
 */
 export function wrapExpression(input: string) {
-  return `eval(\`
+  return `
     try {
       ${sanitizeInput(input)}
     } catch (e) {
       e
     }
-  \`)`.trim();
+  `.trim();
 }
```

PR #4225, opened 2017-09-30
Fix some lingering POE + Exception bugs

jasonLaster:
Associated Issue: #4220, #4223, ...

### Summary of Changes

* the `eval` was unnecessary and made us susceptible to `eval` monkey patches
* our old approach of ignoring re-evaluating when pausing and resuming did help protect the server. It turns out that if you have "pause on all exceptions" you will naively re-evaluate when you're pausing in and resuming from an expression that threw. We should catch both cases...

STR:

1. set POE to all (blue)
2. add `foo.bar` *(you should only pause once)*
3. add `foo.bat` *(you should only pause once)*
4. add `3` *(you should see the result)*

### Test Plan

Yeah... we'll write some unit tests :)
And bring back the mochitest

### DIFF
--- a/src/actions/pause.js
+++ b/src/actions/pause.js
@@ -5,10 +5,12 @@ import { PROMISE } from "../utils/redux/middleware/promise";
 
 import {
   getPause,
+  pausedInEval,
   getLoadedObject,
   isStepping,
   isPaused,
-  getSelectedSource
+  getSelectedSource,
+  isEvaluatingExpression
 } from "../selectors";
 import {
   updateFrameLocations,
@@ -57,12 +59,14 @@ export function resumed() {
       return;
     }
 
+    const wasPausedInEval = pausedInEval(getState());
+
     dispatch({
       type: "RESUME",
       value: undefined
     });
 
-    if (!isStepping(getState())) {
+    if (!isStepping(getState()) && !wasPausedInEval) {
       dispatch(evaluateExpressions());
     }
   };
@@ -120,7 +124,11 @@ export function paused(pauseInfo: Pause) {
     if (hiddenBreakpointLocation) {
       dispatch(removeBreakpoint(hiddenBreakpointLocation));
     }
-    dispatch(evaluateExpressions());
+
+    if (!isEvaluatingExpression(getState())) {
+      dispatch(evaluateExpressions());
+    }
+
     dispatch(
       selectSource(frame.location.sourceId, { line: frame.location.line })
     );
--- a/src/utils/expressions.js
+++ b/src/utils/expressions.js
@@ -1,5 +1,6 @@
 // @flow
 
+import { correctIndentation } from "./indentation";
 import type { Expression } from "debugger-html";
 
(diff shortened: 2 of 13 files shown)


src/actions/pause.js
```diff
@@ -120,7 +124,11 @@ export function paused(pauseInfo: Pause) {
     if (hiddenBreakpointLocation) {
       dispatch(removeBreakpoint(hiddenBreakpointLocation));
     }
-    dispatch(evaluateExpressions());
+
+    if (!isEvaluatingExpression(getState())) {
```
jasonLaster: I prefer checking the command for whether we're evaluating an expression...

src/actions/pause.js
```diff
@@ -57,12 +59,14 @@ export function resumed() {
...
 
+    const wasPausedInEval = pausedInEval(getState());
+
     dispatch({
       type: "RESUME",
       value: undefined
     });
 
-    if (!isStepping(getState())) {
+    if (!isStepping(getState()) && !wasPausedInEval) {
```
jasonLaster: we check the `frame.displayName`, because the debugger is now resuming, it's no longer evaluating an expression

src/utils/expressions.js
```diff
@@ -14,13 +15,14 @@ export function sanitizeInput(input: string) {
...
+  return correctIndentation(`
+    eval(\`
     try {
       ${sanitizeInput(input)}
     } catch (e) {
       e
-    }
-  \`)`.trim();
+    }\`)
+  `);
```
jasonLaster: switching to `correctIndentation` because it's prettier...

src/utils/function.js
```diff
@@ -1,20 +1,5 @@
...
-  const _getIndentation = line => line && line.match(/^\s*/)[0].length;
-
-  const indentations = [
-    _getIndentation(firstLine),
-    _getIndentation(secondLine),
-    _getIndentation(lastLine)
-  ];
-
-  return Math.max(...indentations, 0);
-}
```
jasonLaster: cleaning up indentation because it makes me 😄

src/actions/pause.js
```diff
@@ -57,12 +59,14 @@ export function resumed() {
       return;
     }
 
+    const wasPausedInEval = pausedInEval(getState());
```
codehag: it feels like this should be `isPausedInEval` ?

src/utils/tests/__snapshots__/expressions.spec.js.snap
```diff
@@ -1,21 +1,17 @@
...
       e
-    }
-  \`)"
+    }"
 `;
 
 exports[`expressions wrap exxpression should wrap expression with a comment 1`] = `
-"eval(\`
-    try {
+"try {
```
codehag: is this ok?

jasonLaster: yep, the `eval` is unnenessary here

PR #4227, opened 2017-10-01
Switch Popover PropTypes to flow props

ohana54:
Associated Issue: #2804

### Summary of Changes

Switch Popover PropTypes to flow props

### Test Plan
- `yarn test`


jasonLaster:
Thanks!

Issue #4136, opened 2017-09-23
[Preview] Show React Components


jasonLaster:
We currently show react components with all of the properties. This makes it pretty hard to see what's truly important. We should design a *special* preview popup for react components:

1. Recognize it's a react component
2. Highlight state / props
3. Show the component class name (displayName)
4. Show the other properties pelow

### current

<img src='https://shipusercontent.com/929728a9e94a27ed32dfe2acd269cd6a/Screen%20Shot%202017-09-22%20at%2010.40.47%20PM.png' title='Screen Shot 2017-09-22 at 10.40.47 PM.png' width=652>

### proposal

<img src='https://shipusercontent.com/b7eb666f00efbe8351d8ef5a6a531119/Screen%20Shot%202017-09-22%20at%2010.40.21%20PM.png' title='Screen Shot 2017-09-22 at 10.40.21 PM.png' width=661>

jasonLaster:
Here's a good sketch for a V1:

when it sees a react component, it will only render props + state

```diff
➜  debugger.html gst
diff --git a/src/components/Editor/Preview/Popup.js b/src/components/Editor/Preview/Popup.js
index 5961b29..92a218d 100644
--- a/src/components/Editor/Preview/Popup.js
+++ b/src/components/Editor/Preview/Popup.js
@@ -34,6 +34,10 @@ type Props = {
   openLink: string => void
 };

+function isReactComponent(roots) {
+  return roots.any(root => root.name === "_reactInternalInstance");
+}
+
 export class Popup extends Component {
   marker: any;
   pos: any;
@@ -118,12 +122,16 @@ export class Popup extends Component {
     const { loadObjectProperties, loadedObjects, openLink } = this.props;

     const getObjectProperties = id => loadedObjects[id];
-    const roots = this.getChildren(root, getObjectProperties);
+    let roots = this.getChildren(root, getObjectProperties);

     if (!roots) {
       return null;
     }

+    if (isReactComponent(roots)) {
+      roots = roots.filter(root => ["state", "props"].includes(root.name));
+    }
+
     return (
       <ObjectInspector
         roots={roots}
```

PR #4228, opened 2017-10-02
React preview

samuel-peers:
Associated Issue: #4136

### Summary of Changes

Recognize it's a react component
Highlight state and props

### Screenshots
reactpreview


src/components/Editor/Preview/ReactPopup.js
```diff
@@ -0,0 +1,69 @@
...
+class ReactPopup extends Component {
+  props: {
+    roots: Array<Object>
+  };
+
+  renderComponentAttributes(root: Object) {
+    if (!("preview" in root.contents.value)) {
+      return <span className="none">none</span>;
+    }
+    const attributes = root.contents.value.preview.ownProperties;
```
samuel-peers: I don't know if this is the best way to grab properties... `ObjectInspector` uses `const getObjectProperties = id => loadedObjects[id];` but I couldn't get that to work :/ Any insight is appreciated ;)

jasonLaster:
I fear I lead you astray with the mockups :) but besides the purple headers, it's great.

I'll take a look at the ReactPopup implementation tomorrow. This is very cool!

src/components/Editor/Preview/Popup.js
```diff
@@ -34,6 +36,10 @@ type Props = {
   openLink: string => void
 };
 
+function isReactComponent(roots) {
+  return roots.some(root => root.name === "_reactInternalInstance");
```
codehag: can we use https://reactjs.org/docs/react-api.html#isvalidelement ?

src/components/Editor/Preview/ReactPopup.css
```diff
@@ -0,0 +1,47 @@
...
+  width: calc(100% - 1px); /* 1px fixes the hidden right border */
+}
+
+.react-popup li {
+  font-size: 12px;
+  color: var(--theme-body-color);
+  padding-top: 5px;
+  display: flex;
+  justify-content: space-between;
+  border: 1px solid transparent;
```
codehag: why are we adding a border?

src/components/Editor/Preview/ReactPopup.css
```diff
@@ -0,0 +1,47 @@
...
+  width: 70%;
```
codehag: where does this width come from?

src/components/Editor/Preview/ReactPopup.css
```diff
@@ -0,0 +1,47 @@
+.react-popup h2 {
+  margin: 0 0 5px 0;
+  font-size: 15px;
```
codehag: where does this number come from? maybe we should have a variable instead?

src/components/Editor/Preview/ReactPopup.js
```diff
@@ -0,0 +1,73 @@
...
+    if (!("preview" in root.contents.value)) {
+      return <span className="attrKey">none</span>;
+    }
+    const attributes = root.contents.value.preview.ownProperties;
+
+    return Object.keys(attributes).map(key => {
+      const val = attributes[key].value;
+      let content = "";
+
+      switch (typeof val) {
```
codehag: this might be moved into its own function, and instead have this:

```javascript
function getContent(val) {
  switch (typeof val) {
  case "object":
    if (val.class === "Function") {
      return`${val.displayName}()`;
      // .. and so on
}

const content = getContent(val);
```

src/components/Editor/Preview/ReactPopup.js
```diff
@@ -0,0 +1,73 @@
...
+            content = `${val.displayName}()`;
+          } else {
+            content = val.type;
+          }
+          break;
+        case "string":
+          content = `"${val}"`;
+          break;
+        default:
+          content = attributes[key].value.toString();
```
codehag: `attributes[key].value === val` so lets use `val.toString()` instead :)

src/components/Editor/Preview/ReactPopup.js
```diff
@@ -0,0 +1,73 @@
...
+      }
+
+      return (
+        <li key={key}>
```
codehag: do we have any errors thrown from this in the console? (asking because I can't test it at the moment 😅 )

src/components/Editor/Preview/ReactPopup.js
```diff
@@ -0,0 +1,73 @@
...
+    const roots = this.props.roots.filter(r =>
+      ["state", "props"].includes(r.name)
+    );
+
+    return (
+      <div className="react-popup">
+        <div className="react-popup-section">
+          <h2>State</h2>
+          <ul className="popup-list">
+            {this.renderComponentAttributes(roots[1])}
```
codehag: do we always have two roots? could we maybe destructure them like so:
```javascript
const [rootState, rootProps] = this.roots.filter( //...
```

src/components/Editor/Preview/ReactPopup.css
```diff
@@ -0,0 +1,47 @@
...
+.react-popup ul {
+  list-style: none;
+  margin: 0px;
+  padding: 0px;
+  overflow: auto;
+  width: calc(100% - 1px); /* 1px fixes the hidden right border */
+}
+
+.react-popup li {
+  font-size: 12px;
```
codehag: we are adding a magic number here, maybe we can use existing styles?

src/components/Editor/Preview/Popup.js
```diff
@@ -96,9 +102,22 @@ export class Popup extends Component {
   }
 
   renderObjectPreview(expression: string, root: Object) {
-    return (
-      <div className="preview-popup">{this.renderObjectInspector(root)}</div>
-    );
+    let content = null;
+    const { loadedObjects } = this.props;
+    const getObjectProperties = id => loadedObjects[id];
+    const roots = this.getChildren(root, getObjectProperties);
```
jasonLaster: could we filter the roots to just be two properties:

```js
roots = roots.filter(root => ["state", "props"].includes(root.name));

```

src/components/Editor/Preview/Popup.js
```diff
@@ -96,9 +102,22 @@ export class Popup extends Component {
...
+
+    if (!roots) {
+      return null;
+    }
+
+    if (isReactComponent(roots)) {
+      content = <ReactPopup roots={roots} />;
```
jasonLaster: At the moment, i want to avoid styling so we have some more flexibility. I think this means we can avoid the `ReactPopup` and just re-use the OI :)

samuelpeers-mhi: Lol kay. I actually did that initially, I just found it to be too ugly compared to your pretty mockup

src/components/Editor/Preview/Popup.js
samuel-peers: ya that's already being done in `ReactPopover.js`

samuel-peers:
Without styling (using just oi):
barebones

src/components/Editor/Preview/Popup.js
```diff
@@ -96,8 +100,24 @@ export class Popup extends Component {
...
+    const getObjectProperties = id => loadedObjects[id];
+    let roots = this.getChildren(root, getObjectProperties);
+
+    if (!roots) {
+      return null;
+    }
+
+    if (isReactComponent(roots)) {
+      roots = roots.filter(r => ["state", "props"].includes(r.name));
+      expandDepth = 1;
```
jasonLaster: nice... for now though, we can leave that out too. i'm okay w/ users expanding :) 

sometimes you'll want state, sometimes props. Either way, we'll iterate quickly

jasonLaster:
One question, did @codehag's comment on identifying a react component work?

samuel-peers:
ah totally forgot to check. one sec

src/components/Editor/Preview/Popup.js
samuel-peers: Hmm it doesn't seem to work for our purposes. Something like:
`  const reactElem = <span>hi sam</span>;
  React.isValidElement(reactElem);`
yields `true`, but in Popup.js we're not dealing w/ react objects, we get `roots` which I'm not totally sure what is lol (its definitely an array of elements that do not look like react elements). I know that `roots.some(root => React.isValidElement(root))` yields false even when hovering on a react component

src/components/Editor/Preview/Popup.js
```diff
@@ -34,6 +34,10 @@ type Props = {
   openLink: string => void
 };
 
+function isReactComponent(roots) {
+  return roots.some(root => root.name === "_reactInternalInstance");
```
jasonLaster: haha - we'll get back to this... as we do other objects

and yeah, i kinda like looking at a super secret internal prop

PR #4229, opened 2017-10-02
Remove this:any references

darkwing:
The `this: any` references are rare and, IMO, don't mean much; if anything, they're bowing to an unnecessary practice.


codehag:
This might have been there to satisfy flow.. i agree that it isn't meaningful -- is there perhaps a way to create a flow type that is meaningful in these cases?

jasonLaster:
Hi @darkwing i'm closing because we'll need a different solution for tackling the `self: any = this`

Issue #4221, opened 2017-09-29
Clear up react test warnings


wldcordeiro:
When you run the test suite we get lots of warnings because of some deprecations and stuff over in React. We should update the appropriate packages and import paths.

```
  ● Console

    console.warn node_modules/react-addons-test-utils/index.js:31
      Warning: ReactTestUtils has been moved to react-dom/test-utils. Update references to remove this warning.
    console.error node_modules/fbjs/lib/warning.js:33
      Warning: Shallow renderer has been moved to react-test-renderer/shallow. Update references to remove this warning.
```

pradeepgangwar:
I would like to work on this. Please help me how should i proceed.

jasonLaster:
@pradeepgangwar thanks. Try running the debugger locally and see if you can reproduce the errors

jasonLaster:
![Screen Shot 2017-10-02 at 12.18.07 AM.png](https://shipusercontent.com/c2ebcae71e7fe2652bea804018da212a/Screen%20Shot%202017-10-02%20at%2012.18.07%20AM.png)


I think a good way to reproduce is to run `yarn test`

pradeepgangwar:
I also got same set of errors on running test suite by `yarn test-all`. So how should i proceed further to correct this.
![debugger-err](https://user-images.githubusercontent.com/21259802/31064527-a9741118-a75a-11e7-973b-99c251c60a6d.png)

jasonLaster:
@pradeepgangwar i'm not sure. i'd look up the errors in google and try to track down the code that causes them

wldcordeiro:
Looking further into this I believe it may be an issue in Enzyme itself. I'm digging further in and reporting it in their repo. Marking not-available for now till I get a bit more info.

ahuth:
Picked up and resolved many of the warnings in #4248.

There are a few more for `PropTypes` and `createClass` that will take some more work to resolve.

wldcordeiro:
@ahuth the `PropType` ones will be silenced soon since we're going full with Flow. As for `createClass` I assume it maybe coming from `devtools-core` as searching for `createClass` in the debugger only has a handful of results in docs.

PR #4248, opened 2017-10-03
Fix some react test warnings

ahuth:
Resolves many of the console warnings when running tests.

Per Enzyme's documentation for v15.5 and up, react-test-renderer should be included instead of the old test utils. See https://github.com/airbnb/enzyme/pull/876.

There are still some warnings about using `React.PropTypes` that I haven't tracked down yet, and they may be coming from dependencies. Additionally, there are some for using `React.createClass` that will take some more code changes to fix.

Associated Issue: #4221 

### Summary of Changes

* Removed `react-addons-test-utils`
* Added `react-test-renderer`

### Test Plan

Ran the tests, all of which are green.

### DIFF
--- a/package.json
+++ b/package.json
@@ -138,7 +138,7 @@
     "npm-run-all": "^4.0.2",
     "prettier": "^1.6.1",
     "react-addons-perf": "^15.4.2",
-    "react-addons-test-utils": "^15.6.2",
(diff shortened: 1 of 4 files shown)


jasonLaster:
oh neat. Thanks.

@wldcordeiro was this what you had in mind?

PR #4256, opened 2017-10-03
Adds readme with debugger.html version information to mc.

yurydelendik:
Associated Issue: will help with bisecting

### DIFF
--- a/bin/writeReadme.js
+++ b/bin/writeReadme.js
@@ -1,0 +1,70 @@
+const fs = require("fs");
+const path = require("path");
+const spawn = require("child_process").spawn;
+
+function getGitSha() {
+  return new Promise((resolve, reject) => {
+    const proc = spawn("git", ["log", "--format=%H", "-n", "1"]);
+
+    let version = "";
+
+    proc.stdout.on("data", data => {
+      version += data;
+    });
+    proc.on("close", code => resolve(version.trim()));
+    proc.on("error", error => {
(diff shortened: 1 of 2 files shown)


bin/writeReadme.js
```diff
@@ -0,0 +1,70 @@
...
+  } catch (ex) {
+    return { name, version: "n/a" };
+  }
+}
+
+const packageOfInterest = [
+  "babel-plugin-transform-es2015-modules-commonjs",
+  "babel-preset-react",
+  "react",
+  "react-dom",
```
jasonLaster: we exclude some of these from the bundle: 

https://github.com/ahuth/debugger.html/blob/fix-warnings/webpack.config.js#L53-L61

bin/writeReadme.js
```diff
@@ -0,0 +1,70 @@
...
+    proc.on("close", code => resolve(version.trim()));
+    proc.on("error", error => {
+      if (error.code == "ENOENT") {
+        reject(new Error("Could not find git."));
+        return;
+      }
+      reject(error);
+    });
+  });
+}
```
jasonLaster: :thumbsup:

bin/writeReadme.js
yurydelendik: right, we might need to know difference anyway

yurydelendik: list is incomplete too

PR #4260, opened 2017-10-03
Remove usage of create factory

ahuth:
Per [React's top-level API documentation](https://reactjs.org/docs/react-api.html#createfactory), using `createFactory` isn't really intended for a project to use if it's using JSX. Also, more developers will be familiar with using JSX.

For these reasons, this PR removes most instances of `createFactory` and replaces them with JSX.

Whereas before there would be something like:

```js
const Component = createFactory(_Component)
const rendered = new Component({some: "prop"});
```

Now there will be:

```jsx
const rendered = <Component some="prop" />
```

The `ManagedTree` component still uses `createFactory`. Changing that over revealed some prop-type failures, and will take some more digging to resolve.

Let me know if there's some good reasons for staying with `createFactory` instead of just using JSX. However, I can't really think of any, and being consistent with the community and with how we generally create React elements in this project probably makes sense.

### Test Plan

Ran the tests, which are green.

### DIFF
--- a/src/components/Editor/tests/SearchBar.spec.js
+++ b/src/components/Editor/tests/SearchBar.spec.js
@@ -1,4 +1,4 @@
-import { createFactory } from "react";
+import React from "react";
 import { shallow } from "enzyme";
 import SearchBar from "../SearchBar";
 import "../../../workers/search";
@@ -11,8 +11,6 @@ jest.mock("../../../workers/search", () => ({
 jest.mock("../../../utils/editor", () => ({
   find: () => ({ ch: "1", line: "1" })
 }));
-
-const SearchBarComponent = createFactory(SearchBar.WrappedComponent);
 
 function generateDefaults() {
   return {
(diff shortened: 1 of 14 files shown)


src/components/Editor/tests/SearchBar.spec.js
```diff
@@ -39,7 +37,7 @@ function generateDefaults() {
 function render(overrides = {}) {
   const defaults = generateDefaults();
   const props = { ...defaults, ...overrides };
-  const component = shallow(new SearchBarComponent(props));
+  const component = shallow(<SearchBar.WrappedComponent {...props} />);
```
jasonLaster: lets set a variable above.

`const SearchBarComponent = SearchBar.WrappedComponent;`

jasonLaster:
Merging because this PR touches a lot of files, but a follow up PR for the comment above would be helpful.

src/components/Editor/tests/SearchBar.spec.js
ahuth: 👍 Sounds good

Issue #4168, opened 2017-09-26
[SourceTree] 30000 files in the tree


yurydelendik:
STR:
1. Open and debug http://firefox-dev.tools/debugger-examples/examples/f30k/index.html
2. Browse source tree

Actual result:

Browser hangs

Expected result:

Able to open and browse source tree with 30000 files.

jasonLaster:
```diff
diff --git a/src/utils/sources-tree/createTree.js b/src/utils/sources-tree/createTree.js
index de18287..fd007f6 100644
--- a/src/utils/sources-tree/createTree.js
+++ b/src/utils/sources-tree/createTree.js
@@ -7,10 +7,25 @@ import { addToTree } from "./addToTree";

 import type { SourcesMap } from "../../reducers/types";

+function addSources(sources, uncollapsedTree) {
+  for (const source of sources) {
+    addToTree(uncollapsedTree, source);
+  }
+}
+
 export function createTree(sources: SourcesMap, debuggeeUrl: string) {
   const uncollapsedTree = createNode("root", "", []);
-  for (const source of sources.valueSeq()) {
-    addToTree(uncollapsedTree, source, debuggeeUrl);
+
+  const sourceList = sources.valueSeq();
+
+  if (huge) {
+    // 30K
+    // batches of 1000
+    sourceListPartitions.forEach(
+      requestIdleCallback(() => addSources(sources, uncollapsedTree))
+    );
+  } else {
+    addSources(sources, uncollapsedTree);
   }

   const sourceTree = sortEntireTree(collapseTree(uncollapsedTree), debuggeeUrl);
(END)

```

yurydelendik:
I measured performance of source tree with 3,000 files (10x less of the target), and following pain points are:

* addTree operation ~470ms
* tree sort ~160ms
* tree batch component update ~5900ms

I was able to reduce time to 30%, by:

1. Removing SVG {icon} (improvement)-> tree batch component update ~2100ms
2. Not parsing debuggee domain all the time -> tree sort ~5ms
3. Having an name index for nodes -> addTree operation ~205ms

See blueprint at https://gist.github.com/yurydelendik/0b10038753f4facf750f727a9db49844

dholbert:
>I was able to reduce time to 30%, by ... Removing SVG {icon}

[...]

PR #4263, opened 2017-10-03
Optimize addToTree function.

yurydelendik:
Associated Issue: #4168

### Summary of Changes

* don't use sortTree  for sources
* improve the performance of addToTree

### Test Plan

Added console.time() in the SourcesTree.js and createTree.js around addToTree+sortTree operations.
Measured performance of https://yurydelendik.github.io/old-man-sandbox/create-30k/f3k.html tree.

Before:
screen shot 2017-10-03 at 2 59 13 pm

After:
screen shot 2017-10-03 at 3 03 50 pm

Performance improved from 1186ms->291ms also affected subsequent/incremental tree update operations 180ms->2ms.

### DIFF
--- a/src/utils/sources-tree/addToTree.js
+++ b/src/utils/sources-tree/addToTree.js
@@ -7,15 +7,29 @@ import {
   partIsFile,
   createNode
 } from "./utils";
+import {
+  createTreeNodeMatcher,
+  findNodeInContents,
+  getDomain
+} from "./treeOrder";
 import { getURL, getFilenameFromPath } from "./getURL";
 
 import type { Node } from "./types";
 import type { SourceRecord } from "../../reducers/types";
 
-function createNodeInTree(part: string, path: string, tree: Node) {
+function createNodeInTree(
+  part: string,
+  path: string,
+  tree: Node,
+  index: number
+) {
   const node = createNode(part, path, []);
+
   // we are modifying the tree
-  tree.contents = [...tree.contents, node];
+  const contents = tree.contents.slice(0);
+  contents.splice(index, 0, node);
+  tree.contents = contents;
+
   return node;
 }
 
@@ -31,23 +45,28 @@ function findOrCreateNode(
   path: string,
   part: string,
   index: number,
-  url: Object
+  url: Object,
+  debuggeeHost: ?string
 ) {
-  const child = subTree.contents.find(c => c.name === part);
+  const addedPartIsFile = partIsFile(index, parts, url);
+  const { found: childFound, index: childIndex } = findNodeInContents(
+    subTree,
+    createTreeNodeMatcher(part, !addedPartIsFile, debuggeeHost)
+  );
 
   // we create and enter the new node
-  if (!child) {
-    return createNodeInTree(part, path, subTree);
+  if (!childFound) {
+    return createNodeInTree(part, path, subTree, childIndex);
   }
 
   // we found a path with the same name as the part. We need to determine
   // if this is the correct child, or if we have a naming conflict
-  const addedPartIsFile = partIsFile(index, parts, url);
+  const child = subTree.contents[childIndex];
   const childIsFile = !nodeHasChildren(child);
 
   // if we have a naming conflict, we'll create a new node
   if ((childIsFile && !addedPartIsFile) || (!childIsFile && addedPartIsFile)) {
-    return createNodeInTree(part, path, subTree);
+    return createNodeInTree(part, path, subTree, childIndex);
   }
 
   // if there is no naming conflict, we can traverse into the child
@@ -58,7 +77,7 @@ function findOrCreateNode(
  * walk the source tree to the final node for a given url,
  * adding new nodes along the way
  */
-function traverseTree(url: Object, tree: Node) {
+function traverseTree(url: Object, tree: Node, debuggeeHost: ?string) {
   url.path = decodeURIComponent(url.path);
 
   const parts = url.path.split("/").filter(p => p !== "");
@@ -67,7 +86,16 @@ function traverseTree(url: Object, tree: Node) {
   let path = "";
   return parts.reduce((subTree, part, index) => {
     path = `${path}/${part}`;
-    return findOrCreateNode(parts, subTree, path, part, index, url);
+    const debuggeeHostIfRoot = index === 0 ? debuggeeHost : null;
+    return findOrCreateNode(
+      parts,
+      subTree,
+      path,
+      part,
+      index,
+      url,
+      debuggeeHostIfRoot
+    );
   }, tree);
 }
 
@@ -84,18 +112,24 @@ function addSourceToNode(node: Node, url: Object, source: SourceRecord) {
   }
 
   const name = getFilenameFromPath(url.path);
-  const existingNode = node.contents.find(childNode => childNode.name === name);
+  const { found: childFound, index: childIndex } = findNodeInContents(
+    node,
+    createTreeNodeMatcher(name, false, null)
+  );
 
   // if we are readding an existing file in the node, overwrite the existing
   // file and return the node's contents
-  if (existingNode) {
+  if (childFound) {
+    const existingNode = node.contents[childIndex];
     existingNode.contents = source;
     return node.contents;
   }
 
   // if this is a new file, add the new file;
   const newNode = createNode(name, source.get("url"), source);
-  return [...node.contents, newNode];
+  const contents = node.contents.slice(0);
+  contents.splice(childIndex, 0, newNode);
+  return contents;
 }
 
 /**
@@ -108,11 +142,12 @@ export function addToTree(
   debuggeeUrl: string
 ) {
   const url = getURL(source.get("url"), debuggeeUrl);
+  const debuggeeHost = getDomain(debuggeeUrl);
 
   if (isInvalidUrl(url, source)) {
     return;
   }
 
-  const finalNode = traverseTree(url, tree);
+  const finalNode = traverseTree(url, tree, debuggeeHost);
   finalNode.contents = addSourceToNode(finalNode, url, source);
 }
--- a/src/utils/sources-tree/treeOrder.js
+++ b/src/utils/sources-tree/treeOrder.js
@@ -1,0 +1,131 @@
+// @flow
+
+import { parse } from "url";
+
+import { nodeHasChildren } from "./utils";
+
+import type { Node } from "./types";
+
+/*
+ * Gets domain from url (without www prefix)
+ */
+export function getDomain(url: string): ?string {
+  const { host } = parse(url);
+  if (!host) {
+    return null;
+  }
+  return host.startsWith("www.") ? host.substr("www.".length) : host;
+}
+
+/*
+ * Checks if node name matches debugger host/domain.
+ */
+function isExactDomainMatch(part: string, debuggeeHost: string): boolean {
+  return part.startsWith("www.")
+    ? part.substr("www.".length) === debuggeeHost
+    : part === debuggeeHost;
+}
+
+/*
+ * Function to assist with node search for a defined sorted order, see e.g.
+ * `createTreeNodeMatcher`. Returns negative number if the node
+ * stands earlier in sorting order, positive number if the node stands later
+ * in sorting order, or zero if the node is found.
+ */
+export type FindNodeInContentsMatcher = (node: Node) => number;
+
+/*
+ * Performs a binary search to insert a node into contents. Returns positive
+ * number, index of the found child, or negative number, which can be used
+ * to calculate a position where a new node can be inserted (`-index - 1`).
+ * The matcher is a function that returns result of comparision of a node with
+ * lookup value.
+ */
+export function findNodeInContents(
+  tree: Node,
+  matcher: FindNodeInContentsMatcher
+) {
+  const { contents } = tree;
+  if (contents.length === 0) {
+    return { found: false, index: 0 };
+  }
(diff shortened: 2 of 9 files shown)


src/utils/sources-tree/addToTree.js
```diff
@@ -1,5 +1,7 @@
 // @flow
 
+import { parse } from "url";
```
jasonLaster: does it make sense to have this in `getURL` to have one module for URL parsing?

src/utils/sources-tree/addToTree.js
```diff
@@ -12,14 +14,118 @@ import { getURL, getFilenameFromPath } from "./getURL";
...
+ */
+function isExactDomainMatch(part: string, debuggeeHost: string): boolean {
+  return part.indexOf("www.") === 0
+    ? part.substr(4) === debuggeeHost
+    : part === debuggeeHost;
+}
+
+type FindNodeInContentsMatcher = any => number;
+/*
+ * Performs a binary search to insert a node into contents. Returns positive
```
jasonLaster: :+1:

src/utils/sources-tree/addToTree.js
```diff
@@ -12,14 +14,118 @@ import { getURL, getFilenameFromPath } from "./getURL";
...
+    return null;
+  }
+  return host.indexOf("www.") === 0 ? host.substr(4) : host;
+}
+
+/*
+ * Checks if node name matches debugger host/domain.
+ */
+function isExactDomainMatch(part: string, debuggeeHost: string): boolean {
+  return part.indexOf("www.") === 0
```
adrian17: just curious: why not `part.startsWith("www.")` ?

src/utils/sources-tree/addToTree.js
```diff
@@ -12,14 +14,118 @@ import { getURL, getFilenameFromPath } from "./getURL";
...
+
+type FindNodeInContentsMatcher = any => number;
+/*
+ * Performs a binary search to insert a node into contents. Returns positive
+ * number, index of the found child, or negative number, which can be used
+ * to calculate a position where a new node can be inserted (`-index - 1`).
+ * The matcher is a function that returns result of comparision of a node with
+ * lookup value.
+ */
+function findNodeInContents(tree: Node, matcher: FindNodeInContentsMatcher) {
```
codehag: describing this function in plain english, and using the ternary values I defined elsewhere (`T_TRUE = 1, T_UNKNOWN = 0, T_FALSE = -1`):

* we take the tree contents which is an array. if that array is empty, we return -1.

* if it is not empty, we set a start point ('l' for left) and an end point ('r' for right)

* so long as the starting point is less than the end point, we also select a center point ('c')

* if the center point is `T_FALSE` ( -1 ) -- then it should be at the bottom of the list, so we set the start point (l) to c + 1

* otherwise, the center point is `T_UNKNOWN`, or `T_TRUE`, so we set the ending point to C

example:

```
0|1|2|3|4
----------
l| |c| |r

if (matcher(contents[c]) === `T_FALSE`) 
0|1|2|3|4
----------
 | | |l, c|r // ignore the first half of the array

else
0|1|2|3|4
----------
l|c|r| |  // ignore the second half of the array
```

* Once the loop completes, (and our starting point "l" is greater than our endpoint "r") 

```
0|1|2|3|4
----------
 |r|l| |
0|1|2|3|4
----------
 | | | |r|l // outside of the original array if all matcher checks return `T_FALSE`
```

we check if the result is `T_UNKNOWN`. If it is unknown, we return the starting value ("l").

If it is either `T_FALSE` (-1) or `T_TRUE` (1), then 

- check if it is `T_FALSE`, if it is, return the starting point ('l'). 
- if it is `T_TRUE`, return the starting point ('l') plus 1. 

whatever the result is, make it negative and -1 from it.

```
0|1|2|3|4
----------
 |r|l|x|   // if `T_FALSE` is true, the returned position will be -(l + 1) ( also see above for the off by one case, where  | | | |r|l|x -- is outside of the existing array by two spots)
0|1|2|3|4
----------
 |r|l| |x| // if `T_TRUE` is true, the returned position will be -(l + 2)
```

this code is essentially taking an array, seeing if the matching position would be above or below the center, and halving it to the side where the element might be found. That isnt immediately clear from the code. there are two things contributing to this: 

1) bad naming. Single character variable names shouldn't be used except in special situations, and this isn't one of them.
2) return value is doing too much. it is unclear to me why we need it to be negative, or why we need to add +1 to it. can you clarify why this is happening? also, there are a lot of magic things happening, for example +1 is being added in a number of places without any reason explained behind it. return value is a good example of "why are we adding 1 and then making the whole thing negative and then adding 1 to it again". at the very least the return value needs to be split up


it also seems like we are avoiding complex data structures, but in this case i think we need them. we shouldnt be checking for negative indexs to tell if a node is new or not.

src/utils/sources-tree/addToTree.js
```diff
@@ -12,14 +14,118 @@ import { getURL, getFilenameFromPath } from "./getURL";
...
+
+/*
+ * Creates a matcher for findNodeInContents.
+ * The sorting order of nodes during comparison is:
+ * - "(index)" node
+ * - root node with the debuggee host/domain
+ * - hosts/directories (not files) sorted by name
+ * - files sorted by name
+ */
+function createTreeNodeMatcher(
```
codehag: this should be moved to its own file, each HOF can be split into its own function and named. that way you wont have to add so many comments

```javascript
function indexMatcher(indexName) {
  return (node) => (node.name === IndexName ? 0 : 1);
}
function domainMatcher(indexName, debuggeeHost) {
 return node => {
   if (node.name === IndexName) {
      return -1;
    }
    return isExactDomainMatch(node.name, (debuggeeHost: any)) ? 0 : 1;
  };
}

createTreeNodeMatcher(...) {
  //...
  if (part === IndexName) {
    return IndexMatcher(indexName)
  } 
  if (debuggeeHost && isExactDomainMatch(part, debuggeeHost)) {
    return domainMatcher(indexName, debuggeeHost)
  }
  // ... etc
```

src/utils/sources-tree/addToTree.js
```diff
@@ -12,14 +14,118 @@ import { getURL, getFilenameFromPath } from "./getURL";
...
+  return part.indexOf("www.") === 0
+    ? part.substr(4) === debuggeeHost
+    : part === debuggeeHost;
+}
+
+type FindNodeInContentsMatcher = any => number;
+/*
+ * Performs a binary search to insert a node into contents. Returns positive
+ * number, index of the found child, or negative number, which can be used
+ * to calculate a position where a new node can be inserted (`-index - 1`).
```
codehag: so, if an index is negative, it is a new node? and if it is positive it is an existing node? that is difficult to understand. why not use a tuple? { index: <number>, newNode: <boolean> }

src/utils/sources-tree/addToTree.js
```diff
@@ -31,23 +137,28 @@ function findOrCreateNode(
...
+  const childIndex = findNodeInContents(
+    subTree,
+    createTreeNodeMatcher(part, !addedPartIsFile, debuggeeHost)
+  );
 
   // we create and enter the new node
-  if (!child) {
-    return createNodeInTree(part, path, subTree);
+  if (childIndex < 0) {
+    return createNodeInTree(part, path, subTree, -childIndex - 1);
```
codehag: see other comments

src/utils/sources-tree/addToTree.js
```diff
@@ -84,18 +204,24 @@ function addSourceToNode(node: Node, url: Object, source: SourceRecord) {
...
+    const existingNode = node.contents[childIndex];
     existingNode.contents = source;
     return node.contents;
   }
 
   // if this is a new file, add the new file;
   const newNode = createNode(name, source.get("url"), source);
-  return [...node.contents, newNode];
+  const contents = node.contents.slice(0);
+  contents.splice(-childIndex - 1, 0, newNode);
```
codehag: it is so unclear what is going on here. we shouldn't be using negative numbers to determine if something is a new node

src/utils/sources-tree/addToTree.js
```diff
@@ -12,14 +14,118 @@ import { getURL, getFilenameFromPath } from "./getURL";
...
+  }
+  const result = matcher(contents[l]);
+  if (result === 0) {
+    return l;
+  }
+  return -(result > 0 ? l : l + 1) - 1;
+}
+
+/*
+ * Creates a matcher for findNodeInContents.
```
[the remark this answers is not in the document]
yurydelendik: Actually, it's similar to what any comparison function will return: negative for less, 0 for equial and positive for greater

src/utils/sources-tree/addToTree.js
yurydelendik: Sorry, I don't follow your analysis. The algorithm above is basic binary search algorithm. I renamed 'l' and 'r' to 'left' and 'right' for clarity.

src/utils/sources-tree/addToTree.js
```diff
@@ -12,14 +14,118 @@ import { getURL, getFilenameFromPath } from "./getURL";
...
+    if (debuggeeHost && isExactDomainMatch(node.name, debuggeeHost)) {
+      return -1;
+    }
+    const nodeIsDir = nodeHasChildren(node);
+    if (nodeIsDir && !isDir) {
+      return -1;
+    } else if (!nodeIsDir && isDir) {
+      return 1;
+    }
+    return node.name.localeCompare(part);
```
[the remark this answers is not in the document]
yurydelendik: Added comments to the FindNodeInContentsMatcher instead.

src/utils/sources-tree/addToTree.js
yurydelendik: Changed return type to return an object (I hope JIT will not create lots of objects in memory to GC them later). Measured performance for 3k file -- there was no real change in the performance.

src/utils/sources-tree/addToTree.js
yurydelendik: No reason, probably just a habit to support old browsers. Changed to startsWith for readability, since all browsers we support have it.

src/utils/sources-tree/addToTree.js
yurydelendik: > describing this function in plain english, and using the ternary values I defined elsewhere

The algorithm is using return values similar to comparefn defined in http://www.ecma-international.org/ecma-262/5.1/#sec-15.4.4.11 , and these are not compatible with  ternary values.

src/utils/sources-tree/addToTree.js
yurydelendik: Not at the moment -- I did not want to affect any other logic that might be in "assumed" in the code.

src/utils/sources-tree/addToTree.js
yurydelendik: Code was moved to the ./treeOrder.js file.

src/utils/sources-tree/addToTree.js
codehag: :+1: i mean ternary as in: https://en.wikipedia.org/wiki/Three-valued_logic -- which i see as being similar to what is going on in the sort / comparefn that you linked (the difference being that 0 is used for "equal" instead of "undefined"). But i could be wrong... anyway, thank you for the changes it reads much better!

src/utils/sources-tree/treeOrder.js
```diff
@@ -0,0 +1,131 @@
...
+      left = middle + 1;
+    } else {
+      right = middle;
+    }
+  }
+  const result = matcher(contents[left]);
+  if (result === 0) {
+    return { found: true, index: left };
+  }
+  return { found: false, index: result > 0 ? left : left + 1 };
```
codehag: thanks for clarifying why you were doing what you were doing before! The garbage collection might be an issue, true. I hadn't thought of that, and you are totally right

Perhaps a way to make the code readable and more memory efficient (trying to balance things :sweat_smile:), we could keep the code you had (with a comment about what it was doing, a more explicit comment)

then where the value is used, have a wrapper function that translates the value? 

ie)

```javascript
export function findNodeInContents(
  // ...
  // returns a negative value for a new node, offset by one
  // returns a positive value for an existing node
  return -(result > 0 ? left : left + 1 ) - 1;
```

with helpers
```javascript
function childFound(child) {
  return child > -1
} 

function determineIndex(child) {
  // return a valid index whether the child is new (an therefore a negative number off by one)
  // or if the child is an existing index
  return child > -1 ? child : -child - 1;
}
```


then where it is used:
```javascript
if (!childFound(childIndex)) {
  return createNodeInTree(part, path, subTree, determineIndex(childIndex));
}
```

src/utils/sources-tree/addToTree.js
yurydelendik: there is no guarantee that localeCompare() will return only -1, 0, or 1 (based on spec). Writing `const result = name1.localeCompare(name2); return result === 0 ? T_UNKNOWN : result < 0 ? T_FALSE : T_TRUE;` to accommodate ternary values does not make any sense (and unreadable) IMHO.

codehag:
Thanks for the clarifications yury! i made one more comment for your feedback

src/utils/sources-tree/treeOrder.js
yurydelendik: I would keep that as is for now for readability. We can re-open the issue later. It's hard to measure atm -- it's impossible to run it with 10,000 or 30,000 load.

codehag:
awesome work!

Issue #4261, opened 2017-10-03
[Breakpoints] sort breakpoints by line number


nyrosmith:
Currently the Breakpoint component displays the breakpoints in the same order as they get added.
This is not optimal and breakpoints should get ordered by line number

### screenshot

![unbenannt](https://user-images.githubusercontent.com/2511026/31144394-ebdef046-a880-11e7-828c-a1c40ce96fde.png)

PR #4264, opened 2017-10-03
Fix 4261 - Sort breakpoints by line number

darkwing:
Started this before realizing that @nyrosmith had started it :/ .  

Anyways, this was my solution.  Hoping @nyrosmith has something prettier coming :)

### DIFF
--- a/src/components/SecondaryPanes/Breakpoints.js
+++ b/src/components/SecondaryPanes/Breakpoints.js
@@ -20,7 +20,7 @@ import { getFilename } from "../../utils/source";
 import { showMenu, buildMenu } from "devtools-launchpad";
 import CloseButton from "../shared/Button/Close";
 import "./Breakpoints.css";
-import { get } from "lodash";
+import { get, sortBy } from "lodash";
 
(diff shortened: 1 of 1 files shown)


nyrosmith:
@darkwing I wanted to prepare this with with some patch so new contributors could have jumped into https://github.com/devtools-html/debugger.html/issues/4261

jasonLaster:
@nyrosmith i guess we missed our chance :)

src/components/SecondaryPanes/Breakpoints.js
```diff
@@ -393,7 +420,10 @@ class Breakpoints extends PureComponent {
       breakpoints.size === 0 ? (
         <div className="pane-info">{L10N.getStr("breakpoints.none")}</div>
       ) : (
-        breakpoints.valueSeq().map(bp => this.renderBreakpoint(bp))
+        breakpoints
```
jasonLaster: [sortBy][https://lodash.com/docs/4.17.4#sortBy] could be amazing here:


```js
sortBy(bps, ['filename', 'line'])
```

Issue #3172, opened 2017-06-14
[flow] Improve covariant style


jasonLaster:
We currently follow an awkward style for covariants 

```js

  constructor() {
    super();

    const self: any = this;
    self.closeConditionalPanel = this.closeConditionalPanel.bind(this);
    self.onEscape = this.onEscape.bind(this);
    self.onGutterClick = this.onGutterClick.bind(this);
    self.onGutterContextMenu = this.onGutterContextMenu.bind(this);
    self.onScroll = this.onScroll.bind(this);
    self.onSearchAgain = this.onSearchAgain.bind(t
```

@thejameskyle shared some wisdom on how we can improve the style

---

Class methods are covariant in Flow by default, which in effect means they are "read-only" and cannot be assigned to.

```js
class Error extends React.Component {
  constructor(props) {
    super(props);
    this.onClick = this.onClick.bind(this); // Error
    //   ^ property `onClick`. Covariant property `onClick` incompatible with contravariant use in
    //     assignment of property `onClick`
  }
  
  onClick(event: ButtonClickEvent) {
    // ...
  }
  
  render() {
    return <button onClick={this.onClick}>Click Me</button>;
  }
}
```

In order to make this work, you need to use an additional class field, which are invariant (read & write) by default, like this:

```js
class WorksWithClassPropertyType extends React.Component {
  constructor(props) {
    super(props);
    this.onClick = this.onClick.bind(this); // Works!
  }
  
  onClick: (ButtonClickEvent) => void;
  onClick(event) {
    // ...
  }
}
```

Alternatively you can use the class properties proposal to get around all of this:

```js
class WorksWithClassPropertyArrowFunction extends React.Component {
  onClick = (event: ButtonClickEvent) => {
    // ...
  };
}
```

[...]

PR #4266, opened 2017-10-03
Proposal for closing #3172

maniator:
Associated Issue: #3172

Here's the Pull Request Doc

### Summary of Changes
Using https://babeljs.io/docs/plugins/transform-class-properties/ to use class properies instead of giant ```self.<fn> = <fn>.bind(this); ... ``` blocks

Also in the es class field proposal https://github.com/tc39/proposal-class-fields

### DIFF
--- a/src/components/Editor/index.js
+++ b/src/components/Editor/index.js
@@ -92,15 +92,6 @@ class Editor extends PureComponent {
       highlightedLineRange: null,
       editor: null
     };
-
-    const self: any = this;
-    self.closeConditionalPanel = this.closeConditionalPanel.bind(this);
-    self.onEscape = this.onEscape.bind(this);
-    self.onGutterClick = this.onGutterClick.bind(this);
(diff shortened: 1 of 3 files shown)


src/components/Editor/index.js
```diff
@@ -92,15 +92,6 @@ class Editor extends PureComponent {
       highlightedLineRange: null,
       editor: null
     };
```
maniator: We can also use the class properties for state, and all of the other vars set in this constructor

maniator:
This will also get rid of a lot of `: any`s in the code

maniator:
Also if this is approved we can apply this to other classes as well to make the code a bit cleaner :-)

src/components/Editor/index.js
maniator: (and then this overridden constructor can be removed)

thejameskyle:
Sweet :) This looks great

codehag:
works for me :+1:

PR #4268, opened 2017-10-03
Remove object.assign from selectors.js close #4267

### LINKED ISSUE #4267: selectors.js should not use Object.assign as it goes against es6 paradigms
Removing Object.assign will allow for ease of use with IDEs and it is not completely needed.

maniator:
Associated Issue: #4267

Here's the Pull Request Doc

### Summary of Changes

Written in the related issue


maniator:
Maybe not needed.

Can be closed.

jasonLaster:
I like it

Issue #4213, opened 2017-09-29
[Shortcuts] modal polish


jasonLaster:
#### Todos

- [ ] we should add a shortcut for launching it via `cmd + /` i think
- [ ] we should move K+ to the right and soften the color (whatever the accordion headers are)
- [ ] the section headers in the mocks looked pretty nice. Lets do some small typography cleanup.

```diff
diff --git a/src/components/App.js b/src/components/App.js
index 1dd5132..62f5b35 100644
--- a/src/components/App.js
+++ b/src/components/App.js
@@ -100,6 +100,9 @@ class App extends Component {
       L10N.getStr("symbolSearch.search.key2"),
       this.toggleSymbolModal
     );
+
+    shortcuts.on(shortcutModalKey, this.toggleShortcutsModal);
+
     shortcuts.on("Escape", this.onEscape);
   }

diff --git a/src/components/SecondaryPanes/index.js b/src/components/SecondaryPanes/index.js
index 55cd6f9..174f661 100644
--- a/src/components/SecondaryPanes/index.js
+++ b/src/components/SecondaryPanes/index.js
@@ -222,6 +222,7 @@ class SecondaryPanes extends Component {
       return;
     }

+    // we can style this in utils bar.css
     return (
       <UtilsBar
         horizontal={this.props.horizontal}
diff --git a/src/components/ShortcutsModal.css b/src/components/ShortcutsModal.css
index b3edaaa..c29b771 100644
--- a/src/components/ShortcutsModal.css
+++ b/src/components/ShortcutsModal.css
@@ -5,6 +5,7 @@
   column-width: 250px;
 }

+/* HERE */
 .shortcuts-content h2 {
```

codehag:
K+ is not clear in my opinion. perhaps we should write it out, or use `?` for help or something like this?

jasonLaster:
Yeah, now that you say it I think `?` is a more universal button for shortcuts

jasonLaster:
Here is the design for shortcuts. I think the only thing style wise is the headers

![](https://user-images.githubusercontent.com/55994/30822975-9ab5ff7c-a1df-11e7-80ee-7c486667a416.png)

PR #4273, opened 2017-10-04
Changed K+ to K?

Subtlesnow:
Associated Issue: #4213

Here's the Pull Request Doc

### Summary of Changes

- [x] Took out SVG component that was for K+ and added K? in text because we couldn't find another SVG icon.
- [x] Moved the button to the right.
- [x] Soften Color to match accordion

### DIFF
--- a/src/components/App.js
+++ b/src/components/App.js
@@ -72,6 +72,7 @@ class App extends Component {
   renderVerticalLayout: Function;
   toggleSymbolModal: Function;
   onEscape: Function;
+  onCommandSlash: Function;
 
   constructor(props) {
     super(props);
@@ -88,6 +89,7 @@ class App extends Component {
     this.renderEditorPane = this.renderEditorPane.bind(this);
     this.renderVerticalLayout = this.renderVerticalLayout.bind(this);
     this.onEscape = this.onEscape.bind(this);
+    this.onCommandSlash = this.onCommandSlash.bind(this);
   }
 
   getChildContext() {
@@ -101,6 +103,7 @@ class App extends Component {
       this.toggleSymbolModal
     );
     shortcuts.on("Escape", this.onEscape);
+    shortcuts.on("Cmd+/", this.onCommandSlash);
   }
 
   componentWillUnmount() {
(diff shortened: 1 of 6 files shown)


src/components/App.js
```diff
@@ -72,6 +72,7 @@ class App extends Component {
   renderVerticalLayout: Function;
   toggleSymbolModal: Function;
   onEscape: Function;
+  onCommandSlash: Function;
```
codehag: Someone mentioned that there is another command for opening the shortcuts modal... perhaps we should double check before doing this?  @jasonLaster thoughts? maybe this would be better in a separate pr?

src/components/SecondaryPanes/CommandBar.css
```diff
@@ -2,6 +2,7 @@
   flex: 0 0 29px;
   border-bottom: 1px solid var(--theme-splitter-color);
   display: flex;
+  justify-content: flex-end;
```
codehag: this also moves all of the stepping buttons over to the right hand side of the bar, which is not something we want to do right now. lets remove this rule

src/components/SecondaryPanes/UtilsBar.js
```diff
@@ -15,7 +14,7 @@ function debugBtn(onClick, type, className, tooltip, disabled = false) {
 
   return (
     <button className={classnames(type, className)} {...props}>
-      <Svg name={type} />
+      K?
```
codehag: I was thinking that the button could be a simple `?` (since im used to it from other editors) with no K (since K is not necessarily meaningful).. not sure, what do you think?

jasonLaster:
Hey @Subtlesnow do you think you'll have time to make these changes?

Subtlesnow:
Yes I should be able to fix them in the next week.

jasonLaster:
Thanks @Subtlesnow!

jasonLaster:
Hey @Subtlesnow how's your availability?

Subtlesnow:
Hello,

I apologize for getting around to fixing those bugs. I have been very busy
this past week.

jasonLaster:
Hey @Subtlesnow, I re-opened your PR on #4382 w/ some small tweaks. Thanks for the help at the meetup!

Issue #4232, opened 2017-10-02
[Stepping] step over yield expressions


jasonLaster:
We currently support stepping over `await` expressions. We should do the same thing for `yield` expressions. [here][h]

[h]: https://astexplorer.net/#/gist/6e82de985482df2701e84e37c65a9128/6e9fb16ba6cf9bcf44fafc20ed28b5df03adf013

jasonLaster:
Here is a quick sketch of how we can extend our await stepping to also support yields.

```diff
diff --git a/src/utils/parser/steps.js b/src/utils/parser/steps.js
index 29af7f4..335531f 100644
--- a/src/utils/parser/steps.js
+++ b/src/utils/parser/steps.js
@@ -5,7 +5,7 @@ import { isAwaitExpression } from "./utils/helpers";
 import type { NodePath } from "babel-traverse";

 export function getNextStep(source: Source, pausedPosition: AstPosition) {
-  const awaitExpression = getAwaitExpression(source, pausedPosition);
+  const awaitExpression = getAsyncExpression(source, pausedPosition);
   if (!awaitExpression) {
     return null;
   }
@@ -13,18 +13,18 @@ export function getNextStep(source: Source, pausedPosition: AstPosition) {
   return getLocationAfterAwaitExpression(awaitStatement, pausedPosition);
 }

-function getAwaitExpression(source: Source, pausedPosition: AstPosition) {
+function getAsyncExpression(source: Source, pausedPosition: AstPosition) {
   const closestPath = getClosestPath(source, pausedPosition);

   if (!closestPath) {
     return null;
   }

-  if (isAwaitExpression(closestPath)) {
+  if (isAwaitExpression(closestPath) || isYieldExpression(closestPath)) {
     return closestPath;
   }

-  return closestPath.find(p => p.isAwaitExpression());
+  return closestPath.find(p => p.isAwaitExpression() || p.isYieldExpression());
 }
```

```diff
diff --git a/src/utils/parser/tests/steps.spec.js b/src/utils/parser/tests/steps.spec.js
index f00c523..1879f25 100644
--- a/src/utils/parser/tests/steps.spec.js
+++ b/src/utils/parser/tests/steps.spec.js
@@ -2,34 +2,40 @@ import { getNextStep } from "../steps";
 import { getSource } from "./helpers";

 describe("getNextStep", () => {
-  it("first await call", () => {
-    const source = getSource("async");
[...]

PR #4276, opened 2017-10-04
Added step over functionality for yield (#4232)

jayjaycross:
Associated Issue: #4232

### Summary of Changes

* Added step over functionality for `yield`

### Test Plan

Example test plan:

- [x] Wrote local HTML to try out step over functionality for `await` and `yield`
- [x] Wrote unit tests to cover `yield` step over scenarios.

Opened with @longitude @mjfrancoeur.

### DIFF
--- a/src/workers/parser/steps.js
+++ b/src/workers/parser/steps.js
@@ -1,11 +1,11 @@
 import { Source } from "../../../flow-typed/debugger-html";
 import { AstPosition } from "./types";
 import { getClosestPath } from "./utils/closest";
-import { isAwaitExpression } from "./utils/helpers";
+import { isAwaitExpression, isYieldExpression } from "./utils/helpers";
 import type { NodePath } from "babel-traverse";
 
 export function getNextStep(source: Source, pausedPosition: AstPosition) {
-  const awaitExpression = getAwaitExpression(source, pausedPosition);
+  const awaitExpression = getAsyncExpression(source, pausedPosition);
   if (!awaitExpression) {
     return null;
   }
@@ -13,18 +13,18 @@ export function getNextStep(source: Source, pausedPosition: AstPosition) {
   return getLocationAfterAwaitExpression(awaitStatement, pausedPosition);
 }
 
-function getAwaitExpression(source: Source, pausedPosition: AstPosition) {
+function getAsyncExpression(source: Source, pausedPosition: AstPosition) {
   const closestPath = getClosestPath(source, pausedPosition);
 
   if (!closestPath) {
     return null;
   }
 
-  if (isAwaitExpression(closestPath)) {
+  if (isAwaitExpression(closestPath) || isYieldExpression(closestPath)) {
     return closestPath;
   }
 
-  return closestPath.find(p => p.isAwaitExpression());
+  return closestPath.find(p => p.isAwaitExpression() || p.isYieldExpression());
 }
 
 function getLocationAfterAwaitExpression(
--- a/src/workers/parser/steps.js
+++ b/src/workers/parser/steps.js
@@ -5,15 +5,15 @@ import { isAwaitExpression, isYieldExpression } from "./utils/helpers";
 import type { NodePath } from "babel-traverse";
 
 export function getNextStep(source: Source, pausedPosition: AstPosition) {
-  const awaitExpression = getAsyncExpression(source, pausedPosition);
-  if (!awaitExpression) {
+  const currentExpression = getSteppableExpression(source, pausedPosition);
+  if (!currentExpression) {
     return null;
   }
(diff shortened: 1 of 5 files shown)


src/workers/parser/steps.js
```diff
@@ -1,30 +1,30 @@
 import { Source } from "../../../flow-typed/debugger-html";
 import { AstPosition } from "./types";
 import { getClosestPath } from "./utils/closest";
-import { isAwaitExpression } from "./utils/helpers";
+import { isAwaitExpression, isYieldExpression } from "./utils/helpers";
 import type { NodePath } from "babel-traverse";
 
 export function getNextStep(source: Source, pausedPosition: AstPosition) {
-  const awaitExpression = getAwaitExpression(source, pausedPosition);
+  const awaitExpression = getAsyncExpression(source, pausedPosition);
```
codehag: this variable should also be renamed for consistency

src/workers/parser/tests/fixtures/async.js
```diff
@@ -8,3 +8,8 @@ async function stuff() {
   await foo(1);
   await foo(2);
 }
+
+function* bar() {
```
codehag: this should be moved into its own file (called "generators")

reason being:
`yield` isnt asynchronous. its a generator. generators are processes that can be paused and resumed. it is the building block of the async/await keyword, but is not itself asynchronous.

here is a good write up, in case you are curious in reading further :) http://exploringjs.com/es6/ch_generators.html

src/workers/parser/tests/steps.spec.js
```diff
@@ -2,34 +2,53 @@ import { getNextStep } from "../steps";
...
+    });
   });
 
-  it("second call expression", () => {
-    const source = getSource("async");
-    const pausePosition = { line: 9, column: 9, sourceId: "async" };
-    expect(getNextStep(source, pausePosition)).toEqual(null);
+  describe("yield", () => {
+    it("first yield call", () => {
+      const source = getSource("async");
```
codehag: as mentioned above, yield is synchronous, so we should have it in its own file

src/workers/parser/tests/steps.spec.js
```diff
@@ -2,34 +2,53 @@ import { getNextStep } from "../steps";
...
+      const pausePosition = { line: 9, column: 9, sourceId: "async" };
+      expect(getNextStep(source, pausePosition)).toEqual(null);
+    });
   });
 
-  it("second call expression", () => {
-    const source = getSource("async");
-    const pausePosition = { line: 9, column: 9, sourceId: "async" };
-    expect(getNextStep(source, pausePosition)).toEqual(null);
+  describe("yield", () => {
```
codehag: 👍 nice test!

src/workers/parser/steps.js
```diff
@@ -1,30 +1,30 @@
...
-function getAwaitExpression(source: Source, pausedPosition: AstPosition) {
+function getAsyncExpression(source: Source, pausedPosition: AstPosition) {
   const closestPath = getClosestPath(source, pausedPosition);
 
   if (!closestPath) {
     return null;
   }
 
-  if (isAwaitExpression(closestPath)) {
+  if (isAwaitExpression(closestPath) || isYieldExpression(closestPath)) {
```
codehag: this should be separated, or the function renamed, since yield is not async and this would lead to misunderstandings later on.

src/workers/parser/steps.js
```diff
@@ -1,30 +1,30 @@
...
 import type { NodePath } from "babel-traverse";
 
 export function getNextStep(source: Source, pausedPosition: AstPosition) {
-  const awaitExpression = getAwaitExpression(source, pausedPosition);
+  const awaitExpression = getAsyncExpression(source, pausedPosition);
   if (!awaitExpression) {
     return null;
   }
   const awaitStatement = awaitExpression.getStatementParent();
   return getLocationAfterAwaitExpression(awaitStatement, pausedPosition);
```
codehag: this also needs to be renamed. in general it is too long. Maybe `_getNextStep` since that is what it is doing

src/workers/parser/steps.js
codehag: this should be totally renamed i now realize. it should probably be `currentExpression`

src/workers/parser/steps.js
jayjaycross: I ended up going with `getSteppableExpression` for now as separating out the two cases led to more conditional logic living in `getNextStep`.

jayjaycross:
@codehag I think I addressed all your comments, let me know if any other changes are needed!

codehag:
:tada: awesome work!

PR #4277, opened 2017-10-04
modal height and position changed

9andresc:
### Summary of Changes

* modal max-height changed to 40vh to make it scrollable when there are too many items to fit vertically
* Jason asked to put the modal at the top so we did it!

### Test Plan

Example test plan:

- [x] Open the shortcut modal with Shift+CmdOrCtrl+O
- [x] Enjoy!

### DIFF
--- a/src/components/shared/Modal.css
+++ b/src/components/shared/Modal.css
@@ -17,7 +17,7 @@
   overflow-y: scroll;
   background-color: var(--theme-toolbar-background);
   transition: transform 150ms cubic-bezier(0.07, 0.95, 0, 1);
-  box-shadow: 0 0 10px 2px var(--popup-shadow-color);
+  box-shadow: 1px 1px 3px 1px var(--popup-shadow-color);
 }
 
 .modal.entering,
--- a/src/components/shared/Modal.css
+++ b/src/components/shared/Modal.css
@@ -3,23 +3,19 @@
   display: flex;
   flex-direction: column;
   align-items: center;
-  justify-content: flex-start;
   width: 100%;
   height: 100%;
   top: 0;
   left: 0;
-  padding-top: 30px;
   transition: z-index 200ms;
   z-index: 100;
 }
 
 .modal {
-  top: 0;
-  width: 500px;
-  max-height: 40vh;
-  height: 100%;
+  width: 80%;
+  height: auto;
+  overflow-y: scroll;
   background-color: var(--theme-toolbar-background);
-  transform: translateY(-250px);
   transition: transform 150ms cubic-bezier(0.07, 0.95, 0, 1);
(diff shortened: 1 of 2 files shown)


src/components/shared/Modal.css
```diff
@@ -1,19 +1,23 @@
 .modal-wrapper {
   position: fixed;
   display: flex;
+  flex-direction: column;
   align-items: center;
+  justify-content: flex-start;
```
codehag: we don't need this property since it is the default. making it explicit as it is here would suggest that it is overridden somewhere else.

src/components/shared/Modal.css
```diff
@@ -1,19 +1,23 @@
...
   transition: z-index 200ms;
   z-index: 100;
 }
 
 .modal {
   top: 0;
-  margin: auto;
   width: 500px;
+  max-height: 40vh;
+  height: 100%;
```
codehag: we don't need this property, because it is overriden by `max-height: 40vh` in all cases

src/components/shared/Modal.css
```diff
@@ -1,19 +1,23 @@
...
+  padding-top: 30px;
   transition: z-index 200ms;
   z-index: 100;
 }
 
 .modal {
   top: 0;
-  margin: auto;
   width: 500px;
+  max-height: 40vh;
```
codehag: im a bit torn here because i don't think `40vh` is a good solution in this case. when the panel is horizontal, it will be really squished:

screen shot 2017-10-04 at 09 27 52

i would instead recommend that we do a media query for horizontal layout, and override with the vertical layout, like so:

```css

  .modal { // use the smallest screen size as the default size, and override from there
    height: 100%;
    overflow-y: scroll;
  }
@media (min-height: 680px) {
  .modal {
    height: auto; // override height at 100% and prefer content height
    max-height: 80vh; // this is a max height, if the content is too large
  }
}
```

src/components/shared/Modal.css
```diff
@@ -1,19 +1,23 @@
...
   position: fixed;
   display: flex;
+  flex-direction: column;
   align-items: center;
+  justify-content: flex-start;
   width: 100%;
   height: 100%;
   top: 0;
   left: 0;
+  padding-top: 30px;
```
codehag: this will look a bit weird when in horizontal mode. might also justify a media query. it might also make sense to use this property on the modal itself and set the translate to 30px or 0px, since we have that positioning there anyway.

bomsy:
@9andresc How is this going?

codehag:
hi @9andresc i cleaned this up because there are a few things waiting on it! If you have any questions please feel free to ask :)

jasonLaster:
Thanks!

PR #4279, opened 2017-10-04
Extract temp variables for wrapped components in their tests

ahuth:
Follow up to #4260. This PR extracts some temp variables in a couple of the tests.


codehag:
Looks really good, thank you!

PR #4283, opened 2017-10-04
Removes unneeded SVG DOM elements for source tree arrows.

yurydelendik:
Associated Issue: #4168

I cannot properly measure, but for 3k files win is about 10%-20% (e.g. 2500ms->2000ms) during forceUpdate() of the SourceTree.


jasonLaster:
this could really help:

PR #4286, opened 2017-10-04
(wip) Show class functions in the outline!

jasonLaster:
### Summary of Changes

* groups functions in the UI. This goes from showing a list of functions to showing the class name and the functions below that.
* add choo support for the call stack
* supports the most common class patterns

#### Backbone
```js
const TodoView = Backbone.View.extend({
  render: function() {
    console.log("yo")
  }
})
```

#### React 

```js
const TodoClass = createClass({
  render: function() {
    console.log("yo")
  }
})

```

#### Prototypes
```js
function Button () {}

Button.prototype.createElement = function (color) {}
```

### Test Plan

unit test all the things

@yoshuawuyts and i looked into this. He deserves all the credit, i deserve all the blame.


src/components/PrimaryPanes/Outline.js
```diff
@@ -56,10 +56,40 @@ export class Outline extends Component {
...
+    const classFunctions = functions.filter(
+      func => func.name != "anonymous" && !!func.klass
+    );
+
+    if (classFunctions.length == 0) {
+      return null;
+    }
+
+    const klass = classFunctions[0].klass;
+    const klassFunc = find(functions, func => func.name === klass);
```
digitarald: Could `Array::find` work here as well, @jasonLaster ? We seem to have mixed use of lodash and native find across the code.

jasonLaster: Yeah, that's a good find! that would be a nice PR because we have good test coverage here

PR #4295, opened 2017-10-05
docs most-common-issues

Anzumana:
fixed#4167

### DIFF
--- a/docs/most-common-issues.md
+++ b/docs/most-common-issues.md
@@ -1,0 +1,14 @@
+Before you do anything else try to remove the project again from your computer and clone it again
+
+```bash
+rm -rf [yourpath]/debugger.html
+git clone https://github.com/devtools-html/debugger.html
+cd debugger.html
+yarn install
+yarn start
+```
+
+Don't forget to ask in our [slack room][slack] or fill an [issue][yarn-issue] if you get stuck.
+We are always happy to help.
+
+[slack]:https://devtools-html-slack.herokuapp.com/
--- a/docs/most-common-issues.md
+++ b/docs/most-common-issues.md
@@ -1,4 +1,53 @@
-Before you do anything else try to remove the project again from your computer and clone it again
+## Common Issues
+
+* [Node](#node)
+* [Yarn](#yarn)
+* [Repo](#repo)
+* [babel cache](#babel)
+* [global node packages](#global-packages)
+* [firefox plugin](#firefox-plugin)
+* [windows / linux](#windows--linux)
+
+We do our best to have a 3 step / 1 minute install process,
+but we use bleeding edge technology and problems do arise. Here are our
+most seen issues. Please stop by and ask us a question in [slack],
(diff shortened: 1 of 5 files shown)


docs/most-common-issues.md
```diff
@@ -0,0 +1,14 @@
+Before you do anything else try to remove the project again from your computer and clone it again
```
nyrosmith: This should only be mentioned as a last resort.
I think it'd be a good idea to tell the user to check the yarn version and point that there are issues with yarn !== 0.28.4

docs/most-common-issues.md
```diff
@@ -0,0 +1,14 @@
...
+
+```bash
+rm -rf [yourpath]/debugger.html
```
nyrosmith: This is not working on windows and we should avoid just pointing out platform specific solutions.

docs/most-common-issues.md
```diff
@@ -0,0 +1,14 @@
...
+git clone https://github.com/devtools-html/debugger.html
+cd debugger.html
+yarn install
+yarn start
+```
+
+Don't forget to ask in our [slack room][slack] or fill an [issue][yarn-issue] if you get stuck.
```
nyrosmith: It might be a good idea to make our slack more prominent and not mention it at the end of the doc. Let's tell the people we love to help

docs/most-common-issues.md
jasonLaster: lets keep this in the appendix of getting setup

docs/most-common-issues.md
```diff
@@ -0,0 +1,14 @@
...
+We are always happy to help.
+
+[slack]:https://devtools-html-slack.herokuapp.com/
```
jasonLaster: lets mention that you should check your version of node as well.

Also, if that doesn't work.... use nvm to switch to a clean version of node w/ no global dependencies

docs/most-common-issues.md
jasonLaster: should we suggest `yarn nom` somewhere.
also, perhaps checkout master... and git pull --rebase

GarethSharpe:
Quick comment: 
Was not able to get debugger.html running locally due to an issue with my `~/.babel` file. Although I did not know that this file was the root of the problem, `yarn install` and `yarn run` were giving me command prompt error messages. **Discovered the issue was the** `~/.bable` **file being hidden**. Once I unchecked the 'hidden' checkbox, everything worked. After a `yarn install` and `yarn start`, I could even re-hide the file and it would still work. 

Windows 10
node: v6.11.4
yarn: 1.2.0
error: `Error: EPERM: operation not permitted, open 'C:\Users\<username>\.babel.json'`

jasonLaster:
we should also mention:

* global NPM caches
* webpack loading 
* installed firefox plugins

Anzumana:
great additions @jasonLaster

PR #4296, opened 2017-10-05
Support framework frames for Preact

ohana54:
Associated Issue: #2980 

### Summary of Changes

Support framework frames for Preact

### Test Plan

- Run the debugger locally
- Debug Preact TodoMVC (https://preact-todomvc.surge.sh/)

Note:
Since I'm not a regex expert, Preact regex match needs to run before React.
To mitigate that, I added a unit test that makes sure it is not broken.

Screenshot:
image


src/utils/frame.js
```diff
@@ -95,6 +99,11 @@ export function getLibraryFromUrl(frame: Frame) {
     return "jQuery";
   }
 
+  // Needs to remain before "react", otherwise "react" can also match "preact"
```
bomsy: We could make react more specific by maybe changing the react regex to `/^react/i`

ohana54: If I'm not mistaken, we check the entire url, which means `react` can be anywhere in the url and not just in the beginning. We can try to get the filename from the url and then it would be easier.

bomsy: aaaahhhh ... my bad ! ur right. what of this `/(react)/i` instead

jasonLaster:
@ohana54 you missed the best part. Screenshots!

src/utils/frame.js
ohana54: It would create a capture group, but still match `react` as part of `preact`.

ohana54: The closest thing I found is reverse look-ahead (use look-ahead after reversing the string) but I think it's too complex.

bomsy: https://regexr.com/3gt7c

ohana54:
@jasonLaster The one time screenshots are a must :)
Added!

PR #4298, opened 2017-10-05
Updating docs

jfo:
debugging-the-debugger.md included a couple of misleading, outdated
diffs that didn't reflect the current state of the world.

I updated the diffs to work against the current HEAD. 

This might not be the optimal place to put the pause examples but it's [where I ended up when trying to go through the document myself.](https://gist.github.com/jfo/a005b89d926bdd46fec5901a14055f6f)

img

[Here](https://gist.githubusercontent.com/jfo/ae3c69c43069242cd3ca0f96eeec4d8a/raw/8a4e4f6917a44d501bd03e914900f77d9d32ee66/foo.diff) is a diff which can be applied to current master and which illustrates the proposed diffs for the doc. Debugging the debugger and diffing diffs for doc diffs too 😮

### DIFF
--- a/docs/debugging-the-debugger.md
+++ b/docs/debugging-the-debugger.md
@@ -23,18 +23,19 @@ Adding a breakpoint is a critical piece in the inception game...
 Lets make the debugger do something special when a breakpoint is added.
 
 ```diff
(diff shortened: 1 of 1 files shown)


jasonLaster:
@jfo i love it!

Issue #4299, opened 2017-10-05
Update component bound syntax


jasonLaster:
We've been worried about how we bind functions in our components for some time. See this issue #3172. We recently adopted the new class properties syntax as a way to address this in #4266.

We should now update the other components with this pattern:

- [x] CallSite.js
- [x] SearchBar.js
- [x] Frames/Group.js
- [x] Accordion.js
- [x] ManagedTree.js
- [x] Modal.js
- [x] SymbolModal.js

#### before and after:

```diff
    constructor() {
      super();
- 
-     const self: any = this;
-     self.closeConditionalPanel = this.closeConditionalPanel.bind(this);
-     self.onEscape = this.onEscape.bind(this);
-     self.onGutterClick = this.onGutterClick.bind(this);
-     self.onGutterContextMenu = this.onGutterContextMenu.bind(this);
-     self.onSearchAgain = this.onSearchAgain.bind(this);
-     self.onToggleBreakpoint = this.onToggleBreakpoint.bind(this);
-     self.toggleConditionalPanel = this.toggleConditionalPanel.bind(this);
    }
    
-   onEscape(key, e) {
+   onEscape = (key, e) => {
      if (!this.state.editor) {
        return;
      }
  
      const { codeMirror } = this.state.editor;
      if (codeMirror.listSelections().length > 1) {
        codeMirror.execCommand("singleSelection");
        e.preventDefault();
      }
-   }

-   onSearchAgain(_, e) {
+   onSearchAgain = (_, e) => {
      const { query, searchModifiers } = this.props;
      const { editor: { codeMirror } } = this.state.editor;
      const ctx = { ed: this.state.editor, cm: codeMirror };
  
      const direction = e.shiftKey ? "prev" : "next";
      traverseResults(e, ctx, query, direction, searchModifiers.toJS());
-   }
+   };
```

ohana54:
I'm on it

codehag:
Thanks @ohana54 !

maniator:
:+1:

PR #4302, opened 2017-10-05
Change component function binding to class properties

ohana54:
Associated Issue: #4299

### Summary of Changes

Changed the following components to have functions as class properties instead of using bind:
CallSite.js
SearchBar.js
Frame.js
Group.js
Accordion.js
ManagedTree.js
Modal.js
SymbolModal.js

### Test Plan

`yarn prepush`

### DIFF
--- a/src/components/shared/Modal.js
+++ b/src/components/shared/Modal.js
@@ -17,15 +17,9 @@ type ModalProps = {
 export class Modal extends Component {
   props: ModalProps;
 
-  constructor(props: ModalProps) {
-    super(props);
-    const self: any = this;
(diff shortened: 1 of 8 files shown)


src/components/shared/Modal.js
```diff
@@ -19,13 +19,11 @@ export class Modal extends Component {
 
   constructor(props: ModalProps) {
```
bomsy: we won't need the constructor here anymore .. right?

ohana54: Whoops, was on autopilot towards the end and missed it. Fixing now.

Issue #4061, opened 2017-09-18
[SourceTree] alt click should expand/collapse all the nodes


jasonLaster:
It's a common pattern that <alt> clicking on a node in a tree will expand it and all of its descendant nodes. Conversely, if it's open, it will close the node and all of its descendent nodes.

Examples:
* mac finder: alt click will expand a folder and it's children and children's folders
* chrome: element tree and object tree work this way too

PR #4309, opened 2017-10-06
Make alt click expand/collapse all the nodes

Fischer-L:
Associated Issue: #4061

### Summary of Changes
Expand/Collapse all nodes in the source tree when doing alt click on the tree item(folder)

### Screenshots
4061-altkey-expand-collapse-all-nodes

### DIFF
--- a/src/components/shared/ManagedTree.js
+++ b/src/components/shared/ManagedTree.js
@@ -1,5 +1,6 @@
 // @flow
 import React, { createFactory, Component } from "react";
+import { nodeHasChildren } from "../../utils/sources-tree";
 import "./ManagedTree.css";
 
 import { Tree as _Tree } from "devtools-components";
@@ -68,13 +69,35 @@ class ManagedTree extends Component {
     }
   }
 
-  setExpanded = (item: Item, isExpanded: boolean) => {
+  setExpanded = (item: Item, isExpanded: boolean, isRecursive: boolean) => {
     const expanded = this.state.expanded;
-    const itemPath = this.props.getPath(item);
+    let itemPath = this.props.getPath(item);
     if (isExpanded) {
       expanded.add(itemPath);
     } else {
       expanded.delete(itemPath);
+    }
+    if (isRecursive) {
+      let children = null;
+      let parents = [item];
+      while (parents.length) {
+        children = [];
+        for (const parent of parents) {
+          if (!nodeHasChildren(parent)) {
+            continue;
+          }
+          for (const child of parent.contents) {
+            itemPath = this.props.getPath(child);
+            if (isExpanded) {
+              expanded.add(itemPath);
+            } else {
+              expanded.delete(itemPath);
+            }
+            children.push(child);
+          }
+        }
+        parents = children;
+      }
     }
     this.setState({ expanded });
 
@@ -125,8 +148,8 @@ class ManagedTree extends Component {
       isExpanded: item => expanded.has(this.props.getPath(item)),
       focused: focusedItem,
       getKey: this.props.getPath,
-      onExpand: item => this.setExpanded(item, true),
-      onCollapse: item => this.setExpanded(item, false),
+      onExpand: item => this.setExpanded(item, true, false),
+      onCollapse: item => this.setExpanded(item, false, false),
       onFocus: this.focusItem,
       renderItem: (...args) =>
         this.props.renderItem(...args, {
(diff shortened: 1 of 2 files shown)


jasonLaster:
Nice @Fischer-L!
@amitzur?

amitzur:
I didn't know this feature existed in other places like finder and chrome!
Re UX I think it's proper to back this by a button, possibly in a top bar on the right pane, where a "scroll from source" button can also exist. Just a thought

Fischer-L:
@amitzur 
> I didn't know this feature existed in other places like finder and chrome!
> Re UX I think it's proper to back this by a button, possibly in a top bar on the right pane, where a "scroll from source" button can also exist. Just a thought
Yes this alt click feature exists on Chrome's devtool, finder and Sublime, which is handy.

src/components/shared/ManagedTree.js
```diff
@@ -68,14 +69,36 @@ class ManagedTree extends Component {
     }
   }
 
-  setExpanded = (item: Item, isExpanded: boolean) => {
+  setExpanded = (item: Item, isExpanded: boolean, isRecursive: boolean) => {
```
bomsy: i'm thinking something like `shouldIncludeChildren` feels better and clearer then `isRecursive`

src/components/shared/ManagedTree.js
```diff
@@ -68,14 +69,36 @@ class ManagedTree extends Component {
...
     const expanded = this.state.expanded;
-    const itemPath = this.props.getPath(item);
+    let itemPath = this.props.getPath(item);
     if (isExpanded) {
       expanded.add(itemPath);
     } else {
       expanded.delete(itemPath);
     }
+    if (isRecursive) {
+      let children = null;
```
bomsy: we can remove this line ...

src/components/shared/ManagedTree.js
```diff
@@ -68,14 +69,36 @@ class ManagedTree extends Component {
...
+      let parents = [item];
+      while (parents.length) {
+        children = [];
```
bomsy: ... and make this `const children = [];`

src/components/shared/ManagedTree.js
```diff
@@ -68,14 +69,36 @@ class ManagedTree extends Component {
...
-  setExpanded = (item: Item, isExpanded: boolean) => {
+  setExpanded = (item: Item, isExpanded: boolean, isRecursive: boolean) => {
     const expanded = this.state.expanded;
-    const itemPath = this.props.getPath(item);
+    let itemPath = this.props.getPath(item);
     if (isExpanded) {
       expanded.add(itemPath);
     } else {
       expanded.delete(itemPath);
     }
```
bomsy: We could pull this into a local function like this here
```js 
    function toggleItem() {
      const path = this.props.getPath(item);
      if (isExpanded) {
        expanded.add(path);
      } else {
        expanded.delete(path);
      }
    };
```

bomsy: Then  do `toggleItem(item);` here

src/components/shared/ManagedTree.js
```diff
@@ -68,14 +69,36 @@ class ManagedTree extends Component {
...
+          }
+          for (const child of parent.contents) {
+            itemPath = this.props.getPath(child);
+            if (isExpanded) {
+              expanded.add(itemPath);
+            } else {
+              expanded.delete(itemPath);
+            }
+            children.push(child);
+          }
```
bomsy: Then this whole for loop becomes
```js
          for (const child of parent.contents) {
            toggleItem(child);
            children.push(child);
          }
```

src/components/shared/ManagedTree.js
```diff
@@ -68,14 +69,36 @@ class ManagedTree extends Component {
...
+        }
```
bomsy: This for loop can become
```js
        for (const parent of parents) {
          if (nodeHasChildren(parent)) {
            for (const child of parent.contents) {
              toggleItem(child);
              children.push(child);
            }
          }
        }
```

bomsy: We should not need the `continue` ...

bomsy: And may not need `nodeHasChildren` since we only use `parent.contents` anyway, we can just check that

src/components/shared/ManagedTree.js
```diff
@@ -1,5 +1,6 @@
 // @flow
 import React, { createFactory, Component } from "react";
+import { nodeHasChildren } from "../../utils/sources-tree";
```
bomsy: We can remove this line

bomsy:
@Fischer-L Great Stuff here! left some comments..

bomsy:
@Fischer-L i fixed up the few comments so we could landing this ...

Thanks for the amazing Work!!!!

PR #4321, opened 2017-10-07
Improve triaging docs

jasonLaster:
- reference steve's article
- outline the steps for making a bug actionable

### DIFF
--- a/docs/maintainer.md
+++ b/docs/maintainer.md
@@ -2,13 +2,17 @@
 
 Helping maintain a project is the best way to contribute to its overall health.
 
-+ [Pushing to a branch](#pushing-to-a-branch)
-+ [Triaging issues](#triaging-issues)
++ [Triaging Issues](#triaging-issues)
++ [Making Bugs Actionable](#making-bugs-actionable)
++ [Reviewing Stale Issues](#reviewing-stale-issues)
++ [Making Issues available](#making-issues-available)
++ [Following up on In Progress work](#following-up-on-in-progress-work)
 + [Adding a Patch](#adding-a-patch)
++ [Pushing to a branch](#pushing-to-a-branch)
 
 ### Triaging Issues
 
-We encourage the community to help make bugs actionale, make features available,
+We encourage the community to help make bugs actionable, make features available,
  and close stale issues. Triaging is one of the most important contributions a
 community member can make for a projects health. Stave Klabnik's
 [article][gardening] on how to be an open source gardener expresses our values.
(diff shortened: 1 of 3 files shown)


docs/maintainer.md
```diff
@@ -17,7 +32,7 @@ We define *stale* as issues that are 30 days or older. Stale is often an indicat
 * *broad* - create more focused issues, such as "add flow types to `WhyPaused` component"
 * *low priority* - close it and add it to the [enhancements board]
 
-#### Making Issues `available`
+#### Making features `available`
```
bomsy: features?

jasonLaster: i wanted to differentiate btw bugs and features... but it might be a bad dstinction...

docs/maintainer.md
```diff
@@ -8,6 +8,21 @@ Helping maintain a project is the best way to contribute to its overall health.
...
+
+#### Making Bugs Actionable
+
+Bugs are labeled [`not-actionable`][na] when they are filed. Here is a checklist
+for making bugs actionable:
+
+1. Check for duplicate bugs
+2. Verify that the bug is reproducable
+3. Define steps to reproduce with expected and actual results
+4. Create a small test case. We have a [glitch] template for sharing examples.
```
andreicristianpetcu: I don't know how glitch works. Can you please add a link to a Glitch tutorial?

docs/maintainer.md
bomsy: yeah ...  i think of bugs/ features as all issues, we probably need to make the distinction clearer in the doc . If we are changing it here, might need to re-word the section below it as well.

docs/maintainer.md
```diff
@@ -8,6 +8,21 @@ Helping maintain a project is the best way to contribute to its overall health.
...
+
 #### Closing Stale Issues
```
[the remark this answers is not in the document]
jasonLaster: Hmm, lets rename this "Review Stale Issues"

jnachtigall:
FWIW, I noticed that this maintainer.md file is nowhere linked to as is. That is, as a whole like a chapter in docs. I can only find that it is linked to certain parts, see https://github.com/devtools-html/debugger.html/search?utf8=%E2%9C%93&q=maintainer.md&type=  

I guess it's meant for new fellow maintainers and PR reviewers and as such I would expect it to be linked for, say, [CONTRIBUTING.md at Maintainer tips](https://github.com/devtools-html/debugger.html/blob/master/CONTRIBUTING.md#maintainer-tips) should link to it as a whole I guess

docs/maintainer.md
jnachtigall: Why restrict the process to features (enhancements) and not also to bugs? Bug reports can also go stale, not just feature request. So I'd stick to "Issues" here instead of only "feature requests".

docs/maintainer.md
jnachtigall: I think that there might also be a 5th step for "Categorize the issue report with a suitable tag" linking to kind of an overview on what you have ('needs-ux', 'performance'). FWIW, I cannot find such an overview. This could be just a section of contributing.md or - well, even better - in this maintainer.md ;)

Issue #4210, opened 2017-09-29
goto line functionality


codehag:
a user has requested goto line functionality similar to the old debugger

something like adding `:100` in the search field to go to line 100

this should be easy to implement using the selectLocation functionality using the current source. 

### GIST

```diff
diff --git a/src/components/App.js b/src/components/App.js
index 1dd5132..ed7e64e 100644
--- a/src/components/App.js
+++ b/src/components/App.js
@@ -100,6 +100,13 @@ class App extends Component {
       L10N.getStr("symbolSearch.search.key2"),
       this.toggleSymbolModal
     );
+
+    // cmd g should toggle go to line
+    shortcuts.on(
+      L10N.getStr("symbolSearch.search.key2"),
+      this.toggleSymbolModal
+    );
+
     shortcuts.on("Escape", this.onEscape);
   }
 
@@ -143,6 +150,30 @@ class App extends Component {
     setActiveSearch("symbol");
   }
 
+  toggleGoToLineModal(_, e: SyntheticEvent) {
+    const {
+      selectedSource,
+      activeSearch,
+      closeActiveSearch,
+      setActiveSearch
+    } = this.props;
+
+    e.preventDefault();
+    e.stopPropagation();
+
+    // this is beautiful. we don't wan to topen if there is no source :)
+    // yay
+    if (!selectedSource) {
+      return;
+    }
+
+    if (activeSearch == "symbol") {
+      return closeActiveSearch();
+    }
+
+    setActiveSearch("line");
+  }
+
   onLayoutChange() {
     if (isVisible()) {
       this.setState({ horizontal: verticalLayoutBreakpoint.matches });
diff --git a/src/components/SymbolModal.js b/src/components/SymbolModal.js
index e515a1b..9db62fa 100644
--- a/src/components/SymbolModal.js
+++ b/src/components/SymbolModal.js
@@ -1,5 +1,10 @@
 // @flow
 
+// copy this file and re-name it to go to line
+
+// the key detail is that enter should call the action `selectSource` with the current
+// selected source and the line
+
 import PropTypes from "prop-types";
 import React, { Component } from "react";
 import { connect } from "react-redux";
```

jasonLaster:
Agreed, lets find a good UI for this. Chrome uses `ctrl+g` which is the standard shortcut. They also use their modal, which is a good pattern here. I think they had an alert box until recently :)

[...]

PR #4323, opened 2017-10-07
add a GotoLineModal

jfo:
Associated Issue: #4210 

### Summary of Changes

This commit adds a gotoline modal/functionality triggered with `:` in
the main debugger source window.

- [x] `Cmd+:` opens the modal
- [x] Entering a line number takes you to that line in the current source
- [x] unusable input (i.e. letters) result in a noop

goto

This is not ready to merge but it is ready for feedback. I've commented on the diff with some things! Also needs tests, would like some guidance on that via [SymbolModal.spec.js](https://github.com/devtools-html/debugger.html/blob/master/src/components/tests/SymbolModal.spec.js), I assume.

### DIFF
--- a/assets/panel/debugger.properties
+++ b/assets/panel/debugger.properties
@@ -643,6 +643,11 @@ watchExpressionsSeparatorLabel2=\u0020→
 # and its real name (if available).
 functionSearchSeparatorLabel=←
 
+# LOCALIZATION NOTE(gotoLineModal.placeholder): The placeholder
+# text displayed when the user searches for specific lines in a file
+gotoLineModal.placeholder=Go to line…
+gotoLineModal.commandKey=CmdOrCtrl+Shift+;
+
 # LOCALIZATION NOTE(symbolSearch.search.functionsPlaceholder): The placeholder
 # text displayed when the user searches for functions in a file
 symbolSearch.search.functionsPlaceholder=Search functions…
--- a/src/components/App.js
+++ b/src/components/App.js
@@ -47,6 +47,8 @@ import EditorTabs from "./Editor/Tabs";
 
 import SymbolModal from "./SymbolModal";
 
+import GotoLineModal from "./GotoLineModal";
+
 type Props = {
   selectSource: Function,
   selectedSource: SourceRecord,
@@ -71,6 +73,7 @@ class App extends Component {
   renderEditorPane: Function;
   renderVerticalLayout: Function;
   toggleSymbolModal: Function;
+  toggleGoToLineModal: Function;
   onEscape: Function;
   onCommandSlash: Function;
 
@@ -86,6 +89,7 @@ class App extends Component {
     this.getChildContext = this.getChildContext.bind(this);
     this.onLayoutChange = this.onLayoutChange.bind(this);
     this.toggleSymbolModal = this.toggleSymbolModal.bind(this);
+    this.toggleGoToLineModal = this.toggleGoToLineModal.bind(this);
     this.renderEditorPane = this.renderEditorPane.bind(this);
     this.renderVerticalLayout = this.renderVerticalLayout.bind(this);
     this.onEscape = this.onEscape.bind(this);
@@ -98,10 +102,17 @@ class App extends Component {
 
   componentDidMount() {
     verticalLayoutBreakpoint.addListener(this.onLayoutChange);
+
     shortcuts.on(
       L10N.getStr("symbolSearch.search.key2"),
       this.toggleSymbolModal
     );
+
+    shortcuts.on(
+      L10N.getStr("gotoLineModal.commandKey"),
+      this.toggleGoToLineModal
+    );
+
     shortcuts.on("Escape", this.onEscape);
     shortcuts.on("Cmd+/", this.onCommandSlash);
   }
@@ -112,6 +123,9 @@ class App extends Component {
       L10N.getStr("symbolSearch.search.key2"),
       this.toggleSymbolModal
     );
+
+    shortcuts.off(L10N.getStr("gotoLineModal.key"), this.toggleGoToLineModal);
+
     shortcuts.off("Escape", this.onEscape);
   }
 
@@ -148,6 +162,28 @@ class App extends Component {
     }
 
     setActiveSearch("symbol");
+  }
+
+  toggleGoToLineModal(_, e: SyntheticEvent) {
+    const {
+      selectedSource,
+      activeSearch,
+      closeActiveSearch,
+      setActiveSearch
+    } = this.props;
+
+    e.preventDefault();
+    e.stopPropagation();
+
+    if (!selectedSource) {
+      return;
+    }
+
+    if (activeSearch == "line") {
+      return closeActiveSearch();
+    }
+
+    setActiveSearch("line");
   }
 
   onLayoutChange() {
@@ -275,6 +311,21 @@ class App extends Component {
     );
   }
 
+  renderGotoLineModal() {
+    const { selectSource, selectedSource, activeSearch } = this.props;
+
+    if (activeSearch !== "line") {
+      return;
+    }
+
+    return (
+      <GotoLineModal
+        selectSource={selectSource}
+        selectedSource={selectedSource}
+      />
+    );
+  }
+
   renderShortcutsModal() {
     const additionalClass = isMacOS ? "mac" : "";
 
@@ -298,6 +349,7 @@ class App extends Component {
           ? this.renderHorizontalLayout()
           : this.renderVerticalLayout()}
         {this.renderSymbolModal()}
+        {this.renderGotoLineModal()}
         {this.renderShortcutsModal()}
       </div>
     );
--- a/src/components/GotoLineModal.js
+++ b/src/components/GotoLineModal.js
@@ -1,0 +1,123 @@
+// @flow
+
+import React, { Component } from "react";
+import { connect } from "react-redux";
+import { bindActionCreators } from "redux";
+import { getActiveSearch, getSelectedSource } from "../selectors";
+import actions from "../actions";
+
+import Modal from "./shared/Modal";
+
+import SearchInput from "./shared/SearchInput";
+
+import type { SourceRecord } from "../reducers/sources";
+import type { SelectSourceOptions } from "../actions/sources";
+
+type GotoLineModalState = {
+  query: ?string
+};
+
+import "./SymbolModal.css";
+
+class GotoLineModal extends Component {
+  state: GotoLineModalState;
+
+  props: {
+    enabled: boolean,
+    selectSource: (string, ?SelectSourceOptions) => void,
+    selectedSource?: SourceRecord,
+    closeActiveSearch: () => void,
+    highlightLineRange: ({ start: number, end: number }) => void,
+    clearHighlightLineRange: () => void
+  };
+
+  constructor(props) {
+    super(props);
+    this.state = { query: "" };
+  }
+
+  onClick = (e: SyntheticEvent) => {
+    e.stopPropagation();
+  };
+
+  onChange = (e: SyntheticInputEvent) => {
+    const { selectedSource } = this.props;
+    if (!selectedSource || !selectedSource.get("text")) {
+      return;
+    }
+
+    this.setState({ query: e.target.value });
+  };
+
+  closeModal = () => {
+    this.props.closeActiveSearch();
+    this.props.clearHighlightLineRange();
+  };
+
+  onKeyUp = (e: SyntheticKeyboardEvent) => {
+    e.preventDefault();
+    const { selectSource, selectedSource, enabled } = this.props;
+    const { query } = this.state;
+
+    if (!enabled || !selectedSource) {
+      return;
+    }
+
+    if (e.key === "Enter" && query != null) {
+      const linenumber = parseInt(query.replace(/[^\d+]/g, ""), 10);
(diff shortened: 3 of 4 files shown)


assets/panel/debugger.properties
```diff
@@ -648,6 +648,14 @@ watchExpressionsSeparatorLabel2=\u0020→
...
 functionSearchSeparatorLabel=←
 
+# LOCALIZATION NOTE(gotoLineModal.placeholder): The placeholder
+# text displayed when the user searches for specific lines in a file
+gotoLineModal.placeholder=Go to line...
+
+# LOCALIZATION NOTE(symbolSearch.search.key2): The Key Shortcut for
+# searching for a function or variable
+gotoLineModal.key=:
+
```
jfo: idk if this is the right place for this.

src/components/App.js
```diff
@@ -47,6 +47,8 @@ import EditorTabs from "./Editor/Tabs";
 
 import SymbolModal from "./SymbolModal";
 
+import GotoLineModal from "./GotoLineModal";
```
jfo: this whole pattern is from @codehag's diff on the associated issue. 

It feels like these (SymbolModal, ShortcutModal, GotoLineModal) should all be the same thing, but I thought this would be a good start before merging them together.

src/components/GotoLineModal.js
```diff
@@ -0,0 +1,118 @@
...
+import SearchInput from "./shared/SearchInput";
+
+import type { SourceRecord } from "../reducers/sources";
+import type { SelectSourceOptions } from "../actions/sources";
+
+type GotoLineModalState = {
+  query: ?string
+};
+
+import "./SymbolModal.css";
```
jfo: I'm pretty sure this file is actually not being used anywhere (it's also called in `SymbolModal.js` but the `.symbol-modal-wrapper` and `.symbol-modal` classes are never applied. I might suggest merging the [remaining rules](https://github.com/devtools-html/debugger.html/blob/master/src/components/SymbolModal.css#L18-L31) into `shared/Modal.css` instead but that's just outside the scope of this pr.

assets/panel/debugger.properties
```diff
@@ -648,6 +648,14 @@ watchExpressionsSeparatorLabel2=\u0020→
 # and its real name (if available).
 functionSearchSeparatorLabel=←
 
+# LOCALIZATION NOTE(gotoLineModal.placeholder): The placeholder
+# text displayed when the user searches for specific lines in a file
+gotoLineModal.placeholder=Go to line...
```
flodolo: Use …, not ...

assets/panel/debugger.properties
```diff
@@ -648,6 +648,14 @@ watchExpressionsSeparatorLabel2=\u0020→
...
+
+# LOCALIZATION NOTE(symbolSearch.search.key2): The Key Shortcut for
```
flodolo: Please stop copying localization comments without updating them

assets/panel/debugger.properties
```diff
@@ -648,6 +648,14 @@ watchExpressionsSeparatorLabel2=\u0020→
...
+# searching for a function or variable
+gotoLineModal.key=:
```
flodolo: Does this even work? IIRC you need to press shift to get to that keyboard in English, and that's a modifier.

jfo: It does work. There is an attached gif.

codehag: Hi, thanks for helping our contributors with their prs. I would like to point out that the wording of this comment can be read as aggressive, and we want our contributors to feel safe and that their changes are welcome. Phrasing such as "Does this even work" can be off-putting for a contributor, and can also turn away potential contributors who are looking through our PRs to get a sense of the community. It is especially problematic to have this on something that is clearly labelled as a work in progress, since the person may be looking for feedback.

If you find something amiss, think of how to word things in a way that are not judgemental or polarizing.

jfo: Sorry. Yes. You do have to press shift to get to that symbol, but this syntax works to trigger that. In any case, it's a temporary key that can be changed to anything that makes the most sense.

assets/panel/debugger.properties
codehag: A more constructive way to say this would be "Please update this localization note to reflect the new key"

flodolo: It wasn't meant as an hostile comment towards the author, if it came across like that, I apologize.

For context: this is happening in several PRs every week, so I think there's a general issue to solve. Commenting strings is not mandatory, comments are there to give context to translators. On the other hand, I have the feeling most developers think they are, but copying and paste an existing comment is not the solution.

assets/panel/debugger.properties
flodolo: As above, this comment wasn't meant as hostile or derogatory. 

I can't know in which state a PR is when it's open (I'm not a developer, I work on localization), if it's testable or just an early WIP (it looked like that), hence the question "Does : this even work?". There was no underlying implication, i.e. "are you even testing your code?", in my question.

I don't know exactly how devtools or debugger manage shortcuts internally, but I don't think it would work for a feature in Firefox, since we stumbled upon it for shortcuts on special characters like [ or ]

assets/panel/debugger.properties
```diff
@@ -648,6 +648,14 @@ watchExpressionsSeparatorLabel2=\u0020→
 # and its real name (if available).
 functionSearchSeparatorLabel=←
 
+# LOCALIZATION NOTE(gotoLineModal.placeholder): The placeholder
+# text displayed when the user searches for specific lines in a file
+gotoLineModal.placeholder=Go to line…
+
+# LOCALIZATION NOTE(gotoLineModal.key): The Key Shortcut for
```
flodolo: You still need to update the comment itself (you only updated the reference to the string ID, but the text describes a different function).

At this point I would suggest to drop this comment completely, put this string near the one above (that already has a comment explaining the function), and call this string `gotoLineModal.commandkey` to avoid any potential confusion. 

There's no rule set in stone for key naming, but `.key` is more prone to misinterpretation that `.commandkey`.

assets/panel/debugger.properties
codehag: thanks for clarifying! And again thanks for helping :)

assets/panel/debugger.properties
jfo: Oh that's a good idea.

To reiterate, `Cmd+:` is definitely temporary, just needed something to use for testing b/c the suggested (in the linked issue) `Cmd+g` is already being used. I think ultimately just `:` would be a good shortcut _from an already opened modal_ but definitely not as a global shortcut key.

assets/panel/debugger.properties
jfo: It did come across like that, so thank you for apologizing. I have no knowledge of what other PR's look like (this is my very first one of any substance) or that the comments are optional, I'm just trying to match what's around the usage with the expectation of being corrected if necessary, so thanks for catching it, it was an oversight.

assets/panel/debugger.properties
jfo: It took me a while to figure out what you meant by this, that the first is a unicode character is hard to discern from the text. Putting them in `pre` fences would help distinguish them i.e.
 `...` vs `…`.

assets/panel/debugger.properties
flodolo: Good point, I'll keep it in mind.

assets/panel/debugger.properties
flodolo: Good point, I'll keep it in mind.

(in case you're seeing this comment twice, I re-added comment in the right thread).

src/components/GotoLineModal.js
```diff
@@ -0,0 +1,118 @@
...
+
+    if (e.key === "Enter" && query != null) {
+      const linenumber = parseInt(query.replace(/[^\d+]/g, ""), 10);
+      if (!isNaN(linenumber)) {
+        selectSource(selectedSource.get("id"), { line: linenumber });
+      }
+      this.closeModal();
+    } else if (e.key === "Tab") {
+      this.closeModal();
+    }
```
jasonLaster: I think we might want to handle "Escape" too

src/components/GotoLineModal.js
```diff
@@ -0,0 +1,118 @@
...
+      return;
+    }
+
+    if (e.key === "Enter" && query != null) {
+      const linenumber = parseInt(query.replace(/[^\d+]/g, ""), 10);
+      if (!isNaN(linenumber)) {
+        selectSource(selectedSource.get("id"), { line: linenumber });
+      }
+      this.closeModal();
+    } else if (e.key === "Tab") {
```
jasonLaster: this is just a style thing, but i tend to prefer early returns

```js
if (e.key === "Enter" && query != null) {
 if (!isNaN(linenumber)) {
   selectSource(selectedSource.get("id"), { line: linenumber })
 }
 return this.closeModal();
}

if (e.key === "Tab") {
  return this.closeModal();
}
```

assets/panel/debugger.properties
```diff
@@ -648,6 +648,14 @@ watchExpressionsSeparatorLabel2=\u0020→
...
+# searching for a specific line
+gotoLineModal.key=:
```
jasonLaster: I'm surprised that this is not `cmd+:`

src/components/App.js
jasonLaster: Yeah... we do want to consolidate the symbol + gotoline modal and search for file.

+ `cmd+p`
+ `cmd+:`
+ `cmd+shift+o`

src/components/App.js
```diff
@@ -150,6 +161,28 @@ class App extends Component {
...
+
+    if (!selectedSource) {
+      return;
+    }
+
+    if (activeSearch == "line") {
+      return closeActiveSearch();
+    }
+
+    setActiveSearch("line");
```
jasonLaster: we'll likely need to update the flow definition for the possible search types in `reducers/ui.js`

jasonLaster:
by the way, we have an open PR to fix the position of the modal: https://github.com/devtools-html/debugger.html/pull/4277

it should always be pinned to the top...

src/components/GotoLineModal.js
wldcordeiro: It looks the those rules are applied?

https://github.com/devtools-html/debugger.html/blob/master/src/components/SymbolModal.js#L227 `.input-wrapper`

and then the result list? https://github.com/devtools-html/debugger.html/blob/master/src/components/SymbolModal.js#L213

jfo: Sorry, I wasn't clear, I meant that only `.symbol-modal-wrapper` and `.symbol-modal` classes don't seem to be applied and the remaining rules that I linked to might be incorporated into `Modal.css` in the process of merging the modals together more, but I'm not planning to do that on this PR.

src/components/GotoLineModal.js
jfo: Escape already closes out the modal in both SymbolModal and GotoLineModal, it's handled more globally [here](https://github.com/devtools-html/debugger.html/blob/master/src/actions/ui.js#L7) via [here](https://github.com/devtools-html/debugger.html/blob/master/src/components/App.js#L118) and [here](https://github.com/devtools-html/debugger.html/blob/master/src/components/App.js#L105).

src/components/App.js
jfo: do you mean [here](https://github.com/devtools-html/debugger.html/blob/master/src/reducers/ui.js#L21)? I am surprised flow did not catch that for me in a warning... I guess I don't really know how flow works 😃

src/components/App.js
jfo: - [cmd+p](https://github.com/devtools-html/debugger.html/blob/master/assets/panel/debugger.properties#L141)
- [cmd+shift+o](https://github.com/devtools-html/debugger.html/blob/master/assets/panel/debugger.properties#L153)

For those shortcuts... `CmdOrCtrl+Shift+;` works, but `CmdOrCtrl+:` does not work, nor does `CmdOrCtrl+Shift+:`. Is this correct? This could be revisted when you unify them, whatever you'd like for right now is fine by me :)

src/components/GotoLineModal.js
jfo: I'll defer to you on that 😃 ! But, as those return values aren't being dispatched on in the caller, I'd prefer 

```js
if (e.key === "Enter" && query != null) {
 if (!isNaN(linenumber)) {
   selectSource(selectedSource.get("id"), { line: linenumber })
 }
 this.closeModal();
 return;
}

if (e.key === "Tab") {
  this.closeModal();
  return;
}
```

When I was moving some logic to this from `SymbolModal.js` I had trouble grokking [this line](https://github.com/devtools-html/debugger.html/blob/master/src/components/SymbolModal.js#L109), because I assumed the `return` value was significant. It can't be, since [updateResults](https://github.com/devtools-html/debugger.html/blob/master/src/components/SymbolModal.js#L131-L145) always returns undefined, but semantically that was unclear.

Is there a technical reason to avoid `else if` in favor of multiple `if` blocks? Like a branching implementation thing?

jfo:
Thanks all for the good feedback! Will this require tests?

jasonLaster:
@jfo i don't think so

jfo:
Hi all, I've addressed the outstanding reviews and rebased against master, I think this should be ready to merge now but let me know if there is anything else needed.

Thanks a lot! 🙌

jasonLaster:
🎁

Issue #4346, opened 2017-10-11
threadClient._parent.listWorkers is not a function


codehag:
Steps to Reproduce:

1) open the browser toolbox and select the debugger

expected:: debugger to appear
Actual: no debugger

-- in the console there is the following error: 

```
TypeError: threadClient._parent.listWorkers is not a function
Stack trace:
exports.onConnect</_ref<@resource://devtools/shared/base-loader.js -> resource://devtools/client/debugger/new/debugger.js:30385:5
step@resource://devtools/shared/base-loader.js -> resource://devtools/client/debugger/new/debugger.js:30420:183
step/<@resource://devtools/shared/base-loader.js -> resource://devtools/client/debugger/new/debugger.js:30420:361
promise callback*step@resource://devtools/shared/base-loader.js -> resource://devtools/client/debugger/new/debugger.js:30420:314
_asyncToGenerator/</<@resource://devtools/shared/base-loader.js -> resource://devtools/client/debugger/new/debugger.js:30420:437
_asyncToGenerator/<@resource://devtools/shared/base-loader.js -> resource://devtools/client/debugger/new/debugger.js:30420:99
onConnect@resource://devtools/shared/base-loader.js -> resource://devtools/client/debugger/new/debugger.js:30410:12
onConnect</_ref<@resource://devtools/shared/base-loader.js -> resource://devtools/client/debugger/new/debugger.js:30266:23
step@resource://devtools/shared/base-loader.js -> resource://devtools/client/debugger/new/debugger.js:30320:183
_asyncToGenerator/</<@resource://devtools/shared/base-loader.js -> resource://devtools/client/debugger/new/debugger.js:30320:437
_asyncToGenerator/<@resource://devtools/shared/base-loader.js -> resource://devtools/client/debugger/new/debugger.js:30320:99
onConnect@resource://devtools/shared/base-loader.js -> resource://devtools/client/debugger/new/debugger.js:30300:12
bootstrap@resource://devtools/shared/base-loader.js -> resource://devtools/client/debugger/new/debugger.js:120:15
open@resource://devtools/shared/base-loader.js -> resource://devtools/client/debugger/new/panel.js:29:15
async*onLoad@resource://devtools/shared/base-loader.js -> resource://devtools/client/framework/toolbox.js:1665:19
```

PR #4347, opened 2017-10-11
fix browser toolbox issues

codehag:
related issue: #4346 

There was an issue with debuggerUrl not being defined in the browser toolbox context and listWorkers being present but not a function on the threadClient._parent

both need more investigation, this is just to get the browser toolbox working again.

### DIFF
--- a/src/utils/sources-tree/treeOrder.js
+++ b/src/utils/sources-tree/treeOrder.js
@@ -9,7 +9,11 @@ import type { Node } from "./types";
 /*
  * Gets domain from url (without www prefix)
  */
-export function getDomain(url: string): ?string {
+export function getDomain(url?: string): ?string {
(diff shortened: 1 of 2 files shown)


src/utils/sources-tree/treeOrder.js
```diff
@@ -9,7 +9,11 @@ import type { Node } from "./types";
 /*
  * Gets domain from url (without www prefix)
  */
-export function getDomain(url: string): ?string {
+export function getDomain(url?: string): ?string {
+  // TODO: define how files should be ordered on the browser debugger
+  if (!url) {
+    return null;
+  }
```
jasonLaster: should we have a test for this?

jasonLaster: landing because i want to do a release

Issue #4350, opened 2017-10-11
Switch to CSS SVGs


jasonLaster:
We would like to switch from using react-based inline SVGs to CSS-based SVGs. The primary advantage is performance. SVGs created by React are slow to render. CSS drawn SVGs have a significant perf benefit.

Todolist:

- [x] convert source tree arrows #4351
- [ ] convert Object Inspector arrows *(will be done in devtools-core)*
- [x] convert source tree domain / file / folder
- [ ] convert close button (tabs, breakpoints, watch expressions)
- [x] convert step buttons
- [x] convert blackbox / pretty print

jasonLaster:
### Gist for Pretty Print / Blackbox

```diff
diff --git a/src/components/Editor/Tabs.css b/src/components/Editor/Tabs.css
index 4d85385..4738b13 100644
--- a/src/components/Editor/Tabs.css
+++ b/src/components/Editor/Tabs.css
@@ -71,7 +71,15 @@
   line-height: 0;
 }

-.source-tab .prettyPrint svg {
+.source-tab img.prettyPrint {
+  mask: url(/images/prettyPrint.svg);
+}
+.source-tab img.blackBox {
+  mask: url(/images/prettyPrint.svg);
+}
+
+.source-tab img.prettyPrint,
+.source-tab img.blackBox {
   height: 12px;
   width: 12px;
 }
@@ -86,11 +94,6 @@
   align-self: center;
 }

-.source-tab .blackBox svg {
-  height: 12px;
-  width: 12px;
-}
-
 .source-tab .blackBox path {
   fill: var(--theme-textbox-box-shadow);
 }
 
 diff --git a/src/components/Editor/Tabs.js b/src/components/Editor/Tabs.js
index d78d4e7..026fcf4 100644
--- a/src/components/Editor/Tabs.js
+++ b/src/components/Editor/Tabs.js
@@ -482,10 +482,10 @@ class SourceTabs extends PureComponent {
     const sourceObj = source.toJS();

     if (isPretty(sourceObj)) {
-      return <Svg name="prettyPrint" />;
+      return <img className="prettyPrint" />;
     }
     if (sourceObj.isBlackBoxed) {
-      return <Svg name="blackBox" />;
+      return <img className="blackBox" />;
     }
   }
```

jasonLaster:
### Gist for close button

```diff 
diff --git a/src/components/shared/Button/Close.css b/src/components/shared/Button/Close.css
[...]

PR #4351, opened 2017-10-11
Update sources arrow

jasonLaster:
Associated Issue: #4350

### Summary of Changes

Switch to the tree arrow to a CSS SVG

screen shot 2017-10-11 at 11 01 47 am


jasonLaster:
some small tweaks, the before/after should be the same now

#### before




#### after

PR #4354, opened 2017-10-11
Tidies up reducer comments

lukaszsobek:
Associated Issue: #4338

### Summary of Changes

* Unified comments across header of reducer files
* added license information


jasonLaster:
looks like you angered the Flow gods when you added a flow file. How about removing that and then making a separate issue for it

Issue #4345, opened 2017-10-11
[L10N] Hard-coded accesskeys


StoyanDimitrov:
This is the list of the hard-coded accesskeys I see in the Firefox UI.

[EditorMenu.js#L82](https://github.com/devtools-html/debugger.html/blob/cfe847a2b5dc14ebc1bae54c154e5a9320fc274d/src/components/Editor/EditorMenu.js#L82)

[EditorMenu.js#L89](https://github.com/devtools-html/debugger.html/blob/cfe847a2b5dc14ebc1bae54c154e5a9320fc274d/src/components/Editor/EditorMenu.js#L89)

[GutterMenu.js#L68](https://github.com/devtools-html/debugger.html/blob/cfe847a2b5dc14ebc1bae54c154e5a9320fc274d/src/components/Editor/GutterMenu.js#L68)

[GutterMenu.js#L82](https://github.com/devtools-html/debugger.html/blob/cfe847a2b5dc14ebc1bae54c154e5a9320fc274d/src/components/Editor/GutterMenu.js#L82)

[GutterMenu.js#L106](https://github.com/devtools-html/debugger.html/blob/cfe847a2b5dc14ebc1bae54c154e5a9320fc274d/src/components/Editor/GutterMenu.js#L106)

I think they should be migrated to l10zable ones to help with translation consistency.

lukaszsobek:
Just went through the whole src folder, the ones you mentioned seem to be the only ones still hard-coded.

PR #4378, opened 2017-10-15
Adds L10N for  previously hardcoded keybindings

lukaszsobek:
Associated Issue: #4345

### Summary of Changes

* replaced hard-coded access keys with a reference to debugger.properties
* added access keys if they did not already exist

### DIFF
--- a/assets/panel/debugger.properties
+++ b/assets/panel/debugger.properties
@@ -389,7 +389,7 @@ editor.editBreakpoint=Edit breakpoint
 # LOCALIZATION NOTE (editor.addConditionalBreakpoint): Editor gutter context
 # menu item for adding/editing a breakpoint condition on a line.
 editor.addConditionalBreakpoint=Add/Edit conditional breakpoint
-editor.addConditionalBreakpoint.accesskey=C
+editor.addConditionalBreakpoint.accesskey=c
 
 # LOCALIZATION NOTE (editor.conditionalPanel.placeholder): Placeholder text for
 # input element inside ConditionalPanel component
@@ -402,7 +402,7 @@ editor.conditionalPanel.close=Cancel edit breakpoint and close
 # LOCALIZATION NOTE (editor.jumpToMappedLocation1): Context menu item
 # for navigating to a source mapped location
 editor.jumpToMappedLocation1=Jump to %S location
-editor.jumpToMappedLocation1.accesskey=C
+editor.jumpToMappedLocation1.accesskey=m
 
 # LOCALIZATION NOTE (framework.disableGrouping): This is the text that appears in the
 # context menu to disable framework grouping.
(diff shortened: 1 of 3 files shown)


assets/panel/debugger.properties
```diff
@@ -386,8 +387,9 @@ editor.removeBreakpoint=Remove breakpoint
 editor.editBreakpoint=Edit breakpoint
 
 # LOCALIZATION NOTE (editor.addConditionalBreakpoint): Editor gutter context
-# menu item for adding a breakpoint condition on a line.
-editor.addConditionalBreakpoint=Add conditional breakpoint
+# menu item for adding/editing a breakpoint condition on a line.
+editor.addConditionalBreakpoint=Add/Edit conditional breakpoint
+editor.addConditionalBreakpoint.accesskey=C
```
flodolo: Please use 'c' (lowercase), it's less confusing than relying on the [fallback](https://developer.mozilla.org/en-US/docs/Mozilla/Tech/XUL/Attribute/accesskey) C -> c.

assets/panel/debugger.properties
```diff
@@ -400,6 +402,7 @@ editor.conditionalPanel.close=Cancel edit breakpoint and close
 # LOCALIZATION NOTE (editor.jumpToMappedLocation1): Context menu item
 # for navigating to a source mapped location
 editor.jumpToMappedLocation1=Jump to %S location
+editor.jumpToMappedLocation1.accesskey=C
```
flodolo: You need to pick a letter available in the associated label, otherwise this will show up as 

> Jump to %S location (C)

With C underlined.

I know you're just de-hardcoding the values, but it should be fixed nonetheless, unless you've completely run out of characters.

assets/panel/debugger.properties
```diff
@@ -420,6 +423,7 @@ original=original
 # LOCALIZATION NOTE (expressions.placeholder): Placeholder text for expression
 # input element
 expressions.placeholder=Add watch expression
+expressions.placeholder.accesskey=E
```
flodolo: As above, please use lowercase 'e'

flodolo:
@jasonLaster Do you want to take a second look here?

codehag:
This looks good to me, thanks @lukaszsobek & @flodolo !

lukaszsobek:
Thanks!

assets/panel/debugger.properties
```diff
@@ -386,8 +387,9 @@ editor.removeBreakpoint=Remove breakpoint
 editor.editBreakpoint=Edit breakpoint
 
 # LOCALIZATION NOTE (editor.addConditionalBreakpoint): Editor gutter context
-# menu item for adding a breakpoint condition on a line.
-editor.addConditionalBreakpoint=Add conditional breakpoint
+# menu item for adding/editing a breakpoint condition on a line.
+editor.addConditionalBreakpoint=Add/Edit conditional breakpoint
```
flodolo: Good thing I review strings…

GitHub diff clearly didn't help here, but you changed a string without using a new ID :-\
https://developer.mozilla.org/en-US/docs/Mozilla/Localization/Localization_content_best_practices#Changing_existing_strings

We need a follow up to rename this string ID

flodolo: @lukaszsobek  BTW, is this change expected? Now it describes a completely different function, but there are no code changes associated in this PR.

Issue #4369, opened 2017-10-13
[Previews] some lines are incorrectly marked as out of scope


jasonLaster:
STR:

1. go to [todomvc-react](http://todomvc.com/examples/react/#/active)
2. add a bp in TodoModel.js#46
3. add a bp
4. click toggle all
5. hover on "this"

ER: see something
AR: you don't :)

NOTE a breakpoint below works 

![Screen Shot 2017-10-13 at 11.34.07 AM.png](https://shipusercontent.com/d1a9f8da1c9a7382da3c34a8ec12574c/Screen%20Shot%202017-10-13%20at%2011.34.07%20AM.png)

#### GIST 

```diff
diff --git a/src/selectors/linesInScope.js b/src/selectors/linesInScope.js
index 23b5512..7b241c5 100644
--- a/src/selectors/linesInScope.js
+++ b/src/selectors/linesInScope.js
@@ -27,6 +27,7 @@ export default function getInScopeLines(state: OuterState) {
     return;
   }
 
+  // what's going on here when calculating out of scope lines
   const linesOutOfScope = getOutOfScopeLines(
     outOfScopeLocations,
     source.toJS()
diff --git a/src/utils/editor/expression.js b/src/utils/editor/expression.js
index b36cd2c..0864e6e 100644
--- a/src/utils/editor/expression.js
+++ b/src/utils/editor/expression.js
@@ -48,6 +48,12 @@ export function updatePreview(
 
   const inScope = linesInScope && linesInScope.includes(location.line);
 
+  if (tokenText == "this") {
+    // note that the line is out of scope... why is that :)
+    console.log({ inScope, linesInScope });
+    debugger;
+  }
+
   const invaildType =
     target.className === "cm-string" ||
     target.className === "cm-number" ||
diff --git a/src/workers/parser/getOutOfScopeLocations.js b/src/workers/parser/getOutOfScopeLocations.js
index 6e38ff5..ad9a819 100644
--- a/src/workers/parser/getOutOfScopeLocations.js
+++ b/src/workers/parser/getOutOfScopeLocations.js
@@ -80,6 +80,8 @@ function getOutOfScopeLocations(
   const { functions, comments } = findSymbols(source);
   const commentLocations = comments.map(c => c.location);
 
+  // function responsible for figuring this out
+
   return functions
     .map(getLocation)
     .concat(commentLocations)
[...]

PR #4381, opened 2017-10-16
4369-[Previews] some lines are incorrectly marked as out of scope

Fischer-L:
Associated Issue: #4369 

### Screenshots/Videos 
x0lc81erpu

### DIFF
--- a/src/workers/parser/tests/contains.spec.js
+++ b/src/workers/parser/tests/contains.spec.js
@@ -1,0 +1,308 @@
+import {
+  containsPosition,
+  containsLocation,
+  nodeContainsPosition
+} from "../utils/contains";
+
+function getTestLoc() {
+  return {
+    start: {
+      line: 10,
+      column: 2
+    },
+    end: {
+      line: 12,
+      column: 10
+    }
+  };
+}
+
+function startPos(lineOffset, columnOffset) {
+  const { start } = getTestLoc();
+  return {
+    line: start.line + lineOffset,
+    column: start.column + columnOffset
+  };
+}
+
+function endPos(lineOffset, columnOffset) {
+  const { end } = getTestLoc();
+  return {
+    line: end.line + lineOffset,
+    column: end.column + columnOffset
+  };
+}
+
+function startLine(lineOffset = 0) {
+  const { start } = getTestLoc();
+  return {
+    line: start.line + lineOffset
+  };
+}
+
+function endLine(lineOffset = 0) {
+  const { end } = getTestLoc();
+  return {
+    line: end.line + lineOffset
+  };
+}
+
+function testContains(pos, bool) {
+  const loc = getTestLoc();
+  expect(containsPosition(loc, pos)).toEqual(bool);
+}
+
+function testContainsPosition(pos, bool) {
+  const loc = getTestLoc();
+  expect(nodeContainsPosition({ loc }, pos)).toEqual(bool);
+}
+
+describe("containsPosition", () => {
+  describe("location and postion both with the column criteria", () => {
+    it("should contain position within the location range", () =>
+      testContains(startPos(1, 1), true));
+
+    it("should not contain position out of the start line", () =>
+      testContains(startPos(-1, 0), false));
+
+    it("should not contain position out of the start column", () =>
+      testContains(startPos(0, -1), false));
+
+    it(`should contain position on the same start line and
+       within the start column`, () => testContains(startPos(0, 1), true));
+
+    it("should not contain position out of the end line", () =>
+      testContains(endPos(1, 0), false));
+
+    it("should not contain position out of the end column", () =>
+      testContains(endPos(0, 1), false));
+
+    // eslint-disable-next-line max-len
+    it("should contain position on the same end line and within the end column", () =>
+      testContains(endPos(0, -1), true));
+  });
+
+  describe("position without the column criterion", () => {
+    it("should contain position on the same start line", () =>
+      testContains(startLine(0), true));
+
+    it("should contain position on the same end line", () =>
+      testContains(endLine(0), true));
+  });
+
+  describe("location without the column criterion", () => {
+    it("should contain position on the same start line", () => {
+      const loc = getTestLoc();
+      loc.start.column = undefined;
+      const pos = {
+        line: loc.start.line,
+        column: 1
+      };
+      expect(containsPosition(loc, pos)).toEqual(true);
+    });
+
+    it("should contain position on the same end line", () => {
+      const loc = getTestLoc();
+      loc.end.column = undefined;
+      const pos = {
+        line: loc.end.line,
+        column: 1
+      };
+      expect(containsPosition(loc, pos)).toEqual(true);
+    });
+  });
+
+  describe("location and postion both without the column criterion", () => {
+    it("should contain position on the same start line", () => {
+      const loc = getTestLoc();
+      loc.start.column = undefined;
+      const pos = startLine();
+      expect(containsPosition(loc, pos)).toEqual(true);
+    });
+
+    it("should contain position on the same end line", () => {
+      const loc = getTestLoc();
+      loc.end.column = undefined;
+      const pos = endLine();
+      expect(containsPosition(loc, pos)).toEqual(true);
+    });
+  });
+});
+
+describe("containsLocation", () => {
+  describe("locations both with the column criteria", () => {
+    it("should contian location within the range", () => {
+      const locA = getTestLoc();
+      const locB = {
+        start: startPos(1, 1),
+        end: endPos(-1, -1)
+      };
+      expect(containsLocation(locA, locB)).toEqual(true);
+    });
+
+    it("should not contian location out of the start line", () => {
+      const locA = getTestLoc();
+      const locB = getTestLoc();
+      locB.start.line--;
+      expect(containsLocation(locA, locB)).toEqual(false);
+    });
+
+    it("should not contian location out of the start column", () => {
+      const locA = getTestLoc();
+      const locB = getTestLoc();
+      locB.start.column--;
+      expect(containsLocation(locA, locB)).toEqual(false);
+    });
+
+    it("should not contian location out of the end line", () => {
+      const locA = getTestLoc();
+      const locB = getTestLoc();
+      locB.end.line++;
+      expect(containsLocation(locA, locB)).toEqual(false);
+    });
+
+    it("should not contian location out of the end column", () => {
+      const locA = getTestLoc();
+      const locB = getTestLoc();
+      locB.end.column++;
+      expect(containsLocation(locA, locB)).toEqual(false);
+    });
+
+    it(`should contain location on the same start line and
+        within the start column`, () => {
+      const locA = getTestLoc();
+      const locB = {
+        start: startPos(0, 1),
+        end: endPos(-1, -1)
+      };
+      expect(containsLocation(locA, locB)).toEqual(true);
+    });
+
+    it(`should contain location on the same end line and
+        within the end column`, () => {
+      const locA = getTestLoc();
+      const locB = {
+        start: startPos(1, 1),
+        end: endPos(0, -1)
+      };
+      expect(containsLocation(locA, locB)).toEqual(true);
+    });
+  });
+
+  describe("location A without the column criterion", () => {
+    it("should contain location on the same start line", () => {
+      const locA = getTestLoc();
+      locA.start.column = undefined;
+      const locB = getTestLoc();
+      expect(containsLocation(locA, locB)).toEqual(true);
+    });
+
+    it("should contain location on the same end line", () => {
+      const locA = getTestLoc();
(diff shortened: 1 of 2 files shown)


src/workers/parser/utils/contains.js
```diff
@@ -4,11 +4,21 @@ import type { AstLocation, AstPosition } from "../types";
...
+    startsBefore =
+      a.start.column >= 0 && b.column >= 0 ? a.start.column <= b.column : true;
+  }
+
+  const endsAfter = a.end.line > b.line;
+  if (!endsAfter && a.end.line === b.line) {
+    // We do the same check as the startsBefore case for the endsAfter case.
+    endsAfter =
+      a.end.column >= 0 && b.column >= 0 ? a.end.column >= b.column : true;
+  }
```
Fischer-L: @jasonLaster Thanks for the gist hint.

Before proceeding to the test, would like to check with you if this solution is the right direction since this touches the util function. Below is my analysis, please have a look, thank you:

While getting out-of-scope locations, it uses the parsed source locations and the bp position to filter out those not contained within scope [1]
The bp position only has the column criterion [2] and the source locations have one location [3].
During filtering, the `containsPosition` doesn’t consider the case without the column criterion. 
As a result [4]
```js
// This evaluates to false although a.start.line === b.line === 46,
// because a.start.column <= b.column is false (b.column === undefined)
startsBefore = (a.start.line === b.line && a.start.column <= b.column); 
```
then the line 46 where the bp exists is judged as out of scope.

So the solution is to take the case without column criterion into consideration.

[1] https://github.com/devtools-html/debugger.html/blob/c5177581a8af4b797f9aa56f82214216de0176c6/src/workers/parser/getOutOfScopeLocations.js#L86

[2] Object { sourceId: "server1.conn26.child2/source32", line: 46 }

[3] {
  end: Object { line: 48, column: 3 }
  start: Object { line: 46, column: 30 }
}

[4] https://github.com/devtools-html/debugger.html/blob/c5177581a8af4b797f9aa56f82214216de0176c6/src/workers/parser/utils/contains.js#L9

jasonLaster:
Thanks @Fischer-L This looks really good! I think it will help document the fix if you could add some unit tests.

* `parser/tests/contains.spec.js`
* some tests that document the basic behavior and some others that show the edge case that you fixed

Fischer-L:
@jasonLaster The test is added. Please have a look thanks.

jasonLaster:
@Fischer-L looks like there are some small test issues

src/workers/parser/utils/contains.js
```diff
@@ -4,11 +4,21 @@ import type { AstLocation, AstPosition } from "../types";
...
-    a.start.line < b.line ||
-    (a.start.line === b.line && a.start.column <= b.column);
-  const endsAfter =
-    a.end.line > b.line || (a.end.line === b.line && a.end.column >= b.column);
+  let startsBefore = a.start.line < b.line;
+  if (!startsBefore && a.start.line === b.line) {
+    // If the start line is equal to the target line
+    // and there is no column criterion, we judge it as true.
+    // Otherwise we have to check if the start column is really before the target colimn.
+    startsBefore =
```
jasonLaster: perhaps we need two smaller functions:

```js
function startsBefore(a, b) {
  if (!(a.start.column > 0 && b.column)) {
    return true
  }
  
  return a.start.column <= b.column;
}
```

Fischer-L: Definitely, will update

src/workers/parser/utils/contains.js
```diff
@@ -3,12 +3,30 @@
...
-    a.end.line > b.line || (a.end.line === b.line && a.end.column >= b.column);
+  let startsBefore = a.start.line < b.line;
+  if (!startsBefore && a.start.line === b.line) {
+    startsBefore = startsBeforeColumn(a, b);
+  }
+
+  let endsAfter = a.end.line > b.line;
+  if (!endsAfter && a.end.line === b.line) {
+    endsAfter = endsAfterColumn(a, b);
+  }
```
Fischer-L: @jasonLaster I was thinking the following as well (but seems not really different from the current one):
```js
function startsBefore(a: AstLocation, b: AstPosition) {
  let startsBefore = a.start.line < b.line;
  if (a.start.line === b.line && a.start.column && b.column) {
    startsBefore = a.start.column <= b.column;
  }
  return startsBefore;
}

function endsAfter(a: AstLocation, b: AstPosition) {
  // The similar codes as startsBefore
}

export function containsPosition(a: AstLocation, b: AstPosition) {
  return startsBefore(a, b) && endsAfter(a, b);
}
```
Please let me know which is better you think, thanks

amitzur:
I think this is not fully accurate. In the example above line 47 is actually out of scope. With this code it is considered in scope, so when you hover over `todo` in line 47 you see `undefined` in the preview, which is not good. It's not defined since it's out of scope.

src/workers/parser/utils/contains.js
jasonLaster: I like it better :) 

```js
function startsBefore(a: AstLocation, b: AstPosition) {
  if (a.start.line === b.line && a.start.column && b.column) {
    return a.start.column <= b.column;
  }
  
  return a.start.line < b.line;
}

```

it's easier for me to read

Fischer-L:
> I think this is not fully accurate. In the example above line 47 is actually out of scope. With this code it is considered in scope, so when you hover over todo in line 47 you see undefined in the preview, which is not good. It's not defined since it's out of scope.

I would say Line 47 is in scope and `todo` is just not yet initialized. When entering `toggleAll` function, that function scope is on. For some not-yet-initialized variable, `undefined` for them should be expected.

See another case:
1. go to [todomvc-react](http://todomvc.com/examples/react/#/active)
2. add a bp in TodoModel.js#32 (the 1st line inside `app.TodoModel.prototype.addTodo `)
3. Add one todo in todomvc-react
4. Observe the moment we stop at the bp in TodoModel.js#32
We can see the variables inside that function scope are observable and no-yet-initialized variables are undefined.
oodqdou8nh

Consider another case
```js
function doA(n) {
   let result = doSomeJob({ id: n + 1  }); // Set a bp A here
   return result;
}
function doB(n) {
   let result = doSomeJob( // Set a bp B here
      { id: n + 1  }
   );
   return result;
}
```
`doA` and `doB` are totally the same just different in line breaks. When stopping at the bp A, `{ id: n + 1 }` is in scope but undefined because it is not yet initialized. What about stopping at the bp B? The situation should be the same for the bp B.

Chrome is doing the same thing and just not displaying not-yet-initialized `todo`
ylxfo9wkct

Fischer-L:
@jasonLaster Updated to use the 2nd approach. Please look thank you

Fischer-L:
@jasonLaster and @amitzur
Elaborate why go updating contains.js to fix this issue.
This is a fix without affecting the total architecture plus improving contains.js.

For the whole issue, there are 2 aspects to consider:
1. Currently, `containsPosition` may face the condition without the column criterion but it didn't handle that so the issue happened. Maybe we could fix this issue using another approach. However, `containsPosition` still should be improved in case the similar issue happens again. As a result, no matter what, we should improve `containsPosition` anyway. Go for it to fix this issue is more reasonable.

2. What if try to fix this issue from the *in-scope* angle? The thing will get complicated. The current architecture takes *out-of-scope* angle, which is to remove those not in scope given by `containsPosition`. There are some options for the *in-scope* way, for instances: 
    - Touch `getInScopeLines`[1] to consider the case with the same line but out-of-scope column. This approach is to *re-translate* the job of `getInScopeLines`, which is like saying `getInScopeLines` only cares the same line and column doesn't matter. To some degree that's acceptable(but maybe still debatable). However we now are facing a problem that we don't have the bp info.
      [1] https://github.com/devtools-html/debugger.html/blob/085c871b1dd6b84956525cca1b08ed0bac8b4ace/src/selectors/linesInScope.js#L22

    - Do *in-scope* from the ground up. OK, then we have to touch actions and reducers [2] and more  architecture, which is too risky and requires deeper and broader analysis.
       [2] https://github.com/devtools-html/debugger.html/search?utf8=%E2%9C%93&q=OUT_OF_SCOPE_LOCATIONS&type=

    - Add fix in other places, I can foresee some side-effect or impact to the architecture.

Fischer-L:
>>  I think this is not fully accurate. In the example above line 47 is actually out of scope. With this code it is considered in scope, so when you hover over todo in line 47 you see undefined in the preview, which is not good. It's not defined since it's out of scope.

> I would say Line 47 is in scope and todo is just not yet initialized. When entering toggleAll function, that function scope is on. For some not-yet-initialized variable, undefined for them should be expected.

Correct myself: Precisely speaking `todo` is not initialized and not visible from outside. But for the closure variables `Utils.extend` and `checked`  should be visible.

jasonLaster:
@Fischer-L could you fix up the tests so we can land this fix. 👍

src/actions/tests/sources.spec.js
```diff
@@ -54,7 +54,7 @@ describe("sources", () => {
     expect(source.get("id")).toEqual("foo1");
 
     const locations = getOutOfScopeLocations(getState());
-    expect(locations.length).toEqual(1);
+    expect(locations.length).toEqual(0);
```
Fischer-L: This tests loads a fake "foo1" source, which is [1] and set a bp at Line 1. Originally it judged this was out of scope but that contradicts what we wanna fix in this bug. In this bug this case in fact should be ruled as in-scope, otherwise the similar case [2] would be out of scope too, so update the expected test result.

[1] "function foo1() {\n  return 5;\n}": https://github.com/devtools-html/debugger.html/blob/9a2977bdb5bbb3a85effd63a315d716d9656fda3/src/actions/tests/helpers/threadClient.js#L71

[2]
```js
let obj = { bar: "bar", foo: function foo1() {  return 5; } };
```

src/workers/parser/utils/contains.js
```diff
@@ -3,14 +3,26 @@
...
-  const startsBefore =
-    a.start.line < b.line ||
-    (a.start.line === b.line && a.start.column <= b.column);
-  const endsAfter =
-    a.end.line > b.line || (a.end.line === b.line && a.end.column >= b.column);
+function startsBefore(a: AstLocation, b: AstPosition) {
+  let startsBefore = a.start.line < b.line;
+  if (a.start.line === b.line) {
+    startsBefore =
+      a.start.column >= 0 && b.column >= 0 ? a.start.column <= b.column : true;
```
Fischer-L: The getOutOfScopeLocations.spec.js test shows "0"[1] matters so `>= 0` is required.

[1] https://github.com/devtools-html/debugger.html/blob/c5177581a8af4b797f9aa56f82214216de0176c6/src/workers/parser/tests/getOutOfScopeLocations.spec.js#L29

jasonLaster:
@Fischer-L still seeing some errors

Fischer-L:
> @Fischer-L still seeing some errors

Sorry. This was embarrassing, didn't notice some eslint error out there... Fixed

jasonLaster:
Hey @Fischer-L I pushed some _opinionated_ test refactorings. The goal was to make them more readable from the outside, ideally show all the permutations in code as well as the overview.

Let me know what you think. By the way, the tests really are great :) thanks for the coverage.

### Tests 💃 

screen shot 2017-10-24 at 11 44 15 pm

Fischer-L:
@jasonLaster Thank you for the update. Looks nice.

src/workers/parser/tests/contains.spec.js
```diff
@@ -0,0 +1,307 @@
...
+}
+
+function testContainsPosition(pos, bool) {
+  const loc = getTestLoc();
+  expect(nodeContainsPosition({ loc }, pos)).toEqual(bool);
+}
+
+describe("containsPosition", () => {
+  describe("location and postion both with the column criteria", () => {
+    it("should contain position within the location range", () =>
```
Fischer-L: Was prettier moving the test description up again? I didn't touch the line break...

src/actions/tests/sources.spec.js
Fischer-L: No longer have to update sources.spec.js test because it has been modified by PR #4143 to include the column criterion.

src/workers/parser/tests/contains.spec.js
```diff
@@ -0,0 +1,307 @@
...
+    it(`should contain position on the same start line and
+       within the start column`, () => testContains(startPos(0, 1), true));
+
+    it("should not contain position out of the end line", () =>
+      testContains(endPos(1, 0), false));
+
+    it("should not contain position out of the end column", () =>
+      testContains(endPos(0, 1), false));
+
+    it("should contain position on the same end line and within the end column", () =>
```
Fischer-L: @jasonLaster the eslint complains about a over-80-length line but looks like it was prettier which formatted this line...

jasonLaster:
hmm, yeah... i'll see what we can do about that :)

PR #4386, opened 2017-10-16
Converted blackBox/prettyPrint SVG's to CSS

jselmani:
Associated Issue: #4350 

Converted `blackBox` and `prettyPrint` SVG's to images utilizing CSS styles and `img` tags. 
 
### Screenshots
convertedtabsandfooter


src/components/Editor/Footer.css
```diff
@@ -53,9 +53,20 @@
   fill: var(--theme-selection-color);
 }
 
-.source-footer > .commands > .action svg {
+.source-footer > .commands > .action > img.prettyPrint {
+  mask: url("images/prettyPrint.svg") no-repeat;
+  mask-size: 100%;
   height: 16px;
   width: 16px;
+  background: black;
```
jasonLaster: hmm, we should find the variable color that other buttons use. the variable helps us with dark theme :)

src/components/Editor/Tabs.css
```diff
@@ -67,13 +67,13 @@
...
 }
 
-.source-tab .prettyPrint {
-  line-height: 0;
-}
-
-.source-tab .prettyPrint svg {
+.source-tab img.prettyPrint {
+  mask: url("/images/prettyPrint.svg") no-repeat;
+  mask-size: 100%;
```
jasonLaster: do we need `mask-size` now?

src/components/Editor/Tabs.css
```diff
@@ -67,13 +67,13 @@
   fill: var(--theme-body-color);
 }
 
-.source-tab .prettyPrint {
-  line-height: 0;
-}
-
-.source-tab .prettyPrint svg {
+.source-tab img.prettyPrint {
+  mask: url("/images/prettyPrint.svg") no-repeat;
```
jasonLaster: lets always use a `/` i.e. `mask: url("/images/prettyPrint.svg") no-repeat;` 

so lets change the other sites

src/components/Editor/Tabs.css
jselmani: For reference because we discussed this through Slack, we need `mask-size` in `Tabs.css` because otherwise the SVG mask will be the size of the SVG itself, so `16x x 16x`.  By using `mask-size`, we limit the size to the `height` and `width` provided.

jselmani:
Do not merge.  This PR works great in Firefox, however an error occurs in Chrome when clicking on the Blackbox button.  Here is the error:

src/components/Editor/Tabs.css
```diff
@@ -67,13 +67,13 @@
...
 }
 
-.source-tab .prettyPrint {
-  line-height: 0;
-}
-
-.source-tab .prettyPrint svg {
+.source-tab img.prettyPrint {
+  mask: url(/images/prettyPrint.svg) no-repeat;
+  mask-size: 100%;
```
jasonLaster: do we need this?

jasonLaster:
dude - running the launchpad in chrome is fine

jselmani:
Then there must be something wrong with my machine for causing errors because that error showed up.  I am glad that it worked the way it is.

PR #4391, opened 2017-10-17
Use a monospace font for editor line numbers

darkwing:
IMO a monospaced font for the editor looks better:

Before:

linenumbers

After:
monospaced

### DIFF
--- a/src/components/Editor/Editor.css
+++ b/src/components/Editor/Editor.css
@@ -120,8 +120,9 @@ html[dir="rtl"] .editor-mount {
 }
 
 .CodeMirror-linenumber {
-  font-size: 11px;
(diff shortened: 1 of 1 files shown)


darkwing:
What do we think about monospacing the code too?

monospacecode

jasonLaster:
haha - it's crazy how much better this feels

jasonLaster:
I like monospacing the code as well... i wonder if we can pull in a nice editor font like menlo or something...

darkwing:
@jasonLaster I think we need to back this out since Nightly is showing monospaced font; I think it's just a local dev issue.

jasonLaster:
okay... mind making a revert PR

PR #4395, opened 2017-10-17
Move the Editor font to Monospace

darkwing:
Monospace code is an industry standard for text editors and code displays.  Let's do it!

monocode


darkwing:
Ooops, wrong image.  Anyways, updated.

@jasonLaster mentioned using another font. We'll use this PR as a starting point.

PR #4398, opened 2017-10-17
🔨: Rework `SourceSearch` into `SourcesModal`

wldcordeiro:

### Summary of Changes

* Refactor `SourceSearch` into `FilesModal`

### Test Plan

Example test plan:

- [x] Command-P opens the modal
- [x] Enter a query to see results
- [x] up down to traverse results
- [x] enter on result to select it

Light Theme
Dark Theme


jasonLaster:
I like the hammer :)

Nice work!

wldcordeiro:
Odd, traversing results doesn't seem to work because the sources tree is stealing focus. 🤔

src/components/shared/Autocomplete.js
```diff
@@ -1,158 +0,0 @@
...
-        {children}
-        {this.renderResults(searchResults)}
-      </div>
-    );
-  }
-}
-
-Autocomplete.defaultProps = {
-  size: ""
-};
```
jasonLaster: :bomb:

src/components/App.js
```diff
@@ -110,6 +114,12 @@ class App extends Component {
       this.toggleSymbolModal
     );
 
+    const searchKeys = [
+      L10N.getStr("sources.search.key2"),
+      L10N.getStr("sources.search.alt.key")
+    ];
+    searchKeys.forEach(key => shortcuts.on(key, this.toggleSourceSearch));
```
jasonLaster: maybe `this.toggleSourceSearch` should be `this.fileSourceSearch`

jasonLaster:
I think we should call this the SourceModal to differentiate between source search and file search :)

wldcordeiro:
> I think we should call this the SourceModal to differentiate between source search and file search :)

I can do that. Figured the name could be temporary till I merge them into a `GoToAnythingModal`

jasonLaster:
seeing some empty files for todomvc...

jasonLaster:
we should do overflow-y auto

wldcordeiro:
> we should do overflow-y auto

@jasonLaster I think we get that for free? Looks like the result list has that applied.

wldcordeiro:
I dug into the empty items issue and that is present in master as well. The only reason we see it here is because I'm showing all the items when our query is `""`.

wldcordeiro:
I tried to just filter out the ones that were empty but this causes our mochitest to fail as the sources used there lack URLs and therefore get empty titles/values. I want to open a follow-up issue for fixing the empty names @jasonLaster what do you think?

src/components/shared/Modal.css
```diff
@@ -20,9 +20,23 @@
...
+
+.modal .result-list {
+  height: calc(100% - 28px);
+  overflow-y: auto;
+}
+
 .modal.entering,
 .modal.exited {
-  transform: translateY(-250px);
+  transform: translateY(-101%);
```
wldcordeiro: This makes sure we put it far enough "up" that even a taller modal slides down. The extra 1% is to account for the drop-shadow that would peak over briefly.

src/test/mochitest/browser_dbg-search-sources.js
```diff
@@ -13,7 +13,6 @@ add_task(async function() {
 
   pressKey(dbg, "sourceSearch");
   await waitForElement(dbg, "input");
-  findElementWithSelector(dbg, "input").focus();
```
wldcordeiro: We don't need to manually focus our elements they do that themselves. Leaving these in caused failures.

src/components/SourcesModal.js
```diff
@@ -0,0 +1,208 @@
...
+    .filter(source => !isPretty(source) && !isThirdParty(source))
+    .map(source => ({
+      value: getSourcePath(source),
+      title: getSourcePath(source)
+        .split("/")
+        .pop(),
+      subtitle: endTruncateStr(getSourcePath(source), 100),
+      id: source.id
+    }))
+    .filter(formattedSource => formattedSource.value != "");
```
wldcordeiro: Some sources have no value or title and I filter them out but I think we should do something better. This would make a good follow-up issue.

src/test/mochitest/browser_dbg-search-sources.js
jasonLaster: nice

src/components/SourcesModal.js
```diff
@@ -0,0 +1,208 @@
...
+
+    this.setState({ selectedIndex: nextIndex });
+  };
+
+  onChange = (e: SyntheticInputEvent) => {
+    this.props.setSourceSearchQuery(e.target.value);
+    this.searchSources(e.target.value);
+  };
+
+  onKeyDown = (e: SyntheticKeyboardEvent) => {
```
wldcordeiro: By using `onKeyDown` instead of `onKeyUp` we can let the user hold down a key to continue scrolling through a list.

jasonLaster:
noticing a couple things:

1. the modal has an overflow set to scroll
2. the animation on key down is a bit funky... maybe we debounce it?

http://recordit.co/poNIAcEL1Q

wldcordeiro:
Fixes made. I think the scrolling is a little nicer now.

PR #4405, opened 2017-10-17
Adds flow types to the Scope declaration.

yurydelendik:
Associated Issue: FIXME comments at debugger-html.js

### DIFF
--- a/src/utils/scopes.js
+++ b/src/utils/scopes.js
@@ -3,25 +3,36 @@
 import { toPairs } from "lodash";
 import { get } from "lodash";
 import { simplifyDisplayName } from "./frame";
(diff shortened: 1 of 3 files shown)


src/utils/scopes.js
```diff
@@ -3,25 +3,36 @@
 import { toPairs } from "lodash";
 import { get } from "lodash";
 import { simplifyDisplayName } from "./frame";
-import type { Frame, Pause, Scope } from "debugger-html";
+import type { Frame, Pause, Scope, BindingContents } from "debugger-html";
 
-type ScopeData = {
+export type NamedValue = {
```
jasonLaster: can we make this more explicit

Issue #3950, opened 2017-09-08
[Shortcut] cmd+shift+o is used twice


jasonLaster:
The debugger overrides the default options for `cmd+shift+o` which opens the settings pane. We should choose one action and change the other, basically it should either open function search or settings not either.

jasonLaster:
@lukaszsobek I'm not sure what the fix is for this :)

lukaszsobek:
@jasonLaster for what issue? So far the bot seems to be working as intended :P

jasonLaster:
haha - i meant the issue :P

jasonLaster:
but yeah... you're the assignee! sweet

lukaszsobek:
FOBG are taken, O needs to be removed so maybe S for settings?

PR #4409, opened 2017-10-17
Changes keyboard shortcut letter from O to S from to resolve conflict

lukaszsobek:
Associated Issue: #3950

### Summary of Changes

* changes shortcut for search from CmdCtrl + Shift + O to CmdCtrl + Shift + S


jasonLaster:
I like cmd+shift+o because it's consistent w/ chrome and editors. let me see if we can change settings

jasonLaster:
closing because it looks like we'll remove the other use case

lukaszsobek:
Nice!

Issue #4109, opened 2017-09-21
[Stepping] Improve Stepping UX


jasonLaster:
We've discussed several ideas for stepping in the past:

1. only show stepping buttons and paused panels when we're paused & when we're stepping.
2. move pausing buttons (Pause on Exception and Next Statement) to the breakpoints panel

There are a couple of advantages:

1. the active state is less noisy
2. the pause state is more obvious
3. we have more space for additional pausing actions. Think event listeners, dom mutations, xhr, watch points ...
4. we can finally solve the FLASHING pane problem when stepping

### Horizontal Mode

![](http://g.recordit.co/DkhpaTpRNo.gif)

### Vertical Mode

![](http://g.recordit.co/sixI8Wnl1c.gif)

### Pause Actions

When the user selects "pause on next statement" or "pause on exception", we can show it in the breakpoint pane. 

<img width="696" alt="screen shot 2017-09-21 at 12 06 01 am" src="https://user-images.githubusercontent.com/254562/30678516-b9fbdb2c-9e60-11e7-8aa8-083842c29ff7.png">

### actions in the breakpoints panel

* the dropdown is one possibility. VS code shows these actions in the [panel][pa]. 
[pa]: https://cloud.githubusercontent.com/assets/323878/15724622/b4d94908-27fc-11e6-981e-2bf39907493e.png

<img width="762" alt="screen shot 2017-09-21 at 12 05 54 am" src="https://user-images.githubusercontent.com/254562/30678517-b9fc3d88-9e60-11e7-80bd-b4f526c2a0f5.png">

Here's the [branch][b] where i'm working on it. Pull it down and play with it. 

[b]:https://github.com/devtools-html/debugger.html/compare/master...jasonLaster:active-view?expand=1

digitarald:
Just to throw this in as hypothesis that we could test with some user research or an experiment: The stepping controls to the side of the code might throw some people off, as they navigate execution flow navigate the source but the controls are in the sidepanel.

VSCode's solution that overlays the stepping controls seems like a good alternative and also avoids the empty button space that needs to be kept available for the buttons.

jasonLaster:
we've discussed moving the stepping buttons before

jasonLaster:
For me, it's an incremental thing:

1. move pause buttons, which are rare
2. hide step buttons so they're only visible when paused
3. improve the paused UI holistically

jasonLaster:
[...]

PR #4410, opened 2017-10-17
Special Breakpoints Dropdown

GarethSharpe:
Associated Issue: #4109

### DIFF
--- a/src/actions/pause.js
+++ b/src/actions/pause.js
@@ -120,6 +120,7 @@ export function pauseOnExceptions(
   shouldIgnoreCaughtExceptions: boolean
 ) {
   return ({ dispatch, client }: ThunkArgs) => {
+    console.log("PauseOnExceptions()")
     dispatch({
       type: "PAUSE_ON_EXCEPTIONS",
       shouldPauseOnExceptions,
@@ -220,6 +221,7 @@ export function resume() {
 export function breakOnNext() {
   return ({ dispatch, client }: ThunkArgs) => {
     client.breakOnNext();
+    console.log("breakOnNext()")
 
     return dispatch({
       type: "BREAK_ON_NEXT",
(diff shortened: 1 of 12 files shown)


src/actions/pause.js
```diff
@@ -141,12 +142,15 @@ export function pauseOnExceptions(
  */
 export function command(type: CommandType) {
   return async ({ dispatch, client }: ThunkArgs) => {
-    // execute debugger thread command e.g. stepIn, stepOver
-    dispatch({ type: "COMMAND", value: { type } });
-
-    await client[type]();
-
-    dispatch({ type: "CLEAR_COMMAND" });
+    dispatch({
```
jasonLaster: you don't need to do this refactor. That was me messing around

src/reducers/pause.js
```diff
@@ -141,10 +141,10 @@ function update(state: PauseState = State(), action: Action): PauseState {
       });
 
     case "COMMAND":
-      return { ...state, command: action.value.type };
-
-    case "CLEAR_COMMAND":
-      return { ...state, command: "" };
+      return {
```
jasonLaster: you don't need this

jasonLaster:
Hey! I just took a look and it seems like you pulled in my previous PR pretty literally...

So, this is a tempting thing to do... but I actually think it gets you in more trouble than it's worth:

1. the changes are out of date
2. the PR has tons of stuff that's not really needed
3. the PR is big and hard to understand... really hard
4. it's not clear what to do (obviously)

Here's what I recommend.

1. start on a new up to date branch
2. add a dropdown to the the breakpoints header (look at the PR to see how i did it). Maybe make a small PR for customizing the dropdown button :) Maybe make a small PR that adds this dropdown behind a feature [flag][f]
3. add one action to the dropdown (perhaps pause on exceptions)... make sure that works. Add all the other actions...
4. start removing the buttons from the command bar
5. profit

GarethSharpe:
Hey man - Ha yes I did. I was under the (false it seems) impression that
everything that you did in that PR was required for the functionality so I
just started trying to understand how it all worked. Thank you very much
for your recommendations though. I really like that approach and I think it
will give you a little more to go off of if I ever message you again and
say "I'm stuck". Thanks again for putting in the time - I'll start this
Saturday morning. #profit

PR #4425, opened 2017-10-18
Fix "3" expressions

jasonLaster:
### Summary of Changes

@bomsy and i found that "3" watch expressions were showing a syntax error. The problem was how we escaped the watch expression: 

this is what we want:

```
"try {
 \\"2\\"
} catch (e) {
 e
}"
```

### Test Plan

unit tests

### before

screen shot 2017-10-18 at 10 43 28 am

### after

screen shot 2017-10-18 at 10 38 50 am


bomsy:
LGTM!

Issue #3822, opened 2017-08-30
Memory Leak in Preview Component


jvilk:
The `Preview` component adds four event listeners to CodeMirror that remain after it unmounts.

```typescript
class Preview extends PureComponent {
  // (snip)
  componentDidMount() {
    const { codeMirror } = this.props.editor;
    const codeMirrorWrapper = codeMirror.getWrapperElement();

    codeMirror.on("scroll", this.onScroll);
    codeMirrorWrapper.addEventListener("mouseover", this._onMouseOver);
    codeMirrorWrapper.addEventListener("mouseup", this._onMouseUp);
    codeMirrorWrapper.addEventListener("mousedown", this._onMouseDown);
  }
 // (snip)
}
```

This leak adds around 500KB to `debugger.html`'s heap size every time a source document is opened and closed.

Forgive me for using a competing browser for capturing the heap; it's the setup I'm currently working with!

### Experiment Setup

* Fresh installation from master (last commit: 91f5c632380678ce9eb7ccdb0965fb4e87f58164).
* Running via `yarn start`
* Connected to an instance of Firefox with one tab pointing to the [SensorWeb debugging example](http://aws-sensorweb-static-site.s3-website-us-west-2.amazonaws.com/).

### Methodology

Perform the following tasks with an unmodified version of debugger.html, and a version of debugger.html with the leaks fixed:

* Navigate to `http://localhost:8000/?firefox-tab=child1/tab1`
* Do the following 5 times within one session:
  * Click on `main.js` to open it in the source viewer.
  * Once loaded, click on the `x` to close `main.js`.
  * [Take a heap snapshot.](https://developer.chrome.com/devtools/docs/heap-profiling#basics_snapshot)
* For each snapshot:
  * Click on the dropdown and select the *statistics* view.
  * Subtract *System Objects* from the *Total*; this is debugger.html's controllable heap size.
    * (As far as I can tell, System Objects correspond to internal Chrome state with no clear direct link to your application's behavior. They can add considerable noise to the graphs.)

### Result

As you can see in the image below, debugger.html adds about 500KB to its heap every time `main.js` is opened and closed. After this leak is fixed, `debugger.html` grows normally (the remaining heap growth, which stabilizes after a few iterations, is likely Chrome's JIT compiler in action).

![firefox_devtools](https://user-images.githubusercontent.com/1141042/29889008-92a2258a-8d77-11e7-8c30-ef447b8e589a.png)

### Suggested Actions

[...]

PR #4426, opened 2017-10-18
removes declared hooks on componentWillUnmount

lukaszsobek:
Associated Issue: #3822

### Summary of Changes

* Removal of listeners on componentWillUnmount

### Test Plan

Tested the memory footprint by repeatedly opening and closing a tab 

Shows memory consumption before the change, and with 2 variations of changes implemented (the fix contains variation 2). 

http://prntscr.com/gyziln

### DIFF
--- a/src/components/Editor/Preview/index.js
+++ b/src/components/Editor/Preview/index.js
@@ -39,9 +39,9 @@ class Preview extends PureComponent {
     const self = this;
     self.onScroll = this.onScroll.bind(this);
     self.onMouseOver = debounce(this.onMouseOver, 40);
-    self._onMouseOver = e => this.onMouseOver(e);
-    self._onMouseUp = e => this.onMouseUp(e);
-    self._onMouseDown = e => this.onMouseDown(e);
+    self.onMouseOver = this.onMouseOver.bind(this);
+    self.onMouseUp = this.onMouseUp.bind(this);
+    self.onMouseDown = this.onMouseDown.bind(this);
   }
 
   componentDidMount() {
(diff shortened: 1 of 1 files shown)


src/components/Editor/Preview/index.js
```diff
@@ -39,16 +39,29 @@ class Preview extends PureComponent {
     const self = this;
     self.onScroll = this.onScroll.bind(this);
     self.onMouseOver = debounce(this.onMouseOver, 40);
+    self._onMouseOver = e => this.onMouseOver(e);
+    self._onMouseUp = e => this.onMouseUp(e);
+    self._onMouseDown = e => this.onMouseDown(e);
```
jasonLaster: can we bind here: `self.onMouseDown = this.onMouseDown.bind(this)`

jasonLaster:
What does this show us:



what is the current result? is there a leak, if so ... how much?

src/components/Editor/Preview/index.js
lukaszsobek: Sure. Will update.

jasonLaster:
@lukaszsobek thanks for the detailed write up. 

Let's dig a bit:

1. Source documents has a map of code mirror docs. Do we add to the map?
2. Redux has a store of sources. Does the redux sources store change?

lukaszsobek:
@jasonLaster Here is another screenshot set comparing the memory - this time with snapshorts of editor window closed as well. All images are about the same debugger session on the same window with just opening and closing the main.js editor panel (notice how the memory use increases on strings and when the window gets closed http://prntscr.com/h0rx6w

wldcordeiro:
Looks like this is ready to roll? From my read-through @lukaszsobek has responded to review and this looks like it gives us an improvement.

Issue #3846, opened 2017-09-01
Source Tabs and Gutter Column Don't Line Up


darkwing:
Hard to see in the light theme but pronounced in the dark theme:

<img width="127" alt="lineuplight" src="https://user-images.githubusercontent.com/46655/29952015-2d9f42f6-8e8c-11e7-9f10-a225e10ce68a.png">

<img width="264" alt="lineupdark" src="https://user-images.githubusercontent.com/46655/29952018-3047cc6c-8e8c-11e7-8818-4e3743148632.png">

jasonLaster:
This fix can be made in here: https://github.com/rohanprasad/debugger.html/blob/search-store-mapping/src/components/Editor/Tabs.css#L1

Ronsho:
will take it!

diessica:
Is anyone on it? I'd like to take this.

bomsy:
@Ronsho are you still on this?

jasonLaster:
@diessica it's yours

pradeepgangwar:
@jasonLaster Can i try resolving this issue, if @diessica is busy with something else.

jasonLaster:
@pradeepgangwar go for it

jasonLaster:
here are some nottes

https://github.com/devtools-html/debugger.html/blob/master/src/utils/editor/source-documents.js#L41

it could literally do document.querySelector(‘tab-foo’).style.marginLeft = whatever the gutter is…


we could also update this:

https://github.com/devtools-html/devtools-core/blob/master/packages/devtools-source-editor/src/utils.js#L47-L50

pradeepgangwar:
@jasonLaster If I understand bug correctly, then it is talking of difference between line column and tab. 
But I ran the debugger today and found that it was not like that. I think either this bug has been resolved or i have misunderstood it.

![screenshot-2017-10-6 debugger](https://user-images.githubusercontent.com/21259802/31294049-17c1f2b2-aaf7-11e7-9d7d-ed956b5f57c0.png)

jasonLaster:
try looking at files w/ more than 100 lines of code

pradeepgangwar:
Yes, I got it there is some issue with it.

PR #4427, opened 2017-10-18
Fixes 3846 Source Tabs and Gutter Column Don't Line Up

jainsneha23:
Associated Issue: #3846 

### Summary of Changes

* Pane toggle button padding changed to make it wider
* Margin left removed from first tab

### Test Plan

- [x] Open the debugger and check the view

screen shot 2017-10-18 at 11 12 15 pm

### DIFF
--- a/src/utils/editor/source-documents.js
+++ b/src/utils/editor/source-documents.js
@@ -110,12 +110,18 @@ function showSourceText(editor: Object, source: Source) {
 function resizeBreakpointGutter(editor: Object) {
   const gutters = editor.display.gutters;
   const breakpoints = gutters.querySelector(".breakpoints");
-  breakpoints.style.width = `${getLineNumberWidth(editor)}px`;
+  if (breakpoints) {
+    breakpoints.style.width = `${getLineNumberWidth(editor)}px`;
+  }
 }
 
 function resizeToggleButton(editor: Object) {
-  const toggleButton = document.querySelector(".toggle-button-start");
-  toggleButton.style.width = `${getLineNumberWidth(editor)}px`;
+  const toggleButton = document.querySelector(
+    ".source-header .toggle-button-start"
+  );
+  if (toggleButton) {
+    toggleButton.style.width = `${getLineNumberWidth(editor)}px`;
+  }
 }
 
 export {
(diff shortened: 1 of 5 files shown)


jasonLaster:
nice... can you try and support cases where there are > 100 or > 1000 lines?

I think i have a comment in the issue for it too

jainsneha23:
Sure..

codehag:
This will probably have to be calculated based off of the codemirror width, do you need any help with that?

jasonLaster:
here's where we calculate the gutter width now:

https://github.com/devtools-html/devtools-core/blob/master/packages/devtools-source-editor/src/utils.js#L40-L52

we could use it like this 

```js
import { SourceEditorUtils  } from "devtools-source-editor";
const { resizeBreakpointGutter } = SourceEditorUtils
```

jasonLaster:
After seeing the core PR let's just stop using it :) and add getGW to the debugger utils

src/utils/editor/source-documents.js
```diff
@@ -105,6 +107,23 @@ function showSourceText(editor: Object, source: Source) {
...
+  }
+}
+
+function resizeToggleButton(editor: Object) {
+  const toggleButton = document.querySelector(
+    ".source-header .toggle-button-start"
+  );
+  if (toggleButton) {
+    toggleButton.style.width = `${getLineNumberWidth(editor)}px`;
+  }
```
jasonLaster: I think this is he Ron module for these functions. This m is or code mirror documents. Let's move them to utils/editor/index

jainsneha23: sure

jainsneha23: @jasonLaster I will move the code to debugger utils. Will you take care of cleaning up the function in core modules? I will close my PR then?

src/utils/ui.js
```diff
@@ -3,7 +3,44 @@
...
  */
-export function isVisible() {
+function isVisible() {
   const el = document.querySelector("#mount");
   return el && el.getBoundingClientRect().width;
 }
+
+/* Gets the line numbers width in the code editor
+ */
+function getLineNumberWidth(editor) {
```
jasonLaster: small nit: lets add `export` to each function.

jainsneha23:
screen shot 2017-10-22 at 11 37 38 am
screen shot 2017-10-22 at 11 37 53 am
screen shot 2017-10-22 at 11 38 06 am
screen shot 2017-10-22 at 11 37 08 am
screen shot 2017-10-22 at 11 36 56 am

codehag:
Brilliant stuff!

PR #4428, opened 2017-10-18
[Previews] Fix computation of out of scope locations

amitzur:
Associated Issue: #4369 

### Summary of Changes

Thanks @liorpr for working on this together!

The reason for the bug was that it was ignoring column data. If the token is at the same line where an out of scope range begins, but earlier, then the token was considered out of scope. We added consideration of the column position.

* Add outOfScopeLocations to props of preview
* Use column data to accurately compute out of scope

NOTE: the code is dirty, maybe a bit of refactoring is in place. We did it in #goodnessSquad and wanted to create a PR. It works, but it's complicated and can probably be simplified.

### Test Plan

`updatePreview` probably needs testing, but we didn't add it.

### DIFF
--- a/src/utils/editor/expression.js
+++ b/src/utils/editor/expression.js
@@ -18,7 +18,7 @@ export function getTokenLocation(codeMirror: any, tokenEl: HTMLElement) {
 export function updatePreview(
   target: HTMLElement,
   editor: any,
-  { linesInScope, preview, setPreview, clearPreview }: any
+  { linesInScope, preview, setPreview, clearPreview, outOfScopeLocations }: any
 ) {
   const location = getTokenLocation(editor.codeMirror, target);
   const tokenText = target.innerText ? target.innerText.trim() : "";
@@ -46,7 +46,25 @@ export function updatePreview(
 
   const isUpdating = preview && preview.updating;
 
-  const inScope = linesInScope && linesInScope.includes(location.line);
+  let inScope;
+  const isLineInScope = linesInScope && linesInScope.includes(location.line);
+  if (isLineInScope) {
+    inScope = true;
+  } else {
+    const startLineLocation = outOfScopeLocations.find(
+      loc => loc.start.line === location.line
+    );
+    if (startLineLocation) {
+      inScope = startLineLocation.start.column > location.column;
+    } else if (!inScope) {
+      const endLineLocation = outOfScopeLocations.find(
+        loc => loc.end.line === location.line
+      );
+      if (endLineLocation) {
+        inScope = endLineLocation.end.column < location.column;
+      }
+    }
+  }
 
   const invaildType =
     target.className === "cm-string" ||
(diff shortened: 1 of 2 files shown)


jasonLaster:
Hey @amitzur + @liorpr! 

I think this was worked on here too: #4381. Could you see how the implementations are similar / different and help land one of them. The other is missing unit tests. I think this one is too :)

src/utils/editor/expression.js
```diff
@@ -46,7 +46,25 @@ export function updatePreview(
 
   const isUpdating = preview && preview.updating;
 
-  const inScope = linesInScope && linesInScope.includes(location.line);
+  let inScope;
+  const isLineInScope = linesInScope && linesInScope.includes(location.line);
+  if (isLineInScope) {
```
codehag: lets move this into its own function which returns a value that can be assigned to `inScope`

here is how it would looks like:
```javascript
function isInScope(linesInScope, location) {
  const isLineInScope = linesInScope && linesInScope.includes(location.line);
  if (isLineInScope) {
    return true;
  }

  const startLineLocation = outOfScopeLocations.find(
    loc => loc.start.line === location.line
  );
  if (startLineLocation) {
    return startLineLocation.start.column > location.column;
  }
  
  const endLineLocation = outOfScopeLocations.find(
    loc => loc.end.line === location.line
  );
  if (endLineLocation) {
    return endLineLocation.end.column < location.column;
  }    

  // if none of these conditions are true, return false
  return false;
}

// use
const inScope = isInScope(linesInScope, location)
```

src/utils/editor/expression.js
```diff
@@ -46,7 +46,25 @@ export function updatePreview(
...
+    inScope = true;
+  } else {
+    const startLineLocation = outOfScopeLocations.find(
+      loc => loc.start.line === location.line
+    );
+    if (startLineLocation) {
```
codehag: we seem to be using the out of scope locations to determine if the lien is in scope, but we have a linesInScope array -- why was this not working as intended? perhaps the bug is steming from linesInScope not representing the correct information?

src/utils/editor/expression.js
amitzur: love it

src/utils/editor/expression.js
amitzur: yes, actually `linesInScope` is not really mandatory here, since we need the column information.
you could write a function which works only with `outOfScopeLocations` and calculates if a location is out of scope.
I kept `linesInScope` since it was easier to calculate only in for cases where the line isn't in that array.

hope that makes sense o_O

Fischer-L:
I was working on the issue #4369 from 5 days ago but forgot to leave the message (sorry for causing the double effort). I will suggest another approach for the issue #4369 . Please see the detailed explanation: https://github.com/devtools-html/debugger.html/pull/4381#issuecomment-338382838

jasonLaster:
I think @Fischer-L's solution w/ containsLocation is sufficient here. Happy to re-open if there is still ongoing work.

PR #4430, opened 2017-10-18
Support framework frames for Marko

jainsneha23:
Associated Issue: #2980 

### Summary of Changes

* Support framework frames for Marko

### Test Plan

Example test plan:

- [x] Run the debugger locally
- [x] Debug Marko Sample [https://github.com/marko-js-samples/marko-webpack](https://github.com/marko-js-samples/marko-webpack)

### Screenshots/Videos
screen shot 2017-10-19 at 12 06 18 am
screen shot 2017-10-19 at 12 06 58 am

### DIFF
--- a/src/utils/frame.js
+++ b/src/utils/frame.js
@@ -23,6 +23,10 @@ const libraryMap = [
     pattern: /jquery/i
   },
   {
+    label: "Preact",
+    pattern: /preact/i
+  },
+  {
     label: "React",
     pattern: /react/i
   },
@@ -85,10 +89,6 @@ const libraryMap = [
   {
     label: "Dojo",
     pattern: /dojo/i
-  },
-  {
-    label: "Preact",
-    pattern: /preact/i
   },
   {
     label: "Marko",
--- a/src/utils/frame.js
+++ b/src/utils/frame.js
@@ -91,6 +91,10 @@ function isPreact(frame) {
   return getFrameUrl(frame).match(/preact/i);
 }
 
+function isMarko(frame) {
+  return getFrameUrl(frame).match(/marko/i);
+}
+
 export function getLibraryFromUrl(frame: Frame) {
   // @TODO each of these fns calls getFrameUrl, just call it once
   // (assuming there's not more complex logic to identify a lib)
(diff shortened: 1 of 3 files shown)


src/utils/frame.js
```diff
@@ -14,163 +14,113 @@ function getFrameUrl(frame) {
 }
 
 function isBackbone(frame) {
-  return getFrameUrl(frame).match(/backbone/i);
+  return getFrameUrl(frame).match(/backbone/i) && "Backbone";
```
jasonLaster: do we need these: ` && "Backbone";`?

src/utils/frame.js
```diff
@@ -14,163 +14,113 @@ function getFrameUrl(frame) {
...
+    isLodash(frame) ||
+    isEmber(frame) ||
+    isVueJS(frame) ||
+    isRxJs(frame) ||
+    isAngular(frame) ||
+    isRedux(frame) ||
+    isDojo(frame) ||
+    isImmutable(frame) ||
+    isMarko(frame)
+  );
```
jasonLaster: thanks... but i'm not sure if we want to clean up the code yet. I'm not sure what API i want yet :)

jainsneha23: The function had 20 if conditions, giving me linting error of function complexity 20. So, I was forced to do the cleanup. I can revert back. But then I have to disable this particular linting error. Do you have any other suggestion?

wldcordeiro: I think a solution that would be nicer might look something like this. I wouldn't say it needs to be done here in this PR though.  

```js
const libraryMap = {
  react: {
    label: "React",
    pattern: /react/i,
  },
  ...
};

export function getLibraryFromUrl(frame: Frame) {
  const frameUrl = getFrameUrl(frame);
  return Object
    .values(libraryMap)
    .find(({ pattern }) => frameUrl.match(pattern))
    .label; 
}
```

jasonLaster: haha - okay :)

i think @wldcordeiro has a nice refactor... lets do it 😛

jainsneha23: Great. Even I think a map is better. I will refactor this.

jainsneha23: @jasonLaster Refactoring done.

Issue #3683, opened 2017-08-16
[flow] update to v 56


jasonLaster:
It would be great to get help upgrading flow from 52 to 53.

Flow 53 is blocked on some new flow rules: https://circleci.com/gh/devtools-html/debugger.html/6192?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

arthur801031:
I'll take a stab at this :)

jasonLaster:
Oh thanks @arthur801031!

jasonLaster:
This PR is a good start, but it's possible things have changed in 54 and 55.

[flow-upgrade][fu] could help 

[fu]: https://github.com/facebook/flow/tree/master/packages/flow-upgrade

ahuth:
I've started on this as well. Will take a shot and see how it goes. Got about half of the flow errors resolved.

jasonLaster:
@ahuth that's great!

jasonLaster:
Open a WIP PR so we can provide feedback

jasonLaster:
what version of flow are you using?

ahuth:
> Open a WIP PR so we can provide feedback

Will do 👍 

> what version of flow are you using?

I'm in the middle of upgrading to 0.53.0 as a first shot, which seems to be where the significant React changes were made. Planning on going to the latest version, though (0.57.3) by the end of the effort.

jasonLaster:
sounds good. We'd obviously be happy to go to 53 before we try 54, ...

ahuth:
An interesting thing is that flow-upgrade doesn't seem to be able to actually change anything:

```
Running 3 codemods...

Processing 139 files...
Spawning 3 workers...
Sending 47 files to free worker...
Sending 47 files to free worker...
Sending 45 files to free worker...
All done.
Results:
0 errors
62 unmodified
0 skipped
0 ok
Time elapsed: 42.583seconds
✨  Done in 137.17s.
```

Doing the upgrade by hand is partially done, but was hoping that the codemods could make it a bit easier.

jasonLaster:
heh - yeah... weird

ahuth:
/claim

Opened a WIP PR

ahuth:
@jasonLaster, should I close this, since we've upgraded to Flow 53? I still want to upgrade this to 57, but that will take some more work on a couple of the dependencies.

jasonLaster:
Good call

PR #4433, opened 2017-10-19
Upgrade flow

ahuth:
Associated Issue: #3683 

### Summary of Changes

* Updated Flow to 0.53.0.
* Moved `defaultProps` for class components to `static` properties, which Flow understands.
* Updated the syntax of the flow typing. Note that I have not tried to improve on any of the type signatures at all.

In particular, instead of defining classes like

```js
class Foo extends Component {
  props: Props
  state: State
}
```

it will now be

```js
class Foo extends Component<Props, State>
```

### Test Plan

* Tests
* Manual QA


src/components/shared/Modal.js
```diff
@@ -1,23 +1,20 @@
 // @flow
 
 import PropTypes from "prop-types";
-import React, { Component } from "react";
-import type { Children } from "react";
+import * as React from "react";
```
ahuth: This is required to make use of the `React.Node` or `React.Element` types. Alternatively, we could do something like

```js
import type { Node } from "react"
```

ahuth:
Right now I'm getting flow warnings for `node_modules/devtools-reps/src/object-inspector/index.js`. We may need to update that library, as well.

src/components/shared/Autocomplete.js
```diff
@@ -0,0 +1,157 @@
...
+  items: Array<Object>,
+  close: (value: any) => void,
+  inputValue: string,
+  placeholder: string,
+  size: string,
+  onChangeHandler: (queryString: string) => void,
+  children: any
+};
+
+export default class Autocomplete extends Component<Props, State> {
```
ahuth: Uh oh, something is up with this diff. Something may have gone wrong while merging in master from upstream.

jasonLaster:
@ahuth just read through it and it looks great! Thank you so much for your hard work

jasonLaster:
@ahuth feel free to disable flow checks for it in `flow_config` I want this to land :)

it's too valuable

ahuth:
@jasonLaster, this should be good to go, although I haven't manually QA'd it, yet.

Upgraded to 0.53.0. Going farther than that will take some more work. I'll do that sometime, and may also come back and cleanup some of the types.

codehag:
this looks good

ahuth:
Thanks

src/components/shared/Modal.js
jasonLaster: I like this

jasonLaster:
Thanks for the help! You should stop by our [slack](https://devtools-html-slack.herokuapp.com) :wave:

Issue #1743, opened 2017-01-18
[debuggee] Show a pause indicator


jasonLaster:
We currently show a pause reason in the debugger [1] , but we're not showing anything in the debuggee

We should have a popup and a mask.
https://bugzilla.mozilla.org/show_bug.cgi?id=1318824

[1] Pause Reason
![screen shot 2017-01-18 at 12 27 55 pm](https://cloud.githubusercontent.com/assets/254562/22075358/c1132580-dd79-11e6-8078-a42f3522a79d.png)

clarkbw:
I believe this might require server side changes.

We currently have something in the inspector that pops up to indicate that the page is currently paused.  This only appears if you're paused and you try to use the inspector.

Here's the code:
https://dxr.mozilla.org/mozilla-central/source/devtools/client/inspector/inspector.js#238

Screenshot:
<img width="425" alt="screen shot 2017-01-20 at 3 06 57 pm" src="https://cloud.githubusercontent.com/assets/2134/22168552/388014b2-df22-11e6-9aa6-2bd43da3147b.png">

However the Chrome version: 
* Displays across the entire page no matter what tool is selected
* Offers the resume action
* Has a configurable message
* Is brought up via the client API (ours detects that the debugger is paused)

There are some positives and negatives there, but we should definitely be taking the good.

jasonLaster:
Thanks for the detailed write-up.

Let's talk to Ryan and see what he thinks the complexity is based on rdm.

clarkbw:
Also, this [envision](https://projects.invisionapp.com/share/AY8OEG3C6#/screens/117937554_Debugger) has designs for such a pause indicator. It doesn't have the reason but I think we could update it to include that.

jasonLaster:
Here's where the inspector show's the message:
https://dxr.mozilla.org/mozilla-central/source/devtools/client/inspector/inspector.js#235-258

clarkbw:
Maybe @bwinton could help us with this?

Garbee:
The overlay is primarily useful in Chrome's case since you can have an un-docked debugger and lose track of it. Which if you can't click on your page and the debugger is hidden... is an issue. It also helps with just that though, knowing you can't click/hover on things since that could cause the expectation of JS to run, which it can't. Just some background context on why it happens there.

I do love the idea of the full-size Command Bar at the top of the page though. That looks very nice in the mocks.

jryans:
[...]

PR #4438, opened 2017-10-19
WIP: add paused overlay

codehag:
show paused overlay

associated issue: #1743

I need help on the mochi tests: how do i detect if the paused overlay is being shown?

this relies on https://bugzilla.mozilla.org/show_bug.cgi?id=1354679

### DIFF
--- a/src/actions/pause.js
+++ b/src/actions/pause.js
@@ -117,6 +117,7 @@ export function paused(pauseInfo: Pause) {
         )
       : frameScopes;
 
+    client.showPausedOverlay();
     dispatch({
       type: "PAUSED",
       pauseInfo: { why, frame, frames },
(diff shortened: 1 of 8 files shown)


jasonLaster:
ooh, this is cool.

I think it might be better to add the client logic [here](https://dxr.mozilla.org/mozilla-central/source/devtools/client/framework/attach-thread.js#24) though. The rationale is this:

1. it's nice to have the logic next to where we highlight the tool 
2. we can pause before the debugger is open. It's rare, but possible... think `debugger` in the console.

codehag:
thanks for the comments jason! I also thought it would be nice to have it all in one place, but here is why i have the front end call "clientPaused" explicitly:

1) by having it in the client, it is easier to pass the localized "pause reason" message back to the server, instead of having to redo how we do localization in order to synchronize it between server and client. It would save us work later on.

2) the server pauses before the client, and there is a disconnect between the debugger ui being available for debugging, and the pause overlay being shown. I found this frustrating as a user, though it may be alleviated later on if we improve performance

Issue #4297, opened 2017-10-05
Optimize initial pause 


jasonLaster:
We're currently optimizing stepping speed in #4284, but the initial pause is a special case that should also be investigated.

My theory is that we will do additional work on the first pause like parsing large scripts.

STR:
1. go to [perf-html][p] locally
2. add a breakpoint in TabBar.js#34
3. click "Timeline" or one of the other tabs

Observe that it takes awhile to pause

[p]:http://localhost:4242/public/fa1302b5668d67f1434932edd5fa60932d9895ac/markers/?hiddenThreads=&markerSearch=&range=2.1344_3.2600&react_perf&thread=5&threadOrder=0-2-3-4-5-1
[p2]:https://github.com/devtools-html/perf.html

NOTE: [perf-html][p2] is a great project that is easy to git clone, npm install, npm start :)

jasonLaster:
i'm going to look into this and see if i can simplify

how / when we map scopes and get symbols

yurydelendik:
Reduced test case for parsing https://yurydelendik.github.io/old-man-sandbox/babel-parse/index.html

jasonLaster:
Here's the source: https://github.com/yurydelendik/old-man-sandbox/blob/master/babel-parse/package.json

PR #4439, opened 2017-10-19
(wip) Refactor pause actions

jasonLaster:
Associated Issue: #4297 

### Summary of Changes

* cleanup actions
* defer expensive work
* remove ensure text

### Test Plan

more unit tests :)


jasonLaster:
This change brings the time it takes to pause `perf.html` from 7s to 163ms. I kinda feel bad bragging about this because 7s is so large.

How did we get these savings? We deferred mapping the scopes until after we've saved the set the pause data. We're also not blocking on that data as well.

src/actions/sources/tests/createPrettySource.spec.js
```diff
@@ -2,9 +2,11 @@ import { actions, createStore, makeSource } from "../../../utils/test-head";
 import { createPrettySource } from "../createPrettySource";
 import { getSourceByURL } from "../../../selectors";
 
+import { sourceThreadClient } from "../../tests/helpers/threadClient.js";
+
 describe("createPrettySource", () => {
-  it("returns a pretty source for a minified file", async () => {
-    const { dispatch, getState } = createStore();
+  fit("returns a pretty source for a minified file", async () => {
```
codehag: forced test here

src/actions/pause/commands.js
```diff
@@ -0,0 +1,113 @@
...
+ * Debugger commands like stepOver, stepIn, stepUp
+ *
+ * @param string $0.type
+ * @memberof actions/pause
+ * @static
+ */
+export function command(type: CommandType) {
+  return async ({ dispatch, client }: ThunkArgs) => {
+    // execute debugger thread command e.g. stepIn, stepOver
+    dispatch({ type: "COMMAND", value: { type } });
```
codehag: command data doesnt seem to be consumed anywhere? this might be replaced with an api call and then we can skip the redux cycle (since no front end data needs to be updated)

codehag: is the data from COMMAND used anywhere?

src/actions/pause/paused.js
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+ * Debugger has just paused
+ *
+ * @param {object} pauseInfo
+ * @memberof actions/pause
+ * @static
+ */
+export function paused(pauseInfo: Pause) {
+  return async function({ dispatch, getState, client, sourceMaps }: ThunkArgs) {
```
codehag: do we need to use async here? or can we make this dispatch utilize thunk, and put the async (mapped frames, and frameScopes) stuff in the [PROMISE] ?

src/actions/pause/resumed.js
```diff
@@ -0,0 +1,30 @@
...
+ * @memberof actions/pause
+ * @static
+ */
+export function resumed() {
+  return ({ dispatch, client, getState }: ThunkArgs) => {
+    if (!isPaused(getState())) {
+      return;
+    }
+
+    const wasPausedInEval = pausedInEval(getState());
```
codehag: perhaps we need a comment about this

src/actions/sources/loadSourceText.js
```diff
@@ -1,8 +1,25 @@
 // @flow
 import { PROMISE } from "../../utils/redux/middleware/promise";
 import { setEmptyLines, setSymbols } from "../ast";
+import { getSource } from "../../selectors";
+import { setSource } from "../../workers/parser";
 import type { Source } from "../../types";
 import type { ThunkArgs } from "../types";
+
+async function loadSource(source: Source, { sourceMaps, client }) {
```
codehag: 👍

src/actions/tests/sources.spec.js
```diff
@@ -131,8 +133,8 @@ describe("sources", () => {
     const fooSource = getSource(getState(), "foobar.js");
     const barSource = getSource(getState(), "barfoo.js");
 
-    expect(fooSource.get("text").indexOf("return 2")).not.toBe(-1);
-    expect(barSource.get("text").indexOf("return 3")).not.toBe(-1);
+    expect(fooSource.get("text").indexOf("return foobar.js")).not.toBe(-1);
```
codehag: text is a string? if it is maybe `includes` would be clearer

src/components/SecondaryPanes/Scopes.js
```diff
@@ -41,11 +41,12 @@ class Scopes extends PureComponent<Props, State> {
   }
 
   componentWillReceiveProps(nextProps) {
-    const { pauseInfo, selectedFrame } = this.props;
+    const { pauseInfo, selectedFrame, frameScopes } = this.props;
     const pauseInfoChanged = pauseInfo !== nextProps.pauseInfo;
-    const selectedFrameChange = selectedFrame !== nextProps.selectedFrame;
+    const selectedFrameChanged = selectedFrame !== nextProps.selectedFrame;
+    const frameScopesChanged = frameScopes !== nextProps.frameScopes;
```
codehag: is this part of the pause refactor?

src/components/shared/SearchInput.js
```diff
@@ -6,7 +6,7 @@ import CloseButton from "./Button/Close";
 import "./SearchInput.css";
 
 const arrowBtn = (onClick, type, className, tooltip) => {
-  var props = {
+  const props = {
```
codehag: :+1:

src/reducers/breakpoints.js
```diff
@@ -185,24 +184,19 @@ export function getBreakpointForLine(
...
 }
 
-export const getHiddenBreakpoint = createSelector(getBreakpoints, function(
-  breakpoints
-) {
-  const hiddenBreakpoints = breakpoints
+export function getHiddenBreakpoint(state: OuterState) {
+  return getBreakpoints(state)
     .valueSeq()
     .filter(breakpoint => breakpoint.hidden)
```
codehag: maybe we could use `find`. it will return the first match, though it doesnt need to be done in this pr.. this also looks like it isnt part of the pause refactor. I bring this up mostly so that we have a clear chronology when looking back on the commit messages. don't worry about it for this one, but it might be worth splitting out these driveby refactors into their own prs

jasonLaster: yeah... i did it because `createSelector` was no longer defined... but that might be more troubling :/

src/actions/tests/sources.spec.js
jasonLaster: agreed, but lets do this later :)

PR #4445, opened 2017-10-20
fix bug where the line being 0 returns a falsy

bomsy:
Associated Issue: none

### Summary of Changes

* As a follow-up to #4143
* Fixes an issue where the `line` being `0` returns a falsy, causing the not to highlight for single line minifed files
* Use `loadedState` to check if the text has fully loaded

### Test Plan

Example test plan:

- [x] Go to the `source mapped js.min files` example
- [x] Press `cmd + shift + f`
- [x] search for `log-message`
- [x] Click the match result for `example-websites/source-mapping/js/main.min.js`
- [x] The line should flash

### DIFF
--- a/src/components/Editor/index.js
+++ b/src/components/Editor/index.js
@@ -245,7 +245,7 @@ class Editor extends PureComponent<Props, State> {
     // Only update and jump around in real source texts. This will
     // keep the jump state around until the real source text is
     // loaded.
-    if (selectedSource && selectedSource.get("loadedState") === "loaded") {
+    if (selectedSource && isLoaded(selectedSource.toJS())) {
       this.highlightLine();
     }
   }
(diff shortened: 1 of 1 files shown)


src/components/Editor/index.js
```diff
@@ -245,7 +245,7 @@ class Editor extends PureComponent<Props, State> {
     // Only update and jump around in real source texts. This will
     // keep the jump state around until the real source text is
     // loaded.
-    if (selectedSource && selectedSource.has("text")) {
+    if (selectedSource && selectedSource.get("loadedState") === "loaded") {
```
jasonLaster: lets use the `isLoaded` [helper](https://github.com/devtools-html/debugger.html/blob/master/src/utils/source.js#L245-L247)

PR #4449, opened 2017-10-20
Add Jest Cases

jasonLaster:
### Summary of Changes

Adds support for cases aka data providers, via the `jest-in-case` package. Note our tests don't change at all


bomsy:
@jasonLaster this is cool!

PR #4454, opened 2017-10-21
[Refacoring] Make Redux Connect more explicit in Sources Tree

amelzer:
Hey folks, when working on SourcesTree.js I noticed that the redux usage can be made more readable using a more explicit approach.

### Summary of Changes

* explicit import of used actions
* ditch bindActionsCreators in favour of `() => dispatch(myImportedAction())`
* explicitly name and define connect props

I think this makes it more readable and more accessible for newbies who don't have to follow a trail of imports to see where props are coming from.

### DIFF
--- a/src/components/PrimaryPanes/SourcesTree.js
+++ b/src/components/PrimaryPanes/SourcesTree.js
@@ -9,7 +9,6 @@ import React, { Component } from "react";
 import classnames from "classnames";
 
 // Redux
-import { bindActionCreators } from "redux";
 import { connect } from "react-redux";
 import {
   getShownSource,
@@ -19,7 +18,9 @@ import {
   getProjectDirectoryRoot,
   getSources
 } from "../../selectors";
-import actions from "../../actions";
+
+import { setExpandedState } from "../../actions/source-tree";
+import { selectSource } from "../../actions/sources";
 
 // Types
 import type { SourcesMap } from "../../reducers/types";
@@ -47,13 +48,13 @@ import { features } from "../../utils/prefs";
 import { setProjectDirectoryRoot } from "../../actions/ui";
 
 type Props = {
-  sources: SourcesMap,
   selectSource: string => void,
+  setExpandedState: any => void,
+  sources: SourcesMap,
   shownSource?: string,
   selectedSource?: SourceRecord,
   debuggeeUrl: string,
   projectRoot: string,
-  setExpandedState: any => void,
   expanded?: any
 };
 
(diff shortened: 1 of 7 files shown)


src/components/PrimaryPanes/SourcesTree.js
```diff
@@ -353,15 +362,21 @@ class SourcesTree extends Component<Props, State> {
...
+    selectedSource: getSelectedSource(state),
+    debuggeeUrl: getDebuggeeUrl(state),
+    expanded: getExpandedState(state),
+    projectRoot: getProjectDirectoryRoot(state)
+  };
+};
+
+const mapDispatchToProps = dispatch => {
+  return {
+    setExpandedState: () => dispatch(setExpandedState()),
```
wldcordeiro: Redux automatically binds dispatch to an object of functions so we could make this look like
```js
const actionCreators = {
  setExpandedState,
  selectSource
};

export default connect(mapStateToProps, actionCreators)(SourcesTree);
```

src/components/PrimaryPanes/SourcesTree.js
```diff
@@ -301,7 +302,7 @@ class SourcesTree extends Component<Props, State> {
   }
 
   render() {
-    const { setExpandedState, expanded } = this.props;
+    const expanded = this.props.expanded;
```
wldcordeiro: I don't get this change. You could have left it and made `onExpand/onCollapse` call it without `this.props`

src/components/PrimaryPanes/SourcesTree.js
```diff
@@ -310,6 +311,14 @@ class SourcesTree extends Component<Props, State> {
       highlightItems
     } = this.state;
 
+    const onExpand = (item, expandedState) => {
```
wldcordeiro: Both the functions are the same so how about `toggleExpanded`?

src/components/PrimaryPanes/SourcesTree.js
amelzer: The change was necessary after the implicit import of `setExpandedState` could have also renamed it, I guess.

src/components/PrimaryPanes/SourcesTree.js
amelzer: This is also a great way to do it!

src/components/PrimaryPanes/SourcesTree.js
jasonLaster: haha - it's small enough where this is fine too. Good either way

jasonLaster:
ping @amelzer

codehag:
This looks great! cant wait to see it merged!

amelzer:
@jasonLaster @codehag @wldcordeiro So with just the action creators (no usage of dispatch) this would be good to go? I can update it then.

jasonLaster:
yep

PR #4461, opened 2017-10-21
Converted source tree SVG's to images using CSS styles

jselmani:
Associated Issue: #4350 
#### Description
Converted the Source Tree SVG's to Images using CSS styles.  This should have a significant performance improvement for the project as rendering of SVG's is no longer required. Can perhaps also close Issue #4168 should the improvement be significant enough.

In regards to the Webpack SVG, with so many different paths being filled different colours, converting to an image would likely not reflect the same image that is portrayed currently as it will all be one colour if changed. 

##### Webpack SVG
webpacksvg
##### Fill Paths
fillpaths

### DIFF
--- a/src/components/shared/Svg.css
+++ b/src/components/shared/Svg.css
@@ -1,12 +1,15 @@
 .arrow,
-.folder,
-.domain,
-.file,
 .worker,
 .refresh,
 .shortcut,
 .add-button {
   fill: var(--theme-splitter-color);
+}
+
+.file,
+.folder,
+.domain {
+  background-color: var(--theme-splitter-color);
 }
 
 .worker,
@@ -24,8 +27,6 @@
   top: 1px;
 }
 
-.domain svg,
-.folder svg,
 .worker svg,
 .refresh svg,
 .shortcut svg,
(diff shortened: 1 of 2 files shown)


src/components/shared/Svg.css
```diff
@@ -1,14 +1,17 @@
...
 .refresh,
 .shortcut,
 .add-button {
   fill: var(--theme-splitter-color);
 }
 
+.file,
+.folder,
+.domain {
+  background-color: var(--theme-splitter-color);
```
James300: Out of curiosity, why does this use `background-color` whereas before it used `fill`?

jselmani: Fill is used to fill in paths of SVG's that are loaded into a webpage whereas background color loads a color into the div.  

With a mask being used over the div, the background color is what shows through the mask.  Hope that helps @James300

jasonLaster:
@jselmani mind sharing a screenshot?

jselmani:
Here they are @jasonLaster 

### Firefox
firefox

### Chrome
chrome

src/components/shared/Svg.css
James300: Thank you, that makes a lot of sense :+1:

jselmani:
Closing as new PR (#4477) was made to resolve these issues.

Issue #4259, opened 2017-10-03
[Sources] style the debuggee domain


jasonLaster:
It would be awesome if we could highlight the debuggee domain.

![](https://files.slack.com/files-tmb/T3V7H1198-F7DK9D8R4-7f39db2f6b/screen_shot_2017-10-03_at_11.39.22_am_480.png)

```diff
diff --git a/src/components/PrimaryPanes/SourcesTree.js b/src/components/PrimaryPanes/SourcesTree.js
index 151a774..1fb1f78 100644
--- a/src/components/PrimaryPanes/SourcesTree.js
+++ b/src/components/PrimaryPanes/SourcesTree.js
@@ -21,9 +21,11 @@ import {
   sortEntireTree,
   collapseTree,
   createTree,
-  getDirectories
+  getDirectories,
+  getUrl
 } from "../../utils/sources-tree";

+
 import ManagedTree from "../shared/ManagedTree";

 import actions from "../../actions";
@@ -230,9 +232,16 @@ class SourcesTree extends Component {
           : "paddingRight";
     }

+    const isDebuggeeDomain = getUrl(this.props.debuggeeUrl) ....
+    // we'll likely want to just look at the HOST because we only
+    // want to style the domain and not everything in the domain directory
+
     return (
       <div
-        className={classnames("node", { focused })}
+        className={classnames("node", {
+          focused,
+          debuggee: isDebuggee
+        })}
         style={{ [paddingDir]: `${depth * 15}px` }}
         key={item.path}
```

jasonLaster:
Hey @svitlana-galianova how's it going?

svitlana-galianova:
Hello @jasonLaster I am setting up a testing environment. Had some fun figuring how to make debugger run and make Firefox listen to the right port

svitlana-galianova:
@jasonLaster The code has been modified a bit and it's different from the one in the issue description. I have tried different things, but I feel like without understanding which domain is debuggee, I will be stuck in the same place for a while. 

Is using "focusedItem" state property a good idea to determine which domain is debuggee or maybe "highlightItems"? From every property I saw in the code, I feel like those two are the most suitable.

jasonLaster:
Hi @svitlana-galianova!

thanks for reaching out, the debuggee url is [here](https://github.com/devtools-html/debugger.html/blob/master/src/components/PrimaryPanes/SourcesTree.js#L363-L364)

[...]

PR #4463, opened 2017-10-21
[SourcesTree] Style debuggee URL domain

svitlana-galianova:
Associated Issue: #4259

### Summary of Changes

* added new svg style for domain icon
* wrote code to highlight debuggee domain in light green

### Test Plan

Example test plan:

- start debugger
- start debugging a web page
- observe the main domain being light green


James300:
That's really cool, looks good!

src/components/shared/Svg.css
```diff
@@ -75,6 +75,10 @@ html .arrow.expanded svg {
   transform: rotate(0deg);
 }
 
+.domain.debuggee {
+  fill: lightgreen;
```
wldcordeiro: Note: Maybe there's a green defined in the themes we can use here @jasonLaster?

svitlana-galianova:
@wldcordeiro thanks! It won't be a problem to change a color to whatever you think is the most suitable.

src/components/PrimaryPanes/SourcesTree.js
```diff
@@ -216,7 +216,10 @@ class SourcesTree extends Component {
     }
 
     if (depth === 0) {
-      return <Svg name="domain" />;
+      return <Svg name="domain" 
+       className={classnames({
+          debuggee: (this.props.debuggeeUrl != null && item.name != null && this.props.debuggeeUrl.indexOf(item.name) !== -1) ? true : false
```
nt1m: This can be simplified to:

```js
debuggee: this.props.debuggeeUrl && item.name && this.props.debuggeeUrl.includes(item.name),
```
You don't need the ternary that does `true : false` since your condition is already a boolean.

src/components/shared/Svg.css
nt1m: var(--theme-toolbar-highlighted-color) ?

src/components/PrimaryPanes/SourcesTree.js
wldcordeiro: You do need the `!= null` checks to avoid Flow complaining about sketchy existence checks.

svitlana-galianova: @nt1m Thanks for your comment, I thought it would be smarter to use indexOf since it's supported in all the browsers and includes is only supported in the following browsers (according to w3school):
browser support

codehag: We use `includes` in other areas of the code base, you can feel safe using it :). We are preferring readable code to compatibility with old browsers. The reason for this is we are updating along with Firefox, which is currently at version 56, going on to 57, and includes is available there. We also support recent versions of chrome, but supporting anything from  years ago does not make sense given the scope of the project. It is great that you are thinking of this and is a really useful skill in doing front end!

svitlana-galianova: @codehag Thanks for your clarification!

src/components/shared/Svg.css
svitlana-galianova: @nt1m I tried your suggestion and here is a screenshot of what I got:

screenshot

To tell you the truth, I like green color more.

jasonLaster: Actually, we don't need a color... We'll style it later :)

wldcordeiro: We have `--green-50` that would work. http://design.firefox.com/photon/visuals/color.html#green

jasonLaster: The green in the mockup was a throw away color. I'd like to look at all the colors in tree holistically soon.

jasonLaster:
I think this needs a rebase

wldcordeiro:
Hmm so this is problematic. Since we've been moving the SVGs in the Debugger to images this PR can no longer work (I tried pulling down and rebasing). @jasonLaster ideas?

src/components/PrimaryPanes/SourcesTree.js
```diff
@@ -216,7 +216,10 @@ class SourcesTree extends Component {
     }
 
     if (depth === 0) {
-      return <Svg name="domain" />;
+      return <Svg name="domain" 
+       className={classnames({
```
jasonLaster: can we add a className to the new `img`?

wldcordeiro: Unfortunately while we can add a className it doesn't have any effect. SVGs have to be inlined for you to be able to style them with CSS.

jasonLaster: we can do a mask/background color trick. we do it elsewhere

nt1m: mozilla-central uses -moz-context-properties, but that only works for Firefox's chrome. As for debugger.html, we can either use an SVG filter, then the fill property works fine from the CSS (but SVG filters have a perf-cost). The other possibly is a mask (but that instead uses background-color instead of fill).

svitlana-galianova: @jasonLaster I am not quite clear, is there any action required from me at this point?

jasonLaster: @svitlana-galianova yep. just add the className to the new `img`. we'll style it later

wldcordeiro:
Uh oh, this looks like it needs a rebase. It's gotten a lot of additional commits on it that aren't part of the PR.

jasonLaster:
I feel like this pr got stuck in the proverbial git mud, so i cherrypicked the good stuff and put it in a new PR https://github.com/devtools-html/debugger.html/pull/4536

@svitlana-galianova you're still the author of the new PR's commit. Thanks for the hard work while we figured out what to do here.

flodolo:
Passing by comment: it looks like you have worked directly on the master branch of your fork. Sadly, when using a fork, things will become unmanageable really fast, especially on a repository with a lot of changes like Debugger.

The safest workflow is to keep your master in sync with the original repository's master, and create a branch to work on pull requests. At that point, if needed, you'd only need to rebase your branch against master.

Once you have completed all the pending work, my personal suggestion would be to "reset" your master and sync it with this repository's master, so you have a clean starting point for future work.

```
git remote add upstream https://github.com/devtools-html/debugger.html
git fetch upstream
git reset --hard origin/upstream
git push -f
```

Issue #4396, opened 2017-10-17
"Add Condition" Context Menu Triggers Error


darkwing:
Right-clicking a breakpoint in the right column, then selecting "Add Condition", triggers the following error:

"TypeError: toggleConditionalBreakpointPanel is not a function
<img width="1400" alt="error" src="https://user-images.githubusercontent.com/46655/31643659-7d31090e-b2b7-11e7-9869-89d5b92d16ad.png">
"

nyrosmith:
@lukaszsobek 
```diff
diff --git a/src/components/SecondaryPanes/Breakpoints.js b/src/components/SecondaryPanes/Breakpoints.js
index 2f4b58c..9333d89 100644
--- a/src/components/SecondaryPanes/Breakpoints.js
+++ b/src/components/SecondaryPanes/Breakpoints.js
@@ -44,7 +44,7 @@ type Props = {
   toggleAllBreakpoints: boolean => void,
   toggleDisabledBreakpoint: number => void,
   setBreakpointCondition: Location => void,
-  toggleConditionalBreakpointPanel: number => void
+  openConditionalBreakpointPanel: number => void
 };
 
 function isCurrentlyPausedAtBreakpoint(pause, breakpoint) {
@@ -271,7 +271,7 @@ class Breakpoints extends PureComponent {
       accesskey: addConditionKey,
       click: () => {
         this.selectBreakpoint(breakpoint);
-        toggleConditionalBreakpointPanel(breakpoint.location.line);
+        this.toggleConditionalBreakpointPanel(breakpoint.location.line);
       }
     };
 
@@ -281,7 +281,7 @@ class Breakpoints extends PureComponent {
       accesskey: editConditionKey,
       click: () => {
         this.selectBreakpoint(breakpoint);
-        toggleConditionalBreakpointPanel(breakpoint.location.line);
+        this.toggleConditionalBreakpointPanel(breakpoint.location.line);
       }
     };
 
@@ -343,7 +343,8 @@ class Breakpoints extends PureComponent {
   }
 
   toggleConditionalBreakpointPanel(line) {
-    this.props.toggleConditionalBreakpointPanel(line);
+    debugger
+    this.props.openConditionalPanel(line);
   }
 
   renderBreakpoint(breakpoint) {
```

PR #4465, opened 2017-10-22
Fix conditional breakpoints

nyrosmith:
Associated Issue: #4396 

### Summary of Changes
* fixed opening the ConditionalPanel
* fixed setting the breakpoint id

### Test Plan
- [x] Manual testing only


jasonLaster:
Nice!

Still need to test and think about it. Saw some test failures :/

jasonLaster:
@nyrosmith waiting on the test cleanups

nyrosmith:
@jasonLaster I needed to wait for the CRLF fixes. Otherwise the test would always fail, no matter what I do.
Will work on this during the weekend

jasonLaster:
ahh, great... perhaps i can pull and rebase

jasonLaster:
@nyrosmith - not sure where the test fails are coming from...

src/actions/breakpoints/tests/__snapshots__/syncing.js.snap
```diff
@@ -18,11 +20,34 @@ Object {
       "sourceUrl": "http://localhost:8000/gen.js",
     },
     "hidden": false,
+    "id": Object {
```
jasonLaster: why are storing the ID?

src/actions/breakpoints/tests/__snapshots__/syncing.js.snap
```diff
@@ -18,11 +20,34 @@ Object {
...
+        "sourceUrl": "http://localhost:8000/gen.js",
+      },
+      "hidden": false,
+      "location": Object {
+        "column": undefined,
+        "line": 3,
+        "sourceId": "magic.js",
+        "sourceUrl": "http://localhost:8000/magic.js",
+      },
+    },
```
jasonLaster: this is a lot of new data to be storing...

nyrosmith: @jasonLaster we are already looking into it.
@bomsy is helping me out finding what's going on here

src/actions/breakpoints/tests/__snapshots__/syncing.js.snap
codehag: this was an issue earlier on, when we were not syncing the id -- the result was that synced breakpoints could not be deleted/turned into conditionals.

codehag: we need to store the id in the sync step because earlier we were not, and this was creating breakpoints that could not be changed/deleted. however it shouldnt be such a large object, just a string

src/actions/breakpoints/syncBreakpoint.js
```diff
@@ -102,20 +103,21 @@ export async function syncClientBreakpoint(
...
   // the breakpoint might have slid server side, so we want to get the location
   // based on the server's return value
-  const newGeneratedLocation = clientBreakpoint.actualLocation;
+  const newGeneratedLocation = actualLocation;
   const newLocation = await sourceMaps.getOriginalLocation(
     newGeneratedLocation
   );
 
   return createSyncData(
+    id,
```
codehag: we are missing this change on line 92 of this file

src/actions/breakpoints/syncBreakpoint.js
```diff
@@ -88,6 +89,7 @@ export async function syncClientBreakpoint(
   // send update only to redux
   if (pendingBreakpoint.disabled || (existingClient && isSameLocation)) {
     return createSyncData(
+      pendingBreakpoint.location.sourceId,
```
codehag: this should be `existingClient.id`

Issue #4394, opened 2017-10-17
"Copy Function" Context Menu Item Seems Broken


darkwing:
Right-clicking inside a function does not enable the "Copy Function" context menu.  
<img width="353" alt="context" src="https://user-images.githubusercontent.com/46655/31642233-7b1b9b14-b2af-11e7-86d1-1878af65d140.png">
Something may be off.

PR #4144, opened 2017-09-23
Shortcut to jump to mapped location

GarethSharpe:
Associated Issue: #4062 

Here's the Pull Request Doc
https://devtools-html.github.io/debugger.html/CONTRIBUTING.html#pull-requests

### Summary of Changes

* Created mouse click event listener for codeMirror
* Created onClick(e: MouseEvent) function shortcut to jump to source location

### Test Plan

- [x] When click + metaKey + altKey, jump to source location

Here's the Debugger's Testing doc
https://docs.google.com/document/d/1oBMRxV8A2ag2t22YsQOxTdEv0mXKzIg0tggJjRkU1S0/edit#.
Feel free to improve it!

### Screenshots/Videos

debugger - mozilla firefox - jump to context

PR #4466, opened 2017-10-22
Fix "Copy function"

nyrosmith:
Associated Issue: #4394 

I also checked the other occurences of  `coordsChar` but they are ok.


jasonLaster:
Do you know when this broke?

nyrosmith:
@jasonLaster https://github.com/devtools-html/debugger.html/pull/4144/files#diff-631f065a7bf985e3183635b0acc27205R52

jasonLaster:
Thanks!

PR #4281, opened 2017-10-04
(WIP) Editor highlight errors

nyrosmith:
Associated Issue: #3112 

This is incomplete and needs some fixes. But I need feedback from CI

Current state
highlight errors

Issue #3112, opened 2017-06-06
[Paused] Highlight the debug line red on errors


jasonLaster:
It would be sweet if the debugger could show you that it's paused on an exception based on the debug line color:

<img src='https://shipusercontent.com/a7255d9c46e615fbcf70f212ea94f2b0/Screen%20Shot%202017-06-06%20at%204.14.42%20PM.png' title='Screen Shot 2017-06-06 at 4.14.42 PM.png' width=794>

Here's a rough sketch
```diff
diff --git a/src/components/Editor/index.js b/src/components/Editor/index.js
index d85e265..16ec036 100644
--- a/src/components/Editor/index.js
+++ b/src/components/Editor/index.js
@@ -166,10 +166,10 @@ class Editor extends PureComponent {
     }

     if (this.props.outOfScopeLocations !== nextProps.outOfScopeLocations) {
-      clearLineClass(this.editor.codeMirror, "out-of-scope");
+      clearLineClass(this.editor.codeMirror, "out-of-scope", this.props.pauseInfo.why);
     }

-    this.setDebugLine(nextProps.selectedFrame, selectedLocation);
+    this.setDebugLine(nextProps.selectedFrame, selectedLocation, nextProps.pauseInfo.why);
     resizeBreakpointGutter(this.editor.codeMirror);
   }

@@ -588,25 +588,29 @@ class Editor extends PureComponent {
     }
   }

+  getDebugLineClass(whyPaused) {
+    return whyPaused.error ? "new-debug-line" : "new-debug-line-error"
+  }
+
   clearDebugLine(selectedFrame) {
     if (selectedFrame) {
       const line = selectedFrame.location.line;
       this.editor.codeMirror.removeLineClass(
         line - 1,
         "line",
-        "new-debug-line"
+        this.getDebugLineClass(whyPaused)
       );
     }
   }
:
```

PR #4467, opened 2017-10-22
Highlight errors in editor

nyrosmith:
Associated Issue: #3112 
I closed #4281 because after the editor refactoring I wasn't able to merg/rebase anymore.

highlight_editor_error

### DIFF
--- a/src/components/Editor/DebugLine.js
+++ b/src/components/Editor/DebugLine.js
@@ -4,12 +4,17 @@ import { markText, toEditorPosition } from "../../utils/editor";
 import { getDocument } from "../../utils/editor/source-documents";
 
 import { connect } from "react-redux";
-import { getSelectedLocation, getSelectedFrame } from "../../selectors";
+import {
+  getSelectedLocation,
+  getSelectedFrame,
+  getPause
+} from "../../selectors";
 
 type Props = {
   editor: Object,
   selectedFrame: Object,
-  selectedLocation: Object
+  selectedLocation: Object,
+  pauseInfo: Object
 };
 
 type State = {
@@ -26,6 +31,7 @@ export class DebugLine extends Component<Props, State> {
 
   componentDidMount() {
     this.setDebugLine(
+      this.props.pauseInfo,
       this.props.selectedFrame,
       this.props.selectedLocation,
       this.props.editor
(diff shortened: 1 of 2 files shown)


src/components/Editor/Editor.css
```diff
@@ -3,6 +3,8 @@
   --debug-expression-background: rgba(202, 227, 255, 0.5);
   --editor-searchbar-height: 27px;
   --editor-second-searchbar-height: 27px;
+  --debug-line-error-border: rgb(255, 0, 0);
+  --debug-expression-error-background: rgba(255, 127, 127, 0.5);
```
nyrosmith: Not sure which color to use as background.
This looked ok for me...

src/components/Editor/DebugLine.js
```diff
@@ -90,7 +109,11 @@ export class DebugLine extends Component<Props, State> {
       return;
     }
 
+    // if (pauseInfo && pauseInfo.why.type === "exception") {
```
nyrosmith: I forgot to remove this

src/components/Editor/Editor.css
bomsy: - the background feels a bit strong , maybe lighter, or no background with just the red border lines

jasonLaster: Let's do something light :)

src/components/Editor/DebugLine.js
```diff
@@ -46,32 +53,44 @@ export class DebugLine extends Component<Props, State> {
...
-    this.setState({ debugExpression });
+    if (pauseInfo && pauseInfo.why.type === "exception") {
+      doc.addLineClass(line, "line", "new-debug-line-error");
+      const debugExpression = markText(editor, "debug-expression-error", {
+        start: { line, column },
+        end: { line, column: null }
+      });
+      this.setState({ debugExpression });
+    } else {
+      doc.addLineClass(line, "line", "new-debug-line");
```
codehag: these two branches are almost exactly the same, with the exception of the the class names. how about something like this:

```javascript
type TextClasses = {
  markTextClass: string,
  lineClass: string
}
getTextClasses(pauseInfo: Object): TextClasses {
  if (pauseInfo && pauseInfo.why.type === "exception") {
    return { markTextClass: "debug-expression-error", lineClass: "debug-expression-error" }
  }
  return { markTextClass: "debug-expression", lineClass: "new-debug-line" }
}

// ...

const { markTextClass, lineClass } = getTextClasses(pauseInfo);
doc.addLineClass(line, "line", markTextClass);
const debugExpression = markText(editor, lineClass, {
  start: { line, column },
  end: { line, column: null }
});
this.setState({ debugExpression });
```

this way, it is clear what is changing in the code, its not the entire if block, but just the class names

nyrosmith: You are right. Will change that

Issue #4020, opened 2017-09-14
[Preview] Tooltip shows undefined for variable used as object key


juliandescottes:
STRs:

run in the console

```js
(function (key) {
  debugger;
  let obj = {};
  obj[key] = 0;
  return obj;
})("a");
```
hover on "key" in obj[key]

Expected result: should show `"a"`
Actual result: shows `undefined`

juliandescottes:
![](https://screenletstore.appspot.com/img/d05e00d4-9983-11e7-b141-ddd9e672c8a3.png)

bomsy:
The issue relates to the computed and uncomputed MemberExpressions. we are looking at `obj[key]` (computed) as `obj.key` (uncomputed)

juliandescottes:
Thanks for fixing this @bomsy ! Just happened to test this in the last debugger bundle update and this works fine now :)

PR #4468, opened 2017-10-22
Differentiate between computed and uncomputed member expressions

bomsy:
Associated Issue: #4020

### Summary of Changes

* Add the computed property to the member expression objects returned from getSymbols
* Differentiate computed from un-computed memberexpressions

Example test plan:

- [x] Debug todomvc example
- [x] open `backbone.js` and set a breakpoint on line 223
- [x] Refresh the debuggee (which hits the breakpoint)
- [x] Hover over `id` in `listeningTo[id] = obj`
- [x] It should show `l2`

### DIFF
--- a/src/actions/ast.js
+++ b/src/actions/ast.js
@@ -117,7 +117,8 @@ function findBestMatch(symbols, tokenPos, token) {
     const overlaps =
       expression.location.start.line == line &&
       expression.location.start.column <= column &&
-      expression.location.end.column >= column;
+      expression.location.end.column >= column &&
+      !expression.computed;
 
     if (overlaps) {
       return expression;
--- a/src/actions/ast.js
+++ b/src/actions/ast.js
@@ -18,6 +18,8 @@ import {
   getOutOfScopeLocations
 } from "../workers/parser";
 
+import { findBestMatchExpression } from "../utils/ast";
+
 import { isGeneratedId } from "devtools-source-map";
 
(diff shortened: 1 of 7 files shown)


src/actions/ast.js
```diff
@@ -117,7 +117,8 @@ function findBestMatch(symbols, tokenPos, token) {
     const overlaps =
       expression.location.start.line == line &&
       expression.location.start.column <= column &&
-      expression.location.end.column >= column;
+      expression.location.end.column >= column &&
+      !expression.computed;
```
jasonLaster: this is not related to this patch, but it would be nice to move `findBestMatch` to a utility and add some unit tests.

jasonLaster:
Looks great!

src/actions/ast.js
bomsy: Yeah! thanks for the confirm, thought about that ..but was wondering if we do it in another PR. i think it'll be cool to do here..

src/utils/ast.js
```diff
@@ -0,0 +1,17 @@
...
+      expression.location.end.column >= column &&
+      !expression.computed;
+
+    if (overlaps) {
+      return expression;
+    }
+
+    return found;
+  }, {});
+}
```
bomsy: @jasonLaster was not sure if this is the best place to put this. ur tots!

jasonLaster: not a bad spot :)

Issue #4469, opened 2017-10-23
Right-click in editor window shows 2 context menus


lukaszsobek:
* open editor window
* right click inside the text
2 context menus appear (debugger + default)
* press esc to hide one
* press esc to hide the other

![31880096-d33ad848-b7df-11e7-95e0-4673e8debf74](https://user-images.githubusercontent.com/23530054/31881679-8e5e9d40-b7e4-11e7-8aa5-c3c0e612b56a.gif)

Expected:
Only the debugger context menu shows

jainsneha23:
@bomsy Please close the issue.

PR #4470, opened 2017-10-23
Fix for #4469 Right-click in editor window shows 2 context menus

jainsneha23:
Associated Issue: #4469 

### Summary of Changes

* Added event stopPropagation and prevent default to context menu event

### Test Plan

- [x] open editor window
- [x] right click inside the text
- [x] 1 context menu appear (debugger)


src/components/Editor/index.js
```diff
@@ -166,7 +166,11 @@ class Editor extends PureComponent<Props, State> {
         return this.onGutterContextMenu(event);
       });
 
-      codeMirror.on("contextmenu", (cm, event) => this.openMenu(event));
+      codeMirror.on("contextmenu", (cm, event) => {
+        event.stopPropagation();
+        event.preventDefault();
+        return this.openMenu(event);
```
jasonLaster: minor point, but for aesthetics can we move `event.stopPropagation` and `event.preventDefault` to `openMenu`. It'll only ever be called here

jainsneha23:
There is one more bug here. 
right click -> context menu appears -> right click on a different point
there will be 2 context menu - one debugger , and other browser

Seems like codemirror issue

screen shot 2017-10-23 at 10 13 07 pm

jasonLaster:
we can ignore the 2nd issue.. it's only true in the launchpad

PR #4475, opened 2017-10-23
Fix cross-configuration problems

Garbee:
### Summary of Changes

* Added gitattributes to force LF line endings in checkouts on Windows.
* Converted existing CLRF endings to LF
* No colors on tests for cross-configuration stability.

### Test Plan

Run `yarn test` on multiple platforms and configs (Windows and Linux in my case) and see it succeed on fresh clones without any other changes.

Using `--no-colors` for the test parameter is a hack since we don't have a community serializer (that I can find) to strip colors out of the snapshots. So we have a not-so-good looking test output. But, it functions across more platforms and configurations than running it with colors does do. So until a serializer is found/built to do this transparently having `--no-colors` as the default is our best option to get things working.


jasonLaster:
Looks good!

jasonLaster:
@Garbee seeing some test failures

Also, i saw this today in the lerna project... might be interesting: https://www.npmjs.com/package/normalize-newline

jasonLaster:
lol, it's tiny: https://github.com/sindresorhus/normalize-newline/blob/master/index.js#L7-L11

PR #4477, opened 2017-10-24
Converted Source Tree SVG's (including Blackbox) to images using CSS styles (#4350)

jselmani:
Associated Issue: #4350 

Converted the Source Tree SVG's to images using CSS styles.  This also includes the newly added Blackbox SVG.

#### Firefox
firefox

#### Chrome
chrome


jselmani:
@jasonLaster can we update issue #4350.  I'll get started on the step buttons next 💯

Issue #4418, opened 2017-10-18
"Reveal in Tree" Context Menu Item Should Switch to "Sources" Tab if "Outline" tab selected.


darkwing:
At the moment, if the "Outline" tab is selected, nothing appears to happen.

darkwing:
Same with using "Reveal in Tree" in a source tab's context menu;

PR #4480, opened 2017-10-24
"Reveal in Tree": jump to sources tree

nyrosmith:
Associated Issue: #4418 

### Summary of Changes

* Move local state out of PrimaryPane into ui reducer
* fire action `setPrimaryPaneTab`

reveal_in_tree


src/actions/ui.js
```diff
@@ -1,14 +1,24 @@
...
+} from "../reducers/ui";
 
 export function setContextMenu(type: string, event: any) {
   return ({ dispatch }: ThunkArgs) => {
     dispatch({ type: "SET_CONTEXT_MENU", contextMenu: { type, event } });
   };
 }
 
+export function setPrimaryPaneTab(tabName: SelectedPrimaryPaneTabType) {
+  return ({ dispatch }: ThunkArgs) => {
```
wldcordeiro: I don't think this action needs to have a thunk. Could do with a plain action?

wldcordeiro:
A note about an action creator but this looks great @nyrosmith!

Issue #4420, opened 2017-10-18
Context Menu "Jump to Original Location" should be disabled when not pretty-printed


darkwing:
I don't believe this context menu item has any value outside of a pretty-printed tab (i.e. an original source tab).  It simply highlights the current line.

jainsneha23:
@darkwing  @jasonLaster 
Shouldn't we be disabling "Reveal in tree" and "Blackbox script" for pretty printed files as well?

PR #4497, opened 2017-10-26
#4420 Context Menu "Jump to Original Location" should be disabled when not pretty-printed

jainsneha23:
Associated Issue: #4420 

### Summary of Changes

* Get if selected source is pretty-printed. If yes, disable "Reveal in tree" and "Blackbox script"
* Get open tabs list. Get pretty url for the selected source. If pretty tab is open, disable "Jump to Original Location"

### Test Plan

- [x] Open a tab, "Jump to Original Location" is disabled
- [x] Click on pretty print in the footer
- [x] "Jump to Original Location" is enabled
- [x] Go to the pretty printed tab
- [x] "Reveal in tree" and "Blackbox script" are disabled

### DIFF
--- a/src/components/Editor/EditorMenu.js
+++ b/src/components/Editor/EditorMenu.js
@@ -6,6 +6,7 @@ import { PureComponent } from "react";
 import { showMenu } from "devtools-launchpad";
 import { isOriginalId } from "devtools-source-map";
 import { copyToTheClipboard } from "../../utils/clipboard";
+import { isPretty, getPrettySourceURL } from "../../utils/source";
 import { getSourceLocationFromMouseEvent } from "../../utils/editor";
 import { bindActionCreators } from "redux";
 import { connect } from "react-redux";
@@ -15,7 +16,8 @@ import {
   getContextMenu,
   getSelectedLocation,
   getSelectedSource,
-  getSymbols
+  getSymbols,
+  getSourceTabs
 } from "../../selectors";
 
 import actions from "../../actions";
@@ -30,6 +32,7 @@ function getMenuItems(
     editor,
     selectedLocation,
     selectedSource,
+    tabs,
     showSource,
     onGutterContextMenu,
     jumpToMappedLocation,
@@ -83,14 +86,19 @@ function getMenuItems(
     event
   );
 
-  const pairedType = isOriginalId(selectedLocation.sourceId)
-    ? L10N.getStr("generated")
-    : L10N.getStr("original");
+  const isOriginal = isOriginalId(selectedLocation.sourceId);
+  const isPrettyTabOpen =
+    !isOriginal && tabs.includes(getPrettySourceURL(selectedSource.get("url")));
+  const hasSourceMap = selectedSource.toJS().sourceMapURL;
+  const isPrettyPrinted = isPretty(selectedSource.toJS());
 
   const jumpLabel = {
     accesskey: L10N.getStr("editor.jumpToMappedLocation1.accesskey"),
-    disabled: false,
-    label: L10N.getFormatStr("editor.jumpToMappedLocation1", pairedType),
+    disabled: isOriginal || hasSourceMap ? false : !isPrettyTabOpen,
+    label: L10N.getFormatStr(
+      "editor.jumpToMappedLocation1",
+      isOriginal ? L10N.getStr("generated") : L10N.getStr("original")
+    ),
     click: () => jumpToMappedLocation(sourceLocation)
   };
 
@@ -104,7 +112,7 @@ function getMenuItems(
     id: "node-menu-blackbox",
     label: toggleBlackBoxLabel,
     accesskey: blackboxKey,
-    disabled: false,
+    disabled: isOriginal || hasSourceMap,
     click: () => toggleBlackBox(selectedSource.toJS())
   };
 
@@ -115,7 +123,7 @@ function getMenuItems(
     id: "node-menu-show-source",
     label: revealInTreeLabel,
     accesskey: revealInTreeKey,
-    disabled: false,
+    disabled: isPrettyPrinted,
     click: () => showSource(selectedSource.get("id"))
   };
 
@@ -184,9 +192,11 @@ class EditorMenu extends PureComponent {
 export default connect(
   state => {
     const selectedSource = getSelectedSource(state);
+    const tabs = getSourceTabs(state);
     return {
       selectedLocation: getSelectedLocation(state),
       selectedSource,
+      tabs,
       contextMenu: getContextMenu(state),
       getFunctionText: line =>
         findFunctionText(
--- a/src/components/Editor/EditorMenu.js
+++ b/src/components/Editor/EditorMenu.js
@@ -4,9 +4,9 @@
 
 import { PureComponent } from "react";
 import { showMenu } from "devtools-launchpad";
-import { isOriginalId } from "devtools-source-map";
+import { isOriginalId, isGeneratedId } from "devtools-source-map";
 import { copyToTheClipboard } from "../../utils/clipboard";
-import { isPretty, getPrettySourceURL } from "../../utils/source";
+import { isPretty } from "../../utils/source";
 import { getSourceLocationFromMouseEvent } from "../../utils/editor";
 import { bindActionCreators } from "redux";
 import { connect } from "react-redux";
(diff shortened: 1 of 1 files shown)


src/components/Editor/EditorMenu.js
```diff
@@ -75,14 +78,18 @@ function getMenuItems(
...
+  const isGenerated = isOriginalId(selectedLocation.sourceId);
+  const isPrettyTabOpen =
+    !selectedSource.get("isPrettyPrinted") &&
+    tabs.includes(getPrettySourceURL(selectedSource.get("url")));
 
   const jumpLabel = {
     accesskey: L10N.getStr("editor.jumpToMappedLocation1.accesskey"),
-    disabled: false,
-    label: L10N.getFormatStr("editor.jumpToMappedLocation1", pairedType),
+    disabled: isGenerated ? false : !isPrettyTabOpen,
```
jasonLaster: I'm a bit confused by this so i'll share the spec as i see it:

- original files can not be pretty printed
- generated files that have a source map cannot be pretty printed
- if the source has been pretty printed, jumping should jump between the pretty and "ugly" file :) this might be new behavior...

src/components/Editor/EditorMenu.js
```diff
@@ -96,7 +103,7 @@ function getMenuItems(
     id: "node-menu-blackbox",
     label: toggleBlackBoxLabel,
     accesskey: blackboxKey,
-    disabled: false,
+    disabled: isGenerated,
```
jasonLaster: so... i think the comment before might help...
we want it to be disabled for pretty printed files, but nothing else...

jainsneha23:
@jasonLaster 
The spec for this is:

```
jumpLabel = {
  isOriginal ? false : !isPrettyTabOpen
}
```
If the file opened is pretty printed(i.e. isOriginal), u will see "jump to generated location" always
Else if its generated, "Jump to Original Location" will be disabled if "!isPrettyTabOpen". As if its not pretty printed, you cannot jump to the original location

`
const blackBoxMenuItem = {
    disabled: isOriginal
  };
`
"Blackbox source" and "Reveal in tree" are also disabled for pretty printed (i.e, original source)

src/components/Editor/EditorMenu.js
```diff
@@ -75,14 +78,17 @@ function getMenuItems(
     event
   );
 
-  const pairedType = isOriginalId(selectedLocation.sourceId)
-    ? L10N.getStr("generated")
-    : L10N.getStr("original");
+  const isOriginal = isOriginalId(selectedLocation.sourceId);
```
jainsneha23: Check if the source is original, or generated ?

src/components/Editor/EditorMenu.js
```diff
@@ -75,14 +78,17 @@ function getMenuItems(
...
+  const isPrettyTabOpen =
```
jainsneha23: If the source is generated, check if there is a open tab with its original

src/components/Editor/EditorMenu.js
```diff
@@ -75,14 +78,17 @@ function getMenuItems(
...
     accesskey: L10N.getStr("editor.jumpToMappedLocation1.accesskey"),
-    disabled: false,
-    label: L10N.getFormatStr("editor.jumpToMappedLocation1", pairedType),
+    disabled: isOriginal ? false : !isPrettyTabOpen,
+    label: L10N.getFormatStr(
+      "editor.jumpToMappedLocation1",
+      isOriginal ? L10N.getStr("generated") : L10N.getStr("original")
+    ),
     click: () => jumpToMappedLocation(sourceLocation)
   };
```
jainsneha23: Jump to generated location is always enabled for original source
    
Jump to original location is enabled only if,  its original location is open in a tab

src/components/Editor/EditorMenu.js
```diff
@@ -96,7 +102,7 @@ function getMenuItems(
     id: "node-menu-blackbox",
     label: toggleBlackBoxLabel,
     accesskey: blackboxKey,
-    disabled: false,
+    disabled: isOriginal,
     click: () => toggleBlackBox(selectedSource.toJS())
```
jainsneha23: Blackbox menu is disabled for original source (prettified)

jasonLaster:
Hey @jainsneha23 lets discuss O/G here:

https://github.com/devtools-html/debugger.html/pull/4512/files?short_path=c474cab#diff-c474cab07b2dd9f949f47376f7d47fd2

i'm still a bit confused about how you're using the term... we might also be inconsistent in the app

jainsneha23:
@jasonLaster Even I am confused here for terminology. For me the pretty printed file is generated, and the one i open first is original. But the terminology used in the same file before was the opposite. So, I sticked to it. If the code looks fine apart from the terminology for generated/original, please suggest the proper names. I will add it.

jasonLaster:
^ haha - let me check to see what we do :)

jainsneha23:
So, if the pretty source is original , the terminology being used right now is correct. Then the code can is ready to review?

src/components/Editor/EditorMenu.js
```diff
@@ -75,14 +78,17 @@ function getMenuItems(
     event
   );
 
-  const pairedType = isOriginalId(selectedLocation.sourceId)
-    ? L10N.getStr("generated")
-    : L10N.getStr("original");
+  const isOriginal = isOriginalId(selectedLocation.sourceId);
+  const isPrettyTabOpen =
+    !isOriginal && tabs.includes(getPrettySourceURL(selectedSource.get("url")));
```
jasonLaster: would this fail if we were looking at a different original file?

I think we should be able to get the selectedSource, which would have the URL and ignore the tab stuff

src/components/Editor/EditorMenu.js
jasonLaster: true, we don't want to blackbox original sources. This is because we don't know how to blackbox an original source that is part of a larger bundled (generated) source

src/components/Editor/EditorMenu.js
```diff
@@ -107,7 +113,7 @@ function getMenuItems(
     id: "node-menu-show-source",
     label: revealInTreeLabel,
     accesskey: revealInTreeKey,
-    disabled: false,
+    disabled: isOriginal,
```
jasonLaster: pretty printed files are not in the tree, but other original files that come from a bundle are.

src/components/Editor/EditorMenu.js
```diff
@@ -168,9 +174,11 @@ class EditorMenu extends PureComponent {
 export default connect(
   state => {
     const selectedSource = getSelectedSource(state);
+    const tabs = getSourceTabs(state);
     return {
       selectedLocation: getSelectedLocation(state),
       selectedSource,
```
jasonLaster: oh good, we have a selectedSource

src/components/Editor/EditorMenu.js
jasonLaster: i like the idea of having an `isPrettyPrinted` variable

src/components/Editor/EditorMenu.js
jainsneha23: Correct. I will fix this.

src/components/Editor/EditorMenu.js
jainsneha23: Yes, we will need both isOriginal and isPrettyPrinted. As they can be different.

src/components/Editor/EditorMenu.js
```diff
@@ -75,14 +78,19 @@ function getMenuItems(
...
+  const isPrettyTabOpen =
+    !isOriginal && tabs.includes(getPrettySourceURL(selectedSource.get("url")));
+  const hasSourceMap = selectedSource.toJS().sourceMapURL;
+  const isPrettyPrinted = isPretty(selectedSource.toJS());
 
   const jumpLabel = {
     accesskey: L10N.getStr("editor.jumpToMappedLocation1.accesskey"),
-    disabled: false,
-    label: L10N.getFormatStr("editor.jumpToMappedLocation1", pairedType),
+    disabled: isOriginal || hasSourceMap ? false : !isPrettyTabOpen,
```
bomsy: Can we use `!isPrettyPrinted` here instead of `!isPrettyTabOpen`
then we won't need line 82??

jasonLaster:
did a rebase here: https://github.com/jasonLaster/debugger.html/tree/jainsneha23-bug-4420

sorry, i couldn't push to your branch

jasonLaster:
I think this is good to go. Sorry for the delay on reviewing.

I ended up doing some small tweaks to clean it up as it's super confusing

jainsneha23:
Great

jasonLaster:
screen shot 2017-11-14 at 11 16 09 am

Issue #4519, opened 2017-10-27
Breakpoint is set, but not visible


yurydelendik:
STR (in Nightly):

1. Open devtools on the Debugger tab
2. Load https://mdn.github.io/webassembly-examples/js-api-examples/memory.html in the current tab.
3. Set a breakpoint before the ‘i32.load’ operation.
4. Reload the page.
5. Observe paused location.

Actual result:
On step 3, the breakpoint disappears; and on step 5, still not visible (even if it was paused)

Expected:
The breakpoint is visible and listed in the breakpoints panel

yurydelendik:
I still can observe this issue in the nightly

wldcordeiro:
Not sure the status here. I see two merged PRs @yurydelendik could you update?

yurydelendik:
We can close it as fixed by #4817 -- there was issues that prevented to verify it.

PR #4520, opened 2017-10-27
Fix parser returning failing to clone data

jasonLaster:
Associated Issue: #4519

### Summary of Changes

The parser was returning an object that could not be serialized.

### Test Plan

adding a mochitest to prevent new regressions


src/workers/parser/getSymbols.js
```diff
@@ -80,6 +80,14 @@ function getComments(ast) {
   }));
 }
 
+function getSpecifiers(specifiers) {
+  if (!specifiers) {
+    return;
```
yurydelendik: can you return `null` or `undefined` here? just to know result is expected.

src/workers/parser/utils/formatSymbols.js
```diff
@@ -23,7 +23,10 @@ export function formatSymbols(source: Source) {
     const expression = symbol.expression || "";
     const klass = symbol.klass || "";
 
-    return `${loc} ${exprLoc} ${expression} ${symbol.name}${params} ${klass}`.trim(); // eslint-disable-line max-len
+    const name = symbol.name || "";
+    const names = (symbol.specifiers || []).join(", ");
```
yurydelendik: to avoid garbage array creation `const names = symbol.specifiers ? symbol.specifiers.join(", ") : "";`

jasonLaster: haha - yeah, that makes sense. `format` is just for testing

PR #4521, opened 2017-10-27
More advanced original source scopes mapping.

yurydelendik:
Associated Issue: Display synthesized scopes (e.g. ES6) instead of transpiled

Test case:

1. Open https://yurydelendik.github.io/old-man-sandbox/sorted-es6/test.html in debugger
2. Set breakpoint at webpack/sorted.js line 9 (`return na < nb ? -1 : na > nb ? 1 : 0;`)
3. Reload the page


src/actions/sources.js
```diff
@@ -135,6 +135,7 @@ function loadSourceMap(generatedSource) {
 
     await dispatch(loadSourceText(generatedSource));
     originalSources.forEach(async source => {
+      await dispatch(loadSourceText(source));
```
jasonLaster: why is this necessary? It could be a very expensive operation

src/utils/scopes.js
```diff
@@ -113,6 +122,233 @@ function getThisVariable(frame: any, path: string) {
...
+    const lastGeneratedScope = generatedScopes[generatedScopes.length - 1];
+    const isLastGeneratedScopeFn =
+      lastGeneratedScope && lastGeneratedScope.type === "function";
+    title =
+      isLastGeneratedScopeFn && lastGeneratedScope.function.displayName
+        ? simplifyDisplayName(lastGeneratedScope.function.displayName)
+        : L10N.getStr("anonymous");
+  } else {
+    title = L10N.getStr("scopes.block");
+  }
```
jasonLaster: perhaps a small helper for getting the scope title

```js
function getScopeTitle(...) {
  if (type === "function") {
    // FIXME Use original function name here
    const lastGeneratedScope = generatedScopes[generatedScopes.length - 1];
    const isLastGeneratedScopeFn =
      lastGeneratedScope && lastGeneratedScope.type === "function";
    title =
      isLastGeneratedScopeFn && lastGeneratedScope.function.displayName
        ? simplifyDisplayName(lastGeneratedScope.function.displayName)
        : L10N.getStr("anonymous");
  } else {
    title = L10N.getStr("scopes.block");
  }
}
```

src/utils/scopes.js
```diff
@@ -113,6 +122,233 @@ function getThisVariable(frame: any, path: string) {
...
+    generatedScopes.push(s);
+    s = s.parent;
+  }
+
+  const { result } = syntheticScopes.scopes.reduce(synthesizeScope, {
+    actor,
+    scopeIndex,
+    lastScopeIndex: syntheticScopes.scopes.length - 1,
+    generatedScopes,
+    foundGeneratedNames: Object.create(null),
```
jasonLaster: style nit:  `foundGeneratedNames: {}`

src/utils/scopes.js
```diff
@@ -113,6 +122,233 @@ function getThisVariable(frame: any, path: string) {
...
+      }
+    }
+
+    vars.push({
+      name,
+      generatedName,
+      path: `${key}/${generatedName}`,
+      contents: { value: { type: "undefined" } }
+    });
+  });
```
jasonLaster: can we move this to a small helper:

```js
function findOriginalBindings(...) {
   let vars = [];
   bindingsNames.forEach(name => {
     // Find binding name in the original source bindings
     const generatedScope = generatedScopes.find(
       gs => gs.sourceBindings && name in gs.sourceBindings
     );
     if (!generatedScope || !generatedScope.sourceBindings) {
       return;
     }
     // .. and map it to the generated name
     const generatedName = generatedScope.sourceBindings[name];
     // Skip if we already use the generated name
     if (generatedName && !foundGeneratedNames[generatedName]) {
       if (generatedScope.bindings.variables[generatedName]) {
         vars.push({
           name,
           generatedName,
           path: `${key}/${generatedName}`,
           contents: generatedScope.bindings.variables[generatedName]
         });
         foundGeneratedNames[generatedName] = true;
         return;
       }
 
       const arg = generatedScope.bindings.arguments.find(
         arg_ => arg_[generatedName]
       );
       if (arg) {
         vars.push({
           name,
           generatedName,
           path: `${key}/${generatedName}`,
           contents: arg[generatedName]
         });
         foundGeneratedNames[generatedName] = true;
         return;
       }
     }
 
     vars.push({
       name,
       generatedName,
       path: `${key}/${generatedName}`,
       contents: { value: { type: "undefined" } }
     });
   });
}
```

src/utils/scopes.js
```diff
@@ -113,6 +122,233 @@ function getThisVariable(frame: any, path: string) {
...
+  if (vars && vars.length) {
+    vars.sort((a, b) => a.name.localeCompare(b.name));
+    result.push({
+      name: title,
+      path: key,
+      contents: vars
+    });
+  }
+  return acc;
+}
```
jasonLaster: synthesize scope is complicated enough where it takes some time to see what it is trying to do. I think it would be nice to extract the "real" work so that the code flow is more readable.

```js
function synthesizeScope(
   acc: SynthesizeScopeContext,
   syntheticScope: SyntheticScope,
   index: number
  ): SynthesizeScopeContext {

  let bindings = findOriginalBindings(...)
  
  if (isLast) {
    bindings = [...bindings, ...findUnusedBindings(...)]
  }
  
  if (isFirst) {
    bindings = [...bindings, ...getSpecialVariables(...)]
    // ...
  }
  
  if (bindings.length > 0) {
    const title = getScopeTitle(...);
    result.push({
      name: title,
      path: key,
      contents: vars
    });
  }

  return acc
}
```

src/utils/scopes.js
```diff
@@ -113,6 +122,233 @@ function getThisVariable(frame: any, path: string) {
...
+  for (
+    let count = syntheticScopes.groupLength, s = scope;
+    count > 0 && s;
+    count--
+  ) {
+    generatedScopes.push(s);
+    s = s.parent;
+  }
+
+  const { result } = syntheticScopes.scopes.reduce(synthesizeScope, {
```
jasonLaster: I think the fact that synthesizeScope is mutating makes synthesize a bit more complicated... perhaps, we could have it return the bindings...

```js
  const { result } = syntheticScopes.scopes.reduce((bloby, syntheticScope, index) => {
    const bindings = getBindingsFromScope(bloby,..) 
    // it's tempting to update foundGeneratedNames here too so that `getBindingsFromScope` is pure, 
    // but not sure if that is needed... 
    if (bindings.length) {
      const title = getScopeTitle(...);
      blob.result.push({
        name: title,
        path: key,
        contents: vars
      });
    }
  }, blobOStuff)

```

src/utils/scopes.js
jasonLaster: It would be nice to create the synthetic scope bindings in the map-bindings space. This would be closer to the data structure we'd get from source-maps

* we can add the special vars in the debugger side

src/utils/pause.js
```diff
@@ -28,37 +30,25 @@ export function updateFrameLocations(
...
+    async getSourceMapsScopes(location) {
+      const astScopes: ?(SourceScope[]) = await getScopes(location);
+      const generatedScopes = await sourceMaps.getLocationScopes(
+        location,
+        astScopes
+      );
+      return generatedScopes;
+    },
+    async getOriginalSourceScopes(location) {
+      return getScopes(location);
```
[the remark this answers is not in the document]
yurydelendik: How can we track if we already loaded the source?

src/actions/pause/fetchScopes.js
```diff
@@ -18,12 +18,12 @@ export function fetchScopes() {
...
     }
 
-    const sourceRecord = getSource(
+    const generatedSourceRecord = getSource(
       getState(),
       frame.generatedLocation.sourceId
     );
 
-    if (sourceRecord.get("isWasm")) {
+    if (generatedSourceRecord.get("isWasm")) {
```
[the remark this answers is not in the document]
yurydelendik: https://github.com/devtools-html/debugger.html/pull/4637

src/utils/pause.js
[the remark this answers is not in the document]
jasonLaster: we can call `await loadSourceText` and if it is loaded it will come back quickly

ryanjduffy:
I ran a quick test with this and wanted to verify what I'm seeing.

ES6 Source:
```
function test () {
	let a = 1;
	if (a) {
	  const b = 2;
	}
}

document.addEventListener('click', test);
```

Transpiled source:
```
'use strict';

function test() {
	var a = 1;
	if (a) {
		var b = 2;
	}
}

document.addEventListener('click', test);

//# sourceMappingURL=scopes-compiled.js.map
```

When pausing on line 4, `const b = 2;`, both `a` and `b` show within the `test` scope and only `<this>` within the `if` block.

image

I expected `<this>` to be restricted to function scope and `b` to be within the block scope.

Thoughts?

jasonLaster:
@ryanjduffy could you "remix" the dbg example: http://dbg.glitch.me/

yurydelendik:
> maybe devtools-examples.

I would rather to see it as a complete example online with source maps. I wonder if babel spits out names.

jasonLaster:
i'll look for a codepen that has babel

yurydelendik:
> I expected <this> to be restricted to function scope and b to be within the block scope.

Based on debugger.html logic at https://github.com/devtools-html/debugger.html/blob/60c59811df6e82ab7831d8faab4cee23ec48bd63/src/utils/scopes.js#L173 we place `<this>` on the selected scope.

ryanjduffy:
Here's a live version of the sample: https://ryanjduffy.github.io/debugger-examples/scopes/

yurydelendik:
(The devtools/client/shared/source-map/worker.js needs to be updated to address the issue above, works in debugger.html now)

yurydelendik:
> Here's a live version of the sample: https://ryanjduffy.github.io/debugger-examples/scopes/

@ryanjduffy the names section is empty at https://ryanjduffy.github.io/debugger-examples/scopes/scopes-compiled.js.map , what did you use to generate the .map file?

ryanjduffy:
Babel cli

src/utils/pause/scopes/translateScope.js
```diff
@@ -0,0 +1,84 @@
...
+function getScopeTitle(type, scope) {
+  if (type === "function") {
+    return scope.function.displayName
+      ? simplifyDisplayName(scope.function.displayName)
+      : L10N.getStr("anonymous");
+  }
+  return L10N.getStr("scopes.block");
+}
+
+export function translateScope(
```
jasonLaster: does the name translate mean anything special?

would `getScope`, which is parallel to `getScopes` work?

Issue #4436, opened 2017-10-19
[Editor] Add some unit tests


jasonLaster:
We have simplified the Editor component a lot, which makes it a great candidate for some Jest unit tests.

I'd like to test the render and update lifecycle and make sure we'e appropriately setting the correct text and showing the correct line:

- [ ] flashLine
- [ ] setText

We can model these tests after other component unit tests :smile:

PR #4522, opened 2017-10-27
Add Units for Editor

James300:
Associated Issue: #4436

### Summary of Changes

* Add Unit Test for `SetText`
* Add Unit Tests for `SetText` while loading


src/components/Editor/tests/Editor.spec.js
```diff
@@ -38,4 +38,20 @@ describe("Editor", () => {
...
+    expect(component).toMatchSnapshot();
+  });
+
+  it("should highlight line", async () => {
+    const { component } = render();
+    component.setProps({
+      selectedLocation: { line: 1, column: 1 },
+      selectedFrame: { location: { line: 1, column: 1 } }
+    });
+    expect(component).toMatchSnapshot();
```
jasonLaster: i don't think the snapshots are showing the highlight line or the text... perhaps we need to also spy on codemirror or do something to make sure they render...

jasonLaster:
Such a small change, yet... so interesting

Issue #4506, opened 2017-10-27
Find in Files Shows "0 results" Before Search Executed


darkwing:
Nit: showing a search result count before a search is executed doesn't have value; we should hide the result count until a search is actually executed.

jasonLaster:
Here's a good place to start digging:

https://github.com/devtools-html/debugger.html/blob/master/src/components/ProjectSearch/TextSearch.js#L53-L60

PR #4525, opened 2017-10-29
Displays result count only after initiating search

lukaszsobek:
Associated Issue: #4506 

### Summary of Changes

* moved variables to state to be able to assign initial state and update
* update based on boolean updateCount and difference of current state from previous state

### Test Plan

tested by performing searches

http://g.recordit.co/bEQ4fkre8s.gif

### DIFF
--- a/src/components/ProjectSearch/TextSearch.js
+++ b/src/components/ProjectSearch/TextSearch.js
@@ -16,8 +16,7 @@ export default class TextSearch extends Component {
   constructor(props: Props) {
     super(props);
     this.state = {
-      inputValue: this.props.query || "",
-      resultCount: 0
+      inputValue: this.props.query || ""
     };
 
     this.focusedItem = null;
@@ -37,19 +36,6 @@ export default class TextSearch extends Component {
   componentWillUnmount() {
     const shortcuts = this.context.shortcuts;
     shortcuts.off("Enter", this.onEnterPress);
-  }
-
-  componentDidUpdate(prevProps, prevState) {
-    if (this.props.query === "") {
-      return;
-    }
-
-    const resultCount = this.getResultCount();
-    if (resultCount === prevState.resultCount) {
-      return;
-    }
-
-    this.setState({ resultCount });
   }
 
   selectMatchItem(matchItem) {
@@ -185,18 +171,17 @@ export default class TextSearch extends Component {
   }
 
   renderInput() {
+    const resultCount = this.getResultCount();
+
     return (
       <SearchInput
         query={this.state.inputValue}
-        count={this.state.resultCount}
+        count={resultCount}
         placeholder={L10N.getStr("projectTextSearch.placeholder")}
         size="big"
         summaryMsg={
           this.props.query !== ""
-            ? L10N.getFormatStr(
-                "sourceSearch.resultsSummary1",
-                this.state.resultCount
-              )
+            ? L10N.getFormatStr("sourceSearch.resultsSummary1", resultCount)
             : ""
         }
         onChange={e => this.inputOnChange(e)}
--- a/src/components/ProjectSearch/TextSearch.js
+++ b/src/components/ProjectSearch/TextSearch.js
@@ -16,7 +16,10 @@ export default class TextSearch extends Component {
   constructor(props: Props) {
     super(props);
     this.state = {
-      inputValue: this.props.query || ""
+      inputValue: this.props.query || "",
+      resultCount: -1,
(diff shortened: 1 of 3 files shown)


src/components/ProjectSearch/TextSearch.js
```diff
@@ -170,19 +194,13 @@ export default class TextSearch extends Component {
   }
 
   renderInput() {
-    const resultCount = this.getResultCount();
-    const summaryMsg = L10N.getFormatStr(
-      "sourceSearch.resultsSummary1",
-      resultCount
-    );
```
jasonLaster: could we keep summary msg here? it seems weird to store that?

src/components/ProjectSearch/TextSearch.js
```diff
@@ -38,6 +41,24 @@ export default class TextSearch extends Component {
...
+
+    if (resultCount === prevState.resultCount) {
+      return;
+    }
+
+    const summaryMsg = L10N.getFormatStr(
+      "sourceSearch.resultsSummary1",
+      resultCount
+    );
+    this.setState({ resultCount, summaryMsg });
```
jasonLaster: nice

src/components/ProjectSearch/TextSearch.js
```diff
@@ -70,6 +91,9 @@ export default class TextSearch extends Component {
     }
     this.focusedItem = null;
     this.props.searchSources(this.state.inputValue);
+
+    const updateCount = true;
```
jasonLaster: can we avoid storing this?

wldcordeiro: I think we could signal the need to update in a way other than the flag.

src/components/ProjectSearch/TextSearch.js
lukaszsobek: Something like this?
screenshot_2

jasonLaster: i still don't see the need for storing `summaryMsg`, why can't we calculate each time we render?

lukaszsobek: ah, I think because this would display the message before enter is pressed. Entering data into the field itself is not a sign of wanting to commit to a search and as such displaying the "0 Results message" does not seem to be the thing to do.

jasonLaster: hmm, can we have a check there to see that we are not currently searching?

lukaszsobek: I think we could put a function that would construct the output based on the value of `this.updateCount`, similar to what SearchBar.js does. Something along the lines of:
screenshot_3

src/components/ProjectSearch/TextSearch.js
```diff
@@ -83,6 +99,12 @@ export default class TextSearch extends Component {
     }
   }
 
+  buildSummaryMsg(updateCount, resultCount) {
+    return !updateCount
+      ? ""
+      : L10N.getFormatStr("sourceSearch.resultsSummary1", resultCount);
```
jasonLaster: sorry... i think i'm missing the point here:

* will this only show a value if updateCount is true? when is it false?
* is `updateCount` representing that we have searched? can we do this another way? like looking at whether there is a current search query string? or adding a boolean to the text-search reducer? 

what is this trying to do again? :)

lukaszsobek: Not quite sure I'm getting the direction. 

The idea behind it is to show a count only after the enter key has been pressed (which will cause `updateCount = true`). Since the code automatically adjusts the `resultCount` disregarding of whether enter was pressed or not it makes the variable unreliable as means of determining whether or not to show something - we have a 0 results state while searching and a 0 results state that should not be shown because the enter key has not yet been pressed.

src/components/ProjectSearch/TextSearch.js
```diff
@@ -38,6 +39,19 @@ export default class TextSearch extends Component {
...
+    if (this.props.query === "") {
+      return;
+    }
+
+    const resultCount = this.getResultCount();
+    if (resultCount === prevState.resultCount) {
+      return;
+    }
+
+    this.setState({ resultCount });
```
jasonLaster: do we need to store resultCount? why can't we use it in the render method... perhaps, `this.getResultCount` should return null if `this.props.query === ""` or we don't render the summary if `this.props.query === ""`

lukaszsobek:
Glad you do - good practice!

Issue #4024, opened 2017-09-14
[scopes] safely access properties


jasonLaster:
We want a safe way to access grip properties without causing a warning on the server.

This attempt should be fine, but some grips live beyond the pause lifetime. The best example is watch expressions.

```diff
diff --git a/src/actions/pause.js b/src/actions/pause.js
index ca92c09..feff145 100644
--- a/src/actions/pause.js
+++ b/src/actions/pause.js
@@ -243,7 +243,7 @@ export function loadObjectProperties(object: any) {
   return ({ dispatch, client, getState }: ThunkArgs) => {
     const objectId = object.actor || object.objectId;

-    if (getLoadedObject(getState(), objectId)) {
+    if (!getPause(getState()) || getLoadedObject(getState(), objectId)) {
       return;
     }
```

jasonLaster:
this was addressed by our waitForPaused work

PR #4538, opened 2017-10-30
Clear pause state on step action

jasonLaster:
Associated Issue: #4024 

### Summary of Changes

* will  also help address  some of the intermittents we see w/ wasm and other tests by removing the chance for unsafe calls

### Tests

* unit tests

### DIFF
--- a/src/actions/pause/commands.js
+++ b/src/actions/pause/commands.js
@@ -5,6 +5,7 @@ import { getPausedPosition } from "../../utils/pause";
 import { getNextStep } from "../../workers/parser";
 import { addHiddenBreakpoint } from "../breakpoints";
 import { features } from "../../utils/prefs";
+import { PROMISE } from "../../utils/redux/middleware/promise";
 
 import type { ThunkArgs } from "../types";
 type CommandType = "stepOver" | "stepIn" | "stepOut" | "resume";
@@ -18,12 +19,11 @@ type CommandType = "stepOver" | "stepIn" | "stepOut" | "resume";
  */
 export function command(type: CommandType) {
   return async ({ dispatch, client }: ThunkArgs) => {
-    // execute debugger thread command e.g. stepIn, stepOver
-    dispatch({ type: "COMMAND", value: { type } });
-
-    await client[type]();
-
-    dispatch({ type: "CLEAR_COMMAND" });
+    return dispatch({
+      type: "COMMAND",
+      command: type,
+      [PROMISE]: client[type]()
+    });
   };
 }
 
--- a/src/reducers/pause.js
+++ b/src/reducers/pause.js
@@ -12,6 +12,7 @@
 
 import { createSelector } from "reselect";
 import { prefs } from "../utils/prefs";
+import { isEmpty } from "lodash";
 
 import type { Action } from "../actions/types";
 
@@ -44,6 +45,7 @@ export const State = (): PauseState => ({
 const emptyPauseState = {
   pause: null,
   frames: null,
+  frameScopes: {},
   selectedFrameId: null,
   loadedObjects: {}
 };
@@ -222,6 +224,11 @@ export function getLoadedObject(state: OuterState, objectId: string) {
   return getLoadedObjects(state)[objectId];
 }
 
+export function hasLoadingObjects(state: OuterState) {
+  const objects = getLoadedObjects(state);
+  return Object.values(objects).some(isEmpty);
+}
+
 export function getObjectProperties(state: OuterState, parentId: string) {
   return getLoadedObjects(state).filter(obj => obj.parentId == parentId);
 }
@@ -242,8 +249,17 @@ export function getFrames(state: OuterState) {
   return state.pause.frames;
 }
 
-export function getFrameScopes(state: OuterState, frameId: string) {
+export function getFrameScope(state: OuterState, frameId: ?string) {
+  if (!frameId) {
+    return null;
+  }
+
   return state.pause.frameScopes[frameId];
+}
+
+export function getSelectedScope(state: OuterState) {
+  const frameId = getSelectedFrameId(state);
+  return getFrameScope(state, frameId);
 }
 
 export function getScopes(state: OuterState) {
--- a/src/reducers/pause.js
+++ b/src/reducers/pause.js
@@ -41,6 +41,13 @@ export const State = (): PauseState => ({
   command: ""
 });
 
+const emptyPauseState = {
+  pause: null,
+  frames: null,
+  selectedFrameId: null,
+  loadedObjects: {}
+};
+
 function update(state: PauseState = State(), action: Action): PauseState {
   switch (action.type) {
     case "PAUSED": {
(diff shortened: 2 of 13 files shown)


src/reducers/pause.js
```diff
@@ -80,12 +87,7 @@ function update(state: PauseState = State(), action: Action): PauseState {
         frameScopes: { ...state.frameScopes, [selectedFrameId]: scopes }
       };
     case "RESUME":
-      return Object.assign({}, state, {
-        pause: null,
-        frames: null,
-        selectedFrameId: null,
-        loadedObjects: {}
-      });
+      return { ...state, emptyPauseState };
```
jimblandy: Doesn't this create an object that looks like state, but with a new property named `emptyPauseState`?

src/actions/pause/commands.js
```diff
@@ -18,12 +19,11 @@ type CommandType = "stepOver" | "stepIn" | "stepOut" | "resume";
...
-    dispatch({ type: "COMMAND", value: { type } });
-
-    await client[type]();
-
-    dispatch({ type: "CLEAR_COMMAND" });
+    return dispatch({
+      type: "COMMAND",
+      command: type,
+      [PROMISE]: client[type]()
+    });
```
jimblandy: Is this neglecting to put a `status` property on this action, or is this not a COMMAND `PauseAction`?

jimblandy: I see; this communicates with the promise middleware.

src/reducers/pause.js
```diff
@@ -153,10 +155,9 @@ function update(state: PauseState = State(), action: Action): PauseState {
       });
 
     case "COMMAND":
-      return { ...state, command: action.value.type };
-
-    case "CLEAR_COMMAND":
-      return { ...state, command: "" };
+      return action.status === "start"
+        ? { ...state, ...emptyPauseState, command: action.command }
+        : { ...state, command: "" };
```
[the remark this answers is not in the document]
jimblandy: So what this is really doing is replacing the `COMMAND` / `CLEAR_COMMAND` pair with a single `COMMAND` action, and using the `AsyncStatus` attached by the promise middleware to discriminate between the two behaviors.

src/reducers/pause.js
```diff
@@ -165,7 +166,7 @@ function update(state: PauseState = State(), action: Action): PauseState {
       };
 
     case "NAVIGATE":
-      return { ...state, debuggeeUrl: action.url };
+      return { ...state, ...emptyPauseState, debuggeeUrl: action.url };
```
jimblandy: This seems like an unrelated change. Is it somehow necessitated by the changes to the `COMMAND` action?

src/test/mochitest/head.js
```diff
@@ -343,13 +344,22 @@ function isPaused(dbg) {
...
  */
 async function waitForPaused(dbg) {
-  // We want to make sure that we get both a real paused event and
-  // that the state is fully populated. The client may do some more
-  // work (call other client methods) before populating the state.
-  let loading = waitForDispatch(dbg, "LOAD_OBJECT_PROPERTIES");
-  await waitForThreadEvents(dbg, "paused");
-  await waitForState(dbg, state => isPaused(dbg));
-  await loading;
+  function scopeIsLoaded(dbg) {
```
jimblandy: Rather than digging around in the state and counting properties, would it be possible for, say, something in `src/reducers` to provide a predicate on a state that indicates whether the scopes are fully loaded? It seems to me that duplicating the knowledge of that structure's details here is asking for trouble.

src/test/mochitest/head.js
```diff
@@ -343,13 +344,22 @@ function isPaused(dbg) {
...
+    if (loadedObjects.length == 0) {
+      return false;
+    }
+
+    return Object.keys(loadedObjects[0]).length !== 0;;
+  }
+
+  await waitForState(
+    dbg,
+    state => isPaused(dbg) && scopeIsLoaded(dbg),
```
jimblandy: Is the `state` being passed to this function the same value retrieved by `dbg.getState()`? It seems like it would make sense for `scopeIsLoaded` to be a predicate on states.

src/test/mochitest/head.js
```diff
@@ -679,9 +688,9 @@ function togglePauseOnExceptions(
  * @return {Promise}
  * @static
  */
-function invokeInTab(fnc) {
+async function invokeInTab(fnc) {
   info(`Invoking function ${fnc} in tab`);
-  return ContentTask.spawn(gBrowser.selectedBrowser, fnc, function*(fnc) {
+  ContentTask.spawn(gBrowser.selectedBrowser, fnc, function*(fnc) {
```
jimblandy: Won't failing to return the promise here break `uncaughtException` and `caughtException` in `src/test/mochitest/browser_dbg-pause-exceptions.js`?

src/test/mochitest/browser_dbg-sourcemaps.js
```diff
@@ -78,13 +78,14 @@ add_task(async function() {
   assertPausedLocation(dbg);
 
   await stepIn(dbg);
-  // await waitForMappedScopes(dbg);
+  await waitForSelectedSource(dbg, "times2.js")
```
jimblandy: Is this somehow related to the change to the way the pause state is managed? If not, it should be a separate pull request.

src/reducers/pause.js
jasonLaster: Yep, good catch. Turns out that we can drop this entirely now that we no longer update the state when we are resumed

src/reducers/pause.js
jasonLaster: It is something that happened before, but this change makes it more explicit. It is helpful to do it here to prevent future regressions. Also simpler to test

src/test/mochitest/head.js
jasonLaster: 👍

src/test/mochitest/head.js
```diff
@@ -343,13 +344,13 @@ function isPaused(dbg) {
...
-  // work (call other client methods) before populating the state.
-  let loading = waitForDispatch(dbg, "LOAD_OBJECT_PROPERTIES");
-  await waitForThreadEvents(dbg, "paused");
-  await waitForState(dbg, state => isPaused(dbg));
-  await loading;
+  const { getSelectedScope, hasLoadingObjects } = dbg.selectors;
+
+  return waitForState(
+    dbg,
+    state => isPaused(dbg) && getSelectedScope(state) && !hasLoadingObjects(state),
```
jasonLaster: we're now checking to see if we have a fetched scope and that there aren't any pending object fetches which is pretty nice

src/test/mochitest/head.js
jasonLaster: we aren't using the promise, but i shouldn't have removed it either

src/reducers/pause.js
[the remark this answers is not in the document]
jasonLaster: yep

codehag:
This seems to have been split into smaller prs, should it be closed?

Issue #3878, opened 2017-09-04
Syntax highlighting is crazy for some JSX construct


julienw:
Here is the code that makes the syntax highlighting crazy:

```javascript
  render() {
    this._scheduleDraw();
    return (
      <div className={this.props.className}>
        <canvas
          className={classNames(
            `${this.props.className}Canvas`,
            'threadStackGraphCanvas'
          )}
          ref={ref => (this._canvas = ref)}
          onMouseUp={this._onMouseUp}
        />
      </div>
    );
  }
```

See how it renders:
![capture d ecran de 2017-09-04 17-59-00](https://user-images.githubusercontent.com/454175/30033296-c49c05e8-919a-11e7-85ed-fe135275e147.png)

jasonLaster:
It would be nice to see what is available w/ codemirror, maybe others have found a good solution?

wldcordeiro:
Maybe CodeMirror's latest version has fixed it? I don't see a note in the changelog about the JSX mode specifically. Also maybe worth reporting to [them](https://github.com/codemirror/CodeMirror) if it does seem broken.

julienw:
It was easy to actually try:
![screenshot-2017-10-30 codemirror](https://user-images.githubusercontent.com/454175/32163532-e2344aaa-bd5c-11e7-9958-dbd07fa49617.png)

Yep, it looks like Code Mirror's newer versions actually fix this.

julienw:
Though some of the colors are still weird: see the angle bracket in the ref's arrow function, or `onMouseUp` not having the right color, or the closing `div` tag.

But it's not as broken as it was.

julienw:
I added this example in this issue: https://github.com/codemirror/CodeMirror/issues/3745#issuecomment-340389826

julienw:
Ah, they answered that we should be in JSX mode, and this actually works in JSX mode (while JS mode still shows the same issue as initially shown). I don't know if there's any drawbacks using this mode for all JS code, I asked there.

jasonLaster:
Thanks @julienw we can be pretty smart about the mode if we parse the source text and see a JSX node

jasonLaster:
when was the JSX mode introduced? should we try to do this w/ 28? or jump and do it?

nyrosmith:
@jasonLaster jsx mode was introduced in [v5.11.0.](https://github.com/codemirror/CodeMirror/releases/tag/5.11.0) in Jan 2016

[...]

PR #4539, opened 2017-10-30
Improve JSX syntax highlighting

nyrosmith:
Associated Issue: #3878 

Using perf.html as an example also revealed a weakness in our logic to detect if it's a react component. So I extended  `utils/source.js` to also take `PureComponent` into account.

Syntax highlighting of JSX after this changes:
jsx


wldcordeiro:
One concern is that JSX is used by more than React so we might wish to be more agnostic but I like this solution so far.

nyrosmith:
@wldcordeiro of course it can be improved but I think it's a good start.

src/utils/source.js
```diff
@@ -9,6 +9,7 @@ import { isOriginalId } from "devtools-source-map";
 import { endTruncateStr } from "./utils";
 import { basename } from "../utils/path";
 import { parse as parseURL } from "url";
+import { isReactComponent } from "../workers/parser/frameworks";
```
jasonLaster: ooh, this call will have to be async because it requires parsing...

if we want to do this, we'll need to add it to another place, like on `loadSourceText`.
i'm not sure if we're already parsing 


```diff

+export function setSourceMetaData(sourceId: SourceId) {
+  return async ({ dispatch, getState }: ThunkArgs) => {
+    const sourceRecord = getSource(getState(), sourceId);
+    if (!sourceRecord) {
+      return;
+    }
+
+    const source = sourceRecord.toJS();
+    if (!source.text || source.isWasm || hasSymbols(getState(), source)) {
+      return;
+    }
+
+    const isReactComponent = await isReactComponent(source.get("id"));
+    const hasJSX = await hasJSX(source.get("id"));
+
+    // would update the source record...
+    dispatch({
+      type: "SOURCE_META_DATA",
+      isReactComponent,
+      hasJSX
+    });
+  };
+}
```

nyrosmith: @jasonLaster  I looked at https://github.com/devtools-html/debugger.html/blob/c5f08e7dd63c7fc0f238c70ca016aa474e831105/src/workers/parser/tests/framework.spec.js#L6
There is no indication that it is async. 
Also https://github.com/devtoolshtml/debugger.html/blob/master/src/workers/parser/frameworks.js doesn't look to me if it's async. What am I missing/overseeing?

jasonLaster: Everything in the worker is async. Also needs to be parsed

src/components/Editor/Breakpoint.js
```diff
@@ -52,7 +53,7 @@ class Breakpoint extends Component<Props> {
     const sourceId = selectedSource.get("id");
     const line = toEditorLine(sourceId, breakpoint.location.line);
 
-    showSourceText(editor, selectedSource.toJS());
+    showSourceText(editor, selectedSource.toJS(), metaData);
```
jasonLaster: @codehag do we need `showSourceText` after the lifecycle work?

src/reducers/ast.js
```diff
@@ -38,7 +38,8 @@ export type ASTState = {
   symbols: SymbolsMap,
   emptyLines: EmptyLinesMap,
   outOfScopeLocations: ?Array<AstLocation>,
-  preview: Preview
+  preview: Preview,
+  metaData: Object
```
jasonLaster: lets type it like this:

```js
  type SourceId = string
  export type SourceMetaData = { .. }
  export type MetaDataMap = Map<SourceId, SourceMetaData>;
```

src/reducers/ast.js
```diff
@@ -111,6 +113,10 @@ function update(
       return initialState();
     }
 
+    case "SET_METADATA": {
```
jasonLaster: perhaps `SET_SOURCE_METADATA` would be clearer

src/components/Editor/Breakpoint.js
codehag: I am not sure why this is in addBreakpoint to be honest. this shouldnt be taken care of here -- it should be done with the lifecycle update... in fact the same can be said about everything that is happening with the `editor.codemirror` here. this is outside of the scope of this pr... good catch though

codehag: ahh ok i understand better. the naming is a bit off -- this should probably be renderBreakpint

jasonLaster: Filed #4579 and we can come back

src/reducers/ast.js
```diff
@@ -167,4 +179,8 @@ export function getPreview(state: OuterState) {
   return state.ast.get("preview");
 }
 
+export function getMetaData(state: OuterState) {
```
jasonLaster: hmm, i think this should be `getSourceMetaData(state: OuterState, source: Source)` so that we scope it down.

src/actions/ast.js
```diff
@@ -29,6 +30,29 @@ const extraProps = {
   react: { displayName: "this._reactInternalInstance.getName()" }
 };
 
+export function setMetaData(sourceId: SourceId) {
```
jasonLaster: lets add a jest unit test for this next to `setSymbols`

src/reducers/ast.js
nyrosmith: What do we need source for in this selector?

src/reducers/ast.js
```diff
@@ -111,6 +119,10 @@ function update(
       return initialState();
     }
 
+    case "SET_SOURCE_METADATA": {
+      return state.set("metaData", action.metaData);
```
jasonLaster: ahh, I'd change this to look like empty lines: `return state.setIn(["emptyLines", source.id], emptyLines);`

nyrosmith:
@jasonLaster Changed to your suggestions.
I also refactored to have unified naming and added some tests

jasonLaster:
A little bit of context on this PR. It turns out that codemirror has a great "JSX" mode, which is better than "JS" for showing... JSX. 

This PR, takes advantage of @James300's isReactComponent function in the parser, to say "if the file has a component, lets use the JSX" mode. 

The question is where should we store the "isReactComponent":

1. a new `Ast.sourceMetaData` redux map
2. a new field in the `Sources.sources` redux map
3. a new field in `symbols`

The argument for (1) and (2) is that we will want to know if a source is special, i.e. a component, route, or any other special type of framework file. (1) is a normal form, (2) is denormalized.

The argument for (3) is that symbols is already storing lots of meta data about sources, such as all the functions, variables, ..., why not also store meta data about the source: "is a react component"....

src/actions/sources/loadSourceText.js
```diff
@@ -45,5 +45,6 @@ export function loadSourceText(source: Source) {
     await setSource(newSource);
     await dispatch(setSymbols(source.id));
     await dispatch(setEmptyLines(source.id));
+    await dispatch(setSourceMetaData(source.id));
```
codehag: @jasonLaster, I think we should keep this like this for now -- it looks fine and we are doing the same thing anyway with setSymbols and setEmptyLines; but we should eventually revisit this, and ask some hard questions. Are these four dispatches truly independent, or are they all concerned with the same piece of data? If we step away from the implementation of loadSourceText; what is loadSourceText doing -- and when is it finished? what is the action to an outside observer (and not to someone who knows how the inside works)?  

I would say this is the best approach for now, and we can change it later on, when we refactor this in earnest and ask these questions. for more information about redux antipatterns: https://github.com/reactjs/redux/issues/857; https://stackoverflow.com/questions/33220776/should-actions-like-showing-hiding-loading-screens-be-handled-by-reducers-to-rel/33226443#33226443

jasonLaster: Yep... this is definitely an interesting place to investigate. A bit of context:

1. these actions could potentially be slow, so some of the dispatches are staggered to get the data when it's ready
2. some of this is "do the simplest thing" because when we created `actions/ast` it wasn't clear if AST was a separate thing, or if `setPreview` should live w/ pause and `setSymbols` should live with sources...

would love to revisit this too :)

nyrosmith:
@julianw why does codemirror have different modes for js & jsx? What are the implications if we always use jsx mode?
I thought it's safer to only use jsx mode when needed but I'll stay on topic and do further investigation

julienw:
[They say there should be no drawback](https://github.com/codemirror/CodeMirror/issues/3745#issuecomment-340395138). But because they say it doesn't mean it's true ;)

I don't know why they have different modes -- I suppose this is historical.
One thing that comes to my mind is for scripts inlined in an HTML page -- maybe here JS is better ? What happens for JSX code in JS code in an inline script ? :) This is of course unusual but not impossible.

PR #4540, opened 2017-10-30
Add timing data

jasonLaster:
screen shot 2017-10-30 at 5 17 15 pm

We can add this to the bottom of mochitests, we can also use it in other ways too :)


codehag:
i updated this to fix the linting error and to make the function names consistent. it looks good from my side, @jasonLaster you can merge if you approve!

Issue #4488, opened 2017-10-25
[Search] Quick Open follow up


jasonLaster:
#### Next Steps
- [ ] jump to call stack frame
- [ ] jump to breakpoint
- [x] style matched fuzzy characters
- [x] show indicator if symbols are still being parsed

#### Styling
- [ ] style focused result item
- [ ] style modal border shadow
- [x] style line numbers (i think we should drop the colon)
- [ ] style the input bar (i think it's a bit small)

![](http://g.recordit.co/RLOuU9WjeQ.gif)

#### style matched fuzzy characters (@princiya)

I love how atom styles matched characters. It might be nice to experiment w/ [downshift][d] and see if it could help us. Otherwise, we could do something what [fuzzyaldrin][f] gives us. Or, do something custom like what [projectSearch][p] does

![Screen Shot 2017-10-25 at 10.03.17 AM.png](https://shipusercontent.com/0301ef58052672b38cbb518275aeeacc/Screen%20Shot%202017-10-25%20at%2010.03.17%20AM.png)

[d]:https://github.com/paypal/downshift
[p]:https://github.com/wldcordeiro/debugger.html/blob/master/src/components/ProjectSearch/textSearch/utils/highlight.js#L17
[f]:https://github.com/jeancroy/fuzz-aldrin-plus

princiya:
@jasonLaster at this point, to what I understand, is this an issue to experiment with the below possibilities?
- downshift
- fuzzyaldrin
- projectSearch

wldcordeiro:
@princiya We're using fuzz-aldrin-plus already. I don't think we have a serious need to bring in something like downshift as it's just another dependency when we've built a decent solution already. As for projectSearch? I'm not sure what you mean by that.

jasonLaster:
yeah - lets build the simple thing first for styling the matched chars... making them blue :) or something like that. I think what we did for project search is pretty close... The basic algorithm for finding the chars for a query like "foo" is to take the match result string and split it up into matching and not-matching spans. so if the result is` "af/my/o/o.js",` => `"a", "f","/my/","o","/","o",".js"`

princiya:
ok, that sounds good. so I shall focus on the styling part for now.

jasonLaster:
yep

wldcordeiro:
Looks like the action items from #4431 got moved but this comment from @violasong is useful for anyone following up on the shadow fix.

[...]

PR #4541, opened 2017-10-31
Style line numbers in search results

princiya:
Associated Issue: #4488 

### Summary of Changes

* `:` is removed from the line numbers in search results

### Before

screen shot 2017-10-31 at 13 28 09

### After

screen shot 2017-10-31 at 13 25 56

### DIFF
--- a/src/test/mochitest/browser_dbg-quick-open.js
+++ b/src/test/mochitest/browser_dbg-quick-open.js
@@ -99,7 +99,7 @@ add_task(async function() {
   quickOpen(dbg, "#");
   is(resultCount(dbg), 1, "one variable result");
   const results = findAllElements(dbg, "resultItems");
-  results.forEach(result => is(result.textContent, "x:13"));
+  results.forEach(result => is(result.textContent, "x13"));
   await waitToClose(dbg);
 
   info("Testing goto line:column");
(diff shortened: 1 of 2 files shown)


princiya:
I have been trying to understand the file structure for the `search` component.

And, the `:` is coming from [here](https://github.com/devtools-html/debugger.html/blob/master/src/utils/quick-open.js#L74).

What is the best way to drop this in `ResultList`?
- Remove it from `utils/quick-open.js`?
  - There might be other dependencies involved here.
  - In other words, what is the motivation to add `:` in this file?
- Parse `subtitle` to display only the number without the `:` [here](https://github.com/devtools-html/debugger.html/blob/master/src/components/shared/ResultList.js#L45)?

jasonLaster:
deferring to @wldcordeiro

wldcordeiro:
I like the change. It looks nice! The CI failure is legit but it's an easy fix. 

Change this line's assertion https://github.com/devtools-html/debugger.html/blob/master/src/test/mochitest/browser_dbg-quick-open.js#L102

```
// from
results.forEach(result => is(result.textContent, "x:13"));
// to
results.forEach(result => is(result.textContent, "x13"));
```

wldcordeiro:
Something to note. Atom formats their results for symbols like this, perhaps we can do this. Though it'd be a bit more of a change.
```
symbolName
Line 13
```

screen shot 2017-11-01 at 9 21 50 am

jasonLaster:
Hmm, i like that wellington, but lets hold off on that for now

VS Code is more compact and i think we should stick to one thing at a time

PR #4550, opened 2017-11-01
Refactor map scopes to fetch scopes

jasonLaster:
### Summary of Changes

A refactor based on talking w/ @jimblandy about how scopes are fetched from the server. I realized that map scopes does more than advertised


codehag:
Fixed merge conflict and naming error

PR #4554, opened 2017-11-01
Refactor command action to use PROMISE

jasonLaster:
### Summary of Changes

Simplifies command so that it uses the promise middleware. Depends on the prior PR to finish first.

### DIFF
--- a/src/reducers/pause.js
+++ b/src/reducers/pause.js
@@ -12,6 +12,7 @@
 
 import { createSelector } from "reselect";
 import { prefs } from "../utils/prefs";
+import { isEmpty } from "lodash";
 
 import type { Action } from "../actions/types";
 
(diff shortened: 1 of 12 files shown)


src/test/mochitest/head.js
```diff
@@ -343,13 +343,13 @@ function isPaused(dbg) {
...
-  // We want to make sure that we get both a real paused event and
-  // that the state is fully populated. The client may do some more
-  // work (call other client methods) before populating the state.
-  let loading = waitForDispatch(dbg, "LOAD_OBJECT_PROPERTIES");
-  await waitForThreadEvents(dbg, "paused");
-  await waitForState(dbg, state => isPaused(dbg));
-  await loading;
+  const { getSelectedScope, hasLoadingObjects } = dbg.selectors;
+
+  return waitForState(
```
codehag: much easier to read. One question, is `isPaused` equivalent to the `waitForThreadEvents`?

jasonLaster: isPaused, is not the same as `waitForThreadEvents`, but I intentionally dropped `waitForThreadEvents(dbg, "paused");`

* waiting for an event means we have to start listening before we do an action otherwise we can miss it
* it is sufficient to check the pause state to see if we're paused

src/reducers/pause.js
```diff
@@ -223,6 +223,11 @@ export function getLoadedObject(state: OuterState, objectId: string) {
   return getLoadedObjects(state)[objectId];
 }
 
+export function hasLoadingObjects(state: OuterState) {
```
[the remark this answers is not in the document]
jasonLaster: ` Object.values(objects).none(isEmpty);` would be `hasLoadedObjects`...

I'm okay doing either, don't really care

PR #4557, opened 2017-11-01
Optimistically clear pause state

jasonLaster:


src/reducers/pause.js
```diff
@@ -38,9 +40,18 @@ export const State = (): PauseState => ({
   shouldPauseOnExceptions: prefs.pauseOnExceptions,
   shouldIgnoreCaughtExceptions: prefs.ignoreCaughtExceptions,
   debuggeeUrl: "",
-  command: ""
+  command: "",
+  history: []
```
codehag: why do we want to keep a history of this?

jasonLaster: this is useful for showing the pause trail, and we need the previous pause reason now that we're clearing pause state optimistically

codehag: since we are always pushing here, and never clearing the history, this could potentially grow quite large, especially when stepping repeatedly. we seem to be only using the [0] element in the history. Lets be a bit more conservative for now and just update this field rather than keeping everything

PR #4561, opened 2017-11-02
Breakpoint Dropdown

GarethSharpe:
Associated Issue: #4410 

### Summary of Changes

* Added icon prop to dropdown (every dropdown requires icon input)
* Breakpoint dropdown (BD) toggle .svg to "+"
* BD returns dropdown of breakpoint actions
1. Pause on next statement 
2. Pause on exceptions
3. Pause on uncaught exceptions
4. Ignore exceptions

Test plan:

- [x] Clicking '+' in the breakpoints pane toggles the BD
- [x] Clicking an action causes .svg image highlight
- [x] Clicking an action also causes '+' image highlight
- [x] Selected action executes on exception (or exception is ignored)
- [x] Resuming program execution allows program to continue until next exception
- [x] Clicking '+' again (without selecting an action) closes the BD

TODO:
- [x] Adjust dropdown width to fit parent
- [x] Remove dropdown scroll-bars
- [x] Update` L10N.getStr` for each dropdown option
- [x] Add menu items (with corresponding images)
- [x] Add menu item implementation
- [x] Add hover animation
- [x] Change `Svg` implementation of `dbgButton` to `img` (to accommodate mask)
- [x] '+' icon highlights when in execution state (when execption are not ignored)
- [ ] Remove `CommandBar` options as needed


assets/panel/prefs.js
```diff
@@ -44,3 +44,4 @@ pref("devtools.debugger.features.shortcuts", true);
 pref("devtools.debugger.project-directory-root", "");
 pref("devtools.debugger.features.root", false);
 pref("devtools.debugger.features.column-breakpoints", false);
+pref("devtools.debugger.features.dropdown", false);
```
jasonLaster: lets be more specific :) `devtools.debugger.features.breakpoints-dropdown`

src/components/shared/Dropdown.js
```diff
@@ -49,7 +50,7 @@ class Dropdown extends Component<Props, State> {
   renderButton() {
     return (
       <button className="dropdown-button" onClick={this.toggleDropdown}>
-        »
+        <Svg name="plus" />
```
jasonLaster: lets not forget about the renderDropdown prop

src/components/SecondaryPanes/index.js
```diff
@@ -107,7 +108,28 @@ class SecondaryPanes extends Component<Props> {
         : L10N.getStr("breakpoints.disable")
     };
 
-    return <input {...inputProps} />;
+    if (!features.breakpointsDropdown) {
+      return;
```
jasonLaster: if the feature is not enabled we still want ` <input {...inputProps} />;` right?

jasonLaster:
Noticed a couple things for the next iteration:

1. clicking closes the pane
2. the scroll bars are present 
3. the width is quite narrow



The big thing though (and reason why the test is failing) is that the toggle all bps button is missing when the pref is off

you can toggle it like this: `getGlobalsForTesting().features.breakpointsDropdown = false`

src/components/SecondaryPanes/index.js
GarethSharpe: Yes good catch. I must have forgot to put that back.

GarethSharpe:
@jasonLaster Thanks for the review. I'm looking to implement the `getGlobalsForTesting().features.breakpointsDropdown = false` part but am unsure as to where to add this. 😢

jasonLaster:
no worries. you'll want to do something like this

```
if (!features.breakpointsDropdown) {
  return <input ...>
}
```

i use this in the console to flip the flag
```js
getGlobalsForTesting().features.breakpointsDropdown = false
getGlobalsForTesting().features.breakpointsDropdown = true
```

GarethSharpe:
😃

src/components/SecondaryPanes/index.js
```diff
@@ -107,7 +108,28 @@ class SecondaryPanes extends Component<Props> {
...
+        {this.renderBreakpointsDropdown()}
+        <input {...inputProps} />
+      </div>
+    );
+  }
+
+  renderBreakpointsDropdown() {
+    const Panel = (
+      <ul>
+        <li>Pause on Next Statement</li>
```
codehag: these should be i18n texts (for translation)

src/components/SecondaryPanes/SecondaryPanes.css
```diff
@@ -42,3 +42,13 @@
 .theme-dark .secondary-panes .accordion .arrow svg {
   fill: var(--theme-content-color3);
 }
+
+.secondary-panes .breakpoints-buttons {
+  display: flex;
```
codehag: do we need flex here? could we achieve the same with floats?

src/components/SecondaryPanes/index.js
jasonLaster: https://github.com/devtools-html/debugger.html/blob/master/src/components/SecondaryPanes/CommandBar.js#L215

GarethSharpe:
I was able to put in the action functionality. Unsure as to what you would like from the first element of the drop down. I am currently using L10N.getStr(), but I think I will have to change the contents of those eventually. They are the same strings that are used for the tool tips of the command bar buttons (so they are a little long). There is no highlighting the currently selected option so I will put that in the TODO. Haven't yet looked at how to stop the pane from opening/collapsing at every touch of the button.

src/components/SecondaryPanes/index.js
```diff
@@ -107,7 +179,105 @@ class SecondaryPanes extends Component<Props> {
...
+          {" "}
+          {this.renderIgnoreExceptions()}
+          {L10N.getStr("ignoreExceptions")}
+        </li>
+      </ul>
+    );
+
+    return (
+      <Dropdown class="dropdown" panel={Panel} icon={<Svg name="plus" />} />
+    );
```
jasonLaster: Yep... we'll need to move this into it's own component now :) it's too damn big

`SecondaryPanes/BreakdpointsDropdown`

```js
export default function renderBreakpointsDropdown() {
  // stuff
  return <Dropdown class="dropdown" panel={Panel} icon={<Svg name="plus" />} />
}
```

src/components/SecondaryPanes/index.js
```diff
@@ -107,7 +179,105 @@ class SecondaryPanes extends Component<Props> {
...
+        <li> {this.renderPauseButton()} Pause on Next Statement</li>
+        <li>
+          {" "}
+          {this.renderPauseOnUncaughtExceptions()}
+          {L10N.getStr("pauseOnUncaughtExceptions")})
+        </li>
+        <li>
+          {" "}
+          {this.renderPauseOnExceptions()}
+          {L10N.getStr("pauseOnExceptions")}
```
jasonLaster: these strings are good for the tooltip, but you're right that it's too long. 
You can easily go into `assets/panel/properties` and add new shorter strings there too. We can add a `tooltip` suffix to the older ones

pauseOnExceptions=the new shorter oner
pauseOnExceptions.tooltip=the old long one

src/components/SecondaryPanes/index.js
```diff
@@ -107,7 +179,105 @@ class SecondaryPanes extends Component<Props> {
...
+    const Panel = (
+      <ul>
+        <li> {this.renderPauseButton()} Pause on Next Statement</li>
+        <li>
+          {" "}
+          {this.renderPauseOnUncaughtExceptions()}
+          {L10N.getStr("pauseOnUncaughtExceptions")})
+        </li>
+        <li>
+          {" "}
```
jasonLaster: you're cheating :)

Add some CSS.

src/components/SecondaryPanes/index.js
```diff
@@ -107,7 +179,105 @@ class SecondaryPanes extends Component<Props> {
...
+  renderBreakpointsDropdown() {
+    const Panel = (
+      <ul>
+        <li> {this.renderPauseButton()} Pause on Next Statement</li>
+        <li>
+          {" "}
+          {this.renderPauseOnUncaughtExceptions()}
+          {L10N.getStr("pauseOnUncaughtExceptions")})
+        </li>
+        <li>
```
jasonLaster: you'll want to make these also clickable... I think you will want to change this up so the li has the onClick `<li onClick={actions.pauseOnExceptions(true, true)}` and the button is just an `` element

assets/panel/debugger.properties
```diff
@@ -47,6 +47,10 @@ expandPanes=Expand panes
 # button when the debugger is in a running state.
 pauseButtonTooltip=Pause %S
 
+# LOCALIZATION NOTE (pauseButton): The label that is displayed for the dropdown pause
+# list item when the debugger is in a running state, excluding the shortcut key.
```
flodolo: Can you explain what "excluding the shortcut key" means in all these localization comments?

assets/panel/debugger.properties
```diff
@@ -55,6 +59,10 @@ pausePendingButtonTooltip=Waiting for next execution
 # button when the debugger is in a paused state.
 resumeButtonTooltip=Resume %S
 
+# LOCALIZATION NOTE (resumeButtonTooltip): The label that is displayed for the dropdown
```
flodolo: Please double check that the string ID referenced in the comment matches the actual string (it doesn't here).

assets/panel/debugger.properties
```diff
@@ -539,15 +547,27 @@ sourceSearch.noResults2=No results found
 
 # LOCALIZATION NOTE (ignoreExceptions): The pause on exceptions button tooltip
```
flodolo: You need to update the string ID here.

assets/panel/debugger.properties
```diff
@@ -539,15 +547,27 @@ sourceSearch.noResults2=No results found
...
 # when the debugger will not pause on exceptions.
-ignoreExceptions=Ignore exceptions. Click to pause on uncaught exceptions
+ignoreExceptionsTooltip=Ignore exceptions. Click to pause on uncaught exceptions
+
+# LOCALIZATION NOTE (ignoreExceptions): The pause on exceptions button description
+# when the debugger will not pause on exceptions.
+ignoreExceptions=Ignore exceptions
```
flodolo: Is the case (sentence case) correct for a button? Same question for other strings used as button labels later in the PR

assets/panel/debugger.properties
```diff
@@ -539,15 +547,27 @@ sourceSearch.noResults2=No results found
...
 
 # LOCALIZATION NOTE (pauseOnUncaughtExceptions): The pause on exceptions button
```
flodolo: String ID.

src/components/SecondaryPanes/BreakpointsDropdown.js
```diff
@@ -0,0 +1,97 @@
+import React from "react";
+import Svg from "../shared/Svg";
+import Dropdown from "../shared/Dropdown";
+
+export default function renderBreakpointsDropdown(_this, actions) {
+  function debugBtn(onClick, type, className, tooltip) {
```
jasonLaster: pull this out so it's outside of this function:
```js
function debugBtn(onClick, type, className, tooltip) {
}

 export default function renderBreakpointsDropdown(_this, actions) {
}
```

src/components/SecondaryPanes/BreakpointsDropdown.js
```diff
@@ -0,0 +1,97 @@
...
+        className={`${type} ${className}`}
+        key={type}
+        title={tooltip}
+      >
+        <Svg name={type} title={tooltip} aria-label={tooltip} />
+      </button>
+    );
+  }
+
+  function renderPauseButton() {
```
jasonLaster: same here

src/components/SecondaryPanes/BreakpointsDropdown.js
```diff
@@ -0,0 +1,97 @@
...
+  function renderPauseOnUncaughtExceptions() {
+    return debugBtn(
+      () => actions.pauseOnExceptions(true, false),
+      "pause-exceptions",
+      "uncaught enabled",
+      L10N.getStr("pauseOnUncaughtExceptionsTooltip")
+    );
+  }
+
+  function renderIgnoreExceptions() {
```
jasonLaster: all of these functions should be pulled out actually :)

src/components/SecondaryPanes/BreakpointsDropdown.js
```diff
@@ -0,0 +1,97 @@
...
+  const Panel = (
+    <ul>
+      <li onClick={renderPauseButton}>{L10N.getStr("pauseButton")}</li>
+      <li onClick={renderPauseOnUncaughtExceptions}>
+        {L10N.getStr("pauseOnUncaughtExceptions")}
+      </li>
+      <li onClick={renderPauseOnExceptions}>
+        {L10N.getStr("pauseOnExceptions")}
+      </li>
+      <li onClick={renderIgnoreExceptions}>
```
jasonLaster: why is this in an onClick?

`onClick={pauseOnUncaughtExceptions()}`

src/components/SecondaryPanes/BreakpointsDropdown.js
```diff
@@ -0,0 +1,72 @@
...
+    <button className="button-spacer">
+      <Svg name={type} title={tooltip} aria-label={tooltip} />
+    </button>
+  );
+}
+
+function renderPause(pauseData, isWaitingOnBreak) {
+  if (pauseData) {
+    return debugBtn("resume", L10N.getFormatStr("resumeButton"));
+  }
```
jasonLaster: we don't need the resume button here... because we're just creating a new breakpoint

src/components/SecondaryPanes/BreakpointsDropdown.js
```diff
@@ -0,0 +1,72 @@
...
+
+  if (isWaitingOnBreak) {
+    return debugBtn("pause", L10N.getStr("pausePendingButtonTooltip"));
+  }
```
jasonLaster: we also don't this one...

src/components/SecondaryPanes/BreakpointsDropdown.js
```diff
@@ -0,0 +1,72 @@
...
+
+  return debugBtn("pause", L10N.getFormatStr("pauseButtonItem"));
```
jasonLaster: we just need to be able to add a "pause on next statement" button

assets/panel/debugger.properties
```diff
@@ -537,17 +545,29 @@ sourceSearch.search=Search sources…
...
 
-# LOCALIZATION NOTE (ignoreExceptions): The pause on exceptions button tooltip
+# LOCALIZATION NOTE (ignoreExceptionsTooltip): The pause on exceptions button tooltip
+# when the debugger will not pause on exceptions.
+ignoreExceptionsTooltip=Ignore exceptions. Click to pause on uncaught exceptions
+
+# LOCALIZATION NOTE (ignoreExceptionsItem): The pause on exceptions button description
 # when the debugger will not pause on exceptions.
-ignoreExceptions=Ignore exceptions. Click to pause on uncaught exceptions
+ignoreExceptionsItem=Ignore exceptions
```
flodolo: The question about case still stands: is this supposed to use title case (like `pauseButtonItem`) or sentence case? I don't have visibility on the feature, so it might be indeed expected.

assets/panel/debugger.properties
```diff
@@ -537,17 +545,29 @@ sourceSearch.search=Search sources…
...
 
 # LOCALIZATION NOTE (pauseOnUncaughtExceptions): The pause on exceptions button
 # tooltip when the debugger will pause on uncaught exceptions.
 pauseOnUncaughtExceptions=Pause on uncaught exceptions. Click to pause on all exceptions
 
-# LOCALIZATION NOTE (pauseOnExceptions): The pause on exceptions button tooltip
+# LOCALIZATION NOTE (pauseOnUncaughtExceptionsItem): The pause on exceptions dropdown 
+# item shown when a user is adding a new breakpoint.
+pauseOnUncaughtExceptionsItem=Pause on uncaught exceptions
```
flodolo: Same question about title case vs sentence case.

assets/panel/debugger.properties
```diff
@@ -537,17 +545,29 @@ sourceSearch.search=Search sources…
...
+
+# LOCALIZATION NOTE (pauseOnExceptionsTooltip): The pause on exceptions button tooltip
+# when the debugger will pause on all exceptions.
+pauseOnExceptionsTooltip=Pause on all exceptions. Click to ignore exceptions
+
+# LOCALIZATION NOTE (pauseOnExceptionsItem): The pause on exceptions button description
 # when the debugger will pause on all exceptions.
-pauseOnExceptions=Pause on all exceptions. Click to ignore exceptions
+pauseOnExceptionsItem=Pause on all exceptions
```
flodolo: Same question about title case vs sentence case.

assets/panel/debugger.properties
jasonLaster: I think sentence case is fine. It's what we're using for other lists:

flodolo: So, pauseButtonItem needs fixing instead ("Pause on Next Statement", title case)?

GarethSharpe:
Ran into an issue with the coloring. I can access the image color fine from css, however the image itself doesn't update. The inspector displays the following error:
error

As you can see, the color of the img has been changed to be highlighted, but this change is not reflected in the actual image. I'm unable to decipher the error. Any help on this would be appreciated.

jasonLaster:
Hey man, i sat down and made some tweaks to your branch

- dropped debugBtn
- copied the commandbar.css and made changes to it in your css file

the UI still needs work, but i think it's getting closer

- the tooltips/titles, commands could be off... i guessed
- the css still needs work

but we're close. lets get this into a good state by monday

GarethSharpe:
dropdown

GarethSharpe:
dropdown

src/components/SecondaryPanes/BreakpointsDropdown.js
```diff
@@ -0,0 +1,133 @@
+import React from "react";
+import Svg from "../shared/Svg";
+import Dropdown from "../shared/Dropdown";
+import classnames from "classnames";
+
+import "./BreakpointsDropdown.css";
+
+function renderPause(isWaitingOnBreak) {
+  if (isWaitingOnBreak) {
+    return (
```
jasonLaster: we can simplify this:

```js
     <div className={classnames("pause-next", {inactive: isWaitingOnBreak, active: !isWaitingOnBreak})} >
         <img className="pause-next" />
         <span className="icon-spacer">{L10N.getStr("pauseButtonItem")}</span>
     </div>
```

jasonLaster:
okay... this is a complicated UI. I should have made a spec document beforehand, but i didn't realize all the edge cases till you shared your last version.

GarethSharpe:
A few questions about the UI:
- If `Pause on all exceptions' is selected, should `Pause on uncaught exceptions' also be selected since and uncaught exceptions is an element of all exceptions? 
- Should it be impossible to disable `Pause on next`? I thought perhaps not since any selection other than `Pause on next` does not necessarily mean that it is guaranteed to break on the next statement.

jasonLaster:
Good questions:

If Pause on all exceptions' is selected, shouldPause on uncaught exceptions' also be selected since and uncaught exceptions is an element of all exceptions?

> i see your point, but I think that makes it a bit more complicated.

Should it be impossible to disable Pause on next? I thought perhaps not since any selection other than Pause on next does not necessarily mean that it is guaranteed to break on the next statement.

> i don't follow. i'm pretty sure that once you select PON then it's impossible to disable it... basically you click it the, you're going to have to pause when you click in the app...

GarethSharpe:
Alright - I'll stick with having them be separate. As for the PON yes I believe your right. The only way to really disable it is to click the 'play' button in which case we revert back to ignore exceptions. Thanks. I'll finish this up right now.

GarethSharpe:
Just needs border-top to divide the sections.

PR #4565, opened 2017-11-02
Validate scripts

jasonLaster:
### Summary of Changes

switches to validating scripts as opposed to expressions to accomodate new mapped expressions


codehag:
looks good

Issue #3390, opened 2017-07-20
[Scopes]  does not show function's return value


tromey:
The old debugger would show the function's return value on stepping out, like this:

![screenshot from 2017-07-20 10-21-00](https://user-images.githubusercontent.com/1557670/28428020-3e0579f8-6d35-11e7-9d46-ba84f7e969c1.png)

I don't see this in the new debugger.

I just made a simple function and then repeatedly clicked the step over button.  Eventually it stops at the end of the function and shows the return value.

jasonLaster:
You can test it out w/ this test page locally: http://localhost:8000/integration/examples/doc-return-values.html

tromey:
I can reproduce it with the booleans example without having to clone the repository: https://devtools-html.github.io/debugger-examples/examples/booleans/

tromey:
To reproduce, open that page, put a breakpoint in the `makeBooleans` function, then click the button.  Then click "step over" repeatedly.  It will double stop at the "}" (which is what I was looking at when I found this); the second stop ought to show the return value.

bomsy:
i tried to reproduce this,  i get this

![](http://g.recordit.co/o8UNOp6Mq5.gif)

Looks like it works now. Or am i misunderstanding something?

tromey:
That looks correct to me.

bomsy:
Cool thanks! this looks like its fixed, will do a bit more testing and close this

codehag:
I think I found what is happening. the last bracket requires that you step over twice, it doesnt show the return value on the first step.

I think we don't want to have two steps at the end of the function. any ideas on how to avoid that?

yurydelendik:
> I think we don't want to have two steps at the end of the function. any ideas on how to avoid that?

So Debugger API shows return value at [Debugger.Frame::onPop](https://developer.mozilla.org/en-US/docs/Tools/Debugger-API/Debugger.Frame) -- you cannot ignore this one. It comes after last step, which can be somehow ignored, but it is hard. A frame has location (offset) property, but it does not give you much.

tromey:
> I think we don't want to have two steps at the end of the function. any ideas on how to avoid that?

This is what I was looking into when I found that I couldn't see the <return> at all.

I'm not sure whether I'll implement it or not, but the basic idea is to add a new debugger API so the server can tell whether it's stopped at a return opcode, and, if so, keep going until the frame pop.

jasonLaster:
@tromey That would be nice.

[...]

PR #4569, opened 2017-11-02
Fix display of return value / thrown exception in scopes

jimblandy:
Associated Issue: #3390

### Summary of Changes

* **[`scopes.spec.js`]: Don't supply scope in `pauseData`.**

    With [Bug 1370923], frames delivered to the new debugger no longer have `scope`
    properties. This commit changes the mock pause data to omit the scope property.
    However, since the tests don't actually check for `<return>` or `<throw>`
    pseudovars, nothing fails --- yet.

* **[`getScopes`]: Fix `<return>` and `<exception>` pseudo-vars.**

    As of [Bug 1370923], frames in "paused" packets from the server no longer include
    an environment form. This broke [`getScopes`], which assumed it could recognize the
    scope of the frame that is currently returning by comparing the scope's actor
    name with that given in the "paused" packet's frame.

    Since only one frame at a time will ever have a `frameFinished` property, it's
    good enough to simply watch for the top scope of the selected frame, and append
    the values whenever the frame has a `frameFinished` property.

    This commit adds tests to [`scopes.spec.js`].

* **Rename [`getSpecialVariables`] to `getFramePopVariables`.**

    This better reflects the function's actual purpose and behavior.

[Bug 1370923]: https://bugzilla.mozilla.org/show_bug.cgi?id=1370923
[`getScopes`]: https://github.com/devtools-html/debugger.html/blob/5bfe03c/src/utils/scopes.js#L158-L161
[`scopes.spec.js`]: https://github.com/devtools-html/debugger.html/blob/5bfe03c/src/utils/tests/scopes.spec.js
[`getSpecialVariables`]: https://github.com/devtools-html/debugger.html/blob/5bfe03c/src/utils/scopes.js#L73

### Test Plan

Each change includes changes or additions to [`scopes.spec.js`].

### DIFF
--- a/src/utils/scopes.js
+++ b/src/utils/scopes.js
@@ -165,7 +165,7 @@ export function getScopes(
       // the return value or the exception being thrown as special variables.
       if (
         scope.actor === selectedScope.actor &&
-        selectedFrame.id == pauseInfo.frame.id
+        selectedFrame.id === pauseInfo.frame.id
       ) {
         vars = vars.concat(getFramePopVariables(pauseInfo, key));
       }
--- a/src/utils/scopes.js
+++ b/src/utils/scopes.js
@@ -5,7 +5,6 @@
 // @flow
 
 import { toPairs } from "lodash";
-import { get } from "lodash";
 import { simplifyDisplayName } from "./frame";
 import type { Frame, Pause, Scope, BindingContents } from "debugger-html";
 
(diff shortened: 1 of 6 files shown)


src/utils/tests/scopes.spec.js
```diff
@@ -175,5 +169,103 @@ describe("scopes", () => {
...
+    });
+
+    it("throwing scope", () => {
+      const pauseData = {
+        frame: {
+          this: {}
+        },
+        why: {
+          frameFinished: {
+            throw: "a party"
```
jasonLaster: :cake:

jimblandy:
> It might be nice to also include the mochitest in this PR as well.

I had entirely forgotten that I had already done the work to resuscitate that test.

src/utils/scopes.js
```diff
@@ -155,9 +155,13 @@ export function getScopes(
...
         : getBindingVariables(bindings, key);
 
-      // show exception, return, and this variables in innermost scope
-      if (scope.actor === pausedScopeActor) {
-        vars = vars.concat(getSpecialVariables(pauseInfo, key));
+      // On the innermost scope of a frame that is just about to be popped, show
+      // the return value or the exception being thrown as special variables.
+      if (
+        scope.actor === selectedScope.actor &&
+        selectedFrame.id == pauseInfo.frame.id
```
codehag: we should use `===` because this could hold true for `selectedFrame.id = undefined; pauseInfo.frame.id = 0`

src/utils/scopes.js
```diff
@@ -70,13 +69,14 @@ function getSourceBindingVariables(
   return bound.concat(unused);
 }
 
-export function getSpecialVariables(pauseInfo: Pause, path: string) {
-  const thrown = get(pauseInfo, "why.frameFinished.throw", undefined);
-
-  const returned = get(pauseInfo, "why.frameFinished.return", undefined);
+export function getFramePopVariables(pauseInfo: Pause, path: string) {
```
codehag: :+1: nice renaming. might need a comment for those who are not familiar with frames

src/utils/tests/scopes.spec.js
```diff
@@ -175,5 +169,103 @@ describe("scopes", () => {
...
+            arguments: [],
+            variables: {}
+          },
+          parent: null
+        },
+        this: {}
+      };
+
+      const scopes = getScopes(pauseData, selectedFrame);
+      expect(scopes).toMatchObject([
```
codehag: :+1: i like this better than toMatchSnapshot.

jimblandy:
@jasonLaster Okay, I've added the mochitests. Between some problems they uncovered and some flow typing issues, a bunch has changed here, so it probably needs a re-review.

jasonLaster:
added Yulia's suggestion for a `===` and merged. Great work!

jimblandy:
@codehag Ah, sorry I missed your comments! I'll look more closely in the future.

PR #4584, opened 2017-11-04
Disallow uncaught exceptions

jasonLaster:
We were allowing uncaught exceptions in our unit tests, which is a bad practice:

We now throw when we see it.

screen shot 2017-11-04 at 11 56 07 am

### DIFF
--- a/src/actions/tests/pause.spec.js
+++ b/src/actions/tests/pause.spec.js
@@ -2,7 +2,8 @@ import {
   actions,
   selectors,
   createStore,
-  getHistory
+  getHistory,
+  makeSource
 } from "../../utils/test-head";
 
 const { isStepping } = selectors;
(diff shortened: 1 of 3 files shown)


src/actions/tests/pause.spec.js
```diff
@@ -2,7 +2,8 @@ import {
   actions,
   selectors,
   createStore,
-  getHistory
+  getHistory,
+  makeSource
```
jimblandy: Are these calls fixes for errors that had been previously being ignored? Usually one should submit unrelated fixes as a separate PR, and treat the harness fixes as blocked on that.

src/test/tests-setup.js
```diff
@@ -28,4 +28,5 @@ setConfig(config);
 
 process.on("unhandledRejection", (reason, p) => {
   console.log("Unhandled Rejection at:", p, "reason:", reason);
+  throw reason;
```
jimblandy: A thousand times yes.

src/actions/tests/pause.spec.js
jasonLaster: yep, this is a tad lazy... 

I don't think we gain anything in version control land by doing that though here

PR #4585, opened 2017-11-04
Add some additional helper methods

jasonLaster:


src/test/mochitest/head.js
```diff
@@ -363,6 +363,24 @@ async function waitForPaused(dbg) {
   );
 }
 
+/*
+ * useful for when you want to see what is happening
+ * e.g await waitForever()
+ */
+function waitForever() {
```
codehag: this seems like a method used while debugging, but not something that is used in the code...

jasonLaster: they both are

jasonLaster: they both are, i think they are helpful patterns though that we should make easier for newcomers

PR #4586, opened 2017-11-04
Mochi intermittent script

jasonLaster:
Mochi Intermittents is a small script for running the full mochitest
suite several times to look for intermittent failures.

There are two commands:
* run - runs the full suite N times
* uniqErrors - scans the logs for uniq test errors and their count

It is easy to pick up on trends by running `less logs`
and looking for `TEST-START <test-name>`

screen shot 2017-11-04 at 4 24 03 pm

### DIFF
--- a/bin/mochi-intermittents.js
+++ b/bin/mochi-intermittents.js
@@ -1,0 +1,52 @@
+const shell = require("shelljs");
+const _ = require("lodash");
+const fs = require("fs");
+
+/*
+Mochi Intermittents is a small script for running the full mochitest
+suite several times to look for intermittent failures.
+
+There are two commands:
+* run - runs the full suite N times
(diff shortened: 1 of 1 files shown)


bin/mochi-intermittents.js
```diff
@@ -0,0 +1,52 @@
...
+  errorCounts = uniq_errors.map(error => ({
+    error,
+    count: errors.filter(e => e == error).length
+  }))
+
+  console.log(errors.length)
+}
+
+(() => {
+  // run(50);
```
bomsy: Should these be commented out?

jasonLaster: yeah... it will require some fine tuning when run

bomsy: cool

PR #4600, opened 2017-11-06
Converted tab, breakpoint, watch expression, and search bar Close Button SVG's to Images

jselmani:
Associated Issue: #4350 

### Summary of Changes
* Converted Close Button SVG's in tab, breakpoint, watch expression, and the search bar.

I used the close button styling from the Atom text editor.  Their close button change in color when they are hovered over.  A GIF follows for representation:

atomsourcetabs

##### Source Tabs
sourcetabs

##### Watch Expressions/Breakpoints
watchexpressionandbreakpoints

##### Find In Files
findinfiles

##### Source File Search
sourcefilesearch

### DIFF
--- a/src/components/Editor/SearchBar.css
+++ b/src/components/Editor/SearchBar.css
@@ -14,12 +14,6 @@
 
 .search-field .close-btn {
   margin-top: 7px;
-  display: inline-flex;
-}
-
-.search-field .close-btn img.close {
-  margin-top: 3px;
-  margin-left: -1.5px;
 }
 
 .search-bottom-bar * {
--- a/src/components/Editor/SearchBar.css
+++ b/src/components/Editor/SearchBar.css
@@ -12,8 +12,14 @@
   height: var(--editor-searchbar-height);
 }
 
-.search-bar .close-btn {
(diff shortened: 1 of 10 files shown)


jasonLaster:
will look today

jasonLaster:
Hey @jselmani just looked and saw some differences in the dark theme:

* the size is bigger
* the background color is blue

### new


### old

jasonLaster:
pushed snapshots, `jest -u src`.
pushed a rebase

noticed some of the `x` were off




lets try looking at what the old CSS did to line it up because this is tough to get right and keep right :) we've broken it many times

codehag:
It looks good; this might need a rebase, im seeing an unrelated linting error

Garbee:
The lint error is a mispelling in the mochitest readme. Not caused by these changes and safe to merge since this won't overwrite that at all.

codehag:
Yes, but it isn't good practice to merge a branch with something failing even if it isn't related to the changes of this branch. if we start doing this we become sloppy, and it becomes easy to miss true failures. It is better to be strict for this reason.

Issue #3755, opened 2017-08-22
[Preview] add a mochitest


jasonLaster:
We have plenty of unit tests for preview, but it would be good to have atleast one mochitest to demonstrate hovering on a token

Fortunately the old debugger had some:

https://searchfox.org/mozilla-central/source/devtools/client/debugger/test/mochitest/browser_dbg_variables-view-popup-01.js

[reOpenPopup](https://searchfox.org/mozilla-central/source/devtools/client/debugger/test/mochitest/head.js#948)

#### Steps 

1. add a new test https://github.com/devtools-html/debugger.html/blob/master/docs/mochitests.md#L124-L125
2. `yarn mochi browser-dbg_preview`

jasonLaster:
@bomsy is on it

jasonLaster:
We sh

jasonLaster:
@bomsy taking your head off it because it's been a month

jasonLaster:
removing from our milestone as Sam, from ucosp can work on it

PR #4603, opened 2017-11-06
Preview mochitest

samuel-peers:
Associated Issue: #3755 

### Summary of Changes

* emulate hovering on simple variable, confirm contents of tooltip
* emulate hovering on object variable, confirm contents of popup


jasonLaster:
looks like the changes to simple1 broke this... perhaps, add a simple3 that is similar...

Then run `yarn mochi` to see what breaks. `yarn mochih` does fail w/ some input related tests

Issue #4575, opened 2017-11-03
Class name "watch-expressions-pane" is different in each locales, because it is translated


magicp-github:
**Steps to reproduce:**
Confirm class name "watch-expressions-pane" in each locales.

**Actual Results:**
In Japanese, "watch-expressions-pane" is "監視式-pane".
![watch-expressions-pane-class-name](https://user-images.githubusercontent.com/13914314/32373593-4dde245a-c0dd-11e7-927b-4a449b66dfa7.png)

**Expected Results:**
"watch-expressions-pane"

PR #4611, opened 2017-11-07
Fixes secondary pane css classes being localized

lukaszsobek:
Associated Issue: #4575

### Summary of Changes

* Added new property cssClass to getScopeItem, getWatchItem and getStartItems responsible for keeping the css class name of the item.
* added new property to typeCheck
* removed generation of class name based on header as it caused bugs now and might complicate things further in future.

### Test + Screenshot

End result:
screenshot_7

### DIFF
--- a/src/components/SecondaryPanes/index.js
+++ b/src/components/SecondaryPanes/index.js
@@ -129,7 +129,7 @@ class SecondaryPanes extends Component<Props> {
 
     return {
       header: L10N.getStr("scopes.header"),
-      cssClass: "scopes-pane",
+      className: "scopes-pane",
       component: Scopes,
       opened: prefs.scopesVisible,
       onToggle: opened => {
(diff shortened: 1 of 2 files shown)


src/components/SecondaryPanes/index.js
```diff
@@ -129,6 +129,7 @@ class SecondaryPanes extends Component<Props> {
 
     return {
       header: L10N.getStr("scopes.header"),
+      cssClass: "scopes-pane",
```
jasonLaster: can we use `className` or is that reserved?

src/components/shared/Accordion.js
```diff
@@ -62,12 +63,9 @@ class Accordion extends Component<Props, State> {
 
   renderContainer = (item: AccordionItem, i: number) => {
     const { opened, created } = this.state;
```
jasonLaster: :+1: nice

src/components/SecondaryPanes/index.js
lukaszsobek: The original version I made had className, but it felt confusing to use className property to pass values to className as we would have the same name in two places doing two different things.

Issue #4526, opened 2017-10-29
[Project Search] Closing Find window does not remove results from memory


lukaszsobek:
STR + AR:
- go to [this page](https://developmentarc.gitbooks.io/react-indepth/content/life_cycle/update/postrender_with_componentdidupdate.html)
- select gitbook.js
- press ctrl/cmd+shift+f in editor window
- type in "input" and press enter
- wait for results to appear
- close find window by pressing x
- press ctrl/cmd+shift+f again
- wait ...
- find window appears after a long delay with results from previous search 

ER:
- find window opens with an empty state

wldcordeiro:
@lukaszsobek Thanks for the report! Just FYI it's nice to include the UI component in question in the issue title. I wasn't sure if this was for file search (cmd+f) or project search (cmd+shift+f) so I've updated the title.

lukaszsobek:
@wldcordeiro Thank you! Will mention it next time.

princiya:
@wldcordeiro again, trying to understand the whole structure for the `search`. I am able to reproduce this bug and I want to take up this issue. Please assign it to me :)

I want to make sure, when the `close` button is clicked, the `state` is set right. In my attempt to debug, can you please help me figure out where is [this](https://github.com/devtools-html/debugger.html/blob/master/src/components/App.js#L14) function being declared? I don't see it in `selectors`.

princiya:
Oops, I see the issue is being claimed. Never mind, but please help me understand the source for that function :)

wldcordeiro:
@princiya I think you can most easily see this by using the project search, closing the tab/view and then investigating the store via something like the redux-devtools to see that the contents aren't cleared.

lukaszsobek:
@princiya it's yours for taking!

PR #4621, opened 2017-11-08
[Search] Clear results while closing project search

princiya:
Associated Issue: #4526 

### Steps

- Press ⌘ + ⇧ + F
- Enter a search term
- Press 'x'
- Press ⌘ + ⇧ + F
- The older search results are now cleared

### Before

Before

### After

After

### DIFF
--- a/src/actions/project-text-search.js
+++ b/src/actions/project-text-search.js
@@ -39,6 +39,10 @@ export function updateSearchStatus(status: string) {
   return { type: "UPDATE_STATUS", status };
 }
 
+export function closeProjectSearch() {
+  return { type: "CLOSE_PROJECT_SEARCH" };
+}
+
 export function searchSources(query: string) {
   return async ({ dispatch, getState }: ThunkArgs) => {
     await dispatch(clearSearchResults());
--- a/src/actions/tests/project-text-search.spec.js
+++ b/src/actions/tests/project-text-search.spec.js
@@ -113,4 +113,23 @@ describe("project text search", () => {
     dispatch(actions.updateSearchStatus(mockStatus));
     expect(getTextSearchStatus(getState())).toEqual(mockStatus);
   });
+
+  it("should close project search", async () => {
+    const { dispatch, getState } = createStore(threadClient);
+    const mockQuery = "foo";
+
+    await dispatch(actions.newSource(makeSource("foo1")));
+    await dispatch(actions.searchSources(mockQuery));
+
+    expect(getTextSearchResults(getState())).toMatchSnapshot();
+
+    dispatch(actions.closeProjectSearch());
+
+    expect(getTextSearchQuery(getState())).toEqual("");
+
+    const results = getTextSearchResults(getState());
+
+    expect(results).toMatchSnapshot();
+    expect(results.size).toEqual(0);
+  });
 });
--- a/src/reducers/project-text-search.js
+++ b/src/reducers/project-text-search.js
@@ -70,6 +70,12 @@ function update(
       return state.merge({
         results: state.get("results").clear()
       });
+
+    case "CLOSE_PROJECT_SEARCH":
+      return state.merge({
+        query: "",
+        results: state.get("results").clear()
+      });
   }
   return state;
 }
--- a/src/actions/tests/__snapshots__/project-text-search.spec.js.snap
+++ b/src/actions/tests/__snapshots__/project-text-search.spec.js.snap
@@ -70,6 +70,76 @@ Immutable.List [
 ]
 `;
 
+exports[`project text search should close project search 1`] = `
+Immutable.List [
+  Object {
+    "filepath": "http://localhost:8000/examples/foo1",
+    "matches": Array [
+      Object {
+        "column": 9,
+        "line": 1,
+        "match": "foo",
+        "sourceId": "foo1",
+        "text": "function foo1() {",
+        "value": "function foo1() {",
+      },
+      Object {
+        "column": 8,
+        "line": 2,
+        "match": "foo",
+        "sourceId": "foo1",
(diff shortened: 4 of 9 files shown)


princiya:
I have the bug fix in this PR as reported in this issue #4526.

- Should we clear the search query and results for `escape` key too?
- Atom editor does not clear the query or results when you press `escape` key or click the `x` button.

@wldcordeiro what do you think?

princiya:
And, I don't know how to update the snapshot :). The CI fails.

bomsy:
@princiya updating the snapshot should be `yarn jest -- -u`. if i remember correctly.

src/actions/project-text-search.js
```diff
@@ -34,6 +36,14 @@ export function clearSearchResults() {
   };
 }
 
+export function closeProjectSearch() {
+  return ({ dispatch, getState }: ThunkArgs) => {
+    dispatch(clearSearchQuery());
```
wldcordeiro: I think I would prefer having the reducers listen for the actions rather than a thunk that fires three.


```js
export function closeProjectSearch() {
  return { type: 'CLOSE_PROJECT_SEARCH' }
}
```

wldcordeiro: In the project search reducer...

```js
    case "CLOSE_PROJECT_SEARCH":
      return state.merge({
        query: "",
        results: state.get("results").clear()
      });
```

in the ui reducer

```js
    case "CLOSE_PROJECT_SEARCH": {
      if (state.get("activeSearch") === "project") {
        return state.set("activeSearch", null);
      }
      return state;
    }
```

wldcordeiro:
> Should we clear the search query and results for escape key too?
> Atom editor does not clear the query or results when you press escape key or click the x button.

Performance is even more of a priority for us so I think clearing the results is good but we could likely leave the query in. That way if you want to search the same stuff again you can easily but we don't store all the results in memory.

princiya:
@wldcordeiro Updated changes. Thanks for the pointers. 

TIL, we can have the same action dispatcher in two different reducers (same name, like `CLOSE_PROJECT_SEARCH`).

The next question, do I add in the changes to the `escape` key (clear results, but keep the query) in this PR or create a new issue and PR?

princiya:
@bomsy Updated the snapshot. The CI still fails :(

src/actions/ui.js
```diff
@@ -24,6 +24,10 @@ export function closeActiveSearch() {
   };
 }
 
+export function closeProjectSearch() {
```
wldcordeiro: No need for the duplicate. Just the one above is fine. The `type` is still the same.

src/actions/project-text-search.js
```diff
@@ -34,6 +34,10 @@ export function clearSearchResults() {
   };
 }
 
+export function closeProjectSearch() {
+  return { type: "CLOSE_PROJECT_SEARCH" };
```
codehag: this isn't dispatching... and the only place a dispatch is used is in the test, not in the click handler. any ideas about what is going on here?

src/actions/tests/project-text-search.spec.js
```diff
@@ -101,4 +101,17 @@ describe("project text search", () => {
 
     expect(getTextSearchResults(getState())).toMatchSnapshot();
   });
+
+  it("should close project search", () => {
+    const { dispatch, getState } = createStore(threadClient);
```
codehag: hmm, this test is almost there but it isn't testing what we need to test. we need to have a populated project search before closing it to determine of the close function is working properly. otherwise we are testing the default behavior (query is empty by default, results are empty by default). Maybe we can do this for now by dispatching a search query and mocking out the function that populates the list?

src/actions/project-text-search.js
princiya: @codehag It is used in the click handler [here](https://github.com/princiya/debugger.html/blob/b17a22194cca393d9888fdc1bf2315ed8dc1f7c6/src/components/ProjectSearch/TextSearch.js#L191)

src/actions/tests/project-text-search.spec.js
princiya: ok, will update the test spec. 👍

src/actions/project-text-search.js
codehag: but there is no dispatching that is going on, it is just calling the action which is returning an object, or maybe i am missing a wrapper somewhere that does the dispatch?

codehag: or i just forgot how action creators work 😅

princiya: @codehag I guess the latter one! :)

src/actions/tests/project-text-search.spec.js
```diff
@@ -101,4 +101,23 @@ describe("project text search", () => {
...
+    await dispatch(actions.searchSources(mockQuery));
+
+    expect(getTextSearchResults(getState())).toMatchSnapshot();
+
+    dispatch(actions.closeProjectSearch());
+
+    expect(getTextSearchQuery(getState())).toEqual("");
+
+    const results = getTextSearchResults(getState());
+
```
princiya: @codehag updated the test spec. let me know what you think.

src/reducers/project-text-search.js
```diff
@@ -53,6 +53,12 @@ function update(
       return state.merge({
         results: state.get("results").clear()
       });
+
+    case "CLOSE_PROJECT_SEARCH":
+      return state.merge({
+        query: "",
```
princiya: @wldcordeiro should we use `state.remove("query")` instead? not sure why one is preferred over the other.

jasonLaster:
just rebased and looks great!

Issue #4639, opened 2017-11-09
Clicking on a URL in the scopes panel replaces the debugger  


juliandescottes:
STRs:
- go to mozilla.org
- break anywhere
- go to Scopes panel, find window > location >href
- click on a link 

The URL gets loaded and replaces the debugger UI. Only way to restore the debugger is to close and reopen the toolbox.

See attached GIF

![debugger_navigates_to_url](https://user-images.githubusercontent.com/1141550/32603742-2cf817c4-c54b-11e7-92ee-bba9d23c74d3.gif)

juliandescottes:
This should be fixed by https://github.com/devtools-html/devtools-core/commit/bc7bb952d66685c709216a675b3d674856df1c32 which was released in devtools-reps 0.14.0

It adds an omitLinkHref prop to the String rep, which is true by default.

juliandescottes:
Looks like 0.14.0 brings some non-backward compatible changes (new props for ObjectInspector), and I'm not sure how we feel about bringing a new version in at the last minute. I would really like that to be fixed before we branch FF58 though.

@nchevobbe how would feel about doing a v0.12.4 release of reps only with the fix for  omitLinkHref ?

PR #4641, opened 2017-11-09
Fix 4639 - stop rendering href attributes in String reps for URLs

juliandescottes:
This PR switches to the newly released reps v0.12.4 which by default won't render href attributes for links in String reps.

This fixes the issue for #4639 , hopefully nothing else in the debugger actually relied on this behavior?


juliandescottes:
tests are passing, merging.

Issue #3991, opened 2017-09-12
Hang when going from Console to Debugger


digitarald:
https://perfht.ml/2joU0UH

- Codemirror's `updateHeightsInViewport` is expensive due to mostly reflows: https://perfht.ml/2jnqRsU
- There are 2 cycles of `updateHeightsInViewport` for loading one file, which might be overhead. One is caused by `setOption` and the other by `resetLineNumberFormat`: https://perfht.ml/2joUJ8n

digitarald:
Maybe related to #3400

jasonLaster:
Looks like this is coming from `resetLineNumberFormat`, which was added for wasm


![Screen Shot 2017-09-19 at 10.57.47 AM.png](https://shipusercontent.com/1de2605b4fcebc313490ba415942d51a/Screen%20Shot%202017-09-19%20at%2010.57.47%20AM.png)

jasonLaster:
@digitarald do you have an STR here?

digitarald:
Tried different cases of clicking from Console logs to the JS source and could not reproduce a noticeable super long hangs anymore (just long delays on loading the sources).

jasonLaster:
I found one :) 

when the debugger is open locally

digitarald:
The root cause of this issue seems fixed.

Amazon: https://perfht.ml/2V1Rx1c - snappy.
Github: https://perfht.ml/2V0YTlq - we seem to be waiting for downloading the source maps and the worker before drawing any UI, which is an odd choice. I will file that as separate issue though.

PR #4653, opened 2017-11-09
Fix Console Hang

jasonLaster:
Associated Issue: #3991

### Summary of Changes

This fixes a super long hang (10seconds) when parsing the debugger bundle locally. Generally we should try not to wait for something when it is on the critical path

### Test Plan

all mochitests pass


bomsy:
i like!

Issue #3696, opened 2017-08-17
[stepping] add some unit tests


jasonLaster:
It would be great to boost our coverage of `src/actions/pause`

```diff
diff --git a/src/actions/tests/pause.spec.js b/src/actions/tests/pause.spec.js
index 46c5b3c..58708e3 100644
--- a/src/actions/tests/pause.spec.js
+++ b/src/actions/tests/pause.spec.js
@@ -22,15 +22,15 @@ const mockThreadClient = {
   }
 };

+const mockPauseInfo = {
+  frames: [{ id: 1, scope: [], location: { sourceId: "foo1", line: 4 } }],
+  loadedObjects: [],
+  why: {}
+};
+
 describe("pause", () => {
   it("should set and clear the command", async () => {
     const { dispatch, getState } = createStore(mockThreadClient);
-    const mockPauseInfo = {
-      frames: [{ id: 1, scope: [], location: { sourceId: "foo1", line: 4 } }],
-      loadedObjects: [],
-      why: {}
-    };
-
     await dispatch(actions.paused(mockPauseInfo));
     dispatch(actions.stepIn());
     expect(isStepping(getState())).toBeTruthy();
@@ -46,4 +46,21 @@ describe("pause", () => {
     await dispatch(actions.resumed());
     expect(client.evaluate.mock.calls.length).toEqual(0);
   });
+
+  it("should only step when paused", async () => {
+    const client = { stepIn: jest.fn() };
+    const { dispatch } = createStore(client);
+
+    dispatch(actions.stepIn());
+    expect(client.stepIn.calls.length).toEqual(0);
+  });
+
+  it("should step when paused", async () => {
+    const client = { stepIn: jest.fn() };
+    const { dispatch } = createStore(client);
+
+    await dispatch(actions.paused(mockPauseInfo));
+    dispatch(actions.stepIn());
+    expect(client.stepIn.calls.length).toEqual(1);
+  });
```

PR #4655, opened 2017-11-10
Breakpoint setting unit tests

James300:
Associated Issue: #3696

### Summary of Changes

* Added a unit test for making sure breakpoint stepping work while paused
* Added a unit test for making sure breakpoint stepping doesn't work while not paused

### DIFF
--- a/src/actions/tests/pause.spec.js
+++ b/src/actions/tests/pause.spec.js
@@ -51,6 +51,25 @@ describe("pause", () => {
       await stepped;
       expect(isStepping(getState())).toBeFalsy();
     });
+
+    fit("should only step when paused", async () => {
+      const client = { stepIn: jest.fn() };
+      const { dispatch } = createStore(client);
+
+      dispatch(actions.stepIn());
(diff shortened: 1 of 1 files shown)


src/actions/tests/pause.spec.js
```diff
@@ -51,6 +51,24 @@ describe("pause", () => {
...
+      expect(client.stepIn.mock.calls.length).toEqual(0);
+    });
+
+    it("should step when paused", async () => {
+      const { dispatch, getState } = createStore(mockThreadClient);
+      const mockPauseInfo = createPauseInfo();
+
+      await dispatch(actions.newSource(makeSource("foo1")));
+      await dispatch(actions.paused(mockPauseInfo));
+      dispatch(actions.stepIn());
```
jasonLaster: how do we know this hasn't finished?

James300: This one should finish right? This is checked that it does work while paused

jasonLaster: oh yeah. true

Issue #4659, opened 2017-11-10
Show the number of stack frames in the collapsed framework frame group


amitzur:
I though it would be nice to show the number of collapsed frames.
Going to file a PR now.

See in these screenshots:
![screen shot 2017-11-10 at 2 39 21 pm](https://user-images.githubusercontent.com/394320/32659226-6b6cefe4-c626-11e7-8ce9-15e8c3df20aa.png)
![screen shot 2017-11-10 at 2 39 28 pm](https://user-images.githubusercontent.com/394320/32659227-6bddda2e-c626-11e7-83f5-76d517dcde89.png)

PR #4660, opened 2017-11-10
Frame count badge

amitzur:
Associated Issue: #4659 

### Summary of Changes

* new react component - `<Badge>{...}</Badge>` is intended to show a number with rounded background

* added some css utility classes a la bootstrap. If this is considered good, I'd add more and do some refactoring in existing css.

### Test Plan

I updated the snapshot for Group.js test.

This is pretty similar to GitHub's UI (even though I saw that only after I implemented):
screen shot 2017-11-10 at 2 57 21 pm

### DIFF
--- a/src/components/App.css
+++ b/src/components/App.css
@@ -80,3 +80,7 @@ button:focus {
 .text-center {
   text-align: center;
 }
+
+.min-width-0 {
+  min-width: 0;
+}
--- a/src/components/App.css
+++ b/src/components/App.css
@@ -60,3 +60,23 @@ button:focus {
   left: 50%;
   transform: translate(-50%, -50%);
 }
+
+.d-flex {
+  display: flex;
+}
+
+.align-items-center {
+  align-items: center;
+}
+
+.rounded-circle {
+  border-radius: 50%;
+}
+
+.text-white {
+  color: white;
+}
+
+.text-center {
+  text-align: center;
+}
--- a/src/components/SecondaryPanes/Frames/Frames.css
+++ b/src/components/SecondaryPanes/Frames/Frames.css
@@ -51,7 +51,7 @@
 .frames .title {
   text-overflow: ellipsis;
   overflow: hidden;
-  margin: 7px 1em 7px 0;
+  margin: 7px 0.5em 7px 0;
 }
 
 .frames ul li:hover,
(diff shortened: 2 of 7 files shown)


jasonLaster:
I like this!

* lets clean up the patch
* add a screenshot so we can review UI (long badges, narrow sidebar, ..)
* review the css style... I like the utility class names, but it is different, so lets get some buy in (@codehag, @wldcordeiro)

amitzur:
- What cleanup is needed? The irrelevant linted files?
- Here's a low res screencast:
(i think the edge case where it overlaps isn't that important, but let me know otherwise)
fw_count_pr
- Would be happy to discuss and elaborate on slack or in one of the weekly calls. @codehag @wldcordeiro

jasonLaster:
> What cleanup is needed? The irrelevant linted files?

yep
 
> UI nit

I think i would prefer a smaller badge... but that's a minor thing.

@violasong what do you think?

violasong:
Yes, a bit less top and bottom padding on the badge would be nice. Console has a badge style you could match.

jasonLaster:
thanks. here's the css
https://searchfox.org/mozilla-central/source/devtools/client/themes/webconsole.css#100

we probably don't want to pull in all of webconsole.css, but we can copy it into a file that we can eventually share... if that's what we want to do

amitzur:
> Edit: The Console badge class name is .message-repeats
Where in the console's UI can I find it?

amitzur:
> > What cleanup is needed? The irrelevant linted files?

> yep

I wanna do this myself, but not sure - how do I pull commits out of the branch? Do I create another branch for this PR?

amitzur:
> Yes, a bit less top and bottom padding on the badge would be nice. Console has a badge style you could match.

@violasong So, you're saying no circle style? And do you think the colors here work?

jasonLaster:
Hmm, look into `git rebase interactive` or just make another commit on top, or squash commits ...

amitzur:
- [x] UI cleaned up
- [x] patch cleaned up

Have a look at how it shows up with made up number of 1-3 digits:
screen shot 2017-11-13 at 8 00 45 pm
screen shot 2017-11-13 at 8 00 59 pm

wldcordeiro:
👏 I dig this! Great job @amitzur! Like @jasonLaster mentioned cleaning up the CSS would be nice but can be done in a follow-up.

jasonLaster:
I don't know. I like the CSS, and it's nice to have the full example w the
util vlasses. It just brings up the qs do we want to adopt this style :)

Issue #4631, opened 2017-11-08
[flow] drop react-immutable-types


jasonLaster:
We are switching from PropTypes to Flow, so we should try and drop Improptypes. The good news is we'll also be able to save some space too.

```diff
--- a/src/components/SecondaryPanes/index.js
+++ b/src/components/SecondaryPanes/index.js
@@ -9,7 +9,6 @@ import React, { Component } from "react";
 import { connect } from "react-redux";
 import { bindActionCreators } from "redux";
 import { features } from "../../utils/prefs";
-import ImPropTypes from "react-immutable-proptypes";
 
 import actions from "../../actions";
 import {
@@ -271,7 +270,7 @@ SecondaryPanes.propTypes = {
   evaluateExpressions: PropTypes.func.isRequired,
   pauseData: PropTypes.object,
   horizontal: PropTypes.bool,
-  breakpoints: ImPropTypes.map.isRequired,
+  breakpoints: PropTypes.object,
breakpointsDisabled: PropTypes.bool,
```

PR #4667, opened 2017-11-10
Drop react-immutable-proptypes

sharathnarayanph:
Associated Issue: #4631 

Summary of Changes

Removed react-immutable-proptypes and replaced with proptypes

Test Plan

Secondary pane functionality should work as intended


sharathnarayanph:
I assumed immutable types would be used in a lot of places. Looks like it is the only file where it is used!

bomsy:
Yeah! it was used in a lot of places but has been removed over time

Issue #4579, opened 2017-11-03
[Editor] Remove Breakpoint Editor management


jasonLaster:
The breakpoint component will attempt to show the source text if it is not available: [here][h]

```js
showSourceText(editor, selectedSource.toJS());
```

This was done to accommodate some weird edge cases prior to cleaning up the editor's lifecyle. We should be able to safely remove it now.

[h]: https://github.com/devtools-html/debugger.html/blob/master/src/components/Editor/Breakpoint.js#L55

PR #4677, opened 2017-11-12
Remove Breakpoint Editor management

nyrosmith:
Associated Issue: #4579 

### Summary of Changes

* Remove `showSourceText` from Breakpoint component


jasonLaster:
Awesome! Lets double and triple check this because we added it before because there were some race conditions:

I think the STR for the old bug, was to have a BP that is in an original source:

1. add the bp
2. toggle between original and generated and make sure you see it
3. remove the bp jump to the other source make sure it's gone...


...

jasonLaster:
Works great!

Issue #3307, opened 2017-07-11
[Parser] Parse Failed on "application/ld+json" SCRIPT


darkwing:
Many WordPress sites (mine included) will have script tags that look like this:

```
<script type='application/ld+json'>{"@context":"http:\/\/schema.org","@type":"Person","url":"https:\/\/davidwalsh.name\/","sameAs":["https:\/\/www.facebook.com\/davidwalshblog","https:\/\/plus.google.com\/114538814489633467974?rel=author","https:\/\/twitter.com\/davidwalshblog"],"@id":"#person","name":"David Walsh"}</script>
```

Essentially a `script` tag with JSON.  Debugger.html throws the following:

```
ast.js:33 parse failed 
```

I know this type of script is an outlier but I figured it was worth reporting.  Also somewhat curious why a script tag containing only JSON would bomb during parsing.

jasonLaster:
nice find.

@ryanjduffy do you think this is related to `parse-script-tags`?

ryanjduffy:
Likely. I'll look into it

ryanjduffy:
The issue here is that JSON isn't itself valid, stand-alone JavaScript. So when the string is passed to the [debugger.html parser](https://github.com/devtools-html/debugger.html/blob/master/src/utils/parser/utils/ast.js#L23), it fails. It shouldn't inhibit the debugger since the failure only means that particular script will be omitted from the AST for the HTML page.

@darkwing Are you able to debug other scripts on the page?

I could build in more intelligence into `parse-script-tags` to respect the `type` attribute and thereby not try to parse non-JS code but it wouldn't prevent that error from printing when a script tag contained otherwise invalid JS. Provided debugging works well otherwise, a simpler solution might be to either make the warning more descriptive or suppress the warning altogether in this scenario.

codehag:
@ryanjduffy are you still looking into this?

jasonLaster:
:wave: @ryanjduffy, just checking in cause it looks like this issue might have fallen off the radar. If you're busy w/ the parser work, we can take your head off this.

jasonLaster:
Also, you should have an invitation to the community team in your inbox
![screen shot 2017-11-04 at 6 10 26 pm](https://user-images.githubusercontent.com/254562/32409883-821440b6-c18b-11e7-9b0c-b984280b0dd7.png)

ryanjduffy:
I haven't looked at this but I will next week

ryanjduffy:
[...]

PR #4684, opened 2017-11-13
Fix including non-JS scripts from HTML source

ryanjduffy:
Associated Issue: #3307 

### Summary of Changes

Updates the dependency to `parse-script-tags` to version 0.1.4 which supports   omitting script tags that have non-JS types.

### Test Plan

* Added breakpoints on a Wordpress site (http://davidwalsh.name) with non-JS `<script>`s
* Created a simple sample with non-JS `<script>`:

```html
<html>
<script type="application/ld+json">
var notScript = true;
alert(notScript);
</script>
<script>
	console.log('from script');
</script>
<body>
Not Script
</body>
</html>
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -85,7 +85,7 @@
     "lodash": "^4.17.4",
     "lodash.kebabcase": "^4.1.1",
     "md5": "^2.2.1",
-    "parse-script-tags": "^0.1.1",
+    "parse-script-tags": "^0.1.4",
     "pretty-fast": "^0.2.2",
     "prop-types": "^15.6.0",
     "react": "^15.6.2",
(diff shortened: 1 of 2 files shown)


jasonLaster:
screen shot 2017-11-21 at 9 50 18 am

codehag:
this change needed to be backed out (https://github.com/devtools-html/debugger.html/pull/4840), it was causing a memory leak on linux machines. it looks like it was coming from an imported dependancy, can you take a look @ryanjduffy ?

ryanjduffy:
Will do! I've started on a smaller, special-purposed sub-module to replace the imported dependency which should both address the leak (assuming I code it correctly) and the file size concern @jasonLaster referenced in ryanjduffy/parse-script-tags#6.

Issue #4151, opened 2017-09-24
[Tabs] Show Framework logos


jasonLaster:
It would be really cool if we could show a framework logo in the source tabs. This would help us differentiate between the Expressions component, actions, and reducer.

To figureout if a file is a component, actions, or reducer we need to check the AST.

Component: imports react and uses createClass or the Component class
Action: has a function with an inner that has an object with dispatch
Reducer: has an update function with a switch statement

canonical actions

```js
export function resumed() {
  return ({ dispatch }) => {
  };
}

export function resumed() {
  return function ({ dispatch }) {
  };
}
```

canonical reducer:
```js
function update(
  state: Record<BreakpointsState> = initialState(),
  action: Action
) {
  switch (action.type) {
  }
}
```

jasonLaster:
Here's a [patch][p] to help get started

[p]:https://gist.github.com/jasonLaster/9045e18ce2a74369187e9bb33eee4610

jasonLaster:
James is working on it

jasonLaster:
We landed a util for [`isReactComponent`](https://github.com/devtools-html/debugger.html/blob/master/src/workers/parser/frameworks.js#L3).

We can now easily check :)

and add it here:

```diff
diff --git a/src/components/Editor/Tabs.js b/src/components/Editor/Tabs.js
index 50aba30..927b8a2 100644
--- a/src/components/Editor/Tabs.js
+++ b/src/components/Editor/Tabs.js
@@ -451,6 +451,10 @@ class SourceTabs extends PureComponent<Props, State> {
   getSourceAnnotation(source) {
     const sourceObj = source.toJS();
 
+    if (isReactComponent) {
+      return <img className="react" />;
+    }
+
     if (isPretty(sourceObj)) {
       return <img className="prettyPrint" />;
     }
```

jasonLaster:
```diff
diff --git a/src/components/Editor/Tabs.js b/src/components/Editor/Tabs.js
index ae3f5ab..4c910d5 100644
--- a/src/components/Editor/Tabs.js
+++ b/src/components/Editor/Tabs.js
@@ -453,8 +453,13 @@ class SourceTabs extends PureComponent<Props, State> {
   }
 
   getSourceAnnotation(source) {
+    const { sourceMetaData } = this.props;
     const sourceObj = source.toJS();
 
+    if (sourceMetaData && sourceMetaData.isReactComponent) {
+      return <img className="react" />;
+    }
+
[...]

PR #4696, opened 2017-11-15
WIP - Action and Reducer Parser

James300:
Associated Issue: #4151

### Summary of Changes

* Added a way to determine if a file is a redux action file (or at least contains an action)
* (Coming soon) Added a way to determine if a file is a redux reducer file (or at least contains a reducer)

### Test Plan

* Added unit tests for new functions
* I plan to add more before this is done


jasonLaster:
going to close because we're going to try a different tactic for reducers / actions

Issue #4459, opened 2017-10-21
[Preview] Mis-judge the in-scope inner location as out-of-scope from its parent scope


Fischer-L:
STR:
1. visit http://todomvc.com/examples/react/#/active
2. add a bp in TodoModel.js#46
3. click toggle all in todomvc
4. step over to proceed to Line 50
5. hover `this` on Line 46

ER: see the preview of `this` on Line 46

AR: no preview of `this` on Line 46. See the gif below
![0pusqajliw](https://user-images.githubusercontent.com/5627487/31851735-3d6c88bc-b69e-11e7-98b6-07269e2a9d44.gif)

Root cause:
This issue was found during working on #4369 .
In the step 4, while getting out-of-scope locations[1], we have the bp position [2] and the location of `app.TodoModel.prototype.toggleAll` [3] and the location of the function arg for `this.todos.map` [4] (See [5] for the `toggleAll` method).

The location [4] now is considered out-of-scope by `containsPosition` because it doesn't contains the [2] position. This is correct from the view point of location-position relationship. As a result the Line 46 was totally ruled as out-of-scope.

The fix probably should be improve `getOutOfScopeLocations` function in handling this tricky scope thing such as concat comment locations, handling overlapping-inner location before calling `containsPosition` to do its job. (`containsPosition` did the right job and should keep focusing on location-position relationship)

[1] https://github.com/devtools-html/debugger.html/blob/c5177581a8af4b797f9aa56f82214216de0176c6/src/workers/parser/getOutOfScopeLocations.js#L86

[2] Object { sourceId: "server1.conn173.child2/source34", line: 50 }

[3]
```
{ // the location of `app.TodoModel.prototype.toggleAll` 
  end: Object { line: 51, column: 2 }
  start: Object { line: 41, column: 37 }
}
```
[4] 
```
{ // the location of the function arg for `this.todos.map`. Now will be filtered as out-of-scope
  end: Object { line: 48, column: 3 }
  start: Object { line: 46, column: 30 }
}
```

[5]
```js
 app.TodoModel.prototype.toggleAll = function (checked) {
	// Note: it's usually better to use immutable data structures since they're
	// easier to reason about and React works very well with them. That's why
	// we use map() and filter() everywhere instead of mutating the array or
	// todo items themselves.
	this.todos = this.todos.map(function (todo) { // Line 46
		return Utils.extend({}, todo, {completed: checked});
	});
       
        this.inform(); // Line 50
};
```

[...]

Issue #1574, opened 2016-12-30
Hover on a variable


jasonLaster:
Checklist
- [x] Show a popup in CodeMirror #1840
- [x] Add a JS parser in a worker #1841
- [x] Add an `expressionAtLocation` function for getting the expression to eval #1842
- [x] Preview values in the popup #1843

SebastianZ:
For what it's worth, here's a screenshot of how this feature looked in the old UI:

![Variable value infotip](https://cloud.githubusercontent.com/assets/958943/22205208/022b7220-e176-11e6-86f0-6b3c7dbf46a3.png)

And people are already [starting to wonder](http://stackoverflow.com/q/40800183/432681) where this feature is gone.

Because this is a regression in regard of the old UI and a feature that is presumably used quite often, I believe this should get some priority.

Sebastian

clarkbw:
Lets start with an MVP of hover on:

* Variable
* Function

## Variable

Hover over a variable to inspect.  Popup should display:
* ObjectInspector for variable value
* Action to Reveal variable in Scopes
* Action to Add to Watch Expressions

## Function

Hover over a function to inspect.  Popup should display:
* Action to jump to function definition
* Action to find occurrences of function (in current file)

### Chrome

![](https://cloud.githubusercontent.com/assets/95570/18097151/9dbd7b6e-6e91-11e6-9515-2aa261a17772.png)

### DXR

<img width="300" alt="screen shot 2017-01-11 at 3 11 10 pm" src="https://cloud.githubusercontent.com/assets/2134/22262151/9c935090-e224-11e6-91a6-ee98a3fe378b.png">

jasonLaster:
closing this in favor of a more modern tracking issue

Nexxado:
@SebastianZ you're right on that, i'm not using Firefox solely because its missing this critical feature which chrome does flawlessly.

jasonLaster:
@Nexxado what is missing in the current version?

Nexxado:
@jasonLaster sorry,  I should have been more specific.
I'm referring to not being able to see variables value (on hover)  in the debugger

jasonLaster:
@Nexxado that's still pretty vague.

for instance this example should work: https://small-vals.glitch.me/

we intentionally dont show preview popups for out of scope functions.

Nexxado:
@jasonLaster Here (see image) when hovering over the function's arguments I don't see its value.

Could it be related to the fact that the code is minified?
[...]

PR #4699, opened 2017-11-15
[Preview] mis-judge in scope inner locations

Fischer-L:
Associated Issue: #4459 

### Screenshots/Videos 
4459_verification

### DIFF
--- a/src/workers/parser/getOutOfScopeLocations.js
+++ b/src/workers/parser/getOutOfScopeLocations.js
@@ -38,26 +38,26 @@ function getLocation(func) {
 }
 
 /**
- * Reduces an array of locations to remove items that are completely enclosed
- * by another location in the array.
+ * Return an new locations array which excludes
+ * items that are completely enclosed by another location in the input locations
+ *
+ * @param locations Notice! The locations MUST be sorted by `sortByStart`
+ *                  so that we can do linear time complexity operation.
  */
-function removeOverlaps(
-  locations: AstLocation | AstLocation[],
-  location: AstLocation
-) {
-  // support reducing without an initializing array
-  if (!Array.isArray(locations)) {
-    locations = [locations];
+function removeOverlaps(locations: AstLocation[]) {
+  const newLocs = [];
+  const length = locations.length;
+  if (length) {
+    newLocs.push(locations[0]);
+    for (let i = 1, parentIdx = 0; i < length; i++) {
+      const loc = locations[i];
+      if (containsLocation(locations[parentIdx], loc) == false) {
+        newLocs.push(loc);
+        parentIdx = i;
+      }
+    }
   }
-
-  const contains =
-    locations.filter(a => containsLocation(a, location)).length > 0;
-
-  if (!contains) {
-    locations.push(location);
-  }
-
-  return locations;
+  return newLocs;
 }
 
 /**
@@ -83,13 +83,12 @@ function getOutOfScopeLocations(
 ): AstLocation[] {
   const { functions, comments } = findSymbols(source);
   const commentLocations = comments.map(c => c.location);
-
-  return functions
+  const locations = functions
     .map(getLocation)
     .concat(commentLocations)
-    .filter(loc => !containsPosition(loc, position))
-    .reduce(removeOverlaps, [])
-    .sort(sortByStart);
+    .sort(sortByStart)
+    .filter(loc => !containsPosition(loc, position));
+  return removeOverlaps(locations);
 }
 
 export default getOutOfScopeLocations;
--- a/src/workers/parser/getOutOfScopeLocations.js
+++ b/src/workers/parser/getOutOfScopeLocations.js
@@ -8,6 +8,8 @@ import type { Source } from "debugger-html";
 import type { AstLocation, AstPosition } from "./types";
 
 import get from "lodash/fp/get";
+import findIndex from "lodash/findIndex";
+import findLastIndex from "lodash/findLastIndex";
 
 import { containsLocation, containsPosition } from "./utils/contains";
 
@@ -38,6 +40,42 @@ function getLocation(func) {
 }
 
 /**
+ * Find the nearest location containing the input position and
+ * return new locations without inner locations under that nearest location
+ *
+ * @param locations Notice! The locations MUST be sorted by `sortByStart`
+ *                  so that we can do linear time complexity operation.
+ */
+function removeInnerLocations(locations: AstLocation[], position: AstPosition) {
+  // First, let's find the nearest position-enclosing function location,
+  // which is to find the last location enclosing the position.
+  const newLocs = locations.slice();
+  const parentIndex = findLastIndex(newLocs, loc =>
+    containsPosition(loc, position)
+  );
+  if (parentIndex < 0) {
+    return newLocs;
+  }
+
+  // Second, from the nearest location, loop locations again, stop looping
+  // once seeing the 1st location not enclosed by the nearest location
+  // to find the last inner locations inside the nearest location.
+  const innerStartIndex = parentIndex + 1;
+  const parentLoc = newLocs[parentIndex];
+  const outerBoundaryIndex = findIndex(
+    newLocs,
+    loc => !containsLocation(parentLoc, loc),
+    innerStartIndex
+  );
+  const innerBoundaryIndex =
+    outerBoundaryIndex < 0 ? newLocs.length - 1 : outerBoundaryIndex - 1;
+
+  // Third, remove those inner functions
+  newLocs.splice(innerStartIndex, innerBoundaryIndex - parentIndex);
+  return newLocs;
+}
+
+/**
  * Return an new locations array which excludes
  * items that are completely enclosed by another location in the input locations
  *
@@ -45,19 +83,19 @@ function getLocation(func) {
  *                  so that we can do linear time complexity operation.
  */
 function removeOverlaps(locations: AstLocation[]) {
-  const newLocs = [];
-  const length = locations.length;
-  if (length) {
-    newLocs.push(locations[0]);
-    for (let i = 1, parentIdx = 0; i < length; i++) {
-      const loc = locations[i];
-      if (containsLocation(locations[parentIdx], loc) == false) {
-        newLocs.push(loc);
-        parentIdx = i;
-      }
-    }
+  if (locations.length == 0) {
+    return [];
   }
-  return newLocs;
+  const firstParent = locations[0];
+  return locations.reduce(deduplicateNode, [firstParent]);
+}
+
+function deduplicateNode(nodes, location) {
+  const parent = nodes[nodes.length - 1];
+  if (!containsLocation(parent, location)) {
+    nodes.push(location);
+  }
+  return nodes;
 }
 
 /**
@@ -83,11 +121,15 @@ function getOutOfScopeLocations(
 ): AstLocation[] {
   const { functions, comments } = findSymbols(source);
   const commentLocations = comments.map(c => c.location);
-  const locations = functions
+  let locations = functions
     .map(getLocation)
     .concat(commentLocations)
-    .sort(sortByStart)
-    .filter(loc => !containsPosition(loc, position));
+    .sort(sortByStart);
+  // Must remove inner locations then filter, otherwise,
+  // we will mis-judge in-scope inner locations as out of scope.
+  locations = removeInnerLocations(locations, position).filter(
+    loc => !containsPosition(loc, position)
+  );
   return removeOverlaps(locations);
 }
 
--- a/src/workers/parser/tests/__snapshots__/getOutOfScopeLocations.spec.js.snap
+++ b/src/workers/parser/tests/__snapshots__/getOutOfScopeLocations.spec.js.snap
@@ -7,7 +7,8 @@ exports[`Parser.getOutOfScopeLocations should exclude function for locations on 
 (35, 20) -> (37, 1)
 (39, 26) -> (41, 1)
 (43, 20) -> (45, 1)
-(47, 31) -> (49, 1)"
+(47, 31) -> (49, 1)
+(51, 19) -> (62, 1)"
 `;
 
 exports[`Parser.getOutOfScopeLocations should exclude non-enclosing function blocks 1`] = `
@@ -21,20 +22,26 @@ exports[`Parser.getOutOfScopeLocations should exclude non-enclosing function blo
 (35, 20) -> (37, 1)
 (39, 26) -> (41, 1)
 (43, 20) -> (45, 1)
-(47, 31) -> (49, 1)"
+(47, 31) -> (49, 1)
+(51, 19) -> (62, 1)"
 `;
 
-exports[`Parser.getOutOfScopeLocations should roll up function blocks 1`] = `
+exports[`Parser.getOutOfScopeLocations should not exclude in-scope inner locations 1`] = `
 "(1, 0) -> (1, 16)
-(4, 16) -> (6, 3)
-(8, 16) -> (10, 3)
-(12, 22) -> (14, 3)
-(16, 16) -> (18, 3)
-(20, 27) -> (22, 3)
-(24, 9) -> (26, 3)
+(3, 14) -> (27, 1)
 (29, 16) -> (33, 1)
 (35, 20) -> (37, 1)
 (39, 26) -> (41, 1)
 (43, 20) -> (45, 1)
 (47, 31) -> (49, 1)"
+`;
+
+exports[`Parser.getOutOfScopeLocations should roll up function blocks 1`] = `
+"(1, 0) -> (1, 16)
+(29, 16) -> (33, 1)
+(35, 20) -> (37, 1)
+(39, 26) -> (41, 1)
+(43, 20) -> (45, 1)
+(47, 31) -> (49, 1)
+(51, 19) -> (62, 1)"
 `;
--- a/src/workers/parser/tests/fixtures/outOfScope.js
+++ b/src/workers/parser/tests/fixtures/outOfScope.js
@@ -47,3 +47,16 @@ globalAssignment = (function() {
 const globalIifeDeclaration = (function() {
   const x = 1;
 })();
+
+function parentFunc() {
+  let MAX = 3;
+  let nums = [0, 1, 2, 3];
+  let x = 1;
+  let y = nums.find(function(n) {
+    return n == x;
+  });
+  function innerFunc(a) {
+    return Math.max(a, MAX);
+  }
+  return innerFunc(y);
+}
--- a/src/workers/parser/tests/getOutOfScopeLocations.spec.js
+++ b/src/workers/parser/tests/getOutOfScopeLocations.spec.js
@@ -51,4 +51,12 @@ describe("Parser.getOutOfScopeLocations", () => {
       { end: { column: 15, line: 1 }, start: { column: 0, line: 1 } }
     ]);
   });
+
+  it("should not exclude in-scope inner locations", () => {
+    const actual = getOutOfScopeLocations(getSource("outOfScope"), {
+      line: 61,
+      column: 0
+    });
+    expect(formatLines(actual)).toMatchSnapshot();
+  });
 });


Fischer-L:
Waiting the test running...

Fischer-L:
@jasonLaster 
I need to investigate some mochitest error but didn't find how to run it locally at [1].
That would be great you could show me how to do that at local, thanks

[1] https://github.com/devtools-html/debugger.html/blob/master/docs/local-development.md#testing

jasonLaster:
check out our docs https://github.com/devtools-html/debugger.html/blob/master/docs/mochitests.md

ryanjduffy:
@Fischer-L I don't think including inner scopes of the current scope is the right approach here. By doing so, we've brought the inner function body into scope when it isn't. I'd prefer to convert our out of scope calculation to respect column *and* line rather than just line. We're already receiving this information from the parser but it's been simplified to lines by the {{linesInScope}} selector.

@jasonLaster - Thoughts?

jasonLaster:
@ryanjduffy i agree that it would be nicer to out of scope lines & columns. I have not looked at this solution though :) so can't speak to what is going on here

Fischer-L:
> @Fischer-L I don't think including inner scopes of the current scope is the right approach here. By doing so, we've brought the inner function body into scope when it isn't. I'd prefer to convert our out of scope calculation to respect column and line rather than just line. We're already receiving this information from the parser but it's been simplified to lines by the {{linesInScope}} selector.

OK, this leads questions to discuss:
1. Is it ok to display the closure variable referenced both from outer and inner scope? With the current approach, will be able to see the closure variable both referenced but unable to see local ones (as `undefined`) Chrome devtool is doing the similar, please see the below gifs. It is, for sure, wrong to see the local variable's value, which at least should be `undefined` to the outside. 

uuwa6ghyuj

uuwa6ghy

To allow developers to peek into these closure variables is like providing a convenient shortcut during debugging. Yes, we could be strict on this but just would lose some friendly ability for developers.
```js
function doA(obj) {
   let result = obj.array && obj.array.map( v => {    // bp
        // convenient to be able to preview `a_closure_var_a_bit_away` on mouse over
        return v + a_closure_var_a_bit_away.base;
   });
   return result
}
``` 

2. Respecting the column is correct I think so too. For doing that, we have to rethink the current approach deciding what is in-scope. A bp only tells us which line we are stop at. We have to decide in/out scope on mouse over so for the below case during counting our-of-scope locations, we only see 2 locations: Line10 -> 14 and Line 11 -> 12 plus one bp at Line 13
And we should not rule out Line 11 -> 12 since there are in-scope codes on Line 11.
Line 11 is *potentially* in-scope.
```js
function doB(a, b) {                                       // Line 10
   let v = a > 0 ? a : nums.find(v => (v <= b &&           // Line 11
                                 v >= 0 && v % 2 == 0));   // Line 12
   return v * 2;                                           // Line 13, bp
}                                                          // Line 14
```

3. Combine (1) and (2). Assume we now respect column, do we still want to provide the convenient shortcut to closure variables to developers?

jasonLaster:
Thanks for the write up! Great points, and i agree that hovering on an inner scope should work. Even though it technically is out of scope...

> A bp only tells us which line we are stop at. 

Fortunately pause data has the currently line & column. I suggest looking at the selected frame because this has to work regardless of which frame is selected.

Thanks for the hard work here, it's a super interesting problem!

ryanjduffy:
Is the line/column discussion a red herring here, at least partially? The question seems to be "Is the hovered token in scope?" rather than "Is the location in scope?". To your point:

> To allow developers to peek into these closure variables is like providing a convenient shortcut during debugging.

It make sense to be able to preview a variable that is in the current scope even if it resides in a different scope. So, in the following example, if I pause on `return c()`, I should be able to preview `b` on either line 2 or line 5 but not `d` even though it's also on line 5.

```
function a () {
  let b = 1;
  let c = () => {
    let d = 2;
    return b + d;
  }
  return c();
}
```

src/workers/parser/getOutOfScopeLocations.js
```diff
@@ -38,26 +38,26 @@ function getLocation(func) {
...
+      const loc = locations[i];
+      if (containsLocation(locations[parentIdx], loc) == false) {
+        newLocs.push(loc);
+        parentIdx = i;
+      }
+    }
   }
-
-  const contains =
-    locations.filter(a => containsLocation(a, location)).length > 0;
```
Fischer-L: This is O(n) complexity.
Consider there are 10 locations to test: [ L1, L2, ..., L10 ] and only L7 contains L8 ~ L10.
Loop 1: we get [ L1 ]
Loop 2: we test L2 against [ L1 ]
Loop 3: we test L3 against [ L1, L2 ]
... ...
Loop 8: we test L8 against [ L1, .., L7 ] 
Loop 9: we test L9 against [ L1, .., L7 ] 
Loop 10: we test L10 against [ L1, .., L7 ]

src/workers/parser/getOutOfScopeLocations.js
```diff
@@ -83,13 +83,12 @@ function getOutOfScopeLocations(
...
-  return functions
+  const locations = functions
     .map(getLocation)
     .concat(commentLocations)
-    .filter(loc => !containsPosition(loc, position))
-    .reduce(removeOverlaps, [])
-    .sort(sortByStart);
+    .sort(sortByStart)
+    .filter(loc => !containsPosition(loc, position));
+  return removeOverlaps(locations);
```
Fischer-L: `sort` should be O(nlogn). The old `removeOverlaps` is O(n^2) so the old way was doing O(n^2) then O(nlogn).
Try to improve that, just do once O(nlogn) then the following operations can do O(n)

src/workers/parser/tests/__snapshots__/getOutOfScopeLocations.spec.js.snap
```diff
@@ -26,12 +26,6 @@ exports[`Parser.getOutOfScopeLocations should exclude non-enclosing function blo
 
 exports[`Parser.getOutOfScopeLocations should roll up function blocks 1`] = `
 "(1, 0) -> (1, 16)
-(4, 16) -> (6, 3)
-(8, 16) -> (10, 3)
-(12, 22) -> (14, 3)
-(16, 16) -> (18, 3)
-(20, 27) -> (22, 3)
-(24, 9) -> (26, 3)
```
Fischer-L: These are in fact the inner locations under the location enclosing the bp so shouldn't be out of scope

src/workers/parser/utils/contains.js
```diff
@@ -7,7 +7,7 @@
 import type { AstLocation, AstPosition } from "../types";
 import type { Node } from "babel-traverse";
 
-function startsBefore(a: AstLocation, b: AstPosition) {
+export function startsBefore(a: AstLocation, b: AstPosition) {
```
Fischer-L: Reuse the function we did in PR #4381 we can do some optimization in `removeInnerLocations`

src/workers/parser/getOutOfScopeLocations.js
```diff
@@ -38,26 +42,72 @@ function getLocation(func) {
...
+      nearestPos = i;
+    } else if (startsBefore(loc, position) == false) {
+      // Optimization: If a location dosen't even start before the position,
+      // then, all locations following won't contain the position so early break
+      break;
+    }
   }
 
-  const contains =
-    locations.filter(a => containsLocation(a, location)).length > 0;
```
Fischer-L: This is O(n^2) complexity.
Consider there are 10 locations to test: [ L1, L2, ..., L10 ] and only L7 contains L8 ~ L10.
Loop 1: we get [ L1 ]
Loop 2: we test L2 against [ L1 ]
Loop 3: we test L3 against [ L1, L2 ]
... ...
Loop 8: we test L8 against [ L1, .., L7 ]
Loop 9: we test L9 against [ L1, .., L7 ]
Loop 10: we test L10 against [ L1, .., L7 ]

src/workers/parser/getOutOfScopeLocations.js
```diff
@@ -83,13 +133,16 @@ function getOutOfScopeLocations(
...
     .concat(commentLocations)
-    .filter(loc => !containsPosition(loc, position))
-    .reduce(removeOverlaps, [])
     .sort(sortByStart);
+  // Must remove inner locations then filter, otherwise,
+  // we will mis-judge in-scope inner locations as out of scope.
+  locations = removeInnerLocations(locations, position).filter(
+    loc => !containsPosition(loc, position)
+  );
+  return removeOverlaps(locations);
```
Fischer-L: `sort` should be O(nlogn). The old `removeOverlaps` is O(n^2) so the old way was doing O(n^2) then O(nlogn).
Try to improve that, just do once O(nlogn) then the following operations can do O(n)

Fischer-L:
@jasonLaster Would you have a look the patch, thanks
Now we can handle inner locations under the bp-enclosing location.

In the below test, I also try to step into the inner location. We can see the outer `this` which should be unknown to the inner is not revealed (thanks to the correct resolution on call stack)
4459_solution

##### More to discuss:
- This patch should help #1574 somehow and the regression on #4369 because now we start handling inner locations

- The frame info. Plus the frame info we should be able to do early return at [1] and not even show `undefined`. That requires some thoughts on the current structure  and handling on locations as well so I chose to do it in the future. With this approach at least we are improved and not worse-off.

[1] https://github.com/devtools-html/debugger.html/blob/e12186b556563e67ee8c1dd37094a1c1be5042bb/src/utils/editor/expression.js#L61

src/workers/parser/getOutOfScopeLocations.js
Fischer-L: `sort` should be O(nlogn). The old `removeOverlaps` is O(n^2) so the old way was doing O(n^2) then O(nlogn).
Try to improve that, just do once O(nlogn) then the following operations can do O(n)

jasonLaster:
So... this is a bit hard to follow, looking at our test coverage and the fixture: https://github.com/Fischer-L/debugger.html/blob/4459-mis-judge-in-scope-inner-locations/src/workers/parser/tests/fixtures/outOfScope.js#L1

I wonder if it would help to have new tests that help cover these cases

jasonLaster:
I think it is a bit strange to talk about `bp-enclosing` functions as opposed to just enclosing function. We want this to be generic

src/workers/parser/getOutOfScopeLocations.js
```diff
@@ -38,26 +42,72 @@ function getLocation(func) {
...
-  locations: AstLocation | AstLocation[],
-  location: AstLocation
-) {
-  // support reducing without an initializing array
-  if (!Array.isArray(locations)) {
-    locations = [locations];
+function removeInnerLocations(locations: AstLocation[], position: AstPosition) {
+  // First, let's find the nearest bp-enclosing function location,
+  // which is to loop locations to find the last location enclosing the bp.
+  let nearestPos = -1;
```
jasonLaster: I think we can replace this logic w/ a lodash helper:

this attempt is likely flawed, but you should see what I have in mind :P 

```js
const firstInnerLocation = findIndex(locations,  loc => containsPosition(loc, position)
```

src/workers/parser/getOutOfScopeLocations.js
```diff
@@ -38,26 +42,72 @@ function getLocation(func) {
...
+    const parentLoc = locations[nearestPos];
+    for (let i = nearestPos + 1; i < locations.length; i++) {
+      const loc = locations[i];
+      if (containsLocation(parentLoc, loc)) {
+        lastInnerPos = i;
+      } else {
+        break;
+      }
+    }
+  }
```
jasonLaster: can we replace this with a lodash helper?

src/workers/parser/getOutOfScopeLocations.js
jasonLaster: Nice!

src/workers/parser/getOutOfScopeLocations.js
Fischer-L: @jasonLaster I was trying to import and utilize lodash. It works fine on the local page but while running the mochi test it fails because
```
JavaScript error: , line 0: uncaught exception: TypeError: _lodash is undefined
  JS Error: 
    "uncaught exception: TypeError: _lodash is undefined"]
```
Tried to search "lodash" under the mochi test folder bug found nothing to set. Did I miss some settings?

jasonLaster: oh yeah :) that's tricky. I forgot that this is running in a worker

so the typical way to use lodash is `import {x} from "lodash"` in the worker you need to do `import x from "lodash/x"` because we can't require lodash at runtime in the worker...

Fischer-L:
> So... this is a bit hard to follow, looking at our test coverage and the fixture: https://github.com/Fischer-L/debugger.html/blob/4459-mis-judge-in-scope-inner-locations/src/workers/parser/tests/fixtures/outOfScope.js#L1
> I wonder if it would help to have new tests that help cover these cases

Original wanted to reuse the `outer` function. Later because I wanted to make some explicit and might-be-seen case, added another test function for that.

> I think it is a bit strange to talk about bp-enclosing functions as opposed to just enclosing function. We want this to be generic and sometimes you're stepping and the original BP is far far away :)

Removed the bp wording in comments

p.s: during the local mochi test, saw the below error. At the 1st look this shouldn't relates to us here, wanna see the result on the test result
```
 TEST-UNEXPECTED-FAIL browser_dbg-chrome-debugging.js:23 
 TypeError: DebuggerServer.addBrowserActors is not a function
     initDebuggerClient browser/devtools/client/debugger/new/test/mochitest/browser_dbg-chrome-debugging.js 23:5
      browser/devtools/client/debugger/new/test/mochitest/browser_dbg-chrome-debugging.js 72:13
     Async*Tester_execTest/< browser-test.js 1065:21
     Tester_execTest browser-test.js 1056:9
     Tester.prototype.nextTest</< browser-test.js 956:9
     SimpleTest.waitForFocus/waitForFocusInner/focusedOrLoaded/< tests/SimpleTest/SimpleTest.js 795:59
```

Fischer-L:
@jasonLaster The test looks fine. Would you look at the patch, thanks?

src/workers/parser/getOutOfScopeLocations.js
```diff
@@ -38,26 +40,63 @@ function getLocation(func) {
...
+  const length = locations.length;
+  if (length) {
+    newLocs.push(locations[0]);
+    for (let i = 1, parentIdx = 0; i < length; i++) {
+      const loc = locations[i];
+      if (containsLocation(locations[parentIdx], loc) == false) {
+        newLocs.push(loc);
+        parentIdx = i;
+      }
+    }
```
jasonLaster: can we use `findIndex` here too?

src/workers/parser/tests/__snapshots__/getOutOfScopeLocations.spec.js.snap
```diff
@@ -21,20 +22,26 @@ exports[`Parser.getOutOfScopeLocations should exclude non-enclosing function blo
...
+(51, 19) -> (62, 1)"
+`;
+
+exports[`Parser.getOutOfScopeLocations should not exclude in-scope inner locations 1`] = `
+"(1, 0) -> (1, 16)
+(3, 14) -> (27, 1)
+(29, 16) -> (33, 1)
+(35, 20) -> (37, 1)
+(39, 26) -> (41, 1)
+(43, 20) -> (45, 1)
```
jasonLaster: This might be a crazy idea, but i think it would be fun to write a helper function for our tests so that we can visualize what is in scope and what is out of scope.

for instance, we could write `expect(highlightInScopeLocations(..)).toMatchSnapshot()` and the locations in the code that are inscope are bolded with https://github.com/chalk/chalk

This would make it really fun to debug this code because you could quickly visualize what the algorithm outputs!

```
function parentFunc() {
  let MAX = 3;
  let nums = [0, 1, 2, 3];
  let x = 1;
  let y = nums.find(function(n) {
    return n == x;
  });
  function innerFunc(a) {
    return Math.max(a, MAX);
  }
  return innerFunc(y);
}
```

src/workers/parser/getOutOfScopeLocations.js
```diff
@@ -38,26 +40,63 @@ function getLocation(func) {
...
+  let lastInnerPos = -1;
+  if (nearestPos >= 0) {
+    let parentLoc = locations[nearestPos];
+    let firstNotContained = findIndex(
+      locations,
+      loc => !containsLocation(parentLoc, loc),
+      nearestPos + 1
+    );
+    lastInnerPos =
+      firstNotContained < 0 ? locations.length - 1 : firstNotContained - 1;
```
jasonLaster: lets move this to a helper function... that returns `lastInnerPos`

src/workers/parser/getOutOfScopeLocations.js
```diff
@@ -38,26 +40,63 @@ function getLocation(func) {
...
-) {
-  // support reducing without an initializing array
-  if (!Array.isArray(locations)) {
-    locations = [locations];
+function removeInnerLocations(locations: AstLocation[], position: AstPosition) {
+  // First, let's find the nearest position-enclosing function location,
+  // which is to loop locations to find the last location enclosing the bp.
+  let nearestPos = findLastIndex(locations, loc =>
+    containsPosition(loc, position)
+  );
```
jasonLaster: lets move this out to a helper function too with the comments above the function

src/workers/parser/tests/__snapshots__/getOutOfScopeLocations.spec.js.snap
Fischer-L: @jasonLaster 

Do you mean printing out in-scope function in tests/fixtures/outOfScope.js like this?
``` 
// This parentFunc is in-scope so from here codes are in green
function parentFunc() {
  let MAX = 3;
  let nums = [0, 1, 2, 3];
  let x = 1;
  let y = nums.find(function(n) {
    return n == x;
  });
  function innerFunc(a) {
    return Math.max(a, MAX);
  }
  return innerFunc(y);
}
// to here codes are in green
```

src/workers/parser/getOutOfScopeLocations.js
```diff
@@ -38,26 +40,63 @@ function getLocation(func) {
...
+ * items that are completely enclosed by another location in the input locations
+ *
+ * @param locations Notice! The locations MUST be sorted by `sortByStart`
+ *                  so that we can do linear time complexity operation.
+ */
+function removeOverlaps(locations: AstLocation[]) {
+  const newLocs = [];
+  const length = locations.length;
+  if (length) {
+    newLocs.push(locations[0]);
```
codehag: this is easier to read as a reduce:

```javascript
function removeOverlappingNodes(locations: AstLocation[]) {
  const firstParent = locations[0];
  locations.reduce((parents, location) => {
    const parent = parents[parents.length - 1];
    if (containsLocation(parent, location)) {
      return [ ...parents, location ];
    }
    return parents;
  }, [ firstParent ])
```

src/workers/parser/getOutOfScopeLocations.js
```diff
@@ -38,26 +40,63 @@ function getLocation(func) {
...
+      firstNotContained < 0 ? locations.length - 1 : firstNotContained - 1;
   }
 
-  const contains =
-    locations.filter(a => containsLocation(a, location)).length > 0;
-
-  if (!contains) {
-    locations.push(location);
+  // Third, remove those inner functions
+  const newLocs = locations.slice();
```
codehag: why does locations need to be duplicated?

src/workers/parser/getOutOfScopeLocations.js
```diff
@@ -38,26 +40,63 @@ function getLocation(func) {
...
+    containsPosition(loc, position)
+  );
+
+  // Second, from the nearest location, loop locations again, stop looping
+  // once seeing the 1st location not enclosed by the nearest location
+  // to find the last inner locations inside the nearest location.
+  let lastInnerPos = -1;
+  if (nearestPos >= 0) {
+    let parentLoc = locations[nearestPos];
+    let firstNotContained = findIndex(
```
codehag: the naming here is confusing. maybe something like `outerBoundaryIndex`?

src/workers/parser/getOutOfScopeLocations.js
```diff
@@ -38,26 +40,63 @@ function getLocation(func) {
...
-  locations: AstLocation | AstLocation[],
-  location: AstLocation
-) {
-  // support reducing without an initializing array
-  if (!Array.isArray(locations)) {
-    locations = [locations];
+function removeInnerLocations(locations: AstLocation[], position: AstPosition) {
+  // First, let's find the nearest position-enclosing function location,
+  // which is to loop locations to find the last location enclosing the bp.
+  let nearestPos = findLastIndex(locations, loc =>
```
codehag: this naming is a bit confusing, how about `parentIndex` or `outerStartIndex` (because nearest can also be adjacent, and Pos suggests Position similar to the ASTPosition, but this is an index. more specifically it is a boundary)

also we use it twice with +1; it makes sense to have that as a variable, maybe

```javascript
const outerStartIndex = findLastIndex(locations, loc => containsPosition(loc, position));
const innerStartIndex = outerStartIndex + 1;
```

src/workers/parser/getOutOfScopeLocations.js
```diff
@@ -38,26 +40,63 @@ function getLocation(func) {
...
+    containsPosition(loc, position)
+  );
+
+  // Second, from the nearest location, loop locations again, stop looping
+  // once seeing the 1st location not enclosed by the nearest location
+  // to find the last inner locations inside the nearest location.
+  let lastInnerPos = -1;
+  if (nearestPos >= 0) {
```
codehag: Sometimes doing a positive check is making the code harder to read and adds unnecessary complexity. For this reason we usually prefer doing negative checks, because then we can return early. it also saves some work, in this case the memory allocation that is happening on 72 and also the other checks. You also wouldnt need to reassign `lastInnerPos` which is making this a bit more confusing to read. 

Usually you want to use `if` statements sparingly. they force a reader to remember things, which is always extra mental load.

So maybe something like this would be a good idea:

```javascript
if (nearestPos < 0) {
  return locations;
}
```

src/workers/parser/getOutOfScopeLocations.js
```diff
@@ -38,26 +40,63 @@ function getLocation(func) {
...
+    let parentLoc = locations[nearestPos];
+    let firstNotContained = findIndex(
+      locations,
+      loc => !containsLocation(parentLoc, loc),
+      nearestPos + 1
+    );
+    lastInnerPos =
```
codehag: since this isnt a position but an index. maybe `innerBoundaryIndex` is clearer?

src/workers/parser/getOutOfScopeLocations.js
```diff
@@ -38,26 +40,63 @@ function getLocation(func) {
...
   }
 
-  const contains =
-    locations.filter(a => containsLocation(a, location)).length > 0;
-
-  if (!contains) {
-    locations.push(location);
+  // Third, remove those inner functions
+  const newLocs = locations.slice();
+  if (nearestPos >= 0 && lastInnerPos >= 0) {
```
codehag: i dont think it is possible for lastInnerPos to be -1, since you have a check on line 68

src/workers/parser/getOutOfScopeLocations.js
```diff
@@ -38,26 +40,62 @@ function getLocation(func) {
...
-  locations: AstLocation | AstLocation[],
-  location: AstLocation
-) {
-  // support reducing without an initializing array
-  if (!Array.isArray(locations)) {
-    locations = [locations];
+function removeInnerLocations(locations: AstLocation[], position: AstPosition) {
+  // First, let's find the nearest position-enclosing function location,
+  // which is to loop locations to find the last location enclosing the position.
+  const newLocs = locations.slice();
```
Fischer-L: The reason returning new locations is because:
1. Both `removeOverlaps` and Array `filter` return new locations. Since `removeInnerLocations` also does removal operation, that is for consistency, not to bring surprise. 
2. Although this is not an exported function, for the removal operation probably less surprise without mutating the input arg. This is not necessary, however, consider the reason 1, consistency is good for less error.

src/workers/parser/getOutOfScopeLocations.js
```diff
@@ -38,26 +40,62 @@ function getLocation(func) {
...
-    locations.push(location);
+/**
+ * Return an new locations array which excludes
+ * items that are completely enclosed by another location in the input locations
+ *
+ * @param locations Notice! The locations MUST be sorted by `sortByStart`
+ *                  so that we can do linear time complexity operation.
+ */
+function removeOverlaps(locations: AstLocation[]) {
+  if (locations.length == 0) {
```
Fischer-L: Important to take care of an empty array. Otherwise
```js
// firstParent will be undefined if locations are empty
const firstParent = locations[0]; 
````

Fischer-L:
@codehag Updated per comments, thanks

src/workers/parser/tests/__snapshots__/getOutOfScopeLocations.spec.js.snap
jasonLaster: Yep, i might hack on this now that it has landed :)

Issue #4171, opened 2017-09-26
[PrettyPrint] auto-pretty-print minified sources


jasonLaster:
The old debugger would automatically pretty print minified sources. This was a nice feature, which removed an extra "pretty print" action.

codehag:
This could be done via an action on select source?

jasonLaster:
The way i see it working is, when we load the source text we can check for 
a: min.js, 
b: a line w/ > 1000 chars

and if either is true we can fire the `togglePrettyPrint` action

bomsy:
There is an `isMinified` function which the old debugger uses as heuristics  to decide, we could use that ..
https://searchfox.org/mozilla-central/source/devtools/client/debugger/utils.js#74

jasonLaster:
lets steal it:

```js
 function (key, text) {
    if (this._minifiedCache.has(key)) {
      return this._minifiedCache.get(key);
    }

    let isMinified;
    let lineEndIndex = 0;
    let lineStartIndex = 0;
    let lines = 0;
    let indentCount = 0;
    let overCharLimit = false;

    // Strip comments.
    text = text.replace(/\/\*[\S\s]*?\*\/|\/\/(.+|\n)/g, "");

    while (lines++ < SAMPLE_SIZE) {
      lineEndIndex = text.indexOf("\n", lineStartIndex);
      if (lineEndIndex == -1) {
        break;
      }
      if (/^\s+/.test(text.slice(lineStartIndex, lineEndIndex))) {
        indentCount++;
      }
      // For files with no indents but are not minified.
      if ((lineEndIndex - lineStartIndex) > CHARACTER_LIMIT) {
        overCharLimit = true;
        break;
      }
      lineStartIndex = lineEndIndex + 1;
    }

    isMinified =
      ((indentCount / lines) * 100) < INDENT_COUNT_THRESHOLD || overCharLimit;

    this._minifiedCache.set(key, isMinified);
    return isMinified;
  }
  ```

peterschussheim:
@jasonLaster can I take this one?

jasonLaster:
sure, it's yours

jasonLaster:
:wave: @peterschussheim, just checking in to see how it's going and if you have any questions.

Also, if you're currently busy, feel free to `/unclaim` the issue. You can always `/claim` another one later. And, this is OSS, we're guilt free!

Also, always happy to chat on [slack] or in a 'hangout!

[...]

PR #4700, opened 2017-11-15
[PrettyPrint] auto-pretty-print minified sources (#4171)

assafdd:
Associated Issue: #4171

### Summary of Changes
* copied isMinified heuristic from old debugger
* auto togglePrettyPrint on setText 

### Test Plan:
run tests on 
1. New Tab - about:home (some react  resources under vendor folder) 
2. http://devtools-html.github.io/debugger-examples/examples/todomvc/ (bower_components/backbone/local-storage folder)

One odd is that if you close  pretty-print source while it's selected it will reappear while if you close it while the original version is visible it won't.

### DIFF
--- a/src/actions/sources.js
+++ b/src/actions/sources.js
@@ -25,6 +25,7 @@ import { loadSourceText } from "./sources/loadSourceText";
 import { prefs } from "../utils/prefs";
 import { removeDocument } from "../utils/editor";
 import { isThirdParty, isMinified } from "../utils/source";
+import { isOriginalId } from "devtools-source-map";
 import { getGeneratedLocation } from "../utils/source-maps";
 
 import {
@@ -208,7 +209,7 @@ export function selectSource(id: string, options: SelectSourceOptions = {}) {
         await dispatch(loadSourceText(source.toJS()));
         await dispatch(setOutOfScopeLocations());
         const src = getSource(getState(), id).toJS();
-        if (src.text && isMinified(src.id, src.text)) {
+        if (src.text && !isOriginalId(src.id) && isMinified(src.id, src.text)) {
           await dispatch(togglePrettyPrint(src.id));
         }
       })()
--- a/src/actions/sources.js
+++ b/src/actions/sources.js
@@ -24,7 +24,7 @@ import { loadSourceText } from "./sources/loadSourceText";
 
 import { prefs } from "../utils/prefs";
 import { removeDocument } from "../utils/editor";
-import { isThirdParty } from "../utils/source";
+import { isThirdParty, isMinified } from "../utils/source";
 import { getGeneratedLocation } from "../utils/source-maps";
 
 import {
@@ -207,6 +207,10 @@ export function selectSource(id: string, options: SelectSourceOptions = {}) {
       [PROMISE]: (async () => {
         await dispatch(loadSourceText(source.toJS()));
         await dispatch(setOutOfScopeLocations());
+        const src = getSource(getState(), id).toJS();
+        if (isMinified(src.id, src.text)) {
+          await dispatch(togglePrettyPrint(src.id));
+        }
       })()
     });
   };
--- a/src/actions/sources.js
+++ b/src/actions/sources.js
@@ -24,8 +24,7 @@ import { loadSourceText } from "./sources/loadSourceText";
 
 import { prefs } from "../utils/prefs";
 import { removeDocument } from "../utils/editor";
-import { isThirdParty, isMinified } from "../utils/source";
-import { isOriginalId } from "devtools-source-map";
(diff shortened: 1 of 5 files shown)


assafdd:
the commit also broke Editor.spec.js with to the following exception
  ● Editor › should set text
    TypeError: Cannot read property 'match' of undefined
      at isOriginalId (node_modules/devtools-source-map/src/utils/index.js:19:14)

src/components/Editor/index.js
```diff
@@ -488,6 +492,15 @@ class Editor extends PureComponent<Props, State> {
...
     if (selectedSource) {
+      // auto-prettyprint the source
+      const selected = selectedSource.toJS();
+      if (
+        selected.text &&
+        !isOriginalId(selected.id) &&
+        isMinified(selected.id, selected.text)
+      ) {
+        togglePrettyPrint(selectedSource.get("id"));
+      }
```
jasonLaster: can we do this in `actions/sources#selectSource` we like to have the actions control the business logic

src/components/Editor/index.js
```diff
@@ -224,7 +226,9 @@ class Editor extends PureComponent<Props, State> {
   }
 
   componentWillUpdate(nextProps) {
-    this.setText(nextProps);
+    if (nextProps.selectedSource !== this.props.selectedSource) {
+      this.setText(nextProps);
+    }
```
jasonLaster: why do we need this?

bomsy: It was resetting the text each time, we setEmptylines and other stuff even though the selected source has not changed

assafdd:
Move the login to the action.
@bomsy I don't think the if statement is needed any more. WDYT? 
`if (nextProps.selectedSource !== this.props.selectedSource) {`

assafdd:
Still some jest test are failing on my machine.

jasonLaster:
kill the IF

bomsy:
@assafdd yeah ..not relevant

assafdd:
@jasonLaster I removed the IF, fixed expressions.spec.js 
I can't figures out why createPrettySource.spec.js is breaking and if it's releted to my changes at all.
Any suggestions how to continue?

jasonLaster:
will look

src/actions/sources.js
```diff
@@ -207,6 +207,10 @@ export function selectSource(id: string, options: SelectSourceOptions = {}) {
       [PROMISE]: (async () => {
         await dispatch(loadSourceText(source.toJS()));
         await dispatch(setOutOfScopeLocations());
+        const src = getSource(getState(), id).toJS();
+        if (src.text && isMinified(src.id, src.text)) {
```
bomsy: We should also be checking if this is a generated file or not before auto pretty printing. That's probably the issue with the tests

bomsy: adding  `&&  !isOriginalId(src.id)` should do help

assafdd: thanks @bomsy I'm on it

assafdd: done

bomsy:
tiny lint issues
image

assafdd:
Well, I think it's finally ready :)

src/actions/sources.js
```diff
@@ -207,6 +208,10 @@ export function selectSource(id: string, options: SelectSourceOptions = {}) {
       [PROMISE]: (async () => {
         await dispatch(loadSourceText(source.toJS()));
         await dispatch(setOutOfScopeLocations());
+        const src = getSource(getState(), id).toJS();
+        if (src.text && !isOriginalId(src.id) && isMinified(src.id, src.text)) {
```
bomsy: hey @assafdd,
 just noticed one last tiny thing. we should do add `shouldPrettyPrint(src)`. it'll look something like this
```js
if (src.text && !isOriginalId(src.id) && (isMinified(src.id, src.text) || shouldPrettyPrint(src))) { 
```
 so we make sure all holes are covered.

jasonLaster: could we do `shouldPrettyPrint(src)) && isMinified(src.id, src.text)`

bomsy: That might be to much as there might be files that don't meet the `isMinified` heuristics but we might still want to pretty print if `shouldPrettyPrint(src)` is `true`

assafdd: Yes we can. I move the empty text check to isMinified util. 
Otherwise "should evaluate expressions in specific scope" in expressions.spec.js breaks.

assafdd: @bomsy I tried to play  with shouldPrettyPrint()  and its seems to be truthy on any JS source that I have checked.

bomsy: cool! it's fine, i had a scenario which i found but can't find it any more so its fine

bomsy:
Thanks! @assafdd

PR #4707, opened 2017-11-16
[Outline] Use magenta for param color, fixes #4604

### LINKED ISSUE #4604: [Outline] Change param color to magenta
We're trying to keep red for errors only, so it would be great if we could change the red params in Outline to syntax magenta (var(--theme-highlight-red)).

image

(Ideally it would match the syntax highlighting, but in the code it looks like this is just colored as blue and blue.)

bogas04:
Associated Issue: #4604 

### Summary of Changes

* As suggested in the associated issue #4604, the color of param in Outline is changed.

### Test Plan

- [x] Outline pane in debugger uses magenta color for function param as opposed to red.

### Screenshots

Before  | After
:--|:--
screen shot 2017-11-16 at 8 42 40 pm| screen shot 2017-11-16 at 8 42 47 pm

### DIFF
--- a/src/components/shared/PreviewFunction.css
+++ b/src/components/shared/PreviewFunction.css
@@ -11,7 +11,7 @@
 }
 
 .function-signature .param {
-  color: var(--string-color);
+  color: var(--theme-highlight-red);
 }
 
 .function-signature .paren {


codehag:
Looks great! thanks!

PR #4710, opened 2017-11-16
[Accessibility] command bar buttons accessibility

### LINKED ISSUE #4078: [Accessibility] command bar buttons accessibility.
Checklist:

- [x]  aria-label and title for each button are redundant. Just title will suffice.
- [x]  Enabled state should be accompanied but "aria-pressed" attribute (pause on exceptions for example) as there's no way for screen reader users to know when it's activated.
- [ ]  Improve focus styling. Right now dotted outline does not line up well with icons.

cvetanovskaa:
Associated Issue: #4078 

### Summary of Changes

* Aria-label from Command Bar Buttons is removed, since it was redundant
* Aria-pressed attribute was added to the PauseOnExceptions button

### Test Plan

1. run yarn start and open the debugger.
2. Right click on the pause on exceptions button and click inspect element
3. Then click the pause on exceptions button at least 3 times and look at the elements and make sure the boolean value change correctly.

### Screenshots

The code of the debug button, after the changes: 
screenshot from 2017-11-14 11-33-05

The code of the PauseOnExceptions button, after the changes:
exceptionsbtn


jasonLaster:
Thanks! this looks good. I'll defer to @Garbee and @yzen for accessibility reviews

yzen:
Looks good from accessibility standpoint. Thanks!

PR #4720, opened 2017-11-17
Add content for profiling

jasonLaster:
bumping to pad 110 to enable auto react perf timings


digitarald:
LGTM, only nit is not putting Chrome over perf.html (or pointing to bugs that need to be addressed to change the ordering 🙏 )

Issue #4685, opened 2017-11-14
[Search][UX] "Results Not Found" before the results


karlcow:
Using Firefox Nightly 59.0a1 (2017-11-13) (64-bit)

1. Go to https://trends.google.com/trends/
2. Open developer tools. 
3. Go to the debugger
4. Search in all files (command + shift + F on macOS) with a string.

Example: backdrop

![capture d ecran 2017-11-14 a 10 15 53](https://user-images.githubusercontent.com/505230/32757890-14017d4a-c925-11e7-85b5-a21ef7749b9d.png)

After a **couple of seconds** (long enough to have the feeling to be waiting but on the cliff of anxiety that they will never come. Cue dramatic music effects).

![capture d ecran 2017-11-14 a 10 17 12](https://user-images.githubusercontent.com/505230/32757905-26002fc8-c925-11e7-844c-5abf939da763.png)

Expected: 
Have something in the UX that the search is going on. 

Actual:
The  Not results found is misleading and encourages to abandon the search.

codehag:
This could be achieved using redux-thunk, and have a "loading" message while we are waiting

varenya:
@jasonLaster I have a fix for this but I am not sure how to style the "loading.." part. are there already any components for this?

codehag:
@varenya we don't have anything yet, what about just using a text for now? we can iterate on it later :)

lukaszsobek:
Fix merged, closing issue.

PR #4726, opened 2017-11-17
added a loading message while search results are being fetched 

varenya:
Associated Issue: #4685

Here's the Pull Request Doc

### Summary of Changes

*  Added a new action creator for updating the status of search result fetching
*  based on the above action, in reducer I have added a new case branch which basically updates the status variable in the reducer.

### Test Plan

Example test plan:

- [x] Command-Shift-F opens the "search in files" box
- [x] Enter any text you want to search 
- [x] Press enter
- [x] A "Loading.." message should come up before showing the results.

### DIFF
--- a/src/actions/project-text-search.js
+++ b/src/actions/project-text-search.js
@@ -35,7 +35,7 @@ export function clearSearchResults() {
   };
 }
 
-export function updateStatus(status: string) {
+export function updateSearchStatus(status: string) {
   return { type: "UPDATE_STATUS", status };
 }
 
@@ -43,7 +43,7 @@ export function searchSources(query: string) {
   return async ({ dispatch, getState }: ThunkArgs) => {
     await dispatch(clearSearchResults());
     await dispatch(addSearchQuery(query));
-    dispatch(updateStatus(statusEnum.fetching));
+    dispatch(updateSearchStatus(statusEnum.fetching));
     await dispatch(loadAllSources());
     const sources = getSources(getState());
     const validSources = sources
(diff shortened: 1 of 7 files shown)


src/actions/project-text-search.js
```diff
@@ -34,17 +35,21 @@ export function clearSearchResults() {
   };
 }
 
+export function updateStatus(status: string) {
```
jasonLaster: hmm how about? `updateSearchStatus`

src/actions/project-text-search.js
```diff
@@ -34,17 +35,21 @@ export function clearSearchResults() {
...
+  return { type: "UPDATE_STATUS", status };
+}
+
 export function searchSources(query: string) {
   return async ({ dispatch, getState }: ThunkArgs) => {
     await dispatch(clearSearchResults());
     await dispatch(addSearchQuery(query));
+    dispatch(updateStatus(statusEnum.fetching));
```
jasonLaster: can we set the status to fetching when we set the query or done when we set the results?

varenya: I thought of that at first but this seemed more explicit than your approach

src/actions/project-text-search.js
varenya: yeah that sounds better will change it

src/reducers/project-text-search.js
```diff
@@ -22,16 +23,28 @@ export type Search = {
   matches: I.List<any>
 };
 
+export const statusEnum = {
```
bomsy: nit: can we use `statusType` instead of `statusEnum` feels more intutive

bomsy:
Great work @varenya

varenya:
Thanks everyone :)

Issue #4524, opened 2017-10-29
Watch Expressions Shouldn't Be Allowed with Syntax Errors


darkwing:
I created a bogus watch expression:

<img width="299" alt="watchsyntaxerror" src="https://user-images.githubusercontent.com/46655/32139971-f9d1a764-bc1e-11e7-8175-b479f1fcc827.png">

If a syntax error is detected for a watch expression, we shouldn't allow its creation -- we should instead highlight the input box red and notify the user that their expression is invalid.

wldcordeiro:
The title was off based on your description.

jasonLaster:
hmm. that's a good point... hadn't thought of that. That'll actually simplify the evaluate code a bit

jasonLaster:
```diff
diff --git a/src/actions/expressions.js b/src/actions/expressions.js
index 85968ea..521e513 100644
--- a/src/actions/expressions.js
+++ b/src/actions/expressions.js
@@ -32,6 +32,11 @@ export function addExpression(input: string) {
       return;
     }
 
+    const error = await parser.hasSyntaxError(input);
+    if (error) {
+      return false;
+    }
+
     const expression = getExpression(getState(), input);
     if (expression) {
       return dispatch(evaluateExpression(expression));
@@ -102,14 +107,6 @@ function evaluateExpression(expression: Expression) {
     }
 
     let input = expression.input;
-    const error = await parser.hasSyntaxError(input);
-    if (error) {
-      return dispatch({
-        type: "EVALUATE_EXPRESSION",
-        input: expression.input,
-        value: { input: expression.input, result: error }
-      });
-    }
 
     const frame = getSelectedFrame(getState());
 
diff --git a/src/components/SecondaryPanes/Expressions.js b/src/components/SecondaryPanes/Expressions.js
index 5362c50..2a604f3 100644
--- a/src/components/SecondaryPanes/Expressions.js
+++ b/src/components/SecondaryPanes/Expressions.js
@@ -41,7 +41,8 @@ class Expressions extends PureComponent<Props, State> {
     super(...args);
 
     this.state = {
-      editing: null
+      editing: null,
+      error: null
     };
 
     this.renderExpression = this.renderExpression.bind(this);
@@ -163,26 +164,34 @@ class Expressions extends PureComponent<Props, State> {
     }
   }
 
-  renderNewExpressionInput() {
-    const onKeyPress = e => {
-      if (e.key !== "Enter") {
-        return;
-      }
-
[...]

PR #4734, opened 2017-11-18
Watch Expressions Shouldn't Be Allowed with Syntax Errors

khal0988:
Associated Issue: #4524

Here's the Pull Request Doc

### Summary of Changes

* If a syntax error is detected for a watch expression, we don't allow its creation -- we instead highlight the input box red and notify the user that their expression is invalid.

### Test Plan

- [ ] Open a source file
- [x] Highlight an expression, right click it, click add watch expression:
         * if no syntax error. expression should be added to list
         * if syntax error. expression isn't added
- [x] Open watch expression panel, type in an expression to be watched:
         * if no syntax error. expression should be added to list
         * if syntax error, text is coloured red, delete text to go back to normal colour.

Feel free to improve it!

### Screenshot
screen shot 2017-12-19 at 5 05 27 pm

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -1,15 +1,8 @@
 [ignore]
-.*node_modules/cjson.*
-.*node_modules/fbjs.*
-.*node_modules/npm.*
-.*node_modules/stylelint.*
-.*node_modules/radium.*
-.*node_modules/config-chain.*
 .*node_modules/documentation.*
 .*node_modules/devtools-reps/src/object-inspector/index.js
 .*node_modules/devtools-mc-assets/.*
 .*node_modules/immutable/dist/immutable.js.flow
-.*node_modules/jest-in-case/index.js
 <PROJECT_ROOT>/firefox/.*
 .*mozilla-central/.*
 .*flow-coverage-report/.*
--- a/assets/panel/debugger.properties
+++ b/assets/panel/debugger.properties
@@ -441,7 +441,7 @@ original=original
 expressions.placeholder=Add watch expression
 # LOCALIZATION NOTE (expressions.errorMsg): Error text for expression
 # input element
-expressions.errorMsg=Invalid expression, try again…
+expressions.errorMsg=Invalid expression…
 expressions.label=Add watch expression
 expressions.accesskey=e
 
--- a/assets/panel/debugger.properties
+++ b/assets/panel/debugger.properties
@@ -439,6 +439,9 @@ original=original
 # LOCALIZATION NOTE (expressions.placeholder): Placeholder text for expression
 # input element
 expressions.placeholder=Add watch expression
+# LOCALIZATION NOTE (expressions.errorMsg): Error text for expression
+# input element
+expressions.errorMsg=Invalid expression, try again…
 expressions.label=Add watch expression
 expressions.accesskey=e
 
--- a/src/actions/expressions.js
+++ b/src/actions/expressions.js
@@ -34,7 +34,7 @@ export function addExpression(input: string) {
 
     const error = await parser.hasSyntaxError(input);
     if (error) {
-      return false;
+      return;
     }
 
     const expression = getExpression(getState(), input);
@@ -48,7 +48,7 @@ export function addExpression(input: string) {
     });
 
     const newExpression = getExpression(getState(), input);
-    dispatch(evaluateExpression(newExpression));
+    return dispatch(evaluateExpression(newExpression));
   };
 }
 
--- a/src/actions/expressions.js
+++ b/src/actions/expressions.js
@@ -34,8 +34,6 @@ export function addExpression(input: string) {
 
     const error = await parser.hasSyntaxError(input);
     if (error) {
-      // for debugging only
-      alert("'" + input + "' " + "has syntax error");
       return false;
     }
 
--- a/src/actions/expressions.js
+++ b/src/actions/expressions.js
@@ -35,7 +35,7 @@ export function addExpression(input: string) {
     const error = await parser.hasSyntaxError(input);
     if (error) {
       // for debugging only
-      console.log("has syntax error");
+      alert("'" + input + "' " + "has syntax error");
       return false;
     }
 
--- a/src/actions/expressions.js
+++ b/src/actions/expressions.js
@@ -7,6 +7,7 @@
 import {
   getExpression,
   getExpressions,
+  getExpressionError,
   getSelectedFrame,
   getSelectedFrameId,
   getSource
@@ -28,13 +29,19 @@ import type { ThunkArgs } from "./types";
  */
 export function addExpression(input: string) {
   return async ({ dispatch, getState }: ThunkArgs) => {
-    if (!input) {
+    // clear previous error.
+    const prevErr = getExpressionError(getState());
+    const error = await parser.hasSyntaxError(input);
+    if (prevErr && !error) {
+      dispatch(setExpressionError(false));
+    }
+
+    if (!input && !prevErr) {
       return;
     }
 
-    const error = await parser.hasSyntaxError(input);
     if (error) {
-      return;
+      return dispatch(setExpressionError());
     }
 
     const expression = getExpression(getState(), input);
@@ -52,16 +59,24 @@ export function addExpression(input: string) {
   };
 }
 
+export function setExpressionError(value: boolean = true) {
+  return { type: "EXPRESSION_ERROR", value };
+}
+
 export function updateExpression(input: string, expression: Expression) {
-  return ({ dispatch, getState }: ThunkArgs) => {
+  return async ({ dispatch, getState }: ThunkArgs) => {
+    const error = await parser.hasSyntaxError(input);
     if (!input || input == expression.input) {
       return;
+    }
+    if (error) {
+      return dispatch(setExpressionError());
     }
 
     dispatch({
       type: "UPDATE_EXPRESSION",
       expression,
-      input: input
+      input
     });
 
     dispatch(evaluateExpressions());
--- a/src/actions/expressions.js
+++ b/src/actions/expressions.js
@@ -28,7 +28,6 @@ import type { ThunkArgs } from "./types";
  */
 export function addExpression(input: string) {
   return async ({ dispatch, getState }: ThunkArgs) => {
-    const expressionError = await parser.hasSyntaxError(input);
     if (!input) {
       return;
     }
@@ -38,6 +37,7 @@ export function addExpression(input: string) {
       return dispatch(evaluateExpression(expression));
     }
 
+    const expressionError = await parser.hasSyntaxError(input);
     dispatch({ type: "ADD_EXPRESSION", input, expressionError });
 
     const newExpression = getExpression(getState(), input);
@@ -53,10 +53,11 @@ export function clearExpressionError() {
 
 export function updateExpression(input: string, expression: Expression) {
   return async ({ dispatch, getState }: ThunkArgs) => {
-    const expressionError = await parser.hasSyntaxError(input);
     if (!input || input == expression.input) {
       return;
     }
+
+    const expressionError = await parser.hasSyntaxError(input);
     dispatch({
       type: "UPDATE_EXPRESSION",
       expression,
--- a/src/actions/expressions.js
+++ b/src/actions/expressions.js
@@ -32,6 +32,13 @@ export function addExpression(input: string) {
       return;
     }
 
+    const error = await parser.hasSyntaxError(input);
+    if (error) {
+      // for debugging only
+      console.log("has syntax error");
+      return false;
+    }
+
     const expression = getExpression(getState(), input);
     if (expression) {
       return dispatch(evaluateExpression(expression));
@@ -102,15 +109,6 @@ function evaluateExpression(expression: Expression) {
     }
 
     let input = expression.input;
-    const error = await parser.hasSyntaxError(input);
-    if (error) {
-      return dispatch({
-        type: "EVALUATE_EXPRESSION",
-        input: expression.input,
-        value: { input: expression.input, result: error }
-      });
-    }
-
     const frame = getSelectedFrame(getState());
 
     if (frame) {
--- a/src/actions/expressions.js
+++ b/src/actions/expressions.js
@@ -7,7 +7,6 @@
 import {
   getExpression,
   getExpressions,
-  getExpressionError,
   getSelectedFrame,
   getSelectedFrameId,
   getSource
@@ -29,19 +28,9 @@ import type { ThunkArgs } from "./types";
  */
 export function addExpression(input: string) {
   return async ({ dispatch, getState }: ThunkArgs) => {
-    // clear previous error.
-    const prevErr = getExpressionError(getState());
-    const error = await parser.hasSyntaxError(input);
-    if (prevErr && !error) {
-      dispatch(setExpressionError(false));
-    }
-
-    if (!input && !prevErr) {
+    const expressionError = await parser.hasSyntaxError(input);
+    if (!input) {
       return;
-    }
-
-    if (error) {
-      return dispatch(setExpressionError());
     }
 
     const expression = getExpression(getState(), input);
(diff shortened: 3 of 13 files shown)


src/components/SecondaryPanes/Frames/index.js
```diff
@@ -204,6 +204,7 @@ class Frames extends Component<Props, State> {
   }
 }
 
+// Should be converted to flow types... what is flow types ?!
```
jasonLaster: check other components for flow props 

https://flow.org/

khal0988: *thumbs up*

src/actions/expressions.js
```diff
@@ -32,6 +32,13 @@ export function addExpression(input: string) {
       return;
     }
 
+    const error = await parser.hasSyntaxError(input);
+    if (error) {
+      // for debugging only
+      alert("'" + input + "' " + "has syntax error");
```
jasonLaster: lol - didn't realize this was a WIP pr

src/components/SecondaryPanes/Expressions.js
```diff
@@ -9,17 +9,19 @@ import { bindActionCreators } from "redux";
 import actions from "../../actions";
 import { getExpressions, getLoadedObjects, getPause } from "../../selectors";
 import { getValue } from "../../utils/expressions";
-
+import * as parser from "../../workers/parser";
```
jasonLaster: lets just import the one function we need `import { hasX`

src/components/SecondaryPanes/Expressions.js
```diff
@@ -174,18 +178,30 @@ class Expressions extends PureComponent<Props, State> {
...
       e.stopPropagation();
-      e.target.value = "";
-      this.props.addExpression(value);
+
+      const error = await parser.hasSyntaxError(value);
+      this.setState({ error });
+
+      if (error) {
+        e.target.value = value;
+        e.target.style.color = "red";
```
jasonLaster: we will want to style it w/ `CSS` so set a class when we're rendering the input that it has an error

src/components/SecondaryPanes/Expressions.js
```diff
@@ -174,18 +178,30 @@ class Expressions extends PureComponent<Props, State> {
...
+      e.persist();
       e.stopPropagation();
-      e.target.value = "";
-      this.props.addExpression(value);
+
+      const error = await parser.hasSyntaxError(value);
+      this.setState({ error });
+
+      if (error) {
+        e.target.value = value;
```
jasonLaster: do we need this?

src/components/SecondaryPanes/Expressions.js
```diff
@@ -174,18 +178,30 @@ class Expressions extends PureComponent<Props, State> {
         return;
       }
 
+      e.persist();
```
jasonLaster: ? never seen this

src/actions/expressions.js
khal0988: Oops! I forgot to delete the comment lol

src/components/SecondaryPanes/Expressions.js
khal0988: sure

src/components/SecondaryPanes/Expressions.js
khal0988: this allows references to the event to be retained, without it the event becomes null after calling hasSyntaxtError

src/components/SecondaryPanes/Expressions.js
khal0988: fixed

codehag:
@khal0988 can you take a look at the CI errors? This looks pretty close

```
yarn lint-js

yarn lint-js v0.28.4
$ eslint *.js "src/**/*.js" --fix
The babel/generator-star-spacing rule is deprecated. Please use the built in generator-star-spacing rule instead.

/home/ubuntu/debugger.html/src/actions/expressions.js
  146:27  error  'parser' is not defined  no-undef
  157:10  error  'parser' is not defined  no-undef

/home/ubuntu/debugger.html/src/components/SecondaryPanes/Expressions.js
  16:8  error  'classnames' is defined but never used  no-unused-vars

/home/ubuntu/debugger.html/src/components/shared/Badge.js
  4:18  warning  'children' is missing in props validation  react/prop-types

✖ 4 problems (3 errors, 1 warning)

yarn flow

yarn flow v0.28.4
$ flow
Launching Flow server for /home/ubuntu/debugger.html
Spawned flow server (pid=19739)
Logs will go to /tmp/flow/zShomezSubuntuzSdebugger.html.log
Error: src/actions/expressions.js:146
146:   const astScopes = await parser.getScopes(generatedLocation);
                               ^^^^^^ identifier `parser`. Could not resolve name
```

codehag:
we have a few tests failing, can you take a look?
```
Test Suites: 2 failed, 69 passed, 71 total
Tests:       7 failed, 429 passed, 436 total
Snapshots:   2 failed, 126 passed, 128 total
```

bomsy:
Looking into landing this ... it being open for a while

bomsy:
image

src/actions/tests/expressions.spec.js
```diff
@@ -60,11 +60,12 @@ describe("expressions", () => {
   it("should evaluate expressions global scope", async () => {
     const { dispatch, getState } = createStore(mockThreadClient);
 
-    dispatch(actions.addExpression("foo"));
-    dispatch(actions.addExpression("bar"));
+    await dispatch(actions.addExpression("foo"));
+    await dispatch(actions.addExpression("bar"));
+
+    expect(selectors.getExpression(getState(), "foo").value).toBe("bla");
+    expect(selectors.getExpression(getState(), "bar").value).toBe("bla");
```
bomsy: @jasonLaster i'll love your a quick glance on these tests. i'm not feeling they do what i think they should

jasonLaster:
Hey @bomsy i think you used a merge to get this branch up to date, that makes it difficult to rebase the commits from master. In the future you should try doing `git rebase master`.

jasonLaster:
I just cleaned it up by 

- creating a new branch that squashed Johnny's 4 initial commits. 
- cherrypicking them on top of master
- and then cherrypicking your touchups

jasonLaster:
Found two things:

1. you can still edit and save a bad expression
  - create a good expression
  - edit it to become bad
  - er: the ui goes red and prevents you from saving
  - ar: you can save it just fine
  
2. when you save a bad expression and lose focus the placeholder goes red
  - we should remove the invalid classname so the placeholder is always normal

jasonLaster:
:thumbsup:

src/components/SecondaryPanes/Expressions.js
```diff
@@ -78,18 +79,31 @@ class Expressions extends PureComponent<Props, State> {
     deleteExpression(expression);
   }
 
-  inputKeyPress(e, expression) {
+  async inputKeyPress(e, expression) {
+    const target = e.target;
+    target.className = "input-expression";
```
wldcordeiro: Is this part needed? I think the reason it exists is to make sure we reset to the non-error state? I ask because it appears to already be set to this value by default [here](https://github.com/khal0988/debugger.html/blob/55c80e90e8c3b2b9ba18a20a23689e44222347b8/src/components/SecondaryPanes/Expressions.js#L114)

src/components/SecondaryPanes/Expressions.js
```diff
@@ -78,18 +79,31 @@ class Expressions extends PureComponent<Props, State> {
...
+
     if (value == "") {
       return;
     }
 
+    const syntaxError = await hasSyntaxError(value);
+    this.setState({ error: syntaxError });
+
+    if (syntaxError) {
+      target.className = "input-expression-error";
```
wldcordeiro: What if instead of editing `className` to `input-expression-error` we used the [classList](https://developer.mozilla.org/en-US/docs/Web/API/Element/classList) API (either the `.add()` or `.toggle()` methods) to add/remove an `error` CSS class so we have

```css
.input-expression
```

and

```css
.input-expression.error
```

src/components/SecondaryPanes/Expressions.js
```diff
@@ -164,18 +178,30 @@ class Expressions extends PureComponent<Props, State> {
   }
 
   renderNewExpressionInput() {
-    const onKeyPress = e => {
+    const onKeyPress = async e => {
+      e.stopPropagation();
+
+      const target = e.target;
+      target.className = "input-expression";
```
wldcordeiro: See previous comment regarding class names.

src/components/SecondaryPanes/Expressions.css
```diff
@@ -12,6 +12,16 @@
   color: var(--theme-body-color);
 }
 
+.input-expression-error {
```
wldcordeiro: If we did something like I suggested we would want

```css
.input-expression.error {
```

wldcordeiro (approved):
There are a few comments regarding how we're setting the error CSS class but I'm fine with following up with that.

assets/panel/debugger.properties
```diff
@@ -439,6 +439,9 @@ original=original
 # LOCALIZATION NOTE (expressions.placeholder): Placeholder text for expression
 # input element
 expressions.placeholder=Add watch expression
+# LOCALIZATION NOTE (expressions.placeholder): Placeholder text for expression
```
flodolo: Comments are not mandatory, but if you add one:
* Update the text to match the new string, don't simply copy over an existing comment. This string is an error message, not a placeholder.
* Make sure that the string ID between parenthesis matches the actual string (`expressions.errorMsg`, not `expressions.placeholder`).

assets/panel/debugger.properties
```diff
@@ -439,6 +439,9 @@ original=original
...
+# input element in error state
+expressions.errorMsg=Invalid expression, try again...
```
flodolo: Always use a single unicode character `…` vs 3 dots `...`

wldcordeiro:
Went through a pretty large change in this [commit](https://github.com/devtools-html/debugger.html/pull/4734/commits/81f7e14803bed91ae3cbf96006843ca7a3bb1992) so this needs some additional review @jasonLaster @bomsy

flodolo (changes requested):

assets/panel/debugger.properties
wldcordeiro: @flodolo bullet 2 was an oversight, but I did bullet one on purpose as both are put in the placeholder attribute. But it makes sense to have it match its purpose.

flodolo (approved):
Thanks, no concerns on the properties part

jasonLaster:
> It's in redux because we check for errors in the action creators for add/update expressions.

could we check twice?

wldcordeiro:
@jasonLaster we could but I don't think I see the point really. We do all the work of evaluation in the action creators already so it seemed like a good home. We'd also need to do some async work in the component to support it since the parse is on a worker.

jasonLaster (commented):
This PR is amazing! I had so much fun reading component and learned a ton of flow + react :) 

Oh and the animation is fun!

One comment on the redux, but lets get it in soon!

src/actions/expressions.js
```diff
@@ -28,10 +29,21 @@ import type { ThunkArgs } from "./types";
  */
 export function addExpression(input: string) {
   return async ({ dispatch, getState }: ThunkArgs) => {
-    if (!input) {
+    // clear previous error.
+    const prevErr = getExpressionError(getState());
+    const error = await parser.hasSyntaxError(input);
+    if (prevErr && !error) {
+      dispatch(setExpressionError(false));
+    }
```
jasonLaster: two things seem off here:

1. a clearExpressionError action could consolidate the logic
2. we could clear the error field in the `ADD_EXPRESSION` reducer 

```js
      case "ADD_EXPRESSION":
        return appendToList(state, ["expressions"], {
          input: action.input,
          value: null,
          updating: true
        })
          .set("error", ???)
+     case "EXPRESSION_ERROR":

```

.flowconfig
```diff
@@ -1,15 +1,8 @@
...
-.*node_modules/cjson.*
-.*node_modules/fbjs.*
-.*node_modules/npm.*
-.*node_modules/stylelint.*
-.*node_modules/radium.*
-.*node_modules/config-chain.*
 .*node_modules/documentation.*
 .*node_modules/devtools-reps/src/object-inspector/index.js
 .*node_modules/devtools-mc-assets/.*
 .*node_modules/immutable/dist/immutable.js.flow
```
jasonLaster: YESSS!

src/actions/expressions.js
wldcordeiro: Huh I didn't realize I could do it right in the `add/update` cases. I agree on the clear expression action, it's a little clearer to have a distinct function over a flag.

src/actions/expressions.js
```diff
@@ -37,26 +38,30 @@ export function addExpression(input: string) {
...
 
-    dispatch({
-      type: "ADD_EXPRESSION",
-      input
-    });
+    dispatch({ type: "ADD_EXPRESSION", input, expressionError });
 
     const newExpression = getExpression(getState(), input);
-    dispatch(evaluateExpression(newExpression));
+    if (newExpression) {
```
jasonLaster: is it possible to add expression and not have one?

do we need the IF?

jasonLaster (commented):
a couple of additional action refactors… i might be missing something here so feel free to push back

src/actions/expressions.js
```diff
@@ -37,26 +38,30 @@ export function addExpression(input: string) {
       return dispatch(evaluateExpression(expression));
     }
 
-    dispatch({
-      type: "ADD_EXPRESSION",
-      input
-    });
+    dispatch({ type: "ADD_EXPRESSION", input, expressionError });
```
jasonLaster: I think this makes sense, but just confirming that we can add an expression w/ an error

src/actions/expressions.js
```diff
@@ -37,26 +38,30 @@ export function addExpression(input: string) {
...
     if (!input || input == expression.input) {
       return;
     }
-
     dispatch({
       type: "UPDATE_EXPRESSION",
       expression,
-      input: input
+      input: expressionError ? expression.input : input,
+      expressionError
```
jasonLaster: maybe we want to have a separate action for `setExpressionError`

```js
if (error) {
  dispatch(setExpressionError(..))
else 
  dispatch(
    type: "UPDATE_EXPRESSION",
    ...
  )
```

the advantage is that that will simplify update expression. And we can do the same thing as w/ add expression where if we are updating the expression we can clear the error...

src/actions/tests/expressions.spec.js
```diff
@@ -75,13 +76,13 @@ describe("expressions", () => {
...
+    await dispatch(actions.addExpression("foo"));
+    await dispatch(actions.addExpression("bar"));
 
-    expect(selectors.getExpression(getState(), "foo").value).toBe(null);
-    expect(selectors.getExpression(getState(), "bar").value).toBe(null);
+    expect(selectors.getExpression(getState(), "foo").value).toBe("boo");
+    expect(selectors.getExpression(getState(), "bar").value).toBe("boo");
 
-    await dispatch(actions.evaluateExpressions("boo"));
+    await dispatch(actions.evaluateExpressions());
```
jasonLaster: should we test adding an expression / updating etc w/ errors?

wldcordeiro: I like that idea. Adding a test for error would be great.

src/actions/expressions.js
wldcordeiro: The line that was guarded was producing a type error when the guard wasn't present.

src/actions/expressions.js
wldcordeiro: I had the additional action before and the reason I have the line above isn't for setting the expression it's rather so we don't change the content of the expression to the new invalid value.

src/actions/expressions.js
```diff
@@ -38,6 +37,7 @@ export function addExpression(input: string) {
       return dispatch(evaluateExpression(expression));
     }
 
+    const expressionError = await parser.hasSyntaxError(input);
     dispatch({ type: "ADD_EXPRESSION", input, expressionError });
```
wldcordeiro: Sorry accidentally deleted both yours and one I added accidentally @jasonLaster. So the reducer prevents the creation of the expression. So no expression created.

src/actions/tests/expressions.spec.js
wldcordeiro: Tests added.

codehag:
What is the status here?

wldcordeiro:
I believe it should be ready to go @codehag but @jasonLaster may have some additional feedback.

jasonLaster:
Yay!

PR #4735, opened 2017-11-19
Prevent unecessary renders

### LINKED ISSUE #4692: Performance regression when loading sources
Performance regression spotted with the [11-7 bundle release](https://bugzilla.mozilla.org/show_bug.cgi?id=1415300).

See screenshot of DAMP perf charts, reload regressed of 10%:
image

The test case from https://bugzilla.mozilla.org/show_bug.cgi?id=1366693 really highlights the issue. The test page is simply sequentially loading scripts as fast as possible and measuring the time it takes to load all scripts. 

On my machine: 
- without devtools opened: < 1s
- with debugger opened (version 10-25): 3s ([profile](https://perfht.ml/2zCcNCz))
- with debugger opened (version 11-7): 10s ([profile](https://perfht.ml/2zY3rld))

The test case is a zip at https://bugzilla.mozilla.org/attachment.cgi?id=8869957

Precise STRs:
- serve the content of the zip e.g. on http://127.0.0.1:8080
- open http://127.0.0.1:8080
- (the page should display "Loaded in X seconds", X hopefully below 1)
- open the debugger (be careful about opening only the debugger and not any other tools as they can impact the perf)
- reload the page
- page will now take several seconds to load

jasonLaster:
Fixes #4692

### Summary of Changes

- the JSX patch introduced sourceMetaData, which when not set would be a new empty object every time, this fixes that by sharing the same object.
- i think it can be a bit cleaner

### Testing

I'd like to add a unit test that adds a source and checks to make sure the editor did not render.

### DIFF
--- a/src/reducers/ast.js
+++ b/src/reducers/ast.js
@@ -183,8 +183,9 @@ export function getPreview(state: OuterState) {
   return state.ast.get("preview");
 }
 
+const emptySourceMetaData = {};
 export function getSourceMetaData(state: OuterState, sourceId: string) {
-  return state.ast.getIn(["sourceMetaData", sourceId]) || {};
+  return state.ast.getIn(["sourceMetaData", sourceId]) || emptySourceMetaData;
 }
 
 export default update;


James300:
This is a really cool optimization for making sure React doesn't think things have changed 😄

src/reducers/ast.js
```diff
@@ -183,8 +183,9 @@ export function getPreview(state: OuterState) {
   return state.ast.get("preview");
 }
 
+const emptySourceMetaData = {};
 export function getSourceMetaData(state: OuterState, sourceId: string) {
-  return state.ast.getIn(["sourceMetaData", sourceId]) || {};
+  return state.ast.getIn(["sourceMetaData", sourceId]) || emptySourceMetaData;
```
jasonLaster: i bet getIn takes a default arg

codehag:
looks good!

Issue #4730, opened 2017-11-18
IsReactComponent needs improvements to support other implementations of a component


James300:
Add functionality and tests for the following scenarios:
- React.Component
- Using require instead of import

James300:
Woops, thought that said close cancel comment 😪

lukaszsobek:
Merged. Closing.

PR #4737, opened 2017-11-19
Improve isReactComponent

James300:
Associated Issue: #4730 

### Summary of Changes

* Detect `require("react")` and accept it as an import
* Detect `React.Component` and accept as a component extension

### DIFF
--- a/src/workers/parser/frameworks.js
+++ b/src/workers/parser/frameworks.js
@@ -5,13 +5,11 @@
 import getSymbols from "./getSymbols";
 
 export function isReactComponent(source) {
-  const { imports, classes } = getSymbols(source);
-
-  if (!imports || !classes) {
-    return false;
-  }
-
-  return importsReact(imports) && extendsComponent(classes);
+  const { imports, classes, callExpressions } = getSymbols(source);
+  return (
+    (importsReact(imports) || requiresReact(callExpressions)) &&
+    extendsComponent(classes)
+  );
 }
 
 function importsReact(imports) {
@@ -22,12 +20,21 @@ function importsReact(imports) {
   );
 }
 
+function requiresReact(callExpressions) {
+  return callExpressions.some(
+    callExpression =>
+      callExpression.name === "require" &&
+      callExpression.arguments.some(arg => arg.value === "react")
+  );
+}
+
 function extendsComponent(classes) {
   let result = false;
   classes.some(classObj => {
     if (
       classObj.parent.name === "Component" ||
-      classObj.parent.name === "PureComponent"
+      classObj.parent.name === "PureComponent" ||
+      classObj.parent.property.name === "Component"
     ) {
       result = true;
     }
--- a/src/workers/parser/frameworks.js
+++ b/src/workers/parser/frameworks.js
@@ -24,7 +24,7 @@ function requiresReact(callExpressions) {
   return callExpressions.some(
     callExpression =>
       callExpression.name === "require" &&
-      callExpression.arguments.some(arg => arg.value === "react")
+      callExpression.values.some(value => value === "react")
   );
 }
 
--- a/src/workers/parser/getSymbols.js
+++ b/src/workers/parser/getSymbols.js
@@ -23,8 +23,7 @@ export type SymbolDeclaration = {|
   parameterNames?: string[],
   identifier?: Object,
   computed?: Boolean,
-  arguments?: Object[],
-  values?: string
+  values?: string[]
 |};
 
 export type FunctionDeclaration = SymbolDeclaration & {|
--- a/src/workers/parser/getSymbols.js
+++ b/src/workers/parser/getSymbols.js
@@ -165,7 +165,7 @@ function extractSymbols(source: Source) {
           const { start, end, identifierName } = callee.loc;
           callExpressions.push({
             name: identifierName,
-            arguments: args,
+            values: args.filter(arg => arg.value).map(arg => arg.value),
             location: { start, end }
           });
         }
--- a/src/workers/parser/getSymbols.js
+++ b/src/workers/parser/getSymbols.js
@@ -23,7 +23,8 @@ export type SymbolDeclaration = {|
   parameterNames?: string[],
   identifier?: Object,
   computed?: Boolean,
-  arguments?: Object[]
+  arguments?: Object[],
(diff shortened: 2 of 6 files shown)


src/workers/parser/getSymbols.js
```diff
@@ -47,33 +47,49 @@ function getFunctionParameterNames(path: NodePath): string[] {
 }
 
 function getVariableNames(path: NodePath): SymbolDeclaration[] {
+  let callees = path.node.declarations.reduce((callees, declaration) => {
+    return callees.concat(declaration.init.callee);
+  }, []);
+
+  let args = path.node.declarations.reduce((args, declaration) => {
```
James300: I wasn't sure if there was a better way to detect a `require` call but I found that `isVariable` worked and this lead me down this rabbit hole, any ideas if there is a better way or if these reduce calls are a good idea?

James300:
I've tried quite a few things now but I can't figure out why things are breaking in the CI tests, anyone got any ideas?

jasonLaster:
no worries, i'll look tomorrow

src/workers/parser/frameworks.js
```diff
@@ -22,12 +20,21 @@ function importsReact(imports) {
   );
 }
 
+function requiresReact(callExpressions) {
+  return callExpressions.some(
+    callExpression =>
+      callExpression.name === "require" &&
+      callExpression.arguments.some(arg => arg.value === "react")
+  );
+}
```
jasonLaster: nice!

for the record, this is another way it was done. not sure which is better

```diff
+ function importsReact(imports, identifiers) {
+   return (
+     imports.some(
+       importObj =>
+         importObj.source === "react" &&
+         importObj.specifiers.some(specifier => specifier === "React")
+     ) || identifiers.some(identifier => identifier.name === "React")

```

src/workers/parser/getSymbols.js
```diff
@@ -159,10 +160,12 @@ function extractSymbols(source: Source) {
 
       if (t.isCallExpression(path)) {
         const callee = path.node.callee;
+        const args = path.node.arguments;
         if (!t.isMemberExpression(callee)) {
           const { start, end, identifierName } = callee.loc;
           callExpressions.push({
             name: identifierName,
+            arguments: args,
```
jasonLaster: Lets add this to the formatFixtures function so we can see the output.

src/workers/parser/getSymbols.js
```diff
@@ -159,10 +160,12 @@ function extractSymbols(source: Source) {
 
       if (t.isCallExpression(path)) {
         const callee = path.node.callee;
+        const args = path.node.arguments;
```
jasonLaster: We likely just want to store the value

`path.node.arguments.map(arg => arg.value)`

jasonLaster:
can you try rebasing so we remove the indentation changes

src/workers/parser/getSymbols.js
James300: If we do this, then we can't confirm that the expression's name is 'require', I'm not sure how many other functions would have a value of 'react', what do you think? Any that do contain a function like this probably are a react file anyway right 😛

src/workers/parser/getSymbols.js
James300: Where is the `formatFixtures` function?

src/workers/parser/getSymbols.js
jasonLaster: I'm confused. Wouldn't we have the call expression name and the list of arguments?

src/workers/parser/getSymbols.js
jasonLaster: oops: https://github.com/devtools-html/debugger.html/blob/master/src/workers/parser/utils/formatSymbols.js#L23

src/workers/parser/getSymbols.js
James300: Yeah that's my bad, changing it now

jasonLaster:
why do we need the full arg object?

i think we could do this:
`callExpression.arguments.some(arg => arg === "react")`

src/workers/parser/getSymbols.js
```diff
@@ -159,10 +160,12 @@ function extractSymbols(source: Source) {
...
         if (!t.isMemberExpression(callee)) {
           const { start, end, identifierName } = callee.loc;
           callExpressions.push({
             name: identifierName,
+            values: args.filter(arg => arg.value).map(arg => arg.value),
```
James300: Is just storing values better? Also, I use a filter first to trim the null / undefined values

jasonLaster: this looks good to me

src/workers/parser/getSymbols.js
codehag: this still needs to be changed, right?

James300: I used just the value lower down to try and save a bit of performance because it does a map, is there any benefit to changing it here?

src/workers/parser/getSymbols.js
```diff
@@ -22,7 +22,9 @@ export type SymbolDeclaration = {|
   expressionLocation?: BabelLocation,
   parameterNames?: string[],
   identifier?: Object,
-  computed?: Boolean
+  computed?: Boolean,
+  arguments?: Object[],
+  values?: string
```
jasonLaster: do we need both of these? is this a string or a list of strings?

src/workers/parser/tests/__snapshots__/getSymbols.spec.js.snap
```diff
@@ -740,8 +740,8 @@ variables:
 [(14, 4), (14, 25)]   child
 
 callExpressions:
-[(8, 14), (8, 20)]   square
-[(10, 15), (10, 22)]   squaare
+[(8, 14), (8, 20)]   square   2
+[(10, 15), (10, 22)]   squaare   4
```
jasonLaster: nice

src/workers/parser/getSymbols.js
James300: @codehag @jasonLaster do you want me to make this change or is it good?

codehag: it looks good

Issue #4733, opened 2017-11-18
[SourceTree] show angular icon as a domain


jasonLaster:
We currently show the webpack logo when the domain is webpack, we should do the same for angular :)

- the path might be slightly off - easy to check in `addToTree.js`
- the icon will need styling w/ light and dark themes

```diff
diff --git a/src/components/PrimaryPanes/SourcesTree.js b/src/components/PrimaryPanes/SourcesTree.js
index e2d6981..3f8adf8 100644
--- a/src/components/PrimaryPanes/SourcesTree.js
+++ b/src/components/PrimaryPanes/SourcesTree.js
@@ -232,6 +232,10 @@ class SourcesTree extends Component<Props, State> {
       return <Svg name="webpack" />;
     }
 
+    if (item.path === "/ng:") {
+      return <Svg name="angular" />;
+    }
+
     if (depth === 0) {
       return (
         <img
```

<img src='https://shipusercontent.com/3fc075cf7a02d2004ce2aef195e2e3b3/Screen%20Shot%202017-11-18%20at%204.54.41%20PM.png' title='Screen Shot 2017-11-18 at 4.54.41 PM.png' width=319>

borian:
Hey Jason,
https://github.com/borian/debugger.html/commit/762f7ced050930db5086a066a9a800a28a66540f
The `angular.svg` was already there so I only added the path check and styling.

Because the icon is already used in the call stack view, I needed to add specific styling for the source pane

`.managed-tree .tree-node .angular svg {`

![light](https://user-images.githubusercontent.com/1525694/32992085-ae98ed28-cd46-11e7-9322-6a17fd220d74.png)
![dark](https://user-images.githubusercontent.com/1525694/32992086-b12a40c8-cd46-11e7-82f5-9c6db84bd648.png)

If this is ok, I will do a PR

zaggy:
Maybe a good thing is to write `Angular (ng://)` instead of just plain `ng://` or like we do with Webpack, to ditch the address part altogether.

jasonLaster:
Yep, I'd just show Angular. Most users don't care

borian:
updated the PR https://github.com/devtools-html/debugger.html/pull/4738
Now it just showes Angular


![ng-light](https://user-images.githubusercontent.com/1525694/33102301-69ee4ed4-cf1c-11e7-9363-2663085d53e7.png)
[...]

PR #4738, opened 2017-11-19
added angular icon in the source pane for angular projects

borian:
Associated Issue: #4733

### Summary of Changes

added angular icon in the source pane for angular projects

light
dark

### DIFF
--- a/src/components/shared/Svg.css
+++ b/src/components/shared/Svg.css
@@ -115,3 +115,14 @@ html .arrow.expanded svg {
 .theme-dark .webpack {
   opacity: 0.5;
 }
+
+.managed-tree .tree-node .angular svg {
+  width: 15px;
+  height: 15px;
+  margin-right: 5px;
+  vertical-align: sub;
+}
+
+.theme-dark .managed-tree .tree-node .angular {
+  opacity: 0.5;
+}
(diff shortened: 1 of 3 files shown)


src/components/shared/Svg.css
```diff
@@ -115,3 +115,14 @@ html .arrow.expanded svg {
 .theme-dark .webpack {
   opacity: 0.5;
 }
+
+.managed-tree .tree-node .angular svg {
```
James300: Any reason this isn't just `.angular`, I understand the difference but I'm curious if this is safer / more maintainable / something else?

borian: Sorry, I only mentioned this in the issue https://github.com/devtools-html/debugger.html/issues/4733.
The icon is already used in the call stack view, thats why I added specific styling for the source pane

jasonLaster:
@borian can we replace `ng://` and `Angular`

borian:
@jasonLaster shure, I changed the name and squashed the commit
ng-light
ng-dark

jasonLaster:
Thanks i made a couple of tweaks including the shorter classname, hopefully we'll be okay :)

borian:
This css will add additional styling to the icon used in the call stack view.
But it will only be 3px larger, i think 😄

PR #4747, opened 2017-11-20
Clicking a console link should navigate to the correct line

jasonLaster:
Associated Issue: #3411

### Summary of Changes

When we switched to selectSource accepting a location we missed some call sites. This fixes that an adds a mochitest.

### DIFF
--- a/src/test/mochitest/examples/script-switching-02.js
+++ b/src/test/mochitest/examples/script-switching-02.js
@@ -11,3 +11,4 @@ function secondCall() {
 }
 
 var x = true;
+console.log("hi")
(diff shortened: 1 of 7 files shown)


src/test/mochitest/examples/script-switching-02.js
```diff
@@ -11,3 +11,4 @@ function secondCall() {
 }
 
 var x = true;
+console.log("hi")
```
bomsy: rouge one! :)

jasonLaster: haha - we need it for the mochitest

PR #4753, opened 2017-11-20
only load sources once

jasonLaster:
Associated Issue: #4705 

### Summary of Changes

* updates load source text so only one source is loaded at a time
* starts a unit test

### DIFF
--- a/src/actions/sources/loadSourceText.js
+++ b/src/actions/sources/loadSourceText.js
@@ -8,16 +8,25 @@ import { PROMISE } from "../utils/middleware/promise";
 import { setSymbols } from "../ast";
 import { getSource } from "../../selectors";
 import { setSource } from "../../workers/parser";
+import { isLoading, isLoaded } from "../../utils/source";
 import type { Source } from "../../types";
 import type { ThunkArgs } from "../types";
 
+const requests = new Map();
+
 async function loadSource(source: Source, { sourceMaps, client }) {
+  console.log("load!");
   if (sourceMaps.isOriginalId(source.id)) {
     return await sourceMaps.getOriginalSourceText(source);
   }
 
+  console.log("...load!");
   const response = await client.sourceContents(source.id);
-
+  console.log("load! ...", {
+    id: source.id,
+    text: response.source,
+    contentType: response.contentType || "text/javascript"
+  });
   return {
     id: source.id,
     text: response.source,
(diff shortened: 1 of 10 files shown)


src/actions/sources/loadSourceText.js
```diff
@@ -31,17 +40,29 @@ async function loadSource(source: Source, { sourceMaps, client }) {
...
 export function loadSourceText(source: Source) {
   return async ({ dispatch, getState, client, sourceMaps }: ThunkArgs) => {
+    console.log("lst", source);
     // Fetch the source text only once.
-    if (source.text) {
+    if (isLoaded(source)) {
       return Promise.resolve(source);
     }
 
+    if (isLoading(source)) {
```
codehag: this requires that there has been a state update before this happened, however it could be that two requests with the same data will pass this check, thus resulting in the same issue we currently have. it would be better to do 

```javascript
const existingRequest = requests.get(source.id);
if (existingRequest) {
  return existingRequest
}

Comment (jasonLaster) [reply src/actions/sources/loadSourceText.js]: Ugh... sources of truth :)

**Comment by jasonLaster** [comment]
browser_dbg-outline

**Comment by codehag** [comment]
remove changes to assets manifest

File: src/actions/sources/loadSourceText.js [hunk 2]
```diff
@@ -25,29 +28,51 @@ async function loadSource(source: Source, { sourceMaps, client }) {
...
+
 /**
  * @memberof actions/sources
  * @static
  */
 export function loadSourceText(source: Source) {
   return async ({ dispatch, getState, client, sourceMaps }: ThunkArgs) => {
+    const deferred = defer();
+
+    deferred.resolve();
```
codehag: i dont understand what this is doing. why do we want to resolve here?

src/actions/sources/loadSourceText.js
```diff
@@ -25,29 +28,51 @@ async function loadSource(source: Source, { sourceMaps, client }) {
...
-    if (source.text) {
+    if (isLoaded(source)) {
       return Promise.resolve(source);
     }
 
+    if (isLoading(source) || requests.has(source.id)) {
+      return requests.get(source.id);
+    }
+
+    requests.set(source.id, deferred.promise);
```
codehag: could we set loadSource here instead?

src/actions/sources/loadSourceText.js
```diff
@@ -25,29 +28,51 @@ async function loadSource(source: Source, { sourceMaps, client }) {
...
+    requests.delete(source.id);
+
     const newSource = getSource(getState(), source.id).toJS();
     if (newSource.isWasm) {
       return;
     }
 
     await setSource(newSource);
     dispatch(setSymbols(source.id));
+    return deferred.resolve();
```
codehag: i dont think that this is doing more than the async function, we might not need it

src/utils/test-head.js
```diff
@@ -50,7 +50,7 @@ function commonLog(msg: string, data: any = {}) {
 function makeSource(name: string, props: any = {}) {
   return {
     id: name,
-    loadedState: "loaded",
+    loadedState: "unloaded",
```
codehag: changed this, because we more often need the unloaded state than the loaded state

jasonLaster: makes sense

jasonLaster:
Thanks for the tweaks. Looks great!

src/actions/sources/loadSourceText.js
```diff
@@ -25,17 +28,35 @@ async function loadSource(source: Source, { sourceMaps, client }) {
...
-    if (source.text) {
+    if (isLoaded(source)) {
       return Promise.resolve(source);
     }
 
+    if (isLoading(source) || requests.has(source.id)) {
+      return requests.get(source.id);
+    }
+
+    requests.set(source.id, deferred.promise);
```
jasonLaster: we set the promise here because we only want to resolve it after symbols and other async work is done

src/utils/test-head.js
jasonLaster: nice

jasonLaster: nice

PR #4757, opened 2017-11-21
[WIP] add damp test runner with compare and group commits

codehag:
This shouldn't be merged, I am just making it available in case anyone else needs it

what this does:
- allows you to run damp tests from the github project
- allows you to specify a series of commits that you want to run
- allows you to compare previously run commits

flags:

`--target` is where your firefox install is located. if it is located in your debugger directory you would do something like this: `--target ./firefox`

`--commits` here you specify the commits you want to run. each commit is separate. for example
`--commits df2cd2fa55e79a6f69f72aeab5828b75cb06a126 3741eeb267fd2f0a1b42107389e31d3887798957  b800f2eb6f575622ade6448307b5151db385274c`

this would run and compare these three commits

`--from` allows you to specify a first commit from which to walk up the git log

`--to` allows you to specify a last commit were to end, can only be used with from


codehag:
this is, by the way, totally hacked together an in no way ready

codehag:
moved to an npm module: https://www.npmjs.com/package/damp-git-runner

jasonLaster:
Great! Just like mochi! Can't wait to hack on this

Issue #4715, opened 2017-11-16
[Pause] Locations should be symmetric  for original and generated files


jasonLaster:
One of the coolest features we've worked on IMO, is debugger symmettry. Basically, we expect users to mostly care about debugging the original code, but we'll try to help users when they want to see the generated code.

One of the biggest use cases for trying to see the generated code is when you want to inspect variables in the generated code.

We currently map breakpoint locations, so you can see where the breakpoint is. We do not map the debug line and call stack locations, but we totally could and should

here's a [patch](https://gist.github.com/a25952d5dc9caac5cd6178d106be529d)

<img src='https://shipusercontent.com/0cdcafbbe2cd423f38cc58bb6299ccef/symmettry.jpg' title='symmettry.jpg' width=1773>

yurydelendik:
It is worth to notice that "symmetric" might not be a right word here. Single generated location might be mapped to multiple original locations (see e.g. [allGeneratedPositionsFor](https://github.com/mozilla/source-map#sourcemapconsumerprototypeallgeneratedpositionsfororiginalposition) ), but probably to simplicity of this issue the first generated location (or one returned by `generatedPositionFor`) must be selected for symmetry.

PR #4758, opened 2017-11-21
[Pause] Locations should be symmetric  for original and generated files (#4715)

assafdd:
[Pause] Locations should be symmetric  for original and generated files (#4715)

Associated Issue: #4715
### Summary of Changes

* moved getCallStackFrames logic from frames/index.js to a separate selector.
* Frames location use generatedLocation if source is not  original
* Debug line also use generatedLocation if source is not  original

### Test Plan

Example test plan:
- used http://localhost:7999/increment/ for testing
- placed BP in increment.js line 4
- clicked on increment
- switched between original and sourcedMap tabs.
- line numbers in breaks point reflects the selected source (original or sourcedMap)
- debug line in editor pane show the correct line based on the selected soruce  

debugger-symmettry

### DIFF
--- a/src/components/Editor/DebugLine.js
+++ b/src/components/Editor/DebugLine.js
@@ -7,6 +7,7 @@ import { Component } from "react";
 import { markText, toEditorPosition } from "../../utils/editor";
 import { getDocument } from "../../utils/editor/source-documents";
 
+import { isOriginalId } from "devtools-source-map";
 import { connect } from "react-redux";
 import {
   getSelectedLocation,
@@ -70,7 +71,11 @@ export class DebugLine extends Component<Props, State> {
       return;
     }
 
-    const { location, location: { sourceId } } = selectedFrame;
+    const isGeneratedSource = !isOriginalId(selectedLocation.sourceId);
+    const location = isGeneratedSource
+      ? selectedFrame.generatedLocation || selectedFrame.location
+      : selectedFrame.location;
+    const sourceId = location.sourceId;
     const doc = getDocument(sourceId);
     if (!doc) {
       return;
@@ -130,6 +135,9 @@ export class DebugLine extends Component<Props, State> {
 
 export default connect(state => ({
   selectedLocation: getSelectedLocation(state),
+
+  // we should give `getSelectedFrame` the breakpoints treatment
+  // and have it return the correct location: original or generated ...
   selectedFrame: getSelectedFrame(state),
   pauseInfo: getPause(state)
 }))(DebugLine);
--- a/src/components/Editor/DebugLine.js
+++ b/src/components/Editor/DebugLine.js
@@ -8,11 +8,11 @@ import { markText, toEditorPosition } from "../../utils/editor";
 import { getDocument } from "../../utils/editor/source-documents";
 
 import { connect } from "react-redux";
-import { getSelectedFrameLocation, getPause } from "../../selectors";
+import { getVisibleSelectedFrame, getPause } from "../../selectors";
 
 type Props = {
   editor: Object,
-  selectedLocation: Object,
+  selectedFrame: Object,
   pauseInfo: Object
 };
 
@@ -35,36 +35,36 @@ export class DebugLine extends Component<Props, State> {
   componentDidMount() {
     this.setDebugLine(
       this.props.pauseInfo,
-      this.props.selectedLocation,
+      this.props.selectedFrame,
       this.props.editor
     );
   }
 
   componentWillReceiveProps(nextProps: Props) {
-    this.clearDebugLine(this.props.selectedLocation, this.props.editor);
+    this.clearDebugLine(this.props.selectedFrame, this.props.editor);
     this.setDebugLine(
       nextProps.pauseInfo,
-      nextProps.selectedLocation,
+      nextProps.selectedFrame,
       nextProps.editor
     );
   }
 
   componentWillUnmount() {
-    this.clearDebugLine(this.props.selectedLocation, this.props.editor);
+    this.clearDebugLine(this.props.selectedFrame, this.props.editor);
   }
 
-  setDebugLine(pauseInfo: Object, selectedLocation: Object, editor: Object) {
-    if (!selectedLocation) {
+  setDebugLine(pauseInfo: Object, selectedFrame: Object, editor: Object) {
+    if (!selectedFrame) {
       return;
     }
 
-    const sourceId = selectedLocation.sourceId;
+    const sourceId = selectedFrame.location.sourceId;
     const doc = getDocument(sourceId);
     if (!doc) {
       return;
     }
 
-    const { line, column } = toEditorPosition(sourceId, selectedLocation);
+    const { line, column } = toEditorPosition(sourceId, selectedFrame.location);
 
     // make sure the line is visible
     if (editor && editor.alignLine) {
@@ -81,11 +81,11 @@ export class DebugLine extends Component<Props, State> {
     this.setState({ debugExpression });
   }
 
-  clearDebugLine(selectedLocation: Object, editor: Object) {
-    if (!selectedLocation) {
+  clearDebugLine(selectedFrame: Object, editor: Object) {
+    if (!selectedFrame) {
       return;
     }
-    const { line, sourceId } = selectedLocation;
+    const { line, sourceId } = selectedFrame.location;
     const { debugExpression } = this.state;
     if (debugExpression) {
       debugExpression.clear();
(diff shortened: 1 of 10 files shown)


src/components/Editor/DebugLine.js
```diff
@@ -130,6 +135,9 @@ export class DebugLine extends Component<Props, State> {
 
 export default connect(state => ({
   selectedLocation: getSelectedLocation(state),
+
```
assafdd: I didn't get it. do we want to refactor getSelectedLocation to refer to original or sourcedMap location or change the getSelectedFrame to return location?

src/components/Editor/DebugLine.js
```diff
@@ -130,6 +135,9 @@ export class DebugLine extends Component<Props, State> {
...
+  // we should give `getSelectedFrame` the breakpoints treatment
+  // and have it return the correct location: original or generated ...
   selectedFrame: getSelectedFrame(state),
```
assafdd: I didn't get it. do we want to refactor getSelectedLocation to refer to original or sourcedMap location or change the getSelectedFrame to return location?

src/components/Editor/DebugLine.js
```diff
@@ -70,7 +71,11 @@ export class DebugLine extends Component<Props, State> {
       return;
     }
 
-    const { location, location: { sourceId } } = selectedFrame;
+    const isGeneratedSource = !isOriginalId(selectedLocation.sourceId);
+    const location = isGeneratedSource
+      ? selectedFrame.generatedLocation || selectedFrame.location
+      : selectedFrame.location;
+    const sourceId = location.sourceId;
```
jasonLaster: this should work, but i think it would be better in the long run to keep our components dumb - and not let them know about this.

One good reason for this is because when you jump between original and generated contexts, it's good for the component to see a new location so it's updated

src/components/Editor/DebugLine.js
jasonLaster: hmm. probably getSelectedFrame

src/selectors/getCallStackFrames.js
```diff
@@ -0,0 +1,48 @@
...
+  if (!frames) {
+    return null;
+  }
+
+  return frames
+    .filter(frame => getSourceForFrame(sources, frame))
+    .map(frame => appendSource(sources, frame, selectedSource))
+    .filter(frame => !get(frame, "source.isBlackBoxed"))
+    .map(annotateFrame);
+}
```
jasonLaster: looks great. Nice!

assafdd:
@jasonLaster move the logic out of the component. 
Is there anything we want to do with visibleBreakpoints.js ? looks fine to me.

assafdd:
@jasonLaster the last test that is still failing is "Blackboxed Frames"  in tests/frame.js
Since this logic is now moved to getCallStackFrames() in selector folder I do not see a value in that unit test any more. If you think it's important I can create test getCallStackFrames tough I do not see any test in selector folder yet.

jasonLaster:
Thanks @assafdd the component test should still pass as the functionality has not changed. I'm traveling, so a bit slower w/ reviews :)

assafdd:
@jasonLaster fix frames test. I think it's ready.

src/selectors/getSelectedFrameLocation.js
```diff
@@ -0,0 +1,21 @@
...
+  const selectedLocation = getSelectedLocation(state);
+  const isGeneratedSource = !isOriginalId(selectedLocation.sourceId);
+  const selectedFrame = getSelectedFrame(state);
+
+  if (!selectedFrame) {
+    return;
+  }
+
+  return isGeneratedSource
+    ? selectedFrame.generatedLocation || selectedFrame.location
```
jasonLaster: are there cases when there won't be `selectedFrame.generatedLocation`?

src/selectors/visibleBreakpoints.js
```diff
@@ -18,6 +18,9 @@ function formatBreakpoint(breakpoint, selectedSource) {
   const isGeneratedSource = isGeneratedId(sourceId);
 
   return {
+    // when we get the location for the breakpoint
+    // we do the smart thing of returning the relevant location
+    // either the generated or original location!!!
```
jasonLaster: we don't need this comment :)

src/selectors/getCallStackFrames.js
```diff
@@ -0,0 +1,58 @@
...
+function appendSource(sources, frame, selectedSource) {
+  const isGeneratedSource = !isOriginalId(selectedSource.get("id"));
+  return {
+    ...frame,
+    location: getLocation(frame, isGeneratedSource),
+    source: getSourceForFrame(sources, frame).toJS()
+  };
+}
+
+export function getAndProcessFrames(
```
jasonLaster: lets call this function `getCallStackFrames`

src/components/SecondaryPanes/Frames/tests/__snapshots__/Frames.js.snap
```diff
@@ -40,7 +42,9 @@ exports[`Frames Blackboxed Frames filters blackboxed frames 1`] = `
           "location": Object {
             "sourceId": "1",
           },
-          "source": Object {},
+          "source": Object {
+            "id": "1",
```
jasonLaster: i wonder how this changed :)

src/components/SecondaryPanes/Frames/index.js
```diff
@@ -8,32 +8,23 @@ import PropTypes from "prop-types";
...
 import { copyToTheClipboard } from "../../../utils/clipboard";
 
 import {
-  getFrames,
   getFrameworkGroupingState,
   getSelectedFrame,
-  getSourceInSources,
-  getSources,
-  getPause
+  getPause,
```
jasonLaster: I love seeing this complexity go away. Especially lodash, reselect, the selectors...

src/selectors/getSelectedFrameLocation.js
```diff
@@ -0,0 +1,21 @@
...
+    : selectedFrame.location;
+}
```
jasonLaster: I prefer what we do w/ getVisibleBreakpoints and returning the entire object for a couple of reasons

- the debug line still needs to know if we're paused on an error
- the debug line could need the frame ID...
- seems more future proof

jasonLaster:
Shit, looks like we need to clear the debug line

1. pause,
2. go to the bundle
3. go to the original
4. step
5. go to the bundle

(the first bundle clear line wasn't cleared...)

src/selectors/getSelectedFrameLocation.js
assafdd: I did not encountered such use case. I just used the same logic we have wrote in getCallStackFrames and also saw in visibleBreakpoint , assuming there could be such use case. 
I can remove it if you feel comfortable with it.

src/selectors/getCallStackFrames.js
assafdd: thats name is used by the default  export function. 
we can call is formatCallStackFrames.

src/selectors/getSelectedFrameLocation.js
assafdd: Cool. beside location, which other props from the frame would you like to copy?
right now it's the only one that is really needed in DebugLine.js

assafdd:
@jasonLaster I could not reproduce the scenario where we don't clear the debugLine in the bundled source. 

One thing I did saw is that the source name in of the sources in the callstack is not change to 'bundle' when we switch to the generated code while the line numbers are changed. 

callstack_source_names

jasonLaster:
Here's a picture, i got these timing issues when i worked on debugline earlier

there are some traps

jasonLaster:
@assafdd no problem, lets try to address the file names and the mochitests. have you run the mochitests locally yet: https://github.com/devtools-html/debugger.html/blob/master/docs/mochitests.md

they look pretty busted.

codehag:
fixed conflicts

jasonLaster:
Nice find @bomsy. 

We use the selected file to determine if the call stack should be original / generated. The interesting thing is that some frames are ALWAYS generated because HTML files are never mapped. As a result, switching to those frames will change the context, making it difficult to go back. This is also the case when the map does not have a good mapping. 

I think we can live w/ this because in practice HTML frames are rare and no-mappings are largely transpiler code, which no-one wants to see either.

bomsy:
Ok cool ... makes sense!

assafdd:
technically we can store on each frame it's last source state (generated or original)  and use if we find it when rendering the frames source file names. But it would be a little ugly to store UI state in store. so if it's rare use-case probably better to keep it clean and simple

Issue #4071, opened 2017-09-18
[Accessibility] Search sources autocomplete functionality needs to be accessible


yzen:
Some background can be found [here]. Keyboard accessibility is pretty great, missing is the semantics:

- [ ] aria-autocomplete="list" must be used in the search [input field][if]
- [ ] aria-controls should be used in the search input to point to the result-list
- [ ] aria-expanded should be used appropriately when the results are shown/hidden
- [ ] result-list should be an aria-live="polite" so screen readers could be notified when new results are rendered
- [ ] result-list should have a role of ['listbox'][lb] and each individual result should have a role of 'option'.
- [ ] aria-activedescendant (set on the listbox) should be pointing to the ID of the currently highlighted result item.
- [ ] Ideally, for each result item the labeling should be as follows: aria-labelledby should be pointing to ".title" child and aria-describedby should be pointing to ".subttitle". This will require assigning ID's to each one of them. This will help with the unnecessary verbosity of each result item where the screen reader would not have to utter a complete path for each one of them.

[here]:  https://www.marcozehe.de/2014/03/11/easy-aria-tip-7-use-listbox-and-option-roles-when-constructing-autocomplete-lists/
[if]: https://www.w3.org/TR/wai-aria/states_and_properties#aria-autocomplete 
[lb]: https://www.w3.org/TR/wai-aria/roles#listbox

### Screenshot

![Screen Shot 2017-09-18 at 8.08.43 PM.png](https://shipusercontent.com/5f0c12dd786fa5deeec037101a48606a/Screen%20Shot%202017-09-18%20at%208.08.43%20PM.png)

yzen:
### aria-autocomplete="list" must be used in the search input field
```diff
diff --git a/src/components/shared/SearchInput.js b/src/components/shared/SearchInput.js
index b8966fa1..4dd1a66c 100644
--- a/src/components/shared/SearchInput.js
+++ b/src/components/shared/SearchInput.js
@@ -118,6 +118,7 @@ class SearchInput extends Component {
       onKeyUp,
       onFocus,
       onBlur,
+      "aria-autocomplete": "list",
       placeholder,
       value: query,
       spellCheck: false,
```

yzen:
[...]

PR #4760, opened 2017-11-21
Search sources autocomplete functionality is accessible (#4071)

davisv007:
Associated Issue: #4071

### Summary of Changes

- [x] aria-autocomplete="list" added in the search input field
- [x] aria-controls= ‘result-list’ added in the search input
- [x] aria-expanded added to show when the results are shown/hidden
- [x] aria-live="polite" added to result-list
- [x] result-list has a role of 'listbox' and each individual result has a role of 'option'.
- [x] aria-activedescendant points to the ID of the currently highlighted result item.
- [x]  aria-labelledby points to ".title" child and aria-describedby points to ".subttitle" using IDs.
 
### Test Plan

We used ChromeVox for Google Chrome as our screenreader. 

Example test plan:
1. Open TodoMVC example

From debugger:

2. Command-P opens the panel
3. type in ‘js’
4. you should here the first result’s filepath
5. you should be read the next items filepath as you scroll through the result list

### DIFF
--- a/src/components/shared/ResultList.js
+++ b/src/components/shared/ResultList.js
@@ -16,14 +16,16 @@ type Props = {
     item: any,
     index: number
   ) => void,
-  size: string
+  size: string,
+  role: string
 };
 
 export default class ResultList extends Component<Props> {
   displayName: "ResultList";
 
   static defaultProps = {
-    size: "small"
+    size: "small",
+    role: "listbox"
   };
 
   constructor(props: Props) {
@@ -38,6 +40,7 @@ export default class ResultList extends Component<Props> {
       key: `${item.id}${item.value}${index}`,
       ref: String(index),
       title: item.value,
+      role: "option",
       className: classnames("result-item", {
         selected: index === selected
       })
@@ -52,10 +55,10 @@ export default class ResultList extends Component<Props> {
   }
 
   render() {
-    const { size, items } = this.props;
+    const { size, items, role } = this.props;
 
     return (
-      <ul className={classnames("result-list", size)}>
+      <ul className={classnames("result-list", size)} role={role}>
         {items.map(this.renderListItem)}
       </ul>
     );
--- a/src/components/shared/ResultList.js
+++ b/src/components/shared/ResultList.js
@@ -58,7 +58,15 @@ export default class ResultList extends Component<Props> {
     const { size, items, role } = this.props;
 
     return (
-      <ul className={classnames("result-list", size)} role={role}>
+      <ul
+        className={classnames("result-list", size)}
+        role={role}
+        aria-activedescendant={
+          typeof items !== undefined && items.length > 0
+            ? items[selected].id + "-title"
+            : null
+        }
+      >
         {items.map(this.renderListItem)}
       </ul>
     );
--- a/src/components/shared/ResultList.js
+++ b/src/components/shared/ResultList.js
@@ -40,6 +40,8 @@ export default class ResultList extends Component<Props> {
       key: `${item.id}${item.value}${index}`,
       ref: String(index),
       title: item.value,
+      "aria-labelledby": `${item.id}-title`,
+      "aria-describedby": `${item.id}-subtitle`,
       role: "option",
       className: classnames("result-item", {
         selected: index === selected
@@ -48,8 +50,12 @@ export default class ResultList extends Component<Props> {
 
     return (
       <li {...props}>
-        <div className="title">{item.title}</div>
-        <div className="subtitle">{item.subtitle}</div>
+        <div id={`${item.id}-title`} className="title">
+          {item.title}
+        </div>
+        <div id={`${item.id}-subtitle`} className="subtitle">
+          {item.subtitle}
+        </div>
       </li>
     );
   }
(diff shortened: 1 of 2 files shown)


src/components/shared/ResultList.js
```diff
@@ -38,24 +40,41 @@ export default class ResultList extends Component<Props> {
...
   render() {
-    const { size, items } = this.props;
+    const { size, items, role, selected } = this.props;
 
     return (
-      <ul className={classnames("result-list", size)}>
+      <ul
+        className={classnames("result-list", size)}
+        role={role}
+        aria-activedescendant={
```
jasonLaster: Nice

sanginovs: Hi @jasonLaster , is there anything we can do or should we just wait for our pull request to be reviewed by others?

src/components/shared/SearchInput.js
```diff
@@ -139,6 +139,8 @@ class SearchInput extends Component {
       onKeyUp,
       onFocus,
       onBlur,
+      "aria-autocomplete": "list",
+      "aria-controls": "result-list",
```
yzen: aria-controls has to point to an id ref (e.g. other element's id attribute). Since you are referencing "result-list" here, it has to be set as an id for the result-list <UL>.

src/components/shared/ResultList.js
```diff
@@ -38,24 +40,41 @@ export default class ResultList extends Component<Props> {
...
+          typeof items !== undefined && items.length > 0
+            ? items[selected].id + "-title"
+            : null
+        }
+        aria-expanded={typeof items !== undefined && items.length > 0}
```
yzen: aria-expanded belongs to the parent of the input field instead of the list. Move it there. 

That parent ("search-field"), should also have role="combobox" and aria-haspopup="listbox" attributes (For more details and a really good example see https://www.w3.org/TR/wai-aria/roles#combobox).  

Because "search-field" element is not an ancestor or the results list, lets add aria-owns="result-list" to it as well.

src/components/shared/ResultList.js
yzen: I was doing some more research on how best to implement accessible autocomplete and I was hoping it's OK I ask for a couple of changes (from my original comments), here in particular:

aria-activedescendant should belong to the input element instead of the results list.

davisv007:
Checklist of changes needed:
- [ ] aria-activedescendant should belong to the input element instead of the results list.
- [ ] aria-expanded belongs to the parent of the input field instead of the list. Move it there.
- [ ] "search-field" should have role="combobox" 
- [ ] "search-field" should have aria-owns="result-list" 
- [ ] "search-field" should have aria-haspopup="listbox" 
- [ ] aria-controls has to point to another element's id attribute ("result-list" has to be set as an id for the result-list)

src/components/shared/ResultList.js
davisv007: Hi, @yzen. We are trying to move aria-activedescendant to the input field in SearchInput.js, but we are having trouble accessing the items list and the selected items index from ResultList.js. 

We have tried importing ResultList in SearchInput but I think the issue is that the components in  ResultList have not yet loaded when we try to access them in SearchInput.

One of my team members has also looked into passing the onSelectResultItem function from QuickOpenModal.js to SearchInput.js. 

So far, neither have worked.

davisv007:
@yzen,
So most of the changes we needed to make were pretty minor, however, we have had some challenges with aria-expanded and aria-activedescendant.
We've tried to move aria-activedescendant to the input field in SearchInput.js, but we have had trouble accessing the items list and the selected items index from ResultList.js. 
We have tried importing ResultList in SearchInput but I think the issue is that the components in  ResultList have not yet loaded when we try to access them in SearchInput.
One of my team members has also looked into passing the onSelectResultItem function from QuickOpenModal.js to SearchInput.js. 
So far, neither have worked.

I think our team have ran out of time as this semester is about to end. What would you like us to do to wrap up before we leave this pull request?

yzen:
Hi @davisv007, I looked at the aria-expanded case first. I believe in case of the quick open dialog,
you can infer your expanded state for the input field based on the following: at time of render you know if your the result is enabled based on the value of this.props.enabled (props are props of QuickOpenModal) and you also know if you are rendering any results based on the value of this.state.results (if it's non-empty, and state, again is the state of QuickOpenModal). So you can safely set aria-expanded on the input to true if enabled prop is true and results state is non-empty, otherwise aria-expanded is false.

yzen:
I think you should be able to solve the aria-activedescendant problem by elevating the calculation of the aria-activedescendant id from ResultList to QuickOpenModal component. You can then just set aria-activedescendant prop in renderInput function since you know (at the time of rendering) the value of selectedIndex and results.

wldcordeiro:
How is this going? @davisv007 and @yzen what would you say is still needed? I could take this under my wing to get it in.

davisv007:
@wldcordeiro My team and I have not met up since Dec. 5th, and I don't expect that we will make much more progress on this issue. So I welcome you to undertake it. If there's anything I can do for you or @yzen to make that transition easier, please let me know.

yzen:
Can be no closed in favour of #5285

Issue #4756, opened 2017-11-20
[Tabs] Remove stray dark line underneath Debugger tab strip


violasong:
Easiest fix is to simply remove the background-color from .editor-pane.

Ideally, could instead figure out why there's this 1px-tall extra bottom spacing in the first place.

Here it is highlighted in red:
![image](https://user-images.githubusercontent.com/55994/33043790-760f79d0-cdfb-11e7-8147-e5d03cbfccb5.png)

Bonus nitpick:
![image](https://user-images.githubusercontent.com/55994/33043836-988a30ea-cdfb-11e7-82c7-97c15e2c58dd.png)
After removing background-color from .editor-pane I see this tiny gray line right at the top of the line number gutter that's making that little area look weird. Would be cool if we could get rid of that too.

pradeepgangwar:
@codehag This issue looks fixed. Please close this to avoid confusion. 😃

violasong:
It looks like this is happening again in Nightly:
<img width="566" alt="image" src="https://user-images.githubusercontent.com/55994/37616237-826f8fd4-2b6c-11e8-8120-3b770e5f199e.png">

jasonLaster:
I'm seeing it in nightly as well

Malachi-Holden:
I am not able to replicate this issue.
I am using Ubuntu 16.04, with Firefox nightly Version 61.0a1 (2018-04-03) (64-bit)
![screenshot from 2018-04-03 14-22-47](https://user-images.githubusercontent.com/35282360/38267959-ec756054-374a-11e8-8521-6eff49c76f71.png)

darkwing:
This should be gone now, I've not seen in ages

PR #4598, opened 2017-11-06
Panel fixes

jasonLaster:
### Summary of Changes

@jbhoosreddy and I looked at some panel bugs today and fixed some of the big ones

* gap in the welcome box
* jump to line when selecting a breakpoint
* jump to line when stepping
* gutter disappearing while stepping

I'll add some mochitests for the gutter and jumping, they're tougher to test because we need to actually test to see if the line is "visible" not just highlighted 

screen shot 2017-11-05 at 6 12 58 pm

PR #4762, opened 2017-11-21
#4756 Remove stray line

cristidrg:
Associated Issue: #4756 

### Summary of Changes
Modified header variable to actually match the given value: https://github.com/devtools-html/debugger.html/blob/master/src/components/Editor/Tabs.css#L8

### Test Plan
Visual change, looks neat now.

### Screenshot
The issue was actually always there, not just in the dark theme. See the picture below:
purple

It was not obvious in the other themes because the background was white.

### DIFF
--- a/src/components/Editor/Editor.css
+++ b/src/components/Editor/Editor.css
@@ -9,7 +9,7 @@
   --editor-second-searchbar-height: 27px;
   --debug-line-error-border: rgb(255, 0, 0);
   --debug-expression-error-background: rgba(231, 116, 113, 0.3);
-  --editor-header-height: 29px;
+  --editor-header-height: 30px;
 }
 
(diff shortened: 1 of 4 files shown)


codehag:
Hi! Thanks for the pr!

A bit of context around this, since I recently merged a pr changing this same line. it used to be 29px, but was changed to 30px because of this bug: https://github.com/devtools-html/debugger.html/pull/4598

the code is visible under the panel in certain situations. Can you do a bit more investigation and see if there is a better solution that will fix both?

cristidrg:
Sure, I will take a look later today !

Thanks for the additional info

codehag:
@cristidrg thanks for taking a look at that! looks good, i will merge once travis agrees!

PR #4766, opened 2017-11-22
Toggle Panes on Pause

GarethSharpe:
### Summary of Changes

* Call-stack pane and scopes pane hidden if debugger is not paused

### Test Plan

- [x] Pausing the debugger reveals call-stack pane and scopes pane
- [x] Clicking the down arrow reveals call-stack and scope(s)
- [x] Resuming the debugger hides call-stack and scopes pane
- [x] Pausing the debugger again displays panes in the same state (open/closed) as the previous pause

A few test cases:
1. Neither pane is open when resumed (clean resume)
2. One of the panes are open when resumed (minor choppy resume)
3. Both panes are open when resumed (visible choppiness on resume)

- Cases 1 and 3 are displayed bellow

toggle-panes

### DIFF
--- a/assets/panel/prefs.js
+++ b/assets/panel/prefs.js
@@ -46,3 +46,4 @@ pref("devtools.debugger.features.root", false);
 pref("devtools.debugger.features.column-breakpoints", false);
 pref("devtools.debugger.features.map-scopes", true);
 pref("devtools.debugger.features.breakpoints-dropdown", false);
+pref("devtools.debugger.features.remove-command-bar-options", false);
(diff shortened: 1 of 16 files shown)


GarethSharpe:
Failed the TI test with the following error:

`TEST-UNEXPECTED-FAIL Not paused message is shown 
 Got , expected Not paused`

I wonder if this test needs to be removed given the pane is hidden, rather then displaying 'Not paused'?

jasonLaster:
Yep - lets remove that check :)

You can try running the tests locally too
https://github.com/devtools-html/debugger.html/blob/master/docs/mochitests.md

jasonLaster:
The PR looks funky. Mind rebasing

PR #4769, opened 2017-11-22
use newSources for sourcemapped sources

codehag:
there was an issue in which new sources that are added via sourcemapped
sources, would disappear on reload.

### DIFF
--- a/src/actions/sources.js
+++ b/src/actions/sources.js
@@ -144,6 +144,9 @@ function loadSourceMap(generatedSource) {
         }: Source)
     );
 
+    // TODO: check if this line is really needed, it introduces
+    // a lot of lag to the application.
+    await dispatch(loadSourceText(generatedSource));
     dispatch(newSources(originalSources));
   };
 }
--- a/src/actions/sources.js
+++ b/src/actions/sources.js
@@ -144,13 +144,7 @@ function loadSourceMap(generatedSource) {
         }: Source)
     );
 
-    dispatch({ type: "ADD_SOURCES", sources: originalSources });
-
-    await dispatch(loadSourceText(generatedSource));
-    originalSources.forEach(async source => {
-      await checkSelectedSource(getState(), dispatch, source);
-      checkPendingBreakpoints(getState(), dispatch, source.id);
-    });
+    dispatch(newSources(originalSources));
   };
 }
 


src/actions/sources.js
```diff
@@ -131,13 +131,7 @@ function loadSourceMap(generatedSource) {
...
     );
 
-    dispatch({ type: "ADD_SOURCES", sources: originalSources });
-
-    await dispatch(loadSourceText(generatedSource));
-    originalSources.forEach(async source => {
-      await checkSelectedSource(getState(), dispatch, source);
-      checkPendingBreakpoints(getState(), dispatch, source.id);
-    });
+    dispatch(newSources(originalSources));
```
jasonLaster: this works except it's missing the `loadSourceText(generatedSource)` for the generated source. 

this would work:

```js
 await dispatch(loadSourceText(generatedSource));
 dispatch(newSources(originalSources));
```

the use case is that pending breakpoints need the generated source text to do the syncing...

codehag: why do we need to load the source text here? we are loading it in newSource after we check the sourcemap, so this is a second check, isnt it?

jasonLaster: This is a small cleanup. I have another PR that removes the need for `loadSourceText`

We'll fix the timing issues in other PRs.

codehag: this comment isn't clear to me

jasonLaster: Oh sorry, I was saying I think we should keep `await dispatch(loadSourceText(generatedSource));`

codehag: I tested this for every edge case I could think of, including ast pending breakpoints, on load and on reload, but there is no difference in behavior by removing this await. 

Why we want to remove this:
generated sources with source maps are often huge, for example bundle.js can be over 40,000 lines. We are loading it here eagerly, which is using resources for something that may never be displayed or used in any way. 

I will make an issue for it so if I missed any possible use case we can test it, but i do not see any reason to keep it. no tests fail, and i could not find any difference while testing.

bomsy: Looks Good! @codehag Do we want create an issue with these details, and link that in TODO comment, so they don't get lost for when we get to removing that line?

codehag: @bomsy sounds like a great idea!

Issue #4773, opened 2017-11-23
[Tabs] the react logo is based off the selected source


jasonLaster:
We currently show a react logo for all the tabs if the selected source is a react component. We should check each tab separately. 

![](http://g.recordit.co/2EjfMRvmU2.gif)

jasonLaster:
At the moment, when you're viewing one tab that is a react file, then all the tabs show the icon. I think we fixed this earlier, but somehow it regressed.


![](http://g.recordit.co/HWJwTcUPLt.gif)

bomsy:
The issue was with getting the sourcemetadata of the selected source rather than the current source.

Here is the some work on the fix.
```diff
diff --git a/src/components/Editor/Tab.js b/src/components/Editor/Tab.js
index ccbec14e..469021b5 100644
--- a/src/components/Editor/Tab.js
+++ b/src/components/Editor/Tab.js
@@ -47,7 +47,7 @@ type Props = {
   showSource: string => void,
   source: SourceRecord,
   activeSearch: string,
-  sourceMetaData: string => any
+  getMetaData: string => any
 };

 class Tab extends PureComponent<Props> {
@@ -144,7 +144,7 @@ class Tab extends PureComponent<Props> {
       selectSource,
       closeTab,
       source,
-      sourceMetaData
+      getMetaData
     } = this.props;
     const src = source.toJS();
     const filename = getFilename(src);
@@ -154,7 +154,7 @@ class Tab extends PureComponent<Props> {
       sourceId == selectedSource.get("id") &&
       (!this.isProjectSearchEnabled() && !this.isSourceSearchEnabled());
     const isPrettyCode = isPretty(source);
-    const sourceAnnotation = getSourceAnnotation(source, sourceMetaData);
+    const sourceAnnotation = getSourceAnnotation(source, getMetaData(sourceId));

     function onClickClose(e) {
       e.stopPropagation();
@@ -202,10 +202,7 @@ export default connect(
       closeTab,
       source,
-      sourceMetaData
+      getMetaData
     } = this.props;
     const src = source.toJS();
     const filename = getFilename(src);
@@ -154,7 +154,7 @@ class Tab extends PureComponent<Props> {
       sourceId == selectedSource.get("id") &&
       (!this.isProjectSearchEnabled() && !this.isSourceSearchEnabled());
     const isPrettyCode = isPretty(source);
-    const sourceAnnotation = getSourceAnnotation(source, sourceMetaData);
[...]

PR #4779, opened 2017-11-23
Display React logo only on active tab

fivepointseven:
Associated Issue: #4773

### Summary of Changes

* Fixed displaying react logo on non-React tabs when React tab is selected

### DIFF
--- a/src/components/Editor/Tabs.js
+++ b/src/components/Editor/Tabs.js
@@ -32,7 +32,6 @@ import Dropdown from "../shared/Dropdown";
 import type { List } from "immutable";
 import type { SourceRecord } from "../../reducers/sources";
 import type { ActiveSearchType } from "../../reducers/ui";
-import type { SourceMetaDataMap } from "../../reducers/ast";
 type SourcesList = List<SourceRecord>;
 
 /*
@@ -98,7 +97,7 @@ type Props = {
   startPanelCollapsed: boolean,
   endPanelCollapsed: boolean,
   searchOn: boolean,
-  sourceMetaData: SourceMetaDataMap
+  sourceTabsMetaData: any
 };
 
 type State = {
@@ -389,7 +388,7 @@ class SourceTabs extends PureComponent<Props, State> {
       source.get("id") == selectedSource.get("id") &&
       (!this.isProjectSearchEnabled() && !this.isSourceSearchEnabled());
     const isPrettyCode = isPretty(source.toJS());
-    const sourceAnnotation = this.getSourceAnnotation(source, active);
+    const sourceAnnotation = this.getSourceAnnotation(source);
 
     function onClickClose(ev) {
       ev.stopPropagation();
@@ -455,11 +454,12 @@ class SourceTabs extends PureComponent<Props, State> {
     );
   }
 
-  getSourceAnnotation(source, active) {
+  getSourceAnnotation(source) {
     const sourceObj = source.toJS();
-    const { sourceMetaData } = this.props;
+    const sourceId = source.get("id");
+    const sourceMetaData = this.props.sourceTabsMetaData[sourceId];
 
-    if (sourceMetaData && sourceMetaData.isReactComponent && active) {
+    if (sourceMetaData && sourceMetaData.isReactComponent) {
       return <img className="react" />;
     }
     if (isPretty(sourceObj)) {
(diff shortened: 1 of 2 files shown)


fivepointseven:
This has room for improvement beyond just fixing the bug (I'd prefer to have framework logos on the appropriate tabs all the time like in a code editor) but I've been familiar with the project for about 30 minutes, so I only did the bug for now :D

jasonLaster:
@fivepointseven I see what you're saying! And i see how you fixed the problem, which is a step in the right direction. I'd like to take a stab at showing the correct logo for all the tabs though. 

I'm traveling, so i'm a bit less responsive, but I bet you can try getting the source meta data for all the tabs. It's a bit inefficient the way we're doing it now, but we can refactor it later.

fivepointseven:
@jasonLaster I tried doing that yesterday and came up with a bunch of different ways to do it actually, but the problem is that for some reason `isReactComponent` was `false` the whole evening for me even with the current code. Weird.
Anyway I’ll try to get that to work and commit it later today :)

fivepointseven:
@jasonLaster okay, done :)
React logos are now consistently displayed:

src/components/Editor/Tabs.js
```diff
@@ -484,16 +484,20 @@ class SourceTabs extends PureComponent<Props, State> {
 
 export default connect(
   state => {
-    const selectedSource = getSelectedSource(state);
-    const sourceId = selectedSource ? selectedSource.get("id") : "";
+    const sourceTabs = getSourcesForTabs(state);
+    const sourceTabsMetaData = {};
+    sourceTabs.forEach(source => {
+      const sourceId = source ? source.get("id") : "";
```
codehag: im wondering, what is the case when a source does not have an id? do you have an example?

also, meta thought: it feels like the meta data belongs to the tab as a data structure too..

src/components/Editor/Tabs.js
```diff
@@ -98,7 +97,7 @@ type Props = {
   startPanelCollapsed: boolean,
   endPanelCollapsed: boolean,
   searchOn: boolean,
-  sourceMetaData: SourceMetaDataMap
+  sourceTabsMetaData: any
```
codehag: hmmm perhaps we can make a type annotation?

fivepointseven: Roger that

src/components/Editor/Tabs.js
fivepointseven: This was copied from the latest code, so I'm not entirely sure:
```diff
- const selectedSource = getSelectedSource(state);
- const sourceId = selectedSource ? selectedSource.get("id") : "";
+ sourceTabs.forEach(source => {
+   const sourceId = source ? source.get("id") : "";
```
Is there a case where source doesn't have an id?

----
A funny thing about metadata btw, it's set on "LOAD_SOURCE_TEXT", which means that in order for `isReactComponent` to be set tab needs to be visited at least once (if you open a few tabs and reload the window, logos won't show up until you click on respective tabs). It doesn't affect UX that much, and when I tried moving it to reducer it disintegrated my CPU, so I probably didn't entirely understand how it works outside of Tabs component, so I didn't touch it.

jasonLaster: @fivepointseven that's a good observation. We should probably re-think source metadata so it is a field in sources and tabs, so that when we re-show a tab on reload it is saved...

that way you don't need to revisit it.

fivepointseven: @jasonLaster What's the starting point in source's "life"? Cause I tried adding that at `ADD_SOURCES` action but it ended up eating all of my resources over 10 seconds after loading so I probably did it wrong.

jasonLaster: haha - yeah. we don't want to parse every source! it would kill us :)

don't worry about that bug now, I'll add some context in another issue

fivepointseven: Okay, cool then. So, do you have any issues with this PR I should fix? :)

src/workers/parser/frameworks.js
```diff
@@ -5,20 +5,22 @@
...
-    importObj =>
-      importObj.source === "react" &&
-      importObj.specifiers.some(specifier => specifier === "React")
+function importsReact(imports, identifiers) {
+  return (
+    imports.some(
+      importObj =>
+        importObj.source === "react" &&
+        importObj.specifiers.some(specifier => specifier === "React")
+    ) || identifiers.some(identifier => identifier.name === "React")
```
jasonLaster: this is an interesting change. Why is this needed? Also, could we add some new tests that would fail w/o it

jasonLaster:
Minor point, when you open your next PR use a non master branch as it's easier for us to pull down and to push changes to

https://github.com/devtools-html/debugger.html/blob/master/docs/pull-requests.md#git-workflow

src/workers/parser/frameworks.js
jasonLaster: I bet this change could be in its own PR actually.

jasonLaster:
Just checked and it works nicely! 

Happy to land it after we discuss the parser change

src/workers/parser/frameworks.js
fivepointseven: This is what caused the problem I mentioned earlier:
>  for some reason isReactComponent was false the whole evening for me even with the current code

`import React` was expected here and it didn't trigger by `const React = require('react')`, so I added the identifier check as well

jasonLaster: Ahh brilliant. Lets do this in a separate PR. It fixes another problem :)

jasonLaster: Actually, #4737 is already working on it

fivepointseven: Oh, didnt realize that
I just kinda did it cause it was interfering with what I was doing :D

jasonLaster: Yep! I do that all the time too :)

glad you got to see the parser code. It's my favorite part of the codebase

Issue #4608, opened 2017-11-07
[Editor] add selectLocation


jasonLaster:
We use `actions.selectSource(sourceId, {location: {line, column}})` in a lot of places. I mean a lot a lot of places: stepping, selecting a breakpoint, search, ....

It would be nice to have a simpler API that assumes you want to jump to a `Location` as defined in `debugger-html`.

```js
actions.selectLocation(location);
```

for the time being we can have it call `selectSource`.

```js
 export function selectLocation(location: Location) {
   return selectSource(location.sourceId, { location });
}
```

Refactoring this function will require refactoring many of the places where `selectSource` is used:

```diff
    selectBreakpoint(breakpoint) {
-     const sourceId = breakpoint.location.sourceId;
-     const line = breakpoint.location.line;
-     this.props.selectSource(sourceId, { line });
+     this.props.selectLocation(breakpoint.location);
    }
```  

This is just a small code cleanliness thing.

PR #4789, opened 2017-11-25
Replaces selectSource with selectLocation

lukaszsobek:
Associated Issue: #4608

### Summary of Changes

* Replaces occurences of selectSource with selectLocation except in mochitests
* Reformats input to selectLocation to conform with the new requirements

### Test Plan
- yarn test
- yarn flow check

Still needs mochitest

### DIFF
--- a/assets/panel/panel.js
+++ b/assets/panel/panel.js
@@ -98,8 +98,9 @@ DebuggerPanel.prototype = {
     return { frames, selected };
   },
 
-  selectSource(sourceURL, sourceLine) {
-    this._actions.selectSourceURL(sourceURL, { location: { line: sourceLine }});
+  selectLocation(sourceOptions) {
+    const { id, line } = sourceOptions;
+    this._actions.selectSourceURL(id, { location: { line } });
   },
 
   getSource(sourceURL) {
--- a/assets/panel/panel.js
+++ b/assets/panel/panel.js
@@ -98,9 +98,8 @@ DebuggerPanel.prototype = {
     return { frames, selected };
   },
 
-  selectLocation(sourceOptions) {
-    const { id, line } = sourceOptions;
-    this._actions.selectSourceURL(id, { location: { line } });
+  selectSource(url, line) {
+    this._actions.selectSourceURL(url,  { line });
   },
 
   getSource(sourceURL) {
--- a/src/actions/sources.js
+++ b/src/actions/sources.js
@@ -52,12 +52,12 @@ async function checkSelectedSource(state: State, dispatch, source) {
   const pendingLocation = getPendingSelectedLocation(state);
 
   if (pendingLocation && !!source.url && pendingLocation.url === source.url) {
-    await dispatch(selectSource(source.id, { location: pendingLocation }));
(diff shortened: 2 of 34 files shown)


bomsy:
@lukaszsobek there is a `selectSource` in this file `devtools/client/debugger/new/test/mochitest/browser_dbg-console-link.js` causing the travis error. You might need to switch that to `selectLocation` as well.

lukaszsobek:
@bomsy - some of these seem to call the mochitest version of selectSource (which is intentionally left as selectSource to distinguish it from selectLocation). Will get to the mochitest as soon as I get it installed.

src/actions/sources.js
```diff
@@ -174,19 +174,28 @@ export function selectSourceURL(
...
+  column?: ?number,
+  sourceUrl?: ?string
+};
+
 /**
  * @memberof actions/sources
  * @static
  */
-export function selectSource(id: string, options: SelectSourceOptions = {}) {
+export function selectLocation(location: LocationObject) {
```
bomsy: @lukaszsobek for now don't replace this function as it is used in MC here 
https://dxr.mozilla.org/mozilla-central/source/devtools/client/shared/view-source.js#65
maybe just create a new `selectLocation` function and call `selectSource` from there

bomsy: Might have to change that eventually

assets/panel/panel.js
```diff
@@ -98,8 +98,9 @@ DebuggerPanel.prototype = {
     return { frames, selected };
   },
 
-  selectSource(sourceURL, sourceLine) {
-    this._actions.selectSourceURL(sourceURL, { location: { line: sourceLine }});
+  selectLocation(sourceOptions) {
```
bomsy: @lukaszsobek for now don't replace this function, as it is used in MC as well here 
https://dxr.mozilla.org/mozilla-central/source/devtools/client/shared/view-source.js#65.
Maybe just create a new `selectLocation` function and call `selectSource` from there.
That's probably the issue with the tests as well

src/actions/sources.js
bomsy: @lukaszsobek sorry! ignore this comment here... i've move the comment up to the right location [here](https://github.com/devtools-html/debugger.html/pull/4789/files#r153234382)

jasonLaster:
just pushed a rebase

jasonLaster:
@lukaszsobek good news, i think everything is working. Lets see if we go green :)

lukaszsobek:
All green :)

jasonLaster:
This is amazing work. Thank you so much for helping us upgrade this API.

Changes like this are how we keep the code base a place we are happy to work in as opposed to letting it slip into something that we just shrug and accept. 

There is no way, that I could have found the time or energy to complete a migration of this size. So, I'm really grateful that you were able to do it. 

I hope you learned a bit along the way about how some of these components work.

lukaszsobek:
It was so good I even wrote a blog post about it. An incredibly rewarding experience! Thank you.

bomsy:
Nice one! Share your blogpost with us

Issue #4551, opened 2017-11-01
backslash \ shall be allowed in expressions


yurydelendik:
STR:
1. Add `N\u{61}N` as watch expression
2. Add `'\u{61}'` as watch expression

Expected results:
1. `NaN`
2. `"a"`

Actual results:
1. Syntax Error
2. `"\\u{61}"`

yurydelendik:
The sanitizeInput function has to be removed at https://github.com/devtools-html/debugger.html/blob/495339a8e958051e1aaee949fc22b6e12b6a7f48/src/utils/expressions.js#L6-L9

PR #4790, opened 2017-11-25
Permits backslash in expressions

lukaszsobek:
Associated Issue: #4551

Here's the Pull Request Doc

### Summary of Changes

* Removed replace of backslashes in expressions.js
* Removed test case referencing the replace in expressions.spec.js

screenshot-20171125111327-326x275

### DIFF
--- a/src/utils/tests/expressions.spec.js
+++ b/src/utils/tests/expressions.spec.js
@@ -29,6 +29,14 @@ describe("expressions", () => {
     it("sanitizes 2 quotes", () => {
       expect(sanitizeInput('"3"')).toEqual('"3"');
     });
+
+    it("evaluates \\u{61} as a", () => {
+      expect(sanitizeInput("\u{61}")).toEqual("a");
+    });
+
+    it("evaluates N\\u{61}N as NaN", () => {
+      expect(sanitizeInput("N\u{61}N")).toEqual("NaN");
+    });
   });
 
   describe("getValue", () => {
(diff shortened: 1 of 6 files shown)


src/utils/tests/expressions.spec.js
```diff
@@ -29,10 +29,6 @@ describe("expressions", () => {
     it("sanitizes 2 quotes", () => {
       expect(sanitizeInput('"3"')).toEqual('"3"');
     });
-
-    it("sanitizes forward slashes", () => {
-      expect(sanitizeInput("foo\\\\")).toEqual("foo\\\\\\\\");
-    });
   });
```
jasonLaster: lets add the two cases

src/utils/tests/expressions.spec.js
```diff
@@ -29,10 +29,6 @@ describe("expressions", () => {
...
 
   describe("getValue", () => {
```
jasonLaster: and are there others that we need? perhaps where we want to guard the back slash?

src/utils/tests/expressions.spec.js
codehag: yep, we need to add the two test cases!

  1.  Add N\u{61}N as watch expression
  2.  Add '\u{61}' as watch expression

Expected results:

  1.  NaN
  2.  "a"

lukaszsobek: Test cases added

src/utils/tests/expressions.spec.js
lukaszsobek: I can't think of any cases where we would need it. If you have any I'm more than happy to add them.

Issue #4767, opened 2017-11-22
Duplicate conditional-breakpoint-panel when send "Ctrl + B"


magicp-github:
**Steps to reproduce:**
1. Open any scripts
2. Focus (set cursor) on any lines
3. Add conditional breakpoint on the same line
4. Send "Ctrl + B"

**Actual Results:**
conditional-breakpoint-panel will be duplicated.

![duplicate-conditional-breakpoint-panel](https://user-images.githubusercontent.com/13914314/33115954-3e8e173a-cfa6-11e7-9d45-bb3142f651ae.png)

**Expected Results:**
No any actions?

nyrosmith:
@magicp-github Did you see this behavior in launchpad or Firefox panel? In launchpad I am unable to reproduce it

magicp-github:
I can reproduce in the latest Nightly build (20171123100420).

nyrosmith:
@magicp-github can you reproduce it in launchpad too? I tried several times but it seems to me it's already fixed here

magicp-github:
Yes, I can reproduce it in launchpad too.

nyrosmith:
Hmm maybe I am doing something wrong. Habe you already joined our slack? Maybe we can do a quick hangout so this can get tackled fast

magicp-github:
I don't use slack. Step 2-3 is important. Focused line and conditional breakpoint should be on the same line. "Ctrl+B" will make a breakpoint on the focused line, then conditional breakpoint panel will be duplicated.

nyrosmith:
Thx for clarification!

PR #4793, opened 2017-11-26
Editor: close ConditionalPanel before adding breakpoint

nyrosmith:
Associated Issue: #4767 
### Summary of Changes

* Check if ConditonalPanel is open and closing it before trying to set a regular breakpoint

### DIFF
--- a/src/components/Editor/index.js
+++ b/src/components/Editor/index.js
@@ -265,7 +265,7 @@ class Editor extends PureComponent<Props, State> {
 
   onToggleBreakpoint = (key, e) => {
     e.preventDefault();
-    const { selectedSource } = this.props;
+    const { selectedSource, conditionalPanelLine } = this.props;
 
     if (!selectedSource) {
       return;
@@ -275,7 +275,10 @@ class Editor extends PureComponent<Props, State> {
 
     if (e.shiftKey) {
       this.toggleConditionalPanel(line);
+    } else if (!conditionalPanelLine) {
+      this.props.toggleBreakpoint(line);
     } else {
+      this.toggleConditionalPanel(line);
       this.props.toggleBreakpoint(line);
     }
   };


src/components/Editor/index.js
```diff
@@ -275,7 +275,10 @@ class Editor extends PureComponent<Props, State> {
 
     if (e.shiftKey) {
       this.toggleConditionalPanel(line);
+    } else if (!conditionalPanelLine) {
+      this.props.toggleBreakpoint(line);
     } else {
+      this.toggleConditionalPanel(line);
       this.props.toggleBreakpoint(line);
     }
```
jasonLaster: I think this might be a good way to do it:

```js
   if (e.shiftKey) {
       this.toggleConditionalPanel(line);
     } else {
       if (conditionalPanelLine) {
          this.toggleConditionalPanel(line);
       }
       this.props.toggleBreakpoint(line);
     }
```

Perhaps an even better way to do it would be

```js
   if (e.shiftKey) {
       this.toggleConditionalPanel(line);
     } else {
       this.props.toggleBreakpoint(line);
     }
```
and then reset the conditional panel in the redux action

nyrosmith: Your first suggested attempt is forbidden by some EsLint rule and I really wouldn't move it into the toggleBreakpoint  action as it makes things really unclear

nyrosmith: We added all of this kind of logic into actions on another project I worked on and things became really nasty over time. We refactored all of that just to get back to a healthy source.
Just think to how this action should be named to reflect its purpose: `addOrRemoveBreakpointAndCloseConditionalPanelIfOpen`
This sounds to me this action would do way to much...

jasonLaster: > addOrRemoveBreakpointAndCloseConditionalPanelIfOpen

lol

It's true, that we have been moving some of our component complexity to the actions. It's quite possible that we will find that they become too complicated as well.

Another interesting location for this would be the reducers, which could listen for ADD_BREAKPOINT, and reset the conditional pane.

Either way, we are splitting hairs here and i'm happy leaving the logic in the component.

codehag: Continuing with splitting hairs: it seems like `toggleConditionalPanel` is doing too much, as it can open or close the panel. We probably need a simple `closeConditionalPanel` and it should be idempotent. In other words, regardless of what the state of the conditional panel is, the result should be a closed panel. this would get rid of the else if case and would make the code more precise. We should move away from using `toggle`s in any case, unless it is in very special circumstances, since it is too ambiguous for most use cases. 

Doing it in a reducer is a good idea and would likely be simplest to implement. Actions should describe what happened, they shouldn't be used as setters. I don't know if we have a reducer for this logic, but if we do, that could be a clean way of ensuring that it is always closed.

Issue #4794, opened 2017-11-26
[Tabs] New tabs open at tabIndex 0


lukaszsobek:
STR: 
- Open File by clicking it in the sources panel
- Wait for file to load in editor
- Open another file by clicking it in the sources panel

ER:
- New tab opens at the end of the list of tabs

Rationale:
- Almost every text editor opens tabs the ER way. Giving the user a feeling of familiarity and security should increase user satisfaction. 

AR:
- New tab opens with a tab index of 0

![ezgif-4-4f813d1d84](https://user-images.githubusercontent.com/23530054/33239511-6beed8fe-d2a3-11e7-8d50-7fadb97823d8.gif)

lukaszsobek:
Requires further UI/UX discussion.

PR #4797, opened 2017-11-26
Makes new tabs open after old ones

lukaszsobek:
Associated Issue: #4794

### Summary of Changes

* Makes new tabs open after old ones by checking the size of the tabs list and using it as reference for opening new ones

### Screenshots/Videos
new-tab

### DIFF
--- a/src/reducers/sources.js
+++ b/src/reducers/sources.js
@@ -216,7 +216,7 @@ function updateTabList(state: OuterState, url: ?string, tabIndex?: number) {
       tabs = tabs.delete(urlIndex).insert(tabIndex, url);
     }
   } else {
-    tabs = tabs.insert(0, url);
+    tabs = tabs.insert(tabs.size, url);
   }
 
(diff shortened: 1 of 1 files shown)


jasonLaster:
Nice!

I think this is an improvement, but i'm hesitant landing this w/o thinking through the UX of having the hidden tabs dropdown

jasonLaster:
lets close this as we're blocked on UX and come back once we spec it out

src/reducers/sources.js
```diff
@@ -216,7 +216,7 @@ function updateTabList(state: OuterState, url: ?string, tabIndex?: number) {
       tabs = tabs.delete(urlIndex).insert(tabIndex, url);
     }
   } else {
-    tabs = tabs.insert(0, url);
+    tabs = tabs.insert(tabs.size, url);
```
Zacqary: `tabs.push(url)` might be clearer about what this is doing

Issue #4694, opened 2017-11-14
[Editor] detect if files use JSX


jasonLaster:
The debugger currently detects if a file includes a react component and uses that to assume we should use JSX mode in the editor.

The better way to do this would be to check to see if the file has a jsx element in the text and if so use jsx mode. The rationale is that other frameworks like preact also use JSX.

here is an ast for jsx:
https://astexplorer.net/#/gist/d390e9a9b92911d197385283f4ca5754/0602aa2592d9562097b4f0043152963a789c134c

jasonLaster:
```diff
diff --git a/src/reducers/ast.js b/src/reducers/ast.js
index 29278c8..e396c61 100644
--- a/src/reducers/ast.js
+++ b/src/reducers/ast.js
@@ -150,6 +150,11 @@ export function getSymbols(
   return symbols || emptySet;
 }
 
+export function hasJSX(state: OuterState, source: Source) {
+  const symbols = getSymbols(state, source);
+  return symbols.hasJSX;
+}
+
 export function hasSymbols(state: OuterState, source: Source): boolean {
   if (!source) {
     return false;
diff --git a/src/workers/parser/getSymbols.js b/src/workers/parser/getSymbols.js
index 89ac650..e588795 100644
--- a/src/workers/parser/getSymbols.js
+++ b/src/workers/parser/getSymbols.js
@@ -104,6 +104,7 @@ function extractSymbols(source: Source) {
   const identifiers = [];
   const classes = [];
   const imports = [];
+  const hasJSX = false;
 
   const ast = traverseAst(source, {
     enter(path: NodePath) {
@@ -121,6 +122,10 @@ function extractSymbols(source: Source) {
         });
       }
 
+      if (isJSXElement(path)) {
+        hasJSX = true;
+      }
+
       if (t.isClassDeclaration(path)) {
         classes.push({
           name: path.node.id.name,
@@ -218,7 +223,8 @@ function extractSymbols(source: Source) {
     comments,
     identifiers,
     classes,
-    imports
+    imports,
+    hasJSX
   };
 }
 
diff --git a/src/workers/parser/tests/getSymbols.spec.js b/src/workers/parser/tests/getSymbols.spec.js
index 58850c6..b62e750 100644
--- a/src/workers/parser/tests/getSymbols.spec.js
+++ b/src/workers/parser/tests/getSymbols.spec.js
@@ -25,6 +25,7 @@ cases(
     },
     { name: "component", file: "component" },
[...]

PR #4800, opened 2017-11-27
[WIP] detect if files use JSX

DavideDeCenzo:
Associated Issue: #4694 

Here's the Pull Request Doc

Feel free to improve it!


src/workers/parser/getSymbols.js
```diff
@@ -5,14 +5,15 @@
...
-import { isVariable, isFunction } from "./utils/helpers";
+import { isVariable, isFunction, isJSXElement } from "./utils/helpers";
 import { inferClassName } from "./utils/inferClassName";
 import * as t from "babel-types";
 import getFunctionName from "./utils/getFunctionName";
 
 import type { Source } from "debugger-html";
 import type { NodePath, Node, Location as BabelLocation } from "babel-traverse";
 let symbolDeclarations = new Map();
+var hasJSX = false;
```
jasonLaster: this should be defined in the function it's used
also, we use `let` and `const` :)

src/workers/parser/getSymbols.js
```diff
@@ -121,6 +126,10 @@ function extractSymbols(source: Source) {
         });
       }
 
+      if (isJSXElement(path)) {
+        hasJSX = true;
```
jasonLaster: hmm, i think we should store it the same way as `imports` & `classes`

src/workers/parser/getSymbols.js
```diff
@@ -95,6 +96,10 @@ function getSpecifiers(specifiers) {
   return specifiers.map(specifier => specifier.local && specifier.local.name);
 }
 
+export function getHasJSX(): boolean {
+  return hasJSX;
```
jasonLaster: this function won't know the difference between sources...

src/workers/parser/tests/getSymbols.spec.js
```diff
@@ -25,6 +25,7 @@ cases(
     },
     { name: "component", file: "component" },
     { name: "react component", file: "frameworks/component" },
-    { name: "flow", file: "flow" }
+    { name: "flow", file: "flow" },
+    { name: "jsx", file: "frameworks/jsx" }
```
jasonLaster: you should run `jest -u src/workers/parser` to see this run.

Also, because each test uses `formatSymbols` we should make sure it shows the boolean

src/workers/parser/utils/helpers.js
```diff
@@ -67,3 +67,10 @@ export function getMemberExpression(root: Node) {
   const expr = _getMemberExpression(root, []);
   return expr.join(".");
 }
+
+export function isJSXElement(path: NodePath): boolean {
+  if (t.isJSXElement(path)) {
+    return true;
+  }
+  return false;
+}
```
jasonLaster: we can simplify this to `return t.isJSXElement(path)` at which point i am not sure if we need to have a helper function :)

jasonLaster:
looks like it needs a rebase
https://github.com/devtools-html/debugger.html/blob/master/docs/pull-requests.md#git-workflow

DavideDeCenzo:
@jasonLaster  Thanks for the comments ,  I'll do a rebase then  I didn't know what happened

jasonLaster:
haha - no worries. OSS is all about learning GIT

DavideDeCenzo:
Ok this should be better @jasonLaster

src/workers/parser/getSymbols.js
```diff
@@ -121,6 +122,17 @@ function extractSymbols(source: Source) {
         });
       }
 
+      if (t.isJSXElement(path)) {
+        hasJSX.push({
+          location: path.node.loc,
+          openingElement: path.node.openingElement,
+          closingElement: path.node.closingElement,
+          children: path.node.children
+        });
```
jasonLaster: could it just be `hasJSX = true` :)

src/workers/parser/getSymbols.js
```diff
@@ -235,6 +248,11 @@ export default function getSymbols(source: Source): SymbolDeclarations {
   return symbols;
 }
 
+export function getHasJSX(source: Source): boolean {
+  const symbols = getSymbols(source);
+  return symbols.hasJSX.length > 0;
```
jasonLaster: we should store this property in `sourceMetaData` in `reducers/ast` along side the `isReact` field. This way we don't need to ask the worker every time

DavideDeCenzo:
I'm not sure of these changes but I think it's worth trying :) @jasonLaster

src/workers/parser/getSymbols.js
```diff
@@ -121,6 +122,12 @@ function extractSymbols(source: Source) {
         });
       }
 
+      if (t.isJSXElement(path)) {
+        setJSXMetaData(source, true);
```
jasonLaster: `setJSXMetaData` unfortunately won't work because you can't call reducer methods directly.

I recommend adding a `jsx` field to the object that is returned. That way, we'll be able to look up the jsx value wherever symbols are used

jasonLaster:
Hey @DavideDeCenzo I'd be happy to review this with you if you're free to chat. Find me on slack :)

jasonLaster:
@DavideDeCenzo can we create a new PR for just the parser work?

DavideDeCenzo:
Yes , I create it with the last changes @jasonLaster

PR #4802, opened 2017-11-27
Do work in the background

jasonLaster:
Associated Issue: #4705 

### Summary of Changes

* Removes await 
* moves loadSourceText to checkPendingBreakpoints (we might not need it at all)

### DIFF
--- a/src/actions/sources.js
+++ b/src/actions/sources.js
@@ -41,8 +41,7 @@ import {
   removeSourcesFromTabList,
   removeSourceFromTabList,
   getTextSearchQuery,
-  getActiveSearch,
-  getGeneratedSource
+  getActiveSearch
 } from "../selectors";
 
 import type { Source, Location } from "../types";
@@ -71,16 +70,12 @@ function checkPendingBreakpoints(sourceId) {
       getState(),
       source.url
     );
+
     if (!pendingBreakpoints.size) {
       return;
     }
 
     await dispatch(loadSourceText(source));
-
-    if (isOriginalId(source.id)) {
-      const generatedSource = getGeneratedSource(getState(), source);
-      await dispatch(loadSourceText(generatedSource.toJS()));
-    }
 
     const pendingBreakpointsArray = pendingBreakpoints.valueSeq().toJS();
     for (const pendingBreakpoint of pendingBreakpointsArray) {
--- a/src/actions/sources.js
+++ b/src/actions/sources.js
@@ -27,6 +27,7 @@ import { prefs } from "../utils/prefs";
 import { removeDocument } from "../utils/editor";
 import { isThirdParty, isMinified, shouldPrettyPrint } from "../utils/source";
 import { getGeneratedLocation } from "../utils/source-maps";
+import { isOriginalId } from "devtools-source-map";
 
 import {
   getSource,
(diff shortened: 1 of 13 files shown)


jasonLaster:
Looks pretty close (2 test fails)

src/actions/sources.js
```diff
@@ -66,6 +67,12 @@ async function checkPendingBreakpoints(state, dispatch, sourceId) {
 
   // load the source text if there is a pending breakpoint for it
   await dispatch(loadSourceText(source));
+
+  if (isOriginalId(source.id)) {
+    const generatedSource = getGeneratedSource(state, source);
+    await dispatch(loadSourceText(source.toJS()));
+  }
```
jasonLaster: @codehag persuaded me last week that we might not need this :)

src/actions/sources.js
```diff
@@ -145,9 +152,8 @@ function loadSourceMap(generatedSource) {
 
     dispatch({ type: "ADD_SOURCES", sources: originalSources });
 
-    await dispatch(loadSourceText(generatedSource));
```
jasonLaster: @codehag in another PR i mentioned that we should leave this line in because I was removing it here :)

In this PR, as it is now, we can remove this because we load it if we're syncing pending BPs for a source. We might not need it there either!

If we discover that we do need it, and for some reasons our tests pass without it, we should update our tests!

jasonLaster:
Looks like this affects our mochitests more (which is why i love integration tests)!

my guess is that we'll be able to update our `head.js` helpers and resolve some of the timing issues

* browser_dbg-editor-highlight.js
* browser_dbg-minified.js
* browser_dbg-sourcemaps-reloading.js
* browser_dbg-sourcemaps2.js
* browser_dbg-sourcemaps3.js
* browser_dbg-sourcemaps3.js

jasonLaster:
Made some tweaks which fixed editor-highlight, but the other mochitests still need some tending too...

src/actions/sources.js
```diff
@@ -40,37 +41,52 @@ import {
...
 import type { Source, Location } from "../types";
 import type { ThunkArgs } from "./types";
-import type { State } from "../reducers/types";
 
 // If a request has been made to show this source, go ahead and
 // select it.
-async function checkSelectedSource(state: State, dispatch, source) {
-  const pendingLocation = getPendingSelectedLocation(state);
+function checkSelectedSource(source) {
+  return async ({ dispatch, getState }: ThunkArgs) => {
```
jasonLaster: I made both `checkSelectedSource` and `checkPendingBreakpoints` action creators because that simplified things / made them more consistent.

src/actions/sources.js
```diff
@@ -40,37 +41,48 @@ import {
...
+    const pendingBreakpoints = getPendingBreakpointsForSource(
+      getState(),
+      source.url
+    );
+
+    if (!pendingBreakpoints.size) {
+      return;
+    }
+
+    await dispatch(loadSourceText(source));
```
bomsy: Curious! Why does the source needs to be loaded for checking the pending breakpoints?

jasonLaster: glad you asked :) because pending breakpoints need the symbols to compute the ast locations. thus we need to load the file

src/actions/sources/loadSourceText.js
```diff
@@ -16,13 +17,11 @@ const requests = new Map();
 
 async function loadSource(source: Source, { sourceMaps, client }) {
   if (sourceMaps.isOriginalId(source.id)) {
```
bomsy: switch to `isOriginalId` here

src/actions/sources/loadSourceText.js
```diff
@@ -15,14 +16,12 @@ import type { ThunkArgs } from "../types";
 const requests = new Map();
 
 async function loadSource(source: Source, { sourceMaps, client }) {
-  if (sourceMaps.isOriginalId(source.id)) {
-    return await sourceMaps.getOriginalSourceText(source);
+  if (isOriginalId(source.id)) {
+    return sourceMaps.getOriginalSourceText(source);
```
bomsy: why is `sourceMaps` used here and passed into functions in other places? if we  are importing `devtools-source-map` already at the top, could we not just pull in all the other functions
e.g `import { isOriginalId, getOriginalSourceText } from "devtools-source-map";` or am i missing something?

bomsy: ahhh ..its for the workers, right?

jasonLaster: yeah - and the source map worker is different from the others because it is shared w/ the other panels

bomsy: cool thanks!

jasonLaster:
Overall load time is 25% less and there are fewer blocking frames (red bars)

### Before the change





### After the change

bomsy:
Ahhh thanks for this!!

src/actions/sources.js
```diff
@@ -112,13 +123,28 @@ export function newSources(sources: Source[]) {
...
-      await checkSelectedSource(getState(), dispatch, source);
-      await checkPendingBreakpoints(getState(), dispatch, source.id);
+      dispatch(loadSourceMap(source));
+      dispatch(checkSelectedSource(source));
+      dispatch(checkPendingBreakpoints(source.id));
     }
   };
 }
 
+function createOriginalSource(
```
codehag: :+1:

Issue #4703, opened 2017-11-16
Conditional breakpoint overlay is cut off in vertical mode


digitarald:
STR:

- Vertical mode in Debugger
- Make breakpoint conditional

ER: Correctly positioned conditional breakpoint form, responsively fitting into the space available.

AR: <img width="661" alt="image" src="https://user-images.githubusercontent.com/8599/32876270-ef4d12a2-ca50-11e7-95ba-34d857b8e596.png">

borian:
Mhh, I was not able to reproduce the issue

![condition-breakpoint](https://user-images.githubusercontent.com/1525694/32994083-a83822de-cd62-11e7-9309-b48ae3cbf695.png)

magicp-github:
Probably, CodeMirror was scrolled horizontally in AR image?

Zacqary:
It looks like there's a responsiveness bug even outside of vertical mode. Notice how the close button is obscured because it's too far to the right:
![screen shot 2017-11-27 at 12 17 49 pm](https://user-images.githubusercontent.com/1445834/33282064-0a7e8ebe-d36d-11e7-9299-269017cf3410.png)

This is fixable in horizontal mode by setting the `CodeMirror-linewidget` to have `width: calc(100% - 30px)` but that doesn't work in vertical mode when the user can scroll horizontally. 

I think we need to clear up what the expected behavior is in vertical mode. Do we want something like this?
![screen shot 2017-11-27 at 12 21 50 pm](https://user-images.githubusercontent.com/1445834/33282244-b1409800-d36d-11e7-9636-388bfbea82a3.png)
and have it fixed in that position as the user scrolls to the right?
![screen shot 2017-11-27 at 12 23 38 pm](https://user-images.githubusercontent.com/1445834/33282309-dbb4acac-d36d-11e7-9583-d626e96a04c3.png)

(This is a mockup, I haven't yet figured out how to do that in proper CSS)

jasonLaster:
👋 @Zacqary!

We have a UI redux field for the sidebar widths. Could help...

Zacqary:
So here's what I was able to achieve by adding `padding-right: 16px` to `.conditional-breakpoint-panel` and then removing the `min-width` from `.CodeMirror-sizer` — the condition input field is now fully responsive, and the only thing that's still broken is horizontal scrolling:

[...]

PR #4808, opened 2017-11-27
Conditional breakpoint panel responsiveness and visual fixes

Zacqary:
Associated Issue: #4703 

#### Fixes conditional breakpoint panel shrinking
This is achieved by adding some right padding to the breakpoint panel, and removing the minimum width of the `.CodeMirror-sizer`. I haven't noticed any side effects related to disabling the min-width.
#### Fixes conditional breakpoint panel horizontally when the user scrolls
The gutters seem to be fixed in place the same way: they listen for `scroll` events on the `.CodeMirror-scroll` div, and then move themselves `x` pixels to the left, where `x` is `.CodeMirror-scroll`'s `scrollLeft`. If we need to fix more things this way, we might be able to abstract the scroll-listening logic from `ConditionalPanel.js` in the future.

### Test Plan
- [ ] Open the conditional breakpoint panel
- [ ] Make the screen smaller, and ensure the Close Button (x) is always visible
- [ ] Make the screen small enough so that the horizontal scrollbar appears
- [ ] Scroll from left to right, and ensure the breakpoint panel stays fixed in place

#### Firefox
conditionalpanel
firefoxresize
#### Chrome
chromeresize
chromeresize

(scroll repositioning is a bit jankier in horizontal mode, though the number gutters suffer from this too)

### DIFF
--- a/src/components/Editor/ConditionalPanel.css
+++ b/src/components/Editor/ConditionalPanel.css
@@ -11,6 +11,7 @@
   background: var(--theme-toolbar-background);
   border-top: 1px solid var(--theme-splitter-color);
   border-bottom: 1px solid var(--theme-splitter-color);
+  padding-right: 16px;
 }
 
 .conditional-breakpoint-panel .prompt {
--- a/src/components/Editor/ConditionalPanel.js
+++ b/src/components/Editor/ConditionalPanel.js
@@ -30,6 +30,8 @@ type Props = {
 export class ConditionalPanel extends PureComponent<Props> {
   cbPanel: null | Object;
   input: ?HTMLInputElement;
+  panelNode: ?HTMLDivElement;
+  scrollParent: ?HTMLElement;
 
   constructor() {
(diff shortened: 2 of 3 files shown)


jasonLaster:
Hey, mind sharing GIFs w/ vertical mode and horizontal mode. 
Also good to test in firefox and chrome. And, try resizing the window and the sidebars w/ the panel open :)

Zacqary:
Found a bug: it looks like the `.CodeMirror-scroll` is deleted and recreated when the user resizes the window so that it changes from horizontal mode to vertical mode. This breaks the scroll event listener.

On `master` it **also** makes it impossible to:
- Open a conditional breakpoint widget
- Resize from horizontal to vertical, or vice versa
- Close the conditional breakpoint widget

So I might be able to fix another pre-existing bug here.

Zacqary:
Fixed the bug, added additional gifs.

Issue #4070, opened 2017-09-18
[Accessibility] Search sources widget needs to be accessible.


yzen:
Here is a list of accessibility improvements for the search sources interface (note: autocomplete accessibility should be worked on in #4071 ):

#### Search Input
- [x] Magnifying glass icon should have an ARIA role of "presentation" as it doesn't carry any meaningful semantics.
- [x] input element should have a type of "search".
- [x] input element should have keyboard focus styling.
- [ ] Results number section "X results" should have an ID set. This way we can apply an ARIA attribute [aria-describedby][by] on the input field.
- [x] Close button should be an actual button or at least have a role "button". It should have styling for focused-with-keyboard case. It should also be focusable and actionable with the keyboard.

#### Bottom Bar

- [ ] "Search sources..." and "Find in files…" are not accessible. Semantically they should be [radio][rb] buttons that are mutually exclusive. User should also be able to navigate to them using the keyboard only (tabbing to). They should also be actionable when Enter/Space are pressed.
- [ ] When "Search sources..." is selected and keyboard focus is inside the input there is a keyboard trap. There is no way to tab out of the input field (neither forward nor back).

### Screenshot

![Screen Shot 2017-09-18 at 8.08.43 PM.png](https://shipusercontent.com/42c1531e51e518e514c496914189cb4d/Screen%20Shot%202017-09-18%20at%208.08.43%20PM.png)

[by]: https://www.w3.org/TR/wai-aria/states_and_properties#aria-describedby
[rb]:https://www.w3.org/TR/wai-aria/roles#radio

yzen:
### Magnifying glass icon should have an ARIA role of "presentation" as it doesn't carry any meaningful semantics.
This is an example of how to add an appropriate role to an element. In this case role "presentation" (https://www.w3.org/TR/wai-aria/roles#presentation) to an icon.
```diff
diff --git a/assets/images/Svg.js b/assets/images/Svg.js
index 956fe45b..ab67d81e 100644
--- a/assets/images/Svg.js
+++ b/assets/images/Svg.js
@@ -58,10 +58,11 @@ type SvgType = {
   name: string,
   clasName?: string,
   onClick?: () => void,
-  "aria-label"?: string
+  "aria-label"?: string,
+  role?: string,
 };
 
-function Svg({ name, className, onClick, "aria-label": ariaLabel }) {
+function Svg({ name, className, onClick, "aria-label": ariaLabel, role }) {
   if (!svg[name]) {
[...]

PR #4815, opened 2017-11-28
Added Further Accessiblity to Search Sources Widget [#4070]

asharron:
Associated Issue: #4070

### Summary of Changes

* Added the role of presentation to the magnifying glass in Search Sources
* Added focus styling to Search Sources input element
* Made the close button next to Search Sources and Search in Files an actual button
* Can now tab to the close button next to Search in Files to close input box

### Test Plan

- [x] Control-P opens search sources and it has focus styling on the input field
- [x] Control-P opens search sources and Orca ignores magnifying glass
- [x] Control-Shift-F opens search in files and you can tab to the close button 
- [x] Pressing enter or clicking on the close button closes the search in files input box

### Comments
* We Took Feedback from the previous pull request we made and implemented the recommended changes

### DIFF
--- a/src/components/shared/Button/Close.css
+++ b/src/components/shared/Button/Close.css
@@ -36,16 +36,23 @@
   background-color: var(--theme-selection-background);
 }
 
+.close-btn:focus img.close {
+  background-color: white;
+}
+
 .close-btn.big {
   width: 16px;
   height: 16px;
 }
 
-.close-btn.big img {
-    padding: 5px;
-}
-
 .close-btn.big .close {
   width: 9px;
   height: 9px;
+}
+
+img.close::before {
+  width: 100%;
+  height: 100%;
+  padding: 5px;
+
 }
--- a/src/components/shared/SearchInput.css
+++ b/src/components/shared/SearchInput.css
@@ -66,6 +66,7 @@
 
 .search-field input:focus {
   background: var(--theme-toolbar-background-hover);
+  padding-inline-start: 10px;
   outline-width: 0;
 }
 
@@ -80,6 +81,7 @@
 .search-field .summary {
   line-height: 27px;
   padding-right: 10px;
+  padding-left: 10px;
   color: var(--theme-body-color-inactive);
 }
 


src/components/shared/SearchInput.js
```diff
@@ -79,7 +79,7 @@ class SearchInput extends Component {
       return <Svg name="sad-face" />;
     }
 
-    return <Svg name="magnifying-glass" />;
+    return <Svg name="magnifying-glass" role="presentation" />;
```
yzen: 👍

src/components/shared/Button/Close.css
```diff
@@ -24,6 +24,11 @@
   margin-top: 0;
 }
 
+.close-btn:hover {
+  display: block;
+  background: var(--theme-selection-background);
```
yzen: This is already present on line 37. You probably want to have a keyboard focus styling instead.

src/components/shared/Button/Close.js
```diff
@@ -14,13 +14,13 @@ type Props = {
 
 function CloseButton({ handleClick, buttonClass, tooltip }: Props) {
   return (
-    <div
+    <button
```
yzen: This breaks close buttons (internal 'x' svg image now seems to have width of 0). You can also see how close buttons within source tabs are also broken. I think you could use implementation of toggle-button-end and toggle-button-start buttons (in the same directory) as a good example of how to make real buttons out of "div" style buttons.

src/components/shared/Button/Close.css
```diff
@@ -32,12 +32,24 @@
   background-color: var(--theme-selection-background);
 }
 
+.close-btn:focus {
+  background-color: var(--theme-selection-background);
```
yzen: You also need the same styling like the one on line 27 but for when the button has :focus . Right now the 'x' has a gray background where it should be white.

src/components/shared/SearchInput.css
```diff
@@ -65,6 +65,7 @@
 }
 
 .search-field input:focus {
+  background: var(--theme-toolbar-background-hover);
```
yzen: Couple of suggestions:
The background is too flush with controls that are before and after the input field. Could you please add a padding-inline-start: 10px; for the input field so the cursor is not at the very edge of the input field. Also could you add 10px padding-left to the '.summary' element so there is a little bit of space between the summary and the input field (but only when summary has any results).

src/components/shared/Button/Close.css
Jessonsotoventura: Styling for line 27, is hardcoding the background to white for the image, so we have a white x on a blue background. Are you saying, that the entire button area should be white 
screenshot-2017-12-7 debugger
^ Like so when focused?

wldcordeiro:
How is this going @asharron? @yzen if you could list what is remaining I can get this ready to go.

Jessonsotoventura:
The last update should have resolved the issues listed by @yzen, last I remember we were awaiting a final approval on the changes.

asharron:
Hello, @yzen . I would like to wrap up this issue and have it be solved. Do you have any suggestions for next steps? I know that we tried to make a fix to the styling, but we are unsure as to whether or not our solution was acceptable.

yzen:
Sure, Ill go over the current state and comment where appropriate.

src/components/shared/SearchInput.css
yzen: Yeah this does not seem to work. I think it's fine to add it to not just :focus styling but to the overall .search-field input styling. I would add:
```
padding-inline-start: 10px;
padding-inline-end: 10px;
margin-inline-end: 10px;
```
What it does is it makes sure that there is always space between text and background borders: text inside input field and text for the number of results.

src/components/shared/Button/Close.js
yzen: This seems fine now.

PR #4818, opened 2017-11-28
Fixed whitespace at the bottom of center pane after searching files

CharlesStocky:
**Post fix:**
screenshot from 2017-11-28 14-06-16
screenshot from 2017-11-28 14-06-48

**Prior to fix:**
screenshot from 2017-11-28 14-19-25
screenshot from 2017-11-28 14-19-34

### DIFF
--- a/src/components/ProjectSearch/TextSearch.css
+++ b/src/components/ProjectSearch/TextSearch.css
@@ -88,7 +88,7 @@
 
 .project-text-search .managed-tree {
   overflow-y: auto;
-  height: calc(100% - 81px);
+  height: 100%;
 }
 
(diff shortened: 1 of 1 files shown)


jasonLaster:
Thanks @CharlesStocky could you test this in chrome as well and see what it looks like if you look for something that only has one match.

But great fix!

CharlesStocky:
@jsonLaster I'm working on a way of opening the debugger with chrome through launchpad (on ubuntu 16.04). I had a friend open it however, this is what it looks like in chrome at half size: 

screen_shot_2017-12-01_at_7 48 08_pm

and Firefox with one match: 

screenshot from 2017-12-02 17-07-56

Issue #4570, opened 2017-11-02
[Preview] Popup can flash if you’re off center


jasonLaster:
We might have filed this in another place, but I couldn't find it so filing again:

STR:
1. pause
2. hover on a pause token (move the mouse to the border of the token top,bottom)...

![](http://g.recordit.co/XhZSLCUM7E.gif)

```diff
diff --git a/src/actions/ast.js b/src/actions/ast.js
index 769d649..0b857f6 100644
--- a/src/actions/ast.js
+++ b/src/actions/ast.js
@@ -122,6 +122,7 @@ export function setPreview(
       return;
     }
 
+    // this is where we do the work to get the new preview values
     await dispatch({
       type: "SET_PREVIEW",
       [PROMISE]: (async function() {
diff --git a/src/components/Editor/Preview/index.js b/src/components/Editor/Preview/index.js
index 540b539..66ab981 100644
--- a/src/components/Editor/Preview/index.js
+++ b/src/components/Editor/Preview/index.js
@@ -66,6 +66,7 @@ class Preview extends PureComponent {
 
   onMouseOver(e) {
     const { target } = e;
+    console.log("is this fired ofteh");
     if (this.props.selectedFrameVisible) {
       updatePreview(target, this.props.editor, this.props);
     }
diff --git a/src/utils/editor/expression.js b/src/utils/editor/expression.js
index b36cd2c..659b52b 100644
--- a/src/utils/editor/expression.js
+++ b/src/utils/editor/expression.js
@@ -25,6 +25,7 @@ export function updatePreview(
   const cursorPos = target.getBoundingClientRect();
 
   if (preview) {
+    // we try to prevent showing unecessary popups here
     // We are mousing over the same token as before
     if (isEqual(preview.tokenPos, location)) {
       return;
@@ -44,6 +45,7 @@ export function updatePreview(
       !target.parentElement.closest(".CodeMirror-line")) ||
     cursorPos.top == 0;
 
+  // are we updating
   const isUpdating = preview && preview.updating;
 
   const inScope = linesInScope && linesInScope.includes(location.line);
@@ -53,6 +55,7 @@ export function updatePreview(
     target.className === "cm-number" ||
     target.className === "cm-atom";
 
+  // is there something else going on that is moving the mouse off the token
   if (invalidTarget || !inScope || isUpdating || invalidToken || invaildType) {
     return;
   }
```

tohmjudson:
/claim
Going to give this one a try

[...]

PR #4820, opened 2017-11-29
Prevent flashing preview

jasonLaster:
Associated Issue: #4570

### Summary of Changes

The preview popup can go crazy when the debugger is struggling w/ a big bundle or other computations like mapping scopes... This makes it much much simpler.

- remove redundant renders (caused by empty symbols obj)
- move getInScopeLocations to an action so it's only computed once
- remove parser getSymbols call, which was async and went over the wire
- moved utils/expression/updatePreview to an action
- some other stuff as well... 

We might need to break this up to make it easier to land...

#### Next Steps

- i don't think we need 2 actions, one should do...

### Test Plan

- Just manual testing
- I'd love to add more unit tests


src/components/Editor/Preview/index.js
```diff
@@ -126,21 +127,20 @@ class Preview extends PureComponent {
...
-  setPreview,
+  updatePreview,
   clearPreview
 } = actions;
 
 export default connect(
   state => ({
     preview: getPreview(state),
     selectedSource: getSelectedSource(state),
-    linesInScope: getInScopeLines(state),
```
jasonLaster: this prop is now in the action.
We can move `selectedFrameVisible` as well, which simplifies the data dependencies for the component quite a bit

src/utils/test-head.js
```diff
@@ -78,6 +78,10 @@ function makeSymbolDeclaration(name: string, line: number) {
  */
 function waitForState(store: any, predicate: any): Promise<void> {
   return new Promise(resolve => {
+    if (predicate(store.getState())) {
+      resolve();
+    }
```
jasonLaster: Not needed, but it is now consistent w/ the mochitest head helper.

jasonLaster:
created a PR for setInScopeLines

src/components/Editor/Preview/index.js
tohmjudson: It looks like you can remove `getInScopeLines` from the import (line 14) as well then

PR #4821, opened 2017-11-29
Extracts isMinified

jasonLaster:
### Summary of Changes

* extracts  isMinified
* adds a unit test
* we should add several more

Next Steps
* i don't think we want to consider small files w/ < 200 chars minified... the case i have in the test is the current, behavior but it feels off to me.
* I think we should remove the minified cache and replace it w/ a Source isMinified field... that would be simpler going forward because caches can go stale, consume memory, etc...


codehag:
flow is throwing an error. looks like this is also a refactor?

jasonLaster:
Yeah - it should be pretty easy to appease flow...

PR #4829, opened 2017-11-29
immutable variable hover

samuel-peers:
### Summary of Changes

* check if variable is an Immutable
* show "Immutable" in the header
* filter relevant roots (`tail`, `root`, and `size`)

### Screenshots
Before:
immutable_before

After:
immutable_after_map
immutable_after_list

### DIFF
--- a/src/actions/ast.js
+++ b/src/actions/ast.js
@@ -157,6 +157,7 @@ export function setPreview(
       type: "SET_PREVIEW",
       [PROMISE]: (async function() {
         let iType = null;
+        let iContents = null;
         const source = getSelectedSource(getState());
         const _symbols = await getSymbols(source.toJS());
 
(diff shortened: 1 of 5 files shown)


jasonLaster:
This is a great start!

Here's my next question, 
1) is there an easy way to show the immutable type in the header "List"
2) is there a way to format the list so it looks like an array or a map looks like an object
3) can we throw a small immutable logo in the header :P

samuel-peers:
immutable_logo_list
immutable_logo_map
immutable_logo_stack

samuel-peers:
> is there a way to format the list so it looks like an array or a map looks like an object?
I couldn't find an easy way to do this :/ I think `ObjectInspector` would have to be modified

jasonLaster:
Hmm one thought would be to load root and use that as the OI roots lol. Naming collision

samuel-peers:
immutable_entries
immutable_entries_expanded

Issue #4807, opened 2017-11-27
Add PrettyPrint Icon to Tabs Dropdown


darkwing:
It would be nice to repeat the Pretty Print icon in the tabs dropdown:

<img width="730" alt="dropdownicon" src="https://user-images.githubusercontent.com/46655/33291215-b2365868-d38a-11e7-8926-6cd184731fe3.png">

Without the `{}`, it's impossible to know which tab you're clicking on if both are open

lukaszsobek:
Feature implemented.

PR #4830, opened 2017-11-29
Adds file, prettyPrint and blackBox icons to dropdown

lukaszsobek:
Associated Issue: #4807

### Summary of Changes

* Adds file, prettyPrint and blackBox icons to dropdown

### Test Plan
dropdown-icons

### DIFF
--- a/src/components/Editor/Tabs.js
+++ b/src/components/Editor/Tabs.js
@@ -316,6 +316,12 @@ class SourceTabs extends PureComponent<Props, State> {
     });
   }
 
+  getIconClass(source: SourceRecord) {
+    if (isPretty(source.toJS())) return "prettyPrint";
+    if (source.get("isBlackBoxed")) return "blackBox";
+    return "file";
+  }
+
   renderDropdownSource(source: SourceRecord) {
     const { selectLocation } = this.props;
     const filename = getFilename(source.toJS());
@@ -324,6 +330,7 @@ class SourceTabs extends PureComponent<Props, State> {
       selectLocation({ sourceId: source.get("id"), line: 0 });
     return (
       <li key={source.get("id")} onClick={onClick}>
+        <img className={"dropdown-icon " + this.getIconClass(source)} />
         {filename}
       </li>
     );
--- a/src/components/shared/Dropdown.css
+++ b/src/components/shared/Dropdown.css
@@ -48,6 +48,30 @@ html[dir="rtl"] .dropdown {
   background-color: var(--search-overlays-semitransparent);
 }
 
+.dropdown-icon {
+  width: var(--icon-size);
+  height: var(--icon-size);
+  margin-right: 5px;
+  vertical-align: middle;
+}
+
+.dropdown-icon.prettyPrint {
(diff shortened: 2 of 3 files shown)


src/components/Editor/Tabs.js
```diff
@@ -317,13 +317,24 @@ class SourceTabs extends PureComponent<Props, State> {
...
   renderDropdownSource(source: SourceRecord) {
+    const sourceObj = source.toJS();
+    const sourceId = source.get("id");
+    let fileIconClass = "file";
+
+    if (isPretty(sourceObj)) {
+      fileIconClass = "prettyPrint";
+    } else if (source.get("isBlackBoxed")) {
+      fileIconClass = "blackBox";
+    }
```
jasonLaster: lets move this to its' own function

lukaszsobek: Done.

codehag:
looks like a rebase might have gone wrong here!

src/components/shared/Dropdown.css
```diff
@@ -48,6 +48,30 @@ html[dir="rtl"] .dropdown {
   background-color: var(--search-overlays-semitransparent);
 }
 
+.dropdown li img.prettyPrint {
+  -webkit-mask: url(/images/prettyPrint.svg) no-repeat;
```
codehag: we have autoprefixer, so you should be able to use just mask :)

src/components/shared/Dropdown.css
```diff
@@ -48,6 +48,30 @@ html[dir="rtl"] .dropdown {
...
+  mask: url(/images/prettyPrint.svg) no-repeat;
+  -webkit-mask-size: 100%;
+  mask-size: 100%;
+  background: var(--theme-highlight-blue);
+  width: 13px;
```
codehag: this looks like a magic number, maybe we can make it a variable and name it?

src/components/shared/Dropdown.css
```diff
@@ -48,6 +48,30 @@ html[dir="rtl"] .dropdown {
...
+  height: 13px;
+  margin-right: 5px;
+  vertical-align: middle;
+}
+
+.dropdown li img.blackBox {
```
codehag: the styles here are largely the same, i think we can factor the similarity out into its own class, `.dropdown--icon` with two modifications `.dropdown--icon_blackbox` and `.dropdown--icon_prettyprint`

src/components/shared/Dropdown.css
lukaszsobek: Changed.

src/components/shared/Dropdown.css
lukaszsobek: Exported to variables.css as `--icon-size`

src/components/shared/Dropdown.css
lukaszsobek: using combined `.dropdown-icon` that holds the majority of the css + `.blackBox`, `.prettyPrint` & `.file`. It seems that you cannot move the mask size declaration to a different class than declaring the actual mask, interesting.

Issue #4573, opened 2017-11-03
[PrettyPrint] and New Source Scrolling


darkwing:
I think I've nailed this down but if you open a source, then pretty print it, switch to the pretty print tab and scroll down to line 300ish, then pretty print another source, the new pretty printed source will be shown at line 300ish.

Newly created tabs should always start showing line 1.

lukaszsobek:
Related to: #4393

Issue #4393, opened 2017-10-17
[Editor] Navigating to a new source should scroll the editor to the top


darkwing:
STR:

1.  Open a source in the editor and scroll down.
2.  Open another source

In my instance, I scrolled to line 70, and when I opened another source, the editor scroll was at line 70.

When opening a new source, we should scroll the editor to the top.

jasonLaster:
I'm guessing this is a recent regression that was triggered by our Editor lifecyle PR:

I thought this [test][t] exercised that.

[t]: https://github.com/devtools-html/debugger.html/blob/master/src/test/mochitest/browser_dbg-editor-highlight.js#L6

lukaszsobek:
This seems to happen because the Editor uses the same editor tab to display the new content without resetting the line and column.

jasonLaster:
Hmm, weird it should create a new document for the new source

lukaszsobek:
It is a bug where the scroll state of the current window is preserved when opening a new one. I have the part with navigating a new window to the top done, but would love to handle the 4573 issue at the same time. However in order for that to happen we'd need #4832 to be taken care of.

jasonLaster:
Yep. Although, this is a different patch so it could easily be two PRs.

jasonLaster:
We like small PRs :)

lukaszsobek:
Closing. Issue resolved apart from edge cases, which are already mentioned in another issue.

Issue #4832, opened 2017-11-30
[PrettyPrint] Pretty document loaded first time has wrong properties


lukaszsobek:
STR:
- open debugger on minimized file `https://apis.google.com/js/api.js`
- press `ctrl+shift+i`
- open `src/components/Editor/index.js`
- place breakpoint  ~line 237 inside `componentDidUpdate` after the line `const { selectedLocation, selectedSource } = this.props;`
- prettify js
- look at Scope > Local > selectedSource > loadedState
- look at Scope > Local > selectedSource > isPrettyPrinted

ER:
first load:
- selectedSource.get("loadedState") == "loading"
- selectedSource.get("isPrettyPrinted") == "true"
All loads after:
- selectedSource.get("loadedState") == "loaded"
- selectedSource.get("isPrettyPrinted") == "true"

AR:
first load:
- selectedSource.get("loadedState") == "loaded"
- selectedSource.get("isPrettyPrinted") == "false"
All loads after:
- selectedSource.get("loadedState") == "loaded"
- selectedSource.get("isPrettyPrinted") == "true"

Image:
![pretty_print_bug_overlay](https://user-images.githubusercontent.com/23530054/33458001-cadcf4a2-d624-11e7-9020-5cd12efc522a.png)

jasonLaster:
Nice find. Can you share a test case and the steps to follow...

lukaszsobek:
updated desc to make it clearer.

PR #4836, opened 2017-12-01
Makes windows with new sources open at line 1 column 1

lukaszsobek:
Associated Issue: #4393

### Summary of Changes

* Rationale: Due to new tabs opened inheriting scroll state from previously focused tabs we have to manually scroll the container up if the source is new. 
* Makes container scroll to top if `loadedState != loaded` i.e. the document has not been loaded before.
* Disables flow checking on property .scrollTo, since flow does not understand it

### Test Plan

- open long tab
- scroll down to a line outside of the original view
- open new tab
- new tab is scrolled to the top
(in the old version it would be at the same line as the tab before)

### DIFF
--- a/src/actions/sources.js
+++ b/src/actions/sources.js
@@ -247,6 +247,11 @@ export function selectLocation(location: Location, tabIndex: string = "") {
 
     dispatch(addTab(source.toJS(), 0));
 
+    if (source.get("loadedState") == "unloaded") {
+      if (!location.line) location.line = 1;
+      if (!location.column) location.column = 1;
+    }
+
     return dispatch({
       type: "SELECT_SOURCE",
       source: source.toJS(),
--- a/src/components/Editor/Preview/index.js
+++ b/src/components/Editor/Preview/index.js
@@ -11,6 +11,7 @@ import Popup from "./Popup";
 import {
   getPreview,
   getSelectedSource,
+  getInScopeLines,
   isSelectedFrameVisible
 } from "../../../selectors";
 import actions from "../../../actions";
@@ -134,6 +135,7 @@ export default connect(
   state => ({
     preview: getPreview(state),
     selectedSource: getSelectedSource(state),
+    linesInScope: getInScopeLines(state),
     selectedFrameVisible: isSelectedFrameVisible(state)
   }),
   {
--- a/src/components/Editor/index.js
+++ b/src/components/Editor/index.js
@@ -442,6 +442,7 @@ class Editor extends PureComponent<Props, State> {
     // to flash the previously saved selected line.
     if (
       line !== null &&
+      this.lastJumpLine !== 1 &&
       this.lastJumpLine !== null &&
       (!selectedFrame || selectedFrame.location.line !== line)
     ) {
--- a/src/actions/tests/ast.spec.js
+++ b/src/actions/tests/ast.spec.js
@@ -180,7 +180,9 @@ describe("ast", () => {
         const locations = getOutOfScopeLocations(getState());
         const lines = getInScopeLines(getState());
 
-        expect(locations).toEqual(null);
+        expect(locations).toEqual([
+          { end: { column: 21, line: 1 }, start: { column: 13, line: 1 } }
+        ]);
         expect(lines).toEqual([1]);
       });
     });
@@ -222,8 +224,11 @@ describe("ast", () => {
 
         const locations = getOutOfScopeLocations(getState());
         const lines = getInScopeLines(getState());
-        expect(locations).toEqual(null);
-        expect(lines).toEqual([1, 2, 3]);
+
+        expect(locations).toEqual([
+          { end: { column: 1, line: 2 }, start: { column: 13, line: 1 } }
+        ]);
+        expect(lines).toEqual([2, 3]);
       });
     });
   });


lukaszsobek:
This should also automatically fix #4573 after #4832 is taken care of

src/components/Editor/index.js
```diff
@@ -234,6 +234,13 @@ class Editor extends PureComponent<Props, State> {
     // responsible for updating the editor annotations.
     const { selectedLocation, selectedSource } = this.props;
 
+    // To stop newly opened sources from inheriting the scroll-state
+    // of the currently open tab we force a scrollTo
+    if (selectedSource && selectedSource.get("loadedState") != "loaded") {
+      const newSource = document.querySelector(".CodeMirror-scroll");
+      /*::`*/ if (newSource) newSource.scrollTo(0, 0); /*::`*/
```
codehag: our linter is complaining about these inline comments, im also wondering what do you want to express with them? maybe there is a better way?

im also wondering if there are any edge cases for this. We don't always want to open a source and scroll to 0,0. sometimes we want to be at a line. maybe what we want to do, when we select location, is default to 0 if no other location is specified?

lukaszsobek: The linter is complaining because it didn't expect the property .toTop on an object that might not be there. Looking for a solution to this I've moved the code to a better place and rewritten it,  working on disabling the scroll if there is a location set now.

codehag: hmm this is the error on the ci (lint usually doesnt complain about properties on objects, do you mean flow?):

```
yarn lint-js

yarn lint-js v0.28.4
$ eslint *.js "src/**/*.js" --fix
The babel/generator-star-spacing rule is deprecated. Please use the built in generator-star-spacing rule instead.

/home/ubuntu/debugger.html/src/components/Editor/index.js
  241:7  error  Unexpected comment inline with code  no-inline-comments
  243:9  error  Unexpected comment inline with code  no-inline-comments

✖ 2 problems (2 errors, 0 warnings)

yarn lint-js returned exit code 1
```

codehag: an str for opening a new file at a location is to have breakpoints in a few files, close all of them, and click the breakpoint

lukaszsobek: Sounds like a great str, thanks!

src/components/Editor/Preview/index.js
```diff
@@ -47,10 +47,30 @@ class Preview extends PureComponent {
...
+    )
+      return;
+
+    try {
+      this.props.preview.result.location.line;
+    } catch (e) {
+      scrollContainer.scrollTo(0, 0);
+    }
+  }
+
```
bomsy: @lukaszsobek This try catch feels a bit hacky, can't we just default the `selectedLocation.line` to 0 if there's not a line already set? that should just work.

lukaszsobek: That was a train of thought I had considered and also the first working version used, but setting `line` means we want to scroll to that specific line and ping it. In this case we want to get around the new tab inheriting the scroll state. While using `line: 0` should do the job it did not feel like the "right" tool to use.

The try catch is only there because the object we are testing against is nested and I wanted to avoid having to call eval or `if(a && a.b && a.b.c && a.b.c.d)`

The approach used here is actually an idea taken from https://github.com/ReactTraining/react-router/blob/master/packages/react-router-dom/docs/guides/scroll-restoration.md

lukaszsobek: Changed it - less code, easier to read, fewer dependencies.

bomsy:
@lukaszsobek looks nice ...

src/components/Editor/Preview/index.js
```diff
@@ -47,10 +47,24 @@ class Preview extends PureComponent {
...
   }
 
+  // To stop newly opened sources from inheriting the scroll-state
+  // of the currently open tab we force a scroll to top.
+  topScrollNewTab(dispSource) {
+    if (dispSource.get("loadedState") == "loaded") return;
+
+    const sourceLocation = dispSource.location;
+    if (!sourceLocation || !sourceLocation.line) {
+      sourceLocation = { line: 0, column: 0 };
```
jasonLaster: what does this do?

src/components/Editor/Preview/index.js
```diff
@@ -47,10 +47,24 @@ class Preview extends PureComponent {
...
+    }
+  }
+
   componentDidMount() {
     const { codeMirror } = this.props.editor;
+    const { selectedSource } = this.props;
     const codeMirrorWrapper = codeMirror.getWrapperElement();
 
+    this.topScrollNewTab(selectedSource);
```
jasonLaster: why is it in Preview/index and not Editor/index?

I would have expected this kind of logic in an action

src/components/Editor/Preview/index.js
lukaszsobek: It checks if the new document was called with a location. We only want to scroll to the top if there is no line set in the first place, otherwise it would overwrite an intended scroll.

src/components/Editor/Preview/index.js
lukaszsobek: It is on the preview because the componentDidMount action occurs once for every tab mounted. In editor it would have to be in componentDidUpdate which seemed the less efficient place to put it in due to all the checks involved.

Sure, will put it in an action.

codehag:
@lukaszsobek should we close this? it is probably fixed by selectLocation

lukaszsobek:
@codehag - unfortunately it doesn't seem so. Just checked with the freshest version.

src/components/Editor/Preview/index.js
```diff
@@ -49,6 +49,7 @@ class Preview extends PureComponent {
 
   componentDidMount() {
     const { codeMirror } = this.props.editor;
+    const { selectedSource } = this.props;
```
lukaszsobek: Leftover, will remove in next iteration.

src/components/Editor/index.js
```diff
@@ -442,6 +442,7 @@ class Editor extends PureComponent<Props, State> {
     // to flash the previously saved selected line.
     if (
       line !== null &&
+      this.lastJumpLine !== 0 &&
```
lukaszsobek: if last jumpline was 0 it means we are jumping to the start of the document. and we don't want highlight.

src/actions/sources.js
```diff
@@ -220,6 +220,11 @@ export function selectLocation(location: LocationObject) {
 
     dispatch(addTab(source.toJS(), 0));
 
+    if (source.get("loadedState") == "unloaded") {
+      if (!location.line) location.line = 1;
```
lukaszsobek: I'd really love to use line 0 & column 0, but we get this:
screenshot-20171206102351-914x77

src/actions/sources.js
```diff
@@ -220,6 +220,11 @@ export function selectLocation(location: LocationObject) {
...
+      if (!location.column) location.column = 1;
```
jasonLaster: maybe we just do `location = {...location, line: 1, column: 1}`

lukaszsobek: Checked, it bears different results as we are unconditionally rewriting line and column. For example, when someone clicks a link to follow a source that is not yet loaded the above would overwrite the line and column the person wanted to go to.

src/components/Editor/Preview/index.js
```diff
@@ -70,7 +70,7 @@ class Preview extends PureComponent {
   onMouseOver(e) {
     const { target } = e;
     if (this.props.selectedFrameVisible) {
-      this.props.updatePreview(target, this.props.editor);
+      updatePreview(target, this.props.editor, this.props);
```
jasonLaster: looks like this needs a rebase :)

src/actions/sources.js
```diff
@@ -247,6 +247,11 @@ export function selectLocation(location: Location, tabIndex: string = "") {
...
```
jasonLaster: So lets see here:

1. user wants to go to line 23 / col 47, we shouldn't change the location
2. user doesn't specify a line or column, we default to 1/1

- why is this different for loaded / unloaded sources?

lukaszsobek: A loaded source might already have a state in itself, e.g. the user scrolled the window to line 254 manually and then switched tabs or closed tabs.

src/components/Editor/Preview/index.js
lukaszsobek: This is after a rebase with the newest upstream at the time of pushing :/

lukaszsobek: fixed it manually.

jasonLaster:
@lukaszsobek i just landed my refactor for highlight line, debug line, and scrolling.

I'll try and do a rebase and see what we can do

src/components/Editor/index.js
```diff
@@ -442,6 +442,7 @@ class Editor extends PureComponent<Props, State> {
     // to flash the previously saved selected line.
     if (
       line !== null &&
+      this.lastJumpLine !== 1 &&
```
jasonLaster: the new work drops the notion of `lastJumpLine` and perhaps fixes the problem :)

not sure

lukaszsobek: @jasonLaster  First impression: it's a lot faster than the old one, yummy! Getting to work.

jasonLaster: awesome!

lukaszsobek: @jasonLaster We can completely remove this pr, the issue is no more. There is however another bug introduced. If you scroll a document, reload the debugger, the document will reload on line 1 column 1 instead of on the scroll state it was before.

jasonLaster: hmm - i wonder how that is new... we don't really store the scroll state and we'll probably remove the source while it is loading

lukaszsobek:
Issue seems to have been fixed. Closing.

src/components/Editor/index.js
lukaszsobek: Also another bug, if you open a document after a prettified source has been created and scrolled, the new document opens scrolled.
pretty-scroll

Issue #4844, opened 2017-12-01
revisit selectLocation action


codehag:
selectLocation is not yet consistent. we are selecting locations in two ways, and the id has been doubled

<img width="560" alt="screen shot 2017-12-01 at 16 56 15" src="https://user-images.githubusercontent.com/26968615/33490809-9241b062-d6b8-11e7-99ed-431cbadf53a5.png">

<img width="553" alt="screen shot 2017-12-01 at 16 54 40" src="https://user-images.githubusercontent.com/26968615/33490757-71ffb722-d6b8-11e7-8fb9-1bd39631a804.png">

Also, we have a failing mochitest

https://travis-ci.org/devtools-html/debugger.html/builds/309222663?utm_source=github_status&utm_medium=notification

PR #4845, opened 2017-12-01
fix selectLocation action

codehag:
Associated Issue: #4844 

This fixes the inconsistency and the test. however it also seems to have fixed pretty print, which was not working and not noticed until now. the CI tests are failing, they need to be updated to reflect that we pretty print automatically now. will revisit in the morning!

### DIFF
--- a/src/actions/sources.js
+++ b/src/actions/sources.js
@@ -42,7 +42,7 @@ import {
   getActiveSearch
 } from "../selectors";
 
-import type { Source } from "../types";
+import type { Source, Location } from "../types";
 import type { ThunkArgs } from "./types";
 import type { State } from "../reducers/types";
 
@@ -52,7 +52,7 @@ async function checkSelectedSource(state: State, dispatch, source) {
   const pendingLocation = getPendingSelectedLocation(state);
 
   if (pendingLocation && !!source.url && pendingLocation.url === source.url) {
-    await dispatch(selectLocation({ id: source.id, ...pendingLocation }));
+    await dispatch(selectLocation({ ...pendingLocation, sourceId: source.id }));
   }
 }
 
@@ -155,7 +155,7 @@ function loadSourceMap(generatedSource) {
 
 declare type SelectSourceOptions = {
   tabIndex?: number,
-  location?: { line: ?number, column?: ?number }
+  location?: { line: number, column?: ?number }
 };
 
 /**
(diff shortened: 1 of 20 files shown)


src/actions/sources.js
```diff
@@ -322,7 +321,7 @@ export function closeTabs(urls: string[]) {
     const sourceId = getNewSelectedSourceId(getState(), tabs);
 
     dispatch({ type: "CLOSE_TABS", urls, tabs });
-    dispatch(selectLocation({ id: sourceId }));
+    dispatch(selectLocation({ sourceId, line: 0 }));
```
codehag: this is a left over

src/components/Editor/Tabs.js
```diff
@@ -320,7 +320,8 @@ class SourceTabs extends PureComponent<Props, State> {
     const { selectLocation } = this.props;
     const filename = getFilename(source.toJS());
 
-    const onClick = () => selectLocation({ id: source.get("id") });
+    const onClick = () =>
+      selectLocation({ sourceId: source.get("id"), line: 0 });
```
codehag: leftover

jasonLaster:
Thanks @codehag!

Lets see if we can tighten up our flow types to catch this. Maybe Location should be a strict type `{| |}`.

src/components/QuickOpenModal.js
```diff
@@ -140,17 +140,16 @@ export class QuickOpenModal extends Component<Props, State> {
       if (selectedSource == null) {
         return;
       }
-      selectLocation({
-        id: selectedSource.get("id"),
-        ...(item.location != null ? { line: item.location.start.line } : null)
-      });
+      const line =
```
lukaszsobek: The 0 part would be covered at component mount time by https://github.com/devtools-html/debugger.html/pull/4836 , I'd rather null this here so as not to have this functionality hardcoded in too many places.

src/components/QuickOpenModal.js
```diff
@@ -140,17 +140,16 @@ export class QuickOpenModal extends Component<Props, State> {
...
     } else if (searchType === "gotoSource") {
       const location = parseLineColumn(query);
       if (location != null) {
-        selectLocation({ id: item.id, ...location });
+        selectLocation({ ...location, sourceId: item.id });
       }
     } else {
-      selectLocation({ id: item.id });
+      selectLocation({ sourceId: item.id, line: 0 });
     }
```
lukaszsobek: Same as above.

src/components/QuickOpenModal.js
```diff
@@ -168,10 +167,9 @@ export class QuickOpenModal extends Component<Props, State> {
...
 
     if (searchType === "variables") {
-      selectLocation({
-        id: selectedSource.get("id"),
-        ...(item.location != null ? { line: item.location.start.line } : null)
-      });
+      const line =
+        item.location && item.location.start ? item.location.start.line : 0;
```
lukaszsobek: here too

lukaszsobek:
@codehag @jasonLaster 
- Tests update pending via pr on codehag's repo
- We are using` Location` as type in several places, it requires i.a. line explicitly. selectLocation (which is currently using `Location`) on the other hand is perfectly fine without having a location set. The problem is that is passes the location on directly, so dependent functions will not get their line if it wasn't already set. Some solutions might be:
1.) give `selectLocation` its own type that does not require line nor column, but inside the function make sure line and column are set to 0 if not provided.
2.) make every `selectLocation` call have line explicitly set to 0 if there is nothing else

Issue #4837, opened 2017-12-01
[Performance] Replace subsequent .toJS() calls with reference to stored value


lukaszsobek:
.toJS() seems to be an expensive function. Wherever possible it would be nice to call it once, store the value and use the value instead of calling the function over and over again. 

At the moment of writing there are 94 instances of .toJS() in the code. This issue would require looking at the functions where .toJS() is called and, if it is called two or more times within the same function to assign the result of .toJS() to a variable and reference that instead of executing the function several times.

Todo List
- [ ] Actions
- [ ] Components
  - [ ]  Source Tree #5738

cristidrg:
I can work on this

bomsy:
@cristidrg go for it! You can claim it if you like

codehag:
this might need to be broken up, since this is a pretty large change, and will require a lot of things to be changed. @cristidrg let us know if you need help keeping it managable!

cristidrg:
Sure, I'll share my insights after I'll work on it for a bit.

jasonLaster:
This will be 10X easier now that sources uses a record.

We should be able replace ALL of the toJS calls now because `source.text` works as well as `source.get("text")`

jasonLaster:
So... I ran into a problem where i accidentally tried to set a value in a source record. Now that the objects _look_ like objects we have to be careful :)

jasonLaster:
I think we've basically finished this. @lukaszsobek thoughts?

lukaszsobek:
@jasonLaster I'd agree, and we got rid of a lot of items.
Just doing a final sanity test for the remaining items to see if there is anything more to squeeze out of it.

jasonLaster:
sounds good

lukaszsobek:
We're down to about 40 or so. The remaining toJS are either of no relevance to the performance of the working application or needed. Closing this.

PR #4847, opened 2017-12-01
Replace subsequent calls of toJS() #4837

cristidrg:
Associated Issue: #4837 

### Summary of Changes
I memoized values of subsequent toJS() calls to improve overall performance.

### Test Plan
I reran all the tests which were affected by the changes made.

### DIFF
--- a/src/actions/ast.js
+++ b/src/actions/ast.js
@@ -157,7 +157,8 @@ export function setPreview(
       type: "SET_PREVIEW",
       [PROMISE]: (async function() {
         const source = getSelectedSource(getState());
-        const _symbols = await getSymbols(source.toJS());
+        const sourceRecord = source.toJS();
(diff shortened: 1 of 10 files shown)


cristidrg:
I'm not sure why the travis build fails...

jasonLaster:
Thanks!

A lot of the toJS called are for the source object. It would be cool to explore immutable records which would let us avoid all toJS calls for that object :)

https://tonyhb.gitbooks.io/redux-without-profanity/content/using_immutablejs_records.html

jasonLaster:
Hi @cristidrg, we recently landed #4890 which had a lot of the work for sources.

I'll close this now, because that was the majority of the PR, but feel free to re-open this after a rebase or with a new commit that touches non-source toJS's

Issue #4833, opened 2017-11-30
Massive hang after clicking pretty print a few times


digitarald:
STR:

- Click pretty print a few times application.js on Slack

AR:
- Even after the pretty printed script shows up, the whole UI beachballs

Profile with 20s main process hang: https://perfht.ml/2BnIjS2

lukaszsobek:
Just to make sure, what does to beachball mean?

codehag:
beachball occurs on macs, and it is an indicator that the system is waiting for something

digitarald:
Also called Eternal Flame: https://xkcd.com/961/

AnshulMalik:
This can be closed ?

PR #4850, opened 2017-12-02
Call prettyprint only once

AnshulMalik:
Associated Issue: #4833

### Summary of Changes

* Prevent multiple execution of pretty print heavy task when button is pressed again.

### Test Plan

- Need to represent this in a test case, need help with this

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "lint-css": "stylelint \"src/components/**/*.css\"",
     "lint-js": "eslint *.js \"src/**/*.js\" --fix",
     "lint-md":
-      "remark -u devtools-linters/markdown/preset -qf *.md configs docs",
+      "remark -u devtools-linters/markdown/preset -qf *.md src configs docs",
     "lint-fix": "yarn lint-js -- --fix",
     "mochi":
       "mochii --mc ./firefox --interactive --default-test-path devtools/client/debugger/new",
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "lint-css": "stylelint \"src/components/**/*.css\"",
     "lint-js": "eslint *.js \"src/**/*.js\" --fix",
     "lint-md":
-      "remark -u devtools-linters/markdown/preset -qf *.md src configs docs",
+      "remark -u devtools-linters/markdown/preset -qf *.md configs docs",
     "lint-fix": "yarn lint-js -- --fix",
     "mochi":
       "mochii --mc ./firefox --interactive --default-test-path devtools/client/debugger/new",
--- a/src/actions/sources/createPrettySource.js
+++ b/src/actions/sources/createPrettySource.js
@@ -2,9 +2,8 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
 
-import { getFrames, getSource } from "../../selectors";
+import { getSource } from "../../selectors";
 import { prettyPrint } from "../../workers/pretty-print";
-import { updateFrameLocations } from "../../utils/pause";
 import { getPrettySourceURL } from "../../utils/source";
 
 export function createPrettySource(sourceId) {
@@ -17,46 +16,17 @@ export function createPrettySource(sourceId) {
       url,
       id,
       isPrettyPrinted: true,
-      text: "Loading...",
       contentType: "text/javascript",
-      frames: [],
-      loadedState: "loadeding"
+      loadedState: "loading"
     };
     dispatch({ type: "ADD_SOURCE", source: prettySource });
 
-    dispatch({
-      type: "UPDATE_SOURCE",
-      source: {
-        id: sourceId,
-        loadedState: "loading"
-      }
-    });
-
-    const { code, mappings } = await prettyPrint({
-      source,
-      url
-    });
-
+    const { code, mappings } = await prettyPrint({ source, url });
     await sourceMaps.applySourceMap(source.get("id"), url, code, mappings);
 
-    let frames = getFrames(getState());
-    if (frames) {
-      frames = await updateFrameLocations(frames, sourceMaps);
-    }
-
-    prettySource.text = code;
-    prettySource.frames = frames;
-    prettySource.loadedState = "loaded";
-
     dispatch({
-      type: "UPDATE_SOURCES",
-      sources: [
-        {
-          id: sourceId,
-          loadedState: "loaded"
-        },
-        prettySource
-      ]
+      type: "UPDATE_SOURCE",
+      source: { ...prettySource, text: code, loadedState: "loaded" }
     });
 
     return prettySource;
--- a/src/actions/sources/createPrettySource.js
+++ b/src/actions/sources/createPrettySource.js
@@ -13,6 +13,25 @@ export function createPrettySource(sourceId) {
     const url = getPrettySourceURL(source.get("url"));
     const id = await sourceMaps.generatedToOriginalId(sourceId, url);
 
+    const prettySource = {
+      url,
+      id,
+      isPrettyPrinted: true,
+      text: "Loading...",
+      contentType: "text/javascript",
+      frames: [],
+      loadedState: "loadeding"
+    };
+    dispatch({ type: "ADD_SOURCE", source: prettySource });
+
+    dispatch({
+      type: "UPDATE_SOURCE",
+      source: {
+        id: sourceId,
+        loadedState: "loading"
+      }
+    });
+
     const { code, mappings } = await prettyPrint({
       source,
       url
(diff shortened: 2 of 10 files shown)


src/actions/sources.js
```diff
@@ -366,6 +366,7 @@ export function togglePrettyPrint(sourceId: string) {
     }
 
     const newPrettySource = await dispatch(createPrettySource(sourceId));
+    console.log(newPrettySource);
```
bomsy: rouge console.log

AnshulMalik: Yep, doing cleanup now

jasonLaster:
Thanks, I like this work :)

So a couple of thoughts:

There are several use-cases with async actions: 

- we do one action and cancel subsequent actions
- we do one action and subsequent calls wait on the initial call
- we start an action and show a loading state

If you want to wait on subsequent calls you need this pattern with the requests map because the map stores the promises. 

If you just want to cancel subsequent calls, you can keep track of a pending state in the store. The advantage of adding the pending state to the store is that we can use that in the UI:

- disable/hide the pretty print button
- show a progress ui

In this case, I think we should probably follow the source loadState pattern with "unloaded", "loading", and "loaded". This means we'd add a new field to source or update the prettyPrinted field to be "unprettyprinted", "pretty-printing", or "pretty-printed" or something better than that :)

bomsy:
I like the idea of storing the pending state, that info would really useful in structuring a workflow for the user. 
For e.g if the user has a visual que that prettyprinting has started, they would not need to hit the button again and again.

Also disabling / hiding the pretty print button also saves from kicking off other uneccesary actions.

src/utils/source.js
```diff
@@ -340,7 +340,19 @@ function isLoading(source: Source) {
...
+  const promise: Promise<Source> = new Promise((_res, _rej) => {
+    resolve = _res;
+    reject = _rej;
+  });
+
+  return { resolve, reject, promise };
+}
+
 export {
+  defer,
```
jasonLaster: lol - yeah, this should be in a new file. `utils/defer`

codehag:
can you rebase this?

AnshulMalik:
Rebase ? squash ? @codehag

codehag:
@AnshulMalik the tests are failing on travis at the moment, because there was an issue with master :) if you rebase the travis tests should pass and we can trust the tests a bit more

AnshulMalik:
Looks like mochitest fails in travis after rebase

jasonLaster:
I'll look at this on monday :)

jasonLaster:
Hey @AnshulMalik not sure why travis didn't run, but circle has some easy fixes...

jasonLaster:
Hey @AnshulMalik, I'm now seeing something more reasonable:      3259 INFO TEST-UNEXPECTED-FAIL | devtools/client/debugger/new/test/mochitest/browser_dbg-pretty-print.js | A promise chain failed to handle a rejection: No such actor for ID: server1.conn29.child1/source27/originalSource-d096ee6b9228104ac049e96c5926c143 - stack: promiseMiddleware/</</</</< devtools/client/debugger/new/debugger.js 7355:11

src/components/Editor/Footer.js
```diff
@@ -48,15 +48,18 @@ class SourceFooter extends PureComponent<Props> {
...
+    const isPrettyPrinting =
+      selectedSource.get("prettyStatus") == "prettifying";
 
+    const type = "prettyPrint";
     return (
       <button
         onClick={() => togglePrettyPrint(selectedSource.get("id"))}
         className={classnames("action", type, {
-          active: sourceLoaded,
+          active: sourceLoaded || isPrettyPrinting,
```
jasonLaster: why do we need `|| isPrettyPrinting`

AnshulMalik: I thought, when active class is added, button's ui changes somehow to represent that it's active, so that user can know that the action is in progress.
But atm, adding active doesn't do anything.

jasonLaster: Got it

jasonLaster:
Hey, just thinking about it a bit more after seeing your patch and I think we can do something clever:

1. I think we can use `loadedState` instead of adding `prettyStatus`. When we create the new pretty source we can set the `loadedState` to loading and when it is done, set it to `loaded`. The advantage is that we can piggy back on the existing loading functionality.
2. It would be nice to add a new UPDATE_SOURCE action so we don't abuse ADD_SOURCE to do the update

AnshulMalik:
I tried using `loadedState` in place of `prettyStatus`, but when I prettify a source, it's text get's replaced with pretty text and also the prettified source tab opens.

jasonLaster:
oh no. we want a new source object for the pretty source

AnshulMalik:
a

src/actions/sources/createPrettySource.js
```diff
@@ -13,6 +13,14 @@ export function createPrettySource(sourceId) {
     const url = getPrettySourceURL(source.get("url"));
     const id = await sourceMaps.generatedToOriginalId(sourceId, url);
 
+    dispatch({
+      type: "UPDATE_SOURCE",
+      source: {
+        id: sourceId,
+        loadedState: "loading"
+      }
+    });
```
jasonLaster: I was imagining that we would add the pretty source here w/ the loadedState `loading` and below, we would update the pretty source w/ the text and loadedState `loaded`

lukaszsobek: Yes please .. that should take care of #4832

package.json
```diff
@@ -27,7 +27,7 @@
     "lint-css": "stylelint \"src/components/**/*.css\"",
     "lint-js": "eslint *.js \"src/**/*.js\" --fix",
     "lint-md":
-      "remark -u devtools-linters/markdown/preset -qf *.md src configs docs",
+      "remark -u devtools-linters/markdown/preset -qf *.md configs docs",
```
AnshulMalik: I am not confident about this change, it's just a workaround

flow-typed/debugger-html.js
```diff
@@ -2,35 +2,35 @@
 
 declare module "debugger-html" {
   /**
- * Breakpoint ID
- *
- * @memberof types
- * @static
- */
```
jasonLaster: i believe this is fixed if you rebase

src/actions/sources/createPrettySource.js
```diff
@@ -13,6 +13,28 @@ export function createPrettySource(sourceId) {
...
+      contentType: "text/javascript",
+      frames: [],
+      loadedState: "loadeding"
+    };
+    dispatch({
+      type: "ADD_SOURCE",
+      source: prettySource
+    });
+
+    dispatch({
```
jasonLaster: why do we need to update the non-pretty source?

src/actions/sources/createPrettySource.js
```diff
@@ -13,6 +13,28 @@ export function createPrettySource(sourceId) {
...
+      isPrettyPrinted: true,
+      text: "Loading...",
+      contentType: "text/javascript",
+      frames: [],
+      loadedState: "loadeding"
+    };
+    dispatch({
+      type: "ADD_SOURCE",
+      source: prettySource
+    });
```
jasonLaster: this can be one line 

`  dispatch({    type: "ADD_SOURCE",      source: prettySource }); `

src/actions/sources/createPrettySource.js
```diff
@@ -25,19 +47,19 @@ export function createPrettySource(sourceId) {
...
-      source: prettySource
+      type: "UPDATE_SOURCES",
+      sources: [
+        {
+          id: sourceId,
+          loadedState: "loaded"
+        },
+        prettySource
+      ]
     });
```
jasonLaster: we should be updating the pretty source here.

src/actions/sources/createPrettySource.js
AnshulMalik: This is to indicate on UI that this source is being pretty printed

src/actions/sources/createPrettySource.js
AnshulMalik: It's update above https://github.com/devtools-html/debugger.html/pull/4850/files#diff-69b4b3dbfca379d059dd7360f8c53a52R50

src/actions/sources/createPrettySource.js
```diff
@@ -13,6 +13,25 @@ export function createPrettySource(sourceId) {
...
     const id = await sourceMaps.generatedToOriginalId(sourceId, url);
 
+    const prettySource = {
+      url,
+      id,
+      isPrettyPrinted: true,
+      text: "Loading...",
+      contentType: "text/javascript",
+      frames: [],
+      loadedState: "loadeding"
```
jasonLaster: `loading` lol

jasonLaster:
quick rebase and squash

jasonLaster:
@AnshulMalik I added some refactoring here:

1. moved toggle pretty print to its own module
2. pulled out the updateFrame stuff because that made it much harder.

Not sure what needs to be done here

jasonLaster:
- polished the experience so that auto-pretty print closes the ugly file. 
- we only auto-pretty print the first time, before we've created a pretty file.

jasonLaster:
@AnshulMalik i've run with this a bit to see what it would look like to do some pause refactoring. 

- replace pause state with isPaused and getPauseReason calls
- ...

jasonLaster:
I pulled the refactor stuff out so this could land :)

lukaszsobek:
@AnshulMalik @jasonLaster try this:
1.) open slack.com in debugger,
2.) Find rollup-marketing.8908908908.min.js,
3.) click file in sidebar 5-15 times 

Results may include:
- several pretty print windows
- browser hanging
- heavy beachball

Issue #4532, opened 2017-10-30
Update the mapDispatchToProps calls for our components


wldcordeiro:
Currently most of our `connect()` calls look like this.

```js
function mapStateToProps(state) {
  const selectedSource = getSelectedSource(state);
  let symbols = null;
  if (selectedSource != null) {
    symbols = getSymbols(state, selectedSource.toJS());
  }
  return {
    enabled: getQuickOpenEnabled(state),
    sources: formatSources(getSources(state)),
    selectedSource,
    symbols: formatSymbols(symbols),
    query: getQuickOpenQuery(state),
    searchType: getQuickOpenType(state)
  };
}

export default connect(mapStateToProps, dispatch =>
  bindActionCreators(actions, dispatch)
)(QuickOpenModal);
```

It'd be a lot nicer if we updated the second argument from `dispatch => bindActionCreators(actions, dispatch)` to something like this.

```js
function mapStateToProps(state) {
  const selectedSource = getSelectedSource(state);
  let symbols = null;
  if (selectedSource != null) {
    symbols = getSymbols(state, selectedSource.toJS());
  }
  return {
    enabled: getQuickOpenEnabled(state),
    sources: formatSources(getSources(state)),
    selectedSource,
    symbols: formatSymbols(symbols),
    query: getQuickOpenQuery(state),
    searchType: getQuickOpenType(state)
  };
}

const actionCreators = {
  selectSource,
  setQuickOpenQuery,
  highlightLineRange,
  closeQuickOpen,
}

export default connect(mapStateToProps, actionCreators)(QuickOpenModal);
```

This way our components will only have the action creators available to them that they need (making things like the React devtools easier to use) and because Redux automatically binds an function map we won't need to bring `bindActionCreators` into each file.

One problem here though is that since we have this ESLint rule (`"no-shadow": 2`) enabled we'll run into errors from shadowing the function names, we could disable this rule though.

Fischer-L:
Just update the issue status: the merged PR #4640 is the start. Next there will be more PRs coming. And in each PR, probably better to touch one component and derived components at once. So lower risk and easier to review.

Fischer-L:
/unclaim
As the team's discussion [1], the current approach is fine.
[1] https://github.com/devtools-html/debugger.html/pull/4859#issuecomment-353219834

darkwing:
I believe this is complete.

PR #4640, opened 2017-11-09
4532 - Part 1: Update the mapDispatchToProps calls for Breakpoints.js, Popup.js, Preview/index.js

Fischer-L:
Associated Issue: #4532

### Summary of Changes

Update the mapDispatchToProps calls for 
- src/components/Editor/Breakpoints.js
- src/components/Editor/Preview/Popup.js
- src/components/Editor/Preview/index.js

PR #4859, opened 2017-12-05
4532 - Part 4: Update mapDispatchToProps for Footer.js and GutterManu.js

Fischer-L:
Associated Issue: #4532 

### Summary of Changes
Update mapDispatchToProps for Footer.js and GutterManu.js

### DIFF
--- a/src/components/Editor/Footer.js
+++ b/src/components/Editor/Footer.js
@@ -5,7 +5,6 @@
 // @flow
 import React, { PureComponent } from "react";
 import { connect } from "react-redux";
-import { bindActionCreators } from "redux";
 import actions from "../../actions";
 import {
   getSelectedSource,
@@ -199,6 +198,14 @@ class SourceFooter extends PureComponent<Props> {
   }
 }
 
+const {
+  jumpToMappedLocation,
+  togglePrettyPrint,
+  toggleBlackBox,
+  togglePaneCollapse,
+  recordCoverage
+} = actions;
+
 export default connect(
   state => {
     const selectedSource = getSelectedSource(state);
@@ -211,5 +218,11 @@ export default connect(
       endPanelCollapsed: getPaneCollapse(state, "end")
     };
   },
-  dispatch => bindActionCreators(actions, dispatch)
+  {
+    jumpToMappedLocation,
+    togglePrettyPrint,
+    toggleBlackBox,
+    togglePaneCollapse,
+    recordCoverage
+  }
 )(SourceFooter);
--- a/src/components/Editor/GutterMenu.js
+++ b/src/components/Editor/GutterMenu.js
@@ -4,7 +4,6 @@
 
 import { PureComponent } from "react";
 import { showMenu } from "devtools-contextmenu";
-import { bindActionCreators } from "redux";
 import { connect } from "react-redux";
 import { lineAtHeight } from "../../utils/editor";
 import {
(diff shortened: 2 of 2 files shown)


src/components/Editor/GutterMenu.js
```diff
@@ -154,6 +153,19 @@ class GutterContextMenuComponent extends PureComponent {
   }
 }
 
+const {
```
codehag: this would be better done in line 19

src/components/Editor/Footer.js
```diff
@@ -199,6 +198,14 @@ class SourceFooter extends PureComponent<Props> {
   }
 }
 
+const {
```
codehag: this could be done on line 9, since actions are not being used anywhere but here

src/components/Editor/GutterMenu.js
Fischer-L: @codehag 
Is it required to move this up?
Put here for the following reasons:
- closer to where the actions are used, which is `connect` call
- put at the top, the eslint would complain the no-shadow issue. (this has also been mentioned in the issue description [1])
- this style might be thought 
  ```js
  import { ... } from actions
  ```
   but we will face some transpile issue [2]

[1] https://github.com/devtools-html/debugger.html/issues/4532#issue-269498439

[2] https://github.com/devtools-html/debugger.html/pull/4640#discussion_r149972541

jasonLaster:
I wonder if we want to reconsider passing in *just* the actions we use.

PROs: the component has fewer props
CONs: it is extra-code, it might encourage people to try and call the function directly and not use the bound action `this.props.foo()`

Fischer-L:
> I wonder if we want to reconsider passing in just the actions we use.
> PROs: the component has fewer props
> CONs: it is extra-code, it might encourage people to try and call the function directly and not use the bound action this.props.foo()

No strong opinion. Either way has its own Pros/Cons. Just have the team agree on the approach.

> The biggest win here for me is that we remove the bindActionCreators import. I would like to just pass in what we use but I don't know if we have a JS syntax that supports this.

Yes, removal of some variables is good.

codehag:
we could always combine the two:

```javascript
// defined somewhere we have access to the actions object
getDispatchActions(...actionsList) {
  return actionsList.reduce((actionsObject, actionName) => {
     return { ...actionsObject, [actionName]: actions[actionName]};
  }, {});
};

// in component
export default connect(
   state => {
   state => {
@@ -211,5 +218,11 @@ export default connect(
       endPanelCollapsed: getPaneCollapse(state, "end")
       endPanelCollapsed: getPaneCollapse(state, "end")
     };
     };
   },
   },
   getActions(
     "jumpToMappedLocation",
     "togglePrettyPrint",
      "toggleBlackBox",
      // etc
  }
 )(SourceFooter);
 )(SourceFooter);
```

jasonLaster:
I think I'm in favor of passing the actions object in to connect for now. It's better to do that and improve it when there is a clear winner.

codehag:
@jasonLaster can you clarify what you mean with a code example?

jasonLaster:
sure:

```js
export default connect(
   state => {
      endPanelCollapsed: getPaneCollapse(state, "end")
     };
   },
   actions
 )(SourceFooter);
```

jasonLaster:
closing as i think it's simpler to pass the entire actions object in for now

PR #4861, opened 2017-12-05
Synchronize browser_dbg-chrome-debugging.js from m-c

juliandescottes:
Synchronize changes from https://bugzilla.mozilla.org/show_bug.cgi?id=1416711

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ env:
   global:
     - DISPLAY=':99.0'
     - YARN_VERSION='0.24.5'
-    - MC_COMMIT='f27105b62753' # https://hg.mozilla.org/mozilla-central/shortlog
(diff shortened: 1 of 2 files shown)


juliandescottes:
I'm wondering why the previous test wasn't failing though, because addBrowserActors() is no longer available on DebuggerServer. Is it possible that the image used for mochitests is outdated?

codehag:
hmm good question, lets see if it fails now

juliandescottes:
As discussed, failed because we are based on mozilla central rev [f27105b62753](https://hg.mozilla.org/mozilla-central/rev/f27105b62753) which is too old. Will update the PR

juliandescottes:
Travis is green with the new m-c changeset.

PR #4867, opened 2017-12-07
Toggle scopes & callstack panes on pause

GarethSharpe:
### Summary of Changes

* Only renders scopes pane when in paused state

### Test Plan

- [x] Pause debugger and scopes/call-stack panes are shown
- [x] Resume debugger and scopes/call-stack panes are not rendered

### Cases
- [x] Scopes/call-stack panes are not rendered when respective accordion(s) are closed
- [x] Scopes/call-stack panes are not rendered when respective accordion(s) are open

toggle-scopes-done


GarethSharpe:
@codehag so here it is. I've tried various other solutions such as splicing the scopes pane out of the array of items being rendered or using `classnames` to enable/disable the accordion, but each produces a similar (if not exact) error.

jasonLaster:
Nice work @GarethSharpe! Can't wait to check it out

GarethSharpe:
Thanks. Just making a gif now. Will have a few questions to throw your way.

PR #4874, opened 2017-12-07
[DebugLine] Fix editor jumping when debug line is set

jasonLaster:
### Summary of Changes

* Prevent adding a breakpoint from jumping to the pause line
* Fix an issue where iterating through file search results would not move the editor
* Fix an issue where the previous debug line would flash when you step

### Test Plan

mochitests :)

### Screenshots/Videos 

Searching - the current match does not stay in focus because the pause line is always in focus

Breakpoints - adding a breakpoint jumps to the debug line...

Previous line would flash - stepping flashes the previous debug line


src/components/Editor/DebugLine.js
```diff
@@ -8,43 +8,44 @@ import { markText, toEditorPosition } from "../../utils/editor";
...
   lineClass: string
 };
-export class DebugLine extends Component<Props, State> {
-  constructor() {
-    super();
-    this.state = { debugExpression: { clear: () => {} } };
-  }
+
+export class DebugLine extends Component<Props> {
+  debugExpression: null;
```
jasonLaster: I'm switching `debugExpression` to a property so that setting it does not trigger an update.

src/components/Editor/DebugLine.js
```diff
@@ -53,18 +54,23 @@ export class DebugLine extends Component<Props, State> {
...
 
-    const sourceId = selectedFrame.location.sourceId;
+    const { sourceId } = selectedLocation;
     const doc = getDocument(sourceId);
     if (!doc) {
       return;
     }
 
-    const { line, column } = toEditorPosition(sourceId, selectedFrame.location);
+    const { line, column } = toEditorPosition(sourceId, selectedLocation);
```
jasonLaster: i actually am not sure if i need to switch back to `selectedLocation`...

src/components/Editor/index.js
```diff
@@ -247,10 +247,18 @@ class Editor extends PureComponent<Props, State> {
...
+    // keep the jump state around until the real source text is loaded.
+    // 2. Only highlight a line when the selected location changes. This
+    // will prevent highlight other updates like text searches,
+    // and toggling breakpoints.
+    if (
+      selectedSource &&
+      isLoaded(selectedSource.toJS()) &&
+      selectedLocation &&
+      selectedLocation != prevProps.selectedLocation
+    ) {
```
jasonLaster: This was half of the fix for the extra flash on stepping.

src/components/Editor/index.js
```diff
@@ -443,6 +451,7 @@ class Editor extends PureComponent<Props, State> {
     if (
       line !== null &&
       this.lastJumpLine !== null &&
+      this.lastJumpLine !== line &&
```
jasonLaster: this was the other half. I'd like to investigate simplifying this logic with a component that is similar to DebugLine

src/selectors/visibleSelectedFrame.js
```diff
@@ -2,26 +2,30 @@
...
+          location: getLocation(selectedFrame, isGeneratedSource)
+        };
   }
+);
 
-  return {
-    location: getLocation(selectedFrame, isGeneratedSource)
-  };
-}
+export default getVisibleSelectedFrame;
```
jasonLaster: so... everytime we have an issue where we need to use `reselect` to prevent selectors from returning new objects every time I think of @asolove and am grateful for his help this past spring figuring this out.

This issue was introduced when @assafdd moved the debug line to using this new selector so that we grab the correct location based on the selected location. 

It's really hard to detect when a selector change causes a component to re-render aggressively (like 20 times). We should think about a monitoring strategy so we're notified.

src/test/mochitest/browser_dbg-pause-ux.js
```diff
@@ -0,0 +1,44 @@
...
+  pressKey(dbg, "fileSearch");
+  type(dbg, "check");
+  await waitForMatch(dbg, { matchIndex: 0, count: 2 });
+  const matchScrollTop = getScrollTop(dbg);
+  debugger;
+  // ok(isVisibleInEditor(dbg, findElement(dbg, "gutter", 27)), "search match is visible");
+  ok(pauseScrollTop != matchScrollTop, "did not jump to debug line");
+
+  // await waitForever(dbg);
+  // debugger
```
jasonLaster: not done yet, but i like the idea of adding a test that performs some basic tasks like searching, adding breakpoints when we're paused... and makes sure nothing stupid happens

src/components/Editor/DebugLine.js
```diff
@@ -66,29 +54,24 @@ export class DebugLine extends Component<Props, State> {
 
     const { line, column } = toEditorPosition(sourceId, selectedFrame.location);
```
jasonLaster: we should only scroll the editor from the Editor component. And, anyways this should be redundant.

src/components/Editor/Editor.css
```diff
@@ -190,7 +190,7 @@ debug-expression-error {
 }
 
 .highlight-line .CodeMirror-line {
-  animation: fade-highlight-out 1.5s normal forwards;
+  animation: fade-highlight-out 5s normal forwards;
```
jasonLaster: oops - this should be reverted

src/components/Editor/HighlightLine.js
```diff
@@ -0,0 +1,104 @@
...
+      return false;
+    }
+
+    // Don't highlight the current debug line
+    if (this.isDebugLine(nextProps.selectedFrame, nextProps.selectedLocation)) {
+      return false;
+    }
+
+    return true;
+  }
```
jasonLaster: is it nice to have a verbose function or a more consolidated function?

src/components/Editor/HighlightLine.js
```diff
@@ -0,0 +1,104 @@
...
+  editor: Object,
+  selectedFrame: Frame,
+  selectedLocation: Location
+};
+
+export class DebugLine extends Component<Props> {
+  componentDidMount() {
+    const { selectedLocation, selectedFrame, editor } = this.props;
+    // this.setHighlightLine(selectedLocation, selectedFrame, editor);
+  }
```
jasonLaster: we should be able to remove this function because we dont need to show the highlight when we mount

src/components/Editor/HighlightLine.js
```diff
@@ -0,0 +1,104 @@
...
+  isDebugLine(selectedFrame: Frame, selectedLocation: Location) {
+    if (!selectedFrame) {
+      return;
+    }
+
+    return (
+      selectedFrame.location.sourceId == selectedLocation.sourceId &&
+      selectedFrame.location.line == selectedLocation.line
+    );
+  }
```
jasonLaster: we can move the out of the class

src/components/Editor/index.js
```diff
@@ -415,48 +379,17 @@ class Editor extends PureComponent<Props, State> {
...
-    // class, etc).
-    if (this.lastJumpLine !== null) {
-      clearLineClass(this.state.editor.codeMirror, "highlight-line");
-    }
-
-    let line = null;
-    if (selectedLocation.line >= 0) {
-      line = this.scrollToPosition();
-    }
+  scrollToLocation(nextProps) {
```
jasonLaster: this function is causing us to jump a bit... so not quite ready for showtime

jasonLaster: cleaned it up by using `createLocation` in selectLocation so everything had a nice default.

src/components/Editor/HighlightLine.js
lukaszsobek: Why not both, 4 lines of comments - one for each condition above a consolidated if statement. Feels easier to read and fewer lines of code as well.

jasonLaster:
seeing some mochitest issues:

+ browser_dbg-breaking.js
+ browser_dbg-console-link.js
+ browser_dbg-editor-highlight.js
+ browser_dbg-editor-select.js
+ browser_dbg-expressions-error.js
+ browser_dbg-iframes.js
+ browser_dbg-navigation.js
+ browser_dbg-outline.js
+ browser_dbg-sourcemaps.js

will investigate tomorrow

src/components/Editor/DebugLine.js
```diff
@@ -4,110 +4,107 @@
...
+  getVisibleSelectedFrame,
+  getPause,
+  getSelectedSource
+} from "../../selectors";
+
+import type { Frame, Pause } from "debugger-html";
+import type { SourceRecord } from "../../reducers/types";
 
 type Props = {
   editor: Object,
```
codehag: this is no longer used in this component

src/components/Editor/tests/DebugLine.spec.js
```diff
@@ -0,0 +1,132 @@
...
+  const doc = createMockDocument(clear);
+  setDocument(props.selectedSource.get("id"), doc);
+
+  const component = shallow(<DebugLine.WrappedComponent {...props} />, {
+    lifecycleExperimental: true
+  });
+  return { component, props, clear, editor, doc };
+}
+
+describe("DebugLine Component", () => {
```
codehag: great stuff :+1:

src/components/Editor/tests/DebugLine.spec.js
```diff
@@ -0,0 +1,132 @@
...
+  describe("pausing at the first location", () => {
+    it("should show a new debug line", async () => {
+      const { component, props, doc } = render({
+        selectedSource: makeSourceRecord("foo", { loadedState: "loaded" })
+      });
+
+      const selectedFrame = createFrame(2);
```
codehag: the `2` here might be better in a const. that way the relationship between it and the expected result is clear and descriptive. 

```javascript
const lineNumber = 2;
const editorNumber = lineNumber - 1;
createFrame(lineNumber)

//.. 
expect(doc.addLineClass.mock.calls).toEqual([
  [editorLine,  "line", "new-debug-line"]
]);

File: src/components/Editor/tests/DebugLine.spec.js [hunk 16]
```diff
@@ -0,0 +1,132 @@
...
+    });
+
+    describe("pausing at a new location", () => {
+      it("should replace the first debug line", async () => {
+        const { props, component, clear, doc } = render({
+          selectedSource: makeSourceRecord("foo", { loadedState: "loaded" })
+        });
+
+        component.instance().debugExpression = { clear: jest.fn() };
+        component.setProps({ ...props, selectedFrame: createFrame(2) });
```
codehag: same comment here -- its not clear wht the relationship between the numbers are, would be better if they were name constants

src/components/Editor/tests/Editor.spec.js
```diff
@@ -42,38 +29,123 @@ function render(overrides = {}) {
...
-      selectedSource: I.fromJS({ loadedState: "loading" })
+  describe("When loaded", () => {
+    it("should show text", async () => {
+      const { component, mockEditor, props } = render({});
+
+      await component.setState({ editor: mockEditor });
+      await component.setProps({
+        ...props,
+        selectedSource: createMockSource({ loadedState: "loaded" }),
+        selectedLocation: { sourceId: "foo", line: 3, column: 1 }
```
codehag: would be better to get the mock values out so that we can test against them, and its clearer when reading the tests what is happening

src/components/Editor/tests/Editor.spec.js
```diff
@@ -42,38 +29,123 @@ function render(overrides = {}) {
...
+        ...props,
+        selectedSource: createMockSource({
+          loadedState: "loaded"
+        }),
+        selectedLocation: { sourceId: "foo", line: 1, column: 1 }
+      });
+
+      expect(mockEditor.setText.mock.calls).toEqual([
+        ["Loading…"],
+        ["the text"]
```
codehag: same here -- it is unclear what relationship these text strings have to the expected behavior

src/reducers/sources.js
```diff
@@ -294,8 +294,8 @@ export function getSourceByURL(state: OuterState, url: string): ?SourceRecord {
   return getSourceByUrlInSources(state.sources.sources, url);
 }
 
-export function getGeneratedSource(state: OuterState, source: any) {
-  if (!isOriginalId(source.id)) {
+export function getGeneratedSource(state: OuterState, source: ?Source) {
```
codehag: nice!

src/selectors/visibleSelectedFrame.js
```diff
@@ -2,26 +2,34 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
 
+// @flow
```
codehag: 🎉

src/utils/source.js
```diff
@@ -293,9 +293,14 @@ function getMode(source: Source, sourceMetaData: SourceMetaDataType) {
   return { name: "text" };
 }
 
-function isLoaded(source: SourceRecord) {
+function isLoaded(source: ?SourceRecord) {
+  if (!source) {
```
codehag: should this ever happen?

jasonLaster: it could if the caller doesn't check if there's a source... Dropped it because the places we use isLoaded should be safe now.

jasonLaster:
[repeats](https://treeherder.mozilla.org/#/jobs?repo=try&revision=b9137ebdc6604f229c94bdd13d17d0984586f1b7)
[debugger](https://treeherder.mozilla.org/#/jobs?repo=try&revision=4a654da2b122fee24e4e77c7b84827a9c631ba15)

Issue #4868, opened 2017-12-07
Regressions on closing and reopening the debugger


codehag:
in the latest release, two regressions were spotted:

when closing and reopening the debugger:
- source tab is blank
- breakpoints are persisted but not visible in the gutter

https://bug1423158.bmoattachments.org/attachment.cgi?id=8935329

We should get this fixed before the next release, next week

PR #4875, opened 2017-12-07
Show source text when devtools opens

jasonLaster:
Associated Issue: #4868

### Summary of Changes

This addresses a recent issue where reloading the debugger would no longer show the selected source.

The issue was timing related where the source loaded before the editor had been setup. This fixes it by showing the source text when the editor is created.

### DIFF
--- a/src/components/Editor/index.js
+++ b/src/components/Editor/index.js
@@ -228,7 +228,7 @@ class Editor extends PureComponent<Props, State> {
     this.setSize(nextProps);
   }
 
-  componentDidUpdate(prevProps) {
+  componentDidUpdate(prevProps, prevState) {
     // This is in `componentDidUpdate` so helper functions can expect
     // `this.props` to be the current props. This lifecycle method is
     // responsible for updating the editor annotations.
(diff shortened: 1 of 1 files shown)


src/components/Editor/index.js
```diff
@@ -253,6 +253,13 @@ class Editor extends PureComponent<Props, State> {
     if (selectedSource && isLoaded(selectedSource.toJS())) {
       this.highlightLine();
     }
+
+    // NOTE: when devtools are opened, the editor is not set when the source loads
```
lukaszsobek: The linter thinks this line is too long :)

jasonLaster: lol - i'll help it out

src/components/Editor/index.js
```diff
@@ -253,6 +253,14 @@ class Editor extends PureComponent<Props, State> {
     if (selectedSource && isLoaded(selectedSource.toJS())) {
       this.highlightLine();
     }
+
+    // NOTE: when devtools are opened, the editor is not set when
+    // the source loads so we need to wait until the editor is
+    // set to update the text and size.
+    if (!prevState.editor && this.state.editor) {
```
wldcordeiro: Only note if you're looking for existence it maybe better to do this as sometimes you may get null or undefined.
```js
if (prevState.editor == null && this.state.editor != null)`
```

wldcordeiro:
Merge conflicts but otherwise looks good 😄

Issue #4437, opened 2017-10-19
[Editor] is blank when devtools opens


jasonLaster:
I'm seeing a blank editor when i open devtools w/ a selected tab.

STR:

1. go to [todomvc][t]
2. select todo-view.js
3. close devtools
4. re-open devtools

ER: see the source
AR: editor is blank

It would be nice to get a mochitest here

![](https://bugzilla.mozilla.org/show_bug.cgi?id=1409101)

[t]:https://bugzilla.mozilla.org/show_bug.cgi?id=1409101

jainsneha23:
@jasonLaster Can you check if the bug still exists in master?

jasonLaster:
hey @jainsneha23 this is a panel bug which is tough to debug. I recommend grabbing something else :)

lukaszsobek:
So basically we want to preserve the state complete with focus and pretty printed tabs? How are we keeping the tabs across reloads now?

lukaszsobek:
Issue opened here: https://github.com/devtools-html/devtools-core/issues/839

lukaszsobek:
Issue should be fixed now @jasonLaster ?

PR #4878, opened 2017-12-07
Adds rejected promise handling to bpClients

lukaszsobek:
Associated Issue: #4437

### Summary of Changes

* Added promise handling in case client.onConnect rejects

### Screenshots/Videos
empty-tab

### DIFF
--- a/src/client/index.js
+++ b/src/client/index.js
@@ -45,7 +45,13 @@ async function onConnect(
   });
 
   bootstrapWorkers();
-  const { bpClients } = await client.onConnect(connection, actions);
+  const { bpClients } = await client
+    .onConnect(connection, actions)
+    .then(clients => clients)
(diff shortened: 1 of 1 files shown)


src/client/index.js
```diff
@@ -45,7 +45,13 @@ async function onConnect(
...
 
   bootstrapWorkers();
-  const { bpClients } = await client.onConnect(connection, actions);
+  const { bpClients } = await client
+    .onConnect(connection, actions)
+    .then(clients => clients)
+    .catch(err => {
+      console.log(err);
+      return {};
+    });
```
jasonLaster: we should do this in the launchpad `onConnect` function

lukaszsobek:
Moved here: https://github.com/devtools-html/devtools-core/pull/840

PR #4881, opened 2017-12-07
Run mochitests in debug mode

jasonLaster:
### Summary of Changes

we've discussed running our mochitests in debug mode on CI for awhile to look for leaks. this gives it a try

questions: 
1. does it run at all?
2. how slow is it?
3. do we need extra checks to parse the new output?


jasonLaster:
timed out. i'll see if it is easy to bump the timeout to get a sense of how long this run would be...

Issue #4883, opened 2017-12-08
[Flow] Improve flow behavior


lukaszsobek:
currently flow creates a lot of terminal spam and caches results, which has a negative influence on debugging workflow. It would be nice if we changed the default `yarn flow` behavoir to be `yarn flow check`, not only is it less verbose to the output, but it also forces a re-check without using cache.

lukaszsobek:
Testing proved the proposed method is slower than what we have now.

PR #4884, opened 2017-12-08
Changes flow behavior from server to on demand basis

lukaszsobek:
Associated Issue: #4883

Rationale:
Currently we use `flow` in its default behavior, which is short for `flow status`. `flow status` asks the currently running flow server for a list of errors which the flow server is supposed to be continually monitoring. If there is no server `flow status` will spawn one. Also it will sometimes cache results.

However, for the purpose of testing we use `flow` on an on demand basis - expecting a full and fresh set of results every time. This is the default behavior of `flow check`. A new flow server is initiated, errors are checked, the server exits. No caching, no verbose server initiated messages.

### Summary of Changes

* changed flow behavior from keeping the flow server running to on demand execution

### Test Plan
executed `yarn flow`

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,17 +17,20 @@
   "scripts": {
     "start": "node bin/dev-server",
     "start-app": "TARGET=application node bin/dev-server",
-    "flow": "flow",
-    "eslint-check": "eslint --print-config .eslintrc.js | eslint-config-prettier-check",
+    "flow": "flow check",
+    "eslint-check":
+      "eslint --print-config .eslintrc.js | eslint-config-prettier-check",
     "prettier": "node bin/prettier.js",
     "license-check": "devtools-license-check",
     "links": "ls -l node_modules/ | grep ^l || echo 'no linked packages'",
     "lint": "run-p lint-css lint-js lint-md",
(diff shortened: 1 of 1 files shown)


jasonLaster:
@lukaszsobek i'm worried that this will be slower. What is the time difference between runnung `yarn check` twice and running `yarn flow` twice?

lukaszsobek:
@jasonLaster - It's really interesting. And you are right, flow check is slower:
First run they preform the same (about 70 seconds) to fire up the flow server and everything, the following iterations make the numbers really interesting. Because `flow check` takes 18 seconds for every consecutive run it really seems to fire up the server, check the code and kill the server after. So even if there are no changes the server has to be instantiated again. The regular `flow status` keeps the server alive for a certain amount of time, so even though the checks are the same there is no overhead in terms of winding the server up, so a consecutive check within the server lifetime only takes 0.2 seconds.

Conclusion: the idea of using `flow check` as standard of flow validation can definitely be discarded as the standard way of doing it is faster. Keeping the server alive benefits us about 15 seconds per check and besides, if we do need `yarn flow check` we can easily execute it manually.

Issue #4896, opened 2017-12-09
[Performance] Debugger in background causes hangs


digitarald:
STR
- Opened https://alpha.giveth.io/ with Debugger set to `Pause on uncaught exception`
- Paused on `value` or similar, showing a non-pretty minified JS file
- Switch to a different tab (I opened https://github.com/devtools-html/debugger.html/issues/new to file a bug about that the `js` file doesn't have a prettify option)

AR:
- Debugger in background causes massive hangs on the main process, even with tab and devtools in the background

ER:
- Debugger in background behaves nicely

https://perfht.ml/2iGWuuc - Rendering a lot of React, with the typical expensive updates on source editor

jasonLaster:
The first half of the time is spent in `setText` the second half is spent in `getCallSites`. I opened a PR, which fixes the callsite feature flag. We'll have to look into setText separately.


<img width="801" alt="screen shot 2017-12-09 at 11 31 25 am" src="https://user-images.githubusercontent.com/254562/33797387-d0082e2c-dcd4-11e7-9fc3-087e5bd0c706.png">

We'll want to look into speeding up `setText` as well.

<img width="1552" alt="screen shot 2017-12-09 at 11 30 01 am" src="https://user-images.githubusercontent.com/254562/33797395-f240096a-dcd4-11e7-8924-31bf4f856b45.png">
<img width="1552" alt="screen shot 2017-12-09 at 11 29 50 am" src="https://user-images.githubusercontent.com/254562/33797396-f24a8868-dcd4-11e7-84a0-3213b7e5351e.png">

digitarald:
We we have mechanisms to not do heavy UI updates when in background?

jasonLaster:
I believe @ochameau has a patch for this.

ochameau:
I did the work for the console as an example here:
  https://bugzilla.mozilla.org/show_bug.cgi?id=1399242
But work is still needed all other panels, to use the VisibilityHandler React component:
  https://searchfox.org/mozilla-central/source/devtools/client/shared/components/VisibilityHandler.js
I filled a bug to do that for the debugger here:
  https://bugzilla.mozilla.org/show_bug.cgi?id=1419352
  I filled on bugzilla as it would require some tweaks to DAMP test.

jasonLaster:
Thanks. Maybe we can give this a try when we're together this week

ochameau:
[...]

PR #4898, opened 2017-12-09
Update CallSites flag

jasonLaster:
Associated Issue: #4896

### Summary of Changes

Our callsite flag which is used to feature flag the Column Breakpoints UI previously rendered the breakpoints in the background and just disabled the event handling so they wouldn't be shown. 

This means that we were paying the price of fetching and rendering the breakpoints with codemirror, with none of the benefits.

Here are the perf wins we'll see in minified files. Most of the time is spent fetching the column breakpoints:

screen shot 2017-12-09 at 11 24 11 am

screen shot 2017-12-09 at 11 23 57 am

### Further Improvements

The first half of the time is spent in `setText` the second half is spent in `getCallSites`

screen shot 2017-12-09 at 11 31 25 am

We'll want to look into speeding up `setText` as well.

screen shot 2017-12-09 at 11 30 01 am
screen shot 2017-12-09 at 11 29 50 am

### DIFF
--- a/src/components/Editor/CallSites.js
+++ b/src/components/Editor/CallSites.js
@@ -4,7 +4,6 @@
 
 import React, { Component } from "react";
 import { connect } from "react-redux";
-import { features } from "../../utils/prefs";
 
 import { range, keyBy, isEqualWith } from "lodash";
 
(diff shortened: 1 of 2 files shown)


src/components/Editor/CallSites.js
```diff
@@ -58,22 +57,18 @@ class CallSites extends Component {
...
   componentDidUnMount() {
     const { editor } = this.props;
     const codeMirrorWrapper = editor.codeMirror.getWrapperElement();
 
-    if (features.columnBreakpoints) {
-      codeMirrorWrapper.addEventListener("click", e => this.onTokenClick(e));
-      document.body.removeEventListener("keydown", e => this.onKeyDown);
-      document.body.removeEventListener("keyup", this.onKeyUp);
-    }
+    codeMirrorWrapper.addEventListener("click", e => this.onTokenClick(e));
```
bomsy: we should be removing the listener here right?

src/components/Editor/CallSites.js
```diff
@@ -58,22 +57,18 @@ class CallSites extends Component {
...
+    document.body.removeEventListener("keydown", e => this.onKeyDown);
```
bomsy: this might not be actually removing this event listener as the listener function is different from what passed when added.
this should probably be `document.body.removeEventListener("keydown", this.onKeyDown);`

src/components/Editor/CallSites.js
```diff
@@ -58,22 +57,18 @@ class CallSites extends Component {
...
-      codeMirrorWrapper.addEventListener("click", e => this.onTokenClick(e));
-      document.body.addEventListener("keydown", this.onKeyDown);
-      document.body.addEventListener("keyup", this.onKeyUp);
-    }
+    codeMirrorWrapper.addEventListener("click", e => this.onTokenClick(e));
+    document.body.addEventListener("keydown", this.onKeyDown);
+    document.body.addEventListener("keyup", this.onKeyUp);
   }
 
   componentDidUnMount() {
```
NickMalt: hm, correct me if I'm wrong here, but there is no `componentDidUnMount` lifecycle method in React. Only [componentWillUnmount](https://reactjs.org/docs/react-component.html#componentwillunmount).

bomsy: Nice catch! Thanks!

Issue #4886, opened 2017-12-08
name collision: getOutOfScopesLocations


codehag:
We have a naming collision between getOutOfScopesLocation in `workers/parser/getOutOfScopesLocation.js` and getOutOfScopesLocation in `reducers/ast.js`

the two functions do different thing, one determines which scopes are out of scope, the other retrieves cached data. since we have a standard way of naming selectors with the `get` prefix, we should rename the parser code to `findOutOfScopeLocations`.

the easiest way to find this is using the arguments, the parser takes two arguments, a source id, and an astlocation. the other one only accepts the global state.

we should try to limit such naming collisions, I will list them as I find them.

This issue is beginner friendly, but a bit frustrating. make sure you run the tests while doing the renaming.

PR #4900, opened 2017-12-10
Fix name collisions: getOutOfScopeLocations

pradeepgangwar:
Associated Issue: #4886

### Summary of Changes

* Fix naming collisions between `getOutOfScopeLocations` in `workers/parser`/getOutOfScopeLocations.js and `getOutOfScopeLocations` in `reducers/ast.js`


pradeepgangwar:
@codehag Please review.

codehag:
Shoot, i think i wasn't entirely clear in the issue.

we don't want to rename both, just one. I was thinking we should rename the parser (which you did, perfect!) but not the ast reducer, since we use the `get` prefix on most of our reducer selectors. The ast reducer should keep the name `getOutOfScopeLocations`..

Let me know if anything is unclear!

pradeepgangwar:
@codehag I am not getting where i have done mistake. I changed `src/actions/ast.js` to this 
```
import {
   getSymbols,
   getEmptyLines,
-  getOutOfScopeLocations,
+  findOutOfScopeLocations,
   isReactComponent
 } from "../workers/parser";
```
because it is imported from parser. Other than this I haven't touched `ast reducer` code. Please correct me if i am doing wrong somewhere.

codehag:
oh, my mistake! you are absolutely right. I will read through the pr again when i have more time :)

thanks for the correction!

Issue #1989, opened 2017-02-10
[Breakpoints] can't set breakpoint in prettified source


tromey:
I tried the test case in [bug 1327713](https://bugzilla.mozilla.org/show_bug.cgi?id=1327713).  After prettifying the source, when I try to set a breakpoint, the breakpoint briefly appears, and then disappears.

I don't see an stack trace in the console, but enabling pause-on-exception shows an exception coming from `addBreakpoint` saying: "query object's 'line' property is neither undefined nor an integer".

[TEST PAGE](https://hex-lines.glitch.me/)

jasonLaster:
**STR 1:**
1. Open attached "testcase 1"
2. Open debugger, prettify script containing word "click"
3. Set breaking on "click" event, click on the page
4. Keep clicking "Step in" button until line "console.log(123);" is executed

**AR:**
 When debugger steps into a function, then, if the first token in that function is "function",
 debugger jumps to line 1. That line becomes highlighted with green; caret is displayed in that line

**ER:**
 When debugger steps into a function, some sensible line inside that function should be selected.
 In other words, debugger should select lines in the very same sequence as in STR_2 (see below).


**STR_2:**
1. Open attached "testcase 1"
2. Set breaking on "wheel" event, hover mouse over the page, rotate mouse wheel down once
3. Keep clicking "Step in" button until line "console.log(123);" is executed

AR:  Everything's OK


Here's the testcase 1:

```html
<script src="data:text/javascript,
(function(){
  var A = {};
  window.A = A;
  A.F = function(){
    A.G();
  },
  A.G = function(){
    function _G(){};
    A.H();
  },
  A.H = function(){
    A.I()
  },
  A.I = function(){
    function _I(){};
    A.J();
  };
  A.J = function(){
    function _J(){};
    A.K();
  };
  A.K = function(){
    console.log(123);
  };
  
  addEventListener('click',A.F,false);
})()
//.js"></script>

<!-- a bit confusing -->

<script src="data:text/javascript,
(function () {
%0A  var A = {
%0A  };
%0A  window.A = A;
%0A  A.F = function () {
%0A    A.G();
%0A  },
%0A  A.G = function () {
%0A    function _G() {
%0A    };
%0A    A.H();
%0A  },
%0A  A.H = function () {
%0A    A.I()
%0A  },
%0A  A.I = function () {
%0A    function _I() {
%0A    };
%0A    A.J();
%0A  };
%0A  A.J = function () {
[...]

PR #4902, opened 2017-12-12
[Breakpoints] can't set breakpoint in prettified source

NickMalt:
Associated Issue: #1989 

### Summary of Changes
Unsuccessfully trying to reproduce bug in mochitest. Everything seems ok 🤔 . 
Can anyone play with it? Do I need to change something in STR?

### DIFF
--- a/src/test/mochitest/browser.ini
+++ b/src/test/mochitest/browser.ini
@@ -59,7 +59,6 @@ support-files =
   examples/simple1.js
   examples/simple2.js
   examples/simple3.js
-  examples/simple4.js
   examples/frames.js
   examples/script-mutate.js
   examples/script-switching-02.js
--- a/src/test/mochitest/browser.ini
+++ b/src/test/mochitest/browser.ini
@@ -45,6 +45,7 @@ support-files =
   examples/doc-sources.html
   examples/doc-return-values.html
   examples/doc-wasm-sourcemaps.html
+  examples/doc-prettified-breakpoints.html
   examples/asm.js
   examples/async.js
   examples/bogus-map.js
@@ -58,6 +59,7 @@ support-files =
   examples/simple1.js
   examples/simple2.js
   examples/simple3.js
+  examples/simple4.js
   examples/frames.js
   examples/script-mutate.js
   examples/script-switching-02.js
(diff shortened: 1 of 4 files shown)


jasonLaster:
Thanks @NickMalt, definitely a great start! Try using the inline script with the funky `%0A  `

jasonLaster:
Hey @NickMalt, here is an HTML test case w/ the crazy `%0A` new lines...

https://bug1327713.bmoattachments.org/attachment.cgi?id=8822992

I just tested and i think i was able to reproduce the bug

NickMalt:
After some debugging I think that I caught this bug:

1) Here we send correct parameters but got `{ line: null, column: null }`.  Which is why breakpoint flashing. https://github.com/devtools-html/debugger.html/blob/223d35a570d4762000c736e00e5a16b09c9ae8c6/src/utils/source-maps.js#L17-L20

2) [Here](https://dxr.mozilla.org/mozilla-central/source/devtools/client/shared/source-map/worker.js?q=const+%7B+line%2C+column+%7D+%3D+map.generatedPositionFor%28%7B&redirect_type=single#674) we handle this request in source-map worker. But [_findSourceIndex](https://dxr.mozilla.org/mozilla-central/source/devtools/client/shared/source-map/worker.js#2912) return `-1` as if this source did not exist.
￼
image
image


3) The problem is in property `sourceUrl` mismatch.
This one we send to worker:
image
This one we have in worker: 
image
And we have this mismatch because of path [normalization](https://dxr.mozilla.org/mozilla-central/source/devtools/client/debugger/new/parser-worker.js#28487) in [worker](https://dxr.mozilla.org/mozilla-central/source/devtools/client/shared/source-map/worker.js#2474)
image

4) And if we add this patch to [normalize function](https://dxr.mozilla.org/mozilla-central/source/devtools/client/debugger/new/parser-worker.js#28487), everything works ok.
image

### Before/After
bp before
bp after



@jasonLaster The question is, how to properly change normalize function? I’m probably miss something, because do not see the whole picture.

And how should I continue this? I mean how to contibute to `mozila-central`?

jasonLaster:
Nice hunting!

jasonLaster:
Hey @NickMalt, great analysis I think you're right about how we're escaping the source url.

The normalize function comes from this file: https://github.com/mozilla/source-map/blob/master/lib/util.js in the mozilla source-map lib.

I think we can probably change how we create the URL in the debugger so it's not escaped... not sure where though.

jasonLaster:
Hey how's it going?

PR #4915, opened 2017-12-13
[WIP] Use devtools-reps-0.19.0

nchevobbe:
Updating to the latest devtools-reps means that we need to change how we instantiate ObjectInspector.

This is a wip because : 
- ~~devtools-reps-0.17.0 is not published yet~~
- ~~I didn't checked the tests~~
- ~~The Popup does not seem to work.~~

Yet, watch expressions and scopes are fine.


jasonLaster:
This is looking good @nchevobbe. Nice work w/ `createObjectClient`

nchevobbe:
So the only thing with this patch is that the preview show you the root (which is then autoexpanded)

before: 
screen shot 2017-12-15 at 12 59 43

after: 
screen shot 2017-12-15 at 13 03 08

nchevobbe:
Also: photon colors !

nchevobbe:
Here's the "flashing" issue I talked you about @jasonLaster : 

dec-17-2017 12-38-01

Basically, it's the time we have to wait for properties to be loaded. This is a big case here since we target document, which has a lot of things.
We could prevent to render until we have everything loaded to make it better, but the latency to actually show something in the preview might be an issue ? (probably not more than what we were doing previously).

nchevobbe:
Okay so yeah, just tested the same case in Nightly, and we do have this latency on document, the only thing being we don't show anything until we have all properties loaded.

jasonLaster:
@nchevobbe i don't mind delaying showing the preview. It should be snappy on most objects.

I suppose we could also try some animations too. Perhaps there's something fun we could do with animating the yellow highlight? dunno

nchevobbe:
okay, nice
So do we have clear next steps to make this PR landed ?

jasonLaster:
Lets treat the preview loading behavior as a next step. And just focus on getting the tests passing now

nchevobbe:
Sure, CI is failing because of Flow. Let me fix this now

nchevobbe:
Erf, at this point, we would break preview for React and Immutable https://github.com/devtools-html/debugger.html/pull/4915/files#diff-2e80e25aeaf08281d8427771196d61e1R101 . 😕

@MikeRatcliffe do you think you could have a look and see how we could do thing in the ObjectInspector to autoLoad properties and get them back in some way, so we don't have to change much code for the Preview here ?

jasonLaster:
@nchevobbe happy to look in a bit

jasonLaster:
Did a quick look

- Immutable did not seem to regress
- the preview font is bigger
- the source tree seems clipped
- previewing a react component breaks STR: go to this [site](https://devtools-reporter-hzsvksiqym.now.sh) and add a breakpoint in App.js 

screen shot 2017-12-19 at 9 17 12 am
screen shot 2017-12-19 at 9 13 51 am
screen shot 2017-12-19 at 9 13 09 am

MikeRatcliffe:
When we are trying to preview `this` inside a React component the following error is thrown:
>Warning: setState(...): Can only update a mounted or mounting component. This usually means you called setState() on an unmounted component. This is a no-op. Please check the code for the ObjectInspector component. 13 warning.js:33
	printWarning warning.js:33
	warning warning.js:57
	getInternalInstanceReadyForUpdate ReactUpdateQueue.js:46
	enqueueSetState ReactUpdateQueue.js:207
	ReactComponent.prototype.setState ReactBaseClasses.js:62
	setExpanded index.js:310:8

nchevobbe:
Okay, I guess we need some ObjectInspector changes here

nchevobbe:
source tree is now fixed.
We use devtools-reps-0.17.0

I think the error about setState is when we hover quickly a variable and then mouseleave.
This causes the ObjectInspector to be mounted, rendered, making him fetching the first level properties, and thus trying to do a setState. But at this point, the OI is unmounted because we not longer hover the variable in the debugger.
I am not sure how this can be fixed, but I think I want to keep the same behavior as before , which means we need to modify reps.

nchevobbe:
@jasonLaster The popup is working just like before now, but i did not tested the React and Immutable cases.
I didn't fixed the mochitests and did not refactor the loadObjectProperties action much because I wanted to check if you are okay with this approach.

This is based on https://github.com/devtools-html/devtools-core/pull/884, which will probably be in a 0.19.0 reps package.

There is also a minor styling issue (see https://github.com/devtools-html/devtools-core/issues/886)

src/client/firefox.js
```diff
@@ -8,11 +8,19 @@ import { setupCommands, clientCommands } from "./firefox/commands";
 import { setupEvents, clientEvents } from "./firefox/events";
 import { features } from "../utils/prefs";
 
+let DebuggerClient;
+
+function createObjectClient(grip: Object) {
```
codehag: could we use the Grip type here from flow-typed/debugger-html? https://github.com/devtools-html/debugger.html/blob/master/flow-typed/debugger-html.js#L235

src/components/Editor/Preview/Popup.js
```diff
@@ -46,24 +53,32 @@ export class Popup extends Component<Props> {
...
       value,
-      editor,
-      range
+      expression,
+      loadObjectProperties,
+      loadedObjects
     } = this.props;
+    const root = createNode(null, expression, expression, { value });
+
+    if (!nodeIsPrimitive(root) && value.actor && !loadedObjects[value.actor]) {
```
codehag: before, the value was potentially null -- is that no longer the case? maybe we can type this?

src/components/Editor/Preview/Popup.js
```diff
@@ -46,24 +53,32 @@ export class Popup extends Component<Props> {
...
     const {
-      loadObjectProperties,
-      loadedObjects,
       value,
-      editor,
-      range
+      expression,
+      loadObjectProperties,
+      loadedObjects
     } = this.props;
```
codehag: The function has changed to componentWillMount, so this.props will be representing the previous props. do we want to use this.props or nextProps?

src/components/Editor/Preview/Popup.js
```diff
@@ -72,13 +87,18 @@ export class Popup extends Component<Props> {
     }
   }
 
-  getChildren(root: Object, getObjectProperties: Function) {
-    const actors = {};
+  getChildren(root: Object) {
```
codehag: is root of type Node? https://github.com/devtools-html/debugger.html/blob/master/src/utils/sources-tree/types.js#L14

src/components/Editor/Preview/Popup.js
```diff
@@ -103,16 +123,18 @@ export class Popup extends Component<Props> {
   }
 
   renderObjectPreview(expression: string, root: Object, extra: Object) {
```
codehag: also maybe here we can use the node type for root:  https://github.com/devtools-html/debugger.html/blob/master/src/utils/sources-tree/types.js#L14

it will need to be moved to a more general type

src/components/Editor/Preview/Popup.js
```diff
@@ -166,31 +188,20 @@ export class Popup extends Component<Props> {
...
         autoExpandDepth={0}
         disableWrap={true}
         disabledFocus={true}
         openLink={openLink}
-        getObjectProperties={getObjectProperties}
-        loadObjectProperties={loadObjectProperties}
-        // TODO: See https://github.com/devtools-html/debugger.html/issues/3555.
-        getObjectEntries={actor => {}}
-        loadObjectEntries={grip => {}}
+        createObjectClient={grip => createObjectClient(grip)}
```
codehag: :+1: nice!

src/components/SecondaryPanes/Expressions.js
```diff
@@ -141,11 +138,7 @@ class Expressions extends PureComponent<Props, State> {
               this.editExpression(expression, options)
             }
             openLink={openLink}
-            getObjectProperties={id => loadedObjects[id]}
-            loadObjectProperties={loadObjectProperties}
-            // TODO: See https://github.com/devtools-html/debugger.html/issues/3555.
-            getObjectEntries={actor => {}}
-            loadObjectEntries={grip => {}}
+            createObjectClient={grip => createObjectClient(grip)}
```
codehag: 🎉

src/components/SecondaryPanes/tests/__snapshots__/Expressions.spec.js.snap
```diff
@@ -151,12 +142,9 @@ exports[`Expressions should render 1`] = `
     >
       <ObjectInspector
         autoExpandDepth={0}
+        createObjectClient={[Function]}
```
codehag: its just a code deletion party! hurray!

src/client/firefox.js
nchevobbe: sure

src/components/Editor/Preview/Popup.js
nchevobbe: it is, thanks for pointing to the type

src/components/Editor/Preview/Popup.js
nchevobbe: yes

nchevobbe:
I think everything is working fine now, except: 
- it requires devtools-reps-0.19.0 (on its way)
- it certainly breaks ChromeScope
- and mochitest who rely on LOAD_OBJECT_PROPERTIES (the action was changed to only set popup object properties)

jasonLaster:
i can `git rm ChromeScope` it was an experiment for CDP but hasnt been touched in awhile

nchevobbe:
Okay, so everything is working now, devtools-reps-0.19.0 has been published, we're in a good place.
Except for those tests.
A bunch of them are using something like `await waitForDispatch(dbg, "LOAD_OBJECT_PROPERTIES");` which awaits forever since we don't have this action anymore (or even something that could replace it).

I'm not sure how to tackle this, should I edit all the tests so they don't rely on this action, but more on the UI ? Something like "wait until this .tree element has more than 1 node" ?

That's how it's done in the console, but here it will differ from what the test are doing 🤔 

@jasonLaster do you have an idea how I can tackle this without rewriting all the tests ?

jasonLaster:
yep - lets wait for the dom to have rendered loaded scope variable...

nchevobbe:
okay, i'm down to 3 mochitests failing.
At least 2 of them are failing because the PR breaks the Scopes "remapping", I need to check what's going on

nchevobbe:
okay I get it.
The ObjectInspector keeps a Map of cached nodes, which is never cleaned up.
When remapping the scope, we use the same OI instance, and so auto-expanded node, which were cached, are not updated.

This means one more change in Reps to fix this :/

nchevobbe:
@jasonLaster everything is working now (with https://github.com/devtools-html/devtools-core/pull/905), so this PR is good for review.
It seems that there are conflicts I'll resolve tomorrow

wldcordeiro:
@nchevobbe you need to add `setPopupObjectProperties` to `/assets/dictionary.txt` I believe.

src/client/firefox.js
```diff
@@ -7,12 +7,20 @@
 import { setupCommands, clientCommands } from "./firefox/commands";
 import { setupEvents, clientEvents } from "./firefox/events";
 import { features } from "../utils/prefs";
+import type { Grip } from "debugger-html";
+let DebuggerClient;
+
+function createObjectClient(grip: Grip) {
+  return DebuggerClient.createObjectClient(grip);
```
wldcordeiro: Mochitest is failing here because the assignment in the `onConnect` below hadn't happened yet. Should you put a guard in to check if the client isn't undefined?

nchevobbe: I think it was failing because of https://bugzilla.mozilla.org/show_bug.cgi?id=1430799, which now landed in mozilla-central.
Locally, all the mochitest pass with `--verify` :)

nchevobbe:
> @nchevobbe you need to add setPopupObjectProperties to /assets/dictionary.txt I believe.

Yes ! Thanks, I fixed it in the PR (and took this opportunity to sort the dictionary as well).

src/client/firefox.js
wldcordeiro: Ah good to know. I saw the failure complaining about undefined and figured it was some async funkiness. Guess we just need to update the CI image?

nchevobbe: i'm not familiar with this, but if this mean the CI image needs to be updated to pick up mozilla-central changes, then yes :)

jasonLaster: @nchevobbe you can update the `.travis.yml` w/ the new mc scha

src/test/mochitest/browser_dbg-expressions.js
```diff
@@ -32,7 +32,13 @@ function assertEmptyValue(dbg, index) {
...
 
 function toggleExpression(dbg, index) {
-  findElement(dbg, "expressionNode", index).click();
+  const node = findElement(dbg, "expressionNode", index);
+  const objectInspector = node.closest(".object-inspector");
+  const properties = objectInspector.querySelectorAll(".node").length;
+  node.click();
+  return waitUntil(
+    () => objectInspector.querySelectorAll(".node").length !== properties
+  );
```
jasonLaster: shouldn't we wait till it is the same?

can we move this to a helper?

src/test/mochitest/browser_dbg-minified.js
```diff
@@ -3,10 +3,6 @@
 
 // Tests minfied + source maps.
 
-async function foo() {
-  return new Promise(r => setTimeout(r, 3000000));
-}
-
 function getScopeNodeLabel(dbg, index) {
```
jasonLaster: lol

src/test/mochitest/head.js
```diff
@@ -388,13 +388,12 @@ function isPaused(dbg) {
...
-async function waitForLoadedObjects(dbg) {
-  const { hasLoadingObjects } = dbg.selectors;
-  return waitForState(
-    dbg,
-    state => !hasLoadingObjects(state),
-    "loaded objects"
-  );
+async function waitForLoadedScopes(dbg) {
+  await waitUntil(() => findElement(dbg, "scopes"));
+  const scopes = findElement(dbg, "scopes");
```
jasonLaster: this can be `const scopes = await waitUntil(() => findElement(dbg, "scopes"));`

src/test/mochitest/browser_dbg-expressions-error.js
```diff
@@ -21,18 +21,14 @@ function getValue(dbg, index) {
...
 function toggleExpression(dbg, index) {
-  findElement(dbg, "expressionNode", index).click();
+  const node = findElement(dbg, "expressionNode", index);
+  const objectInspector = node.closest(".object-inspector");
+  const properties = objectInspector.querySelectorAll(".node").length;
+  node.click();
+  return waitUntil(
+    () => objectInspector.querySelectorAll(".node").length !== properties
+  );
 }
```
jasonLaster: Perhaps we can have helpers like this in `mochitest/head.js`?

```js
async function toggleExpression() {
  const objectInspector = findElement(dbg, "expressionNode", index).closest(".object-inspector");
  await toggleObjectInspectorNode(dbg, objectInspector);
}

async function toggleObjectInspectorNode(dbg, objectInspector) {
   const getNumProperties = () => objectInspector.querySelectorAll(".node").length;
   const numProperties = getNumProperties();
   node.click();
   return waitUntil(() => numProperties !== getNumProperties());
}
```

src/test/mochitest/browser_dbg-scopes.js
```diff
@@ -2,7 +2,13 @@
...
 
 function toggleNode(dbg, index) {
-  clickElement(dbg, "scopeNode", index);
+  const node = findElement(dbg, "scopeNode", index);
+  const objectInspector = node.closest(".object-inspector");
+  const properties = objectInspector.querySelectorAll(".node").length;
+  node.click();
+  return waitUntil(
+    () => objectInspector.querySelectorAll(".node").length !== properties
+  );
```
jasonLaster:
```js
function toggleNode(dbg, index) {
  const objectInspector = findElement(dbg, "scopeNode", index).closest(".object-inspector");
  return toggleObjectInspectorNode(dbg, objectInspector);
}
```

src/components/Editor/Preview/Popup.js
```diff
@@ -44,24 +53,37 @@ export class Popup extends Component<Props> {
...
+      value &&
+      value.actor &&
+      !popupObjectProperties[value.actor]
+    ) {
+      const onLoadItemProperties = loadItemProperties(root, createObjectClient);
+      if (onLoadItemProperties !== null) {
+        const properties = await onLoadItemProperties;
+        setPopupObjectProperties(value, properties);
+      }
+    }
```
jasonLaster: this looks like it could be moved to a util in the debugger or reps project.

I'm thinking this because it pulls in a bunch of dependencies and is pretty complicated...

could we move some of this logic to the `actions/pause/setupPopup...`

src/test/mochitest/browser_dbg-expressions.js
nchevobbe: nope, since we expand a node, the result should be that the objectInspector has a different number of nodes

src/test/mochitest/head.js
nchevobbe: mh, I'll try, but I looked up the waitUntil function, and it was resolving with true, not the result of the condition.
Though, I did not try what you suggest here, so I will

src/test/mochitest/browser_dbg-expressions-error.js
nchevobbe: yeah, sounds like a good idea

src/test/mochitest/head.js
nchevobbe: this makes test fail because `scopes.querySelector` is not a function.

This is explained by https://searchfox.org/mozilla-central/rev/2031c0f517185b2bd0e8f6f92f9491b3410c1f7f/devtools/client/framework/test/shared-head.js#453-462 , where waitUntil resolves with `true`.

jasonLaster: oh, i think we have `waitForElement` :)

src/test/mochitest/browser_dbg-expressions-error.js
nchevobbe: ✅

src/test/mochitest/browser_dbg-scopes.js
nchevobbe: ✅

PR #4918, opened 2017-12-14
Add longer timeout for browser_dbg-iframes.js

juliandescottes:
Add longer timeout x3 for browser_dbg-iframes.js


juliandescottes:
Let's close this, looks like the current release is not getting backed out because of the iframes test.

jasonLaster:
Should we create an issue for iframes being intermittent

Issue #4864, opened 2017-12-06
Right-click callstack entry causes a 800ms hang by updating codemirror


digitarald:
STR:
- Pause in a large file
- Right-click callstack

AR:
- 1s to show the menu; 800ms hang

Profile: https://perfht.ml/2itMAfk

I see code mirror and other things in here that should not really render when right-clicking the callstack.

PR #4921, opened 2017-12-14
Do not select the frame if the frame is already selected

bomsy:
Associated Issue: #4864

### Summary of Changes

* Do not select the frame if its already selected

### Test Plan

Example test plan:

- [x] Go to https://www.google.com/
- [x] Open the debugger
- [x] select `(index)` 
- [x] Set a breakpoint on line 4
- [x] Refresh the web page
- [x] Right-click in the callstack
- [x] The context menu should not delay showing up

### DIFF
--- a/src/components/SecondaryPanes/Frames/Frame.js
+++ b/src/components/SecondaryPanes/Frames/Frame.js
@@ -87,10 +87,7 @@ export default class FrameComponent extends Component<FrameComponentProps> {
     frame: Frame,
     selectedFrame: Frame
   ) {
-    if (
-      (e.nativeEvent.which == 3 && selectedFrame.id != frame.id) ||
-      selectedFrame.id === frame.id
-    ) {
+    if (e.nativeEvent.which == 3 || selectedFrame.id === frame.id) {
       return;
     }
     this.props.selectFrame(frame);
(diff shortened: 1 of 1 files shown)


darkwing:
@bomsy Do you know of a URL/source which usually takes a long time so I can test a real case?

bomsy:
@darkwing https://www.google.com/ has some massive minified files.

src/components/SecondaryPanes/Frames/Frame.js
```diff
@@ -87,7 +87,10 @@ export default class FrameComponent extends Component<FrameComponentProps> {
     frame: Frame,
     selectedFrame: Frame
   ) {
-    if (e.nativeEvent.which == 3 && selectedFrame.id != frame.id) {
+    if (
+      (e.nativeEvent.which == 3 && selectedFrame.id != frame.id) ||
+      selectedFrame.id === frame.id
+    ) {
```
jasonLaster: i feel like we might want: `e.nativeEvent.which == 3 || selectedFrame.id === frame.id`

bomsy: true

jasonLaster:
try rebasing to get the fix for the intermittent

jasonLaster:
Perfect fix! I love it when it comes together this nicely

Issue #4906, opened 2017-12-13
[CommandBar] buttons are mis-aligned


jasonLaster:
It looks like the command bar buttons recently shifted a bit up and are no longer aligned in the center with the toggle button.

<img src='https://shipusercontent.com/d6a26e69bbbab527b023935a36814481/Screen%20Shot%202017-12-13%20at%2011.07.18%20AM.png' title='Screen Shot 2017-12-13 at 11.07.18 AM.png' width=376>

ryanjduffy:
Looks like the toggle button is (loosely?) centered with the tabs whereas the command bar buttons are (loosely?) centered in the bar. To which should they align?

I'd like to do a bit of refactoring to standardize these components to ensure some consistency across their usage.

lukaszsobek:
@ryanjduffy if you remove this class
`.command-bar > button {
   padding: 6px 5px;
}`
they will be perfectly aligned again. Alternatively you can also add it to:
`.toggle-button-start.collapsed, .toggle-button-end.collapsed`
which will have a similar effect

ryanjduffy:
@lukaszsobek Thanks! I saw the same. While that does produce the right result, I think we can improve the layout. After removing that rule, the button has an 8px top/bottom margin leaving 13px for height while the contained image is actually 16px. So we're overflowing the button's content box and effectively treating it like it has a top padding of 8px and bottom padding of 5px.

However, none of the padding gymnastics is really even necessary because both the editor pane and the command bar is using flex layout. I refactored the shared button component out and reworked the CSS to simplify it and rely on the flex layout. I also adjusted the resume SVG which was offset to the right of the viewbox. Here's gif of my change compared to just removing the padding rule.

![4809-diff](https://user-images.githubusercontent.com/788456/34007599-341743f4-e0b7-11e7-8b58-fe036714a324.gif)

wldcordeiro:
Looks like this is closed! Thanks @ryanjduffy

PR #4922, opened 2017-12-14
Refactor common button code into CommandBarButton

ryanjduffy:
Associated Issue: #4906

### Summary of Changes

* Refactored common code into `CommandBarButton` component and used in `PaneToggle` and command bar
* Cleaned up CSS to resolve alignment problem

image

### DIFF
--- a/src/components/Editor/Tabs.css
+++ b/src/components/Editor/Tabs.css
@@ -7,7 +7,6 @@
   width: 100%;
   height: 30px;
   display: flex;
-  align-items: flex-end;
 }
 
 .source-header * {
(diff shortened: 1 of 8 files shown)


jasonLaster:
Nice work @ryanjduffy!

src/components/Editor/Tabs.css
```diff
@@ -7,7 +7,6 @@
   width: 100%;
   height: 30px;
   display: flex;
-  align-items: flex-end;
```
wldcordeiro: This seems out of place?

nyrosmith:
Looks good to me

src/components/Editor/Tabs.css
ryanjduffy: It pushes the toggle buttons to the bottom rather than the middle. I had assumed that was necessary to align the tabs with the bottom but they didn't move when this was removed.

darkwing:
Excellent work @ryanjduffy !!!  Thank you!

Issue #3942, opened 2017-09-08
[Editor] Add a menu item to execute selected text in the console


captainbrosset:
Filing from some feedback received today on IRC #devtools:

"Selecting text in debugger to evaluate in console.  Couldn't figure out how to do that, but seems like a super easy convenience function.
Just an extra right click entry that copies it over and runs it."

jasonLaster:
Yes, that's a great win. Chrome also uses the selection to show a preview. They started doing this 8 months or so ago.

lukaszsobek:
Seems we need something along the lines of:
- add new text entry to the localization file "Evaluate in console"
- add this entry to a new menu item
- the menu item only appears when we right-click a selection of text
- the new menu item calls a function
- the function console logs the output

jasonLaster:
I think we can add a `consoleEvaluate` action in `actions/panel` 

```js
export function consoleEvaluate(input) {
    return ({ client }: ThunkArgs) => {
      // check to see if there is a syntax error ... 
      
      // then evaluate
      client.evaluate(`console.log(input)`)
    }
}
```

lukaszsobek:
@jasonLaster @captainbrosset 
So when we run evaluateInConsole we seem to have 3 scenarios:

- we are evaluating a variable, in which case we want to return:
a) which variable was evaluated
b) what the variable contained or undefined

- we are evaluating a function content, in which case we want to return:
a) which function was evaluated
b) what the contents of the functions are

- we are evaluating a function execution, in which case we want to return:
a) which function was being executed
b) what the contents of the functions are
c) what the return value of the function is

Does that sound about right?

lukaszsobek:
Merged. Closing.

PR #4934, opened 2017-12-16
Adds option to execute selected text in editor right-click menu

lukaszsobek:
Associated Issue: #3942

### Changes
* introduces a context menu item to evaluate code in the console.

### DIFF
--- a/src/actions/toolbox.js
+++ b/src/actions/toolbox.js
@@ -5,6 +5,7 @@
 // @flow
 
 const { isDevelopment } = require("devtools-config");
+const { getWorker, getSelectedFrameId } = require("../selectors");
 
 import type { ThunkArgs } from "./types";
 import type { Worker } from "../types";
@@ -34,5 +35,15 @@ export function openWorkerToolbox(worker: Worker) {
     } else {
       openWorkerToolboxCommand(worker);
     }
+  };
+}
+
+export function evaluateInConsole(inputString: string) {
+  return async ({ client, getState }: ThunkArgs) => {
+    const frameId = getSelectedFrameId(getState());
+    client.evaluate(
+      `console.log("${inputString}"); console.log(${inputString})`,
+      { frameId }
+    );
   };
 }
--- a/src/actions/toolbox.js
+++ b/src/actions/toolbox.js
@@ -5,7 +5,7 @@
 // @flow
 
 const { isDevelopment } = require("devtools-config");
-const { getWorker, getSelectedFrameId } = require("../selectors");
+const { getSelectedFrameId } = require("../selectors");
 
 import type { ThunkArgs } from "./types";
 import type { Worker } from "../types";
--- a/src/components/Editor/EditorMenu.js
+++ b/src/components/Editor/EditorMenu.js
@@ -34,6 +34,7 @@ function getMenuItems(
   {
     addExpression,
     editor,
+    evaluateInConsole,
     flashLineRange,
     getFunctionLocation,
     getFunctionText,
@@ -62,7 +63,7 @@ function getMenuItems(
     selectedLocation,
     event
   );
-  const textSelected = editor.codeMirror.somethingSelected();
+  const isTextSelected = editor.codeMirror.somethingSelected();
 
   // localizations
   const blackboxKey = L10N.getStr("sourceFooter.blackbox.accesskey");
@@ -77,6 +78,7 @@ function getMenuItems(
   const copySourceLabel = L10N.getStr("copySource");
   const copySourceUri2Key = L10N.getStr("copySourceUri2.accesskey");
   const copySourceUri2Label = L10N.getStr("copySourceUri2");
+  const evaluateInConsoleLabel = L10N.getStr("evaluateInConsole.label");
   const jumpToMappedLocKey = L10N.getStr(
     "editor.jumpToMappedLocation1.accesskey"
   );
@@ -157,6 +159,12 @@ function getMenuItems(
     click: () => addExpression(editor.codeMirror.getSelection())
   };
 
+  const evaluateInConsoleItem = {
+    id: "node-menu-evaluate-in-console",
+    label: evaluateInConsoleLabel,
+    click: () => evaluateInConsole(selectionText)
+  };
+
   // construct menu
   const menuItems = [
     copySourceItem,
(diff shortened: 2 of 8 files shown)


src/actions/execute-in-console.js
```diff
@@ -0,0 +1,9 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
+
+// @flow
+
+export function executeInConsole(string: String) {
+  console.log(string);
```
bomsy: this is not finished, right?

src/actions/index.js
```diff
@@ -4,40 +4,40 @@
...
+  ...pause,
+  ...preview,
   ...projectTextSearch,
   ...quickOpen,
+  ...sources,
   ...sourceTree,
-  ...loadSourceText,
-  ...debuggee,
   ...toolbox,
-  ...preview
```
jasonLaster: lets merge this clean up in it's own PR

src/components/Editor/EditorMenu.js
```diff
@@ -151,6 +152,18 @@ function getMenuItems(
     click: () => addExpression(editor.codeMirror.getSelection())
   };
 
+  const executeInConsoleItem = {
+    id: "node-menu-execute-in-console",
+    label: executeInConsoleLabel,
+    click: () => {
+      new Promise((res, rej) => {
+        res(eval(selectionText));
```
jasonLaster: we probably want to the action `evaluateExpression("console.log(${selectionText})")` so that we have the server try to evaluate the expression `console.log(foo)` or whatever

src/actions/index.js
lukaszsobek: Done. :)

jasonLaster:
```js
import { actions, selectors, createStore } from "../../utils/test-head";

describe("toolbox", () => {
  describe("evaluate", () => {
    it("simple", () => {
      const threadClient = {
        evaluate: jest.fn()
      };
      const { dispatch, getState } = createStore(threadClient);
      dispatch(actions.executeInConsole("foo"));
      expect(threadClient.evaluate).toHaveBeenCalledWith("foo");
    });
  });
});
```

src/actions/toolbox.js
```diff
@@ -37,3 +37,19 @@ export function openWorkerToolbox(url: string) {
...
   };
 }
+
+export function evaluateInConsole(inputString: string) {
+  return async ({ client, getState }: ThunkArgs) => {
+    const frameId = getSelectedFrameId(getState());
+
+    console.log(inputString);
+
+    client.evaluate(`${inputString}.toString()`, { frameId }).then(res => {
```
jasonLaster: the `toString` worries me :)

src/actions/toolbox.js
```diff
@@ -37,3 +37,19 @@ export function openWorkerToolbox(url: string) {
     }
   };
 }
+
+export function evaluateInConsole(inputString: string) {
+  return async ({ client, getState }: ThunkArgs) => {
+    const frameId = getSelectedFrameId(getState());
+
+    console.log(inputString);
```
jasonLaster: oops

src/actions/toolbox.js
```diff
@@ -37,3 +37,19 @@ export function openWorkerToolbox(url: string) {
...
+
+    client.evaluate(`${inputString}.toString()`, { frameId }).then(res => {
+      if (res.exception) {
+        const docLink = res.exceptionDocURL ? `\n${res.exceptionDocURL}` : "";
+        return console.error(res.exceptionMessage, docLink);
```
jasonLaster: what's this?

src/actions/toolbox.js
lukaszsobek: This one is in order to reassure the person evaluating what they are evaluating before it is being evaluated.

src/actions/toolbox.js
lukaszsobek: Not sure how to do it differently if we have e.g. a function and we expect the body of it to be displayed.

src/actions/toolbox.js
lukaszsobek: It's a try to add context to an error message if it is provided via exceptionDocURL.

jasonLaster: i see. we probably need to evaluate `client.evaluate("console.error(${res.exceptionMessage}, ${docLink})")` or something like that

src/actions/toolbox.js
```diff
@@ -37,3 +37,19 @@ export function openWorkerToolbox(url: string) {
...
+      }
+      return console.log(res.result);
```
jasonLaster: i dont think we want the console.log here...

src/actions/toolbox.js
jasonLaster: we should drop this

src/actions/toolbox.js
jasonLaster: I think we should get rid of toString

src/actions/toolbox.js
lukaszsobek: This is to reassure the user that the correct piece of code has been executed. Chrome does it in a similar fashion.

src/actions/toolbox.js
lukaszsobek: Not quite sure how to catch execution output (we evaluate function return with `()`), function body text (it is a function without the `()`) or variable content at the same time a different way.

src/actions/toolbox.js
lukaszsobek: I think I see two possible scenarios here, not sure which is being addressed `I dont think we want the _console.log_ here` or `I dont think we want the console.log _here_`

jasonLaster:
so... right click is a bit weird

lukaszsobek:
XD - split personality - it wants to do two things at once maybe...

jasonLaster:
@lukaszsobek we can fix the right click on hover in a separate issue. The STR is just select text, hover, and right click... it's weird

jasonLaster:
rebased and re-running tests. it should pass :)

jasonLaster:
I updated the tests. Other than that though. it looks great

lets focus on hover / right click behavior when there are selections

PR #4943, opened 2017-12-18
[Editor] detect if files use JSX

DavideDeCenzo:
Associated Issue: #4694 

Here's the Pull Request Doc

Feel free to improve it!

### DIFF
--- a/src/actions/README.md
+++ b/src/actions/README.md
@@ -1,0 +1,26 @@
+## Actions
+
+### Best Practices
+
+#### Scheduling Async Actions
+
+There are several use-cases with async actions that involve scheduling:
+
+* we do one action and cancel subsequent actions
+* we do one action and subsequent calls wait on the initial call
+* we start an action and show a loading state
+
+If you want to wait on subsequent calls you need to store action promises.
+[ex][req]
+
+If you just want to cancel subsequent calls, you can keep track of a pending
+state in the store. [ex][state]
+
+The advantage of adding the pending state to the store is that we can use that
+in the UI:
+
+* disable/hide the pretty print button
+* show a progress ui
+
+[req]: https://github.com/devtools-html/debugger.html/blob/master/src/actions/sources/loadSourceText.js
+[state]: https://github.com/devtools-html/debugger.html/blob/master/src/reducers/sources.js
--- a/src/actions/sources/createPrettySource.js
+++ b/src/actions/sources/createPrettySource.js
@@ -2,9 +2,8 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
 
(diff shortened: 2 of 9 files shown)


src/components/Editor/index.js
```diff
@@ -500,10 +502,13 @@ class Editor extends PureComponent<Props, State> {
     }
 
     if (selectedSource) {
+      const hasJsx = getSymbols(selectedSource);
```
bomsy: Does this mean that, if symbols exist for the selected source, then it has JSX? Not sure i understand this...

DavideDeCenzo: My bad you're right I got confused with another thing Thanks for the review!

jasonLaster:
Hey @DavideDeCenzo this is getting closer and I bet this version works, which is cool.

Couple of architectural comments:
1) we shouldn't call `getSymbols` from the component, because calling it can be slow
2) we should import `getSymbols` from the `selectors` file and call it in connect.

```diff
diff --git a/src/components/Editor/index.js b/src/components/Editor/index.js
index aa3706e..4ce645d 100644
--- a/src/components/Editor/index.js
+++ b/src/components/Editor/index.js
@@ -22,7 +22,8 @@ import {
   getHitCountForSource,
   getCoverageEnabled,
   getSourceMetaData,
-  getConditionalPanelLine
+  getConditionalPanelLine,
+  getSymbols
 } from "../../selectors";
 
 // Redux actions
@@ -627,6 +628,7 @@ const mapStateToProps = state => {
   return {
     selectedLocation: getSelectedLocation(state),
     selectedSource,
+    symbols: getSymbols(state),
     searchOn: getActiveSearch(state) === "file",
     hitCount: getHitCountForSource(state, sourceId),
     selectedFrame: getSelectedFrame(state),
```

The difference is that here we get the symbols from the redux store and don't have to ask the worker for them. Sorry this is complicated, might help to watch some egghead videos :)

DavideDeCenzo:
Ok I'll try to figure out these aspects , Thanks for the advices @jasonLaster

jasonLaster:
This looks perfect. Will review carefully soon though!

src/utils/source.js
```diff
@@ -235,7 +240,7 @@ function getMode(source: Source, sourceMetaData: SourceMetaDataType) {
 
   if (
     (url && url.match(/\.jsx$/i)) ||
-    (sourceMetaData && sourceMetaData.isReactComponent)
+    (sourceMetaData && (sourceMetaData.isReactComponent || symbols.hasJsx))
```
jasonLaster: lets simplify this:

`sourceMetaData && (sourceMetaData.isReactComponent || symbols.hasJsx)` => `symbols.hasJsx`

DavideDeCenzo: I've got "invalid left-hand side in arrow function parameters" and some token errors. I've tried to fix that but nothing

PR #4749, opened 2017-11-20
Throttle new sources

jasonLaster:
Associated Issue: #4705 

### Summary of Changes

This throttles the firefox newSource event so that we fire at most one newSources action every 50ms. 

* Not sure if we need a test
* Not sure if we want to debounce in the event or the actions. Events seem easier now

The performance savings is significant ⏲ 

#### No throttling, w/o the editor fix

~ 9 seconds. Lots of thrashing here :P

screen shot 2017-11-20 at 11 44 04 am

#### Throttling w/o the editor fix

3 seconds. The editor update holds up the main thread so we also do more newSources calls :) vicious cycle

screen shot 2017-11-20 at 11 47 12 am

#### Throttling w/ the editor fix

1.5 sec

screen shot 2017-11-20 at 11 48 40 am

PR #4947, opened 2017-12-18
Revert "Throttle new sources (#4749)"

jasonLaster:
Reverts the throttle optimization because it had a memory leak w/ try.

We're working on a fix now


jasonLaster:
landing so that master can be green. Yulia is working on a fix now

PR #4960, opened 2017-12-21
Remove unecessary bindActionCreator calls

jasonLaster:
### Summary of Changes

Quick refactor inspired by @Fischer-L's work cleaning up our connect calls.

### DIFF
--- a/src/components/Editor/GutterMenu.js
+++ b/src/components/Editor/GutterMenu.js
@@ -4,7 +4,6 @@
 
 import { PureComponent } from "react";
 import { showMenu } from "devtools-contextmenu";
-import { bindActionCreators } from "redux";
 import { connect } from "react-redux";
 import { lineAtHeight } from "../../utils/editor";
 import {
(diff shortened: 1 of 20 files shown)


jasonLaster:
Hmm, looks like this approach doesn't work without bindActionCreators... Maybe i'm missing something.

CC @wldcordeiro @Fischer-L

wldcordeiro:
@jasonLaster I think the object that `actions` is, isn't a normal JS object so Redux doesn't bind properly. 🤔

jasonLaster:
ooh interesting... it should just come the actions/index module. 

Worth investigating :8ball:

src/components/Editor/GutterMenu.js
```diff
@@ -154,19 +153,16 @@ class GutterContextMenuComponent extends PureComponent {
...
-      pauseData: getPause(state),
-      contextMenu: getContextMenu(state),
-      emptyLines: selectedSource
-        ? getEmptyLines(state, selectedSource.toJS())
-        : []
-    };
-  },
-  dispatch => bindActionCreators(actions, dispatch)
-)(GutterContextMenuComponent);
+export default connect(state => {
```
lukaszsobek: This seems very hard to read, maybe we just make a mapStateToProps and mapDispatchToProps function to handle the logic. 1.) It is the default way of handling these kinds of things and 2.) will make it easier to see what is going on and also 3.) eliminate mistakes due to nested bracketing.

jasonLaster: I'd like to figure out how to have `connect(mapStateToProps, actions)(GutterContextMenuComponent);` everywhere

`

jasonLaster:
closing for now

PR #4974, opened 2017-12-22
Panes

GarethSharpe:
### Summary of Changes

* Only renders scopes pane when in paused state

### Test Plan

- [x] Pause debugger and scopes/call-stack panes are shown
- [x] Resume debugger and scopes/call-stack panes are not rendered

### Known Issues

- Panes do not render open when pref is set

#### Cases
- [x] Scopes/call-stack panes are not rendered when respective accordion(s) are closed
- [x] Scopes/call-stack panes are not rendered when respective accordion(s) are open

toggle-scopes-done


jasonLaster:
rebase?

jasonLaster:
@GarethSharpe turns out that the panes were not being shown because of some faulty logic in the Accordion component.

When that was fixed it was easy to take out the togglePanes functions in the tests.

PR #4978, opened 2017-12-23
Cleanup the documentation 

bomsy:
### Summary of Changes

As discussed with @jasonLaster 
This is a first pass at fixing up the documentation landing page, this will make it easy to find documentation, and also so we do not duplicate information that might already exist.
 
Down the line, we should look into detailed solutions.

Now for any documentaion added, there should be a link added to this documentation linking there


jasonLaster:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=efb66a0598c1afa29d5acaa0e5a565bd8b207a5c

Issue #4863, opened 2017-12-06
[Sources] Show “Loading sources” label


nchevobbe:
Steps to reproduce:
1. Go to https://debugger-html-css-grid.glitch.me/
2. Open the debugger 
3. Reload

Expected results:
The source panel shows a "Loading sources" label
Actual results:
The sources panel shows "this page has no sources" label, and then is updated

![dec-06-2017 15-41-40](https://user-images.githubusercontent.com/578107/33667077-1cec1202-da9c-11e7-8a44-38f84ecf0961.gif)

In the gif case, it is a bit confusing, but fine. But in some cases you have a flicker of the sources panel because it quickly switches from "no sources" to the tree.

Having a "loading sources" which is displayed a decent amount of time so we don't have the flicker would be nice.

----

I don't know how feasible this is since sources can be added at anytime, but I guess we might have some initial source fetching we can rely on to handle this.

jasonLaster:
Closing this issue and moving to the [enhancements](http://github.com/devtools-html/debugger.html/projects/5) board. 

* We will hopefully be able to prioritize it in the near future. 
* The issue is closed temporarily, but feel free to continue discussing the issue

PR #4979, opened 2017-12-25
WIP - Adds loading message for source tree if sources are loading and no items are displayed yet

lukaszsobek:
Associated Issue: #4863

### Summary of Changes

* Makes SourceTree component aware of isSourceTreeLoaded component state
* Adds display message for loading sources + changes in css
* Orders props & state alphabetically

### Screenshots/Videos
1st reload = reload of debugger (no loading indicator since sources are loaded before component is mounted)
2nd reload = reload of target window (loading indicator as sources are being loaded while debugger is displayed)
loading-indicator

### DIFF
--- a/src/components/PrimaryPanes/SourcesTree.js
+++ b/src/components/PrimaryPanes/SourcesTree.js
@@ -48,24 +48,25 @@ import { features } from "../../utils/prefs";
 import { setProjectDirectoryRoot } from "../../actions/ui";
 
 type Props = {
-  selectLocation: Object => void,
-  setExpandedState: any => void,
-  sources: SourcesMap,
-  shownSource?: string,
-  selectedSource?: SourceRecord,
   debuggeeUrl: string,
+  expanded?: any,
   projectRoot: string,
-  expanded?: any
+  selectLocation: Object => void,
+  selectedSource?: SourceRecord,
+  setExpandedState: any => void,
+  shownSource?: string,
+  sources: SourcesMap
 };
 
 type State = {
   focusedItem?: any,
-  parentMap: any,
-  sourceTree: any,
-  projectRoot: string,
-  uncollapsedTree: any,
+  highlightItems?: any,
+  isSourceTreeLoaded: boolean,
   listItems?: any,
-  highlightItems?: any
+  parentMap: any,
+  projectRoot: string,
+  sourceTree: any,
+  uncollapsedTree: any
 };
 
 class SourcesTree extends Component<Props, State> {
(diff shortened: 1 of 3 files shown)


src/components/PrimaryPanes/SourcesTree.js
```diff
@@ -48,24 +48,25 @@ import { features } from "../../utils/prefs";
...
-  debuggeeUrl: string,
-  projectRoot: string,
-  expanded?: any
+  sources: SourcesMap
 };
 
 type State = {
   focusedItem?: any,
+  highlightItems?: any,
+  isSourceTreeLoaded: boolean,
```
lukaszsobek: Seems better to use `treeProps.key == empty` for checking if there is anything in the tree instead of introducing a new property in state.

jasonLaster:
this is a tough one :)

My intuition is to save some state in redux like: we're navigating or loading...

Here are some interesting locations in the code:

https://gist.github.com/05d4ccdca7a917aebf160eb2d9369ba4

lukaszsobek:
@jasonLaster Thinking about it I was more caught by the display side of things ... do we make the "Loading..." go away immediately after at least one source loads (like it is now), or do we wait for the full duration of the loading process and only then make the message go away. The first might cause things to flash quickly, the other might make the interface feel sluggish. On the other hand if we decide to display the message for 2 seconds it feels like we have to buffer the results and return them after the time has passed totally ignoring the render cycle.

ps. Just noticed we could have both ... display a loading tag,  spinning circle or loading cursor while the content is being loaded. This way the newly arriving content would not replace the loading message, but they would exist next to each other to inform the user what is happening and how much progress has been made. Something along the lines of:

On example of a spinning circle:
1. User reloads
2. Only spinning circle displays
3. Content is slowly being loaded while circle continues to spin. the source tree changes while sources are being added.
4. All sources have been loaded. Spinning circle disappears to indicate we are done.

Rationale:
1.) The interface does not flash very abruptly
2.) the user can clearly differentiate between loading and loaded without having to read
3.) Progress indication and progress information are on distinctly separate display layers
4.) the difference layers operate independently from each other without conflicting with each other.

jasonLaster:
that sounds good. I think we want to improve our loading UX. Atom has a nice spinner at the bottom right that often shows that it's working.

lukaszsobek:
@jasonLaster how does this look?
loading-circle

jasonLaster:
@lukaszsobek I'm not sure what the best way to model this is. Also, I think i'd prefer a spinner for the debugger... kinda like how atom and other IDEs have one spinner.

Going to close this for now as it's not a priority.

lukaszsobek:
@jasonLaster Sure, was interesting working on it though.

PR #4983, opened 2017-12-27
⬆️: Upgrading Flow to 0.61

wldcordeiro:
Associated Issue: N/A

### Summary of Changes

* Upgrade Flow to 0.61.0 

### Test Plan

* Ran test suite
* Ran through app scenarios (stepping, breakpoints, etc to ensure no odd regressions)

### DIFF
--- a/src/reducers/pending-breakpoints.js
+++ b/src/reducers/pending-breakpoints.js
@@ -80,7 +80,6 @@ function addBreakpoint(state, action) {
   if (action.status !== "done") {
     return state;
   }
-
   // when the action completes, we can commit the breakpoint
   const { value: { breakpoint } } = action;
   const locationId = makePendingLocationId(breakpoint.location);
--- a/package.json
+++ b/package.json
@@ -126,7 +126,7 @@
     "eslint-plugin-prettier": "^2.3.1",
     "eslint-plugin-react": "^7.2.1",
     "expect.js": "^0.3.1",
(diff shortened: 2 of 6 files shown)


src/reducers/pending-breakpoints.js
```diff
@@ -81,12 +81,16 @@ function addBreakpoint(state, action) {
     return state;
   }
 
-  // when the action completes, we can commit the breakpoint
-  const { value: { breakpoint } } = action;
-  const locationId = makePendingLocationId(breakpoint.location);
-  const pendingBreakpoint = createPendingBreakpoint(breakpoint);
+  if (action.type === "ADD_BREAKPOINT") {
```
wldcordeiro: Not sure about this change yet. Somewhat related to the remaining error. Just involves some type refinements in certain places.

jasonLaster:
Nice!

src/reducers/pending-breakpoints.js
wldcordeiro: I noticed this function is called for both add and enable breakpoints but I don't see the `value` property on the enable breakpoint action @jasonLaster Could you help with a little clarification?

wldcordeiro:
`flow-bin` 0.62 seems to be a little delayed so I'm removing the WIP and making this just the upgrade to 0.61 and will follow-up with the next one.

Issue #4560, opened 2017-11-02
[QuickOpen] `?` should show options


jasonLaster:
VS Code and Chrome do a great job of showing possible commands with `?`

We should do the same thing. 

<img src='https://shipusercontent.com/ceb5b7d6c605f16aacf7a6a78c2fe651/Screen%20Shot%202017-11-02%20at%2012.21.21%20PM.png' title='Screen Shot 2017-11-02 at 12.21.21 PM.png' width=731>

---

<img src='https://shipusercontent.com/52a47f22149ce874bf258ca900af6364/Screen%20Shot%202017-11-02%20at%2012.22.12%20PM.png' title='Screen Shot 2017-11-02 at 12.22.12 PM.png' width=642>

pradeepgangwar:
@jasonLaster I have not done something big change to this code repository. So although i may be knowing little about fixing it. But I would like to try it and learn something. Please guide me a little to start. And i will try it if i can fix it.

jasonLaster:
Great, @wldcordeiro will mentor this issue, but I would start by working on QuickOpen showing a list of buttons 

- jump to go to line,
- jump to sources

these actions are all in redux :)

pradeepgangwar:
@jasonLaster If i am right, main task is to open a drop down suggestion list when user types **?**

wldcordeiro:
@pradeepgangwar I'll update this with a patch later today for you to base off but the overview goes like this
1. User opens quick open modal (cmd+p)
2. User enters `?`
3. A list shows up that show our options
  * `@ for functions`
  * `# for variables`
  * `: for go to line`
  * `term: search for term and go to line`
4. User arrows down to option and hits enter
5. Quick open updates its query to use the option chosen.

This list might not be what we actually want listed but there's the basis.

nyrosmith:
The PR got merged so I'll close this

PR #5000, opened 2017-12-30
Quickopen '?' shows different options.

pradeepgangwar:
Associated Issue: #4560 

* Typing '?' in quickopen shows different options.
* Currently some code has been hard coded
* I need suggestions to improve code and working .

### Screenshots/Videos 
debugger


src/utils/quick-open.js
```diff
@@ -96,6 +104,31 @@ export function formatSymbols(
   };
 }
 
+export function formatGeneralQuery(): Array<FormattedGeneralQuery> {
+  const queryResult = [
+    {
+      value: "Seach for a function in a file",
+      title: "@ Function Search",
```
jasonLaster: we should use l10n here
https://github.com/devtools-html/debugger.html/blob/master/docs/local-development.md#l10n

src/utils/quick-open.js
```diff
@@ -24,7 +24,8 @@ export function parseQuickOpenQuery(query: string): QuickOpenType {
     const modifiers = {
       "@": "functions",
       "#": "variables",
-      ":": "goto"
+      ":": "goto",
+      "?": "query"
```
jasonLaster: wow! that's really cool

src/components/QuickOpenModal.js
```diff
@@ -121,33 +127,59 @@ export class QuickOpenModal extends Component<Props, State> {
...
     }
+    if (this.isGeneralQuery()) {
+      return this.generalQuery(query);
+    }
     return this.searchSources(query);
   };
 
+  updateQuery = (item: FormattedGeneralQuery) => {
+    const { setQuickOpenQuery } = this.props;
+    if (item.id == "id.123") {
```
jasonLaster: lol - perhaps @wldcordeiro has a better idea for the ids. I'm inclined to use a shortened name like gotoLine or searchFunctions

wldcordeiro: I'm gonna agree the IDs should be stuff like `gotoLine`, `searchFunctions`, `searchVariables`, etc. Maybe just reuse the values from the `modifiers` map in the `quick-open.js` util file?

src/utils/quick-open.js
wldcordeiro: Seems like this is also the source of those `id.123` values too.

src/components/QuickOpenModal.js
```diff
@@ -121,33 +127,59 @@ export class QuickOpenModal extends Component<Props, State> {
     this.setState({ results });
   };
 
+  generalQuery = (query: string) => {
```
wldcordeiro: Note: I don't know if the name `generalQuery` makes sense. It took me a while to figure out that you meant the `?` as the "general" query. I would have thought of the default (sources) as the "general" in this context. I'd maybe go with a name like `shortcutQuery` or `hintQuery` to more accurately convey what details are shown.

src/utils/quick-open.js
```diff
@@ -96,6 +104,31 @@ export function formatSymbols(
   };
 }
 
+export function formatGeneralQuery(): Array<FormattedGeneralQuery> {
+  const queryResult = [
```
wldcordeiro: If this constant isn't modified it can just be returned.

src/components/QuickOpenModal.js
```diff
@@ -121,33 +127,59 @@ export class QuickOpenModal extends Component<Props, State> {
...
+      setQuickOpenQuery("#");
+    } else {
+      setQuickOpenQuery(":");
+    }
+  };
+
   selectResultItem = (
     e: SyntheticEvent<HTMLElement>,
-    item: ?FormattedSource | ?FormattedSymbolDeclaration
+    item:
```
wldcordeiro: Really thinking about it this type definition shouldn't be a union and should be a single type with optional properties.

src/components/QuickOpenModal.js
```diff
@@ -41,6 +43,7 @@ type Props = {
   sources: Array<Object>,
   selectedSource?: SourceRecord,
   query: string,
+  generalItems: Array<Object>,
```
wldcordeiro: I don't think this prop is ever used? Also wouldn't it be `FormattedGeneralQuery[]`?

wldcordeiro:
Looking great! There are some notes and we have CI failures currently. But I dig the look of the hints/shortcuts. 😄

pradeepgangwar:
@wldcordeiro Thanks! for reviewing and suggestions. I will surely improve this PR based on review. 👍

wldcordeiro:
I've gone through the PR and updated to cover the notes and fixed the Flow issues.

pradeepgangwar:
Thanks! @wldcordeiro for the help. I was busy and there were some test failures last time i tried. Let me know if it requires some more changes. 😄

src/utils/quick-open.js
```diff
@@ -96,7 +89,27 @@ export function formatSymbols(
   };
 }
 
-export function formatSources(sources: SourcesMap): Array<FormattedSource> {
+export function formatShortcutResults(): Array<QuickOpenResult> {
+  return [
+    {
+      value: "Seach for a function in a file",
+      title: "@ Function Search",
+      id: "functions"
```
jasonLaster: how about `id: @`

src/components/QuickOpenModal.js
```diff
@@ -121,41 +122,62 @@ export class QuickOpenModal extends Component<Props, State> {
...
     }
+    if (this.isShortcutQuery()) {
+      return this.showShortcuts(query);
+    }
     return this.searchSources(query);
   };
 
+  setModifier = (item: QuickOpenResult) => {
+    const { setQuickOpenQuery } = this.props;
+    if (item.id === QUICKOPEN_MODIFIERS["@"]) {
```
jasonLaster: how about `if(["@","#",":"].includes(item.id)) setModifiers(item.id)`

src/components/tests/QuickOpenModal.spec.js
```diff
@@ -71,6 +71,22 @@ describe("QuickOpenModal", () => {
...
+        searchType: "shortcuts",
+        symbols: {
+          functions: [],
+          variables: []
+        }
+      },
+      "mount"
+    );
+    expect(wrapper).toMatchSnapshot();
+  });
```
jasonLaster: :clap:

src/components/QuickOpenModal.js
wldcordeiro: Good call if, the IDs are changed to the modifiers it works great.

assets/panel/debugger.properties
```diff
@@ -677,14 +677,17 @@ functionSearchSeparatorLabel=←
 # text displayed when the user searches for specific lines in a file
 gotoLineModal.placeholder=Go to line…
 gotoLineModal.key=CmdOrCtrl+Shift+;
+gotoLineModal.title=Go to a line number in a file
```
wldcordeiro: Good call on the translations! 😄 It might be nice at some point to unify all the translations under `quickOpenModal` rather than split across `gotoLineModal`, `symbolSearch`, etc.

src/components/tests/QuickOpenModal.spec.js
```diff
@@ -71,6 +71,22 @@ describe("QuickOpenModal", () => {
...
+        query: "?",
+        searchType: "shortcuts",
+        symbols: {
+          functions: [],
+          variables: []
+        }
+      },
+      "mount"
+    );
+    expect(wrapper).toMatchSnapshot();
```
jasonLaster: maybe we can simplify this with `wrapper.find("ResultList")`

src/components/QuickOpenModal.js
```diff
@@ -121,41 +121,57 @@ export class QuickOpenModal extends Component<Props, State> {
...
   ) => {
     if (item == null) {
       return;
     }
     const { selectLocation, selectedSource, query, searchType } = this.props;
-    if (this.isSymbolSearch()) {
+    if (this.isShortcutQuery()) {
+      this.setModifier(item);
+      return;
+    } else if (searchType === "gotoSource") {
```
jasonLaster: it'd be nice to have an `isGotoQuery`

Issue #5003, opened 2017-12-30
[SourceTree] Refactor source display to remove WET code


lukaszsobek:
Currently SourceTree.js has a verbose implementation of code on line 180 that is virtually identical to the function of createTree.js. It would be nice to get these two things in line so that sourceTree actually uses createTree.js.

lukaszsobek:
Merged. Closing.

PR #5004, opened 2017-12-30
Refactors displaying source tree updates

lukaszsobek:
Associated Issue: #5003

### Workbook

- [x] move updating tree into `updateTree` function
- [x] minor visual cleanup + alphabetical ordering
- [x] update jest suite to try `updateTree`

### DIFF
--- a/src/actions/pause/fetchScopes.js
+++ b/src/actions/pause/fetchScopes.js
@@ -64,13 +64,13 @@ export function fetchScopes() {
       frame.generatedLocation.sourceId
     );
 
-    if (generatedSourceRecord.get("isWasm")) {
+    if (generatedSourceRecord && generatedSourceRecord.get("isWasm")) {
       return;
     }
 
     const sourceRecord = getSource(getState(), frame.location.sourceId);
 
-    if (sourceRecord.get("isPrettyPrinted")) {
+    if (sourceRecord && sourceRecord.get("isPrettyPrinted")) {
       return;
     }
 
--- a/src/actions/pause/fetchScopes.js
+++ b/src/actions/pause/fetchScopes.js
@@ -64,13 +64,13 @@ export function fetchScopes() {
       frame.generatedLocation.sourceId
     );
 
-    if (generatedSourceRecord && generatedSourceRecord.get("isWasm")) {
+    if (generatedSourceRecord.get("isWasm")) {
       return;
     }
 
     const sourceRecord = getSource(getState(), frame.location.sourceId);
 
-    if (sourceRecord && sourceRecord.get("isPrettyPrinted")) {
+    if (sourceRecord.get("isPrettyPrinted")) {
       return;
     }
 
--- a/src/components/PrimaryPanes/SourcesTree.js
+++ b/src/components/PrimaryPanes/SourcesTree.js
@@ -77,7 +77,12 @@ class SourcesTree extends Component<Props, State> {
 
   constructor(props) {
     super(props);
-    this.state = createTree(this.props);
+    const { projectRoot, debuggeeUrl, sources } = this.props;
+    this.state = createTree({
+      projectRoot,
+      debuggeeUrl,
+      sources
+    });
   }
 
   componentDidMount() {
@@ -110,6 +115,7 @@ class SourcesTree extends Component<Props, State> {
       nextProps.sources.size === 0
     ) {
       // early recreate tree because of changes
+      // to project root, debugee url or lack of sources
       return this.setState(
         createTree({
           projectRoot,
--- a/src/components/PrimaryPanes/SourcesTree.js
+++ b/src/components/PrimaryPanes/SourcesTree.js
@@ -58,7 +58,7 @@ type Props = {
 type State = {
   focusedItem?: any,
   parentMap: any,
-  sourceTree: Node,
+  sourceTree: Object,
   projectRoot: string,
   uncollapsedTree: any,
   listItems?: any,
--- a/src/components/PrimaryPanes/SourcesTree.js
+++ b/src/components/PrimaryPanes/SourcesTree.js
@@ -25,9 +25,6 @@ import { setExpandedState } from "../../actions/source-tree";
 import { selectLocation } from "../../actions/sources";
 import { setProjectDirectoryRoot } from "../../actions/ui";
 
-// Types
-import type { Props, State } from "../../utils/sources-tree/types";
-
 // Components
 import ManagedTree from "../shared/ManagedTree";
 import Svg from "../shared/Svg";
@@ -45,6 +42,29 @@ import { copyToTheClipboard } from "../../utils/clipboard";
 import { throttle } from "../../utils/utils";
 import { features } from "../../utils/prefs";
 
+import type { SourcesMap, SourceRecord } from "../../reducers/types";
+
+type Props = {
+  selectLocation: Object => void,
+  setExpandedState: any => void,
+  sources: SourcesMap,
+  shownSource?: string,
+  selectedSource?: SourceRecord,
+  debuggeeUrl: string,
+  projectRoot: string,
+  expanded?: any
+};
+
+type State = {
+  focusedItem?: any,
+  parentMap: any,
+  sourceTree: Node,
+  projectRoot: string,
+  uncollapsedTree: any,
+  listItems?: any,
+  highlightItems?: any
+};
+
 class SourcesTree extends Component<Props, State> {
   focusItem: Function;
   selectItem: Function;
@@ -57,23 +77,7 @@ class SourcesTree extends Component<Props, State> {
 
   constructor(props) {
     super(props);
-
     this.state = createTree(this.props);
-
-    this.focusItem = this.focusItem.bind(this);
-    this.selectItem = this.selectItem.bind(this);
-    this.getPath = this.getPath.bind(this);
-    this.getIcon = this.getIcon.bind(this);
-    this.onContextMenu = this.onContextMenu.bind(this);
-    this.renderItem = this.renderItem.bind(this);
-
-    this.queueUpdate = throttle(function() {
-      if (!this.mounted) {
-        return;
-      }
-
-      this.forceUpdate();
-    }, 50);
   }
 
   componentDidMount() {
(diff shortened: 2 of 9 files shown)


jasonLaster:
does this change any of the behavior or does it just consolidate it?

lukaszsobek:
Currently just consolidation intended, haven't looked into changes - do we need something?

jasonLaster:
Cool. This looks good.

src/components/PrimaryPanes/SourcesTree.js
```diff
@@ -157,52 +150,18 @@ class SourcesTree extends Component<Props, State> {
...
-
-    const uncollapsedTree = this.state.uncollapsedTree;
 
     // TODO: recreating the tree every time messes with the expanded
     // state of ManagedTree, because it depends on item instances
     // being the same. The result is that if a source is added at a
     // later time, all expanded state is lost.
-    let sourceTree = this.state.sourceTree;
-    if (newSet.size > 0) {
-      for (const source of newSet) {
```
jasonLaster: i think this was done for performance reasons because recreating the tree every time is expensive.

src/components/PrimaryPanes/SourcesTree.js
```diff
@@ -157,52 +150,18 @@ class SourcesTree extends Component<Props, State> {
...
     }
 
-    if (nextProps.sources === this.props.sources) {
-      return;
-    }
-
-    if (nextProps.sources.size === 0) {
-      // remove all sources
-      this.setState(
-        createTree(
```
jasonLaster: i agree it's verbose, but it's nice to see the params

src/components/PrimaryPanes/SourcesTree.js
```diff
@@ -116,17 +114,12 @@ class SourcesTree extends Component<Props, State> {
 
   componentWillReceiveProps(nextProps) {
     if (
```
jasonLaster: we might want to move all of this code into its own util and think through utility

src/components/PrimaryPanes/SourcesTree.js
lukaszsobek: Tried the behavior of the tree in various scenarios and it is recreating anyway - so really not quite sure if there are any cases where the code brings performance benefits that have not been covered by the checks going before it.

src/components/PrimaryPanes/SourcesTree.js
lukaszsobek: Went with the train of thought that for a quick look we only need to know that we create the tree based on the params. If we want to understand what the function does exactly we have to look into the file anyway. Inside the function itself we have a 1 line statement with all the constants we are going to use. So understanding the details of the work of the function felt easier and skimming the file inside which the function is used seemed easier as well.  

No problem changing it back though. Just felt cleaner that way.

src/components/PrimaryPanes/SourcesTree.js
lukaszsobek: Feels good ... SourcesTree seems pretty long with lots of things in it.

jasonLaster:
Looking good. Perhaps a unit test for `updateTree` that is similar to the `createTree` tests

jasonLaster:
Looking better!

src/utils/sources-tree/tests/updateTree.spec.js
```diff
@@ -0,0 +1,34 @@
...
+    };
+    props.sources = new Map();
+    props.sources.set("1", "str1");
+
+    // set nextProps
+    const nextProps = {
+      debuggeeUrl: "blah",
+      projectRoot: ""
+    };
+    nextProps.sources = new Map();
```
codehag: it looks like this isnt an immutable map, but a plain js Map https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map

This doesn't have a valueSeq method and is probably why the tests are failing!

you can fix this by importing immutable's "Map" and using that here

src/utils/sources-tree/types.js
```diff
@@ -16,3 +16,34 @@ export type Node = {
   path: string,
   contents: SourceRecord | Array<Node>
 };
+
+/**
+ * Props are exported as property type for this.props and NextProps
+ * @memberof utils/sources-tree
+ * @static
+ */
+export type Props = {
```
jasonLaster: I think these types should still live in the component file. We can export them if you wish so that they can be used in the util...

This is one of the reasons though that i think it's simpler to pass some specific params to the utils

src/utils/sources-tree/tests/updateTree.spec.js
```diff
@@ -0,0 +1,63 @@
...
+        name: "root",
+        path: ""
+      }
+    };
+
+    const returnvalue = updateTree(nextProps, props, state);
+    expect(
+      returnvalue.uncollapsedTree.contents[0].contents[0].contents.get("url")
+    ).toBe("https://davidwalsh.name/source1.js");
+  });
```
jasonLaster: :thumbsup:

src/utils/sources-tree/updateTree.js
```diff
@@ -0,0 +1,38 @@
...
+import type { Props, State } from "./types";
+
+import { addToTree } from "./addToTree";
+import { collapseTree } from "./collapseTree";
+import { createParentMap } from "./utils";
+
+function compareProps(nextProps: Props, props: Props) {
+  const next = Set(nextProps.sources.valueSeq());
+  const prev = Set(props.sources.valueSeq());
+  return next.subtract(prev);
```
jasonLaster: ooh - nice find w/ Immutable.Set.subtract

I think we can simplify this and save a bunch of CPU cycles :clock:

`return next.sources.subtract(prev.sources)`

lukaszsobek: `next` and `prev` are the sourceMaps already, so I don't think it goes any deeper.

src/utils/sources-tree/tests/updateTree.spec.js
lukaszsobek: @bomsy suggested a snapshot instead of this - seems like a good idea.

lukaszsobek:
`TEST-UNEXPECTED-FAIL browser_dbg-chrome-debugging.js:23 
 TypeError: DebuggerServer.addBrowserActors is not a function
     initDebuggerClient browser/devtools/client/debugger/new/test/mochitest/browser_dbg-chrome-debugging.js 23:5
      browser/devtools/client/debugger/new/test/mochitest/browser_dbg-chrome-debugging.js 72:13
     Async*Tester_execTest/< browser-test.js 1065:21
     Tester_execTest browser-test.js 1056:9
     Tester.prototype.nextTest</< browser-test.js 956:9
     SimpleTest.waitForFocus/waitForFocusInner/focusedOrLoaded/< tests/SimpleTest/SimpleTest.js 795:59`

lukaszsobek:
oops, some of the fixes got overwritten.

src/components/PrimaryPanes/SourcesTree.js
```diff
@@ -80,25 +77,7 @@ class SourcesTree extends Component<Props, State> {
...
-    this.renderItem = this.renderItem.bind(this);
-
-    this.queueUpdate = throttle(function() {
-      if (!this.mounted) {
-        return;
-      }
-
-      this.forceUpdate();
-    }, 50);
+    this.state = createTree(this.props);
```
lukaszsobek: Think this should be more verbose, spreading the parameters for easier reading instead of putting the whole object in there.

src/components/PrimaryPanes/SourcesTree.js
```diff
@@ -61,7 +58,7 @@ type Props = {
 type State = {
   focusedItem?: any,
   parentMap: any,
-  sourceTree: any,
+  sourceTree: Object,
```
lukaszsobek: the any type makes flow complain about properties.

src/utils/sources-tree/index.js
```diff
@@ -7,13 +7,15 @@
...
 export { collapseTree } from "./collapseTree";
-export { getDirectories } from "./getDirectories";
 export { createTree } from "./createTree";
-export { getURL, getFilenameFromPath } from "./getURL";
+export { formatTree } from "./formatTree";
+export { getDirectories } from "./getDirectories";
+export { sortEntireTree, sortTree } from "./sortTree";
+export { updateTree } from "./updateTree";
+
+export { getFilenameFromPath, getURL } from "./getURL";
```
lukaszsobek: This should be alphabetical.

src/utils/sources-tree/tests/updateTree.spec.js
```diff
@@ -0,0 +1,94 @@
+import { Map } from "immutable";
```
lukaszsobek: missing license header

src/components/PrimaryPanes/SourcesTree.js
```diff
@@ -115,28 +94,32 @@ class SourcesTree extends Component<Props, State> {
...
+      nextProps.sources.size === 0
     ) {
-      // Recreate tree because the sort order changed
-      this.setState(
-        createTree(
-          nextProps.sources,
-          nextProps.debuggeeUrl,
-          nextProps.projectRoot
-        )
+      // early recreate tree because of changes
```
bomsy: feels a little vague, can this be abit more detailed?

src/actions/pause/fetchScopes.js
```diff
@@ -64,13 +64,13 @@ export function fetchScopes() {
...
 
-    if (generatedSourceRecord.get("isWasm")) {
+    if (generatedSourceRecord && generatedSourceRecord.get("isWasm")) {
       return;
     }
 
     const sourceRecord = getSource(getState(), frame.location.sourceId);
 
-    if (sourceRecord.get("isPrettyPrinted")) {
+    if (sourceRecord && sourceRecord.get("isPrettyPrinted")) {
```
jasonLaster: does this need to be here? I think we should take this out because it's unrelated

lukaszsobek: Yeah, I'll pull it out into a separate issue.

src/components/PrimaryPanes/SourcesTree.js
lukaszsobek: Sure.

PR #5015, opened 2018-01-03
🔨: Cleaning up some quick open code and adding tests.

wldcordeiro:
Associated Issue: N/A

### Summary of Changes

* Add some additional tests for `QuickOpenModal`
* Add tests for `parseLineColumn` utility
* trim some code around the modal

### DIFF
--- a/src/components/QuickOpenModal.js
+++ b/src/components/QuickOpenModal.js
@@ -6,7 +6,6 @@
 
 import React, { Component } from "react";
 import { connect } from "react-redux";
-import { bindActionCreators } from "redux";
 import { filter } from "fuzzaldrin-plus";
 
 import actions from "../actions";
@@ -193,7 +192,7 @@ export class QuickOpenModal extends Component<Props, State> {
 
   traverseResults = (direction: number) => {
     const { selectedIndex, results } = this.state;
-    const resultCount = this.resultCount();
+    const resultCount = this.getResultCount();
     const index = selectedIndex + direction;
     const nextIndex = (index + resultCount) % resultCount;
 
@@ -201,6 +200,17 @@ export class QuickOpenModal extends Component<Props, State> {
 
     if (results != null) {
       this.onSelectResultItem(results[nextIndex]);
+    }
+  };
+
+  gotoLocation = () => {
+    const { selectLocation, selectedSource, query } = this.props;
+    if (!selectedSource) {
+      return;
+    }
+    const location = parseLineColumn(query);
+    if (location != null) {
+      selectLocation({ ...location, sourceId: selectedSource.get("id") });
     }
   };
 
@@ -215,7 +225,7 @@ export class QuickOpenModal extends Component<Props, State> {
   };
 
   onKeyDown = (e: SyntheticKeyboardEvent<HTMLElement>) => {
-    const { selectLocation, selectedSource, enabled, query } = this.props;
+    const { enabled } = this.props;
     const { results, selectedIndex } = this.state;
 
     if (!enabled || !results) {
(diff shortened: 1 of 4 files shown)


jasonLaster:
I feel like we can go further with this: 

https://gist.github.com/f38b210f2a200b82306a6f9de08e54b8

wldcordeiro:
Good suggestion! I'll update this PR with that when the shortcut PR lands. 😄

src/components/tests/QuickOpenModal.spec.js
```diff
@@ -101,4 +109,55 @@ describe("QuickOpenModal", () => {
...
+          functions: [],
+          variables: []
+        }
+      },
+      "mount"
+    );
+    wrapper
+      .find("input")
+      .simulate("change", { target: { value: "somefil:33" } });
+    expect(filter).toHaveBeenCalled();
```
jasonLaster: can we be more specific w/ the params passed to filter?

For instance i would guess filter would not be called if we include a `:##` param

src/components/QuickOpenModal.js
```diff
@@ -215,27 +225,20 @@ export class QuickOpenModal extends Component<Props, State> {
...
       return;
     }
 
     const canTraverse = !this.isGotoQuery();
-    if (e.key === "ArrowUp" && canTraverse) {
-      return this.traverseResults(-1);
-    } else if (e.key === "ArrowDown" && canTraverse) {
-      return this.traverseResults(1);
+    const isKeyUpDown = e.key === "ArrowUp" || e.key === "ArrowDown";
+    if (canTraverse && isKeyUpDown) {
```
jasonLaster: if we handle the goTo case first we can drop the `canTraverse` case.

src/components/QuickOpenModal.js
```diff
@@ -215,27 +225,20 @@ export class QuickOpenModal extends Component<Props, State> {
...
+      return this.traverseResults(e.key === "ArrowUp" ? -1 : 1);
     } else if (e.key === "Enter") {
```
jasonLaster: we can replace `else if` with `if` because each case is returning early when hit

src/components/QuickOpenModal.js
```diff
@@ -215,27 +225,20 @@ export class QuickOpenModal extends Component<Props, State> {
...
     }
 
     const canTraverse = !this.isGotoQuery();
-    if (e.key === "ArrowUp" && canTraverse) {
-      return this.traverseResults(-1);
-    } else if (e.key === "ArrowDown" && canTraverse) {
-      return this.traverseResults(1);
+    const isKeyUpDown = e.key === "ArrowUp" || e.key === "ArrowDown";
+    if (canTraverse && isKeyUpDown) {
+      return this.traverseResults(e.key === "ArrowUp" ? -1 : 1);
```
jasonLaster: if we pass `e` to `this.traverseResults` it can figure out the inc

src/components/tests/QuickOpenModal.spec.js
wldcordeiro: `.toHaveBeenCalledWith(...)` over `toHaveBeenCalled()`?

jasonLaster: `toHaveBeenCalledWith` would be good because we can assert the params that are passed as well

src/components/QuickOpenModal.js
```diff
@@ -204,6 +211,20 @@ export class QuickOpenModal extends Component<Props, State> {
     }
   };
 
+  gotoLocation = (location: ?GotoLocationType) => {
+    const { selectLocation, selectedSource } = this.props;
+    const selectedSourceId = selectedSource ? selectedSource.get("id") : "";
```
wldcordeiro: This function isn't ever called in a situation where the selected source won't exist and neither will the `location.sourceId` so this shushes Flow.

The times when the selectedSource doesn't exist we're passing in the ID to use.

jasonLaster:
This is great!

will review tomorrow, but first glance is :cake:

PR #5016, opened 2018-01-03
🔨: Consolidate the project text search components

wldcordeiro:
Associated Issue: N/A

### Summary of Changes

* consolidate the project text search into one component and flow type it.

### Test Plan

Example test plan:

- [x] <kbd>cmd</kbd>+<kbd>shift</kbd>+<kbd>f</kbd> to open project text search
- [x] Test search queries to see if search is performed
- [x] Test navigating the results with keyboard


wldcordeiro:
Still trying to figure out a couple bugs with Flow and then there's a setState call I have to investigate.

jasonLaster:
Nice! I like where this is heading

src/components/ProjectSearch/index.js
```diff
@@ -20,28 +23,77 @@ import {
...
+    this.state = {
+      inputValue: this.props.query || "",
+      inputFocused: false,
+      focusedItem: null
+    };
+
+    this.inputOnChange = this.inputOnChange.bind(this);
+    this.onKeyDown = this.onKeyDown.bind(this);
+    this.onEnterPress = this.onEnterPress.bind(this);
+    this.selectMatchItem = this.selectMatchItem.bind(this);
```
jasonLaster: can we move these to the new format `foo = () => {}`

jasonLaster:
i started this refactor but likely broke stuff :)

src/components/ProjectSearch/index.js
```diff
@@ -20,84 +23,256 @@ import {
...
+  renderResults() {
+    const results = this.getResults().filter(
+      result => result.matches.length > 0
+    );
+
+    const { status } = this.props;
+
+    function getFilePath(item: Match, index?: number) {
+      return item.filepath
+        ? `${item.sourceId}-${index || "$"}`
```
jasonLaster: why the $?

src/components/ProjectSearch/index.js
```diff
@@ -20,84 +23,256 @@ import {
...
+        query={this.state.inputValue}
+        count={resultCount}
+        placeholder={L10N.getStr("projectTextSearch.placeholder")}
+        size="big"
+        summaryMsg={
+          this.props.query !== ""
+            ? L10N.getFormatStr("sourceSearch.resultsSummary1", resultCount)
+            : ""
+        }
+        onChange={e => this.inputOnChange(e)}
```
jasonLaster: could we do? `onChange={this.inputOnChange}`

src/components/ProjectSearch/index.js
```diff
@@ -20,84 +23,256 @@ import {
...
-        closeProjectSearch={closeProjectSearch}
-        selectLocation={selectLocation}
-        query={textSearchQuery}
+      <SearchInput
+        query={this.state.inputValue}
+        count={resultCount}
+        placeholder={L10N.getStr("projectTextSearch.placeholder")}
+        size="big"
+        summaryMsg={
+          this.props.query !== ""
```
jasonLaster: this is copy pasta from before, but I think it could be in a helper function...

src/components/ProjectSearch/index.js
```diff
@@ -20,84 +23,256 @@ import {
...
+      return item.filepath
+        ? `${item.sourceId}-${index || "$"}`
+        : `${item.sourceId}-${item.line}-${item.column}-${index || "$"}`;
+    }
+
+    const renderItem = (item, depth, focused, _, expanded, { setExpanded }) => {
+      return item.filepath
+        ? this.renderFile(item, focused, expanded, setExpanded)
+        : this.renderMatch(item, focused);
+    };
```
jasonLaster: why define `renderItem` in the scope and not on the class?

src/components/ProjectSearch/index.js
```diff
@@ -20,84 +23,256 @@ import {
...
+          getChildren={file => file.matches || []}
+          itemHeight={24}
+          autoExpandAll={true}
+          autoExpandDepth={1}
+          getParent={item => null}
+          getPath={getFilePath}
+          renderItem={renderItem}
+        />
+      );
+    } else if (
```
jasonLaster: minor thing, but it'd be nice to have the fetching check first... 

also, we dont need the else if... could just be `if`

src/components/ProjectSearch/index.js
```diff
@@ -20,84 +23,256 @@ import {
...
+    const { results } = this.props;
+    return results
+      .toJS()
+      .filter(result => result.filepath && result.matches.length > 0);
+  };
+
+  getResultCount = () => {
+    const results = this.getResults();
+    return results.reduce(
+      (count, file) => count + (file.matches ? file.matches.length : 0),
```
jasonLaster: out of curiosity, how can a file not have `matches`?

src/components/ProjectSearch/index.js
```diff
@@ -20,84 +23,256 @@ import {
...
+  sourceId: string,
+  line: number,
+  column: number,
+  match: string,
+  value: string,
+  text: string
+};
+type Result = {
+  filepath: string,
+  matches: Array<Match>,
```
jasonLaster: it looks like matches is not optional, but we assume it is optional in some places...

src/components/ProjectSearch/index.js
wldcordeiro: `ManagedTree` doesn't ever directly pass the index back but `Tree` which `ManagedTree` wraps does, but the effect of this is that you sometimes do and don't have an index so to avoid stringifying `undefined` I put the `$`. This speaks to issues in `ManagedTree` and maybe `Tree` or just the relationship between the two but that should be addressed on its own imo.

src/components/ProjectSearch/index.js
wldcordeiro: I had lifted this from the previous implementation in most cases so some of it wasn't scrutinized much. Good question though!

src/components/ProjectSearch/index.js
wldcordeiro: Yeah I think so. Like listed above, lots was moved without a ton of scrutiny on my part of whether it's done the best way.

src/components/ProjectSearch/index.js
wldcordeiro: Moving the `renderItem` body up to the class does require adding a function signature to make Flow happy too. So that may have been why it was originally put within.

src/components/ProjectSearch/index.js
wldcordeiro: I don't know if it ever doesn't have matches. I had lifted it. 🙁

wldcordeiro: Checked we don't need this ternary. Fixing.

src/components/ProjectSearch/index.js
```diff
@@ -20,84 +23,266 @@ import {
...
+
+import type { List } from "immutable";
+import type { Location } from "../../types";
+import type { ActiveSearchType } from "../../reducers/types";
+import type { StatusType } from "../../reducers/project-text-search";
+
 import "./ProjectSearch.css";
 
+type Match = {
+  type: "MATCH",
```
wldcordeiro: To make `renderItem` work easily adding a type makes a disjoint union easier. Otherwise we'd have compare object members to determine if it's a result or match.

PR #5020, opened 2018-01-03
Fix destructuring

jasonLaster:
### Summary of Changes

#5013 fixed the issue where we could throw an exception while getting symbols. This fixes the bug we were seeing in the debugger bundle.

As a fun fact, I found the bug by dropping the bundle into a test fixture and running it directly. This was easier than debugging the browser toolbox's parser worker :)

### DIFF
--- a/src/workers/parser/tests/__snapshots__/getSymbols.spec.js.snap
+++ b/src/workers/parser/tests/__snapshots__/getSymbols.spec.js.snap
@@ -371,6 +371,7 @@ variables:
 [(12, 17), (12, 27)]   foooo
 [(14, 6), (14, 16)]   foo
 [(14, 23), (14, 31)]   z
+[(16, 7), (16, 15)]   prefName
 
 callExpressions:
 [(1, 21), (1, 28)]   compute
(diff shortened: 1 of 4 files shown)


src/workers/parser/utils/helpers.js
```diff
@@ -78,7 +78,9 @@ export function getVariables(dec: Node) {
       return [];
     }
 
-    return dec.id.elements.map(element => {
+    // NOTE: it's possible that an element is empty
+    // e.g. const [, a] = arr
+    return dec.id.elements.filter(element => element).map(element => {
```
wldcordeiro: I use this at times and still forget that it's possible! 👍 Love the addition of the test fixture too.

src/workers/parser/tests/__snapshots__/getSymbols.spec.js.snap
```diff
@@ -441,6 +443,10 @@ identifiers:
 [(14, 7), (14, 10)]  key key
 [(14, 13), (14, 16)]  foo foo
 [(14, 23), (14, 24)]  z z
+[(16, 4), (16, 47)]
```
AnshulMalik: What's this line for? It has no identifier

wldcordeiro: From what I can tell the test prints the value and location for each symbol and I guess for the iteration it produces `undefined` and that got stringified to blank?

jasonLaster: hmm.. good find, lets try and filter it out before this lands

wldcordeiro: This is happening because of the `formatSymbols` helper in `src/workers/parser/utils/formatSymbols.js` specifically because of [this](https://github.com/devtools-html/debugger.html/blob/master/src/workers/parser/utils/formatSymbols.js#L31)

jasonLaster:
Nice catch @AnshulMalik!

PR #5041, opened 2018-01-06
Refactor Editor utils

jasonLaster:
### Summary of Changes

* Bring devtools-source-editor back into the debugger
* Switch `module.exports` to `export function`
* Add some more types 
* Add some safety guards e.g. `if (!line)`


lukaszsobek:
Does this need a rebase? Flow is complaining about a completely different file here:
flow-error

Issue #5044, opened 2018-01-06
[Docs] Mention docs as component in issue.md


lukaszsobek:
It would be great if documentation issues were perpended with [Docs] or [Doc] in order for contributors to be able to quickly assess what the issue is about. 

One of the steps to achieve that is to add information about doc labels to https://github.com/devtools-html/debugger.html/blob/master/docs/issues.md

PR #5045, opened 2018-01-06
[Docs] Adds docs as possible issue title start tag

lukaszsobek:
Associated Issue: #5044

### Summary of Changes

* adds docs as possible issue title start tag
* expands on examples on how to use start tags

### DIFF
--- a/docs/issues.md
+++ b/docs/issues.md
@@ -14,15 +14,25 @@
 **Components**
 
 Issues are organized in terms of components.
-Issue titles should include a component at the front e.g. `[Editor]`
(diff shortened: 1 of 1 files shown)


wldcordeiro:
Circle is failing on Markdown stuff but it was something else. 🤔 As for dep-ci I swear that isn't working right now so false positive there.

lukaszsobek:
Reworded the issues. Don't know why the discrepancies did not get picked up earlier - maybe they were changes from back before the linter was introduced and since I was the first to touch the doc it was my privilege to spring the trap.

Issue #5050, opened 2018-01-07
[TextSearch] stuck at "Loading..." when there is no result


Loirooriol:
https://bugzilla.mozilla.org/show_bug.cgi?id=1428586

1. Go to http://example.com
2. Open the debugger
3. Press Ctrl+Shift+F
4. Write 123
5. Press enter

Result: A "Loading..." message
Expected: A "No results found" message

Regressed in https://hg.mozilla.org/integration/autoland/pushloghtml?fromchange=003d418a3516b9fffc40f4a5f6f9f6e75f2de448&tochange=b7b279271a15188af2ba3cb6c8e4f119545fd1c7

PR #5051, opened 2018-01-08
Changed error to 'No results found'

Vasusena:
Associated Issue: #5050 

Here's the Pull Request Doc

### Summary of Changes

Here is the screen shot for the rectified error message
img-4069

### DIFF
--- a/src/components/ProjectSearch/TextSearch.js
+++ b/src/components/ProjectSearch/TextSearch.js
@@ -184,11 +184,10 @@ export default class TextSearch extends Component<Props> {
           renderItem={renderItem}
         />
       );
-    } else if (status === statusType.fetching) {
-      return (
-        <div className="no-result-msg absolute-center">No results found.</div>
-      );
-    } else if (this.props.query && !results.length) {
+    } else if (
+      (this.props.query && !results.length) ||
+      status === statusType.fetching
+    ) {
       return (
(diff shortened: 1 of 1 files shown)


src/components/ProjectSearch/TextSearch.js
```diff
@@ -185,7 +185,9 @@ export default class TextSearch extends Component<Props> {
         />
       );
     } else if (status === statusType.fetching) {
-      return <div className="no-result-msg absolute-center">Loading...</div>;
+      return (
+        <div className="no-result-msg absolute-center">No results found.</div>
```
lukaszsobek: Nicely done! One improvement would be that the `No results found` text needs to be localized through L10N. (if you search the file for `L10N` you'll see how it works)

src/components/ProjectSearch/TextSearch.js
```diff
@@ -185,7 +185,9 @@ export default class TextSearch extends Component<Props> {
         />
       );
     } else if (status === statusType.fetching) {
```
lukaszsobek: If you look below you will find that `} else if (this.props.query && !results.length) {` also returns the same piece of code, so it might be a good idea to join conditions with a `||`

darkwing:
Made a small update here to group the conditionals and use localized text.  Thanks for the awesome start @Vasusena !

PR #5055, opened 2018-01-08
Show vertical mode on open

jasonLaster:
### Summary of Changes

Opening the debugger below the width bp was not triggering vertical mode.

### Test Plan

mochitests


jasonLaster:
### Before

screen shot 2018-01-08 at 12 13 40 pm

### After

screen shot 2018-01-08 at 12 13 26 pm

PR #5058, opened 2018-01-08
Fix pane ordering

jasonLaster:
### Summary of Changes

When we landed the #4974 work, we accidentally flipped the ordering of the scopes and call stack panes.


jasonLaster:
screen shot 2018-01-08 at 6 16 38 pm
screen shot 2018-01-08 at 6 16 31 pm

PR #5060, opened 2018-01-09
Add the services argument

bomsy:
### Summary of Changes

* Added the `services` which was being used but never defined.

### DIFF
--- a/src/actions/event-listeners.js
+++ b/src/actions/event-listeners.js
@@ -12,8 +12,9 @@
 
 import { reportException } from "../utils/DevToolsUtils";
 import { isPaused, getSourceByURL } from "../selectors";
-import waitUntilService from "./utils/middleware/wait-service";
-const services = { WAIT_UNTIL: waitUntilService.NAME };
+const services = {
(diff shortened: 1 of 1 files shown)


src/actions/event-listeners.js
```diff
@@ -47,7 +47,7 @@ async function asPaused(state: any, client: any, func: any) {
  * @static
  */
 export function fetchEventListeners() {
-  return ({ dispatch, getState, client }) => {
+  return ({ dispatch, getState, client, services }) => {
```
jasonLaster: i don't know if that has `WAIT_UNTIL`...

src/actions/event-listeners.js
```diff
@@ -12,6 +12,9 @@
 
 import { reportException } from "../utils/DevToolsUtils";
 import { isPaused, getSourceByURL } from "../selectors";
+const services = {
+  WAIT_UNTIL: require("./utils/middleware/wait-service").NAME
+};
```
jasonLaster: lets use import

`import { NAME as WAIT_UNTIL } from "./utils/middleware/wait-service"`

Issue #5023, opened 2018-01-04
[CallStack] Broken Stacktrace is back


TitanNano:
![screen shot 2018-01-04 at 09 52 27](https://user-images.githubusercontent.com/4500858/34556375-4f574396-f136-11e7-941e-0446a95d88a5.png)
This happened after I stepped into an Angular internal method. 
```
TypeError: Argument 1 of Node.removeChild is not an object.[Weitere Informationen] react-dom.js:1379:3
```

TitanNano:
```
Unknown SVG: rxjs  debugger.js:33573:5
Error: Minified React error #105; 
visit http://facebook.github.io/react/docs/error-decoder.html?invariant=105&args[]=Svg 
for the full message or use the non-minified dev environment for full errors and additional helpful warnings.  react-dom.js:16637:15
```

This seems to be one of the causes for this issue.

TitanNano:
Steps to Reproduce:
1. go to https://jovandemo.modulentwicklung.de:8001/admin
2. login with User: anyone@mozilla.org PW: 123456789
3. go to Orders > Orders
4. Open Debugger and search for file: Shared/Directives/DataTable/DataTableController.ts
5. Set a breakpoint in line 197
6. Move cursor in side the view-port to trigger angular

Result: No call stack is shown, no scopes are shown either.

Additional:

1. Right click and select jump to generated source

Result: a new tab appears with the generated source, on the left the broken call stack appears too.

bomsy:
Thanks @TitanNano !  will look into this!

PR #5062, opened 2018-01-09
fix to return null instead of undefined

bomsy:
Associated Issue: #5023

### Summary of Changes

* return `null` from instead of `undefined` which broke react rendering

### Test Plan

Example test plan:

Steps to Reproduce:

- go to https://jovandemo.modulentwicklung.de:8001/admin
- login with User: anyone@mozilla.org PW: 123456789
- go to Orders > Orders
- Open Debugger and search for file: Shared/Directives/DataTable/DataTableController.ts
- Set a breakpoint in line 197
 - Move cursor inside the view-port to trigger angular

image

### DIFF
--- a/assets/images/Svg.js
+++ b/assets/images/Svg.js
@@ -77,7 +77,7 @@ function Svg({ name, className, onClick, "aria-label": ariaLabel }) {
     }
 
     console.warn(error);
-    return;
+    return null;
   }
 
   className = `${name} ${className || ""}`;
--- a/assets/images/Svg.js
+++ b/assets/images/Svg.js
@@ -59,7 +59,8 @@ const svg = {
   nextjs: require("./nextjs.svg"),
   showSources: require("./showSources.svg"),
   showOutline: require("./showOutline.svg"),
-  nuxtjs: require("./nuxtjs.svg")
+  nuxtjs: require("./nuxtjs.svg"),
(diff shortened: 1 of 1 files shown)


jasonLaster:
how was it breaking?

assets/images/Svg.js
```diff
@@ -59,7 +59,8 @@ const svg = {
   nextjs: require("./nextjs.svg"),
   showSources: require("./showSources.svg"),
   showOutline: require("./showOutline.svg"),
-  nuxtjs: require("./nuxtjs.svg")
+  nuxtjs: require("./nuxtjs.svg"),
+  rxjs: require("./rxjs.svg")
```
codehag: interesting, why are we adding this back in?

bomsy: We have an issue which was looking to render the `rxjs` svg and couldn't find it so it return `undefined` which broke rendering. The svg file but it was not included in the map above.

bomsy: Not sure why the tests are failing ..

jasonLaster: no worries. i've got it

TitanNano:
@jasonLaster how long until this will land in nightly?

jasonLaster:
it will land tomorrow.

Issue #5057, opened 2018-01-08
[Pause] Immediately pausing in an original file


jasonLaster:
I noticed while debugging that when the debugger pauses immediately in an original location it causes an exception in the debugger UI that prevents the source from being shown:

<img src='https://shipusercontent.com/e7427b81981c75829228c4128022d5d4/Screen%20Shot%202018-01-08%20at%206.10.52%20PM.png' title='Screen Shot 2018-01-08 at 6.10.52 PM.png' width=598>

jasonLaster:
I think this might be related:

```
 Adding a new tab with URL: http://example.com/browser/devtools/client/debugger/new/test/mochitest/examples/doc-debugger-statements.html
   1 ADD_SOURCES server1.conn15.child1/source27 - doc-debugger-statements.html
   1 CONNECT http://example.com/browser/devtools/client/debugger/new/test/mochitest/examples/doc-debugger-statements.html
   1 SET_EXPANDED_STATE 
   1 SET_ORIENTATION 
   1 SET_WORKERS 
   Waiting for state change: paused
   1 NAVIGATE 
   1 PAUSED debuggerStatement: server1.conn15.child1/source27:11:6
   1 MAP_FRAMES 
   1 CLEAR_SELECTED_SOURCE 
   1 ADD_SCOPES 
   Finished waiting for state change: paused
   Waiting for state change: loaded source
 JavaScript error: resource://devtools/shared/base-loader.js -> resource://devtools/client/debugger/new/debugger.js, line 15288: TypeError: generatedSourceRecord is undefined
  JS Error: 
    "TypeError: generatedSourceRecord is undefined" {file: "resource://devtools/shared/base-loader.js -> resource://devtools/client/debugger/new/debugger.js" line: 15288}]
 TEST-UNEXPECTED-FAIL Test timed out 
 TEST-UNEXPECTED-FAIL A promise chain failed to handle a rejection: generatedSourceRecord is undefined 
 stack: null
```

PR #5076, opened 2018-01-10
Support pausing immediately

jasonLaster:
Associated Issue: #5057 

### Summary of Changes

When we pause we need to wait for sources to be registered in redux and the parser worker. This is why when we pause we flush the newSources queue. The issue is we also need to wait on original sources  to be fetched from the source map worker :)

### Screenshots

| Before | After |
| -- | --- |
| ![b] | ![a] |

[b]: https://user-images.githubusercontent.com/254562/34797795-97a8bbe8-f627-11e7-83d9-92519b673696.png
[a]: https://user-images.githubusercontent.com/254562/34797796-97b52748-f627-11e7-97a1-53ec01d44faa.png

### DIFF
--- a/src/actions/navigation.js
+++ b/src/actions/navigation.js
@@ -41,13 +41,13 @@ export function willNavigate(_: any, event: Object) {
     clearASTs();
     clearScopes();
     clearSources();
-    sourceQueue.clear();
-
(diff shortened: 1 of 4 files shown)


jasonLaster:
@codehag found that to be a good interval

PR #5078, opened 2018-01-10
Cleanup Preview

jasonLaster:
### Summary of Changes

* cleans up the reducer a bit
* prevents pre-mature highlighting and arrows
* simplifies logic a bit

### DIFF
--- a/src/components/Editor/Preview/Popup.js
+++ b/src/components/Editor/Preview/Popup.js
@@ -53,11 +53,11 @@ export class Popup extends Component<Props> {
       range
     } = this.props;
 
-    this.marker = markText(editor, "preview-selection", range);
-
     if (!value || !value.type == "object") {
       return;
     }
+
+    this.marker = markText(editor, "preview-selection", range);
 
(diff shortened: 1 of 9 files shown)


src/reducers/ast.js
```diff
@@ -99,23 +99,9 @@ function update(
...
-
       return state.set("preview", {
-        updating: false,
-        expression,
-        location,
-        result,
-        tokenPos,
-        cursorPos,
-        extra
+        ...action.value,
```
bomsy: 👍

src/components/Editor/Preview/Popup.js
```diff
@@ -85,9 +99,10 @@ export class Popup extends Component<Props> {
     return null;
   }
 
-  renderFunctionPreview(value: Object, root: Object) {
-    const { selectSourceURL } = this.props;
+  renderFunctionPreview() {
+    const { selectSourceURL, value } = this.props;
     const { location } = value;
+    const root = this.getRoot();
```
bomsy: this `root` here does not seem to be used within this function

Issue #4547, opened 2017-10-31
[WASM] Show a notice on WASM/asm.js modules to refresh the page


digitarald:
Before reload with Debugger open asm.js/WASM modules are not debuggable. If a user tried to debug such a module, we should show a simple notice `Please refresh to debug this module [Learn more]`.

yurydelendik:
Fixed by the #5090

PR #5090, opened 2018-01-12
Fixes loading error

yurydelendik:
Associated Issue: #4547 and also general error during source text loading

TODO:
- [x] Fix "Editor › When load..."-tests
- [x] When panel with error message shown, during refresh not all sources present in the source tree #5094
- [x] Test

### DIFF
--- a/assets/panel/debugger.properties
+++ b/assets/panel/debugger.properties
@@ -581,6 +581,10 @@ pauseOnExceptions=Pause on all exceptions. Click to ignore exceptions
 # yet.
 loadingText=Loading\u2026
 
+# LOCALIZATION NOTE (wasmIsNotAvailable): The text that is displayed in the
+# script editor when the WebAssembly source is not available.
+wasmIsNotAvailable=Please refresh to debug this module
+
 # LOCALIZATION NOTE (errorLoadingText3): The text that is displayed in the debugger
 # viewer when there is an error loading a file
 errorLoadingText3=Error loading this URI: %S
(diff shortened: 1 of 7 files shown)


src/actions/sources/loadSourceText.js
```diff
@@ -82,20 +82,24 @@ export function loadSourceText(source: SourceRecord) {
...
-    });
+    try {
+      await dispatch({
+        type: "LOAD_SOURCE_TEXT",
+        sourceId: id,
+        [PROMISE]: loadSource(source, { sourceMaps, client })
+      });
+    } catch (e) {
+      // Ignoring the rejected promise -- source already has 'error' property.
+    }
```
jasonLaster: can we move the catch into `loadSource` so that it doesn't reject the promise?

src/components/Editor/index.js
```diff
@@ -464,6 +465,7 @@ class Editor extends PureComponent<Props, State> {
     }
 
     resetLineNumberFormat(editor);
+    showErrorMessage(editor, msg);
```
jasonLaster: nice catch

assets/panel/debugger.properties
```diff
@@ -581,6 +581,10 @@ pauseOnExceptions=Pause on all exceptions. Click to ignore exceptions
 # yet.
 loadingText=Loading\u2026
 
+# LOCALIZATION NOTE (wasmIsNotAvailable): The text that is displayed in the
+# script editor when the WebAssembly source is not avaiable.
```
flodolo: typo: available

src/actions/sources/loadSourceText.js
yurydelendik: Nope, you need dispatch() to send the "error" status.

jasonLaster: i see...

PR #5095, opened 2018-01-12
Make loading source maps safer

jasonLaster:
### Summary of Changes

I noticed while testing https://reactjs.org/ that it had a bad map, which was causing an exception so i wrapped the API.

### DIFF
--- a/src/actions/sources/newSources.js
+++ b/src/actions/sources/newSources.js
@@ -41,14 +41,23 @@ function createOriginalSource(
   };
 }
 
+// TODO: It would be nice to make getOriginalURLs a safer api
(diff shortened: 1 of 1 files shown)


src/actions/sources/newSources.js
```diff
@@ -41,14 +41,23 @@ function createOriginalSource(
   };
 }
 
+// TODO: It would be nice to make getOriginalURLs a safer api
```
tromey: FWIW the toolbox wraps this in a proxy that reports errors to the console.

jasonLaster: great

PR #5096, opened 2018-01-12
Speedup sync bp

jasonLaster:
### Summary of Changes

When the page loads with breakpoints, the debugger can take a really long time to show the saved breakpoints. 

The reason for this is that syncing an original bp involves loading & parsing the generated source. We need to load the generated source, but we do not need to parse it!

Secondary perf wins:

* we pass the entire source object to the worker every time we make a call. We can save serializing time by avoiding that
* we can avoid unnecessary calls.

This PR is probably too big to land by itself, I hope to land a couple of smaller changes:

- [ ] refactor the api to source id: #5098
- [x] refactor newSource #5099
- [ ] addBreakpoint should use symbols from the store
- [x] getSymbols when we select the source #5105
- [x]  speedup generated file parsing #5101

### DIFF
--- a/src/actions/sources/loadSourceText.js
+++ b/src/actions/sources/loadSourceText.js
@@ -6,7 +6,6 @@
 
 import { isOriginalId } from "devtools-source-map";
 import { PROMISE } from "../utils/middleware/promise";
-import { setSymbols } from "../ast";
 import {
   getSource,
   getGeneratedSource,
@@ -97,8 +96,9 @@ export function loadSourceText(source: SourceRecord) {
 
     if (!newSource.isWasm) {
       await parser.setSource(newSource);
-      dispatch(setSymbols(id));
     }
+
+    console.log("LOAD_SOURCE_TEXT LOADED");
 
     // signal that the action is finished
     deferred.resolve();
--- a/src/actions/sources/newSources.js
+++ b/src/actions/sources/newSources.js
@@ -20,7 +20,8 @@ import { prefs } from "../../utils/prefs";
 import {
   getSource,
   getPendingSelectedLocation,
-  getPendingBreakpointsForSource
+  getPendingBreakpointsForSource,
+  getGeneratedSource
 } from "../../selectors";
 
 import type { Source } from "../../types";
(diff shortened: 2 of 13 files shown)


src/actions/sources/newSources.js
```diff
@@ -105,8 +119,19 @@ function checkPendingBreakpoints(sourceId) {
...
 
+    console.log(
+      `checkPendingBreakpoints ${source.get("url")}`,
+      pendingBreakpoints.toJS()
+    );
+
     // load the source text if there is a pending breakpoint for it
     await dispatch(loadSourceText(source));
+    console.log("checkPendingBreakpoints LOADED");
+    // const generatedSource = getGeneratedSource(getState(), source.toJS());
```
codehag: probably dont need the comments here

src/workers/parser/getSymbols.js
```diff
@@ -401,3 +389,17 @@ function getSnippet(path, prevPath, expression = "") {
 export function clearSymbols() {
   symbolDeclarations = new Map();
 }
+
+export default function getSymbols(sourceId: string): SymbolDeclarations {
```
codehag: :+1:

src/workers/parser/utils/ast.js
```diff
@@ -60,15 +61,12 @@ export function parseScript(text: string, opts?: Object) {
...
-  }
-
-  if (ASTs.has(source.id)) {
-    return ASTs.get(source.id);
+export function getAst(sourceId: string) {
+  if (ASTs.has(sourceId)) {
+    return ASTs.get(sourceId);
   }
 
+  const source = getSource(sourceId);
```
codehag: :+1:

src/actions/sources/loadSourceText.js
```diff
@@ -97,9 +96,10 @@ export function loadSourceText(source: SourceRecord) {
 
     if (!newSource.isWasm) {
       await parser.setSource(newSource);
-      dispatch(setSymbols(id));
     }
 
+    console.log("LOAD_SOURCE_TEXT LOADED");
```
codehag: this can go

src/actions/sources/loadSourceText.js
```diff
@@ -97,9 +96,10 @@ export function loadSourceText(source: SourceRecord) {
 
     if (!newSource.isWasm) {
       await parser.setSource(newSource);
-      dispatch(setSymbols(id));
```
codehag: im guessing we are saving a lot here -- why did we have this originally?

PR #5098, opened 2018-01-12
[Parser] switch to source id API

jasonLaster:
Associated Issue: #4849

### Summary of Changes

This PR continues @yurydelendik's work to switch to a `sourceId` based API. This saves us ~100ms when we're passing large objects back and forth like a source w/ text.

screen shot 2018-01-12 at 3 05 24 pm

PR #5099, opened 2018-01-12
Refactor newSource

jasonLaster:
### Summary of Changes

Small refactor to consolidate the logic between `newSource` and `newSources`. We don't use `newSource` in the app, but we do in the tests. In the future, it might be nice to have around too...


jasonLaster:
This PR is probably too big to land by itself, I hope to land a couple of smaller changes:

1. refactor the api to source id: #5098
2. refactor newSource #5099
3. addBreakpoint should use symbols from the store
4. getSymbols when we select the source

PR #5101, opened 2018-01-13
Speed up generated source parsing

jasonLaster:
Associated Issue: #5096 

### Summary of Changes

Working on #5096, the biggest slow down comes from parsing + traversing the large generated files (aka bundles). The debugger bundles can take ~12 seconds in the debugger worker. When it's parsed in a jest test it's about ~7 seconds. 

When we parse the file without the babylon plugins, the time goes down to ~3 seconds. This could be a big savings in the debugger as we only need to parse + traverse the bundle once.

### DIFF
--- a/src/utils/test-head.js
+++ b/src/utils/test-head.js
@@ -57,6 +57,11 @@ function makeSource(name: string, props: any = {}) {
   };
 }
 
+function makeOriginalSource(name: string, props?: Object) {
+  const source = makeSource(name, props);
+  return { ...source, id: `${name}-original` };
+}
+
 function makeSourceRecord(name: string, props: any = {}) {
   return I.Map(makeSource(name, props));
 }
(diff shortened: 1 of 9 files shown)


src/workers/parser/utils/ast.js
```diff
@@ -74,7 +78,9 @@ export function getAst(source: Source) {
   if (contentType == "text/html") {
     ast = parseScriptTags(source.text, htmlParser) || {};
   } else if (contentType && contentType.match(/(javascript|jsx)/)) {
-    ast = parse(source.text);
+    const type = source.id.includes("original") ? "original" : "generated";
```
wldcordeiro: This is really smart! it doesn't make sense to use all the plugins on bundles and things that have been processed. 🚀

src/utils/test-head.js
```diff
@@ -57,6 +57,11 @@ function makeSource(name: string, props: any = {}) {
   };
 }
 
+function makeOriginalSource(name: string, props?: Object) {
```
wldcordeiro: 👍 Love little utilities like this.

PR #5105, opened 2018-01-15
Get symbols in select

jasonLaster:
Associated Issue: #5096 

### Summary of Changes

* Moves `getSymbols` from `loadSourceText` to `selectLocation`.
* The rationale is that loadSource could happen much on debuggee load, which is unnecessary

### DIFF
--- a/src/actions/sources/select.js
+++ b/src/actions/sources/select.js
@@ -11,7 +11,7 @@
 
 import { isOriginalId } from "devtools-source-map";
 
-import { setOutOfScopeLocations } from "../ast";
+import { setOutOfScopeLocations, setSymbols } from "../ast";
 import { closeActiveSearch } from "../ui";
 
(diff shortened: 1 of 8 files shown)


src/actions/sources/select.js
```diff
@@ -119,16 +119,22 @@ export function selectLocation(location: Location, tabIndex: string = "") {
...
     const selectedSource = getSelectedSource(getState());
+    if (!selectedSource) {
+      return;
+    }
+
+    const sourceId = selectedSource.get("id");
     if (
       prefs.autoPrettyPrint &&
-      !getPrettySource(getState(), selectedSource.get("id")) &&
+      !getPrettySource(getState()) &&
```
bomsy: is the second argument optional? if not do we need to update getPrettySource definition?

jasonLaster: nice catch

Issue #5100, opened 2018-01-13
Test Coverage Push


wldcordeiro:
This is a tracking issue for the test coverage push. The goal is to increase our test coverage 10-20% by targeting some lower hanging fruits. I had started this push around January 9th 2018 and the output of `yarn test --coverage` gave me ~*56%* as our total coverage. Let's see where we're at on February 9th!

## Stretch goal bump total coverage to ~*76%*

* [ ] `src/components/shared
  - [x] `Button/`
    - [x] Close
    - [x] CommandBarButton
    - [x] PaneToggle
  - [x] Accordion
  - [x] Badge
  - [x] BracketArrow
  - [x] Dropdown
  - [ ] ManagedTree
  - [x] Modal
  - [x] Popover
  - [x] PreviewFunction
  - [x] ResultList
  - [x] SearchInput
  - [ ] Svg
* [ ] `src/utils/editor/`
  - [ ] create-editor
  - [ ] getTokenLocation
  - [ ] index.js
  - [ ] source-documents.js
  - [ ] source-editor.js
  - [ ] source-search.js
...

jasonLaster:
Here's the first PR https://github.com/devtools-html/debugger.html/pull/5097

It's a GREAT example for how to write jest unit tests!

jasonLaster:
I'll work on `scr/utils/frame`

wldcordeiro:
Working on `src/utils/editor/`

wldcordeiro:
Updating the issue with adjusted goals as we had been measuring the coverage wrong and Jest was missing stuff. Lower percentage now. 🙁 Let's get this up there! 🎉

mutdmour:
Covered SourcesTree. Will work on Outline next 😀

adancarrasco:
Hi, I'm planning to help on this task, just wondering if the list is up-to-date? I was planning to contribute to `astBreakpointLocation test` but I'm seeing it in the code from yesterday's cloned repo. 😃

adancarrasco:
I will work on `src/utils/ui`.

mutdmour:
I will work on `source-documents`

jasonLaster:
@kenjiO go for it!

marcogallo566:
Can I work on `src/utils/fromJS`?

jasonLaster:
@marcogallo566 we're using that file less and less, id recommend a different one

marcogallo566:
[...]

PR #5118, opened 2018-01-16
Moar testing

wldcordeiro:
Associated Issue: #5100 

### Summary of Changes

* Fix a flow issue in Popover code
* Adding tests for majority of ManagedTree behavior

### DIFF
--- a/src/components/shared/ManagedTree.js
+++ b/src/components/shared/ManagedTree.js
@@ -40,8 +40,7 @@ type State = {
 
 class ManagedTree extends Component<Props, State> {
   constructor(props: Props) {
-    super();
-
+    super(props);
     this.state = {
       expanded: props.expanded || new Set(),
       focusedItem: null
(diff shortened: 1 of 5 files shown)


src/components/shared/ManagedTree.js
```diff
@@ -40,21 +40,18 @@ type State = {
...
     this.state = {
       expanded: props.expanded || new Set(),
       focusedItem: null
     };
   }
 
-  componentWillReceiveProps(nextProps: Props) {
-    const listItems = nextProps.listItems;
+  componentWillReceiveProps({ listItems, highlightItems, focused }: Props) {
     if (listItems && listItems != this.props.listItems && listItems.length) {
```
jasonLaster: minor nit:

```js
   componentWillReceiveProps(nextProps: Props) {
     const {listItems} = this.props;
```

typically we're destructuring on the next line and using props

wldcordeiro: I destructed the nextProps because I noticed we only used those specific props.

Issue #5108, opened 2018-01-15
[ProjectSearch] sad face while typing


jasonLaster:
It looks like the recent consolidation #5016 is causing the sad face to appear before the user does a search:

STR:
1. cmd+shift+f
2. start typing 

see: sad face

<img src='https://shipusercontent.com/c3bee147c1b711b002ddb28cccc5aae7/Screen%20Shot%202018-01-15%20at%205.51.19%20PM.png' title='Screen Shot 2018-01-15 at 5.51.19 PM.png' width=575>

jbhoosreddy:
done in #5124

PR #5124, opened 2018-01-17
(refactor) [Project search] Modify status update and error emoji (#5108)

jbhoosreddy:
Associated Issue: #5108

### Summary of Changes

With this commit, the Status behavior is stored differently in reducers.

- INITIAL -> Search not yet performed
- FETCHING -> Search in progress
- DONE -> Search completed | Smiley Shown | After all sources searched
    - (not fired after each source as before)

- Render error emoji only on done
    - Removes a slight jitter in the UI when the results are still inflight.

- Use the shouldShowErrorEmoji predicate instead of passing status

### Test Plan
- CmdOrCtrl+Shift+F
- As per above

### DIFF
--- a/src/actions/project-text-search.js
+++ b/src/actions/project-text-search.js
@@ -38,8 +38,7 @@ export function clearSearchResults() {
 
 export function clearSearch() {
   return ({ dispatch, getState }: ThunkArgs) => {
-    dispatch(clearSearchQuery());
-    dispatch(clearSearchResults());
+    dispatch({ type: "CLEAR_SEARCH" });
   };
 }
 
--- a/src/actions/project-text-search.js
+++ b/src/actions/project-text-search.js
@@ -53,9 +53,6 @@ export function closeProjectSearch() {
 
 export function searchSources(query: string) {
   return async ({ dispatch, getState }: ThunkArgs) => {
-    if (!query) {
-      return;
-    }
     await dispatch(clearSearchResults());
     await dispatch(addSearchQuery(query));
     dispatch(updateSearchStatus(statusType.fetching));
--- a/src/actions/project-text-search.js
+++ b/src/actions/project-text-search.js
@@ -18,28 +18,19 @@ import { statusType } from "../reducers/project-text-search";
 import type { ThunkArgs } from "./types";
 
 export function addSearchQuery(query: string) {
-  return ({ dispatch, getState }: ThunkArgs) => {
-    dispatch({ type: "ADD_QUERY", query });
-  };
+  return { type: "ADD_QUERY", query };
 }
 
 export function clearSearchQuery() {
-  return ({ dispatch, getState }: ThunkArgs) => {
-    dispatch({ type: "CLEAR_QUERY" });
-    dispatch(updateSearchStatus(statusType.initial));
-  };
+  return { type: "CLEAR_QUERY" };
 }
 
 export function clearSearchResults() {
-  return ({ dispatch, getState }: ThunkArgs) => {
-    dispatch({ type: "CLEAR_SEARCH_RESULTS" });
-  };
+  return { type: "CLEAR_SEARCH_RESULTS" };
 }
 
 export function clearSearch() {
-  return ({ dispatch, getState }: ThunkArgs) => {
-    dispatch({ type: "CLEAR_SEARCH" });
-  };
+  return { type: "CLEAR_SEARCH" };
 }
 
 export function updateSearchStatus(status: string) {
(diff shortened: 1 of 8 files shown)


src/actions/project-text-search.js
```diff
@@ -45,6 +53,9 @@ export function closeProjectSearch() {
 
 export function searchSources(query: string) {
   return async ({ dispatch, getState }: ThunkArgs) => {
+    if (!query) {
```
jbhoosreddy: wondering if I should move this guard logic to the component instead?

src/components/ProjectSearch.js
```diff
@@ -76,6 +78,11 @@ function getFilePath(item: Item, index?: number) {
     : `${item.sourceId}-${item.line}-${item.column}-${index || "$"}`;
 }
 
+function sanitizeQuery(query) {
+  // no '\' at end of query
```
jbhoosreddy: prevents wierd console error

jbhoosreddy: hopefully inefficient regexs are not too bad here.

src/components/ProjectSearch.js
```diff
@@ -170,7 +177,11 @@ export class ProjectSearch extends Component<Props, State> {
 
   inputOnChange = (e: SyntheticInputEvent<HTMLInputElement>) => {
     const inputValue = e.target.value;
+    this.props.updateSearchStatus(statusType.initial);
```
wldcordeiro: Each time the user types we're setting the status to initial?

src/actions/project-text-search.js
```diff
@@ -35,6 +36,13 @@ export function clearSearchResults() {
   };
 }
 
+export function clearSearch() {
```
wldcordeiro: Rather than dispatching two actions couldn't we have one action that is listened to in the reducer and does the work?

src/actions/project-text-search.js
wldcordeiro: Probably a good idea.

wldcordeiro:
One suggestion I think would maybe simplify things. The search input is used by the project search, quick open modal and file search. In 2 of 3 instances we only want it to apply search and smiley on the user pressing enter. While the third we want it to apply on each key press. We could potentially add a prop on `SearchInput` that changes the behavior from only on enter or not.

src/components/ProjectSearch.js
jbhoosreddy: I added this line in later. Would really leave it to you.

By only changing it to initial, it signified to me that we're about to make a new search once we press enter.

src/actions/project-text-search.js
jbhoosreddy: sure.. I'll hook in a proper action in place.

src/components/ProjectSearch.js
wldcordeiro: We should avoid the work either here or in the action if the status is already initial.

jbhoosreddy:
@wldcordeiro could you elaborate your thoughts on adding an additional prop (reactive?) to SearchInput. I tried to sketch it out, but it's behaviour on userinput is driven externally, so is the smiley behavior. It seems to add complexity to make it reactive. But we can also punt on it.

Any ways, up for review again.

jbhoosreddy:
I think I can do the reactive SearchInput stuff in a followup PR.

src/actions/project-text-search.js
```diff
@@ -35,6 +36,12 @@ export function clearSearchResults() {
...
+  return ({ dispatch, getState }: ThunkArgs) => {
```
wldcordeiro: Since this only fires the one action you can remove the wrapping `return ({ dispatch, getState }: ThunkArgs) => {...}` and return just the object.

src/reducers/project-text-search.js
```diff
@@ -72,6 +72,12 @@ function update(
         results: state.get("results").clear()
       });
 
+    case "CLEAR_SEARCH":
+      const newSate = state.remove("query");
```
wldcordeiro: newState?

jbhoosreddy:
@wldcordeiro sorry for the delay. should be good to go. thanks

src/actions/tests/project-text-search.spec.js
```diff
@@ -8,8 +8,8 @@ import {
 const {
   getTextSearchQuery,
   getTextSearchResults,
-  getSource,
-  getTextSearchStatus
+  getTextSearchStatus,
+  getSource
```
lukaszsobek: Alphabetical order would be great here.

PR #5128, opened 2018-01-17
Use pushPref helper in mochitests to set temporary pref values

juliandescottes:
Use the pushPref helper whenever possible in mochitests, otherwise the preference changes leak from one test to another. 

Also simplified some tests that used to set the preference and restore it on cleanup. First, there was no added value compared to pushPref. Secondly, the restore was not wrong. It should restore to the original value (read at the beginning of the test) rather than to a random set value (eg. false).

Let's see if tests still pass.

### DIFF
--- a/src/test/mochitest/browser_dbg-async-stepping.js
+++ b/src/test/mochitest/browser_dbg-async-stepping.js
@@ -3,7 +3,7 @@
 
 // Tests async stepping will step over await statements
 add_task(async function test() {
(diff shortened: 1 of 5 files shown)


juliandescottes:
Can't use it for devtools.toolbox.splitconsoleEnabled in browser_dbg-console.js because this pref is automatically cleared by shared-head.js and pushPref has a tendency to freeze when the prefs get modified by something else during the test.

PR #5129, opened 2018-01-17
Tests for editor utils

wldcordeiro:
Associated Issue: #5100 

### Summary of Changes

* adding tests in `src/utils/editor/`


jasonLaster:
> can we avoid this work?

move the conversion to their own PR...

wldcordeiro:
@jasonLaster we could. I had just seen a types file in the editor utils directory that I was working in and I saw it was missing the flow header and I enabled it. When I did I saw type errors because some of the wasm functions return only inside if blocks and have implicit void returns. So I put the `?` on the type and had to provide a default. 🙁  I didn't think much of it at the time but that could be problematic.

src/utils/editor/tests/editor.spec.js
```diff
@@ -0,0 +1,263 @@
...
+      column: 31
+    });
+    expect(editor.codeMirror.coordsChar).toHaveBeenCalledWith({
+      left: 30,
+      top: 60
+    });
+  });
+});
+
+describe("forEachLine", () => {
```
jasonLaster: i like the individual describes

src/utils/editor/index.js
```diff
@@ -99,14 +99,12 @@ export function toSourceLocation(
   location: EditorPosition
 ): AstPosition {
   return {
-    line: toSourceLine(sourceId, location.line),
+    line: toSourceLine(sourceId, location.line || 0),
```
jasonLaster: whats the rationale with `|| 0` again?

wldcordeiro:
Dropping the WIP status and I'll be following this up with additional PRs.

src/utils/editor/index.js
yurydelendik: It will be really hard to get rid of this pesky `|| 0`. I introduced `EditorPosition` when adding wasm capability, so intent was to not have `line: ?number` there. Can we investigate this and make sure wasm related utilites will not have possible void return?

jasonLaster: @yurydelendik what would it take to investigate?

can you outline the STR w/ fib?
http://firefox-dev.tools/debugger-examples/examples/wasm/fib/fib.index.html

do you need to poke around w/ this PR?

wldcordeiro: @yurydelendik the only problem was that you must have missed the Flow header in the file where you defined `EditorPosition` because when I added it the errors appeared. 🙁 I could revert the changes including the header to avoid this here. But I think it is problematic that we have void returns.

jasonLaster:
[Screen Capture 1-24-18, 12.30.02 PM.mp4](https://shipusercontent.com/bd3a1ce86fe454533373fb9e62fdd674/Screen%20Capture%201-24-18%2C%2012.30.02%20PM.mp4)

wldcordeiro:
😕 @jasonLaster I am confused.

jasonLaster:
haha - i'm testing retro clip :)

Issue #3279, opened 2017-07-07
[Tabs] Decouple tabs functionality from sources


jasonLaster:
We currently show the current selected source tab as active.

We should show a new tab when you're doing a project text search. And it should be a bit active.

This is a bit challenging because tabs are currently associated with sources and URLs

bomsy:
A couple of thoughts about it
* Use the HOC pattern for tabs (i.e split to `Tabs.js` and `Tab.js`)
* Use more generic naming  i.e `SourceTabs` to `Tabs`
* Separate tab actions from source actions file
* Have separate reducers for the tabs
* Remove dependencies on source urls

codehag:
Is this currently in progress?

bomsy:
no i've not started it yet. was going to pick it up next..

codehag:
Wanted to add @ametzler's observations here as well:

"the Tab thing seems to affect more that the storybook. If I click the the '+' symbol and search the project, it does not open a new tab but loads the file in the current one"

jasonLaster:
https://github.com/devtools-html/debugger.html/pull/4199

jasonLaster:
The PR has all but the source tests passing in Jest. The next step is to finish up the jest tests and then focus on the mochitests.

PR #5133, opened 2018-01-18
[WIP] Refactor tabs

bomsy:
Associated Issue: #3279

Here's the Pull Request Doc

### Summary of Changes

* Add `Tab.js`
* Add tab utility


src/utils/tabs.js
```diff
@@ -0,0 +1,28 @@
+/*
```
jasonLaster: missing flow and the copyright comment

jasonLaster:
so much cleaner!

src/components/Editor/Tab.js
```diff
@@ -0,0 +1,217 @@
+import React, { PureComponent } from "react";
```
wldcordeiro: Flow + copyright header missing.

src/components/Editor/Tab.js
```diff
@@ -0,0 +1,217 @@
...
+  closeTabs: (List<string>) => void,
+  togglePrettyPrint: string => void,
+  showSource: string => void,
+  source: SourceRecord,
+  sourceTabsMetaData: {
+    [key: string]: SourceMetaDataMap
+  }
+};
+
+export default class Tab extends PureComponent<Props, State> {
```
wldcordeiro: I like that tabs are now their own component!

src/components/Editor/Tabs.js
```diff
@@ -337,108 +147,49 @@ class SourceTabs extends PureComponent<Props, State> {
...
-        onContextMenu={e => this.onTabContextMenu(e, source.get("id"))}
-        title={getFileURL(source.toJS())}
-      >
-        {sourceAnnotation}
-        <div className="filename">{filename}</div>
-        <CloseButton
-          handleClick={onClickClose}
-          tooltip={L10N.getStr("sourceTabs.closeTabButtonTooltip")}
-        />
+        {tabSources.map((source, index) => {
```
wldcordeiro: I think this looks a little nicer as

```jsx
{tabSources.map((source, index) => (
  <Tab
    key={index}
    source={source}
    tabSources={tabSources}
    selectedSource={selectedSource}
    closeTab={closeTab}
    togglePrettyPrint={togglePrettyPrint}
    showSource={showSource}
    sourceTabsMetaData={sourceTabsMetaData}
    selectSource={selectSource}
  />
))}
```

src/components/Editor/Tab.js
```diff
@@ -0,0 +1,217 @@
...
+  constructor() {
```
wldcordeiro: If we have a constructor you should take the props argument and pass it to super. But in this case since the body is empty save for the super call we should remove the constructor.

src/components/Editor/Tab.js
```diff
@@ -0,0 +1,217 @@
...
+    const tabURLs = tabSources.map(t => t.get("url"));
+    const otherTabURLs = otherTabs.map(t => t.get("url"));
+
+    if (!tabID) {
+      return;
+    }
+
+    const isPrettySource = isPretty(tabID);
+
+    const closeTabMenuItem = {
```
wldcordeiro: Feel like these menu items could be moved to a utility file for most of their body, something like

```js
export const tabMenuItems = {
  closeTab: {
    id: "node-menu-close-tab",
    label: L10N.getStr("sourceTabs.closeTab"),
    accesskey: L10N.getStr("sourceTabs.closeTab.accesskey"),
    disabled: false
  },
  ...
};
```

Then this way when you're building up the list here all you're adding is the `click` properties and can spread in the hard coded stuff from elsewhere.

src/components/Editor/Tab.js
wldcordeiro: One thing though. This component takes lots of props that are from Redux but are being passed through in `src/components/Editor/Tabs`, maybe instead of default exporting the class we bring in redux `connect()`.

src/components/Editor/Tab.js
jasonLaster: yep - lets do this in a separate PR though to keep this smaller

src/components/Editor/Tab.js
bomsy: good idea!

Issue #5140, opened 2018-01-18
[Breakpoints] Syncing fails w/ line `null` 


jasonLaster:
I found while debugging that breakpoints sometimes fail to sync. The reason for this is 

a) they try to set a breakpoint at line `null`
b) the ast location is off
c) the original location of the bp does not have a mapping...

jasonLaster:
Okay, looks like the bug is actually with disabled breakpoints and ast locations:

1. clone https://github.com/devtools-html/perf.html/
2. add a breakpoint in Home.js on line 92 in constructor
3. disable the breakpoint
4. add a bunch of empty lines above constructor
5. reload perf

ER: you see the breakpoint moved

AR: 
1. it is moved
2. localstorage shows that we have a new pending breakpoint in the correct location, but we didn't remove the old one...

PR #5146, opened 2018-01-19
Fix syncing disabled breakpoints

jasonLaster:
Associated Issue: #5140

### Summary of Changes

This fixes a bug where we were not removing old disabled breakpoints when the breakpoint moved on sync.

I'm working on a follow up PR where I cleanup the unit tests and increase the coverage

### DIFF
--- a/src/actions/breakpoints/tests/syncing.js
+++ b/src/actions/breakpoints/tests/syncing.js
@@ -235,4 +235,42 @@ describe("reloading debuggee", () => {
     expect(findScopeByName).toHaveBeenCalled();
     expect(update).toMatchSnapshot();
   });
+
+  it("syncs with changed source and an existing disabled BP", async () => {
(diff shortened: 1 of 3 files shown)


jasonLaster:
just added an STR https://github.com/devtools-html/debugger.html/issues/5140#issuecomment-359078036

src/actions/breakpoints/tests/syncing.js
```diff
@@ -235,4 +235,59 @@ describe("reloading debuggee", () => {
     expect(findScopeByName).toHaveBeenCalled();
     expect(update).toMatchSnapshot();
   });
+  it("syncs with changed source and an existing disabled BP", async () => {
```
lukaszsobek: Maybe put these two tests inside their own describe in order to separate them, put the warning message in the describe and avoid repeating the complicated test description.

Issue #5131, opened 2018-01-17
Copy Function Off by 1 Line


darkwing:
STR:

1.  Open the following URL in the debugger:  https://davidwalsh.name/wp-content/themes/punky/js/main.js

2.  Right-click line 2 (`function t()`) and select "Copy funciton"

3.  Error:  `TypeError: getFunctionLocation(...) is null`

I changed line 113 of `EditorMenu.js` to `getFunctionLocation(line + 1)` and the error went away.

darkwing:
A few things to check out:

1.  Is this only happening on pretty-printed resources?

2.  Do we have a larger off-by-one-line issue in the debugger?

PR #5147, opened 2018-01-19
Fix 5131 - Use correct line for copying function

darkwing:
Associated Issue: #5131

### Summary of Changes

* Changing the line number of the copy functionality to prevent error described in the issue

### Test Plan

Open a JS file, right-click on a function line and select `Copy function`.  There should be no error.

### DIFF
--- a/src/components/Editor/EditorMenu.js
+++ b/src/components/Editor/EditorMenu.js
@@ -10,7 +10,10 @@ import { isOriginalId } from "devtools-source-map";
 import { copyToTheClipboard } from "../../utils/clipboard";
 import { findFunctionText } from "../../utils/function";
 import { findClosestScope } from "../../utils/breakpoint/astBreakpointLocation";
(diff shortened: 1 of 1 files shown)


src/components/Editor/EditorMenu.js
```diff
@@ -110,7 +110,7 @@ function getMenuItems(
     accesskey: copyFunctionKey,
     disabled: !functionText,
     click: () => {
-      const { location: { start, end } } = getFunctionLocation(line);
+      const { location: { start, end } } = getFunctionLocation(line + 1);
```
jasonLaster: i think we have a helper function like `toEditorLine` or `toSourceLine`

darkwing:
Would that also be used on line 106?  Is there a difference?

jasonLaster:
yeah - probably

darkwing:
Updated to use `toSourceLine`!

Issue #5086, opened 2018-01-11
[QuickOpen] go to line shows matching files


jasonLaster:
It looks like queries that begin with `:` will show files that match the remaining query.

<img src='https://shipusercontent.com/ab6a0630252ae38db3f1b0c21357a668/Screen%20Shot%202018-01-11%20at%2011.29.45%20AM.png' title='Screen Shot 2018-01-11 at 11.29.45 AM.png' width=640>

wldcordeiro:
So if we have the goto line prefix we should error if the following stuff doesn't match the `line:column` format (column is optional)

PR #5161, opened 2018-01-21
[QuickOpen] Fix for go to line showing matching files

atwalg2:
Associated Issue: #5086

### Summary of Changes

* Added case for preventing goto from performing a search

### Test Plan

Tested by typing alphabetical characters after ":", and observed the behavior to work how it should.

before: 
image

after:
image

### DIFF
--- a/src/components/QuickOpenModal.js
+++ b/src/components/QuickOpenModal.js
@@ -102,7 +102,9 @@ export class QuickOpenModal extends Component<Props, State> {
       const results = this.props.sources;
       return this.setState({ results });
     }
-
+    if (this.isGotoQuery()) {
+      return this.setState({ results: [] });
+    }
     if (this.isGotoSourceQuery()) {
       const [baseQuery] = query.split(":");
       const results = filter(this.props.sources, baseQuery);
@@ -149,9 +151,6 @@ export class QuickOpenModal extends Component<Props, State> {
 
     if (this.isShortcutQuery()) {
       return this.showShortcuts(query);
-    }
-    if (this.isGotoQuery()) {
-      return this.searchSources(null);
     }
     return this.searchSources(query);
   };
--- a/src/components/QuickOpenModal.js
+++ b/src/components/QuickOpenModal.js
@@ -152,6 +152,9 @@ export class QuickOpenModal extends Component<Props, State> {
     if (this.isShortcutQuery()) {
       return this.showShortcuts(query);
     }
+    if (this.isGotoQuery()) {
+      return this.searchSources(null);
+    }
     return this.searchSources(query);
   };
 
--- a/src/components/QuickOpenModal.js
+++ b/src/components/QuickOpenModal.js
@@ -102,9 +102,6 @@ export class QuickOpenModal extends Component<Props, State> {
       const results = this.props.sources;
       return this.setState({ results });
     }
-    if (this.isGotoQuery()) {
-      return this.setState({ results: [] });
-    }
     if (this.isGotoSourceQuery()) {
       const [baseQuery] = query.split(":");
       const results = filter(this.props.sources, baseQuery);
@@ -141,6 +138,10 @@ export class QuickOpenModal extends Component<Props, State> {
   };
 
   updateResults = (query: string) => {
+    if (this.isGotoQuery()) {
+      return;
+    }
+
     if (query == "") {
       return this.showTopSources();
     }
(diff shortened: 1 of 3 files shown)


jbhoosreddy:
@atwalg2 instead of making the searches and not showing the results, I think it would be better if we never make a search in this case, to start with. could you investigate that approach?

jbhoosreddy:
could you add a small test case in QuickOpenModal.spec.js?

atwalg2:
sorry for taking so long, I've added a test case. It's my first attempt at one fyi.

src/components/tests/QuickOpenModal.spec.js
```diff
@@ -163,4 +163,20 @@ describe("QuickOpenModal", () => {
       .simulate("change", { target: { value: "@someFunc" } });
     expect(filter).toHaveBeenCalledWith([], "someFunc", { key: "value" });
   });
+
+  test("Simple goto search query = :abc & searchType = goto", () => {
```
wldcordeiro: Did you run this test locally? It doesn't seem to have a snapshot and normally this would modify the snapshot.

src/components/tests/QuickOpenModal.spec.js
```diff
@@ -163,4 +163,20 @@ describe("QuickOpenModal", () => {
...
+        query: ":abc",
+        searchType: "goto",
+        symbols: {
+          functions: [],
+          variables: []
+        }
+      },
+      "mount"
+    );
+    expect(wrapper).toMatchSnapshot();
```
wldcordeiro: I'd add an assertion like `expect(wrapper.state().results).toEqual([])`

src/components/tests/QuickOpenModal.spec.js
atwalg2: I just added the snapshot right now.

src/components/tests/QuickOpenModal.spec.js
atwalg2: add right underneath the toMatchSnapshot() line or replace it?

src/components/QuickOpenModal.js
```diff
@@ -95,7 +95,9 @@ export class QuickOpenModal extends Component<Props, State> {
       const results = this.props.sources;
       return this.setState({ results });
     }
-
+    if (this.isGotoQuery()) {
```
wldcordeiro: You could avoid the setState entirely by changing line 91 to this: 

```js
    if (!this.isGotoQuery() && (nowEnabled || queryChanged)) {
```
and line 248 to:

```js
    if ((this.isSymbolSearch() && noSource) || this.isGotoQuery()) {
```

wldcordeiro:
Condensing my review notes here.

```patch
index 1fde7423..e7f2693d 100644
@@ -88,7 +88,7 @@ export class QuickOpenModal extends Component<Props, State> {
 
     const nowEnabled = !prevProps.enabled && this.props.enabled;
     const queryChanged = prevProps.query !== this.props.query;
-    if (nowEnabled || queryChanged) {
+    if (!this.isGotoQuery() && (nowEnabled || queryChanged)) {
       this.updateResults(this.props.query);
     }
   }
@@ -245,7 +245,7 @@ export class QuickOpenModal extends Component<Props, State> {
     const { selectedSource, setQuickOpenQuery } = this.props;
     setQuickOpenQuery(e.target.value);
     const noSource = !selectedSource || !selectedSource.get("text");
-    if (this.isSymbolSearch() && noSource) {
+    if ((this.isSymbolSearch() && noSource) || this.isGotoQuery()) {
       return;
     }
     this.updateResults(e.target.value);
```

src/components/tests/QuickOpenModal.spec.js
wldcordeiro: Under it is best then you get two tests. 😄

wldcordeiro:
Something I'd like to follow-up the logic for showing the error emoji may need tweaking as it shouldn't be an error when we're entering a goto query. I'll open an issue for that as a separate task.

atwalg2:
travis-ci keeps failing. Have any idea why this might be happening?

atwalg2:
I was trying to update my forked repo, and now its adding all these other commits to this PR.

How can I fix this.

jasonLaster:
travis is currently broken...

this is helpful :)
https://github.com/k88hudson/git-flight-rules/blob/master/README.md

atwalg2:
@jasonLaster Thank you, I got it fixed now.

jasonLaster:
Hey @atwalg2 I see some flow issues in the circle run.

https://flow.org/

wldcordeiro:
Fixed the flow issues. Should be good to go. Just waiting on the CI runs.

Issue #5151, opened 2018-01-19
[Scopes] are missing padding


jasonLaster:
It looks like the scopes pane lost its padding after we refactored the reps and tree component

<img src='https://shipusercontent.com/2b16536c141786dc575d7b26d9660ead/Screen%20Shot%202018-01-19%20at%204.03.34%20PM.png' title='Screen Shot 2018-01-19 at 4.03.34 PM.png' width=324>

lukaszsobek:
Fixed, closing.

PR #5163, opened 2018-01-21
[Scopes]-fixed missing padding in scopes tree

darrylpargeter:
Associated Issue: #5151 

### Summary of Changes
* Added a 4px padding to the left-hand side of ```scopes-list``` class

### Test Plan
* A visual check that the Block node aligns with the Parent Scopes node.

Before added padding
4px padding added

### DIFF
--- a/src/components/SecondaryPanes/Scopes.css
+++ b/src/components/SecondaryPanes/Scopes.css
@@ -29,6 +29,10 @@
   overflow: auto;
 }
 
+.scopes-list {
+  padding-left: 4px;
+}
+
 .scopes-list .function-signature {
   display: inline-block;
 }


src/components/SecondaryPanes/Scopes.css
```diff
@@ -29,6 +29,10 @@
   overflow: auto;
 }
 
+.scopes-list {
+  padding-left: 4px;
```
nchevobbe: not sure how we deal with LTR languages, but `padding-inline-start` might be a bit better

lukaszsobek: Oh, that's interesting!

jasonLaster: we should lint for `padding-left` `padding-right`

lukaszsobek: @nchevobbe @jasonLaster we use https://github.com/gasolin/postcss-bidirection as way to handle ltr & rtl which polyfills ...start into ...left or ...right depending on the direction.

Issue #5149, opened 2018-01-19
[Sources] the selection width is clipped


jasonLaster:
I think that upgrading to reps and the new tree component caused the sources tree node to be clipped.

this seems like a good fix

```diff
diff --git a/src/components/PrimaryPanes/Sources.css b/src/components/PrimaryPanes/Sources.css
index 0e49ca0..c32840c 100644
--- a/src/components/PrimaryPanes/Sources.css
+++ b/src/components/PrimaryPanes/Sources.css
@@ -33,6 +33,7 @@

 .sources-list .managed-tree .tree .node {
   padding: 0px 0px 0px 3px;
+  width: 100%;
 }
 ```

![](http://g.recordit.co/HMfj8CuYRn.gif)

calebt97:
Hey my name is Caleb (noob for short). I'd love to help you!

lukaszsobek:
@calebt97 Hey Caleb, we'd love your help. I think this issue is pretty much covered though, so maybe take a look at some of our other good first issues here: https://github.com/devtools-html/debugger.html/labels/good%20first%20issue and pick one you like.

lukaszsobek:
Fixed, closing.

PR #5167, opened 2018-01-21
[Sources] - fix paddings in sources tree and outline list

martinfojtik:
### Summary of Changes

* fix height in sources tree, there was blue line under selected item 
* unify padding around items in outline list
* and fix https://github.com/devtools-html/debugger.html/issues/5149

* sources tree BEFORE
firefox-tree
* sources tree AFTER
firefox-tree-after
* outline list BEFORE
firefox-outline
* outline list AFTER
firefox-outline-after

### DIFF
--- a/src/components/shared/ManagedTree.css
+++ b/src/components/shared/ManagedTree.css
@@ -12,6 +12,9 @@
   white-space: nowrap;
   overflow: auto;
   min-width: 100%;
+
+  display: grid;
+  align-content: start;
 }
 
 .managed-tree .tree button {
--- a/src/components/PrimaryPanes/Outline.css
+++ b/src/components/PrimaryPanes/Outline.css
@@ -17,7 +17,7 @@
 
 .outline-list {
   list-style-type: none;
-  width: 100%;
+  flex: 1 0 100%;
   padding: 10px 0px;
   margin: 0;
 }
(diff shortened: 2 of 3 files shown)


lukaszsobek:
An after image is always nice, as it allows to compare changes visually ;)

martinfojtik:
@lukaszsobek sure,I added screenshots to description

martinfojtik:
@lukaszsobek I fixed right paddings in lists and finded solution for https://github.com/devtools-html/debugger.html/issues/5149

src/components/shared/ManagedTree.css
```diff
@@ -12,6 +12,9 @@
   white-space: nowrap;
   overflow: auto;
   min-width: 100%;
+
+  display: grid;
+  align-content: start;
```
jasonLaster: what does this do?

martinfojtik: this will fix bug https://github.com/devtools-html/debugger.html/issues/5149, basically, this will extend items in the tree to full width. I don't know why, but `width: 100%` or `display: flex` was not able to fix the issue.

lukaszsobek: On the current update of master the line already seems to be the full width, so not sure if this is needed.

martinfojtik: The problem appears when you have long and short short names on the list, and try to vertically scroll, as you can see in the gif.
devtools-source

jasonLaster:
Thanks for the fix! Looks great

jasonLaster:
Hey @martinfojtik, I'm sorry to do this to you but I had to revert your commit earlier today because it was causing an issue with travis CI.

Earlier this week we had false positives with CI, where the tests were failing but not reporting that.

Would you be able to re-open the PR and we can sort out the problem?

martinfojtik:
Hi, @jasonLaster , where can I find more info about the problem? How do i re-open this PR?

jasonLaster:
you revert the reverted commit and open a new PR. My theory is that display grid was throwing off the linux version of firefox. will test that now in a new PR

PR #5168, opened 2018-01-21
Disable CSS auto-prefixes for old IE

martinfojtik:
### Summary of Changes

I think a majority of user use modern browsers for running this debugger, so some prefixes for really old browsers are not needed.

### DIFF
--- a/postcss.config.js
+++ b/postcss.config.js
@@ -34,7 +34,11 @@ module.exports = ({ file, options, env }) => {
   return {
     plugins: [
       require("postcss-bidirection"),
-      require("autoprefixer"),
+      require("autoprefixer")({
+				browsers: ['last 2 Firefox versions', 'last 2 Chrome versions'],
+        flexbox: false,
+        grid: false
+			}),
       require("postcss-class-namespace")(),
       mapUrl(mapUrlDevelopment)
     ]


lukaszsobek:
What would you say were the benefits of the change?

martinfojtik:
@lukaszsobek I think there will be a very very tiny increase in building speed, smaller build size. And there will be no old prefixes, which are not needed for newest browsers.

postcss.config.js
```diff
@@ -34,7 +34,11 @@ module.exports = ({ file, options, env }) => {
   return {
     plugins: [
       require("postcss-bidirection"),
-      require("autoprefixer"),
+      require("autoprefixer")({
+				browsers: ['> 3%', 'not ie <= 10'],
```
jasonLaster: Haha yeah. Let’s do latest chrome and latest Firefox

martinfojtik: should I change this to `['last 2 Firefox versions', 'last 2 Chrome versions']` ?

note: browser list is there: https://github.com/ai/browserslist#queries

jasonLaster: sure

martinfojtik: done

postcss.config.js
```diff
@@ -34,7 +34,11 @@ module.exports = ({ file, options, env }) => {
   return {
     plugins: [
       require("postcss-bidirection"),
-      require("autoprefixer"),
+      require("autoprefixer")({
+				browsers: ['last 2 Firefox versions', 'last 2 Chrome versions'],
```
bomsy: should this be "browserslist: [ ..." rather than "browsers: [ ..."?

martinfojtik: it should be `browsers`

https://github.com/postcss/autoprefixer#options

https://github.com/postcss/autoprefixer/blob/83de1e82c0d418aac4a5fa0faebf71a354babdc3/lib/autoprefixer.js#L56

bomsy: cool nice work!

wldcordeiro: While we technically only support Firefox and Chrome you could just make this the 'last 2 version' value instead of explicitly just Firefox/Chrome.

jasonLaster:
Looks good to me. I prefer being explicit because who knows what other features other browsers are missing :)

Issue #5172, opened 2018-01-23
[fetchScopes] Tries to read get from possibly empty object 


lukaszsobek:
Pause.spec.js is calling fetchScopes without supplying a Map for the function to read from which results in an error when running jest. Would be nice to verify the object exists before trying to get from it.

lukaszsobek:
Closing in favor of issue with broader scope.

PR #5173, opened 2018-01-23
Adds check to see if object exists before trying to get from it

lukaszsobek:
Associated Issue: #5172

pause.spec.js was throwing errors during the test

### Before
screenshot-20180123094324-612x759

### After
screenshot-20180123094613-534x172

### DIFF
--- a/src/actions/pause/fetchScopes.js
+++ b/src/actions/pause/fetchScopes.js
@@ -64,13 +64,13 @@ export function fetchScopes() {
       frame.generatedLocation.sourceId
     );
 
-    if (generatedSourceRecord.get("isWasm")) {
+    if (generatedSourceRecord && generatedSourceRecord.get("isWasm")) {
       return;
     }
 
     const sourceRecord = getSource(getState(), frame.location.sourceId);
 
-    if (sourceRecord.get("isPrettyPrinted")) {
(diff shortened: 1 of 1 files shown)


src/actions/pause/fetchScopes.js
```diff
@@ -64,13 +64,13 @@ export function fetchScopes() {
...
     );
 
-    if (generatedSourceRecord.get("isWasm")) {
+    if (generatedSourceRecord && generatedSourceRecord.get("isWasm")) {
       return;
     }
 
     const sourceRecord = getSource(getState(), frame.location.sourceId);
 
-    if (sourceRecord.get("isPrettyPrinted")) {
```
jasonLaster: I'd like to figure out why these sources aren't around when we're doing `fetchScopes`

jasonLaster:
I feel like this change makes it okay to not have source objects. When that is a requirement at this stage.

jasonLaster:
I'm not sure what the best way to hunt down the "WHY" though..

lukaszsobek:
@jasonLaster We can start with the pause.spec.js as this was where the bug first appeared. There was something distinctly odd about the way the test was running, I'll try to recreate what it was once I'm logged over to the workstation.

jasonLaster:
awesome

lukaszsobek:
Closing in favor of issue with broader scope.

PR #5175, opened 2018-01-23
Add basic UI for rewinding when server indicates it can rewind

bhackett1024:
### Summary of Changes

- Allow the debugger client to keep track of whether the server can rewind, as indicated by the server via the 'canRewind' property of the traits sent while attaching.

- Modify the debugger UI when interacting with a rewindable server, adding rewind and reverse-step buttons, and always showing the pause button (handy to have because a replaying process will zip forward to the end of the replay as fast as it can).

- Add some checks for the presence of a frame when paused.  Recording/replaying processes may pause when there are no JS frames on the stack.

### DIFF
--- a/src/client/firefox.js
+++ b/src/client/firefox.js
@@ -54,7 +54,8 @@ export async function onConnect(connection: any, actions: Object): Object {
   // bfcache) so explicity fire `newSource` events for all returned
   // sources.
   const sources = await clientCommands.fetchSources();
-  await actions.connect(tabTarget.url);
+  const traits = tabTarget.activeTab ? tabTarget.activeTab.traits : null;
+  await actions.connect(tabTarget.url, traits && traits.canRewind);
   await actions.newSources(sources);
 
   // If the threadClient is already paused, make sure to show a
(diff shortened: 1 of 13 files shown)


jasonLaster:
this will be tough to test as its the firefox support is not there... we might be able to simulate some of it in a redux JEST unit test though

jasonLaster:
I'll look soon!

jasonLaster:
sorry, i'm getting held up with some other work...

src/client/firefox.js
```diff
@@ -54,7 +54,8 @@ export async function onConnect(connection: any, actions: Object): Object {
   // bfcache) so explicity fire `newSource` events for all returned
   // sources.
   const sources = await clientCommands.fetchSources();
-  await actions.connect(tabTarget.url);
+  const traits = tabTarget.activeTab ? tabTarget.activeTab.traits : null;
+  await actions.connect(tabTarget.url, traits && traits.canRewind);
```
jasonLaster: in the future we may want to pass the traits through `actions.connect(tabTarget.url, traits);` but for now it is fine.

jasonLaster:
closed and opening a new rebased PR. In the future, if you open your PR w/ a non-master branch i can push directly to that.

PR #5181, opened 2018-01-24
Cleanup tabs

bomsy:
* Move part of the items to a util


src/components/Editor/Tab.js
```diff
@@ -90,82 +70,55 @@ class Tab extends PureComponent<Props> {
...
     if (!isPrettySource) {
-      items.push({ item: showSourceMenuItem });
-      items.push({ item: prettyPrint });
+      items.push({
+        item: { ...tabMenuItems.showSource, click: () => showSource(tab) }
+      });
+      items.push({
+        item: {
+          ...tabMenuItems.prettyPrint,
+          click: () => showSource(tab)
```
bomsy: this should be click: () => togglePrettyPrint(sourceTab.get("id"))

src/utils/tabs.js
```diff
@@ -61,3 +61,51 @@ export function getSourceAnnotation(
     return <img className="blackBox" />;
   }
 }
+
+export function getTabMenuItems() {
+  const items = {
```
bomsy: maybe return { ... }

wldcordeiro: Since it's always the same object you could skip the function and just

```js
export const tabMenuItems = {
...
};
```

wldcordeiro:
Sketch of an idea...


```js
const handlers = {
  closeTab: {
    click: () => closeTab(sourceTab.get("url"))
  },
  // ...
};

const items = {
  closeTab: {
    id: "node-menu-close-tab",
    label: L10N.getStr("sourceTabs.closeTab"),
    accesskey: L10N.getStr("sourceTabs.closeTab.accesskey"),
    disabled: false
  },
  // ...
};

function mapTabItems(handlers, items) {
  return Object.entries(items).reduce((acc, [key, item]) => {
    return [...acc, { item, ...handlers[key] }]
  }, []);
}

jasonLaster:
so... frankly i'm okay with a longer menu function. It's cool to try and consolidate it, but lets not add too much indirection in

Issue #5138, opened 2018-01-18
[Stepping] aligns line every time


jasonLaster:
When we refactored scrolling last month, we removed a check that used to ensure that we'd only scroll if the line were not visible. We should bring that back because now it's very jump :runner:

STR: 
1. go to [todomvc](http://devtools-html.github.io/debugger-examples/examples/todomvc/)
2. add a breakpoint in todoview.js#33
3. add a todo in the app
4. start stepping

notice how it always scrolls!

![](http://g.recordit.co/sZG97dpyZe.gif)

### GIST

```diff
diff --git a/src/utils/editor/index.js b/src/utils/editor/index.js
index 661b01b..9e2d5d8 100644
--- a/src/utils/editor/index.js
+++ b/src/utils/editor/index.js
@@ -91,7 +91,26 @@ export function scrollToColumn(codeMirror: any, line: number, column: number) {
   const centeredX = Math.max(left - scroller.offsetWidth / 2, 0);
   const centeredY = Math.max(top - scroller.offsetHeight / 2, 0);
 
-  codeMirror.scrollTo(centeredX, centeredY);
+  if (!isVisible(codeMirror, line, column)) {
+    // we probably also want to check to see if the column is visible
+    codeMirror.scrollTo(centeredX, centeredY);
+  }
+}
+
+// this approach comes from alignLine
+// we could also check the top / left we have above or some other strategy...
+function isVisible(codeMirror, line, column) {
+  const editorClientRect = codeMirror.getWrapperElement().getBoundingClientRect();
+
+  const from = codeMirror.lineAtHeight(editorClientRect.top, "page");
+  const to = codeMirror.lineAtHeight(
+    editorClientRect.height + editorClientRect.top,
+    "page"
+  );
+
+  const isLineVisible = line >= to && line <= from;
+  const isColumnVisible = false;
+  return isLineVisible && isColumnVisible;
 }
 
 export function toSourceLocation(
```

PR #5182, opened 2018-01-24
Change horizontal and vertical realignment to only occur when line is out of bounds.

mmcote:
Associated Issue: #5138 

Here's the Pull Request Doc

### Summary of Changes

-Check if the line is currently visible.
-Only realign the line to be centered if the line is out of bounds.

### Test Plan
N/A

Feel free to improve it!

Before: 
- As can be seen before, on every step in the debugger the editor was being realigned horizontally and vertically.
beforestepdebug

After:
- After the changes the editor will only be realigned when the line goes out of the bounds of the editor.
afterstepper

Before:
- Negative on the margin on the right side of the scroll area was increasing the clientWidth of the scrollArea.
screen shot 2018-01-23 at 6 49 05 pm

After:
- Removed extra margin on the right side of the scroll area.
screen shot 2018-01-23 at 6 49 21 pm

Also added some necessary functions to be mocked in Editor.spec.js, that are used in the isVisible function.

### DIFF
--- a/src/components/Editor/Editor.css
+++ b/src/components/Editor/Editor.css
@@ -126,6 +126,10 @@ html[dir="rtl"] .editor-mount {
   height: 100%;
 }
 
+.CodeMirror-scroll {
+  margin-right: 0px;
+}
+
 .editor-wrapper .editor-mount {
   width: 100%;
   background-color: var(--theme-body-background);
--- a/src/components/Editor/Editor.css
+++ b/src/components/Editor/Editor.css
@@ -126,10 +126,6 @@ html[dir="rtl"] .editor-mount {
   height: 100%;
 }
 
-.CodeMirror-scroll {
-  margin-right: 0px;
-}
-
 .editor-wrapper .editor-mount {
   width: 100%;
   background-color: var(--theme-body-background);
--- a/src/utils/editor/index.js
+++ b/src/utils/editor/index.js
@@ -108,7 +108,7 @@ function isVisible(codeMirror: any, top: number, left: number) {
   const inXView = withinBounds(
     left,
     scrollArea.left,
-    scrollArea.left + scrollArea.clientWidth - charWidth
+    scrollArea.left + (scrollArea.clientWidth - 30) - charWidth
   );
 
   const fontHeight = codeMirror.defaultTextHeight();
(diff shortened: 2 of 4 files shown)


src/utils/editor/index.js
```diff
@@ -87,11 +87,37 @@ export function scrollToColumn(codeMirror: any, line: number, column: number) {
...
+    scrollArea.left,
+    scrollArea.left + scrollArea.clientWidth - charWidth
+  );
+
+  const fontHeight = codeMirror.defaultTextHeight();
+  const inYView = withinBounds(
+    top,
+    scrollArea.top,
+    scrollArea.top + scrollArea.clientHeight - fontHeight
+  );
```
jasonLaster: how does this work?

src/components/Editor/Editor.css
```diff
@@ -126,6 +126,10 @@ html[dir="rtl"] .editor-mount {
   height: 100%;
 }
 
+.CodeMirror-scroll {
+  margin-right: 0px;
+}
```
jasonLaster: why is this needed?

mmcote: When I was using the codeMirror, I noticed that the clientWidth that was being returned from the getScrollInfo() was not matching what was seen in the inspector. It was off by 30px. I found that by default it had a margin-right: -30px. Although once I removed this extra negative margin, it properly represented the scroll area, the bounds of the scroll view.

src/utils/editor/index.js
mmcote: I did this just because if your scroll view area was not exactly a multiple of the font height then the last line may be cut off when you go to debug the last line, as it will still stay in view. So instead I said that if the top is below the top of the last lines fontHeight just realign and then you can see the whole line.

src/components/Editor/Editor.css
lukaszsobek: @jasonLaster @mmcote you might find this discussion interesting: https://github.com/codemirror/CodeMirror/issues/2984
not the same margin, but maybe the same reason.

mmcote: Thank you for bringing this to my attention :) I will make some corresponding changes when I get a chance later on in the day.

Issue #4777, opened 2017-11-23
[Tabs] Show framework logos


jasonLaster:
We just started showing React logos in the editor tabs for react components, but there's no reason we cant generalize it to other frameworks:

### checklist 

- [ ] React
- [ ] Ember
- [ ] Angular
- [ ] Vue

...

jasonLaster:
```diff
diff --git a/src/workers/parser/frameworks.js b/src/workers/parser/frameworks.js
index d5bbfe9..7afa5dc 100644
--- a/src/workers/parser/frameworks.js
+++ b/src/workers/parser/frameworks.js
@@ -4,6 +4,7 @@
 
 import getSymbols from "./getSymbols";
 
+// this is how we currently check for react components
 export function isReactComponent(source) {
   const { imports, classes } = getSymbols(source);
 
@@ -35,3 +36,13 @@ function extendsComponent(classes) {
 
   return result;
 }
+
+export function isFramework(source) {
+  if (isReactComponent(source)) {
+    return "React";
+  }
+
+  if (isAngularComponent(source)) {
+    return "Angular";
+  }
+}
diff --git a/src/workers/parser/tests/fixtures/frameworks/component.js b/src/workers/parser/tests/fixtures/frameworks/component.js
deleted file mode 100644
index 37f3ac4..0000000
--- a/src/workers/parser/tests/fixtures/frameworks/component.js
+++ /dev/null
@@ -1,3 +0,0 @@
-import React, { Component } from "react";
-
-class PrimaryPanes extends Component {}
diff --git a/src/workers/parser/tests/framework.spec.js b/src/workers/parser/tests/framework.spec.js
index 416ac06..07bcd43 100644
--- a/src/workers/parser/tests/framework.spec.js
+++ b/src/workers/parser/tests/framework.spec.js
@@ -5,4 +5,10 @@ describe("Parser.frameworks", () => {
   it("should be a react component", () => {
     expect(isReactComponent(getSource("frameworks/component"))).toBe(true);
   });
+
+  it("should be an angular component", () => {
+    expect(isFramework(getSource("frameworks/angular-component"))).toBe(
+      "Angular"
+    );
+  });
 });
```

codehag:
i understand that this might be an easy patch to do, but i would rather take this out right now. we have a lot of other work and this increases the maintenance burden. Lets try to make performance bugs available rather than features. for this reason i am making it unavailable.

codehag:
[...]

PR #5186, opened 2018-01-24
Set up logic to allow checking for frameworks

lukaszsobek:
Associated Issue: #4777

- [x] Introduces `getFramework()` as parent function for `isReactComponent()` to be able to check for more frameworks.
- [x] Makes framework class the lowercase version of the name assigned in check

### DIFF
--- a/src/utils/tabs.js
+++ b/src/utils/tabs.js
@@ -51,9 +51,15 @@ export function getSourceAnnotation(
   const sourceId = source.get("id");
   const sourceMetaData = getMetaData(sourceId);
 
-  if (sourceMetaData && sourceMetaData.isReactComponent) {
-    return <img className="react" />;
+  const framework =
+    sourceMetaData && sourceMetaData.framework
+      ? sourceMetaData.framework
+      : false;
+
+  if (framework) {
+    return <img className={framework.toLowerCase()} />;
   }
+
   if (isPretty(source)) {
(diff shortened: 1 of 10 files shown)


src/workers/parser/frameworks.js
```diff
@@ -4,7 +4,13 @@
 
 import getSymbols from "./getSymbols";
 
-export function isReactComponent(source) {
+export function whichFramework(source) {
```
jasonLaster: i think i would prefer `getFramework`. We use `get` and `fetch` for most of our query functions

src/utils/tabs.js
```diff
@@ -50,10 +50,15 @@ export function getSourceAnnotation(
 ) {
   const sourceId = source.get("id");
   const sourceMetaData = getMetaData(sourceId);
+  const whichFramework =
+    sourceMetaData && sourceMetaData.whichFrameworkComponent
```
jasonLaster: i think `sourceMetaData.framework` would be simpler here... we can always add more specificity when we need it.

src/workers/parser/frameworks.js
lukaszsobek: Ok, will change it to match.

src/utils/tabs.js
lukaszsobek: Sure, will be much shorter.

PR #5191, opened 2018-01-25
[breakpoints] Use symbols from store

jasonLaster:
Associated Issue: #5096

### Summary of Changes

Replaces a call to the parser getSymbols w/ a fetch from the redux store.

### DIFF
--- a/src/actions/breakpoints/addBreakpoint.js
+++ b/src/actions/breakpoints/addBreakpoint.js
@@ -9,7 +9,7 @@ import {
   getASTLocation,
   assertLocation
 } from "../../utils/breakpoint";
-import { getSource } from "../../selectors";
+import { getSource, getSymbols } from "../../selectors";
 import { getGeneratedLocation } from "../../utils/source-maps";
 
 export default async function addBreakpoint(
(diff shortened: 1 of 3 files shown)


src/actions/breakpoints/addBreakpoint.js
```diff
@@ -50,7 +50,8 @@ export default async function addBreakpoint(
     newGeneratedLocation
   );
 
-  const astLocation = await getASTLocation(sourceRecord, newLocation);
+  const symbols = getSymbols(getState, sourceRecord);
```
bomsy: - `getSymbols` seems not to be defined.
- should be `getSymbols(getState(), sourceRecord)` or `getSymbols(state, sourceRecord)` right?

src/utils/breakpoint/astBreakpointLocation.js
```diff
@@ -48,13 +48,13 @@ export function findClosestScope(functions: Scope[], location: Location) {
 
 export async function getASTLocation(
   source: Source,
+  symbols: Object,
```
bomsy: this might break other places `getASTLocation` is used, maybe we put it on the end
```js
export async function getASTLocation(
   source: Source, 
   location: Location, 
   symbols: Object): Promise<ASTLocation> {

PR #5193, opened 2018-01-25
🔧: Fixing Jest coverage collection and cleaning up config for Jest 22 prep

wldcordeiro:
Associated Issue: #5100 

### Summary of Changes

* Add `collectCoverageFrom` to the Jest config to get a more accurate report
* consolidate test/tests-setup and test/tests-framework as they both were just setting up Jest environment.
* upgrade jest-junit-reporter and jest-localstorage-mock since they had no breaking changes or effect.
* Rename some test files to match the `.spec.js` pattern we've moved to
* Prettier found somethings in the mochis it didn't like? 🤷‍♂️ 

Sad but good news, our coverage dropped from ~62% to ~43%, it's lame to see a lower number but it's more accurate so it's a win in the long run!


wldcordeiro:
@jasonlaster11 the only file I deleted was the tests-framework file but I moved its content to tests-setup since both were only preparing the Jest environment.

jasonLaster:
weird it looks like some specs were deleted

wldcordeiro:
Nah I just renamed a few that were missing the ".spec.js" in their name.

PR #5195, opened 2018-01-25
[blackbox] blackbox sources should be persisted

Fischer-L:
Associated Issue: #3601 

### Summary of Changes
Persist the blackboxed sources 

### Screenshots/Videos
[https://streamable.com/frfoh](https://streamable.com/frfoh)

### DIFF
--- a/src/actions/sources/newSources.js
+++ b/src/actions/sources/newSources.js
@@ -8,7 +8,7 @@
  * Redux actions for the sources state
  * @module actions/sources
  */
-
+import { toggleBlackBox } from "./blackbox";
 import { syncBreakpoint } from "../breakpoints";
 import { loadSourceText } from "./loadSourceText";
 import { togglePrettyPrint } from "./prettyPrint";
@@ -16,6 +16,7 @@ import { selectLocation } from "../sources";
 import { getRawSourceURL, isPrettyURL } from "../../utils/source";
 
 import {
+  getBlackBoxList,
   getSource,
   getPendingSelectedLocation,
   getPendingBreakpointsForSource
@@ -122,6 +123,20 @@ function checkPendingBreakpoints(sourceId) {
   };
 }
 
+function checkBlackBoxedSources(sources: Source[]) {
+  return async ({ dispatch }: ThunkArgs) => {
+    const tabs = getBlackBoxList();
+    if (tabs.length <= 0) {
+      return;
+    }
+    for (const source of sources) {
+      if (tabs.includes(source.url) && !source.isBlackBoxed) {
+        dispatch(toggleBlackBox(source));
+      }
+    }
+  };
+}
+
 /**
  * Handler for the debugger client's unsolicited newSource notification.
  * @memberof actions/sources
@@ -152,6 +167,7 @@ export function newSources(sources: Source[]) {
       dispatch(checkSelectedSource(source));
       dispatch(checkPendingBreakpoints(source.id));
     }
+    dispatch(checkBlackBoxedSources(filteredSources));
 
     return Promise.all(
       filteredSources.map(source => dispatch(loadSourceMap(source)))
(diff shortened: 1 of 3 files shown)


src/utils/prefs.js
```diff
@@ -24,6 +24,7 @@ if (isDevelopment()) {
   pref("devtools.debugger.start-panel-collapsed", false);
   pref("devtools.debugger.end-panel-collapsed", false);
   pref("devtools.debugger.tabs", "[]");
+  pref("devtools.debugger.tabsBlackBoxed", "[]");
```
Fischer-L: This patch follows the logic we save a url list of scripts to dispaly in the `devtools.debugger.tabs` pref, creating another `devtools.debugger.tabsBlackBoxed` pref to save a list of baclboxed scripts.

The reason for another `devtools.debugger.tabsBlackBoxed` pref is as below.
There are some issues if switch to `[ { uri, isBlackBoxed }, ... ]` data structure.

- The backward compatibility: Currently we save `[url, ...]`  in the `devtools.debugger.tabs` pref. So we have to change `[url, ...]` to  `[ { uri, isBlackBoxed }, ... ]`. This is possible by doing at `restoreTabs`[1] when `initialState`[2] but increases some cost at the init state (maybe acceptable since only once).

- Touch selectors and helper functions: Since the tab list structure changes, we have to update `getSourceTabs`, `getSearchTabs`, `getSourcesForTabs` and `getNewSelectedSourceId`. Places rely on these selectors and helper functions as well as rely on the data structure returned by these functions. Risky to touch these parts.

- The case of closing a tab: Assume the following common case,

  1. User opens and blackboxes some_lib.js
  2. User closes some_lib.js and starts debugging
  3. User reloads and continues debugging
  4. While debugging, user expect some_lib.js still blackboxed

  For this case, now there is one more state to save, which a tab is closed BUT blackboxed.
  As a result, have to introduce another `shouldDisplay` property in the tab list so `[ { uri, isBlackBoxed, shouldDisplay }, ... ]`.
  Then while closing a tab[3], we couldn't simply remove a closing tab from the tab list but kept it and set `shouldDisplay = false` if it was blackboxed. The introduction of `shouldDisplay` is not really nice but required for the persisting-blackbox state goal here.

From the point of data grouping, `[ { uri, isBlackBoxed, shouldDisplay }, ... ]` is good. However, have to touch too many places. That is error-prone and regression-risky. Just test some normal cases then quite easy to see errors (not to mention other cases). So I try another approach which save a baclboxed tab list in the `devtools.debugger.tabsBlackBoxed` pref. The good sides for it are far less places to touch, make blackboxing independent from closing a tab, and follow the logic we save a url list of tabs to dispaly in the `devtools.debugger.tabs` pref.

[1] https://github.com/devtools-html/debugger.html/blob/8a1e8b0542af5572be70f2a9ecb6fb9689eb76f3/src/reducers/sources.js#L189
[2] https://github.com/devtools-html/debugger.html/blob/8a1e8b0542af5572be70f2a9ecb6fb9689eb76f3/src/reducers/sources.js#L45
[3] https://github.com/devtools-html/debugger.html/blob/257279690e771aa3ef819cbe367df5caeec48369/src/actions/sources/tabs.js#L47

src/actions/sources/newSources.js
```diff
@@ -122,6 +123,20 @@ function checkPendingBreakpoints(sourceId) {
   };
 }
 
+function checkBlackBoxedSources(sources: Source[]) {
+  return async ({ dispatch }: ThunkArgs) => {
+    const tabs = prefs.tabsBlackBoxed;
```
jasonLaster: we should get the list of blackboxed tabs from the reducer

Fischer-L: > we should get the list of blackboxed tabs from the reducer

@jasonLaster updated, thanks

bomsy:
i'm currently working on refactoring the tabs in a PR which should be coming out in a few days. it should make these quite easy to do

We have two options
- We can land this , and refactor it in that PR
- Or wait for that to land before proceeding

If this PR is good to go, i would say we go for the first option

jasonLaster:
this is a great start. I'd vote for waiting on bomsy's PR and then we can do a small refactor on top of it. Closing for now :) We can re-open soon though

PR #5197, opened 2018-01-25
CI: add mochitests using the debugger and located in non-debugger test suites

juliandescottes:
WIP.

Added all the tests I could find the devtools that rely on the new debugger. In theory they should all be tested as well. If it increases the testing time too much, we can look in more details on try to remove  tests that seem redundant regarding their usage of the debugger.

Let's see what the CI says.

### DIFF
--- a/bin/ci/run-tests.js
+++ b/bin/ci/run-tests.js
@@ -58,7 +58,17 @@ Object.keys(fileSizes).forEach(key => {
 });
 
 const otherTests = [
-  "browser_net_open_in_debugger.js"
+  "devtools/client/framework/test/browser_browser_toolbox_debugger.js",
+  "devtools/client/framework/test/browser_keybindings_01.js",
+  "devtools/client/framework/test/browser_toolbox_getpanelwhenready.js",
+  "devtools/client/inspector/markup/test/browser_markup_links_06.js",
+  "devtools/client/netmonitor/test/browser_net_open_in_debugger.js",
+  "devtools/client/netmonitor/test/browser_net_view-source-debugger.js",
+  "devtools/client/webconsole/new-console-output/test/mochitest/browser_jsterm_autocomplete_in_debugger_stackframe.js",
+  "devtools/client/webconsole/new-console-output/test/mochitest/browser_webconsole_eval_in_debugger_stackframe1.js",
+  "devtools/client/webconsole/new-console-output/test/mochitest/browser_webconsole_eval_in_debugger_stackframe2.js",
+  "devtools/client/webconsole/new-console-output/test/mochitest/browser_webconsole_location_debugger_link.js",
+  "devtools/client/webconsole/new-console-output/test/mochitest/browser_webconsole_stacktrace_location_debugger_link.js",
 ];
 
 console.log(chalk.blue("Running Tests"));


bin/ci/run-tests.js
```diff
@@ -58,7 +58,21 @@ Object.keys(fileSizes).forEach(key => {
 });
 
 const otherTests = [
-  "browser_net_open_in_debugger.js"
+  "devtools/client/framework/test/browser_browser_toolbox_debugger.js",
+  "devtools/client/framework/test/browser_keybindings_01.js",
+  "devtools/client/framework/test/browser_source_map-01.js",
+  "devtools/client/framework/test/browser_source_map-absolute.js",
+  "devtools/client/framework/test/browser_source_map-init.js",
+  "devtools/client/framework/test/browser_source_map-inline.js",
```
jasonLaster: are the source-map tests all using the debugger?

bin/ci/run-tests.js
```diff
@@ -58,7 +58,21 @@ Object.keys(fileSizes).forEach(key => {
...
+  "devtools/client/framework/test/browser_source_map-reload.js",
+  "devtools/client/framework/test/browser_toolbox_getpanelwhenready.js"
+  "devtools/client/netmonitor/test/browser_net_open_in_debugger.js",
+  "devtools/client/netmonitor/test/browser_net_view-source-debugger.js",
+  "devtools/client/webconsole/new-console-output/test/mochitest/browser_jsterm_autocomplete_in_debugger_stackframe.js",
+  "devtools/client/webconsole/new-console-output/test/mochitest/browser_webconsole_eval_in_debugger_stackframe1.js",
+  "devtools/client/webconsole/new-console-output/test/mochitest/browser_webconsole_eval_in_debugger_stackframe2.js",
+  "devtools/client/webconsole/new-console-output/test/mochitest/browser_webconsole_location_debugger_link.js",
+  "devtools/client/webconsole/new-console-output/test/mochitest/browser_webconsole_stacktrace_location_debugger_link.js",
```
jasonLaster: any inspector tests?

bin/ci/run-tests.js
```diff
@@ -58,7 +58,21 @@ Object.keys(fileSizes).forEach(key => {
 });
 
 const otherTests = [
-  "browser_net_open_in_debugger.js"
+  "devtools/client/framework/test/browser_browser_toolbox_debugger.js",
```
jasonLaster: yes!!

bin/ci/run-tests.js
juliandescottes: Ah yes there seems to at least be one: devtools/client/inspector/markup/test/browser_markup_links_06.js not sure how I missed it my first pass. Thanks!

bin/ci/run-tests.js
juliandescottes: They need a toolbox and to do so they open it on the debugger. I assumed they needed to start the debugger for the tests to work. But starting the toolbox on any other tool seems to be sufficient to make the tests pass. I'll remove those.

jasonLaster: I'd rather not include them, because a smaller list is easier...

we can always re-consider if they fail in a release :P

juliandescottes:
Fixed a typo + applied comments

juliandescottes:
We should have at least one failure here, we will need a m-c changeset with my patch from https://bugzilla.mozilla.org/show_bug.cgi?id=1432842 before browser_browser_toolbox_debugger.js can pass.

juliandescottes:
I hadn't updated the PR by commenting out the failing test, so I don't know if the timeout is related to setup issues.

PR #5198, opened 2018-01-25
Travis artifacts

jasonLaster:
This should save the mochitest logs as artifacts


bin/ci/upload-artifact.js
```diff
@@ -0,0 +1,75 @@
...
+    });
+
+    // uploader.on("progress", function() {
+    //   console.log(
+    //     "progress",
+    //     uploader.progressMd5Amount,
+    //     uploader.progressAmount,
+    //     uploader.progressTotal
+    //   );
+    // });
```
bomsy: do we need to take the comment out? or would u want to turn it on later?

jasonLaster: meh... not sure if it will work for us because it doesn't support forks...

.travis.yml
```diff
@@ -5,42 +5,21 @@ cache:
...
-before_script:
-  - ./bin/ci/build-firefox.sh
-  - sh -e /etc/init.d/xvfb start
-
 script:
-  - node --version
-  - du -sh firefox
-  - node ./bin/ci/run-tests.js
+  - echo "hi" > foo3.log
+  - node bin/ci/upload-artifact.js ./foo3.log
```
bomsy: is this a test?

PR #5200, opened 2018-01-25
Reset SourcesTree when necessary

MikeRatcliffe:
Pull request #5004 introduced an issue where the SourcesTree was not updated when navigating to a new URL. In reality it was setting the tree using `props.blah` instead of `nextProps.blah`.

This also fixes a couple of undefined errors I was randomly getting in the console.

### DIFF
--- a/src/components/PrimaryPanes/SourcesTree.js
+++ b/src/components/PrimaryPanes/SourcesTree.js
@@ -4,12 +4,13 @@
 
 // @flow
 
-// React
+// Dependencies
 import React, { Component } from "react";
 import classnames from "classnames";
-
-// Redux
+import { showMenu } from "devtools-contextmenu";
 import { connect } from "react-redux";
+
+// Selectors
 import {
   getShownSource,
   getSelectedSource,
@@ -19,12 +20,10 @@ import {
   getSources
 } from "../../selectors";
 
+// Actions
 import { setExpandedState } from "../../actions/source-tree";
 import { selectLocation } from "../../actions/sources";
-
-// Types
-import type { SourcesMap } from "../../reducers/types";
-import type { SourceRecord } from "../../reducers/sources";
+import { setProjectDirectoryRoot } from "../../actions/ui";
 
 // Components
 import ManagedTree from "../shared/ManagedTree";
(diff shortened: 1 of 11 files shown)


jasonLaster:
Oh wow - that's a nice catch. Would it be possible to add a new unit test for SourcesTree?

Here is a unit test for the Outline component which is similar
https://github.com/devtools-html/debugger.html/blob/master/src/components/tests/Outline.spec.js#L1

lukaszsobek:
Nice catch!

src/components/PrimaryPanes/SourcesTree.js
```diff
@@ -184,6 +184,8 @@ class SourcesTree extends Component<Props, State> {
     const { sources } = this.props;
     const blackBoxedPart =
       item.contents.get &&
+      item.contents.has("id") &&
+      sources.has(item.contents.get("id")) &&
       sources.get(item.contents.get("id")).get("isBlackBoxed")
```
lukaszsobek: This might be split into two parts to reduce the time it takes to parse.

src/components/PrimaryPanes/SourcesTree.js
```diff
@@ -212,7 +214,7 @@ class SourcesTree extends Component<Props, State> {
 
     if (!nodeHasChildren(item)) {
       const source = sources.get(item.contents.get("id"));
```
lukaszsobek: We could set this a bit higher and reuse source.

lukaszsobek: I think we could set this a bit further above and reuse here.

src/components/PrimaryPanes/SourcesTree.js
MikeRatcliffe: Done

src/components/PrimaryPanes/SourcesTree.js
MikeRatcliffe: Done

MikeRatcliffe:
Created test but somebody should probably review it.

jasonLaster:
The tests are a good start. I think we want to have a test that simulates changing the sources via setting new props:

this test has a good example of that https://github.com/devtools-html/debugger.html/blob/master/src/components/Editor/tests/Editor.spec.js#L68-L73

MikeRatcliffe:
Should be good to go now.

jasonLaster:
just rebased.

PR #5207, opened 2018-01-26
Revert "[breakpoints] syncing when the breakpoint is removed (#5176)"

jasonLaster:
This reverts commit e79b8f181bf0fd7769e462562834ecfb8d933360.

This PR landed because mochitest returns an exit code `0` if the last assert was positive


jasonLaster:
Alright, master should now be green, but we'll need to work on re-landing these 4 commits.

PR #5208, opened 2018-01-26
Re-add other tests

jasonLaster:
@juliandescottes recently added other tests that touched the debugger

I had to take them out because the way they were setup were timing out... Here's a new branch that attempts to land them so that they all run

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,3 +46,15 @@ script:
   - node ./bin/copy-assets.js --mc ./firefox
   - node ./bin/ci/check-file-sizes.js
   - ./node_modules/.bin/mochii --ci --mc ./firefox --headless devtools/client/debugger/new
+  - ./node_modules/.bin/mochii --ci --mc ./firefox --headless
+    browser_toolbox_debugger.js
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -46,3 +46,16 @@ script:
   - node ./bin/copy-assets.js --mc ./firefox
   - node ./bin/ci/check-file-sizes.js
   - ./node_modules/.bin/mochii --ci --mc ./firefox --headless devtools/client/debugger/new
+  - ./node_modules/.bin/mochii --ci --mc ./firefox --headless
+    browser_toolbox_debugger.js
```
juliandescottes: the real name of the test is browser_browser_toolbox_debugger.js (and it will time out until we have a m-c changeset with the fix from https://bugzilla.mozilla.org/show_bug.cgi?id=1432842)

juliandescottes:
> I had to take them out

I still see the tests in tree. Are we missing a commit?

juliandescottes:
See https://github.com/jasonLaster/mochii/issues/56

PR #5209, opened 2018-01-26
CI: add mochitests using the debugger and located in non-debugger tes…

juliandescottes:
Reopening this with a proper list of tests


juliandescottes:
Travis doesn't run here, not sure why, let's close this in favor of #5210

PR #5213, opened 2018-01-26
Bump devtools-components to 0.1.4

MikeRatcliffe:
Couldn't be simpler... bump devtools-components to 0.1.4

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,7 +71,6 @@
     "babel-types": "^6.26.0",
     "babylon": "^6.18.0",
     "codemirror": "^5.28.0",
-    "devtools-components": "^0.0.2",
     "devtools-launchpad": "^0.0.115",
     "devtools-linters": "^0.0.4",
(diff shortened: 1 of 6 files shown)


wldcordeiro:
Failing snapshots so I think you need to run `yarn test -u`

MikeRatcliffe:
> Failing snapshots so I think you need to run `yarn test -u`

Because I was working on a pull request I decided to mark it WIP so that nobody else would work on it.

Thanks for giving me the magical incantation though... that save me some time.

MikeRatcliffe:
I am testing all debugger functionality because I don't want to recreate all of the snapshots if the debugger is doing the wrong thing.

MikeRatcliffe:
As soon as we land this I will create a bundle and get on with upgrading all the things to React 16.

MikeRatcliffe:
> I think it is safe to remove the devtools-components dependency from package.json...

Done

PR #5215, opened 2018-01-26
First-pass implementation of mapping original scopes to server scopes

loganfsmyth:
Here ya go @jasonLaster 

### Summary of Changes

* Moved AST visitor for scope processing into debugger.html
* Improved it a bunch
* Used those scopes to build a renderable scope list representing the original file

With Babel generating sourcemaps but not actually doing transformations
screen shot 2018-01-26 at 1 00 19 pm

With babel-preset-env enabled
screen shot 2018-01-26 at 1 03 45 pm


src/test/mochitest/browser.ini
```diff
@@ -108,6 +108,7 @@ skip-if = os == "linux" # bug 1351952
 skip-if = os == "win"
 [browser_dbg-navigation.js]
 [browser_dbg-minified.js]
+skip-if = true
```
jasonLaster: we should prioritize getting these tests passing soon :)

Issue #2545, opened 2017-04-05
[Sources] use file type icons


jasonLaster:
We can use the icon font that atom uses for its source tree

- [x] React Components - #2690 [badge](https://vorillaz.github.io/devicons/#/singleicon/react) [1]
- [x] JavaScript - [badge](https://vorillaz.github.io/devicons/#/singleicon/javascript_badge)
- [x] Webpack - source map domains. We'll need to get an SVG
- [ ] Style the tree now that we have some icons
- [x] style the webpack icon #2859

[1] We can be conservative and only highlight component files.
  1. use babel to find a file that uses `createClass` or extends `Component`.
  2. check the file path to see if it is in a component dir...
  
  ---

- https://vorillaz.github.io/devicons/#/dafont
- we can likely use dafont's SVGs instead of using the font library and getting webpack to work.

irfanhudda:
Do we want to display different icons for `.jsx` files or coffeescript files in source tree of debugger instead of generic file icons?

jasonLaster:
@wldcordeiro is on it

Issue #4141, opened 2017-09-23
[Outline] Include icons from DevIcons


jasonLaster:
It would be sweet to use a consistent set of icons as Atom. This is a two step process:

1. setup a font-loader in webpack for base64 encoding fonts
2. pull in octicons

Here are some links

* https://github.com/atom/atom/blob/master/static/variables/octicon-utf-codes.less
* https://github.com/atom/atom/blob/master/static/octicons-2.1.2.woff
* http://fizzed.com/oss/font-mfizz
* https://vorillaz.github.io/devicons/#/dafont
* https://atom.io/packages/file-icons
* https://github.com/nicksrandall/base64-font-loader

jasonLaster:
and here are the SVGs: https://github.com/vorillaz/devicons/tree/master/!SVG

jasonLaster:
here are octicon icons: https://github.com/primer/octicons

jasonLaster:
![Screen Shot 2017-10-10 at 10.12.12 AM.png](https://shipusercontent.com/8a4017ab19a653453c8dac52ff2fe486/Screen%20Shot%202017-10-10%20at%2010.12.12%20AM.png)

and here's where the octicons font lives: `/Applications/Atom.app/Contents/Resources/app/static`

i still think if we figureout which svg is used, we can use that...

jasonLaster:
The goal here is to re-use atom's function and class icons:

<img src='https://shipusercontent.com/5538bde5840778578eafe03d5f71ad92/Screen%20Shot%202017-10-11%20at%201.51.34%20PM.png' title='Screen Shot 2017-10-11 at 1.51.34 PM.png' width=910>

sarahlim:
Do you know what Atom plugins are in the screenshot above? I was hoping to inspect the webview, since I haven't had any luck finding those exact icons from [Octicons](https://octicons.github.com/) or [devicons](https://vorillaz.github.io/devicons/#/dafont). (Unless we don't actually need to get the same ones? But I'm having trouble finding plain lambda icons, too 😛)

lukaszsobek:
Can this be closed, as the lambda issue is resolved or are we going for a full icon rework?

PR #5216, opened 2018-01-26
feat(SourceTree) - JS and React icons for source files

calebstdenis:
Associated Issue: #2545

### Summary of Changes
Source tree icons for JavaScript, CoffeeScript, React, and TypeScript.

### Test Plan
Tested with TypeScript, CoffeeScript, and React debugger examples, as seen in the screenshots below. In the case of the React example I had to go in and change the component file from ".js" to ".jsx" so it would show the React icon.

### Screenshots
image
------
image
------
image

^ It annoys me that the TS icon doesn't match the JS one, but it's all that was readily available. I'm thinking of making a separate issue for hunting down / making a matching SVG?


calebstdenis:
Thinking of encapsulating `SourceTree` node logic into a `SourceTreeNode` component so I can snapshot test these guys. Thoughts?

lukaszsobek:
For the icons you might find this interesting: https://github.com/devtools-html/debugger.html/issues/4141

jasonLaster:
Thanks @lukaszsobek. I think these icons all look good :)

assets/images/coffeescript.svg
```diff
@@ -0,0 +1,6 @@
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
```
wldcordeiro: Are we actually cool with putting the Mozilla license header on images of these library logos? I swear some may have trademarking or something.

jasonLaster:
Looks like we need to handle the dark theme unselected case:

assets/images/coffeescript.svg
calebstdenis: I'm no lawyer but afaik it's OK to use trademarks as long as you're not misleading people into thinking that you're associated with the company who owns it.

calebstdenis:
@jasonLaster re dark theme: On it.

calebstdenis:
@jasonLaster Fixed by setting the icons to the same color as the text

image
----
image

jasonLaster:
looks good to me!

Issue #5185, opened 2018-01-24
[Tests] Testing for promise does not fail


lukaszsobek:
We have several situations in our tests where we want the return value of a function and use `.toBe(...)` to compare it with the desired outcome. There is no problem with **synchronous** functions to be handled this way. But there might be a problem with **asynchronous** functions, as they require an additional step, where we do not compare the result of the function itself but chain the `expect.toBe` into the `.then` part of the test. e.g.

```
test('the data is peanut butter', () => {
  expect.assertions(1);
  return fetchData().then(data => {
    expect(data).toBe('peanut butter');
  });
});

```
Or wait for the state to update with `waitForState()`. Otherwise we will pass tests even though they should be failing:
![screenshot-20180124074207-544x781](https://user-images.githubusercontent.com/23530054/35318376-5b349862-00db-11e8-8114-549c1108e9b0.png)

lukaszsobek:
Merged. Closing.

PR #5217, opened 2018-01-26
Calls the correct pause info

lukaszsobek:
Associated Issue: #5185

- A non-existent pauseInfo was called which resulted in an empty selector and the test passing with an error in the console despite the generatedSourceRecord in fetchScopes.js being undefined.  It seems the test now fails like it should.


jasonLaster:
Nice work!

PR #5222, opened 2018-01-28
Truncate css fix

CharlesStocky:
Summary of Change

This fix truncates location names if the location string expands beyond the default width of an element. I eventually found a utility function (endTruncateStr) to be unnecessary, and may hamper future development. This fix replaces that utility function, wherever present, with CSS (text-overflow: ellipsis, except for the frames within callstack-pane), which fixes both issues.

I haven't found a solution for truncating (with ellipsis) of the frames within callstack-pane. This fix does hide/reveal more of the location strings within that element upon pane resizing.
Test Plan

Dynamic truncation checklist:
[x] Upon hitting a breakpoint, locations that expand beyond an element's width should be truncated.
[x] Pane resizing should increase/decrease the visibility of the location string when increasing/decreasing pane width (dynamic truncation).

Screenshots

Firefox: 

screen shot 2018-01-25 at 7 33 07 pm

screen shot 2018-01-25 at 7 49 15 pm

Chrome: 

screen shot 2018-01-27 at 6 58 24 pm

screen shot 2018-01-27 at 6 58 33 pm

### DIFF
--- a/src/components/Editor/Tabs.js
+++ b/src/components/Editor/Tabs.js
@@ -385,7 +385,6 @@ class SourceTabs extends PureComponent<Props, State> {
         onContextMenu={e => this.onTabContextMenu(e, source)}
         title={tabName(source)}
       >
-        {sourceAnnotation}
         <div className="filename">{tabName(source)}</div>
         <CloseButton
           handleClick={onClickClose}
@@ -403,7 +402,6 @@ class SourceTabs extends PureComponent<Props, State> {
       source.get("id") == selectedSource.get("id") &&
       (!this.isProjectSearchEnabled() && !this.isSourceSearchEnabled());
     const isPrettyCode = isPretty(source);
-    const sourceAnnotation = this.getSourceAnnotation(source);
 
     function onClickClose(ev) {
       ev.stopPropagation();


lukaszsobek:
Good work! I think that if it shortens the name in the panes it would be nice if it showed the full version in a tooltip when hovering over the item.

src/utils/utils.js
```diff
@@ -34,21 +34,6 @@ function promisify(context: any, method: any, ...args: any): Promise<mixed> {
   });
 }
 
-/**
- * @memberof utils/utils
- * @static
- */
-function endTruncateStr(str: any, size: number) {
```
lukaszsobek: This will probably break tests as they are still expecting the string to be truncated manually. You'd need to remove tests that check if the text strings are truncated in order for this to pass.

src/components/Editor/Tabs.css
```diff
@@ -45,12 +45,12 @@
   align-items: center;
   position: relative;
   transition: all 0.15s ease;
-  min-width: 40px;
   overflow: hidden;
   padding: 5px;
   margin-inline-start: 3px;
   margin-top: 3px;
   cursor: default;
+  max-width: 30%;
```
lukaszsobek: Oh, what is this for?

src/components/Editor/Tabs.js
```diff
@@ -422,7 +421,7 @@ class SourceTabs extends PureComponent<Props, State> {
         onContextMenu={e => this.onTabContextMenu(e, source.get("id"))}
         title={getFileURL(source.toJS())}
       >
-        {sourceAnnotation}
+        {}
```
lukaszsobek: I like curly braces :)

src/components/SecondaryPanes/Breakpoints.js
```diff
@@ -70,8 +69,7 @@ function getBreakpointFilename(source) {
 function renderSourceLocation(source, line, column) {
   const filename = getBreakpointFilename(source);
   const isWasm = source && source.get("isWasm");
-  const columnVal =
-    features.columnBreakpoints && column ? `:${column}` : "";
+  const columnVal = features.columnBreakpoints && column ? `:${column}` : "";
```
lukaszsobek: Multi-line ternary usually go like this
```
const columnVal = features.columnBreakpoints && column
? `:${column}`
: "";
```

lukaszsobek:
It would be nice if you could check how this behaves in RTL settings (in the launchpad click `settings` and choose `rtl` instead of `ltr`)

CharlesStocky:
@lukaszsobek cool, I'll look at removing the tests asap. I'll keep toying around with truncating the call stack frames with ellipsis as well, it hasn't been as simple as applying `text-overflow: ellipsis` to it.

CharlesStocky:
Also, I'm not sure why I set `max-width` for tabs. I'll push with `width` when I remove the tests.

CharlesStocky:
Also also, the tooltip that I added to callstack-pane shows just the line number. My thinking is that the location is viewable by tooltip in multiple locations (tab, footer elements), however the line number of the breakpoints is not. I thought to add a column in the callstack-pane that displayed the line number, and perhaps a toggle button (so all or no line numbers would be viewable). If you think a tooltip would be preferable, I could figure out a way of seperating the location/line number within the tooltip.

jasonLaster:
Hmm, not sure about the tooltip. One advantage of adding the filename is that you can hover on a frame you haven't clicked.

Looks great though!

CharlesStocky:
Here's the tooltip: 

screen shot 2018-01-29 at 2 44 01 am

Let me know if the line break is necessary. 

Here are the changes in the rtl view: 

screen shot 2018-01-29 at 2 50 35 am

jasonLaster:
Wow those tooltips are crazy. I'm not sure how I feel. They usually less-invasive. I think I would prefer leaving out the line number because it does make the tooltip quite big.

CharlesStocky:
Here's the tool tip without the line break/"Line number": 

screen shot 2018-01-29 at 6 15 59 pm

jasonLaster:
lets drop `:line number`

CharlesStocky:
I'm still learning Git, so bear with me if possible :). I went ahead and reforked debugger.html, and I'm redoing the changes in that fork on the same branch (css-truncate-fix). It shouldn't take very long at all to re-implement the fixes, although some of the changes I made don't look so good in the most recent version of debugger.html. I'll be around in the Slack channel in a bit.

CharlesStocky:
I'm also looking into decreasing the width of the tabs only if necessary, instead of decreasing them when the editor pane narrows by default.

src/components/Editor/Tabs.css
darkwing: Without the 30% max-width, we have the tab go under the toggle button:

nomaxwidth

src/components/SecondaryPanes/Frames/Frame.js
```diff
@@ -42,7 +42,10 @@ function FrameLocation({ frame }: FrameLocationProps) {
 
   const filename = getFilename(frame.source);
   return (
-    <div className="location">{`${filename}: ${frame.location.line}`}</div>
+    <div className="location" title={"Line number:" + frame.location.line}>
+      <div className="fileName">{filename}: </div>
+      <div className="lineNumber">{frame.location.line}</div>
```
jasonLaster: do we want separate `<div>`s or spans?

CharlesStocky: Spans :). Still working the width issue with the tabs. Would you be against adding state to the Tab component? I haven't found a way of implementing dynamic resizing that isn't convoluted and that goes with best-practices.

jasonLaster: sure, give it a try!

darkwing:
Closing in favor of: https://github.com/devtools-html/debugger.html/pull/5325

PR #5224, opened 2018-01-28
Revert "Revert "[Sources] - fix paddings in sources tree and outline …

jasonLaster:
This brings back the padding fix, which had to be reverted because it broke dbg-sources

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,8 +45,8 @@ script:
   - du -sh firefox
   - node ./bin/copy-assets.js --mc ./firefox
   - node ./bin/ci/check-file-sizes.js
-  - ./node_modules/.bin/mochii --ci --mc ./firefox --headless devtools/client/debugger/new
-  - ./node_modules/.bin/mochii --ci --mc ./firefox --headless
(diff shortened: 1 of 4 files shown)


jasonLaster:
Alright - we now know that `onClick` is never received :)

jasonLaster:
Weird i don't see the effect of the grid...

jasonLaster:
CC @martinfojtik lets see if rolling back the grid stuff fixes it... just guessing now. Would love your thoughts here. Also, feel free to open your own PR so you can test what travis is doing too...

this might be a first for me.

jasonLaster:
hmm - this helps the tests go green... can we try and avoid the `display: grid` stuff?

martinfojtik:
We should be able using grid without any problems. If I disable it, there is a problem as you can see on video Gif.
devtools-sources

martinfojtik:
@jasonLaster I just created the same PR with display: grid and all tests passed. Can you try to Rebase your branch? Maybe something from this PR https://github.com/devtools-html/debugger.html/pull/5230 was causing the problem.

jasonLaster:
@martinfojtik i'm not sure - travis wasn't included in your run. try adding an extra comment somewhere to get a different sha...

Issue #5220, opened 2018-01-27
[Expressions] Cursor moves to the end when writing watch expression


Loirooriol:
Like #4693, but with watch expressions.

1. Click *Add watch expression*
2. Write `123`, press enter
3. Double click the watch expression
4. Press home key
5. Write `abc`

Result: the field shows `a123bc`
Expected: `abc123`

lukaszsobek:
lol, this one is good. Here is a visual on the bug:
![abc123](https://user-images.githubusercontent.com/23530054/35481127-ad850c4c-041d-11e8-802d-75cfc5bfb6e2.gif)

Loirooriol:
I still don't understand why somebody thought using `setSelectionRange` to move the cursor to the end would be useful. I would just get rid of that.

https://github.com/devtools-html/debugger.html/blob/8fe140cce02dd4a7181b436b7ac46875dbc1251d/src/components/SecondaryPanes/Expressions.js#L87

It's also in SearchInput.js, maybe it's what causes #5214.

https://github.com/devtools-html/debugger.html/blob/8fe140cce02dd4a7181b436b7ac46875dbc1251d/src/components/shared/SearchInput.js#L60

jasonLaster:
yeah - i forgot why ...

CC @wldcordeiro

lukaszsobek:
Merged. Closing.

PR #5226, opened 2018-01-28
Highlight all text on watch expression edit

AnshulMalik:
Associated Issue: #5220 

watch

### DIFF
--- a/src/components/SecondaryPanes/Expressions.js
+++ b/src/components/SecondaryPanes/Expressions.js
@@ -83,9 +83,8 @@ class Expressions extends PureComponent<Props, State> {
 
   componentDidUpdate(prevProps, prevState) {
     if (this._input && !prevState.editing) {
-      // Put the cursor to the end of input text, when start editing
       const input = this._input;
-      input.setSelectionRange(input.value.length + 1, input.value.length + 1);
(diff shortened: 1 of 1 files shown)


jasonLaster:
looks great. It would be nice to have a comment on what the code is trying to do :)

AnshulMalik:
Hehe, I thought we hate comments :P

jasonLaster:
this is the exception to the rule lol

jasonLaster:
do we need to touch SearchInput?

jasonLaster:
CC @Loirooriol

AnshulMalik:
It's already a one-time thing in `SearchInput` because it's in `componentDidMount`. So we are good without this change.

AnshulMalik:
We should probably make this a helper function, `moveCursorToEnd`

Loirooriol:
Moving the cursor to the end when edition starts also seems a bit arbitrary. Selecting all the text may be more useful, this way you can just start typing and replace the old expression with the new one.

jasonLaster:
I think given that we're double clicking to edit the user would expect the text to be highlighted.

AnshulMalik:
watch

jasonLaster:
:fire:

PR #5229, opened 2018-01-28
[QuickOpen] style matched fuzzy characters

atwalg2:
Associated Issue: #4488 

### Summary of Changes

* title highlighting
* subtitle highlighting

### To Do

* Create Tests

### Screens

image

image

### DIFF
--- a/src/components/QuickOpenModal.css
+++ b/src/components/QuickOpenModal.css
@@ -1,6 +1,6 @@
-.result-item .title .matching-character {
+.result-item .title .fuzzy-match {
   font-weight: bold;
 }
-.result-item .subtitle .matching-character {
+.result-item .subtitle .fuzzy-match {
   color: black;
 }
--- a/src/components/QuickOpenModal.css
+++ b/src/components/QuickOpenModal.css
@@ -1,8 +1,7 @@
-.result-item .title .matching-character {
+
+.result-item .title .matching-character{
   font-weight: bold;
 }
-
-.result-item .subtitle .matching-subtitle {
-  color: black;
-  font-weight: bold;
+.result-item .subtitle .matching-character {
+  color:black;
 }
--- a/src/components/QuickOpenModal.css
+++ b/src/components/QuickOpenModal.css
@@ -1,3 +1,3 @@
-.result-item .title .matching-letter {
+.result-item .title .matching-character {
   font-weight: bold;
 }
--- a/src/components/QuickOpenModal.css
+++ b/src/components/QuickOpenModal.css
@@ -1,0 +1,3 @@
+.result-item .title .matching-letter {
+  font-weight: bold;
+}
--- a/src/components/QuickOpenModal.css
+++ b/src/components/QuickOpenModal.css
@@ -1,7 +1,6 @@
-
-.result-item .title .matching-character{
+.result-item .title .matching-character {
   font-weight: bold;
 }
 .result-item .subtitle .matching-character {
-  color:black;
+  color: black;
 }
--- a/src/components/QuickOpenModal.css
+++ b/src/components/QuickOpenModal.css
@@ -1,3 +1,8 @@
 .result-item .title .matching-character {
   font-weight: bold;
 }
+
+.result-item .subtitle .matching-subtitle {
+  color:black;
+  font-weight: bold;
+}
--- a/src/components/QuickOpenModal.css
+++ b/src/components/QuickOpenModal.css
@@ -1,6 +1,12 @@
 .result-item .title .fuzzy-match {
   font-weight: bold;
 }
+
 .result-item .subtitle .fuzzy-match {
   color: black;
+}
+
+.theme-dark .result-item .title .fuzzy-match,
+.theme-dark .result-item .subtitle .fuzzy-match {
+  color: white;
 }
--- a/src/components/QuickOpenModal.css
+++ b/src/components/QuickOpenModal.css
@@ -3,7 +3,7 @@
 }
 
 .result-item .subtitle .fuzzy-match {
-  color: black;
+  color: var(--grey-90);
 }
 
 .theme-dark .result-item .title .fuzzy-match,
--- a/src/components/QuickOpenModal.css
+++ b/src/components/QuickOpenModal.css
@@ -3,6 +3,6 @@
 }
 
 .result-item .subtitle .matching-subtitle {
-  color:black;
+  color: black;
   font-weight: bold;
 }
--- a/src/components/QuickOpenModal.js
+++ b/src/components/QuickOpenModal.js
@@ -302,6 +302,17 @@ export class QuickOpenModal extends Component<Props, State> {
   isGotoSourceQuery = () => this.props.searchType === "gotoSource";
   isShortcutQuery = () => this.props.searchType === "shortcuts";
 
+  renderHighlight = (part: string[] | string, i: number) => {
+    if (Array.isArray(part)) {
+      return (
+        <span key={`${part.join("")}-${i}`} className="matching-character">
+          {part.join("")}
+        </span>
+      );
+    }
+    return part;
+  };
+
   highlightMatching = (query: string, results: QuickOpenResult[]) => {
     if (query === "") {
       return results;
@@ -310,44 +321,19 @@ export class QuickOpenModal extends Component<Props, State> {
     return results.map(result => {
       const resultParts = result.title.toLowerCase().split("");
       const title = groupSharedChars(resultParts, queryLetters);
-      let subtitle = [""];
-      if (
-        result.subtitle !== undefined &&
-        Object.keys(MODIFIERS).includes(queryLetters[0])
-      ) {
-        const resultSubParts = result.subtitle.split("");
-        subtitle = groupSharedChars(resultSubParts, queryLetters);
-        console.log(subtitle);
-      }
+      const subtitle =
+        result.subtitle != null
+          ? groupSharedChars(
+              result.subtitle.toLowerCase().split(""),
+              queryLetters
+            )
+          : null;
       return {
         ...result,
-        title: title.map((part, i) => {
-          if (Array.isArray(part)) {
-            return (
-              <span
-                key={`${part.join("")}-${i}`}
-                className="matching-character"
-              >
-                {part.join("")}
-              </span>
-            );
-          }
-          return part;
-        }),
-        subtitle: subtitle.map((partX, iX) => {
-          if (Array.isArray(partX)) {
-            return (
-              <span
-                key={`${partX.join("")}-${iX + 50}`}
-                className="matching-subtitle"
-                // style={{ color: "blue" }}
-              >
-                {partX.join("")}
-              </span>
-            );
-          }
-          return partX;
-        })
+        title: title.map(this.renderHighlight),
+        ...(subtitle != null
+          ? { subtitle: subtitle.map(this.renderHighlight) }
+          : null)
       };
     });
   };
--- a/src/components/QuickOpenModal.js
+++ b/src/components/QuickOpenModal.js
@@ -23,7 +23,7 @@ import {
   formatSources,
   parseLineColumn,
   formatShortcutResults,
-  groupSharedChars
+  groupFuzzyMatches
 } from "../utils/quick-open";
 import Modal from "./shared/Modal";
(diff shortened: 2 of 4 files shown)


atwalg2:
I opened this to get some advice, on how I can improve my logic, I think it could probably be better.

src/components/QuickOpenModal.js
```diff
@@ -68,6 +68,10 @@ function filter(values, query) {
   });
 }
 
+const styles = {
```
wldcordeiro: Style in the CSS file.

src/components/QuickOpenModal.js
```diff
@@ -286,6 +290,40 @@ export class QuickOpenModal extends Component<Props, State> {
     return results && results.length ? results.length : 0;
   };
 
+  fuzzyhighlight = (query, results) => {
```
wldcordeiro: Something like this maybe... I haven't had a chance to test so consider this a sketch. 😄 

```jsx
return results.map(result => {
  if (MODIFIERS.includes(query[0]) {
    return result;
  }

  const matching = new Set(...query.split(), ...result.title.split());
  return {
    ...result,
    title: result.title
      .split()
      .map(letter => matching.includes(letter) ? <span className="matched">letter</span> : letter)
  };
});
```

MODIFIERS can be found in utils/quick-open (but you need to export it)

One idea is instead of a span around each letter you could map the ranges and put spans just around the characters that you match.

```
// query = food
<span>foo</span>bar
```

src/components/QuickOpenModal.js
```diff
@@ -302,6 +340,10 @@ export class QuickOpenModal extends Component<Props, State> {
       return null;
     }
 
+    if (results) {
```
wldcordeiro: Based on my note above you should do this also even if you were to do it differently I don't think you should be mutating `results`:

```js
let highlightedResults;
if (results) {
  highlightedResults = this.fuzzyhighlight(query, results);
}
```

src/components/QuickOpenModal.js
atwalg2: for some reason when I try it this way I can't seem to get it to continue highlighting. I imported the css file, and tried to attach className='highlight'.

I may just be messing up the syntax?

wldcordeiro:
Hey @atwalg2 So based on the note above I tried to produce a working version. Here is a git patch!

```patch
index a822e21b..9fefd659 100644
@@ -22,7 +22,8 @@ import {
   formatSymbols,
   formatSources,
   parseLineColumn,
-  formatShortcutResults
+  formatShortcutResults,
+  MODIFIERS
 } from "../utils/quick-open";
 import Modal from "./shared/Modal";
 import SearchInput from "./shared/SearchInput";
@@ -37,6 +38,8 @@ import type { Location } from "debugger-html";
 import type { SourceRecord } from "../reducers/sources";
 import type { QuickOpenType } from "../reducers/quick-open";
 
+import "./QuickOpenModal.css";
+
 type Props = {
   enabled: boolean,
   sources: Array<Object>,
@@ -296,6 +299,29 @@ export class QuickOpenModal extends Component<Props, State> {
   isGotoSourceQuery = () => this.props.searchType === "gotoSource";
   isShortcutQuery = () => this.props.searchType === "shortcuts";
 
+  highlightMatching = (query, results) => {
+    if (!query || !Object.keys(MODIFIERS).includes(query[0])) {
+      return results;
+    }
+    const queryLetters = query.split();
+    return results.map(result => {
+      const shared = new Set(...queryLetters, ...result.title.split());
+      return {
+        ...result,
+        title: result.title.split("").map(
+          (letter, i) =>
+            shared.has(letter) ? (
+              <span key={`${letter}-${i}`} className="matching-letter">
+                {letter}
+              </span>
+            ) : (
+              letter
+            )
+        )
+      };
+    });
+  };
+
   render() {
     const { enabled, query, searchType } = this.props;
     const { selectedIndex, results } = this.state;
@@ -330,7 +356,7 @@ export class QuickOpenModal extends Component<Props, State> {
         {newResults && (
           <ResultList
             key="results"
-            items={newResults}
+            items={this.highlightMatching(query, newResults)}
             selected={selectedIndex}
             selectItem={this.selectResultItem}
             ref="resultList"
diff --git a/src/utils/quick-open.js b/src/utils/quick-open.js
index d3e1a0cc..df757b2c 100644
--- a/src/utils/quick-open.js
+++ b/src/utils/quick-open.js
@@ -14,7 +14,7 @@ import type {
   SymbolDeclarations
 } from "../workers/parser/types";
 
-const MODIFIERS = {
+export const MODIFIERS = {
   "@": "functions",
   "#": "variables",
   ":": "goto",
```

For some reason it wouldn't add the new CSS file to the patch but I created a `QuickOpenModal.css` next to the JS file with this in it

```css
.result-item .title .matching-letter {
  color: var(--theme-selection-background);
}

```

To use the patch you can copy paste it into a file `something.patch` and run `git apply something.patch`

wldcordeiro:
We might want to try different highlight colors also.

screen shot 2018-01-28 at 5 55 27 pm

wldcordeiro:
@atwalg2 https://gist.github.com/wldcordeiro/8824d28e37a534f745be398285d73aed Here is a patch.

atwalg2:
@wldcordeiro So I tried out your patch. However their are some issues.

src/components/QuickOpenModal.js
```diff
@@ -50,6 +50,12 @@ type Props = {
   closeQuickOpen: () => void
 };
 
+const styles = {
```
lukaszsobek: We should really try to take this into its own css. I know it didn't work out the first time but having css styling in our js files is going to make the code really ugly.

src/components/QuickOpenModal.js
```diff
@@ -286,6 +292,51 @@ export class QuickOpenModal extends Component<Props, State> {
...
+
+      for (resultElem; resultElem < results.length; resultElem++) {
+        let queryElem = 0;
+        if (query[0] === "@" || query[0] === "#" || query[0] === ":") {
+          queryElem++;
+        }
+
+        let titleElem = 0;
+        const title = results[resultElem].title;
+        const styledResults: Array<any> = [];
```
lukaszsobek: Any chance to avoid any here? :)

atwalg2:
@wldcordeiro I tried out your patch, however their are some potential issues with it, based on my understanding of how it is meant to be implemented.

1. The line if (!query || !Object.keys(MODIFIERS).includes(query[0])) prevents the highlighting when no modifier is in front. Is this something we really want?
2. From what @jasonLaster told me before I started, we should be matching the characters only on their first occurrence in the string

src/components/QuickOpenModal.js
```diff
@@ -286,6 +292,51 @@ export class QuickOpenModal extends Component<Props, State> {
...
+
+        results[resultElem].title = "";
+
+        for (queryElem; queryElem < query.length; queryElem++) {
+          for (titleElem; titleElem < title.length; titleElem++) {
+            const q = query[queryElem];
+            const t = title[titleElem];
+            if (q === t || queryUpper[queryElem] === t) {
+              styledResults.push(
+                <span key={q + titleElem} style={styles}>
```
lukaszsobek: Maybe something like `className="match"` instead of `style={styles}` so css can take care of it. If `.match` is not specific enough to trigger the css we can always amp up the specificity in the css file until it sticks, like ".result-list .result-item .title .match".

src/utils/quick-open.js
```diff
@@ -54,7 +54,7 @@ export function parseLineColumn(query: string) {
 export type QuickOpenResult = {|
   id: string,
   value: string,
-  title: string,
+  title: string | any,
```
lukaszsobek: Do we expect the title to be anything else than a string?

src/components/QuickOpenModal.js
```diff
@@ -302,6 +353,11 @@ export class QuickOpenModal extends Component<Props, State> {
       return null;
     }
 
+    let highlightedResults = results;
```
lukaszsobek: I feel like we could do this differently considering the fact that we only need `highlightedResults` when there actually is something to highlight and we do not use `results` for the remainder of our story.

src/utils/quick-open.js
atwalg2: Yea actually I was puting <span></span> into it, and if I convert span to a string it renders as Object Object.

src/components/QuickOpenModal.js
atwalg2: Currently no, because I have the highlighted element that has to be some JSX/HTML that render as Object object if I put in something like a string for example.

src/components/QuickOpenModal.js
atwalg2: This was very helpful, thanks. I got it working with your advice.

src/components/QuickOpenModal.js
```diff
@@ -319,7 +315,10 @@ export class QuickOpenModal extends Component<Props, State> {
             const t = title[titleElem];
             if (q === t || queryUpper[queryElem] === t) {
               styledResults.push(
-                <span key={q + titleElem} style={styles}>
+                <span
+                  key={q + titleElem}
+                  className="result-item title matching-letter"
```
lukaszsobek: You should just need `matching-letter` here. The rest is already there in terms of classes. The `matching-letter` is already inside a container with the class `title` which in turn is inside a container classed `result-item`.

src/components/QuickOpenModal.js
```diff
@@ -353,9 +351,11 @@ export class QuickOpenModal extends Component<Props, State> {
       return null;
     }
 
-    let highlightedResults = results;
+    let highlightedResults;
     if (results) {
       highlightedResults = this.fuzzyhighlight(query, results);
```
lukaszsobek: How about `const highlightedResults = results ? this.fuzzyhighlight(query, results) : []; `

wldcordeiro:
> The line if (!query || !Object.keys(MODIFIERS).includes(query[0])) prevents the highlighting when no modifier is in front. Is this something we really want?

Not sure if we do. I was trying to use your work as a basis and saw a check for the first character. We could remove that condition though.

wldcordeiro:
Utilizing my patch seems to pass Flow with the latest [revision](https://gist.github.com/wldcordeiro/8824d28e37a534f745be398285d73aed/revisions) and removes that condition.

wldcordeiro:
> From what @jasonLaster told me before I started, we should be matching the characters only on their first occurrence in the string

I dunno about that, looks like it's all occurrences in Atom. Dunno about VS Code or others though.

screen shot 2018-01-29 at 9 57 23 am

atwalg2:
It looks like its both in atom lol. 😕 
image
image

VSCode does first instance... except in the second example where it matches package.json it skips the first instance.
image
image

atwalg2:
I have this one flow issue, any advice on how to fix it?

image

jasonLaster:
@atwalg2 looks like it's not sure result is an object... try flipping the IF to check for the property

atwalg2:
turns out I just needed to change my || to && ... oops hahaha

jasonLaster:
This data structure might be cleaner: `[{type:"match", value: 'foo'}, {type:"miss", value: 'bar'}]` than `[['foo'], 'b', 'a', 'r']`.

src/utils/quick-open.js
```diff
@@ -126,27 +126,43 @@ export function formatSources(sources: SourcesMap): Array<QuickOpenResult> {
...
+  const queryChars = query.toLowerCase().split("");
+  const shared = parts.filter(char => queryChars.includes(char));
   const title = [];
   let matched;
+  let missed;
   parts.forEach((char, i) => {
     if (shared.includes(char)) {
       if (!matched) {
-        matched = [char];
+        matched = { type: "match", value: input[i] };
```
wldcordeiro: It's important that we use `input[i]` for the value portion as the others are all lowercased.

src/components/QuickOpenModal.css
```diff
@@ -0,0 +1,12 @@
...
+  font-weight: bold;
+}
+
+.result-item .subtitle .fuzzy-match {
+  color: black;
+}
+
+.theme-dark .result-item .title .fuzzy-match,
+.theme-dark .result-item .subtitle .fuzzy-match {
+  color: white;
```
jasonLaster: we should use css variables

PR #5160, opened 2018-01-20
Improve Breakpoint Syncing

jasonLaster:
Associated Issue: #5140

### Summary

1. Adds an integration test w/ 3 versions of the same original file
2. Adds a check for clearing null mapped generated lines, this check removes a breakpoint in the mochitest when the line the breakpoint is on no longer exists!!!

PR #5230, opened 2018-01-29
Revert "Revert "Improve Breakpoint Syncing (#5160)""

jasonLaster:
Re-add improve breakpoint syncing


jasonLaster:
re-adding as it is green... at a minimum we'll try and land this in release 12

PR #5236, opened 2018-01-29
Fix jest directory excludes

jasonLaster:
when firefox was installed, jest was searching there as well


AnshulMalik:
This was really needed 👍

PR #5238, opened 2018-01-29
Fix firefox.mcPath documentation.

ochameau:
I had to do a local.json file like this in order to make `yarn copy-assets` to work.

### DIFF
--- a/docs/running-as-firefox-panel.md
+++ b/docs/running-as-firefox-panel.md
@@ -48,7 +48,9 @@ However if you have firefox installed elsewhere you can update this in `configs/
 called
 
 ```json
-mcPath: './firefox' // expecting firefox to be under `debugger.html/firefox`
+{
+  "firefox.mcPath": "./firefox" // expecting firefox to be under `debugger.html/firefox`
+}
 ```
 
 You can change this to what works for you!


jasonLaster:
strange, that seems like the default. When i'm copying to another dir, i use the node script `node bin/copy-assets.js --mc ../gecko --assets` maybe we should just document that. local.json can be a pain.

ochameau:
I don't have much opinion here as I just discover all this ;)
The only important thing here is to have a correct and complete doc to build the m-c bundle manually.

ochameau:
Ok, so you have to pass --assets to have the same behavior than what is shipped.
And also --symlink false to prevent having `devtools/client/debugger/new/test/mochitest` to be deleted by the symlink function from launchpad.

May be it is easier to document setting that in the json file rather than saying you have to pass this to yarn copy-assets every time?

jasonLaster:
yeah - i'll review this tomorrow and try and simplify the steps. it's gotten pretty stale

codehag:
ping @jasonLaster did you have a chance to take a look? should we update the docs?

jasonLaster:
yep - just busy :)

Issue #5254, opened 2018-01-31
[Scopes] Render generated scopes when viewing generated code


loganfsmyth:
The Frames pane toggles source locations based on whether the user is viewing the original or generated file. With https://github.com/devtools-html/debugger.html/pull/5215 enabled, the Scopes pane renders the original scopes, but toggling back to the generated code continues to show the original scopes, which can make debugging difficult.

PR #5276, opened 2018-01-31
Render mapped scope list only when viewing original file

loganfsmyth:
Associated Issue: #5254

### Summary of Changes

* Add a general "Loading..." state for the scope list pane
* Change the scope state to be asynchronous with status
* Split the scope list state into two separate sets for generated and mapped scopes, and render based on which file you are viewing.

### Test Plan

- [ ] Stop debugger in original sourcemap file
- [ ] Sources pane will show scopes based on original file
- [ ] Click to open the generated file and note that the scopes reflect the generated code

scope-toggling

### DIFF
--- a/src/actions/pause/mapScopes.js
+++ b/src/actions/pause/mapScopes.js
@@ -27,6 +27,8 @@ import type {
 
 import type { ThunkArgs } from "../types";
 
+export type OriginalScope = RenderableScope;
+
 export function mapScopes(scopes: Promise<Scope>, frame: Frame) {
   return async function({ dispatch, getState, client, sourceMaps }: ThunkArgs) {
     const generatedSourceRecord = getSource(
@@ -46,23 +48,22 @@ export function mapScopes(scopes: Promise<Scope>, frame: Frame) {
       type: "MAP_SCOPES",
       frame,
       [PROMISE]: (async function() {
-        let mappedScopes;
-        if (shouldMapScopes) {
-          await dispatch(loadSourceText(sourceRecord));
-
-          try {
-            mappedScopes = await buildMappedScopes(
-              sourceRecord.toJS(),
-              frame,
-              await scopes,
-              sourceMaps
-            );
-          } catch (e) {
-            mappedScopes = null;
-          }
+        if (!shouldMapScopes) {
+          return null;
         }
 
-        return mappedScopes || scopes;
+        await dispatch(loadSourceText(sourceRecord));
+
+        try {
+          return await buildMappedScopes(
+            sourceRecord.toJS(),
+            frame,
+            await scopes,
+            sourceMaps
+          );
+        } catch (e) {
+          return null;
+        }
       })()
     });
   };
@@ -73,7 +74,7 @@ async function buildMappedScopes(
   frame: Frame,
   scopes: Scope,
   sourceMaps: any
-): Promise<?RenderableScope> {
+): Promise<?OriginalScope> {
   const originalAstScopes = await getScopes(frame.location);
   const generatedAstScopes = await getScopes(frame.generatedLocation);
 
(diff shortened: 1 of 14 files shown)


src/reducers/pause.js
```diff
@@ -11,16 +11,32 @@
...
   frames: ?(any[]),
-  frameScopes: any,
+  frameScopes: {
+    generated: {
+      [string]: {
+        pending: boolean,
+        scope: Scope
+      }
+    },
+    mapped: {
```
jasonLaster: perhaps `original` to be consistent

src/reducers/pause.js
```diff
@@ -11,16 +11,32 @@
  */
 
 import { createSelector } from "reselect";
+import { isGeneratedId } from "devtools-source-map";
 import { prefs } from "../utils/prefs";
+import { getSelectedSource } from "./sources";
 
+import type { RenderableScope } from "../utils/pause/scopes/getScope";
```
jasonLaster: could this be `OriginalScope` to be parallel with Scope

src/reducers/pause.js
```diff
@@ -11,16 +11,32 @@
...
+import type { Why, Scope, SourceId } from "debugger-html";
 
 type PauseState = {
   why: ?Why,
   isWaitingOnBreak: boolean,
   frames: ?(any[]),
-  frameScopes: any,
+  frameScopes: {
+    generated: {
+      [string]: {
```
jasonLaster: could this be `FrameId` not sure if we have it... alias for string

src/reducers/pause.js
```diff
@@ -198,22 +252,36 @@ export function getFrames(state: OuterState) {
...
   }
 
-  return state.pause.frameScopes[frameId];
+  return state.pause.frameScopes.generated[frameId];
 }
 
-export function getSelectedScope(state: OuterState) {
-  const frameId = getSelectedFrameId(state);
-  return getFrameScope(state, frameId);
+export function getFrameScope(
```
jasonLaster: ha not needed now but `getFrameScopes` would be simpler... its just so hard to different between plural and signular!

src/reducers/pause.js
```diff
@@ -198,22 +252,36 @@ export function getFrames(state: OuterState) {
...
+  sourceId: ?SourceId,
+  frameId: ?string
+) {
+  if (!frameId || !sourceId) {
+    return null;
+  }
+
+  const frameScopes = isGeneratedId(sourceId)
+    ? state.pause.frameScopes.generated
+    : state.pause.frameScopes.mapped;
```
jasonLaster: :thumbsup:

src/components/SecondaryPanes/Scopes.js
```diff
@@ -96,12 +104,18 @@ class Scopes extends PureComponent<Props, State> {
...
-    const frameScopes = selectedFrame
-      ? getFrameScope(state, selectedFrame.id)
-      : null;
+    const selectedSource = getSelectedSource(state);
+
+    const { scope: frameScopes, pending } = getFrameScope(
+      state,
+      selectedSource && selectedSource.get("id"),
+      selectedFrame.id
+    ) || { pending: false };
```
jasonLaster: should this `{ pending: false };` be in the selector?

so if we dont have scopes we just return `{ pending: false, scope: null };`

src/actions/pause/mapScopes.js
```diff
@@ -24,57 +27,91 @@ import type {
...
+      !sourceRecord.get("isPrettyPrinted") &&
+      !isGeneratedId(frame.location.sourceId);
 
     dispatch({
       type: "MAP_SCOPES",
       frame,
-      scopes: mappedScopes
+      [PROMISE]: (async function() {
+        let mappedScopes;
+        if (shouldMapScopes) {
```
jasonLaster: minor thing:

```js
if (!shouldMapScopes) {
  return scopes;
}

...
return mappedScopes
```

loganfsmyth: I didn't do that at the time because the `|| scopes` logic is needed even when `shouldMapScopes` is true, but since we talked about changing that to store `null` and have the selector fallback to generated, I can definitely do `return null` early here now.

I generally find that early returns any time there's something more complicated than `null` just end up leading to cases where you assume something always runs and end up being wrong.

src/components/SecondaryPanes/Scopes.js
loganfsmyth: I don't love the idea of returning something when it wasn't actually in the state. It also depends on if you want to differentiate between "loaded but not found", and "not loaded at all". You're right that we don't care here, but I don't really love making that assumption for all possible uses of `getFrameScope`.

src/actions/pause/mapScopes.js
jasonLaster: lol that is fair

src/reducers/pause.js
loganfsmyth: The type is called `Scope`, so unless we want to rename all of that too, this seems more consistent.

src/components/SecondaryPanes/Scopes.js
jasonLaster: yeah - that is true.

Issue #5241, opened 2018-01-29
[Preview] Show empty values


jasonLaster:
We recently regressed when it comes to showing tooltips for empty/falsey values

STR:
1. navigate to https://index-empties.glitch.me/
2. click "stuff"
3. hover on falsyString, falsyBool, falsyNumber, undefinedThing, nullThing

```js
function stuff() {
  let falsyString = "";
  let falsyBool = false;
  let falsyNumber = 0;
  let undefinedThing = undefined;
  let nullThing = null;

  let truthyString = " ";
  let truthyBool = true;
  let truthyNumber = 1;

  debugger;
}
```

Fischer-L:
I can reproduce this bug but one thing confuses me, which is, while hovering on `falsyNumber` I could observed[1] the preview prop like
```
Object { expression: "falsyNumber", result: 0, location: {…}, tokenPos: {…}, cursorPos: DOMRect, extra: {}, updating: false }
```
But no preview showed for `falsyNumber`.

The same prop for `undefinedThing` also observed
```
Object { expression: "undefinedThing", result: {…}, location: {…}, tokenPos: {…}, cursorPos: DOMRect, extra: {}, updating: false }
```
and a preview showed for `undefinedThing`

[1][https://github.com/devtools-html/debugger.html/blob/6d4220ecc43e4613f5db8feda6b2c25676f46259/src/components/Editor/Preview/index.js#L137](https://github.com/devtools-html/debugger.html/blob/6d4220ecc43e4613f5db8feda6b2c25676f46259/src/components/Editor/Preview/index.js#L137)

Fischer-L:
@jasonLaster Would you see PR #5279? thanks

PR #5279, opened 2018-02-01
5241 - [Preview] Show empty values

Fischer-L:
Associated Issue: #5241

### Screenshots/Videos
jufkidihcu

### DIFF
--- a/src/components/Editor/Preview/Popup.js
+++ b/src/components/Editor/Preview/Popup.js
@@ -211,7 +211,7 @@ export class Popup extends Component<Props> {
     );
   }
 
-  renderSimplePreview(value: Object) {
+  renderSimplePreview(value: any) {
     const { openLink } = this.props;
     return (
       <div className="preview-popup">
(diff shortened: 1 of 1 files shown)


src/components/Editor/Preview/Popup.js
```diff
@@ -241,9 +241,6 @@ export class Popup extends Component<Props> {
 
   renderPreview() {
     const { value } = this.props;
-    if (!value) {
```
Fischer-L: Found this is a regression from PR #4915
In that PR, added one `if (!value)` early return[1] so `this.renderSimplePreview(value)` below for simple typed value, such "", 0, null, wouldn't be reached. 

[1] [https://github.com/devtools-html/debugger.html/blob/6d4220ecc43e4613f5db8feda6b2c25676f46259/src/components/Editor/Preview/Popup.js#L244](https://github.com/devtools-html/debugger.html/blob/6d4220ecc43e4613f5db8feda6b2c25676f46259/src/components/Editor/Preview/Popup.js#L244)

nyrosmith: Mind leaving a comment so it becomes obvious that there should not be an early return because this breaks things?

AnshulMalik:
Looking good to me, just need to hack it to please flow, `value.class === "Function"` => `value && value.class === "Function"`

src/components/Editor/Preview/Popup.js
Fischer-L: @nyrosmith That helps in the future. Added

Fischer-L:
@AnshulMalik 

> Looking good to me, just need to hack it to please flow,
value.class === "Function" => value && value.class === "Function"

Thanks for reminding this improvement.

PR #5290, opened 2018-02-01
Fix #5258 - Group class functions within Outline Panel

### LINKED ISSUE #5258: In the Outline panel, classes should be more visible
At the moment, in the outline panel, if you have a class we show it's name, then its methods, which is really nice.

```
A
  λ foo
  λ bar
```

I think it would be nicer if we could have a more visible class so if you have multiple classes in the file, it is more visible

```
class A
  λ foo
  λ bar
```

here the  `class` keyword was added, and it would be nice if it could have the same color as in the editor.

It could be something like: 

image

Note that the class name could also be of the same color than in the editor (blue)

darkwing:
Associated Issue: #5258 

### Summary of Changes

* Adds a separate class heading and list for functions of the class
* Will use the same color highlighting as the editor highlighter

### Test Plan

I'm using the following test file:  https://gist.github.com/darkwing/3946523182fb40cdd1d10e1401ef481d

Example test plan:

- Open file linked above, ensure that functions are under correct class heading

### Screenshots/Videos

5258-screenshot

### TODO
- [ ] Test(s)
- [x] Add syntax highlighting for `class` and function name keywords

### DIFF
--- a/src/components/PrimaryPanes/Outline.js
+++ b/src/components/PrimaryPanes/Outline.js
@@ -61,21 +61,17 @@ export class Outline extends Component<Props> {
     );
   }
 
-  renderClassFunctions(functions: SymbolDeclaration[]) {
-    const classFunctions = functions.filter(
-      func => func.name != "anonymous" && !!func.klass
-    );
-
-    if (classFunctions.length == 0) {
+  renderClassFunctions(klass: string, functions: SymbolDeclaration[]) {
+    if (klass == null || functions.length == 0) {
       return null;
     }
 
-    const klass = classFunctions[0].klass;
-    const klassFunc = functions.find(func => func.name === klass);
+    const classFunc = functions.find(func => func.name === klass);
+    const classFunctions = functions.filter(func => func.klass === klass);
 
     return (
-      <div className="outline-list__class">
-        <h2>{klassFunc ? this.renderFunction(klassFunc) : klass}</h2>
+      <div className="outline-list__class" key={klass}>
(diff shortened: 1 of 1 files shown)


src/components/PrimaryPanes/Outline.js
```diff
@@ -61,21 +61,17 @@ export class Outline extends Component<Props> {
...
-
-    if (classFunctions.length == 0) {
+  renderClassFunctions(klass: string, functions: SymbolDeclaration[]) {
+    if (klass == null || functions.length == 0) {
       return null;
     }
 
-    const klass = classFunctions[0].klass;
-    const klassFunc = functions.find(func => func.name === klass);
+    const classFunc = functions.find(func => func.name === klass);
```
jasonLaster: hehe `klass` is from my ruby days :) no fun!

src/components/PrimaryPanes/Outline.js
```diff
@@ -61,21 +61,17 @@ export class Outline extends Component<Props> {
...
+    const classFunctions = functions.filter(func => func.klass === klass);
```
darkwing: This filtering can probably be moved to the `renderClassFunctions` call.

src/components/PrimaryPanes/Outline.js
darkwing: Ha! It makes sense here because `klass` is the preceding const but with that being moved it makes sense to change its spelling.

darkwing:
I think this is ready for a look @jasonLaster !  Thank you!

jasonLaster:
Looking at the editor and outline made me think we might want to try to re-use the editor font

Issue #5252, opened 2018-01-31
[Scopes] Support <this> and arguments bindings in new original scopes work


loganfsmyth:
With the work in https://github.com/devtools-html/debugger.html/pull/5215 DevTools can now render scopes based on the original source file, then attempts to map that back to the generated file.

That new work doesn't yet attempt to support `this` and `arguments` properly. I'm not entirely sure if it is 100% doable, but I'd like to look into it a bit before doing firmly one way or the other.

PR #5291, opened 2018-02-02
Map original 'this' and 'arguments' to correct bindings where possible

loganfsmyth:
Associated Issue: #5252

### Summary of Changes

* Fixes a few bugs in the visitor scope analysis
* Uses scope information to try to map `this` and `arguments`

### Test Plan

Kind of hard to test this one because Babel at least doesn't generate maps with good mappings for these. I'll make a PR for babel to map them properly.

These shots are taken with a local copy of Babel that actually generates good maps for `this` and `arguments` 😬 

* Note the successful inclusion of the correct `this` in this case, even though the user is inside of a transformed arrow function.

    screen shot 2018-02-01 at 5 16 40 pm

* Here note that `this` can only be transformed if there is a `this` reference in the same function scope as the breakpoint itself, as the debugger server does not send us binding information about `this` in other scopes. `arguments` still works because the parent scope has a reference to `arguments`. If neither location referenced either one, they would be excluded from the scope list entirely.

    screen shot 2018-02-01 at 5 33 57 pm

### DIFF
--- a/src/actions/pause/mapScopes.js
+++ b/src/actions/pause/mapScopes.js
@@ -84,7 +84,8 @@ async function buildMappedScopes(
 
   const generatedAstBindings = buildGeneratedBindingList(
     scopes,
-    generatedAstScopes
+    generatedAstScopes,
+    frame.this
   );
 
   const mappedOriginalScopes = await Promise.all(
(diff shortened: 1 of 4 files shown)


loganfsmyth:
I've submitted a PR for Babel so in the future these will map better. https://github.com/babel/babel/pull/7312

Issue #5262, opened 2018-01-31
Remove λ Svg in Outline view


nchevobbe:
In the Outline view we use a SVG to show a "lamba" letter before functions.
Turns out there is a unicode sign for lambda: `λ`

This does not look too bad I think (even better I'd say) : 

![Current vs Proposal](https://user-images.githubusercontent.com/578107/35613290-f6410aa8-066b-11e8-9bbc-2db94e389563.png)

Also, we know SVG can be perf sensitive, so having a plain text here is probably better.

WIP: 
```diff
diff --git a/src/components/PrimaryPanes/Outline.css b/src/components/PrimaryPanes/Outline.css
index 382c214b..e4397f3c 100644
--- a/src/components/PrimaryPanes/Outline.css
+++ b/src/components/PrimaryPanes/Outline.css
@@ -43,6 +43,10 @@
   white-space: nowrap;
 }

+.outline-list__class .outline-list__element {
+  padding-inline-start: 2rem;
+}
+
 .outline-list__element:hover {
   background: var(--theme-toolbar-background-hover);
 }
diff --git a/src/components/PrimaryPanes/Outline.js b/src/components/PrimaryPanes/Outline.js
index 8d1fe746..9c6c6dc0 100644
--- a/src/components/PrimaryPanes/Outline.js
+++ b/src/components/PrimaryPanes/Outline.js
@@ -9,7 +9,6 @@ import { bindActionCreators } from "redux";
 import { connect } from "react-redux";
 import actions from "../../actions";
 import { getSelectedSource, getSymbols } from "../../selectors";
-import Svg from "../shared/Svg";

 import "./Outline.css";
 import PreviewFunction from "../shared/PreviewFunction";
@@ -55,13 +54,12 @@ export class Outline extends Component<Props> {
         className="outline-list__element"
         onClick={() => this.selectItem(location)}
       >
-        <Svg name="function" />
-        <PreviewFunction func={{ name, parameterNames }} />
+        λ <PreviewFunction func={{ name, parameterNames }} />
       </li>
     );
   }

```

iansu:
Hi, I'd like to work on this.

PR #5293, opened 2018-02-02
Replace function SVG with unicode character in Outline pane

iansu:
Associated Issue: #5262 

### Summary of Changes

* Replace function SVG icon with the unicode function symbol in the Outline pane

### Test Plan

- [x] Open the Outline pane and look at the function symbol to the left of any function name

### Screenshots

|Before|After|
|----------|------|
|Before|After|


src/components/PrimaryPanes/Outline.js
```diff
@@ -55,8 +54,7 @@ export class Outline extends Component<Props> {
         className="outline-list__element"
         onClick={() => this.selectItem(location)}
       >
-        <Svg name="function" />
-        <PreviewFunction func={{ name, parameterNames }} />
+        λ <PreviewFunction func={{ name, parameterNames }} />
```
lukaszsobek: We do potentially want a way to style this element individually, so a tag around it would be nice.

iansu: Will do. I think that might also fix the failing test.

jasonLaster:
should be an easy fix

jasonLaster:
Welcome aboard @iansu!

iansu:
I was just about to fix the failing test but it looks like you beat me to it. Thanks! 😀

jasonLaster:
haha - no worries. mochitests are weird

Issue #5265, opened 2018-01-31
[Outline] default value parameters are skipped


nchevobbe:
**Steps to reproduce:**
1. Go to `data:text/html,hi<script>function myFunction(first = defaultValue, second){};</script>`
2. Open the debugger
3. Select the only script and go to the outline panel

**Expected results:**

There is the following function displayed
`λ myFunction(first = defaultValue, second)`

or at least 

`λ myFunction(first, second)`

**Actual results:**

The following function is displayed

`λ myFunction(second)`

i.e. the first parameter is not displayed.

nchevobbe:
So here: 
https://github.com/devtools-html/debugger.html/blob/56848d5a65d41121efb2015ed777095e0fbec254/src/components/PrimaryPanes/Outline.js#L50

parameterNames is `[undefined, "second"]`, so I guess this is an issue when parsing the functions in the file.

PR #5294, opened 2018-02-02
show default value parameters in Outline

nyrosmith:
Associated Issue: #5265

default_paramst


src/workers/parser/tests/__snapshots__/getSymbols.spec.js.snap
```diff
@@ -841,97 +841,28 @@ hasJsx: false"
 
 exports[`Parser.getSymbols function names 1`] = `
 "functions:
```
jasonLaster: why are we losing the snapshot data?

src/workers/parser/getSymbols.js
```diff
@@ -44,7 +44,18 @@ export type SymbolDeclarations = {
...
+        return param.name;
+      } else if (
+        // Parameter with default value
+        param.type === "AssignmentPattern" &&
+        param.left.type === "Identifier" &&
+        param.right.type === "Identifier"
+      ) {
+        return `${param.left.name} = ${param.right.name}`;
+      }
+    });
```
jasonLaster: is there now a case that we're not handling?

src/workers/parser/tests/__snapshots__/getSymbols.spec.js.snap
nyrosmith: I don't know I just updated the snapshots to reflect the new test case

src/workers/parser/getSymbols.js
nyrosmith: Yeah. Eg objects, arrays or functions as default values.
Right now I have no idea how these should be displayed

nyrosmith: I'll try to sketch out how to display the other cases over the weekend

darkwing:
I was talking to @jasonLaster yesterday about this same idea!  Great minds!

I was thinking we could do more of an "expand to view args" functionality; one thing I like about this pane is that it's simple.  Thoughts?

jasonLaster:
+1 on keeping it simple!

In the future i'd love to show imports and exports here too!

src/workers/parser/tests/__snapshots__/getSymbols.spec.js.snap
```diff
@@ -881,6 +882,8 @@ variables:
 [(25, 13), (25, 31)]   foo
 [(26, 12), (26, 42)]
 [(26, 13), (26, 36)]   foo
```
jasonLaster: looks like we're missing `first` variable here

src/workers/parser/tests/fixtures/functionNames.js
```diff
@@ -38,3 +38,5 @@ class Cls {
 (function(){});
 
 export default function (){}
+
+function def(first = defaultValue, second){}
```
jasonLaster: we didn't get any values because `function(first = defaultValue, second){}` is a syntax error

nyrosmith:
default_params

We are now showing much more cases than before. I checked every default assignment I could think of.
Caveats: Inline objects and arrays are always displayed as empty. I couldn't find an appropriate UI representation if they have values...

nyrosmith:
default_params

We are now showing much more cases than before. I checked every default assignment I could think of.
Caveats: Inline objects and arrays are always displayed as empty. I couldn't find an appropriate UI representation if they have values...

src/workers/parser/getSymbols.js
```diff
@@ -44,7 +44,50 @@ export type SymbolDeclarations = {
...
+          param.left.type === "Identifier" &&
+          param.right.type === "NullLiteral"
+        ) {
+          return `${param.left.name} = null`;
+        }
+      } else {
+        return param.name;
+      }
+
+      // if (param.name) {
```
nyrosmith: I forgot to delete this

jasonLaster:
Great work!

PR #5296, opened 2018-02-02
[WIP] Clear the sourceMapURL on fail to load source maps in order to enable pretty printing

AlexisDeschamps:
Associated Issue: #1661
Reliant on https://github.com/devtools-html/devtools-core/pull/921 so `fetchSourceMap` properly throws error

### Summary of Changes
*  Started clearing the sourceMapURL on fail to load source maps
* Added unit test to test the above

### Test Plan
Setup to get a source with a source map that fails on load:
- [x] Clone https://github.com/spotify/web-api-auth-examples
- [x] Run the following :
    $ npm install
    $ cd authorization_code
    $ node app.js
- [x] Simulate a fail to load the source map. Example method: Add `return null` to line 45 of `actions/newSources`

To test the patch
- [x] Open `code.jquery.com` folder from the sources
- [x] Click on `jquery-1.10.1.min.js`
- [x] `jquery-1.10.1.min.js` is automatically pretty printed (as opposed to being unable to pretty print before the patch)

### Screenshots
Before:
image

After:
image

### DIFF
--- a/src/actions/sources/newSources.js
+++ b/src/actions/sources/newSources.js
@@ -39,25 +39,25 @@ function createOriginalSource(
   };
 }
 
-// TODO: It would be nice to make getOriginalURLs a safer api
-async function loadOriginalSourceUrls(sourceMaps, generatedSource) {
-  try {
-    return await sourceMaps.getOriginalURLs(generatedSource);
-  } catch (e) {
-    console.error(e);
-    return null;
-  }
-}
-
 /**
  * @memberof actions/sources
  * @static
  */
-function loadSourceMap(generatedSource) {
+export function loadSourceMap(generatedSource: Source) {
   return async function({ dispatch, getState, sourceMaps }: ThunkArgs) {
-    const urls = await loadOriginalSourceUrls(sourceMaps, generatedSource);
+    let urls;
+    try {
+      urls = await sourceMaps.getOriginalURLs(generatedSource);
+    } catch (e) {
+      console.error(e);
+      urls = null;
+    }
     if (!urls) {
-      // If this source doesn't have a sourcemap, do nothing.
+      // If this source doesn't have a sourcemap, enable it for pretty printing
+      dispatch({
+        type: "UPDATE_SOURCE",
+        source: { ...generatedSource, sourceMapURL: "" }
+      });
       return;
     }
 
--- a/src/test/mochitest/browser_dbg-sourcemaps-reload.js
+++ b/src/test/mochitest/browser_dbg-sourcemaps-reload.js
@@ -53,7 +53,6 @@ add_task(async function() {
   await reload(dbg);
   await waitForSource(dbg, "v1");
   await syncBp;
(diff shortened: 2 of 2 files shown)


AlexisDeschamps:
Hmm, I don't actually have any reproduction steps. I'll try to find some.

codehag:
Hi! Do you need any help with tests or anything?

AlexisDeschamps:
@codehag Yeah, thanks. I updated the branch with changes if you want to review.

But the Travis CI build is failing. Not too sure why...

jasonLaster:
@AlexisDeschamps where is this at?

jasonLaster:
rebased and pushed. it should pass

src/actions/sources/newSources.js
```diff
@@ -40,11 +40,14 @@ function createOriginalSource(
 }
 
 // TODO: It would be nice to make getOriginalURLs a safer api
-async function loadOriginalSourceUrls(sourceMaps, generatedSource) {
+async function loadOriginalSourceUrls(dispatch, sourceMaps, generatedSource) {
   try {
```
jasonLaster: this function might have outlived it's usefulness. Perhaps we kill it and inline it back into `loadSourceMap`

src/test/mochitest/browser_dbg-sourcemaps-reload.js
```diff
@@ -53,7 +53,6 @@ add_task(async function() {
   await reload(dbg);
   await waitForSource(dbg, "v1");
   await syncBp;
```
jasonLaster: wow why is that not needed?

AlexisDeschamps: From my debugging, it seems that the source is already selected, so the `await` would just wait forever, causing the timeout. It also doesn't seem to be necessary for the last assessment of the number of breakpoints.

src/actions/sources/newSources.js
AlexisDeschamps: I think that would make sense.

src/test/mochitest/browser_dbg-sourcemaps-reload.js
jasonLaster: how does it pass now?

also if it is already selected it should pass immediately

AlexisDeschamps: Well it passes because I removed the line, I guess. It passed without the line removal before these changes.

I don't really see how my change would affect the behavior of the test though. It seems kind of convoluted overall since it's putting code from different files (`v1.js`, `v2.js`, `v3.js`) under the `v1` source.

If `waitForSelectedSource` is supposed to pass immediately if it is already selected then I guess ` return dbg.selectors.hasSymbols(state, source.toJS());` would be hanging forever for some reason.

jasonLaster:
Cool. This is ready to go once the one function is inlined back into the caller.

jasonLaster:
:boom:

PR #5299, opened 2018-02-02
Toggling the conditional bp pane should not open bookmarks

jasonLaster:
### Summary of Changes

cmd+shift+b should toggle the cb panel without opening the bookmarks bar :)


src/components/Editor/index.js
```diff
@@ -257,6 +258,13 @@ class Editor extends PureComponent<Props, State> {
     }
   };
 
+  onTogleConditionalPanel = (key, e) => {
```
darkwing: typo: `togle` -> `toggle`

darkwing:
This works but IMO we should focus on the breakpoint's `input` when opened -- looking into fixing that now

jasonLaster:
@darkwing there is an issue for it. And Mike cote is on it :)

PR #5304, opened 2018-02-03
improve dbg

jasonLaster:
### Summary of Changes

* adds docs for `dbg`
* adds `dbg.helpers.findSource`, `dbg.helpers.evaluate`
* binds `dbg.selectors` which makes it super easy to fire off queries like `dbg.selectors.getBreakpoints()`

### DIFF
--- a/docs/dbg.md
+++ b/docs/dbg.md
@@ -3,7 +3,7 @@
 
 ### store
 
-`dbg.store` is a convenient way for getting the redux store.
+`dbg.store` redux store.
 
 ```js
 const store = dbg.store.getState()
@@ -14,7 +14,7 @@ const source = dbg.store.getState().sources.sources.first();
 
 ### actions
 
-`dbg.actions` is a convenient way for getting the redux actions.
+`dbg.actions` redux actions.
 The actions are bound so you can call them directly!
 
 ```js
@@ -24,7 +24,7 @@ dbg.actions.selectSource(source.get("id")))
 
 ### selectors
 
-`dbg.selectors` is a convenient way for getting the redux selectors.
+`dbg.selectors` redux selectors.
 The selectors are bound so you can call them without `store.getState()`!
 
 ```js
@@ -33,7 +33,7 @@ const source = dbg.selectors.getSelectedSource();
 
 ### client
 
-`dbg.client` is a convenient way for getting the firefox commands.
+`dbg.client` firefox commands.
 The commands are the interface for talking to the debugger server.
 
 ```js
@@ -42,48 +42,6 @@ dbg.client
   .setBreakpoint({line: 24, sourceId: source.get("id")})
   .then(console.log)
 ```
-
-### connection
-
-`dbg.connection` has several useful properties for seeing the current connection to the tab.
-If you're looking for the current threadClient, debuggerClient, or tabTarget it is your friend!
-
-```js
-dbg.connection.tab
-/*
-{
-  "title": "Backbone.js • TodoMVC",
-  "url": "http://firefox-dev.tools/debugger-examples/examples/todomvc/",
-  "id": "server1.conn12.child1/tab1",
-  "tab": {
-    "actor": "server1.conn12.child1/tab1",
-    "title": "Backbone.js • TodoMVC",
-    "url": "http://firefox-dev.tools/debugger-examples/examples/todomvc/",
-    "outerWindowID": 2147483649,
-    "consoleActor": "server1.conn12.child1/consoleActor2",
-    "inspectorActor": "server1.conn12.child1/inspectorActor3",
-    "callWatcherActor": "server1.conn12.child1/callWatcherActor4",
-  },
-  "clientType": "firefox"
-}
-*/
-
-dbg.connection.connTarget // { type: "firefox", param: "child1/tab1" }
-
-dbg.connection.tabConnection // debuggerClient, threadClient, tabTarget
-```
-
-### bpClients
-
-`dbg.bpClients` stores the current set of breakpoint clients.
-
-### services
-
-`dbg.services` has a reference to the SourceMap worker, which is available to the redux actions.
-
-`dbg.services.sourceMaps`
-
-> originalToGeneratedId, generatedToOriginalId, isGeneratedId, isOriginalId, hasMappedSource, getOriginalURLs, getGeneratedLocation, getOriginalLocation, getLocationScopes, getOriginalSourceText, applySourceMap, clearSourceMaps, startSourceMapWorker, stopSourceMapWorker
 
 ### prefs
 
@@ -110,7 +68,7 @@ dbg.features.codeCoverage // true
 
 #### findSource
 
-Sometimes you want to quickly grab a source object from the store, but you dont know the full url or source id. The `findSource` function is your friend here :)
+Sometimes you want to quickly grab a source object from the store, but you don't know the full url or source id. The `findSource` function is your friend here :)
 
 ```js
 dbg.helpers.findSource("todo-view")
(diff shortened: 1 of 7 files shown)


jasonLaster:
after thinking about it... i think the markdown docs should just be inline comments...

wldcordeiro:
You could make `dbg.help()` and have it return the different helper docs you wrote? Like if you do `dbg.help()` it shows a quick list of topics and then you can pass in the string like `dbg.help('selectors')`?

src/utils/dbg.js
```diff
@@ -2,18 +2,37 @@ import { bindActionCreators } from "redux";
...
+  const source = sources.find(s => (s.get("url") || "").includes(url));
+
+  if (!source) {
+    return;
+  }
+
+  return source.toJS();
+}
+
+function sendPacket(dbg, packet, cbk) {
```
codehag: cbk stands for callback? this could be more readable if its fully written out

src/utils/dbg.js
```diff
@@ -2,18 +2,37 @@ import { bindActionCreators } from "redux";
...
+  dbg.connection.tabConnection.debuggerClient
+    .request(packet)
+    .then(cbk || console.log);
```
codehag: do we want to console.log if there is no callback? perhaps a better pattern would be to have a default empty function ?

src/utils/dbg.js
```diff
@@ -2,18 +2,37 @@ import { bindActionCreators } from "redux";
...
+}
+
+function evaluate(dbg, expression, cbk) {
+  dbg.client.evaluate(expression).then(cbk || console.log);
```
codehag: same comment as above

docs/dbg.md
```diff
@@ -0,0 +1,162 @@
...
+The actions are bound so you can call them directly!
+
+```js
+const source = dbg.store.getState().sources.sources.first();
+dbg.actions.selectSource(source.get("id")))
+```
+
+### selectors
+
+`dbg.selectors` is a convenient way for getting the redux selectors.
```
codehag: `is a convenient way` is used quite often, perhaps this sentence fragment is not needed in the above three cases, and we can keep this shorter -- that way there is less to read :)

docs/dbg.md
```diff
@@ -0,0 +1,162 @@
...
+The selectors are bound so you can call them without `store.getState()`!
+
+```js
+const source = dbg.selectors.getSelectedSource();
+```
+
+### client
+
+`dbg.client` is a convenient way for getting the firefox commands.
```
codehag: same as above

docs/dbg.md
```diff
@@ -0,0 +1,162 @@
...
+```js
+const source = dbg.selectors.getSelectedSource();
+dbg.client
+  .setBreakpoint({line: 24, sourceId: source.get("id")})
+  .then(console.log)
+```
+
+### connection
+
+`dbg.connection` has several useful properties for seeing the current connection to the tab.
```
codehag: maybe this could be clearer about what " connection" is (or what it is representing). At the moment I know from this that it has properties, but i don't know if it is something useful itself -- is like a grab-bag of these properties but not an actual connection instance, or is it a connection instance that reflects some data elsewhere is the question i find myself asking here

docs/dbg.md
```diff
@@ -0,0 +1,162 @@
...
+dbg.connection.tabConnection // debuggerClient, threadClient, tabTarget
+```
+
+### bpClients
+
+`dbg.bpClients` stores the current set of breakpoint clients.
+
+### services
+
+`dbg.services` has a reference to the SourceMap worker, which is available to the redux actions.
```
codehag: same comment as the connection comment. -- what *is* services? is it just a path to get to the sourcemaps? 

maybe if it is just a path we can omit "services" even

docs/dbg.md
```diff
@@ -0,0 +1,162 @@
...
+
+```js
+dbg.prefs.pauseOnExceptions // false
+dbg.prefs.pauseOnExceptions = true
+dbg.prefs.pauseOnExceptions // true
+```
+
+### features
+
+`dbg.features` references the debugger's feature flags. You can use `dbg.features` to see or change the state of any flag.
```
codehag: "`dbg.features` references the debugger's feature flags." is a great descriptive sentence about what features is!

docs/dbg.md
```diff
@@ -0,0 +1,162 @@
...
+
+```js
+dbg.features.codeCoverage // false
+dbg.features.codeCoverage = true
+dbg.features.codeCoverage // true
+```
+
+
+### helpers
+
```
codehag: maybe 

"the dbg.helpers are a collection of frequently used functions that help manage the data from the dbg. A good example of this is looking for sources...."

docs/dbg.md
```diff
@@ -0,0 +1,162 @@
...
+  "contentType": "text/javascript",
+  "isBlackBoxed": false,
+  "id": "server1.conn16.child1/source33"
+}
+*/
+```
+
+#### sendPacket
+
+`dbg.sendPacket` is an easy way to send a packet to the server. This is a useful helper for prototyping new APIs that aren't used in the UI yet.
```
codehag: this looks like it should be `dbg.helpers.sendPacket`

also, "an easy way" suggests there is another way... maybe to avoid confusion this can say: 

"dbg.sendPacket communicates with the server using the packet pattern outlined [here under "packets"](http://docs.firefox-dev.tools/backend/protocol.html)..."

in case we want people to have more context. plus linking is great!

docs/dbg.md
```diff
@@ -0,0 +1,162 @@
...
+  .sendPacket({
+    to: dbg.selectors.getSelectedFrame().id,
+    type: "evaluateExpressions"
+  })
+  .then(console.log)
+```
+
+#### evaluate
+
+`dbg.evaluate` is a convenient way of evaluating expressions in the context of the debuggee
```
codehag: nitpicking again -- using the wording " is a convenient way" ... suggests that there are other ways, but chances are we always want them to use this.

also, it looks like this should be dbg.helper.evaluate? according to line 150?

src/utils/dbg.js
```diff
@@ -2,18 +2,37 @@ import { bindActionCreators } from "redux";
...
+}
+
+function bindSelectors(obj) {
+  return Object.keys(obj.selectors).reduce((bound, selector) => {
     bound[selector] = (a, b, c) =>
```
codehag: maybe we can use the spread operator here to not lose any arguments?

docs/debugging.md
```diff
@@ -62,9 +61,10 @@ dbg.store.getState().sources.toJS();
 You can also test out a selector in the console with the `dbg helper:
 
 ```js
-dbg.selectors.getBreakpoints(dbg.store.getState());
+dbg.selectors.getBreakpoints();
```
codehag: should this be `dbg.helpers.selectors`?

docs/debugging.md
```diff
@@ -62,9 +61,10 @@ dbg.store.getState().sources.toJS();
 You can also test out a selector in the console with the `dbg helper:
```
codehag: this looks like its missing the closing \`

docs/dbg.md
jasonLaster: hmm... it can be other things too...

docs/debugging.md
jasonLaster: nope

src/utils/dbg.js
jasonLaster: i think it's nice to always see the response, if you don't care you can pass in `() => {}`

src/utils/dbg.js
jasonLaster: i don't know if that works here..

Issue #5306, opened 2018-02-03
[FileSearch] should not select ":" or "@" on re-open


lukaszsobek:
Currently when re-opening the `ctrl-shift-f` search modal that has a value inside we select the whole text. This is not really intuitive if the first character is `:` or `@`. In such cases the prefix should be omitted from the selection.

lukaszsobek:
Merged. Closing.

PR #5307, opened 2018-02-03
Omits '@:#' from selection in search modal

lukaszsobek:
Associated Issue: #5306 

Omits selecting the prefix from the input of the search input based on a `hasPrefix` flag set on the `<SearchInput/>` element.

Todo:
- [x] Add selection logic in QuickOpenModal.js
- [ ] Add selection logic in SearchBar.js
- [ ] Add selection logic in ProjectSearch.js
- [x] Update tests


darkwing:
I had a different take in `src/components/shared/SearchInput.js`:

```javascript
shouldShowErrorEmoji = () => {
    const { count, query, showErrorEmoji } = this.props;

    return (
      showErrorEmoji &&
      count === 0 &&
      query.trim() !== "" &&
      // Allow for gotoline format
      !query.match(/^:\d*$/)
    );
  };
```

That provides a check that will match ":", ":9", ":99999", but not ":99x9".  Thoughts?

lukaszsobek:
@darkwing The first thing I was wondering about is why we need to spread `this.props.showErrorEmoji` as I'm not sure what benefit referring the current state has.

isNaN was the first approach that came to mind, so I went with it. Now that you mantioned it - the regexp is a lot shorter, also being reasonably readable at the same time, so it seems the preferable solution. Will update the check.

lukaszsobek:
@darkwing 

Lol, just got confused about the issues when I started to answer. This is the pr for selecting the text :)

The first thing I was not really sure about was why we want to spread this.props.showErrorEmoji. One thing that came to mind was speed, but then it doesn't seem trim and match are that expensive in terms of processing time.

darkwing:
Ahh, sorry, I was addressing a different bug; carry on :)

src/components/shared/SearchInput.js
```diff
@@ -60,9 +60,9 @@ class SearchInput extends Component<Props> {
     if (this.$input) {
       const input = this.$input;
       input.focus();
-      if (input.value != "") {
-        input.select();
-      }
+
+      const selectStart = /^[@:#]/.test(input.value) ? 1 : 0;
+      input.setSelectionRange(selectStart, input.value.length + 1);
```
jasonLaster: SearchInput is a shared component now so i dont think it should know about the prefixes... `/^[@:#]` 

maybe we should have something like `showErrorEmoji` for disabling the select behavior...

Maybe we need to begin asking weather Search Input should be a shared component or just a part of quick open.

wldcordeiro:
One approach is making the shouldShowErrorEmoji thing in SearchInput work in one of two modes.

1. Only show if the prop is true
2. Use the prop + the count info (current implementation)

This way for scenarios like what we're trying to fix, we can validate in the consumer and not add to the SearchInput in ways that are specific to a single consumer.

codehag:
rebased and updated the snapshots! is this still a wip or ready to merge? if it is still a wip, do you have a todo list of what is left?

src/components/shared/SearchInput.js
```diff
@@ -60,9 +60,9 @@ class SearchInput extends Component<Props> {
     if (this.$input) {
       const input = this.$input;
       input.focus();
-      if (input.value != "") {
-        input.select();
-      }
+
+      const selectStart = this.props.searchType == "quickOpenModal" ? 1 : 0;
+      input.setSelectionRange(selectStart, input.value.length + 1);
```
jasonLaster: In this case - it's not that we don't want to select the `@` in `@fo` it's that we don't want to select the text `@` if that is the only thing in the input bar.

I'm not a big fan of the search input having logic based on all of the types. e.g. quick open also has types like function search. The reason for this is that when classes know about their consumers, it creates coupling that makes the complexity a function like N x M where M is the number of consumers.

It'd be better if SearchInput exposed an API about it's behavior for instance whether we should select the input text on mount `shouldSelectText`. This way, QuickOpen could have a predicate like `/^(#|@|:)$/`.

Issue #5268, opened 2018-01-31
[ToggleButton] Left side collapse pane icon is blurry


juliandescottes:
On my non-retina dispay the left-side collapse-icon is really blurry:

![image](https://user-images.githubusercontent.com/1141550/35620266-f0dbe0f0-0680-11e8-9380-922bc180cf2b.png)

Part of the issue seems to come from the 2.5px padding at
https://github.com/devtools-html/debugger.html/blob/e61cda3247ee8f593326b1186b665bffc0619938/src/components/shared/Button/PaneToggle.css#L12

Using padding 2px improves it but it still remains more blurry than the right one.

PR #5309, opened 2018-02-03
[ToggleButton] Fix for left side collapse pane icon was blurry

martinfojtik:
Associated Issue: #5268

Fix for left side collapse pane icon

#### BEFORE
devtoools-icon-before

#### AFTER
devtoools-icon-after

### DIFF
--- a/src/components/shared/Button/PaneToggle.css
+++ b/src/components/shared/Button/PaneToggle.css
@@ -8,12 +8,13 @@
   padding: 5px;
 }
 
-.toggle-button.vertical {
-  padding: 2.5px 2.5px;
+.toggle-button .togglePanes {
+  vertical-align: -2px;
 }
 
 .toggle-button svg {
   fill: var(--theme-comment);
+  vertical-align: 0;
 }
 
 :root.theme-dark .toggle-button svg {


src/components/shared/Button/CommandBarButton.css
```diff
@@ -27,9 +27,8 @@
   color: var(--theme-body-color);
 }
 
-.command-bar-button > * {
```
lukaszsobek: Is this the only icon affected by the rule?

src/components/shared/Button/PaneToggle.css
```diff
@@ -8,10 +8,6 @@
   padding: 5px;
 }
 
-.toggle-button.vertical {
```
lukaszsobek: I use horizontal mode, so not sure what removing this part changes. Is there insight you can share on this?

lukaszsobek:
The result looks great! Good job.

src/components/shared/Button/CommandBarButton.css
martinfojtik: yes, I didn't find any other icon which uses this class name

src/components/shared/Button/PaneToggle.css
martinfojtik: The `vertical` class is not referring to the vertical position of Devtools. Actually, I don't know why is there.

src/components/shared/Button/CommandBarButton.css
```diff
@@ -33,3 +33,7 @@
   display: inline-block;
   vertical-align: middle;
 }
+
+.command-bar-button .togglePanes {
```
lukaszsobek: Introducing this makes the top and bottom padding uneven. I'd rather avoid it if possible - can this be dropped with the fix working as intended?

martinfojtik: `vertical-align: middle;` will not force top and bottom paddings to be even, I added vertical-align in pixels.

darkwing:
I've been playing around with the CSS in this PR but it doesn't seem to have any effect:

notdoinganything

Could I be missing something?

martinfojtik:
@darkwing `vertical-align: -2px` is mainly for Windows OS

Issue #5157, opened 2018-01-20
[QuickOpen] Go To line should only show a sad face when the query is bad


jasonLaster:
STR:
1. cmd+:
2. type 12

AR: sad face
ER: should not be sad

If the user types `:sdlkjdsf` it should show a sad face because that query would be cray

![](http://g.recordit.co/dc6mZUbe1q.gif)

jbhoosreddy:
#5108

jbhoosreddy:
#5086 fixes this.

jasonLaster:
Hey @jbhoosreddy @atwalg2 @wldcordeiro I still see the issue.

Also the prefix is selected lol.. Damn this is a tough feature :)

<img src='https://shipusercontent.com/b3d944a0a33d42670172f260cae8e07b/Screen%20Shot%202018-02-02%20at%205.50.40%20PM.png' title='Screen Shot 2018-02-02 at 5.50.40 PM.png' width=769>

<img src='https://shipusercontent.com/96c8af125e402f6c006b1d3a3aeea13a/Screen%20Shot%202018-02-02%20at%205.51.07%20PM.png' title='Screen Shot 2018-02-02 at 5.51.07 PM.png' width=796>

lukaszsobek:
@jasonLaster the prefix is selected because we select everything there is in the input field on reopening. It will not be selected on the first open.

jasonLaster:
hmm, i feel like if you do `cmd+;`,`<esc>`,`cmd+;` it shouldnt be selected... can we make that work?

lukaszsobek:
@jasonLaster - will give it a go.

lukaszsobek:
Merged. Closing.

PR #5310, opened 2018-02-03
Makes error emoji behave sensibly

lukaszsobek:
Associated Issue: #5157 

Gist:
The error emoji should appear only when there is an actual error.  This creates two situations:
a.) where we are searching for a line number - in this case anything but a number should trigger the error emoji and
b.) where we are searching for text. In this case the emoji should only trigger if there is a search query AND the result count is 0.

p.s. The case of an @ or # is already handled well, so no need to address it

Vid:
source-search

### DIFF
--- a/src/components/Editor/SearchBar.js
+++ b/src/components/Editor/SearchBar.js
@@ -280,13 +280,16 @@ class SearchBar extends Component<Props, State> {
     );
   };
 
+  shouldShowErrorEmoji() {
+    return true;
+  }
+
   render() {
     const { searchResults: { count }, searchOn } = this.props;
 
     if (!searchOn) {
       return <div />;
     }
-
     return (
       <div className="search-bar">
         <SearchInput
@@ -295,6 +298,7 @@ class SearchBar extends Component<Props, State> {
           placeholder={L10N.getStr("sourceSearch.search.placeholder")}
           summaryMsg={this.buildSummaryMsg()}
           onChange={this.onChange}
+          showErrorEmoji={this.shouldShowErrorEmoji()}
           onKeyDown={this.onKeyDown}
           handleNext={e => this.traverseResults(e, false)}
           handlePrev={e => this.traverseResults(e, true)}
--- a/src/components/Editor/SearchBar.js
+++ b/src/components/Editor/SearchBar.js
@@ -281,7 +281,8 @@ class SearchBar extends Component<Props, State> {
   };
 
   shouldShowErrorEmoji() {
-    return true;
+    const { query, searchResults: { count } } = this.props;
+    return !!query && !count;
   }
 
   render() {
--- a/src/components/QuickOpenModal.js
+++ b/src/components/QuickOpenModal.js
@@ -356,6 +356,24 @@ export class QuickOpenModal extends Component<Props, State> {
     });
   };
 
+  shouldShowErrorEmoji() {
+    const { query } = this.props;
+
+    if (this.isGotoQuery()) {
+      return !/^:\d*$/.test(query);
+    }
+
+    if (
+      this.isVariableQuery() ||
+      this.isFunctionQuery() ||
+      this.isShortcutQuery()
+    ) {
+      return !this.getResultCount();
+    }
+
+    return false;
+  }
+
   render() {
     const { enabled, query } = this.props;
     const { selectedIndex, results } = this.state;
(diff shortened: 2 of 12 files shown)


src/components/shared/SearchInput.js
```diff
@@ -67,8 +67,8 @@ class SearchInput extends Component<Props> {
   }
 
   shouldShowErrorEmoji = () => {
-    const { count, query, showErrorEmoji } = this.props;
-    return showErrorEmoji && count === 0 && query.trim() !== "";
+    const { count, query } = this.props;
+    return count === 0 && query.trim() !== "" && !/^:\d*$/.test(query);
```
jasonLaster: So... I don't think our search input should know about go to line queries `!/^:\d*$/.test(query);`

Instead, it'd be better if we could use the `showErrorEmoji` flag so that QuickOpen can disable the error emoji when gotoline is set.

wldcordeiro: I had suggested this in the other PR similar to this and I think I agree. Maybe we do one of two things if showErrorEmoji is provided it turns on whether we even show it. Next we either just use the count and query or we use a consumer provided boolean value.

src/components/QuickOpenModal.js
```diff
@@ -356,6 +356,24 @@ export class QuickOpenModal extends Component<Props, State> {
...
+    if (this.isGotoQuery()) {
+      return !/^:\d*$/.test(query);
+    }
+
+    if (
+      this.isVariableQuery() ||
+      this.isFunctionQuery() ||
+      this.isShortcutQuery()
+    ) {
+      return !this.getResultCount();
```
lukaszsobek: It seems that going by this logic we could completely eliminate any checks in `shouldShowErrorEmoji()` in `SearchInput.js` in favor of just passing a true/false based on the checks in the corresponding dedicated module.

src/components/tests/__snapshots__/QuickOpenModal.spec.js.snap
```diff
@@ -343,7 +343,7 @@ exports[`QuickOpenModal Basic render with mount 1`] = `
               placeholder="Search sources…"
               query=""
               selectedItemId=""
-              showErrorEmoji={true}
+              showErrorEmoji={false}
```
lukaszsobek: The snapshots might need a rework after the logic is done since we are shifting the checks to a different place and `showErrorEmoji` will have different values.

lukaszsobek:
Current state:

Search modal:
modal-emoji

Project search:
project-emoji

Searchbar:
file-emoji

src/components/QuickOpenModal.js
jasonLaster: yeah - that's not a bad idea. 

We could also remove this ` !this.getResultCount();` check because it should be redundant

src/components/Editor/SearchBar.js
```diff
@@ -280,13 +280,16 @@ class SearchBar extends Component<Props, State> {
     );
   };
 
+  shouldShowErrorEmoji() {
+    return true;
```
jasonLaster: why do we always want to show the error?

lukaszsobek: There are three conditions for the error emoji to appear - this flag passed into SearchInput + count being 0 + search input not being empty. if we don't set should `shouldShowErrorEmoji` to true the sad face will never show since the flag will always exit on false.

src/components/Editor/SearchBar.js
```diff
@@ -281,7 +281,8 @@ class SearchBar extends Component<Props, State> {
   };
 
   shouldShowErrorEmoji() {
-    return true;
+    const { query, searchResults: { count } } = this.props;
```
codehag: this looks better!

src/components/shared/tests/SearchInput.spec.js
```diff
@@ -4,27 +4,32 @@ import { shallow } from "enzyme";
...
     />
   );
-  it("render", () => expect(wrapper).toMatchSnapshot());
-  it("show svg (emoji)", () => {
-    wrapper.setProps({ query: "test" });
+
+  it("renders", () => expect(wrapper).toMatchSnapshot());
+
+  it("shows svg error emoji", () => {
+    wrapper.setProps({ showErrorEmoji: true });
```
codehag: this is testing the behavior by just changing the emoji... however this change should reflect the count.. perhaps using the count was a better way of verifying the behavior?

A nice way to think of tests is to ask these two questions:

what is the expected input?
what is the expected output?

the input isn't really the showEmoji flag, its the number of results that come back.

Maybe a test to return to!

lukaszsobek: I'd say this is definitely a test to return to. I just had enough time to play around with it and see how it functions. Now comes the second part is rewriting it so that it actually tests something meaningful.

PR #5311, opened 2018-02-03
Fix [QuickOpenModal] initial scroll

### LINKED ISSUE #5284: [QuickOpen] modal scrolls when displayed
Steps to reproduce:
1. In a page with a good number of files, open the debugger in the firefox devtools, docked at the bottom (not sure this is important but I could not reproduce with the launchpad)
2. Open the Source modal (Cmd + P)

Expected results:
I see the modal, scrolled at the top

Actual results:

The modal is displayed, but already scrolled

feb-01-2018 16-21-52

kootoopas:
Fixes #5284.

### Summary of Changes

* Apply delay if modal has just been enabled or if query didn't change. *(although `nowEnabled` is always `false` since `prevProps.enabled` is always `true` for some reason 🤔. maybe it's the because the `Modal` is wrapped with a `Transition`)*
* modify `scrollList` to accept optional delay flag

### Test Plan

quickopen-modal-initial-scroll-fix

### DIFF
--- a/src/components/QuickOpenModal.js
+++ b/src/components/QuickOpenModal.js
@@ -96,12 +96,17 @@ export class QuickOpenModal extends Component<Props, State> {
   }
 
   componentDidUpdate(prevProps: Props) {
-    if (this.refs.resultList && this.refs.resultList.refs) {
-      scrollList(this.refs.resultList.refs, this.state.selectedIndex);
-    }
-
     const nowEnabled = !prevProps.enabled && this.props.enabled;
     const queryChanged = prevProps.query !== this.props.query;
+
+    if (this.refs.resultList && this.refs.resultList.refs) {
+      scrollList(
+        this.refs.resultList.refs,
+        this.state.selectedIndex,
+        nowEnabled || !queryChanged
+      );
+    }
+
     if (nowEnabled || queryChanged) {
       this.updateResults(this.props.query);
     }
--- a/src/components/shared/Modal.js
+++ b/src/components/shared/Modal.js
@@ -20,6 +20,8 @@ type ModalProps = {
   handleClose: () => any
 };
 
+export const transitionTimeout = 175;
+
 export class Modal extends React.Component<ModalProps> {
   onClick = (e: SyntheticEvent<HTMLElement>) => {
     e.stopPropagation();
@@ -59,7 +61,7 @@ export default function Slide({
   handleClose
 }: SlideProps) {
   return (
-    <Transition in={inProp} timeout={175} appear>
+    <Transition in={inProp} timeout={transitionTimeout} appear>
       {(status: TransitionStatus) => (
         <Modal
           status={status}
--- a/src/utils/result-list.js
+++ b/src/utils/result-list.js
@@ -3,9 +3,7 @@
  * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
 
 import { isFirefox } from "devtools-config";
-import {
-  transitionTimeout as modalTransitionTimeout
-} from "../components/shared/Modal";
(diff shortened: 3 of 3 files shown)


kootoopas:
Does circleci use prettier? The line that it throws for doesn't exceed 80.

src/components/QuickOpenModal.js
```diff
@@ -96,12 +96,17 @@ export class QuickOpenModal extends Component<Props, State> {
   }
 
   componentDidUpdate(prevProps: Props) {
+    const nowEnabled = !prevProps.enabled && this.props.enabled;
+    const queryChanged = prevProps.query !== this.props.query;
```
lukaszsobek: Not really needed to strict compare a value unless you expect it to be 0 or 1, this one is going to be a string.

src/components/shared/Modal.js
```diff
@@ -20,6 +20,8 @@ type ModalProps = {
   handleClose: () => any
 };
 
+export const transitionTimeout = 175;
```
lukaszsobek: Just curious, why should it be exactly 175?

kootoopas: That was the value that the `timeout` prop of `<Transition/>` already had. I just made it exportable. I'm not sure why we have a delay for the transition of modals to be honest. :|

src/components/QuickOpenModal.js
kootoopas: It is a string. :) I dunno, I have the habit of always using strict comparisons since they're more precise. Are there any performance decrements?

jasonLaster:
Hey @wldcordeiro mind taking a look. This touches your transition CSS logic which is kinda funky

src/components/shared/Modal.js
wldcordeiro: https://reactcommunity.org/react-transition-group/#Transition-prop-timeout

> The duration of the transition, in milliseconds. Required unless addEventListener is provided.

src/components/QuickOpenModal.js
AnshulMalik: Strict comparison should be faster

wldcordeiro: Personally I prefer the strict checks for all things except for when you're comparing to `null`/`undefined` where `if (foo == null)` is equivalent to `if (foo === null || foo === undefined)` but yeah. If I had a thing I'd prefer here it'd be this but only because I like explicitly plucking the props/prevProps I care about.

```js
componentDidUpdate({ enabled: prevEnabled, query: prevQuery }) {
  const { enabled, query } = this.props;
  const nowEnabled = !prevEnabled && enabled;
  const queryChanged = prevQuery !== query;
  // ...
}

File: src/utils/result-list.js [hunk 3]
```diff
@@ -3,19 +3,32 @@
  * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
 
 import { isFirefox } from "devtools-config";
+import {
```
wldcordeiro: Weird that CI is tripping here. Maybe just leave the original name?

src/components/shared/Modal.js
kootoopas: Huh, weird. I thought that the timeout was just a delay and when it ends, the transition is fired. I gave it a value of 2000ms just to check what it does and the transition speed of the modal appearance is the same. So maybe it is a delay. Am I missing something? 🤔 

I think that we can make it faster and give the modal appearance a more responsive feel if we make the timeout 25ms.

src/utils/result-list.js
kootoopas: done did it :)

Issue #4544, opened 2017-10-31
[WelcomeBox] is too low-contrast


bzbarsky:
This is the "Cmd-P to search for sources" and the like text.

The background here is rgb(249, 249, 250), looks like.  The text color is rgb(204, 209, 213).

Running through the basic calculations at https://www.w3.org/TR/AERT/#color-contrast (and note that these colors are all gray, so really color brightness is all that matters), we have a brightness of 249 for the background and a brightness of 207 for the foreground, for a brightness difference of 42.  The minimum recommendation for brightness difference is 125.

(That said, if you compute the color difference in this case, per the formulas on the linked page, that's 112, which is also way lower than the recommended minimum of 500).

jasonLaster:
Here are some screenshots:

<img src='https://shipusercontent.com/0976c23fb036567aea6ab8a55e3231ac/Screen%20Shot%202018-01-25%20at%201.35.03%20PM.png' title='Screen Shot 2018-01-25 at 1.35.03 PM.png' width=508>

<img src='https://shipusercontent.com/fb3781086aa3a0acfbd26ba9955cc729/Screen%20Shot%202018-01-25%20at%201.34.59%20PM.png' title='Screen Shot 2018-01-25 at 1.34.59 PM.png' width=351>

violasong:
Should match the colors of the other panels like so:

- File search keyboard shortcuts, other important elements like Modifiers buttons, source tree icons: var(--theme-body-color)
- Input placeholder color: Gabe says if .search-field input::placeholder is changed to black like the other panels, it will apply the right translucency to match.

ericawright:
@violasong 
<img width="853" alt="screen shot 2018-01-30 at 2 43 50 pm" src="https://user-images.githubusercontent.com/10803178/35587602-486811be-05cc-11e8-81a7-ada6d11a4427.png">
<img width="621" alt="screen shot 2018-01-30 at 2 44 00 pm" src="https://user-images.githubusercontent.com/10803178/35587600-4692a3ae-05cc-11e8-9321-cb26c58bb05a.png">

this is the current result - is the "modifiers" section a bit too dark? and did you want the search icon and "x" to also be darker? What about the x in the file tab?

@jasonLaster Also, is there a way I can see it in in dark theme, or do I have to run it with a firefox build for that?

jasonLaster:
@ericawright you can go back to the launchpad and then settings

jasonLaster:
also seems a bit dark here

[...]

PR #5314, opened 2018-02-05
color fixes

ericawright:
@violasong 
@jasonLaster 
Associated Issue: #4544

### Summary of Changes

* slightly darken "add watch expression" placeholder - matches search input now.
* slightly lighten welcomebox shortcut message - `var(--theme-comment)`;
* darken, override styling from https://github.com/devtools-html/devtools-core/blob/master/packages/devtools-components/src/tree.css#L73-L79 to have the svg arrows on the right side be `var(--theme-comment)` color.
* darken the svg arrows on the left side to be the same color.
* darken the x button in the search input
* darken the magnifying glass and ellipse in the search input

### Screenshots
screen shot 2018-02-05 at 11 09 00 am
screen shot 2018-02-05 at 11 08 33 am

screen shot 2018-02-05 at 11 07 44 am

screen shot 2018-02-05 at 10 30 26 am

#### notes:
* I'm not sure about the modifiers - right now they are all set to `var(--theme-body-color);` perhaps they should also be `var(--theme-comment);`

### DIFF
--- a/src/components/PrimaryPanes/Sources.css
+++ b/src/components/PrimaryPanes/Sources.css
@@ -51,10 +51,6 @@
   transform: rotate(0deg);
 }
 
-.theme-dark .sources-list .tree .node:not(.focused) img.arrow {
-  background: var(--theme-content-color3);
-}
-
 .theme-dark .source-list .tree .node.focused {
   background-color: var(--theme-tab-toolbar-background);
 }
--- a/src/components/SecondaryPanes/Expressions.css
+++ b/src/components/SecondaryPanes/Expressions.css
@@ -39,8 +39,7 @@
 .input-expression::placeholder {
   text-align: center;
   font-style: italic;
-  color: var(--theme-comment-alt);
-  opacity: 1;
+  color: var(--theme-comment);
 }
 
 .input-expression:focus {
--- a/src/components/SecondaryPanes/SecondaryPanes.css
+++ b/src/components/SecondaryPanes/SecondaryPanes.css
@@ -43,10 +43,6 @@
   cursor: default;
 }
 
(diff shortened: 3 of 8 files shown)


violasong:
This is looking GREAT - thanks for working on this Erica!

Couple little things:
- The disclosure arrows look a bit darker than they do in Inspector. I just checked and it looks like the Inspector actually uses a png from the "twisty" theme rather than color-able svg? ("chrome://devtools/skin/images/controls.png") That's pretty weird, but I think for now, it would be best for Debugger to match those colors (seems to be #B2B2B2 for light, #7F7F81 for dark).
- It looks like the source tree text may have too much contrast (we were recently lessening the dark mode contrast in this bug - maybe these changes didn't make it over? https://bugzilla.mozilla.org/show_bug.cgi?id=1421389) 
- If the globe, folder, and file icons in the source tree could also be updated to match this new magnifying glass color, that would be great

ericawright:
@violasong yep, the sidebar backgrounds are using grey-90, rather than #1B1B1D; which was changed in the bug you referenced.

ericawright:
@violasong it seems to be a variable issue. in moz central `--theme-sidebar-background` is defined as #1B1B1D, so my understanding is that it will appear differently when live in the browser. 
would result in this:
screen shot 2018-02-05 at 1 16 04 pm
screen shot 2018-02-05 at 1 15 44 pm

ericawright:
@jasonLaster all of @violasong's requests have been changed now.

wldcordeiro:
This was great. Nice improvements to the overall contrast and consistency!

violasong:
Just saw this in Nightly and my eyeballs breathed a big sigh of relief at the hearty levels of contrast! Thanks so much for working on this, @ericawright!

Issue #5084, opened 2018-01-11
[CallStack] clicking the selected frame does not jump


jasonLaster:
STR:
1. pause
2. scroll 
3. click the selected frame

ER: you jump to the line
AR: nothing happens

The reason for that is that the redux state does not change. One solution is to temporarily remove the selected frame and then re-add it.

PR #5316, opened 2018-02-05
5084 - [CallStack] clicking the selected frame does not jump

Fischer-L:
Associated Issue: #5084 

### Summary of Changes
Do `selectFrame` still on `onMouseDown`/`onKeyUp` even the frame given is selected already. To re-select the frame will trigger `selectLocation` again [1] so it will jump back to the paused line if scroll the line out of the view. If the line was within the view, then it just select the selected frame more one time.

[1] [https://github.com/devtools-html/debugger.html/blob/7985f7e5e17de7ae026b609b7e740696663c8a2b/src/actions/pause/selectFrame.js#L25](https://github.com/devtools-html/debugger.html/blob/7985f7e5e17de7ae026b609b7e740696663c8a2b/src/actions/pause/selectFrame.js#L25)

### DIFF
--- a/src/components/SecondaryPanes/Frames/Frame.js
+++ b/src/components/SecondaryPanes/Frames/Frame.js
@@ -87,7 +87,7 @@ export default class FrameComponent extends Component<FrameComponentProps> {
     frame: Frame,
     selectedFrame: Frame
   ) {
-    if (e.nativeEvent.which == 3 || selectedFrame.id === frame.id) {
+    if (e.nativeEvent.which == 3) {
       return;
     }
     this.props.selectFrame(frame);
@@ -98,7 +98,7 @@ export default class FrameComponent extends Component<FrameComponentProps> {
     frame: Frame,
     selectedFrame: Frame
   ) {
-    if (event.key != "Enter" || selectedFrame.id == frame.id) {
+    if (event.key != "Enter") {
       return;
     }
     this.props.selectFrame(frame);


jasonLaster:
Thanks!

jasonLaster:
i wonder how this updates if you just scroll out of view, because redux wont have changed...

Fischer-L:
@jasonLaster 
To re-select the frame will trigger `selectLocation` again [1], which will dispatch "SELECT_SOURCE" action. The "SELECT_SOURCE" reducer [2] will set the state again so the state's reference changes.

[1] [https://github.com/devtools-html/debugger.html/blob/7985f7e5e17de7ae026b609b7e740696663c8a2b/src/actions/pause/selectFrame.js#L25](https://github.com/devtools-html/debugger.html/blob/7985f7e5e17de7ae026b609b7e740696663c8a2b/src/actions/pause/selectFrame.js#L25)

[2] [https://github.com/devtools-html/debugger.html/blob/04a6801de5493f2567fe1342d8bb1406140e07d4/src/reducers/sources.js#L73](https://github.com/devtools-html/debugger.html/blob/04a6801de5493f2567fe1342d8bb1406140e07d4/src/reducers/sources.js#L73)

jasonLaster:
@bomsy i know you were in here recently. What do you think?

bomsy:
The change looks ok! i think selectSource might be remounting the source again..maybe ... will test and see

- We should test right click behaviors, before landing

bomsy:
Will test fully in a bit!

Fischer-L:
> @Fischer-L How are the right click behaviors?

@bomsy The code guards on the mouse event `which` == 3 so the right click would return early, The keyboard is similar; only `enter` works.

Issue #5184, opened 2018-01-24
[Stepping] Upon stepping onto certain statements, the editor realigns vertically to the leftmost frame of the editor.


mmcote:
When we refactored scrolling last month, we removed a check that used to ensure that we'd only scroll if the line were not visible. We should bring that back because now it's very jump 🏃

STR:

    

- Go to a webpage with heavily nested statements, such as the file given below.
- Add a breakpoint to the first if statement.
- Shrink the horizontal scroll view to as small as possible, ensuring some of the statements are out of sight.
- Step through the code to the most nested element and note how the code realigns vertically to the debugger statements, but realigns to the leftmost frame of the editor when debugging if statements.

```
<html>
  <head>
    <title>Debugger Statements</title>
  </head>

  <body>
    <script>
      test();

    function test() {
        if (1 + 1 == 2)
        {
            debugger;
            if (2 + 2 == 4)
            {
                debugger;
                if (3 + 3 == 6) 
                {
                    debugger;
                    if (4 + 4 == 8) 
                    {
                        debugger;
                        if (1 + 1 == 2) 
                        {
                            debugger;
                            if (2 + 2 == 4) 
                            {
                                debugger;
                                if (3 + 3 == 6) 
                                {
                                    if (4 + 4 == 8) 
                                    {
                                        debugger;
                                    }
                                }
                            }
                        }
                    }
                }
            }
        }
    }
    </script>
  </body>

</html>
```

![skippingbackandforth](https://user-images.githubusercontent.com/14250545/35311744-48a08f50-0075-11e8-9eeb-fc8e4841f07b.gif)

-----------------------------------------------

PR #5322, opened 2018-02-06
Determine the column of the current line being stepped on by the lead…

mmcote:
Determine the column of the current line being stepped on by the leading spaces, also used in debugLine to highlight the statement starting at the right column.

Associated Issue: #5184 

### Summary of Changes

- Parsed debug line code for leading whitespace to determine the column the line begins on.
- This change was implemented in the automatic scrolling of debugger and to fix the highlighting of the line being debugged.

### Screenshots
**Before**
Before the fix, it can be seen that as we step through the code being debugged, on certain statements the debugger does not center correctly, most notably when the code on the line being debugged is out of view. It is seen in these scenarios to realign the view with the leftmost edge of the editor.

Also before the fix, it can be seen that when stepping these statements the line was highlighted from the leftmost edge all the way to the right, where it is supposed to start highlighting the debugged line at the start of the code on the line.
01beforeiffix

**After**
After the fix, it can be seen that the editor realigns itself and starts highlighting the lines with the leftmost character of the example if statements. 
01afteriffix

### DIFF
--- a/src/components/Editor/index.js
+++ b/src/components/Editor/index.js
@@ -418,7 +418,13 @@ class Editor extends PureComponent<Props, State> {
     const { editor } = this.state;
 
     if (this.shouldScrollToLocation(nextProps)) {
-      const { line, column } = toEditorPosition(nextProps.selectedLocation);
+      let { line, column } = toEditorPosition(nextProps.selectedLocation);
+
+      const lineText = editor.editor.getLine(line);
+      const spaces = lineText.match(/^\s+/);
+      if (spaces) {
+        column = Math.max(column, spaces[0].length);
+      }
       scrollToColumn(editor.codeMirror, line, column);
     }
   }
--- a/src/components/Editor/tests/Editor.spec.js
+++ b/src/components/Editor/tests/Editor.spec.js
@@ -30,6 +30,9 @@ function createMockEditor() {
       defaultTextHeight: () => 0,
       display: { gutters: { querySelector: jest.fn() } }
     },
+    editor: {
+      getLine: line => ""
+    },
     setText: jest.fn(),
     on: jest.fn(),
     off: jest.fn(),
--- a/src/components/Editor/DebugLine.js
+++ b/src/components/Editor/DebugLine.js
@@ -61,9 +61,15 @@ export class DebugLine extends Component<Props> {
     const sourceId = selectedFrame.location.sourceId;
     const doc = getDocument(sourceId);
 
-    const { line, column } = toEditorPosition(selectedFrame.location);
+    let { line, column } = toEditorPosition(selectedFrame.location);
     const { markTextClass, lineClass } = this.getTextClasses(why);
     doc.addLineClass(line, "line", lineClass);
+
+    const lineText = doc.getLine(line);
+    const spaces = lineText.match(/^\s+/);
+    if (spaces) {
+      column = Math.max(column, spaces[0].length);
+    }
 
     this.debugExpression = doc.markText(
       { ch: column, line },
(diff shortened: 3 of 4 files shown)


src/components/Editor/index.js
```diff
@@ -418,7 +418,13 @@ class Editor extends PureComponent<Props, State> {
...
 
     if (this.shouldScrollToLocation(nextProps)) {
-      const { line, column } = toEditorPosition(nextProps.selectedLocation);
+      let { line, column } = toEditorPosition(nextProps.selectedLocation);
+
+      const lineText = editor.editor.getLine(line);
+      const spaces = lineText.match(/^\s+/);
+      if (spaces) {
+        column = Math.max(column, spaces[0].length);
+      }
```
jasonLaster: lets see if we can reuse the indentation helper:

this way we can write something like `column = Math.max(column, getIndentation(lineText));`

```diff
diff --git a/src/utils/indentation.js b/src/utils/indentation.js
index 8452f5e..579c6b6 100644
--- a/src/utils/indentation.js
+++ b/src/utils/indentation.js
@@ -2,7 +2,17 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
 
-function getIndentation(lines) {
+// @flow
+
+export function getIndentation(line?: string) {
+  if (!line) {
+    return 0;
+  }
+
+  return line.match(/^\s*/)[0].length;
+}
+
+function getFunctionIndentation(lines) {
   const firstLine = lines[0];
   const secondLine = lines[1];
   const lastLine = lines[lines.length - 1];
@@ -20,7 +30,7 @@ function getIndentation(lines) {
 
 export function correctIndentation(text) {
   const lines = text.trim().split("\n");
-  const indentation = getIndentation(lines);
+  const indentation = getFunctionIndentation(lines);
   const formattedLines = lines.map(_line =>
     _line.replace(new RegExp(`^\\s{0,${indentation - 1}}`), "")
   );
```

mmcote: Okay sounds good, I'll take a look now :)

mmcote:
@jasonLaster, just updated the pr with a new commit utilizing the indentation util :)

src/utils/indentation.js
```diff
@@ -2,17 +2,19 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
 
+export function getIndentationLength(line) {
+  return line && line.match(/^\s*/)[0].length;
```
jasonLaster: hmm, i think we should do:

```js
export function getIndentation(line) {
  if (!line) {
    return 0;
  }
  return line.match(/^\s*/)[0].length;
}
```

and we can rename the other private function

jasonLaster: also maybe a test for `getIndentation`

jasonLaster:
I see a bunch of mochitest issues, likely because this changes the behavior a bit. checkout our mochitest docs for help getting setup :)

src/components/Editor/tests/Editor.spec.js
```diff
@@ -30,10 +30,15 @@ function createMockEditor() {
       defaultTextHeight: () => 0,
       display: { gutters: { querySelector: jest.fn() } }
     },
+    editor: {
+      getLine: line => ""
+    },
```
jasonLaster: do we need this?

mmcote: No sorry, the merge ended up a little wonky, thanks for the double check, I'll remove it right now.

Issue #5312, opened 2018-02-04
[Sources] should indicate if they contain JS


lukaszsobek:
We are working towards pretty printing in-line JavaScript. One of the first steps to do that seems to be for the debugger to actually know the document contains in-line JavaScript. This could be done by modifying src/utils/source.js 

An outline would be: 
- add a `hasJavascript` function
- check if the source text has javascript with `indexOf` or maybe even `DOMParser`
- since `hasJavascript` might have a lot of data to deal with we might want to only call it together with a `isJavascript` that returns false.

lukaszsobek:
Better alternative in pr. Closing.

PR #5329, opened 2018-02-06
Introduces hasJavaScript()

lukaszsobek:
Associated Issue: #5312 

Introduces a function that checks if the source text contains inline javascript for future use of pretty printing.

Currently there is no check in place if the source text actually exists, so if we implement one there seem to be two scenarios:
1) we make the function return `undefined` to indicate no source text was provided as opposed to the regular `false` return for there not being a match.
2) we make the function return `false` since technically there was no match.

Also, the solution assumes

### DIFF
--- a/src/utils/source.js
+++ b/src/utils/source.js
@@ -59,6 +59,20 @@ export function shouldPrettyPrint(source: SourceRecord) {
 }
 
 /**
+ * Returns true if the specified source text contains inline javascript
+ *
+ * @return boolean
+ *         True if the source text likely contains javascript.
+ *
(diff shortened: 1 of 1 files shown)


jasonLaster:
hmm, i'm not sure how this would help. At the moment inline scripts are inside the HTML and we would need to extract that script from the document.

I'm currently looking at another bug w/ inline scripts where i think the fix is to have the server send us the inline script source. If we have that, it'll be much easier!

lukaszsobek:
More potential in the alternative solution. Closing.

Issue #4928, opened 2017-12-15
[Welcome] Hide content for narrow widths


jasonLaster:
When you open the debugger when it's docked to the right, the welcome content is wrapped like crazy.

Perhaps we want to hide the content after a certain min-width.

<img src='https://shipusercontent.com/3e74c5310afdf23f35618cbc3d89ee97/Screen%20Shot%202017-12-15%20at%2010.06.26%20AM.png' title='Screen Shot 2017-12-15 at 10.06.26 AM.png' width=589>

CC @violasong?

pradeepgangwar:
@jasonLaster May I try fixing this one if available. :)

violasong:
Oof, yeah, "fix vertical mode" is a general todo item on my list since it suffers from not enough dogfooding - lots of issues including can't get narrow enough, imo. 

Interesting to see it looks like this when you switch from bottom-dock to side. 

<img width="462" alt="image" src="https://user-images.githubusercontent.com/55994/34281344-21644466-e68c-11e7-8cf3-446796240f48.png">

The above actually looks fine if the breakpoints part is just a lot shorter, and the source tree is narrower as well, perhaps something like this: 

<img width="463" alt="image" src="https://user-images.githubusercontent.com/55994/34281424-9fa21420-e68c-11e7-9e3e-c42d2a9a78f3.png">

(as a bonus you can see the buggy sourcemap message which needs truncation). It doesn't seem to save my sidebar adjustments when I switch to bottom dock and back.

If we stick with a three-column idea, the content area is fine if we make the sidebars smaller by default. This still doesn't seem very usable however, so I think I'd prefer the solution above, "breakpoints below, but shorter, and narrow source tree."

<img width="462" alt="image" src="https://user-images.githubusercontent.com/55994/34281525-50419120-e68d-11e7-9034-e1239488fdc3.png">

pradeepgangwar:
I found that currently we use https://github.com/devtools-html/debugger.html/blob/master/src/components/WelcomeBox.css#L69 to maintain welcome box text, but recently i figured out that @media property is concerned about screen size but here width of welcome box also depends on sidebars widths. 
This can be handled by writing proper JS, so i want to ask where should i write such JS code as most JS files contain react code. If there is another workaround then please let me know.

jasonLaster:
Maybe the welcomeBox component? Not sure...

violasong:
[...]

PR #5330, opened 2018-02-06
correctly show vertical mode when toggling between modes

ericawright:
@jasonLaster 
Associated Issue: #4928

### Summary of Changes

* add a second match listener
* listen for either one to change and trigger prop changes only when one is correctly matched.

### Test Plan
to view the bug:
* open dev tools
* click on console
* toggle to the side
* flip to debugger
* observe that it is in horizontal layout
* if you click toggle while the debugger tab is displayed it does not do this

with this fix:
* It was tested on mozilla-central code, since with the web debugger you cannot toggle between vertical and horizontal modes.
* test by pasting this code into the appropriate spots in debugger.js and building firefox.

### DIFF
--- a/src/components/App.js
+++ b/src/components/App.js
@@ -21,7 +21,6 @@ import {
 } from "../selectors";
 
 import type { SourceRecord, OrientationType } from "../reducers/types";
-import { isVisible } from "../utils/ui";
 
 import { KeyShortcuts, Services } from "devtools-modules";
 const shortcuts = new KeyShortcuts({ window });
@@ -30,7 +29,10 @@ const { appinfo } = Services;
 
 const isMacOS = appinfo.OS === "Darwin";
 
-const verticalLayoutBreakpoint = window.matchMedia("(min-width: 800px)");
+const horizontalLayoutBreakpoint = window.matchMedia("(min-width: 800px)");
+const verticalLayoutBreakpoint = window.matchMedia(
+  "(min-width: 10px) and (max-width: 800px)"
+);
 
 import "./variables.css";
 import "./App.css";
@@ -105,6 +107,7 @@ class App extends Component<Props, State> {
   }
 
   componentDidMount() {
+    horizontalLayoutBreakpoint.addListener(this.onLayoutChange);
     verticalLayoutBreakpoint.addListener(this.onLayoutChange);
     this.setOrientation();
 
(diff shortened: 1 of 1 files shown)


src/components/App.js
```diff
@@ -30,7 +29,10 @@ const { appinfo } = Services;
 
 const isMacOS = appinfo.OS === "Darwin";
 
-const verticalLayoutBreakpoint = window.matchMedia("(min-width: 800px)");
+const horizontalLayoutBreakpoint = window.matchMedia("(min-width: 800px)");
```
lukaszsobek: So if we have 3 cases (invisible, vertical, horizontal) with the conditions being (< 10px; 10px - 800px; 800px+) do we  care what layout the debugger is if we cannot see it?

src/components/App.js
```diff
@@ -194,11 +198,13 @@ class App extends Component<Props, State> {
...
   setOrientation() {
-    const orientation = verticalLayoutBreakpoint.matches
-      ? "horizontal"
-      : "vertical";
-    if (isVisible()) {
-      this.props.setOrientation(orientation);
+    // If the orientation does not match (if it is not visible) it will
+    // not setOrientation, or if it is the same as before, calling
+    // setOrientation will not cause a rerender.
+    if (horizontalLayoutBreakpoint.matches) {
```
lukaszsobek: Same as above - does it seem feasible to just address the horizontal layout and treat everything else as vertical?

ericawright: We could, but, that would cause a re-render every time we change away from the debugger tab, and another re-render when we change back.

src/components/App.js
ericawright: That was the intention with this. if it matches positively with either case, then change to that case, if it is below 10px, leave it the way it previously was.

jasonLaster:
hmm, i might be a bit dense, but how does this help hiding the welcome content when it's narrow?

ericawright:
@jasonLaster now, the window remaining in "horizontal mode" when it is too thin doesn't happen. the content displaying like that was a bug. It should have been displaying like this all along:
screen shot 2018-02-07 at 1 14 06 pm
The three panel layout is only for horizontal mode/when it is wide enough. The four panel layout pictured is vertical mode, previously it was not correctly switching to vertical mode.

ericawright:
feb-07-2018 13-27-23
I'm not sure if this gif will work, but maybe it will help illustrate the problem

jasonLaster:
oh nice. looks great

Issue #5278, opened 2018-02-01
[ProjectSearch] "Find in files" show "No results found" when search is running


nchevobbe:
Steps to reproduce:
1. Run the debugger launchpad
2. Open the debugger on the tab
3. Search in file for "reveal"

Expected results:

A label/icon is shown to indicate the search is running

Actual results:

There's a "No results found" label displayed, even if the search is not finished yet

![feb-01-2018 08-43-41](https://user-images.githubusercontent.com/578107/35667008-5e77c3da-072c-11e8-9124-9a724059366a.gif)

You can even see the search counter being incremented at the right of the input, but no results are shown in the main panel.

PR #5336, opened 2018-02-07
Bring back loading message to project text search

sravan-s:
Associated Issue: #5278

### Summary of Changes

* Show loading message while results are fetched
* Add test for loading

### Test Plan

- Open some large website 
- Do text search
- See if `Loading` is displayed while search is going on
Follow the gif please 

### Screenshots/Videos (OPTIONAL)]
Gif

### DIFF
--- a/src/components/ProjectSearch.js
+++ b/src/components/ProjectSearch.js
@@ -247,6 +247,9 @@ export class ProjectSearch extends Component<Props, State> {
   renderResults = () => {
     const results = this.getResults();
     const { status } = this.props;
+    if (!this.props.query) {
+      return;
+    }
     if (results.length && status === statusType.done) {
       return (
         <ManagedTree
@@ -261,20 +264,11 @@ export class ProjectSearch extends Component<Props, State> {
         />
       );
     }
-    if (status === statusType.fetching) {
-      return (
-        <div className="no-result-msg absolute-center">
-          {L10N.getStr("loadingText")}
-        </div>
-      );
-    }
-    if (this.props.query && !results.length) {
-      return (
-        <div className="no-result-msg absolute-center">
-          {L10N.getStr("projectTextSearch.noResults")}
-        </div>
-      );
-    }
+    const msg =
+      status === statusType.fetching
+        ? L10N.getStr("loadingText")
+        : L10N.getStr("projectTextSearch.noResults");
+    return <div className="no-result-msg absolute-center">{msg}</div>;
   };
 
   renderSummary = () => {
--- a/src/components/ProjectSearch.js
+++ b/src/components/ProjectSearch.js
@@ -261,10 +261,14 @@ export class ProjectSearch extends Component<Props, State> {
         />
       );
     }
-    if (
-      status === statusType.fetching ||
-      (this.props.query && !results.length)
-    ) {
+    if (status === statusType.fetching) {
+      return (
+        <div className="no-result-msg absolute-center">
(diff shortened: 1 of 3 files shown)


src/components/ProjectSearch.js
```diff
@@ -261,10 +261,14 @@ export class ProjectSearch extends Component<Props, State> {
         />
       );
     }
-    if (
-      status === statusType.fetching ||
-      (this.props.query && !results.length)
-    ) {
+    if (status === statusType.fetching) {
+      return (
+        <div className="no-result-msg absolute-center">
```
lukaszsobek: It seems we have a bit of duplicated html here and might want to refactor the code to shave off some lines.

src/components/tests/ProjectSearch.spec.js
```diff
@@ -84,6 +85,14 @@ describe("ProjectSearch", () => {
     expect(component).toMatchSnapshot();
   });
 
+  it("should display loading message while search is in progress", () => {
```
lukaszsobek: Yummy 👍

src/components/ProjectSearch.js
sravan-s: Aaha let me see what I can shave off 😉

codehag: has this already been fixed? *looks for duplicated html

sravan-s: @codehag @lukaszsobek 
sorry, no I was super busy last few days. I could take a look only today... :( 
Do you have any specific suggestions?

codehag: we can probably simplify the code, as it looks like we are accounting for every case. we can add a 
```javascript
if (!this.props.query) return
```
but its probably not necessary.

for the rest we can do:

```javascript
const msg = status === statusType.fetching 
  ? L10N.getStr("loadingText") 
  : L10N.getStr("projectTextSearch.noResults");

return (
  <div className="no-result-msg absolute-center">
    {msg}
  </div>
```

darkwing:
Hello @sravan-s; I've tried rebasing this for you, I push the rebase to your branch, but unfortunately the PR isn't updating, which is a bit crazy.

Do you feel comfortable managing this merge conflict and rebasing?

sravan-s:
I updated some tests, can anyone verify?

src/components/tests/ProjectSearch.spec.js
```diff
@@ -204,15 +213,15 @@ describe("ProjectSearch", () => {
...
+        status: statusType.fetching,
         results: testResults
       });
       expect(component).toMatchSnapshot();
     });
 
     it("false if not done & no results", () => {
       const component = render({
-        status: "searching"
+        status: statusType.fetching
```
sravan-s: I guess this is the correct status

src/components/tests/__snapshots__/ProjectSearch.spec.js.snap
```diff
@@ -731,7 +766,7 @@ exports[`ProjectSearch showErrorEmoji false if not done & no results 1`] = `
     <div
       className="no-result-msg absolute-center"
     >
-      No results found
+      Loading…
```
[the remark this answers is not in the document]
codehag: this is getting the text from L10N, which is translated according to locale. You can see the expected translation here: https://github.com/devtools-html/debugger.html/blob/master/assets/panel/debugger.properties#L612

so, this looks like it is what it should be. :)

PR #5332, opened 2018-02-06
WIP: Fix of dbg-sources test

darkwing:
This is a work in progress.  Please ignore.

PR #5337, opened 2018-02-07
fix failing mochitest dbg-sources

codehag:
fixed mochitest that was failing due to lack of focus in source tree

issue was we had unified the two paths and now the expected functionality was broken


wldcordeiro:
Looks like you and @darkwing had both attempted to fix this test  -> #5332

darkwing:
We merged @jasonLaster's PR (https://github.com/devtools-html/debugger.html/pull/5357) so let's close this for now.

Issue #4985, opened 2017-12-27
[Selectors] Move selectors to the selectors folder


lukaszsobek:
Currently some of the reselect selectors reside in various reducers, whereas other selectors already reside in the src/selectors folder. It would be nice if all the selectors were moved to the already existing src/selectors folder so we can:
1.) shorten the length of reducers files
2.) have code in a designated place 
3.) separate the functionality of reducers from selectors
4.) make import statements clearer
5.) reduce tracing of components

jasonLaster:
@codehag and i have discussed this and think it would help.

It would nice if many of the selectors were in the selectors directory and were exported via `selectors/index`

jasonLaster:
at the moment the selectors in the `src/selectors` dir pull state from different reducers, which is why they cant be in one reducers dir. I'm not sure how i feel about moving the selectors in the reducer files into their own module... 

The major reason not to do it is that currently the selectors are easier to follow because they're next to the reducer and the state type.

I'm not necessarily opposed either because moving the selectors out will make space for the reducer update helper methods...

@codehag

lukaszsobek:
Ok, I'll keep working on it. Then we can take a look at the finished thing. If it looks good we can use it, if not we don't. I like having the selector in a separate file as it makes the reducer much cleaner + there is only one import.

codehag:
do we have a candidate reducer for this? something we can play around with and see how it looks?

lukaszsobek:
Rescoping in another issue. Closing this one.

PR #5344, opened 2018-02-07
Moves selectors to selector folder

lukaszsobek:
Associated Issue: #4985

- [x] Move items from selectors.js to `index.js` in `src/selectors/` folder


src/components/SecondaryPanes/Breakpoints.js
```diff
@@ -8,23 +8,16 @@ import { bindActionCreators } from "redux";
...
-  getSources,
-  getSourceInSources,
-  getBreakpoints,
-  getPauseReason,
-  getTopFrame
-} from "../../selectors";
-import { isInterrupted } from "../../utils/pause";
-import { makeLocationId } from "../../utils/breakpoint";
+
+import { getBreakpointSelector } from "../../selectors/";
```
jasonLaster: I think the selector should make it clear what it's used for `getBreakpointsForPane` because it's different if we want them for the editor

lukaszsobek: Thanks, the first version was `whatchamacallit` as i was not really sure what to name it :)

src/selectors/getBreakpoints.js
```diff
@@ -0,0 +1,52 @@
...
+
+function isCurrentlyPausedAtBreakpoint(frame, why, breakpoint) {
+  if (!frame || !isInterrupted(why)) {
+    return false;
+  }
+
+  const bpId = makeLocationId(breakpoint.location);
+  const pausedId = makeLocationId(frame.location);
+  return bpId === pausedId;
+}
```
bomsy: Is this file going to contain all the selectors for breakpoints or is it one file per selector? if it is the former , then this file may probably be  "src/selectors/breakpoints" rather than "src/selectors/getBreakpoints"?

jasonLaster: yeah - getBreakpointsForPane

codehag:
As per our discussion, with components I think this makes a lot of sense! One place where I find myself a little frustrated with our codebase is that when we add new selectors, we often have to thread them through the selectors file from the reducer. Is this also what you are feeling or are you seeing another issue?

jasonLaster:
> we often have to thread them through the selectors file from the reducer

what do you mean?

jasonLaster:
@lukaszsobek what are the next steps here?

lukaszsobek:
@jasonLaster The idea is to  separate the selectors from the reducers and clean up src/selectors.js so that the selectors are neatly contained within the selector folder instead of being slightly in many places.

src/selectors/getBreakpoints.js
jasonLaster: hmm, i think i misread this comment... 

I thought you were referring to the selector function for the breakpoints component. Not sure about the reducers.

jasonLaster:
This seems to be growing fairly quickly. I'm not sure if we want to take all of the selectors out of the reducers, either way we should do this slowly and discuss. I had questions about the initial PR, now i've got lots more questions

lukaszsobek:
@jasonLaster Sure, now that you have seen more or less the direction it is going there seem to be two questions:
1.) Is this issue necessary?
2.) If yes, what should be the scope for now?

I think that at the very basic level we should move the contents of `selector.js` to `src/selectors/index.js` so it conforms with the pattern we are using in the rest of debugger. Other than that I'd also move the selectors that can be moved without too much impact to start separating concerns, but that feels like a question of preference and we might not want to do that since selectors are easier to get at in reducers.

jasonLaster:
Okay, 

- Lets make this PR just the `src/selectors/index.js` migration
- Lets make a new PR for components Breakpoints selector
- Lets talk about what to do about the selectors currently in reducer files...

I think a structure like this could make sense:

- selectors
  - breakpoints
    - index.js
    - getBreakpointsForPane
    - getVisibleBreakpoints
    
Here all the breakpoints selectors are defined or exported in `selectors/breakpoints/index.js`. And `reducers/breakpoints` is just for updating the state.

lukaszsobek:
Sounds great. On it.

PR #5345, opened 2018-02-07
Fix #5288 - Save directory root preference upon refresh; remove throttling

### LINKED ISSUE #5288: [SourceTree] The directory root should be saved
We should save the root in prefs so that once the user sets it, it will be saved when they close the debugger and re-open it.

darkwing:
Associated Issue: #5288 

### Summary of Changes

* Remove unnecessary throttling of source tree
* Sets default `projectDirectoryRoot` to preference value.
* You can now set a directory root, refresh, and see the source tree default to that root

### Test Plan

1.  Set a root directory
2.  Refresh the page
3.  Ensure that root directory is still the root of the source tree
4.  Unset root directory
5.  All resources should display

Example test plan:

Feel free to improve it!

### DIFF
--- a/src/components/PrimaryPanes/SourcesTree.js
+++ b/src/components/PrimaryPanes/SourcesTree.js
@@ -43,7 +43,6 @@ import {
 } from "../../utils/sources-tree";
 
 import { copyToTheClipboard } from "../../utils/clipboard";
-import { throttle } from "../../utils/utils";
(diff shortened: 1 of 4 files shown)


darkwing:
Pegged another commit on this @jasonLaster to remove a throttle util function we're no longer using.  There's one other `throttle` function usage in the project but we're using lodash's for that.

PR #5347, opened 2018-02-07
The (index) is not a directory

darkwing:
At present we can right-click the `(index)` item in the source tree and set as root directory -- that isn't ideal.  It was happening because `isDirectory` was returning `true` for `(index)`, so we should change that.


jasonLaster:
Seeing an issue here:

jasonLaster:
failing for other mochitests

PR #5349, opened 2018-02-07
[WIP] Fix 5341 - Add 'copy text' menu item to source tree context menu

darkwing:
This is more of a conversation starter.  

## ToDo
- [ ] Do we want this functionality?
- [ ] If so, do we also want this as a tab context menu item? (probably)
- [ ] What would be a good access key?
- [ ] Add a test
- [ ] Figure out the case where "undefined" is returned

### DIFF
--- a/assets/panel/debugger.properties
+++ b/assets/panel/debugger.properties
@@ -24,6 +24,11 @@ copySource.accesskey=y
 copySourceUri2=Copy source URI
 copySourceUri2.accesskey=u
 
+# LOCALIZATION NOTE (copySourceText): This is the text that appears in the
+# context menu to copy the source text of file open.
+copySourceText=Copy source code
+copySourceText.accesskey=s
+
 # LOCALIZATION NOTE (setDirectoryRoot.label): This is the text that appears in the
 # context menu to set a directory as root directory
 setDirectoryRoot.label=Set directory root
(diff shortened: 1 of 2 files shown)


assets/panel/debugger.properties
```diff
@@ -24,6 +24,11 @@ copySource.accesskey=y
 copySourceUri2=Copy source URI
 copySourceUri2.accesskey=u
 
+# LOCALIZATION NOTE (copySourceText): This is the text that appears in the
+# context menu to copy the source text of file open.
```
flodolo: nit: of the open file

assets/panel/debugger.properties
```diff
@@ -24,6 +24,11 @@ copySource.accesskey=y
...
+copySourceText=Copy source code
```
flodolo: Please use `copySourceText.label` for this string ID (also in the comment above). That's the best way to name strings, unfortunately a lot of the old ones use "foo" and "foo.accesskey".

bomsy:
@darkwing How would the clipboard fair with handling really massive files with a lot of lines of code?

codehag:
Did i understand correctly that this copies the text of the file name? rather than the file? can you describe a bit what you are thinking with this feature and where you see it as being useful? Maybe with a user story! that would help with evaluating it.

darkwing:
@bomsy Great question -- I'll test a huge file today.

@codehag This would copy the source's contents to the clipboard.

I'm not super insistent on this feature -- I just wanted to start the conversation and experiment.

darkwing:
Closing this for now.  While I think it would be useful, there are considerations, like massive file size, that aren't a priority right now.

PR #5357, opened 2018-02-08
Fix the sources tree test

jasonLaster:
### Summary of Changes

Taking another whack at the test, by cleaning up the sources tree code a bit more

### DIFF
--- a/src/components/PrimaryPanes/SourcesTree.js
+++ b/src/components/PrimaryPanes/SourcesTree.js
@@ -314,9 +314,11 @@ class SourcesTree extends Component<Props, State> {
         key={item.path}
         onClick={e => {
           e.stopPropagation();
+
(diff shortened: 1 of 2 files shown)


src/components/PrimaryPanes/SourcesTree.js
```diff
@@ -301,6 +301,10 @@ class SourcesTree extends Component<Props, State> {
         className={classnames("arrow", {
           expanded: expanded
         })}
+        onClick={e => {
```
darkwing: Since we're allowing click on the entire tree node, I think we can remove this all together -- the icon at this point is nothing but a UI hint.

jasonLaster: this is what fixed it... :)

Issue #5324, opened 2018-02-06
"No result found" label breaks on letter while there's plenty of space


nchevobbe:
Steps to reproduce:
1. Go to data:text/html,<script>var hello;</script>
2. Open the debugger
3. Search in files ( Cmd + Shift + F ) for "boo"

Expected results:

The "No result found" label is displayed, and word wraps to a new line when needed

Actual results:

![image](https://user-images.githubusercontent.com/578107/35863811-208a80e4-0b50-11e8-8aab-23ae244cca1a.png)

The words break on letters while there's plenty of space to break on word. At worst, I expect hyphens at the right spot.

ericawright:
fixed with #5359

PR #5359, opened 2018-02-08
Issue 5324

tripurari001:
Associated Issue: #5324

### Summary of Changes

* removed the word-break property from `no-result-msg` css class in `ProjectSearch.css` file
* added `overflow-wrap` to break word if it exceed the parent (`max-width: 100%`)
* `hyphens: auto` to insert hyphens when word break on letters

#### before 
before

#### after
after

### DIFF
--- a/src/components/ProjectSearch.css
+++ b/src/components/ProjectSearch.css
@@ -66,7 +66,6 @@
   color: var(--theme-body-color-inactive);
   font-size: 24px;
   padding: 4px;
-  word-break: break-all;
 }
 
 .project-text-search .file-result {
--- a/src/components/ProjectSearch.css
+++ b/src/components/ProjectSearch.css
@@ -65,7 +65,10 @@
 .project-text-search .no-result-msg {
   color: var(--theme-body-color-inactive);
   font-size: 24px;
-  padding: 4px;
+  padding: 4px 15px;
+  max-width: 100%;
+  overflow-wrap: break-word;
+  hyphens: auto;
 }
 
 .project-text-search .file-result {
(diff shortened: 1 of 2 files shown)


jasonLaster:
CC @ericawright

ericawright:
This works well for English, but once we translate, it's not guaranteed that the words will fit.
Then this happens:
screen shot 2018-02-08 at 2 07 15 pm

I think it would be nice if the words would break on letters if they have to.

tripurari001:
cc @ericawright  Thanks for the suggestion

Added the `overflow-wrap: break-word` to break word if it exceed parent width (` max-width: 100% `)
also added `left-right padding` to `15px` and `hyphen:auto` (not working in firefox 59.0b7 and working in chrome)

after change  

final

ericawright:
@tripurari001 thanks, it looks good! 
I don't know why `hyphens` is not working in this case, it should work on Firefox. I tried on another page and `hyphens` works as expected there, so there's something unique about this. Removing the `overflow-wrap` rule didn't seem to make it work either. I tried fiddling with some sizing rule sand couldn't figure it out. I think this is good for now, and maybe we can open a new bug about the hyphens not working.
Do you know how to rebase and squash these commits so there's only 1? Otherwise I can do it for you.

tripurari001:
@ericawright  Thanks for your help. 
This was my first open-source code contribution and I am Looking forward to do more .
One thing I noticed while fixing this issue is `0 result x` at top right breaks in two line at smaller width ( as you can see in 3rd GIF) 
not sure if it's expected or not

AnshulMalik:
Yep @tripurari001. Thanks for the contribution. It'll be an awesome journey

tripurari001:
@ericawright  Notice this in 3rd gif 
screen shot 2018-02-09 at 10 04 46 pm

jasonLaster:
hmm - we should fix this

PR #4932, opened 2017-12-16
Add go to line shortcut in shortcuts modal

pradeepgangwar:
Associated Issue: #4804 

### Summary of Changes

* Add go to line shortcut `CtrlorCmd+:` in shortcut modal.

### Screenshots

screenshot-2017-12-16 debugger 2

PR #5360, opened 2018-02-08
Fix localizations

jasonLaster:
### Summary of Changes

Fixes issues found in #4932

### DIFF
--- a/assets/panel/debugger.properties
+++ b/assets/panel/debugger.properties
@@ -567,16 +567,16 @@ watchExpressions.refreshButton=Refresh
 # LOCALIZATION NOTE (welcome.search): The center pane welcome panel's
 # search prompt. e.g. cmd+p to search for files. On windows, it's ctrl, on
 # a mac we use the unicode character.
-welcome.search=%S To search for sources
+welcome.search=%S to search for sources
 
 # LOCALIZATION NOTE (welcome.findInFiles): The center pane welcome panel's
 # search prompt. e.g. cmd+f to search for files. On windows, it's ctrl+shift+f, on
 # a mac we use the unicode character.
-welcome.findInFiles=%S To find in files
+welcome.findInFiles=%S to find in files
 
 # LOCALIZATION NOTE (welcome.searchFunction): Label displayed in the welcome
 # panel. %S is replaced by the keyboard shortcut to search for functions.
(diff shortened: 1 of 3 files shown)


assets/panel/debugger.properties
```diff
@@ -707,7 +707,7 @@ functionSearchSeparatorLabel=←
 # Do not localize "CmdOrCtrl+;", or change the format of the string. These are
 # key identifiers, not messages displayed to the user.
 gotoLineModal.placeholder=Go to line…
-gotoLineModal.key=CmdOrCtrl+;
+gotoLineModal.key2=CmdOrCtrl+;
```
flodolo: Can we split the part of comment about the shortcut into a separate comment? Right now it's on the placeholder string's comment.

jasonLaster:
fixed. @flodolo  what do you think?

juliandescottes:
I don't think the comment from @flodolo was actually addressed here? The idea was to move the comment right above `gotoLineModal.key2=CmdOrCtrl+;` if I read that correctly.

Also this is not included in release 15. @jasonLaster we should fix this asap, can you cherry pick this one in the release?

flodolo:
> Also this is not included in release 15. @jasonLaster we should fix this asap, can you cherry pick this one in the release?

I'm confused, I'm sure I've seen a version of this with the comment split (there's about an hour between my comment and the second one).

flodolo:
> I'm confused, I'm sure I've seen a version of this with the comment split (there's about an hour between my comment and the second one).

OK, I'm not crazy. Here was [the commit](https://github.com/devtools-html/debugger.html/commit/f9209bc) I've approved, resurrected from my emails. No clue what happened in-between.

juliandescottes:
I would guess that a `-force` push erased the commit you reviewed and it was the one that got merged into master?

flodolo:
It's still listed above the merge. 

screenshot-2018-2-9 fix localizations by jasonlaster pull request 5360 devtools-html debugger html

As far as I've seen, a force push would hide it from the PR, while the commit itself would be still reachable (that's why I went straight to my email).

jasonLaster:
got it. I'll push a fix.

re: release 15, this commit landed after that was cut. I'll get 16 out soon

PR #5369, opened 2018-02-09
babel 7 :star2:

jasonLaster:
### Summary of Changes

upgrades babel to 7, which unlocks the faster traverse function


codehag:
Looks like there are some unexpected failure here! need any help on this?

PR #5371, opened 2018-02-09
[QuickOpen] show indicator if symbols are still being parsed

atwalg2:
Associated Issue: #4488

### Summary of Changes

* Added loading text, and styled
* Logic
* Added snapshot

### Test Plan
- [x]  Added new snapshot

To test load a project into the debugger, then open up quick open using ctrl + o, and you should see the loading text appear, then disappear on it's own when the symbols have been parsed. 

screencap


jasonLaster:
We don't have any spinners yet, so lets just re-use the "Loading..." message we have in some other places

atwalg2:
The spinner is just a temporary filler proof of concept. Where is the loading message used elsewhere?

jasonLaster:
https://github.com/devtools-html/debugger.html/blob/release-15/src/components/SecondaryPanes/Scopes.js#L90-L91

atwalg2:
Thanks. I set it up the right way I think now (Visually). I mainly have to configure the logic to work exactly how it should now, it's a bit wrong right now.

codehag:
The travis test is failing at the moment regarding the quick-open... can you take a look? do you have experience running mochi tests? Happy to help if needed!

atwalg2:
@codehag no I haven't got any experience yet using mochi tests

jasonLaster:
looking better. How about we center the loading message and add some vertical padding

jasonLaster:
@atwalg2 please share a screenshot

atwalg2:
image

src/components/QuickOpenModal.css
```diff
@@ -11,3 +11,8 @@
 .theme-dark .result-item .subtitle .highlight {
   color: white;
 }
+
+.loadIndicator {
```
AnshulMalik: How about `loading-indicator`? CSS rules generally are not camel case

src/components/QuickOpenModal.js
```diff
@@ -358,11 +370,12 @@ export class QuickOpenModal extends Component<Props, State> {
 
   render() {
     const { enabled, query } = this.props;
-    const { selectedIndex, results } = this.state;
+    let { selectedIndex, results, isLoading } = this.state;
```
AnshulMalik: `const` -> `let` am I missing something here?

codehag: agreeing with @AnshulMalik -- since none of these variables are reassigned (and since they are part of state and should only be updated via setState), we can use const here!

src/components/QuickOpenModal.js
```diff
@@ -162,21 +167,26 @@ export class QuickOpenModal extends Component<Props, State> {
...
   updateResults = (query: string) => {
     if (this.isGotoQuery()) {
-      return;
+      return this.setState({ isLoading: false });
     }
 
     if (query == "") {
-      return this.showTopSources();
+      this.showTopSources();
+      return this.setState({ isLoading: false });
```
codehag: `this.setState({ isLoading: false });` can be moved  to the top of this function (since it always happens) so that we don't have it repeating in each if statement

src/components/QuickOpenModal.css
atwalg2: I'll fix this right away

src/components/QuickOpenModal.js
atwalg2: I'll fix this right away

src/components/QuickOpenModal.js
atwalg2: I'm glad you said this because it made me re examine my code, and realized I was causing wasteful re renders, I moved the isLoading: false to the instances when results is already being set through setState

src/components/QuickOpenModal.js
```diff
@@ -134,35 +135,39 @@ export class QuickOpenModal extends Component<Props, State> {
...
     } else {
-      this.setState({ results: sources.slice(0, 100) });
+      this.setState({ results: sources.slice(0, 100), isLoading: false });
     }
   };
 
   updateResults = (query: string) => {
     if (this.isGotoQuery()) {
-      return;
+      return this.setState({ isLoading: false });
```
jasonLaster: hmm, re isLoading... i dont think we need it. Let me explain

the only time we're really loading is when `symbols` are not fetched. We can check this.props.symbols and if it doesn't have any functions it is still loading otherwise we're not

onChange and updateResults dont matter because it's instant

atwalg2:
Travis Ci Error
image

from test/mochitest
image

jasonLaster:
hmm... i think that's happening for others

jasonLaster:
do you need a rebase

src/components/QuickOpenModal.js
```diff
@@ -390,10 +388,15 @@ export class QuickOpenModal extends Component<Props, State> {
           onChange={this.onChange}
           onKeyDown={this.onKeyDown}
           handleClose={this.closeModal}
-          hasPrefix={this.hasPrefix()}
           expanded={expanded}
           selectedItemId={expanded ? items[selectedIndex].id : ""}
         />
+        {!symbols ||
+          (symbols.functions.length == 0 && (
```
jasonLaster: sorry, this should be `symbols.identifiers` i think i told you the wrong thing

atwalg2: image

These are the only properties of symbols.

darkwing:
Hello @atwalg2 !  It looks to me like the `Loading text` never disappears:

foreverloading

src/components/QuickOpenModal.js
```diff
@@ -384,12 +382,17 @@ export class QuickOpenModal extends Component<Props, State> {
...
           onKeyDown={this.onKeyDown}
           handleClose={this.closeModal}
-          hasPrefix={this.hasPrefix()}
           expanded={expanded}
           selectedItemId={
             expanded && items[selectedIndex] ? items[selectedIndex].id : ""
           }
         />
+        {!symbols ||
+          (symbols.functions.length == 0 && (
```
darkwing: This `symbols.functions.length` will always fail if there's no query text, so we should add that to this conditional.

darkwing: Also, when I do add text to the input, and there are many results, `symbols.functions.length` is still 0, so maybe we should be doing a different check?

jasonLaster:
We'll be updating the loading logic in #5588 so it's not a big deal

PR #5374, opened 2018-02-09
Address localization comments

jasonLaster:
Follow up https://github.com/devtools-html/debugger.html/pull/5360


jasonLaster:
CC @flodolo @juliandescottes

PR #5382, opened 2018-02-09
Update to Babel 7

loganfsmyth:
Updated version of #5369 with passing tests and a few small fixes.


src/workers/parser/tests/__snapshots__/closest.spec.js.snap
```diff
@@ -78,7 +78,6 @@ FunctionDeclaration (1:0,3:1)
             right: Identifier (2:13,2:14)
               name: 'n'
     directives: []
-  expression: false
```
loganfsmyth: This field was deleted from Babel 7's AST because it is just duplicate data. If you want to know if the body of an arrow function is an expression, people are expected to check `node.body.type !== "BlockStatement"`

src/workers/parser/utils/ast.js
```diff
@@ -14,28 +14,28 @@ import { getSource } from "../sources";
 let ASTs = new Map();
 
 function _parse(code, opts) {
-  return babylon.parse(code, opts);
+  return babylon.parse(code, {
+    ...opts,
+    tokens: true
```
loganfsmyth: This `_parse` function is also called from `parse-script-tags` so it needs `tokens: true` too. This was causing the mochitest failures. Even worse, that module assumes tokens will always be present, so instead of resulting in a `tokens` result of `undefined`, it returns `[undefined]` as an array with a single item.

src/workers/parser/utils/ast.js
```diff
@@ -14,28 +14,28 @@ import { getSource } from "../sources";
...
-    tokens: true,
     plugins: [
       "jsx",
       "flow",
       "doExpressions",
       "objectRestSpread",
       "classProperties",
-      "exportExtensions",
+      "exportDefaultFrom",
+      "exportNamespaceFrom",
```
loganfsmyth: This plugin got split in two in 7.x.

src/workers/parser/utils/ast.js
```diff
@@ -14,28 +14,28 @@ import { getSource } from "../sources";
...
       "asyncGenerators",
       "functionBind",
       "functionSent",
-      "dynamicImport",
-      "templateInvalidEscapes"
```
loganfsmyth: This has landed in the spec, so it is on automatically.

src/workers/parser/utils/ast.js
loganfsmyth: @JLast if `parse-script-tags` is actually something used anywhere else, you probably want to have it pass this option automatically, or update it to handle the case where the user didn't pass in `tokens: true`.

src/workers/parser/utils/ast.js
wldcordeiro: There are new plugins in 7.x that I think we could enable like the TypeScript one

loganfsmyth: Can't enable both `flow` and `typescript` at the same time, so there'd need to be some logic to check the original file extensions, but should otherwise be workable.

Issue #5342, opened 2018-02-07
Panel size changes


ericawright:
Some changes that were missed with #4928
![image](https://user-images.githubusercontent.com/10803178/35937561-84b8ee7e-0c14-11e8-8327-b02b68f6bf5d.png)

ericawright:
1 - this is no longer relevant since it switches to vertical-mode styling
2 - done
3 - done
4 - not possible without an overhaul of dev-tools core since we need to listen for the click event when toggling between vertical and horizontal. Currently, the size measurement is the only thing we have to detect which mode we are in.

PR #5383, opened 2018-02-09
reduce to one layout type and rearrange sizes of panels

ericawright:
Associated Issue: #5342 
### Summary of Changes

* combine renderHorizontalLayout and renderVerticalLayout methods into one.
* change the sizes of the panels

in horizontal mode: 
* primary panel is 250px wide 
* secondary panel is 300px wide

in vertical mode:
* primary panel is 150px wide
* secondary panel is 400px tall

This is in conjunction with a change necessary to devtools-splitter component.
PR: https://github.com/devtools-html/devtools-core/pull/930

### Screenshots
screen shot 2018-02-09 at 3 18 14 pm
screen shot 2018-02-09 at 3 45 04 pm

### DIFF
--- a/src/components/App.js
+++ b/src/components/App.js
@@ -80,7 +80,7 @@ class App extends Component<Props, State> {
   onLayoutChange: Function;
   getChildContext: Function;
   renderEditorPane: Function;
-  renderVerticalLayout: Function;
+  renderLayout: Function;
   toggleQuickOpenModal: Function;
   onEscape: Function;
   onCommandSlash: Function;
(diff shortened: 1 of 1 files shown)


jasonLaster:
CC @darkwing - this might be a good PR for you

nchevobbe:
As far as I remember, switching to a Grid solved some issues that might not be the case here: 
- As Yulia said, it will allow us to remove a lot of complexity in the code with all the `isHorizontal` calls. 
- It will allow us to remove the resize listener to switch from Horizontal to Vertical, which, even not done frequently, is a nice when "snappy". At the moment we just re-render everything which is slow
- With the 2 previous points, I suspect a nice improvement in performances. At least, the layout is handled by the browser, and not by JS
- It will also simplify element alignment, i.e. no more fixed heigth to have the Tab srop and the breakpoint toolbar be the same height, and other cases I don't remember. Basically, a lot saner way of doing layout.

With all this being said, this change looks nice since it already removes complexity, but I do still want the debugger to use a CSS grid (very much). I did not had the time to look into the AreaResizer we talked about back in december since this work has been de-prioritized, but if we have someone willing to work on the switch to a css grid, i'll happily look into it.

jasonLaster:
Thanks @ericawright, looks awesome! I'll review soon.

re: CSS grid, imo that is a bigger change and should not be done as a bug fix.

darkwing:
Awesome!  I think at some point we should define a bunch of the width/numeric values at the top of this file, detailing what they mean and why, but shouldn't block merge.  Thank you @ericawright !

Issue #5150, opened 2018-01-19
[CallStack] Simplify Babel call stacks


jasonLaster:
It's really common for babel to create a bunch of call stack frames when you're paused in an async function.

Here is an example

| displayName | url |
|--|--|
| _callee2$ | webpack:///entry.js |
| tryCatch | webpack:///Users/jlaster/src/mozilla/debugger-examples/node_modules/regenerator-runtime/runtime.js |
| invoke | webpack:///Users/jlaster/src/mozilla/debugger-examples/node_modules/regenerator-runtime/runtime.js |
| defineIteratorMethods/</prototype[method] | webpack:///Users/jlaster/src/mozilla/debugger-examples/node_modules/regenerator-runtime/runtime.js |
| step | undefined |
| step/< | undefined |
| run | webpack:///Users/jlaster/src/mozilla/debugger-examples/node_modules/core-js/modules/es6.promise.js |
| notify/< | webpack:///Users/jlaster/src/mozilla/debugger-examples/node_modules/core-js/modules/es6.promise.js |
| flush | webpack:///Users/jlaster/src/mozilla/debugger-examples/node_modules/core-js/modules/_microtask.js |

It would be awesome if we could simplify it to 

| displayName | url |
|--|--|
| _callee2$ | webpack:///entry.js |
| Async | Babel |

STR:

1. go to http://firefox-dev.tools/debugger-examples/examples/babel/
2. add a breakpoint in entry.js#31

calebstdenis:
@jasonLaster Should we take into account the possibility of `regenerator-runtime` (the top frames) appearing in a non-Babel context? I'm leaning towards "yes" but if the answer is "no" then the implementation for this should be pretty trivial.

jasonLaster:
haha - yeah :) lets go for it!

PR #5385, opened 2018-02-09
WIP - Group async babel call stack frames

calebstdenis:
Associated Issue: #5150  

Work in progress:
- [x] Added some tests to cover existing grouping functionality.
- [x] Added a test covering the case outlined in #5150.
- [ ] Investigate what other cases might exist since it seems Babel has [many ways of dealing with async](https://babeljs.io/docs/plugins/syntax-async-functions)
- [ ] Make implementation pass tests (one is currently failing)

What do you guys think of the approach so far?


src/utils/frame.js
```diff
@@ -213,57 +213,60 @@ export function formatCopyName(frame: LocalFrame) {
...
-      currentGroup = [frame];
-    } else if (prevLibrary && prevLibrary == frame.library) {
-      currentGroup.push(frame);
-    } else {
-      items = addGroupToList(currentGroup, items);
-      currentGroup = [frame];
-    }
-
-    prevItem = frame;
+function collapseLastWebpackFrames(frames: Frame[]) {
```
AnshulMalik: What is this function doing?

src/utils/tests/frame.spec.js
```diff
@@ -118,4 +119,135 @@ describe("function names", () => {
...
+          id: "4",
+          displayName: "(global)",
+          source: {
+            url: "https://foobar.com/bundle.js"
+          }
+        }
+      ];
+      const collapsedFrames = collapseFrames(frames);
+
+      expect(collapsedFrames).toEqual([
```
codehag: you can use `toMatchSnapshot()` which will be easier to maintain in the future

src/utils/tests/frame.spec.js
```diff
@@ -118,4 +119,135 @@ describe("function names", () => {
       });
     });
   });
+
+  describe("collapseFrames", () => {
+    it("collapses React component render frames and Webpack bootstrap frames", () => {
+      const frames = [
```
codehag: what is the minimum information that we need for this to run? do we need id and display name for example? maybe we can either reduce this so that it only shows relevant info for the test, or we can create a factory that will generate frames for us when needed. something like https://github.com/devtools-html/debugger.html/blob/5026582ce0547f3a964b31cfd53aee1059db3cc6/src/actions/tests/expressions.spec.js#L109 but more general and that could replace other uses of this for tests

src/utils/tests/frame.spec.js
```diff
@@ -118,4 +119,135 @@ describe("function names", () => {
...
+        frames[0],
+        [frames[1], frames[2]],
+        [frames[3], frames[4]]
+      ]);
+    });
+  });
+
+  it("groups async frames generated by Babel", () => {
+    const frames = [
+      {
```
codehag: same here, lots of repeated info, which isnt necessarily crucial for the test. let me know if you need help with this!

darkwing:
Hey @calebstdenis !  Let us know if we can help push this to finish line!

src/utils/tests/frame.spec.js
calebstdenis: collapseFrames is a util function, though. The test isn't rendering any component. 

I actually started out with snapshot tests, then realized, "what's the point of having big snapshots when what I'm really testing is a basic JavaScript algorithm?"

src/utils/tests/frame.spec.js
calebstdenis: "library" might be the only one absolutely necessary. I'll look into that.

For the record, I was trying strike a balance between having minimal data and data that resembled what you would actually see in the wild. But for Jest tests I guess it does make sense to only worry about the former.

src/utils/frame.js
calebstdenis: This is pre-existing functionality that I've put into a separate function.  Sometimes ([here](https://github.com/devtools-html/debugger.html/pull/5385/files#diff-7a84ca451c1af01a43cf706fa6736146R159), for example) you want the bottom frame to be considered a "webpack" frame even if the sourcemap URL doesn't identify it as such. @jasonLaster Did I get that right?

Issue #5377, opened 2018-02-09
[Preview] implement fast preview


jasonLaster:
We currently have a really smart preview that handles evaluating a bunch of interesting expressions. The downside, is that it requires parsing and traversing the AST to get the data. This makes preview powerful, but potentially slow when large sources are loaded.

Chrome on the other hand, only supports limited expressions, but because of this is always fast.

I think we can have the best of both worlds, we can use our existing preview when symbols are available and we can fall back on a fast/simple preview when it is not available. 

Here are two examples that we can support, but chrome can not. 
1. expressions with arrays or object properties
2. multi-line chained expressions

```js

  const firstAuthor = collection.books[1].author;
  const firstSciFiAuthor = collection.genres["sci-fi"].books[0].author;

  const directors = collection
    .movies
    .map(movie => movie.director)
```

### Fast Preview

The implementation is borrowed from [chrome]
```js
function tokenAtTextPosition(cm, lineNumber, columnNumber) {
  if (lineNumber < 0 || lineNumber >= cm.lineCount()) {
    return null;
  }

  var token = cm.getTokenAt(lineNumber, (columnNumber || 0) + 1));
  if (!token) {
    return null;
  }
  
  return {startColumn: token.start, endColumn: token.end, type: token.type};
}

const token = tokenAtTextPosition(line, column);
const startHighlight = token.startColumn;
const endHighlight = token.endColumn - 1;

while (startHighlight > 1 && line.charAt(startHighlight - 1) === '.') {
    var tokenBefore = tokenAtTextPosition(lineNumber, startHighlight - 2);

    if (!tokenBefore || !tokenBefore.type) {
      return null;
    }
    
    startHighlight = tokenBefore.startColumn;
}
```

### Initial Patch

https://gist.github.com/43a8a4dd965540b33a6508b9ffced13b

[chrome]:https://github.com/ChromeDevTools/devtools-frontend/blob/master/front_end/sources/JavaScriptSourceFrame.js#L414

jasonLaster:
yep! that's what i'm thinking.

jasonLaster:
here's what i got so far - https://github.com/devtools-html/debugger.html/pull/5386/files#diff-bb4a5ccb760e4629be966f69f285d154R107

PR #5386, opened 2018-02-10
[WIP] Fast Preview

jasonLaster:
Associated Issue: #5377 

### Summary of Changes

* Early hacky version. Give it a go and let me what you think.
* It somewhat works... 
  - finds lots of expression
  - highlighting is off
  - is flickery

### DIFF
--- a/src/actions/preview.js
+++ b/src/actions/preview.js
@@ -9,10 +9,11 @@ import { getTokenLocation } from "../utils/editor";
 import { isReactComponent, isImmutable } from "../utils/preview";
 import { isGeneratedId } from "devtools-source-map";
 import { PROMISE } from "./utils/middleware/promise";
+import { getExpressionFromCoords } from "../utils/editor/get-expression";
 
 import {
   getPreview,
-  getInScopeLines,
+  isLineInScope,
   getSelectedSource,
   getSelectedFrame,
   getSymbols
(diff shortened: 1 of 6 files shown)


src/actions/preview.js
```diff
@@ -61,16 +62,43 @@ async function getExtraProps(expression, result, evaluate) {
...
+function isInvalidTarget(target: HTMLElement) {
+  if (!target || !target.innerText) {
+    return true;
+  }
+
+  const tokenText = target.innerText.trim();
+  const cursorPos = target.getBoundingClientRect();
+
+  // exclude literal tokens where it does not make sense to show a preview
+  const invaildType = ["cm-string", "cm-number", "cm-atom"].includes(
```
codehag: "invalid"

src/actions/preview.js
```diff
@@ -61,16 +62,43 @@ async function getExtraProps(expression, result, evaluate) {
...
+    const tokenPos = getTokenLocation(editor.codeMirror, target);
     const cursorPos = target.getBoundingClientRect();
     const preview = getPreview(getState());
 
     if (preview) {
-      // We are mousing over the same token as before
-      if (isEqual(preview.tokenPos, location)) {
+      // Return early if we are currently showing another preview or
+      // if we are mousing over the same token as before
+      if (preview.updating || isEqual(preview.tokenPos, tokenPos)) {
```
codehag: `preview.updating` might be why we are seeing the flicker

src/actions/preview.js
```diff
@@ -80,72 +108,44 @@ export function updatePreview(target: HTMLElement, editor: any) {
       }
     }
 
-    const source = getSelectedSource(getState());
-
-    const symbols = getSymbols(getState(), source.toJS());
-    if (symbols.identifiers.length == 0) {
+    if (isInvalidTarget(target)) {
```
codehag: :+1: nice change

src/actions/preview.js
```diff
@@ -80,72 +108,44 @@ export function updatePreview(target: HTMLElement, editor: any) {
...
-
-    const isUpdating = preview && preview.updating;
+    if (!isLineInScope(getState(), tokenPos.line)) {
+      return;
+    }
 
-    const linesInScope = getInScopeLines(getState());
-    const inScope = linesInScope && linesInScope.includes(location.line);
+    const source = getSelectedSource(getState());
+    const symbols = getSymbols(getState(), source.toJS());
```
codehag: hmmm im a bit surprised that getSymbols needs `toJS` -- is it because its used outside of the parser in this case?

jasonLaster:
the flicker was fixed. that was coming from a timing issue :)

src/actions/preview.js
jasonLaster: the toJS is on the source

src/actions/preview.js
jasonLaster: the flicker was a timing issue that is fixed in this version

GCorbel:
I did some tests with your branch and it seems to very light and sources seems to appears instantly. As you can see [in this memory report](https://github.com/devtools-html/debugger.html/files/1725613/memory-report.json.gz), a single tab opened with the debugger took only 271mb in the "explicit allocations" compared to 320 in the nightly release.

As I said [in my bug report](https://bugzilla.mozilla.org/show_bug.cgi?id=1432051#c14), speed was really improved in latest releases so it's a bit hard to know if this PR does a real difference or not. I attach a report to the comment.

Is it something I should check to see if there is a real difference? Is there a way to be sure I'm using Firefox with your version of the code ?

Thanks a lot for your work.

GCorbel:
I just did a few more tests and I also have 270mb with Firefox nightly so I really don't know there is some improvement.

jasonLaster:
that's fine. i dont think this will actually help memory, just performance. as in preview should show up instantly

GCorbel:
By preview, you mean the little popup which is supposed to appear when we stop the execution on a line with a break point and we pass hover a variable, right ?

In my case, I simply have nothing. When I pass over a variable, nothing appears.

jasonLaster:
yep. oh weird. mind sharing a screenshot of the code?

GCorbel:
Here is the screenshot.

capture du 2018-02-14 17-55-05

My mouse is hover `mod` and there is something in it as you can see in the console. It does the same on the nightly release and the "normal" one too.

jasonLaster:
strange. thanks

GCorbel:
It works but it takes a while for the debugger to be ready. 

Here is a screenshot with my CPU when open the debugger with **Firefox nightly**, wait for the breakpoint and pass hover a variable : 

image

The graph represents 60 seconds and the  60 is the oldest. At second 40, the debugger is ready. From seconds 40 to 20, I pass hover variables but nothing is shown. At seconds 15, I can see the preview. After that, it continues to have a high memory usage but it works.

Here is the same **with your branch** :

image.

It takes the same time to be ready but the previews are correct very soon an it takes a lot less time to complete the same thing.

My conclusion is : your branch is much more efficient.

jasonLaster:
thanks

Issue #4935, opened 2017-12-16
[Sources] Use Immutable records


jasonLaster:
It would be great to explore immutable [records][r1] with the Source I.Map. 

Benefits:
+ we can do `source.id` instead of `source.get("id")`
+ we can potentially drop the `SourceRecord` Flow type from a lot of places
+ Our utils don't have to worry if they're getting an Immutable object or PoJo

I've learned so far that Records are a bit different:
+ They always have all the fields (with a default). this is a good thing
+ Not sure if they destructure well i.e. `const newSource = {...source, {url: "foo"}}`
+ Not sure if they will serialize automatically passed to a worker.task call
+ Not sure if we should pass Records to reducers, e.g. when updating a source... Perhaps the substitute here is that we just pass the source id and new fields

Here's a [branch][b] i created for hacking on it. Already lots of passing tests :confetti_ball: 

https://medium.com/azendoo-team/immutable-record-react-redux-99f389ed676

[r1]: https://gist.github.com/jareware/5f492d47fae45d577e922c431c267c67
[b]:https://github.com/devtools-html/debugger.html/compare/master...jasonLaster:immutable-records?expand=1

DavideDeCenzo:
I've read the articles https://medium.com/azendoo-team/immutable-record-react-redux-99f389ed676 and  https://gist.github.com/jareware/5f492d47fae45d577e922c431c267c67  . From what I read records should destructure well also  on https://facebook.github.io/immutable-js/docs/#/Record They said Records use a custom internal rapresentation for serialization But I think it's not very exhaustive.  I've looked in your branch and I've seen the SourceRecordClass and source.id instead of source.get("id") I guess. Now what should I do? @jasonLaster

jasonLaster:
Thanks @DavideDeCenzo. haha - make it work :)

I didnt spend too much time on it, so i'm sure it's broken in a couple of ways, but hopefully it's not too bad

Also, it could use a rebase!

PR #5389, opened 2018-02-12
[Sources] Use Immutable records

DavideDeCenzo:
Associated Issue: #4935

Here's the Pull Request Doc

Feel free to improve it!

### DIFF
--- a/src/reducers/sources.js
+++ b/src/reducers/sources.js
@@ -47,6 +47,19 @@ export function initialSourcesState(): Record<SourcesState> {
   )();
 }
 
+export const SourceRecordClass = new I.Record({
+  id: undefined,
+  url: undefined,
+  sourceMapURL: undefined,
+  isBlackBoxed: false,
+  isPrettyPrinted: false,
+  isWasm: false,
+  text: undefined,
+  contentType: "",
+  error: undefined,
+  loadedState: "unloaded"
+});
+
 function update(
   state: Record<SourcesState> = initialSourcesState(),
   action: Action
@@ -60,7 +73,8 @@ function update(
     }
 
     case "ADD_SOURCE": {
-      return updateSource(state, action.source);
+      const source = action.source;
+      return updateSource(state, source);
     }
 
     case "ADD_SOURCES": {
@@ -133,7 +147,7 @@ function update(
 
     case "NAVIGATE":
       const source = getSelectedSource({ sources: state });
-      const url = source && source.get("url");
+      const url = source && source.url;
 
       if (!url) {
         return initialSourcesState();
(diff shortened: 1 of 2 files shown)


src/reducers/sources.js
```diff
@@ -324,7 +343,9 @@ function getSourceByUrlInSources(sources: SourcesMap, url: string) {
     return null;
   }
 
-  return sources.find(source => source.get("url") === url);
+  return sources.find(source => {
+    return source.get("url") === url;
```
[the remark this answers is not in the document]
darkwing: Do you happen to know which?  Travis says everything is all good :)

Also, can we do:

sources.find(source => source.get("url") === url);

DavideDeCenzo:
I get these errors If I put `` source.url == url `` 

testsfailed

darkwing:
Wow, you're right; `source.get("url")` returns a URL while `source.url` return `undefined`

DavideDeCenzo:
I've done some logs and I discovered that even the source in the function getPrettySource with `source.url` that I put returns an undefined. Actually It returns an URL with  `source.get("url").`It's strange because the tests are all passed.

sourcesundefined

sourceundefined2

mapvalue

src/reducers/sources.js
```diff
@@ -174,8 +188,13 @@ function updateSource(state: Record<SourcesState>, source: Source | Object) {
...
     return state;
   }
+  const existingSource = state.getIn(["sources", source.id]);
 
-  return state.mergeIn(["sources", source.id], source);
+  if (existingSource) {
+    const updatedSource = existingSource.merge(source);
+    return state.setIn(["sources", source.id], updatedSource);
+  }
+  return state.mergeIn(["sources", source.id], new SourceRecordClass(source));
```
giatorta: this converts the new Record into a Map, see this [immutable bug](https://github.com/facebook/immutable-js/issues/975). This causes the failures to retrieve attributes without using `get()`.

jasonLaster: Thanks! is there a good work around?

jasonLaster:
I think this PR has kinda gone out of scope. @DavideDeCenzo how do you feel about picking up something else while we try and figureout immutable.js!

src/reducers/sources.js
giatorta: hey @jasonLaster... I changed `state.mergeIn()` in L197 to:
`state.setIn(["sources", source.id], new SourceRecordClass(source);`
and every `source.get("x")` with `source.x` and all tests now pass! I think @DavideDeCenzo may try to finalize it

jasonLaster: oh thanks. that's promising.  Re-opening

DavideDeCenzo:
Thanks @giatorta Now It works!

DavideDeCenzo:
@jasonLaster Now it should be better  , the tests are passed also

jasonLaster:
We're in! Great job @DavideDeCenzo

PR #5390, opened 2018-02-12
Fix previewing a file w/ no functions

jasonLaster:
### Summary of Changes

- Previously we assumed files would always have one function if we wanted to preview them.
- now we check for atleast one identifier, which is more general.

### DIFF
--- a/src/test/mochitest/browser.ini
+++ b/src/test/mochitest/browser.ini
@@ -72,6 +72,7 @@ support-files =
   examples/doc-debugger-statements.html
   examples/doc-minified.html
   examples/doc-minified2.html
+  examples/doc-on-load.html
   examples/doc-sourcemaps.html
   examples/doc-sourcemaps2.html
   examples/doc-sourcemaps3.html
@@ -87,6 +88,7 @@ support-files =
   examples/long.js
   examples/math.min.js
   examples/nested/nested-source.js
+  examples/top-level.js
   examples/opts.js
   examples/output.js
   examples/simple1.js
@@ -137,6 +139,8 @@ skip-if = os == "win"
 [browser_dbg-pretty-print-console.js]
 [browser_dbg-pretty-print-paused.js]
 [browser_dbg-preview.js]
+skip-if = os == "win"
+[browser_dbg-preview-module.js]
 skip-if = os == "win"
 [browser_dbg-preview-source-maps.js]
 skip-if = os == "win"
--- a/src/test/mochitest/browser_dbg-preview-module.js
+++ b/src/test/mochitest/browser_dbg-preview-module.js
@@ -1,28 +1,27 @@
 /* Any copyright is dedicated to the Public Domain.
  * http://creativecommons.org/publicdomain/zero/1.0/ */
 
-// Test hovering on an object, which will show a popup and on a
-// simple value, which will show a tooltip.
+// Test hovering in a script that is paused on load
+// and doesn't have functions.
 add_task(async function() {
   const dbg = await initDebugger("doc-scripts.html");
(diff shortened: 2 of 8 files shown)


nchevobbe:
could we add a test to prevent regressing this ?

src/test/mochitest/browser.ini
```diff
@@ -138,6 +140,8 @@ skip-if = os == "win"
 [browser_dbg-pretty-print-paused.js]
 [browser_dbg-preview.js]
 skip-if = os == "win"
+[browser_dbg-preview-module.js]
+skip-if = os == "win"
```
nchevobbe: why do we skip on windows ?

src/test/mochitest/browser_dbg-preview-module.js
```diff
@@ -0,0 +1,27 @@
...
+
+// Test hovering in a script that is paused on load
+// and doesn't have functions.
+add_task(async function() {
+  const dbg = await initDebugger("doc-scripts.html");
+  const { selectors: { getSelectedSource }, getState } = dbg;
+
+  navigate(dbg, "doc-on-load.html")
+
+  await waitForPaused(dbg);
```
nchevobbe: could you add a comment to say that the test page has a debugger statement that will pause the debugger, hence the waitForPaused here ?

src/test/mochitest/browser_dbg-preview-module.js
```diff
@@ -0,0 +1,27 @@
...
+
+  const popupPreviewed = waitForDispatch(dbg, "SET_PREVIEW");
+  hoverAtPos(dbg, { line: 1, ch: 6 });
+  await popupPreviewed;
+  await assertPopup(dbg, { field: "foo", value: "1", expression: "obj" });
+  await assertPopup(dbg, { field: "bar", value: "2", expression: "obj" });
+
+  // hover over an empty position
```
nchevobbe: I suspect this is done to hide the tooltip ? Would you mind adding a comment or an info call to explain that ?

src/test/mochitest/head.js
```diff
@@ -1084,6 +1084,47 @@ function getCM(dbg) {
   return el.CodeMirror;
 }
 
+
+function getCoordsFromPosition(cm, { line, ch }) {
+  return cm.charCoords({ line: ~~line, ch: ~~ch });
```
nchevobbe: I had to lookup this :) So this is done to have a Math.floor, do we ever trigger this with floating numbers ?

src/test/mochitest/head.js
```diff
@@ -1084,6 +1084,47 @@ function getCM(dbg) {
...
+  const previewEl = await waitForElement(dbg, "tooltip");
+  is(previewEl.innerText, result, "Preview text shown to user");
+
+  const preview = dbg.selectors.getPreview(dbg.getState());
+  is(`${preview.result}`, result, "Preview.result");
+  is(preview.updating, false, "Preview.updating");
+  is(preview.expression, expression, "Preview.expression");
+}
+
+async function assertPopup(dbg, { field, value, expression }) {
```
nchevobbe: nit: Could this be assertPreviewPopup ? We do have other things that might be seen as "Popup" (like, the Help modal, even if we call it a modal :D )

src/test/mochitest/browser.ini
jasonLaster: windows is weird w/r/t to the mouse events

src/test/mochitest/head.js
jasonLaster: I believe this helped before and it is copy pasta from elsewhere

PR #5391, opened 2018-02-12
Watch for null selectedFrame.

bhackett1024:
This adds back a change from PR #5175 that was regressed by a more recent change (it will be nice when there are automated tests for this stuff).

### DIFF
--- a/src/components/SecondaryPanes/Scopes.js
+++ b/src/components/SecondaryPanes/Scopes.js
@@ -109,7 +109,7 @@ export default connect(
     const { scope: frameScopes, pending } = getFrameScope(
       state,
       selectedSource && selectedSource.get("id"),
-      selectedFrame.id
+      selectedFrame ? selectedFrame.id : null
     ) || { scope: null, pending: false };
 
     return {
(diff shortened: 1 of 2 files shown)


src/components/SecondaryPanes/Scopes.js
```diff
@@ -109,7 +109,7 @@ export default connect(
     const { scope: frameScopes, pending } = getFrameScope(
       state,
       selectedSource && selectedSource.get("id"),
-      selectedFrame.id
+      selectedFrame ? selectedFrame.id : null
```
jasonLaster: we should have an invariant higher up

e.g. I'm surprised this passed. We can change the definition of isPaused i suppose to
https://github.com/devtools-html/debugger.html/blob/master/src/components/SecondaryPanes/index.js#L240-L247

codehag:
Maybe this is a place where adding flow type checking makes sense as well? Since this is a type error that might be a better approach than tests

jasonLaster:
@bhackett1024 this is what i had in mind. My question is that if `isPaused` is defined was defined as has a frame as opposed to not null, would that solve it?

```diff
diff --git a/src/reducers/pause.js b/src/reducers/pause.js
index c7a097d..131819d 100644
--- a/src/reducers/pause.js
+++ b/src/reducers/pause.js
@@ -244,7 +244,7 @@ export function isStepping(state: OuterState) {
 }

 export function isPaused(state: OuterState) {
-  return !!getFrames(state);
+  return getFrames(state).length > 0;
 }
```

jasonLaster:
we landed an alternate

PR #5395, opened 2018-02-12
Mapping - log exceptions in development

jasonLaster:
we should log in development what broke


codehag:
Can you be a bit more descriptive about what is going on here? Difficult to review with so little information

Issue #5366, opened 2018-02-08
[Search] 2 search bars, one does not work, the other is not focused


Loirooriol:
![screenshot](https://user-images.githubusercontent.com/7477678/35996288-67bfd4b0-0d16-11e8-9567-1904a8c1cd9b.png)

When I press Ctrl+F to search something, a 2nd search bar appears below the real one. And this extra bar does not seem to work properly. The one that works is not focused by Ctrl+F, I must click it manually.

So search is so broken.

jasonLaster:
hm, can you check the browser console?

also do you have any plugins. this is the source editor search showing up.

@gabrielluong did we touch the source editor recently?

Loirooriol:
Regressed by https://bugzilla.mozilla.org/show_bug.cgi?id=1432842

It seems the first time I use Ctrl+F the proper search bar opens, but other uses open the other search bar.

No errors in the browser console, reproducible in clean profile.

![screencast](https://user-images.githubusercontent.com/7477678/35998178-28037524-0d1c-11e8-9f8e-a1754deaaa55.gif)

jasonLaster:
nice find. also i love the keyboard!

darkwing:
I could not reproduce this on my Windows 10 machine.

jasonLaster:
@Loirooriol is this an issue for you on nightly now?

Loirooriol:
Yes, on latest nightly ea3da643422c58d65335f1778dd6c89c09911585

jasonLaster:
@darkwing was testing in the launchpad. We'll now verify in the panel.

darkwing:
I finally got edge debugger working in a Firefox build and can confirm that this is happening. :(

darkwing:
A few additional notes:

* On edge build for Windows:  One Ctrl+F shows the "Search in file" bar but doesn't focus on the input, another Ctrl+F does nothing, then another Ctrl+F shows the second search box, this time focusing on it.

* On LaunchPad for Windows:  Everything works as intended: only the "Search File" box displays and the focus is set properly every time.

* On edge build for Mac:  Everything works as intended: only the "Search File" box displays and the focus is set properly every time.

*

jasonLaster:
Nice writeup!

Here is a useful link to start looking for inconcistencies
https://github.com/devtools-html/debugger.html/blob/master/src/utils/editor/create-editor.js#L15-L32

you can either make changes here and run `node ./bin/copy-assets` w/ the right flags

[...]

PR #5399, opened 2018-02-12
Ensure SearchInput retains value when open and clears when closed

darkwing:
In looking at #5366, I noticed a problem:

* Command+F opens the search and focuses on the `input`, but then clicking into the editor and Command+F'ing again does not.  I would expect the input to be focused and its text selected.

### Summary of Changes

* Clearing the state: query is important 
* Retaining the value when Command+F is pressed is important

### Test Plan

* Open a source in the editor
* Press Command+F, type a keyword
* Click into the editor, press Command+F again, ensure the search text is still there and selected
* Press Escape, the search box should close
* Press Command+F again, the input should display but be clear of text ("")

### Feedback Wanted

* `this.$input.$input.value` seems gross; what's a better way of having these two components interact more cleanly?


src/components/Editor/SearchBar.js
```diff
@@ -136,13 +136,13 @@ class SearchBar extends Component<Props, State> {
...
   closeSearch = (e: SyntheticEvent<HTMLElement>) => {
     const { editor, searchOn } = this.props;
-
     if (editor && searchOn) {
       this.clearSearch();
       this.props.closeFileSearch(editor);
       e.stopPropagation();
       e.preventDefault();
     }
+    this.setState({ query: "" });
```
darkwing: This is required so that the search input value is cleared upon close (i.e. Escape key)

src/components/Editor/SearchBar.js
```diff
@@ -156,7 +156,8 @@ class SearchBar extends Component<Props, State> {
 
     if (this.props.searchOn && editor) {
       const selection = editor.codeMirror.getSelection();
-      this.setState({ query: selection });
+      let query = selection || this.$input.$input.value || "";
```
darkwing: This likely needs some work but you get the idea.

codehag: Interesting! I see how you are solving this, and i agree: its inverting the flow of the two components and directly reaching into the subcomponent, and that doesn't feel right. 

not 100% sure this will work, but have you tried this?:

```javascript
const query = selection ||this.state.query || "";
this.setState({ query: "" }); // clear it to reset the searchInput;
this.setState({ query });  // reset it as needed
```

src/components/Editor/SearchBar.js
```diff
@@ -304,6 +305,7 @@ class SearchBar extends Component<Props, State> {
           handleNext={e => this.traverseResults(e, false)}
           handlePrev={e => this.traverseResults(e, true)}
           handleClose={this.closeSearch}
+          ref={c => (this.$input = c)}
```
codehag: i agree that this feels odd

src/components/Editor/SearchBar.js
codehag: :+1:

src/components/Editor/SearchBar.js
darkwing: @jasonLaster Instead of using a `ref` to directly call a method on a child component, would using redux here be a better option?  If so, can we very briefly pair on it?

src/components/Editor/SearchBar.js
```diff
@@ -304,6 +313,7 @@ class SearchBar extends Component<Props, State> {
...
```
jasonLaster: hmm, we've generally kept only json in redux. @wldcordeiro might have a good idea here

src/components/Editor/SearchBar.js
```diff
@@ -155,10 +157,17 @@ class SearchBar extends Component<Props, State> {
...
+      if (this.$input) {
+        this.$input.setFocus();
+      }
+
+      if (query !== "") {
+        this.setState({ query });
+        this.doSearch(query);
+      } else {
+        this.setState({ query: "" });
       }
```
jasonLaster: looks good

src/components/shared/SearchInput.js
```diff
@@ -55,6 +55,10 @@ class SearchInput extends Component<Props> {
   };
 
   componentDidMount() {
+    this.setFocus();
```
jasonLaster: :thumbsup:

src/components/Editor/SearchBar.js
```diff
@@ -155,10 +157,17 @@ class SearchBar extends Component<Props, State> {
...
-      if (selection !== "") {
-        this.doSearch(selection);
+      const query = editor.codeMirror.getSelection() || this.state.query;
+
+      if (this.$input) {
+        this.$input.setFocus();
+      }
+
+      if (query !== "") {
+        this.setState({ query });
```
wldcordeiro: Maybe use function form?

```js
this.setState(() => {
  this.doSearch(query);
  return { query }
})

File: src/components/Editor/SearchBar.js [hunk 8]
```diff
@@ -155,10 +157,17 @@ class SearchBar extends Component<Props, State> {
...
 
     if (this.props.searchOn && editor) {
-      const selection = editor.codeMirror.getSelection();
-      this.setState({ query: selection });
-      if (selection !== "") {
-        this.doSearch(selection);
+      const query = editor.codeMirror.getSelection() || this.state.query;
+
+      if (this.$input) {
+        this.$input.setFocus();
```
wldcordeiro: An idea here to avoid the inner ref would be making a `focus` state field and having a prop to searchInput like `isFocused` that you assign it to.

src/components/Editor/SearchBar.js
wldcordeiro: Refs aren't that weird anymore they're a pretty common pattern. But I would suggest adding a `isFocused` prop to `SearchInput` and handling the toggling of that value in the state here.

darkwing:
Updated to use state instead of direct child access!

darkwing:
It just occurred to me that using this method of sending props breaks this, in that it will continually call `select` upon each key press; looking into it now.

darkwing:
@jasonLaster @wldcordeiro This is ready for a look!  Thank you!

jasonLaster:
Nice work! satisfy flow and it's in!

PR #5400, opened 2018-02-13
Ensure correct URL copied from tab context menu

darkwing:
The context menu's "Copy Source URI" menu item for pretty tabs is outputting the `:formatted` URL; let's remote that internal `:formatted` signifier from the source.


jasonLaster:
@darkwing wasn't this one of the first things you did two years ago? :)
or was that truncating?

darkwing:
I've come full circle hahaha.  I think it was in the tab name that I removed `:formatted`

Issue #4823, opened 2017-11-29
[Editor] Context menu `Add conditional breakpoints` should focus condition input


digitarald:
STR:
- Open https://mozilladevelopers.github.io/sample-todo/03-conditional-breakpoints/
- Right click on a line number in Debugger and click `Add conditional breakpoints`

AR:
- Input isn't focused

ER:
- Focus input field to start typing condition

This works as expected when the breakpoint is created normally and then made conditional by right clicking `Add condition` in the `Breakpoints` section.

GarethSharpe:
I wasn't able to reproduce the issue. Possible to send a gif of the AR?

![field-focus](https://user-images.githubusercontent.com/22062107/34470903-a1ed6a60-ef09-11e7-8386-8ca9afb9eca1.gif)

digitarald:
Not working for me in Nightly/OSX using the STR.

AlexisDeschamps:
I was not able to reproduce on Nightly and Chrome on Windows 10. Could this be specific to OSX?

jasonLaster:
@AlexisDeschamps yeah - it could be...

AlexisDeschamps:
It seems that the problem arises when the Debugger is used within a panel instead of just the launchpad.

mmcote:
Found that the .focus() is correctly setting itself below in the renderToWidget and am currently searching for where it gets switched to focus elsewhere.

mmcote:
Doing this simple check:

`
const activeElementIsFocused =
document.activeElement == this.input && document.hasFocus();
console.log("Input is focused:", activeElementIsFocused);
`

Right after the this.input.focus() in the renderToWidget it will always return true whether the conditional panel input is focused or not. Whereas, if you add in a componentDidUpdate section and check under that lifecycle event this will accurately return true when the input box shows focused and false when the input box is not focused. 

This tells me that somewhere in between renderToWidget and componentDidUpdate the input loses focus.. sometimes.

Originally I fixed the issue with a setTimeout in the renderToWidget, thereafter I changed the fix to just put the this.input.focus() into the componentDidUpdate.

Currently, I am trying to write a mochi test for the issue.

lukaszsobek:
@digitarald is it fixed for you now?

jasonLaster:
@lukaszsobek not yet. We're still working on landing a patch that passes mochitests

PR #5409, opened 2018-02-13
Added componentDidUpdate to ConditionalPanel.js to only focus on input after render.

mmcote:
Associated Issue: #4823

### Summary of Changes

* Changed the focus of the conditional breakpoint panel's input to after the render in the componentDidUpdate instead of the renderToWidget.

### Test Plan

Added one test, browser_dbg-breakpoints-cond_focus.js, to ensure that the input does focus.

### Before: 
It can be seen that after adding a conditional breakpoint the input does not automatically focus.
before

### After: 
It can be seen now that the input box is acting as expected. Each time a conditional breakpoint is added, the input is focused.
after


mmcote:
Created this pull request looking for some help with the mochi tests, currently the test I uploaded I expect to pass only after the fix. Currently, it works prior to and after the fix. While debugging the corresponding mochi test I noticed that it is passing for good reason as as 100% of the time prior in the mochi test the conditional breakpoint panel is focusing. Although, if I pause the test and add conditional panels myself, I see the same behavior. As a test to ensure I was not missing the input box being focused elsewhere, I removed the one this.input.focus() in the renderToWidget and sure enough it no longer focused. I am just looking for suggestions on how I can possibly better reproduce the bug in a test. 

I tried a couple changes to the click to add a conditional panel but no success.

jasonLaster:
i love this change and the test!

jasonLaster:
great work!

PR #5411, opened 2018-02-13
Sync Bug 1434374 - use ChromeUtils.import instead Cu.import

juliandescottes:
Needed for release 16

Sync changes from https://bugzilla.mozilla.org/show_bug.cgi?id=1434374


juliandescottes:
Travis failed on  browser_dbg-preview-module.js, but looks like this is busted on other PRs as well ...

Issue #5302, opened 2018-02-02
[Editor] Last few columns of editor are cut off.


mmcote:
If the code being debugged has lines long enough enable the horizontal scroll, and you attempt to scroll horizontally to the end of the editor, you see that the last few columns are missing. To be specific the last 30 px of the editor are missing.

Reproduce: 
- Debug a file that has lines long enough to enable the horizontal scroll.
- Try scrolling to the rightmost column of the editor. 
- Observe that your longest line of code will be missing a few columns.

![02beforehorizontalcutoff](https://user-images.githubusercontent.com/14250545/35754155-a7d244b2-081f-11e8-94b6-ec26bca39e0e.gif)

mmcote:
Hey @lukaszsobek, do you mind if I take this bug on, as I was actually working on this prior and found the source of the problem, I was just waiting for the issue to be claimable. Sorry for the trouble :)

lukaszsobek:
@mmcote Sure thing, go ahead! /unclaim

mmcote:
Thank you @lukaszsobek

PR #5412, opened 2018-02-14
Show rightmost columns of the editor.

mmcote:
Associated Issue: #5302

### Summary of Changes

* Reverted min-width: 0 for CodeMirror-sizer that modified the codemirror's fundamental behaviour, hiding the rightmost columns.
* Removed close button which was aligned with the previously moving code-mirror-sizer as it was not needed since the close conditional breakpoint panel action is also invoked onBlur.
* Switched CodeMirror-linewidget overflow to hidden to prevent infinite horizontal scrolling when repositioning the conditional breakpoint panel (when the conditional breakpoint panel is open).

### Screenshots/Videos 
Before, as can be seen in the gif below, the rightmost columns are unable to be seen when horizontally scrolling to the right.
before

After, as can be seen in the gif below, the rightmost columns are now accessible when scrolling horizontally to the right.
aftercolumns

### DIFF
--- a/src/components/Editor/ConditionalPanel.css
+++ b/src/components/Editor/ConditionalPanel.css
@@ -11,7 +11,6 @@
   background: var(--theme-toolbar-background);
   border-top: 1px solid var(--theme-splitter-color);
   border-bottom: 1px solid var(--theme-splitter-color);
-  padding-right: 16px;
 }
 
 .conditional-breakpoint-panel .prompt {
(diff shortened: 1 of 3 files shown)


jasonLaster:
This looks great @mmcote!

Does this address the lifecycle event w/ focusing?

mmcote:
Thank @jasonLaster :) no I put the focusing issue in another pull request: https://github.com/devtools-html/debugger.html/pull/5409.

PR #5413, opened 2018-02-14
Map live import bindings in the original file to member expressions in the generated code

loganfsmyth:
### Summary of Changes

* Expose metadata about the expression a given binding reference is part of in the scope walking visitor
* Use that information to map live import bindings in the original code back to member expressions in the generated code

### Test Plan

Added a bunch more mochitests and some Jest snapshots for the scope binding metadata logic

##### Babel's Module compilation
Currently still fails because Babel does not generate sourcemaps good enough to use for imports. I'll be submitting a PR over there to fix that in the future.
screen shot 2018-02-13 at 5 33 41 pm

##### Webpack's imports post-babel-loader

Webpack's standard imports don't get mapped well, but if the file is passed through Babel _first_ the bindings get identifier-level maps before Webpack does its own module transform.

Note in this case the named imports show as `Getter` because the imported modules use getters.
screen shot 2018-02-13 at 5 34 06 pm

##### Webpack's imports post-babel-loader, but without Babel transforming let/const

Same as previous, except that the code is left as `export const foo = "";`, Webpack can skip the getter entirely in favor of just exporting the values, so they show up nicer.

screen shot 2018-02-13 at 5 34 16 pm


src/actions/pause/mapScopes.js
```diff
@@ -291,6 +284,168 @@ async function findGeneratedBinding(
...
+      const hasCall = operations.some(op => op.type === "call");
+      if (hasCall) {
+        // In theory this could fall back to trying to use client.evaluate.
+        return null;
+      }
+
+      let desc = val.desc;
+      if (desc) {
+        for (const op of operations) {
+          if (op.type === "call" || op.type === "inherit") {
```
jasonLaster: can you add some comments here :)

src/actions/pause/mapScopes.js
```diff
@@ -291,6 +284,168 @@ async function findGeneratedBinding(
...
+  if (
+    !allowExpressionMatch &&
+    mapped.start.line === binding.loc.start.line &&
+    locColumn(mapped.start) >= locColumn(binding.loc.start) - 1 &&
+    locColumn(mapped.start) <= locColumn(binding.loc.end)
+  ) {
+    return [];
+  }
+
+  if (allowExpressionMatch) {
```
jasonLaster: could this be an early return 

```js
if (!allowExpressionMatch) {
  return;
}
```

src/actions/pause/mapScopes.js
```diff
@@ -291,6 +284,168 @@ async function findGeneratedBinding(
...
+    //
+    // Unfortunately, Webpack also has a tendancy to over-map past the call
+    // expression to the start of the next line, at least when there isn't
+    // anything else on that line that is mapped, e.g.
+    //
+    //   Object(foo.bar)()
+    //   ^^^^^^^^^^^^^^^^^
+    //   ^                 // wrapped to column 0 of next line
+
+    if (mappingContains(mapped, binding.loc)) {
```
jasonLaster: same here:
```js
    if (!mappingContains(mapped, binding.loc)) {
      return null;
    }
      ```

File: src/actions/pause/mapScopes.js [hunk 4]
```diff
@@ -291,6 +284,168 @@ async function findGeneratedBinding(
...
+      const { meta } = binding.loc;
+
+      // Limit to 2 simple property or inherits operartions, since it would
+      // just be more work to search mroe and it is very unlikely that
+      // bindings would be mapped to more than a single member + inherits
+      // wrapper.
+      const operations = [];
+      for (let op = meta, i = 0; op && i < 2; i++, op = op.parent) {
+        // Allow calls as the first access, "foo().bar", not but "foo.bar()"
```
jasonLaster: > not but => ,but not

src/actions/pause/mapScopes.js
```diff
@@ -291,6 +284,168 @@ async function findGeneratedBinding(
...
+
+    if (mappingContains(mapped, binding.loc)) {
+      const { meta } = binding.loc;
+
+      // Limit to 2 simple property or inherits operartions, since it would
+      // just be more work to search mroe and it is very unlikely that
+      // bindings would be mapped to more than a single member + inherits
+      // wrapper.
+      const operations = [];
+      for (let op = meta, i = 0; op && i < 2; i++, op = op.parent) {
```
jasonLaster: silly, but can `i` be `index`

src/actions/pause/mapScopes.js
```diff
@@ -291,6 +284,168 @@ async function findGeneratedBinding(
...
+        // because that is too likely to accidentally pick up a call that
+        // was in the user's own code, and it would be too risky. The chance
+        // of an expression mapping "one" to "_one()" isn't a common transform
+        // and is also one that is less likely to be screwed up during
+        // sourcemap generation, so it is allowed.
+        if (op.type === "call" && i !== 0) {
+          break;
+        }
+
+        if (mappingContains(mapped, op)) {
```
jasonLaster: could this be an early return

```js
if (!mappingContains(mapped, op)) {
  break;
}
```

src/actions/pause/mapScopes.js
```diff
@@ -291,6 +284,168 @@ async function findGeneratedBinding(
...
+        name: val.name,
+        desc: desc
+      };
+    }
+
+    return null;
+  }, null);
+}
+
+function mapBindingToGeneratedRange(
```
jasonLaster: can this be commented

hmm... is there a better way to describe returning operations... I wouldnt guess that is what's happening with the method name

src/actions/pause/mapScopes.js
```diff
@@ -291,6 +284,168 @@ async function findGeneratedBinding(
   };
 }
 
+async function findGeneratedBindingFromPosition(
```
jasonLaster: this might just be my style, but it would be easier to follow if we had a `utils/pause/mapScopes` which had a module `findGeneratedBindingFromPosition.js` so i can see what's needed for this algorithm and what's public and what's private to it

src/actions/pause/mapScopes.js
loganfsmyth: Good call.

src/actions/pause/mapScopes.js
loganfsmyth: So make `mapScopes` a folder with `index.js` and `findGeneratedBindingFromPosition.js`?

jasonLaster: yep

jasonLaster:
here's what it looked like before

Issue #5416, opened 2018-02-14
Hide "file modal" when pressing Cmd + Shift + F


nchevobbe:
Steps to reproduce:
1. Open this page
2. Hit <kbd>Cmd</kbd>+<kbd>P</kbd> (this should show the file modal)
3. Hit <kbd>Cmd</kbd>+<kbd>Shift</kbd>+<kbd>F</kbd> (this start the search in project)

Expected results:
The file modal is dismissed and I can type what i'm searching for

Actual results:
The search UI appears underneath the file modal, which is still visible: 

<img width="829" alt="screen shot 2018-02-14 at 08 03 17" src="https://user-images.githubusercontent.com/578107/36191509-dfcc0d70-115d-11e8-9c8b-ed57cd0de0df.png">

---

We should hide the modal when showing the search UI.

lukaszsobek:
Merged. Closing.

PR #5417, opened 2018-02-14
Closes QOM when file or project search opens

lukaszsobek:
Associated Issue: #5416

Closes quick open modal when either file search or project search are being opened.

Vid:
closes-qom


src/components/Editor/SearchBar.js
```diff
@@ -148,13 +149,14 @@ class SearchBar extends Component<Props, State> {
...
     e.stopPropagation();
     e.preventDefault();
-    const { editor } = this.props;
+    const { editor, closeQuickOpen, searchOn, setActiveSearch } = this.props;
 
-    if (!this.props.searchOn) {
-      this.props.setActiveSearch("file");
+    if (!searchOn) {
+      closeQuickOpen();
+      setActiveSearch("file");
```
jasonLaster: can we have `setActiveSearch` close quick open if it's open? that will simplify the component

src/components/ProjectSearch.js
```diff
@@ -117,14 +118,16 @@ export class ProjectSearch extends Component<Props, State> {
...
+    const { closeProjectSearch, setActiveSearch, closeQuickOpen } = this.props;
     if (e) {
       e.preventDefault();
     }
 
     if (this.isProjectSearchEnabled()) {
       return closeProjectSearch();
     }
+
+    closeQuickOpen();
```
jasonLaster: same here

src/components/Editor/SearchBar.js
lukaszsobek: Sure.

src/actions/ui.js
```diff
@@ -38,6 +38,8 @@ export function setActiveSearch(activeSearch?: ActiveSearchType) {
       return;
     }
 
+    dispatch({ type: "CLOSE_QUICK_OPEN" });
```
jasonLaster: lets add an if statement so we only fire it if it's open.

we can also add a unit test here where we open the modal and then fire SAS

jasonLaster:
try rebasing

darkwing:
I'm still seeing layering with the PR:

dialogs

I did CMD+F, then CMD+P, and they both display.

lukaszsobek:
@darkwing - sure, from what I understood it was supposed to close the modal when we open projectsearch or searchbar, not prevent opening it altogether. So:
- when the modal `cmd+p` is open and you open `cmd+shift+f` or `cmd+f` the modal will close.
- when `cmd+shift+f` or `cmd+f` is already open and you press `cmd+p` it will not prevent opening the modal since you clearly intended to open the modal.

We might be talking about another issue here: "if projectsearch/searchbar is open and we open the modal, should we close the projectsearch/searchbar?". The issue is very much related, but I see two things here:
1.) do we want that behavior?
2.) should this issue be handled separately as it seems to refer to a different part of UI behavior?

PR #5435, opened 2018-02-15
WIP: Fix #5434: Avoid selectLocation action dance with tabs and pretty print

### LINKED ISSUE #5434: [PrettyPrint] selectLocation action opens unwanted original tab
At present the `autoPrettyPrint` pref is true, and so clicking a source in the source tree should simply open the pretty printed tab.

Unfortunately the current process is a bit hacky:

1.  Open source tab
2.  Open pretty tab (via `togglePrettyPrint`)
3.  Close source tab

We're doing a ton of expensive work we shouldn't need to do.  The only thing we should be doing is step 2.

darkwing:
This is a WIP to get feedback.  A few notes:

1.  I left the console.log messages there for reviewer's manual testing

2.  I feel like `preserveOriginalSource` may not be the best answer to the pickle we're in, but finding the line between automagically pretty printing vs. expressly telling selectLocation to preserve original source is going to be really hard.  Maybe it's best we document all the places we'd want original source preserved?

3.  One issue right now is if you pretty print, then click to see the original source (in the footer), then go back to the pretty print tab, then click the original source's tab, nothing happens, because `selectLocation` is directing the pretty print to happen.

4.  Need to add tests.

### DIFF
--- a/src/actions/sources/select.js
+++ b/src/actions/sources/select.js
@@ -195,7 +195,7 @@ export function jumpToMappedLocation(location: Location) {
       );
     }
 
-    return dispatch(selectLocation({ ...pairedLocation }, null, true));
+    return dispatch(selectLocation({ ...pairedLocation }, "", true));
   };
 }
 
--- a/src/actions/sources/select.js
+++ b/src/actions/sources/select.js
@@ -111,28 +111,38 @@ export function selectLocation(location: Location, tabIndex: string = "") {
 
     dispatch(addTab(source.toJS(), 0));
 
-    dispatch({
-      type: "SELECT_SOURCE",
-      source: source.toJS(),
-      tabIndex,
-      location
-    });
-
     await dispatch(loadSourceText(source));
-    const selectedSource = getSelectedSource(getState());
+
+    console.log('source text should be loaded!', source.get("text"));
+
+    const selectedSource = getSelectedSource(getState()) || source;
     if (!selectedSource) {
       return;
     }
 
     const sourceId = selectedSource.get("id");
+
+    console.log('--------- ', source, selectedSource);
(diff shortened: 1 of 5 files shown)


src/actions/sources/select.js
```diff
@@ -90,15 +91,21 @@ export function selectSource(sourceId: string, tabIndex: string = "") {
  * @memberof actions/sources
  * @static
  */
-export function selectLocation(location: Location, tabIndex: string = "") {
+export function selectLocation(
+  location: Location,
+  tabIndex: string = "",
+  preserveOriginalSource: boolean = false
```
jasonLaster: re: `preserveOriginalSource`
Lets see what else we can do here...

maybe selectLocation needs to differentiate between the first time we are going to the source where we 
a) might want to create the pretty source
b) need to setSymbols, setOutOfScopeLocations

and selecting a source in general. 

Lastly, could we move the decision for which source to another function because *that* is complicated. Here are two ideas:

1. source tree on click checks to see if the source has a pretty file and if so calls selectLocation with that source
2. source tree calls a function like `selectBestLocation` where that function checks to see if there is a pretty file and if so calls that... i dunno it's hard as we discussed :)

darkwing: I somewhat agree that `selectLocation` is doing too much; the problem is that `selectLocation` gets called from *many* different places:

1.  SourceTree
2.  Tab (individual tab click)
3.  Quick open / search
..... and so on

The only two spots we *know* which tab (original or pretty) should be selected is clicking the individual tab from the tab bar and clicking the footer link from the pretty tab -- they should both be original.  In all other cases, we will want to go with the prefs.autoPrettyPrint choice, right?  Just thinking out loud.

PR #5440, opened 2018-02-15
Use babel7 imports

jasonLaster:
this reduces the parser worker bundle from 60K to 54K

dropping babel-traverse will bring us down to 30K...

screen shot 2018-02-15 at 6 35 57 pm
screen shot 2018-02-15 at 6 35 49 pm


darkwing:
Whoa!

PR #5442, opened 2018-02-16
Truncated frames text (2nd of 4 EndTruncateStr PRs) 

CharlesStocky:
I couldn't (for the life of me) figure out how to truncate with ellipsis, however this CSS does truncate the title string. I'll eventually come back and figure out how to truncate with ellipsis.  I also added a tooltip to the frames. 

Prior to fix: 

screen shot 2018-02-15 at 9 58 17 pm
screen shot 2018-02-15 at 9 58 28 pm
screen shot 2018-02-15 at 10 45 12 pm

Post fix: 

screen shot 2018-02-15 at 10 48 12 pm

screen shot 2018-02-15 at 10 48 21 pm

### DIFF
--- a/src/actions/preview.js
+++ b/src/actions/preview.js
@@ -9,10 +9,11 @@ import { getTokenLocation } from "../utils/editor";
 import { isReactComponent, isImmutable } from "../utils/preview";
 import { isGeneratedId } from "devtools-source-map";
 import { PROMISE } from "./utils/middleware/promise";
+import { getExpressionFromCoords } from "../utils/editor/get-expression";
 
 import {
   getPreview,
-  getInScopeLines,
+  isLineInScope,
   getSelectedSource,
   getSelectedFrame,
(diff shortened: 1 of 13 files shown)


darkwing:
Yeah, it looks like Frames is going to be a tough one; after a bit of width toggling, I see:

framesdots

The word `(global)`, as is the file name, is being obscured, and the line number goes away too:

callstackline

Try removing `display: flex` to `.frames .location` -- that may do the trick! :

CharlesStocky:
Latest push should fix the issue. I'm starting to gain some footing with git, so things should be a lot cleaner from now on :D.

darkwing:
Thanks for contributing this @CharlesStocky!  After talking with our UX person, we're going to reimagine the design, including a grid layout, so I think we'll close these for the time being and re-open when the grid and such is ready.  Thank you!

Issue #5428, opened 2018-02-15
Typing filename:linenumber in the File search should navigate me to the file at given line


nchevobbe:
**Steps to reproduce:**
1. Go to http://firefox-dev.tools/debugger-examples/examples/todomvc/
2. Open the debugger
3. Open the File search (<kbd>Cmd</kbd>+<kbd>P</kbd>)
4. Type `todo-view.js:62`

**Expected results:**

The debugger show me the todo-view.js file as a potential result.
And when I press <kbd>Enter</kbd>, I'm taken to todo-view.js, at line 62.
Eventually, it would be nice if `todo-view.js:62:100` would take me to line 62 and column 100.

**Actual results:**
I get a sad red face : 

<img width="172" alt="image" src="https://user-images.githubusercontent.com/578107/36248122-a86ba064-1236-11e8-966e-6574fb3f9296.png">

---

This can be very useful when copying a stacktrace frame. vscode does this quite well.

jasonLaster:
so... it should support line/column jumping. The sad face shouldn't be there :)

lukaszsobek:
A little more insight:
Typing the query does actually seem to produce the desired results. The thing that does not work is pasting the line - it doesn't trigger the onChange event which in turn does not update the query type from `sources` to `gotoSource` and that gives us a modal trying to look for a file that is literally named `test.js:60` instead of trying to go to `test.js` on line "60".

PR #5444, opened 2018-02-16
Removes sad face when source+goto

lukaszsobek:
Associated Issue: #5428

- Removes sad face when using source+goto queries in QOM search
- handles source goto queries

Vid:
removes-sad-goto

### DIFF
--- a/src/components/QuickOpenModal.js
+++ b/src/components/QuickOpenModal.js
@@ -110,10 +110,14 @@ export class QuickOpenModal extends Component<Props, State> {
     this.props.closeQuickOpen();
   };
 
+  dropGoto = (query: string) => {
+    return query.split(":")[0];
+  };
+
   searchSources = (query: string) => {
     const { sources } = this.props;
     const results =
-      query == "" ? sources : filter(sources, query.split(":")[0]);
+      query == "" ? sources : filter(sources, this.dropGoto(query));
     return this.setState({ results });
   };
 
--- a/src/components/QuickOpenModal.js
+++ b/src/components/QuickOpenModal.js
@@ -111,18 +111,10 @@ export class QuickOpenModal extends Component<Props, State> {
   };
 
   searchSources = (query: string) => {
-    if (query == "") {
-      const results = this.props.sources;
-      return this.setState({ results });
-    }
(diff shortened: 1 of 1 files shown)


src/components/QuickOpenModal.js
```diff
@@ -115,7 +115,7 @@ export class QuickOpenModal extends Component<Props, State> {
       const results = this.props.sources;
       return this.setState({ results });
     }
-    if (this.isGotoSourceQuery()) {
+    if (this.isGotoSourceQuery() || this.isSourcesQuery()) {
```
lukaszsobek: Not really sure we need such a rigid separation between `isGotoSourceQuery` and `isSourcesQuery` as `isGotoSourceQuery` seems more like a subtype of `isSourcesQuery` than a fully fledged equivalent. Then there is `isSourceSearch`, which should have been used above in checking the type in the scenario we have now.

jasonLaster: hmm not sure what this does. can you add a unit test?

lukaszsobek: Sure, this causes the error emoji to disappear if the source has a `:`, but it is only half of the issue - the other half is that when the input is in `SourcesQuery` mode it does not consider the location suffix.

src/components/QuickOpenModal.js
```diff
@@ -111,18 +111,10 @@ export class QuickOpenModal extends Component<Props, State> {
...
-      const [baseQuery] = query.split(":");
-      const results = filter(this.props.sources, baseQuery);
-      this.setState({ results });
-    } else {
-      const results = filter(this.props.sources, query);
-      this.setState({ results });
-    }
+    const { sources } = this.props;
+    const results =
+      query == "" ? sources : filter(sources, query.split(":")[0]);
```
jasonLaster: lets move this to its own function `query.split(":")[0]`

`filter(sources, dropGoto(query))`

i dont like the term drop but it is nice to think about the language we use

darkwing:
Part of me wonders if this is too simplistic.  I've seen ":" used in URLs for files before, plus we have our own ":formatted" syntax. Maybe we should try to make it smarter, looking for :[digits]?

PR #5450, opened 2018-02-16
Fix locale property

jasonLaster:


assets/panel/debugger.properties
```diff
@@ -47,7 +47,7 @@ copyStackTrace.accesskey=c
 # that expands the left and right panes in the debugger UI.
 expandPanes=Expand panes
 
-# LOCALIZATION NOTE (executeInConsole): Editor right-click menu item
+# LOCALIZATION NOTE (evaluateInConsole): Editor right-click menu item
```
flodolo: Should be evaluateInConsole.label

jasonLaster: cool

PR #5452, opened 2018-02-16
Fix stepping w/ await

jasonLaster:
currently we miss the opportunity to use our await stepping if we're paused at `await <here>foo()` this fixes that by updating the regexy check


jasonLaster:
still have a test to fix

PR #5453, opened 2018-02-16
Revert "Fixes yarn warnings (#5419)"

jasonLaster:
This reverts commit 8b930f56f7460000f06090f1324e5f86443586a0.
For some reason we need the dashes so that `yarn mochih dbg-sources` only runs the sources tests


jasonLaster:
:banana:

PR #5455, opened 2018-02-16
Ensure the source tree highlights the correct item, regardless of pretty or original print

darkwing:
So this has been really bugging me, had to fix it.

At present, if you open a pretty print tab (which now happens automagically when selecting a source from tree), the file is no longer highlighted in the tree.  This is because we were comparing sources; since PP and Original are different sources, and the pretty print URL was requested to highlight, but doesn't exit in the tree, nothing would be highlighted.

Additionally, I tried using an additional check:

```js
!selectedSource || selectedSource.get("url") != nextProps.selectedSource.get("url")
```

... but if you clicked on the tab you were already looking at, the highlight would go away.

In the end, I get why we did the checks we did but they have side effects.  I think simplifying this highlight check will do us wonders!

### DIFF
--- a/src/components/PrimaryPanes/SourcesTree.js
+++ b/src/components/PrimaryPanes/SourcesTree.js
@@ -124,11 +124,15 @@ class SourcesTree extends Component<Props, State> {
     }
 
     if (nextProps.selectedSource) {
-      const highlightItems = getDirectories(
-        getRawSourceURL(nextProps.selectedSource.get("url")),
-        sourceTree
-      );
-      return this.setState({ highlightItems });
+      const highlightUrl = nextProps.selectedSource.get("url");
+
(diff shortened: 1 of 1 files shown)


darkwing:
BTW, as I test this, I think it may inadvertently fix the left-arrow-going-AWOL problem, mainly because the highlight doesn't disappear when the pretty print feature is opened.

darkwing:
With regard to calls:  

```
WITH THIS PR:
    Opening a new source:  5
    Changing sources:  1

WITH OLD CODE:
    Opening a new source:  3
    Changing sources: 1
```

darkwing:
And, of course, all tests pass for me locally. *sigh*

jasonLaster:
yep we can afford 2ms :P

PR #5459, opened 2018-02-17
[CallStack] - Refactor and test existing frame grouping behavior

calebstdenis:
Associated Issue: #5150 

This precursor PR includes some refactoring and adds tests to pave the way for the implementation, which will be the next PR. 

### Summary of Changes
* Moved frame grouping logic so that it runs in the selector, instead of the component render function.
* Added some tests to cover the existing grouping behavior.

### Test Plan

1. Debug the [React example](http://firefox-dev.tools/debugger-examples/examples/my-app/build/).
2. Set a breakpoint on App.js line 14. *React and Webpack frames still grouped as expected in the call stack area.*
4. Open the call stack context menu and click "Disable framework grouping." *Frames are ungrouped as expected.*
5. Re-enable framework grouping. *Frames are grouped as expected*

### Screenshots
This is what the call stack looks like when breaking in an async function transformed by Babel.

Before an `await`:
image

After an `await`:
image 1


src/selectors/tests/getCallStackFrames.spec.js
```diff
@@ -0,0 +1,228 @@
...
+        [
+          expect.objectContaining(state.frames[0]),
+          expect.objectContaining(state.frames[1]),
+          expect.objectContaining(state.frames[2])
+        ]
+      ]);
+    });
+
+    describe("groups Babel frames for transforming async functions", () => {
+      // eslint-disable-next-line max-nested-callbacks
```
calebstdenis: I had a choice here between:
* Having two tests that exceed max level of nesting (what I did)
* Having two tests with really long names that exceed the max line length.
* Having two tests with short but rather cryptic names.

I felt I picked the lesser evil but I'm open to being told otherwise. :)

jasonLaster: haha - i dont care you can always bump the max level of nesting

src/selectors/getCallStackFrames.js
```diff
@@ -39,23 +40,30 @@ function appendSource(sources, frame, selectedSource) {
...
 
-  return frames
+  let formattedFrames = frames
     .filter(frame => getSourceForFrame(sources, frame))
     .map(frame => appendSource(sources, frame, selectedSource))
     .filter(frame => !get(frame, "source.isBlackBoxed"))
     .map(annotateFrame);
+
+  if (frameworkGroupingOn) {
+    formattedFrames = collapseFrames(formattedFrames);
```
jasonLaster: i think the reason i avoided doing this earlier was that the collapsing messed with the count of frames which was important for expanding and collapsing.

The other reason was that it made toggling `frameworkGroupingOn` a bit different... 

This way works, but I wonder if it is needed. If the main benefit of doing the collapsing in the selector is testing, could we re-think how we're doing the testing?

for instance what is the input to the unit test? if it is just an array of frames which have the library field, we dont need to think about annotateFrame or the selector at all.

calebstdenis: >i think the reason i avoided doing this earlier was that the collapsing messed with the count of frames which was important for expanding and collapsing.
I suspected this might be the case but couldn't find any scenario where this problem would occur. I'm not 100% confident though so I guess I shouldn't be changing that. 

>This way works, but I wonder if it is needed. If the main benefit of doing the collapsing in the selector is testing, could we re-think how we're doing the testing? for instance what is the input to the unit test? if it is just an array of frames which have the library field, we dont need to think about annotateFrame or the selector at all.
The issue with that is, deciding what library a frame belongs to is at least half of the logic for framework grouping. I think we should still have tests to cover that. So, one test suite to make sure frames are annotated properly, and another to ensure frames are grouped based on the annotation. 

It's less than ideal to test against frame annotation, because that's an implementation detail. That was my thinking for this PR. Then again, maybe I've changed more than I need to for the "good enough" solution :/

jasonLaster: makes sense

Issue #4879, opened 2017-12-07
Update to React 16.2.0


nchevobbe:
This is needed so the whole devtools toolbox can also switch to React 16 as well.
Here's what we should do: 

- [x] Update React packages to 16.2.0 (use separate packages for proptypes and react-dom-factories). Maybe @MikeRatcliffe could help us here.
- [x] Update all the dependencies that would not work in their current version with React 16 (see also https://github.com/devtools-html/debugger.html/issues/4848)
- [x] Because devtools-reps will be updated, the `<ObjectInspector>`instances should use the new API (pass a `createObjectClient`prop instead of using `loadedProperties` and `loadProperties`). @jasonLaster commited to this
- [ ] Because of the update of devtools-reps, switching from horizontal to vertical layout (or the other way around) will collapse any expanded nodes by the user. To overcome this, we'll switch to a css grid layout and media-queries to switch from one layout to another (See https://github.com/devtools-html/debugger.html/issues/4839). React 16 allow us to return Fragments from components so we can put the main parts as direct children of the `.debugger` node. @codehag commited to this.
- [ ] Because of previous point, we won't be able to use the `<Splitter>` component (which works by wrapping 2 children, which we don't want). A new `<GridAreaResizer>` component will be created to be able to resize a grid track. @nchevobbe commited to this.

jasonLaster:
@nchevobbe surprised me earlier when he informed me of the large `set_operation` performance time. It looks like it's about 200ms:

My take away is that we can probably optimize this quite a bit. We probably should as well as this is likely our perf when opening the debugger for the first time.

<img src='https://shipusercontent.com/ea8d10fcb6bd461bd6cbc9c346bee55f/Screen%20Shot%202017-12-07%20at%204.44.42%20PM.png' title='Screen Shot 2017-12-07 at 4.44.42 PM.png' width=1074>
<img src='https://shipusercontent.com/af959227c1a608a2050839305ef65485/Screen%20Shot%202017-12-07%20at%204.44.31%20PM.png' title='Screen Shot 2017-12-07 at 4.44.31 PM.png' width=887>

---

<img src='https://shipusercontent.com/38cd705a2cb141f1962a5b9a0d73da2a/Screen%20Shot%202017-12-07%20at%204.46.11%20PM.png' title='Screen Shot 2017-12-07 at 4.46.11 PM.png' width=730>
[...]

PR #5465, opened 2018-02-19
[WIP] Upgrade to react 16.2.0

AnshulMalik:
Associated Issue: #4879 

### Summary of Changes

* React and React DOM to 16.2.0


jasonLaster:
Thanks @AnshulMalik could you update the jest tests.

i can look into these license

AnshulMalik:
Yep, working on the tests :)

darkwing:
Hey @AnshulMalik !  Thank you for this!

Could you add a checklist to the description so we can track the remaining WIP items?  It's something I'd like to try out to keep the PR queue "clean" moving forward.  Thank you!

jasonLaster:
^ @nchevobbe did we address this in core?

nchevobbe:
just run license-check on devtools-core without issues.
I guess we should evaluate those dependencies and their licences

src/components/Editor/tests/Editor.spec.js
```diff
@@ -117,7 +117,6 @@ describe("Editor", () => {
       });
 
       expect(mockEditor.setText.mock.calls).toEqual([
-        [""],
```
AnshulMalik: Is it okay that this one call is being skipped?

jasonLaster:
I'm pleasantly surprised that we don't pull in react 15.x

src/components/Editor/tests/Editor.spec.js
jasonLaster: yep

src/components/tests/__snapshots__/QuickOpenModal.spec.js.snap
```diff
@@ -1458,14 +1458,14 @@ exports[`QuickOpenModal updateResults on enable 2`] = `
     >
       <Modal
         handleClose={[Function]}
-        status="entering"
+        status="exited"
```
nchevobbe: is this intended ?

jasonLaster: yeah - i'm scared of that too... i feel like we have that problem in the panel perhaps, so we should land this and then test that...

jasonLaster: i guess i could also test it now... before it lands :wink:

nchevobbe: yeah, i think it would be safer :)

darkwing:
When toggling the right pane, I see: "Warning: Expressions has a method called shouldComponentUpdate(). shouldComponentUpdate should not be used when extending React.PureComponent. Please extend React.Component if shouldComponentUpdate is used"

That doesn't appear to happen on master.

darkwing:
`Warning: EditorMenu has a method called shouldComponentUpdate(). shouldComponentUpdate should not be used when extending React.PureComponent. Please extend React.Component if shouldComponentUpdate is used.`

It looks like our PureComponent usage is wrong for this upgrade.

src/components/tests/__snapshots__/QuickOpenModal.spec.js.snap
jasonLaster: Quick open slides in and slides out. It is also fine in the other 10 unit tests. 

I am going to move forward and we can figure this mystery out later. I believe it is just a timing thing.

jasonLaster:
Same w/ the jest tests. I think this can also be a follow up as it is a warning, but not an error:

jasonLaster:
The necessity for moving forward is that we need to be on the same version of react as the panel to catch bugs that only come up between the differences.

PR #5469, opened 2018-02-19
Update devtools-reps to 0.20.0

nchevobbe:


nchevobbe:
There was a failing mochitest which I fixed in another commit. Let's see what CI says now

nchevobbe:
there's still a failure in mochi, but seems unrelated. Can we merge this ?

jasonLaster:
yep

PR #5471, opened 2018-02-19
Remove double tilde

jasonLaster:
Drops the double tilde from the mochitest because its unnecessary in this context.


jasonLaster:
it wasnt the tilde that was the problem

Issue #4964, opened 2017-12-21
[Breakpoint] Conditional breakpoint UI can be opened more than once (and can't be closed afterwards)


juliandescottes:
STRs:

- Go to http://firefox-dev.tools/debugger-examples/examples/todomvc/
- open Debugger
- open any source (e.g. app-view,js)
- right click on the gutter to add a conditional breakpoint on line 11
- right click on the gutter to add a conditional breakpoint on line 21
- right click on the gutter to add a conditional breakpoint on line 30
- click in the UI that was opened in the previous step
=> This last UI is closed

After that, all the remaining inputs for conditional breakpoints are broken. They can't be validated or closed.

![conditional_bp_bug](https://user-images.githubusercontent.com/1141550/34267080-d50bec28-e67b-11e7-8377-8f17ae66402e.gif)

lukaszsobek:
What's the expected behavior in this case?

nyrosmith:
I can reproduce it only in the panel. In launchpad it works as expected.
Not sure what's going wrong there...

jasonLaster:
Lets see if we can investigate a bit before we mark it available...

nyrosmith:
Why shouldn't this work in the panel too? https://github.com/devtools-html/debugger.html/blob/f9149b7d312f902a532c235d1e5bc90a2dca30ca/src/components/Editor/ConditionalPanel.js#L150

jasonLaster:
this should be fixed because of the onblur

PR #5480, opened 2018-02-20
Prevent the re-rendering of a conditional breakpoint panel (cbp) when…

mmcote:
Prevent the re-rendering of a conditional breakpoint panel (cbp) when opening a new cbp or removing or disabling an open cbp.

Associated Issue: #4964 

### Summary of Changes

- When rendering, if the conditional breakpoint panel is open, then close it before rendering a new instance.
- For situations where the renderToWidget is called before the closeConditionalPanel action can be sent out, send out the action before the conditional panel gets dereferenced and return. (ex. Removing or disabling a conditional breakpoint when editing)

### Screenshots/Videos

Before:
- As can be seen when adding multiple conditional panels without closing the previous instances a new panel would be made but the old panel would remain and would have no ability to clear it out.
- When editing a conditional breakpoint, if we remove or disable it the breakpoint will once again remain and be unable to be interacted with.
beforefix

After:
- Can add conditional breakpoint panels with the prior instance being cleared.
- Can add a new conditional breakpoint panel/remove/disable the conditional breakpoint panel even while editing and it will properly be cleared out.
afterfix

### DIFF
--- a/src/components/Editor/ConditionalPanel.js
+++ b/src/components/Editor/ConditionalPanel.js
@@ -104,6 +104,13 @@ export class ConditionalPanel extends PureComponent<Props> {
   }
 
   renderToWidget(props: Props) {
+    if (this.cbPanel) {
+      if (this.props.line && this.props.line == props.line) {
+        return props.closeConditionalPanel();
+      }
+      this.clearConditionalPanel();
+    }
+
     const { selectedLocation, line, editor } = props;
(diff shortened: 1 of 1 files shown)


jasonLaster:
Nice. Looks good!

darkwing:
I rebased, let's see what CI says now

darkwing:
Everything checks out!  Nice work @mmcote !

mmcote:
Thanks @darkwing !

PR #5485, opened 2018-02-21
Use original source location to automatically stepOver locations with useless mappings.

loganfsmyth:
### Summary of Changes

* Add parser utility to detect useless mappings
* Check for useless or repeated positions when pausing.

There's probably more than can be done, but this is a self-contained first step.

original-stepping


jasonLaster:
I pushed some opinionated refactorings onto this branch that are also not tested. Feel free to accept/reject any of it. I shared it this way because it was easier for me to think in a commit than a review

src/actions/pause/paused.js
```diff
@@ -4,19 +4,27 @@
...
+  getPreviousPauseFrameLocation
 } from "../../selectors";
+
 import { mapFrames } from ".";
 import { removeBreakpoint } from "../breakpoints";
 import { evaluateExpressions } from "../expressions";
 import { selectLocation } from "../sources";
 import { togglePaneCollapse } from "../ui";
+import { command } from "./commands";
+// NOTE: we might want to move utils/pause.js into utils/pause/index... so we can replace the `pause/index` with `pause`
```
loganfsmyth: Agreed, having to call out `/index` isn't good.

src/reducers/pause.js
```diff
@@ -190,10 +196,18 @@ function update(
...
-      return action.status === "start"
-        ? { ...state, ...emptyPauseState, command: action.command }
-        : { ...state, command: "" };
+    case "COMMAND": {
+      if (action.status === "start") {
+        return {
+          ...state,
+          ...emptyPauseState,
+          command: action.command,
+          previousLocation: _prev(state, action)
```
loganfsmyth: Nice.

jasonLaster: oh _prev was a bad name :P

Issue #5482, opened 2018-02-20
[Preview] Flow types should be not previewed


jasonLaster:
We currently let the user evaluate flow types. They're always null because there's nothing in the code. We can avoid that by excluding them from the symbols identifiers.

![](http://g.recordit.co/MLgF4vyEyv.gif)

here are some places to dig:

```diff
diff --git a/src/actions/preview.js b/src/actions/preview.js
index 8f911cd..3575141 100644
--- a/src/actions/preview.js
+++ b/src/actions/preview.js
@@ -120,6 +120,7 @@ export function updatePreview(target: HTMLElement, editor: any) {
     const symbols = getSymbols(getState(), source.toJS());
 
     let match;
+    // why does it think the flow type is a good thing to preview?
     if (!symbols || symbols.identifiers) {
       match = findBestMatchExpression(symbols, tokenPos, tokenText);
     } else {
diff --git a/src/workers/parser/tests/fixtures/flow.js b/src/workers/parser/tests/fixtures/flow.js
index 0294f19..d5d1875 100644
--- a/src/workers/parser/tests/fixtures/flow.js
+++ b/src/workers/parser/tests/fixtures/flow.js
@@ -1,4 +1,6 @@
+
 class App extends Component {
+                    /* here */
   renderHello(name: string) {
     return `howdy ${name}`;
   }
diff --git a/src/workers/parser/tests/getSymbols.spec.js b/src/workers/parser/tests/getSymbols.spec.js
index 621c18a..1b4b88f 100644
--- a/src/workers/parser/tests/getSymbols.spec.js
+++ b/src/workers/parser/tests/getSymbols.spec.js
@@ -31,7 +31,7 @@ cases(
     },
     { name: "component", file: "component", original: true },
     { name: "react component", file: "frameworks/component", original: true },
-    { name: "flow", file: "flow", original: true },
+    { name: "flow", file: "flow", original: true, only: true },
     { name: "jsx", file: "jsx", original: true },
     { name: "destruct", file: "destructuring" }
   ]
```

jasonLaster:
I thought of a couple of other cases where we don't want to preview:

- class functions keys:
- keys for objects inside of call expressions

![](http://g.recordit.co/u2fxCcEkys.gif)

Here is an [example](https://dbg-weird-preview.glitch.me/)

```js

class Parret {
  
  flyAway() {
    // fly();
  }
  
  speak(words) {
    console.log(german(words, { enlarge: true, umlaut: true }))
[...]

PR #5486, opened 2018-02-21
Exclude flow annotations from symbol identifiers

AnshulMalik:
Associated Issue: #5482 

### Summary of Changes

* Exclude identifier from symbols if it's a type annotation

### Test Plan

- [x] Updated `getSymbols` test to incorporate this change

a

### DIFF
--- a/src/workers/parser/tests/__snapshots__/getSymbols.spec.js.snap
+++ b/src/workers/parser/tests/__snapshots__/getSymbols.spec.js.snap
@@ -70,6 +70,7 @@ identifiers:
 [(23, 6), (23, 11)]  Ultra Ultra
 [(25, 4), (25, 8)] [(25, 4), (25, 8)] this this
 [(25, 9), (25, 16)]  awesome awesome
+[(28, 12), (28, 18)]  person person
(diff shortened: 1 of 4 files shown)


src/workers/parser/tests/__snapshots__/getSymbols.spec.js.snap
```diff
@@ -755,10 +755,13 @@ hasJsx: false"
 
 exports[`Parser.getSymbols flow 1`] = `
 "functions:
-[(2, 2), (4, 3)]   renderHello(name) App
+[(2, 2), (4, 3)]   renderHello(name, action, ) App
 
 variables:
 [(2, 14), (2, 26)]   name
+[(2, 28), (2, 47)]   action
+[(2, 49), (2, 65)]
```
jasonLaster: should we exclude this?

jasonLaster: Okay, so i think this is a different bug with object keys `foo( { todos }: Props) {...}`

lets move forward and fix this up later

Issue #5301, opened 2018-02-02
[Sources] Debugger is blank when breaking on a debugger statement


gregtatum:
I don't have an STR other than setting a debugger statement in perf.html.

<img width="684" alt="image" src="https://user-images.githubusercontent.com/1588648/35753794-afe93df2-0826-11e8-8fba-c3b180d8723e.png">

The error in the browser toolbox is:

<img style='border: 1px solid #000' width="981" alt="image" src="https://user-images.githubusercontent.com/1588648/35753811-c811defc-0826-11e8-980d-3aadbe50c238.png">

---

<img width="688" alt="image" src="https://user-images.githubusercontent.com/1588648/35753821-d272f19c-0826-11e8-9fe3-2bc8ea8fb950.png">

---

<img width="726" alt="image" src="https://user-images.githubusercontent.com/1588648/35753834-def286f8-0826-11e8-8288-79ac560b0b7a.png">

---

<img width="683" alt="image" src="https://user-images.githubusercontent.com/1588648/35753855-f182af3c-0826-11e8-8168-84b9752ea3a0.png">

gregtatum:
Just did a quick search, as Flow is pretty good at catching errors like the `isLoaded()` one, where source is undefined. It looks like only about 7 out of 15 calls to it had Flow coverage.

codehag:
thanks for the issue greg!

codehag:
I think i have an idea why this is happening

jasonLaster:
I think this was also an issue: https://github.com/devtools-html/debugger.html/issues/5509

codehag:
Hey @gregtatum !

We made a change to hopefully counteract what you were saying. I will close this since we do not have an str and there is no way to be certain that its fixed. If you see it again and can reproduce, please reopen

gregtatum:
Thanks, will do!

jasonLaster:
#5490 was reverted

jasonLaster:
This should be easy to bring back. The issue is with the new mochitest

PR #5490, opened 2018-02-21
try to address empty debugger bug by keeping track of ongoing work

codehag:
trys to solve #5301


jasonLaster:
Yulia and I spoke offline. This is R+ with a mochitest that exercises the bug.

PR #5501, opened 2018-02-22
Provide a method for Alphabetizing the outline functions and classes

darkwing:
I find the Outline pane much easier to use when the functions are in alphabetical order.  This PR provides a place for discussion about a feature for alphabetizing and restoring original order:

outlinealphabetize

## ToDo:

- [x] Use variable (not hardcoded) height for the Outline footer and its styles
- [x] Add a "toggled" UI state for the "Az" button so it's clear whether the current state is set to alphabetize or not
- [x] Sort classes as well
- [x] Make the alphabetize setting persist (?) so that moving between sources keeps the alphabetized state
- [x] Tests

### DIFF
--- a/src/components/PrimaryPanes/Outline.css
+++ b/src/components/PrimaryPanes/Outline.css
@@ -3,7 +3,12 @@
  * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
 
 .outline {
-  overflow-y: auto;
+  overflow-y: hidden;
+}
+
+.outline > div {
+  width: 100%;
+  position: relative;
 }
 
 .outline .outline-pane-info {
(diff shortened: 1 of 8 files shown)


src/components/PrimaryPanes/Outline.css
```diff
@@ -70,3 +70,19 @@
...
+  position: absolute;
+  display: flex;
+  bottom: 0;
+  left: 0;
+  right: 0;
+  opacity: 1;
+  z-index: 1;
+  -moz-user-select: none;
+  user-select: none;
+  height: 24px /* var(--editor-footer-height) */;
```
darkwing: `--editor-footer-height` needs to be globally available; should also look at how the SecondaryPane handles the footer height, and possibly consolidate

src/components/PrimaryPanes/Outline.js
```diff
@@ -89,22 +99,44 @@ export class Outline extends Component<Props> {
     );
   }
 
+  toggleAlphabetize() {
```
jasonLaster: this can use the class bind notation

```js
  toggleAlphabetize = () => {
      this.setState({ alphabetize: !this.state.alphabetize });
  }

```

darkwing:
For some reason the WIP check doesn't leave after it's added...

jasonLaster:
Love it!

Issue #5266, opened 2018-01-31
[Outline] Should be disabled when there isn’t a selected source


nchevobbe:
if there are no source selected, we should't be able to select the outline tab, or at least, we should have a better message displayed.
At the moment, it says: "No functions", it could be "No file selected" or something meaningful

jasonLaster:
I think the issue title is misleading here. I like nicolas' suggestion that we change the message to:
` "No file selected" `

nyrosmith:
```diff
diff --git a/assets/panel/debugger.properties b/assets/panel/debugger.properties
index 26d795d..1ce238c 100644
--- a/assets/panel/debugger.properties
+++ b/assets/panel/debugger.properties
@@ -533,6 +533,9 @@ outline.header=Outline
 # LOCALIZATION NOTE (outline.noFunctions): Outline text when there are no functions to display
 outline.noFunctions=No functions
 
+# LOCALIZATION NOTE (outline.noFileSelected): Outline text when there are no functions to display
+outline.noFileSelected=No file selected
+
 # LOCALIZATION NOTE (sources.search): Sources left sidebar prompt
 # e.g. Cmd+P to search. On a mac, we use the command unicode character.
 # On windows, it's ctrl.
diff --git a/src/components/PrimaryPanes/Outline.js b/src/components/PrimaryPanes/Outline.js
index 8d1fe74..8926750 100644
--- a/src/components/PrimaryPanes/Outline.js
+++ b/src/components/PrimaryPanes/Outline.js
@@ -39,11 +39,10 @@ export class Outline extends Component<Props> {
   }
 
   renderPlaceholder() {
-    return (
-      <div className="outline-pane-info">
-        {L10N.getStr("outline.noFunctions")}
-      </div>
-    );
+    const displayMessage = !this.props.selectedSource
+      ? L10N.getStr("outline.noFileSelected")
+      : L10N.getStr("outline.noFunctions");
+    return <div className="outline-pane-info">{displayMessage}</div>;
   }
 
   renderFunction(func: SymbolDeclaration) {
```

sakshivij:
Hi! Can I work on this issue?

AnshulMalik:
Yep @sakshivij, go ahead :) And feel free to ask us on slack if you need any help

PR #5506, opened 2018-02-22
Display message to indicate no selected source in Outline tab (#5266)

Bilie:
Associated Issue: #5266 

### Summary of Changes

'No file selected' message is displayed in the Outline tab when there is no source selected.

### DIFF
--- a/assets/panel/debugger.properties
+++ b/assets/panel/debugger.properties
@@ -555,6 +555,9 @@ outline.header=Outline
 # LOCALIZATION NOTE (outline.noFunctions): Outline text when there are no functions to display
 outline.noFunctions=No functions
 
+# LOCALIZATION NOTE (outline.noFileSelected): Outline text when there are no functions to display
+outline.noFileSelected=No file selected
+
 # LOCALIZATION NOTE (sources.search): Sources left sidebar prompt
 # e.g. Cmd+P to search. On a mac, we use the command unicode character.
 # On windows, it's ctrl.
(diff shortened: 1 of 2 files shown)


assets/panel/debugger.properties
```diff
@@ -555,6 +555,9 @@ outline.header=Outline
 # LOCALIZATION NOTE (outline.noFunctions): Outline text when there are no functions to display
 outline.noFunctions=No functions
 
+# LOCALIZATION NOTE (outline.noFileSelected): Outline text when there are no functions to display
```
juliandescottes: The comment is wrong "when there are no functions to display" should be "Outline text when there are no files selected"

assets/panel/debugger.properties
```diff
@@ -555,6 +555,9 @@ outline.header=Outline
...
+outline.noFileSelected=No file selected
```
flodolo: Should this be "… no file**s**"?

juliandescottes: Since we can only ever select one file I'd keep the singular here? Same as in https://searchfox.org/mozilla-central/source/dom/locales/en-US/chrome/layout/HtmlForm.properties#17-22

juliandescottes:
Thanks for addressing my comment @Bilie 
Flagging @darkwing for reviewing the code change.

assets/panel/debugger.properties
flodolo: I'm not a native speaker, but I usually use a plural after "no". I guess both singular and plural work though.

juliandescottes: Yes I was unsure about that too. googling + looking through the codebase did not really help getting a definitive answer, so I think leaving as is is fine.

darkwing:
Awesome work @Bilie !  I made a tiny change to use the positive conditional check but other than that it's perfect! 👍 💯

PR #5511, opened 2018-02-23
Highlight original source for sources tree

darkwing:
More updates are needed to make the source tree behave but this should be fairly easy:  don't try to highlight pretty-printed `:formatted` addresses that don't exist in the tree.  Instead highlight its original counterpart.

### DIFF
--- a/src/components/PrimaryPanes/SourcesTree.js
+++ b/src/components/PrimaryPanes/SourcesTree.js
@@ -42,6 +42,7 @@ import {
   getExtension
 } from "../../utils/sources-tree";
 
+import { getRawSourceURL } from "../../utils/source";
 import { copyToTheClipboard } from "../../utils/clipboard";
 import { features } from "../../utils/prefs";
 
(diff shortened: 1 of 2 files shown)


src/components/PrimaryPanes/SourcesTree.js
```diff
@@ -131,10 +132,11 @@ class SourcesTree extends Component<Props, State> {
       nextProps.selectedSource &&
       nextProps.selectedSource != selectedSource
     ) {
-      const highlightItems = getDirectories(
-        nextProps.selectedSource.get("url"),
-        sourceTree
-      );
+      let url = nextProps.selectedSource.get("url");
+      if (url) {
+        url = getRawSourceURL(url);
```
darkwing: Some `url`'s are `null`, so this `if` check needs to be done.

AnshulMalik: Should we add a comment here? it might be hard to figure out why this raw source was required.

jasonLaster: hmm, what about making getRawSourceURL safe and return null if there is no URL?

PR #5518, opened 2018-02-23
[CallStack] Simplify call stacks involving Babel async transforms

calebstdenis:
Associated Issue: #5150 

### Summary of Changes
* Annotate frames with certain displayNames or URLs as being Babel frames, so that frames related to async transforms get grouped together.

### Test Plan
1. Go to http://firefox-dev.tools/debugger-examples/examples/babel/
2. Set breakpoints in `entry.js` lines 28 and 31.
3. Hit the `asyncExample` button.
4. On either breakpoint, frames related to async transforms should be grouped together as being Babel frames. (see gif below)

Regression test:
1. Go to http://firefox-dev.tools/debugger-examples/examples/my-app/build/
2. Set a breakpoint on App.js#14
3. Grouping for React and Webpack frames should work like before.

### Screenshots/Videos
babelcallstack


src/utils/frame.js
```diff
@@ -146,12 +150,19 @@ const displayNameMap = {
...
+    (displayNameMap[library] && displayNameMap[library][displayName]) ||
+    displayName
+  );
 }
 
-export function annotateFrame(frame: Frame) {
+export const annotateFrames = flow(
+  map(annotateFrame),
+  annotateBabelAsyncFrames
+);
```
jasonLaster: wow, this is a really elegant solution.

- I've never seen `flow` before!
- I like how this *just* solves it by annotating!

src/utils/frame.js
```diff
@@ -123,6 +124,9 @@ export function getLibraryFromUrl(frame: Frame) {
 }
 
 const displayNameMap = {
+  Babel: {
+    tryCatch: "Async"
```
jasonLaster: So... This is definetly the best way to do it given that we just match on the first frame's display name. 

I think this is fine to ship, but we might need another way in the future to check all the frames display names and decide based on another name like asyntToGenerator

src/utils/frame.js
calebstdenis: Thank you! It took some trial and error to realize, "hey wait a minute, grouping already works fine. why change that?"

src/utils/frame.js
calebstdenis: Agreed.

src/utils/frame.js
jasonLaster: :thumbsup:

jasonLaster:
The only thing I see at the moment, is that the "B" has to be babel's shade of yellow

src/utils/frame.js
```diff
@@ -160,6 +171,35 @@ export function annotateFrame(frame: Frame) {
...
 
+function annotateBabelAsyncFrames(frames: Frame[]) {
+  const firstIndex = findIndex(
+    frames,
+    frame =>
+      getFrameUrl(frame).match(/regenerator-runtime/i) &&
+      frame.displayName === "tryCatch"
+  );
+
+  const lastIndex = findIndex(
```
jasonLaster: lets have a helper for this: 

```js
function getFrameIndex(frames, url, displayName) {
}
```

jasonLaster: can `getFrameUrl(frame)` return null? if so, we should have a check to see if the URL is null before calling match

src/components/SecondaryPanes/Frames/Group.js
```diff
@@ -14,12 +14,11 @@ import "./Group.css";
...
 
 import type { LocalFrame } from "./types";
-import type { Frame } from "../../../types";
 import Badge from "../../shared/Badge";
 
-type FrameLocationProps = { frame: Frame };
+type FrameLocationProps = { frame: LocalFrame };
 function FrameLocation({ frame }: FrameLocationProps) {
-  const library = getLibraryFromUrl(frame);
+  const library = frame.library || getLibraryFromUrl(frame);
```
jasonLaster: why does this have to change?

src/utils/frame.js
```diff
@@ -160,6 +171,35 @@ export function annotateFrame(frame: Frame) {
   return frame;
 }
 
+function annotateBabelAsyncFrames(frames: Frame[]) {
+  const firstIndex = findIndex(
```
loganfsmyth: I can't quite tell how the grouping here is done. Is the approach of firstIndex and lastIndex going to have problems if one async function is called inside of another one, so you potentially have multiple `tryCatch`s in the frame list?

calebstdenis:
@jasonLaster The babel yellow seems a little low-contrast to me on light theme. What do you think?
image


On dark theme, the icon colors are overridden either way
image

jasonLaster:
Lets use it and then consult @violasong.

src/utils/frame.js
jasonLaster: @calebstdenis do you know how to test this with babel...
The main thing is to add this to a file and then run babel to create the output file


```js
async function first() {
  debugger;
}

async function second() {
  await first()
}

second
```

src/components/SecondaryPanes/Frames/Group.js
calebstdenis: It needs to change because now figuring out the library is a little more complicated than just looking at the url. The `getLibraryFromUrl` call doesn't actually need to be there at all, because the selector is already doing the work of figuring out the library. I was being defensive by keeping it in with an `||` (perhaps overly so.)

src/utils/frame.js
calebstdenis: Will do.

calebstdenis: The flowtype says sourceUrl is optional. Will add a null check.

src/utils/frame.js
calebstdenis: @loganfsmyth You're right. Thanks for catching this!
image

src/utils/frame.js
calebstdenis: nevermind. getFrameUrl does a null check

src/utils/frame.js
calebstdenis: re: Flow. Careful! I'll do more FP if you encourage me. Not everyone likes it when I do that :P

src/utils/frame.js
calebstdenis: @loganfsmyth Commit 1fac0ea980c493e063df60729d441f23ac9bd2e9 should address this. Also, the code is hopefully more self-explanatory now. 

image

jasonLaster:
@calebstdenis can you share some new screenshots?

src/components/SecondaryPanes/Frames/Group.js
jasonLaster: could we get away w/ just `getLibraryFromUrl`

src/utils/frame.js
```diff
@@ -160,6 +171,48 @@ export function annotateFrame(frame: Frame) {
...
+  const startIndexes = frames.reduce(
+    (accumulator, frame, index) =>
+      isStartOfBabelGroup(frame) ? [...accumulator, index] : accumulator,
+    []
+  );
+  const endIndexes = frames.reduce(
+    (accumulator, frame, index) =>
+      isEndOfBabelGroup(frame) ? [...accumulator, index] : accumulator,
+    []
+  );
```
jasonLaster: can we generalize this so we have a new `getIndices` function that takes frames, and a predicate?

calebstdenis: absolutely.

calebstdenis:
New screenshot
image

jasonLaster:
The screenshot looks good, but without mapping _callee$ and _calle2$ it is still really hard to read

src/components/SecondaryPanes/Frames/Group.js
calebstdenis: We cannot because getting the library is no longer as simple as getting the library from the URL. However, I'm confident we can get away with `frame.library`.

src/utils/frame.js
```diff
@@ -160,6 +170,49 @@ export function annotateFrame(frame: Frame) {
...
+      (url.match(/_microtask/i) && displayName === "flush")
+  );
+
+  if (startIndexes.length != endIndexes.length || startIndexes.length === 0) {
+    return frames;
+  }
+
+  return flatMap(zip(startIndexes, endIndexes), ([start, end]) =>
+    range(start, end + 1)
+  );
```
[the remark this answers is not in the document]
jasonLaster: hmm, perhaps `start` should be `startIndex`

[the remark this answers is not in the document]
jasonLaster: but yeah, it is hard to see how `[[1,3], [5,8]]` might expand to `[[1,2,3], [5,6,7,8]]` but it is also hard to see how it would fail, once you understand what it does.

There are no obvious bugs, and once you study it, it is obviously correct... :wink:

PR #5522, opened 2018-02-24
Ignore node_modules in flow

jasonLaster:
### Summary of Changes

Follows perf.html's lead - https://github.com/devtools-html/perf.html/pull/810/files
This should help the memory and local perf usage for the debugger


jasonLaster:
closing as we'll need to update libdefs

PR #5524, opened 2018-02-24
Fix re-evaluate on resume

### LINKED ISSUE #5509: [Expressions] Debugger.Frame is not live error from evalWithDebugger
Here is a slightly updated STR, it turns out this requires watch expressions. I reproduce this in Firefox 58-60.

STR:
 * Open up attachment debugger-frame-bug.html in Nightly
 * Open up the DevTools' Debugger
 * Add the watch expression "asdf" (any expression really)
 * Refresh and catch on the breakpoint.
 * Comment out the "debugger" statement from the code in an editor.
 * Refresh the page in Nightly.

Results:

evaluateJS: Error: Debugger.Frame is not live
```
Stack:
  evalWithDebugger@resource://devtools/server/actors/webconsole.js:1397:11
  onEvaluateJS@resource://devtools/server/actors/webconsole.js:949:20
  onPacket@resource://devtools/server/main.js:1767:15
  receiveMessage@resource://devtools/shared/transport/transport.js:760:7
```
https://bugzilla.mozilla.org/show_bug.cgi?id=1440472#c1

jasonLaster:
Fixes #5509

### Summary of Changes

The issue was that we were only clearing the pause state when the debugger would fire a command like step or resume. Another time we resume is when we navigate. In this case there is a navigate event and resumed event. We should also make sure we clear the pause state when we were resumed by the server.

### DIFF
--- a/src/actions/pause/resumed.js
+++ b/src/actions/pause/resumed.js
@@ -20,13 +20,12 @@ export function resumed() {
   return async ({ dispatch, client, getState }: ThunkArgs) => {
     const why = getPauseReason(getState());
     const wasPausedInEval = inDebuggerEval(why);
+    const wasStepping = isStepping(getState());
 
-    if (!isStepping(getState()) && !wasPausedInEval) {
+    dispatch({ type: "RESUME" });
+
+    if (!wasStepping && !wasPausedInEval) {
       await dispatch(evaluateExpressions());
     }
-
-    dispatch({
-      type: "RESUME"
-    });
   };
 }
--- a/src/reducers/pause.js
+++ b/src/reducers/pause.js
@@ -69,7 +69,6 @@ export const createPauseState = (): PauseState => ({
 });
 
 const emptyPauseState = {
-  pause: null,
   frames: null,
   frameScopes: {
     generated: {},
@@ -77,7 +76,8 @@ const emptyPauseState = {
   },
   selectedFrameId: null,
   loadedObjects: {},
-  previousLocation: null
+  previousLocation: null,
+  why: null
 };
 
 function update(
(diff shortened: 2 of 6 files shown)


src/actions/pause/resumed.js
```diff
@@ -20,13 +20,12 @@ export function resumed() {
   return async ({ dispatch, client, getState }: ThunkArgs) => {
     const why = getPauseReason(getState());
     const wasPausedInEval = inDebuggerEval(why);
+    const wasStepping = isStepping(getState());
 
-    if (!isStepping(getState()) && !wasPausedInEval) {
+    dispatch({ type: "RESUME" });
+
+    if (!wasStepping && !wasPausedInEval) {
```
jasonLaster: we should evaluate the expressions after the state is updated from the resume dispactch.

src/reducers/pause.js
```diff
@@ -69,15 +69,15 @@ export const createPauseState = (): PauseState => ({
 });
 
 const emptyPauseState = {
-  pause: null,
```
jasonLaster: this is no longer used

src/reducers/pause.js
```diff
@@ -69,15 +69,15 @@ export const createPauseState = (): PauseState => ({
...
   frames: null,
   frameScopes: {
     generated: {},
     original: {}
   },
   selectedFrameId: null,
   loadedObjects: {},
-  previousLocation: null
+  previousLocation: null,
+  why: null
```
jasonLaster: this is part of the pause state

src/reducers/pause.js
```diff
@@ -210,7 +210,7 @@ function update(
     case "RESUME":
       // We clear why on resume because we need it to decide if
       // we shoul re-evaluate watch expressions.
-      return { ...state, why: null };
+      return { ...state, ...emptyPauseState };
```
jasonLaster: why is in the empty pause state
and this is the fix

Issue #5526, opened 2018-02-24
[Parser] Make parser recognize Angular 1


lukaszsobek:
With the recent addition of `getFramework` we have the basis for making the parser recognize various frameworks. The tracking issue is here #4777 

This issue is for enabling the parser to see Angular files.

lukaszsobek:
Merged. Closing this issue. Noted outstanding parts in tracking issue to be addressed in a future pr.

PR #5527, opened 2018-02-24
[WIP] Makes parser recognize Angular

lukaszsobek:
Associated Issue: #5526

- Adds logic to test against Angular files
- Adds tests for angular 1 and undefined
- Adds & renames fixture files

- [ ] Add recognizing later versions of angular


nyrosmith:
I don't think this will work for ng2 +

lukaszsobek:
@nyrosmith I think we have to write different tests for different versions of angular then. Right we seem to have 2,4 and 5 ...

jasonLaster:
do we want to add support to other versions here or another PR?

lukaszsobek:
@jasonLaster as we discussed adding the next versions is dependent on being able to parse decorators, which is a whole other issue. I think wrapping up this PR and introducing Angular 1 support is beneficial over waiting for another issue to be implemented only to be able to expand this issue.

Issue #5491, opened 2018-02-21
[Preview] should ignore some labels


jasonLaster:
There are some cases where Preview could ignore some identifiers because we know that they'll be undefined.

- class functions keys: e.g. `flyAway`, `speak`
- keys for objects inside of call expressions e.g. `umlaut`

![](http://g.recordit.co/u2fxCcEkys.gif)

Here is an [example](https://dbg-weird-preview.glitch.me/)

```js

class Parret {
  
  flyAway() {
    // fly();
  }
  
  speak(words) {
    console.log(german(words, { enlarge: true, umlaut: true }))
  }
}
```

These changes will definitely be made in the parser symbols module

jasonLaster:
```diff
diff --git a/.flowconfig b/.flowconfig
index 170cb25..42ebf16 100644
--- a/.flowconfig
+++ b/.flowconfig
@@ -1,8 +1,5 @@
 [ignore]
-.*node_modules/documentation.*
-.*node_modules/devtools-reps/src/object-inspector/index.js
-.*node_modules/devtools-mc-assets/.*
-.*node_modules/immutable/dist/immutable.js.flow
+.*node_modules/.*
 <PROJECT_ROOT>/firefox/.*
 .*mozilla-central/.*
 .*flow-coverage-report/.*
diff --git a/src/workers/parser/getSymbols.js b/src/workers/parser/getSymbols.js
index c83265f..815df43 100644
--- a/src/workers/parser/getSymbols.js
+++ b/src/workers/parser/getSymbols.js
@@ -214,6 +214,11 @@ function extractSymbol(path, symbols) {
   if (t.isIdentifier(path)) {
     let { start, end } = path.node.loc;
 
+    // https://astexplorer.net/#/gist/80223c44df0de43d9a991df1f2afe89a/4c628d018ba2af93afff2adfaff18dca3b2431c4
+    if (t.isProperty(path.parent)) {
+      return;
+    }
+
     if (path.node.typeAnnotation) {
       const column = path.node.typeAnnotation.loc.start.column;
       end = { ...end, column };
diff --git a/src/workers/parser/tests/fixtures/es6.js b/src/workers/parser/tests/fixtures/es6.js
index 90b5314..899b5a3 100644
--- a/src/workers/parser/tests/fixtures/es6.js
+++ b/src/workers/parser/tests/fixtures/es6.js
@@ -1 +1,3 @@
-dispatch({ ...action, [PROMISE]: promise });
[...]

PR #5531, opened 2018-02-25
[Preview] should ignore some labels

atwalg2:
Associated Issue: #5491

### Summary of Changes

- [x]  add case for class functions keys: e.g. flyAway, speak
- [x]  add case handling keys for objects inside of call expressions e.g. umlaut

### Test Plan

Will be adding
- [ ] test case for class functions keys
- [ ] test case for objects inside of call expressions

Before
before
After
screencapafter

### DIFF
--- a/src/workers/parser/getSymbols.js
+++ b/src/workers/parser/getSymbols.js
@@ -214,6 +214,14 @@ function extractSymbol(path, symbols) {
   if (t.isIdentifier(path)) {
     let { start, end } = path.node.loc;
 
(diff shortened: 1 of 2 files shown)


jasonLaster:
I like it! Just needs some unit tests :)

jasonLaster:
The CI issue is jest complaining the previous snapshots changed, which is a good thing.

Perhaps just running `jest -u` is enough because we might have coverage elsewhere

PR #5535, opened 2018-02-26
fix tabs using the selected source

bomsy:
Associated Issue: #4773

Here's the Pull Request Doc

### Summary of Changes

* remove `sourceMetaData` which uses the selectedSource
* added a `getMetaData` function which would take the current source id

### Test Plan

Example test plan:

- [x] Select react tab
- [x] Switch between other tabs and react tab

### DIFF
--- a/src/components/Editor/Tab.js
+++ b/src/components/Editor/Tab.js
@@ -47,7 +47,7 @@ type Props = {
   showSource: string => void,
   source: SourceRecord,
   activeSearch: string,
-  sourceMetaData: string => any
+  getMetaData: string => any
 };
 
 class Tab extends PureComponent<Props> {
@@ -144,7 +144,7 @@ class Tab extends PureComponent<Props> {
       selectSource,
       closeTab,
       source,
-      sourceMetaData
+      getMetaData
     } = this.props;
     const src = source.toJS();
     const filename = getFilename(src);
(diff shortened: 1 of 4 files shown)


src/components/Editor/Tab.js
```diff
@@ -202,10 +202,7 @@ export default connect(
     return {
       tabSources: getSourcesForTabs(state),
       selectedSource: selectedSource,
-      sourceMetaData: getSourceMetaData(
-        state,
-        selectedSource && selectedSource.get("id")
-      ),
+      getMetaData: sourceId => getSourceMetaData(state, sourceId),
```
jasonLaster: so the problem with this approach is that it is not memoized

I think we need a new selector

```js
export const getSelectedSourceMetaData = ...
```

bomsy: Do we want the meta data of the selected source?

src/components/Editor/Tab.js
```diff
@@ -197,15 +198,13 @@ class Tab extends PureComponent<Props> {
   }
 }
 export default connect(
-  state => {
+  (state, ownProps) => {
```
lukaszsobek: I usually just call them `props`.

src/components/Editor/Tab.js
```diff
@@ -197,15 +198,13 @@ class Tab extends PureComponent<Props> {
...
     const selectedSource = getSelectedSource(state);
+    const { source } = ownProps;
     return {
       tabSources: getSourcesForTabs(state),
       selectedSource: selectedSource,
-      sourceMetaData: getSourceMetaData(
-        state,
-        selectedSource && selectedSource.get("id")
-      ),
+      sourceMetaData: getSourceMetaData(state, source.get("id")),
```
lukaszsobek: This changes it from using the root url to the actual document that is displayed, right?

src/components/Editor/Tab.js
```diff
@@ -14,6 +14,7 @@ import CloseButton from "../shared/Button/Close";
 
 import type { List } from "immutable";
 import type { SourceRecord } from "../../reducers/sources";
+import type { SourceMetaDataType } from "../../reducers/ast";
```
lukaszsobek: Maybe types should have their own folder instead of being together with reducers  ... but that's for another issue I guess.

Issue #5528, opened 2018-02-25
[ProjectSearch] should show the result count


jasonLaster:
Project search (`cmd+shift+f`) should show the number of found items. 

I believe this PR #5387, might of unintentionally removed that functionality.

lukaszsobek:
Merged. Closing.

PR #5536, opened 2018-02-26
Restores result count for Project Search

lukaszsobek:
Associated Issue: #5528

- Restores result count for project search
- Removes unused code after removal of count from search bar and QOM.

Vid:
result_count


wldcordeiro:
Looks good but I think SearchBar should retain the result count as well. It's a common pattern for in-file search to show results. The only one with no result count should be QuickOpen.

lukaszsobek:
@wldcordeiro Sure thing.

PR #5537, opened 2018-02-26
Pause points

jasonLaster:
### Summary of Changes
It would be nice to have a notion of good places to pause in the debugger.
We can use this data set for 2 use cases:

1. allow breakpoints to be added on a line
2. tell the server where we are cool with it stopping

If there is a program

```js
const arr = [
  '1',
  2,
  foo()
]

const obj = {
  a: '1',
  b: 2,
  c: foo(),
}

debugger; debugger;
console.log("a"); console.log("a");


foo(1, '2', bar())

```

The tests will output this 

```
    <1>const arr = [
      '1',
      2,
      <2>foo()
    ]

    <3>const obj = {
      a: '1',
      b: 2,
      c: <4>foo(),
    }

    <5>debugger; <6>debugger;
    <7>console.log("a"); <8>console.log("a");

    <9>foo(1, '2', <10>bar())
```


jasonLaster:
I hacked on this a bit and made some progress

- pause points is now used to calculate empty lines, this is much more accurate than looking at tokens
- pause points now has a types field for breakpoints and stepOver because some times you want to be able to set a breakpoint where you wouldnt other wise want to pause
- i've extended the definition of a pause point to include control flow, functions, imports

I'm evolving my thinking on what a step over really is... when the user is paused at a mutlilne expression like an array, object, or call expression and steps over, I believe they expect to jump over the expression. This means, that a step over would not land at an inner call expression. 
- If the user wants to go to the call expression, they can step in. 
- If they want to go to the second call expression, they can play dolphin and step in, step out...
- I'm not sure about what to do if the user is paused at the first inner expression in an array and steps over... The user would only be able to get there by setting a breakpoint though. e.g. sets a bp at a(), then steps over. 

```js
var x = [
  1,
  a(),
  2, 
  b()
]
```

darkwing:
I love this -- it also seems "faster".  It's nice clicking around from line to line without having the breakpoint marker jump like it did with EmptyLines

Issue #5514, opened 2018-02-23
Highlighted Tree Item is lost when switching between Sources and Outline


darkwing:
Ideally we keep track of the highlighted item:

![losthighlight](https://user-images.githubusercontent.com/46655/36604051-43c1c7fe-1882-11e8-96d6-8957d886ef87.gif)

darkwing:
I started looking at this:

```diff
diff --git a/src/components/PrimaryPanes/SourcesTree.js b/src/components/PrimaryPanes/SourcesTree.js
index 411211019..9fda76010 100644
--- a/src/components/PrimaryPanes/SourcesTree.js
+++ b/src/components/PrimaryPanes/SourcesTree.js
@@ -80,13 +80,22 @@ class SourcesTree extends Component<Props, State> {
 
   constructor(props: Props) {
     super(props);
-    const { debuggeeUrl, sources, projectRoot } = this.props;
+    const { debuggeeUrl, sources, projectRoot, selectedSource } = this.props;
 
-    this.state = createTree({
+    const tree = createTree({
       projectRoot,
       debuggeeUrl,
       sources
     });
+
+    this.state = tree;
+
+    if (selectedSource) {
+      this.state.highlightItems = getDirectories(
+        selectedSource,
+        tree.sourceTree
+      );
+    }
   }
 
   componentWillReceiveProps(nextProps: Props) {
```

When you switch back to the Sources tab, only the `constructor` runs, so focused item / highlighting doesn't happen.

This snippet doesn't fix the issue but it's a start in the right direction!

darkwing:
I got a bit further with this:

```diff
diff --git a/src/components/PrimaryPanes/SourcesTree.js b/src/components/PrimaryPanes/SourcesTree.js
index 752b0f5f6..992befcb2 100644
--- a/src/components/PrimaryPanes/SourcesTree.js
+++ b/src/components/PrimaryPanes/SourcesTree.js
@@ -80,13 +80,23 @@ class SourcesTree extends Component<Props, State> {
 
   constructor(props: Props) {
     super(props);
-    const { debuggeeUrl, sources, projectRoot } = this.props;
+    const { debuggeeUrl, sources, projectRoot, selectedSource } = this.props;
 
-    this.state = createTree({
+    const state = createTree({
       projectRoot,
       debuggeeUrl,
       sources
     });
+
+    if (selectedSource) {
+      const items = getDirectories(selectedSource, state.sourceTree);
+      state.highlightItems = items;
[...]

PR #5538, opened 2018-02-26
Preserve highlighted Tree Item when switching between Sources and Out…

nyrosmith:
…line

Associated Issue: #5514

### DIFF
--- a/src/components/PrimaryPanes/SourcesTree.js
+++ b/src/components/PrimaryPanes/SourcesTree.js
@@ -376,7 +376,6 @@ class SourcesTree extends Component<Props, State> {
       roots = () => sourceTree.contents[0].contents;
     }
 
-    const isEmpty = sourceTree.contents.length === 0;
     const treeProps = {
       autoExpandAll: false,
       autoExpandDepth: expanded ? 0 : 1,
--- a/src/components/PrimaryPanes/SourcesTree.js
+++ b/src/components/PrimaryPanes/SourcesTree.js
@@ -92,6 +92,19 @@ class SourcesTree extends Component<Props, State> {
     });
   }
 
+  componentDidMount() {
+    const { selectedSource } = this.props;
+    if (selectedSource) {
+      const highlightItems = getDirectories(
+        selectedSource.get("url"),
+        this.state.sourceTree
+      );
+
+      // eslint-disable-next-line react/no-did-mount-set-state
+      this.setState({ highlightItems });
+    }
+  }
+
   componentWillReceiveProps(nextProps) {
     const {
       projectRoot,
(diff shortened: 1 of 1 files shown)


lukaszsobek:
Looks good to me - highlight stays in place, any edge cases removing the lines might cause?

src/components/PrimaryPanes/SourcesTree.js
```diff
@@ -91,6 +91,19 @@ class SourcesTree extends Component<Props, State> {
...
+  componentDidMount() {
+    const { selectedSource } = this.props;
+    if (selectedSource) {
+      const highlightItems = getDirectories(
+        selectedSource.get("url"),
+        this.state.sourceTree
+      );
+
+      // eslint-disable-next-line react/no-did-mount-set-state
+      this.setState({ highlightItems });
```
lukaszsobek: Could we move this to the constructor?

jasonLaster:
CC @darkwing

src/components/PrimaryPanes/SourcesTree.js
nyrosmith: That was my first attempt but it didn't work

darkwing: "react/no-did-mount-set-state" is scaring me; let me think on this

darkwing:
@nyrosmith What do you think about moving the highlightedItems to a parent state like I did here:  

https://github.com/devtools-html/debugger.html/pull/5501

That way we don't need the extra renders that `react/no-did-mount-set-state` brings?

nyrosmith:
Good idea. Let's make it so!

darkwing:
@nyrosmith Would you like to update this or should we?  Thanks!

darkwing:
I'm taking this over, will update soon

darkwing:
Hey @nyrosmith !  I spent a while with this issue today and I shared your frustration.  Your solution here is the easy way but I'm not sure it's the best way.

I started from scratch and *almost* got there.  Here's what I did:

https://gist.github.com/darkwing/8edf5a6e7076b735f423386c80ea373f

Essentially I:

- Calculated the `highlightItems` and `focusedItem` in the `render` method, not relying on state, because that state disappears when you switch to the `Outline` tab

- Used the `props.focused` prop within `ManagedTree` to set the initial `focusedItem`

This solved 90% of the problem but, since I was using the currently selected source in the render method, there was a bit of "jank" when expanding and contracting items.

I thought I'd pass this on for you to look at and see if you can cross the finish line :)

darkwing:
Let's close this for now and pick it up when it has more priority.  Thank you for your effort @nyrosmith !

Issue #5534, opened 2018-02-26
[commandbar] the hover background should be flush with the bar


jasonLaster:
We recently converted the command bar buttons from divs to buttons. When we did that the hover background stopped filling the bar. 

This should be an easy CSS Fix.

<img src='https://shipusercontent.com/0f1e6298f4e7ce014236f7a5e0c0f8c1/Screen%20Shot%202018-02-26%20at%2011.57.36%20AM.png' title='Screen Shot 2018-02-26 at 11.57.36 AM.png' width=256>

jasonLaster:
Here is a GIF of what we do with the toggle buttons
![](http://g.recordit.co/goOS750LpX.gif)

PR #5548, opened 2018-02-27
Give command bar buttons full background on hover

berraknil:
Associated Issue: #5534

### Summary of Changes

* Changed the flexbox alignment as it was preventing buttons from reaching full height and has no visible effect on icons being centered
* Also removed the margin-inline-end property as it was separating the buttons from one another on hover, if that is the intended effect while keeping the button height full on hover, I can put it back :)

### Test Plan

- [x] Opened a few projects and hovered and clicked on the command bar icons

### Screenshots/Videos 

full-background-on-hover

### DIFF
--- a/src/components/SecondaryPanes/CommandBar.css
+++ b/src/components/SecondaryPanes/CommandBar.css
@@ -113,7 +113,8 @@ img.resume {
 
 .command-bar .step-position {
   color: var(--theme-comment-alt);
-  margin-inline-end: 1em;
+  padding-top: 8px;
+  margin-inline-end: 4px;
 }
 
 .command-bar .replay-active {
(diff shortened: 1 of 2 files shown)


darkwing:
Thank you for your contribution @berraknil !!!  We're glad to have your help! 

I made an addition to fix the positioning of the step position:

position

berraknil:
Hi @darkwing completely missed that part, as I am new to the debugger :) That looks good but now I am wondering if we should just make all of that line a flexbox and align all the items there (including the step positions) vertically (without squishing them). Haven't looked into the rest of the code yet though, will check and see if it is more than I have time for now. In that case this commit looks good to me!

darkwing:
Awesome @berraknil!  Just provide an update here when you get a moment :)

PR #5551, opened 2018-02-28
Clean up the stepping tests

jasonLaster:
Two things here

1) i organized the test functions so it's easy to run one at a time...
2) i changed one value to get the test to pass! we should check this ✅


src/test/mochitest/browser_dbg-babel-stepping.js
```diff
@@ -104,23 +103,27 @@ add_task(async function() {
...
+// and Babel doesn't map the _loop() call, so we step past it automatically.
+function testStepOveForOfClosure(dbg) {
+  return breakpointSteps(
     dbg,
     "step-over-for-of-closure",
     { line: 6, column: 2 },
     [
       ["stepOver", { line: 8, column: 2 }],
-      ["stepOver", { line: 12, column: 2 }]
+      ["stepOver", { line: 8, column: 2 }]
```
jasonLaster: I changed this because the tests were failing locally. they were failing on CI too, but we weren't catching that...
https://travis-ci.org/devtools-html/debugger.html/branches

loganfsmyth: We should expand this to be
```

      ["stepOver", { line: 8, column: 2 }],
      ["stepOver", { line: 8, column: 2 }],
      ["stepOver", { line: 8, column: 2 }],
      ["stepOver", { line: 12, column: 2 }]
```
so it still steps all the way to line 12.

loganfsmyth:
Looks good other than the one comment.

PR #5573, opened 2018-03-01
Better tab widths & the death of endTruncateStr (3 of 3 endTruncateStr PRs)

CharlesStocky:
A tab's width (if the string's length is 40 or above) is now 40% of `source-tab`'s width. `source-tab`'s width is now 100% (minus the pixels the buttons to right take up) by default, this is what allows this method of tab-sizing to work. `Tab`'s width can easily be lowered from `Tab.js`, it probably should.

`source-tabs` is rendered before a `source-tab` is rendered, so its width can be grabbed before a `source-tab` is rendered.

screen shot 2018-03-01 at 1 13 16 pm

`endTruncateString` is no more, so some things are wonky. I also had a PR that dealt with the effects of removing `endTruncateStr` from breakpoints, but I was made aware the breakpoints are going through redesign anyway, so I'll hold off on that PR.

### DIFF
--- a/src/components/Editor/Tab.js
+++ b/src/components/Editor/Tab.js
@@ -156,6 +156,12 @@ class Tab extends PureComponent<Props> {
     const isPrettyCode = isPretty(source);
     const sourceAnnotation = getSourceAnnotation(source, sourceMetaData);
 
+    const sourceTabsWidth = document.getElementsByClassName("source-tabs");
+    const tabWidth =
+      filename.length < 40
+        ? "auto"
+        : sourceTabsWidth[0].clientWidth * 0.4 + "px";
+
     function onClickClose(ev) {
       ev.stopPropagation();
       closeTab(source.get("url"));
@@ -173,6 +179,7 @@ class Tab extends PureComponent<Props> {
         onClick={() => selectSource(sourceId)}
         onContextMenu={e => this.onTabContextMenu(e, sourceId)}
         title={getFileURL(src)}
+        style={{ width: tabWidth }}
       >
         {sourceAnnotation}
         <div className="filename">{filename}</div>
--- a/src/components/Editor/Tabs.js
+++ b/src/components/Editor/Tabs.js
@@ -139,11 +139,9 @@ class Tabs extends PureComponent<Props, State> {
       return;
     }
 
-    return (
-      <div className="source-tabs" ref="sourceTabs">
-        {tabSources.map((source, index) => <Tab key={index} source={source} />)}
-      </div>
-    );
+    return tabSources.map((source, index) => (
+      <Tab key={index} source={source} />
+    ));
   }
 
   renderDropdown() {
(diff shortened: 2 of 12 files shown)


src/utils/frame.js
```diff
@@ -202,7 +201,7 @@ export function formatDisplayName(
   }
 
   displayName = simplifyDisplayName(displayName);
```
lukaszsobek: Maybe `return simplifyDisplayName(displayName);` would be enough.

src/utils/source.js
```diff
@@ -124,7 +123,7 @@ function resolveFileURL(
 ) {
   url = getRawSourceURL(url || "");
   const name = transformUrl(url);
```
lukaszsobek: Maybe just `return transformUrl(url);`

src/components/SecondaryPanes/Breakpoints.js
```diff
@@ -80,11 +79,7 @@ function renderSourceLocation(source, line, column) {
     return null;
   }
 
-  return (
-    <div className="location">
-      {`${endTruncateStr(filename, 30)}: ${bpLocation}`}
-    </div>
-  );
+  return <div className="location">{`$filename}: ${bpLocation}`}</div>;
```
lukaszsobek: Not really a fan of putting everything in one line as it makes it really difficult to read, besides - we might be missing an opening `{` here.

src/components/Editor/Tabs.js
```diff
@@ -187,7 +185,7 @@ class Tabs extends PureComponent<Props, State> {
     return (
       <div className="source-header">
         {this.renderStartPanelToggleButton()}
-        {this.renderTabs()}
+        <div className="source-tabs">{this.renderTabs()}</div>
```
lukaszsobek: Encapsulated `{}` feel better in separate lines for me as they are easier to absorb when reading.

src/components/Editor/Tab.js
```diff
@@ -156,6 +156,12 @@ class Tab extends PureComponent<Props> {
     const isPrettyCode = isPretty(source);
     const sourceAnnotation = getSourceAnnotation(source, sourceMetaData);
 
+    const sourceTabsWidth = document.getElementsByClassName("source-tabs");
```
lukaszsobek: Might need a change of name as assigning DOM elements to a `width` variable feels difficult to track later.

src/components/Editor/Tab.js
```diff
@@ -156,6 +156,12 @@ class Tab extends PureComponent<Props> {
...
+    const tabWidth =
+      filename.length < 40
+        ? "auto"
+        : sourceTabsWidth[0].clientWidth * 0.4 + "px";
```
lukaszsobek: This might go into a separate variable as it is rather long.

src/components/Editor/Tab.js
```diff
@@ -173,6 +179,7 @@ class Tab extends PureComponent<Props> {
         onClick={() => selectSource(sourceId)}
         onContextMenu={e => this.onTabContextMenu(e, sourceId)}
         title={getFileURL(src)}
+        style={{ width: tabWidth }}
```
lukaszsobek: Love padding curly braces like this as it is really easy to read, but for the sake of consistency it seems we should remove the spaces.

darkwing:
Thanks for contributing this @CharlesStocky!  After talking with our UX person, we're going to reimagine the design, including a grid layout, so I think we'll close these for the time being and re-open when the grid and such is ready.  Thank you!

PR #5579, opened 2018-03-01
Get rid of prop-types

nyrosmith:
Associated Issue: #2804


jasonLaster:
I love this!

nyrosmith:
I need to cleanup the stories. Then we can get rid of the prop-types package at all

jasonLaster:
good point

PR #5592, opened 2018-03-02
[Parser] Stop using babel-traverse

loganfsmyth:
Alternative to the partial implementation in #5557

Instead of trying to rewrite all of getSymbols to handle an array instead of NodePath, I built a wrapper on the array to make it behave like a NodePath. I think on the whole it makes this PR much easier to follow and makes the rewrite less likely to break stuff.

I'm not sure how to update the module manifest like you have in your PR @jasonLaster.


jasonLaster:
In a quick performance test, the new traversal was .8 sec as opposed to 1.7 sec when parsing a 50K bundle.

This is a huge win!

```js
  setSource(source);
    getAst(source.id);
    const t = Date.now();

    getSymbols(source.id);
    getSymbols(source.id);
    getSymbols(source.id);
    getSymbols(source.id);
    getSymbols(source.id);

    const t1 = Date.now();
    console.log(`time ${t1 - t}`);

    // w/ babel/types traverse    time 4113
    // w/ babel-traverse          time 8844
```

PR #5593, opened 2018-03-02
(WIP) Re-enable "try to address empty debugger bug"

jasonLaster:
Associated Issue: #5301

### Summary of Changes

* Re-enables the patch
* changes the test to have new assertions
* adds some blacklisted actions to the log to simplify the output


jasonLaster:
@codehag I'd like to get your thoughts on the test assertions here. Is this what you had in mind?

jasonLaster:
If we want to update the test we can do it as a follow up

PR #5604, opened 2018-03-05
Update timeout in browser_dbg-babel.js (sync m-c bug 1440102)

juliandescottes:
Synchronizing changes from https://bugzilla.mozilla.org/show_bug.cgi?id=1440102 

I haven't changed the browser.ini in the sync since the previous m-c change (skip on win + ccov) had not been synchronized here.


juliandescottes:
It looks like the skip-if is somehow visible on the release branches? 
How are supposed to work in this case? Make PRs against the release branches?

I know diverging from master has been helpful to get releases out while not blocking master but we can't leave commits like this out of master.

Issue #4629, opened 2017-11-08
[bundle] debugger.js is too large


jasonLaster:
The firefox debugger bundle has always been difficult to understand. 

* why is it so many lines?
* what packages make it large?
* why are those packages included?

There are some good tools and we should improve some of the documentation around investigating these questions.

- [x] document best practices
- [x] we should lint for bundles growing too large... 

There are also some known issues

- [x] devtools-map-bindings uses babylon
- [ ] devtools-launchpad is included which is huge
  - [ ] we should switch to using `devtools-contextmenu`
  - [ ] we should have a small renderRoot function, which we can include directly
- [ ] we bundle packages we can externalize
  - [ ] immutable
  - [ ] react-redux
  - [ ] redux 
- [ ] we bundle packages not needed in the bundle
  - [ ] properties files
  - [ ] ...

jasonLaster:
### Some Tools

![Screen Shot 2017-11-08 at 9.42.07 AM.png](https://shipusercontent.com/05ace0ec040dc7af3067cb50b528d717/Screen%20Shot%202017-11-08%20at%209.42.07%20AM.png)


![Screen Shot 2017-11-08 at 9.41.09 AM.png](https://shipusercontent.com/d00336549a3b754be1d6669c1dee2fd7/Screen%20Shot%202017-11-08%20at%209.41.09%20AM.png)

jasonLaster:
Removing launchpad should be a top priority

- [ ] create a new renderRoot
- [ ] separate showMenu

The next steps which should be nice are

- [ ] get rid of immutable
- [ ] externalize assets/panel/properties

http://g.recordit.co/URWKiXqD0E.gif

jasonLaster:
here are some visuals of what it would look like to trim down the bundle

<img src='https://shipusercontent.com/d555561dc53c881f2b8dd14764dc90b9/bundles.jpg' title='bundles.jpg' width=2730>

jasonLaster:
```diff
diff --git a/webpack.config.js b/webpack.config.js
index 89d94b5..f826166 100644
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -65,7 +65,15 @@ function buildConfig(envConfig) {
       "wasmparser/dist/WasmParser": "devtools/client/shared/vendor/WasmParser",
       "wasmparser/dist/WasmDis": "devtools/client/shared/vendor/WasmDis",
       "devtools-connection": "devtools/shared/flags",
[...]

PR #5609, opened 2018-03-05
Improve Bundle Size

jasonLaster:
Associated Issue: #4629 

### Summary of Changes

* Stop using launchpad
* exclude libs

This brings the bundle size from 1.5 mb to 1.2mb. It also means we should see smaller diffs w/ the assets properties

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -60,12 +60,18 @@ function buildConfig(envConfig) {
       "./source-editor": "devtools/client/sourceeditor/editor",
       "./test-flag": "devtools/shared/flags",
       react: "devtools/client/shared/vendor/react",
+      redux: "devtools/client/shared/vendor/redux",
       "react-dom": "devtools/client/shared/vendor/react-dom",
       lodash: "devtools/client/shared/vendor/lodash",
+      immutable: "devtools/client/shared/vendor/immutable",
+      "react-redux": "devtools/client/shared/vendor/react-redux",
+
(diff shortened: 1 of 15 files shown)


webpack.config.js
```diff
@@ -60,12 +60,18 @@ function buildConfig(envConfig) {
...
       react: "devtools/client/shared/vendor/react",
+      redux: "devtools/client/shared/vendor/redux",
       "react-dom": "devtools/client/shared/vendor/react-dom",
       lodash: "devtools/client/shared/vendor/lodash",
+      immutable: "devtools/client/shared/vendor/immutable",
+      "react-redux": "devtools/client/shared/vendor/react-redux",
+
       "wasmparser/dist/WasmParser": "devtools/client/shared/vendor/WasmParser",
       "wasmparser/dist/WasmDis": "devtools/client/shared/vendor/WasmDis",
+      "../assets/panel/debugger.properties": "devtools/shared/flags",
```
juliandescottes: If I understand correctly this redirect works because those dependencies are not really used when isFirefoxPanel is true, and flags is just used randomly because it is one of our externalized dependencies? Could we add a comment here to explain the trick?

juliandescottes: By the way, I'm just thinking that we could both clarify the intent by importing a dedicated external module that throws a clear error if it is required. Something like "forbidden-panel-import.js". It would make it easier to investigate regressions where a require to an excluded lib made its way back to the panel version.

jasonLaster: That's a nice idea. Lets land this and if we regress we can do that fix as a follow up.

PR #5613, opened 2018-03-05
Change PR template for issue clarification

darkwing:
As discussed earlier

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE
+++ b/.github/PULL_REQUEST_TEMPLATE
@@ -1,4 +1,4 @@
-Associated Issue: #<issue number>
+Fixes Issue: #<issue number>
 
 Here's the Pull Request Doc
 https://devtools-html.github.io/debugger.html/CONTRIBUTING.html#pull-requests


.github/PULL_REQUEST_TEMPLATE
```diff
@@ -1,4 +1,4 @@
-Associated Issue: #<issue number>
+Fixes Issue: #<issue number>
```
jasonLaster: so, this is where the language is a bit difficult. 

It'd be great to say something like:

If this PR fixes the issue use (`fixes <#issue number>`) syntax if it is just related to the issue just mention the `<#issue number>`.

This is kinda verbose, not sure if there's a better way to shorten it though

darkwing: I don't think we should include "logic" in the PR template; IMO we should just run with "Fixes" because most of our issues describe a single-purpose, single-problem.  There's a decent argument that our issues are too complicated if we're describing multiple problems. And even if we do have a few of those, most describe just the one issue.  If we use a hook to close issues based on "Fixes #", we can simply open up the more complicated issue because we're likely working on that, as opposed to a contributor.  Thoughts?

jasonLaster: I like it!

Does `Fixes issue: #234` work? or should we just do `Fixes #<issue number>`

darkwing: I like `Fixes Issue: ####` because I'm guessing people will submit with "Fixes issue where ____".  I like pushing people toward a specific *issue*, and adding the text "issue" makes it clear.

jasonLaster: cool. Lets do it

bomsy: Also i like `Fixes issue `. It feels more active. `Associated Issue` was more passive & informative.

This is totally minor and is likely a non-issue, but if the PR fixes more than one issue,
can we use `Fixes issue(s):  #234, #456` or just go as `Fixes issue:  #234, #456`?

Issue #5448, opened 2018-02-16
[SourceTree] it should be easy to clear the root


jasonLaster:
At the moment it can be easy to forget that there is a root set. It is also difficult to figure out how to clear the root.

We should solve that problem by showing something in the UI for clearing the root. We've also discussed breadcrumbs, but because the path can be deep i don't know if there is a good way of showing it...

lukaszsobek:
Maybe an icon with a tooltip next to the set root item and the setting available to be cleared via a rightclick menu - the same way it was set.

jasonLaster:
yeah - i think that's right. @violasong thoughts?

violasong:
Oh interesting, I didn't realize this feature existed! It's nice that you can already right-click to clear it, although instead of "Remove directory root" it should probably say something like "Unset as root".  Instead of an icon (since there's already an icon in the row), maybe it could say "(branch)" next to the name of the item, or something more accurate than that? 

...I feel like there may be an existing pattern for this somewhere out there that could help, but can't think of anything at the moment

or .../loading/

violasong:
Or maybe it could be displayed as:
▷ .../directoryname

darkwing:
I love the `▷ .../directoryname` idea!  `..` is a well known symbol that people would recognize and saves us from needing to add a button and/or message somewhere.  Nice @violasong !

Issue #5432, opened 2018-02-15
[Sources tree] Add UI Element for Reverting to Main Source Tree


darkwing:
An idea from @juliandescottes:

Right now we can set a root directory in the sources tree (sweet!) but there's no UI element which reminds the user they are looking at a piece of the source tree and not the whole thing.  We do have a context menu item to remove the root directory and show the full tree, but the context menu item isn't visible enough.

darkwing:
This is @violasong's mockup:

![image](https://user-images.githubusercontent.com/46655/37120969-80f8c322-2221-11e8-98bb-e471baed9511.png)

We will implement the home button method.

PR #5632, opened 2018-03-07
Provide a clear button for clearing directory root

darkwing:
Fixes Issue: #5448, #5432

A new button appears when the user has set a new directory root:


darkwing:
This PR needs the following updates:

[ ] -  Implement @violasong's new mockup (icon here: https://design.firefox.com/icons/viewer/)
[ ] -  If the root has no sources (i.e. via a page change), we should clear the directory root

jasonLaster:
@darkwing we might not need to clear the root, just not use it to filter the results

darkwing:
screeny

jasonLaster:
- home might be too dark
- home needs some pading
- not sure about that divider, maybe there's a simpler >
- the JS directory is pretty indented. Can we normalize indentation so the top directory is always the same

darkwing:
Per your points:

- [ ] Yes, I agree the home icons should be lighter; I'll mess with opacity there.  @violasong said it would be a bigger, though.

- [ ] The padding issue is one I saw coming, since the tree padding is a bit eradic in the case of subdirs and files below them.  I was a bit worried about this. I added the `margin-inline-start: 10px;` per @violasong's single screenshot, but that only handled a single case.

- [ ] I kinda like the divider

- [ ] The only indentation I touched was the source-list indentation when it's inside a custom root.  What do you think without the 10px margin addition?  Or do we want to re-write all indentation within that bottom pane?  Sounds like....pain :D

- [ ] I'll look into the horizontal scrollbar -- just noticed it.

- [ ] The globe is currently incorrect without this PR; it's simply being added when depth is 0

jasonLaster:
cool, re - tree depth is the tree preserving the depth some how which is affecting padding between js and js/views?

darkwing:
It looks like `.tree-node[data-expandable="false"][aria-level="0"]` is the culprit for why the indent is so massive:

styletoggle

darkwing:
The big blocker here is the scrollbar issue.

src/components/PrimaryPanes/SourcesTree.js
```diff
@@ -193,7 +193,7 @@ class SourcesTree extends Component<Props, State> {
       return <Svg name="angular" />;
     }
 
-    if (depth === 0) {
+    if (depth === 0 && projectRoot === "") {
```
darkwing: Correcting the root / globe issue is actually this easy.

darkwing:
This should be ready for another look.

darkwing:
screeny2

  Looks good in RTL too.

Issue #4852, opened 2017-12-03
[Editor] Adding a breakpoint should not cause the editor to scroll


magicp-github:
**Steps to reproduce:**
1. Launch the latest Nightly
2. Go to about:home
3. Open Debugger
4. Open resource://activity-stream/data/content/activity-stream.bundle.js
5. Scroll down about one or two page <- _This is an important step for reproducing_
6. Add breakpoint (e.g. Click line number 104) <- _This step works_
7. Remove breakpoint
8. Switch DevTools tab, then go back Debugger
9. Add breakpoint in line 104 <- _No response (or make a breakpoint in other line)_
10. Click in code view -> cursor will be jump to other line

**Actual Results:**
Add breakpoint is no response (or make a breakpoint in other line) after switching DevTools tab. If you click in code view, cursor will be jump to other line.

**Expected Results:**
Add breakpoint works even if after switching DevTools tab.
Cursor does not jump to other line.

juliandescottes:
Note that this behavior seems to depend on your toolbox height. I was about to log a different bug, but I think it would be a duplicate so I'll just add my STRs here.

STRs:
- make sure to use a rather small toolbox height (you can set devtools.toolbox.footer.height to 250 in about:config for instance, assuming toolbox docked to the bottom)
- go to http://firefox-dev.tools/debugger-examples/examples/todomvc/
- set a breakpoint in app-view.js, render() method, line 52
- add to todo to break on your breakpoint
- switch to console, switch back to debugger
- click on your breakpoint in the gutter to delete it

AR: Breakpoint is added at line 15
ER: Breakpoint is deleted 

Note that the behavior really changes depending on the toolbox height. With 300px instead of 250px, clicking on the breakpoint does nothing. With 400px, the buggy breakpoint gets added on line 24 instead of 15.

![debugger_breakpoint_wrong_location](https://user-images.githubusercontent.com/1141550/33583050-a83ecdbc-d958-11e7-94c6-5d49bf6ad5bc.gif)

jasonLaster:
Oh yeah, this is certainly an interesting codemirror bug. I think it will likely turn out that codemirror needs to be forced to update when we re-display the document or make the UI visible.

Great catch!

by the way, I reproduced by going to todomvc and todo-view.js:104
http://firefox-dev.tools/debugger-examples/examples/todomvc/

[...]

PR #5643, opened 2018-03-08
Added page visibility and observer to refresh CodeMirror editor after…

mmcote:
Added page visibility and observer to refresh CodeMirror editor after tab is visible.

Fixes Issue: #4852 

### Summary of Changes

1. Upgraded CodeMirror to the latest version 5.34.
2. Added a visibilitychange event listener to detect when the debugger tab in the panel is visible.
3. Added a mutationObserver to call a CodeMirror `refresh()` after tab is fully visible and not collapsed.

### Screenshots/Videos

- Before: 
After scrolling down the document and switching tabs elsewhere and back to the debugger tab. If the user is to click to make a new breakpoint, the breakpoint will appear on a line much higher up than selected. If the user is to click in the text portion of the editor, the editor will re-focus on a line much higher up than selected.
beforefixjumping

- After:
afterfixjump

### Alternative approaches that were also explored

1. [~Worked] Scrolling to the location upon receiving the specific CodeMirror `signal(...)` to the location that was properly stored in the instance, just not updated to. -- This approach was passed on as it wasn't very generic. For each breaking case you would have to have this event handler, possibly missing some cases. For example, the breakpoint would be firing this handler upon a `"gutterclick"` signal and for the click in the text of the editor we would fire upon a `"focus"` signal. 

2. [Failed] Sending different CodeMirror signals, `"refresh"`, `"update"`, `"scroll"`, `"focus"`, `"viewportChange"`, etc.

3. [Failed] Throwing a CodeMirror `refresh()` upon the change in visibility of the page.

4. [Failed] Forcing an update with `forceUpdate()`, using an instance variable, `this.visibilityChange = true`, to track if the update was just forced from the visibilityChange, then doing a CodeMirror `refresh()` in `componentDidUpdate()`.

5. [Failed] Using React's `ref` in hopes that it will only be called back when the actual DOM has been flushed out. Then trying to do a refresh, but this is called before `componentDidMount` and `componentDidUpdate`.

6. [Failed] Spawning an event handler when the tab is visible, to listen for some HTML DOM events such as `"load"`.

- 3, 4, 5, and 6, all failed because the refresh is called while the tab is collapsed.

6. [Worked] Forcing an update and in componentDidUpdate doing a CodeMirror `refresh()` (allowing the `refresh()` on every update). This worked.. but because the original `refresh()` spawned two more update cycles, in which the third one, or last one occurred after the tab is visible again or no longer collapsed. 

7. [Worked] Method proposed in this PR.

### DIFF
--- a/src/components/Editor/index.js
+++ b/src/components/Editor/index.js
@@ -128,21 +128,25 @@ class Editor extends PureComponent<Props, State> {
   }
 
   initPageVisibility = () => {
-    window.addEventListener("visibilitychange", () => {
-      if (document.visibilityState == "visible") {
-        const boxEl = window.parent.document.getElementById("toolbox-deck");
+    window.addEventListener(
+      "visibilitychange",
+      () => {
+        if (document.visibilityState == "hidden") {
+          const vBoxEl = window.parent.document.getElementById("toolbox-panel-jsdebugger");
 
-        if (boxEl) {
-          const observer = new MutationObserver((mutationsList) => {
-            if (isEditorVisible(mutationsList)) {
-              this.state.editor.codeMirror.refresh();
-              observer.disconnect();
-            }
-          });
-          observer.observe(boxEl, { attributes: true });
+          if (vBoxEl) {
+            const observer = new MutationObserver(mutationsList => {
+              if (isEditorVisible(mutationsList)) {
+                this.state.editor.codeMirror.refresh();
+                observer.disconnect();
+              }
+            });
+            observer.observe(vBoxEl, { attributes: true });
+          }
         }
-      }
-    }, false);
+      },
+      false
+    );
   };
 
   setupEditor() {
@@ -160,7 +164,6 @@ class Editor extends PureComponent<Props, State> {
     const { codeMirror } = editor;
     const codeMirrorWrapper = codeMirror.getWrapperElement();
 
-    this.initPageVisibility.bind(this);
     this.initPageVisibility();
 
     resizeBreakpointGutter(codeMirror);
--- a/src/components/Editor/index.js
+++ b/src/components/Editor/index.js
@@ -126,6 +126,34 @@ class Editor extends PureComponent<Props, State> {
     this.scrollToLocation(nextProps);
   }
 
+  initPageVisibility = () => {
+    function changeHandler() {
+      if (document.visibilityState == "visible") {
+        const boxEl = window.parent.document.getElementById("toolbox-deck");
+        const config = { attributes: true };
+        const callback = function(mutationsList) {
+          for (const mutation of mutationsList) {
+            if (
+              mutation.type == "attributes" &&
+              mutation.attributeName == "collapsed" &&
+              mutation.target.collapsed == false
+            ) {
+              this.state.editor.codeMirror.refresh();
+              observer.disconnect();
+            }
+          }
+        };
+
+        const observer = new MutationObserver(callback.bind(this));
+        observer.observe(boxEl, config);
+      }
+    }
+
+    const visibilityChange = "visibilitychange";
+    window.addEventListener(visibilityChange, changeHandler.bind(this), false);
+  };
+    
+
   setupEditor() {
     const editor = createEditor();
 
@@ -140,6 +168,9 @@ class Editor extends PureComponent<Props, State> {
 
     const { codeMirror } = editor;
     const codeMirrorWrapper = codeMirror.getWrapperElement();
+
+    this.initPageVisibility.bind(this);
+    this.initPageVisibility();
 
     resizeBreakpointGutter(codeMirror);
     resizeToggleButton(codeMirror);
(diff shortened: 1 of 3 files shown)


src/components/Editor/index.js
```diff
@@ -126,6 +126,34 @@ class Editor extends PureComponent<Props, State> {
...
+          }
+        };
+
+        const observer = new MutationObserver(callback.bind(this));
+        observer.observe(boxEl, config);
+      }
+    }
+
+    const visibilityChange = "visibilitychange";
+    window.addEventListener(visibilityChange, changeHandler.bind(this), false);
```
jasonLaster: this could use arrow expressions
```js
    window.addEventListener("visibilitychange", () => {
      ...
    }, false);
```

src/components/Editor/index.js
```diff
@@ -126,6 +126,34 @@ class Editor extends PureComponent<Props, State> {
     this.scrollToLocation(nextProps);
   }
 
+  initPageVisibility = () => {
+    function changeHandler() {
+      if (document.visibilityState == "visible") {
+        const boxEl = window.parent.document.getElementById("toolbox-deck");
+        const config = { attributes: true };
+        const callback = function(mutationsList) {
+          for (const mutation of mutationsList) {
```
jasonLaster: this check might benefit from being abstracted to:

```js
function isEditorVisible(mutationList) {
  for (const mutation of mutationsList) {
             if (
               mutation.type == "attributes" &&
               mutation.attributeName == "collapsed" &&
               mutation.target.collapsed == false
             ) {
      return true;
 }
    return false
}
```

jasonLaster:
Can you share some alternatives that you tried and what worked / didn't?

mmcote:
Hey @jasonLaster, at the bottom of the PR description I posted the alternative solutions that I explored.

src/components/Editor/index.js
```diff
@@ -141,6 +160,9 @@ class Editor extends PureComponent<Props, State> {
     const { codeMirror } = editor;
     const codeMirrorWrapper = codeMirror.getWrapperElement();
 
+    this.initPageVisibility.bind(this);
+    this.initPageVisibility();
```
jasonLaster: do we need the bind if we're using the assignment syntax `initPageVisibility = () => {}`

src/utils/editor/index.js
```diff
@@ -180,3 +180,16 @@ export function getTextForLine(codeMirror, line) {
 export function getCursorLine(codeMirror) {
   return codeMirror.getCursor().line;
 }
+
+export function isEditorVisible(mutationsList) {
+  for (const mutation of mutationsList) {
```
jasonLaster: how do we know this is the editor that is no longer collapsed?

src/components/Editor/index.js
juliandescottes: method.bind(this) returns a new method but doesn't modify the existing one, so I think the first statement is simply not needed here?

juliandescottes:
Thanks for working on this! 

I see you are using visibility handlers here. I checked if the toolbox provided any helpers about this, but only found: https://searchfox.org/mozilla-central/source/devtools/client/shared/components/VisibilityHandler.js

(Won't help you much here, but dropping this just for reference :) )

nchevobbe:
> (Won't help you much here, but dropping this just for reference :) )

There's also a custom react-redux connect function built on top of the VisibilityHandler: https://searchfox.org/mozilla-central/source/devtools/client/shared/redux/visibility-handler-connect.js

src/utils/editor/index.js
mmcote: Good call, I just checked, and this fix doesn't work for all the tabs, that collapse is triggered for the console which has its own `<box ...>` element. The other tabs seem to be in `<vbox ...>`'s in the same box as the debugger. It think it would be better to observe if the debugger tab is selected or not. Which is set in the debugger's `<vbox ...>`.

image

src/components/Editor/index.js
mmcote: Thanks for the help from both of you :) Yes, that line seems to be unneeded after testing without it.

jasonLaster:
Hmm - maybe we get some toolbox events we can listen too...
https://searchfox.org/mozilla-central/source/devtools/client/framework/target.js#599


This lists all the events that emit:
https://searchfox.org/mozilla-central/search?q=emit(&case=false&regexp=false&path=devtools%2Fclient

`devtools/client/framework/toolbox.js` is a great place to hunt


There should be a nice way to listen for these events

jasonLaster:
https://searchfox.org/mozilla-central/search?q=VisibilityHandler&case=false&regexp=false&path=

mmcote:
Closing this PR, as we will be trying a new approach where we unmount when we leave the tab, hence possibly creating a new CodeMirror instance when going back to the debugger tab and solving this issue.

PR #5644, opened 2018-03-08
Improve evaluating expressions when pausing

jasonLaster:
Fixes Issue: #5509

### Summary of Changes

This is a bunch of improvements to how we think about console evaluations and pause timing. I don't know how much of this we'll want to include in the final version

- clears the frames when we resume
- adds a new `evaluateMessages` helpers
- adds a new `waitForEvaluations` helpers

### DIFF
--- a/src/actions/pause/fetchExtra.js
+++ b/src/actions/pause/fetchExtra.js
@@ -11,11 +11,7 @@ import type { ThunkArgs } from "../types";
 export function fetchExtra() {
   return async function({ dispatch, getState }: ThunkArgs) {
     const frame = getSelectedFrame(getState());
-    if (!frame) {
-      return;
-    }
-
-    const extra = await dispatch(getExtra("this;", frame.this, frame));
+    const extra = await dispatch(getExtra("this;", frame.this));
     dispatch({
       type: "ADD_EXTRA",
       extra: extra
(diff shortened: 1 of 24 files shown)


src/actions/pause/fetchExtra.js
```diff
@@ -11,11 +11,7 @@ import type { ThunkArgs } from "../types";
 export function fetchExtra() {
```
bomsy: FetchExtra??? This feels a bit vague, down the line this might be really confusing. i feel like this can be much clearer.

src/actions/pause/fetchExtra.js
```diff
@@ -11,11 +11,7 @@ import type { ThunkArgs } from "../types";
...
   return async function({ dispatch, getState }: ThunkArgs) {
     const frame = getSelectedFrame(getState());
-    if (!frame) {
-      return;
-    }
-
-    const extra = await dispatch(getExtra("this;", frame.this, frame));
+    const extra = await dispatch(getExtra("this;", frame.this));
```
bomsy: same ... getExtra?? not sure what it returns.  getExtraProperties maybe?

jasonLaster: good point!

src/actions/pause/fetchExtra.js
jasonLaster: can you file an issue for renaming this

bomsy: Ok will do.

Issue #5639, opened 2018-03-08
[Editor] does not update on attach


jasonLaster:
I'm seeing some funny business when i open the devtools panel. 

1. the editor does not show the text
2. switching tabs does not update the editor
3. breakpoints are missing

I think React 16 is to blame, but i'll need to dig deeper :)

lukaszsobek:
Noticed that too, especially on reloading the debugger with already open tabs - you would need to manually trigger an update by clicking the position in the sidebar.

PR #5645, opened 2018-03-09
Fix permanent loading state

jasonLaster:
Fixes Issue: #5639

### Summary of Changes

This PR #5296 started updating every source that did not have a generatedSourceURL, the effect was that sources would go from being loaded to unloaded. This fixes that and cleans up loadSourceMap a bit.

Another nice effect is that it cleaned up our redux actions quite a bit

### Testing

Given that this was a race condition it is difficult to have a mochitest for it. I'll investigate some options tomorrow.

### DIFF
--- a/src/actions/sources/newSources.js
+++ b/src/actions/sources/newSources.js
@@ -22,7 +22,7 @@ import {
   getPendingBreakpointsForSource
 } from "../../selectors";
 
-import type { Source } from "../../types";
+import type { Source, SourceId } from "../../types";
 import type { ThunkArgs } from "../types";
 
 function createOriginalSource(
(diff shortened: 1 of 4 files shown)


src/actions/sources/newSources.js
```diff
@@ -44,15 +44,20 @@ function createOriginalSource(
  * @memberof actions/sources
  * @static
  */
-export function loadSourceMap(generatedSource: Source) {
+export function loadSourceMap(generatedSourceId: SourceId) {
   return async function({ dispatch, getState, sourceMaps }: ThunkArgs) {
-    let urls;
+    const generatedSource = getSource(getState(), generatedSourceId);
```
jasonLaster: the reason why the source state reverted to the loading state (i believe) is that we were using stale source data.

Re-querying the store helps with that

src/actions/sources/newSources.js
```diff
@@ -44,15 +44,20 @@ function createOriginalSource(
...
+    if (!generatedSource.sourceMapURL) {
```
jasonLaster: this new check keeps us from attempting to fetch source maps if the sources don't have a source map url.

AlexisDeschamps: Makes sense.

Issue #5616, opened 2018-03-06
[Frames] Refactor utility methods


jasonLaster:
I've noticed that we have added a lot of utilities in the frame module.

- some of them are complicated and deserve their own module and tests
- some should be moved to a functions util directory
- we should move to a `/frame/index.js` pattern

Here is the [branch] i started

[branch]: https://github.com/devtools-html/debugger.html/compare/master...jasonLaster:refactor-frames?expand=1

PR #5646, opened 2018-03-09
[Frames] Refactor utility methods

calebstdenis:
Fixes Issue: #5616

### Summary of Changes
* Moved `frame.js` util module to index pattern, where each public util function in is its own file and is re-exported from  `index.js`. 
* Split specs accordingly.

### Test Plan
This is just a refactor of an already well-tested module, so linting and current tests should cover it well enough.

### Screenshots 
What the directory structure looks like now:
image


src/utils/pause/frames/simplifyDisplayName.js
```diff
@@ -0,0 +1,36 @@
...
+
+  for (const reg of scenarios) {
+    const match = reg.exec(displayName);
+    if (match) {
+      return match[1];
+    }
+  }
+
+  return displayName;
+}
```
jasonLaster: maybe we collapse three modules into one:

simplifyDN, formatDN, and formatCN

all of these could go into `displayName` perhaps... 

not a big deal of course, just not sure if we need to go all the way towards one function per module

calebstdenis: sure thing

calebstdenis: @jasonLaster Done in fb87d2b8ad63ec7e21d35b18d66d20072e9058fe

I noticed the index pattern made this refactor easy to do. :)

IMO things were easier to comprehend at one function per module.

Issue #5584, opened 2018-03-02
[Windows] Ctrl+P `.highlight` markers font looks blurry on Windows


jnachtigall:
I noticed this since a 3, 4 weeks. The highlighted, bold parts of the Search matches of `Ctrl+P` look blurry on Windows. No idea why, maybe some odd fonts settings. I also captured the browser.xul output, but cannot see anything obvious in the CSS:

![image](https://user-images.githubusercontent.com/3307404/36890421-0f680514-1dfe-11e8-9597-1042e7d2a4e6.png)

See how bold, kinda ugly the bold `api` match looks. The letter spacing is kind of too narrow.

I looked at this also on Linux and there it looks fine.

PS Looking at the source I see that the match is hightlighted using `<strong class="hightlight">`. While this is not totally incorrect afaik, I think it should rather be `<mark>` to make it semantically 100% correct and better accessibly. By the way there's also https://github.com/bvaughn/react-highlight-words (which also uses `mark` not `strong` here).

lukaszsobek:
@jasonLaster I read up on the code and fuzzy-aldrin-plus uses strong internally. Technically it should accept options as its third parameter in which we are then able to set the way it highlights, but this options object seems to be completely ignored. Referring the issue: https://github.com/jeancroy/fuzz-aldrin-plus/issues/37

jasonLaster:
Thanks, lets see what they say

jeancroy:
Does the usage of mark vs strong make a difference in anti-aliasing ?

lukaszsobek:
Merged. Closing.

jnachtigall:
@lukaszsobek It still (or again) looks blurry on Nightly:

![image](https://user-images.githubusercontent.com/3307404/40174899-be6b96b6-59d6-11e8-8db8-9bdd4adb48e8.png)

For instance, check the 4th entry `PathUtils.js` letters. So using `<mark>` instead of `<strong>` did not seem to have fixed the issue. I am on Windows 10.

Can you reopen?

PS Looks like the 
```css
.result-item .title .highlight {

    font-weight: bold;
}
```
causes this. I am not a fonts expert, but maybe the used font on Windows is not great for bold.

PR #5647, opened 2018-03-09
Changes highlight from <strong> to <mark>

lukaszsobek:
Fixes Issue: #5584

Replaces `<strong>` in highlights with `<mark>`

## Vid:
mark_instead_of_strong

### DIFF
--- a/src/components/QuickOpenModal.js
+++ b/src/components/QuickOpenModal.js
@@ -320,8 +320,13 @@ export class QuickOpenModal extends Component<Props, State> {
 
   /* eslint-disable react/no-danger */
   renderHighlight = (candidateString: string, query: string, name: string) => {
-    const html = fuzzyAldrin.wrap(candidateString, query);
-
+    const options = {
+      wrap: {
+        tagOpen: '<mark class="highlight">',
+        tagClose: "</mark>"
+      }
+    };
+    const html = fuzzyAldrin.wrap(candidateString, query, options);
     return <div dangerouslySetInnerHTML={{ __html: html }} />;
   };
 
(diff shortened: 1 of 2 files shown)


src/components/QuickOpenModal.js
```diff
@@ -320,8 +320,13 @@ export class QuickOpenModal extends Component<Props, State> {
 
   /* eslint-disable react/no-danger */
   renderHighlight = (candidateString: string, query: string, name: string) => {
-    const html = fuzzyAldrin.wrap(candidateString, query);
-
+    const options = {
+      wrap: {
+        tagOpen: '<mark class="highlight">',
```
darkwing: At face value I like this change; two questions though:

1.  Do we need to update a CSS  style in our codebase for this?
2.  `<mark>` seems like the correct semantic choice, which is why I love this change, but any idea why they wouldn't use `<mark>` by default?  Any clue why `<strong>`?

lukaszsobek: ad 1. the class 'highlight' is not tag dependent, however replacing 'strong' with 'mark' makes the items background yellow.

ad 2. The connected issue provides all information and links. In brief: The author was not aware it could be an issue and is considering it because we made him aware that it might. As far as I understood it he wants to bundle this change with some others so it might take some more time.
Speaking to Jason there was the notion to include this fix in favor of waiting for a native fix that may or may not come soon as it will fix the issue now and should have no negative impact in future where we might be able to get rid of the configuration option altogether.

jasonLaster: hmm, so I'm not a fan of the yellow background. if the markup makes it clearer, we can just remove the background via css

jeancroy:
Hi, can I ask how the change from `strong` to `mark` fix the blurry font issue ?
Is there something specific to the strong that that affect rendering ?

jasonLaster:
CC @lukaszsobek

Issue #4849, opened 2017-12-02
[Performance] Webpack devtool modes slow down page load


ChrisGitter:
I tried to use the debugger with a big React/Redux project using Webpack for bundling. At first load the debugger opens slowly. The site load itself is not effected at first. But when I set a breakpoint somewhere and then reload, the site load takes up to a minute or so. It also freezes the inspector tabs and the CPU is under some load.

I recently tried to switch from chrome to Firefox and I use the same settings as I do in Chrome, where debugger and breakpoints work without any delay or preformance issues.

My webpack config uses the 'cheap-module-eval-source-map' setting.

To reproduce I can take any React projects that use webpack, for example: https://github.com/andrewngu/sound-redux
I cloned it, changed the devtool setting in webpack to use source maps and it showed the same issues.

I have tried this with ArchLinux and Ubuntu, both on latest versions and both using the latest Firefox developer edition.

jasonLaster:
Thanks @ChrisGitter! We'll take a look :)

tbaud0n:
I'm not using webpack but my webapp takes more than 12 seconds to load if the devtools are open and only 2-3 seconds if devtools panels are closed.
It's a local development and the page loads 280 files (mainly js + css files). 
@jasonLaster do you think it's linked to this issue or do I need to feel a new issue?

jasonLaster:
Hey @tbaud0n - thanks for filing :) we're working on this now and have 3 fixes for it

1. throttle updating the redux store for each source
2. schedule how we load source maps so we don't load them immediately
3. schedule async work like syncing breakpoints so they're not immediately done

2 & 3 are now on Firefox Nightly. 1 is a huge win and we're just fixing up a memory leak with it. Could you try Firefox nightly and let us know what you think?

Also, what's your site? We can test it w/ 1 locally :)

And, of course ping you when it lands

tbaud0n:
I've tried using nightly and there is a little improvement (9 sec instead of 12 sec) but the problem is still present as it should take only 3 seconds to load the page.

On discourse.mozilla.org, there's a [discussion](https://discourse.mozilla.org/t/devtools-slow-and-occasionally-unresponsive-on-some-sites/19936/35) about this issue. Julian Descottes helped me to find that the problem happens when the network panel has been opened.

Here is the perf without opening the network panel : https://perfht.ml/2CYFNmU
The perf with the network panel open : https://perfht.ml/2oFPlAh

[...]

PR #5648, opened 2018-03-09
Improve loading sources

jasonLaster:
Fixes Issue: #4849

### Summary of Changes

Previously, loading perf.html when it used webpack `eval-source-map` took about 17 seconds of heavy work. Now it takes about 4 seconds of less intensive work and feels pretty snappy.

* adds some checks from #5645, which prevent unnecessary updates
* batches all of the calls to `newSources` from `loadSourceMap`

Loading is now pretty snappy

NOTE: this also works for cheap-source-map

#### Before

- lots of loadSourceMap calls
- lots of updateTree calls

#### After

- a few updateTree calls
- minimal loadSourceMap time

### DIFF
--- a/src/actions/sources/newSources.js
+++ b/src/actions/sources/newSources.js
@@ -8,13 +8,16 @@
  * Redux actions for the sources state
  * @module actions/sources
  */
+
+import { isGeneratedId } from "devtools-source-map";
+import { flatten } from "lodash";
+
 import { toggleBlackBox } from "./blackbox";
 import { syncBreakpoint } from "../breakpoints";
 import { loadSourceText } from "./loadSourceText";
 import { togglePrettyPrint } from "./prettyPrint";
 import { selectLocation } from "../sources";
 import { getRawSourceURL, isPrettyURL } from "../../utils/source";
-
 import {
   getBlackBoxList,
   getSource,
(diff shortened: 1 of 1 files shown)


src/actions/sources/newSources.js
```diff
@@ -14,7 +14,7 @@ import { loadSourceText } from "./loadSourceText";
 import { togglePrettyPrint } from "./prettyPrint";
 import { selectLocation } from "../sources";
 import { getRawSourceURL, isPrettyURL } from "../../utils/source";
-
+import { flatten } from "lodash";
```
AlexisDeschamps: This seems very cool.

src/actions/sources/newSources.js
```diff
@@ -171,11 +179,10 @@ export function newSources(sources: Source[]) {
       dispatch(checkPendingBreakpoints(source.id));
     }
 
-    await Promise.all(
-      filteredSources.map(source => dispatch(loadSourceMap(source.id)))
-    );
+    await dispatch(loadSourceMaps(filteredSources));
```
AnshulMalik: We should load source maps for only the generated sources, how about we incorporate that change here as well?

darkwing: Thoughts on this @jasonLaster or can we merge?

jasonLaster: done. good idea!

Issue #5633, opened 2018-03-07
[Editor] should not flash when leaving a location


jasonLaster:
The editor should smoothly go from one location to another when stepping. 

In this GIF, we should only see the blue line

![](http://g.recordit.co/scFZIS9shy.gif)

In this GIF, we should see the grey lines :) 
![](http://g.recordit.co/1ct6gqnPCf.gif)

The grey lines serve as a way of highlighting where we're going too (if we aren't using the blue debug line). At the moment they also flash the line we're leaving and that's the problem.

jasonLaster:
We just landed a PR, which fixed many of the cases

we now just have this case: reloading and seeing the highlight... but otherwise this is really great.

![](http://g.recordit.co/Y2RFphqxYb.gif)

PR #5650, opened 2018-03-09
[Editor] [WIP] Stop highlighting the line being left when stepping

AlexisDeschamps:
Fixes Issue: #5633

### Summary of Changes

* Stopped highlighting the line being left when using stepping functions

### Test Plan

- [x] Open http://firefox-dev.tools/debugger-examples/examples/todomvc/ as debugee
- [x] In `Sources`: open `firefox-dev.tools`, open `debugger-examples/examples/todomvc`, open `bower_components`, open `js`, open `views`, open `app-view.js`
- [x] Put a breakpoint on line 31 of `app-view.js`
- [x] Refresh http://firefox-dev.tools/debugger-examples/examples/todomvc/
- [x] In the top right of the Debugger, press any of the `Resume`, `Step over`, `Step in`, or `Step out` buttons
- [x] Notice the line being left not being highlighted in grey
- [x] Breakpoints can be clicked repeatedly and they will highlight the appropriate location in grey

## Notes
Cases this seems to fix
- Clicking `Resume`, `Step over`, `Step in`, or `Step out` buttons—removes the unwanted grey line on the previous location
- Repeatedly clicking the same breakpoint in the secondary pane—after every click, the breakpoint location is selected and highlighted in grey, as desired
- Clicking on a breakpoint in the secondary pane and then clicking a debugger button (`Step over`, `Step in`, etc.)—the breakpoint location is selected and highlighted in gray, then, when a debugger button is clicked, no line is highlighted in grey, but the new selected location is highlighted in blue as desired

Why this works: upon clicking a debugger button, `isInSteppingUpdate ` is set to `true` and no line is highlighted until the component runs through its updates. On the last one, the previous and current lines are different, so `isInSteppingUpdate` can be set to `false`. At this point, the component allows highlighting.

Repeatedly clicking a breakpoint still works because `isInSteppingUpdate ` will not be set to true.

### Videos
Before
5633-before

After
5633-after

### DIFF
--- a/src/components/Editor/HighlightLine.js
+++ b/src/components/Editor/HighlightLine.js
@@ -18,6 +18,7 @@ import {
 import type { Frame, Location, SourceRecord } from "../../types";
 
 type Props = {
+  pauseCommand: string,
   selectedFrame: Frame,
   selectedLocation: Location,
   selectedSource: SourceRecord
@@ -43,14 +44,50 @@ function isDocumentReady(selectedSource, selectedLocation) {
 }
 
 export class HighlightLine extends PureComponent<Props> {
+  isStepping: boolean;
+  previousEditorLine: ?number;
+
+  constructor(props: Props) {
+    super(props);
+    this.isStepping = false;
+    this.previousEditorLine = null;
+  }
+
+  shouldComponentUpdate(nextProps: Props) {
+    const { selectedLocation, selectedSource } = nextProps;
+    return this.shouldSetHighlightLine(selectedLocation, selectedSource);
+  }
+
+  shouldSetHighlightLine(
+    selectedLocation: Location,
+    selectedSource: SourceRecord
+  ) {
+    const { sourceId, line } = selectedLocation;
+    const editorLine = toEditorLine(sourceId, line);
+    if (this.isStepping && editorLine === this.previousEditorLine) {
+      return false;
+    }
+    if (!isDocumentReady(selectedSource, selectedLocation)) {
+      return false;
+    }
+    return true;
+  }
+
   componentDidUpdate(prevProps: Props) {
-    const { selectedLocation, selectedFrame, selectedSource } = this.props;
+    const {
+      pauseCommand,
+      selectedLocation,
+      selectedFrame,
+      selectedSource
+    } = this.props;
+    if (pauseCommand) {
+      this.isStepping = true;
+    }
 
     this.clearHighlightLine(
       prevProps.selectedLocation,
       prevProps.selectedSource
     );
-
     this.setHighlightLine(selectedLocation, selectedFrame, selectedSource);
   }
 
@@ -59,17 +96,18 @@ export class HighlightLine extends PureComponent<Props> {
     selectedFrame: Frame,
     selectedSource: SourceRecord
   ) {
-    if (!isDocumentReady(selectedSource, selectedLocation)) {
+    const { sourceId, line } = selectedLocation;
+    if (!this.shouldSetHighlightLine(selectedLocation, selectedSource)) {
       return;
     }
-
-    const { sourceId, line } = selectedLocation;
+    this.isStepping = false;
+    const editorLine = toEditorLine(sourceId, line);
+    this.previousEditorLine = editorLine;
 
     if (!line || isDebugLine(selectedFrame, selectedLocation)) {
       return;
     }
 
-    const editorLine = toEditorLine(sourceId, line);
     const doc = getDocument(sourceId);
     doc.addLineClass(editorLine, "line", "highlight-line");
   }
@@ -91,6 +129,7 @@ export class HighlightLine extends PureComponent<Props> {
 }
 
 export default connect(state => ({
+  pauseCommand: state.pause.command,
   selectedFrame: getVisibleSelectedFrame(state),
   selectedLocation: getSelectedLocation(state),
   selectedSource: getSelectedSource(state)
--- a/src/components/Editor/HighlightLine.js
+++ b/src/components/Editor/HighlightLine.js
@@ -12,13 +12,15 @@ import { connect } from "react-redux";
 import {
   getVisibleSelectedFrame,
   getSelectedLocation,
-  getSelectedSource
+  getSelectedSource,
+  getPauseCommand
 } from "../../selectors";
 
 import type { Frame, Location, SourceRecord } from "../../types";
+import type { Command } from "../../reducers/types";
 
 type Props = {
-  pauseCommand: string,
+  pauseCommand: Command,
   selectedFrame: Frame,
   selectedLocation: Location,
   selectedSource: SourceRecord
@@ -44,14 +46,8 @@ function isDocumentReady(selectedSource, selectedLocation) {
 }
 
 export class HighlightLine extends PureComponent<Props> {
-  isStepping: boolean;
-  previousEditorLine: ?number;
-
-  constructor(props: Props) {
-    super(props);
-    this.isStepping = false;
-    this.previousEditorLine = null;
-  }
+  isStepping: boolean = false;
+  previousEditorLine: ?number = null;
 
   shouldComponentUpdate(nextProps: Props) {
     const { selectedLocation, selectedSource } = nextProps;
@@ -64,12 +60,15 @@ export class HighlightLine extends PureComponent<Props> {
   ) {
     const { sourceId, line } = selectedLocation;
     const editorLine = toEditorLine(sourceId, line);
-    if (this.isStepping && editorLine === this.previousEditorLine) {
-      return false;
-    }
+
     if (!isDocumentReady(selectedSource, selectedLocation)) {
       return false;
     }
+
+    if (this.isStepping && editorLine === this.previousEditorLine) {
+      return false;
+    }
+
     return true;
   }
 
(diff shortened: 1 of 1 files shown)


jasonLaster:
Hmm I'm a bit confused. We want to go to the location of the frame when we pause or click a frame. The issue is that we shouldn’t show the old grey highlight line when we do.

So for instance if you step from line ten to eleven. We want the blue line to flash on eleven but we want to avoid the grey one on ten.

jasonLaster:
Hey, this is pretty cool. Can you add a comment for what cases this covers. Also a comment in the PR for what you tried and why this works

AlexisDeschamps:
Hmm, I don't think this works quite yet. It doesn't work properly when repeatedly clicking the same frame (only highlights in grey on the first click).

src/components/Editor/HighlightLine.js
```diff
@@ -52,6 +52,15 @@ export class HighlightLine extends PureComponent<Props> {
...
 
+    if (
+      !selectedLocation.forced &&
+      prevProps.selectedLocation.line &&
+      selectedLocation.line &&
+      prevProps.selectedLocation.line === selectedLocation.line
+    ) {
+      return;
+    }
+    selectedLocation.forced = false;
```
AlexisDeschamps: Not sure if this should be done through an action or something.

AlexisDeschamps:
So `isStepping` doesn't seem to really work because when `HighlightLine` `componentDidUpdate` fires, `isStepping` is no longer true. So I made clicking on the breakpoint frames give a special property to `selectedLocation` that makes `HighlightLine` always highlight.

src/components/Editor/HighlightLine.js
jasonLaster: lol, this seems a bit forced.

I'd prefer to use the `prevLine` hack than to pass a highlight flag into selectLocation

src/components/Editor/HighlightLine.js
```diff
@@ -59,17 +77,22 @@ export class HighlightLine extends PureComponent<Props> {
...
       return;
     }
+    this.isInSteppingUpdate = false;
+    this.previousLine = editorLine;
 
-    const { sourceId, line } = selectedLocation;
-
+    if (!isDocumentReady(selectedSource, selectedLocation)) {
+      return;
+    }
```
jasonLaster: some of this logic makes me think we might want to use the lifecycle function `shouldComponentUpdate() {}` to prevent renders...

src/components/Editor/HighlightLine.js
```diff
@@ -59,17 +77,22 @@ export class HighlightLine extends PureComponent<Props> {
...
     selectedSource: SourceRecord
   ) {
-    if (!isDocumentReady(selectedSource, selectedLocation)) {
+    const { sourceId, line } = selectedLocation;
+
+    const editorLine = toEditorLine(sourceId, line);
+    if (this.isInSteppingUpdate && editorLine === this.previousLine) {
       return;
     }
+    this.isInSteppingUpdate = false;
```
jasonLaster: this is a clever way to keep the state a bit longer

src/components/Editor/HighlightLine.js
AlexisDeschamps: Makes sense, I will look into that.

AlexisDeschamps:
@jasonLaster Updated with your suggestions in mind if you want to take a look.

src/components/Editor/HighlightLine.js
```diff
@@ -59,17 +86,22 @@ export class HighlightLine extends PureComponent<Props> {
     selectedFrame: Frame,
     selectedSource: SourceRecord
   ) {
-    if (!isDocumentReady(selectedSource, selectedLocation)) {
+    const { sourceId, line } = selectedLocation;
+
+    const editorLine = toEditorLine(sourceId, line);
+    if (this.isInSteppingUpdate && editorLine === this.previousEditorLine) {
```
jasonLaster: could this condition move to `shouldComponentUpdate`

src/components/Editor/HighlightLine.js
```diff
@@ -59,17 +86,22 @@ export class HighlightLine extends PureComponent<Props> {
...
       return;
     }
+    this.isInSteppingUpdate = false;
+    this.previousEditorLine = editorLine;
 
-    const { sourceId, line } = selectedLocation;
-
+    if (!isDocumentReady(selectedSource, selectedLocation)) {
```
jasonLaster: could this condition move?

jasonLaster:
@AlexisDeschamps this is looking good. Two thoughts:

Lets keep pushing this forward further and simplifying it


I thought a bit more about your idea of having `selectLocation` and `selectSource` take an options object with `{ highlight: true}` and saving it in `reducers/ui` `case SELECT_SOURCE: { this.setState("highlightLine" : actions.highlight })`. 

- it does require the different use cases to know whether they want to highlight, but ... who knows it might be good...
- one concern is that the user clicks a breakpoint, so we set highlight true. then the user pauses, we'll need to clear the highlight so it doesn't flash when leaving

src/components/Editor/HighlightLine.js
AlexisDeschamps: I added it to `shouldComponentUpdate`, not sure if I should remove it from here as well since React can render despite the results of `shouldComponentUpdate` from what I understand.

AlexisDeschamps: From what I have tried, `if (!line || isDebugLine(selectedFrame, selectedLocation))` cannot be put in `shouldComponentUpdate` since it needs to run the remove the current highlight.

src/components/Editor/HighlightLine.js
```diff
@@ -91,6 +126,7 @@ export class HighlightLine extends PureComponent<Props> {
 }
 
 export default connect(state => ({
+  isResumingOrStepping: getIsResumingOrStepping(state),
```
jasonLaster: there should be a better term for this... `getIsResumingOrStepping` maybe `isRunning` dunno... @jimblandy @darkwing what do you think?

AlexisDeschamps:
@jasonLaster Should be updated again. Thanks.

src/components/Editor/HighlightLine.js
```diff
@@ -43,14 +44,50 @@ function isDocumentReady(selectedSource, selectedLocation) {
...
+    return this.shouldSetHighlightLine(selectedLocation, selectedSource);
+  }
+
+  shouldSetHighlightLine(
+    selectedLocation: Location,
+    selectedSource: SourceRecord
+  ) {
+    const { sourceId, line } = selectedLocation;
+    const editorLine = toEditorLine(sourceId, line);
+    if (this.isInSteppingUpdate && editorLine === this.previousEditorLine) {
```
jasonLaster: lets change this to `this.isStepping`

src/components/Editor/HighlightLine.js
```diff
@@ -43,14 +44,50 @@ function isDocumentReady(selectedSource, selectedLocation) {
...
+
   componentDidUpdate(prevProps: Props) {
-    const { selectedLocation, selectedFrame, selectedSource } = this.props;
+    const {
+      pauseCommand,
+      selectedLocation,
+      selectedFrame,
+      selectedSource
+    } = this.props;
+    if (["resume", "stepIn", "stepOver", "stepOut"].includes(pauseCommand)) {
```
jasonLaster: i think this is all of the options :)
could we do something like `!!pauseCommand` which will be the same?

src/components/Editor/HighlightLine.js
```diff
@@ -59,17 +96,18 @@ export class HighlightLine extends PureComponent<Props> {
     selectedFrame: Frame,
     selectedSource: SourceRecord
   ) {
-    if (!isDocumentReady(selectedSource, selectedLocation)) {
+    if (!this.shouldSetHighlightLine(selectedLocation, selectedSource)) {
       return;
     }
-
+    this.isInSteppingUpdate = false;
```
jasonLaster: lets move this down below the destructure

src/components/Editor/HighlightLine.js
```diff
@@ -59,17 +96,18 @@ export class HighlightLine extends PureComponent<Props> {
...
     const { sourceId, line } = selectedLocation;
+    const editorLine = toEditorLine(sourceId, line);
+    this.previousEditorLine = editorLine;
```
jasonLaster: should this be before the IF?

AlexisDeschamps: Hmm, no according to my understanding, since the IF wants to look at the previous value of `this.previousEditorLine`.

AlexisDeschamps:
@jasonLaster Thanks for the suggestions. They should be implemented except for the one I replied to.

src/components/Editor/HighlightLine.js
```diff
@@ -43,14 +44,50 @@ function isDocumentReady(selectedSource, selectedLocation) {
 }
 
 export class HighlightLine extends PureComponent<Props> {
+  isStepping: boolean;
+  previousEditorLine: ?number;
+
+  constructor(props: Props) {
+    super(props);
+    this.isStepping = false;
```
jasonLaster: this can move up to line 47 and 48

src/components/Editor/HighlightLine.js
```diff
@@ -91,6 +129,7 @@ export class HighlightLine extends PureComponent<Props> {
 }
 
 export default connect(state => ({
+  pauseCommand: state.pause.command,
```
jasonLaster: we should use a selector here

jasonLaster:
I pushed some tweaks and opened #5804 for commands.

I noticed some issues reloading and seeing the highlight... but otherwise this is really great.

jasonLaster:
Lets tweek the reload case in a new PR1

jasonLaster:
lets tweak the reload case in a new PR

Issue #5623, opened 2018-03-07
[QuickOpen] Loading message never disappears


AnshulMalik:
<img width="1025" alt="screen shot 2018-03-07 at 4 55 32 pm" src="https://user-images.githubusercontent.com/7821757/37089910-5fa0dd02-2228-11e8-9f80-e1ae9a98dd5c.png">

When you open Quick Open dialog(Cmd+P), you can see `Loading...` text displayed even after the results are available.

Expected:
`Loading...` should disappear when results are available.

atwalg2:
Hello @atiqueansari1987 
Since I caused this bug. I came up with this solution. 
I'm going to make a pull request for it tonight.

In src/components/QuickOpenModal.js
```    const { enabled, query, symbols, selectedSource } = this.props;
```
...
```        {selectedSource != null &&
          this.isSymbolSearch() &&
          (!symbols ||
            (symbols.functions.length == 0 && (
              <div className="loading-indicator">
                {L10N.getStr("loadingText")}
              </div>
            )))}
```

PR #5653, opened 2018-03-11
[QuickOpen] Fix loading for searching symbols

atwalg2:
Fixes Issue: #5623 

### Summary of Changes

* Loading indicator to show only when searching up symbols, and only with a source tab selected
* update snapshot

### Test Plan
- [x] cntrl + shift + o with no tabs open (Shouln't show)
- [x] cntrl + shift + o with tabs(source) open (Should Show)

### Screenshots/Videos
loadingindicator

### DIFF
--- a/src/components/QuickOpenModal.js
+++ b/src/components/QuickOpenModal.js
@@ -362,6 +362,21 @@ export class QuickOpenModal extends Component<Props, State> {
     return !this.getResultCount() && !!query;
   }
 
+  
+  renderSymLoading = (symbols: FormattedSymbolDeclarations, selectedSource: any, 
+    isSymbolSearch: boolean ) => {
+
+    if(isSymbolSearch && selectedSource) {
+      if(!symbols || symbols.functions.length == 0) {      
+        return(
+          <div className="loading-indicator">
+            {L10N.getStr("loadingText")}
+          </div>
+        )
+      }
+    }
+  }
+
   render() {
     const { enabled, query, symbols, selectedSource } = this.props;
     const { selectedIndex, results } = this.state;
@@ -388,14 +403,8 @@ export class QuickOpenModal extends Component<Props, State> {
             expanded && items[selectedIndex] ? items[selectedIndex].id : ""
           }
         />
-        {this.isSymbolSearch() &&
-          selectedSource &&
-          (!symbols ||
-          (symbols.functions.length == 0 && (
-            <div className="loading-indicator">
-              {L10N.getStr("loadingText")}
-            </div>
-          )))}
+
+        {this.renderSymLoading(symbols, selectedSource, this.isSymbolSearch())}
         {newResults && (
           <ResultList
             key="results"
--- a/src/components/QuickOpenModal.js
+++ b/src/components/QuickOpenModal.js
@@ -362,23 +362,20 @@ export class QuickOpenModal extends Component<Props, State> {
     return !this.getResultCount() && !!query;
   }
 
-  
-  renderSymLoading = (symbols: FormattedSymbolDeclarations, selectedSource: any, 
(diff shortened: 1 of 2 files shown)


src/components/QuickOpenModal.js
```diff
@@ -383,12 +383,14 @@ export class QuickOpenModal extends Component<Props, State> {
             expanded && items[selectedIndex] ? items[selectedIndex].id : ""
           }
         />
-        {!symbols ||
+        {this.isSymbolSearch() &&
```
AnshulMalik: This is becoming hard to read, how about we move this logic to its own function?

And run `jest -u` to update the tests :)

atwalg2:
@AnshulMalik @jasonLaster I noticed that when we open quickopen menu with '@' prefixed, when we finish loading the symbols it does not show the list automatically. It requires changing the text field first. This is shown by the gif in this PR description.

Should this be added as a separate issue? it does not seem to be influenced by the loading indicator.

src/components/QuickOpenModal.js
```diff
@@ -357,8 +357,23 @@ export class QuickOpenModal extends Component<Props, State> {
     return !this.getResultCount() && !!query;
   }
 
+  
+  renderSymLoading = (symbols: FormattedSymbolDeclarations, selectedSource: any,
```
AnshulMalik: Do you think we can extract these variables from props? Maybe we don't need to pass them here.

src/components/QuickOpenModal.js
```diff
@@ -383,12 +398,8 @@ export class QuickOpenModal extends Component<Props, State> {
...
           }
         />
-        {!symbols ||
-          (symbols.functions.length == 0 && (
-            <div className="loading-indicator">
-              {L10N.getStr("loadingText")}
-            </div>
-          ))}
+
+        {this.renderSymLoading(symbols, selectedSource, this.isSymbolSearch())}
```
jasonLaster: It's okay for this to be `this.renderLoading` and for the function to know which modes we show loading messages for

atwalg2: by that do you mean something like.... 
```renderLoading( mode: string ) {
     { symbols, selectedSource } = this.props
      if(mode === symbol) {
            {code for symbol loading}
```

atwalg2: by that do you mean something like.... 
```renderLoading( mode: string ) {
     { symbols, selectedSource } = this.props
      if(mode === symbol) {
            {code for symbol loading}
```

src/components/QuickOpenModal.js
```diff
@@ -357,8 +357,20 @@ export class QuickOpenModal extends Component<Props, State> {
     return !this.getResultCount() && !!query;
   }
 
+  renderLoading = (isSymbolSearch: boolean) => {
```
jasonLaster: there's no reason to pass the argument `isSymbolSearch`

atwalg2: I tried to use it in the function, but I guess you can't do that, or at least the way I tried. Do you mean I should put it inside the function or just use it in the render like... 
> ( this.isSymbolSearch() && this.renderLoading() )

src/components/QuickOpenModal.js
```diff
@@ -357,8 +357,20 @@ export class QuickOpenModal extends Component<Props, State> {
     return !this.getResultCount() && !!query;
   }
 
+  renderLoading = () => {
+    const { symbols, selectedSource } = this.props;
+
+    if (selectedSource) {
```
jasonLaster: we should still check to see if we're looking for a function or variable?

maybe `isFunctionQuery() || isVariableQuery()`

jasonLaster:
@lukaszsobek not having toJS is so nice

Issue #5378, opened 2018-02-09
[Preview] Support object properties


jasonLaster:
`getSymbols` should support object properties like `1` or `sci-fi` when building up the expressions. We probably should rename memberExpressions to expressions too :)

```js
const firstAuthor = collection.books[1].author;
const firstActionDirector = collection.genres["sci-fi"].movies[0].director;
```

```
memberExpressions:
[(1, 40), (1, 46)] [(1, 20), (1, 46)] collection.books.undefined.author author
[(1, 37), (1, 38)] [(1, 20), (1, 39)] collection.books
[(1, 31), (1, 36)] [(1, 20), (1, 36)] collection.books books
[(2, 66), (2, 74)] [(2, 28), (2, 74)] collection.genres.undefined.movies.undefined.director director
[(2, 63), (2, 64)] [(2, 28), (2, 65)] collection.genres.undefined.movies
[(2, 56), (2, 62)] [(2, 28), (2, 62)] collection.genres.undefined.movies movies
[(2, 46), (2, 54)] [(2, 28), (2, 55)] collection.genres
[(2, 39), (2, 45)] [(2, 28), (2, 45)] collection.genres genres
```

PR #5655, opened 2018-03-11
Fix computed member expressions

bomsy:
Fixes Issue: #5378

### Summary of Changes

* Added a new `isComputedExpression` util to check if an expression string  is computed or not.
* using `extra.raw` instead of `.value`adds the quotes for string literals so that we do not need to try to handle that.


jasonLaster:
Looks good. Just some flow noise

PR #5665, opened 2018-03-13
Removes toJS from 3 files

lukaszsobek:
Edited 3 files to remove possible toJS calls from actions:
- ast.js,
- breakpoints.js,
- addBreakpoint.js

Investigate: `getGeneratedLocation` uses source records, so it would be nice to remove the toJS from here as well

### DIFF
--- a/src/actions/ast.js
+++ b/src/actions/ast.js
@@ -25,13 +25,8 @@ import type { ThunkArgs } from "./types";
 
 export function setSourceMetaData(sourceId: SourceId) {
   return async ({ dispatch, getState }: ThunkArgs) => {
-    const sourceRecord = getSource(getState(), sourceId);
-    if (!sourceRecord) {
-      return;
-    }
-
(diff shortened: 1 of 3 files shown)


src/reducers/ast.js
```diff
@@ -17,6 +17,7 @@ import type { SymbolDeclarations, AstLocation } from "../workers/parser/types";
 import type { Map } from "immutable";
 import type { Source } from "../types";
 import type { Action } from "../actions/types";
+import type { SourceRecord } from "./sources";
```
lukaszsobek: how does that relate to the `SourceRecord` type in reducers/type?

jasonLaster: it might be an Immutable object...

jasonLaster:
something clearly broke w/ sourcemaps. Try debugging  http://firefox-dev.tools/debugger-examples/examples/increment/ or one of the other source map examples

browser_dbg-sourcemaps3.js

PR #5674, opened 2018-03-14
Lock down parser flow type exports

jasonLaster:
### Summary of Changes

* adds flow types for utils/ast
* exports types from parser/index
* fixes a recursion issue caused by defining types in getSymbols

### DIFF
--- a/src/actions/preview.js
+++ b/src/actions/preview.js
@@ -21,7 +21,6 @@ import {
 } from "../selectors";
 
 import { getMappedExpression } from "./expressions";
-
 import { isEqual } from "lodash";
 
 import type { ThunkArgs } from "./types";
@@ -91,7 +90,6 @@ function isInvalidTarget(target: HTMLElement) {
 
 export function updatePreview(target: HTMLElement, editor: any) {
   return ({ dispatch, getState, client, sourceMaps }: ThunkArgs) => {
-    const tokenText = target.innerText ? target.innerText.trim() : "";
     const tokenPos = getTokenLocation(editor.codeMirror, target);
     const cursorPos = target.getBoundingClientRect();
     const preview = getPreview(getState());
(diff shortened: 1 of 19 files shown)


src/actions/preview.js
```diff
@@ -126,9 +126,9 @@ export function updatePreview(target: HTMLElement, editor: any) {
 
     let match;
     if (!symbols || symbols.loading) {
-      match = findBestMatchExpression(symbols, tokenPos, tokenText);
-    } else {
       match = getExpressionFromCoords(editor.codeMirror, tokenPos);
+    } else {
+      match = findBestMatchExpression(symbols, tokenPos);
```
jasonLaster: yikes. really helpful catch

src/workers/parser/getSymbols.js
```diff
@@ -23,40 +23,10 @@ import {
...
 import getFunctionName from "./utils/getFunctionName";
 
-let symbolDeclarations = new Map();
-
-export type ClassDeclaration = {|
-  name: string,
-  location: BabelLocation,
-  parent?: ClassDeclaration
-|};
+import type { SimplePath } from "./utils/simple-path";
```
jasonLaster: just being consistent. i kinda like how it can be more concise, but we dont do it anywhere else

src/utils/ast.js
```diff
@@ -2,7 +2,15 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
 
-export function findBestMatchExpression(symbols, tokenPos, token) {
+// @flow
+
+import type { Position } from "../types";
+import type { SymbolDeclarations } from "../workers/parser";
+
+export function findBestMatchExpression(
```
loganfsmyth: Your fix seems more like the right one, but I'll add that the immediate thing that jumped out to me when I looked is that this function's return type is still weird. It should be `SymbolDeclaration | null` or something, but it is `SymbolDeclaration | {}` right now. That means if you were in a file that doesn't have any identifiers, it still looks like it found something.

jasonLaster: yeah, true

jasonLaster:
cool, this will be fixed in the pause points work

Issue #5204, opened 2018-01-25
[Expression] syntax errors are cleared on save


Loirooriol:
Create a watch expression with a syntax error, e.g. `[1,2,3`

Result: the expression is cleared and "Invalid expression..." is displayed.

Expected: the expression is not cleared so that you can fix the syntax error without having to rewrite it all.

jasonLaster:
@Loirooriol this was added to the debugger in github. It should be in nightly tomorrow or monday.

Loirooriol:
Still bad after [bug 1432842](https://bugzilla.mozilla.org/show_bug.cgi?id=1432842) (v10).

jasonLaster:
CC @bomsy @wldcordeiro what do you think?

bomsy:
I think it's a good idea that we do not clear the invalid expression. 
It can be frustrating to retype really long expressions because of tiny mistakes.
Maybe making the border (and also maybe the text) red should be quite obvious to the user.

bomsy:
Maybe we shoould fix this as well here

![](http://g.recordit.co/RVaVrdnNUD.gif)

The field should not remain red (in the invalid expression state) after the invalid expression has been field.

jasonLaster:
```diff
diff --git a/src/actions/expressions.js b/src/actions/expressions.js
index bf237e9..62fde2e 100644
--- a/src/actions/expressions.js
+++ b/src/actions/expressions.js
@@ -48,6 +48,7 @@ export function addExpression(input: string) {
 }
 
 export function clearExpressionError() {
+  // maybe this will help
   return { type: "CLEAR_EXPRESSION_ERROR" };
 }
 
diff --git a/src/components/SecondaryPanes/Expressions.js b/src/components/SecondaryPanes/Expressions.js
index fb22daf..dbd22d0 100644
--- a/src/components/SecondaryPanes/Expressions.js
+++ b/src/components/SecondaryPanes/Expressions.js
@@ -57,6 +57,7 @@ class Expressions extends Component<Props, State> {
   }
 
   clear = () => {
+    // some useful code
     this.setState(() => {
       this.props.clearExpressionError();
       return { editing: false, editIndex: -1, inputValue: "" };
@@ -207,6 +208,8 @@ class Expressions extends Component<Props, State> {
   renderExpressionEditInput(expression: Expression) {
     const { expressionError } = this.props;
     const { inputValue, editing } = this.state;
+
+    /// is this doing the right thing?
[...]

PR #5678, opened 2018-03-15
Expression syntax errors are cleared on save

marcogallo566:
Fixes Issue: #5204 

Here's the Pull Request Doc

### Summary of Changes

* Now the wrong expression is not cleaned
* The red borders now are displaying only when there is an error

Feel free to improve it!

### GIF
dpfqqfod2x

### DIFF
--- a/src/components/SecondaryPanes/Expressions.js
+++ b/src/components/SecondaryPanes/Expressions.js
@@ -121,16 +121,22 @@ class Expressions extends Component<Props, State> {
   ) => {
     e.preventDefault();
     e.stopPropagation();
+
     this.props.updateExpression(this.state.inputValue, expression);
   };
 
   handleNewSubmit = async (e: SyntheticEvent<HTMLFormElement>) => {
     const { inputValue } = this.state;
-    this.props.clearExpressionError();
     e.preventDefault();
     e.stopPropagation();
+
+    this.props.clearExpressionError();
     await this.props.addExpression(this.state.inputValue);
-    this.setState({ editing: false, editIndex: -1, inputValue: this.props.expressionError ? inputValue : "" });
+    this.setState({
+      editing: false,
+      editIndex: -1,
+      inputValue: this.props.expressionError ? inputValue : ""
+    });
   };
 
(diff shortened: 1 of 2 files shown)


src/components/SecondaryPanes/Expressions.js
```diff
@@ -125,10 +125,16 @@ class Expressions extends Component<Props, State> {
   };
 
   handleNewSubmit = async (e: SyntheticEvent<HTMLFormElement>) => {
+    const { inputValue } = this.state;
+    this.props.clearExpressionError();
```
jasonLaster: hmm, I would prefer not firing this unless we need to

marcogallo566: I used this because I want to be sure that i cleaned the red borders when I'm writing a correct expression, but also permit to do an animation every time I write a wrong expression (before if I wrote 2 consecutives wrong expression, the second one didn't have an animation).

jasonLaster: @wldcordeiro mind taking a look at this?

src/components/SecondaryPanes/Expressions.js
```diff
@@ -125,10 +125,16 @@ class Expressions extends Component<Props, State> {
...
     e.preventDefault();
     e.stopPropagation();
-    this.props.addExpression(this.state.inputValue);
-    this.setState({ editing: false, editIndex: -1, inputValue: "" });
+    await this.props.addExpression(this.state.inputValue);
+    if (!this.props.expressionError) {
+      this.setState({ editing: false, editIndex: -1, inputValue: "" });
+    } else {
+      this.setState({ editing: false, editIndex: -1, inputValue: inputValue });
```
darkwing: Can we use a ternary here?

```js
this.setState({ editing: false, editIndex: -1, inputValue: this.props.expressionError ? inputValue : "" });
```

marcogallo566: Yes, it's much better, thank you!

darkwing:
The Travis errors seem unrelated, so kicking again

Issue #5447, opened 2018-02-16
[SourceTree] directory root can cause the tree to show no sources


jasonLaster:
When a directory root does not match, the user sees a message "No Sources..." which can be very confusing. Confession, i was stumped for 4 days when i saw this!

Here are a couple thoughts:

1. The user can add a root for one domain and then debug another page and nothing will match. Do we want to key roots by domain?
2. If nothing matches, should we clear the root so that we show everything? Should we show a better error message and let the user clear it?

<img src='https://shipusercontent.com/0edac5885beeaa987d4590887a8ef3d1/Screen%20Shot%202018-02-16%20at%209.45.43%20AM.png' title='Screen Shot 2018-02-16 at 9.45.43 AM.png' width=1425>

darkwing:
I think we should maybe do both?  Show a message that there are no matches and then show the entire tree.

lukaszsobek:
Maybe remember which root was set for a given url and show the full tree for other urls. However, this would require us to store the settings and could lead to bloat when ancient url items start to haunt our local storage.

jasonLaster:
Hmm, @darkwing how would that work?

jasonLaster:
@darkwing i think we should consider this blocked by #5448

darkwing:
Agreed; I think we can/should add a bottom bar to the sources pane, as we have for outline alphabetization.  We can have a "Clear directory root" button there.

PR #5682, opened 2018-03-15
WIP: Show project root based on sources count

darkwing:
Fixes Issue: #5447 

This is an early WIP to show progress, get feedback, and allow me to quickly check out other PRS.

## Summary
- If a custom project root is set and that root exists in the sources tree, show only sources under that root.
- If a custom project root is set and that root does *not* exist in the sources tree, show the full tree.

## ToDo
- [x] Ensure functionality works for `createTree` (initial display of the tree upon render)
- [ ] Ensure functionality works for `updateTree`
- [ ] Test(s)

### DIFF
--- a/src/actions/ui.js
+++ b/src/actions/ui.js
@@ -162,11 +162,15 @@ export function clearProjectDirectoryRoot() {
 export function setProjectDirectoryRoot(newRoot: string) {
   return ({ dispatch, getState }: ThunkArgs) => {
     const curRoot = getProjectDirectoryRoot(getState());
+
+    // What is this for?  It's causing problems by appending successive root settings
+    /*
     if (newRoot && curRoot) {
       const temp = newRoot.split("/");
       temp.splice(0, 2);
       newRoot = `${curRoot}/${temp.join("/")}`;
     }
+    */
 
     dispatch({
       type: "SET_PROJECT_DIRECTORY_ROOT",
--- a/src/components/PrimaryPanes/SourcesTree.js
+++ b/src/components/PrimaryPanes/SourcesTree.js
@@ -145,20 +145,14 @@ class SourcesTree extends Component<Props, State> {
     // only when a new source is added
     if (nextProps.sources != this.props.sources) {
       this.setState(
-        /*
         updateTree({
           newSources: nextProps.sources,
           prevSources: sources,
           debuggeeUrl,
           projectRoot,
+          projectRootUsed: this.state.projectRootUsed,
           uncollapsedTree,
           sourceTree
-        })
-        */
-        createTree({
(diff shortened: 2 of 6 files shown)


darkwing:
The functionality here is working well (yay) but I removed `updateTree` usage completely, so we'll need to see what I'm doing wrong in incorporating that the fix there.

## `updateTree` Problem

So the problem here is that an odd root is being chosen when you go between sites; inexplicably the tree seems to add a random node as the root; a problem I haven't figured out yet.

src/components/PrimaryPanes/SourcesTree.js
```diff
@@ -144,6 +145,7 @@ class SourcesTree extends Component<Props, State> {
     // only when a new source is added
     if (nextProps.sources != this.props.sources) {
       this.setState(
+        /*
         updateTree({
```
darkwing: Ideally we can use this function but it's causing problems atm.

src/components/PrimaryPanes/SourcesTree.js
```diff
@@ -292,6 +300,7 @@ class SourcesTree extends Component<Props, State> {
     return (
       <div
         className={classnames("node", { focused })}
+        title={item.path}
```
darkwing: Used for debugging -- this can be removed.

darkwing:
@jasonLaster @lukaszsobek I'd love a sanity check here in the case I'm missing something really obvious

jasonLaster:
hmm, i think that was for when you set a root on a directory and then a new root on a descendent directory. @AnshulMalik would know better

src/components/PrimaryPanes/SourcesTree.js
darkwing: By doing more testing, I'm seeing that updateTree is a necessity, because directories from previous pages sometimes display when they shouldn't.

src/actions/ui.js
```diff
@@ -162,11 +162,15 @@ export function clearProjectDirectoryRoot() {
 export function setProjectDirectoryRoot(newRoot: string) {
   return ({ dispatch, getState }: ThunkArgs) => {
     const curRoot = getProjectDirectoryRoot(getState());
+
+    // What is this for?  It's causing problems by appending successive root settings
+    /*
```
AnshulMalik: Let's say we have this as root: `/example.com/route`
So, our tree nodes will now be updated with the path which will start from `/route` since it's the root.

And now if you try to set project root again, then this functionality comes. So if the new root is `/route/ex`, so we'll append `ex` with the current root.
Hope it makes things clear. I'd love to see if we can make it simpler.

PR #5683, opened 2018-03-15
Fix message and formatting of Outline when no functions

darkwing:
At present, if no source is selected, the Outline tab shows `Loading...`, which is not good.  The formatting was also off, so I fixed that.


jasonLaster:
nice, do you have a before and after screenshot?

darkwing:
Before:
outlinebefore

After:

outlineaftetr

Issue #5561, opened 2018-02-28
[Scopes] Tracking issue for known original scope mapping tasks


loganfsmyth:
Not necessarily all required, but a general tracking issue for known tasks around the original scope work

#### Bugs
* [x] Exclude Flowtype bindings from original scope binding lists
* [x] Generate proper scope bindings when bindings are referenced before they are declared
* [x] Imported bindings that have been optimized out may cause errors
* [x] Handle Webpack's eval-source-map `devtool` option (generated code for map runs inside `eval()`)
* [x] Some mappings include multiple locations, and often the simple binary search in `source-map` finds one that isn't ideal. We should expose `allGeneratedPositionsFor` so we can decide for ourselves. [PR](https://github.com/devtools-html/devtools-core/pull/1014)
* [x] Resolving generated locations can sometimes result in ranges larger than a single variable. Most commonly, this is if the identifier is the first thing on its own line, because generally the mapped range includes the indentation whitespace
  * [x] `this` is not mapped in perf.html's `TabBar#mouseDownListener` #5967
  * [x] `Preview/index.js`'s `addExpression` (and other destructured/shorthand-referenced bindings) show as `(unmapped)` #5967
* [x] Investigate performance of mapping logic
  * [x] Optimize the perf of generated binding search [PR](https://github.com/devtools-html/debugger.html/pull/5919)
  * [x] Optimize the perf of original binding mapping by batching the lookup into a single call
    * [x] Implement batching in the devtools-core worker https://github.com/devtools-html/devtools-core/pull/1029
    * [x] Use the batched logic to optimize the debugger generated range lookup #5967
* [x] Bindings sometimes show as `(unmapped)` instead of `(optimized out)` for module imports. #5984
* [x] dependencies are unmapped in debugger.html `actions/tabs.js:closeTab` #6040
  * If a mapping has multiple ranges, we should search for mappings in all of them
* [x] `JSXIdentifier` nodes are not included in visitor.js's binding results #6047
* [x] this is mapped to `_this` in a class property in this [app](https://dbg-class-prop-this.glitch.me/) #6072 
* [x] Expression variable remapping also renames variables that are shadowed by re-declaration in the expression #6087
* [x] Consider bailing out of mapping if a range has too many applicable bindings #6099
* [x] 'this' doesn't map fully in arrow functions #6163
* [x] Handle single-identifier mapped ranges with more broad matching to addresshttps://github.com/Microsoft/TypeScript/issues/22834
[...]

PR #5684, opened 2018-03-15
Use original scope mappings to rewrite preview and watch expressions

loganfsmyth:
Related Issue: #5561

### Summary of Changes

* Adds expression-mapping logic to scope mappings
* Uses expression mappings to process watch and preview expressions


jasonLaster:
i pushed a fix to an unrelated intermittent

jasonLaster:
i landed the fixes to the console test, so we should see this go green in 20 mins

PR #5690, opened 2018-03-16
Handle out-of-order declarations in scopes.

loganfsmyth:
Refs Issue: #5561

### Summary of Changes

* Collects variable references on `enter` but doesn't try to find their related bindings until `exit`

Trying to associate the references with declarations on `enter` leads to errors if the references is earlier in the file that the declaration.


src/workers/parser/getScopes/visitor.js
```diff
@@ -576,25 +572,31 @@ const scopeCollectionVisitor = {
...
     } else if (t.isIdentifier(node) && t.isReferenced(node, parentNode)) {
-      const identScope = findIdentifierInScopes(state.scope, node.name);
-      if (identScope) {
-        identScope.bindings[node.name].refs.push({
-          type: "ref",
-          start: fromBabelLocation(node.loc.start, state.sourceId),
-          end: fromBabelLocation(node.loc.end, state.sourceId),
-          meta: buildMetaBindings(state.sourceId, node, ancestors)
-        });
+      let freeVariables = state.freeVariables.get(node.name);
```
jasonLaster: what's this doing? can you add some comments here for what a free variable is

loganfsmyth: Totally.

src/workers/parser/getScopes/visitor.js
```diff
@@ -133,11 +134,19 @@ export function parseSourceScopes(sourceId: SourceId): ?Array<ParsedScope> {
...
     scopeStack: []
   };
   t.traverse(ast, scopeCollectionVisitor, state);
 
+  for (const [key, freeVariables] of state.freeVariables) {
+    const binding = global.bindings[key];
+    if (binding) {
+      binding.refs = freeVariables.concat(binding.refs);
+    }
+  }
```
jasonLaster: does it also help to have a unit test for the visitor so that we have more direct tests of this case, which would be easier to fix than the mochitest

jasonLaster:
i think all of these PRs are blocked by unrelated travis fails... i'll fix them in #5689 which was in hindsight too narrow

PR #5692, opened 2018-03-16
Exclude Flowtype bindings from original scope list.

loganfsmyth:
Refs Issue: #5561

### Summary of Changes

* Exclude Flowtype bindings from original scope list.

### Test Plan


src/test/mochitest/browser_dbg-babel-scopes.js
```diff
@@ -253,6 +253,12 @@ add_task(async function() {
 
   await breakpointScopes(dbg, "non-modules", { line: 7, column: 2 }, []);
 
+  await breakpointScopes(dbg, "flowtype-bindings", { line: 6, column: 2 }, [
+    "Module",
+    ["aConst", '"a-const"'],
+    "root()"
+  ]);
```
jasonLaster: nice

src/workers/parser/getScopes/visitor.js
```diff
@@ -540,7 +540,10 @@ const scopeCollectionVisitor = {
           state.sourceId
         );
       });
-    } else if (t.isImportDeclaration(node)) {
+    } else if (
+      t.isImportDeclaration(node) &&
+      (!node.importKind || node.importKind === "value")
+    ) {
```
jasonLaster: do we also want a unit test for this too? 

it should be simpler and would give us some quicker feedback than the mochitest, which is also helpful but less direct

jasonLaster:
Hmm, maybe this is legit

PR #5702, opened 2018-03-16
Render a toggle for original/generated scopes when available.

loganfsmyth:
Refs Issue: #5561

### Summary of Changes

* Renders "Show generated/original scope" toggle in scope list

original-toggle


src/components/SecondaryPanes/Scopes.js
```diff
@@ -59,27 +66,45 @@ class Scopes extends PureComponent<Props, State> {
...
+          {hasOriginalScopes ? (
+            <div className="scope-type-toggle">
+              <a
+                href=""
+                onClick={e => {
+                  e.preventDefault();
+                  this.setState({ showOriginal: !showOriginal });
+                }}
+              >
+                Show {showOriginal ? "generated" : "original"} scope
```
jasonLaster: hmm, i think we should have two strings in `assets/panel/debugger.properties` not a big deal... See 115 for the L10.getStr api

src/components/SecondaryPanes/Scopes.js
```diff
@@ -112,12 +137,27 @@ export default connect(
...
-      isLoading: pending,
+      isLoading,
       why: getPauseReason(state),
-      frameScopes: frameScopes
+      frameScopes,
+      generatedFrameScopes,
+      hasOriginalScopes:
+        isLoading || !generatedFrameScopes || !frameScopes
+          ? false
+          : generatedFrameScopes.actor !== frameScopes.actor
```
jasonLaster: hmm, this seems like something that we should have in `reducers/pause` so that the mochitests can also listen for it to

darkwing:
@jasonLaster Do we want to hide this behind a pref?  Is this something most users will want?

jasonLaster:
@darkwing good point. this should use the `mapScopes` feature flag

src/components/SecondaryPanes/Scopes.js
loganfsmyth: Sounds good.

loganfsmyth:
@darkwing @jasonLaster The toggle only shows up if there are original scopes, so it should already be hidden based on that flag since there aren't original scopes when the flag is off.

assets/panel/debugger.properties
```diff
@@ -542,6 +542,14 @@ scopes.notAvailable=Scopes unavailable
 # for when the debugger is not paused.
 scopes.notPaused=Not paused
 
+# LOCALIZATION NOTE (scopes.toggleToGenerated): Scopes right sidebar pane label
+# for link to change scope list to show scopes for generated code.
```
flodolo: I think both comments need some rewording, but I'm not sure I can help a lot, not being familiar with the matter. E.g. "Link displayed in the right sidebar to display scopes for generated code" makes sense?

loganfsmyth: Haha, yeah that's what I get for copy/pasting half the text from the one above this :P

Issue #5602, opened 2018-03-05
[Stepping] Speed up stepping over await calls


jasonLaster:
We currently check to see if we're at an await expression and if we are we look ahead for the next statement's location so that we if the user wants to step to it, we can set a hidden breakpoint and resume.

This strategy has one key drawback, every time we think we might be at an await expression, we have to check and calculate the next statement. In big files, this means expensive traversals on the fly. We can speed this up by getting *all* of the await expressions while we extract the symbols.

While we're at it, we can get the previous statement location as well as the next. This is useful because in the case of babel, stepping to an await is just as bad as stepping off off of it.

Here is what the code will basically look like:

Note: that getting the sibling locations might be a bit challenging because we no longer have babel's path object.

```diff
diff --git a/src/workers/parser/getSymbols.js b/src/workers/parser/getSymbols.js
index 331cd6c..bb76c4c 100644
--- a/src/workers/parser/getSymbols.js
+++ b/src/workers/parser/getSymbols.js
@@ -42,6 +42,16 @@ export type FunctionDeclaration = SymbolDeclaration & {|
   parameterNames: string[]
 |};
 
+export type AwaitDeclaration = {|
+  locaton: BabelLocation,
+
+  // used to see if we're about to step to an await
+  previousLocation: BabelLocation,
+
+  // used to see where to step to
+  nextLocation: BabelLocation
+|}
+
 export type SymbolDeclarations = {
   classes: Array<ClassDeclaration>,
   functions: Array<SymbolDeclaration>,
@@ -50,7 +60,8 @@ export type SymbolDeclarations = {
   callExpressions: Array<SymbolDeclaration>,
   objectProperties: Array<SymbolDeclaration>,
   identifiers: Array<SymbolDeclaration>,
-  comments: Array<SymbolDeclaration>
+  comments: Array<SymbolDeclaration>,
+  awaits: Array<AwaitDeclaration>
 };
 
 function getFunctionParameterNames(path: SimplePath): string[] {
@@ -215,6 +226,17 @@ function extractSymbol(path: SimplePath, symbols) {
     }
   }
 
+  if (isAwaitExpression(path) || isYieldExpression(path)) {
+    const { start, end } = path.loc;
+    symbols.await.push({
+      location: { start, end },
+      previousLocation: {}
+      nextLocation: {}
+    });
+
+  }
+
+
   if (t.isIdentifier(path) && !t.isGenericTypeAnnotation(path.parent)) {
[...]

PR #5711, opened 2018-03-19
[Parser] (WIP) Speedup stepping await

nyrosmith:
Fixes Issue: #5602

- [x] parse awaits
- [ ] get previous location
- [ ] get next location

I still need to find a performant way to get previous & next location

### DIFF
--- a/src/workers/parser/getSymbols.js
+++ b/src/workers/parser/getSymbols.js
@@ -20,6 +20,12 @@ import getFunctionName from "./utils/getFunctionName";
 
 let symbolDeclarations = new Map();
 
+export type AwaitDeclaration = {|
+  locaton: BabelLocation,
+  previousLocation: BabelLocation,
+  nextLocation: BabelLocation
+|};
+
 export type ClassDeclaration = {|
   name: string,
   location: BabelLocation,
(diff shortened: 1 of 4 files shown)


src/workers/parser/getSymbols.js
```diff
@@ -265,10 +272,20 @@ function extractSymbol(path: SimplePath, symbols) {
       location: { start, end }
     });
   }
+
+  if (t.isAwaitExpression(path) || t.isYieldExpression(path)) {
+    const { start, end } = path.node.loc;
```
jasonLaster: I think we could try something like this for the left/right sibling.

So for this example: [ast](https://astexplorer.net/#/gist/cb2b761d20387a311a1d7187ce18c694/fec1187db78c84128d299cd816262c6255a6997c)
```js
async function foo(x, y) {
  console.log()
  await slower();
  return bar;
}
```

We can get `console.log` and `bar` locations by:

1. checking if the parent is an expression statement
2. using `simple-path.js` getSibling, to get the previous and next sibling.

```js
if (t.isExpressionStatement(path.parent)) {
  const prevNode = path.getSibling(-1);
  const prev = prevNode && t.isExpressionStatement(prevNode) ? prevNode.loc : null;
  const nextNode = path.getSibling(1);
  const next = nextNode && t.isExpressionStatement(nextNode) ? nextNode.loc : null;
}


We will want guards to make sure that we're:

1. handling cases where await is the first or last statement
2. we're not getting locations when await is not a child of a statement e.g `foo(1, await slow(), 2)`

jasonLaster:
Hey @nyrosmith. When do you think you'll have a chance to look at this? I'm tempted to close it temporarily...

jasonLaster:
closing, we can pick this up later. @nyrosmith feel free to re-open if you have a chance to work on it.

PR #5714, opened 2018-03-19
Reps 0.21.1

nchevobbe:
Fixes Issue #5713


nchevobbe:
Travis is complaining about the spread operator 🤔

jasonLaster:
it should be easy to bump circle's node here: https://github.com/devtools-html/debugger.html/blob/master/circle.yml#L1-L4

not sure about the mochitests... does this work with `yarn start`?

Issue #5649, opened 2018-03-09
Switch to pretty-quick


jasonLaster:
Our prettier git hook is great, but a bit slow. This will speed it up

This blog post explains it
https://www.wisdomgeek.com/web-development/using-prettier-format-javascript-code/

link
https://github.com/azz/pretty-quick

```diff
diff --git a/bin/prettier.js b/bin/prettier.js
index 694af47..4643e38 100644
--- a/bin/prettier.js
+++ b/bin/prettier.js
@@ -1,3 +1,5 @@
+// how much of this can be removed :)
+
 const path = require("path");
 const spawn = require("child_process").spawn;

diff --git a/package.json b/package.json
index 7388cb0..f350d37 100644
--- a/package.json
+++ b/package.json
@@ -20,6 +20,7 @@
     "flow": "flow",
     "eslint-check":
       "eslint --print-config .eslintrc.js | eslint-config-prettier-check",
+    /*  something */
     "prettier": "node bin/prettier.js",
     "license-check": "devtools-license-check",
     "links": "ls -l node_modules/ | grep ^l || echo 'no linked packages'",
@@ -157,6 +158,7 @@
     "webpack-visualizer-plugin": "^0.1.11",
     "workerjs": "github:jasonLaster/workerjs"
   },
+  /* what we have now */
   "lint-staged": {
     "*.js": ["prettier", "git add"],
     "src/*.js": ["prettier", "git add"],
     ```

PR #5719, opened 2018-03-19
added pretty quick

Rohanhacker:


Rohanhacker:
#5649

jasonLaster:
Thanks. How does it feel? Is it faster? How is the ux?

Rohanhacker:
@jasonLaster It feels same as prettier. I can't feel any significant difference. Maybe it'll faster in case of large number of files.

PR #5723, opened 2018-03-20
Reformat the getIcon code

darkwing:
Just a bit of code cleanup


src/components/PrimaryPanes/SourcesTree.js
```diff
@@ -72,6 +72,13 @@ type State = {
   highlightItems?: any
 };
 
+const sourceTypes = {
```
darkwing: Moving this up here so we aren't recreating the same object over and over

Issue #5352, opened 2018-02-08
[Outline] Add "Copy Function" context menu item to functions


darkwing:
@jasonLaster added a "Copy function" context menu item to the editor source pane.  It would be cool if we added a context menu for Outline pane functions to do the same thing.

codehag:
would this copy the function in the outline in their entirety? that would be interesting!

jasonLaster:
Nice idea!

AnshulMalik:
I am unclaiming this as this can be a good issue for new contributor, feel free to claim it :)

Here is my branch, it might help
https://github.com/AnshulMalik/debugger.html/commit/bfe6583fefd3ab98f0983a88d58c8b93d90f908b

AnshulMalik:
unclaim

escaton:
@darkwing could you close this issue please?

PR #5725, opened 2018-03-20
[Outline] Add "Copy Function" context menu item to functions

escaton:
Fixes Issue: #5352

### Summary of Changes

* Add "Copy Function" context menu item to functions in Outline tab

### Test Plan

- [x] Open any source contains functions
- [x] Open Outline tab
- [x] Open context menu on any function
- [x] Clicking "copy function" copies function body

### DIFF
--- a/src/components/PrimaryPanes/Outline.js
+++ b/src/components/PrimaryPanes/Outline.js
@@ -6,9 +6,19 @@
 
 import React, { Component } from "react";
 import { bindActionCreators } from "redux";
+import { showMenu } from "devtools-contextmenu";
 import { connect } from "react-redux";
+
+import { copyToTheClipboard } from "../../utils/clipboard";
+import { findFunctionText } from "../../utils/function";
+import { findClosestScope } from "../../utils/breakpoint/astBreakpointLocation";
+
 import actions from "../../actions";
-import { getSelectedSource, getSymbols } from "../../selectors";
+import {
+  getSelectedSource,
+  getSymbols,
+  getSelectedLocation
+} from "../../selectors";
 
 import "./Outline.css";
 import PreviewFunction from "../shared/PreviewFunction";
@@ -25,7 +35,11 @@ type Props = {
   selectLocation: ({ sourceId: string, line: number }) => void,
   selectedSource: ?SourceRecord,
   onAlphabetizeClick: Function,
-  alphabetizeOutline: boolean
+  alphabetizeOutline: boolean,
+  getFunctionText: Function,
+  getFunctionLocation: Function,
+  flashLineRange: Function,
+  selectedLocation: any
 };
 
 export class Outline extends Component<Props> {
@@ -37,6 +51,46 @@ export class Outline extends Component<Props> {
     const selectedSourceId = selectedSource.get("id");
     const startLine = location.start.line;
     selectLocation({ sourceId: selectedSourceId, line: startLine });
+  }
+
+  onContextMenu(event: SyntheticEvent<HTMLElement>, func: SymbolDeclaration) {
+    event.stopPropagation();
+    event.preventDefault();
+
+    const {
+      selectedSource,
+      getFunctionText,
+      getFunctionLocation,
+      flashLineRange,
+      selectedLocation
+    } = this.props;
+
+    const copyFunctionKey = L10N.getStr("copyFunction.accesskey");
+    const copyFunctionLabel = L10N.getStr("copyFunction.label");
+
+    if (!selectedSource) {
+      return;
+    }
+
+    const sourceLine = func.location.start.line;
+    const functionText = getFunctionText(sourceLine);
+    const copyFunctionItem = {
+      id: "node-menu-copy-function",
+      label: copyFunctionLabel,
+      accesskey: copyFunctionKey,
+      disabled: !functionText,
+      click: () => {
+        const { location: { start, end } } = getFunctionLocation(sourceLine);
+        flashLineRange({
+          start: start.line,
+          end: end.line,
+          sourceId: selectedLocation.sourceId
+        });
+        return copyToTheClipboard(functionText);
+      }
+    };
+    const menuOptions = [copyFunctionItem];
+    showMenu(event, menuOptions);
   }
 
   renderPlaceholder() {
@@ -61,6 +115,7 @@ export class Outline extends Component<Props> {
         key={`${name}:${location.start.line}:${location.start.column}`}
         className="outline-list__element"
         onClick={() => this.selectItem(location)}
+        onContextMenu={e => this.onContextMenu(e, func)}
       >
         <span className="outline-list__element-icon">λ</span>
         <PreviewFunction func={{ name, parameterNames }} />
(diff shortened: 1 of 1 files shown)


src/components/PrimaryPanes/Outline.js
```diff
@@ -162,7 +217,20 @@ export default connect(
...
+      selectedSource,
+      selectedLocation: getSelectedLocation(state),
+      getFunctionText: line =>
+        findFunctionText(
+          line,
+          selectedSource.toJS(),
+          getSymbols(state, selectedSource.toJS())
+        ),
+      getFunctionLocation: line =>
+        findClosestScope(getSymbols(state, selectedSource.toJS()).functions, {
```
escaton: There is flow error 
```
Cannot get `getSymbols(...).functions` because property `functions` is missing in null or undefined
```
Should i check the existence of `getSymbols(...)` right here, return `undefined` from `getFunctionLocation` call and again check it in `onContextMenu` ?

src/components/PrimaryPanes/Outline.js
```diff
@@ -39,6 +53,46 @@ export class Outline extends Component<Props> {
...
+        flashLineRange({
+          start: start.line,
+          end: end.line,
+          sourceId: selectedLocation.sourceId
+        });
+        return copyToTheClipboard(functionText);
+      }
+    };
+    const menuOptions = [copyFunctionItem];
+    showMenu(event, menuOptions);
```
escaton: In some other code i saw dispatching action instead of imperative call, not sure which is right?

escaton:
General question: there is non trivial logic about getting function source, which is copy-pasted from `EditorMenu`. Maybe i should do some refactoring here and move common logic in one place? What would you suggest?

src/components/PrimaryPanes/Outline.js
```diff
@@ -39,6 +53,46 @@ export class Outline extends Component<Props> {
...
+
+    const copyFunctionKey = L10N.getStr("copyFunction.accesskey");
+    const copyFunctionLabel = L10N.getStr("copyFunction.label");
+
+    if (!selectedSource) {
+      return;
+    }
+
+    const sourceLine = func.location.start.line;
+    const functionText = getFunctionText(sourceLine);
```
AnshulMalik: We can probably move this inside of click since the user might not always want to copy the function, so we can save some CPU cycles there.

src/components/PrimaryPanes/Outline.js
AnshulMalik: This looks good to me :)

src/components/PrimaryPanes/Outline.js
```diff
@@ -39,6 +53,46 @@ export class Outline extends Component<Props> {
...
+    const copyFunctionItem = {
+      id: "node-menu-copy-function",
+      label: copyFunctionLabel,
+      accesskey: copyFunctionKey,
+      disabled: !functionText,
+      click: () => {
+        const { location: { start, end } } = getFunctionLocation(sourceLine);
```
AnshulMalik: Do you think we can reuse `func.location` here?

src/components/PrimaryPanes/Outline.js
AnshulMalik: I'd recommend to save the result of `getSymbols` to a variable and check if it's not null before accessing `functions`.

Or if you like, you can fix `getSymbols` to always return `SymbolDeclarations`

escaton: Turns out i don't need this helper at all

src/components/PrimaryPanes/Outline.js
escaton: Unfortunately no, we need `functionText` to decide wether "copy function" is available

escaton: But maybe there is better way to check it, this code was copied from `EditorMenu`

src/components/PrimaryPanes/Outline.js
escaton: Oh, i missed it, thank you)

AnshulMalik:
`yarn jest -u` will fix circle ci

src/components/PrimaryPanes/Outline.js
AnshulMalik: Empty functions are really rare. So its very least likely that the text will be empty, that's why I suggest we move it inside.

And we can do the same at other places as well.

escaton: Sounds reasonable for this particular place.
Not sure we should do same optimization in `EditorMenu`

src/components/PrimaryPanes/Outline.js
```diff
@@ -68,19 +68,18 @@ export class Outline extends Component<Props> {
...
-
     const copyFunctionItem = {
       id: "node-menu-copy-function",
       label: copyFunctionLabel,
       accesskey: copyFunctionKey,
-      disabled: !functionText,
+      disabled: false,
       click: () => {
+        const startLine = func.location.start.line;
+        const endLine = func.location.end.line;
```
AnshulMalik: Destructuring would look nice here :)

escaton: you think so?
```javascript
const {
          start: { line: startLine },
          end: { line: endLine }
        } = func.location;

Comment (darkwing) [reply src/components/PrimaryPanes/Outline.js]: `EditorMenu.js` uses:

```js
const { location: { start, end } } = getFunctionLocation(sourceLine);
```

AnshulMalik: Hahaha, ok, let's keep it this way only :)

src/components/PrimaryPanes/Outline.js
```diff
@@ -68,19 +68,18 @@ export class Outline extends Component<Props> {
...
-
-    const sourceLine = func.location.start.line;
-    const functionText = getFunctionText(sourceLine);
-
     const copyFunctionItem = {
       id: "node-menu-copy-function",
       label: copyFunctionLabel,
       accesskey: copyFunctionKey,
-      disabled: !functionText,
+      disabled: false,
```
darkwing: If you open the "original source" tab for a given pretty-printed tab, the menu item was disabled, so I don't know that we can hardcode `false` -- the Editor suffers from the same issue.

darkwing:
Great work so far @escaton!!  Almost there! 👍

src/components/PrimaryPanes/Outline.js
escaton: @darkwing why use `getFunctionLocation(sourceLine);` when we already have `func.location` ?

src/components/PrimaryPanes/Outline.js
escaton: If i got it right, current way to get function text is limited for some cases, for example minified sources. So we can't be sure that it'll work 100% even we have list of detected functions in Outline.
I suppose for now i just revert this change and create an issue to investigate what can be improved in `findFunctionText` helper?

darkwing: Yeah, let's go back to `!functionText` to be consistent.

escaton: Done

Issue #5630, opened 2018-03-07
[QuickOpen] File Search rows can be quite tall


jasonLaster:
Copying @violasong's comments from #5034

I wonder if it would make sense to move the path to the same line as the filename to condense the rows. Some truncation would be needed, similar to #5019.

---

One easy step would be to use CSS to keep the path to one line, which would help quite a bit. I'm hesitant to have the path on the same line as that could become pretty tight.

I would be happy to explore a condensed option, but would have to see how it looks before we move forward. 

Also, for reference we always have the tooltip for the full path.

PR #5729, opened 2018-03-20
QuickOpen File Search rows can be quite tall

marcogallo566:
Fixes Issue: #5630

Here's the Pull Request Doc

### Summary of Changes

* Now the path is in the same line of the filename
* Added truncation to filename and path

Feel free to improve it!

### Screenshots
c02ca444-6b3a-4f34-8541-d21d546fe6e1

### DIFF
--- a/src/components/QuickOpenModal.js
+++ b/src/components/QuickOpenModal.js
@@ -335,16 +335,7 @@ export class QuickOpenModal extends Component<Props, State> {
     return results.map(result => {
       return {
         ...result,
-        title: this.renderHighlight(result.title, basename(newQuery), "title"),
-        ...(result.subtitle != null && !this.isSymbolSearch()
-          ? {
-              subtitle: this.renderHighlight(
-                result.subtitle,
-                newQuery,
-                "subtitle"
-              )
-            }
-          : null)
+        title: this.renderHighlight(result.title, basename(newQuery), "title")
       };
     });
   };
--- a/src/components/shared/ResultList.css
+++ b/src/components/shared/ResultList.css
@@ -59,7 +59,6 @@
   text-overflow: ellipsis;
   overflow: hidden;
   white-space: nowrap;
-  max-width: 50%;
 }
 
 .result-list li.selected .title {
@@ -77,7 +76,6 @@
   text-overflow: ellipsis;
   overflow: hidden;
   white-space: nowrap;
-  max-width: 50%;
 }
 
 .result-list.big li .subtitle {
--- a/src/components/shared/ResultList.css
+++ b/src/components/shared/ResultList.css
@@ -72,10 +72,7 @@
 .result-list.big li .subtitle {
   word-break: break-all;
   color: var(--theme-body-color-inactive);
-  margin-left: 5px;
-  text-overflow: ellipsis;
-  overflow: hidden;
-  white-space: nowrap;
+  margin-left: 15px;
 }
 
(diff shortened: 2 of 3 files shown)


darkwing:
@violasong Do you have an opinion on this?  I'm thinking the file name should be right-justified.

jasonLaster:
I find having the filename and path on the same line to be a bit dense...
I think i would prefer 2 lines, with css truncation so that we don't wrap lines

marcogallo566:
I tried this other solution, it's more readable.
Otherwise I move again the path under the filename, but with truncation, as @jasonLaster suggested.

Both `max-width: 45%` | Left 40%, right 50% | No `max-width`
------------ | ---------------- | -------------
45d31da8-16d7-421f-a3f1-fb732867ed73 | 24d1eb5a-633d-469e-9ab0-038a2fe403de | 4e435d51-cf54-410a-8e6f-f0b62aef63ab

violasong:
I'm currently working on a full mockup of quick search and other panels, and I'm pattering the rows after the Firefox awesomebar - here's the WIP:
image
More soon!

jasonLaster:
@violasong that looks really nice. Based on this suggestion, perhaps this PR should

1. left align the path
2. add comparable spacing

Going forward:

1. I love the tab icon
2. It would be really cool if we could drop SHAs e.g. `frameworks-a4d22334fsdf.js` became `frameworks.js`. Some will say they use it all the time, but I think if we expose it through a tooltip that will satisfy their use case while still preserving the usability of the modal.

marcogallo566:
This is the result now:

b3caf9b2-0cf6-45a6-9599-a0c5ab541056

jasonLaster:
Small nit, and maybe out of scope, but lets drop the filename from the URL on the right. That will really shorten it.

violasong:
Thanks @marcogallo566 for working on this! It's looking good. Will your patch preserve the white-on-dark-blue selection styling that's currently in the quick open on nightly?

Once I finish my mockups I can write up my full recommendations, including icons and visual tweaks, for a separate issue. 

> It would be really cool if we could drop SHAs e.g. frameworks-a4d22334fsdf.js became frameworks.js. Some will say they use it all the time, but I think if we expose it through a tooltip that will satisfy their use case while still preserving the common use case for the modal.

Yeah, this would be great! Maybe we could truncate it after the first couple characters of the hash like `frameworks-a4d...` (JS icon removes the need to have the .js in there)

jasonLaster:
> frameworks-a4d... 

Lets follow up on ellipsis and extensions in another issue. Have to think about this some more :)

marcogallo566:
Yes, it preserves the changes @violasong.

marcogallo566:
@jasonLaster this is the result without the right filename.

573e6f82-71fe-4176-b753-bb9e4f127394

jasonLaster:
Looking great. I think it might be nice to drop the first and last slash. They seem to add very little information.

marcogallo566:
Yes, it's better but I have one doubt : if there is a file in the root folder the URL would be empty...should I leave the first slash in case of empty URL?

This is an example:

a64f215d-4414-4bbf-af23-bbf8fd32b4d2

jasonLaster:
I think that is okay. @violasong's mockups did not have a slash and this is definitely softer on the eyes :)

darkwing:
Working on a quick update to this to prevent the following wrapping:

longfilenameproblem

Should be in shortly!

PR #5733, opened 2018-03-20
Add a proper type to  scopes Block nodes

### LINKED ISSUE #5426: [Scopes] Scope nodes are a bit too similar to "regular" variable nodes
The Scopes groups variables under "Block" nodes. Sometimes they are named, which is great.
But it's hard to distinguish those "blocks" node from the "variable" nodes. See: 

image

They are of the same color as variable names, and I feel like we could find something different for these top-level nodes.
We could start with having a different color, or maybe a background color, or an icon, … to they stand out from the other nodes.
This way you can quickly distinguish them when you have multiple block and variables expanded.

What do you think ? poke @violasong , @ericawright

nchevobbe:
Now that Reps 0.21.0 landed, we can use the BLOCK node type to style scopes block.

Here's how it looks: 

image

Fixes #5426

### DIFF
--- a/src/utils/pause/scopes/getScope.js
+++ b/src/utils/pause/scopes/getScope.js
@@ -3,6 +3,7 @@
  * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
 
 // @flow
+import { ObjectInspectorUtils } from "devtools-reps";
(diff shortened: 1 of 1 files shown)


nchevobbe:
Fixes #5426

Issue #5698, opened 2018-03-16
[QuickOpen] up arrow should not move the cursor to the front


jasonLaster:
When quick open is open and the user is iterating over the results with the arrow keys, up arrow moves the cursor to the front of the input and down arrow moves it to the bottom. the up/down arrow keys should not affect the cursor.

![](http://g.recordit.co/aAckOZyZlf.gif)

haoRchen:
I'm unable to reproduce the up arrow key moving cursor forward. Both up and down causes the cursor to disappear. 

![1 - reproducing issue](https://user-images.githubusercontent.com/19397242/37561127-0f29cc9c-2a1d-11e8-9002-5436083e826b.gif)

jasonLaster:
@haoRchen i'm refering to the input text cursor... if you type `abc` up arrow moves the cursor to c, down brings it to a.

haoRchen:
got it!

PR #5742, opened 2018-03-21
Fixes issue#5698

haoRchen:
Fixes Issue: #5698 

### Summary of Changes

* Prevent input cursor position from changing when pressing down/up arrow key in Quick Open.

### Screenshots/Videos 
Before:
5 5 - before fix behavior

After: Up/Down arrow key no longer moves input cursor
5 - fixed result


jasonLaster:
Nice! could you add a GIF w/ search results

haoRchen:
@jasonLaster  GIF has been updated. A before and after comparison with search results.

darkwing:
Awesome work @haoRchen !!

PR #5743, opened 2018-03-21
Provide a method to clear the root if no child sources are found

darkwing:
So this is an alternative approach to https://github.com/devtools-html/debugger.html/pull/5682 .  It was quick to put together so I thought I'd share.

Instead of reverting to showing the entire source tree, this keeps the "clear root" button but shows the "no sources" message.  Example:

clear

Positives:

-  There's no appearing/disappearing root button (consistency)
-  The code change to make this work is minimal

Negatives:

-  It is cumbersome to need to click that button?

### DIFF
--- a/src/components/PrimaryPanes/SourcesTree.js
+++ b/src/components/PrimaryPanes/SourcesTree.js
@@ -336,7 +336,13 @@ class SourcesTree extends Component<Props, State> {
 
     // The "sourceTree.contents[0]" check ensures that there are contents
     // A custom root with no existing sources will be ignored
-    if (isCustomRoot && sourceTree.contents[0]) {
+    if (isCustomRoot) {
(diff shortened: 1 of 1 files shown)


jasonLaster:
both options seem reasonable.

darkwing:
I can fix the test failures if we think this is a useful avenue.

assets/panel/debugger.properties
```diff
@@ -233,6 +233,9 @@ projectTextSearch.noResults=No results found
 # LOCALIZATION NOTE (sources.noSourcesAvailable): Text shown when the debugger
 # does not have any sources.
 sources.noSourcesAvailable=This page has no sources
+# LOCALIZATION NOTE (sources.noSourcesAvailableRoot): Text shown when the debugger
```
jasonLaster: Lets add a line space

PR #5744, opened 2018-03-21
Revert "Clear the directory root when navigating domains"

darkwing:
Reverts devtools-html/debugger.html#5740

We may need to remove this, as it clears upon refresh, not domain change;  we'll need to add a domain change check to ensure we clear only when we want to.


darkwing:
*sigh*, this grabbed stuff I didn't want reverted...ffs

PR #5749, opened 2018-03-21
Handle original source maps used on code running inside an `eval()`

loganfsmyth:
Refs Issue: #5561

### Summary of Changes

* Rework the processing of generated code so that it handles being run inside `eval()`
* Unrelated small fix: Include the final line in a file as part of the empty line list


src/utils/ast.js
```diff
@@ -45,6 +45,6 @@ export function findEmptyLines(
     return [];
   }
   const lineCount = selectedSource.text.split("\n").length;
-  const sourceLines = range(1, lineCount);
+  const sourceLines = range(1, lineCount + 1);
```
loganfsmyth: @jasonLaster This was the issue in the screenshot I sent you with the last line looking debuggable:
screen shot 2018-03-19 at 4 37 22 pm

darkwing: Awesome catch!

PR #5752, opened 2018-03-22
Add additional mocks

jasonLaster:
We were getting some exceptions in the Jest tests


darkwing:
Is there documentation that should get axed with this too?

jasonLaster:
sorry wrong, title

PR #5755, opened 2018-03-22
Removes toJS in more files

lukaszsobek:
- Removed a bunch more toJS
- cleaned up unused modifier references

### DIFF
--- a/src/actions/pause/continueToHere.js
+++ b/src/actions/pause/continueToHere.js
@@ -17,9 +17,9 @@ import type { ThunkArgs } from "../types";
 
 export function continueToHere(line: number) {
   return async function({ dispatch, getState }: ThunkArgs) {
-    const source = getSelectedSource(getState()).toJS();
+    const selectedSource = getSelectedSource(getState());
 
-    if (!isPaused(getState())) {
+    if (!isPaused(getState()) || !selectedSource) {
       return;
     }
 
@@ -36,7 +36,7 @@ export function continueToHere(line: number) {
       addHiddenBreakpoint({
         line,
         column: undefined,
-        sourceId: source.id
+        sourceId: selectedSource.id
       })
     );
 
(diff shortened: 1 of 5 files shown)


src/actions/file-search.js
```diff
@@ -86,7 +86,7 @@ export function searchContents(query: string, editor: Object) {
     const matches = await getMatches(
```
jasonLaster: can we remove `      const _modifiers = modifiers.toJS();`?

src/actions/file-search.js
```diff
@@ -86,7 +86,7 @@ export function searchContents(query: string, editor: Object) {
...
       query,
       selectedSource.get("text"),
-      _modifiers
+      modifiers
     );
```
jasonLaster: is it safe to pass an immutable object here? we might not have good coverage here so please test manually.

src/actions/pause/continueToHere.js
```diff
@@ -36,7 +34,7 @@ export function continueToHere(line: number) {
       addHiddenBreakpoint({
         line,
         column: undefined,
-        sourceId: source.id
+        sourceId: getSelectedSource(getState()).id
```
jasonLaster: if there isn't a source, this will fail.

Perhaps we get the source above and if there isn't one we return early?

src/actions/file-search.js
lukaszsobek: I think the tests threw when I did this last time, will retest to be sure.

src/actions/file-search.js
lukaszsobek: This shouldn't be here iirc, will check.

jasonLaster:
getting closer. just some flow things
https://circleci.com/gh/devtools-html/debugger.html/11420?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

PR #5761, opened 2018-03-22
Avoid listing workers if connected to Fennec v59

juliandescottes:
This is a client side workaround for fixing connection to Fennec v59, for the issue mentioned on [Bug 1443550](https://bugzilla.mozilla.org/show_bug.cgi?id=1443550)

It assumes that the root issue will be fixed after 59. We get device information and check the application type and the version. If we are on Fennec 59, we do not attempt to fetch workers.

Another possible workaround is to use Promise.race to timeout the listWorkers call after N seconds. I will submit another PR for this second approach and we can decide which one we like best.

### DIFF
--- a/src/client/firefox/commands.js
+++ b/src/client/firefox/commands.js
@@ -322,9 +322,9 @@ async function fetchSources() {
 
 /**
  * Temporary helper to check if the current server will support a call to
- * listWorkers. On Fennec 59 or older, the call will silently crash and prevent
+ * listWorkers. On Fennec 60 or older, the call will silently crash and prevent
  * the client from resuming.
- * XXX: Remove when FF59 for Android is no longer used or available.
+ * XXX: Remove when FF60 for Android is no longer used or available.
  *
  * See https://bugzilla.mozilla.org/show_bug.cgi?id=1443550 for more details.
  */
@@ -342,16 +342,14 @@ async function checkServerSupportsListWorkers() {
   }
 
   // We are only interested in Fennec release versions here.
-  // We assume that the server fix for Bug 1443550 will be uplifted to FF60.
-  // This means that we will not attempt to fetch workers for any early 60
-  // version such as 60.0b6.
+  // We assume that the server fix for Bug 1443550 will land in FF61.
   const version = description.platformversion;
-  return Services.vc.compare(version, "60.0") >= 0;
+  return Services.vc.compare(version, "61.0") >= 0;
 }
 
 async function fetchWorkers(): Promise<{ workers: Worker[] }> {
   // Temporary workaround for Bug 1443550
-  // XXX: Remove when FF59 for Android is no longer used or available.
+  // XXX: Remove when FF60 for Android is no longer used or available.
   const supportsListWorkers = await checkServerSupportsListWorkers();
 
   // NOTE: The Worker and Browser Content toolboxes do not have a parent
--- a/src/client/firefox/commands.js
+++ b/src/client/firefox/commands.js
@@ -31,6 +31,9 @@ import { makePendingLocationId } from "../../utils/breakpoint";
 
 import { createSource, createBreakpointLocation } from "./create";
 
+import { getDeviceFront } from "./fronts-device";
+import { Services } from "devtools-modules";
+
 let bpClients: BPClients;
 let threadClient: ThreadClient;
 let tabTarget: TabTarget;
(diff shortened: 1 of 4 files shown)


src/client/firefox/commands.js
```diff
@@ -317,6 +319,12 @@ async function fetchSources() {
   return sources.map(source => createSource(source, { supportsWasm }));
 }
 
+async function getDeviceDescription() {
+  const listTabsForm = await debuggerClient.listTabs();
+  const deviceFront = await getDeviceFront(debuggerClient, listTabsForm);
```
jasonLaster: could we add a method to the debuggerClient?

`debuggerClient.getDeviceFront()`

I think we do this in other places too

src/client/firefox/commands.js
```diff
@@ -329,6 +337,13 @@ async function fetchWorkers(): Promise<{ workers: Worker[] }> {
     return Promise.resolve({ workers: [] });
   }
 
+  const description = await getDeviceDescription();
+  const isFennec = description.apptype === "mobile/android";
+  const is59 = description.version.startsWith("59.");
```
jryans: Aren't all versions <=59 affected?

src/client/firefox/commands.js
jryans: Hmm, do you have an example of another place this is done?

I imagine the `debuggerClient` as for "interacting with the root actor", so `getDeviceFront` is not really related to that in my mental model.

jasonLaster: Hmm, was thinking of cases where clients isntantiate other clients or manage relation them:
- https://searchfox.org/mozilla-central/source/devtools/shared/client/thread-client.js#493
- https://searchfox.org/mozilla-central/source/devtools/shared/client/tab-client.js#67-70

jryans: Ah, I see... These are all examples from older, pre-protocol.js actors, whereas device actor uses protocol.js. We don't currently have an obvious API to jump between old style actor parent to new style actor child like this. At the moment, it's typically done manually in a manner similar to this PR:

https://searchfox.org/mozilla-central/search?q=new.*Front%5C(&case=true&regexp=true&path=

Seems like it would be worth exploring a natural way to do this for all actors in a follow up bug.

src/client/firefox/commands.js
```diff
@@ -317,6 +319,12 @@ async function fetchSources() {
   return sources.map(source => createSource(source, { supportsWasm }));
 }
 
+async function getDeviceDescription() {
+  const listTabsForm = await debuggerClient.listTabs();
```
jryans: `listTabs` can actually be expensive to call, so it's best to avoid it if we can.

The `Target` has the root form cached as `target.root`, and I believe that should work for your needs here.

jryans: Here's more detail: https://searchfox.org/mozilla-central/source/devtools/client/framework/target.js#282

It looks like it's not currently listed on the debugger's target type info, so you may need to add it there first.

src/client/firefox/commands.js
```diff
@@ -317,14 +320,48 @@ async function fetchSources() {
...
+    // devtools-modules, used when this code runs in a tab.
+    return true;
+  }
+
+  // We are only interested in Fennec release versions here.
+  // We assume that the server fix for Bug 1443550 will be uplifted to FF60.
+  // This means that we will not attempt to fetch workers for any early 60
+  // version such as 60.0b6.
+  const version = description.platformversion;
+  return Services.vc.compare(version, "60.0") >= 0;
```
juliandescottes: @jryans how do you feel about this comparison. Means pre-release v60 will not list workers either (which is not a big deal I think). But this way all 59.x versions should avoid the bug (assuming we can't uplift to 59).

juliandescottes:
Thanks for the comments! Cleaned up and updated, can you do a 2nd pass? @jryans @jasonLaster

juliandescottes:
Thinking a bit more about the change here, I think uplifting the debugger workaround will be easier and safer than uplifting a change for the package manifest for android. Adding a second changeset here also including FF60 in the versions where we avoid to list workers.

src/client/firefox/fronts-device.js
```diff
@@ -0,0 +1,15 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
+
+export function getDeviceFront() {
+  return {
+    getDescription: function() {
+      return {
+        // Return anything that will not match Fennec v59
```
jryans: Maybe this should say Fennec 60, now that you've changed the rest to mention 60 as well?

jryans:
I also opened #5768 to remove stray references to `listTabs` in the debugger, to help future patch authors avoid the trap.

jasonLaster:
I think this came from the launchpad which used it to get tabs and passed the tab along to the debugger.

jryans:
> I think this came from the launchpad which used it to get tabs and passed the tab along to the debugger.

Right, that makes sense. It used to be that you had to call `listTabs` even for a single tab, but these days we have a more efficient `getTab` to avoid the perf hit. So, `listTabs` should now only be used if you really do want a complete list of every tab for something like launchpad or about:debugging.

juliandescottes:
@jlast there was still one comment to address from @jryans  :/

jasonLaster:
Oh sorry, feel free to land a follow up a PR or "revert" and land the full change set.

PR #5763, opened 2018-03-22
Timeout listWorkers after 2000ms

juliandescottes:
Alternate approach to https://github.com/devtools-html/debugger.html/pull/5761

The advantage with this one is that if we manage to uplift the server fixes to 59, worker debugging will work even for this version.

Cons are that the timeout value is pretty arbitrary. 

cc @jasonLaster @jryans 

Any opinion on which workaround we prefer here? Or any other idea?


juliandescottes:
Thanks for the feedback @jryans closing this one in favor of #5761

PR #5769, opened 2018-03-23
followup to #5761: update comment about Fennec version

juliandescottes:


juliandescottes:
Simple comment followup, merging.

PR #5771, opened 2018-03-23
Pause points tweaks

jasonLaster:
### Summary of Changes

* fixes up two tests whose behavior changed (largely) for the better
* removes a check about adding a breakpoint w/ sliding bc pausepoints prevents users from adding bps in that spot
* added a `return foo()` check in pause points to prevent pausing at a return when there is a subsequent call

### DIFF
--- a/src/workers/parser/pausePoints.js
+++ b/src/workers/parser/pausePoints.js
@@ -1,5 +1,22 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
+
+// @flow
+
 import { traverseAst } from "./utils/ast";
 import * as t from "@babel/types";
+
+import type { AstLocation } from "./types";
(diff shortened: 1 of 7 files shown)


src/workers/parser/pausePoints.js
```diff
@@ -36,17 +36,24 @@ function formatNode(location, types) {
...
+    if (t.isCallExpression(node.argument)) {
+      state.push(
+        formatNode(node.loc.start, { breakpoint: false, stepOver: false })
+      );
+    } else {
+      state.push(
+        formatNode(node.loc.start, { breakpoint: true, stepOver: true })
+      );
+    }
+  }
```
darkwing: Should we try to shorten this code a bit, like:

```js
const shouldBreakStep = !t.isCallExpression(node.argument);

state.push(formatNode(node.loc.start, { breakpoint: shouldBreakStep, stepOver: shouldBreakStep }))
```

jasonLaster: yeah - formatNode can be cleaned

darkwing:
Fixed merge conflict, will merge when Travis goes green.

Issue #5753, opened 2018-03-22
[PausePoints] find some additional locations


jasonLaster:
Here are some examples of locations we should find pause points:

- [x] imports
- [x] decorators
- [x] exports
- [x] class definitions
- [x] class assignments
- [x] class and switch statements
- [x] chained function calls on different lines

### Input

Check out the [ast](https://astexplorer.net/#/gist/49d7da9282189cb960f6561aeece4ff6/beb3dfc220ab9e05e97f0b64e2856c1d263b6a24)!

```js
import {component} from "@angular/core"

@Component({
  selector: "app-root",
  templateUrl: './app.component.html',
  styleUrls: ['./app.component.css']
})

export class AppComponent {
  title = 'app'
}
```

### Output

```js
/*bs 1*/import {component} from "@angular/core"

/*bs 2*/@Component({
  selector: "app-root",
  templateUrl: './app.component.html',
  styleUrls: ['./app.component.css']
})

/*bs 3*/export /*bs 4*/class AppComponent {
  /*bs 5*/title = 'app'
/*bs 6*/}
```

---

### Code Sample

```diff
diff --git a/src/workers/parser/pausePoints.js b/src/workers/parser/pausePoints.js
index 3aa018f..496bfbf 100644
--- a/src/workers/parser/pausePoints.js
+++ b/src/workers/parser/pausePoints.js
@@ -13,7 +13,9 @@ const isControlFlow = node =>
   t.isForStatement(node) || t.isWhileStatement(node) || t.isIfStatement(node);

 const isAssignment = node =>
-  t.isVariableDeclaration(node) || t.isAssignmentExpression(node);
+  t.isVariableDeclaration(node) ||
+  t.isAssignmentExpression(node) ||
+  t.isStaticAssignment(node);

 const isImport = node => t.isImport(node) || t.isImportDeclaration(node);
 const isReturn = node => t.isReturnStatement(node);
@@ -37,9 +39,10 @@ function onEnter(node, ancestors, state) {

   if (
     isAssignment(node) ||
-    isImport(node) ||
+    isImport(node) || // why didn't this work?
     isControlFlow(node) ||
-    isReturn(node)
+    isReturn(node) ||
+    isExport(node)
   ) {
     state.push(
       formatNode(node.loc.start, { breakpoint: true, stepOver: true })
@@ -63,7 +66,7 @@ function onEnter(node, ancestors, state) {
     );
   }

-  if (t.isFunction(node)) {
[...]

PR #5777, opened 2018-03-23
Add additional pause point locations

jasonLaster:
Fixes Issue: #5753 

### Summary of Changes

- Adds a bunch of new syntax (case/switch)
- Adds a bunch of new scenarios like assignments with calls

### Tests

- I'll add a mochitest to document this behavior


jasonLaster:
### Todos

- [ ] add jsx support
- [ ] add a fallback for when pausepoints are not found

darkwing:
Are these to do's meant to be done in this PR?  Should we move that list to a tracking issue?

darkwing:
This all looks reasonable but going to mark as pr-wip until tests are fixed.

Issue #5695, opened 2018-03-16
[Preview] small syntax improvements


jasonLaster:
Our preview expression engine is really powerful and handles lots of cases. Here are some cases we don't support yet that we could:

- [x] destructuring #5706 (:runner: @marcogallo566)
- [x] console.log @bomsy 
- [x] Object properties with an assignment (:runner: @atwalg2)
- [x] object keys (:runner @bomsy )
- [x] booleans (this seems to work already)

#### Example 1: destructuring

hovering on `variables` should show the value

```js
const { variables, functions } = symbols;
```

#### Example 2: console.log

hovering on `console` and any of its properties should not do anything. 
```js
console.log(2)
console.warn('3');
```

#### Example 3: Object properties with an assignment

hover on `frameActor` should show the value of `frameId` and only highlight `frameActor` not the entire line.

```js
const params = frameId ? { frameActor: frameId } : {}
```

#### Example 4: object keys

highlighting on `sci-fi` or `1` should return values.

```js
const director = collection.genres['sci-fi'].movies[1].director
```

#### Example 5: booleans
~~highlighting `x` should show `true`~~ 
```js
const x = true
```

tommai78101:
One more thing you can add to the list is the following:

**Example 6: Evaluating expressions**

highlighting the text `GetValue(10)` or `GetValue(10);`, and then hover above it should return the result of `GetValue(10)`.

```
const GetValue = function(x) { return x + 10; }

GetValue(10);
```

jasonLaster:
Here is a glitch [example](https://dbg-obj-properties.glitch.me/) for testing.

PR #5780, opened 2018-03-25
add function to check for console object or props

bomsy:
Fixes Issue: #5695

### Summary of Changes

* added `isConsole` function
* hide

### Test Plan

Example test plan:

- [x] debug https://loud-bowl.glitch.me/
- [x] set breakpoint on line 26
- [x] refresh debuggee
- [x] hover over all `console`, `log`, `warn` and `group`


darkwing:
Part of me thinks we should check `console.`, so as not to match on `consoleShim.someCustomMethod()`, but I don't want to block on that.  Nice @bomsy !

Issue #5760, opened 2018-03-22
[QuickOpen] File search should use the directory root


jasonLaster:
The new debugger project root is amazing! It really simplifies the source tree when work working with large projects. We should try and use the directory root to simplify Quick Open:

1. we should only show files that match the root
2. we should truncate the file paths so you only see the part of the relevant part of the path

### Quick Open

![Screen Shot 2018-03-22 at 6.30.24 PM.png](https://shipusercontent.com/00ae4c856f25082093e9e25d0daeb3ac/Screen%20Shot%202018-03-22%20at%206.30.24%20PM.png)

### Directory Roots

![Screen Shot 2018-03-22 at 5.59.29 PM.png](https://shipusercontent.com/9b76e3b378ca00fc27bc69daa1778ee5/Screen%20Shot%202018-03-22%20at%205.59.29%20PM.png)

darkwing:
Whoa, great idea!

PR #5784, opened 2018-03-26
[QuickOpen] Filter file search if a project directory root is set

darkwing:
Fixes Issue: #5760

### Summary of Changes

* Limits the possible result set when doing a source search when there's a project root set

### Test Plan

2.  Set the project root to `/davidwalsh.name/wp-content/themes/punky/js`
3.  Search "m"
4.  You should only see files in that root

### To Do
- [x] Filter sources even when there is no query
- [x] Remove the directory root from the item subtitle


jasonLaster:
Can you share a screenshot?

src/components/QuickOpenModal.js
```diff
@@ -68,10 +70,18 @@ type GotoLocationType = {
...
     maxResults: 1000
   });
+
+  if (prefs.projectDirectoryRoot) {
+    results = results.filter(result =>
+      result.url.includes(prefs.projectDirectoryRoot)
+    );
+  }
+
+  return results;
```
jasonLaster: this is awesome!

I think we should also update:

`formatSources(getSources(state), prefs.projectDirectoryRoot)`  so that we can truncate the paths, which will really clean up the UI!

darkwing:
screen

src/components/QuickOpenModal.js
```diff
@@ -68,10 +70,18 @@ type GotoLocationType = {
...
 function filter(values, query) {
-  return fuzzyAldrin.filter(values, query, {
+  let results = fuzzyAldrin.filter(values, query, {
     key: "value",
     maxResults: 1000
   });
+
+  if (prefs.projectDirectoryRoot) {
+    results = results.filter(result =>
+      result.url.includes(prefs.projectDirectoryRoot)
```
AnshulMalik: How about `startsWith`?

jasonLaster: works for me

darkwing: We can't do `startsWith` because our projectDirectoryRoot is saved as `/davidwalsh.name/path/to/whatever` and the result.url is the true URL: `https://davidwalsh.name/path/to/whatever`.

Maybe we should save the root as true URL and that would simplify things.

src/components/QuickOpenModal.js
darkwing: Good shout -- added as a "to do" and marked the PR WIP

src/components/QuickOpenModal.js
jasonLaster: includes is good for now. i like keeping the tree data

src/components/QuickOpenModal.js
```diff
@@ -68,10 +70,18 @@ type GotoLocationType = {
 };
 
 function filter(values, query) {
-  return fuzzyAldrin.filter(values, query, {
+  let results = fuzzyAldrin.filter(values, query, {
     key: "value",
     maxResults: 1000
   });
+
+  if (prefs.projectDirectoryRoot) {
```
jasonLaster: we should be getting this from the store via connect

src/components/QuickOpenModal.js
```diff
@@ -68,10 +70,18 @@ type GotoLocationType = {
...
+    results = results.filter(result =>
```
jasonLaster: I feel like we should have a selector for getting the filtered sources... Not sure if getSources should do it, but that could be a next step based on the issue in discussion

src/utils/quick-open.js
```diff
@@ -49,6 +50,56 @@ export function parseLineColumn(query: string) {
...
+  if (sourcePath) {
+    const sourcePathSplit = sourcePath.split("/");
+
+    // Remove leading "/"
+    if (sourcePathSplit[0] === "") {
+      sourcePathSplit.shift();
+    }
+
+    // Make the title the file name
+    title = sourcePathSplit.pop().split("?")[0];
```
jasonLaster: we should have a helper for this... perhaps where we have getFileName

src/utils/quick-open.js
```diff
@@ -49,6 +50,56 @@ export function parseLineColumn(query: string) {
...
+
+    // Rebuild the sourcePath
+    let newSourcePath = sourcePathSplit.join("/");
+
+    if (projectDirectoryRoot != "") {
```
jasonLaster: This should be in a helper too

```js
truncateFilePathRoot(path, root)
```

this ofcourse could be tested in several different ways too

src/utils/quick-open.js
```diff
@@ -49,6 +50,56 @@ export function parseLineColumn(query: string) {
...
+    // Make the subtitle the remaining info
+    subtitle = endTruncateStr(newSourcePath, 100);
+  }
+
+  return {
+    value: sourcePath,
+    title,
+    subtitle,
+    id: source.get("id"),
+    url: source.get("url")
```
jasonLaster: lets help lukas and change this to `source.id` & `source.url`

jasonLaster:
I made a couple of tweaks to the code here:

- i added a new `getRelativeSources` selector which returns sources below the root with a new `relativeUrl` field
- updated formatSources so that it focuses on taking a relative source and returning a list node with title, subtitle, etc

src/reducers/types.js
```diff
@@ -31,7 +31,7 @@ export type PendingSelectedLocation = {
   column?: number
 };
 
-export type { SourcesMap } from "./sources";
+export type { SourcesMap, OuterState } from "./sources";
```
jasonLaster: We should be able to use `State` from line 15 here instead of `OuterState`

src/selectors/getRelativeSources.js
```diff
@@ -0,0 +1,43 @@
...
+export type RelativeSource = Source & {
+  +relativeUrl: string
+};
+
+function getRelativeUrl(url, root) {
+  if (!root) {
+    return dropScheme(url)
+      .split("/")
+      .slice(2)
+      .join("/");
```
jasonLaster: hmm, I was thinking it would be nice to include the domain/host... If we want to drop that too, it would be easier use `getSourcePath` or another helper...

src/selectors/test/getRelativeSources.spec.js
```diff
@@ -0,0 +1,24 @@
...
+    dispatch(actions.setProjectDirectoryRoot("/js"));
+    const filteredSources = getRelativeSources(getState());
+    const firstSource = filteredSources[0];
+
+    expect(firstSource.url).toEqual(
+      "http://localhost:8000/examples/js/scopes.js"
+    );
+    expect(firstSource.relativeUrl).toEqual("scopes.js");
+  });
+});
```
jasonLaster: I'd like to move this test to `actions/tests/ui.spec.js`

src/selectors/visibleBreakpoints.js
```diff
@@ -2,10 +2,14 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
 
+// @flow
+
 import { getBreakpoints } from "../reducers/breakpoints";
 import { getSelectedSource } from "../reducers/sources";
 import { isGeneratedId } from "devtools-source-map";
 
+import type { OuterState } from "../reducers/types";
```
jasonLaster: oops, this should have been `State`

jasonLaster:
Thanks for the help! This getting really close

src/selectors/getRelativeSources.js
darkwing: So if no root was set, you'd want the subtitle to be: "/davidwalsh.name/some/other/thing"?

jasonLaster: that's what I was thinking, because there could be several domains... does it matter, probably not.

I think I'm persuaded that the path is sufficient :)

src/reducers/ast.js
```diff
@@ -31,7 +31,7 @@ export type SymbolsMap = Map<string, Symbols>;
 export type EmptyLinesMap = Map<string, EmptyLinesType>;
 
 export type SourceMetaDataType = {
-  framework: ?string
+  framework: string | void
```
jasonLaster: so... this caused flow to crash...

jasonLaster:
### Old



### New

PR #5786, opened 2018-03-26
Restore sources tree icons

darkwing:
lighticons
darkicons


darkwing:
Updated screenshots using img tags:

darkicons2
lighticons2

Issue #5738, opened 2018-03-21
[Sources] Remove contents.get calls


jasonLaster:
We recently refactored our immutable Source object to be a Record instead of a Map. As a result, we can now replace `contents.get("id")` calls with `contents.id` calls. 

This refactor also removes some oversights in the code too :)

```diff
diff --git a/src/components/PrimaryPanes/SourcesTree.js b/src/components/PrimaryPanes/SourcesTree.js
index 078096d..2a90af5 100644
--- a/src/components/PrimaryPanes/SourcesTree.js
+++ b/src/components/PrimaryPanes/SourcesTree.js
@@ -163,28 +163,20 @@ class SourcesTree extends Component<Props, State> {
 
   selectItem = item => {
     if (!nodeHasChildren(item)) {
-      this.props.selectLocation({ sourceId: item.contents.get("id") });
+      this.props.selectLocation({ sourceId: item.contents.id });
     }
   };
 
   getPath = item => {
-    const { sources } = this.props;
-    const obj = item.contents.get && item.contents.get("id");
-
-    let blackBoxedPart = "";
-
-    if (
-      typeof obj !== "undefined" &&
-      sources.has(obj) &&
-      sources.get(obj).get("isBlackBoxed")
-    ) {
-      blackBoxedPart = "update";
+    if (nodeHasChildren(item)) {
+      return item.path;
     }
 
+    const blackBoxedPart = item.contents.isBlackBoxed ? "update" : "";
     return `${item.path}/${item.name}/${blackBoxedPart}`;
   };
 
-  getIcon = (sources, item, depth) => {
+  getIcon = (item, depth) => {
     const { debuggeeUrl, projectRoot } = this.props;
 
     if (item.path === "/Webpack") {
@@ -205,9 +197,8 @@ class SourcesTree extends Component<Props, State> {
     }
 
     if (!nodeHasChildren(item)) {
-      const obj = item.contents.get("id");
-      const source = sources.get(obj);
-      if (source && source.get("isBlackBoxed")) {
+      const source = item.contents;
+      if (source && source.isBlackBoxed) {
         return <img className="blackBox" />;
       }
       const sourceType = {
@@ -239,7 +230,7 @@ class SourcesTree extends Component<Props, State> {
     const menuOptions = [];
 
     if (!isDirectory(item)) {
-      const source = item.contents.get("url");
+      const source = item.contents.url;
       const copySourceUri2 = {
         id: "node-menu-copy-source",
[...]

PR #5788, opened 2018-03-26
Removes .get - Part 1

lukaszsobek:
Fixes Issue: #5738

- First part of removing .get calls
- 1 updated test with Immutable removed


jasonLaster:
Wow!

Issue #5659, opened 2018-03-12
[CallStack] map display names


jasonLaster:
We currently map call stack frame locations, but we don't map the frame display names. 

This causes some surprising results when the line points to the current function, but the name refers to the generated function.

![](https://user-images.githubusercontent.com/7321311/36925581-890136e6-1e41-11e8-8df3-e10b4eda9142.png)

---

A good way to solve this, is to extend `mapFrames` so that the action maps both the locations and the display names. This way, when we pause, `getCallStackFrames` can show the original display names in the call stack panel.

Here is a [branch](https://github.com/devtools-html/debugger.html/compare/master...jasonLaster:map-frames)

PR #5790, opened 2018-03-27
Map frame display names

calebstdenis:
Fixes Issue: #5659

### Summary of Changes

* 756fedb8955a002bb2b11cb5750dd62d598fea05: When pausing in a generated source, map call stack frame names to the closest corresponding function in the original source. This gets rid of cryptic function names that show up in generated source.
* c418541a45c7743f95baf3b423ae1c6c072889fe Slight update to the Babel async grouping to make things a little easier to read than this  (see how `slowest` and `slower` appear twice): 
image

### Test Plan
1. Go to http://firefox-dev.tools/debugger-examples/examples/babel/
2. Set breakpoint at entry.js:23
3. Click "asyncExample" button. 
4. Expect second screenshot below

### Screenshots
image

image

### DIFF
--- a/src/actions/ast.js
+++ b/src/actions/ast.js
@@ -11,6 +11,7 @@ import {
   isPaused
 } from "../selectors";
 
+import { mapFrames } from "./pause";
 import { setInScopeLines } from "./ast/setInScopeLines";
 import {
   getSymbols,
@@ -58,6 +59,10 @@ export function setSymbols(sourceId: SourceId) {
       source: source.toJS(),
       [PROMISE]: getSymbols(source.id)
     });
+
+    if (isPaused(getState())) {
+      await dispatch(mapFrames());
+    }
 
     await dispatch(setPausePoints(sourceId));
     await dispatch(setSourceMetaData(sourceId));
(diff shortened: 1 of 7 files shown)


src/actions/ast.js
```diff
@@ -59,6 +60,10 @@ export function setSymbols(sourceId: SourceId) {
       [PROMISE]: getSymbols(source.id)
     });
 
+    if (isPaused(getState())) {
+      await dispatch(mapFrames());
```
calebstdenis: There's a little bit of unnecessary work going on here. Not sure if it's enough to matter.

src/actions/pause/mapFrames.js
```diff
@@ -26,12 +28,32 @@ function updateFrameLocations(
...
+      return frame;
+    }
+    const originalFunction = findClosestFunction(
+      symbols.functions,
+      frame.location
+    );
+    if (!originalFunction) {
+      return frame;
+    }
+    const originalDisplayName = originalFunction.name;
```
jasonLaster: lets add some line spaces between some of this stuff

src/actions/ast.js
jasonLaster: I think that's fine. looks good to me!

jasonLaster:
The mapDisplayName work reminds me of this saying

> There are two ways of constructing a software design: One way is to make it so simple that there are obviously no deficiencies, and the other way is to make it so complicated that there are no obvious deficiencies. The first method is far more difficult.

PR #5793, opened 2018-03-27
Fix quick open line format

darkwing:
linenumbers


jasonLaster:
I think we want to right align the line numbers and drop the colon

darkwing:
I think it looks way better because:

1.  The `{file}:{line number}` syntax is used elsewhere and devs are used to seeing file numbers presented that way

2.  On a large screen, there's eye fatigue looking left to see the function name and then tracing the same row to see the line number.

I'll revert to previous layout though.

darkwing:
Since this is undesirable it's probably just best to start over.

jasonLaster:
I can see that. I think for me it is a question of importance. Line numbers are a lot less important

PR #5799, opened 2018-03-27
Fix alignment of line numbers in quick open

darkwing:
PR https://github.com/devtools-html/debugger.html/pull/5729 introduced a bug where alignment of line numbers was no longer to the right.  This PR fixes said bug.

smallopen
bigopen

### DIFF
--- a/src/components/shared/ResultList.css
+++ b/src/components/shared/ResultList.css
@@ -28,6 +28,10 @@
   border-bottom: 1px solid var(--theme-splitter-color);
 }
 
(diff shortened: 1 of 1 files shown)


src/components/shared/ResultList.css
```diff
@@ -28,6 +28,10 @@
   border-bottom: 1px solid var(--theme-splitter-color);
 }
 
+.result-list.small li {
+  justify-content: space-between;
```
jasonLaster: nice

Issue #3800, opened 2017-08-28
Framework Features


jasonLaster:
- [x] 1. Highlighting framework frames in the call stack. More work to do here.
- [ ] 2. Show component names in the call stack. React team excited about this. Right now we show filenames. They want to map it to something semantic to the framework. Would be unique. #3792
- [x] 3. Showing framework components in the source tree.
- [x] 4. Previewing library objects on hover. Custom formatters with a nice navigation UI. (Sam)
- [x] 5. Show component instances in the sidebar. New pane showing all instances of a component.

jasonLaster:
###  5. Show component instances in the sidebar. New pane showing all instances of a component.


```diff

diff --git a/src/components/SecondaryPanes/index.js b/src/components/SecondaryPanes/index.js
index a801f64..8afc2e5 100644
--- a/src/components/SecondaryPanes/index.js
+++ b/src/components/SecondaryPanes/index.js
@@ -142,6 +142,26 @@ class SecondaryPanes extends Component<Props> {
     ];
   }
 
+  getComponentItem() {
+    if (!sourceMetaData.isReactComponent) {
+      return null;
+    }
+
+    const componentName = //borrow preview's extra field which has the displayName
+    const preview = // borrow what the preview does w/ the Object Inspector
+
+    return {
+      header: componentName
+      className: "scopes-pane",
+      component: preview,
+      opened: prefs.scopesVisible,
+      onToggle: opened => {
+        prefs.scopesVisible = opened;
+      },
+      shouldOpen: isPaused
+    };
+  }
+
   getScopeItem() {
     const isPaused = () => !!this.props.pauseData;
 
@@ -243,6 +263,10 @@ class SecondaryPanes extends Component<Props> {
     const items: Array<SecondaryPanesItems> = [];
 
     if (!this.props.horizontal) {
+      items.unshift(this.getComponentItem());
+    }
+
+    if (!this.props.horizontal) {
       items.unshift(this.getScopeItem());
     }
 
```

jasonLaster:
It's absolutely okay to have several ppl working on it as long as we're talking and sharing what we learn :)

decadef20:
I am trying to finish #5

jasonLaster:
here are some notes on how we get the react preview. 

```diff
diff --git a/src/components/Editor/Preview/Popup.js b/src/components/Editor/Preview/Popup.js
index 89c0177..abca23b 100644
--- a/src/components/Editor/Preview/Popup.js
[...]

PR #5801, opened 2018-03-27
[Preview] React Component pane

mmcote:
Fixes Issue: #3800 -- 5: Show component instances in the sidebar.

### Summary of Changes

* Added new FrameworkComponent.
* Separated out function to get extra props.

### Test Plan

* Currently validated through the browser, will add tests.

### Final Screenshot
screen shot 2018-03-27 at 10 50 29 am

### DIFF
--- a/src/actions/preview.js
+++ b/src/actions/preview.js
@@ -24,7 +24,7 @@ import { getMappedExpression } from "./expressions";
 import { isEqual } from "lodash";
 
 import type { ThunkArgs } from "./types";
-import type { Range, Position } from "../types";
+import type { Frame, Range, Position } from "../types";
 
 async function getReactProps(evaluate) {
   const reactDisplayName = await evaluate(
@@ -93,7 +93,7 @@ function isInvalidTarget(target: HTMLElement) {
 export function getExtra(
   expression: string,
   result: Object,
-  selectedFrame: any
+  selectedFrame: Frame
 ) {
   return async ({ dispatch, getState, client, sourceMaps }: ThunkArgs) => {
     const extra = await getExtraProps(expression, result, expr =>
--- a/src/actions/preview.js
+++ b/src/actions/preview.js
@@ -90,6 +90,20 @@ function isInvalidTarget(target: HTMLElement) {
   return invalidTarget || invalidToken || invaildType;
 }
 
+export function getExtra(
+  expression: string,
+  result: Object,
+  selectedFrame: any
+) {
+  return async ({ dispatch, getState, client, sourceMaps }: ThunkArgs) => {
+    const extra = await getExtraProps(expression, result, expr =>
+      client.evaluateInFrame(selectedFrame.id, expr)
+    );
+
+    return extra;
+  };
+}
+
 export function updatePreview(target: HTMLElement, editor: any) {
   return ({ dispatch, getState, client, sourceMaps }: ThunkArgs) => {
     const tokenPos = getTokenLocation(editor.codeMirror, target);
@@ -185,8 +199,8 @@ export function setPreview(
           return;
         }
 
-        const extra = await getExtraProps(expression, result, expr =>
-          client.evaluateInFrame(selectedFrame.id, expr)
+        const extra = await dispatch(
+          getExtra(expression, result, selectedFrame)
         );
 
         return {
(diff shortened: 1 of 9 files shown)


jasonLaster:
Awesome!

where is `t` coming from. It looks like the component is `App`

mmcote:
@jasonLaster, yeah for sure, I was curious about this as well, currently I am just utilizing what was there for the popup, which returns the same name, as they are both using the `getReactProps`. Which 
essentially is just calling `this._reactInternalFiber.type.name` or `this._reactInternalInstance.getName()` depending on the react version. I will look into each react object and find a way to return the proper name. 

screen shot 2018-03-27 at 11 18 01 am

mmcote:
^^^ Change that, I am not looking for a new place to get the name, just looking for why I'm currently getting t, and what needs to be changed for these statements to return the correct name.

jasonLaster:
sounds good

mmcote:
@jasonLaster, the reason the `t` was coming appearing was because the react applications I was testing this feature on were deployed react apps that have been minified for production. 

This was stated here, as they had the same problem: [https://github.com/facebook/react-native/issues/10550](https://github.com/facebook/react-native/issues/10550), and repeated here in the last comment: [https://stackoverflow.com/questions/43800784/get-component-name-in-react](https://stackoverflow.com/questions/43800784/get-component-name-in-react).

Tested this myself:
Development
screen shot 2018-03-27 at 12 35 10 pm

Production
screen shot 2018-03-27 at 12 33 52 pm

I will look if there is a way of getting the name in production.

jasonLaster:
I see, that's not a big deal. We have one other option too you know :)

If we know that, 

a) `this` is a react component, 
b) that the debug line is in a class we can use the symbols to get the closest class
c) that the debug line is not in a class, we can use the symbols table to get the closest function.

mmcote:
Oh awesome, okay I will look at implementing this as an alternative when not in a dev environment. :)

jasonLaster:
Sounds good, i'll see if we can merge this as is for now

src/actions/preview.js
```diff
@@ -90,6 +90,20 @@ function isInvalidTarget(target: HTMLElement) {
   return invalidTarget || invalidToken || invaildType;
 }
 
+export function getExtra(
+  expression: string,
+  result: Object,
+  selectedFrame: any
```
jasonLaster: `selectedFrame: Frame`

src/components/SecondaryPanes/FrameworkComponent.js
```diff
@@ -0,0 +1,92 @@
...
+
+import { ObjectInspector, ObjectInspectorUtils } from "devtools-reps";
+import { isReactComponent } from "../../utils/preview";
+
+const { createNode, getChildren } = ObjectInspectorUtils.node;
+const { loadItemProperties } = ObjectInspectorUtils.loadProperties;
+
+type Props = {
+  setPopupObjectProperties: (Object, Object) => void,
+  frame: any,
```
jasonLaster: `frame: Frame`

src/components/SecondaryPanes/index.js
```diff
@@ -151,6 +154,20 @@ class SecondaryPanes extends Component<Props> {
     };
   }
 
+  getComponentItem() {
+    const { extra: { react } } = this.props;
+
+    return {
+      header: react.displayName,
```
jasonLaster: could this be null?

src/components/SecondaryPanes/index.js
```diff
@@ -241,6 +258,10 @@ class SecondaryPanes extends Component<Props> {
       items.push(this.getCallStackItem());
       if (this.props.horizontal) {
         items.push(this.getScopeItem());
+
+        if (extra.react) {
```
jasonLaster: guess not :)

src/utils/prefs.js
```diff
@@ -19,6 +19,7 @@ if (isDevelopment()) {
   pref("devtools.debugger.ignore-caught-exceptions", false);
   pref("devtools.debugger.call-stack-visible", true);
   pref("devtools.debugger.scopes-visible", true);
+  pref("devtools.debugger.component-visible", true);
```
jasonLaster: we'll also have to add this to assets/prefs

PR #5804, opened 2018-03-27
Refactor pause commands

jasonLaster:


src/reducers/pause.js
```diff
@@ -278,8 +290,12 @@ export function getPauseReason(state: OuterState): ?Why {
   return state.pause.why;
 }
 
+export function getPauseCommand(state: OuterState): Command {
+  return state.pause && state.pause.command;
```
lukaszsobek: where are conditional properties when you need them...

PR #5806, opened 2018-03-28
Provide Tab Keyboard Navigation

darkwing:
Making source tabs navigatable by key is important for usability and accessibility.  This is my prototype code.

keyboardnavigation

## ToDo
- [x] Comment cleanup
- [x] `shortcuts.off` for both events in `componentWillUnmount`
- [x] Windows support - `CommandorControl` most likely

### DIFF
--- a/src/components/App.js
+++ b/src/components/App.js
@@ -132,8 +132,8 @@ class App extends Component<Props, State> {
     shortcuts.on("Escape", this.onEscape);
     shortcuts.on("Cmd+/", this.onCommandSlash);
 
-    shortcuts.on("Command+Left", this.goToPreviousTab);
-    shortcuts.on("Command+Right", this.goToNextTab);
+    shortcuts.on("CmdOrCtrl+Left", this.goToPreviousTab);
+    shortcuts.on("CmdOrCtrl+Right", this.goToNextTab);
   }
 
   componentWillUnmount() {
@@ -154,8 +154,8 @@ class App extends Component<Props, State> {
 
     shortcuts.off("Escape", this.onEscape);
 
-    shortcuts.off("Command+Left", this.goToPreviousTab);
-    shortcuts.off("Command+Right", this.goToNextTab);
+    shortcuts.off("CmdOrCtrl+Left", this.goToPreviousTab);
+    shortcuts.off("CmdOrCtrl+Right", this.goToNextTab);
   }
 
   goToNextOrPreviousTab(e, goNext) {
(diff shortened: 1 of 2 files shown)


src/components/App.js
```diff
@@ -141,6 +150,40 @@ class App extends Component<Props, State> {
...
+    const currentIndex = tabSources.indexOf(this.props.selectedSource);
+    if (currentIndex < 0) {
+      return;
+    }
+
+    // Calculate the next index
+    let nextIndex = 0;
+
+    // Next
+    if (offset) {
```
AnshulMalik: In my opinion, `offset` can be a bit confusing.
Can we separate this into two functions? that would make it more readable.

src/components/App.js
```diff
@@ -141,6 +154,40 @@ class App extends Component<Props, State> {
     shortcuts.off(L10N.getStr("gotoLineModal.key2"), this.toggleQuickOpenModal);
 
     shortcuts.off("Escape", this.onEscape);
+
+    shortcuts.off("CmdOrCtrl+Left", this.goToPreviousTab);
+    shortcuts.off("CmdOrCtrl+Right", this.goToNextTab);
+  }
+
+  goToNextOrPreviousTab(e, goNext) {
```
jasonLaster: nit, lets use this syntax `    onEscape = (_, e) => {`

src/components/App.js
```diff
@@ -141,6 +154,40 @@ class App extends Component<Props, State> {
...
+    // Calculate the next index
+    let nextIndex = 0;
+
+    // Next
+    if (goNext) {
+      nextIndex = currentIndex === tabSources.size - 1 ? 0 : currentIndex + 1;
+    } else {
+      // Previous
+      nextIndex = currentIndex === 0 ? tabSources.size - 1 : currentIndex - 1;
+    }
```
jasonLaster: I've found over time that this logic is better in an action:

1. the component logic is easier to understand (get the event and surrounding state and pass it to the action)
2. the action is easier to understand because it is not wrapped in a class and its clear how it will be fired
3. the action is easier to test independently
4. the action has access to `getState` which keeps the component from needing more props via connect

src/utils/tabs.js
```diff
@@ -11,7 +11,7 @@ import type { SourceRecord } from "../types";
 import type { SourceMetaDataType } from "../reducers/ast";
 import { isPretty } from "./source";
 
-type SourcesList = List<SourceRecord>;
+export type SourcesList = List<SourceRecord>;
```
jasonLaster: this type might be better in reducers/sources

jasonLaster:
This is really great! Nice work

darkwing:
Closing for now but referenced in the issue so that someone can drag it across the finish line!

PR #5809, opened 2018-03-28
Make getMappedExpression a real action and fix double-binding action creators

loganfsmyth:
Refs Issue: #5561

### Summary of Changes

* Convert `getMappedExpression` into a real action
* Avoid double-binding all of the actions in `dbg.actions`

### DIFF
--- a/src/utils/dbg.js
+++ b/src/utils/dbg.js
@@ -1,4 +1,3 @@
-import { bindActionCreators } from "redux";
 import * as timings from "./timings";
 import { prefs, features } from "./prefs";
 import { isDevelopment } from "devtools-config";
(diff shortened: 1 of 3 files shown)


src/utils/dbg.js
```diff
@@ -52,11 +51,9 @@ function _formatPausePoints(dbg, url) {
 
 export function setupHelper(obj) {
   const selectors = bindSelectors(obj);
-  const actions = bindActionCreators(obj.actions, obj.store.dispatch);
```
loganfsmyth: The actions here are already bound in https://github.com/devtools-html/debugger.html/blob/1d24d9875d7fa4d75804a5709f7afdbc4001fd28/src/utils/bootstrap.js#L55 and binding them again causes value-returning actions like `getMappedExpression` to try to re-dispatch their return value.

PR #5810, opened 2018-03-28
Removes getIn

lukaszsobek:
Where possible removes immutable `getIn` calls not needed due to switch to source records.

### DIFF
--- a/src/reducers/expressions.js
+++ b/src/reducers/expressions.js
@@ -40,20 +40,20 @@ function update(
       if (action.expressionError) {
         return state.set("expressionError", !!action.expressionError);
       }
-      return appendToList(state, ["expressions"], {
+      return appendExpressionToList(state, {
         input: action.input,
         value: null,
         updating: true
       });
     case "UPDATE_EXPRESSION":
       const key = action.expression.input;
-      return updateItemInList(state, ["expressions"], key, {
+      return updateExpressionInList(state, key, {
         input: action.input,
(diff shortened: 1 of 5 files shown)


src/reducers/expressions.js
```diff
@@ -111,8 +108,9 @@ function appendToList(
   value: any
 ) {
   const newState = state.updateIn(path, () => {
-    return state.getIn(path).push(value);
+    return state[path].push(value);
```
lukaszsobek: This has to be .get(path) because of flow complaining

src/reducers/expressions.js
```diff
@@ -124,7 +122,7 @@ function updateItemInList(
   value: any
 ) {
   const newState = state.updateIn(path, () => {
-    const list = state.getIn(path);
+    const list = state[path];
```
lukaszsobek: same here

src/reducers/expressions.js
```diff
@@ -96,38 +96,32 @@ function restoreExpressions() {
...
+function updateExpressionInList(
   state: Record<ExpressionState>,
-  path: string[],
   key: string,
   value: any
 ) {
-  const newState = state.updateIn(path, () => {
-    const list = state.getIn(path);
+  const newState = state.update("expressions", () => {
+    const list = state.get("expressions");
```
jasonLaster: why not `state.expressions`?

PR #5812, opened 2018-03-29
Document source map semantics

jasonLaster:
### Summary of Changes

* document some of the test cases we have in examples/babel
* include links to the source map visualization where helpful
* describe some of the mappings based on language semantics (calls, imports, etc) and different tool configurations
* try to make some opinions on what should happen

### Where should this document live

- I tried google doc first, but the links are bigger than what docs supports
- I thought about a new repo but it makes sense to have the doc next to the examples

### Next steps

- we might want to copy our tests into a new dedicated source map repo
- we might want more examples for calls
- we might want to do a hosted model of source map visualization that can read from our list of tests...


src/test/mochitest/examples/babel/source-maps-semantics.md
```diff
@@ -0,0 +1,49 @@
...
+
+![](https://shipusercontent.com/c0939aecadad12902646bd687b05840f/Screen%20Shot%202018-03-28%20at%2010.00.30%20PM.png)
+
+![](https://shipusercontent.com/fadbe965f836d304e84989f89529276e/Screen%20Shot%202018-03-28%20at%2010.07.57%20PM.png)
+![](https://shipusercontent.com/cc5ff07e169e2583e9e4e8859b912810/Screen%20Shot%202018-03-28%20at%2010.07.52%20PM.png)
+
+When Webpack is left to its own devices then dependencies are not mapped at all. In these cases, the original dependencies all map to the generated location `0,0`, which is equivelant to null.
+
+- example 1: ![webpack-es-modules]  
+- example 2: ![webpack-modules]
```
jasonLaster: @loganfsmyth what's the difference between es-modules and modules here? I always forget :)

we might want to put a readme in some of the examples.

loganfsmyth: Fair. The `es` one runs with no Babel plugins, and the non-es one runs with preset-env, so the `es` one still has ES20XX features in it.

PR #5819, opened 2018-03-29
Combines toJS calls in select.js

lukaszsobek:
Refactor due to possible performance gains in selectLocation and selectSpecificLocation, which should make the whole application feel snappier by about a quarter of a second each time any of these functions is called.

### DIFF
--- a/src/actions/sources/select.js
+++ b/src/actions/sources/select.js
@@ -98,8 +98,8 @@ export function selectLocation(location: Location) {
       return;
     }
 
-    const source = getSource(getState(), location.sourceId);
-    if (!source) {
+    const sourceRecord = getSource(getState(), location.sourceId);
+    if (!sourceRecord) {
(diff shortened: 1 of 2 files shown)


src/actions/sources/select.js
```diff
@@ -162,21 +164,22 @@ export function selectSpecificLocation(location: Location) {
...
     }
 
-    dispatch(addTab(source.toJS(), 0));
+    const source = sourceRecord.toJS();
 
+    dispatch(addTab(source, 0));
     dispatch({
       type: "SELECT_SOURCE",
-      source: source.toJS(),
+      source,
```
jasonLaster: I think I'd prefer `source: sourceRecord.toJS()` so it is clear why we're doing the serialization

jasonLaster:
@darkwing what do you think

PR #5823, opened 2018-03-29
Get breakpoint text

darkwing:
I wanted to show progress on this.  This:

- Removes the `snippet` element as it hasn't been used and we'll not be using it in the redesign
- Stores breakpoint text at a given line

## ToDo

- [x] Tests?
- [x]  Fix `syncBreakpoints` stuff

### DIFF
--- a/src/actions/breakpoints/addBreakpoint.js
+++ b/src/actions/breakpoints/addBreakpoint.js
@@ -20,8 +20,8 @@ export default async function addBreakpoint(
   breakpoint
 ) {
   const state = getState();
-
   const source = getSource(state, breakpoint.location.sourceId);
+
   const location = { ...breakpoint.location, sourceUrl: source.url };
   const generatedLocation = await getGeneratedLocation(
     state,
@@ -29,6 +29,8 @@ export default async function addBreakpoint(
     location,
     sourceMaps
   );
+
+  const generatedSource = getSource(state, generatedLocation.sourceId);
 
   assertLocation(location);
   assertLocation(generatedLocation);
@@ -53,17 +55,8 @@ export default async function addBreakpoint(
   const symbols = getSymbols(getState(), source);
   const astLocation = await getASTLocation(source, symbols, newLocation);
 
-  // generatedSource.text, in the case of pretty print or original tab
-  // returns the minified contents, which is why it's used
-  // when creating the "originalText"
-  const generatedSource = getSource(state, generatedLocation.sourceId);
-
-  const text = getTextAtPosition(source.text, location.line, location.column);
-  const originalText = getTextAtPosition(
-    generatedSource && generatedSource.text,
-    actualLocation.line,
-    actualLocation.column
-  );
+  const originalText = getTextAtPosition(source, location);
+  const text = getTextAtPosition(generatedSource, actualLocation);
 
   const newBreakpoint = {
     id,
--- a/src/actions/breakpoints/addBreakpoint.js
+++ b/src/actions/breakpoints/addBreakpoint.js
@@ -11,6 +11,7 @@ import {
 } from "../../utils/breakpoint";
 import { getSource, getSymbols } from "../../selectors";
 import { getGeneratedLocation } from "../../utils/source-maps";
+import { getTextAtPosition } from "../../utils/source";
 
 export default async function addBreakpoint(
   getState,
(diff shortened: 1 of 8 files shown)


src/actions/breakpoints/syncBreakpoint.js
```diff
@@ -78,6 +86,21 @@ export async function syncClientBreakpoint(
...
+    location.column,
+    location.column + 100
+  );
+
+  const originalText = source.text[location.line].slice(
+    location.column,
+    location.column + 100
+  );
+  */
+
```
jasonLaster: :)

jasonLaster:
This looks good sans syncBreakpoints!

src/actions/breakpoints/syncBreakpoint.js
```diff
@@ -78,6 +87,17 @@ export async function syncClientBreakpoint(
...
+  // generatedSource.text, in the case of pretty print or original tab
+  // returns the minified contents, which is why it's used
+  // when creating the "originalText"
+  const generatedSource = getSource(getState(), generatedLocation.sourceId);
+  const text = getTextAtPosition(source.text, location.line, location.column);
+  const originalText = getTextAtPosition(
+    generatedSource.text,
+    actualLocation.line,
+    actualLocation.column
+  );
```
jasonLaster: this likely should be done closer to where the `createSyncData` call is made. the rationale for that is that the bp could move bc of ast bps and other factors...

darkwing:
Note to self; the following is useful in seeing a view of all the important information within `addBreakpoint` and `syncBreakpoint`:

```js
console.table([
    ["source.url", source.url],
    ["generatedSource.text", generatedSource.text],
    ["source.text", source.text],
    ["location.line", location.line],
    ["actualLocation.line", actualLocation.line],
    ["actualLocation.column", actualLocation.column],
    ["text", text],
    ["originalText", originalText]
  ]);
```

src/utils/source.js
```diff
@@ -313,3 +313,19 @@ export function isLoaded(source: SourceRecord) {
 export function isLoading(source: SourceRecord) {
   return source.get("loadedState") === "loading";
 }
+
+export function getTextAtPosition(
+  text: string,
+  line: ?number,
+  column: ?number
+) {
+  if (text === undefined) {
```
jasonLaster: this is probably better because `""` could qualify too
```js
if (!text) {
  return "" 
}
```

src/actions/breakpoints/addBreakpoint.js
```diff
@@ -52,6 +53,21 @@ export default async function addBreakpoint(
...
+  // generatedSource.text, in the case of pretty print or original tab
+  // returns the minified contents, which is why it's used
+  // when creating the "originalText"
+  const generatedSource = getSource(state, generatedLocation.sourceId);
+
+  const text = getTextAtPosition(source.text, location.line, location.column);
+  const originalText =
+    generatedSource && generatedSource.text
+      ? getTextAtPosition(
+          generatedSource.text,
```
jasonLaster: nit: this could be simplified

```diff
+        getTextAtPosition(
+           generatedSource && generatedSource.text,
+           actualLocation.line,
+           actualLocation.column
+         )
+       ;
```

```

File: src/actions/breakpoints/addBreakpoint.js [hunk 5]
```diff
@@ -52,6 +53,21 @@ export default async function addBreakpoint(
   const symbols = getSymbols(getState(), source);
   const astLocation = await getASTLocation(source, symbols, newLocation);
 
+  // generatedSource.text, in the case of pretty print or original tab
+  // returns the minified contents, which is why it's used
+  // when creating the "originalText"
```
jasonLaster: Why don't we get the original text from the original source? `source`

darkwing: As I mentioned last week, `addBreakpoint` is receiving `:formatted` (pretty-printed) sources.  I wish that wasn't the case, so this task would be as easy just using `originalSource` and `generatedSource`, but this has been part of what's delaying this problem.

darkwing:
As a general update, I'm looking at the `syncBreakpoint` flow.  What I'm seeing thus far:

1.  Creating breakpoints works and they display in the UI
2.  Refresh the source page
3.  WTF #1:  No breakpoints display in the breakpoints accordion pane but the debugger does pause on the breakpoint line
4.  WTF #2:  The `checkPendingBreakpoints` fires every time I hover over an object property in the editor (i.e. when the preview pane pops up)

jasonLaster:
WTF #2: The checkPendingBreakpoints function fires every time I hover over an object property in the editor (i.e. when the preview pane pops up)

> we do a console evaluate in those cases so it actually creates a new source :)

jasonLaster:
> WTF #1: No breakpoints display in the breakpoints accordion pane but the debugger does pause on the breakpoint line

not sure about that

darkwing:
Having turned off autoprettyprint, which is enabled in dev mode but disabled in production, everything here works.  The text is there when I first add the breakpoint (you can see this by hovering over teh breakpoint in the breakpoint pane and seeing its `title` attribute) and after I refresh the sample page.  W00t!

darkwing:
👍

PR #5828, opened 2018-03-30
Fix TypeError:  lastStatement is undefined

darkwing:
I've been seeing the following TypeError in the console when opening HTML files:  `lastStatement is undefined`


jasonLaster:
This is nice. Lets add a quick test here too

PR #5832, opened 2018-03-30
Remove executeSoon()

lukaszsobek:
If I understand correctly we use executeSoon in the promises to delay dispatching an action until the moment when there is a free spot in the event loop. This leads to nested promises and a situation where I believe we could achieve a very similar thing just by returning one promise.

The suggestion contains a full version of the implementation. I'd be very interested in insights on this matter.

### DIFF
--- a/src/actions/utils/middleware/promise.js
+++ b/src/actions/utils/middleware/promise.js
@@ -27,38 +27,34 @@ function promiseMiddleware({
     if (!(PROMISE in action)) {
       return next(action);
     }
-
     const promiseInst = action[PROMISE];
+
     const seqId = seqIdGen().toString();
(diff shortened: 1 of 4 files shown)


src/actions/utils/middleware/promise.js
```diff
@@ -27,38 +26,30 @@ function promiseMiddleware({
...
-              error: error.message || error
-            });
-            reject(error);
-          });
-        }
-      );
-    });
+    return promiseInst
+      .then(value => {
+        dispatch({ ...action, status: "done", value: value });
```
jasonLaster: I think we should keep the `setTimeout` ...

Issue #5770, opened 2018-03-23
Implement Photon ResultList Design


darkwing:
@violasong has provided the following mockups for our result list design:

![photo-dropdown-design](https://user-images.githubusercontent.com/46655/37837918-84fd156c-2e84-11e8-9326-4dd52c181bf2.png)

Once @marcogallo566's PR https://github.com/devtools-html/debugger.html/pull/5729 has been merged, the last remaining piece is adding the tab/"JS" image, which can be found here:

https://design.firefox.com/icons/viewer/

MicahRamirez:
Just making sure I understand the behavior for showing "JS" versus "Tab" image: 
We show the tab image in a ResultList item if there exists a tab with path equal to the ResultList item path?

I wasn't able to find the "JS" image within the provided link. Tab was there, but JS wasn't.

PR #5839, opened 2018-04-01
Implement Photon ResultList Design

MicahRamirez:
Fixes Issue: #5770

### Description 
Add an icon (as svg) to the ResultList that denotes either whether a tab exists or what type of file is being searched.  

### Summary of Changes
WIP

### Test Plan
WIP

## Original 
Pre Search Sources

Pre Search With Query

## Updated
WIP
## New Design
Design Goal


jasonLaster:
@MicahRamirez Can you update the description with an overview and some before and after screenshots

src/reducers/sources.js
```diff
@@ -56,7 +56,8 @@ export const SourceRecordClass = new I.Record({
   text: undefined,
   contentType: "",
   error: undefined,
-  loadedState: "unloaded"
+  loadedState: "unloaded",
+  tabExists: false
```
jasonLaster: I think we'll want to do this with either a redux selector or a check in the component

src/components/shared/ResultList.js
```diff
@@ -53,6 +53,11 @@ export default class ResultList extends Component<Props> {
 
     return (
       <li {...props}>
+        {item.svg && (
+          <div id={`${item.id}-svg`} className="svg">
```
jasonLaster: We generally try and avoid `id`s

src/utils/quick-open.js
```diff
@@ -122,7 +128,8 @@ export function formatSources(sources: SourcesMap): Array<QuickOpenResult> {
           .replace(sourcePath.split("/").pop(), "")
           .slice(1, -1),
         id: source.id,
-        url: source.url
+        url: source.url,
+        svg: <Svg name={svgName} />
```
jasonLaster: I think we'll want to do this in the component

src/utils/quick-open.js
```diff
@@ -112,6 +114,10 @@ export function formatSources(sources: SourcesMap): Array<QuickOpenResult> {
     .filter(source => !isPretty(source))
     .map(source => {
       const sourcePath = getSourcePath(source.url);
+      const sourcePathSplitByPeriod = sourcePath.split(".");
+      const svgName = source.tabExists
+        ? "tab"
+        : sourcePathSplitByPeriod[sourcePathSplitByPeriod.length - 1];
```
jasonLaster: we'll likely wan to do this in the component

src/utils/quick-open.js
```diff
@@ -112,6 +114,10 @@ export function formatSources(sources: SourcesMap): Array<QuickOpenResult> {
     .filter(source => !isPretty(source))
     .map(source => {
       const sourcePath = getSourcePath(source.url);
+      const sourcePathSplitByPeriod = sourcePath.split(".");
```
jasonLaster: we might want to have an extension field here which could be a function shared with SourcesTree

src/reducers/sources.js
MicahRamirez: Thanks for the early feedback, part of my motivation for throwing this out as WIP early : ).

That was my initial thought as well, but was running into some headaches with the augmenting the Source record. 

I still have that code, so I can pivot back np.

src/utils/quick-open.js
MicahRamirez: An extension field in the Source record? or are you talking about a helper function? Seems like the latter.

MicahRamirez:
So beyond the last reverting of the sources reducer file, I think I've got the first round of comments handled. 

I figured out how to work with immutable, but then that wasn't going to work because you can't add extra fields not defined in the record xD. 

@jasonLaster A question though around  the`<Svg/>` name attribute.

I dont know all possible file types that would be used with our debugger. We have a map `const svg = {...}` in the `<Svg/>` component but in most cases the entries are full names not file extension. I can either add ~duplicate~ require statements 
For example:
```
  js: require(jsSvg) 
  javascript: require(jsSvg),
  react: require(reactSvg),
  jsx: require(reactSvg),
```

OR I can have an extensionMap that will map from extension .js, .jsx, .vue to whatever names people specify in the `const svg ={...}` map

latter would be more to maintain and slightly more mental overhead imo. 

Left to do: fixing/adding tests and updating the styles

darkwing:
I think the `require` statement will work for importing icons :)

assets/images/Svg.js
```diff
@@ -24,6 +24,7 @@ const svg = {
   home: require("./home.svg"),
   javascript: require("./javascript.svg"),
   jquery: require("./jquery.svg"),
+  js: require("./javascript.svg"),
```
darkwing: We already have a `javascript` key, so you don't need to add this.

src/utils/source.js
```diff
@@ -313,3 +313,11 @@ export function isLoaded(source: SourceRecord) {
 export function isLoading(source: SourceRecord) {
   return source.get("loadedState") === "loading";
 }
+
+export function getExtensionType(url: string) {
```
darkwing: This would break if the URL had a "?", for instance.

Instead of this function, it would be best to factor out `SourceTree`'s `getIcon` method as well as `sourceTypes` o bject.  We should extract as much of that logic as we can and make it a utility method in `src/utils/source.js`

assets/images/Svg.js
MicahRamirez: I guess the idea here was to simplify the mapping between a sources extension .js, .css, .jsx .anything. In this case, simplifying by adding an additional key .js that maps to the correct svg. 

This might be a silly question, but what all file types can we have on the browser ? Just html, js, css? 

If I don't do it this way, this then becomes N number of if... checks where N is the number of different file types possible where I'd be checking `if (extension) svgName=extensionTypeFullName` eg `if (js) svgName = javascript` or we maintain a mapping from file extension to svgName in utils or something.

MicahRamirez: Ya'll have an extension map I think I see what you are looking for now.

src/utils/source.js
MicahRamirez: Question, why do we use <Svg/>  sometimes over  ? What's the preferred usage?

src/components/PrimaryPanes/SourcesTree.js
```diff
@@ -214,13 +206,7 @@ class SourcesTree extends Component<Props, State> {
...
       const obj = item.contents.get("id");
       const source = sources.get(obj);
-      if (source && source.get("isBlackBoxed")) {
-        return <img className="blackBox" />;
-      }
-
-      const sourceType = sourceTypes[getExtension(source)];
-      const classNames = classnames("source-icon", sourceType || "file");
-      return <img className={classNames} />;
+      return <img className={getSourceClassnames(source.toJS())} />;
```
MicahRamirez: @darkwing I've only extracted the logic that dealt directly with source. It sounded like you were potentially looking for more of a refactor ?

src/utils/source.js
jasonLaster: + CSS is more performant than inline svgs

src/components/shared/ResultList.js
```diff
@@ -50,9 +51,14 @@ export default class ResultList extends Component<Props> {
         selected: index === selected
       })
     };
-
+    const svgName = item.tabExists
+      ? "source-icon tab"
+      : getSourceClassnames(item);
     return (
       <li {...props}>
+        <div className="svg">
```
jasonLaster: should we have this <div> if the svgName does not exist?

src/components/shared/ResultList.js
```diff
@@ -50,9 +51,14 @@ export default class ResultList extends Component<Props> {
         selected: index === selected
       })
     };
-
+    const svgName = item.tabExists
+      ? "source-icon tab"
+      : getSourceClassnames(item);
```
jasonLaster: this logic should probably live in QuickOpenModal

src/reducers/quick-open.js
```diff
@@ -72,3 +74,19 @@ export function getQuickOpenQuery(state: OuterState) {
...
+  getTabs,
+  (formattedSources, tabs) => {
+    const tabsAsArray = tabs.toArray();
+    const formattedSourcesWithTabExistence = formattedSources.map(source => {
+      source.tabExists = tabsAsArray.includes(source.url);
+      return source;
+    });
+    return formattedSourcesWithTabExistence;
+  }
+);
```
jasonLaster: this is the first `sources` related selector in quick-open, it is also the first time a reducer has imported ../selectors, which makes me nervous of circular dependencies.

Perhaps, you were right the first time to `source.tabExists = tabsAsArray.includes(source.url);` in formatSources :wink:

src/components/PrimaryPanes/SourcesTree.js
darkwing: I like what you've done :)

src/components/shared/ResultList.js
darkwing: I think `result-item-icon` may be a better className.

darkwing:
Wow @MicahRamirez, this is look decent! :)

A few issues I found:

- [ ] The icon looks a bit lower than the file name
- [ ] The icon needs a specific styling (`.result-list.big li.selected .svg`?) for when it's selected -- it's too dark when selected:

darktab

The dark theme colors look great:

yess

Awesome work so far! :)

src/components/shared/ResultList.js
MicahRamirez: Not 100% on the question from jason, but I'll take a stab at it
I would think the div should still be there to keep the spacing consistent?

```
| icon | someSource/source.js
sourceSource/typeWithoutSvg.weirdExtension  // compare below
|         | sourceSource/typeWithoutSvg.weirdExtension // compare above
| icon | someSource/source.js
```

@darkwing I'm fine with being more specific with the svgname. It's just not consistent with the other html elements defined in renderListItem (e.g. classnames for the other respective divs are just "title" and "subtitle")

MicahRamirez:
@darkwing thanks for following up. Yeah I noticed the colors weren't inverting when selected. I was trying to take some inspiration from how we were doing the styling for the source tree highlighting, but was having difficulty getting that working. Will be taking another swing at it shortly. 

@jasonLaster thanks! sounds good. I'd probably prefer writing up a follow up issue, especially since I haven't fixed the styles nor addressed the tests. 

Will be working on testing the new changes with some local changes I have and then getting the styling fixed hopefully tonight. 

Really appreciate the feedback and insights along the way. Makes me excited to be contributing!

MicahRamirez:
I'm having some trouble with styles, specifically the sizing and positioning. We use flexbox for the list item, but then we switch back to block for the items that compose the list item. Are we supposed to use explicit sizes for icon containers and then the respective image? 

I'm not a css ninja, but it looks like we are using similar styles for the title and subtitle between the QuickOpenModal.css and the ResultListItem.css, which is throwing me off a little.

MicahRamirez:
Will take a deeper look at this tomorrow!

src/utils/quick-open.js
```diff
@@ -67,6 +70,7 @@ export type QuickOpenResult = {|
   id: string,
   value: string,
   title: string,
+  badge?: boolean,
```
jasonLaster: should the be icon?

src/utils/source.js
```diff
@@ -329,3 +340,16 @@ export function getTextAtPosition(source: Source, location: Location) {
 
   return lineText.slice(column, column + 100).trim();
 }
+
+export function getSourceClassnames(
+  source: Object,
+  ...additionalClasses: Array<string>
+) {
+  const sourceRecord = new SourceRecordClass(source);
```
jasonLaster: why is this needed?

src/utils/source.js
```diff
@@ -329,3 +340,16 @@ export function getTextAtPosition(source: Source, location: Location) {
...
+  if (source && sourceRecord.get("isBlackBoxed")) {
```
jasonLaster: this seems like it is doing too much...

It is probably sufficient to return blackbox or the sourceType

src/utils/quick-open.js
```diff
@@ -50,14 +51,16 @@ export function parseLineColumn(query: string) {
...
   const title = getFilename(source);
   const subtitle = endTruncateStr(source.relativeUrl, 100);
-
   return {
     value: source.relativeUrl,
     title,
     subtitle,
+    icon: tabs.includes(source.url)
+      ? "tab result-item-icon"
+      : getSourceClassnames(source, "result-item-icon"),
```
jasonLaster: we can do classnames("result-item-icon",  getSourceClassnames(source))` to keep the util simpler

src/components/PrimaryPanes/SourcesTree.js
```diff
@@ -214,13 +206,9 @@ class SourcesTree extends Component<Props, State> {
...
       const source = sources.get(obj);
-      if (source && source.get("isBlackBoxed")) {
-        return <img className="blackBox" />;
-      }
-
-      const sourceType = sourceTypes[getExtension(source)];
-      const classNames = classnames("source-icon", sourceType || "file");
-      return <img className={classNames} />;
+      return (
+        <img className={getSourceClassnames(source.toJS(), "source-icon")} />
```
jasonLaster: same here... `classnames("source-icon", getSourceClassnames(source))`

src/utils/source.js
MicahRamirez: to use the .get method it needs to be a SourceRecord. getExtension also uses .get even though the method signature says the source parameter is a plain object. My initial inclination was to maintain the contract there.

However, `getExtension` is only used in`getSourceClassnames`, so as far as I am concerned we should just be working with plain js objects here. Good catch.

src/utils/quick-open.js
MicahRamirez: my thought process was that this would typically be used on an image tag, so it's highly likely you'd want to add additional classes that are relevant to the source class name but not derived from the extension type. 

This does expand on its defined behavior maybe a bit too much.

src/components/QuickOpenModal.css
```diff
@@ -14,7 +14,8 @@
 }
 
 .theme-dark .result-item .title .highlight,
-.theme-dark .result-item .subtitle .highlight {
+.theme-dark .result-item .subtitle .highlight,
+.theme-dark .result-item .svg .highlight {
```
MicahRamirez: - [ ]  Remove the .svg style

src/utils/source.js
```diff
@@ -329,3 +339,12 @@ export function getTextAtPosition(source: Source, location: Location) {
...
   return lineText.slice(column, column + 100).trim();
 }
+
+export function getSourceClassnames(source: Object) {
+  if (source && source.isBlackBoxed) {
+    return "blackBox";
+  }
+  const sourceType = sourceTypes[getExtension(source)];
+  const classNames = classnames(sourceType || "file");
+  return classNames;
```
jasonLaster: do we need `classnames(sourceType || "file")`

could it just be `return sourceType || "file"`
`

darkwing:
I made a few small positioning and code updates; if Travis we'll merge! :)

darkwing:
@jasonLaster @bomsy 

Can I get advice here?  The jest errors are triggering because of the following line in `getExtension(source: Object)`:

```js
const parsedUrl = parse(source.url).pathname;
```

All UI functionality works with `.url` but jest is unhappy because I'm not using `source.get("url")`.  For example, if I change this function to use:

```js
const parsedUrl = parse(source.get ? source.get("url") : source.url).pathname;
```

...everything works properly.  If I use only `source.get("url")`, the Tree builds properly but using QuickOpen bombs because `source.get` is not a function.

In the case of this function, `source` can be a `RelativeSource` (quick open) or `Source` (tree).

Any idea why Jest would be upset but the UI is in fine shape with `.url`?

MicahRamirez:
Thank you for pushing the rest of this through. Really sorry for not taking the lead on getting this pushed through. Rough past week. I appreciate ya'lls guidance and feedback.

jasonLaster:
Thanks for the help! It was great working with you

darkwing:
@MicahRamirez You did great!  Thank you so much for your help! :)

PR #5843, opened 2018-04-02
Add stricter types for Breakpoint and AST actions

ryanjduffy:
Fixes Issue: #3080 

### Summary of Changes

* Add PromiseAction generic type
* Update Breakpoint and AST actions to use exact object types and covariant props
* Adopt PromiseAction for Breakpoint and AST action where used


src/reducers/breakpoints.js
```diff
@@ -92,7 +92,8 @@ function addBreakpoint(state, action) {
 
   // when the action completes, we can commit the breakpoint
   if (action.status === "done") {
-    return syncBreakpoint(state, action.value);
+    const { value } = ((action: any): DonePromiseAction);
+    return syncBreakpoint(state, value);
```
jasonLaster: could we move the type check into `syncBreakpoint`?

src/actions/ast.js
```diff
@@ -34,13 +34,15 @@ export function setSourceMetaData(sourceId: SourceId) {
...
+    const action: Action = {
       type: "SET_SOURCE_METADATA",
       sourceId: source.id,
       sourceMetaData: {
         framework
       }
-    });
+    };
+
+    dispatch(action);
```
jasonLaster: we might be able to use [this](https://github.com/devtools-html/debugger.html/blob/master/src/reducers/sources.js#L45) strategy:

```js
dispatch({
  type: "SET_SOURCE_METADATA",
  sourceId: source.id,
  sourceMetaData: {
    framework
  }
}: Action);
```

ryanjduffy: I started with that but prettier(?) wanted it to be wrapped with parenthesis and indented which seemed less readable than extracting it to variable.

```
    dispatch(
      ({
        type: "SET_SOURCE_METADATA",
        sourceId: source.id,
        sourceMetaData: {
          framework
        }
      }: Action)
    );
```

jasonLaster: Hmm, i think I'd opt for this... having a separate var feels awkward. Another option is exporting a dispatch function from `actions/index` that asserts the param is Action...

src/reducers/breakpoints.js
ryanjduffy: We'd have to refactor a bit because the action `syncBreakpoint()` expects isn't a PromiseAction so the structures are different. In this case, this seems like the least intrusive approach but open to updating if you'd like.

jasonLaster: cool. I'm good with that

src/actions/ast.js
ryanjduffy: :+1: I'll update those instances to the inline type

PR #5851, opened 2018-04-03
Switched getExtraProps to use findClosestFunction to get the component name.

mmcote:
### Summary of Changes
- Changed getExtraProps to use findClosestFunction, which has now been changed to findClosestofSymbolDeclaration to find the name of the component even when in production  code.

### Screenshots
Before
screen shot 2018-04-03 at 1 58 12 pm

After
screen shot 2018-04-03 at 2 02 36 pm

### DIFF
--- a/src/utils/ast.js
+++ b/src/utils/ast.js
@@ -8,11 +8,14 @@ import { without, range } from "lodash";
 
 import type { Location, Source, ColumnPosition } from "../types";
 
-import type { AstPosition, AstLocation, PausePoint } from "../workers/parser";
 import type {
+  AstPosition,
+  AstLocation,
+  PausePoint,
+  SymbolDeclaration,
   SymbolDeclarations,
-  FunctionDeclaration
-} from "../workers/parser/getSymbols";
+  ClassDeclaration
+} from "../workers/parser";
 
 export function findBestMatchExpression(
   symbols: SymbolDeclarations,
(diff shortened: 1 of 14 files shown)


src/utils/ast.js
```diff
@@ -68,11 +71,11 @@ export function containsPosition(a: AstLocation, b: AstPosition) {
   return startsBefore && endsAfter;
 }
 
-export function findClosestFunction(
-  functions: FunctionDeclaration[],
+function findClosestofSymbolDeclaration(
```
jasonLaster: nit: `findClosestofSymbol` is smaller :)

src/utils/ast.js
```diff
@@ -100,3 +103,25 @@ export function findClosestFunction(
...
+}
+
+export function findClosestClass(
+  symbols: SymbolDeclarations,
+  location: Location
+) {
+  const { classes } = symbols;
+  if (!classes) {
+    return null;
+  }
```
jasonLaster: we can move this to `findClosestofSymbol`

darkwing:
Wow, this looks slick @mmcote!  Nice work!

So I'm trying to test this on a React app (debugger.html) and I'm not seeing the `t` or `App` panes;  any idea what I could be doing wrong?

Thanks!

jasonLaster:
We can look into the debugger case separately... This logic works well and the code reads well.

Issue #3829, opened 2017-08-30
[ProjectSearch] search the source when a match is selected


jasonLaster:
When clicking on a line result in the search results, it would be nice if the line or keyword instance would be highlighted when the source is opened.

Atom does this really nicely when you select a match. For us it would be as easy as firing this [action](https://github.com/devtools-html/debugger.html/blob/master/src/actions/ui.js#L45-L50)

### 1. project search

![Screen Shot 2017-08-30 at 5.43.16 PM.png](https://shipusercontent.com/f610a683bef2a6c357d7553f7c1de7e8/Screen%20Shot%202017-08-30%20at%205.43.16%20PM.png)

### 2. bin/mochi

![Screen Shot 2017-08-30 at 5.43.09 PM.png](https://shipusercontent.com/c9f6c096ea24c7f3b5cd12ddf26e3701/Screen%20Shot%202017-08-30%20at%205.43.09%20PM.png)

AlexisDeschamps:
``` export class ProjectSearch extends Component<Props, State> {

   selectMatchItem = (matchItem: Match) => {
     this.props.selectLocation({ ...matchItem });
+    this.props.doSearch(this.state.inputValue, editor);
   };

   getResults = (): Result[] => {```
This change will add the search on source selection, but the editor is not available within this component and would need to be obtained.

urielha:
I am not sure if this helps, but anyway 
The old FireFox debugger had this feature as you can see in this GIF :
<hr/>

![https://i.imgur.com/yTMfi8t.gif](https://i.imgur.com/yTMfi8t.gif)
<hr/>

Comparing to the new debugger which doesn't highlight the matching clause :
<hr/>

![https://i.imgur.com/4NBhQG2.gif](https://i.imgur.com/4NBhQG2.gif)
<hr/>

jasonLaster:
Thanks @urielha

jasonLaster:
I think we can get the editor if we set it in module state:

```diff
diff --git a/src/utils/editor/index.js b/src/utils/editor/index.js
index 00d689a..fa6e15c 100644
--- a/src/utils/editor/index.js
+++ b/src/utils/editor/index.js
@@ -20,6 +20,20 @@ import type { AstLocation } from "../../workers/parser";
 import type { EditorPosition, EditorRange } from "../editor/types";
 import type { Location } from "../../types";

+let editor = null;
+
+export function setEditor(_editor) {
+  editor = _editor;
+}
+
[...]

PR #5854, opened 2018-04-04
Started highlighting the searched terms when clicking on a ProjectSearch result

AlexisDeschamps:
Issue: #3829

### Summary of Changes

I took the function flow used to do what I believe to be file searches and removed some of lines that did not seem necessary if only highlighting was needed.

File search function flow:
`doSearch` (`actions/file-search.js`) --> `searchContents` (`actions/file-search.js`) --> `find` (`utils/editor/source-search.js`)  --> `doSearch` (`utils/editor/source-search.js`) --> `searchNext` (`utils/editor/source-search.js`)

I added these new functions to only get the highlighting without touching the current file search:
`doSearchForHighlight` (`actions/file-search.js`) --> `searchContentsForHighlight` (`actions/file-search.js`) --> `searchSourceForHighlight` (`utils/editor/source-search.js`)  --> `findNextOnLine` (`utils/editor/source-search.js`)

- Started calling `doSearchForHighlight` when a `ProjectSearch` result is selected
- Added functions to get the `Editor` object in the `ProjectSearch` component

### Test Plan
- [ ] In Debugee, go to http://firefox-dev.tools/debugger-examples/examples/todomvc/
- [ ] In Debugger, open Project Search (`Ctrl/Cmd + Shift + F`)
- [ ] In Debugger, project search for "app" (type "app", then press `Enter`)
- [ ] In Debugger, click a project search result
- [ ] See that all instances of "app" in the file have a box highlight
- [ ] See that the specifically selected project search result is fully highlighted

### Screenshots
Before
image

After
image

### DIFF
--- a/src/utils/editor/source-search.js
+++ b/src/utils/editor/source-search.js
@@ -157,6 +157,31 @@ function doSearch(ctx, rev, query, keepSelection, modifiers: SearchModifiers) {
   });
 }
 
+export function searchSourceForHighlight(
+  ctx: Object,
+  rev: boolean,
+  query: string,
+  keepSelection: boolean,
+  modifiers: SearchModifiers,
+  line: number,
+  ch: number
+) {
+  const { cm } = ctx;
+  if (!cm) {
+    return;
+  }
+
+  return cm.operation(function() {
+    const state = getSearchState(cm, query);
+    const isNewQuery = state.query !== query;
+    state.query = query;
+
+    updateOverlay(cm, state, query, modifiers);
+    updateCursor(cm, state, keepSelection);
+    findNextOnLine(ctx, rev, query, isNewQuery, modifiers, line, ch);
+  });
+}
+
 function getCursorPos(newQuery, rev, state) {
   if (newQuery) {
     return rev ? state.posFrom : state.posTo;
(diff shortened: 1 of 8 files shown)


src/utils/editor/source-search.js
```diff
@@ -208,6 +233,46 @@ function searchNext(ctx, rev, query, newQuery, modifiers) {
...
+    if (!cm.state.selectingText) {
+      ed.alignLine(cursor.from().line, "center");
+      cm.setSelection(cursor.from(), cursor.to());
+    }
+
+    nextMatch = { from: cursor.from(), to: cursor.to() };
+  });
+
+  return nextMatch;
+}
```
jasonLaster: how is this different from searchNext? could we re-use `searchNext` are there are parts that we want to share?

src/utils/editor/source-search.js
```diff
@@ -157,6 +157,31 @@ function doSearch(ctx, rev, query, keepSelection, modifiers: SearchModifiers) {
...
+  return cm.operation(function() {
+    const state = getSearchState(cm, query);
+    const isNewQuery = state.query !== query;
+    state.query = query;
+
+    updateOverlay(cm, state, query, modifiers);
+    updateCursor(cm, state, keepSelection);
+    findNextOnLine(ctx, rev, query, isNewQuery, modifiers, line);
+  });
+}
```
jasonLaster: this looks similar to doSearch, are there are parts we want to share? Why do we need two functions?

src/utils/editor/source-search.js
```diff
@@ -157,6 +157,31 @@ function doSearch(ctx, rev, query, keepSelection, modifiers: SearchModifiers) {
...
+
```
jasonLaster: do we want to add test coverage here:
https://github.com/AlexisDeschamps/debugger.html/blob/3829-search-on-match/src/utils/editor/tests/source-search.spec.js#L1

src/utils/editor/source-search.js
AlexisDeschamps: Yeah, the only differences are:
1. accepts `line` and `ch` so `findNextOnLine` can know which line to full highlight
2. calls `findNextOnLine` instead of `findNext`, which are pretty different right now

src/utils/editor/source-search.js
AlexisDeschamps: Yeah, so I guess `searchNext` needs to actually find the next instance of the query to full highlight, while `findNextForSearch` would already know the starting `line` and `ch`, so it doesn't need to do a bunch of logic to find the starting location.

src/utils/editor/source-search.js
AlexisDeschamps: Yes, I will investigate further for the tests.

darkwing:
A good future update would be putting the cursor on the highlighted word when selected.  Nice work @AlexisDeschamps !

AlexisDeschamps:
@darkwing OK, cool, thanks!

PR #5861, opened 2018-04-05
Fix webpack roots

jasonLaster:
### Summary of Changes

We had an issue where webpack based urls had their paths changed in the tree from `webpack:///src/file.js` to `Webpack/src/file.js` changing the source url meant that the root no longer lined up with the source urls. This patch fixes that!

---

### Before
screen shot 2018-04-05 at 9 38 16 am

### After
screen shot 2018-04-05 at 9 19 00 am

### DIFF
--- a/src/selectors/getRelativeSources.js
+++ b/src/selectors/getRelativeSources.js
@@ -34,6 +34,6 @@ export function getRelativeSources(state: State): RelativeSource[] {
   return sources
     .valueSeq()
     .toJS()
-    .filter(({ url }) => url && url.startsWith(root))
+    .filter(({ url }) => url && url.includes(root))
     .map(source => formatSource(source, root));
 }
--- a/src/selectors/getRelativeSources.js
+++ b/src/selectors/getRelativeSources.js
@@ -35,6 +35,6 @@ export function getRelativeSources(state: State): RelativeSource[] {
   return sources
     .valueSeq()
     .toJS()
-    .filter(({ url }) => url && url.includes(root))
+    .filter(({ url }) => url && url.startsWith(root))
     .map(source => formatSource(source, root));
 }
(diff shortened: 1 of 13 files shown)


src/selectors/getRelativeSources.js
```diff
@@ -31,10 +31,9 @@ function formatSource(source: Source, root): RelativeSource {
 export function getRelativeSources(state: State): RelativeSource[] {
   const sources = getSources(state);
   const root = getProjectDirectoryRoot(state);
-
   return sources
     .valueSeq()
     .toJS()
-    .filter(({ url }) => url && url.includes(root))
+    .filter(({ url }) => url && url.startsWith(root))
```
darkwing: This change, from `includes` to `startsWith`, is causing no sources to display when a root is set.  Do we need this?

jasonLaster: i think startsWith should now work because of the other changes

jasonLaster:
oh nice catch

jasonLaster:
```
Waiting on sources: nested-source
   Finished waiting on sources: nested-source
   Waiting for state change: selected source
   Finished waiting for state change: selected source
 TEST-UNEXPECTED-FAIL Nodes in path are automatically expanded 
   Adding a new tab with URL: http://example.com/browser/devtools/client/debugger/new/test/mochitest/examples/doc-sources.html
  JS warning: 
   "Loading failed for the <script> with source “http://example.com/browser/devtools/client/debugger/new/test/mochitest/examples/nested/deeper/deeper-source.js”." {file: "http://example.com/browser/devtools/client/debugger/new/test/mochitest/examples/doc-sources.html" line: 23}]
   
   ```
   
   oh no, looks like this broke dbg-sources

darkwing:
My local disagrees!  Kicking it.

Issue #5849, opened 2018-04-03
[Breakpoints] Syntax highlight the breakpoint text


darkwing:
Parent Issue:  https://github.com/devtools-html/debugger.html/issues/5773

Now that we have the breakpoint text, we need to syntax highlight that code using CodeMirror.

PR #5864, opened 2018-04-05
Syntax-highlight breakpoints

darkwing:
Fixes Issue: #5849

### Summary of Changes

* Adds syntax highlighting to Breakpoints pane breakpoints
* Improves the layout of close button (more padding to avoid hover / close button movement)

### Test Plan

Add breakpoints, remove breakpoints, disable breakpoints -- ensure all looks good.

### Screenshots

updatedscreenshot

### ToDo

- [x] Figure out how to get text-ellipsis working
- [x] Investigate weird case where sometimes a CM instance isn't created and the plain text version of the breakpoint code persists
- [x] Performance - the breakpoints pane seems a bit slower.
- [x] Side Effects - Disable a breakpoint and then add another -- the disabled breakpoint syntax-highlighted code appears enabled for a second and then fades to disabled opacity, despite no action on that element being taken.

### DIFF
--- a/src/components/SecondaryPanes/BreakpointItem.js
+++ b/src/components/SecondaryPanes/BreakpointItem.js
@@ -39,9 +39,14 @@ class BreakpointItem extends Component<Props> {
   componentDidMount() {
     this.setupEditor();
   }
+  componentDidUpdate() {
+    this.setupEditor();
+  }
 
   componentWillUnmount() {
-    this.editor.destroy();
+    if (this.editor) {
+      this.editor.destroy();
+    }
   }
 
   shouldComponentUpdate(nextProps: Props) {
--- a/src/components/SecondaryPanes/BreakpointItem.js
+++ b/src/components/SecondaryPanes/BreakpointItem.js
@@ -11,7 +11,9 @@ import CloseButton from "../shared/Button/Close";
 
 import { createEditor } from "../../utils/breakpoint";
 import { features } from "../../utils/prefs";
+
 import type { LocalBreakpoint } from "./Breakpoints";
+import type SourceEditor from "../../utils/editor/source-editor";
 
 type Props = {
   breakpoint: LocalBreakpoint,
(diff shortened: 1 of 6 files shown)


src/components/SecondaryPanes/BreakpointItem.js
```diff
@@ -30,6 +34,41 @@ function getBreakpointLocation(source, line, column) {
...
+  }
+
+  componentWillUnmount() {
+    this.editor.destroy();
+  }
+
+  shouldComponentUpdate(nextProps: Props) {
+    // This is needed, IMO, so we don't need to create
+    // loads of CM instances for no reason
+    return this.props.breakpoint != nextProps.breakpoint;
```
darkwing: CodeMirror seems like it could be a taxing tool, and since the other BreakpointItem props shouldn't change, I thought this could be useful.

src/components/SecondaryPanes/BreakpointItem.js
```diff
@@ -30,6 +34,41 @@ function getBreakpointLocation(source, line, column) {
...
+
+    // disables the default search shortcuts
+    // $FlowIgnore
+    this.editor._initShortcuts = () => {};
+
+    const node = ReactDOM.findDOMNode(this);
+    if (node instanceof HTMLElement) {
+      const mountNode = node.querySelector(".breakpoint-label");
+      if (node instanceof HTMLElement) {
+        // $FlowIgnore
```
darkwing: I hated to use this but flow keeps yelling at me:  `Cannot assign empty string to mountNode.innerHTML because property innerHTML is missing in null [1].`

darkwing:
Re: Performance:

- [ ] Every breakpoint gets re-rendered when any breakpoint is enabled or disabled, added or deleted

darkwing:
I think `.map((bp, i) => this.renderBreakpoint(bp, `${filename}-${i}`))` in `Breakpoints.js` is the problem -- if adding/removing a breakpoint causes an index change (`${i}`), the key changes and thus a re-render is probably triggered!

jasonLaster:
> `.map((bp, i) => this.renderBreakpoint(bp,${filename}-${i}))`

oh interesting, I believe that! 

I think we can likely remove the shouldUpdate call by doing a pretty extreme refactoring [branch](https://github.com/jasonLaster/debugger.html/tree/breakpoint-highlight2). The reason i think we need it is that every time Breakpoints renders we create new local breakpoints, so BreakpointItem feels like it should re-render. We can change the connect function to be more pure, basically give us all the data unchanged, and then on the component side do what we need. 

As i said yesterday, this component is one of the first components we wrote so its got some early ideas.

jasonLaster:
I'm landing because this version is pretty awesome and we can refactor as a follow up

jasonLaster:
I filed #5875 as a follow up

Issue #5800, opened 2018-03-27
[QuickOpen] Escape no longer closes the modal


jasonLaster:
STR:

1. open QO via `cmd+p`
2. press `<escape>`

ER: Modal should close
AR: Modal is unaffected

darkwing:
Oddly enough I could make this happen once but then it worked every other time, whether there was a source open or not.

jasonLaster:
nope... not sure how either

dimitardanailov:
Greetings, 

We should discuss this Boolean flag: 

```javascript
if (quickOpenEnabled === true) {
  closeQuickOpen();
}
```

If @wldcordeiro share with us why the Boolean flag is required we could discuss how to restructure the source code.

I was able to reproduce the issue but I can't share which steps are required. 
The issue is very hard for reproduce.

Demo with my success: https://www.youtube.com/watch?v=ew80SQx77CA&feature=youtu.be

jasonLaster:
can we get rid of that block if `closeActiveSearch` does the same thing?

darkwing:
I suggested the same :)

dimitardanailov:
I'll have time on Thursday to play with source code.

> can we get rid of that block if closeActiveSearch does the same thing?
I agree with you.

PR #5866, opened 2018-04-05
Escape no longer closes the modal

dimitardanailov:
Fixes Issue: #5800 

I had a conversation with @jasonLaster  and @darkwing about the `if` statement. 
The issue was very hard for reproduce (Video: https://www.youtube.com/watch?v=ew80SQx77CA&feature=youtu.be) and our decision was `if` statement must be removed. 

### Summary of Changes

* change 1 - `if (quickOpenEnabled === true) {` was removed.

### Test Plan

STR:
open QO via cmd+p
press <escape>

ER: Escape key should close the modal dialog


jasonLaster:
@dimitardanailov i'm not sure what's going on... Looks like we need the IF after all. I added a tweak, which might help

dimitardanailov:
@jasonLaster 

1)
I didn't like this section: 

```
if (quickOpenEnabled) {
  closeQuickOpen();
}

if (activeSearch || quickOpenEnabled) {
  e.preventDefault();
}
```

2) 
Why this flag (`quickOpenEnabled`) is required ?

darkwing:
Add a commit that only uses `preventDefault` when an action is executed; also added back destructuring of used properties.

PR #5871, opened 2018-04-06
Handle global-rewrite (Ember-related) original scope resolution

loganfsmyth:
Refs Issue: #5561

### Summary of Changes

* Allow resolution of original scope values that have been rewritten to reference globals
* Allow resolution of original scope values that map to multiple generated locations
* Add a mochitest for scope resolution in an ember quickstart application


loganfsmyth:
@jasonLaster The tests are failing on this because I forgot I made that local change to my mochitest environment to use the new `devtools-source-map` worker that has `getAllGeneratedLocations`. How do we update the main environment to do that?

Issue #5120, opened 2018-01-16
[Flow] add eslint-plugin-flowtype - require-valid-file-annotation


gregtatum:
It would be great to use this [eslint rule](https://github.com/gajus/eslint-plugin-flowtype#require-valid-file-annotation) to ensure that we don't mistakenly leave out `// @flow` headers.

Here are this list of files that are missing the header. Lets add the header and the minimal types needed to satisfy flow. It is okay to add a lot of loose types like `any`, `Object`, etc... 

### Example

here is an example patch that could be a PR

```diff
diff --git a/src/utils/ui.js b/src/utils/ui.js
index efc8b03..599989d 100644
--- a/src/utils/ui.js
+++ b/src/utils/ui.js
@@ -2,6 +2,8 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */

+// @flow
+
 /* Checks to see if the root element is available and
  * if the element is visible. We check the width of the element
  * because it is more reliable than either checking a focus state or
@@ -14,7 +16,7 @@ export function isVisible() {

 /* Gets the line numbers width in the code editor
  */
-export function getLineNumberWidth(editor) {
+export function getLineNumberWidth(editor: Object) {
   const gutters = editor.display.gutters;
   const lineNumbers = gutters.querySelector(".CodeMirror-linenumbers");
   return lineNumbers && lineNumbers.clientWidth;
```

### Todo List

- [ ] src/actions
  - [ ] ast/setInScopeLines.js
  - [ ] breakpoints/addBreakpoint.js
  - [ ] breakpoints/remapLocations.js
  - [ ] event-listeners.js
  - [ ] source-tree.js
  - [ ] utils/middleware/log.js
  - [ ] utils/middleware/timing.js
- [ ] src/components
  - [ ] SecondaryPanes/Breakpoints/BreakpointsContextMenu.js
  - [ ] SecondaryPanes/Frames/types.js
- [ ] src/reducers
  - [ ] /async-requests.js
  - [ ] src/reducers/event-listeners.js
  - [ ] src/reducers/index.js
  - [ ] src/reducers/quick-open.js
  - [x] src/reducers/types.js
- [ ] src/selectors/
  - [ ] breakpointAtLocation.js
  - [ ] getCallStackFrames.js
  - [ ] isSelectedFrameVisible.js
  - [x] visibleBreakpoints.js
- [ ] src/utils
[...]

PR #5872, opened 2018-04-06
[Flow] src/utils/DevtoolsUtils.js - add type defs for flow coverage

atwalg2:
Fixes Issue: #5120

### DIFF
--- a/src/utils/DevToolsUtils.js
+++ b/src/utils/DevToolsUtils.js
@@ -11,7 +11,7 @@ export function reportException(who: string, exception: any[]) {
   console.error(msg, exception);
 }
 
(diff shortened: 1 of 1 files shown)


src/utils/DevToolsUtils.js
```diff
@@ -2,14 +2,16 @@
...
 import assert from "./assert";
 
-export function reportException(who, exception) {
+export function reportException(who: string, exception: any[]) {
   const msg = `${who} threw an exception: `;
   console.error(msg, exception);
 }
 
-export function executeSoon(fn) {
+export function executeSoon(fn: () => void) {
```
AnshulMalik: `fn: Function` would look pretty here :)

PR #5880, opened 2018-04-06
Enable reading of the wasm name section to display debug names in text

yurydelendik:
Fixes Issue: the wasm name section is not used for names

### Summary of Changes

* Attempt to read the section and use it during disassembly

### Test Plan

* Open and debug https://yurydelendik.github.io/sqlite-playground/src/playground.html
* Notice that function names are not autogenerated (i.e. starts with `$func`)

### DIFF
--- a/src/utils/wasm.js
+++ b/src/utils/wasm.js
@@ -5,7 +5,7 @@
 /* @flow */
 
 import { BinaryReader } from "wasmparser/dist/WasmParser";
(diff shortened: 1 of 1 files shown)


yurydelendik:
Depends on https://github.com/devtools-html/debugger.html/pull/5870

Related change in disassembler: https://github.com/wasdk/wasmparser/commit/0498c32b622e8a1683e94719f99586520f4d0f68

Issue #5869, opened 2018-04-05
[Sources] Reveal in Tree should open the left sidebar


jasonLaster:
I just noticed that reveal in sidebar doesn't open the left sidebar if it is closed. Should be a nice fix :)

PR #5885, opened 2018-04-07
[Sources] Reveal in tree opens the left sidebar

marcobeltempo:
**Fixes Issue: #5869** 

### Summary of Changes

When selecting ` Reveal in tree` from the context menu, it now checks:
 - if the left panel is collapsed
   - opens if it is 

### Test Plan

- [x] Select `Reveal in tree` with the left panel **closed** to make sure it opened
- [x] Select `Reveal in tree` with the left panel **opened** to make sure it didn't collapse
- [x] With the context menu opened, enter `Shift` + `Alt` + `r` will also open the left panel

### Screenshots/Videos

debugger_html_issue5869

### DIFF
--- a/src/actions/ui.js
+++ b/src/actions/ui.js
@@ -68,6 +68,14 @@ export function showSource(sourceId: string) {
   return ({ dispatch, getState }: ThunkArgs) => {
     const source = getSource(getState(), sourceId);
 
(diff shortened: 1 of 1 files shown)


jasonLaster:
Thanks for the help!

By the way, you should join our slack https://devtools-html-slack.herokuapp.com/ and say hi!

PR #5892, opened 2018-04-09
[scopes] Batch generated locations

jasonLaster:
Fixes Issue: #5561

### Summary of Changes

This is an attempt to speedup the mapping logic by consolidating the source map location fetching.
I dont know if this was the slowest path though... I do know that this is faster when there are 19K locations...

Here's a new profile after the fact

screen shot 2018-04-09 at 7 22 38 pm


src/utils/pause/mapScopes/generateClientScope.js
```diff
@@ -0,0 +1,68 @@
+export function generateClientScope(
```
loganfsmyth: Missing `// @flow`

src/utils/pause/mapScopes/isReliableScope.js
```diff
@@ -0,0 +1,27 @@
+/**
```
loganfsmyth: Missing `// @flow`

src/utils/pause/mapScopes/findGeneratedBindingFromPosition.js
```diff
@@ -341,22 +339,31 @@ function mappingContains(mapped, item) {
...
+  generatedLocations: any
 ): Promise<{
   start: Location,
   end: Location
 } | null> {
-  const start = await sourceMaps.getGeneratedLocation(pos.start, source);
-  const end = await sourceMaps.getGeneratedLocation(pos.end, source);
+  // console.log(`> getting locations ${pos.start.line} ${pos.end.line}`);
+  const start = getGeneratedLocation(pos.start, generatedLocations);
+  const end = getGeneratedLocation(pos.end, generatedLocations);
```
loganfsmyth: We may want to delay this, because this is the logic I'm already going to be changing to updating how the ranges work. Maybe it would make more sense for me to make that a batched lookup?

jasonLaster: that works!

PR #5718, opened 2018-03-19
(WIP) Upgrade to flow 68

nyrosmith:
Fixes Issue: #5687 

* Updated to flow 68
* added failing types flow complained

PR #5893, opened 2018-04-09
 Upgrade to flow 68

darkwing:
Supercedes @nyrosmith's awesome https://github.com/devtools-html/debugger.html/pull/5718


darkwing:
Closing this not because it's bad work but because it's only going to go stale since it's not presently a priority -- myself or someone else can reopen and use @nyrosmith 's awesome work at the appropriate time 👍 💯

PR #5897, opened 2018-04-10
[Flow] src/utils/editor/source-editor.js 

atwalg2:
Fixes Issue: #5120

### DIFF
--- a/src/utils/editor/source-editor.js
+++ b/src/utils/editor/source-editor.js
@@ -2,6 +2,8 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+// @flow
+
 /**
  * CodeMirror source editor utils
  * @module utils/source-editor
@@ -70,7 +72,7 @@ export default class SourceEditor {
     }
   }
 
-  get codeMirror(): any {
+  get codeMirror(): CodeMirror {
     return this.editor;
   }
 
@@ -78,7 +80,7 @@ export default class SourceEditor {
     this.editor.setValue(str);
   }
 
-  getText() {
+  getText(): any {
     return this.editor.getValue();
   }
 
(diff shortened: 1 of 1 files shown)


src/utils/editor/source-editor.js
```diff
@@ -70,15 +72,15 @@ export default class SourceEditor {
     }
   }
 
-  get codeMirror(): any {
+  get codeMirror(): CodeMirror {
```
AnshulMalik: Umm, not sure if CodeMirror is a type defined somewhere. We probably need to get its types from somewhere.

darkwing: IMO we should just use `object`; defining props for third party libraries seems unnecessary.

jasonLaster: it is really nice to have 3rd party types because then we have better API docs :)

atwalg2:
adding flow here causes some weird behavior in components/editor/index.

jasonLaster:
@AnshulMalik @ryanjduffy thoughts?

jasonLaster:
looks like getSource should be looser and handle nulls `?SourceId`
@AnshulMalik @ryanjduffy thoughts?

AnshulMalik:
Yeah, we probably need to make it `?editor`

atwalg2:
image

AnshulMalik:
Well, this is strange

darkwing:
I had a go at this, albeit it with a major headache, and ran into a hornets nest of issues.  Will look again tomorrow.

atwalg2:
Awesome. Yea this is a weird bug I'm gonna go at it tonight and tomorrow as well

ryanjduffy:
In this case, I think you're invalidating the type refinement. Flow doesn't know if `this.state.editor` is still valid after `resizeBreakpointGutter` since it _could_ have altered the object. The solution is to extract the `codeMirror` member after the check but before the function calls.

https://flow.org/en/docs/lang/refinements/#toc-refinement-invalidations

```js
const { codeMirror } = this.state.editor;

resizeBreakpointGutter(codeMirror);
resizeToggleButton(codeMirror);
```

jasonLaster:
closing as this looks stuck. Feel free to re-open it if you figure it out...

Issue #5552, opened 2018-02-28
Highlighted SourceTree item loses highlight when ContextMenu -> Copy Source URI is clicked


darkwing:
Clicking a context menu item within the source tree makes the selected source lose focus.

darkwing:
A few more notes:

1.  The focus is definitely lost during the context menu/item's `onMousedDown` event.

2.  The focus is even lost when the mouse-down'd context menu item is disabled, making me think it's an entire context menu thing.

jasonLaster:
this might be a launchpad thing, but not a panel thing

darkwing:
This happens within the panel as well

PR #5900, opened 2018-04-10
[WIP] [Sources] keep focus on source item if copying source uri

mutdmour:
Fixes Issue: #5552 

Here's the Pull Request Doc

### Summary of Changes

* check if popup is shown before unfocusing an item

### Test Plan

- [x] Opening context menu does not remove focus
- [x] Closing context menu by clicking on `Copy Source Uri` item does not remove focus
- [x] Closing context menu by clicking outside of the menu does not remove focus
- [x] Keyboard navigation still works afterward

Feel free to improve it!

### Screenshots/Videos 
http://recordit.co/10Fok8kU2jv

### DIFF
--- a/src/components/shared/ManagedTree.js
+++ b/src/components/shared/ManagedTree.js
@@ -143,6 +143,10 @@ class ManagedTree extends Component<Props, State> {
   }
 
   focusItem = (item: Item) => {
+    const isPopupShown = document.querySelector("#contextmenu-mask.show");
+    if (isPopupShown !== null && typeof item === "undefined") {
+      return;
+    }
     if (!this.props.disabledFocus && this.state.focusedItem !== item) {
       this.setState({ focusedItem: item });
 
--- a/src/components/shared/ManagedTree.js
+++ b/src/components/shared/ManagedTree.js
@@ -144,7 +144,7 @@ class ManagedTree extends Component<Props, State> {
 
   focusItem = (item: Item) => {
     const isPopupShown = document.querySelector("#contextmenu-mask.show");
-    if (isPopupShown !== null && typeof item === "undefined") {
+    if (isPopupShown != null && typeof item == "undefined") {
       return;
     }
     if (!this.props.disabledFocus && this.state.focusedItem !== item) {


src/components/shared/ManagedTree.js
```diff
@@ -143,6 +143,10 @@ class ManagedTree extends Component<Props, State> {
   }
 
   focusItem = (item: Item) => {
+    const isPopupShown = document.querySelector("#contextmenu-mask.show");
+    if (isPopupShown !== null && typeof item === "undefined") {
```
lukaszsobek: Not really sure what the state of `isPopupShown` is. In case it's just null or true we shouldn't really need a tripple equal, if on the other hand it can be `false`, `true` or `null` and we need to differentiate between `null` and `false` we do actually need the === here. Again, this is before looking into the code, so it's a general comment.

Checking against the string type of "undefined" we can omit the === here and replace it with a == since this equation never needs to be coerced. (or in other words: since `typeof` always returns a string and we compare it to a string `"undefined"` we don't need the additional type conversion that would result from ===)

mutdmour: Oh I just use `===` out of habit. From [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelector), it's either `null` or an object. So it makes sense in both cases just to use `==`..

darkwing:
Wow, this is some clever thinking! 👍   Let me think on this a bit more; this change breaks keyboard navigation of the tree but that may not be avoidable.  Nice work, I'll let you know what I found.

darkwing:
Another option could be:

```js
const copySourceUri2 = {
    id: "node-menu-copy-source",
    label: copySourceUri2Label,
    accesskey: copySourceUri2Key,
    disabled: false,
    click: () => {
        copyToTheClipboard(source);
        this.focusItem(item);
        this.setState({ highlightItems: [item] });
    }
};
```

But that "flashes" the item (i.e. the item loses highlight and is then immediately reapplied).  Still thinking on all of this.

PR #5901, opened 2018-04-10
Fix memory leak by setting editor to null

darkwing:
As discussed in meeting.


darkwing:
Thanks for the tweak!

PR #5904, opened 2018-04-10
[Flow] src/utils/function.js

atwalg2:
Fixes Issue: #5120

SymbolDeclarations? or should I just put Object?

### DIFF
--- a/src/utils/ast.js
+++ b/src/utils/ast.js
@@ -8,21 +8,20 @@ import { xor, range } from "lodash";
 import { convertToList } from "./pause/pausePoints";
 
 import type { Location, Source, ColumnPosition } from "../types";
+import type { Symbols } from "../reducers/ast";
 
-import type {
-  AstPosition,
-  AstLocation,
-  PausePoints,
-  SymbolDeclarations
-} from "../workers/parser";
+import type { AstPosition, AstLocation, PausePoints } from "../workers/parser";
 
 export function findBestMatchExpression(
-  symbols: SymbolDeclarations,
+  symbols: Symbols,
   tokenPos: ColumnPosition
 ) {
-  const { memberExpressions, identifiers, literals } = symbols;
-  const { line, column } = tokenPos;
+  if (symbols.loading) {
+    return null;
+  }
 
+  const { line, column } = tokenPos;
+  const { memberExpressions, identifiers, literals } = symbols;
   const members = memberExpressions.filter(({ computed }) => !computed);
 
   return []
@@ -107,18 +106,18 @@ function findClosestofSymbol(declarations: any[], location: Location) {
   }, null);
 }
 
-export function findClosestFunction(
-  symbols: SymbolDeclarations,
-  location: Location
-) {
-  const { functions } = symbols;
-  return findClosestofSymbol(functions, location);
+export function findClosestFunction(symbols: ?Symbols, location: Location) {
+  if (!symbols || symbols.loading) {
+    return null;
+  }
+
+  return findClosestofSymbol(symbols.functions, location);
 }
 
-export function findClosestClass(
-  symbols: SymbolDeclarations,
-  location: Location
-) {
-  const { classes } = symbols;
-  return findClosestofSymbol(classes, location);
+export function findClosestClass(symbols: Symbols, location: Location) {
+  if (!symbols || symbols.loading) {
+    return null;
+  }
+
+  return findClosestofSymbol(symbols.functions, location);
 }
--- a/src/utils/function.js
+++ b/src/utils/function.js
@@ -2,21 +2,28 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
 
+// @flow
 import { findClosestFunction } from "./ast";
 import { correctIndentation } from "./indentation";
+import type { Source } from "../types";
+import type { Symbols } from "../reducers/ast";
 
-export function findFunctionText(line, source, symbols) {
+export function findFunctionText(
+  line: number,
+  source: Source,
+  symbols: ?Symbols
+): ?string {
   const func = findClosestFunction(symbols, {
+    sourceId: source.id,
     line,
     column: Infinity
   });
-  if (!func) {
+
+  if (!func || !source.text) {
     return null;
   }
 
-  const {
-    location: { start, end }
-  } = func;
+  const { location: { start, end } } = func;
   const lines = source.text.split("\n");
   const firstLine = lines[start.line - 1].slice(start.column);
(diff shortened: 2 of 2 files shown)


src/utils/function.js
```diff
@@ -2,10 +2,15 @@
...
+// @flow
 import { findClosestFunction } from "./ast";
 import { correctIndentation } from "./indentation";
 
-export function findFunctionText(line, source, symbols) {
+export function findFunctionText(
+  line: number,
+  source: any,
+  symbols: SymbolDeclarations
+): string | null {
```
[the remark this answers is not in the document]
atwalg2: replace any with any?

atwalg2:
ok so issue is here:

```
export function findFunctionText(
  line: number,
  source: any,
  symbols: SymbolDeclarations
): string | null {
  const func = findClosestFunction(symbols, {
    line,
    column: Infinity
  });
  if (!func) {
    return null;
  }
```

the second arguemnt of findClosestFunction is a location type.. i.e 
```
export type Location = {
  sourceId: SourceId,
  line: number,
  column: ?number,
  sourceUrl?: string
};
```

simple solution is to make sourceId an optional param. Thoughts?

src/utils/function.js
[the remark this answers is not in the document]
jasonLaster: oops, i meant Source

src/types.js
```diff
@@ -54,7 +54,7 @@ export type ActorId = string;
  * @static
  */
 export type Location = {
-  sourceId: SourceId,
+  sourceId?: SourceId,
```
jasonLaster: sorry, this shouldnt be optional, the param for the reducer should be

atwalg2: This is still a problem if I change this back and edit the reducer param. 
In ast.js, location is declared multiple times and if sourceId is a required param then it should be included.. seperated... or made optional

atwalg2: image

src/utils/function.js
```diff
@@ -2,25 +2,39 @@
...
-  if (!func) {
+  if (symbols) {
+    func = findClosestFunction(symbols, {
+      sourceId: source.id,
+      line,
+      column: Infinity
+    });
+  } else {
+    return null;
+  }
```
jasonLaster: small suggestion here. I think i'd prefer putting the guard statement in `findClosestFunction`. We prefer safer calls than checks before the calls

```diff
diff --git a/src/utils/ast.js b/src/utils/ast.js
index 56c30f3..6464345 100644
--- a/src/utils/ast.js
+++ b/src/utils/ast.js
@@ -104,17 +104,25 @@ function findClosestofSymbol(declarations: any[], location: Location) {
 }
 
 export function findClosestFunction(
-  symbols: SymbolDeclarations,
+  symbols: ?SymbolDeclarations,
   location: Location
 ) {
+  if (!symbols) {
+    return null;
+  }
+
   const { functions } = symbols;
   return findClosestofSymbol(functions, location);
 }
 
 export function findClosestClass(
-  symbols: SymbolDeclarations,
+  symbols: ?SymbolDeclarations,
   location: Location
 ) {
+  if (!symbols) {
+    return null;
+  }
+
   const { classes } = symbols;
   return findClosestofSymbol(classes, location);
 }
```

src/reducers/ast.js
```diff
@@ -151,7 +151,10 @@ function update(
 // https://github.com/devtools-html/debugger.html/blob/master/src/reducers/sources.js#L179-L185
 type OuterState = { ast: Record<ASTState> };
 
-export function getSymbols(state: OuterState, source: Source): ?Symbols {
+export function getSymbols(
+  state: OuterState,
+  source: Source
+): ?SymbolDeclarations {
```
jasonLaster: It's important that this is `Symbols` and not `SymbolDeclarations` look at the definition above

atwalg2: hey @jasonLaster were you referring to only this case or the other ones as well. I may have read this wrong initially.. I just realized you may have meant only this case and not the others .. oops. Please clarify this for me I can make a quick fix

src/utils/function.js
```diff
@@ -2,25 +2,34 @@
...
   }
 
   const { location: { start, end } } = func;
-  const lines = source.text.split("\n");
-  const firstLine = lines[start.line - 1].slice(start.column);
-  const lastLine = lines[end.line - 1].slice(0, end.column);
-  const middle = lines.slice(start.line, end.line - 1);
-  const functionText = [firstLine, ...middle, lastLine].join("\n");
-  const indentedFunctionText = correctIndentation(functionText);
+  const lines: string[] = source.text.split("\n");
```
AnshulMalik: Do we need these types?

src/utils/ast.js
```diff
@@ -7,21 +7,22 @@
...
-  symbols: SymbolDeclarations,
+  symbols: Symbols,
   tokenPos: ColumnPosition
 ) {
-  const { memberExpressions, identifiers, literals } = symbols;
-  const { line, column } = tokenPos;
+  if (
+    !(symbols.memberExpressions && symbols.identifiers && symbols.identifiers)
+  ) {
+    return null;
```
AnshulMalik: This means both `memberExpressions` and `identifiers` should be present for this to return results. Is this what we want?

jasonLaster: > !(symbols.memberExpressions && symbols.identifiers && symbols.identifiers)

could we just do `!symbols`

src/utils/function.js
```diff
@@ -2,25 +2,34 @@
...
+  const firstLine: string = lines[start.line - 1].slice(start.column);
+  const lastLine: string = lines[end.line - 1].slice(0, end.column);
+  const middle: string[] = lines.slice(start.line, end.line - 1);
+  const functionText: string = [firstLine, ...middle, lastLine].join("\n");
+  const indentedFunctionText: string = correctIndentation(functionText);
```
jasonLaster: do we need these for it to pass? they should be inferred...

src/utils/ast.js
atwalg2: well no because as per type def for Symbols it can be either SymbolDeclaration | {| loading: true |}. So we have to account for the other case. Really it seems like we are only dealing with SymbolDeclarations here. What do you think?

src/utils/function.js
atwalg2: No we don't. I just put them in their to try and improve the file coverage, which most likely wouldn't affect it.

src/utils/function.js
atwalg2: oh ok I can get rid of these. They are def unnecessary.

src/utils/ast.js
jasonLaster: so `symbols.identifiers && symbols.identifiers` is unneccessary right?

could we do `!symbols.loading && symbols`

PR #5924, opened 2018-04-11
Use addons icons for extensions

darkwing:
I thought it would be cool to show an icon for Firefox extensions:

extensions


jasonLaster:
I like it. What if we changed moz-extension:// to "Extensions"...

While you'r in there we should fix angular:// to be consistent with webpack's scheme :wink:

jasonLaster:
lol you already fixed angular

jasonLaster:
Also, this feels a bit big, might be fine though

darkwing:
Updated the text value and icon size:

extensionsicon

PR #5929, opened 2018-04-11
Fix blank debugger when debugging service workers

juliandescottes:
Service worker debugging was broken with https://github.com/devtools-html/debugger.html/pull/5761 , my bad! Looks like not all debugging target have a root client defined.


juliandescottes:
Thanks for merging!

PR #5930, opened 2018-04-11
Fix App.js import order

jasonLaster:


src/components/App.js
```diff
@@ -139,18 +138,17 @@ class App extends Component<Props, State> {
...
     } = this.props;
 
     if (activeSearch) {
       e.preventDefault();
       closeActiveSearch();
     }
 
-    if (quickOpenEnabled) {
-      e.preventDefault();
+    if (quickOpenEnabled === true) {
```
AnshulMalik: Is there any reason for this change?

Sekhmet: Looks like this reverted this PR: https://github.com/devtools-html/debugger.html/pull/5866/files

jasonLaster: oops. that should be left out!

PR #5934, opened 2018-04-11
Move debugBtn to a better spot

darkwing:
I thought I could let this go but it bugged me so I just changed it quick.


jasonLaster:
lol this is better

PR #5942, opened 2018-04-12
Use mc help svg

jasonLaster:
### summary

this updates the svg logic so we can re-use mc svgs

### DIFF
--- a/src/components/SecondaryPanes/CommandBar.css
+++ b/src/components/SecondaryPanes/CommandBar.css
@@ -92,7 +92,7 @@ img.shortcuts {
 }
 
 .command-bar img.shortcuts {
-  mask: url(/images/help.svg) no-repeat;
+  mask: url(/mc/help.svg) no-repeat;
 }
 
 .command-bar .replay-inactive {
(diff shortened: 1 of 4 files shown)


jasonLaster:
:thumbsup:

darkwing:
Added a second commit to remove the asset.  I'm still not seeing the icon though.

darkwing:
Whether a comedy of errors on my part or legit weirdness, using `/mc/` created nothing but problems for me; I was thinking the conflict was here:

```js
.replace(/(chrome:\/\/|resource:\/\/)/, "/mc/")
```

But no changes I made fixed the issue.  I'm also a bit dubious about using `mc` in our CSS, as most contributors wont know what it represents; I think a more explicit path would be helpful.

src/components/SecondaryPanes/CommandBar.css
```diff
@@ -92,7 +92,7 @@ img.shortcuts {
 }
 
 .command-bar img.shortcuts {
-  mask: url(/images/help.svg) no-repeat;
+  mask: url(/mozillacentral/help.svg) no-repeat;
```
jasonLaster: weird maybe the problem was that images should have been mc

jasonLaster:
Could you check one more time:

PR #5945, opened 2018-04-12
 Made the close button an actual button

BadInternetGuy:
Fixes Issue: #4070 

This changes the div to an actual button so that it will work with a screen reader. There is still a slight issue with the button, as the padding makes it seem larger than the image when it is pressed.

Ctrl+Shift+F to open a search bar. 
Then simply press the close button. 
Padding makes the button slightly to large.

clickedbutton

### DIFF
--- a/src/components/shared/SearchInput.js
+++ b/src/components/shared/SearchInput.js
@@ -132,7 +132,7 @@ class SearchInput extends Component<Props> {
       selectedItemId,
       showErrorEmoji,
       size,
-      summaryMsg
+      summaryMsg,
     } = this.props;
 
     const inputProps = {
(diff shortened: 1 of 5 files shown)


jasonLaster:
Hi, please update the description

BadInternetGuy:
This change changes the Div to a button so it can be read by a screen reader. The button works, although there still seems to be a slight issue with the padding around it, as the actual button appears to be larger than the image attached to it.

src/components/shared/SearchInput.js
```diff
@@ -132,7 +132,7 @@ class SearchInput extends Component<Props> {
       selectedItemId,
       showErrorEmoji,
       size,
-      summaryMsg
+      summaryMsg,
```
darkwing: We can remove the extra `,`

darkwing:
Closing as this is errantly from a local `master` branch -- I look forward to seeing a new PR on a feature branch! 👍

Issue #5581, opened 2018-03-01
[Proposal] remove chrome support


jasonLaster:
The debugger has had 80% chrome/cdp support since June of 2016. We've hoped to get to CDP support in 3-6 months ever since June of 2016. 

I think we should acknowledge that this is not on the 3-6 month roadmap and we should remove it for now. It will still be in git, so it won't be hard to bring back when the time is right.

### What would removing it look like:

1. the `src/client` directory will be simpler
2. the `src/components/../ChromeScopes` component will be removed
3. the `src/reducer/pause` will be simpler
3. the launchpad won't show chrome/node debugging options by default

### Will we ever get to Chrome support? 

I believe we will. CDP is still a Mozilla goal. We believe standardizing around a single protocol is a priority for editors and the web. We also believe there is an advantage for us supporting CDP, which would improve our tooling integration with editors like vs code. There is a Firefox adapter, but that is difficult for others to maintain.

I would also like to see us offer a JS Debugger at some point. I personally think that the Node community and JS community would like to see another option. And it would be great for the Mozilla brand to introduce a new product just for developers. But, this is a longer term effort that will need to be planned for.

lukaszsobek:
A standalone debugger? Count me in on that one!

Back on topic:
I think hiding the chrome support for now is a good option in terms of trading it for more focus on debugger being a great product.

AnshulMalik:
I think we should keep the chrome protocol, there is no harm? or there is?

bomsy:
Since we are at 80% could this be removed from master and moved to dedicated branch where anyone interested can still be hacking and making slow progress ...having said that , it will probably need regular updates with master.

I guess one of the other reasons to remove it for now would be reduced build size ... right?

vxcamiloxv:
It should remove Node.js tab (and from main page and docs) if no have plans to support it, that create a fake expectation: https://github.com/devtools-html/debugger.html/issues/5739#issuecomment-374972414

btw, could be great have an alternative to chrome/chromium to debug Node.js apps.

PR #5950, opened 2018-04-12
Remove Chrome Scopes

jasonLaster:
#5581

### Summary of Changes

- drops chrome scopes
- we can bring it back when we do v8 support
- we might likely want to do it over from scratch though


darkwing:
I'm seeing reference to "Chrome Scopes" in:
- [ ] `configs/ci.json`
- [ ] `configs/local.sample.json`
- [ ] `src/utils/prefs.js`

Are these config files still relevant?

jasonLaster:
lets get rid of all of these

Issue #5951, opened 2018-04-12
[Pause] debugger doesn’t show the pause state


jasonLaster:
Sorry for the delay... I'm finding a different but equally scary bug,

0. With Firefox Desktop Responsive Design mode. (window 360px/640px, Firefox Android UA, touch activated). 61.0a1 (2018-03-27) (64-bit)
1. Go to http://siberiastationspa.com/infos-pratiques/
2. Open the debugger
3. Find `function initHashClick()` in `default.min.js`
4. Set a breakpoint at the beginning of the code in initHashClick(). See screenshot
5. Go back to the viewport window.
6. click on the 3 stripes menu

I expect to see the menu, but instead i think the debugger pauses but doesnt tell us why...

![Screen Shot 2018-04-12 at 2.20.33 PM.png](https://shipusercontent.com/8b0705f8afac9deac663f69519bd5d19/Screen%20Shot%202018-04-12%20at%202.20.33%20PM.png)

jasonLaster:
Hmm, it is now pausing for me. the location is a bit funky, but i think that is because of pretty printed source maps.

PR #5954, opened 2018-04-12
Use pausePoints to determine valid pause locations

jasonLaster:
Fixes Issue: #5951

### Summary of Changes

- previously isInValidLocation would not respond in some cases 
- now we use pause points which is pre-computed
- pause points has some new empty points to account for places we don't want to pause


src/workers/parser/pausePoints.js
```diff
@@ -11,15 +11,20 @@ import isEqual from "lodash/isEqual";
 import type { BabelNode } from "@babel/types";
 import type { SimplePath } from "./utils/simple-path";
 
+const isForStatement = node =>
+  t.isForStatement(node) || t.isForOfStatement(node);
```
darkwing: Oh nice!

darkwing:
Code looks solid; do we have any manual testing steps that can be done reliably?

jasonLaster:
manually uses babel-stepping

http://localhost:8000/integration/examples/doc-babel.html

Issue #3792, opened 2017-08-27
[CallStack] show react components


jasonLaster:
Create React App has begun showing Error Stacks with component information. The benefit of this UI is that it shows users the component tree so they can see where something went wrong.

It would be great if we could show the component names in the call stack. In the screenshot below I include the react render frames, but those might not be useful

@sebmarkbage has a spec for custom stack frames [here](https://gist.github.com/sebmarkbage/bdefa100f19345229d526d0fdd22830f).

---

### Component Call Stack

![Screen Shot 2017-08-27 at 10.34.23 AM.png](https://shipusercontent.com/2044bfcfa7435966b4e83d447d7209c8/Screen%20Shot%202017-08-27%20at%2010.34.23%20AM.png)

### Create React App 

![Screen Shot 2017-08-27 at 10.00.42 AM.png](https://shipusercontent.com/3987fa490eba7c76175f8f5b5093dbee/Screen%20Shot%202017-08-27%20at%2010.00.42%20AM.png)

![](https://video.twimg.com/tweet_video/DIH2113UAAQ3SDA.mp4)

jasonLaster:
cc @gaearon @bvaughn

gaearon:
To be clear we haven't enabled this in React yet but I'm working on it.

jasonLaster:
moving this to the icebox. we can re-evaluate in the next release cycle

jasonLaster:
Here's an initial re-write of the components pane, with the goals of

- separating out components from extra
- adding children
- making it behave like a tree (WIP - we should use the Tree component)

https://github.com/jasonLaster/debugger.html/tree/components

jasonLaster:
here is a branch for the component tree - https://github.com/devtools-html/debugger.html/compare/master...jasonLaster:component-tree?expand=1

jasonLaster:
Moving to the icebox

PR #5955, opened 2018-04-12
WIP - React component stack

calebstdenis:
The beginnings of #3792 

#### Done
* Add list of React component names in the Redux store (`pause.reactComponentStack`). Consists of the current component being paused in, and all its ancestors up the root component. 
#### To-Do
* Get source file locations for each component 
* Prototype a UI
* Anything else?

### Test Plan
1. Hit a breakpoint at [Frame.js#115](https://github.com/devtools-html/debugger.html/blob/b079ab8c167f0e41004a14aa905c1c383c5e103f/src/components/SecondaryPanes/Frames/Frame.js#L115)
2. Check Redux store
image

### DIFF
--- a/src/actions/pause/paused.js
+++ b/src/actions/pause/paused.js
@@ -21,7 +21,8 @@ import { selectLocation } from "../sources";
 import { loadSourceText } from "../sources/loadSourceText";
 import { togglePaneCollapse } from "../ui";
 import { command } from "./commands";
-import { shouldStep } from "../../utils/pause";
+import { getReactComponentStack, shouldStep } from "../../utils/pause";
+import { isReactComponent } from "../../utils/preview";
 
 import { updateFrameLocation } from "./mapFrames";
 
@@ -33,6 +34,7 @@ import type { ThunkArgs } from "../types";
 async function getOriginalSourceForFrame(state, frame: Frame) {
   return getSources(state).get(frame.location.sourceId);
 }
+
 /**
  * Debugger has just paused
  *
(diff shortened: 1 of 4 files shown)


src/actions/pause/paused.js
```diff
@@ -44,8 +46,13 @@ export function paused(pauseInfo: Pause) {
   return async function({ dispatch, getState, client, sourceMaps }: ThunkArgs) {
     const { frames, why, loadedObjects } = pauseInfo;
     const rootFrame = frames.length > 0 ? frames[0] : null;
+    let reactComponentStack;
 
     if (rootFrame) {
+      if (isReactComponent(rootFrame.this)) {
+        reactComponentStack = await getReactComponentStack(rootFrame, client);
+      }
```
jasonLaster: i think we should try to put this in the `extra` field. that way we have one place for react stuff

src/utils/pause/getReactComponentStack.js
```diff
@@ -0,0 +1,21 @@
...
+      `this._reactInternalFiber${new Array(i)
+        .fill("._debugOwner")
+        .join("")}.type`
+    );
+    const { type, userDisplayName, displayName } = reactInternalType.result;
+    if (type === "undefined") {
+      break;
+    }
+    components.push(userDisplayName || displayName);
+    i++;
```
jasonLaster: cute, but we should minimize the number of calls :) so do this in one evaluate

calebstdenis: tru dat

src/actions/pause/paused.js
calebstdenis: will do

PR #5956, opened 2018-04-12
Consolidate extra fields

jasonLaster:
### Summary of Changes

organizes all the extra stuff in one place

### Test

- pause in a react component
- check to see if the component pane is there w/ the name


src/actions/pause/extra.js
```diff
@@ -0,0 +1,89 @@
...
+import { isReactComponent, isImmutable } from "../../utils/preview";
+import { findClosestClass } from "../../utils/ast";
+
+import type { ThunkArgs } from "../types";
+
+async function getReactProps(evaluate) {
+  const reactDisplayName = await evaluate(
+    "this.hasOwnProperty('_reactInternalFiber') ? " +
+      "this._reactInternalFiber.type.name : " +
+      "this._reactInternalInstance.getName()"
```
calebstdenis: This is pre-existing code but I'm confused. The function is called `getReactProps`, but doesn't it just return the component name as a string?

src/actions/pause/extra.js
```diff
@@ -0,0 +1,89 @@
...
+    if (symbols && symbols.classes) {
+      const originalClass = findClosestClass(symbols, selectedFrame.location);
+
+      if (originalClass) {
+        props.react = { displayName: originalClass.name };
+      }
+    }
+
+    if (!props.react) {
+      props.react = await getReactProps(evaluate);
```
mmcote: I know originally when I was adding to this function we kept this evaluate function here as a safety net just in case the closest class can't be found, although we thought that going through the symbols should be the standard, as it is, and that this safety-net may not really be needed. Could this be removed, and therefore be able to remove the getReactProps all together?

darkwing:
So here's what I've done:

0.  Using debugger.html as my React app -- wow!
1.  In `App.js`, I set a breakpoint at the fist line of the `render()` method.
2.  Reload source debugger.html app so the breakpoint will be hit in other browser
3.  Breakpoint is hit, but no Component panel displays and `props.react` is `undefined`
4.  `dbg.store.getState().pause.extra` is empty

nopane

Could I be testing this incorrectly?

src/actions/pause/extra.js
jasonLaster: yeah - we'll definitely simplify this

src/actions/pause/extra.js
calebstdenis: woops I skipped over the lines below this. Makes total sense now

PR #5960, opened 2018-04-12
Cleanup the ember test

jasonLaster:
This change makes it pretty hard to update the ember test we have, but it should be sufficient for our current use cases and perhaps some new perf tests in the future...


jasonLaster:
If we need to make a change to the bundle, i can go back to this commit, make some tweaks and cp the bundles somewhere 😉

PR #5967, opened 2018-04-13
[Scopes] Allow bindings to not start at the mapping start in some cases, and optimize performance of mapping range loading

loganfsmyth:
Refs Issue: #5561

### Summary of Changes

* Fix two misc scope visitor bugs
* Dispatch all of the sourceMap worker requests synchronously so they queue into a single message, making them much faster
* Use the new sourceMaps.getGeneratedRanges to get the original range
* Make use of the filtered range list to allow more broad mapping interpretation when a binding is the first match


src/actions/pause/mapScopes.js
```diff
@@ -85,6 +86,65 @@ export function mapScopes(scopes: Promise<Scope>, frame: Frame) {
...
+      }
+    }
+  }
+
+  return {
+    async getGeneratedRanges(pos, s) {
+      const key = buildLocationKey(pos);
+
+      if (s !== source || !precalculatedRanges.has(key)) {
+        throw new Error("Bad precalculated mapping");
```
jasonLaster: so... do we want to throw in firefox? if not we should use `utils/assert`

i guess we are catching this globally...

src/test/mochitest/examples/doc-babel.html
```diff
@@ -29,6 +29,8 @@
     <button onclick="functions()">Run functions</button>
     <script src="babel/fixtures/imported-bindings/output.js"></script>
     <button onclick="importedBindings()">Run importedBindings</button>
+    <script src="babel/fixtures/line-start-bindings-es6/output.js"></script>
+    <button onclick="lineStartBindingsEs6()">Run lineStartBindingsEs6</button>
```
jasonLaster: It would be nice to add documentation here maybe so that it is clear what each case is for..

| example | description  |
| -- | -- |
| lineStartBindingsEs6 | tests es6 bindings on many lines to see if ... |
| ... | .. |

Presumably this doc would be with a `<table><tr><td>` markup :)

src/actions/pause/mapScopes.js
loganfsmyth: Mostly this is to catch programming errors. I can't really imagine a case where this would throw in production but not during whatever work we're doing. If we did want to avoid this though, I'd probably say we update it to throw when `isDevelopment` and otherwise just call through to `sourceMaps` to do the query without batching.

src/test/mochitest/examples/doc-babel.html
loganfsmyth: Yeah sounds like a good idea.

loganfsmyth: Should I just add that to the tracking issue?

jasonLaster: already did.

src/actions/pause/mapScopes.js
jasonLaster: yeah, lets do that... small fix

loganfsmyth: Updated!

PR #5969, opened 2018-04-13
[WIP] Stopped highlighting the current line location on reload.

AlexisDeschamps:
Fixes Issue: #5633

### Summary of Changes
* Made `HighlightLine` component only set the highlight if it `isStepping` or the location was requested by clicking on a breakpoint
* Made breakpoint-clicking give the property of `isLocatingBreakpoint` to `selectedLocation` object

### Test Plan
- [x] Open http://firefox-dev.tools/debugger-examples/examples/todomvc/ as debugee
- [x] In `Sources`: open `firefox-dev.tools`, open `debugger-examples/examples/todomvc`, open `bower_components`, open `js`, open `views`, open `app-view.js`
- [x] Put a breakpoint on line 31 of `app-view.js`
- [x] Refresh http://firefox-dev.tools/debugger-examples/examples/todomvc/
- [x] Notice the current location line is not highlighted in grey

### Videos
Before
before

After
after

### DIFF
--- a/src/components/Editor/HighlightLine.js
+++ b/src/components/Editor/HighlightLine.js
@@ -61,14 +61,15 @@ export class HighlightLine extends Component<Props> {
     const { sourceId, line } = selectedLocation;
     const editorLine = toEditorLine(sourceId, line);
 
+    if (!this.isStepping && !selectedLocation.isLocatingBreakpoint) {
+      return false;
+    }
     if (!isDocumentReady(selectedSource, selectedLocation)) {
       return false;
     }
-
     if (this.isStepping && editorLine === this.previousEditorLine) {
       return false;
     }
-
     return true;
   }
 
@@ -100,15 +101,14 @@ export class HighlightLine extends Component<Props> {
       return;
     }
     this.isStepping = false;
-    const editorLine = toEditorLine(sourceId, line);
-    this.previousEditorLine = editorLine;
+    this.previousEditorLine = toEditorLine(sourceId, line);
 
     if (!line || isDebugLine(selectedFrame, selectedLocation)) {
       return;
     }
 
     const doc = getDocument(sourceId);
-    doc.addLineClass(editorLine, "line", "highlight-line");
+    doc.addLineClass(this.previousEditorLine, "line", "highlight-line");
   }
 
   clearHighlightLine(selectedLocation: Location, selectedSource: SourceRecord) {
--- a/src/actions/sources/select.js
+++ b/src/actions/sources/select.js
@@ -91,7 +91,10 @@ export function selectSource(sourceId: string) {
  * @memberof actions/sources
  * @static
  */
-export function selectLocation(location: Location) {
+export function selectLocation(
+  location: Location,
+  isLocatingBreakpoint: boolean = false
+) {
   return async ({ dispatch, getState, client }: ThunkArgs) => {
     if (!client) {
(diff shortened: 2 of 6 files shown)


AlexisDeschamps:
I'm not sure if https://github.com/devtools-html/debugger.html/blob/master/src/test/mochitest/browser_dbg-editor-highlight.js should be removed if no locations are supposed to be highlighted when a source is selected.

jasonLaster:
hmm perhaps

jasonLaster:
so i'm hesitant to add a new field to the redux store...

I think we can use `pendingSelectedLocation` as well as the previous `selectedLocation` to figure out if we should show a highlight;

This patch takes advantage of the fact that we dont want to show a highlight if the line hasn't changed... 


```diff
diff --git a/src/components/Editor/HighlightLine.js b/src/components/Editor/HighlightLine.js
index 98e66b3..e641edf 100644
--- a/src/components/Editor/HighlightLine.js
+++ b/src/components/Editor/HighlightLine.js
@@ -37,6 +37,10 @@ function isDebugLine(selectedFrame: Frame, selectedLocation: Location) {
   );
 }

+function getLine(location: ?Location) {
+  return location && location.line;
+}
+
 function isDocumentReady(selectedSource, selectedLocation) {
   return (
     selectedLocation &&
@@ -51,6 +55,11 @@ export class HighlightLine extends Component<Props> {

   shouldComponentUpdate(nextProps: Props) {
     const { selectedLocation, selectedSource } = nextProps;
+
+    if (getLine(selectedLocation) == getLine(this.props.selectedLocation)) {
+      return false;
+    }
+
     return this.shouldSetHighlightLine(selectedLocation, selectedSource);
   }
```

I reviewed this [code][c] closely to see if we could take advantage of other behavior:

[c]: https://github.com/devtools-html/debugger.html/blob/master/src/reducers/sources.js#L84-L116

AlexisDeschamps:
@jasonLaster From what I tried, using `pendingSelectedLocation` works to remove highlights on reload. However, it does not allow for breakpoint highlighting if the breakpoint is at the currently selected location. I'm not sure if there's some kind of data that would allow for knowing if the currently selected location should still be highlighted if the update is caused by a breakpoint being clicked, besides what I did for this PR.

jasonLaster:
I'll live with that. haha one step forward at a time :)

src/components/Editor/HighlightLine.js
```diff
@@ -61,14 +61,15 @@ export class HighlightLine extends Component<Props> {
     const { sourceId, line } = selectedLocation;
     const editorLine = toEditorLine(sourceId, line);
 
+    if (!this.isStepping && !selectedLocation.isLocatingBreakpoint) {
```
jasonLaster: hmm, i'd like to find a way to avoid adding this field `selectedLocation.isLocatingBreakpoint`

sorry i dont have time to jump in and help though

AlexisDeschamps:
So, I did a `git bisect` It seems that this problem was actually fixed in #6042. It seems that the breakpoints are applied later.

The rest of the highlight functionality seems to work still as well. Everything seems to work as expected to on the latest master.

Closing this PR as a result, but I am up for hearing thoughts about the situation.

jasonLaster:
oh cool! That's pretty crazy, i'll give it a try soon

jasonLaster:
Yes! that definitely is fixed. phew!

Issue #5966, opened 2018-04-13
[Preview] isReactComponent isn't Properly Detecting React Components


darkwing:
The `isReactComponent` inside `src/utils/preview.js` isn't properly detecting React components.

`ownProperties` contains only a list of the component's methods and properties, of which `_reactInternalInstance` and `_reactInternalFiber` are not.  For example, App.js' ownProperties are:

```js
(10) […]
0: "props"
1: "context"
2: "refs"
3: "updater"
4: "getChildContext"
5: "onEscape"
6: "onCommandSlash"
7: "toggleQuickOpenModal"
8: "onLayoutChange"
9: "renderEditorPane"
​```

**Comment by jasonLaster** [comment]
So... I think it would be good to add a static way to see if we're in a react component. This logic formalizes some of these things by 

1. seeing if we're paused in a class
2. seeing if we're in a file w/ a react component
3. checking if we're in fact in a component class


```diff
diff --git a/src/selectors/inComponent.js b/src/selectors/inComponent.js
new file mode 100644
index 0000000..02f7f64
--- /dev/null
+++ b/src/selectors/inComponent.js
@@ -0,0 +1,23 @@
+const classes = ["Component", "PureComponent"];
+
+export function inComponent(state) {
+  const selectedFrame = getSelectedFrame(state);
+  const source = getSource(state, selectedFrame.location.sourceId);
+  const symbols = getSymbols(state, source);
+
+  const closestClass = findClosestClass(symbols, selectedFrame.location);
+  if (!closestClass) {
+    return null;
+  }
+
+  const sourceMetaData = getSourceMetaData(state, source.id);
+  if (!sourceMetaData) {
+    return;
+  }
+
+  const inReactFile = sourceMetaData.framework == "react";
+  const inComponent =
+    closestClass.parent && classes.includes(closestClass.parent.name);
+
+  return inReactFile && inComponent;
+}
```

PR #5971, opened 2018-04-13
Use inComponent to detect if the current pause is in a React component

darkwing:
Fixes Issue: #5966

I started a different approach but this was much cleaner.  @jasonLaster always wins.

## To Do:
- [ ] Replace all other instances of isReactComponent with this (?)


src/actions/tests/__snapshots__/preview.spec.js.snap
```diff
@@ -46,11 +46,7 @@ exports[`setPreview react instance 1`] = `
 Object {
   "cursorPos": undefined,
   "expression": "this",
-  "extra": Object {
-    "react": Object {
-      "displayName": "Foo",
-    },
-  },
+  "extra": Object {},
```
darkwing: Grrr.  `extra` is there when I debug; will look at this again.

src/actions/tests/__snapshots__/preview.spec.js.snap
```diff
@@ -46,14 +46,7 @@ exports[`setPreview react instance 1`] = `
...
   "expression": "this",
-  "extra": Object {
-    "react": Object {
-      "componentStack": Array [
-        "Foo",
-      ],
-      "displayName": "Foo",
-    },
-  },
+  "extra": Object {},
```
darkwing: Booo

darkwing:
With this change I now see the component's name panel in the right pane (yay) but I don't see any contents:

nocontents

Is this a side effect or unrelated?

Issue #5970, opened 2018-04-13
[Breakpoints] Show Special Icon in Group File Name


darkwing:
We have this awesome feature where we show a React icon if the file is a React file; maybe we should try showing the same icon in next to the file name in the breakpoints pane:

<img width="1182" alt="react" src="https://user-images.githubusercontent.com/46655/38753663-ba58f73e-3f24-11e8-9877-19a07b766f0a.png">

jasonLaster:
Yeah! Even the file icon might add a nice visual call out.

jasonLaster:
<img src='https://shipusercontent.com/264049eeacdeb8e0eff623f0097c5407/Screen%20Shot%202018-04-13%20at%205.06.29%20PM.png' title='Screen Shot 2018-04-13 at 5.06.29 PM.png' width=300>

darkwing:
Closing as dupe https://github.com/devtools-html/debugger.html/issues/6000

PR #5972, opened 2018-04-13
Style the right sidebar

jasonLaster:
Fixes Issue: #5970 

### Summary of Changes

* add file icon to file header in breakpoints pane
* consolidate logic for icons
* hide stepping buttons when we are not paused
* hide refresh button when there are 0 expressions
* remove breakpoint toggle button (this is probably pre-mature)

Next steps
* clean up the code
* style the other icon locations...
* maybe cache the framework data...

screen shot 2018-04-13 at 5 13 35 pm


jasonLaster:
I hid the bp toggle button here, but that was probably pre-mature. We should add it back until we have the new disable pausing button in the top right.

darkwing:
Superceded by: https://github.com/devtools-html/debugger.html/pull/5985

PR #5932, opened 2018-04-11
Add checkboxes for caught exceptions

darkwing:
Fixes Issue: #5795

exceptioncheckboxes

## ToDo

- [x] Wire checkbox states to their representative functionality

PR #5976, opened 2018-04-14
Fix l10n key (pauseOnExceptions)

jasonLaster:
#5932  #5972

Fixes l10n issue.

Is this the appropriate fix?


jasonLaster:
nope. this will land now and both will go into release 39

Issue #5603, opened 2018-03-05
[Breakpoints] show breakpoint on hover


jasonLaster:
I recently was looking at "Continue to Line" for time travel debugging. CtL got me thinking it about two things:

1) it would be nice to have a hover effect in the gutter. For instance, hovering on a line where you can add a breakpoint shows a light breakpoint. With continue to here, hovering on a line that you can jump to with the alt key pressed could show a light background color indicating you can jump there.

2) the way we currently show empty lines (greying out the line number). Can create a zebra effect that is quite distracting. Most of time, the user doesn't care which lines are empty or not, bu the color difference is always visible. 

Lastly, I think disabling some lines so that you can't add a breakpoint or jump is helpful, but we just need a soft nudge at the right moment. The hover effect should validate whether it's possible to jump or set the breakpoint, without being distracting at other times.

### Breakpoints On Hover

In this example, the user is hovering on 127 and can add a breakpoint. The user is also hovering on line 133 with the alt key pressed and can jump to 133.

![Screen Shot 2018-03-04 at 10.59.10 PM.png](https://shipusercontent.com/a4f545be5c6c59867e7169527e473315/Screen%20Shot%202018-03-04%20at%2010.59.10%20PM.png)

### Empty Lines

![Screen Shot 2018-03-04 at 10.40.37 PM.png](https://shipusercontent.com/ac8cfc62314fc2b679ab35e79e3320c4/Screen%20Shot%202018-03-04%20at%2010.40.37%20PM.png)

---

Other options:

- remove the opacity, perhaps the user doesn't need a visual indicator that the line is valid
- remove the opacity, and use a different visual indicator like bold the line number on hover

darkwing:
Maybe a light grey when you hover, so it's not confused with disabled breakpoint?  I love the idea though.

jasonLaster:
yeah - that's a good point. i was worried about disabled bp state too

violasong:
Love this idea! More informative than the empty lines, and lol that's so true re: zebra effect. 

Sounds like the same line could have either a breakpoint added or CtL activated, so there may be a need for a combined indicator?

jasonLaster:
ooh good point. You could jump to a line with a breakpoint. 

I kinda was lazy when i chose the background color for CtL maybe we need breakpoint like icon that we can show?

violasong:
[...]

PR #5979, opened 2018-04-15
[Editor] Add gutter hover effect

Sekhmet:
Fixes Issue: #5603 

### Summary of Changes

* Add hover effect to gutter line numbers.

### Test Plan

- [x] Open JS file.
- [x] Hover on the gutter line numbers.
- [x] Line numbers should be highlighted.
- [x] Empty lines shouldn't be highlighted.
- [x] Existing breakpoints shouldn't be highlighted.
- [x] Highlight should disappear after leaving gutter area.

### Screenshots/Videos

Before:
gutter_before

After:
gutter_after

### DIFF
--- a/src/components/Editor/Editor.css
+++ b/src/components/Editor/Editor.css
@@ -64,7 +64,8 @@ html[dir="rtl"] .editor-mount {
 
 :not(.empty-line):not(.new-breakpoint) > .CodeMirror-gutter-wrapper:hover > .CodeMirror-linenumber {
   color: white;
-  left: 27px !important;
+  left: auto !important;
+  right: 4px;
 }
 
 .editor-wrapper .breakpoints {
--- a/src/components/Editor/Editor.css
+++ b/src/components/Editor/Editor.css
@@ -51,17 +51,27 @@ html[dir="rtl"] .editor-mount {
   direction: ltr;
 }
 
+.theme-light {
+  --gutter-hover-background-color: #dde1e4;
+}
+
+.theme-dark {
+  --gutter-hover-background-color: #414141;
+}
+
 :not(.empty-line):not(.new-breakpoint) > .CodeMirror-gutter-wrapper:hover {
   width: 60px;
   height: 13px;
   left: -55px !important;
-  background-color: var(--theme-highlight-gray) !important;
+  background-color: var(--gutter-hover-background-color) !important;
   mask: url(/images/breakpoint.svg) no-repeat;
   mask-size: 100%;
   mask-position: 0 1px;
 }
 
-:not(.empty-line):not(.new-breakpoint) > .CodeMirror-gutter-wrapper:hover > .CodeMirror-linenumber {
+:not(.empty-line):not(.new-breakpoint)
+  > .CodeMirror-gutter-wrapper:hover
+  > .CodeMirror-linenumber {
   left: auto !important;
   right: 6px;
   color: var(--theme-body-color);
--- a/src/components/Editor/Editor.css
+++ b/src/components/Editor/Editor.css
@@ -54,20 +54,18 @@ html[dir="rtl"] .editor-mount {
 :not(.empty-line):not(.new-breakpoint) > .CodeMirror-gutter-wrapper:hover {
   width: 60px !important;
   height: 13px;
-  left: -57px !important;
-  background-color: var(--theme-highlight-blue) !important;
+  left: -55px !important;
+  background-color: var(--theme-highlight-gray) !important;
   mask: url(/images/breakpoint.svg) no-repeat;
   mask-size: 100%;
   mask-position: 0 1px;
-  opacity: 0.8;
 }
 
 :not(.empty-line):not(.new-breakpoint)
   > .CodeMirror-gutter-wrapper:hover
   > .CodeMirror-linenumber {
-  color: white;
   left: auto !important;
-  right: 4px;
+  right: 6px;
 }
 
 .editor-wrapper .breakpoints {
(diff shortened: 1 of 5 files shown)


jasonLaster:
Nice. Does this work when a file has 100+ or 1000+ lines?

Sekhmet:
I couldn't reproduce this problem (both on Firefox and Chrome on Linux). Is there anything specific that might cause this?
Maybe there is something that changes line number color, if it's the case does that patch solve the problem?
```diff
diff --git a/src/components/Editor/Editor.css b/src/components/Editor/Editor.css
index 12f2ed8..68f924f 100644
--- a/src/components/Editor/Editor.css
+++ b/src/components/Editor/Editor.css
@@ -65,7 +65,7 @@ html[dir="rtl"] .editor-mount {
 :not(.empty-line):not(.new-breakpoint)
   > .CodeMirror-gutter-wrapper:hover
   > .CodeMirror-linenumber {
-  color: white;
+  color: white !important;
   left: auto !important;
   right: 4px;
 }
```


What should transition apply to? Background and text color?

bomsy:
Random Q! should the arrow head pop of the edge of the gutter, and maybe the bottom of the triangle should line up with the edge of the gutter.

Sekhmet:
That's something that Jason pointed out as well. Right now hover mark is styled the same way as regular breakpoint. I can update both of them in this PR. It should look better that way.

jasonLaster:
@Sekhmet can you try styling it so on hover we get the same soft grey hover (w/o a bp) as we get when hovering a button? Curious how that would look

jasonLaster:
I started a transparent hack here 
https://gist.github.com/56463fa5ae15cf72a4ca0b281a40ed55

i couldnt figure out how to show the line number...

Sekhmet:
Grey background:
gutter_grey

Grey background withour icon:
gutter_grey_without_arrow

Sekhmet:
Regular version with opacity removed and icon aligned to the edge of gutter:
gutter_normal_aligned

darkwing:
@jasonLaster If you like grey (I think I like that more than blue), this var we're already using would work:  

```css
.editor.new-breakpoint svg {
  fill: var(--theme-highlight-gray)
}

```

hover

jasonLaster:
yeah, i think the grey might be nice

jasonLaster:
This is looking really good. 

- lets get circle green
- perhaps lets darken the highlighted line number to give some pop

src/components/Editor/EmptyLines.css
```diff
@@ -1,7 +1,7 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
-
+/*
```
darkwing: I believe we still want this, right?

src/components/Editor/Editor.css
```diff
@@ -51,6 +51,23 @@ html[dir="rtl"] .editor-mount {
   direction: ltr;
 }
 
+:not(.empty-line):not(.new-breakpoint) > .CodeMirror-gutter-wrapper:hover {
+  width: 60px !important;
```
darkwing: I'm slightly concerned with all of the `!important` declarations -- can we adjust our selector specificity so we don't need them?

src/components/Editor/EmptyLines.css
jasonLaster: oh... lets rm this file

src/components/Editor/Editor.css
jasonLaster: this might be hard to fix.. but yes, we should try

src/components/Editor/EmptyLines.css
darkwing: Sorry!  I wasn't aware -- remove away! :)

src/components/Editor/Editor.css
jasonLaster: i see. makes sense

src/components/Editor/Editor.css
```diff
@@ -51,6 +51,22 @@ html[dir="rtl"] .editor-mount {
...
+  background-color: var(--theme-highlight-gray) !important;
+  mask: url(/images/breakpoint.svg) no-repeat;
+  mask-size: 100%;
+  mask-position: 0 1px;
+}
+
+:not(.empty-line):not(.new-breakpoint) > .CodeMirror-gutter-wrapper:hover > .CodeMirror-linenumber {
+  left: auto !important;
+  right: 6px;
+  color: var(--theme-body-color);
```
Sekhmet: This looks good on light theme, but on dark theme it's hard to read. Is there a color that might work well for this?

jasonLaster:
hmm, we still want to prevent highlighting empty lines

src/components/Editor/Editor.css
darkwing: I love the detail in testing the dark theme too! Great work!  👍 

How about `--theme-body-color-alt` for the dark theme?  Seem reasonable to me :)

violasong:
It's helpful to see all the variations of this - thanks all! One idea - Could we try using the row-hover background colors from the inspector light and dark themes? (It's a very subtle blue - maybe too subtle.) White-on-gray could be good for the disabled breakpoints.

jasonLaster:
@violasong i went with a solid grey as it's the least intrusive now:

- row-hover had opacity which meant it'd clip
- disable bps is light blue and changing it seems like something we can do once this lands and we know we like it...

violasong:
Ah yes, that makes sense to keep disable bps the existing light blue for now - sounds good!

PR #5980, opened 2018-04-15
[Sources] keep focus on source item if copying source uri

mutdmour:
Fixes Issue: #5552 

Here's the Pull Request Doc

### Summary of Changes

* prevent blur after context menu is shown, enable it after blur is prevented

### Test Plan

- [x] Opening context menu does not remove focus
- [x] Closing context menu by clicking on `Copy Source Uri` item does not remove focus
- [x] Closing context menu by clicking outside of the menu does not remove focus
- [x] Keyboard navigation still works afterward

Feel free to improve it!


mutdmour:
this might be overengineering a simpler problem. My initial approach in #5900 might have been simpler + keeping the focus on the tree I added here.
All would be needed in ManagedTree.js `focusItem` function is
```javascript
focusItem = (item: Item) => {
   const isPopupShown = document.querySelector("#contextmenu-mask.show");
    if (isPopupShown != null && typeof item == "undefined") {
      setTimeout(() => {
        // keep focus on tree to maintain keyboard navigation
        const tree = document.querySelector(".tree." + this.props.className);
        if (tree) {
          tree.focus();
        }
      }, 0);
      return;
    }
    if (!this.props.disabledFocus && this.state.focusedItem !== item) {
    ...
}
```
I will leave it for the reviewers to consider which is better.

darkwing:
Thanks for giving this a shot @mutdmour -- it's a real challenge, much harder than it should be :(

I'm ultra sensitive when it comes to adding complication to the two Tree components; they're probably due for a massive rewrite soon.

I'm really concerned about the `setTimeout` used here; `setTimeout` promises (no pun intended) magic but it's really, really unreliable in the long run.  I also think that `allowBlur` and `preventBlur` can be confusing/conflicting. 

Is there a way we can try to avoid the blur being fired at all, rather than trying to do a hot swap of highlightItems?

mutdmour:
@darkwing this is the simplest fix I have come up with so far. I get the concerns about setTimeout though. 
The blur event seems to be triggered by a shift in focus towards body when the context menu shows. I have not found a fix for that. 
To note, I don't see this problem in Firefox and Developer Firefox on Mac. So this fix might not be necessary.

Issue #3074, opened 2017-06-02
If you’re paused at a breakpoint and try to declare something in the “Console” panel, it doesn’t work.


vladikoff:
If you’re paused at a breakpoint and try to declare something in the “Console” panel, it doesn’t work.

![image](https://cloud.githubusercontent.com/assets/128755/26747119/befc393c-47c1-11e7-99da-3104247500aa.png)

(This is from a series of bugs reported by Firefox teams, I'm making sure we file all of those. Please close if dup)

jimblandy:
Curious: if you change the example to use `var` instead of `let`, does it work?

SpiderMonkey itself does have the ability to inject these declarations into the frame. The Debugger.Frame.prototype.eval function is specified to inject them in non-strict code:

> If <i>code</i> is not strict mode code, then variable declarations in <i>code</i> affect the environment of this frame. (In the terms used by the ECMAScript specification, the `VariableEnvironment` of the execution context for the eval code is the `VariableEnvironment` of the execution context that this frame represents.) If implementation restrictions prevent SpiderMonkey from extending this frame's environment as requested, this call throws an Error exception.

The reason strict mode code is treated differently is that `eval` was meant to behave like an actual `eval` call placed in the code. In strict mode, eval cannot affect its caller's environment. And in fact, a `let` in an eval (shown in the screen shot) never affects its caller's environment, strict mode or not.

However, it may be the case that, for developer tools, the old, non-strict, invasive behavior is what's always wanted, even in strict mode code, and even for `let`. I *think* this can be implemented.

jasonLaster:
Thanks for the overview. Out of curiosity, what do you think it would take?

jimblandy:
Probably changes to [DebuggerGenericEval].

[DebuggerGenericEval]: http://searchfox.org/mozilla-central/rev/d441cb24482c2e5448accaf07379445059937080/js/src/vm/Debugger.cpp#8176-8264

jasonLaster:
reposting in the console component: https://bugzilla.mozilla.org/show_bug.cgi?id=1375300

PR #5983, opened 2018-04-16
(WIP) Fix console assignments

jasonLaster:
Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1375300

### Summary of Changes

One of the biggest issues with debugging today is that it is difficult to declare variables when paused in an instance. For example `var a = 3` does not work.

We discussed our options last week in the spider monkey / debugger collaboration call [here][h]. There are definitely some good server/platform options, but they are likely not ready before 61. This is a client-side short term solution for declaring variables when paused.

#### User goals

1. be able to set a temporary variable and use it immediately
2. be able to set a variable and have it when paused the next time (step, or resume)

#### What does chrome do?

it looks like they add the variable to the window, unless the window already has this property. This mirrors that behavior with a small babel transform

#### New behavior

input `var a = 3`
output `if (!window.hasOwnProperty("a")) window.a = 3;`

input `const a = 3`
output `if (!window.hasOwnProperty("a")) window.a = 3;`

input `let a = 3, b = 4`,
output `if (!window.hasOwnProperty("a")) window.a = 3;if (!window.hasOwnProperty("b")) window.b = 4;` 

### Caveats

- If there's already a local by that name, it won't shadow it, for example.
- allows re-declarations for const/let/var

### GIF

[h]: https://docs.google.com/document/d/1z9Xu3oe90SJ4ow2akK7pnD3zYETDmQFVkTKUK2ax22k/edit#


darkwing:
Nice!  Testing this out right now!

jasonLaster:
@darkwing sorry, should have mentioned this earlier. We should talk with @jimb and @nchevobbe first

nchevobbe:
Would we do the same thing if the variable are declared in a block, e.g. 
```js
{
  let a = 3, b = 4;
  console.log(a, b);
}
```
?
In the example above, I'm not sure we'd want to tie `a` and `b` to the window.

jasonLaster:
we can special case blocks and functions e.g.

```js
function foo() {
  let a = 2;
  console.log(a)
}

foo();
```

jasonLaster:
closing so we can evaluate the scope of server work

jasonLaster:
I'm going to look into making this more safe with the hope of landing a fix before we can resolve the platform issues

jasonLaster:
Alright, so this PR should be good to go:

we only map top-level variable declarators. This protects us from accidentally breaking shadowing

jasonLaster:
So a couple of observations without the patch:

when i'm paused and there is no `this`, I see declarations when paused e.g. https://dbg-obj-properties.glitch.me/

src/workers/parser/mapOriginalExpression.js
```diff
@@ -81,22 +77,67 @@ export default function mapOriginalExpression(
...
+        didUpdate = true;
+        replaceNode(ancestors, t.cloneNode(replacement));
+      }
+    }
+
 
-    const replacement = replacements.get(locationKey(node.loc.start));
-    if (replacement) {
-      replaceNode(ancestors, t.cloneNode(replacement));
+    if (t.isVariableDeclaration(node) && !t.isBlockStatement(parentNode)) {
```
wldcordeiro: 💯 Though I think it could use some comments to clarify for those unfamiliar with the `t.*` syntax

jasonLaster:
closing again. Lets see what we can do from the spidermonkey side

jasonLaster:
This is related to #3074

src/workers/parser/tests/mapOriginalExpression.spec.js
```diff
@@ -6,55 +6,87 @@ import mapOriginalExpression from "../mapOriginalExpression";
...
+    expect(expression.replace("\n", "")).toEqual(
+      'if (!window.hasOwnProperty("a")) window.a = 3;'
+    );
+  });
+
+  it("variable declaration", () => {
+    const expression = mapOriginalExpression("var a = 3, b = 4", {});
+    expect(expression.replace("\n", "")).toEqual(
+      'if (!window.hasOwnProperty("a")) window.a = 3;if ' +
+        '(!window.hasOwnProperty("b")) window.b = 4;'
```
jasonLaster: we should be careful about workers which wont have a window

src/workers/parser/mapOriginalExpression.js
```diff
@@ -81,22 +77,74 @@ export default function mapOriginalExpression(
...
   }
 
-  if (replacements.size === 0) {
-    // Avoid the extra code generation work and also avoid potentially
-    // reformatting the user's code unnecessarily.
-    return expression;
+  return replacements;
+}
+
+// input `var a = 3` => `if (!window.hasOwnProperty("a")) window.a = 3 `
```
loganfsmyth: `self` now, right?

src/workers/parser/mapOriginalExpression.js
```diff
@@ -81,22 +77,74 @@ export default function mapOriginalExpression(
...
+      (t.isIdentifier(node) || t.isThisExpression(node))
+    ) {
+      const replacement = replacements.get(locationKey(node.loc.start));
+      if (replacement) {
+        didUpdate = true;
+        replaceNode(ancestors, t.cloneNode(replacement));
+      }
+    }
+
+    if (t.isVariableDeclaration(node) && !t.isBlockStatement(parentNode)) {
```
loganfsmyth: Assuming the `isBlockStatement` here is meant to check if something is a top-level declaration, this won't quite work since you can also have 
```
{
  var foo = 4;
}
```
which is a top-level declaration of `foo` since the `var` is hoisted, but it'll fail this check. Similarly,
```
function foo(){
  if (true) var foo = 4;
}
```
is function-scoped variable, but here you'll try to globalize it.

src/workers/parser/mapOriginalExpression.js
```diff
@@ -81,22 +77,74 @@ export default function mapOriginalExpression(
...
+      const parts = globalizeAssignment(node);
 
-    const replacement = replacements.get(locationKey(node.loc.start));
-    if (replacement) {
-      replaceNode(ancestors, t.cloneNode(replacement));
+      didUpdate = true;
+      const lastAncestor = ancestors[ancestors.length - 1];
+      const { index } = lastAncestor;
+      parent.node[parent.key].splice(index, 1, ...parts);
```
loganfsmyth: This will throw, since `if (true) var foo = 4` is a case where the parent is a single property, not an array of nodes.

src/workers/parser/mapOriginalExpression.js
loganfsmyth: This approach will also skip globalizing functions and classes, which doesn't seem ideal.

loganfsmyth: I think ideally this logic, if it stays as-is, should use the `buildScopeList` data to decide which things are top-level declarations.

src/workers/parser/mapOriginalExpression.js
```diff
@@ -81,22 +77,74 @@ export default function mapOriginalExpression(
...
+    if (init) {
+      return t.ifStatement(
+        t.unaryExpression(
+          "!",
+          t.callExpression(
+            t.memberExpression(
+              t.identifier("self"),
+              t.identifier("hasOwnProperty")
+            ),
+            [t.stringLiteral(id.name)]
```
loganfsmyth: This is going to fail if users are using destructuring since `id` is a general pattern.

src/workers/parser/mapOriginalExpression.js
```diff
@@ -81,22 +77,74 @@ export default function mapOriginalExpression(
...
-    // reformatting the user's code unnecessarily.
-    return expression;
+  return replacements;
+}
+
+// input `var a = 3` => `if (!window.hasOwnProperty("a")) window.a = 3 `
+function globalizeAssignment(node) {
+  return node.declarations.map(({ id, init }) => {
+    if (init) {
+      return t.ifStatement(
```
loganfsmyth: Two major concerns with this pattern:

1. It seems like this should probably be `var _tmp = 3; if (!window.hasOwnProperty("a")) window.a = _tmp`, otherwise  doing `var foo = someComplexThing();` would silently skip evaluating `someComplexThing()` when `window.foo` exists.
2. This means if you do `var foo = 3;` in the console, and then realize that you want to change it, the easy option is to press UP to get the existing entry and change the value to `4` or whatever, but now since the first attempt created `window.foo`, the second line would skip assigning to `window.foo` entirely.

src/workers/parser/mapOriginalExpression.js
loganfsmyth: Oh and this will also hit `for (let foo = 4; false;){}` since the variable isn't in a block.

src/workers/parser/mapBindings.js
```diff
@@ -0,0 +1,69 @@
...
+function hasDestructuring(node) {
+  return node.declarations.some(declaration => t.isPattern(declaration.id));
+}
+
+function alreadyDefined(node, bindings) {
+  const names = node.declarations.map(dec => dec.id.name);
+  return names.some(name => bindings.includes(name));
+}
+
+export default function mapExpressionBindings(
```
loganfsmyth: Should we skip entirely if a `with` statement is found? I think probably? For example:
```
var obj = { a: 5 };
with(obj) {
  var a = 4;
  a = 6;
}
```
if the `var a = 4` becomes `self.a = 4` then `a = 6` is suddenly modifying the value on the object instead.

src/workers/parser/mapExpression.js
```diff
@@ -0,0 +1,14 @@
+import mapOriginalExpression from "./mapOriginalExpression";
```
loganfsmyth: Missing `// @flow`

src/workers/parser/mapBindings.js
loganfsmyth: If we're entirely bailing out there, it also makes be wonder if we'd be better bailing entirely when destructuring is used. I worry that it might be even more confusing to users if part of their executed snippet behaves one way, and part behaves another way.

src/workers/parser/mapExpression.js
```diff
@@ -0,0 +1,14 @@
...
+import mapExpressionBindings from "./mapBindings";
+
+export default function mapExpression(
+  expression: string,
+  mappings: {
+    [string]: string | null
+  },
+  bindings: string[]
+): string {
+  const safeExpression = mapExpressionBindings(expression, bindings);
```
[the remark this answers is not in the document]
jasonLaster: good idea

src/workers/parser/mapBindings.js
jasonLaster: > should we skip entirely if a `with` is found
yep.

> bailing entirely
I agree. better to be safe than sorry

src/workers/parser/mapBindings.js
```diff
@@ -0,0 +1,69 @@
...
+  expression: string,
+  bindings: string[] = []
+): string {
+  const ast = parseScript(expression);
+
+  t.traverse(ast, (node, ancestors) => {
+    const parent = ancestors[ancestors.length - 1];
+    if (
+      t.isVariableDeclaration(node) &&
+      isTopLevel(ancestors) &&
```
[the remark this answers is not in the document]
jasonLaster: true. i have a test that documents that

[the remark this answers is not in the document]
loganfsmyth: Should we bail out if that's found? It seems fine to not handle all the cases, but since users have no insight into what we've decided to handle and what to ignore, it seems like partial-transformation will be pretty confusing.

src/reducers/pause.js
```diff
@@ -399,6 +399,33 @@ export function getFrameScopes(state: OuterState) {
...
+  if (!frameScope || frameScope.pending) {
+    return;
+  }
+
+  let currentScope = frameScope.scope;
+  let frameBindings = [];
+  while (currentScope && currentScope.type != "object") {
+    const bindings = Object.keys(currentScope.bindings.variables);
+    const args = currentScope.bindings.arguments.map(
+      argument => Object.keys(argument)[0]
```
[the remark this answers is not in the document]
jasonLaster: good call

src/workers/parser/mapExpression.js
```diff
@@ -0,0 +1,14 @@
...
+  const originalExpression = mapOriginalExpression(safeExpression, mappings);
```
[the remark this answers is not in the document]
jasonLaster: yes!

src/workers/parser/mapBindings.js
```diff
@@ -0,0 +1,108 @@
...
+    if (t.isWithStatement(node)) {
+      shouldUpdate = false;
+      return;
+    }
+
+    if (!isTopLevel(ancestors)) {
+      return;
+    }
+
+    if (t.isAssignmentExpression(node) && !t.isMemberExpression(node.left)) {
```
loganfsmyth: Better to check `t.isIdentifier(node.left)` since `node.left` can also be a destructuring pattern. We'll also want to clear `shouldUpdate` in that case.

src/workers/parser/mapBindings.js
```diff
@@ -0,0 +1,108 @@
...
+  });
+}
+
+function globalizeAssignment(node, bindings) {
+  if (bindings.includes(node.left.name)) {
+    return node;
+  }
+
+  const identifier = t.memberExpression(t.identifier("self"), node.left);
+  return t.expressionStatement(
```
loganfsmyth: Since this is replacing an expression, it shouldn't be wrapped in a statement.

src/workers/parser/mapBindings.js
```diff
@@ -0,0 +1,108 @@
...
+    t.assignmentExpression("=", identifier, node.right)
```
loganfsmyth: We should use `node.operator` instead of hard-coding `"="` since it this could also be `foo += 4` or something like that.

src/workers/parser/mapBindings.js
```diff
@@ -0,0 +1,108 @@
...
+  const ancestor = ancestors[ancestors.length - 1];
+
+  if (typeof ancestor.index === "number") {
+    ancestor.node[ancestor.key][ancestor.index] = node;
+  } else {
+    ancestor.node[ancestor.key] = node;
+  }
+}
+
+function replaceParent(ancestors, newNodes) {
```
loganfsmyth: Should we maybe combined this with `replaceNode` as

```
  if (typeof ancestor.index === "number") {
    if (Array.isArray(node)) {
      ancestor.node[ancestor.key].splice(ancestor.index, 1, ...node);
    } else {
      ancestor.node[ancestor.key][ancestor.index] = node;
    }
  } else {
    ancestor.node[ancestor.key] = node;
  }
```

src/workers/parser/mapBindings.js
```diff
@@ -0,0 +1,108 @@
...
+  const parent = ancestors[ancestors.length - 1];
+  const lastAncestor = ancestors[ancestors.length - 1];
```
loganfsmyth: If this does stick around, `lastAncestor` can be removed here since it's the same as `parent`.

src/workers/parser/mapBindings.js
```diff
@@ -0,0 +1,108 @@
...
+
+    if (!t.isVariableDeclaration(node)) {
+      return;
+    }
+
+    if (hasDestructuring(node)) {
+      shouldUpdate = false;
+      return;
+    }
+
```
loganfsmyth: Should we also bail out if `node.name === "self"` or for assignment if someone does `self = 4`?

src/workers/parser/mapBindings.js
jasonLaster: but makes sense

src/test/mochitest/browser_dbg-console-map-bindings.js
```diff
@@ -0,0 +1,45 @@
...
+
+  if (!win) {
+    win = toolbox.win;
+  }
+
+  if (!toolbox.splitConsole) {
+    pressKey(dbg, "Escape");
+  }
+
+  return new Promise(resolve => {
```
loganfsmyth: I know it was copy-pasted, but this antipattern always just jumps out at me.

jasonLaster: hmm, are you suggesting? I think this would be nicer :)

```js
async function getSplitConsole(dbg) {
  // before
  await toolbox.getPanelWhenReady("webconsole");
  // after
  return jsterm
}
```

loganfsmyth: Yep! Or at a minimum
```
return toolbox.getPanelWhenReady("webconsole").then(() => {
  ok(toolbox.splitConsole, "Split console is shown.");
  return toolbox.getPanel("webconsole").hud.jsterm;
});
```
without the `new Promise`.

jasonLaster: oh ha

PR #5991, opened 2018-04-17
Changes the close button to an actual button

BadInternetGuy:
Fixes Issue: #4070 

This changes to div to a button, so that it can work with a screen reader. The padding around the button is still slightly to large, but the button works as needed.

Ctrl+Shift+F to open a search bar.
Press the close button.
Padding makes the button slightly to large. 

clickedbutton

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 [![slack-badge]][slack] ![][ci-status] [![npm-version]][npm-package] [![PRs Welcome]][make-a-pull-request]
 
-debugger.html is a very hackable debugger for post-modern times, built from the ground down using [React] and [Redux].  It is designed to be appropriate, yet powerful.  And it is engineering us to be predictable, understandable, and testable.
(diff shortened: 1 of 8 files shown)


darkwing:
I added two commits:

1.  Update to jest tests (`yarn jest -u`)
2.  Lint fixes (`yarn lint`)

darkwing:
This is looking really good!  A few notes:

- [ ] The button looks wider than it was before, which is easy to see when you hover over it;  `<button>` elements are given padding by default in every browser so I think the fix is:

```css
.close-btn {
  padding: 0;
}
```

I looked at the update in Chrome (we don't officially support using the debugger in Chrome but we do our best) and the "x" somehow doesn't display...until my fix above.  Padding saves the day!

Once you update padding I think this is good to go! :) . Make sure to pull down my changes `git pull origin close-button` before making updates!  Cheers!

Issue #5995, opened 2018-04-17
[testing] babel-preview is intermittent


jasonLaster:
I'm seeing some failures when i run babel-preview. `yarn mochih babel-preview`

The issue i see is that we're hovering and a popup doesn't show up...

PR #5996, opened 2018-04-17
Fix preview int

jasonLaster:
Fixes Issue: #5995 

### Summary of Changes

- we're now clearing via actions as opposed to flaky hovers


darkwing:
This has an extra commit :p

PR #6025, opened 2018-04-19
Fixes outline issue from our earlier fix to the button

BadInternetGuy:
Fixes Issue: #4070 

Finally managed to remove the outline from our button. 
We simply added a new attribute to the CSS file, namely, adding
outline: none !important;

Before:
clickedbutton

After:
fixedpadding


darkwing:
Hello @BadInternetGuy!  This is one of the weirder issues I've seen in a while.  What's causing the problem is `:moz-focusring` (https://developer.mozilla.org/en-US/docs/Web/CSS/:-moz-focusring):

`--theme-focus-outline: 1px dotted var(--theme-focus-outline-color);`

What's weird is why the focusring is stretching the entire height of its neighboring input; if you remove `moz-focusring`, the outline is only the height of the button, which is what we desire.

I'm going to play around with this and try to figure it out.

darkwing:
OK -- updated to use the `:focus` pseudo instead of `!important`

darkwing:
Nice work @BadInternetGuy and team!

Issue #4080, opened 2017-09-18
[Accessibility] accordion accessibility


yzen:
This issue tracks the work to make the accordion accessible. A PR can do one or several items in the checklist:

- [x] Visually (CSS focus styling) there is no way to focus on the accordion headers and reload button
- [x] Need to be able to navigate to headers with keyboard
- [ ] Need to be able to collapse expand headers with keyboard
- [x] Headers/Panels should at least have list/listitem semantics including expanded and collapsed states.
- [ ] Headers listitem should have expanded and collapsed states exposed.
- [ ] When reaching a tree widget inside an accordion section focus should remain on the container for efficiency. The tree should handle keyboard using arrow keys (similar to source list tree widget).
- [ ] Breakpoint related check boxes must be properly accessible. Text label (or code line that serves as one) must be set as the formal label for the checkbox.

### Screenshot 
![Screen Shot 2017-09-18 at 8.12.38 PM.png](https://shipusercontent.com/623f4c6a99b4a6184f5f7815090626d6/Screen%20Shot%202017-09-18%20at%208.12.38%20PM.png)

jasonLaster:
#### Style the focus state:

This is a patch for styling the focus state. I'm not sure what style we should apply. 

```diff
diff --git a/src/components/shared/Accordion.css b/src/components/shared/Accordion.css
index 0cd96de..3035a11 100644
--- a/src/components/shared/Accordion.css
+++ b/src/components/shared/Accordion.css
@@ -30,6 +30,10 @@
   cursor: default;
 }

+.accordion ._header:focus {
+  background-color: yellow; /* not sure what color makes sense */
+}
+
 .accordion ._header:hover {
   background-color: var(--theme-toolbar-background-hover);
 }
```


#### Need to be able to navigate to headers with keyboard

Do we need to do something to make the header focusable?

#### Need to be able to collapse expand headers with keyboard

```diff
diff --git a/src/components/shared/Accordion.js b/src/components/shared/Accordion.js
index 9f45ce2..fb69473 100644
--- a/src/components/shared/Accordion.js
+++ b/src/components/shared/Accordion.js
@@ -35,6 +35,7 @@ class Accordion extends Component {

     const self: any = this;
     self.renderContainer = this.renderContainer.bind(this);
+    self.onKeyDown = this.onKeyDown.bind(this);
   }

   componentWillReceiveProps(nextProps: Props) {
@@ -93,9 +94,15 @@ class Accordion extends Component {
     );
[...]

PR #5949, opened 2018-04-12
Converted accordian to list

jawyoonis:
Fixes Issue: #4080 checkbox #4 -Headers/Panels should at least have list/listitem semantics including expanded and collapsed states.

Here's the Pull Request Doc
https://devtools-html.github.io/debugger.html/CONTRIBUTING.html#pull-requests

### Summary of Changes
We removed the divs and changes to lists and we also used aria roles to use list/lisitems for headers/panels.

* change 1- Changed divs to list 
* change 2-Added aria roles to list/listitems semantics 

### Test Plan
We used the debugger tool's console to visualize whether our changes are effective or not. We seen all the changes that we made on the project are effective and working with the project.

image

PR #6026, opened 2018-04-19
Accordion accessibility enhancement

jawyoonis:
Fixes Issue: #4080 checkbox #4 -Headers/Panels should at least have list/listitem semantics including expanded and collapsed states.

### Summary of Changes
We changed the header panels from divs to lists and we also added aria roles to use list/lisitems for headers/panels.

* change 1- Changed divs to list 
* change 2-Added aria roles to list/listitems semantics
* change 3-addressed requests from @darkwing in [PR#5949 ](https://github.com/devtools-html/debugger.html/pull/5949)

### Test Plan
You can use debugger tool's console to verify that header/panels are now lists and have aria roles.


darkwing:
Awesome work here @jawyoonis !  The only update I made was a padding update for RTL languages.  If tests pass we'll get this in.  AWESOME!

Issue #5748, opened 2018-03-21
[Editor] switching panes causes the editor to lose its scroll position


jasonLaster:
STR:

1. go to todomvc
2. go to todo-view
3. scroll to line 70
4. switch to console
5. switch back

ER: you're on line 70
AR: you're on line 1

---

This was caused by the new VisibilityHelper

PR #6030, opened 2018-04-19
Revert "Added VisibilityHandler to handle the mounting and unmounting…

jasonLaster:
Fixes Issue: #5748

### Summary of Changes

* We no longer unmount the app when we navigate to another tool
* This means we will naturally preserve the scroll position

Trade-offs
- I do not believe we saw a significant performance win by unmounting the application.
- Unmounting the app means removing codemirror's underlying dom
- saving the scroll position on scroll is possible, but requires scrolling at the right time. There's also quite a bit of overhead in re-mounting...


darkwing:
Tested in panel and scrolling persists.

juliandescottes:
Great that this managed to be fixed before the end of 61 Nightly!

PR #6035, opened 2018-04-20
42 fixes

jasonLaster:


jasonLaster:
we should rebase this commit so that both commits are preserved

PR #6019, opened 2018-04-18
Speed up initial load

jasonLaster:
### Summary of Changes

* The initial react App Mount used to take 220ms now it takes 55ms
* The old App Mount used to build the Source Tree, instantiate CodeMirror, and calculate hidden tabs
* This work is now postponed...

### Old
screen shot 2018-04-18 at 4 58 28 pm

### New 
screen shot 2018-04-18 at 4 48 41 pm

### New Profile
we can toggle devtools really fast now!

screen shot 2018-04-18 at 4 56 00 pm

PR #6041, opened 2018-04-20
Defer updating hidden tabs

jasonLaster:
### Summary of Changes

- Follows up on #6019 
- defers updating hidden tabs which prevents doing expensive work on mount

### DIFF
--- a/src/components/Editor/Tabs.js
+++ b/src/components/Editor/Tabs.js
@@ -77,7 +77,7 @@ class Tabs extends PureComponent<Props, State> {
   }
 
   componentDidMount() {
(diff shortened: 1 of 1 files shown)


src/components/Editor/Tabs.js
```diff
@@ -77,7 +77,7 @@ class Tabs extends PureComponent<Props, State> {
   }
 
   componentDidMount() {
-    this.updateHiddenTabs();
+    window.requestIdleCallback(this.updateHiddenTabs);
```
bomsy: i like!

PR #6043, opened 2018-04-20
[Sources] Defer creating the source tree

jasonLaster:
### Summary

- Follows up on #6019
- Prevent creating the source tree until we have some sources


darkwing:
At first check the tree never seems to load:

doesntload

darkwing:
After rebasing and testing again, sources do now load initially but switching to the outline tab and back to the sources tab clears the Sources tree (nothing displays), presumably because sources haven't changed.

PR #6052, opened 2018-04-21
fix "set directory root" failure

yalooong:
### Summary of Changes

This pull request is a fix to the two issues illustrated below.
Both of the issues related to "Set directory root" on the left panel of the debugger.

Issue 1 - STR:
1. Go to https://firefox-debugger-example-react-js.glitch.me/
2. On the left panel, right click on "Webpack" folder and Click on "Set directory root"
3. Then right click on "app" foler and Click on "Set directory root" ( notice the content is missing ).
kennugbuqh

Issue 2 - STR:
1. Go to https://davidwalsh.name/
2. On the left panel, right click on "davidwalsh.name" and Click on "Set directory root"
3. Then expand the subfolders; right click on "libs" and Click on "Set directory root" ( notice the content is missing ).
v5bvqw7akf

### DIFF
--- a/src/actions/ui.js
+++ b/src/actions/ui.js
@@ -171,9 +171,12 @@ export function setProjectDirectoryRoot(newRoot: string) {
   return ({ dispatch, getState }: ThunkArgs) => {
     const curRoot = getProjectDirectoryRoot(getState());
     if (newRoot && curRoot) {
-      const temp = newRoot.split("/");
-      temp.splice(0, 2);
-      newRoot = `${curRoot}/${temp.join("/")}`;
+      const temp1 = newRoot.replace(/\/+/g, "/").split("/");
+      const temp2 = curRoot.replace(/\/+/g, "/").split("/");
+      if (temp1[0] !== temp2[0]) {
+        temp1.splice(0, 2);
+        newRoot = `${curRoot}/${temp1.join("/")}`;
+      }
     }
 
     dispatch({


jasonLaster:
Oh nice, looks like there are some ci issues

this might be a good opportunity to improve our test coverage too
https://github.com/devtools-html/debugger.html/blob/master/src/actions/tests/ui.spec.js#L96-L97

yalooong:
I added test coverage for the above issues.

darkwing:
I'm still seeing an "off by 1" error with which Tree item is being selected -- like, the item that should be the root is actually first in the tree, and its parent is above it:

sourcetreeoffbyone

yalooong:
@darkwing I updated the code to fix the "off by 1" issue. Please check it out.

src/components/PrimaryPanes/SourcesTree.js
```diff
@@ -350,8 +350,8 @@ class SourcesTree extends Component<Props, State> {
     if (isCustomRoot) {
       let rootLabel = projectRoot.split("/").pop();
       if (sourceTree.contents[0]) {
```
jbhoosreddy: can we put `sourceTree.contents[0]` into a const?

src/actions/ui.js
```diff
@@ -171,9 +171,15 @@ export function setProjectDirectoryRoot(newRoot: string) {
   return ({ dispatch, getState }: ThunkArgs) => {
     const curRoot = getProjectDirectoryRoot(getState());
     if (newRoot && curRoot) {
-      const temp = newRoot.split("/");
-      temp.splice(0, 2);
-      newRoot = `${curRoot}/${temp.join("/")}`;
+      const temp1 = newRoot.replace(/\/+/g, "/").split("/");
```
jbhoosreddy: better naming? temp1, temp2?

yalooong:
@jbhoosreddy updated the naming issues

SourcesTree.js:
screen shot 2018-04-23 at 23 55 56
ui.js:
screen shot 2018-04-23 at 23 55 29

src/actions/ui.js
```diff
@@ -171,14 +171,14 @@ export function setProjectDirectoryRoot(newRoot: string) {
   return ({ dispatch, getState }: ThunkArgs) => {
     const curRoot = getProjectDirectoryRoot(getState());
     if (newRoot && curRoot) {
-      const temp1 = newRoot.replace(/\/+/g, "/").split("/");
-      const temp2 = curRoot
+      const tmpNew = newRoot.replace(/\/+/g, "/").split("/");
```
jbhoosreddy: nit: I was thinking more like newRootArr, or something. I generally try to avoid variable names like tmp. There's always better names which better communicate intent.

jbhoosreddy: but since it was unfortunately already called temp1, I guess I'm fine with it.

yalooong: I can quickly make the tweaks.

Issue #5406, opened 2018-02-13
[Editor] Preview tooltip can get stuck in visible state when mouse leaves code window too fast


digitarald:
STR:

- Pause on a larger code base
- Quickly hover a variable and move mouse out of code panel before tooltip appears (made easy by it being slow on large code cases, see #3984)

AR: Tooltip stays open until mouse re-enters code panel

ER: Tooltip never appears or closes when mouse leaves

Maybe the fix is improving performance; but we'll see Telemetry if we hit a level that would mitigate the issue described here.

jasonLaster:
How is it going @haoRchen?

haoRchen:
@jasonLaster progressing. I have found the mouseover() call within the editor canvas
![10 - possible mask](https://user-images.githubusercontent.com/19397242/37311862-5325c092-261f-11e8-88cd-f5a0b292c3ee.png)
as well as the mouseleave() within the popover
![11 - onmouseleave popover](https://user-images.githubusercontent.com/19397242/37311875-5feb88ca-261f-11e8-8ec0-f13e88736ebe.png)
Manually testing mouseover() is a bit troublesome as it rarely registers the debug variable for creating the popover, the event gets triggered when I move my cursor onto the editor; which will process another element as the target of the mouseover(). eg. the scroll bar UI as shown below
![12 - debugging mouseover](https://user-images.githubusercontent.com/19397242/37312565-e172fc14-2621-11e8-935c-def996d533fe.png)

Will be looking into mouseleave() shortly.

Any advice?

jasonLaster:
Hmm, sometimes when this is the case, I cheat and add a new condition that is only true when i type a letter like "f" :) you can do this with a new keyDown event.

I'll defer to @darkwing, or @wldcordeiro who are great w/ the UI

jasonLaster:
how's this going @haoRchen?

haoRchen:
@jasonLaster  I'm having trouble getting past the editor mask. 
Using your suggested method, pressing "A" on top of the variable grabs nothing as the target:
![17 - target for onkeydown](https://user-images.githubusercontent.com/19397242/37543507-3e17466e-2938-11e8-8b5e-26c941f87754.png)
I am going to see if this issue was introduced in an earlier release via git bisect.

jasonLaster:
Thanks @haoRchen nice work so far. I pinged the core team to see if someone could help mentor

haoRchen:
[...]

PR #6053, opened 2018-04-22
[Preview] avoid windows getting stuck

haoRchen:
Fixes Issue: #5406 

### Summary of Changes

The preview window no longer gets stuck when quickly hovering over a given debug expression.

### DIFF
--- a/src/actions/preview.js
+++ b/src/actions/preview.js
@@ -49,6 +49,12 @@ function isInvalidTarget(target: HTMLElement) {
       !target.parentElement.closest(".CodeMirror-line")) ||
     cursorPos.top == 0;
 
+  console.log(
+    target,
+    tokenText,
+    { invalidTarget, invalidToken, invalidType },
+    target.outerHTML
+  );
   return invalidTarget || invalidToken || invalidType;
 }
 
@@ -71,11 +77,14 @@ export function updatePreview(target: HTMLElement, editor: any) {
 
       // We are mousing over a new token that is not in the preview
       if (!target.classList.contains("debug-expression")) {
+        console.log(`mousing over a new token`, target.innerHTML);
         dispatch(clearPreview());
       }
     }
 
     if (isInvalidTarget(target)) {
+      console.log(`mousing over an invalid targt`, target.outerHTML);
+      dispatch(clearPreview());
       return;
     }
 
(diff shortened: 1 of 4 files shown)


jasonLaster:
love how small the patch is!

jasonLaster:
do we run the risk of also previewing on mouse out... probably not.

And i wonder why the yellow hover is still there. Did you look into how that's added and removed?

haoRchen:
I have tested for previewing onMouseOut, shouldn't occur. I'll look further into the yellow highlight

jasonLaster:
does this help?
Perhaps look into the react code?
https://reactjs.org/docs/react-component.html#componentwillunmount

src/actions/preview.js
```diff
@@ -103,7 +103,7 @@ export function updatePreview(target: HTMLElement, editor: any) {
     if (preview) {
       // Return early if we are currently showing another preview or
```
jbhoosreddy: update comment?

src/actions/preview.js
```diff
@@ -103,7 +103,7 @@ export function updatePreview(target: HTMLElement, editor: any) {
...
       // if we are mousing over the same token as before
-      if (preview.updating || isEqual(preview.tokenPos, tokenPos)) {
+      if (isEqual(preview.tokenPos, tokenPos)) {
```
jbhoosreddy: `updatePreview` was never added to the unit tests.
https://github.com/devtools-html/debugger.html/blob/master/src/actions/tests/preview.spec.js

Could it be possible to add this as a test so to prevent future regressions?

I'd be happy to mentor with that.

src/actions/preview.js
jbhoosreddy: update comment.

jasonLaster:
Haven't tested, but code looks!

jasonLaster:
still seeing some funny flickering

jasonLaster:
I've had this bug on my mind for the past couple months so i sat down and added some additional fixes. Apologies that this is a different branch, there is definitely a lot of overlap:
https://github.com/devtools-html/debugger.html/compare/master...jasonLaster:fix-popup?expand=1


1. target element onMouseLeave
2. document.body onMouseLeave
3. editor onMouseLeave
3. removed the debounce (it is scary)
4. added some additional cases to the popover, those will need to move to the onClose handler to be generalized

lastly, i have a suspicion that there is a MUCH better way to do all of this :) 
I kinda think we should land a bunch of small fixes in this PR and then let it sit a bit. My guess is that the better fix will come to one of us soon

jasonLaster:
I think a bunch of my checks were likely redundant and not needed. the debounce as likely the biggest source of evil

haoRchen:
Hovering over a variable no longer generate a preview, and the highlight is still stuck

jasonLaster:
@haoRchen @darkwing could you manually test this?

darkwing:
Works awesome!  Great work all! 🥇

Issue #6048, opened 2018-04-21
[Sources] Set directory root uses the parent


jasonLaster:
1. go to https://firefox-debugger-example-react-js.glitch.me/
2. set directory root for App

Expected: App is the root
Actual: webpack is the root

![](http://g.recordit.co/1XX8nqjIaL.gif)

PR #6054, opened 2018-04-22
Fixed issue #6048

pranoy10:
Fixes Issue: #6048 

### Summary of Changes

* Fixed a bug in the file ui.js preventing from adding the correct folder as root.

### DIFF
--- a/src/actions/ui.js
+++ b/src/actions/ui.js
@@ -170,7 +170,7 @@ export function clearProjectDirectoryRoot() {
 export function setProjectDirectoryRoot(newRoot: string) {
   return ({ dispatch, getState }: ThunkArgs) => {
     const curRoot = getProjectDirectoryRoot(getState());
-    if (newRoot && curRoot) {
+    if (newRoot || curRoot) {
       const temp = newRoot.split("/");
       temp.splice(0, 2);
       newRoot = `${curRoot}/${temp.join("/")}`;


jasonLaster:
nice, looks like a good fix. It might be nice to add a jest test too
https://github.com/devtools-html/debugger.html/blob/master/src/actions/tests/ui.spec.js#L96-L97

jasonLaster:
small thing, but can you update the PR title?

jasonLaster:
Ping @pranoy10 I'm seeing some issues with the other Jest tests

src/actions/tests/ui.spec.js
```diff
@@ -118,4 +118,11 @@ describe("setProjectDirectoryRoot", () => {
 
     expect(firstSource.relativeUrl).toEqual("scopes.js");
   });
+
+  it("should set the selected directory as the root directory ", () => {
+    const { dispatch, getState } = createStore();
+    dispatch(actions.setProjectDirectoryRoot("webpack://"));
```
jbhoosreddy: `dispatches` are async. do we want to await on them?

src/actions/ui.js
```diff
@@ -170,7 +170,7 @@ export function clearProjectDirectoryRoot() {
 export function setProjectDirectoryRoot(newRoot: string) {
   return ({ dispatch, getState }: ThunkArgs) => {
     const curRoot = getProjectDirectoryRoot(getState());
-    if (newRoot && curRoot) {
+    if (newRoot || curRoot) {
```
jbhoosreddy: `|| curRoot` doesn't really do anything here.
1. What is it's intended purpose? Why are we not using it? Is it not needed?

jbhoosreddy:
Notice that sources disappear.
Please provide gifs which fixes directory root correctly. Thanks!

Issue #6032, opened 2018-04-20
[Breakpoints] Shows the generated source text


jasonLaster:
![Screen Shot 2018-04-20 at 9.33.48 AM.png](https://shipusercontent.com/d6f3e21b36b9b5b894effa6f7a146bd5/Screen%20Shot%202018-04-20%20at%209.33.48%20AM.png)

PR #6055, opened 2018-04-22
[Breakpoints] Show original breakpoint text

jasonLaster:
Fixes Issue: #6032

### Summary of Changes

Previously we were only showing the generated line. Now we show the correct text based on the context


bomsy:
OOOOoH nice!

bomsy:
Found 2 issues

- When remove and re-add the breakpoints in the original file, the breakpoint text no longer switches correctly (see gif below). Is this a known issue?

 

- Minor one, the breakpoint text looses the code styling on context switch.

I can add issues and we can fix separately if you like.

darkwing:
Added two commits:

1 - Ensure the syntax highlighting works upon update
2 - Make it more efficient

PR #6057, opened 2018-04-22
[Sources] increase test coverage for source tree

mutdmour:
Fixes Issue: #5100

Here's the Pull Request Doc

### Summary of Changes

* increase coverage for source tree
`  SourcesTree.js                              |    98.11 |    98.72 |       92 |      100 |               124 |`
* could not get that last line 💩


jasonLaster:
Thanks @mutdmour. Looks great. I think the SourceTree is our most complicated component as well

CC @wldcordeiro @codehag curious to see what you think

src/components/tests/SourcesTree.spec.js
```diff
@@ -85,4 +86,622 @@ describe("SourcesTree", () => {
...
+        const componentWithRoot = render({
+          projectRoot: "project-root"
+        }).component;
+
+        const node = componentWithRoot
+          .instance()
+          .renderItem(itemOne, 0, false, null, false, {
+            setExpanded: jest.fn()
+          });
+        expect(node).toMatchSnapshot();
```
jasonLaster: it might be nice to replace some of the match snapshots with element tests like: exists or has classes:
https://github.com/airbnb/enzyme/blob/master/docs/api/selector.md

src/components/tests/SourcesTree.spec.js
```diff
@@ -85,4 +86,622 @@ describe("SourcesTree", () => {
...
+      const item = createMockItem("root", "http://mdn.com", []);
+      const node = component
+        .instance()
+        .renderItem(item, 0, true, null, false, { setExpanded: jest.fn() });
+      expect(node).toMatchSnapshot();
+    });
+
+    it(
+      "should not show domain item if projectRoot is not empty " +
+        "even if depth is 0",
```
jasonLaster: `should not show domain item when the projectRoot exists`

src/components/tests/SourcesTree.spec.js
```diff
@@ -85,4 +86,622 @@ describe("SourcesTree", () => {
...
+  describe("on receiving new props", () => {
+    let component;
+    let props;
+    let currState;
+
+    beforeEach(() => {
+      const rendered = render({});
+      component = rendered.component;
+      props = rendered.props;
+      currState = component.state();
```
codehag: `defaultState` would be more accurate

src/components/tests/SourcesTree.spec.js
```diff
@@ -85,4 +86,622 @@ describe("SourcesTree", () => {
...
+    };
+
+    const rootDirectory = {
+      contents: new Map(),
+      name: "root",
+      path: "root/"
+    };
+
+    beforeEach(() => {
+      const rendered = render({
```
codehag: a better pattern for this would be to create an overrides object for this, rather than doing this in the before each. Component and props are returned by `render` by default. having this colocated in a test makes it easier to read than having it in a beforeEach

src/components/tests/SourcesTree.spec.js
```diff
@@ -85,4 +86,622 @@ describe("SourcesTree", () => {
...
+      const menuOptions = [
+        {
+          accesskey: "r",
+          click: expect.any(Function),
+          disabled: false,
+          id: "node-set-directory-root",
+          label: "Set directory root"
+        }
+      ];
+      await component.instance().onContextMenu(event, directory);
```
codehag: we use instance quite a bit, maybe it should be returned from the render function?

src/components/tests/SourcesTree.spec.js
```diff
@@ -85,4 +86,622 @@ describe("SourcesTree", () => {
...
+
+      showMenu.mock.calls[0][1][0].click();
+      expect(props.setProjectDirectoryRoot).not.toHaveBeenCalled();
+      expect(props.clearProjectDirectoryRoot).toHaveBeenCalled();
+      expect(copyToTheClipboard).not.toHaveBeenCalled();
+    });
+  });
+
+  describe("renderItem", () => {
+    const { component, props } = render();
```
codehag: this is a bit dangerous. in future modifications someone might miss that this isn't in a before each.

I would however always colocate this kind of setup in the test itself. it makes it more difficult to make a mistake, and mistakes will be made.

src/components/tests/SourcesTree.spec.js
```diff
@@ -85,4 +86,622 @@ describe("SourcesTree", () => {
...
+      const node = component
+        .instance()
+        .renderItem(itemOne, 1, false, null, false, { setExpanded: jest.fn() });
+      expect(node).toMatchSnapshot();
+    });
+
+    it("should show source item with source icon with focus", async () => {
+      const node = component
+        .instance()
+        .renderItem(itemOne, 1, true, null, false, { setExpanded: jest.fn() });
```
codehag: it is unclear to me what each of these args means, can you think of a way to make it clearer to a reader?

src/components/tests/SourcesTree.spec.js
```diff
@@ -85,4 +86,622 @@ describe("SourcesTree", () => {
...
+          .renderItem(itemOne, 0, false, null, false, {
+            setExpanded: jest.fn()
+          });
+        expect(node).toMatchSnapshot();
+      }
+    );
+
+    it("should show menu on contextmenu event", async () => {
+      const event = { event: "contextmenu" };
+      const node = shallow(
```
codehag: shallow is not used in the above tests, should it be?

src/components/tests/SourcesTree.spec.js
```diff
@@ -85,4 +86,622 @@ describe("SourcesTree", () => {
...
+        "http://mdn.com/blackboxed.js",
+        "blackboxed.js",
+        I.Map({ id: "server1.conn13.child1/59" })
+      );
+      const path = component.instance().getPath(item);
+      expect(path).toEqual("http://mdn.com/blackboxed.js/blackboxed.js/update");
+    });
+  });
+
+  describe("test redux connect", () => {
```
codehag: you shouldnt need to test redux (we trust redux to work and have its own tests, it is a dependancy). We should be able to assert that this works correctly from how the component works in the other tests.

This level of "in-component-testing" is very brittle, and makes the tests hard to maintain.

src/components/tests/SourcesTree.spec.js
```diff
@@ -6,54 +6,55 @@ import React from "react";
...
+const sourceThree = I.Map({
+  "server1.conn13.child1/41": createMockSource(
+    "server1.conn13.child1/41",
+    "http://mdn.com/three.js"
+  )
+});
 
-  return { component, props };
-}
+const itemOne = createMockItem(
```
codehag: what does itemOne mean?

src/components/tests/SourcesTree.spec.js
```diff
@@ -85,4 +86,622 @@ describe("SourcesTree", () => {
...
+    });
+  });
+
+  describe("focusItem", () => {
+    it("update the focused item", async () => {
+      const { component } = render({});
+
+      expect(component.state("focusedItem")).toEqual(null);
+
+      await component.instance().focusItem(itemOne);
```
codehag: this is reaching into the internals of the component, rather than testing from the outside. a better way would be to use an event to trigger this.

Testing component internals in a unit tests makes them difficult to maintain, as these internals might change. input -> output tests are more effective and maintainable

src/components/tests/SourcesTree.spec.js
mutdmour: this is targeting the `onContextMenu` of the item, so had to use `shallow` to simulate an event. 

Before when testing for `onContextItem`, we were testing the one of the tree. I could make that clearer in the description of the tests.

src/components/tests/SourcesTree.spec.js
mutdmour: this was more to get coverage on `mapToProps`. I don't really need to test each function, just that mapToProps is being called. Or export `mapToProps` and mock it and test it's been called.

src/components/tests/SourcesTree.spec.js
mutdmour: item from source `one.js` 😆

src/components/tests/SourcesTree.spec.js
mutdmour: well, it's only used in `onKeyDown` so can look at the output of that

src/components/tests/SourcesTree.spec.js
mutdmour: not sure it's better to do this. I think it will make it more messy now that I have co-located the different components and would need a different instance for each component.

src/components/tests/SourcesTree.spec.js
```diff
@@ -6,32 +6,617 @@ import React from "react";
...
+      expect(path).toEqual("http://mdn.com/one.js/one.js/");
+    });
+
+    it("should return path for blackboxedboxed item", async () => {
+      const path = blackboxedComponent.instance().getPath(blackboxedMockItem);
+      expect(path).toEqual("http://mdn.com/blackboxed.js/blackboxed.js/update");
+    });
+  });
+
+  describe("test redux connect", () => {
```
jasonLaster: Yeah, we don't need to get the coverage here, it is okay not to have 100% coverage :)

src/components/tests/SourcesTree.spec.js
```diff
@@ -6,32 +6,617 @@ import React from "react";
...
+  name: "folder",
+  path: "folder/"
+};
+
+const mockEvent = {
+  preventDefault: jest.fn(),
+  stopPropagation: jest.fn()
+};
+
+let defaultComponent;
```
codehag: it looks like there is a misunderstanding -- colocate would be to define the variables in the test. for example
```javascript
it('asserts condition 1', () => {
  const { component, props } = render()
  const newProps = { ...props, ...modifiedProps };
  // ... assertion
});

it('asserts condition  2', () => {
  const { component, props } = render()
  // ... assertion
});
```
This current change is making everything global and is quite dangerous. One can easily miss a step and be using the wrong version of a variable that was defined in another test. the test will pass, however there is a possibility of testing the wrong thing. Generally tests should be as isolated from one another as possible. running them in random order should not have any failures.

It is generally better to have something defined close to where you use it. Can you change this either back to what you had or to what I am suggesting?

src/components/tests/SourcesTree.spec.js
```diff
@@ -41,48 +626,22 @@ function render(overrides = {}) {
   return { component, props };
```
codehag: i mentioned in an earlier review that since instance is being used so often, you can return it from the render function. 

you can do that here:

```
function render(overrides = {}) {
  const props = generateDefaults(overrides);
  const component = shallow(<SourcesTree.WrappedComponent {...props} />);
  const instance = component.instance;
  return { component, props, instance };
}
```

src/components/tests/SourcesTree.spec.js
```diff
@@ -6,32 +6,617 @@ import React from "react";
...
+
+    rendered = render({
+      projectRoot: "mdn"
+    });
+    componentWithMdnRoot = rendered.component;
+    propsWithMdnRoot = rendered.props;
+
+    blackboxedComponent = render({
+      sources: blackboxedMockSource
+    }).component;
```
codehag: a lot of the setup being done here is not necessary for every test. the only shared element is `defaultComponent`. Also, the previous naming was a bit better. I think this is due to the misunderstanding of what I meant when i said colocate -- generally it is better to keep a reader from going up and down the code. It is hard to keep your mental model clear. Unless this setup is really needed for all the tests, then we are doing unnecessary work here computationally, and we are doing unnecessary work here mentally. 

so, once again -- try to keep variables in the test that you use them in. do not define extra stuff that is used in only one test in the before each. Given that we have the render function which will define everything that we need in a single like, it is unlikely that you need this before each at all. If this is confusing, please let me know, I will try to clarify. (also on slack you can find me and ask -- look for yulia)

src/components/tests/SourcesTree.spec.js
```diff
@@ -6,32 +6,617 @@ import React from "react";
...
+};
+
+const mockEvent = {
+  preventDefault: jest.fn(),
+  stopPropagation: jest.fn()
+};
+
+let defaultComponent;
+let defaultProps;
+let defaultState;
```
codehag: I did recommend changing `currState` to `defaultState`. component didn't need to be renamed to `defaultComponent` because it is mutated and you continue to use the same object. calling it default component makes it more confusing, as you expect it to change.

src/components/tests/SourcesTree.spec.js
```diff
@@ -6,32 +6,617 @@ import React from "react";
...
+
+    emptyComponent = render({
+      projectRoot: "custom",
+      sources: I.Map()
+    }).component;
+
+    rendered = render({
+      projectRoot: "mdn"
+    });
+    componentWithMdnRoot = rendered.component;
```
codehag: regarding lines 100-114. You do not want to reassign variables in general. In this case it is very confusing, because I have to really pay attention to which `rendered` object has which effect. it makes the code really hard to follow. One trick to avoid doing this kind of thing is -- imagine the next person who reads this code is drunk -- would they be able to tell you what it is doing? 

again, for this kind of setup -- put the variables in the test, not in the before each or in globals. Avoid globals unless they do not change.  

change means: reassignment or state change related to what you are testing. Something like the mocks you are using is fine. However: creating a lot of test state related variables in the globals scope will lead to problems.

src/components/tests/SourcesTree.spec.js
```diff
@@ -6,32 +6,617 @@ import React from "react";
...
+      expect(propsWithRoot.setProjectDirectoryRoot).not.toHaveBeenCalled();
+      expect(propsWithRoot.clearProjectDirectoryRoot).toHaveBeenCalled();
+      expect(copyToTheClipboard).not.toHaveBeenCalled();
+    });
+  });
+
+  describe("renderItem", () => {
+    it("should show icon for webpack item", async () => {
+      const item = createMockItem("webpack://", "webpack://");
+      const node = renderItem(defaultComponent, item);
```
codehag: I will use this as an example to help make clear what i mean.

We have a test, where we are using a variable called `defaultComponent`. in order for me, an outsider, to know what this test is doing (for example, "why is this failing?", imagine being in a frustrated state of mind), we have to scroll up to line 83. There we see that `defaultComponent` can be modified, because it is an undefined `let`, meaning that it isn't the same thing all the time! Oh no!

next, we have to go through every time `defaultComponent` is defined. it is defined at line 96. then, in every test it goes through some kind of modification. So as the person reading this code, i have to remember -- ok it is in a before each, so it gets reset, are there other beforeEachs? is there some case where this is not reset? is there more nesting that i need to keep in mind? how much of the scope do i need to know about? are there other ways this could get modified?

As you can see, the mental load is _high_. I need to remember a lot, and i need to read a lot of code to know what is happening. 

Ok, now let's define what we are working with, in the test:

```javascript
 describe("renderItem", () => {
    it("should show icon for webpack item", async () => {
      const { component } = render();
      const item = createMockItem("webpack://", "webpack://");
      const node = renderItem(component, item);
      expect(node).toMatchSnapshot();
    });
```

i don't have to go anywhere, and this only costs a single extra line! `component` can mean **only one thing** here. this is a massive improvement for readability. 

DRY code is only good if it makes sense, and I would first write code out WET before drying it up. do not dry up code if it makes it harder to read, and try to keep an outsiders (often emotional and tired) perspective in mind! you always have an audience for your code :)

src/components/tests/SourcesTree.spec.js
```diff
@@ -6,32 +6,617 @@ import React from "react";
...
+    defaultProps = rendered.props;
+    defaultState = defaultComponent.state();
+
+    rendered = render({
+      projectRoot: "root/"
+    });
+    componentWithRoot = rendered.component;
+    propsWithRoot = rendered.props;
+
+    emptyComponent = render({
```
codehag: this was in the test that was using it before, can you move it back?

src/components/tests/SourcesTree.spec.js
```diff
@@ -6,32 +6,617 @@ import React from "react";
...
+      sources: I.Map()
+    }).component;
+
+    rendered = render({
+      projectRoot: "mdn"
+    });
+    componentWithMdnRoot = rendered.component;
+    propsWithMdnRoot = rendered.props;
+
+    blackboxedComponent = render({
```
codehag: this was in the test before, and should be moved back.

src/components/tests/SourcesTree.spec.js
mutdmour: oh ya, I just assumed co-locate meant put everything in the same place. 

Oh so you want the render function to handle the different cases.

mutdmour:
@codehag I think I have addressed all the comments. Let me know if there's anything else 😃

mutdmour:
oh it's failing. will fix it later

mutdmour:
looks like it's failing because of the bug fixed here #6085

Issue #6012, opened 2018-04-18
[Tabs] Update the show more button 


jasonLaster:
It would be nice to use the new chevron for the show more button in the tabbar

Here is how you can get it w/ CSS

```diff
diff --git a/src/components/SecondaryPanes/CommandBar.css b/src/components/SecondaryPanes/CommandBar.css
index c99c97d..32eca25 100644
--- a/src/components/SecondaryPanes/CommandBar.css
+++ b/src/components/SecondaryPanes/CommandBar.css
@@ -48,7 +48,7 @@ img.shortcuts {
 }

 .command-bar img.stepOut {
-  mask: url(/images/stepOut.svg) no-repeat;
+  mask: url(/mc/command-chevron.svg) no-repeat;
 }
```

### Tabbar
<img src='https://shipusercontent.com/4a8efb5e6fc3444e8d2536d1fef402aa/Screen%20Shot%202018-04-18%20at%2010.16.25%20AM.png' title='Screen Shot 2018-04-18 at 10.16.25 AM.png' width=384>

### Show More dropdown
<img src='https://shipusercontent.com/8553368a44822016c484efaba03f7c0d/Screen%20Shot%202018-04-17%20at%2010.02.03%20PM.png' title='Screen Shot 2018-04-17 at 10.02.03 PM.png' width=277>

bomsy:
@jasonLaster we are a bit confused by this ticket

jasonLaster:
@bomsy what's up?

I might have been vague w/ the css patch... you'll need to change it somewhere else :)

PR #6059, opened 2018-04-23
[Tabs] Update the show more button 

pranoy10:
Fixes Issue: #6012

### Summary of Changes

Changed the icon. Although not sure if this is the fix that is required.

### DIFF
--- a/src/components/SecondaryPanes/CommandBar.css
+++ b/src/components/SecondaryPanes/CommandBar.css
@@ -49,7 +49,7 @@ img.skipPausing {
 }
 
 .command-bar img.stepOut {
-  mask: url(/images/stepOut.svg) no-repeat;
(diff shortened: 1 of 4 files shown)


src/components/SecondaryPanes/CommandBar.css
```diff
@@ -49,7 +49,7 @@ img.skipPausing {
 }
 
 .command-bar img.stepOut {
-  mask: url(/images/stepOut.svg) no-repeat;
+  mask: url(/mc/command-chevron.svg) no-repeat;
```
jasonLaster: this was sample code for the svg... we'll need to add the chevron to the correct button in the tabs bar :)

darkwing:
Updated:

chevron

Annoyingly I can't use the `/mc/` path to the image so I've added it for now

Issue #6049, opened 2018-04-21
[Component] the content should be padded


jasonLaster:
Our new component pane should get the same padding as the other panes

STR: 
1. go to https://firefox-debugger-example-react-js.glitch.me/
2. set a breakpoint in App.jsx/onSortClick
3. click the button

Notice the left arrows need some padding in the "HelloWorld" pane

<img src='https://shipusercontent.com/f5b6725feb94f1cf80da209125a54d6d/Screen%20Shot%202018-04-21%20at%209.12.52%20AM.png' title='Screen Shot 2018-04-21 at 9.12.52 AM.png' width=378>

CamStan:
There's a group of four (@wmeulink, @No-one-alone, @Launia and myself) of us here at a university who are going to try to work on this bug together in order to get familiar with this process. So there may be comments, questions, or pull requests coming from any one of us. :grinning:

CamStan:
@jasonLaster, we seem to be having an issue replicating the bug. After navigating to [https://firefox-debugger-example-react-js.glitch.me/](url), the app.jsx file (ours is lower case) does not have an onSortClick function.

![image](https://user-images.githubusercontent.com/12384825/39606163-bf80d6c2-4ee8-11e8-9be6-6520c423d885.png)

So using the inspector I clicked on the button and had it take me to the onSortClick funciton on line 911 in main.js which is where I put a breakpoint. However, after clicking the button, no "HelloWorld" pane shows up.

![image](https://user-images.githubusercontent.com/12384825/39606241-44957d54-4ee9-11e8-8be8-e5078056dd1c.png)

We are all new to this, so any help would be appreciated.

CamStan:
Just realized a pull request was made for this 11 days ago, but the bug is still labeled as open. Was this already fixed?

jasonLaster:
yes, sorry

PR #6060, opened 2018-04-23
Added Padding for Accordion List

EmdaviesSeneca:
Fixes Issue: #6049

Here's the Pull Request Doc

### Summary of Changes

* Added padding for the Accordion list elements. (In this example, HelloWorld's pane now has padding).

### Test Plan
   1. Go to https://firefox-debugger-example-react-js.glitch.me/
   2. Set a breakpoint in App.jsx/onSortClick
   3. Click the button on the page
   4. Check the HelloWorld pane. There should be padding now.

### DIFF
--- a/src/components/SecondaryPanes/Frames/Frames.css
+++ b/src/components/SecondaryPanes/Frames/Frames.css
@@ -5,8 +5,7 @@
 .frames ul {
   list-style: none;
   margin: 0;
- 
-}
+ }
 
 .frames ul li {
   padding: 0 10px 0 21px;
@@ -57,8 +56,7 @@
 .frames ul li:focus {
   background-color: var(--theme-toolbar-background-alt);
   outline: none;
- 
-}
+ }
 
 .theme-dark .frames ul li:focus {
   background-color: var(--theme-tab-toolbar-background);
@@ -109,5 +107,4 @@
 
 .framework-component div {
 	padding: 4px;
-	
 }
--- a/src/components/SecondaryPanes/Frames/Frames.css
+++ b/src/components/SecondaryPanes/Frames/Frames.css
@@ -5,7 +5,7 @@
 .frames ul {
   list-style: none;
   margin: 0;
-  padding: 0;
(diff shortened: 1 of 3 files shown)


src/components/shared/Accordion.css
```diff
@@ -54,6 +54,7 @@
 .accordion ._content {
   border-bottom: 1px solid var(--theme-splitter-color);
   font-size: 12px;
+   padding: 3.5px;
```
bomsy: any need for `.5`? is `4px` not good?

bomsy:
Nice work! @EmdaviesSeneca. Just added a comment.

src/components/shared/Accordion.css
darkwing: Nice @EmdaviesSeneca -- I can't update this PR because you've done this work on master branch, but once you update this to `4px` we should be good :)

EmdaviesSeneca:
Hey there - I've changed the px size. Let me know if anything else needs to be done.

src/components/SecondaryPanes/Frames/Frames.css
```diff
@@ -105,3 +104,7 @@
 :root.theme-dark .annotation-logo svg path {
   fill: var(--theme-highlight-blue);
 }
+
+.framework-component div {
```
darkwing: This selector is too general, as it would add 4px to every DIV within the panel.

src/components/shared/Accordion.css
```diff
@@ -54,6 +54,7 @@
 .accordion ._content {
   border-bottom: 1px solid var(--theme-splitter-color);
   font-size: 12px;
+   padding: 4px;
```
darkwing: This was the original problem with the PR and this line needs to go away.

darkwing:
Thanks for looking at this @EmdaviesSeneca !  Since the work is done on master,  I can't jump in and help. Could you resubmit your patch on a feature branch?  Thank you!

PR #6062, opened 2018-04-23
Improve Editor gutter hover 

jasonLaster:
Fixes Issue: #5979 

### Summary of Changes

* highlights active lines on hover
* softer highlight for empty lines
* moves gutter styles to gutter.css

Todo

- it would be nice to animate off the highlight
- fix a bug with scrolling

### DIFF
--- a/src/components/Editor/Editor.css
+++ b/src/components/Editor/Editor.css
@@ -53,23 +53,36 @@ html[dir="rtl"] .editor-mount {
 
 .theme-light {
   --gutter-hover-background-color: #dde1e4;
+  --empty-gutter-hover-background-color: #f2f2f3;
 }
 
 .theme-dark {
   --gutter-hover-background-color: #414141;
 }
 
-:not(.empty-line):not(.new-breakpoint) > .CodeMirror-gutter-wrapper:hover {
+.editor-wrapper :not(.new-breakpoint) > .CodeMirror-gutter-wrapper:hover {
   width: 60px;
   height: 13px;
   left: -55px !important;
-  background-color: var(--gutter-hover-background-color) !important;
   mask: url(/images/breakpoint.svg) no-repeat;
   mask-size: 100%;
   mask-position: 0 1px;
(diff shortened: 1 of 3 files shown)


darkwing:
This PR scares me a little bit.  Prior to very, very recent changes, a lighter line number means you can't interact with it -- now we're changing it to the default state?  Seems like it would give the impression you couldn't interact with any lines.  Just my two cents.

jasonLaster:
Yeah, that could be true... 

I was thinking that this is softer on the eyes... but we could be overthinking this by a factor of 10x

jasonLaster:
@violasong @digitarald what do you think?

digitarald:
@jasonLaster I like it. There is a slight risk, but overall I don't think our short spell of styling disabled gutter entries will keep users from hovering over the gutter to delightfully discover this.

nit, hard to tell from low fps GIF, but the animation could be a little snappier with `0.1s` (vs `0.2s`). To animate out, could this use `transition` (vs `animation`)?

jasonLaster:
Thanks @digitarald i think that makes sense.

I'll look into transition. First time using these commands.

@mcote - http://g.recordit.co/JGYhTujZdr.gif

the str is just scroll a lot over the gutter, and then mouse over it...

violasong:
I like the idea of this lighter line number color by default because they're more visually heavy than they need to be currently. Without the zebra effect, I don't think the light color looks disabled.

I'm a bit worried that the very subtle different grays for both numbers and hover arrows in this prototype might not be clear enough. Maybe when hovering over the gutter, the non-breakable lines should actually go lighter, and not be hover-able? (I know we thought this looked broken previously, but I'm curious if it looks less broken when the zebra is in effect.)

jasonLaster:
Closing this as it's not clear how we want to move forward

jasonLaster:
seeing something weird
http://g.recordit.co/Ubanf7yEZ1.gif

Issue #6066, opened 2018-04-23
[Breakpoints] (Nit) Clicking on file name should selectLocation in Editor


darkwing:
Clicking on a filename (group label) above breakpoints should select the tab for the given source.

darkwing:
I started a PR for this feature but didn't finish it here:

https://github.com/devtools-html/debugger.html/pull/6011

That would be a an awesome starting point!

PR #6077, opened 2018-04-23
Fix 6066 - Improve Breakpoint Grouping Uniqueness and Functionality

darkwing:
Fixes Issue: #6066

Winding down my day by reopening this.

- Files of the same name are no longer grouped
- Files are sorted by name
- Clicking a group heading opens the given resource

### DIFF
--- a/src/components/SecondaryPanes/Breakpoints.js
+++ b/src/components/SecondaryPanes/Breakpoints.js
@@ -202,14 +202,26 @@ class Breakpoints extends Component<Props> {
         .map(url => {
           const split = getRawSourceURL(url).split("/");
           const file = split[split.length - 1];
+          const groupBreakpoints = groupedBreakpoints[url].filter(
+            bp => !bp.hidden && (bp.text || bp.originalText)
+          );
+
+          if (!groupBreakpoints.length) {
+            return null;
+          }
 
           return [
-            <div className="breakpoint-heading" title={url} key={url}>
+            <div
+              className="breakpoint-heading"
+              title={url}
+              key={url}
+              onClick={() =>
+                this.props.selectLocation(groupBreakpoints[0].location)
+              }
+            >
               {file}
             </div>,
-            ...groupedBreakpoints[url]
-              .filter(bp => !bp.hidden && bp.text)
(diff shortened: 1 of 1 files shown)


src/components/SecondaryPanes/Breakpoints.js
```diff
@@ -190,20 +187,33 @@ class Breakpoints extends Component<Props> {
...
     );
 
     return [
       ...Object.keys(groupedBreakpoints)
-        .sort()
-        .map(filename => {
+        .sort((urlA, urlB) => urlA.split("/").pop() > urlB.split("/").pop())
+        .map(url => {
+          const split = getRawSourceURL(url).split("/");
+          const file = split[split.length - 1].split("?")[0];
```
jasonLaster: It would be nice if this were a helper method next to `getFilename` perhaps `getFilenameFromRawUrl`:

- the helper function is easier to document the rationale
- easier to unit test 
- can be re-used in other places

src/components/SecondaryPanes/Breakpoints.js
```diff
@@ -190,20 +187,33 @@ class Breakpoints extends Component<Props> {
...
+          const groupBreakpoints = groupedBreakpoints[url].filter(
+            bp => !bp.hidden && (bp.text || bp.originalText)
+          );
+
+          if (!groupBreakpoints.length) {
+            return null;
+          }
```
jasonLaster: Nice! I think i saw this show

src/components/SecondaryPanes/Breakpoints.js
```diff
@@ -190,20 +187,33 @@ class Breakpoints extends Component<Props> {
...
+
           return [
-            <div className="breakpoint-heading" title={filename} key={filename}>
-              {filename}
+            <div
+              className="breakpoint-heading"
+              title={url}
+              key={url}
+              onClick={() => this.props.selectSourceURL(url)}
```
jasonLaster: We should use `selectSource` where we can because two sources can share the same URLs.

https://github.com/pranoy10/debugger.html/blob/chevronfix/src/actions/sources/select.js#L83

PR #6079, opened 2018-04-24
Add devtools-utils and devtools-components

jasonLaster:
### Summary of Changes

* copy over devtools-utils
* copy over devtools-components


lukaszsobek:
Luckily just a "few" lint and flow errors :)

PR #6068, opened 2018-04-23
Move devtools-reps to the debugger repo

nchevobbe:
Had to fix a ton of eslint issues because of the line length limit.

PR #6080, opened 2018-04-24
Use devtools-reps 0.23 

nchevobbe:
The patch to get reps into debugger starts to get big. Let's do this part here at least


jasonLaster:
closing as we'll likely add #6068. Feel free to re-open if i'm wrong.

PR #6081, opened 2018-04-24
Removes toJS from modifiers

lukaszsobek:
Currently modifiers are being converted to an object at a very high level in `searchContents`, `searchContentsForHighlight`, `traverseResults` when their value is being used in `buildQuery` or `doSearch`. 

This here is a try to remove the toJS statements and see if mochi passes (jest works already)

Also: It may be possible to get the search modifiers from the state instead of passing them through X functions for a cleaner workflow and a definite source of truth with fewer dependencies. Insight on that would be great.

Structure of passing modifiers through functions:
```
searchContents
	_modifiers = modifiers.toJS();

	functions using objects
	- getMatches
		- buildQuery (destructures modifiers)
	- find
		- doSearch
			- updateOverlay
				- searchOverlay
					- buildQuery
			- searchNext
				- getSearchCursor
					- buildQuery

searchContentsForHighlight
	_modifiers = modifiers.toJS();

	functions using objects
	- searchSourceForHighlight
		- updateOverlay
		- findNextOnLine
			- getSearchCursor
				- buildQuery

traverseResults
	- findPrev (toJS)
		- doSearch
	- findNext (toJS)
		- doSearch
```

### DIFF
--- a/src/actions/file-search.js
+++ b/src/actions/file-search.js
@@ -105,10 +105,9 @@ export function searchContents(query: string, editor: Object) {
     }
 
     const ctx = { ed: editor, cm: editor.codeMirror };
-    const _modifiers = modifiers.toJS();
-    const matches = await getMatches(query, selectedSource.text, _modifiers);
+    const matches = await getMatches(query, selectedSource.text, modifiers);
 
-    const res = find(ctx, query, true, _modifiers);
+    const res = find(ctx, query, true, modifiers);
     if (!res) {
       return;
     }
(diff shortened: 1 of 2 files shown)


darkwing:
@lukaszsobek Seems worth investigating if doing the file search changes in a separate PR would make this change easier since tests are melting down a bit.

lukaszsobek:
@darkwing Yep, you were right :)

jasonLaster:
we should be careful here and make sure this passes manually because i dont think mochitests test this

darkwing:
@lukaszsobek Would you like to take a shot at writing a test for this?

lukaszsobek:
Just did - it breaks, will update this PR soon.

lukaszsobek:
Closing. With workers we do not want immutable constructs, but only plain objects.

PR #6086, opened 2018-04-24
[Exceptions] Update Checkbox Options Language

### LINKED ISSUE #6064: [Breakpoints] (Nit) Change Exception Checkbox Labels
Needs to change to:

- [ ] Pause on Exceptions
  - [ ] Pause on Caught Exceptions

darkwing:
Fixes Issue: https://github.com/devtools-html/debugger.html/issues/6064

pausing

## ToDo
- [x] Upon refresh of the debugger I see the following error: "Error: non-bool passed to setBoolPref (Services.js:79:12)"; this error bricks the entire debugger.  For some reason `pause-on-caught-exceptions` is being set to `undefined`.  Tracing the `prefs` object and `dbg.store.getState().pause.shouldPauseOnCaughtExceptions` shows the values are being changed properly, so no idea where this error is coming from.

- [ ] I'm trying this out in the "Exceptions" example in the `debugger-examples` repo (http://localhost:7999/exceptions.html) and ... it's not working -- the caught exceptions are being stopped at regardless of this checkbox setting.  *Actually, with all checkboxes off in current master, the behavior is the same*


assets/panel/debugger.properties
```diff
@@ -104,6 +104,10 @@ pauseOnExceptionsItem=Pause on exceptions
 # when the debugger will not pause on any caught exception
 ignoreCaughtExceptionsItem=Ignore caught exceptions
 
+# LOCALIZATION NOTE (pauseOnCaughtExceptionsItem): The pause on exceptions checkbox description
+# when the debugger should puase on caught exceptions
```
flodolo: typo: puase

src/client/firefox/commands.js
```diff
@@ -272,11 +272,13 @@ async function getFrameScopes(frame: Frame): Promise<*> {
...
   shouldPauseOnExceptions: boolean,
-  shouldIgnoreCaughtExceptions: boolean
+  shouldPauseOnCaughtExceptions: boolean
 ): Promise<*> {
   return threadClient.pauseOnExceptions(
     shouldPauseOnExceptions,
-    shouldIgnoreCaughtExceptions
+    // Providing opposite value because server
+    // uses "shouldIgnoreCaughtExceptions"
+    !shouldPauseOnCaughtExceptions
```
darkwing: The way I'm thinking about this, we can change everything on our side, but simply reverse the value we're using when sending to server since the server still works on ignoreCaughtExceptions.

src/actions/pause/pauseOnExceptions.js
```diff
@@ -14,16 +14,16 @@ import type { ThunkArgs } from "../types";
...
   return ({ dispatch, client }: ThunkArgs) => {
     dispatch({
       type: "PAUSE_ON_EXCEPTIONS",
       shouldPauseOnExceptions,
-      shouldIgnoreCaughtExceptions,
+      shouldPauseOnCaughtExceptions,
       [PROMISE]: client.pauseOnExceptions(
         shouldPauseOnExceptions,
-        shouldIgnoreCaughtExceptions
+        !shouldPauseOnCaughtExceptions
```
darkwing: I *think* we need the `!` here.

assets/panel/debugger.properties
darkwing: WHY DO I KEEP DOING THAT?!  Seriously, I did that like 100 times while working on this stuff.  Making me so mad.

jasonLaster:
Looks great! I'll test soon

jasonLaster:
code looks good, but i'm seeing this error when i refresh




Also, when i use this page to test
http://firefox-dev.tools/debugger-examples/examples/exceptions.html

it looks like i'm always pausing on caught exceptions, even w/o the checkbox checked

darkwing:
Updates made, tests passing, worth another look @jasonLaster

Issue #6065, opened 2018-04-23
[Breakpoints] (Nit) Tighten Breakpoints Spacing


darkwing:
The spacing between breakpoints at the moment is a bit much.  @violasong will provide a more specific target for spacing :)

<img width="304" alt="breakpointsspacing" src="https://user-images.githubusercontent.com/46655/39133508-fee94630-46d9-11e8-808f-50e2f2529201.png">

violasong:
Here are some nits/suggestions :). I'm providing the CSS just so you can see what I think it should be visually but any visual equivalent of these is good too. Forgot to take a screenshot of the result, oops.

.breakpoints-exceptions
padding-bottom: .5em;
padding-top: .5em;

.breakpoints-list .breakpoint-heading
padding-top: .75em

.breakpoints-list .breakpoint-label
font-size: 11px

.breakpoints-list .breakpoint-heading, .breakpoints-list .breakpoint, .breakpoints-exceptions
padding: 0.25em 1em .25em 0.5em;

.pane.breakpoints-list
padding-bottom: .35em;

.breakpoints-exceptions-caught
    margin-top: -.25em;

Potential bug I noticed - when hovering over a breakpoint code snippet, I'm unable to scroll the panel (using 2-finger scroll on MacOS)

I'm looking at the code snippet and realizing that for short lines of code, it includes the next line. I feel like this might be easier to parse if we just do at most the one line? Not sure about this though. An extra space between line would help.

PR #6098, opened 2018-04-25
Breakpoints Nits:  Update breakpoints spacing per Victoria's suggestion

darkwing:
victoriaschanges

Issue: #6065

### DIFF
--- a/src/components/SecondaryPanes/Breakpoints.css
+++ b/src/components/SecondaryPanes/Breakpoints.css
@@ -7,7 +7,7 @@
 }
 
 .pane.breakpoints-list {
-  padding-bottom: .35em;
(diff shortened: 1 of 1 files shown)


darkwing:
@violasong Thoughts?

violasong:
@darkwing Ah, oops, I think I might have left out a style or two since this doesn't quite match what I was looking at - I think there may be a 3px padding that needs to be removed so that the checkbox rows can be closer together. Also needs about 3px more spacing above the first file name. For next time I can put the mockups into https://zeplin.io/ (or I could add you to the Mozilla Invision instance for their inspection feature) so you can hover to look at the exact pixel amounts for these things

PR #6099, opened 2018-04-25
[scopes] If there are too many possible bindings, assume they are all bad.

loganfsmyth:
Refs Issue: #5561

### Summary of Changes

* Consider all bindings bad if there are too many

All the tests pass with this, so I'm happy with these numbers. We can always tune them in the future.


jasonLaster:
this is a nice check! I'd love to get some data on it... but we'll see

PR #6112, opened 2018-04-26
Revert edge cases

jasonLaster:
### Summary of Changes

we're seeing issues with the preview-sourcemaps test. This should fix that:

screen shot 2018-04-26 at 9 28 46 am


loganfsmyth:
Is there a link to the failed builds somewhere? I'm not able to reproduce that error locally.

loganfsmyth:
Okay! This test fails when `mapScopes` is enabled, which it isn't locally for me and as far as I know, isn't in debugger.html's CI, but it must be on whatever environment this error was triggered in.

It's definitely an issue of the logic in those PRs being overly broad, so I'll see what I can do.

PR #6159, opened 2018-04-26
Revert "Revert "[Exceptions] Update Checkbox Options Language (#6086)…

jasonLaster:


jasonLaster:
seeing this error
https://travis-ci.org/devtools-html/debugger.html/builds/371650228?utm_source=github_status&utm_medium=notification

darkwing:
@jasonLaster You said not to touch this on Friday because you knew what was going on here -- any insight?

jasonLaster:
looking soon

jasonLaster:
running some travis runs here:
- [baseline](https://travis-ci.org/devtools-html/debugger.html/builds/373202728)
- [prefs change](https://travis-ci.org/devtools-html/debugger.html/builds/373201672)

jasonLaster:
so the tweaks lead to helpful travis errors:
https://travis-ci.org/devtools-html/debugger.html/builds/373202728

darkwing:
So is Firefox using the pref and not what we're sending from the thread client?  Any tiny bit of background here?

darkwing:
I'm slightly concerned that we're going to have to manage two prefs, or should we bail on the new pref, revert to old, and simply set a reverse value?

jasonLaster:
I think the issue is that the old pref is used by the old debugger...
we shouldnt use it

darkwing:
Travis seems happy

Issue #5978, opened 2018-04-15
[Breakpoints] Scrolling over a breakpoint is suppressed


jasonLaster:
It looks like we have a funky bug with codemirror and scrolling in the right sidebar:

STR:

1. add some breakpoints
2. scroll in the right sidebar so that the mouse is over breakpoint text

Notice how the panel won't scroll any more

![](http://g.recordit.co/rshcWzgOr7.gif)

PR #6162, opened 2018-04-26
Prevent CodeMirror from preventing scroll in breakpoints

darkwing:
Fixes Issue: #5978 

### Test Plan

Pull this down, build into Firefox, then try to scroll over breakpoints

I figured this out via:

``` ./mach run --jsdebugger --devtools ```, setting this CSS value with the console, and then scrolling worked as desired.


lukaszsobek:
Wow, didn't even know such a property existed. Great job!

PR #6164, opened 2018-04-26
removes toJS from ResultsList

lukaszsobek:
Since .map() and .reverse() work on Immutable maps there might be no need for the .toJS() at all.


lukaszsobek:
ouch - I'll close this for now until we have more security.

PR #6169, opened 2018-04-27
Removes repeating .toJS() calls in Outline.js

lukaszsobek:
since every .toJS() call is about a quarter of a second we should be able to shave off half a second with this refactor every time Outline is called.


lukaszsobek:
@jasonLaster that looks goooood 👍

nchevobbe:
so, we can do that because we still get a record in which we can access properties, right ?

lukaszsobek:
@nchevobbe exactly, for performing operations on it a map is perfectly fine. The only place we really want a plain object is when dispatching.

PR #6171, opened 2018-04-28
Cleans up redux connect statements

lukaszsobek:
Clean up and order connect statements, remove nesting so redux connect calls are more readable and less error prone. Remove bindActionCreators.

### DIFF
--- a/src/components/Editor/Footer.js
+++ b/src/components/Editor/Footer.js
@@ -5,7 +5,6 @@
 // @flow
 import React, { PureComponent } from "react";
 import { connect } from "react-redux";
-import { bindActionCreators } from "redux";
 import actions from "../../actions";
 import {
   getSelectedSource,
@@ -199,17 +198,17 @@ class SourceFooter extends PureComponent<Props> {
   }
 }
 
-export default connect(
-  state => {
-    const selectedSource = getSelectedSource(state);
-    const selectedId = selectedSource.get("id");
-    const source = selectedSource.toJS();
-    return {
-      selectedSource,
-      mappedSource: getGeneratedSource(state, source),
-      prettySource: getPrettySource(state, selectedId),
-      endPanelCollapsed: getPaneCollapse(state, "end")
-    };
-  },
-  dispatch => bindActionCreators(actions, dispatch)
-)(SourceFooter);
+const mapStateToProps = state => {
+  const selectedSource = getSelectedSource(state);
+  const selectedId = selectedSource.get("id");
+  const source = selectedSource.toJS();
+
+  return {
+    selectedSource,
+    mappedSource: getGeneratedSource(state, source),
+    prettySource: getPrettySource(state, selectedId),
+    endPanelCollapsed: getPaneCollapse(state, "end")
+  };
+};
+
+export default connect(mapStateToProps, actions)(SourceFooter);
--- a/src/components/SecondaryPanes/Breakpoints.js
+++ b/src/components/SecondaryPanes/Breakpoints.js
@@ -263,7 +263,7 @@ const _getBreakpoints = createSelector(
       .filter(bp => bp.source && !bp.source.isBlackBoxed)
 );
 
-const mapStateToProps = (state, props) => ({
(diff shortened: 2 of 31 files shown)


src/components/Editor/Footer.js
```diff
@@ -199,17 +198,17 @@ class SourceFooter extends PureComponent<Props> {
...
+
+  return {
+    selectedSource,
+    mappedSource: getGeneratedSource(state, source),
+    prettySource: getPrettySource(state, selectedId),
+    endPanelCollapsed: getPaneCollapse(state, "end")
+  };
+};
+
+export default connect(mapStateToProps, actions)(SourceFooter);
```
wldcordeiro: You seem to jump around between passing in `actions` and a `mapDispatchToProps` object. I kind of prefer the latter as it then limits the number of action creators passed to the component (making it easier to read prop lists in react devtools 😄 )

src/components/SecondaryPanes/EventListeners.js
```diff
@@ -99,20 +98,19 @@ class EventListeners extends Component<Props> {
...
+      ...l,
+      breakpoint: getBreakpoint(state, {
+        sourceId: l.sourceId,
+        line: l.line,
+        column: null
+      })
+    };
+  });
+
+  return { listeners };
```
wldcordeiro:
```js
const mapStateToProps = state => ({
  listeners: getEventListeners(state).map(listener => ({
    ...listener,
    breakpoint: getBreakpoint(state, {
      sourceId: listener.sourceId,
      line: listener.line,
      column: null
    })
  }),
});
```

src/components/Editor/Footer.js
lukaszsobek: I agree with the notion, the actions should either be extracted or passed as a whole - having it done a different way every time creates confusion. That is the scope for a different pr however.

The aim here was to clean up the connect, get rid of bindActionCreators etc.

wldcordeiro:
Merge conflicts here. 😦

src/components/SecondaryPanes/index.js
```diff
@@ -400,18 +400,16 @@ SecondaryPanes.contextTypes = {
...
-    breakpointsLoading: getBreakpointsLoading(state),
-    isWaitingOnBreak: getIsWaitingOnBreak(state),
-    shouldPauseOnExceptions: getShouldPauseOnExceptions(state),
-    shouldIgnoreCaughtExceptions: getShouldIgnoreCaughtExceptions(state),
-    workers: getWorkers(state)
-  }),
-  dispatch => bindActionCreators(actions, dispatch)
-)(SecondaryPanes);
+const mapStateToProps = state => ({
+  extra: getExtra(state),
```
lukaszsobek: Misses expressions

src/components/SecondaryPanes/index.js
```diff
@@ -16,7 +15,6 @@ import {
   getBreakpoints,
   getBreakpointsDisabled,
   getBreakpointsLoading,
-  getExpressions,
   getIsWaitingOnBreak,
```
lukaszsobek: Misses getExpressions

src/components/SecondaryPanes/EventListeners.js
lukaszsobek: Done

src/components/SecondaryPanes/Breakpoints.js
```diff
@@ -264,10 +263,9 @@ const _getBreakpoints = createSelector(
...
 );
 
-export default connect(
-  (state, props) => ({
-    breakpoints: _getBreakpoints(state),
-    selectedSource: getSelectedSource(state)
-  }),
-  dispatch => bindActionCreators(actions, dispatch)
-)(Breakpoints);
+const mapStateToProps = (state, props) => ({
```
lukaszsobek: no props used

jasonLaster:
I love this! Lets review/qa this soon so we can land before the conflicts come

PR #6173, opened 2018-04-29
Removes a toJS()

lukaszsobek:
`getSourceClassnames()` works the same way when it gets a clean object or something from immutable. The result is the same.


src/utils/source-maps.js
```diff
@@ -16,7 +16,7 @@ export async function getGeneratedLocation(
 
   const { line, sourceId, column } = await sourceMaps.getGeneratedLocation(
     location,
-    source
+    source.toJS()
```
darkwing: Was this intended?

lukaszsobek: Yes, it was - moving the toJS closer to the source.

Issue #5467, opened 2018-02-19
[WatchExpressions] Autocomplete for editing watch expressions


Tazmain:
Can we please add the functionality to have autocomplete, when editing a watch expression in the debugger tab?

jasonLaster:
Here is the actor endpoint

https://searchfox.org/mozilla-central/source/devtools/server/actors/webconsole.js#1068


Here is the client support for the webconsole

https://searchfox.org/mozilla-central/source/devtools/client/webconsole/jsterm.js#1513-1535

It'd be fun to see if someone wants to try and prototype this feature in the debugger. Perhaps without a UI.

bomsy:
So some of the things we need to do here
- [ ] Add client api for autocomplete support
- [ ] Add UI for autocomplete functionality
- [ ] Support caching so no need to call the api always (improve perfomance)

bomsy:
Hey @marcogallo566 , yes i've  got a solution for it. Sure you can help ... i'll write out a diff for you to start off with in a bit.

marcogallo566:
@bomsy Any news about the UI?

bomsy:
Hey @marcogallo566 
Hey is a gif of the my idea of what this will look like
![](http://g.recordit.co/ZObtEYOqfe.gif)

bomsy:
Hey @marcogallo566 

Here is my start on it. This diff is quite buggy so u'll have to fix some of the issues.
if you need any help! Let us know in the slack channel!

Cheers!

```diff
diff --git a/src/components/SecondaryPanes/Expressions.js b/src/components/SecondaryPanes/Expressions.js
index fe188555..e8e15693 100644
--- a/src/components/SecondaryPanes/Expressions.js
+++ b/src/components/SecondaryPanes/Expressions.js
@@ -9,10 +9,16 @@ import classnames from "classnames";
 import { ObjectInspector } from "devtools-reps";

 import actions from "../../actions";
-import { getExpressions, getExpressionError } from "../../selectors";
+import {
+  getExpressions,
+  getExpressionError,
+  getAutocompleteMatchset
+} from "../../selectors";
 import { getValue } from "../../utils/expressions";
 import { createObjectClient } from "../../client/firefox";

+import { debounce } from "lodash";
+
 import CloseButton from "../shared/Button/Close";

 import type { List } from "immutable";
@@ -37,7 +43,9 @@ type Props = {
   deleteExpression: (expression: Expression) => void,
   openLink: (url: string) => void,
[...]

PR #6177, opened 2018-04-30
add the client api for autocomplete

bomsy:
Fixes Issue: #5467

### Summary of Changes

* Added `autocomplete` client api
* Added a new `autocomplete` action
* Added tests for the action
* Add reducer to locally cache the matches


lukaszsobek:
Flow complaining.

Issue #6166, opened 2018-04-26
(Nit) [Expressions] Misc tweaks to empty state/add behavior


violasong:
- [x] Plus button should match stepping buttons in color

- [x] When no Watch Expressions, should hide Refresh button

- [x] Dotted-outline focus ring that appears when clicking + seems unnecessary

- [x] Clicking + should always open the accordion and change focus to the input field

- [x] Watch expressions rows should be taller, to match the new breakpoint styling

phoenixking25:
row should be taller mean have to increase height of the rows in css? @darkwing

darkwing:
In a previous conversation @violasong mentioned that `1em` may be the height we want -- give it a shot :)

violasong:
It should visually match the height of the breakpoint rows, so borrowing some styles from there might help :)

phoenixking25:
![screenshot from 2018-05-05 19-10-51](https://user-images.githubusercontent.com/22527880/39664057-0e501ec6-509b-11e8-98b3-f31381a51d3f.png)
![screenshot from 2018-05-05 19-11-11](https://user-images.githubusercontent.com/22527880/39664058-0ead57ee-509b-11e8-9ec4-3ee07b5283af.png)

phoenixking25:
added `min-height: 3em` to `.expression-container`  Is it looking alright with this or add some padding too? @violasong

PR #6182, opened 2018-04-30
Fix a few nits from #6166

darkwing:
This addresses two nits from #6166:

- Removes outline
- Opens accordion item if not open and the "+" is clicked

Also:

- Now always shows the "+", since the pane could be closed and someone might know how to open it to add an expression

Test plan:

- Click the "+" icon with the pane closed and opened, ensure the input gets focused
- Edit an existing expression, ensure the editing experience is good.

I'll address the last item in that issue once we prefect breakpoints.

### DIFF
--- a/src/components/SecondaryPanes/index.js
+++ b/src/components/SecondaryPanes/index.js
@@ -213,6 +213,10 @@ class SecondaryPanes extends Component<Props, State> {
   }
 
   getWatchItem(): AccordionPaneItem {
+    if (this.state.showExpressionsInput) {
+      prefs.expressionsVisible = true;
+    }
+
     return {
       header: L10N.getStr("watchExpressions.header"),
       className: "watch-expressions-pane",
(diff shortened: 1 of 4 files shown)


src/components/SecondaryPanes/index.js
```diff
@@ -213,6 +213,10 @@ class SecondaryPanes extends Component<Props, State> {
   }
 
   getWatchItem(): AccordionPaneItem {
+    if (this.state.showExpressionsInput) {
+      prefs.expressionsVisible = true;
+    }
```
jasonLaster: what does this do?

darkwing: Bailed on that change, found a cleaner way.

darkwing:
What kid of sorcery did you perform to get that?

outline

jasonLaster:
oof

darkwing:
Going to merge -- this PR wouldn't have created that issue.  I still have no idea how you did that, so we'll monitor it.

jasonLaster:
sounds good

PR #6187, opened 2018-05-01
Works on project-search

lukaszsobek:
- Refactors project-search
- Adds project-search tests

### DIFF
--- a/src/workers/search/project-search.js
+++ b/src/workers/search/project-search.js
@@ -4,7 +4,7 @@
 
 // Maybe reuse file search's functions?
 
-export function findSourceMatches(source = {}, queryText) {
+export function findSourceMatches(source, queryText) {
   const { id, loadedState, text } = source;
   if (loadedState != "loaded" || !text || queryText == "") {
     return [];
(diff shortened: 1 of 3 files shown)


src/workers/search/project-search.js
```diff
@@ -4,23 +4,23 @@
 
 // Maybe reuse file search's functions?
 
-export function findSourceMatches(source, queryText) {
-  const text = source.text;
-  if (source.loadedState !== "loaded" || !text || queryText == "") {
+export function findSourceMatches(source = {}, queryText) {
```
lukaszsobek: Not sure if we really want `findSourceMatches` to gracefully handle the lack of a source parameter or just outright throw. I'd say throwing is better for the application in terms of bug hunting and will save development time.

darkwing: Yeah, I think we should throw if there's no source, otherwise we're just sort of gracefully doing work we don't need to be doing.

bomsy: i wouldagree we should not default to an empty object as `findSourceMatches` should always be called with a source. if it is not, it should probably fail . How do we differentiate between an invalid or broken source (e.g source is null ) and a valid source that has no matches ?

PR #6191, opened 2018-05-01
[Preview] an additional tweak

jasonLaster:
Fixes Issue:  #5406

### Summary of Changes

I feel like we're writing industrial strength hover protection. Anyways, this fixes two phenomenons 

1. there is no relatedTarget (small mistake on the last pr)
2. the scenario that the user moused away before the preview was shown...


jasonLaster:
so yeah... mochitests moves the mouse away from the element immediately... forgot about that :)

jasonLaster:
closing... i think the new approach can circumvent this check

PR #6195, opened 2018-05-02
Adds some source-tree tests

lukaszsobek:
Adds source-tree reducer tests

### DIFF
--- a/src/reducers/tests/source-tree.spec.js
+++ b/src/reducers/tests/source-tree.spec.js
@@ -1,0 +1,33 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
+
+import update, { InitialState, getExpandedState } from "../source-tree";
+
+describe("source-tree", () => {
+  it("creates initial state", () => {
+    const state = InitialState();
+    const action = { type: "@@INIT" };
+    const updatedState = update(state, action);
+    expect(updatedState.toJS()).toEqual({
+      expanded: null
+    });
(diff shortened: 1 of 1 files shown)


src/reducers/tests/source-tree.spec.js
```diff
@@ -0,0 +1,33 @@
...
+
+  it("updates state", () => {
+    const expanded = "updateTest";
+    const state = InitialState();
+    const action = {
+      type: "SET_EXPANDED_STATE",
+      expanded
+    };
+    const updatedState = update(state, action);
+    expect(updatedState.toJS()).toEqual({ expanded });
```
lukaszsobek: should be just `expect(updatedState).toEqual(state);`

src/reducers/tests/source-tree.spec.js
```diff
@@ -0,0 +1,33 @@
...
+ * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
+
+import update, { InitialState, getExpandedState } from "../source-tree";
+
+describe("source-tree", () => {
+  it("creates initial state", () => {
+    const state = InitialState();
+    const action = { type: "@@INIT" };
+    const updatedState = update(state, action);
+    expect(updatedState.toJS()).toEqual({
```
lukaszsobek: should be just `expect(updatedState).toEqual(state);`

jasonLaster:
we actually dont write reducer tests because they're in theory tested by the action tests
https://github.com/devtools-html/debugger.html/blob/master/src/actions/tests/source-tree.spec.js#L18

could you check here and see if there are any missing tests?

darkwing:
Closing this per @jasonLaster's comments, but please feel free to reopen if you feel appropriate @lukaszsobek !

Issue #6051, opened 2018-04-21
[Preview] Popup gets clipped on top


jasonLaster:
When you're hovering over a big object that won't fit below, we try to show it above the variable. Unfortunately, showing it above is often not a good option either.

In these cases, it would be great if we could show the popup to the right of the variable and try and use all of the available space needed...

<img src='https://shipusercontent.com/5869fd0f83eded871fce7a50c3a32ddf/Screen%20Shot%202018-04-21%20at%209.23.27%20AM.png' title='Screen Shot 2018-04-21 at 9.23.27 AM.png' width=587>

shirlyb3:
/claim
I think I found the location where it can be changed:
https://github.com/devtools-html/debugger.html/blob/master/src/components/shared/Popover.js#L76

jasonLaster:
How is it going @shirlyb3? Anything i can help with?

shirlyb3:
@jasonLaster going well, almost done - there's only one case the preview is not located in the way I wanted it to, so I'm checking now what I missed.

I might need help with the tests, I have 2 popover tests failing - 
it expects the preview to be located over the token, but now according to the `editorRect` and `targetRect` passed to the popover component in the test, the position calculation in my implementation places it on the right side of the token.
This is the one failing -  https://github.com/devtools-html/debugger.html/blob/master/src/components/shared/tests/Popover.spec.js#L73
But I'm not 100% sure the test is wrong (and also I don't want to be one of those programmers saying "tests are wrong my code is awesome"), so I feel like I need some approval. Also, I never created or updated a .snap file, so I'm not sure what's the procedure - can I just change the expected component props?

PR #6196, opened 2018-05-02
Preview to the right

shirlyb3:
Fixes Issue: #6051 

### Summary of Changes
Most of the changes are in getPopoverCoords, now it's doing 3 things mainly:
1. calculates the popover orientation - should it be placed on top/bottom/right of the token, using the target, popover and editor positions.
2. Using the orientation to calculate the left and top of the popover
3. To determine where the popover arrow should be - calculates `targetMid`

right

### Test Plan

- Set a breakpoint and stop

On large window, behavior should remain unchanged:
- Enlarge the debugger window to max size
- Pick an object to test
- Place the token in the top of the editor.
- Preview opens **under** the token.
- Place the token in the vertical middle of the editor.
- Preview opens **under** the token.
- Place the token in the bottom of the editor
- Preview opens **above** the token.

On Small window, right orientations should affect:
- Decrease the debugger window dimensions so it fits the preview popover fully, but not much more
- Place the token in the top of the editor.
- Preview opens **under** the token.
- Place the token in the 1/4 of the editor.
- Preview opens **right** to the token,the popover's top is clipped to the editor's top.
- Place the token in the vertical middle of the editor.
- Preview opens **right** to the token, the token is centered vertically to the popover.
- Place the token in the 3/4 of the editor.
- Preview opens **right** to the token,the popover's bottom is clipped to the editor's bottom.
- Place the token in the bottom of the editor
- Preview opens **above** the token.

- Decrease the debugger window dimensions so it's smaller then the popover
- Place the token anywhere
- Preview opens to the right of the token, clipped to the top of the window.

### DIFF
--- a/src/components/shared/BracketArrow.css
+++ b/src/components/shared/BracketArrow.css
@@ -44,3 +44,20 @@
   border-top-color: var(--theme-body-background);
   top: -1px;
 }
+
+.bracket-arrow.left::before {
+  border-left-color: transparent;
+  border-right-color: var(--theme-splitter-color);
+  top: 0px;
+}
+
+.theme-dark .bracket-arrow.left::before {
+  border-right-color: var(--theme-body-color);
+}
+
+.bracket-arrow.left::after {
+  border-left-color: transparent;
+  border-right-color: var(--theme-body-background);
+  top: 0px;
+  left: 1px;
+}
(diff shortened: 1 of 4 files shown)


jasonLaster:
Great job. It looks really good

I checked to see what chrome does and their algorithm is really nice too. They shrink the height of the popup so that it fits.



I think this is better though because we can show the user more information!

amitzur:
I also think this is better! But still there's a height overflow problem when the popover doesn't fit. But we believed this belongs to another issue :)

Issue #6073, opened 2018-04-23
Identify react functional component as react source


yaelballa:
When debugging react sources there is an icon on the tab of react component files. But if the code includes a react functional component, there is no class definition, and therefore the icon does not appear.  

A solution to this problem should be to look for 'import react' when detecting framework

PR #6198, opened 2018-05-02
Find react by import and function

darkwing:
Fixes Issue: #6073
Supercedes: https://github.com/devtools-html/debugger.html/pull/6074

Moved @yaeballa's PR to a feature branch and added a modification to prevent test issues.


loganfsmyth:
FYI, this exposed an existing issue with `extendsReactComponent` since it is being called more often now.

PR #6204, opened 2018-05-03
Renames SELECT_SOURCE_URL action to setSourceLocation

lukaszsobek:
It seems the action SELECT_SOURCE_URL actually sets/selects the (pending?) source location instead of just setting the URL. It would be great to:
1.) rename this in order to mitigate a naming conflict with `selectSourceURL`
2.) externalize it for easier testing

### DIFF
--- a/src/actions/sources/setSourceLocation.js
+++ b/src/actions/sources/setSourceLocation.js
@@ -1,0 +1,11 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
+
+import type { Action } from "../types";
+
+export const setSourceLocation: Action = (url, options) => ({
+  type: "SET_SOURCE_LOCATION",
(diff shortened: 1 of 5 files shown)


src/actions/sources/setSourceLocation.js
```diff
@@ -0,0 +1,11 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
+
+import type { Action } from "../types";
+
+export const setSourceLocation: Action = (url, options) => ({
+  type: "SET_SOURCE_LOCATION",
+  url: url,
+  line: options.location ? options.location.line : null
+});
```
bomsy: Maybe we can move this to a common file with other functions? so we don't end up with many single files with only one function in them.

lukaszsobek: Sounds like a great idea, there are a few of these actions and having 20 files sounds like a bit too much - what would you call that file?

lukaszsobek:
Closing due to rescoping: #6205

PR #6205, opened 2018-05-03
Moves action creators in selected.js to own functions

lukaszsobek:
moves action creators in selected.js to own functions because:
1.) easier testability
2.) reduced risk of typos


src/actions/sources/clearSelectedSource.js
```diff
@@ -0,0 +1,9 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
+
+import type { Action } from "../types";
+
+export const clearSelectedSource: Action = () => ({
+  type: "CLEAR_SELECTED_SOURCE"
+});
```
bomsy: Maybe we can move this to a common file with other functions? so we don't end up with many single files with only one function in them.

lukaszsobek: Great idea, how would you name it?

src/actions/sources/index.js
```diff
@@ -3,6 +3,8 @@
  * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
 
 // @flow
+import { clearSelectedSource } from "./clearSelectedSource";
+export { clearSelectedSource };
```
darkwing: Can we just do `export * from ./clearSelectedSource`?

src/actions/sources/tests/sources.spec.js
```diff
@@ -0,0 +1,12 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
+
+import { clearSelectedSource } from "../";
+
+test("clearSelectedSource", () => {
```
lukaszsobek: 2 tests here and they need to be more meaningful than this.

src/actions/sources/select.js
```diff
@@ -113,13 +115,7 @@ export function selectLocation(location: Location) {
...
 
     dispatch(addTab(source, 0));
-    dispatch(
-      ({
-        type: "SELECT_SOURCE",
-        source,
-        location
-      }: Action)
-    );
+    dispatch(setSelectedLocation(source, location));
```
lukaszsobek: feels like an appropriate name and avoids a naming conflict with existing `selectSource`

src/actions/sources/index.js
```diff
@@ -3,10 +3,10 @@
  * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
 
 // @flow
-
+export * from "./blackbox";
```
lukaszsobek: alphabetical sorting

darkwing:
What's the status of this @lukaszsobek ?

lukaszsobek:
@darkwing - if you think the tests are ok this should be ready to go.

src/actions/sources/select.js
```diff
@@ -32,14 +32,30 @@ import {
...
+
+export const setPendingSelectedLocation = (url: string, options: Object) => ({
+  type: "SET_PENDING_SELECTED_LOCATION",
+  url: url,
+  line: options.location ? options.location.line : null
+});
+
+export const clearSelectedLocation = () => ({
+  type: "CLEAR_SELECTED_LOCATION"
+});
```
jasonLaster: I think i might be old school, but it would be nice for these to still be functions:

```js
export function clearSelectedLocation() {
  return {type: "CLEAR_SELECTED_LOCATION"}
}
```

this is pure style though

jasonLaster: hmm, what do you mean?

wldcordeiro: > hmm, what do you mean?

Sounds like @lukaszsobek is talking about `this` context and how `function() {}` produces its own.

PR #6206, opened 2018-05-03
Adds addTab and moveTab tests

lukaszsobek:
Adds missing tabs tests

### DIFF
--- a/src/actions/tests/tabs.spec.js
+++ b/src/actions/tests/tabs.spec.js
@@ -12,6 +12,22 @@ const { getSelectedSource, getSourceTabs } = selectors;
 
 import { sourceThreadClient as threadClient } from "./helpers/threadClient.js";
 
+import { addTab, moveTab } from "../sources/tabs";
+
+test("addTab returns expected action", () => {
+  const source = "testSource";
+  const tabIndex = "testTabIndex";
+  const output = addTab(source, tabIndex);
+  expect(output).toEqual({ type: "ADD_TAB", source, tabIndex });
+});
+
+test("moveTab returns expected action", () => {
+  const url = "testUrl";
+  const tabIndex = "testTabindex";
+  const output = moveTab(url, tabIndex);
+  expect(output).toEqual({ type: "MOVE_TAB", tabIndex, url });
+});
+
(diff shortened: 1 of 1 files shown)


src/actions/tests/tabs.spec.js
```diff
@@ -12,6 +12,22 @@ const { getSelectedSource, getSourceTabs } = selectors;
 
 import { sourceThreadClient as threadClient } from "./helpers/threadClient.js";
 
+import { addTab, moveTab } from "../sources/tabs";
+
+test("addTab returns expected action", () => {
+  const source = "testSource";
+  const tabIndex = "testTabIndex";
+  const output = addTab(source, tabIndex);
+  expect(output).toEqual({ type: "ADD_TAB", source, tabIndex });
```
bomsy: i'm sure if this check gives us any gains, as we need no certainty the tab was actually added.

We should probably instead use a selector to check if the tab was actually added to the redux store.

src/actions/tests/tabs.spec.js
```diff
@@ -12,6 +12,22 @@ const { getSelectedSource, getSourceTabs } = selectors;
...
+});
+
+test("moveTab returns expected action", () => {
+  const url = "testUrl";
+  const tabIndex = "testTabindex";
+  const output = moveTab(url, tabIndex);
+  expect(output).toEqual({ type: "MOVE_TAB", tabIndex, url });
```
bomsy: same as above

src/actions/tests/tabs.spec.js
lukaszsobek: I'd say that having the tab added is an integration test of action and reducer. We still have to know where the error occured and that information seems easier to obtain from  two unit tests where we know where the fault is instead of having an integration tests where we would still have to hunt down the bug.

bomsy: The `addTab` action just returns an object, i'm not sure even if an error occurs from dispatching the action, you get an error back here.

lukaszsobek:
Closing for now so as not to pollute the PR queue, will re-create at a later time.

PR #6207, opened 2018-05-03
Split mochitest/head.js in head.js and helpers.js

juliandescottes:
This PR splits the mochitest head.js in two files:
- head.js will contain everything that is needed to setup the test environment for the debugger mochitests (set preferences, load shared-head etc...)
- helpers.js will contain all the APIs that can drive the debugger

The added value is that helpers.js can be loaded from other panels without any issue. You cannot load head.js directly because shared-head.js is already imported by the panel and would throw.

I need this in order to test the service worker debugging in the application panel (see https://bugzilla.mozilla.org/show_bug.cgi?id=1458808). 

However note that this will make the following test fail: `client/framework/test/browser_browser_toolbox_debugger.js`. In order to fix the test we need to modify it as follows:

```
diff --git a/devtools/client/framework/test/browser_browser_toolbox_debugger.js b/devtools/client/framework/test/browser_browser_toolbox_debugger.js
--- a/devtools/client/framework/test/browser_browser_toolbox_debugger.js
+++ b/devtools/client/framework/test/browser_browser_toolbox_debugger.js
@@ -13,16 +13,17 @@ const { PromiseTestUtils } = scopedCuImp
 PromiseTestUtils.whitelistRejectionsGlobally(/File closed/);
 
 // On debug test runner, it takes about 50s to run the test.
 requestLongerTimeout(4);
 
 const { fetch } = require("devtools/shared/DevToolsUtils");
 
 const debuggerHeadURL = CHROME_URL_ROOT + "../../debugger/new/test/mochitest/head.js";
+const helpersURL = CHROME_URL_ROOT + "../../debugger/new/test/mochitest/helpers.js";
 const testScriptURL = CHROME_URL_ROOT + "test_browser_toolbox_debugger.js";
 
 add_task(async function runTest() {
   await new Promise(done => {
     let options = {"set": [
       ["devtools.debugger.prompt-connection", false],
       ["devtools.debugger.remote-enabled", true],
       ["devtools.chrome.enabled", true],
@@ -116,21 +117,25 @@ add_task(async function runTest() {
         }, interval);
       });
     }
   }).toSource().replace(/^\(function\(\) \{|\}\)$/g, "");
   /* eslint-enable no-unused-vars */
   // Stringify testHead's function and remove `(function {` prefix and `})` suffix
   // to ensure inner symbols gets exposed to next pieces of code
   // Then inject new debugger head file
-  let { content } = await fetch(debuggerHeadURL);
-  let debuggerHead = content;
+  let { content: debuggerHead } = await fetch(debuggerHeadURL);
   // We remove its import of shared-head, which isn't available in browser toolbox process
   // And isn't needed thanks to testHead's symbols
-  debuggerHead = debuggerHead.replace(/Services.scriptloader.loadSubScript[^\)]*\);/, "");
+  debuggerHead = debuggerHead.replace(/Services.scriptloader.loadSubScript[^\)]*\);/g, "");
+
+  // Also include the debugger helpers which are separated from debugger's head to be
+  // reused in other modules.
+  let { content: debuggerHelpers } = await fetch(helpersURL);
+  debuggerHead = debuggerHead + debuggerHelpers;
 
   // Finally, fetch the debugger test script that is going to be execute in the browser
   // toolbox process
   let testScript = (await fetch(testScriptURL)).content;
   let source =
     "try { let testUrl = \"" + testUrl + "\";" + testHead + debuggerHead + testScript + "} catch (e) {" +
     "  dump('Exception: '+ e + ' at ' + e.fileName + ':' + " +
     "       e.lineNumber + '\\nStack: ' + e.stack + '\\n');" +
```


juliandescottes:
An option is to stop testing browser_browser_toolbox_debugger.js for now and reenable it when this lands in mozilla central?

jasonLaster:
@juliandescottes this is great. I've wanted to do this as well so that we can begin organizing the helpers a bit more as well.

- Lets land this when tests pass
- Feel free to use helpers.js in your work now 
- When we land helpers.js in mc we can fix up the toolbox test

jasonLaster:
looks like the only failure was `client/framework/test/browser_browser_toolbox_debugger` because travis is not happy when you change its config in the middle

PR #6208, opened 2018-05-03
[ObjectInspector] Add some fulltext tests & minor optimisations

kootoopas:
Since I didn't have time to add these when the longstring expansion feature was
shipped, due to the then receeding deadline, I had left these out.

There's still need for some more unit tests (`setNodeFullText`,
`getFullText`) and integration tests such as the expand funtionality on the OI
component, but it's a start!

### DIFF
--- a/packages/devtools-reps/src/object-inspector/tests/component/create-long-string-client.js
+++ b/packages/devtools-reps/src/object-inspector/tests/component/create-long-string-client.js
@@ -72,7 +72,7 @@ describe("createLongStringClient", () => {
       const stub = longStringStubs.get("testLoadedFullText");
       const substring = jest.fn(() => Promise.resolve({ fullText: "" }));
       const createLongStringClient = jest.fn(grip =>
-        LongStringClient(grip, { a: x => x })
+        LongStringClient(grip, { substring })
       );
 
       mount(
@@ -95,5 +95,4 @@ describe("createLongStringClient", () => {
       expect(substring.mock.calls).toHaveLength(0);
     });
   });
-
 });
--- a/packages/devtools-reps/src/object-inspector/tests/component/create-long-string-client.js
+++ b/packages/devtools-reps/src/object-inspector/tests/component/create-long-string-client.js
@@ -42,9 +42,6 @@ describe("createLongStringClient", () => {
     it("is called for longStrings with unloaded full text", () => {
       const stub = longStringStubs.get("testUnloadedFullText");
       const substring = jest.fn(() => Promise.resolve({ fullText: "" }));
-      const createLongStringClient = jest.fn(grip =>
-        LongStringClient(grip, { substring })
-      );
 
       mount(
         ObjectInspector({
@@ -58,22 +55,19 @@ describe("createLongStringClient", () => {
             }
           ],
           createObjectClient: ObjectClient(stub),
-          createLongStringClient
+          createLongStringClient: grip => LongStringClient(grip, { substring })
         })
       );
 
       expect(substring.mock.calls[0]).toHaveLength(3);
-      const [initialLength, fullTextLength] = substring.mock.calls[0];
-      expect(initialLength).toBe(stub.initial.length);
-      expect(fullTextLength).toBe(stub.length);
+      const [start, length, callback] = substring.mock.calls[0];
+      expect(start).toBe(stub.initial.length);
+      expect(length).toBe(stub.length);
     });
 
     it("is not called for longString node w/ loaded full text", () => {
       const stub = longStringStubs.get("testLoadedFullText");
       const substring = jest.fn(() => Promise.resolve({ fullText: "" }));
-      const createLongStringClient = jest.fn(grip =>
-        LongStringClient(grip, { substring })
-      );
 
       mount(
         ObjectInspector({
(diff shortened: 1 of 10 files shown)


kootoopas:
Seems that `SourcesTree` tests are failing. I checked #6207 and there are failing `SourcesTree` tests there too. Same result when I run `yarn` test locally on master.

packages/devtools-reps/src/object-inspector/tests/component/create-long-string-client.js
```diff
@@ -0,0 +1,98 @@
...
+                value: stub
+              }
+            }
+          ],
+          createObjectClient: ObjectClient(stub),
+          createLongStringClient
+        })
+      );
+
+      expect(substring.mock.calls[0]).toHaveLength(3);
```
nchevobbe: what's the third one for ?

packages/devtools-reps/src/object-inspector/tests/component/create-long-string-client.js
```diff
@@ -0,0 +1,98 @@
...
+      const [initialLength, fullTextLength] = substring.mock.calls[0];
```
nchevobbe: could we name it firstArg, secondArg, or start and length ? This would better convey what they are

packages/devtools-reps/src/object-inspector/tests/component/create-long-string-client.js
```diff
@@ -0,0 +1,98 @@
...
+                value: stub
+              }
+            }
+          ],
+          createObjectClient: ObjectClient(stub),
+          createLongStringClient
+        })
+      );
+
+      expect(createLongStringClient.mock.calls[0][0]).toBe(stub);
```
nchevobbe: isn't it a bit weird that we do create the client if we aren't going to use it ?

packages/devtools-reps/src/object-inspector/utils/client.js
```diff
@@ -87,11 +89,18 @@ async function getPrototype(
...
-  object: NodeContents
+  item: Node
 ): Promise<{ fullText?: string }> {
-  const { initial, length } = object;
+  const { initial, fullText, length } = getValue(item);
 
   return new Promise((resolve, reject) => {
+    // Return fullText property if it exists so that it can be added to the
+    // loadedProperties map.
+    if (nodeHasFullText(item)) {
```
nchevobbe: could we do this outside of the Promise callback ? 
```js
if (nodeHasFullText(item)) {
  return Promise.resolve({ fullText })
}

return new Promise(…
```

It feels a bit easier to follow

packages/devtools-reps/src/reps/stubs/long-string.js
```diff
@@ -9,14 +9,19 @@ const fullTextLength = multilineFullText.length;
...
+
+const multilineLongString = {
   type: "longString",
   initial: initialText,
   length: fullTextLength,
   actor: "server1.conn1.child1/longString58"
-});
+};
+
+stubs.set("testMultiline", multilineLongString);
```
nchevobbe: maybe we could drop this and always use testUnloadedFullText ? What do you think ?

packages/devtools-reps/src/object-inspector/tests/component/create-long-string-client.js
kootoopas: It's the callback, which has access to the client response, since substring is async.

packages/devtools-reps/src/object-inspector/tests/component/create-long-string-client.js
kootoopas: There's no other way to verify that substring isn't called for loaded fullTexts, other than to spy on substring and pass it to the LongStringClient.

packages/devtools-reps/src/object-inspector/utils/client.js
kootoopas: indeed 👍

packages/devtools-reps/src/reps/stubs/long-string.js
kootoopas: In some other contexts in the test base we care about it being multiline. Hmm, probably the proper way to go is to make 2 separate stubs: one for multiline strings and one with no fullText.

packages/devtools-reps/src/object-inspector/tests/component/create-long-string-client.js
```diff
@@ -58,22 +55,19 @@ describe("createLongStringClient", () => {
             }
           ],
           createObjectClient: ObjectClient(stub),
-          createLongStringClient
+          createLongStringClient: grip => LongStringClient(grip, { substring })
```
kootoopas: Omitted LongStringClient spy and assertion since it's checked on a previous test.

packages/devtools-reps/src/object-inspector/tests/component/create-long-string-client.js
```diff
@@ -87,11 +81,10 @@ describe("createLongStringClient", () => {
...
```
kootoopas: Ditto

darkwing:
Assigning to @nchevobbe until I can get a better grasp of ObjectInspector and the work that got merged during the Paris week.  I look forward to it @nchevobbe !

kootoopas:
@darkwing Oh nice! Welcome to Europe. 🇪🇺

PR #6209, opened 2018-05-03
Add a disableFrameTruncate to the Frames component.

nchevobbe:
This prop prevents the debugger from truncating the frames
and show all of them on first render, as well as hide the
toggle button.
This will be useful for the console when we share this module
so we have a similar behavior with the current console stacktrace.


kootoopas:
Seems that `SourcesTree` tests are failing for all of us who have submitted prs today. :)

PR #6211, opened 2018-05-03
Render object as grip when noGrip is false

Loirooriol:
`noGrip` defaults to `false` so the default rep should be `Grip` instead of `Obj`. The current behavior causes problems like https://bugzil.la/1377668 and https://bugzil.la/1457701. Better fail fast and loud and require `noGrip = true` for raw objects or arrays.

### Summary of Changes

 - The default rep is now `Grip` instead of `Obj`.
 - The `Obj` and `ArrayRep` reps will refuse to render the object if `noGrip = true` is not specified.
 - Fixed various tests that relied on the previous bad behavior.

### Test Plan

I ported the changes to my mozilla-central repo and ensured that [browser_net_json-long.js](https://searchfox.org/mozilla-central/source/devtools/client/netmonitor/test/browser_net_json-long.js) fails, because netmonitor needs `noGrip = true` (https://bugzil.la/1457701). Better fail fast and loud than only in hard to detect edge cases. In this case, without `noGrip = true`, an object with properties is rendered as `{}` instead of `{…}`.

### DIFF
--- a/packages/devtools-reps/src/reps/array.js
+++ b/packages/devtools-reps/src/reps/array.js
@@ -142,10 +142,11 @@ function getLength(object) {
   return object.length;
 }
 
-function supportsObject(object) {
+function supportsObject(object, noGrip = false) {
   return (
-    Array.isArray(object) ||
-    Object.prototype.toString.call(object) === "[object Arguments]"
+    noGrip &&
+    (Array.isArray(object) ||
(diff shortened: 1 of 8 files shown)


nchevobbe:
Thanks Oriol, everything looks good to me. Let's merge this

PR #6216, opened 2018-05-03
Remove busted test until fixed in mc

darkwing:
We know this test is busted based on the fix/helpers PR, so I'm checking to see if we can remove it until we fix the test in MC


jasonLaster:
wait, is it busted? i think the other test for the toolbox was failing?

jasonLaster:
I think this is the busted test
devtools/client/framework/test/browser_browser_toolbox_debugger.js 

the config was somehow not cleared... it still has this test. Does this PR need to be rebased to pass?

screen shot 2018-05-03 at 5 34 14 pm

PR #6220, opened 2018-05-04
[WIP] Transpile debugger modules for DevTools loader

juliandescottes:
This is a work in progress PR, intended for collaboration and discussion, do not merge!

cc @jasonLaster @ochameau  

I started doing some cleanup, and rebased on latest master.

In my opinion, TODOs:
- [x] cleanup and comment copy-modules.js
- [x] cleanup and comment transform-mc.js
- [x] find alternative to hardcoding `window.DebuggerConfig` in debugger's index
- [x] find alternative to workaround in CallSite.js
- [x] fix mochitest failures
- [x] what to do about allowed dupes -> https://hg.mozilla.org/try/comparison/9b6fdd663985/browser/installer/allowed-dupes.mn
- [x] use copy-modules on travis

### DIFF
--- a/.babel/transform-mc.js
+++ b/.babel/transform-mc.js
@@ -1,0 +1,73 @@
+const _path = require("path");
+const fs = require("fs");
+
+function isRequire(t, node) {
+  return node && t.isCallExpression(node) && node.callee.name == "require";
+}
+
+module.exports = function({ types: t }) {
+  return {
+    visitor: {
+      ModuleDeclaration(path, state) {
+        const source = path.node.source;
+        const value = source && source.value;
+        if (value && value.includes(".css")) {
+          path.remove();
+        }
+      },
+
+      StringLiteral(path, state) {
+        const { mappings, vendors, filePath } = state.opts;
+        let value = path.node.value;
+
+        if (!isRequire(t, path.parent)) {
+          return;
+        }
+
+        // Transform mappings
+        if (Object.keys(mappings).includes(value)) {
+          path.replaceWith(t.stringLiteral(mappings[value]));
+          return;
+        }
+
+        // Transform vendors
+        if (vendors.includes(value)) {
+          path.replaceWith(
+            t.stringLiteral("devtools/client/debugger/new/vendors")
+          );
+
+          let name = value;
+          if (
+            path.parentPath &&
+            path.parentPath.parent &&
+            path.parentPath.parent.id &&
+            path.parentPath.parent.id.name
+          ) {
+            name = path.parentPath.parent.id.name;
+          }
+
+          path.parentPath.replaceWith(
+            t.memberExpression(path.parent, t.stringLiteral(name), true)
+          );
+          return;
+        }
+
+        const dir = _path.dirname(filePath);
+        const depPath = _path.join(dir, `${value}.js`);
+        const exists = fs.existsSync(depPath);
+        if (
+          !exists &&
+          !value.endsWith("index") &&
+          !value.startsWith("devtools")
+        ) {
+          path.replaceWith(t.stringLiteral(`${value}/index`));
+          return;
+        }
+      }
+    }
+  };
+};
+
+// const Services = require("Services");
+// const devtoolsModules = require("devtools/.../vendors")['devtools-modules']
+// const Services = devtoolsModules;
--- a/.babel/transform-mc.js
+++ b/.babel/transform-mc.js
@@ -80,7 +80,7 @@ module.exports = function({ types: t }) {
 
         // Handle require() to files bundled in vendor.js.
         // e.g. require("some-module");
-        //   -> require("devtools/client/debugger/new/vendors").vendored["some-module"];
+        //   -> require("devtools/client/debugger/new/dist/vendors").vendored["some-module"];
         const isVendored = VENDORS.some(vendored => value.endsWith(vendored));
         if (isVendored) {
           // components/shared/Svg is required using various relative paths.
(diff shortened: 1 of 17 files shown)


jasonLaster:
> find alternative to workaround in CallSite.js

could we just do `import "callsite.css"` basically what we do everywher?

juliandescottes:
Applied Jason's suggestion for Callsite.css. 
Cleaned up transform-mc and copy-modules.
Mutualized the m-c mappings between webpack-config and copy-modules/transform-mc

.babel/transform-mc.js
```diff
@@ -0,0 +1,107 @@
...
+ * This Babel plugin is used to transpile a single Debugger module into a module that
+ * can be loaded in Firefox via the regular DevTools loader.
+ */
+module.exports = function({ types: t }) {
+  return {
+    visitor: {
+      ModuleDeclaration(path, state) {
+        const source = path.node.source;
+        const value = source && source.value;
+        if (value && value.includes(".css")) {
```
juliandescottes: @jasonLaster I don't remember exactly the purpose of this part, can you comment?

mozilla-central-mappings.js
```diff
@@ -0,0 +1,30 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
juliandescottes: @jasonLaster I put this under the root right now, maybe we should have a firefox-integration folder or something?

webpack.config.js
```diff
@@ -44,6 +37,17 @@ const webpackConfig = {
   }
 };
 
+if (isDevelopment()) {
+  // In local development, use the debugger as a single bundle
+  webpackConfig.entry.debugger = getEntry("src/main.js");
+} else {
```
juliandescottes: @jasonLaster When we are not in development, we no longer build the debugger bundle, but we still build other bundles. Maybe we can simplify some things here?

juliandescottes:
The DebuggerConfig issue should be resolved by https://github.com/devtools-html/debugger.html/pull/6223, so marking this one as resolved. Still mochitests to fix now!

juliandescottes:
There is also a new dupe for the chevron:

browser/chrome/devtools/skin/images/command-chevron.svg
browser/chrome/devtools/skin/images/debugger/command-chevron.svg

juliandescottes:
Mochitests should now be fixed: the issue was with [build-query.js](https://github.com/devtools-html/debugger.html/blob/master/src/workers/search/build-query.js) where I earlier changed:

```
- import escapeRegExp from "lodash/escapeRegExp";
+ import { escapeRegExp } from "lodash";
```

This worked well when the file is used in the context of the debugger, but it is also used in a worker. In this case, `from "lodash"` was triggering the usage of an m-c mapping, trying to require lodash from our devtools vendors folder, which of course doesn't work in a worker.

First comment, we probably should not use the mc-mappings when bundling workers, they won't be able to require them anyway. And second comment, since those workers will need to pull and bundle everything we should be careful not to bundle huge libraries with them.

The solution here is to handle the special case of "lodash/subModule" requires in our babel plugin.

juliandescottes:
Hmmmpf I rebased this on release 46 in order to male sure to have a clean baseline for tests, which is making the whole PR unreadable

juliandescottes:
Excluded types from copy modules.

juliandescottes:
Here is a try run with the latest version, this one should hopefully be green:

https://treeherder.mozilla.org/#/jobs?repo=try&revision=c9b09c93e98597b1ce52ff7fbba0e3d2c6e67f04

juliandescottes:
The travis build is failing right now because it can't see the files transferred via copy modules. This is most likely because we need to build again after `copy-modules`

juliandescottes:
All _my_ TODOs are addressed, we have a green try, a green travis. We can move to the review step!

jasonLaster:
This looks great!

2 things:
1. do you think we can land this on master?
2. do you think the mappings file could go in the `config` dir?

jasonLaster:
also - one other thing that might be a nice follow up, but I've been thinking about the directory structure and i think this might be more future forward:

```
/new 
  /dist
     parser-worker.js
     search-worker.js
     vendors.js
  /actions
  /components
  /...
  /test
  index.html
  panel.js
  README.md
```

juliandescottes:
Rebased + moved mozilla-central-mappings.

> the src files are similar to other panels

We actually have `src/` in netmonitor and application panel. While discussing it before creating the application panel, this seemed like the less messy option, although we haven't tried yet to refactor other panels to do the same. So I would rather keep `src` for now here.

> the bundles are in a directory that hopefully will go away over time

This sounds like a good step. I'll try to fit it in here.

bin/copy-modules.js
```diff
@@ -0,0 +1,147 @@
...
+
+  console.log("[copy-modules] creating moz.build files");
+  createMozBuildFiles();
+
+  const projectPath = path.resolve(__dirname, "..");
+  const mcPath = args.mc ? args.mc : feature.getValue("firefox.mcPath");
+  const mcDebuggerPath = path.join(mcPath, "devtools/client/debugger/new");
+
+  console.log("[copy-modules] copying files to: " + mcDebuggerPath);
+  shell.cp("-r", "./out/src", mcDebuggerPath);
```
jasonLaster: it would be nice to `rm ./out` when we're done.

.babel/transform-mc.js
```diff
@@ -0,0 +1,130 @@
...
```
jasonLaster: we need to remove the CSS imports e.g. `import "CallSite.css"`

PR #6221, opened 2018-05-04
[parser] Pause at multi-line returns

jasonLaster:
### Summary of Changes

This is a small tweak which now lets us add breakpoints at return statements when the argument is on another line. This is actually quite common for JSX.


darkwing:
I made a minor adjustment to cut down on code; feel free to remove it if you'd like.

jasonLaster:
hmm, the jest updates seem off...

darkwing:
I was trying to explain that to you in Slack yesterday, must have misworded that

jasonLaster:
yep.. strange

jasonLaster:
the jest issues were the result of adding to the fixture example

PR #6223, opened 2018-05-04
[devtools-config] stop using devtools-config in app

jasonLaster:
supports #6220 

### Summary of Changes

We are only using devtools-config for a couple simple environment flags isDevelopment, isFirefoxPanel, ... These checks can be moved into the debugger and simplified.


jasonLaster:
after doing this, i'm having some second thoughts... if we want other tools to use a launchpad in the future, they will want to use this too. I think the appropriate thing to do is to create a new package `devtools-environment` in devtools-core. 

- this is inline with core being about supporting the launchpad
- this makes devtools-config much smaller and just about launchpad config stuff like what port, favicon we use, and logging the websocket connection

jasonLaster:
https://github.com/devtools-html/devtools-core/pull/1063

jasonLaster:
this is blocked on landing devtools-environment

src/utils/bootstrap.js
```diff
@@ -63,13 +58,16 @@ export function bootstrapStore(client: any, { services, toolboxActions }: any) {
 }
 
 export function bootstrapWorkers() {
+  const workerPath = isDevelopment()
```
juliandescottes: nit: since isDevelopment() = !isFirefoxPanel(), it could be clearer to have only one test.

PR #6226, opened 2018-05-04
Removes selectSpecificLocation

lukaszsobek:
Since selectSpecificLocation is virtually the same as selectLocation and selectSpecificLocation is only accessed by one function it may be a cheap idea to just toggle whether or not we want to check the pretty print with the default set to true.

### DIFF
--- a/src/actions/sources/select.js
+++ b/src/actions/sources/select.js
@@ -86,7 +86,7 @@ export function selectSource(sourceId: string) {
  * @memberof actions/sources
  * @static
  */
-export function selectLocation(location: Location) {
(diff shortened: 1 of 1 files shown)


darkwing:
We added `selectSpecificLocation` as a very preliminary step in simplifying `selectLocation`; unfortunately I think removing it and complicating `selectLocation` with parameters is the opposite direction we want to go.  I'm going to close this, not because it's bad (I like it), but I think we'll go a different direction with `selectLocation`.  Thank you @lukaszsobek !

lukaszsobek:
@darkwing - cool, glad to see it is just a preliminary step. With this pr it was just faster making the pr to show the changes than trying to explain it.

Issue #5663, opened 2018-03-13
[SourceTree] Show original files when the files have the same URL


jasonLaster:
STR:
1. go to https://q7mmol9xy9.codesandbox.io/
2. go to src/Hello.js

Code Sandbox ships each file down with its own base64 sourcemap and no sourceUrl. The result is that the original file has the same url as its generated file equivalent. This means that the source tree at the moment only shows the generated file. cmd+p shows both but doesn't differentiate them.

We have two options:

1. show both and differentiate them e.g ([sm] suffix)
2. show the original file and let the user jump back to the generated file

### Chrome
<img src='https://shipusercontent.com/2aa202c580b77384c3fe7b544c91999d/Screen%20Shot%202018-03-13%20at%2011.21.16%20AM.png' title='Screen Shot 2018-03-13 at 11.21.16 AM.png' width=788>

### Firefox 
![Screen Shot 2018-03-13 at 11.21.08 AM.png](https://shipusercontent.com/dd26dc6cb3278fbcf85540208c4c9e80/Screen%20Shot%202018-03-13%20at%2011.21.08%20AM.png)

joshunger:
@buttersnipps are you working on this issue?

joshunger:
@buttersnipps I'm going to take a look in at this issue later this week can you let me know if you're working on it? If I don't hear from you I'll start working on it.  If you want to ✋ it off let me know.

joshunger:
Mouse over -
![image](https://user-images.githubusercontent.com/2301847/37977856-e3014f32-31a1-11e8-94b1-d14dad76f018.png)

joshunger:
![image](https://user-images.githubusercontent.com/2301847/38197571-21226e42-3646-11e8-84e8-94907bf76055.png)

jasonLaster:
After exploring changing the url to be clearer we decided that it would cause more confusion about how the url came from the source map.

joshunger:
@jasonLaster the React tree components use path as the `key` which becomes a problem because these two files have the same path, but different `id`s. How would you like me to fix?

- https://github.com/devtools-html/debugger.html/blob/master/src/components/PrimaryPanes/SourcesTree.js#L290
- https://github.com/devtools-html/debugger.html/blob/master/src/components/shared/ManagedTree.js#L153

jasonLaster:
i think we do something like this for blackboxing where we add something to the path adding `-original` to the end should work.

joshunger:
```
[...]

PR #6227, opened 2018-05-04
Makes addTab only dependent on url

lukaszsobek:
`addTab` and `ADD_TAB` only use the source url, so in order to clear up some things further down the line it would be really useful to decouple `addTab` from needing the whole source object.

### DIFF
--- a/src/actions/sources/select.js
+++ b/src/actions/sources/select.js
@@ -107,7 +107,7 @@ export function selectLocation(location: Location) {
 
     const source = sourceRecord.toJS();
 
-    dispatch(addTab(source, 0));
+    dispatch(addTab(source.url, 0));
     dispatch(
       ({
         type: "SELECT_SOURCE",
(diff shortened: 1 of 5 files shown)


darkwing:
This is fine for now but I just discovered that we may want to move away from using URLs for tab display due to #5663; it's possible that two sources have the same URL and in `updateTabList` we're comparing URLs :/

jasonLaster:
we might need to use a URL + strategy where it is `url/original|generated/blackboxed`

lukaszsobek:
@jasonLaster sounds like a plan. But even if we decide to do it differently there will probably be 2 properties at max to pass which we could get off of the source before passing it to addTab.

Whatever direction is chosen decoupling addTab (and a few others) from needing full objects will enable us to remove another couple of toJS() calls for performance gains without impeding the flexibility of the application.

PR #6230, opened 2018-05-04
[scopes] Add another original scope line mapping test.

loganfsmyth:
Refs Issue: #5561

### Summary of Changes

* Add one more test to show that we fall back to generated scopes


jasonLaster:
i just saw a false positive travis run, lets be careful before merging

darkwing:
@jasonLaster Let's talk about this Monday; we can read the code super carefully but tests are a huge part of green light; should we run Travis multiple times or do so more locally?

darkwing:
The cited build was green but didn't complete propertly so I've kicked it

jasonLaster:
Hmm, this is showing an early error, which i think we fixed on travis...
screen shot 2018-05-07 at 3 34 01 pm

jasonLaster:
I'm going to gamble and say it is alright... if i'm wrong we can revert

PR #6232, opened 2018-05-05
Makes setSymbols dependent on Id only 

lukaszsobek:
`setSymbols` currently passes the whole source when it only usues the Id. This allows us to get rid of an unneeded toJS() call


jasonLaster:
we should do this everywhere we do `source.toJS()` in the dispatches

lukaszsobek:
@jasonLaster been working on that the last few days - will probably need some pairing or so for the mochitests.

PR #6233, opened 2018-05-05
Removes toJS from setPausePoints

lukaszsobek:
setPausePoints only really uses id and text, so in order to get rid of the toJS it needed some scope changes in arguments passed.


jasonLaster:
i love how you're making small PRs. So much easier to go green and to land

lukaszsobek:
And much easier to review.

PR #6238, opened 2018-05-06
increased expression row height #6166

phoenixking25:
Addresses issue #6166

|Increased height|Original|
|----------|------|
|Increased|original|


src/components/SecondaryPanes/Expressions.css
```diff
@@ -72,6 +72,7 @@
   background-color: var(--theme-body-background);
   display: block;
   position: relative;
+  min-height: 2.8em;
```
darkwing: I think we'll need to do a bit more than this; specifically, I think we should set the same height for both expression items and breakpoint items, so we'll need the height chosen to be in both places.

phoenixking25:
|Increased height|Original|
|----------|------|
|Increased|original|

darkwing:
Hello @phoenixking25 ; I re-read my comment and maybe wasn't specific enough.  I've added a commit along the lines of what I was thinking.

setheight

@violasong The image above sets expressions items, breakpoint items, and the checkbox rows to 2.5em height -- how does this look?

violasong:
Thanks so much for working on this @darkwing and @phoenixking25! 

After seeing this last screenshot, I realized that while the watch expressions do match the breakpoints (yay!)  there's something off about the difference between those two rows and the two Pause on Exceptions rows, which actually look the best in their spacing. I screenshotted the PoE part and pasted it at 50% opacity over the other two parts, and it looks like both breakpoints and watch expressions have 12px more spacing between their rows. If we could tighten that up to match the PoE rows, that would be great!

phoenixking25:
@violasong Sorry, if i am wrong but instead of tighten them up(Watch expressions and breakpoint rows). Can we increase the top margin of PoE part so that it can match the other two?

violasong:
@phoenixking25 Sorry if this has been confusing — my comments are based on my original mockup which has the right breakpoint spacing. https://mozilla.invisionapp.com/share/TNIA9FGHER8#/screens I don't think we should change the PoE spacing as that's the one that looks perfect.  Ideally the others can match it, and they just need 12px less spacing between their rows.

phoenixking25:
Okey, will do that thanks. :)

PR #6244, opened 2018-05-07
add --watch to copy-modules

jasonLaster:
### Summary of Changes

- adds the ability to watch files so that we can incrementally transpile changed files while testing...
- adds a new `yarn watch` command that replaces `copy-assets-watch`

### DIFF
--- a/bin/copy-modules.js
+++ b/bin/copy-modules.js
@@ -8,6 +8,8 @@ const fs = require("fs");
 const path = require("path");
 var shell = require("shelljs");
 const minimist = require("minimist");
+var chokidar = require('chokidar');
+
 
 const feature = require("devtools-config");
 const getConfig = require("./getConfig");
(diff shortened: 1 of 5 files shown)


darkwing:
Best way to test?

jasonLaster:
The best way to test is to 
1. pull it down
2. run `yarn watch`
3. run `yarn mochih dbg-exceptions` and change the test and a file it uses

bin/copy-modules.js
```diff
@@ -44,13 +53,19 @@ function transformSingleFile(filePath) {
   return out.code;
 }
 
+function transpileFile(file) {
+  if (ignoreFile(file)) {
+    return;
+  }
+
+  const filePath = path.join(__dirname, "..", file);
+  const code = transformSingleFile(filePath);
```
juliandescottes: we could inline transformSingleFile in here, no need for two separate methods.

bin/copy-modules.js
```diff
@@ -18,13 +20,20 @@ const envConfig = getConfig();
 feature.setConfig(envConfig);
 
 const args = minimist(process.argv.slice(1), {
```
juliandescottes: Nothing to do with your change but I remember changing the slice(2) (from copy-assets) to slice(1) and forgetting to check the impact. Do you think this should be slice(2)?

jasonLaster: should be fine...

jasonLaster:
If any issues come up we can easily tweak

Issue #6180, opened 2018-04-30
[QuickOpen] Photon UI tweaks


jasonLaster:
We recently landed some improvements for the Quick Open Modal [here](https://github.com/devtools-html/debugger.html/issues/5988). These changes were a big step forward, but we will need to land some tweaks to complete the work

- The blue border is not strong enough
- The cursor padding is too far to the right
- The close button should be removed
- The loading message should be on the right
- The secondary row color should be softer
- The modal shadow should be stronger

jasonLaster:
I believe @aaronmoodie  is on it

jasonLaster:
I started a branch that i believe will hide the quickopen close button [here](https://github.com/devtools-html/debugger.html/compare/master...jasonLaster:qo-close-btn?expand=1)

The trick for jest was running `jest -u` which updated the snapshots

marvinkome:
About the blue border not being strong enough, the border (more like box-shadow) was created by a variable in devtools-mc-assets. It's either it's replaced or edited from the devtools-mc-assets project

jasonLaster:
feel free to use a different style.

PR #6245, opened 2018-05-07
Photon ui tweaks for issue #6180

marvinkome:
Fixes Issue: #6180 

### Summary of Changes

* remove close button
* add the loading message

### Test Plan

- [x] Clicking ctrl (cmd) + shift + 'f'

### DIFF
--- a/src/components/shared/SearchInput.css
+++ b/src/components/shared/SearchInput.css
@@ -102,6 +102,11 @@
   color: var(--theme-body-color-inactive);
 }
 
+.search-field .search-loading {
+  margin-top: 6px;
+  color: #424242bf;
+}
+
 .search-field .search-nav-buttons {
   display: flex;
   user-select: none;
--- a/src/components/shared/SearchInput.js
+++ b/src/components/shared/SearchInput.js
@@ -6,8 +6,6 @@
 
 import React, { Component } from "react";
 
-import CloseButton from "./Button/Close";
-
 import Svg from "./Svg";
 import classnames from "classnames";
 import "./SearchInput.css";
@@ -153,7 +151,6 @@ class SearchInput extends Component<Props, State> {
   render() {
     const {
       expanded,
-      handleClose,
       onChange,
       onKeyDown,
       onKeyUp,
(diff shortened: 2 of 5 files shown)


darkwing:
I rebased this to get the linting issues out (they were already fixed on master) A few improvements we should make before merging:

- [ ] We should make the "Loading text" appear lower; would "align-self: center" work?
- [ ] The loading text should only display while a search is occurring, then hide when the serach is complete

jasonLaster:
I believe @aaronmoodie might be working on this as well...

- The search input is used in other components, so we'll want to hide it conditionally.
- Perhaps "loading" should be passed to the input through some prop like statusMsg, that way other UIs could put different messages there...

darkwing:
Hello again @marvinkome.  Let's pinch the scope of this PR to just removing the close button, since the "Loading..." message is going to be a lot tougher.  I'm thinking we can do something like this:

```js
index 0caabcef..709098be 100644
@@ -397,6 +397,7 @@ export class QuickOpenModal extends Component<Props, State> {
           selectedItemId={
           }
+          hideCloseButton={true}
         />
         {this.renderLoading()}
         {newResults && (
diff --git a/src/components/shared/SearchInput.js b/src/components/shared/SearchInput.js
index 704dd941..b84e891f 100644
--- a/src/components/shared/SearchInput.js
+++ b/src/components/shared/SearchInput.js
@@ -46,7 +46,8 @@ type Props = {
   shouldFocus?: boolean,
   showErrorEmoji: boolean,
   size: string,
-  summaryMsg: string
+  summaryMsg: string,
+  hideCloseButton: boolean
 };
 
 type State = {
@@ -162,7 +163,8 @@ class SearchInput extends Component<Props, State> {
       selectedItemId,
       showErrorEmoji,
       size,
-      summaryMsg
+      summaryMsg,
+      hideCloseButton
     } = this.props;
 
     const inputProps = {
@@ -201,7 +203,9 @@ class SearchInput extends Component<Props, State> {
           <input {...inputProps} />
           {summaryMsg && <div className="summary">{summaryMsg}</div>}
           {this.renderNav()}
-          <CloseButton handleClick={handleClose} buttonClass={size} />
+          {!hideCloseButton && (
+            <CloseButton handleClick={handleClose} buttonClass={size} />
+          )}
         </div>
       </div>
     );
```

aaronmoodie:
@jasonLaster had some changes to remove the close button which are in this PR.
https://github.com/devtools-html/debugger.html/pull/6247

Issue #5443, opened 2018-02-16
[QuickOpen] "Go to line" should have a better hint


nchevobbe:
At the moment, when you hit <kbd>Cmd</kbd>+<kbd>;</kbd> , you are prompted with an input containing a colon, without much more information.

We could have a better hint for the user (a placeholder, a label, …)

wldcordeiro:
This is a little harder just because if we put a `:` in the input the placeholder is cleared, but it should be doable and would be a nice UX enhancement.

marvinkome:
Since it also uses the SearchInput component. I was thinking we add "Go to line" in the place where "Loading ..." or close button will be

PR #6248, opened 2018-05-08
Add 'Go to line' as hint for issue #5443

marvinkome:
Fixes Issue: #5443

### Summary of Changes

* add a hint when `cmd or ctrl` + `;` is pressed

### Test Plan

- [x] Command (or ctrl) + ; opens the go-to line search input
- [x] if it starts with ':' then the goto line hint should show else it shouldn't


darkwing:
Made a bunch of updates, will merge if tests pass.

jasonLaster:
I think we should get the summary message in Quick Open and pass it down. Search Input is supposed to be pretty general purpose and not really know about any features. Basically, SearchInput could in theory move to a lib that others use.

wldcordeiro:
Yep it should be something we can provide from QuickOpen and the input should know as little as possible. 🙂

PR #6261, opened 2018-05-09
Add relativeUrl to Source records; fix getPath in SourcesTree

darkwing:
The next step will be using `getRelativeSources` on the `SourcesTree` component.


jasonLaster:
you can use `SourceRecord` to satisfy flow i think...

src/utils/quick-open.js
```diff
@@ -68,7 +68,7 @@ export function formatSourcesForList(source: RelativeSource, tabs: TabList) {
 
 export type QuickOpenResult = {|
   id: string,
-  value: string,
+  value?: string,
```
jasonLaster: we are filtering for value, so that should be fine

src/utils/quick-open.js
```diff
@@ -125,11 +125,13 @@ export function formatShortcutResults(): Array<QuickOpenResult> {
...
-  return sources
-    .filter(source => !isPretty(source))
-    .map(source => formatSourcesForList(source, tabs))
-    .filter(({ value }) => value != "");
+  return [
+    ...sources
+      .filter(source => !isPretty(source))
+      .map(source => formatSourcesForList(source, tabs))
+      .filter(({ value }) => value != "")
+  ];
```
jasonLaster: why can't we do: 

`sources.filter().map().filter()`

if we need to turn this into a JS list we can  do ``sources.filter().map().filter().toJS()`

darkwing:
I'm going to rip out a few updates to make this PR as small as possible; will push in a moment

darkwing:
OK, I drastically simplified this PR

src/types.js
```diff
@@ -299,6 +299,11 @@ export type Source = {
   +loadedState: "unloaded" | "loading" | "loaded"
 };
 
+export type RelativeSourceRecord = RecordOf<RelativeSource>;
```
AnshulMalik: Did you check if `RecordOf` works? I had hard time with this.

src/selectors/getRelativeSources.js
```diff
@@ -22,22 +19,24 @@ function getRelativeUrl(url, root) {
   return url.slice(url.indexOf(root) + root.length + 1);
 }
 
-function formatSource(source: Source, root): RelativeSource {
-  return {
-    ...source,
+function formatSource(source: SourceRecord, root): RelativeSource {
+  // return source.set("relativeUrl", getRelativeUrl(source.url, root));
+  return new RelativeSourceRecordClass({
+    ...source.toJS(),
```
AnshulMalik: `toJS()` is required?

PR #6262, opened 2018-05-10
Stop using Services.telemetry directly. Fixes #6258

### LINKED ISSUE #6258: Debugger should not use Services.telemetry directly
All Services.telemetry calls should be calling:
https://searchfox.org/mozilla-central/source/devtools/client/shared/telemetry.js

Of course, this needs a stub when running outside the browser.

MikeRatcliffe:
This patch adds a telemetry stub and points it at the real `telemetry.js` for code that lives on Mozilla Central.

We also now use `telemetry.start()` and `telemetry.finish()` for the timing inside `src/actions/sources/loadSourceText.js` instead of calculating the duration ourselves.

### DIFF
--- a/src/actions/sources/loadSourceText.js
+++ b/src/actions/sources/loadSourceText.js
@@ -9,6 +9,7 @@ import { PROMISE } from "../utils/middleware/promise";
 import { getGeneratedSource, getSourceFromId } from "../../selectors";
 import * as parser from "../../workers/parser";
 import { isLoaded } from "../../utils/source";
+import { Telemetry } from "devtools-modules";
 
 import defer from "../../utils/defer";
 import type { ThunkArgs } from "../types";
@@ -16,10 +17,9 @@ import type { ThunkArgs } from "../types";
 import type { Source } from "../../types";
 
 const requests = new Map();
-import { Services } from "devtools-modules";
-const loadSourceHistogram = Services.telemetry.getHistogramById(
-  "DEVTOOLS_DEBUGGER_LOAD_SOURCE_MS"
-);
+
+const loadSourceHistogram = "DEVTOOLS_DEBUGGER_LOAD_SOURCE_MS";
+const telemetry = new Telemetry();
 
 async function loadSource(source: Source, { sourceMaps, client }) {
   const id = source.id;
@@ -51,16 +51,19 @@ export function loadSourceText(source: Source) {
       return Promise.resolve();
     }
 
-    const telemetryStart = performance.now();
     const deferred = defer();
     requests.set(id, deferred.promise);
 
+    // The file is not loaded so we need to load it. This telemetry probe times
+    // the loading of the source file.
+    telemetry.start(loadSourceHistogram, source);
     try {
       await dispatch({
         type: "LOAD_SOURCE_TEXT",
         sourceId: id,
         [PROMISE]: loadSource(source, { sourceMaps, client })
       });
+      telemetry.finish(loadSourceHistogram, source);
     } catch (e) {
       deferred.resolve();
       requests.delete(id);
@@ -82,9 +85,6 @@ export function loadSourceText(source: Source) {
     deferred.resolve();
     requests.delete(id);
 
-    const telemetryEnd = performance.now();
-    const duration = telemetryEnd - telemetryStart;
-    loadSourceHistogram.add(duration);
     return source;
   };
 }
--- a/src/actions/sources/loadSourceText.js
+++ b/src/actions/sources/loadSourceText.js
@@ -18,6 +18,7 @@ import type { Source } from "../../types";
 
 const requests = new Map();
 
+// Measures the time it takes for a source to load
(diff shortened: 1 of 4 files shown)


src/utils/telemetry.js
```diff
@@ -0,0 +1,492 @@
...
+   *
+   * @param {dataset}
+   *        DATASET_RELEASE_CHANNEL_OPTOUT or DATASET_RELEASE_CHANNEL_OPTIN.
+   * @param {boolean} [clear = false]
+   *        Whether to clear out the scalars after snapshotting.
+   */
+  snapshotKeyedScalars(dataset, clear) {}
+}
+
+export { Telemetry };
```
jasonLaster: can we add this ship to devtools-modules

https://github.com/devtools-html/devtools-core/tree/master/packages/devtools-modules/src

the benefits are:

1. any tools using the launchpad can use the shim
2. all the shims are in one place
3. the shims are easier to test

MikeRatcliffe: Yup, that would make a lot of sense.

MikeRatcliffe:
Obviously, this pull request depends on https://github.com/devtools-html/devtools-core/pull/1065 and a devtools-modules version bump.

jasonLaster:
I just pushed some tweaks for using telemetry from d-m

darkwing:
Hello @MikeRatcliffe !  Is this ready or does it require more?  I'm not familiar with our telemetry usage so any more detail you can provide would be great.  Thanks!

MikeRatcliffe:
@darkwing It should be good to go... to be honest I am not sure why @jasonLaster didn't land it.

jasonLaster:
oh, didn't realize that it was ready...

jasonLaster:
@MikeRatcliffe seeing some ci failures: https://circleci.com/gh/devtools-html/debugger.html/13179?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

src/actions/sources/loadSourceText.js
```diff
@@ -9,16 +9,17 @@ import { PROMISE } from "../utils/middleware/promise";
...
 
 const requests = new Map();
-import { Services } from "devtools-modules";
-const loadSourceHistogram = Services.telemetry.getHistogramById(
-  "DEVTOOLS_DEBUGGER_LOAD_SOURCE_MS"
-);
+
+const loadSourceHistogram = "DEVTOOLS_DEBUGGER_LOAD_SOURCE_MS";
+const telemetry = new Telemetry();
+telemetry.start(loadSourceHistogram);
```
jasonLaster: hmm, what is this trying to measure?

it might be capturing the time from module load to the completed load source...

also, it'd be good to be careful that you don't deal w/ race conditions...

like three sources loading and the 2nd finishing first

MikeRatcliffe: The probes are in the same place as they previously were but starting on module load seems wrong to me as well... I will try to find a better place.

src/actions/sources/loadSourceText.js
```diff
@@ -63,6 +65,7 @@ export function loadSourceText(source: SourceRecord) {
           [PROMISE]: loadSource(source, { sourceMaps, client })
         }: Action)
       );
+      telemetry.finish(loadSourceHistogram, source);
```
jasonLaster: I'm not sure about the telemetry api, does this support potential race conditions with timigh

```js    
telemetry.start(loadSourceHistogram, {id: 1});
telemetry.start(loadSourceHistogram, {id: 2});
telemetry.finish(loadSourceHistogram, {id: 2});
telemetry.finish(loadSourceHistogram, {id: 1});
```

also, do we want to handle timing when the load fails? Should we note that in the probe?

jasonLaster: it might be nice to test these cases in `actions/sources/tests`

MikeRatcliffe: As far as race conditions are concerned we are covered because we pass in source as an object and the timing event is then keyed on that source object:
`telemetry.finish(loadSourceHistogram, source);`

It is up to you whether you want to handle timing when the load fails... we could always move it to a finally block.

src/actions/sources/loadSourceText.js
```diff
@@ -51,10 +51,12 @@ export function loadSourceText(source: SourceRecord) {
       return Promise.resolve();
     }
 
-    const telemetryStart = performance.now();
     const deferred = defer();
     requests.set(id, deferred.promise);
 
+    // The file is not loaded so we need to load it. This telemetry probe times
+    // the loading of the source file.
+    telemetry.start(loadSourceHistogram, source);
```
jasonLaster: I don't think we need the comment here, it kinda describes the action. Perhaps we can add a description to the jsdoc of the action

src/actions/sources/loadSourceText.js
jasonLaster: the finally block might be a good idea so that we always measure something.

Issue #6193, opened 2018-05-02
[Preview] use new token events


jasonLaster:
The current preview mouse events have several flaws:

- mouseover fires many times for the same token
- mouseover fires for invalid tokens
- there is no notion of on token leave

Here is a [branch](https://github.com/devtools-html/debugger.html/compare/master...jasonLaster:token-over?expand=1) i started with bomsy and hao

and here is a diagram of the token lifecycle

<img src='https://shipusercontent.com/e958e111c4d041e8f80a744e97dffa3a/token%20leave.jpg' title='token leave.jpg' width=1135>

PR #6263, opened 2018-05-10
[Preview] Switch to token-over

haoRchen:
Fixes Issue: #6193 

### Summary of Changes

Goals:

1. Display/hide preview via OnToken events
2. A preview should be cleared via OnPreviewLeave when hovering off the popup.

### DIFF
--- a/src/actions/preview.js
+++ b/src/actions/preview.js
@@ -28,79 +28,37 @@ import type { Action, ThunkArgs } from "./types";
 import type { ColumnPosition } from "../types";
 import type { AstLocation } from "../workers/parser";
 
-function isInvalidTarget(target: HTMLElement) {
-  if (!target || !target.innerText) {
-    return true;
+function findExpressionMatch(state, codeMirror, tokenPos) {
+  const source = getSelectedSource(state);
+  const symbols = getSymbols(state, source);
+
+  let match;
+  if (!symbols || symbols.loading) {
+    match = getExpressionFromCoords(codeMirror, tokenPos);
+  } else {
+    match = findBestMatchExpression(symbols, tokenPos);
   }
-
-  const tokenText = target.innerText.trim();
-  const cursorPos = target.getBoundingClientRect();
-
-  // exclude literal tokens where it does not make sense to show a preview
-  const invalidType = ["cm-atom", ""].includes(target.className);
-
-  // exclude syntax where the expression would be a syntax error
-  const invalidToken =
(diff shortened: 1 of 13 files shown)


jasonLaster:
Thanks @haoRchen. This will be great to get in

darkwing:
Hello @haoRchen!  Is this still a WIP or do you have more coming?  Thank you!

jasonLaster:
I believe he has more coming

darkwing:
Any update @haoRchen ?  I'd love to help!

haoRchen:
Hey @darkwing. Currently, the preview displays onTokenEnter but does not clear. Something is causing mouseOver to be called twice and somehow skips the call for TokenLeave(which clears the preview). Could you help me look into whats causing this?

jasonLaster:
Alright, I looked into this and confirmed what you were seeing @haoRchen. CodeMirror is in fact replacing the ENTIRE line everytime we update a marker, thus no mouseleave events...

by the way, i used chromes break on dom event here...

jasonLaster:
I took a look at this today and found some interesting things:

1. token events could ignore entering editor-mount
2. we could stop using markText
3. we could simplify our rules for entering and leaving

ultimately, i think this is pretty close to being perfect.

I left logging in to help w/ the debugging

jasonLaster:
Alright, I'm really excited to share this work. @haoRchen and I have been working on a new system for Preview events that I believe will be much more robust:


It starts with *tokenEvents*, where we introduce two new codeMirror events `tokenEnter` and `tokenLeave`. These events have really helpful properties: 

- tokenEnter is triggered once per token
- tokenEnter is only fired for tokens that we could preview or jump to. Not whitespace, syntax, ...
- tokenLeave is triggered immediately after the mouse leaves the element

*tokenEvents* simplifies the problem and makes everything below possible.

Preview component:
- `tokenEnter` - kick off a preview 
- `tokenLeave` - clear the preview, unless we're hovering on the popup 
- `componentDidUpdate` - updates the highlight
- `render` - renders the popup if the Preview is ready...

Popup Component:
- `onMouseLeave` - clears the preview, unless we're hovering on the highlight...

Redux:

- `updatePreview` is now much simpler because it doesn't need to check if the token is valid or if we're hovering on the same element.
- `setPreview` used to have a race conflict when you cleared before it finished, we now check that the preview is updating before we set the final value

haoRchen:
looks good!

jasonLaster:
- i removed the shared helper because the two were different
- cleaned up circle ci
- fixed the travis tests

Issue #5875, opened 2018-04-06
[Breakpoints] make Breakpoint Pure


jasonLaster:
I think we can likely remove the `shouldUpdate` call by doing a pretty extreme refactoring [branch][bp]. The reason i think we need `componentShouldUpdate` is that every time Breakpoints renders we create new local breakpoints, so Breakpoint feels like it should re-render. We can change the connect function to be more pure, basically give us all the data unchanged, and then on the component side do what we need. 

To clarify, this is in `components/SecondaryPane/Breakpoint.js`

#### componentShouldUpdate
```diff
-  shouldComponentUpdate(nextProps: Props) {
-    const prevBreakpoint = this.props.breakpoint;
-    const nextBreakpoint = nextProps.breakpoint;
-
-    return (
-      !prevBreakpoint ||
-      (prevBreakpoint.text != nextBreakpoint.text ||
-        prevBreakpoint.disabled != nextBreakpoint.disabled ||
-        prevBreakpoint.condition != nextBreakpoint.condition ||
-        prevBreakpoint.hidden != nextBreakpoint.hidden ||
-        prevBreakpoint.isCurrentlyPaused != nextBreakpoint.isCurrentlyPaused)
-    );
-  }
-
```

#### PureComponent
```diff
-class Breakpoint extends Component<Props> {
+class Breakpoint extends PureComponent<Props> {
```

[bp]: https://github.com/jasonLaster/debugger.html/tree/breakpoint-highlight2

PR #6265, opened 2018-05-10
Add mappedBreakpoints

jasonLaster:
### Summary of Changes

* cleans up breakpoint location logic
* will now select the exact location

### DIFF
--- a/src/components/SecondaryPanes/Breakpoint.js
+++ b/src/components/SecondaryPanes/Breakpoint.js
@@ -10,10 +10,9 @@ import classnames from "classnames";
 
 import { CloseButton } from "../shared/Button";
 
-import { createEditor } from "../../utils/breakpoint";
+import { createEditor, makeLocationId } from "../../utils/breakpoint";
 import { features } from "../../utils/prefs";
 import { isInterrupted } from "../../utils/pause";
-import { makeLocationId } from "../../utils/breakpoint";
 
 import type { LocalBreakpoint } from "./Breakpoints";
 import type SourceEditor from "../../utils/editor/source-editor";
@@ -159,10 +158,10 @@ class Breakpoint extends Component<Props> {
   }
 
   renderLineClose() {
-    const { breakpoint, onCloseClick, selectedSource } = this.props;
+    const { breakpoint, onCloseClick } = this.props;
     const { location } = breakpoint;
 
-    let { line, column } = location;
+    const { line, column } = location;
 
     return (
       <div className="breakpoint-line-close">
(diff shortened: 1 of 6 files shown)


darkwing:
This makes me happy:

paused

jasonLaster:
Yes!

darkwing:
@jasonLaster I made a bunch of changes here so you may want to take another look at the general direction of this, flow issues aside

jasonLaster:
sounds good

src/components/SecondaryPanes/Breakpoint.js
```diff
@@ -38,21 +40,22 @@ function getBreakpointLocation(source, line, column) {
...
+  why: Why
+) {
+  if (!frame || isInterrupted(why)) {
+    return false;
   }
 
-  return originalText;
+  const bpId = getLocationWithoutColumn(breakpoint.location);
+  const pausedId = getLocationWithoutColumn(frame.location);
+  return bpId === pausedId;
```
jasonLaster: maybe we just do the comparisons here

```js
bp.location.sourceId == frame.locaton.sourceId ...
```

darkwing:
Weirdest way to find a bug ever but STR:

1.  Make a few breakpoints, refresh source document, wait for breakpoint to hit
2.  Close all tabs in the tab bar
Weirdest way to find a bug ever but STR:

1.  Make a few breakpoints, refresh source document, wait for breakpoint to hit
2.  Close all tabs in the tab bar

mappedbreakpointsbug

What's the best course of action if we suddenly don't have a `selectedSource`?  Returning or returning null?

darkwing:
The error cited above is also persistent; i.e. if I try to refresh the debugger page the same error appears, so it seems we need to be super careful with storing breakpoint data.

jasonLaster:
yep, returning early is good

src/selectors/mappedBreakpoints.js
```diff
@@ -0,0 +1,57 @@
...
+    location = breakpoint.generatedLocation || breakpoint.location;
+    text = breakpoint.text;
+  }
+  const localBP = { locationId, location, text, source, condition };
+
+  return localBP;
+}
+
+function _getMappedBreakpoints(breakpoints, sources, selectedSource) {
+  return breakpoints
```
jasonLaster: maybe check for the selectedSource here so that we can be extra cautious

darkwing: Returning a `I.Map()` seems like the best idea, IMO, because it can prevent breakage elsewhere.

darkwing:
I'm also seeing this periodically; trying to figure out if it's an existing issue or new to this PR:

mapscopeserror

jasonLaster:
it would be great to do this https://github.com/devtools-html/debugger.html/issues/5875 after this lands because then we wont need shouldComponentUpdate

Issue #6255, opened 2018-05-08
Search failure doesn't update when hitting Enter to search


hcs64:
Steps to reproduce:
1. Open files in tabs in the debugger
2. Ctrl+F search for something that is in one file but not the current tab (frowny appears and search term turns red)
3. Switch to the tab with the file that contains the search term
4. Click in the search box and hit Enter to re-run the search

Even though the search term is found, the frowny and red text remain.

jasonLaster:
Thanks for the report!

lukaszsobek:
Merged. Closing.

PR #6270, opened 2018-05-11
[Search] Makes SearchBar update results on enter press

lukaszsobek:
Fixes Issue: #6255

Changes:
makes SearchBar recalculate results

Cause:
Pressing enter after a tab switch relied on the old state and did not update the result count. This lead to a series of undesired outcomes including bad error emoji state and stale count of found items.

Vid after change:
update_search

### DIFF
--- a/src/components/Editor/SearchBar.js
+++ b/src/components/Editor/SearchBar.js
@@ -211,13 +211,14 @@ class SearchBar extends Component<Props, State> {
     this.setState({ inputFocused: false });
   };
 
-  onKeyDown = (e: SyntheticKeyboardEvent<HTMLElement>) => {
+  onKeyDown = (e: any) => {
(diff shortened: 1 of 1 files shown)


src/components/Editor/SearchBar.js
```diff
@@ -218,6 +218,7 @@ class SearchBar extends Component<Props, State> {
 
     this.traverseResults(e, e.shiftKey);
     e.preventDefault();
+    return this.doSearch(e.target.value);
```
wldcordeiro: I love when it's just  a 1 line fix 😆

wldcordeiro:
Approve the code but CI failures. 😢

src/components/Editor/SearchBar.js
lukaszsobek: Yeah, an an hour searching for it XD

PR #6275, opened 2018-05-11
Fix Photon styling for search inputs

darkwing:
When we (I) merged the initial QuickOpen style updates for Photon, it became apparent the new CSS was too broad and that elements were enlarged that shouldn't have been -- just QuickOpen should have been increased in size.  This has been bugging me so I've taken the time to fix it.

Command+F (Find in File) and Command+G (Search Again)

Command+P (Quick Open)

Command + Shift + F (Full Project Search)

Command + Shift + O (Function Search)

Command + ; (Go to line)

### DIFF
--- a/src/components/QuickOpenModal.js
+++ b/src/components/QuickOpenModal.js
@@ -398,6 +398,7 @@ export class QuickOpenModal extends Component<Props, State> {
           selectedItemId={
             expanded && items[selectedIndex] ? items[selectedIndex].id : ""
           }
+          {...(this.isSourceSearch() ? { size: "big" } : {})}
         />
         {this.renderLoading()}
         {newResults && (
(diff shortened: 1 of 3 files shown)


src/components/QuickOpenModal.js
```diff
@@ -398,6 +398,7 @@ export class QuickOpenModal extends Component<Props, State> {
           selectedItemId={
             expanded && items[selectedIndex] ? items[selectedIndex].id : ""
           }
+          {...(this.isSourceSearch() ? { size: "big" } : {})}
```
wldcordeiro: I prefer the big here all the time as they're all one modal. But I agree with the change in project search.

jasonLaster:
Hmm, i also would prefer one input height for all of quick open... Lets see with @violasong thinks

jasonLaster:
On second thought, lets land this and then we can evaluate how we want to support other types of QO searches

wldcordeiro:
It's a one line change so that sounds like a good plan.

Issue #6251, opened 2018-05-08
Startpage's `Ctrl+P` and `Ctrl+Shift+F` should be <button/>s


jnachtigall:
From the mouse cursor and visually it is not clear that the 'buttons' are clickable. Looks like these are divs/spans with event listeners. Would be better if these where real `<button>`'s, also for semantics/accessibility. Like it is for the Performance or Memory tab:

![debugger-buttons](https://user-images.githubusercontent.com/3307404/39756222-02b74e60-52c9-11e8-9f5c-ea1d32e9889e.gif)

Probably it would be easy to fix by:
1. Changing the corresponding source code html from `div/span` to `<button>`
2. Apply the same CSS to it like it is for the buttons on Memory / Performance tab
3. Optionally: Probably nicer to write the shortcuts with `+` as it is in other shortcuts, at least in Windows/Linux (not sure about Mac). That is, `Ctrl+P` and `Ctrl+Shift+F` instead of `Ctrl P` and `Ctrl Shift F`. All other Firefox shortcuts like `Ctrl+C` or `Ctrl+F` are written with `+` (just click on the top-right hamburger menu to check)

jnachtigall:
@phoenixking25 almost perfect, thank you! :) (though I'm not a maintainer, so wait for them for a proper code review).

What I'm seeing that still could be improved:
1. Both bothes should have the same width so they align more nicely. (The find files button is a bit wider at the moment). Not sure how this would be best done. Setting a `min-width` on both probably does not work this the strings also get translated and depending on the language the length might vary/change.
2. Well, still I think it should be That is, `Ctrl+P` and `Ctrl+Shift+F` instead of `Ctrl P` and `Ctrl Shift F` for consistency. At least on Windows and Linux there's always a `+` for shortcuts, in Firefox and in all other languages.

phoenixking25:
![min-width](https://user-images.githubusercontent.com/22527880/39783886-2c21b496-5334-11e8-81d9-86055785d933.gif)
@jnachtigall Thanks for the help :) Setting `min-width: 250px` did the thing. And the `+` sign will work on that :)

jnachtigall:
> Setting min-width: 250px did the thing.

[...]

PR #6285, opened 2018-05-11
changed div into button

phoenixking25:
pull
#6251 converted div/span into buttons


jasonLaster:
Thanks @phoenixking25 can you update the css so that it doesnt change aesthetically

phoenixking25:
@jasonLaster you mean to remove on hover effect?

jasonLaster:
and the button styling

phoenixking25:
updated
@jasonLaster is it looking fine now?

jasonLaster:
hmm, lets also get rid of the background hover. I think its somewhat distracting

phoenixking25:
The hover effect is set on button element for all. If i remove that then wouldnt be it same as before?

jasonLaster:
it would be more accessible. think tabs

phoenixking25:
@jasonLaster Done :+1:

jasonLaster:
small jest issue, youll need to run `jest -u`

mind sharing another gif?

phoenixking25:
tabbed
@jasonLaster Now can access with tabs and removed background hover

AnshulMalik:
Can we get rid of the text outline and keep the button outline?

jasonLaster:
@phoenixking25 to might be nice to undlerine the text on hover and change the cursor to a pointer

AnshulMalik:
And also the cursor should be changed to pointer @phoenixking25

phoenixking25:
screenshot from 2018-05-14 20-57-27
@jasonLaster changed the cursor and underlined it!!

darkwing:
Hello @phoenixking25 -- great work so far.  I took a look, threw a few ideas around with colleagues, and what we came up with two options:

1.  If we switch to `<button>`, we should style the button to look just as the `<div>s` look now; same color, font-size, and font family

2.  If we stay as a `<div>`, add `role=button` (https://www.w3.org/TR/2016/WD-wai-aria-practices-1.1-20160317/examples/button/button.html), set a `tabindex=0` (we'll see how this effects other stuff), add a pointer cursor, and make the text a bit darker upon hover.

I like option 2 the best but have fun and try a few things!  Thank you so much for your help! 👍

phoenixking25:
div
@darkwing done as you said, used `<p>`  tags instead of `<div>` because using div was changing  the alignment and applied onKeyPress, made the text bit bolder but in case of commands like `Ctrl P` we are using some font-family so it is not getting bolder. ignore the cursor its my gif making tool fault, actually its pointer cursor.

darkwing:
Hello again @phoenixking25 ; this seemingly small change (not your code, just the design change in general) seems to have kicked off a bunch of different opinions 😸  and our design person want to look at it a bit more.  I'm closing this PR until they make a decision but I wanted to let you know this is really awesome work and we'll likely re-open and use it soon.  In the mean time, please stick with us and pick another issue to fix -- you do great work! 👍

PR #6287, opened 2018-05-11
Cleans up shared buttons

lukaszsobek:
Changes:
- renames files to contain the word "button"
- creates single export point for buttons
- moves css to `/styles/` folder
- adds more tests
- moves tests to Button folder
- makes tests more readable by using `.hasClass` instead `.toMatchSnapshot` where possible.


src/components/Editor/Footer.js
```diff
@@ -18,7 +18,7 @@ import { isPretty, isLoaded, getFilename } from "../../utils/source";
 import { getGeneratedSource } from "../../reducers/sources";
 import { shouldShowFooter, shouldShowPrettyPrint } from "../../utils/editor";
 
-import PaneToggleButton from "../shared/Button/PaneToggle";
+import { PaneToggleButton } from "../shared/Button";
```
lukaszsobek: Really tempted to go with "Buttons", since the folder contains a number bigger than one.

jasonLaster:
so, i'm a bit worried about this because Buttons that relate to their container, PaneToggle, CommandBar, should be near their container...

jasonLaster:
I like all the testing and clean up, so tough one. I'd like to see what @wldcordeiro and @darkwing think

lukaszsobek:
@jasonLaster not quite sure I understand. The buttons are where they were before, the only thing that has changed is a unified export endpoint to access the buttons from `/shared/buttons` instead of grabbing each individually through their files, moving the css to a separate folder since we are not interested in the css most of the time and moving the test inside the folder with the components being tested.

jasonLaster:
i see. lets rebase and land it. dont worry about travis, those are known issues that are unrelated.

src/components/shared/Button/PaneToggleButton.js
```diff
@@ -6,8 +6,8 @@
 import React, { PureComponent } from "react";
 import classnames from "classnames";
 import Svg from "../Svg";
-import CommandBarButton from "./CommandBarButton";
-import "./PaneToggle.css";
+import { CommandBarButton } from "./";
+import "./styles/PaneToggleButton.css";
```
jasonLaster: You know, `PaneToggleButton` is cool, but itd be even cooler to do `PaneToggle`

lukaszsobek: Since the only thing this component returns is a button it feels more descriptive to let the person reading know by adding the word Button to the name of the files even though the component already is in a button directory. We might also have a pane toggle reducer and other things that share the same name. By keeping the name button we have all the info we need about the files in the filename accessible at once instead of having to refer to the folder name for additional input as to what PaneToggle might be. 

Your call. I'd stick with PaneToggleButton, but will change it to PaneToggle if you think that fits better.

PR #6272, opened 2018-05-11
✅: Migrate linting to use jest-runner-eslint

wldcordeiro:
Fixes Issue: N/A

### Summary of Changes

* Changing JS linting strategy to incorporate Jest Runners

### Test Plan

Tell us a little a bit about how you tested your patch.

Example test plan:

- Run `yarn test`
- Run `yarn lint`
- Runt `yarn jest`

### Screenshots/Videos (OPTIONAL)

screen shot 2018-05-11 at 9 33 56 am
screen shot 2018-05-11 at 9 33 37 am

PR #6291, opened 2018-05-11
⬆️: Upgrading Jest to 22

wldcordeiro:
Fixes Issue: #<issue number>

### Summary of Changes

* This is a cherry-pick of the bulk of #6272 to break up the work that we want for sure (updating
Jest) vs something we want to discuss further (using Jest as a platform to handle linting for JS and
CSS via jest-runner-eslint and jest-runner-stylelint)


package.json
```diff
@@ -126,17 +126,18 @@
...
     "eslint-plugin-react": "^7.2.1",
     "expect.js": "^0.3.1",
     "flow-bin": "0.67.1",
     "glob": "^7.0.3",
     "husky": "^0.14.2",
-    "jest": "^20.0.4",
+    "jest": "^22.4.3",
+    "jest-environment-jsdom": "^22.4.3",
     "jest-in-case": "^1.0.2",
-    "jest-junit-reporter": "^1.1.0",
```
wldcordeiro: This reporter has gone over a year without an update and officially only supports Jest up to 18 (it runs our configs through validation so if we use any new config stuff it fails automatically.)

packages/devtools-reps/package.json
```diff
@@ -8,10 +8,7 @@
     "lint-js": "eslint src",
     "firefox": "./node_modules/.bin/start-firefox --start --location https://devtools-html.github.io/debugger-examples/",
     "chrome": "./node_modules/.bin/start-chrome",
-    "copy-assets": "node bin/copy-assets",
```
wldcordeiro: The three scripts do not exist here.

PR #6293, opened 2018-05-12
✨: Utilize Jest Runners for ESlint and Stylelint

wldcordeiro:
Fixes Issue: #<issue number>

### Summary of Changes

* Introduce `jest-runner-eslint` and `jest-runner-stylelint` in place of normal runners.

This is essentially a reopening of #6272 but split up from the work updating Jest.

@jasonLaster suggested doing this and opening up for discussion.

A good starting point is this [Jest as a Platform](https://www.youtube.com/watch?v=NtjyeojAOBs) talk  but I've bought in and love the idea of Jest as a platform because of the number of possibilities it opens up and the best in class tooling that Jest has. Jest is an awesome platform and having runners for linting makes for a nicer experience in terms of consistency but also pure developer experience as you can now have `yarn jest --watch` running and it'll lint your recently changed code and test it concurrently.

I'd love more discussion and feedback here. 😄 

### Basic Screenshot

screen shot 2018-05-11 at 10 37 14 am

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,4 @@ tags
 webpack-stats/
 package-lock.json
 out
+junit.xml
--- a/circle.yml
+++ b/circle.yml
@@ -22,12 +22,10 @@ test:
   override:
     - mkdir -p $CIRCLE_TEST_REPORTS/flow-coverage
     - mkdir -p $CIRCLE_TEST_REPORTS/jest-coverage
-    - yarn jest -i --no-colors --testResultsProcessor="jest-junit" --coverage --coverageDirectory=$CIRCLE_TEST_REPORTS/jest-coverage
-    - cd ./packages/devtools-source-map; yarn test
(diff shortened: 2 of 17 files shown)


jasonLaster:
Thanks @wldcordeiro i like that we could move to one system. It also could be faster. My concern is that the output for a passing lint run is now quite verbose, where as before it was just "no errors".

I think the terminal output can be distracting and perhaps intimidating for new comers.

wldcordeiro:
Hey peeps my internet has been down today. I was looking at Jest's CLI options for something like a `--simple` but there doesn't appear to be one. There is an issue open for it.

That said I really like the full output myself and it's one of my favorite parts of Jest. Just so satisfying watching all the reds go to green.

Jest watch mode feels like magic with the runners 🙂

darkwing:
We'll give this a shot and revert if it turns out to be a problem; thank you @wldcordeiro !!

PR #6298, opened 2018-05-12
[src/components] increase test coverage

mutdmour:
Fixes Issue: #5100 

Here's the Pull Request Doc

### Summary of Changes

* ShortcutsModal
* QuickOpenModal
* Welcomebox

### Test Plan
Feel free to improve it!

screen shot 2018-05-12 at 7 02 23 pm


src/components/QuickOpenModal.js
```diff
@@ -293,10 +293,6 @@ export class QuickOpenModal extends Component<Props, State> {
       }
 
       if (results) {
-        if (this.isShortcutQuery()) {
```
mutdmour: this is same function is already called in `this.selectResultItem` below. just redundant

jasonLaster:
Looks great!

jasonLaster:
@codehag will look at this tomorrow. Thanks for your patience

src/components/test/ShortcutsModal.spec.js
```diff
@@ -6,15 +6,34 @@ import React from "react";
...
+      enabled: false
+    });
+    expect(component.text()).toBe("");
+  });
+
+  it("renders with additional classname", () => {
+    const { component } = render({
+      additionalClass: "additional-class"
+    });
+    expect(component).toMatchSnapshot();
```
lukaszsobek: it would be great if we could just use `.hasClass` here, as it is the thing we are testing for. A snapshot seems to make the test more difficult to read and brittle.

src/components/test/ShortcutsModal.spec.js
```diff
@@ -6,15 +6,34 @@ import React from "react";
 import { shallow } from "enzyme";
 import { ShortcutsModal } from "../ShortcutsModal";
 
+function render(overrides = {}) {
```
lukaszsobek: Love this rewrite, so sophisticated :)

src/components/test/ShortcutsModal.spec.js
wldcordeiro: Heh, I agree generally when you use normal snapshots because of the size produced but I'm gonna shamelessly plug my tool [`snapshot-diff-serializer`](https://github.com/wldcordeiro/snapshot-diff-serializer) as this is an awesome place for this kind of snapshot.

```js
const { component: defaultComp } = render()
const { component } = render({ additionalClass: "additional-class" })
expect({
  diffA: defaultComp,
  diffB: component
}).toMatchSnapshot();
```

src/components/test/QuickOpenModal.spec.js
```diff
@@ -210,6 +287,542 @@ describe("QuickOpenModal", () => {
...
+        }
+      };
+      wrapper.setState(() => ({
+        results: [{ id: "0", location }, { id: "1" }, { id: "2" }],
+        selectedIndex: 1
+      }));
+      wrapper.find("SearchInput").simulate("keydown", event);
+      expect(event.preventDefault).toHaveBeenCalled();
+      expect(wrapper.state().selectedIndex).toEqual(2);
+      expect(props.selectLocation).toHaveBeenCalledWith({
```
codehag: these results are bit confusing, why are we at line 0? is it taking us to the top of the file on `arrowDown` within the variables search?

src/components/test/QuickOpenModal.spec.js
```diff
@@ -210,6 +287,542 @@ describe("QuickOpenModal", () => {
...
+        preventDefault: jest.fn(),
+        key: "ArrowDown"
+      };
+      const location = {
+        start: {
+          line: 7
+        }
+      };
+      wrapper.setState(() => ({
+        results: [{ id: "0", location }, { id: "1" }, { id: "2" }],
```
codehag: are these distinct for variables somehow? if not do we need to distinguish between these two?

src/components/test/ShortcutsModal.spec.js
mutdmour: fair. `hasSnapshot` is just the lazy option

src/components/test/ShortcutsModal.spec.js
mutdmour: not going to claim this. Learned this pattern from other files from this project.

src/components/test/QuickOpenModal.spec.js
```diff
@@ -210,6 +287,542 @@ describe("QuickOpenModal", () => {
...
+      const { wrapper, props } = generateModal(
+        {
+          enabled: true,
+          query: "@test",
+          searchType: "shortcuts"
+        },
+        "shallow"
+      );
+      wrapper.setState(() => ({
+        results: [{}, { id: ":" }],
```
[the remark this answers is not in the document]
mutdmour: 💩

src/components/test/QuickOpenModal.spec.js
```diff
@@ -210,6 +287,542 @@ describe("QuickOpenModal", () => {
...
+        selectedIndex: 0
+      }));
+      const event = {
+        key: "Enter"
+      };
+      wrapper.find("SearchInput").simulate("keydown", event);
+      expect(props.setQuickOpenQuery).toHaveBeenCalledWith("@");
+    });
+
+    it("on Enter with empty results, handle no item", () => {
```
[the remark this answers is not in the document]
mutdmour: did the iterative

src/components/test/QuickOpenModal.spec.js
```diff
@@ -210,6 +287,542 @@ describe("QuickOpenModal", () => {
...
+      };
+      wrapper.find("SearchInput").simulate("keydown", event);
+      expect(props.setQuickOpenQuery).toHaveBeenCalledWith(":");
+    });
+
+    it("on Enter with results, handle # shortcut", () => {
+      const { wrapper, props } = generateModal(
+        {
+          enabled: true,
+          query: "@test",
```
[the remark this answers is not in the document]
mutdmour: not really. the query does not matter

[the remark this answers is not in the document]
mutdmour: I just changed it to an empty string for clarity

src/components/test/QuickOpenModal.spec.js
```diff
@@ -210,6 +287,542 @@ describe("QuickOpenModal", () => {
...
+          searchType: "shortcuts"
+        },
+        "shallow"
+      );
+      wrapper.setState(() => ({
+        results: [{}, { id: ":" }],
```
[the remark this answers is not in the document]
mutdmour: the number of results has no effect. It's my way of testing that it's getting the right index, bundled up with the test for symbol search. Will just split them up for clarity and create a separate test to make sure it's getting the right index.

src/components/test/QuickOpenModal.spec.js
```diff
@@ -210,6 +287,542 @@ describe("QuickOpenModal", () => {
...
+      const sourceId = "sourceId";
+      const { wrapper, props } = generateModal(
+        {
+          enabled: true,
+          query: "test",
+          searchType: "variables",
+          selectedSource: I.Map({ id: sourceId }),
+          symbols: {
+            functions: [{ title: "functionA" }, { title: "anonymous" }],
+            variables: {}
```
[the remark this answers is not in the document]
mutdmour: it does not really use them. but `onChange` throws an error sometimes if functions are not defined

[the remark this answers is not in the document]
mutdmour: though `functions` can be an empty array. will change it to that for clarity

src/components/test/QuickOpenModal.spec.js
mutdmour: I think the code expects all variables to have a location, I just only give `location` to the one it expects to match. will throw an error on any of the others.

src/components/test/QuickOpenModal.spec.js
mutdmour: ya, so it's opening the file with that variable, but that variable does not have a location, so it defaults to line 0

src/components/test/QuickOpenModal.spec.js
codehag: ah sorry i wasnt too clear with my question: is there a difference between what happens when we traverse functions and variables?

src/components/test/QuickOpenModal.spec.js
codehag: that might be worth putting into a separate test, because here the test is expecting variable traversal and what we are testing is if the default behavior is what we expect

codehag:
Looks really great! i think one final push and it will be done! thanks so much for your hard work!

src/components/test/QuickOpenModal.spec.js
mutdmour: traversal is the same, but selection is different. there's other traversal tests looking at functions

src/components/test/QuickOpenModal.spec.js
mutdmour: done

jasonLaster:
By the way @mutdmour, we're refactoring the source tree in #6328 and your tests made that so much easier

jasonLaster:
This is so great. QuickOpen does so much, now we can test its public API

PR #6300, opened 2018-05-13
[Frames] Show Components

jasonLaster:
Fixes Issue: #3792 

### Summary of Changes

Here's an initial re-write of the components pane, with the goals of

- separating out components from extra
- adding children
- making it behave like a tree (WIP - we should use the Tree component)

Next steps:

1. land this branch
2. Use the managed tree component to let users expand and see other component children
3. on select navigate to the components render function (we have location)
4. on select update the props&state in the component pane

https://github.com/jasonLaster/debugger.html/tree/components


darkwing:
OMG dude, this was the first thing I thought of when I woke up this morning.  AND YOU STILL BEAT ME TO IT!

jasonLaster:
Haha I need all the help I can get here. This feature is big. It reminds me
of what it felt like building v1 of the call stack

src/actions/pause/components.js
```diff
@@ -0,0 +1,115 @@
...
+            class: child.type
+          }))
+      }
+    } else {
+      []
+    }
+  `);
+}
+
+function getAncestors(evaluate: Function) {
```
gaearon: 😢 

All of this can break in any patch release. *Please* don't rely on these. We don't want to break Firefox debugger.

We are happy to talk about turning these into semi-public APIs you could use. Let's work together on this.

gaearon: Specifically I believe we mentioned the `console.reactStack()` proposal before. It's not hard to add to React.

jasonLaster: @gaearon it would be nice to wrap revisit `console.reactStack` and see what kind of API makes sense.

darkwing:
Nice! Was able to give this a run and the functionality look sweet!

What are next steps?

jasonLaster:
hacking on it now. I suppose we could try to land this behind a flag.

The next steps are

1) have it look like the source tree / call stack
2. on select navigate to the component

jasonLaster:
just pushed some updates

component tree

jasonLaster:
feels much nicer

jasonLaster:
Alright, this is ready for review

The best way to test is to go [here](https://dbg-class-prop-this.glitch.me/) and add a breakpoint in ListItem#render. You should see the new component Stack.

darkwing:
Side note: found this during the testing process:  

https://github.com/devtools-html/debugger.html/issues/6385

darkwing:
There also doesn't appear to be a `ListItem` component at that address

jasonLaster:
@darkwing oops. any render will do

jasonLaster:
closing. we can bring this back when we're focusing again on framework features

PR #6308, opened 2018-05-14
Add test for utils/ui.

adancarrasco:
Fixes Issue: #5100

Hi, 
This is my first contribution to the project.

### Summary of Changes

* Add Test for utils/ui
`src/utils/tests/ui.spec.js`

I was planning to add more tests for the other methods, but after @jasonLaster feedback on Slack would be better to stub those.

image

Thanks!


darkwing:
This is a cool first test, @adancarrasco ; nice work!

PR #6309, opened 2018-05-15
Fix #6201 - Prevent relatedTarget.classList error

### LINKED ISSUE #6201: Error: relatedTarget.classList is undefined in Popup.js
classlist

I haven't been able to reproduce this yet but just something we should be aware of; at the very least check that classList exists.

darkwing:
Fixes Issue: #6201


darkwing:
The failing test here doesn't appear to have anything to do with this PR

PR #6311, opened 2018-05-15
Fix onCmdCtrlClick handler.

nchevobbe:
We weren't checking the appropriate event property on non-mac OSes.

This still need some webpack configuration, because I don't want to get devtools-modules into the reps bundle.


packages/devtools-reps/src/object-inspector/component.js
```diff
@@ -9,6 +9,10 @@ const dom = require("react-dom-factories");
 const { connect } = require("react-redux");
 const { bindActionCreators } = require("redux");
 
+import { Services } from "devtools-modules";
```
juliandescottes: We can't import Services in reps because of the JSON viewer :/

Issue #5999, opened 2018-04-17
[Breakpoints] Enhance New Breakpoints UI


darkwing:
This serves as a tracking issue for the enhancements to the new Breakpoints UI

## Tasks
- [x] Create SourceIcon component #6001
- [x] Show Source Icons with Group/File Name #6000

Issue #6001, opened 2018-04-17
[Breakpoints] Create SourceIcon Component


darkwing:
Parent issue:  #5999

We presently have icons for sources in the SourcesTree, Tabs, and we'll be adding them to the Breakpoints panel with #6000 .

It would be nice to have a unified Component to make icon display easier.  There may be code to borrow here:

https://github.com/devtools-html/debugger.html/pull/5985/files

Issue #6000, opened 2018-04-17
[Breakpoints] Show Source Icons with Group/File Name


darkwing:
Parent issue:  #5999

We show icons in the Source Tree and sometimes Tabs for a given resource.  Example tabs include React, Angular, Pretty Print, Blackboxed, etc.

It would be nice if we carried those over to the breakpoints UI:

<img width="307" alt="breakpointicon" src="https://user-images.githubusercontent.com/46655/38900067-ba2121ea-425f-11e8-8035-a428f4d92bc5.png">

@jasonLaster has some code in #5985 that can be borrowed.

jasonLaster:
How's it going @abitravers1989? Anything i can do to help?

PR #6312, opened 2018-05-15
Introduce and Implement a SourceIcon Component

darkwing:
Fixes Issue: #5999, #6000, #6001(?)

### Summary of Changes

Debugger uses many icons throughout the App, most of the time based on characteristics of a source, yet we have multiple places where icons are calculated, and thus having one component to do all that repeated work would help with consistency of UI and maintainability of code.

### Test Plan

Since this PR is only meant to cover the breakpoints pane and tabs, open tabs in blackboxed, pretty-print, and react states, and ensure they look good.  Ensure the same in the breakpoints pane.  Check out the tree that icons look good there as well.

### ToDo

I don't want this PR to grow out of control but there are a few things that still need to happen:

#### *Musts* Before Merge

- [x] Fix the vertical alignment of icons in tabs
- [x] For some really weird reason the React icon doesn't show up in a tab until its tab is selected -- really weird.  Maybe we're calculating data in a lazy fashion?

### Total Components that will Require Implementation
- [ ] SourcesTree - this will be difficult because we're creating icons by `item`, not source
- [ ] ResultList - specifically the QuickOpen, where we also need to be tab-aware
- [ ] Workers pane - lowest priority, IMO

### DIFF
--- a/src/components/shared/SourceIcon.js
+++ b/src/components/shared/SourceIcon.js
@@ -6,7 +6,10 @@
 
 import React, { PureComponent } from "react";
 
+import { connect } from "react-redux";
+
 import { getSourceClassnames } from "../../utils/source";
+import { getSourceMetaData } from "../../selectors";
 
 import type Source from "../../types";
 import type { SourceMetaDataType } from "../../reducers/ast";
@@ -16,7 +19,7 @@ import "./SourceIcon.css";
 type Props = {
   source: Source,
   // sourceMetaData will provide framework information
-  sourceMetaData?: SourceMetaDataType,
+  sourceMetaData: SourceMetaDataType,
   // Array of strings representing cases where we prefer to get no image
   renderNothingIfIncludes?: Array
 };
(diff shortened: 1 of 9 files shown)


jasonLaster:
Looks great!

wldcordeiro:
Looking awesome, love the consolidation work.

src/components/shared/SourceIcon.js
```diff
@@ -0,0 +1,47 @@
...
+
+class SourceIcon extends PureComponent<Props> {
+  render() {
+    const { renderNothingIfIncludes, source, sourceMetaData } = this.props;
+    const iconClass = getSourceClassnames(source, sourceMetaData);
+
+    if (
+      renderNothingIfIncludes &&
+      renderNothingIfIncludes.includes(iconClass)
+    ) {
```
jasonLaster: nit: this could be a more generic prop:

```js
<SourceIcon shouldHide={icon => ["js"].includes(icon)}
```

```js
if(shouldHide(souceClassNames)) {
  return null
}
```

PR #6317, opened 2018-05-16
[WIP][Tree] Virtualize tree

AnshulMalik:
Fixes Issue: #4168

### Summary of Changes

* Virtualize the 🌲 component so that we can handle huge trees

Currently, the source tree works really nice and fast. But there are some side effects, the scopes panel has some issues, when we expand, it doesn't expand the first time, but expands second time. I think that's because the properties are fetched asynchronously and the re-render doesn't take place when the results arrive.

### DIFF
--- a/packages/devtools-components/src/tests/tree.js
+++ b/packages/devtools-components/src/tests/tree.js
@@ -43,7 +43,6 @@ function mountTree(overrides = {}) {
                 },
                 getRoots: () => ["A", "M"],
                 getKey: x => `key-${x}`,
-                itemHeight: 1,
                 onFocus: x => {
                   this.setState(previousState => {
                     return { focused: x };
--- a/packages/devtools-components/src/tree.js
+++ b/packages/devtools-components/src/tree.js
@@ -372,8 +372,8 @@ class Tree extends Component {
       traversal: []
     };
 
-    this._onExpand = oncePerAnimationFrame(this._onExpand).bind(this);
-    this._onCollapse = oncePerAnimationFrame(this._onCollapse).bind(this);
+    this._onExpand = this._onExpand.bind(this);
+    this._onCollapse = this._onCollapse.bind(this);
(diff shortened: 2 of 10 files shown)


jasonLaster:
Oh neat, i believe in the past we might have tried something like this. At first glance your implementation looks great. Did you compare this approach to what is in react-virtualized?

@nchevobbe curious what you think

AnshulMalik:
@jasonLaster I had a look at react-virtualized, it seemed to me that we are using a lot of custom functionality, like expand and collapse and handling keyboard navigation, I am not sure if we can do those things without modifying the codebase of react-virtualized. So I went the other way.

jasonLaster:
Oh yeah, was mostly asking because it is nice to see how others do things. Have you checked to see how this does with scrolling? I believe that was an issue with react-virtualized

https://hacks.mozilla.org/2017/11/async-panzoom-apz-lands-in-firefox-quantum/

AnshulMalik:
Thanks @nchevobbe, those link will really help.

And yeah, when height is not fixed, then my current solution won't work. I'll see how react-virtualised achieves that.

I'll also explore IntersectionObserver.

jasonLaster:
### Editor/componentWillReceiveProps

We try and resize the gutter and bp every time, that is kind of crazy. We should only do this, if the source has changed or some other indicator:
`resizeBreakpointGutter`, `resizeToggleButton`



### Editor/componentWillUpdate

can we prevent doing costly work here?

    this.setText(nextProps);
    this.setSize(nextProps);
    this.scrollToLocation(nextProps);

 

### Editor/setupEditor

4 seconds, can we speed this up?



### SourcesTree/renderItem

It's by far the slowest part of the tree, and renderItem is disproportionately spending time in getIcon, which is URL.parse heavy...

[video](http://recordit.co/s02VZpD7Yb)

jasonLaster:
closing this so we can focus on the quick wins above

PR #6320, opened 2018-05-16
Removes toJS from getGeneratedSource

lukaszsobek:
`getGeneratedSource` only checks if source exists exists and grabs the `id` off of it. The same can be done with a `Record`, so there should be no need for a `toJS()` here.


jasonLaster:
i think this will pass after a rebase

PR #6324, opened 2018-05-16
Start failing if the build fails

jasonLaster:
### Summary of Changes

Teach travis to fail if the build fails


darkwing:
Oddly enough it says the build is successful

jasonLaster:
yeah. not sure about that... :)

jasonLaster:
not sure what's up w/ this test: `browser_dbg-scopes-mutations` 

also not thrilled that we have the firefox build output on exit...

PR #6327, opened 2018-05-17
✨: Use `jest-runner-write-prettier` to perform formatting

wldcordeiro:
### Summary of Changes

* In line with my work using Jest runners for eslint, stylelint and testing this adds prettier into the mix!

### Test Plan

Example test plan:

- [x] `yarn lint`
- [x] `yarn test`


package.json
```diff
@@ -162,19 +161,20 @@
     "remark-preset-lint-recommended": "^3.0.0",
     "remark-validate-links": "^6.1.0",
     "rimraf": "^2.6.1",
+    "rxjs-compat": "^6.1.0",
```
wldcordeiro: This is here because `lint-staged` depends on something called `any-observable` that only works with RxJS up to 5, but the jest runner is using rxjs 6 so this causes a breakage, by adding the compat layer we get working `lint-staged` until they update to support.

PR #6334, opened 2018-05-17
Fix #6333 - Prevent overflow of contents for expressions

### LINKED ISSUE #6333: [Expressions] expanding a node does not expand the pane
STR

1. pause and add an expression, which is an object
2. expand some keys

darkwing:
Fixes Issue: #6333


jasonLaster:
okay. saw that error somewhere else as well.

PR #6337, opened 2018-05-17
add a search field for getOriginalLocation

jasonLaster:
### Summary of Changes

It is possible that a generated location does not map to a specific original location, in that case we might want to search for the best match given a bias.

### DIFF
--- a/packages/devtools-source-map/src/source-map.js
+++ b/packages/devtools-source-map/src/source-map.js
@@ -155,7 +155,13 @@ async function getAllGeneratedLocations(
   }));
 }
 
-async function getOriginalLocation(location: Location): Promise<Location> {
+type locationOptions = {
+  search?: "LEAST_UPPER_BOUND" | "GREATEST_LOWER_BOUND"
+};
+async function getOriginalLocation(
+  location: Location,
+  { search }: locationOptions = {}
+): Promise<Location> {
   if (!isGeneratedId(location.sourceId)) {
     return location;
   }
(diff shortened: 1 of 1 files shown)


src/actions/sources/select.js
```diff
@@ -194,6 +194,32 @@ export function selectSpecificSource(sourceId: string) {
   };
 }
 
+export function selectVisibleLocation(clientLocation: Location) {
+  return async ({ dispatch, getState, client, sourceMaps }: ThunkArgs) => {
+    const selectedSource = getSelectedSource(getState());
+    const locationSource = getSourceByURL(getState(), clientLocation.url);
```
jasonLaster: this probably should move out of here so we're just using normal locations.

packages/devtools-source-map/src/source-map.js
```diff
@@ -183,6 +183,40 @@ async function getOriginalLocation(location: Location): Promise<Location> {
...
+  let offset = 0;
+  while (!match || !match.source) {
+    match = map.originalPositionFor({
+      line: location.line + offset++,
+      column: location.column == null ? 0 : location.column,
+      bias: SourceMapConsumer.LEAST_UPPER_BOUND
+    });
+  }
+
+  const { source: sourceUrl, line, column } = match;
```
jasonLaster: I should probably assume we might never have a match and return null there

packages/devtools-source-map/src/source-map.js
```diff
@@ -165,11 +171,31 @@ async function getOriginalLocation(location: Location): Promise<Location> {
...
     column: location.column == null ? 0 : location.column
   });
 
+  // If there is not an exact match, look for a match with a bias at the
+  // current location and then on subsequent lines
+  if (search) {
+    let line = location.line;
+    let column = location.column == null ? 0 : location.column;
+
+    while (!match.sourceUrl) {
```
loganfsmyth: I think this is an infinite loop. You want `match.source` or probably better `match.source === null`.

packages/devtools-source-map/src/source-map.js
```diff
@@ -165,11 +171,31 @@ async function getOriginalLocation(location: Location): Promise<Location> {
...
+      match = map.originalPositionFor({
+        line,
+        column,
+        bias: SourceMapConsumer[search]
+      });
+
+      line = search == "LEAST_UPPER_BOUND" ? line++ : line--;
+      line = search == "LEAST_UPPER_BOUND" ? 0 : Infinity;
```
loganfsmyth: Is this supposed to be `column = `?

packages/devtools-source-map/src/source-map.js
```diff
@@ -165,11 +171,31 @@ async function getOriginalLocation(location: Location): Promise<Location> {
...
+    let column = location.column == null ? 0 : location.column;
+
+    while (!match.sourceUrl) {
+      match = map.originalPositionFor({
+        line,
+        column,
+        bias: SourceMapConsumer[search]
+      });
+
+      line = search == "LEAST_UPPER_BOUND" ? line++ : line--;
```
loganfsmyth: I think assigning and doing `line++` on the same line is a little confusing. Maybe

```
line += search == "LEAST_UPPER_BOUND" ? 1 : -1;
```
?

PR #6338, opened 2018-05-17
sync changes from release 56

jasonLaster:
### Summary of Changes

A couple of things that were in our release branch, but should have also been in master

Note, I'm also copying over the changes that were in master, but should have been in release [here](https://github.com/devtools-html/debugger.html/compare/28f2817...release-56). Let me know if i missed something that should go here. For example, if some of the CSS was correct in the release branch and wrong here, we should add it to this commit so it doesnt get changed in MC!

### DIFF
--- a/src/test/mochitest/browser_dbg-preview.js
+++ b/src/test/mochitest/browser_dbg-preview.js
@@ -18,8 +18,8 @@ add_task(async function() {
   await selectSource(dbg, "preview.js");
 
   await previews(dbg, "empties", [
-    { line: 2, column: 9, expression: "a", result: '""' },
-    { line: 3, column: 9, expression: "b", result: "false" },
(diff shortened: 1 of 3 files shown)


src/test/mochitest/browser_dbg-preview.js
```diff
@@ -18,8 +18,8 @@ add_task(async function() {
   await selectSource(dbg, "preview.js");
 
   await previews(dbg, "empties", [
-    { line: 2, column: 9, expression: "a", result: '""' },
-    { line: 3, column: 9, expression: "b", result: "false" },
+    // { line: 2, column: 9, expression: "a", result: '""' },
+    // { line: 3, column: 9, expression: "b", result: "false" },
```
jasonLaster: I don't love doing this, but it seems appropriate for now

PR #6341, opened 2018-05-18
[reps] Don't include devtools-modules in the reps bundle.

nchevobbe:
We add an alias for Services so we use the one from mozilla-central.
This means we also need to add a moduleMapper entry for jest test to pass.

---

**TEST PLAN**

- `cd packages/devtools-reps; yarn; yarn start`
- Make sure you can access the reps test app and evaluate things

- Creating the reps bundle should work, and it shouldn't include devtools-modules code
- Running the debugger in launchpad should work as well

### DIFF
--- a/configs/mozilla-central-mappings.js
+++ b/configs/mozilla-central-mappings.js
@@ -24,5 +24,6 @@ module.exports = {
   "../assets/panel/debugger.properties": "devtools/shared/flags",
   "devtools-connection": "devtools/shared/flags",
   "chrome-remote-interface": "devtools/shared/flags",
-  "devtools-launchpad": "devtools/shared/flags"
+  "devtools-launchpad": "devtools/shared/flags",
+  "devtools-services": "Services"
 };
--- a/packages/devtools-reps/package.json
+++ b/packages/devtools-reps/package.json
@@ -42,6 +42,7 @@
     "devtools-launchpad": "^0.0.125",
     "devtools-license-check": "^0.5.1",
     "devtools-modules": "^0.0.36",
+    "devtools-services": "^0.0.1",
     "enzyme": "^3.3.0",
     "enzyme-adapter-react-16": "^1.1.1",
     "enzyme-to-json": "^3.3.1",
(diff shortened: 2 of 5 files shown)


jasonLaster:
Alright, so i did a couple of things to research this. 

1. i created a devtools-services package which would simplify what you're doing [here][p]. I published it so you can try it out locally.

2. I looked at what we were doing before, which turns out to be pretty good. We were mapping services [here][h]. So, this patch gets you a smaller Reps bundle, but i believe you were safe w/ respect to services.

I think the best outcome would be to land services so you can require it directly. And then the reps bundle will not need all of modules.

My primary concern with what you have is the aliases add a lot of logic because they're based on paths and it is simpler to work on the package level where you can just add one mapping at the end.

[p]; https://github.com/devtools-html/devtools-core/pull/1066
[h]: https://github.com/devtools-html/devtools-core/blob/master/packages/devtools-launchpad/webpack.config.devtools.js#L28-L32

webpack.config.js
```diff
@@ -40,6 +40,10 @@ const webpackConfig = {
     path: path.join(__dirname, "assets/build"),
     filename: "[name].js",
     publicPath: "/assets/build"
+  },
+
+  resolve: {
+    alias: {}
   }
```
jasonLaster: do we need this?

webpack.config.js
```diff
@@ -57,6 +61,10 @@ function buildConfig(envConfig) {
   if (isDevelopment()) {
     webpackConfig.module = webpackConfig.module || {};
     webpackConfig.module.rules = webpackConfig.module.rules || [];
+    webpackConfig.resolve.alias.Services = path.join(
+      __dirname,
+      "node_modules/devtools-modules/src/Services"
+    );
   } else {
```
jasonLaster: do we need this?

webpack.config.js
nchevobbe: oh yeah, probably not, I forgot to check what could be removed, let me see and push again

webpack.config.js
nchevobbe: same, probably not

Issue #5019, opened 2018-01-03
[Tabs] Improve truncation of long filenames, smaller max tab size 


violasong:
STR: 
- Open Debugger when viewing a github issues page like this one 
- Navigate to `assets-cdn.github.com > assets > frameworks-alsxsljd...`
- The tab is too wide and the info in it is hashy gibberish
- Bonus problem: click Pretty Print
The source map message mentions the filename with the same poor truncation rules, and when the Debugger is narrow it breaks in this odd way:
<img width="462" alt="image" src="https://user-images.githubusercontent.com/55994/34536772-16a511ba-f07b-11e7-9d16-3508b0d54610.png">

Suggestions:
- Max truncation level should be at most half what it is now, perhaps 20 chars or so. This will keep the max tab size smaller, even when there's only one tab.
- Do the truncation at the end instead of beginning. So with this file it would end up looking like `frameworks-fbf9e3a...`
- When the window is very narrow, more truncation would happen so that you can always see the full tab UI. (The tab would look like `[ framew...  x ]` - I think we could do this with text-overflow: ellipsis;
- Hm, for the "source mapped from" message, would it make sense to remove that message, and instead keep the pretty print icon to the left but in a highlighted state?
- Hovering over anything truncated should show a tooltip with full text

(Sprouted from discussions in #4928)

CharlesStocky:
This is about to be PR'ed (https://github.com/devtools-html/debugger.html/pull/5415)

PR #6349, opened 2018-05-18
[Tabs] Improve truncation of long filenames

adancarrasco:
Fixes Issue: #5019 

### Summary of Changes

* Create a utility method to truncate text in the middle (...)
* Create test for this utility method

### Test Plan
- [x] Open Debugger when viewing a github issues page
- [x] Navigate to `assets-cdn.github.com > assets > frameworks-alsxsljd...`
- [x] See the name has been truncated in the middle (showing the extension and first part of the name)

### Screenshots
image

### DIFF
--- a/src/utils/tests/text.spec.js
+++ b/src/utils/tests/text.spec.js
@@ -8,7 +8,7 @@ describe("text", () => {
   it("should truncate the text in the middle", () => {
     const sourceText = "this is a very long text and ends here";
     expect(truncateMiddleText(sourceText, 30)).toMatch(
-      "this is a ver...and ends here"
+      "this is a ver... and ends here"
     );
   });
   it("should keep the text as it is", () => {
--- a/src/utils/tests/text.spec.js
+++ b/src/utils/tests/text.spec.js
@@ -1,0 +1,20 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
+
+import { truncateMiddleText } from "../text";
+
+describe("text", () => {
+  it("should truncate the text in the middle", () => {
+    const sourceText = "this is a very long text and ends here";
+    expect(truncateMiddleText(sourceText, 30)).toMatch(
+      "this is a ver...and ends here"
+    );
+  });
+  it("should keep the text as it is", () => {
+    const sourceText = "this is a short text ends here";
+    expect(truncateMiddleText(sourceText, 30)).toMatch(
+      "this is a short text ends here"
+    );
+  });
+});
--- a/src/utils/text.js
+++ b/src/utils/text.js
@@ -39,4 +39,26 @@ function formatKeyShortcut(shortcut: string): string {
     .replace(/Shift\+/g, "Shift ");
 }
 
-export { formatKeyShortcut };
+/**
+ * Truncates the received text to the maxLength in the format:
+ * Original: 'this is a very long text and ends here'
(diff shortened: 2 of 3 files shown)


darkwing:
Ooooh, I love this!

adancarrasco:
I'm glad you liked it @darkwing, it's a small change but I think it makes a big difference when working with long filenames. 🦊

darkwing:
Thoughts on this strategy @jasonLaster?  I'm a big fan of this.

src/utils/tests/text.spec.js
```diff
@@ -0,0 +1,12 @@
...
+ * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
+
+import { truncateText } from "../text";
+
+describe("text", () => {
+  it("should truncate the text", () => {
+    const text = "this is a very long text and ends here";
+    expect(truncateText(text, 30)).toMatch("this is a ver...and ends here");
+  });
+});
```
darkwing: I know this seems weird but a test with a short string confirming it's unchanged would be good. :)

src/utils/text.js
```diff
@@ -39,4 +39,24 @@ function formatKeyShortcut(shortcut: string): string {
...
+/**
+ * Truncates the received text to the maxLength in the format:
+ * Original: 'this is a very long text and ends here'
+ * Truncated: 'this is a ver...nd ends here'
+ * @param {String} text - Original text
+ * @param {Number} maxLength - Max allowed length
+ * @memberof utils/text
+ * @static
+ */
+function truncateText(text: string, maxLength: number): string {
```
jasonLaster: i kinda think of truncating at the end as the default. Do want to name this function `truncateMiddleText`

Also do we need the word text, that makes me think of source text

src/utils/text.js
```diff
@@ -39,4 +39,24 @@ function formatKeyShortcut(shortcut: string): string {
...
+  if (text.length > maxLength) {
+    truncatedText = `${text.substring(
+      0,
+      Math.round(maxLength / 2) - 2
+    )}...${text.substring(text.length - Math.round(maxLength / 2 - 2))}`;
+  }
+  return truncatedText;
+}
+
+export { formatKeyShortcut, truncateText };
```
jasonLaster: looks like we have endTruncateStr
https://github.com/devtools-html/debugger.html/blob/master/src/utils/utils.js#L41-L46

should we move this there and rename both to truncateEnd, truncateMiddle?

jasonLaster: good call

src/utils/text.js
adancarrasco: I have updated the name as suggested. :)

src/utils/tests/text.spec.js
adancarrasco: Done.

src/utils/tests/text.spec.js
```diff
@@ -0,0 +1,20 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
+
+import { truncateMiddleText } from "../text";
+
+describe("text", () => {
+  it("should truncate the text in the middle", () => {
+    const sourceText = "this is a very long text and ends here";
+    expect(truncateMiddleText(sourceText, 30)).toMatch(
+      "this is a ver...and ends here"
```
darkwing: It turns out this is actually only 29 characters; I think we should add the extra character to the end

adancarrasco: Sorry @darkwing, I got some work and haven't been able to look at this before. Thanks for helping out!

PR #6350, opened 2018-05-18
📝: Tweak developer experience for testing and linting

wldcordeiro:
### Summary of Changes

* Also, document the `test:js` command which runs just the Jest unit test project
* Turns out there is an issue with the list format lint-staged uses and Jest. This fixes it by using
the CLI flag added here https://github.com/facebook/jest/pull/1678

### DIFF
--- a/package.json
+++ b/package.json
@@ -169,12 +169,12 @@
     "workerjs": "github:jasonLaster/workerjs"
   },
   "lint-staged": {
-    "bin/*.js": ["yarn lint", "git add"],
-    "src/*.js": ["yarn lint", "git add"],
-    "src/*/*.js": ["yarn lint", "git add"],
-    "src/*/!(mochitest)**/*.js": ["yarn lint", "git add"],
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -169,12 +169,12 @@
     "workerjs": "github:jasonLaster/workerjs"
   },
   "lint-staged": {
-    "bin/*.js": ["yarn lint", "git add"],
-    "src/*.js": ["yarn lint", "git add"],
-    "src/*/*.js": ["yarn lint", "git add"],
-    "src/*/!(mochitest)**/*.js": ["yarn lint", "git add"],
-    "src/*/!(mochitest)*/**/*.js": ["yarn lint", "git add"],
-    "packages/**/src/*.js": ["yarn lint", "git add"]
+    "bin/*.js": ["yarn lint --findRelatedTests", "git add"],
```
wldcordeiro: Long story short, lint-staged takes the staged files and makes them a space separated list. Jest doesn't understand this by default. This flag makes it understand.

Issue #6347, opened 2018-05-18
[Tabs] Chevron overlaps other tabs


jasonLaster:
It looks like our new chevron image overlaps other tabs.

We should find a way to either:

1. have padding on the right so that the tabs do not go under the chevron.
2. hide the chevron when space gets tight :)

![](http://g.recordit.co/RxZrICI3Ui.gif)

lzzluca:
@jasonLaster Hi! This fix works for me:

https://github.com/devtools-html/debugger.html/pull/6356

Checking the debugger code for the first time as part of React Europe 2018 Hackathon, hope the fix can be of interest! :)

jasonLaster:
Oh fantastic. Happy hacking

adancarrasco:
Seems that #6356 fixed it. Just wondering if we can close this issue as well? :)

PR #6356, opened 2018-05-19
[Tabs] chevron image overlaps other tabs

lzzluca:
https://github.com/devtools-html/debugger.html/issues/6347

### Summary of Changes

The problem: when one only tab is left in the debugger editor, it does not scale down with the editor width. That creates the overlapping issue as from ticket.
Tuned up `.source-tab`'s width to prevent the situation described above.

Tested on FF 60 and Chrome 67

### Test Plan

Open multiple tabs on the debugger editor and drag the editor pane as much as possible, to reduce the space availabale for tabs to the minimum as possible.

Expected: the last tab displayed will reduce its width accordingly to the space available. Also the dropdown button (to see the hidden tabs) is expected to render correctly (a bug prevent the button icon to render properly, but clicking on it is still possible to see it).

_lot of space to render tabs: all ok_
screen shot 2018-05-19 at 16 01 26

_suddenly no more space to render tabs: all hidden but one, that resize accordingly to the space available_
screen shot 2018-05-19 at 16 01 42

_also dropdown button reachable 👍_
screen shot 2018-05-19 at 16 01 50

### DIFF
--- a/src/components/Editor/Tabs.css
+++ b/src/components/Editor/Tabs.css
@@ -52,6 +52,8 @@
   margin-top: 3px;
   cursor: default;
   height: 27px;
+  width: 100%;
+  max-width: fit-content;
 }
 
 .source-tab:first-child {


darkwing:
Thank you for your efforts @lzzluca !  Welcome to the team!

TIL: `fit-contents` 👍   I was, however, able to simplify your code to just:

```css
max-width: 100%;
```

It appears to have the same effect.  Could you try that?  Thanks!

lzzluca:
@darkwing I think you are very right, well spotted!

darkwing:
Superceded by this:  https://github.com/devtools-html/debugger.html/pull/6397

Great work @lzzluca !

Issue #6354, opened 2018-05-19
[Search] results label should not wrap


jasonLaster:
STR:
1. open the launchpad
2. select a file
3. cmd+f
4. search for something which appears a lot
5. narrow the editor or the debugger quite a bit...

I think we likely want to just overflow hidden it so that at a certain point it goes away... Likely the same treatment as the buttons to the right.

![](http://g.recordit.co/RJl6Suqy1K.gif)

darkwing:
Tip: Adding `white-space: nowrap` to `.search-field .summary` will prevent this wrapping 👍

PR #6361, opened 2018-05-19
[Search] results label should not wrap 

reescott:
Fixes Issue: #6354

### Summary of Changes

* Added CSS rule to prevent search results label from wrapping.

### Test Plan

- [x] Find a file to search
- [x] Search for something with a lot of results
- [x] Resize window
- [x] "results" label no longer wraps to next line.


darkwing:
Nice!  Thank you @reescott !

PR #6363, opened 2018-05-20
Removes toJS from getEmptyLines

lukaszsobek:
`getEmptyLines` just uses `id` of the source record. No need for `toJS()` here.


src/reducers/ast.js
```diff
@@ -187,12 +187,12 @@ export function isEmptyLineInSource(
...
 
-export function getEmptyLines(state: OuterState, source: Source) {
-  if (!source) {
+export function getEmptyLines(state: OuterState, sourceRecord: SourceRecord) {
+  if (!sourceRecord) {
     return null;
   }
 
-  return state.ast.emptyLines.get(source.id);
+  return state.ast.emptyLines.get(sourceRecord.id);
```
lukaszsobek: Maybe instead of the full source record just pass in the `id` to `getEmptyLines`? Then we could just return if `!id`. Question if there are any edge cases where there is a source without an id that would break this approach...

jasonLaster: good idea, lets do that

Issue #6370, opened 2018-05-21
[Search] Results should line up with the directory


jasonLaster:
The project search results should line up w/ the directory. The line numbers should shift left to be at that blue bar :)

![Screen Shot 2018-05-21 at 2.32.59 PM.png](https://shipusercontent.com/5a5e79f29064a360111840694b8871ce/Screen%20Shot%202018-05-21%20at%202.32.59%20PM.png)

PR #6372, opened 2018-05-21
[Search] Updated padding for search results

steffiau:
Fixes Issue: #6370 

### Screenshots
screen shot 2018-05-21 at 1 12 39 pm

:)

### DIFF
--- a/src/components/ProjectSearch.css
+++ b/src/components/ProjectSearch.css
@@ -51,6 +51,7 @@
 .project-text-search .result .line-number {
   padding-left: 5px;
   min-width: 33px;
+  width: 100%;
   text-align: right;
   grid-column: 1/2;
 }
--- a/src/components/ProjectSearch.css
+++ b/src/components/ProjectSearch.css
@@ -24,7 +24,7 @@
 }
 
 .project-text-search .result {
-  display: flex;
+  display: contents;
   cursor: default;
   padding: 4px 0 4px 5px;
   line-height: 16px;
@@ -49,9 +49,14 @@
 }
 
 .project-text-search .result .line-number {
-  margin-right: 5px;
-  min-width: 28px;
+  padding-left: 5px;
+  min-width: 33px;
   text-align: right;
+  grid-column: 1/2;
+}
+
+.project-text-search .result .line-value {
+  grid-column: 2/3;
 }
 
 .project-text-search .tree-indent {
@@ -73,6 +78,7 @@
   cursor: default;
   padding: 2px 0 2px 5px;
   font-size: 12px;
+  grid-column: 1/3;
 }
 
 .project-text-search .file-result .arrow {
@@ -101,6 +107,13 @@
 
 .project-text-search .managed-tree .tree {
   height: 100%;
+  display: grid;
+  grid-template-columns: auto 1fr;
+  grid-column-gap: 5px;
+}
+
+.project-text-search .managed-tree .tree .tree-node {
+  display: contents;
 }
 
 /* Focus values */
--- a/src/components/ProjectSearch.css
+++ b/src/components/ProjectSearch.css
@@ -26,7 +26,7 @@
 .project-text-search .result {
   display: flex;
   cursor: default;
-  padding: 4px 0 4px 30px;
+  padding: 4px 0 4px 0;
   line-height: 16px;
   font-size: 10px;
   width: 100%;
--- a/src/components/ProjectSearch.css
+++ b/src/components/ProjectSearch.css
@@ -58,6 +58,7 @@
 
 .project-text-search .result .line-value {
   grid-column: 2/3;
+  padding-left: 5px;
 }
 
 .project-text-search .tree-indent {
@@ -110,7 +111,6 @@
   height: 100%;
   display: grid;
   grid-template-columns: auto 1fr;
-  grid-column-gap: 5px;
 }
 
 .project-text-search .managed-tree .tree .tree-node {
@@ -120,12 +120,13 @@
 /* Focus values */
 
 .project-text-search .file-result.focused,
-.project-text-search .result.focused {
+.project-text-search .result.focused .line-value {
   background-color: var(--theme-selection-background);
 }
 
 .project-text-search .result.focused .line-number {
   font-weight: bolder;
+  background-color: var(--theme-selection-background);
 }
 
 .project-text-search .result.focused .query-match {
--- a/src/components/ProjectSearch.css
+++ b/src/components/ProjectSearch.css
@@ -26,7 +26,7 @@
 .project-text-search .result {
   display: flex;
   cursor: default;
-  padding: 4px 0 4px 0;
+  padding: 4px 0 4px 5px;
   line-height: 16px;
   font-size: 10px;
   width: 100%;
@@ -49,9 +49,13 @@
 }
 
 .project-text-search .result .line-number {
-  margin-right: 1em;
-  width: 2em;
-  margin-left: 0.8em;
+  margin-right: 5px;
+  min-width: 28px;
+  text-align: right;
+}
+
+.project-text-search .tree-indent {
+  display: none;
 }
 
 .project-text-search .no-result-msg {
--- a/src/test/mochitest/helpers.js
+++ b/src/test/mochitest/helpers.js
@@ -1059,7 +1059,7 @@ const selectors = {
   sourceNodes: ".sources-list .tree-node",
   sourceDirectoryLabel: i => `.sources-list .tree-node:nth-child(${i}) .label`,
   resultItems: ".result-list .result-item",
-  fileMatch: ".managed-tree .result",
+  fileMatch: ".project-text-search .line-value",
   popup: ".popover",
   tooltip: ".tooltip",
   previewPopup: ".preview-popup",


jasonLaster:
Looks great

nchevobbe:
Thanks @steffiau !
Do you think we could align the numbers to the right of the icon, and the search result items to the beginning of the folder name ?

jasonLaster:
I think that's a good point @nchevobbe

steffiau:
Like this?
screen shot 2018-05-22 at 5 47 08 pm

nchevobbe:
> Like this?

Yes, that looks nice.
Could you post a screenshot with a file of ~100 000 lines so we can see how the "gutter" acts when the line number has 6 numbers ?

darkwing:
@nchevobbe I've made a few modifications to some CSS:

updates

I don't think there's a catch-all solution to ensure numbers line up properly.  Switching to a min-width instead of width seems like a good compromise.  Thoughts?

steffiau:
@darkwing I think that's a good idea. The only other idea I can think of right now is to wrap the number onto a second line, but that would likely be confusing.

jasonLaster:
what we do in the editor is measure the max width and then use that.
could we also have a separate div for the line numbers and then use CSS to fit that?

darkwing:
Find in files uses our Tree widget so we *might* be slightly constrained; I'll check that out

steffiau:
@darkwing If you don't mind me asking, how did you remove the '.tree-indent' spans next to the results?
screen shot 2018-05-26 at 10 20 51 am
Do you remove/modify the code below in tree.js? is that okay since another file seems to be using the tree widget (SourcesTree.js)?
`const treeIndent = dom.span({ className: "tree-indent" }, "\u200B");`
`    const indents = Array.from({ length: depth }).fill(treeIndent);
    const items = indents.concat(
      renderItem(item, depth, focused, arrow, expanded)
    );`

I've only recently picked up React, so I'm still trying to wrap my head around it. I'd like to understand the codebase better too. Sorry if this seems like a trivial question!

darkwing:
I would just use CSS to `display: none;` them. :)

steffiau:
Oh good idea haha! I haven't thought of that.

steffiau:
I've only been able to make it work using min-width so far.

screen shot 2018-06-02 at 2 32 17 pm

darkwing:
Thank you for working on this @steffiau -- it's a tough problem to tackle.  I was thinking about this over the weekend and I have another strategy you can try:

https://codepen.io/darkwing/pen/vrLqRN

If we can get `display: table / table-row / table-cell` working for `.tree` and `.tree-node` in the Find in Files result list, the line number divs will size all together, like my CodePen example.

steffiau:
Hi @darkwing ,

Thanks for the suggestion! I've tried fiddling using tables and css grid, but in both cases I haven't found a solution yet. Mainly because with our tree structure, we have an extra wrapper div in between the parent container and the targeted child item (In our case: .tree as container and .line-number/ .line-value as children, with 2 wrapper divs .tree-node and .result in between).

https://codepen.io/steffiau/pen/BVWyMK

I've been able to achieve what we need with css grid and `display: contents`, however IE 11 and Edge doesn't support `display: contents` yet.
https://codepen.io/steffiau/pen/rKyOja

jasonLaster:
Thanks steffi!

2 thoughts:

1. We just need to support latest chrome and Firefox

2. We don’t need the tree component, we could make a new component if it
supports collapsing and other arrow key shortcuts.

darkwing:
@steffiau Let's roll with the `display: contents` solution -- IE and Edge aren't important at the moment. Great work!

steffiau:
@jasonLaster 
> 1. We just need to support latest chrome and Firefox

Those are great words to hear!

@darkwing 
I got it working like the below:
screen shot 2018-06-14 at 9 35 27 pm

Smaller viewport:
screen shot 2018-06-14 at 9 46 45 pm

Multiple files:
screen shot 2018-06-14 at 9 52 25 pm


What do you think? Let me know if I should make any changes. If not I will commit my changes and update the pull request. Thanks!

jasonLaster:
Looks great to me!

steffiau:
I've ran tests by doing `yarn run test:all` locally and they all pass, but didn't realize I need to do extra setup for the mochi tests. I'm assuming it must've skipped over them? I'm setting up mochi to try to run it locally, but am having problems with installing mercurial and autoconf. I'll try to look into it more in the coming days.

jasonLaster:
alright, @darkwing or I could also look. I have no problem doing some mochi fix ups :wink:

darkwing:
So the reason this is failing is (I believe) because the `.managed-tree .result` elements no longer have a height due to `display: contents`.  Clicking items in the tree still works, however; working on figuring out how to get the test passing.  Nice work!

darkwing:
The other side effect of the `height` essentially being 0 is that the highlight color no longer appears when selecting or using the keyboard to navigate the list.  I'm hoping I can find a solution for this as well.

steffiau:
@darkwing thanks for looking into it! I will also try to see if there's a way around that.

steffiau:
The problem with display:contents is it removes the borders, padding and background of the element and only keep the content. That was a good catch, I should've tested it out with keyboard!

Without display:contents to make those elements become a direct children of the grid container (.tree), I don't think I can make the line numbers style like what we need using css grid. I think I might have to go back to the drawing board, likely have to look at modifying the structure.

wldcordeiro:
@steffiau would nesting grids possibly work? Make the result items into 1 grid inside the larger grid that is the list/tree? Or removing the outer in favor of just the one in the result items?

steffiau:
@wldcordeiro I think the problem with nesting grids or making result item the grid itself is each result item is now independent grids, so they won't know what the width of line number div in other result items. Hence unable to have a column that expands to the largest line number div. I might be understanding you wrong though, I've tried it here, let me know if you were thinking something else! https://codepen.io/steffiau/pen/VdBOXR

steffiau:
sorry I haven't had a chance to work on this the past week, I will try to make some time over this weekend!

steffiau:
I've tried looking into using display: tables again and flexbox but with no luck. I've considered replicating Tree component into a new component and modifying the rendered markup, but this seems like an overkill and will result in duplicated code. I might be missing something since I'm a little lost on where to start modifying the generated markup from tree.js (Looking at TreeNodeFactory).

jasonLaster:
@steffiau that might be simpler :) we have a much simple tree... Not saying you should do it, but just offering a quick take

darkwing:
Wow, this has been quite the journey -- thank you so much @steffiau for tackling this.  It even pushed me to learn more about `display: contents`!  It's so frustrating when a solution gets you soooo close but a tiny issue pops in like this!

I'm wondering if you can use, as a last resort, a `:before` or `:after` rule on tree node to give it some height?  Just my last suggestion!

steffiau:
@jasonLaster @darkwing Thanks for the suggestions! It's been a bit frustrating indeed but I wouldn't have otherwise learned the differences between grids vs html tables vs display: table either. :) I do want to see this through. I will first try using :before :after, and if that doesn't work I will see if I can create a simple tree. I have to admit I am not familiar with reactjs or the code base so any pointers would be immensely helpful! else I can slowly stab at it but it will just take a while :)

jasonLaster:
I wonder if @gabrielluong, @captainbrosset, or @oslego would have some ideas? They know a thing or two about layout 😉

captainbrosset:
What you really want here is subgrid support (CSS Grid Level 2). It is being implemented in Firefox as we speak 😄 Not sure about Chrome yet.
With subgrid, you don't have to `display:contents` all of the intermediary wrappers. This means you can use them for background color changes and such, while still preserving the column definition from the parent grid.

Having said, we don't have subgrid now ...

So the only other thing I can think of is moving the background coloring to the grid items: the line number and file name. Because they stretch to fill their grid cells entirely, their background will fill up the same room as their wrappers.

Using a ::before or ::after could work because grid supports overlapping items. So you could, in theory, place a ::before pseudo-element per line, with a lower z-index, and positioned exactly at the same place as the line number and file name (so it would be in the background). But the problem is, you are relying on automatic placement of line numbers and file names at the moment. I.e. you don't specific `grid-row` for each of them, nor should you, because that would require to do this from JS. And because of this, there's no way to also place those pseudo-elements at the right rows.

steffiau:
@captainbrosset I initially wanted to use subgrid but had to settle with `display: contents` for the exact reason that its not available yet 😅  Can't wait till its finally implemented!

I was able to get the highlight color working with your idea by moving the background colors to the grid items instead. Thank you for your suggestion!
screen shot 2018-07-17 at 8 29 02 pm

darkwing:
Hello @steffiau!  If you could push that up I'd love to help with tests!  Thank you so much for all of your efforts!

steffiau:
@darkwing Done! Sorry I haven't been able to figure out how to get mochi running on my local yet, thanks for helping!

darkwing:
I love it @steffiau !  Thank you so much for your persistence!  Amazing!

steffiau:
Sorry it took so long, thank you for everyone that helped!

PR #6389, opened 2018-05-22
accordion accessibility #4080 Need to be able to collapse expand headers with keyboard

nastark519:
Fixes Issue: #<issue number>

Here's the Pull Request Doc

### Summary of Changes

* We have a bug fix though the we can expand and collapse the headers with any key press not only the space bar.  No functionality has been compromised. We can still change selection with `tab`

### Test Plan

+ Going though the debugger panel we can shift to the next object with `tab` when landing on the headers "Watch Expressions" and "Breeakpoints" we can then expand and collapse them with using the `space bar` and any other key on the key board. with the functionality of being able to move on to the next item/object with the `tab` key.


assets/panel/debugger.properties
```diff
@@ -597,7 +597,7 @@ sources.search=%S to search
 
 # LOCALIZATION NOTE (watchExpressions.header): Watch Expressions right sidebar
 # pane header.
-watchExpressions.header=Watch expressions
+watchExpressions.header=Watch Expressions
```
flodolo: Is title case expected here? There's at least another header that doesn't use it, but I'm not sure if the context is different

darkwing:
Let's leave the language as-is @nastark519;  I think what you want to do is something like this:

```js
onKeyPress={e => if(e && e.keyCode === 32) { this.handleHeaderClick(i); } }
```

assets/panel/debugger.properties
darkwing: This shouldn't change -- will remove before merging @flodolo

src/components/shared/Accordion.js
```diff
@@ -57,6 +57,7 @@ class Accordion extends Component<Props, State> {
         <h2
           className="_header"
           tabIndex="0"
+          onKeyPress={() => this.handleHeaderClick(i)}
```
yzen: Just to add a little bit of context. The click event is not handling keypress "Enter" events because it's not an actionable (in this case clickable) tag such as <BUTTON>. This is why we explicitly need onKeyPress.

Another suggestion I have is to use key property instead of keyCode (which is deprecated) to check if the value is equal to "Enter" or " ".

BeyondUsGames:
@darkwing I'm on the team with @nastark519 and we tried for several hours to get your code suggestion to work, but it always came up with an error, telling us the if check was an unexpected token.

Do you have any suggestions about how we can change that to work properly?

darkwing:
Oops @BeyondUsGames, this would be the proper code:

```js
onKeyPress={e => { if(e && e.keyCode === 32) { this.handleHeaderClick(i); } } }
```

This diff I created looks a bit more structured:

```diff
diff --git a/src/components/shared/Accordion.js b/src/components/shared/Accordion.js
index 43a00571..be370586 100644
--- a/src/components/shared/Accordion.js
+++ b/src/components/shared/Accordion.js
@@ -49,6 +49,12 @@ class Accordion extends Component<Props, State> {
     this.forceUpdate();
   }

+  onKeyPress(e, i) {
+    if (e && (e.key === " " || e.key === "Enter")) {
+      this.handleHeaderClick(i);
+    }
+  }
+
   renderContainer = (item: AccordionItem, i: number) => {
     const { opened } = item;

@@ -58,6 +64,7 @@ class Accordion extends Component<Props, State> {
           className="_header"
           tabIndex="0"
           onClick={() => this.handleHeaderClick(i)}
+          onKeyPress={e => this.onKeyPress(e, i)}
         >
```

Let me know if you have questions!

darkwing:
Superceded by https://github.com/devtools-html/debugger.html/pull/6438

PR #6392, opened 2018-05-23
Removes toJS from findSourceMatches

lukaszsobek:
findSourcematches depends on 3 properties of the source and as such the .toJS() of the full source is not really needed.


lukaszsobek:
Will revisit later. Closing for now.

PR #6393, opened 2018-05-23
Refactors getExtension to only use url

lukaszsobek:
`getExtension` only ever uses the url, it doesn't need the whole source object.

### DIFF
--- a/src/utils/source.js
+++ b/src/utils/source.js
@@ -375,6 +375,5 @@ export function getSourceClassnames(
   if (source.isBlackBoxed) {
     return "blackBox";
   }
-
-  return sourceTypes[getExtension(source)] || defaultClassName;
+  return sourceTypes[getExtension(source.url)] || defaultClassName;
 }
--- a/src/utils/sources-tree/tests/utils.spec.js
+++ b/src/utils/sources-tree/tests/utils.spec.js
@@ -3,7 +3,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
 
-import I, { Map } from "immutable";
+import { Map } from "immutable";
 
 import {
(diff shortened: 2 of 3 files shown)


src/utils/source.js
```diff
@@ -375,6 +375,5 @@ export function getSourceClassnames(
   if (source.isBlackBoxed) {
     return "blackBox";
   }
-
-  return sourceTypes[getExtension(source)] || defaultClassName;
+  return sourceTypes[getExtension(source.url)] || defaultClassName;
```
lukaszsobek: maybe name it getFileExtension or so for more clarity?

src/utils/sources-tree/tests/utils.spec.js
```diff
@@ -157,54 +157,47 @@ describe("sources tree", () => {
   describe("getUrl", () => {
     it("handles normal url with http and https for filename", function() {
       const urlObject = getURL("https://a/b.js");
-      const urlObject2 = getURL("http://a/b.js");
-
       expect(urlObject.filename).toBe("b.js");
+
+      const urlObject2 = getURL("http://a/b.js");
```
lukaszsobek: just ordering, definition 1, test 1 - definition 2, test 2

src/utils/sources-tree/utils.js
```diff
@@ -37,8 +37,7 @@ export function isDirectory(url: Object) {
   );
 }
 
-export function getExtension(source: Object): string {
-  const url = source.get ? source.get("url") : source.url;
+export function getExtension(url: string = ""): string {
```
lukaszsobek: Our snapshot tests are written in such a way that having the url as undefined triggers a ton of snapshot errors, this is why the empty string default.

src/utils/source.js
jasonLaster: yeah, that would be nice

PR #6397, opened 2018-05-23
Add a max-width to sourcetabs to avoid chevron overlap

darkwing:
chevy


darkwing:
Supercedes https://github.com/devtools-html/debugger.html/pull/6356

PR #6398, opened 2018-05-23
[breakpoints] switch Breakpoint to a PureComponent

jasonLaster:
Fixes Issue: #5875

### Summary of Changes

This PR builds on mappedBreakpoints, which pulled the bp logic into a selector, to finally pull the source logic out as well. The result is that the Breakpoints component is now *only* responsible for showing a list of bps and the poe checkobx. 

The Breakpoint component now knows to re-render when its props change, but that is pretty safe. For example, we used to attempt to re-render a bp whenever a bp was added or removed, now the breakpoints are de-coupled and a bp will only re-render when it changes or one of its other props changes

changes:

* switches to a PureComponent, which no longer needs shouldComponentUpdate
* moves source grouping logic to a selector
* replaces codemirror editor w/ runMode, which does inline syntax highlighting

### Test Plan

Lets add some more unit tests

#### New
screen shot 2018-05-23 at 3 28 52 pm

#### Old
screen shot 2018-05-23 at 3 30 19 pm


jasonLaster:
awesome QAing

jasonLaster:
Great Feedback:

> 1. When refreshing the page, the breakpoint text doesn't always display for the current or other breakpoints:

I believe I fixed the issue where text was not being shown when you re-loaded. The issue was that we were clearing the editor on navigate...

> 2. The complete breakpoint list does not display -- other files and their breakpoints not shown (I believe we are aware of this issue)

not sure what is going on here. is there an STR?

> 3. When paused at a breakpoint, pretty-printing a file and then switching back to the original loses the green left border on the paused breakpoint


will look into the current pause indicator with pretty printing

> 4. Pretty printing allows the breakpoint file to display twice:

with regard to duplicate sources, the underlying issue is that we have 4 different BPs the first two fail to get new original locations when they are remapped. See remapBreakpoints and remapLocations... Given that they are different, i think we should show both sources and all 4 bps.





5. I fixed a minor issue where we would show the bp before we had text, which was a flicker

jasonLaster:
> When paused at a breakpoint, pretty-printing a file and then switching back to the original loses the green left border on the paused breakpoint

Just looked into this and found a bug with comparing locations. I fixed it my using MappedLocation, but it is a good reminder that *everything* needs to be mapped. 

Like the sources bug, the underlying problem here is that not every generated location maps to an original location, so... many of the BPs added before mapping do not map to original locations. As a result, when you switch between pretty and ugly files you'll see the BP only exists in the ugly version.



I think we can attack this problem later with our new `search` param for getOriginalLocation, but that is a bit risky...

darkwing:
Seems that the app file isn't there yet, per the tree:

where

darkwing:
OK, ignore me, looking at the index file now (weird that the index isn't showing in the tree), ToDo is kicked off before App is loaded and run, so it isn't showing the app.js file, which makes sense.  Sorry about my point 2, it's not applicable.

jasonLaster:
hmm, the retries did not do it...

src/selectors/breakpointSources.js
```diff
@@ -0,0 +1,66 @@
...
+  breakpoints: List<Breakpoint>
+}>;
+
+function getBreakpointsForSource(
+  source: SourceRecord,
+  breakpoints: BreakpointsMap
+) {
+  return breakpoints
+    .valueSeq()
+    .filter(bp => bp.location.sourceId == source.id && !bp.hidden && bp.text)
```
darkwing: Should we make this `bp.text || bp.condition`?

Issue #6400, opened 2018-05-23
[Scopes] Tracking issue for original scope mapping tasks


loganfsmyth:
This is a general tracking issue, which replaces #5561 and starts fresh with our current goals.

#### Bugs

* [x] Handle function body scope with lex and nonlex decls #6387
* [x] Ensure that regenerator maps `yield` statements to `return` statements [regenerator issue #342](https://github.com/facebook/regenerator/issues/342) [regenerator PR #344]( https://github.com/facebook/regenerator/pull/344) :runner: 
* [x] Use the original file ranges to more effectively exclude whole-line mappings #6405
* [x] Multiple classes in the same module (Wheel demo)
* [x] `cycleItems` is no longer mapped #6564
* [x] Flow `import { type Foo } from` bindings are included incorrectly #6655
* [x] Flow/TS binding references should be entirely skipped in scope analysis #6655
* [x] Fix bindings incorrectly marked `(optimized out)` #6710 (From code introduced in #6674)
* [x] `this` in the debugger
* [x] create react app source mappings

#### Features 

* [ ] Include original binding data in console autocomplete
* [ ] Verify/add support for "sections" in sourcemaps
* [ ] Handle complex expression mappings better
  * [x] Expose server APIs for calling functions and getters [devtools server PR](https://bugzilla.mozilla.org/show_bug.cgi?id=1473996)
  * [ ]  We should execute module namespace getters to show the real value. Modules generally use getters for live values, rather than side-effects, so it should be within user expectations to trigger those
  * [ ] Some module imports are mapped to function calls instead of property accesses, but we cannot currently call these functions to get their values

#### `source-map`

* [ ] Enhance `applySourceMap` to use range matching instead of closest-matching
* [ ] Expose both relative and absolute paths to allow more easily passing sources from consumers to generators
* [ ] Return additional contexual data from results of `generatedPositionFor` and `originalPositionFor` at least, so users know where they `bias` value took them
* [ ] Consider exposing an async-initialization API for `source-map` that would allow for synchronous consumption once initialized
* [ ] Implement official type definitions for the library
* [ ] Explore if there are internal validation changes that can be made to ensure mappings are valid (line/column always positive, for example)

#### `babel`/`typescript`

* [ ] Explore defining clear rules of _how_ syntax is mapped.
  * Are mappings per-token? 
  * Expressions are complex nested structures, but maps can only represent flat ranges, so what are the expectations there?
  * What are the expectations for whitespace? 
[...]

PR #6405, opened 2018-05-24
[Scopes] Exclude original bindings if the original range is too broad

loganfsmyth:
Refs Issue: #6400

### Summary of Changes

This PR makes it so that original bindings are considered unmapped if their range on the _original_ file contains multiple bindings (except one specific case), or contains a single binding in an unexpected location.

This should help reduce the chance of false positive matches for files that have line mappings, since there are often multiple bindings on a single line.


src/utils/pause/mapScopes/index.js
```diff
@@ -298,12 +314,23 @@ function generateClientScope(
...
 }
 
+function hasValidIdent(range: MappedOriginalRange, pos: BindingLocation) {
+  return (
+    range.type === "match" ||
+    // For declarations, we allow the range on the identifier to be a
+    // more general "contains" to increase the chances of a match.
+    (pos.type !== "ref" && range.type === "contains")
+  );
+}
```
jasonLaster: what is this?

loganfsmyth: I can try to give it a better name if that would help. This makes sure that the original range only contains a single binding, and if it's a reference and not a declaration, it also requires that the binding be at the start of the range.

Issue #5759, opened 2018-03-22
[SourceTree] Missing path segment for resource:// and extension:// URLs


fvsch:
In Firefox Nightly 2018-03-22:

<img width="242" alt="screenshot" src="https://user-images.githubusercontent.com/243601/37796352-61cdeac0-2e17-11e8-8219-3e347a37243a.png">

Sources on the `resource://` and `extension://` protocols are missing the first path segment after the protocol (or is that considered a domain maybe?).

Example with two source URLs:

```
moz-extension://39d86772-227a-3143-96fc-2e1cee5d25ae/content.js
resource://gre/modules/ExtensionContent.jsm
```

Displayed:

```
moz-extension://
    content.js

resource://
    modules
        ExtensionContent.jsm
```

Expected:

```
moz-extension://39d86772-227a-3143-96fc-2e1cee5d25ae
    content.js

resource://gre
    modules
        ExtensionContent.jsm
```

or maybe:

```
moz-extension://
    39d86772-227a-3143-96fc-2e1cee5d25ae
        content.js

resource://
    gre/modules
        ExtensionContent.jsm
```

PS: please disregard the dark boxes in the screen shot, they're probably due to WebRender.

jasonLaster:
Thanks for filing @fvsch. Sorry for the delay. Could you share a good STR?

jasonLaster:
Thanks @fvsch

PR #6417, opened 2018-05-26
Separates plugins and resources by host

lukaszsobek:
Fixes Issue: #5759

Adds separation for plugins and resources displayed in source tree by host:

Screenshot:
screenshot-20180526084059-280x189

### DIFF
--- a/src/utils/sources-tree/getURL.js
+++ b/src/utils/sources-tree/getURL.js
@@ -43,19 +43,19 @@ export function getURL(sourceUrl: string, debuggeeUrl: string = ""): ParsedURL {
       // Ignore `javascript:` URLs for now
       return def;
 
-    case "webpack:":
-      // A Webpack source is a special case
+    case "moz-extension:":
+    case "resource:":
       return merge(def, {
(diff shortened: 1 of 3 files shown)


src/utils/sources-tree/getURL.js
```diff
@@ -43,19 +43,19 @@ export function getURL(sourceUrl: string, debuggeeUrl: string = ""): ParsedURL {
...
-    case "webpack:":
-      // A Webpack source is a special case
+    case "moz-extension:":
+    case "resource:":
       return merge(def, {
-        path: path,
-        group: "webpack://",
-        filename: filename
+        path,
+        group: `${protocol}//${host || ""}`,
```
lukaszsobek: flow complained `host` might be undefined, hence the failsafe.

jasonLaster: looking good. Can we get some more unit tests, maybe for just `getUrl`

lukaszsobek: Sure, on it.

PR #6418, opened 2018-05-26
Moves export statements in front of functions and consts

lukaszsobek:
Tracking functions and constants seems easier with all information in one place.

If we separate out the export statement there is an additional step required to check if the function we are looking at is actually being exported.


lukaszsobek:
browser_dbg-scopes-mutations.js ...

Issue #6403, opened 2018-05-24
[Breakpoints] Extract Exceptions component


jasonLaster:
We're currently refactoring the Breakpoints pane. See #6398. It would be awesome to move the Exceptions logic to its own component as well...

Here is a [gist](https://gist.github.com/jasonLaster/f923e9a473fb0dd258f465035725746c) where I started the cleanup.

PR #6423, opened 2018-05-27
[SecondaryPanel] Makes createExceptionOption a component

lukaszsobek:
Fixes Issue: #6403

- refactors createExceptionOption to be ExceptionOption component
- adds render tests to component

### DIFF
--- a/src/components/SecondaryPanes/Breakpoints/ExceptionOption.js
+++ b/src/components/SecondaryPanes/Breakpoints/ExceptionOption.js
@@ -7,21 +7,15 @@ type ExceptionOptionProps = {
   className: string,
   isChecked: boolean,
   label: string,
-  onChange: Function,
-  shouldRender: boolean
+  onChange: Function
 };
 
 export default function ExceptionOption({
   className,
   isChecked = false,
   label,
-  onChange,
-  shouldRender = true
+  onChange
 }: ExceptionOptionProps) {
-  if (!shouldRender) {
-    return null;
-  }
-
   return (
     <div className={className} onClick={onChange}>
       <input
--- a/src/components/SecondaryPanes/Breakpoints/ExceptionOption.js
+++ b/src/components/SecondaryPanes/Breakpoints/ExceptionOption.js
@@ -1,0 +1,35 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
+import React from "react";
+
+type ExceptionOptionProps = {
+  className: string,
+  isChecked: boolean,
+  label: string,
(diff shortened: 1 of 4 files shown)


src/components/SecondaryPanes/Breakpoints/index.js
```diff
@@ -60,28 +44,28 @@ class Breakpoints extends Component<Props> {
...
-        {exceptionsBox}
-        {shouldPauseOnExceptions ? ignoreCaughtBox : null}
+        <ExceptionOption
+          className="breakpoints-exceptions"
+          label={L10N.getStr("pauseOnExceptionsItem2")}
+          isChecked={shouldPauseOnExceptions}
+          onChange={() => pauseOnExceptions(!shouldPauseOnExceptions, false)}
+        />
+
+        <ExceptionOption
```
jasonLaster: I think a connected component would help us modularize a bit more here:

these props could all move inside of the component
```js
  shouldPauseOnExceptions: boolean,
  shouldPauseOnCaughtExceptions: boolean,
  pauseOnExceptions: Function
```

some day we might want to move this UI and it would be nice if it were self contained

lukaszsobek: Do you want this done now, or make a component now and promote it to a container later?

jasonLaster: up to yo

jasonLaster:
lets test this and land it if it works well

src/components/SecondaryPanes/Breakpoints/ExceptionOption.js
```diff
@@ -0,0 +1,35 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
+import React from "react";
+
+type ExceptionOptionProps = {
+  className: string,
+  isChecked: boolean,
+  label: string,
+  onChange: Function,
+  shouldRender: boolean
```
darkwing: Instead of shouldRender, IMO we should just avoid putting the component in the parent's JSX.  Is there any reason to render it at all if the first checkbox isn't checked?

src/components/SecondaryPanes/Breakpoints/index.js
darkwing: Let's do this the right way the first time to avoid more maintenance down the road.  This isn't a pressing task, though it is important.  Nice work so far @lukaszsobek !

darkwing:
How's it going @lukaszsobek?  Anything we can help with?

src/components/SecondaryPanes/Breakpoints/index.js
jasonLaster: Hmm, on second glance, i think this is fine as is.

jasonLaster:
should be ready to go!

PR #6426, opened 2018-05-28
Unbundle reps.

nchevobbe:
Adds a new script to transpile and move Reps files to mozilla-central.
At some point, it would be nice to simply have an option in copy-modules
to handle both debugger and reps modules.

### DIFF
--- a/.babel/transform-mc.js
+++ b/.babel/transform-mc.js
@@ -5,7 +5,7 @@ const mappings = require("../configs/mozilla-central-mappings");
 
 // Add two additional mappings that cannot be reused when creating the
 // webpack bundles.
-mappings["devtools-reps"] = "devtools/client/shared/components/reps/reps.js";
+mappings["devtools-reps"] = "devtools/client/shared/components/reps/src/index.js";
 mappings["devtools-source-map"] = "devtools/client/shared/source-map/index.js";
 
 function isRequire(t, node) {
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,8 +46,7 @@ install:
   - python --version
   - node --version
   - du -sh firefox
-  - node ./bin/copy-assets.js --mc ./firefox
-  - node ./bin/copy-modules.js --mc ./firefox
+  - node ./bin/copy
   - ./bin/ci/build-firefox.sh
   - node ./bin/ci/check-file-sizes.js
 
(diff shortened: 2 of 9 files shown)


nchevobbe:
Travis is failing because there would need to be some work client side to handle the changes

jasonLaster:
looks great. I think it might be nice to dry up the shared utilities. 


Did you try extending `copy-modules`? Perhaps it might be simpler to run that with an option... If that feels bad, maybe move the utils to a separate file...

nchevobbe:
What do you mean by dry up ? I think it means removing some stuff :) But which one ?
I didn't tried to extend copy-modules yet. I wanted to have something working on its own as a first step. But I agree that this is what we should aim for

darkwing:
Could I get more background on this?  I don't feel comfortable merging this as I don't know the background or goal.

nchevobbe:
> Could I get more background on this? I don't feel comfortable merging this as I don't know the background or goal.

Reps are still bundle in one reps.js giant file. The goal is to have a debugger-like approach where we land individual-transpiled files in mozilla-central so we get more meaningful stacktraces when things go wrong for example.
The secondary goal is that the ObjectInspector now requires "Services" in order to check the platform the user is on. This is causing an issue in the JSON Viewer, which uses Reps to render the json values - but not the ObjectInspector, because its code lives in the content process and uses RequireJS to load modules. This requires some tweaks in order to be able to load some modules, and adding Services to the list would take some time (which we shouldn't, since it won't even be used).
First solution would be to have a dedicated ObjectInspector bundles, but I'd rather totally un-bundle things instead.

You can let this PR like this for now, I need to test if I can re-use copy-modules instead of creating a dedicated script for reps.

nchevobbe:
@jasonLaster I pushed a version where I only use copy-modules with options we pass to it.
What do you think of it ?

nchevobbe:
Also, I'm trying to make this work in m-c, but there's something going wrong :  `
devtools/client/debugger/new/dist/vendors.js, line 1922: ReferenceError: window is not defined` , which points to : 

```js
function inToolbox() {
  return window.parent.document.documentURI == "about:devtools-toolbox";
}
```

I don't know why this is called now and wasn't before

jasonLaster:
it looks good to me

nchevobbe:
> Am i doing something wrong?

No, but this patch needs extensive m-c changes (and github ones too).
Basically the requires need to change in m-c, and the reps file need to be usable by the jsonviewer (which uses requirejs).

I'm not sure what's the best way of doing this.

darkwing:
I see we've hit merge conflicts; should we jump on a call to figure this out, table the discussion?  How should we proceed?

nchevobbe:
This requires more work on mozilla-central and I don't have the bandwidth to do it for now.
Let's close this PR and revisit later.

PR #6430, opened 2018-05-29
[Breakpoints] touch up breakpoint class

jasonLaster:
### Summary of Changes

small tweaks needed to land the new pure component

### DIFF
--- a/src/components/SecondaryPanes/Breakpoints/Breakpoint.js
+++ b/src/components/SecondaryPanes/Breakpoints/Breakpoint.js
@@ -18,9 +18,8 @@ import { CloseButton } from "../../shared/Button";
 import { getLocationWithoutColumn } from "../../../utils/breakpoint";
 
 import { features } from "../../../utils/prefs";
-import { getEditor } from "../../../utils/editor";
+import { getCodeMirror } from "../../../utils/editor";
 
-import type SourceEditor from "../../../utils/editor/source-editor";
 import type {
   Frame,
   Source,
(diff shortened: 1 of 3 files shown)


jasonLaster:
we can land this once 60 goes to central...

src/components/SecondaryPanes/Breakpoints/Breakpoint.js
```diff
@@ -125,18 +122,14 @@ class Breakpoint extends PureComponent<Props> {
 
   highlightText() {
     const text = this.getBreakpointText();
-    const sourceEditor = getEditor();
+    const codeMirror = getCodeMirror();
 
-    if (!text || !sourceEditor || !sourceEditor.editor) {
+    if (!text || !codeMirror) {
       return { __html: "" };
```
darkwing: Due to https://github.com/devtools-html/debugger.html/issues/6439, I'm thinking we should consider showing the escaped text in the case that there's no text or codemirror.

jasonLaster: hmm, i wonder if this was with the first mapped breakpoints PR and not pure components? do we see it now with the latest 61 release in nightly?

jasonLaster: thinking of this 
https://github.com/devtools-html/debugger.html/blob/master/src/selectors/breakpointSources.js#L28-L31

jasonLaster: I see... we should change this strategy so that we just get the CodeMirror class, we don't really need the instance around to do it.

jasonLaster: I think I can fix #6439 in a bit

jasonLaster:
landing this so that we're in sync with MC (this landed there first). now we can move on to #6450 and do it right :tm:

Issue #6441, opened 2018-05-29
Clean up "source.get ? source.get('url') : source.url" Usage


darkwing:
There are a few usages of the following code to get the SourcesTree rework passing:

```js
source.get ? source.get("url") : source.url
```

Most of the work is just test updates.  I've started a branch here:

https://github.com/darkwing/debugger.html/commit/90037cee19d7c0720b06e83ef648017ea6359e23

PR #6443, opened 2018-05-29
Cleanup source tree test updates

darkwing:
Fixes: #6441

We need to get rid of the `source.get ? source.get('url') : source.url` stuff.


darkwing:
@lukaszsobek I know you've worked with the immuatables and getting rid of `get`; is there something I'm missing?

src/utils/sources-tree/addToTree.js
```diff
@@ -169,7 +169,13 @@ export function addToTree(
   debuggeeUrl: string,
   projectRoot: string
 ) {
-  const url = getURL(source.get ? source.get("url") : source.url, debuggeeUrl);
+
+  //console.log(source.url);
+  if (!source.url) {
+    console.log('NOOOOOOO',  source, source.url, source.get("url"));
+  }
+
```
jasonLaster: 😛

src/utils/sources-tree/utils.js
```diff
@@ -52,7 +52,7 @@ export function isNotJavaScript(source: Object): boolean {
 export function isInvalidUrl(url: Object, source: SourceRecord) {
   return (
     IGNORED_URLS.indexOf(url) != -1 ||
-    !(source.get ? source.get("url") : source.url) ||
+    !source.get("url") ||
```
jasonLaster: we should be able to do `source.url` now

lukaszsobek:
btw, according to the docs the dot notation only works for I.records, not for I.maps.

darkwing:
w00t; changing `mergeIn` to `setIn` was the key!

PR #6444, opened 2018-05-30
Bug 1463780 - Set cursor style to element-node rep

dadaa:
### Summary of Changes

* Set pointer cursor to `.objectBox-node` element in case of the element was clickable
* Set pointer cursor to `.open-inspector` element

This changes are for https://bugzilla.mozilla.org/show_bug.cgi?id=1463780

### Test Plan

Please run yarn test

### DIFF
--- a/packages/devtools-reps/src/reps/element-node.js
+++ b/packages/devtools-reps/src/reps/element-node.js
@@ -50,7 +50,8 @@ function ElementNode(props) {
   if (isInTree) {
     if (onDOMNodeClick) {
       Object.assign(baseConfig, {
-        onClick: _ => onDOMNodeClick(object)
+        onClick: _ => onDOMNodeClick(object),
+        className: `${baseConfig.className} clickable`
       });
     }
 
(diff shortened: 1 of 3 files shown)


nchevobbe:
Thanks @dadaa, this looks good.
Should we also add the `clickable` class to the text-node Rep too ?

dadaa:
Hi @nchevobbe !
In animation inspector, we are using only element-node. 
(Also this change affects to FlexboxItem and GridItem[1])
However, if we need, I'll do that! (maybe another PR?). Please let me know if we need.

[1] https://searchfox.org/mozilla-central/search?q=defaultRep%3A+ElementNode%2C&case=false&regexp=false&path=

nchevobbe:
Yes, we should do it for TextNode as well, since they can be displayed in the console/debugger and we should be consistent about how we render similar things (e.g. text node can also be clicked to select them in the inspector).

dadaa:
Ok! Will make the PR for text-node as well.
May I make also the cursor style of text-node to be same?

nchevobbe:
yes, it should be the same

dadaa:
Hi @nchevobbe !
I have looked the text-node. However I think that we could not append `clickable`, since the component does not have `onDOMNodeClick` which I had add last time. 
Otherwise I think that we could not detect whether the element has click event listeners.

Issue #6448, opened 2018-05-30
Pass event object to openLink function when linkifying in reps.


yzen:
Right now event object is not passed when calling openLink (https://github.com/devtools-html/debugger.html/blob/master/packages/devtools-reps/src/reps/string.js#L209). This would be more awesome if it did so we could check for things like meta or crtl keys.

PR #6449, opened 2018-05-30
Pass event object when calling openLink in reps(#6448)

yzen:
Fixes Issue: #6448

### Summary of Changes

* Pass event object when openLink is called on click in the string rep.

### Test Plan

Components that specify openLink will now be able to access click event's properties like (meta/ctrl/shift keys).

### DIFF
--- a/packages/devtools-reps/src/reps/tests/string-with-url.js
+++ b/packages/devtools-reps/src/reps/tests/string-with-url.js
@@ -16,6 +16,21 @@ const renderRep = (string, props) =>
     })
   );
 
+const testLinkClick = (link, openLink, url) => {
+  let syntheticEvent;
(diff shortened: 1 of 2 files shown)


nchevobbe:
It looks like a test (reps/tests/string-with-url.js:462) needs to be updated

packages/devtools-reps/src/reps/tests/string-with-url.js
```diff
@@ -16,6 +16,21 @@ const renderRep = (string, props) =>
     })
   );
 
+const simulateClick = (link, openLink, url) => {
```
nchevobbe: it would be nice to call it "testLinkClick" or something similar since we do the click and also run some assertions related to that click.

Issue #6439, opened 2018-05-29
Breakpoint text not displaying in breakpoints pane


darkwing:
I was tinkering with breakpoints here (https://coherent-potato.glitch.me/) and this occurred:

<img width="1281" alt="wuhhhhh" src="https://user-images.githubusercontent.com/46655/40678577-7c8dd8c0-6346-11e8-8119-338fb220ab07.png">

The text is gone on the right, but breakpoints exist and their title attributes show the proper code.

darkwing:
<img width="1278" alt="whu2" src="https://user-images.githubusercontent.com/46655/40678799-1eb67102-6347-11e8-9d0b-2a2a2a3ddaff.png">

darkwing:
Derp, closing this; the function I modified requires it return a promise.  Nothing to see here...

PR #6450, opened 2018-05-30
[Breakpoints] show before a file is shown

jasonLaster:
Fixes Issue: #6439 

### Summary of Changes

* adds a new CodeMirror field to the SourceEditor so that we can use the class before an editor is intantiated 
* this will need a similar fix in the m-c SourceEditor before the tests pass
* hopefully this simplifies things too

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ env:
   global:
     - DISPLAY=':99.0'
     - YARN_VERSION='0.24.5'
-    - MC_COMMIT='???' # just grab the top sha from this URL --> https://hg.mozilla.org/mozilla-central/shortlog
+    - MC_COMMIT='fa5724780fe7'
 
 notifications:
   slack:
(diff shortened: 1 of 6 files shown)


jasonLaster:
NOTE: we'll likely need to change this file https://searchfox.org/mozilla-central/source/devtools/client/sourceeditor/editor.js when we land the commit in MC. The editor is weeeeird

darkwing:
Added a test update because `editor.CodeMirror.runMode` was choking when `breakpoint.text` and `breakpoint.originalText` was undefined.

darkwing:
What needs to be updated in mc?  The `get CodeMirror` modification?

jasonLaster:
yep - get CodeMirror

jasonLaster:
yep - I addressed this in 62

jasonLaster:
we'll be able to update the travis.yml with the latest version on monday when 62 lands and this should pass

Issue #6384, opened 2018-05-22
[Stability] Implement componentDidCatch to Prevent App Bricking


darkwing:
We have some high risk failure point components but we aren't using `componentDidCatch` (https://reactjs.org/docs/error-boundaries.html) so we risk the debugger completely bricking, which is something we should avoid at all costs.

## Important Components
- [ ] App
    -  Solution:  TBD
- [ ] Editor
    -  Solution:  TBD
- [ ] Preview
    -  Solution:  TBD
- [ ] SourcesTree
    -  Solution:  TBD
- [ ] PrimaryPanes/index.js ?
    -  Solution:  TBD
- [ ] SecondaryPanes/index.js ?
    -  Solution:  TBD

## Ideas
-  Move errors into telemetry or sentry
-  Display error for user?
-  Have different solution based on development mode and production

loganfsmyth:
Closing in favor of https://bugzilla.mozilla.org/show_bug.cgi?id=1589265

PR #6452, opened 2018-05-31
Implement componentDidCatch to prevent complete app breakage

darkwing:
Fixes Issue: #6384

This is a discussion/kickoff PR to implementing overall error handling for the component.  The alternative to `componentDidCatch` (i.e. what we have right now) is an error completely bricking the debugger and nothing being shown in that Firefox panel.  Yuck.

## Questions
-  How does this look as a general rule?
-  How can we send the error data via Telemetry?  (@MikeRatcliffe)

## Testing
-  In any component, add `throw new Error("blah blah")`
-  Refresh the debugger page
-  See something like this:

errorpane

All thoughts welcome

### DIFF
--- a/src/components/App.css
+++ b/src/components/App.css
@@ -28,10 +28,6 @@ button:focus {
     padding: 20px;
 }
 
-.debugger-error details {
-  white-space: pre-wrap;
-}
-
 .editor-pane {
   display: flex;
   position: relative;
--- a/src/components/App.css
+++ b/src/components/App.css
@@ -23,6 +23,15 @@ button:focus {
   height: 100%;
 }
 
+.debugger-error {
+    display: block;
+    padding: 20px;
+}
+
+.debugger-error details {
+  white-space: pre-wrap;
+}
+
 .editor-pane {
   display: flex;
   position: relative;
--- a/src/components/App.js
+++ b/src/components/App.js
@@ -66,7 +66,9 @@ type Props = {
 type State = {
   shortcutsModalEnabled: boolean,
   startPanelSize: number,
-  endPanelSize: number
+  endPanelSize: number,
+  error: string,
+  errorInfo: Object
 };
 
 class App extends Component<Props, State> {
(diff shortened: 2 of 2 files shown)


jasonLaster:
Hmm, I'm not sure if we want to show the error. In the ideal world, we would be able to catch the error, report it, and then recover from it so the user does not notice something went wrong. e.g. the editor is blank, but they can keep working and re-load the tab...

darkwing:
I agree that keeping the app recoverable is ideal but we'll need to add a `componentDidCatch` to every widget to prevent complete bricking.  I was hoping to initially introduce the bare minimum which would tell us what was problematic via telemetry while telling the user that the debugger broke.

I included the error in the debugger because there's not much value in telling the user to go to another devtool to see an error that the debugger already knows about.

jasonLaster:
I see. I'm not sure how I feel about showing an error message in the UI. It feels equivelant to the fail whale in the twitter, which had a bad brand effect. If we go this route, we should pull in Harald and Victoria.

It might be nice to add cDC handlers at the pane level, which will disable a part of the UI, but would let the user recover, or close and re-open the debugger. [commit][commit]
[commit]: https://github.com/jasonLaster/debugger.html/commit/47b094ce36c1f5e35ed173017f6a7f9a12ccc64a

jasonLaster:
re: telemetry, we might need to hook this into Sentry's raven client later when it is better baked

jasonLaster:
CC @digitarald @violasong what do you think?

digitarald:
Wording-wise this needs instructions to get out of this, which is probably recommending closing and re-opening the tools. Link to file an issue would be interesting as well (or a note that we got notified and will look into it; if we have error reporting set up).

Design-wise, Firefox has error pages that can guide the design here (centered).

violasong:
What Harald said sounds good to me. I think the fail whale was damaging to Twitter's brand in those days because it happened all the time, and sometimes it remained broken for minutes/hours at a time 😂 . I'm assuming this type of Debugger failure is a lot more rare.

Design system page on errors: https://design.firefox.com/photon/patterns/errors.html

darkwing:
As discussed in our triage meeting, this should have a broader rule for all of DevTools so I'm tabling this PR for now.

PR #6453, opened 2018-06-01
[components] Feature flag component pane

jasonLaster:
### Summary of Changes

* flag off the component pane
* remove component stack as it will be replaced by component frames in the future


jasonLaster:
we should prioritize landing this soon because it makes things simpler

Issue #6027, opened 2018-04-19
[Sources] reveal in tree doesn’t work with directory roots


jasonLaster:
STR:

1. go to todomvc
2. select `js` as the root in the tree
3. navigate to todo-view.js
4. select reveal in tree

Expected Result: it expands the tree to show the file
Actual Result: it doesnt do anything

darkwing:
Note:

It appears that everything is going through the correct routine but `getDirectories` is returning `Object { name: "root", path: "", contents: […] }`

darkwing:
So this gets us closer:

https://gist.github.com/darkwing/a2feb06e193014c778a5ed87265ab070

The *only* side effect I can find is that trying to set a root under a root under a root doesn't show the sources they should, mostly like because of changing line 127 from `-1` to `-2`

PR #6455, opened 2018-06-01
[tree] switch to relative sources

jasonLaster:
Fixes Issue: #6027

### Summary of Changes

I believe this should fix this use case by re-creating the source tree with relative sources. There are a couple of up-sides:

1. perf - the work is done in the selector so we no longer need to figure out what goes in and out of the tree
2. simplicity - i started removing projectRootUrl from all of the tree code, which is a great cleanup

### Testing

yep, we should write some more mochitests here

### DIFF
--- a/src/components/PrimaryPanes/SourcesTree.js
+++ b/src/components/PrimaryPanes/SourcesTree.js
@@ -18,7 +18,7 @@ import {
   getDebuggeeUrl,
   getExpandedState,
   getProjectDirectoryRoot,
-  getSources
+  getRelativeSources
 } from "../../selectors";
 
 // Actions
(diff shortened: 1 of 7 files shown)


darkwing:
Looking at this, I think the current test snapshot value is clearly incorrect.  Also, however, I don't believe a quick `yarn jest -u` is correct either, because I'm seeing `resource://` there.  Looking more into this.  A manual test works great though

jasonLaster:
yep - didnt look at any of the tests. threw this together fairly quickly as a proof of concept

darkwing:
That test in question failed because the sources handed to addToTree are pre-filtered, thus it's going to return everything it was given.

wldcordeiro:
👍 Looks very nice

PR #6465, opened 2018-06-02
WatchExpressions: Added UI for Autocomplete

marcogallo566:
Issue no: #5467

### Summary of Changes

* Added UI 

I'm stuck with a problem that is visible in the following GIF: the results are given too late, there are no results until I delete the expression. I discovered that the following method
````
 findAutocompleteMatches = debounce(target => { 
    this.props.autocomplete(target.value, target.selectionStart); 
  }, 250);
````
remains pending.

Feel free to improve it!

### GIF
7cda2agnt8

## TODO:
- [ ] Fix bugs found
- [ ] Enable this functionality behind a flag

### CODE REVIEW

**Comment by darkwing** [comment]
Great start @marcogallo566 !  One quick thing I noticed:

letters

When emptying the box, the properties that start with the last letter display; we should either show all possible properties or none when the input is "" (empty)

File: src/components/SecondaryPanes/Expressions.js [hunk 1]
```diff
@@ -225,6 +239,21 @@ class Expressions extends Component<Props, State> {
...
+    if(autocompleteMatches) {
+      return (
+        <datalist id="autocomplete-matches">
+          {autocompleteMatches.map((match, index) => {
+            return <option key={index} value={match} />
+          })}
+        </datalist>
+      );
+    } else {
+      return <datalist id="autocomplete-matches" />;
```
darkwing: Should we just render `null` if no matches are found?

bomsy:
Hey @marcogallo566 i think i found the bug. Fix this in `Expression.js`
```diff
@@ -88,11 +88,12 @@ class Expressions extends Component<Props, State> {

   shouldComponentUpdate(nextProps, nextState) {
     const { editing, inputValue, focused } = this.state;
-    const { expressions, expressionError, showInput } = this.props;
+    const { expressions, expressionError, showInput, autocompleteMatches } = this.props;

     return (
       expressions !== nextProps.expressions ||
       expressionError !== nextProps.expressionError ||
+      autocompleteMatches !== nextProps.autocompleteMatches ||
       editing !== nextState.editing ||
       inputValue !== nextState.inputValue ||
       nextProps.showInput !== showInput ||
@@ -138,10 +139,9 @@ class Expressions extends Component<Props, State> {
     this.findAutocompleteMatches(target);
   };

-  findAutocompleteMatches = debounce(target => {
-    console.log(this.props.autocompleteMatches);
+  findAutocompleteMatches(target) {
     this.props.autocomplete(target.value, target.selectionStart);
-  }, 250);
+  }
```

bomsy:
Also you might need to do a rebase.

marcogallo566:
@darkwing There was an error about a pending promise, this was the cause of the bug.
Thank you @bomsy but @giatorta just resolved it.
Yes I'll rebase soon.

src/components/SecondaryPanes/Expressions.js
```diff
@@ -125,9 +133,24 @@ class Expressions extends Component<Props, State> {
...
-    this.setState({ inputValue: e.target.value });
+    const target = e.target;
+    this.findAutocompleteMatches(target);
   };
 
+  findAutocompleteMatches = debounce(target => {
+    this.props.autocomplete(target.value, target.selectionStart);
+  }, 250);
+
+  findAutocompleteMatches = (target) => {
```
bomsy: `findAutocompleteMatches = async (target) => {`

src/components/SecondaryPanes/Expressions.js
```diff
@@ -125,9 +133,24 @@ class Expressions extends Component<Props, State> {
...
+    const value = target.value;
+    const prom = this.props.autocomplete(target.value, target.selectionStart);
+    prom.then(() => {this.helper(value);});
```
bomsy: Nice work! we can try this aswell
```js
 const input  = await this.props.autocomplete(target.value, target.selectionStart);
this.setState({ inputValue: input })
```

bomsy:
Ah nice one @marcogallo566 . Sorry i didnt seee the fix well done!

marcogallo566:
@bomsy Thank you so much, now it works fine!
Yes, I also think it's one step behind.
Now I'll work for the remaining issues, I found that if I click on watch expression the autocomplete appears without any expression typed.

isibaxqrlj

darkwing:
Great work so far @marcogallo566 and @bomsy!  Could you add a "ToDo" section in the PR description so we can all track the scope of this PR?

Thank you!

marcogallo566:
@bomsy What do you mean with "Enable this functionality behind a flag"?
An option to activate/deactivate the autocomplete?

The last fix is made in the right way?

darkwing:
@marcogallo566 He means that we should keep this disabled until we've done much more user testing and iron out some bugs; this will allow us to continue working on the functionality without exposing users to broken stuff.

marcogallo566:
Ah now I understand. I'll implement it soon.

bomsy:
@marcogallo566 This would help in creating a pref flag  to enable  autocomplete behind.

https://github.com/devtools-html/debugger.html/blob/master/docs/local-development.md#prefs

The pref could be
```js
pref("devtools.debugger.features.autocomplete", false);
```

wldcordeiro:
@bomsy I'd go with `pref("devtools.debugger.features.autocomplete-expressions", false)` just to be a bit more specific if we add additional autocomplete features.

src/components/SecondaryPanes/Expressions.js
```diff
@@ -256,8 +254,7 @@ class Expressions extends Component<Props, State> {
 
   renderAutoCompleteMatches() {
     const { autocompleteMatches } = this.props;
-    const { enableAutocomplete } = this.state;
-    if (autocompleteMatches && enableAutocomplete) {
+    if (autocompleteMatches && features.autocompleteExpression) {
```
wldcordeiro: `(features.autocompleteExpression && autocompleteMatches)` would be slight better just so it fails immediately when it isn't enabled.

wldcordeiro:
This is looking great! I'm loving this addition of an autocomplete.

src/components/SecondaryPanes/Expressions.js
```diff
@@ -125,9 +140,18 @@ class Expressions extends Component<Props, State> {
   }
 
   handleChange = (e: SyntheticInputEvent<HTMLInputElement>) => {
-    this.setState({ inputValue: e.target.value });
+    const target = e.target;
```
wldcordeiro: Slight improvement here since `findAutocompleteMatches` is a redux action you're debouncing let's do this and give it a slight head start on the react state's update.

```js
this.setState(() => {
  const { target } = e;
  if (features.autocompleteExpression) {
    this.findAutocompleteMatches(target.value, target.selectionStart);
  }
  return { inputValue: target.value }
});
```

src/components/SecondaryPanes/Expressions.js
```diff
@@ -226,6 +252,20 @@ class Expressions extends Component<Props, State> {
     );
   };
 
+  renderAutoCompleteMatches() {
+    const { autocompleteMatches } = this.props;
+    if (autocompleteMatches && features.autocompleteExpression) {
```
wldcordeiro: Feature check first here too.

wldcordeiro:
```jsx
renderAutoCompleteMatches() {
  const { autocompleteMatches } = this.props;
  if (!features.autocompleteExpression) {
    return null;
  }
  if (autocompleteMatches) {
    return (
      <datalist id="autocomplete-matches">
        {autocompleteMatches.map((match, index) => {
           return <option key={index} value={match} />;
         })}
      </datalist>
    );
  }
  return <datalist id="autocomplete-matches" />;
}
```

src/utils/prefs.js
```diff
@@ -55,6 +55,7 @@ if (isDevelopment()) {
   pref("devtools.debugger.features.pause-points", true);
   pref("devtools.debugger.features.component-pane", false);
   pref("devtools.debugger.features.skip-pausing", false);
+  pref("devtools.debugger.features.autocomplete-expressions", false);
```
jasonLaster: this will also need to be in `assets/panel/prefs` devtools.debugger.features.autocomplete-expressions

src/components/SecondaryPanes/Expressions.js
marcogallo566: I tried this solution, but I learned now that a synthetic event is pooled. Can I write `e.persist()`? I read that the event is pooled for performance reasons and I'm not sure that I can do it.

wldcordeiro:
I'm gonna give this a deeper look today and see if I can get CI passing. Thanks for your work @marcogallo566 it's a great addition. :smile:

marcogallo566:
@wldcordeiro Thank you for your help!

src/components/SecondaryPanes/Expressions.js
wldcordeiro: Hmm yeah let's avoid the event pooling for now. We could still put the call to `findAutocompleteMatches` before `setState`. Thanks for following up on that.

src/components/SecondaryPanes/Expressions.js
```diff
@@ -125,9 +140,18 @@ class Expressions extends Component<Props, State> {
...
+    if (features.autocompleteExpression) {
+      this.findAutocompleteMatches(target.value, target.selectionStart);
+    }
+    this.setState({ inputValue: target.value });
   };
 
+  findAutocompleteMatches = debounce((value, selectionStart) => {
+    const { autocomplete } = this.props;
+    autocomplete(value, selectionStart);
+  }, 250);
```
wldcordeiro: Minor note, maybe 250 is a little high? I think we debounce ~150ms elsewhere.

bomsy:
Nice one! @marcogallo566

Issue #6478, opened 2018-06-05
Can't right click, select text or navigate links in Scopes


digitarald:
Trying to copy variable states for filings bugs is impossible, as users can't select text or have the rich right click that Console offers. Links are highlighted but clicking left and right doesn't do anything.

![screenflow](https://user-images.githubusercontent.com/8599/41000506-14fabc5a-68c3-11e8-8452-59cd8dccea49.gif)

PR #6479, opened 2018-06-05
Allow for clicking and copying scopes information

darkwing:
Fixes Issue: #6478

Allows users to copy scope properties and values.  This declaration was overly broad.

## Testing
- Set a breakpoint, refresh your source page, wait for breakpoint to hit
- Hover over and copy values in the Scopes' pane tree

## Bonus
- @digitarald will be very happy when this is in


darkwing:
Have a look when you get a chance @bomsy @wldcordeiro @juliandescottes

Issue #6331, opened 2018-05-17
[Symbols] should recognize React.Component as a parent class


jasonLaster:
We miss `React.Component` because it is a member expresssion

```js
import React from 'react';

export default class UnorderedList extends React.Component {
  render() {  
    const {items, highlight} = this.props;
    return (
      <ul>
        {items.map((item, i) => {
          return <li key={i} className={i === highlight ? 'highlighted' : null}>{item}</li>;
        })}
      </ul>
    );
  }
}


// WEBPACK FOOTER //
// ./app/components/UnorderedList.jsx
```

Here is a patch to help get started. The trick is to get the name if it is not easy to find. and a small clean up to just return relevant fields

```diff
diff --git a/src/workers/parser/getSymbols.js b/src/workers/parser/getSymbols.js
index cde1021..8868635 100644
--- a/src/workers/parser/getSymbols.js
+++ b/src/workers/parser/getSymbols.js
@@ -16,7 +16,8 @@ import {
   getObjectExpressionValue,
   getVariableNames,
   getComments,
-  getSpecifiers
+  getSpecifiers,
+  getCode
 } from "./utils/helpers";

 import { inferClassName } from "./utils/inferClassName";
@@ -143,9 +144,15 @@ function extractSymbol(path: SimplePath, symbols) {
   }

   if (t.isClassDeclaration(path)) {
+    const { name, loc, superClass } = path.node;
     symbols.classes.push({
       name: path.node.id.name,
-      parent: path.node.superClass,
+      parent: {
+        name: t.isMemberExpression(superClass)
+          ? getCode(superClass)
+          : superClass.name,
+        location: superClass.loc
+      },
       location: path.node.loc
     });
   }
diff --git a/src/workers/parser/utils/helpers.js b/src/workers/parser/utils/helpers.js
index 0d86919..2abc00d 100644
--- a/src/workers/parser/utils/helpers.js
+++ b/src/workers/parser/utils/helpers.js
@@ -61,6 +61,10 @@ export function getObjectExpressionValue(node: Node) {
   return shouldWrap ? `(${code})` : code;
 }

+export function getCode(node: Node) {
+  return generate(node).code;
+}
+
 export function getVariableNames(path: SimplePath): SymbolDeclaration[] {
   if (t.isObjectProperty(path.node) && !isFunction(path.node.value)) {
     if (path.node.key.type === "StringLiteral") {
     
[...]

PR #6481, opened 2018-06-06
[Preview] Recognize React.Component as a parent class.

adancarrasco:
Fixes Issue: #6331 

### Summary of Changes

* Pretty much followed @jasonLaster comments on the issue. 

### Test Plan
- Open https://firefox-debugger-example-react-js.glitch.me/ debugger test page
- Select parse-worker from workers
- Command-P opens the panel and look for getSymbols.js
- Put a breakpoint in line 148
- Reload the target debug page
- It'll hit the breakpoint, then step over
- `symbols.classes` array has the `UnorderedList` object with the parent as shown in the screenshot

### Screenshots/Videos
Object in parser-worker.js, parent added as React.Component
image

### DIFF
--- a/src/actions/tests/__snapshots__/preview.spec.js.snap
+++ b/src/actions/tests/__snapshots__/preview.spec.js.snap
@@ -46,7 +46,14 @@ exports[`setPreview react instance 1`] = `
 Object {
   "cursorPos": undefined,
   "expression": "this",
-  "extra": Object {},
+  "extra": Object {
+    "react": Object {
+      "componentStack": Array [
+        "Foo",
+      ],
+      "displayName": "FixtureComponent",
+    },
+  },
   "location": Object {
     "end": Object {
       "column": 18,
--- a/src/actions/tests/__snapshots__/preview.spec.js.snap
+++ b/src/actions/tests/__snapshots__/preview.spec.js.snap
@@ -46,14 +46,7 @@ exports[`setPreview react instance 1`] = `
 Object {
   "cursorPos": undefined,
   "expression": "this",
-  "extra": Object {
-    "react": Object {
-      "componentStack": Array [
-        "Foo",
-      ],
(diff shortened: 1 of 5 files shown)


src/actions/tests/__snapshots__/preview.spec.js.snap
```diff
@@ -46,14 +46,7 @@ exports[`setPreview react instance 1`] = `
 Object {
   "cursorPos": undefined,
   "expression": "this",
-  "extra": Object {
```
adancarrasco: After I updated `inComponent.js` this got empty. I don't think this should be the right snapshot. I'll investigate later.

darkwing: Yep -- that's an issue.

src/workers/parser/getSymbols.js
```diff
@@ -33,7 +34,7 @@ export type SymbolDeclaration = {
 };
 
 export type ClassDeclaration = SymbolDeclaration & {
-  parent: string
+  parent: Object | null
```
jasonLaster: we should improve the flow type here:

```
name: string,
location: Location
```

adancarrasco: Thanks for the feedback Jason, I would work on it this week. :)

src/actions/tests/__snapshots__/preview.spec.js.snap
adancarrasco: Thanks @darkwing. I will review it this week. :)

darkwing:
How's it going @adancarrasco ?  Anything we can help with?

jasonLaster:
Thanks @adancarrasco!

adancarrasco:
Thank you guys, sorry for the delay. I was off and couldn't make it as promised.

Issue #6355, opened 2018-05-19
[Breakpoints] breakpoint does not appear on reload


jasonLaster:
It's pretty hard to reproduce, but there is a scenario where after you reload and go to the file with a breakpoint you can see it in the pane, but not in the editor

![Screen Shot 2018-05-19 at 9.24.39 AM.png](https://shipusercontent.com/1dafac1fe18f66f6cb2fefea63e33971/Screen%20Shot%202018-05-19%20at%209.24.39%20AM.png)

bomsy:
i might have and STR that works
-  debug http://firefox-dev.tools/debugger-examples/examples/todomvc/
- open `backbone.js` in the debugger
- set breakpoint on line 12
- close `backbone.js`
- refresh the debugger
- click to open `backbone.js` from the source tree

AR:
No breakpoint set on line 12 in the editor

ER:
Breakpoint should be set on line 12

PR #6486, opened 2018-06-07
allow update always

bomsy:
Fixes Issue: #6355

### Summary of Changes

* remove `shouldComponentUpdate` in Breakpoint.js

If the editor is not ready before the breakpoint tries to render, the breakpoint never gets shown.
`shouldComponentUpdate` stops the breakpoint from updating after the editor is ready.

**Note:** One concern was the possible performance hit, but i don't think that is an issue because it still feels instant, as i think react figures out how to not re-render the breakpoints. see (after screenshot)

### Test Plan

-  debug http://firefox-dev.tools/debugger-examples/examples/todomvc/
- open `backbone.js` in the debugger
- set breakpoint on line 12
- close `backbone.js`
- refresh the debugger
- click to open `backbone.js` from the source tree

AR:
No breakpoint set on line 12 in the editor

ER:
Breakpoint should be set on line 12

#### Before
On reload & reopening `backbone.js`  the breakpoints do not show up.

#### After
On reload & reopening `backbone.js` the breakpoints now show up.

### DIFF
--- a/src/components/Editor/Breakpoint.js
+++ b/src/components/Editor/Breakpoint.js
@@ -71,18 +71,6 @@ class Breakpoint extends Component<Props> {
     }
   };
 
-  shouldComponentUpdate(nextProps: any) {
-    const { editor, breakpoint, selectedSource } = this.props;
-    return (
-      editor !== nextProps.editor ||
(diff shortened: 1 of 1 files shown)


darkwing:
This seems reasonable; if we see gross side effects, we can back out easily.

PR #6488, opened 2018-06-07
[SourceTree] Don't blur when tree is no longer focused

darkwing:
One thing that bugs me is when I click an item in the source tree and then navigate over to the editor, the focused source's highlight disappears.  IMO it shouldn't -- the focused file should always be highlighted in the tree.

### DIFF
--- a/packages/devtools-components/src/tree.js
+++ b/packages/devtools-components/src/tree.js
@@ -341,7 +341,9 @@ class Tree extends Component {
       // Additional classes to add to the root element.
       className: PropTypes.string,
       // style object to be applied to the root element.
-      style: PropTypes.object
+      style: PropTypes.object,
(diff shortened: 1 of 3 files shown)


jasonLaster:
Nice

packages/devtools-components/src/tree.js
```diff
@@ -341,7 +341,9 @@ class Tree extends Component {
       // Additional classes to add to the root element.
       className: PropTypes.string,
       // style object to be applied to the root element.
-      style: PropTypes.object
+      style: PropTypes.object,
+      // Prevents blur when Tree loses focus
+      preventBlur: PropTypes.bool
```
wldcordeiro: Maybe add a default prop value for this? You're just checking the negation so not having one should be fine but it's good to have a default.

Issue #5918, opened 2018-04-10
SourcesTree: (index) click weirdness


darkwing:
The first time I click on "(index)" in the SourcesTree, the following item is focused:

```js
// console.log'ing from inside `focusItem()`
Object { name: "(index)", path: "https://davidwalsh.name/", contents: {…} }
Object { name: "(index)", path: "https://davidwalsh.name/", contents: {…} }
Object { name: "root", path: "", contents: […] }
```

If I click it again, I get:

```js
Object { name: "(index)", path: "https://davidwalsh.name/", contents: {…} }
Object { name: "(index)", path: "https://davidwalsh.name/", contents: {…} }
```

The `Object { name: "root", path: "", contents: […] }` item appears to be what throws the highlight off.  We need to figure out where that's coming from to fix this bug.

darkwing:
The problematic `Object { name: "root", path: "", contents: […] }` item is coming from `ManagedTree`:

```js
highlightItem(highlightItems: Array<Item>) {
  // ....

  //... the last line:
  this.focusItem(highlightItems[index]);
}
```

janodvarko:
@darkwing How can I reproduce this issue? STR?

Honza

PR #6491, opened 2018-06-08
[SourceTree]  Prevent index weirdness

### LINKED ISSUE #6462: [SourcesTree] Left Arrow Key Sometimes Loses Focused/Highlight State
Expanding and contracting tree items works really well but occasionally hitting LeftArrow on a source fails to highlight its parent directory.  I've painfully looked at this issue numerous times over the last months and I've found that the Tree component only calls `renderItem` (from within the `render` method) on one item, which is the highlighted source, but doesn't on the (should-be) highlighted directory;  because the highlighted directory isn't re-rendered via `renderItem`, the `focused` CSS class is not applied.

STR:

1.  Go to https://perf-html.io/
2.  In the sources tree, click the www.google-analytics.com directory
3.  Down arrow to "analytics.js"
4.  Hight left arrow key

analytics.js is no longer highlighted (Yay) but www.google-analytics.com is not

@nchevobbe Can you provide any insight?  I don't think it's Debugger's ManagedTree/SourcesTree causing the problem.  I get the feeling something may need to be changed in this:

```js
const indents = Array.from({ length: depth }).fill(treeIndent);
    const items = indents.concat(
      renderItem(item, depth, focused, arrow, expanded)
    );
```

Because the parent dir isn't being re-rendered.

darkwing:
Fixes Issue: #5918

This `else` clause is causing mischief in the SourcesTree, and I don't see its value, so removing.

## Testing
2.  Click on "(index)"; the item should highlight
3.  Expand `ad.doubleclick.net` and its subfolders; click on the oddly named files; they should also highlight

## Note
1.  This also seems to fix #6462

### DIFF
--- a/src/components/shared/ManagedTree.js
+++ b/src/components/shared/ManagedTree.js
@@ -127,8 +127,14 @@ class ManagedTree extends Component<Props, State> {
       // closed folder and highlights this folder
       const index = highlightItems
         .reverse()
-        .findIndex(item => !expanded.has(this.props.getPath(item)));
-      this.focusItem(highlightItems[index]);
+        .findIndex(
+          item =>
+            !expanded.has(this.props.getPath(item)) && item.name !== "root"
+        );
+
+      if (highlightItems[index]) {
+        this.focusItem(highlightItems[index]);
+      }
     }
   }
 
--- a/src/components/test/SourcesTree.spec.js
+++ b/src/components/test/SourcesTree.spec.js
@@ -195,6 +195,17 @@ describe("SourcesTree", () => {
         .simulate("keydown", { keyCode: 13 });
       expect(props.selectSource).toHaveBeenCalledWith(item.contents.id);
     });
+
+    it("allows focus on the (index)", async () => {
+      const { component, instance, props } = render();
+      const item = createMockItem("https://davidwalsh.name/", "(index)");
+      await instance.focusItem(item);
+      await component.update();
+      await component
+        .find(".sources-list")
+        .simulate("keydown", { keyCode: 13 });
+      expect(props.selectSource).toHaveBeenCalledWith(item.contents.id);
+    });
   });
 
   describe("with custom root", () => {
--- a/src/utils/sources-tree/addToTree.js
+++ b/src/utils/sources-tree/addToTree.js
@@ -85,12 +85,10 @@ function findOrCreateNode(
  * adding new nodes along the way
  */
 function traverseTree(
-  url: Object,
+  url: ParsedURL,
   tree: TreeDirectory,
   debuggeeHost: ?string
 ): TreeNode {
-  url.path = decodeURIComponent(url.path);
-
   const parts = url.path.split("/").filter(p => p !== "");
   parts.unshift(url.group);
 
--- a/src/utils/sources-tree/tests/__snapshots__/addToTree.spec.js.snap
+++ b/src/utils/sources-tree/tests/__snapshots__/addToTree.spec.js.snap
@@ -35,10 +35,26 @@ exports[`sources-tree addToTree does not attempt to add two of the same file 1`]
 "
 `;
 
+exports[`sources-tree addToTree does not mangle encoded URLs 1`] = `
+" - root path= 
+  - example.com path=example.com 
+    - foo path=example.com/foo 
+      - B9724220.131821496;dc_ver=42.111;sz=468x60;u_sd=2;dc_adk=2020465299;ord=a53rpc;dc_rfl=1,https%3A%2F%2Fdavidwalsh.name%2F$0;xdt=1 path=example.com/foo/B9724220.131821496;dc_ver=42.111;sz=468x60;u_sd=2;dc_adk=2020465299;ord=a53rpc;dc_rfl=1,https%3A%2F%2Fdavidwalsh.name%2F$0;xdt=1 source_id=undefined 
+"
+`;
+
 exports[`sources-tree addToTree excludes javascript: URLs from the tree 1`] = `
 " - root path= 
   - example.com path=example.com 
     - source1.js path=example.com/source1.js source_id=undefined 
+"
+`;
+
+exports[`sources-tree addToTree name does not include query params 1`] = `
+" - root path= 
+  - example.com path=example.com 
+    - foo path=example.com/foo 
+      - name.js path=example.com/foo/name.js source_id=undefined 
 "
 `;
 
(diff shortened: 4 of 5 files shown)


AnshulMalik:
I think the else block is important with current implementation.

Consider this directory structure:

- root
  - dir1(expanded)
    - file1
  - dir2(collapsed)
    - file 2
    - file 3
  
So the tree would look like
- root
  - dir1
    - file1
  - dir2

Now if we want to highlight file 3, it's not rendered yet, so we try to find it's parent which is visible and highlight that in the else statement.

src/components/shared/ManagedTree.js
```diff
@@ -122,13 +122,6 @@ class ManagedTree extends Component<Props, State> {
...
     if (expanded.has(this.props.getPath(highlightItems[0]))) {
       this.focusItem(highlightItems[0]);
-    } else {
-      // Look at folders starting from the top-level until finds a
-      // closed folder and highlights this folder
-      const index = highlightItems
-        .reverse()
-        .findIndex(item => !expanded.has(this.props.getPath(item)));
-      this.focusItem(highlightItems[index]);
     }
```
jasonLaster: hmm, this might benefit from some unit tests that document the behavior. I honestly don't really understand what it is doing. for instance what is highlightItems?

jasonLaster:
CC @arthur801031 @bomsy

darkwing:
@AnshulMalik Interesting.  Unfortunately it's returning `{ name: "root" ... }` for `index` and similarly oddly named items, which is making the highlight disappear.

jasonLaster:
Hmm @darkwing can you elaborate?

darkwing:
@jasonLaster The reason the tree highlight/focus disappears is because that `reverse.findIndex`'s result `0` is "root" from `const uncollapsedTree = createNode("root", "", []);` ; that has no visual representation and thus the focus disappears.

darkwing:
Restored the `else` clause and ensured `root` is not acceptable

AnshulMalik:
Umm, I did some digging here, and found that the actual problem is in `getDirectories`.

I found that the highlightItems which is used to represent the path of the item to be highlighted, is not being calculated right.

For some cases, when the url we are looking for doesn't math any of the source urls, `getDirectories` returns `root` node, which gets assigned to highlightItems and in turn leads to return of `-1`.

Why the Url doesn't match when there is a source with the url, 
In `getDirectories` -> `findSource`, we are checking for `subtree.path == sourceUrl`

For one of the oddly named files, we get,

SourceUrl: `ad.doubleclick.net/ddm/adj/N366601.452584BUYSELLADS.COM/B9724220.131821496;dc_ver=42.111;sz=468x60;u_sd=2;dc_adk=2020465299;ord=wr6azv;dc_rfl=1,https%3A%2F%2Fdavidwalsh.name%2F$0;xdt=1;crlt=A!%27N55Evr-;sttr=109;prcl=s`

subtree.path: ad.doubleclick.net/ddm/adj/N366601.452584BUYSELLADS.COM/B9724220.131821496;dc_ver=42.111;sz=468x60;u_sd=2;dc_adk=2020465299;ord=wr6azv;dc_rfl=1,https:/davidwalsh.name/$0;xdt=1;crlt=A!'N55Evr-;sttr=109;prcl=s

There is some issue with url encoding, let's use decodeURIComponent
decodeURIComponent(sourceUrl): 
`ad.doubleclick.net/ddm/adj/N366601.452584BUYSELLADS.COM/B9724220.131821496;dc_ver=42.111;sz=468x60;u_sd=2;dc_adk=2020465299;ord=wr6azv;dc_rfl=1,https://davidwalsh.name/$0;xdt=1;crlt=A!'N55Evr-;sttr=109;prcl=s`


Now still they don't match, because of an extra `/` in sourceUrl (https://davidwalsh.name/$0)

darkwing:
Furthering this, the item path is stored incorrectly deeper in the tree; 

`ad.doubleclick.net/ddm/adj/N366601.452584BUYSELLADS.COM/B9724220.131821496;dc_ver=42.111;sz=468x60;u_sd=2;dc_adk=2020465299;ord=wr6azv;dc_rfl=1,https%3A%2F%2Fdavidwalsh.name%2F$0;xdt=1;crlt=A!%27N55Evr-;sttr=109;prcl=s`

vs.

oneslash

It could be as easy as simple splitting by slashes that broke it.

darkwing:
OK, so the root issue is the following line in `addToTree.js`'s `traverseTree` function:

```js
const parts = url.path.split("/").filter(p => p !== "");
```

The `.filter(p => p !== "")` piece ruins the integrity of the item's URL.

jasonLaster:
oh nice find!

jasonLaster:
here's where we added decodeURIComponent... https://github.com/devtools-html/debugger.html/commit/22dbc958a11391e8110261891acb04d2ceb1f19b#diff-fbb5b5145e272c781d90834c20b1d404R79

I don't remember why we needed this, but maybe @jbhoosreddy would have some recollection. 

CC @codehag

darkwing:
I looked at a blame there but didn't see a compelling reason for the decoding so we'll see how Travis pans out.

darkwing:
Tests passing but welcome to hear feedback about the removal of URL decoding.

AnshulMalik:
Yep, looks really good. But now I think we probably don't need `item.name !== "root"`.

darkwing:
@AnshulMalik We do want to keep that because, without it, we can still receive `Object { name: "root", path: "", contents: (8) […] }`, which causes more problems; I'd rather we didn't try to highlight something that visually doesn't exist.

AnshulMalik:
I am afraid it might hide the actual issue which is causing this behaviour, like the case of url encoding.

darkwing:
@AnshulMalik Sure but I'd prefer not to keep this process going; we can do that in a follow up.

src/components/shared/ManagedTree.js
```diff
@@ -127,8 +127,14 @@ class ManagedTree extends Component<Props, State> {
...
-        .findIndex(item => !expanded.has(this.props.getPath(item)));
-      this.focusItem(highlightItems[index]);
+        .findIndex(
+          item =>
+            !expanded.has(this.props.getPath(item)) && item.name !== "root"
+        );
+
+      if (highlightItems[index]) {
+        this.focusItem(highlightItems[index]);
+      }
```
jasonLaster: We now have lots of good tests for SourcesTree
https://github.com/devtools-html/debugger.html/blob/master/src/components/PrimaryPanes/SourcesTree.js#L184

It might be nice to add a new passing test for the non-root case and perhaps add a failing test for the root case so that we have documented this behavior

PR #6485, opened 2018-06-06
Fix #6451 - Ensure breakpoint hover is always correct

darkwing:
Fixes Issue: #6451

When you scroll the editor horizontally and then hover a line, the breakpoint hover arrow doesn't display because we aren't allowing the `left` property to change.  Setting a negative margin works much better.

## Testing
1.  Hover over a line, ensure the breakpoint arrow displays
2.  Scroll horizontally, hover over a line, ensure the breakpoint arrow displays

PR #6492, opened 2018-06-08
Reverts 6485; Restore breakpoint arrow calculation

darkwing:
While my #6485 works well for most JS files, it doesn't look good in WASM, so we should revert and try again.


darkwing:
Easy source to use:  https://mdn.github.io/webassembly-examples/js-api-examples/fail.html

jasonLaster:
it might also be worth checking on files with 1000+ lines too

PR #6496, opened 2018-06-11
Tests for utils/editor/get-expressions

kenjiO:
Here is a draft of some tests.  Can you let me know how they look so far?

### DIFF
--- a/src/utils/editor/tests/get-expression.spec.js
+++ b/src/utils/editor/tests/get-expression.spec.js
@@ -2,9 +2,32 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
 
+import CodeMirror from "codemirror";
 import { getExpressionFromCoords } from "../get-expression";
 
 describe("get-expression", () => {
+  let isCreateTextRangeDefined;
+
+  beforeAll(() => {
+    if (document.body.createTextRange) {
+      isCreateTextRangeDefined = true;
+    } else {
+      isCreateTextRangeDefined = false;
+      // CodeMirror needs createTextRange
+      // https://discuss.codemirror.net/t/working-in-jsdom-or-node-js-natively/138/5
+      document.body.createTextRange = () => ({
+        getBoundingClientRect: () => {},
+        getClientRects: () => ({})
+      });
+    }
+  });
+
+  afterAll(() => {
+    if (!isCreateTextRangeDefined) {
+      delete document.body.createTextRange;
+    }
+  });
+
   describe("getExpressionFromCoords", () => {
     it("returns null when location.line is greater than the lineCount", () => {
       const lineCount = 1;
@@ -105,21 +128,12 @@ describe("get-expression", () => {
     });
 
     it("includes previous tokens in the expression", () => {
-      const codemirrorMock = {
-        lineCount: () => 2,
-        getTokenAt: location => {
-          if (location.ch >= 5 && location.ch <= 7) {
-            return { start: 4, end: 7, type: "property" };
-          } else if (location.ch >= 1 && location.ch <= 3) {
-            return { start: 0, end: 3, type: "variable" };
-          }
-        },
-        doc: {
-          getLine: () => "foo.bar;"
-        }
-      };
+      const cm = CodeMirror(document.body, {
+        value: "foo.bar;\n",
+        mode: "javascript"
+      });
 
-      const result = getExpressionFromCoords(codemirrorMock, {
+      const result = getExpressionFromCoords(cm, {
         line: 1,
         column: 5
       });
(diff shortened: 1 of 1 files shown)


src/utils/editor/tests/get-expression.spec.js
```diff
@@ -0,0 +1,183 @@
...
+      const result = getExpressionFromCoords(codemirrorMock, {
+        line: 1,
+        column: 1
+      });
+      expect(result.expression).toEqual("foo");
+      expect(result.location.start).toEqual({ line: 1, column: 0 });
+      expect(result.location.end).toEqual({ line: 1, column: 3 });
+    });
+
+    it("gets the expression when not the first token on the line", () => {
```
AnshulMalik: It's very similar to the above case `gets the expression when first token on the line`?

src/utils/editor/tests/get-expression.spec.js
```diff
@@ -0,0 +1,183 @@
...
+      expect(result.location.start).toEqual({ line: 1, column: 4 });
+      expect(result.location.end).toEqual({ line: 1, column: 7 });
+    });
+
+    it("includes previous tokens in the expression", () => {
+      const codemirrorMock = {
+        lineCount: () => 2,
+        getTokenAt: location => {
+          if (location.ch >= 5 && location.ch <= 7) {
+            return { start: 4, end: 7, type: "property" };
```
AnshulMalik: I believe in here we are testing `getTokenAt` and not `getExpressionFromCoords`

src/utils/editor/tests/get-expression.spec.js
```diff
@@ -0,0 +1,183 @@
...
+      });
+      expect(result.expression).toEqual("foo.bar");
+      expect(result.location.start).toEqual({ line: 1, column: 0 });
+      expect(result.location.end).toEqual({ line: 1, column: 7 });
+    });
+
+    it("includes multiple previous tokens in the expression", () => {
+      const codemirrorMock = {
+        lineCount: () => 2,
+        getTokenAt: location => {
```
AnshulMalik: same here (testing `getTokenAt`)

src/utils/editor/tests/get-expression.spec.js
```diff
@@ -0,0 +1,183 @@
...
+      });
+      expect(result.expression).toEqual("foo.bar.baz");
+      expect(result.location.start).toEqual({ line: 1, column: 0 });
+      expect(result.location.end).toEqual({ line: 1, column: 11 });
+    });
+
+    it("does not include tokens not part of the expression", () => {
+      const codemirrorMock = {
+        lineCount: () => 2,
+        getTokenAt: location => {
```
AnshulMalik: same here (testing `getTokenAt`)

src/utils/editor/tests/get-expression.spec.js
kenjiO: For this I couldn't figure out a nice way to mock the codemirror `getTokenAt` method.  When there are more than one token in an expression this function gets called multiple times and returns different results.  Do you have any suggestions on how I could better mock the `getTokenAt` method?  I had also tried creating a code mirror instance instead of mocking it but I couldn't get it to work in the test environment.

kenjiO:
I figured out how to use a CodeMirror instance instead of a mock for testing.  In bdbb5ec I used it for the last 3 tests.  Do you prefer this way?  If so, should I do it this way for all the tests?

src/utils/editor/tests/get-expression.spec.js
AnshulMalik: Do you think `does not include tokens not part of the expression` covers this one too?

AnshulMalik:
I love the codemirror method :) Great work!

src/utils/editor/tests/get-expression.spec.js
```diff
@@ -0,0 +1,175 @@
...
+    if (document.body.createTextRange) {
+      isCreateTextRangeDefined = true;
+    } else {
+      isCreateTextRangeDefined = false;
+      // CodeMirror needs createTextRange
+      // https://discuss.codemirror.net/t/working-in-jsdom-or-node-js-natively/138/5
+      document.body.createTextRange = () => ({
+        getBoundingClientRect: () => {},
+        getClientRects: () => ({})
+      });
```
jasonLaster: does jest have a `before` block that is only run once? I'd like to avoid `isCreateTextRangeDefined`

wldcordeiro: No, I believe if you only need it once you would write it in the top of the `test` body. Or if it's needed in all the tests but only needs to be created once you can do it at the top of the `describe` or module if I'm recalling right.

src/utils/editor/tests/get-expression.spec.js
```diff
@@ -0,0 +1,175 @@
...
+  let isCreateTextRangeDefined;
+
+  beforeAll(() => {
+    if (document.body.createTextRange) {
+      isCreateTextRangeDefined = true;
+    } else {
+      isCreateTextRangeDefined = false;
+      // CodeMirror needs createTextRange
+      // https://discuss.codemirror.net/t/working-in-jsdom-or-node-js-natively/138/5
+      document.body.createTextRange = () => ({
```
wldcordeiro: `jest.fn()` over empty arrow functions.

wldcordeiro:
Awesome job, love seeing more tests come in. Feedback from a few peeps but you're on the right track here. 😄

src/utils/editor/tests/get-expression.spec.js
kenjiO: Yes, that one also covers this case so I will go ahead to take this test out.

src/utils/editor/tests/get-expression.spec.js
kenjiO: I think that `document.body` is a global variable or something like that so if I added `createTextRange` to it I wanted to delete it at the end of these tests so it wouldn't affect other tests.  Is there another way to keep track of whether or not I should delete `createTextRangeDefined` at the end of the tests?

jasonLaster:
Thanks @kenjiO. These tests are awesome

Issue #5708, opened 2018-03-17
[Flow] type reselect


jasonLaster:
Many of our selectors use [reselect](https://github.com/reactjs/reselect) so that they're memoized. This is great, but the downside is that none of these selectors are typed!

We should fix that by adding the reselect flow-typed definitions. This [branch][b] starts the effort and gets pretty close. 

---

By the way, when i did this Flow found some amazing errors in the code that were great to squash, like our sources reducer still had a field `sourcesText` which we stopped using months ago!

Here are some errors that I did not have time to fix:

<img src='https://shipusercontent.com/c9e3c2b89ce2633c67a6a326355cbf20/Screen%20Shot%202018-03-17%20at%202.34.50%20PM.png' title='Screen Shot 2018-03-17 at 2.34.50 PM.png' width=1138>
<img src='https://shipusercontent.com/dff421981913a3332385b97d504a1ddd/Screen%20Shot%202018-03-17%20at%202.34.43%20PM.png' title='Screen Shot 2018-03-17 at 2.34.43 PM.png' width=1131>

This fix will definitely be a crash course in Flow, but I'm optimistic that it will help us a lot and will be a great way to learn our codebase!

[b]: https://github.com/devtools-html/debugger.html/compare/master...jasonLaster:reselect-types?expand=1

jasonLaster:
@nyrosmith I unassigned you since it's been two weeks. Feel free to re-claim if you have time.

jasonLaster:
Just updated this [branch](https://github.com/devtools-html/debugger.html/compare/master...jasonLaster:reselect-types?expand=1). I love how the first thing that flow told me here was that this function should be killed, it's overkill to have a type checker tell you a function doesn't make sense anymore, but why not!

```diff
-export const getSelectedSourceText = createSelector(
-  getSelectedSource,
-  getSourcesState,
-  (selectedSource, sources) => {
-    const id = selectedSource.id;
-    return id ? sources.sourcesText.get(id) : null;
-  }
-);
-
```

jasonLaster:
done

Issue #5122, opened 2018-01-16
[Flow] Run flow-typed to get coverage of some third party libraries


gregtatum:
There is no coverage for most of the types, as they get coerced to `any` types when they run through  third party libraries. This is particularly bad for reselect, which powers all of the selectors. 

Flow-typed has definition files for the following libraries:

  • enzyme_v3.x.x.js
  • reselect_v3.x.x.js
  • prop-types_v15.x.x.js
  • react-redux_v5.x.x.js
  • lodash_v4.x.x.js

The tool is available here:
https://github.com/flowtype/flow-typed

Also be warned that it generates lots of "stubs" for libraries that aren't supported, they can be a bit of noise if committed.

gregtatum:
The tool is available here:
https://github.com/flowtype/flow-typed

Also be warned that it generates lots of "stubs" for libraries that aren't supported, they can be a bit of noise if committed.

wldcordeiro:
I think it's best to manually add each dependencies types rather than add all + generate stubs.

jasonLaster:
@wldcordeiro why?

jasonLaster:
@gregtatum does perf.html flow type react-redux?

gregtatum:
I built some manual definitions since the ones from `flow-typed` weren't accurate enough and were not causing errors in the props. They are pretty verbose and explicit in order to be correct. I know that people have been working on the official types, but I haven't seen what they've done. I'd be happy to have a bigger discussion around this if you're interested. We still have a couple of small correctness issues with our definition (involving ThunkActions) where we need to fix things.

Some example typing: https://github.com/devtools-html/perf.html/blob/529cb9a1286d7f3b38bf740855dcb5c09a70159e/src/components/marker-table/Settings.js#L63-L70

jasonLaster:
we're now typing all of the 3rd party libs we use

gregtatum:
😮 whoa, nice job!

PR #6498, opened 2018-06-12
[flow] add react-redux

jasonLaster:
Fixes Issue: #5708

### Summary of Changes

This makes progress on #5122 so that we can have well typed 3rd party libraries and begin using flow to its full potential.


jasonLaster:
`connect(mapStateToProps, () => actions)(com)` did not work in development.

PR #6499, opened 2018-06-12
[flow] add types for reselect

jasonLaster:
Fixes Issue: #5122 

### Summary of Changes

This adds flow type definitions for reselect, which is important for validating PropTypes and improving our redux action and selector coverage

### DIFF
--- a/src/actions/ast.js
+++ b/src/actions/ast.js
@@ -61,13 +61,11 @@ export function setSymbols(sourceId: SourceId) {
       return;
     }
 
-    await dispatch(
-      ({
-        type: "SET_SYMBOLS",
-        sourceId,
-        [PROMISE]: getSymbols(sourceId)
-      }: Action)
-    );
+    await dispatch({
+      type: "SET_SYMBOLS",
+      sourceId,
+      [PROMISE]: getSymbols(sourceId)
+    });
 
     if (isPaused(getState())) {
       await dispatch(fetchExtra());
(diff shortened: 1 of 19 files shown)


src/actions/ast.js
```diff
@@ -61,13 +61,11 @@ export function setSymbols(sourceId: SourceId) {
...
-        type: "SET_SYMBOLS",
-        sourceId,
-        [PROMISE]: getSymbols(sourceId)
-      }: Action)
-    );
+    await dispatch({
+      type: "SET_SYMBOLS",
+      sourceId,
+      [PROMISE]: getSymbols(sourceId)
+    });
```
jasonLaster: I temporarily stopped checking Promise actions, I'd like to find a solution before this lands.

src/reducers/pause.js
```diff
@@ -299,7 +300,7 @@ function getPauseLocation(state, action) {
 // top-level app state, so we'd have to "wrap" them to automatically
 // pick off the piece of state we're interested in. It's impossible
 // (right now) to type those wrapped functions.
-type OuterState = { pause: PauseState };
+type OuterState = State;
```
jasonLaster: So... this could be really cool, let me explain:

Our selectors in the reducers use `OuterState` as a hack to pretend they receive the full AppState type. Now it looks like we can just import that type from `reducers/types`

src/actions/ast.js
wldcordeiro: 👍

Issue #5687, opened 2018-03-15
[Flow] upgrade to 68


jasonLaster:
68 just came out and looks pretty great

https://circleci.com/gh/devtools-html/debugger.html/11247?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

We would get stricter action types and native types, both of which are helpful

darkwing:
@nyrosmith has an amazing start on this here:

https://github.com/devtools-html/debugger.html/pull/5893

Working off of that gets us most of the way there!

PR #6500, opened 2018-06-12
[Flow] Upgrade to 70

jasonLaster:
Fixes Issue: #5687

### Summary of Changes

* starting with 68 here #6501

### DIFF
--- a/src/components/QuickOpenModal.js
+++ b/src/components/QuickOpenModal.js
@@ -323,7 +323,15 @@ export class QuickOpenModal extends Component<Props, State> {
   isSourceSearch = () => this.isSourcesQuery() || this.isGotoSourceQuery();
 
   /* eslint-disable react/no-danger */
-  renderHighlight = (candidateString: string, query: string, name: string) => {
+  renderHighlight(
+    candidateString: string | React$Element<"div">,
+    query: string,
+    name: string
+  ): string | React$Element<"div"> {
+    if (typeof candidateString != "string") {
+      return "";
+    }
+
     const options = {
       wrap: {
         tagOpen: '<mark class="highlight">',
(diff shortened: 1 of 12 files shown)


src/components/QuickOpenModal.js
```diff
@@ -323,7 +323,15 @@ export class QuickOpenModal extends Component<Props, State> {
...
   /* eslint-disable react/no-danger */
-  renderHighlight = (candidateString: string, query: string, name: string) => {
+  renderHighlight(
+    candidateString: string | React$Element<"div">,
+    query: string,
+    name: string
+  ): string | React$Element<"div"> {
+    if (typeof candidateString != "string") {
+      return "";
+    }
```
jasonLaster: maybe we should only call this if the candidate is a string

src/components/QuickOpenModal.js
```diff
@@ -332,21 +340,26 @@ export class QuickOpenModal extends Component<Props, State> {
...
     return <div dangerouslySetInnerHTML={{ __html: html }} />;
-  };
+  }
+
+  updateResult(result: QuickOpenResult, newQuery: string): QuickOpenResult {
+    return {
+      ...result,
+      title: this.renderHighlight(result.title, basename(newQuery), "title")
+    };
+  }
```
jasonLaster: i moved this to its own function to simplify the typing but i think we can move it back...

src/components/SecondaryPanes/Frames/Frame.js
```diff
@@ -87,7 +87,7 @@ export default class FrameComponent extends Component<FrameComponentProps> {
     frame: Frame,
     selectedFrame: Frame
   ) {
-    if (e.nativeEvent.which == 3) {
+    if (e.which == 3) {
```
jasonLaster: i couldnt find nativeEvent in the types... lets see if this still works

src/utils/quick-open.js
```diff
@@ -70,7 +71,7 @@ export function formatSourcesForList(source: RelativeSource, tabs: TabList) {
 export type QuickOpenResult = {|
   id: string,
   value: string,
-  title: string,
+  title: string | React$Element<"div">,
```
jasonLaster: when we use fuzzaldrin to fuzzy highlight we return a React Element

src/workers/parser/getSymbols.js
```diff
@@ -145,7 +148,7 @@ function extractSymbol(path: SimplePath, symbols) {
   if (t.isClassDeclaration(path)) {
     symbols.classes.push({
       name: path.node.id.name,
-      parent: path.node.superClass,
+      parent: { name: path.node.superClass },
```
jasonLaster: we had a bug in inComponent that needed this

src/workers/parser/getSymbols.js
```diff
@@ -291,7 +294,8 @@ function extractSymbols(sourceId): SymbolDeclarations {
     imports: [],
     literals: [],
     hasJsx: false,
-    hasTypes: false
+    hasTypes: false,
+    loading: false
```
jasonLaster: we can no longer check if a property exists w/ flow... with that said, we might want to try another option here

wldcordeiro: 👍 interesting note, just saw that in 0.68 release notes.

src/utils/quick-open.js
wldcordeiro: 👍

src/workers/parser/getSymbols.js
wldcordeiro: Hmm, this seems to break preview however as the snapshots are not generating properly.

wldcordeiro: I'm going to revert this change, if there's a bug we should open an issue for it. 😄 

screen shot 2018-06-18 at 9 19 45 am

PR #6501, opened 2018-06-12
[Flow] Upgrade to 68

jasonLaster:
Fixes Issue: #5687

### DIFF
--- a/src/reducers/sources.js
+++ b/src/reducers/sources.js
@@ -126,12 +126,12 @@ function update(
 
     case "ADD_TAB":
       return state.merge({
-        tabs: updateTabList({ sources: state }, action.url)
+        tabs: updateTabList(state.tabs, action.url)
       });
 
     case "MOVE_TAB":
       return state.merge({
-        tabs: updateTabList({ sources: state }, action.url, action.tabIndex)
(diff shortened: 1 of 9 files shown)


src/reducers/sources.js
```diff
@@ -126,12 +126,12 @@ function update(
...
     case "ADD_TAB":
       return state.merge({
-        tabs: updateTabList({ sources: state }, action.url)
+        tabs: updateTabList(state.sources.tabs, action.url)
       });
 
     case "MOVE_TAB":
       return state.merge({
-        tabs: updateTabList({ sources: state }, action.url, action.tabIndex)
+        tabs: updateTabList(state.sources.tabs, action.url, action.tabIndex)
```
darkwing: Would it be easier if we did `state.sources.tabs || []`?

jasonLaster: hmm, what's the benefit of that?

darkwing: I (thought I) saw a circle failure due to `updateTabList` returning undefined

jasonLaster: got it. We might be able to add a sensible default in initializer

wldcordeiro:
👀 looking at this and then the 70 PR now.

PR #6502, opened 2018-06-12
[flow] upgrade to Immutable 4

jasonLaster:
### Summary of Changes

* Upgrades to immutable 4, which has significant improvements for performance and type-inference.
* We can now start using the flow inference and fix some of the inconsistencies we have had...

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -2,7 +2,6 @@
 .*node_modules/documentation.*
 .*node_modules/devtools-reps/src/object-inspector/index.js
 .*node_modules/devtools-mc-assets/.*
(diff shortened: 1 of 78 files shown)


wldcordeiro:
Oof, loads of Flow errors here (seeing 66), but this is a great start! 😄 I was looking into their road to 4.0, there are a few last outstanding issues, some of which are type definition related so we should review to see if they're affecting us to make sure we don't chase anything that's already being addressed.

jasonLaster:
yes! that'd be really helpful

jasonLaster:
I think I'd like to try going in the other direction and statically enforcing immutability.

Issue #6495, opened 2018-06-11
Search next result and previous result arrow buttons are in opposite order of typical user experience


joshunger:
The search next result and previous result arrow buttons are displayed in opposite order of typical user experience.  EXPECTING: the previous result arrow button followed by the next result arrow button.

debugger.html
<img width="1103" alt="image" src="https://user-images.githubusercontent.com/2301847/41244543-991f26f8-6d62-11e8-940e-9836b393e457.png">

FireFox Nightly
![image](https://user-images.githubusercontent.com/2301847/41244396-290d80da-6d62-11e8-93f7-76621d47f0e4.png)

Chrome
![image](https://user-images.githubusercontent.com/2301847/41244410-2fdbe00a-6d62-11e8-852f-503a31c04376.png)

PR #6504, opened 2018-06-13
fix for issue #6495 (https://github.com/devtools-html/debugger.html/i…

hardgroi:
…ssues/6495) where it was suggested to match the order of the input search arrows to 'previous result' then 'next result' which is more common.

Fixes Issue: #6495

### Summary of Changes

* Updated the order of the search input arrows to follow the order "previous result" then "next result" as suggested by the issue.
* Updated the snapshots to match the swapped arrows

### Test Plan

Updated the existing snapshots

### Screenshots/Videos
screen shot 2018-06-13 at 12 53 40 am

### DIFF
--- a/src/components/shared/SearchInput.js
+++ b/src/components/shared/SearchInput.js
@@ -109,16 +109,16 @@ class SearchInput extends Component<Props, State> {
 
     return [
       arrowBtn(
-        handleNext,
-        "arrow-down",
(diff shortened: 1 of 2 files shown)


darkwing:
Awesome work @hardgroi !  Feel free to grab another issue when you get the time.

PR #6506, opened 2018-06-13
[Scopes] fix class declarations

jasonLaster:
#6400

### Summary of Changes

Classes have two bindings, one in the inner class, one in the outer scope. This caused MappedOriginalRanges to accidentally assume the class binding range was "multiple".

```js
export default function root() {
  let one = 1;

  class Thing {
    one() {
      // fails because `Thing` is not referenced locally
      console.log("pause here");
    }
  }

  class Another {
    method() {
      // works because `Another` is referenced locally
      console.log("pause here", Another, Thing, root);
    }
  }

  new Another().method();
  new Thing().one();
}
```

### Test plan

add a mochitest


jasonLaster:
I'll go in and update the test references in a bit, but here is a small test case.

jasonLaster:
this should be ready for review

src/utils/pause/mapScopes/index.js
```diff
@@ -279,7 +279,7 @@ function generateClientScope(
 
       return {
         // Flow doesn't like casting 'parent'.
```
loganfsmyth: Looks like this comment could be removed then.

src/test/mochitest/helpers.js
```diff
@@ -814,6 +814,8 @@ async function expandAllScopes(dbg) {
 async function assertScopes(dbg, items) {
   await expandAllScopes(dbg);
 
+  // dump(JSON.stringify(items.entries()));
```
loganfsmyth: Leftovers

jasonLaster: nice catch

Issue #6474, opened 2018-06-05
Preview leaks out of Editor Window


darkwing:
I spotted this small issue on Windows:

![img_1320](https://user-images.githubusercontent.com/46655/40993949-81fd4d34-68c0-11e8-84ba-893344716abf.jpg)

The `Element`'s preview leaked into the tab bar; we should keep these tooltips in the editor area.

ATCraiger:
Hey @darkwing, I could work on this. Do you have any suggestions on how I start on this? I've worked a bit on the javascript side of the debugger, but this seems like a different type of issue.

/claim

darkwing:
Awesome @ATCraiger!  I think a good place to start would be looking at the following file:

src/components/shared/Popover.js

Within that file we do a lot of calculations for where to place a preview popup; work was recently done on this file to get previews to display on the right:

https://github.com/devtools-html/debugger.html/pull/6196

darkwing:
I was able to reproduce on both Mac and Windows by:

1.  Go to https://davidwalsh.name
2.  Open "davidwalsh.name/wp-content/themes/punky/js/main.js
3.  Pretty-print the resource.
4.  Add a breakpoint to line 46
5.  Refresh the page, let the debugger pause on the breakpoint, and scroll the editor down so that the breakpoint line is at the top of the editor pane.  
6.  Hover over "Element"

The preview/tooltip will appear in the tab bar.

AarontCraig:
Ok, I was able to replicate this as well through your steps. I also found that all of the tool tips do this.
I'll get working on the JavaScript and report back soon.

AarontCraig:
My term is coming to an end, so I'll open this up for anyone else to work on.
/unclaim

PR #6509, opened 2018-06-13
display tooltip below target when not enough space above

kenjiO:
Fixes Issue: #6474 

### Summary of Changes

* Added code to the `getTooltipCoords` function in `Popover.js` to set the top position below the target when there is not enough room between the editor.top and target.top.
* Added 2 new test cases and updated a snapshot.

### Test Plan

**Tell us a little a bit about how you tested your patch.**
I ran the debugger and verified that tooltips for targets at the top line of the editor now show below the target and no longer extend past the editor.  I also added two unit tests but will need help getting them to work if you want to keep them.  The problem is that in the test environment [this line](https://github.com/devtools-html/debugger.html/blob/7eb88759712140709d56f6df2020396701c6bddd/src/components/shared/Popover.js#L139) reports the height of the popover as `0` which messes up the calculations. 

Before:
before

After:
after

### DIFF
--- a/src/components/shared/Popover.js
+++ b/src/components/shared/Popover.js
@@ -192,12 +192,16 @@ class Popover extends Component<Props, State> {
       const editorRect = editor.getBoundingClientRect();
       const targetRect = this.props.targetPosition;
       const left = this.calculateLeft(targetRect, editorRect, tooltipRect);
-      const top = targetRect.top - tooltipRect.height;
+      const enoughRoomForTooltipAbove =
+        targetRect.top - editorRect.top > tooltipRect.height;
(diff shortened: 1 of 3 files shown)


darkwing:
Great work @kenjiO !  I tested this out and it works great; just fix up the unit test and we're good to go!  Thank you!

kenjiO:
I figured out how to get the unit tests to pass.  I don't know why the Travis build failed though.

darkwing:
@kenjiO Travis has been unreliable in as far as cloning Firefox -- it's not your fault.  I'll rebuild :)

PR #6510, opened 2018-06-14
remove unused code in devtools-components/tree.js

AnshulMalik:
This code was not being used so removed it.


jasonLaster:
cc @nchevobbe

PR #6514, opened 2018-06-17
 Fix Travis intermittent failures

jasonLaster:
### Summary of Changes

We have been seeing intermittent travis failures for the past month or so. David noticed early on that it looks like it is an out of memory issue. 

This PR tries to fix that by doubling our RAM from 4gb to 7.5gb (i think). [link](https://docs.travis-ci.com/user/reference/overview/).

I think I have some evidence to show that we have an OOM issue:

A recent build had this output for the `hg clone`. The `killed!` output is weird, but I think the important thing is the `137` exit code which looks like a standard linux and docker code for memory issues.

screen shot 2018-06-17 at 2 18 44 am

Another indicator that the clone failed is when the update includes the message `untracked file differs` and has this error message `abort: untracked files in working directory differ from files in requested revision`

screen shot 2018-06-17 at 2 23 26 am

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,5 @@
+sudo: required
+dist: precise
 language: python
   - "3.6"
 
--- a/bin/ci/clone-gecko.sh
+++ b/bin/ci/clone-gecko.sh
@@ -1,9 +1,12 @@
 #!/bin/bash
-
+python --version
 hg --version
+df -h 
+
 rm -rf firefox/
 hg clone https://hg.mozilla.org/mozilla-unified/ firefox
 
 cd firefox
+hg update --clean
 hg co $MC_COMMIT
 cd ..


jasonLaster:
i kicked off another 10ish runs here
https://travis-ci.org/devtools-html/debugger.html/branches

i'll clean up this branch in the mean time and if the others are green we can land

jasonLaster:
All 10 runs successfully cloned and built. One run had a test failure, but that's different. 

🍷

darkwing:
Nice.  I was going to go a different route and, instead of cloning MC, would simply `curl` the code down at the designated commit ID.  Could still be worth a try.  Much less metadata and work, in theory.

jasonLaster:
oh yeah! I've explored this route as well in the past, the challenge is artifact builds need HG which is sad!

PR #6520, opened 2018-06-19
[Breakpoints] Enable skip all pausing button in dev

### LINKED ISSUE #6512: Enable "Disable All Pausing" Button
Now that @jasonLaster 's server-side patch has been merged, we should enable the feature-flag for this button.  It will require an API change because the param name is now "skipPausing".

darkwing:
Fixes Issue: #6512

## Testing
- `yarn copy`
- `cd firefox && ./mach build faster`
- `./mach run`
- Go into `about:config` and turn on `debugger.features.skip-pausing`
- Restart Firefox (`./mach run`)
- Toggle on and off the button and ensure breakpoints are either hit or not.

### DIFF
--- a/src/test/mochitest/browser_dbg-breakpoint-skipping.js
+++ b/src/test/mochitest/browser_dbg-breakpoint-skipping.js
@@ -1,0 +1,29 @@
+/* Any copyright is dedicated to the Public Domain.
+ * http://creativecommons.org/publicdomain/zero/1.0/ */
+
+add_task(async function() {
+  await pushPref("devtools.features.skip-pausing", true);
+
+  let dbg = await initDebugger("doc-scripts.html");
+
+  await selectSource(dbg, "simple3");
+  await addBreakpoint(dbg, "simple3", 2);
+  await addBreakpoint(dbg, "simple3", 3);
+
+  let syncedBps = waitForDispatch(dbg, "SYNC_BREAKPOINT", 2);
+  await reload(dbg, "simple3");
+  await waitForSelectedSource(dbg, "simple3");
+  await syncedBps;
+
+  assertNotPaused(dbg);
+
+  // Turn "skip pausing" on
+  clickElementWithSelector(dbg, ".command-bar-skip-pausing");
+
+  syncedBps = waitForDispatch(dbg, "SYNC_BREAKPOINT", 2);
+  await reload(dbg, "simple3");
+  await waitForSelectedSource(dbg, "simple3");
+  await syncedBps;
+
+  assertPaused(dbg);
+});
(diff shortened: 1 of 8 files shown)


darkwing:
👍

src/test/mochitest/browser_dbg-breakpoints.js
```diff
@@ -73,3 +73,37 @@ add_task(async function() {
   bp2 = findBreakpoint(dbg, "simple2", 5);
   is(bp2.disabled, false, "second breakpoint is enabled");
 });
+
+add_task(async function() {
+  Services.prefs.setBoolPref("devtools.features.skip-pausing", true);
```
jasonLaster: i believe we want to use pushPref so it resets at the end

https://github.com/jasonlaster/debugger.html/blob/1b737fc00f7356e9ea5ceda3a5938f9924d4d39f/src/test/mochitest/browser_dbg-async-stepping.js#L6-L7

src/test/mochitest/browser_dbg-breakpoints.js
```diff
@@ -73,3 +73,37 @@ add_task(async function() {
...
+  await addBreakpoint(dbg, "simple2", 3);
+  await addBreakpoint(dbg, "simple2", 5);
+
+  let syncedBps = waitForDispatch(dbg, "SYNC_BREAKPOINT", 2);
+  await reload(dbg, "simple2");
+  await waitForSelectedSource(dbg, "simple2");
+  await syncedBps;
+
+  assertPaused(dbg);
+});
```
jasonLaster: we try to avoid multiple `add_task` calls in the same test file because they run at the same time in different tabs. It might be nicer to have a new mochitest file and  consolidate the two tasks.

darkwing: Thanks for the tip!

src/test/mochitest/browser_dbg-breakpoints.js
darkwing: Ohhh, nice!

src/test/mochitest/browser_dbg-breakpoint-skipping.js
```diff
@@ -0,0 +1,29 @@
+/* Any copyright is dedicated to the Public Domain.
+ * http://creativecommons.org/publicdomain/zero/1.0/ */
+
+add_task(async function() {
+  await pushPref("devtools.features.skip-pausing", true);
```
darkwing: @jasonLaster This isn't enabling the button within tests; is there a further update required?

jasonLaster: did you add the flag to `assets/panel/prefs.js`?

jasonLaster:
do we need to flip this too? 
https://github.com/devtools-html/debugger.html/blob/master/assets/panel/prefs.js#L66

darkwing:
I did that but for some reason the test is still wonk; what I'm trying to do is very straight-forward so I'm leaving this for the night and will come back with a fresh mind tomorrow

jasonLaster:
i made a bunch of tweaks, but the main one is to assert that `simple()` returns 3, which would only happen if we don't pause.

Issue #6519, opened 2018-06-18
[Breakpoints] Right clicking on a breakpoint does not open context menu


kenjiO:
STR
Set a breakpoint. In the Secondary Panes breakpoints list, right click on the breakpoint.  Then instead of opening the context menu an error comes up in the console: `TypeError: breakpoints is undefined`

![](http://g.recordit.co/lwHV2A3spj.gif)

I was able to get the error to stop by adding `breakpoints: getBreakpoints(state)` to mapStateToProps in [Breakpoint.js](https://github.com/devtools-html/debugger.html/blob/cfa74f7b77b29496e1e0beacd970a6c7778f0768/src/components/SecondaryPanes/Breakpoints/Breakpoint.js#L175-L178).  Would you like me to do a PR for this?

jasonLaster:
Thanks @kenjiO. That'd be great

jbhoosreddy:
fixed in #6521

PR #6521, opened 2018-06-20
[Breakpoints] fix missing redux prop

kenjiO:
Fixes Issue: #6519 

### Summary of Changes

* Add a breakpoints prop to Breakpoint compnent that gets data from redux selector

### Test Plan

Ran the debugger as a web app and set some breakpoints.  Then I right clicked on a breakpoint and verified the context menu displayed and the error did not come up in the console anymore.  I tried the features of the context menu and they all seemed to work properly.  

**Note:** after thinking about this I realized the data from the selector I chose might not be what the component is expecting.  Someone more familiar with the design of the component might want to verify this is the correct data for the breakpoints prop.

### DIFF
--- a/src/components/SecondaryPanes/Breakpoints/Breakpoint.js
+++ b/src/components/SecondaryPanes/Breakpoints/Breakpoint.js
@@ -20,6 +20,8 @@ import { getLocationWithoutColumn } from "../../../utils/breakpoint";
 import { features } from "../../../utils/prefs";
 import { getEditor } from "../../../utils/editor";
 
+import type { BreakpointsMap } from "../../../reducers/types";
+
 import type {
   Frame,
   Source,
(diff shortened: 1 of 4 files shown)


jasonLaster:
This is great. We should also try and add a mochitest or jest test before this lands to make sure we don't regress.

kenjiO:
I came up with 3 different ways to write a test for this.  Can you let me know if you want to use any of them?  I also added the breakpoints property to the flow Props but I don't know flow well so let me know if that needs to be changed.  Thanks

src/components/SecondaryPanes/Breakpoints/tests/Breakpoint.spec.js
```diff
@@ -50,6 +55,46 @@ describe("Breakpoint", () => {
...
     expect(component).toMatchSnapshot();
   });
+
+  it("gets breakpoints prop from Connect version #1", () => {
+    const store = createStore();
+    const component = shallow(<Breakpoint />, { context: { store } });
+    expect(component.prop("breakpoints")).toBe(
+      store.getState().breakpoints.breakpoints
+    );
+  });
```
jasonLaster: I think i like #1 the most.

It would be nice to change the test a bit so that instead of asserting that we have the same breakpoints, we assert that the render was successful and that the user can infact click an option like `disable`.

This might be better as a mochitest, at which point, i'd be happy to create one which is similar to `breakpoints-cond` which also uses a context menu i believe

jasonLaster:
CC @wldcordeiro what do you think?

jasonLaster:
I switched the test over to a small mochitest and updated the flow type

Issue #6394, opened 2018-05-23
[Tabs] Title Not Matching Tree Title


darkwing:
I found this while testing a debugger release on YouTube:

<img width="868" alt="tree2" src="https://user-images.githubusercontent.com/46655/40434331-c071594a-5e73-11e8-9cb2-c590089827b6.png">

<img width="1056" alt="treevstabname" src="https://user-images.githubusercontent.com/46655/40434336-c2704e90-5e73-11e8-8764-1e9f5721c549.png">

jasonLaster:
nice, we should totally fix that

darkwing:
OK, so I found the issue here.  The URL of `video_masthead` is actually:

https://www.youtube.com/video_masthead?video_id=AOnFe2aaX14&textLine1=Get+20%25+Off+Sitewide+-+Today+only&textLine2=Shop+Now+and+Save+20%25+on+EVERYTHING&autocrop=0&autoplay_start_time=0&autoplay_duration=15000&channel_banner=1&video_wall=0&subscribe_button=1&subscriber_count=1&site_cta=1&cta_label=Shop+now&layout=material&line_item_id=4691152491&creative_id=138235268831

The tree is removing the querystring.  The tabs and breakpoint pane are not:

<img width="1277" alt="sourcenames" src="https://user-images.githubusercontent.com/46655/41060080-146e106c-6994-11e8-9626-411bf1744837.png">

@digitarald What shall we do here?  I like the cleanliness of `video_masthead` but are we concerned that multiple usages of `script.js?my-module` and `script.js?my-other-module` will cause problems?

digitarald:
How would the sources tree handle `serve.php?file=a.js` and `serve.php?file=b.js` atm? This might be already affecting some users with dynamic server-side build steps.

Assuming filename is more important than query parameters for static assets; I would propose to try to always show the filename and show the query string in muted colors and overflow into an ellipse.

Unsure about (index); it seems that we know the real name internally and should probably just show it in Sources?

darkwing:
Here's a step in the correct direction:

```diff
diff --git a/src/components/PrimaryPanes/SourcesTree.js b/src/components/PrimaryPanes/SourcesTree.js
index a0377544d..595d70418 100644
--- a/src/components/PrimaryPanes/SourcesTree.js
+++ b/src/components/PrimaryPanes/SourcesTree.js
@@ -42,7 +42,8 @@ import {
   updateTree
 } from "../../utils/sources-tree";
[...]

PR #6525, opened 2018-06-21
[Tabs] Title Not Matching Tree Title

jasonLaster:
Fixes Issue: #6394

### Summary of Changes

* re-uses getURL in getFilename which is the main other call site
* memoizes getURL so that we only parse a url once, this is one of the top slow downs for the tree
* removes getFileNameFromUrl because it can't be memoized

### Test Plan

* unit tests

Go to youtube.com, look at video_masthead, embed (index)

* hover on tabs,  breakpoint file groups, source tree items

### DIFF
--- a/src/components/PrimaryPanes/SourcesTree.js
+++ b/src/components/PrimaryPanes/SourcesTree.js
@@ -115,13 +115,12 @@ class SourcesTree extends Component<Props, State> {
     }
 
     if (nextProps.shownSource && nextProps.shownSource != shownSource) {
-      const listItems = getDirectories(nextProps.shownSource, sourceTree);
-
-      if (listItems && listItems[0]) {
-        this.selectItem(listItems[0]);
-      }
-
-      return this.setState({ listItems });
+      // const listItems = getDirectories(nextProps.shownSource, sourceTree);
+      //
+      // if (listItems && listItems[0]) {
+      //   this.selectItem(listItems[0]);
+      // }
+      // return this.setState({ listItems });
     }
 
     if (
@@ -129,7 +128,7 @@ class SourcesTree extends Component<Props, State> {
       nextProps.selectedSource != selectedSource
     ) {
       const highlightItems = getDirectories(
-        getRawSourceURL(nextProps.selectedSource.url),
+        nextProps.selectedSource,
         sourceTree
       );
 
--- a/src/components/PrimaryPanes/SourcesTree.js
+++ b/src/components/PrimaryPanes/SourcesTree.js
@@ -124,7 +124,9 @@ class SourcesTree extends Component<Props, State> {
           sources[matchingSources[0]],
           sourceTree
         );
-        this.selectItem(listItems[0]);
+        if (listItems && listItems.length) {
+          this.selectItem(listItems[0]);
+        }
         return this.setState({ listItems });
       }
     }
--- a/src/components/PrimaryPanes/SourcesTree.js
+++ b/src/components/PrimaryPanes/SourcesTree.js
@@ -20,7 +20,9 @@ import {
   getSourceCount
 } from "../../selectors";
 
+// Actions
 import actions from "../../actions";
+import { getRawSourceURL } from "../../utils/source";
 
 // Components
 import SourcesTreeItem from "./SourcesTreeItem";
@@ -36,8 +38,6 @@ import {
   nodeHasChildren,
   updateTree
 } from "../../utils/sources-tree";
-
-import { getRawSourceURL } from "../../utils/source";
 
 import type {
   TreeNode,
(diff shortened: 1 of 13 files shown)


jasonLaster:
This is a pretty rough version, but i wanted to get it up before i headed out. The nice thing here, which we didn't discuss, is that getURL is memoized, which will really help our source tree perf. it might even help @yurydelendik show a couple more thousand sources. cc @AnshulMalik

src/components/PrimaryPanes/SourcesTree.js
```diff
@@ -123,21 +123,20 @@ class SourcesTree extends Component<Props, State> {
...
 
     if (nextProps.shownSource && nextProps.shownSource != shownSource) {
-      const listItems = getDirectories(nextProps.shownSource, sourceTree);
-
-      if (listItems && listItems[0]) {
-        this.selectItem(listItems[0]);
-      }
-
-      return this.setState({ listItems });
+      // const listItems = getDirectories(nextProps.shownSource, sourceTree);
```
jasonLaster: will figure out how to make this a source tomorrow...

not sue why selectItem fires selectSource again, that is a bit weird...

src/utils/sources-tree/tests/addToTree.spec.js
```diff
@@ -224,7 +224,7 @@ describe("sources-tree", () => {
       expect(formatTree(tree)).toMatchSnapshot();
     });
 
-    it("uses debuggeeUrl as default", () => {
+    xit("uses debuggeeUrl as default", () => {
```
jasonLaster: i think we can remove this

src/utils/sources-tree/tests/getUrl.spec.js
```diff
@@ -0,0 +1,78 @@
...
+  it("handles normal url with http and https for filename", function() {
+    const urlObject = getURL({ url: "https://a/b.js" });
+    expect(urlObject.filename).toBe("b.js");
+
+    const urlObject2 = getURL({ url: "http://a/b.js" });
+    expect(urlObject2.filename).toBe("b.js");
+  });
+
+  it("handles url with querystring for filename", function() {
+    const urlObject = getURL({ url: "https://a/b.js?key=randomeKey" });
```
darkwing: Don't kill me but there's an extra `e` in the `key` value...

jasonLaster: haha - love it

src/components/PrimaryPanes/SourcesTree.js
```diff
@@ -123,21 +123,26 @@ class SourcesTree extends Component<Props, State> {
...
     if (nextProps.shownSource && nextProps.shownSource != shownSource) {
-      const listItems = getDirectories(nextProps.shownSource, sourceTree);
+      const matchingSources = Object.keys(sources).filter(
+        sourceId => getRawSourceURL(sources[sourceId].url) === shownSource
+      );
 
-      if (listItems && listItems[0]) {
+      if (matchingSources.length) {
+        const listItems = getDirectories(
+          sources[matchingSources[0]],
```
jasonLaster: lets see if we can use `getSource` here and we might need to return if there isn't a source...

darkwing:
@jasonLaster For the life of me I cannot identify why Flow is having a problem when this looks exactly the same as other usages.  Could you put eyes on it?

jasonLaster:
Seeing one travis error
> sources-named-eval.js

darkwing:
The test fails here because it only ever see *2* sources;

## master

Found elements: 3; example.com|​ browser/devtools/client/debugger/new/test/mochitest/examples|​ evaled.js

## on this branch

Found elements: 2; example.com|​ browser/devtools/client/debugger/new/test/mochitest/examples

Somehow "evaled.js" isn't being added to the tree.

Debugged using:

```diff
diff --git a/src/test/mochitest/browser_dbg-sources-named-eval.js b/src/test/mochitest/browser_dbg-sources-named-eval.js
index fad49e60d..8b669db1c 100644
--- a/src/test/mochitest/browser_dbg-sources-named-eval.js
+++ b/src/test/mochitest/browser_dbg-sources-named-eval.js
@@ -7,6 +7,10 @@ async function waitForSourceCount(dbg, i) {
   // We are forced to wait until the DOM nodes appear because the
   // source tree batches its rendering.
   await waitUntil(() => {
+    const foundElements = findAllElements(dbg, "sourceNodes");
+    const names = [...foundElements].map(el => el.textContent.trim()).join("|");
+    info(`Found elements: ${foundElements.length}; ${names}`);
+    info(`------------------------------`);
     return findAllElements(dbg, "sourceNodes").length === i;
   });
 }
```

src/utils/sources-tree/getURL.js
```diff
@@ -45,65 +47,75 @@ export function getURL(sourceUrl: string, debuggeeUrl: string = ""): ParsedURL {
...
       if (pathname && pathname.startsWith("/")) {
         // use file protocol for a URL like "/foo/bar.js"
-        return merge(def, {
+        return {
+          ...def,
           path: path,
-          group: "file://",
-          filename: filename
-        });
-      } else if (host === null) {
```
darkwing: Finally figured it out; removing this `else if` makes `browser_dbg-sources-named-eval.js` fail because it doesn't start with "/" and we have no other handling here, so the source essentially gets "lost".

When I make this change:

```diff
diff --git a/src/utils/sources-tree/getURL.js b/src/utils/sources-tree/getURL.js
index 8c0687dee..d3c354f18 100644
--- a/src/utils/sources-tree/getURL.js
+++ b/src/utils/sources-tree/getURL.js
@@ -90,6 +90,14 @@ function _getURL(source: Source): ParsedURL {
           group: "file://"
         };
       }
+
+      return {
+        ...def,
+        path: path,
+        filename,
+        group: "example.com"
+      };
+
```

The test passes because I've hardcoded the domain.  I think we need to use the `defaultDomain` var that was removed from this first commit.

Issue #5134, opened 2018-01-18
[Preview] 'Split console' overlays the debugger variable tooltip


jnachtigall:
*STR:*
1. Set a breakpoint and also enable `Toggle split console [esc])`
2. Hover a variable at bottom

*AR:*
The variable tooltip is cropped:
![image](https://user-images.githubusercontent.com/3307404/35091459-06ff6616-fc3d-11e7-94cd-e4cc982ce692.png)

*ER:*
Don't crop. Either have the tooltip above the `Split console`. Or: Take the opened console into consideration while calculation into which direction the tooltop should open (So in this case not to the bottom but to the top).

lukaszsobek:
Thinking about it there seem to be two situations worth considering: 
1.) the debugger is too low to hold the tooltip.
2.) the debugger is too narrow to hold the tooltip.

What would you say should happen in such situations?

jasonLaster:
@juliandescottes do you think we could use a XUL wrapper here?

jnachtigall:
> What would you say should happen in such situations?

From a user perspective I would expect the tooltip to overlay just like normal, that is, with part of it overlaying over the normal html content (and not be confined to the debugger window). Don't know if this is technically possible though, because afaik the debugger is in its own `iframe` or so. (But at least this might be possible with regards to the `Console` split area)

juliandescottes:
XUL is on the way out, we shouldn't invest in this direction. We should design tooltips to fit in the devtools frame.

Regardless of XUL, the DevTools tooltip is creating tooltip containers in the topmost devtools frame, so even without XUL we have no overlapping issues between devtools panels (eg splitconsole, devtools toolbar etc). It also takes care of calculating best position for the tooltip. 

I would be concerned with having the debugger relying on a different tooltip implementation on github vs mc. Given the current workflow I think I'd prefer the debugger to keep a custom implementation, even if it's more limited.

Calculate the available space in the debugger frame and anchor the tooltip accordingly.

ochameau:
@bgrins is there any plan to replace xul:popup/xul:panel with something else? If not a plan, has it been discussed? I find it quite frustrating to limit devtools UX because of XUL being deprecated.

tripurari001:
I would like to work on this 

My approach would be to calculate the best position to open the tool-tip ( top or bottom ) based on the position of element ( hovered one )

darkwing:
Great @tripurari001 !  Let us know if we can help along the way :)

[...]

PR #6537, opened 2018-06-22
[Preview] Don't allow popover to move outside editor boundaries

darkwing:
Fixes Issue: #5134

Took a while to identify the problem element but this should be the right strategy.

### DIFF
--- a/src/components/Editor/Preview/Popup.css
+++ b/src/components/Editor/Preview/Popup.css
@@ -5,12 +5,9 @@
 .popover .preview-popup {
   background: var(--theme-body-background);
   width: 350px;
-  min-height: 80px;
   border: 1px solid var(--theme-splitter-color);
   padding: 10px;
   height: auto;
-  min-height: inherit;
-  max-height: 200px;
   overflow: auto;
   box-shadow: 1px 2px 3px var(--popup-shadow-color);
 }
--- a/src/components/shared/Popover.js
+++ b/src/components/shared/Popover.js
@@ -4,6 +4,7 @@
 
 // @flow
 import React, { Component } from "react";
+import ReactDOM from "react-dom";
 import classNames from "classnames";
 import BracketArrow from "./BracketArrow";
 
@@ -26,8 +27,11 @@ type State = {
   left: number,
   top: number,
   targetMid: TargetMid,
-  orientation: Orientation
+  orientation: Orientation,
+  maxHeight: number
 };
+
+const DEFAULT_MAX_HEIGHT = 200;
 
 class Popover extends Component<Props, State> {
   $popover: ?HTMLDivElement;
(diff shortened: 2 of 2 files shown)


jasonLaster:
nice, mind sharing a GIF

jasonLaster:
@darkwing is this ready for review? If so, can you share a before/after GIF?

darkwing:
Not ready yet -- see the "ToDo" in the PR description

jasonLaster:
got it. yeah, this is a tricky case. I suppose one option if we can't put the popup below or above, is to always go to the right and then use the full height... I believe this is what chrome does. Worth checking out

darkwing:
This PR is 75+% of the way there.  Where this PR falls short is when the popover appears *above* the highlighted token:

- An incorrect maxHeight is calculated because the maxHeight is currently taking the {all space} - {top}
- Since the maxHeight is wrong, the arrow shifts downward and no longer points to the highlighted token when the `ObjectInspector` on the inside is expanded.

I think the best course of action here would be to recalculate the maxHeight an positioning of the popover when the `ObjectInspector` is expanded or contracted; that should fix the maxHeight problem.  An additional need is adding the orientation as a condition to the maxHeight positioning so that the popover appears in the correct spot.

darkwing:
Current status:

heightyessss

I need to change the code, however, so that the top arrow content doesn't look weird, but this is super encouraging.

darkwing:
@jasonLaster I think this is ready for evaluation.  I don't want the scope to creep out and all my testing shows this works

jasonLaster:
can we use react `style={}` while rendering?
https://reactjs.org/docs/dom-elements.html#style

darkwing:
Instead of `popupNode.style.maxHeight = `${maxHeight}px`;`?  Nope; we're styling an element inside the component, not the component wrapper itself.

jasonLaster:
I wonder what this would look like if we used a [render prop](https://reactjs.org/docs/render-props.html) strategy where Popup can get the coords from popover and use them intelligently...

Here's a quick sketch i made. I never tried to run it and i did get a bit distracted.
- coords becomes a regular field i.e. not state
- the default coords are saved once
- coords are passed up to Popup, which uses it when rendering 

Why is this strategy better?

I think it is dangerous for Popover to know too much about what it is showing. It is a bit cleaner for Popup to be responsible for doing its own styling... Anyways, this is just a hunch i'm not in the code so just guessing.

[sketch]: https://gist.github.com/jasonLaster/459f61ca03ef7b4a92933410d86229f1

darkwing:
Render props are cool but I don't see how it applies here -- `this.getChildren()` is the closest handle I get

jasonLaster:
https://gist.github.com/jasonLaster/459f61ca03ef7b4a92933410d86229f1

darkwing:
Superceded by: https://github.com/devtools-html/debugger.html/pull/6654

PR #6542, opened 2018-06-25
Sorce tree refactor

jasonLaster:
### Summary of Changes

* extracts source tree item from the source tree class
* this should make the code more readable
* it will make it easier to memoize expensive properties that we'd otherwise have to recompute every render

### Test Plan

jest/flow/mochitests


jasonLaster:
This should be ready to go. i believe reload.js is a separate bug

darkwing:
Works great!

PR #6548, opened 2018-06-26
fix(sources): Make self sourcemapped urls unique

jbhoosreddy:
Fixes Issue: #5663 

### Summary of Changes

* By making the source url unique, it fixes both SourcesTree and TabsList. Maybe in the future, we need to get a better plan, but for now it's probably fine.

## ToDo:
- [x] Allow breakpoint creation for `[sm]` files; *Note*: a breakpoint is getting added to the store but the `isVisible()` function inside `getVisibleBreakpoints()` is filtering it out.
- [x] Right-clicking the tab and doing "copy to clipboard" keeps the "[sm]" in the URL
- [x] If you open a source and refresh, the editor displays as empty (never even renders) upon refresh

## Notes:
- At present, if you add a breakpoint in a `[sm]` file, it only displays when you go to the ugly file.

### DIFF
--- a/packages/devtools-source-map/src/utils/index.js
+++ b/packages/devtools-source-map/src/utils/index.js
@@ -23,12 +23,18 @@ function isGeneratedId(id: string) {
   return !isOriginalId(id);
 }
 
-function formatUrl(generatedUrl: ?string, originalUrl: string) {
+function formatUrl(generatedUrl: ?string, originalUrl: string): string {
   return originalUrl === generatedUrl ? `${originalUrl} [sm]` : originalUrl;
 }
 
-function unformatUrl(url: ?string) {
-  return (url || "").replace(" [sm]", "");
+function unformatUrl(url: ?string): string {
+  if (!url) {
+    return "";
+  }
+  if (!url.endsWith("[sm]")) {
+    return url;
+  }
+  return url.replace(/( |%20)\[sm\]$/, "");
 }
 
 /**
--- a/packages/devtools-source-map/src/utils/index.js
+++ b/packages/devtools-source-map/src/utils/index.js
@@ -23,6 +23,14 @@ function isGeneratedId(id: string) {
   return !isOriginalId(id);
 }
 
+function formatUrl(generatedUrl: ?string, originalUrl: string) {
+  return originalUrl === generatedUrl ? `${originalUrl} [sm]` : originalUrl;
+}
+
+function unformatUrl(url: ?string) {
+  return (url || "").replace(" [sm]", "");
+}
+
 /**
  * Trims the query part or reference identifier of a URL string, if necessary.
  */
@@ -79,5 +87,7 @@ module.exports = {
   isOriginalId,
   isGeneratedId,
   getContentType,
+  formatUrl,
+  unformatUrl,
   contentMapForTesting: contentMap
 };
--- a/src/actions/sources/newSources.js
+++ b/src/actions/sources/newSources.js
@@ -80,15 +80,13 @@ function loadSourceMap(sourceId: SourceId) {
 
     if (!urls) {
       // If this source doesn't have a sourcemap, enable it for pretty printing
-      dispatch(
+      return dispatch(
         ({
           type: "UPDATE_SOURCE",
           source: { ...source, sourceMapURL: "" }
         }: Action)
       );
-      return;
     }
-
     return urls.map(url => createOriginalSource(url, source, sourceMaps));
   };
 }
--- a/src/components/PrimaryPanes/SourcesTreeItem.js
+++ b/src/components/PrimaryPanes/SourcesTreeItem.js
@@ -18,7 +18,9 @@ import { features } from "../../utils/prefs";
 import type { TreeNode } from "../../utils/sources-tree/types";
 
 import type { Source } from "../../types";
-const { unformatUrl: sourceMapUnformatUrl } = require("devtools-source-map");
+const {
+  unformatUrl: sourceMapUnformatUrl
+} = require("devtools-source-map/src/utils");
 
 type Props = {
   debuggeeUrl: string,
--- a/src/components/PrimaryPanes/SourcesTreeItem.js
+++ b/src/components/PrimaryPanes/SourcesTreeItem.js
@@ -153,7 +153,8 @@ export default class SourceTreeItem extends Component<Props, State> {
       case "moz-extension://":
         return L10N.getStr("extensionsText");
       default:
-        return name;
+        // we add " [sm]" to original sources if they have same url as generated
+        return name.replace("%20[sm]", " [sm]");
     }
   }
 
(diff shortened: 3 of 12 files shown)


src/actions/sources/newSources.js
```diff
@@ -79,15 +79,19 @@ function loadSourceMap(sourceId: SourceId) {
...
           source: { ...source, sourceMapURL: "" }
         }: Action)
       );
-      return;
     }
-
+    dispatch(
+      ({
+        type: "UPDATE_SOURCE",
+        source: { ...source, url: `${source.url}[sm]` }
```
jasonLaster: i think we only want to add [sm] *if* there is a generated source with the same URL

src/actions/sources/newSources.js
```diff
@@ -79,15 +79,19 @@ function loadSourceMap(sourceId: SourceId) {
...
           type: "UPDATE_SOURCE",
           source: { ...source, sourceMapURL: "" }
         }: Action)
       );
-      return;
     }
-
+    dispatch(
+      ({
+        type: "UPDATE_SOURCE",
```
jasonLaster: will this change the generated source or the originals?

src/actions/sources/newSources.js
```diff
@@ -79,15 +79,19 @@ function loadSourceMap(sourceId: SourceId) {
...
+        source: { ...source, url: `${source.url}[sm]` }
+      }: Action)
+    );
     return urls.map(url => createOriginalSource(url, source, sourceMaps));
```
jasonLaster: we might want to do something here where we compare source.url with the url ...

jbhoosreddy:
@jasonLaster So, when I update the sm'd url it throws because SourceMaps has not seen that url in its cache. The only url I could safely update was the non-sourcemapped one.  Maybe we can call it ":original" or something. Thoughts?

darkwing:
Wow, excellent work @jbhoosreddy !

jasonLaster:
i see, we might need to make the change to the URLs in `packages/devtools-source-maps/...`

jbhoosreddy:
@jasonLaster should be good to go.

jbhoosreddy:
I'll fix mochitests later today

jbhoosreddy:
yay! failing test was not me. ready for review!

jasonLaster:
This work looks wonderful! I'll be sure to QA as soon as possible.

jasonLaster:
The bug i ran into was adding a breakpoint here:

1. https://w6q8w3lz47.codesandbox.io/

screen shot 2018-07-03 at 12 46 36 pm

src/components/PrimaryPanes/SourcesTreeItem.js
```diff
@@ -153,7 +153,8 @@ export default class SourceTreeItem extends Component<Props, State> {
       case "moz-extension://":
         return L10N.getStr("extensionsText");
       default:
-        return name;
+        // we add " [sm]" to original sources if they have same url as generated
+        return name.replace("%20[sm]", " [sm]");
```
jasonLaster: we will likely need to abstract this later so that we can reuse it in other places

jasonLaster:
i'm seeing other issues reloading this [page ](https://w6q8w3lz47.codesandbox.io/) when it's paused. Seems like we just need more testing

jasonLaster:
hey @jbhoosreddy what's your timing here?

darkwing:
ToDo:
- [ ] Allow breakpoint creation for `[sm]` files
- [ ] Right-clicking the tab and doing "copy to clipboard" keeps the "[sm]" in the URL

packages/devtools-source-map/src/utils/index.js
```diff
@@ -23,6 +23,14 @@ function isGeneratedId(id: string) {
   return !isOriginalId(id);
 }
 
+function formatUrl(generatedUrl: ?string, originalUrl: string) {
+  return originalUrl === generatedUrl ? `${originalUrl} [sm]` : originalUrl;
+}
+
+function unformatUrl(url: ?string) {
+  return (url || "").replace(" [sm]", "");
```
darkwing: We should change this to a regex to remove only `[sm]` at the end of a file

src/utils/sources-tree/utils.js
```diff
@@ -51,7 +50,9 @@ export function isSource(item: TreeNode) {
 }
 
 export function getFileExtension(source: Source): string {
-  const parsedUrl = getURL(source).path;
+  const strippedUrl = source.url.replace(/ \[sm\]/, "");
+  const parsedUrl = parse(strippedUrl).pathname;
```
jasonLaster: maybe we should do this inside of `parse(source)`, so that every source has the `[sm]` stripped out of the fields?

jasonLaster:
These changes look great! Can't wait to test it tomorrow.

How's reloading with tabs and breakpoints?

jbhoosreddy:
@jasonLaster if I move `unformatUrl` inside `parse()`, we lose the `%20[sm]` string from the name early on.
screen shot 2018-08-05 at 2 25 05 pm
But navigating still works.

jbhoosreddy:
still fixing issues with breakpoints

jbhoosreddy:
at least breakpoints are working well. also copy to clipboard is working.

src/utils/url.js
```diff
@@ -20,14 +21,15 @@ const defaultUrl = {
 };
 
 export function parse(url: string): URL | object {
+  const strippedUrl = unformatUrl(url);
```
jbhoosreddy: probably we want to take this out of parse, if we want to preserve " [sm]" in the text.

jasonLaster: Yep. Perhaps. We could also format it ourselves in the render function. [sm] is kinda ugly :) we just need to know it's original

jbhoosreddy: or we just don't render ' [sm]' as it is right now?

jbhoosreddy:
@jasonLaster @darkwing could you guys also test this thing. I'm sure I missed something.

src/utils/url.js
jasonLaster: sure

jasonLaster:
@jbhoosreddy happy to!

darkwing:
@jasonLaster :  As a general rule, I think it would be good to rename "formatUrl" and "unformatUrl" to something much more explicit, like "addSourcemapPrefix" and "removeSourcemapPrefix".  Thoughts?  This is the second case where we're customizing URLs (`:formatted` being the other), and I want to make it apparent exactly what's going on.

jasonLaster:
makes sense @darkwing

jasonLaster:
so... @darkwing's line of thinking around _not_ changing the URL is slowly seeping into my brain... anyways, I put together this proof of concept [here](https://github.com/devtools-html/debugger.html/compare/master...jasonLaster:sm-urls?expand=1)

It can do some things:

- show the ` [sm]` suffix in the tree
- select the correct source and persist it on reload
- add breakpoints (note i redid a lot of actions/sources/newSources, but that is not too bad

It cannot persist tabs, but with the current tabs reducer work... i think we might have a shot at doing that in the future...

jbhoosreddy:
I can close out this PR in favor of your changeset @jasonLaster

jbhoosreddy:
didn't realize we switched to IDs, closing.

Issue #5072, opened 2018-01-10
[Flow] Connected components are not covered


gregtatum:
I discovered this in perf.html and wanted to pass it on so you all were aware.

Steps to reproduce: break some assumption in the `react-redux` `connect()` function.
Expected Result: Flow errors out, revealing your broken types.
Actual Result: There are no Flow errors.

For instance in the `PrimaryPanes/Outline` component, the props look like:

```js
type Props = {
  symbols: SymbolDeclarations,
  selectLocation: ({ sourceId: string, line: number }) => void,
  selectedSource: ?SourceRecord
};
```

Then in the `connect()` function, I can break the symbols type.

```diff
diff --git a/src/components/PrimaryPanes/Outline.js b/src/components/PrimaryPanes/Outline.js
index 8d1fe746..450fd660 100644
--- a/src/components/PrimaryPanes/Outline.js
+++ b/src/components/PrimaryPanes/Outline.js
@@ -120,7 +120,7 @@ export default connect(
   state => {
     const selectedSource = getSelectedSource(state);
     return {
-      symbols: getSymbols(state, selectedSource && selectedSource.toJS()),
+      symbols: "foobar",
       selectedSource
     };
   },
```

And I get no errors.

I have a solution for this that I'm working out on perf.html if you all are interested in the fix, I can provide my solution. Unfortunately it's a large amount of refactor work that will take a day or two of work to clean up with some code churn on every component.

jasonLaster:
Thanks for filing. Please share your solution :)

gregtatum:
Ok, my solution can be tracked with this issue: https://github.com/devtools-html/perf.html/issues/703

I'll post back a recommendation when I'm done.

ahuth:
Will take a look, as well, later this week. If you get something done first, @gregtatum, than that's great, too!

AnshulMalik:
Had a look on this, 

```diff --git a/src/components/PrimaryPanes/Outline.js b/src/components/PrimaryPanes/Outline.js
index 8d1fe746..764a060d 100644
--- a/src/components/PrimaryPanes/Outline.js
+++ b/src/components/PrimaryPanes/Outline.js
@@ -14,6 +14,9 @@ import Svg from "../shared/Svg";
 import "./Outline.css";
 import PreviewFunction from "../shared/PreviewFunction";
 import { uniq } from "lodash";
+
+import type { Connector } from "react-redux";
+import type { Dispatch } from "redux";
 import type {
[...]

PR #6551, opened 2018-06-26
[flow] type react-redux

jasonLaster:
Fixes Issue: #5072

### Summary of Changes

* Adds flow support for react-redux
* Fixes all the ambiguous flow types - i like that one consequence is that we now only pass in the actions that we use
* This inspires me to add coverage to some components and component types that are lacking

We now catch warnings like this!

screen shot 2018-06-26 at 9 41 26 am

### DIFF
--- a/src/components/PrimaryPanes/SourcesTree.js
+++ b/src/components/PrimaryPanes/SourcesTree.js
@@ -20,13 +20,7 @@ import {
   getSourceCount
 } from "../../selectors";
 
-// Actions
-import { setExpandedState } from "../../actions/source-tree";
-import { selectSource } from "../../actions/sources";
-import {
-  setProjectDirectoryRoot,
-  clearProjectDirectoryRoot
-} from "../../actions/ui";
+import actions from "../../actions";
 
 // Components
 import SourcesTreeItem from "./SourcesTreeItem";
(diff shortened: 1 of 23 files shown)


src/components/PrimaryPanes/SourcesTree.js
```diff
@@ -57,17 +51,17 @@ import type { SourcesMap } from "../../reducers/types";
...
   shownSource?: string,
   selectedSource?: Source,
   debuggeeUrl: string,
   projectRoot: string,
-  expanded?: boolean
+  expanded?: boolean,
+  selectSource: string => *,
+  setExpandedState: (Set<string>) => *,
+  setProjectDirectoryRoot: string => *,
+  clearProjectDirectoryRoot: void => *
```
jasonLaster: we dont care about the return type

src/components/SecondaryPanes/Breakpoints/Breakpoint.js
```diff
@@ -39,12 +39,12 @@ type Props = {
...
   selectedSource: ?Source,
-  disableBreakpoint: Function,
+  source: Source,
+  frame: ?Frame,
   enableBreakpoint: Function,
   removeBreakpoint: Function,
-  selectSpecificLocation: Function,
-  source: Source,
-  frame: ?Frame
+  disableBreakpoint: Function,
```
jasonLaster: we should stop being lazy and using Function. The best would be `typeof actions.disableBreakpoint` where flow does the work for us

julienw: note `typeof` doesn't work well with thunk actions; but yes otherwise

src/components/PrimaryPanes/SourcesTree.js
julienw: then it's better to use "any" or "mixed" than "*", if you don't care. With "*" you tell Flow to infer it for you from what is passed, which could be an issue if you have several instances of the component where you pass different functions returning different things.
I'd suggest "mixed".

jasonLaster: okay, makes sense

src/components/SecondaryPanes/Breakpoints/Breakpoint.js
jasonLaster: it might still capture arguments, right?

julienw: yeah it should be fine if you're only interested in the arguments.

julienw: mmm nope sorry: Flow might prevent you from passing the function, because `typeof thunkAction` returns a function, while the function you pass to your component returns whatever _that_ function returns. This is a case where `typeof` is too precise if you only care about the arguments.
This works for simple actions because what the action creator returns is what the `dispatch(action)` call returns too.

But please check what I'm saying against a real-world use case.

PR #6553, opened 2018-06-26
[breakpoints] Fix timing issue with syncing breakpoints

jasonLaster:
### Summary of Changes

* fixes an intermittent with [dbg-navigation][dbg] [graph][g]
* The underlying issue is that the test does 3 navigates, the second navigate starts before we are done syncing a breakpoint, the result is that a source is not available mid-sync and the command fails. 

#### Lessons

It's hard to see a timing issue and not have some thoughts:

1. the flow types are really helpful - we should continue investing
2. the code assumes via `getSourceFromId` that we will definitely have it, that is a bit naive. It is better to be conservative.
3. it would be cool if before navigating we we could cancel in progress actions 
4. alternatively, it would be cool if we could see that there were in progress actions and wait for them to finish...

these are just thoughts... largely, i think our system is getting more stable and it is getting harder to introduce these issues

[dbg]: https://bugzilla.mozilla.org/show_bug.cgi?id=1453815
[g]: https://treeherder.mozilla.org/intermittent-failures.html#/bugdetails?startday=2018-06-18&endday=2018-06-24&tree=trunk&bug=1453815

### DIFF
--- a/src/actions/sources/newSources.js
+++ b/src/actions/sources/newSources.js
@@ -136,10 +136,10 @@ function checkPendingBreakpoints(sourceId: string) {
     // load the source text if there is a pending breakpoint for it
     await dispatch(loadSourceText(source));
 
-    const pendingBreakpointsArray = pendingBreakpoints.valueSeq().toJS();
-    for (const pendingBreakpoint of pendingBreakpointsArray) {
-      await dispatch(syncBreakpoint(sourceId, pendingBreakpoint));
-    }
+    const breakpoints = pendingBreakpoints.valueSeq().toJS();
+    await Promise.all(
+      breakpoints.map(bp => dispatch(syncBreakpoint(sourceId, bp)))
+    );
   };
 }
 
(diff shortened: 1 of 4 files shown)


src/actions/sources/newSources.js
```diff
@@ -136,10 +136,12 @@ function checkPendingBreakpoints(sourceId: string) {
...
-    const pendingBreakpointsArray = pendingBreakpoints.valueSeq().toJS();
-    for (const pendingBreakpoint of pendingBreakpointsArray) {
-      await dispatch(syncBreakpoint(sourceId, pendingBreakpoint));
-    }
+    const breakpoints = pendingBreakpoints.valueSeq().toJS();
+    await Promise.all(
+      breakpoints.map(breakpoint =>
+        dispatch(syncBreakpoint)(sourceId, breakpoint)
+      )
+    );
```
jasonLaster: I'm switching to `all` so that every sync is started at the same time.

PR #6556, opened 2018-06-27
Rollback PR #6335.

nchevobbe:
This introduced a regression in the ObjectInspector (see [Bug 1469959](https://bugzilla.mozilla.org/show_bug.cgi\?id\=1469959)).
Let's see how to handle the removal of explicitOriginalTarget later.

### DIFF
--- a/packages/devtools-components/src/tree.js
+++ b/packages/devtools-components/src/tree.js
@@ -823,14 +823,14 @@ class Tree extends Component {
             return;
           }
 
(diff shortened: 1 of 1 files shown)


jasonLaster:
the sourcemaps-reload.js test is currently flakey in travis, but fine in mc :/

nchevobbe:
The patch is already in Nightly and uplifted, it looks like we should be safe to merge it here ?

PR #6558, opened 2018-06-27
[breakpoints] Improve perf of toggle and removing breakpoints

darkwing:
Let's see how this works.

### DIFF
--- a/src/actions/breakpoints.js
+++ b/src/actions/breakpoints.js
@@ -270,11 +270,12 @@ export function toggleBreakpoints(
  */
 export function removeAllBreakpoints() {
   return async ({ dispatch, getState }: ThunkArgs) => {
-    const promises = getBreakpoints(getState())
+    const breakpointList = getBreakpoints(getState())
       .valueSeq()
-      .toJS()
-      .map(([, breakpoint]) => dispatch(removeBreakpoint(breakpoint.location)));
-    await Promise.all(promises);
+      .toJS();
+    return Promise.all(
+      breakpointList.map(bp => dispatch(removeBreakpoint(bp.location)))
+    );
   };
 }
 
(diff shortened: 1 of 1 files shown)


src/actions/breakpoints.js
```diff
@@ -266,10 +267,10 @@ export function toggleBreakpoints(
  */
 export function removeAllBreakpoints() {
   return async ({ dispatch, getState }: ThunkArgs) => {
-    const breakpoints = getBreakpoints(getState());
-    for (const [, breakpoint] of breakpoints) {
-      await dispatch(removeBreakpoint(breakpoint.location));
-    }
+    const promises = [...getBreakpoints(getState())].map(([, breakpoint]) =>
```
jasonLaster: other places we do `valueSeq().toJS()` is this the equivelant?

src/actions/breakpoints.js
```diff
@@ -248,13 +248,14 @@ export function toggleBreakpoints(
...
-    for (const [, breakpoint] of breakpoints) {
+    const promises = [...breakpoints].map(([, breakpoint]) => {
       if (shouldDisableBreakpoints) {
-        await dispatch(disableBreakpoint(breakpoint.location));
-      } else {
-        await dispatch(enableBreakpoint(breakpoint.location));
+        return dispatch(disableBreakpoint(breakpoint.location));
       }
-    }
+      return dispatch(enableBreakpoint(breakpoint.location));
```
jasonLaster: this is a minor nit, but we can do something cute like:

```js
...breakpoints].map(([, breakpoint]) => 
  shouldDisableBreakpoints 
   ? dispatch(disableBreakpoint(breakpoint.location))
   : dispatch(enableBreakpoint(breakpoint.location));
)
```

Issue #6544, opened 2018-06-25
"Search in file" number of occurrences does not update when switching file


nchevobbe:
**Steps to reproduce**

1. Go to http://todomvc.com/examples/angularjs/#/
1. Open the debugger
1. Open todoStorage.js
1. Open app.js
1. Search (cmd + F) for angular
1. There should be 3 results
1. Switch to todoStorage.js

**Expected results**

The number of occurrences is updated to 10

**Actual results**

The number of occurrences is not updated and still says 3, until you hit enter again (which then shows 10 results). See following GIF: 

![Jun-25-2018 16-00-37.gif](https://shipusercontent.com/6be56af66bbf7b2cc4519919f01faf7f/Jun-25-2018%2016-00-37.gif)

darkwing:
Somewhat related:  https://github.com/devtools-html/debugger.html/issues/6466

PR #6559, opened 2018-06-27
[Sources] Update file search results when source changes

kenjiO:
Fixes Issue: #6544

Can you let me know if I am on the right track by doing it this way?

### DIFF
--- a/src/actions/sources/select.js
+++ b/src/actions/sources/select.js
@@ -100,7 +100,56 @@ export function selectSource(sourceId: string) {
  * @static
  */
 export function selectLocation(location: Location) {
-  return locationSelect(location, {checkPrettyPrint: true});
+  return async ({ dispatch, getState, client }: ThunkArgs) => {
+    const currentSource = getSelectedSource(getState());
+
+    if (!client) {
+      // No connection, do nothing. This happens when the debugger is
+      // shut down too fast and it tries to display a default source.
+      return;
+    }
+
+    const source = getSource(getState(), location.sourceId);
+    if (!source) {
+      // If there is no source we deselect the current selected source
+      return dispatch(clearSelectedLocation());
+    }
+
+    const activeSearch = getActiveSearch(getState());
+    if (activeSearch !== "file") {
+      dispatch(closeActiveSearch());
+    }
+
+    dispatch(addTab(source.url, 0));
+    dispatch(setSelectedLocation(source, location));
+
+    await dispatch(loadSourceText(source));
+    const loadedSource = getSource(getState(), source.id);
+
+    if (!loadedSource) {
+      // If there was a navigation while we were loading the loadedSource
+      return;
+    }
+
+    if (
+      prefs.autoPrettyPrint &&
+      !getPrettySource(getState(), loadedSource.id) &&
+      shouldPrettyPrint(loadedSource) &&
+      isMinified(loadedSource)
+    ) {
+      await dispatch(togglePrettyPrint(loadedSource.id));
+      dispatch(closeTab(loadedSource.url));
+    }
+
+    dispatch(setSymbols(loadedSource.id));
+    dispatch(setOutOfScopeLocations());
+
+    // If a new source is selected update the file search results
+    const newSource = getSelectedSource(getState());
+    if (!currentSource || (newSource && currentSource.id != newSource.id)) {
+      dispatch(clearFileSearch());
+    }
+  };
 }
 
 /**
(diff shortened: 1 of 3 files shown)


src/actions/sources/select.js
```diff
@@ -137,6 +141,19 @@ export function selectLocation(location: Location) {
...
+    const isLocationChanged =
+      !previousLocation || previousLocation.sourceId !== location.sourceId;
+    if (isLocationChanged) {
+      const isFileSearchOpen = getActiveSearch(getState()) === "file";
+      const fileSearchQuery = getFileSearchQuery(getState());
+      if (isFileSearchOpen && fileSearchQuery) {
+        const editor = getEditor();
+        dispatch(searchContents(fileSearchQuery, editor));
+      }
+    }
```
jasonLaster: lets try and simplify this a bit...

```js
const newSource = getSelectedSource(getState());
if (currentSource != newSource) {
  dispatch(clearFileSearch())
}
```

we can do a couple of things in `clearFileSearch`

1. check if there is an active file search
2. dispatch `searchContents`

src/actions/sources/select.js
```diff
@@ -142,17 +144,10 @@ export function selectLocation(location: Location) {
...
-    if (isLocationChanged) {
-      const isFileSearchOpen = getActiveSearch(getState()) === "file";
-      const fileSearchQuery = getFileSearchQuery(getState());
-      if (isFileSearchOpen && fileSearchQuery) {
-        const editor = getEditor();
-        dispatch(searchContents(fileSearchQuery, editor));
-      }
+    // If a new source is selected update the file search results
+    const newSource = getSelectedSource(getState());
+    if (!currentSource || (newSource && currentSource.id != newSource.id)) {
```
kenjiO: I tried doing `(currentSource != newSource)` to check if the source was changed but that was evaluating to true even when the source wasn't changed.  So I used this instead

kenjiO:
I noticed `selectLocation` and `selectSpecificLocation` are the same except one has an if statement that does some prettyPrint actions.  Do you want to refactor this so there is not so much repeated code?

jasonLaster:
Take a stab. :)

kenjiO:
How does this look?

src/actions/sources/select.js
```diff
@@ -96,87 +100,15 @@ export function selectSource(sourceId: string) {
...
-      isMinified(loadedSource)
-    ) {
-      await dispatch(togglePrettyPrint(loadedSource.id));
-      dispatch(closeTab(loadedSource.url));
-    }
-
-    dispatch(setSymbols(loadedSource.id));
-    dispatch(setOutOfScopeLocations());
-  };
+  return locationSelect(location, {checkPrettyPrint: true});
```
jasonLaster: looks good, maybe instead of `locatonSelect`, we add a param  here `selectLocation(location: Location, {checkPrettyPrint: true})`

also lets do this in a separate PR

kenjiO:
I took out the refactor and will do that in a separate PR.  Is there anything else that I should change for this PR?

src/actions/sources/select.js
```diff
@@ -137,6 +143,12 @@ export function selectLocation(location: Location) {
 
     dispatch(setSymbols(loadedSource.id));
     dispatch(setOutOfScopeLocations());
+
+    // If a new source is selected update the file search results
+    const newSource = getSelectedSource(getState());
+    if (!currentSource || (newSource && currentSource.id != newSource.id)) {
```
jasonLaster: maybe we can be even simpler. This will say, there had to have been a current source and it can't be the same as the new one. I don't know if the IDs matter

```js
if (currentSource && currentSource !== newSource) {
}
```

src/actions/sources/select.js
```diff
@@ -227,3 +247,17 @@ export function jumpToMappedSelectedLocation() {
...
+function clearFileSearch() {
+  return async ({ dispatch, getState }: ThunkArgs) => {
+    const isFileSearchOpen = getActiveSearch(getState()) === "file";
+    const fileSearchQuery = getFileSearchQuery(getState());
+    if (isFileSearchOpen && fileSearchQuery) {
+      const editor = getEditor();
+      dispatch(searchContents(fileSearchQuery, editor));
+    }
+  };
+}
```
jasonLaster: lets move this to a different action file that is focused on search

src/actions/sources/select.js
kenjiO: I tried it this way but it would sometimes return true even when the source did not change, such as when clicking on the editor tab or source file of the file that is already active.  The objects were different even though they were pointing to the same source.
screen shot 2018-07-05 at 10 44 21 am

src/actions/sources/select.js
kenjiO: I moved it to the ui.js actions file because that has the similar actions `closeActiveSearch` and `setActiveSearch`.  But it if you prefer it being in the file-seach.js actions file I could move it there instead.  I also renamed the action to `updateActiveFileSeach` since I think that better describes what it does rather than `clearFileSearch`

src/actions/sources/select.js
jasonLaster: nice find. try changing this and see if that is still the case. If it is, lets figure out why the object is changing!

```diff

diff --git a/src/reducers/sources.js b/src/reducers/sources.js
index bb18ebc..a74bf7c 100644
--- a/src/reducers/sources.js
+++ b/src/reducers/sources.js
@@ -416,15 +416,7 @@ export const getSelectedSource = createSelector(
       return;
     }

-    const source = sources[selectedLocation.sourceId];
-
-    // TODO: remove this when the immutable refactor lands in m-c
-    if (isTesting) {
-      const testSource: any = { ...source, get: field => source[field] };
-      return testSource;
-    }
-
-    return source;
+    return sources[selectedLocation.sourceId];
   }
 );
```

jasonLaster: just noticed this should be ` if (isTesting()) {`

kenjiO: I added the parentheses and that fixed it and now comparing for the same object works :)

jasonLaster:
Wow! I love how clean this PR became and how we uncovered a bug from the immutable refactoring work

PR #6561, opened 2018-06-27
Use Promise.all for checkPendingBreakpoints

darkwing:
I can't think of a reason these needs to be synced in any order so throwing this out there.


darkwing:
Closing that because Jason had this in another PR

PR #6568, opened 2018-06-28
[Editor] Type editor

jasonLaster:
### Summary of Changes

* adds flow types for Editor/index
* there is a funny thing with `nextProps.selectedSource` where we had to refactor to help flow
* the main issue now is that codeMirror is not typed

### DIFF
--- a/src/components/Editor/index.js
+++ b/src/components/Editor/index.js
@@ -66,6 +66,7 @@ import "./Highlight.css";
 
 import type SourceEditor from "../../utils/editor/source-editor";
 import type { SymbolDeclarations } from "../../workers/parser";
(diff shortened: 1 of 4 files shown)


src/components/Editor/index.js
```diff
@@ -332,7 +342,7 @@ class Editor extends PureComponent<Props, State> {
     // ignore right clicks in the gutter
     if (
       (ev.ctrlKey && ev.button === 0) ||
-      ev.which === 3 ||
+      ev.button === 2 ||
```
AnshulMalik: Is this required?

jasonLaster: yes... i checked and it was the same.

PR #6567, opened 2018-06-28
[Telemetry] Fixed telemetry import to point at devtools module using transform-mc.js

MikeRatcliffe:
Fixed telemetry import to point at devtools module using transform-mc.js

PR #6571, opened 2018-06-28
[Telemetry] Add recordTelemetryEvent() helper to debugger util fixes #6529

### LINKED ISSUE #6529: Add `recordTelemetryEvent()` helper to debugger util
This helper will allow us to log simple telemetry events using `recordTelemetryEvent("add_breakpoint")` etc.

Internally this would be something like this:

```js
import { Telemetry } from "devtools-modules";

const telemetry = new Telemetry();

export function recordTelemetryEvent(eventName, fields = {}) {
  telemetry.recordEvent("devtools.main", eventName, "debugger", null, {
    "session_id": this.toolbox.sessionId,
    ...fields
  });
}
```

MikeRatcliffe:
# requires: https://github.com/devtools-html/debugger.html/pull/6567

This greatly simplifies debugger telemetry... of course, the appropriate events need to be present inside `Events.yaml` for them to work.

### Usage

```js
import { recordEvent } from "src/utils/telemetry"; // Correct the path as appropriate

// Event without extra properties

// Event with extra properties
recordEvent("pause", {
  "reason": "debugger-statement",
  "collapsed_callstacks": 1
});
```

### DIFF
--- a/src/utils/telemetry.js
+++ b/src/utils/telemetry.js
@@ -1,0 +1,68 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
+
+/**
+ * Usage:
+ *
+ * import { recordEvent } from "src/utils/telemetry";
+ *
+ * // Event without extra properties
+ * recordEvent("add_breakpoint");
+ *
+ * // Event with extra properties
+ * recordEvent("pause", {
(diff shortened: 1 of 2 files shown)


MikeRatcliffe:
I moved it to it's own module and created a test.

spyOn didn't work... no idea why but a common problem with external modules so I mocked devtools-modules::Telemetry instead and that works just fine.

MikeRatcliffe:
Fixed flow type annotations.

MikeRatcliffe:
Corrected usage information (in telemetry.js comments).

PR #6581, opened 2018-06-29
Fix #6503 SourcesTree highlighting problem

### LINKED ISSUE #6503: [SourcesTree] SourcesTree & Reveal in Tab Weirdness
Reveal in tab doesn't work for the following URLs (using https://davidwalsh.name as the debuggee URL):

- https://davidwalsh.name/    (`(index)`)
- https://ad.doubleclick.net/ddm/adj/N366601.452584BUYSELLADS.COM/B9719141.131867274;dc_ver=42.111;sz=180x150;u_sd=2;dc_adk=2036841227;ord=g3xfdx;dc_rfl=1,https%3A%2F%2Fdavidwalsh.name%2F$0;xdt=1;crlt=9'6NAAQxjg;sttr=29;prcl=s?

After some quick sleuthing, I'm seeing getDirectories.js' `findSource` function returns:

 `Object { name: "root", path: "", contents: (8) […] }` 

...as the source to be highlighted, and thus no SourcesTree item is highlighted.  Something probably needs to be fixed in the `findSource` function.  Worth noting that both URLs are "complex"; `"(index)"` is a special case because it doesn't have a file name, and the doubleclick URL is...interesting.

darkwing:
Fixes Issue: #6503

Spent too long finding this 

/me cries

So `(index)` wasn't getting highlighted because `/davidwalsh.name/` didn't match `davidwalsh.name`

### DIFF
--- a/src/utils/sources-tree/getDirectories.js
+++ b/src/utils/sources-tree/getDirectories.js
@@ -16,7 +16,7 @@ function findSource(sourceTree: TreeDirectory, sourceUrl: string): TreeNode {
         _traverse(child);
       }
     } else if (!returnTarget) {
-      if (subtree.path.replace(/http(s)?:\//, "") == sourceUrl) {
+      if (subtree.path.replace(/http(s)?:\/\//, "") == sourceUrl) {
         returnTarget = subtree;
         return;
       }


src/utils/sources-tree/getDirectories.js
```diff
@@ -16,7 +16,7 @@ function findSource(sourceTree: TreeDirectory, sourceUrl: string): TreeNode {
         _traverse(child);
       }
     } else if (!returnTarget) {
-      if (subtree.path.replace(/http(s)?:\//, "") == sourceUrl) {
+      if (subtree.path.replace(/http(s)?:\/\//, "") == sourceUrl) {
```
jasonLaster: do you think `http(s)?:\/{1,2}` makes sense?

jasonLaster:
What is the STR to test this?

src/utils/sources-tree/getDirectories.js
jasonLaster: we should add a test for getDirectories that reproduces the bug 
https://github.com/jbhoosreddy/debugger.html/blob/sourcemap-self-url-unique/src/utils/sources-tree/tests/utils.spec.js#L155

also, i wonder why this line was written this way, is it handling a case we're missing?
could we re-write it to be more general with getURL?

Issue #6541, opened 2018-06-25
[releases] 64 slowed pausing and stepping


jasonLaster:
v64 regressed pause by 16%, step-in by 6.8%

[bugzilla](https://bugzilla.mozilla.org/show_bug.cgi?id=1467805)
[comparison](https://github.com/devtools-html/debugger.html/compare/release-63...release-64)

We think it was this [commit][c]

[c]: https://github.com/devtools-html/debugger.html/commit/3bc6b752d07e49f041961e828df7db2f5c13dbb4

jasonLaster:
@juliandescottes confirmed that it was -  Allow breakpoints in editor to always update (#6486) 
https://github.com/devtools-html/debugger.html/commit/3bc6b752d07e49f041961e828df7db2f5c13dbb4

The issue is that we now re-show breakpoints in the editor too often

jasonLaster:
This might fix it:

```diff
diff --git a/src/components/Editor/Breakpoint.js b/src/components/Editor/Breakpoint.js
index 99fc8c7..7d3d019 100644
--- a/src/components/Editor/Breakpoint.js
+++ b/src/components/Editor/Breakpoint.js
@@ -4,7 +4,7 @@

 // @flow

-import React, { Component } from "react";
+import React, { PureComponent } from "react";
 import ReactDOM from "react-dom";
 import classnames from "classnames";
 import Svg from "../shared/Svg";
@@ -33,13 +33,9 @@ type Props = {
   editor: Object
 };

-class Breakpoint extends Component<Props> {
+class Breakpoint extends PureComponent<Props> {
   addBreakpoint: Function;

-  constructor() {
-    super();
-  }
-
   addBreakpoint = () => {
     const { breakpoint, editor, selectedSource } = this.props;

```

jasonLaster:
This was fixed here, lets see if we get the boost in 68

juliandescottes:
@jasonLaster did a talos compare for 68 at https://treeherder.mozilla.org/perf.html#/compare?originalProject=try&originalRevision=d233ed9676c2abf4c22d7adf342e5c1a78b9d6de&newProject=try&newRevision=eeefa57a5246832cef51c19459d8ab654b456b07&framework=1

No improvement. But I actually don't see the commit in https://github.com/devtools-html/debugger.html/compare/release-67...release-68 ?

PR #6583, opened 2018-06-30
use PureComponent

bomsy:
Fixes Issue: #6541
### Summary of Changes

* Change to use `PureComponent`


jasonLaster:
Hey @bomsy @jbhoosreddy when you tested this, did you see the multiple re-renders go away with this change?

bomsy:
@jasonLaster Yeah i did and the multiple re-renders went away. Also the perf capture was a lot better!

Issue #6588, opened 2018-07-03
[Editor] HighlightLine updates too frequently


jasonLaster:
Recently we've seen an issue where the highlightline flashes somewhat frequently on page load, this is both a performance issue as well as a bug.

https://perfht.ml/2tRmoRP

STR:

1. go to https://andreicristianpetcu-ngcli.glitch.me/
2. select my-new-component.component.ts
3. add a breakpoint in ngOnInit
4. reload
5. remove the breakpoint

Notice the line keeps flashing

![](http://g.recordit.co/ufbdBWfFhU.gif)

![screen shot 2018-07-03 at 9 24 20 am](https://user-images.githubusercontent.com/254562/42222352-f44281aa-7ea2-11e8-80a7-03c1b6a85f33.png)

jasonLaster:
I think this should be fixed

PR #6590, opened 2018-07-03
[Editor] do not update highlight line by default

jasonLaster:
Fixes Issue: #6588

### Summary of Changes

Previously we were updating the highlight line if two conditions were not met. Now, we update it if those conditions are not met and a some others are. This will prevent spurious re-renders.

### DIFF
--- a/src/components/Editor/HighlightLine.js
+++ b/src/components/Editor/HighlightLine.js
@@ -66,7 +66,14 @@ export class HighlightLine extends Component<Props> {
       return false;
     }
 
-    return true;
+    if (
+      editorLine !== this.previousEditorLine ||
+      this.props.selectedSource !== selectedSource
+    ) {
+      return true;
+    }
+
+    return false;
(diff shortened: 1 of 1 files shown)


src/components/Editor/HighlightLine.js
```diff
@@ -66,7 +66,14 @@ export class HighlightLine extends Component<Props> {
       return false;
     }
 
-    return true;
+    if (
+      editorLine !== this.previousEditorLine ||
+      this.props.selectedSource !== selectedSource
```
bomsy: having the `this.props.selectedSource !== selectedSource` check in, seems to make the highlight flash a really quick, that sometimes you might not see the line highlighted.

jasonLaster:
We might not need this once #6601 lands

bomsy:
is this still needed?

jasonLaster:
I don't think so!

Issue #4075, opened 2017-09-18
[Accessibility] Editor Tabs and Source Tabs should be accessible


yzen:
There are several improvements that can be done for the source tabs to make them awesome for accessibility:

- [ ] Most Importantly both tab widgets (source and editor tabs) should be the same shared component component that would be styled/configured differently for a particular location.
- [ ] Reachable by keyboard: tabs should be accessible with the keyboard, right now there's not way to focus on the tab (if there is, it is not visible)
- [ ] Tabs should be navigatable via keyboard -- there's a blueprint for adding this functionality here;  https://github.com/devtools-html/debugger.html/pull/5806
- [ ] Styling for keyboard focus for tabs
- [ ] Close tab and add new (+) should be real buttons (or with role button), alternative text and be keyboard accessible: tabbable, activatable and have focus styling
- [x] source-tabs should semantically be tabs. Marco has a good [article] about this.

[article]:https://www.marcozehe.de/2013/02/02/advanced-aria-tip-1-tabs-in-web-apps/

### Editor Tabs

![Screen Shot 2017-09-18 at 8.13.49 PM.png](https://shipusercontent.com/328ae611c2f13327e81689f4bd377f9a/Screen%20Shot%202017-09-18%20at%208.13.49%20PM.png)

### Source Tabs

<img width="254" alt="screen shot 2018-04-27 at 10 33 09" src="https://user-images.githubusercontent.com/561275/39368693-453c6442-4a08-11e8-9235-a5b97e09f062.png">

jasonLaster:
@garbee looked at our tabs before and found them difficult to make accessible given the HTML structure. Does it make sense to try and refactor the HTML or do some JS controls around where tab sound take you?

Garbee:
Styling would be handled the same as buttons in most cases. As that is essentially what tabs are at the end of it all.

The specific problem in carrying this out with tabs though, is the close trigger. Those should be nested within the tab node, and buttons can't be nested in buttons. That's not nice. One solution to this though is borrowing from the linked tutorial. Have the list items, with anchors for the triggers and the close buttons as a button beside the anchor in the markup. This would allow for the anchor to be properly accessible and stylable, while allowing a button node to handle the close trigger which is what makes the most sense there. Win-Win on both sides. All the triggers are fully accessible and easily stylable.

<figure>

```html
<ul> <!-- Because tabs are lists -->
  <li>
    <a href>app-view.js</a><button>close</button>
  <li>
[...]

PR #6595, opened 2018-07-05
Accessible Tabs

ahuth:
Fixes Issue: #4075 

### Summary of Changes

- [x] Make the source tabs accessible
- [ ] Make the editor tabs accessible

### Test Plan

Tested all items in the [WAI ARIA practices section on tabs](https://www.w3.org/TR/wai-aria-practices-1.1/#tabpanel), including:

- Tabbing onto the tabs focuses the active tab
- When focused, the left and right arrow keys activates and focuses the next/previous tabs
- Tabbing from one of the tabs focuses on the next focusable element that's **_not_** a tab

### Screenshots/Videos

Hard to tell in this gif, but I'm alternating between

- Using the arrow keys to change tabs
- Using the tab and shift-tab keys to tab from and to the tabs

a11y-tabs


package.json
```diff
@@ -87,8 +87,10 @@
     "parse-script-tags": "^0.1.6",
     "pretty-fast": "^0.2.3",
     "prop-types": "^15.6.0",
-    "react": "16.2.0",
-    "react-dom": "16.2.0",
+    "react": "16.4.1",
+    "react-aria-components":
+      "https://github.com/juanca/react-aria-components.git",
+    "react-dom": "16.4.1",
```
ahuth: react-aria-components needs at least React 16.3, because it uses the new `React.createRef` api to manage focusing.

package.json
```diff
@@ -87,8 +87,10 @@
     "parse-script-tags": "^0.1.6",
     "pretty-fast": "^0.2.3",
     "prop-types": "^15.6.0",
-    "react": "16.2.0",
-    "react-dom": "16.2.0",
+    "react": "16.4.1",
+    "react-aria-components":
+      "https://github.com/juanca/react-aria-components.git",
```
ahuth: I realize we may not want to pull in another 3rd-party dependency. This one in particular is intended to be lightweight, and @juanca and I may publish components from it individually, potentially reducing even further the dependency.

Also, the tab component from this library is used in production at the company we both work for, and is "stable".

src/components/PrimaryPanes/index.js
```diff
@@ -22,6 +22,11 @@ import SourcesTree from "./SourcesTree";
 
 import type { SourcesMap } from "../../reducers/types";
 
+import Tab from "react-aria-components/src/tab";
+import Tabs from "react-aria-components/src/tabs";
+import TabList from "react-aria-components/src/tab-list";
+import TabPanels from "react-aria-components/src/tab-panels";
```
ahuth: The 4 separate imports isn't ideal, and is WIP for now. I'm intending to update react-aria-components so that we can do:

```jsx
import { Tab, Tabs, TabList, TabPanels } from 'react-aria-components/src/tabs'
```

src/components/PrimaryPanes/index.js
```diff
@@ -38,13 +43,6 @@ type Props = {
 };
 
 class PrimaryPanes extends Component<Props, State> {
-  renderShortcut: Function;
-  selectedPane: string;
-  showPane: Function;
-  renderTabs: Function;
-  renderChildren: Function;
-  onAlphabetizeClick: Function;
```
ahuth: Many of these fields don't exist anymore or are unnecessary, so I removed them. Let me know if any of these were a mistake, though!

src/components/PrimaryPanes/index.js
```diff
@@ -63,77 +61,61 @@ class PrimaryPanes extends Component<Props, State> {
...
-      return (
-        <span className="sources-header-info" dir="ltr" onClick={onClick}>
-          {L10N.getFormatStr(
-            "sources.search",
-            formatKeyShortcut(L10N.getStr("sources.search.key2"))
-          )}
-        </span>
-      );
-    }
-  };
```
ahuth: This renderShortcut function does not seem to be used anywhere, anymore, and I deleted it. Let me know if I missed something, though!

src/components/PrimaryPanes/Sources.css
```diff
@@ -146,6 +146,8 @@
   user-select: none;
   box-sizing: border-box;
   height: 30px;
+  margin: 0;
+  padding: 0;
```
ahuth: Needed to add these because the container of the tabs is a `ul` now, instead of a `div`.

webpack.config.js
```diff
@@ -40,10 +38,20 @@ const webpackConfig = {
...
     publicPath: "/assets/build"
+  },
+
+  module: {
+    rules: [
+      {
+        test: /\.js$/,
+        loader: "babel-loader",
+        include: /node_modules\/react-aria-components/
+      }
```
ahuth: Unfortunately, we still need babel to process react-aria-components, as it's non-transpiled ES6/React.

webpack.config.js
```diff
@@ -11,9 +11,7 @@ const path = require("path");
 var Visualizer = require("webpack-visualizer-plugin");
 const ObjectRestSpreadPlugin = require("@sucrase/webpack-object-rest-spread-plugin");
 
-function isDevelopment() {
-  return process.env.NODE_ENV !== "production";
-}
+const isProduction = process.env.NODE_ENV === "production";
```
ahuth: I **_think_** this check can just be a constant. Let me know if that's wrong

circle.yml
```diff
@@ -1,6 +1,6 @@
 machine:
   node:
-    version: 8.9.4
+    version: 8.11.1
```
ahuth: Seems that one of the packages being used has a requirement of ^8.11.1

package.json
```diff
@@ -87,8 +87,10 @@
     "parse-script-tags": "^0.1.6",
     "pretty-fast": "^0.2.3",
     "prop-types": "^15.6.0",
-    "react": "16.2.0",
-    "react-dom": "16.2.0",
+    "react": "16.4.1",
```
jasonLaster: We need to first try to go to react 16.4 in M-C before we do it here. Is it possible to do this work with 16.3 or 16.2?

src/components/PrimaryPanes/index.js
jasonLaster: CC @yzen, @darkwing, this is pretty cool. I don't know if we can use it yet because we need React 16.3, but wanted to share

webpack.config.js
jasonLaster: yep

jasonLaster:
CC @MikeRatcliffe what do you think about upgrading to react 16.3 or 16.4?

https://reactjs.org/blog/2018/03/29/react-v-16-3.html
https://reactjs.org/blog/2018/05/23/react-v-16-4.html

both releases will help us fix issues early on and are backward compatible

src/components/PrimaryPanes/index.js
yzen: This is really nice, just looking through the repo it seems to be doing the right stuff.

package.json
ahuth: 16.3 or 16.4 is ideal because the new ref API works really nicely with how we manage focus in the Tab component.

But if it'll be a while before being able to upgrade, I'll look into ways of using the old API to achieve the same result.

jasonLaster: @MikeRatcliffe is looking into upgrading now. I'll have more information soon

src/components/PrimaryPanes/index.js
jasonLaster: thanks

src/components/PrimaryPanes/index.js
```diff
@@ -63,77 +57,61 @@ class PrimaryPanes extends Component<Props, State> {
...
-        })}
-        onClick={() => this.showPane("outline")}
+      </Tab>,
+      <Tab
+        className={classnames("tab outline-tab", { active: isOutline })}
         key="outline-tab"
       >
         {outline}
-      </div>
+      </Tab>
```
jasonLaster: great cleanup!

webpack.config.js
jasonLaster: we might have a good way to do this here: https://github.com/devtools-html/devtools-core/blob/master/packages/devtools-launchpad/webpack.config.js#L33-L41

package.json
jasonLaster: do you have time to do a call and discuss this tomorrow?

src/components/PrimaryPanes/index.js
```diff
@@ -6,6 +6,7 @@
 
 import React, { Component } from "react";
 import { connect } from "react-redux";
+import { Tab, Tabs, TabList, TabPanels } from "react-aria-components/src/tabs";
```
jasonLaster: We have a somewhat counter intuitive way of vendoring our packages currently:

```diff
diff --git a/src/vendors.js b/src/vendors.js
index 30ee124..8687324 100644
--- a/src/vendors.js
+++ b/src/vendors.js
@@ -23,6 +23,7 @@ import * as devtoolsModules from "devtools-modules";
 import * as devtoolsUtils from "devtools-utils";
 import * as fuzzaldrinPlus from "fuzzaldrin-plus";
 import * as transition from "react-transition-group/Transition";
+import * as reactAriaComponents from "react-aria-components/src/tabs";
 import * as reselect from "reselect";
 import * as url from "url";
 import * as lodashMove from "lodash-move";
@@ -48,6 +49,7 @@ export const vendored = {
   "fuzzaldrin-plus": fuzzaldrinPlus,
   "lodash-move": lodashMove,
   "react-transition-group/Transition": transition,
+  "react-aria-components/src/tabs": reactAriaComponents,
   reselect,
   // Svg is required via relative paths, so the key is not imported path.
   // See .babel/transform-mc.js
```

the rationale is that this lets us separate the files we land directly in firefox repo and the vendor bundle we need to keep separate. Basically src & dist [here][h]

[h]: https://github.com/mozilla/gecko-dev/tree/master/devtools/client/debugger/new

jasonLaster: I think this is why the integration tests are failing

package.json
ahuth: Yes I do 👍

webpack.config.js
ahuth: Ah, I did see that but wasn't sure we wanted to add a special case there. If we're cool with that, I can certainly open up a PR for that.

src/components/PrimaryPanes/index.js
ahuth: Gotcha. Thanks for the info!

webpack.config.js
ahuth: Updated devtools-launchpad in https://github.com/devtools-html/devtools-core/pull/1075. I'll update this PR to pull in that branch for testing.

jasonLaster:
Hey @ahuth we're now using the react 16.4!

Could you rebase this branch, upgrade to `devtools-launchpad@0.0.128-rc1`, and update the webpack.config.js?

128-rc1 has `babelIncludes`, which should help https://github.com/devtools-html/devtools-core/pull/1075

ahuth:
Will do 👍 

Thanks for pulling the React upgrade through, @jasonLaster. I know that took some effort on your (and others) part.

I'll let you know when this branch is updated and testing out `devtools-launchpad@0.0.128-rc1`.

package.json
```diff
@@ -88,6 +88,7 @@
     "pretty-fast": "^0.2.3",
     "prop-types": "^15.6.0",
     "react": "16.4.1",
+    "react-aria-components": "https://github.com/juanca/react-aria-components.git",
```
ahuth: I'll publish a new version to npm tomorrow, and if we're good with this library I'll point this to that version.

jasonLaster:
looking good. The big thing for me is to better understand the impact of `react-aria-components`. Can you include a GIF of how the new new UI is more accessible?

jasonLaster:
by the way, i assume i can safely land the launchpad webpack work?

ahuth:
Sure! Tonight I'll put together a gif + some docs on how the component is more accessible.

And yes, the devtools-launchpad work is safe/good to go.

ahuth:
Alright, here's the deal with accessibility here. [Per the WAI-ARIA spec for tabs](https://www.w3.org/TR/wai-aria-practices-1.1/#tabpanel):

1. Focus moving onto the tab list focuses the active tab. We can see that in the following gif, with the sources/outline tabs (the editor tabs I haven't converted, yet). Note that the second tab is active, and tabbing onto the tabs focuses on that tab and no others.
   
   tabs2

2. When focused on the tab list, the left and right arrows change the active tab:

   arrow keys

3. Beyond that, there are a variety of roles and aria attributes per the spec that are applied.

As far as I can tell, everything is in order in terms of WAI-ARIA Authoring Practices 1.1. Please let me know if there any more details I can provide. Happy to talk about this some more!

jasonLaster:
Hey @ahuth curious what you think of this as well - https://github.com/devtools-html/debugger.html/issues/6628

ahuth:
Awesome, thanks @darkwing! I missed those visual regressions, and will look into those tonight.

@jasonLaster, regarding the hiding focus outlines on click: if doing so let's us remove `outline: 0` then we should definitely do it.

It's not my favorite technique, as seeing focus outlines doesn't bother me. Having said that, it can be a good compromise between accessibility/usability and visual preferences.

Want me to take a shot at implementing it?

jasonLaster:
@ahuth i agree. the `outline: 0` effects are bad especially when we are doing work to support keyboard navigation. I'd love it if you could help us out there afterwards.

jasonLaster:
@ahuth i just bumped the launchpad. The last question for me is if we want to pull react-aria-components in. It'd be nice if @yzen or @darkwing could take a look.

ahuth:
Thanks @jasonLaster. And I still have a couple of visual regressions on here to look into, as well.

Also figured I'd leave the editor tabs for another PR, as those are more complicated.

jasonLaster:
Sounds good

ahuth:
Quick update: I'm traveling a bit the second half of this week, and I'll come back to this then.

ahuth:
Visual regression fixes:

Here's the sources tab (has full height now):
sources

And here's the outline tab (also has full height):
outline

PR #6598, opened 2018-07-05
[SourceTree] getDirectories should compare source identities

jasonLaster:
Fixes Issue:  #5918 

### Summary of Changes

- I took a stab at cleaning up how we find a source in the tree and get the ancestor path.
- this approach removes the need for #6503
- the benefit of this approach is that we rely on source.id's as opposed to tree paths, which can be complicated.

### Test Plan

- Jest


src/utils/sources-tree/getDirectories.js
```diff
@@ -9,48 +9,42 @@ import { getURL } from "./getURL";
...
-        _traverse(child);
-      }
-    } else if (!returnTarget) {
-      if (subtree.path.replace(/http(s)?:\//, "") == sourceUrl) {
-        returnTarget = subtree;
-        return;
-      }
+function _traverse(subtree: TreeNode, source: Source) {
+  if (subtree.type === "source") {
+    if (subtree.contents.id === source.id) {
```
jasonLaster: I thought we would be able to do`subtree.contents === source`, but alas when we  blackbox a source we create a new reference...

src/utils/sources-tree/getDirectories.js
```diff
@@ -5,52 +5,46 @@
...
     }
-    directories.push(node);
+    directories.push(item);
   }
 }
+
+export function getDirectories(source: Source, sourceTree: TreeDirectory) {
+  const item = findSourceItem(sourceTree, source);
+  const ancestors = getAncestors(sourceTree, item);
+  return ancestors && ancestors.length ? ancestors : [sourceTree];
```
darkwing: This change, ensuring the `[sourceTree]` is returned if no ancestors is found, keeps the functionality.

jasonLaster:
the failing travis test was sourcemaps-reload, it should be an intermittent...

darkwing:
No idea why netlify is bombing but this is worth a look @jasonLaster

jasonLaster:
I removed netlify because we're not using it.

src/utils/sources-tree/getDirectories.js
jasonLaster: Could this be `ancestors ? ancestors : [sourceTree];` because flow tells us that getAncestors is null or an array?

PR #6600, opened 2018-07-05
[Sources] remove get shim

jasonLaster:
### Summary of Changes

this is a shim we added so that one of the console mochitests would still pass. We should remove it now that we've landed the immutable work and can easily update the MC mochitest.


jasonLaster:
I should be able to finish this now that release 69 landed

PR #6603, opened 2018-07-06
Update React

ahuth:
### Summary of Changes

- Updates React in preparation for #6595

@jasonLaster


jasonLaster:
Looks like 16.4 is going out today in MC, so we should be able to upgrade soon

ahuth:
Gotcha. Extracted refactors out to https://github.com/devtools-html/debugger.html/pull/6606

PR #6618, opened 2018-07-10
Prevent :formatted from tab names, only show mappedSource footer for original sources

darkwing:
Fixes these issues:

bugz

### DIFF
--- a/src/components/Editor/Footer.js
+++ b/src/components/Editor/Footer.js
@@ -155,7 +155,7 @@ class SourceFooter extends PureComponent<Props> {
 
   renderSourceSummary() {
     const { mappedSource, jumpToMappedLocation, selectedSource } = this.props;
-    if (mappedSource) {
+    if (mappedSource && mappedSource != selectedSource) {
       const filename = getFilename(mappedSource);
       const tooltip = L10N.getFormatStr(
         "sourceFooter.mappedSourceTooltip",
--- a/src/components/Editor/Tab.js
+++ b/src/components/Editor/Tab.js
@@ -192,7 +192,9 @@ class Tab extends PureComponent<Props> {
           source={source}
           shouldHide={icon => ["file", "javascript"].includes(icon)}
         />
-        <div className="filename">{getTruncatedFileName(source)}</div>
+        <div className="filename">
+          {getRawSourceURL(getTruncatedFileName(source))}
+        </div>
         <CloseButton
           handleClick={onClickClose}
           tooltip={L10N.getStr("sourceTabs.closeTabButtonTooltip")}


src/components/Editor/Tab.js
```diff
@@ -192,7 +192,9 @@ class Tab extends PureComponent<Props> {
           source={source}
           shouldHide={icon => ["file", "javascript"].includes(icon)}
         />
-        <div className="filename">{getTruncatedFileName(source)}</div>
+        <div className="filename">
+          {getRawSourceURL(getTruncatedFileName(source))}
```
jasonLaster: maybe we can have `getTruncatedFileName` do `getRawSourceURL` because I assume we always want to do this.

It'd be nice to have a unit test for `getTruncatedFileName` that demonstrates the behavior.
Or have a unit test for Tab.js that demonstrates it

src/components/Editor/Footer.js
```diff
@@ -155,7 +155,7 @@ class SourceFooter extends PureComponent<Props> {
 
   renderSourceSummary() {
     const { mappedSource, jumpToMappedLocation, selectedSource } = this.props;
-    if (mappedSource) {
+    if (mappedSource && mappedSource != selectedSource) {
```
jasonLaster: what's this doing?

src/components/Editor/Tab.js
darkwing: Good shout.

src/components/Editor/Footer.js
darkwing: It's preventing the "(from source.js)" message from displaying when you're on source.js.  Recently `mappedSource` has been introduce for all sources.  Still digging to find out how/when.

src/utils/source.js
```diff
@@ -156,7 +156,7 @@ export function getFilename(source: Source) {
 
   const { filename } = getURL(source);
 
-  return filename;
+  return getRawSourceURL(filename);
```
jasonLaster: should we call it before getURL i.e. `getURL(getRawSourceURL(source))`?

we could also have a param:

`getURL(..., {raw: true})`

jasonLaster: are there cases where we're using getRawSourceURL and wouldnt need it anymore after this change?
are there cases whee we're calling getFilename and dont expect it to be raw?

src/components/Editor/Footer.js
jasonLaster: perhaps we should move this into its own PR so we can look into it separately and land it separately?

src/utils/source.js
darkwing: Per your comments:

1.  We can't pass `getRawSourceURL` a source because it's used in cases where we don't have a `source` object
2.  We're currently using `getRawSourceURL` in places where it's reasonable to keep the function independent, i.e. copying source URL 
3.  It doesn't appear so

PR #6619, opened 2018-07-10
Footer.js nit

darkwing:
Just spotted this and needed to fix it ASAP


jasonLaster:
wait, how is this different?

darkwing:
It's not -- it simply removes a massive `if` block and instead returns early if the condition isn't met.

Issue #6605, opened 2018-07-06
[PrettyPrint] disable prettyprinting when sourcemaps are disabled


jasonLaster:
Pretty printing uses source maps, which means when it is disabled, we should not let users pretty print, because it will fail.

```diff
diff --git a/src/utils/source.js b/src/utils/source.js
index 623dacf..57d1db8 100644
--- a/src/utils/source.js
+++ b/src/utils/source.js
@@ -59,6 +59,7 @@ export function shouldPrettyPrint(source: Source) {
   const _isJavaScript = isJavaScript(source);
   const isOriginal = isOriginalId(source.id);
   const hasSourceMap = source.sourceMapURL;
+  const sourceMapsEnabled = prefs.sourceMaps; // or something like that?

   if (_isPretty || isOriginal || hasSourceMap || !_isJavaScript) {
     return false;
```

CC #6328

AgtLucas:
@jasonLaster Can I take this one?

jasonLaster:
Go for it!

PR #6620, opened 2018-07-10
Disable pretty-print when sourcemaps are enabled

AgtLucas:
Fixes Issue: #6605

### Summary of Changes

* Add `sourceMapsEnabled` check condition pretty-print condition.

I'm confused on how to test this, I'm running the debugger locally but I can't find settings like the way we do on Firefox. CC @jasonLaster help 😂

### DIFF
--- a/src/utils/source.js
+++ b/src/utils/source.js
@@ -17,6 +17,7 @@ import { parse as parseURL } from "url";
 import { getUnicodeUrl } from "devtools-modules";
 export { isMinified } from "./isMinified";
 import { getURL, getFileExtension } from "./sources-tree";
+import { prefs } from './prefs';
 
 import type { Source, Location } from "../types";
 import type { SourceMetaDataType } from "../reducers/ast";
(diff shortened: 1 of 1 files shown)


darkwing:
Could it be the case that someone would want to prettyprint a third party library, for example, that doesn't provide a sourcemap?  Is this a bit too general?

darkwing:
@jasonLaster thoughts on my comment above?

jasonLaster:
image

src/utils/source.js
```diff
@@ -59,8 +60,17 @@ export function shouldPrettyPrint(source: Source) {
...
+  const options = [
+    _isPretty,
+    !_isJavaScript,
+    isOriginal,
+    hasSourceMap,
+    sourceMapsEnabled
+  ];
 
-  if (_isPretty || isOriginal || hasSourceMap || !_isJavaScript) {
+  if (options.some(Boolean)) {
```
jasonLaster: I think I'd prefer something a bit more straight forward:

```js
if (
  isJavaScript(source) ||
  isOriginalId(source.id)
) {
  // stuff
} 
```

darkwing:
@AgtLucas Thank you for starting on this!  We just need @jasonLaster 's code style update as well as test updates and we're golden.  Thank you for picking this up!

AgtLucas:
@darkwing No problem! :)

I’ll update that ASAP!

src/utils/source.js
AgtLucas: @jasonLaster Is it safe to ignore `isPretty(source)`?

Issue #6608, opened 2018-07-06
[Search] after deleting a search string, the last character matches are still highlighted


kenjiO:
STR
  open the file search and type some characters. Matches will be highlighted.
  press backspace to delete all the characters you typed

ER
 After deleting all the characters from the search, there will be no matches highlighted
 
AR
 matches are still highlighted for the last character deleted

![](http://g.recordit.co/yROLvZRgE7.gif)

Jiberson:
Is it fine if I try to fix this bug?

PR #6613, opened 2018-07-07
[Editor] Footer messaging should have consistent left spacing

iisa:
Fixes Issue: #6565 

### Summary of Changes
Footer - Made `blackbox message` left padding of 2.5px that is consistent with `mapped source` button's [custom padding](https://github.com/devtools-html/debugger.html/blob/master/src/components/Editor/Footer.css#L98).

### Test Plan
Tell us a little a bit about how you tested your patch.
- [x] Go to TODO MVC
- [x] Open debugger
- [x] Toggle blackbox button on & off 
- [x] Padding between the blackbox button and messaging should no longer jump

### Screenshots/Videos (OPTIONAL)
debugger-blackbox

PR #6622, opened 2018-07-10
fix(search): highlight stays after deleting a search string

Jiberson:
Fixes Issue: #6608

### Summary of Changes

During the file search, instead of simply ignoring the case where the search string is empty, I call the clearSearch function which is imported from source-search.js.

### Test Plan

- [x] Open a source
- [x] Open the search bar
- [x] Type a word that can be found in the file
- [x] Delete it
- [x] Highlights are gone

### DIFF
--- a/src/actions/file-search.js
+++ b/src/actions/file-search.js
@@ -5,6 +5,7 @@
 // @flow
 
 import {
+  clearSearch,
   find,
   findNext,
   findPrev,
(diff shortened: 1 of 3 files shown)


src/actions/file-search.js
```diff
@@ -94,17 +95,17 @@ export function searchContents(query: string, editor: Object) {
...
       return;
     }
 
     const ctx = { ed: editor, cm: editor.codeMirror };
+
+    if (!query) {
+      clearSearch(ctx.cm, query);
+      return;
+    }
+
```
jasonLaster: i wonder if we can create a unit test for this
57c385b [Tabs] Prevent :formatted from tab names, only show mappedSource footer for original sources (#6618)
70cedc6 [Editor] Footer messaging should have consistent left spacing (#6613)

Jiberson: @jasonLaster Do you mean I should add a unit test method for this fix?
Sorry this is my very first PR on a projet, I need some clarification.

src/utils/editor/source-search.js
```diff
@@ -274,7 +274,7 @@ export function removeOverlay(ctx: any, query: string) {
  * @memberof utils/source-search
  * @static
  */
-function clearSearch(cm, query: string) {
+export function clearSearch(cm, query: string) {
```
darkwing: `clearSearch(cm: any, query: string)` will fix the flow issue

Jiberson:
I don't get why CI failed, doesn't seem linked with my commits when I look through the logs?

darkwing:
@Jiberson that's an intermittent test failure -- nothing related to your work :)

PR #6625, opened 2018-07-11
Add types to parser worker methods

ryanjduffy:
Fixes Issue: #<issue number>

I had discussed a couple options for typing the functions. The first, using a generic `Task` type, worked fine for flow typing and was more terse but didn't give useful information to developers (e.g. the flow tooltip in vscode) because it only presented the generic (e.g. `Task<[SourceId], string>`) which required understanding of the generic.

The second option is to document the function signature explicitly (e.g. `(sourceId: SourceId) => Promise<string>`). While more verbose, the type support is identical and provides a better DX so was chosen for this PR.

### Summary of Changes

* Adds types to dispatcher tasks for the parser worker
* Fixes a few related type issues
* Removes `replaceOriginalVariableName` and `getClosestExpression` worker exports which do not appear to be exist/referenced.

From VS Code, the first line of the tooltip shows the default typing for a dispatcher task which is what would be used for a task that hadn't had typing defined for it. I'm not sure why VS code displays it here. The second line (under [Flow]) shows the actual typing for this method as seen by Flow.

image

### DIFF
--- a/src/actions/ast.js
+++ b/src/actions/ast.js
@@ -20,7 +20,8 @@ import {
   getSymbols,
   findOutOfScopeLocations,
   getFramework,
-  getPausePoints
+  getPausePoints,
+  type AstPosition
 } from "../workers/parser";
 
 import { PROMISE } from "./utils/middleware/promise";
@@ -38,7 +39,9 @@ export function setSourceMetaData(sourceId: SourceId) {
     }
 
     const framework = await getFramework(source.id);
-    dispatch(updateTab(source, framework));
+    if (framework) {
+      dispatch(updateTab(source, framework));
+    }
 
     dispatch(
       ({
@@ -87,7 +90,10 @@ export function setOutOfScopeLocations() {
 
     let locations = null;
     if (location.line && source && !source.isWasm && isPaused(getState())) {
-      locations = await findOutOfScopeLocations(source.id, location);
+      locations = await findOutOfScopeLocations(
+        source.id,
+        ((location: any): AstPosition)
+      );
     }
 
     dispatch(
@@ -106,7 +112,7 @@ function compressPausePoints(pausePoints) {
     compressed[line] = {};
     for (const col in pausePoints[line]) {
       const point = pausePoints[line][col];
-      compressed[line][col] = (point.break && 1) | (point.step && 2);
+      compressed[line][col] = (point.break ? 1 : 0) | (point.step ? 2 : 0);
     }
   }
 
--- a/src/reducers/ast.js
+++ b/src/reducers/ast.js
@@ -32,7 +32,7 @@ export type SymbolsMap = Map<string, Symbols>;
 export type EmptyLinesMap = Map<string, EmptyLinesType>;
 
 export type SourceMetaDataType = {
-  framework: string | void
+  framework: ?string
 };
 
 export type SourceMetaDataMap = Map<string, SourceMetaDataType>;
@@ -221,8 +221,10 @@ export function getPausePoint(
     return;
   }
 
-  const linePoints = pausePoints[line];
-  return linePoints && linePoints[column];
+  const linePoints = pausePoints[String(line)];
+  if (linePoints && column) {
+    return linePoints[String(column)];
+  }
 }
 
 export function hasPausePoints(state: OuterState, sourceId: string): boolean {
--- a/src/workers/parser/index.js
+++ b/src/workers/parser/index.js
@@ -7,30 +7,78 @@
 import { workerUtils } from "devtools-utils";
 const { WorkerDispatcher } = workerUtils;
 
-const dispatcher = new WorkerDispatcher();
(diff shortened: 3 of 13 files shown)


src/actions/ast.js
```diff
@@ -85,8 +86,11 @@ export function setOutOfScopeLocations() {
     const source = getSourceFromId(getState(), location.sourceId);
 
     let locations = null;
-    if (location.line && source && isPaused(getState())) {
-      locations = await findOutOfScopeLocations(source.id, location);
+    if (location.line && location.column && source && isPaused(getState())) {
```
jasonLaster: why do we have to do this?
would `if (location && ..) {}` do the job?

src/actions/ast.js
```diff
@@ -105,7 +109,7 @@ function compressPausePoints(pausePoints) {
     compressed[line] = {};
     for (const col in pausePoints[line]) {
       const point = pausePoints[line][col];
-      compressed[line][col] = (point.break && 1) | (point.step && 2);
+      compressed[line][col] = (point.break ? 1 : 0) | (point.step ? 2 : 0);
```
jasonLaster: haha - but this was the most clever code I ever wrote :)

src/reducers/ast.js
```diff
@@ -221,8 +221,10 @@ export function getPausePoint(
     return;
   }
 
-  const linePoints = pausePoints[line];
-  return linePoints && linePoints[column];
+  const linePoints = pausePoints[String(line)];
+  if (linePoints && column != null) {
```
jasonLaster: I suppose `linePoints && column` would not be clear enough?

src/workers/parser/index.js
```diff
@@ -7,30 +7,84 @@
...
+  pausedPosition: AstPosition
+) => Promise<?Location>);
+
+export const clearASTs = ((dispatcher.task("clearASTs"): any): () => Promise<
+  void
+>);
+
+export const getScopes = ((dispatcher.task("getScopes"): any): (
+  location: Location
+) => Promise<SourceScope[]>);
```
jasonLaster: would this be any clearer?

```js
export const getScopes =  async (location:Location): Promise<SourceScope[]> => 
   dispatcher.invoke("getScopes", location)
```

we'd need to add a new `invoke` function to dispatcher, but shouldn't be too bad

src/workers/parser/types.js
```diff
@@ -8,4 +8,4 @@ export type AstPosition = { line: number, column: number };
 export type AstLocation = { end: AstPosition, start: AstPosition };
 
 export type PausePoint = { break: boolean, step: boolean };
-export type PausePoints = { line: { column: PausePoint } };
+export type PausePoints = { [line: string]: { [column: string]: PausePoint } };
```
jasonLaster: 👍

src/actions/ast.js
ryanjduffy: Neither are actually necessary to pass typing due to the `any` casting below. However, I added it for clarity because `Location` defines `column` as optional but `AstPosition` defines it as required so I wanted to check before blindly casting.

src/actions/ast.js
ryanjduffy: It is clever; too clever for flow it seems. :)

src/reducers/ast.js
ryanjduffy: That'd work if you prefer (assuming `column` is always non-zero).

ryanjduffy:
Good suggestion on creating new methods for the worker tasks. I think that'll be more readable than the type coersion. I've updated this PR to use the `invoke` method but I can't seem to find the source for`devtools-utils` to add that method. Mind giving me a pointer?

jasonLaster:
here ya go - https://github.com/jasonLaster/debugger.html/blob/fix-console-assignments/packages/devtools-utils/src/worker-utils.js#L23

ryanjduffy:
Pushed a commit to update `WorkerDispatcher`. Hoping to fix up the failing tests tomorrow.

jasonLaster:
Here's a conversation I had with @suchipi on the topic
https://twitter.com/suchipi/status/1017123968166195200

jasonLaster:
and a flow bug - https://github.com/facebook/flow/issues/4296

jasonLaster:
hey @ryanjduffy what's your timing here?

ryanjduffy:
@jasonLaster The failures appear to be due to making changes to `WorkerDispatcher` which aren't available to debugger.html yet until a release occurs resulting in it unable to find the new `invoke()` method.

How do you want to proceed? I could separate that into a new PR to be merged and released and then adopt it in this PR. Seems like you'd want to have some inter-module linking happening to avoid these sorts of gymnastics, though.

ryanjduffy:
Will do 🤞

src/actions/ast.js
jasonLaster: haha aww shucks

Issue #6627, opened 2018-07-11
Create Breakpoint Hover Effect


darkwing:
An awesome contributor made a valiant effort in creating a breakpoint hover effect in this PR:

#5979

Unfortunately it doesn't work properly when the editor is scrolled to the right (#6451) and creates a weird shifting side effect when code-folding is enabled.

What "breaks" the contributor's solution is that the "left" CSS value is hardcoded and thus scrolling to the right "hides" the hover because of that value -- it doesn't move with the editor scrolling.  It's also important to note that breakpoint arrow displays works properly for set breakpoints because a DOM node is introduced, whereas the contributor tried a CSS-only solution (which *is* ideal).

The old debugger did not have a CSS hover effect, nor does Chrome now.  

We *could* introduce a DOM node for the hover effect and simply move it around when the user hovers in the gutter, but even with debouncing, it will be a perf drain and could look janky.

The ideal solution is much like the contributor's but without an explicit pixel width and `left` position.

## Testing for the new solution
[ ] - Hover arrow displays properly when scrolled to the right
[ ] - Hover arrow expands with long line numbers (and wasm)
[ ] - Hover does not display if a breakpoint is already set for the given line

Issue #6585, opened 2018-07-02
[wasm] Line offset jumps if hovered by mouse


janodvarko:
STR:

1) Load http://janodvarko.cz/test/wasm/basic1/
2) Press `Load Module & Execute Exported Function` button
3) Go to the debugger and select `test.wasm` file
4) Hover mouse over any line offset number in the left source editor gutter.
5) The number jumps to the left and is only partially visible -> BUG

See the offset 0x28 on the screenshot:

![image](https://user-images.githubusercontent.com/37785/42159572-919b3faa-7df4-11e8-83a9-6ade4c3d3583.png)

Honza

jasonLaster:
This looks similar to #6451

darkwing:
Fixed by reverting the hover breakpoint effect.

PR #6632, opened 2018-07-12
Add breakpoint hover effect in gutter

pb03:
Fixes Issue: #5603, #6627

Here's the Pull Request Doc

### Summary of Changes
Add grey hover effect to the breakpoints in gutter

### Test Plan

* Open a JS file.
* Hover over the gutter line numbers.
* Line numbers should be highlighted.
* Empty lines shouldn't be highlighted.
* Existing breakpoints shouldn't be highlighted.

#### Before:
before

#### After:
ff

### DIFF
--- a/src/components/Editor/Editor.css
+++ b/src/components/Editor/Editor.css
@@ -59,22 +59,29 @@ html[dir="rtl"] .editor-mount {
   --gutter-hover-background-color: #414141;
 }
 
-:not(.empty-line):not(.new-breakpoint) > .CodeMirror-gutter-wrapper:hover {
-  width: 48px;
+:not(.empty-line):not(.new-breakpoint)
+  > .CodeMirror-gutter-wrapper:hover
+  > .CodeMirror-linenumber {
   height: 13px;
-  background-color: var(--gutter-hover-background-color) !important;
-  mask: url(/images/breakpoint.svg) no-repeat;
-  mask-size: 60px 14px;
-  mask-position: -13px 0px;
+  color: var(--theme-body-color);
+  /* Add 1px offset to the background to match it
+    with the actual breakpoint image dimensions */
+  background: linear-gradient(to bottom, transparent 1px, var(--gutter-hover-background-color) 0);
 }
 
 :not(.empty-line):not(.new-breakpoint)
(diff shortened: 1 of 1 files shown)


pb03:
This is almost the same PR which was reverted (https://github.com/devtools-html/debugger.html/pull/6626) except it fixes the bug introduced by it.

jasonLaster:
Wow! This looks great. Out of curiosity, how does this succeed when the prior approach failed? 

Couple questions:

- Did you test this while scrolling?
- Did you test this when there were less then 10 lines and more than 1000?
- Could you test this approach with a wasm file? I'm paranoid 😉 #6585

jasonLaster:
unrelated, but i wonder if this PR would now work now that we've fixed this weird effect 🤔 
https://github.com/devtools-html/debugger.html/pull/6062

jasonLaster:
dont worry about the failing test, that is a known intermittent

darkwing:
Thank you for giving this a shot @pb03 !  I'm seeing this when I pull down your code:

width

I removed the following code which prevented the odd jump:

```css
:not(.empty-line):not(.new-breakpoint)
  > .CodeMirror-gutter-wrapper:hover
  > .CodeMirror-linenumber {
  left: auto !important;
  right: 6px;
}
```

yay

Yay!  The last piece is avoiding the explicit `width`, since the line number could be very short or long (wasm).  I love it so far!

darkwing:
The other reason not setting an explicit width is important is because code-folding will extend the line number column width in some cases.

pb03:
@jasonLaster I believe previously explicit value of `left` was used to achieve an appropriate arrow size, I just removed it and fixed the shape size using `mask-size`.

@darkwing Yup, wasn't aware that gutter width can vary. 
I've a better solution in mind (without using explicit width). Let me try that and update the PR.

pb03:
fixed
It's fixed now!

jasonLaster:
Wow! @pb03 i'm learning more MASK magic ✨ in this PR

pb03:
press
Note: When selecting a breakpoint if there's wide gutter, the blue arrow gets cut.
It's a separate issue and I think it would be a good idea to show it with CSS background, instead of adding it to the DOM.

PR #6633, opened 2018-07-12
Enable code-folding and always show markers

darkwing:
I did a very scientific poll and people seem to like this feature:

https://twitter.com/davidwalshblog/status/1017040616025452544

This PR enables code-folding and always shows markers.  The default behavior is to only show when the row is hovered over, but the hit area on the arrow is small so it flickers.

My only beef with folding enabled is that the gutter now feels "noisy".  We could make the arrows .5 opacity to grey them out until hovered, but let's take this as a starting point:

folding

### DIFF
--- a/assets/panel/prefs.js
+++ b/assets/panel/prefs.js
@@ -57,7 +57,7 @@ pref("devtools.debugger.features.remove-command-bar-options", false);
 pref("devtools.debugger.features.workers", true);
 pref("devtools.debugger.features.code-coverage", false);
 pref("devtools.debugger.features.event-listeners", false);
-pref("devtools.debugger.features.code-folding", false);
+pref("devtools.debugger.features.code-folding", true);
 pref("devtools.debugger.features.outline", true);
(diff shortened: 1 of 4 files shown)


darkwing:
@violasong @digitarald Would like your input here on look, priority, etc.

jasonLaster:
great work!

I agree that it feels a bit noisy. 
- I like how Atom only shows it on gutter hover for instance. 
- It also seems to clash a bit with the gutter border. 
- Opacity could help
- I'm surprised there is a marker on line 5
- down the road, i'd soften the purple collapse indicator

darkwing:
This hasn't warranted much discussion so I'm going to close this and we can revisit later.

violasong:
(Ah, I thought I had posted a comment on Friday but I must've closed the tab before it fully posted or something)

Just wanted to suggest that we could use the [...] folded icon from Inspector; and we could also show the Inspector-style triangles right next to the first line of the code block, visible only when hovering over that first line.

Issue #6466, opened 2018-06-04
[Search] Close "Find in Files" Search and Clear Results Upon Page Change


darkwing:
STR:

1.  Use Find in Files on Site A
2.  Navigate to Site B
3.  Find in Files is still open; close it
4.  Open Find in Files again; the previous results are still there.

jasonLaster:
This should be fixed now via #6544

PR #6636, opened 2018-07-13
[Search] clear search queries and results when navigating

kenjiO:
Fixes Issue: #6466 

Here's what I have so far.  I noticed that a similar issue happens with the file-search so I am working on fixing that too and will push another commit here when I get that done.

### DIFF
--- a/src/actions/file-search.js
+++ b/src/actions/file-search.js
@@ -180,9 +180,8 @@ export function traverseResults(rev: boolean, editor: Editor) {
 
 export function closeFileSearch(editor: Editor) {
   return ({ getState, dispatch }: ThunkArgs) => {
-    const query = getFileSearchQuery(getState());
-
     if (editor) {
+      const query = getFileSearchQuery(getState());
       const ctx = { ed: editor, cm: editor.codeMirror };
       removeOverlay(ctx, query);
     }
--- a/src/actions/file-search.js
+++ b/src/actions/file-search.js
@@ -91,7 +91,7 @@ export function updateSearchResults(
 }
 
 export function clearSearchResults() {
-  return updateSearchResults(-1, -1, []);
+  return { type: "CLEAR_FILE_SEARCH_RESULTS" };
 }
 
 export function searchContents(query: string, editor: Object) {
--- a/src/actions/file-search.js
+++ b/src/actions/file-search.js
@@ -184,22 +184,11 @@ export function traverseResults(rev: boolean, editor: Editor) {
 
 export function closeFileSearch(editor: Editor) {
   return ({ getState, dispatch }: ThunkArgs) => {
-    if (editor) {
-      let codeMirror;
-      // if editor.appendToLocalElement() has not been called, editor.codeMirror
-      // throws an error.  Check for this error and if it occurs there is no
-      // need to call removeOverlay.
-      try {
-        codeMirror = editor.codeMirror;
-      } catch (err) {
-        codeMirror = null;
-      }
+    const query = getFileSearchQuery(getState());
 
-      if (codeMirror) {
-        const query = getFileSearchQuery(getState());
-        const ctx = { ed: editor, cm: codeMirror };
-        removeOverlay(ctx, query);
-      }
+    if (editor) {
+      const ctx = { ed: editor, cm: editor.codeMirror };
+      removeOverlay(ctx, query);
     }
 
     dispatch(setFileSearchQuery(""));
--- a/src/actions/file-search.js
+++ b/src/actions/file-search.js
@@ -184,11 +184,22 @@ export function traverseResults(rev: boolean, editor: Editor) {
 
 export function closeFileSearch(editor: Editor) {
   return ({ getState, dispatch }: ThunkArgs) => {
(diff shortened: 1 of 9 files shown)


src/components/ProjectSearch.js
```diff
@@ -123,6 +123,13 @@ export class ProjectSearch extends Component<Props, State> {
     shortcuts.off("Enter", this.onEnterPress);
   }
 
+  componentDidUpdate(prevProps: Props) {
+    // If the query changes in redux, also change it in the UI
+    if (prevProps.query !== this.props.query) {
+      this.setState({ inputValue: this.props.query });
```
kenjiO: I know generally deriving state from props is considered bad practice but this was the simplest way I came up with to keep the query string from persisting in the UI between navigations.  Let me know if you have other ideas of how to do this.  Or a case could me made to keep the query there if it is likely someone would want to search for the same term on a different page.

src/actions/file-search.js
```diff
@@ -189,5 +193,6 @@ export function closeFileSearch(editor: Editor) {
     dispatch(setFileSearchQuery(""));
     dispatch(closeActiveSearch());
     dispatch(clearHighlightLineRange());
+    dispatch(clearSearchResults());
```
kenjiO: I think this will also affect closing the SearchBar so I can change this if you don't won't the results cleared when the search bar is closed.

src/components/ProjectSearch.js
darkwing: @jasonLaster Do you have a suggestion here?

darkwing:
I'm having problems running tests locally at the moment but functionally this appears to work great.  I also really appreciate the test additions! :)

@kenjiO Could you look at the test failures to figure out what's up?  To do so:

```
./bin/download-firefox-artifact
yarn copy
./firefox/mach build
./firefox/mach run # for manual testing
yarn mochid # or mochih to run headless.
```

src/actions/file-search.js
```diff
@@ -179,15 +183,27 @@ export function traverseResults(rev: boolean, editor: Editor) {
...
-      removeOverlay(ctx, query);
+      let codeMirror;
+      // if editor.appendToLocalElement() has not been called, editor.codeMirror
+      // throws an error.  Check for this error and if it occurs there is no
+      // need to call removeOverlay.
+      try {
+        codeMirror = editor.codeMirror;
+      } catch (err) {
+        codeMirror = null;
+      }
```
kenjiO: An error is being thrown in the devtools version of code mirror [here](https://dxr.mozilla.org/mozilla-central/source/devtools/client/sourceeditor/editor.js#237).  If Editor.codeMirror is used before Editor._setup has been called it throws an error instead of just returning null or undefined.  I found a way to get the test to pass by adding  a try block to catch this error.  There may be a way to change the mochi tests to instantiate the editor but I couldn't figure that out yet.  In the debugger, the codeMirror is instantiated in the Editor component componentDidUpdate function.

jasonLaster: Thanks for finding the problem. I'd like to see why codeMirror might be called to soon (or too late) and find ways to prevent that)

kenjiO: Is this something I should look for in the mochi test code or the application code?

src/components/ProjectSearch.js
kenjiO: I think that `props.query` is the query used for the current search results and `state.inputValue` is the current value in the UI input box so the state is not quite redundant.  When the user starts typing in the input box, `inputValue` changes but the `query` doesn't change until the user executes the search.

src/actions/navigation.js
```diff
@@ -42,6 +44,8 @@ export function willNavigate(event: Object) {
     clearScopes();
     clearSources();
     dispatch(navigate(event.url));
+    dispatch(closeFileSearch(getEditor()));
+    dispatch(closeProjectSearch());
```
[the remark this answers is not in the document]
kenjiO: I think doing it in the reducer works well.  And it gets the mochi tests to pass!

src/actions/file-search.js
jasonLaster: maybe we should just clear the search results when we call `setFileSearchQuery`? It is nice to keep the number of discrete actions down..

src/components/ProjectSearch.js
jasonLaster: ahh thanks

src/reducers/file-search.js
```diff
@@ -87,6 +96,15 @@ function update(
       return state.setIn(["modifiers", action.modifier], actionVal);
     }
 
+    case "NAVIGATE": {
+      return state.set("query", "").set("searchResults", {
+        matches: [],
+        matchIndex: -1,
+        index: -1,
+        count: 0
+      });
```
jasonLaster: are these the default params? maybe we can just pull that out into its own object...

` state.set("query", "").set("searchResults", emptySearchResults)`

src/actions/file-search.js
kenjiO: I just realized that the reducers are now listening for the NAVIGATE action to do this. So I'll just take this out.

Issue #6589, opened 2018-07-03
Change the way we access toolbox.sessionId


MikeRatcliffe:
At the moment we do this to make the toolbox ID available:

```js
let sessionId = -1;

if (typeof window === "object" && window.parent.frameElement) {
  sessionId = window.parent.frameElement.getAttribute("session_id");
}

telemetry.recordEvent("devtools.main", eventName, "debugger", null, {
  session_id: sessionId,
  ...fields
});
```

But it is a very bad idea to rely upon the HTML structure of the toolbox in order to get the session ID.

@jasonLaster Said that we would be better creating a redux field to somehow make the toolbox session ID available everywhere.

jasonLaster:
Hmm, maybe redux doesnt make sense for a util, here's how i'd pipe the value to the module:

https://gist.github.com/jasonLaster/c3f459c8e6c66e96cedf43560bc1eb29

1. the panel.js should get the sessionId
2. onConnect should set the ID in the module
3. the module should have the ID and the telemetry object

jasonLaster:
As a side-note, it would also be nice if the telemetry instance had the ID, because that should be the same for the lifetime of the object.

> NOTE: not a priority

```js
import { Telemetry } from "devtools-modules";

const telemetry = null;

export function setupTelemetry(id: number) {
  telemetry = new Telemetry(id);
}

/**
 * @memberof utils/telemetry
 * @static
 */
export function recordEvent(eventName: string, fields: {} = {}) {
  /* eslint-disable camelcase */
  telemetry.recordEvent("devtools.main", eventName, "debugger", null, fields);
  /* eslint-enable camelcase */
}
```

MikeRatcliffe:
> As a side-note, it would also be nice if the telemetry instance had the ID, because that should be the same for the lifetime of the object.

You make a good point... if we stored the ID inside telemetry.js we wouldn't need to send it with every ping.

We still need the toolbox to get the ID in the first place though.

jasonLaster:
I think this should be done in MC first and then we can pull it in later

PR #6638, opened 2018-07-15
codeHealth(telemetry) retrieve session id from toolbox

Jiberson:
Fixes Issue: #6589

### Summary of Changes

I followed what jasonLaster proposed :

   1) the panel.js should get the sessionId
   2) onConnect should set the ID in the module
   3) the module should have the ID and the telemetry object
   4) telemetry instance has the ID

### Test Plan

See telemetry.spec.js : expect record event to be called without the session id.

### DIFF
--- a/src/utils/telemetry.js
+++ b/src/utils/telemetry.js
@@ -46,23 +46,18 @@
 
 import { Telemetry } from "devtools-modules";
 
-const telemetry = new Telemetry();
+let telemetry = null;
+
+export function setupTelemetry(id: number) {
+  telemetry = new Telemetry(id);
+}
 
 /**
  * @memberof utils/telemetry
  * @static
  */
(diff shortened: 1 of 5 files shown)


src/utils/telemetry.js
```diff
@@ -46,23 +46,18 @@
 
 import { Telemetry } from "devtools-modules";
 
-const telemetry = new Telemetry();
+let telemetry = null;
+
+export function setupTelemetry(id: number) {
+  telemetry = new Telemetry(id);
```
jasonLaster: thanks, my guess is that `Telemetry` does not currently accept `id` in the constructor and we'll have to add that. Is that correct @MikeRatcliffe?

jasonLaster:
I am going to close this as we're blocked on an improved telemetry helper in MC. We can re-open it when it lands.

Issue #6640, opened 2018-07-16
lodash-move is not properly vendored to mozilla-central


juliandescottes:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1475421 for STRs

lodash-move uses `export default` and should be required as `import move from "lodash-move";`. Instead it is required as `import { move } from "lodash-move";` and imported via wildcard in vendors.js. 

Not sure this even worked outside of m-c? 

On a sidenote, lodash-move is a pretty small util, rather than complexifying our vendoring story it could be better to simply duplicate the method in debugger.html.

jasonLaster:
> On a sidenote, lodash-move is a pretty small util, rather than complexifying our vendoring story it could be better to simply duplicate the method in debugger.html.

Yeah, i can see that point.

PR #6641, opened 2018-07-16
Fix vendoring of lodash-move

juliandescottes:
Fixes Issue: #6640

### Summary of Changes

* import lodash-move as `import move from "lodash-move"`
* import without destructuring in vendors.js


juliandescottes:
Tested with STRs from https://bugzilla.mozilla.org/show_bug.cgi?id=1475421

Issue #5833, opened 2018-03-30
[Breakpoints] Removing Breakpoints Between Original and Generated Sources Has Weird Side Effect


darkwing:
## Steps to Repeat

1.  Go to https://davidwalsh.name, open the "https://davidwalsh.name/wp-content/themes/punky/js/main.js" pretty source
2.  Create breakpoints at multiple lines: 10, 14, 16
3.  Open it's original source tab (click "From main.js" in the footer)
4.  (Notice the breakpoint marker at line 1)
5.  Click the breakpoint marker on line 1:  the breakpoint on line 10 is removed from the breakpoints pane
6.  Click line 1 again (there's no longer a marker):  the breakpoint on line 16 is removed from the breakpoints pane
7.  Click line 1 again (there's still no marker):  the breakpoint on line 14 is removed from the breakpoints pane

## Desired Behavior
I think it would be best if we removed all breakpoints on a given original source line if created generated source.

PR #6646, opened 2018-07-16
Fix 5833 - Remove all breakpoints on a line

darkwing:
Fixes Issue: #5833

There's a fair amount of repetition here but it's a good starting off point for discussion.  

## Testing
1.  Go to https://davidwalsh.name/ , open tab for `main.js`, pretty print it
2.  Create breakpoints at various lines in the pretty printed tab
3.  Switch over to the original `main.js` tab
4.  Click the breakpoint marker on line 1
5.  All breakpoint should disappear.

### DIFF
--- a/src/actions/breakpoints.js
+++ b/src/actions/breakpoints.js
@@ -429,28 +429,13 @@ export function toggleBreakpointsAtLine(line: number, column?: number) {
         addBreakpoint({
           sourceId: selectedSource.id,
           sourceUrl: selectedSource.url,
-          line: line,
-          column: column
+          line,
+          column
         })
       );
     }
 
-    const promises = bps.map(bp => {
-      if (bp.loading) {
-        return null;
-      }
-
-      return dispatch(
-        removeBreakpoint({
-          sourceId: bp.location.sourceId,
-          sourceUrl: bp.location.sourceUrl,
-          line: bp.location.line,
-          column: column || bp.location.column
-        })
-      );
-    });
-
-    return Promise.all(promises);
+    return Promise.all(bps.map(bp => dispatch(removeBreakpoint(bp.location))));
   };
 }
 
--- a/src/actions/breakpoints.js
+++ b/src/actions/breakpoints.js
@@ -14,7 +14,8 @@ import {
   getBreakpoint,
   getBreakpoints,
   getSelectedSource,
-  getBreakpointAtLocation
+  getBreakpointAtLocation,
+  getBreakpointsAtLine
 } from "../selectors";
 import { createBreakpoint, assertBreakpoint } from "../utils/breakpoint";
 import addBreakpointPromise from "./breakpoints/addBreakpoint";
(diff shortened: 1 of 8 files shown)


src/actions/breakpoints.js
```diff
@@ -407,6 +408,52 @@ export function toggleBreakpoint(line: ?number, column?: number) {
...
+        removeBreakpoint({
+          sourceId: bp.location.sourceId,
+          sourceUrl: bp.location.sourceUrl,
+          line: bp.location.line,
+          column: column || bp.location.column
+        })
+      );
+    });
+
+    return Promise.all(promises);
```
jasonLaster: I wonder if we can do something like this:

```js
return Promise.all(bps.map(bp => dispatch(removeBreakpoint(bp.location))
```

the advantage is that we are not duplicating logic

src/actions/breakpoints.js
```diff
@@ -407,6 +408,52 @@ export function toggleBreakpoint(line: ?number, column?: number) {
   };
 }
 
+export function toggleBreakpointsAtLine(line: number, column?: number) {
```
jasonLaster: do we need the column here: `column?: number`

src/actions/breakpoints.js
```diff
@@ -407,6 +408,52 @@ export function toggleBreakpoint(line: ?number, column?: number) {
...
+
+    if (bps.size === 0) {
+      return dispatch(
+        addBreakpoint({
+          sourceId: selectedSource.id,
+          sourceUrl: selectedSource.url,
+          line: line,
+          column: column
+        })
+      );
```
jasonLaster: Can we simplify this:

```js
return dispatch(addBreakpoint({ line, sourceId:  selectedSource.id }))
```

src/selectors/breakpointAtLocation.js
```diff
@@ -66,3 +76,10 @@ export function getBreakpointAtLocation(state, location) {
 
   return findBreakpointAtLocation(breakpoints, selectedSource, location);
 }
+
+export function getBreakpointsAtLine(state: OuterState, line: number) {
+  const selectedSource = getSelectedSource(state);
+  const breakpoints = getBreakpointsForSource(state, selectedSource);
+
+  return findBreakpointsAtLine(breakpoints, selectedSource, line);
```
jasonLaster: minor nit, it might be nice to inline this function as it's only used in one place:


```js
  return breakpoints.filter(
    breakpoint => getLocation(breakpoint, selectedSource).line === line
  );
```

src/test/mochitest/browser_dbg-editor-gutter.js
```diff
@@ -46,4 +46,13 @@ add_task(async function() {
...
   assertEditorBreakpoint(dbg, 4, false);
+
+  // Ensure that clicking the gutter removes all breakpoints on a given line
+  await addBreakpoint(dbg, source, 4, 0);
+  await addBreakpoint(dbg, source, 4, 1);
+  await addBreakpoint(dbg, source, 4, 2);
+  clickGutter(dbg, 4);
+  await waitForDispatch(dbg, "REMOVE_BREAKPOINT", 3);
+  is(getBreakpoints(getState()).size, 0, "No breakpoints exist");
+  assertEditorBreakpoint(dbg, 4, false);
```
jasonLaster: Nice!

jasonLaster: It'd be nice to also have a jest unit test for `toggleBreakpointsAtLine`, which will help us refactor that action if we ever touch it in the future

src/test/mochitest/browser_dbg-editor-gutter.js
```diff
@@ -46,4 +46,13 @@ add_task(async function() {
   await waitForDispatch(dbg, "REMOVE_BREAKPOINT");
   is(getBreakpoints(getState()).size, 0, "No breakpoints exist");
   assertEditorBreakpoint(dbg, 4, false);
+
+  // Ensure that clicking the gutter removes all breakpoints on a given line
+  await addBreakpoint(dbg, source, 4, 0);
+  await addBreakpoint(dbg, source, 4, 1);
+  await addBreakpoint(dbg, source, 4, 2);
+  clickGutter(dbg, 4);
+  await waitForDispatch(dbg, "REMOVE_BREAKPOINT", 3);
```
jasonLaster: I think this is one of our first tests, so it is using this older pattern of waiting for actions to fire. The problem with it is that sometimes there are timing issues around when these actions are sequenced w/ thunks. The newer style is to wait for state.

```js
await waitForState(dbg, state => getBreakpointsForLine(state, source.id, 4).size == 0)
```

it'd be nice to update the other `waitForDispatch` calls in this test as well

src/actions/breakpoints.js
darkwing: Yeah, we can tighten that up.

src/actions/breakpoints.js
darkwing: Yes, because this code also *adds* a breakpoint, so it keeps parity with `toggleBreakpoint`; I know we don't use a column breakpoint UI now but we will in the future.

src/actions/breakpoints.js
darkwing: Oh, sure.  Oops.

src/selectors/breakpointAtLocation.js
darkwing: Sure.

src/test/mochitest/browser_dbg-editor-gutter.js
darkwing: Oh nice, I'll do that too.

PR #6650, opened 2018-07-16
[Parser] teach pause points about sequence expressions

jasonLaster:
### Summary of Changes

While we were testing stepping @loganfsmyth and I found two interesting cases
`(1, f())` and `[ f() ]` both cases had some surprising results.

1. `(1, f())` => we should pause at f() not 1
2. `[ f() ]` => we should pause at f(), not [

### Test Plan

Jest

### DIFF
--- a/src/workers/parser/pausePoints.js
+++ b/src/workers/parser/pausePoints.js
@@ -124,11 +124,7 @@ function onEnter(node: BabelNode, ancestors: SimplePath[], state) {
   }
 
   if (isControlFlow(node)) {
-    if (isForStatement(node)) {
-      addStopPoint(state, startLocation);
-    } else {
-      addEmptyPoint(state, startLocation);
-    }
+    addPoint(state, startLocation, isForStatement(node));
 
     const test = node.test || node.discriminant;
     if (test) {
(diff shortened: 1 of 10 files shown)


src/workers/parser/pausePoints.js
```diff
@@ -50,6 +51,52 @@ function getStartLine(node) {
...
+
+  if (t.isObjectProperty(parentNode)) {
+    children = grandParentNode.properties.map(({ value }) => value);
+  }
+
+  if (t.isSequenceExpression(parentNode)) {
+    children = parentNode.expressions;
+  }
+
+  if (t.isCallExpression(parentNode)) {
```
loganfsmyth: It's a little hard to grasp the _why_ for some of this logic. Would this need to consider `parentNode.callee` for instance?

src/workers/parser/pausePoints.js
```diff
@@ -50,6 +51,52 @@ function getStartLine(node) {
...
 
+// Finds the first call item in a step expression so that we can step
+// to the beginning of the list and either step in or over. e.g. [], x(), { }
+function isFirstCall(node, parentNode, grandParentNode) {
+  let children = [];
+  if (t.isArrayExpression(parentNode)) {
+    children = parentNode.elements;
+  }
+
+  if (t.isObjectProperty(parentNode)) {
```
loganfsmyth: To clarify, this will allow stepping to the call in `{ prop: foo() }`, but wouldn't allow `{ prop: foo() + 1 }`, is that right?

jasonLaster: yes... this is a bit awkward I admit :/ I guess a depth first search would be best

src/workers/parser/pausePoints.js
jasonLaster: Yeah, i feel like each line could get it's own comment. Will look

src/workers/parser/pausePoints.js
jasonLaster: I'd like to switch to a smarter system in a follow up phase. I started this branch, but got a bit tripped up

https://github.com/devtools-html/debugger.html/compare/master...jasonLaster:pp-traverse?expand=1

I'll file a follow up issue

loganfsmyth: Sounds good!

PR #6655, opened 2018-07-17
[Scopes] Fully exclude Flow and TS bindings in places that were not handled properly

loganfsmyth:
Refs Issue: #6400

### Summary of Changes

* `import { type Foo } from ""` incorrectly introduced a `Foo` binding into the module scope
* Nested type binding references are incorrectly included in the scope

This change addresses both of those by much more aggressively ignoring AST nodes related to Flowtype and Typescript.


darkwing:
Test failure is an intermittent.  Kicking Travis.

Issue #6005, opened 2018-04-17
[Breakpoints] Files with the same filename


jasonLaster:
Our breakpoints are grouped by filename, unfortunately when there are two files with the same name the breakpoints are not grouped separately.

I think we should do something similar to editors and show the directory he file is in as well when there are duplicates.

<img src='https://shipusercontent.com/ba5284458b877df85ffc6627253c8682/Screen%20Shot%202018-04-17%20at%203.41.58%20PM.png' title='Screen Shot 2018-04-17 at 3.41.58 PM.png' width=362>

![](http://g.recordit.co/V7i5MQ2iz3.gif)

buttersnipps:
Hey,
Just to be clear we want the breakpoint panel to show the directory like the editor does if files are the same name?

darkwing:
@buttersnipps Eventually but for now simply showing the same file name label, with maybe a `title` attribute with the full path, would be great.

jainsneha23:
@darkwing We already have title attribute with the fullpath in the filename. Can you please tell the expected output here.

<img width="632" alt="screen shot 2018-05-03 at 5 01 07 pm" src="https://user-images.githubusercontent.com/8366397/39574157-b72363fe-4ef3-11e8-8c17-43528f91689d.png">
<img width="623" alt="screen shot 2018-05-03 at 5 02 35 pm" src="https://user-images.githubusercontent.com/8366397/39574176-d327fc2c-4ef3-11e8-90bf-3ab4df9ecb7d.png">

jasonLaster:
Hey @jainsneha23!


Oh nice point, i think this basically works. I think it would be nice to include the directory name next to the filename if they are the same. e.g:

- index.js (education)
- index.js (job)

the directory name can have a secondary grey color...

jainsneha23:
@jasonLaster Hey there can be situations with parent folder name same as well. So, we will have to check until we get different parent folder. And we can specify the same as picture below.

<img width="698" alt="screen shot 2018-05-03 at 9 24 53 pm" src="https://user-images.githubusercontent.com/8366397/39588048-9210537c-4f18-11e8-8043-81cde42144ac.png">

jasonLaster:
yeah, that kind of thing. good point. we should go up until its unique

jainsneha23:
@jasonLaster And then the same should apply on the editor as well??

jainsneha23:
@kootoopas i am working on it. 
[...]

PR #6656, opened 2018-07-18
Solves issue 6005 - files with same filename

jainsneha23:
Gets unique name with path for breakpoints

Fixes Issue: #6005

Here's the Pull Request Doc

### Summary of Changes

* Show directories for files with same names in the breakpoint list

### Test Plan

* Open a project which has files with same filenames
* Add breakpoints in line of two files with same name, and one breakpoint in a file with unique filename
* Directory should be visible with the common filenames in the breakpoint list, and should not be visible in the file with unique filename

Example test plan:

- [x] Add breakpoints
- [x] Check breakpoint list for filenames with directories

Feel free to improve it!

screen shot 2018-07-18 at 12 05 07 pm

### DIFF
--- a/src/utils/source.js
+++ b/src/utils/source.js
@@ -176,7 +176,7 @@ export function getTruncatedFileName(source: Source, length: number = 30) {
  * @static
  */
 
-export function getUniqueFileName(mySource: Source, sources: Sources) {
+export function getUniqueFileName(mySource: Source, sources: Source[]) {
   const myFileName = getFilename(mySource);
 
   let myPathSegments;
--- a/src/utils/source.js
+++ b/src/utils/source.js
@@ -182,6 +182,9 @@ export function getUniqueFileName(mySource: Source, sources: Source[]) {
   let myPathSegments;
   const openEditorPathSegmentsWithSameFilename = [];
   for (const source of sources) {
+    if (!source.url) {
+      continue;
+    }
     const fileName = getFilename(source);
     if (fileName === myFileName) {
       const idx = source.url.lastIndexOf("/");
--- a/src/utils/source.js
+++ b/src/utils/source.js
@@ -169,6 +169,51 @@ export function getTruncatedFileName(source: Source, length: number = 30) {
   return truncateMiddleText(getFilename(source), length);
 }
 
+/* Show a source url's unique filename for editor tabs, breakpoints, etc.
+ * Pass the url of file, and the list of file urls
+ *
+ * @memberof utils/source
+ * @static
+ */
+
+export function getUniqueFileName(mySource: Source, sources: Sources) {
+  const myFileName = getFilename(mySource);
+
+  let myPathSegments;
+  const openEditorPathSegmentsWithSameFilename = [];
+  for (const source of sources) {
+    const fileName = getFilename(source);
+    if (fileName === myFileName) {
+      const idx = source.url.lastIndexOf("/");
+      const pathSegments = source.url.slice(0, idx).split("/");
+      openEditorPathSegmentsWithSameFilename.push(pathSegments);
+      if (source.url === mySource.url) {
+        myPathSegments = pathSegments;
+      }
+    }
+  }
+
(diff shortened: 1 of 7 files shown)


src/utils/source.js
```diff
@@ -169,6 +169,54 @@ export function getTruncatedFileName(source: Source, length: number = 30) {
...
 }
 
+/* Show a source url's unique filename for editor tabs, breakpoints, etc.
+ * Pass the url of file, and the list of file urls
+ *
+ * @memberof utils/source
+ * @static
+ */
+
+export function getUniqueFileName(mySource: Source, sources: Source[]) {
```
jasonLaster: lets add some unit tests for this function

src/utils/source.js
```diff
@@ -169,6 +169,54 @@ export function getTruncatedFileName(source: Source, length: number = 30) {
...
+
+  let myPathSegments;
+  const openEditorPathSegmentsWithSameFilename = [];
+  for (const source of sources) {
+    if (!source.url) {
+      continue;
+    }
+    const fileName = getFilename(source);
+    if (fileName === myFileName) {
+      const idx = source.url.lastIndexOf("/");
```
jasonLaster: can we use `getURL` for some of this? (path, filename)

src/utils/source.js
```diff
@@ -169,6 +169,54 @@ export function getTruncatedFileName(source: Source, length: number = 30) {
...
+    const myPathSegment = myPathSegments[i];
+    if (
+      openEditorPathSegmentsWithSameFilename.some(
+        segments => segments.length === i + 1 || segments[i] !== myPathSegment
+      )
+    ) {
+      commonPathSegmentCount = i;
+      break;
+    }
+  }
```
jasonLaster: this feels like something we can simplify... but i don't know how yet 😉

src/utils/source.js
```diff
@@ -169,6 +169,54 @@ export function getTruncatedFileName(source: Source, length: number = 30) {
...
+  return `${myFileName} \u2014 ${[
+    ...myPathSegments.slice(commonPathSegmentCount + 1)
+  ].join("/")}`;
+}
```
jasonLaster: maybe we have this just return the common paths so on the component level we can do `${myFileName} \u2014  ${commonPaths}`

darkwing:
I *love* this!  Wow!  We should also investigate usage in tab titles.  What a great start!

src/utils/source.js
jainsneha23: nice idea. Also any nice name for the function, currently I named it 'getPath'

jasonLaster:
One style nit, I think it'd be nice to drop the `--` and instead of the path be slightly greyer

jainsneha23:
@jasonLaster As you recommended to drop the --, and instead the path be slightly greyer, you have any example in mind, any UX? All editors , browsers use - , using greyer text might not give a clear idea.

jasonLaster:
Hey @jainsneha23 I think the vscode example is pretty good. 

> All editors , browsers use -

where is that?

darkwing:
I'm so excited for this feature :D :D :D

I think one additional test that would be good would be a test that opens ~3-4 tabs, all with the same file name, but with different paths and ensure they are all as unique as they should be.

The styling of the display path can be adjusted afterward.  I think we can tie this first iteration up with @jasonLaster 's function rewrite if it works and one more test.

Amazing work @jainsneha23 !  Wow!

jainsneha23:
@jasonLaster Please check the various styles in the image, and let me know which one to incorporate.

Also I was wondering, keeping the getDisplayPath in render function will impact performance. Shouldn't this be moved somewhere else? 

For first 2 styles in below image, there will be 2 calls for each path, to check if there is any value, if yes, render.
For other 2 styles, simple space or hypen and path, it is still fine.

screen shot 2018-07-27 at 3 36 30 pm

jainsneha23:
Current style (Matches vscode) :

screen shot 2018-07-30 at 11 41 44 am

jasonLaster:
Thanks so much! This is a really nice feature :)

PR #6665, opened 2018-07-19
[Parser] tweak pause points to wrt end of functions

jasonLaster:
### Summary of Changes

Fixes some issues introduced by #6660

### DIFF
--- a/src/test/mochitest/browser_dbg-stepping.js
+++ b/src/test/mochitest/browser_dbg-stepping.js
@@ -24,6 +24,6 @@ add_task(async function test() {
   await stepIn(dbg);
   await stepIn(dbg);
 
-  assertDebugLine(dbg, 42264);
+  assertDebugLine(dbg, 42309);
   assertPausedLocation(dbg);
 });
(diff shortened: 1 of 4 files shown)


src/test/mochitest/browser_dbg-stepping.js
```diff
@@ -24,6 +24,6 @@ add_task(async function test() {
   await stepIn(dbg);
   await stepIn(dbg);
 
-  assertDebugLine(dbg, 42264);
+  assertDebugLine(dbg, 42309);
```
loganfsmyth: Is this also a case where we'd be better to remove one of the `stepIn()` calls, or is changing the line more in line with the goal of this test?

jasonLaster: perhaps, frankly i'm less interested in the final destination here and more that we do not crash and consistently get the same output.

loganfsmyth: Fair enough, just couldn't tell if there was something specific on `42264` that this was testing.

PR #6666, opened 2018-07-19
[Breakpoints] guarantee we show breakpoints in the bp list

jasonLaster:
### Summary of Changes

The way breakpointSources was setup, it was possible we would have a source that was not associated with any breakpoints. This would mean we could show a source in the bp list without any breakpoints 😱


darkwing:
@jasonLaster Is this ready?

jasonLaster:
i guess it needs a quick rebase, but yes

Issue #6617, opened 2018-07-10
[intermittent] sourcemaps-reload


jasonLaster:
[bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1434792) [tree]
	- count 25
	- started july 2nd

[tree]: https://treeherder.mozilla.org/intermittent-failures.html#/bugdetails?startday=2018-07-02&endday=2018-07-08&tree=trunk&bug=1434792

darkwing:
```shell
$ git log --after="2018-06-30" --before="2018-07-03" --oneline
636c0ff21 Add recordTelemetryEvent() helper to debugger util fixes #6529 (#6571)
94d5e0639 [Telemetry] Fixed telemetry import to point at devtools module using transform-mc.js (#6567)
990c52150 [Editor] get mode from editor (#6563)
2f990aa0e [flow] add types for Editor/index (#6568)
7f92836ba When replaying, fix resume button bug and show pause button when unpaused. (#6582)
1840902f9 [Breakpoint] use PureComponent (#6583)
```

darkwing:
Recent error output:

```shell
TEST-START browser_dbg-sourcemaps-reload.js
   Adding a new tab with URL: http://example.com/browser/devtools/client/debugger/new/test/mochitest/examples/doc-minified.html
   1 ADD_SOURCES server1.conn52.child1/source25 - math.min.js
   1 SET_WORKERS 
   1 CONNECT http://example.com/browser/devtools/client/debugger/new/test/mochitest/examples/doc-minified.html
   1 PAUSE_ON_EXCEPTIONS [start] 
   1 PAUSE_ON_EXCEPTIONS [done] 
   1 SET_EXPANDED_STATE 
   1 SET_ORIENTATION 
 >> meh
   Navigating to sourcemaps-reload/doc-sourcemaps-reload.html
   Waiting for NAVIGATE to dispatch 1 time(s)
   1 NAVIGATE 
   NAVIGATE dispatched 1 time(s)
   Waiting on sources: v1
   Waiting for state change: source v1 exists
   2 ADD_SOURCES server1.conn52.child1/source80 - sjs_code_reload.sjs?foo
   3 ADD_SOURCES 
     - server1.conn52.child1/source80/originalSource-8cbe013208dfe08fb4edebc23f0b3c67 - bootstrap 8bcf0d04c821a99f6859
[...]

PR #6672, opened 2018-07-20
[sourcemaps] update sourcemaps-reload test

jasonLaster:
Fixes Issue: #6617 

### Summary of Changes

- so after much digging i learned when things went well we would try to load the v2.bundle.js.map twice. This is no bueno.
- I have no faith in the sjs file and reloading.
- I changed the test up to be really explicit and avoid all the crazy sjs stuff


src/test/mochitest/browser_dbg-sourcemaps-reload.js
```diff
@@ -25,33 +25,29 @@ function getBreakpoints(dbg) {
...
 
-  await waitForPaused(dbg);
   await syncBp;
-  assertDebugLine(dbg, 72);
   breakpoint = getBreakpoints(dbg)[0];
 
   is(breakpoint.location.line, 9);
   is(breakpoint.generatedLocation.line, 73);
 
-  await resume(dbg);
```
jasonLaster: no more pausing, but  i dont really care. we test that elsewhere

Issue #6645, opened 2018-07-16
[releases] Use native URL class


jasonLaster:
We're currently importing the node `url` package, which is less performant than the native class. 

- We need the module for the unit tests to pass
- We should be able to use webpack's externals feature to exclude it from the bundles

Here's my first attempt at fixing it:

https://gist.github.com/jasonLaster/8aeffacb8bdc4c1e21bce1c6d06007a2

btw, the vendors module is where we put all the 3rd party packages.

Issue #5102, opened 2018-01-15
[Tabs] Sources without URLs should have a tab


digitarald:
STR:

- Open Console
- type Tdebugger`
- Hit Enter

AR: Broken UI state that doesn't show a tab for the source, scope or and pausing

![image](https://user-images.githubusercontent.com/8599/44860979-5e928800-ac2c-11e8-9950-f454d4603647.png)

jasonLaster:
There isn't a tab here because there isn't a URL associated with the script. We likely want to indicate that with a tab title: `eval (5)` or something like that.

digitarald:
Bug, UX polish or feature, @jasonLaster ?

darkwing:
Hello @atwalg2 !  How is this going?

atwalg2:
Hey @darkwing, I got pulled away from this to fix some mistakes I made on previous PR's. I hope to get back to it asap.

darkwing:
I've seen lots of stuff going in from you @atwalg2  -- awesome!  Is this still on your radar or should we unclaim it?

atwalg2:
Yea I could probably just unclaim it now

digitarald:
I just saw something similar on https://angular-quickstart.glitch.me/ . does the `ngfactory` file fall into the same issue, @jasonLaster ?

![screenflow](https://user-images.githubusercontent.com/8599/40452945-692d6424-5e98-11e8-9227-968839080ab1.gif)

jasonLaster:
looks more like this one #5663

jasonLaster:
STR:
1. in the console, evaluate `debugger`

![image](https://user-images.githubusercontent.com/254562/44860927-63b00100-ac45-11e8-8d91-7b71c0d75c42.png)

darkwing:
I've been looking into why eval`d cases weren't given a tab and I think I've found the root cause.  We're using the following selector for tabs:

```js
export const getSourceTabs = createSelector(
  getTabs,
  getSources,
  getUrls,
  (tabs, sources, urls) =>
    tabs.filter(tab =>
      getSourceByUrlInSources(sources, urls, tab.url, tab.isOriginal)
    )
);

export const getSourcesForTabs = createSelector(
  getSourceTabs,
  getSources,
  getUrls,
  (tabs, sources, urls) => {
    return tabs
      .map(tab =>
        getSourceByUrlInSources(sources, urls, tab.url, tab.isOriginal)
      )
      .filter(Boolean);
  }
);
[...]

PR #6678, opened 2018-07-22
releases : use native URL class

Jiberson:
Fixes Issue: #6645

### Summary of Changes

* Remove references to url in vendors.js and transform-mc.js
* Use webpack externals to exclude it from the bundles

### Test Plan

I'm not sure I am doing it right. Please add a comment if you see something is missing, I'll take it into account.

### DIFF
--- a/src/utils/sources-tree/getURL.js
+++ b/src/utils/sources-tree/getURL.js
@@ -4,7 +4,7 @@
 
 // @flow
 
-import { parse } from "url";
+import { parse } from "../../utils/url";
 import { getUnicodeHostname, getUnicodeUrlPath } from "devtools-modules";
 
 import type { Source } from "../../types";
--- a/src/utils/sources-tree/getURL.js
+++ b/src/utils/sources-tree/getURL.js
@@ -80,7 +80,7 @@ function _getURL(source: Source, defaultDomain: string): ParsedURL {
         filename: url
       };
 
-    case null:
+    case "":
       if (pathname && pathname.startsWith("/")) {
         // use file protocol for a URL like "/foo/bar.js"
         return {
@@ -89,7 +89,7 @@ function _getURL(source: Source, defaultDomain: string): ParsedURL {
           filename,
           group: "file://"
         };
-      } else if (host === null) {
+      } else if (!host) {
         return {
           ...def,
           path: url,
--- a/src/utils/sources-tree/tests/getUrl.spec.js
+++ b/src/utils/sources-tree/tests/getUrl.spec.js
@@ -6,6 +6,9 @@
 
 import { getURL } from "../getURL";
 import { createSource } from "../../../reducers/sources";
+import * as Url from "../../url";
+
+let spy;
 
 function createMockSource(props) {
   return createSource(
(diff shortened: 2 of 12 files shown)


jasonLaster:
looks good, not sure if the build worked though - https://travis-ci.org/devtools-html/debugger.html/builds/406901630?utm_source=github_status&utm_medium=notification.

I can help take a look tomorrow

darkwing:
I just pushed a small update that adds a utility for "safely" using `new URL()`.  I'm getting an odd error locally and want to see if Travis sees the same.

darkwing:
OK, so I figured out the build issues that's been blocking me:

https://github.com/devtools-html/devtools-core/blob/master/packages/devtools-launchpad/src/tools/mc/make-bundle.js

This file expects the export from webpack.config.js to be the webpack config, and this PR moves that to a property of what's returned, thus `bin/copy-assets.js` fails with:

```js
WebpackOptionsValidationError: Invalid configuration object. Webpack has been initialised using a configuration object that does not match the API schema.
 - configuration misses the property 'entry'.
   object { <key>: non-empty string | [non-empty string] } | non-empty string | [non-empty string] | function
   -> The entry point(s) of the compilation.
```

darkwing:
I started making some changes here but still no dice:

https://gist.github.com/darkwing/1d15a4af7eadd9dcf8278b606cb79f57

The

webpack.config.js
```diff
@@ -88,4 +88,11 @@ function buildConfig(envConfig) {
...
 
-module.exports = buildConfig(envConfig);
+module.exports = {
+  externals: {
+    url: {
+      commonjs: "URL"
+    }
+  },
+  webpackConfig: buildConfig(envConfig)
+};
```
jasonLaster: this seems weird. wouldn't we want to do externals in `buildConfig`

src/utils/url.js
```diff
@@ -0,0 +1,26 @@
...
+  search: "",
+  searchParams: {}, // This should be a "URLSearchParams" object
+  username: ""
+};
+
+export function parse(url: string): URL | object {
+  try {
+    return new URL(url);
+  } catch (err) {
+    return defaultUrl;
```
jasonLaster: at this point, we should know we have URL

webpack.config.js
darkwing: Yes, it needs to be moved there, per my gist

src/utils/url.js
darkwing: Using `new URL(input)` is dangerous, as it can throw an error which would brick out our panel.  It's the same bug I fixed for NetMonitor.  Unless we have a URL validation process, I think this is a decent bet

darkwing: This also comes to mind:  https://github.com/devtools-html/debugger.html/issues/5102

jasonLaster: got it. then maybe we should do something like `new URL(<minimal viable url>)`

jasonLaster:
I was thinking about this and I bet we can do two things here:

1. *launchpad & panel* just use `new URL` in some places
2. *jest* in tests-setup do `global.URL = require("url")`

basically, we don't need any webpack stuff

src/utils/url.js
```diff
@@ -0,0 +1,34 @@
...
+  protocol: "",
+  search: "",
+  searchParams: {}, // This should be a "URLSearchParams" object
+  username: ""
+};
+
+export function parse(url: string): URL | object {
+  try {
+    const urlObj = new URL(url);
+    urlObj.path = urlObj.pathname + urlObj.search;
```
darkwing: "path" isn't a native URL property -- I found this in URL's readme

jasonLaster: nice find

src/utils/sources-tree/tests/getUrl.spec.js
```diff
@@ -85,52 +82,4 @@ describe("getUrl", () => {
...
-      const url = getURL(source2);
-      expect(spy).toHaveBeenCalledTimes(2);
-
-      expect(url).toEqual({
-        filename: "baz.js",
-        group: "example.com",
-        path: "/foo/bar/baz.js"
-      });
-    });
-  });
```
jasonLaster: Hmm, it'd still nice to make sure we don't parse too often.

perhaps we need to spy on `./utils/url` parse. Jest has pretty good module mocking

darkwing:
@wldcordeiro @bomsy Do you have any ideas here for restoring Jest spy functionality?  I've tried a variety of items here (https://stackoverflow.com/questions/40465047/how-can-i-mock-an-es6-module-import-using-jest) and none seem to work.

src/utils/sources-tree/getURL.js
```diff
@@ -89,7 +89,7 @@ function _getURL(source: Source, defaultDomain: string): ParsedURL {
           filename,
           group: "file://"
         };
-      } else if (host === null) {
+      } else if (!host) {
```
darkwing: The `url` lib returned `null` when info didn't exist, whereas the native `URL` returns "".  TIL.

src/utils/sources-tree/tests/getUrl.spec.js
wldcordeiro: Normally when I do this kind of mocking I use `jest.mock('module')` and if I need to specifically mock one of its exports I can do `const mockSpy = jest.fn(); jest.mock('module', () => ({ theExport: mockSpy }));`

Issue #6679, opened 2018-07-23
debugger.html should use local version of devtools-utils


juliandescottes:
debugger.html depends on version 0.0.13 of devtools-utils, while the local version in packages in only version 0.0.12. This prevents the debugger from using the local version for the workspace.

We should bump the local version at https://github.com/devtools-html/debugger.html/blob/master/packages/devtools-utils/package.json#L3 and synchronize it with the version in debugger.html's main package.json.

Additionally, all packages coming from workspaces could use a more explicit version number (eg "local"), and have a comment explaining how they are resolved.

jasonLaster:
> Additionally, all packages coming from workspaces could use a more explicit version number (eg "local"), and have a comment explaining how they are resolved.

Why do you think we should have an explicit "local" version number?

juliandescottes:
> Why do you think we should have an explicit "local" version number?

For documentation purposes, that's a least the 3rd time I have been confused by the lack of documentation around the usage of "workspaces" :)

jasonLaster:
I can see that. I agree we should try and find something more obvious. Lets leave this issue open so that we can prioritize fixing it up. It would be nice to see if we can make version numbers optional.

https://twitter.com/arcanis/status/1021415846663655424

juliandescottes:
`0.0.0` works for me as well, it's pretty clear it's non-versioned.

Regarding adding comments, the `"//"` might not in the middle of the "dependencies" object, like: 

```
  "dependencies": {
    "some-npm-package": "0.10.0",
    [...]
    "//": "Warning all the 0.0.0 versioned packages below are resolved from yarn workspaces!",
    "devtools-utils": "0.0.0",
    [...]
  } 
```

To be honest I haven't tried but I thought "hmmmm this won't work in the middle of `dependencies`".

jasonLaster:
I'm going to close this and wait for yarn to let us drop versions for local packages.

PR #6680, opened 2018-07-23
Use local devtools-utils, use explicit version for workspace packages

juliandescottes:
Should fix issue #6679 

@nchevobbe I am absolutely not sure about what I did, especially specifying "workspaces" inside of the packages themselves. The one change I would love to see in the next debugger bundle is https://github.com/devtools-html/debugger.html/commit/9048eb7e0bfc88946411bde0c7af223acc30e6f8

Have fun poking at this, mobile network is not the best to `yarn install` stuff :)

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,7 +75,7 @@
     "devtools-reps": "0.23.0",
     "devtools-source-map": "0.16.0",
     "devtools-splitter": "^0.0.6",
-    "devtools-utils": "0.0.13",
+    "devtools-utils": "0.0.14",
     "enzyme-adapter-react-16": "^1.1.1",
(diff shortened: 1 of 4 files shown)


jasonLaster:
Hmm, I think I would like to keep the version numbers, but rarely update them:

1. the version numbers are simpler if we want to npm in the future
2. they're not as long :)

juliandescottes:
Let's first address the devtools-utils issue, I'd like to get this long standing issue resolved quickly.

jasonLaster:
I think it would be nice to make version numbers optional going forward
https://twitter.com/arcanis/status/1021415846663655424

PR #6682, opened 2018-07-23
[releases] exclude worker sources

jasonLaster:
### Summary of Changes

We bundle the worker files, so we should not land them in m-c because it gives the impression that they are useful and if you change the code it should be reflected in the debugger.


bin/copy-modules.js
```diff
@@ -21,6 +21,11 @@ feature.setConfig(envConfig);
 
 
 function ignoreFile(file) {
+  // We need to ignore all the files except the index.js
+  if (file.match(/\/workers/) && !file.match(/index.js/)) {
```
juliandescottes: index.js -> index\.js

bin/copy-modules.js
```diff
@@ -21,6 +21,11 @@ feature.setConfig(envConfig);
 
 
 function ignoreFile(file) {
+  // We need to ignore all the files except the index.js
```
juliandescottes: Can we add some more details about why workers are not included (they are bundled) and why the index files are included (they are API entry points)

darkwing:
What's the best way to test this?

jasonLaster:
`yarn copy`

bin/copy-modules.js
juliandescottes: erf github comments.... 

Please escape the dot

Issue #6106, opened 2018-04-26
Remove lambda chars in front of the frame function name


nchevobbe:
_From @nchevobbe on April 23, 2018 7:39_

It was added recently but some feedback suggest that it isn't clear for the user what it is about.
Let's remove it before we ship 61. We can then find a better iconography for this.
We may directly change the reps.css file in [Bug 1456061](https://bugzilla.mozilla.org/show_bug.cgi?id=1456061), so we should propagate back it here.

_Copied from original issue: devtools-html/devtools-core#1057_

PR #6687, opened 2018-07-24
Make stacktrace copy/paste more useful

nchevobbe:
### Summary of Changes

At the moment when one try to copy and paste the stacktrace
of an error object, everything gets mangled in the same line,
making it very hard to read.
This patch adds a tab character before each function name as
well as a new line after each frame so what the user gets in
the clipboard looks more like what they see on the output.

### Test Plan

1. Launch the reps test app (`cd packages/devtools-reps && yarn start`)
1. Open a firefox window and select a tab as a target
1. In the reps test app (with the purple header), click on the "errors" button. This will log 3 different error object in different mode
1. Copy an error object with a stacktrace (either small or long mode)
1. Paste it in an editor/textarea

-> It should be nicely formatted

---

A mochitest will probably be added in mozilla-central to make sure copy/pasting to work as expected (node is not ideal to est clipboard interactions)

### DIFF
--- a/packages/devtools-reps/src/reps/error.js
+++ b/packages/devtools-reps/src/reps/error.js
@@ -144,6 +144,7 @@ function getStacktraceElements(props, preview) {
     }
 
     stack.push(
+      "\t",
       span(
         {
           key: `fn${index}`,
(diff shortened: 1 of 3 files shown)


nchevobbe:
Fixes #6106

darkwing:
Yes!  I experienced this the other day and was hoping we could do better!  Nice!

Issue #6527, opened 2018-06-22
Add "pause_on_exceptions" telemetry


MikeRatcliffe:
#### HOWTO

Bugzilla bug https://bugzil.la/1463117

If all properties are available then adding the probe is easy:

```js
import { recordEvent } from "src/utils/telemetry"; // Change the path as appropriate

recordEvent("pause_on_exceptions", {
  "exceptions": true,
  "caught_exceptio": false
});
```

MikeRatcliffe:
This is linked with https://bugzil.la/1463117

jasonLaster:
maybe we can wrap recordEvent


```js
import { Telemetry } from "devtools-modules";

const telemetry = new Telemetry();

telemetry.recordEvent("devtools.main", "continue", "debugger", null, {
  "session_id": this.toolbox.sessionId
});

export recordEvent(name, fields) {
  telemetry.recordEvent("devtools.main", name, null, {
    "session_id", // somehow has this field,
    ...fields
  })
}
```

PR #6671, opened 2018-07-20
 Fix #6533 - Add simple telemetry

darkwing:
Fixes Issue: #6533

### Events to be Added

- [x]  "continue"  https://bugzil.la/1463122
- [x]  "add_breakpoint"  https://bugzil.la/1463123
- [x]  "remove_breakpoint"  https://bugzil.la/1463124
- [x]  "pretty_print"  https://bugzil.la/1463125
- [x]  "blackbox"  https://bugzil.la/1463126

PR #6689, opened 2018-07-24
Fix 1463117 - Add telemetry for pause on (caught) exceptions

darkwing:
Fixes Issue: https://bugzilla.mozilla.org/show_bug.cgi?id=1463117

## ToDo
- [x] Add test

We should wait on https://github.com/devtools-html/debugger.html/pull/6671 being merged first

### DIFF
--- a/src/actions/pause/tests/pauseOnExceptions.spec.js
+++ b/src/actions/pause/tests/pauseOnExceptions.spec.js
@@ -1,0 +1,26 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
+
+import {
+  actions,
+  createStore,
+  getTelemetryEvents
+} from "../../../utils/test-head";
+
+describe("pauseOnExceptions", () => {
+  it("should track telemetry for pauseOnException changes", async () => {
+    const { dispatch } = createStore({ pauseOnExceptions: () => {} });
+    dispatch(actions.pauseOnExceptions(true, false));
+
+    const pauseTelemetry = getTelemetryEvents("pause_on_exceptions");
+
+    expect(pauseTelemetry).toHaveLength(1);
+    /* eslint-disable camelcase */
+    expect(pauseTelemetry[0]).toEqual({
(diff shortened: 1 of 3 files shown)


src/actions/pause/tests/pauseOnExceptions.spec.js
```diff
@@ -0,0 +1,26 @@
...
+
+    expect(pauseTelemetry).toHaveLength(1);
+    /* eslint-disable camelcase */
+    expect(pauseTelemetry[0]).toEqual({
+      exceptions: true,
+      caught_exceptio: false
+    });
+    /* eslint-enable camelcase */
+  });
+});
```
jasonLaster: wow is this really the first `pauseOnExceptions`!

jasonLaster: I think it would be nice to just match snapshot `expect(getTelemetryEvents("pause_on_exceptions")).toMatchSnapshot()`. These tests are mostly about preventing regressions.

src/actions/pause/tests/pauseOnExceptions.spec.js
```diff
@@ -0,0 +1,26 @@
...
+import {
+  actions,
+  createStore,
+  getTelemetryEvents
+} from "../../../utils/test-head";
+
+describe("pauseOnExceptions", () => {
+  it("should track telemetry for pauseOnException changes", async () => {
+    const { dispatch } = createStore({ pauseOnExceptions: () => {} });
+    dispatch(actions.pauseOnExceptions(true, false));
```
jasonLaster: maybe we should also have a test for the redux selectors as well?

darkwing: Could you expound?  Which selectors?

darkwing:
Will fix https://github.com/devtools-html/debugger.html/issues/6527

src/actions/pause/tests/pauseOnExceptions.spec.js
jasonLaster:
```js
expect(shouldPauseOnExceptions(store.getState()).toBe(true);
expect(shouldPauseOnExceptionsOnUncaught(store.getState()).toBe(true);
```

The reason for this is it looks like it is our first POE test.

PR #6697, opened 2018-07-25
Disable pretty print when sourcemaps are disabled

darkwing:
Supercedes https://github.com/devtools-html/debugger.html/pull/6620

I've added style changes


AnshulMalik:
Shouldn't it be when source maps are disabled?

PR #6702, opened 2018-07-25
[sync] backport changes from release-75

jasonLaster:
### Summary of Changes

syncs work that was in release-75, that had to be backported to master


jasonLaster:
just one css lint issue

src/components/Editor/Editor.css
 69:13  ✖  Expected newline after ":" with a    declaration-colon-newline-after
           multi-line declaration

interestingly stylelint and prettier disagree

configs/mozilla-central-mappings.js
```diff
@@ -9,25 +9,26 @@ const EXCLUDED_FILES = {
   "devtools-connection": "devtools/shared/flags",
   "chrome-remote-interface": "devtools/shared/flags",
   "devtools-launchpad": "devtools/shared/flags",
+  "devtools-services": "Services"
```
juliandescottes: Why is this moved here? 

I tried to make the name and comment explicit enough so that only files mapped to "devtools/shared/flags" should be in this object.

jasonLaster: nice catch - yeah, i think i was pretty careless here...

PR #6707, opened 2018-07-26
[scopes] Test each set of fixtures across more build targets.

loganfsmyth:
Refs Issue: #6400

cc @jasonLaster 

### Summary of Changes

* Make use of the changes from #6701 to build the fixtures with each build tool
* More broadly run the tests with the various build tools to cover our bases.


src/test/mochitest/browser_dbg-sourcemapped-scopes.js
```diff
@@ -28,513 +28,1134 @@ add_task(async function() {
...
+async function testESModules(dbg) {
+  // TODO: The behavior on Webpack 3 seems to be broken.
+  await breakpointScopes(
+    dbg,
+    "webpack3",
+    "esmodules",
+    { line: 20, column: 0 },
+    [
+      "Module",
+      ["aDefault", "(optimized away)"],
```
loganfsmyth: FYI, one benefit of expanding these is that I came across this. Definitely not desirable. This should be falling back to generated bindings, but isn't. If there were more non-import bindings, it'd tip it over the edge I think, but since this file is almost exclusively import bindings and they are all detected as optimized out, it isn't falling back properly.

jasonLaster: Ooh, nice find

loganfsmyth: Fixed by #6710

PR #6710, opened 2018-07-27
Perform the line-range check later so that we still search for applicable bindings for (optimized out) handling.

loganfsmyth:
Refs Issue: #6400

cc @jasonLaster 

### Summary of Changes

* Remove the range-merging logic so we work directly with the ranges from the sourcemap, since I don't think it's actually doing anything useful.
* Move the full-line range test to _after_ we assign `hadApplicableBindings` so that bindings in full-line ranges properly cause a binding to not count as `(optimized out)`.

### DIFF
--- a/src/test/mochitest/browser_dbg-sourcemapped-scopes.js
+++ b/src/test/mochitest/browser_dbg-sourcemapped-scopes.js
@@ -722,7 +722,7 @@ async function testTypescriptClasses(dbg) {
     "webpack3",
     "rollup",
   ]) {
-    const { rollupOptimized } = targetToFlags(target);
+    const { isRollup, rollupOptimized } = targetToFlags(target);
 
     await breakpointScopes(dbg, target, "typescript-classes", { line: 50, column: 2 }, [
       "Module",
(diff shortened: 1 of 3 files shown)


src/test/mochitest/browser_dbg-sourcemapped-scopes.js
```diff
@@ -1004,30 +1010,33 @@ async function testESModules(dbg) {
...
+      "__WEBPACK_IMPORTED_MODULE_1__src_mod2__",
+      "__WEBPACK_IMPORTED_MODULE_10__src_optimized_out__",
+      "__WEBPACK_IMPORTED_MODULE_2__src_mod3__",
+      "__WEBPACK_IMPORTED_MODULE_3__src_mod4__",
+      "__WEBPACK_IMPORTED_MODULE_4__src_mod5__",
+      "__WEBPACK_IMPORTED_MODULE_5__src_mod6__",
+      "__WEBPACK_IMPORTED_MODULE_6__src_mod7__",
+      "__WEBPACK_IMPORTED_MODULE_7__src_mod9__",
+      "__WEBPACK_IMPORTED_MODULE_8__src_mod10__",
+      "__WEBPACK_IMPORTED_MODULE_9__src_mod11__",
```
jasonLaster: Are these dependencies that can't be mapped?

Issue #6704, opened 2018-07-26
Setting breakpoints in pretty-printed source code does not work


msiemens:
**Browser version:** 63.0a1 (2018-07-26) (64-Bit) on macOS 10.12.6

**Steps to reproduce:**

1. Visit http://todomvc.com/examples/vanilladart/build/web/, open the dev tools debugger and open `examples/vanilladart/dart/app.dart.js`: ![screenshot 2018-07-26 09 56 44](https://user-images.githubusercontent.com/1873922/43249216-6eec6762-90ba-11e8-9275-9301f9710b59.png)
2. Pretty-print the code: ![screenshot 2018-07-26 09 57 15](https://user-images.githubusercontent.com/1873922/43249236-7a029176-90ba-11e8-97bd-9d31f2e53cbc.png)
3. Set a breakpoint at line 3: ![screenshot 2018-07-26 09 57 17](https://user-images.githubusercontent.com/1873922/43249250-826f9084-90ba-11e8-938a-6f818a8b5b55.png)
4. Reload: ![screenshot 2018-07-26 09 57 33](https://user-images.githubusercontent.com/1873922/43249269-89d6c4a0-90ba-11e8-8e41-7afeee05ca53.png)

**Expected:** see the breakpoint and pause the debugger at this location

**Actual:** breakpoint is suddenly missing, debugger is breaking at seemingly random lines of code<sup>1</sup>

1: Seems like it steps through all lines of the formatted source code starting at line 1, which is rather unfortunate if setting a breakpoint at line 10,000+

darkwing:
Thank you for the report @msiemens !  

I followed the steps cited and:

0. (remember I added a breakpoint to line 3)
1.  The ugly source opened first, with line 1 highlighted
2.  (2 seconds pass)
3.  The pretty tab opens, with line 1 highlighted
4.  No breakpoints display in the list on the right

PR #6713, opened 2018-07-28
breakpoints set on pretty source visible on reload

AnshulMalik:
Fixes Issue: #6704 

Since `addSources` doesn't gets called for pretty sources, so we missed checking for pending breakpoints.


AnshulMalik:
Now there is one more issue that I left for another PR.

It happens when:
- Pretty print a source
- Add breakpoints on pretty source
- Close the pretty source
- Make sure that the ugly source is selected
- Reload

You'd see that we can't see the breakpoints in ugly source.

jasonLaster:
CC @darkwing

Issue #6279, opened 2018-05-11
Fix devtools.debugger.pending-breakpoints overflow issue


digitarald:
Data in `devtools.debugger.pending-breakpoints` is invalid JSON: https://gist.github.com/digitarald/a0ccbe43a1ddf4a484a7d30ea3536746

Browser Console throws this several times:

`Warning: attempting to write 7034 bytes to preference devtools.debugger.pending-breakpoints. This is bad for general performance and memory usage. Such an amount of data should rather be written to an external file. This preference will not be sent to any content processes.`

P1, because this might affect more frequent debugger users.

jasonLaster:
Here is a formatted gist.
http://jsoneditoronline.org/?id=7fa7d6b45f0d498bb609086641743abd

it's just 17 breakpoints, which is not that much

Loirooriol:
Also see #5219

bomsy:
OLD 

```js 
 "https://ton.twimg.com/tweetdeck-web/web/dist/bundle.f8cdc25670.js:formatted:5:": {
    condition: null,
    disabled: false,
    location: {
      sourceUrl:
        "https://ton.twimg.com/tweetdeck-web/web/dist/bundle.f8cdc25670.js:formatted",
      line: 5,
      column: 0
    },
    astLocation: {
      offset: {
        sourceId:
          "server1.conn9.child1/source36/originalSource-002da27b62870739bb2396d425b57db5",
        sourceUrl:
          "https://ton.twimg.com/tweetdeck-web/web/dist/bundle.f8cdc25670.js:formatted",
        line: 5,
        column: 0
      }
    },
    generatedLocation: {
      sourceUrl:
        "https://ton.twimg.com/tweetdeck-web/web/dist/bundle.f8cdc25670.js",
      line: 1,
      column: 47
    }
  }
```

NEW

```js 
 "https://ton.twimg.com/tweetdeck-web/web/dist/bundle.f8cdc25670.js:formatted:5:": {
    condition: null,
    disabled: false,
    astLocation: {
      offset: "server1.conn9.child1/source36/originalSource-002da27b62870739bb2396d425b57db5"
    },
    generatedLocation: "https://ton.twimg.com/tweetdeck-web/web/dist/bundle.f8cdc25670.js:1:47"
  }
```

Thinking about this, i feel the sample OLD json above can be reduced to NEW json above. Some key changes
[...]

PR #6719, opened 2018-07-30
[Breakpoints] Switch Pending BPs from Immutable to Pojo

jasonLaster:
issue: #6279

### Summary of Changes

* switches pending breakpoints to a plain object
* this will make it easier to persist / serialize 
* it also will help us flow type breakpoints

### Test Plan

flow and jest were clutch here

### DIFF
--- a/src/reducers/pending-breakpoints.js
+++ b/src/reducers/pending-breakpoints.js
@@ -23,31 +23,18 @@ import type { PendingBreakpoint } from "../types";
 import type { Action, DonePromiseAction } from "../actions/types";
 import type { Record } from "../utils/makeRecord";
 
-export type PendingBreakpointsMap = I.Map<string, PendingBreakpoint>;
+export type PendingBreakpointsState = { [string]: PendingBreakpoint };
 
-export type PendingBreakpointsState = {
-  pendingBreakpoints: PendingBreakpointsMap
-};
-
-export function initialPendingBreakpointsState(): Record<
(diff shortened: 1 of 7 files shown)


src/reducers/pending-breakpoints.js
```diff
@@ -87,25 +70,23 @@ function update(
 }
 
 function addBreakpoint(state, action) {
-  if (action.status !== "done") {
+  if (action.breakpoint.hidden || action.status !== "done") {
```
wldcordeiro: Slightly different condition? Does this fix a bug too?

jasonLaster: cleans up the code a bit

src/reducers/pending-breakpoints.js
```diff
@@ -21,33 +18,19 @@ import { prefs } from "../utils/prefs";
...
 
 function update(
-  state: Record<PendingBreakpointsState> = initialPendingBreakpointsState(),
+  state: PendingBreakpointsState = initialPendingBreakpointsState(),
   action: Action
 ) {
   switch (action.type) {
     case "ADD_BREAKPOINT": {
-      if (action.breakpoint.hidden) {
-        return state;
```
AnshulMalik: This is not required?

jasonLaster: It is moved to `addBreakpoint`. Thanks for asking

Issue #6718, opened 2018-07-30
[Sources] Vertical scrollbar is missing in the sourcepanel


nchevobbe:
**Steps to reproduce**
1. open the debugger on a page with a good number of sources (any todomv exa,ples should do)
2. Reduce the debugger window height so the source panel should show a scrollbar

**Expected results**

There's a vertical scrollbar

**Actual results**

There's no vertical scrollbar: 

![image](https://user-images.githubusercontent.com/578107/43402777-c7e8e758-9413-11e8-8990-f751adc592bc.png)

---

Reproduced on latest master and Nightly

darkwing:
Yep, this happened with the accessibility stuff; `.sources-panel` has `overflow: hidden;` which is causing the problem.  Looking at it now

PR #6727, opened 2018-07-30
Fix 6718 - Allow scrolling in sources tree and outline panes

darkwing:
Fixes Issue: #6718 

## Testing

1.  Open the sources pane so that there should be scrollbars.  Ensure there's only one scrollbar for vertical and one for horizontal.  Same for Outline pane.

2.  Set a directory root.  Ensure there's only one scrollbar for vertical and one for horizontal.  Same for outline pane.

### DIFF
--- a/src/components/PrimaryPanes/Sources.css
+++ b/src/components/PrimaryPanes/Sources.css
@@ -189,6 +189,7 @@
 
 .source-outline-panel {
   flex: 1;
+  overflow: auto;
 }
 
 .sources-list .managed-tree .tree .node img.blackBox {


ahuth:
Thanks for fixing this, and that's my bad for not seeing this in #6595

src/components/PrimaryPanes/Sources.css
```diff
@@ -189,6 +189,7 @@
 
 .source-outline-panel {
   flex: 1;
+  overflow: auto;
```
jasonLaster: oh interesting. I wonder why auto is not the default?

PR #6737, opened 2018-08-02
 Instrument object expansion with telemetry

nchevobbe:
This calls recordTelemetry event with the object_expanded event name
when the object inspector is expanded.
A test is added to make sure this works as expected.

Also, we now pass a recordTelemetryEvent prop to the ObjectInspector in the reps-launchpad for manual testing sake.

### DIFF
--- a/packages/devtools-reps/src/object-inspector/component.js
+++ b/packages/devtools-reps/src/object-inspector/component.js
@@ -192,7 +192,8 @@ class ObjectInspector extends Component<Props> {
       loadedProperties,
       nodeExpand,
       nodeCollapse,
-      roots
+      recordTelemetryEvent,
+      roots,
     } = this.props;
 
     if (expand === true) {
@@ -212,6 +213,10 @@ class ObjectInspector extends Component<Props> {
         createObjectClient,
         createLongStringClient
       );
+
+      if (recordTelemetryEvent) {
+        recordTelemetryEvent("object_expanded");
+      }
     } else {
       nodeCollapse(item);
     }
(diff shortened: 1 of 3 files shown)


packages/devtools-reps/src/object-inspector/component.js
```diff
@@ -212,6 +213,10 @@ class ObjectInspector extends Component<Props> {
         createObjectClient,
         createLongStringClient
       );
+
+      if (recordTelemetryEvent) {
+        recordTelemetryEvent("object_expanded");
+      }
```
jasonLaster: I think it would be nice to have a default prop here which would let you just do `recordTelemetryEvent`. There are obviously other strategies too with context or modules that would help too, so if you want to add the probes and then refactor, tht works too

packages/devtools-reps/src/object-inspector/tests/component/expand.js
```diff
@@ -308,4 +308,61 @@ describe("ObjectInspector - state", () => {
...
+    const store = wrapper.instance().getStore();
+
+    let nodes = wrapper.find(".node");
+    const root1 = nodes.at(0);
+    const root2 = nodes.at(1);
+
+    // Expanding a node calls recordTelemetryEvent.
+    root1.simulate("click");
+    expect(recordTelemetryEvent.mock.calls).toHaveLength(1);
+    expect(recordTelemetryEvent.mock.calls[0][0]).toEqual("object_expanded");
```
jasonLaster: `expect(recordTelemetryEvent.mock.calls).toMatchSnapshot()` is probably fine as you likely just want to prevent regressions.

Again, no big deal.

nchevobbe: yeah, I think I like this way better because it better conveys what we want to test exactly

jasonLaster: sounds good

nchevobbe:
Thanks for the review Jason. I'll merge for now, and we'll refactor if we add more probes to the objectInspector.

PR #6741, opened 2018-08-02
[breakpoints] Save breakpoints to asyncStorage

### LINKED ISSUE #6279: Fix devtools.debugger.pending-breakpoints overflow issue

jasonLaster:
Fixes #6279 

### Summary of Changes

- Switches from prefs to asyncStore.
- because asyncStore is _async_ we need to fetch the initial state before creating the store and pass it in.

### Test Plan

unit and integration tests should pass

### DIFF
--- a/src/utils/asyncStoreHelper.js
+++ b/src/utils/asyncStoreHelper.js
@@ -1,0 +1,55 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
+
+// @flow
+
+import { asyncStorage } from "devtools-modules";
+
+/*
+ * asyncStoreHelper wraps asyncStorage so that it is easy to define project
+ * specific properties. It is similar to PrefsHelper.
+ *
+ * e.g.
+ *   const asyncStore = asyncStoreHelper("r", {a: "_a"})
+ *   asyncStore.a         // => asyncStorage.getItem("r._a")
+ *   asyncStore.a = 2     // => asyncStorage.setItem("r._a", 2)
+ */
+export function asyncStoreHelper(root: string, mappings: Object) {
+  let store: any = {};
+
+  function getMappingKey(key) {
+    return Array.isArray(mappings[key]) ? mappings[key][0] : mappings[key];
+  }
+
+  function getMappingDefaultValue(key) {
+    return Array.isArray(mappings[key]) ? mappings[key][1] : null;
+  }
+
+  Object.keys(mappings).map(key =>
+    Object.defineProperty(store, key, {
+      async get() {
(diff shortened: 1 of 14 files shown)


src/utils/asyncStoreHelper.js
```diff
@@ -0,0 +1,32 @@
...
+      get() {
+        return asyncStorage.getItem(`${root}.${mappings[key]}`);
+      },
+      set(value) {
+        return asyncStorage.setItem(`${root}.${mappings[key]}`, value);
+      }
+    })
+  );
+
+  return (store: { [$Keys<typeof mappings>]: any });
```
jasonLaster: this is pretty cool...

if you try and do `asyncStore.foo` it will fail unless `foo` is in a key in the `mappings` object!
I tried this with `asyncStore.pendingBreakpoints2` and got a flow error

src/utils/dbg.js
```diff
@@ -61,6 +61,7 @@ export function setupHelper(obj: Object) {
     ...obj,
     selectors,
     prefs,
+    asyncStore,
```
jasonLaster: we can now do `dbg.asyncStore.pendingBreakpoints`

src/utils/asyncStoreHelper.js
AnshulMalik: awesome

src/utils/asyncStoreHelper.js
```diff
@@ -0,0 +1,34 @@
...
+// @flow
+
+import { asyncStorage } from "devtools-modules";
+
+/*
+ * asyncStoreHelper wraps asyncStorage so that it is easy to define project
+ * specific properties. It is similar to PrefsHelper.
+ *
+ * e.g.
+ *   const asyncStore = asyncStoreHelper("r", {a: "_a"})
```
AnshulMalik: shouldn't this be asyncStoreHelper("r", {a: ["_a", 1]})

jasonLaster: good point. We probably should support both types as well

darkwing:
I've just started testing this manually, but when I add a few breakpoints and refresh, the breakpoints go away and `dbg.asyncStore.pendingBreakpoints.then(bp => console.log(bp))` shows me an empty object.

jasonLaster:
Thanks for testing... I'm curious to see if you can pin point that

darkwing:
`dbg.selectors.getPendingBreakpoints()` is empty too. Still digging.

PR #6742, opened 2018-08-02
Add top-level await support

jasonLaster:
CC https://bugzilla.mozilla.org/show_bug.cgi?id=1410820

### Summary of Changes

- Adds client-side top-level await support. 
- returns the pending promise and logs the return value when it is complete
- relies on the fact that babel fails to parse top-level awaits, and attempts to wrap top level awaits in an async iife. We probably could do better than that :)

### Test Cases

 ✅  `await sleep(10)` => `10`
  ✅   `let x = await sleep(10); x + x` => `20`

### DIFF
--- a/assets/panel/prefs.js
+++ b/assets/panel/prefs.js
@@ -66,3 +66,9 @@ pref("devtools.debugger.features.async-stepping", true);
 pref("devtools.debugger.features.skip-pausing", true);
 pref("devtools.debugger.features.autocomplete-expressions", false);
 pref("devtools.debugger.features.map-expression-bindings", true);
+
+#if defined(RELEASE_OR_BETA)
+pref("devtools.debugger.features.map-await-expression", false);
+#else
+pref("devtools.debugger.features.map-await-expression", true);
+#endif
--- a/src/workers/parser/mapAwaitExpression.js
+++ b/src/workers/parser/mapAwaitExpression.js
@@ -1,0 +1,44 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
+
+// @flow
+
+import template from "@babel/template";
+import generate from "@babel/generator";
+import * as t from "@babel/types";
+
+import { parse, hasNode } from "./utils/ast";
+import { isTopLevel } from "./utils/helpers";
+
+function hasTopLevelAwait(expression: string) {
+  const ast = parse(expression, { allowAwaitOutsideFunction: true });
+  const hasAwait = hasNode(
+    ast,
+    (node, ancestors, b) => t.isAwaitExpression(node) && isTopLevel(ancestors)
+  );
+
+  return hasAwait && ast;
+}
+
+function wrapExpression(ast) {
+  const statements = ast.program.body;
+  const lastStatement = statements[statements.length - 1];
+  const body = statements
+    .slice(0, -1)
+    .concat(t.returnStatement(lastStatement.expression));
+
+  const newAst = t.arrowFunctionExpression([], t.blockStatement(body), true);
+  return generate(newAst).code;
+}
+
+export default function handleTopLevelAwait(expression: string) {
+  const ast = hasTopLevelAwait(expression);
+  if (ast) {
+    const func = wrapExpression(ast);
+    return generate(template.ast(`(${func})().then(r => console.log(r));`))
+      .code;
+  }
+
+  return expression;
+}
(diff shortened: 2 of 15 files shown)


src/workers/parser/mapExpression.js
```diff
@@ -7,6 +7,22 @@
...
 import mapExpressionBindings from "./mapBindings";
 
+import { hasSyntaxError } from "./validate";
+import { buildScopeList } from "./getScopes";
+import generate from "@babel/generator";
+import * as t from "@babel/types";
+
+function handleTopLevelAwait(expression) {
+  if (hasSyntaxError(expression) && expression.match(/await/)) {
+    const newExpression = `(async () => { return ${expression} })().then(r => console.log(r))`;
```
jasonLaster: This could be improved...

We likely would want to do the transform on the AST level

1. create an async function
2. add the first n-1 statements
3. add a return statement with the value of nth statement.

src/workers/parser/mapExpression.js
```diff
@@ -7,6 +7,22 @@
 import mapOriginalExpression from "./mapOriginalExpression";
 import mapExpressionBindings from "./mapBindings";
 
+import { hasSyntaxError } from "./validate";
+import { buildScopeList } from "./getScopes";
+import generate from "@babel/generator";
+import * as t from "@babel/types";
+
+function handleTopLevelAwait(expression) {
+  if (hasSyntaxError(expression) && expression.match(/await/)) {
```
jasonLaster: we should probably be smarter here :)

loganfsmyth: I'd re-parse with https://github.com/babel/babel/blob/master/packages/babel-parser/src/options.js#L32 enabled and then use an AST transform to do this wrapping.

jasonLaster: makes sense!

src/workers/parser/mapAwaitExpression.js
```diff
@@ -0,0 +1,43 @@
...
+  const newAst = t.arrowFunctionExpression([], t.blockStatement(body), true);
+
+  return generate(newAst).code;
+}
+
+export default function handleTopLevelAwait(expression: string) {
+  const ast = hasTopLevelAwait(expression);
+  if (ast) {
+    const func = wrapExpression(ast);
+    return `(${func})().then(r => console.log(r))`;
```
loganfsmyth: Should this also `console.error` on error?
Also, do we actually want to log the promise, and then log the result afterward, or should this include a `, void 0` at the end so that it actually evaluates to `undefined` at first or something?

src/workers/parser/mapAwaitExpression.js
```diff
@@ -0,0 +1,43 @@
...
+
+// @flow
+
+import { parse } from "./utils/ast";
+import generate from "@babel/generator";
+import * as t from "@babel/types";
+
+function hasTopLevelAwait(expression: string) {
+  const ast = parse(expression, { allowAwaitOutsideFunction: true });
+  const hasAwait = ast.program.body.some(
```
loganfsmyth: Seems like this should be a traversal, since the `await` can also be in nested expressions somewhere, as long as it's not in some other nested async function.

src/workers/parser/mapAwaitExpression.js
```diff
@@ -0,0 +1,43 @@
...
+function wrapExpression(ast) {
+  const statements = ast.program.body;
+  const lastStatement = statements[statements.length - 1];
+  const body = statements
+    .slice(0, -1)
+    .concat(t.returnStatement(lastStatement.expression));
+
+  const newAst = t.arrowFunctionExpression([], t.blockStatement(body), true);
+
+  return generate(newAst).code;
```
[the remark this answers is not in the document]
loganfsmyth: You can always use `@babel/template` too, e.g.
```
import template from "@babel/template";
// ...
return template.program.ast`
  (${newAst})().then(r => console.log(r));
`;
```

src/workers/parser/mapAwaitExpression.js
loganfsmyth: It's definitely a case of balancing between what devtools wants and what the engine actually provides :(

src/workers/parser/mapAwaitExpression.js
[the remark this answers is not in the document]
jasonLaster: didn't know that existed!

assets/panel/prefs.js
```diff
@@ -66,3 +66,4 @@ pref("devtools.debugger.features.async-stepping", true);
 pref("devtools.debugger.features.skip-pausing", true);
 pref("devtools.debugger.features.autocomplete-expressions", false);
 pref("devtools.debugger.features.map-expression-bindings", true);
+pref("devtools.debugger.features.map-await-expression", true);
```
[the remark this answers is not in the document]
jasonLaster: i suppose it could be, but then we wouldn't get feedback from nightly...

Issue #5570, opened 2018-03-01
[Breakpoints] improve breakpoint syncing cases


jasonLaster:
Julia, from QA, did a fantastic stress test of breakpoint syncing and found some great edge cases.

1. breakpoints saved in the module context (w/ multiple code changes, toggles)
2. breakpoints saved in the module context (w/ the line deleted)
   - i think this is the expected behavior

[overview]

[overview]:https://docs.google.com/document/d/1G9Dt2KpsO07u4RE3w23ws3j7At9TqVTm2xoTPLvr8js/edit#

PR #6743, opened 2018-08-03
[breakpoints] Fix syncing cases with toggling

jasonLaster:
Fixes Issue:  #5570 

### Summary of Changes

addresses the syncing issues, which would come about when enabling a breakpoint after the code has moved.

### Test Plan

followed case #1 in this [doc](https://docs.google.com/document/d/1VDSyP4N2WPlmbi1s3ZiJWm9_HBB4nOsngZaeX1lH81s/edit).

### DIFF
--- a/src/actions/breakpoints/addBreakpoint.js
+++ b/src/actions/breakpoints/addBreakpoint.js
@@ -20,7 +20,12 @@ async function addBreakpointPromise(getState, client, sourceMaps, breakpoint) {
   const state = getState();
   const source = getSource(state, breakpoint.location.sourceId);
 
-  const location = { ...breakpoint.location, sourceUrl: source.url };
+  const location = {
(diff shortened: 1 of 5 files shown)


darkwing:
I tested this and found the following:

missing

The breakpoint is there in the editor but doesn't display in the sidebar.

darkwing:
The breakpoint that was enabled, however, did move as expected!  Nice!

darkwing:
I'm a complete and total disaster this Monday morning...I did this on master.  Rechecking :(

darkwing:
I followed the STR and I'm still seeing the same result in panel; the disabled breakpoint doesn't display in the right column and the breakpoints didn't shift with the empty lines -- still at the original line.

jasonLaster:
hmm, lets do a call and look together

Issue #6353, opened 2018-05-19
[Editor] does not go to the correct location (when loading)


jasonLaster:
STR:

1. open the launchpad and make the viewport height narrow (200px)
2. go to Preview/index.js add a breakpoint on #72 (updatePreview)
3. hove in the debuggee editor and pause on 72
4. step in 

Expected Result: scrolled to line 3
Actual Result: scrolled to somewhere in the 40s :)

![](http://g.recordit.co/NmBnLnjsWs.gif)

jasonLaster:
@AlexisDeschamps @mmcote, i'm curious if you see anything obvious here. You've both looked at Editor scrolling [here][h] quite a bit.

[h]: https://github.com/devtools-html/debugger.html/blob/master/src/components/Editor/index.js#L400-L423

AlexisDeschamps:
I think the `selectLocation` function call is supposed to bring the editor to the right spot from what I have seen. Maybe the values sent to `selectLocation` are wrong. I can try taking a look eventually.

jasonLaster:
yep. my guess is it tries to scroll too soon. and then assumes we're in a good spot

jasonLaster:
Another case from @gregtatum 

This STR may not reproduce, I'm not sure, but this has happened to me a few times now so I'm filing it:

STR:
 * Add a debugger statement to some code in perf.html.
 * Trigger the code to catch on that statement.

Actual Result:
<img width="655" alt="image" src="https://user-images.githubusercontent.com/1588648/40332717-b58e47ae-5d1a-11e8-9776-8c3de4a479e6.png">

Expected Result:
<img width="655" alt="image" src="https://user-images.githubusercontent.com/1588648/40332787-01f0f7ea-5d1b-11e8-980d-36ca030115e6.png">

darkwing:
I haven't been able to duplicate this issue.  I do wonder if a good solution could be to, if only temporarily, set the cursor at the paused line; that would let CodeMirror handle the scroll position.

janodvarko:
Are there any update STR? I can't follow the steps in the first comment. The Preview/index.js file doesn't implement `onMouseOver` anymore.

Honza

darkwing:
I've been able to duplicate this but I've not yet figured out why it's happening.  It almost appears as though the editor is staying at the last position scrolled to in the file instead of moving to the breakpoint, despite the breakpoint line being highlighted.

PR #6748, opened 2018-08-03
[Editor] fix scrolling while loading

jasonLaster:
Fixes Issue: #6353

### Summary of Changes

The issue is determing what the scroll offset of the editor is, which is a bit tricky when the document is loading


jasonLaster:
new STR:

1. add a breakpoint at ProfileViewer.js 94
2. click on a tab in perf.html
3. pause and step in



screen shot 2018-08-06 at 4 12 31 pm

darkwing:
One interesting note is that the `scrollTop` is 0 at the time of calculation but when I inspect that element I see its scrollTop is a large number, so it appears to be a timing issue.

jasonLaster:
Nice find!

Issue #6746, opened 2018-08-03
[pause] currentScope.bindings is undefined


jasonLaster:
```
devtools/client/debugger/new/src/reducers/pause.js, line 403: 
TypeError: currentScope.bindings is undefined
```

PR #6749, opened 2018-08-03
[pause] accomodate with blocks

jasonLaster:
Fixes: #6746

### Summary of Changes

initial fix, but needs some help with flow

### DIFF
--- a/src/types.js
+++ b/src/types.js
@@ -319,10 +319,10 @@ export type ScopeBindings = {
  * @memberof types
  * @static
  */
-export type Scope = {
+export type Scope = {|
   actor: ActorId,
   parent: ?Scope,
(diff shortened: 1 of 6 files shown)


src/types.js
```diff
@@ -319,10 +319,10 @@ export type ScopeBindings = {
  * @memberof types
  * @static
  */
-export type Scope = {
+export type Scope = {|
   actor: ActorId,
   parent: ?Scope,
-  bindings: {
+  bindings?: {
```
jasonLaster: most of the changes are a result of making `bindigns` optional

jasonLaster:
Merging as this fixes a small type-error that was breaking the console in the browser toolbox when paused

Issue #6755, opened 2018-08-04
Package "url" is still vendored after #6678


juliandescottes:
I think we can remove "url" from vendors.js

PR #6756, opened 2018-08-04
Remove "url" from vendors.js

juliandescottes:
See #6755


juliandescottes:
Frees 30kB on vendors.js, not amazing but better than nothing!

jasonLaster:
hey, 30kb is good. babel 6 --> 7 this week was not cheep

Issue #6628, opened 2018-07-11
[accessibility] hide focus ring on clicks


jasonLaster:
@marcysutton mentioned in her fantastic talk [empathy driven development][edd] that it is possible to hide focus rings on mouse clicks with a small library [what-input]. Another option is to use [focus-visible][fs], which seems to line up better with some of the spec work.

In the future, i believe she mentioned that there will be a css selector that could help us here.

This will let us clean up some of our focus ring usage [here](https://github.com/devtools-html/debugger.html/search?q=%22outline%3A+none%22&unscoped_q=%22outline%3A+none%22).

[what-input]: https://www.npmjs.com/package/what-input
[edd]: https://youtu.be/l95VFLj3e2w?t=836
[fs]: https://github.com/WICG/focus-visible

jasonLaster:
CC @ahuth still interested?

ahuth:
Yes, @jasonLaster. Thanks for the reminder.

darkwing:
Thank you for taking this @ahuth!  Let us know if we can help!

PR #6757, opened 2018-08-05
Conditionally show focus indicators

ahuth:
Fixes Issue: #6628 

### Summary of Changes

* Introduces a React component which detects keyboard vs mouse interaction
* Applies a class indicating what mode of interaction is being used
* Styles focus outlines so that they only show up when the keyboard is being used
* Removes `outline: 0` from everywhere else, so that keyboard users can see what has focus

### Screenshots/Videos

Coming soon...

### DIFF
--- a/src/components/A11yIntention.css
+++ b/src/components/A11yIntention.css
@@ -1,0 +1,3 @@
+.A11y-mouse :focus {
+  outline: 0;
+}
--- a/src/components/App.js
+++ b/src/components/App.js
@@ -8,6 +8,7 @@ import React, { Component } from "react";
 import { connect } from "react-redux";
 import { features } from "../utils/prefs";
 import actions from "../actions";
+import A11yIntention from "./A11yIntention";
 import { ShortcutsModal } from "./ShortcutsModal";
 
 import {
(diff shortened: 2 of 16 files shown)


src/components/App.js
```diff
@@ -8,6 +8,7 @@ import React, { Component } from "react";
 import { connect } from "react-redux";
 import { features } from "../utils/prefs";
 import actions from "../actions";
+import A11yIntention from "./A11yIntention";
```
ahuth: Decided to introduce a component here, instead of bringing in [what-input](https://github.com/ten1seven/what-input) or a polyfill for `:focus-visible`.

Those solutions are really great, but our usecase here is much simpler than the problems they solve. For example, I don't believe we care about touch or pointer interaction.

src/components/App.js
```diff
@@ -297,14 +298,16 @@ class App extends Component<Props, State> {
...
+        <A11yIntention>
+          {this.renderLayout()}
+          {quickOpenEnabled === true && (
+            <QuickOpenModal
+              shortcutsModalEnabled={this.state.shortcutsModalEnabled}
+              toggleShortcutsModal={() => this.toggleShortcutsModal()}
+            />
+          )}
+          {this.renderShortcutsModal()}
+        </A11yIntention>
```
ahuth: This introduces another `<div>` around the app content. Figured that was okay.

If not, we can combine it with the parent `div.debugger`.

src/components/A11yIntention.css
```diff
@@ -0,0 +1,3 @@
+.A11y-mouse :focus {
+  outline: 0;
+}
```
jasonLaster: this is really cool! 

Can't wait to test it tomorrow

jasonLaster:
This is so great!

ahuth:
Thanks!

PR #6758, opened 2018-08-05
Auto-format package.json

ahuth:
When running `yarn install` the package.json gets auto-formatted for me, and this is the result.

I'm on yarn 1.9.4. Tried with Node 8.11.3 and 10.6.0.

Not sure if we actually want these changes or not, and if we don't that's cool.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,41 +18,31 @@
     "start": "node bin/dev-server",
     "start-app": "TARGET=application node bin/dev-server",
     "flow": "flow",
-    "eslint-check":
-      "eslint --print-config .eslintrc.js | eslint-config-prettier-check",
+    "eslint-check": "eslint --print-config .eslintrc.js | eslint-config-prettier-check",
     "prettier": "node bin/prettier.js",
     "license-check": "devtools-license-check",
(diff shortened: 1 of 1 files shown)


jasonLaster:
so, i think this is different than what prettier will format it. So the trick is to run prettier after `yarn install` or something like that

ahuth:
Interesting. Thanks for the info!

danwulff:
As a side note, this is why I traditionally add a `.prettierignore` file for `package.json`, `package-lock.json` (and yarn) files. (aka files that are generated by some CLI process that prettier wouldn't normally be touching, since devs wouldn't be touching it within their IDE)

jasonLaster:
Oh nice. That could work. I don't really care about the formatting. I just
don't like it bothering folks

ahuth:
Ah, makes sense. Will give it a shot, see how it works

ahuth:
I've learned some interesting things. As far as I can tell, this PR **_is_** how Prettier wants to format things.

Seems that f70b36bf5a9465d189d63237a0f64f703700f2ba the following:

```json
"install": "prettier --write package.json"
```

which causes prettier to run on `yarn install`. Not 100% sure, but we may actually want these changes.

PR #6760, opened 2018-08-05
Dev dependencies

ahuth:
Probably doesn't matter much, but the following packages were listed under regular dependencies, instead of dev dependencies:

- enzyme-adapter-react-16
- eslint-plugin-import

This PR moves them to the dev dependencies list, which is where I would expect them..


jasonLaster:
Yep! loving these little dev ux changes

ahuth:
Thanks!

PR #6765, opened 2018-08-06
Update local-development.md

AhmedSayed77:
use right property name

In creating a new feature flag, the feature was named "awesome", but in the example it was using the value "search" instead.


jasonLaster:
nice close reading!

jasonLaster:
thanks for the help

AhmedSayed77:
with pleasure

Issue #6711, opened 2018-07-27
[Tabs] feature to re-order editor tabs


kenjiO:
I thought it will be nice if people can re-order the tabs of the open files in the editor by dragging and dropping them.  I would like to try to implement this using the method used [for the toolbox tabs](https://searchfox.org/mozilla-central/source/devtools/client/framework/toolbox-tabs-order-manager.js).  Can you let me know your thoughts about this?

darkwing:
Wow, that would be an awesome feature @kenjiO !  I'd love to see something like that :)

kenjiO:
/claim
Cool.  I'll be able to get started on it next week.

jasonLaster:
Thanks, after thinking about it. I love the feature, but I'm not sure if sharing the code makes sense. Let's see what it looks like on tabs.js

Also there's a great ticket for refactoring the trans redux logic. This might be a good prerequisite.  CC @bomsy

bomsy:
The ticket @jasonLaster is probably referring to is #3279. it has some useful PRs that might need little to finish off.

kenjiO:
I'll wait until #3279 gets finished then.  What still needs to be done for that one?

jasonLaster:
@kenjiO it was close, but everyone got busy... 

https://github.com/devtools-html/debugger.html/pull/5672/files honestly is pretty old and we've come a long way since then. For instance we have much better flow types and no longer use Immutable for sources. If i were to start it now, I'd draw inspiration from it, but then start from scratch:

1) create a reducers/tabs.js file
2) separate tabs actions
3) wire it all up

darkwing:
I feel like a re-ordering feature may be a bit much but the idea of a "pinned" tab makes sense to me.  I can see a scenario where someone would want 2-3 given tabs always displaying first.

jasonLaster:
I feel like both of these features can be added to the icebox. Neither feels like an immediate goal for the team, but we would be happy to review PRs.

kenjiO:
OK.  I was putting this off because there were a lot of other PRs dealing with tabs that I didn't want to interefere with.  If you ever want to implement this feature let me know.

PR #5672, opened 2018-03-14
cleanup tabs, seperate reducers - rebased

darkwing:
Associated Issue: #3279

## Summary of Changes
- Adds tab reducer
- cleanup work

## Todo

- Fix lint issues
- Fix tests

This is the rebased work from https://github.com/devtools-html/debugger.html/pull/5313

PR #6768, opened 2018-08-06
[Tabs] Move actions and reducer for the editor tabs into their own files

### LINKED ISSUE #3279: [Tabs] Decouple tabs functionality from sources

kenjiO:
Fixes #3279.  This is to finish the work started in PR #5672 as discussed in #6711.

Tasks:
- [x] create a reducers/tabs.js file
- [x]  separate tabs actions
- [x]  wire it all up

### DIFF
--- a/src/reducers/tabs.js
+++ b/src/reducers/tabs.js
@@ -1,0 +1,172 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
+
+// @flow
+
+/**
+ * Tabs reducer
+ * @module reducers/tabs
+ */
+
+import { createSelector } from "reselect";
+import move from "lodash-move";
+
+import { prefs } from "../utils/prefs";
+import {
+  getSource,
+  getSources,
+  getSourceByURL,
+  getSourceByUrlInSources
+} from "./sources";
+
+import type { Action } from "../actions/types";
+import type { SourcesMap, SourcesState } from "./sources";
+
+type Tab = string;
+export type TabList = Tab[];
+export type TabsState = Tab[];
+
+export function initialTabsState(): TabsState {
+  return restoreTabs();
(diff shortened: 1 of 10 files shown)


kenjiO:
@jasonLaster  can you let me know how this looks so far?

jasonLaster:
looking good. We can land this once travis is happy

kenjiO:
How do you feel about moving the functions `removeSourceFromTabList`, `removeSourcesFromTabList` and `getNewSelectedSourceId` from `reducers/tab.js` to `actions/tab.js` or `utils/tab.js`?  Since they are not used by the reducer or selectors it makes more sense for me to not have them in actions where they are used or in utils.

src/reducers/tabs.js
```diff
@@ -0,0 +1,160 @@
...
+  getSourceByURL,
+  getSourceByUrlInSources
+} from "./sources";
+
+import type { Action } from "../actions/types";
+import type { SourcesMap, SourcesState } from "./sources";
+
+type Tab = string;
+export type TabList = Tab[];
+export type TabsState = Tab[];
```
kenjiO: I don't know whether or not to keep both `TabList` and `TabState` or just use one.

jasonLaster:
Either is fine

jasonLaster:
This is great! 

This will really help @AnshulMalik out with persisting better tab state. It will also help me with a similar tab related sourcemap bug!

kenjiO:
I changed it to just use the one type `TabList`.   Can you let me know if you want to move the `removeSourceFromTabList`, `removeSourcesFromTabList` and `getNewSelectedSourceId` functions or make any other changes?

jasonLaster:
Thank you! This is a huge step forward

jasonLaster:
Thanks for all the help over the past month. Feel free to join our slack so we can chat more regularly

https://devtools-html-slack.herokuapp.com/

PR #6773, opened 2018-08-07
[webreplay] tweak preview and continue to here

jasonLaster:
### Summary of Changes

* we should allow inline preview now for time travel 
* continue to here should be cmd click. I've been meaning to switch it and have always forgotten. This is really useful for time travel


darkwing:
Best way to test?

jasonLaster:
it's okay not to test as it is experimental. Continue to Here we will want to document in MDN though

PR #6784, opened 2018-08-08
[sources] add [sm] to source tree

jasonLaster:
Fixes Issue: #5663 

### Summary of Changes

This addresses the first of three issues with original sources that share a URL with a generated source. The other two issues are saved breakpoints and tabs. I think this is safe to land by itself and we can tackle the other two in their own PRs.

NOTE: the first commit is in another PR

screen shot 2018-08-08 at 5 57 06 pm

### DIFF
--- a/src/components/PrimaryPanes/SourcesTreeItem.js
+++ b/src/components/PrimaryPanes/SourcesTreeItem.js
@@ -4,19 +4,23 @@
 
 // @flow
 
+import { isOriginalId } from "devtools-source-map";
 import React, { Component } from "react";
+import { connect } from "react-redux";
 import classnames from "classnames";
 import { showMenu } from "devtools-contextmenu";
 
 import SourceIcon from "../shared/SourceIcon";
 import Svg from "../shared/Svg";
 
+import { getSourcesByURL } from "../../selectors";
+import actions from "../../actions";
+
 import { isDirectory } from "../../utils/sources-tree";
 import { copyToTheClipboard } from "../../utils/clipboard";
 import { features } from "../../utils/prefs";
 
 import type { TreeNode } from "../../utils/sources-tree/types";
-
 import type { Source } from "../../types";
 
 type Props = {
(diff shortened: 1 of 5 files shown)


src/components/PrimaryPanes/SourcesTreeItem.js
```diff
@@ -172,3 +178,23 @@ export default class SourceTreeItem extends Component<Props, State> {
...
+    return false;
+  }
+
+  const sources = getSourcesByURL(state, source.url);
+  return isOriginalId(source.id) && sources.length > 1;
+}
+
+const mapStateToProps = (state, props) => {
+  return {
+    isDup: hasMatchingGeneratedSource(state, props.source)
```
jasonLaster: what's a good name for this property?

src/components/PrimaryPanes/SourcesTreeItem.js
```diff
@@ -172,3 +178,23 @@ export default class SourceTreeItem extends Component<Props, State> {
...
+  };
+};
+
+export default connect(mapStateToProps, {
+  setProjectDirectoryRoot: actions.setProjectDirectoryRoot,
+  clearProjectDirectoryRoot: actions.clearProjectDirectoryRoot
+})(SourceTreeItem);
```
jasonLaster: it's now connected so it is easy to check if it has a dup

src/components/PrimaryPanes/SourcesTreeItem.js
darkwing: `hasMatchingGeneratedSource` seems a reasonable property name to me :)

darkwing:
We're still storing `prefs.debugger.tabs` as URLs (`["https://url1.com", ...]`) which is no longer viable if we're going to allow duplicate URLs, so I'm proposing we turn `prefs.debugger.tabs` into the following structure:

```js
[
    { url: "https://url1.com", generated: true }, 
    { url: "https://url1.com", generated: false }
]
```

This would allow duplicate tab URLs and maintain functionality of remembering tabs and connecting tabs to sources upon reload.  Thoughts?

jasonLaster:
this is green, sans lint issue

darkwing:
...and in turn, I think we'll need to use async storage (indexeddb) for tabs since someone could have many tabs open and we don't want to bloat prefs.

Issue #6772, opened 2018-08-07
[stepping] step over misses a location in mapExpression


jasonLaster:
STR: 

1. set a breakpoint in the browser toolbox worker/parser/mapExpression.js
2. add a console.log in the content toolbox
3. pause in mapExpression
4. step over a couple of times, observe some of the IF statements are missed

PR #6788, opened 2018-08-09
[PausePoint] pause at control flow

jasonLaster:
Fixes Issue: #6772

### Summary of Changes

We tried to be a bit too smart with pause points and only pause at conditions, the issue is that it is often nice to pause at the `if` or `else if`. We now always pause there


src/workers/parser/pausePoints.js
```diff
@@ -124,7 +124,7 @@ function onEnter(node: BabelNode, ancestors: SimplePath[], state) {
   }
 
   if (isControlFlow(node)) {
-    addPoint(state, startLocation, isForStatement(node));
+    addStopPoint(state, startLocation);
 
     const test = node.test || node.discriminant;
     if (test) {
```
jasonLaster: i'd like to comment why we have this test for iteration with `while` and for loops

PR #6795, opened 2018-08-10
Replace/expand stack frames with original source details

yurydelendik:
Fixes Issue: generated frames can correspond to multiple original stack frames

### Summary of Changes

* modify mapFrames() to replace single frame with multiple
* add stub for getOriginalStackFrames

### Test Plan

- See unit test in src/actions/pause/tests/pause.spec.js

### DIFF
--- a/src/actions/ast.js
+++ b/src/actions/ast.js
@@ -85,7 +85,7 @@ export function setOutOfScopeLocations() {
     const source = getSourceFromId(getState(), location.sourceId);
 
     let locations = null;
-    if (location.line && source && isPaused(getState())) {
+    if (location.line && source && !source.isWasm && isPaused(getState())) {
       locations = await findOutOfScopeLocations(source.id, location);
     }
 
--- a/src/actions/pause/mapFrames.js
+++ b/src/actions/pause/mapFrames.js
@@ -5,13 +5,19 @@
 // @flow
 
 import { getFrames, getSymbols, getSourceFromId } from "../../selectors";
+import assert from "../../utils/assert";
 import { findClosestFunction } from "../../utils/ast";
 
 import type { Frame } from "../../types";
 import type { State } from "../../reducers/types";
 import type { ThunkArgs } from "../types";
 
+import { isGeneratedId } from "devtools-source-map";
+
 export function updateFrameLocation(frame: Frame, sourceMaps: any) {
+  if (frame.isOriginal) {
+    return Promise.resolve(frame);
+  }
   return sourceMaps.getOriginalLocation(frame.location).then(loc => ({
     ...frame,
     location: loc,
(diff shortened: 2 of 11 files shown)


src/actions/pause/mapFrames.js
```diff
@@ -55,6 +64,63 @@ export function mapDisplayNames(
...
+      // Keep outer most frame with true actor ID, and generate uniquie
+      // one for the nested frames.
+      const id = j == 0 ? frame.id : `${frame.id}-originalFrame${j}`;
+      result.push({
+        id,
+        displayName: originalFrame.displayName,
+        location: originalFrame.location,
+        scope: frame.scope,
+        this: frame.this,
+        isOriginalSourceFrame: true,
```
jasonLaster: NIT:`isOriginal` is probably fine. We us `id` instead of `frameId` and `sourceId` and i think the same rule applies here

src/actions/pause/mapFrames.js
```diff
@@ -55,6 +64,63 @@ export function mapDisplayNames(
...
+    const originalFrames = await getOriginalStackFrames(
+      frame.generatedLocation
+    );
+    if (!originalFrames) {
+      result.push(frame);
+      continue;
+    }
+
+    assert(originalFrames.length > 0, "Expected at least one original frame");
+    originalFrames[0].originalLocation = frame.location;
```
jasonLaster: what is `originalLocation` used for? it looks like a new field

src/actions/pause/mapFrames.js
```diff
@@ -55,6 +64,63 @@ export function mapDisplayNames(
...
+function isWasmOriginalSourceFrame(frame, getState: () => State): boolean {
+  if (isGeneratedId(frame.location.sourceId)) {
+    return false;
+  }
+  const generatedSource = getSourceFromId(
+    getState(),
+    frame.generatedLocation.sourceId
+  );
+  return generatedSource.isWasm;
+}
```
jasonLaster: we might want to move this to `utils/frame.js` in the future if it has other use cases.

jasonLaster: NIT: what do you think of `isOriginalWasmFrame`? We use the term original to refer to other locations without having the source connotation. This might read a bit better...

jasonLaster:
One additional thought. It would be nice to have a mochitest for this as well.

src/actions/pause/mapFrames.js
yurydelendik: that's mistake (last minute refactoring) -- fixed and will add a comment

yurydelendik:
> One additional thought. It would be nice to have a mochitest for this as well.

I'll need to stub getOriginalSourceFrames somehow for mochitest. Do you have an idea how to test a new logic in unit tests?

jasonLaster:
> I'll need to stub getOriginalSourceFrames somehow for mochitest. Do you have an idea how to test a new logic in unit tests?

why would you need to stub it?

src/actions/ast.js
```diff
@@ -85,7 +85,7 @@ export function setOutOfScopeLocations() {
     const source = getSourceFromId(getState(), location.sourceId);
 
     let locations = null;
-    if (location.line && source && isPaused(getState())) {
+    if (location.line && source && !source.isWasm && isPaused(getState())) {
```
yurydelendik: Otherwise unit test fails

PR #6797, opened 2018-08-10
[sources] add support for only adding one bp per source

jasonLaster:
Fixes Issue: #5663

### Summary of Changes

- Adds support for only adding one bp.
- tested it here https://q7mmol9xy9.codesandbox.io/
- blocked on the `[sm]` pr because we need `getSourcesByUrl`

### Test Plan

- one jest test
- added a breakpoint in `index.js [sm]` on line 12 and refreshed. when it worked there was one bp, when it failed there were 2, the second was on line 11


darkwing:
Not sure how I did this but I went back/forward and saw this:

backforwardwuuuut

jasonLaster:
ooh not sure if that error is still there...

darkwing:
Yes!  Baller work dude!

PR #6800, opened 2018-08-10
[breakpoints] memoize Breakpoints component

jasonLaster:
### Summary of Changes

I wanted to file a bug for the debuggers pausing performance and when i gathered a profile I found a ridiculous number of Breakpoint.addBreakpoint calls, so instead I put together a prototype fix that memoizes all the things. To be fair, i don't think this is great code, but it is interesting and I'd like to continue exploring some of these things

screen shot 2018-08-10 at 4 55 17 pm

### DIFF
--- a/src/selectors/visibleBreakpoints.js
+++ b/src/selectors/visibleBreakpoints.js
@@ -7,8 +7,10 @@
 import { getBreakpoints } from "../reducers/breakpoints";
 import { getSelectedSource } from "../reducers/sources";
 import { isGeneratedId } from "devtools-source-map";
+import { createSelector } from "reselect";
 
-import type { State } from "../reducers/types";
+import type { BreakpointsMap } from "../reducers/types";
+import type { Source } from "../types";
 
 function getLocation(breakpoint, isGeneratedSource) {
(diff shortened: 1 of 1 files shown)


jasonLaster:
This should be ready, I ultimately decided to cache the redux side and leave the react component alone. It should be okay imo.

What's changed:
1. adding a breakpoint doesnt re-add all the other bps
2. stepping doesnt trigger the selector to re-compute all the bps

darkwing:
I took a quick look at this and I'm seeing an odd behavior.  I opened the following resource:

https://davidwalsh.name/wp-content/themes/punky/js/MooTools-Core-1.6.0.js

I put a breakpoint on all 3 lines and refreshed the debugger; the breakpoint on line 3 was gone.  Then I refreshed again and the breakpoint on line 3 was back but on line 2 was gone.

darkwing:
Could we move the `memoize` function to a more general util so it can be used elsewhere?

darkwing:
OK, I can't duplicate the issue in panel, so I think this is good to go.  We can move memoize the next time we use it.

PR #6805, opened 2018-08-11
Introduce utils/result-list spec

huoxito:
Related https://github.com/devtools-html/debugger.html/issues/5100 

Trying to get familiar with the code base here. Let me know if I missed something and this is already tested somewhere pls.

Two of the tests below don't make much sense because I'm not sure how to set `clientHeight` so we can assert properly on `scrollTop` for the chrome scenario. Should it provide a complete mock object to the test instead?

Also for the [delayed branch](https://github.com/devtools-html/debugger.html/blob/3a530479dd10ed05a0394e714153dcb30a8bce03/src/utils/result-list.js#L23) would assert on `setTimeout`being called enough?

cheers

### DIFF
--- a/src/utils/tests/result-list.spec.js
+++ b/src/utils/tests/result-list.spec.js
@@ -1,0 +1,40 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
+
+jest.mock("devtools-environment", () => ({
+  isFirefox: jest.fn()
+}));
+
+import { isFirefox } from "devtools-environment";
+import { scrollList } from "../result-list.js";
+
+describe("scrollList", () => {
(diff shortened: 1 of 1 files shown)


darkwing:
Welcome to the team @huoxito !  I've made a few adjustments, most notably because we don't support Chrome at the moment.

src/utils/result-list.js
```diff
@@ -13,11 +12,7 @@ export function scrollList(resultList, index, delayed = false) {
   const resultEl = resultList[index];
 
   const scroll = () => {
-    if (isFirefox()) {
-      resultEl.scrollIntoView({ block: "center", behavior: "smooth" });
-    } else {
-      chromeScrollList(resultEl, index);
-    }
+    resultEl.scrollIntoView({ block: "center", behavior: "smooth" });
```
jasonLaster: i think we should keep `chromeScrollList` because otherwise this will crash in when you use the launchpad in chrome.

darkwing:
Good shout, restored.

Issue #6821, opened 2018-08-14
Provide [sm] Style Enhancement


darkwing:
We recently added support for sourcemap files with the same URL as the minified file.  At the moment, we only show "[sm]" next to the file name in the sources tree, but it would look better with a bit more styling, a la:

<img width="162" alt="43419230-3a1ca270-93f5-11e8-8399-0600f9b27134" src="https://user-images.githubusercontent.com/46655/44096955-34191be8-9fa2-11e8-8457-e834d593e05f.png">

You can use https://q7mmol9xy9.codesandbox.io/ as a test side, as there are two sourcemapped files in that tree.

ekaoddlass:
/claim

Hi, I would like to try this! Is this still available?

PR #6823, opened 2018-08-14
Move suffix to separate element for easier styling

darkwing:
Helps Issue: #6821

Moving the suffix to its own element will help style it for #6821

### DIFF
--- a/src/components/PrimaryPanes/SourcesTreeItem.js
+++ b/src/components/PrimaryPanes/SourcesTreeItem.js
@@ -150,8 +150,7 @@ class SourceTreeItem extends Component<Props, State> {
   }
 
   renderItemName() {
-    const { item, hasMatchingGeneratedSource } = this.props;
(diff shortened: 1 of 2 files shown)


src/components/PrimaryPanes/SourcesTreeItem.js
```diff
@@ -150,21 +150,23 @@ class SourceTreeItem extends Component<Props, State> {
...
       default:
-        return `${item.name}${suffix}`;
+        return `${item.name}`;
     }
   }
 
   render() {
-    const { item, depth, focused } = this.props;
+    const { item, depth, focused, hasMatchingGeneratedSource } = this.props;
+    const suffix = hasMatchingGeneratedSource ? (
```
darkwing: We're going to need a follow up PR for this which creates/moves `getSuffixForSource()` to a source util so I can use it in other places, like tabs and breakpoint sidebar, but I'll do that with tabs work.

Issue #6774, opened 2018-08-08
Hyperactive debugger in background tab


digitarald:
STR:
- Open https://connorholyday.github.io/angular-pig-latin/ and Debugger
- Put in background tab and just browse the web

Debugger causes 120-140ms hangs on main thread every. Fan spinning 🔥. 

https://perfht.ml/2vMg3HM (this profile has 2 other occurrences, zoom out to see more)

jasonLaster:
yep, we'll have to clean up some of this perf. Good news is that some of this work is already underway

digitarald:
@jasonLaster which tickets would be relevant that we should mention here?

PR #6825, opened 2018-08-14
[sources] cache relative sources in redux store

jasonLaster:
Fixes Issue: #6774 

### Summary of Changes

This switches getRelativeSources from a memoized selector to a redux field. This is similar to how we now cached sources by url as well. The trade-off is memory vs computation. Ideally all the computed data would be cached in a selector, but that is difficult to do and maintain the same performance properties.

- [new version](https://perfht.ml/2vL57e7)
- [old version](https://perfht.ml/2vKElm9)

### old
image

### new
image


src/test/tests-setup.js
```diff
@@ -80,6 +80,7 @@ beforeEach(() => {
   clearSymbols();
   clearHistory();
   clearDocuments();
+  prefs.projectDirectoryRoot = "";
```
jasonLaster: the fact that we weren't clearing prefs prior is pretty scary...

src/reducers/sources.js
```diff
@@ -184,12 +191,49 @@ function updateSource(state: SourcesState, source: Object) {
...
+
+  const relativeSource = {
+    ...source,
+    relativeUrl: getRelativeUrl(source, root)
+  };
+
+  return {
+    ...state,
+    relativeSources: {
+      ...state.relativeSources,
```
AnshulMalik: We call `updateRelativeSource` for every source. And here it seems it's doing a lot of work. Your opinion?

jasonLaster: oh wow - good point. I didn't think of that!

jasonLaster: we should definitely have a mutable helper here

jasonLaster: just updated. what do you think?

AnshulMalik: Looking good :)

PR #6836, opened 2018-08-15
Don't repeat filename in subtitle for result list

darkwing:
I spotted this issue and it bugged me.  If the subtitle and title match, there's no point in showing a subtitle:

wuuuut

### DIFF
--- a/src/components/shared/ResultList.js
+++ b/src/components/shared/ResultList.js
@@ -61,9 +61,11 @@ export default class ResultList extends Component<Props> {
         <div id={`${item.id}-title`} className="title">
           {item.title}
         </div>
-        <div id={`${item.id}-subtitle`} className="subtitle">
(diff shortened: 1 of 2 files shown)


src/components/shared/ResultList.js
```diff
@@ -61,9 +61,11 @@ export default class ResultList extends Component<Props> {
         <div id={`${item.id}-title`} className="title">
           {item.title}
         </div>
-        <div id={`${item.id}-subtitle`} className="subtitle">
-          {item.subtitle}
-        </div>
+        {item.subtitle && item.subtitle != item.title ? (
```
jasonLaster: when will we not have a subtitle?

darkwing: ...good shout.  Will update.

Issue #6819, opened 2018-08-14
[scopes] map this in the debugger


jasonLaster:
It's currently possible to see  "This is optimized out" in the console when you're paused and this is not used.

STR: 

1. add a breakpoint in SourceTreeItem.js#renderItemName
2. reload a page

The easiest way to do this is in the browser toolbox, but perhaps the launchpad would work too

![screen shot 2018-08-07 at 5 30 24 pm](https://user-images.githubusercontent.com/254562/43803740-a58c0af0-9a67-11e8-8655-90fe466f46d0.png)
---
![screen shot 2018-08-07 at 5 30 21 pm](https://user-images.githubusercontent.com/254562/43803741-a59df152-9a67-11e8-86ee-8ffdc3d3ac1a.png)
---

This is fixed with this change:

```diff
diff --git a/src/utils/pause/mapScopes/index.js b/src/utils/pause/mapScopes/index.js
index 25dfdaa..08d0d11 100644
--- a/src/utils/pause/mapScopes/index.js
+++ b/src/utils/pause/mapScopes/index.js
@@ -484,7 +484,7 @@ async function findGeneratedBinding(
       },
       expression: null
     };
-  } else if (!hadApplicableBindings) {
+  } else if (!hadApplicableBindings && name !== "this") {
     // If there were no applicable bindings to consider while searching for
     // matching bindings, then the source map for this file didn't make any
     // attempt to map the binding, and that most likely means that the
```

the problem is that `this` is not used so there are no mappings

PR #6846, opened 2018-08-17
[parser] original language: Accommodate class properties

jasonLaster:
Fixes Issue: #6819 

### Summary of Changes

It looks like when you pause in a class that has a class-property that is mapped with babel, mapScopes gets confused. The reason I think it is confused is 

- the generated scopes includes the scope for the constructor which has a `this`. 
- the `this` does not have any applicable bindings and as a result we assume that it is optimized out... Perhaps this is a babel source map issue as well with that transform...

### Test Plan

- I started updating the classes example to have a class property. If this is right, we'll want to update the other builds to also have the transform.
- We should also update the console test to pause and evaluate `this` successfully


src/test/mochitest/examples/sourcemapped/output/parcel/classes.js
```diff
@@ -120,6 +120,10 @@ function root() {
   }
 
   class Another {
+    that() {
```
loganfsmyth: Where did this come from?

src/test/mochitest/examples/sourcemapped/fixtures/classes/input.js
```diff
@@ -7,10 +7,15 @@ export default function root() {
     }
   }
 
+
+
   class Another {
+    bound = () => {
```
loganfsmyth: I think the build script needs to be run again, since this isn't reflected in the output files.

src/test/mochitest/examples/doc-sourcemapped.html
```diff
@@ -13,120 +13,6 @@
       Content generated by examples/sourcemapped/build.js.
       Run "yarn build" to update.
     -->
-    <h2>parcel</h2>
```
loganfsmyth: Looks like you had these commented out when you ran the build?

PR #6847, opened 2018-08-17
Bug 1482401 - Fix preference name devtools.debugger.features.map-await-expression in test

Archaeopteryx:
### Summary of Changes

* change preference name in call to match the real preference's name which fixes test in beta simulations


jasonLaster:
Thanks!

Issue #6670, opened 2018-07-20
Watch expression "delete" button is shown on top of the Reps arrow icon


nchevobbe:
**Steps to reproduce**
1. Open the debugger
2. Add the following watch expression: `({a: 1})`

**Expected results**

I can click the arrow icon

**Actual results**

The remove expression icon is shown on top of the arrow icon:
<img width="231" alt="image" src="https://user-images.githubusercontent.com/578107/43006979-5ca7c2b8-8c37-11e8-8a96-e3e475cb4142.png">

jaikathuria:
@nchevobbe can you please help me reproduce the issue because I was able to get the **Expected results**
![2018-08-06 19 38 36](https://user-images.githubusercontent.com/13422807/43721481-83a1f560-99b0-11e8-86d4-9e9982388f7d.gif)

jasonLaster:
@jaikathuria interesting. I still see in the panel.
I wonder what is different...

![screen shot 2018-08-06 at 1 35 29 pm](https://user-images.githubusercontent.com/254562/43731785-9c452da4-997d-11e8-9fcc-4d7c66945c31.png)

jaikathuria:
I don't know if OS and Browser could make a difference here, I am on High Sierra 10.13.4 and my broswer is Firefox Version 61.0.1 (64-bit)

darkwing:
...weird.  So I'm seeing that in launchpad the following appears:

```css
html[dir="ltr"] .expression-container__close-btn {
    right: 0px;
}
```

But doesn't when I inspect a Firefox build.  Upon more inspection, the following *does* appear in our CSS:

```css
offset-inline-end: 0px;
```

So that's being translated in launchpad to `html[dir="ltr"] .expression-container__close-btn {right: 0px; }` but *not* inside panel -- it's just showing `offset-inline-end: 0px;` as in invalid style:

<img width="416" alt="offsetinlineendwtf" src="https://user-images.githubusercontent.com/46655/44275011-52a8ce00-a209-11e8-90c4-07866e49915a.png">

fvsch:
The direction-aware `offset-*` properties were renamed to `inset-*` in the CSS draft.
Firefox seems to have implemented that change in [bug 1464782](https://bugzilla.mozilla.org/show_bug.cgi?id=1464782), targetting Firefox 63.

fvsch:
See also:

[...]

PR #6848, opened 2018-08-17
Fix position of watch expression "delete" button

mbochynski:
Fixes Issue: #6670

### Summary of Changes

* Add renamed css properties `offset-*` to `inset-*` following https://bugzilla.mozilla.org/show_bug.cgi?id=1464782

### Test Plan

1.  Test at your localhost:8000 that the close button still displays properly
2.  Open Firefox, open Tools -> Web Developer -> Browser Toolbox, and use the inspector to manually set the same CSS change and confirm the [x] shows up in the right spot! :)

### Screenshots/Videos
delete-expression-button-placement

### DIFF
--- a/src/components/SecondaryPanes/Breakpoints/Breakpoints.css
+++ b/src/components/SecondaryPanes/Breakpoints/Breakpoints.css
@@ -188,7 +188,9 @@ html[dir="rtl"] .breakpoints-list .breakpoint .breakpoint-line {
 
 .breakpoint .close-btn {
   offset-inline-end: 15px;
+  inset-inline-end: 15px;
   offset-inline-start: auto;
+  inset-inline-start: auto;
   position: absolute;
   top: 8px;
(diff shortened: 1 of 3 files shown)


darkwing:
Wonderful @mbochynski !  Thank you so much!  Welcome to the team! :)

mbochynski:
@jasonLaster according to https://www.w3.org/TR/css-logical-1/#inset-properties this is the same as `left` and `right` properties, so in postcss-bidirection it would duplicate case of offset-inline-*. I will take a look and create a PR there.

jasonLaster:
> is the same as left and right properties

Duplicating makes sense

Wonderful! There's lots to do :)

gasolin:
@jasonLaster thanks for the heads up. 
@mbochynski issue is created in https://github.com/gasolin/postcss-bidirection/issues/18  feel free to pick it

Issue #6686, opened 2018-07-24
Horizontal scrollbar of preview popup is only visible at bottom


gerdneuman:
Using Windows 10 Nightly with Dark Theme.

This is from a real world problem I had, here's a minimal reproducable example.

**Steps to reproduce:**
1. Go to https://30926wj5o5.codesandbox.io/
2. Set breakpoint on line 31 (Interestingly, this also adds a breakpoint on line 32, but I think this is bug #6685) 
3. Reload
4. Hover over variable `myLongObject`

**Actual results:**
The horizontal scrollbar is not visible so it looks like the value is cropped. It was not obvious to me that the value goes on. The horizontal scrollbar is only visible when scrolling to the bottom.

![horizontal-scrollbar](https://user-images.githubusercontent.com/12525563/43132176-335dc946-8f3b-11e8-8e9e-83b50774eb8c.gif)

**Expected Results:**
The horizontal should always be visible.

jasonLaster:
Hey @darkwing is it possible we fixed this yesterday?

darkwing:
I doubt it; my guess is that the container doesn't allow vertical overflow but the tree inside it does.  Famous last words but this doesn't look too difficult to fix.

AnshulMalik:
`overflow: none` [here](https://github.com/devtools-html/debugger.html/blob/master/packages/devtools-components/src/tree.css#L6) seems to resolve this, can you please verify?

PR #6849, opened 2018-08-18
Fix preview popup horizontal scrollbar

luisbd:
Fixes Issue: #6686 

### Summary of Changes

* removed unneeded css block for `.tree` with the declaration `overflow: auto`

### Test Plan

1. At your localhost:8000 debugging, for example:`https://30926wj5o5.codesandbox.io/`.
2. Open debugger tool > Sources > `index.js`
2. Add a breakpoint at line 34
3. Reload `https://30926wj5o5.codesandbox.io/`
4. When the execution is halted at line 34 hover over `myLongObject` variable
3. Notice how after the fix the horizontal scrollbar is visible at any time independently from  the vertical scroll position.

#### Before fix
Before fix

#### After fix
After fix

### DIFF
--- a/packages/devtools-components/src/tree.css
+++ b/packages/devtools-components/src/tree.css
@@ -2,10 +2,6 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-.tree {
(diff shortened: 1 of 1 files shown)


darkwing:
Great!  Thank you so much @luisbd !  Welcome!

Issue #6820, opened 2018-08-14
[scopes] Improve Variable Mapping UI


jasonLaster:
Tools like CRA use line-based mappings. We should detect it and notify the user that they should change their source map mode to something else for better variables 

Looks like create react app is having some issues with the mappings:
https://github.com/devtools-html/debugger.html/issues/6762

---

This was caused by create-react-app using cheap-module-source-map, which has inline source maps. We can do 2 things:

1) detect in the sm-worker that it is 100% line mappings and not try
2) tell the user to change their source map strategy
3) see if CRA can choose a different mapping

jasonLaster:
Here is a [gist][g] of how to implement it. I'm currently stuck on the design and copy

[g]: https://gist.github.com/jasonLaster/67fb8be47dce83cb4f5e1426194cd9cc

jasonLaster:
we'll now track two things here:

1. the error message
2. improvements to how we show the mappings

janodvarko:
Not on our roadmap, closing.

Firefox Developer Tools (including Debugger tool) development is now entirely happening in Bugzilla and new issues should be reported here: https://bugzilla.mozilla.org/enter_bug.cgi?product=DevTools&component=Debugger

Honza

PR #6850, opened 2018-08-20
[scopes] detect line-based mappings

jasonLaster:
Issue: #6820 

### Summary of Changes

* checks for line mappings and bails early
* does not try to show a helpful error message

### DIFF
--- a/src/utils/pause/mapScopes/index.js
+++ b/src/utils/pause/mapScopes/index.js
@@ -65,17 +65,21 @@ export async function buildMappedScopes(
     return null;
   }
 
-  const generatedAstBindings = buildGeneratedBindingList(
-    scopes,
-    generatedAstScopes,
-    frame.this
-  );
-
   const originalRanges = await loadRangeMetadata(
     source,
     frame,
     originalAstScopes,
     sourceMaps
+  );
+
+  if (hasLineMappings(originalRanges)) {
+    return null;
+  }
+
+  const generatedAstBindings = buildGeneratedBindingList(
+    scopes,
+    generatedAstScopes,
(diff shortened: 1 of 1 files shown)


src/utils/pause/mapScopes/index.js
```diff
@@ -65,19 +65,23 @@ export async function buildMappedScopes(
...
-  );
-
   const originalRanges = await loadRangeMetadata(
     source,
     frame,
     originalAstScopes,
     sourceMaps
   );
 
+  if (originalRanges.every(range => range.columnEnd === Infinity)) {
```
darkwing: Just for my knowledge, why would a `columnEnd` be `Infinity`?

jasonLaster: that is what the source-map library includes

jasonLaster:
Hey @loganfsmyth could you take a look at this quickly?

src/utils/pause/mapScopes/index.js
loganfsmyth: We should check `columnStart === 0` too I think.

loganfsmyth: @darkwing Mappings are basically line/col => line/col, so you might have a mapping saying "original line 4 column 2 to generated line 6 column 4". If there aren't any other mappings on original line 4, you'll end up with a mapping range saying "line 4 column 2 to column Infinity" because it doesn't have any other information about where the mapping locally might end.

darkwing: Thank you for the explanation @loganfsmyth !

Issue #6829, opened 2018-08-14
Broken interactions & rendering in Source Map


digitarald:
STR:
- Open reddit.com and Debugger on a fresh profile
- Click through Webpack folder (all non-existent source files from reddit ads)

AR: Things go haywire at some point, with files duplicating and being unclickable.

![screenflow](https://user-images.githubusercontent.com/8599/44119982-b3004f3a-9fcf-11e8-8b60-9e93d0bf81ba.gif)

State after a plain refresh, `bootstrap` is already duped several times:

<img width="1280" alt="image" src="https://user-images.githubusercontent.com/8599/44120156-447096aa-9fd0-11e8-9f38-f0585c0d3d3c.png">

darkwing:
I'm fairly certain this is the sourcemaps stuff that @jasonLaster has just completed

darkwing:
Just tested and can confirm this is still an issue; the only difference is that the tree items all now show mapping:

![theydontloveyoulikeiloveyou](https://user-images.githubusercontent.com/46655/44127449-ee167620-a002-11e8-9d2d-5a4f4c278768.png)

darkwing:
With @jasonLaster 's mapped work, I can no longer get the duplicate item crap happen in panel.  Yessss.  :D

darkwing:
Grrr, scratch that, does still happen :/

jasonLaster:
yeah - it still happens for me too. Even after the performance fix. Not sure what is going on there though

PR #6856, opened 2018-08-21
[sources] improve updateSource performance

jasonLaster:
Issue: #6829 

### Summary of Changes

* speeds up updateSource by a lot. 

### old

https://perfht.ml/2OUOQKH

screen shot 2018-08-21 at 1 39 52 pm

### new

https://perfht.ml/2OVCMsz

screen shot 2018-08-21 at 1 32 48 pm

now the time spent is almost entirely in the UI which is the bigger blips next to the longer bars in the old perf graph

image

### DIFF
--- a/src/reducers/sources.js
+++ b/src/reducers/sources.js
@@ -68,19 +68,16 @@ function update(
   switch (action.type) {
     case "UPDATE_SOURCE": {
       const source = action.source;
-      return updateSource(state, source);
+      return updateSources(state, [source]);
     }
 
     case "ADD_SOURCE": {
       const source = action.source;
-      return updateSource(state, source);
+      return updateSources(state, [source]);
     }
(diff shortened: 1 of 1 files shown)


src/reducers/sources.js
```diff
@@ -187,19 +198,20 @@ function updateSource(state: SourcesState, source: Object) {
     ? { ...existingSource, ...source }
     : createSource(source);
 
+  state.sources[source.id] = updatedSource;
+
   const existingUrls = state.urls[source.url];
-  const urls = existingUrls ? [...existingUrls, source.id] : [source.id];
+  state.urls[source.url] = existingUrls
```
darkwing: Do we need to take into account origin vs. generated here?  Since a sourcemap can have the same URL?

Issue #6771, opened 2018-08-07
[Breakpoints] Use `loading` property


darkwing:
At present the Breakpoint object has a `loading` property but it doesn't appear we're actually using it anywhere; i.e. it default to false and never changes.

Ideally we default the value to `true` and then set to `false` when the server returns successful breakpoint syncing.

Once this is working properly, we can update `breakpointSources.js`'s `getBreakpointsForSource` function to check for `!bp.loading` to determine if the breakpoint should display.

jaikathuria:
Maybe, I have solved the first part of the issue. Now by the default value for loading will be `true` and then `false` when the `addBreakpointPromise` promise completes i.e. the server returns successful breakpoint syncing.

<img width="508" alt="screen shot 2018-08-11 at 12 29 56 am" src="https://user-images.githubusercontent.com/13422807/43976922-0cb13710-9d00-11e8-9d85-ad7bd4057b1e.png">


@darkwing please have a look at it and let me know if I am going right!

jasonLaster:
@jaikathuria  are you still interested in working on this?

Art-Vanderlay:
@jasonLaster I'll look into this one if @jaikathuria isn't working on this anymore. What else needs to be done here?

jasonLaster:
Go for it. Here's a PR with the latest attempt. 

https://github.com/devtools-html/debugger.html/pull/6857

PR #6799, opened 2018-08-10
Create Breakpoints with loading set

jaikathuria:
Fixes Issue: #6771 

### Summary of Changes

* This will make sure that when a Breakpoint is created it's `loading` property is set 
* When the `addBreakpointPromise` promise completes it will set the value of `loading` to `false`

### Test Plan

For now, I used Redux DevTools to test this, but moving forward I think I can write a test for this as well. 

### Screenshots

Screenshot for Redux DevTools depicting the change in `loading` property
screen shot 2018-08-11 at 12 29 56 am

PR #6857, opened 2018-08-21
Initialize a breakpoint as loading true

darkwing:
Fixes Issue: #6771

Supercedes https://github.com/devtools-html/debugger.html/pull/6799

Initializes a breakpoint as loading.

### DIFF
--- a/src/utils/breakpoint/index.js
+++ b/src/utils/breakpoint/index.js
@@ -133,7 +133,7 @@ export function createBreakpoint(
     condition: condition || null,
     disabled: disabled || false,
     hidden: hidden || false,
-    loading: false,
+    loading: true,
     astLocation: astLocation || defaultASTLocation,
     generatedLocation: generatedLocation || location,
     location,


darkwing:
@jaikathuria I moved this to a feature branch so we can further development on it -- let's work on getting the test working and get this in!  Great work so far!

darkwing:
I did some rough investigation here and the breakpoint's `loading` property is being saved as `true`; I can see that by executing the following in the console:

```js
dbg.store.getState().breakpoints.toJS().breakpoints
```

Within `src/actions/breakpoints/index.js`, the following is being hit when clicking the `[x]` button in the right side panel:

```js
removeBreakpoint(...) {
    ....
    if (!bp || bp.loading) {
        return;
    }
}
```

So upon reload, the close button doesn't work and there's no breakpoint arrow in the left sidebar.

Essentially we need to pinpoint that correct spot to turn `loading` back to false.

jaikathuria:
@darkwing you are right, with the small change made in the properties while creating the breakpoint now `loading` property is being saved as `true`.

Also, when the action to `addBreakpoint` is dispatched and `addBreakpointPromise` resolves the loading property is set back to `false` and I guess, this is the correct spot for this change to happen.

darkwing:
I'm going to close this for someone else to pick up ; lots of more important stuff to focus on right now.

Art-Vanderlay:
Hey @darkwing, I'd like to pick this one up. How can I help here?

darkwing:
I'd love for you to check this out @Art-Vanderlay !  If you can get this passing properly, I'd love it!

PR #6863, opened 2018-08-22
Bump prefs for new tabs format

darkwing:
Since the tabs pref went from an array of strings to an array of objects, we should bump prefs.  The change doesn't appear to be causing an error but this seems like the right thing to do.

### DIFF
--- a/src/utils/prefs.js
+++ b/src/utils/prefs.js
@@ -9,7 +9,7 @@ import { isDevelopment } from "devtools-environment";
 import Services from "devtools-services";
 import { asyncStoreHelper } from "./asyncStoreHelper";
 
-const prefsSchemaVersion = "1.0.3";
+const prefsSchemaVersion = "1.0.4";
 
 const pref = Services.pref;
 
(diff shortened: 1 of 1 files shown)


src/utils/prefs.js
```diff
@@ -9,7 +9,7 @@ import { isDevelopment } from "devtools-environment";
 import Services from "devtools-services";
 import { asyncStoreHelper } from "./asyncStoreHelper";
 
-const prefsSchemaVersion = "1.0.3";
+const prefsSchemaVersion = "1.0.4";
```
jasonLaster: we need to also clear tabs

```diff
 
@@ -120,5 +120,6 @@ export const asyncStore = asyncStoreHelper("debugger", {
 if (prefs.debuggerPrefsSchemaVersion !== prefsSchemaVersion) {
   // clear pending Breakpoints
   prefs.pendingBreakpoints = {};
+  prefs.tabs = [];
   prefs.debuggerPrefsSchemaVersion = prefsSchemaVersion;
}
```

jasonLaster:
how are you testing this?

Issue #6822, opened 2018-08-14
Fix [sm] tabs


darkwing:
Test site:  https://q7mmol9xy9.codesandbox.io/

The legendary @jasonLaster created functionality so that sourcemaps that share a URL with the minified URL can both display in the sources tree.  Unfortunately that breaks tabbing a bit; tabs don't display in the tab bar for the `[sm]` sources.

Instead of storing tabs as an array of URL strings, we should store them as:

```js
[{ url: "", isOriginal: bool }]
```

PR #6867, opened 2018-08-22
[tabs] Create tabs for sourcemapped files

darkwing:
Fixes Issue: #6822

This is a very early WIP for creating tabs based on URL *and* original/generated state.

## At the moment
Opening tabs from sources tree works *OK* but sometimes the tab refuses to switch to generated vs. original; it's kinda weird.  Close tab closes both tabs but that's because I've not implemented that functionality yet.

## ToDo
- [x] Tests
- [x] Close tab
- [x] Move tab

### DIFF
--- a/src/actions/ast.js
+++ b/src/actions/ast.js
@@ -39,7 +39,7 @@ export function setSourceMetaData(sourceId: SourceId) {
     }
 
     const framework = await getFramework(source.id);
-    dispatch(updateTab(source.url, framework));
+    dispatch(updateTab(source, framework));
 
     dispatch(
       ({
--- a/src/actions/tabs.js
+++ b/src/actions/tabs.js
@@ -65,13 +65,10 @@ export function moveTab(url: string, tabIndex: number): Action {
 export function closeTab(source: Source) {
   return ({ dispatch, getState, client }: ThunkArgs) => {
     const { id, url } = source;
+
     removeDocument(id);
 
-    const tabs = removeSourceFromTabList(
-      getSourceTabs(getState()),
-      url,
-      isOriginalId(id)
-    );
+    const tabs = removeSourceFromTabList(getSourceTabs(getState()), source);
     const sourceId = getNewSelectedSourceId(getState(), tabs);
     dispatch(({ type: "CLOSE_TAB", url, tabs }: Action));
     dispatch(selectSource(sourceId));
@@ -84,15 +81,18 @@ export function closeTab(source: Source) {
  */
 export function closeTabs(urls: string[]) {
   return ({ dispatch, getState, client }: ThunkArgs) => {
-    urls.forEach(url => {
-      const source = getSourceByURL(getState(), url);
-      if (source) {
-        removeDocument(source.id);
-      }
-    });
+    const sources = urls
+      .map(url => {
+        const source = getSourceByURL(getState(), url);
+        if (source) {
+          removeDocument(source.id);
+          return source;
+        }
+      })
+      .filter(source => source);
 
-    const tabs = removeSourcesFromTabList(getSourceTabs(getState()), urls);
-    dispatch(({ type: "CLOSE_TABS", urls, tabs }: Action));
+    const tabs = removeSourcesFromTabList(getSourceTabs(getState()), sources);
+    dispatch(({ type: "CLOSE_TABS", sources, tabs }: Action));
 
     const sourceId = getNewSelectedSourceId(getState(), tabs);
     dispatch(selectSource(sourceId));
--- a/src/actions/tabs.js
+++ b/src/actions/tabs.js
@@ -15,7 +15,7 @@ import { removeDocument } from "../utils/editor";
 import { selectSource } from "./sources";
 
 import {
-  getSourceByURL,
+  getSourcesByURLs,
   getSourceTabs,
   getNewSelectedSourceId,
   removeSourceFromTabList,
@@ -81,15 +81,8 @@ export function closeTab(source: Source) {
  */
 export function closeTabs(urls: string[]) {
   return ({ dispatch, getState, client }: ThunkArgs) => {
-    const sources = urls
-      .map(url => {
-        const source = getSourceByURL(getState(), url);
-        if (source) {
-          removeDocument(source.id);
-          return source;
-        }
-      })
-      .filter(source => source);
+    const sources = getSourcesByURLs(getState(), urls);
+    sources.map(source => removeDocument(source.id));
 
     const tabs = removeSourcesFromTabList(getSourceTabs(getState()), sources);
     dispatch(({ type: "CLOSE_TABS", sources, tabs }: Action));
--- a/src/actions/tabs.js
+++ b/src/actions/tabs.js
@@ -9,6 +9,8 @@
  * @module actions/tabs
  */
 
+import { isOriginalId } from "devtools-source-map";
+
 import { removeDocument } from "../utils/editor";
 import { selectSource } from "./sources";
 
@@ -21,20 +23,30 @@ import {
 } from "../selectors";
 
 import type { Action, ThunkArgs } from "./types";
+import type { Source } from "../types";
 
-export function updateTab(url: string, framework: string): Action {
+export function updateTab(source: Source, framework: string): Action {
+  const { url } = source;
+  const isOriginal = isOriginalId(source.id);
+
   return {
     type: "UPDATE_TAB",
     url,
-    framework
+    framework,
+    isOriginal
   };
 }
 
-export function addTab(url: string, framework?: string): Action {
+export function addTab(source: Source): Action {
+  const { url } = source;
+  const framework = "";
+  const isOriginal = isOriginalId(source.id);
+
   return {
     type: "ADD_TAB",
     url,
-    framework
+    framework,
+    isOriginal
   };
 }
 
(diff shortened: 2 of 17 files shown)


src/actions/tabs.js
```diff
@@ -30,11 +30,17 @@ export function updateTab(url: string, framework: string): Action {
...
+  isOriginal?: boolean
+): Action {
+  console.log("ADD_TAB", url, framework, isOriginal);
   return {
     type: "ADD_TAB",
     url,
-    framework
+    framework,
+    isOriginal
   };
```
jasonLaster: we might want to change this API to `addTab(source)`, `closeTab(source)` and do everything internally in the action

src/actions/ast.js
```diff
@@ -39,7 +39,7 @@ export function setSourceMetaData(sourceId: SourceId) {
     }
 
     const framework = await getFramework(source.id);
-    dispatch(updateTab(source.url, framework));
+    dispatch(updateTab(source, framework));
```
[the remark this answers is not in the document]
jasonLaster: i think we can move `const framework = await getFramework(source.id);` into `updateTab` so that whenever we call update it does this regardless of where we call it from

src/actions/types/SourceAction.js
```diff
@@ -61,6 +61,6 @@ export type SourceAction =
     |}
   | {|
       +type: "CLOSE_TABS",
-      +urls: string[],
+      +sources: Array<Source>,
```
darkwing: @jasonLaster Can you advise on why flow doesn't like this?  I tried `Source[]` and that blew up too.

src/actions/tabs.js
darkwing: Done!

src/actions/ast.js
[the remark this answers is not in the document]
darkwing: So I looked at this and I actually like where it is now.  This meta gets set once and it's when we know what that framework info is.  We don't want to be grabbing framework info every time that `updateTab` is called.

darkwing:
@jasonLaster Can you do a full review on this?  I'm hoping for advice on the flow issue and/or the test.  This is 99% ready to rock!

jasonLaster:
@darkwing sure, can you do a quick rebase first?

darkwing:
Rebased!

src/reducers/sources.js
```diff
@@ -312,7 +330,11 @@ export function getPrettySource(state: OuterState, id: string) {
     return;
   }
 
-  return getSourceByURL(state, getPrettySourceURL(source.url));
+  return getSourceByURL(
+    state,
+    getPrettySourceURL(source.url),
+    isOriginalId(source.id)
+  );
```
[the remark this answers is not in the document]
darkwing: The way this is being used at present, it appears, is to check if a pretty source for a given generated source exists, so I think it makes sense.

I'm very glad you asked this, however, because I think this *could be* what's causing the lone test to fail; I think for pretty-printed tabs the `isOriginalId` may not be correct which might cause a that failure.

src/reducers/tabs.js
```diff
@@ -46,23 +48,39 @@ function update(state: TabList = prefs.tabs || [], action: Action): TabList {
...
-export function removeSourceFromTabList(tabs: TabList, url: string): TabList {
-  return tabs.filter(tab => tab.url !== url);
+export function removeSourceFromTabList(
+  tabs: TabList,
+  source: Source
+): TabList {
+  return tabs.filter(
+    tab =>
+      tab.url !== source.url ||
+      (tab.url === source.url && tab.isOriginal != isOriginalId(source.id))
```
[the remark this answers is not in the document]
darkwing: We can't just do a URL comparison because we don't want to close both resources if they just want to close the original version with the same URL.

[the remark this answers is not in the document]
jasonLaster: agreed. my comment is that two url checks seems weird... haven't thought too much about it though

  tab.url !== source.url ||
      (tab.url === source.url && tab.isOriginal != isOriginalId(source.id))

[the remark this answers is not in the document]
jasonLaster: how is that different from 

> Remove the tab when the URLs don't match or their isOriginal values don't match

darkwing:
Did a quick audit of `SourcesTree`'s `componentWillReceiveProps` method and, as suspected, the following block wasn't triggered:

```js
if (
    nextProps.selectedSource &&
    nextProps.selectedSource != selectedSource
) {
    const highlightItems = getDirectories(
        nextProps.selectedSource,
        sourceTree
    );
    this.setState({ highlightItems });
}
```

That's what is causing the mochitest failure.  Looking into it now

darkwing:
@jasonLaster So my last commit is still correct; if we receive a prettyprinted tab, we do want to explicitly pass `false` because we need the source's URL to match what's in the tree, and a pretty URL will never be in the tree.  All set.

src/actions/tabs.js
```diff
@@ -50,11 +60,13 @@ export function moveTab(url: string, tabIndex: number): Action {
...
  * @static
  */
-export function closeTab(url: string) {
+export function closeTab(source: Source) {
   return ({ dispatch, getState, client }: ThunkArgs) => {
-    removeDocument(url);
+    const { id, url } = source;
 
-    const tabs = removeSourceFromTabList(getSourceTabs(getState()), url);
+    removeDocument(id);
```
jasonLaster: is it okay to switch from ID to URL? Would we need to change setDocument to also use ID?

jasonLaster: also, if we're switching from URL, maybe we want to actually do something smarter... 

```diff

-let sourceDocs: SourceDocuments = {};
+let sourceDocs: SourceDocuments = new WeakMap();

-export function getDocument(key: string) {
-  return sourceDocs[key];
+export function getDocument(source: Source) {
+  return sourceDocs.get(source);
 }

-export function hasDocument(key: string): boolean {
-  return !!getDocument(key);
+export function hasDocument(source: Source): boolean {
+  return sourceDocs.has(source, doc)
 }

-export function setDocument(key: string, doc: any) {
-  sourceDocs[key] = doc;
+export function setDocument(source: Source, doc: any) {
+  sourceDocs.set(source, doc)
 }

-export function removeDocument(key: string) {
-  delete sourceDocs[key];
-}
-
-export function clearDocuments() {
-  sourceDocs = {};
+export function removeDocument(source: Source) {
+  sourceDocs.remove(source);
 }
```

this means we won't ever need to call `clearDocuments`

src/actions/types/SourceAction.js
jasonLaster: `Source[]` should be the same... strange

src/actions/types/index.js
```diff
@@ -48,13 +48,15 @@ type ProjectTextSearchResult = {
 type AddTabAction = {|
   +type: "ADD_TAB",
   +url: string,
-  +framework?: string
+  +framework?: string,
```
jasonLaster: do we need this?

src/actions/tabs.js
darkwing: I like this idea but I think we should do this as a follow up; this PR is big enough already and `WeakMap` doesn't have a solution for clearing itself, so I don't think your `this means we won't ever need to call clearDocuments` statement is true.

darkwing: This is the only case I saw `removeDocument` used with a `url`; every other function uses a source ID

src/reducers/sources.js
```diff
@@ -286,12 +291,39 @@ export function getSourceFromId(state: OuterState, id: string): Source {
...
+  // Pretty sources should always be original
+  if (isPrettyURL(url)) {
+    isOriginal = true;
+  }
+
+  return getSourceByUrlInSources(
+    getSources(state),
+    getUrls(state),
+    url,
+    isOriginal
```
jasonLaster: I think it would be nice to pass in the correct isOriginal flag so that it includes formatted


```diff
diff --git a/src/actions/tabs.js b/src/actions/tabs.js
index 53060be..9b138e8 100644
--- a/src/actions/tabs.js
+++ b/src/actions/tabs.js
@@ -9,7 +9,7 @@
  * @module actions/tabs
  */

-import { isOriginalId } from "devtools-source-map";
+import { isOriginal } from "../utils/source";

 import { removeDocument } from "../utils/editor";
 import { selectSource } from "./sources";
@@ -27,7 +27,7 @@ import type { Source } from "../types";

 export function updateTab(source: Source, framework: string): Action {
   const { url } = source;
-  const isOriginal = isOriginalId(source.id);
+  const isOriginal = isOriginal(source.id);

   return {
     type: "UPDATE_TAB",
@@ -39,7 +39,7 @@ export function updateTab(source: Source, framework: string): Action {

 export function addTab(source: Source): Action {
   const { url } = source;
-  const isOriginal = isOriginalId(source.id);
+  const isOriginal = isOriginal(source.id);

   return {
     type: "ADD_TAB",
diff --git a/src/utils/source.js b/src/utils/source.js
index 1837d4c..defbca4 100644
--- a/src/utils/source.js
+++ b/src/utils/source.js
@@ -87,6 +87,10 @@ export function isJavaScript(source: Source): boolean {
   );
 }

+export function isOriginal(source: Source) {
+  return isOriginalId(source.id) || isPrettyURL(source.url)
+}
+
```

src/reducers/sources.js
```diff
@@ -322,20 +354,22 @@ export function hasPrettySource(state: OuterState, id: string) {
...
+  url: string,
+  isOriginal: boolean
 ) {
   const foundSources = getSourcesByUrlInSources(sources, urls, url);
   if (!foundSources) {
     return null;
   }
 
-  return foundSources[0];
+  return foundSources.find(source => isOriginalId(source.id) == isOriginal);
```
jasonLaster: we can switch this case to `isOriginal` as well, which will prevent `:formatted` edge cases

src/reducers/tabs.js
```diff
@@ -46,23 +52,37 @@ function update(state: TabList = prefs.tabs || [], action: Action): TabList {
...
 }
 
-export function removeSourceFromTabList(tabs: TabList, url: string): TabList {
-  return tabs.filter(tab => tab.url !== url);
+export function removeSourceFromTabList(
+  tabs: TabList,
+  source: Source
+): TabList {
+  return tabs.filter(
+    tab => tab.url !== source.url || tab.isOriginal != isOriginalId(source.id)
```
jasonLaster: same here

jasonLaster:
This is a great start, we can follow up on UI and isOriginal in follow ups

Issue #6868, opened 2018-08-22
[preview] preview can initially lag


jasonLaster:
It's unfortunately common for the first preview to lag by a second or two. The reason for this is that in order to show the preview, the debugger first has to compute the mapped expression. The mapping is performed by the parser worker, which in the case of original files is busy computing the original and generated ast scopes.

darkwing:
Yeah, that can take a while.  Might be nice to have a message somewhere just so the user knows work is happening and the info will be shown shortly.

jasonLaster:
We have a loading message in the scopes pane, but could do more there for sure.

PR #6869, opened 2018-08-22
Preview - speed up the previews

jasonLaster:
Fixes Issue: #6868

### Summary of Changes

- prevents mapping expressions when there arent mappings
- this fixes a slow case in the debugger when the debugger tries to map an expression while the parser worker is buy calculating original and generated scopes
- next steps - in the future it would be better to show a "loading" indicator when hovering and wait until we have the mappings to calculate the mapped expression. This is a significant amount of work that I think we can prioritize later.

### Testing

1. debug the debugger or another [app] w/ sourcemaps 
2. add a breakpoint in an original file
3. pause and quickly hover or use the console

[app]: https://firefox-debugger-example-react-js.glitch.me/

### DIFF
--- a/src/actions/expressions.js
+++ b/src/actions/expressions.js
@@ -167,7 +167,14 @@ export function getMappedExpression(expression: string) {
     const mappings = getSelectedScopeMappings(getState());
     const bindings = getSelectedFrameBindings(getState());
 
-    if (!mappings && !bindings && !expression.includes("await")) {
+    // We bail early if we do not need to map the expression. This is important
+    // because mapping an expression can be slow if the parser worker is
+    // busy doing other work.
+    //
+    // 1. there are no mappings - we do not need to map original expressions
+    // 2. does not contain `await` - we do not need to map top level awaits
+    // 3. does not contain `=` - we do not need to map assignments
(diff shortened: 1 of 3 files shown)


jasonLaster:
I think we might have the same issue with the console initially as well. I'm going to explore simplifying this tomorrow

jasonLaster:
the test failure is pretty cool 

browser_dbg-preview-source-maps

src/actions/expressions.js
```diff
@@ -167,7 +167,14 @@ export function getMappedExpression(expression: string) {
...
     const bindings = getSelectedFrameBindings(getState());
 
-    if (!mappings && !bindings && !expression.includes("await")) {
+    // We bail early if we do not need to map the expression. This is important
+    // because mapping an expression can be slow if the parser worker is
+    // busy doing other work.
+    //
+    // 1. there are no mappings - we do not need to map original expressions
+    // 2. does not contain `await` - we do not need to map top level awaits
+    // 3. does not contain `=` - we do not need to map assignments
```
darkwing: I love these types of comments in code! 👍

darkwing:
Snappy!  Nice!

PR #6870, opened 2018-08-22
Experiment with directly reading WASM DWARF format

yurydelendik:
NODE: The patch is just a plan for future patches.

This patch allows to fetch wasm file directly, parse its DWARF sections, convert to source maps (JSON). Only .debug_line custom section provide useful information, but that's enough for proof-of-work.

Content of the DWARF format documented at https://yurydelendik.github.io/webassembly-dwarf/ .

dwarf_to_json.wasm generated from https://github.com/yurydelendik/dwarf-to-json


darkwing:
@yurydelendik What is the next step for this?

yurydelendik:
TODOs:
- [ ] check if we can include dwarf-to-json as part of the debugger.html build
- [ ] and/or, package dwarf-to-json as external dependency
- [ ] determine how to fix local source maps references, currently wasm sourceURLPrefixes section needs to be present to re-map local paths, see [this section content example](https://github.com/yurydelendik/old-man-sandbox/blob/a2e0cea56684a7588c1b2445ddad187ef22caeac/rust-wasm-hey/misc/remappings.json)
- [ ] how debugger.html knows that it can use generated source as DWARF data source, currently wasm file adds the sourceMappingURL section that refers to itself

jasonLaster:
@yurydelendik i think we should try and land this so it is easier to discuss the new scopes data that you're handling.

The one issue i see in the Jest test is that TextDecoder needs a shim...  maybe this... https://www.npmjs.com/package/text-encoding

https://github.com/devtools-html/debugger.html/blob/master/src/test/tests-setup.js#L39-L48

src/client/firefox/commands.js
```diff
@@ -365,7 +364,7 @@ async function checkServerSupportsListWorkers() {
     return false;
   }
 
-  const deviceFront = await getDeviceFront(debuggerClient, root);
```
jasonLaster: we'll land the sync PR first

jasonLaster:
one failing jest test: `packages/devtools-utils/src/tests/network-request.js` which looks pretty easy to fix. Nice work getting this ready to land!

jasonLaster:
opened it here https://github.com/devtools-html/debugger.html/pull/6892

Issue #6348, opened 2018-05-18
[sources] tweak selectSpecificLocation


jasonLaster:
I noticed we call two actions unnecessarily.

1. closeActiveSearch
2. addTab 

NOTE: i made up hasTab, i have no idea whether that exists :)

```diff
diff --git a/src/actions/sources/select.js b/src/actions/sources/select.js
index 3c673fc..0ac5c17 100644
--- a/src/actions/sources/select.js
+++ b/src/actions/sources/select.js
@@ -162,13 +162,16 @@ export function selectSpecificLocation(location: Location) {
     }

     const activeSearch = getActiveSearch(getState());
-    if (activeSearch !== "file") {
+    if (activeSearch && activeSearch !== "file") {
       dispatch(closeActiveSearch());
     }

     const source = sourceRecord.toJS();

-    dispatch(addTab(source, 0));
+    if (!hasTab(getState(), source)) {
+      dispatch(addTab(source, 0));
+    }
+
     dispatch(setSelectedLocation(source, location));

     await dispatch(loadSourceText(sourceRecord));
     ```

jasonLaster:
@reescott still interested?

kenjiO:
@jasonLaster, Can I take this one?

jasonLaster:
sure @kenjiO

WenInCode:
@jasonLaster I was just browsing through the issues and I think this can be closed now since the pr was merged

PR #6878, opened 2018-08-23
[sources] don't displatch ADD_TAB and TOGGLE_ACTIVE_SEARCH when not needed

kenjiO:
Fixes Issue: #6348

### Summary of Changes
Made minor changes to selectLocation action
* add check if activeSearch is truthy before trying to close it
* add check if a source is already in the tabs state before adding it

### Test Plan

- [x] Open redux devtools and verify the ADD_TAB action is not being disptached when selecting a source with an existing tab
- [x] Open redux devtools and verify the TOGGLE_ACTIVE_SEARCH action is not being disptached when selecting a source while no search is active

### DIFF
--- a/src/actions/sources/select.js
+++ b/src/actions/sources/select.js
@@ -12,6 +12,7 @@
 import { isOriginalId } from "devtools-source-map";
 
 import { getSourceFromId } from "../../reducers/sources";
+import { getSourcesForTabs } from "../../reducers/tabs";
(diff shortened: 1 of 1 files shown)


kenjiO:
I noticed in the `selectLocation` funtion, `getState()` is called several times.  Would it make any noticable difference in performance to just call it once and assign it to a variable?

jasonLaster:
@kenjiO the reason we don't save getState() is that the state can change :)

Issue #6457, opened 2018-06-01
Paused state on refresh keeps the Debugger in half-loaded state


digitarald:
Firefox 62.0a1 (2018-06-01) (64-bit)

STR
- Opened perf-html.io
- Set a breakpoint in `webpack/src/index.js`
- Hard refresh, breakpoint gets hit
- Soft refresh a few times

ER: Page loads and breakpoint gets hit
AR: Half-loaded state for debugger, with correctly paused panel-tab icon but incomplete UI (no sources or breakpoints)

No errors in browser console.

<img width="500" alt="image" src="https://user-images.githubusercontent.com/8599/40860544-a0b2d970-659a-11e8-8e53-f42384d2202e.png">

darkwing:
I could reproduce this just as described.  I pressed "Resume" and nothing happened.  I then refreshed again and the breakpoint was hit again and the debugger continued to work properly.

darkwing:
@jimblandy Could this be a server bug?  Do you have any insights as to what could cause this?

jasonLaster:
Just tried this by refreshing fairly quickly and got an error here:

```js

function getPendingBreakpointsForSource(state, source) {
  // source is not defined
  const sources = (0, _sources.getSourcesByURL)(state, source.url);

  if (sources.length > 1 && (0, _devtoolsSourceMap.isGeneratedId)(source.id)) {
    // Don't return pending breakpoints for duplicated generated sources
    return [];
  }

  return getPendingBreakpointList(state).filter(pendingBreakpoint => pendingBreakpoint.location.sourceUrl === source.url);
}

```

digitarald:
Reproduced on recent Nightly with fresh profile, happened 2 times in this session:

![screenflow](https://user-images.githubusercontent.com/8599/47321301-4118e380-d609-11e8-9173-22cfd11d434f.gif)

Profile recorded while loading was stuck: https://perfht.ml/2S9wmsv

janodvarko:
I can't reproduce that anymore, closing.

We are moving our bug tracking into Bugzilla, so please report bugs here:
https://bugzilla.mozilla.org/enter_bug.cgi?product=DevTools&component=Debugger

Thanks,
Honza

Issue #6437, opened 2018-05-29
Error: Assertion failure: location must have a source url


darkwing:
STR:

1.  Go to https://coherent-potato.glitch.me/ ; (source contents are this:  https://gist.github.com/darkwing/bb7eb44bd7e655017f89e9436f297f99)

2.  Add a few breakpoints, then refresh the glitch page

3.  https://github.com/devtools-html/debugger.html/issues/6385 may occur; If so, just stop the glitch page and refresh the debugger page again.

4.  Refresh the glitch page.

See this error.

PR #6879, opened 2018-08-23
[sources] double check whether a source is present

jasonLaster:
Fixes Issue: #6457

### Summary of Changes

* fixed an issue i saw in the browser console here

### Test Plan

We should re-try this STR

### DIFF
--- a/src/actions/pause/mapFrames.js
+++ b/src/actions/pause/mapFrames.js
@@ -4,7 +4,7 @@
 
 // @flow
 
-import { getFrames, getSymbols, getSourceFromId } from "../../selectors";
+import { getFrames, getSymbols, getSource } from "../../selectors";
 import assert from "../../utils/assert";
 import { findClosestFunction } from "../../utils/ast";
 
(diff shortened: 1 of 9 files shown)


jasonLaster:
okay, i ran into two bugs:

```
TypeError: action.value is null[Learn More] sources.js:183:5 
```
```
TypeError: response.frames is undefined[Learn More] create.js:67:5 
```

src/utils/editor/source-documents.js
```diff
@@ -70,7 +70,7 @@ export function clearEditor(editor: SourceEditor) {
   const doc = editor.createDocument();
   editor.replaceDocument(doc);
   editor.setText("");
-  editor.setMode({ name: "text" });
+  // editor.setMode({ name: "text" });
```
darkwing: Do these calls need to be restored?

bomsy:
code looks good ..
testing now...

darkwing:
Nice!  I manually test the STR on master and debugger bombed, but on your branch I'm always brought to where I should go -- excellent!

We just need to fix the test.  Woot!

darkwing:
This may also fix; https://github.com/devtools-html/debugger.html/issues/6437

digitarald:
This did not seem to fix #6457

PR #6884, opened 2018-08-27
Roll back Tree component onFocus change.

nchevobbe:
The check that was made there prevented the first
item of the Tree to be selected if the click happened
inside it. Without this check, the first item is automatically
selected and scrolled to, which is causing some weirdness in
the console (probably in the debugger as well).
The fix was made in mozilla-central directly in the reps bundle
as well.


juliandescottes:
Ah, just read your explanation in Phabricator, looks like this is fine.

nchevobbe:
yeah, TLDR; : nchevobbe has no brain :D

Issue #5782, opened 2018-03-26
[Breakpoints] U+2028 and U+2029 characters throw off pause locations


karlcow:
Issues met when working with this issue https://webcompat.com/issues/16130

1. With Firefox Nightly 61.0a1 (2018-03-25) (64-bit)
2. Open devtools
3. Go to http://www.ximalaya.com/14647138/album/6527131/
4. Go to debugger
5. Find almond.js ( http://s1.xmcdn.com/js/almond.js?v=20180319144900 )
6. Unminify it (long script 109,510 lines once unminified)
7. Try to set a breakpoint by clicking on the left side of the line.

Actual: Nothing is happening

Expected: breakpoint is set.

jasonLaster:
hmm, that's interesting @AnshulMalik. can you share a bit more... Maybe some surrounding code?

jasonLaster:
wow great discovery, i wouldn't have thought of that. Can you please walk us through your process a bit more

AnshulMalik:
So, here is the process, how I discovered those chars:
- Download `almond.js` locally (`wget http://s1.xmcdn.com/js/almond.js?v=20180319144900`)
- Open it in editor(vscode in my case)
- Go to line 193, you'll see there are some characters which are binary and can not be displayed so there is a `?` kind of symbol.
- We can also see the ascii or hex value of character present at that location with `hexdump -s 6645 -n 100 -C almond.js`

And if we remove those `?` symbol chars, the parser works as expected.

jasonLaster:
CC @jimblandy

loganfsmyth:
There may be other cases I'm missing, but there is two primary issues here. One is easy to fix, but the other is harder.

Essentially, the issue is that JS has _5_ newline tokens (`\r\n`, `\r`, `\n`, `\u2028`, and `\u2028`), but at least a few places in the codebase assume that it is all `\n`.

* https://github.com/devtools-html/debugger.html/blob/ecc967e59b97ba95a25191abf308470ee6345d46/src/utils/source.js#L269
* https://github.com/devtools-html/debugger.html/blob/ecc967e59b97ba95a25191abf308470ee6345d46/src/utils/source.js#L388

While those two are easy to fix, there's one more that I came across. CodeMirror itself. CodeMirror splits all input code on `/\r?\n/` to split the user code into multiple lines. You can _kind_ of configure this in that we can:
[...]

PR #6906, opened 2018-08-30
Migrate to Monaco Editor

rebornix:
As title.


jasonLaster:
thanks @rebornix

src/components/Editor/Preview/index.js
```diff
@@ -33,90 +33,102 @@ type State = {
...
+  //   if (!inPopup(e)) {
+  //     this.props.clearPreview();
+  //   }
+  // }
+
+  onMouseOver = e => {
+    const { target } = e;
+    if (this.props.selectedFrameVisible) {
+      console.log("target");
+      this.props.updatePreview(target, this.props.editor);
```
jasonLaster: getting the preview popup to work well has been one of the biggest challenges. Can't wait to test that

src/components/Editor/Tabs.js
```diff
@@ -68,9 +82,25 @@ class Tabs extends PureComponent<Props, State> {
...
 
   componentDidUpdate(prevProps) {
-    if (!(prevProps === this.props)) {
-      this.updateHiddenTabs();
+    if (prevProps === this.props) {
+      return;
+    }
+
+    // @todo verify its correctness.
+    if (
```
jasonLaster: this has very little to do with codemirror - is it affected by monaco or are you trying to help us tackle a separate hard problem. This is interesting to me because the old approach has some style calculations which affect the dom ...

src/components/Editor/index.js
```diff
@@ -436,14 +427,17 @@ class Editor extends PureComponent<Props, State> {
...
 
       if (selectedSource && hasDocument(selectedSource.id)) {
         const doc = getDocument(selectedSource.id);
-        const lineText: ?string = doc.getLine(line);
+        const lineText = doc.getLineContent(line);
         column = Math.max(column, getIndentation(lineText));
       }
-      scrollToColumn(editor.codeMirror, line, column);
+
+      editor.monaco.revealPosition({ lineNumber: line, column: column });
```
jasonLaster: one of our concerns is we don't want to scroll every time the user moves, just when it is not visible. Does this do that?

src/components/Editor/index.js
```diff
@@ -558,11 +586,10 @@ class Editor extends PureComponent<Props, State> {
 
     return (
       <div>
-        <DebugLine />
-        <HighlightLine />
-        <EmptyLines editor={editor} />
```
jasonLaster: 👍

src/components/shared/Popover.js
```diff
@@ -136,43 +136,31 @@ class Popover extends Component<Props, State> {
...
-      };
-    } else {
-      targetMid = {
-        x: targetRect.left - popoverLeft + targetRect.width / 2 - 8,
-        y: 0
-      };
+      const top = targetRect.top + targetRect.height;
+      const targetMid = targetRect.width / 2 - 8;
+
+      return { left: targetRect.left, top, orientation: "down", targetMid };
```
jasonLaster: lol

src/components/shared/Popover.js
```diff
@@ -184,27 +172,22 @@ class Popover extends Component<Props, State> {
...
+    const top = targetRect.top - tooltipRect.height;
+    // const editorRect = editor.getBoundingClientRect();
+    // const targetRect = this.props.targetPosition;
+    // const left = this.calculateLeft(targetRect, editorRect, tooltipRect);
+    // const enoughRoomForTooltipAbove =
+    //   targetRect.top - editorRect.top > tooltipRect.height;
+    // const top = enoughRoomForTooltipAbove
+    //   ? targetRect.top - tooltipRect.height
+    //   : targetRect.bottom;
+    return { left: targetRect.left, top, orientation: "up", targetMid: { x: 0, y: 0 } };
```
jasonLaster: this is so much simpler. I can't wait to play with it

src/utils/monaco/source-editor.js
```diff
@@ -0,0 +1,76 @@
...
+
+// (1) Desired editor features:
+import "monaco-editor/esm/vs/editor/browser/controller/coreCommands.js";
+import "monaco-editor/esm/vs/editor/contrib/hover/hover.js";
+import "monaco-editor/esm/vs/editor/contrib/folding/folding.js";
+import "monaco-editor/esm/vs/editor/standalone/browser/quickOpen/gotoLine.js";
+import { ModelDecorationOptions } from "monaco-editor/esm/vs/editor/common/model/textModel";
+import * as monaco from "monaco-editor/esm/vs/editor/editor.api.js";
+
+import "monaco-editor/esm/vs/basic-languages/monaco.contribution";
```
jasonLaster: 👍

webpack.config.js
```diff
@@ -27,6 +27,7 @@ const webpackConfig = {
     // We always generate the debugger bundle, but we will only copy the CSS
     // artifact over to mozilla-central.
     debugger: getEntry("src/main.js"),
+    "editor.worker": "monaco-editor/esm/vs/editor/editor.worker.js",
```
jasonLaster: 👍

jasonLaster:
wow this works really well! I can't believe how good it is off the bat. Many things that we spent a lot of time on, seem to just work, which is huge.

I found a couple of rough edges:
- hovering on variables (hovering on the token, moving the mouse off the token, moving the mouse to the preview, moving off of preview, scrolling while in the preview)
- shortcuts like cmd+b/cmd+shift+b should toggle breakpoints (open a conditional panel). the panel does open if you right click on a breakpoint in the panel and select add condition. The panel works beautifully, horizontal scrolling is great
- right click should open context menus. both click events and keyboard events should be easy to fix.
- this is so minor, but it'd be nice to have a light theme while testing cause it is so distracting

darkwing:
@rebornix I managed to capture on camera my reaction when this came in:

awesome

Wow, this is amazing work!  I'll be looking it over and testing a few things, much as Jason has, over the next few hours.  Thank you so much!

src/components/Editor/index.js
```diff
@@ -104,80 +116,111 @@ type State = {
...
   }
 
   setupEditor() {
-    const editor = getEditor();
-
-    // disables the default search shortcuts
-    // $FlowIgnore
-    editor._initShortcuts = () => {};
+    const editor = new SourceEditor({
+      theme: "vs",
```
rebornix: @jasonLaster , any idea of how I can access the current theme option? It seems that themes are controlled by css but not options directly, or maybe I miss something.

src/components/Editor/Tabs.js
rebornix: Agreed, removed this out from this PR.

rebornix:
@jasonLaster it worked better before I merged the code with upstream, it has been 6 months since my last update, so many things changed ;) I made some changes real quick


> shortcuts like cmd+b/cmd+shift+b should toggle breakpoints (open a conditional panel). the panel does open if you right click on a breakpoint in the panel and select add condition. The panel works beautifully, horizontal scrolling is great

It now works ;)

> right click should open context menus. both click events and keyboard events should be easy to fix.

Right click can now invoke the proper context menu. The problem here right now is we can right click on the context menu, which will invoke system's context menu. I'd love to know how you fix this issue with CodeMirror.

> this is so minor, but it'd be nice to have a light theme while testing cause it is so distracting
 
I switched back to light theme for now. Once I know how to access the theme option we make it align.

> I haven't checked the profile yet. I wonder how it compares.

I like the performance gain of `setText` but I'm also surprised by Code Mirror's syntax highliting, which is lazy and fast. Apparently there are tons of things we can learn from Code Mirror.

marijnh:
Aww, it'd be sad to lose my favorite user, but so it goes. Could you enumerate the pain points that motivated this? Also see the plans for the future at https://codemirror.net/6/ , which should, once realized, put CodeMirror ahead in terms of architecture, speed, and accessibility again.

darkwing:
@rebornix I don't believe getting the theme is available at the moment, but I'll do more digging to find out how to make it available.

src/components/Editor/index.js
rebornix: The position is revealed only when necessary, which means it only scrolls when it's not visible. We can also use `revealPositionInCenter` if we want to reveal the position to the center of the viewport when necessary.

rebornix:
Cleaned up the code a bit more, and I tried to make the code change minimal by reusing the same API signatures. I have revisited popover/hover yet but will do later. With current code change, breakpoint/conditionPanel/contextMenu/search should work, probably not perfectly.

I also ran into one issue with the layout change. When I drag the sash between different panes, I didn't receive any event/updates in editor component, which means we don't have a chance to resize the editor properly. Anyone has ideas on this?

src/components/Editor/ConditionalPanel.js
```diff
@@ -67,21 +72,20 @@ export class ConditionalPanel extends PureComponent<Props> {
...
+    const { editor } = this.props;
+    if (this.expressionOverlay !== null) {
+      editor.monaco.removeOverlayWidget(this.expressionOverlay);
+      this.expressionOverlay = null;
     }
-  }
+    if (this.expressionViewZone >= 0) {
+      editor.monaco.changeViewZones(accessor => {
+        accessor.removeZone(this.expressionViewZone);
+        this.expressionViewZone = -1;
```
jasonLaster: what's going on here?

src/components/Editor/Preview/index.js
```diff
@@ -72,52 +57,125 @@ class Preview extends PureComponent<Props, State> {
...
+
+          // CONTENT_TEXT
+          if (e.target.type !== 6) {
+            this.onTokenLeave(e);
+            return;
+          }
+
+          const text = editor.monaco
+            .getModel()
+            .getWordAtPosition(e.target.position);
```
jasonLaster: this looks better than relying on the spans. i think we might want to use the cursor position if we can, but that might be fraught... This would help us differentiate in the case of hovering on `a`, `b`, or `c` e.g. `a.b.c.`

jasonLaster:
)

src/components/Editor/ConditionalPanel.js
rebornix: The way we draw the conditional panel in Monaco is 

* Render a whitespace area after Line A
* Put a DOM node, which is the conditional panel on top of the whitespace area (so in our case, the DOM node is managed by React)


When we hide the conditional panel, we need to remove the whitespace area from the editor. That's what the code is doing here.

darkwing:
Wanted to mention quickly; clicking the pretty print ("{}") button of a source breaks the debugger UI.  I tracked it down to:

```js
// /src/utils/monaco/index.js
export function shouldShowFooter(selectedSource, horizontal) {
  if (!horizontal) {
    return true;
  }
  if (!selectedSource) {
    return false;
  }
  return (
    shouldShowPrettyPrint(selectedSource) ||
    isOriginalId(selectedSource.get("id"))
  );
}
```

I think changing to `selectedSource.id` should fix the problem.

jasonLaster:
does monaco have an idea of documents so that we can load a source once and not have to tokenize/highlight it each time?

rebornix:
Some quick answers to @jasonLaster 's question:

> does monaco have an idea of documents so that we can load a source once and not have to tokenize/highlight it each time?

Monaco's text model will hold reference to both the content and tokens (if available), as long as the model is not disposed, the tokens will be not deleted. The model is created the first time we load the file https://github.com/rebornix/debugger.html/blob/monaco/src/utils/monaco/source-documents.js#L50-L57 and cached in https://github.com/rebornix/debugger.html/blob/monaco/src/utils/monaco/source-documents.js#L11 . If we don't remove the document when switching between different files (I don't think we are eagerly disposing them), then we are good, the document will only be tokenized once.

> I'm seeing some funny business with the flex layout, and the horizontal scrollbar.

As we discussed in community call two weeks ago, right now I didn't get any layout update event from editor components passed in props https://github.com/rebornix/debugger.html/blob/bb8611e795d02d28a53d07f25739a3c3b3ea3115/src/components/Editor/index.js#L86-L87 , if we can get proper size update event, we just need to call `editor.layout()` and Monaco will align its size with its container. @jasonLaster or @janodvarko probably have ideas how the hook up the events with the relayout ;)

>  this looks better than relying on the spans. i think we might want to use the cursor position if we can, but that might be fraught... This would help us differentiate in the case of hovering on a, b, or c e.g. a.b.c.

We discussed this in the community call as well and it is about what kind of tokens we want to fetch from the editor when hover. Instead of relying on tokenizers (which majorly focus on highlighting), in Monaco, we provide word around cursor by separating the line content by word separators https://github.com/Microsoft/vscode/blob/master/src/vs/monaco.d.ts#L2568 , by default its value is 

    `~!@#$%^&*()-=+[{]}\\|;:\'",.<>/?`

as you can see above, `.` is being seen as a word separator. So we can know for sure, which character we are hovering on in the case of `a.b.c`.

rebornix:
@darkwing @janodvarko thanks for pointing out the `selectedSource.get()` bug, it seems this syntax was removed in last half year but now I already removed all legacy code so the bug is gone.

src/components/Editor/Preview/index.js
```diff
@@ -72,52 +57,125 @@ class Preview extends PureComponent<Props, State> {
     this.updateHighlight(prevProps);
   }
 
+  componentWillUnmount() {
+    console.log("unmount");
```
JimLiu: could be removed :)

jasonLaster:
I'm closing because this is a bit dated. This was a fantastic experiment and will be a really useful point of comparison. 

Thanks!

jasonLaster:
Hey @rebornix do you think monaco would help with #5782 ?

Issue #6904, opened 2018-08-30
Source mapped label not readable


janodvarko:
The `(mapped)` label used for source mapped file in the Source tree is almost not visible when selected. See the attached screenshot.

The label text should use white color when selected, just like the file name text.

![image](https://user-images.githubusercontent.com/37785/44858051-69234080-ac71-11e8-8f3e-22de2880dfb4.png)

Honza

darkwing:
This should fix the problem:

```diff
diff --git a/src/components/PrimaryPanes/Sources.css b/src/components/PrimaryPanes/Sources.css
index 775de636b..d0c5cdf79 100644
--- a/src/components/PrimaryPanes/Sources.css
+++ b/src/components/PrimaryPanes/Sources.css
@@ -99,6 +99,10 @@
   color: var(--theme-comment);
 }

+.sources-list .tree .focused .label .suffix {
+  color: inherit;
+}
+
 .sources-list .tree img.arrow.expanded {
   transform: rotate(0deg);
 }
```

<img width="445" alt="yaymapped" src="https://user-images.githubusercontent.com/46655/44879052-30359c80-ac6e-11e8-959d-2d5ee44e9915.png">

PR #6908, opened 2018-08-30
Fix source mapped label

egdbear:
Fixes Issue: #6904


darkwing:
Ooops @egdbear, I think you want `.sources-list .tree .focused .label .suffix`; there's a missing `s` in the first part of the selector.

Issue #6700, opened 2018-07-25
[SourceTree] add integration tests for keyboard arrow behavior


jasonLaster:
We should test what happens when the user:

- **left arrow** on a closed directory
- **left arrow** on an open directory
- **left arrow** on a source
- **right arrow** on a closed directory
- **right arrow** on an open directory
- **right arrow** on a source
- **up arrow** on a directory
- **up arrow** on a source
- **down arrow** on a directory
- **down arrow** on a source
- **down arrow** at the bottom of the tree
- **up arrow** at the top of the tree

Garbee:
While someone is writing all these assertions here a few more that are helpful beyond just arrow keys:

* **home** takes you to the first node of the tree
* **end** takes you to the last node of the tree
* **enter** toggles the current selected node state if a parent item or trigger the nodes action if it is a leaf.

It also may be prudent to test things such as, right arrow on a leaf node does nothing. Try to cover as much as the [WAI-ARIA guidelines](https://www.w3.org/TR/wai-aria-practices-1.1/#TreeView) for a tree view as possible in testing. That way changes later don't do something unexpected without setting a bell off.

WenInCode:
Hey @jasonLaster, this looks like it could be a good first issue to take on. I have a good idea of how react components work in general, but I haven't written any integration tests for a react codebase before. This looks like it could be a great introduction into that. Mind if I take a go at this?

[This](https://github.com/devtools-html/debugger.html/blob/master/src/components/PrimaryPanes/SourcesTree.js) is the component we want to test, right?

jasonLaster:
Yep. You could write the test with jest. Or mochitests. We document both in
the docs dir. Mochitests in some ways is better, but harder to setup.

WenInCode:
@jasonLaster Hey again,

I spent some time reading up on mochitests, and I watched the [video](https://www.youtube.com/watch?v=5K9Sx5529JE&t=1373s&list=WL&index=2) you have on youtube about debugging & mochitests. I thought the video was great. I am running into an issue getting mochitests to run properly though.

I have successfully ran `./bin/prepare-mochitests-dev` and when I first ran the existing sources tests they passed:
```
// I also ran it in headless
yarn mochid browser_dbg-sources
```

[...]

PR #6912, opened 2018-08-31
[SourceTree] Keyboard arrow behaviour integration tests

WenInCode:
Fixes Issue: #6700 

### Summary of Changes
* Adds keyboard integrations tests for source-tree

### Test Plan
- [x] left arrow on a closed directory
- [x] left arrow on an open directory
- [x] left arrow on a source
- [x] right arrow on a closed directory
- [x] right arrow on an open directory
- [x] right arrow on a source
- [x] up arrow on a directory
- [x] up arrow on a source
- [x] down arrow on a directory
- [x] down arrow on a source
- [x] down arrow at the bottom of the tree
- [x] up arrow at the top of the tree

### A few questions
- I wasn't sure if it was okay to add a second `add_task` inside one mochitest file. It looks like it has been done before, but I am not sure if it's not recommended.
- I had some issues knowing _what to wait for_ after triggering a `pressKey`. I made a function that waits for a specific node to gain focus. This does work, but if that test is going to fail and the node won't gain focus, the test will just timeout not giving the best error message. Any idea what I should be waiting for instead?

### DIFF
--- a/src/test/mochitest/browser.ini
+++ b/src/test/mochitest/browser.ini
@@ -727,6 +727,7 @@ skip-if = os == "win" || (verify) # Bug 1434792
 [browser_dbg-sourcemaps-bogus.js]
 skip-if = os == 'linux' && !asan # bug 1447118
 [browser_dbg-sources.js]
+[browser_dbg-sources-arrow-keys.js]
 [browser_dbg-sources-named-eval.js]
 [browser_dbg-stepping.js]
 skip-if = debug || (verify && (os == 'win')) || (os == "win" && os_version == "6.1")
--- a/src/test/mochitest/browser_dbg-sources-arrow-keys.js
+++ b/src/test/mochitest/browser_dbg-sources-arrow-keys.js
@@ -1,0 +1,79 @@
+/* Any copyright is dedicated to the Public Domain.
+ * http://creativecommons.org/publicdomain/zero/1.0/ */
+
+ // Test keyboard arrow behaviour
+
+async function waitForNodeToGainFocus(dbg, index) {
+  await waitUntil(() => {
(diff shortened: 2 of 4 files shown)


WenInCode:
Haha I think that timeout is exactly what happened

darkwing:
Thank you so much for this @WenInCode !  I love it!  We probably want to put this set of tests in a new file; that way we can isolate the `add_task` calls and keep the test suite more focused.

I do see a lot of `TypeError: findElement(...) is null, can't access property "classList" of it`, so we probably need to make `return findElement(dbg, "sourceNode", index).classList` safer.

WenInCode:
I'm having trouble reproducing the failure that's on Travis locally. Do you have any recommendations?

When I run the test with `yarn mochih browser_dbg-sources-arrow-keys` it passes.
I tried copying the command that is run on Travis, `./node_modules/.bin/mochii --ci --mc ./firefox --headless devtools/client/debugger/new`. The tests pass, but a different unrelated test fails. 

I've only worked with CircleCi before, and there I would probably try SSHing in and running the test. I'm not sure if there is an equivalent in Travis, or perhaps a recommended debug flow in these kinds of situations.

darkwing:
A few points:
-  You'll want to back out the `module-manifest` and `yarn.lock` files.
-  90% of the time the tests pass locally, but every so often I get a timeout at `TEST-PASS 2 node is focused`

You may want to try adding this to the top of the test new test script file:

`requestLongerTimeout(5);

WenInCode:
Thanks for the help 😄 . I spent some time looking into why it was getting hung up, and it seemed like the test was not waiting properly following clicking the bottom source in the tree. I removed the changes to the `yarn.lock` and the `module-manifest`

Issue #6789, opened 2018-08-09
[Editor] highlights  selected line on reload


jasonLaster:
STR:

1. pause somewhere
2. refresh

ER: see the source
AR: see the source and see the old pause location highlighted

![](http://g.recordit.co/qYa0mYmi6m.gif)

darkwing:
This is happening because of `HighlightLine.js`'s `setHighlightLine` function; the conditionals in `setHighlightLine` and `shouldSetHighlightLine` aren't preventing this from happening on initial load.  A few notes:

## setHighlightLine

* `selectedLocation` does have a value: 
```js
column: 0
line: 4
sourceId: "server1.conn0.child1/source25"
url: "https://d2zah9y47r7bi2.cloudfront.net/releases/current/tracker.js"
<prototype>: {…}
```
* `selectedFrame` has a value:

```js
id: "server1.conn0.child1/frame164"
location: {…}

column: 21856
line: 1
sourceId: "server1.conn0.child1/source34"
```
* `selectedSource` is the source of the opened tab

Upon load, if there's no breakpoint, there shouldn't be a selectedFrame, should there?

kenjiO:
I think this is happening because [paused.js](https://github.com/devtools-html/debugger.html/blob/020dbd9af2bad4a5f683256c5d5606184ce4e256/src/actions/pause/paused.js#L75) calls [selectLocation](https://github.com/devtools-html/debugger.html/blob/020dbd9af2bad4a5f683256c5d5606184ce4e256/src/actions/sources/select.js#L95) which dispatches `SET_SELECT_LOCATION` with the paused location including the line number.  Then in the [reducer](https://github.com/devtools-html/debugger.html/blob/020dbd9af2bad4a5f683256c5d5606184ce4e256/src/reducers/sources.js#L88) it sets `prefs.pendingSelectedLocation` to the currently paused file and line number.  When the page is refreshed it retrieves `prefs.pendingSelectedLocation` and gets the file with line number which it highlights.

Is there a need to save the line number in `prefs.pendingSelectedLocation`?   A simple fix would be to not save the line and column numbers in `prefs.pendingSelectedLocation`.

PR #6916, opened 2018-08-31
[editor] add check in HighlightLine to fix unwanted highlighting

kenjiO:
Fixes Issue: #6789

I think the issue is happening because [paused.js](https://github.com/devtools-html/debugger.html/blob/020dbd9af2bad4a5f683256c5d5606184ce4e256/src/actions/pause/paused.js#L75) calls [selectLocation](https://github.com/devtools-html/debugger.html/blob/020dbd9af2bad4a5f683256c5d5606184ce4e256/src/actions/sources/select.js#L95) which dispatches `SET_SELECT_LOCATION` with the paused location including the line number.  Then in the [reducer](https://github.com/devtools-html/debugger.html/blob/020dbd9af2bad4a5f683256c5d5606184ce4e256/src/reducers/sources.js#L88) it sets `prefs.pendingSelectedLocation` to the currently paused file and line number.  When the page is refreshed it retrieves `prefs.pendingSelectedLocation` and gets the file with line number which it highlights.

Is there a need to save the line number in `prefs.pendingSelectedLocation`?   If so this simple change fixes it.  If you want to keep the line and column numbers in the prefs than maybe I could add a flag somewhere where it could indicate to ignore them during a paused action.

EDIT: I abandoned the strategy of not saving the line number per Jason's comments.  Now the strategy I'm trying is to save the `prefs.pendingSelectedLocation` with a flag indicating not to highlight the line.


jasonLaster:
So, it'd be nice to be able to scroll to the correct location, but not highlight it. I think we should try and improve the Highlight Line component

kenjiO:
@jasonLaster Are there any circumstances where you want to scroll to the correct location and highlight it?   If so do you think we should add some kind of test in the HighlightLine component to determine whether or not to highlight or not?

src/components/Editor/HighlightLine.js
```diff
@@ -46,60 +46,32 @@ function isDocumentReady(selectedSource, selectedLocation) {
 }
 
 export class HighlightLine extends Component<Props> {
-  isStepping: boolean = false;
```
kenjiO: I found a way to use `props.pauseCommand` in one location rather than having to set and track this value.

src/components/Editor/HighlightLine.js
```diff
@@ -46,60 +46,32 @@ function isDocumentReady(selectedSource, selectedLocation) {
...
+    return isDocumentReady(selectedSource, selectedLocation);
   }
 
   componentDidUpdate(prevProps: Props) {
-    const {
-      pauseCommand,
-      selectedLocation,
-      selectedFrame,
-      selectedSource
-    } = this.props;
```
kenjiO: I moved these to `setHighlightLine` since all this did was pass them on to that function anyway.

kenjiO:
I refactored the HighlightLine component because it was hard to for me to follow what it was doing. Hopefully this makes it clearer.  Let me know if you want to keep it and if you want me to do a separate PR for the refactor.

jasonLaster:
Wow. this is really interesting!

kenjiO:
Here is a way I found that fixes the problem.  I don't like having to add a `noHighlightLine` flag to the `pendingSelectedLocation` and `selectedLocation` objects but so far that is the only way I have come up with where it still saves the line to scroll to.

darkwing:
This looks good and appears to function as desired.  @jasonLaster Can you take a quick peek?  I know you're better versed with the use cases than I.  Also, can we think of a relevant test @kenjiO ?

darkwing:
@jasonLaster This looks good but I'd love a few more eyes.  I'll merge otherwise.

jasonLaster:
Will look today

kenjiO:
I've been working on a mochi test for this. I'll post a commit with it tomorrow.

jasonLaster:
Great, a mochitest will really help us here.

Why can't we do something like this? It feels like `previousEditorLine` should be sufficient. 

```diff
diff --git a/src/components/Editor/HighlightLine.js b/src/components/Editor/HighlightLine.js
index ff463bab..a899556f 100644
--- a/src/components/Editor/HighlightLine.js
+++ b/src/components/Editor/HighlightLine.js
@@ -95,11 +95,18 @@ export class HighlightLine extends Component<Props> {
     selectedSource: Source
   ) {
     const { sourceId, line } = selectedLocation;
+    const editorLine = toEditorLine(sourceId, line);
+
     if (!this.shouldSetHighlightLine(selectedLocation, selectedSource)) {
       return;
     }
+
+    if (!this.previousEditorLine) {
+      this.previousEditorLine = editorLine;
+      return;
+    }
+
     this.isStepping = false;
-    const editorLine = toEditorLine(sourceId, line);
     this.previousEditorLine = editorLine;

     if (!line || isDebugLine(selectedFrame, selectedLocation)) {
```

kenjiO:
@jasonLaster,  I like the way you came up with a lot better than adding a flag to the prefs.  This was the first time I wrote a mochi test.  Can you look over it to make sure it I did it right?  Thanks!

jasonLaster:
it looks good. You're a natural.

Does it fail without your fix?

kenjiO:
Yes, when I commented out the fix the mochi test fails.

jasonLaster:
Works great!

Issue #6913, opened 2018-08-31
Add feature to cycle through previous searches using up/down arrows


darkwing:
Most text editors save a history of searches and update the search input upon up and down keys, allowing you to access previous searches easily.  We should add that! 👍 

![arrows](https://user-images.githubusercontent.com/46655/44921488-4d1eae00-ad08-11e8-8baf-7140f4548788.gif)

darkwing:
In the short term I think we can simply keep the search terms in memory, no need for prefs storage.

samuelfullerthomas:
@darkwing this issue has been resolved

PR #6919, opened 2018-09-03
Add arrow-key-navigable search history

samuelfullerthomas:
Addresses Issue: #6913 

### Summary of Changes

* Add arrow-key-navigable search history to SearchBar and ProjectSearch components

### Test Plan

- [x] Tested in the global files search bar
- [x] Tested in the file search bar

### Screenshots/Videos
scroll

How do you guys feel about the code getting duplicated like this? I wasn't sure if it would be within the project style to throw it in a util folder somewhere or not.


darkwing:
Very nice start @samuelfullerthomas !  I love it!  A few improvements I think we can make:

[ ] - If I search for a term many times over, the array fills up with that same item.  I think we should try to shift the term or at least prevent the same term from being added multiple times.

[ ] - I'm wondering if there's a way we can cut down on the repeated code.  Maybe add the functionality to the shared SearchInput widget?

This is so great to see!  Keep up the awesome work!

src/components/shared/SearchInput.js
```diff
@@ -141,6 +146,47 @@ class SearchInput extends Component<Props, State> {
     }
   };
 
+  onKeyDown = (e: any) => {
+    const { onHistoryScroll, onKeyDown } = this.props;
+    if (onHistoryScroll) {
```
jasonLaster: maybe we should invert this: `if (!onHistoryScroll) {return}` or do `if (onHist) { return onHistoryScroll(e) }`

darkwing:
Awesome work so far @samuelfullerthomas !  A few things:

-  A jest test would be really nice and they aren't too difficult to write, so I think we'll need one
-  Hitting the "up" arrow immediately after submitting a search (`Enter` key) does nothing, you need to "up" arrow again to move

This is so cool so far :D

samuelfullerthomas:
good shout on the array index. I've updated and added a few tests @darkwing @jasonLaster.

darkwing:
Continued awesome work @samuelfullerthomas, I love it!

I'm thinking maybe we don't need a `historyPosition`, because we can get the position based on the value.

Also consider this case:  I search for  a term, then delete it, then press "Up"; nothing happens but I think we should show the last term.

Great work so far!  We're almost there!

darkwing:
Interesting concepts and thank you for the updates! 👍 

I searched for "<meta", pressed up, and saw "<met", and then "<me"; I was confused, thinking I was somehow hitting the backspace key.  I think we can nix the debouncing for that effect that's causing that autosave.

samuelfullerthomas:
Alright @darkwing @jasonLaster I think this is ready for review

PR #6937, opened 2018-09-06
[sources] add a webconsole sourcemap test

jasonLaster:
### Summary of Changes

- adds a failing webconsole test
- it fails because #6867 changed the definition of getSourceByURL to either look for original or generated sources

we might need to reintroduce an agnostic search. This could look like three separate selectors: `getSourceByUR`, `getOriginalSourceByURL`, `getGeneratedSourceByURL`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -64,4 +64,5 @@ script:
     browser_webconsole_eval_in_debugger_stackframe2.js
     browser_webconsole_location_debugger_link.js
     browser_webconsole_stacktrace_location_debugger_link.js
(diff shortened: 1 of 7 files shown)


jasonLaster:
I cannot land the isOriginal refactoring until this is ready... we should prioritize getting this fixed so these two commits can go to mc

darkwing:
Apologies for not knowing this was waiting on me; I for some reason thought this was a WIP.

Let's talk about this tomorrow morning; I mostly understand what's happening here but I'd like to talk more about a better solution.

Issue #6932, opened 2018-09-05
Reps: add a rep for Accessible actor.


yzen:
Accessibility panel sidebar will soon need to be able to accessible actors[fronts] for accessible object's relations. The rep for the Accessible actor is very similar (and simpler) to Element node.

nchevobbe:
@yzen how do I do to get an Accessible actor logged in the console ? is it something that will be accessible in the content page or just in chrome ?

yzen:
At this point it will only be used with accessibility inspector. In browser console you can query XPCOM accessible objects directly. None of it, AFAIK, is available in non privileged content.

PR #6939, opened 2018-09-06
Add accessible rep to reps package (#6932)

yzen:
Fixes Issue: #6932 

### Summary of Changes

* Added accessibility rep, similar but simpler than Element node rep.
* Added tests

### Test Plan

Added unit tests for the rep, also tried using it in m-c with a local copy.

### DIFF
--- a/packages/devtools-reps/src/reps/stubs/accessible.js
+++ b/packages/devtools-reps/src/reps/stubs/accessible.js
@@ -1,0 +1,67 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
+
+const stubs = new Map();
+stubs.set("Document", {
+  actor: "server1.conn1.child1/accessible31",
+  typeName: "accessible",
+  preview: {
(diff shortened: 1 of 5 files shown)


packages/devtools-reps/src/reps/stubs/accessible.js
```diff
@@ -0,0 +1,67 @@
...
+  actor: null,
+  typeName: "accessible",
+  preview: {
+    name: null,
+    role: "section",
+    isConnected: false
+  }
+});
+
+const name = "a".repeat(1000);
```
nchevobbe: can accessible have an actual longString name (i.e. if the name length is > 9999, in the majority of object we don't send the whole string, but only it's first N char - See https://github.com/devtools-html/debugger.html/blob/e7f212ee3916d97c799549b6b89cb43fbc1ce34f/packages/devtools-reps/src/reps/stubs/element-node.js#L224-L229)

yzen: It's quite possible. Platform a11y creates accessible objects for most text nodes. So if we have a text node that has more than 9999 chars, which is pretty reasonable, then it is the case.

yzen:
Just added a small change to parameterize separator text + a test case.

PR #6942, opened 2018-09-06
[reps] centralize reducer

jasonLaster:
### Summary of Changes

This is an experiment to see what it would look like if the Object Inspector reducer were instantiated alongside the rest of the parent application (debugger/console). There were some really nice benefits when i tried it:

- the actions receive client, hudProxy so we do not need to pass down `createObjectClient`, `releaseActor`
- actions were able to query the store, which simplified the components quite a bit. We could have likely done this anyways... fields like loadedProperties and actors can be queried 

Progress
- the debugger currently works. I'm sure I broke a lot of unit tests though :)
- i'd like to see how easy it would be to also support the console

Follow up
- I'm going to hold off working on persisting expand state as that is separate

### DIFF
--- a/packages/devtools-reps/src/object-inspector/actions.js
+++ b/packages/devtools-reps/src/object-inspector/actions.js
@@ -4,17 +4,10 @@
 
 // @flow
 
-import type {
-  CreateLongStringClient,
-  CreateObjectClient,
-  GripProperties,
-  LoadedProperties,
-  Node,
-  Props,
-  ReduxAction
-} from "./types";
+import type { GripProperties, Node, Props, ReduxAction } from "./types";
 
 const { loadItemProperties } = require("./utils/load-properties");
+const { getLoadedProperties, getActors } = require("./reducer");
 
 type Dispatch = ReduxAction => void;
 
@@ -27,30 +20,10 @@ type ThunkArg = {
  * This action is responsible for expanding a given node, which also means that
  * it will call the action responsible to fetch properties.
  */
-function nodeExpand(
-  node: Node,
-  actor?: string,
-  loadedProperties: LoadedProperties,
-  createObjectClient: CreateObjectClient,
-  createLongStringClient: CreateLongStringClient
-) {
-  return async ({ dispatch }: ThunkArg) => {
-    dispatch({
-      type: "NODE_EXPAND",
-      data: { node }
-    });
-
-    if (!loadedProperties.has(node.path)) {
-      dispatch(
-        nodeLoadProperties(
-          node,
-          actor,
-          loadedProperties,
-          createObjectClient,
-          createLongStringClient
-        )
-      );
-    }
+function nodeExpand(node: Node, actor) {
+  return async ({ dispatch, getState }: ThunkArg) => {
+    dispatch({ type: "NODE_EXPAND", data: { node } });
+    dispatch(nodeLoadProperties(node, actor));
   };
 }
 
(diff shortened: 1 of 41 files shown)


packages/devtools-reps/src/object-inspector/actions.js
```diff
@@ -72,22 +54,21 @@ function nodeFocus(node: Node) {
...
-function nodeLoadProperties(
-  item: Node,
-  actor?: string,
-  loadedProperties: LoadedProperties,
-  createObjectClient: CreateObjectClient,
-  createLongStringClient: CreateLongStringClient
-) {
-  return async ({ dispatch }: ThunkArg) => {
+function nodeLoadProperties(node: Node, actor) {
+  return async ({ dispatch, client, getState }: ThunkArg) => {
```
jasonLaster: We have access to `client` because of the debugger's thunkArgs. This was one of the best decisions we made early on in the debugger.  We can easily land this for the console as well.

https://github.com/devtools-html/debugger.html/blob/master/src/utils/bootstrap.js#L48-L50

packages/devtools-reps/src/object-inspector/actions.js
```diff
@@ -105,6 +86,15 @@ function nodePropertiesLoaded(
   };
 }
 
+function closeObjectInspector() {
+  return async ({ getState, hudProxy }: ThunkArg) => {
```
jasonLaster: we can pass `hudProxy` into thunk args as well, which should simplify releaseActor as well

packages/devtools-reps/src/object-inspector/component.js
```diff
@@ -126,16 +129,12 @@ class ObjectInspector extends Component<Props> {
...
-        )) ||
-      (expandedPaths.size === nextProps.expandedPaths.size &&
-        [...nextProps.expandedPaths].some(key => !expandedPaths.has(key))) ||
-      focusedItem !== nextProps.focusedItem
+
+    const hasLoaded = [...nextProps.expandedPaths].every(path =>
+      nextProps.loadedProperties.has(path)
     );
+
+    return hasLoaded || focusedItem !== nextProps.focusedItem;
```
jasonLaster: so this logic is probably too naive, but it seems to work :)

1. hasLoaded will be false if you expand and things are not loaded... so it might cover all of the cases... either way, it was me goofing around and should be separate.

packages/devtools-reps/src/object-inspector/component.js
```diff
@@ -187,33 +178,15 @@ class ObjectInspector extends Component<Props> {
...
-      const actor = isRoot || !value ? null : value.actor;
-      nodeExpand(
-        item,
-        actor,
-        loadedProperties,
-        createObjectClient,
-        createLongStringClient
-      );
-
+      const actor = getActor(item, roots);
```
jasonLaster: i feel like this could also be written differently to say something like, fetch every item that has an actor and isn't an environment... 

If we did this, then we would not need to pass actor in at all. And noadLoadProperties would look like this:

```js
function nodeLoadProperties(node: Node, actor) {
  return async ({ dispatch, client, getState }: ThunkArg) => {
    const loadedProperties = getLoadedProperties(getState());
    if (node.actor && !isEnvironment(node.actor) && loadedProperties.has(node.path)) {
      return;
    }
    ...
}
```

The advantage of this approach is 

1. we don't have to do the complicated roots check
2. we can express the logic in the action, which is easier to test and understand

nchevobbe:
So there's only 3 tests failing, and all are because of the use of `setProps` in the test (or the handling of `roots` changes, but my manual testing + mochitests in the console revealed no issues).

@jasonLaster do you have ideas on how to make those tests pass again ?

nchevobbe:
okay, Circle is happy now.

nchevobbe:
Here's a try push with these changes: https://treeherder.mozilla.org/#/jobs?repo=try&revision=bf6ca031259db1ade998b34f7da3e4042d0519f9

There's quite some failures.
2 are debugger tests, which I'm not sure why are failing.
the other ones are devtools inspector extension tests, and this is more concerning.

The code for inspector extension uses the ObjectInspector. So it means it is impacted by this change.
The issue is that it's not using its own store, but [re-use the inspector one](https://searchfox.org/mozilla-central/rev/ffe6eaf2f032e58ec3b0650a87df2c62ae4ca441/devtools/client/inspector/extensions/extension-sidebar.js#39). Now this means we need to modify the inspector store, which does not use thunk yet. I don't know how I feel about that.

jasonLaster:
we'll land this with the backport changes

Issue #6940, opened 2018-09-06
Rename "Search in Sources" to "Go to file"


digitarald:
Came to this when I looked at https://github.com/devtools-html/debugger.html/pull/6917

_"Search in Sources"_ can be misunderstood in how it searches; while its purpose is to only filter the file list to quickly jump to a specific file. Even harder to understand is how it compares to _"Find in Files"_.

Quick survey of other editors:

- Atom: _Find File_
- VSCode: _Go to File_
- WebStorm:  _Find class, file, or symbol_ , _Navigate to Class, File or Symbol by Name_

The three occurrences I am aware of are
- the welcome screen
- the search input field
- the shortcut cheat sheet, mixing both styles to _Source File Search_

My proposal is to align the copy to to _Go to file_.

_Edit: Switched case to be consistent with other copy._

PR #6944, opened 2018-09-06
Consistent use of find and files

digitarald:
Fixes Issue: #6940

### Summary of Changes

* Search in Sources → Go to File
* Consistent lowercase

### Test Plan

- [x] Check wording in welcome screen
- [x] Check wording in go to file input
- [x] Check wording in find in file input
- [x] Check wording in go to line input
- [x] Check wording in find function input

### Screenshot

image

image


darkwing:
I *believe* we need new keys when updating a l10n item; isn't that right @flodolo ?

darkwing:
@flodolo Good shout; I searched for all the changed terms and one wasn't use so it has been removed.

assets/panel/debugger.properties
```diff
@@ -261,6 +261,10 @@ sourceSearch.search.key2=CmdOrCtrl+F
 # the source search input bar
 sourceSearch.search.placeholder=Search in file…
```
flodolo: Is this string still used after this PR? If not, it should be removed

Issue #6953, opened 2018-09-08
File expand / collapse broken in ProjectSearch view


samuelfullerthomas:
Hey All:

Collapsing and expanding of files seems to be broken in search. I've got a PR to fix it though! Just thought I'd file an issue to follow procedure?

Steps to reproduce:
1. In the `ProjectSearch` view, search for a term that is in at least one file
2. In the results view of `ProjectSearch`, try to expand or collapse a files results

Gif:
![broken-collapsing](https://user-images.githubusercontent.com/10165959/45256638-4ee41380-b391-11e8-99c3-5fc1ccb96a6f.gif)

janodvarko:
I can confirm that this has been fixed.
Thanks for the patch!
Honza

PR #6954, opened 2018-09-08
Make search results files collapse / expand

samuelfullerthomas:
Fixes Issue: #6953 

### Summary of Changes
* Remove `setExpanded` from `ProjectSearch` `renderFile` method
* the `Tree` component handles expanding and collapsing on its own

### Test Plan
- [x] Search for a search term with at least 1 result
- [x] click the arrow to expand or collapse a result

### In action
collapse-expand-fixed

### DIFF
--- a/src/components/ProjectSearch.js
+++ b/src/components/ProjectSearch.js
@@ -224,7 +224,6 @@ export class ProjectSearch extends Component<Props, State> {
       <div
         className={classnames("file-result", { focused })}
         key={file.sourceId}
-        onClick={e => setExpanded(file, !expanded)}
       >
         <Svg name="arrow" className={classnames({ expanded })} />
         <img className="file" />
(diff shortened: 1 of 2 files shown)


AnshulMalik:
It's great that you found a neat solution. But after little testing, I got to know that this complexity of additional `setExpanded` calls are to support the `onEnterPress` functionality.
 We should expand/collapse the node if we press enter on node, so we need to manually call the `setExpanded` method of `ManagedTree` from `ProjectSearch`.

We should have this in tests.

samuelfullerthomas:
Thanks for the review @AnshulMalik  - you're right, I should leave that in the method. For some reason I thought it was extraneous, but it's clearly not.

darkwing:
Nice work @samuelfullerthomas !  Thank you!

Issue #6733, opened 2018-07-31
[Pretty Print] Better format duplicate filename text


darkwing:
We should incorporate logic that doesn't do the extended file information when they're the same file, only pretty-printed.

<img width="1057" alt="probs" src="https://user-images.githubusercontent.com/46655/43487660-e2897374-94dc-11e8-9025-a720a7299b82.png">

jainsneha23:
@darkwing Have you thought on the behavior is the path name is too large? Should we truncate as we do for the filenames?

jaikathuria:
I guess something can be done here for large paths but truncate them won’t be good idea. Assume the following case when paths are 

```
abc/project2/html/landing/
abc/project3/html/landing/
```

If we truncate both for say 15 length the main difference between two paths that we actually want to display to differentiate between the files would be missing.
Just my thoughts 🙄

darkwing:
I'm wondering if there's a logical way to identify the difference between the two paths and show that, i.e. project2 vs project 3 in @jaikathuria 's example.

jasonLaster:
hmm, if the preceding part is long  we could show `.../abc/project2/...`

@jainsneha23 is this easy to modify with your existing logic?

jainsneha23:
@jasonLaster @darkwing @jaikathuria 
We also have very long paths like in www.google.com, which made me feel like truncating.
@jaikathuria  your example also makes sense. 
So, we have 2 cases here.

<img width="1414" alt="screen shot 2018-08-14 at 4 36 03 pm" src="https://user-images.githubusercontent.com/8366397/44088287-3851dbc6-9fe0-11e8-80d1-667cc0a6ae3b.png">

jasonLaster:
i think we should drop query strings from paths. perhaps show them for titles where you want more info...

ryzokuken:
Please close this.

jainsneha23:
Seems like the original requirement of the issue is fixed and merged. What else is pending to be done here? @darkwing 
Else can we close this issue please?

darkwing:
Your work was amazing @jainsneha23!  Thank you so much!

I'm going to look into improving the logic this week, so let's keep it open for now.  Thank you!

PR #6955, opened 2018-09-08
Hiding the extended file information for pretty printed code tab

Narrator:
Fixes Issue: #6733 

### Summary of Changes

* Added condition to check if source is being pretty printed and if so, hide the extended file information.

### Test Plan

- [x] Clicking the pretty print icon opens the pretty printed source tab
- [x] The extended file information is not shown in the pretty printed source tab's title.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -3364,6 +3364,10 @@ devtools-config@^0.0.16:
   version "0.0.16"
   resolved "https://registry.yarnpkg.com/devtools-config/-/devtools-config-0.0.16.tgz#e7251788422f42a16aa1372b6adbebfcb7a74994"
 
+devtools-connection@0.0.12:
+  version "0.0.12"
+  resolved "https://registry.yarnpkg.com/devtools-connection/-/devtools-connection-0.0.12.tgz#a1bb6add55e4b317dd5122c25d3bb47bc896918c"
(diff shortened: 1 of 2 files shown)


darkwing:
I think this may be a bit too broad; I think my intention was to hide the info *if the other tab is open*.  Thoughts?

Narrator:
By other tab, do you mean the original minified counterpart of the pretty-printed tab? 

If that's the case, should the extended file information appear in the pretty-printed tab when someone closes the original source tab?

darkwing:
*should the extended file information appear in the pretty-printed tab when someone closes the original source tab?* . Yep!  That seems reasonable to me!

yarn.lock
```diff
@@ -3364,6 +3364,10 @@ devtools-config@^0.0.16:
   version "0.0.16"
   resolved "https://registry.yarnpkg.com/devtools-config/-/devtools-config-0.0.16.tgz#e7251788422f42a16aa1372b6adbebfcb7a74994"
 
+devtools-connection@0.0.12:
```
jasonLaster: we shouldnt change the yarn.lock

darkwing:
I'm going to close this for now as I think we misunderstand the idea and scope.

PR #6967, opened 2018-09-12
Remove inToolbox function in production.

nchevobbe:
When we're generating bundles for mozilla-central,
there's no reason we should keep the inToolbox
functions as the answer is always true.
We remove the function declaration and replace
the call to the function by true.
This fixes an issue we have while trying to start
the parser worker from the toolbox (window is undefined).


nchevobbe:
Didn't looked to much in a cleaner solution, but I don't really like how we redeclare everything for the `production` environment in the babel config

nchevobbe:
Also, I should have mentioned that it fixes the issue we were seeing with the parser worker in the toolbox, and does not affect the debugger in the launchpad nor in the toolbox.

PR #6970, opened 2018-09-13
feat: Set accelerator text for context menu items

### LINKED ISSUE #3673: ContextMenus should show keyboard shortcuts
The old debugger used to show keyboard shortcuts. It would be nice to have this behavior as well as it's a good education

### New Debugger
screen shot 2017-07-06 at 10 26 13 pm

### Old Debugger
screen shot 2017-07-06 at 10 19 47 pm

jbhoosreddy:
Fixes #3673 

Associated PR: https://github.com/devtools-html/devtools-core/pull/1087

### Summary of Changes

* Add context menu accelerator for existing shortcuts
* Also add a shortcut for watch expressions
* Update string for copy source to be more accurate

### Test Plan
- Manual testing

#### Editor
screen shot 2018-09-12 at 11 11 10 pm

#### Gutter
screen shot 2018-09-12 at 11 11 22 pm

### DIFF
--- a/assets/panel/debugger.properties
+++ b/assets/panel/debugger.properties
@@ -19,10 +19,11 @@ collapsePanes=Collapse panes
 copyToClipboard.label=Copy to clipboard
 copyToClipboard.accesskey=C
 
-# LOCALIZATION NOTE (copySource): This is the text that appears in the
+# LOCALIZATION NOTE (copySource2): This is the text that appears in the
 # context menu to copy the selected source of file open.
-copySource=Copy source text
-copySource.accesskey=y
+copySource=Copy
+copySource2=Copy source text
+copySource2.accesskey=y
 
 # LOCALIZATION NOTE (copySourceUri2): This is the text that appears in the
 # context menu to copy the source URI of file open.
--- a/assets/panel/debugger.properties
+++ b/assets/panel/debugger.properties
@@ -19,11 +19,11 @@ collapsePanes=Collapse panes
 copyToClipboard.label=Copy to clipboard
 copyToClipboard.accesskey=C
 
-# LOCALIZATION NOTE (copySource2): This is the text that appears in the
+# LOCALIZATION NOTE (copySource.label): This is the text that appears in the
 # context menu to copy the selected source of file open.
 copySource=Copy
-copySource2=Copy source text
-copySource2.accesskey=y
+copySource.label=Copy source text
+copySource.accesskey=y
 
 # LOCALIZATION NOTE (copySourceUri2): This is the text that appears in the
 # context menu to copy the source URI of file open.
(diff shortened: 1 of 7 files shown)


src/components/Editor/EditorMenu.js
```diff
@@ -170,7 +172,8 @@ function getMenuItems(
     id: "node-menu-add-watch-expression",
     label: watchExpressionLabel,
     accesskey: watchExpressionKey,
-    click: () => addExpression(editor.codeMirror.getSelection())
+    click: () => addExpression(editor.codeMirror.getSelection()),
+    accelerator: formatKeyShortcut(L10N.getStr("expressions.key"))
```
jbhoosreddy: actually, I did more research and it seems the actual Electron API which this seems to try to emulate, actually takes unformatted strings in the form of "CmdOrCtrl+Shift+E". And does that mapping internal to the API. Which might lead to a slight duplication of the code. I guess I can refactor to expose that from a common library

assets/panel/debugger.properties
```diff
@@ -19,10 +19,11 @@ collapsePanes=Collapse panes
 copyToClipboard.label=Copy to clipboard
 copyToClipboard.accesskey=C
 
-# LOCALIZATION NOTE (copySource): This is the text that appears in the
+# LOCALIZATION NOTE (copySource2): This is the text that appears in the
 # context menu to copy the selected source of file open.
 copySource=Copy
-copySource.accesskey=y
+copySource2=Copy source text
```
flodolo: One last nit: it would be great to use `.label` in this type of string (older strings don't have that, newer one have, like the one right above)

```
# LOCALIZATION NOTE (copySource2.label): This is the text that appears in the
# context menu to copy the selected source of file open.
copySource2.label=Copy source text 
copySource2.accesskey=y
```

jbhoosreddy:
I tried to add the watch expression shortcut from the old debugger, in this PR, but it's causing issues so I'll remove it for now and try again later.

jasonLaster:
For fun, i played with showing the shortcuts in the launchpad because it is a cool CSS hack. Anyways, it's pretty easy to have a proof of concept :P 

screen shot 2018-09-17 at 9 56 03 am
screen shot 2018-09-17 at 9 55 52 am

assets/panel/debugger.properties
```diff
@@ -19,9 +19,10 @@ collapsePanes=Collapse panes
 copyToClipboard.label=Copy to clipboard
 copyToClipboard.accesskey=C
 
-# LOCALIZATION NOTE (copySource): This is the text that appears in the
+# LOCALIZATION NOTE (copySource.label): This is the text that appears in the
 # context menu to copy the selected source of file open.
 copySource=Copy
```
flodolo: Just noticed while translating: is the copySource string still used anywhere after this PR? If not, it should be removed

jasonLaster: good call. I can check

jasonLaster: actually we're close to removing the old debugger. Lets wait till then and then remove a lot of strings

flodolo: Sounds good, thanks (there's another PR with a lot more strings)

PR #6975, opened 2018-09-13
Async storage tabs

### LINKED ISSUE #6546: [perf] Storing too much data into devtools.debugger.tabs pref
I am seeing the following warning when stepping:

Warning: attempting to write 16114 bytes to preference devtools.debugger.tabs. This is bad for general performance and memory usage. Such an amount of data should rather be written to an external file. This preference will not be sent to any content processes.

Fixing this should have positive impact on stepping performance.

Here is example value of the pref: https://pastebin.com/3ABGfm9X

Honza

bomsy:
Fixes #6546 

### Summary of Changes

* Switch to use async storage

### DIFF
--- a/src/actions/sources/tests/select.spec.js
+++ b/src/actions/sources/tests/select.spec.js
@@ -89,17 +89,22 @@ describe("sources", () => {
   it("should select next tab on tab closed if no previous tab", async () => {
     const { dispatch, getState } = createStore(sourceThreadClient);
 
-    const fooSource = makeSource("foo.js");
-    await dispatch(actions.newSource(fooSource));
+    const bazSource = makeSource("baz.js");
+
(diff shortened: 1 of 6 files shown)


src/actions/sources/tests/select.spec.js
```diff
@@ -89,16 +89,20 @@ describe("sources", () => {
   it("should select next tab on tab closed if no previous tab", async () => {
     const { dispatch, getState } = createStore(sourceThreadClient);
 
-    const fooSource = makeSource("foo.js");
+    const bazSource = makeSource("baz.js");
+
     await dispatch(actions.newSource(fooSource));
```
bomsy: argh.. this should be `await dispatch(actions.newSource(makeSource("foo.js")));`

jasonLaster:
Thanks @AnshulMalik and @bomsy!

Issue #6947, opened 2018-09-07
[Pretty Print] Pretty print feature suggestions from Bug 1488654


juliandescottes:
From https://bugzilla.mozilla.org/show_bug.cgi?id=1488654
Forwarding from bugzilla, feel free to split more.

- [ ] add feature to always pretty print
- [ ] add feature to hide original file if pretty printed version exists
 
cc @birtles

loganfsmyth:
Closing in favor of Bugzilla

PR #6977, opened 2018-09-13
Show some progress when pretty printing

AnshulMalik:
### Summary of Changes

* Display loader when pretty print is in progress.
* Close the original file when pretty printed.

pretty-loading

### DIFF
--- a/src/actions/sources/prettyPrint.js
+++ b/src/actions/sources/prettyPrint.js
@@ -24,6 +24,7 @@ import {
 } from "../../selectors";
 
 import type { Action, ThunkArgs } from "../types";
+import { selectSource } from "./select";
 import type { JsSource } from "../../types";
 
 export function createPrettySource(sourceId: string) {
@@ -44,6 +45,7 @@ export function createPrettySource(sourceId: string) {
     };
 
     dispatch(({ type: "ADD_SOURCE", source: prettySource }: Action));
+    dispatch(selectSource(prettySource.id));
 
     const { code, mappings } = await prettyPrint({ source, url });
     await sourceMaps.applySourceMap(source.id, url, code, mappings);
@@ -84,7 +86,6 @@ export function togglePrettyPrint(sourceId: string) {
     if (!source.isPrettyPrinted) {
       recordEvent("pretty_print");
     }
-    dispatch({ type: "TOGGLE_PRETTY_PRINTING" });
 
     if (!isLoaded(source)) {
       await dispatch(loadSourceText(source));
@@ -125,7 +126,6 @@ export function togglePrettyPrint(sourceId: string) {
       })
     );
 
-    dispatch({ type: "TOGGLE_PRETTY_PRINTING" });
     return newPrettySource;
   };
 }
--- a/src/actions/sources/prettyPrint.js
+++ b/src/actions/sources/prettyPrint.js
@@ -8,6 +8,7 @@ import assert from "../../utils/assert";
 import { recordEvent } from "../../utils/telemetry";
 import { remapBreakpoints } from "../breakpoints";
 
+import { closeTab } from "../tabs";
 import { setPausePoints, setSymbols } from "../ast";
 import { prettyPrint } from "../../workers/pretty-print";
 import { setSource } from "../../workers/parser";
@@ -84,6 +85,7 @@ export function togglePrettyPrint(sourceId: string) {
     if (!source.isPrettyPrinted) {
       recordEvent("pretty_print");
     }
+    dispatch({ type: "TOGGLE_PRETTY_PRINTING" });
 
     if (!isLoaded(source)) {
       await dispatch(loadSourceText(source));
@@ -124,6 +126,8 @@ export function togglePrettyPrint(sourceId: string) {
       })
     );
 
+    dispatch(closeTab(source));
+    dispatch({ type: "TOGGLE_PRETTY_PRINTING" });
     return newPrettySource;
   };
 }
--- a/src/actions/sources/prettyPrint.js
+++ b/src/actions/sources/prettyPrint.js
@@ -8,7 +8,6 @@ import assert from "../../utils/assert";
 import { recordEvent } from "../../utils/telemetry";
 import { remapBreakpoints } from "../breakpoints";
 
-import { closeTab } from "../tabs";
 import { setPausePoints, setSymbols } from "../ast";
 import { prettyPrint } from "../../workers/pretty-print";
 import { setSource } from "../../workers/parser";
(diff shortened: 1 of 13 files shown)


src/actions/sources/prettyPrint.js
```diff
@@ -124,6 +126,8 @@ export function togglePrettyPrint(sourceId: string) {
       })
     );
 
+    dispatch(closeTab(source));
+    dispatch({ type: "TOGGLE_PRETTY_PRINTING" });
```
jasonLaster: hmm, could we hook into the fact that the pretty printed source is loading and use that field?

https://github.com/devtools-html/debugger.html/blob/master/src/actions/sources/prettyPrint.js#L35-L44

src/actions/sources/prettyPrint.js
```diff
@@ -124,6 +126,8 @@ export function togglePrettyPrint(sourceId: string) {
       })
     );
 
+    dispatch(closeTab(source));
```
jasonLaster: why do we need this?

src/components/Editor/Footer.js
```diff
@@ -41,13 +42,21 @@ type Props = {
...
   prettyPrintButton() {
-    const { selectedSource, togglePrettyPrint } = this.props;
+    const { selectedSource, togglePrettyPrint, isPrettyPrinting } = this.props;
     const sourceLoaded = selectedSource && isLoaded(selectedSource);
 
     if (!shouldShowPrettyPrint(selectedSource)) {
       return;
     }
 
+    if (isPrettyPrinting) {
```
jasonLaster: could this be selectedSource isLoading and isPrettyPrinted?

src/actions/sources/prettyPrint.js
AnshulMalik: So, the Footer component doesn't know about that, and I think it would be hacky if we try to find what is the new pretty source and it it's loading or loaded?

src/actions/sources/prettyPrint.js
AnshulMalik: So, I picked it from https://github.com/devtools-html/debugger.html/issues/6947,

So close the original file when pretty printed one exists. It's a step at least.

src/components/Editor/Footer.js
AnshulMalik: Yeah, that will be awesome

src/actions/sources/prettyPrint.js
darkwing: I don't think we want this, or if we do use it, do it in a separate PR so it's easy to back out if needed.

src/components/Editor/Footer.js
```diff
@@ -41,12 +41,19 @@ type Props = {
...
 class SourceFooter extends PureComponent<Props> {
   prettyPrintButton() {
-    const { selectedSource, togglePrettyPrint } = this.props;
+    const { selectedSource, togglePrettyPrint, isPrettyPrinting } = this.props;
     const sourceLoaded = selectedSource && isLoaded(selectedSource);
 
     if (!shouldShowPrettyPrint(selectedSource)) {
       return;
     }
+    if (isLoading(selectedSource) || isPrettyPrinting) {
```
jasonLaster: do we need `isPrettyPrinting` or could we use `selectedSource.prettyPrinted`

AnshulMalik: When we pretty print a source, we generate a new pretty printed source, and then select it, we don't update the current source ands set prettyPrinted to be true.

jasonLaster: so if we pretty print and then immediately select it we should be good. I think that will help to because the editor will show loading immediately as well

src/utils/editor/source-documents.js
```diff
@@ -75,11 +75,9 @@ export function clearEditor(editor: SourceEditor) {
 }
 
 export function showLoading(editor: SourceEditor) {
-  if (hasDocument("loading")) {
-    return;
-  }
+  const docExists = hasDocument("loading");
 
-  const doc = editor.createDocument();
+  const doc = docExists ? getDocument("loading") : editor.createDocument();
```
jasonLaster: maybe we just want to return if we have the doc...

AnshulMalik: If we just return from there, it doesn't display `loading...` since we are not replacing the editor text, so I see the text of original source and the spinner.

jasonLaster: got it. maybe get and then replace..

src/components/Editor/index.js
```diff
@@ -577,7 +577,7 @@ class Editor extends PureComponent<Props, State> {
         <EmptyLines editor={editor} />
         <Breakpoints editor={editor} />
         <Preview editor={editor} editorRef={this.$editorWrapper} />;
-        <Footer editor={editor} horizontal={horizontal} />
+        <Footer horizontal={horizontal} />
```
jasonLaster: i dont think we needed the prior empty UI.

src/utils/editor/index.js
```diff
@@ -64,6 +64,10 @@ export function shouldShowPrettyPrint(selectedSource) {
     return false;
   }
 
+  if (isLoading(selectedSource)) {
+    return true;
+  }
```
jasonLaster: shouldnt we also check if the source is pretty?

src/components/Editor/Footer.js
```diff
@@ -47,6 +46,13 @@ class SourceFooter extends PureComponent<Props> {
     if (!shouldShowPrettyPrint(selectedSource)) {
```
jasonLaster: it seems like this is weird now that we have a new case.

I propose doing: 

```diff

  if (isLoading(selectedSource) && selectedSource.pretty) {
      // show loading UI
  }
 if (!shouldShowPrettyPrint(selectedSource)) {
    return
  }
  ...
```

jasonLaster:
http://g.recordit.co/DQNVX54oY2.gif

digitarald:
I love how the new tab with loading state appears right after click.

src/utils/editor/index.js
AnshulMalik: Now that we are showing some loading, we should do it for any source while it's loading?

darkwing:
I played around with this :  https://gist.github.com/darkwing/120ee1afb15988373c20619344cff366

One thing that bugged me about the loader is how it abruptly disappeared.  This update keeps the prettyprint button there at all times but shows it as disabled when the source is pretty.

disabled

darkwing:
I love this 👍 💯

PR #6980, opened 2018-09-13
Making sure that accessible rep only supports grips of accessible ob…

### LINKED ISSUE #6932: Reps: add a rep for Accessible actor.

yzen:
…jects and not plain accessible fronts (#6932).

Fixes #6932

Currently a match for accessible rep checks for typeName this is too general since typeName property also exists is front accessible objects. We need to make it more specific and also check if preview property is available which indicates it is a grip.

### Summary of Changes

Added a check for existence of the ```preview``` property when matching accessible rep.

### DIFF
--- a/packages/devtools-reps/src/reps/accessible.js
+++ b/packages/devtools-reps/src/reps/accessible.js
@@ -106,7 +106,7 @@ function supportsObject(object, noGrip = false) {
     return false;
   }
 
-  return object.typeName && object.typeName === "accessible";
+  return object.preview && object.typeName && object.typeName === "accessible";
 }
 
 // Exports from this module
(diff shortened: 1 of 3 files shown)


nchevobbe:
could you add a test that is failing without the fix and make it pass ? This way we don't regress in the future.
Also,  Travis is failing, but I think I saw that failure in other PRs. @darkwing @jasonLaster could you confirm the failure is not related ? Thanks !

yzen:
> could you add a test that is failing without the fix and make it pass ? This way we don't regress in the future.
Added a test.

PR #6983, opened 2018-09-14
Return what was mapped in mapExpression.

nchevobbe:
In `mapExpression`, we now return an object instead of the mapped expression directly.
The object contains the mapped expression, as well as a `mapped` object containing
all the mapping the happened on the expression or not (await, originalExpression, bindings).

We take this as an opprtunity to clean-up `mapAwaitExpression`:
 - the default exported function is renamed to `mapTopLevelAwait` to match other mapping functions
 - we add a `.catch` in which we `console.error` to catch any promise rejection or Exception.

### DIFF
--- a/src/actions/expressions.js
+++ b/src/actions/expressions.js
@@ -178,10 +178,12 @@ export function getMappedExpression(expression: string) {
     // 2. does not contain `await` - we do not need to map top level awaits
     // 3. does not contain `=` - we do not need to map assignments
     if (!mappings && !expression.match(/(await|=)/)) {
-      return null;
+      // TODO: return null when
+      // https://bugzilla.mozilla.org/show_bug.cgi?id=1491354 lands.
+      return expression;
     }
 
-    const mapResult = parser.mapExpression(
+    const mapResult = await parser.mapExpression(
       expression,
       mappings,
       bindings || [],
--- a/src/actions/expressions.js
+++ b/src/actions/expressions.js
@@ -144,7 +144,8 @@ function evaluateExpression(expression: Expression) {
         !isGeneratedId(sourceId) &&
         !isGeneratedId(selectedSource.id)
       ) {
-        input = await dispatch(getMappedExpression(input));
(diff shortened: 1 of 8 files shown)


nchevobbe:
the returned `mapped` object will be used in the console to *not* print the Promise we create by wrapping the await in an async function.

nchevobbe:
Travis is rightfully not happy since the change here would need some modifications in mozilla-central

nchevobbe:
> I'm not sure if doing a string comparison is correct because the transform can re-format the expression, but not change its meaning

Yeah I know. In the test for example, there are some cases that go from `a` to `a;`, so it's marked as mapped, but ideally should not.

jasonLaster:
JSTerm is getting a bad result: 
https://searchfox.org/mozilla-central/source/devtools/client/webconsole/components/JSTerm.js

This looks alright:
https://github.com/devtools-html/debugger.html/blob/master/assets/panel/panel.js#L100

jasonLaster:
this should help

```diff
diff --git a/src/workers/parser/mapOriginalExpression.js b/src/workers/parser/mapOriginalExpression.js
index efb9955..ad9afe0 100644
--- a/src/workers/parser/mapOriginalExpression.js
+++ b/src/workers/parser/mapOriginalExpression.js
@@ -42,7 +42,7 @@ export default function mapOriginalExpression(
 ): string {
   const ast = parseScript(expression, { allowAwaitOutsideFunction: true });
   const scopes = buildScopeList(ast, "");
-
+  let shouldUpdate = false;
   const nodes = new Map();

   const replacements = new Map();
@@ -94,9 +94,14 @@ export default function mapOriginalExpression(

     const replacement = replacements.get(locationKey(node.loc.start));
     if (replacement) {
+      shouldUpdate = true;
       replaceNode(ancestors, t.cloneNode(replacement));
     }
   });

+  if (!shouldUpdate) {
+    return expression;
+  }
+
   return generate(ast).code;
```

PR #6992, opened 2018-09-18
Use F8 for both pause and breakOnNext commands

### LINKED ISSUE #6963: [shortcuts] pause on f8
the debugger documents that F8 should trigger "break on next statement", yet the client doesn't actually have that command. 

https://github.com/devtools-html/debugger.html/blob/master/src/components/SecondaryPanes/CommandBar.js#L30-L31

https://github.com/devtools-html/debugger.html/blob/master/src/components/SecondaryPanes/CommandBar.js#L34-L51

kenjiO:
Fixes #6963

### Summary of Changes
Update the `CommandBar.js` component so the handler for the F8 key press will call either `props.breakOnNext` or `props.resume` depending on the state of `props.isPaused`.  Previously two F8 event handlers were created.  One called `props.resume` always regardless of paused state.  The other one called a non-existant `props.pause` function.

### Test Plan

- [x] jest tests
- [x] test `command \` shortcut key breaks on next and pauses on a mac
- [x] test `F8` shortcut key breaks on next and pauses on a mac
- [ ] test `F8` shortcut key breaks on next and pauses on windows
- [ ] test `F8` shortcut key breaks on next and pauses on linux

### DIFF
--- a/src/components/SecondaryPanes/CommandBar.js
+++ b/src/components/SecondaryPanes/CommandBar.js
@@ -177,10 +177,7 @@ class CommandBar extends Component<Props> {
         () => this.resume(),
         "resume",
         "active",
-        L10N.getFormatStr(
-          "resumeButtonTooltip",
-          formatKey("resume")
-        )
+        L10N.getFormatStr("resumeButtonTooltip", formatKey("resume"))
       );
     }
 
--- a/src/components/SecondaryPanes/CommandBar.js
+++ b/src/components/SecondaryPanes/CommandBar.js
@@ -26,24 +26,26 @@ import { appinfo } from "devtools-services";
 
 const isMacOS = appinfo.OS === "Darwin";
 
-const COMMANDS = ["resumeOrBreakOnNext", "stepOver", "stepIn", "stepOut"];
+// NOTE: the "resume" command will call either the resume or breakOnNext action
+// depending on whether or not the debugger is paused or running
+const COMMANDS = ["resume", "stepOver", "stepIn", "stepOut"];
 
 const KEYS = {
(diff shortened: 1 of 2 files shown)


kenjiO:
I can't figure out what is causing this flow error:

```
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ src/components/SecondaryPanes/CommandBar.js:59:20

Cannot get osActions[action] because:
 • property resume is missing in object literal [1].
 • property resume is missing in object literal [2].

 [1] 32│   WINNT: {
     33│     resumeOrBreakOnNext: "F8",
     34│     stepOver: "F10",
     35│     stepIn: "F11",
     36│     stepOut: "Shift+F11"
     37│   },
       :
 [2] 45│   Linux: {
     46│     resumeOrBreakOnNext: "F8",
     47│     stepOver: "F10",
     48│     stepIn: "Ctrl+F11",
     49│     stepOut: "Ctrl+Shift+F11"
     50│   }
       :
     56│
     57│ function getKeyForOS(os, action) {
     58│   const osActions = KEYS[os] || KEYS.Linux;
     59│   return osActions[action];
     60│ }
     61│
     62│ function formatKey(action) {
```

darkwing:
You're seeing that error message because you changed the "resume" key in each OS type to "resumeOrBreakOnNext"; maybe from line 16/17.  Do we need to rename that property?

kenjiO:
@darkwing Should I change the name back to 'resume' even though this is now being used for both resume and break on next?

darkwing:
Yeah, let's keep it simple.

PR #6998, opened 2018-09-20
[Parser] getFramework false positive angularjs detection

### LINKED ISSUE #6833: [tab] debugger.js tab has an angular icon!
So, when I am debugging the debugger and open the debugger's webpack bundle, the bundle is given an angular icon! This is quite silly because the debugger is built with React and bundled with Webpack. Angular is great, but we don't use angular.

This is definitely a top 3 silly bug! 

The reason for the bug is that our framework check is a bit relaxed for angular. We check for two things: 

1. a variable with the label angular
2. an expression `controller.` or `module.` 

[here](https://github.com/devtools-html/debugger.html/blob/master/src/workers/parser/frameworks.js#L68-L76)

these two things can happen without us actually using angular!

The fix is to tighten the check, but i'm not quite sure how because I'm not familiar with angular :)

screen shot 2018-08-15 at 9 23 08 am

brunoscopelliti:
Fixes #6833 

### Summary of Changes

* Fix `hasAngularExpressions` method, so that it checks not just the property name, but also the host object name (eg. occurrences of `angular.controller`, instead of simply `controller`). 
* Added test case for false positive detection.
* Organised unit test in framework specific

### Test Plan

debugger

I wasn't able to reproduce the issue in a running instance of devtools; I've written a failing test that exposed the problem, and then made it turns green.

### DIFF
--- a/src/workers/parser/frameworks.js
+++ b/src/workers/parser/frameworks.js
@@ -71,7 +71,9 @@ const identifiesAngular = identifiers => {
 
 const hasAngularExpressions = memberExpressions => {
   return memberExpressions.some(
-    item => item.name == "controller" || item.name == "module"
+    item =>
+      item.expression == "angular.controller" ||
+      item.expression == "angular.module"
   );
 };
 
--- a/src/workers/parser/tests/fixtures/frameworks/angular1FalsePositive.js
+++ b/src/workers/parser/tests/fixtures/frameworks/angular1FalsePositive.js
@@ -1,0 +1,11 @@
+const frameworkStars = {
+  angular: 40779,
+  react: 111576,
+  vue: 114358,
+};
+
+const container = new Container();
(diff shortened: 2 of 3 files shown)


brunoscopelliti:
One consequence of this changeset is that now we can get rid of `identifiesAngular` check, because it implicitly passes everytime `hasAngularExpressions` returns true.

https://github.com/devtools-html/debugger.html/blob/4081589d55da350244d9ad6884c9f7db7e6b8f22/src/workers/parser/frameworks.js#L61-L66

I haven't changed that line yet, cause I'm not sure I'm missing something.

jasonLaster:
Looks good. Feel free to drop `identifiesAngular,

brunoscopelliti:
Done, the check now is done directly in `isAngularComponent` (similarly to `isVueComponent`).

brunoscopelliti:
Thanks Jason, I'll give a look to jest cases.

PR #7006, opened 2018-09-21
UX changes for issue #6803

### LINKED ISSUE #6803: Consistent top bar height and baseline alignment
Two spacing issues with the debugger's top bars:
debugger-top-bar-height

1. The top bar height *excluding borders* in the Debugger is 29px. Compare to the DevTools tab bar’s 28px. That’s not a huge difference, but it would be nice to be fully consistent.

2. The file tab labels are 2px too low, compared to the Sources|Outline tabs and other icons.

My recommended fix for the second issue (which is the most visible IMO) would be to ditch the tab style with a border on top, and go with a flatter style like for "Sources"|"Outline". It might also be possible to use the devtools-tab style (like the Inspector and Netmonitor do, at 24px-high), making the Debugger feel a bit less out of place.

Or the alignment can be achieved with the current design, by tweaking a few pixels.

Doomwyte:
Fixes #6803 

### Summary of Changes

* Changed height of Debugger's tab bar to 28px excluding borders
* Aligned the file tab labels and changed to look like the Source | Outline tabs

### Test Plan

-Tested hovering action
-Tested multiple file tabs open

### Screenshots
screen shot 2018-09-21 at 4 48 06 pm

### DIFF
--- a/src/components/Editor/Tabs.css
+++ b/src/components/Editor/Tabs.css
@@ -5,7 +5,7 @@
 .source-header {
   border-bottom: 1px solid var(--theme-splitter-color);
   width: 100%;
-  height: 30px;
+  height: 29px;
   display: flex;
 }
 
@@ -50,9 +50,8 @@
   overflow: hidden;
   padding: 5px;
   margin-inline-start: 3px;
-  margin-top: 3px;
   cursor: default;
-  height: 27px;
+  height: 29px;
 }
 
 .source-tab:first-child {
@@ -61,7 +60,7 @@
 
 .source-tab:hover {
   background-color: var(--theme-toolbar-background-alt);
-  border-color: var(--theme-splitter-color);
+  border-bottom-color: var(--theme-splitter-color);
 }
 
 .source-tab.active {
@@ -69,6 +68,7 @@
   background-color: var(--theme-body-background);
   border-color: var(--theme-splitter-color);
   border-bottom-color: transparent;
+  border-top-color: transparent;
 }
 
 .source-tab.active path,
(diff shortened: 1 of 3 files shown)


darkwing:
This has the same bottom border issue for the selected tab as https://github.com/devtools-html/debugger.html/pull/7011 ; both are almost ready for design review!   Nice!

darkwing:
@violasong What do you think about @fvsch 's suggestion of removing the "tab" interface and going with a flatter tab which looks like the sources and outline panes?  I'd really like your feedback.  It appears the only tab-like interfaces we have are in the perf pane, and they look way different:

perftabs

Thoughts?

jasonLaster:
I am okay dropping the tab interface. 

I think i would prefer re-using the blue border over the background though.

Chrome does this pretty well imo

screen shot 2018-09-24 at 6 52 01 am

fvsch:
>  It appears the only tab-like interfaces we have are in the perf pane

Other tools using tabs: Inspector and Network (request details, also seen in the Console).
For the sake of consistency it might be better to use the same tabs.

jasonLaster:
The inspector tabs are nice:

Doomwyte:
I have created a new issue to change the Debugger tabs #7034

violasong:
Thanks all for looking at this - I love the idea of changing both the file tabs and the sources/outline tabs to the Inspector-style simpler blue border tabs. 

Debugger's file tabs are more similar to the Firefox main tabs in having highly variable, often truncated lengthy text, so it is possible we'll need to do a slight modification on these to give it enough separation. I'd love to see how just Inspector-style will look in Nightly, and can propose more UI ideas if needed.

PR #7011, opened 2018-09-22
Alternative UX changes for issue 6803

### LINKED ISSUE #6803: Consistent top bar height and baseline alignment

Doomwyte:
Fixes #6803

The alternative approach suggested in the issue where the file tab design does not change. Original approach: https://github.com/devtools-html/debugger.html/pull/7006

### Summary of Changes

* Changed height of Debugger's tab bar to 28px excluding borders
* Adjusted file tab labels to align

### Test Plan

-Tested hovering action
-Tested multiple tabs open

### Screenshots
screen shot 2018-09-22 at 11 02 15 am

### DIFF
--- a/src/components/Editor/Tabs.css
+++ b/src/components/Editor/Tabs.css
@@ -52,7 +52,7 @@
   margin-inline-start: 3px;
   margin-top: 3px;
   cursor: default;
-  height: 26px;
+  height: 27px;
 }
 
 .source-tab:first-child {
--- a/src/components/Editor/Tabs.css
+++ b/src/components/Editor/Tabs.css
@@ -5,7 +5,7 @@
 .source-header {
   border-bottom: 1px solid var(--theme-splitter-color);
   width: 100%;
-  height: 30px;
+  height: 29px;
   display: flex;
 }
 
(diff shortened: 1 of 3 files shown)


darkwing:
Oooooh, this is so close!  The only issue is I'm seeing an odd bottom border here:
weirdbottomborder

I think it's a result of changing the height of one of these elements.  Once we fix this border issue, I think this will be ready to merge!

PR #7015, opened 2018-09-23
Improve regex pattern for determining React library files

### LINKED ISSUE #6792: Framework collapsing also collapses user-land component
STR:
- Debug https://firefox-debugger-example-create-react-app.glitch.me/ with Pause on Exception
- Set an X and try to set an O on the same field to raise an error

ER: `handleClick` has its own frame in the callstack
AR: `handleClick` is part of a collapsed React callstack entry

image

Is the collapsing logic too broad in this case, maybe thrown off my too many `react` words in the project?

danny6514:
Improve regex pattern for determining whether a particular frame belongs to the React library or not.

Fixes #6792 

### Summary of Changes

* Improve React regex pattern to detect React files correctly. Previously, if the word `react` appeared anywhere in the URL, the debugger will classify it as React library file. Now, the debugger will look for the word `react` in the filename rather than just the path.

### Test Plan

- [x] Test using the provided URL in the issue, ensure the functions in `Game.js` and `Board.js` are showing separate from the React group.

### Screenshots/Videos

Before:
image
image

After:
image
image

### DIFF
--- a/src/utils/pause/frames/tests/getLibraryFromUrl.spec.js
+++ b/src/utils/pause/frames/tests/getLibraryFromUrl.spec.js
@@ -77,7 +77,9 @@ describe("getLibraryFromUrl", () => {
         "https://debugger-example.com/react.development.js",
         "https://debugger-example.com/react.production.min.js",
         "https://debugger-react-example.com/react.js",
-        "https://debugger-react-example.com/react/react.js"
+        "https://debugger-react-example.com/react/react.js",
+        "/node_modules/react/test.js",
+        "/node_modules/react-dom/test.js"
       ];
       reactUrlList.forEach(reactUrl => {
         const frame = {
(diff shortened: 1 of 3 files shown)


danny6514:
Changes made in this PR is pending the discussion in #6792

jasonLaster:
Looks good to me.

Lets add some unit tests for this change too

danny6514:
@jasonLaster : Tests added

src/utils/pause/frames/tests/getLibraryFromUrl.spec.js
```diff
@@ -49,4 +49,50 @@ describe("getLibraryFromUrl", () => {
...
+          location: {
+            line: 12
+          },
+          source: {
+            url: reactUrl
+          }
+        };
+        expect(getLibraryFromUrl(frame)).toEqual("React");
+      });
+    });
```
jasonLaster: have you seen how we use `cases` in our test suite. I think you'll like it :)

PR #7023, opened 2018-09-25
prevent a regex infinite loop in the search worker

### LINKED ISSUE #6896: [search] file search with regex using lots of memory
When searching a file for a regex the memory usage goes up by gigabytes.  I am able to get this to happen on my mac with regexes that contain the `*` and `^` special characters.

kenjiO:
Fixes #6896

### Summary of Changes

* Add check in the search worker regex search loop to catch if the regex index is not being updated

### Test Plan

Example test plan:

- [x] Add a jest test
- [x] Ran the debugger and verified searching with regex 'a*' did not cause memory to increase wildly
- [x] Ran the debugger and verified searching with regex '' did not cause memory to increase wildly

### DIFF
--- a/src/workers/search/get-matches.js
+++ b/src/workers/search/get-matches.js
@@ -20,8 +20,17 @@ export default function getMatches(
   for (let i = 0; i < lines.length; i++) {
     let singleMatch;
     const line = lines[i];
+    let previousLastIndex = regexQuery.lastIndex;
     while ((singleMatch = regexQuery.exec(line)) !== null) {
       matchedLocations.push({ line: i, ch: singleMatch.index });
+
+      // exec() is supposed to update the lastIndex property but sometimes
+      // doesn't and can cause this to be an infinite loop by matching on the
+      // same index.  A work around is to update the lastIndex manually
+      if (regexQuery.lastIndex === previousLastIndex) {
+        regexQuery.lastIndex++;
+      }
+      previousLastIndex = regexQuery.lastIndex;
     }
   }
   return matchedLocations;
--- a/src/workers/search/get-matches.js
+++ b/src/workers/search/get-matches.js
@@ -20,7 +20,7 @@ export default function getMatches(
   for (let i = 0; i < lines.length; i++) {
     let singleMatch;
     const line = lines[i];
-    let previousLastIndex = regexQuery.lastIndex;
+    let previousLastIndex = 0;
     while ((singleMatch = regexQuery.exec(line)) !== null) {
       matchedLocations.push({ line: i, ch: singleMatch.index });
 
--- a/src/workers/search/get-matches.js
+++ b/src/workers/search/get-matches.js
@@ -29,7 +29,7 @@ export default function getMatches(
       // increment it manually in that case.  See issue #7023
       if (singleMatch[0] === "") {
         assert(
-          regexQuery.unicode,
+          !regexQuery.unicode,
           "lastIndex++ can cause issues in unicode mode"
         );
         regexQuery.lastIndex++;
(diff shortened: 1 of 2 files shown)


src/workers/search/get-matches.js
```diff
@@ -20,8 +20,17 @@ export default function getMatches(
   for (let i = 0; i < lines.length; i++) {
     let singleMatch;
     const line = lines[i];
+    let previousLastIndex = regexQuery.lastIndex;
     while ((singleMatch = regexQuery.exec(line)) !== null) {
```
kenjiO: According to [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/exec#Finding_successive_matches) this should work fine but I found that in some cases this is creating an infinite loop because `regexQuery.exec()` is not updating `regexQuery.lastMatch` and it keeps returning the same match over and over again.  Is this a bug in the javascript engine?

src/workers/search/tests/get-matches.spec.js
```diff
@@ -71,5 +71,17 @@ describe("search", () => {
       });
       expect(matchLocations).toHaveLength(0);
     });
+
+    // regression test for #6896
+    it("doesn't crash on the regex 'a*'", () => {
```
kenjiO: If you comment out the fix and run this test it will crash with an out of memory error.  I couldn't figure out a way to prevent the crash and fail more elegantly.

jasonLaster:
Hey @loganfsmyth I'm curious what you think

src/workers/search/get-matches.js
```diff
@@ -20,8 +20,17 @@ export default function getMatches(
...
+    let previousLastIndex = 0;
     while ((singleMatch = regexQuery.exec(line)) !== null) {
       matchedLocations.push({ line: i, ch: singleMatch.index });
+
+      // exec() is supposed to update the lastIndex property but sometimes
+      // doesn't and can cause this to be an infinite loop by matching on the
+      // same index.  A work around is to update the lastIndex manually
+      if (regexQuery.lastIndex === previousLastIndex) {
+        regexQuery.lastIndex++;
+      }
```
kenjiO: After looking over this again I didn't like it as much as it seemed to hacky.  But I when I looked up [lastIndex on MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/lastIndex) it seems like setting `lastIndex` yourself to indicate the next index to start matching at is something there by design and is in the spec.  So maybe this is OK after all.

src/workers/search/get-matches.js
```diff
@@ -20,8 +20,17 @@ export default function getMatches(
...
     let singleMatch;
     const line = lines[i];
+    let previousLastIndex = 0;
     while ((singleMatch = regexQuery.exec(line)) !== null) {
       matchedLocations.push({ line: i, ch: singleMatch.index });
+
+      // exec() is supposed to update the lastIndex property but sometimes
+      // doesn't and can cause this to be an infinite loop by matching on the
+      // same index.  A work around is to update the lastIndex manually
+      if (regexQuery.lastIndex === previousLastIndex) {
```
loganfsmyth: Instead of saving `previousLastIndex` here, we can check `singleMatch[0] === ""` to keep things simpler.

Also, I think it's fine, but it might also be good to add an assertion here that `!regexQuery.unicode` since incrementing this index by a single codeunit could cause issues if the regex was trying to match in unicode mode.

If you're curious, I didn't think about the unicode issue either, but I looked at https://tc39.github.io/proposal-string-matchall/#%RegExpStringIteratorPrototype%.next since it has to deal with this same issue. You can see step `11.a.ii.2` it does the same type of thing.

loganfsmyth:
Thanks!

PR #7024, opened 2018-09-25
In getMappedExpression, return the result of parser.mapExpression.

nchevobbe:
Now that Bug 1491354 landed, the console will expect
`getMappedExpression` function to return the same object
than parser's `mapExpression` function, or null if the
expression doesn't need any mapping.


jasonLaster:
Thanks, the PR looks good. I'm seeing a build issue though

>  0:05.48         Error processing test manifest /home/travis/build/devtools-html/debugger.html/firefox/devtools/client/application/test/browser.ini: entry in support-files not present in the srcdir: !/devtools/client/debugger/new/test/mochitest/helpers/context.js

PR #7027, opened 2018-09-25
[breakpoints] map sources in pane

jasonLaster:
### Summary of Changes

* changes the breakpoint locations in the pane based on the selected file context

### Test Plan

* jest tests

### Screenshots

### DIFF
--- a/src/utils/source-maps.js
+++ b/src/utils/source-maps.js
@@ -6,7 +6,9 @@
 
 import { isOriginalId } from "devtools-source-map";
 import { getSource } from "../selectors";
-import type { Location, Source } from "../types";
+
+import type { Location, MappedLocation, Source } from "../types";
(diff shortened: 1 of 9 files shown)


src/utils/source-maps.js
```diff
@@ -46,3 +49,9 @@ export async function getMappedLocation(
 
   return sourceMaps.getOriginalLocation(location, source);
 }
+
+export function getSelectedLocation(mappedLocation: MappedLocation, selectedSource: Source) {
+  return selectedSource && isGenerated(selectedSource)
```
darkwing: You don't need the `selectedSource` check since `selectedSource: Source` doesn't have a `?`, right?

jasonLaster:
breakpoints-actions.js

PR #7031, opened 2018-09-25
[WIP] Issue #5515 - ParsedURL with queue string.

### LINKED ISSUE #5515: [Sources] Debugger does not distinguish scripts with different GET parameters
Create a page with

```html
<script src="script.php?param=1"></script>
<script src="script.php?param=2"></script>
```

where that script is

```php
<?php
if ($_GET["param"] == "1") {
  echo "var a = 1; debugger;\n";
} else {
  echo "var a = 2;\n";
}
?>
```

Due to `debugger`, the debugger will show a script.php tab with the first script. Resume the script, right-click the tab and choose "Reveal in tree". This will open a tab with the second script!

In Sources, script.php is listed only once, and refers to the second one.

ZhangYue-vhyt:
Fixes #5515 

# Done
The `ParsedURL` from `getURL.js` now has `searchParams` (queue string).

# Not Done
Display the queue string got from ParsedURL in the source tree.


.gitignore
```diff
@@ -24,3 +24,7 @@ out
 junit.xml
 .cache/
 .rpt2_cache
+.vs/slnx.sqlite
```
darkwing: Can we do a blanket `.vs/` instead?

ZhangYue-vhyt: Sure.

src/utils/sources-tree/getURL.js
```diff
@@ -37,7 +38,8 @@ function _getURL(source: Source, defaultDomain: string): ParsedURL {
     return def;
   }
 
-  const { pathname, protocol, host } = parse(url);
+  const { pathname, searchParams, protocol, host } = parse(url);
```
darkwing: It would be better to use the `search` property instead.  The value of `searchParams` is an object, and we don't need an object; `search` provides just the string `s=1&t=2`, etc.

ZhangYue-vhyt: I have replaced `searchParams` with `search` on my computer. Should I push to the origin and make a new pull request? Can I do the changes in this pull request?

darkwing:
Hello @ZhangYue-vhyt ; I just caught this but your changes are on `master` branch; you always want to be making your changes on a feature branch, for example:

```
git checkout master
git pull upstream master # make sure you have the latest code from mozilla's repo
git checkout -b my-branch-name
```

Since you've made these changes on master, I can't jump in and change anything

src/utils/sources-tree/getURL.js
```diff
@@ -10,6 +10,7 @@ import { getUnicodeHostname, getUnicodeUrlPath } from "devtools-modules";
 import type { Source } from "../../types";
```
darkwing: We don't need any of these changes anymore; these were left over from your first attempt.

src/reducers/sources.js
```diff
@@ -377,6 +381,18 @@ function getSourcesByUrlInSources(
   return urls[url].map(id => sources[id]);
 }
 
+function getSourcesByUrlInSources5515(
```
darkwing: We shouldn't need this function; we should be able to do any structuring of the data in `getHasSiblingOfSameName`

ZhangYue-vhyt: Do you mean all the changes should be done in the `src/components/PrimaryPanes/SourcesTreeItem.js`?

darkwing: Yes; it isn't apparent to me why we would need this extra function when we could do filtering in `getHasSiblingOfSameName`.  Also, why are we checking for `undefined`?

darkwing:
Closing due to a new PR from a new branch

PR #7032, opened 2018-09-25
Tweak memoize to use n keys

### LINKED ISSUE #6938: [Editor] Removing breakpoint from pretified file doesn't remove it from the original
STR:

1) Load https://impossible-education.glitch.me/
2) Go to the debugger UI and open `impossible-education/client.js` file
3) Click the `Pretty print source` at the bottom of the toolbox UI (looks like `{}`)
3) You should be auto navigated to the new file (i.e. no spaces at the top)
4) Create breakpoints on lines  2, 5, 8
5) Go to the original, breakpoints are on wrong lines
6) Go to pretty printed file and remove all breakpoints
7) Go back to the original, breakpoints are still there -> BUG
(btw. they  can't be removed in the original file atm)

Looks to me like source-maps problem.

But could be: breakpoints, source-maps or pretty-print.

Honza

AnshulMalik:
Fixes #6938

### Summary of Changes

* We can now use Memoize to cache upto n arguments serially

memoize

### DIFF
--- a/src/utils/tests/memoize.spec.js
+++ b/src/utils/tests/memoize.spec.js
@@ -1,0 +1,47 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
+
+import memoize from "../memoize";
+
+const a = {
+  number: 3
+};
+const b = {
+  number: 4
+};
+const c = {
+  number: 5
+};
+const d = {
+  number: 6
+};
+
+function add(...numberObjects) {
+  return numberObjects.reduce((prev, cur) => prev + cur.number, 0);
+}
+
+describe("memozie", () => {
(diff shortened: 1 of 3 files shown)


src/utils/tests/memoize.spec.js
```diff
@@ -0,0 +1,47 @@
...
+  it("should work for two args as key", () => {
+    const memoizedAdd = memoize(2, add);
+    expect(memoizedAdd(a, b, d)).toEqual(13);
+    expect(memoizedAdd(a, b, c)).toEqual(13);
+    expect(memoizedAdd(b, c)).toEqual(9);
+    expect(memoizedAdd(b, a)).toEqual(7);
+  });
+
+  it("should work for three args as key", () => {
+    const memoizeAdd = memoize(3, add);
```
jasonLaster: can avoid having an explicit `n`?

AnshulMalik: It was intentional, just in case we don't want to use all arguments as keys.

PR #7033, opened 2018-09-25
Fix 7026 - Only show (mapped) when duplicate file names

### LINKED ISSUE #7026: [sources] only show mapped when there is a conflict
STR:

1. go to https://firefox-dev.tools/debugger-examples/examples/increment/
2. view sources tree

We should only show mapped if there is both an original and generated source with the same name for example: https://w6q8w3lz47.codesandbox.io/

image

darkwing:
Fixes #7026

I was going to give this issue to @ZhangYue-vhyt but, in reviewing the code, I realized I wasn't very concise with my original `isOriginal` pull request.

`getSourcesByURL` should return all sources with a given URL; `isOriginal` shouldn't be in play with that function.

### DIFF
--- a/src/components/PrimaryPanes/SourcesTreeItem.js
+++ b/src/components/PrimaryPanes/SourcesTreeItem.js
@@ -187,13 +187,15 @@ class SourceTreeItem extends Component<Props, State> {
 }
 
 function getHasMatchingGeneratedSource(state, source: ?Source) {
-  if (!source) {
+  if (!source || !isOriginalSource(source)) {
     return false;
   }
 
-  const isOriginal = isOriginalSource(source);
-  const sources = getSourcesByURL(state, source.url, isOriginal);
-  return isOriginal && sources.length > 0;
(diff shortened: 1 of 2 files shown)


darkwing:
@ZhangYue-vhyt Could you review this?  URLs to test:

- https://firefox-dev.tools/debugger-examples/examples/increment/ (Should *not* show `(mapped)`)
- https://q7mmol9xy9.codesandbox.io/ (Should show mapped and two sources with the same name)

ZhangYue-vhyt:
I did the test. Both show (mapped) under the "Webpack" source.

darkwing:
Are you sure you brought down my changes using git?  I'm not seeing the `(mapped)` text at https://firefox-dev.tools/debugger-examples/examples/increment/ anyore.

ZhangYue-vhyt:
I have copied your changes and do a test. Both of them don't show (mapped).

jasonLaster:
I did not manually test though

bomsy:
@darkwing @jasonLaster Tested!

PR #7044, opened 2018-09-26
[Editor] continue resizing breakpoints

jasonLaster:
### Summary of Changes

when we introduced some of the codemirror operations we began to see the breakpoint sizes update  late when we switched files. This fixes that by removing one of the operation groups.


bomsy:
Found this scenario where if we land on a page with larger gutter sizes  first, its initially broken

jasonLaster:
Nice!

PR #7061, opened 2018-09-27
Fix 2 mochitests failure issues

loganfsmyth:


.gitignore
```diff
@@ -1,5 +1,5 @@
 .DS_Store
-build
+/assets/build
```
loganfsmyth: This was too general, so were ignoring the `build` folder we actually wanted.

src/actions/ast.js
```diff
@@ -56,7 +56,7 @@ export function setSymbols(sourceId: SourceId) {
   return async ({ dispatch, getState }: ThunkArgs) => {
     const source = getSourceFromId(getState(), sourceId);
 
-    if (source.isWasm || hasSymbols(getState(), source)) {
+    if (source.isWasm || hasSymbols(getState(), source) || !isLoaded(source)) {
```
loganfsmyth: This was causing a race condition making a bunch of pretty-printing tests flaky. Looks like the flakiness was introduced by this line: https://github.com/devtools-html/debugger.html/pull/6977/files#diff-87aac1ffafaa2e75da681cb07b73be1cR48

Because it selects the source before it is loaded, it has the potential to call `setSymbols` before there is actually code to process. That was causing us to generate empty symbol data, but the actual test failure was caused because `setSourceMetaData` checks `isLoaded`, which meant it was possible for it to bail out, and then never run again because this `hasSymbols(getState(), source)` would be `true` on the second call to `setSymbols` even though it wasn't loaded and thus the first `setSourceMetaData` didn't actually set the metadata. That meant the tests for pretty-printing were blocking forever waiting for metadata.

PR #7063, opened 2018-09-28
[WIP] ObjectInspector - Added expanded node

### LINKED ISSUE #6893: ObjectInspector expanded node could be more useful
In [Bug 1486381](https://bugzilla.mozilla.org/show_bug.cgi?id=1486381), @fvsch suggest that the way we represent expanded ObjectInspector nodes is not the best.

Currently, the nodes are being rendered in `TINY` mode, but we could introduce a new `expanded` prop to use in Reps.

So instead of having an object rendered as

```
▼ { … }
|   a: 1
|   b: 2
|   <prototype>
```

It could be 

```
▼ Object {
|   a: 1
|   b: 2
|   <prototype>
}
```
or at least

```
▼ Object
|   a: 1
|   b: 2
|   <prototype>
```
if we don't want to deal with opening and closing brackets.

marcogallo566:
Fixes #6893

Here's the Pull Request Doc

### Summary of Changes

* Removed unnecessary brackets

Feel free to improve it!

### Videos
sep-28-2018 11-57-19

### DIFF
--- a/packages/devtools-reps/src/object-inspector/component.js
+++ b/packages/devtools-reps/src/object-inspector/component.js
@@ -285,8 +285,13 @@ class ObjectInspector extends Component<Props> {
       if (depth > 0) {
         repProps.mode = this.props.mode === MODE.LONG ? MODE.SHORT : MODE.TINY;
       }
+
       if (expanded) {
-        repProps.mode = MODE.EXPANDED;
+        if (this.props.autoExpandDepth === 0) {
+          repProps.mode = MODE.TINY;
+        } else {
+          repProps.mode = MODE.EXPANDED;
+        }
       }
 
       if (nodeIsLongString(item)) {
--- a/packages/devtools-reps/src/object-inspector/component.js
+++ b/packages/devtools-reps/src/object-inspector/component.js
@@ -286,7 +286,7 @@ class ObjectInspector extends Component<Props> {
         repProps.mode = this.props.mode === MODE.LONG ? MODE.SHORT : MODE.TINY;
       }
       if (expanded) {
-        repProps.mode = MODE.TINY;
+        repProps.mode = MODE.EXPANDED;
       }
 
       if (nodeIsLongString(item)) {
--- a/packages/devtools-reps/src/object-inspector/tests/component/__snapshots__/expand.js.snap
+++ b/packages/devtools-reps/src/object-inspector/tests/component/__snapshots__/expand.js.snap
@@ -32,7 +32,7 @@ exports[`ObjectInspector - state does not throw when expanding a block node 2`] 
 
 exports[`ObjectInspector - state expanding a getter returning a longString does not throw 1`] = `
 "
-▼ {…}
+▼ Object
 |  ▼ baseVal: \\"data:image/png;base64,initial<<<<\\"
 ▶︎ Proxy { <target>: {…}, <handler>: (3) […] }
 "
@@ -47,7 +47,7 @@ exports[`ObjectInspector - state expands if user selected some text and clicked 
 
 exports[`ObjectInspector - state expands if user selected some text and clicked the arrow 2`] = `
 "
-[ ▼ {…} ]
+[ ▼ Object ]
   |    a: 1
   |    Symbol(): \\"hello\\"
   |  ▶︎ <prototype>: Object { … }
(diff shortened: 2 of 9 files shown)


packages/devtools-reps/src/reps/grip-map.js
```diff
@@ -49,6 +49,9 @@ function GripMap(props) {
   }
 
   const propsArray = safeEntriesIterator(props, object, maxLengthMap.get(mode));
+  if (mode == MODE.EXPANDED) {
```
nchevobbe: I think we could add a condition on the if statement of line 48, since we return the same thing

packages/devtools-reps/src/reps/grip.js
```diff
@@ -78,6 +78,10 @@ function GripRep(props) {
 
   const propsArray = safePropIterator(props, object, maxLengthMap.get(mode));
 
+  if (mode === MODE.EXPANDED) {
```
nchevobbe: could we move this before building `propsArray` ? `safePropIterator` can be a bit costly, and here we don't use it when in expanded mode.

packages/devtools-reps/src/reps/grip-map.js
marcogallo566: Ops! I'll fix it now.

packages/devtools-reps/src/reps/grip.js
marcogallo566: Right observation! I think it's possible, now I move it.

nchevobbe:
Okay, this is looking quite good !
I only have one comment I think.
Things are a bit weird when the Rep is originally in TINY mode and then get expanded, i.e. : 

```
▶︎ {…}
```

becomes

```
▼ Object
| …
```

which is a bit weird (same for `▶︎ (2) […]` and `▼ Array (2)`.
I think it would be better to not change the mode at all when the node is already in `TINY`

packages/devtools-reps/src/object-inspector/component.js
```diff
@@ -286,7 +286,7 @@ class ObjectInspector extends Component<Props> {
         repProps.mode = this.props.mode === MODE.LONG ? MODE.SHORT : MODE.TINY;
       }
       if (expanded) {
```
nchevobbe:
```suggestion
      if (expanded && this.props.mode !== MODE.TINY) {
```

I think this would be better

marcogallo566: If I change the code as you suggested, nothing change. I printed the `MODE` and I found that they are in `MODE.LONG`. I don't understand why.

nchevobbe: could you paste a diff of what you have?

marcogallo566: I've resolved by myself (I think), look at the last commit if that's what you mean.

packages/devtools-reps/src/object-inspector/tests/component/__snapshots__/expand.js.snap
```diff
@@ -32,7 +32,7 @@ exports[`ObjectInspector - state does not throw when expanding a block node 2`]
 
 exports[`ObjectInspector - state expanding a getter returning a longString does not throw 1`] = `
 "
-▼ Object
+▼ {…}
```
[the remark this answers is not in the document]
marcogallo566: I'm a bit confused, do you want something like this `▼ Object {...}`? Because I've restored the TINY mode in this case.

packages/devtools-reps/src/object-inspector/component.js
```diff
@@ -285,8 +285,13 @@ class ObjectInspector extends Component<Props> {
       if (depth > 0) {
         repProps.mode = this.props.mode === MODE.LONG ? MODE.SHORT : MODE.TINY;
       }
+
       if (expanded) {
-        repProps.mode = MODE.EXPANDED;
+        if (this.props.autoExpandDepth === 0) {
```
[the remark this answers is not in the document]
marcogallo566: When a Rep is in the format of `{...}` or `[...]` has autoExpandDepth == 0, when the Rep look like `foo {...}` has autoExpandDepth != 0. It's a check to understand if is better to use TINY or EXPANDED mode.

darkwing:
@nchevobbe Did you want to push this across the finish line or shall we close?

nchevobbe:
@marcogallo566 I'm really sorry for letting this slips. 
I'm in the middle of a consequent refactoring of the ObjectInspector and the patch probably won't apply cleanly.
Let's close this for now, and once I land my refactor, re-open a new one to fix this issue.
I'll do my best to help you finish this soon!

Issue #7028, opened 2018-09-25
[Breakpoints] Column Breakpoints


jasonLaster:
### Feature

Column breakpoints lets users set breakpoints inside of a line. We added this feature behind a flag, which can be flipped in the launchpad with `dbg.features.columnBreakpoints` or in firefox at about:config `devtools.debugger.features.column-breakpoints`

### Tasks
- [ ] holding down option should only show possible bps for the function you're hovering over
- [x] member expressions should be shown e.g a.b().c()  (b and c should now be shown) in addition to a(b(c())...
- [ ] switch UI from highlight to markers

####  holding down option should only show possible bps for the function you're hovering over

why: adding markers for every location in a file will be too slow. 
how: 

- we have a list of functions and their start and end locations for each file
- codemirror knows how to find the current line you're mousing over
- we can get the current hovered function and the column breakpoints for that function. 

https://github.com/devtools-html/debugger.html/blob/master/src/components/Editor/CallSites.js#L1

Also see highlightLines or copy function body

#### member expressions should be shown e.g a.b().c()  (b and c should now be shown) in addition to a(b(c())...

when we wrote the feature initially the engine did not support breakpoints at member properties. Now it does!

```diff
diff --git a/src/components/Editor/CallSites.js b/src/components/Editor/CallSites.js
index ad4f12c0..ebbd8676 100644
--- a/src/components/Editor/CallSites.js
+++ b/src/components/Editor/CallSites.js
@@ -174,6 +174,11 @@ function getCallSites(symbols, breakpoints) {
     return;
   }

+  // I think we can also get memberExpressions here too. It's not pretty, but would work...
+
+  // a possibly more reliable technique would be to start using the pause points
+  // data structure which should have all the locations we could set a breakpoint at!
+
   const callSites = symbols.callExpressions;

   // NOTE: we create a breakpoint map keyed on location
```

darkwing:
How Chrome does it:

<img width="825" alt="columnbps" src="https://user-images.githubusercontent.com/46655/46158287-2a1de600-c243-11e8-96ec-e7e4b251e26c.png">

How it works now:

![columnbp](https://user-images.githubusercontent.com/46655/46158301-3013c700-c243-11e8-9a28-cf5f1041a77d.gif)

PR #7064, opened 2018-09-28
extend callExpressions to include a.b().c()

AnshulMalik:
Fixes part of #7028

call


jasonLaster:
Nice!

So two ideas: 

1. could we use the memberExpressions we already save
2. better yet, could we switch to using pause points?

PR #7079, opened 2018-10-01
Change jump to line shortcut from Cmd+; to Ctrl+G

### LINKED ISSUE #6946: same keyboard shortcut (⌘;) is used for two functions: "step in" and "go to line"
From https://bugzilla.mozilla.org/show_bug.cgi?id=1488662

same keyboard shortcut (⌘;) is used for two functions: "step in" and "go to line".

steps to reproduce:
1. set a breakpoint and hit the breakpoint with the debugger
2. press ⌘; to step into a function

expected:
- step-in action performed
- nothing else

actual:
- step-in action performed
- the "search sources" popup is shown, with ":" populated as the query string; ie. the "go to line" feature

note the ⌘; and ⌘' shortcuts are quite useful on macos, where F11 is mapped to "show desktop" by default.

gregtatum:
Fixes #6946

### Summary of Changes

The Cmd+; shortcut conflicts with the step-in shortcut, which is
especially needed on newer Macbooks, as the function keys are off by
default with the new touchbar.

The shortcut was replaced to use the same as Atom Editor, and other
editors, with Ctrl+G.

### Test Plan

This shortcut was covered by a snapshot test, and I performed manual testing.

### DIFF
--- a/assets/panel/debugger.properties
+++ b/assets/panel/debugger.properties
@@ -796,11 +796,11 @@ gotoLineModal.placeholder=Go to line…
 # to open the go to line modal
 gotoLineModal.title=Go to a line number in a file
 
-# LOCALIZATION NOTE(gotoLineModal.key2): The shortcut for opening the
+# LOCALIZATION NOTE(gotoLineModal.key3): The shortcut for opening the
 # go to line modal
-# Do not localize "CmdOrCtrl+;", or change the format of the string. These are
+# Do not localize "Ctrl+G", or change the format of the string. These are
 # key identifiers, not messages displayed to the user.
-gotoLineModal.key2=CmdOrCtrl+;
+gotoLineModal.key3=Ctrl+G
 
 # LOCALIZATION NOTE(symbolSearch.search.functionsPlaceholder): The placeholder
 # text displayed when the user searches for functions in a file
--- a/src/components/App.js
+++ b/src/components/App.js
@@ -111,7 +111,7 @@ class App extends Component<Props, State> {
     ];
     searchKeys.forEach(key => shortcuts.on(key, this.toggleQuickOpenModal));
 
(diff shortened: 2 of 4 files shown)


assets/panel/debugger.properties
```diff
@@ -798,9 +798,9 @@ gotoLineModal.title=Go to a line number in a file
 
 # LOCALIZATION NOTE(gotoLineModal.key2): The shortcut for opening the
```
darkwing: I can't jump into this right away but note to self or @gregtatum ; I think we need to change `key2` to `key3` here.

gregtatum: Thanks for catching that, fixed! For some reason my editor is not wanting to search through the assets folder.

assets/panel/debugger.properties
```diff
@@ -796,11 +796,11 @@ gotoLineModal.placeholder=Go to line…
...
 gotoLineModal.title=Go to a line number in a file
 
-# LOCALIZATION NOTE(gotoLineModal.key2): The shortcut for opening the
+# LOCALIZATION NOTE(gotoLineModal.key3): The shortcut for opening the
 # go to line modal
-# Do not localize "CmdOrCtrl+;", or change the format of the string. These are
+# Do not localize "Ctrl+G", or change the format of the string. These are
 # key identifiers, not messages displayed to the user.
-gotoLineModal.key2=CmdOrCtrl+;
+gotoLineModal.key3=Ctrl+G
```
jasonLaster: hmm, cmd+g is another shortcut, should we limit it to cmd+g?

does case matter ctrl+g or ctrl+G?

darkwing: Any thoughts here @gregtatum

gregtatum:
I was out on PTO, was this closed due to inactivity?

darkwing:
Yep -- feel free to open again with updates :)

PR #7082, opened 2018-10-02
Add tabs for sources without URLs WIP

### LINKED ISSUE #5102: [Tabs] Sources without URLs should have a tab

Doomwyte:
Fixes #5102 (WIP)

### Summary of Changes

* Add tabs for URL-less sources
* Store URL-less tabs in redux but not in prefs

### Test Plan

- [x] Typing `debugger;` in Console brings up a URL-less source in Debugger, make sure that this shows a tab

### Screenshots
Before (no tab)
screen shot 2018-10-01 at 11 02 11 pm

After (yes tab)
screen shot 2018-10-01 at 11 02 27 pm

### DIFF
--- a/src/reducers/sources.js
+++ b/src/reducers/sources.js
@@ -374,6 +374,22 @@ export function getSourceByUrlInSources(
   return foundSources.find(source => isOriginalSource(source) == isOriginal);
 }
 
+export function getSourcesBySourceId(
+  sources: SourcesMap,
+  sourceId: string,
+  isOriginal: boolean
+) {
+  const foundSource = sources[sourceId];
+  if (!foundSource) {
+    return null;
+  }
+  if (isOriginalSource(foundSource) == isOriginal) {
+    return foundSource;
+  } else {
+    return null;
+  }
+}
+
 function getSourcesByUrlInSources(
   sources: SourcesMap,
   urls: UrlsMap,
--- a/src/reducers/tabs.js
+++ b/src/reducers/tabs.js
@@ -19,7 +19,8 @@ import {
   getSources,
   getUrls,
   getSourceByURL,
-  getSourceByUrlInSources
+  getSourceByUrlInSources,
+  getSourcesBySourceId
 } from "./sources";
 
 import type { Action } from "../actions/types";
@@ -75,20 +76,29 @@ export function removeSourcesFromTabList(tabs: TabList, sources: Source[]) {
  */
 function updateTabList(
   tabs: TabList,
-  { url, framework = null, isOriginal = false }
+  { url, framework = null, sourceId, isOriginal = false }
 ) {
   const currentIndex = tabs.findIndex(tab =>
     isSimilarTab(tab, url, isOriginal)
   );
 
   if (currentIndex === -1) {
-    tabs = [{ url, framework, isOriginal }, ...tabs];
+    tabs = [{ url, framework, sourceId, isOriginal }, ...tabs];
   } else if (framework) {
     tabs[currentIndex].framework = framework;
   }
 
-  asyncStore.tabs = tabs;
+  asyncStore.tabs = persistTabs(tabs);
   return tabs;
+}
+
+function persistTabs(tabs) {
+  return tabs.filter(tab => tab.url).map(tab => {
+    const newTab = { ...tab };
+    delete newTab.sourceId;
+    console.log(newTab);
+    return newTab;
+  });
 }
 
 function moveTabInList(tabs: TabList, { url, tabIndex: newIndex }) {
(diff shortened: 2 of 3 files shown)


src/reducers/sources.js
```diff
@@ -374,6 +374,22 @@ export function getSourceByUrlInSources(
   return foundSources.find(source => isOriginalSource(source) == isOriginal);
 }
 
+export function getSourcesBySourceId(
+  sources: SourcesMap,
+  sourceId: string,
+  isOriginal: boolean
+) {
+  const foundSource = sources[sourceId];
```
darkwing: The logic here could be shortened to something like:

```js
if (foundSource && isOriginalSource(foundSource) === isOriginal) {
  return foundSource;
}

return null;
```

src/reducers/tabs.js
```diff
@@ -75,22 +76,31 @@ export function removeSourcesFromTabList(tabs: TabList, sources: Source[]) {
...
-  asyncStore.tabs = tabs;
+  asyncStore.tabs = persistTabs(tabs);
   return tabs;
 }
 
+function persistTabs(tabs) {
+  return tabs.filter(tab => tab.url).map(tab => {
+    const newTab = { ...tab };
+    delete newTab.sourceId;
+    console.log(newTab);
```
darkwing: nit: remove console.log if you no longer need it.

src/reducers/tabs.js
```diff
@@ -184,9 +194,13 @@ export const getSourceTabs = createSelector(
   getSources,
   getUrls,
   (tabs, sources, urls) =>
-    tabs.filter(tab =>
-      getSourceByUrlInSources(sources, urls, tab.url, tab.isOriginal)
-    )
+    tabs.filter(tab => {
```
darkwing: The `tab => { }` function looks the same so maybe create a function to reuse the logic in both spots?

```js
.filter(getTabWithOrWithoutUrl(tab, sources, url))
```

Something like that.

darkwing:
As for a test here, I think you could add another test to `browser_dbg-tabs.js`.  There are a few there which show basic tab interaction.  I'm looking at `browser_dbg-breaking-from-console.js` and see the following:

```js
const toolbox = await openNewTabAndToolbox(url, "webconsole");

  // Type "debugger" into console
  let jsterm = toolbox.getPanel("webconsole").hud.jsterm;
  jsterm.execute("debugger");

  // Wait for the debugger to be selected and make sure it's paused
  await waitOnToolbox(toolbox, "jsdebugger-selected");
```

That looks relevant for executing a `debugger`, so you can probably copy that over; from there you can (1) ensure the tab count went up 1 and (2) that the tab has the correct name 👍

Doomwyte:
Hi @darkwing ! For the refreshing page and the source-with-url tab not getting focused, what should the expected behaviour be? 

I was thinking originally that maybe the wp-embeded.min.js tab should be focused, since it is the only available tab. But I was thinking, if there are multiple source-with-url tabs, which one should I focus on? I was thinking maybe focus on the first tab, but then it made me think...because originally we were on the source-without-url tab and then it disappears when we refresh the page, so we focus on the first tab, which had nothing to do with the source-without-url tab. Would that change the behaviour of which tab to select as the active one?

darkwing:
@Doomwyte I love that you're not only focused on completing the task but also thinking about UX -- that's awesome.  To keep it simple, I think the easiest route is focusing on the first tab that displays in the tab list.

Doomwyte:
@darkwing Hi! To display the first tab in the tab list where I refresh the page when I have a URL-less tab focused, I think something like this might do it:
```
export const getSelectedSource = createSelector(
  getSelectedLocation,
  getSources,
  getTabs,
  (selectedLocation: ?Location, sources: SourcesMap): ?Source => {
    if (!selectedLocation) {
      return getSourceByURL(tabs.first.url);
    }

    return sources[selectedLocation.sourceId];
  }
);
```
But this means that in the case where without URL-less tabs, I have some URL tabs open but none are focused. When I refresh the page, it will select the first tab, even though I did not have any selected previously. Is that ok? 

An alternative would be to set selectedLocation for URL-less tabs, maybe something like `selectedLocation: {urlLess: true}` but I think that will require more code changes.

darkwing:
Do you know the "root" cause of why that tab isn't focused?  Is there a way to only add urlLess to selectLocation upon initial load?

Doomwyte:
The root cause is because when I refresh the page, the debugger tries add a list of known sources and tries to check if the source it's trying to add matches with the one that was selected. But because the URL-less tab does not have a real source, it goes through the list of sources but does not find the source of the URL-less tab I was focused on.

darkwing:
Oh, yep, right.  So, would it be possible to:

1.  Before tabs are added to the UI, determine if that source exists/has a URL, then...
2.  If not, focus on the first tab?

darkwing:
Superceded by https://github.com/devtools-html/debugger.html/pull/7102

PR #7083, opened 2018-10-02
Function signatures

### LINKED ISSUE #6580: [Breakpoints] Breakpoint changes its place on page refresh
### STR
```javascript
a = {
    sameName: () => {
        return 1; /* After refresh breakpoint is now set on this line instead. */
    }
}
b = {
    sameName: () => {
        return 42; /* Set breakpoint here and refresh.*/
    }
}
```
1. Set breakpoint on line with 'return 42'
2. Refresh page

Actual results:

Breakpoint is now set on line with 'return 1'.

Expected results:

Breakpoint should stay where it was set.

This happens when two or more methods have the same name even when they are part of different objects. 
If I had to guess debugger stores breakpoint line numbers as an offset from the line where method is defined. Then on refresh it finds the first time this method's name appears and sets the breakpoint with that offset there disregarding that there could be more than one method in the code with such name.

bomsy:
Fixes #6580 

### Summary of Changes

* Add a `signature` property to uniquely identify the function scope 

The `signature` is a new property added to the function symbol, it is meant to uniquely identify the current function scope based on certain criteria. etc
```js
const a  = {
      b: function() {} 
} 
// signature => a.b
const c  = {
      b: function() {} 
} 
// signature => c.b

function b (){}
//signature => b
```

### Test Plan
- Load https://lumpy-plier.glitch.me/
- Open the debugger and create a breakpoint on line 11 return 42
- Refresh the page
- The breakpoint is now on line 6 -> BUG


jasonLaster:
i think this would benefit from some unit tests

bomsy:
Yep ... on the way!

darkwing:
@bomsy Outside of jest updates, what's the state of this PR?  Thanks!

bomsy:
So we've fixed the jest issue will finish this up today..

darkwing:
Wow @bomsy!  This looks amazing!  I'll dig into it today!

src/workers/parser/getSymbols.js
```diff
@@ -123,6 +124,57 @@ function getFunctionParameterNames(path: SimplePath): string[] {
...
+    return path.node.id.name;
+  }
+
+  if (!signature) {
+    signature = getFunctionName(path.node, path.parent);
+    path = path.parentPath && path.parentPath.parentPath;
+  }
+
+  if (t.isMemberExpression(path)) {
+    if (!path || !path.node.object.name) {
```
loganfsmyth: I think duck typing like this makes the code harder to follow. I'd prefer `!path || !t.isIdentifier(path.node.object)`.

src/workers/parser/getSymbols.js
```diff
@@ -123,6 +124,57 @@ function getFunctionParameterNames(path: SimplePath): string[] {
...
+      return signature;
+    }
+    return `${path.node.object.name}.${signature}`;
+  }
+
+  if (t.isAssignmentExpression(path)) {
+    if (!path || !path.node.left.name) {
```
loganfsmyth: Same here for `t.isIdentifier`

src/workers/parser/getSymbols.js
```diff
@@ -123,6 +124,57 @@ function getFunctionParameterNames(path: SimplePath): string[] {
...
+
+  if (path && t.isObjectProperty(path)) {
+    return getFunctionSignature(
+      path.parentPath,
+      `${path.node.key.name}.${signature}`
+    );
+  }
+
+  if (path && (t.isVariableDeclarator(path) || t.isClassDeclaration(path))) {
+    if (!path.node.id.name) {
```
loganfsmyth: And here.

src/workers/parser/getSymbols.js
```diff
@@ -123,6 +124,57 @@ function getFunctionParameterNames(path: SimplePath): string[] {
...
+  if (t.isFunctionDeclaration(path)) {
+    if (path.node.id == null) {
+      return "anonymous";
+    }
+    return path.node.id.name;
+  }
+
+  if (!signature) {
+    signature = getFunctionName(path.node, path.parent);
+    path = path.parentPath && path.parentPath.parentPath;
```
loganfsmyth: This feels worth a comment. Why jump up 2 paths?

src/workers/parser/getSymbols.js
```diff
@@ -123,6 +124,57 @@ function getFunctionParameterNames(path: SimplePath): string[] {
...
 
+function getFunctionSignature(
+  path: SimplePath | null,
+  signature: string | null
+) {
+  if (!path) {
+    return signature;
+  }
+
+  if (t.isFunctionDeclaration(path)) {
```
loganfsmyth: If I'm reading this right, this will mean that the signature for a function inside of a function declaration can be the declaration itself? I might be missing an edge case. e.g.

````
function someThing(callback) {
  callback(() => {

  });
}
```
is `someThing` the expected signature for the arrow function?

src/workers/parser/getSymbols.js
bomsy: nice didn't think about that .. thansk!

src/utils/breakpoint/astBreakpointLocation.js
```diff
@@ -27,15 +27,22 @@ export function getASTLocation(
...
+  source: Source,
+  name: ?string,
+  signature: ?string
+) {
   const symbols = await getSymbols(source.id);
   const functions = symbols.functions;
 
-  return functions.find(node => node.name === name);
+  return functions.find(
+    node => node.name === name && node.signature === signature
```
bomsy: @loganfsmyth In answer to your question on bail out of `signatures`, Yes! if the `signature` is not unique, then the result of the check of `name` and `signature` would match that of checking just the `name`.

jasonLaster:
Great work so far. This option seems like it will do the trick. 

Out of curiosity, how do you feel it compares with the other option of keeping the function's index? I feel like most functions will appear only once in a file, but if they appear multiple times then the index (0, 1, 2, ...) would help us track it as well.

I'm happy to go with either option, but just wanted to see what you're thinking.

bomsy:
Thanks! Yeah. The solution with the function index is much easier as compared to this.
This solution is quite involved (sometimes i think if this solution is an overkill).
- The reason i opted for it though was that  the signature is more consistently unique, than when using indexes .i.e if the functions are moved around or removed, the breakpoints added would get messed up.
If we can live with that, i don't mind us going with that.

jasonLaster:
I think i'm okay with a simpler solution at the moment... It is kinda tough to maintain all these edge cases. I suppose we can always revisit it if we need too...

PR #7084, opened 2018-10-02
6864/keep item focused on tree update

### LINKED ISSUE #6864: [sources] source tree loses focuses when new items are added
STR:

1. go to https://www.cnn.com/
2. reload
3. click on items in the source tree

ER: it stays focused
AR: it loses its focus

WenInCode:
Fixes #6864 

**WIP**
The issue is that the source tree looses focus when new sources are added to it. One of the reasons that's happening is because we return a `null` `focusedItem` from the [updateTree](https://github.com/devtools-html/debugger.html/blob/master/src/utils/sources-tree/updateTree.js#L55) call. However, even if we were to return the current `focusedItem` in that call it would not stay focused because we build a _new tree_ each time we call `updateTree` so this [equality check](https://github.com/devtools-html/debugger.html/blob/master/packages/devtools-components/src/tree.js#L778) will fail as the old focused item is no longer referencing the same item in the tree.

### Summary of Changes
* Search `newSourceTree` for `focusedItem` and set it in the `SourcesTree` state.

### Current Issues
I left some debug logging in the function that searches for the `focusedItem`. It looks like the `focusedItem` is consistently found, but only passes the [equality check](https://github.com/devtools-html/debugger.html/blob/master/packages/devtools-components/src/tree.js#L778) if it is a `source` and not a directory. The result of that is that `sources` remain focused as the tree updates, but `directories` do not. I am a little stuck on that, as it seems like the `focusedItem` is always found in the tree.

Right now I'm focused on getting the code working, before I focus on making the code better. I'd love some help with the above issue 😄 

#### STR
* Open debugger
* Set network throttling (optional)
* Open cnn.com 
* click on a source (it remains focused)
* click on a directory (loses focus as the sources are added)

### Test Plan
**Test Plan has not been created yet**

### DIFF
--- a/src/utils/sources-tree/updateTree.js
+++ b/src/utils/sources-tree/updateTree.js
@@ -34,7 +34,12 @@ function findFocusedItemInTree(
     const parts = focusedItem.path.split("/").filter(p => p !== "");
     let path = "";
 
+    console.log("------ START OF TREE SEARCH ------");
     focusedItem = parts.reduce((subTree, part, index) => {
+      console.log("SUB TREE:", subTree);
+      console.log("PART:", part);
+      console.log("PARTS:", parts);
+      console.log("IS DIRECTORY:", isPartDir(focusedItem, parts.length, index));
       path = path ? `${path}/${part}` : part;
       const { index: childIndex } = findNodeInContents(
         subTree,
@@ -44,8 +49,11 @@ function findFocusedItemInTree(
           debuggeeHost
         )
       );
+      console.log("------ END OF LOOP ------");
       return subTree.contents[childIndex];
     }, newSourceTree);
+    console.log("------ END OF TREE SEARCH ------");
+    console.log("FOCUSED ITEM:", focusedItem);
   }
 
   return focusedItem;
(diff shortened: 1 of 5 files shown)


jasonLaster:
CC @AnshulMalik what do you think?

src/utils/sources-tree/updateTree.js
```diff
@@ -22,13 +25,55 @@ function newSourcesSet(newSources, prevSources) {
...
+  }
+
+  return focusedItem;
+}
+
+function isPartDir(focusedItem: ?TreeNode, partsLength, index) {
+  if (focusedItem && isDirectory(focusedItem)) {
+    return true;
+  }
+  return partsLength - 1 != index;
```
WenInCode: I'll probably leverage the actual [partIsFile helper](https://github.com/devtools-html/debugger.html/blob/master/src/utils/sources-tree/utils.js#L75) which looks like it's doing a similar thing. However, `url` will only be available if `focusedItem` is a source _I think_.

WenInCode:
Updated this a bit. I figured out why the `subTree` was occasionally coming through as `null`. The directory losing focus issue remains. After the discussion from the community call last week, it sounds like we may want to move forward with this even if the directory issue remains. If anyone has any time to check it out that would be great though, a little stumped on that. Other than that I added a few asserts to an existing mochitest before/after a source is added to the tree.

PR #7088, opened 2018-10-02
[outline] Enable filtering the outline list

### LINKED ISSUE #7054: [Outline] should have a search box
It would be really nice if our outline pane let you filter the results.

- this would be similar in practice to Quick Open and likely use the same fuzz-aldrin filter
- this is similar to how atom has an outline view with a filter input

screen shot 2018-09-27 at 10 15 51 am

kenjiO:
Fixes #7054

### Summary of Changes
* Add react component for an input box to filter the outline items
* Add code to the Outline component that filters based on the input in the new component

### Test Plan
- [x] manual testing
- [x] automated tests
 

screen shot 2018-10-02 at 3 19 54 pm
screen shot 2018-10-02 at 3 20 11 pm


kenjiO:
Can you let me know how this looks so far?  I am still planning on writing some tests and maybe doing some small changes.

src/components/PrimaryPanes/Outline.js
```diff
@@ -41,7 +43,18 @@ type Props = {
...
-export class Outline extends Component<Props> {
+type State = {
+  filter: string
+};
+
+export class Outline extends Component<Props, State> {
+  constructor(props: Props) {
+    super(props);
+    (this: any).filterOutlineItems = this.filterOutlineItems.bind(this);
+    (this: any).onUpdateFilter = this.onUpdateFilter.bind(this);
```
kenjiO: If we use the ES6 arrow function definitions for `filterOutlineItems` and `onUpdateFilter` we wouldn't need to bind them here.  Do you want to do this?  If so do you want to change all the function definitions in the class to this style for consistancy?

assets/panel/debugger.properties
```diff
@@ -606,6 +606,10 @@ sources.header=Sources
 # LOCALIZATION NOTE (outline.header): Outline left sidebar header
 outline.header=Outline
 
+# LOCALIZATION NOTE (outline.placeholder): Placeholder text for the filter input
+# input element
```
flodolo: nit: you have two "input" in the comment

src/components/PrimaryPanes/OutlineFilter.js
```diff
@@ -0,0 +1,60 @@
...
+
+  onChange = (e: SyntheticInputEvent<HTMLElement>) => {
+    this.props.updateFilter(e.target.value);
+  };
+
+  onKeyDown = (e: SyntheticKeyboardEvent<HTMLInputElement>) => {
+    if (e.key === "Escape" && this.props.filter !== "") {
+      e.preventDefault();
+      this.props.updateFilter("");
+    }
```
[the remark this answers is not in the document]
AnshulMalik: That sounds right to me

AnshulMalik:
This is awesome :)

kenjiO:
@darkwing I've finished getting this to filter the list by function names.  At the call we discussed also filtering by class name but I don't know what the desired behavior for that would be.  For example if the filter matched a class name would you show all the functions in that class or just show the class name?

darkwing:
Wow, that's a great question.  Let's focus on functions only for now, with the caveat that if the function was a class method, we should show the class name above it, IMO.  Thoughts?

kenjiO:
@darkwing I agree at keeping it this way now.  Usually there are not that many classes defined in a file so I don't think being able to filter them will be that useful.  Currently if any class functions match the filter it will show all the class names with the matching functions underneath their associated class.   I don't have any other pending changes for this PR so you can merge it if you don't need any further changes.

PR #7091, opened 2018-10-05
Revert "[editor] add check in HighlightLine to fix unwanted highlighting"

darkwing:
Reverts devtools-html/debugger.html#6916

This provided a false positive travis run and breaks two tests


loganfsmyth:
Looks like this accidentally included a commit that is empty and we aren't actually reverting.

PR #7094, opened 2018-10-07
Issue 7035: [Search] Bit of polish for new Find in File UI

### LINKED ISSUE #7035: [Search] Bit of polish for new Find in File UI
To follow up on the great work by @audreyfang in #7010:

This part which I hadn't considered in my mockup looks a bit busy. 
image

I assume we do want to show both the navigational buttons and the modifiers at the same time when in this state, in case the user wants to change a modifier while the search in progress. 

If so, I think adding a vertical line separator line between the navigation and modifiers should help with this, similar to how this image looks (see main DevTools toolbar, Netmonitor toolbar, for examples of separator code)
image 

Spacing: There needs to be less spacing between the next/prev arrows to give them more of a sense of being related (could be about the same amount of spacing as the modifier buttons, if that seems to work fine for easy clickability.)

Re: Close button: I think it needs to always be the right most button, with a separator line between it and other buttons.

In light mode, search bars should have a white background.

Thanks for working on this!

adancarrasco:
Fixes #7035 

Here's the Pull Request Doc

### Summary of Changes

* Move Close Button implementation from `SearchInput.js` to `SearchBar.js`.

### Test Plan
Open a file in the debugger, then find for something, you'll see the close button is now in the desired place.

Example test plan:

- [x] Search for a file or open it
- [x] Cmd/Ctrl + F

Feel free to improve it!

Theme Light:
image
Theme Dark:
image


src/components/Editor/SearchBar.css
```diff
@@ -47,7 +47,7 @@
   flex-shrink: 0;
   justify-content: flex-end;
   background-color: var(--theme-toolbar-background);
-  padding: 0 13px;
+  padding: 0 1px 0 13px;
```
darkwing: We should use `padding-inline-start` and `padding-inline-end` here so the design is consistent in RTL.

src/components/Editor/SearchBar.css
```diff
@@ -65,6 +65,22 @@
...
+.search-bottom-bar .search-modifiers button.close-btn.big img {
+  width: 12px;
+  height: 12px;
+}
+
+.search-bottom-bar .search-modifiers .pipe-divider {
+  height: 70%;
+  border: 1px solid #e0e0e1;
+  vertical-align: middle;
+  margin: 0 0.3em 0 0.8em;
```
darkwing: We should use `margin-inline-start` and `margin-inline-end` for RTL.

src/components/Editor/SearchBar.js
```diff
@@ -297,6 +306,12 @@ class SearchBar extends Component<Props, State> {
           svgName="whole-word-match"
           tooltip={L10N.getStr("symbolSearch.searchModifier.wholeWord")}
         />
+        {showClose && (
+          <React.Fragment>
+            <span className={"pipe-divider"} />
```
darkwing: The `{` and `}` aren't needed here.

darkwing:
Wow, this is awesome @adancarrasco !  Quick thing I noticed is the hover color of the close button should be dark:

hover

Also, I think you'll need to run `yarn jest -u` to update tests.

Great work so far! 👍

src/components/Editor/SearchBar.css
```diff
@@ -65,6 +65,22 @@
...
 }
 
+.search-bottom-bar .search-modifiers button.close-btn.big img {
+  width: 12px;
+  height: 12px;
+}
+
+.search-bottom-bar .search-modifiers .pipe-divider {
+  height: 70%;
+  border: 1px solid #e0e0e1;
```
darkwing: We should probably use an existing variable color instead of explicitly setting them.

src/components/Editor/SearchBar.css
adancarrasco: Good catch! I'll update it.

audreyfang:
The close button on the SearchInput is also used in Find in Project (Ctrl Shift P) so moving the close button from SearchInput to SearchBar removes the close button from Find in Project. Find in Project by can still be closed moving to another tab, but we do lose the explicit close button.

darkwing:
Hello again @adancarrasco !  This one's almost there!  Do you know when you may have time to pull it across the finish line?  Keep up the great work!

adancarrasco:
@audreyfang Thanks for the feedback. I will check it. :)
@darkwing Tomorrow I will dedicate some time to update this PR and hopefully get it merged. 😎

PR #7095, opened 2018-10-08
Column breakpoints UI

audreyfang:
### Summary of Changes

Very rough start to the column breakpoint changes @darkwing and I discussed earlier. When alt/option is pressed column breakpoints are shown only for the function where the cursor is currently located. 

This is still quite buggy but works in some cases and I will also be further looking into changing the >> to actual icons.

### Screenshots/Videos

column-breakpoint


src/components/Editor/CallSites.js
```diff
@@ -144,15 +145,34 @@ class CallSites extends Component {
...
-    const { editor, callSites, selectedSource } = this.props;
+    const { editor, callSites, selectedSource, symbols } = this.props;
     const { showCallSites } = this.state;
+
     let sites;
     if (!callSites) {
       return null;
     }
 
+    const { line } = editor.codeMirror.getCursor();
```
AnshulMalik: I think `getCursor` gives us the position of the blinking cursor, and not mouse position, I am not sure though.

darkwing:
Interesting that you've gone the CSS route instead of adding an element as a marker -- I look forward to seeing where this goes.  The important test for it is ensuring that clicking in that area of the marker does in fact activate/deactivate the column breakpoint -- CSS can be weird like that sometimes.

src/components/Editor/CallSite.css
```diff
@@ -4,19 +4,20 @@
...
-  position: absolute;
+  content: ">>";
   width: 100%;
   height: calc(100% - 2px);
   border-bottom: 2px solid #aed3ef;
 }
 
 .call-site-bp {
   position: relative;
+  border-bottom: 2px solid red;
```
darkwing: I'm guessing this is for debugger?

src/components/Editor/CallSite.css
```diff
@@ -25,8 +26,7 @@
 }
 
 .call-site-bp::before {
-  content: "";
-  position: absolute;
+  content: ">>";
```
darkwing: You should be able to use the same breakpoint svg image as the sidebar.

src/components/Editor/CallSites.js
```diff
@@ -144,15 +145,34 @@ class CallSites extends Component {
...
+
+    let callSitesFiltered;
+    if (closestFunc !== null) {
+      callSitesFiltered = callSites.filter(
+        ({ location }) =>
+          location.start.line >= closestFunc.location.start.line &&
+          location.start.line <= closestFunc.location.end.line
+      );
+    } else {
+      callSitesFiltered = callSites;
```
darkwing: You don't need the `else`; just initialize `callSitesFiltered` to `callSites`:

```js
let callSitesFiltered = callSites;
if ( ....
```

src/components/Editor/CallSites.js
audreyfang: It does give the position of the blinking cursor, which was my intention for this first iteration. Do you happen to know of any way to get the mouse hover position?

src/components/Editor/CallSite.css
audreyfang: Not completely sure I understand the question, but the red border is the current way of indicating that a breakpoint is selected. I added it here so that both the icon and text continue to be underlined.

darkwing: Ahh, I see.  We probably want to find a different color, but that's a simple change so don't hang up on it.

src/components/Editor/CallSites.js
AnshulMalik: I did read the docs but couldn't find the mouse position, I think we need to come up with our own way to capture mouse position.

darkwing: Let's table the mouse position experiment for now.  Let's just focus on clicking the icon adding the column breakpoint, and then having different states like enabled, disabled, off, as well as dark mode 💯

darkwing:
Update as to what's left here @audreyfang ?

audreyfang:
@darkwing I need to make a few minor fixes to the ui colours, look at your suggestions from last week and the failing tests. I should have another commit up later tonight.

audreyfang:
@darkwing Do you have a specific example for when clicking on a breakpoint doesn't remove it? It works for me in most cases, although I do seem to have some issues with one particular function in the file I'm using.

assets/module-manifest.json
```diff
@@ -20574,7 +20574,8 @@
       "../../parse5/lib/serializer/serializer_stream.js": 1752,
```
darkwing: This doesn't belong in this PR, we'll remove it before merging.

assets/panel/prefs.js
```diff
@@ -50,7 +50,7 @@ pref("devtools.debugger.skip-pausing", false);
 pref("devtools.debugger.features.wasm", true);
 pref("devtools.debugger.features.shortcuts", true);
 pref("devtools.debugger.features.root", true);
-pref("devtools.debugger.features.column-breakpoints", false);
+pref("devtools.debugger.features.column-breakpoints", true);
```
darkwing: 👍

darkwing: Actually, as things stand, we'll want to change this back to `false` before merging.  I just pressed `option` in a minified file and my browser hung, so we'll need to measure performance implications here.  Keep it true for now, if only for the sake of everyone testing and tests running.

darkwing:
OMG I love how this looks so far.  Awesome awesome awesome!

I'm seeing a really weird issue and I'm wondering if @loganfsmyth or @AnshulMalik could help.  I prettified the following URL during testing:

https://davidwalsh.name/wp-content/themes/punky/js/main.js

I press option and see a bunch of lovely column breakpoint locations (yay!).  When I click one, however, the following issues occur:

1.  A breakpoint is added at the *next line*
2.  The column's arrow very briefly stays dark (yay!) but then goes back to inactive state (the root cause is probably the same as 1)
3.  I can't add multiple column breakpoints for the same line; instead the existing breakpoint's location is changed.

@loganfsmyth Can you look into the breakpoint location issue?  It's acting a bit odd.

@audreyfang In the mean time, could you explore adding mochitest for this functionality?  i.e. clicking a column breakpoint on and off does what's expected?  I expect you'll probably have questions here which we can discuss in Slack or 1:1 video. 

Great work so far!

AnshulMalik:
It's the issue with source-map, either the source map generated from prettifying is not right(I think this is the case). Or `source-map` package is producing wrong results for generatedLocation.

audreyfang:
@darkwing Will the column breakpoints remain enabled by default? If so, I need to update the jest tests to reflect that. If not, then this should be fine to merge.

darkwing:
@audreyfang I think you'll only need to change the pref in `assets/panel/prefs.js`; I think the other prefs file only controls the development environment (launchpad) preference value.  Also, please reset the `assets/module-manifest.json` file.  I think we can merge after that 👍

Issue #7085, opened 2018-10-02
Upgrade greenkeeper


jasonLaster:
Greenkeeper's job is to update our dependencies so that we are always prompted when something changes. At the moment, we have a forked version which attempts to make a new commit with the yarn.lock, but it fails at the moment.

We should try to:
a) get back on current [greenkeeper](https://greenkeeper.io/) which is better or 
b) switch to [renovate](https://renovatebot.com/)

---
👍  If you'd like to help and are not an admin, ping me and i can promote you temporarily :)

AnshulMalik:
We need to set these environment variables in circles 
- GK_LOCK_COMMIT_NAME
   
- GK_LOCK_COMMIT_EMAIL

darkwing:
I've added these environment vars; let's hope things work properly.

AnshulMalik:
We have renovate now :)

PR #7096, opened 2018-10-09
use latest greenkeeper

AnshulMalik:
Let's try using the official greenkeeper

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,7 +79,7 @@
     "react-inlinesvg": "^0.8.1",
     "react-redux": "^5.0.7",
     "react-transition-group": "^2.2.1",
-    "redux-saga": "^0.16.0",
+    "redux-saga": "^0.16.1",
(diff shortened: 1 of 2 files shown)


darkwing:
Maybe I missed it but what's the issue with current greenkeeper?

WenInCode:
@darkwing https://github.com/devtools-html/debugger.html/issues/7085
according to this issue, the fork seems to sometimes fail to update the `yarn.lock`.

package.json
```diff
@@ -79,7 +79,7 @@
     "react-inlinesvg": "^0.8.1",
     "react-redux": "^5.0.7",
     "react-transition-group": "^2.2.1",
-    "redux-saga": "^0.16.0",
+    "redux-saga": "^0.16.1",
```
kenjiO: Does the debugger use redux-saga anywhere?

PR #7102, opened 2018-10-11
Add tabs for sources without URLs

### LINKED ISSUE #5102: [Tabs] Sources without URLs should have a tab

Doomwyte:
Fixes #5102

Continuation of #7082 

### Summary of Changes

* Add tabs for URL-less sources
* Store URL-less tabs in redux but not in prefs

### To do:
* When having a URL-less tab open and focused, when refreshing the page, the URL-less tab doesn't get focused
* Add mochi test

### DIFF
--- a/src/reducers/sources.js
+++ b/src/reducers/sources.js
@@ -92,7 +92,9 @@ function update(
         url: action.source.url
       };
 
-      prefs.pendingSelectedLocation = location;
+      if (action.source.url) {
+        prefs.pendingSelectedLocation = location;
+      }
 
       return {
         ...state,
--- a/src/reducers/sources.js
+++ b/src/reducers/sources.js
@@ -417,6 +417,22 @@ export function getSourceByUrlInSources(
   return foundSources[0];
 }
 
+export function getSourcesBySourceId(
+  sources: SourcesMap,
+  sourceId: string,
+  isOriginal: boolean
+) {
+  const foundSource = sources[sourceId];
+  if (!foundSource) {
+    return null;
+  }
+  if (isOriginalSource(foundSource) == isOriginal) {
+    return foundSource;
+  } else {
+    return null;
+  }
+}
+
 function getSourcesByUrlInSources(
   sources: SourcesMap,
   urls: UrlsMap,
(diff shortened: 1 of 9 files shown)


darkwing:
To Do:

[ ] - Disable the second set of items in the tab's context menu
[ ] - Test
[ ] - Figure out how to select the first tab after refreshing while having been on a URL-less tab.

darkwing:
One issue I found was that only one URL-less source would display at a given time but I figured out how to fix that 👍 
multipleurllesssources

src/reducers/sources.js
```diff
@@ -92,7 +92,9 @@ function update(
         url: action.source.url
       };
 
-      prefs.pendingSelectedLocation = location;
+      if (action.source.url) {
+        prefs.pendingSelectedLocation = location;
+      }
```
darkwing: @Doomwyte This change allows the editor to select the first available tab upon refresh; `prefs.pendingSelectedLocation` saves the last source (and its location) to localStorage so that it can be selected next time around.  If the source doesn't have a URL, however, we don't want that to be saved.

src/reducers/tabs.js
```diff
@@ -75,22 +80,32 @@ export function removeSourcesFromTabList(tabs: TabList, sources: Source[]) {
...
+  { url, framework = null, sourceId, isOriginal = false }
 ) {
-  const currentIndex = tabs.findIndex(tab =>
-    isSimilarTab(tab, url, isOriginal)
-  );
+  // Set currentIndex to -1 for URL-less tabs so that they aren't
+  // filtered by isSimilarTab
+  const currentIndex = url
+    ? tabs.findIndex(tab => isSimilarTab(tab, url, isOriginal))
+    : -1;
```
darkwing: During manual testing I noticed that only one URL-less source tab would appear, and quickly figured out this was the problem.  This snippet was finding a tab without a URL, and saying, "OK, we already have this, don't create a new tab"

src/utils/source.js
```diff
@@ -151,7 +151,7 @@ export function getFormattedSourceId(id: string) {
  */
 export function getFilename(source: Source) {
   const { url, id } = source;
-  if (!url) {
+  if (!getRawSourceURL(url)) {
```
darkwing: This is pretty print optimization.

PR #7107, opened 2018-10-12
Change tab interface to have blue borders WIP

### LINKED ISSUE #7034: Change tab interface to have blue borders
The tab styling between the Debugger tool and the Inspector and Network tools are different. Perhaps change the tab interface for the Debugger tool to be consistent with the blue border style that the  Inspector and Network tools have?

This is suggested from a discussion from #7006 
 
--------------------------------------------------------------------------------------------------------
Debugger:

--------------------------------------------------------------------------------------------------------
Inspector:

--------------------------------------------------------------------------------------------------------
Network:

Doomwyte:
Fixes #7034 

### Summary of Changes

* Added blue border style to the Source/Outline and file tabs

### To-do

- [x]  Add blue border style to the Sources / Outline tabs
- [x]  Get rid of line between Sources and Outline (image 1 below)
- [x] Get rid of space between file tabs (image 2 below)
- [x] Fix colouring for dark theme
- [x] No hover effect for currently selected tab
- [x] Create separator lines between tabs
    - [x] Remove bottom line for selected tabs
    - [x] Fix jittering when selecting a file tab
- [x] Copy over animation in hovered tabs from top-level DevTools tabs

### Screenshots
Light theme
screen shot 2018-10-18 at 8 48 58 pm

Dark theme
screen shot 2018-10-18 at 8 48 14 pm

### DIFF
--- a/src/components/Editor/Tab.js
+++ b/src/components/Editor/Tab.js
@@ -192,6 +192,7 @@ class Tab extends PureComponent<Props> {
         onContextMenu={e => this.onTabContextMenu(e, sourceId)}
         title={getFileURL(source)}
       >
+        <span className="devtools-tab-line"></span>
         <SourceIcon
           source={source}
           shouldHide={icon => ["file", "javascript"].includes(icon)}
--- a/src/components/Editor/Tab.js
+++ b/src/components/Editor/Tab.js
@@ -192,7 +192,7 @@ class Tab extends PureComponent<Props> {
         onContextMenu={e => this.onTabContextMenu(e, sourceId)}
         title={getFileURL(source)}
       >
-        <span className="devtools-tab-line"></span>
+        <span className="source-tab-line"></span>
         <SourceIcon
           source={source}
           shouldHide={icon => ["file", "javascript"].includes(icon)}
--- a/src/components/Editor/Tabs.css
+++ b/src/components/Editor/Tabs.css
@@ -58,13 +58,13 @@
 }
 
 .source-tab:hover {
-  border-top: 2px solid rgba(0,0,0,.2);
+  border-top: 2px solid rgba(0, 0, 0, 0.2);
   background-color: var(--theme-toolbar-hover);
 }
 
 .source-tab.active {
-  color: rgb(0, 96, 223);;
-  border-top: 2px solid rgb(0, 96, 223);;
+  color: rgb(0, 96, 223);
+  border-top: 2px solid rgb(0, 96, 223);
   border-bottom-color: transparent;
 }
 
--- a/src/components/Editor/Tabs.css
+++ b/src/components/Editor/Tabs.css
@@ -1,6 +1,10 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
+ :root {
+  --animation-easing-function: cubic-bezier(.07, .95, 0, 1);
+  --theme-toolbar-selected: rgb(10, 132, 255);
+}
 
 .source-header {
   border-bottom: 1px solid var(--theme-splitter-color);
@@ -38,7 +42,8 @@
 }
 
 .source-tab {
-  border-top: 2px solid transparent;
+  border-left: 1px solid transparent;
+  border-right: 1px solid transparent;
   display: inline-flex;
   align-items: center;
   position: relative;
@@ -57,17 +62,11 @@
 }
 
 .source-tab:not(.active):hover {
-  border-top: 2px solid rgba(0, 0, 0, 0.2);
   background-color: var(--theme-toolbar-hover);
-}
-
-.theme-dark .source-tab:hover {
-  border-top: 2px solid var(--tab-line-hover-color);
 }
 
 .source-tab.active {
   color: rgb(0, 96, 223);
-  border-top: 2px solid rgb(10, 132, 255);
   border-bottom-color: transparent;
   border-left: 1px solid var(--theme-splitter-color);
   border-right: 1px solid var(--theme-splitter-color);
@@ -84,10 +83,6 @@
 
 .theme-dark .source-tab.active {
   color: var(--theme-toolbar-selected-color);
-}
-
-.theme-dark .source-tab.active:hover {
-  border-top: 2px solid rgb(10, 132, 255);
 }
 
 .source-tab.active path,
@@ -167,4 +162,16 @@ html[dir="rtl"] img.moreTabs {
 
 .source-tab:hover .close-btn {
   visibility: visible;
+}
+
+.source-tab:not(.active):hover .devtools-tab-line {
+  background: var(--tab-line-hover-color);
+  opacity: 1;
+  transform: scaleX(1);
+}
+
+.source-tab.active .devtools-tab-line {
+  background: var(--theme-toolbar-selected);
+  opacity: 1;
+  transform: scaleX(1);
 }
--- a/src/components/Editor/Tabs.css
+++ b/src/components/Editor/Tabs.css
@@ -62,8 +62,8 @@
   background-color: var(--theme-toolbar-hover);
 }
 
-.theme-dark .source-tab:hover { 
-  border-top: 2px solid rgb(177, 177, 179);;
+.theme-dark .source-tab:hover {
+  border-top: 2px solid rgb(177, 177, 179);
 }
 
 .source-tab.active {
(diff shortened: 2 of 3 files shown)


darkwing:
Whoa, nice!  I'd love your thoughts before merging @violasong 
tabs

Doomwyte:
I'm on it! Thanks for the suggestions!

fvsch:
Sorry to mention this very late, but has there been some discussion about using the existing tab implementation used in the toolbox (devtools top tabs), Inspector, and Network monitor? If this component can be reused in the Debugger, and maybe extended to add support for the close button, that could avoid some duplication.

It's the `.devtools-tab` component on the CSS side (defined in `devtools/client/themes/toolbox.css`). Not sure if it's an easily reusable React component on the JS side (might be older code).

If we need to reimplement it in the Debugger (e.g. because Debugger can be used outside of DevTools and needs at the very least a copy of this component), we can at least copy or pick from the styles for this element, and especially the styles for the `.devtools-tab-line` animation. There's a specific easing curve to use: https://design.firefox.com/photon/motion/duration-and-easing.html

violasong:
Makes sense - in case it matters, there are a bunch more changes beyond close buttons to support variably-named file tabs as separate from section tabs: truncation rules, main Firefox-style separator lines/colors/hover behaviors.

This might be best for a separate patch, but: I just realized though that the Sources and Outline tabs should probably match regular DevTools sidebar tabs, and the whole Debugger tab/toolbar should probably match the other DevTools shorter toolbars in height.

fvsch:
Toolbar heights, excluding borders, are:
- 28px for the main DevTools toolbar
- 24px for tool-specific toolbars, historically and still true in some tools such as the Inspector, Performance, Memory, Storage
- 28px for the top-level toolbar in Console and Network

So there's a bit of discrepancy nowadays (not counting all the sizing bugs with toolbars which are 1px too small or too big).

@violasong, were you suggesting using 24px or 28px for the debugger top toolbars?

violasong:
A thing to test for later is whether 24px works for Console and Network toolbars. Emailing myself @violasong

violasong:
Sounds good

Doomwyte:
For the `.devtools-tab-line` animation, I am unable to copy it over to the current css class that implements the blue border on top, because with `.devtools-tab-line`, the class selects only the blue border. With our current implementation, it selects the entire tab and the blue border on top is just a `border-top` styling, so when I apply the animation, it makes the entire tab disappear and not just the blue border on top.

A possibility I thought of is to remove the `border-top` styling from the current css class and create a `.devtools-tab-line`-like class that creates the blue border on top. Is this a good idea / is there a better way to approach this / should I just keep it as it is? (I notice that besides the top-level DevTools tabs, the other tabs inside Inspector and Network don't seem to have any animations). Thanks for your help!

darkwing:
@Doomwyte Hello Lily!   Do you have what you need here to complete this or did you have further questions?  Thank you!

Doomwyte:
@darkwing Hi! I still have that question I asked on slack yesterday, couldn't figure it out

darkwing:
Hello @Doomwyte!  I looked into this.  A few things:

1.  Your `:before` code does register:

itsthere

The animation names aren't there, however.

2.  I think there's a few items in the CSS you've copied that is causing you problems.  Here's what I did:

```css
.source-tab::before { 
  content: "";
  position: absolute;
  top: 0;
  left: 0; 
  right: 0; /* added this */
  width: 100%;  
  height: 2px;
  /*
  background: transparent;
  transition: transform 250ms var(--animation-easing-function), opacity 250ms var(--animation-easing-function);
  opacity: 0;
  transform: scaleX(0);
  */
  background: pink;
}
```

yayyay

This should help you move forward 👍

Doomwyte:
@darkwing Hi again! I was getting stuck on using the pseudo-element, so I tried the other method of outputting a `<span className="devtools-tab-line"></span>` and I think the animation is working okay for the file tabs now. It wasn't working for the Source/Outline tabs, so I need to look at that more...

Also, your suggestion for adding in transparent borders on each side solved the side to side jitter problem, thanks! I'm still seeing the up and down jitter though, but I am not sure how to solve it...I think it has to do with me increasing the active tab to 30px and to adjust for it doing `bottom: 2px;` for non-active tabs, but I don't know how to get around it.

Doomwyte:
Oops, another question is I added some `:root {}` variables to `Tabs.css`, but I'm not sure if this is where I should put it. Is there a more reusable place I can put these variables or is this okay? Thanks!

src/components/Editor/Tabs.css
```diff
@@ -49,26 +52,37 @@
...
+  border-right: 1px solid var(--theme-splitter-color);
+
+  /* added this part to connect active file tab to editor */
+  background-color: var(--theme-body-background);
+  height: 30px;
+}
+
+/* added this part to balance out height of the not active file tabs */
+.source-tab:not(.active) {
+  bottom: 2px;
```
fvsch: Shouldn't the default state and active state have the same height? That's how the `devtools-tab` component is designed and implemented.

src/components/Editor/Tab.js
```diff
@@ -192,6 +192,7 @@ class Tab extends PureComponent<Props> {
         onContextMenu={e => this.onTabContextMenu(e, sourceId)}
         title={getFileURL(source)}
       >
+        <span className="devtools-tab-line"></span>
```
fvsch: Using a decorative span is alright, but can we use a class such as `source-tab-line` since it's a visual element of the `source-tab` component?

fvsch:
> I think it has to do with me increasing the active tab to 30px and to adjust for it doing bottom: 2px; for non-active tabs, but I don't know how to get around it.

Don't change the element's box model metrics (height, border-width, padding, margin, font-weight, line-height, etc.) for different visual states, if you don't want this element to move or make its siblings move. The solution is to define those metrics once for the default state, and not change them in other states (hover, focus, active, etc.).

src/components/PrimaryPanes/Sources.css
```diff
@@ -160,33 +161,41 @@
...
-  border-color: var(--theme-splitter-color);
+.theme-dark .source-outline-tabs .tab:hover {
+  border-top: 2px solid var(--tab-line-hover-color);
 }
 
 .source-outline-tabs .tab.active {
-  background-color: var(--theme-body-background);
-  border-bottom-color: transparent;
+  border-top: 2px solid rgb(10, 132, 255);
+  color: rgb(0, 96, 223);
```
fvsch: Can we use colors from the Photon design system here?
If these are already Photon colors, can we use `var(--blue-50)` and other DevTools-wide theme variables here to be more explicit and consistent?

@darkwing do you know if and how the Debugger codebase can use DevTools variables canonically defined in `mozilla-central/devtools/client/themes/variables.css`? Are these imported in the Debugger codebase in some way?

src/components/PrimaryPanes/Sources.css
```diff
@@ -160,33 +161,41 @@
...
+}
+
+.theme-dark .source-outline-tabs .tab.active {
+  color: var(--theme-toolbar-selected-color);
+}
+
+.theme-dark .source-outline-tabs .tab.active:hover {
+  border-top: 2px solid rgb(10, 132, 255);
```
fvsch: We should not change box-model metrics (border-top-width, here) for a specific theme (`.theme-dark`). Themes should only change colors, and if possible we should change colors globally. Something like:

```css
:root {
  --source-tab-hover-border-color: pink;
  --source-tab-active-border-color: red;
}
:root.theme-dark {
  --source-tab-hover-border-color: green;
  --source-tab-active-border-color: lime;
}
```

Doomwyte:
1. It was something @violasong suggested in the comments above, so I thought I'd give it a try. My bad if I misunderstood. ```Hm, one nice thing about separator lines between the tabs is that then we can remove the bottom line for selected tabs, further connecting the selected tab to selected file.```
2. That's a good idea, I'll look into it! Thanks!

src/components/PrimaryPanes/Sources.css
Doomwyte: Ah I did not know that, thanks for letting me know! And to your other comments as well, thank you 👍

darkwing:
Hello @Doomwyte!  I figured it out!  Check this out:

https://gist.github.com/darkwing/4eb3ebb58d1f37f1dca8952a949f9cf7

We can use a gradient background to keep the border on the bottom without height manipulation 👍

darkwing:
```css
linear-gradient(var(--theme-toolbar-hover) 28px, transparent 1px)
```

darkwing:
This is a great place to merge and enhance.  Awesome work @Doomwyte !

PR #7125, opened 2018-10-21
Added Flow in src/actions/ast/setInScopeLines.js

Kanikashrivastav:
Fixes a part #5120

### DIFF
--- a/src/actions/ast/setInScopeLines.js
+++ b/src/actions/ast/setInScopeLines.js
@@ -2,10 +2,14 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
 
+// @flow
+
 import { getOutOfScopeLocations, getSelectedSource } from "../../selectors";
 import { getSourceLineCount } from "../../utils/source";
 
 import { range, flatMap, uniq, without } from "lodash";
+import type { AstLocation } from "../../workers/parser";
+import type { ThunkArgs } from "../types";
 
 function getOutOfScopeLines(outOfScopeLocations: AstLocation[]) {
   if (!outOfScopeLocations) {
@@ -13,7 +17,7 @@ function getOutOfScopeLines(outOfScopeLocations: AstLocation[]) {
   }
 
   return uniq(
-    flatMap(outOfScopeLocations, location =>
+    flatMap((outOfScopeLocations:any), location =>
       range(location.start.line, location.end.line)
     )
   );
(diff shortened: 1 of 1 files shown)


src/actions/ast/setInScopeLines.js
```diff
@@ -2,18 +2,22 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
 
+// @flow
+
 import { getOutOfScopeLocations, getSelectedSource } from "../../selectors";
 import { getSourceLineCount } from "../../utils/source";
 
 import { range, flatMap, uniq, without } from "lodash";
+import type { AstLocation } from "../../workers/parser";
```
darkwing: These probably aren't required.

src/actions/ast/setInScopeLines.js
```diff
@@ -2,18 +2,22 @@
...
+import type { ThunkArgs } from "../types";
 
 function getOutOfScopeLines(outOfScopeLocations: AstLocation[]) {
   if (!outOfScopeLocations) {
     return null;
   }
 
   return uniq(
-    flatMap(outOfScopeLocations, location =>
+    flatMap((outOfScopeLocations:any), location =>
```
darkwing: We should avoid `:any` and use its actual type.

src/actions/ast/setInScopeLines.js
```diff
@@ -2,12 +2,15 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
 
+// @flow
+
 import { getOutOfScopeLocations, getSelectedSource } from "../../selectors";
 import { getSourceLineCount } from "../../utils/source";
 
 import { range, flatMap, uniq, without } from "lodash";
+import type { ThunkArgs } from "../types";
```
darkwing: I don't think this is necessary.

src/actions/ast/setInScopeLines.js
```diff
@@ -2,12 +2,15 @@
...
 
-function getOutOfScopeLines(outOfScopeLocations: AstLocation[]) {
+function getOutOfScopeLines(outOfScopeLocations) {
```
darkwing: Why remove the `AstLocation` type here?

Kanikashrivastav:
I am getting these error if I do the changes that you suggested, I have installed everything using 

> yarn install

 
image
image

darkwing:
Right -- you'll need to import that type.

Kanikashrivastav:
So, shouldn't this be merged rather than closed? @darkwing

darkwing:
I was hoping you could take another shot at this @Kanikashrivastav; the changes here are going in the wrong direction and could use a fresh start.

PR #7128, opened 2018-10-22
Add a displayFullUrl prop to the Frame component.

nchevobbe:
This props will allow the component to be rendered with the full URL of the frame, instead of just the filename as its currently the case.
This is needed by the webconsole where we do have a lot more space to display information.

### Screenshot

image


nchevobbe:
See [Bug 1390768](https://bugzilla.mozilla.org/show_bug.cgi?id=1390768)

nchevobbe:
See [Bug 1390768](https://bugzilla.mozilla.org/show_bug.cgi?id=1390768)

PR #7129, opened 2018-10-22
Export non-connected Frames component.

nchevobbe:
This will allow other panels (e.g. Console, Netmonitor, …) to import it and passing the appropriate props.

For some reasons, this triggered Flow errors (which I fixed)

### DIFF
--- a/src/components/SecondaryPanes/Frames/index.js
+++ b/src/components/SecondaryPanes/Frames/index.js
@@ -55,7 +55,7 @@ class Frames extends Component<Props, State> {
   toggleFrameworkGrouping: Function;
   renderToggleButton: Function;
 
-  constructor(props) {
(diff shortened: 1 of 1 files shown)


loganfsmyth:
How will the other tools be importing this exactly? Should we be planning to reorganize this code so that `Frame` is in a new shared package?

nchevobbe:
> How will the other tools be importing this exactly? Should we be planning to reorganize this code so that Frame is in a new shared package?

Ultimately, yes, that would be the goal. For now, the console would directly import the component from the debugger path.

In the end, the goal would be to replace https://searchfox.org/mozilla-central/source/devtools/client/shared/components/StackTrace.js with this component, so it would be great to emphasize that it's not restricted to the debugger.

PR #7137, opened 2018-10-22
Match outline functions

### LINKED ISSUE #4443: [PrettyPrint] Add test to ensure original source and pretty printed outline tab functions match
The sample URI I used was:

https://davidwalsh.name/wp-includes/js/wp-embed.min.js?ver=4.8.2

The outline tab for the original source shows:

```
c()
receiveEmbedMessage(c)
```

The pretty-printed outline tab shows:

```
"No functions"
```

Unless I'm missing something obvious, there's no real structure change between the two, so the Outline tab function listing shouldn't change.

akashs09:
Fixes #4443 

Here's the Pull Request Doc

### Summary of Changes

* change 1
* change 2

### Test Plan

Example test plan:

Feel free to improve it!


src/test/mochitest/browser_dbg-outline.js
```diff
@@ -9,6 +9,20 @@ function getNthItem(dbg, index) {
...
+add_task(async function () {
+  const dbg = await initDebugger("doc-scripts.html");
+  const {
+    selectors: { getSelectedSource },
+    getState
+  } = dbg;
+  await selectSource(dbg,"simple1", 1);
+  findElementWithSelector(dbg, ".outline-tab").click();
+  const originalSource = getItems(dbg);
+  findElementWithSelector(dbg, ".prettyPrint").click();
```
loganfsmyth: When writing new tests, it's always a good idea to look for existing tests that do similar things. For example in https://github.com/devtools-html/debugger.html/blob/6c82067dc511efe01186745a7d21fa17e076d920/src/test/mochitest/browser_dbg-tabs-pretty-print.js#L10 we have an example that clicks this button.

This also demonstrates another piece that is missing here. Given that pretty-printing will not happen instantly, there likely needs to be some kind of delay before you continue like the one at https://github.com/devtools-html/debugger.html/blob/6c82067dc511efe01186745a7d21fa17e076d920/src/test/mochitest/browser_dbg-tabs-pretty-print.js#L11

src/test/mochitest/browser_dbg-outline.js
```diff
@@ -9,6 +9,20 @@ function getNthItem(dbg, index) {
...
 }
 
+// Tests that the outline functions for original and pretty printed source matches
+add_task(async function () {
+  const dbg = await initDebugger("doc-scripts.html");
+  const {
+    selectors: { getSelectedSource },
+    getState
+  } = dbg;
+  await selectSource(dbg,"simple1", 1);
```
loganfsmyth: Missing space between `,`, and `"`.

src/test/mochitest/browser_dbg-outline.js
```diff
@@ -9,6 +9,20 @@ function getNthItem(dbg, index) {
...
+  findElementWithSelector(dbg, ".outline-tab").click();
+  const originalSource = getItems(dbg);
+  findElementWithSelector(dbg, ".prettyPrint").click();
+  const prettySource = getItems(dbg);
+  is(originalSource.length, prettySource.length, "Outline functions match");
+});
 // Tests that the editor highlights the correct location when the
```
loganfsmyth: It would be good to add a newline above this to have some visual separation.

src/test/mochitest/browser_dbg-outline.js
```diff
@@ -9,6 +9,20 @@ function getNthItem(dbg, index) {
   return findElement(dbg, "outlineItem", index);
 }
 
+// Tests that the outline functions for original and pretty printed source matches
```
loganfsmyth: I'd probably clarify this, since this only verifies that they have the same number of items, not that they match.

src/test/mochitest/browser_dbg-outline.js
```diff
@@ -9,6 +9,25 @@ function getNthItem(dbg, index) {
...
+
+  await selectSource(dbg, "simple1");
+  findElementWithSelector(dbg, ".outline-tab").click();
+  const originalSource = getItems(dbg);
+
+  clickElement(dbg, "prettyPrintButton");
+  await waitForSource(dbg, "simple1:formatted");
+  const prettySource = getItems(dbg);
+
+  is(originalSource.length, prettySource.length, "Length of outline functions for both prettyPrint and originalSource same");
```
darkwing: The test fails for me here but it's a timing issue.  Use `waitForElement` to wait for the function list to be present.  The timing issue is that the pane starts out with content `Loading...`, and so that list wont be there immediately.

PR #7140, opened 2018-10-23
Use renderStacktrace prop if provided to render Error rep stacktrace

### LINKED ISSUE #7130: Use the `renderStacktrace` prop to render the stacktrace in Error reps
In [Bug 1390768](https://bugzilla.mozilla.org/show_bug.cgi?id=1390768), we want to use debugger's `Frames` component to render stacktraces in the console.
We also want to be able to use the same component that we will create there for Errors as well.
This means that in the `Error` rep, if a `renderStacktrace` prop is passed, it should be used to render the stacktrace.
if it is not passed, then we should default to the current rendering.

nchevobbe:
This will allow us to use the same component
in the console and in reps for example.
If the prop is not passed, it defers to the
current stacktrace rendering.

Fixes #7130

### DIFF
--- a/packages/devtools-reps/src/reps/error.js
+++ b/packages/devtools-reps/src/reps/error.js
@@ -20,7 +20,8 @@ const IGNORED_SOURCE_URLS = ["debugger eval code"];
 ErrorRep.propTypes = {
   object: PropTypes.object.isRequired,
   // @TODO Change this to Object.values when supported in Node's version of V8
-  mode: PropTypes.oneOf(Object.keys(MODE).map(key => MODE[key]))
+  mode: PropTypes.oneOf(Object.keys(MODE).map(key => MODE[key])),
+  renderStacktrace: PropTypes.func
 };
 
 function ErrorRep(props) {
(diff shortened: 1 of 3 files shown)


nchevobbe:
> I don't think we want the change to `assets/module-manifest.json`

good catch!

darkwing:
@janodvarko Just a heads up you were requested for review here

packages/devtools-reps/src/reps/error.js
```diff
@@ -20,7 +20,8 @@ const IGNORED_SOURCE_URLS = ["debugger eval code"];
 ErrorRep.propTypes = {
   object: PropTypes.object.isRequired,
   // @TODO Change this to Object.values when supported in Node's version of V8
-  mode: PropTypes.oneOf(Object.keys(MODE).map(key => MODE[key]))
+  mode: PropTypes.oneOf(Object.keys(MODE).map(key => MODE[key])),
+  renderStacktrace: PropTypes.func
```
janodvarko: It would be great to have an explanatory comment for the new prop

```// Custom function used to render the stack trace (optional)
  renderStacktrace: PropTypes.func
```

nchevobbe: done ✅

PR #7147, opened 2018-10-23
Properly set permissions for testing, label nit

darkwing:
- Label nit was something @flodolo had pointed out but we missed.
- The condition for a test goes underneath it, so we accidentally added a condition to the xhr breakpoint test and removed it from its desired test.

### DIFF
--- a/src/test/mochitest/browser.ini
+++ b/src/test/mochitest/browser.ini
@@ -655,8 +655,8 @@ support-files =
 [browser_dbg-asm.js]
 [browser_dbg-async-stepping.js]
 [browser_dbg-sourcemapped-breakpoint-console.js]
-[browser_dbg-xhr-breakpoints.js]
 skip-if = (os == "win" && ccov) # Bug 1453549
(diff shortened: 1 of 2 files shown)


src/test/mochitest/browser.ini
```diff
@@ -655,8 +655,8 @@ support-files =
 [browser_dbg-asm.js]
 [browser_dbg-async-stepping.js]
 [browser_dbg-sourcemapped-breakpoint-console.js]
-[browser_dbg-xhr-breakpoints.js]
 skip-if = (os == "win" && ccov) # Bug 1453549
+[browser_dbg-xhr-breakpoints.js]
```
AnshulMalik: So we are skipping it for windows build?

darkwing: No; the `skip-if` conditions apply to the test *above* the `skip-if` text, so in this case line 657

AnshulMalik: ohh, Thanks

PR #7148, opened 2018-10-23
Backport mc changes from release 94 and 95

darkwing:
I ran into these collisions when doing releases 94 and 95.


loganfsmyth:
Looks good. I assume the flow failures are expected because the function signatures changed and the types didn't get updated.

darkwing:
Yep!  I'll fix the flow stuff tomorrow; I only had the energy to send the backported code in tonight.

PR #7151, opened 2018-10-24
Use --error-color for --string-color (#7149)

### LINKED ISSUE #7149: Use --error-color for --string-color
In Bug [1494789](https://bugzilla.mozilla.org/show_bug.cgi?id=1494789#c17) we updated the color for strings inside errors in the console. Part of that touched reps.css and we need to land this change here to be in sync with changes made in bugzilla.

kriskristin:
Fixes #7149

We need to land this so when the next devtools update lands in mozilla central, we don't overwrite the following change (landed in 1494789): https://hg.mozilla.org/mozilla-central/rev/5566199eda70

r? @nchevobbe

### DIFF
--- a/packages/devtools-reps/src/reps/reps.css
+++ b/packages/devtools-reps/src/reps/reps.css
@@ -6,7 +6,7 @@
 .theme-dark,
 .theme-light {
   --number-color: var(--theme-highlight-green);
-  --string-color: var(--theme-highlight-red);
+  --string-color: var(--error-color);
   --null-color: var(--theme-comment);
   --object-color: var(--theme-highlight-blue);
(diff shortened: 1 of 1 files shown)


nchevobbe:
Oh, I didn't realize yesterday that this change impacts *all* the strings in objects, not only in error message: 

image

nchevobbe:
Sorry, this lacks explanation (I shouldn't write things in the morning :) )
So, I think we shouldn't make this PR. The only thing that we should do is to ensure to update the `--string-color` to `--error-color` for error messages in webconsole.css.

Sorry I made you create this PR in the first place, I should have ensure that this was needed.
I'm going to close the PR, and we can discuss in the issue if you want (or in the bugzilla bug)

PR #7152, opened 2018-10-24
Render Frame title attribute with a getFrameTitle prop

### LINKED ISSUE #7133: Allow to have a custom `title` in the Frame component
Since we plan to use the Frames component in the console, we need to have almost the same feature as the one we have in the current console stacktrace.
One of them is the tooltip we currently have when hovering a location:
image

(note the `View source in debugger - ${FULL URL}`).

---

The same thing should be achievable in the debugger Frame component. We could have a `getFrameTooltip(url: String) : String` prop in `Frames`, `Group` and finally `Frame` that would be use to compute the `title` property of the top element rendered in `Frame`.

This could also benefits the debugger, where we could put something like `Jump to frame` tooltiip (or something better, this is not good wording here :D )

nchevobbe:
Fixes #7133

### DIFF
--- a/src/components/SecondaryPanes/Frames/Frame.js
+++ b/src/components/SecondaryPanes/Frames/Frame.js
@@ -44,13 +44,8 @@ function FrameLocation({ frame, displayFullUrl = false }: FrameLocationProps) {
   const filename = displayFullUrl
     ? getFileURL(source, false)
     : getFilename(source);
-  const title = `${getFileURL(source, false)}:${location.line}`;
 
-  return (
-    <div className="location" title={title}>{`${filename}:${
-      location.line
-    }`}</div>
-  );
+  return <div className="location">{`${filename}:${location.line}`}</div>;
 }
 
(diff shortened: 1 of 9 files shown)


src/components/SecondaryPanes/Frames/Frame.js
```diff
@@ -118,12 +114,20 @@ export default class FrameComponent extends Component<FrameComponentProps> {
...
 
     const className = classNames("frame", {
       selected: selectedFrame && selectedFrame.id === frame.id
     });
+
+    const title = getFrameTitle
+      ? getFrameTitle(
+          `${getFileURL(frame.source, false)}:${frame.location.line}`
+        )
+      : undefined;
```
darkwing: Would `""` be better here?

nchevobbe: `""` would create an empty `title` attribute in the dom node, whereas `undefined` do not.
So, that's not a really big deal, but I'd rather not render anything on the DOM that is not useful.

PR #7153, opened 2018-10-24
Add a maxLength option to formatDisplayName.

nchevobbe:
The maxLength will default to the current length
we use (25), but could be overrided to whatever
the consumer wants. Setting it to null will not
truncate the string at all.

### DIFF
--- a/src/utils/pause/frames/displayName.js
+++ b/src/utils/pause/frames/displayName.js
@@ -74,10 +74,13 @@ function mapDisplayNames(frame, library) {
   );
 }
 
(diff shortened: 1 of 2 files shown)


darkwing:
Is this something we want for blanket debugger usage?

nchevobbe:
Yes, I'll use it in the Frame component (overflow should be done in CSS
there)

Le mer. 24 oct. 2018 à 20:51, David Walsh <notifications@github.com> a
écrit :

darkwing:
Ooops, I see we were already truncating to 25. 👍

PR #7157, opened 2018-10-24
testing out flow annotations and acceptance of static types

### LINKED ISSUE #5120: [Flow] add eslint-plugin-flowtype - require-valid-file-annotation

akashs09:
Fixes #5120 
some redundant static types annotations may have been added but wanted to know the length/scope of annotations needed for merging.
Here's the Pull Request Doc

### Summary of Changes

* change 1
* change 2

### Test Plan

Example test plan:

Feel free to improve it!

### DIFF
--- a/src/actions/breakpoints/addBreakpoint.js
+++ b/src/actions/breakpoints/addBreakpoint.js
@@ -2,6 +2,8 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
 
+// @flow
+
 import { isOriginalId } from "devtools-source-map";
 import {
   locationMoved,
@@ -28,18 +30,21 @@ async function addBreakpointPromise(getState, client, sourceMaps, breakpoint) {
   };
 
   const generatedLocation = await getGeneratedLocation(
-    state,
-    source,
-    location,
-    sourceMaps
+    (state: State),
+    (source: Source),
+    (location: Location),
+    (sourceMaps: SourceMaps)
   );
 
-  const generatedSource = getSource(state, generatedLocation.sourceId);
+  const generatedSource = getSource(
+    (state: State),
+    (generatedLocation.sourceId: String)
+  );
 
-  assertLocation(location);
-  assertLocation(generatedLocation);
+  assertLocation((location: Location));
+  assertLocation((generatedLocation: GeneratedLocation));
 
-  if (breakpointExists(state, location)) {
+  if (breakpointExists((state: State), (location: Location))) {
     const newBreakpoint = { ...breakpoint, location, generatedLocation };
     assertBreakpoint(newBreakpoint);
     return { breakpoint: newBreakpoint };
@@ -48,7 +53,7 @@ async function addBreakpointPromise(getState, client, sourceMaps, breakpoint) {
   const { id, hitCount, actualLocation } = await client.setBreakpoint(
     generatedLocation,
     breakpoint.condition,
-    isOriginalId(location.sourceId)
+    isOriginalId((location.sourceId: String))
   );
 
   const newGeneratedLocation = actualLocation || generatedLocation;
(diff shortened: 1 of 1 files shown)


loganfsmyth:
In the 99% of cases, a typecast like this should not be needed. When I'm adding type annotations to a codebase, essentially I want to add as few annotations as possible since the vast majority Flow will be able to infer on its own. Flow requires type annotations on functions that are exported, and I might add type annotations to other functions if it helps document how they behave.

When adding annotations to existing code, I would add the `// @flow` annotation to the file and then run Flow to see what errors come up, then work to figure out the cause of the errors and fix them.

src/actions/breakpoints/addBreakpoint.js
```diff
@@ -28,18 +30,21 @@ async function addBreakpointPromise(getState, client, sourceMaps, breakpoint) {
   };
```
loganfsmyth: For instance, if you run Flow, the first error in this file is the 2 lines above here, because `source` here could technically be `null` if the `breakpoint.location.sourceId` was actually for a source that it couldn't find.

If I were writing this fresh, I personally might just make `getSource` throw instead of returning `null`, but that's probably too aggressive of a change for you to make, so instead I'd probably add
```diff
  const source = getSource(state, breakpoint.location.sourceId);
+ if (!source) {
+   throw new Error(`Failed to add breakpoint because ${breakpoint.location.sourceId} is unknown`);
+ }
```
because this means that Flow can know for sure that `source` _must_ point at a valid source file by the time it has to read `source.id`.

Adding Flow to an existing codebase often comes down to 

1. Adding annotations where strictly needed
2. Tweaking your code to prove to Flow that the types actually fit together

loganfsmyth:
I think I'm going to close this, since realistically any update will be an entirely new set of changes.

PR #7161, opened 2018-10-25
Change onClick to onMouse/KeyDown for search bar modifiers

### LINKED ISSUE #7013: [SearchBar] Modifers need two clicks to activate when input bar is highlighted
**Steps to Reproduce**

* Ctrl-F to open Find in File search.
* Click on a modifier. Input goes out of focus, but modifier does not highlight.
* Click on modifier again. Modifier highlights.

search-click

audreyfang:
Fixes #7013

### Summary of Changes

* Modifiers on the Find in File search bar no longer need to be clicked twice when the search bar is in focus.

### Test Plan

- [x] Ctrl-F opens the search bar
- [x] Search for something
- [x] Clicking a modifier makes the search results change immediately.

### DIFF
--- a/src/components/Editor/SearchBar.js
+++ b/src/components/Editor/SearchBar.js
@@ -276,6 +276,12 @@ class SearchBar extends Component<Props, State> {
             toggleFileSearchModifier(modVal);
             doSearch(query);
           }}
+          onKeyDown={(e: any) => {
+            if (e.key === "Enter") {
+              toggleFileSearchModifier(modVal);
+              doSearch(query);
+            }
+          }}
           title={tooltip}
         >
           <Svg name={svgName} />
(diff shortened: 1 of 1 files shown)


darkwing:
@audreyfang Were you able to chase down why `click` stopped working?

audreyfang:
@darkwing part of it is the ordering of events. onclick fires after onblur so the click event gets lost for some reason when the search input blurs. I’m not completely sure exactly why the event is lost. Do you want me to look into that?

darkwing:
I think it would be worth some time, just to make sure we aren't papering over a larger issue.

darkwing:
I'm going to merge this for now since the need to click twice on an icon is annoying.  We should continue researching the click issue though.

PR #7162, opened 2018-10-25
Handle Frame overflow (function and location) in CSS.

### LINKED ISSUE #7131: Allow to not truncate the name of the function in Frame component
At the moment, the function name is truncated (to 25 I think ?).
We should not truncate it using JS, but with CSS in order to be suited for all sizes.

nchevobbe:
Fixes #7131 

The Frames component has limited space available,
and it might happen that we need to truncate some
information.
Previouysly, this was done by manually setting a
length limit on the string we were displaying.
This is a bit unfortunate since CSS has some useful
properties we can use to show whatever fits on the
screen.
So now, in Frames, we don't truncate strings anymore,
and, in CSS, we set the text-overflow we want.
Furthermore, we make the Frame element wrap, so when
the panel gets smaller, the location is put on a
new line instead of shrinking (and then shrinks if
it needs to).
In order to have such feature, we needed to have the
location element to not have display: flex. This was
not hard to fix since flex was only used to place an
optional icon after the framework name, which can be
do with inline display.

### Video
(full URL was enabled only to have long location, we'll still only display the filename in the debugger)

oct-25-2018 10-39-31

### DIFF
--- a/src/components/SecondaryPanes/Frames/Frame.js
+++ b/src/components/SecondaryPanes/Frames/Frame.js
@@ -19,8 +19,8 @@ type FrameTitleProps = {
   options: Object
 };
 
-function FrameTitle({ frame, options }: FrameTitleProps) {
-  const displayName = formatDisplayName(frame, options);
+function FrameTitle({ frame, options = {} }: FrameTitleProps) {
+  const displayName = formatDisplayName(frame, { ...options, maxLength: null });
   return <div className="title">{displayName}</div>;
 }
 
--- a/src/components/SecondaryPanes/Frames/Frames.css
+++ b/src/components/SecondaryPanes/Frames/Frames.css
@@ -9,13 +9,16 @@
 }
 
 .frames ul li {
-  padding: 0 10px 0 21px;
+  padding: 7px 10px 7px 21px;
   overflow: hidden;
   display: flex;
   justify-content: space-between;
+  column-gap: 0.5em;
   flex-direction: row;
   align-items: center;
   margin: 0;
+  max-width: 100%;
+  flex-wrap: wrap;
 }
 
 .frames ul li * {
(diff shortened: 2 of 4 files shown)


darkwing:
I have two concerns here:

1.  The text on the left looks a lot like the text on the right, in color and font-size, so I worry this change will cause a "mash" of text

2.  Are we concerned that the height of this area will get massive?

@violasong @fvsch Any help?

nchevobbe:
The wrapping part can be disabled if needed (also, debugger won't be that dramatic since it only displays filename and not the fill URL)
It would be fine now that we can have a tooltip on the frame so the user can still get the full frame information.
For the text color, I don't think the PR changes that, but I agree.
When I'll be using the component in the console. I'll switch the color to match what we currently have: 

image

nchevobbe:
Here's how it would actually look in the debugger
oct-25-2018 18-10-54

fvsch:
On the UI side, I'm not 100% on board but it looks like an improvement over what we have today. Choosing what and how we crop in CSS is better IMO than hardcoding it by truncating the output string.

For example on this very GitHub page, I can get a call stake looking like this:

invisible function name in call stack

The filenames are long due to cache busting string in them (`bundleName-cacheBustingHash.js` pattern), not truncated aggressively enough, and when making the sidebar not very wide it obliterates the function names (flex-shrink + no min-width, I presume).

So @nchevobbe's result is a nice improvement.
Given the `bundleName<then some hash here>.js` pattern, which seems to be more common, I would truncate text on the right rather than on the left, though.

---

Ultimately, I think I would prefer a design where:

1. If the sidebar is smaller than a given small width (could be as small as 200px), always wrap the file name to the next line, and left-align it.
2. If the sidebar is wider than this breakpoint, always keep two columns (flexbox, space-between, maybe shrink the file name more using different `flex-shrink` values, and 30% or 40% min-widths).

But for now I would go with Nicolas' proposal.

PR #7169, opened 2018-10-25
added flow

ravi610:
Fixes a part of #5120

### DIFF
--- a/src/reducers/sources.js
+++ b/src/reducers/sources.js
@@ -431,7 +431,7 @@ function getSourcesByUrlInSources(
   return urls[url].map(id => sources[id]);
 }
 
-export function getSourceInSources(sources: SourcesMap, id: string): Source {
+export function getSourceInSources(sources: SourcesMap, id: string): ?Source {
   return sources[id];
 }
 
--- a/src/reducers/sources.js
+++ b/src/reducers/sources.js
@@ -431,7 +431,7 @@ function getSourcesByUrlInSources(
   return urls[url].map(id => sources[id]);
 }
 
-export function getSourceInSources(sources: SourcesMap, id: string): ?Source {
+export function getSourceInSources(sources: SourcesMap, id: string): Source {
   return sources[id];
 }
 
--- a/src/utils/source-maps.js
+++ b/src/utils/source-maps.js
@@ -41,7 +41,11 @@ export async function getMappedLocation(
   sourceMaps: Object,
   location: Location
 ): Promise<Location> {
-  const source: Source = getSource(state, location.sourceId);
+  const source = getSource(state, location.sourceId);
+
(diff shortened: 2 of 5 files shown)


src/utils/source-queue.js
```diff
@@ -10,7 +10,7 @@ let supportsWasm = false;
 let queuedSources;
 let currentWork;
 
-async function dispatchNewSources() {
+async function dispatchNewSources(){
```
AnshulMalik: Lit, we need this space :)

src/utils/source-maps.js
```diff
@@ -37,8 +40,8 @@ export async function getMappedLocation(
   state: Object,
   sourceMaps: Object,
   location: Location
-) {
-  const source = getSource(state, location.sourceId);
+): Promise<Location> {
+  const source: Source = getSource(state, location.sourceId);
```
AnshulMalik: We are good without this `Source` type.

src/reducers/sources.js
```diff
@@ -431,7 +431,7 @@ function getSourcesByUrlInSources(
   return urls[url].map(id => sources[id]);
 }
 
-export function getSourceInSources(sources: SourcesMap, id: string): ?Source {
+export function getSourceInSources(sources: SourcesMap, id: string): Source {
```
loganfsmyth: I don't think this is safe. If the `id` doesn't exist in the map, we'll get `undefined` as the result. If we _did_ want this type, I think we'd want this to be

```
const source = sources[id];
if (!source) {
  throw new Error(`Unknown source ${id}`);
}
return source;
```
but looking at places like https://github.com/devtools-html/debugger.html/blob/cbdcc6f424e13c6e919288c31235675f818c8937/src/actions/project-text-search.js#L74 there appear to be places where we explicitly handle the fact that the source might not exist. 

Given that, it might be better to leave this as `?Source` for now and handle it explicitly in `getMappedLocation` by throwing there instead, because otherwise I'd want to do a pass over the whole codebase trying to figure out if we can get away throwing in `getSourceInSources`.

src/utils/source-maps.js
```diff
@@ -37,9 +40,13 @@ export async function getMappedLocation(
   state: Object,
   sourceMaps: Object,
   location: Location
-) {
+): Promise<Location> {
   const source = getSource(state, location.sourceId);
 
+  if (!source) {
+    throw new Error("Unknown source for location");
```
loganfsmyth: @ravi610 Actually, looking at this again it _might_ be more in line with the old behavior to do `return location;` here. Would you be up for making a new PR with that tweak?

loganfsmyth: Done in https://github.com/devtools-html/debugger.html/pull/7176

PR #7174, opened 2018-10-26
Handle anonymous functions more consistently.

nchevobbe:
Anonymous function name was not displayed
at all in the callstack panel, neither it was
in PreviewFunction. In the scopes panel, it was
displayed in another way (`(anonymous)`) than
in the toolbox, where we use `<anonymous>`.
With this patch, we make sure to have anonymous
functions displayed in a consistent way in the debugger,
and matching other panels in the toolbox.

In order to do that, the anonymous item in the properties
file is updated to match the existing ones in mozilla-central.

### DIFF
--- a/assets/panel/debugger.properties
+++ b/assets/panel/debugger.properties
@@ -919,9 +919,10 @@ whyPaused.other=Debugger paused
 # keyboard shortcuts that use the control key
 ctrl=Ctrl
 
-# LOCALIZATION NOTE (anonymous): The text that is displayed when the
-# display name is null.
-anonymous=(anonymous)
+# LOCALIZATION NOTE (anonymous): this string is used to display JavaScript
+# functions that have no given name - they are said to be
+# anonymous.
+anonymousFunction=<anonymous>
 
 # LOCALIZATION NOTE (shortcuts.toggleBreakpoint): text describing
 # keyboard shortcut action for toggling breakpoint
--- a/assets/panel/debugger.properties
+++ b/assets/panel/debugger.properties
@@ -919,8 +919,8 @@ whyPaused.other=Debugger paused
 # keyboard shortcuts that use the control key
 ctrl=Ctrl
 
-# LOCALIZATION NOTE (anonymous): this string is used to display JavaScript
-# functions that have no given name - they are said to be
+# LOCALIZATION NOTE (anonymousFunction): this string is used to display
+# JavaScript functions that have no given name - they are said to be
 # anonymous.
 anonymousFunction=<anonymous>
 
--- a/src/client/firefox/create.js
+++ b/src/client/firefox/create.js
@@ -21,7 +21,10 @@ export function createFrame(frame: FramePacket): ?Frame {
   if (frame.type == "call") {
     const c = frame.callee;
     title =
-      c.name || c.userDisplayName || c.displayName || L10N.getStr("anonymous");
+      c.name ||
+      c.userDisplayName ||
+      c.displayName ||
+      L10N.getStr("anonymousFunction");
   } else {
     title = `(${frame.type})`;
   }
--- a/src/client/firefox/create.js
+++ b/src/client/firefox/create.js
@@ -20,11 +20,7 @@ export function createFrame(frame: FramePacket): ?Frame {
   let title;
   if (frame.type == "call") {
     const c = frame.callee;
-    title =
-      c.name ||
-      c.userDisplayName ||
-      c.displayName ||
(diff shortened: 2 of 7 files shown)


assets/panel/debugger.properties
```diff
@@ -919,9 +919,10 @@ whyPaused.other=Debugger paused
 # keyboard shortcuts that use the control key
 ctrl=Ctrl
 
-# LOCALIZATION NOTE (anonymous): The text that is displayed when the
-# display name is null.
-anonymous=(anonymous)
+# LOCALIZATION NOTE (anonymous): this string is used to display JavaScript
```
flodolo: Need to update also the string reference here

nchevobbe: oops, done

src/utils/pause/frames/displayName.js
```diff
@@ -82,13 +82,22 @@ export function formatDisplayName(
...
-  let { displayName, originalDisplayName, library } = frame;
-  displayName = originalDisplayName || displayName;
+  let {
+    displayName,
+    originalDisplayName,
+    userDisplayName,
+    name,
+    library
+  } = frame;
+  displayName = originalDisplayName || userDisplayName || displayName || name;
```
jasonLaster: this seems like an opportunity to consolidate some logic the `utils/frame` dir

```js
function getDisplayName(frame: Frame) {
  let {
    displayName,
    originalDisplayName,
    userDisplayName,
    name,
    library
  } = frame;
  return originalDisplayName || userDisplayName || displayName || name;
}
```

src/utils/pause/frames/displayName.js
```diff
@@ -82,13 +82,22 @@ export function formatDisplayName(
...
     displayName = mapDisplayNames(frame, library);
   }
 
-  displayName = simplifyDisplayName(displayName);
+  displayName = displayName
+    ? simplifyDisplayName(displayName)
+    : L10N.getStr("anonymousFunction");
+
   return Number.isInteger(maxLength)
     ? endTruncateStr(displayName, maxLength)
```
jasonLaster: this is a tangent, but are we using maxLength anywhere now?

src/utils/pause/frames/displayName.js
```diff
@@ -19,7 +19,7 @@ const annonymousProperty = /([\w\d]+)\(\^\)$/;
 
 export function simplifyDisplayName(displayName: string) {
   // if the display name has a space it has already been mapped
-  if (/\s/.exec(displayName)) {
+  if (!displayName || /\s/.exec(displayName)) {
```
jasonLaster: should we update the type `simplifyDisplayName(displayName: string)`?

src/client/firefox/create.js
```diff
@@ -21,7 +21,10 @@ export function createFrame(frame: FramePacket): ?Frame {
   if (frame.type == "call") {
     const c = frame.callee;
     title =
-      c.name || c.userDisplayName || c.displayName || L10N.getStr("anonymous");
+      c.name ||
+      c.userDisplayName ||
+      c.displayName ||
+      L10N.getStr("anonymousFunction");
```
jasonLaster: do we need `L10N.getStr("anonymousFunction");` if we're going to add it later when we format the display name?

src/utils/pause/frames/displayName.js
nchevobbe: Yeah, `formatCopyName` still does (it doesn't pass anything, but `maxLength` gets defaulted to 25).
Anyway, since `formatCopyName` seems only used to copy the stacktrace, I think it would be good to have the full length there too. Which means we can then remove the `maxLength` option.

I'll do that in a follow-up PR since the concern is a bit different from this one.

src/utils/pause/frames/displayName.js
nchevobbe: sounds good indeed

src/client/firefox/create.js
nchevobbe: probably not, let's remove it.

src/utils/pause/frames/displayName.js
jasonLaster: seems like it accepts a string, but then we check if there is no string. I wonder if we should do

`simplifyDisplayName(displayName: string | void)`

nchevobbe: oh, okay, I'll do that

nchevobbe:
Modified some flow typing

PR #7181, opened 2018-10-29
show wasm breakpoints in panel

### LINKED ISSUE #6570: [wasm] Breakpoints in WASM module are not visible in Breakpoints panel
STR:

1) Load http://janodvarko.cz/test/wasm/basic1/
2) Press `Load Module & Execute Exported Function` button
3) Go to the debugger and select test.wasm file
4) Create a breakpoint at offset 26 (`get_local $var0`)
5) Check out the Breakpoints side panel, there should be one breakpoint
6) Create a breakpoint at offset 28 (`get_local $var1`)
7) Check out the Breakpoints side panel, there should be two breakpoints, but there is still one -> BUG

Honza

AnshulMalik:
Fixes #6570 

### Summary of Changes

* Change `getTextAtPosition` to consider wasm sources

### Test Plan
- Add a breakpoint in wasm file

We can do it two ways, one is we can directly store the parsed text in the `source` but the downside with this is that the text will only be needed when user opens the wasm file. So the whole parsing will be a waste of time.

Currently, we delay the conversion of binary to text till the user opens the wasm file.

### DIFF
--- a/src/utils/source.js
+++ b/src/utils/source.js
@@ -15,6 +15,8 @@ import { getUnicodeUrl } from "devtools-modules";
 import { endTruncateStr } from "./utils";
 import { truncateMiddleText } from "../utils/text";
 import { parse as parseURL } from "../utils/url";
+import { renderWasmText } from "./wasm";
+import { toEditorPosition } from "./editor";
 export { isMinified } from "./isMinified";
 import { getURL, getFileExtension } from "./sources-tree";
 import { prefs } from "./prefs";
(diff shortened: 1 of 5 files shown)


src/utils/source.js
```diff
@@ -378,13 +380,19 @@ export function isLoading(source: Source) {
...
 
   const line = location.line;
   const column = location.column || 0;
 
+  if (source.isWasm) {
+    const { line: editorLine } = toEditorPosition(location);
+    const lines = renderWasmText(source.id, source.text);
+    return lines[editorLine];
+  }
+
```
jasonLaster: it'd be nice to have a unit test for this:

Maybe it is similar to this test:
https://github.com/devtools-html/debugger.html/blob/master/src/utils/tests/wasm.spec.js#L51

yurydelendik: lgtm, though renderWasmText is (very) expensive operation, it will be nice to cache renderWasmText result in case if several breakpoints are set.

jasonLaster:
@yurydelendik how does the flow types + memoization look?

jasonLaster:
@AnshulMalik @yurydelendik how does this look?

yurydelendik:
> how does this look?

this looks beautiful

PR #7186, opened 2018-10-29
[breakpoints] support breakpoint loading

### LINKED ISSUE #6771: [Breakpoints] Use `loading` property

navgurukul-shivani18:
Fixes #6771 
the loading is now true
Added !bp for the bp.condition, didn't quite get by "bp.loading" mentioned in the issue

### DIFF
--- a/src/selectors/breakpointSources.js
+++ b/src/selectors/breakpointSources.js
@@ -28,7 +28,7 @@ function getBreakpointsForSource(
       bp =>
         bp.location.sourceId == source.id &&
         !bp.hidden &&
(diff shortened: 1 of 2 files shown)


darkwing:
@navgurukul-shivani18 How did you test this?  It looks like it breaks one of our tests.

Issue #6631, opened 2018-07-11
Consistent vertical rhythm in debugger panels


digitarald:
Did a quick analysis of line-heights in the side-panel. Turns out that we have a lot going on. I have not even started on font variations, this might be also considered as part of the issue.

My hypothesis is that cleaning this up a bit for a better design could make the information easier to scan.

![image](https://user-images.githubusercontent.com/8599/42604367-3aa7326e-8527-11e8-8591-1cd25e5ad750.png)

@violasong how far does https://design.firefox.com/photon/visuals/grid.html apply to devtools and could guide this? Or could this be used as case-study for a devtools-specific design system?

Edit: Updated image with grid lines and added font differences as an issue.

violasong:
Just wanted to include the comments I wrote on Slack to keep things together :)
- one thing that will help is when we replace the blue background for current frame with an arrow icon as mentioned in the callstack PRD
- I think the difference in spacing between the three visible sections is distracting, esp call stack compared to scopes - they look like three separate worlds (lazy bug I filed about this, I can be more helpful later: https://github.com/devtools-html/debugger.html/issues/6091) 
- Do we need the bold text in the frames? Perhaps the breakpoints should be indented so the checkbox lines up with the file name. I really want to replace the dotted line focus ring with a different styling
- There's some vertical alignment polish that could be done across the sections/accordion headers - even just indenting the Scopes content seems like it would help

digitarald:
@darkwing want to take a shot at it and try to tweak font styles and sizing to make the look a little more coherent? Not sure what will have the biggest impact here, but it seems like a set of low hanging design improvements.

cc @nchevobbe to be aware of reps and line heights consistency.

violasong:
This arrow icon, horizontally flipped, could be used for current frame indicator
https://bugzilla.mozilla.org/show_bug.cgi?id=1479861#c7

Mockup: https://mozilla.invisionapp.com/share/N6K0U23U3ZP

fvsch:
Would be happy to pick it up in the next week(s).

@violasong Do we have a guideline for alignment in the devtools? Things like toolbar height, item height, distance from edges? I believe we should use multiples of 4px as suggested in https://design.firefox.com/photon/visuals/grid.html

For the vertical rythm, Photon recommends 32px rows, but the DevTools seem to use smaller lines:
- 28px for the main tabs, and for some some toolbars (e.g. the Netmonitor's main toolbar)
[...]

PR #7188, opened 2018-10-29
[CallStack] Adjust height for WhyPaused and ExpandRows

### LINKED ISSUE #7136: Allow to show more sidebar information on small viewport heights
On my HD screen the call stacks take a up most of the vertical space:

image

Detailed shot of call stacks show that it does take a lot of estate for the button and paused state; which both could be compressed a bit more:

image

kenjiO:
Fixes #7136

In the Call Stack ui, the `<li>` for each stack frame are 28.5667px high.  I changed the height for the "paused on breakpoint" text from 35px to 28.5667  and changed the "expand rows" button height from 47px to 28.5667 so everything is uniform and about 25px of space is saved.

Before and After:
screen shot 2018-10-29 at 1 27 36 pm

### DIFF
--- a/src/components/SecondaryPanes/Frames/Frames.css
+++ b/src/components/SecondaryPanes/Frames/Frames.css
@@ -91,7 +91,8 @@
   text-align: center;
   border: 1px solid var(--theme-splitter-color);
   background-color: var(--theme-tab-toolbar-background);
-  width: 100%;
+  margin: 0 auto;
+  border-radius:2px;
   font-size: inherit;
   color: inherit;
 }
--- a/src/components/SecondaryPanes/Frames/Frames.css
+++ b/src/components/SecondaryPanes/Frames/Frames.css
@@ -83,12 +83,11 @@
 
 .show-more-container {
   display: flex;
-  height: 28.5667px;
-  padding: 3px 10px;
+  height: 28px;
+  padding: 4px 0;
 }
 
 .show-more {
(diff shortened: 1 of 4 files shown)


kenjiO:
I put in the change in the second commit because I think it makes it more clear that the Expand Rows box is a button.  Feel free to drop it though if you like the original better.
screen shot 2018-10-29 at 1 54 36 pm

kenjiO:
I just noticed in #6631 there is discussion about using 4px multiples for grid sizing to better adhere to photon guidelines.  Let me know if I should re-work this so it uses 4px multiples for spacing and sizing.

darkwing:
Hello @kenjiO !  #6631 takes priority in this case, so this could use some small rework 👍   Thank you!

kenjiO:
I updated the sizing.  Now the "Paused on Breakpoint" div is height 24px normally and expands if there is a message below it.  The div with the "Expand Rows" button is height 28px.

screen shot 2018-10-31 at 3 55 31 pm

darkwing:
Thank you for taking this on @kenjiO !  Any reason for the button style change, with radius?  I want to make sure we stick to branding.

kenjiO:
I thought it looks more like a button when not taking up the whole width of the container.  Also it looks similar to the buttons on the Memory and Performance tools.  I will be glad to revert it back to the original style if you think that looks better.

Issue #7194, opened 2018-10-29
[XHRBreakpoints] Input box disappears when checkbox clicked/unclicked


darkwing:
The input box for XHR breakpoints should always display if there are no XHR breakpoints, just as the Watch Expressions input does.

There's something odd happening where the value of the checkbox effects whether the input displays.

darkwing:
After a bit of experimenting, the reason the XHR breakpoint input doesn't display when there are no breakpoints listed is because there's a "hidden" (for lack of better term) breakpoint for the "Pause on any URL" checkbox, which looks like:

```js
// Checked
Object { path: "", method: "ANY", disabled: false, loading: false, text: "URL contains \"\"" }
```

or:

```js
// Unchecked
Object { path: "", method: "ANY", disabled: true, loading: false, text: "URL contains \"\"" }
```

In the short term we could filter this list on `path` before doing a `xhrBreakpoints.size` check in `renderBreakpoints`; in the longer term I think it would be nice to possibly store that catch-all XHR breakpoint in the function that interacts with the server, so that components downstream don't have that item in the xhrBreakpoints list.

AnshulMalik:
Yeah, that's a hack :P sorry

PR #7196, opened 2018-10-29
Always display the XHR breakpoint input if no explicit XHR breakpoints

darkwing:
References:  #7194 

This ensures text-based XHR breakpoints are checked before choosing not to display the XHR breakpoint input element.

### DIFF
--- a/src/components/SecondaryPanes/XHRBreakpoints.js
+++ b/src/components/SecondaryPanes/XHRBreakpoints.js
@@ -146,14 +146,23 @@ class XHRBreakpoints extends Component<Props, State> {
     }
   };
 
-  renderBreakpoint = ({ path, text, disabled, method }, index) => {
+  renderBreakpoint = breakpoint => {
+    const { path, text, disabled, method } = breakpoint;
     const { editIndex } = this.state;
-    const { removeXHRBreakpoint } = this.props;
+    const { removeXHRBreakpoint, xhrBreakpoints } = this.props;
+
+    // The "pause on any" checkbox
+    if (!path) {
+      return;
+    }
+
+    // Finds the xhrbreakpoint so as to not make assumptions about position
+    const index = xhrBreakpoints.findIndex(
+      bp => bp.path === path && bp.method === method
+    );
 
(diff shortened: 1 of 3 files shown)


src/components/SecondaryPanes/XHRBreakpoints.js
```diff
@@ -179,23 +179,25 @@ class XHRBreakpoints extends Component<Props, State> {
...
-    const { showInput, xhrBreakpoints } = this.props;
+    const { xhrBreakpoints } = this.props;
+    const explicitXhrBreakpoints = xhrBreakpoints.filter(bp => bp.path !== "");
+
     return (
       <ul className="pane expressions-list">
-        {xhrBreakpoints.map(this.renderBreakpoint)}
-        {(showInput || !xhrBreakpoints.size) &&
+        {explicitXhrBreakpoints.map(this.renderBreakpoint)}
+        {explicitXhrBreakpoints.size === 0 &&
```
AnshulMalik: we might want to keep `showInput` for the case when we have few bps and want to add more

darkwing:
As it stands, this PR breaks removal.

src/test/mochitest/browser_dbg-xhr-breakpoints.js
```diff
@@ -4,44 +4,76 @@
...
+add_task(async function() {
+  const dbg = await initDebugger("doc-xhr.html");
+
+  await Promise.all([
+    dbg.actions.togglePauseOnAny(),
+    dbg.actions.setXHRBreakpoint("1", "GET"),
+    dbg.actions.setXHRBreakpoint("2", "GET"),
+    dbg.actions.setXHRBreakpoint("3", "GET"),
+    dbg.actions.setXHRBreakpoint("4", "GET"),
+  ]);
```
jasonLaster: given that we're testing XHR breakpoints pane here, i think it would nice to use the helpers for clicking and typing.

We type in the watch expressions tests as well. Our rule of thumb is test the UI in the few dedicated tests and then use redux actions outside of those tests

PR #7200, opened 2018-10-30
Add tooltip to remove expression

### LINKED ISSUE #7199: Expressions' <CloseButton> should pass a tooltip prop
The `Expressions` component renders the watch expression panel: 

image

You can see there that you can add expressions, and then remove them when you don't need them anymore, with the close button.

But the button itself does not have a title, which means that when you hover on it, it's the title of a parent element which is displayed (the expression one, which displays the expression string): 

image

It would be nice if the button had an actual title that would say "Remove watch expression" for example. 

We're lucky because the `CloseButton` component that we are using here: 

https://github.com/devtools-html/debugger.html/blob/7f49272177728ced8acd26ce9e5ad2cdbd7c444a/src/components/SecondaryPanes/Expressions.js#L248

takes a `tooltip` prop already, that is then used to create the `title` attribute in the DOM: 
https://github.com/devtools-html/debugger.html/blob/7f49272177728ced8acd26ce9e5ad2cdbd7c444a/src/components/shared/Button/CloseButton.js#L20

So, in Expressions.js, we need to pass a `tooltip={"a title"}` prop. Since we want the debugger to be localized, we should use the `L10N` utility, and add an item in the `debugger.properties` file (for example after https://github.com/devtools-html/debugger.html/blob/7f49272177728ced8acd26ce9e5ad2cdbd7c444a/assets/panel/debugger.properties#L512-L515).

jacted:
Fixes #7199 

### Summary of Changes

* Added item to debugger.properties
* Added tooltip prop to CloseButton in `debugger.html/src/components/SecondaryPanes/Expressions.js`

### DIFF
--- a/assets/panel/debugger.properties
+++ b/assets/panel/debugger.properties
@@ -513,6 +513,7 @@ expressions.errorMsg=Invalid expression…
 expressions.label=Add watch expression
 expressions.accesskey=e
 expressions.key=CmdOrCtrl+Shift+E
+expressions.remove.tooltip=Remove watch expression
 
 # LOCALIZATION NOTE (xhrBreakpoints.header): The pause on any XHR breakpoints headings
 xhrBreakpoints.header=XHR Breakpoints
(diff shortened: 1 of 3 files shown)


nchevobbe:
That looks great to me, thanks @jacted !
Let's see if flod has any concerns about the properties file, and then we can merge this

darkwing:
@flodolo Can you weigh in?

darkwing:
Thank you @jacted !

PR #7203, opened 2018-10-30
[WIP] Add a utils to get the l10n object.

### LINKED ISSUE #7132: Do not rely on a global `L10N` properties in components
In many components, we are using L10N as it's defined as a global in the debugger panel.
But since we plan to re-use some of the debugger components in other panels, we should switch to an explicit usage of L10N, by providing it as a prop.

nchevobbe:
Fixes #7132.

This allow us to *not* rely on a global `L10N`
property, which might not exist in some case,
for example if a  component is imported from other
panels than the debugger.

### DIFF
--- a/src/components/SecondaryPanes/Frames/index.js
+++ b/src/components/SecondaryPanes/Frames/index.js
@@ -17,6 +17,7 @@ import renderWhyPaused from "./WhyPaused";
 import actions from "../../../actions";
 import { collapseFrames, formatCopyName } from "../../../utils/pause/frames";
 import { copyToTheClipboard } from "../../../utils/clipboard";
+import { getL10n } from "../../../utils/l10n";
 
 import {
   getFrameworkGroupingState,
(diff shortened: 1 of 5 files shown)


nchevobbe:
This fixes the issues I have when trying to import the `Frames` component in mozilla-central.

Aside from the crappy webpack config, how does that feel ?

webpack.config.js
```diff
@@ -71,6 +71,7 @@ webpackConfig.plugins.push(new ObjectRestSpreadPlugin());
 if (!isProduction) {
   webpackConfig.module = webpackConfig.module || {};
   webpackConfig.module.rules = webpackConfig.module.rules || [];
+  webpackConfig.externals = [/devtools\/shared\/l10n/];
```
jasonLaster: can we update `tranform-mc` to avoid the need for this. I think it is better to do all import logic in one place

src/utils/pause/frames/displayName.js
```diff
@@ -91,7 +92,9 @@ export function formatDisplayName(
     displayName = mapDisplayNames(frame, library);
   }
 
-  return simplifyDisplayName(displayName) || L10N.getStr("anonymousFunction");
```
jasonLaster: perhaps we should not have `L10N` in the util. I'd suggest doing L10N in the component that calls the util. Given that we have many call sites, we might want to have a shared functional component.

src/components/SecondaryPanes/Frames/index.js
```diff
@@ -160,9 +161,10 @@ class Frames extends Component<Props, State> {
   }
 
   renderToggleButton(frames: LocalFrame[]) {
+    const l10n = getL10n();
```
jasonLaster: I think it would be nice to see if we could use context to expose l10n

PR #7206, opened 2018-10-31
Share querystring logic with tabs & breakpoints sidebar 

### LINKED ISSUE #7159: Extend Source Querystring Logic to Tabs and Sidebar
In https://github.com/devtools-html/debugger.html/pull/7087 @ZhangYue-vhyt implemented functionality to show the querystring of a source if there is a sibling source with the same filename within the SourcesTree component.  For example:

- script.js?v=1
- script.js?v=2

That functionality is great for the SourcesTree component but tabs and the right sidebar (breakpoints pane, for example) could use this same logic.  We would ideally have a single function which would check for the presence of sibling sources (and possibly `isOriginal` status) and return a label with querystring included.

WenInCode:
Fixes #7159 

### Summary of Changes
- [x] Share query string logic with tabs
- [x] Share query string logic with breakpoints sidebar

### Test Plan

1. Open the test URL https://shimmer-ostrich.glitch.me/
2. Open debugger.
3. Reload the test URL.
4. Open one of the `script.js` source and observe the query string appended to the name in the tab
5. Add a debugger to the `script.js` and observe the query string appended to the name in the breakpoints sidebar

Mochitests
- [x] Add check for querystring in tab name to `dbg-sources-querystring`
- [x] Add check for querystring in breakpoints side bar in `dbg-sources-querystring`

### Screenshots
screen shot 2018-11-04 at 4 45 44 pm

### Questions/Thoughts
I considered moving `https://github.com/wenincode/debugger.html/blob/7159-source-querystring-logic-tabs-sidebar/src/reducers/sources.js#L434-L449` into a separate `utils` because they aren't actual `reducers` and aren't actually being used by the `reducer` in this file. I _think_ they are in here because the code is a bit tied to the `OuterState` type. I was going to move it and change the types to just be an array of sources to try and decouple it from this file, but the more I work with it the more I see why it is the way it is now. Just seemed a bit odd to have these functions in here.

For the adding logic to sidebar stuff I am thinking of moving [this code] into a sub-component. The reasoning being that I _think_ I could leverage a similar `mapStateToProps` with `OuterState` approach as the other two areas. I am not _entirely_ sure how that all works yet, still pretty new to redux land. I just don't currently see how I can get the right state to pass into those functions in the reducer without moving this code. Just wanted to capture my thoughts before I shut down for the night 😄

### DIFF
--- a/src/components/Editor/Tab.js
+++ b/src/components/Editor/Tab.js
@@ -185,6 +185,7 @@ class Tab extends PureComponent<Props> {
     });
 
     const path = getDisplayPath(source, tabSources);
+    const query = hasSiblingOfSameName ? getSourceQueryString(source) : "";
 
     return (
       <div
@@ -194,15 +195,14 @@ class Tab extends PureComponent<Props> {
         // Accommodate middle click to close tab
         onMouseUp={e => e.button === 1 && closeTab(source)}
         onContextMenu={e => this.onTabContextMenu(e, sourceId)}
-        title={getFileURL(source)}
+        title={getFileURL(source, false)}
       >
         <SourceIcon
           source={source}
           shouldHide={icon => ["file", "javascript"].includes(icon)}
         />
         <div className="filename">
-          {getTruncatedFileName(source)}
-          {hasSiblingOfSameName && getSourceQueryString(source)}
+          {getTruncatedFileName(source, query)}
           {path && <span>{`../${path}/..`}</span>}
         </div>
         <CloseButton
--- a/src/components/Editor/Tab.js
+++ b/src/components/Editor/Tab.js
@@ -22,7 +22,8 @@ import {
   getRawSourceURL,
   getTruncatedFileName,
   getDisplayPath,
-  isPretty
+  isPretty,
+  getSourceQueryString
 } from "../../utils/source";
 import { copyToTheClipboard } from "../../utils/clipboard";
 import { getTabMenuItems } from "../../utils/tabs";
@@ -35,7 +36,6 @@ import {
 } from "../../selectors";
 
 import classnames from "classnames";
-import { parse } from "../../utils/url";
 
 type SourcesList = List<Source>;
 
@@ -185,8 +185,6 @@ class Tab extends PureComponent<Props> {
     });
 
     const path = getDisplayPath(source, tabSources);
-    const query =
-      source && hasSiblingOfSameName ? parse(source.url).search : null;
 
     return (
       <div
@@ -204,7 +202,7 @@ class Tab extends PureComponent<Props> {
         />
         <div className="filename">
           {getTruncatedFileName(source)}
-          {query}
+          {hasSiblingOfSameName && getSourceQueryString(source)}
           {path && <span>{`../${path}/..`}</span>}
         </div>
         <CloseButton
--- a/src/components/Editor/Tab.js
+++ b/src/components/Editor/Tab.js
@@ -30,10 +30,12 @@ import { getTabMenuItems } from "../../utils/tabs";
 import {
   getSelectedSource,
   getActiveSearch,
-  getSourcesForTabs
+  getSourcesForTabs,
+  getHasSiblingOfSameName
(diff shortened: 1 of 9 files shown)


src/components/Editor/Tab.js
```diff
@@ -198,6 +204,7 @@ class Tab extends PureComponent<Props> {
         />
         <div className="filename">
           {getTruncatedFileName(source)}
+          {query}
```
WenInCode: The source tree item stuff [has this in a span](https://github.com/devtools-html/debugger.html/blob/master/src/components/PrimaryPanes/SourcesTreeItem.js#L187), but I wasn't sure if this needed to be. I _think_ it looks better sans span in the tab, but I am open to feedback 😄

src/test/mochitest/browser_dbg-sources-querystring.js
```diff
@@ -28,4 +28,8 @@ add_task(async function() {
     true,
     "simple1.js?x=1 and simple2.jsx=2 exist"
   )
+
+  await clickElement(dbg, "sourceNode", 3);
+  const tab = findElement(dbg, "activeTab");
+  is(tab.innerText, "simple1.js?x=2", "Tab label is simple1.js?x=2");
```
WenInCode: Looks like the order of this can't be guaranteed. Might be better off selecting the node in the tree based on the source name.

src/reducers/sources.js
```diff
@@ -440,6 +440,14 @@ export function getSourcesUrlsInSources(state: OuterState, url: string) {
   return [...new Set(Object.keys(urls).filter(Boolean))];
 }
 
+export function getHasSiblingOfSameName(state: OuterState, source: ?Source) {
```
darkwing: I think this function is in the `SourcesTree` file, so I think we should move it somewhere in `utils/source.js`

darkwing: Ooops, missed the change about it :f ;

darkwing:
This works great!  👍

WenInCode:
I wanted to add a breakpoint to one of the files and assert that the query shows in the panel. I'm having issues getting the breakpoint to show up in the panel when adding it with the helpers.

```
diff --cc src/test/mochitest/browser_dbg-sources-querystring.js
index f381197a,f381197a..6b7067e8
--- a/src/test/mochitest/browser_dbg-sources-querystring.js
+++ b/src/test/mochitest/browser_dbg-sources-querystring.js
@@@ -33,4 -33,4 +33,7 @@@ add_task(async function()
    const activeSourceLabel = getLabel(dbg, 3);
    const tab = findElement(dbg, "activeTab");
    is(tab.innerText, activeSourceLabel, `Tab label is ${activeSourceLabel}`);
++
++  await addBreakpoint(dbg, activeSourceLabel, 4, 0);
++  await waitForever();
  });
```

darkwing:
Hello @WenInCode!  It took me a while to track down but try this:

```js
// Select the source so that the breakpoint displays in sidebar
  const source = findSource(dbg, "simple1.js?x=1");
  await selectSource(dbg, source);
  await addBreakpoint(dbg, "simple1.js?x=1", 6);
  await waitForever();
```

WenInCode:
@darkwing that makes a lot of sense 😂 thanks! I also didn't realize you could just use `selectSource` with the name of the source. I've added the test for the `BreakpointHeading`.

I am going to start looking more into getting the project search to [work with params again](https://github.com/devtools-html/debugger.html/issues/7217). That work can either land in this, or separate, whatever works best 👍

WenInCode:
@darkwing I truncated the long file names 👍 I also noticed that the title attributes for the heading and the tab were not consistent. The heading showed the entire file url, whereas the tab showed a truncated portion which ended up just being part of the query. I updated them to be consistent. Let me know if you think it should be different 😄 

**New:**
screen shot 2018-11-06 at 8 50 21 pm
screen shot 2018-11-06 at 8 50 07 pm

**Old:**
screen shot 2018-11-06 at 8 51 38 pm

PR #7207, opened 2018-10-31
Pass the l10n object through React context.

nchevobbe:
This allow us to not rely on a global L10N variable.


nchevobbe:
Looks quite simple on out-of-debugger side too: 

```diff
diff --git a/devtools/client/shared/components/SmartTrace.js b/devtools/client/shared/components/SmartTrace.js
--- a/devtools/client/shared/components/SmartTrace.js
+++ b/devtools/client/shared/components/SmartTrace.js
@@ -9,7 +9,7 @@ const PropTypes = require("devtools/clie
 const { LocalizationHelper } = require("devtools/shared/l10n");
 
 const l10n = new LocalizationHelper("devtools/client/locales/components.properties");
-
+const debuggerL10n = new LocalizationHelper("devtools/client/locales/debugger.properties");
 const Frames = createFactory(require("devtools/client/debugger/new/src/components/SecondaryPanes/Frames/index").Frames);
 const { annotateFrames } = require("devtools/client/debugger/new/src/utils/pause/frames/annotateFrames");
 
@@ -31,6 +31,10 @@ class SmartTrace extends Component {
     this.state = { hasError: false };
   }
 
+  getChildContext() {
+    return { l10n: debuggerL10n };
+  }
+
   componentWillMount() {
     if (this.props.sourceMapService) {
       this.sourceMapServiceUnsubscriptions = this.props.stacktrace.map((frame, index) => {
@@ -131,4 +135,8 @@ class SmartTrace extends Component {
   }
 }
 
+SmartTrace.childContextTypes = {
+  l10n: PropTypes.object,
+};
+
 module.exports = SmartTrace;
```

---

The only downsides that I see is that we need to pass the l10n object to functions (like `formatDisplayName`, `formatCopyName`, …), and tests need to be modified to explicitly pass the context.

@jasonLaster , how does it look in comparison to https://github.com/devtools-html/debugger.html/pull/7203 ?

src/components/SecondaryPanes/Frames/Frame.js
```diff
@@ -16,11 +18,12 @@ import type { LocalFrame } from "./types";
...
   frame: Frame,
-  options: Object
+  options: Object,
+  l10n: Object
 };
 
-function FrameTitle({ frame, options = {} }: FrameTitleProps) {
-  const displayName = formatDisplayName(frame, options);
+function FrameTitle({ frame, options = {}, l10n }: FrameTitleProps) {
+  const displayName = formatDisplayName(frame, options, l10n);
```
jasonLaster: I'd do something like this

```suggestion
  const displayName = formatDisplayName(frame, options) || l10n.getStr("anonymousFunction");
```

src/components/SecondaryPanes/Frames/Frame.js
```diff
@@ -148,3 +156,4 @@ export default class FrameComponent extends Component<FrameComponentProps> {
 }
 
 FrameComponent.displayName = "Frame";
+FrameComponent.contextTypes = contextTypes;
```
jasonLaster: I'd prefer not drying up the `contextTypes`

1. it is clearer to read the types
2. it is possible we'll have different context types for each component, i.e. does this use shortcuts?

src/components/SecondaryPanes/Frames/Group.js
```diff
@@ -155,3 +156,4 @@ export default class Group extends Component<Props, State> {
 }
 
 Group.displayName = "Group";
+Group.contextTypes = contextTypes;
```
jasonLaster: does group need contextTypes?

nchevobbe: not for now, but soon there should be a localized tooltip there (see https://github.com/devtools-html/debugger.html/issues/7209)

src/components/SecondaryPanes/Frames/Frame.js
jasonLaster: agreed, i guess this is okay...

nchevobbe: not ideal for sure. that's why I did https://github.com/devtools-html/debugger.html/pull/7203 first, seemed a bit more easy to use.

src/components/SecondaryPanes/Frames/Group.js
jasonLaster: lets add it in #7209 then.

nchevobbe:
@jasonLaster I made some changes according to your comments. Does that sound good to be merged?

PR #7219, opened 2018-11-01
Revert "Highlight selected source when switching between panes, auto …

jasonLaster:
this change caused a significant perf regression --

https://treeherder.mozilla.org/perf.html#/comparesubtest?originalProject=mozilla-central&newProject=try&newRevision=4ff4e116841a2158a1b8a2c94d62d097c8965b63&originalSignature=28c4fee1ef7d4e5d184aa628df7c94de208b7678&newSignature=28c4fee1ef7d4e5d184aa628df7c94de208b7678&framework=12&selectedTimeRange=172800


jasonLaster:
@darkwing do you know why this would cause dbg-asm to fail?

darkwing:
Yes and no.  

Yes:  The test fails on a wrong node count because the tree no longer expands all parent directories down to the selected source, thus the files don't display in the tree and the count is wrong.

No:  I don't know why/how this hadn't been failing for a long time.

PR #7226, opened 2018-11-05
[Breakponts] remove immutable.js wrappers

### LINKED ISSUE #7202: [breakpoints] remove immutable wrapper
When we started we used immutable to wrap our data so that we had guarantees that the data would only be updated in a reducer and that we could rely on the data being immutable, which made component updates cheap.

This summer we refactored sources away from immutable because immutable is hard to type check with flow, the api is harder to use, it is slower, and we have gotten good enough with `...` destructuring. We should now refactor breakpoints, which is the second biggest data source.

Steps:

1. change the reducer to be plain objects
2. run `yarn test` or `jest src/actions` and start fixing the unit tests
3. run `yarn flow` and start fixing the flow errors
4. pull out a good amount of hair 😢, ultimately achieve eternal wisdom 🙏 

This is a good amount of work, but will be a big help and teach you some interesting details about breakpoints in the debugger!

---

I started refactoring the reducer here

```diff
diff --git a/src/reducers/breakpoints.js b/src/reducers/breakpoints.js
index a86b5ac52..6cb9168e1 100644
--- a/src/reducers/breakpoints.js
+++ b/src/reducers/breakpoints.js
@@ -19,8 +19,8 @@ import type { XHRBreakpoint, Breakpoint, Location } from "../types";
 import type { Action, DonePromiseAction } from "../actions/types";
 import type { Record } from "../utils/makeRecord";
 
-export type BreakpointsMap = I.Map<string, Breakpoint>;
-export type XHRBreakpointsList = I.List<XHRBreakpoint>;
+export type BreakpointsMap = { [string]: Breakpoint };
+export type XHRBreakpointsList = XHRBreakpoint[];
 
 export type BreakpointsState = {
   breakpoints: BreakpointsMap,
@@ -29,14 +29,12 @@ export type BreakpointsState = {
 
 export function initialBreakpointsState(
   xhrBreakpoints?: any[] = []
-): Record<BreakpointsState> {
-  return makeRecord(
-    ({
-      breakpoints: I.Map(),
-      xhrBreakpoints: I.List(xhrBreakpoints),
-      breakpointsDisabled: false
-    }: BreakpointsState)
-  )();
+): BreakpointsState {
+  return {
+    breakpoints: {},
+    xhrBreakpoints: xhrBreakpoints,
+    breakpointsDisabled: false
+  };
 }
 
 function update(
@@ -148,11 +146,18 @@ function updateXHRBreakpoint(state, action) {
   return state.set("xhrBreakpoints", xhrBreakpoints.set(index, breakpoint));
 }
 
+function setBreakpoint(state, locationId, breakpoint) {
+  return {
+    ...state,
+    breakpoints: { ...state.breakpoints, [locationId]: breakpoint }
+  };
+}
+
 function addBreakpoint(state, action) {
   if (action.status === "start" && action.breakpoint) {
     const { breakpoint } = action;
     const locationId = makeLocationId(breakpoint.location);
-    return state.setIn(["breakpoints", locationId], breakpoint);
+    return setBreakpoint(state, locationId, breakpoint);
   }
 
   // when the action completes, we can commit the breakpoint
@@ -182,20 +187,20 @@ function syncBreakpoint(state, data) {
   }
 
   const locationId = makeLocationId(breakpoint.location);
-  return state.setIn(["breakpoints", locationId], breakpoint);
+  return setBreakpoint(state, locationId, breakpoint);
 }
 
 function updateBreakpoint(state, action) {
   const { breakpoint } = action;
   const locationId = makeLocationId(breakpoint.location);
-  return state.setIn(["breakpoints", locationId], breakpoint);
+  return setBreakpoint(state, locationId, breakpoint);
 }
 
 function updateAllBreakpoints(state, action) {
   const { breakpoints } = action;
   breakpoints.forEach(breakpoint => {
     const locationId = makeLocationId(breakpoint.location);
-    state = state.setIn(["breakpoints", locationId], breakpoint);
+    state = setBreakpoint(state, locationId, breakpoint);
   });
   return state;
}
```

kenjiO:
Fixes #7202

Updates breakpoint reducer to not use immutable.js wrappers.  Also updates anything that uses the reducer and selectors to work with the updated data structures.

### DIFF
--- a/src/actions/breakpoints/index.js
+++ b/src/actions/breakpoints/index.js
@@ -242,7 +242,6 @@ export function setBreakpointCondition(
 
     if (bp.disabled) {
       await dispatch(enableBreakpoint(location));
-      bp.disabled = !bp.disabled;
     }
 
     await client.setBreakpointCondition(
@@ -252,7 +251,7 @@ export function setBreakpointCondition(
       isOriginalId(bp.location.sourceId)
     );
 
-    const newBreakpoint = { ...bp, condition };
+    const newBreakpoint = { ...bp, disabled: false, condition };
 
     assertBreakpoint(newBreakpoint);
 
--- a/src/actions/breakpoints/index.js
+++ b/src/actions/breakpoints/index.js
@@ -13,7 +13,7 @@ import { isOriginalId } from "devtools-source-map";
 import { PROMISE } from "../utils/middleware/promise";
 import {
   getBreakpoint,
-  getBreakpoints,
+  getBreakpointsList,
   getXHRBreakpoints,
   getSelectedSource,
   getBreakpointAtLocation,
@@ -33,7 +33,6 @@ import { isEmptyLineInSource } from "../../reducers/ast";
 
 import type { ThunkArgs, Action } from "../types";
 import type { Breakpoint, Location, XHRBreakpoint } from "../../types";
-import type { BreakpointsMap } from "../../reducers/types";
 
 import { recordEvent } from "../../utils/telemetry";
 
@@ -113,11 +112,11 @@ export function disableBreakpoint(location: Location) {
  */
 export function toggleAllBreakpoints(shouldDisableBreakpoints: boolean) {
   return async ({ dispatch, getState, client }: ThunkArgs) => {
-    const breakpoints = getBreakpoints(getState());
+    const breakpoints = getBreakpointsList(getState());
 
     const modifiedBreakpoints = [];
 
-    for (const [, breakpoint] of breakpoints) {
+    for (const breakpoint of breakpoints) {
       if (shouldDisableBreakpoints) {
         await client.removeBreakpoint(breakpoint.generatedLocation);
         const newBreakpoint: Breakpoint = { ...breakpoint, disabled: true };
@@ -154,18 +153,15 @@ export function toggleAllBreakpoints(shouldDisableBreakpoints: boolean) {
  */
 export function toggleBreakpoints(
   shouldDisableBreakpoints: boolean,
-  breakpoints: BreakpointsMap
+  breakpoints: Breakpoint[]
 ) {
   return async ({ dispatch }: ThunkArgs) => {
-    const promises = breakpoints
-      .valueSeq()
-      .toJS()
-      .map(
-        ([, breakpoint]) =>
-          shouldDisableBreakpoints
-            ? dispatch(disableBreakpoint(breakpoint.location))
-            : dispatch(enableBreakpoint(breakpoint.location))
-      );
+    const promises = breakpoints.map(
+      breakpoint =>
+        shouldDisableBreakpoints
+          ? dispatch(disableBreakpoint(breakpoint.location))
+          : dispatch(enableBreakpoint(breakpoint.location))
+    );
 
     await Promise.all(promises);
   };
@@ -179,9 +175,7 @@ export function toggleBreakpoints(
  */
 export function removeAllBreakpoints() {
   return async ({ dispatch, getState }: ThunkArgs) => {
-    const breakpointList = getBreakpoints(getState())
-      .valueSeq()
-      .toJS();
+    const breakpointList = getBreakpointsList(getState());
     return Promise.all(
       breakpointList.map(bp => dispatch(removeBreakpoint(bp.location)))
     );
@@ -194,18 +188,17 @@ export function removeAllBreakpoints() {
  * @memberof actions/breakpoints
  * @static
  */
-export function removeBreakpoints(breakpoints: BreakpointsMap) {
+export function removeBreakpoints(breakpoints: Breakpoint[]) {
   return async ({ dispatch }: ThunkArgs) => {
-    const breakpointList = breakpoints.valueSeq().toJS();
     return Promise.all(
-      breakpointList.map(bp => dispatch(removeBreakpoint(bp.location)))
+      breakpoints.map(bp => dispatch(removeBreakpoint(bp.location)))
     );
   };
 }
 
 export function remapBreakpoints(sourceId: string) {
   return async ({ dispatch, getState, sourceMaps }: ThunkArgs) => {
-    const breakpoints = getBreakpoints(getState());
+    const breakpoints = getBreakpointsList(getState());
     const newBreakpoints = await remapLocations(
       breakpoints,
       sourceId,
@@ -322,7 +315,7 @@ export function toggleBreakpointsAtLine(line: number, column?: number) {
     const bps = getBreakpointsAtLine(state, line);
     const isEmptyLine = isEmptyLineInSource(state, line, selectedSource.id);
 
-    if (bps.size === 0 && !isEmptyLine) {
+    if (bps.length === 0 && !isEmptyLine) {
       return dispatch(
         addBreakpoint({
           sourceId: selectedSource.id,
--- a/src/reducers/breakpoints.js
+++ b/src/reducers/breakpoints.js
@@ -290,15 +290,15 @@ export function getBreakpointsLoading(state: OuterState): boolean {
 export function getBreakpointsForSource(
   state: OuterState,
   sourceId: string
-): BreakpointsMap {
+): Breakpoint[] {
   if (!sourceId) {
-    return {};
+    return [];
   }
 
   const isGeneratedSource = isGeneratedId(sourceId);
   const breakpoints = getBreakpointsMap(state);
 
-  const breakpointsForSource = {};
+  const breakpointsForSource = [];
 
   for (const key in breakpoints) {
     const bp = breakpoints[key];
@@ -306,7 +306,7 @@ export function getBreakpointsForSource(
       ? bp.generatedLocation || bp.location
       : bp.location;
     if (location.sourceId === sourceId) {
-      breakpointsForSource[key] = { ...bp };
+      breakpointsForSource.push({ ...bp });
     }
   }
 
--- a/src/reducers/breakpoints.js
+++ b/src/reducers/breakpoints.js
@@ -10,16 +10,14 @@
  */
 
 import * as I from "immutable";
-import makeRecord from "../utils/makeRecord";
 
 import { isGeneratedId } from "devtools-source-map";
 import { makeLocationId } from "../utils/breakpoint";
 
 import type { XHRBreakpoint, Breakpoint, Location } from "../types";
 import type { Action, DonePromiseAction } from "../actions/types";
-import type { Record } from "../utils/makeRecord";
 
-export type BreakpointsMap = I.Map<string, Breakpoint>;
+export type BreakpointsMap = { [string]: Breakpoint };
 export type XHRBreakpointsList = I.List<XHRBreakpoint>;
 
 export type BreakpointsState = {
@@ -29,20 +27,18 @@ export type BreakpointsState = {
 
 export function initialBreakpointsState(
   xhrBreakpoints?: any[] = []
-): Record<BreakpointsState> {
-  return makeRecord(
-    ({
-      breakpoints: I.Map(),
-      xhrBreakpoints: I.List(xhrBreakpoints),
-      breakpointsDisabled: false
-    }: BreakpointsState)
-  )();
+): BreakpointsState {
+  return {
+    breakpoints: {},
+    xhrBreakpoints: I.List(xhrBreakpoints),
+    breakpointsDisabled: false
+  };
 }
 
 function update(
-  state: Record<BreakpointsState> = initialBreakpointsState(),
+  state: BreakpointsState = initialBreakpointsState(),
   action: Action
-) {
+): BreakpointsState {
   switch (action.type) {
     case "ADD_BREAKPOINT": {
       return addBreakpoint(state, action);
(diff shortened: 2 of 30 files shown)


kenjiO:
@jasonLaster Can you review the first commit and let me know how this looks so far?  I noticed that most of the selectors use the breakpoints as an array rather than a map.  Would you prefer to store the breakpoints in the redux store as an array rather than a map to avoid converting it to an array?

src/actions/breakpoints/index.js
```diff
@@ -179,9 +176,7 @@ export function toggleBreakpoints(
  */
 export function removeAllBreakpoints() {
   return async ({ dispatch, getState }: ThunkArgs) => {
-    const breakpointList = getBreakpoints(getState())
-      .valueSeq()
-      .toJS();
+    const breakpointList = (Object.values(getBreakpoints(getState())): any);
```
jasonLaster: I think it would be nice to have a selector `getBreakpointList` that returns an array of breakpoints and can be flow typed

src/reducers/breakpoints.js
```diff
@@ -29,20 +29,18 @@ export type BreakpointsState = {
...
-    ({
-      breakpoints: I.Map(),
-      xhrBreakpoints: I.List(xhrBreakpoints),
-      breakpointsDisabled: false
-    }: BreakpointsState)
-  )();
+): BreakpointsState {
+  return {
+    breakpoints: {},
+    xhrBreakpoints: I.List(xhrBreakpoints),
```
jasonLaster: we should probably leave this as immutable for now as that might be its own PR

jasonLaster:
one lint error and needs a rebase. Lets get this in! :)

src/actions/breakpoints/index.js
```diff
@@ -113,11 +113,11 @@ export function disableBreakpoint(location: Location) {
  */
 export function toggleAllBreakpoints(shouldDisableBreakpoints: boolean) {
   return async ({ dispatch, getState, client }: ThunkArgs) => {
-    const breakpoints = getBreakpoints(getState());
+    const breakpoints = getBreakpointsArray(getState());
```
jasonLaster: this is a nit, but i think I prefer `getBreakpointsList` as it is more descriptive than implementation specific

src/actions/breakpoints/index.js
```diff
@@ -157,15 +157,12 @@ export function toggleBreakpoints(
...
-    const promises = breakpoints
-      .valueSeq()
-      .toJS()
-      .map(
-        ([, breakpoint]) =>
-          shouldDisableBreakpoints
-            ? dispatch(disableBreakpoint(breakpoint.location))
-            : dispatch(enableBreakpoint(breakpoint.location))
-      );
+    const promises = (Object.values(breakpoints): any).map(
```
jasonLaster: could we use `getBreakpointsArray`?

src/actions/breakpoints/index.js
```diff
@@ -196,7 +191,7 @@ export function removeAllBreakpoints() {
  */
 export function removeBreakpoints(breakpoints: BreakpointsMap) {
   return async ({ dispatch }: ThunkArgs) => {
-    const breakpointList = breakpoints.valueSeq().toJS();
+    const breakpointList = (Object.values(breakpoints): any);
```
jasonLaster: could we use `getBreakpointsArray`?

src/reducers/breakpoints.js
```diff
@@ -118,12 +114,17 @@ function addXHRBreakpoint(state, action) {
   );
 
   if (existingBreakpointIndex === -1) {
-    return state.set("xhrBreakpoints", xhrBreakpoints.push(breakpoint));
+    // TODO make this immutable?
+    return {
```
jasonLaster: it is okay to have mutable state as long as you copy it and ensure that it does not change. We ensure this with source records via flow 

e.g. `type Source = {| +a: 2 |}; let s: Source = {a: 2}`. Now `a` is read only :)

src/reducers/breakpoints.js
```diff
@@ -214,60 +246,73 @@ function remapBreakpoints(state, action) {
...
 }
 
 // Selectors
 // TODO: these functions should be moved out of the reducer
 
-type OuterState = { breakpoints: Record<BreakpointsState> };
+type OuterState = { breakpoints: BreakpointsState };
 
-export function getBreakpoints(state: OuterState) {
+// TODO make this immutable?
```
jasonLaster: i don't think it needs to be immutable as long as we prevent writes via flow

src/reducers/breakpoints.js
```diff
@@ -214,60 +246,73 @@ function remapBreakpoints(state, action) {
...
+export function getBreakpointsMap(state: OuterState): BreakpointsMap {
   return state.breakpoints.breakpoints;
 }
 
+export function getBreakpointsArray(state: OuterState): Breakpoint[] {
```
jasonLaster: I think this would read a bit nicer:

```js
 export function getBreakpointsList(state: OuterState): Breakpoint[] { 
```

src/reducers/breakpoints.js
```diff
@@ -214,60 +246,73 @@ function remapBreakpoints(state, action) {
...
   }
 
   const isGeneratedSource = isGeneratedId(sourceId);
-  const breakpoints = getBreakpoints(state);
+  const breakpoints = getBreakpointsMap(state);
 
-  return breakpoints.filter(bp => {
+  const breakpointsForSource = {};
+  const keys = Object.keys(breakpoints);
+  keys.forEach(key => {
```
jasonLaster: i think it might be nicer to do `for (const key in breakpoints)`

src/reducers/tests/breakpoints.spec.js
```diff
@@ -11,20 +11,66 @@ import {
...
-  return state.set("breakpoints", mappedData);
+  state.breakpoints = data;
+  return state;
 }
 
 describe("Breakpoints Selectors", () => {
   it("it gets a breakpoint for an original source", () => {
     const sourceId = "server1.conn1.child1/source1/originalSource";
-    const matchingBreakpoints = { id1: { location: { sourceId } } };
+    const matchingBreakpoints = {
```
jasonLaster: why is this so different?

src/selectors/breakpointSources.js
```diff
@@ -67,10 +66,7 @@ function findBreakpointSources(
   breakpoints: BreakpointsMap
 ): Source[] {
   const sourceIds: string[] = uniq(
-    breakpoints
-      .valueSeq()
-      .map(bp => getSelectedLocation(bp, selectedSource).sourceId)
-      .toJS()
+    (Object.values(breakpoints): any).map(bp => bp.location.sourceId)
```
jasonLaster: can we use the new selector?

src/test/mochitest/browser_dbg-editor-gutter.js
```diff
@@ -38,20 +38,20 @@ add_task(async function() {
...
   await waitForDispatch(dbg, "ADD_BREAKPOINT");
-  is(getBreakpoints(getState()).size, 1, "One breakpoint exists");
+  is(getBreakpointsArray(getState()).length, 1, "One breakpoint exists");
   assertEditorBreakpoint(dbg, 4, true);
 
   // Make sure clicking at the same place removes the icon.
   clickGutter(dbg, 4);
   await waitForDispatch(dbg, "REMOVE_BREAKPOINT");
-  is(getBreakpoints(getState()).size, 0, "No breakpoints exist");
+  is(getBreakpointsArray(getState()).length, 0, "No breakpoints exist");
```
jasonLaster: I see this cropping up a lot, should we have a selector `getBreakpointCount`?

src/actions/breakpoints/index.js
kenjiO: It looks like the code that calls this function is already passing in an array instead of a map so I can just change the type and drop the Object.values() call

src/reducers/tests/breakpoints.spec.js
kenjiO: flow complained about all the missing properties so I added them

jasonLaster:
Looks great. If it's green, I'll do some QA testing and fully expect to press the green button tomorrow.

jasonLaster:
It would be nice to update the Flow types for Breakpoints now as well and make sure we *actually* are immutable. I'd be nervous landing this without some guarantees that we have flow coverage to make sure we don't accidentally mutate outside of the reducers.

```diff
diff --git a/src/types.js b/src/types.js
index 961b136fa..aebdc18a1 100644
--- a/src/types.js
+++ b/src/types.js
@@ -93,31 +93,31 @@ export type ASTLocation = {|
  * @memberof types
  * @static
  */
-export type Breakpoint = {
-  id: BreakpointId,
-  location: Location,
-  astLocation: ?ASTLocation,
-  generatedLocation: Location,
-  loading: boolean,
-  disabled: boolean,
-  hidden: boolean,
-  text: string,
-  originalText: string,
-  condition: ?string
-};
+export type Breakpoint = {|
+  +id: BreakpointId,
+  +location: Location,
+  +astLocation: ?ASTLocation,
+  +generatedLocation: Location,
+  +loading: boolean,
+  +disabled: boolean,
+  +hidden: boolean,
+  +text: string,
+  +originalText: string,
+  +condition: ?string
+|};

 /**
  * XHR Breakpoint
  * @memberof types
  * @static
  */
-export type XHRBreakpoint = {
-  path: string,
-  method: "GET" | "POST" | "DELETE" | "ANY",
-  loading: boolean,
-  disabled: boolean,
-  text: string
-};
+export type XHRBreakpoint = {|
+  +path: string,
+  +method: "GET" | "POST" | "DELETE" | "ANY",
+  +loading: boolean,
+  +disabled: boolean,
+  +text: string
+|};
```

src/selectors/visibleBreakpoints.js
```diff
@@ -46,8 +45,8 @@ function isVisible(breakpoint, selectedSource) {
   */
 export const getVisibleBreakpoints = createSelector(
```
kenjiO: Do you know how to add a flow type to getVisibleBreakpoints?  I think the issue with the "disable other breakpoints" has to do with a Breakpoint being compared to a FormattedBreakpoint from this selector that gets a false negative.

jasonLaster: hmm, i forgot. what do we we do with other memozed selectors?

jasonLaster:
@kenjiO i just updated this PR w/ some small tweaks and squashed your commits down.

src/reducers/tests/breakpoints.spec.js
```diff
@@ -11,20 +11,24 @@ import {
...
+      id1: createBreakpoint({ line: 1, sourceId: sourceId })
+    };
+
     const otherBreakpoints = {
-      id2: { location: { sourceId: "not-this-source" } }
+      id2: createBreakpoint({ line: 1, sourceId: "not-this-source" })
```
jasonLaster: this is a fun trick to keep the fixture data lean. We do the same thing for sources in some other tests

kenjiO: Cool, I will update the tests with that

kenjiO:
RE: 8f47451, can we add these '+' to the types?  I tried adding the '|' to the types but that got a lot of flow errors.  Maybe save that for another PR.  I still haven't fixed the issue with 'disable other breakpoints' yet so please don't merge it unless you fixed it.

jasonLaster:
> disable other breakpoints

thanks for the heads up. I definitely want to land this oon.

> other `+` 

thanks!

jasonLaster:
> disable other breakpoints

so i think this is a big enough change, where we likely will have some small bugs. Lets land this and then fix the disable other breakpoints bug. Maybe add a unit test for it as well.

kenjiO:
@jasonLaster I think this should fix it.  I'll work on some tests for it.  Should I do jest tests, mochi tests or both?

```diff
diff --git a/src/components/SecondaryPanes/Breakpoints/BreakpointsContextMenu.js b/src/components/SecondaryPanes/Breakpoints/BreakpointsContextMenu.js
index 510ef51a..09b7bd42 100644
--- a/src/components/SecondaryPanes/Breakpoints/BreakpointsContextMenu.js
+++ b/src/components/SecondaryPanes/Breakpoints/BreakpointsContextMenu.js
@@ -4,6 +4,20 @@
 
 import { buildMenu, showMenu } from "devtools-contextmenu";
 
+/**
+ * Check if a Breakpoint type object and a FormattedBreakpoint type object
+ * are the same breakpoint
+ */
+function checkIfSameBreakpoint(breakpoint, formattedBreakpoint) {
+  const location1 = breakpoint.location;
+  const location2 = formattedBreakpoint.selectedLocation;
+  return (
+    location1.sourceId === location2.sourceId &&
+    location1.line === location2.line &&
+    location1.column === location2.column
+  );
+}
+
 export default function showContextMenu(props) {
   const {
     removeBreakpoint,
@@ -74,14 +88,16 @@ export default function showContextMenu(props) {
     "breakpointMenuItem.addCondition2.accesskey"
   );
 
-  const otherBreakpoints = breakpoints.filter(b => b !== breakpoint);
+  const otherBreakpoints = breakpoints.filter(
+    b => !checkIfSameBreakpoint(b, breakpoint)
+  );
   const enabledBreakpoints = breakpoints.filter(b => !b.disabled);
   const disabledBreakpoints = breakpoints.filter(b => b.disabled);
   const otherEnabledBreakpoints = breakpoints.filter(
-    b => !b.disabled && b !== breakpoint
+    b => !b.disabled && !checkIfSameBreakpoint(b, breakpoint)
   );
   const otherDisabledBreakpoints = breakpoints.filter(
-    b => b.disabled && b !== breakpoint
+    b => b.disabled && !checkIfSameBreakpoint(b, breakpoint)
   );
 
   const deleteSelfItem = {
```

jasonLaster:
lets move this patch to a new WIP pr.

My first thoughts are you could use `bp.id` to compare values. Also, why does `breakpoint != b` fail? we should be returning the same breakpoint? 

Lines like [this](https://github.com/devtools-html/debugger.html/blob/master/src/reducers/breakpoints.js#L309-L310) where we are creating new objects when nothing changed are probably a mistake. Actually, it could be a big perf regression if this means components will start re-rendering because they think its a new object :/

jasonLaster:
it looks like the breakpoint patch has a performance regression... do you have some time to dig in and investigate?

kenjiO:
@jasonLaster Yes, I will look into it.  What tool do you reccommend to measure peformance and is there a place I kind find more details about the performance regression?

jasonLaster:
>  What tool do you reccommend to measure peformance

http://perf-html.io/ and chrome devtools performance

> is there a place I kind find more details about the performance regression?

not really...

PR #7232, opened 2018-11-05
Show column breakpoints when line breakpoint is set, remove Alt/Option key dependency

audreyfang:
### Summary of Changes

* Pressing Option/Alt only shows column breakpoints on lines that have a breakpoint already set.

column-breakpoints

### DIFF
--- a/src/components/Editor/CallSites.js
+++ b/src/components/Editor/CallSites.js
@@ -142,17 +142,34 @@ class CallSites extends Component {
       });
     }
   }
+  filterCallSites() {
+    const { callSites, breakpoints } = this.props;
+
+    const breakpointLines = new Set();
+    if (breakpoints._root !== undefined) {
+      for (const key in breakpoints._root.entries) {
+        breakpointLines.add(breakpoints._root.entries[key][1].location.line);
+      }
+    }
+
+    return callSites.filter(({ location }) =>
+      breakpointLines.has(location.start.line)
+    );
+  }
 
   render() {
     const { editor, callSites, selectedSource } = this.props;
     const { showCallSites } = this.state;
+
     let sites;
     if (!callSites) {
       return null;
     }
 
+    const callSitesFiltered = this.filterCallSites();
+
     editor.codeMirror.operation(() => {
-      const childCallSites = callSites.map((callSite, index) => {
+      const childCallSites = callSitesFiltered.map((callSite, index) => {
         const props = {
           key: index,
           callSite,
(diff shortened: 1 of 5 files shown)


darkwing:
Hello @audreyfang !  Could you update this with next steps?

src/components/Editor/CallSites.js
```diff
@@ -142,17 +142,34 @@ class CallSites extends Component {
       });
     }
   }
+  filterCallSites() {
```
darkwing: I'd love a more specific name on this function; i.e. filtered by what?  Additionally, a comment above the function as to why the given filter is there would be nice.

darkwing:
After a bit of quick testing:

-  I'd love if we removed the crosshair cursor; it doesn't seem applicable anymore
- When using https://firefox-dev.tools/debugger-examples/examples/todomvc/app.js as my test file, I add a breakpoint to line 25 by clicking the line number, then press `option` to view the possible column breakpoint locations.  I click the column breakpoint icon and the column breakpoint is added (yay!).  I then click the column breakpoint again to remove it, and the line breakpoint arrow (the blue one that displays on the line number) is removed; we'll want to prevent that.

This is looking awesome.  Great work!

src/components/Editor/CallSites.js
```diff
@@ -142,17 +142,34 @@ class CallSites extends Component {
...
+    const { callSites, breakpoints } = this.props;
+
+    const breakpointLines = new Set();
+    if (breakpoints._root !== undefined) {
```
loganfsmyth: Naming conventions in JS generally treat `_` as an indication that something is private. In this case, the specifics of `_root` are internal to the `breakpoints` object and shouldn't be used directly. The `breakpoints` object is an `Immutable.js` `Map` object, so if you peak at the docs, you've for all sorts of options. https://facebook.github.io/immutable-js/docs/#/Map

In this case, I'd probably do

```
const breakpointLines = new Set(breakpoints.toIndexedSeq().map(bp => bp.location.line));
```
which takes all the values and discards the keys, and then converts each value into its line number.

audreyfang: Thanks for pointing me in the right direction for this! I'll definitely take a look and clean this up.

audreyfang:
@darkwing In terms of next steps, I wanted a bit of feedback before I did a some more extensive testing. I've definitely noticed some issues like the one you mentioned with removing column breakpoints.

PR #7237, opened 2018-11-06
Enable line-highlighting for debugger editor

darkwing:
I can't decided if I like this or not.  I like the additional attention on the selected line, but when we look at interaction, we don't allow text modification so I don't know that an editor-like highlight feels right. Thoughts?

line-highlight


jasonLaster:
I find it a bit distracting personally. For the most part, the current cursor location doesn't really matter when your debugging.

Issue #7160, opened 2018-10-25
[SourcesTree] Sort SourcesTree when Querystring present


darkwing:
@ZhangYue-vhyt implemented showing querystrings for items in the sources tree (https://github.com/devtools-html/debugger.html/pull/7087) when there are sources with sibling filenames of the same name. 

At present the SourcesTree only sorts by filename and not by querystring, so we'll need to update the logic in the tree to reliably sort by file + querystring value.

PR #7238, opened 2018-11-06
Sort SourcesTree when Querystring present (#7160)

ZhangYue-vhyt:
Not done #7160
 ## What I did
Display the files with query strings by ascending order in the source tree.

## Test steps
1. Open Test URL: https://davidwalsh.name/demo/devtools-qstring-contribute.php.
2. Open the debugger.
3. Reload the test URL.
4. Reload the debugger.
5. Look at the source tree.

## Issue
- The query strings of all files have the same name will be changed if one of them is selected.
- Failed on breakpoint test.


darkwing:
This was really awesome work and research @ZhangYue-vhyt !  I think for now we need to table this feature for now; your implementation works pretty well but I'm worried about the performance ramifications of all of this sorting.  We'd then also want to sort sources in the right pane, and I'm worried that we need a higher level of sorting overall.  This was well worth the time though; thank you @ZhangYue-vhyt !

ZhangYue-vhyt:
> This was really awesome work and research @ZhangYue-vhyt ! I think for now we need to table this feature for now; your implementation works pretty well but I'm worried about the performance ramifications of all of this sorting. We'd then also want to sort sources in the right pane, and I'm worried that we need a higher level of sorting overall. This was well worth the time though; thank you @ZhangYue-vhyt !

Thank you, David. Can you review the diff of `src\reducers\sources.js`? I'm not sure does the old function return the same array as the new function. I guess the check of the same pathname of urls is required.

Issue #7241, opened 2018-11-06
[Tracking] Column Breakpoints 


darkwing:
Supersedes #7028.

[Server-side Meta Bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1528417).

Column breakpoints are currently flagged as `devtools.debugger.features.column-breakpoints`.  Once we turn them on, there are a number of tasks remaining:

## MVP
- [x] Update the marker logic to only display if the column breakpoint location is within the viewport.
- [x] Performance: Switching between tabs is painfully slow when column breakpoints are enabled.
- [x] Disable column breakpoints for pretty printed files and original sources with line breakpoints (https://github.com/devtools-html/debugger.html/issues/7248)

## Polish
- [x] When clicking on the line in the gutter to add a breakpoint, add the breakpoint to the first possible column, not just `0` or `undefined` (*Status*: PR:  https://github.com/devtools-html/debugger.html/pull/7392)
- [x] Ensure that the breakpoint symbol on the editor line isn't removed when a column breakpoint is deactivated, unless the only activated column breakpoint is removed.  (*Status*: PR:  https://github.com/devtools-html/debugger.html/pull/7410)
- [x] Add a context menu to column breakpoints   (*Status*: Experiment:  https://github.com/darkwing/debugger.html/tree/experiment-column-context)
- [x] Ensure that jumping to a pause location still scrolls to the right location.  (*Status*: None) 
- [x] Implement new UI
- [x] Memoize getCallSites (*Status*: None; *Bug:* https://github.com/devtools-html/debugger.html/issues/7309)
- [x] Enrich the gutter and breakpoint menus to include "Disable all on line", "Enable all on line", "Remove all on line" (*Status*: None)

## Todos

- [x] add server integration for breakable positions
- [x] Review how breakpoint sync behaves as files with column breakpoints change to make sure that still works well
- [x] should be able to preview
- [x]  enable column breakpoints for the tests
- [x]  enable column breakpoints
- [x] update selected location support #8025 
- [x] load breakpoint positions when a source is loaded
- [x] finish column bps test [pr]
- [x] review [Critical Path Interactions][cpi]
- [x] Resolve onSources exception: #8106
- [ ] Breakpoints in HTML files with more than one `<script>` are broken: [bugzilla 1535246](https://bugzilla.mozilla.org/show_bug.cgi?id=1535246)

### Polish
[...]

Issue #6592, opened 2018-07-03
[PrettyPrint] Improve location quality


jasonLaster:
There are a couple of issues with how we handle locations when prettyprinting

- [x] pauses without a bp on load #5519 #5827 🥇
- [x] jsx sets a breakpoint on the line above #6685
- [ ] selectLocation to do less work with regard to pretty print #5618 
- [ ] prettifying should remember the user intent to jump to a specific line #3778 
- [ ] breakpoints are not persisted #3532

janodvarko:
Most of the bugs are closed (except of #5618)

Closing

Honza

PR #7246, opened 2018-11-07
Don't show column breakpoints for pretty files

darkwing:
References:  #7241


AnshulMalik:
Do we really want to disable column breakpoints of pretty files?

darkwing:
We found them to be unreliable.

digitarald:
@darkwing I assume this will work correctly and can be offered when #6592 is addressed?

PR #7256, opened 2018-11-09
Update yarn-copy so that assets are accurately created

jasonLaster:
### Summary of Changes
- this updates yarn copy so that the next release uses the new images format
- `v=2` was a silly hack so that 2 SVGs could be in debugge/new/images and devtools/client/themes even though they are the same
- `updateManifest` will copy the manifest over so that we don't get crazy bundle changes when we run `yarn copy`


jasonLaster:
still a work in progress...

PR #7257, opened 2018-11-09
Add a backport script

jasonLaster:
Now that we have original sources in the debugger, it should be easier to backport changes in m-c for a given sha back to github.

This script looks at a commit, filters out the files outside of the debugger, and copies the others back to github. 

I'm sure that this can be improved:

1. we will now copy any changes including dist changes
2. we will miss changes from reps...
3. we could also write a commit message if we want
4. this assumes you're using hg... we can generalize this as well...


jasonLaster:
merging so that it's on *master* and we can easily iterate. This is definitely WIP.

PR #7262, opened 2018-11-12
Support destructuring assignments

### LINKED ISSUE #7222: Top level await should support destructuring assignments
We should gracefully support this kind of syntax.

input:
```js
const {a,b} = await {a: 2, b:3}
```

output:
```js
(async () => { 
  const __await1__ = {a:2, b: 3}; 
  self.a = __await1__; 
  self.b = __await1__;
})();
```

P0lip:
Fixes #7222

Here's the Pull Request Doc

### Summary of Changes

* Added support for simple top-level destructuring assignments - ~still working on the nested destructuring with default values~ (already implemented). The change affects all cases, not just the ones with await statements.
* Built an IIFE with `@babel/types` in mapAwaitExpression.js.

### Test Plan

Just by writing tests. No specific test plan.

Feel free to improve it!

### DIFF
--- a/src/workers/parser/mapBindings.js
+++ b/src/workers/parser/mapBindings.js
@@ -16,6 +16,61 @@ function getIdentifierName(identifier, bindings) {
     : t.memberExpression(t.identifier("self"), identifier);
 }
 
+function destructurize(body, bindings, baseObj, property) {
+  let name = property.value;
+  let value = null;
+
+  if (t.isAssignmentPattern(property.value)) {
+    name = property.value.left;
+
+    value = t.conditionalExpression(
+      t.binaryExpression(
+        "===",
+        t.memberExpression(baseObj, property.key),
+        t.identifier("undefined")
+      ),
+      property.value.right,
+      t.memberExpression(baseObj, property.key)
+    );
+  }
+
+  if (t.isIdentifier(name)) {
+    body.push(
+      t.assignmentExpression(
+        "=",
+        getIdentifierName(name, bindings),
+        value === null ? t.memberExpression(baseObj, property.key) : value
+      )
+    );
+  } else if (t.isObjectPattern(name)) {
+    const identifier =
+      value === null
+        ? baseObj
+        : t.identifier(
+            `${baseObj.object ? baseObj.object.name : baseObj.name}${
+              property.key.name
+            }__`
+          );
+
+    if (value !== null) {
+      body.push(
+        t.variableDeclaration("let", [t.variableDeclarator(identifier, value)])
+      );
+    }
+
+    for (const childProperty of name.properties) {
+      destructurize(
+        body,
+        bindings,
+        value === null
+          ? t.memberExpression(identifier, property.key)
+          : identifier,
+        childProperty
+      );
+    }
+  }
+}
+
 // translates new bindings `var a = 3` into `self.a = 3`
 // and existing bindings `var a = 3` into `a = 3` for re-assignments
 function globalizeDeclaration(node, bindings) {
@@ -30,15 +85,7 @@ function globalizeDeclaration(node, bindings) {
       );
 
       for (const property of declaration.id.properties) {
-        const identifier = getIdentifierName(property.value, bindings);
-
-        body.push(
-          t.assignmentExpression(
-            "=",
-            identifier,
-            t.memberExpression(varIdentifier, property.key)
-          )
-        );
+        destructurize(body, bindings, varIdentifier, property);
       }
     } else {
       const identifier = getIdentifierName(declaration.id, bindings);
@@ -79,12 +126,6 @@ function replaceNode(ancestors, node) {
   }
 }
 
-function hasNestedDestructuring(node) {
-  return node.declarations.some(declaration =>
-    t.isObjectPattern(declaration.value)
-  );
-}
-
 export default function mapExpressionBindings(
   expression: string,
   bindings: string[] = []
@@ -94,6 +135,11 @@ export default function mapExpressionBindings(
   let shouldUpdate = true;
 
   t.traverse(ast, (node, ancestors) => {
+    // the isMapped check is to avoid processing the newly added nodes
+    if (isMapped) {
+      return;
+    }
+
     const parent = ancestors[ancestors.length - 1];
 
     if (t.isWithStatement(node)) {
@@ -122,13 +168,7 @@ export default function mapExpressionBindings(
       return;
     }
 
-    if (hasNestedDestructuring(node)) {
-      shouldUpdate = false;
-      return;
-    }
-
-    // the isMapped check is to avoid processing the newly added nodes
-    if (!t.isForStatement(parent.node) && !isMapped) {
+    if (!t.isForStatement(parent.node)) {
       const newNodes = globalizeDeclaration(node, bindings);
       isMapped = true;
       replaceNode(ancestors, newNodes);
--- a/src/workers/parser/mapBindings.js
+++ b/src/workers/parser/mapBindings.js
@@ -10,18 +10,48 @@ import { isTopLevel } from "./utils/helpers";
 import generate from "@babel/generator";
 import * as t from "@babel/types";
 
+function getIdentifierName(identifier, bindings) {
+  return bindings.includes(identifier.name)
+    ? identifier
+    : t.memberExpression(t.identifier("self"), identifier);
+}
+
 // translates new bindings `var a = 3` into `self.a = 3`
 // and existing bindings `var a = 3` into `a = 3` for re-assignments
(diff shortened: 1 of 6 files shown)


src/workers/parser/tests/mapExpression.spec.js
```diff
@@ -178,6 +281,19 @@ describe("mapExpression", () => {
         bindings: false,
         originalExpression: false
       }
+    },
+    {
+      name: "bindings + destructuring without mappings",
+      expression: "({ a } = { a: 4 })",
+      newExpression: "({ a } = {\n a: 4 \n})",
```
P0lip: The following test case fails.
Shall we map the assignment here? I assume we want to have `({ a: _a }) = { a: 4 })`, but `bindings without mapping` does not seem to work as expected, as it does not map anything.

src/workers/parser/mapBindings.js
```diff
@@ -49,8 +79,10 @@ function replaceNode(ancestors, node) {
   }
 }
 
-function hasDestructuring(node) {
-  return node.declarations.some(declaration => t.isPattern(declaration.id));
+function hasNestedDestructuring(node) {
```
P0lip: Just a temporary change to make sure we don't process nested destructuring for now.

src/workers/parser/mapBindings.js
```diff
@@ -90,12 +122,13 @@ export default function mapExpressionBindings(
...
 
-    if (hasDestructuring(node)) {
+    if (hasNestedDestructuring(node)) {
       shouldUpdate = false;
       return;
     }
 
-    if (!t.isForStatement(parent.node)) {
+    // the isMapped check is to avoid processing the newly added nodes
+    if (!t.isForStatement(parent.node) && !isMapped) {
```
P0lip: Happy to rename/add new variable or slightly adjust the way we iterate.
Currently, the newly inserted/replaced nodes are iterated again, so
`let __decl1__ = x;` is transformed to `self.__decl1__ = x`;

P0lip: Moved the condition to the top. I suppose we will want to adjust it, as the check does might be confusing.

src/workers/parser/tests/mapExpression.spec.js
```diff
@@ -140,6 +285,20 @@ describe("mapExpression", () => {
         originalExpression: false
       }
     },
+    {
+      name: "declaration + destructuring",
+      expression: "var { a } = { a: 3 };",
+      newExpression: `let __decl0__ = {\n a: 3 \n}
```
P0lip: Should we wrap the code in a block statement?

src/workers/parser/mapBindings.js
```diff
@@ -62,6 +135,11 @@ export default function mapExpressionBindings(
   let shouldUpdate = true;
 
   t.traverse(ast, (node, ancestors) => {
+    // the isMapped check is to avoid processing the newly added nodes
+    if (isMapped) {
```
loganfsmyth: This isn't a safe way to bail because we may well want to perform more changes, even if part of the AST has been mapped. Ideally the newly-added nodes would be simplified enough that they wouldn't trigger further processing. Do you have example of what this is trying to avoid?

src/workers/parser/mapBindings.js
```diff
@@ -49,10 +126,6 @@ function replaceNode(ancestors, node) {
   }
 }
 
-function hasDestructuring(node) {
-  return node.declarations.some(declaration => t.isPattern(declaration.id));
```
loganfsmyth: As it is, this logic will still fail with array patterns, so if we do end up going with this pattern, we'll still want to bail out here for array patterns, just not object patterns.

src/workers/parser/mapBindings.js
```diff
@@ -10,18 +10,95 @@ import { isTopLevel } from "./utils/helpers";
...
+
 // translates new bindings `var a = 3` into `self.a = 3`
 // and existing bindings `var a = 3` into `a = 3` for re-assignments
 function globalizeDeclaration(node, bindings) {
-  return node.declarations.map(declaration => {
-    const identifier = bindings.includes(declaration.id.name)
-      ? declaration.id
-      : t.memberExpression(t.identifier("self"), declaration.id);
+  return node.declarations.reduce((body, declaration, i) => {
+    if (t.isPattern(declaration.id)) {
```
loganfsmyth: Clearly a lot of work went into this, but unless I'm missing something, I don't think we strictly need to do so much to implement this. For instance, for
```
expression: "const { a, c: y } = await b()",
newExpression: formatAwait(`
  let __decl0__ = await b();
  self.a = __decl0__.a;
  return (self.y = __decl0__.c);
`)
```
we could do
```
expression: "const { a, c: y } = await b()",
newExpression: formatAwait(`
  ({ a: self.a, c: self.y } = await b());
`)
```
to avoid re-implementing object destructuring, and also allow us to handle array pattern destructuring.

P0lip: Well, I just adapted the solution to the proposed output. Frankly speaking, I had a similar feeling during implementation. Nonetheless, I changed the implementation to preserve destructuring, as the code is way cleaner now and we can easily support array destructuring.

src/workers/parser/mapBindings.js
P0lip: Support for array destructuring has been added.

src/workers/parser/mapBindings.js
P0lip: Well, basically in my case it covered the new declarations, such as `let __decl{index}__` and transformed it to `self.__decl{index}__`.
I applied the suggested changes and that check is not required anymore.

src/workers/parser/mapBindings.js
```diff
@@ -10,20 +10,51 @@ import { isTopLevel } from "./utils/helpers";
...
 // translates new bindings `var a = 3` into `self.a = 3`
 // and existing bindings `var a = 3` into `a = 3` for re-assignments
 function globalizeDeclaration(node, bindings) {
   return node.declarations.map(declaration => {
-    const identifier = bindings.includes(declaration.id.name)
-      ? declaration.id
-      : t.memberExpression(t.identifier("self"), declaration.id);
+    if (t.isPattern(declaration.id)) {
+      return t.expressionStatement(
+        t.assignmentExpression("=", declaration.id, declaration.init)
```
loganfsmyth: I think ideally we'd want to call `globalizeObjectDestructuring` and `globalizeArrayDestructuring` from here as a general
```
t.assignmentExpression("=", globalizeDestructuring(declaration.id), declaration.init)
```
otherwise we're temporarily putting the AST into a broken state, since
```
var {a: self.a } = {}
```
is not valid syntax. It would also make things easier to follow in my opinion.

loganfsmyth: We could also potentially instead change the `getIdentifierName(declaration.id, bindings);` below to `getAssignmentTarget(declaration.id, bindings)` what would automatically return the right thing to assign to instead.

src/workers/parser/mapBindings.js
```diff
@@ -80,18 +107,22 @@ export default function mapExpressionBindings(
...
 
-      if (t.isPattern(node.left)) {
-        shouldUpdate = false;
-        return;
-      }
+      return;
     }
 
-    if (!t.isVariableDeclaration(node)) {
+    if (t.isObjectPattern(node)) {
```
loganfsmyth: As mentioned above, I think we'd want to handle these explicitly in `globalizeAssignment` and `globalizeDeclaration` rather than relying on traversal to reach them.

src/workers/parser/mapBindings.js
```diff
@@ -10,20 +10,51 @@ import { isTopLevel } from "./utils/helpers";
...
       t.assignmentExpression("=", identifier, declaration.init)
     );
   });
 }
 
+function globalizeArrayDestructuring(node, bindings) {
+  for (const [i, element] of node.elements.entries()) {
+    if (t.isIdentifier(element)) {
+      node.elements[i] = getIdentifierName(element, bindings);
+    }
```
loganfsmyth: Looks like this also needs to handle the `AssignmentPattern` case.

src/workers/parser/tests/mapExpression.spec.js
P0lip: Turned out to be caused most likely by `mapOriginalExpression`. Will file a bug report once entirely certain.

src/workers/parser/mapBindings.js
P0lip: I think it's handled now. Correct me if I'm wrong.

jasonLaster:
Great job @P0lip!

PR #7264, opened 2018-11-12
Simplify the getBreakpointsForSource selector function

kenjiO:
A simplification of getBreakpointsForSource based on [this comment](https://github.com/devtools-html/debugger.html/pull/7226#issuecomment-437555166).  I made it so it no longer re-creates new breakpoint objects and also used getBreakpointsList instead of getBreakpointsMap so it can just use Array.filter to get the results.

### DIFF
--- a/src/reducers/breakpoints.js
+++ b/src/reducers/breakpoints.js
@@ -296,21 +296,13 @@ export function getBreakpointsForSource(
   }
 
   const isGeneratedSource = isGeneratedId(sourceId);
-  const breakpoints = getBreakpointsMap(state);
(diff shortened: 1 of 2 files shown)


jasonLaster:
okay, i dug a bit into disable others and i don't think you broke it :)
I think i broke it awhile back and we never tested it...

This should fix it
https://gist.github.com/4f26899b282ebc2e735aebc060d41525

we should add a mochitest

jasonLaster:
https://github.com/devtools-html/debugger.html/issues/7268

PR #7265, opened 2018-11-13
Prevent removing column breakpoint from from removing line's breakpoint when there are multiple breakpoints on the same column

darkwing:
This PR prevents the line breakpoint marker from being removed when a column breakpoint is removed when there are multiple breakpoints on a line.

### DIFF
--- a/src/components/Editor/Breakpoints.js
+++ b/src/components/Editor/Breakpoints.js
@@ -5,6 +5,7 @@
 // @flow
 import { connect } from "react-redux";
 import React, { Component } from "react";
+import { uniq, isEqual } from "lodash";
 
 import Breakpoint from "./Breakpoint";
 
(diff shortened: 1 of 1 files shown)


jasonLaster:
So, the premise of this PR confuses me...

if the user adds 2 column bps for a line visibleBreakpoints should return both breakpoints. 
- At that point we should try to show both markers in the gutter (weird)

if the user removes one column bp for a line visibleBreakpoints should return the other breakpoint
- At that point we should try to show the remaining marker in the gutter (why doesn't that work?)

darkwing:
The idea of this PR is that removing a single column breakpoint should not remove the arrow over the line number, so the logic is to get all breakpoint line numbers and do nothing if the old line number listing matches the new line number listing (i.e. there's still a breakpoint on that line, so don't remove the arrow).

I changed `componentShouldUpdate`, not component did update, if that means anything

jasonLaster:
why should it matter if we render? Is this a perf improvement? my fear is that sometimes these optimizations cause us to not update when we should sometimes...

darkwing:
It's a bug fix, in theory; if you add a breakpoint to a line, then add two column breakpoints, then deactivate one column breakpoint, the gutter marker for the breakpoint is removed, which we definitely do not want, because there's an active breakpoint there.

darkwing:
Closing as @jasonLaster and I discussed a different approach.

PR #7266, opened 2018-11-13
WIP:  Only show column breakpoints in current viewport

audreyfang:
### Summary of Changes

* @darkwing I've been experimenting with this by forcing an artificial state change onscroll. This only appears to work on some scrolls. I think this is due to asynchronous scrolling: https://developer.mozilla.org/en-US/docs/Mozilla/Performance/Scroll-linked_effects. I'm looking for some help in either figuring out what's going wrong or a way around asynchronous scrolling if that is the problem.

### DIFF
--- a/src/components/Editor/CallSites.js
+++ b/src/components/Editor/CallSites.js
@@ -47,15 +47,23 @@ class CallSites extends Component {
   componentDidMount() {
     const { editor } = this.props;
     const codeMirrorWrapper = editor.codeMirror.getWrapperElement();
+    const scroller = editor.codeMirror.getScrollerElement();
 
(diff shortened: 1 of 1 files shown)


jasonLaster:
@audreyfang i think this version will call codemirror for each check. Can we refactor the code so that we calculate the viewport coord positions first. e.g. `{lineStart: 1, lineEnd: 10, columnStart: 1, columnEnd: 100}` or `{start: {line: 1, column:1}, end: {line:10, column: 100}}`

darkwing:
Closing in favor of :  https://github.com/devtools-html/debugger.html/pull/7311

PR #7269, opened 2018-11-13
Use spread to get array instead of slice.call

darkwing:
Just trying to get some momentu m


jasonLaster:
Yep, it would be nice we had an up to date Flow type and double check the behavior because i doubt this functionality is tested.

loganfsmyth:
Oh I didn't mean to derail this, it seems safe to me. The majority of array-like things are also iterable, so this should be pretty much identical.

PR #7271, opened 2018-11-14
[Breakpoints] fix breakpoint context menu logic

### LINKED ISSUE #7268: [breakpoints] disable others no longer works
STR:

1. add a bunch of breakpoints
2. in the breakpoint panel right click a breakpoint and select disable others

This fixes it
https://gist.github.com/4f26899b282ebc2e735aebc060d41525

But, we should also add a jest or mochitest test

kenjiO:
Fixes #7268 

### Summary of Changes

* update BreakpointContextMenu.js logic that finds other breakpoints.  A formatted breakpoint object was being compared to a breakpoint object using object equality which was always returning that they were not the same breakpoint.  I changed it to compare using breakpoint id.
* added the breakpoint id property to FormattedBreakpoint objects

### DIFF
--- a/src/test/mochitest/browser_dbg-breakpoints-actions.js
+++ b/src/test/mochitest/browser_dbg-breakpoints-actions.js
@@ -1,9 +1,8 @@
 /* Any copyright is dedicated to the Public Domain.
  * http://creativecommons.org/publicdomain/zero/1.0/ */
 
-async function removeBreakpoint(dbg) {
-  rightClickElement(dbg, "breakpointItem", 3)
-  selectMenuItem(dbg, 1);
+function openFirstBreakpointContextMenu(dbg){
+  rightClickElement(dbg, "breakpointItem", 3);
 }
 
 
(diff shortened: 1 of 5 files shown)


src/test/mochitest/browser_dbg-breakpoints-actions.js
```diff
@@ -14,8 +13,58 @@ add_task(async function() {
...
+    dbg.selectors.getBreakpointsList(state).every(bp => bp.disabled)
+  );
+  ok("all breakpoints are disabled")
+
+  openFirstBreakpointContextMenu(dbg)
+  // select "Enable Others"
+  selectMenuItem(dbg, 3);
+  await waitForState(dbg, state =>
+    dbg.selectors.getBreakpointsList(state)
+      .every(bp => (bp.location.line === 1) === bp.disabled)
```
jasonLaster: okay, this is the craziest thing I've seen in awhile, but it's a test so lets go crazy

jasonLaster:
This is beautiful. I love that you opened two PRs by the way, that really helps keep things moving

PR #7276, opened 2018-11-15
Bug 1505701 - Propagate all errors from promiseMiddleware.

arai-a:
Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1505701

### Summary of Changes

Changed `promiseMiddleware` to propagate error which is thrown from `executeSoon`
this can happen if the window that executes `setTimeout` (`executeSoon`'s impl) is closed
before the timeout happens.

The file that needs fix for https://bugzilla.mozilla.org/show_bug.cgi?id=1505701 is only  src/actions/utils/middleware/promise.js , but I've applied the same fix to packages/devtools-reps/src/launchpad/utils/redux/middleware/promise.js given it looks like another copy of the same file which has the same issue.

### Test Plan

This has no testcase, given the case this patch handles is already inside existing testcase in m-c:
    * devtools/client/canvasdebugger/test/browser_canvas-frontend-call-stack-01.js
    * devtools/client/inspector/markup/test/browser_markup_links_06.js

The detail of the issue is described in https://bugzilla.mozilla.org/show_bug.cgi?id=1505701
and they hit error if I fix https://bugzilla.mozilla.org/show_bug.cgi?id=1498775

### DIFF
--- a/packages/devtools-reps/src/launchpad/utils/redux/middleware/promise.js
+++ b/packages/devtools-reps/src/launchpad/utils/redux/middleware/promise.js
@@ -2,7 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
 
-const { defer, executeSoon, filterByKey } = require("../../utils");
+const { executeSoon, filterByKey } = require("../../utils");
 
 const PROMISE = (exports.PROMISE = "@@dispatch/promise");
 let seqIdVal = 1;
@@ -29,32 +29,28 @@ function promiseMiddleware({ dispatch, getState }) {
     dispatch(Object.assign({}, action, { status: "start" }));
 
     // Return the promise so action creators can still compose if they want to.
-    const deferred = defer();
-    promiseInst.then(
-      value => {
-        executeSoon(() => {
+    return Promise.resolve(promiseInst)
+      .finally(() => new Promise(resolve => executeSoon(resolve)))
+      .then(
+        value => {
           dispatch(
             Object.assign({}, action, {
               status: "done",
               value: value
             })
           );
-          deferred.resolve(value);
-        });
-      },
-      error => {
-        executeSoon(() => {
+          return value;
+        },
+        error => {
           dispatch(
             Object.assign({}, action, {
               status: "error",
               error: error.message || error
             })
           );
-          deferred.reject(error);
-        });
-      }
-    );
-    return deferred.promise;
+          return Promise.reject(error);
+        }
+      );
   };
 }
 
--- a/src/actions/utils/middleware/promise.js
+++ b/src/actions/utils/middleware/promise.js
@@ -82,26 +82,22 @@ function promiseMiddleware({
 
     // Return the promise so action creators can still compose if they
     // want to.
-    return new Promise((resolve, reject) => {
-      promiseInst.then(
+    return Promise.resolve(promiseInst)
+      .finally(() => new Promise(resolve => executeSoon(resolve)))
+      .then(
         value => {
-          executeSoon(() => {
-            dispatch({ ...action, status: "done", value: value });
-            resolve(value);
-          });
+          dispatch({ ...action, status: "done", value: value });
+          return value;
         },
         error => {
-          executeSoon(() => {
-            dispatch({
-              ...action,
-              status: "error",
-              error: error.message || error
-            });
-            reject(error);
+          dispatch({
+            ...action,
+            status: "error",
+            error: error.message || error
           });
+          return Promise.reject(error);
         }
       );
-    });
   };
 }
 


jasonLaster:
It looks like `finally` is not in node yet. We might need to add it via this polyfill https://www.npmjs.com/package/promise.prototype.finally here 
https://github.com/devtools-html/debugger.html/blob/master/src/test/tests-setup.js#L1

loganfsmyth:
@arai-a If you change https://github.com/devtools-html/debugger.html/blob/55f610ae00e4a5e74108e6e90a377796b977aaee/.travis.yml#L6 to `10` I think the tests will pass.

arai-a:
thanks for handling incompatibility :)
now looks like tests are passed.

loganfsmyth:
@arai-a Unfortunately I wasn't able to include this in our most recent release from Github to MC because we were seeing some MC test failures on `dt8` with flakiness in `dt1` and `dt7`: https://treeherder.mozilla.org/#/jobs?repo=try&revision=9288630ff3f60a8d30b1531eec05249435873ca9&selectedJob=212363871

Would you be able to look into that?

arai-a:
looks like it's pre-existing unhandled rejection (or newly added, given it didn't fail on last try)
I'll take a look.

arai-a:
here's the trace

promiseMiddleware/</</</<@resource://devtools/client/debugger/new/src/actions/utils/middleware/promise.js:46:79
```
function promiseMiddleware({
  dispatch,
  getState
}) {
  return next => action => {
...
    return Promise.resolve(promiseInst).finally(() => new Promise(resolve => (0, _DevToolsUtils.executeSoon)(resolve))).then(value => {
------------------------------------------------------------------------------^
```

promiseMiddleware/</</<@resource://devtools/client/debugger/new/src/actions/utils/middleware/promise.js:46:55
```
function promiseMiddleware({
  dispatch,
  getState
}) {
  return next => action => {
...
    return Promise.resolve(promiseInst).finally(() => new Promise(resolve => (0, _DevToolsUtils.executeSoon)(resolve))).then(value => {
------------------------------------------------------^
```

Async*promiseMiddleware/</<@resource://devtools/client/debugger/new/src/actions/utils/middleware/promise.js:46:41
```
function promiseMiddleware({
  dispatch,
  getState
}) {
  return next => action => {
...
    return Promise.resolve(promiseInst).finally(() => new Promise(resolve => (0, _DevToolsUtils.executeSoon)(resolve))).then(value => {
----------------------------------------^
```

thunk/</</<@resource://devtools/client/debugger/new/src/actions/utils/middleware/thunk.js:21:100
```
function thunk(makeArgs) {
    const args = { dispatch, getState };

    return next => action => {
      return typeof action === "function" ? action(makeArgs ? makeArgs(args, getState()) : args) : next(action);
---------------------------------------------------------------------------------------------------^
```

dispatch@resource://devtools/client/shared/vendor/redux.js:755:18
```
function applyMiddleware() {
...
  return function (createStore) {
    return function (reducer, preloadedState, enhancer) {
...
      var middlewareAPI = {
        getState: store.getState,
        dispatch: function dispatch(action) {
          return _dispatch(action);
-----------------^
```

enableBreakpoint/<@resource://devtools/client/debugger/new/src/actions/breakpoints/addBreakpoint.js:120:12
```
function enableBreakpoint(location) {
  return async ({ dispatch, getState, client, sourceMaps }) => {
...
    return dispatch({
-----------^
```

async*thunk/</</<@resource://devtools/client/debugger/new/src/actions/utils/middleware/thunk.js:21:45
```
function thunk(makeArgs) {
    const args = { dispatch, getState };

    return next => action => {
      return typeof action === "function" ? action(makeArgs ? makeArgs(args, getState()) : args) : next(action);
--------------------------------------------^
```

dispatch@resource://devtools/client/shared/vendor/redux.js:755:18
```
function applyMiddleware() {
...
  return function (createStore) {
    return function (reducer, preloadedState, enhancer) {
...
      var middlewareAPI = {
        getState: store.getState,
        dispatch: function dispatch(action) {
          return _dispatch(action);
-----------------^
```

toggleBreakpoints/</promises<@resource://devtools/client/debugger/new/src/actions/breakpoints/index.js:165:130
```
function toggleBreakpoints(shouldDisableBreakpoints, breakpoints) {
  return async ({ dispatch }) => {
    const promises = breakpoints.map(breakpoint => shouldDisableBreakpoints ? dispatch(disableBreakpoint(breakpoint.location)) : dispatch((0, _addBreakpoint.enableBreakpoint)(breakpoint.location)));
---------------------------------------------------------------------------------------------------------------------------------^

    await Promise.all(promises);
  };
```

toggleBreakpoints/<@resource://devtools/client/debugger/new/src/actions/breakpoints/index.js:165:22
```
function toggleBreakpoints(shouldDisableBreakpoints, breakpoints) {
  return async ({ dispatch }) => {
    const promises = breakpoints.map(breakpoint => shouldDisableBreakpoints ? dispatch(disableBreakpoint(breakpoint.location)) : dispatch((0, _addBreakpoint.enableBreakpoint)(breakpoint.location)));
---------------------^

    await Promise.all(promises);
  };
```

async*thunk/</</<@resource://devtools/client/debugger/new/src/actions/utils/middleware/thunk.js:21:45
```
function thunk(makeArgs) {
    const args = { dispatch, getState };

    return next => action => {
      return typeof action === "function" ? action(makeArgs ? makeArgs(args, getState()) : args) : next(action);
--------------------------------------------^
```

bindActionCreator/<@resource://devtools/client/shared/vendor/redux.js:644:12
```
function bindActionCreator(actionCreator, dispatch) {
  return function () {
    return dispatch(actionCreator.apply(undefined, arguments));
-----------^
```

click@resource://devtools/client/debugger/new/src/components/SecondaryPanes/Breakpoints/BreakpointsContextMenu.js:105:18
```
  const enableOthersItem = {
    id: "node-menu-enable-others",
    label: enableOthersLabel,
    accesskey: enableOthersKey,
    disabled: false,
    click: () => toggleBreakpoints(false, otherDisabledBreakpoints)
-----------------^
```

Menu.prototype._createMenuItems/</<@resource://devtools/client/debugger/new/dist/vendors.js:1930:50
```
      menuitem.addEventListener("command", () => item.click());
-------------------------------------------------^
```

EventListener.handleEvent*Menu.prototype._createMenuItems/<@resource://devtools/client/debugger/new/dist/vendors.js:1930:7
```
      menuitem.addEventListener("command", () => item.click());
------^
```

This time the unhandled rejection is caught by test harness.
(the rejection of `toggleBreakpoints` is surely not handled)

the actual issue is the interaction between the middleware and the consumers,
that is, whether that the consumer expect the failure from middleware's implementation.

in this case the issue is that the tab gets closed before the setTimeout (which depends on the tab's lifetime) used in the middleware finishes.
With this PR's fix, the middleware itself expects that setTimeout can fail,
but if the consumer doesn't expect it and just report it as an error,
such tests (which closes the tab before setTimeout) will still fail.

arai-a:
The solution is either:
  - Change the testcase to wait until setTimeout and all remaining handlers are finished before closing tab (given the testsuite runs very quickly compared to human, it might be reasonable to not expect such error on outside of automation)
  - Handle the rejection, maybe just explicitly ignore it

jasonLaster:
> Change the testcase to wait until setTimeout and all remaining handlers are finished before closing tab

This seems pretty good to me.

jasonLaster:
I wonder if Redux's middleware api would let us observes the async dispatchers (thunks) and trigger an event when all of the thunks are done?

```diff
diff --git a/src/actions/utils/middleware/thunk.js b/src/actions/utils/middleware/thunk.js
index 6e5e57bdc..f214fb921 100644
--- a/src/actions/utils/middleware/thunk.js
+++ b/src/actions/utils/middleware/thunk.js
@@ -8,6 +8,21 @@
 
 import type { ThunkArgs, ActionType } from "../../types";
 
+const thunks = new Set();
+
+function checkThunkQueue(thunk) {
+  thunks.delete(thunk);
+  if (thunks.size == 0) {
+    // this might be too hacky
+    document.dispatchEvent(new Event("thunks:cleared"));
+  }
+}
+
+function newThunk(thunk) {
+  thunks.add(thunk);
+  thunk.then(checkThunkQueue);
+}
+
 /**
  * A middleware that allows thunks (functions) to be dispatched. If
  * it's a thunk, it is called with an argument that contains
@@ -20,9 +35,13 @@ export function thunk(makeArgs: any) {
     const args = { dispatch, getState };
 
     return (next: Function) => (action: ActionType) => {
-      return typeof action === "function"
-        ? action(makeArgs ? makeArgs(args, getState()) : args)
-        : next(action);
+      if (action === "function") {
+        const thunk = action(makeArgs ? makeArgs(args, getState()) : args);
+        newThunk(thunk);
+        return thunk;
+      }
+
+      return next(action);
     };
   };
}
```

arai-a:
waiting for all thunk doesn't work if there's any thunk that is not related to the test and if it doesn't resolved,
and looks like that's the case here (there's 1 remaining thunk in the set after the toggleBreakpoint's handling)

I'll check what it is, and also try other solution as well, maybe tonight

arai-a:
if I ignore the failure from executeSoon and continue to fulfillment case,
it fails with the following error:

```
  FAIL A promise chain failed to handle a rejection: No such actor for ID: server1.conn1.child1/source24 - stack: (No stack available.)
Rejection date: Thu Nov 22 2018 12:29:37 GMT+0900 (Japan Standard Time) - false == true - JS frame :: resource://testing-common/PromiseTestUtils.jsm :: assertNoUncaughtRejections :: line 257
Stack trace:
resource://testing-common/PromiseTestUtils.jsm:assertNoUncaughtRejections:257
chrome://mochikit/content/browser-test.js:nextTest:736
chrome://mochikit/content/browser-test.js:testScope/test_finish/<:1395
chrome://mochikit/content/browser-test.js:run:1332
```

the remaining part also depends on the lifetime of the tab and the debugging session.
so I think waiting for completion is the simplest fix.

arai-a:
added "TOGGLE_BREAKPOINTS" dispatch, and it runs successfully locally.
try is running:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=d61f144a2358d8bc90e7b042bcdb2ef9c98908e3

then, where is devtools/client/debugger/new/test/mochitest/browser_dbg-breakpoints-actions.js stored on this repo?
I cannot locate the same file.
(there's src/test/mochitest/browser_dbg-breakpoints-actions.js but the content seems to be different)

loganfsmyth:
Alright, I've looked this over because I was surprised it wasn't _already_ working and have tracked down the issue.

These tests wait for the `disabled` flag on the breakpoints to be toggled with https://github.com/devtools-html/debugger.html/blob/ec5b4e2565c9df3d69e6918eb212c15441f9d7a6/src/test/mochitest/browser_dbg-breakpoints-actions.js#L58

The problem is that we optimistically set `disabled: false` in https://github.com/devtools-html/debugger.html/blob/master/src/actions/breakpoints/addBreakpoint.js#L121 without actually waiting for the action's promise to finish dispatching in https://github.com/devtools-html/debugger.html/blob/master/src/actions/breakpoints/addBreakpoint.js#L152. This means that the test will advance without waiting for the actions to finish dispatching, which means we can potentially dispatch actions after the test has finished evaluating and the tab has been closed.

It looks like breakpoints technically have a `loading` flag, but it doesn't appear to ever be set to `true` anywhere that I can see so I'm not sure if it is what we'd want.

loganfsmyth:
Alright, we landed #7350 so this should be fixed in one of the next releases.

PR #7280, opened 2018-11-15
Prevent duplicated generated locations from displaying in call sites

### LINKED ISSUE #7248: Avoid rendering column breakpoints that would not map properly to the generated source
One issue we want to address with column breakpoints is hiding them in cases where they won't render in a meaningful way. This happens currently in sourcemapped content that only has line-granularity mappings. This also arises for pretty-printed files, though we will ideally fix that eventually.

The core of the issue is that if you look at an original file, it may have many useful column locations, but if the source mapping is not fine-grained enough, you can easily take a line/column location on the original file and simply end up at the start of the line in the generated file, essentially making the column breakpoint useless.

We want to be able to detect this failure mode and skip rendering column breakpoints for a given line if it is detected to have poor mappings. One way to do this would be to inspect the map itself to get a sense of its structure, but this can end up being relatively complex.

What is probably easier would be to take advantage of the high likelyhood that the bad mappings mean that multiple column offsets will all map to the same column in the generated code. We could then consider

```
if (callSites.length > 1 && !hasConflictingGeneratedPosition(callSites)){
  // only render the callsites if this passes
}
```
where `hasConflictingGeneratedPosition` would take the original line/column of each callsite and get its generated location. If the mappings are fine-grained, we'd expect each callsite to have a unique generated line/column, but if they are line-level mappings, it is very likely that multiple callsites would share the same generated line/column offset, meaning that we'd likely want to skip rendering the callsites entirely.

darkwing:
Fixes #7248

This is far enough along that I need eyes on it @loganfsmyth @jasonLaster ; ignore the flow errors.

I used @jasonLaster's `hasConflictingGeneratedPosition` suggestion but I'd like more scrutiny on that.

### DIFF
--- a/src/actions/utils/symbols.js
+++ b/src/actions/utils/symbols.js
@@ -16,11 +16,12 @@ function updateSymbolLocation(
   source: Source,
   sourceMaps: any
 ) {
-  site.location.start.sourceId = source.id;
   return sourceMaps
-    .getGeneratedLocation(site.location.start, source)
+    .getGeneratedLocation(
+      { ...site.location.start, sourceId: source.id },
+      source
+    )
     .then(loc => {
-      // site.generatedLocation = { line: loc.line, column: loc.column };
       return {
         ...site,
         generatedLocation: { line: loc.line, column: loc.column }
@@ -44,5 +45,6 @@ export async function updateSymbolLocations(
   );
 
   const newSymbols = { ...symbols, callExpressions: mappedCallExpressions };
+
   return Promise.resolve(newSymbols);
 }
--- a/src/actions/utils/symbols.js
+++ b/src/actions/utils/symbols.js
@@ -43,9 +43,6 @@ export async function updateSymbolLocations(
     )
   );
 
-  symbols.callExpressions = mappedCallExpressions;
-
-  console.log(mappedCallExpressions);
-
-  Promise.resolve(symbols);
+  const newSymbols = { ...symbols, callExpressions: mappedCallExpressions };
+  return Promise.resolve(newSymbols);
 }
--- a/src/actions/utils/symbols.js
+++ b/src/actions/utils/symbols.js
@@ -1,0 +1,51 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
+
+// @flow
+
+import {
+  type SymbolDeclaration,
+  type SymbolDeclarations
+} from "../../workers/parser";
+
+import type { Source } from "../../types";
+
+function updateSymbolLocation(
+  site: SymbolDeclaration,
+  source: Source,
+  sourceMaps: any
+) {
+  site.location.start.sourceId = source.id;
+  return sourceMaps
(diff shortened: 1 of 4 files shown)


src/actions/utils/symbols.js
```diff
@@ -0,0 +1,51 @@
...
+export async function updateSymbolLocations(
+  symbols: SymbolDeclarations,
+  source: Source,
+  sourceMaps: any
+): Promise<SymbolDeclarations> {
+  if (!symbols || !symbols.callExpressions) {
+    return Promise.resolve(symbols);
+  }
+
+  const mappedCallExpressions = await Promise.all(
```
jasonLaster: this could probably be

```js
return await Promise.all(
...
) 
```

src/actions/utils/symbols.js
```diff
@@ -0,0 +1,51 @@
...
+    symbols.callExpressions.map(site =>
+      updateSymbolLocation(site, source, sourceMaps)
+    )
+  );
+
+  symbols.callExpressions = mappedCallExpressions;
```
jasonLaster: we want to avoid mutating

src/actions/utils/symbols.js
```diff
@@ -0,0 +1,51 @@
...
+
+  console.log(mappedCallExpressions);
+
+  Promise.resolve(symbols);
```
jasonLaster: this no longer needs to be there, but if it did we should return it right?

src/actions/utils/symbols.js
```diff
@@ -0,0 +1,50 @@
...
+
+  const mappedCallExpressions = await Promise.all(
+    symbols.callExpressions.map(site =>
+      updateSymbolLocation(site, source, sourceMaps)
+    )
+  );
+
+  const newSymbols = { ...symbols, callExpressions: mappedCallExpressions };
+
+  return Promise.resolve(newSymbols);
```
jasonLaster: why not just?

```js
return Promise.all(symbols.callExpressions.map(site =>
    updateSymbolLocation(site, source, sourceMaps)
  )
);
```

jasonLaster: Oh, i see yeah... this makes sense :)

darkwing:
WINNING!

yesssss

jasonLaster:
Yes! One small thing, that i likely wasn't clear about earlier. We only what to show column breakpoints IF there are 2 or more column breakpoints.

jasonLaster:
Do we have a mochitest to ensure that a pretty printed file does not have column breakpoints?

src/components/Editor/CallSites.js
```diff
@@ -123,21 +123,40 @@ class CallSites extends Component {
...
+    // Find all callsites that have a breakpoint set on their start line
+    const callSitesForBreakpointLines = callSites
+      .filter(({ location }) => breakpointLines.has(location.start.line))
+      // Filter by only lines with multiple column breakpoints
+      .filter(({ location }, index, arr) => {
+        return (
+          arr.filter(site => site.location.start.line === location.start.line)
+            .length > 1
+        );
+      });
```
darkwing: This block needs to be rewritten.  We've continued to add conditions so a fresh look will do us good.  I think we should index the sites by line, then eliminate those with only one column bp.

PR #7285, opened 2018-11-16
Add a title to Group component description element.

### LINKED ISSUE #7209: In the Call stack panel, Groups should have a title
In https://github.com/devtools-html/debugger.html/blob/6daa22e0253ab6e30f8d250784e59ae6cbf4327e/src/components/SecondaryPanes/Frames/Group.js#L128-L133 

we should have a `title` attribute which would contain the following: `Show all ${frame.library} frames` / `Collapse ${frame.library} frames`.

In order to do that, 2 new strings should be added in `debugger.properties`, (e.g. `callStack.group.expandTooltip` and `callStack.group.collapseTooltip`): 

```
callStack.group.expandTooltip=Show all %S frames
```

Then, in Group.js, `L10N.getFormatStr("callStack.group.expandTooltip", frame.library)` can be used to replace the `%S` with the library name.

nchevobbe:
Fixes #7209.

This adds a title to the group component
that indicates it can be expanded/collapsed.
Two location entries are added to handle both
collapsed and expanded state.

### DIFF
--- a/assets/panel/debugger.properties
+++ b/assets/panel/debugger.properties
@@ -416,6 +416,20 @@ callStack.collapse=Collapse rows
 # message to show more of the frames.
 callStack.expand=Expand rows
 
+# LOCALIZATION NOTE (callStack.group.expandTooltip): The text that will appear
+# when hovering a collapsed Group of frames in the callStack panel. `frames` is
+# always plural since a group can only exist if it contain more that 1 frame.
+# %S is replaced by the name of the library of the frames in the group.
+# example: `Show React frames`.
+callStack.group.expandTooltip=Show %S frames
+
+# LOCALIZATION NOTE (callStack.group.collapseTooltip): The text that will appear
+# when hovering an expanded Group of frames in the callStack panel. `frames` is
+# always plural since a group can only exist if it contain more that 1 frame.
+# %S is replaced by the name of the library of the frames in the group.
+# example: `Collapse React frames`.
+callStack.group.collapseTooltip=Collapse %S frames
+
 # LOCALIZATION NOTE (components.header): Header for the
 # Framework Components pane in the right sidebar.
 components.header=Components
--- a/src/components/SecondaryPanes/Frames/Group.js
+++ b/src/components/SecondaryPanes/Frames/Group.js
@@ -4,6 +4,7 @@
 
 // @flow
 import React, { Component } from "react";
+import PropTypes from "prop-types";
 import classNames from "classnames";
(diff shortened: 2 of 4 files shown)


assets/panel/debugger.properties
```diff
@@ -416,6 +416,18 @@ callStack.collapse=Collapse rows
 # message to show more of the frames.
 callStack.expand=Expand rows
 
+# LOCALIZATION NOTE (callStack.group.expandTooltip): The text that will appear
+# when hovering a collapsed Group of frames in the callStack panel. `frames` is
+# always plural since a group can only exist if it contain more that 1 frame.
+# %S is replaced by the name of the library of the frames in the group.
+callStack.group.expandTooltip=Show %S frames
```
flodolo: Unfortunately "always plural" (i.e. more than 1) doesn't really matter, we need a proper plural form here.
https://developer.mozilla.org/en-US/docs/Mozilla/Localization/Localization_and_Plurals

There should be already [usages in the code](https://github.com/devtools-html/debugger.html/blob/4bb14abb94fbb4361710600ac5e1c9c2da75c3f6/assets/panel/debugger.properties#L758-L765).

assets/panel/debugger.properties
```diff
@@ -416,6 +416,18 @@ callStack.collapse=Collapse rows
...
+
+# LOCALIZATION NOTE (callStack.group.collapseTooltip): The text that will appear
+# when hovering an expanded Group of frames in the callStack panel. `frames` is
+# always plural since a group can only exist if it contain more that 1 frame.
+# %S is replaced by the name of the library of the frames in the group.
+callStack.group.collapseTooltip=Collapse %S frames
```
flodolo: Same here, this needs an actual plural

assets/panel/debugger.properties
nchevobbe: I'm not sure what should be done here. What I meant by "always plural" is that it will always be: `Show React frames` (for example).
So there's no number I can pass to it to decide between singular and plural. Should I still be using a proper PluralForm if I don't have any numbers?

flodolo: Sorry, I completely missed that was the result before seeing the example, and was thrown off by the "always plural". You're correct: without a number, you can't have a correct plural form, so this should be interpreted as a generic "frames".

In the end, this is correct. I would only add the the "Show React frames" as an example in the comment for clarity.

nchevobbe: will do, thanks :)

PR #7286, opened 2018-11-16
Add context check in getLibraryFromUrl.

nchevobbe:
Some file used by libraries can't get really be identified
as part of the framework since their name can be quite
common (like zone.js).
In those case, it would be nice to check the whole callstack
to find if there's at least another frame matching a library.
(for zone.js, we should identify as Angular only if there's
another frame identified as an Angular frame).

---

See [Bug 1507327 🐛](https://bugzilla.mozilla.org/show_bug.cgi?id=1507327)

### DIFF
--- a/src/utils/pause/frames/getLibraryFromUrl.js
+++ b/src/utils/pause/frames/getLibraryFromUrl.js
@@ -78,7 +78,8 @@ const libraryMap = [
   },
   {
     label: "Angular",
-    pattern: /angular/i
+    pattern: /angular/i,
+    contextPattern: /(zone\.js)/
   },
   {
     label: "Redux",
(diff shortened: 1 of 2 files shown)


nchevobbe:
Not a hundred percent sure of this, so feel free to push back if you feel like it's adding unwanted complexity

src/utils/pause/frames/getLibraryFromUrl.js
```diff
@@ -106,14 +107,34 @@ const libraryMap = [
...
-    return null;
+
+  // Let's first check if the frame match a defined pattern.
+  let match = libraryMap.find(o => frameUrl.match(o.pattern));
+  if (match) {
+    return match.label;
+  }
+
+  // If it does not, it might still be one of the case where the file is used
+  // by a library but the name has not enough specify. In such case, we want to
```
jasonLaster: > specificity

Issue #6140, opened 2018-04-26
Add a way to evaluate getter properties


nchevobbe:
_From @nchevobbe on July 19, 2017 11:15_

When https://github.com/devtools-html/devtools-core/issues/397 , we'll be able to show getters and setters in Reps/ObjectInspector.
It would be nice if we could provide a way for the user to quickly evaluate a property directly in-place in the ObjectInspector.

We could show a button, with a title warning the user of potential side-effects. This button would be shown only in ObjectInspector nodes (i.e. not for the roots).

Here's a rough first approach of what it could look like : 

<img width="573" alt="screen shot 2017-07-19 at 11 25 37" src="https://user-images.githubusercontent.com/578107/28364151-3a64fba8-6c7b-11e7-9975-717a48c7614a.png">

---

On the technical side, I am not sure how to handle this for now (that's why I don't include it in #397 since it's a nice-to-have, not a critical feature).
At the moment, we don't have any actor reference for this property. We have them for the getter and the setter though, so maybe there's a way to fire the function with that reference (really not sure about this).
We may have to pass the parent reference and the property name to the new `Accessor` rep so we could evaluate the property.

_Copied from original issue: devtools-html/devtools-core#510_

nchevobbe:
_From @Loirooriol on August 14, 2017 18:23_

What if the getter is in the prototype chain? Some class getters expect the receiver to be an instance, not the `prototype` where they are defined.

nchevobbe:
_From @davidboom95 on November 17, 2017 19:47_

Any news on this feature?

nchevobbe:
_From @wyatt8740 on February 23, 2018 2:42_

Just going to chip in and point out that Firebug can do this in older versions of Firefox and in Seamonkey.

nchevobbe:
_From @jasonLaster on February 23, 2018 3:48_

This is something that has come up in the debugger as well. I might be able to get to it in 3-6 weeks.

I'd be happy to mentor somebody who is interested in doing it sooner :)
We're all on slack [here](https://devtools-html-slack.herokuapp.com/)

nchevobbe:
_From @Loirooriol on February 23, 2018 14:56_

How does Firebug determine the receiver? Does it always use the object that owns the getter?

nchevobbe:
_From @wyatt8740 on February 23, 2018 19:6_

[...]

PR #7290, opened 2018-11-16
 Display <get> and <set> node as sibling as the property with accessors.

### LINKED ISSUE #7288: Display getters and setters in separate nodes
Since we want to have a way to evaluate getters (#6140), it brings some question about the UI, and what to do with evaluated getters (how to show getter/setter nodes, replace them if they were expanded, …)
The simplest way would be to show independent nodes for those, so given the following object: 
```js
x = {
  get y() {
    return "yyyy";
  },
  set y(a) {
    console.log("setting y")
  }
}
```

It would render the following: 
- when collapsed: `Object { y: Getter & Setter}` (same thing we currently have)
- when expanded: 
```
▼ {…}
|   y: (…)
| ▶︎ <get y()>: function
| ▶︎ <set y()>: function
```

nchevobbe:
Fixes #7288.

### Screenshots

image

---

i'm not super happy with the tests changes, but, maybe we can revamp them a bit when doing #6140 (which I will as soon as this is merged)

### DIFF
--- a/packages/devtools-reps/src/object-inspector/tests/component/__snapshots__/getter-setter.js.snap
+++ b/packages/devtools-reps/src/object-inspector/tests/component/__snapshots__/getter-setter.js.snap
@@ -2,22 +2,25 @@
 
 exports[`ObjectInspector - getters & setters renders getters and setters as expected 1`] = `
 "
-▼ x: Getter & Setter
-|  ▶︎ <get>: function x()
-|  ▶︎ <set>: function x()
+▼ root
+|    x: Getter & Setter
+|  ▶︎ <get x()>: function x()
+|  ▶︎ <set x()>: function x()
 "
 `;
 
 exports[`ObjectInspector - getters & setters renders getters as expected 1`] = `
 "
-▼ x: Getter
-|  ▶︎ <get>: function x()
+▼ root
+|    x: Getter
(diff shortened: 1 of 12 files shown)


packages/devtools-reps/src/object-inspector/tests/component/__snapshots__/getter-setter.js.snap
```diff
@@ -2,22 +2,25 @@
 
 exports[`ObjectInspector - getters & setters renders getters and setters as expected 1`] = `
 "
-▼ x: Getter & Setter
-|  ▶︎ <get>: function x()
-|  ▶︎ <set>: function x()
+▼ root
+|    x: Getter & Setter
+|  ▶︎ <get x()>: function x()
+|  ▶︎ <set x()>: function x()
```
jasonLaster: why are we `x: Getter & Setter` and the getter setter instead of just the getter and setter?

nchevobbe: because we want to be able to evaluate `x` at some point :) 
This is an intermediary step, and at some point, instead of `Getter & Setter`, we'll have a button to invoke the getter (and hide setter only properties I think)

PR #7291, opened 2018-11-16
WIP: Introduce maximum column breakpoint range

darkwing:
Fixes #<issue number>

Here's the Pull Request DocThis will prevent massive performance problems with column breakpoints.

### Summary of Changes

* change 1
* change 2

### Test Plan

Example test plan:

Feel free to improve it!

### DIFF
--- a/src/components/Editor/CallSites.js
+++ b/src/components/Editor/CallSites.js
@@ -21,6 +21,8 @@ import { isWasm } from "../../utils/wasm";
 
 import actions from "../../actions";
 
+const MAX_COLUMN_BREAKPOINT_COLUMN = 300;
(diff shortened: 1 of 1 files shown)


src/components/Editor/CallSites.js
```diff
@@ -21,6 +21,8 @@ import { isWasm } from "../../utils/wasm";
 
 import actions from "../../actions";
 
+const MAX_COLUMN_BREAKPOINT_COLUMN = 300;
```
darkwing: * This number is arbitrary, based on nothing more than a guess
* We could make this a pref so users could yolo a large number

PR #7298, opened 2018-11-17
7283/fix :formatted showing in tabs/breakpoints sidebar

### LINKED ISSUE #7283: :formatted displaying in Tabs and Breakpoints Sidebar again
Just spotted this:

prettyformatted

I think this may be a result of @WenInCode's work with querystrings.  We should also add a unit test to check for this, as it's regressed a few times now.

WenInCode:
Fixes #7283 

### Summary of Changes
* `getHasSiblingOfSameName` was returning true if there was ever more than one source returned from `getSourcesUrlsInSources`, which means that it would be true in almost any case. I've updated it to only return true if there is actually a match. I would like to write a unit test for this function, but I was having a hard time understanding how to build the initial `OuterState`
* Update `getSourceQueryString` to return the sources query string, with any `:formatted` stripped from the end. I was going to check `source.isPrettyPrinted` before replacing, but I believe that value is set using pretty much the same regex I was using.

### Test Plan
Added the following to the existing query string mochitest:
* Pretty print a source and ensure the tab has the query string still, but no `:formatted`

Hands-on QA:
1. Open https://skinny-clutch.glitch.me/
2. In the debugger, select `/script.js?s=13`
3. Click the pretty print button
4. the active (pretty print) tab should still show `script.js?s=13` but with no `:formatted`

### Screenshots
screen shot 2018-11-17 at 4 08 14 pm

### DIFF
--- a/src/reducers/sources.js
+++ b/src/reducers/sources.js
@@ -19,6 +19,7 @@ import {
 } from "../utils/source";
 import { originalToGeneratedId } from "devtools-source-map";
 import { prefs } from "../utils/prefs";
+import { parse as parseURL } from "../utils/url";
 
 import type { Source, SourceId, SourceLocation } from "../types";
 import type { PendingSelectedLocation } from "./types";
@@ -451,7 +452,9 @@ export function getHasSiblingOfSameName(state: OuterState, source: ?Source) {
     return false;
   }
 
-  return getSourcesUrlsInSources(state, source.url).length > 1;
+  const urls = getSourcesUrlsInSources(state, source.url);
+  const { origin, pathname } = parseURL(source.url);
+  return urls.filter(url => url.includes(`${origin}${pathname}`)).length > 1;
 }
 
 export function getSourceInSources(sources: SourcesMap, id: string): ?Source {
--- a/src/reducers/sources.js
+++ b/src/reducers/sources.js
@@ -19,7 +19,6 @@ import {
 } from "../utils/source";
 import { originalToGeneratedId } from "devtools-source-map";
 import { prefs } from "../utils/prefs";
-import { parse as parseURL } from "../utils/url";
 
 import type { Source, SourceId, SourceLocation } from "../types";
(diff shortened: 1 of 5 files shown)


src/reducers/sources.js
```diff
@@ -445,7 +446,9 @@ export function getHasSiblingOfSameName(state: OuterState, source: ?Source) {
     return false;
   }
 
-  return getSourcesUrlsInSources(state, source.url).length > 1;
+  const urls = getSourcesUrlsInSources(state, source.url);
+  const { origin, pathname } = parseURL(source.url);
+  return urls.filter(url => url.includes(`${origin}${pathname}`)).length > 1;
```
WenInCode: It seems like @darkwing has a different solution for this same issue. Will update mine once https://github.com/devtools-html/debugger.html/pull/7303 lands

WenInCode:
Interesting that github tells you if you force-pushed now. Anywho, this is updated now.

src/utils/source.js
```diff
@@ -462,5 +462,6 @@ export function isGenerated(source: Source) {
 }
 
 export function getSourceQueryString(source: ?Source) {
-  return source ? parseURL(source.url).search : "";
+  const query = source ? parseURL(source.url).search : "";
+  return query.replace(/:formatted$/, "");
```
darkwing: We have a function that does this replace; `getRawSourceURL`; if you update to that we can get this in.  Thank you!

src/utils/source.js
```diff
@@ -462,5 +462,6 @@ export function isGenerated(source: Source) {
 }
 
 export function getSourceQueryString(source: ?Source) {
-  return source ? parseURL(source.url).search : "";
+  const query = source ? parseURL(source.url).search : "";
```
darkwing: We could just return if there's no source.

src/utils/source.js
WenInCode: Thanks!

PR #7302, opened 2018-11-19
Fix debugger hanging with slow loading source maps

### LINKED ISSUE #7301: Debugger panel hangs for minutes if loading one or more source maps times out
Steps to reproduce:

1) Go on airbnb.com
2) Open the debugger

screen shot 2018-11-19 at 13 40 38

The problem is that loading source maps for most of Airbnb's bundles timeouts, as the files seem to be served only inside their network.

If you wait long enough for all the source maps to timeout (many minutes), eventually the debugger UI shows up and is functional.

alessandrod:
Fixes #7301 

The fix involves:
- making newSources() not wait for source maps to be loaded
- making loadSourceMaps() add each map to the state asap, instead of doing one big add once all the source maps are loaded

I added two tests that model the current bug and that fail if the two changes above are reverted.

### DIFF
--- a/src/actions/sources/newSources.js
+++ b/src/actions/sources/newSources.js
@@ -26,6 +26,7 @@ import {
 } from "../../selectors";
 
 import { prefs } from "../../utils/prefs";
+import sourceQueue from "../../utils/source-queue";
 
 import type { Source, SourceId } from "../../types";
 import type { Action, ThunkArgs } from "../types";
@@ -46,20 +47,24 @@ function createOriginalSource(
   };
 }
 
-function loadSourceMaps(sources) {
-  return async function({ dispatch, sourceMaps }: ThunkArgs) {
+function loadSourceMaps(sources: Source[]) {
+  return async function({
+    dispatch,
+    sourceMaps
+  }: ThunkArgs): Promise<Promise<Source>[]> {
     if (!prefs.clientSourceMapsEnabled) {
-      return;
+      return [];
     }
 
-    let originalSources = await Promise.all(
-      sources.map(({ id }) => dispatch(loadSourceMap(id)))
+    return flatten(
+      await Promise.all(
+        sources.map(async ({ id }) => {
+          const originalSources = await dispatch(loadSourceMap(id));
+          sourceQueue.queueSources(originalSources);
+          return originalSources;
+        })
+      )
     );
-
-    originalSources = flatten(originalSources).filter(Boolean);
-    if (originalSources.length > 0) {
-      await dispatch(newSources(originalSources));
-    }
   };
 }
 
@@ -68,11 +73,15 @@ function loadSourceMaps(sources) {
  * @static
  */
 function loadSourceMap(sourceId: SourceId) {
-  return async function({ dispatch, getState, sourceMaps }: ThunkArgs) {
+  return async function({
+    dispatch,
+    getState,
+    sourceMaps
+  }: ThunkArgs): Promise<Source[]> {
     const source = getSource(getState(), sourceId);
 
     if (!source || isOriginal(source) || !source.sourceMapURL) {
-      return;
+      return [];
     }
 
     let urls = null;
@@ -97,7 +106,7 @@ function loadSourceMap(sourceId: SourceId) {
           source: (({ ...currentSource, sourceMapURL: "" }: any): Source)
         }: Action)
       );
-      return;
+      return [];
     }
 
     return urls.map(url => createOriginalSource(url, source, sourceMaps));
@@ -197,7 +206,7 @@ export function newSources(sources: Source[]) {
 
     dispatch(({ type: "ADD_SOURCES", sources: sources }: Action));
 
-    await dispatch(loadSourceMaps(sources));
+    dispatch(loadSourceMaps(sources));
 
     for (const source of sources) {
       dispatch(checkSelectedSource(source.id));
--- a/src/utils/source-queue.js
+++ b/src/utils/source-queue.js
@@ -2,11 +2,12 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
 
+// @flow
+
 import { throttle } from "lodash";
+import type { Source } from "../types";
 
 let newSources;
-let createSource;
-let supportsWasm = false;
 let queuedSources;
 let currentWork;
 
@@ -14,24 +15,25 @@ async function dispatchNewSources() {
   const sources = queuedSources;
   queuedSources = [];
 
-  currentWork = await newSources(
-    sources.map(source => createSource(source, { supportsWasm }))
-  );
+  currentWork = await newSources(sources);
 }
 
 const queue = throttle(dispatchNewSources, 100);
 
 export default {
-  initialize: options => {
-    newSources = options.actions.newSources;
-    createSource = options.createSource;
-    supportsWasm = options.supportsWasm;
+  initialize: (actions: Object) => {
+    newSources = actions.newSources;
     queuedSources = [];
   },
-  queue: source => {
+  queue: (source: Source) => {
     queuedSources.push(source);
     queue();
   },
+  queueSources: (sources: Source[]) => {
+    queuedSources = queuedSources.concat(sources);
+    queue();
+  },
+
   flush: () => Promise.all([queue.flush(), currentWork]),
   clear: () => queue.cancel()
 };
--- a/src/actions/sources/tests/newSources.spec.js
+++ b/src/actions/sources/tests/newSources.spec.js
@@ -14,6 +14,7 @@ const {
   getSelectedSource,
   getSourceByURL
 } = selectors;
+import sourceQueue from "../../../utils/source-queue";
 
 // eslint-disable-next-line max-len
 import { sourceThreadClient as threadClient } from "../../tests/helpers/threadClient.js";
@@ -66,7 +67,7 @@ describe("sources - new sources", () => {
   });
 
   // eslint-disable-next-line
-  it("should no attempt to fetch original sources if it's missing a source map url", async () => {
+  it("should not attempt to fetch original sources if it's missing a source map url", async () => {
     const getOriginalURLs = jest.fn();
     const { dispatch } = createStore(threadClient, {}, { getOriginalURLs });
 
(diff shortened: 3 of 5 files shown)


jasonLaster:
This is a difficult issue to get right because if we update the Redux store, too often we cause performance issues for apps like (angular todos) which have one source map per file. 

https://w2x2oypmqk.codesandbox.io/

I'd love your help fine tuning this because what we have now clearly doesn't work either...

src/actions/sources/newSources.js
```diff
@@ -184,7 +185,7 @@ export function newSources(sources: Source[]) {
 
     dispatch(({ type: "ADD_SOURCES", sources: sources }: Action));
 
-    await dispatch(loadSourceMaps(sources));
```
jasonLaster: I think this is safe to do

src/actions/sources/newSources.js
```diff
@@ -50,14 +49,16 @@ function loadSourceMaps(sources) {
...
-      sources.map(({ id }) => dispatch(loadSourceMap(id)))
+    await Promise.all(
+      sources.map(({ id }) =>
+        dispatch(loadSourceMap(id)).then(originalSource => {
+          if (originalSource) {
+            dispatch(newSources(originalSource));
+          }
+          return originalSource;
+        })
+      )
```
jasonLaster: I don't think we need to return anything, so if do this it could be much simpler

```js
sources.map(async ({ id }) => {
  const originalSources = await dispatch(loadSourceMap(id));
   if (originalSources) {
      dispatch(newSources(originalSource));
    }
 })
```

My concern with this approach is that in the case of an angular app this could be 500+ redux calls to newSources..  Of course, we can add some throttling in `newSources` so at most we dispatch the new sources every 50ms.

jasonLaster: I was trying to think what a throttled addSources would look like and I think this is a reasonable approximation

```diff
diff --git a/src/actions/sources/newSources.js b/src/actions/sources/newSources.js
index 2c5b52d98..5466039e0 100644
--- a/src/actions/sources/newSources.js
+++ b/src/actions/sources/newSources.js
@@ -174,6 +174,14 @@ export function newSource(source: Source) {
   };
 }

+let throttledSources = []
+const addSources = debounce(() => {
+  return ({ dispatch, getState }: ThunkArgs) => {
+    dispatch(({ type: "ADD_SOURCES", sources: throttledSources }: Action));
+    throttledSources = [];
+  }
+}
+
 export function newSources(sources: Source[]) {
   return async ({ dispatch, getState }: ThunkArgs) => {
     sources = sources.filter(source => !getSource(getState(), source.id));
@@ -182,7 +190,8 @@ export function newSources(sources: Source[]) {
       return;
     }

-    dispatch(({ type: "ADD_SOURCES", sources: sources }: Action));
+    throttledSources = [...throttledSources, ...sources]
+    dispatch(addSources())

     await dispatch(loadSourceMaps(sources));
```

jasonLaster: Maybe an even smarter way to do this would be to re-use the same throttled source queue we use in [firefox/events.js](https://github.com/devtools-html/debugger.html/blob/master/src/client/firefox/events.js#L102-L103)

https://github.com/devtools-html/debugger.html/blob/master/src/utils/source-queue.js#L20


```diff
diff --git a/src/actions/sources/newSources.js b/src/actions/sources/newSources.js
index 2c5b52d98..3dd0d1e3f 100644
--- a/src/actions/sources/newSources.js
+++ b/src/actions/sources/newSources.js
@@ -25,6 +25,7 @@ import {
   getPendingBreakpointsForSource
 } from "../../selectors";
 
+import sourceQueue from "../../utils/source-queue";
 import type { Source, SourceId } from "../../types";
 import type { Action, ThunkArgs } from "../types";
 
@@ -50,14 +51,10 @@ function loadSourceMaps(sources) {
       return;
     }
 
-    let originalSources = await Promise.all(
-      sources.map(({ id }) => dispatch(loadSourceMap(id)))
-    );
-
-    originalSources = flatten(originalSources).filter(Boolean);
-    if (originalSources.length > 0) {
-      await dispatch(newSources(originalSources));
-    }
+    sources.map(async ({ id }) => {
+      const originalSources = await dispatch(loadSourceMap(id));
+      queueSources(originalSource);
+    });
   };
 }
 
diff --git a/src/utils/source-queue.js b/src/utils/source-queue.js
index 4c6428df8..0338edb8b 100644
--- a/src/utils/source-queue.js
+++ b/src/utils/source-queue.js
@@ -32,6 +32,11 @@ export default {
     queuedSources.push(source);
     queue();
   },
+  queueSources: sources => {
+    queuedSources = queuedSources.concat(sources);
+    queue();
+  },
+
   flush: () => Promise.all([queue.flush(), currentWork]),
   clear: () => queue.cancel()
};
```

This solution strikes me as particularly elegant as we're re-using the throttling that already exists and would be able to add new original sources pretty smoothly

alessandrod: Alright, I've made newSources use source-queue and updated the unit tests accordingly. Mochitests are failing on travis now, going to try to reproduce those failures locally now.

alessandrod: Running mochitests on my laptop takes a while so I've only managed to do 3 complete runs, the result is this:

```
Overall Summary
===============

mochitest-browser
~~~~~~~~~~~~~~~~~
Ran 2878 checks (73 tests, 2805 subtests)
Expected results: 2872
Skipped: 3 tests
Unexpected results: 2
  subtest: 2 (2 fail)

Unexpected Results
------------------
devtools/client/debugger/new/test/mochitest/browser_dbg-breakpoints-columns.js
  FAIL Test timed out -
devtools/client/debugger/new/test/mochitest/browser_dbg-tabs-keyboard.js
  FAIL Uncaught exception - at chrome://mochitests/content/browser/devtools/client/debugger/new/test/mochitest/helpers.js:605 - TypeError: findElement(...) is null; can't access its "children" property
Stack trace:
countTabs@chrome://mochitests/content/browser/devtools/client/debugger/new/test/mochitest/helpers.js:605:10
@chrome://mochitests/content/browser/devtools/client/debugger/new/test/mochitest/browser_dbg-tabs-keyboard.js:15:6
Async*Tester_execTest/<@chrome://mochikit/content/browser-test.js:1093:34
async*Tester_execTest@chrome://mochikit/content/browser-test.js:1084:16
nextTest/<@chrome://mochikit/content/browser-test.js:982:9
SimpleTest.waitForFocus/waitForFocusInner/focusedOrLoaded/<@chrome://mochikit/content/tests/SimpleTest/SimpleTest.js:803:59
```

Those two tests fail reproducibly. I'm not seeing all those time outs that seem to happen on travis.

AnshulMalik:
This is awesome. I have a patch to fix the tests :)

src/actions/sources/newSources.js
```diff
@@ -52,14 +52,12 @@ function loadSourceMaps(sources) {
...
-    let originalSources = await Promise.all(
-      sources.map(({ id }) => dispatch(loadSourceMap(id)))
-    );
-
-    originalSources = flatten(originalSources).filter(Boolean);
-    if (originalSources.length > 0) {
-      await dispatch(newSources(originalSources));
-    }
+    sources.map(async ({ id }) => {
+      const originalSources = await dispatch(loadSourceMap(id));
```
jasonLaster: can we have always return an array? and enforce it with flow?

src/utils/source-queue.js
```diff
@@ -32,6 +32,16 @@ export default {
     queuedSources.push(source);
     queue();
   },
+  queueSources: sources => {
+    queuedSources = queuedSources.concat(
+      sources.map(source => {
+        source.actor = source.id;
```
alessandrod: Thanks for fixing the tests!

Why is this required here tho? Shouldn't the callers set source.actor if it's required?

AnshulMalik: There are few possibilities where we can add actor as property, or use `id` everywhere(it's what we should do in future for CDP compatibility). I just wanted to see if this fixes things.

We can either say `source.actor || source.id` [here](https://github.com/devtools-html/debugger.html/blob/master/src/client/firefox/create.js#L48)
and fix flow, this would be what I'd do but flow is upset with me.

src/actions/sources/newSources.js
alessandrod: Right, done.

src/utils/source-queue.js
alessandrod: The type `Source` defines an [id property](https://github.com/devtools-html/debugger.html/blob/fcff67bcf43a118ba7c967a0445df9006b75176a/src/types.js#L294) while `SourcePayload` defines an [actor property](https://github.com/devtools-html/debugger.html/blob/5bc3da76eab9ac5d160c88e1dc5aaeb63b56f20a/src/client/firefox/types.js#L86). Is SourcePayload supposed to be a serialized Source? If so SourcePayload.actor should probably become SourcePayload.id?

AnshulMalik: So, SourcePayload is for what we get from the server, that is, source id as `actor` property.
But on the client side, we don't use the actor property instead we use `id`.
Ideally, we'd want the server to send us `id` and not `actor`
But for now, we'd have to make a patch somewhere.

src/actions/sources/newSources.js
```diff
@@ -46,20 +47,24 @@ function createOriginalSource(
...
-      return;
+      return [];
     }
 
-    let originalSources = await Promise.all(
-      sources.map(({ id }) => dispatch(loadSourceMap(id)))
+    return flatten(
+      sources.map(async ({ id }) => {
+        const originalSources = await dispatch(loadSourceMap(id));
+        if (originalSources !== undefined) {
```
AnshulMalik: I think now we don't need this check, since we know it'd never be `undefined`

jasonLaster: nope, no objection.

alessandrod: Alright, I've annotated `loadSourceMap` to return `Promise<Source[]>` and fixed the return points. I've updated `test-head.js` so that all tests initialize the `sourceQueue` as per my previous comment.

Then I've actually reinstated the following check tho:
```diff
+        const originalSources = await dispatch(loadSourceMap(id));
+        if (originalSources.length > 0) {
+          sourceQueue.queueSources(originalSources);
+        }
```

Otherwise whenever called - even with an empty array - `queueSources` internally will create a new array and schedule a new dispatchNewSources call, which seems not ideal. I opted for checking that the array is not empty before calling `queueSources` since it's closer to where the array is created, and it's obvious why it'd be valid for it to be empty. You could argue that `queueSources` should return early if `sources.length == 0` tho and I'm happy to do that if anyone thinks it's better.

src/utils/source-queue.js
AnshulMalik: So we call this `createSource` [function](https://github.com/devtools-html/debugger.html/blob/master/src/client/firefox/create.js#L43) when we get a new source from the server, which has the actor property.

But we also use this function to create a source from original urls(which are generated on client side) and has `id` instead of `actor`. So those original sources gets generated without an id, which causes further issues.

I have not looked into it in much detail as to find a nice solution. But I am confident that you'll find some nice answer.

alessandrod: Thanks that really helped me understand the problem!

The problem is that now the same `sourceQueue` is used to create sources coming from the firefox server (received as instances of type `SourcePayload`), and from the client (instances of type `Source`). That's problematic since the types don't have matching fields, and the client is supposed to abstract over server implementations anyway. 

One possible solution is to refactor `source-queue.js` so that it's not a module-singleton and use two separate queues, one for sources coming from the server and another for sources coming from the client. There's other simpler solutions similar to your fix, but applied to `src/client/firefox`. @jasonLaster opinions?

jasonLaster: Good question. 

I think I would change the newSource function so it calls createSource before enquing the source. 

This should be simpler too

src/actions/sources/newSources.js
```diff
@@ -46,20 +47,24 @@ function createOriginalSource(
...
+    sourceMaps
+  }: ThunkArgs): Promise<Promise<Source>[]> {
     if (!prefs.clientSourceMapsEnabled) {
-      return;
+      return [];
     }
 
-    let originalSources = await Promise.all(
-      sources.map(({ id }) => dispatch(loadSourceMap(id)))
+    return flatten(
```
jasonLaster: we don't really use the return value of this function, but it is nice to keep it as an array of original sources for debugging purposes.

jasonLaster:
just rebased, hopefully it will be green now

jasonLaster:
The tests should now be green. Lets just test one more time to make things are good

alessandrod:
This is great! Thanks so much for your help Jason and Anshul! :)

PR #7303, opened 2018-11-19
Fix getSourcesUrlsInSources so it returns only matching URLs

darkwing:
I noticed that `getSourcesUrlsInSources` is returning all sources and not a matching source, so querystrings are showing up in the tree when they don't need to.

This PR ensures that querystrings only display when they need to.

### DIFF
--- a/src/reducers/sources.js
+++ b/src/reducers/sources.js
@@ -439,7 +439,6 @@ export function getSourcesUrlsInSources(
   if (!url || !urls[url]) {
     return [];
   }
-
   const plainUrl = url.split("?")[0];
 
   return Object.keys(urls)
--- a/src/reducers/sources.js
+++ b/src/reducers/sources.js
@@ -431,13 +431,20 @@ function getSourcesByUrlInSources(
   return urls[url].map(id => sources[id]);
 }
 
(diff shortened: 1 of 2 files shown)


src/reducers/sources.js
```diff
@@ -431,13 +431,20 @@ function getSourcesByUrlInSources(
...
   if (!url || !urls[url]) {
     return [];
   }
 
-  return [...new Set(Object.keys(urls).filter(Boolean))];
+  const plainUrl = url.split("?")[0];
+
+  return Object.keys(urls)
+    .filter(Boolean)
+    .filter(sourceUrl => sourceUrl.split("?")[0] === plainUrl);
```
jasonLaster: can we add a unit test that checks this. maybe in `actions/sources/tests`?

WenInCode:
@darkwing I noticed this too in my PR, but it wasn't clear to me that it was `getSourcesUrlsInSources` problem or `getHasSiblingOfSameName` (https://github.com/devtools-html/debugger.html/pull/7298/files#diff-701eb9ae1383078fc8b17995acdd81bbR444). I'll update mine once this lands.

PR #7308, opened 2018-11-19
Fix #7307 - Prevent enter key from submitting in outline

### LINKED ISSUE #7307: Crash when hitting Enter in Outline filter input
**Steps to reproduce**
1. Go to `data:text/html,<meta charset=utf8><script>function b(){}</script>`
1. Open the debugger
1. In the sources panel, select the only source available
1. Open the Outline panel
1. In the filter input, hit <kbd>Enter</kbd>

**Expected results**

if there's only 1 item in the list, (for this example, it is the case with `b`), we should jump to the function definition.

**Actual results**

⚠️ **The debugger panel goes blank** ⚠️

---

DevEdition is not affected, which is good. We should prioritize this in order to fix it before the 65 cycle.

darkwing:
Good catch by @nchevobbe

### DIFF
--- a/src/components/PrimaryPanes/OutlineFilter.js
+++ b/src/components/PrimaryPanes/OutlineFilter.js
@@ -35,6 +35,8 @@ export default class OutlineFilter extends Component<Props, State> {
       // also bound to the ESC key
       e.preventDefault();
       this.props.updateFilter("");
+    } else if (e.key === "Enter") {
+      e.preventDefault();
     }
   };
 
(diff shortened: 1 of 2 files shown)


src/components/PrimaryPanes/OutlineFilter.js
```diff
@@ -35,6 +35,8 @@ export default class OutlineFilter extends Component<Props, State> {
       // also bound to the ESC key
       e.preventDefault();
       this.props.updateFilter("");
+    } else if (e.key === "Enter") {
+      e.preventDefault();
```
nchevobbe: could we add a comment explaining why it's needed ? With maybe a reference to the issue.

PR #7310, opened 2018-11-19
[Tests] In mochi tests, use id instead of index to select context menu items

kenjiO:
Changes the `selectMenuItem` helper function in mochi tests

- renames it to `selectContextMenuItem` to better describe which menu it is used for
- Instead of selecting a menu item by index, use a css id.  So if the menu items change the test doesn't break.  Also the id makes the test more readable compared to an arbitrary number.

### DIFF
--- a/src/test/mochitest/browser_dbg-breakpoints-cond.js
+++ b/src/test/mochitest/browser_dbg-breakpoints-cond.js
@@ -93,7 +93,7 @@ add_task(async function() {
   //right click breakpoint in breakpoints list
   rightClickElement(dbg, "breakpointItem", 3)
   // select "remove condition";
-  selectContextMenuItem(dbg, "#node-menu-remove-condition");
+  selectContextMenuItem(dbg, selectors.breakpointContextMenu.removeCondition);
   await bpCondition;
   bp = findBreakpoint(dbg, "simple2", 5);
   is(bp.condition, undefined, "breakpoint condition removed");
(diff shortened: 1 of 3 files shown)


src/test/mochitest/browser_dbg-breakpoints-cond.js
```diff
@@ -39,7 +39,10 @@ async function assertConditionalBreakpointIsFocused(dbg) {
 
 async function setConditionalBreakpoint(dbg, index, condition) {
   rightClickElement(dbg, "gutter", index);
-  selectMenuItem(dbg, 2);
+  selectContextMenuItem(
+    dbg,
+    "#node-menu-add-conditional-breakpoint,#node-menu-edit-conditional-breakpoint"
+  );
```
kenjiO: This one is kind of tricky.  When this is called from line 60 the context menu item at index 2 is **add conditional breakpoint**.  When it is called from line 66 the context menu item at index 2 is **edit conditional breakpoint**.  So I used a selector that will get either one of these possibilities.  A better alternative might be making this into separate functions for adding and editing.

darkwing:
Getting these by a name reference is a really good idea -- the index would be unreliable in the long run.  I think using an object to store these selectors would be even better:

https://github.com/devtools-html/debugger.html/blob/master/src/test/mochitest/helpers.js#L1017

It would be consistent with how we're doing things elsewhere and would be one point of reference for the context items! :)

PR #7311, opened 2018-11-19
Show only column breakpoints in viewport

darkwing:
CodeMirror has a really hard time when we inject every marker for a given line if the line is very long:

codemirrorperf

One idea was to "cut off" column breakpoints after a certain column value (https://github.com/devtools-html/debugger.html/pull/7291), but this may be a better, more flexible solution, as I attempt to accommodate only lines and characters in range of the view port.  It debounces for additional perf win.

This already feels way better than current performance, when there's massive lag when switching between tabs.

## To Do
- [x] The initial state is arbitrary values that we can get real values for
- [x] I don't yet trust my column calculation -- I need to find a more scientific way of managing that.
- [x] Figure out why sometimes column breakpoints don't display upon initial page load when breakpoints exist.

### DIFF
--- a/src/components/Editor/CallSites.js
+++ b/src/components/Editor/CallSites.js
@@ -101,9 +101,7 @@ class CallSites extends Component {
     // Update state which should trigger appropriate re-renders
     const leftColumn = Math.floor(scrollLeft > 0 ? scrollLeft / charWidth : 0);
     const rightPosition = scrollLeft + (scrollArea.clientWidth - 30);
-    const rightCharacter = Math.floor(
-      rightPosition > 0 ? rightPosition / charWidth : 0
-    );
+    const rightCharacter = Math.floor(rightPosition / charWidth);
 
     return {
       start: {
@@ -233,11 +231,20 @@ class CallSites extends Component {
       callSitesFilteredByLine
     );
 
+    console.log(
+      "callSitesInViewport: ",
+      callSitesInViewport.length,
+      JSON.stringify(this.state),
+      JSON.stringify(callSitesInViewport)
+    );
+
     let sites;
     editor.codeMirror.operation(() => {
-      const childCallSites = callSitesInViewport.map((callSite, index) => {
+      const childCallSites = callSitesInViewport.map(callSite => {
         const props = {
-          key: index,
+          key: `${callSite.location.start.line}:${
+            callSite.location.start.column
+          }`,
           callSite,
           editor,
           source: selectedSource,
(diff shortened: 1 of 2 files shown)


src/components/Editor/CallSites.js
```diff
@@ -44,20 +56,59 @@ class CallSites extends Component {
...
+      "window"
+    );
+
+    // Update state which should trigger appropriate re-renders
+    const rightCharacter = scrollLeft + (scrollArea.clientWidth - 30);
+    this.setState({
+      topVisibleLine,
+      bottomVisibleLine,
+      leftVisibleColumn: scrollLeft > 0 ? scrollLeft / charWidth : 0,
+      rightVisibleColumn: rightCharacter > 0 ? rightCharacter / charWidth : 0
```
jasonLaster: I think i would prefer  `{start: {line, column}, end: {line: column}}`

src/components/Editor/CallSites.js
```diff
@@ -44,20 +56,59 @@ class CallSites extends Component {
...
+    const scrollArea = editor.codeMirror.getScrollInfo();
+    const { scrollLeft } = editor.codeMirror.doc;
+
+    // Get scroll position
+    const rect = editor.codeMirror.getWrapperElement().getBoundingClientRect();
+    const topVisibleLine = editor.codeMirror.lineAtHeight(rect.top, "window");
+    const bottomVisibleLine = editor.codeMirror.lineAtHeight(
+      rect.bottom,
+      "window"
+    );
```
jasonLaster: can we move this into an editor helper that the current scroll logic could use too?

`utils/editor`

src/components/Editor/CallSites.js
```diff
@@ -44,18 +52,69 @@ class CallSites extends Component {
...
+    const { scrollLeft } = editor.codeMirror.doc;
+    const rect = editor.codeMirror.getWrapperElement().getBoundingClientRect();
+    const topVisibleLine = editor.codeMirror.lineAtHeight(rect.top, "window");
+    const bottomVisibleLine = editor.codeMirror.lineAtHeight(
+      rect.bottom,
+      "window"
+    );
+
+    // Update state which should trigger appropriate re-renders
+    const leftColumn = Math.floor(scrollLeft > 0 ? scrollLeft / charWidth : 0);
```
audreyfang: Why is there a > 0 check here and for rightCharacter? It seems unnecessary to me since without the check 0 is still 0 and negative values will floor as expected for a Cartesian plane.

darkwing:
With regard to the disappearing stuff, there's something weird going on but I think it's a small adjustment somewhere.  Check this out:

bp-weirdness

In the image above, you can see the last scroll position and that only one call site meets that criteria: `getElementsByName`;  however, the one remaining callsite in the frame is the `push` call.  So, from a "new logic" perspective (i.e. this PR), we're doing everything right.  The problem is how react is seeing its children, so we need to do something different to handle the individual callsites.

PR #7314, opened 2018-11-20
Make sources search work with query params

### LINKED ISSUE #7217: Can't find files with query strings
1. Load a page containing
    ```html
    <script src="test.js?foo"></script>
    <script src="test.js?bar"></script>
    ```
2. Open the debugger
3. Use Ctrl+P to find sources
4. Enter `test.js?foo`

Expected: you get a result
Actual: no result

bug1

You need to write `test.js` in order to find the results, but then you get two identical entries and you don't know which is which.

bug2

All of this used to work properly and was regressed in https://hg.mozilla.org/integration/mozilla-inbound/pushloghtml?fromchange=da7e0c710c7d7a7ff3e124db32d4eae9ffebd2a1&tochange=598305280dc37b6b0136795508e4ff7c8ab7b480

https://github.com/devtools-html/debugger.html/compare/release-81...release-82

WenInCode:
Fixes #7217 

### Summary of Changes
- It seems like the search was expecting `tabs` to be an array of strings, but it was really a `TabList`. Updated the `showTopSources` function to convert the tab array into array of source urls for comparison. Now if you have no query in your search, it will show the top sources (open tabs).
- Updated the search source subtitle to be a truncated copy of the sources `relativeUrl` + `query`.
- Updated the search source value to be the sources `relativeUrl` + `query` so that searching by query string works.

### Test Plan
- [x] Open http://skinny-clutch.glitch.me/
- [x] Open a source file
- [x] Cmd+P to open search
- [x] with the search empty, the source in the tab should show up in the search dropdown
- [x] Search for a source with a query string
- [x] It should work!

### To Do
- [ ] Add a mochitest

### Screenshots/Videos

Before:
screen shot 2018-11-20 at 9 02 05 am
searchsourcesquery-before

After:
screen shot 2018-11-20 at 9 07 12 am
searchsourcesquery-after

### DIFF
--- a/src/components/QuickOpenModal.js
+++ b/src/components/QuickOpenModal.js
@@ -37,6 +37,7 @@ import type {
 
 import type { Location, Source } from "../types";
 import type { QuickOpenType } from "../reducers/quick-open";
+import type { Tab } from "../reducers/tabs";
 
 import "./QuickOpenModal.css";
 
@@ -48,7 +49,7 @@ type Props = {
   searchType: QuickOpenType,
   symbols: FormattedSymbolDeclarations,
   symbolsLoading: boolean,
-  tabs: string[],
+  tabs: Tab[],
   shortcutsModalEnabled: boolean,
   selectSpecificLocation: Location => void,
   setQuickOpenQuery: (query: string) => void,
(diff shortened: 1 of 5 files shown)


WenInCode:
Having a _time_ trying to get this mochitest change in. My last commit is a bit of a guess at the mochitest. It seems like the firefox that mochitests runs on isn't picking up changes to the debugger code, despite me running `yarn watch`. 

I tried deleting my local repo all together, re-cloning, and running the prepare script. Once I checked out this branch I ran `yarn watch` and then `yarn mochid browser_dbg-sources-querystring`. The tests run fine, but it doesn't pick up any changes since running the prepare script (which was on master).

PR #7316, opened 2018-11-20
[testing] Wait for sources to load

jasonLaster:


loganfsmyth:
Seems like it might be good to do a pass through the tests to see if there are places where we should be passing sources and aren't, because I bet a lot of tests are calling `selectSource` without `waitForSources`. Maybe a "good first issue"?

src/test/mochitest/helpers.js
```diff
@@ -496,10 +496,12 @@ function clearDebuggerPreferences() {
  * @return {Promise} dbg
  * @static
  */
-async function initDebugger(url) {
+async function initDebugger(url, ...sources) {
   clearDebuggerPreferences();
   const toolbox = await openNewTabAndToolbox(EXAMPLE_URL + url, "jsdebugger");
-  return createDebuggerContext(toolbox);
+  const dbg = createDebuggerContext(toolbox);
+  await waitForSources(dbg, sources)
```
loganfsmyth: Looks like this needs to do `...sources`.

src/test/mochitest/browser_dbg-xhr-breakpoints.js
```diff
@@ -65,10 +64,7 @@ add_task(async function() {
 
 // Tests the "pause on any URL" checkbox works properly
 add_task(async function() {
-  const dbg = await initDebugger("doc-xhr.html");
-  await waitForSources(dbg, "fetch.js");
-
-  info("HERE 1");
```
loganfsmyth: Good catch!

PR #7317, opened 2018-11-20
[breakpoints] experiment with using pause points

jasonLaster:
### Summary of Changes

This gives us a lot more locations

screen shot 2018-11-20 at 6 22 06 pm


jasonLaster:
Thanks for the fixes @darkwing.

re: styling I think this looks better, but we should open a new issue to discuss the options. 

We have white space on the left to show where you currently are. I think we can likely improve the look and feel, but I'd like to preserve some indication of where you are paused in the line..

PR #7318, opened 2018-11-20
Sort breakpoints by line number, then column number

darkwing:
This sorts breakpoints in the right sidebar by: line number, line number length (so `143` is after `62`), and line number.

### DIFF
--- a/src/components/SecondaryPanes/Breakpoints/index.js
+++ b/src/components/SecondaryPanes/Breakpoints/index.js
@@ -7,6 +7,7 @@
 import React, { Component } from "react";
 import classnames from "classnames";
 import { connect } from "react-redux";
+import { sortBy } from "lodash";
 
 import ExceptionOption from "./ExceptionOption";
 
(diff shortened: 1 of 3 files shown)


src/components/SecondaryPanes/Breakpoints/index.js
```diff
@@ -79,14 +80,23 @@ class Breakpoints extends Component<Props> {
...
       ...breakpointSources.map(({ source, breakpoints, i }) => {
         const path = getDisplayPath(source, sources);
+        const sortedBreakpoints = sortBy(
+          breakpoints,
+          ({ selectedLocation }) => [
+            selectedLocation.line,
+            (selectedLocation.column || 0).toString().length,
+            selectedLocation.column || 0
+          ]
+        );
```
jasonLaster: lets move this sorting to the selector

src/components/SecondaryPanes/Breakpoints/index.js
```diff
@@ -79,14 +80,23 @@ class Breakpoints extends Component<Props> {
     return [
       ...breakpointSources.map(({ source, breakpoints, i }) => {
         const path = getDisplayPath(source, sources);
+        const sortedBreakpoints = sortBy(
+          breakpoints,
+          ({ selectedLocation }) => [
+            selectedLocation.line,
+            (selectedLocation.column || 0).toString().length,
+            selectedLocation.column || 0
```
jasonLaster: why is this so complicated?

is this shared with some other logic that should be in a util?

darkwing:
Yay!
bporder

PR #7320, opened 2018-11-21
Ensure that we don't overwrite a loaded source with a loading source to fix flakiness.

loganfsmyth:
In release 103 we had to exclude the source-map upgrade because of flakiness on Windows. It seems to be that we were just uncovering existing flakiness in the existing test, but it varies depending on source-map timing.

### DIFF
--- a/src/actions/sources/newSources.js
+++ b/src/actions/sources/newSources.js
@@ -81,12 +81,18 @@ function loadSourceMap(sourceId: SourceId) {
     }
 
     if (!urls) {
+      // The source might have changed while we looked up the URLs, so we need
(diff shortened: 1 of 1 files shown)


src/actions/sources/newSources.js
```diff
@@ -81,12 +81,16 @@ function loadSourceMap(sourceId: SourceId) {
     }
 
     if (!urls) {
+      // The source might have changed while we looked up the URLs, so we
```
darkwing: Wow, nice find.  I guess my only suggestion would be being more specific in the comment, as to what could have possibly changed with the source.  Loading status?

PR #7331, opened 2018-11-22
WIP: Column breakpont context menu

darkwing:
This is just a progress marker WIP.  One thing I've noticed is that the column number provided by right-clicking a marker is one off of the breakpoint position, so it's possible that the marker itself counts as a character, and thus breakpoints aren't being found because of the column number being off by however many markers are present.  Hopefully @jasonLaster' switch to widgets fixes the location issue.

### DIFF
--- a/src/actions/ui.js
+++ b/src/actions/ui.js
@@ -155,14 +155,17 @@ export function clearHighlightLineRange() {
   };
 }
 
-export function openConditionalPanel(line: ?number) {
+export function openConditionalPanel(line: ?number, column: ?number) {
+  console.log("openConditionalPanel", line, column);
+
   if (!line) {
     return;
   }
 
   return {
     type: "OPEN_CONDITIONAL_PANEL",
-    line
+    line,
+    column
   };
 }
 
--- a/src/components/Editor/CallSite.js
+++ b/src/components/Editor/CallSite.js
@@ -32,9 +32,11 @@ export default class CallSite extends Component<Props> {
 
   addCallSite = (nextProps: ?Props) => {
     const { editor, callSite, breakpoint, source } = nextProps || this.props;
-    const className = !breakpoint ? "call-site" : "call-site-bp";
-    const sourceId = source.id;
-    const editorRange = toEditorRange(sourceId, callSite.location);
+    let className = breakpoint ? "call-site-bp" : "call-site";
+    className = `${className} cl-${callSite.location.start.line}-${
+      callSite.location.start.column
+    }`;
+    const editorRange = toEditorRange(source.id, callSite.location);
     this.marker = markText(editor, className, editorRange);
   };
 
--- a/src/components/Editor/GutterMenu.js
+++ b/src/components/Editor/GutterMenu.js
@@ -25,6 +25,7 @@ type Props = {
 export function gutterMenu({
   breakpoint,
   line,
+  column,
   event,
   isPaused,
   toggleBreakpoint,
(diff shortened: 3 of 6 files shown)


src/actions/ui.js
```diff
@@ -155,14 +155,17 @@ export function clearHighlightLineRange() {
   };
 }
 
-export function openConditionalPanel(line: ?number) {
+export function openConditionalPanel(line: ?number, column: ?number) {
+  console.log("openConditionalPanel", line, column);
```
jasonLaster: loggg

src/actions/ui.js
```diff
@@ -155,14 +155,17 @@ export function clearHighlightLineRange() {
...
+
   if (!line) {
     return;
   }
 
   return {
     type: "OPEN_CONDITIONAL_PANEL",
-    line
+    line,
+    column
```
jasonLaster: maybe we should save this as a location?

`location: {line, column, sourceId}`

The benefit of a sourceId is that 

a) it is a real location
b) in the future you can change sources it will go away and then come back and it will be there...

src/components/Editor/CallSite.js
```diff
@@ -32,9 +32,11 @@ export default class CallSite extends Component<Props> {
...
   addCallSite = (nextProps: ?Props) => {
     const { editor, callSite, breakpoint, source } = nextProps || this.props;
-    const className = !breakpoint ? "call-site" : "call-site-bp";
-    const sourceId = source.id;
-    const editorRange = toEditorRange(sourceId, callSite.location);
+    let className = breakpoint ? "call-site-bp" : "call-site";
+    className = `${className} cl-${callSite.location.start.line}-${
+      callSite.location.start.column
+    }`;
+    const editorRange = toEditorRange(source.id, callSite.location);
```
jasonLaster: looks out of date

src/components/Editor/GutterMenu.js
```diff
@@ -139,16 +140,36 @@ class GutterContextMenuComponent extends Component {
...
+
+    const classMatch = event.target.className.match(/cl-([0-9]+)-([0-9]+)/);
+    let line = 0;
+    let column = undefined;
+    if (classMatch) {
+      line = classMatch[1];
+      column = classMatch[2];
+    } else {
+      line = lineAtHeight(props.editor, sourceId, event);
+    }
```
jasonLaster: yikes...

okay - so two things

1. if the user right clicks anywhere in the line we can either not show the breakpoint options or use the first pause point colum
2. if the user right clicks on a column breakpoint we can intercept use the line/column of that column breakpoint

This regex seems evil... and lets find a way not to rely on codemirror classes...

src/components/Editor/GutterMenu.js
```diff
@@ -139,16 +140,36 @@ class GutterContextMenuComponent extends Component {
...
+
     const breakpoint = nextProps.breakpoints.find(
-      bp => bp.location.line === line
+      bp => bp.location.line === line && bp.location.column === column
+    );
```
jasonLaster: same here, lets try and go off of pause points or visible column breakpoints which the component could have access too.

src/components/Editor/GutterMenu.js
```diff
@@ -139,16 +140,36 @@ class GutterContextMenuComponent extends Component {
...
+
+    window.BREAKS = nextProps.breakpoints;
+
+    console.log(
+      "showMenu event is: ",
+      line,
+      column,
+      breakpoint,
+      nextProps.breakpoints
```
jasonLaster: yep...

src/reducers/ui.js
```diff
@@ -37,7 +37,10 @@ export type UIState = {
     end?: number,
     sourceId?: number
   },
-  conditionalPanelLine: null | number
+  conditionalPanelLine: {
```
jasonLaster: i think this should be:

`conditionalPanelLocation: Location | null`

darkwing:
This was meant as a proof of concept.  Closing as it will require a different plan.

PR #7335, opened 2018-11-23
Able to blackbox a script from multiple places WIP

### LINKED ISSUE #7287: Context menu Source actions should be available from the SourcePanel and Source Tab
Someone was asking on IRC how one could blackbox a script, so I tried myself by: 
- doing a right click on the sources panel, on the script I wanted to blackbox: -> no luck
- opening the source, right click on the source tab -> no luck
- right click *in* the source content -> jackpot

So this is pretty confusing to me that we have to open a source and specifically right-click in the content.
We should be able to perform any source-related action from whatever represent a source in the UI (so, from the sources panel and the source header, maybe from the callstack as well)

I think it's pretty important because at the moment users might miss that we have such feature

Doomwyte:
Fixes #7287 WIP

### Summary of Changes

* Blackbox a script from the sources panel WIP

### Screenshots/Videos 
screen shot 2018-11-22 at 5 20 35 pm

### DIFF
--- a/src/actions/sources/blackbox.js
+++ b/src/actions/sources/blackbox.js
@@ -16,12 +16,16 @@ import type { Source } from "../../types";
 import type { ThunkArgs } from "../types";
 
 export function toggleBlackBox(source: Source) {
+  console.log("~~~~Entering function toggleBlackBox and outputting source~~~~");
+  console.log(source);
   return async ({ dispatch, getState, client, sourceMaps }: ThunkArgs) => {
     const { isBlackBoxed, id } = source;
 
     if (!isBlackBoxed) {
+      console.log("~~~~!isBlackboxed~~~~");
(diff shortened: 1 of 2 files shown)


AnshulMalik:
I see that you are importing the action from actions file and directly calling it. One of my friends did the same thing yesterday and we lost two hours debugging the same :P. 

You should bind the action like [these](https://github.com/devtools-html/debugger.html/pull/7335/files#diff-45dfb11e0a09c9a6a049d49c7fc99d60R249) and use it from props.

Doomwyte:
Awesome I'll give that a try, thanks!

Doomwyte:
Clicking on blackbox source in the sources panel now works. However, it does not update and I am unable to unblackbox the source through the sources panel without refreshing the page. Working on it.

PR #7342, opened 2018-11-25
Function indexes

### LINKED ISSUE #6580: [Breakpoints] Breakpoint changes its place on page refresh

bomsy:
Fixes #6580 

### Summary of Changes

* Add a `index` property to uniquely identify the function scope 

### Test Plan
- Load https://lumpy-plier.glitch.me/
- Open the debugger and create a breakpoint on line 11 return 42
- Refresh the page
- The breakpoint is now on line 6 -> BUG


src/utils/breakpoint/astBreakpointLocation.js
```diff
@@ -17,7 +17,7 @@ export function getASTLocation(
   location: Location
 ): ASTLocation {
   if (source.isWasm || !symbols || symbols.loading) {
-    return { name: undefined, offset: location };
+    return { name: undefined, offset: location, index: undefined };
```
jasonLaster: maybe we should default to 0?

bomsy: sure

src/actions/tests/helpers/breakpoints.js
```diff
@@ -19,7 +19,8 @@ export function mockPendingBreakpoint(overrides = {}) {
       name: undefined,
       offset: {
         line: line || 5
-      }
+      },
+      index: undefined
```
jasonLaster: should this be undefined or 0?

bomsy: oooppps missed this!

jasonLaster:
Sorry for the delay in landing this. This is a super nice bug fix!

PR #7347, opened 2018-11-27
Add an onInvokeGetterButtonClick prop to Accessor rep

nchevobbe:


nchevobbe:
@jasonLaster this is a spinoff of https://github.com/devtools-html/debugger.html/pull/7255 , should make the latter a bit smaller for the next review round

PR #7324, opened 2018-11-21
Set breakpoint to first pause point when gutter clicked

darkwing:
This is an early WIP for finding the first pause point on a given line when none is provided at the time of breakpoint creation.

PR #7352, opened 2018-11-27
Set breakpoint at first pause point when gutter line is clicked

darkwing:
Supercedes https://github.com/devtools-html/debugger.html/pull/7324

Focuses only on choosing the first pause point when the gutter line is clicked.

### DIFF
--- a/packages/devtools-source-map/src/source-map.js
+++ b/packages/devtools-source-map/src/source-map.js
@@ -195,6 +195,19 @@ async function getGeneratedLocation(
       column: location.column == null ? 0 : location.column,
       bias: SourceMapConsumer.LEAST_UPPER_BOUND
     });
+
+    if (map.originalPositionFor(match).line !== location.line) {
+      const previousMatch = map.generatedPositionFor({
+        source: originalSource.url,
+        line: location.line,
+        column: location.column == null ? 0 : location.column,
+        bias: SourceMapConsumer.GREATEST_LOWER_BOUND
+      });
+
+      if (map.originalPositionFor(previousMatch).line === location.line) {
+        match = previousMatch;
+      }
+    }
   }
 
   return {
--- a/src/actions/breakpoints/addBreakpoint.js
+++ b/src/actions/breakpoints/addBreakpoint.js
@@ -16,7 +16,7 @@ import {
   getSource,
   getSymbols,
   getBreakpoint,
-  getPausePoints
+  getFirstPausePointLocation
 } from "../../selectors";
 import { getGeneratedLocation } from "../../utils/source-maps";
 import { getTextAtPosition } from "../../utils/source";
@@ -151,16 +151,11 @@ export function addBreakpoint(
     recordEvent("add_breakpoint");
 
     if (location.column === undefined) {
-      const pausePoints = getPausePoints(getState(), location.sourceId);
-
-      if (pausePoints) {
-        const pausesAtLine = pausePoints[location.line];
-
-        const column = pausesAtLine
-          ? Object.keys(pausesAtLine).find(col => pausesAtLine[col].break)
-          : undefined;
-        location = { ...location, column };
-      }
+      const pausePointLocation = getFirstPausePointLocation(
+        getState(),
+        location
+      );
+      location = pausePointLocation;
     }
 
     const breakpoint = createBreakpoint(location, { condition, hidden });
--- a/src/actions/breakpoints/addBreakpoint.js
+++ b/src/actions/breakpoints/addBreakpoint.js
@@ -12,7 +12,12 @@ import {
   assertLocation
 } from "../../utils/breakpoint";
 import { PROMISE } from "../utils/middleware/promise";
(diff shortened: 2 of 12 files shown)


src/actions/breakpoints/addBreakpoint.js
```diff
@@ -142,10 +147,24 @@ export function addBreakpoint(
...
+  return async ({ dispatch, getState, sourceMaps, client }: ThunkArgs) => {
     recordEvent("add_breakpoint");
 
+    if (location.column === undefined) {
+      const pausePoints = getPausePoints(getState(), location.sourceId);
+
+      if (pausePoints) {
+        const pausesAtLine = pausePoints[location.line];
+
+        const column = pausesAtLine
```
darkwing: ToDo:  Add selector for getting the first pause point.

src/reducers/ast.js
```diff
@@ -207,6 +207,32 @@ export function getPausePoints(
...
+  if (!location) {
+    return;
+  }
+
+  const pausePoints = getPausePoints(state, location.sourceId);
+  if (!pausePoints) {
+    return location;
+  }
+
+  const pausesAtLine = pausePoints[String(location.line)];
```
jasonLaster: the key should be a number, if we're not doing that we should fix it

src/reducers/ast.js
```diff
@@ -207,6 +207,32 @@ export function getPausePoints(
   return state.ast.pausePoints.get(sourceId);
 }
 
+export function getFirstPausePointLocation(
+  state: OuterState,
+  location: ?Location
+): ?Location {
+  if (!location) {
+    return;
+  }
```
jasonLaster: can we require a loc?

src/reducers/ast.js
```diff
@@ -207,6 +207,32 @@ export function getPausePoints(
...
+    return location;
+  }
+
+  const pausesAtLine = pausePoints[String(location.line)];
+  if (pausesAtLine) {
+    const column = Object.keys(pausesAtLine).find(
+      col => pausesAtLine[col].break
+    );
+    if (column !== undefined) {
+      return { ...location, column: Number(column) };
```
jasonLaster: we should have a number :)

i bet i recently broke this

src/reducers/ast.js
```diff
@@ -207,6 +207,32 @@ export function getPausePoints(
...
+  }
+
+  const pausePoints = getPausePoints(state, location.sourceId);
+  if (!pausePoints) {
+    return location;
+  }
+
+  const pausesAtLine = pausePoints[String(location.line)];
+  if (pausesAtLine) {
+    const column = Object.keys(pausesAtLine).find(
```
jasonLaster: would Object.values be simpler?

darkwing: No, because I wouldn't have access to the key (line number), which is what I need.

src/utils/pause/pausePoints.js
```diff
@@ -22,8 +22,8 @@ function insertStrtAt(string, index, newString) {
 export function convertToList(pausePoints: PausePoints): PausePoint[] {
   const list = [];
   for (const line in pausePoints) {
-    for (const column in pausePoints[line]) {
-      const point = pausePoints[line][column];
+    for (const column in pausePoints[Number(line)]) {
+      const point = pausePoints[Number(line)][Number(column)];
```
jasonLaster: I think we can simplify this a lot if we have a location field in pause points from the beginning:

```diff
diff --git a/src/workers/parser/pausePoints.js b/src/workers/parser/pausePoints.js
index fa12cd70b..ef6bc40be 100644
--- a/src/workers/parser/pausePoints.js
+++ b/src/workers/parser/pausePoints.js
@@ -170,17 +170,19 @@ function hasPoint(state, { line, column }) {

 function addPoint(
   state,
-  { line, column },
+  location,
   types: boolean | { break?: boolean, step?: boolean }
 ) {
   if (typeof types === "boolean") {
     types = { step: types, break: types };
   }

+  const { line, column } = location;
+
   if (!state[line]) {
     state[line] = {};
   }
-  state[line][column] = types;
+  state[line][column] = { types, location };
   return state;
 }
```

I added location in my map pause points PR, but I don't think i was thinking [big] enough. I really should have done it in the worker so we always have it.

[big]: https://github.com/devtools-html/debugger.html/pull/7349/files#diff-44b80fbbfa2e03e79842d1e4b853ff20R42

packages/devtools-source-map/src/source-map.js
```diff
@@ -195,6 +195,19 @@ async function getGeneratedLocation(
       column: location.column == null ? 0 : location.column,
       bias: SourceMapConsumer.LEAST_UPPER_BOUND
     });
+
+    if (map.originalPositionFor(match).line !== location.line) {
```
darkwing: @loganfsmyth Is this still relevant or should we bail?

loganfsmyth: I think we should ditch this, since using the generated location + only showing things when there are more than a single column bp should hopefully avoid the need for it.

PR #7355, opened 2018-11-28
Restyling breakpoint colors

audreyfang:
### Summary of Changes

* Break point colors have been changed and a border has been added according to this mockup:
newdesign

* There's still one issue that I'm not sure how to fix. The --blue-40 variable is not set, so those borders don't actually appear.

### Test Plan
- [x] Click on various breakpoints in both light and dark mode.

### Screenshots/Videos
breakpoint

### DIFF
--- a/src/components/Editor/ColumnBreakpoint.js
+++ b/src/components/Editor/ColumnBreakpoint.js
@@ -26,12 +26,12 @@ type Props = {
 };
 
 const breakpointImg = document.createElement("div");
+ReactDOM.render(<Svg name={"column-marker"} />, breakpointImg);
 function makeBookmark(isActive, { onClick }) {
   const bp = breakpointImg.cloneNode(true);
   const className = isActive ? "active" : "disabled";
   bp.className = classnames("call-site", className);
   bp.onclick = onClick;
-  ReactDOM.render(<Svg name={"column-marker"} />, bp);
   return bp;
 }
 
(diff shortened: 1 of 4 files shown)


darkwing:
Great start!  A few points:

1.  I'm seeing the marker above the paused line:  
breapointposition

2.  The use of `<SVG>` we've found to be a performance hit; this would be much faster if it were a `<div>` that used the SVG as a background image.

3.  We need to add the missing color to our launchpad -- don't worry about that.

Great start!

src/components/Editor/ColumnBreakpoint.js
```diff
@@ -26,10 +28,10 @@ type Props = {
 const breakpointImg = document.createElement("div");
 function makeBookmark(isActive, { onClick }) {
   const bp = breakpointImg.cloneNode(true);
-  bp.className = classnames("call-site", {
-    active: isActive
-  });
+  const className = isActive ? "active" : "disabled";
+  bp.className = classnames("call-site", className);
   bp.onclick = onClick;
+  ReactDOM.render(<Svg name={"column-marker"} />, bp);
```
jasonLaster: I think for this to be fast we'll need to keep the SVG so we can set stroke and fill, but do what we do in the Gutter w/ breakpoints and have one SVG and clone it every time

https://github.com/devtools-html/debugger.html/blob/master/src/components/Editor/Breakpoint.js#L17-L18

darkwing:
Awesome!!!

PR #7376, opened 2018-11-30
Fix conditional breakpoint marker stroke

darkwing:
## Before

beforecolcond

## After

orange

The orange color used wasn't provided by @mcroud but we should address the yucky blue ASAP


darkwing:
Additionally, all column breakpoints were also given orange, which we don't want.  At present, the `ColumnBreakpoint` type doesn't include `condition`, so we can't color specific column breakpoints yet.  That can be a separate PR.

PR #7366, opened 2018-11-28
Change conditional panel to use location

darkwing:
Supercedes https://github.com/devtools-html/debugger.html/pull/7324 for having the conditional breakpoint input update the correct breakpoint.

PR #7377, opened 2018-11-30
Provide conditional comment styling

darkwing:
conditionalcolumnbps

This lays the groundwork for having column breakpoints show the proper color when they have a condition.  You have to hardcode a condition in the code now to test until https://github.com/devtools-html/debugger.html/pull/7366 is merged.

### DIFF
--- a/src/selectors/visibleColumnBreakpoints.js
+++ b/src/selectors/visibleColumnBreakpoints.js
@@ -14,7 +14,8 @@ import type { SourceLocation } from "../types";
 
 export type ColumnBreakpoint = {|
   +location: SourceLocation,
-  +enabled: boolean
+  +enabled: boolean,
+  +condition: ?string
 |};
 
 function contains(location, range) {
@@ -101,10 +102,20 @@ export function getColumnBreakpoints(pausePoints, breakpoints, viewport) {
     ({ location: { line } }) => lineCount[line] > 1
   );
 
-  return columnBreakpoints.map(({ location }) => ({
-    location,
-    enabled: isEnabled(location, breakpointMap)
-  }));
+  return columnBreakpoints.map(({ location }) => {
+    const foundBreakpoint = breakpoints.find(breakpoint => {
+      return (
+        breakpoint.location.line === location.line &&
+        breakpoint.location.column === location.column
+      );
+    });
+
+    return {
+      location,
+      enabled: isEnabled(location, breakpointMap),
+      condition: foundBreakpoint ? foundBreakpoint.condition : ""
+    };
+  });
 }
 
 export const visibleColumnBreakpoints = createSelector(
--- a/src/selectors/visibleColumnBreakpoints.js
+++ b/src/selectors/visibleColumnBreakpoints.js
@@ -2,7 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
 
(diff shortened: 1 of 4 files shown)


src/selectors/visibleColumnBreakpoints.js
```diff
@@ -101,10 +102,20 @@ export function getColumnBreakpoints(pausePoints, breakpoints, viewport) {
     ({ location: { line } }) => lineCount[line] > 1
   );
 
-  return columnBreakpoints.map(({ location }) => ({
-    location,
-    enabled: isEnabled(location, breakpointMap)
-  }));
+  return columnBreakpoints.map(({ location }) => {
+    const foundBreakpoint = breakpoints.find(breakpoint => {
```
loganfsmyth: I think it is fine as a first pass, but as column bps move forward, I'm going to be very curious to see how often this fails to match. I could easily see our breakpoints sliding a causing thins to fail to find the condition.

src/selectors/visibleColumnBreakpoints.js
```diff
@@ -101,10 +102,20 @@ export function getColumnBreakpoints(pausePoints, breakpoints, viewport) {
...
+      return (
+        breakpoint.location.line === location.line &&
+        breakpoint.location.column === location.column
+      );
+    });
+
+    return {
+      location,
+      enabled: isEnabled(location, breakpointMap),
+      condition: foundBreakpoint ? foundBreakpoint.condition : ""
```
loganfsmyth: Should we default this to `null`? I see the type is `?string`, but it's not clear if there is a difference between empty string and null/undefined here.

src/selectors/visibleColumnBreakpoints.js
```diff
@@ -101,10 +102,20 @@ export function getColumnBreakpoints(pausePoints, breakpoints, viewport) {
...
-  return columnBreakpoints.map(({ location }) => ({
-    location,
-    enabled: isEnabled(location, breakpointMap)
-  }));
+  return columnBreakpoints.map(({ location }) => {
+    const foundBreakpoint = breakpoints.find(breakpoint => {
+      return (
+        breakpoint.location.line === location.line &&
+        breakpoint.location.column === location.column
+      );
```
jasonLaster: lets use lodash's isEqual

src/selectors/visibleColumnBreakpoints.js
```diff
@@ -101,10 +102,20 @@ export function getColumnBreakpoints(pausePoints, breakpoints, viewport) {
...
+    });
+
+    return {
+      location,
+      enabled: isEnabled(location, breakpointMap),
```
jasonLaster: isn't isEnabled trying to find the breakpoint too?

can we re-use this?

src/selectors/visibleColumnBreakpoints.js
darkwing: We're comparing these objects so that seems like overkill:

nahh

darkwing:
This is blocked on https://github.com/devtools-html/debugger.html/pull/7366 ;

src/selectors/visibleColumnBreakpoints.js
darkwing: Good shout; we'll be able to move to `.get` instead of `hasIn` when https://github.com/devtools-html/debugger.html/pull/7366 lands.

darkwing:
Feedback addressed!

darkwing:
Outside of @loganfsmyth's comment, this is ready for re-review.  Can we document the ways that his case would happen, and address that separately?

PR #7384, opened 2018-11-30
Enable getter evaluation in ObjectInspector.

### LINKED ISSUE #6140: Add a way to evaluate getter properties

nchevobbe:
This makes use of ObjectClient's getPropertyValue function.
A new evaluations property is added in the OI redux store,
which the returned value of the invokation.

Fixes #6140
---

Looks like Github could not continue on https://github.com/devtools-html/debugger.html/pull/7255


nchevobbe:
TRY https://treeherder.mozilla.org/#/jobs?repo=try&revision=4653a73776e0ed590a2d8f953d289382a615736f

nchevobbe:
flow is unhappy, i'll fix that this afternoon (waiting for TRY results see if everything is okay)

packages/devtools-reps/src/object-inspector/components/ObjectInspectorItem.js
```diff
@@ -163,6 +167,18 @@ class ObjectInspectorItem extends Component<Props> {
         };
       }
 
+      if (nodeHasGetter(item)) {
+        const parentNode = getParent(item);
+        const parentGripNode = parentNode && getClosestGripNode(parentNode);
+        const parentGrip = parentGripNode && getValue(parentGripNode);
```
jasonLaster: nit, but maybe we move this to a helper: `getParentGrip(item)`.

packages/devtools-reps/src/object-inspector/components/ObjectInspectorItem.js
```diff
@@ -163,6 +167,18 @@ class ObjectInspectorItem extends Component<Props> {
...
+        if (parentGrip) {
+          Object.assign(repProps, {
+            onInvokeGetterButtonClick: () =>
```
jasonLaster: still not sure why it is conditionally availabe. Could we have an onClick callback that does this check and then invokeGetter?

packages/devtools-reps/src/object-inspector/components/ObjectInspector.js
```diff
@@ -130,25 +147,58 @@ class ObjectInspector extends Component<Props> {
...
+    }
+
+    return [];
   }
 
   getRoots(): Array<Node> {
     return this.props.roots;
   }
 
+  getEvaluatedItem(item: Node): Node {
```
jasonLaster: it might be simpler to have an evaluations utils module that has some of this tree node logic...

packages/devtools-reps/src/object-inspector/components/ObjectInspector.js
```diff
@@ -130,25 +147,58 @@ class ObjectInspector extends Component<Props> {
...
       cachedNodes,
       item
     });
+    if (Array.isArray(children)) {
+      return children.map(i => this.getEvaluatedItem(i));
+    }
+
+    if (children) {
+      return this.getEvaluatedItem(children);
+    }
```
jasonLaster: this still seems a bit complicated...

packages/devtools-reps/src/object-inspector/components/ObjectInspector.js
```diff
@@ -93,21 +97,34 @@ class ObjectInspector extends Component<Props> {
       if (this.props.rootsChanged) {
         this.props.rootsChanged();
       }
+      return;
+    }
+
+    // if there are new evaluations, we want to remove the existing cached
+    // nodes from the cache.
+    if (nextProps.evaluations.size > this.props.evaluations.size) {
```
jasonLaster: could we do an identity check?

jasonLaster: maybe move this up to a small utility `removeOldNodes`

packages/devtools-reps/src/object-inspector/reducer.js
```diff
@@ -49,6 +50,20 @@ function reducer(
...
+  if (type === "GETTER_INVOKED") {
+    return cloneState({
+      actors: data.actor
+        ? new Set(state.actors || []).add(data.result.from)
+        : state.actors,
+      evaluations: new Map(state.evaluations).set(data.node.path, {
+        getterValue:
+          data.result &&
+          data.result.value &&
+          (data.result.value.return || data.result.value.throw)
```
jasonLaster: what's this doing?

packages/devtools-reps/src/object-inspector/components/ObjectInspectorItem.js
nchevobbe: sure

packages/devtools-reps/src/object-inspector/components/ObjectInspectorItem.js
nchevobbe: because we only want to show the button in expanded nodes.
So given the following object: 

```js
x = {
  get a() {}
}
```

when printed, we display: `▶︎ Object { a : Getter}`. Here we don't want to be able to invoke the getter, because it would add too much complexity for now.

But we do want to show it when expanded, i.e. 
```
▼ Object
| a: Getter >>
| <get a>: function
```

does it make sense ?

packages/devtools-reps/src/object-inspector/components/ObjectInspector.js
nchevobbe: > could we do an identity check?

sure

> maybe move this up to a small utility

okay

packages/devtools-reps/src/object-inspector/components/ObjectInspector.js
nchevobbe: The alternative would be to pass `evaluations` to `getChildren`, and pipe it through all the different function that can create nodes, which I find more complex than that.

Can you think of anything else that would make this more simple?

packages/devtools-reps/src/object-inspector/components/ObjectInspector.js
nchevobbe: I'm not a 100% convinced here. We could have a util function for that, e.g. `getEvaluatedItem(item: Node, evaluations: Map)` , but the only consumer would eventually be this component.

Am I missing something?

packages/devtools-reps/src/object-inspector/reducer.js
nchevobbe: assigning the right result to `getterValue`, since the getter might throw.
This can be moved elsewhere in a small util function (or in the action)

packages/devtools-reps/src/object-inspector/components/ObjectInspector.js
nchevobbe: moved to node utils

packages/devtools-reps/src/object-inspector/components/ObjectInspector.js
nchevobbe: moved to node utils

PR #7391, opened 2018-11-30
Pass explicit column values for non-breakpoint location usage.

loganfsmyth:
More prep work around breakpoint column. This defaults to `column: 0` for all the places that reference file locations but _aren't_ breakpoints.

### DIFF
--- a/src/actions/sources/newSources.js
+++ b/src/actions/sources/newSources.js
@@ -129,7 +129,10 @@ function checkSelectedSource(sourceId: string) {
           sourceId: source.id,
           line:
             typeof pendingLocation.line === "number" ? pendingLocation.line : 0,
-          column: pendingLocation.column
+          column:
+            typeof pendingLocation.column === "number"
+              ? pendingLocation.column
+              : 0
         })
       );
     }
--- a/src/utils/source-maps.js
+++ b/src/utils/source-maps.js
@@ -33,7 +33,12 @@ export async function getGeneratedLocation(
   return {
     line,
     sourceId,
-    column: column === 0 ? undefined : column,
+    // If the input location didn't have a column number, and we got back
+    // the start of the line, we consider the result to also not have a column.
+    // This is so that we can preserve our current behavior of having
+    // breakpoints with optional columns, while allowing operations on
+    // locations unrelated to breakpoints to always have a column.
+    column: location.column === undefined && column === 0 ? undefined : column,
     sourceUrl: generatedSource.url
   };
 }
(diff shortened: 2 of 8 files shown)


darkwing:
Were we going to pref this?

loganfsmyth:
I'll pref the breakpoint column location, but I don't think we need to pref the non-breakpoint `0`-default stuff. Does that seem reasonable?

src/actions/sources/newSources.js
```diff
@@ -129,7 +129,10 @@ function checkSelectedSource(sourceId: string) {
           sourceId: source.id,
           line:
             typeof pendingLocation.line === "number" ? pendingLocation.line : 0,
-          column: pendingLocation.column
+          column:
+            typeof pendingLocation.column === "number"
+              ? pendingLocation.column
+              : 0
```
jasonLaster: could we do this when we set `pendingLocation.column`

src/utils/source-maps.js
```diff
@@ -33,7 +33,12 @@ export async function getGeneratedLocation(
   return {
     line,
     sourceId,
-    column: column === 0 ? undefined : column,
+    // If the input location didn't have a column number, and we got back
+    // the start of the line, we consider the result to also not have a column.
+    // This is so that we can preserve our current behavior of having
+    // breakpoints with optional columns, while allowing operations on
+    // locations unrelated to breakpoints to always have a column.
+    column: location.column === undefined && column === 0 ? undefined : column,
```
jasonLaster: is it fair to change `getGeneratedLocation` to always have a column and assume we get a good column from  location?

src/actions/sources/newSources.js
loganfsmyth: Sure!

src/utils/source-maps.js
loganfsmyth: Hmm, I think if we are concerned about things breaking and want it behind a pref, this is definitely one we'd want behind a pref. My intention with this is that the breakpoint pref would leave `column: undefined` and therefore propagate through into here, but if the pref for default-0 is one, we'll always get a column out of this too.

src/actions/sources/newSources.js
loganfsmyth: Hmmm, this is actually a bit difficult because this datatype is shared between both bending location selections, and pending breakpoints. Since we still want to hide the breakpoint logic behind a pref, breakpoint pending locations need to continue having an optional column in Flow at least, which means that even if we move this default setting to the creating of the pending non-breakpoint locations, we'd still need an assertion in here checking the type anyway to make sure it's always set, or something.

PR #7392, opened 2018-11-30
Set breakpoint at first pause point when gutter line is clicked

darkwing:
Supercedes https://github.com/devtools-html/debugger.html/pull/7352 since so much had changed since the original.

ToDo:
- [ ] Use `generatedLocation` to avoid showing column breakpoint markers where they'd actually get shifted to the next line
- [ ] Use a selector for `getFirstPausePointLocation`


src/reducers/ast.js
```diff
@@ -221,12 +221,32 @@ export function getPausePoint(
...
+  }
+  const pausesAtLine = pausePoints[location.line];
+  if (pausesAtLine) {
+    const column = Object.keys(pausesAtLine).find(
+      col => pausesAtLine[Number(col)].types.break
+    );
+    if (column !== undefined) {
+      return { ...location, column: Number(column) };
+    }
+  }
```
jasonLaster: minor point, but i think you can do  
```js
const point = Object.values(pausesAtLine).find(point => point.types.break)
if (point) {
  return point.location
}
```

or some variant of this

src/utils/pause/pausePoints.js
```diff
@@ -22,8 +22,8 @@ function insertStrtAt(string, index, newString) {
 export function convertToList(pausePoints: PausePoints): PausePoint[] {
   const list = [];
   for (const line in pausePoints) {
-    for (const column in pausePoints[line]) {
-      list.push(pausePoints[line][column]);
+    for (const column in pausePoints[Number(line)]) {
+      list.push(pausePoints[Number(line)][Number(column)]);
```
jasonLaster: if you rebase this should not need to change...

src/utils/pause/pausePoints.js
```diff
@@ -50,7 +50,7 @@ export async function mapPausePoints<T>(
 
   const newPausePoints = {};
   for (const line in pausePoints) {
-    const linePoints = pausePoints[line];
+    const linePoints = pausePoints[Number(line)];
```
jasonLaster: this should not be needed ...

src/workers/parser/types.js
```diff
@@ -13,5 +13,5 @@ export type PausePoint = {|
   generatedLocation: AstPosition
 |};
 export type PausePoints = {
-  [line: string]: { [column: string]: PausePoint }
+  [line: number]: { [column: number]: PausePoint }
```
jasonLaster: hmm, why is all of this changing 😉

darkwing:
With regard to the sourcemaps test failure, whereby the test wants `line 18, column 6`, but instead gets `line 18, column 10`, I don't think that this PR does anything wrong.  The only pause point found on said line is at column 10, so I need advice on where to go.  Do we need to update our pause point detection code to include whatever condition hits column 6, or should the test be adjusted?  @loganfsmyth

darkwing:
Closing this until we're sold on actually doing this.

src/actions/breakpoints/addBreakpoint.js
```diff
@@ -152,10 +157,15 @@ export function addBreakpoint(
   location: SourceLocation,
   { condition, hidden }: addBreakpointOptions = {}
 ) {
-  const breakpoint = createBreakpoint(location, { condition, hidden });
-  return ({ dispatch, getState, sourceMaps, client }: ThunkArgs) => {
+  return async ({ dispatch, getState, sourceMaps, client }: ThunkArgs) => {
     recordEvent("add_breakpoint");
 
+    if (location.column === undefined) {
+      location = getFirstPausePointLocation(getState(), location);
```
loganfsmyth: Did we want this behavior toggled based on the column breakpoint feature flag?

darkwing: Probably a good idea; I'll add it!

PR #7394, opened 2018-11-30
[Parser] Support object destructuring in mapOriginalExpression 

### LINKED ISSUE #7274: [Parser] mapOriginalExpression does not handle destructuring assignments well
When working on #7222 I found out that destructuring assignments are not handled properly by `mapOriginalExpression` function.

I created the following test case for `mapOriginalExpression` function and it does not pass.

```js
it("object destructuring", () => {
  const generatedExpression = mapOriginalExpression("({ a } = { a: 4 })", {
    a: "_mod.foo",
  });
  
  expect(generatedExpression).toEqual("({ a: _mod.foo } = { a: 4 })");
});
```

This is the actual output I received:

```js
({ _mod.foo: _mod.foo } = {  a: 4 });
``` 
It's obviously incorrect (it's not even a valid syntax).

Expected state:

Mapping is done correctly, i.e.
`({ a: _mod.foo } = { a: 4 })`

or `_mod.foo = ({ a: 4 }).a`

or no mapping is done at all (?).

P0lip:
Fixes #7274

### Summary of Changes

* Add special case covering ObjectProperty

### Test Plan

No idea, I relied on tests.

### DIFF
--- a/src/workers/parser/mapOriginalExpression.js
+++ b/src/workers/parser/mapOriginalExpression.js
@@ -10,14 +10,11 @@ import { buildScopeList } from "./getScopes";
 import generate from "@babel/generator";
 import * as t from "@babel/types";
 
-// NOTE: this will only work
-// if we are replacing an original identifier or an object property
+// NOTE: this will only work if we are replacing an original identifier
 function replaceNode(ancestors, node) {
   const ancestor = ancestors[ancestors.length - 1];
 
-  if (t.isObjectProperty(ancestor.node)) {
-    ancestor.node.value = node;
-  } else if (typeof ancestor.index === "number") {
+  if (typeof ancestor.index === "number") {
     ancestor.node[ancestor.key][ancestor.index] = node;
   } else {
     ancestor.node[ancestor.key] = node;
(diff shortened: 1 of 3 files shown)


src/workers/parser/mapOriginalExpression.js
```diff
@@ -10,11 +10,14 @@ import { buildScopeList } from "./getScopes";
...
 import * as t from "@babel/types";
 
-// NOTE: this will only work if we are replacing an original identifier
+// NOTE: this will only work
+// if we are replacing an original identifier or an object property
 function replaceNode(ancestors, node) {
   const ancestor = ancestors[ancestors.length - 1];
 
-  if (typeof ancestor.index === "number") {
+  if (t.isObjectProperty(ancestor.node)) {
```
loganfsmyth: I don't think this approach is the right way to go about this.

There should already be a `node.value` in the AST, AFAIK, but it looks like we are failing to find it, I assume because it has the same location as the existing `node.key`.

It _might_ actually be enough to do
```
  t.traverse(ast, (node, ancestors) => {
    if (!t.isIdentifier(node) && !t.isThisExpression(node)) {
      return;
    }

    const ancestor = ancestors[ancestors.length - 1];
    if (ancestor.node.type === "ObjectProperty" && ancestor.key !== "value") {
      return;
    }

    // ...
```

src/workers/parser/mapOriginalExpression.js
```diff
@@ -93,6 +93,11 @@ export default function mapOriginalExpression(
       return;
     }
 
+    const ancestor = ancestors[ancestors.length - 1];
+    if (t.isObjectProperty(ancestor.node) && ancestor.key !== "value") {
```
loganfsmyth: Can you add a comment here making it clear why we do this? Something like
```
// Shorthand properties can have a key and value with `node.loc.start` value, and we only want to replace the value.
```

darkwing:
Apologies for amount of time review is taking here @P0lip -- we're all traveling ATM.

PR #7408, opened 2018-12-04
Fix typo in debugger-html-react-redux-overview.md

notjaril:
### Summary of Changes

Fixed spelling: debbuger.html -> debugger.html.

### Test Plan

Added `debbuger.html` string to dictionary.txt so that it can be caught in case the typo occurs again.

### DIFF
--- a/assets/dictionary.txt
+++ b/assets/dictionary.txt
@@ -205,3 +205,4 @@ bytecode
 eval
 asyncStore
 CPG
+debbuger.html
--- a/assets/dictionary.txt
+++ b/assets/dictionary.txt
@@ -205,4 +205,3 @@ bytecode
 eval
 asyncStore
 CPG
-debbuger.html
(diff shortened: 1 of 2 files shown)


assets/dictionary.txt
```diff
@@ -205,3 +205,4 @@ bytecode
 eval
 asyncStore
 CPG
+debbuger.html
```
AnshulMalik: Ohh, it also have two `b`s :P

darkwing:
Good catch @notjaril !  Could you remove the dictionary addition though?  We don't need it.

notjaril:
Fixed. Not a problem - just caught it while getting familiarized with the docs.

assets/dictionary.txt
notjaril: Yes, I added the misspelled version("debbuger.html") to `dictionary.txt`. As I understood it from `docs/pull-requests.md`, the dictionary.txt should have the misspelled words. Is that correct?

This doesn't matter as much as I've removed it, but would be useful to clarify for reference later.

PR #7410, opened 2018-12-04
Make gutter marker or breakpoint represent first breakpoint on line

darkwing:
I'm super excited about this.

At present, a breakpoint gutter item is added for every breakpoint (whether column is `undefined` or present), which makes the gutter weird;

1.  Clicking an active column breakpoint marker removes the gutter marker, even though there are more breakpoints active on that column

2.  It's hard to understand which breakpoint that marker's context menu actually represents.

This PR makes the gutter marker represent a single column breakpoint (the first found in the list);  in the future, we'd want to sort the breakpoints first so that the gutter marker represents the first breakpoint *on that line*, instead of the first one found.

## To Do:
- [x]  Sort breakpoints before `uniqBy` (will be available once `getFirstPausePoint` lands (https://github.com/devtools-html/debugger.html/pull/7392)

### DIFF
--- a/src/components/Editor/Breakpoints.js
+++ b/src/components/Editor/Breakpoints.js
@@ -5,11 +5,12 @@
 // @flow
 import { connect } from "react-redux";
 import React, { Component } from "react";
+import { uniqBy } from "lodash";
 
 import Breakpoint from "./Breakpoint";
 
 import { getSelectedSource, getVisibleBreakpoints } from "../../selectors";
-import { makeLocationId } from "../../utils/breakpoint";
+import { makeLocationId, sortBreakpoints } from "../../utils/breakpoint";
 import { isLoaded } from "../../utils/source";
 
 import type { Breakpoint as BreakpointType, Source } from "../../types";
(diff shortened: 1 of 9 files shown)


darkwing:
gutter

src/components/Editor/Breakpoints.js
```diff
@@ -54,6 +53,8 @@ class Breakpoints extends Component<Props> {
 }
 
 export default connect(state => ({
-  breakpoints: getVisibleBreakpoints(state),
+  // Retrieves only the first breakpoint per line so that the
+  // breakpoint marker represents only the first breakpoint
+  breakpoints: getFirstVisibleBreakpoints(state),
```
jasonLaster: `getFirstVisibleBreakpoint`

darkwing: Yeah, the naming here is a bit hard. `getFirstVisibleBreakpoint` seems to imply it returns only one breakpoint.  Maybe `getFirstVisibleBreakpointPerLine`?

jasonLaster: or `getFirstVisibleBreakpointOnLine` 

per feels like for each line

jasonLaster: good call

jasonLaster: good call

PR #7415, opened 2018-12-04
Make empty lines line number lighter

darkwing:
empty-lines-light

empty-lines-dark

### DIFF
--- a/src/components/Editor/Editor.css
+++ b/src/components/Editor/Editor.css
@@ -66,6 +66,10 @@ html[dir="rtl"] .editor-mount {
   --breakpoint-stroke-disabled: var(--blue-60);
 }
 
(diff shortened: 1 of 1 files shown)


darkwing:
@violasong @mcroud This simply uses 50% opacity to show empty line ; what color do you suggest instead?

AnshulMalik:
IMO, for dark theme, the line numbers are invisible, we can try to make the little more brighter, both empty and non-empty ones.

mcroud:
I have some agreement with @AnshulMalik comment here that Dark mode in particular struggles with disabled colors which saw a lot of experimentation with the column breakpoint colors for dark mode. It may be that we need to look at the non-empty numbers too. Let me make a a single mockup of best fit suggestions and I'll see what feedback I can get.

PR #7420, opened 2018-12-05
Fix tab context menu item size checks

darkwing:
The `.size` checks were undefined since we likely moved away from immutables.  Additionally, I dislike show/hide of menu items; simply disabling them keeps menu integrity while only allowing desired functionality.

### DIFF
--- a/src/components/Editor/Tab.js
+++ b/src/components/Editor/Tab.js
@@ -89,9 +89,9 @@ class Tab extends PureComponent<Props> {
       {
         item: {
           ...tabMenuItems.closeOtherTabs,
-          click: () => closeTabs(otherTabURLs)
-        },
-        hidden: () => tabSources.size === 1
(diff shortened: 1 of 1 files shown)


src/components/Editor/Tab.js
```diff
@@ -89,21 +89,21 @@ class Tab extends PureComponent<Props> {
...
             closeTabs(tabURLs.filter((t, i) => i > tabIndex));
-          }
-        },
-        hidden: () =>
-          tabSources.size === 1 ||
-          tabSources.some((t, i) => t === tab && tabSources.size - 1 === i)
+          },
+          disabled: () =>
+            tabSources.length === 1 ||
+            tabSources.some((t, i) => t === tab && tabSources.length - 1 === i)
```
jasonLaster: maybe we should have a `tabCount` variable?

PR #7421, opened 2018-12-05
Remove 4px gap between the source file name and the primary pane separator

darkwing:
I looked at the PR where this came in and it's unclear why it was added.  In any event, we've changed the styling the three a lot and I don't believe we need it.


darkwing:
## Before
before 1


## After

after1

Issue #5250, opened 2018-01-30
[Sources] The indentation of the files is off


jasonLaster:
It looks like the tree indentation is a bit off, which could come from a variety of places:

1. https://github.com/devtools-html/debugger.html/issues/5239
2. https://github.com/devtools-html/debugger.html/issues/5216

<img src='https://shipusercontent.com/245a1d94ac4f19099bf30f024f224518/Screen%20Shot%202018-01-30%20at%205.17.36%20PM.png' title='Screen Shot 2018-01-30 at 5.17.36 PM.png' width=697>

MikeRatcliffe:
Here is a screenshot from an MDN tutorial for comparison:

![Before](https://user-images.githubusercontent.com/116941/35628962-4c5788da-0695-11e8-83b0-56794003efa3.png)

MikeRatcliffe:
I am debugging this using http://todomvc.com/examples/dojo/ because it has a reasonably deep tree.

MikeRatcliffe:
I have created a pull request at https://github.com/devtools-html/devtools-core/pull/919 so that this can be fixed upstream.

MikeRatcliffe:
In the future we may consider switching to the arrows provided by core but for consistency in the debugger let's make the CSS change. The debugger arrows are a different size than core and I think they fit the debugger better.

lukaszsobek:
Merged. Closing.

PR #7423, opened 2018-12-05
Float cursor position to the right

darkwing:
I don't like this informational lock being flush up against commands for the given source.  It looks a lot better on the opposite side.

cursorright


jasonLaster:
Mind trying this with blackboxed sources as well?

AnshulMalik:
The change was for the indentation fix for this issue https://github.com/devtools-html/debugger.html/issues/5250

The non-expandable nodes go out of the way. May be we can find another simple solution.

PR #7428, opened 2018-12-05
Prevent the breakpoint exception labels from being highlighted

darkwing:
I just noticed that clicking the "Pause on Exception" and "Pause on any xhr" checkboxes quickly results in their text being highlighted, which we don't want.


jasonLaster:
love these!

PR #7455, opened 2018-12-07
Provide initial implementation of opening DOM nodes in inspector

### LINKED ISSUE #1427: Add button to hightlight (hover) or select (click) DOM nodes
The old debugger had way to hightlight/select DOM nodes. Really missing it:

Old debugger:
2016-12-12 11_32_10-program manager

debugger.html:
2016-12-12 11_23_31-html-element

darkwing:
Fixes #1427

This provides a method to open DOM elements from within the Object inspector.  This works in all places the OI is used.

### DIFF
--- a/assets/panel/panel.js
+++ b/assets/panel/panel.js
@@ -35,7 +35,22 @@ DebuggerPanel.prototype = {
       toolboxActions: {
         // Open a link in a new browser tab.
         openLink: this.openLink.bind(this),
-        openWorkerToolbox: this.openWorkerToolbox.bind(this)
+        openWorkerToolbox: this.openWorkerToolbox.bind(this),
+        openElementInInspector: async function(grip) {
+          await this.toolbox.initInspector();
+          const onSelectInspector = this.toolbox.selectTool("inspector");
+          const onGripNodeToFront = this.toolbox.highlighterUtils.gripToNodeFront(grip);
+          const [
+            front,
+            inspector,
+          ] = await Promise.all([onGripNodeToFront, onSelectInspector]);
+
+          const onInspectorUpdated = inspector.once("inspector-updated");
+          const onNodeFrontSet = this.toolbox.selection
+            .setNodeFront(front, { reason: "debugger" });
+
+          return Promise.all([onNodeFrontSet, onInspectorUpdated]);
+        }.bind(this)
       }
     });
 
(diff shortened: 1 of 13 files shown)


packages/devtools-reps/src/reps/reps.css
```diff
@@ -243,7 +243,7 @@
 
 button.open-accessibility-inspector,
 button.open-inspector {
-  mask: url(/images/open-inspector.svg) no-repeat;
+  mask: url("chrome://devtools/skin/images/vview-open-inspector.png") no-repeat;
```
darkwing: This is required to show the "open in inspector" icon; even the console appears broken in Nightly.

nchevobbe: How's that? I just tested on latest Nightly and it seems fine in the console: 

image

Would you have the STR for the broken console?

darkwing: My bad; the `./firefox` directory hash was behind and it wasn't showing the icon.  I updated and we're good.  Sorry!

darkwing:
@captainbrosset Yay, inspector and (new) debugger will be friends again!

assets/panel/panel.js
```diff
@@ -35,7 +35,22 @@ DebuggerPanel.prototype = {
...
+            front,
+            inspector,
+          ] = await Promise.all([onGripNodeToFront, onSelectInspector]);
+
+          const onInspectorUpdated = inspector.once("inspector-updated");
+          const onNodeFrontSet = this.toolbox.selection
+            .setNodeFront(front, { reason: "debugger" });
+
+          return Promise.all([onNodeFrontSet, onInspectorUpdated]);
+        }.bind(this)
```
jasonLaster: lets move this to a method in Panel or Toolbox. My preference would be if it were in the Toolbox class because it is not really debugger specific

darkwing: @nchevobbe @captainbrosset Could you have a quick look at this function before I move it to https://searchfox.org/mozilla-central/source/devtools/client/framework/toolbox.js?

captainbrosset: The code in this function looks good to me. I think this is the right thing to do.
When you do move the code to toolbox.js, can you please ask Alex or Yulia to review the patch? I just want to make sure they take a look at it because they've been changing related code with Fission.

PR #7463, opened 2018-12-08
Updated the "skip pausing" button to "Deactivate breakpoints".

### LINKED ISSUE #7360: [breakpoints] update the skip pausing button copy
Currently hovering on the button shows this copy "Skip all pausing". That is kind of a lie because in reality you'll just skip pausing due to breakpoints. Chrome's copy is "Deactivate breakpoints". I think that is a better option.

image

acnair:
Fixes #7360 

Here's the Pull Request Doc

### Summary of Changes

*  Replaced "Skip all pausing" with "Deactivate breakpoints" in assets/panel/debugger.properties

### Test Plan

* Ran ```yarn start``` and made sure it was changed to "Deactivate breakpoints". 

Feel free to improve it!

hover

### DIFF
--- a/assets/panel/debugger.properties
+++ b/assets/panel/debugger.properties
@@ -84,7 +84,7 @@ stepOutTooltip=Step out %S
 
 # LOCALIZATION NOTE (skipPausingTooltip): The tooltip text for disabling all
 # breakpoints and pausing triggers
-skipPausingTooltip=Skip all pausing
+skipPausingTooltip=Deactivate breakpoints
 
 # LOCALIZATION NOTE (pauseButtonItem): The label that is displayed for the dropdown pause
 # list item when the debugger is in a running state.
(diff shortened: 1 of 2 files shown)


acnair:
@darkwing @flodolo I read the documentation that @flodolo sent me which said that all the instances have to be changed. So I did a rough search and there were close to 50 places where ``skipPausing`` is used! There are even files and import statements which has ``skipPausing`` so does all the ``skipPausing`` needs to be changed?

acnair:
@darkwing but then how did it work when I ran the debugger? Just curious. 
@flodolo oh! I will look into it. :smile:

darkwing:
My concern here is that we're introducing a new term, "Deactiviate", which could be confusing. 

@jasonLaster:  What about "Skip all debugger pauses"?

jasonLaster:
@darkwing two thoughts:

1. the button is specific to breakpoints. we'll still pause in other circumstances like debugger statements, interrupts, and exceptions
2. deactivate is the term that chrome uses, so consistency might help us

digitarald:
👍 for consistency.

Issue #7113, opened 2018-10-18
bogus sourcemap leads to being unable to prettify the file


karlcow:
Original file 

![capture d ecran 2018-10-18 a 14 54 24](https://user-images.githubusercontent.com/505230/47133952-ecfac000-d2e5-11e8-973d-8f34f4a0ae0d.png)

Formatted file. It failed with the message:

> Error while fetching an original source: Error: "https://images.medicinenet.com/oocommon/js/20181004/legacy.js:formatted" is not in the SourceMap.
> Source URL: https://images.medicinenet.com/oocommon/js/20181004/legacy.js:formatted

![capture d ecran 2018-10-18 a 14 56 32](https://user-images.githubusercontent.com/505230/47133977-06037100-d2e6-11e8-9b1f-fda955df1767.png)

which is not extremely practical in this case :)

![capture d ecran 2018-10-18 a 14 58 29](https://user-images.githubusercontent.com/505230/47134056-511d8400-d2e6-11e8-9ee6-176bc0f6f7d2.png)

Site URL is https://www.medicinenet.com/acne/article.htm
Context https://github.com/webcompat/web-bugs/issues/19784

karlcow:
This seems to depends on 

```js
                        n[i].parentNode.insertBefore(d, n[i]), c.default.nextTick(function() {
                            Stickyfill.add(document.querySelectorAll("#apPgNum"))
                        }), _.forEach(["DOMContentLoaded", "load", "scroll", "resize", "popstate", "hashchange"], function(t) {
                            document.getElementById("content").addEventListener(t, _.framerateThrottle(function(t) {
                                if ((0, v.elementInView)(s, 300, 300)) {
                                    if (!o) {
                                        if (p) webmd.ads2.refresh({
                                            id: u,
                                            keys: {
                                                pg: e,
                                                LL: 1
                                            }
                                        });
                                        else {
                                            var n = {
                                                targets: (0, a.default)({}, window.ooAdTarget, {
                                                    pvid: window.s_pageview_id,
                                                    pg: e,
                                                    LL: 1
                                                }),
                                                id: u,
                                                pos: "2026",
                                                sizes: [
                                                    [300, 250],
                                                    [300, 253]
                                                ],
                                                blockCodes: window.ooAdTarget.blockCodes
                                            };
                                            webmd.ads2Consumer.defineAd(n), setTimeout(function() {
                                                l = n.slot
                                            }, 0), p = !0
                                        }
[...]

PR #7466, opened 2018-12-10
Use same logic for pretty print option in footer and tab

darkwing:
References;  https://github.com/devtools-html/debugger.html/issues/7113

We're using different logic for showing the pretty print option in the tab context menu and the footer.  This PR ensures the same logic is used in both places.

### DIFF
--- a/src/utils/editor/index.js
+++ b/src/utils/editor/index.js
@@ -58,22 +58,18 @@ export function endOperation() {
   codeMirror.endOperation();
 }
 
-export function shouldShowPrettyPrint(selectedSource) {
-  if (!selectedSource) {
-    return false;
-  }
-
-  return shouldPrettyPrint(selectedSource);
+export function shouldShowPrettyPrint(source) {
+  return shouldPrettyPrint(source);
 }
 
-export function shouldShowFooter(selectedSource, horizontal) {
+export function shouldShowFooter(source, horizontal) {
   if (!horizontal) {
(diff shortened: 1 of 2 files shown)


darkwing:
WIP'ing because I'm seeing the following:

1uncaught exception: Error: Line numbers must be >= 1`

src/utils/editor/index.js
```diff
@@ -58,22 +58,18 @@ export function endOperation() {
...
 }
 
-export function shouldShowPrettyPrint(selectedSource) {
-  if (!selectedSource) {
-    return false;
-  }
-
-  return shouldPrettyPrint(selectedSource);
+export function shouldShowPrettyPrint(source) {
+  return shouldPrettyPrint(source);
```
AnshulMalik: Now that we are just calling `shouldPrettyPrint`, the additional function, `shouldShowPrettyPrint`. Can't we just use `shouldPrettyPrint`?

darkwing: I kind of like having a function with a presentational name wrapping this.  To me "shouldPrettyPrint" sounds like it determines if an action should take place; "shouldShowPrettyPrint" seems like a good symbol of if something should be presented.

AnshulMalik: Ahan, we can do something like `shouldShowPrettyPrint = shouldPrettyPrint`, that would mean on less call on stack :P

PR #7467, opened 2018-12-10
Clear the watch expression input box on blur

### LINKED ISSUE #7464: "Invalid Expression" text and style should go away when blurred
When we blur from the expressions element, we should clear its state so that the error styling and message does not persist.

invalidexpression

notjaril:
Fixes #7464

### Summary of Changes

Fixed expression-input-container so that it clears the input box on blur.

dec-09-2018 16-49-58

### DIFF
--- a/src/components/SecondaryPanes/Expressions.js
+++ b/src/components/SecondaryPanes/Expressions.js
@@ -26,6 +26,8 @@ import type { Expression } from "../../types";
 
 import "./Expressions.css";
 
+import * as parser from "../../workers/parser";
+
 const { ObjectInspector } = objectInspector;
 
 type State = {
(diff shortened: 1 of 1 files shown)


notjaril:
I think I now understand what you had in mind, in which case my original commit was over-complicated. Some thoughts:

```
+            onBlur={() => {
+              this.hideInput();
+              if (!inputValue) {
+                this.props.clearExpressionError();
+              }
+            }}
```

This is good - was unclear about what you had intended to fix with the original issue, but this clears it up. That said, how about cases where the user modified the invalid input, but did not empty it? Upon blurring, this would keep the modified input and a red border around it. Options might be:

1) Keep error state (default)
2) Clear error state regardless of input value on blur
3) Validate input value on blur, only clear the error state if the current input value is no longer invalid

I'm leaning towards 2 since it's simpler, and I can't currently think of why anybody would want to be notified of an invalid input value if they're focused on other parts of the debugger. Also, it's less urgent to remind users that their last entry did not go through, as they're bound to notice it right away seeing as they're trying to watch those variables.

```
+    if (!value) {
+      this.props.clearExpressionError();
+    }
```

Don't like this as much - while the box is in focus, I think it's better the error state persists to prompt the user to fix the previous invalid input. That way it's clear that whichever expression they tried to add was invalid, and it prompts the user to either fix their input or add something else entirely. If the error state were cleared while the input stays on focus, that (useful) information/prompt is lost.

Let me know what you end up deciding is best, and I can make the changes.

PR #7468, opened 2018-12-10
Add separator between search box/navigation and modifiers

darkwing:
This was brought up by @violasong during our meeting last week.


darkwing:
Looks like:

sep

PR #7471, opened 2018-12-10
Provide title text for line and column position

darkwing:
I can see how the line and column text could be confusing or meaningless to some devs.  This tooltip lets them know what the numbers mean.

### DIFF
--- a/assets/panel/debugger.properties
+++ b/assets/panel/debugger.properties
@@ -601,11 +601,11 @@ sourceFooter.codeCoverage=Code coverage
 
 # LOCALIZATION NOTE (sourceFooter.currentCursorPosition): Text associated
 # with the current cursor line and column
-sourceFooter.currentCursorPosition=(%S, %S)
+sourceFooter.currentCursorPosition.label=(%1$S, %2$S)
 
 # LOCALIZATION NOTE (sourceFooter.currentCursorPosition.tooltip): Text associated
 # with the current cursor line and column
-sourceFooter.currentCursorPosition.tooltip=(Line %S, column %S)
+sourceFooter.currentCursorPosition.tooltip=(Line %1$S, column %2$S)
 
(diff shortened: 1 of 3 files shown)


assets/panel/debugger.properties
```diff
@@ -603,6 +603,10 @@ sourceFooter.codeCoverage=Code coverage
 # with the current cursor line and column
 sourceFooter.currentCursorPosition=(%S, %S)
 
+# LOCALIZATION NOTE (sourceFooter.currentCursorPosition.tooltip): Text associated
+# with the current cursor line and column
+sourceFooter.currentCursorPosition.tooltip=(Line %S, column %S)
```
flodolo: One thing that I missed in the line above: we should try to explicitly use ordered parameters when there's more than one.

```
 sourceFooter.currentCursorPosition.tooltip=(Line %1$S, column %2$S) 
```

AnshulMalik: And should the first letter `c` in `column` be capitalized?

assets/panel/debugger.properties
```diff
@@ -601,11 +601,11 @@ sourceFooter.codeCoverage=Code coverage
 
 # LOCALIZATION NOTE (sourceFooter.currentCursorPosition): Text associated
 # with the current cursor line and column
-sourceFooter.currentCursorPosition=(%S, %S)
+sourceFooter.currentCursorPosition.label=(%1$S, %2$S)
```
flodolo: No need to update the ID for this one, it's a safe change (localizers can still change it, they're equivalent)

assets/panel/debugger.properties
```diff
@@ -601,7 +601,11 @@ sourceFooter.codeCoverage=Code coverage
...
```
flodolo: No need to update the ID here, localizers can already change it if they want to.

PR #7475, opened 2018-12-10
Fix 6121 - Crop attributes and provide title after specified length

### LINKED ISSUE #6121: ElementNode attribute are not cropped
_From @nchevobbe on December 7, 2017 9:4_

**Steps to reproduce:**
1. Evaluate:
```js
var div = document.createElement("div");
div.setAttribute("data-test","a".repeat(5000));
div;
```

**Expected results:**
The `data-test attribute` is cropped 

**Actual results:**
It is not cropped (and does not wrap)

_Copied from original issue: devtools-html/devtools-core#835_

darkwing:
Fixes #6121

This PR crops long attributes and provides a `title` attribute for the attribute itself.

## To Do
- [x] Test
- [x] Despite having a `title` attribute, I don't see the tooltip every display when I hover or move around

### DIFF
--- a/packages/devtools-reps/src/reps/element-node.js
+++ b/packages/devtools-reps/src/reps/element-node.js
@@ -14,6 +14,8 @@ const nodeConstants = require("../shared/dom-node-constants");
 const dom = require("react-dom-factories");
 const { span } = dom;
 
+const maxAttributeLength = 50;
+
 /**
  * Renders DOM element node.
  */
@@ -144,7 +146,12 @@ function getElements(grip, mode) {
       {},
       span({ className: "attrName" }, name),
       span({ className: "attrEqual" }, "="),
-      StringRep({ className: "attrValue", object: value })
+      StringRep({
+        className: "attrValue",
+        object: value,
+        cropLimit: maxAttributeLength,
+        title: value.length > maxAttributeLength ? value : null
+      })
     );
 
     return arr.concat([" ", attribute]);
--- a/packages/devtools-reps/src/reps/element-node.js
+++ b/packages/devtools-reps/src/reps/element-node.js
@@ -7,14 +7,14 @@ const PropTypes = require("prop-types");
 
 // Utils
 const { isGrip, wrapRender } = require("./rep-utils");
-const { rep: StringRep } = require("./string");
+const { rep: StringRep, isLongString } = require("./string");
 const { MODE } = require("./constants");
 const nodeConstants = require("../shared/dom-node-constants");
 
 const dom = require("react-dom-factories");
 const { span } = dom;
 
-const maxAttributeLength = 50;
+const MAX_ATTRIBUTE_LENGTH = 50;
 
 /**
  * Renders DOM element node.
@@ -142,6 +142,12 @@ function getElements(grip, mode) {
   }
   const attributeElements = attributeKeys.reduce((arr, name, i, keys) => {
     const value = attributes[name];
+
+    let title = isLongString(value) ? value.initial : value;
+    if (title.length < MAX_ATTRIBUTE_LENGTH) {
+      title = null;
+    }
+
     const attribute = span(
       {},
       span({ className: "attrName" }, name),
@@ -149,8 +155,8 @@ function getElements(grip, mode) {
       StringRep({
         className: "attrValue",
         object: value,
-        cropLimit: maxAttributeLength,
-        title: value.length > maxAttributeLength ? value : null
+        cropLimit: MAX_ATTRIBUTE_LENGTH,
+        title
       })
     );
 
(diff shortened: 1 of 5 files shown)


darkwing:
@nchevobbe This work in progress is sufficient enough to look at now, as I get familiar with the codebabse

packages/devtools-reps/src/reps/string.js
```diff
@@ -32,7 +32,8 @@ StringRep.propTypes = {
   member: PropTypes.object,
   object: PropTypes.object.isRequired,
   openLink: PropTypes.func,
-  className: PropTypes.string
+  className: PropTypes.string,
+  title: PropTypes.title
```
nchevobbe: should it be `PropTypes.string` ?

nchevobbe:
The patch looks good to me, and it does work well.
Regarding the `title, it is added to the DOM, but I can't have a tooltip to be displayed, that's weird.

nchevobbe:
The `title` issue seems to be a Firefox bug. I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1513170.

You can check with `var div = document.createElement("div"); div.setAttribute("data-test","a ".repeat(2000)); div;` that the tooltip is rendered fine when there's whitespaces (which I supposed there will be in most cases).

This lead to another issue with the current code with the attribute is a long string.
if you evaluate: 
```js
var div = document.createElement("div"); 
div.setAttribute("data-test","a ".repeat(5000)); 
div;
```
the title is `[object Object]`

This is because the string is now a longString (> 9999 chars), which does not contain the full text. I think we can simply display the `initial` property we get in those case.

packages/devtools-reps/src/reps/stubs/element-node.js
```diff
@@ -206,6 +206,22 @@ stubs.set("SvgNodeInXHTML", {
...
+  type: "object",
+  actor: "server1.conn1.child1/obj32",
+  class: "HTMLParagraphElement",
+  ownPropertyLength: 0,
+  preview: {
+    kind: "DOMNode",
+    nodeType: 1,
+    nodeName: "p",
+    attributes: {
+      "data-test": "a".repeat(100)
```
nchevobbe: I don't think this qualifies as a `LongString`, as we acknowledge it in DevTools.
Here's a stub representing a longString: https://github.com/devtools-html/debugger.html/blob/30b0065dceb3933e116df1a15635c4bacf79f4b2/packages/devtools-reps/src/reps/stubs/long-string.js#L20-L27

I guess we could re-use it here for `"data-test"`

Here's the result we get for `var div = document.createElement("div"); div.setAttribute("data-test","a ".repeat(5000)); div;` (which creates a long string attribute)

```js
{
  "type": "object",
  "actor": "server1.conn0.child1/obj26",
  "class": "HTMLDivElement",
  "extensible": true,
  "frozen": false,
  "sealed": false,
  "ownPropertyLength": 0,
  "preview": {
    "kind": "DOMNode",
    "nodeType": 1,
    "nodeName": "div",
    "isConnected": false,
    "attributes": {
      "data-test": {
        "type": "longString",
        "initial": "a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a ",
        "length": 10000,
        "actor": "server1.conn0.child1/longString27"
      }
    },
    "attributesLength": 1
  }
}
```

darkwing: I think this is a naming issue; the usage of this stub is simply to have an attribute > max (50); I used the `NodeWithLongStringAttribute` to test a true LongString

packages/devtools-reps/src/reps/element-node.js
```diff
@@ -170,6 +183,7 @@ function supportsObject(object, noGrip = false) {
 
 // Exports from this module
 module.exports = {
-  rep: wrapRender(ElementNode),
-  supportsObject
+  rep: wrapRender(Element, Node),
```
nchevobbe:
```suggestion
  rep: wrapRender(ElementNode),
```

PR #7476, opened 2018-12-10
Sync 106

jasonLaster:
a bunch of stuff landed post 106...


src/components/Editor/Editor.css
```diff
@@ -240,7 +240,7 @@ debug-expression-error {
   display: none;
 }
 
-.highlight-line .CodeMirror-line {
+.debugger:not(.can-rewind) .highlight-line .CodeMirror-line {
```
AnshulMalik: what's this for?

jasonLaster: It was a quick fix so that WebReplay demos would be smoother

AnshulMalik: can we do this for dark theme too?

PR #7481, opened 2018-12-11
Set min-width for left toggle to ensure line marker and toggle button border line up

darkwing:
Both @violasong and I noticed each time that the first tab border and the line number don't line up.  This PR makes them line up initially:

align

When the user scrolls down to the 100+ lines, the CodeMirror vertical line number separator moves right and from that point forward the alignment is broken from the first tab.

I do believe, however, that having this aesthetically pleasing alignment upon initial open is worth this patch.

### DIFF
--- a/src/components/shared/Button/styles/PaneToggleButton.css
+++ b/src/components/shared/Button/styles/PaneToggleButton.css
@@ -28,6 +28,7 @@
 
 .toggle-button.start {
   margin-inline-start: 0px;
+  min-width: 29px;
 }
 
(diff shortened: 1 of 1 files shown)


jasonLaster:
cc @jainsneha23

violasong:
Yeah agreed with the closing of this issue for now - it seems too tricky to solve for how minor it is. I looked at other text editors and their approach is to not even have a separator line along the gutter. Not sure if we want to go that minimal.  Anyway Chrome has the same mis-alignment ;)

PR #7482, opened 2018-12-11
7190: Save generated file and source map

### LINKED ISSUE #7190: [sourcemap] save file and map
One of the challenging aspects of debugging sourcemaps today is that it can 

a) be hard to downlaod the generated file and sourcemap.
b) be hard to visualize the map because source map visualizer has a cap on the size of the visualization

we can address part a by having an editor context menu item for, which saves the generated file and source map to a file

Here is a gist for creating a generic *save file prompt*

```js
   var saveData = (function () {
    var a = document.createElement("a");
    document.body.appendChild(a);
    a.style = "display: none";
    return function (data, fileName) {
        var json = JSON.stringify(data),
            blob = new Blob([json], {type: "text/json"}),
            url = window.URL.createObjectURL(blob);
        a.href = url;
          a.download = fileName;
          a.click();
          window.URL.revokeObjectURL(url);
      };
  }());

  var data = { x: 42, s: "hello, world", d: new Date() },
      fileName = "my-download.json";

  saveData(data, fileName);
```

jainsneha23:
Fixes #7190 

### Summary of Changes

* Created a utility to download data with a given filename
* Changed debugger properties to introduce text for context menu
* Change in editor menu to introduce an item to save generated file

### Test Plan

- [x] Go to any generated file
- [x] Right click to see context menu
- [x] There's an item saying "Download Generated File". Click it
- [x] The file is downloaded with filename of the generated file


jasonLaster:
Yep, i think download file is cleaner. We can tack on other features if it is a generated file...

assets/panel/debugger.properties
```diff
@@ -501,6 +501,11 @@ editor.conditionalPanel.close=Cancel edit breakpoint and close
 editor.jumpToMappedLocation1=Jump to %S location
 editor.jumpToMappedLocation1.accesskey=m
 
+# LOCALIZATION NOTE (editor.downloadGeneratedFile): Context menu item
```
flodolo: Wrong reference to the string (it doesn't have `editor.`).

assets/panel/debugger.properties
```diff
@@ -501,6 +501,11 @@ editor.conditionalPanel.close=Cancel edit breakpoint and close
...
+# for downloading generated file and its source map
+downloadGeneratedFile=Download Generated File
```
flodolo: Can we use `downloadGeneratedFile.label`?

src/components/Editor/EditorMenu.js
```diff
@@ -197,6 +213,10 @@ function getMenuItems(
     menuItems.push(watchExpressionItem, evaluateInConsoleItem);
   }
 
+  if (!isOriginal) {
```
jasonLaster: maybe we should just have a download file item

yarn.lock
```diff
@@ -852,10 +835,6 @@ assert@^1.1.1:
   dependencies:
     util "0.10.3"
 
-assertion-error@^1.0.1:
-  version "1.1.0"
-  resolved "https://registry.yarnpkg.com/assertion-error/-/assertion-error-1.1.0.tgz#e60b6b0e8f301bd97e5375215bda406c85118c0b"
-
```
jasonLaster: the yarn.lock changes look erroneous

assets/panel/debugger.properties
```diff
@@ -510,6 +510,11 @@ editor.conditionalPanel.close=Cancel edit breakpoint and close
 editor.jumpToMappedLocation1=Jump to %S location
 editor.jumpToMappedLocation1.accesskey=m
 
+# LOCALIZATION NOTE (downloadFile.label): Context menu item
+# for downloading a source's content
+downloadFile.label=Download file
+downloadFile.accesskey=d
```
flodolo: Comment for posterity: access keys are case sensitive. By using lowercase, the last character will be underlined, instead of the first one. Not sure if it's the desired behavior.

If there's no perfect match (same letter and case), it would fall back to the other case, so (`Down` and `d` would underline the first character as fall back).

Issue #7223, opened 2018-11-02
Stop converting spread syntax into Object.assign


Loirooriol:
https://github.com/devtools-html/debugger.html/blob/e7f212ee3916d97c799549b6b89cb43fbc1ce34f/packages/devtools-reps/src/reps/grip.js#L169

```js
properties = { ...properties, ...object.preview.safeGetterValues };
```

uses spread syntax because of https://github.com/devtools-html/devtools-core/issues/986

However, somehow this is transformed into https://searchfox.org/mozilla-central/rev/50ba1dd30cf013bddce1ae756f1b3c95b26f0628/devtools/client/shared/components/reps/reps.js#1429

```js
properties = _extends({}, properties, object.preview.safeGetterValues);
```

where `_extends === Object.assign`.

This causes https://bugzilla.mozilla.org/show_bug.cgi?id=1496044

Regressed by https://bugzilla.mozilla.org/show_bug.cgi?id=1457517

https://github.com/devtools-html/debugger.html/compare/release-45...release-46, probably https://github.com/devtools-html/debugger.html/commit/03010e173028716de3c0ddb04e61487c3de05b98

jasonLaster:
Thanks @Loirooriol I'll fix this next week.

pradeepgangwar:
@jasonLaster @Loirooriol I don't think this needs changes in this repo. Please let me know if i am missing something.

Loirooriol:
@pradeepgangwar This didn't happen when the code was in devtools-core. I guess babel is misconfigured here and breaks the code.

jasonLaster:
I think this was done in m-c. Thanks

Loirooriol:
@jasonLaster This is still broken, please reopen.

nchevobbe:
i'm looking into this

PR #7484, opened 2018-12-11
Invoke getter in the root object.

### LINKED ISSUE #7397: Consider evaluating getters in the root object
It's usual to define getters in a function prototype so that they will be inherited by the instances. Usually it makes no sense to call the getter in the prototype itself.

For example, consider:

```js
new class {
  constructor() {
    this.num = Math.random();
  }
  get double() {
    return this.num * 2;
  }
}();
```

You can inspect the `<prototype>` and run the `double` getter but you will only get `NaN` because the prototype has no `num` property.

I think it would be more useful to evaluate the getter in the root object instead of in the owner one.

Alternatively, `double` could appear in both objects, so that the user can decide in which one it should be evaluated. The downside is that there could be lots of inherited getters and the user might lonly be interested in the own properties, then this would be a nuisance. So maybe group them in a `<inherited getters>`, similar to `<default properties>`.

nchevobbe:
To invoke getter, we were retrieving the parent node
so we could access the getter. This works fine unless
the getter is itself somewhere in the prototype chain
and refers to a property set by an upper level.
This patch changes how we retrieve the context to
invoke the getter in to be the top-most, non-prototype
grip, which should cover this case.

Fixes #7397 

---

Still WIP as I'm trying to find ways I could break it :) 
@Loirooriol , would you have complex test cases I can test against to ensure this is a solid fix? :)

### DIFF
--- a/packages/devtools-reps/src/object-inspector/utils/node.js
+++ b/packages/devtools-reps/src/object-inspector/utils/node.js
@@ -895,22 +895,6 @@ function getNonPrototypeParentGripValue(item: Node | null): Node | null {
   return getValue(parentGripNode);
 }
 
-function getParentGripValue(
-  item: Node
-): RdpGrip | ObjectInspectorItemContentsValue | null {
-  const parentNode = getParent(item);
-  if (!parentNode) {
-    return null;
-  }
-
-  const parentGripNode = getClosestGripNode(parentNode);
-  if (!parentGripNode) {
-    return null;
-  }
-
-  return getValue(parentGripNode);
-}
-
 module.exports = {
   createNode,
   createGetterNode,
@@ -921,7 +905,6 @@ module.exports = {
   getClosestGripNode,
   getClosestNonBucketNode,
   getParent,
-  getParentGripValue,
   getNonPrototypeParentGripValue,
   getNumericalPropertiesCount,
   getValue,
--- a/packages/devtools-reps/src/object-inspector/utils/node.js
+++ b/packages/devtools-reps/src/object-inspector/utils/node.js
@@ -877,6 +877,24 @@ function getClosestNonBucketNode(item: Node): Node | null {
   return getClosestNonBucketNode(parent);
 }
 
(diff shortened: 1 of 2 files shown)


Loirooriol:
For example,

```js
({
  value: 1,
  obj: {
    value: 2,
    ["__proto__"]: {
      value: 3,
      __proto__: {
        value: 4,
        __proto__: {
          value: 5,
          get getter() { return this.value; },
        },
      },
    },
  },
})
```

I should be able to expand `obj`, `__proto__`, `<prototype>`, `<prototype>`, run the getter and get `3`.

But of course `__proto__` is still broken (#7223).

nchevobbe:
> I should be able to expand obj, __proto__, <prototype>, <prototype>, run the getter and get 3.

Good, I applied patch from #7489 on top of this patch and tested your test case: 
image

I think this is ready to be reviewed. 

I also added a mochitest in https://bugzilla.mozilla.org/show_bug.cgi?id=1513505 ([TRY](https://treeherder.mozilla.org/#/jobs?repo=try&revision=8e8c8d380afad78f9d027ad2f470f911e25bad48))

packages/devtools-reps/src/object-inspector/utils/node.js
```diff
@@ -877,6 +877,24 @@ function getClosestNonBucketNode(item: Node): Node | null {
...
+  }
+
+  if (getType(parentGripNode) === NODE_TYPES.PROTOTYPE) {
+    return getNonPrototypeParentGripValue(parentGripNode);
+  }
+
+  return getValue(parentGripNode);
+}
+
 function getParentGripValue(
```
juliandescottes: I don't think getParentGripValue is used anywhere anymore?

nchevobbe: good catch. Let me remove it :)

nchevobbe:
how does that look Jason? (last patch before refactor :) I want to uplift this to beta if possible)

PR #7486, opened 2018-12-11
(wip) copy commits

jasonLaster:
this will let us land individual commits in a git cinnabar setup. We can add hg functionality to it fairly easily..


darkwing:
omg

jasonLaster:
this is not perfect, but good enough

PR #7495, opened 2018-12-12
Ensure found breakpoint isn't disabled

darkwing:
Simply because a column breakpoint is found doesn't mean it's enabled...

### DIFF
--- a/src/selectors/visibleColumnBreakpoints.js
+++ b/src/selectors/visibleColumnBreakpoints.js
@@ -112,7 +112,7 @@ export function getColumnBreakpoints(pausePoints, breakpoints, viewport) {
 
     return {
       location,
-      enabled: foundBreakpoint ? !foundBreakpoint.disabled : false,
+      enabled: !!foundBreakpoint && !foundBreakpoint.disabled,
       condition: foundBreakpoint ? foundBreakpoint.condition : null
     };
   });
--- a/src/selectors/visibleColumnBreakpoints.js
+++ b/src/selectors/visibleColumnBreakpoints.js
@@ -112,7 +112,7 @@ export function getColumnBreakpoints(pausePoints, breakpoints, viewport) {
 
     return {
       location,
-      enabled: !!foundBreakpoint,
+      enabled: foundBreakpoint && !foundBreakpoint.disabled,
       condition: foundBreakpoint ? foundBreakpoint.condition : null
     };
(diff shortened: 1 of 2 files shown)


src/selectors/visibleColumnBreakpoints.js
```diff
@@ -112,7 +112,7 @@ export function getColumnBreakpoints(pausePoints, breakpoints, viewport) {
 
     return {
       location,
-      enabled: !!foundBreakpoint,
+      enabled: foundBreakpoint ? !foundBreakpoint.disabled : false,
```
jasonLaster: would this work?

```js
foundBreakpoint && !foundBreakpoint.disabled
```

jasonLaster: i wonder why disabled would be undefined... maybe we should change that so that it is true or false?

darkwing: That's what `foundBreakpoint ? !foundBreakpoint.disabled : false` does

jasonLaster: so if `foundBreakpoint.disabled` were either true or false, `foundBreakpoint && !foundBreakpoint.disabled` would work...

we try our best to keep disabled a boolean in several places

- https://github.com/bomsy/debugger.html/blob/funcition-indexes/src/utils/breakpoint/index.js#L141-L150
- https://github.com/bomsy/debugger.html/blob/funcition-indexes/src/actions/breakpoints/addBreakpoint.js#L77

somehow it must be allowed to become undefined

darkwing: Try executing the following in console:

```js
let foundBreakpoint
foundBreakpoint && !foundBreakpoint.disabled
```

You'll get `undefined`; it's not a bug in our breakpoint code, it's that `foundBreakpoint && !foundBreakpoint.disabled` returns `undefined` because `foundBreakpoint` is `undefined`; I prevent that issue with my ternary:  `enabled: foundBreakpoint ? !foundBreakpoint.disabled : false`

jasonLaster: got it...  i didn't follow.

Okay, so this is a super nit, but i think i'd prefer `!!foundBreakpoint && !foundBreakpoint.disabled`. Go with whatever looks better to you though.

PR #7496, opened 2018-12-13
[Pause points] speed it up

### LINKED ISSUE #7485: Unresponsive script in pausePoints.js when trying to debug pretty printed site
screen shot 2018-12-11 at 9 23 00 am

This reproduces for me in DevEdition 65b2 and Nightly 66 on OSX.

STR: 

1. Navigate to https://docs.arnoldrenderer.com/display/A5AFMUG/Arnold+for+Maya+User+Guide
2. Open DevTools
3. Click on the image of the 3d model space guy
4. Click on the link from the TypeError message to open the source script in the debugger
5. Click PrettyPrint (and wait...)
6. Go back to console, click on source link again to go to the right place in the debugger
7. Click around the debugger for a bit... and eventually Firefox will be locked up and you'll get a unresponsive script message.

Note: oddly, trying to go through my steps again, Nightly decided to auto-pretty print (which is neat?) but I ended up with the same frozen pausePoints.js:49 message.

jasonLaster:
Fixes #7485

### Summary of Changes

- backports https://phabricator.services.mozilla.com/D14346
- also adds a second performance fix, which is slightly less impactful

### DIFF
--- a/src/actions/ast/setPausePoints.js
+++ b/src/actions/ast/setPausePoints.js
@@ -9,6 +9,7 @@ import * as parser from "../../workers/parser";
 import { isGenerated } from "../../utils/source";
 import { mapPausePoints } from "../../utils/pause/pausePoints";
 import { features } from "../../utils/prefs";
+import { getGeneratedLocation } from "../../utils/source-maps";
 
 import type { SourceId } from "../../types";
 import type { ThunkArgs, Action } from "../types";
@@ -26,12 +27,17 @@ function compressPausePoints(pausePoints) {
   return compressed;
 }
 
-async function mapLocations(pausePoints, source, sourceMaps) {
+async function mapLocations(pausePoints, state, source, sourceMaps) {
   const sourceId = source.id;
   return mapPausePoints(pausePoints, async ({ types, location }) => {
-    const generatedLocation = await sourceMaps.getGeneratedLocation(
-      { ...location, sourceId },
-      source
+    const generatedLocation = await getGeneratedLocation(
+      state,
+      source,
+      {
+        ...location,
+        sourceId
+      },
+      sourceMaps
     );
 
     return { types, location, generatedLocation };
@@ -49,7 +55,12 @@ export function setPausePoints(sourceId: SourceId) {
     }
 
     let pausePoints = await parser.getPausePoints(sourceId);
-    pausePoints = await mapLocations(pausePoints, source, sourceMaps);
+    pausePoints = await mapLocations(
+      pausePoints,
+      getState(),
+      source,
+      sourceMaps
+    );
 
     if (isGenerated(source)) {
       const compressed = compressPausePoints(pausePoints);
--- a/src/utils/pause/pausePoints.js
+++ b/src/utils/pause/pausePoints.js
@@ -47,13 +47,14 @@ export async function mapPausePoints<T>(
   iteratee: PausePoint => T
 ) {
   const results = await Promise.all(convertToList(pausePoints).map(iteratee));
+  let index = 0;
 
   const newPausePoints = {};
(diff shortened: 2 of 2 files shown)


src/actions/ast/setPausePoints.js
```diff
@@ -29,10 +30,10 @@ function compressPausePoints(pausePoints) {
 async function mapLocations(pausePoints, source, sourceMaps) {
   const sourceId = source.id;
   return mapPausePoints(pausePoints, async ({ types, location }) => {
-    const generatedLocation = await sourceMaps.getGeneratedLocation(
-      { ...location, sourceId },
-      source
-    );
+    const generatedLocation = await getGeneratedLocation(sourceMaps, source, {
```
jasonLaster: we want to use this version of the helper because we will avoid making calls to the worker when it is already generated...

src/utils/source-maps.js
```diff
@@ -11,10 +11,9 @@ import type { SourceLocation, MappedLocation, Source } from "../types";
 import { isGenerated } from "../utils/source";
 
 export async function getGeneratedLocation(
-  state: Object,
+  sourceMaps: Object,
```
jasonLaster: I think we should avoid passing state into lots of helpers... 

This is an opinionated refactor to prevent this.

src/actions/ast/setPausePoints.js
```diff
@@ -26,12 +27,17 @@ function compressPausePoints(pausePoints) {
...
 }
 
-async function mapLocations(pausePoints, source, sourceMaps) {
+async function mapLocations(pausePoints, state, source, sourceMaps) {
   const sourceId = source.id;
   return mapPausePoints(pausePoints, async ({ types, location }) => {
-    const generatedLocation = await sourceMaps.getGeneratedLocation(
-      { ...location, sourceId },
-      source
+    const generatedLocation = await getGeneratedLocation(
```
jasonLaster: this prevents unnecessary calls to sourceMaps for generated sources.

src/utils/pause/pausePoints.js
```diff
@@ -47,13 +47,14 @@ export async function mapPausePoints<T>(
...
   const results = await Promise.all(convertToList(pausePoints).map(iteratee));
+  let index = results.length - 1;
 
   const newPausePoints = {};
   for (const line in pausePoints) {
     const linePoints = pausePoints[line];
     const newLinePoints = (newPausePoints[line] = {});
     for (const column in linePoints) {
-      newLinePoints[column] = results.shift();
+      newLinePoints[column] = results[index--];
```
jasonLaster: shift is slow

jasonLaster: I think you're right. :/

jasonLaster:
Thanks for the catch @nchevobbe. Fortunately we're only using the mappings logic for column breakpoints.

jasonLaster:
We can definitely cut down on convertToList, which is slow in a follow up PR.

nchevobbe:
Could we write a test for it? Circle and Travis were perfectly fine with
the patch, which is a bit scary

Le jeu. 13 déc. 2018 à 21:03, Jason Laster <notifications@github.com> a
écrit :

jasonLaster:
@nchevobbe i'd prefer to have a mochitest for column breakpoints that would test it end-to-end. I believe that @darkwing will write some as we get closer to the end point. 

Also, when we remove the convert to list use cases, mapLocations will go away too.

jasonLaster:
alright, here's a follow up item which simplifies the data structure so we just use a list. I think it is complicated enough where we should land this and then do a separate review

https://github.com/devtools-html/debugger.html/pull/7504

PR #7506, opened 2018-12-14
Stop using immutable for search results

AnshulMalik:


src/reducers/project-text-search.js
```diff
@@ -38,69 +37,68 @@ export type ProjectTextSearchState = {
...
-  ProjectTextSearchState
-> {
-  return makeRecord(
-    ({
-      query: "",
-      results: I.List(),
-      status: statusType.initial
-    }: ProjectTextSearchState)
-  )();
+export function initialProjectTextSearchState(): ProjectTextSearchState {
```
jasonLaster: can ProjectTextSearchState be read only `+query` and things like that?

Can we drop `I`. I see it here ` matches: I.List<any>`

PR #7511, opened 2018-12-14
[symbols] stop fetching symbols when loading

jasonLaster:
### Summary of Changes

We were previously fetching symbols again when we were loading because hasSymbols ignores loading.

### DIFF
--- a/src/actions/ast.js
+++ b/src/actions/ast.js
@@ -56,15 +56,17 @@ export function setSymbols(sourceId: SourceId) {
   return async ({ dispatch, getState, sourceMaps }: ThunkArgs) => {
     const source = getSourceFromId(getState(), sourceId);
 
-    if (source.isWasm || !isLoaded(source) || getSymbols(getState(), source)) {
+    if (source.isWasm && !isLoaded(source)) {
       return;
     }
 
-    await dispatch({
-      type: "SET_SYMBOLS",
-      sourceId,
-      [PROMISE]: parser.getSymbols(sourceId)
-    });
+    if (!getSymbols(getState(), source)) {
+      await dispatch({
+        type: "SET_SYMBOLS",
+        sourceId,
+        [PROMISE]: parser.getSymbols(sourceId)
+      });
+    }
 
     if (isPaused(getState())) {
       await dispatch(fetchExtra());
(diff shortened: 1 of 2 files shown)


src/actions/ast.js
```diff
@@ -92,7 +87,7 @@ export function setOutOfScopeLocations() {
 
     let locations = null;
     if (location.line && source && !source.isWasm && isPaused(getState())) {
-      locations = await findOutOfScopeLocations(
+      locations = await parser.findOutOfScopeLocations(
```
jasonLaster: adding some more specificity so that it is clear when we use a selector vs worker

src/actions/ast.js
```diff
@@ -61,14 +56,14 @@ export function setSymbols(sourceId: SourceId) {
   return async ({ dispatch, getState, sourceMaps }: ThunkArgs) => {
     const source = getSourceFromId(getState(), sourceId);
 
-    if (source.isWasm || hasSymbols(getState(), source) || !isLoaded(source)) {
+    if (source.isWasm || !isLoaded(source) || getSymbols(getState(), source)) {
```
jasonLaster: `hasSymbols` was false when loading

bomsy:
hmm ... these tests shouldn't fail with these changes

src/actions/ast.js
```diff
@@ -61,15 +56,17 @@ export function setSymbols(sourceId: SourceId) {
...
-      sourceId,
-      [PROMISE]: getSymbols(sourceId)
-    });
+    if (!getSymbols(getState(), source)) {
+      await dispatch({
+        type: "SET_SYMBOLS",
+        sourceId,
+        [PROMISE]: parser.getSymbols(sourceId)
+      });
+    }
```
bomsy: This fixes the  failing tests, since we still want o fetch the extra content even if we don't reset the symbols.

jasonLaster:
@bomsy i reverted some of your changes to see if it would pass after a rebase. I thought this should be simple :/

bomsy:
@jasonLaster that's a good idea. didn't try that first.

It seems though like the `loading` check in `hasSymbols` was there to fix this `extra` property issue in the failing test, which is weird and a bit confusing. Might need a refactor. Would love to pair when you have a chance.

PR #7512, opened 2018-12-14
[symbols] save object property identifiers

jasonLaster:
### Summary of Changes

Fixes a pretty big regression introduced by  #7270, which prevented us from previewing a bunch of identifiers in objects and arrays.

### DIFF
--- a/src/workers/parser/getSymbols.js
+++ b/src/workers/parser/getSymbols.js
@@ -16,7 +16,9 @@ import {
   getPatternIdentifiers,
   getComments,
   getSpecifiers,
-  getCode
+  getCode,
+  nodeHasSameLocation,
+  getFunctionParameterNames
 } from "./utils/helpers";
 
 import { inferClassName } from "./utils/inferClassName";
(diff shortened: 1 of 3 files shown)


src/workers/parser/getSymbols.js
```diff
@@ -122,6 +123,12 @@ function getFunctionParameterNames(path: SimplePath): string[] {
   return [];
 }
 
+function addIdentifier(symbols, newId) {
+  if (!symbols.identifiers.find(id => nodeHasSameLocation(id, newId))) {
```
jasonLaster: i'm a little concerned by the perf here, but i think it is okay because we don't store that many identifiers...

src/workers/parser/utils/helpers.js
```diff
@@ -50,7 +51,7 @@ export function getObjectExpressionValue(node: Node) {
     return value.name;
   }
 
-  if (t.isCallExpression(value)) {
+  if (t.isCallExpression(value) || t.isFunctionExpression(value)) {
```
jasonLaster: this logic is a bit brittle... the idea is we can evaluate object values, if they are safe...

so for instance if the user hovers over `a` in `{a: b}` we'll evaluate `b`... the risk is that b might have side effects... so  we try and prevent that. I'm filtering by function here because it is not a common case and makes the snapshot look bad :) we generally wouldn't want to evaluate the render function either `{render: () => {}}`

src/workers/parser/getSymbols.js
loganfsmyth: Should we add it and then dedup at the end instead maybe? At the end of the day this is fine though.

darkwing:
@loganfsmyth Is this ready?

jasonLaster:
@darkwing i'd like to do the filtering at the end for performance

PR #7519, opened 2018-12-15
Remove white background for source icon images in tree, also remove border

darkwing:
@fvsch Reported the following:

screen shot 2018-12-15 at 18 31 52

Selected source icons have a white background we don't want.

This PR:
* Removes the white background
* Removes what appears to be a border around the `` icons, which I'm presuming is a windows thing.

white

iconbg

### DIFF
--- a/src/components/PrimaryPanes/Sources.css
+++ b/src/components/PrimaryPanes/Sources.css
@@ -89,10 +89,6 @@
   margin-right: 5px;
 }
 
-.sources-list .tree .focused img.arrow {
-  background-color: white;
-}
-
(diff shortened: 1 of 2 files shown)


fvsch:
@darkwing 

> Removes what appears to be a border around the  icons, which I'm presuming is a windows thing.

It's a Gecko platform thing, since Firefox 55.
Are your screenshots made in Firefox DevTools, or with a standalone debugger?

fvsch:
This is most probably not fixed.
I create a follow-up issue with more detailed information in #7520.

PR #7523, opened 2018-12-17
Implement CSS containment on secondary pane and sources tree

darkwing:
Adding to these spaces based on:  https://wiki.mozilla.org/Platform/Layout/CSS_Containment_Best_Practices

I didn't add to Editor because that's being done in MC by `:dholbert`:  https://bugzilla.mozilla.org/show_bug.cgi?id=1502524

## To Do
- [ ] Figure out a good way to measure perf impact from this.
- [ ] Should I add to the editor's container (?) . Maybe :dholbert's work in MC isn't enough?

### DIFF
--- a/src/components/PrimaryPanes/Sources.css
+++ b/src/components/PrimaryPanes/Sources.css
@@ -204,6 +204,10 @@
   overflow: auto;
 }
 
+.sources-list {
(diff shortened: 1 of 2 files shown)


jasonLaster:
closing for the time being

PR #7526, opened 2018-12-17
Upgrade flow to 0.89

ryanjduffy:
* Adds adapter functions for connect and createSelector to suppress
  flow typing errors
* Corrects additional type errors introduced by new rules


jasonLaster:
squashed and rebased. this should be ready to go soon

jasonLaster:
As follow up items, it would be nice to flow type reselect and ideally update our flow-types so we don't need the adapters.

ryanjduffy:
:tada: Thanks @jasonLaster

PR #7527, opened 2018-12-17
Simplify the Sources Tree  icon highlight color

darkwing:
My last attempt at this wasn't helpful in our current image/border predicament, so we should remove it.  This is much simpler.


darkwing:
To test, bring up a website that uses frameworks and highlight each source node, making sure its icon is white.

PR #7528, opened 2018-12-17
Use <span> instead of <img> without SRC

### LINKED ISSUE #7520: src-less images (<img class="foo">) show up with a border in Gecko
We've looked at this problem in a few issues or PRs I believe, but I don't think we've addressed the core issue properly.

## Context

Two things happened recently:

1. Debugger moved to use `` for a bunch of icons (see e.g. #5658).
2. Gecko started rendering "broken images" placeholders in Firefox 65.

"Broken image" placeholders have two visual features:

- a border, something along the lines of `border: 1px inset black`, which *cannot* be overriden with the `border` property in author CSS;
- if the image has a `src` URL which returned an error (500, 404, etc.), the image will also show a "broken file" icon in the top left.

This is supposed to make Gecko match Blink in how it displays missing images, though I'm not sure if it's 100% the same heuristics.

## Visual issue in Debugger

This border painted on the "background" of src-less img elements show up for a lot of Debugger icons in Firefox (Beta and Nightly). If we're using CSS masking, it only shows up if the masking icons has pixels near its edge. When we're not using CSS masking, as for some file type icons, the border shows up even more visibly.

debugger-icon-borders

We can locate src-less images with a bit of debugging CSS:

```css
img:not([src]) {
  mask: none !important;
  background: red !important;
}
```

Which gives us:

screenshot

## Solutions

In my tests, adding `border: none;` to the images didn't work. It's not the  element itself which has a border, it's a platform-specific shadow-domy element.

The best solution would be to avoid using `` without a src at all. It's already a stretch, semantically, to do that, and it may cause accessibility issues (how are src-less and alt-less images read by screen readers, exactly?).

I believe in most cases we should be able to replace that element with a `<span></span>`, or with a pseudo-element (e.g. `before`), especially when we already have an accessible label:

1. In context, e.g. `<button><span class="icon"></span> Some label</button>`
2. On the parent element as aria-label, e.g. `<button aria-label="Some label"><span class="icon"></span></button>`.

The slightly trickier part is when we're trying to show a library's icon. This is relevant information that doesn't exist in context, so we're supposed to have alt text there. Solutions include:

- `Vue`
- `Vue` and a CSS background image to show the actual icon

Transparent PNG image taken from: http://png-pixel.com/

darkwing:
Fixes #7520

This PR prevents the problem brought whereby an `` without an `src` attribute gets a gross border.  I've switched to rendering a `<span>` with a default size and display, and everything's fallen into place nicely.


jasonLaster:
Hey @emilio I'm curious what you think of this bug we're seeing.

darkwing:
@emilio Is there a way via CSS to override that shadow border?  If only for the short term?

darkwing:
https://github.com/devtools-html/debugger.html/pull/7527 Should go in before this

emilio:
Hmm, not really, though I can add a way to force it to not show if you need it. We have the netscape-days `-moz-force-broken-image-icon` property that I could repurpose for this but... :)

jasonLaster:
@emilio `-moz-force-broken-image-icon` nice :)

I don't think we need it at the moment, but good to know about 😉

PR #7549, opened 2018-12-18
(WIP) Enable type checking for actions

gregtatum:
I spent a bit of time on this yesterday, but probably won't have enough time or understanding of some of the type errors I surfaced to take this PR to completion. I found that all the actions were set to `any` types, and I took a quick stab at fixing it. I surfaced some more issues in the types, that may be a bit hard to resolve. I changed a decent amount to get most of the flow errors removed, but there are still some things remaining. I will add inline comments about this investigation.


src/components/App.js
```diff
@@ -63,13 +63,13 @@ type Props = {
...
   quickOpenEnabled: boolean,
-  setActiveSearch: string => void,
-  closeActiveSearch: () => void,
-  closeProjectSearch: () => void,
-  openQuickOpen: (query?: string) => void,
-  closeQuickOpen: () => void,
-  setOrientation: OrientationType => void,
-  canRewind: boolean
+  canRewind: boolean,
+  setActiveSearch: typeof actions.setActiveSearch,
```
gregtatum: In the profiler we use a lot of type inference for our actions like this. I've been a big fan of it from the maintainability of the code.

* There is only one type definition, the action creator.
* When modifying the action creator, all of the consuming components will provide actionable errors when the type signature changes.
* There are not duplicate definitions all over the place.

The one problem with this is that `mapDispatchToProps` actually modifies the type signature of the functions, and the react-redux definition does not take this into account. This can cause a few inconsistencies in approach.

Original type signatures of the action creator:

```
// Action creator
(Arg1, Arg2) => Action
// Thunk action creator
(Arg1, Arg2) => (Dispatch, GetState) => ReturnType
```

Then `mapDispatchToProps` changes them to:

```
// Action creator
(Arg1, Arg2) => void
// Thunk action creator
(Arg1, Arg2) => ReturnType
```

I'm not sure a great solution to this. It turns out some of the profiler's types aren't correct here either.

src/components/Editor/SearchBar.js
```diff
@@ -176,13 +175,15 @@ class SearchBar extends Component<Props, State> {
       return;
     }
 
+    // TODO - What is this editor?
```
gregtatum: Needs better types.

src/components/Editor/SearchBar.js
```diff
@@ -176,13 +175,15 @@ class SearchBar extends Component<Props, State> {
...
     this.props.doSearch(query, this.props.editor);
   };
 
   updateSearchResults = (characterIndex, line, matches) => {
     const matchIndex = matches.findIndex(
       elm => elm.line === line && elm.ch === characterIndex
     );
+    // TODO - This is sending in the wrong values.
```
gregtatum: This looks like an error.

src/components/PrimaryPanes/index.js
```diff
@@ -46,7 +47,7 @@ class PrimaryPanes extends Component<Props, State> {
     };
   }
 
-  showPane = (selectedPane: string) => {
+  showPane = (selectedPane: SelectedPrimaryPaneTabType) => {
```
gregtatum: This type is a bit stricter here.

src/components/ProjectSearch.js
```diff
@@ -329,7 +322,10 @@ export class ProjectSearch extends Component<Props, State> {
         onBlur={() => this.setState({ inputFocused: false })}
         onKeyDown={this.onKeyDown}
         onHistoryScroll={this.onHistoryScroll}
-        handleClose={this.props.closeProjectSearch}
+        handleClose={
+          // TODO - This function doesn't quite match the signature.
+          (this.props.closeProjectSearch: any)
```
gregtatum: The type signature doesn't match the usage. This needs the `SearchInput` component's types to be more permissive to what you give it I think.

src/components/QuickOpenModal.js
```diff
@@ -221,12 +221,13 @@ export class QuickOpenModal extends Component<Props, State> {
     }
 
     if (this.isFunctionQuery()) {
-      return highlightLineRange({
+      const range = {
```
gregtatum: Woops, this change should be reverted. I was playing with the types here.

src/components/QuickOpenModal.js
```diff
@@ -434,11 +435,9 @@ function mapStateToProps(state) {
 export default connect(
   mapStateToProps,
   {
-    shortcutsModalEnabled: actions.shortcutsModalEnabled,
```
gregtatum: These were unused. It's easy to overshadow DispatchProps, with OwnProps. The current typing situation doesn't guard against overshadowing props with multiple things, but it does check for improper type usage.

PR #7550, opened 2018-12-18
Add context menu to breakpoint headings #7474

### LINKED ISSUE #7474: Breakpoint Headings Should Have Context Menu
It would be nice to have a context menu here to:

- "Disable breakpoints" - just in this source
- "Enable breakpoints" - just in this source
- "Remove breakpoints" - just in this source

Transfusion:
Fixes #7474 

### Summary of Changes

* Created another `PureComponent` similar to `BreakpointsContextMenu.js` to show the context menu with items.
* Create `{enable,disable,remove}BreakpointsInSource` as Redux actions in `breakpoints`, using the already available `getBreakpointsForSource` function
* Connect them to `BreakpointHeading.js` and pass them through into the new `PureComponent`, to be called on `click`

Demo video

### DIFF
--- a/assets/panel/debugger.properties
+++ b/assets/panel/debugger.properties
@@ -332,6 +332,15 @@ searchPanelGoToLine=Go to line (%S)
 # filter panel popup for the variables search operation.
 searchPanelVariable=Filter variables (%S)
 
+# LOCALIZATION NOTE (breakpointHeadingMenuItem): The text for all the elements
+# that are displayed in the breakpoint headings menu item popup.
+breakpointHeadingsMenuItem.enableInSource.label=Enable breakpoints
+breakpointHeadingsMenuItem.enableInSource.accesskey=E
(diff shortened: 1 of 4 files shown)


assets/panel/debugger.properties
```diff
@@ -332,6 +332,15 @@ searchPanelGoToLine=Go to line (%S)
 # filter panel popup for the variables search operation.
 searchPanelVariable=Filter variables (%S)
 
+# LOCALIZATION NOTE (breakpointHeadingMenuItem): The text for all the elements
```
flodolo: This comment will be displayed only for the first string, but I don't think there's a lot of value in adding it to all strings. For clarity, maybe `# LOCALIZATION NOTE (breakpointHeadingMenuItem.*)`

darkwing:
Wow, thank you so much for this @Transfusion !  I'll have a look today! :

Issue #6094, opened 2018-04-25
Full Project Search is really slow


andreicristianpetcu:
I use Ctrl+Shift+F in a pretty large Angular project. It takes a lot of time and it seems not to find any results. Also it generates a lot of LOAD_SOURCE_TEXT events.

Here is a video: https://vimeo.com/266454867

I cannot share the app I am working on :( but I can run de debugger locally with whatever profiling tool you would like. I can build it from branch if that helps.

andreicristianpetcu:
I attached the profile zip file. You can just hit F12 in Firefox go to performance and click the import button to see it visually. 
[profile.json.zip](https://github.com/devtools-html/debugger.html/files/1946233/profile.json.zip)

Would a profile with Gecko Profiler help?

nchevobbe:
Thanls a lot @andreicristianpetcu . Are you running the project from firefox devtools (if so which version), or via a clone of this repo ?
Could you also try to profile with http://perf-html.io/ and upload the profile please ?

andreicristianpetcu:
I'm running it from this commit

50e1c03 April 24th (#6084) (Jason Laster, 14 hours ago)

andreicristianpetcu:
Here you go! @nchevobbe https://perfht.ml/2FfedSg Happy hacking! :)

AnshulMalik:
Current issue with full search is that we first request a source text from server and then search in it on the client side.
So when there are a lot of files, a lot of time goes in this source text transfer, from server to client and from client to web worker. 

Can we move this search operation on server side? 
If we do, we will receive few lines per source from server and if user chooses to go to file, we can then request the entire file.

jasonLaster:
@AnshulMalik server-side approach would be great.

We should also limit the number of results shown. so pagination with scolling would be cool. Like show 100 results, then when you scroll down show a link for more or automatically render more...

WenInCode:
I started to take a peak at this. 

From the [debugger ux review doc](https://docs.google.com/document/d/1BhrOmy8kLOMOcJ5UgY7zlWuxnb3t6LfYbbVWULpcr54)
> Search sources
Big performance issue - should show first e.g. 10 results for each file, with more button

I did a quick spike into limiting matches per file, it definitely sped big searches up a lot. I searched for "e" on gmail before and it would time out, but after limiting it only took a few seconds.

jasonLaster:
@AnshulMalik lets close this when your work finishes up

Issue #5328, opened 2018-02-06
[ProjectSearch] makes tab unresponsive


lukaszsobek:
STR:
- go to http://todomvc.com/
- open debugger
- open project search `ctrl+shift+f`
- type `a`
- press `enter`
- wait for the search to finish (about 40k results)

ER: 
- browser tab is responsive

AR:
- browser tab / browser crashes

codehag:
thanks for reporting this! it sounds pretty bad, we will need to look into it

Fischer-L:
There are 2 bottlenecks here:
1. In `searchSources` action: it would block if searched on multiple large files [1].
2. In `renderResults` in `ProjectSearch` component: it would block as well if rendered a huge amount of matched results [2].

Probably we have to divide search operations, such as if there were 30 files, and we got 300 matched results from the first 5 files, we should pause searching, then went rendering. To search then render over 1000+ matched results is slow as well as a bit overwhelming for people.

[1] [https://github.com/devtools-html/debugger.html/blob/a986a9b1ddba05db0e52a0ca0fff34772ed1e8ad/src/actions/project-text-search.js#L59](https://github.com/devtools-html/debugger.html/blob/a986a9b1ddba05db0e52a0ca0fff34772ed1e8ad/src/actions/project-text-search.js#L59)

[2] [https://github.com/devtools-html/debugger.html/blob/7c0deaaf4cf030f7e9027d10bb4eacb2f5ceda98/src/components/ProjectSearch.js#L255](https://github.com/devtools-html/debugger.html/blob/7c0deaaf4cf030f7e9027d10bb4eacb2f5ceda98/src/components/ProjectSearch.js#L255)

jasonLaster:
- we can exclude 3rd party searching by default
- if it is react we can show 1000 results at a time
- we can limit search so you have to have atleast 3 chars

AnshulMalik:
Here are my findings, I searched for `ab` on http://todomvc.com/ 

<img width="1680" alt="screen shot 2018-05-21 at 12 34 52 pm" src="https://user-images.githubusercontent.com/7821757/40294596-e9d6fd02-5cf3-11e8-88ce-cf87c11c54ad.png">

The profile capture took around 150 sec.
Here are major culprits:
Layout: 130000 ms
Tree Component: 433ms

It's probably because the search results contain very long lines.
[...]

PR #7561, opened 2018-12-19
Don't send whole line across

### LINKED ISSUE #7557: Project search takes huge amount of memory
**Steps to reproduce**
1. Open https://nchevobbe.github.io/demo/console-test-app.html (in toolbox)
1. Open the debugger
1. Do a project search with `event`

**Expected results**

I get some results, or a UI telling me there's no result

**Actual results**

The panel shows "Loading", doesn't update. On the Activity Monitor, I can see the memory growing fast (> 2 GB after 10s, with only 1 tab)
if I don't kill the browser, it eats up all my memory and make the OS unusable, forcing a hard reboot.

---

Reproduced on beta (65) and release (64)

AnshulMalik:
Fixes #7557 #6094 #5328

### Summary of Changes

Most of the time just see few words before and after the search keyword and make a decision, so it seems useless to send whole line all the way from worker to main thread.

- This PR tries to start the line few words before the keyword and similarly not going all the way to the end, stops after some characters. 

If someone really wants to see ahead, they can just click the line and verify. But I think this should help 99% cases.

GIF: http://recordit.co/UKLvMy0wuH/gif/notify


src/actions/tests/__snapshots__/project-text-search.spec.js.snap
```diff
@@ -125,23 +125,23 @@ Array [
         "column": 9,
         "line": 1,
         "match": "foo",
-        "sourceId": "foo1",
+        "tc": 9,
```
AnshulMalik: Yes, I am bad at naming things.

src/workers/search/project-search.js
```diff
@@ -4,32 +4,65 @@
...
+);
+
+function truncateLine(text, column) {
+  if (text.length < 100) {
+    return {
+      c: column,
+      value: text
+    };
+  }
+  const offset = Math.max(column - Math.floor(Math.random() * 10 + 30), 0);
```
nchevobbe: from here, things are getting quite cryptic. 
Comments would definitely help :)

Also, why do we have `Math.random()` here?

src/workers/search/project-search.js
```diff
@@ -4,32 +4,65 @@
...
+      line: line + 1,
+      column: ch,
+      tc: c,
+      match: queryText,
+      value
+    };
+  });
+}
+
+const startRegex = /([ !@#$%^&*()_+\-=\[\]{};':"\\|,.<>\/?])/g;
```
nchevobbe: could we explain what those regex are about? They seem pretty complex

src/workers/search/project-search.js
```diff
@@ -4,32 +4,65 @@
...
+  [
+    "([ !@#$%^&*()_+-=[]{};':\"\\|,.<>/?])",
+    '[^ !@#$%^&*()_+-=[]{};\':"\\|,.<>/?]*$"/'
+  ].join("")
+);
+
+function truncateLine(text, column) {
+  if (text.length < 100) {
+    return {
+      c: column,
```
nchevobbe: let's return `column`, and not `c`

src/workers/search/project-search.js
```diff
@@ -4,32 +4,65 @@
...
-    .filter(_matches => _matches.length > 0);
-
-  return [].concat(...matches);
+
+  return getMatches(queryText, text, modifiers).map(({ line, ch }) => {
+    const { value, c } = truncateLine(lines[line], ch);
+    return {
+      line: line + 1,
+      column: ch,
+      tc: c,
```
nchevobbe: what does `tc` stand for? Could we have a better name (it maps to the returned column from `truncateLine`, so maybe it could simply replace `column` ?)

src/workers/search/project-search.js
```diff
@@ -4,32 +4,65 @@
...
+import getMatches from "./get-matches";
+
 export function findSourceMatches(source, queryText) {
-  const { id, loadedState, text } = source;
+  const { loadedState, text } = source;
   if (loadedState != "loaded" || !text || queryText == "") {
     return [];
   }
 
+  const modifiers = {
```
nchevobbe: I guess those should be picked up from the state instead of being hard coded here

src/workers/search/project-search.js
AnshulMalik: Yeah :P

Nice catch! It was useful when I started working on this patch, but no longer required.

src/workers/search/project-search.js
AnshulMalik: Yeah, probably because they are hand made :P

Start one is to find a word separator near the start of cropped line
And end is to find the same near the end, so the results look nice and clean.

src/workers/search/project-search.js
AnshulMalik: So there are two lines here, one is the original line(huge one). and column refers to the column in that line, since we need it to navigate on cliick.

And the is the new cropped line, and `tc` represents the column in the cropped string, since we need to highlight the match found.
I am so bad at naming things, I am sure you can find me a nice name :P

src/workers/search/project-search.js
AnshulMalik: For now, we don't have them for project search

src/workers/search/project-search.js
nchevobbe: so, looks like `tc` could be renamed to `matchIndex` or something similar?

princiya:
@AnshulMalik I was not able to understand how this PR solves issue #5328. Could you please help me understand better? TIA :)

AnshulMalik:
@princiya So in my analysis, I found a few things which cause the full project search to be really slow.

This PR addresses one of them:

Rendering of really long lines takes a lot of time, and generally minified sources are very long and one-liner. So if for such a source file, there are 3000 results, we were transferring 3000 copies of the source back to main thread as result strings. And sometimes it exceeds the max cap of what we can transfer. 
And it makes the rendering slower.

We are doing so much work of transferring and rendering and most of the work is being wasted, since a user just needs some context to find what they are looking for, so if we can just keep some context instead of the whole line, we can make it faster, keeping the same experience.

So this PR is about cutting short the whole line keeping some context to make sense of the results.

I hope it makes it a little bit easier to understand.

princiya:
@AnshulMalik thanks a lot, certainly helps in better understanding.

If you remove the [horizontal scrollbar](https://github.com/devtools-html/debugger.html/pull/7561#issuecomment-450004655), I think we are good to go with this PR 👍

PR #7424, opened 2018-12-05
Re-write getting started docs

b117014:
1. add getting started
2. add first activity
3. consolidate contributing
4. consolidate readme

Fixes #<issue number>

Here's the Pull Request Doc
https://firefox-dev.tools/debugger.html/docs/pull-requests.html

### Summary of Changes

* change 1
* change 2

### Test Plan

Tell us a little a bit about how you tested your patch.

Example test plan:

- [x] Command-P opens the panel
- [x] Clicking “+” opens the panel
- [x] Clicking a source navigates to the source
- [x] Clicking "x" closes the panel

Here's the Debugger's Testing doc
https://docs.google.com/document/d/1oBMRxV8A2ag2t22YsQOxTdEv0mXKzIg0tggJjRkU1S0/edit#.
Feel free to improve it!

PR #7562, opened 2018-12-19
Documentation improvements

darkwing:
Supercedes https://github.com/devtools-html/debugger.html/pull/7424/files

I tried finding the most up to date information, though the original PR appears to be on a very old branch.

### DIFF
--- a/docs/getting-setup.md
+++ b/docs/getting-setup.md
@@ -2,55 +2,66 @@
 
 ![][debugger-intro-gif]
 
-:construction_worker: If something goes wrong, checkout our [most common issues][common-issues] and find us in [Slack][slack]!
-
-### Step 1. Installing the Debugger
-
-First, we'll clone the debugger locally, then we'll install its
-dependencies with [Yarn]. If you don't have Yarn, you can install it [here][yarn-install].
-
-Also, before you start, it's helpful to make sure you have node 7.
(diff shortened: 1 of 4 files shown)


docs/getting-setup.md
```diff
@@ -2,55 +2,66 @@
 
 ![][debugger-intro-gif]
 
-:construction_worker: If something goes wrong, checkout our [most common issues][common-issues] and find us in [Slack][slack]!
+### Step 1. Install Yarn
 
-### Step 1. Installing the Debugger
+```bash
+npm i -g yarn@0.16.1
```
wldcordeiro: This seems like a very old version of Yarn to use. 1.x is stable, shouldn't we use that?

docs/local-development.md
```diff
@@ -486,24 +485,10 @@ yarn run test:all
...
-This will run all the tests that have not been committed. Basically, all the files that are returned by the `git status` command.
-
-If the snapshot changes then update it with:
-
-```bash
-node_modules/jest/bin/jest.js -o -u
-```
+* `yarn test` - Run tests headlessly
+* These are the basic unit tests which must always pass
+* `yarn run mocha-server` - Run tests in the browser once you open `http://localhost:8003`
```
jasonLaster: i don't think this is a thing any more :)

PR #7564, opened 2018-12-20
Improve File Icon Visibility in Search Results On File Selection

### LINKED ISSUE #7558: Icon of selected file in ProjectSearch result should be white
**Steps to reproduce**
1. Do a Project Search that return some results
1. Select a file in the results

**Results**

image

---

The icon is not legible because of the blue background. We should make it the same color as the text.

darkhist:
Fixes #7558 

### Summary of Changes

* Update file icon in Search Results from grey to white when a search result is selected

### Test / Reproduction

1. `yarn start`
2. Visit `localhost:8000` and open Firefox or Chrome
3. Select a sample project, and then press `cmd + shift + f` to search
4. Select one of the files (the file icon is now visible) 

### Screenshots
icon-color-update

### DIFF
--- a/src/components/ProjectSearch.css
+++ b/src/components/ProjectSearch.css
@@ -121,6 +121,10 @@
 
 /* Focus values */
 
+.project-text-search .file-result.focused .img.file {
+  background-color: white;
+}
+
 .project-text-search .file-result.focused,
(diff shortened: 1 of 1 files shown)


jasonLaster:
Thanks @darkhist

PR #7565, opened 2018-12-20
Use CSS animation for the SearchInput loader.

nchevobbe:
This was using SVG previously, which is not as
fast as CSS. The performance win is the most
visible when there is a lot of search results
to render.


nchevobbe:
Before: https://files.slack.com/files-pri/T3V7H1198-FEY3Y9L0H/download/before.mov

After: https://files.slack.com/files-pri/T3V7H1198-FEY3YDMUH/download/after.mov

PR #7573, opened 2018-12-20
Fixes flow typing support for reselect

### LINKED ISSUE #7568: [flow] add types for reselect
We recently [upgraded][u] to flow 89 and had to add an adapter to reselect's createSelector function to keep it from giving us type errors. It would be nice to revisit this and add types, which will let us infer values from all of our memoized functions 

```js
 function createSelector(...args) {
   // $FlowFixMe
   return reselectCreateSelector(...args);
 }
 
 export { createSelector };
````

[u]: https://github.com/devtools-html/debugger.html/pull/7526

### PR DESCRIPTION
**Pull request by ryanjduffy**
Fixes #7568

### Summary of Changes

* Removes temporary `createSelector.js` file
* Adds generic type `Selector<T> = State => T` (borrowed from [perf.html](https://github.com/devtools-html/debugger.html/issues/7568#issuecomment-449083588)) to cast the return function of `createSelector`

### Note for the future

I noticed a case (resolved by 73d7041) in which Flow reported problems figuring out which overloaded version of `createSelector` to use. The root cause turned out to be a mismatch between the return type of an input selector (`?Source`) and the type of the return value in the result function (`Source`). 

The reported error:

```
Cannot call createSelector because:
 • Either no more than 2 arguments are expected by function type [1].
 • Or no more than 2 arguments are expected by function type [2].
````

gave no indication that was the problem so noting here in case it helps others in the future.

### CODE REVIEW

File: src/selectors/visibleBreakpoints.js [hunk 1]
```diff
@@ -47,30 +48,33 @@ function isVisible(breakpoint, selectedSource) {
...
-  (selectedSource: Source, breakpoints: Breakpoint[]) => {
-    if (!selectedSource) {
+  (selectedSource: ?Source, breakpoints: Breakpoint[]) => {
+    if (selectedSource == null) {
       return null;
     }
 
+    // FIXME: Even though selectedSource is checked above, it fails type
+    // checking for isVisible
+    const source: Source = selectedSource;
```
jasonLaster: :/

src/components/Editor/Footer.js
```diff
@@ -248,12 +248,18 @@ class SourceFooter extends PureComponent<Props, State> {
 
 const mapStateToProps = state => {
   const selectedSource = getSelectedSource(state);
-  const selectedId = selectedSource.id;
+  let prettySource = null;
+  let mappedSource = null;
+
+  if (selectedSource) {
+    prettySource = getPrettySource(state, selectedSource.id);
+    mappedSource = getGeneratedSource(state, selectedSource);
```
jasonLaster: i wonder if we can change these functions to just return null if there is no source

src/components/Editor/index.js
```diff
@@ -629,7 +629,7 @@ const mapStateToProps = state => {
     selectedSource,
     searchOn: getActiveSearch(state) === "file",
     conditionalPanelLocation: getConditionalPanelLocation(state),
-    symbols: getSymbols(state, selectedSource)
+    symbols: selectedSource ? getSymbols(state, selectedSource) : null
```
jasonLaster: same here

src/selectors/breakpointSources.js
```diff
@@ -84,12 +83,17 @@ function findBreakpointSources(
   return sortBy(breakpointSources, (source: Source) => getFilename(source));
 }
 
-export const getBreakpointSources = createSelector(
+export const getBreakpointSources: Selector<
+  Array<{
+    source: Source,
+    breakpoints: FormattedBreakpoint[]
+  }>
```
jasonLaster: Maybe we pull this into its own type 

```js
type SourceBreakpoints = {|
   +source: Source,
   +breakpoints: FormattedBreakpoint[]
 |}
```

ryanjduffy: You already had! I just missed it since it wasn't declared with the function.

src/components/Editor/Footer.js
loganfsmyth: I think the downside of that is that it means the return value has to be nullable, so even if we know the input exists, Flow will end up requiring us to assert that the return value is also not nullable.

src/reducers/pause.js
```diff
@@ -556,7 +564,7 @@ export function getSkipPausing(state: OuterState) {
 
 // NOTE: currently only used for chrome
 export function getChromeScopes(state: OuterState) {
-  const frame = getSelectedFrame(state);
+  const frame: ChromeFrame = (getSelectedFrame(state): any);
```
loganfsmyth: Shouldn't this be `?ChromeFrame`?

src/selectors/visibleBreakpoints.js
loganfsmyth: This is because type refinements in outer functions don't stay refined inside callbacks. I think this is a very reasonable workaround for that and doesn't need a FIXME.

PR #7581, opened 2018-12-21
pretty print exception bug

### LINKED ISSUE #7477: PrettyPrint: uncaught exception: Error: Line numbers must be >= 1
STR:

1.  Open any JS tab
2.  Click pretty print button
3.  (Everything works)
4.  Close pretty tab
5.  Click pretty print button again
6.  No pretty tab opens, see error:  `uncaught exception: Error: Line numbers must be >= 1`

jainsneha23:
Fixes #7477 

### Summary of Changes

* Pass line number as is, from utils/location.js

### Test Plan

- [x] Open a file, pretty print it
- [x] Close the pretty printed file tab
- [x] Again pretty print it
- [x] The prettified file opens without error

### DIFF
--- a/src/utils/location.js
+++ b/src/utils/location.js
@@ -15,14 +15,14 @@ type IncompleteLocation = {
 
 export function createLocation({
   sourceId,
-  line,
+  line = 1,
   column,
-  sourceUrl
+  sourceUrl = ""
 }: IncompleteLocation): SourceLocation {
   return {
     sourceId,
-    line: line || 0,
+    line,
     column,
-    sourceUrl: sourceUrl || ""
+    sourceUrl
   };
 }


src/utils/location.js
```diff
@@ -21,7 +21,7 @@ export function createLocation({
 }: IncompleteLocation): SourceLocation {
   return {
     sourceId,
-    line: line || 0,
+    line,
```
jainsneha23: Or should this be 1 by default. As line number starts with 1, and cannot be < 1.

jainsneha23:
Note: This error got introduced in this commit 4a43e9a

After introducing flow, line number for SourceLocation has to be a number. If set to zero, it throws the current error. If set to 1, there's a flash at line 1 while opening a file in the editor. NaN seems to be a workaround, but doesn't make any sense. So setting it as 1 for now. May be we need to handle the flash somehow.

jasonLaster:
This looks good. CC @darkwing

AnshulMalik:
Looking good 👍

PR #7582, opened 2018-12-21
Fix sources/outline pane vertical padding

### LINKED ISSUE #7536: Wrong vertical padding in Sources pane
It might be a regression introduced in the PR for #6804.

The `padding: 4px 0;` we wanted to have on the Sources and Outline panes should be set an element that is either the scrolling context or inside the scrolling context. Currently it's set on a parent, so when a scrollbar appears it does not touch the edges (neither does the content):

screen shot

Garee:
Specify a 4px top padding for the tree element and remove the top
margin from its first child. This fixes the gap between the list
scrollbar and the top of the container when it overflows.

Simplify the overflow styling by removing the overflow styles from
the children of .source-outline-panel. This is now the explicit
scrollable container.

Remove the margin from the outline filter and give it the same 4px
top padding to maintain consistency across tabs.

Updated the expected test results as the inline styles of the tree were changed. 
The default padding and margin of 0 were removed as they looked unnecessary.

Fixes #7536.

### DIFF
--- a/packages/devtools-components/src/tree.js
+++ b/packages/devtools-components/src/tree.js
@@ -806,10 +806,7 @@ class Tree extends Component {
       });
     });
 
-    const style = Object.assign({}, this.props.style || {}, {
-      padding: 0,
-      margin: 0
-    });
+    const style = Object.assign({}, this.props.style || {});
 
     return dom.div(
       {
--- a/packages/devtools-reps/src/object-inspector/tests/component/__snapshots__/classnames.js.snap
+++ b/packages/devtools-reps/src/object-inspector/tests/component/__snapshots__/classnames.js.snap
@@ -9,12 +9,7 @@ exports[`ObjectInspector - classnames has the expected class 1`] = `
   onKeyPress={[Function]}
   onKeyUp={[Function]}
   role="tree"
-  style={
-    Object {
-      "margin": 0,
-      "padding": 0,
-    }
-  }
+  style={Object {}}
   tabIndex="0"
 >
   <div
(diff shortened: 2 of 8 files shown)


nchevobbe:
The tree component is used in many places, possibly outside the debugger.
We should be extra careful when changing style on it.
Here we should target the sources panel specifically.

Le ven. 21 déc. 2018 à 22:30, Gary Blackwood <notifications@github.com> a
écrit :

fvsch:
> Here we should target the sources panel specifically.

@nchevobbe Any pointers how to do that effectively?
Does that mean we should leave ManagedTree.css alone, and only work in `src/components/PrimaryPanes/Sources.css`?

I would also be interested in knowing why we have `overflow-x:hidden;overflow-y:auto;` on a container, then `overflow:auto` on its content. (Plus at least one more container with some `overflow` styles.) This looks fishy, and it's hard to tell if it's accumulated cruft or if it serves a purpose.

Garee:
I see I need to add the overflow style back to `.managed-tree .tree` so that other areas are not effected. I'll update and test.

packages/devtools-components/src/tree.js
```diff
@@ -806,10 +806,14 @@ class Tree extends Component {
...
     });
 
-    const style = Object.assign({}, this.props.style || {}, {
-      padding: 0,
-      margin: 0
-    });
+    const style = Object.assign(
+      {},
+      {
+        padding: 0,
```
nchevobbe: we should maybe check why are those needed, and maybe remove them.
It would be better to style it through CSS when/if needed so we don't have such a string specificity.

nchevobbe: yeah, I don't think it has any effect on the console, nor on the scopes panel.
So let's change it into: 

```js
const style = Object.assign({}, this.props.style || {});
```

and not override it in JS from src/components/PrimaryPanes/SourcesTree.js, but target the CSS class from there (i.e. `.sources-list .managed-tree .tree`).

Garee: Thanks @nchevobbe. I'll make the suggested changes.

PR #7583, opened 2018-12-21
Windowless worker sources

bhackett1024:
### Summary of Changes

All changes below only apply with the windowless workers pref set.

* Sources for different threads are shown separately in sources pane.
* Multiple tabs for the same file may be opened if they are associated with different workers.
* Workers are named in the workers pane, sources pane, source tab titles, and breakpoint sources.

Naming workers here (just done numerically for now, e.g. "Worker #1") is different from what Chrome does but is helpful for when we need to distinguish the same file when it is opened by multiple threads.  The plan is to treat each such source separately, with individual sets of breakpoints and separate tabs in the editor pane.  Behavior in this area is kind of weird in Chrome, as you can get two tabs for the same file that seem to share the same set of breakpoints (i.e. it doesn't seem like breakpoints can be set for particular threads).  Using the worker names avoids ambiguity but in many cases --- where a source has only been opened by one thread --- they could potentially be avoided.

### Test Plan

Automated tests needed.

screen shot 2018-12-21 at 8 42 05 am


bhackett1024:
> Lets think UX to distinguish multiple worker instances of the same file. Do we have some unique ID for a worker instance that could be used to distinguish instances?

Internally we use the actor ID to distinguish between threads, but that shouldn't show up in the UI anywhere.  I think the options for the UI are some combination of the worker's URL, any user-assigned name, and the counter for how many workers have been created.

> Sharing breakpoints between instances of the same file might be the more expected behavior, given how short-lived workers can be and the same file potentially being used for several instances sequentially.

I think that whether breakpoints are shared is tied in with other parts of the UI: if we share breakpoints between sources with the same URL, we should only show one source tab for that URL.  I outlined the choices I see in [bug 1251958](https://bugzilla.mozilla.org/show_bug.cgi?id=1241958#c4), and the preference there was to have a 1:1 correspondence between the UI and internal state like sources and installed breakpoints, as this PR does.

jasonLaster:
Thanks @bhackett1024. I'll review shortly. 

Thanks for the feedback @digitarald. The current "literal" approach is a necessary first step for representing thread specific sources and breakpoints in the UI. 

I expect we will want to create a PRD for worker debugging in January and iron out the different use cases.

src/components/PrimaryPanes/SourcesTree.js
```diff
@@ -352,18 +379,21 @@ function getSourceForTree(state: AppState, source: ?Source): ?Source | null {
...
 
   return {
     shownSource: getSourceForTree(state, shownSource),
     selectedSource: getSourceForTree(state, selectedSource),
     debuggeeUrl: getDebuggeeUrl(state),
-    expanded: getExpandedState(state),
+    expanded: getExpandedState(state, props.thread),
+    focused: focused && focused.thread == props.thread ? focused.item : null,
     projectRoot: getProjectDirectoryRoot(state),
     sources: getRelativeSources(state),
```
jasonLaster: it might be nice to have a selector `getRelativeSourcesForThread` this will simplify some of the logic in the tree and perhaps keep the tree from updating if a source for another thread is received.

src/components/PrimaryPanes/SourcesTree.js
```diff
@@ -352,18 +379,21 @@ function getSourceForTree(state: AppState, source: ?Source): ?Source | null {
...
-    sourceCount: getSourceCount(state)
+    sourceCount: getSourceCount(state),
```
jasonLaster: perhaps `getSourceCountForThread`

src/components/PrimaryPanes/SourcesTree.js
```diff
@@ -352,18 +379,21 @@ function getSourceForTree(state: AppState, source: ?Source): ?Source | null {
...
   const shownSource = getShownSource(state);
+  const focused = getFocusedSourceItem(state);
 
   return {
     shownSource: getSourceForTree(state, shownSource),
     selectedSource: getSourceForTree(state, selectedSource),
     debuggeeUrl: getDebuggeeUrl(state),
-    expanded: getExpandedState(state),
+    expanded: getExpandedState(state, props.thread),
+    focused: focused && focused.thread == props.thread ? focused.item : null,
```
jasonLaster: nit. Maybe `getFocusedSourceItem` could optionally take `thread` and return the item if it matches... could be too cute

src/utils/sources-tree/updateTree.js
```diff
@@ -25,43 +21,13 @@ function newSourcesSet(newSources, prevSources) {
   return uniqSources;
 }
 
-function findFocusedItemInTree(
```
jasonLaster: it's great to see this function go away

src/components/shared/ManagedTree.js
```diff
@@ -129,33 +123,23 @@ class ManagedTree extends Component<Props, State> {
         );
 
       if (highlightItems[index]) {
-        this.focusItem(highlightItems[index]);
+        this.props.onFocus(highlightItems[index]);
       }
     }
   }
 
-  focusItem = (item: Item) => {
```
jasonLaster: it is really nice to simplify this component

src/components/PrimaryPanes/tests/SourcesTree.spec.js
```diff
@@ -238,15 +243,15 @@ describe("SourcesTree", () => {
...
       await component.update();
       await component
         .find(".sources-list")
         .simulate("keydown", { keyCode: 13 });
-      expect(props.selectSource).toHaveBeenCalledWith(
-        "server1.conn13.child1/39"
-      );
+      // expect(props.selectSource).toHaveBeenCalledWith(
+      //  "server1.conn13.child1/39"
+      // );
```
jasonLaster: ?

src/components/PrimaryPanes/index.js
```diff
@@ -118,15 +131,18 @@ class PrimaryPanes extends Component<Props, State> {
 
 const mapStateToProps = state => ({
   selectedTab: getSelectedPrimaryPaneTab(state),
-  sources: getSources(state),
-  sourceSearchOn: getActiveSearch(state) === "source"
+  sources: getRelativeSources(state),
+  sourceSearchOn: getActiveSearch(state) === "source",
+  getWorkerDisplayName: thread => getWorkerDisplayName(state, thread)
```
jasonLaster: we've passed selectors into components in the past a couple of times. Generally, we tend to regret it for a couple of reasons:

1. it is a new function every time which causes react to try re-rendering pretty often
2. it tends to encourage us to put too much logic in the components

In this case, i think we can do two things:

1.  store an index in the debuggee/workers reducer field to represent the index that worker w/ the same URL appeared. 
2. query for the workers here and pass workers into a util `getWorkerDisplayName(workers, thread)` or something like that

src/components/PrimaryPanes/index.js
```diff
@@ -118,15 +131,18 @@ class PrimaryPanes extends Component<Props, State> {
...
   mapStateToProps,
   {
     setPrimaryPaneTab: actions.setPrimaryPaneTab,
     setActiveSearch: actions.setActiveSearch,
     closeActiveSearch: actions.closeActiveSearch
   }
-)(PrimaryPanes);
+);
+
+export default connector(PrimaryPanes);
```
jasonLaster: why do we need `connector`?

src/components/PrimaryPanes/SourcesTree.js
```diff
@@ -110,32 +107,26 @@ class SourcesTree extends Component<Props, State> {
...
       debuggeeUrl != nextProps.debuggeeUrl ||
-      thread != nextProps.thread ||
       nextProps.sourceCount === 0
     ) {
       // early recreate tree because of changes
       // to project root, debugee url or lack of sources
       return this.setState(
         createTree({
-          sources: nextProps.sources[nextProps.thread],
+          sources: nextProps.sources,
```
jasonLaster: i think it is nice to keep the sources tree thread free as for the most part the tree should not know about other threads

src/components/PrimaryPanes/SourcesTree.js
```diff
@@ -110,32 +107,26 @@ class SourcesTree extends Component<Props, State> {
...
         })
       );
     }
 
-    if (
-      nextProps.shownSource &&
-      nextProps.shownSource.thread == thread &&
-      nextProps.shownSource != shownSource
-    ) {
+    if (nextProps.shownSource && nextProps.shownSource != shownSource) {
```
jasonLaster: these checks can be simplified if shownSource and selectedSource or null if they belong to a different thread

src/components/PrimaryPanes/SourcesTree.js
```diff
@@ -371,29 +362,38 @@ class SourcesTree extends Component<Props, State> {
...
+    selectedSource: getSourceForTree(state, selectedSource, thread),
     debuggeeUrl: getDebuggeeUrl(state),
     expanded: getExpandedState(state, props.thread),
     focused: focused && focused.thread == props.thread ? focused.item : null,
     projectRoot: getProjectDirectoryRoot(state),
-    sources: getRelativeSources(state),
-    sourceCount: getSourceCount(state),
-    workerDisplayName: getWorkerDisplayName(state, props.thread)
+    sources: getRelativeSourcesForThread(state, thread),
+    sourceCount: getSourceCount(state, props.thread),
```
jasonLaster: lets limit these to our own thread...

src/components/PrimaryPanes/index.js
```diff
@@ -101,7 +101,7 @@ class PrimaryPanes extends Component<Props, State> {
           ? 1
           : -1
     );
-    return threads.map(thread => <SourcesTree thread={thread} />);
+    return threads.map(thread => <SourcesTree thread={thread} key={thread} />);
```
jasonLaster: this was throwing a warning without a key

src/components/ProjectSearch.js
```diff
@@ -274,7 +274,6 @@ export class ProjectSearch extends Component<Props, State> {
           getParent={item => null}
           getPath={getFilePath}
           renderItem={this.renderItem}
-          onFocus={() => {}}
```
jasonLaster: i switched this to a default prop

src/components/QuickOpenModal.js
```diff
@@ -420,7 +420,7 @@ function mapStateToProps(state) {
 
   return {
     enabled: getQuickOpenEnabled(state),
-    sources: formatSources(getRelativeSources(state), getTabs(state)),
+    sources: formatSources(getRelativeSourcesList(state), getTabs(state)),
```
jasonLaster: this is a minor change, but i'd rather to the data munging in a selector which is easier to type and memoize.

src/reducers/sources.js
```diff
@@ -506,14 +496,22 @@ export function getUrls(state: OuterState) {
...
+    ? sourceList
+    : sourceList.filter(source => source.thread == thread);
 }
 
-export const getSourceCount: Selector<number> = createSelector(
-  getSources,
-  sources => Object.keys(sources).length
-);
+export function getRelativeSourcesList(state: OuterState): Source[] {
+  return ((Object.values(getRelativeSources(state)): any).flatMap(
```
jasonLaster: I needed any any here because flow doesn't know about flatMap

jasonLaster:
It looks like some of the workers' sources do not appear in the tree. Have you seen this as well? It looks like it might be because the sources did not stream down? Also it is possible we need to fetch the sources.

screen shot 2018-12-23 at 4 01 49 pm

jasonLaster:
@bhackett1024 I posted some "tweaks" which are basically opinionated refactorings. Feel free to revert any change, I shared it this way because it was easier for me to think in the editor.

bhackett1024:
> It looks like some of the workers' sources do not appear in the tree. Have you seen this as well? It looks like it might be because the sources did not stream down? Also it is possible we need to fetch the sources.

I've seen this sometimes but refreshing has fixed it for me (also, worker sources won't show when using the standalone debugger.html.)  I've also seen similar problems with the main thread's sources sometimes, and figured they are probably related and would best be fixed in followup.

PR #7585, opened 2018-12-22
Restyle tabs

### LINKED ISSUE #7494: Restyle Tabs
Idea from @jainsneha23:

> Let's remove the right border to this toggle button. And let it be independent of any line marker width. And the tabs bottom border should always be there. Also, we can change the hover CSS of the tabs, so that they doesn't highlight the border of the toggle button. We can have just the top borders, rather than the grey overlay in the tabs, while hover.

carlosangon:
Fixes #7494

### Summary of Changes

* Removed `left and right 1px border` from Tabs in editor's source-header.
* Added a few css properties to `toggle-button` since i tried to add `devtools-separator` but wouldn't display on correct side when panels collapsed, since icon rotates.
* Added `min-width` to `PaneToogleButton` as requested in ticket in order to align the separator to editor number's line, when displaying 2 to 3 digits lines.
* Removed CloseButton.js `transition: all ...;` property as mentioned.
* Removed `.source-tab` and `source-tab active` 1px transparent border from left and right side to reflect demo attached to issue #7585 (img-b)

 `devtools-separator as requested` 
devtools-separator

 `The close button's cross should use the tab's text color (blue when active, gray otherwise) as requested` 
close-btn-blue-when-active-gray on hover

As I was working through this task I realized that the "show more files dropdown icon is not showing, I redownloaded master, and discovered its not showing there either, see image attached.
master-morefiles-icon-missing

I could go ahead in fix it or do we wanna make an additional ticket for that?

[Dropdown icon fix]
Here is fix for the show more files dropdown icon not showing. I pretty much just moved a css class from one file to another. May not be the ultimate or most ideal, please advise.

dropdown-icon

### DIFF
--- a/src/components/shared/Dropdown.css
+++ b/src/components/shared/Dropdown.css
@@ -17,6 +17,10 @@
   overflow: auto;
 }
 
+html[dir="rtl"] .dropdown {
+  right: calc((var(--width) - 11px) * (-1));
+}
+
 .dropdown-block {
   padding: 0px 2px;
   position: relative;
--- a/src/components/shared/Dropdown.css
+++ b/src/components/shared/Dropdown.css
@@ -35,13 +35,18 @@
   width: 24px;
 }
 
-.moreTabs {
+.more-tabs {
   mask: url(/images/command-chevron.svg) no-repeat;
   mask-size: 100%;
   width: 12px;
   display: block;
   background: var(--theme-body-color);
   margin-left: 6px;
+}
+
+html[dir="rtl"] .img.more-tabs {
+  transform: rotate(180deg);
+  margin-right: 6px;
 }
 
 .dropdown li {
--- a/src/components/shared/Dropdown.css
+++ b/src/components/shared/Dropdown.css
@@ -17,10 +17,6 @@
   overflow: auto;
 }
 
-html[dir="rtl"] .dropdown {
-  right: calc((var(--width) - 11px) * (-1));
-}
-
 .dropdown-block {
   padding: 0px 2px;
   position: relative;
@@ -37,6 +33,15 @@ html[dir="rtl"] .dropdown {
   font-size: 14px;
   height: 100%;
   width: 24px;
+}
+
+.moreTabs {
+  mask: url(/images/command-chevron.svg) no-repeat;
+  mask-size: 100%;
+  width: 12px;
+  display: block;
+  background: var(--theme-body-color);
+  margin-left: 6px;
 }
 
 .dropdown li {
--- a/src/components/Editor/Tabs.css
+++ b/src/components/Editor/Tabs.css
@@ -38,12 +38,9 @@
 }
 
 .source-tab {
-  border-left: 1px solid transparent;
-  border-right: 1px solid transparent;
   display: inline-flex;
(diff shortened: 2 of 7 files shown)


carlosangon:
[WIP]

carlosangon:
@princiya I have pushed the

carlosangon:
@princiya Thank you. Yes, I will work on that last requested edit `show more files dropdown icon` today and will ping you when I push it.

carlosangon:
I'm not understanding what is going with the CircleCI test. All eslint test pass before pushing.

Any ideas?

src/components/shared/Dropdown.css
```diff
@@ -17,10 +17,6 @@
   overflow: auto;
 }
 
-html[dir="rtl"] .dropdown {
-  right: calc((var(--width) - 11px) * (-1));
```
AnshulMalik: This is not needed?

src/components/shared/Dropdown.css
```diff
@@ -39,6 +35,15 @@ html[dir="rtl"] .dropdown {
   width: 24px;
 }
 
+.moreTabs {
```
AnshulMalik: The convention in css is to use `-` instead of camel case, don't know how it slipped before. But this should be the right time to fix it :)

carlosangon: Thanks for the suggestion and your time @AnshulMalik 

implementing now.

princiya:
thanks @carlosangon for this PR. it is now 💯 👍 
@AnshulMalik thank you for your feedback too.

AnshulMalik:
This looks great! One thing to mention, that when no sources are open, we can see a separator on right side of `Collapse Pane` button which looks odd, we should only show it if some source is open.
May be we can try moving the style to `first` of the source tabs instead of the collapse button.

carlosangon:
@AnshulMalik Great observation. I can definitely take care of that. With that said, this PR completes all of requisites mentioned issue #7494. Would it be a good idea or ideal to squeeze it the fix in same PR?

AnshulMalik:
Yeah @carlosangon I think it'll be good if we can keep that change with this PR only :)

carlosangon:
Team , I will be working on this today.

darkwing:
Excellent, thank you @carlosangon !

darkwing:
@carlosangon The conflicting file, `debugger.css`, was added by accident at one point, so you can remove that change from the PR.

carlosangon:
@darkwing Thank you for taking the time to mention this to me. I will follow your recommendation and continue with my humble contribution. 

Glad the holidays are over btw.

darkwing:
@violasong @fvsch This is the current display of the PR.  Thoughts?

tabswhite

tabsdark

fvsch:
This looks great to me. Thanks @carlosangon.

carlosangon:
@fvsch Florens, thank you I followed your directions as mentioned in the issue.

darkwing:
Updated to remove the merge conflict.  If this goes green we can merge! 👍

carlosangon:
@darkwing Thanks for the correction David.

PR #7586, opened 2018-12-22
Fixes blurry Line breakpoint marker

### LINKED ISSUE #7569: Line breakpoint marker is blurry on 1x displays
The breakpoint SVG image (shown behind the line number in the gutter) is blurry on 1x displays.

breakpoint-marker

It seems there are two reasons for that:

1. It's using a 1px stroke on a `<path/>` which touches the edges of the viewport. This means that half the stroke is painted outside the viewBox, and clipped. So we end up with a 0.5px border, which looks blurry.
2. The icon is 12px tall, but we change its height to 14px. So everything is scaled, and our 0.5px visible stroke becomes 0.5833px wide.

So I've redrawn it for the actual 14px height and so that it gets a nicer stroke. See the results in the screenshot above.

The SVG for this redrawn image is straightforward:

```svg
<svg viewBox="0 0 60 14" xmlns="http://www.w3.org/2000/svg">
  <path d="M53.07.5H1.5a1 1 0 0 0-1 1v11a1 1 0 0 0 1 1h51.57a2 2 0 0 0 1.53-.7L59.3 7l-4.7-5.8a2 2 0 0 0-1.53-.7z"/>
</svg>
```

I've made the screenshot above by replacing the SVG node in the DOM with the code above. No CSS changes required.

Here's my Figma document for reference:
https://www.figma.com/file/Cb0LV3HXrfGzObAEodN8lt2d/Debugger-Breakpoint-marker

chocos10:
Fixes #7569

### DIFF
--- a/assets/images/breakpoint.svg
+++ b/assets/images/breakpoint.svg
@@ -1,6 +1,6 @@
 <!-- This Source Code Form is subject to the terms of the Mozilla Public
    - License, v. 2.0. If a copy of the MPL was not distributed with this
    - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
-<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" viewBox="0 0 60 12">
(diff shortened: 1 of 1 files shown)


AnshulMalik:
Thanks for your help @chocos10 .

PR #7589, opened 2018-12-23
Fix cmd/ctrl+shift+b shortcut to toggle conditional breakpoint

### LINKED ISSUE #7572: [breakpoints] cmd+shift+b fails to set a conditional bp
STR:

1. open a file in the debugger
2. click a line
3. press cmd+shift+b

ER: see a condtional panel
AR: nothing happens

I think this broke when we made the panel use a location object

Garee:
Currently, if `conditionalPanelLocation` is undefined we attempt to open the conditional panel at that location. Unsurprisingly, this fails as we do not know which source file or line to open at.

This change passes the source and line at which the shortcut was triggered.

While debugging this issue I spotted that the keyboard event for the shortcut is not being fired in Firefox. If the key combination is changed to say, `CmdOrCtrl+E`, it is fired. Chrome has no such issue.

Fixes #7572

### DIFF
--- a/src/components/Editor/index.js
+++ b/src/components/Editor/index.js
@@ -433,14 +433,23 @@ class Editor extends PureComponent<Props, State> {
     const {
       conditionalPanelLocation,
       closeConditionalPanel,
-      openConditionalPanel
+      openConditionalPanel,
+      selectedSource
     } = this.props;
 
     if (conditionalPanelLocation) {
       return closeConditionalPanel();
     }
 
-    return openConditionalPanel(conditionalPanelLocation);
+    if (!selectedSource) {
+      return;
+    }
+
+    return openConditionalPanel({
+      line: line,
+      sourceId: selectedSource.id,
+      sourceUrl: selectedSource.url
+    });
   };
 
   closeConditionalPanel = () => {


jasonLaster:
Hmm, just tried cmd+shift+b on my mac and it seemed to work. One thing i did notice is that I had to turn off column breakpoints, `dbg.features.columnBreakpoints = false` for it to work though.

Garee:
I had `privacy.resistFingerprinting` set to `true` in `about:config`. It works if I disabled that - strange!

princiya:
I tested this PR and it works for me with `cmd+shift+b` on mac.

@Garee this line seems a bit ambiguous to me

> If the key combination is changed to say, CmdOrCtrl+E

What is `CmdOrCtrl+E` meant for?

@jasonLaster do we need to document these somewhere?

- turn off column breakpoints `dbg.features.columnBreakpoints = false`
- disable `privacy.resistFingerprinting` in `about:config`

Garee:
@princiya When `privacy.resistFingerprinting` is `true`,  the keyboard event for `CmdOrCtrl+Shift+B` does not fire, but the event for `CmdOrCtrl+E` does. I temporarily changed the shortcut to `CmdOrCtrl+E` to test this.

princiya:
@Garee ah I see, thank you for clarifying.

I tested on mac, and this PR works 👍

jasonLaster:
Thanks @Garee!

PR #7592, opened 2018-12-23
Log points

bomsy:

### DIFF
--- a/assets/panel/debugger.properties
+++ b/assets/panel/debugger.properties
@@ -512,7 +512,7 @@ editor.conditionalPanel.placeholder=This breakpoint will pause when the expressi
 
 # LOCALIZATION NOTE (editor.conditionalPanel.logPoint.placeholder): Placeholder text for
 # input element inside ConditionalPanel component when a log point is set
-editor.conditionalPanel.logPoint.placeholder=This breakpoint will pause when the expression is true and will always log its result
+editor.conditionalPanel.logPoint.placeholder=This breakpoint will log the result of the expression
 
 # LOCALIZATION NOTE (editor.conditionalPanel.close): Tooltip text for
 # close button inside ConditionalPanel component
--- a/assets/panel/debugger.properties
+++ b/assets/panel/debugger.properties
@@ -490,12 +490,21 @@ editor.removeBreakpoint=Remove breakpoint
 
 # LOCALIZATION NOTE (editor.editBreakpoint): Editor gutter context menu item
 # for setting a breakpoint condition on a line.
-editor.editBreakpoint=Edit breakpoint
+editor.editBreakpoint=Edit condition
 
 # LOCALIZATION NOTE (editor.addConditionalBreakpoint): Editor gutter context
 # menu item for adding a breakpoint condition on a line.
-editor.addConditionalBreakpoint=Add conditional breakpoint
+editor.addConditionalBreakpoint=Add condition
 editor.addConditionalBreakpoint.accesskey=c
+
+# LOCALIZATION NOTE (editor.addLogPoint): Editor gutter context
+# menu item for adding a log point on a line.
+editor.addLogPoint=Add log
+editor.addLogPoint.accesskey=l
+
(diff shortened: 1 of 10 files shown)


assets/panel/debugger.properties
```diff
@@ -490,17 +490,30 @@ editor.removeBreakpoint=Remove breakpoint
 
 # LOCALIZATION NOTE (editor.editBreakpoint): Editor gutter context menu item
 # for setting a breakpoint condition on a line.
-editor.editBreakpoint=Edit breakpoint
+editor.editBreakpoint=Edit condition
```
flodolo: Quick note: to change strings like this, you need to use a new string ID (and updated the comment accordingly).
https://developer.mozilla.org/en-US/docs/Making_String_Changes

bomsy:
depends on https://github.com/devtools-html/debugger.html/pull/7593 landing

jasonLaster:
Thanks @bomsy!

assets/panel/debugger.properties
```diff
@@ -494,13 +494,30 @@ editor.editBreakpoint=Edit breakpoint
 
 # LOCALIZATION NOTE (editor.addConditionalBreakpoint): Editor gutter context
 # menu item for adding a breakpoint condition on a line.
-editor.addConditionalBreakpoint=Add conditional breakpoint
+editor.addConditionalBreakpoint=Add condition
```
flodolo: Sorry, but this shouldn't have passed review. String changes require new string IDs.

@jasonLaster Can we fix this ASAP? This just landed in m-c :-\

assets/panel/debugger.properties
```diff
@@ -494,13 +494,30 @@ editor.editBreakpoint=Edit breakpoint
...
 editor.addConditionalBreakpoint.accesskey=c
 
+# LOCALIZATION NOTE (editor.editBreakpoint): Editor gutter context menu item
```
flodolo: Wrong string ID reference

assets/panel/debugger.properties
```diff
@@ -494,13 +494,30 @@ editor.editBreakpoint=Edit breakpoint
...
+# for setting a breakpoint condition on a line.
+editor.editConditionalBreakpoint=Edit condition
+
+# LOCALIZATION NOTE (editor.addLogPoint): Editor gutter context
+# menu item for adding a log point on a line.
+editor.addLogPoint=Add log
```
flodolo: For the future, better to use X.label and X.accesskey as a couple (no need to change at this point)

assets/panel/debugger.properties
```diff
@@ -494,13 +494,30 @@ editor.editBreakpoint=Edit breakpoint
...
+editor.addLogPoint.accesskey=l
+
+# LOCALIZATION NOTE (editor.editLogpoint): Editor gutter context menu item
+# for setting a log point on a line.
```
flodolo: This would benefit from a comment explaining that "Edit" is a verb, and this doesn't mean "Log of edits".

flodolo:
One more question: initially this PR tried to change `editor.editBreakpoint`, and now it adds a new string. Is the old string still used somewhere? Strings should be removed when they become unused.

src/components/Editor/GutterMenu.js
```diff
@@ -84,6 +92,20 @@ export function gutterMenu({
     ...(breakpoint ? gutterItems.removeBreakpoint : gutterItems.addBreakpoint)
   };
 
+  const logPoint = {
+    accesskey: L10N.getStr("editor.addLogPoint.accesskey"),
```
flodolo: I'm lost in this piece of code: is the accesskey reused for different labels?

bomsy:
@flodolo i've opened https://github.com/devtools-html/debugger.html/pull/7646 to address most of the issues. thanks

rugk:
I've tested this in Nightly and IMHO it's strange that this thing is still labelled "condition"… I mean, it's no condition, it's rather a list of things to log…

image

bomsy:
@rugk Yeah we need to add context menu items for log points in the breakpoints list.

PR #7603, opened 2018-12-27
Shortcuts

### LINKED ISSUE #7498: StartPage shortcuts and their description font proportions are reversed on Windows vs Linux
(Sorry, for nitpicking but I always notice such very small inconsistency with half an eye, which makes the product look less polished than it could be)

On Linux the font of the shortcut is bigger than its description. Compare letter `P` (of `Ctrl P`) with letter `G` (of `Go to file`) on first line:

image

On Windows 10 it is reversed, that is, `P` is smaller than `G`:
image

Same for letter `F` in second line as part of shortcut and the word `Find`. 

In any way it should be consistent.

In my opinion the Linux version (Shortcuts a bit bigger than their description) is better. Windows should be the same. Not sure what it looks like on MacOS.

Though for consistency probably better to ping a Designer here, maybe @fvsch has some idea.

nagpalm7:
Fixes #7498 

### Summary of Changes

* Change font-family: Courier to font-family: var(--monospace-font-family) for keyboard shortcut
* Set line-height and text-size same for both labels and shortcuts to make it proportionate
* Change color for the keyboard shortcut to make it contrast with the label and look nice

### DIFF
--- a/src/components/WelcomeBox.css
+++ b/src/components/WelcomeBox.css
@@ -62,7 +62,7 @@
   font-family: var(--monospace-font-family);
   font-size: 14px;
   line-height: 18px;
-  color: var(--theme-body-color)
+  color: var(--theme-body-color);
 }
 
 .shortcutLabel {
--- a/src/components/WelcomeBox.css
+++ b/src/components/WelcomeBox.css
@@ -59,12 +59,17 @@
 .shortcutKey {
   text-align: right;
   padding-right: 10px;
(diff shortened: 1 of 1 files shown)


princiya:
👍 looks good to me

cc @AnshulMalik @bomsy

jasonLaster:
@fvsch let us know if this looks right to you. It'd be easy to do a follow up tweak

PR #7619, opened 2018-12-31
Bug 7607 no source msg

### LINKED ISSUE #7607: [SourcesTree] No sources while there are sources available
STR:

- Go to http://softwareishard.com/tests/debugger/issue3532/
- Open `index.js`
- Pretty Print 

AR:
Sources panel displays the sources but a new message appears saying `This page has no sources`

ER:
We shouldn't see `This page has no sources` since there are sources.

jainsneha23:
Fixes #7607 

### Summary of Changes

* Show source tree only for threads with valid name. Threads with "" are the threads for pretty printed sources, which need not be displayed in source tree.

### DIFF
--- a/src/components/PrimaryPanes/index.js
+++ b/src/components/PrimaryPanes/index.js
@@ -100,7 +100,10 @@ class PrimaryPanes extends Component<Props, State> {
       this.props.getWorkerDisplayName
     );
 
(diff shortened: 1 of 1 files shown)


jainsneha23:
This fix won't work. It also hides the webpack tree in the source tree pane. Hence closing the PR.

PR #7621, opened 2018-12-31
Add a way to stop an ongoing project search

### LINKED ISSUE #7559: Invent a way to stop full project search
Currently we don't have any option to stop currently ongoing project search.
It'd be really helpful if the search taking time and the user would want to kill it.

Garee:
## Summary

First, the search operation is assigned a cancellation function that can be used to stop the search. Next, when a new search is started, it is saved in state so that it can later be cancelled.

The ongoing search is cancelled when:

- A new project search is started.
- A user clicks the "x" button on the search input.

Fixes #7559

## Testing

- Verified that project searches work as before.
- Performed searches in quick succession and observed that only one search was in progress at a time.
- Tested cancellation via starting a new search and clicking on the "x" button.

## Other

- Do we want the "x" button to cancel the search, or should there be a new button?
- The cancellation isn't immediate -- it occurs after the current source has been searched. This isn't perfect but it's a big improvement.

### DIFF
--- a/src/actions/project-text-search.js
+++ b/src/actions/project-text-search.js
@@ -16,9 +16,14 @@ import { loadSourceText } from "./sources/loadSourceText";
 import { statusType } from "../reducers/project-text-search";
 
 import type { Action, ThunkArgs } from "./types";
+import type { SearchOperation } from "../reducers/project-text-search";
 
 export function addSearchQuery(query: string): Action {
   return { type: "ADD_QUERY", query };
+}
+
+export function addSearch(search: SearchOperation): Action {
+  return { type: "ADD_SEARCH", search };
 }
 
 export function clearSearchQuery(): Action {
@@ -52,8 +57,22 @@ export function closeProjectSearch(): Action {
   return { type: "CLOSE_PROJECT_SEARCH" };
 }
 
+export function stopOngoingSearch() {
+  return ({ dispatch, getState }: ThunkArgs) => {
+    const state = getState().projectTextSearch;
+    const search = state.search;
+    const status = state.status;
+    if (search && status !== statusType.done) {
+      search.cancel();
+    }
+  };
+}
+
 export function searchSources(query: string) {
-  return async ({ dispatch, getState }: ThunkArgs) => {
+  let cancelled = false;
+
+  const search = async ({ dispatch, getState }: ThunkArgs) => {
+    await dispatch(addSearch(search));
     await dispatch(clearSearchResults());
     await dispatch(addSearchQuery(query));
     dispatch(updateSearchStatus(statusType.fetching));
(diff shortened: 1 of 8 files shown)


src/reducers/project-text-search.js
```diff
@@ -90,7 +91,11 @@ function update(
   return state;
 }
 
-type OuterState = { projectTextSearch: ProjectTextSearchState };
+type OuterState = { projectTextSearch: Record<ProjectTextSearchState> };
```
AnshulMalik: I think it's not a record anymore.

src/actions/project-text-search.js
```diff
@@ -53,34 +58,37 @@ export function updateSearchStatus(status: string): Action {
...
-    await dispatch(addSearch(search));
+    stopOngoingSearch(getState());
+    await dispatch(addOngoingSearch(search));
     await dispatch(clearSearchResults());
     await dispatch(addSearchQuery(query));
     dispatch(updateSearchStatus(statusType.fetching));
     const validSources = getSourceList(getState()).filter(
       source => !hasPrettySource(getState(), source.id) && !isThirdParty(source)
     );
+    console.log("Searching...");
```
AnshulMalik: these are for debugging right?

Garee: Yes, I forgot to remove these.

src/reducers/project-text-search.js
Garee: I wasn't able to use `getState()` as an argument to the selectors without changing this.

```javascript
// in src/reducers/types.js
export type State = {
  ...
  projectTextSearch: Record<ProjectTextSearchState>,
  ...
};
```

AnshulMalik: That's strange

AnshulMalik: I think we need to remove `Record` from here https://github.com/devtools-html/debugger.html/blob/master/src/reducers/types.js#L31

Garee:
> Looks good to me, we'll probably need to set the status of the search as `done` or something when we cancel.

I added and set a new `statusType.cancelled` status.

src/actions/project-text-search.js
```diff
@@ -48,24 +57,52 @@ export function updateSearchStatus(status: string): Action {
...
 export function searchSources(query: string) {
-  return async ({ dispatch, getState }: ThunkArgs) => {
+  let cancelled = false;
+
+  const search = async ({ dispatch, getState }: ThunkArgs) => {
+    dispatch(stopOngoingSearch());
+    await dispatch(addOngoingSearch(search));
     await dispatch(clearSearchResults());
     await dispatch(addSearchQuery(query));
     dispatch(updateSearchStatus(statusType.fetching));
```
loganfsmyth: Seems like we'd also at least want to reset this status?

darkwing:
@AnshulMalik What ideas do you have for follow up tweaks?

darkwing:
Great work @Garee !  I love it!

PR #7623, opened 2018-12-31
Update sources and tabs when workers are created and destroyed.

bhackett1024:
### Summary of Changes

* When a worker is started, its sources are added to the sources pane.
* When a worker is destroyed, its sources are removed from the sources pane.
* When a worker is destroyed, any tabs associated with its sources are closed.

### Test Plan

Manually verified behavior is as expected.

### DIFF
--- a/src/actions/debuggee.js
+++ b/src/actions/debuggee.js
@@ -5,10 +5,19 @@
 // @flow
 
 import type { Action, ThunkArgs } from "./types";
+import { closeTabsForMissingThreads } from "./tabs";
+import { getMainThread } from "../reducers/pause";
 
 export function updateWorkers() {
-  return async function({ dispatch, client }: ThunkArgs) {
+  return async function({ dispatch, getState, client }: ThunkArgs) {
     const { workers } = await client.fetchWorkers();
     dispatch(({ type: "SET_WORKERS", workers }: Action));
+
+    closeTabsForMissingThreads(
+      dispatch,
+      getState,
+      getMainThread(getState()),
+      workers
+    );
   };
 }
--- a/src/components/PrimaryPanes/index.js
+++ b/src/components/PrimaryPanes/index.js
@@ -14,7 +14,8 @@ import {
   getRelativeSources,
   getActiveSearch,
   getSelectedPrimaryPaneTab,
-  getWorkerDisplayName
+  getWorkerDisplayName,
+  isValidThread
 } from "../../selectors";
 import { features, prefs } from "../../utils/prefs";
 import "./Sources.css";
@@ -38,7 +39,8 @@ type Props = {
   setPrimaryPaneTab: typeof actions.setPrimaryPaneTab,
   setActiveSearch: typeof actions.setActiveSearch,
   closeActiveSearch: typeof actions.closeActiveSearch,
-  getWorkerDisplayName: string => string
+  getWorkerDisplayName: string => string,
+  isValidThread: string => boolean
 };
 
 class PrimaryPanes extends Component<Props, State> {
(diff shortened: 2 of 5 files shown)


src/actions/debuggee.js
```diff
@@ -5,10 +5,19 @@
...
+import { getMainThread } from "../reducers/pause";
 
 export function updateWorkers() {
-  return async function({ dispatch, client }: ThunkArgs) {
+  return async function({ dispatch, getState, client }: ThunkArgs) {
     const { workers } = await client.fetchWorkers();
     dispatch(({ type: "SET_WORKERS", workers }: Action));
+
+    closeTabsForMissingThreads(
+      dispatch,
```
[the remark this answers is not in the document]
bhackett1024: Sure, I definitely want to improve this but I'm not sure how --- I couldn't find much documentation online or examples in debugger.html of how to either compose actions that are in separate components, or to trigger actions transitively in response to another change to the redux store.  Is there a way for closeTabsForMissingThreads() to have the same signature of other action functions?  Having it return an array of actions instead of doing the dispatching itself would still leave it with different usage from the usual action function.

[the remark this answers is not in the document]
AnshulMalik: How costly is `await client.fetchWorkers()` we can either pass it or re-fetch it in the `closeTabsForMissingThreads` action.
I can try to change the signature :)

src/actions/debuggee.js
```diff
@@ -5,10 +5,13 @@
...
 import type { Action, ThunkArgs } from "./types";
+import { closeTabsForMissingThreads } from "./tabs";
 
 export function updateWorkers() {
-  return async function({ dispatch, client }: ThunkArgs) {
+  return async function({ dispatch, getState, client }: ThunkArgs) {
     const { workers } = await client.fetchWorkers();
     dispatch(({ type: "SET_WORKERS", workers }: Action));
+
+    dispatch(closeTabsForMissingThreads(workers));
```
jasonLaster: I think the cleanest thing to do would be to remove the sources for the removed workers. We could do this like so:

`dispatch(({ type: "SET_WORKERS", workers, destroyed: destroyedWorkers }: Action));`

in reducers/sources

```js
removeSourcesForWorkers(state, removedWorkers)
```

There might be other state that is now out-of-date because of the worker list has changed. For instance the currentThread might need to be reset too. Also, as a general rule of thumb, we try to keep our actions fairly high-level and not too implementation specific. It is better if they are more general-purpose like "newSources" or "updateWorkers"...

src/components/PrimaryPanes/index.js
```diff
@@ -96,7 +98,9 @@ class PrimaryPanes extends Component<Props, State> {
 
   renderThreadSources() {
     const threads = sortBy(
-      Object.getOwnPropertyNames(this.props.sources),
+      Object.getOwnPropertyNames(this.props.sources).filter(
+        this.props.isValidThread
+      ),
```
jasonLaster: if we're removing the sources from the store we shouldn't be worried about rendering them here.

src/actions/debuggee.js
bhackett1024: I can try this, but I implemented the patch this way because there isn't a notion currently of being able to remove sources.  Adding this seems like it can affect quite a few places in the debugger, since sources are manipulated in so many places.  Allowing the sources to continue to exist, but removing them from the UI if they become obsolete (e.g. the worker they are associated with has closed) seemed a more robust design.

While closeTabsForMissingThreads is a pretty specific method, we do need to go through tab specific logic in order to get good behavior when removing sources --- if we just drop all tabs for removed sources then other parts of the editor pane can go out of sync (e.g. showing the contents of a tab that no longer exists).

src/components/PrimaryPanes/index.js
jasonLaster: yep - will be a nice followup, but could be difficult

PR #7631, opened 2019-01-01
removed scroll bar from individual source list and added to the source outline panel.

### LINKED ISSUE #7618: Sources Tree shows odd horizontal scrolling
Saw this today:

treescrolling

We should only have one horizontal scroll for the sources tree and it should display at the bottom of the pane.

ravi610:
…

Fixes #7618 

### Summary of Changes

* added scroll bar to source outline panel
* removed scroll bar from individual sources.

### DIFF
--- a/src/components/PrimaryPanes/Sources.css
+++ b/src/components/PrimaryPanes/Sources.css
@@ -58,25 +58,25 @@
   margin-left: 5px;
 }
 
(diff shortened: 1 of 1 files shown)


darkwing:
Awesome!  Thank you so much @ravi610 !

PR #7638, opened 2019-01-02
Fix linting issue in visibleBreakpoints

darkwing:
Fixing circle


AnshulMalik:
Sorry, tests were failing so I just pushed here instead.

PR #7640, opened 2019-01-02
Update dwarf-to-json library

yurydelendik:
Fixes problem: sometimes Firefox cannot consume and convert .wasm to .map.


packages/devtools-source-map/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "devtools-source-map",
-  "version": "0.16.0",
+  "version": "0.16.1",
```
loganfsmyth: I think the failures are because you bumped this, but not the version in the repo-root `package.json` FYI.

Probably easiest to just leave both as-is?

PR #7644, opened 2019-01-03
Initialize a breakpoint as loading

### LINKED ISSUE #6771: [Breakpoints] Use `loading` property

Art-Vanderlay:
Fixes #6771

Latest attempt to initialize loading as true

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -1,5 +1,3 @@
----
----
 # Contributing to debugger.html
 
 :+1::tada: First off, thanks for taking the time to contribute! :tada::+1:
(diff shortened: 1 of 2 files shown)


darkwing:
Presumably we'd need to set this to `false` somewhere, which may be why tests are failing.

darkwing:
@AnshulMalik or @Bomsy, could you take a look at this?  I've tried this in the past and ran into the same issues.

AnshulMalik:
This is more complex than I thought it would be, I'll try to look into it tonight.

darkwing:
Any progress here @AnshulMalik ?

AnshulMalik:
Yeah, there seems to be some issue with breakpoints not persisting aka pending breakpoints.
Not sure why this is happening though. Will try again

PR #7646, opened 2019-01-03
fix issues found with the string in debugger.properties

bomsy:
Fixes Log point l10n issues

### DIFF
--- a/assets/panel/debugger.properties
+++ b/assets/panel/debugger.properties
@@ -488,26 +488,22 @@ editor.enableBreakpoint=Enable breakpoint
 # for removing a breakpoint on a line.
 editor.removeBreakpoint=Remove breakpoint
 
-# LOCALIZATION NOTE (editor.editBreakpoint): Editor gutter context menu item
-# for setting a breakpoint condition on a line.
-editor.editBreakpoint=Edit breakpoint
-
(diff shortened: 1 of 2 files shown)


assets/panel/debugger.properties
```diff
@@ -488,26 +488,22 @@ editor.enableBreakpoint=Enable breakpoint
...
-# for setting a breakpoint condition on a line.
-editor.editBreakpoint=Edit breakpoint
-
-# LOCALIZATION NOTE (editor.addConditionalBreakpoint): Editor gutter context
+# LOCALIZATION NOTE (editor.addConditionBreakpoint): Editor gutter context
 # menu item for adding a breakpoint condition on a line.
-editor.addConditionalBreakpoint=Add condition
-editor.addConditionalBreakpoint.accesskey=c
+editor.addConditionBreakpoint=Add condition
+editor.addConditionBreakpoint.accesskey=c
```
flodolo: Open question from the other PR: is the accesskey shared across different labels?

bomsy: No it is not at the moment.

flodolo: Thanks, I couldn't figure it out from the code. In case, sharing accesskeys across different labels is a no-go ;-)

PR #7649, opened 2019-01-03
Fix sources tree not displaying when custom root

darkwing:
The PR to add worker sources to the tree (https://github.com/devtools-html/debugger.html/commit/efcda91d682b3e3818ed53e2f430ae70f1a7ac60) sort of broke the sources tree -> custom root functionality.

When a custom root is set, the tree portion of the panel appears empty.  It's not actually empty but the newly introduced `<div>{this.renderThreadSources()}</div>` wrapper displays at 0 height.

This PR fixes that issue.

### DIFF
--- a/src/components/PrimaryPanes/Sources.css
+++ b/src/components/PrimaryPanes/Sources.css
@@ -208,6 +208,10 @@
   overflow: auto;
 }
 
+.source-outline-panel.has-root > div {
+  height: 100%;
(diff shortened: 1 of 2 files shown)


jasonLaster:
@darkwing can you rebase and confirm this is still necessary?

darkwing:
Rebased, still needed.

PR #7651, opened 2019-01-03
7536/improve source pane padding

### LINKED ISSUE #7536: Wrong vertical padding in Sources pane

Garee:
Specify a 4px top/bottom padding for the sources list tree element and remove the top margin from its first child. This fixes the gap between the list scrollbar and the top of the container when it overflows.

Remove the margin from the outline filter and give it the same 4px top padding to maintain consistency across tabs.

Updated the expected test results as the expected styles have changed.

Fixes #7536.

### DIFF
--- a/src/components/PrimaryPanes/Sources.css
+++ b/src/components/PrimaryPanes/Sources.css
@@ -74,8 +74,8 @@
   display: flex;
 }
 
-.sources-list .managed-tree .tree-node:first-child {
-  margin-top: 4px;
+.sources-list .managed-tree .tree {
+  overflow: visible;
 }
 
 .sources-list .managed-tree .tree .node {
--- a/src/components/PrimaryPanes/Sources.css
+++ b/src/components/PrimaryPanes/Sources.css
@@ -62,6 +62,7 @@
   display: flex;
   flex: 1;
   height: 100%;
+  padding-bottom: 4px;
 }
 
 .sources-list {
--- a/src/components/PrimaryPanes/Sources.css
+++ b/src/components/PrimaryPanes/Sources.css
@@ -77,6 +77,7 @@
 
 .sources-list .managed-tree .tree {
   overflow: visible;
+  padding: 4px 0;
 }
 
 .sources-list .managed-tree .tree .node {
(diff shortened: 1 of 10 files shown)


Garee:
@nchevobbe - This is a continuation of [this pull request](https://github.com/devtools-html/debugger.html/pull/7582).

src/components/PrimaryPanes/Sources.css
```diff
@@ -62,6 +62,7 @@
   display: flex;
   flex: 1;
   height: 100%;
+  padding-bottom: 4px;
```
nchevobbe: i'm not sure why we have this one in addition to the bottom padding we have on the tree as well.
Here it looks like we have a 4px padding at the top, then a 8px one at the bottom, which is a bit odd.

Garee: I removed this padding. Thanks.

Garee:
> Since we have padding on the tree, when there are selected items in it, they don't stick to the border of the panel, looking a bit off:

@nchevobbe What should I change it to? I thought this was correct based on the comments by @fvsch in #7536.

fvsch:
> Since we have padding on the tree, when there are selected items in it, they don't stick to the border of the panel, looking a bit off.

That's an expected result. Not ideal, but the padding has other benefits, so for now it's a trade-off we want. (Can't promise that it'll never ever change in a design review though ^^)

src/components/PrimaryPanes/Sources.css
```diff
@@ -74,8 +74,9 @@
   display: flex;
 }
 
-.sources-list .managed-tree .tree-node:first-child {
-  margin-top: 4px;
+.sources-list .managed-tree .tree {
+  overflow: visible;
```
nchevobbe: do we need this? It seems to work fine without it.

Garee: Nope! I've removed it. Thanks.

nchevobbe:
Thanks @Garee !

PR #7662, opened 2019-01-05
Change find from border to highlight

### LINKED ISSUE #6744: Search highlight creates multiple blocks for same search term
STR:
- https://bl.ocks.org/ColinEberhardt/raw/6ceb7ca74aabac9c8534d7120d31b382/
- Open js file
- Search for `wasm.`

AR: Search highlights `wasm` and `.` as separate boxes. Gets worse for `wasm.s`, creating 3 boxes.

screenshot 2018-08-03 13 17 17

darkwing:
Fixes #6744

Changes the editor search from a border to a background highlight:

whitehighlight

darkhighlight

@violasong @mcroud Dark mode is obviously an issue because we have green highlighted text.

Apologies or jumping on this before https://github.com/devtools-html/ux/issues/14 was solved but I've gotten antsy :p


jasonLaster:
oh nice, I hadn't thought about using a background color

1. we should get some thoughts on background-color. e.g. green or what not
2. if we go w/ background color we should distinguish between all matches and selected match
3. i think we can fix this bug with CSS... e.g. `.cm-highlight + .cm-highlight { border-left: none; }` would be half of the fix... 
4. we might want to fix CM's search [tokenizer](https://searchfox.org/mozilla-central/source/devtools/client/debugger/new/src/utils/editor/source-search.js#58-83) CC @bomsy

darkwing:
Closing; we can consider this a good starting spot for someone to work from but we should work for https://github.com/devtools-html/ux/issues/14

PR #7664, opened 2019-01-05
Improve the expand/collapse all behaviour of the Sources tree

### LINKED ISSUE #7663: Inconsistent expand/collapse all behaviour for Sources tree items
Alt-clicking a directory in the Sources tree will expand it and all of its children. However, Alt-clicking again does not collapse the children.

In addition, I also noticed that `setExpanded` within `ManagedTree.js` is invoked twice when you click on a source tree item. Once for `onClick` in `SourcesTreeItem.js` and another for an `onClick` event in `tree.js`. I don't think both of these are required as the tree component handles the expansion.

Garee:
First, remove the expansion code in the `SourcesTreeItem` `onClick` event as the expansion is already handled in the `onClick` event of the tree component.

Next, pass the `altKey` status from the `tree onClick` event to the `ManagedTree onCollapse` function so that we know whether to collapse the children. This allows us to add the ability to collapse all children too.

Fixes #7663

### DIFF
--- a/packages/devtools-components/src/tree.js
+++ b/packages/devtools-components/src/tree.js
@@ -798,7 +798,7 @@ class Tree extends Component {
           // it should be scrolled into view.
           this._focus(item, { preventAutoScroll: true });
           if (this.props.isExpanded(item)) {
-            this.props.onCollapse(item);
+            this.props.onCollapse(item, e.altKey);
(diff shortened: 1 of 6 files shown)


darkwing:
I love this!  Great work @Garee !

PR #7669, opened 2019-01-06
strict flow and remove some unused code in project search

AnshulMalik:
`sources` were unused so removed them.

And updated some types to be more strict.


src/components/ProjectSearch.js
```diff
@@ -84,13 +82,18 @@ function sanitizeQuery(query: string): string {
...
+};
+
 export class ProjectSearch extends Component<Props, State> {
-  focusedItem: ?{
-    setExpanded?: any,
-    file?: any,
-    expanded?: any,
-    match?: Match
-  };
+  focusedItem: ?FileItem | MatchItem;
```
loganfsmyth: I _think_ this ends up being

```
(?FileItem) | MatchItem
```
but is it intended to be?
```
?(FileItem | MatchItem)
```
I see we do a `if (!this.focusedItem` check elsewhere in the patch.

src/components/ProjectSearch.js
```diff
@@ -84,13 +82,18 @@ function sanitizeQuery(query: string): string {
   return query.replace(/\\$/, "");
 }
 
+type FileItem = {
+  setExpanded: (Result, boolean) => any,
```
loganfsmyth: Better to go with `mixed` instead of `any` for things like this so attempting to use the return value would require refinement.

src/components/ProjectSearch.js
AnshulMalik: Yeah, that's what I wanted to do, but this did the trick :P

PR #7688, opened 2019-01-08
fetch symbols before mapFrames in same file

### LINKED ISSUE #7535: [symbols] other actions should be able to wait on symbols completing
We our getSymbols action is currently doing some work that should be done in better places.

```diff

diff --git a/src/actions/ast.js b/src/actions/ast.js
index 0efec8708..665972e57 100644
--- a/src/actions/ast.js
+++ b/src/actions/ast.js
@@ -71,10 +71,10 @@ export function setSymbols(sourceId: SourceId) {

-    if (isPaused(getState())) {
-      await dispatch(fetchExtra());
-      await dispatch(mapFrames());
-    }
     await dispatch(setPausePoints(sourceId));
     await dispatch(setSourceMetaData(sourceId));

diff --git a/src/actions/pause/paused.js b/src/actions/pause/paused.js
index 5f91f68c1..1f36d776d 100644
--- a/src/actions/pause/paused.js
+++ b/src/actions/pause/paused.js
@@ -68,6 +68,7 @@ export function paused(pauseInfo: Pause) {
       dispatch(removeBreakpoint(hiddenBreakpointLocation));
     }

+    await dispatch(setSymbols());
     await dispatch(mapFrames());

     const selectedFrame = getSelectedFrame(getState());
@@ -77,6 +78,7 @@ export function paused(pauseInfo: Pause) {

     dispatch(togglePaneCollapse("end", false));
     await dispatch(fetchScopes());
+    await dispatch(fetchExtra());

     // Run after fetching scoping data so that it may make use of the sourcemap
     // expression mappings for local variables.
```

With this approach the paused thunk can wait for symbols and then expect mapFrames to succeed and later fetch extra information.

For this strategy to work we should borrow ideas from [loadSourceText][ls] where we save the pending thunk in a map so that other actions can await the initial load as well. We could do something similar here.

[ls]: https://github.com/devtools-html/debugger.html/blob/master/src/actions/sources/loadSourceText.js#L52-L55

nagpalm7:
Fixes #7535 

* dispatch setSymbols before mapFrames and extra things can be fetched after map frames are successfully fetched


src/actions/pause/paused.js
```diff
@@ -69,6 +71,7 @@ export function paused(pauseInfo: Pause) {
       dispatch(removeBreakpoint(hiddenBreakpointLocation));
     }
 
+    await dispatch(setSymbols());
```
jasonLaster: i think one concern here is that we could already be getting the symbols for this source and we wouldn't want to try to get it a second time.

This is why we have the logic in loadSourceText to check for this... and some unit tests to also check for it.

nagpalm7: Okay i understood the logic in `loadSourceText`.  I need to change `setSymbols` function in ast.js and perform fetch as an `loadSourceText` correct me if i am wrong. @jasonLaster

jasonLaster: That's right @nagpalm7. We'll need something like this

```js
    if (requests.has(id)) {
      return requests.get(id);
    }


...

    // signal that the action is finished
    deferred.resolve();
    requests.delete(id);
```

nagpalm7:
>/home/mustafa/Documents/OpenSource/debugger.html/node_modules/selenium-webdriver/lib/promise.js:2626
        throw error;

@jasonLaster server crashes when I launch firefox and visit a site for debugging.

src/actions/ast.js
```diff
@@ -53,17 +56,49 @@ export function setSourceMetaData(sourceId: SourceId) {
...
   return async ({ dispatch, getState, sourceMaps }: ThunkArgs) => {
-    const source = getSourceFromId(getState(), sourceId);
+    const id = sourceId;
+    const source = getSourceFromId(getState(), id);
+    // Fetch the source text only once.
+    if (requests.has(id)) {
+      return requests.get(id);
+    }
+
+    if (isLoaded(source)) {
```
jasonLaster: hmm, i think we want to check if there are symbols, and if so return early

nagpalm7: @jasonLaster after checking id or instead of id we need to check for symbols and is there any function predefined

jasonLaster: i don't follow

nagpalm7: Is there any function defined somewhere which returns whether symbols have fetched or not?

jasonLaster: i think `hasSymbols`

darkwing:
Hello @nagpalm7.  How's it going?

darkwing:
Excellent, thank you @nagpalm7 !

darkwing:
Are you asking about PausePoints?  I don't see a pauses.js file....

nagpalm7:
@darkwing  src/actions/pause/paused.js

```

+++ b/src/actions/pause/paused.js
@@ -68,6 +68,7 @@ export function paused(pauseInfo: Pause) {
     }

+    await dispatch(setSymbols());
     await dispatch(mapFrames());

     const selectedFrame = getSelectedFrame(getState());
```

jasonLaster:
@nagpalm7 perhaps `topFrame.location.sourceId`?

nagpalm7:
> @nagpalm7 perhaps `topFrame.location.sourceId`?

done @jasonLaster Can you please review the path and also guide me on how to fix test fails !

darkwing:
Hello @nagpalm7!  Thank your time and work on this -- it helped us to realize this is a tougher issue than we hoped and we need to think more about our internal structure.  Feel free to keep working on this but we'll close the PR for now; if you get it in the state with tests passing, please reopen!  Thank you!

Issue #7671, opened 2019-01-07
[breakpoints] shift+click should add a disabled breakpoint


jasonLaster:
When the user shift+clicks in the gutter, the debugger should create a disabled breakpoint. Currently it creates an enabled breakpoint.

This is a good starting off spot

```diff
diff --git a/src/actions/breakpoints/index.js b/src/actions/breakpoints/index.js
index 830b39a64..4aa386967 100644
--- a/src/actions/breakpoints/index.js
+++ b/src/actions/breakpoints/index.js
@@ -385,7 +385,11 @@ export function toggleBreakpointsAtLine(line: number, column?: number) {
   };
 }

-export function addOrToggleDisabledBreakpoint(line: number, column?: number) {
+export function addOrToggleDisabledBreakpoint(
+  line: number,
+  column?: number,
+  shouldDisable
+) {
   return ({ dispatch, getState, client, sourceMaps }: ThunkArgs) => {
     const selectedSource = getSelectedSource(getState());

@@ -411,7 +415,8 @@ export function addOrToggleDisabledBreakpoint(line: number, column?: number) {
         sourceId: selectedSource.id,
         sourceUrl: selectedSource.url,
         line: line,
-        column: column
+        column: column,
+        enabled: !shouldDisable
       })
     );
   };
diff --git a/src/components/Editor/index.js b/src/components/Editor/index.js
index 7227c11d3..e0cf932a4 100644
--- a/src/components/Editor/index.js
+++ b/src/components/Editor/index.js
@@ -404,7 +404,7 @@ class Editor extends PureComponent<Props, State> {
     }

     if (ev.shiftKey) {
-      return addOrToggleDisabledBreakpoint(sourceLine);
+      return addOrToggleDisabledBreakpoint(sourceLine, ev.shiftKey);
     }
:
```

baronKilljoy:
I would like to work on this. should I alter the existing function or override?

AnshulMalik:
@baronKilljoy altering the existing function should do :)

baronKilljoy:
My idea is to use the preexisting disabled property. However, it kepts being set to false. Am I doing something obviously wrong.
![change1](https://user-images.githubusercontent.com/36493475/50823867-028aef00-1303-11e9-8fb5-38093f747b74.png)
![change2](https://user-images.githubusercontent.com/36493475/50823875-06b70c80-1303-11e9-81e8-9c6fe287e3a2.png)

baronKilljoy:
[...]

PR #7694, opened 2019-01-08
[breakpoints] shift+click should add a disabled breakpoint #7671

baronKilljoy:
Fixes #<issue number>

Here's the Pull Request Doc

### Summary of Changes

* change 1
* change 2

### Test Plan

Example test plan:

Feel free to improve it!

### DIFF
--- a/src/actions/breakpoints/addBreakpoint.js
+++ b/src/actions/breakpoints/addBreakpoint.js
@@ -29,7 +29,13 @@ import type { SourceLocation } from "../../types";
 import type { ThunkArgs } from "../types";
 import type { addBreakpointOptions } from "./";
 
-async function addBreakpointPromise(getState, client, sourceMaps, breakpoint, shiftKey = false) {
+async function addBreakpointPromise(
+  getState,
+  client,
+  sourceMaps,
+  breakpoint,
+  shiftKey = false
+) {
   const state = getState();
   const source = getSource(state, breakpoint.location.sourceId);
 
@@ -90,9 +96,9 @@ async function addBreakpointPromise(getState, client, sourceMaps, breakpoint, sh
     text,
     originalText
   };
-	if(!shiftKey){
-		newBreakpoint.disabled=false;
-	}
+  if (!shiftKey) {
+    newBreakpoint.disabled = false;
+  }
   assertBreakpoint(newBreakpoint);
 
   if (breakpoint.disabled && shiftKey) {
@@ -209,7 +215,13 @@ export function addDisabledBreakpoint(
     return dispatch({
       type: "ADD_BREAKPOINT",
       breakpoint,
-      [PROMISE]: addBreakpointPromise(getState, client, sourceMaps, breakpoint,true)
+      [PROMISE]: addBreakpointPromise(
+        getState,
+        client,
+        sourceMaps,
+        breakpoint,
+        true
+      )
     });
   };
 }
--- a/src/actions/breakpoints/addBreakpoint.js
+++ b/src/actions/breakpoints/addBreakpoint.js
@@ -92,6 +92,9 @@ async function addBreakpointPromise(getState, client, sourceMaps, breakpoint) {
   };
 
   assertBreakpoint(newBreakpoint);
+  if(breakpoint.disabled){
+	  await client.removeBreakpoint(newGeneratedLocation);
+  }
 
   const previousLocation = locationMoved(location, newLocation)
     ? location
--- a/src/actions/breakpoints/addBreakpoint.js
+++ b/src/actions/breakpoints/addBreakpoint.js
@@ -80,7 +80,7 @@ async function addBreakpointPromise(getState, client, sourceMaps, breakpoint) {
 
   const newBreakpoint = {
     id,
-    disabled: false,
+    disabled: breakpoint.disabled,
     hidden: breakpoint.hidden,
     loading: false,
     condition: breakpoint.condition,
@@ -166,6 +166,39 @@ export function addBreakpoint(
     }
 
     const breakpoint = createBreakpoint(location, { condition, hidden });
+
+    return dispatch({
+      type: "ADD_BREAKPOINT",
+      breakpoint,
+      [PROMISE]: addBreakpointPromise(getState, client, sourceMaps, breakpoint)
+    });
+  };
+}
+/**
+ * Add a new disabled breakpoint
+ *
+ * @memberof actions/breakpoints
+ * @static
+ * @param {String} $1.condition Conditional breakpoint condition value
+ * @param {Boolean} $1.disabled Disable value for breakpoint value
+ */
+
+export function addDisabledBreakpoint(
+  location: SourceLocation,
+  { condition, hidden }: addBreakpointOptions = {}
+) {
+  return async ({ dispatch, getState, sourceMaps, client }: ThunkArgs) => {
+    recordEvent("add_breakpoint");
+
+    if (features.columnBreakpoints && location.column === undefined) {
+      location = getFirstPausePointLocation(getState(), location);
+    }
+
+    const breakpoint = createBreakpoint(location, {
+      condition,
+      hidden,
+      disabled: true
+    });
 
     return dispatch({
       type: "ADD_BREAKPOINT",
--- a/src/actions/breakpoints/addBreakpoint.js
+++ b/src/actions/breakpoints/addBreakpoint.js
@@ -29,7 +29,7 @@ import type { SourceLocation } from "../../types";
 import type { ThunkArgs } from "../types";
 import type { addBreakpointOptions } from "./";
 
-async function addBreakpointPromise(getState, client, sourceMaps, breakpoint) {
+async function addBreakpointPromise(getState, client, sourceMaps, breakpoint, shiftKey = false) {
   const state = getState();
   const source = getSource(state, breakpoint.location.sourceId);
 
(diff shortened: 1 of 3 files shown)


notjaril:
This PR has a XHR-related commit that addresses a different issue (#7700). Could you make a separate branch and PR for that one so we can track the issues separately?

baronKilljoy:
it looks like breakpoints have to be initialized then disabled. They cant be created with a state of disabled.

jasonLaster:
@baronKilljoy makes sense

baronKilljoy:
I am at a loss for what the next step is. The promise needs to be resolved either in the function or in the if (ev.shiftKey) statement. Any thoughts?

notjaril:
That's correct - passing `disabled: true` when creating a new breakpoint doesn't actually create a disabled breakpoint. `addBreakpointPromise` sets a new breakpoint on the client regardless of the disabled value. That's why it still pauses on the shift+click-ed breakpoint, even though the redux store might say it's disabled and it shouldn't.

For next steps I'd suggest looking at `/src/client/firefox/commands.js` to understand how breakpoints are set in the client. After that, add`if (breakpoint.disabled) { ... }` logic after the breakpoint is set in `addBreakpointPromise` to handle cases where it's passed `disabled: true`. 

Might get tricky since syncing breakpoint locations always gives us trouble, but this should be enough to get started! :)

baronKilljoy:
What behavior or attribute then tells the client "skip me"

notjaril:
My understanding is the client doesn't retain whether a breakpoint is disabled or not. It only either 1) sets breakpoints or 2) removes breakpoints to pause on. This is in `/src/client/firefox/commands.js`.

When a breakpoint (that's registered in the redux store and displayed in the Breakpoints pane) is skipped, it's because that breakpoint has been removed from the separate list of breakpoints set for the client to look at and pause on. What looks like toggling enable/disable breakpoints in the UI actually adds/removes breakpoints on the client.

baronKilljoy:
so if its created disabled we need to add it disabled, and not send it to the client?

notjaril:
Something along those lines! 👍

baronKilljoy:
@jasonLaster I am looking trying to see at what point its sent to the client, and if i can stop it.

baronKilljoy:
@jasonLaster @jaril  After I said that it occurred to me I could be over thinking it. However fixing this uncovered/creates a new bug where newBreakpoint.disabled must be false or the checkbox breaks.

src/actions/breakpoints/addBreakpoint.js
```diff
@@ -90,9 +96,15 @@ async function addBreakpointPromise(getState, client, sourceMaps, breakpoint) {
     text,
     originalText
   };
-
+  if (!shiftKey) {
```
jaril: I think `assertBreakpoint(newBreakpoint)` could be moved to be immediately after `const newBreakpoint = {...}` since it doesn't check `newBreakpoint.disabled` at all. The way it is right now makes me uncertain whether the `assertBreakpoint(newBreakpoint)` relies on `newBreakpoint.disabled`.

src/actions/breakpoints/addBreakpoint.js
```diff
@@ -174,3 +186,42 @@ export function addBreakpoint(
...
+
+    return dispatch({
+      type: "ADD_BREAKPOINT",
+      breakpoint,
+      [PROMISE]: addBreakpointPromise(
+        getState,
+        client,
+        sourceMaps,
+        breakpoint,
+        true
```
jaril: Are there any cases in which `disabled: true` and `shiftKey: false`, or vice versa? 

This seems to be the only place where `disabled: true`, and if that's the case, the `shiftKey` value passed here is redundant as it can just be inferred from `breakpoint.disabled`.

src/actions/breakpoints/addBreakpoint.js
```diff
@@ -29,7 +29,13 @@ import type { SourceLocation } from "../../types";
 import type { ThunkArgs } from "../types";
 import type { addBreakpointOptions } from "./";
 
-async function addBreakpointPromise(getState, client, sourceMaps, breakpoint) {
+async function addBreakpointPromise(
+  getState,
+  client,
+  sourceMaps,
+  breakpoint,
+  shiftKey = false
```
jaril: See earlier comment - `shiftKey` parameter might be redundant if it's always the same value as `breakpoint.disabled`

src/actions/breakpoints/addBreakpoint.js
```diff
@@ -90,9 +96,15 @@ async function addBreakpointPromise(getState, client, sourceMaps, breakpoint) {
     text,
     originalText
   };
-
+  if (!shiftKey) {
+    newBreakpoint.disabled = false;
+  }
   assertBreakpoint(newBreakpoint);
 
+  if (breakpoint.disabled && shiftKey) {
```
jaril: Are there cases for line 104 in which the values of `breakpoint.disabled` and `shiftKey` would not be the same? If not, it would be clearer if this were put together with the previous `if(!shiftKey)` to form an if/else statement

jaril:
Took this for a spin and it seems like it's now doing what it's supposed to. Left comments for lines where I thought the code could be simplified and where there were could be redundancies.

This branch currently has an unrelated commit 844708c which I think was intended to address a separate issue #7700. That issue has already been adressed by #7766 so you could take that line out from `XHRBreakpoints.js`.

Lastly, same thoughts as @jasonLaster - could you rebase this?

src/actions/breakpoints/addBreakpoint.js
baronKilljoy: add breakpoint is also used with the checkbox in the secondary pane. If someone is enabling the breakpoint by checking the box newBreakpoint needs to be set before assertbreakpoint is called. I didn't add the shiftKey to fix this bug, I added it to fix the bug it created.

src/actions/breakpoints/addBreakpoint.js
baronKilljoy: I didnt know what the behavior should be if someone shift clicks an disabled breakpoint (or why they would) so did it more as a safe guard from malicious users.

jaril: `addOrToggleDisabledBreakpoint` already handles that case (shift+click on disabled breakpoint) by calling `toggleDisabledBreakpoint`, so by the time you're in `addBreakpointPromise` it's not necessary.

src/actions/breakpoints/addBreakpoint.js
jaril: Oh now I understand, thanks for explaining it. Did not take `toggleBreakpoint` into consideration earlier. In that case this order does make sense 👍

baronKilljoy:
i backed out the change in the unrelated fix 

Backing out change #7671 #1 
@jaril

src/actions/breakpoints/addBreakpoint.js
baronKilljoy: ok removed the breakpoint.disabled check.

jaril:
The unrelated line of code from commit 844708c seems like it's still there. Specifically, it's line `198` in `src/components/SecondaryPanes/XHRBreakpoints.js` that needs to be removed:

```javascript
        onClick={() => this.handleCheckbox(index)}
```

After that, I think you just need to rebase and then we're good to go!

darkwing:
Thank you for all the work you've put into this @baronKilljoy!

My immediate concern is that this second function for adding a breakpoint creates a lot of duplication, which will cause us potentially a lot of maintenance issues.

I started fresh, thinking "What if we dispatch two actions?  One to create the breakpoint and another to disable it?"

```diff
diff --git a/src/components/Editor/index.js b/src/components/Editor/index.js
index 1d43e1ced..f992fe5dc 100644
--- a/src/components/Editor/index.js
+++ b/src/components/Editor/index.js
@@ -433,7 +433,7 @@ class Editor extends PureComponent<Props, State> {
       return continueToHere(sourceLine);
     }
 
-    return addBreakpointAtLine(sourceLine);
+    return addBreakpointAtLine(sourceLine, !!ev.shiftKey);
   };
 
   onGutterContextMenu = (event: MouseEvent) => {


diff --git a/src/actions/breakpoints/index.js b/src/actions/breakpoints/index.js
index 65f86eb9f..c5989e865 100644
--- a/src/actions/breakpoints/index.js
+++ b/src/actions/breakpoints/index.js
@@ -330,8 +330,11 @@ export function toggleBreakpointAtLine(line: number) {
   };
 }
 
-export function addBreakpointAtLine(line: number) {
-  return ({ dispatch, getState, client, sourceMaps }: ThunkArgs) => {
+export function addBreakpointAtLine(
+  line: number,
+  shouldDisable: boolean = false
+) {
+  return async ({ dispatch, getState, client, sourceMaps }: ThunkArgs) => {
     const state = getState();
     const source = getSelectedSource(state);
 
@@ -339,14 +342,24 @@ export function addBreakpointAtLine(line: number) {
       return;
     }
 
-    return dispatch(
-      addBreakpoint({
-        sourceId: source.id,
-        sourceUrl: source.url,
-        column: undefined,
-        line
-      })
+    const { breakpoint } = await dispatch(
+      addBreakpoint(
+        {
+          sourceId: source.id,
+          sourceUrl: source.url,
+          column: undefined,
+          line
+        }
+      )
     );
+
+    if (shouldDisable) {
+      breakpoint = await dispatch(disableBreakpoint(breakpoint));
+    }
+
+    return breakpoint;
   };
 }
```

That works but the breakpoint flashes briefly and then disables, which isn't eye-pleasing and probably kicks off some unnecessary re-rendering.

Modifying the promise function seems to work great.  I put the result in a new PR we can work on:

https://github.com/devtools-html/debugger.html/pull/7823

PR #7695, opened 2019-01-09
Fix unhelpful error in top-level await malformed expression.

nchevobbe:
In mapExpression, if we fail to get an AST
from the expression this means the expression
has a syntax error in it.
In order to get a meaningful error when evaluating
top-level await expression, we still need to
wrap the expression in an async iife.
This required some changes in mapExpression in order
to not bail completely if parsing the expression
threw.
We also take this as an opportunity to only parse
the expression once for the all the transforms.

### DIFF
--- a/src/workers/parser/mapExpression.js
+++ b/src/workers/parser/mapExpression.js
@@ -4,6 +4,7 @@
 
 // @flow
 
+import { parseConsoleScript } from "./utils/ast";
 import mapOriginalExpression from "./mapOriginalExpression";
 import mapExpressionBindings from "./mapBindings";
 import mapTopLevelAwait from "./mapAwaitExpression";
@@ -30,22 +31,23 @@ export default function mapExpression(
     originalExpression: false
   };
 
+  const ast = parseConsoleScript(expression);
   try {
-    if (mappings) {
+    if (mappings && ast) {
       const beforeOriginalExpression = expression;
-      expression = mapOriginalExpression(expression, mappings);
+      expression = mapOriginalExpression(expression, ast, mappings);
       mapped.originalExpression = beforeOriginalExpression !== expression;
     }
 
-    if (shouldMapBindings) {
+    if (shouldMapBindings && ast) {
       const beforeBindings = expression;
-      expression = mapExpressionBindings(expression, bindings);
+      expression = mapExpressionBindings(expression, ast, bindings);
       mapped.bindings = beforeBindings !== expression;
     }
 
     if (shouldMapAwait) {
       const beforeAwait = expression;
-      expression = mapTopLevelAwait(expression);
+      expression = mapTopLevelAwait(expression, ast);
       mapped.await = beforeAwait !== expression;
     }
   } catch (e) {
--- a/src/workers/parser/tests/mapExpression.spec.js
+++ b/src/workers/parser/tests/mapExpression.spec.js
@@ -12,7 +12,8 @@ function test({
   bindings,
   mappings,
   shouldMapExpression,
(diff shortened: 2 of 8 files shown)


src/workers/parser/mapExpression.js
```diff
@@ -30,22 +31,36 @@ export default function mapExpression(
...
+      return ast;
+    }
+
+    try {
+      ast = parseConsoleScript(expression);
+    } catch (e) {
+      // no need to do anything
+    }
+    return ast;
+  };
```
jasonLaster: Why is this needed? Are we trying to prevent unnecessary parsing?

nchevobbe: Yes. we may end up doing 3 mappings, or none, so this should prevent doing multiple ones. Also, we don't want to bail out on the entire mapping since we do want to still map the await expression if it has a syntax error

src/workers/parser/mapExpression.js
```diff
@@ -30,22 +32,23 @@ export default function mapExpression(
     originalExpression: false
   };
 
+  const getAst = memoize(parseConsoleScript).bind(null, expression);
```
jasonLaster: do you need? `memoize(parseConsoleScript).bind(null, expression);` 

I feel like `getAst(expression)` would work just as well

src/workers/parser/tests/mapExpression.spec.js
```diff
@@ -294,6 +301,20 @@ describe("mapExpression", () => {
         originalExpression: false
       }
     },
+    {
+      name: "await (with SyntaxError)",
+      expression: "await new Promise())",
+      newExpression: formatAwait("await new Promise())"),
+      parseExpression: false,
```
jasonLaster: what's this doing?

src/workers/parser/tests/mapOriginalExpression.spec.js
```diff
@@ -3,31 +3,35 @@
...
+import { parseConsoleScript } from "../utils/ast";
 import { format } from "prettier";
 
 const formatOutput = output =>
   format(output, {
     parser: "babylon"
   });
 
+const mapOgExpression = (expression, mappings) =>
+  mapOriginalExpression(expression, parseConsoleScript(expression), mappings);
```
jasonLaster: maybe it would be better to use the public API `mapExpression`?

src/workers/parser/mapExpression.js
nchevobbe: right, this can be moved to another PR

src/workers/parser/tests/mapExpression.spec.js
nchevobbe: we don't want to parse the expression like for the other expression, because it would throw (see https://github.com/devtools-html/debugger.html/pull/7695/files/07ef9a76c3f1c5dbed0dfbdd096786dcc15344bd#diff-fff0d7c8c37a491198a0e4f041c09ea5L23) 
So we need to explicitly indicate that we don't want to parse it. And then, we check that an invalid expression (note the extra closing parens) is still wrapped in an async iife.

src/workers/parser/mapExpression.js
nchevobbe: I removed the `getAst` function and just put the result in an `ast` variable. We'll handle memoization in a new PR

src/workers/parser/tests/mapOriginalExpression.spec.js
nchevobbe: done

PR #7697, opened 2019-01-09
Switch from Inline SVGs to images

jasonLaster:
### Summary of Changes

This switches us away from SVGs to images for a couple of reasons:

1. images are faster to render
2. images are better for MC, currently we're building all of the SVGs into vendors.js as well as copying many of them

### notes

- the quick open magninfying glass is broken


darkwing:
Magnifying glass is broken as are the modifier icons:

broken1

darkwing:
Root looks like it's having a hard time:

root

darkwing:
This is super close; I noticed that the Webpack and Angular icons are slightly off:

webpack

mainlevel

The React and Angular icons look fine when they aren't at the top level, so I think it may just be top level icons needing fixing.

PR #7705, opened 2019-01-10
Respect enable extenstions setting in Sources and QuickOpenModal

### LINKED ISSUE #7531: `Sources` and QuickOpenModal should respect `Enable add-on debugging` setting
As most "power-users" I use several add-ons. I use these but I have no intend of debugging these. Now these fill-up my `Sources` tab:
image

And also fill and slow down my `Ctrl+P`:
image

NB: In the above all searches for `react` find only my add-ons.

Expected results:
I do not want to debug add-ons. If I have an intent to do so, I'd be happy to learn that there's an option to turn on Add-on debugging. So please add an option under `F1` aka `Settings`  under `Advanced settings` to `Enable WebExtension (Add-on) debugging`. This checkbox setting should be off by default. Th

konradk:
Fixes #7531 

### Summary of Changes

* add default preference in development environment: "devtools.chrome.enabled" as "false"
* filter sources by extenstions when "devtools.chrome.enabled" is "false"

### Test Plan

- [x] Add extenstion
- [x] Turn off "Enable browser chrome and add-on debugging toolboxes" preferece
- [x] Source list shouldn't list extensions
- [x] QuickOpenModal shouldn't list extenstions

My questions:
- [x] I've only edited `newSources` action, because that's only action I've seen in development mode. Is there any other action I should edit - e.g. `newSource`?
- [x] Is there better way to identify extension than `url.startsWith("moz-extension://"))`?

## To do:

- [x] fix mochitests browser_dbg-content-script-sources.js
- [x] fix mochitests browser_dbg-sources-arrow-keys.js
- [x] Update test snapshots
- [x] Support extension prefix with single slash
- [x] fix type errors


src/actions/sources/newSources.js
```diff
@@ -201,6 +201,11 @@ export function newSource(source: Source) {
 export function newSources(sources: Source[]) {
   return async ({ dispatch, getState }: ThunkArgs) => {
     sources = sources.filter(source => !getSource(getState(), source.id));
+    if (!prefs.chromeAndExtenstionsEnabled) {
+      sources = sources.filter(
+        source => !(source.url && source.url.startsWith("moz-extension://"))
+      );
+    }
```
jasonLaster: i think we could move this flag to a selector like `getRelativeSources`, which will just decide what is shown in the tree and modal.

jasonLaster:
here are links to the tests

https://github.com/devtools-html/debugger.html/blob/master/src/components/PrimaryPanes/tests/SourcesTree.spec.js#L1

https://github.com/devtools-html/debugger.html/blob/master/src/components/test/QuickOpenModal.spec.js#L1

src/client/firefox/create.js
```diff
@@ -54,6 +54,7 @@ export function createSource(
     relativeUrl: source.url,
     isPrettyPrinted: false,
     isWasm: false,
+    isExtension: source.url && source.url.startsWith("moz-extension://"),
```
princiya: Does this address the extensions from Chrome browser?

konradk: No, I've just fixes that - fixed regexp and moved this set to common source action creator

konradk:
@jasonLaster Currently filtering extensions is the responsibility of `getRelativeSources` selector. I don't see the possibility to add the test for this enhancement in SourceTree or QuickOpenModal specs, because it's not their responsibility to filter them. 

I've prepared the reducers test [here](https://github.com/devtools-html/debugger.html/pull/7705/commits/35661a1b52e9109f9740d9f5df7e1d8c8691edf1), but I'm not sure if it's a good approach.

jasonLaster:
this looks good

jasonLaster:
failed test:  devtools/client/debugger/new/test/mochitest/browser_dbg-content-script-sources.js 
failed test:  devtools/client/debugger/new/test/mochitest/browser_dbg-sources-arrow-keys.js 

two mochitests fail -- hopefully the errors are easy to follow...

src/utils/source.js
```diff
@@ -484,3 +484,7 @@ export function getSourceQueryString(source: ?Source) {
 
   return parseURL(getRawSourceURL(source.url)).search;
 }
+
+export function isUrlExtension(url: string) {
+  return /^(chrome|moz)-extension:\/\//.test(url);
```
gerdneuman: @konradk Just a heads up: I also have at least one extension (namely [User Agent Switcher](https://addons.mozilla.org/en-US/firefox/addon/uaswitcher/)) whose scheme seems to end by only *one* trailing slash and not two. The above regex tests with two trailing slashes at the end.

See screenshot:

image

Pinging also @AnshulMalik because I think this is probably relevant for https://bugzilla.mozilla.org/show_bug.cgi?id=1518221 as well.

Note sure how to handle this (and why this is under `blob://`) but testing against 
```js
// one trailing slash
return /^(chrome|moz)-extension:\//.test(url);
```
and not 
```js
// two trailing slashes
return /^(chrome|moz)-extension:\/\//.test(url);
```
would be worthwhile.

If so, probably also good if a test would be added for this edge case.

konradk: @gerdneuman Nice catch, I will fix that.

AnshulMalik: Not sure why it's blob and has just one `/`

PR #7714, opened 2019-01-11
Add 'Collapse all' and 'Expand all' sources context menu options

### LINKED ISSUE #7665: Add Sources tree context menu entries for "Expand all" and "Collapse all"
I recently discovered that you can alt-click on a sources tree item to expand it and all of its children. 

It would be good to increase the visibility of this feature by adding it to the sources tree item context menu.

Garee:
Reintroduce the `setExpanded` function to the `SourcesTreeItem` component. Use this function to collapse/expand all for the selected item.

Add two new resource strings for each menu option.

In the future perhaps we could disable the option if all of the children are already collapsed/expanded.

Fixes #7665

### DIFF
--- a/assets/panel/debugger.properties
+++ b/assets/panel/debugger.properties
@@ -30,6 +30,14 @@ copySource.accesskey=y
 copySourceUri2=Copy source URI
 copySourceUri2.accesskey=u
 
+# LOCALIZATION NOTE (collapseAll.label): This is the text that appears in the
+# context menu to collapse a directory and all of its subdirectories.
+collapseAll.label=Collapse all
+
+# LOCALIZATION NOTE (expandAll.label): This is the text that appears in the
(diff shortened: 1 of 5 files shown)


princiya:
Thanks @Garee 👍 

@flodolo I see you have approved this PR, so I will merge it.

flodolo:
I don't think that was the right thing to do, unless you reviewed the code.
I only check one file (.properties) for localization issues, I'm not a
developer so the rest of the code is up for the product owners to review.

Garee:
I'm not sure this was code reviewed before it was merged. Can someone please have a look and let me know if I need to make any follow up changes? Thanks.

@princiya @jasonLaster @darkwing

darkwing:
I looked over this and it looks pretty good.  We should file a follow up issue to address disabling/enabling the context menu items when possible.

PR #7715, opened 2019-01-11
Implement Event Listeners Panel

darkwing:
This is the first barebones UI for the Event Listeners panel.  I initially implemented it with basic React but this latest commit moves it to the Tree component.

## TODO Before Merging:

- [x] Remove the old implementation
- [x] Get the tree arrow expand/collapse working
- [x] If all checkboxes under a heading are checked, ensure its parent is also checked
- [ ] Tests


darkwing:
@nchevobbe I need your advice on how to get the checkboxes to check/uncheck upon click.  My basic react implementation did so just fine, but within the tree nothing appears to happen UI-wise.  I'd love for you to check this out.

jasonLaster:
Looks good!

nchevobbe:
So, the `renderItem` prop is called from the `TreeNode` component in `tree.js`
The issue you're hitting is that the element does not update because of: 

```js
  shouldComponentUpdate(nextProps) {
    return (
      this.props.item !== nextProps.item ||
      this.props.focused !== nextProps.focused ||
      this.props.expanded !== nextProps.expanded
    );
  }
```

Basically we're only re-rendering if the item changed, or if it is now expanded or focused.
The only way for you to get through a re-render would be to change item (which doesn't look ideal), or implement some prop for `Tree` which we would use in TreeNode's `shouldComponentUpdate` (might be an overkill).

Also, maybe we can not use the Tree component?
From what I see and read in the code, it looks like you'll only have a 1 level deep "tree".
It might not be too complex to handle this directly in `EventListeners`.

src/client/firefox/commands.js
```diff
@@ -386,6 +386,11 @@ function eventListeners(): Promise<*> {
   return threadClient.eventListeners();
 }
 
+function setEventListenerBreakpoints(eventTypes: string[]) {
+  // TODO: Figure out what sendpoint we want to hit
+  threadClient.setEventListenerBreakpoints(eventTypes);
```
jasonLaster: i'd comment this out and logan can add it when he's ready.

jasonLaster: it's not really clear what the ideal API will be anyways

PR #7718, opened 2019-01-14
[Breakpoints] include breakpoint in column breakpoint

jasonLaster:
### Summary of Changes

- adds breakpoint to column breakpoint so that it will be easier to toggle breakpoints in the context menu.
- this also seems more future proof
- added some flow types


jasonLaster:
yep - this will be fixed in the next PR.

PR #7719, opened 2019-01-14
(WIP) Update editor context menus

### LINKED ISSUE #7497: Restructure Context Menus
We have some really awesome context menus that make using the DevTools debugger easier and more effective.  Each context menu was custom created at the time of need which was great when they were implemented but, a few years in, I think we can take a step back and broaden both their code structure and visual structure.

We should create context menu "sections", which would include:

- Breakpoint (enable, disable, remove, add condition, remove condition, etc.)
- Breakpoints (enable all, disable all, remove all, remove all on line, etc.)
- Source (copy to clipboard, copy url, reveal in tree, pretty print, etc.)
- Tab (close, close other, close all, close all to the right, etc.)

The advantage here would be using multiple sections in a single context menu.  For example, the context menu that displays when you right-click a tab could contain: Tab, Source, and possibly even Breakpoint (i.e. to let someone disable all breakpoints in a source).

jasonLaster:
Fixes #7497

### Summary of Changes
 
Before starting on the column breakpoint server-side work, I'd like to clean up how the Editor uses context menus. 

There are a couple of reasons for this:

1. the current menu components (gutter/editor) assume a fixed set of items. We have found that we want to re-use these items in different contexts
2. the gutter menu is used for column breakpoints
3. the menus have a lot of complexity when it comes to breakpoints because they have to support adding/removing a breakpoint based on a location. Removing should be much easier because we by definition have a breakpoint. Going forward the (Column)Breakpoint component handles removing breakpoints.


assets/panel/debugger.properties
```diff
@@ -512,6 +512,14 @@ editor.addConditionBreakpoint.accesskey=c
 # for setting a breakpoint condition on a line.
 editor.editConditionBreakpoint=Edit condition
 
+# LOCALIZATION NOTE (editor.addConditionalBreakpoint): Editor gutter context menu item
+# for creating a breakpoint with a condition
+editor.addConditionalBreakpoint=Add conditional breakpoint
+
+# LOCALIZATION NOTE (editor.addLogBreakpoint): Editor gutter context menu item
+# for creating a breakpoint with a log
+editor.addLogBreakpoint=Add Log point
```
darkwing: We probably want `Add log point`

flodolo: That would be my expectation too.

jasonLaster:
noticing two things:

1. adding a condition by-passes the first possible BP and adds it to the first actual column bp
2. disabling a column bp doesn't work
3. the editor menu fails

...

darkwing:
@jasonLaster I was going to make this update but since you're continually updating it, you'll want to update `ColumnBreakpoints.css` to have:

```css
.column-breakpoint {
  display: inline;
  padding: 0;
  margin-inline-end: 4px;
}

.column-breakpoint:hover {
  background-color: transparent;
}
```

src/actions/debuggee.js
```diff
@@ -6,12 +6,15 @@
...
 
 export function updateWorkers() {
   return async function({ dispatch, getState, client }: ThunkArgs) {
     const workers = await client.fetchWorkers();
     dispatch(({ type: "SET_WORKERS", workers }: Action));
 
-    dispatch(closeTabsForMissingThreads(workers));
+    if (features.windowlessWorkers) {
+      dispatch(closeTabsForMissingThreads(workers));
+    }
```
jasonLaster: this could be in its own PR, but its not too important either way...

src/actions/tabs.js
```diff
@@ -113,10 +113,10 @@ export function closeTabsForMissingThreads(workers: Worker[]) {
...
-    const tabs = removeSourcesFromTabList(oldTabs, removed);
-    dispatch({ type: "CLOSE_TABS", removed, tabs });
-
-    const sourceId = getNewSelectedSourceId(getState(), tabs);
-    dispatch(selectSource(sourceId));
+    // const tabs = removeSourcesFromTabList(oldTabs, removed);
+    // dispatch({ type: "CLOSE_TABS", removed, tabs });
+    //
+    // const sourceId = getNewSelectedSourceId(getState(), tabs);
+    // dispatch(selectSource(sourceId));
```
jasonLaster: same with this...

src/components/Editor/Editor.css
```diff
@@ -92,6 +92,10 @@ html[dir="rtl"] .editor-mount {
   );
 }
 
+.new-breakpoint .CodeMirror-linenumber {
+  pointer-events: none;
+}
```
jasonLaster: this was essential for the Editor/Breakpoint.js onClick and onContextMenu to work!

src/reducers/breakpoints.js
```diff
@@ -315,27 +314,16 @@ export function getBreakpointForLocation(
...
-export function getHiddenBreakpointLocation(
-  state: OuterState
-): ?SourceLocation {
-  const hiddenBreakpoint = getHiddenBreakpoint(state);
-  if (!hiddenBreakpoint) {
-    return null;
-  }
-  return hiddenBreakpoint.location;
-}
-
```
jasonLaster: no longer needed :smile:

src/actions/debuggee.js
jasonLaster: moved this here -- https://github.com/devtools-html/debugger.html/pull/7747

PR #7721, opened 2019-01-15
XHR breakpoints with methods

### LINKED ISSUE #7645: Support XHR breakpoints with methods
Currently, we support breaking on a network request if the URL contains some string.
It'd be a nice improvement if we can also specify the method(GET, POST, etc.) of the request to give the user more control.

Here is how the UI should look like

I'll be happy to mentor if someone would like to give it a try :)

Here is how we render the breakpoints: https://github.com/devtools-html/debugger.html/blob/master/src/components/SecondaryPanes/XHRBreakpoints.js#L171-L210

jaril:
### Preview

XHR breakpoints pane with methods
jan-14-2019 19-12-59

Pausing behavior in action
jan-14-2019 19-48-53

### Summary of Changes

* Closes #7645
* Adds support for XHR breakpoints breaking on specific methods (e.g. GET, POST, etc.)
* Default is `ANY` method, which is how all XHR breakpoints are currently set
* Added tests for XHRBreakpoints component

### Test Plan

Run tests with `npm run test xhrbreakpoints`


src/components/SecondaryPanes/XHRBreakpoints.css
```diff
@@ -7,12 +7,12 @@
 }
 
 .xhr-input {
-  width: 100%;
+  width: 70%;
```
jasonLaster: maybe flexbox would be a good fit here?

jasonLaster:
Wow! Thanks the first pass looks looks great!

src/components/SecondaryPanes/XHRBreakpoints.css
darkwing: Good idea; we can use flexbox to take up available space after before the `<select>` element.

src/components/SecondaryPanes/XHRBreakpoints.css
```diff
@@ -85,3 +85,21 @@
   position: absolute;
   top: 8px;
 }
+
+.xhr-input-method {
```
darkwing: Whenever we use `-right` or `-left`, it makes us vulnerable to RTL.  We use `margin-inline-start/end` in a few spots, so you'll want to check RTL out.

src/components/SecondaryPanes/XHRBreakpoints.js
```diff
@@ -242,6 +271,39 @@ class XHRBreakpoints extends Component<Props, State> {
...
+  renderMethodOption = method => {
+    return (
+      <option key={method} value={method}>
+        {method}
+      </option>
+    );
+  };
+
+  renderMethodSelectElement = () => {
+    const methods = [
```
darkwing: We can declare this as a const at the top of the file so we aren't rebuilding this array each time this method is called.

darkwing:
This looks great @notjaril !  Needs a few updates but very solid!  We may also consider a mochitest here as well -- they've been saving us a lot of pain lately.

src/components/SecondaryPanes/XHRBreakpoints.js
```diff
@@ -103,19 +105,43 @@ class XHRBreakpoints extends Component<Props, State> {
...
+  keepFocusOnInputButNoEdit = () => {
+    this.setState({ focused: true, editing: false });
   };
 
   onFocus = () => {
-    this.setState({ focused: true });
+    this.setState({ focused: true, editing: true });
+  };
+
+  onMouseDown = e => {
```
derek-li: @notjaril Just a nitpick, but I believe passing in this 'e' parameter is not necessary.

notjaril: good catch - thanks!

src/components/SecondaryPanes/XHRBreakpoints.css
notjaril: Worked around this by refactoring CSS to have the same left/right values for paddings and margins. 

`margin/padding-inline-start/end` seems like it would have done the trick as well, but was hesitant to introduce browser compatibility issues if just making the left/right values the same solves the problem.

src/components/SecondaryPanes/XHRBreakpoints.js
notjaril: Fixed!

src/components/SecondaryPanes/XHRBreakpoints.css
notjaril: Sounds good - changed the form to `display: flex`, and the the URL input box now takes up all the space before `<select>`

darkwing:
Regardless of edit or new XHR breakpoint, changing the `<select>`/method resets back to `ANY`.

darkwing:
I believe you'll want to add a `selected` attribute to the `<option>` with the matching value

notjaril:
> Regardless of edit or new XHR breakpoint, changing the `<select>`/method resets back to `ANY`. 

Fixed. Was caused by a specific event behavior on Firefox that re-rendered `<select>` list upon change, reverting the value to the default "ANY" before the onChange callback was fired. 

jan-21-2019 10-46-06

Just need to rewrite mochitests and should be good to go :)

darkwing:
@notjaril Looks like something bad happened here with a merge; could you rebase?

src/components/SecondaryPanes/tests/XHRBreakpoints.spec.js
```diff
@@ -0,0 +1,342 @@
...
+
+  it("should have all 8 methods available as options", function() {
+    const xhrBreakpointsComponent = renderXHRBreakpointsComponent();
+    xhrBreakpointsComponent.find(".xhr-input-url").simulate("focus");
+
+    const xhrInputMethod = xhrBreakpointsComponent.find(".xhr-input-method");
+    expect(xhrInputMethod.children()).toHaveLength(8);
+
+    const actualXHRMethods = [];
+    const expectedXHRMethods = [
```
darkwing: This shouldn't need to be defined more than once.  Also, can this be `export`ed from the XHRBreakpoints.js file instead of duplicated?

darkwing:
@jaril Just one tiny update and I think we should be good here.

src/components/SecondaryPanes/tests/XHRBreakpoints.spec.js
jaril: Extracted the `xhrMethods` array so it's only defined once in `XHRBreakpoints.spec.js`.

It can be exported but doing so would disable that specific test from testing what it's asserting. e.g. tests will still pass even if for some reason, the array in `XHRBreakpoints.js` was edited to have 7 methods instead of 8, or if any of the methods were mispelled.

digitarald:
Summoning @mcroud to give UX feedback.

My preference would be to style the form and the XHR breakpoint entry according to the mockup from the referenced issue:

image

Calling out specifics:
- Visual difference between _method_ and _pattern_ in the entry (color and type) and using monospaced type for the _pattern_
- Layout of the form with separator and dropdown styling aligned with photon dropdowns

jasonLaster:
Perhaps we do a follow up or for styling? 

I'm suggesting this because the functionality was a lot of work. Itd be shame for that to go stale while we sort out some of the design aspects.

digitarald:
> Perhaps we do a follow up or for styling?

Yes, my feedback should not block this PR.

darkwing:
We're 95% of the way there!  Love it!

Two things I noticed:

1.  If I add two XHR strings that are the same but different methods, I see this react error:

```js
Warning: Encountered two children with the same key, `fetch`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.
    in ul (created by XHRBreakpoints)
    in div (created by XHRBreakpoints)
    in XHRBreakpoints (created by Connect(XHRBreakpoints))
    in Connect(XHRBreakpoints) (created by SecondaryPanes)
    in div (created by Accordion)
    in li (created by Accordion)
    in ul (created by Accordion)
    in Accordion (created by SecondaryPanes)
    in div (created by SecondaryPanes)
    in div (created by SecondaryPanes)
    in SecondaryPanes (created by Connect(SecondaryPanes))
    in Connect(SecondaryPanes) (created by App)
    in div (created by SplitBox)
    in div (created by SplitBox)
    in SplitBox (created by App)
    in div (created by A11yIntention)
    in A11yIntention (created by App)
    in div (created by App)
    in App (created by Connect(App))
    in Connect(App)
    in Provider
``` 

You'll just need to adjust the key to be something like `key={path}-{method}`.

2.  If I type a string and press [tab] to tab over to the method dropdown, the inputs disappear.  We should remove the tab handling for that input.

jaril:
Pushed a new commit changing the mochitest implementation fixing false positive test results. 

The problem is with the sequencing. `invokeInTab()`, whose behavior may pause the debugger, is called first and then `assertNotPaused()` is called immediately afterwards. The problem is that pausing the debugger doesn't happen immediately, and what happens is that `assertNotPaused()` is called before the debugger has a chance to be paused. This is what causes tests to pass regardless of whether the debugger is actually paused or not.

A 1 second delay in between `invokeInTab()` and `assertNotPaused()` gives the debugger enough time to pause before the test runner checks where it is paused (or not). It's not the most elegant fix, but it gets the job done. 

Another option is to use await statements (e.g. `await invokeInTab(...)`). This works well if the code is running as expected, which is that it doesn't pause the debugger where we call `assertNotPaused()`. However, if the debugger does get paused in those cases, as would be the case if the XHR breakpoint setting behavior is errant, `invokeInTab()` never gets a chance to resolve and the test stalls until it times out. 

I don't prefer this as much as it's harder to debug - this method would still let you know if something was wrong with XHR Breakpoints, but it would only let you know that the test timed out instead of the actual reason for the test failing.

darkwing:
@jaril The test problem was simply that you needed a rebase.  Now there's two small jest test issue, then we should be good to go!

test/mochitest/browser_dbg-xhr-breakpoints.js
```diff
@@ -28,37 +35,46 @@ async function removeXHRBreakpoint(dbg, index) {
...
   await dbg.actions.removeXHRBreakpoint(0);
   invokeInTab("main", "doc-xhr.html");
+  await wait(1000); // timeout needed before checking whether it's paused
   assertNotPaused(dbg);
 
-  await dbg.actions.setXHRBreakpoint("doc-xhr.html", "POST");
-  invokeInTab("main", "doc");
+  await addXHRBreakpoint(dbg, "doc", "POST");
+  invokeInTab("main", "doc-xhr.html");
+  await wait(1000); // timeout needed before checking whether it's paused
```
darkwing: Oops, we'll want a way around these waits; did you try my suggestion re: polling instead of a generic number?

src/components/SecondaryPanes/XHRBreakpoints.js
```diff
@@ -88,9 +101,21 @@ class XHRBreakpoints extends Component<Props, State> {
...
+      this.hideInput();
+    };
 
-    this.hideInput();
+    // force update inputMethod in state for mochitest purposes
+    // before setting XHR breakpoint
+    this.setState(
+      // $FlowIgnore
+      { inputMethod: e.target.children[1].value },
+      setXHRBreakpoint
```
AnshulMalik: Umm, we are setting this `setXHRBreakpoint` in the state, I don't understand this part.

jaril: We are. This is just a workaround for mochitests to make it testable, since there's no reliable way to simulate a user change event for the `<select>` element without importing `ReactUtils`.

Instead, I just programmatically change the value of the `<select>` element. But changing the value itself doesn't trigger a change event, which doesn't update the `inputMethod` in state, so I added another "force update" on submit to make sure that it updates the `inputMethod` on submit.

`setState` just requests a state change to be queued - it doesn't always immediately update the state as it's called. `setXHRBreakpoint` is passed into `setState` specifically to make sure that it happens after the `inputMethod` is updated on submit. Otherwise, what happens is `setXHRBreakpoint` gets called first before the state is updated.

AnshulMalik: 👍 Makes sense now. Thanks :)

AnshulMalik:
This is awesome @jaril

digitarald:
Follow up bug filed for method changing https://bugzilla.mozilla.org/show_bug.cgi?id=1557780

PR #7723, opened 2019-01-15
Make the Frames component copy-friendly

nchevobbe:
This PR adds a `selectable` props to the Frames component,
which, when true, will cause some white-space only nodes to
be added into the DOM.

We also switch list from `<ul>`s to `<div>`s with proper aria roles. This is unfortunate but this is the only way of not having the engine adding empty lines around the copied text (See [nsXHTMLContentSerializer.cpp#560,567](https://searchfox.org/mozilla-central/rev/c3ebaf6de2d481c262c04bb9657eaf76bf47e2ac/dom/base/nsXHTMLContentSerializer.cpp#560,567))

This is mainly aimed for the console where stacktraces can be
copied and where we want the clipboard output to be as helpful
as what is shown in the console output.

---

The testing is pretty light here because none of the component in the debugger should make use of the `selectable` prop.
It will be used in the console as part of [Bug 1486870](https://bugzilla.mozilla.org/show_bug.cgi?id=1486870), with some mochitests.

At some point that could be used to implement the `Copy stacktrace` context menu entry that we have in the debugger (at the moment the process feels very manual).

### DIFF
--- a/packages/devtools-reps/src/reps/error.js
+++ b/packages/devtools-reps/src/reps/error.js
@@ -57,7 +57,7 @@ function ErrorRep(props) {
     const stacktrace = props.renderStacktrace
       ? props.renderStacktrace(parseStackString(preview.stack))
       : getStacktraceElements(props, preview);
-    content.push("\n", stacktrace);
+    content.push(stacktrace);
   }
 
   return span(
--- a/src/components/SecondaryPanes/Frames/FrameIndent.js
+++ b/src/components/SecondaryPanes/Frames/FrameIndent.js
@@ -1,0 +1,13 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
+
+import React from "react";
+
+export default function FrameIndent() {
+  return (
+    <span className="frame-indent clipboard-only">
+      &nbsp;&nbsp;&nbsp;&nbsp;
+    </span>
+  );
+}
--- a/src/components/SecondaryPanes/Frames/Group.js
+++ b/src/components/SecondaryPanes/Frames/Group.js
@@ -10,13 +10,13 @@ import Svg from "../../shared/Svg";
 import { getLibraryFromUrl } from "../../../utils/pause/frames";
 import FrameMenu from "./FrameMenu";
 import AccessibleImage from "../../shared/AccessibleImage";
+import FrameComponent from "./Frame";
 
 import "./Group.css";
 
-import FrameComponent from "./Frame";
-
 import type { LocalFrame } from "./types";
 import Badge from "../../shared/Badge";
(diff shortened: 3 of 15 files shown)


packages/devtools-reps/src/reps/error.js
```diff
@@ -57,7 +57,7 @@ function ErrorRep(props) {
     const stacktrace = props.renderStacktrace
       ? props.renderStacktrace(parseStackString(preview.stack))
       : getStacktraceElements(props, preview);
-    content.push("\n", stacktrace);
+    content.push(stacktrace);
```
nchevobbe: not needed to add a new line after the error message anymore

src/components/SecondaryPanes/Frames/FrameIndent.js
```diff
@@ -0,0 +1,13 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
+
+import React from "react";
+
+export default function FrameIndent() {
+  return (
+    <span className="frame-indent clipboard-only">
+      &nbsp;&nbsp;&nbsp;&nbsp;
```
nchevobbe: Using spaces because we can't get an inline tab character (it gets mangled in everything but `<pre>`s, but `<pre>` creates line breaks).

src/components/SecondaryPanes/Frames/Group.js
```diff
@@ -105,48 +107,58 @@ export default class Group extends Component<Props, State> {
...
-            toggleBlackBox={toggleBlackBox}
-            toggleFrameworkGrouping={toggleFrameworkGrouping}
-            displayFullUrl={displayFullUrl}
-            getFrameTitle={getFrameTitle}
-            disableContextMenu={disableContextMenu}
-          />
-        ))}
+        {group.reduce((acc, frame) => {
+          if (selectable) {
+            acc.push(<FrameIndent />);
```
nchevobbe: This allow us to have a double indentation for React frames, so the text version is more readable: 

```
Warning: Each child in an array or iterator should have a unique "key" prop. react-dom.development.js:520
    React 10
        warningWithoutStack
        warning
        warnForMissingKey
        warnOnInvalidKey
        reconcileChildrenArray
        reconcileChildFibers
        reconcileChildren
        finishClassComponent
        updateClassComponent
        beginWork
    <anonymous> index.js:6
    js bundle.js:109
    __webpack_require__ bundle.js:20
    <anonymous> bundle.js:84
    <anonymous> bundle.js:1

```

jasonLaster:
This commit is failing try because of this test:

browser_webconsole_context_menu_copy_entire_message.js
https://treeherder.mozilla.org/#/jobs?repo=try&revision=b83e388b82d9401f356d05a04530e27342bc52c4&selectedJob=222601985

Lets land a fix in github for this test and include the test here
https://github.com/devtools-html/debugger.html/blob/release-119/.travis.yml#L62-L68

PR #7730, opened 2019-01-15
replace formatted breakpoint

jasonLaster:
### Summary

This is a small opinionated refactor, which is necessitated by our increased use of flow. The lesson is that `FormattedBreakpoint` can't be passed into functions that expect a `Breakpoint`


src/utils/breakpoint/index.js
```diff
@@ -166,7 +167,7 @@ export function createXHRBreakpoint(
     method,
     disabled: false,
     loading: false,
-    text: L10N.getFormatStr("xhrBreakpoints.item.label", path)
+    text: `URL contains "${path}"`
```
darkwing: This breaks l10n; I fixed this with a PR earlier this week.

PR #7733, opened 2019-01-16
Fix #7566 Use the same expand arrow icon as Inspector, Console etc.

### LINKED ISSUE #7566: Use the same expand arrow icon as Inspector, Console etc.
We updated the expand arrow icon recently ([ux issue](https://github.com/devtools-html/ux/issues/31)) and implemented the new icon in mozilla-central but not yet in Debugger/Reps.

The correct icon can be found here:
https://searchfox.org/mozilla-central/source/devtools/client/themes/images/arrow.svg

It's a 10x10 icon, but is visually just 8x8. It will probably appear a little bit smaller than the current icon in Debugger.

When it's implemented as an icon-only button (e.g. if the user needs to click the icon itself, rather than the surrounding label or row), we should aim for a minimum button size of 14x14 pixels (which is tiny, but better than 10x10, and a bigger button often doesn't work with our small text and line-height).

ash-tamraz:
Fixes #7566 

### Summary of Changes

* Implemented new `arrow.svg` icon found in mozilla-central
* Edited stylesheets to ensure proper sizing of new arrow icon in debugger.

### Test Plan

Examined arrow icons with Firefox's DevTools' Inspector, in the Layout panel, and ran `yarn run test:all`.

### DIFF
--- a/src/components/shared/Svg.css
+++ b/src/components/shared/Svg.css
@@ -316,17 +316,16 @@ span.img.marko {
 
 .img.arrow {
   mask: url(/images/arrow.svg);
-  margin-inline-end: 5px;
-  margin-top: 3px;
-  width: 9px;
-  height: 9px;
+  margin-inline-end: 4px;
+  width: 10px;
+  height: 10px;
   padding-top: 9px;
   background: var(--disclosure-arrow);
   mask-size: 100%;
   display: inline-block;
-  margin-bottom: 1px;
   transform: rotate(-90deg);
   transition: transform 0.18s ease;
+  vertical-align: -1px;
 }
 
 html[dir="rtl"] .img.arrow {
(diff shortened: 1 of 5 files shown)


src/components/shared/Svg.css
```diff
@@ -353,6 +360,30 @@ html[dir="rtl"] .arrow,
   visibility: hidden;
 }
 
+.webpack svg {
```
darkwing: Were these changes intended?

ash-tamraz: It looks like I made a mistake with the mergetool. I've (hopefully) fixed that and I cleaned up the commits. Let me know if there are any other problems and I'll get them fixed!

darkwing:
@nchevobbe This looks good in debugger; could you take a quick look at console?  If so we can merge.

Nice job @ash-tamraz !

nchevobbe:
Looks fine in the console 👍

PR #7749, opened 2019-01-18
Distinguish between condition and log value in breakpoints

bhackett1024:
### Summary of Changes

* Separately track breakpoint conditions and log values, storing them in a BreakpointOptions object.
* The options are converted into a single condition string before passing onto the source client / breakpoint client.  Bug 1520972 will take care of remaining changes to push this down into the server.

### DIFF
--- a/src/client/firefox/commands.js
+++ b/src/client/firefox/commands.js
@@ -6,6 +6,7 @@
 
 import type {
   BreakpointId,
+  BreakpointOptions,
   BreakpointResult,
   EventListenerBreakpoints,
   Frame,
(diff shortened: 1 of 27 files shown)


src/client/firefox/commands.js
```diff
@@ -226,18 +236,17 @@ function removeBreakpoint(
...
-  noSliding: boolean
+  options: BreakpointOptions
 ) {
   const bpClient = bpClients[breakpointId];
   delete bpClients[breakpointId];
 
   const sourceThreadClient = sourceThreads[bpClient.source.actor];
   return bpClient
-    .setCondition(sourceThreadClient, condition, noSliding)
+    .setCondition(sourceThreadClient, transformOptionsToCondition(options))
```
bomsy: should we keep the `noSliding` argument?

bhackett1024: I removed this because BreakpointClient.prototype.setCondition doesn't actually take a third noSliding argument.  It also ignores the thread client argument, which is tidied up by the patch in bug 1520972.

PR #7753, opened 2019-01-18
Use jsterm directly for console logging

### LINKED ISSUE #7668: The "Evaluate in console" context menu item does not work
**Steps to reproduce**

- Set and trigger a break point in a source file.
- Select an expression.
- Right click and click "Evaluate in console"
- View the result in the console.

**Expected Result**

The result of the evaluated expression should be shown in the console.

**Actual Result**

Nothing is logged to the console.

darkwing:
Fixes #7668

Rather than doing what we are now, using the JSTerm API directly may be safer.  This is a discussion item.


assets/panel/panel.js
```diff
@@ -50,6 +50,11 @@ DebuggerPanel.prototype = {
             .setNodeFront(front, { reason: "debugger" });
 
           return Promise.all([onNodeFrontSet, onInspectorUpdated]);
+        }.bind(this),
+        openConsoleAndEvaluate: async function(input) {
+          await this.toolbox.selectTool("webconsole");
+          const jsterm = this.toolbox.getPanel("webconsole").hud.jsterm;
```
jasonLaster: this could be a bit slow, but because this is an infrequent action, it might be better than spinning up our own jsterm

darkwing: Agreed.

test/mochitest/browser_dbg-console-eval.js
```diff
@@ -0,0 +1,34 @@
...
+  clickElement(dbg, "codeMirrorCode");
+  await waitForElementWithSelector(dbg, ".CodeMirror-code");
+
+  // pressKey(dbg, "selectAll"); // This doesn't work properly
+  getCM(dbg).setSelection({ line: 0, ch: 0 }, { line: 8, ch: 0 });
+
+  rightClickElement(dbg, "codeMirrorCode");
+  selectContextMenuItem(dbg, "#node-menu-evaluate-in-console");
+
+  await waitForConsolePanelChange(dbg);
```
darkwing: To this point the test really only ensures that the panel switches to WebConsole; a last step is ensuring the console output is as desired, so I may need to steal a few webbconsole helper functions.  Would also like @nchevobbe's thoughts on the PR overall.

assets/panel/panel.js
```diff
@@ -50,6 +50,11 @@ DebuggerPanel.prototype = {
             .setNodeFront(front, { reason: "debugger" });
 
           return Promise.all([onNodeFrontSet, onInspectorUpdated]);
+        }.bind(this),
+        openConsoleAndEvaluate: async function(input) {
+          await this.toolbox.selectTool("webconsole");
```
nchevobbe: If I read https://searchfox.org/mozilla-central/rev/78e8de968efd77ef3b30ea081bce07614cd6776b/devtools/client/framework/toolbox.js#1952,1956 correctly, you will receive the panel here, so you don't have to do `getPanel` on the next line.

assets/panel/panel.js
```diff
@@ -50,6 +50,11 @@ DebuggerPanel.prototype = {
...
+          const jsterm = this.toolbox.getPanel("webconsole").hud.jsterm;
+          jsterm.execute(input);
```
nchevobbe: that looks reasonable to me. At some point, I think we're going to move `execute` outside of jsterm to the panel, so we can avoid retrieving the jsterm reference.

test/mochitest/browser_dbg-console-eval.js
nchevobbe: we have 2 helpers for messages:
- you can use [findMessage](https://searchfox.org/mozilla-central/rev/78e8de968efd77ef3b30ea081bce07614cd6776b/devtools/client/webconsole/test/mochitest/head.js#240) with `waitFor` in order to wait until a given message shows up.
- [waitForMessage](https://searchfox.org/mozilla-central/rev/78e8de968efd77ef3b30ea081bce07614cd6776b/devtools/client/webconsole/test/mochitest/head.js#186), but this one is based on an event fired on the panel, so you need to have a reference to the panel before doing the evaluation, which looks tricky in this case.

PR #7754, opened 2019-01-18
[breakpoints] Add possible breakpoints positions

jasonLaster:
### Summary of Changes

Adds "possible breakpoints" which are positions that the server says would be good places to place breakpoints. This is something we want for a couple of reasons:

1. when we place a breakpoint we want to know it'll be good
2. we want to know what lines the engine considers empty

### DIFF
--- a/src/actions/breakpoints/addBreakpoint.js
+++ b/src/actions/breakpoints/addBreakpoint.js
@@ -20,12 +20,13 @@ import {
   getSource,
   getSourceActors,
   getSymbols,
-  getFirstPausePointLocation
+  getFirstVisibleBreakpointPosition
 } from "../../selectors";
 import { getGeneratedLocation } from "../../utils/source-maps";
 import { getTextAtPosition } from "../../utils/source";
 import { recordEvent } from "../../utils/telemetry";
 import { features } from "../../utils/prefs";
+import { setBreakpointPositions } from "./breakpointPositions";
 
 import type {
   BreakpointOptions,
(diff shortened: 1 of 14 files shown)


darkwing:
Is this temporary or permanent?  Is this a step toward removing pause points?  My worry is the confusion that now "PossibleBreakpoints" would cause.  Maybe "PossibleBreakPosition" ?

jasonLaster:
this is permanent. when we drop pause points, this will be all that is left

darkwing:
OK, good, that makes sense.  I'd love for us to figure out a better name than "possible breakpoints" though.  "Possible" feels "inexact" or "maybe", and we use "breakpoints" as actual stored breakpoints; maybe we can "Pause positions" instead?  The verbiage just feels a bit shaky.

darkwing:
When I turn on column-breakpoints, clicking on gutters does nothing -- no breakpoint is created.

As part of this PR, maybe we should run breakpoint-centric tests with this pref both on and off.

jasonLaster:
@darkwing we should definitely turn on the pref for the tests soon...

darkwing:
I'm also seeing a white gap between column breakpoints and text when a line i paused; we can fix that with:

```diff
diff --git a/src/components/Editor/ColumnBreakpoints.css b/src/components/Editor/ColumnBreakpoints.css
index 265c7f839..fe8e08bb7 100644
--- a/src/components/Editor/ColumnBreakpoints.css
+++ b/src/components/Editor/ColumnBreakpoints.css
@@ -5,7 +5,7 @@
 .column-breakpoint {
   display: inline;
   padding: 0;
-  margin-inline-end: 4px;
+  padding-inline-end: 4px;
 }
```

darkwing:
Built, tested, working great!  Excellent work @jasonLaster !

PR #7760, opened 2019-01-19
Show command bar if paused and secondary panel hidden

### LINKED ISSUE #7667: [SecondaryPane] The command bar should be visible when the panel is collapsed
Currently, when the user collapses the secondary pane we hide the commandbar... The problem is that in this case the user can no longer access the stepping buttons. I think that it would be nice if the command bar were still visible so that the user could still step...

### Problem
screen shot 2019-01-06 at 1 31 35 pm

### Fix
screen shot 2019-01-06 at 1 22 13 pm

Garee:
## Summary

Display the command bar in the tab bar if the debugger is paused and the secondary panel is closed. This allows users to step while keeping the secondary panel closed.

Fixes #7667 

## Discussion Points

- Can the appearance be improved? Is the command bar is the wrong place here?
- The secondary panel will continue to open automatically if a breakpoint is hit and the user is not stepping.
- Should this be a opt-in via an option?

### DIFF
--- a/src/actions/pause/paused.js
+++ b/src/actions/pause/paused.js
@@ -8,7 +8,8 @@ import {
   isEvaluatingExpression,
   getSelectedFrame,
   getSources,
-  getLastCommand
+  getLastCommand,
+  isStepping
 } from "../../selectors";
 
 import { mapFrames } from ".";
@@ -41,6 +42,7 @@ export function paused(pauseInfo: Pause) {
   return async function({ dispatch, getState, client, sourceMaps }: ThunkArgs) {
     const { thread, frames, why, loadedObjects } = pauseInfo;
     const topFrame = frames.length > 0 ? frames[0] : null;
+    const didStep = isStepping(getState());
 
     // NOTE: do not step when leaving a frame or paused at a debugger statement
     if (topFrame && !why.frameFinished && why.type == "resumeLimit") {
@@ -81,7 +83,10 @@ export function paused(pauseInfo: Pause) {
       await dispatch(selectLocation(selectedFrame.location));
     }
 
-    dispatch(togglePaneCollapse("end", false));
+    if (!didStep) {
+      dispatch(togglePaneCollapse("end", false));
+    }
+
     await dispatch(fetchScopes());
 
     // Run after fetching scoping data so that it may make use of the sourcemap
--- a/src/actions/pause/paused.js
+++ b/src/actions/pause/paused.js
@@ -9,7 +9,7 @@ import {
   getSelectedFrame,
   getSources,
   getLastCommand,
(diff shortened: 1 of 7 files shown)


jasonLaster:
@Garee I like it. Couple questions

1.  how does this looks in vertical mode?
2. what is wasStepping used for?


@darkwing @mcroud what do you think?

mcroud:
Super! Can I suggest that when the comand bar is positioned within the tab bar, it appears left of the Expand Panes button, So that Expand Panes remains the rightmost button in the tab bar.

I feel this upholds the relationship between the Expand Panes button and the (currently) off-canvas panel it controls.

Garee:
@mcroud Thanks for the feedback. Sure, I'll move the command bar to the left of the expand button.

jaril:
I'm not too sure about the vertical layout. To start off with, I'd imagine the reason for somebody collapsing the right sidebar is to have more horizontal space in the code editor available. In horizontal mode this still helps with that - you get more editor horizontal space and then the toolbar takes an insignificant portion of the tabs.

For vertical mode this isn't the case. Collapsing the right sidebar doesn't give you any more horizontal code editor space. Instead it gives you more vertical space in the code editor, while taking up a noticeable portion of the tabs.

Another thing I wanted to bring up - in vertical mode the toolbar ends up appearing in two very different places in the debugger. 
- with right sidebar expanded: below the editor
- with right sidebar collapsed: above the editor

This isn't an issue in horizontal mode since the toolbar stay above the editor regardless of collapse/expand. 

This is all to say I prefer having the toolbar just stay at the bottom similar to how it was in the original 'fixed' spec. What do you think @jasonLaster?

50740089-9d76b280-11b7-11e9-95b0-13c511d203e9

jaril:
> The right collapse sidebar icon seems to have been flipped.

I was wrong here - the right collapse sidebar icon was a separate regression and has been fixed with #7802, so ignore that one!

darkwing:
@digitarald @violasong This needs UX direction to avoid surprises on your side.  How/where would you like the command bar to display when the right pane has been collapsed?

digitarald:
Really nice work, @Garee, and demo!

Thanks @darkwing for bouncing this off. No strong feelings on the solution as this isn't a critical path in the UI. There is arguments for both positions. I'd land this in its current layout and tweak it if it feels off.

darkwing:
I really like this @Garee!  Could you rebase so I can move this along?

Thank you!

Garee:
@darkwing Done, thanks. 

The `yarn flow` command is failing on CircleCI, but I think this is unrelated to this PR.

violasong:
(sorry for the late response, but this sounds great to me too :) )

PR #7766, opened 2019-01-21
[XHR Breakpoints] Clicking text toggles breakpoint

### LINKED ISSUE #7700: XHR Breakpoint items text should be label toggle
With standard breakpoints, we make clicking on the breakpoint open the source at the breakpoint's location, and thus we cannot use its label to toggle the breakpoint on and off.

XHR breakpoints don't need to go to any location, and thus clicking on the "URL contains (text)" should/could toggle the breakpoint on and off.

derek-li:
Hi all!

This PR fixes #7700. 

I added an onClick function to each XHR breakpoint list item that toggles the respective breakpoint.

I had to change the callback for the filter in removeXHRBreakpoint since the new toggle feature was altering the state of the breakpoints (when using isEqual, the **_same_** breakpoint passed in through the `action` and `state` args would now have different toggle statuses, ie. disabled vs enabled, resulting in a falsey value being returned).

In addition to this issue, updateXHRBreakpoint was being called after the breakpoint was removed. This caused an error since updateXHRBreakpoint tried to update with a now non-existent breakpoint. I resolved this through checking if the breakpoint exists when a toggle action is called and returning the state accordingly.

I plan to add tests for this PR later, as I saw @notjaril already has a comprehensive testing spec in the process of being merged.

Please let me know if there are any questions or concerns. Thanks!

GIF:
https://gyazo.com/cdcc6d2f9a36330e030af1f700484ec1

### DIFF
--- a/src/components/SecondaryPanes/XHRBreakpoints.js
+++ b/src/components/SecondaryPanes/XHRBreakpoints.js
@@ -154,7 +154,8 @@ class XHRBreakpoints extends Component<Props, State> {
       </li>
     );
   }
-  handleCheckbox = index => {
+  handleCheckbox = (index, e) => {
+    e.stopPropagation();
     const {
       xhrBreakpoints,
       enableXHRBreakpoint,
@@ -193,13 +194,13 @@ class XHRBreakpoints extends Component<Props, State> {
         key={path}
         title={path}
         onDoubleClick={(items, options) => this.editExpression(index)}
+        onClick={e => this.handleCheckbox(index, e)}
       >
         <input
           type="checkbox"
           className="xhr-checkbox"
           checked={!disabled}
-          onChange={() => this.handleCheckbox(index)}
-          onClick={ev => ev.stopPropagation()}
+          onChange={e => this.handleCheckbox(index, e)}
         />
         <div className="xhr-label">{text}</div>
         <div className="xhr-container__close-btn">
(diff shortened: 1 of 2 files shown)


src/components/SecondaryPanes/XHRBreakpoints.js
```diff
@@ -193,13 +194,13 @@ class XHRBreakpoints extends Component<Props, State> {
         key={path}
         title={path}
         onDoubleClick={(items, options) => this.editExpression(index)}
+        onClick={e => this.handleCheckbox(index, e)}
```
wldcordeiro: Could we maybe achieve this by having a `<label>` wrap the `<input>` and have a `for` attribute?

darkwing:
I agree with @wldcordeiro  -- using a `<label>` would fix this issue without JS 👍

derek-li:
Wow! That's so much simpler. Guess I'm still pretty unfamiliar with HTML, haha. 

Thanks for the direction :) @wldcordeiro @darkwing

src/components/SecondaryPanes/XHRBreakpoints.js
derek-li: Added changes. No `for` attribute needed since association is implicit.

derek-li:
Updated. 

GIF: https://gyazo.com/d02c7085561f071f2a4114fc658e5644

PR #7770, opened 2019-01-21
Run mozilla-central Frames dependent tests on Travis.

nchevobbe:
This adds all the smartrace component tests, as
well as copy message/export all context menu entries tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -64,4 +64,9 @@ script:
     browser_webconsole_location_debugger_link.js
     browser_webconsole_stacktrace_location_debugger_link.js
     browser_application_panel_debug-service-worker.js
+    browser_webconsole_sourcemap_nosource
+    browser_webconsole_context_menu_copy_entire_message.js
+    browser_webconsole_context_menu_export_console_output_clipboard.js
+    browser_webconsole_context_menu_copy_message_with_framework_stacktrace.js
+    test_smart-trace
   - ./node_modules/.bin/mochii --ci --mc ./firefox --headless devtools/client/debugger/new


nchevobbe:
Okay, tests that are expected to fail **do** fail:

```
failed test:  devtools/client/shared/components/test/mochitest/test_smart-trace-grouping.html 
failed test:  devtools/client/shared/components/test/mochitest/test_smart-trace-source-maps.html 
failed test:  devtools/client/webconsole/test/mochitest/browser_webconsole_context_menu_copy_entire_message.js 
```
How should we proceed from there @jasonLaster ? Landing this will cause all the PR to fail (and would need maintainer to look at each travis job manually). I feel like we should wait until the fixed version of those tests land in m-c.
What's your take on this?

jasonLaster:
I think that's right, when the fixes are in we can bump the travis.yml MC sha

nchevobbe:
Patch landed on mc. I updated the `MC_COMMIT` here, so let's see if travis is okay with those changes now.

jasonLaster:
hmm, travis crashed... do we need anything else in github?

jasonLaster:
just landed - https://github.com/devtools-html/debugger.html/pull/7850 which should make this easier

nchevobbe:
```
devtools/client/debugger/new/test/mochitest/browser_dbg-breakpoints-cond.js | A promise chain failed to handle a rejection: bpClient.setCondition is not a function - stack: setBreakpointOptions devtools/client/debugger/new/src/client/firefox/commands.js 215:10
```

re-built a few time but this is still happening. Is this because of the `MC_COMMIT` change ?

PR #7776, opened 2019-01-22
Fix a couple bugs with windowless worker debugging.

### LINKED ISSUE #7710: [tracking] Worker Debugging
Server-side meta bug: [1520698](https://bugzilla.mozilla.org/show_bug.cgi?id=1520698).

#### Todos
- [x] refreshing clears selected source #7708 
- [ ] [Workers] Use one tree instead of one tree per thread #8005 
- [x] remove sources when a worker is removed #7660

bhackett1024:
Addresses #7710 part 2

### Summary of Changes

* Fixes bug where refreshing a page that pauses at startup and has workers ends up in a messed up state where pause information is shown incorrectly.  The problem is that updating the redux state after a thread resumes updated the state of the current thread, instead of the thread that actually resumed.

* Fixes bug where loading devtools where persistent state referred to a worker source could access an undefined source object and throw an exception that prevents the debugger from loading.

### Test Plan

Manually verified issues were fixed.  These are hard to write tests for unfortunately :/

### DIFF
--- a/src/actions/pause/resumed.js
+++ b/src/actions/pause/resumed.js
@@ -4,11 +4,12 @@
 
 // @flow
 
-import { getCurrentThread, isStepping, getPauseReason } from "../../selectors";
+import { isStepping, getPauseReason } from "../../selectors";
 import { evaluateExpressions } from "../expressions";
 import { inDebuggerEval } from "../../utils/pause";
 
 import type { ThunkArgs } from "../types";
+import type { ResumedPacket } from "../../client/firefox/types";
 
 /**
  * Debugger has just resumed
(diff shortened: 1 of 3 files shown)


src/actions/tabs.js
```diff
@@ -106,6 +106,7 @@ export function closeTabsForMissingThreads(workers: Worker[]) {
       if (sourceId) {
         const source = getSourceFromId(getState(), sourceId);
         if (
+          source &&
```
jasonLaster: I think we should use getSource instead of getSourceFromId. This way we're telling flow source can be null.

src/actions/pause/resumed.js
```diff
@@ -4,26 +4,26 @@
...
-export function resumed() {
+export function resumed(packet: ResumedPacket) {
   return async ({ dispatch, client, getState }: ThunkArgs) => {
     const why = getPauseReason(getState());
     const wasPausedInEval = inDebuggerEval(why);
     const wasStepping = isStepping(getState());
-    const thread = getCurrentThread(getState());
 
-    dispatch({ type: "RESUME", thread });
+    dispatch({ type: "RESUME", thread: packet.from });
```
jasonLaster: Nice catch.

src/actions/tabs.js
bhackett1024: Sure, just pushed a fix for this.  Having these two methods with identical implementations and different return types is kind of silly though, I'll try to fix this as part of bug 1518884.

jasonLaster: agreed... `getSourceFromId` was a bad idea :/

PR #7779, opened 2019-01-22
[DebugLine] line is not cleared

### LINKED ISSUE #7755: [DebugLine] occassionally the debug line is not cleared
I've found a couple times that the debug line does not go away when you resume. Some of my theories are 

1. it is no longer the selected location
2. you were currently selecting another frame, that was not the top frame
3. you resumed while viewing another document/another source...

screen shot 2019-01-14 at 1 22 11 pm
screen shot 2019-01-18 at 2 59 55 pm

jasonLaster:
Fixes #7755

### DIFF
--- a/src/components/Editor/DebugLine.js
+++ b/src/components/Editor/DebugLine.js
@@ -45,7 +45,7 @@ export class DebugLine extends Component<Props> {
     const { why, frame, source } = this.props;
 
     startOperation();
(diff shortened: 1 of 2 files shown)


darkwing:
I like your theory behind this and the code is solid.  I can't easily reproduce it so let's see how this goes.

PR #7781, opened 2019-01-23
Change max-width boundaries for verticalLayoutBreakpoint

### LINKED ISSUE #7765: Both horizontalLayoutBreakpoint and verticalLayoutBreakpoint can match at the same time
If the window is exactly 800px wide, both breakpoints here will match:

```js
const horizontalLayoutBreakpoint = window.matchMedia("(min-width: 800px)");
const verticalLayoutBreakpoint = window.matchMedia(
  "(min-width: 10px) and (max-width: 800px)"
);
```

Media queries match at the boundaries, both for `min-width` and `max-width`.

One way to work around this is to make sure you use slightly different values (a 1px or even 0.1px gap usually works). As in:

```js
const horizontalLayoutBreakpoint = window.matchMedia("(min-width: 800px)");
const verticalLayoutBreakpoint = window.matchMedia(
  "(min-width: 10px) and (max-width: 799.9px)"
);
```

pradeepgangwar:
Fixes #7765 

### Summary of Changes

* Changing the max-width boundaries for horizontalLayoutBreakpoint and verticalLayoutBreakpoint so that they do not clash.

### DIFF
--- a/src/components/App.js
+++ b/src/components/App.js
@@ -35,7 +35,7 @@ const isMacOS = appinfo.OS === "Darwin";
 
 const horizontalLayoutBreakpoint = window.matchMedia("(min-width: 800px)");
 const verticalLayoutBreakpoint = window.matchMedia(
-  "(min-width: 10px) and (max-width: 800px)"
+  "(min-width: 10px) and (max-width: 799px)"
 );
 
 import "./variables.css";


darkwing:
I love it @pradeepgangwar !  Thank you!

PR #7796, opened 2019-01-24
Split Source into Source/SourceActor, and Breakpoint into Breakpoint/…

bhackett1024:
### Summary of Changes

* Add a SourceActor object representing a SourceClient connected to a source in a thread.
* Change Source to be independent of threads, and provide a mapping Source -> SourceActor[] in redux state.
* Add a BreakpointActor object representing a BreakpointClient, and which is associated with a SourceActor instead of a Source (as Breakpoint is).

This has the general effect of making sources and breakpoints in the client thread-independent, as was the case before windowless worker debugging landed.  More background on this change can be found in bug 1518884.

### Test Plan

Passes unit tests.
Manually tested on TodoMVC.

There are still some holes in the implementation wrt threads.  Mainly, as sources from new threads appear, BreakpointActors will not be created if there are applicable sites for existing breakpoints in those sources.  I'd like to fix this in followup, as there are a fair number of cleanups to client breakpoint state that can be performed once this lands.

### DIFF
--- a/src/actions/breakpoints/syncBreakpoint.js
+++ b/src/actions/breakpoints/syncBreakpoint.js
@@ -8,13 +8,14 @@ import {
   createBreakpoint,
   assertBreakpoint,
   assertPendingBreakpoint,
-  findScopeByName
+  findScopeByName,
+  makeSourceActorLocation
 } from "../../utils/breakpoint";
 
 import { getGeneratedLocation } from "../../utils/source-maps";
 import { getTextAtPosition } from "../../utils/source";
 import { originalToGeneratedId, isOriginalId } from "devtools-source-map";
-import { getSource } from "../../selectors";
+import { getSource, getSourceActors } from "../../selectors";
 import type { ThunkArgs, Action } from "../types";
 
 import type {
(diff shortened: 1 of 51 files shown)


src/utils/breakpoint/index.js
```diff
@@ -45,6 +47,8 @@ export function locationMoved(
   );
 }
 
+// The ID for a Breakpoint is derived from its location in its Source.
+// FIXME rename this to makeBreakpointId.
```
darkwing: Just calling this FIXME out.

jasonLaster:
I'm seeing this when i have a breakpoint in todo-view.js#34 and reload

screen shot 2019-01-25 at 12 31 57 pm

src/actions/breakpoints/syncBreakpoint.js
```diff
@@ -121,18 +120,32 @@ export async function syncBreakpointPromise(
...
-    const id = pendingBreakpoint.disabled ? "" : existingClient.id;
+  if (pendingBreakpoint.disabled || isSameLocation) {
+    // Make sure the breakpoint is installed on all source actors.
+    if (!pendingBreakpoint.disabled) {
+      for (const sourceActor of sourceActors) {
+        const sourceActorLocation = makeSourceActorLocation(
+          sourceActor,
+          generatedLocation
+        );
+        if (!client.getBreakpointByLocation(sourceActorLocation)) {
```
jasonLaster: why is this needed?

PR #7803, opened 2019-01-25
Initialize a breakpoint as loading true - redo of PR 7644

Art-Vanderlay:
Redo of #7644
Fixes bug #6771


bomsy:
@Art-Vanderlay thanks ... will take a look in a bit

darkwing:
Test failure is legit; if I create a breakpoint, then refresh, the breakpoint is gone.

@Art-Vanderlay Here's an idea.  Let's audit if `loading` is actually being used anywhere.  If it isn't, we could just remove it.

darkwing:
I'm going to park this one @Art-Vanderlay; it's good work but we're restructuring breakpoints work and we'll incorporate this fix as part of the larger change.  Thank you!

PR #7810, opened 2019-01-25
Fix 6139 - Handle longString symbols

darkwing:
I think the code structure still needs changing but it's far enough along to get @nchevobbe 's eyes.


packages/devtools-reps/src/reps/stubs/symbol.js
```diff
@@ -14,4 +14,10 @@ stubs.set("SymbolWithoutIdentifier", {
   actor: "server1.conn1.child1/symbol2"
 });
 
+stubs.set("SymbolWithLongString", {
+  type: "symbol",
+  actor: "server1.conn1.child1/symbol1",
+  name: "aa".repeat(10000)
```
darkwing: I think this needs to be an object of `longString` type

nchevobbe: yes, you're right

packages/devtools-reps/src/reps/symbol.js
```diff
@@ -21,12 +24,28 @@ function SymbolRep(props) {
...
+    symbolText = "";
+  } else if (name.type && name.type === "longString") {
+    symbolText = StringRep({
+      object: name || "",
+      shouldCrop: true,
+      cropLimit: MAX_STRING_LENGTH
+    });
+  } else {
+    const { Rep } = require("./rep");
+    symbolText = Rep({ object: name });
```
nchevobbe: here `name` should be a string, so we can directly assign it to `symbolText`.

nchevobbe:
Tests are failing because using `Rep` instead of a string adds quotes around the symbol's name.
The change I suggested should get rid of those quotes. For the case when we have a longString, and using `StringRep` is what we actually want, we can pass `useQuotes: false` to not have the quotes as well.

nchevobbe:
(and thanks for working on this issue!)

packages/devtools-reps/src/reps/symbol.js
darkwing: Are you suggesting coercing `name` to a string (i.e. `String(name)` or `name+""` or using the String rep?

darkwing:
Merging myself as both seem to approve.

PR #7811, opened 2019-01-26
Fix logpoints bug

bomsy:
#### Issues
* The log points are shown as conditional breakpoints
* On editing the log point nothing shows up in the conditional breakpoint panel

logpoints1

* The context menu for a log point should not show any item relating to conditional breakpoints and vice versa
screenshot 2019-01-25 at 21 51 05

### Summary of Changes

* change  log points color to purple
* cleanup unused menu items
* fix `logValue`

### Test Plan

Example test plan:

- [x] about:config
- [x] set `devtools.debugger.features.log-points` to `true`
- [x] right click editor gutter menu

logpoints3

### DIFF
--- a/assets/panel/debugger.properties
+++ b/assets/panel/debugger.properties
@@ -534,6 +534,12 @@ editor.addLogPoint.accesskey=l
 # LOCALIZATION NOTE (editor.editLogPoint): Editor gutter context menu item
 # for editing a log point already set on a line.
 editor.editLogPoint=Edit log
+editor.editLogPoint.accesskey=E
+
+# LOCALIZATION NOTE (editor.removeLogPoint): Context menu item for removing
+# a log point on a line.
+editor.removeLogPoint.label=Remove log
+editor.removeLogPoint.accesskey=V
 
 # LOCALIZATION NOTE (editor.conditionalPanel.placeholder): Placeholder text for
 # input element inside ConditionalPanel component
--- a/src/components/Editor/Breakpoint.js
+++ b/src/components/Editor/Breakpoint.js
@@ -57,6 +57,7 @@ class Breakpoint extends PureComponent<Props> {
     doc.setGutterMarker(line, "breakpoints", null);
     doc.removeLineClass(line, "line", "new-breakpoint");
     doc.removeLineClass(line, "line", "has-condition");
+    doc.removeLineClass(line, "line", "has-log");
   }
 
   get selectedLocation() {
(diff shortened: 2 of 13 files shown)


jasonLaster:
Nice. I like this!

darkwing:
Awesome @bomsy !  A few things I noticed:

1.  Create a conditional breakpoint still shows purple:

conditionalcolor

2.  In this PR it would also be sweet if we could fix the context menu for the breakpoints pane; we still show "Add condition":

conditioncontext

It seems like it fits in with your other context menu fix.

3.  Test failure looks legit, I think because we removed the context menu item.

bomsy:
Thanks for reviewing @darkwing. Will update asap.

jasonLaster:
Okay, this ux is definitely too vague... I took a stab at handling a couple of cases:

1. context menus (gutter, column, panel)
2. marker colors
3. panel bp text

The heuristics I used were:

1. treat logs and conditions as independent (can be added / removed / edited)
1. give precedence to logs (highlight purple, show bp text)

### Adding a new bp
screen shot 2019-02-10 at 9 35 44 am

### with a condition
screen shot 2019-02-10 at 9 36 20 am

### with a log
screen shot 2019-02-10 at 9 37 47 am

### with both set

screen shot 2019-02-10 at 9 38 02 am

assets/panel/debugger.properties
```diff
@@ -534,6 +534,13 @@ editor.addLogPoint.accesskey=l
 # LOCALIZATION NOTE (editor.editLogPoint): Editor gutter context menu item
 # for editing a log point already set on a line.
 editor.editLogPoint=Edit log
+editor.editLogPoint.accesskey=e
```
flodolo: `E` (better to respect the case)

assets/panel/debugger.properties
```diff
@@ -534,6 +534,13 @@ editor.addLogPoint.accesskey=l
...
+
+
+# LOCALIZATION NOTE (editor.removeLogPoint): Context menu item for removing
+# a log point on a line.
+editor.removeLogPoint=Remove log
```
flodolo: nit: can we call this `editor.removeLogPoint.label`?

assets/panel/debugger.properties
```diff
@@ -534,6 +534,13 @@ editor.addLogPoint.accesskey=l
...
+editor.removeLogPoint.accesskey=p
```
flodolo: Why the external accesskey (not available in the label)?

jasonLaster: i don't follow?

flodolo: Label "Remove log", there's no "p", which will result in "Remove log (p)" (with p underscored) on Windows and Linux.

PR #7812, opened 2019-01-26
Add black box option to context menu for source tree items and source header

### LINKED ISSUE #7287: Context menu Source actions should be available from the SourcePanel and Source Tab

derek-li:
Fixes #7287 

Summary of changes:
- Adds black box option to source header and source items in the tree

Demo:
Source header GIF: https://gyazo.com/441defadd22762e5c3142011241b8e8b
Source tree GIF: https://gyazo.com/5c8354f69db18a3ffd226448d8334006


src/components/Editor/Tab.js
```diff
@@ -135,13 +137,24 @@ class Tab extends PureComponent<Props> {
       }
     ];
 
-    items.push({
-      item: {
-        ...tabMenuItems.prettyPrint,
-        click: () => togglePrettyPrint(tab),
-        disabled: !shouldShowPrettyPrint(source)
+    items.push(
```
darkwing: I think we want to add `shouldBlackbox` here since we use it for other places.

darkwing:
Functionally it works well, just implement `shouldBlackBox` and I think we will be set!

src/components/Editor/Tab.js
derek-li: Added the changes!

derek-li:
Appreciate the review @darkwing :)

darkwing:
@derek-li Looks like you need to do a `yarn jest -u` to update snapshots, then I think we're good!

derek-li:
Hey @darkwing, ran the jest updates, but it looks like more CircleCI tests are failing than before. Do you have any input?

derek-li:
Not sure what changed, but the black box option on the sourcestreeitem context menu stopped working. Currently investigating.

darkwing:
How's the investigation going?  I noticed that the blackbox source context menu item is always disabled, so we need a bit more work here.

derek-li:
Thanks for checking in. Sorry for being a bit slow, still a noob :)

I found that the sources prop in the sourcesTree is passing in state.sources.relativeSources. The problem with that is relativeSources doesn't seem up to date with state.sources.sources.

So even when a source in state.sources.sources has attributes `loadedState: "loaded"` and `isBlackBoxed: true`, we get this as the sources prop:
0d13964fd413ec4da49c1b0602d9d12e

which results in the button being disabled, as well as the blackbox icon not being rendered in the sources tree panel.

I'm currently thinking of potential solutions and am strongly considering using another selector. Do you know why we are using relativeSources in this case?

derek-li:
@darkwing I think this should be good to merge. 
There's just the issue with the inexact source type. Do you know what's causing the error?

jasonLaster:
looks like the flow issues are unrelated. I believe rebasing should fix them

derek-li:
Okay looking into it now.

darkwing:
@derek-li The source flow issue can be fixed with:

```diff
$ git diff
diff --git a/src/components/PrimaryPanes/SourcesTreeItem.js b/src/components/PrimaryPanes/SourcesTreeItem.js
index ac5e58017..2f64d2d9f 100644
--- a/src/components/PrimaryPanes/SourcesTreeItem.js
+++ b/src/components/PrimaryPanes/SourcesTreeItem.js
@@ -246,7 +246,7 @@ class SourceTreeItem extends Component<Props, State> {
         className={classnames("node", { focused })}
         key={item.path}
         onClick={this.onClick}
-        onContextMenu={e => this.onContextMenu(e, item, source)}
+        onContextMenu={e => source && this.onContextMenu(e, item, source)}
       >
         {this.renderItemArrow()}
         {this.getIcon(item, depth)}
```

derek-li:
Thanks for the help guys!

Re based and added the tweak

PR #7819, opened 2019-01-27
Don't wait to load already-loaded source maps

bhackett1024:
This addresses an issue that was exposed by testing PR #7796.  When sources are added to the debugger, source maps for those sources are loaded asynchronously --- the newSources() redux action will finish before the loads have finished.  This is good, because the loads could take an extremely long time if they 404, but causes problems when another devtool (i.e. the console) is trying to open up an original source that has already been mapped by the source map worker.  After loading the debugger the original sources might not be loaded yet.  This patch deals with this by having the worker keep track of source map requests it has finished loading.  The newSources() redux action does not finish until all sources with one of these maps have had that map loaded.  Note that due to how this is implemented a source can go through loadSourceMaps twice --- things are a little cleaner this way.

### Summary of Changes

* Keep track of loaded source maps in the source map worker.
* Don't finish newSources() redux action until existing source maps have been loaded.

### DIFF
--- a/packages/devtools-source-map/src/index.js
+++ b/packages/devtools-source-map/src/index.js
@@ -37,6 +37,7 @@ const applySourceMap = dispatcher.task("applySourceMap");
 const clearSourceMaps = dispatcher.task("clearSourceMaps");
 const hasMappedSource = dispatcher.task("hasMappedSource");
 const getOriginalStackFrames = dispatcher.task("getOriginalStackFrames");
+const hasLoadedSourceMap = dispatcher.task("hasLoadedSourceMap");
 
 module.exports = {
   originalToGeneratedId,
@@ -56,6 +57,7 @@ module.exports = {
   applySourceMap,
   clearSourceMaps,
   getOriginalStackFrames,
+  hasLoadedSourceMap,
   startSourceMapWorker(url: string, assetRoot: string) {
     dispatcher.start(url);
     setAssetRootURL(assetRoot);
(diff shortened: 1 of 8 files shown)


jasonLaster:
I think this can be done a bit simpler:

1. in view-source check to see if the source map worker has seen the url
2. in the worker, have a new function that loops over loaded requests and checks the content of a map

```diff
diff --git a/devtools/client/shared/view-source.js b/devtools/client/shared/view-source.js
index 117d578fe9b5..7e20fc345404 100644
--- a/devtools/client/shared/view-source.js
+++ b/devtools/client/shared/view-source.js
@@ -51,7 +51,8 @@ exports.viewSourceInStyleEditor = async function(toolbox, sourceURL,
 exports.viewSourceInDebugger = async function(toolbox, sourceURL, sourceLine,
                                               reason = "unknown") {
   const dbg = await toolbox.loadTool("jsdebugger");
-  const source = dbg.getSource(sourceURL);
+  const source = dbg.getSource(sourceURL)
+    || toolbox.sourceMapService.hasOriginalURL(sourceURL);
   if (source) {
     await toolbox.selectTool("jsdebugger", reason);
     dbg.selectSource(sourceURL, sourceLine);
```

```diff
diff --git a/packages/devtools-source-map/src/source-map.js b/packages/devtools-source-map/src/source-map.js
index 8eac016bb..f549fe17a 100644
--- a/packages/devtools-source-map/src/source-map.js
+++ b/packages/devtools-source-map/src/source-map.js
@@ -21,7 +21,8 @@ const {
+async function hasOriginalURL(originalURL: string) {
+  for (const request of sourceMapRequests) {
+    if (request.content && request.urls.includes(originalURL)) {
+      return true;
+    }
+  }
+  return false;
+}
+
 const COMPUTED_SPANS = new WeakSet();
```

PR #7823, opened 2019-01-28
Fix 7671 - shift+click should add a disabled breakpoint

### LINKED ISSUE #7671: [breakpoints] shift+click should add a disabled breakpoint

darkwing:
Fixes #7671 

Supercedes: https://github.com/devtools-html/debugger.html/pull/7694

### DIFF
--- a/src/actions/breakpoints/addBreakpoint.js
+++ b/src/actions/breakpoints/addBreakpoint.js
@@ -101,7 +101,7 @@ async function addBreakpointPromise(getState, client, sourceMaps, breakpoint) {
 
   const newBreakpoint = {
     id: makeBreakpointId(generatedLocation),
-    disabled: false,
+    disabled: breakpoint.disabled,
     loading: false,
     options: breakpoint.options,
     location: newLocation,
--- a/src/actions/breakpoints/addBreakpoint.js
+++ b/src/actions/breakpoints/addBreakpoint.js
@@ -34,7 +34,13 @@ import type {
 } from "../../types";
 import type { ThunkArgs } from "../types";
 
-async function addBreakpointPromise(getState, client, sourceMaps, breakpoint) {
+async function addBreakpointPromise(
+  getState,
+  client,
+  sourceMaps,
+  breakpoint,
+  shouldDisable = false
+) {
   const state = getState();
   const source = getSource(state, breakpoint.location.sourceId);
 
@@ -101,7 +107,7 @@ async function addBreakpointPromise(getState, client, sourceMaps, breakpoint) {
 
   const newBreakpoint = {
     id: makeBreakpointId(generatedLocation),
-    disabled: breakpoint.disabled,
+    disabled: shouldDisable,
     loading: false,
     options: breakpoint.options,
     location: newLocation,
@@ -166,11 +172,18 @@ export function addBreakpoint(
     }
 
     const breakpoint = createBreakpoint(location, options);
+    const shouldDisable = options && options.disabled === true;
 
     return dispatch({
       type: "ADD_BREAKPOINT",
       breakpoint,
-      [PROMISE]: addBreakpointPromise(getState, client, sourceMaps, breakpoint)
+      [PROMISE]: addBreakpointPromise(
+        getState,
+        client,
+        sourceMaps,
+        breakpoint,
+        shouldDisable
+      )
     });
   };
 }
--- a/src/actions/breakpoints/index.js
+++ b/src/actions/breakpoints/index.js
@@ -364,7 +364,10 @@ export function toggleBreakpointAtLine(line: number) {
   };
 }
 
-export function addBreakpointAtLine(line: number) {
+export function addBreakpointAtLine(
+  line: number,
+  shouldDisable: boolean = false
+) {
   return ({ dispatch, getState, client, sourceMaps }: ThunkArgs) => {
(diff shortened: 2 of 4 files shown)


darkwing:
@baronKilljoy Let's iterate on this.  We should add a test for sure.  I'd love you to manually test this too :)

baronKilljoy:
@darkwing is there another fix mixed here, the base code is different than mine. What is the !! in !!ev.shiftKey. If shiftkey isn't pressed ev.shiftKey is undefined not false.

baronKilljoy:
@darkwing without my other changed to addBreakpointPromise the checkbox on the secondary box breaks. here is an alternate fix to get rid of the need for the redundant code. 
altfix
The trouble is `export function addBreakpoint(shiftKeyPress,
  location: SourceLocation,
  { condition, hidden }: addBreakpointOptions = {}
) `
addBreakpoint options having a default value means we need to add the bool shiftKeypress cant be defaulted. addBreakpoint ill need to be changed everywhere its called

darkwing:
This is the weirdness that tests caught: oO

wtfbreakpoints

When I click the checkbox, not only does the breakpoint not get enabled, it shifts locations...

src/actions/breakpoints/addBreakpoint.js
```diff
@@ -101,7 +101,7 @@ async function addBreakpointPromise(getState, client, sourceMaps, breakpoint) {
 
   const newBreakpoint = {
     id: makeLocationId(generatedLocation),
-    disabled: false,
+    disabled: breakpoint.disabled,
```
darkwing: `breakpoint` is the data with options, which is why I'm using its `disabled` property.  We can't simply default this to `false`

darkwing:
@jasonLaster Can you think of why initializing a breakpoint to `disabled: true` would cause location shifting and likewise weirdness?

darkwing:
@bomsy @AnshulMalik @princiya @jainsneha23  I'd love your eyes on this as well.

src/actions/breakpoints/addBreakpoint.js
baronKilljoy: newBreakpoint needs to be false or the checkbox breaks.  thats why i manually set it to false if the shiftkey is pressed in my code.

jasonLaster:
@darkwing nope

darkwing:
I did some very advanced console.logging here (/me hides) and found something really weird:

what in the

When clicking on the checkbox to enable the breakpoint, the breakpoint's `id` changes, removing `originalSource` and then losing the line number at the end.

src/actions/breakpoints/addBreakpoint.js
darkwing: Right; this is the current problem; it's as though this is resetting the value somehow

darkwing:
This is a step further in the right direction.  There's only one case that's now causing weirdness:

1.  Shift-click to create a disabled breakpoint
2.  Click its checkbox in the breakpoints pane

Expected:

Breakpoint is enabled

Actual:

The breakpoint is disappearing completely, and `dbg.selectors.getBreakpointsList()` returns an empty array.  oO

baronKilljoy:
@jaril  said some thing that may be relevant in the original bug.
> My understanding is the client doesn't retain whether a breakpoint is disabled or not. It only either 1) sets breakpoints or 2) removes breakpoints to pause on. This is in `/src/client/firefox/commands.js`.
> 
> When a breakpoint (that's registered in the redux store and displayed in the Breakpoints pane) is skipped, it's because that breakpoint has been removed from the separate list of breakpoints set for the client to look at and pause on. What looks like toggling enable/disable breakpoints in the UI actually adds/removes breakpoints on the client.

>

PR #7827, opened 2019-01-28
Use `"".test(regex)` instead of `"".match(regex)` when only a test is being performed

jswalden:


jasonLaster:
Hmm, I'm seeing some issues with test not being a function for strings.

jswalden:
Whoops, my bad, had the wrong ordering of `this` and argument.  Flip 'em and things should work properly.

darkwing:
I pushed a fix for the linting issue; if this goes green, we'll be good.  Thank you @jswalden !

Issue #7797, opened 2019-01-25
Refactor unit tests


bhackett1024:
While working on PR #7796 I had a lot of trouble updating the unit tests for the refactoring being performed.  Many changes needed to be made to unit tests that were only tangentially related to the refactoring, and these changes were mostly non-obvious tweaks to ad hoc objects in the tests.  I think some changes can be incrementally made to the unit tests to make them friendlier to future refactorings and other code changes:

1. Run flow in all test files.  A lack of type checking in the test files makes it hard to see what tests need to change when internal types and interfaces change.

2. Avoid creating ad hoc object literals representing internal state.  This should be done by a shared set of testing methods for creating these objects with a complete set of properties.  This is pretty much necessitated by 1. above, as literals like the below Frame from annotateFrames.spec.js are usually missing almost all properties which flow will insist on.

{ source: { url: "https://cdnjs.cloudflare.com/ajax/libs/angular/angular.js" } }

3. More complex unit tests, mainly those operating on the redux state, should use high level debugger operations simulating user interactions, similar to what is done in debugger mochitests.  Currently, it is possible for a unit test to create situations that cannot occur in practice, like having an original source with no corresponding generated source.  When a code change bumps into these cases they are very hard to fix.  Having a nice interface for interacting with a complete debugger in a unit test would make it easier to write unit tests instead of mochitests when fixing bugs or developing complex new debugger features.  This would be nice, as unit tests are much easier to run and to debug than mochitests.

PR #7828, opened 2019-01-29
Use flow in all remaining tests

bhackett1024:
Partial fix for #7797

### Summary of Changes

* Use flow in all remaining tests under src/


jasonLaster:
I see a couple of issues with travis that are related to the Source Actor PR
https://travis-ci.org/devtools-html/debugger.html/builds/485692195?utm_source=github_status&utm_medium=notification

PR #7829, opened 2019-01-29
[test] disable browser_webconsole_sourcemap_nosource

jasonLaster:
Temporarily disable `browser_webconsole_sourcemap_nosource` while we wait for the view-source fix to land in MC

### DIFF
--- a/test/mochitest/browser.ini
+++ b/test/mochitest/browser.ini
@@ -678,7 +678,7 @@ skip-if = os == "win" # Bug 1448523, Bug 1448450
 [browser_dbg-breakpoints-cond.js]
 [browser_dbg-breakpoints-duplicate-functions.js]
 [browser_dbg-browser-content-toolbox.js]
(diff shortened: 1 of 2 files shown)


test/mochitest/browser.ini
```diff
@@ -678,7 +678,7 @@ skip-if = os == "win" # Bug 1448523, Bug 1448450
 [browser_dbg-breakpoints-cond.js]
 [browser_dbg-breakpoints-duplicate-functions.js]
 [browser_dbg-browser-content-toolbox.js]
-skip-if = !e10s || verify # This test is only valid in e10s
+skip-if = true
```
jasonLaster: this test fails if you run 

```
./mach mochitest --headless --repeat 6 browser-content-toolbox
```

It looks like `addBreakpoint` will try to call client.setBreakpoint for multiple source actors, with i think the same thread. The result is a breakpoint is added on line 5 when it should be line 3.

PR #7834, opened 2019-01-29
Don't map bindings when not paused

nchevobbe:
`mapExpression` used to try to map bindings everytime, even when the debugger wasn't paused, which causes some issue ([See 1494318](https://bugzilla.mozilla.org/show_bug.cgi?id=1494318)).

With this PR, `mapExpression` will map the bindings only when paused, which brings other issue, in `mapAwait`.

In `mapAwait`, we wrap the expression into an async iife in order to get a Promise to listen for. This was working great when the bindings are mapped (because variables were put as `self` properties).
When we don't bind, we have an issue with assignments.

Say the user want to do `let a = await 123`, we can't simply return 
```js
(async () => {
  let a = await 123;
})()
```
because then the user won't have access to `a`.

What we want instead is: 
```js
let a;
(async () => {
  return a = await 123;
})()
```

So we traverse the AST to pull out variable declaration outside of the async iife.
This causes another issue with `const` variables, as we can't re-assign them.
For now, I decided to switch them to `let` to not block us. I don't know if there could be unwanted behavior (on top of being able to re-assign to what the user think is a `const`).

### DIFF
--- a/src/actions/expressions.js
+++ b/src/actions/expressions.js
@@ -13,7 +13,8 @@ import {
   getSourceFromId,
   getSelectedSource,
   getSelectedScopeMappings,
-  getSelectedFrameBindings
+  getSelectedFrameBindings,
+  isPaused
 } from "../selectors";
 import { PROMISE } from "./utils/middleware/promise";
 import { wrapExpression } from "../utils/expressions";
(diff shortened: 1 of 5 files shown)


nchevobbe:
I think it's still missing a mochitest to ensure we don't map the bindings when not paused, but I'm unsure if I should add it in the debugger or in the console yet.

nchevobbe:
yay, all green :)

darkwing:
@nchevobbe Is this ready to go in or are we waiting?

nchevobbe:
I think it's good to go, I'm writing mochitests in console

PR #7836, opened 2019-01-29
Fix 7835 - Focus on inputs if plus is clicked but panel is closed

### LINKED ISSUE #7835: Clicking on the "Add watch expression" does not focus the input if the panel is collapsed
**Steps to reproduce**
1. Go to `data:text/html,<meta charset=utf8>`
1. Open the debugger
1. Ensure the `Watch Expressions` panel is collapsed and empty
1. Click on the "Add watch expression (`+`) button

**Expected results**

The panel is expanded, and the input where I can type my expression is focused

**Actual results**

The panel is expanded, but the input isn't focused, even if I click on the `+` button again.

darkwing:
Fixes #7835

### DIFF
--- a/src/components/SecondaryPanes/Expressions.js
+++ b/src/components/SecondaryPanes/Expressions.js
@@ -71,9 +71,16 @@ class Expressions extends Component<Props, State> {
   }
 
   componentDidMount() {
-    const { expressions, evaluateExpressions } = this.props;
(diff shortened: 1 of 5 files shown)


nchevobbe:
Looks good David, thanks for fixing this.
Should we add mochitests to ensure we don't regress?

PR #7837, opened 2019-01-29
Add context menus for handling breakpoints on a line

darkwing:
It's the last item here:  https://github.com/devtools-html/debugger.html/issues/7241

Adds context menu items for disabling/enabling/removing all breakpoints *on a line*.

## To Do
- [x] Add localization strings and accesskeys
- [x] Conditionally add menu items based on column breakpoints pref
- [x] Disable menu item if no breakpoints are enabled/disabled on the line
- [ ] Mochitests

### DIFF
--- a/assets/panel/debugger.properties
+++ b/assets/panel/debugger.properties
@@ -406,9 +406,9 @@ breakpointMenuItem.removeCondition.accesskey=c
 breakpointMenuItem.editCondition.label=Edit breakpoint condition
 breakpointMenuItem.editCondition.accesskey=n
 breakpointMenuItem.disableAllAtLine.label=Disable breakpoints on line
-breakpointMenuItem.disableAllAtLine.accesskey=Z
+breakpointMenuItem.disableAllAtLine.accesskey=K
 breakpointMenuItem.enableAllAtLine.label=Enable breakpoints on line
-breakpointMenuItem.enableAllAtLine.accesskey=V
+breakpointMenuItem.enableAllAtLine.accesskey=L
 breakpointMenuItem.removeAllAtLine.label=Remove breakpoints on line
 breakpointMenuItem.removeAllAtLine.accesskey=X
 
--- a/assets/panel/debugger.properties
+++ b/assets/panel/debugger.properties
@@ -406,9 +406,10 @@ breakpointMenuItem.removeCondition.accesskey=c
 breakpointMenuItem.editCondition.label=Edit breakpoint condition
 breakpointMenuItem.editCondition.accesskey=n
 breakpointMenuItem.disableAllAtLine.label=Disable breakpoints on line
+breakpointMenuItem.disableAllAtLine.accesskey=d
 breakpointMenuItem.enableAllAtLine.label=Enable breakpoints on line
-breakpointMenuItem.removeAllAtLine.label=Remove breakpoints on line
 breakpointMenuItem.enableAllAtLine.accesskey=l
+breakpointMenuItem.removeAllAtLine.label=Remove breakpoints on line
 breakpointMenuItem.removeAllAtLine.accesskey=x
 
 # LOCALIZATION NOTE (breakpoints.header): Breakpoints right sidebar pane header.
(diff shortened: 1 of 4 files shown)


darkwing:
@digitarald Current, very basic look:

breakpointcontext

How do we want to handle enabling / disabling these new items, as well as existing items?

- What if there's only one breakpoint on the line ; do we want to enable the removeAll/enableAll/disableAll item?
- What order should these be in?

You general thoughts would be good.

src/utils/prefs.js
```diff
@@ -55,7 +55,7 @@ if (isDevelopment()) {
   pref("devtools.debugger.features.remove-command-bar-options", true);
   pref("devtools.debugger.features.code-folding", false);
   pref("devtools.debugger.features.outline", true);
-  pref("devtools.debugger.features.column-breakpoints", false);
```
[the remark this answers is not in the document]
darkwing: Yep, this is just a convenience while I work on it.

digitarald:
@darkwing Looks great. I only would not expect the first 2 menu items, `Remove/Disable Breakpoint` on a line breakpoint indicator. I am basing that on the assumption of column breakpoints aiming to get rid of line breakpoints on the server. Interactions would just set column breakpoints. Line breakpoints would be just a proxy interface for one/many column breakpoints on the line (similar to Chrome).

I am trying to capture this in the [PRD](https://docs.google.com/document/d/1IqkVX_4MAMkJy6gRe8OHfB4vYqzdps8VwIEThIzrEKU/edit#heading=h.1ukwpv9fyebv) as well. Is there anything we should document further to nail down the final behavior, @darkwing ?

assets/panel/debugger.properties
```diff
@@ -405,6 +405,11 @@ breakpointMenuItem.removeCondition.label=Remove breakpoint condition
 breakpointMenuItem.removeCondition.accesskey=c
 breakpointMenuItem.editCondition.label=Edit breakpoint condition
 breakpointMenuItem.editCondition.accesskey=n
+breakpointMenuItem.disableAllAtLine.label=Disable breakpoints on line
```
flodolo: Any reason why this doesn't have an accesskey?

assets/panel/debugger.properties
```diff
@@ -405,6 +405,11 @@ breakpointMenuItem.removeCondition.label=Remove breakpoint condition
...
+breakpointMenuItem.enableAllAtLine.label=Enable breakpoints on line
+breakpointMenuItem.removeAllAtLine.label=Remove breakpoints on line
+breakpointMenuItem.enableAllAtLine.accesskey=l
```
flodolo: Please move the accesskey close to its label, it avoids confusion for both localizers and developers (if they need to update something).

assets/panel/debugger.properties
```diff
@@ -405,6 +405,12 @@ breakpointMenuItem.removeCondition.label=Remove breakpoint condition
 breakpointMenuItem.removeCondition.accesskey=c
 breakpointMenuItem.editCondition.label=Edit breakpoint condition
 breakpointMenuItem.editCondition.accesskey=n
+breakpointMenuItem.disableAllAtLine.label=Disable breakpoints on line
+breakpointMenuItem.disableAllAtLine.accesskey=d
```
flodolo: D, uppercase

darkwing: `D` is already taken

flodolo: Usage: accesskeys are case insensitive, so you have a duplicate anyway.

"Disable breakpoints on line" will still highlight the first uppercase `D`, because it falls back to the other case if the original one is missing.

flodolo: > Usage: accesskeys are case insensitive, so you have a duplicate anyway.

That, unless DevTools do accesskeys in a different way compared to any other software I know.

assets/panel/debugger.properties
```diff
@@ -405,6 +405,12 @@ breakpointMenuItem.removeCondition.label=Remove breakpoint condition
 breakpointMenuItem.removeCondition.accesskey=c
 breakpointMenuItem.editCondition.label=Edit breakpoint condition
 breakpointMenuItem.editCondition.accesskey=n
+breakpointMenuItem.disableAllAtLine.label=Disable breakpoints on line
+breakpointMenuItem.disableAllAtLine.accesskey=Z
```
flodolo: This is going to show up as `Disable breakpoints on line (Z)`. As long as you're OK with that, I don't have an issue.

Note that in preferences we gave up on having unique accesskeys, given the number of controls in in the same scope).

jasonLaster:
Great work @darkwing!

PR #7839, opened 2019-01-29
Monospace conditional panel and expressions

darkwing:
We can start this a discussion point:

conditional-panel-expressions

- Conditional panel input is monospaced
- Expressions and the expression input is monospaced

@digitarald @mcroud


darkwing:
@mcroud The font sizing for Watch Expressions feels a bit big; by contrast, the font-size for breakpoints is 11px.  If I applied 11px there, we'd get:

11px

nchevobbe:
Hey @darkwing , FYI, in the toolbox there's a `devtools-monospace` class that handles the font-family and the font-size (there are some differences between OSes). See [devtools/client/themes/common.css#66-69](https://searchfox.org/mozilla-central/rev/c07aaf12f13037c1f5a343d31f8291549e57373f/devtools/client/themes/common.css#66-69)
Maybe we could use that class in the debugger as well ? (not sure if that'd work in Launchpad though)

jaril:
If for example the watch expression evaluates an object rep, should the rest of that rep also be monospace? If that's the case, it would look different from how the object is represented in the Scopes pane which displays the object rep in sans-serif.

Though the alternative would be to have just the root/parent expression be monospace, and the rest of its children be sans-serif. Not ideal for consistency either.

digitarald:
Thanks @darkwing for taking a shot at this to make it easier to discuss!

I saw the same thing as @jaril stick out. The watch expression _output_ should probably be similar to reps; while the _expression_ itself would be in monospace. Font sizes should not change.

darkwing:
Interesting.  @jaril, could you provide a few sample expressions you used for testing?

jaril:
Just had something simple in mind like this which covers primitive/non-primitive cases nicely:

```javascript
var arr = new Array(10);
arr.myCustomProp = true; // custom prop

var obj = {};
obj.myCustomProp = true;

var primitive = "string";
```

@rald To clarify, do you mean that in the example below the _expressions_ (`arr`/`obj`/`primitive`/`unavailable`) should be in monospace? While the rest of the rep stays as is?

screen shot 2019-01-30 at 9 35 14 pm

darkwing:
I've updated this PR to monospace the inputs of expression and conditional input.

@digitarald Could you provide more detail?  I think @jaril asked a good question.  It's still unclear to me what exactly should or shouldn't be monospaced.  Examples and which parts should be monospaced would be super helpful!

digitarald:
@jaril summed up well. Just to clarify, I mocked it up

image

 @darkwing does that meet your understanding as well?

mcroud:
@darkwing The Watch Expresisons content and all pane body content should be 11px, this was recently concluded as a result from the vertical rhythm discussion work 👍

PR #7845, opened 2019-01-30
Fix pretty printing in source list

### LINKED ISSUE #7784: [PrettyPrint] icon not showing up in source tree
Pretty printing a source file does not change the source icon to the pretty print icon `{}`

My expectation is when I pretty print something, it should change the source icon to the pretty print icon. Instead it stays the same, although the pretty printed tab shows the pretty print icon just fine.

jan-23-2019 10-43-52

Also revisited #7295 and it seems like that the issue remains. Upon pretty printing, we expect it to show a pretty print icon in the pretty printed tab and the source. Instead, the React icon remains in the source.

jan-23-2019 10-45-02

I'm not the most familiar with expected pretty print behavior so if somebody else can chime in with how this is supposed to behave, it would be much appreciated.

jaril:
Fixes #7784 

Pretty printing a source now shows the pretty print icon beside the source in the source list pane.
<details><summary>GIF</summary>

</details><br/>
screen shot 2019-01-30 at 10 55 14 am

### DIFF
--- a/src/components/PrimaryPanes/SourcesTreeItem.js
+++ b/src/components/PrimaryPanes/SourcesTreeItem.js
@@ -15,7 +15,7 @@ import AccessibleImage from "../shared/AccessibleImage";
 import {
   getGeneratedSourceByURL,
   getHasSiblingOfSameName,
-  hasPrettySource
+  hasPrettySource as checkHasPrettySource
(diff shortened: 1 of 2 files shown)


darkwing:
Excellent!

PR #7846, opened 2019-01-30
add types to source-maps API

### LINKED ISSUE #7822: [sourcemaps] refactor API to better use flow
We recently re-wrote the Parser Worker [API][pw] to better use flow and we should do the same thing with the source maps worker [API][sw]:

Here's a gist where i got started: https://gist.github.com/be0012d62c902dd39768672ffce3d5d4

The big difference is going from 

```js
const getOriginalLocation = dispatcher.task("getOriginalLocation");
```

to 

```js
export const getOriginalLocation = async (
  location: SourceLocation,
  options: locationOptions = {}
): Promise<AstLocation[]> =>
  dispatcher.invoke("getOriginalLocation", location, options);

```

The latter is typed with Flow so we have better coverage

[pw]: https://github.com/devtools-html/debugger.html/blob/master/src/workers/parser/index.js
[sw]: https://github.com/devtools-html/debugger.html/blob/master/packages/devtools-source-map/src/index.js

Mrigank11:
Fixes #7822 

Similar to #6625, I added types to the dispatcher tasks for the source maps worker API. Some of the underlying methods were not typed, so I tried to add types to them.

### Queries
- In some of the methods, types are not named. For example:
```
...
): Promise<
  Array<{
    line: number,
    columnStart: number,
    columnEnd: number
  }>
...
```
Should I create named types for them? Or leave it for another PR?
- I couldn't find any reference to `getLocationScopes`. Should it be removed?
- About this snippet:
```
import * as self from "devtools-source-map";
export default self;
```
I am not sure whether this is the best way to deal with mixed: [default](https://github.com/devtools-html/debugger.html/blob/master/src/utils/test-head.js#L13) and [named](https://github.com/devtools-html/debugger.html/blob/5d826e39d3d0a547116968f2e831155b724c677e/src/test/tests-setup.js#L15) imports. Would love to hear your views on this.

### Summary of Changes

* Adds types to dispatcher tasks for the source maps worker
* Fixes a few related type issues

### Test Plan
I ran `yarn test` before creating this PR and there were no failed tests. 
I don't think any new tests are required.

### DIFF
--- a/packages/devtools-source-map/src/index.js
+++ b/packages/devtools-source-map/src/index.js
@@ -5,62 +5,123 @@
 // @flow
 
 const {
+  workerUtils: { WorkerDispatcher }
+} = require("devtools-utils");
+
+import type { SourceLocation, Source, SourceId } from "debugger-html";
+import type { SourceMapConsumer } from "source-map";
+import type { locationOptions } from "./source-map";
+
+export const dispatcher = new WorkerDispatcher();
+
+export const setAssetRootURL = async (assetRoot: string): Promise<void> =>
+  dispatcher.invoke("setAssetRootURL", assetRoot);
+
+export const getOriginalURLs = async (
+  generatedSource: Source
+): Promise<SourceMapConsumer> =>
+  dispatcher.invoke("getOriginalURLs", generatedSource);
+
+export const hasOriginalURL = async (url: string): Promise<boolean> =>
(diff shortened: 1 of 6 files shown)


jasonLaster:
@Mrigank11 could you rebase and fix the jest test?

Mrigank11:
@jasonLaster it's ready for review.

PR #7858, opened 2019-01-31
[prettier] Switch to 90 characters

jasonLaster:
this is a fun PR, which is purely aesthetics.

with that said, i have noticed that mozilla allows 90 characters and we're at 80, which can be awkward sometimes with releases :)


jasonLaster:
ugh - i can't get eslint to respect 90 characters, it keeps reverting to 80 even when i change max-len...

https://eslint.org/docs/rules/max-len

juliandescottes:
@jasonLaster for the record the 90 char limit is /devtools only :) 

Here is our configuration:
```
"max-len": ["error", 90, 2, {
      "ignoreUrls": true,
      "ignorePattern": "data:image\/|\\s*require\\s*\\(|^\\s*loader\\.lazy|-\\*-"
    }],
```

princiya:
> this is a fun PR
@jasonLaster wow, 223 files changed! 😀

darkwing:
Closing this due to merge conflicts; since I'm guessing it's an update of package.json and one command to enforce the 90, it's not worth the trouble of rebasing.

PR #7862, opened 2019-02-01
Fix styling issues with Toggle Panes buttons (fixes #7742)

### LINKED ISSUE #7742: Left "Collapse panes" button icon is blurry
The left "Collapse panes" button (whose label could change, see #7737) has a blurry icon on 1x displays, both on the horizontal and vertical axis.

The SVG icon itself is alright, but it's a 16px icon centered inside an odd-numbered box. (If you center a 16px icon in a 25px-wide content area, you get 9.5px on each side. You will get a blurry rendering at least in Firefox; Chrome may be forcing SVG elements to align with the pixel grid, I'm not sure.)

Causing the issue:
- Width: we have a even `min-width` and a 1px border. The border should probably be removed.
- Height: the button has `height: inherit`, which makes it inherit the height of its parent, which is `29px`. We should remove the height altogether, since the button is a flex child of the toolbar and already stretches on the y axis.

As a side note, we are rotating the full button to reverse the icon. We should only rotate the SVG element (or its parent `<i>` element), not the button itself.

I'm also not found of the transition when rotating. We don't have one for similar buttons in Inspector and Network, and it would be nice to be consistent across DevTools. An animated rotation on an element which also jumps from one place to another (when we show/hide the sidebar) feels a bit strange too.

fvsch:
Fixes #7742

- Gets updated CSS (variables.css, common.css) from mozilla-central
- Use the new `--theme-icon-color` for icon color
- Sets button font-size and font-family to inherit; on Linux they tend to inherit the system font-size, was 16px (and a 21px line-height) in my case, can be 14.5px by default (11pt), etc. This creates a bunch of alignment issues due to the big line-height. Note: there is a similar style in m-c's common.css, but for some reason it doesn't seem to apply here. Maybe the `html|button` selector syntax doesn't work for the Debugger or doesn't work in all contexts?
- Make sure we have two different icon designs for "expand panes" and "collapse panes". The arrow in "expand" should point away from the sidebar, and the arrow in "collapse" should point towards the sidebar. Just flipping the "expand" icon doesn't make it convey "collapse" correctly. Incidentally, this helps us simplify the CSS code because we don't have to mentally compute the correct rotation for different parameters. :D
- Remove the button's `height:inherit`, which make them 31px and causes misalignments and bluriness. They're flex children with `align-self: stretch`, so not having a height at all is fine.
- The PaneToggleButton component was accepting a `collapsed` boolean prop, but treating it as meaning the opposite of the boolean value. When clicking the button, it would basically say "set the pane's collapsed state to this value I received". As a result, when using the PaneToggleButton we had to invert the "collapsed" state we were passing to it. And also, we were outputting the "collapsed" class on buttons for panes which were NOT collapsed and not outputting that class on buttons for panes which WERE collapsed. Not sure that's clear. But basically, we were doing this _backward_s (and _wrongly_, for the element's class). So, well, I tried to fix that.

### DIFF
--- a/images/pane-collapse.svg
+++ b/images/pane-collapse.svg
@@ -1,0 +1,7 @@
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 16 16" width="16" height="16">
+  <path d="M4 3H2v10h2V3zm7 7V6L8 8l3 2z" fill-opacity=".3"/>
+  <path d="M11.47 10.88A1 1 0 0 0 12 10V6a1 1 0 0 0-1.55-.83l-3 2a1 1 0 0 0 0 1.66l3 2c.3.2.7.23 1.02.05zM8 8l3-2v4L8 8zM2 2h12c.6 0 1 .4 1 1v10c0 .6-.4 1-1 1H2c-.6 0-1-.4-1-1V3c0-.6.4-1 1-1zm12 11V3H5v10h9zM2 13h2V3H2v10z"/>
+</svg>
--- a/images/pane-expand.svg
+++ b/images/pane-expand.svg
@@ -1,0 +1,7 @@
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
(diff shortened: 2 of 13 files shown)


fvsch:
Since I changed a few things in PaneToggleButton, the snapshots are out of date. What's the right course of action to fix that?

There's also a test setting the `collapsed` prop to a string instead of a boolean, which I could fix.

fvsch:
I've updated the `PaneToggleButton.spec.js` test and its snapshop to match the changes in `PaneToggleButton.js`. I still need to work my way through other tests and snapshots.

fvsch:
- Fixed tests (and updated 2 snapshots)
- Removed `package.json` change (it's the same dependency bump as what was done on master) and `yarn.lock`.

I think we can move to `pr-needs-review`.

darkwing:
This is wonderful @fvsch !  Thank you so much!

PR #7869, opened 2019-02-02
Improve SourcesTreeItem.mapStateToProps performance by storing plain urls

wartmanm:
Without storing them, every visible tree item does the same work of trimming query parameters from every URL, which gets expensive pretty quickly.

I ended up not using reselect because it didn't seem like a good fit - there's no way to know what changed, so you'd have to rebuild the entire plain URL map every time the URL map changed.

For testing, I navigated the debuggee to https://wartmanm.github.io/debugger-benchmarks/500-scripts.html, made sure the source list in the debugger was fully expanded, and profiled adding a watch expression of "1 + 1".

Before: https://perfht.ml/2Sj4MvH
After: https://perfht.ml/2SgTfNz

I'd like to tackle getHasMatchingGeneratedSource and checkHasPrettySource as well, but this has a much bigger impact.

### DIFF
--- a/src/reducers/sources.js
+++ b/src/reducers/sources.js
@@ -260,7 +260,7 @@ function updatePlainUrl(state: SourcesState, source: Object) {
   if (!source.url || state.urls[source.url]) {
     return;
   }
-  const plainUrl = source.url.split("?")[0];
+  const plainUrl = getPlainUrl(source.url);
   const existing = state.plainUrls[plainUrl] || [];
   state.plainUrls[plainUrl] = [...existing, source.url];
 }
@@ -568,6 +568,11 @@ function getSourcesByUrlInSources(
   return urls[url].map(id => sources[id]);
 }
 
+function getPlainUrl(url: string): string {
+  const queryStart = url.indexOf("?");
+  return queryStart !== -1 ? url.slice(0, queryStart) : url;
+}
+
 export function getSourcesUrlsInSources(
   state: OuterState,
   url: string
@@ -575,7 +580,7 @@ export function getSourcesUrlsInSources(
   if (!url) {
     return [];
   }
-  const plainUrl = url.split("?")[0];
+  const plainUrl = getPlainUrl(url);
   const plainUrls = getPlainUrls(state);
   return plainUrls[plainUrl] || [];
 }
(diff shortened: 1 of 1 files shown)


darkwing:
Awesome work @wartmanm !

jasonLaster:
Hey @wartmanm two things:

1. could you open a new PR with a feature branch
2. what do you think about passing a set of siblings into SourceTreeItem? This might simplify the search quite a bit -- https://gist.github.com/d6fc2835d0ef5c6d84f39e8bc81579ca

darkwing:
Great suggestion @jasonLaster!  Thoughts on this @wartmanm ?

darkwing:
@wartmanm Hi! :) . Would you be able to push this to a feature branch so we could help tweak it?  Thank you!

wartmanm:
Sorry for the extended silence.  I've been delving into the approach jasonLaster described above.

There's a complication in getting it to work as-is: [the tree only re-renders nodes when their `item` prop changes](https://github.com/firefox-devtools/debugger.html/blob/master/packages/devtools-components/src/tree.js#L96), so the sibling prop wouldn't be updated when new sources were added.
You can modify `addToTree` so that neighboring nodes are updated, but I worry that this would still leave other users of `hasSiblingOfSameName`, like `BreakpointHeading`, in the lurch.  (I'm not sure if there is a way around that; I haven't dug into breakpoints enough yet.)

I'm honestly not sure what to do at this point.

wartmanm:
Does pushing a feature branch involve anything besides naming the branch something other than "master"?  Github won't let me change the source branch of this pull request, and I don't want to create another defective one.

jasonLaster:
Yep. Create a new branch locally and then push it and open a PR.

jasonLaster:
Hmm I think you bring up a good point about the complication with item. We've refactored the sources reducer quite a bit. 

Could you update your PR with your initial idea and we can re-evaluate?

jasonLaster:
@wartmanm i was thinking about the assertion that we only update when `item` changes. I think we found a loop hole when doing blackboxing because was want to update the node when the source is blackboxed. The loop hole is updating the SourceTreeItem's key property so that it unmounts and re-mounts. This is fine for sources bc there are no children

PR #7875, opened 2019-02-04
Add better keyboard support for focusable elements inside ObjectInspe…

### LINKED ISSUE #7874: Make ObjectInspector accessible to keyboard when focusable.
At the moment even when ObjectInspector is set to be focusable, focus handling does not work very well for focusable elements that are inside the TreeNode items of its internal Tree. To be consistent with other tools in the toolbox (primarily markup inspector) it'd be ideal to allow to explicitly activate/deactivate a TreeNode for keyboard navigation using Enter[Space]/Escape respectively. Similarly how it happens in the markup view.

yzen:
…ctor and Tree components (fixes #7874)(#7874).

Fixes #7874

* Update Tree component's focus handling. Instead of activate node Tree props, allow TreeNode to handle its inner focus independently when it is set to be active.


packages/devtools-components/src/tests/tree.js
```diff
@@ -195,6 +203,159 @@ describe("Tree", () => {
...
+    });
+    wrapper.getDOMNode().focus();
+    expect(formatTree(wrapper)).toMatchSnapshot();
+    expect(wrapper.find(".active").exists()).toBe(false);
+    expect(wrapper.getDOMNode().ownerDocument.activeElement).toBe(
+      wrapper.getDOMNode()
+    );
+
+    simulateKeyDown(wrapper, "Enter");
+    expect(wrapper.find(".active").prop("id")).toBe("key-L");
```
nchevobbe: maybe we could have a snapshot check here as well `expect(formatTree(wrapper)).toMatchSnapshot();`

src/components/SecondaryPanes/tests/__snapshots__/Expressions.spec.js.snap
```diff
@@ -17,7 +17,6 @@ exports[`Expressions should always have unique keys 1`] = `
         autoExpandDepth={0}
         createObjectClient={[Function]}
         disableWrap={true}
-        focusable={false}
```
nchevobbe: I'm not sure we want those changes in this file?
I guess it could be because `focusable` defaults to `true`.
If we still want that, maybe we should put `focusable: false` on the `Expression` component.

@darkwing should the Object in the Expressions panel be focusable?

packages/devtools-components/src/tests/tree.js
```diff
@@ -653,14 +778,27 @@ function getTreeNodes(wrapper) {
...
     key,
     preventDefault: () => {},
-    stopPropagation: () => {}
+    stopPropagation: () => {},
+    ...options
   });
 }
 
+function renderItemWithFocusableContent(x, depth, focused, arrow) {
+  let children = [{}, arrow, focused ? "[" : null, x, focused ? "]" : null];
```
nchevobbe: nit: we should put the closing bracket after the `if (x === "L")` block so it "wraps" the whole line.

nchevobbe: also, I'd exclude the first object from the array since it's the `dom.div` option object, and not a children

packages/devtools-components/src/tests/tree.js
```diff
@@ -653,14 +778,27 @@ function getTreeNodes(wrapper) {
...
+  if (x === "L") {
+    children = [
+      ...children,
+      dom.a({ id: "active-anchor", href: "#" }, "anchor")
+    ];
+  }
+
+  return dom.div(...children);
```
nchevobbe: If we remove the `{}` from the `children` array, we can change this line into: 

```suggestion
  return dom.div({}, children);
```

packages/devtools-components/src/tests/tree.js
```diff
@@ -653,14 +778,27 @@ function getTreeNodes(wrapper) {
...
+    ...options
   });
 }
 
+function renderItemWithFocusableContent(x, depth, focused, arrow) {
+  let children = [{}, arrow, focused ? "[" : null, x, focused ? "]" : null];
+  if (x === "L") {
+    children = [
+      ...children,
+      dom.a({ id: "active-anchor", href: "#" }, "anchor")
```
nchevobbe: nit: could we put a space before the "anchor" ? It's only to make it a tad better in snapshots.

yzen:
> Excellent work @yzen! I love it!
> 
> How did you manually test this patch? What are our risks with it?

Essentially for the ObjectInspectors that have focusable property set to true (such as scopes, for example) you can ensure that you can tab into the scopes tree and tab out without having to go through the whole tree. In terms of activation you would need a row that has something focusable (for example a open inspector button that nodes reps have). You can also notice that with the trees in the debugger, ArrowExpander is an actual button and it is no longer focusable unless its tree node is "active".

This code is essentially identical to the m-c version: https://bugzilla.mozilla.org/show_bug.cgi?id=1424159 where you can see it in action with the a11y panel sidebar.

packages/devtools-components/src/tree.js
```diff
@@ -12,6 +12,20 @@ require("./tree.css");
 // depth
 const AUTO_EXPAND_DEPTH = 0;
 
+// Simplied selector targetting elements that can receive the focus, full
+// version at
+// http://stackoverflow.com/questions/1599660/which-html-elements-can-receive-focus
+// .
```
nchevobbe: this bit is a bit weird :)

packages/devtools-components/src/tree.js
```diff
@@ -679,10 +812,37 @@ class Tree extends Component {
...
     }
   }
 
+  /**
+   * Current active element within the document.
+   * @return {DOMNode}
+   *         Current active element within the document if available.
+   */
+  get activeElement() {
+    return this.treeRef.current.ownerDocument.activeElement;
```
nchevobbe: should we guard on `this.treeRef` ? (in case this is called before the component is mounted/after it's unmounted) ?

src/components/SecondaryPanes/tests/__snapshots__/Expressions.spec.js.snap
yzen: AFAIK, this should be focusable or it is simply inaccessible. Changes to tree should make it not annoying to tab through.

packages/devtools-components/src/tree.js
yzen: Ah that's taken from:
https://github.com/mozilla/gecko-dev/blob/master/devtools/client/shared/focus.js

nchevobbe: weird!

packages/devtools-components/src/tree.js
yzen: I can add that!

darkwing:
This works great for me!  Awesome work @yzen !

jasonLaster:
Unfortunately, this patch fails browser_webconsole_object_in_sidebar_keyboard_nav.

https://treeherder.mozilla.org/#/jobs?repo=try&revision=9677d62d2832352c2eab422171d641b7c97c7aa6&selectedJob=229134680

jasonLaster:
@yzen that makes sense. I can land the release with the revert so it is easy to fix in MC

PR #7878, opened 2019-02-04
Fix: Double-clicking watch expression values causes editing

### LINKED ISSUE #7867: Double-clicking watch expression values causes editing
Reps in other parts, like the Scopes panel, allow double-click to select text. Watch expressions listen on the whole element for double-clicks and switch to edit; even though only the expression should be editable.

This matches behavior for Chrome as well.

@darkwing, possibly a good first bug?

lawrencek0:
Fixes #7867 

### Summary of Changes

* Moved the function `editExpression` to ObjectInspector to pass it to ObjectInspectorItem


princiya:
@lawrencek0 Please update the Jest snapshot - `yarn test -u` 🙏🏻

darkwing:
This is a great start @lawrencek0 !  Welcome!

One thing I did notice was that if I add the expression `document.body.firstChild`, expand the node, then double-click a property (like `async`), the edit interface appears.  I think we need to isolate the doubleclick-to-edit to *just* the property entered by the user.

packages/devtools-reps/src/object-inspector/components/ObjectInspectorItem.js
```diff
@@ -286,7 +293,8 @@ class ObjectInspectorItem extends Component<Props> {
                 setExpanded: this.props.setExpanded
               });
             }
-          : undefined
+          : undefined,
+        onDoubleClick: onExpressionDblClick
```
darkwing: I see that the `renderLabel` function this appears in has a few interesting things we could try:

1.  The `onClick` has a check for selection:  `if (Utils.selection.documentHasSelection()) {` ; maybe that could be of use?

2.  There is a `depth` check; maybe if `depth` is one, and there's more than one depth, we ignore on depth 0?

lawrencek0: Thanks @darkwing. I implemented a depth check for that

flodolo:
I have the feeling there were some issues with a rebase? I see changes that I'm not sure belong to this PR, at least in the .properties file

lawrencek0:
I am not sure. I rebased from the origin/master branch. Should I close this PR and open a new one in a new branch?

darkwing:
@lawrencek0 It may be easiest to do such, since your changes were small. 👍

lawrencek0:
Closing for #8013

PR #7752, opened 2019-01-18
[breakpoints] update column breakpoints tests

jasonLaster:
here's a wip cb test. I'll come back to this once release 119 lands and my mc is cleaner...

PR #7893, opened 2019-02-06
fixed keyboard shortcuts on MacOS [#7892]

### LINKED ISSUE #7892: Space added between the characters in shortcuts on macOS
Formatted Keyboard shortcuts on MacOS have a space added between characters
image

However on MacOS it's a standard to format shortcuts without any space
image

The fix needs to be implemented in src/utils/text.js (formatKeyShortcut function).

bamanczak:
Fixes #7892

Here's the Pull Request Doc

### Summary of Changes

* Removed spaces from formatKeyShortcut function for MacOS consistency

### Test Plan

1. I verified manually that the fix updated the shortcuts on the initial load of debugger
2. I run `yarn test` (Tests: 10 skipped, 1649 passed, 1659 total)

### DIFF
--- a/src/components/WelcomeBox.css
+++ b/src/components/WelcomeBox.css
@@ -2,12 +2,6 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
 
- :root[platform="mac"] .welcomebox .shortcutKey,
- .launchpad-root[platform="mac"] .welcomebox .shortcutKey {
-   --monospace-font-family: monospace;
-   font-weight: 600;
- }
-
 .welcomebox {
   position: absolute;
   top: var(--editor-header-height);
@@ -67,6 +61,12 @@
   font-size: 14px;
   line-height: 18px;
   color: var(--theme-body-color);
+}
+
+:root[platform="mac"] .welcomebox .shortcutKey,
+.launchpad-root[platform="mac"] .welcomebox .shortcutKey {
+  font-family: system-ui, -apple-system, sans-serif;
+  font-weight: 500;
 }
 
 .shortcutLabel {
--- a/src/components/WelcomeBox.css
+++ b/src/components/WelcomeBox.css
@@ -2,6 +2,12 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
 
+ :root[platform="mac"] .welcomebox .shortcutKey,
+ .launchpad-root[platform="mac"] .welcomebox .shortcutKey {
+   --monospace-font-family: monospace;
+   font-weight: 600;
+ }
+
 .welcomebox {
   position: absolute;
   top: var(--editor-header-height);
--- a/src/utils/text.js
+++ b/src/utils/text.js
@@ -29,10 +29,10 @@ const isMacOS = appinfo.OS === "Darwin";
 export function formatKeyShortcut(shortcut: string): string {
   if (isMacOS) {
     return shortcut
-      .replace(/Shift\+/g, "\u21E7 ")
-      .replace(/Command\+|Cmd\+/g, "\u2318 ")
-      .replace(/CommandOrControl\+|CmdOrCtrl\+/g, "\u2318 ")
-      .replace(/Alt\+/g, "\u2325 ");
+      .replace(/Shift\+/g, "\u21E7")
+      .replace(/Command\+|Cmd\+/g, "\u2318")
+      .replace(/CommandOrControl\+|CmdOrCtrl\+/g, "\u2318")
+      .replace(/Alt\+/g, "\u2325");
   }
   return shortcut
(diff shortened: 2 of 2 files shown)


errietta:
The test failures are unrelated, there's #7752 that should fix that

darkwing:
Nice work.  I'd like others to weigh in here; the key representations look "bunched" or too close to me:

crunched1

I wonder if some `letter-spacing` would help out here.  @violasong @fvsch

fvsch:
See https://github.com/devtools-html/debugger.html/issues/7892#issuecomment-461090972 for my comments about the font family.

For spacing, we could use some letter spacing for macOS only. Or we could use thin spaces:
http://www.fileformat.info/info/unicode/char/2009/index.htm
http://www.fileformat.info/info/unicode/char/200a/index.htm
http://www.fileformat.info/info/unicode/char/202f/index.htm

violasong:
Changing the font-weight to normal may fix the bunched feeling.

bamanczak:
So right now we have this:
image


After trying out the suggestion from [#7892 comment](https://github.com/devtools-html/debugger.html/issues/7892#issuecomment-461090972), it looks like this:
image
Which looks better if you ask me :)
But as I understand from the comment it would require additional testing? I would need some guidance to make sure we won't introduce any weird regression.

Changing font-weight makes no effect.

fvsch:
Applying the styles only to macOS requires a selector like:

```css
:root[platform="mac"] .welcomebox .shortcutKey {
  ...
}
```

Which works for the Debugger in Firefox, but not in "Launchpad" mode (the HTML element doesn't have that attribute). It looks like instead the platform attribute is added to a div in the body, so the relevant selector would become:

```css
:root[platform="mac"] .welcomebox .shortcutKey,
.launchpad-root[platform="mac"] .welcomebox .shortcutKey {
  ...
}
```

@darkwing, do you know if there is some automation in place that converts the first selector form to the second one? Or do we need to write both?

fvsch:
In launchpad mode I'm seeing styles like this, which seem to come from mozilla-central but the selector was tweaked:

```css
:root.html [platform="linux"] {
    --monospace-font-family: monospace;
}
```

Original:

```css
:root[platform="linux"] {
    --monospace-font-family: monospace;
}
```

violasong:
Ah, my font-weight comment was based on David's screenshot but looks like your version is different. Latest version looks good but I miss the bolding of the shortcuts :)

fvsch:
If we use the mac system font we can probably use `font-weight: 500` or `font-weight: 600` to get a semibold variant.

darkwing:
Lots of good discussion here! @fvsch can you summarize the change to be made in one comment?

bamanczak:
I went through the comments and implemented them in a suggested way. Right now the welcomebox looks on Mac as follows:
image
Which I think is nice :)

src/components/WelcomeBox.css
```diff
@@ -2,6 +2,12 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
 
+ :root[platform="mac"] .welcomebox .shortcutKey,
```
fvsch: This rule should be placed after the styles for `.welcomebox .shortcutKey`, to keep a logical order.

src/components/WelcomeBox.css
```diff
@@ -2,6 +2,12 @@
...
+ .launchpad-root[platform="mac"] .welcomebox .shortcutKey {
+   --monospace-font-family: monospace;
```
fvsch: So I'm not very fond of the monospace font here. It looks like with `font-family: monospace` on a mac, Firefox may end up using Courier, which is not something we have elsewhere in DevTools. The `--monospace-font-family` CSS variable is actually defined so that we can specify a specific monospace font on each platform, so we shouldn't override it, even locally in some branch of the DOM.

It would be more interesting here IMO to try to use San Francisco, which seems to be the font used in the issue's screenshots. [According to Can I Use](https://caniuse.com/#search=system-ui), Firefox for Mac supports the `-apple-system` on macOS. For Chrome, it seems that the `system-ui` value is supported. So maybe something like:

```css
[platform="mac"] .welcomebox .shortcutKey {
  font-family: system-ui, -apple-system, sans-serif;
  font-weight: 600;
}
```

src/components/WelcomeBox.css
bamanczak: Do you think now it's the correct place? :)

fvsch: Looks like it. :)

fvsch:
@bamanczak Can you show updated screenshots of the result on Mac, in Firefox and Chrome?
If you think a different `font-weight` works better, feel free to propose that as well.

src/components/WelcomeBox.css
bamanczak: This solutions seemed fine both on Chrome and FF. I also changed font-weight to 500, as it looks better.

bamanczak:
As requested, screenshot from Chrome:
image
And from firefox
image

darkwing:
Thank you for the update @bamanczak !  We'll have a look!

darkwing:
I love this @bamanczak !  Thank you!

PR #7906, opened 2019-02-07
Prevent renovate from reopening PRs

darkwing:
According to https://renovatebot.com/docs/configuration-options/#recreateclosed , we shouldn't need this, but it's worth a shot to prevent renovate from reopening the same PR over and over.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -4,6 +4,7 @@
     "config:base",
     ":preserveSemverRanges"
   ],
+  "recreateClosed": false,
   "statusCheckVerify": true,
   "ignoreDeps": [
     "react",


src/components/SecondaryPanes/Breakpoints/Breakpoint.js
```diff
@@ -102,8 +103,12 @@ class Breakpoint extends PureComponent<Props> {
       return false;
     }
 
-    const bpId = getLocationWithoutColumn(this.selectedLocation);
-    const frameId = getLocationWithoutColumn(frame.selectedLocation);
+    const bpId = features.columnBreakpoints
```
jasonLaster: this probably isn't desired...

PR #7908, opened 2019-02-07
Fix 6734 - Prevent two active breakpoints

### LINKED ISSUE #6734: Two Breakpoints Active in Secondary Pane
I was resuming through breakpoints and saw that two of them were marked as active:

probs

Only one should be active at a time.

darkwing:
Fixes #6734

We'll want to check this closer with breakpoint specificity and column breakpoints landing, but this is a start....

### DIFF
--- a/src/components/SecondaryPanes/Breakpoints/Breakpoint.js
+++ b/src/components/SecondaryPanes/Breakpoints/Breakpoint.js
@@ -16,7 +16,8 @@ import { CloseButton } from "../../shared/Button";
 
 import {
   getLocationWithoutColumn,
-  getSelectedText
+  getSelectedText,
+  makeBreakpointId
 } from "../../../utils/breakpoint";
 import { getSelectedLocation } from "../../../utils/source-maps";
 import { features } from "../../../utils/prefs";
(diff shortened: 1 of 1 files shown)


darkwing:
Testing this further; we may actually not need the ternary at all once breakpoint specificity lands.

darkwing:
This will be the best solution; we can't use `makeLocationId` in both places because frameId returns a column while the breakpoint doesn't:

columnsoff

src/components/SecondaryPanes/Breakpoints/Breakpoint.js
```diff
@@ -102,8 +103,12 @@ class Breakpoint extends PureComponent<Props> {
...
     }
 
-    const bpId = getLocationWithoutColumn(this.selectedLocation);
-    const frameId = getLocationWithoutColumn(frame.selectedLocation);
+    const bpId = features.columnBreakpoints
+      ? makeBreakpointId(this.selectedLocation)
+      : getLocationWithoutColumn(this.selectedLocation);
+    const frameId = features.columnBreakpoints
+      ? makeBreakpointId(frame.selectedLocation)
+      : getLocationWithoutColumn(frame.selectedLocation);
```
jasonLaster:
```js
const bpId = makeBreakpointId(this.selectedLocation)
frameId = makeBreakpointId(frame.selectedLocation)
```

darkwing: ?  Did you see my comment?

jasonLaster: oh i see

PR #7922, opened 2019-02-10
[Breakpoints] Remove pause points

jasonLaster:
### Summary of Changes

- removes the pause points code
- updates the breakpoint position code


src/utils/source-maps.js
```diff
@@ -6,9 +6,10 @@
 
 import { isOriginalId } from "devtools-source-map";
 import { getSource } from "../selectors";
+import { isGenerated } from "../utils/source";
 
 import type { SourceLocation, MappedLocation, Source } from "../types";
-import { isGenerated } from "../utils/source";
+import typeof SourceMaps from "../../packages/devtools-source-map/src";
```
loganfsmyth: Isn't this line the same as
```
import typeof SourceMaps from "devtools-source-map";
```
?

And should this potentially be
```
import typeof * as SourceMaps from "devtools-source-map";
```

PR #7930, opened 2019-02-10
Bump flow-bin to 0.92.1

AnshulMalik:
https://github.com/facebook/flow/releases

It might have some nice things for us :)


jasonLaster:
Wow! that was much better than I feared :P

@ryanjduffy will be happy to see `makeRecord` become simpler

PR #7934, opened 2019-02-11
Change editor line-height to 15px (#7929)

### LINKED ISSUE #7929: Set source code line-height to 15px
Summary: I'm going to make a PR for changing the source code line-height from `auto` (resulting in slightly different values between configs, often around 13.5px) to 15px. This was discussed previously and validated with @violasong but I want to provide context and an opportunity to push back if someone thinks it could be an issue.

/cc @mcroud 

## Context

Currently DevTools panels tend not to specify the line-height of text. Since we mostly have 11px text (base font-size and code font-size; tabs and accordion headers are 12px) with `line-height: auto`, we end up with a line-height that can be 13px, 13.5px or 14px (with perhaps a few more extreme values, depending on the actual font used and the platform).

I've proposed increasing the code line-height a bit to improve readability. At the last All Hands I reviewed a few options with Victoria, mainly 14px, 15px and 16px line-height, using this demo as a reference:
https://fx-devtools-markup-line-height.netlify.com/

We settled on using 11px text with a 15px line-height, and would like to use that consistently across devtools if possible (Inspector markup view and Rules, Debugger sources, Style Editor sources, etc.).

## Prior discussion & related issues

- General discussion: https://github.com/devtools-html/ux/issues/2
- At least one Slack discussion where changing the Debugger's source line-height from 13.5px to 16px was positively received. Since 15px is a bit more conservative, I take it as a hint that it would fit Debugger nicely too.
- Inspector markup view: https://bugzilla.mozilla.org/show_bug.cgi?id=1508036
- Creating shared styles in m-c: https://bugzilla.mozilla.org/show_bug.cgi?id=1520440

## Implementation

I'm probably not going to wait on bug 1520440 to progress, mostly because it's hard making it progress without knowing exactly how we would consume CSS variables or classes in different panels. So my new strategy is "implement first, refactor later". :grin:

fvsch:
Fixes #7929 (rationale for the change is explained there)

### Summary of Changes

- Specify line-height in Editor as `calc(15 / 11)` (resulting in 15px for a 11px font-size; we could used a fixed value but they don't inherit well when changing the font-size, while ratio values do).
- Tweak the breakpoint SVGs and relevant styles to be 15px tall (line breakpoint) and 13px tall (column breakpoint); remove duplicate column-breakpoint.svg
- Change line number hover style that used a background-color on the element itself and a `::after` pseudo-element for the right part to just use a `::after` pseudo-element painted behind the number; removes some extra CSS.
- Remove legacy styles for `.CodeMirror-linenumber:before` that was used for breakpoint indicators in the past, but don't seem used at all anymore.

### DIFF
--- a/images/breakpoint.svg
+++ b/images/breakpoint.svg
@@ -1,6 +1,6 @@
 <!-- This Source Code Form is subject to the terms of the Mozilla Public
    - License, v. 2.0. If a copy of the MPL was not distributed with this
    - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
-<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 60 14">
-  <path d="M53.07.5H1.5a1 1 0 0 0-1 1v11a1 1 0 0 0 1 1h51.57a2 2 0 0 0 1.53-.7L59.3 7l-4.7-5.8a2 2 0 0 0-1.53-.7z"/>
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 60 15" width="60" height="15">
+  <path d="M53.07.5H1.5c-.54 0-1 .46-1 1v12c0 .54.46 1 1 1h51.57c.58 0 1.15-.26 1.53-.7l4.7-6.3-4.7-6.3c-.38-.44-.95-.7-1.53-.7z"/>
 </svg>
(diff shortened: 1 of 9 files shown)


darkwing:
@fvsch Could you rebase this?  I want to make sure I don't break the svg file.

fvsch:
@darkwing Sure, done. :)

fvsch:
@darkwing The travis build failed, but I'm not sure that's related to my changes. Do you know if I should change anything?

darkwing:
I'm looking into this now.  The good news is that the errors all appear to be related to preview, and from my brief testing this morning, the `hoverAtPos` isn't triggering the preview, so I think it's possible your dimension changes have somehow made hovering "miss" the target.  I'm looking into it though.

PR #7935, opened 2019-02-11
Use different labels for start and end PaneToggleButtons (#7741)

### LINKED ISSUE #7741: [Accessibility] "Collapse panes" buttons should have different labels
We have two buttons with the same "Collapse panes" label:

- One collapses the sources navigation
- One collapses the right-hand panes

This creates an accessibility issue because both buttons appear in the same context, and without access to visual information (proximity of the button with either sidebar) you can't differentiate them.

I suggest using a different label for the first button, maybe "Collapse navigation"?

/cc @yzen 

Related issue: #4073.

fvsch:
Fixes #7741

Warning: removes 2 L10N strings and adds 4!
I don't know if there's a special process to follow up with people who work on localization?

I've tried to restrict the flow type for position to an enum with 2 strings, and fix the tests accordingly. I'm pretty new to flow so I did it by ear, but it seems alright.

### DIFF
--- a/assets/panel/debugger.properties
+++ b/assets/panel/debugger.properties
@@ -10,9 +10,13 @@
 # A good criteria is the language in which you'd find the best
 # documentation on web development on the web.
 
-# LOCALIZATION NOTE (collapsePanes): This is the tooltip for the button
-# that collapses the left and right panes in the debugger UI.
-collapsePanes=Collapse panes
+# LOCALIZATION NOTE (collapseSources): This is the tooltip for the button
+# that collapses the Sources and Outlines panes in the debugger UI.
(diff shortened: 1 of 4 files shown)


jasonLaster:
Hmm, this is a bit tough ... I think of the first pane as navigation (source, functions) and the second pane for pausing (expressions, bps, stacks, scopes), but it is a bit of a bucket so tough to describe...

darkwing:
Chrome uses "Hide navigator" and "Hide debugger".

What about "Hide sources" and "Hide debugger" or "Hide breakpoints".  I think "breakpoints" is fine because everything to do with breaking is in that pane.

darkwing:
@flodolo Could you approve please?

PR #7946, opened 2019-02-12
Upgrade Node in Circle, Travis

### LINKED ISSUE #7864: Tests require Node 10, package.json says 7.7
I'm using Node 8 and a bunch of tests fail because Node 8 doesn't have Promise.prototype.finally (Node 10 does).

- Should the minimum version be bumped to Node 10 in `package.json`?
- Or do we want to inject a `.finally` polyfill?

MDN says Promise.prototype.finally is Node 10.0.0+ but not sure how reliable that info is:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/finally

With Node 8.15:

```
Test Suites: 12 failed, 184 passed, 196 total
Tests:       70 failed, 10 skipped, 1576 passed, 1656 total
Snapshots:   7 failed, 413 passed, 420 total
Time:        118.244s
```

With Node 10.15:

```
Test Suites: 1 failed, 195 passed, 196 total
Tests:       1 failed, 10 skipped, 1645 passed, 1656 total
Snapshots:   1 failed, 434 passed, 435 total
Time:        108.516s, estimated 111s
```

(Still have one test to fix, please disregard.)

darkwing:
Fixes #7864

Not competing with @dhyey35, I just want to see if we can normalize our Travis and Circle on the same version of Node

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,7 +16,7 @@ jobs:
       NVM_DIR: ~/.nvm
 
     docker:
-      - image: node:11.9.0
+      - image: node:10.15.0
     # machine: true
     steps:
       - checkout
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,7 @@ install:
   - pip install --user mercurial
   - python --version
   - node --version
-  - . $HOME/.nvm/nvm.sh; nvm install 9; nvm use 9;
+  - . $HOME/.nvm/nvm.sh; nvm install 10.15.0; nvm use 10.15.0;
   - ./bin/ci/install-yarn
   - ./bin/ci/clone-gecko.sh
   - yarn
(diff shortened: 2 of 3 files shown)


darkwing:
@ochameau What version of Node.js do we have baked into MC?  We should probably keep them all in sync.

ochameau:
I'm not 100% sure, but I think it is defined [over here](https://searchfox.org/mozilla-central/source/python/mozbuild/mozbuild/nodeutil.py#17-18):
```
NODE_MIN_VERSION = StrictVersion("8.11.0")
NPM_MIN_VERSION = StrictVersion("5.6.0")
```
@dmose can certainly confirm.

darkwing:
I think using these versions seems reasonable; we can revert if we run into issues.

dmose:
The stuff built in CI and provided to `mach bootstrap` is 8.11.3. 

Note that if you write a bunch of stuff that depends heavily on 10.x features, and then need to revert down the road, it may be harder than you think.

That said, 8.11.3 is supposed to be EOLed at the end of the year, so we won't be on it that long.

PR #7948, opened 2019-02-13
Remove postcss-bidirection

### LINKED ISSUE #7941: Consider removing postcss-bidirection
Using postcss-bidirection creates issues in the CSS codebase because it adds specificity to selectors, e.g. this `0.1.0` selector:

```
.my-component {
  margin-inline-start: 10px;
}
```

becomes a `0.2.1` selector:

```
html[dir="rtl"] .my-component {
  margin-right: 10px;
}
```

This creates small specificity wars, with traces that can be seen in the CSS codebase.
So it would be cleaner to use the native support for direction-relative margin/padding/border/inset, if possible.

Turns out Chrome 69, released 2018-09-04 (current stable is 72) introduced support for all the relevant properties. A few are noted in the MDN compat data, but others are missing, so I filed an issue:
https://github.com/mdn/browser-compat-data/issues/3443

My tests in Chrome 69 are pretty conclusive.

Now, Debugger may also support other browsers for its launchpad mode. So I looked at Edge and Safari:

- Latest Edge: nope, no support.
- Safari 12: only supports `-webkit-(margin|padding|border)-(start|end)`

The good news is that autoprefixer is already adding those webkit prefixes.
The main limit for WebKit would be the lack of an equivalent to `inset-inline-(start|end)`, which we are using in half a dozen places in the CSS codebase.

I guess it depends on the browser support we want.

fvsch:
Fixes #7941

* Remove postcss-bidirection from `postcss.config.js`
* Fix the RTL magnifying glass icon position in RTL
* Change a few text-align values to start|end

I've tested a few things and things look alright in Chrome and Firefox.
More hands on testing would be welcome though. (We may find RTL issues that already exist on master, by the way.)

### DIFF
--- a/docs/local-development.md
+++ b/docs/local-development.md
@@ -80,17 +80,11 @@ Set the `dir` field in the Launchpad's settings pane.
 
 _How do I change how something looks in rtl?_
 
-We use [postcss-bidirection][bidirection] to support [logical CSS][logical] properties. In practice, this means we can write `float:left` or `margin-inline-block: start`, and it just works. Under the hood, `float: left` gets translated into two different CSS rules for `html[dir="rtl"]` and `html:not([dir="rtl"])`.
-
-`public/js/components/SourceFooter.css`
+We use [logical CSS][logical] properties. For instance, if you need to add some padding-left in left-to-right layout and padding-right in right-to-left layout:
 
 ```css
-html:not([dir="rtl"]) .source-footer .command-bar {
-  float: right;
-}
-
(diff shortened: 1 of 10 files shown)


darkwing:
Seeing this in the XHR breakpoints panel:
panelclosexhr

fvsch:
@darkwing Good catch. We were using `offset-inline-*` properties instead of `inset-inline-*`, which postcss-bidirection would fix for us. I moved to using only `inset-inline-*` and removing the older `offset-inline-*` properties (where we were using both).

PR #7950, opened 2019-02-13
Modified the Angular regex

### LINKED ISSUE #7901: Having "angular" in a project name collapses everything into Angular stacks
Live: https://firefox-devtools-angular-log.stackblitz.io/
Editor: https://stackblitz.com/edit/firefox-devtools-angular-log

STR:
- Create a project with angular in the name
- Log, trace or pause in a component

ER: Stacks for the component are not detected as Angular stack
AR: The first entry in the collapsed Angular group, `ngOnInit`, should not be part of the collapsed.

_Hello Component_

image

_Paused with call stack_

image

_Console trace_

image

Possible solutions: Match both `angular` and `core`, as `core` is the sub folder for the main angular code. This is just preliminary checking of paths in the call stack:

image

jaril:
Fixes #7901

### Summary of Changes

Modified `getLibraryFromUrl()` so that it only matches Angular's core framework sub folder, not its components.
- Matches `/turbo_modules/@angular/core@7.0.1/bundles/core.umd.js`
- Doesn't match `/~/src/app/hello.component.ts`

Note that in the image below, `ngOnInit` (which is defined in hello.component.ts) isn't collapsed into the Angular frame group:

screen shot 2019-02-13 at 9 35 38 am

### DIFF
--- a/src/utils/pause/frames/getLibraryFromUrl.js
+++ b/src/utils/pause/frames/getLibraryFromUrl.js
@@ -78,7 +78,13 @@ const libraryMap = [
   },
   {
     label: "Angular",
-    pattern: /angular\.|angular.*(animations|common|core|elements|forms|platform-browser|platform-server|platform-webworker|router|service-worker|upgrade)/i,
+    pattern: new RegExp(
+      "angular\\.|" +
+        "angular.*" +
+        "(animations|common|core|elements|forms|platform-browser|" +
+        "platform-server|platform-webworker|router|service-worker|upgrade)",
+      "i"
+    ),
     contextPattern: /(zone\.js)/
   },
   {
--- a/src/utils/pause/frames/getLibraryFromUrl.js
+++ b/src/utils/pause/frames/getLibraryFromUrl.js
@@ -78,7 +78,7 @@ const libraryMap = [
   },
   {
     label: "Angular",
-    pattern: /angular.*core/i,
+    pattern: /@angular\//,
     contextPattern: /(zone\.js)/
   },
   {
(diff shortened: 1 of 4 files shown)


jaril:
Made this more comprehensive to match all of Angular's framework code. This now matches the following cases:

- AngularJS (AngularJS 1.x) which usually bundled angular code as `angular.js` or `angular.min.js`. Examples are [Google Family Link](https://families.google.com/familylink/) and [United Jetstream](https://jetstream.united.com/#/public-landing).
- Angular (AngularJS 2.x) which bundles Angular code within [directories](https://angular.io/api) (e.g. core, common). For example, most of Angular's core functionality is implemented by `/@angular/core@7.2.4/bundles/core.umd.js`.

src/utils/pause/frames/getLibraryFromUrl.js
```diff
@@ -78,7 +78,13 @@ const libraryMap = [
   },
   {
     label: "Angular",
-    pattern: /angular/i,
+    pattern: new RegExp(
+      "angular\\.|" +
+        "angular.*" +
+        "(animations|common|core|elements|forms|platform-browser|" +
```
darkwing: I'm slightly worried that using these specific strings will cause a maintenance issue, in that every time they rename a file or add one, we'll need to make updates.  Can we do a `angular.([A-z+])` type of regex?  It's more generic but I doubt many people purposely prefix the specific JS with `angular.`

jaril: Good call - switched this out so that it's much simpler :)

jaril:
Changed this so that all it does is it doesn't recognize Angular Components as framework code, which was the issue brought up in the first place. 

Added unit tests to that effect, and to check framework URLs for both AngularJS and Angular (1.x and 2.x)

PR #7957, opened 2019-02-13
Ensure consistent filenames throughout tabs, tree, and search

darkwing:
I noticed when going to https://w2x2oypmqk.codesandbox.io/ that the sources tree cited "external%20file" but, when opened, the tab said "(index)", as did a sources search.

Simply checking for no "." is far too basic.

yaysame

I've also unescaped the text in the tree, as it used to be.

### DIFF
--- a/src/components/PrimaryPanes/SourcesTreeItem.js
+++ b/src/components/PrimaryPanes/SourcesTreeItem.js
@@ -213,7 +213,7 @@ class SourceTreeItem extends Component<Props, State> {
       case "webpack://":
         return "Webpack";
       default:
-        return `${item.name}`;
+        return `${unescape(item.name)}`;
     }
   }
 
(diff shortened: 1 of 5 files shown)


src/components/PrimaryPanes/SourcesTreeItem.js
```diff
@@ -213,7 +213,7 @@ class SourceTreeItem extends Component<Props, State> {
       case "webpack://":
         return "Webpack";
       default:
-        return `${item.name}`;
```
jasonLaster: would we want to unescape this closer to the source? like in getURL or something?

jasonLaster: maybe we should add a jest test for this?

darkwing: This place is the only occurrence of the problem

PR #7975, opened 2019-02-15
Add a BigInt rep.

### LINKED ISSUE #7974: Add support for BigInt
In https://bugzilla.mozilla.org/show_bug.cgi?id=1527867 we're going to add proper grip for BigInt, so we need to have a Rep to display them.

nchevobbe:
Fixes #7974.

The server will send a grip of the following shape: 

```json
{
  "type": "BigInt",
  "value": "1",
}
```
where `value` is `myBigInt.toString()` (we can send the actual BigInt as they're not serializable).

We can display it by taking `value` and appending `n` to it (e.g. `1n`).

### DIFF
--- a/packages/devtools-reps/src/object-inspector/tests/utils/should-render-roots-in-reps.js
+++ b/packages/devtools-reps/src/object-inspector/tests/utils/should-render-roots-in-reps.js
@@ -12,6 +12,7 @@ const gripStubs = require("../../../reps/stubs/grip");
 const gripArrayStubs = require("../../../reps/stubs/grip-array");
 const symbolStubs = require("../../../reps/stubs/symbol");
 const errorStubs = require("../../../reps/stubs/error");
+const bigIntStubs = require("../../../reps/stubs/big-int");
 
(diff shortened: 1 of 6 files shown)


nchevobbe:
Server support is happening in [Bug 1527867](https://bugzilla.mozilla.org/show_bug.cgi?id=1527867)

darkwing:
Yesssssss!

nchevobbe:
> Yesssssss!

Is this a r+ ? ;)

darkwing:
This looks good but do we want a more realistic example / test?  And do we want to do any type of formatting (ellipsizing) of a very long value?

nchevobbe:
> do we want a more realistic example / test?

I'll add other test case.

> And do we want to do any type of formatting (ellipsizing) of a very long value?

I'm not sure. We could not have anything first and see if we get any report that overly long BigInt are an issue.

nchevobbe:
Added a bunch of tests to make sure we always display big ints as expected.

Issue #7868, opened 2019-02-01
Switch delete button for watch expressions to show on hover


digitarald:
Same behavior as breakpoints. Impact of not doing it is loosing 48 pixels (close button + margins).

Possibly a good first bug, @darkwing ?

darkwing:
I think we should consider accessibility here; using hover for a task as important as removing a breakpoint or expressions isn't friendly to keyboard users.

digitarald:
As this is already used for breakpoints, how is the a11y question solved there @darkwing ?

digitarald:
Looks like it doesn't.

This could be solved by making the close button element appear when the main element or a part of it is focused (using `:focus-within`); so tapping through the element shows all options.

keerthanak17:
I'd like to claim this issue if no one else is working on it!

/claim

keerthanak17:
Hi @digitarald , could you tell how I can find the file/ where in the application this issue is? I'm having trouble finding this issue to analyse/ fix it. Thanks!

darkwing:
Hello @keerthanak17 !  The file most important will be `src/components/SecondaryPanes/Expressions.js`, where the JavaScript/HTML is.  There's a `Expressions.css` file where you'll need to add the CSS to show and hide the close button.  A good file to look at will be `src/components/SecondaryPanes/Breakpoints/Breakpoint.js`, where we already implement this functionality. 👍

keerthanak17:
Awesome! Let me do that and get back to you. Thank you @darkwing

keerthanak17:
I think I got this. Do I just give a PR for adding it into the code or is there something else I'm missing?

darkwing:
You can submit a pull request and we'll have a look! :)

fvsch:
For the record I'm generally against hiding UI and revealing it on hover. I've argued against it for breakpoints in #8042.

Not sure freeing up space is such a big concern. Also, if we do end up using the freed up space to display text, that means that on hover we overlay the remove button on top of that text, which is messy.

> Using hover for a task as important as removing a breakpoint or expressions isn't friendly to keyboard users.

This can be worked around by not hiding the button with `display: none`, and using both :focus (or :focus-within) and :hover to reveal it.

```css
.the-row-container {
  position: relative;
  overflow: hidden;
}
.the-close-button {
  /* visually hide by default */
  top: -100px;
}
.the-row-container:hover .the-close-button,
.the-close-button:focus {
  top: 0;
}
```

[...]

PR #7976, opened 2019-02-16
Switch expressions delete button

chujunlu:
Fixes #<#7868>

Hi, I read your comment about the concern of accessibility in #7868. Based on Harald's suggestion, I used 'focus-within.' But with that, a user using a mouse needs to click in the expression container so that the delete button displays; hovering the expression doesn't display the delete button. When moving the mouse out of the container, the delete button doesn't disappear. We need to click somewhere else to make the container to lose focus.

This behavior is inconsistent with breakpoint's. If I replace 'focus-within' with 'hover,' it will behave the same as breakpoints. 

I think no matter which one we choose, it's better to make it consistent for both expressions and breakpoints. Let me know what you think.

### Summary of Changes
* Added display property
* Added visibility property

### Test Plan
- Add watch expression
- Click on the expression container
- Move mouse out of the container
- Click somewhere to turn off 'focus-within'

### Screenshots
http://g.recordit.co/dEXtmKef1C.gif

### DIFF
--- a/src/components/SecondaryPanes/Expressions.css
+++ b/src/components/SecondaryPanes/Expressions.css
@@ -99,6 +99,22 @@
   top: 1px;
 }
 
+.expression-container .close-btn {
+  display: none;
+}
+
+.expression-container:focus-within .close-btn {
+  display: flex;
+}
+
+.expression-container .close {
+  visibility: hidden;
+}
+
+.expression-container:focus-within .close {
+  visibility: visible;
+}
+
 .expression-content {
   position: relative;
 }


darkwing:
Thank you for taking this @chujunlu!  

I appreciate this demo and it gives us something to think about.  Since @digitarald would prefer expressions to look like I think it may be best to implement `focus-within` on Breakpoints first, then port that functionality to Expressions.

chujunlu:
Thanks @darkwing  for reviewing my PR! Yes, that makes sense. I'll keep an eye on what you guys decide. If you open an issue for Breakpoints, or something else, I'm happy to see if I can help.

darkwing:
@chujunlu I'll close this one for now.

Would you like to have a try at implementing this for the Breakpoints panel?  I'd love to share that work with @digitarald when he gets back from vacation! :)

chujunlu:
@darkwing For sure. I'd like to give it a try. I have a question about workflow: Need I create an issue for Breakpoints, then create a PR referring the issue; or just create a PR? Need I refer to the original issue (7868) as well? Thanks for your time!

darkwing:
Yep, creating an issue would be good 👍

PR #7977, opened 2019-02-16
Consolidate sources reducer

bhackett1024:
### Summary of Changes

* Keep track of source actors for a source in the Source itself, instead of a separate table in the sources reducer.  This allows removing CreateSourceResult.
* Rename relativeSources to displayedSources to better fit with how the purpose of this store.
* Store SourceId instead of entire Source in displayed sources table, so that there is only a single place where Source is stored in the reducer.
* Tidy up source related selectors.

### Test Plan

Passes flow/jest.


jasonLaster:
This is much simpler than before. 

I left some comments in this gist because it was easier to edit / think in my editor. I hope that is alright
https://gist.github.com/b0e49b6403aa1e0d438e7ff8a12df1aa

bhackett1024:
Hmm, I can't figure out how to comment inline in the gist, so I've copied parts of it below.

`
+// NOTE: the advantage of having threadUrl here is that some items will be keyed
+// off of thread and some will be threadUrl for instance we show one source per thread url
+// and we add one breakpoint per thread...
+// I'm not sure what the best way to pass threadUrl in is. At the moment we keep the thread url information in redux
+// but perhaps we should keep an array of current worker thread and urls around in the events module...
 export function createSource(
-  thread: string,
+  { thread, threadUrl },
`

For this PR I wanted to avoid changes to how we present sources for different workers alone, so that it can be addressed in followup.  For that followup I'd like to leave SourceActor alone but change the displayed sources to be a map { [threadUrl]: ... } instead of { [ThreadId]: ... }.  We can always determine the URL for a thread given its ThreadId so there isn't a need to store it in SourceActor.

`
-type DisplayedSources = { [ThreadId]: { [SourceId]: boolean } };
+
+// could we switch this to an array of source IDs? would this be any less performant
+type DisplayedSources = { [ThreadId]: SourceId[] };
`

We could, but I was worried about performance if there are a ton of sources for a thread, which seems like it can happen easily with a large project.  I believe that we are O(N) to add a bunch of sources with this patch, but using an array would make this O(N^2).

`
 function updateSources(state, sources: Object[]) {
-  const displayed = { ...state.displayed };
-  for (const thread in displayed) {
-    displayed[thread] = { ...displayed[thread] };
-  }
-
   state = {
     ...state,
     sources: { ...state.sources },
     urls: { ...state.urls },
-    displayed
+    // NOTE: it might be okay not to shallow copy the displayed sets because
+    // we are not relying on the references changing in order to re-compute the data...
+    // with that said, I don't know if we make exceptions like this in other places :P
+    displayed: mapValues(state.displayed, ids => new Set(ids))
};
`

It seems kind of creepy to be modifying objects that are currently referred to by the redux store, even if they'll be replaced later.  The shallow copying in this function is lame, though.  I thought about shallow copying on demand like we do in other places, but it will make adding a bunch of sources at once O(N^2).  What would be really nice would be if we used a smarter data structure than a plain object and could generate modified versions on demand without big performance pitfalls.  I'm hoping we can use immutable.js for this but it seemed like too much for this PR.

`
+    // this is pretty confusing. what's going on here?
`

This is a really ugly way of making sure the new source adds any actors it has to the existing one, instead of overwriting them.  There should be a cleaner way of doing this, e.g. concat the two sets and filter out duplicates.  I'll update the PR.

`
     return mapValues(displayed, threadSourceIds => {
-      const threadSources = mapValues(threadSourceIds, (_, id) => sources[id]);
-      if (chromeAndExtenstionsEnabled) {
-        return threadSources;
-      }
-      return omitBy(threadSources, source => source.isExtension);
+      // is it possible to move this check to where we add sources to the displayed list?
+      const threadSources = [];
+      return threadSourceIds.forEach(id => threadSources.push(sources[id]));
});
`

We could, but I think we would also need to listen to pref changes in the sources reducer so that the displayed list will update immediately if the value of chromeAndExtenstionsEnabled changes.  I think we get the immediate updates for free here by using a selector, but I don't know that for sure nor how important this is.

src/reducers/sources.js
```diff
@@ -640,32 +528,34 @@ export function getProjectDirectoryRoot(state: OuterState): string {
...
-    }
-    return relativeSources;
+  getAllDisplayedSources,
+  (sources, chromeAndExtenstionsEnabled, displayed) => {
+    return mapValues(displayed, threadSourceIds => {
+      const threadSources = mapValues(threadSourceIds, (_, id) => sources[id]);
+      if (chromeAndExtenstionsEnabled) {
+        return threadSources;
+      }
+      return omitBy(threadSources, source => source.isExtension);
```
jasonLaster: we currently do not change the store value when the pref changes. I filed this issue for updating the logic and i think we can re-compute the displayed sources at that point too. 

https://github.com/firefox-devtools/debugger.html/issues/7981

the primary thing i'm concerned here is that this list will still re-compute when sources change like a source is loaded or something, so it is nice to cut down on any of the work..

jasonLaster:
Merging as this works well and it is a big step forward in terms of readability and robustness when there are workers.

src/reducers/sources.js
derek-li: Would having separate stores for attributes that are prone to change (isPrettyPrinted, isBlackBoxed) be a solution to avoid re-computing? 

For example, we have a `state.sources.blackBoxedSources` that stores source ids that we can reference when selecting which sources to blackbox

jasonLaster: Not sure. Feel fee investigating

PR #7984, opened 2019-02-18
[Sources] follow up tweaks for the source consolidation

jasonLaster:
This is a set of a couple follow up tweaks

1. isExtension is field of source
2. we conditionally add sources to the display list based on extension
3. we only keeps sources around in the store if there is one or more actor associated with it


src/actions/sources/newSources.js
```diff
@@ -208,8 +209,12 @@ export function newSource(source: Source) {
   };
 }
 
-export function newSources(sources: Source[]) {
+export function newSources(foundSources: Source[]) {
   return async ({ dispatch, getState }: ThunkArgs) => {
+    const sources = foundSources.filter(
+      source => !getSource(getState(), source.id)
+    );
```
jasonLaster: this is needed so that we don't clobber the original sources data

src/reducers/sources.js
```diff
@@ -289,15 +292,13 @@ function updateWorkers(
...
-      actors: source.actors.filter(({ thread }) => hasThread(thread))
-    };
-  });
+  const threads = [mainThread, ...workers.map(({ actor }) => actor)];
+  const updateActors = source =>
+    source.actors.filter(({ thread }) => threads.includes(thread));
+
+  const sources: Source[] = Object.values(state.sources)
+    .map((source: any) => ({ ...source, actors: updateActors(source) }))
+    .filter(source => source.actors.length > 0);
```
jasonLaster: This filter might be aggressive.

src/reducers/sources.js
```diff
@@ -289,15 +292,13 @@ function updateWorkers(
...
-  const hasThread = thread => {
-    return thread == mainThread || workers.some(({ actor }) => actor == thread);
-  };
-  const sources: Source[] = Object.values(state.sources).map((source: any) => {
-    return {
-      ...source,
-      actors: source.actors.filter(({ thread }) => hasThread(thread))
-    };
-  });
+  const threads = [mainThread, ...workers.map(({ actor }) => actor)];
```
jasonLaster: This will fail if windowless workers is false. If false, we need to use actorID

src/actions/sources/newSources.js
bhackett1024: This filtering looks like it will prevent us from being able to have multiple source actors for the same source.  We rely on being able to handle newSource notifications for the same Source, but with different contents for their source actors.  The sources reducer should merge the old and new information, instead of clobbering all the old state.

PR #7992, opened 2019-02-19
Fixing source selection with the keyboard (fixes #7874) (#7874)

### LINKED ISSUE #7874: Make ObjectInspector accessible to keyboard when focusable.

yzen:
Fixes #7874
Fixes the issue when focused source would not be selected via keyboard using Enter/Space keys.

### DIFF
--- a/src/components/PrimaryPanes/SourcesTree.js
+++ b/src/components/PrimaryPanes/SourcesTree.js
@@ -164,6 +164,10 @@ class SourcesTree extends Component<Props, State> {
     this.props.focusItem({ thread: this.props.thread, item });
   };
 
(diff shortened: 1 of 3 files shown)


darkwing:
Excellent, thank you @yzen !

PR #7994, opened 2019-02-19
Prevent debug line zombie

### LINKED ISSUE #7755: [DebugLine] occassionally the debug line is not cleared

darkwing:
Fixes #7755

It took a while but I was able to finally reproduce the original issue.  I set breakpoints, paused, chose different locations in the call stack, and resumed, then saw the zombie debug line highlight.

After applying Jason's `componentWillUnmount` idea, I no longer see the issue.


darkwing:
WIP'ing as I want to test more manually

jasonLaster:
Can we add a mochitest for this. I think it was bad enough to warrant it. Maybe we just add the test to the existing debug line mochitest

jasonLaster:
Two possible explanations. 

1. is it possible selected location is not equal to selected frame.location? This might be complete bonkers...
2. is it possible we unmount because of this check

```diff
diff --git a/src/components/Editor/index.js b/src/components/Editor/index.js
index 31113ef04..3127ebc79 100644
--- a/src/components/Editor/index.js
+++ b/src/components/Editor/index.js
@@ -597,6 +597,7 @@ class Editor extends PureComponent<Props, State> {
     const { editor, contextMenu } = this.state;

     if (!selectedSource || !editor || !getDocument(selectedSource.id)) {
+      // could we be falling into this location
       return null;
     }
```

src/components/SecondaryPanes/Frames/Frame.js
```diff
@@ -153,7 +153,7 @@ export default class FrameComponent extends Component<FrameComponentProps> {
         role="listitem"
         key={frame.id}
         className={className}
-        onMouseDown={e => this.onMouseDown(e, frame, selectedFrame)}
```
jasonLaster: i think this might actually be an issue

CC @bomsy 
Also, git blame / log could help

test/mochitest/browser_dbg-debug-line.js
```diff
@@ -0,0 +1,38 @@
+/* Any copyright is dedicated to the Public Domain.
+ * http://creativecommons.org/publicdomain/zero/1.0/ */
+
+// Test ensures zombie debug lines do not persist
+// https://github.com/firefox-devtools/debugger/issues/7755
+add_task(async function() {
+  const dbg = await initDebugger("doc-debug-line.html");
```
jasonLaster: do we need another doc for this? could we re-use `dbg-sources` or something?

test/mochitest/browser_dbg-debug-line.js
```diff
@@ -0,0 +1,38 @@
...
+  await addBreakpoint(dbg, "debug-line-2.js", 5);
+
+  // Reload
+  invokeInTab("doThing");
+
+  // Ensure we're paused
+  await waitForPaused(dbg);
+
+  // Click the call stack to get to debugger-line-1
+  const otherLocation = findAllElementsWithSelector(dbg, ".frame")[1];
```
jasonLaster: we should have a frame element in the selectors index. 
Also, it would be nice to use the `nth-child` trick we use in other selectors

test/mochitest/browser_dbg-debug-line.js
```diff
@@ -0,0 +1,38 @@
...
+  otherLocation.click();
+  await waitForSelectedSource(dbg, "debug-line-1.js");
+
+  // Resume
+  resume(dbg);
+
+  // Select the source that had the initial debug line
+  await selectSource(dbg, "debug-line-2.js");
+
+  // Ensure there's no longer a debug line
```
jasonLaster: i think it is nice to use `info("Ensure there's no longer a debug line")` for comments as then we see them in the test.

test/mochitest/browser_dbg-debug-line.js
```diff
@@ -0,0 +1,38 @@
...
+
+  // Ensure we're paused
+  await waitForPaused(dbg);
+
+  // Click the call stack to get to debugger-line-1
+  const otherLocation = findAllElementsWithSelector(dbg, ".frame")[1];
+  otherLocation.click();
+  await waitForSelectedSource(dbg, "debug-line-1.js");
+
+  // Resume
```
jasonLaster: do we need this level of comment? 😉

test/mochitest/browser_dbg-debug-line.js
```diff
@@ -0,0 +1,38 @@
...
+  const dbg = await initDebugger("doc-debug-line.html");
+
+  // Ensure test files load
+  await waitForSources(dbg, "debug-line-1.js", "debug-line-2.js");
+
+  // Add breakpoint to debug-line-2
+  await addBreakpoint(dbg, "debug-line-2.js", 5);
+
+  // Reload
+  invokeInTab("doThing");
```
jasonLaster: are we reloading?

test/mochitest/browser_dbg-debug-line.js
```diff
@@ -0,0 +1,38 @@
+/* Any copyright is dedicated to the Public Domain.
+ * http://creativecommons.org/publicdomain/zero/1.0/ */
+
+// Test ensures zombie debug lines do not persist
+// https://github.com/firefox-devtools/debugger/issues/7755
+add_task(async function() {
+  const dbg = await initDebugger("doc-debug-line.html");
+
+  // Ensure test files load
```
jasonLaster: i think this kind of comment could go above `waitForSources` but might be needed next to the call

test/mochitest/browser_dbg-debug-line.js
```diff
@@ -0,0 +1,38 @@
+/* Any copyright is dedicated to the Public Domain.
+ * http://creativecommons.org/publicdomain/zero/1.0/ */
+
+// Test ensures zombie debug lines do not persist
+// https://github.com/firefox-devtools/debugger/issues/7755
```
jasonLaster: YESSS!!!

test/mochitest/browser_dbg-debug-line.js
darkwing: No, we don't have to.

test/mochitest/browser_dbg-debug-line.js
darkwing: lulz, sorry, pseudo-coded first.

test/mochitest/browser_dbg-debug-line.js
darkwing: Cool, will do.

src/components/SecondaryPanes/Frames/Frame.js
darkwing: I looked at a blame and the last I saw was conversion to JSX...

test/mochitest/browser_dbg-debug-line.js
darkwing: Maybe (?) . I'll have a look.

test/mochitest/browser_dbg-debug-line.js
darkwing: Fair enough; I think `:not(.selected)` could work too.

src/components/SecondaryPanes/Frames/Frame.js
jasonLaster: It looks like the rationale was around focus states which is interesting
https://github.com/firefox-devtools/debugger/pull/1059

If we switch to onClick, we should update the handler as well to `this.onClick`

PR #7995, opened 2019-02-19
Make search result code a bit more compact

darkwing:
Not sure why the repeated args bother me but thought I'd throw this into the mixer.

### DIFF
--- a/src/actions/file-search.js
+++ b/src/actions/file-search.js
@@ -173,9 +173,8 @@ export function traverseResults(rev: boolean, editor: Editor) {
 
     if (modifiers) {
       const matchedLocations = matches || [];
-      const results = rev
-        ? findPrev(ctx, query, true, modifiers.toJS())
-        : findNext(ctx, query, true, modifiers.toJS());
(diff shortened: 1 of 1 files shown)


src/actions/file-search.js
```diff
@@ -173,9 +173,8 @@ export function traverseResults(rev: boolean, editor: Editor) {
 
     if (modifiers) {
       const matchedLocations = matches || [];
-      const results = rev
-        ? findPrev(ctx, query, true, modifiers.toJS())
-        : findNext(ctx, query, true, modifiers.toJS());
+      const findArgs = [ctx, query, true, modifiers.toJS()];
+      const results = rev ? findPrev(...findArgs) : findNext(...findArgs);
```
jasonLaster: while we're nitting we could also do 

`(rev ? findPrev : findNext)(ctx, query, true, modifiers.toJS())`

not sure which i like more...

darkwing: I thought that was too much :D

PR #7996, opened 2019-02-20
Only wait for the main thread to set/clear breakpoints.

bhackett1024:
This is a workaround until issues with setting all breakpoints have been resolved, which will require some work on server side transports.


jasonLaster:
The `sourcemapped-preview` test did not fail for me locally, so re-running the tests.

jasonLaster:
Also, i'm not sure how this change could affect the test.

PR #8001, opened 2019-02-20
[Breakpoints] memoize breakpoint position fetching

jasonLaster:
### Summary of Changes

Memoizes the breakpoint position fetching in a similar manner to loadSourceText

### DIFF
--- a/src/actions/breakpoints/breakpointPositions.js
+++ b/src/actions/breakpoints/breakpointPositions.js
@@ -6,14 +6,12 @@
 
 import { isOriginalId, originalToGeneratedId } from "devtools-source-map";
 
-import {
-  getSourceFromId,
-  getBreakpointPositionsForSource
-} from "../../selectors";
+import { getSourceFromId, hasBreakpointPositions } from "../../selectors";
 
-import type { SourceLocation } from "../../types";
 import type { ThunkArgs } from "../../actions/types";
 import { getOriginalLocation } from "../../utils/source-maps";
+
+const requests = new Map();
 
 async function mapLocations(generatedLocations, state, source, sourceMaps) {
   return Promise.all(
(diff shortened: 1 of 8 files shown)


src/actions/breakpoints/breakpointPositions.js
```diff
@@ -44,10 +47,18 @@ function convertToList(results, sourceId) {
 export function setBreakpointPositions(location: SourceLocation) {
   return async ({ getState, dispatch, client, sourceMaps }: ThunkArgs) => {
     let sourceId = location.sourceId;
+
     if (getBreakpointPositionsForSource(getState(), sourceId)) {
-      return;
+      return getBreakpointPositionsForSource(getState(), sourceId);
```
loganfsmyth: Calling this twice seems unnecessary, can we save the result the first time instead?

src/actions/breakpoints/breakpointPositions.js
```diff
@@ -64,6 +75,10 @@ export function setBreakpointPositions(location: SourceLocation) {
 
     let positions = convertToList(results, sourceId);
     positions = await mapLocations(positions, getState(), source, sourceMaps);
-    return dispatch({ type: "ADD_BREAKPOINT_POSITIONS", positions, location });
+    dispatch({ type: "ADD_BREAKPOINT_POSITIONS", positions, location });
```
loganfsmyth: Can we change this action to take just a `sourceId`? Reviewing this as-is, I was worried that this takes a whole location but we only cache on `sourceId`, so it could easily be a case where the cache key and the used data fall out of sync.

src/actions/breakpoints/breakpointPositions.js
```diff
@@ -44,10 +47,18 @@ function convertToList(results, sourceId) {
 export function setBreakpointPositions(location: SourceLocation) {
```
loganfsmyth: Should we split this in two to keep the caching very clear, e.g.

```
const requests = new Map();
export function setBreakpointPositions({ sourceId }: SourceLocation) {
  return async (thunkArgs: ThunkArgs) => {
    let inProgress = requests.get(sourceId);
    if (!inProgress) {
      inProgress = setBreakpointPositionsPromise(thunkArgs, sourceId).finally(() => {
        requests.delete(sourceId);
      });
      requests.set(sourceId, inProgress);
    }
    return inProgress;
  };
}
async function setBreakpointPositionsPromise(({ getState, dispatch, client, sourceMaps }: ThunkArgs, sourceId: SourceID) {
  // ...
}
```

I think this helps make the general behavior a lot clearer.

src/utils/defer.js
```diff
@@ -5,8 +5,8 @@
 // @flow
 
 export default function defer() {
-  let resolve = () => {};
-  let reject = () => {};
+  let resolve = (...args: any[]) => {};
```
loganfsmyth: This annotations might be better on the `defer` function return value itself, e.g.
```
type Deferred<T> = {
  promise: Promise<T>,
  resolve: (arg: T) => mixed,
  reject: (arg: mixed) => mixed,
};
export default function defer<T>(): Deferred<T> {
  let resolve = () => {};
  let reject = () => {};
  const promise = new Promise((_res, _rej) => {
    resolve = _res;
    reject = _rej;
  });
  return { resolve, reject, promise };
}
```

src/actions/breakpoints/breakpointPositions.js
jasonLaster: oh nice. I was looking for a clean way to do this

src/actions/breakpoints/breakpointPositions.js
```diff
@@ -41,51 +39,72 @@ function convertToList(results, sourceId) {
...
-        }
-      }
-    } else {
-      results = await client.getBreakpointPositions(source.actors[0]);
     }
 
-    let positions = convertToList(results, sourceId);
-    positions = await mapLocations(positions, getState(), source, sourceMaps);
-    return dispatch({ type: "ADD_BREAKPOINT_POSITIONS", positions, location });
+    return requests.get(sourceId);
```
loganfsmyth: nit: this is really just an `await` since the promise doesn't resolve to a value. :P

PR #8006, opened 2019-02-20
[Breakpoints] getBreakpointPositionsForLine should return null when n…

jasonLaster:
small fix that prevents an issue with `getFirstVisibleBreakpoints`


jasonLaster:
yeah, maybe we should change the callee...

jasonLaster:
I think the issue was that the signature suggested it could be null and yet we returned something truthy like an empty array. I think I prefer null, but i could be persuaded in the future.

PR #8009, opened 2019-02-20
Rename repo from debugger.html to debugger

### LINKED ISSUE #7997: Change the repo name to debugger
we recently changed the org from **devtools-html** to **firefox-devtools** to be more clear.  devtools-html was a code name where as firefox-devtools is the product.

We should do the same with the repo and transition to **debugger**.

I think we can just change the setting and at the time move the docs and other URLs over as well.

jasonLaster:
Fixes #7997


kevinmorehouse:
Once this gets the go-ahead, I can replace all instances of _debugger.html_ with _debugger_ in the docs, if necessary.

jasonLaster:
Thanks @kevinmorehouse, that'd be great!

nmicht:
@jasonLaster 
The URL on the description of the project is wrong after the renaming.
We need to update it to be debugger instead of debugger.html

PR #8010, opened 2019-02-21
Fixes Regex for React Libraries

### LINKED ISSUE #7807: Improve the regexp for React library
Currently the regexp to detect the React library is very simple:
https://github.com/devtools-html/debugger.html/blob/192cbd161a136bb42ff8bd9116a052465fa917a2/src/utils/pause/frames/getLibraryFromUrl.js#L24-L25

This means it matches lots of 3rd-party libraries too, especially `react-contextmenu` `react-transition-group` `react-splitter-layout`. I believe React is only `react` and `react-dom` so maybe a good regexp could be:
```
 pattern: /(node_modules\/(?:react|react-dom)/)|(react(\.[a-z]+)*\.js$)/
```
(in case this isn't very visible, there's an added `/` for the `node_modules` part)

championshuttler:
Fixes #7807

### DIFF
--- a/src/utils/pause/frames/getLibraryFromUrl.js
+++ b/src/utils/pause/frames/getLibraryFromUrl.js
@@ -22,7 +22,7 @@ const libraryMap = [
   },
   {
     label: "React",
-    pattern: /(node_modules\/react)|(react(\.[a-z]+)*\.js$)/i
+    pattern: /(node_modules\/(?:react|react-dom)\/)|(react(\.[a-z]+)*\.js$)/
   },
   {
     label: "Immutable",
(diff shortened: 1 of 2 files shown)


jasonLaster:
Thanks @championshuttler! Could you add some jest tests for these new cases?

championshuttler:
Hi @jasonLaster can you help me with it , i never write tests before. 

Thanks

darkwing:
Hello @championshuttler !  Thank you for helping us out!

We have some tests here you can enhance:

src/utils/pause/frames/tests/getLibraryFromUrl.spec.js

You can run them with `yarn jest getLibraryFromUrl.spec.js`; thank you!

PR #8013, opened 2019-02-21
Fix: Double-clicking watch expression values causes editing

### LINKED ISSUE #7867: Double-clicking watch expression values causes editing

midsorbet:
Fixes #7867

### Summary of Changes

* Moved the function `editExpression` to ObjectInspector to pass it to ObjectInspectorItem
* Added a depth check to make sure that double clicking properties does not cause editing

### DIFF
--- a/packages/devtools-reps/src/object-inspector/components/ObjectInspectorItem.js
+++ b/packages/devtools-reps/src/object-inspector/components/ObjectInspectorItem.js
@@ -266,14 +266,7 @@ class ObjectInspectorItem extends Component<Props> {
       return null;
     }
 
-    const {
-      item,
-      depth,
-      focused,
-      expanded,
-      onLabelClick,
-      onExpressionDblClick
-    } = this.props;
+    const { item, depth, focused, expanded, onLabelClick } = this.props;
     return dom.span(
       {
         className: "object-label",
@@ -287,22 +280,6 @@ class ObjectInspectorItem extends Component<Props> {
               }
 
               onLabelClick(item, {
-                depth,
-                focused,
-                expanded,
-                setExpanded: this.props.setExpanded
-              });
-            }
-          : undefined,
-        onDoubleClick: onExpressionDblClick
-          ? event => {
-              event.stopPropagation();
-
-              if (depth > 0) {
-                return;
-              }
-
-              onExpressionDblClick(item, {
                 depth,
                 focused,
                 expanded,
--- a/packages/devtools-reps/src/object-inspector/components/ObjectInspectorItem.js
+++ b/packages/devtools-reps/src/object-inspector/components/ObjectInspectorItem.js
@@ -266,7 +266,14 @@ class ObjectInspectorItem extends Component<Props> {
       return null;
     }
 
-    const { item, depth, focused, expanded, onLabelClick } = this.props;
+    const {
+      item,
+      depth,
+      focused,
+      expanded,
+      onLabelClick,
+      onExpressionDblClick
+    } = this.props;
     return dom.span(
       {
         className: "object-label",
@@ -280,6 +287,22 @@ class ObjectInspectorItem extends Component<Props> {
               }
 
               onLabelClick(item, {
+                depth,
+                focused,
+                expanded,
+                setExpanded: this.props.setExpanded
+              });
+            }
+          : undefined,
+        onDoubleClick: onExpressionDblClick
+          ? event => {
+              event.stopPropagation();
+
+              if (depth > 0) {
+                return;
+              }
+
+              onExpressionDblClick(item, {
                 depth,
                 focused,
                 expanded,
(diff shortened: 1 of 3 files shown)


jaril:
Ran this to double check.

On a primitive, double clicking the node highlights the label + delimiter + value. There's an unrelated node highlighting behavior (#8019) that hides the actual text selection, but I can paste it onto the URL bar in the example below. 

feb-22-2019 11-17-46

On an object, double clicking the node registers as a click first which expands the node, and then on the second click it moves on to edit the label/expression. It ignores `depth > 0` nodes properly so clicking on properties doesn't prompt an edit.

feb-22-2019 11-19-27

lawrencek0:
Thanks for testing it out @jaril. One thing I noticed while using Chrome's DevTools is that double clicking anywhere on the expression block causes editing while on Firefox it does not work on the edge of the expression.
Firefox Dev Tools Behavior
Chrome Dev Tools Behavior

darkwing:
I just pushed a rebase of this branch which hopefully fixes circle

packages/devtools-reps/src/object-inspector/components/ObjectInspectorItem.js
```diff
@@ -266,7 +266,14 @@ class ObjectInspectorItem extends Component<Props> {
...
     }
 
-    const { item, depth, focused, expanded, onLabelClick } = this.props;
+    const {
+      item,
+      depth,
+      focused,
+      expanded,
+      onLabelClick,
+      onExpressionDblClick
```
nchevobbe: I think the naming here is not ideal, since the ObjectInspector is a generic component. Also, it's then bound to `onDoubleClick`, so it could be a bit confusing.
Finally, there's already a `onDoubleClick` props on the `ObjectInspectorItem`, can't we use it?

lawrencek0: Would it be better if I rename `onExpressionDblClick` to `onDoubleClick`?

darkwing: @nchevobbe Could you provide a preference here?  This one is almost good to go! 👍

nchevobbe: Thank you for your patience @lawrencek0 !
So, I think we should try to **not** modify this file at all, and use the `onDoubleClick` prop it already provides.

So, in `Expression.js`, we would do: 
```js
onDoubleClick={(items, options) => {
  const {depth} = options;
  if (depth > 0) {
    return;
  }

  this.editExpression(expression, index)
}}
```

We would have the same functionality, without much changes.
Let me know if you have any question, I'll try to be more responsive this time!

lawrencek0: Thanks, this looks really neat! I always wondered what the options were for.

darkwing:
How does this look @nchevobbe ?

nchevobbe:
code wise it feels fine, but playing with it, it's a bit weird.
Is this only fixing the issue of double clicking on a node from an expanded object?
Also, we lose some affordance, as clicking on the right of the expression result won't edit the expression.
Maybe we could keep the `onDoubleClick` on the `<li>`, and check the event target to see if it's not a "child" node?

```diff
diff --git a/devtools/client/debugger/src/components/SecondaryPanes/Expressions.js b/devtools/client/debugger/src/components/SecondaryPanes/Expressions.js
--- a/devtools/client/debugger/src/components/SecondaryPanes/Expressions.js
+++ b/devtools/client/debugger/src/components/SecondaryPanes/Expressions.js
@@ -257,9 +257,13 @@ class Expressions extends Component<Prop
         className="expression-container"
         key={input}
         title={expression.input}
-        onDoubleClick={(items, options) =>
-          this.editExpression(expression, index)
-        }
+        onDoubleClick={e => {
+          const treeNode = e.target.closest(".tree-node");
+          const level = treeNode && parseInt(treeNode.getAttribute("aria-level"));
+          if (!treeNode || level === 1) {
+            this.editExpression(expression, index)
+          }
+        }}
       >
         <div className="expression-content">
           <ObjectInspector

```

lawrencek0:
Yes, that does feel a bit weird. What do you think @darkwing? Should I change my PR to accommodate that?

darkwing:
Yes, I love @nchevobbe 's idea!

lawrencek0:
Great @darkwing! I will push the changes as soon as I get back!

lawrencek0:
I was away from GitHub. Apologies for not closing it in time!

PR #8014, opened 2019-02-21
Use the union of all mapped and unmapped regions when fetching original breakpoints.

### LINKED ISSUE #7991: Drop usage of getFileGeneratedRange in favor of more specific range checks
The primary issue with `getFileGeneratedRange` is that sourcemaps just don't provide a single range on the generated file for an original source. The current implementation of `getFileGeneratedRange` just takes the first and last mapping in the original file, and uses the generated locations of those mappings to decide on the range. This can break down badly if items in the file are reordered, and in certain contexts with Babel, that can absolutely happen.

`getFileGeneratedRange` was originally added as part of blackboxing to allow us to blackbox the range of an original file. In that context, some of it's shortcomings are mostly acceptable, because blackboxing isn't part of the critical path of the debugging experience.

Recently, however, we added usage of this function for column breakpoint loading to load all of the breakpoints for a given file. This is much more concerning because it could lead to cases where we query the server for breakpoints on a much smaller range than we had expected, leading us to fail to find breakpoints on large portions of a file.

We should update the breakpoint search logic to try to build multiple contiguous ranges over the generated file and query all of these, instead of trying to build one full range to query.

loganfsmyth:
Fixes #7991

Testing locally this looks like it should do the job. Curious to see if there are any test failures.


loganfsmyth:
Alright, this should be ready for review.

PR #8015, opened 2019-02-22
Fix reps launchpad

nchevobbe:
It was broken, not sure for how long, so I fixed it and added some useful buttons


nchevobbe:
r=me because it's only reps launchpad changes

nchevobbe:
mh, circle failing 🤔

nchevobbe:
okay, it's in `src/components/PrimaryPanes/tests/SourcesTreeItem.spec.js` , I don't think I should care.

PR #8016, opened 2019-02-22
Update snapshots for SourcesTreeItem

jaril:
A snapshot update was missed between #8011 and #7984 . This fixes it.
- updated snapshot for `SourceTreeItem renderItem should unescape escaped source URLs 1`


jasonLaster:
Thanks Jaril!

PR #8018, opened 2019-02-22
Update name of repository

kevinmorehouse:
Updated doc to reflect new name of repository (_debugger_) and removed references to old repository name (_debugger.html_), as decided in PR #8009 .

Fixes N/A

### Summary of Changes

* Replaced any instances of _debugger.html_ with _debugger_ (or removed them altogether)
* Fixed capitalization of "debugger" on line 60.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,10 @@
-# [debugger.html][website]
+# [debugger][website]
 
 [website]: http://firefox-dev.tools/debugger/
 
 [![slack-badge]][slack] [![ci-badge]][ci-status] [![PRs Welcome]][make-a-pull-request]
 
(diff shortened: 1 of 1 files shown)


kevinmorehouse:
Am I correct in assuming that the directory name on line 32 should be changed from _debugger.html_ to _debugger_, along with the rest of the changes?

kevinmorehouse:
The _community call_ link (on lines 83 and 114) still points to https://appear.in/debugger.html.

Is the team going to still meet at that url, or are you going to use a new one now?

jasonLaster:
I think we should change debugger.html, but that also means updating appear.in too!

jasonLaster:
Okay, nevermind i think that is taken, lets use `https://appear.in/firefox-debugger`

PR #8022, opened 2019-02-25
Reorder Source Tree (Origins, Webpack, Extensions)

### LINKED ISSUE #7921: [Sources] List webextensions after web origins
When webextension scripts are listed (I think that's only when "Enable browser chrome and addon debugging toolboxes" is checked in DevTools preferences, maybe?), they can appear in-between web origins.

It seems we're currently sorting like this:

- Page origin first
- All other origins, sorted alphabetically (Extensions origins start with "moz-extension")

Which looks like:

screen shot 2019-02-09 at 13 05 37

Could we show the extension origins at the end of the list instead?
So that we have 3 buckets:

1. Page origin
2. Other web origins
3. Webextensions

(We may have more to fit between 1 and 2, for webpack and Angular I believe?)

nagpalm7:
Fixes #7921 

### Summary of Changes

* Reorder function which orders the webpack and extensions at the end of the sourceTree

### DIFF
--- a/src/utils/sources-tree/treeOrder.js
+++ b/src/utils/sources-tree/treeOrder.js
@@ -6,7 +6,12 @@
 
 import { parse } from "../url";
 
-import { nodeHasChildren, isNgBundler, isWebpackBundler, isExactDomainMatch } from "./utils";
+import {
+  nodeHasChildren,
+  isNgBundler,
+  isWebpackBundler,
+  isExactDomainMatch
+} from "./utils";
 import { isUrlExtension } from "../source";
 
 import type { TreeNode } from "./types";
--- a/src/utils/sources-tree/treeOrder.js
+++ b/src/utils/sources-tree/treeOrder.js
@@ -6,7 +6,7 @@
 
 import { parse } from "../url";
 
-import { nodeHasChildren } from "./utils";
+import { nodeHasChildren, isNgBundler, isWebpackBundler, isExactDomainMatch } from "./utils";
 import { isUrlExtension } from "../source";
 
 import type { TreeNode } from "./types";
@@ -26,23 +26,6 @@ export function getDomain(url?: string): ?string {
     return null;
   }
   return host.startsWith("www.") ? host.substr("www.".length) : host;
-}
-
-/*
- * Checks if node name matches debugger host/domain.
- */
-function isExactDomainMatch(part: string, debuggeeHost: string): boolean {
-  return part.startsWith("www.")
-    ? part.substr("www.".length) === debuggeeHost
-    : part === debuggeeHost;
-}
-
-function isNgBundler(part: string): boolean {
-  return part === "ng://";
-}
-
-function isWebpackBundler(part: string): boolean {
-  return part === "webpack://";
 }
 
 /*
--- a/src/utils/sources-tree/treeOrder.js
+++ b/src/utils/sources-tree/treeOrder.js
@@ -7,6 +7,7 @@
 import { parse } from "../url";
 
 import { nodeHasChildren } from "./utils";
+import { isUrlExtension } from "../source";
 
 import type { TreeNode } from "./types";
 
@@ -34,6 +35,14 @@ function isExactDomainMatch(part: string, debuggeeHost: string): boolean {
   return part.startsWith("www.")
     ? part.substr("www.".length) === debuggeeHost
     : part === debuggeeHost;
+}
+
+function isNgBundler(part: string): boolean {
+  return part === "ng://";
+}
+
+function isWebpackBundler(part: string): boolean {
+  return part === "webpack://";
 }
 
 /*
@@ -93,6 +102,57 @@ function createTreeNodeMatcherWithDebuggeeHost(
   };
 }
 
+function createTreeNodeMatcherWithNgBundler(
+  debuggeeHost: ?string
+): FindNodeInContentsMatcher {
+  return (node: TreeNode) => {
+    if (node.name === IndexName) {
+      return -1;
+    }
+    if (debuggeeHost && isExactDomainMatch(node.name, debuggeeHost)) {
+      return -1;
+    }
+    return isNgBundler(node.name) ? 0 : 1;
+  };
+}
+
+function createTreeNodeMatcherWithWebpackBundler(
+  debuggeeHost: ?string
+): FindNodeInContentsMatcher {
+  return (node: TreeNode) => {
+    if (node.name === IndexName) {
+      return -1;
+    }
+    if (debuggeeHost && isExactDomainMatch(node.name, debuggeeHost)) {
+      return -1;
+    }
+    if (isNgBundler(node.name)) {
+      return -1;
+    }
+    return isWebpackBundler(node.name) ? 0 : 1;
+  };
+}
+
+function createTreeNodeMatcherWithExtension(
+  debuggeeHost: ?string
+): FindNodeInContentsMatcher {
+  return (node: TreeNode) => {
+    if (node.name === IndexName) {
+      return -1;
+    }
+    if (debuggeeHost && isExactDomainMatch(node.name, debuggeeHost)) {
+      return -1;
+    }
+    if (isNgBundler(node.name)) {
+      return -1;
+    }
+    if (isWebpackBundler(node.name)) {
+      return -1;
+    }
+    return isUrlExtension(node.name) ? 0 : 1;
+  };
+}
+
 function createTreeNodeMatcherWithNameAndOther(
   part: string,
   isDir: boolean,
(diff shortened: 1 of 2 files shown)


src/components/PrimaryPanes/SourcesTree.js
```diff
@@ -94,6 +94,37 @@ class SourcesTree extends Component<Props, State> {
       debuggeeUrl,
       sources
     });
+    this.orderTree.bind(this);
+  }
+
+  orderTree() {
```
jasonLaster: minor nit, you can do `orderTree = () => {}` and avoid the bind calls

src/components/PrimaryPanes/SourcesTree.js
```diff
@@ -94,6 +94,37 @@ class SourcesTree extends Component<Props, State> {
...
+    const webpack = this.state.sourceTree.contents.filter(item => {
+      if (item.path === "webpack://") {
+        this.state.sourceTree.contents.splice(
+          this.state.sourceTree.contents.indexOf(item),
+          1
+        );
+        return item;
+      }
+    });
+    this.setState({
```
jasonLaster: i'm worried that this can be a performance hit. Would it be possible to limit this just the `addToTree` function in utils which might already have some sorting?

nagpalm7: Okay !

src/components/PrimaryPanes/SourcesTree.js
nagpalm7: thanks @jasonLaster

darkwing:
Jest tests would also be helpful here :)

src/components/PrimaryPanes/SourcesTree.js
nagpalm7: @jasonLaster Actually, `addToTree` is traversing the tree, and appending the nodes to end of the tree.
So, If we implement the sort algorithm there, The time complexity will go to `n^2`  . However, Sorting before passing will make our work done in just 2 loops, i.e. `n` .
So I think the way I did is fine, though some changes like usage of for loops instead of filters can be used.
What do you suggest?

nagpalm7:
@darkwing all tests pass on my local system, Is the fail due to perfomance issue as it shows `timeout error`

darkwing:
I can look into the travis issue.  We should add a jest test, but could you point me to the URL(s) you're using for testing? Thanks!

nagpalm7:
https://www.zoho.com/

BEFORE
-----------
screenshot from 2019-02-28 21-10-23
NOW
---------
screenshot from 2019-02-28 21-10-49

same I have tested for webpacks but in reverse order as i could not find a website in which webpack order was not correct. @darkwing hope it helps. Thanks !

nagpalm7:
https://w2x2oypmqk.codesandbox.io/

Before
--------
screenshot_2019-02-28 debugger

After
-----
after

jasonLaster:
@nagpalm7 could you look into the performance costs of this change? 

Also, would [treeOrder](https://github.com/firefox-devtools/debugger/blob/master/src/utils/sources-tree/treeOrder.js) do something similar?

nagpalm7:
> @nagpalm7 could you look into the performance costs of this change?
> 
> Also, would [treeOrder](https://github.com/firefox-devtools/debugger/blob/master/src/utils/sources-tree/treeOrder.js) do something similar?

@jasonLaster Yes treeOrder is doing same thing but just for main origins, i have added the functions for webpack, angular and web extensions. (Tested webpack and angular, please confirm the changes for extensions as well).

Is there any way to make a general function for bundlers instead of using one for each type ?

**Before**


**Now**
(webpack takes time to load on this particular url so you may see only angular in sources as in one screenshot)

darkwing:
@nagpalm7 Have you tried using just one function?  What problems are you running into?

nagpalm7:
@darkwing Yes, it results in one the bundlers inside the other!

darkwing:
@ravefalcon92 Care to check this out?  You'd mentioned studying the sources tree...

anthonyx:
@darkwing, sorry I changed my username so I didn't see the tag. I'll take a look this week.

nagpalm7:
@anthonyx you are saying to use one function for what, I didn't get you, I think David asked to consolidate all bundlers in one (angular web pack and all) and I tried that It can be consolidated if this is something you saying I will fix this and update in a day.

anthonyx:
@nagpalm7 Yes, I think consolidating the bundlers into one function would be an improvement :)

anthonyx:
It might be a separate issue, but I was saying that we could even refactor the code an additional step and consolidate all custom createTreeNodeMatcherWith*** functions into one function as well. But we can do that on another issue.

nagpalm7:
> It might be a separate issue, but I was saying that we could even refactor the code an additional step and consolidate all custom createTreeNodeMatcherWith*** functions into one function as well. But we can do that on another issue.

Nice Idea! But I think implementation would not be that easy, we need to consider three cases (-1,  0, 1)
, With the above mentioned approach, I think we will miss one. If you have got some hack, we can create an issue and I would love to do that.

darkwing:
@nagpalm7 @anthonyx This is a smaller patch so it would be good to do the consolidation within it -- could you two work together on that?

anthonyx:
Sounds good @darkwing! @nagpalm7: I have an idea for how to consolidate the code, I'll submit a commit for you to look over shortly :)

nagpalm7:
> Sounds good @darkwing! @nagpalm7: I have an idea for how to consolidate the code, I'll submit a commit for you to look over shortly :)

Okay, should I commit the changes I have made to consolidate bundlers?

anthonyx:
> > Sounds good @darkwing! @nagpalm7: I have an idea for how to consolidate the code, I'll submit a commit for you to look over shortly :)
> 
> Okay, should I commit the changes I have made to consolidate bundlers?

Yes, if you have them then that would be helpful. I'll make my changes on top of your changes.

src/utils/sources-tree/treeOrder.js
```diff
@@ -36,6 +37,10 @@ function isExactDomainMatch(part: string, debuggeeHost: string): boolean {
     : part === debuggeeHost;
 }
 
+function isBundler(part: string): boolean {
+  return part === "ng://" || part === "webpack://";
```
anthonyx: Can we move `isBundler` to `src/utils/source.js` since it's more of a utility function for sources and less of a treeOrder function?

src/utils/sources-tree/treeOrder.js
```diff
@@ -146,6 +188,16 @@ export function createTreeNodeMatcher(
     return createTreeNodeMatcherWithDebuggeeHost(debuggeeHost);
   }
 
+  if (isBundler(part)) {
+    // Specialied matcher, when we are looking for angular bundler.
```
anthonyx: Remove 'angular' since we are looking for bundlers in general

src/utils/sources-tree/treeOrder.js
```diff
@@ -93,6 +98,37 @@ function createTreeNodeMatcherWithDebuggeeHost(
...
+  debuggeeHost: ?string
+): FindNodeInContentsMatcher {
+  return (node: TreeNode) => {
+    if (node.name === IndexName) {
+      return -1;
+    }
+    if (debuggeeHost && isExactDomainMatch(node.name, debuggeeHost)) {
+      return -1;
+    }
+    return isBundler(node.name) ? 0 : 1;
```
anthonyx: One edge case:
We call `createTreeNodeMatcher` function in line 174 with `part = "ng://"`
Therefore, line 191 `isBundler(part)` will evaluate to true and we will return `createTreeNodeMatcherWithBundler`

Then, we call `createTreeNodeMatcherWithBundler` where `node.name = "webpack://"`
Therefore, line 111 `isBundler(node.name)` will evaluate to true and we will return 0

However, this is not the correct value to return because webpack and angular are not the same and it should not return 0 because 0 means they are the same. Let me know if this makes sense

src/utils/sources-tree/treeOrder.js
nagpalm7: Yeah we can !

src/utils/sources-tree/treeOrder.js
nagpalm7: Okay, I missed this

src/utils/sources-tree/treeOrder.js
nagpalm7: > @darkwing Yes, it results in one of the bundlers group with the other!

Yes I could not do this earlier because I thought the same, But I think these changes do no effect the tree order/directories, let me confirm again, then I will commit the changes. May be I need to revert this back

anthonyx:
@nagpalm7 I think David also mentioned adding in jest tests so that we don't accidentally regress this desired behavior in the future.

jasonLaster:
closing as this was picked up here
https://phabricator.services.mozilla.com/D31684

Is that right @anthonyx @darkwing?

PR #8024, opened 2019-02-25
Improve SourcesTreeItem.mapStateToProps performance by storing plain urls

wartmanm:
Updated from https://github.com/firefox-devtools/debugger/pull/7869.

I'm not married to this implementation of `updatePlainUrl`.  I'd be fine with implementing it by checking for membership like `updateSourceUrl`, or even rebuilding it from scratch.  It would leave a little bit of performance on the table, but probably not much, and this approach does add complexity.


src/reducers/sources.js
```diff
@@ -464,19 +483,21 @@ export function hasPrettySource(state: OuterState, id: string) {
...
     return [];
   }
-  const plainUrl = url.split("?")[0];
-
-  return Object.keys(urls)
-    .filter(Boolean)
-    .filter(sourceUrl => sourceUrl.split("?")[0] === plainUrl);
+  const plainUrl = getPlainUrl(url);
+  const plainUrls = getPlainUrls(state);
+  return plainUrls[plainUrl] || [];
```
jasonLaster: nit: i think this could be a selector

```js
function getPlainUrlForUrl(url) {
   const plainUrl = getPlainUrl(url);
   return getPlainUrls(state)[plainUrl] || [];
}
```

src/reducers/sources.js
```diff
@@ -236,6 +244,16 @@ function updateSourceUrl(state: SourcesState, source: Source) {
   }
 }
 
+function updatePlainUrl(state: SourcesState, source: Source) {
+  // If we've seen this URL before, it will be in state.urls, and we shouldn't
+  // add it to plainUrls a second time.
+  if (source.url && !state.urls[source.url]) {
+    const plainUrl = getPlainUrl(source.url);
+    const existing = state.plainUrls[plainUrl] || [];
```
jasonLaster: we could re-use the same selector here.

src/reducers/sources.js
```diff
@@ -236,6 +244,16 @@ function updateSourceUrl(state: SourcesState, source: Source) {
   }
 }
 
+function updatePlainUrl(state: SourcesState, source: Source) {
+  // If we've seen this URL before, it will be in state.urls, and we shouldn't
+  // add it to plainUrls a second time.
+  if (source.url && !state.urls[source.url]) {
```
jasonLaster: maybe we should use `hasPlainUrlsForUrl` as well just so we don't make assumptions :)

src/reducers/sources.js
wartmanm: Like so?
```javascript
function getPlainUrlSelectorForUrl(url) {
  const plainUrl = getPlainUrl(url);
  return createSelector(
    getPlainUrls,
    (plainUrls) => plainUrls[plainUrl] || [];
  );
}
```

wartmanm:
I wasn't entirely sure what tests to add that wouldn't be covered by querystrings.spec.js.  I added a couple of simple tests for the new selector, I hope they're close to what you had in mind.

jasonLaster:
Thanks @wartmanm i'll take a look

jasonLaster:
Hey @wartmanm it looks like there is a simple issue with `getPlainUrl` not being imported.

wartmanm:
I accidentally deleted it when I was rebasing to fix the lint errors 😐
It should be fixed now, sorry

jasonLaster:
Thanks! it is nice seeing all the tests 🍏

jasonLaster:
okay - just added some tweaks. I'll merge when it's green.

Thanks for the help!

PR #8025, opened 2019-02-25
[Breakpoints] use the selected location for column breakpoints

jasonLaster:
### Summary of Changes

Fixes an issue where column breakpoints appear in the wrong positions for original locations


packages/devtools-source-map/src/source-map.js
```diff
@@ -367,9 +367,13 @@ async function getGeneratedRangesForOriginal(
...
 
-  const cachedGeneratedMappingsForOriginal = GENERATED_MAPPINGS.get(map);
-  if (cachedGeneratedMappingsForOriginal) {
-    return cachedGeneratedMappingsForOriginal;
+  if (!GENERATED_MAPPINGS.has(map)) {
+    GENERATED_MAPPINGS.set(map, new Map());
+  }
+
+  const generatedRangesMap = GENERATED_MAPPINGS.get(map);
+  if (generatedRangesMap.has(sourceId)) {
```
loganfsmyth: Ugh, thanks! Could we maybe cache this based on `url` instead, since that's the value we actually use for calculating the data?

src/actions/breakpoints/breakpointPositions.js
```diff
@@ -5,55 +5,65 @@
...
 }
 
-async function getBreakpointPositions(
-  sourceId,
-  { client, dispatch, getState, sourceMaps }
-) {
-  let source = getSourceFromId(getState(), sourceId);
+async function _setBreakpointPositions(sourceId, thunkArgs) {
+  const { client, dispatch, getState, sourceMaps } = thunkArgs;
+  let generatedSource = getSourceFromId(getState(), sourceId);
```
loganfsmyth: I think naming this `generatedSource` when it could be original seems potentially confusing. Maybe we can leave the existing variable and declare a new one for the eventual generated source?

src/actions/breakpoints/syncBreakpoint.js
```diff
@@ -106,98 +101,74 @@ export async function syncBreakpointPromise(
...
+      `Could not find relevant sources ${sourceId} - ${generatedSourceId}`
+    );
   }
 
-  const { location, astLocation } = pendingBreakpoint;
+  const { location, generatedLocation, astLocation } = pendingBreakpoint;
   const previousLocation = { ...location, sourceId };
 
-  const scopedLocation = await makeScopedLocation(
+  const newLocation = await findNewLocation(
```
loganfsmyth: Should we always be calling this here, or should we call `findBreakpointPosition` with `previousLocation` position, and fall back to the AST location as a second `findBreakpointPosition` call if nothing was found?

loganfsmyth: I don't think this is strictly related to your changes, so it might be fine for now. My concern is that `findNewLocation` indiscriminantly does
```
  if (func) {
    line = func.location.start.line + offset.line;
  }
```
What if it changes the line to something that is actually wrong, and the existing `line` was actually what we should have checked?

jasonLaster: can you give me an example?

loganfsmyth: Alright say I have
```
function foo() {
  var value = "bar";
  console.log(value);
}
```
with a breakpoint at the start of `console`. It's a breakpoint on line 2, or maybe 3 of the function at column 2.

Say I change the file to
```
function doThing() {
  var value = "bar";
  console.log("bar");
}
function foo() {
  doThing();
}
```
the current logic, I _think_, will take the generated location of the second line of `foo`, which is now the closing `}`, and fail to match, while leaving the breakpoint at the exact original location would have been fine.

It's all edgecasey, so honestly don't worry about it, especially since it's a existing behavior.

jasonLaster:
kicking off a try run
https://treeherder.mozilla.org/#/jobs?repo=try&revision=828c6b10b067b440272ad09e1475e69a9b6a003b

and w/ other gh patches, but not this patch, 
https://treeherder.mozilla.org/#/jobs?repo=try&revision=a533673b883e3be7f8a8ad8f7e727f48e7988163

jasonLaster:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=950752d1bd83d9203d937becbc4ad0c8891193aa

jasonLaster:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=a2fba346c21a67a2a86781473d15774648a5b167

jasonLaster:
Closing this PR in favor of phabricator https://phabricator.services.mozilla.com/D21843

PR #8027, opened 2019-02-25
Update setPreview to properly display falsy values in preview tooltip

### LINKED ISSUE #7956: [Preview] should show "false" in the tooltip
we currently do not show a tooltip wheh you  hover on a token that evaluates to false. I think we should show one because it is confusing when you do not see it.

ravefalcon92:
Fixes #7956 

### Summary of Investigation
Investigated the issue and found that it applied to other falsy values, such as `0` and `''`.
falsy_preview_demo

Within `src/actions/preview.js` we have a piece of code that exits/returns from `setPreview` when the `result` of evaluating the `expression` is a falsy value.

This code originated from PR [#3363](https://github.com/firefox-devtools/debugger/pull/3363). It looks like the use case for this code might be to exit `setPreview` if the `expression` evaluates to something undesired. However, it's also the reason we see unintended behavior with certain falsy values. I  deleted this code to fix the issue.

### Demo of Fix
falsy_preview_solution_demo

Let me know what you think and if there is anything I can improve, thank you!

### DIFF
--- a/src/actions/preview.js
+++ b/src/actions/preview.js
@@ -104,10 +104,6 @@ export function setPreview(
           thread: selectedFrame.thread
         });
 
-        if (!result) {
-          return;
-        }
-
         return {
           expression,
           result,


darkwing:
Wow @ravefalcon92 , great work!

PR #8029, opened 2019-02-26
Add focus-within for showing breakpoint's delete button

### LINKED ISSUE #8012: [Breakpoints][Accessibility]Delete button for breakpoint only shows on hover
Based on the discussions in issue #[7868](https://github.com/firefox-devtools/debugger/issues/7868), we should consider accessibility for both breakpoints and watch expressions' delete buttons. Now the delete button for breakpoints displays when hover, which isn't friendly to keyboard users. The problem might be solved with :focus-within.

screen shot 2019-02-21 at 2 07 03 pm

chujunlu:
Fixes #8012

I keep the `:hover` feature intact for users using a mouse. I added `:focus-with` for keyboard users. I'm new to accessibility practices, so I'd like to listen to feedback and follow up with the discussion. After deciding for breakpoints, we should keep the behavior consistent for watch expressions' delete button.

### Summary of Changes
* Added :focus-within property

### Screenshots/Videos
1) Hovering the label: delete button displays; mouse out: delete button hides.
1emghuu7cu

2) Using tab. Tabbing to the check box: delete button displays; tab out: button hides.
w0hyppvl46

3) One potential "side effect": I use the mouse to check off the box then move my mouse out. Since the checkbox is still in focus, the delete buttons still shows. I need to click somewhere else to get the checkbox out of focus, then delete button can hide. Not sure if it's a problem.
mn5camkubc

### DIFF
--- a/src/components/SecondaryPanes/Breakpoints/Breakpoints.css
+++ b/src/components/SecondaryPanes/Breakpoints/Breakpoints.css
@@ -135,7 +135,7 @@ html .breakpoints-list .breakpoint.paused {
 }
 
 .breakpoints-list .breakpoint:hover .breakpoint-line,
-.breakpoints-list .breakpoint:focus-within .breakpoint-line {
+.A11y-keyboard .breakpoints-list .breakpoint:focus-within .breakpoint-line {
   color: transparent;
 }
 
@@ -188,7 +188,7 @@ html .breakpoints-list .breakpoint.paused {
 }
 
 .breakpoint:hover .close-btn,
-.breakpoint:focus-within .close-btn {
+.A11y-keyboard .breakpoint:focus-within .close-btn {
   display: flex;
 }
 
@@ -197,7 +197,7 @@ html .breakpoints-list .breakpoint.paused {
 }
 
 .breakpoint:hover .close,
-.breakpoint:focus-within .close {
+.A11y-keyboard .breakpoint:focus-within .close {
   visibility: visible;
 }
 
--- a/src/components/SecondaryPanes/Breakpoints/Breakpoints.css
+++ b/src/components/SecondaryPanes/Breakpoints/Breakpoints.css
@@ -56,6 +56,7 @@
 
 .breakpoints-list .breakpoint {
   min-height: var(--breakpoint-expression-height);
+  overflow: hidden;
 }
 
 .breakpoints-exceptions-caught {
@@ -134,8 +135,7 @@ html .breakpoints-list .breakpoint.paused {
   text-align: end;
 }
 
-.breakpoints-list .breakpoint:hover .breakpoint-line,
-.A11y-keyboard .breakpoints-list .breakpoint:focus-within .breakpoint-line {
+.breakpoints-list .breakpoint:hover .breakpoint-line {
   color: transparent;
 }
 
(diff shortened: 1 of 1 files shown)


darkwing:
Wow @chujunlu, this is excellent!  💯

Your third point is a valid concern and I'm glad you've raised it.  Please look at our implementation of `<A11yIntention>`; it's a wrapping component that adds a CSS class based on if the user got to a given element via mouse or keyboard -- I think that could help us out!

In the mean time, maybe @digitarald could add his thoughts on this interaction!

chujunlu:
Thank you @darkwing ! Let me look into that. I might come back with questions :)

chujunlu:
After looking into  `<A11yIntention>` component, point 3 could be solved by adding a class to specify that, only when a user uses keyboard and tab to the checkbox, `:focus-within` gets triggered; clicking with a mouse doesn't trigger the same behavior.

The gif shows that after I check/uncheck the box, move my mouse out, the delete button disappears. I don't need to click somewhere else.
etvjdzmfqr

fvsch:
A simpler fix might be to make sure the button is not using `display: none` but some other way to visually hide it, and then reveal it on `:focus` (rather than `:focus-within` on the parent).

Pseudo-code example in:
https://github.com/firefox-devtools/debugger/issues/7868#issuecomment-468587113

chujunlu:
@fvsch Thanks for the comment. It's simpler. Before I continue or commit anything, I want to make sure is it the desired behavior you have in mind?

1) Mouse
bu7pfpjho0

2) Keyboard tab
v6eyaaivkd

I read that in #[8042](https://github.com/firefox-devtools/debugger/issues/8042), the discussion for `line:column` info hasn't been finalized. So I keep that part of CSS as they are.

fvsch:
@chujunlu Yep it looks like what we want right now. :)

chujunlu:
Sounds good. `position: relative` for the row container, in this case, `.breakpoints-list .breakpoint`, is already there on line 39. So we don't need to add that. 

Let me know what you think :)

src/components/SecondaryPanes/Breakpoints/Breakpoints.css
```diff
@@ -183,22 +183,12 @@ html .breakpoints-list .breakpoint.paused {
...
-}
-
-.breakpoint .close {
-  visibility: hidden;
-}
-
-.breakpoint:hover .close,
-.A11y-keyboard .breakpoint:focus-within .close {
-  visibility: visible;
+.close-btn:focus {
```
fvsch: This publishes `.close-btn` in the global CSS namespace. Any `<button class="close-btn">` anywhere in Debugger would get those styles, leading to conflicts.

A better selector would be `.breakpoints-list .close-btn:focus`.

src/components/SecondaryPanes/Breakpoints/Breakpoints.css
```diff
@@ -183,22 +183,12 @@ html .breakpoints-list .breakpoint.paused {
...
+  top: 25%;
```
fvsch: This only gives us correct vertical centering if the `.breakpoint` container is twice as high as the button.
I believe that button is 16px tall, so we are expecting a 32px tall row. I believe rows are currently a bit smaller than that, and should be changed to be 24px tall in an upcoming update.

Knowing the height of the button is 16px, a better value would be:
`top: calc(50% - 8px);`

src/components/SecondaryPanes/Breakpoints/Breakpoints.css
```diff
@@ -183,22 +183,12 @@ html .breakpoints-list .breakpoint.paused {
   inset-inline-end: 15px;
   inset-inline-start: auto;
   position: absolute;
-  top: 8px;
-  display: none;
+  top: -100px;
```
fvsch: We should probably add a comment to explain this value. Along the lines of "hide the button outside of the row until hovered or focused", maybe.

src/components/SecondaryPanes/Breakpoints/Breakpoints.css
chujunlu: Thanks a lot for pointing these out! I'm not thoughtful enough.
I'm wondering would adding `.breakpoint` before `.close-btn:focus` be able to keep it consistent with line 182 and 189?

fvsch: Either one should be fine. :)

darkwing:
This is excellent @chujunlu !  Great work!

chujunlu:
Florens helped me a lot. Thank you both :)

PR #8034, opened 2019-02-26
Search functionalities

### LINKED ISSUE #7967: Search for text with braces in pattern.
When i try to search in source (ctrl+shift+f) wth braces in pattern i find nothing 
ex: `parameters = Object.assign({node: null}, parameters);`
Then i must escape all braces, which is very annoying, to find what i need.
ex: `parameters = Object.assign\({node: null}, parameters\);`

I propose to search with text pattern by default (like in old versions). Regex search pattern should be an option.

charanjit-singh:
Fixes #7967 

### Summary of Changes

* Added Regex, Case Sensitive, Whole word modifiers in search ( Ctrl + shift + F )
* Added styles for active classes on modifier buttons.

### Test Plan

Tests can be done by searching using combination of all modifiers and checking their results.

Example test plan:

- [x] Open test file in browser.
- [x] Press Ctrl + Shift + F for searching.
- [x] In Input box, type the words to test.
- [x] Check/ uncheck the modifiers.
- [x] Verify the expected results.

### Screenshots
screenshot from 2019-02-26 17-24-15
---

screenshot from 2019-02-26 17-24-23

---

screenshot from 2019-02-26 17-24-44

### DIFF
--- a/src/components/ProjectSearch.js
+++ b/src/components/ProjectSearch.js
@@ -1,5 +1,3 @@
-
-
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
@@ -32,7 +30,6 @@ import type { List } from "immutable";
 import type { ActiveSearchType } from "../reducers/types";
 import type { StatusType } from "../reducers/project-text-search";
 import type { SearchModifiers } from "../types";
-
 
 import "./ProjectSearch.css";
 
@@ -90,14 +87,12 @@ export class ProjectSearch extends Component<Props, State> {
   constructor(props: Props) {
     super(props);
     this.state = {
-
       inputValue: this.props.query || "",
       inputFocused: false,
       focusedItem: null,
       regexMatch: false,
       caseSensitive: false,
       wholeWord: false
-
     };
 
     this.getModifiers = this.getModifiers.bind(this);
@@ -131,12 +126,11 @@ export class ProjectSearch extends Component<Props, State> {
 
   getModifiers() {
     return {
-        caseSensitive: this.state.caseSensitive,
-        regexMatch: this.state.regexMatch,
-        wholeWord: this.state.wholeWord
-      };
+      caseSensitive: this.state.caseSensitive,
+      regexMatch: this.state.regexMatch,
+      wholeWord: this.state.wholeWord
+    };
   }
-
 
   doSearch(searchTerm: string, modifiers: SearchModifiers) {
     this.props.searchSources(searchTerm, modifiers);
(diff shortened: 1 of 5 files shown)


jasonLaster:
This might be a bit of scope creep, but I think it would be nice if the project search bar were now on the bottom and a similar look and feel to file search

src/components/ProjectSearch.js
```diff
@@ -118,8 +124,16 @@ export class ProjectSearch extends Component<Props, State> {
     }
   }
 
-  doSearch(searchTerm: string) {
-    this.props.searchSources(searchTerm);
+  getModifiers() {
```
darkwing: If you change this to `getModifiers = () => {  }`, you can avoid binding on line 98.

src/components/ProjectSearch.js
```diff
@@ -167,7 +181,12 @@ export class ProjectSearch extends Component<Props, State> {
     this.setState({ focusedItem: null });
     const query = sanitizeQuery(this.state.inputValue);
     if (query) {
-      this.doSearch(query);
+      const modifiers = {
```
darkwing: I think you can use your `getModifiers` function here.

src/components/shared/SearchInput.css
```diff
@@ -119,3 +119,15 @@
 .search-field.big .close-btn {
   margin-inline-end: 8px;
 }
+
+.search-field .search-nav-buttons .active {
+  background: #00000030;
```
darkwing: Where did these colors come from?  You should be able to use the same colors as the other search field.

src/components/shared/SearchInput.js
```diff
@@ -107,17 +116,74 @@ class SearchInput extends Component<Props, State> {
     return <AccessibleImage className="search" />;
   }
 
+  // handleModifierRegexMatch()
```
darkwing: These comments need addressing.

src/components/shared/SearchInput.js
```diff
@@ -209,11 +275,20 @@ class SearchInput extends Component<Props, State> {
   renderNav() {
     const { count, handleNext, handlePrev } = this.props;
     if ((!handleNext && !handlePrev) || (!count || count == 1)) {
-      return;
+      return (
+        <div className="search-nav-buttons">
+          <span style={{ padding: "5px" }}>Modifiers:</span>
```
darkwing: We should use CSS for the padding to avoid duplication

darkwing:
This is a really awesome start @charanjit-singh !  I left a few comments to continue you on the journey!

src/components/shared/SearchInput.css
charanjit-singh: Hi @darkwing !
From where Can I get the list of css variables ?
I'm unable to find this in repository.

src/components/shared/SearchInput.js
charanjit-singh: Okay

Issue #6965, opened 2018-09-12
Debugger can't set breakpoint.


johnfound:
Firefox 62.0 (64-bit) on Manjaro Linux.

When trying to set breakpoint in the debugger, it appears for a moment in the breakpoints pane and blue marker appears on the line number of the code, then both instantly disappears. This behavior is consistent for all type of JS code - enclosed in <script> tags of loaded from separate .js file.

janodvarko:
@johnfound: I can't reproduce this issue. It seems to be working for me just fine.

My STR:
1) Load http://janodvarko.cz/firebug/tests/601/Issue601.htm
2) Open Issue601.htm in the Debugger
3) Create a breakpoint on line 35 - the breakpoint is there (in the gutter as well as the right panel)

Does this work for you?
Can you provide a test case that shows the problem?

Honza

mrwweb:
I'm experiencing this issue exactly as described. Windows 10, Firefox 64.0.2. I've disabled all of my add-ons and still have the issue. I followed @janodvarko's test and experienced the issue. Here's a GIF:

![firefox-breakpoint-bug](https://user-images.githubusercontent.com/1238696/51635174-09bd1a00-1f0b-11e9-829e-507de8111874.gif)

jasonLaster:
@mrwweb could you try Firefox 65 or 66 (dev edition or nightly)?

mrwweb:
@jasonLaster It would appear to be working correctly in 65.0beta :)

jasonLaster:
@mrwweb oh good. yeah i think @loganfsmyth fixed this with some of the breakpoint work. it is hard to handle all of the cases :/

th7nder:
I have the same issue. I cannot set a breakpoint. Tried on Chrome, Firefox 65.0.1 and Firefox Developer Edition 66.0b7 on a MacOS 10.14.3.

jasonLaster:
@th7nder can you provide more context on where you're trying to add a breakpoint?

th7nder:
Sure, I am so sorry. Completely forgot about giving you an additional context. 
It happened inside Backbone.js Todo MVC on line 52. Inside function in object which is extended in IIFE.
![image](https://user-images.githubusercontent.com/8305170/52727268-2356fc00-2fb5-11e9-98a3-d12b39806fbd.png)

santicalcagno:
[...]

PR #8044, opened 2019-02-28
Fix 8026 - Prevent breakpoint label from triggering checkbox focus

### LINKED ISSUE #8026: Clicking the breakpoint label triggers the checkbox focus
STR:
- Add a breakpoint
- Slowly lick the breakpoint label (not the checkbox)

AR: Checkbox gets a focused state, looking like it will get toggled
ER: Checkbox doesn't change when clicking, as the label should only jump to the location

screenflow

santicalcagno:
Fixes #8026 

### Summary of Changes

* Replace breakpoint's `label` tag for a `span`, including `aria-labelledby` to the breakpoint's checkbox
* Update breakpoint test snapshot

### Test Plan

Being quite straightforward, I followed @fvsch's advice in the issue's comments but think I bumped into #6965. If anyone can confirm that the change solves the issue I would appreciate it. Thanks!

### DIFF
--- a/src/components/SecondaryPanes/Breakpoints/Breakpoint.js
+++ b/src/components/SecondaryPanes/Breakpoints/Breakpoint.js
@@ -149,6 +149,7 @@ class Breakpoint extends PureComponent<Props> {
     const { breakpoint } = this.props;
     const text = this.getBreakpointText();
     const editor = getEditor();
+    const labelId = `${breakpoint.id}-label`;
     return (
       <div
         className={classnames({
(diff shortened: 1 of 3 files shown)


fvsch:
Tested the PR and it works well for me.

Small detail: the resulting snapshot is not ideal, because the `breakpoint.id` is undefined and since we concatenate it with a string we end up with `undefined-label`. The old snapshot was not great either because the component was using `<label htmlFor={undefined} ...>` so the `for` attribute was dropped from the snapshot.

One fix would be to change the `render` function in `Breakpoint.spec.js`:

```diff
- const props = generateDefaults(overrides, breakpointOverrides);
+ const props = generateDefaults(overrides, {
+   id: "test-breakpoint-id",
+   ...breakpointOverrides
+ });
```

PR #8055, opened 2019-03-01
Fix 7987 - Lessen indent of files

### LINKED ISSUE #7987: [SourceTree] File depth
It looks like we might have an indentation issue for files:

The width for "no source" is now 10px
screen shot 2019-02-19 at 12 07 28 pm

I think it should be 0

screen shot 2019-02-19 at 12 07 37 pm

darkwing:
Fixes #7987

Not sure I necessarily agree with this, since it makes indentation slightly inconsistent, but I thought I'd kick this off as a discussion.

change

### DIFF
--- a/src/components/PrimaryPanes/Sources.css
+++ b/src/components/PrimaryPanes/Sources.css
@@ -77,9 +77,8 @@
   background-color: #ffffff;
 }
 
-/* Use the same width as .img.arrow */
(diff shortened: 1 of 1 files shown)


darkwing:
I agree @fvsch; consistency is key and I like the layout spacing we currently have.

PR #8056, opened 2019-03-01
Revert "Update setPreview to properly display falsy values in preview tooltip"

darkwing:
Reverts firefox-devtools/debugger#8027

While reviewing the latest release (https://phabricator.services.mozilla.com/D21608), the following caused the debugger to go blank:

2.  Add breakpoint to line 1 of main.js
3.  Refresh browser, breakpoint will hit
4.  Hover over "hash"
5.  Debugger bricks

The following error is shown:

```
console.error: (new TypeError("value is null", "resource://devtools/client/debugger/new/src/components/Editor/Preview/index.js", 149))
JavaScript error: resource://devtools/client/debugger/new/src/components/Editor/Preview/index.js, line 149: TypeError: value is null
```

After removing this commit, the debugger no longer bricks.

We need to further investigate why but in the mean time we should revert this commit.

cc @ravefalcon92

### DIFF
--- a/src/actions/preview.js
+++ b/src/actions/preview.js
@@ -104,6 +104,10 @@ export function setPreview(
           thread: selectedFrame.thread
         });
 
+        if (!result) {
+          return;
+        }
+
         return {
           expression,
           result,


ravefalcon92:
I'll take a look into what's causing this and look for an alternative solution. Thanks for finding the issue and letting me know!

darkwing:
Per the instructions above, hovering over the `n` before `().hash` returns the following object from `await client.evaluateInFrame`:

```js
{
  errorMessageName: "JSMSG_NOT_DEFINED"
  exception: {…}
​
  actor: "server1.conn1.child1/obj285"
​
  class: "Error"
​
  extensible: true
​
  frozen: false
​
  ownPropertyLength: 4
​
  preview: {…}
​
  sealed: false
​
  type: "object"
​
  <prototype>: {…}
  exceptionDocURL: "https://developer.mozilla.org/docs/Web/JavaScript/Reference/Errors/Not_defined?utm_source=mozilla&utm_medium=firefox-console-errors&utm_campaign=default"
  exceptionMessage: "ReferenceError: n is not defined"
  frame: {…}
​
  column: 1
​
  line: 1
​
  source: "debugger eval code"
​
  <prototype>: {…}
  helperResult: null
  input: "n"
  notes: null
  result: {…}
​
  type: "undefined"
​
<prototype>: {…}
resultID: 1551742481317
timestamp: 1551742481317
```

The result for `hash`, which bricks the debugger, is:

```js
{ result: null }
```

Issue #8017, opened 2019-02-22
Offset-inline-* not supported in Chrome


jaril:
Noticed that the close buttons for Watch Expressions and XHR Breakpoints weren't displayed correctly on Chrome. I think this is related to the switch from `offset-inline` => `inset-inline` (#7948), where `inset-inline` is [only supported by Firefox](https://developer.mozilla.org/en-US/docs/Web/CSS/inset-inline-end) for now.

@fvsch is there a way to still make `inset-inline` properties work for Chrome, where we wouldn't run into specificity problems(#7941)?

Firefox | Chrome
------------ | -------------
<img width="320" alt="screen shot 2019-02-22 at 10 32 48 am" src="https://user-images.githubusercontent.com/15959269/53252578-3e98ca00-368d-11e9-845c-b626b7c4748d.png"> | <img width="264" alt="screen shot 2019-02-22 at 10 14 15 am" src="https://user-images.githubusercontent.com/15959269/53251868-a5b57f00-368b-11e9-8875-057a9e64327e.png">

darkwing:
Good spot @jaril.  We could duplicate properties but I don't think it's a priority.

fvsch:
So it turns out that Chrome does support `inset-(inline|block)-(start|end)` since Chrome 69, but it's behind the `chrome://flags/#enable-experimental-web-platform-features` flag (unlike the margin/padding/border logical properties).

One solution would be to add `postcss-bidirection` back in, but the extra generated code for margin/padding/border etc. is overkill just for `inset-inline` support.

Searching through the CSS codebase I'm finding only 6 uses of `inset-inline-*`, and:
- half of them could be replaced with flexblox (we use absolute positioning to place a button at the end of a row), with better results for vertical alignment;
- and for the other half we can probably write the direction-dependent styles by hand, e.g.

```css
[dir="ltr"] .search-field .img.search {
  left: 6px;
}
[dir="rtl"] .search-field .img.search {
  right: 6px;
}
```

jasonLaster:
i propose leaving a note in our local-development.md doc for RTL support that if you're using chrome setting the flag in `chrome://flags/#enable-experimental-web-platform-features`  could help. That's the primary use case

PR #8066, opened 2019-03-04
Align watch expressions and scopes pane

jaril:
These are the Watch Expressions and Scopes changes from the [UX alignment issue](https://github.com/firefox-devtools/ux/issues/45) adapted to the [Event Listeners](#7937). I put them together in the same PR since they split the Object Rep styles between the two of them.

* Changed rows to 20px height
* Changed watched expression font to monospace
* Changed monospace font size to 11px
* Changed pane content indent to 20px
* Changed tree indents to 14px to match arrow spacing
* Removed [inset-inline-end](#8017) selectors 

screen shot 2019-03-04 at 11 07 17 am

@fvsch @mcroud Thoughts?

### DIFF
--- a/packages/devtools-components/src/tree.css
+++ b/packages/devtools-components/src/tree.css
@@ -25,7 +25,6 @@
 
 .tree .tree-node {
   display: flex;
-  padding-top: 1px;
 }
 
 .tree .tree-node:not(.focused):hover {
--- a/packages/devtools-components/src/tree.css
+++ b/packages/devtools-components/src/tree.css
@@ -34,9 +34,15 @@
 .tree-indent {
   display: inline-block;
   width: 12px;
-  margin-inline-start: 5px;
+  margin-inline-start: 3px;
   border-inline-start: 1px solid #a2d1ff;
   flex-shrink: 0;
+}
+
+.debugger .tree-indent {
+  width: 16px;
+  margin-inline-start: 0px;
+  border-inline-start: 0;
 }
 
 /* For non expandable root nodes, we don't have .tree-indent elements, so we declare
--- a/packages/devtools-components/src/tree.css
+++ b/packages/devtools-components/src/tree.css
@@ -25,6 +25,7 @@
 
 .tree .tree-node {
   display: flex;
+  padding-top: 1px;
 }
 
 .tree .tree-node:not(.focused):hover {
@@ -37,11 +38,6 @@
   margin-inline-start: 5px;
   border-inline-start: 1px solid #a2d1ff;
   flex-shrink: 0;
-}
-
-/* Align with expandables siblings (where we have the arrow) */
-.tree-node[data-expandable="false"] .tree-indent:last-of-type {
-  margin-inline-end: 15px;
 }
 
 /* For non expandable root nodes, we don't have .tree-indent elements, so we declare
@@ -63,7 +59,7 @@
   border: 0;
   padding: 0;
   margin-inline-start: 1px;
-  margin-inline-end: 4px;
+  margin-inline-end: 6px;
   transform-origin: center center;
   transition: transform 125ms var(--animation-curve);
   background-color: var(--theme-icon-dimmed-color);
--- a/packages/devtools-components/src/tree.css
+++ b/packages/devtools-components/src/tree.css
@@ -57,8 +57,7 @@
   height: 10px;
   border: 0;
   padding: 0;
-  margin-inline-start: 1px;
-  margin-inline-end: 6px;
+  margin-inline-end: 4px;
   transform-origin: center center;
   transition: transform 125ms var(--animation-curve);
   background-color: var(--theme-icon-dimmed-color);
--- a/packages/devtools-reps/src/object-inspector/components/ObjectInspector.css
+++ b/packages/devtools-reps/src/object-inspector/components/ObjectInspector.css
@@ -43,7 +43,7 @@
 .object-inspector .tree-node .arrow {
   display: inline-block;
   vertical-align: middle;
-  margin-inline-start: -1px
+  margin-inline-start: -1px;
 }
 
 /* Focused styles */
--- a/packages/devtools-reps/src/object-inspector/components/ObjectInspector.css
+++ b/packages/devtools-reps/src/object-inspector/components/ObjectInspector.css
@@ -28,8 +28,12 @@
 }
 
 .tree.object-inspector .block .object-label::before {
-  content: "☲ ";
-  font-size: 1.1em;
+  content: "☲";
+  font-size: 1em;
+  display: inline-block;
+  margin-top: -1px;
+  vertical-align: text-top;
+  padding-inline-start: 3px;
 }
 
 .object-inspector .object-delimiter {
--- a/packages/devtools-reps/src/object-inspector/components/ObjectInspector.css
+++ b/packages/devtools-reps/src/object-inspector/components/ObjectInspector.css
@@ -29,11 +29,10 @@
 
 .tree.object-inspector .block .object-label::before {
   content: "☲";
-  font-size: 1em;
-  display: inline-block;
-  margin-top: -1px;
-  vertical-align: text-top;
-  padding-inline-start: 3px;
+  font-size: 1.1em;
+  display: inline;
+  padding-inline-end: 2px;
+  line-height: 14px;
 }
 
 .object-inspector .object-delimiter {
--- a/packages/devtools-reps/src/object-inspector/components/ObjectInspector.css
+++ b/packages/devtools-reps/src/object-inspector/components/ObjectInspector.css
@@ -34,11 +34,13 @@
 
 .object-inspector .object-delimiter {
   color: var(--theme-comment);
+  white-space: pre-wrap;
 }
 
 .object-inspector .tree-node .arrow {
   display: inline-block;
   vertical-align: middle;
+  margin-inline-start: -1px
 }
 
 /* Focused styles */
--- a/src/components/SecondaryPanes/Expressions.css
+++ b/src/components/SecondaryPanes/Expressions.css
@@ -97,11 +97,11 @@
 }
 
 :root.theme-light .expression-container:hover {
-  background-color: var(--theme-selection-background-hover);
+  background-color: var(--search-overlays-semitransparent);
 }
 
 :root.theme-dark .expression-container:hover {
-  background-color: var(--theme-selection-background-hover);
+  background-color: var(--search-overlays-semitransparent);
 }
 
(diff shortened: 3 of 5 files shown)


jaril:
For now I used 12px sans serif, 11px monospace for this instead of the 11px sans serif, 10px monospace in the spec.

It looked smaller once implemented than I thought so figured I should get some feedback on it first. It's still 14px line height for every row so the change doesn't affect the amount of content displayed on the screen. 

<details><summary><b>11px sans serif, 10px monospace</b></summary>
screen shot 2019-03-04 at 11 12 04 am
</details><br>

<details><summary><b>12px sans serif, 11px monospace</b></summary>
screen shot 2019-03-04 at 11 12 49 am
</details>

fvsch:
> - 11px sans serif, 10px monospace
> - 12px sans serif, 11px monospace

I think it should be 11px sans serif & 11px monospace. This was discussed quite a bit. :)

- By default we use 11px text, whether sans-serif or monospace
- 12px text is our "bigger text", used in tabs and headers

I would welcome attempts to make those fonts bigger, but it should be a more global design effort. E.g. at least in Debugger it should impact all panes (including the Sources pane), and perhaps include a code font-size bump too.

jaril:
Fixed - both monospace and serif fonts are now 11px. 

This also sets the rest of the sans-serif fonts on the secondary panes to 12px.

screen shot 2019-03-04 at 6 20 52 pm

packages/devtools-reps/src/object-inspector/components/ObjectInspector.css
```diff
@@ -28,17 +28,23 @@
...
 
 .tree.object-inspector .block .object-label:before {
-  content: "☲ ";
-  font-size: 1.1em;
+  content: "☲";
+  font-size: 1em;
+  display: inline-block;
+  margin-top: -1px;
+  vertical-align: text-top;
+  padding-inline-start: 3px;
```
fvsch: It looks like you're pushing this character to the right by 3px. How does it work if the rest of the text wraps to a second line? Does it look like a strange first-line indent? If so, maybe those 3px should belong to the parent element or some container?

packages/devtools-reps/src/object-inspector/components/ObjectInspector.css
```diff
@@ -28,17 +28,23 @@
 }
 
 .tree.object-inspector .block .object-label:before {
-  content: "☲ ";
-  font-size: 1.1em;
+  content: "☲";
+  font-size: 1em;
+  display: inline-block;
```
fvsch: Do we need inline-block here?
After you switch to inline-block, this pseudo-element will start to impact the parent's height, so you need a bunch of vertical-align and negative margin tricks. If we can keep `display:inline`, it might work out better?

packages/devtools-reps/src/object-inspector/components/ObjectInspector.css
```diff
@@ -28,17 +28,23 @@
...
+  margin-top: -1px;
+  vertical-align: text-top;
```
fvsch: I've had rather different results with text-top/text-bottom depending on the browser/font/operating system.
I'm not sure we can rely on it for precise UI alignment.

src/components/SecondaryPanes/Expressions.css
```diff
@@ -11,8 +11,11 @@
   margin: 0;
   border: 1px;
   background-color: var(--theme-sidebar-background);
-  font-size: 12px;
```
fvsch: With this rule gone, the input text uses the system font-size in Firefox on Linux. That's 16px in my case, a bit big in the context of DevTools. :)
A `font-size: inherit` or even `font: inherit` might help.

src/components/SecondaryPanes/Expressions.css
```diff
@@ -66,20 +69,32 @@
 
 .expression-container {
   border: 1px;
-  padding: 0.6em 1em 0.6em 0.5em;
+  padding-top: 2px;
+  padding-bottom: 2px;
+  padding-inline-start: 20px;
```
fvsch: You're using 19px above to account for the 1px border. We could use the same here.

src/components/SecondaryPanes/Expressions.css
```diff
@@ -66,20 +69,32 @@
...
 }
 
+.expression-container .tree .tree-node[aria-level="1"] {
+  padding-top: 0px;
+}
+
+.expression-container .tree-node[aria-level="1"] .object-label {
+  font-family: var(--monospace-font-family);
+  line-height: 13px;
+  padding-top: 1px;
```
fvsch: Not a fan of this kind of pixel-shifting (the line-height + padding combo). It tends to work for you on one browser with one font at one zoom level, and not so much in other conditions.

Maybe we could use a 14px or even 16px line-height for all text in those tree nodes?

packages/devtools-components/src/tree.css
```diff
@@ -25,6 +25,7 @@
 
 .tree .tree-node {
   display: flex;
+  padding-top: 1px;
```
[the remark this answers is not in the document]
jaril: Good catch - took this out. Instead I used a 15px line height where applicable, with specific selectors for `Expressions.css` and `Scopes.css`

packages/devtools-components/src/tree.css
```diff
@@ -33,17 +34,11 @@
 
 .tree-indent {
   display: inline-block;
-  width: 12px;
-  margin-inline-start: 5px;
+  width: 16px;
```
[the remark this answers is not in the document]
jaril: 👍

packages/devtools-reps/src/object-inspector/components/ObjectInspector.css
jaril: This was meant to be `padding-inline-end`. Fixed this in a recent commit.

packages/devtools-reps/src/object-inspector/components/ObjectInspector.css
jaril: Good call. Took it out and used `display: inline` to make it simpler and more reliable.

src/components/SecondaryPanes/Expressions.css
jaril: Changed this to `font-size: inherit` instead.

src/components/SecondaryPanes/Expressions.css
jaril: Do we? I wasn't sure about this border - it's not really displayed so it never factors into the measurement, unlike the input form. Setting this to 19px would mean we're short by 1px here.

Let me know if I'm missing anything, otherwise it might be good to take the border value out.

src/components/SecondaryPanes/Expressions.css
jaril: Took this out and set the line-height to 14px instead.

src/components/SecondaryPanes/Expressions.css
fvsch: You're right, let's keep the 20px and remove the `border: 1px` to clean things up a bit.

fvsch:
Tried it out again, and it looks pretty good.

One thing I would like to change is the default margins for the `button.arrow` in the tree component (in Scopes).

That component is used in a bunch of places but I'm not sure we're using a `button.arrow` in all those instances (not in Debugger>Sources, not in Storage where we use pseudo-elements).

I would like to make this change:

```diff
.tree-node button.arrow {
  ...
-  margin-inline-start: 1px;
-  margin-inline-end: 6px;
+  margin-inline-end: 4px;
  ...
}
```

Looking at all the places where we use the arrow icon, the margin between that icon and the next content is either 2px or 4px, but never as large as 6px. We use 4px in the Sources pane. We use 1px or 2px in the Inspector. Going for 4px should be fine here. :)

src/components/SecondaryPanes/Expressions.css
jaril: Took this out.

packages/devtools-components/src/tree.css
```diff
@@ -33,17 +34,11 @@
...
   border-inline-start: 1px solid #A2D1FF;
   flex-shrink: 0;
 }
 
-/* Align with expandables siblings (where we have the arrow) */
-.tree-node[data-expandable="false"] .tree-indent:last-of-type {
```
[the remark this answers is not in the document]
jaril: Ran some experiments and you're right. Storage and Accessibility sets their own `tree-indent`. Console uses `tree.css`'s `.tree-indent`. 


<details><summary><b>IMG: Storage and Accessibility in Firefox build with this PR</b></summary>
<p>I set `.tree-indent { ... }` to `background-color: lime !important` to see if it would show in these panels, and it didn't.</p>


</details><br>

Taking out `:last-of-type` only affects Debugger and Console, and they're both aligned properly as expected after taking it out, so we should be okay.

<details><summary><b>IMG: Tree in Console and Debugger(Sources/Scopes)</b></summary>
<p>I set `.tree-indent { ... }` to `background-color: lime !important`.</p>
<h4>Console</h4>

<h4>Debugger (Sources)</h4>

<h4>Debugger (Scopes)</h4>

</details>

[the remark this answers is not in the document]
fvsch: Well, now we know why `.tree-indent` had a `margin-inline-start`: it centers the border (see Console) under the arrow. So we may need to keep that. :/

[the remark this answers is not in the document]
fvsch: @nchevobbe What do you think, do we expect to have this border style only in Console? Should we define it in webconsole.css?

Alternatively, I would love for this style to be opt-in, using an optional class, e.g.:

```html
<span class="tree-indent tree-indent-border">
```
vs:

```html
<span class="tree-indent">
```

jaril:
Based off of the comment thread:

>Well, now we know why .tree-indent had a margin-inline-start: it centers the border (see Console) under the arrow. So we may need to keep that. :/

I added a new commit 5c87fb7 which addresses this b defining separate `.tree-indent` styles for Console and Debugger panel, without modifying any of the existing classes. This seems to be the simplest solution, though I prefer @fvsch's suggestion of having an opt-in class (e.g. `tree-indent-border`).

This at least fixes the problem for now, and we can easily change the implementation based on @nchevobbe's input.

darkwing:
@fvsch After @jaril fixes the merge conflicts, care to have another look?

darkwing:
This is looking really good so far!  Thoughts @fvsch ?

fvsch:
@darkwing In my last tests it worked pretty well, and I think we caught style shared styled issues in the review. We were waiting on feedback from Nicolas but it's on a CSS refactoring detail, we should go with this PR as-is I think.

PR #8070, opened 2019-03-04
Update index.js

### LINKED ISSUE #7952: [Preview] hovering on the debug line does not leave token
One of the important features of inline preview is that mousing into a token we can preview starts the preview, and mousing out stops the preview. This logic is handled [here][h]

It looks like the debug line does not respect this behavior though.

[h]: https://github.com/devtools-html/debugger.html/blob/master/src/utils/editor/token-events.js#L56

miriambudayr:
Fixes #7952 

### Summary of Changes

* Deleted the part of the conditional that looks for the class name "debug-expression." I looked at Git Blame to find out why this line was added. Since "debug-expression" has no relationship to the highlighted token but instead is the class name added to elements along the entire debug line, there is no reason to include a check for it.

### Test Plan
- [x] Mousing out of a token and onto the debug line does end the preview.
- [x] Mousing out of a token but not onto the debug line does end the preview (previous behavior is not broken).
- [x] Mousing into a token from the debug line starts the preview.
- [x] Mousing out of a token but into a tool tip or popup does not end the preview (previous behavior is not broken).

Thanks @darkwing for confirming that this solution works.

### DIFF
--- a/src/components/Editor/Preview/index.js
+++ b/src/components/Editor/Preview/index.js
@@ -42,8 +42,7 @@ function inPopup(e) {
 
   const pop =
     relatedTarget.closest(".tooltip") ||
-    relatedTarget.closest(".popover") ||
-    relatedTarget.classList.contains("debug-expression");
+    relatedTarget.closest(".popover");
 
   return pop;
 }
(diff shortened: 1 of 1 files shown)


miriambudayr:
@darkwing 

Thank you!

PR #8072, opened 2019-03-05
Fix #7956 - Show falsy values in preview

### LINKED ISSUE #7956: [Preview] should show "false" in the tooltip

darkwing:
Fixes #7956

Good manual testing:  
 - Basic: http://breezy-join.glitch.me/
 - Previously found brick case:  https://github.com/firefox-devtools/debugger/pull/8056#issue-257591385

I also updated the test that had these values commented out.

### DIFF
--- a/src/actions/preview.js
+++ b/src/actions/preview.js
@@ -104,9 +104,10 @@ export function setPreview(
           thread: selectedFrame.thread
         });
 
-        // This specific case occurs for a token that follows an errored evaluation
+        // Error case occurs for a token that follows an errored evaluation
         // https://github.com/firefox-devtools/debugger/pull/8056
-        // Accommodating for this case allows us to show preview for falsy values
+        // Accommodating for null allows us to show preview for falsy values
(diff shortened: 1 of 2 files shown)


darkwing:
cc @ravefalcon92

ravefalcon92:
Just tested the solution on a series of different values and it looks like everything is displaying properly in the preview. Thanks for tracking down that bricking behavior @darkwing!

jasonLaster:
@WenInCode are you noticing that sometimes preview doesn't work the first time?

WenInCode:
@jasonLaster I haven't noticed that. I think that looks like the case in the gif I posted, but it was because the debugger window wasn't actually in focus, my screen recorder was.

WenInCode:
Just double checked on the examples a few times to see if i could get the preview to not work on first try. I was unable to get that issue. The preview worked each time 👍

PR #8076, opened 2019-03-05
Selecting tab returns to prior location

### LINKED ISSUE #7951: [Tabs] selecting a tab should return to the prior location
One of the nice features of editor tabs is that clicking them returns to the prior saved location. This is also a feature of codemirror documents.

This used to work too, but i believe a year or so ago, we made clicking a tab synonmous with going to line 1 column 0. We should fix this so that when we do not specify a location we return to the prior location.

This will also fix another nit, where clicking a tab flashes the first line

dhyey35:
Fixes #7951

This was due to #7581 

### Summary of Changes

* Set default line to `NaN` instead of `1`  as suggested by @jainsneha23 [in 7581](https://github.com/firefox-devtools/debugger/pull/7581#issuecomment-449459608)

### Test Plan

Open two source tabs and scroll in first tab, switch to second one and come back again to first one. The first tab should be opened with the line you scrolled to instead of first line.

debugger-7951-1

### DIFF
--- a/src/utils/location.js
+++ b/src/utils/location.js
@@ -8,14 +8,14 @@ import type { SourceLocation, SourceId } from "../types";
 
 type IncompleteLocation = {
   sourceId: SourceId,
-  line?: number,
+  line?: number | null,
   column?: number,
   sourceUrl?: string
 };
 
 export function createLocation({
   sourceId,
-  line = 1,
+  line = null,
   column,
   sourceUrl = ""
 }: IncompleteLocation): SourceLocation {
--- a/src/actions/sources/select.js
+++ b/src/actions/sources/select.js
@@ -138,6 +138,10 @@ export function selectLocation(
     }
 
     const tabSources = getSourcesForTabs(getState());
+    location = {
+      ...location,
+      line: source.wasFocused ? NaN : 1
+    };
     if (!tabSources.includes(source)) {
       dispatch(addTab(source));
     }
--- a/src/actions/sources/select.js
+++ b/src/actions/sources/select.js
@@ -24,6 +24,7 @@ import { prefs } from "../../utils/prefs";
 import { shouldPrettyPrint, isMinified } from "../../utils/source";
 import { createLocation } from "../../utils/location";
 import { getMappedLocation } from "../../utils/source-maps";
+import { hasDocument } from "../../utils/editor/source-documents";
 
 import {
   getSource,
@@ -67,10 +68,7 @@ export const clearSelectedLocation = () => ({
  * @memberof actions/sources
  * @static
  */
-export function selectSourceURL(
-  url: string,
-  options: PartialPosition = { line: 1 }
-) {
+export function selectSourceURL(url: string, options: PartialPosition = {}) {
   return async ({ dispatch, getState, sourceMaps }: ThunkArgs) => {
     const source = getSourceByURL(getState(), url);
     if (!source) {
@@ -87,10 +85,7 @@ export function selectSourceURL(
  * @memberof actions/sources
  * @static
  */
-export function selectSource(
-  sourceId: string,
-  options: PartialPosition = { line: 1 }
-) {
+export function selectSource(sourceId: string, options: PartialPosition = {}) {
   return async ({ dispatch }: ThunkArgs) => {
     const location = createLocation({ ...options, sourceId });
     return dispatch(selectSpecificLocation(location));
@@ -138,14 +133,16 @@ export function selectLocation(
     }
 
     const tabSources = getSourcesForTabs(getState());
-    location = {
-      ...location,
-      line: source.wasFocused ? NaN : 1
-    };
     if (!tabSources.includes(source)) {
       dispatch(addTab(source));
     }
 
+    location = {
+      ...location,
+      // Select first line if opening file for first time otherwise
+      // the line passed or the one user scrolled to
+      line: !location.line && hasDocument(source.id) ? null : location.line || 1
+    };
     dispatch(setSelectedLocation(source, location));
 
     await dispatch(loadSourceText(source));
--- a/src/reducers/sources.js
+++ b/src/reducers/sources.js
@@ -120,14 +120,8 @@ function update(
         prefs.pendingSelectedLocation = location;
       }
 
-      const existingSource = state.sources[action.source.id];
-
       return {
         ...state,
-        sources: {
-          ...state.sources,
-          [action.source.id]: { ...existingSource, wasFocused: true }
-        },
         selectedLocation: {
           sourceId: action.source.id,
           ...action.location
--- a/src/reducers/sources.js
+++ b/src/reducers/sources.js
@@ -120,8 +120,14 @@ function update(
         prefs.pendingSelectedLocation = location;
       }
 
+      const existingSource = state.sources[action.source.id];
+
       return {
         ...state,
+        sources: {
+          ...state.sources,
+          [action.source.id]: { ...existingSource, wasFocused: true }
+        },
         selectedLocation: {
           sourceId: action.source.id,
           ...action.location
--- a/src/types.js
+++ b/src/types.js
@@ -352,7 +352,8 @@ type BaseSource = {|
   +relativeUrl: string,
   +introductionUrl: ?string,
   +isExtension: boolean,
-  +actors: SourceActor[]
+  +actors: SourceActor[],
+  wasFocused?: boolean
 |};
 
 /**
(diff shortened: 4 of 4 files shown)


jasonLaster:
I wonder if we can add a mochitest to prevent this from regressing

jasonLaster:
Nice work hunting down the root cause.

dhyey35:
@jasonLaster Would love to add a test. I will learn how to use mochitest and revert back in a couple of days

src/utils/location.js
```diff
@@ -15,7 +15,7 @@ type IncompleteLocation = {
 
 export function createLocation({
   sourceId,
-  line = 1,
+  line = NaN,
```
jainsneha23: Instead of setting it to NaN, can we change the type of line to be a number or null? Which wd make more sense. And then we wont pass the line for creating a new location

darkwing: I agree that `NaN` doesn't feel right; null or undefined would be fine with me too.

We should then also see what happens when we click a filename heading in the breakpoints pane.

Ideally we leave scroll position alone unless a specific line has reason for focus, like clicking a breakpoint or pausing.

darkwing:
I made a small modification:

```diff
diff --git a/src/utils/location.js b/src/utils/location.js
index c08da62ef..37d3de02e 100644
--- a/src/utils/location.js
+++ b/src/utils/location.js
@@ -8,14 +8,14 @@ import type { SourceLocation, SourceId } from "../types";
 
 type IncompleteLocation = {
   sourceId: SourceId,
-  line?: number,
+  line?: number | null,
   column?: number,
   sourceUrl?: string
 };
 
 export function createLocation({
   sourceId,
-  line = NaN,
+  line = null,
   column,
   sourceUrl = ""
 }: IncompleteLocation): SourceLocation {

```

The last issue I saw was:
1.  Open a source with many lines of code (unminified)
2. Scroll down to the bottom
3.  Open another source with many lines of code
4.  The newly opened source is scrolled down to the previous source's scroll position; since the source has just been opened, it should be focused on line 1

AnshulMalik:
It will be a nice improvement :)

Can we do this? If the source we are opening is not already open, we say the line number as `1`, otherwise `null`?

dhyey35:
@darkwing I tried that but that gives me flow error

```
yarn flow
yarn run v1.12.3
$ flow
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ src/utils/location.js:18:3

null [1] is incompatible with number [2].

 [2][1] 11│   line?: number | null,
        12│   column?: number,
        13│   sourceUrl?: string
        14│ };
        15│
        16│ export function createLocation({
        17│   sourceId,
        18│   line = null,
        19│   column,
        20│   sourceUrl = ""
        21│ }: IncompleteLocation): SourceLocation {


Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ src/utils/location.js:18:10

null [1] is incompatible with number [2].

 [2] 11│   line?: number | null,
     12│   column?: number,
     13│   sourceUrl?: string
     14│ };
     15│
     16│ export function createLocation({
     17│   sourceId,
 [1] 18│   line = null,
     19│   column,
     20│   sourceUrl = ""
     21│ }: IncompleteLocation): SourceLocation {



Found 2 errors
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

I also tried replacing all the `line: number` in [types.js](https://github.com/firefox-devtools/debugger/blob/master/src/types.js) with `line: ?number` but than flow complains :

```
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ src/workers/parser/mapOriginalExpression.js:34:13

Cannot coerce start.line to string because null or undefined [1] should not be coerced.

     src/workers/parser/mapOriginalExpression.js
     31│ }
     32│
     33│ function locationKey(start: Position): string {
     34│   return `${start.line}:${start.column}`;
     35│ }
     36│
     37│ export default function mapOriginalExpression(

     src/types.js
 [1] 80│   +line: ?number,



Found 61 errors

```

darkwing:
The following solution works but is far from ideal:

```diff
diff --git a/src/actions/sources/select.js b/src/actions/sources/select.js
index a37473540..6f13ed40e 100644
--- a/src/actions/sources/select.js
+++ b/src/actions/sources/select.js
@@ -67,10 +67,7 @@ export const clearSelectedLocation = () => ({
  * @memberof actions/sources
  * @static
  */
-export function selectSourceURL(
-  url: string,
-  options: PartialPosition = { line: 1 }
-) {
+export function selectSourceURL(url: string, options: PartialPosition = {}) {
   return async ({ dispatch, getState, sourceMaps }: ThunkArgs) => {
     const source = getSourceByURL(getState(), url);
     if (!source) {
@@ -87,10 +84,7 @@ export function selectSourceURL(
  * @memberof actions/sources
  * @static
  */
-export function selectSource(
-  sourceId: string,
-  options: PartialPosition = { line: 1 }
-) {
+export function selectSource(sourceId: string, options: PartialPosition = {}) {
   return async ({ dispatch }: ThunkArgs) => {
     const location = createLocation({ ...options, sourceId });
     return dispatch(selectSpecificLocation(location));
diff --git a/src/utils/location.js b/src/utils/location.js
index 2e4ea9b2f..54a3a9b7f 100644
--- a/src/utils/location.js
+++ b/src/utils/location.js
@@ -5,20 +5,27 @@
 // @flow
 
 import type { SourceLocation, SourceId } from "../types";
+import { hasDocument } from "./editor/source-documents";
 
 type IncompleteLocation = {
   sourceId: SourceId,
-  line?: number,
+  line?: number | null,
   column?: number,
   sourceUrl?: string
 };
 
 export function createLocation({
   sourceId,
-  line = 1,
+  line = null,
   column,
   sourceUrl = ""
 }: IncompleteLocation): SourceLocation {
+  // Ensures new documents are opened at line 1 and not the line of
+  // the previous document.
+  if (!hasDocument(sourceId)) {
+    line = 1;
+  }
+
   return {
     sourceId,
     line,

```

This solution isn't ideal because of a lack of separation of concerns -- `createLocation` shouldn't care about documents at all.

@jasonLaster had mentioned the scrolling issues could be due to CodeMirror sharing documents, which at a quick look didn't seem to be the case, but it's worth looking into again.

dhyey35:
@AnshulMalik I tried doing it simply by setting line to `NaN` ( can't get `null` to pass flow tests ) in else of `if (!tabSources.includes(source))` [here](https://github.com/firefox-devtools/debugger/blob/master/src/actions/sources/select.js#L141) but that breaks when we reload the tab and few files are already open in the debugger ( It counts them as already opened files ).

So I decided to add a property `wasFocused` to all the sources who were at least focused once, when they are opened from 2nd time on wards they will get `NaN` instead of `1`.

I also tried storing start line number using `getLocationsInViewport(getEditor())` on `scroll` but setting line number actually doesn't give us start line as the line number we pass.

dhyey35:
@darkwing I applied your patches locally and I noticed that it also has the issue that you mentioned

> The last issue I saw was:
> 
> 1. Open a source with many lines of code (unminified)
> 2. Scroll down to the bottom
> 3. Open another source with many lines of code
> 4. The newly opened source is scrolled down to the previous source's scroll position; since the source has just been opened, it should be focused on line 1

AnshulMalik:
We can use the tabs stored in store to check if the source is already open or not, so that way, we can skip adding additional property and things should hopefully work out :)

darkwing:
Closing in favor of: https://github.com/firefox-devtools/debugger/pull/8102

Thank you for having a look @dhyey35 !

PR #8077, opened 2019-03-05
Apply SourceMaps type to redux actions

### LINKED ISSUE #7925: [flow] type sourceMaps in actions
@Mrigank11 recently helped us add flow types for our source map workers [PR]. It would be really great if we could now use that type in our redux actions.

[PR]: https://github.com/devtools-html/debugger.html/commit/17d4a0f44f4d01b64016a49a28904f8041c6d9b6

I started it here and saw 17 warnings which look like legitimate flow warnings

```diff
diff --git a/src/actions/types/index.js b/src/actions/types/index.js
index 1285d332e..4f2d73837 100644
--- a/src/actions/types/index.js
+++ b/src/actions/types/index.js
@@ -17,6 +17,7 @@ import type { PauseAction } from "./PauseAction";
 import type { ASTAction } from "./ASTAction";
 import { clientCommands } from "../../client/firefox";
 import type { Panel } from "../../client/firefox/types";
+import typeof SourceMaps from "../../../packages/devtools-source-map/src";

 /**
  * Flow types
@@ -34,7 +35,7 @@ export type ThunkArgs = {
   dispatch: (action: any) => Promise<any>,
   getState: () => State,
   client: typeof clientCommands,
-  sourceMaps: any,
+  sourceMaps: SourceMaps,
   panel: Panel
 };
```

ryanjduffy:
Fixes #7925 

### Summary of Changes

* Change instances of source maps to use `typeof SourceMaps` instead of `any`/`Object`
* Add a couple more types to address flow inference issue (see below)

### Considerations

I couldn't find a way to import the `typeof` the default export of a module alongside named imports from the same module.

```
import typeof SourceMaps, { isGeneratedId } from "devtools-source-map";
```

Originally, I included two imports and suppressed the eslint error but changed when I realized the frequency this would be required. The current approach is to import the default export and use `typeof` in the instance typing later.

```
import typeof SourceMaps, { isGeneratedId } from "devtools-source-map";
// later
const sourceMaps: typeof SourceMaps = ...;
```

I also ran into a couple issues that seem to trace back to flow. I've documented them inline but noting them here for traceability in GitHub.

Explicitly typing an array wrapped by a Promise
https://github.com/facebook/flow/issues/5294

Unable to correctly type the result of a spread on a union type.
https://github.com/facebook/flow/pull/7298


darkwing:
Thank you for kick-starting this!  Do you plan on continuing this PR?

ryanjduffy:
I do! Thanks for the reminder, @darkwing!

src/actions/pause/mapFrames.js
```diff
@@ -14,11 +14,13 @@ import {
 import assert from "../../utils/assert";
 import { findClosestFunction } from "../../utils/ast";
 
-import type { Frame, ThreadId } from "../../types";
+import type { Frame, OriginalFrame, ThreadId } from "../../types";
 import type { State } from "../../reducers/types";
 import type { ThunkArgs } from "../types";
 
+/* eslint-disable import/no-duplicates */
```
ryanjduffy: I couldn't sort out a way to import `typeof SourceMaps` and `{ isGeneratedId }` in a way that made everything happy. I elected to suppress the linting error since the code style deviation was necessary for typing. Another alternative would be to import the default export and create a type from it but importing "real" code just to then create a type seemed "more bad" to me.

```
import DevToolsSourceMap, {isGeneratedId} from "devtools-source-map";
type SourceMaps = typeof DevToolsSourceMap;
```

src/actions/pause/mapFrames.js
```diff
@@ -130,6 +132,10 @@ async function expandFrames(
     };
 
     originalFrames.forEach((originalFrame, j) => {
+      if (!originalFrame.location || !originalFrame.thread) {
```
ryanjduffy: Not sure if this is the right approach. From what I read in `XScope.search()`, `displayName` will always be present and `location` may be present but `thread` shouldn't be. However, it was accessed below so perhaps I've misread.

src/utils/source-maps.js
```diff
@@ -94,7 +94,7 @@ export async function mapLocation(
     return getGeneratedLocation(state, source, location, sourceMaps);
   }
 
-  return sourceMaps.getOriginalLocation(location, source);
```
ryanjduffy: Not sure why `source` was passed here but seems invalid given the signature of `getOriginalLocation`.

ryanjduffy:
@jasonLaster + @darkwing - I think this is ready for review. Lemme know what you think.

jasonLaster:
quick scan. this looks really good.

src/utils/source-maps.js
jasonLaster: looks like a good catch

Issue #7905, opened 2019-02-07
Debugger CSS uses global classNames, risks conflicts


fvsch:
Working in the CSS codebase, I'm seeing a lot of CSS class names that are both global and very generic:

```css
/* SomeSpecificComponent.css */

.pane {
  /* a bunch of styles */
}

.title {
  /* a bunch of styles */
}

.bottom {
  /* a bunch of styles */
}
```

Sometimes we have class names following a naming convention like BEM or SMACSS or Bootstrap-like names (`.component`, `.component-variant`, `.component-child`), or a "root" class named is used to scope styles (`.component .child`), but that's far from systematic.

CSS selectors share a global namespace. We can see CSS classes as functions, and `<element class="myClass">` as a call site of the `myClass` function. So a lot of the Debugger CSS is the rough equivalent of writing JS like this:

```js
window.pane = function() { ... }
window.title = function() { ... }
window.bottom = function() { ... }
```

I also wonder if contributors to this project may end up publishing very generic names to the global namespace because they have an expectation that the styles they write are scoped to the React component. After all, most stylesheets are named after a React component and intended for styling that component:

```
src/components/PrimaryPanes/Outline.css
src/components/PrimaryPanes/Outline.js
```

So it's fairly logical for the unsuspecting developer to think that since Outline.js is a module with its own local scope, then Outline.css is itself a module with a local scope.

This creates a few issues:

1. We may run into issues when we add generic class names to elements, for instance if I create `<span class="arrow bottom">` somewhere I'm going to inherit the styles for bottom toolbars! Woops.
2. We have a lot of selector specificity fighting going on, with long selectors meant to ramp up the number of classes in the selector so those styles override some other styles.
3. For shared components such as reps or frames, we end up putting DOM in the Console or other consumers that has very short and generic class names, for instance:
```html
<div class="frame">
  <span class="title">React</span>
  <span class="img react"></span>
  <span class="file">...</span>
  <span class="path">...</span>
</div>
```
Are we sure that webconsole.css doesn't already define styles for `.frame`, `.title`, `.img`, `.react`, `.file` or `.path`? Nope. (Those are not the exact class names used but it's similar.)
Also if we export the styles for this DOM and inject it in the Console, we may end up styling completely unrelated elements.

[...]

PR #8078, opened 2019-03-05
#7993 Fix callstack frame-group icon and avoid global selectors

### LINKED ISSUE #7993: Webpack Icon Cut Off in Call Stack Pane
Just spotted this:

webpackcallstack

fvsch:
- Fixes #7993
- Fixes global selectors in SecondaryPanes/Frames (#7905)

Result looks like this:

### React

Monochrome icons use the label's text color. We had some outdated targetting SVG paths that was doing that, so I replicated the effect:

react group 1

react group 2

### Webpack

Multicolor icons don't change depending on the normal/expanded state:

webpack group 1

webpack group 2

### DIFF
--- a/src/components/SecondaryPanes/Frames/Group.css
+++ b/src/components/SecondaryPanes/Frames/Group.css
@@ -2,8 +2,8 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
 
-.frames [role="list"] .frames-group .group,
-.frames [role="list"] .frames-group .group .location {
+.frames-group .group,
+.frames-group .group .location {
   font-weight: 500;
   cursor: default;
   /*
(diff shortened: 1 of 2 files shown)


darkwing:
@nchevobbe I know you had some issues with frame icons so I'd like your review of this if possible.  Thank you!

src/components/SecondaryPanes/Frames/Group.css
```diff
@@ -14,24 +14,33 @@
...
-.frames [role="list"] .frames-group.expanded .group .location {
+.frames-group.expanded .group,
+.frames-group.expanded .group .location {
   color: var(--theme-highlight-blue);
 }
 
-.frames [role="list"] .frames-group .frames-list [role="listitem"] {
-  padding-left: 30px;
+.frames-group .frames-list [role="listitem"] {
+  padding-inline-start: 32px;
```
nchevobbe: is the size change for better alignment?

src/components/SecondaryPanes/Frames/Group.css
```diff
@@ -14,24 +14,33 @@
...
   color: var(--theme-highlight-blue);
 }
 
-.group-description-name {
-  padding-left: 5px;
+.frames-group .img.arrow {
+  margin-inline-start: -1px;
+}
+
+.frames-group .img.annotation-logo {
```
nchevobbe: I think we can have `.annotation-logo` not in `.frames-group`, for example when you have only 1 library frame. Is this change (and the related one in `Frames.css`) impact those?

src/components/SecondaryPanes/Frames/Group.css
fvsch: Yep, it works a bit better in Debugger's call stack.
Nothing important so I can revert it if it's a concern.

fvsch: Reverted it for now, we have some UI work in progress for Debugger panes so we can always revisit later.

src/components/SecondaryPanes/Frames/Group.css
fvsch: It looks like you are rolling your own styles in `devtools/client/shared/components/SmartTrace.css`?
So this probably won't affect Console. But I'll move it back to Frames.css (scoped with `.frames` instead of `.frames-group`) to be on the safe side.

PR #8079, opened 2019-03-05
fix: make search input more responsive

### LINKED ISSUE #8003: [Search] Results clip the modifiers
It looks like the search results clip the modifiers so they do not show. I wonder if we can use flexbox here so that we guarantee that modifiers always appear.

---

screen shot 2019-02-20 at 12 41 33 pm

---
screen shot 2019-02-20 at 12 41 28 pm

th7nder:
Fixes #8003 

### Summary of Changes

* change base width of a search input

### Test Plan
- [x] Launch search bar on any file
- [x] Resize source pane and check behaviour of the search input

### DIFF
--- a/src/components/shared/SearchInput.css
+++ b/src/components/shared/SearchInput.css
@@ -58,6 +58,7 @@
   flex-grow: 1;
   height: 24px;
   min-width: 40px;
+  width: 40px;
   border: none;
   padding: 4px;
   padding-inline-start: 28px;
--- a/src/components/shared/SearchInput.css
+++ b/src/components/shared/SearchInput.css
@@ -57,7 +57,6 @@
   align-self: stretch;
   flex-grow: 1;
   height: 24px;
-  min-width: 40px;
   width: 40px;
   border: none;
   padding: 4px;


src/components/shared/SearchInput.css
```diff
@@ -58,6 +58,7 @@
   flex-grow: 1;
   height: 24px;
   min-width: 40px;
+  width: 40px;
```
darkwing: Functionally this works great but could you explain how `width` is stretching?  Because of the `flex-grow`/`align-self`?

Should we remove the `min-width`?

Awesome work @th7nder~

th7nder: That's a good question. 
Inputs have **a default width,** which is set as a `flex-basis`. I thought, that setting `flex-grow: 1` to allow growth is enough, because `flex-shrink: 1` is set by default and the item will be responsive. 
I have experimented a bit and it turns out that the default `flex-shrink` _does not work on inputs_, thus allowing us only to **grow** an input's width.
This malfunction forces us to set `width` to some minimum value.

`align-self` property affects only the cross axis, it stretches an item vertically.
I think you are right, `min-width` is not necessary there.

darkwing:
Awesome, thank you @th7nder !

PR #8080, opened 2019-03-05
Fix 7674 - Ensure breakpoint text is always highlighted

### LINKED ISSUE #7674: Breakpoint Pane Code Not Highlighted Upon Re-Opening DevTools
Rough STR:

- Open Debugger
- Add a bunch of breakpoints
- Close DevTools
- Open Debugger again

What I saw:

- All breakpoints list items are there but code is *not* syntax highlighted.  No source was open.
- When I opened a source, the breakpoints pane becomes syntax highlighted

I wonder if it's because we're lazyloading(?) CodeMirror?

darkwing:
Fixes #7674 

Breakpoints aren't highlighted if the debugger is opened without a tab open.  To reproduce, you can open the debugger, add a breakpoint, close the source, close devtools, and then reopen the debugger.

This is mostly a POC to get feedback.  There isn't really a pretty way to do this.

### DIFF
--- a/src/components/SecondaryPanes/Breakpoints/Breakpoint.js
+++ b/src/components/SecondaryPanes/Breakpoints/Breakpoint.js
@@ -22,6 +22,7 @@ import {
 import { getSelectedLocation } from "../../../utils/source-maps";
 import { features } from "../../../utils/prefs";
 import { getEditor } from "../../../utils/editor";
+import { createEditor } from "../../../utils/editor/create-editor";
 
 import type {
   Breakpoint as BreakpointType,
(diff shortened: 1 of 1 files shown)


darkwing:
I suppose we could also create the fake editor at <Breakpoints> level, then pass to children.

jasonLaster:
perhaps we want a mechanism to use a headless codeMirror.

cleanup might be a thing too..

darkwing:
This was a good research prototype but we'll need to think more about how to use a headless CodeMirror efficiently.

Issue #8081, opened 2019-03-05
Dropdown icon disappears when filename is too long


fvsch:
The source tabs dropdown items have a few rendering issues since they were switched to `display: flex`.

1. Icon can shrink to 0px width
2. Ellipsis style doesn't work

![dropdown screenshot](https://user-images.githubusercontent.com/243601/53839115-fa63ce80-3f96-11e9-9b09-6f8f0a365812.png)

PR #8083, opened 2019-03-05
Fix #8017 - Work around unsupported inset-inline-start|end in Chrome

### LINKED ISSUE #8017: Offset-inline-* not supported in Chrome

fvsch:
Fixes #8017

I'm setting left:x and right:x properties manually for the `ltr` and `rtl` directions.
Absolute positioning is rare in Debugger, so it's not a lot of work.

I've left two uses of `inset-inline-start|end` around:

- `WelcomeBox.css`: that usage should be removed in #8039
- `Dropdown.css`: addressed in #8081

### DIFF
--- a/src/components/SecondaryPanes/Breakpoints/Breakpoints.css
+++ b/src/components/SecondaryPanes/Breakpoints/Breakpoints.css
@@ -180,10 +180,17 @@ html .breakpoints-list .breakpoint.paused {
 }
 
 .breakpoint .close-btn {
-  inset-inline-end: 15px;
-  inset-inline-start: auto;
   position: absolute;
-  top: -100px; /*For hiding button outside of row until hovered or focused*/
+  /* hide button outside of row until hovered or focused */
+  top: -100px;
+}
+
+[dir="ltr"] .breakpoint .close-btn {
+  right: 12px;
+}
+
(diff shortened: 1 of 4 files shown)


darkwing:
Care to have a look @jaril?

jaril:
Just checked and the alignment issues on Chrome are now all gone. 👍 

The PR (#8066) I'm working will have conflicts with this, but I can just rebase before merging. That way, #8017 can be addressed and fixed right away. 

The redesign uses flexbox for alignment so absolute positioning shouldn't be a concern afterwards. #8066 will affect `Expressions.css`, and I've got a second PR coming tonight that will affect `Breakpoints.css` and `XHRBreakpoints.css`

PR #8086, opened 2019-03-06
Fix #8084 - Vue.js icon missing in Frames

### LINKED ISSUE #8084: [CallStack] Vue icons don't display on frames
I spotted this in both Chrome and Firefox.

STR:
1. Debug http://todomvc.com/examples/vue/
2. In debugger, open the app.js file in the editor pane.
screen shot 2019-03-05 at 6 57 52 pm

3. Set a breakpoint at line 74, within the `addTodo` method.
4. On Vue todomvc site, add a todo.

screen shot 2019-03-05 at 6 05 08 pm
 
I'm not sure if this bug is in any planned future polish.

fvsch:
Fixes #8084

Ideally it would be great if we could return the framework name as `"Vue.js"` and the className as `vuejs`, rather than supporting two names (`"Vue"` and `"VueJS"`) and two classes (`"vue"` and `"vuejs"`), but I'm not sure how to do that and the CSS fix seems less risky.

### DIFF
--- a/src/components/SecondaryPanes/Frames/Frames.css
+++ b/src/components/SecondaryPanes/Frames/Frames.css
@@ -107,6 +107,13 @@
   background-color: currentColor;
 }
 
+/*
+ * We also show the library icon in locations, which are forced to RTL.
+ */
+.frames .location .img.annotation-logo {
+  margin-inline-start: 4px;
+}
+
 /* Some elements are added to the DOM only to be printed into the clipboard
    when the user copy some elements. We don't want those elements to mess with
    the layout so we put them outside of the screen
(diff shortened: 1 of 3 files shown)


darkwing:
I see the icon but it's cut off -- did we have a patch for that?

fvsch:
@darkwing That would be #8078, merged earlier today.

fvsch:
@darkwing The icon should not be cut off if it's the right size (default 16px).

Except maybe for SVG icons which are not 16px (in the source SVG) and use background-image _and_ are not SourceIcons, like some of those.

I think that was discussed in other bug, not sure which, and we said it was safer to add the `background-size: contain;` and related styles to `.img` rather than to just `.source-icon`. So I just added that to the PR.

darkwing:
I love it!  Thank you for catching this!

PR #8087, opened 2019-03-06
Fix #8040 - Correct Accordion header spacing and restore thin plus.svg icon

### LINKED ISSUE #8040: Fix accordion spacing (#7735 follow up)
The accordion CSS changes in #7735 were backed out due to mochitest failures.
As a result we have a few spacing issues in accordions, especially when it comes to the "+" button alignment and the space between the arrow icon and the label.

@mcroud I think you had a UI spec for spacing in accordion headers? Could you share it here?

fvsch:
This restores:
- the old "+" icon, to avoid shipping a lonely Photon icon in Firefox 67
- correct padding in the accordion header ()

Small improvements:
- Small improvements to spacing to match Matt's spec
- Fix button alignment in RTL
- Add spacing between buttons (e.g. in Watch expressions)
- If we are short on room, cut the label, instead of letting the text push the buttons to the right (sometimes outside of the header).

ellipsis screenshot

__Testing plan:__

We need to test this with mochitest because last time we had similar changes to accordion headers, some tests broke. :(

A real possibility is that removing the `height:24px` (so that the header can be 25px, including the bottom border) shuffles some content around and a synthetic click misses. Or there is a synthetic click near the end of the header but the "+" button moved a little bit and this misses. Something like that. Help welcome. :D

### DIFF
--- a/images/plus.svg
+++ b/images/plus.svg
@@ -2,5 +2,5 @@
    - License, v. 2.0. If a copy of the MPL was not distributed with this
    - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
 <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 16 16" width="16" height="16">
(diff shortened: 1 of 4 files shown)


darkwing:
Tested and works/looks excellent!  Thank you @fvsch !

Issue #50, opened 2016-04-12
Investigate Travis CI for running tests


clarkbw:
We want Travis to run tests against every PR however currently it is only running lint tests.

In order to run our unit tests we need the xpcshell binary available on Travis.

There are several ways to possibly accomplish this:

- Download an xpcshell binary (that matches platform) from somewhere and use this to run test
- Build Firefox from source to create the xpcshell binary to run the tests
- Pull a copy of xpcshell into our own repo which could run the tests for us

clarkbw:
Ok, here's what I've found.

From a URL like this: 
https://archive.mozilla.org/pub/firefox/nightly/latest-mozilla-central/

You could download an SDK build like this:
https://archive.mozilla.org/pub/firefox/nightly/latest-mozilla-central/firefox-48.0a1.en-US.linux-i686.sdk.tar.bz2

In that build you'd find xpcshell under `sdk/bin/xpcshell`.  I don't know if it is exactly what we're looking for but it seems pretty close.  I'm going to try a test build and then maybe a test Travis build.

jasonLaster:
Awesome @clarkbw.

jlongster:
@clarkbw neat! that might work. we would need to point the runner to that whole "obj dir" so it can find all the dynamic libraries it needs, but I assume they are all there because regular Firefox needs them too. 

The last bit would be how to get the `head.js` file from the source, which is what we use the "src dir" option for, but if that's really the last piece we can just copy that in and commit it.

clarkbw:
> we would need to point the runner to that whole "obj dir" so it can find all the dynamic libraries it needs

Ah, that is making sense.  I don't know how to run the `xpcshell` command so I'm just getting this error.  But I think we're on the right track here.

```
$ ./run-mozilla.sh ./xpcshell
dyld: Library not loaded: @executable_path/libnss3.dylib
  Referenced from: /Users/clarkbw/Downloads/mac-fx-sdk/sdk/bin/xpcshell
  Reason: image not found
Trace/BPT trap: 5
```

clarkbw:
Closing this out now that #60 has landed and we're going away from xpcshell.

PR #8094, opened 2019-03-07
Aligned breakpoints and XHR breakpoints panes

jaril:
These are the Breakpoints and XHR Breakpoints changes from the [UX Alignment](https://github.com/firefox-devtools/ux/issues/45) issue. I put them together in the same PR since they share some styles in `Breakpoints.css`.

### Summary of Changes

- Changed breakpoints, XHR breakpoint, and option rows to 20px height
- Changed font sizes to 11px
- Changed pane content indentation to 20px
- Changed nesting indentation to 16px
- Removed inset-inline-end selector for XHR Breakpoints

screen shot 2019-03-06 at 10 58 45 pm

### DIFF
--- a/src/components/SecondaryPanes/Breakpoints/Breakpoints.css
+++ b/src/components/SecondaryPanes/Breakpoints/Breakpoints.css
@@ -10,6 +10,11 @@
   user-select: none;
 }
 
+.breakpoints-list {
+  margin-top: 4px;
+  margin-bottom: 4px;
+}
+
 .breakpoints-list .breakpoint-heading {
   text-overflow: ellipsis;
   overflow: hidden;
@@ -18,10 +23,6 @@
   align-items: center;
   padding-top: 3px;
   padding-bottom: 3px;
-}
-
-.breakpoints-list .breakpoint-heading:nth-child(2) {
-  margin-top: 4px;
 }
 
 .breakpoints-list .breakpoint-heading .filename {
@@ -59,10 +60,6 @@
   padding-bottom: 3px;
   padding-top: 3px;
   user-select: none;
-}
-
-.breakpoints-list .breakpoint:last-child {
-  margin-bottom: 4px;
 }
 
 .breakpoints-exceptions-caught {
--- a/src/components/SecondaryPanes/Breakpoints/Breakpoints.css
+++ b/src/components/SecondaryPanes/Breakpoints/Breakpoints.css
@@ -17,18 +17,18 @@
 
 .breakpoints-list .breakpoint-heading {
   text-overflow: ellipsis;
-  overflow: hidden;
-  display: flex;
   width: 100%;
-  align-items: center;
-  padding-top: 3px;
-  padding-bottom: 3px;
+  font-size: 12px;
+  line-height: 16px;
+}
+
+.breakpoint-heading:not(:first-child) {
+  margin-top: 2px;
 }
 
 .breakpoints-list .breakpoint-heading .filename {
   overflow: hidden;
   text-overflow: ellipsis;
-  line-height: 14px;
 }
 
 .breakpoints-list .breakpoint-heading .filename span {
--- a/src/components/SecondaryPanes/Breakpoints/Breakpoints.css
+++ b/src/components/SecondaryPanes/Breakpoints/Breakpoints.css
@@ -6,10 +6,6 @@
   margin: 2px 3px;
 }
 
-.pane.breakpoints-list {
-  padding-bottom: 0.35em;
-}
-
 .breakpoints-list * {
   user-select: none;
 }
@@ -20,11 +16,18 @@
   display: flex;
   width: 100%;
   align-items: center;
+  padding-top: 3px;
+  padding-bottom: 3px;
+}
+
+.breakpoints-list .breakpoint-heading:nth-child(2) {
+  margin-top: 4px;
 }
 
 .breakpoints-list .breakpoint-heading .filename {
   overflow: hidden;
   text-overflow: ellipsis;
+  line-height: 14px;
 }
 
 .breakpoints-list .breakpoint-heading .filename span {
@@ -34,10 +37,8 @@
 
 .breakpoints-list .breakpoint-heading,
 .breakpoints-list .breakpoint {
-  font-size: 12px;
   color: var(--theme-content-color1);
   position: relative;
-  transition: all 0.25s ease;
   cursor: pointer;
 }
 
@@ -45,46 +46,66 @@
 .breakpoints-list .breakpoint,
 .breakpoints-exceptions,
 .breakpoints-exceptions-caught {
-  padding: 0.25em 1em;
+  padding-inline-start: 16px;
+  padding-inline-end: 12px;
 }
 
 .breakpoints-exceptions {
-  padding-bottom: 0.5em;
-  padding-top: 0.5em;
+  padding-bottom: 3px;
+  padding-top: 3px;
   user-select: none;
 }
 
 .breakpoints-list .breakpoint {
-  min-height: var(--breakpoint-expression-height);
   overflow: hidden;
+  padding-top: 2px;
+  padding-bottom: 2px;
+}
+
+.breakpoints-list .breakpoint:last-child {
+  margin-bottom: 4px;
 }
 
 .breakpoints-exceptions-caught {
-  padding: 0 1em 0.5em 3em;
-  margin-top: -0.25em;
+  padding-bottom: 3px;
+  padding-top: 3px;
+  padding-inline-start: 36px;
 }
 
 html[dir="rtl"] .breakpoints-exceptions-caught {
   padding: 0 3em 0.5em 1em;
 }
 
+.breakpoints-exceptions-options {
+  padding-top: 4px;
+  padding-bottom: 4px;
+}
+
+.xhr-breakpoints-pane .breakpoints-exceptions-options {
+  border-bottom: 1px solid var(--theme-splitter-color);
+}
+
 .breakpoints-exceptions-options:not(.empty) {
   border-bottom: 1px solid var(--theme-splitter-color);
-  margin-bottom: 3px;
 }
 
 .breakpoints-exceptions input,
 .breakpoints-exceptions-caught input {
   padding-inline-start: 2px;
+  margin-top: 0px;
+  margin-bottom: 0px;
   margin-inline-start: 0;
+  margin-inline-end: 2px;
   vertical-align: text-bottom;
 }
 
 .breakpoint-exceptions-label {
-  padding-top: 2px;
+  line-height: 14px;
   padding-inline-start: 2px;
   padding-inline-end: 8px;
   cursor: default;
+  overflow: hidden;
+  text-overflow: ellipsis;
 }
 
 .breakpoints-list .breakpoint,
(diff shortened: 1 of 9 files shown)


jaril:
@fvsch We haven't finished discussion on what the XHR input box should look like([#50](https://github.com/firefox-devtools/ux/issues/50)). For now, this input box is a 22px version of what it looked like before:

screen shot 2019-03-06 at 11 03 57 pm

That's likely a separate PR and doesn't need to be included here. Thoughts?

src/components/SecondaryPanes/Breakpoints/Breakpoints.css
```diff
@@ -20,11 +16,18 @@
   display: flex;
   width: 100%;
   align-items: center;
+  padding-top: 3px;
+  padding-bottom: 3px;
+}
+
+.breakpoints-list .breakpoint-heading:nth-child(2) {
```
fvsch: This `:nth-child(2)` looks a bit too circumstantial, and prone to breakage if we add more elements. Not sure what it does, but perhaps a `:not-first-child` or `:nth-child(n+2)` would work too?

src/components/SecondaryPanes/Breakpoints/Breakpoints.css
```diff
@@ -34,57 +37,75 @@
...
 .breakpoints-exceptions-caught {
-  padding: 0 1em 0.5em 3em;
-  margin-top: -0.25em;
+  padding-bottom: 3px;
+  padding-top: 3px;
+  padding-inline-start: 36px;
 }
 
 html[dir="rtl"] .breakpoints-exceptions-caught {
   padding: 0 3em 0.5em 1em;
```
fvsch: We can probably remove this rtl style, since we use `padding-inline-start` above.

src/components/SecondaryPanes/Breakpoints/Breakpoints.css
```diff
@@ -34,57 +37,75 @@
...
-  padding-top: 2px;
+  line-height: 14px;
   padding-inline-start: 2px;
   padding-inline-end: 8px;
   cursor: default;
+  overflow: hidden;
+  text-overflow: ellipsis;
 }
 
 .breakpoints-list .breakpoint,
```
fvsch: It looks like we have the same selector at line 59, with 2px of top/bottom padding.
We should probably merge those two styles.

src/components/SecondaryPanes/Breakpoints/Breakpoints.css
fvsch: We should probably remove this rule, since we're using `padding-inline-start` above.

src/components/SecondaryPanes/Breakpoints/Breakpoints.css
fvsch: We seem to have the same selector at line 59, with 2px padding-top|bottom defined there. We should merge the two rulesets in one place.

fvsch:
The CSS changes look okay, but I haven't checked out the code to give it a spin yet.

A few things to look out for:

1. Since you're using Flexbox, it's worth checking what happens when the content becomes big. E.g. what happens when the main text label is longer than the sidebar width? Does it push the right-hand content out of view? Do we get a scrollbar? Do we clip the label, or does it wrap? (By default I would err on the side of wrapping labels. Sometimes we need to make sure that label uses `flex-shrink: 1` to make it happen.)

2. The checkboxes can have different sizes depending on the OS. It seems you're removing their default margins, which is good. For the record, I'm testing on Linux and the GTK checkboxes seem to be 16x16 (bigger than the Windows and Mac checkboxes which are in the 12-14px range). We can't accomodate everything, but if it works with 16px checkboxes we should be good.

> [Re. the XHR breakpoint input+select style] That's likely a separate PR and doesn't need to be included here.

Agreed, let's live it like this for now and address it in a different issue and PR.
@derek-li was interested in working on it.

src/components/SecondaryPanes/Breakpoints/Breakpoints.css
jaril: This was a way around having to change any of the Component code to add 4px to the start and end of the actual list of breakpoints. This code snippet:

```
.breakpoints-list .breakpoint-heading:nth-child(2) {
  margin-top: 4px;
}
.breakpoints-list .breakpoint:last-child {
  margin-bottom: 4px;
}
```

- adds a 4px margin-top to the start of the breakpoints list, which starts with a breakpoint heading div
- adds a 4px margin-bottom to the end of the breakpoints list, which ends with a breakpoint div

The problem I ran into was that the `.breakpoints-list` div also includes `.breakpoints-exceptions` as its first child, which prevented me from just adding a padding-top+bottom to `.breakpoints-list.`

An alternative is to move `.breakpoints-exceptions` out of `breakpoints-list` so I can just set a top+bottom padding for `breakpoints-list` to do the same thing. 

```html
// BEFORE
<div class="breakpoints-list">
  <div class="breakpoints-exceptions"> ... </div>
  <div class="breakpoint-heading"> ... </div>
  <div class="breakpoint"> ... </div>
</div>

// AFTER
<div>
  <div class="breakpoints-exceptions"> ... </div>
  <div class="breakpoints-list">
    <div class="breakpoint-heading"> ... </div>
    <div class="breakpoint"> ... </div>
  </div>
</div>
```

jaril:
>Since you're using Flexbox, it's worth checking what happens when the content becomes big.

Right now we clip the content with `overflow: hidden`, since we don't really wrap any content on overflow.

| Collapsed | Expanded | 
| --- | --- |
| screen shot 2019-03-07 at 9 15 02 am | screen shot 2019-03-07 at 9 17 51 am | 

>We can't accomodate everything, but if it works with 16px checkboxes we should be good.

Great! The 20px row heights are set off of either a 16px element (either `breakpoint-line` or a close button) with 2px top+bottom padding so it should keep the 20px regardless of checkbox size. :)

src/components/SecondaryPanes/Breakpoints/Breakpoints.css
jaril: Fixed!

src/components/SecondaryPanes/Breakpoints/Breakpoints.css
jaril: Removed.

src/components/SecondaryPanes/Breakpoints/Breakpoints.css
jaril: Fixed.

src/components/SecondaryPanes/Breakpoints/Breakpoints.css
jaril: Removed.

src/components/SecondaryPanes/Breakpoints/index.js
```diff
@@ -81,41 +81,45 @@ class Breakpoints extends Component<Props> {
...
-              )}
-            />
-          ))
-        ];
-      })
-    ];
+    return (
+      <div className="pane breakpoints-list">
+        {[
+          ...breakpointSources.map(({ source, breakpoints, i }) => {
```
fvsch: That was in the original code, but I'm pretty sure we don't need to do `{[ ...someArray.map(foo) ]}` to render a list in JSX, and can directly write `{someArray.map(foo)}`

src/components/SecondaryPanes/Breakpoints/index.js
```diff
@@ -81,41 +81,45 @@ class Breakpoints extends Component<Props> {
...
+          })
+        ]}
+      </div>
+    );
   }
 
   render() {
     return (
-      <div className="pane breakpoints-list">
+      <div>
```
fvsch: It might be useful to still have a specific class that identifies the content of the "Breakpoints" pane.
It's likely we have some mochitests that select DOM nodes using this class, so we could be breaking that.

Maybe we should add a class to the element we're trying to style, rather than move `breakpoints-list` to a different element.

fvsch: mochitest selectors:
```
  breakpointItem: i => `.breakpoints-list div:nth-of-type(${i})`,
  breakpointItems: ".breakpoints-list .breakpoint",
```
But maybe you already accounted for that.

src/components/SecondaryPanes/Breakpoints/index.js
```diff
@@ -81,41 +81,45 @@ class Breakpoints extends Component<Props> {
...
         {this.renderExceptionsOptions()}
-        {this.renderBreakpoints()}
+        {!!this.props.breakpointSources.length && this.renderBreakpoints()}
```
fvsch: Nitpick: my preferred style would be to have this check in `renderBreakpoints`.

```diff
   renderBreakpoints() {
     const { breakpointSources, selectedSource } = this.props;
+    if (!breakpointSources.length) return null;
```

We could also remove the `empty` conditional class in `renderExceptionsOptions` and move this style to the breakpoints list instead (using margin-top + border-top + padding-top):

```css
.breakpoints-exceptions-options:not(.empty) {
  margin-bottom: 3px;
}
```

src/components/SecondaryPanes/Breakpoints/index.js
jaril: Good catch - thanks!

src/components/SecondaryPanes/Breakpoints/index.js
jaril: Tests broke initially because of that, but I've fixed the relevant tests. The reason for structuring it this way was consistency with XHR Breakpoints:

```html
<!-- XHR Breakpoints -->
<div>
  <div class="breakpoints-exceptions-options"> ... </div>
  <ul class="pane expressions-list"> ... </ul>
</div>

<!-- Breakpoints -->
<div>
  <div class="breakpoints-exceptions"> ... </div>
  <div class="pane breakpoints-list"> ... </div>
</div>
```

src/components/SecondaryPanes/Breakpoints/index.js
jaril: Addressed this 👍

fvsch:
Checked it out and it looks pretty good.

My last comment is that we use two different background colors on hover for code breakpoints and XHR breakpoints: a light gray and a light blue. Could we use the light gray for both? (Last related UX discussion I had, on hover color for accordion headers, we wanted to get rid of the light blue shades.)

src/components/SecondaryPanes/XHRBreakpoints.css
```diff
@@ -3,8 +3,9 @@
  * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
 
 .xhr-input-container {
-  display: block;
+  display: flex;
   border: 1px solid transparent;
+
```
derek-li: Nitpick: extra line here

jaril: Good catch - thanks Derek!

darkwing:
One more look @fvsch?

fvsch:
@darkwing @Jaril Tested again and it looks great.

Once thing I'd like to improve is the look of the method <select> in XHR breakpoints, but we can do that in a follow-up bug. I've opened https://github.com/firefox-devtools/debugger/issues/8141 for that.

PR #8104, opened 2019-03-08
Use CompiledModules instead of DebuggerModules in moz.build files

darkwing:
We moved to `CompiledModules` for the JSX work.  This change will make releases 1 billion times easier.


darkwing:
This will require a bump in travis MC hash before it passes tests.

darkwing:
Closing; I did this as part of sync 129

PR #8112, opened 2019-03-11
Sync 129

jasonLaster:
syncs changes from release 129


darkwing:
@flodolo You can ignore this -- we're syncing over from mozilla-central

PR #8115, opened 2019-03-12
[Log Point] Fix log point context menu accelerator

### LINKED ISSUE #7692: [log points] Add Log and Add Condition share the same context menu accelerator
Both features are mapped to CtrlOrCmd+Shift+B , they are both using `accelerator: L10N.getStr("toggleCondPanel.key"),`

https://github.com/devtools-html/debugger.html/blob/55183cd1dfaca2f33b743bc590dd8bdc652d8bcb/src/components/Editor/GutterMenu.js#L103

image

ravefalcon92:
Fixes #7692

### Summary of Changes

* Created new keyboard shortcut(`CmdOrCtrl+Shift+Y`) for toggling conditional panel for log points
* Updated log points context menu accelerator 
* Updated ShortcutsModal with new keyboard shortcut for log points
* Changed shortcut description for `shortcuts.toggleCondPanel` from `Toggle Conditional Panel` to `Edit Conditional Breakpoint`
* Updated snapshot testing

### DIFF
--- a/assets/panel/debugger.properties
+++ b/assets/panel/debugger.properties
@@ -243,17 +243,17 @@ functionSearch.key=CmdOrCtrl+Shift+O
 # key identifiers, not messages displayed to the user.
 toggleBreakpoint.key=CmdOrCtrl+B
 
-# LOCALIZATION NOTE (toggleCondPanel.key): A key shortcut to toggle
+# LOCALIZATION NOTE (toggleCondPanel.breakpoint.key): A key shortcut to toggle
 # the conditional panel for breakpoints.
 # Do not localize "CmdOrCtrl+Shift+B", or change the format of the string. These are
 # key identifiers, not messages displayed to the user.
-toggleCondPanel.key=CmdOrCtrl+Shift+B
+toggleCondPanel.breakpoint.key=CmdOrCtrl+Shift+B
 
-# LOCALIZATION NOTE (toggleLogPanel.key): A key shortcut to toggle
+# LOCALIZATION NOTE (toggleCondPanel.logPoint.key): A key shortcut to toggle
 # the conditional panel for log points.
 # Do not localize "CmdOrCtrl+Shift+Y", or change the format of the string. These are
 # key identifiers, not messages displayed to the user.
-toggleLogPanel.key=CmdOrCtrl+Shift+Y
+toggleCondPanel.logPoint.key=CmdOrCtrl+Shift+Y
 
 # LOCALIZATION NOTE (stepOut.key): A key shortcut to
 # step out.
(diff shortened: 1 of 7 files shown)


assets/panel/debugger.properties
```diff
@@ -1023,8 +1029,12 @@ shortcuts.toggleBreakpoint=Toggle Breakpoint
 shortcuts.toggleBreakpoint.accesskey=B
 
 # LOCALIZATION NOTE (shortcuts.toggleCondPanel): text describing
-# keyboard shortcut action for toggling conditional panel keyboard
-shortcuts.toggleCondPanel=Toggle Conditional Panel
+# keyboard shortcut action for toggling conditional panel for breakpoints
+shortcuts.toggleCondPanel=Edit Conditional Breakpoint
```
flodolo: Please see this document for making changes to existing strings
https://developer.mozilla.org/en-US/docs/Making_String_Changes

In short: you need a new string ID, and update all the code. I'm also not sure I understand how this fits into the PR subject, but that's more for debugger owners.

src/components/Editor/index.js
```diff
@@ -300,11 +305,16 @@ class Editor extends PureComponent<Props, State> {
...
     if (typeof line !== "number") {
       return;
     }
 
-    this.toggleConditionalPanel(line);
+    if (key === L10N.getStr("toggleLogPanel.key")) {
+      this.toggleConditionalPanel(line, true);
+    } else {
+      this.toggleConditionalPanel(line);
+    }
```
jasonLaster: this is a nit, but...

```js
const isLog = key === L10N.getStr("toggleLogPanel.key");
this.toggleConditionalPanel(line, isLog);
```

assets/panel/debugger.properties
ravefalcon92: Thanks for bringing this to my attention. I'll discuss viable solutions with the debugger owners.

src/components/Editor/index.js
ravefalcon92: Fixed. Thanks for the nit

ravefalcon92:
@darkwing, @jasonLaster: Wanted to get your thoughts on how we'd like to move forward with this PR.

The issue is that we currently only have one key/shortcut description for the conditional panel, but we have a conditional panel for breakpoints as well as log points and want to differentiate between the two.

My proposed solution:
~~`toggleCondPanel.key = 'CmdOrCtrl+Shift+B'`~~
`toggleCondPanel.breakpoint.key =  'CmdOrCtrl+Shift+B'`
`toggleCondPanel.logpoint.key = 'CmdOrCtrl+Shift+Y'`
~~`shorcuts.toggleCondPanel = 'Toggle Conditional Panel'`~~
`shorcuts.toggleCondPanel.breakpoint = 'Edit Conditional Breakpoint'`
`shorcuts.toggleCondPanel.logpoint = 'Edit Log Point'`

2 reasons for why I've changed the shortcut description:
1. The conditional panel is more of a term for people who are working on the debugger so it might make more sense to remove mention of conditional panel and switch to talking about breakpoints/log points.
2. Switched the shortcut description from `Toggle` to `Edit` because we are providing the users with an option to add/edit a condition or log. `Toggle` makes sense when it pertains to the conditional panel, but if we decide to remove mention of conditional panel, I think `Edit` is more clear.

assets/panel/debugger.properties
ravefalcon92: I appended a differentiating property to the end of `toggleCondPanel` between log points (`toggleCondPanel.logPoint`)and breakpoints (`toggleCondPanel.breakpoint`) so there is no change to existing strings. @flodolo these changes seem to be consistent with the localization best practices you linked. Let me know if this works

darkwing:
Awesome, thank you!

PR #8125, opened 2019-03-14
Display Date object using `toString` instead of `toISOString`

### LINKED ISSUE #6153: Print dates in local timezone
_From @nchevobbe on June 5, 2017 12:55_

_Moved from https://github.com/devtools-html/reps/issues/114_

See https://bugzilla.mozilla.org/show_bug.cgi?id=1073533

We are using ISO string when we output a date, which might be confusing depending on your timezone.
It'd be more clear to output the local timezone.

_Copied from original issue: devtools-html/devtools-core#400_

GettyOrawo:
Fixes #6153

### Summary of Changes

* change date from` toISOString` to `toString` format

### Test Plan

I modified the already existing tests to accommodate the new feature since the return values of the date time was changed.

- [x] Tested the patch displays the given date in string format e.g `Date Thu Mar 31 2016 00:17:24 GMT+0300 (East Africa Time)`

- [x] Tested the patch defaults to `Date Invalid Date` when the given date is invalid

- [x] Change test timezone to East African Time.

firefox-datetime

### DIFF
--- a/packages/devtools-reps/src/reps/date-time.js
+++ b/packages/devtools-reps/src/reps/date-time.js
@@ -24,9 +24,10 @@ function DateTime(props) {
   try {
     const dateObject = new Date(grip.preview.timestamp);
     dateObject.toISOString();
-    date = span({
-      "data-link-actor-id": grip.actor,
-      className: "objectBox"
+    date = span(
+      {
+        "data-link-actor-id": grip.actor,
+        className: "objectBox"
       },
       getTitle(grip),
       span({ className: "Date" }, dateObject.toString())
@@ -39,9 +40,10 @@ function DateTime(props) {
 }
 
 function getTitle(grip) {
-  return span({
-    className: "objectTitle"
-  },
+  return span(
+    {
+      className: "objectTitle"
+    },
     `${grip.class} `
   );
 }
--- a/packages/devtools-reps/src/reps/date-time.js
+++ b/packages/devtools-reps/src/reps/date-time.js
@@ -30,7 +30,7 @@ function DateTime(props) {
       getTitle(grip),
       span(
         { className: "Date" },
-        new Date(grip.preview.timestamp).toLocaleDateString()
+        new Date(grip.preview.timestamp).toString()
       )
     );
   } catch (e) {
--- a/packages/devtools-reps/src/reps/date-time.js
+++ b/packages/devtools-reps/src/reps/date-time.js
@@ -28,9 +28,7 @@ function DateTime(props) {
         className: "objectBox"
       },
       getTitle(grip),
-      span(
-        { className: "Date" },
-        new Date(grip.preview.timestamp).toString()
+      span({ className: "Date" }, new Date(grip.preview.timestamp).toString()
       )
     );
   } catch (e) {
--- a/packages/devtools-reps/src/reps/date-time.js
+++ b/packages/devtools-reps/src/reps/date-time.js
@@ -25,8 +25,8 @@ function DateTime(props) {
     const dateObject = new Date(grip.preview.timestamp);
     dateObject.toISOString();
     date = span({
-        "data-link-actor-id": grip.actor,
-        className: "objectBox"
+      "data-link-actor-id": grip.actor,
+      className: "objectBox"
       },
       getTitle(grip),
       span({ className: "Date" }, dateObject.toString())
@@ -40,8 +40,8 @@ function DateTime(props) {
 
 function getTitle(grip) {
   return span({
-      className: "objectTitle"
-    },
+    className: "objectTitle"
+  },
     `${grip.class} `
   );
 }
--- a/packages/devtools-reps/src/reps/date-time.js
+++ b/packages/devtools-reps/src/reps/date-time.js
@@ -30,7 +30,7 @@ function DateTime(props) {
       getTitle(grip),
       span(
         { className: "Date" },
-        new Date(grip.preview.timestamp).toISOString()
+        new Date(grip.preview.timestamp).toLocaleDateString()
       )
     );
   } catch (e) {
--- a/packages/devtools-reps/src/reps/date-time.js
+++ b/packages/devtools-reps/src/reps/date-time.js
@@ -28,8 +28,7 @@ function DateTime(props) {
         className: "objectBox"
       },
       getTitle(grip),
-      span({ className: "Date" }, new Date(grip.preview.timestamp).toString()
-      )
+      span({ className: "Date" }, new Date(grip.preview.timestamp).toString())
     );
   } catch (e) {
     date = span({ className: "objectBox" }, "Invalid Date");
--- a/packages/devtools-reps/src/reps/date-time.js
+++ b/packages/devtools-reps/src/reps/date-time.js
@@ -26,7 +26,7 @@ function DateTime(props) {
     // Calling `toISOString` will throw if the date is invalid,
     // so we can render an `Invalid Date` element.
     dateObject.toISOString();
-    
+
     date = span(
       {
         "data-link-actor-id": grip.actor,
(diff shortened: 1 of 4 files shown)


darkwing:
Awesome, thank you so much for this @GettyOrawo !  I look forward to the test :)

packages/devtools-reps/src/reps/date-time.js
```diff
@@ -30,7 +30,7 @@ function DateTime(props) {
       getTitle(grip),
       span(
         { className: "Date" },
-        new Date(grip.preview.timestamp).toISOString()
+        new Date(grip.preview.timestamp).toLocaleDateString()
```
nchevobbe: I think we can safely use `toString`, which will resolve the issue we have :)

```suggestion
        new Date(grip.preview.timestamp).toString()
```

GettyOrawo: Alright on it.

GettyOrawo: and @nchevobbe I find this a little confusing, now that the tests read my current timezone, will that be a problem if the tests are run from any other time-zones?

GettyOrawo: I actually just realized the pipeline fails since I'm testing using my time zone

nchevobbe:
Hello Getty, you're right being worried for the test (and as you noticed, they does fail here since CI does not have the same timezone).

What we can do to fix this is the following: 

```diff
diff --git a/package.json b/package.json
index afc2de0e2..e5574716c 100644
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "mochir": "yarn mochi -- --repeat 10 --",
     "mochih": "yarn mochi -- --headless --",
     "mochici": "mochii --mc ./firefox --ci --default-test-path devtools/client/debugger/new --headless --",
-    "test": "jest",
+    "test": "TZ=Africa/Nairobi jest",
     "test:watch": "jest --watch",
     "test:coverage": "yarn test --coverage",
     "test:all": "yarn test; yarn lint; yarn flow",
diff --git a/packages/devtools-reps/package.json b/packages/devtools-reps/package.json
index 19e768e08..d7b850fce 100644
--- a/packages/devtools-reps/package.json
+++ b/packages/devtools-reps/package.json
@@ -9,7 +9,7 @@
     "firefox": "./node_modules/.bin/start-firefox --start --location https://firefox-devtools.github.io/debugger-examples/",
     "chrome": "./node_modules/.bin/start-chrome",
     "license-check": "devtools-license-check",
-    "test": "jest --projects jest.config.js"
+    "test": "TZ=Africa/Nairobi jest --projects jest.config.js"
   },
   "author": "",
   "license": "MPL-2.0",
diff --git a/packages/devtools-reps/src/reps/tests/date-time.js b/packages/devtools-reps/src/reps/tests/date-time.js
index 750934c62..c63901ee7 100644
--- a/packages/devtools-reps/src/reps/tests/date-time.js
+++ b/packages/devtools-reps/src/reps/tests/date-time.js
@@ -26,7 +26,9 @@ describe("test DateTime", () => {
       })
     );
 
-    expect(renderedComponent.text()).toEqual("Date Thu Mar 31 2016 00:17:24 GMT+0300 (East Africa Time)");
+    expect(renderedComponent.text()).toEqual(
+      "Date Thu Mar 31 2016 00:17:24 GMT+0300 (East Africa Time)"
+    );
     expectActorAttribute(renderedComponent, stub.actor);
   });
 });
```

With those changes, the test pass locally for me (even if I'm not on East Africa Time), so they should pass on CI too.
Can you try to apply those change to your repo and update your PR? Then we can confirm this is working.

GettyOrawo:
> Hello Getty, you're right being worried for the test (and as you noticed, they does fail here since CI does not have the same timezone).
> 
> What we can do to fix this is the following:
> 
> ```diff
> diff --git a/package.json b/package.json
> index afc2de0e2..e5574716c 100644
> --- a/package.json
> +++ b/package.json
> @@ -29,7 +29,7 @@
>      "mochir": "yarn mochi -- --repeat 10 --",
>      "mochih": "yarn mochi -- --headless --",
>      "mochici": "mochii --mc ./firefox --ci --default-test-path devtools/client/debugger/new --headless --",
> -    "test": "jest",
> +    "test": "TZ=Africa/Nairobi jest",
>      "test:watch": "jest --watch",
>      "test:coverage": "yarn test --coverage",
>      "test:all": "yarn test; yarn lint; yarn flow",
> diff --git a/packages/devtools-reps/package.json b/packages/devtools-reps/package.json
> index 19e768e08..d7b850fce 100644
> --- a/packages/devtools-reps/package.json
> +++ b/packages/devtools-reps/package.json
> @@ -9,7 +9,7 @@
>      "firefox": "./node_modules/.bin/start-firefox --start --location https://firefox-devtools.github.io/debugger-examples/",
>      "chrome": "./node_modules/.bin/start-chrome",
>      "license-check": "devtools-license-check",
> -    "test": "jest --projects jest.config.js"
> +    "test": "TZ=Africa/Nairobi jest --projects jest.config.js"
>    },
>    "author": "",
>    "license": "MPL-2.0",
> diff --git a/packages/devtools-reps/src/reps/tests/date-time.js b/packages/devtools-reps/src/reps/tests/date-time.js
> index 750934c62..c63901ee7 100644
> --- a/packages/devtools-reps/src/reps/tests/date-time.js
> +++ b/packages/devtools-reps/src/reps/tests/date-time.js
> @@ -26,7 +26,9 @@ describe("test DateTime", () => {
>        })
>      );
>  
> -    expect(renderedComponent.text()).toEqual("Date Thu Mar 31 2016 00:17:24 GMT+0300 (East Africa Time)");
> +    expect(renderedComponent.text()).toEqual(
> +      "Date Thu Mar 31 2016 00:17:24 GMT+0300 (East Africa Time)"
> +    );
>      expectActorAttribute(renderedComponent, stub.actor);
>    });
>  });
> ```
> 
> With those changes, the test pass locally for me (even if I'm not on East Africa Time), so they should pass on CI too.
> Can you try to apply those change to your repo and update your PR? Then we can confirm this is working.

Oh yes. Let me fix that in. Thank you!

nchevobbe:
You should be able to get rid of this lint failure by running `yarn lint:js -- --fix` from the project's root

GettyOrawo:
I believe it's working fine now.

packages/devtools-reps/src/reps/tests/date-time.js
```diff
@@ -41,6 +43,6 @@ describe("test invalid DateTime", () => {
       })
     );
 
-    expect(renderedComponent.text()).toEqual("Invalid Date");
```
nchevobbe: I'm not sure why this change is needed?

nchevobbe: so, turns out putting an invalid date in the constructor does not throw, it just produces an invalid date. it was the `.toISOString()` call that was then throwing.
So maybe what we could do is keep the call to `toISOString()` just so it can throws and detect Invalid date.

```js
  try {
    const dateObject = new Date(grip.preview.timestamp);
    // this call throws if we are dealing with an invalid date.
    dateObject.toISOString();
    date = span(
      {
        "data-link-actor-id": grip.actor,
        className: "objectBox"
      },
      getTitle(grip),
      span({ className: "Date" }, dateObject.toString())
    );
  } catch (e) {
    date = span({ className: "objectBox" }, "Invalid Date");
  }
```

nchevobbe:
@darkwing @jasonLaster what do you think of the added timezone env?

darkwing:
It seems reasonable to me.

GettyOrawo:
Oh I think I've got it @nchevobbe . Is this okay now?

nchevobbe:
CI is ✅✅✅ , let's merge the PR :)

PR #8126, opened 2019-03-14
Ux 41 on hover classes and functions variations outline panel

yogmel:
Fixes [UX#41](https://github.com/firefox-devtools/ux/issues/41)

### Summary of Changes

* Class title: changed from margin to padding, to match the spacing of the function;
* Class title: insert margin-top only if it's the first element on the list;
* Functions outside Class: they keep the 1st level spacing;
* Changed div to li when creating .outline-list__class elements.

Example:
54314928-aae75900-45bb-11e9-8272-402df01d0cc6

### Test Plan

Opened this file on the Debugger, at the Outline tab:
```javascript
// example script for testing
function eee() {}
class bbb {
  constructor() {
    this.test = 'bbb';
  }
}

class Example {
  constructor(name){
    this.name = name;
  }

  testMe() {
    return `Hello!`;
  }
}

class Example2 {
  constructor(name){
    this.name = name;
  }

  testMeAgain() {
    return `Hello!`;
  }
}
```

Check if the hover state of the Class and functions are spaced correctly.

### DIFF
--- a/src/components/PrimaryPanes/Outline.css
+++ b/src/components/PrimaryPanes/Outline.css
@@ -2,7 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
 
- .outline {
+.outline {
   overflow-y: hidden;
 }
 
--- a/src/components/PrimaryPanes/Outline.css
+++ b/src/components/PrimaryPanes/Outline.css
@@ -2,7 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
 
-.outline {
+ .outline {
   overflow-y: hidden;
 }
 
@@ -12,31 +12,32 @@
 }
 
 .outline-pane-info {
-  padding: 0.5em;
   width: 100%;
   font-style: italic;
   text-align: center;
+  padding: 0.5em;
   user-select: none;
   font-size: 12px;
   overflow: hidden;
 }
 
 .outline-list {
-  margin: 0;
-  padding: 0 0 4px 0;
-  position: absolute;
-  top: 38px;
-  bottom: 25px;
-  left: 0;
-  right: 0;
   list-style-type: none;
+  padding: 4px 0;
+  margin: 0;
+  font-family: var(--monospace-font-family);
   overflow: auto;
+  position: absolute;
+  top: 0;
+  right: 0;
+  left: 0;
+  bottom: 25px;
 }
 
 .outline-list__class-list {
+  list-style: none;
   margin: 0;
   padding: 0;
-  list-style: none;
 }
 
 .outline-list__class-list .function-signature .function-name {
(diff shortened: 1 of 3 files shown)


src/components/PrimaryPanes/Outline.css
```diff
@@ -47,11 +47,16 @@
   color: inherit;
 }
 
-.outline-list h2 {
-  margin: 10px 0 10px 10px;
+.outline-list .outline-list__class h2 {
   font-weight: normal;
   font-size: 1em;
+  padding: 3px 0 3px 10px;
```
fvsch: Can we use text direction dependant padding? Something like:

```
padding: 3px 0;
padding-inline-start: 10px;
```

src/components/PrimaryPanes/Outline.css
```diff
@@ -72,6 +77,10 @@
   white-space: nowrap;
 }
 
+.outline-list > .outline-list__element {
+  padding-left: 0;
```
fvsch: So, instead of removing the list item's padding, and letting the icon's padding do the actual spacing, I would remove the padding-left on the icon, then add more padding to `.outline-list__element` when it's in a sub-list.

src/components/PrimaryPanes/Outline.css
```diff
@@ -72,6 +77,10 @@
...
+}
+
 .outline-list__element-icon {
   padding-right: 0.4rem;
```
fvsch: So here it would be interesting to:
- use `padding-inline-start` and `padding-inline-end` instead of the padding-left/right properties;
- maybe only keep the space between the icon and the label (it makes more sense to manage the indentation with the list item's padding, IMO).

src/components/PrimaryPanes/Outline.css
yogmel: Ok, changes made:

Outline.css:43 - added padding when element is on a class list
Outline.css:90 -- remove padding-left from icon

src/components/PrimaryPanes/Outline.css
```diff
@@ -39,6 +39,10 @@
   list-style: none;
 }
 
+.outline-list__class-list > .outline-list__element {
+  padding-left: 2rem;
```
fvsch: Probably `padding-inline-start` here to be consistent with the style for `.outline-list > .outline-list__element`.

fvsch:
Since you changed a `<div>` to a `<li>`, some tests are failing because we have tests that generate a snapshot of the HTML generated by a component with different inputs, and those snapshots still expect a `<div>`.

You can see the failing test by running:

```sh
yarn test Outline.spec.js
```

That should run only this single test file. (You can run all Jest tests with `yarn test`.)

If you think the HTML generated with your changes is okay, then we must update the snapshots, which you can do with:

```sh
yarn test Outline.spec.js -u
```

Then you need to commit the updated `.snap` file.

yogmel:
ok, @fvsch , thanks for reviewing! The snapshot looks fine.

src/components/PrimaryPanes/Outline.css
darkwing: Yep, let's move to `padding-inline-start`

darkwing:
@yogmel Could you move this small patch over to Phabricator?  That would be a huge help!  Thank you!

yogmel:
Yes, sure. :) 
Do I create a bug there or is it already created?

darkwing:
Please create it :) . Cheers and thank you for the extra effort!

PR #8133, opened 2019-03-16
Pretty Printing done in loadSourceText .

### LINKED ISSUE #8071: loadSourceText should probably be responsible for pretty-printing files
Right now pretty-printing is its own action that kind of hacks its way into the source-loading process. This isn't idea because it means that the promise returned by `loadSourceText` can't consistently handle the case where a file is in the process of being pretty-printed, since don't have a way to know when pretty-printing completes.

nagpalm7:
Fixes #8071 

### Summary of Changes

* Create prettyPrintSources function which returns source with and sets desirable mappings.

### DIFF
--- a/src/actions/sources/prettyPrint.js
+++ b/src/actions/sources/prettyPrint.js
@@ -10,7 +10,6 @@ import { remapBreakpoints, setBreakpointPositions } from "../breakpoints";
 
 import { setSymbols } from "../ast";
 import { prettyPrint } from "../../workers/pretty-print";
-import { setSource } from "../../workers/parser";
 import { getPrettySourceURL, isLoaded } from "../../utils/source";
 import { loadSourceText } from "./loadSourceText";
 import { mapFrames } from "../pause";
@@ -25,11 +24,11 @@ import {
 
 import type { Action, ThunkArgs } from "../types";
 import { selectSource } from "./select";
-import type { JsSource } from "../../types";
+import type { JsSource, Source } from "../../types";
 
 export async function prettyPrintSource(
-  sourceMaps,
-  prettySource,
+  sourceMaps: any,
+  prettySource: Source,
   generatedSource: Source
 ) {
   const url = getPrettySourceURL(generatedSource.url);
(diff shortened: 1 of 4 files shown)


src/actions/sources/prettyPrint.js
```diff
@@ -50,18 +74,11 @@ export function createPrettySource(sourceId: string) {
...
-    for (const sourceActor of source.actors) {
-      await sourceMaps.applySourceMap(sourceActor.actor, url, code, mappings);
-    }
+    const { text } = await prettyPrintSource(sourceMaps, prettySource, source);
 
     const loadedPrettySource: JsSource = {
       ...prettySource,
-      text: code,
+      text: text,
       loadedState: "loaded"
```
loganfsmyth: One of the motivations behind the issue for this is that `loadSourceText` already does this `loaded` toggling as part of the `LOAD_SOURCE_TEXT` action for normal sources. I'm hoping that by doing this in loadSourceText that we can remove this part from here entirely.

nagpalm7: Okay @loganfsmyth, I want to know whether we want to prettyprint all files on their initial loading and remove the dependence on clicking toggle button?

src/actions/tests/__snapshots__/project-text-search.spec.js.snap
```diff
@@ -82,26 +82,7 @@ Array [
...
-        "type": "MATCH",
-        "value": "function bla(x, y) {",
-      },
-    ],
-    "sourceId": "bar:formatted",
-    "type": "RESULT",
-  },
-]
-`;
+exports[`project text search should ignore sources with minified versions 1`] = `Array []`;
```
jasonLaster: hmm, i wonder why this changed?

nagpalm7: @jasonLaster No idea, `yarn jest -u` did this work.

jasonLaster: I think it is worth looking into whether this is desired :)

jasonLaster:
Thanks @nagpalm7! 

This will be landing pretty soon - https://phabricator.services.mozilla.com/D23917?vs=77402&id=77949#toc

PR #8134, opened 2019-03-16
Keep QuickOpenModal resultCount in bounds

### LINKED ISSUE #8119: "Go to file" selected index can be out of bounds when showing limiting the number of results
It looks like the "Go to file" result list is clipped at maybe 100 items, but when the actual result count is higher then users can use the keyboard to lose the currently selected index.
My guess is that we store an index for the currently selected item, and compare it to the result count instead of comparing it to the number of rendered items.

STR:

1. Go to https://firefox-dev.tools/debugger-examples/examples/500-sources.html
2. Hit Ctrl+P to show the "Go to file" input
3. Type "js" to show many results
4. Hit the Up key 5 times.
5. Hit the Down key 5 times.

Expected results:
- In step 3, we should get a list of 100 results displayed.
- In step 4, the first "Up" action should cycle back to the last item (index `99`, if it's a numeric 0-indexed value), then the next "Up" actions should go up in the list (up to index `95`).
- In step 5, the "Down" actions should go down in the list, and the fifth "Down" should cycle back to the first item in the list.

Actual results:
- In step 4, the first "Up" action removes any visible selection.
- In step 5, the fifth "Down" action restores the selected state to the first item.

arosenfeld2003:
Fixes #8119 

Here's the Pull Request Doc
https://firefox-dev.tools/debugger/docs/pull-requests.html

### Summary of Changes

Added a conditional to keep resultCount (number returned from getResultCount) at 100 or less.  
This is contained in the traverseResults method, within QuickOpenModal.js. 

### Test Plan

Tested behavior based upon expected behavior described by @fvsch.

Example test plan:

- [x] There are 100 items displayed in Go To File input.
- [x] Clicking "Up" arrow key from index 0 cycles back to the last item (item 99).
- [x] Clicking the "Down" arrow from index 99 cycles up to the first list item (index 0).
- [x] Passed all unit tests locally.

issue 8119

This is my first code PR, so greatly appreciate any and all feedback.  
Sorry for the extraneous commits (issues with git).
Thanks!

### DIFF
--- a/src/components/QuickOpenModal.js
+++ b/src/components/QuickOpenModal.js
@@ -69,10 +69,12 @@ type GotoLocationType = {
   column?: number
 };
 
+const maxResults = 100;
+
 function filter(values, query) {
   return fuzzyAldrin.filter(values, query, {
     key: "value",
-    maxResults: 1000
+    maxResults: maxResults
   });
 }
 
@@ -80,6 +82,13 @@ export class QuickOpenModal extends Component<Props, State> {
   constructor(props: Props) {
     super(props);
     this.state = { results: null, selectedIndex: 0 };
+  }
+
+  setResults(results: ?Array<QuickOpenResult>) {
+    if (results) {
+      results = results.slice(0, maxResults);
+    }
+    this.setState({ results });
   }
 
   componentDidMount() {
@@ -121,7 +130,7 @@ export class QuickOpenModal extends Component<Props, State> {
     const { sources } = this.props;
     const results =
       query == "" ? sources : filter(sources, this.dropGoto(query));
-    return this.setState({ results });
+    return this.setResults(results);
   };
 
   searchSymbols = (query: string) => {
@@ -133,18 +142,18 @@ export class QuickOpenModal extends Component<Props, State> {
     results = results.filter(result => result.title !== "anonymous");
 
     if (query === "@" || query === "#") {
-      return this.setState({ results });
+      return this.setResults(results);
     }
-
-    this.setState({ results: filter(results, query.slice(1)) });
+    results = filter(results, query.slice(1));
+    return this.setResults(results);
   };
 
   searchShortcuts = (query: string) => {
     const results = formatShortcutResults();
     if (query == "?") {
-      this.setState({ results });
+      this.setResults(results);
     } else {
-      this.setState({ results: filter(results, query.slice(1)) });
+      this.setResults(filter(results, query.slice(1)));
     }
   };
 
@@ -152,12 +161,9 @@ export class QuickOpenModal extends Component<Props, State> {
     const { tabs, sources } = this.props;
     if (tabs.length > 0) {
       const tabUrls = tabs.map((tab: Tab) => tab.url);
-
-      this.setState({
-        results: sources.filter(source => tabUrls.includes(source.url))
-      });
+      this.setResults(sources.filter(source => tabUrls.includes(source.url)));
     } else {
-      this.setState({ results: sources.slice(0, 100) });
+      this.setResults(sources);
     }
   };
 
@@ -374,8 +380,7 @@ export class QuickOpenModal extends Component<Props, State> {
     if (!enabled) {
       return null;
     }
-    const newResults = results && results.slice(0, 100);
-    const items = this.highlightMatching(query, newResults || []);
+    const items = this.highlightMatching(query, results || []);
     const expanded = !!items && items.length > 0;
 
     return (
@@ -398,7 +403,7 @@ export class QuickOpenModal extends Component<Props, State> {
           }
           {...(this.isSourceSearch() ? { size: "big" } : {})}
         />
-        {newResults && (
+        {results && (
           <ResultList
             key="results"
             items={items}
--- a/src/components/test/QuickOpenModal.spec.js
+++ b/src/components/test/QuickOpenModal.spec.js
@@ -209,7 +209,7 @@ describe("QuickOpenModal", () => {
     wrapper.find("input").simulate("change", { target: { value: "somefil" } });
     expect(filter).toHaveBeenCalledWith([], "somefil", {
       key: "value",
-      maxResults: 1000
+      maxResults: 100
     });
   });
 
@@ -230,7 +230,7 @@ describe("QuickOpenModal", () => {
       .simulate("change", { target: { value: "somefil:33" } });
     expect(filter).toHaveBeenCalledWith([], "somefil", {
       key: "value",
-      maxResults: 1000
+      maxResults: 100
     });
   });
 
@@ -257,7 +257,7 @@ describe("QuickOpenModal", () => {
 
       expect(filter).toHaveBeenCalledWith([], "someFunc", {
         key: "value",
-        maxResults: 1000
+        maxResults: 100
       });
     });
 


src/components/QuickOpenModal.js
```diff
@@ -233,10 +233,10 @@ export class QuickOpenModal extends Component<Props, State> {
   traverseResults = (e: SyntheticKeyboardEvent<HTMLElement>) => {
     const direction = e.key === "ArrowUp" ? -1 : 1;
     const { selectedIndex, results } = this.state;
-    const resultCount = this.getResultCount();
+    const resultCount =
+      this.getResultCount() > 100 ? 100 : this.getResultCount();
```
jasonLaster: this looks like we might need to improve Result Count. 

Also, we might want `maxResults` as a variable

fvsch:
So I think the fix to use depends on if we want to use a result count of more than the limit (100) to do some computations and/or show in the interface. For instance, do we want to show "172 results" somewhere in the UI, even though we only show and give access to the first 100 results?

## Option 1: slice results in state

If we're okay with clipping to 100 for all intents and purposes, the simpler fix would be to clip the results before we add them to the component's state, here:

https://github.com/firefox-devtools/debugger/blob/94b063e0b37d6f0c939d97ef7482129712332e86/src/components/QuickOpenModal.js#L124

Basically changing `this.setState({ results });` to:

```js
this.setState({ results: results.slice(0, maxResults) });
```

And removing the corresponding code in https://github.com/firefox-devtools/debugger/blob/94b063e0b37d6f0c939d97ef7482129712332e86/src/components/QuickOpenModal.js#L377 (then we don't need newResults, we can just pass this.state.results directly since it's already sliced).

## Option 2: keep full results in state

Now if we want to keep the full results array in the component's state but make calculations against its length or the limit, whichever is smallest, then the technique you used is alright.

In that scenario, we definitely need a single `const maxResults = 100;` somewhere (near the start of the file, probably), and we need to replace the hardcoded `100` limit in the `render()` method with that constant too.

Nitpick: you can probably simplify your condition a bit by using Math.min:

```js
const resultCount = Math.min(this.getResultCount(), maxResults);
```

arosenfeld2003:
Thank you @jasonLaster and @fvsch.  I had the same question about whether the full results were needed elsewhere in the UI.  Your suggestions make a lot of sense.  I'm happy to go either direction depending on what's needed. 

As an additional option, my first instinct was a simple change of maxResults (set in the filter function, line 75) from 1000 to 100. 

```
  return fuzzyAldrin.filter(values, query, {
    key: "value",
    maxResults: 100
  });
}
```

This solved the problem, and I thought was much cleaner than my initial solution, but it failed three tests that rely on maxResults: 1000 ...

QuickOpenModal.spec.js - lines 212, 233, 260.

If only 100 results are ever needed, the filter could be set at 100, the tests run with maxResults: 100, and all other behavior/tests work as expected.

arosenfeld2003:
@fvsch and @jasonLaster great suggestions!  I've got three workable solutions based on your advice.
 
**Keep Full Results In State**
The simplest change keeps full results in state, updating traverseResults() as follows:
```
   traverseResults = (e: SyntheticKeyboardEvent<HTMLElement>) => {
     const maxResults = 100;
     const direction = e.key === "ArrowUp" ? -1 : 1;
     const { selectedIndex, results } = this.state;
     const resultCount = Math.min(this.getResultCount(), maxResults);
```

**Slice Results In State**
If we slice results in state based on your suggestion above,  the searchSources and render() methods are updated.  Applying `slice()` throws an error when `results` is undefined, which fails a series of tests.  I added a conditional to return `results` if undefined without applying `slice`, which passed all tests.

searchSources: (line 120)
```     
  const { sources } = this.props;
  const results =
    query == "" ? sources : filter(sources, this.dropGoto(query));
     if (results) {
        return this.setState({ results: results.slice(0, 100) });  
     }
   };
```
render: (line 378)
```    
const items = this.highlightMatching(query, results || []);
const expanded = !!items && items.length > 0;
```
render: (line 401)
```
       {results && (
           <ResultList
             key="results"
             items={items}
```

**Limit Results In Filter**
Another solution that works is setting `maxResults: 100` in the filter method, with the same changes to searchSources() and render() as above.  This also required  setting `maxResults: 100` in the three tests based on that value.

filter function: (line 72)
```
   return fuzzyAldrin.filter(values, query, {
     key: "value",
     maxResults: 100
   });
 }
```

I feel like if we want to clip all results, changing `maxResults` in the filter makes more sense?  But I'm happy to go any direction based on what you think is needed.

Thanks for all your help!

fvsch:
Given that it's a "quick open" feature, not "what's the exact count of sources that match this query" feature, I believe we should just limit the number of results we deal with. Can you try the "Limit Results In Filter" approach?

WenInCode:
I think I agree with @fvsch on that we should just limit the results returned from the `filter` function. 
I don't think that will cover _all_ the cases though. For instance, if I open a file with more than 100 symbols and do an empty symbol query `@` it will just return all the symbols before running `filter`. 

```js
searchSymbols = (query: string) => {
    const {
      symbols: { functions }
    } = this.props;

    let results = functions;
    results = results.filter(result => result.title !== "anonymous");

    if (query === "@" || query === "#") {
      return this.setState({ results });  <--- this could be more than 100 results
    }

    this.setState({ results: filter(results, query.slice(1)) });  <--- If the query is not empty it will use filter
  };
```

I'll include a reproduction of that. 
STR:
- go to https://lodash.glitch.me/
- pretty print and inspect `https://code.jquery.com/jquery-2.2.1.min.js` in the devtools
- open the QuickOpenModal and enter `@` to get all the symbols in the file. 
- tap the up arrow key a few times

symbolsearch

Great work so far @arosenfeld2003 - you are asking great questions. I like the filter approach, but I think there are just a few cases that doesn't cover that we should watch out for.

arosenfeld2003:
@WenInCode, great catch.  The edge case above can be handled by slicing results in the searchSymbols method:

<pre>
searchSymbols = (query: string) => {
    const {
      symbols: { functions }
    } = this.props;

    let results = functions;
    results = results
      .filter(result => result.title !== "anonymous")
      <b>.slice(0, 100);</b>
    if (query === "@" || query === "#") {
    }

    this.setState({ results: filter(results, query.slice(1)) });
  };
</pre>

This seems straightforward, and passes all jest tests locally.

I found four methods in QuickOpenModal where `results` is passed to setState: **searchSources()**, **searchSymbols()**, **searchShortcuts()**, and **showTopSources()**.  I didn't come up with any other edge cases in these methods where we'd get more than 100 sources in the 'go to file' box, but please let me know if can think of another instance that I missed, or another reason we'd get more sources??

Otherwise, I'm happy to amend my PR with the above changes if you think they are agreeable?

Thank you again for all your help @WenInCode , @fvsch, @jasonLaster.

WenInCode:
@arosenfeld2003 seems good to me 👌

fvsch:
Sounds like a good approach.

If we're doing `this.setState({ results: results.slice(0, 100) })` in many different functions it might be worthwhile to move that into a function too.

arosenfeld2003:
Good suggestion @fvsch.  

I think it's ok for now, since we don't actually have to call 
`this.setState({ results: results.slice(0, 100) })`.

To catch the edge case above with an empty symbol query, we can slice before `setState`.
Otherwise if the query is not empty, the `filter` will limit `results` to 100.
<pre>
searchSymbols = (query: string) => {
    const {
      symbols: { functions }
    } = this.props;

    let results = functions;
    results = results
      .filter(result => result.title !== "anonymous")
      <b>.slice</b>(0, 100);

    if (query === "@" || query === "#") {
      return <b>this.setState({ results })</b>;  
    }

    this.setState({ results: <b>filter</b>(results, query.slice(1)) });  
  };
</pre>


In showTopSources, we set `results` by slicing the `sources`:
<pre>
this.setState({ results: <b>sources</b>.slice(0, 100) })
</pre>

All the help is much appreciated!

arosenfeld2003:
I'm hoping this is now good to go?  Please let me know if anything else is needed.  Thanks again.

fvsch:
Two things:

1. You're hardcoding the value `100` in a few different places, which has a potential for creating bugs if we ever change one place and not the others. As was suggested before, could you put it in a  `const` declared near the start of the file (probably [just before this](https://github.com/firefox-devtools/debugger/blob/50473140749eca59f828a9514eff0f00712b9177/src/components/QuickOpenModal.js#L72))?

2. Looking at QuickOpenModal, there are 7 places where we call `this.setState({ results: someValue })`. Making sure that all 7 calls are working with a `someValue` clipped at 100 items maximum does not seem safe to me. If we missed something or change one of those calls later on, it could lead to this bug happening again in specific conditions. Perhaps we should do all `results` state update through a single class method, for example:

```js
  setResults = (results) => {
    if (Array.isArray(results) && results.length > maxResults) {
      results = results.slice(0, maxResults);
    }
    this.setState({ results })
  }
```

arosenfeld2003:
Thanks so much for the detailed suggestions @fvsch!  I now understand what you mean.

I think I've got a good solution that creates/calls a `setResults` method and a `const maxResults = 100` just above the filter.  It passes all jest tests locally.

There is one Flow(InferError) `missing type annotation` popping up, as seen here:


Screen Shot 2019-03-23 at 3 39 15 PM



The code still works as expected.  After some research I'm also not sure why this is showing up.  I'm pushing the updated code to show the diff.  Obviously I'm very happy to continue to make improvements!

fvsch:
Are you familiar with Flow type annotations? You can [read up on them here](https://flow.org/en/docs/types/).

I think here you need to tell Flow that `setResults`'s `results` parameter should be an array of objects whose type is `QuickOpenResult`.

src/components/QuickOpenModal.js
```diff
@@ -130,34 +139,30 @@ export class QuickOpenModal extends Component<Props, State> {
     } = this.props;
 
     let results = functions;
-    results = results.filter(result => result.title !== "anonymous");
```
fvsch: What is the reason for removing this filtering operation from the general case and moving it into the `if (query === "@" || query === "#")` case?

arosenfeld2003: @fvsch thanks so much for the question and thorough review.  Honestly this is my mistake.  I was experimenting with this code block to gain deeper understanding, and didn't intend to move the filtering operation for my commit.

darkwing:
Wow, this looks great @arosenfeld2003 !  Just a quick rebase and it'll be good to merge!

arosenfeld2003:
Thank you all so much @fvsch @WenInCode @jasonLaster @darkwing ... did a rebase, hopefully it's good to go.

darkwing:
Hello @arosenfeld2003!  I fixed our CI (Travis) so one more quick rebase and this will be super green! :)

arosenfeld2003:
Second rebase was a little messy, but I think good to go now... sorry for any confusion!

arosenfeld2003:
@darkwing just checking if I need to do anything further here?  Thanks again for your help!

darkwing:
Excellent, thank you @arosenfeld2003 !

PR #8136, opened 2019-03-16
8118/fix quick open modal nan bug

### LINKED ISSUE #8118: "Go to file" loses the ability to select results if hitting up or down key when results.length==0
STR:

1. Go to https://firefox-dev.tools/debugger-examples/examples/todomvc/ and open Debugger
2. Use the Ctrl+P shortcut to show the "Go to file" input
3. Type a search string that matches nothing, e.g. "fjklsfljkks"
4. Hit the Up arrow key once
5. Remove the non-matching search string and type "back"
6. Hit the Down arrow key once

Expected result:
After step 5, two results show up, with the first result (`backbone.js`) highlighted in blue.
In step 6, hitting the Down arrow should put the focus on the second result (`backbone.localStorage.js`).

Actual result:
After step 5, two results show up, and no result is highlighted.
In step 6, no result is highlighted.

WenInCode:
Fixes #8118 

### Summary of Changes
We setting the `selectedIndex` to `NaN` when you used the keys to navigate an empty result list in the `QuickOpenModal`. This would cause keyboard navigation to break going forward, because `NaN + any number = NaN`. 

- [x] Added failing jest test. Realized after we have a test for this already so, just updated that one and removed my new test.
- [x] Set selected index to 0 when results are empty

### Test Plan
- [x] Updated test for this

Manual test:
1. Open https://firefox-dev.tools/debugger-examples/examples/todomvc/
2. `Command + p`
3. type anything that gets no results in the list
4. use the arrow keys to navigate empty list of results
5. search for a file (ex. app)
6. use arrow keys to navigate results

quickopendialogkeystrokes

### DIFF
--- a/src/components/QuickOpenModal.js
+++ b/src/components/QuickOpenModal.js
@@ -235,8 +235,7 @@ export class QuickOpenModal extends Component<Props, State> {
     const { selectedIndex, results } = this.state;
     const resultCount = this.getResultCount();
     const index = selectedIndex + direction;
-    const nextIndex =
-      resultCount > 0 ? (index + resultCount) % resultCount : resultCount;
+    const nextIndex = (index + resultCount) % resultCount || 0;
 
     this.setState({ selectedIndex: nextIndex });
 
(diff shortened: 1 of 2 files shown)


fvsch:
Works great in my tests.

Tiny nitpick: do we want to be more explicit that we default to `0` when there are no results?
As in: `const nextIndex = resultCount > 0 ? (index + resultCount) % resultCount : 0;`

WenInCode:
@fvsch I agree with you. I actually thought I did it that way haha. Making the change.

PR #8137, opened 2019-03-17
[QuickOpen] show the end of the path

### LINKED ISSUE #8124: [QuickOpen] we should show the end of the path
I've found it can be hard to see the relevant paths the way we're truncating

Screen Shot 2019-03-13 at 5 31 26 PM

th7nder:
Fixes #8124 
Implemented a version with showing only the end of the path.
I am not sure if we need ellipsis in the middle, because in most cases we care about last directories.

### Summary of Changes

* removed truncation in JS 
* added an ellipsis on the start of the line

### DIFF
--- a/src/components/shared/ResultList.css
+++ b/src/components/shared/ResultList.css
@@ -113,10 +113,12 @@
   word-break: break-all;
   /** https://searchfox.org/mozilla-central/source/devtools/client/themes/variables.css **/
   color: var(--grey-40);
-  margin-left: 15px;
+  margin-left: 5px;
   text-overflow: ellipsis;
   overflow: hidden;
   white-space: nowrap;
+  direction: rtl; /* make ellipsis on the beginning of a string */
+  text-align: left;
 }
 
 .theme-dark .result-list.big li.selected .subtitle {
--- a/src/components/shared/ResultList.css
+++ b/src/components/shared/ResultList.css
@@ -113,7 +113,7 @@
   word-break: break-all;
   /** https://searchfox.org/mozilla-central/source/devtools/client/themes/variables.css **/
   color: var(--grey-40);
-  margin-left: 5px;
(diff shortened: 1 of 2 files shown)


src/components/shared/ResultList.css
```diff
@@ -113,10 +113,12 @@
   word-break: break-all;
   /** https://searchfox.org/mozilla-central/source/devtools/client/themes/variables.css **/
   color: var(--grey-40);
-  margin-left: 15px;
+  margin-left: 5px;
```
WenInCode: Just curious what your motivation is for changing this margin? Is it just adding more space for the url? I'm not sure if we have a guideline for spacing in this content. I think the larger margin makes it a bit more readable, but I'd love to hear your thoughts on it 😄

th7nder: It felt like there is too much space when I added an ellipsis at the beginning, but now I am not so sure :D good point!
image

darkwing: I think we should keep left and right spacing consistent; IMO this change makes it too close to the right edge.

src/components/shared/ResultList.css
```diff
@@ -113,10 +113,12 @@
...
   text-overflow: ellipsis;
   overflow: hidden;
   white-space: nowrap;
+  direction: rtl; /* make ellipsis on the beginning of a string */
```
darkwing: This is a solid trick but I'm wondering if @yzen could weigh in on accessibility with it.

src/components/shared/ResultList.css
th7nder: Sure, I'll change it back.

src/components/shared/ResultList.css
yzen: It seems to have no negative effect on screen readers. I think this should be fine.

PR #8138, opened 2019-03-17
[Code Health] Remove inline-svg

### LINKED ISSUE #8121: Remove react-inlinesvg
Looks like we only have two inline SVG icons: one for breakpoints in the left gutter, and one for column breakpoints.

Since we use `react-inlinesvg` for that, and that the way we use it seems very limited (we don't do HTTP requests for external SVG icons with cross-domain IE polyfilling or that kind of thing), the feature we are using boils down to this bit of code:

```js
render() {
   return wrapper({
      style,
      className: this.getClassName(),
      dangerouslySetInnerHTML: html,
    }, content);
}
```

We could probably remove that dependency, and replace Svg.js with a more specific BreakpointSvg.js which hardcodes the two relevant SVG files in strings:

```jsx
const LINE_SVG = `<svg ...></svg>`;
const COLUMN_SVG = `<svg ...></svg>`;

...

render() {
  return <span dangerouslySetInnerHTML={this.props.column ? COLUMN_SVG : LINE_SVG} />
}
```

th7nder:
Fixes #8121 

I haven't implemented `onClick`, `aria-label` and `className`, because they seem to be unused.

### Summary of Changes

* removed Svg.js
* added BreakpointSvg.js
* fixed dependencies in shared components
* removed inline-svg from packages

### Test Plan
- [x] set any breakpoint
- [x] set any column breakpoint


src/vendors.js
```diff
@@ -55,5 +55,5 @@ export const vendored = {
   reselect,
   // Svg is required via relative paths, so the key is not imported path.
   // See .babel/transform-mc.js
-  Svg
+  BreakpointSvg
```
th7nder: I don't know if I did it correctly. I recursively checked devtools-core package and saw that no one uses it. devtools-core uses it's own version of `Svg.js`

darkwing:
Thank you for submitting this @th7nder !  I'll have a look today!

darkwing:
Thank you so much @th7nder !  This is excellent!

PR #8139, opened 2019-03-18
Add stylelint-selector-bem-pattern so that components can opt-in on BEM-like rules

fvsch:
Fixes parts of #7905.

Following discussion about global CSS styles in that issue, here's a test implementation of stylelint-selector-bem-pattern, and a few example usages in "weak" mode on a few components.

It seems to work alright. I found two downsides:

1. The dependencies are a bit on the heavy side (374 packages, like +37MB in node_modules).
2. Even in weak mode, it chokes on selectors like `.foo .my-component` when the defined component name is `my-component`. This is mostly an issue for `.theme-(light|dark)` styles. Using `:root.theme-(light|dark)` as a prefix "solves" it, but in practice it disables the rule completely for that selector.

I'm kinda tempted to create a smaller and simpler Stylelint plugin for our needs.

### DIFF
--- a/.stylelintrc
+++ b/.stylelintrc
@@ -1,5 +1,8 @@
 {
-  "plugins": ["stylelint-prettier"],
+  "plugins": [
+    "stylelint-prettier",
+    "stylelint-selector-bem-pattern"
+  ],
   "rules": {
     "block-no-empty": true,
     "color-hex-length": "long",
@@ -11,6 +14,10 @@
     "function-whitespace-after": "always",
     "no-invalid-double-slash-comments": true,
     "no-missing-end-of-source-newline": true,
+    "plugin/selector-bem-pattern": {
+      "componentName": "^[a-zA-Z-]+$",
+      "componentSelectors": "^\\.{componentName}(?:[-_]+[\\w-]+)?(?:[\\.:][\\w-]+|\\[\\S+\\])*$"
+    },
     "prettier/prettier": true,
     "selector-pseudo-element-colon-notation": "double",
     "selector-type-case": "lower",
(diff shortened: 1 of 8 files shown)


jasonLaster:
Up to you, this seems to work pretty well. 

I wonder if there are other linters that are popular for stylelint...

fvsch:
I looked at other stylelint plugins or rules and nothing else seems to match our needs.

Let's go with this solution for now, and I'll try to write a simpler plugin that we can use to replace this one later on.

darkwing:
This looks like a nice solution but seems like it will be a massive undertaking.  Is there any type of validation that occurs with this?  What's a good plan moving forward?

fvsch:
@darkwing The plan as discussed in #7905.

- Against: adding tooling which transforms CSS selectors (like CSS Modules)
- For: trying to use BEM or similar manually-scoped naming conventions for new components / updates to componets
- For: fixing potentially dangerous global CSS as listed in https://github.com/firefox-devtools/debugger/issues/7905#issuecomment-468392022 (hereafter "list of issues")

> This looks like a nice solution but seems like it will be a  massive undertaking.

My strategy would be to work on some of the components from the "list of issues" progressively, and when we do see if we can opt-in to using `/* @define component-name; weak */`.

Debugger doesn't have that many CSS components anyway. I think we could get to 50% coverage reasonably quick.

darkwing:
@fvsch Would you be up for moving this over to MC so we could continue the conversation there?

fvsch:
@darkwing Do you know how and when Eslint and stylelint are run when working on the Debugger on m-c? And then on the Try servers? You can perhaps show me next week. :)

darkwing:
Of course!  Let's sit down next week and I can help you get it all set up! :)

janodvarko:
The related issue has been cloned into Bugzilla: https://bugzilla.mozilla.org/show_bug.cgi?id=1695623

@Florens the patch should be transferred into Phabricator in order to land it.

Honza

PR #8147, opened 2019-03-20
[releases] Include css files in the moz.build

jasonLaster:


darkwing:
@jasonLaster I think this is good, but I'd prefer if we could simply include MC's `moz.build` files in the GitHub repo.  That would cut down on release maintenance.

Issue #7537, opened 2018-12-18
Fix jsx-a11y/no-static-element-interactions warnings raised by jsx-a11y linter.


yzen:
To get a list of accessibility warnings raised by jsx-a11y linter run: `yarn lint:jsx-a11y`.

From that list please note all instances that are flagged to have "jsx-a11y/no-static-element-interactions" problem. Each one needs to be addressed individually.

Once all of the instances are resolved make sure to change this setting in `.eslintrc.jsx-a11y` from:
```
"jsx-a11y/no-static-element-interactions": ["warn", {
  "handlers": ["onClick", "onMouseDown", "onMouseUp", "onKeyPress", "onKeyDown", "onKeyUp"]
}],
```

to
```
"jsx-a11y/no-static-element-interactions": ["error", {
  "handlers": ["onClick", "onMouseDown", "onMouseUp", "onKeyPress", "onKeyDown", "onKeyUp"]
}],
```

Garbee:
So [this error](https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/no-static-element-interactions.md) is essentially, not using buttons for actions and instead using divs/spans/etc.

What we want to do here as detailed is go through the warnings, assess the context of the markup, and devise a strategy to handle the cases as consistently as possible. For example, if we have something making an image or a span with an onClick handler, we'd change those to buttons with the text/image node inside of that.

yzen:
That's correct. Buttons is just one of the interactive elements, there's also anchor and other types of inputs (radios, checkboxes) etc. In worst case scenario, and highly discouraged, ARIA role can be used to set the semantics of the interactive element.

Garbee:
Yup, we need to chose the appropriate interactive element to used depending on the task. But, we'd be better off taking advantage of them rather than using aria roles and scripting all the extra needs manually.

phoenixking25:
@ash-tamraz @Garbee is this issue free ?

ash-tamraz:
@phoenixking25 No, I'm still working on it and close to having a patch ready.

steffiau:
@yzen @ash-tamraz I'd like to help on this if it isn't being worked on anymore?

janodvarko:
Firefox Developer Tools (including Debugger tool) development is entirely happening in Bugzilla and new issues should be reported here: https://bugzilla.mozilla.org/enter_bug.cgi?product=DevTools&component=Debugger

[...]

PR #8152, opened 2019-03-23
Fixed jsx-a11y/no-static-element-interactions warnings raised by jsx-a11y linter

ash-tamraz:
Fixes issue #7537 

I know this generates errors from `yarn run test:all`. I'm looking for feedback on the current state of the fix to make sure I'm going in the right direction and everything looks correct so far. 

### Summary of Changes

* jsx-a11y linter raised "jsx-a11y/no-static-element-interactions" warnings. [Here is a description of the meaning of the warning.](https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/no-static-element-interactions.md)
* This commit makes changes to files in order to remove those generated warnings.
* I attempted to firstly use native HTML elements as opposed to adding unhelpful/inappropriate WAI-ARIA role, and added role values if I was unable to correct the warnings solely with restructuring the code just to use native HTML elements.

### Test Plan

To test:

- Run `yarn jest -u` to update your jest tests
- Then run `yarn run lint:jsx-a11y` to check the warnings from the jsx-a11y linter, or run `yarn run test:all` to run all the available tests

### DIFF
--- a/src/components/PrimaryPanes/SourcesTreeItem.js
+++ b/src/components/PrimaryPanes/SourcesTreeItem.js
@@ -249,6 +249,7 @@ class SourceTreeItem extends Component<Props, State> {
         key={item.path}
         onClick={this.onClick}
         onContextMenu={e => this.onContextMenu(e, item)}
+        role="presentation"
       >
         {this.renderItemArrow()}
         {this.getIcon(item, depth)}
--- a/src/components/shared/Dropdown.js
+++ b/src/components/shared/Dropdown.js
@@ -37,6 +37,8 @@ export class Dropdown extends Component<Props, State> {
         className="dropdown"
         onClick={this.toggleDropdown}
         style={{ display: this.state.dropdownShown ? "block" : "none" }}
+        role="menu"
+        tabIndex="0"
       >
         {this.props.panel}
       </div>
@@ -58,6 +60,7 @@ export class Dropdown extends Component<Props, State> {
         className="dropdown-mask"
         onClick={this.toggleDropdown}
         style={{ display: this.state.dropdownShown ? "block" : "none" }}
+        role="presentation"
       />
     );
   }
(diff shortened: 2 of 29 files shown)


darkwing:
@yzen Have time for feedback?

src/components/shared/Modal.js
```diff
@@ -23,18 +23,20 @@ type ModalProps = {
...
   render() {
     const { additionalClass, children, handleClose, status } = this.props;
 
     return (
-      <div className="modal-wrapper" onClick={handleClose}>
+      <div className="modal-wrapper">
         <div
           className={classnames("modal", additionalClass, status)}
-          onClick={this.onClick}
+          onMouseDown={this.onMouseDown}
```
yzen: Mouse down is not going to be accessible via keyboard, it should either remain onClick or have explicit handling of the " "/"Enter" key event.

src/components/shared/Dropdown.js
```diff
@@ -58,6 +60,7 @@ export class Dropdown extends Component<Props, State> {
         className="dropdown-mask"
         onClick={this.toggleDropdown}
         style={{ display: this.state.dropdownShown ? "block" : "none" }}
+        role="presentation"
```
yzen: it has a click handler so it does not seem to be for "presentation" purposes only.

src/components/PrimaryPanes/SourcesTreeItem.js
```diff
@@ -249,6 +249,7 @@ class SourceTreeItem extends Component<Props, State> {
         key={item.path}
         onClick={this.onClick}
         onContextMenu={e => this.onContextMenu(e, item)}
+        role="presentation"
```
yzen: This does not seem to be for presentation only since there's a click and context menu handlers. It should probably have some semantic meaning.

darkwing:
Closing for inactivity.  Please reopen if you have time to fix @yzen's suggestions, @ash-tamraz !  Thank you!

PR #8154, opened 2019-03-23
Show column number in Call Stack's Debug location / address

### LINKED ISSUE #8123: Call stack does not show column information
When stepping through columns in the same line, I want to see the column information in the callstack, so that I can tell frames apart.

image

ruturajv:
Fixes #8123 

I've ran the test cases, linter, not sure if anything else is pending. Please have a look.

### Summary of Changes

* Show column number in Call stack location
* Doesn't show col number if the column is `0`

### Screenshots
debug

### DIFF
--- a/src/components/SecondaryPanes/Frames/Frame.js
+++ b/src/components/SecondaryPanes/Frames/Frame.js
@@ -54,7 +54,11 @@ function FrameLocation({ frame, displayFullUrl = false }: FrameLocationProps) {
     <span className="location">
       <span className="filename">{filename}</span>:
       <span className="line">{location.line}</span>
-	  {location.column && <span>:<span className="column">{location.column}</span></span>}
+      {location.column && (
+        <span>
+          :<span className="column">{location.column}</span>
+        </span>
+      )}
     </span>
   );
 }
--- a/src/components/SecondaryPanes/Frames/Frame.js
+++ b/src/components/SecondaryPanes/Frames/Frame.js
@@ -27,9 +27,17 @@ function FrameTitle({ frame, options = {}, l10n }: FrameTitleProps) {
   return <span className="title">{displayName}</span>;
 }
 
-type FrameLocationProps = { frame: Frame, displayFullUrl: boolean };
+type FrameLocationProps = {
+  frame: Frame,
+  displayFullUrl: boolean,
+  showCol: boolean
+};
 
-function FrameLocation({ frame, displayFullUrl = false }: FrameLocationProps) {
+function FrameLocation({
+  frame,
+  displayFullUrl = false,
+  showCol = false
+}: FrameLocationProps) {
   if (!frame.source) {
     return null;
   }
@@ -54,11 +62,12 @@ function FrameLocation({ frame, displayFullUrl = false }: FrameLocationProps) {
     <span className="location">
       <span className="filename">{filename}</span>:
       <span className="line">{location.line}</span>
-      {location.column && (
-        <span>
-          :<span className="column">{location.column}</span>
-        </span>
-      )}
+      {showCol &&
+        location.column && (
+          <React.Fragment>
+            :<span className="column">{location.column}</span>
+          </React.Fragment>
+        )}
     </span>
   );
 }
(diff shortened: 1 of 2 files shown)


fvsch:
> Doesn't show col number if the column is `0`

Maybe we should, a col number at `0` is going to be rare anyway, since if you add a line breakpoint in most code you will probably be in a function or class method and the breakpoint would be at column `2`, `4` or more.

How does it look in, say, the methods of a React class component?

jasonLaster:
There was a conversation about when to show columns in the issue.

ruturajv:
> How does it look in, say, the methods of a React class component?

Ya, you're right adding some space before beginning any JS statement, makes the col change to greater than 0. Now the whole col. no. makes me wonder as an end-user, what use is the col. no. to me? the number really makes no difference, rather its the highlighted Blue arrow in the editor that is more relevant and is easier to comprehend. Is it the help that the developer can directly go to the col and make edits :thinking:  

Your thoughts @fvsch ?

src/components/SecondaryPanes/Frames/Frame.js
```diff
@@ -54,6 +54,11 @@ function FrameLocation({ frame, displayFullUrl = false }: FrameLocationProps) {
     <span className="location">
       <span className="filename">{filename}</span>:
       <span className="line">{location.line}</span>
+      {location.column && (
+        <span>
```
fvsch: Can we avoid the extra div here? A React.Fragment might work.

ruturajv: Thanks, I was getting a warning from React to counter - I added the extra span.

ruturajv:
> Well we will just have to update the tests.
Either we land this change on GitHub, and the weekly mercurial commit which imports GitHub changes will have to include a fix for this test.
Or we do the work on mozilla-central (using the Mercurial + Phabricator workflow), which allows making the change to the component and to the tests in the same commit.

I was wondering, to put the conditions in webconsole might be incorrect, instead if we can check for the callee and only then insert the col. no. 

I tried using `arguments.callee` but that is disabled in `strict` mode. I read somewhere about a synthetic `event` object passed everytime, but I couldn't find worthy info in that object.

Any ideas or fixing webconsole's test cases is the right way? @fvsch

fvsch:
> I was wondering, to put the conditions in webconsole might be incorrect, instead if we can check for the callee and only then insert the col. no.

The right way to do that would be with a component prop.

darkwing:
@ruturajv What was the rationale/idea behind the `showCol` option?  Is there somewhere we wouldn't want to show column (if not 0)?

ruturajv:
No, actually it seems this component[s] was getting used in WebConsole, and was troubling the Test cases, so had planned to use this prop. so that we could use `showCol` only for debugger UI and WebConsole UI can run as-is.

Unless we collectively think even Webconsole UI should always show the cols, then we'll have to update those test cases.

nchevobbe:
let's show column number in console as well :) 
The only reason we don't is because we we're using the debugger component (in the message location, we still use them)

ruturajv:
ok, great :)  So now how to get the `mozilla-central` and `debugger` repos in sync? Either of the commits _will_ trigger a bad build in CI. @nchevobbe ?

nchevobbe:
the best would be to do all the work in m-c, so you can fix the jest tests as well as the console mochitests

darkwing:
@ruturajv The best step forward would be closing this PR, setting up mozilla central locally (https://developer.mozilla.org/en-US/docs/Mozilla/Developer_guide/Build_Instructions/Mac_OS_X_Prerequisites#Getting_the_source), and submitting a new patch via Phabricator/moz-phab (https://moz-conduit.readthedocs.io/en/latest/phabricator-user.html#using-moz-phab).

darkwing:
I'll optimistically close this since the patch would break webconsole tests.

ruturajv:
Thanks @darkwing for the clarifications. Yes, I've already setup `m-c`, I'll try to fix it there.

PR #8157, opened 2019-03-24
Fixed Horizontal scrolling in the debugger secondary panel

### LINKED ISSUE #8129: [Scopes] Horizontally scrolling the Scopes panel hides its header
We are using `overflow: auto` on the wrong container for the Scopes panel, and when content overflows to the right and you scroll then the accordion header scrolls along with the content:

Selection_026

helenatxu:
Fixes #8129

### Summary of Changes

* Fixed the horizontal overflow of the headers in the debugger secondary panel.

### Test Plan
- [x] Open the DevTools and go to Debugger tab
- [x] Run the Debugger or inspect Scopes
- [x] You can also increase the font size so elements overflow horizontally
- [x] Now the horizontal overflow scrolls only for the data inside each section, but not for the headers

### DIFF
--- a/src/components/SecondaryPanes/SecondaryPanes.css
+++ b/src/components/SecondaryPanes/SecondaryPanes.css
@@ -3,6 +3,7 @@
  * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
 
 .secondary-panes {
+  overflow-x: hidden;
   display: flex;
   flex-direction: column;
   flex: 1;
--- a/src/components/SecondaryPanes/SecondaryPanes.css
+++ b/src/components/SecondaryPanes/SecondaryPanes.css
@@ -4,6 +4,7 @@
 
 .secondary-panes {
   overflow-x: hidden;
+  overflow-y: auto;
   display: flex;
   flex-direction: column;
   flex: 1;
--- a/src/components/SecondaryPanes/SecondaryPanes.css
+++ b/src/components/SecondaryPanes/SecondaryPanes.css
@@ -3,7 +3,6 @@
  * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
 
 .secondary-panes {
-  overflow: auto;
   display: flex;
   flex-direction: column;
   flex: 1;
--- a/src/components/shared/Accordion.css
+++ b/src/components/shared/Accordion.css
@@ -75,6 +75,7 @@
 .accordion ._content {
   border-bottom: 1px solid var(--theme-splitter-color);
   font-size: var(--theme-body-font-size);
+  overflow: auto;
 }
 
 .accordion div:last-child ._content {
--- a/src/components/shared/Accordion.css
+++ b/src/components/shared/Accordion.css
@@ -75,7 +75,6 @@
 .accordion ._content {
   border-bottom: 1px solid var(--theme-splitter-color);
   font-size: var(--theme-body-font-size);
-  overflow: auto;
 }
 
 .accordion div:last-child ._content {
--- a/src/components/SecondaryPanes/Breakpoints/Breakpoints.css
+++ b/src/components/SecondaryPanes/Breakpoints/Breakpoints.css
@@ -2,6 +2,10 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
 
+.breakpoints-pane > ._content {
+  overflow-x: auto;
+}
+
 .breakpoints-toggle {
   margin: 2px 3px;
 }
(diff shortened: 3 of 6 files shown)


helenatxu:
@fvsch could you add yourself as reviewer? Github is not giving me the option/permission to do it. Thank you!

src/components/shared/Accordion.css
```diff
@@ -75,6 +75,7 @@
 .accordion ._content {
   border-bottom: 1px solid var(--theme-splitter-color);
   font-size: var(--theme-body-font-size);
+  overflow: auto;
```
fvsch: It seems all the other panels are handling overflow themselves. Maybe we should do that for Scopes too, instead of handling it in Accordion?

src/components/SecondaryPanes/SecondaryPanes.css
```diff
@@ -3,7 +3,6 @@
  * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
 
 .secondary-panes {
-  overflow: auto;
```
fvsch: That's a good option, but I wonder if `overflow: hidden;` wouldn't work better here. In case we add more panes in the SecondaryPanes and we forget to manage potential overflow for new panes (or overflow in accordion headers). With `overflow: visible`, we may end up with content that overflows to the right and creates a horizontal scrollbar on the whole window.

src/components/shared/Accordion.css
helenatxu: @fvsch do you mean adding this `overflow: auto;` in each of the panels? I added in accordion because it seemed a more general bug affecting not only the Scopes panel only, as you can see in the image.
I will change it anyway and see how it looks like.
Screen Shot 2019-03-24 at 22 30 35 (2)

src/components/SecondaryPanes/SecondaryPanes.css
helenatxu: Ok, I will try that :)

helenatxu: So I added it as `overflow-x: hidden;` to prevent hiding the vertical overflow, which we actually want.

src/components/shared/Accordion.css
helenatxu: Done! :)

src/components/SecondaryPanes/SecondaryPanes.css
```diff
@@ -3,7 +3,7 @@
...
+  overflow-x: hidden;
```
fvsch: Turns out that using `overflow-x: hidden` automatically sets `overflow-y: auto`.
But if we were to remove that `overflow-x` declaration, then `overflow-y` would revert to `visible`, which we don't want.
Can you add an explicit `overflow-y: auto;` here to prevent that and better communicate our intent?

src/components/shared/Accordion.css
fvsch: My concern is that since we already have `overflow: auto` defined for a lot of SecondaryPanes content, generally with this structure:

```html
<li>
  <div class="_header">...</div>
  <div class="_content">
    <div class="some-container-class">...</div<
  </div>
</li>
```

...where "some-container-class" already has `overflow: auto`, if we add this style to all the `_content` classes than we have two parent/child containers with this style. I'd like to avoid that. For the time being, can you avoid declaring this overflow style for all `.accordion ._content` elements, and only for the panes that are missing this style?

src/components/SecondaryPanes/SecondaryPanes.css
helenatxu: I didn't think about that! Added :)

darkwing:
How does this look @fvsch ?

src/components/shared/Accordion.css
fvsch: We still need to remove this `overflow: auto` here.
Everything else looks great.

helenatxu: Sorry! I forgot about that one. Thanks!

fvsch:
@darkwing We should be good to go.

helenatxu:
Hello @darkwing and @fvsch Travis CI says there is an error, but I don't understand which one.

darkwing:
@helenatxu If you rebase, it should be fixed :)

helenatxu:
Finally the Travis test passing :) @fvsch you cant approve it now if it looks fine for you.

fvsch:
Had forgotten to approve it. I never remember the GitHub UI for that. :stuck_out_tongue:

helenatxu:
@darkwing maybe it needs your approval too?

fvsch:
Yep it needs to be accepted and merged by a team member with the right permissions. Probably tomorrow. :)

Issue #8149, opened 2019-03-21
[Threads] Align threads list with UI spec


fvsch:
UI spec here: https://github.com/firefox-devtools/ux/issues/45#issuecomment-457021260
It doesn't include anything on Threads, but the same general rules should apply.

fvsch:
Related issue: #8145.

nagpalm7:
@fvsch Can I take this issue?

fvsch:
Definitely. :)

fvsch:
Closing this issue in favor of the Bugzilla [bug 1543581](https://bugzilla.mozilla.org/show_bug.cgi?id=1543581), which should be close to landing.

Issue #8141, opened 2019-03-19
[XHR Breakpoints] Make the method selector smaller


fvsch:
With @Jaril's work in #8094 and https://github.com/firefox-devtools/ux/issues/45 we are going to have smaller XHR breakpoint items, 20px tall.

When going to edit mode, though, the input takes more vertical space:

1. The container's 1px border makes the height 22px instead of 20px. Maybe we should try an inset outline instead? As in: `outline: solid 1px; outline-offset: -1px;`.
2. The `<select>` itself is rather big, at least on Linux its easily more than 30px tall.

We have a minimal `<select>` style used in Firefox DevTools, for instance in the Responsive Design Mode toolbar or in the Network panel toolbar. We should reuse that style here.

I'm getting good results with this CSS:

```css
.xhr-input-method {
  margin-inline-start: 4px;
  padding-block: 0;
  padding-inline-start: 2px;
  padding-inline-end: 14px;
  border: none;
  border-radius: 2px;
  font: inherit;
  color: inherit;
  background-color: var(--toolbarbutton-background);
  -moz-appearance: none;
  appearance: none;
}
```

Then the only thing missing is adding the double-arrow icon to the right of the select.

fvsch:
There's a UI spec draft in https://github.com/firefox-devtools/ux/issues/50
But I see the value in being consistent with the select buttons in RDM / Network, so I would still go with that.

blueberryapple:
so I added the css changes suggested, this is what it currenly looks like after adding in the background image for the arrow:

![image](https://user-images.githubusercontent.com/9259833/57118894-14155e00-6d1b-11e9-938a-2d3df94ff9eb.png)

[here is the commit with the changes](https://github.com/blueberryapple/debugger/commit/9d51e45181ceb983b062dc4baa2b5b99c7afd44f#diff-a81c5afc411907c1034050a1a2c8dee7)

I'm not too sure how to transform the svg color, since the other places use `background-color` to change the icon which wouldn't really work in this scenario.

fvsch:
Hmm. We have a way that works in Firefox only, using `fill="context-fill" in the SVG code and this in the CSS:

```
-moz-context-properties: fill;
fill: var(--theme-icon-dimmed-color);
```

It won't work in Chrome but maybe that's an acceptable trade-off?
Especially if we use a medium gray fallback that is still visible in Chrome, e.g.:

```
<path fill="context-fill #808080" d="..." />
```

[...]

PR #8162, opened 2019-03-26
Revert "ligned breakpoints and XHR breakpoints panes (#8094)"

darkwing:
This reverts commit 6b4110eaf3638c3635b765b1b4781d8ca7b6cf7b.

I missed a few CSS issues with this PR that I've spotted later:

1.  There's too much space above and below the the expression and XHR inputs
2.  The breakpoint line-height seems small

FewUIIssues

### DIFF
--- a/src/components/SecondaryPanes/Breakpoints/Breakpoints.css
+++ b/src/components/SecondaryPanes/Breakpoints/Breakpoints.css
@@ -6,24 +6,20 @@
   margin: 2px 3px;
 }
 
+.pane.breakpoints-list {
+  padding-bottom: 0.35em;
+}
+
 .breakpoints-list * {
   user-select: none;
 }
 
-.breakpoints-list {
-  margin-top: 4px;
-  margin-bottom: 4px;
-}
-
 .breakpoints-list .breakpoint-heading {
   text-overflow: ellipsis;
+  overflow: hidden;
+  display: flex;
   width: 100%;
-  font-size: 12px;
-  line-height: 16px;
-}
-
-.breakpoint-heading:not(:first-child) {
-  margin-top: 2px;
+  align-items: center;
 }
 
 .breakpoints-list .breakpoint-heading .filename {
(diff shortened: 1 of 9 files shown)


darkwing:
@jaril Alternatively a new PR with a few fixes would be excellent too!

cc @fvsch

fvsch:
@darkwing As far as I can see, everything is going according to plan. :)

The issues you're seeing are IMO due to two things:

1. Probably the force of habit. We changed some things, giving padding to accordion content and making content rows tighter (20px height, instead of the previous semi-random heights in the 26-30px range). Breakpoint rows were the most airy of all secondary pane content, contrasting a lot with the very squished Scopes, so now the 20px-tall breakpoint rows may seem too small.

2. We haven't landed all the work on secondary panes vertical rhythm, so there are some strong visual differences between the inputs in Watch Expressions (tall, to be fixed) and XHR Breakpoints (narrower, fixed).

My take is that we should land and/or work on the remaining bits that address spacing in Secondary Panes (following @jaril's comprehensive mockups), review the complete result when we're done, and make adjustments then.

Pieces to land and/or work on:

- #8066 (Watch expressions and Scopes)
- #8141 (XHR breakpoint method select)
- #8149 (Threads pane)

> There's too much space above and below the the expression and XHR inputs

The XHR input is actually closer to the spec. The 4px of empty space before and after the input is intented. It's not a great solution, but it's the more rational way to manage situations where the UI has a series of existing rows plus an input row; we would like the input row to use the same height as the saved rows (20px), to make things more cohesive and to avoid vertical jumps.

darkwing:
Thank you for the update @fvsch!  I appreciate the detail!

We're very early in the next FF cycle so I'm fine with additions or fixes.  Not to disparage your or @jaril's work, but the space above and below the Expression and XHR BP feels like an eye sore; as a user, I kind of see it as an obvious "mistake"; it makes the input area seem very big, especially when all of the other "rows" (breakpoints, scopes, etc.) have become so "thin".

I'm good with the thinner rows but I'd really like us to figure out how to make the input areas not seem so large.

fvsch:
The watch expression one is definitely too large, but merging Jaril's PR would help. :)

If we keep the inputs at 20px and the current padding (in XHR Breakpoint), that's a 28px height total, same height as the same pane with one XHR Breakpoint set up. So I'm not worried about it looking too big.

I'm a bit worried about the padding looking like a sore thumb when we just have one input, though. But it will be clearer to see if we need changes there once the PRs land and the dust settles.

PR #8175, opened 2019-04-01
[SourcesTree] Sort SourcesTree by url when files share same name

### LINKED ISSUE #7160: [SourcesTree] Sort SourcesTree when Querystring present

anthonyx:
Fixes #7160 

### Summary of Changes

* Sort by entire url (including parameters) rather than name alone when files share same name
* Only sorts when naming conflicts exist

### Test Plan

* Modified mochitest (`doc-sources-querystring.html`) to test for sources in order

Example test plan:

1. Open https://fortunate-skunk.glitch.me/ 
2. Open debugger
3. See sources in order

Before/After
image

### DIFF
--- a/src/utils/sources-tree/addToTree.js
+++ b/src/utils/sources-tree/addToTree.js
@@ -48,7 +48,8 @@ function findOrCreateNode(
   part: string,
   index: number,
   url: Object,
-  debuggeeHost: ?string
+  debuggeeHost: ?string,
+  source: Source
 ): TreeDirectory {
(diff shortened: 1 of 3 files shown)


darkwing:
Awesome work!!!

PR #8179, opened 2019-04-03
remove skip from grip.js test

### LINKED ISSUE #6147: Grip rep - Handle uninterresting properties
_From @nchevobbe on June 21, 2017 10:8_

This is kind of blocked by https://github.com/devtools-html/devtools-core/pull/466
In this PR, we have a test that we skip because at the moment the rep doesn't output what it should.

In order to fix this issue, we need to:
1. wait for the PR to be merges
2. remove the `.skip` part of the `Grip - Object with uninteresting properties -> renders as expected` test.
3. Make sure the test pass (by running `cd packages/devtools-reps && yarn test`)

This shouldn't be too complex, we already handle those kind of things in [the GripMap rep](https://github.com/devtools-html/devtools-core/blob/master/packages/devtools-reps/src/reps/grip-map.js#L80)

_Copied from original issue: devtools-html/devtools-core#467_

GettyOrawo:
Fixes #6147

### Summary of Changes

* remove skipp from `Grip - Object with uninteresting properties -> renders as expected test`

### Test Plan

Tweaked the tests a little and changed LONG MODE not to use default output for tests to pass

Screenshot from 2019-04-03 10-34-16

### DIFF
--- a/packages/devtools-reps/src/reps/tests/grip.js
+++ b/packages/devtools-reps/src/reps/tests/grip.js
@@ -276,14 +276,12 @@ describe("Grip - Object with uninteresting properties", () => {
     // @TODO This is broken at the moment.
     // See https://bugzilla.mozilla.org/show_bug.cgi?id=1276376
     const renderRep = props => shallowRenderRep(object, props);
-    const defaultOutput = 'Object { a: undefined, c: "c", d: 1, … }';
+    const defaultOutput = 'Object { c: "c", d: 1, a: undefined, … }';
 
     expect(renderRep({ mode: undefined }).text()).toBe(defaultOutput);
     expect(renderRep({ mode: MODE.TINY }).text()).toBe("{…}");
     expect(renderRep({ mode: MODE.SHORT }).text()).toBe(defaultOutput);
-    expect(renderRep({ mode: MODE.LONG }).text()).toBe(
-      'Object { a: undefined, b: undefined, c: "c", d: 1 }'
-    );
+    expect(renderRep({ mode: MODE.LONG }).text()).toBe(defaultOutput);
   });
 });
 
--- a/packages/devtools-reps/src/reps/tests/grip.js
+++ b/packages/devtools-reps/src/reps/tests/grip.js
@@ -272,16 +272,18 @@ describe("Grip - Object with uninteresting properties", () => {
     expect(getRep(object)).toBe(Grip.rep);
   });
 
-  it.skip("renders as expected", () => {
+  it("renders as expected", () => {
     // @TODO This is broken at the moment.
     // See https://bugzilla.mozilla.org/show_bug.cgi?id=1276376
     const renderRep = props => shallowRenderRep(object, props);
-    const defaultOutput = 'Object {c: "c", d: 1, a: undefined, more...}';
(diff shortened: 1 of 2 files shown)


GettyOrawo:
Hey @nchevobbe are the changes okay for this one?

packages/devtools-reps/src/reps/tests/grip.js
```diff
@@ -272,16 +272,18 @@ describe("Grip - Object with uninteresting properties", () => {
     expect(getRep(object)).toBe(Grip.rep);
   });
 
-  it.skip("renders as expected", () => {
+  it("renders as expected", () => {
     // @TODO This is broken at the moment.
     // See https://bugzilla.mozilla.org/show_bug.cgi?id=1276376
     const renderRep = props => shallowRenderRep(object, props);
-    const defaultOutput = 'Object {c: "c", d: 1, a: undefined, more...}';
+    const defaultOutput = 'Object { a: undefined, c: "c", d: 1, … }';
```
nchevobbe: In the test, we only want to remove the `.skip` (so the test will run).
The test should stay the same, and we should fix the `Grip` rep instead :)

GettyOrawo: oh that makes sense. Alright I must have gotten it wrong. Alright let me take a second look at why they fail.

GettyOrawo:
So I almost managed to track where this behaviour was being created. There's this line https://github.com/firefox-devtools/debugger/blob/master/packages/devtools-reps/src/reps/grip.js#L245 
 where we order indexes by ascending order.
Here's the initial error that I am to sort

Screenshot from 2019-04-04 10-38-39

On commenting out line 246-248 I get this error from the `grip.js test ` 

Screenshot from 2019-04-04 11-02-34

`Expected: "Object {c: \"c\", d: 1, a: undefined, more...}"`
` Received: "Object { c: \"c\", d: 1, a: undefined, … }"`

it seems close to the output we are targeting but there's this `more...` part in the tests. Is the output supposed to display that too? and also the output seems to have white spaces after and before the curly-braces. Should these white spaces also be removed?

Also on commenting out the code I mentioned above, other tests seem to break since they are not ordered. Like `event.js`  gives the error below. 

Screenshot from 2019-04-04 11-08-07

Are these two files related? and is there a way I can do this only targeting the grip.js ?

nchevobbe:
oh you're right, the `more...` should be turned into `…`, that's a change we made a while ago (and we did updated this test since it doesn't run at the moment)

GettyOrawo:
Hey @nchevobbe I still have failing tests since the output is being tested to be the same for both the SHORT and LONG modes however, the LONG mode is different. How do I go about this without changing the tests?

Then secondly some other tests are failing since I silenced the line that orders them. Does this change only apply to grip.js or will I have to change the output that the others are testing?

nchevobbe:
Thank you for your patience @GettyOrawo .
So I was looking at the test, and the output feels a bit weird. I checked Chrome, and they don't do such "uninteresting" sort, and I guess it would be weird to not be consistent.
So, what do you think about completely removing the test and reverting the change (where you removed the `sort`) ?

GettyOrawo:
Sure  @nchevobbe if it's the best approach towards fixing this

PR #8191, opened 2019-04-17
Changed gap placement on popover orientation-right

### LINKED ISSUE #8107: [Preview] Mousing from the token to the popup fails
This is a fun bug if you mouse from the token to the popup without being careful it can close. The reason for this is that we only watch to see if your're either on the token, arrow, or popup, not the gap in between which i highlighted red.

gap

kciccolella:
Fixes #8107 

### Summary of Changes

In the unchanged code, the gap is placed above the popover if the orientation is right.

I decided to use `flex-direction: row` in the popover.css file in order to change the placement of the gap, from above the popover to the left.  I then had to make some minor adjustments to the `calculateLeft` function and the `getPopoverArrow` function in the popover.js file to accommodate the changes made.

### Screenshots

**Gap placement before change:**
image

**Gap placement after change:**
image

### Test Plan

- [x] Decrease the width of the editor pane or increase zoom if necessary
- [x] Set a breakpoint
- [x] Scroll until breakpoint is approximately in the middle of the editor pane
- [x] Trigger breakpoint
- [x] Hover over a token

### DIFF
--- a/src/components/shared/Popover.css
+++ b/src/components/shared/Popover.css
@@ -8,8 +8,8 @@
 }
 
 .popover.orientation-right {
-	display: flex;
-	flex-direction: row;
+  display: flex;
+  flex-direction: row;
 }
 
 .popover.orientation-right .gap {
--- a/src/components/shared/Popover.css
+++ b/src/components/shared/Popover.css
@@ -7,7 +7,16 @@
   z-index: 100;
 }
 
-.popover .gap {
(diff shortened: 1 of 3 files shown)


darkwing:
Thank you for contributing @kciccolella !  I'll have a look today :)

darkwing:
This is excellent @kciccolella !  Thank you so much!

PR #8201, opened 2019-04-23
Improve performance of "Go to file"

sorin-davidoi:
### Summary of Changes

* avoid allocating intermediary array
* use sets to avoid linear look-up
* debounce source transformations and filtering
* avoid unnecessary layout computations when scrolling

### Test Plan

- [x] Assert that searching yields the same results as before
- [x] Assert that opened tabs are shown by default

### DIFF
--- a/src/components/QuickOpenModal.js
+++ b/src/components/QuickOpenModal.js
@@ -7,6 +7,7 @@ import React, { Component } from "react";
 import { connect } from "../utils/connect";
 import fuzzyAldrin from "fuzzaldrin-plus";
 import { basename } from "../utils/path";
+import { debounce } from "lodash";
 
 import actions from "../actions";
 import {
@@ -43,7 +44,6 @@ import "./QuickOpenModal.css";
 
 type Props = {
   enabled: boolean,
-  sources: Array<Object>,
   selectedSource?: Source,
   query: string,
   searchType: QuickOpenType,
@@ -55,7 +55,8 @@ type Props = {
   setQuickOpenQuery: typeof actions.setQuickOpenQuery,
   highlightLineRange: typeof actions.highlightLineRange,
   closeQuickOpen: typeof actions.closeQuickOpen,
-  toggleShortcutsModal: () => void
+  toggleShortcutsModal: () => void,
+  computeSources: (tabUrls: Set<$PropertyType<Tab, "url">>) => Array<Object>
 };
 
 type State = {
@@ -69,12 +70,16 @@ type GotoLocationType = {
   column?: number
 };
 
+const updateResultsDebounce = 100;
 const maxResults = 100;
 
 function filter(values, query) {
+  const preparedQuery = fuzzyAldrin.prepareQuery(query);
+
   return fuzzyAldrin.filter(values, query, {
     key: "value",
-    maxResults: maxResults
+    maxResults: maxResults,
+    preparedQuery
   });
 }
 
@@ -127,7 +132,9 @@ export class QuickOpenModal extends Component<Props, State> {
   };
 
   searchSources = (query: string) => {
-    const { sources } = this.props;
+    const { computeSources, tabs } = this.props;
+    const tabUrls = new Set(tabs.map((tab: Tab) => tab.url));
+    const sources = computeSources(tabUrls);
     const results =
       query == "" ? sources : filter(sources, this.dropGoto(query));
     return this.setResults(results);
@@ -158,16 +165,17 @@ export class QuickOpenModal extends Component<Props, State> {
   };
 
   showTopSources = () => {
-    const { tabs, sources } = this.props;
+    const { tabs, computeSources } = this.props;
+    const tabUrls = new Set(tabs.map((tab: Tab) => tab.url));
+    const sources = computeSources(tabUrls);
     if (tabs.length > 0) {
-      const tabUrls = new Set(tabs.map((tab: Tab) => tab.url));
       this.setResults(sources.filter(source => tabUrls.has(source.url)));
     } else {
       this.setResults(sources);
     }
   };
 
-  updateResults = (query: string) => {
+  updateResults = debounce((query: string) => {
     if (this.isGotoQuery()) {
       return;
     }
@@ -185,7 +193,7 @@ export class QuickOpenModal extends Component<Props, State> {
     }
 
     return this.searchSources(query);
-  };
+  }, updateResultsDebounce);
 
   setModifier = (item: QuickOpenResult) => {
     if (["@", "#", ":"].includes(item.id)) {
@@ -383,6 +391,8 @@ export class QuickOpenModal extends Component<Props, State> {
     const items = this.highlightMatching(query, results || []);
     const expanded = !!items && items.length > 0;
 
+    console.log('results', results);
+
     return (
       <Modal in={enabled} handleClose={this.closeModal}>
         <SearchInput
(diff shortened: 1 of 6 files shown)


jasonLaster:
Thanks! I noticed that quick open felt sluggish yesterday. Could you share a before/after profile. It is fine to use https://profiler.firefox.com or chrome's profiler :)

jasonLaster:
By the way, feel free to join us in [slack](https://devtools-html-slack.herokuapp.com/). I couldn't help but notice that you forked zone.js :) Zone is incredibly cool in what they're trying to do.

sorin-davidoi:
The differences are small unfortunately, at least when few tabs are opened. I've been trying different things and thought I should propose this change even though the impact is negligible.

[Before](https://profiler.firefox.com/public/c1e04f3d620b22a42ac77a82bfd408446bdf7d8a/stack-chart/?implementation=js&published&v=3).

[After](https://profiler.firefox.com/public/073340849b17fcdb203bd0e1233b1fb653731c07/stack-chart/?implementation=js&published&v=3).

jasonLaster:
When quick open has a lot of results (1000+) I noticed that the fuzzy search was a bottleneck. Here it looks react related. Are there things we can do to help react not re-render or be further optimized...

https://profiler.firefox.com/public/c1e04f3d620b22a42ac77a82bfd408446bdf7d8a/stack-chart/?globalTrackOrder=0-1-2&implementation=js&localTrackOrderByPid=16850-0~17325-0~&range=9.0730_9.2868&thread=2&v=3

sorin-davidoi:
Adding a small debounce to the query and delaying the computation of sources seems to have improved things quite a bit.

Not sure why some unit tests are failing, will take a look later.

[chrome_performance_trace.zip](https://github.com/firefox-devtools/debugger/files/3113858/chrome_performance_trace.zip)

src/utils/result-list.js
```diff
@@ -42,12 +42,16 @@ function chromeScrollList(elem: Element, index: number): void {
...
 
-  const resultsHeight: number = resultsEl.clientHeight;
-  const itemHeight: number = resultsEl.children[0].clientHeight;
-  const numVisible: number = resultsHeight / itemHeight;
-  const positionsToScroll: number = index - numVisible + 1;
-  const itemOffset: number = resultsHeight % itemHeight;
-  const scroll: number = positionsToScroll * (itemHeight + 2) + itemOffset;
-
-  resultsEl.scrollTop = Math.max(0, scroll);
+  requestAnimationFrame(() => {
```
sorin-davidoi: See https://nolanlawson.com/2018/09/25/accurately-measuring-layout-on-the-web/ for some context.

jasonLaster: nice

sorin-davidoi:
Before I go down the mochitests rabbit hole can I get some feedback on the general approach?

src/components/QuickOpenModal.js
```diff
@@ -422,16 +430,21 @@ export class QuickOpenModal extends Component<Props, State> {
...
     symbolsLoading: isSymbolsLoading(state, selectedSource),
     query: getQuickOpenQuery(state),
     searchType: getQuickOpenType(state),
-    tabs: getTabs(state)
+    tabs,
+    computeSources: (
+      tabUrls: Set<$PropertyType<Tab, "url">>
+    ): Array<Object> => {
+      return formatSources(getDisplayedSourcesList(state), tabUrls);
+    }
```
jasonLaster: including functions in MSPs can trigger unnecessary re-renders. 

What is it trying to do?

sorin-davidoi: This used to be computed in `mapStateToProps`, but if we kept it there we were computing it way too often (making it a bottleneck). The first approach was to declare the function at the top of the file, but then the component has a dependency on the entire state, making it hard to test (you need to make parts of the state). So I decided to keep it in `mapStateToProps` but make it be computed on demand.

You're right about the re-renderes, I think that by declaring the function once and always passing it (the same instance) via props should solve that, correct?

jasonLaster: > You're right about the re-renderes, I think that by declaring the function once and always passing it (the same instance) via props should solve that, correct?

 I think so. Also, you should be safe to use `getDisplayedSourcesList(state)` in the MSP because that is memoized and then call formatSources where ever you'd like...

jasonLaster:
This is looking really good. I just tested it and it felt nice.

Here is a [link][l] to a before after recording (probably similar to what you have).

And a similar GIF without the change it is approximately ~40ms with ~20ms... for todomvc which is small



[l]: https://drive.google.com/drive/folders/1V0IXuQVIP0YERskpw1vMf7MBVutG7ymy?usp=sharing

jasonLaster:
I'd be happy to help fix the one mochitest.

src/components/QuickOpenModal.js
```diff
@@ -158,16 +165,24 @@ export class QuickOpenModal extends Component<Props, State> {
...
-    const { tabs, sources } = this.props;
+    const { displayedSources, tabs } = this.props;
+    const tabUrls = new Set(tabs.map((tab: Tab) => tab.url));
+
     if (tabs.length > 0) {
-      const tabUrls = tabs.map((tab: Tab) => tab.url);
-      this.setResults(sources.filter(source => tabUrls.includes(source.url)));
+      this.setResults(
+        formatSources(
+          displayedSources.filter(
```
sorin-davidoi: We now filter before we map.

sorin-davidoi:
> I'd be happy to help fix the one mochitest.

Would appreciate that, have some trouble setting them up right now.

sorin-davidoi:
@jasonLaster Fixed the mochitests.

src/components/QuickOpenModal.js
```diff
@@ -7,6 +7,7 @@ import React, { Component } from "react";
 import { connect } from "../utils/connect";
 import fuzzyAldrin from "fuzzaldrin-plus";
 import { basename } from "../utils/path";
+import { debounce } from "lodash";
```
jasonLaster: I think we want to use `throttle`

src/components/test/QuickOpenModal.spec.js
```diff
@@ -7,9 +7,14 @@
...
 import React from "react";
 import { shallow, mount } from "enzyme";
+import lodash from "lodash";
 import { QuickOpenModal } from "../QuickOpenModal";
 
 jest.mock("fuzzaldrin-plus");
+jest.unmock("lodash");
+
+// $FlowIgnore
+lodash.debounce = jest.fn(fn => fn);
```
jasonLaster: nice!

src/utils/result-list.js
jasonLaster: Lets turn this into a comment with that links so future readers will see it

test/mochitest/browser_dbg-quick-open.js
```diff
@@ -47,10 +47,13 @@ function resultCount(dbg) {
...
 }
 
-function quickOpen(dbg, query, shortcut = "quickOpen") {
+async function quickOpen(dbg, query, shortcut = "quickOpen") {
   pressKey(dbg, shortcut);
   assertEnabled(dbg);
+
   query !== "" && type(dbg, query);
+
+  await waitForTime(150);
```
jasonLaster: this is fine, but it would be better if we could wait for something UI related like what we do above with resultCount.

We have several helpers for waiting for UI elements. Again, this is totally fine to ignore

darkwing:
Hello @sorin-davidoi !  Jason found a few items we should update before a final merge -- my mistake for merging too early.  Could you address Jason's feedback and then we can get this in?  Thank you! 👍

src/components/QuickOpenModal.js
```diff
@@ -69,12 +70,16 @@ type GotoLocationType = {
   column?: number
 };
 
+const updateResultsDebounce = 100;
 const maxResults = 100;
 
 function filter(values, query) {
+  const preparedQuery = fuzzyAldrin.prepareQuery(query);
```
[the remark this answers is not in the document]
sorin-davidoi: [Parses the query in advance](https://github.com/jeancroy/fuzz-aldrin-plus/blob/master/src/query.coffee#L13) - for some reason `filter` doesn't seem to perform this optimization (although it could).

src/components/QuickOpenModal.js
```diff
@@ -158,16 +165,17 @@ export class QuickOpenModal extends Component<Props, State> {
...
-      const tabUrls = tabs.map((tab: Tab) => tab.url);
-      this.setResults(sources.filter(source => tabUrls.includes(source.url)));
+      this.setResults(sources.filter(source => tabUrls.has(source.url)));
     } else {
       this.setResults(sources);
     }
   };
 
-  updateResults = (query: string) => {
+  updateResults = debounce((query: string) => {
```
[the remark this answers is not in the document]
sorin-davidoi: @jasonLaster Can we reconsider this? Typing a long file name still feels janky when there are more than 1000 files.

PR #8203, opened 2019-04-25
refactor(components/search-input): Don't manage focus via JavaScript

sorin-davidoi:
### Summary of Changes

`SearchInput` listens for `onFocus` & `onBlur` and sets the `.focused` class manually. Each such event triggers a component update (it calls `.setState`). Other components that wrap `SearchInput` (`SearchBar`, `ProjectSearch`) have the same approach, so the events trigger a cascade of component updates.

* Use `:focus-within`to style components, not manually managed class

### Test Plan

Focus the search inputs and check that the styles are applied.

Example test plan:

- [x] Open search in file input
- [x] Open find in files input

focus


src/components/ProjectSearch.js
```diff
@@ -178,8 +176,7 @@ export class ProjectSearch extends Component<Props, State> {
   onEnterPress = () => {
     if (
       !this.isProjectSearchEnabled() ||
-      !this.state.focusedItem ||
-      this.state.inputFocused
+      !this.state.focusedItem
```
sorin-davidoi: I'm not sure how to trigger this code-path, pressing `Enter` doesn't seem to do it.

jasonLaster:
Seems reasonable. Does someone have time to review?

CC @princiya @AnshulMalik @darkwing @fvsch

AnshulMalik:
This is awesome :)

PR #8210, opened 2019-04-30
Update SecondaryPanes splitter

derek-li:
### Summary of Changes
 -Removed unnecessary div from SecondaryPanes.js that was causing the splitter from filling the box

Before:
image1

After:
image2

### DIFF
--- a/src/components/SecondaryPanes/index.js
+++ b/src/components/SecondaryPanes/index.js
@@ -433,21 +433,19 @@ class SecondaryPanes extends Component<Props, State> {
 
   renderVerticalLayout() {
     return (
-      <div>
-        <SplitBox
-          initialSize="300px"
-          minSize={10}
-          maxSize="50%"
-          splitterSize={1}
-          startPanel={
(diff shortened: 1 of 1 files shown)


src/components/SecondaryPanes/index.js
```diff
@@ -433,21 +433,19 @@ class SecondaryPanes extends Component<Props, State> {
...
-          splitterSize={1}
-          startPanel={
-            <div style={{ width: "inherit" }}>
-              <WhyPaused delay={this.props.renderWhyPauseDelay} />
-              <Accordion items={this.getStartItems()} />
-            </div>
-          }
-          endPanel={<Accordion items={this.getEndItems()} />}
-        />
-      </div>
```
jasonLaster: Thanks. I wonder where this came from?

derek-li: I think it was me actually, haha. I was having some issues making it render properly with the WhyPaused redesign.

PR #8214, opened 2019-05-07
refactor(components/editor): Allow one paint before updating the DOM

sorin-davidoi:
### Summary of Changes

Improves perceived responsiveness by rendering the loading state before freezing the UI. One of the things that makes the editor feel sluggish is that sometimes is _seems_ to not respond to clicks (e.g. you click on a tab or a file and nothing happens for a few seconds). This is because CodeMirror sometimes takes seconds to render the text, blocking the UI. This patch makes sure the editor responds immediately to the interaction by rendering the loading state before the UI is blocked by CodeMirror.

### Test Plan

Opened multiple sources, switched between tabs to make sure that they render.

Example test plan:

- [x] Select a large file (I've tested with a 2MB compressed JavaScript file)
- [x] Observe that the panel closes and "Loading..." is displayed (before the panel would stay opened until the source was rendered).

#### Before (the UI freezes while the source is loaded, you can see the hover effect being visible for a few seconds)
before

#### After (the UI freezes while the source is loaded but the "Loading..." message is displayed first - the hover effect also dissapears)
after

### DIFF
--- a/src/components/Editor/index.js
+++ b/src/components/Editor/index.js
@@ -151,9 +151,9 @@ class Editor extends PureComponent<Props, State> {
 
     // Allow one paint before updating the DOM to feel responsive
     requestAnimationFrame(() => {
-      setTimeout(() => {
+      requestAnimationFrame(() => {
         this.setText(nextProps);
-        this.setSize(nextProps);
+        this.setSize(nextProps, prevProps);
         this.scrollToLocation(nextProps, prevProps);
       });
     });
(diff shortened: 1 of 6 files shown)


src/components/Editor/index.js
```diff
@@ -143,9 +143,16 @@ class Editor extends PureComponent<Props, State> {
       return;
     }
 
-    this.setText(nextProps);
-    this.setSize(nextProps);
-    this.scrollToLocation(nextProps);
+    showLoading(this.state.editor);
```
sorin-davidoi: Maybe we could do a quick check for the size of the source and avoid flashing the loading state if it is "small".

jasonLaster:
This is awesome. I'd love to see if we can get these perf wins w/ passing tests.

darkwing:
@sorin-davidoi I'm not familiar with the nested rAF usage -- could you provide some insight?

darkwing:
Could you also rebase @sorin-davidoi ?  Thank you!

sorin-davidoi:
Closing this since sprinkling `rAF`s all over the place is not a viable approach.

jasonLaster:
Thanks for looking into this path. I'd love to continue experimenting and find the right option - we have to make this part faster or react just won't be the right architecture.

PR #8216, opened 2019-05-09
Improve readability of stepping button tool tips labels

### LINKED ISSUE #6993: Confusing tooltips on step buttons
It's hard for me to understand the tooltips showing the keyboard shortcuts for the step buttons.  I think it would be better to use parenthesses and add an 'or' between the two shortcuts. ie: `(⌘\ or F8)`

The tooltips for the toolbox panels use `cmd` instead of `⌘` and also use '+' for multikey shortcuts.  Do you want to do this here as well?

kenjiO:
Fixes #6993

I've been away a while so I thought I would try an easy one to get reacquainted with the project.  It looks like no one has touched this in 3 months so I took it.  I hope you don't mind.

### Summary of Changes
* Add parenthesis and 'or' to step button tool tips to make them more readable

**Before**

**After**

### DIFF
--- a/src/components/SecondaryPanes/CommandBar.js
+++ b/src/components/SecondaryPanes/CommandBar.js
@@ -69,7 +69,7 @@ function formatKey(action) {
     const winKey =
       getKeyForOS("WINNT", `${action}Display`) || getKeyForOS("WINNT", action);
     // display both Windows type and Mac specific keys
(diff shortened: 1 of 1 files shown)


darkwing:
Thank you @kenjiO !  I'll have a look today!

PR #8217, opened 2019-05-10
Update debugger to MC after release 133

darkwing:
WIP.


darkwing:
A few points:

-  The test failure is an intermittent, passes locally every time
-  Jest test failures are present on MC so I'm forgiving those in order to get this in.

PR #8257, opened 2019-08-16
Editing breakpoints - Re-enables breakpoints globally

grlwholifts:
Fixes #1555333 on Bugzilla: https://bugzilla.mozilla.org/show_bug.cgi?id=1555333

In the file: `devtools/client/debugger/src/actions/breakpoints/index.js`, made the following changes: 

1. When a user clicks on a gutter to add a breakpoint, skipPausing is set to false (if it was set to true).
2. When a user enables a breakpoint in the secondary panel, skipPausing is set to false (if it was set to true).

CC: @jasonLaster @digitarald @darkwing

### DIFF
--- a/src/actions/breakpoints/index.js
+++ b/src/actions/breakpoints/index.js
@@ -79,7 +79,7 @@ export function enableBreakpointsInSource(cx: Context, source: Source) {
     for (const breakpoint of breakpoints) {
       if (breakpoint.disabled) {
         dispatch(enableBreakpoint(cx, breakpoint));
-      if (!getState.disabled) {
+      if (!getSkipPausing(getState())) {
         dispatch(toggleBreakpoints(cx, breakpoint));
         }
       }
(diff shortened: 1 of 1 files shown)


src/actions/breakpoints/index.js
```diff
@@ -79,6 +79,9 @@ export function enableBreakpointsInSource(cx: Context, source: Source) {
     for (const breakpoint of breakpoints) {
       if (breakpoint.disabled) {
         dispatch(enableBreakpoint(cx, breakpoint));
+      if (!getState.disabled) {
```
jasonLaster: I think we want to use a selector `getSkipPausing(getState())`

ruchikabgosain: used the selectors, let me know if this works :) @jasonLaster @darkwing @digitarald

janodvarko:
@grlwholifts - I am seeing that you already fixed https://bugzilla.mozilla.org/show_bug.cgi?id=1555333, so I am closing this PR. But, please let me know if there is something in the patch we should land.

Note that Firefox Developer Tools (including Debugger tool) development is now entirely happening in Bugzilla and new issues should be reported here: https://bugzilla.mozilla.org/enter_bug.cgi?product=DevTools&component=Debugger

Honza

Issue #8258, opened 2019-08-16
Pass the object to onDOMNodeMouseOut in the reps


gabrielluong:
For fission, we will need to be able to retrieve the contextual fronts based on the grip's target (See [Bug 1574506](https://bugzilla.mozilla.org/show_bug.cgi?id=1574506). We currently do pass the grip object in onDOMNodeMouseOver and onDOMNodeClick. We will need to do to the same for onDOMNodeMouseOut.

gabrielluong:
Gonna move this to bugzilla.

PR #8259, opened 2019-08-16
Pass the object to onDOMNodeMouseOut in the reps (#8258)

gabrielluong:
Fixes #98258

### Summary of Changes

* Passes the object to onDOMNodeMouseOut in the reps


gabrielluong:
Moving this to bugzilla and phabricator

